PR #81, opened 2017-10-10
WIP - (BOLT-163) Configurable connection timeouts

grimradical:


lib/bolt/node/winrm.rb
```diff
@@ -9,9 +9,14 @@ def initialize(host, port, user, password, shell: :powershell, **kwargs)
...
       @shell = shell
       @endpoint = "http://#{host}:#{port}/wsman"
-      @connection = ::WinRM::Connection.new(endpoint: @endpoint,
-                                            user: @user,
-                                            password: @password)
+      options = { endpoint: @endpoint,
+                  user: @user,
+                  password: @password }
+      if @timeout then
+        options[:receive_timeout] = @timeout
```
Iristyle: So far this is not working as expected... working to further investigate WinRM protocol and the gems behavior.

grimradical:
Okay i think what I'm going to do with this is split things up...I believe I can get the basic connection timeouts (establishing an initial connection with ssh/winrm targets), but anything beyond that (in particular, winrm's command timeout) seems tricky/unreliable enough that I'm inclined to make a separate ticket for those. Stay tuned!

Iristyle:
Yeah, separate effort around winrm sounds fine.  I'm still looking into the specifics there.

Iristyle:
Cleaned up / working in #157

PR #82, opened 2017-10-11
(BOLT-152) Default to secure ssh authentication

joshcooper:
This commit modifies bolt to use the `secure` verifier by default. The
connection will be allowed if the remote host has an entry in our known_hosts
file and the key presented by the host matches our local copy.

Add a command line option (either -k or --insecure) to fallback to the old
lenient behavior. The other net-ssh verifiers `null` and `strict` are not
supported.

### DIFF
--- a/lib/bolt/cli.rb
+++ b/lib/bolt/cli.rb
@@ -130,6 +130,11 @@ HELP
                 "Parameters to a task or plan") do |params|
           results[:task_options] = parse_params(params)
         end
+        results[:insecure] = false
+        opts.on('-k', '--insecure',
+                "Whether to connect insecurely ") do |insecure|
+          results[:insecure] = insecure
+        end
         opts.on_tail('--[no-]tty',
                      "Request a pseudo TTY on nodes that support it") do |tty|
           results[:tty] = tty
@@ -285,7 +290,7 @@ HELP
       else
         nodes = options[:nodes].map do |node|
           Bolt::Node.from_uri(node, options[:user], options[:password],
-                              tty: options[:tty])
+                              tty: options[:tty], insecure: options[:insecure])
         end
 
         results = nil
--- a/lib/bolt/node/errors.rb
+++ b/lib/bolt/node/errors.rb
@@ -1,0 +1,22 @@
+module Bolt
+  class Node
+    class BaseError < StandardError
+      attr_reader :issue_code
+
+      def initialize(message, issue_code)
+        super(message)
+        @issue_code = issue_code
+      end
+
+      def kind
+        'puppetlabs.tasks/node-error'
+      end
+    end
+
+    class ConnectError < BaseError
+      def kind
+        'puppetlabs.tasks/connect-error'
+      end
+    end
+  end
+end
--- a/lib/bolt/node/ssh.rb
+++ b/lib/bolt/node/ssh.rb
@@ -6,13 +6,26 @@ require 'bolt/node/result'
 module Bolt
   class SSH < Node
     def connect
-      options = { logger: @transport_logger,
-                  non_interactive: true }
+      options = {
+        logger: @transport_logger,
+        non_interactive: true
+      }
+
       options[:port] = @port if @port
       options[:password] = @password if @password
+      options[:verify_host_key] = if @insecure
+                                    Net::SSH::Verifiers::Lenient.new
+                                  else
+                                    Net::SSH::Verifiers::Secure.new
+                                  end
 
       @session = Net::SSH.start(@host, @user, options)
       @logger.debug { "Opened session" }
+    rescue Net::SSH::HostKeyError => e
+      raise Bolt::Node::ConnectError.new(
+        "Host key verification failed for #{@uri}: #{e.message}",
+        'HOST_KEY_ERROR'
+      )
     end
 
     def disconnect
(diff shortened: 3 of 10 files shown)


lib/bolt/cli.rb
```diff
@@ -130,6 +130,10 @@ def create_option_parser(results)
                 "Parameters to a task or plan") do |params|
           results[:task_options] = parse_params(params)
         end
+        opts.on('-k', '--insecure',
+                "Whether to connect insecurely ") do |insecure|
+          Bolt.config[:insecure] = insecure
+        end
```
joshcooper: need to add a cli test

lib/bolt/result.rb
```diff
@@ -25,6 +25,16 @@ def success?
     end
   end
 
+  class ErrorResult < Result
+    def to_h
+      { 'value' => @message }
+    end
```
joshcooper: This seems to be sufficient for the `plan` case:

```
$ bundle exec bolt plan run sample::single_task nodes=qkjbgx7gwjt2dqj.delivery.puppetlabs.net --modules spec/fixtures/modules/
ExecutionResult({'qkjbgx7gwjt2dqj.delivery.puppetlabs.net' => 'Host key verification failed for qkjbgx7gwjt2dqj.delivery.puppetlabs.net: fingerprint ac:8c:26:9b:20:d6:71:2d:cd:a9:1f:40:cb:a9:1f:98 is unknown for "qkjbgx7gwjt2dqj.delivery.puppetlabs.net,10.32.123.243"'})
```

But I don't know if that's sufficient for all cases?

pcarlisle: but we want it to be an error, this is making it a success in puppet

pcarlisle: should be something like
```ruby
{
  'error' => {
    'kind' => 'puppetlabs.tasks/connection-error',
    'issue_code' => 'HOST_KEY_SOMETHING',
    'msg' => "Host key verification failed",
    'details' => { ?? }
  }
}
```

joshcooper: The `success?` method returns false below, ah but right, the hash contents need to be error like. Does it need to contain `_error` and `_output`, or only `_error`?

joshcooper: ah, right you were talking about the issue code and parameterizing it.

lib/bolt/node/exception.rb
```diff
@@ -0,0 +1,5 @@
+module Bolt
+  class Node
+    class Exception < RuntimeError; end
```
pcarlisle: the name should indicate what this is for somehow

lib/bolt/node/ssh.rb
```diff
@@ -6,13 +6,24 @@
...
+        non_interactive: true
+      }
+
       options[:port] = @port if @port
       options[:password] = @password if @password
+      options[:verify_host_key] = if @insecure
+                                    Net::SSH::Verifiers::Lenient.new
+                                  else
+                                    Net::SSH::Verifiers::Secure.new
+                                  end
```
joshcooper: This overrides whatever is in `~/.ssh/config` which I don't think is desirable, e.g. if it contains `StrictHostKeyChecking no` for the host/domain you're trying to connect to.

joshcooper: `net-ssh` doesn't parse the `StrictHostKeyChecking` value for a host, e.g. given:

```
  StrictHostKeyChecking no
```

net-ssh doesn't load the value:

```
Net::SSH::Config.for(@host)
{:auth_methods=>["none", "publickey", "password", "keyboard-interactive"], :user=>"root", :keys=>["~/.ssh/id_rsa", "~/.ssh/josh_puppetlabs"], :send_env=>[/^LANG$/, /^LC_.*$/]}
```

joshcooper:
The plan result now indicates an error:

```
$ bx bolt plan run sample::single_task nodes=jgpua7467pa7jqz.delivery.puppetlabs.net --modules spec/fixtures/modules
ExecutionResult({'jgpua7467pa7jqz.delivery.puppetlabs.net' => Error({'message' => 'Host key verification failed for jgpua7467pa7jqz.delivery.puppetlabs.net: fingerprint ac:8c:26:9b:20:d6:71:2d:cd:a9:1f:40:cb:a9:1f:98 is unknown for "jgpua7467pa7jqz.delivery.puppetlabs.net,10.32.124.18"', 'kind' => 'puppetlabs.tasks/connect-error', 'issue_code' => 'NODE_ERROR'})})
```

pcarlisle:
I think the error code is supposed to be more specific than the kind. I was under the impression that it should be specific to the exact error, equivalent to the message but not subject to interpolation or translation.

@hlindberg

joshcooper:
I switched `issue_code` to match the exception class `CONNECT_ERROR`, while the `kind` remains more generic as `puppetlabs.tasks/node-error`.

lib/bolt/node/errors.rb
```diff
@@ -0,0 +1,19 @@
...
+      end
+
+      def issue_code
+        'NODE_ERROR'
+      end
+    end
+
+    class ConnectError < BaseError
+      def issue_code
+        'CONNECT_ERROR'
```
joshcooper: make this more specific, e.g. AUTH_ERROR, HOST_KEY_ERROR, etc

PR #89, opened 2017-10-11
(BOLT-173) empty nodes value results in unclear errors

shawnferry:
OptionParse consumes the next option resulting in unclear erorrs.

This provides a simple solution to the problem.

### DIFF
--- a/lib/bolt/cli.rb
+++ b/lib/bolt/cli.rb
@@ -107,6 +107,15 @@ HELP
           '* protocol is `ssh` by default, may be `ssh` or `winrm`',
           '* port is `22` by default for SSH, `5985` for winrm (Optional)'
         ) do |nodes|
+          # Reject simple mistakes in nodes. Failures are unclear when the
+          # arguemnt to the next (consumed) option is reported as invalid
+          #
+          # simple checking has to happen here instead of validate
+          nodes.each do |node|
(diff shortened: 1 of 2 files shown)


lib/bolt/cli.rb
```diff
@@ -107,6 +107,15 @@ def create_option_parser(results)
           '* protocol is `ssh` by default, may be `ssh` or `winrm`',
           '* port is `22` by default for SSH, `5985` for winrm (Optional)'
         ) do |nodes|
+          # Reject simple mistakes in nodes. Failures are unclear when the
+          # arguemnt to the next (consumed) option is reported as invalid
+          #
+          # simple checking has to happen here instead of validate
+          nodes.each do |node|
```
MikaelSmith: This doesn't appear to be a problem exclusive to `--nodes`. It seems like it applies to any option that expects a value. Hard-coding other CLI short args is a fragile way to address this.

shawnferry: Agreed. I'm not sure how to address it.

It seems like an OptionParse issue not first tokenizing the entire string to find arguments. Without making a fix in the core it may only end up being somewhat messy.

MikaelSmith:
We've decided not to take this direction for improving error message, as it is somewhat error-prone and represents a substantial deviation from standard CLI parsing. Some work has been done to make error messages clearer where possible.

PR #90, opened 2017-10-11
(BOLT-29) Allow hosts to be specified in file or stdin

joshcooper:
Allow hosts to be specified on stdin `--nodes -`, or from a file,
`--nodes @file.txt`. In both cases, the nodes should be specified
one node URI per line. Leading and trailing spaces will be
stripped. Bolt still accepts comma-separated list of nodes.

### DIFF
--- a/lib/bolt/cli.rb
+++ b/lib/bolt/cli.rb
@@ -98,16 +98,18 @@ HELP
     def create_option_parser(results)
       OptionParser.new('') do |opts|
         opts.on(
-          '-n', '--nodes x,y,z', Array,
+          '-n', '--nodes NODES',
           'Node(s) to connect to in URI format [protocol://]host[:port]',
           'Eg. --nodes bolt.puppet.com',
           'Eg. --nodes localhost,ssh://nix.com:2222,winrm://windows.puppet.com',
           "\n",
+          '* NODES can either be comma-separated, \'@<file>\' to read',
+          '* nodes from a file, or \'-\' to read from stdin',
           '* Windows nodes must specify protocol with winrm://',
           '* protocol is `ssh` by default, may be `ssh` or `winrm`',
           '* port is `22` by default for SSH, `5985` for winrm (Optional)'
         ) do |nodes|
-          results[:nodes] = nodes
+          results[:nodes] = parse_nodes(nodes)
         end
         opts.on('-u', '--user USER',
                 "User to authenticate as (Optional)") do |user|
@@ -210,6 +212,11 @@ HELP
                         BANNER
                       end
       puts parser.help
+    end
+
+    def parse_nodes(nodes)
+      list = get_arg_input(nodes)
+      list.split(/[[:space:],]+/).reject(&:empty?)
     end
 
     def parse_params(params)
--- a/spec/bolt/cli_spec.rb
+++ b/spec/bolt/cli_spec.rb
@@ -59,6 +59,38 @@ describe "Bolt::CLI" do
       expect(cli.parse).to include(nodes: %w[foo bar])
     end
 
+    it "reads from stdin when --nodes is '-'" do
+      nodes = <<NODES
+foo
+bar
+NODES
+      cli = Bolt::CLI.new(%w[command run --nodes -])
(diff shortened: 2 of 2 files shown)


MikaelSmith:
Seems sane. Going to test it out.

lib/bolt/cli.rb
```diff
@@ -212,6 +214,15 @@ def print_help(mode)
       puts parser.help
     end
 
+    def parse_nodes(nodes)
+      list = get_arg_input(nodes)
+      if list == nodes
+        list.split(/[, ]+/)
+      else
+        list.split(/[\n]+/)
```
MikaelSmith: Why avoid splitting on commas from a file? Don't like csv-style?

MikaelSmith:
The fact that `bolt command run hostname --nodes "host1 host2"` works is kind of funny, but I can't think of how it would cause any problems.

lib/bolt/cli.rb
joshcooper: I was thinking commas could appear in the URI and we wouldn't want to split on them, but the same is true for line 220. I could probably collapse the regex and always split on whitespace, `\n` or commas.

MikaelSmith: puppet-task reads the file, then splits whatever list it ended up with on `[[:space:],]+`. So splitting on all whitespace and commas would be consistent with it.

MikaelSmith: But ugh, commas are valid parts of a hostname. I'd make them consistent, and if it comes up we can do something to address it in both places.

lib/bolt/cli.rb
```diff
@@ -98,16 +98,18 @@ def initialize(argv)
     def create_option_parser(results)
       OptionParser.new('') do |opts|
         opts.on(
-          '-n', '--nodes x,y,z', Array,
+          '-n', '--nodes NODES',
           'Node(s) to connect to in URI format [protocol://]host[:port]',
           'Eg. --nodes bolt.puppet.com',
           'Eg. --nodes localhost,ssh://nix.com:2222,winrm://windows.puppet.com',
           "\n",
+          '* NODES can either be comma-separated, \'@<file>\' to read',
```
Iristyle: `@` has special meaning in PowerShell, so I would prefer to avoid it here if possible.

joshcooper: We have the same issue with parameters, e.g. `--params @params.json`. I was going to keep this consistent, and fix both in the other ticket.

lib/bolt/cli.rb
jeffmccune: > Why avoid splitting on commas from a file? Don't like csv-style?

FWIW most unix tools are line-oriented and it's generally speaking more difficult to work with csv-style output.  The general use case I expect to work is something along the lines of:

    grep dev nodes.txt | bolt ...

joshcooper: I tried using `[:space]`, but that didn't work:

```
irb(main):001:0> "  foo\nbar  \nbaz\n".split(/[[:space],]+/)
=> ["  foo\nb", "r  \nb", "z\n"]
```

So instead I went with `\s`, but not the first empty string:

```
irb(main):002:0> "  foo\nbar  \nbaz\n".split(/[\s,]+/)
=> ["", "foo", "bar", "baz"]
```

Hence the need to reject empty strings.

lib/bolt/cli.rb
```diff
@@ -212,6 +214,11 @@ def print_help(mode)
       puts parser.help
     end
 
+    def parse_nodes(nodes)
+      list = get_arg_input(nodes)
+      list.split(/[[:space:],]+/).map(&:strip).reject(&:empty?)
```
pcarlisle: this strip should be unnecessary with this regex

joshcooper: IIRC the split doesn’t handle leading spaces for the first entry or trailing spaces for the last entry. I could probably strip before splitting though.

joshcooper: nah, you're right, `strip` isn't needed.

PR #93, opened 2017-10-12
(BOLT-162) Add scaffolding for beaker acceptance

johnduarte:
Adds basic configuration for running beaker acceptance tests against `bolt`.


acceptance/setup/gem/pre-suite/020_install.rb
```diff
@@ -0,0 +1,19 @@
+gem_source = ENV['GEM_SOURCE'] || "https://rubygems.org"
+gem_version = ENV['BOLT_GEM'] || "> 0.1.0"
```
joshcooper: Now that bolt has been released publicly, we can drop the "> 0.1.0" part.

acceptance/README.md
```diff
@@ -0,0 +1,208 @@
...
+Example of running test suite using gem install using an existing hosts file:
+```
+SSH_USER=root                      \
+SSH_PASSWORD='S3@ret3'             \
+WINRM_USER=Administator            \
+WINRM_PASSWORD='S3@ret3'           \
+  bundle exec beaker               \
+      -o config/gem/options.rb     \
+      -h hosts.yaml                \
+      -t tests
```
joshcooper: Can we use beaker's subcommands, e.g. `beaker init`, `beaker provision`, etc?

johnduarte: Yes, the downside is that unless the user wanted to export the password variables they would need to pass them to the pre-suite call and the test run. So, the above command would transform into:

1. Initialize beaker with supplied options file for gem configuration.
```
bundle exec beaker init -h hosts.yaml -o config/gem/options.rb
```
2. Provision hosts
```
bundle exec beaker provision
```
3. Run pre-suite
```
SSH_USER=root                      \
SSH_PASSWORD='S3@ret3'             \
WINRM_USER=Administator            \
WINRM_PASSWORD='S3@ret3'           \
  bundle exec beaker exec pre-suite
```
4. Run tests with specified environment variables
```
SSH_USER=root                      \
SSH_PASSWORD='S3@ret3'             \
WINRM_USER=Administator            \
WINRM_PASSWORD='S3@ret3'           \
  bundle exec beaker exec ./tests
```
5. Destroy hosts
```
bundle exec beaker destroy
```

acceptance/Gemfile
```diff
@@ -1,21 +1,26 @@
...
 
 def location_for(place, fake_version = nil)
-  if place =~ /^(git:[^#]*)#(.*)/
-    [fake_version, { :git => $1, :branch => $2, :require => false }].compact
-  elsif place =~ /^file:\/\/(.*)/
-    ['>= 0', { :path => File.expand_path($1), :require => false }]
+  git_match = place.match(/^(git:[^#]*)#(.*)/)
+  file_match = place.match(%r{^file://(.*)})
+  if git_match
+    git, branch = git_match.captures
```
pcarlisle: this is off by one, the first element of captures is the whole string, so you want git to be git_match[1] and branch to be git_match[2]

acceptance/Gemfile
```diff
@@ -1,21 +1,26 @@
...
+    [fake_version, { git: git, branch: branch, require: false }].compact
+  elsif file_match
+    file_path = file_match.captures
```
pcarlisle: this assigns the whole array of matches as the path, should be file_match[1]

acceptance/Gemfile
pcarlisle: oh this is wrong captures takes care of that

acceptance/tests/command.rb
```diff
@@ -0,0 +1,15 @@
+test_name "C100546: bolt command run should execute command on remote hosts\
+           via ssh" do
+  step "execute `bolt command run` via SSH" do
+    ssh_nodes = select_hosts(roles: ['ssh'])
+    nodes_csv = ssh_nodes.map(&:hostname).join(',')
+    command = 'hostname -f'
+    bolt_command = "bolt command run --nodes #{nodes_csv} '#{command}'"
+    case bolt['platform']
+    when /windows/
+      execute_powershell_script_on(bolt, bolt_command)
```
joshcooper: I think this starts a new `powershell.exe` interpreter each time, which is pretty slow to run a single command. Can we just do `on(bolt, "cmd /c #{bolt_command}")`? /cc @Iristyle

Iristyle: Yes that starts a new `powershell.exe` and yes it's slow...

However, it also simulates closer to what we expect end users to be doing, which may impact argument passing conventions, etc.  I would assume that's desireable here?

johnduarte: The desire is to simulate what we have instructed the user to perform. I think this performance hit is reasonable to endure at the moment. If it is determined not to add value later it can be changed.

acceptance/Rakefile
```diff
@@ -0,0 +1,111 @@
...
+      task.add_env(name: 'GIT_BRANCH', default: 'master')
+      task.add_env(name: 'GIT_SHA',    default: '')
+      task.add_env(name: 'SSH_USER',   default: 'root')
+      task.add_env(name: 'SSH_PASSWORD')
+      task.add_env(name: 'WINRM_USER', default: 'Administrator')
+      task.add_env(name: 'WINRM_PASSWORD')
+      task.add_command do |cmd|
+        cmd.name = 'bundle exec beaker'
+        cmd.add_option(*beaker_options)
+        cmd.add_option(beaker_keyfile) if ENV['BEAKER_KEYFILE']
```
joshcooper: I copied `id_rsa-acceptance*` to the cwd, but beaker was unable to ssh to the host as root.

```
Spent 0.02 seconds tagging VMs
  Warning: Beaker does not support vmhostname to SSH to host, trying next available method.
  Warning: Skipping ip method to ssh to host as its value is not set. Refer to https://github.com/puppetlabs/beaker/tree/master/docs/how_to/ssh_connection_preference.md to remove this warning
  Warning: Beaker does not support hostname to SSH to host, trying next available method.
  Warning: Skipping ip method to ssh to host as its value is not set. Refer to https://github.com/puppetlabs/beaker/tree/master/docs/how_to/ssh_connection_preference.md to remove this warning
:paranoid is deprecated, please use :verify_host_key. Supported values are exactly the same, only the name of the option has changed.
  Warning: Try 1 -- Host lepf9q0hhyp5jri.delivery.puppetlabs.net unreachable: Net::SSH::AuthenticationFailed - Authentication failed for user root@lepf9q0hhyp5jri.delivery.puppetlabs.net
  Warning: Trying again in 3 seconds
:paranoid is deprecated, please use :verify_host_key. Supported values are exactly the same, only the name of the option has changed.
...
```

I deleted the pub/private keys, and ran with `BEAKER_KEYFILE=~/.ssh/id_rsa-acceptance`, which did work. Could you add `BEAKER_KEYFILE` to the list of environment variables that may need setting?

acceptance/Rakefile
```diff
@@ -0,0 +1,111 @@
...
+  # set. This is a rototiller bug QA-2761.
+  beaker_keyfile = {
+    name: '--keyfile',
+    add_argument: {
+      add_env: {
+        name: 'BEAKER_KEYFILE',
+        message: 'The private SSH key beaker will use to connect to SUTs'
+      }
+    }
+  }
```
joshcooper: Can we run with `--debug` enabled by default? I like to see what commands are being executed, not just:

```
Install Bolt via git

  * Clone repo

  * Update submodules

  * Construct version based on SHA
    localhost $ scp /var/folders/8m/d6146zxd6bv_3h8y4ykvvxh40000gn/T/beaker20171101-62411-1a7ivt0 debian9-64-1:bolt/lib/bolt/version.rb {:ignore => }

  * Build gem

  * Install custom gem

  * Ensure install succeeded
```

johnduarte:
@joshcooper This PR has been updated to make the `BEAKER_KEYFILE` a required param and to add `BEAKER_LOG_LEVEL` with a default of `debug`.

johnduarte:
Rebasing on 0.7.0

PR #96, opened 2017-10-12
Add a uniq function to nodes array to ensure we deduplicate list

cyberious:
In the event we execute a list that happens to have a node twice the command will be executed twice, simply running a uniq on the array to ensure it is a unique set.

### DIFF
--- a/lib/bolt/cli.rb
+++ b/lib/bolt/cli.rb
@@ -216,7 +216,7 @@ HELP
 
     def parse_nodes(nodes)
       list = get_arg_input(nodes)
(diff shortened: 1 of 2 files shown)


cyberious:
Resolves issue #97

genebean:
This seems like a very useful addition as I have been putting node lists in files and using them via

```
--nodes `cat nodelist`
```

joshcooper:
In the future, can we add BOLT tickets to the subject line so we can reconcile tickets & commits.

PR #101, opened 2017-10-12
Add new connection scheme `local` that allows to execute command on host that executes bolt

riton:
This is still a work in progress and has not been heavily tested.

I'm opening this issue to open a discussion about the idea of a `local://` scheme.

Example of what it does:

```
riton@myhost $ bundle exec bolt command run -n local://localhost whoami
localhost:

riton

Ran on 1 node in 0.00 seconds
```

### DIFF
--- a/lib/bolt/node/local.rb
+++ b/lib/bolt/node/local.rb
@@ -1,0 +1,117 @@
+require 'json'
+require 'open3'
+require 'bolt/node/result'
+
+module Bolt
+  class Local < Node
+    def connect
+      # this is a local connection, just passing there...
+    end
+
+    def disconnect
+      # this is a local connection, just passing there...
+    end
+
+    def execute(command, options = {})
+      result_output = Bolt::Node::ResultOutput.new
+      status = {}
+
+      @logger.debug { "Executing: #{command}" }
+
+      if options[:stdin]
+        stdout, stderr, rc = Open3.capture3(command, :stdin_data => options[:stdin])
+      else
+        stdout, stderr, rc = Open3.capture3(command)
+      end
+
+      result_output.stdout << stdout.strip unless stdout.nil?
+      @logger.debug { "stdout: #{data}" }
+
+      result_output.stderr << stderr.strip unless stderr.nil?
+      @logger.debug { "stderr: #{data}" }
+
+      status[:exit_code] = rc.to_i
+
+      if status[:exit_code].zero?
+        @logger.debug { "Command returned successfully" }
+        Bolt::Node::Success.new(result_output.stdout.string, result_output)
+      else
+        @logger.info { "Command failed with exit code #{status[:exit_code]}" }
+        Bolt::Node::Failure.new(status[:exit_code], result_output)
+      end
+    end
+
+    def _upload(source, destination)
+      FileUtils.copy_file(source, destination)
+      Bolt::Node::Success.new
+    rescue StandardError => e
+      Bolt::Node::ExceptionFailure.new(e)
+    end
+
+    def make_tempdir
+      # Note(riton): Should be developed to support os that does not have mktemp
+      Bolt::Node::Success.new(`mktemp -d`.chomp)
+    rescue StandardError => e
+      Bolt::Node::ExceptionFailure.new(e)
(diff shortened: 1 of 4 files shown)


riton:
Ok, this seems to work with `task`, `plan` and `command`. 

Currently, with my scheme parsing, `local://localhost` and `local://whatever` will both be considered as _run on local machine_. This is not that pretty, but I'm waiting input from you to discuss more about that.

lib/bolt/node/local.rb
```diff
@@ -0,0 +1,117 @@
...
+    def _upload(source, destination)
+      FileUtils.copy_file(source, destination)
+      Bolt::Node::Success.new
+    rescue StandardError => e
+      Bolt::Node::ExceptionFailure.new(e)
+    end
+
+    def make_tempdir
+      # Note(riton): Should be developed to support os that does not have mktemp
+      Bolt::Node::Success.new(`mktemp -d`.chomp)
```
riton: This will lead to issues if the execution host is _windows_.

lib/bolt/node/local.rb
```diff
@@ -0,0 +1,117 @@
...
+      result = nil
+
+      make_tempdir.then do |value|
+        dir = value
+        local_path = "#{dir}/#{File.basename(file)}"
+        Bolt::Node::Success.new
+      end.then do
+        _upload(file, local_path)
+      end.then do
+        execute("chmod u+x '#{local_path}'")
```
riton: This will lead to issues if the execution host is _windows_.

genebean:
I really like the idea of a local scheme but why not just add a `--local` flag to supplement `--nodes`? It would be much cleaner and would allow for easily using one or the other or both at the same time.

garethr:
+1 to @genebean suggestion. I was thinking something similar. The other option was simply reserving the node name `local` or `localhost` in it's plain form. ie.

```
--nodes node1,node2,localhost
```

If someone really has a remote node called `localhost` (or really wants to access it over SSH) then you could force it using the protocol prefix:

```
--nodes node1,node2,ssh://localhost
```

The reason I prefer this more than the flag is because I may want to store the nodes list in a file, or have a script generate it, and I'd like to be able to include `localhost` in those cases. Making it a flag makes that much harder to do.

The `local://` schema prefix works, but I think feels overly verbose. How often would it not be `local://localhost`? What is the expected behaviour when the host portion doesn't match the hostname of the local host? What happens if you specify multiple `local://` nodes with different hostnames? etc.

adreyer:
--local wouldn't work from plans.

thrnio:
@garethr bare `localhost` should probably be okay, generally. From RFC 6761:

> **6.3.  Domain Name Reservation Considerations for "localhost."**
>
>   The domain "localhost." and any names falling within ".localhost."
   are special in the following ways:
>
>   1.  Users are free to use localhost names as they would any other
       domain names.  Users may assume that IPv4 and IPv6 address
       queries for localhost names will always resolve to the respective
       IP loopback address.
>
>   2.  Application software MAY recognize localhost names as special, or
       MAY pass them to name resolution APIs as they would for other
       domain names.

genebean:
I like @garethr of a reserved name `localhost` that would default to a local transport

nicklewis:
I think this is a great idea, and it makes total sense to me to treat `localhost` specially. @riton Are you still interested in working on this? I'm happy to help however I can!

riton:
@nicklewis yes I am. I plan to work on this feature next week.

I'd be glad to have opinions about the Unix specific stuff like `chmod` and how to _handle_ them properly if Windows hosts has to be supported.

lib/bolt/node/local.rb
MikaelSmith: The winrm transport may provide some hints, and Ruby has methods to detect Windows or non-Windows OS. `Gem.win_platform?` seems reasonable

MikaelSmith:
I'm working on a new version of this I should have up today or tomorrow. I'll close this PR once it's ready.

MikaelSmith:
Closing in favor of https://github.com/puppetlabs/bolt/pull/317.

Issue #100, opened 2017-10-12
bolt command line: add to nodes when --nodes is passed multiple times


jjtoth:
If I run this:

` bolt command run --nodes=web5 --nodes=web6 'rpm -q puppet' `

it only runs on the last host (or hosts given): "web6" in this case.

Motiviation is that i would use it in combination with this shell pattern:

`bolt command run --nodes=web{5,6,7,8}.mydomain.edu 'echo hi, this is bolt'`

or, even more likely:

`bolt command run --nodes={web{5,6,7,8},elasticsearch{1,2,3}.subdomain}.mydomain.edu 'echo hi, this is bolt'`

which would expand to:

`bolt command run --nodes=web5.mydomain.edu --nodes=web6.mydomain.edu --nodes=web7.mydomain.edu --nodes=web8.mydomain.edu --nodes=elasticsearch1.subdomain.mydomain.edu --nodes=elasticsearch2.subdomain.mydomain.edu --nodes=elasticsearch3.subdomain.mydomain.edu 'echo hi, this is bolt'`

jjtoth:
Oops, didn't mean to close it.

cyberious:
@jjtoth PR #103 was just merged to resolve this issue.

joshcooper:
Thanks @jjtoth. This should now be fixed (give bolt#master a try).

PR #103, opened 2017-10-12
Allow --nodes to be declared multiple times

cyberious:
- Allow for the user to declare --nodes multiple times along a cli and merge them   into a single array, previously it would allow for last nodes declaration to win and miss any previous   calls on cli

### DIFF
--- a/lib/bolt/cli.rb
+++ b/lib/bolt/cli.rb
@@ -91,8 +91,10 @@ HELP
     attr_accessor :options
 
     def initialize(argv)
-      @argv = argv
-      @options = {}
+      @argv    = argv
+      @options = {
+        nodes: []
+      }
 
       @parser = create_option_parser(@options)
     end
@@ -111,7 +113,8 @@ HELP
           '* protocol is `ssh` by default, may be `ssh` or `winrm`',
           '* port is `22` by default for SSH, `5985` for winrm (Optional)'
         ) do |nodes|
-          results[:nodes] = parse_nodes(nodes)
+          results[:nodes] += parse_nodes(nodes)
+          results[:nodes].uniq!
         end
         opts.on('-u', '--user USER',
                 "User to authenticate as (Optional)") do |user|
@@ -283,7 +286,7 @@ HELP
               "unknown argument(s) #{options[:leftovers].join(', ')}"
       end
 
-      unless options[:nodes] || options[:mode] == 'plan'
+      unless !options[:nodes].empty? || options[:mode] == 'plan'
         raise Bolt::CLIError, "option --nodes must be specified"
       end
 
--- a/spec/bolt/cli_spec.rb
+++ b/spec/bolt/cli_spec.rb
@@ -73,6 +73,11 @@ describe "Bolt::CLI" do
       expect(cli.parse).to include(nodes: %w[foo bar])
     end
 
+    it "accepts multiple nodes across multiple declarations" do
+      cli = Bolt::CLI.new(%w[command run --nodes foo,bar --nodes bar,more,bars])
+      expect(cli.parse).to include(nodes: %w[foo bar more bars])
+    end
+
     it "reads from stdin when --nodes is '-'" do
       nodes = <<NODES
 foo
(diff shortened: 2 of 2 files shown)


cyberious:
Resolves #100

lib/bolt/cli.rb
```diff
@@ -109,7 +109,8 @@ def create_option_parser(results)
           '* protocol is `ssh` by default, may be `ssh` or `winrm`',
           '* port is `22` by default for SSH, `5985` for winrm (Optional)'
         ) do |nodes|
-          results[:nodes] = parse_nodes(nodes)
+          results[:nodes] = [] unless results[:nodes]
+          results[:nodes] = results[:nodes] + parse_nodes(nodes)
```
joshcooper: This could cause a regression on `uniq`ness as implemented in https://github.com/puppetlabs/bolt/pull/96.

A more idiomatic ruby would be to:

```
parsed = parse_nodes(nodes)
if parsed
  results[:nodes] += parsed
else
  results[:nodes] = []
end
```

cyberious: The problem would be you are assigning the result a nil list if not parsed again, so if it fails the last parse it will than give an empty list?

cyberious: Also at the point of parsing the results[:nodes] would be nil

cyberious: Also now that I am looking more, any reason not to switch to a struct style options parser pattern?

cyberious:
modified to have the options default to empty node list and validate that it is indeed not empty or fail accordingly

cyberious:
I will get back to fix my whitespace issues from my IDE but syntactic review and will goback on other

lib/bolt/cli.rb
```diff
@@ -109,7 +111,10 @@ def create_option_parser(results)
           '* protocol is `ssh` by default, may be `ssh` or `winrm`',
           '* port is `22` by default for SSH, `5985` for winrm (Optional)'
         ) do |nodes|
-          results[:nodes] = parse_nodes(nodes)
+          parsed = parse_nodes(nodes)
+          if parsed
+            results[:nodes] += parsed
+          end
```
joshcooper: If we pass `--nodes a --nodes a` then we'll get duplicates, which would regress on https://github.com/puppetlabs/bolt/commit/555b4ed95e86b231bd92d293c5ab702792778658. Also `parse_nodes` always returns an array (possible empty), so checking `if parsed` isn't needed.

cyberious: correct, but we are ensuring Uniq in a secondary PR that would do away with that is a separate issue can and can exist even without the secondary nodes flag.

joshcooper:
Thanks @cyberious! Can you squash your commits and resolve conflicts?

cyberious:
Squashed and rebased

spec/bolt/cli_spec.rb
```diff
@@ -73,6 +73,11 @@
       expect(cli.parse).to include(nodes: %w[foo bar])
     end
 
+    it "accepts multiple nodes across multiple declarations" do
+      cli = Bolt::CLI.new(%w[command run --nodes foo,bar --nodes more,bars])
+      expect(cli.parse).to include(nodes: %w[foo bar more bars])
```
joshcooper: Can you add the same node name to both `--nodes` flags to verify the uniqueness is working?

spec/bolt/cli_spec.rb
```diff
@@ -73,6 +73,11 @@
       expect(cli.parse).to include(nodes: %w[foo bar])
     end
 
+    it "accepts multiple nodes across multiple declarations" do
+      cli = Bolt::CLI.new(%w[command run --nodes foo,bar --nodes bar,more,bars])
```
cyberious: @joshcooper just added bar in second location to ensure uniq still works

joshcooper:
Thanks @cyberious. One minor nit, it looks like your commit message has your work email address, but our CLA system only knows about your `neoknght` email address. Could you reset the commit author to your personal email?

cyberious:
Updated to my personal

cyberious:
@joshcooper not merged?

cyberious:
Had to take a double check, showed closed and not merged... not sure what is up with github on this one.

joshcooper:
Thanks @cyberious! I updated the commit message to reference the JIRA ticket and merged in 4b2f04374d96d1756df042e32293eae0e629f8c9

PR #107, opened 2017-10-13
(BOLT-168) Add partial gssapi-with-mic support

joshcooper:
Add support for the `gssapi-with-mic` ssh authentication method. To
use it, `gssapi-with-mic` must be specified as a supported method in
your ssh config, e.g.

    PreferredAuthentications gssapi-with-mic,publickey,password

And the `net-ssh-krb` gem must be installed. Note bolt does not support
all of the GSSAPI configuration options such as GSSAPIAuthentication,
GSSAPITrustDNS.

The net-ssh-krb gem cannot be loaded on Windows. So ignore the LoadError
on Windows, as that is expected. Otherwise log a debug warning.


riton:
Awesome. Thanks. As soon as it's merged and released, I'll be able to drop
the ugly hack we're doing in [ccin2p3/bolt](
https://forge.puppet.com/ccin2p3/bolt) and then eventually to eventually
drop the module completely ☺️

joshcooper:
Added `net-ssh-krb` gem to stickler so `gem install bolt` works internally.

joshcooper:
Crap, net-ssh-krb is GPLv2. I'm not sure we can add a hard dependency like this.

lib/bolt/executor.rb
```diff
@@ -1,16 +1,26 @@
...
+      @logger.level = config[:log_level]
+      @logger.formatter = Bolt::Formatter.new
+      @initialized = Set.new
     end
 
     def from_uris(nodes)
       nodes.map do |node|
-        Bolt::Node.from_uri(node, config: @config)
+        result = Bolt::Node.from_uri(node, config: @config)
+        initialize_transport(result.class)
```
joshcooper: Move this to `execute` so that `from_uris` doesn't have side-effects. Also makes more sense if we wanted to add `dispose_transport` lifecycle method in the future.

lib/bolt/executor.rb
```diff
@@ -72,5 +84,15 @@ def results_to_hash(results)
       results.each_pair { |k, v| result_hash[k] = v }
       result_hash
     end
+
+    def initialize_transport(transport)
+      unless @initialized.include?(transport)
+        @initialized.add(transport)
```
joshcooper: Just do nodes.map(&:class).uniq.each { |transport| transport.initialize_transport(@logger) }

lib/bolt/executor.rb
```diff
@@ -72,5 +84,15 @@ def results_to_hash(results)
...
+
+        if transport.respond_to?(:initialize_transport)
```
joshcooper: Define an empty `initialize_transport` method on the Node class.

lib/bolt/executor.rb
joshcooper: Done

lib/bolt/executor.rb
joshcooper: Done

lib/bolt/node.rb
```diff
@@ -27,11 +27,12 @@ def self.from_uri(uri_string, **kwargs)
                 **kwargs)
     end
 
+    def self.initialize_transport(_logger); end
```
joshcooper: @adreyer I didn't change `Bolt::Node::Orch` to load its config and create the orch client, because I have no idea how to test those changes.

adreyer: it's currently just tested in the client library. I'm ok not moving it for now.

PR #108, opened 2017-10-13
(BOLT-124) Use Addressable to parse URIs

joshcooper:
Ruby's URI class does not correctly parse special characters in the `userinfo`
section of a URI. Use Addressable instead. Note it requires ruby 2.0 and
up, which is compatible with Bolt.

### DIFF
--- a/lib/bolt/node_uri.rb
+++ b/lib/bolt/node_uri.rb
@@ -5,18 +5,13 @@ module Bolt
     end
 
     def parse(string)
-      case string
-      when %r{^(ssh|winrm|pcp)://.*:\d+$}
-        URI(string)
-      when %r{^pcp://}
-        URI(string)
-      when %r{^(ssh|winrm)://}
-        uri = URI(string)
-        uri.port = 5985 if uri.scheme == 'winrm'
-        uri
-      else
-        URI("ssh://#{string}")
-      end
+      uri = if string =~ %r{^(ssh|winrm|pcp)://}
+              Addressable::URI.parse(string)
+            else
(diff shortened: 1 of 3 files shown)


lib/bolt/node_uri.rb
```diff
@@ -7,15 +9,13 @@ def initialize(string)
...
       case string
       when %r{^(ssh|winrm|pcp)://.*:\d+$}
-        URI(string)
-      when %r{^pcp://}
-        URI(string)
-      when %r{^(ssh|winrm)://}
-        uri = URI(string)
+        Addressable::URI.parse(string)
+      when %r{^(ssh|winrm|pcp)://}
+        uri = Addressable::URI.parse(string)
```
Iristyle: Looks good, but investigating if this can be further collapsed / simplified.

joshcooper: I have a different PR in progress to further simplify. The port vs non-port can be collapsed, but we still have to differentiate between when a scheme is specified vs not.

Iristyle: This works fine:

```ruby
    def parse(string)
      string = "ssh://#{string}" if !string.match?(%r{^(ssh|winrm|pcp)://.*$})
      uri = Addressable::URI.parse(string)
      uri.port = 5985 if (uri.scheme == 'winrm' && uri.port.nil?)

      uri
    end
```

johnduarte:
👍 
Validated that this patch successfully enabled parsing encoded passwords given in a URI for both WinRM and SSH.

PR #117, opened 2017-10-18
(BOLT-183) Apply CLI options when running plans

joshcooper:
Apply CLI options when running a plan, such as `--user`, `--password`, `--tty`, etc.


joshcooper:
This is blocked on https://github.com/puppetlabs/puppet/pull/6286, and bumping the puppet submodule with that change.

Issue #91, opened 2017-10-11
Bolt should support prompting for the password


TJM:
Putting a password on the command line is not a best practice. Perhaps there should be a `-P` option or `-p -` to support prompting for a password.

joshcooper:
Yep, this is filed under BOLT-28

PR #119, opened 2017-10-18
Add a password prompt without echoing it to console when --user is en…

deezx:
…tered but without --password
Issue- https://github.com/puppetlabs/bolt/issues/91

### DIFF
--- a/lib/bolt/cli.rb
+++ b/lib/bolt/cli.rb
@@ -6,6 +6,7 @@ require 'json'
 require 'bolt/node'
 require 'bolt/version'
 require 'bolt/executor'
+require 'io/console'
 
 module Bolt
   class CLIError < RuntimeError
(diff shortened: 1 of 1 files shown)


ripienaar:
isn't is possible that someone might want to not specify a password on purpose? like when they have a ssh agent running or some other form of SSO?

deezx:
You're right. What about adding a new option called "--prompt-password" that will do this instead of checking if username was entered without password ?

joshcooper:
Thanks @deezx! How about making the existing password option take an optional argument, e.g. `--password [PASSWORD]`. If a value is not specified, then prompt like you're doing?

deezx:
Thanks @joshcooper. Closing this PR, created a new one at- https://github.com/puppetlabs/bolt/pull/121

PR #123, opened 2017-10-18
(maint) Pin to rubocop 0.50.0

joshcooper:
Rubocop dropped support for Ruby 2.0 analysis, see commit
https://github.com/bbatsov/rubocop/commit/2f6f3f297cd957def22419d8d70a1afd9410e497.

Pin to rubocop 0.50.0, while fixing rubocop errors that occur with
0.51.0.


joshcooper:
Talked this over with @pcarlisle, merging.

PR #121, opened 2017-10-18
make --password optional and prompt the user if value is missing

deezx:

PR #124, opened 2017-10-18
(BOLT-28) Make --password optional and prompt the user if value is missing

joshcooper:
Supersedes PR #121


joshcooper:
@deezx Thank you for your contribution! This should go out in the next bolt release.

deezx:
@joshcooper Thanks!

PR #125, opened 2017-10-18
(maint) Relax addressable gem constraint

joshcooper:
Puppet requires addressable < 2.5.0, because that version includes a
dependency that doesn't support ruby 1.9. In order to use puppet and
bolt gems together, e.g. when running `puppet script`, we need to relax
bolt's addressable gem dependency.

### DIFF
--- a/bolt.gemspec
+++ b/bolt.gemspec
@@ -23,7 +23,7 @@ Gem::Specification.new do |spec|
 
   spec.required_ruby_version = "~> 2.0"
 
-  spec.add_dependency "addressable", "~> 2.5"
+  spec.add_dependency "addressable", "~> 2.4"
   spec.add_dependency "concurrent-ruby", "~> 1.0"
   spec.add_dependency "net-sftp", "~> 2.0"
   spec.add_dependency "net-ssh", "~> 4.0"


joshcooper:
FYI @hlindberg @thallgren This makes it possible to run `puppet script` and point puppet's `Gemfile.local` at a version of bolt, e.g.

```ruby
gem "bolt", :path => '/Users/josh/work/bolt'
```

Then  you can run:
```
$ bundle update
...
$ bundle exec puppet script -e "notice(run_command('whoami', 'ssh://k2uhn4v8lsa2har'))"
Notice: Scope(): ExecutionResult({'ssh://k2uhn4v8lsa2har' => Error({'message' => 'Host key verification failed for ssh://k2uhn4v8lsa2har: fingerprint 4f:47:b5:d1:27:4a:09:a8:f3:91:40:e5:0b:78:ae:fe is unknown for "k2uhn4v8lsa2har,10.32.118.156"', 'kind' => 'puppetlabs.tasks/connect-error', 'issue_code' => 'HOST_KEY_ERROR'})})
```

Note there's no way currently to pass user/pass/insecure flags to bolt. But this at least makes it possible to iterate on passing credentials from the plan to the bolt executor, e.g. using hiera eyaml.

PR #129, opened 2017-10-20
(BOLT-122) Allow multiple directories for --modules

pcarlisle:


joshcooper:
@ahpook should we rename the `modules` command line option to `modulepath`?

ahpook:
+1 for `modulepath`, it's more consistent with both the functionality and the rest of the puppet ecosystem.

lib/bolt/cli.rb
```diff
@@ -134,8 +134,9 @@ def create_option_parser(results)
                 "(Optional, defaults to 100)") do |concurrency|
           results[:concurrency] = concurrency
         end
-        opts.on('--modules MODULES', "Path to modules directory") do |modules|
-          results[:modules] = modules
+        opts.on('--modulepath MODULES',
+                "List of directories containing modules") do |modulepath|
```
joshcooper: Could you mention that it's separated/delimited by `#{File.PATH_SEPARATOR}` so it shows the correct value if the controller is windows or not?

PR #131, opened 2017-10-23
(BOLT-192) Require at least net-ssh 4.2, but less than 5.0

joshcooper:
The net-ssh library renamed the `paranoid` key to `verify_host_key` in
commit net-ssh#c4fbcdc882[1]. Using the latter key fails if net-ssh < 4.2,
so bump our minimum net-ssh runtime dependency.

[1] https://github.com/net-ssh/net-ssh/commit/c3fbcdc882f6763fa718f6adbd9d1d60eaf04b62


joshcooper:
The other option is to introspect the net-ssh library and use either `paranoid` or `verify_host_key`, but that seems like more trouble than it's worth.

PR #133, opened 2017-10-24
(BOLT-53) Global transport

joshcooper:
Adds a `--transport` command line option for specifying the default transport to use if one is not specified. The default transport is ssh. This is handy for Windows as you don't need to repeat winrm in each node URI:

```
$ bolt command run whoami --nodes={hghi7uvplkq8kki,qksuptpp9h5gunc}.delivery.puppetlabs.net --transport winrm -u Administrator -p
Please enter your password:
qksuptpp9h5gunc.delivery.puppetlabs.net:

qksuptpp9h5gunc\administrator


hghi7uvplkq8kki.delivery.puppetlabs.net:

hghi7uvplkq8kki\administrator


Ran on 2 nodes in 0.65 seconds
```

### DIFF
--- a/lib/bolt/cli.rb
+++ b/lib/bolt/cli.rb
@@ -95,7 +95,8 @@ HELP
       @argv    = argv
       @options = {
         nodes: [],
-        transport: 'ssh'
+        transport: 'ssh',
+        insecure: false
       }
       @parser = create_option_parser(@options)
     end
@@ -147,7 +148,6 @@ HELP
                 "Parameters to a task or plan") do |params|
           results[:task_options] = parse_params(params)
         end
-        results[:insecure] = false
         opts.on('-k', '--insecure',
                 "Whether to connect insecurely ") do |insecure|
           results[:insecure] = insecure
--- a/lib/bolt/cli.rb
+++ b/lib/bolt/cli.rb
@@ -86,6 +86,7 @@ HELP
 
     MODES = %w[command script task plan file].freeze
     ACTIONS = %w[run upload download].freeze
+    TRANSPORTS = %w[ssh winrm pcp].freeze
 
     attr_reader :parser
     attr_accessor :options
@@ -93,9 +94,9 @@ HELP
     def initialize(argv)
       @argv    = argv
       @options = {
-        nodes: []
+        nodes: [],
+        transport: 'ssh'
       }
-
       @parser = create_option_parser(@options)
     end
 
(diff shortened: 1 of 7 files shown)


lib/bolt/node.rb
```diff
@@ -10,7 +10,7 @@ class Node
     ENVIRONMENT_METHODS = %w[both environment].freeze
 
     def self.from_uri(uri_string, **kwargs)
-      uri = NodeURI.new(uri_string)
+      uri = NodeURI.new(uri_string, kwargs[:config][:transport])
```
Iristyle: I guess `uri_string` here could already include the transport like `winrm://`, so you couldn't just concat like this and leave `node_uri.rb` alone?

```ruby
NodeURI.new("#{kwargs[:config][:transport]}://uri_string)
```

joshcooper: right, it's the default transport, only used if the `uri_string` doesn't specify a scheme.

lib/bolt/node_uri.rb
```diff
@@ -2,19 +2,20 @@
...
 module Bolt
   class NodeURI
-    def initialize(string)
-      @uri = parse(string)
+    def initialize(string, transport = 'ssh')
+      @uri = parse(string, transport)
     end
 
-    def parse(string)
+    def parse(string, transport)
```
Iristyle: Not sure how much you care, but this method now allows for calling `parse` with a `transport` that is not `ssh|winrm|pcp`

Prior behavior wasn't really better though, because if you passed something like `http://foo.bar.com` you'd end up parsing `ssh://http://foo.bar.com`

joshcooper: Let me fix that. We shouldn't be hardcoding valid transports here, we just need to know if a scheme was specified or not.

joshcooper: Updated

lib/bolt/cli.rb
```diff
@@ -146,11 +148,15 @@ def create_option_parser(results)
                 "Parameters to a task or plan") do |params|
           results[:task_options] = parse_params(params)
         end
-        results[:insecure] = false
         opts.on('-k', '--insecure',
                 "Whether to connect insecurely ") do |insecure|
           results[:insecure] = insecure
         end
+        opts.on('--transport TRANSPORT', TRANSPORTS,
+                "Specify which transport to use,
```
adreyer: Specify the default transport.

joshcooper: The line wrapping was wrong too, fixed that

PR #134, opened 2017-10-25
(BOLT-52) Allow arguments to be passed to scripts

joshcooper:
Bolt now accepts arguments to scripts. If an argument contains spaces it must be
quoted, e.g. bolt script run script.sh 'arg with spaces'

```
$ bx bolt script run spec/fixtures/scripts/success.sh 'one two' three -n o4np54hfrljr5wo.delivery.puppetlabs.net -k
o4np54hfrljr5wo.delivery.puppetlabs.net:

arg: one two
arg: three
standard out

standard error

Ran on 1 node in 0.79 seconds
```

```
$ bx bolt script run spec/fixtures/scripts/success.ps1 "one two" three -n winrm://mllw3jr0io5r4zo.delivery.puppetlabs.net -k -u Administrator -p
....
mllw3jr0io5r4zo.delivery.puppetlabs.net:

arg: one two
arg: three
standard output



Ran on 1 node in 2.09 seconds
```

### DIFF
--- a/lib/bolt/node/orch.rb
+++ b/lib/bolt/node/orch.rb
@@ -105,12 +105,13 @@ module Bolt
       _run_task(BOLT_MOCK_FILE, 'stdin', params)
     end
 
-    def _run_script(script, _)
+    def _run_script(script, arguments)
       content = File.open(script, &:read)
       content = Base64.encode64(content)
       params = {
         action: 'script',
-        content: content
+        content: content,
+        arguments: arguments
       }
       unwrap_bolt_result(_run_task(BOLT_MOCK_FILE, 'stdin', params))
     end
--- a/lib/bolt/node/ssh.rb
+++ b/lib/bolt/node/ssh.rb
@@ -136,14 +136,15 @@ module Bolt
       @logger.debug { "arguments: #{arguments}" }
 
       with_remote_file(script) do |remote_path|
-        # should each arg be quoted?
-        command = "'#{remote_path}'"
-        unless arguments.empty?
-          command += " "
-          command += arguments.join(' ')
+        command = ["'#{remote_path}'"]
+        command += arguments.map do |arg|
+          if arg =~ / /
+            "\"#{arg}\""
+          else
+            arg
+          end
         end
-
-        execute(command)
+        execute(command.join(' '))
       end
     end
 
--- a/lib/bolt/node/ssh.rb
+++ b/lib/bolt/node/ssh.rb
@@ -1,6 +1,7 @@
+require 'json'
+require 'shellwords'
 require 'net/ssh'
 require 'net/sftp'
-require 'json'
 require 'bolt/node/result'
 
 module Bolt
@@ -136,15 +137,7 @@ module Bolt
       @logger.debug { "arguments: #{arguments}" }
 
       with_remote_file(script) do |remote_path|
-        command = ["'#{remote_path}'"]
-        command += arguments.map do |arg|
-          if arg =~ / /
-            "\"#{arg}\""
-          else
-            arg
-          end
-        end
-        execute(command.join(' '))
+        execute("'#{remote_path}' #{Shellwords.join(arguments)}")
       end
     end
 
--- a/lib/bolt/node/winrm.rb
+++ b/lib/bolt/node/winrm.rb
@@ -129,12 +129,20 @@ PS
     # 10 minutes in milliseconds
     DEFAULT_EXECUTION_TIMEOUT = 10 * 60 * 1000
 
-    def execute_process(path = '', arguments = '', stdin = nil,
+    def execute_process(path = '', arguments = [], stdin = nil,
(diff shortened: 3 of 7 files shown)


spec/fixtures/scripts/success.ps1
```diff
@@ -0,0 +1,8 @@
+foreach ($i in $args)
+{
+    Write-Output "arg: $i"
+}
+
+Write-Output "standard output"
+
+# Write-Error "standard error"
```
joshcooper: @Iristyle need to talk to you about this

lib/bolt/node/ssh.rb
```diff
@@ -131,10 +131,23 @@ def _run_command(command)
...
+        command = ["'#{remote_path}'"]
+        unless arguments.empty?
+          command += arguments.map do |arg|
+            if arg =~ / /
+              "\"#{arg}\""
+            else
+              arg
+            end
+          end
+        end
```
joshcooper: This same logic is duplicated in winrm, DRY it

lib/bolt/cli.rb
```diff
@@ -56,7 +56,7 @@ class CLI
 HELP
 
     SCRIPT_HELP = <<-HELP.freeze
-Usage: bolt script <action> <script> [options]
+Usage: bolt script <action> <script> [[arg1] ... [argN]] [options]
```
adreyer: Is this going to result in conflicts between bolt options and script args?

spec/fixtures/scripts/success.ps1
Iristyle: Out of curiosity, what happens if you `throw "error!"` instead?

lib/bolt/cli.rb
joshcooper: Yeah, you have to use double-dash:

```
$ bx bolt script run spec/fixtures/scripts/success.sh -n o4np54hfrljr5wo.delivery.puppetlabs.net -k -- --help
o4np54hfrljr5wo.delivery.puppetlabs.net:

arg: --help
standard out

standard error

Ran on 1 node in 0.51 seconds
```

I don't know if there is an equivalent PowerShell way of handling that (when bolt is run on Windows controller).

Iristyle: Yeah, PS doesn't quite have an idea of `--` that I'm aware of... 

As of PS3, they added `--%` which instructs PowerShell itself to stop parsing, but that's not quite the same thing.

On Windows, might have to do something like quote the script arguments, which requires additional parsing... hmm...

Iristyle: Oh wait, I might stand corrected.  

https://stackoverflow.com/questions/15780174/powershell-command-line-parameters-and

lib/bolt/node/orch.rb
```diff
@@ -105,12 +105,13 @@ def _upload(source, destination)
...
 
-    def _run_script(script)
+    def _run_script(script, arguments)
       content = File.open(script, &:read)
       content = Base64.encode64(content)
       params = {
         action: 'script',
-        content: content
+        content: content,
+        arguments: arguments
```
joshcooper: @adreyer do these arguments need to be shell escaped?

lib/bolt/node/winrm.rb
```diff
@@ -214,10 +216,14 @@ def _run_command(command)
...
+    def _run_script(script, arguments)
       @logger.info { "Running script '#{script}'" }
       with_remote_file(script) do |remote_path|
-        args = "#{PS_ARGS} -File \"#{remote_path}\""
+        args = [*PS_ARGS, '-File', "\"#{remote_path}\""].join(' ')
+        unless arguments.empty?
+          args += " "
+          args += escape_arguments(*arguments)
+        end
         execute_process('powershell.exe', args)
```
joshcooper: @Iristyle need to sync with you about quoting with `Invoke-Interpreter`

lib/bolt/node/winrm.rb
```diff
@@ -245,5 +251,19 @@ def _run_task(task, input_method, arguments)
...
+        if arg =~ /^['].*[']$/
+          "\"#{arg}\""
+        elsif arg =~ /^["].*["]$/
+          "'#{arg}'"
+        elsif arg =~ / /
+          "\"#{arg}\""
+        else
+          arg
+        end
+      end.join(' ')
```
joshcooper: Ideally I'd like a Shellwords.escape for powershell. Something like https://github.com/larskanis/shellwords/tree/master/lib/shellwords

joshcooper: Or https://blogs.msdn.microsoft.com/twistylittlepassagesallalike/2011/04/23/everyone-quotes-command-line-arguments-the-wrong-way/

Iristyle: Yeah, I've read that blog post a few times... it's a good one ;0

PR #135, opened 2017-10-26
Added 'bolt task show'

jennymahmoudi:


ux/bolt-CLI-spec.md
```diff
@@ -45,8 +41,49 @@ Option | Description
...
 
 
+
 ### OUTPUT
 
+** View the list of installed tasks **
+
+~~~
+$ bolt task show
+aos_login           Login to AOS server for session token
```
adreyer: We didn't end up implementing this in PE but it may be easier in bolt

PR #136, opened 2017-10-26
(BOLT-197) Validate plan and task names

adreyer:
Failing to specify a plan or task name to run task previously resulted
in a stack trace. After this commit the plan or task name is validated
for the plan or task command.

### DIFF
--- a/lib/bolt/cli.rb
+++ b/lib/bolt/cli.rb
@@ -420,25 +420,32 @@ HELP
       module_name, file_name = name.split('::', 2)
       file_name ||= 'init'
 
-      env = Puppet::Node::Environment.create('bolt', modulepath)
-      Puppet.override(environments: Puppet::Environments::Static.new(env)) do
-        data = Puppet::InfoService::TaskInformationService.task_data(
-          env.name, module_name, name
-        )
+      begin
+        env = Puppet::Node::Environment.create('bolt', modulepath)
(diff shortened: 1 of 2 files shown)


joshcooper:
Does this fix https://tickets.puppetlabs.com/browse/BOLT-187 too?

adreyer:
It may fix some cases where users ran into it but we'll need to catch MissingModule and maybe some other exceptions in load_task_data to fully fix it.

adreyer:
@joshcooper I added a fix for BOLT-187 without including the bolt task show message

lib/bolt/cli.rb
```diff
@@ -410,25 +420,32 @@ def load_task_data(name, modulepath)
...
+            env.name, module_name, name
+          )
 
-        metadata =
-          if data[:metadata_file]
-            JSON.parse(File.read(data[:metadata_file]))
-          else
-            {}
+          file = data[:files].find { |f| File.basename(f, '.*') == file_name }
+          if file.nil?
```
adreyer: I'm not sure how this can happen with the TaskInformationService

joshcooper: I think we were being overly defensive. We have a ticket to replace all of this with the PAL API for loading and validating task parameters, so will leave as is.

joshcooper:
Superseded in #137

PR #138, opened 2017-10-27
(BOLT-158) Run .pp tasks using puppet apply on Windows

joshcooper:
If the task extension is .pp, then run the task using 'puppet apply' on Windows.
The path to puppet matches the default install location for 64-bit puppet on
64-bit Windows, and 32 on 32. Running 32-bit puppet on 64-bit Windows is not
supported.

On *nix, this logic is handled by the script shebang line.


Iristyle:
👍 

We don't support 32-bit agents on 64-bit Windows anymore anyhow, so I don't think we need to do anything special with pathing to handle that scenario.

lib/bolt/node/winrm.rb
```diff
@@ -161,6 +161,11 @@ def process_from_extension(path)
           'powershell.exe',
           "#{PS_ARGS} -File \"#{path}\""
         ]
+      when '.pp'
+        [
+          'C:\Program Files\Puppet Labs\Puppet\bin\puppet.bat',
```
Iristyle: * In #68, `PATH` became appended with `${ENV:ProgramFiles}\\Puppet Labs\\Puppet\\sys\\ruby\\bin\\` - see https://github.com/puppetlabs/bolt/pull/68/files#diff-0ce90a36bcfd738e65f1aa3cc4892943R32
* In #72, `RUBYLIB` was additionally set to make sure that `require 'puppet'` works in cross-platform Ruby tasks

If we want to improve this a bit, we have a few options:

* Add `${ENV:ProgramFiles}\\Puppet Labs\\Puppet\\bin` to `PATH`, similar to #72
* Skip the ability to load `puppet.bat` and do what `puppet.bat` does, given `ruby.exe` is already in `PATH` and `RUBYLIB` is set to be able to load Puppet.  The batch file currently does `ruby -S -- puppet %*` (see https://github.com/puppetlabs/puppet-agent/blob/master/resources/files/windows/puppet.bat), so it would be fairly easy to do something similar as the `when '.rb'` branch of this mapping logic. HOWEVER, this would be subtly different because `puppet.bat` already calls `environment.bat`, which does a few things we're not yet doing:
    * `SET FACTER_env_windows_installdir=%PL_BASEDIR%` per https://github.com/puppetlabs/puppet-agent/blob/master/resources/files/windows/environment.bat#L8 - this could be important for applying a manifest that consumes this fact
    * `SET RUBYOPT=rubygems` per https://github.com/puppetlabs/puppet-agent/blob/master/resources/files/windows/environment.bat#L27 - not sure what will fail if `rubygems` isn't auto-required
    * `RUBYLIB` is missing `%MCOLLECTIVE_DIR%\lib` per https://github.com/puppetlabs/puppet-agent/blob/master/resources/files/windows/environment.bat#L20 - probably won't cause an issue
    * `SET SSL_CERT_FILE` and `SET SSL_CERT_DIR` per https://github.com/puppetlabs/puppet-agent/blob/master/resources/files/windows/environment.bat#L31-L32 - I *don't think* this is generally a problem for running `puppet apply` but could be problematic if the manifest makes network calls (like Azure / AWS modules or devices do)

PR #139, opened 2017-10-27
(BOLT-189) Error if local script doesn't exist

joshcooper:
If the local script to upload and execute doesn't exist, then generate
an error, instead of calling into net-sftp with a nil argument. We
already verify the argument to `bolt upload`, so that did not need to
change.

### DIFF
--- a/lib/bolt/cli.rb
+++ b/lib/bolt/cli.rb
@@ -364,9 +364,7 @@ HELP
               executor.run_command(nodes, options[:object])
             when 'script'
               script = options[:object]
-              unless file_exist?(script)
-                raise Bolt::CLIError, "The script '#{script}' does not exist"
-              end
+              validate_file('script', script)
               executor.run_script(nodes, script, options[:leftovers])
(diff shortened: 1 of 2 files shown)


joshcooper:
```
$ bx bolt script run does_not_exist.sh -n linux -k
The script 'does_not_exist.sh' does not exist
```

lib/bolt/cli.rb
```diff
@@ -363,7 +363,11 @@ def execute(options)
             when 'command'
               executor.run_command(nodes, options[:object])
             when 'script'
-              executor.run_script(nodes, options[:object], options[:leftovers])
+              script = options[:object]
+              unless file_exist?(script)
```
adreyer: We should probably verify that it's readable too

joshcooper: Updated, also check that it's a file, otherwise upload prints an undecipherable error message.

joshcooper:
ping @adreyer, should be good now

PR #140, opened 2017-10-29
Add to the CLI the possibility to choose a specific ssh Private Key 

augabet:
**The problem:**
When connecting to a remote VM without the default ssh private key (for example google_compute_engine), I was not able to choose which private key I want to use through CLI.

**2 Options:** 

1. You can : ```$eval `ssh-add` && ssh-add ~/.ssh/<PRIVATE-KEY>```
2. Run `$bolt -i ~/.ssh/google_compute_engine ...` or `$bolt --private-key ~/.ssh/google_compute_engine`

I added tests for the CLI and to try connecting with specific private key using the Vagrantfile provided.

### DIFF
--- a/lib/bolt/cli.rb
+++ b/lib/bolt/cli.rb
@@ -135,6 +135,10 @@ HELP
           end
         end
         results[:concurrency] = 100
+        opts.on('--private-key KEY',
+                "Private ssh key to authenticate with (Optional)") do |key|
+          results[:key] = key
+        end
         opts.on('-c', '--concurrency CONCURRENCY', Integer,
                 "Maximum number of simultaneous connections " \
                 "(Optional, defaults to 100)") do |concurrency|
--- a/spec/bolt/node/ssh_spec.rb
+++ b/spec/bolt/node/ssh_spec.rb
@@ -15,6 +15,10 @@ describe Bolt::SSH do
   let(:port) { 2224 }
   let(:command) { "pwd" }
   let(:ssh) { Bolt::SSH.new(hostname, port, user, password) }
+  let(:key) {
+    { config: Bolt::Config.new(key: Dir[".vagrant/**/private_key"],
+                               insecure: true) }
+  }
   let(:insecure) { { config: Bolt::Config.new(insecure: true) } }
   let(:echo_script) { <<BASH }
 for var in "$@"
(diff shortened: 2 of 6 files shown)


joshcooper:
Thanks for your contribution @augabet! I'm curious why're not modifying adding a host/domain specific key your `~/.ssh/config` as that is one of the options that `net-ssh` supports, e.g.

```
Host *.example.com
  IdentityFile ~/.ssh/google_compute_engine
```

augabet:
It’s a third option I would say that I forget to mention, thank you ! As a typical use of ssh, you can either choose to give the path to the private key with the `-i` flag or change the `~/.ssh/config` file like you said.
From my point of view, I prefer pass the key to the command line when I cannot predict the DNS or IP of my server/VM, rather than changing the `~/.ssh/config` file.

spec/bolt/node/ssh_spec.rb
```diff
@@ -15,6 +15,10 @@
   let(:port) { 2224 }
   let(:command) { "pwd" }
   let(:ssh) { Bolt::SSH.new(hostname, port, user, password) }
+  let(:key) {
+    { config: Bolt::Config.new(key: Dir[".vagrant/**/private_key"],
+                               insecure: true) }
+  }
```
joshcooper: It looks like this depends on an external file, so probably won't work for anyone else? Note we don't run tests marked with `vagrant: true` in appveyor or travis, but only when running `bundle exec rake test`.

joshcooper:
Thanks @augabet that makes sense. Could you squash your commits (and it looks like your PR picked up some unrelated changes that have already been merged)?

joshcooper:
Sorry didn't mean to close this.

augabet:
@joshcooper , I squashed my commits, and integrated the latests commits. Tell me if there are another things you want me to do.

spec/bolt/node/ssh_spec.rb
augabet: This file is created under the .vagrant directory when you play `vagrant up` before launching `bundle exec rake test`. Unless I am mistaken, `bundle exec rake test` run some tests with vagrant in contrast to `bundle exec rake unit`.

lib/bolt/cli.rb
```diff
@@ -135,6 +135,10 @@ def create_option_parser(results)
           end
         end
         results[:concurrency] = 100
+        opts.on('-i', '--private-key KEY',
```
adreyer: lets just expose this as --private-key for now.

augabet: OK, I am going to delete the "-i" (and associated tests), squash my new commit and push again on my fork repo.

augabet:
I am worried, my tests pass with vagrant, and not anymore with Travis. Have there been any changes to the test protocole ?

augabet:
@joshcooper up

adreyer:
@augabet This needed to be rebased to pick up some new changes which I did https://github.com/adreyer/bolt/commit/d7c9f37404799938b91f2b0ca2a6854410b0d7f9

the travis tests don't use vagrant they ssh back to a different user on localhost. I set up ssh keys for that user but haven't got the tests passing yet.
https://github.com/adreyer/bolt/commit/1bbdc5a5337adef174f38a9863ff67ce902ccce2

adreyer:
merged in with https://github.com/puppetlabs/bolt/pull/172

augabet:
Thank you @adreyer for your work !

PR #141, opened 2017-10-30
(BOLT-52) Verify orch handles script args correctly

joshcooper:
Update unit test to verify script args can be passed through
orchestrator without shell interpolation.


joshcooper:
@adreyer can you double check this, I'm not 100% sure how orch plumbs args all the way through? /cc @pcarlisle

PR #142, opened 2017-10-31
(maint) Enable AppVeyor WinRM testing

Iristyle:
Turns on WinRM tests in AppVeyor, so that connection behavior is tested on PRs.

This results in some additional runtime, but appears to be < 10% in a few PRs I've spot checked.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -7,6 +7,7 @@ matrix:
   fast_finish: true
 
 environment:
+  BOLT_WINRM_USER: roddypiper
   matrix:
     - RUBY_VERSION: 200
     - RUBY_VERSION: 24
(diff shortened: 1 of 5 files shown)


appveyor.yml
```diff
@@ -7,6 +7,10 @@ matrix:
   fast_finish: true
 
 environment:
+  BOLT_WINRM_USER: roddypiper
+  BOLT_WINRM_PASSWORD: r0Wdy999
```
joshcooper: What happened to the commit that generated a secure password?

Iristyle: I removed for the sake of simplicity / being more declarative, but can easily add back random password generation

Iristyle:
FWIW, I think we can do a similar thing for SSH testing in AppVeyor

PR #145, opened 2017-11-02
(BOLT-151) Yield node results as they finish

joshcooper:
All of the run_* methods now take an optional block. If one is provided, then
the node and its result will be yielded to the block as the operation finishes.
Only one result will be yielded at a time, so callers do not need to worry about
concurrency issues. For example, colorize modifies the global stdout stream
color and doesn't need to worry about concurrency issues.

### DIFF
--- a/lib/bolt/cli.rb
+++ b/lib/bolt/cli.rb
@@ -358,11 +358,17 @@ HELP
           results =
             case options[:mode]
             when 'command'
-              executor.run_command(nodes, options[:object])
+              executor.run_command(nodes, options[:object]) do |node, result|
+                print_result(node, result)
+              end
             when 'script'
               script = options[:object]
               validate_file('script', script)
-              executor.run_script(nodes, script, options[:leftovers])
+              executor.run_script(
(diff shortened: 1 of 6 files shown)


joshcooper:
This approach works well for the human readable format and when executing plans, as the latter case could make the results available to the plan to perform some action.

But it won't work for the oneline output format, as we'll need to print output as it occurs, not when the node finishes. Moving to an observable pattern would work better for that.

adreyer:
We could yield events as well. That would help if we want the human readable output to include information when a node starts or the one line. I can also see how it might align better with intermediate output for plans.

This is going to make the JSON format a bit harder but it should be fine.

joshcooper:
I had plans of making the notifier more general so that it could send events as part of the executor lifecycle, start, progress, done, etc. I think that will eventually be necessary to support oneline output format. But I don't really have time to implement those changes. So I'm ok merging this, unless @pcarlisle has concerns?

pcarlisle:
as is is fine with me

PR #147, opened 2017-11-02
(BOLT-206) Correctly handle winrm exit status

Iristyle:
(BOLT-206) Correctly handle winrm exit status

 - Given the winrm command first looks to the $? automatic variable,
   failure status of running PowerShell code could be misinterpreted.

   $? is the status of the last operation in the context of PowerShell
   and doesn't reflect the status of calling the Invoke-Interperter
   function. So with the success of the function executing, the winrm
   gem is considering the operation a success.

   PowerShell does not allow for setting $? manually (a good thing),
   so there are a few options for handling a scenario in which the
   PowerShell helper code ran correctly, but the invoked executable
   (powershell.exe, ruby.exe, puppet.bat, etc) return a non-zero
   exit status:

   * Generate a failing command that doesn't pollute the PowerShell
     $Error collection (which requires some ugly hacks), so that
     $LASTEXITCODE is returned through the winrm gem
   * Call `exit` explicitly with the desired exit code, which appears
     to behave as expected given the new tests

FYI - there are additional tests around exit status behavior in https://github.com/puppetlabs/puppetlabs-powershell/blob/master/spec/integration/puppet_x/puppetlabs/powershell_manager_spec.rb that we may want to investigate in the future.

### DIFF
--- a/lib/bolt/node/winrm.rb
+++ b/lib/bolt/node/winrm.rb
@@ -149,8 +149,9 @@ $invokeArgs = @{
   #{stdin.nil? ? '' : "StdinInput = @'\n" + stdin + "\n'@"}
 }
 
-# winrm gem relies on $LASTEXITCODE
-$LASTEXITCODE = Invoke-Interpreter @invokeArgs
+# winrm gem checks $? prior to using $LASTEXITCODE
+# making it necessary to exit with the desired code to propagate status properly
+exit $(Invoke-Interpreter @invokeArgs)
 PS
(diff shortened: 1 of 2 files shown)


Iristyle:
The code in the winrm gem that makes this necessary is https://github.com/WinRb/WinRM/blob/b4cadddabd64ae27dc39463b1cee04a7ad3244c8/lib/winrm/shells/power_shell.rb#L101

PR #150, opened 2017-11-06
(BOLT-98) Sudo support

pcarlisle:


lib/bolt/node/ssh.rb
```diff
@@ -68,12 +78,22 @@ def execute(command, options = {})
           raise "could not execute command: #{command.inspect}" unless success
 
           channel.on_data do |_, data|
-            result_output.stdout << data
+            if use_sudo && data == sudo_prompt
+              channel.send_data "#{@password}\n"
```
joshcooper: Should this be `sudo_password`?

pcarlisle: yes, per TODO list in the pr description

MikaelSmith:
Oops, BOLT-45 introduced a conflict.

MikaelSmith:
A commit comment about how this works would also be really helpful...

lib/bolt/node/ssh.rb
```diff
@@ -54,9 +54,19 @@ def disconnect
...
       status = {}
+      use_sudo = sudoable && (@sudo || @run_as)
+      sudo_prompt = '[sudo] Bolt needs to run as another user, password: '
+      if use_sudo
+        user_clause = if @run_as
+                        "-u #{@run_as}"
+                      else
+                        ''
+                      end
+        command = "sudo -S #{user_clause} -p '#{sudo_prompt}' #{command}"
```
MikaelSmith: Couldn't this use `"#{@sudo} ...` to make it slightly easier to add different sudo options in the future?

MikaelSmith: I guess I should just look at doing that, doesn't seem too hard to add a few.

lib/bolt/node/ssh.rb
```diff
@@ -68,12 +78,22 @@ def execute(command, options = {})
           raise "could not execute command: #{command.inspect}" unless success
 
           channel.on_data do |_, data|
-            result_output.stdout << data
+            if use_sudo && data == sudo_prompt
```
MikaelSmith: How many times will this retry if the password is wrong?

pcarlisle: as many times as it's prompted

lib/bolt/node/ssh.rb
pcarlisle: Almost every part needs to be parameterized to add another method so it didn't seem worth it to do it until we actually added one.

MikaelSmith:
Ok, that merge attempt didn't work right.

PR #151, opened 2017-11-06
(maint) Change winrm specs to write to stdout

Iristyle:
- The current WinRM specs use the Write-Output cmdlet. The
   Write-Output cmdlet is designed for writing objects to the
   PowerShell pipeline, *not* for writing stdout to a host.

   PowerShell has a few different methods of writing more directly to
   stdout, the most popular of which is Write-Host. Write-Host is
   typically frowned upon by the PowerShell community, but because it
   is not appropriate for use where pipelines may be involved:

   http://www.jsnover.com/blog/2013/12/07/write-host-considered-harmful/

   When the goal is to emit text to be consumed by other tooling that
   doesn't integrate with the PowerShell pipeline, Write-Host becomes
   an important tool, and likely the case the tests are interested in
   (independent ordered text-based writes to stdout).

   Such counterpoints are provided at:

   https://psitem.com/2015/01/16/is-write-host-really-harmful/

   The other alternatives to calling Write-Host are Out-Host,
   [Console]::WriteLine and the lesser known $Host.UI.WriteLine

 - While Write-Output can generate output for both stdout *and* for
   the PowerShell pipeline, there is also a problem with using it if
   output ordering is important.

   In PowerShell 5, changes were introduced to its behavior, so that
   implicit table formatting receives a 300ms delay. This often
   results in out-of-order output when Write-Output is used, and may
   especially be the case when pipeline writes with Write-Output are
   mixed with direct writes to stdout with Write-Host

   https://github.com/PowerShell/PowerShell/issues/4594
   https://windowsserver.uservoice.com/forums/301869-powershell/suggestions/14925213-bug-console-output-appears-out-of-order

### DIFF
--- a/spec/bolt/node/winrm_spec.rb
+++ b/spec/bolt/node/winrm_spec.rb
@@ -18,7 +18,7 @@ describe Bolt::WinRM do
   let(:echo_script) { <<PS }
 foreach ($i in $args)
 {
-    Write-Output $i
+    Write-Host $i
 }
 PS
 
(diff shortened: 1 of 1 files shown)


spec/bolt/node/winrm_spec.rb
```diff
@@ -203,19 +203,50 @@ def stub_winrm_to_raise(klass, message)
...
+      expect(
+        winrm._run_script(file.path, []).value
+      ).to eq([
+        "message 1\r\n",
+        "message 2\r\n",
+        "message 3\r\n",
+        "message 4\r\n",
+        "message 5\r\n",
+        "message 6\r\n"
+      ].join(''))
```
Iristyle: There are other cmdlets for writing output to different streams, but I think we can skip testing those for now - 

* Write-Debug
* Write-Progress
* Write-Verbose
* Write-Warning

spec/bolt/node/winrm_spec.rb
```diff
@@ -203,19 +203,50 @@ def stub_winrm_to_raise(klass, message)
...
+        .to eq("task is running task has run\r\n")
+    end
+  end
+
+  it "will collect on stdout using valid output mechanisms", winrm: true do
+    contents = <<-PS
+    # explicit Format-Table for PS5+ overrides implicit Format-Table which
+    # includes a 300ms delay waiting for more pipeline output
+    # https://github.com/PowerShell/PowerShell/issues/4594
+    Write-Output "message 1" | Format-Table
```
Iristyle: If my theory is correct, then this `Write-Output` call should never end up appearing in output *after* the other writes.

If my theory is incorrect, this is a decent trap for future us.

Iristyle:
This was to address the intermittent fail we saw in output ordering at https://ci.appveyor.com/project/puppetlabs/bolt/build/0.1.0.565-master/job/swpxxswyvqx73po7

PR #152, opened 2017-11-07
(BOLT-208) Execute WinRM scripts by whitelist

Iristyle:
- Change the WinRM _run_script helper to behave similarly to
   _run_task. Look up whitelisted file extensions to determine what
   may be executed on the remote host - currently this includes .rb,
   .ps1 and .pp

- Add some basic mocked tests demonstrating scripts may be invoked
   in this fashion

- Update wording on some tests so its clear when PowerShell is being
   invoked.

### DIFF
--- a/lib/bolt/node/winrm.rb
+++ b/lib/bolt/node/winrm.rb
@@ -272,9 +272,9 @@ PS
     def _run_script(script, arguments)
       @logger.info { "Running script '#{script}'" }
       with_remote_file(script) do |remote_path|
-        args = [*PS_ARGS, '-File', "\"#{remote_path}\""]
+        path, args = *process_from_extension(remote_path)
         args += escape_arguments(arguments)
-        execute_process('powershell.exe', args)
+        execute_process(path, args)
       end
     end
 
--- a/spec/bolt/node/winrm_spec.rb
+++ b/spec/bolt/node/winrm_spec.rb
@@ -122,7 +122,7 @@ PS
     end
   end
 
-  it "can run a script remotely", winrm: true do
+  it "can run a PowerShell script remotely", winrm: true do
     contents = "Write-Output \"hellote\""
     with_tempfile_containing('script-test-winrm', contents) do |file|
       expect(
@@ -131,7 +131,7 @@ PS
     end
   end
 
-  it "can run a script remotely with quoted arguments", winrm: true do
+  it "can run a PowerShell script remotely with quoted args", winrm: true do
     with_tempfile_containing('script-test-winrm-quotes', echo_script) do |file|
       expect(
         winrm._run_script(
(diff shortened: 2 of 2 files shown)


lib/bolt/node/winrm.rb
```diff
@@ -272,9 +272,10 @@ def _run_command(command)
     def _run_script(script, arguments)
       @logger.info { "Running script '#{script}'" }
       with_remote_file(script) do |remote_path|
-        args = [*PS_ARGS, '-File', "\"#{remote_path}\""]
+        path, args = *process_from_extension(remote_path)
+        # TODO: why do we escape in script, but not in task?
         args += escape_arguments(arguments)
-        execute_process('powershell.exe', args)
+        execute_process(path, args, stdin)
```
Iristyle: This will require mocking a test given there is no Puppet / Ruby / etc installed on the remote host by default.

lib/bolt/node/winrm.rb
```diff
@@ -272,9 +272,10 @@ def _run_command(command)
     def _run_script(script, arguments)
       @logger.info { "Running script '#{script}'" }
       with_remote_file(script) do |remote_path|
-        args = [*PS_ARGS, '-File', "\"#{remote_path}\""]
+        path, args = *process_from_extension(remote_path)
+        # TODO: why do we escape in script, but not in task?
```
Iristyle: @joshcooper one open question here...

joshcooper: I think it was just an oversight. I think always quoting via `process_from_extension` is best.

Iristyle:
Tested locally and passing.

Grab fresh VM, install Choco and puppet-agent packages:

```
bundle exec bolt command run 'Set-ExecutionPolicy Bypass; iex ((New-Object System.Net.WebClient).DownloadString("https://chocolatey.org/install.ps1"))' --nodes $BOLT_HOST --user Administrator --password $BOLT_PASSWORD

bundle exec bolt command run 'choco install puppet-agent -y' --nodes $BOLT_HOST --user Administrator --password $BOLT_PASSWORD
```

Create a .PP manifest

```puppet
notify { 'hello world': }

notice('hello notice')
```

Execute

```
bundle exec bolt script run ./spec/fixtures/scripts/success.pp --nodes $BOLT_HOST --user Administrator --password $BOLT_PASSWORD --debug --verbose
```

Result (success)

```
Notice: Scope(Class[main]): hello notice
Notice: Compiled catalog for surc7brrhq2xuje.delivery.puppetlabs.net in environment production in 0.05 seconds
Notice: hello world
Notice: /Stage[main]/Main/Notify[hello world]/message: defined 'message' as 'hello world'
Notice: Applied catalog in 0.14 seconds
```

Similarly, create a `.rb`

```ruby
STDOUT.puts("standard output")
STDERR.puts("error")
```

Execute it

```
bundle exec bolt script run ./spec/fixtures/scripts/success.rb --nodes $BOLT_HOST --user Administrator --password $BOLT_PASSWORD --debug --verbose
```

Result (success)

```
surc7brrhq2xuje.delivery.puppetlabs.net:

standard output

error
```

PR #153, opened 2017-11-08
(BOLT-72) Execute PowerShell scripts in session

Iristyle:
- No longer executes a fresh powershell.exe per command invocation
 - Fixes argument passing problems around quotes (at least for
   PowerShell scripts)
 - Changes what state carries over from script run to script run, so
   that the following is now available in subequent calls to
   _run_script against the same WinRM node:

   * Environment variables
   * script scoped variables
   * global scoped variables

### DIFF
--- a/lib/bolt/node/winrm.rb
+++ b/lib/bolt/node/winrm.rb
@@ -199,6 +199,10 @@ PS
       -NoProfile -NonInteractive -NoLogo -ExecutionPolicy Bypass
     ].freeze
 
+    def powershell_file?(path)
+      Pathname(path).extname.casecmp('.ps1').zero?
+    end
+
     def process_from_extension(path)
       case Pathname(path).extname.downcase
       when '.rb'
@@ -272,9 +276,31 @@ PS
     def _run_script(script, arguments)
       @logger.info { "Running script '#{script}'" }
       with_remote_file(script) do |remote_path|
-        path, args = *process_from_extension(remote_path)
-        args += escape_arguments(arguments)
-        execute_process(path, args)
+        if powershell_file?(remote_path)
+          mapped_args = arguments.map do |a|
+            "$invokeArgs.ArgumentList += @'\n#{a}\n'@"
+          end.join("\n")
+          execute(<<-PS)
+$invokeArgs = @{
+  ScriptBlock = (Get-Command "#{remote_path}").ScriptBlock
+  ArgumentList = @()
+}
+#{mapped_args}
+
+try
+{
+  Invoke-Command @invokeArgs
+}
+catch
+{
+  exit 1
+}
+          PS
+        else
+          path, args = *process_from_extension(remote_path)
+          args += escape_arguments(arguments)
+          execute_process(path, args)
+        end
       end
     end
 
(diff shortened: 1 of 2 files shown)


lib/bolt/node/winrm.rb
```diff
@@ -297,8 +319,22 @@ def _run_task(task, input_method, arguments)
         Bolt::Node::Success.new
       end.then do
         with_remote_file(task) do |remote_path|
-          path, args = *process_from_extension(remote_path)
-          execute_process(path, args, stdin)
+          if is_powershell_file(remote_path) && stdin.nil?
+            execute(<<-PS)
+try
+{
+  # TODO: powershell doesn't support STDIN redirection when executing a script
```
Iristyle: So we end up having to invoke `powershell.exe` anyhow if the task supports STDIN... maybe not a big deal because that combo should never be used in real life... but it makes the code more complex.

spec/bolt/node/winrm_spec.rb
```diff
@@ -161,7 +161,7 @@ def stub_winrm_to_raise(klass, message)
     end
   end
 
-  it "does not reuse host for multiple PowerShell scripts", winrm: true do
```
Iristyle: This 2nd commit shows the behavioral differences around state between `powershell.exe` for every script, and invoking inside the current session.

Environment variables and globally scoped PowerShell variables remain accessible to additional invocations against the same host.

Were we to stop writing the PowerShell scripts to files, and instead invoke them in the current shell, the scoping rules would change yet again.

I figure these tests provide a good initial place to document what state is accessible later and why...

lib/bolt/node/winrm.rb
```diff
@@ -271,10 +276,37 @@ def _run_command(command)
...
-        path, args = *process_from_extension(remote_path)
-        args += escape_arguments(arguments)
-        execute_process(path, args)
+      if powershell_file?(script)
+        mapped_args = arguments.map do |a|
+          "$invokeArgs.ArgumentList += @'\n#{a}\n'@"
+        end.join("\n")
+        # TODO: watch out for File.read encoding - winrm is 65001
+        # TODO: need to cleanup after ScriptBlock? purge shell history?
+        # TODO: document scripts may not include @' '@ style here-strings
```
Iristyle: Workarounds for nested here-strings - https://www.reddit.com/r/PowerShell/comments/589fsw/nested_herestring/

Iristyle:
After further thought and some discussion with @michaeltlombardi, I've removed e746be7, which was an exploration into running PowerShell scripts without writing files.  While it did improve performance, it came with a few downsides:

* The PowerShell scripts themselves could not include `@'` / `'@` style here-strings, which in itself is likely a deal breaker.
* We would have had to deal more directly with the encoding of script files on disk rather than simply passing them to remote hosts to execute (i.e. consider issues with BOMs / UTF-8 vs UTF-16 / etc). Passing files makes things simpler in that case.

While the PowerShell module takes the above approach, ultimately it seems inappropriate for Bolt given the use cases.  Any performance issues with writing scripts to disk and sending them to remote hosts should be mitigated in later efforts that explore using SMB instead of the WinRM gem to send files.

lib/bolt/node/winrm.rb
```diff
@@ -297,8 +323,22 @@ def _run_task(task, input_method, arguments)
         Bolt::Node::Success.new
       end.then do
         with_remote_file(task) do |remote_path|
-          path, args = *process_from_extension(remote_path)
-          execute_process(path, args, stdin)
+          if powershell_file?(remote_path) && stdin.nil?
+            execute(<<-PS)
+try
+{
+  # TODO: powershell doesn't support STDIN redirection when executing a script
```
Iristyle: I don't think that we care about this `TODO`, but should discuss / remove the note prior to merging this PR.

MikaelSmith: This seems fine, let's remove the TODO.

PR #154, opened 2017-11-09
(BOLT-185) Add bolt functions from puppet

hlindberg:
This adds the execution functions from Puppet into an embedded module in Bolt. The unit tests for the functions have also been copied over.
At the moment those tests does not run as they require 'puppet' and sublibraries thereof. (Need help to fix those).


modules/bolt/metadata.json
```diff
@@ -0,0 +1,8 @@
+{
+    "name": "puppetlabs/bolt",
+    "version": "0.7.0",
```
adreyer: We probably need something that bumps this or it should be left out.

hlindberg: I used what we use as minimum meta data in unit tests in puppet. Don't know what we can exclude, or if we need the metadata.json at all.

hlindberg:
@adreyer sure, not good if there is a confusion over which module is which. The name of the embedded module will only show up if there is an error - or if bolt needs to include things in the puppet namespace that require prefixing with the module name. Then users would also see that name. Is one of `boltcore`, `boltlib`, `boltpp`, or `boltmodule`  acceptable perhaps?

modules/bolt/metadata.json
adreyer: I think we can live without maybe here since it's going to be vendored

adreyer:
Of those I like boltlib but they're all fine

hlindberg:
I renamed the embedded module to `boltlib` and dropped the `metadata.json`.
Still need help with how to get the unit testing of the functions to work - they cannot just require 'puppet' as things are configured now.

joshcooper:
@hlindberg I think you just need to `require 'vendored/require_vendored'`. We'll also need bolt's `spec_helper.rb` to register before_each, etc hooks in puppet's `lib/puppet/test/test_helper.rb`. Something like:

```
require 'vendored/require_vendored'
require 'puppet/test/test_helper'

RSpec.configure do |config|
  Puppet::Test::TestHelper.initialize

  config.before :all do
    Puppet::Test::TestHelper.before_all_tests
  end

  config.before :each do
    Puppet::Test::TestHelper.before_each_test
  end

  # etc
end
```

Alternatively, we could use `puppetlabs_spec_helper`...

MikaelSmith:
I'm working on updating this. There's a bit more work needed than Josh outlined.

modules/boltlib/lib/puppet/functions/file_upload.rb
```diff
@@ -0,0 +1,53 @@
...
+# * This function does nothing if the list of targets is empty.
+# * It is possible to run on the target 'localhost'
+# * A target is a String with a targets's hostname or a Target.
+# * The returned value contains information about the result per target.
+#
+# Since > 5.4.0 TODO: Update when version is known
+#
+Puppet::Functions.create_function(:file_upload, Puppet::Functions::InternalFunction) do
+  local_types do
+    type 'TargetOrTargets = Variant[String[1], Target, Array[TargetOrTargets]]'
```
MikaelSmith: Where does the Target type come from? It doesn't appear to be setup during testing, but I'm not even finding where it's defined.

adreyer:
@MikaelSmith I think we probably need to rewrite the tests to use normal module testing tools instead of requiring directly from puppet/spec

MikaelSmith:
Closing in favor of https://github.com/puppetlabs/bolt/pull/161.

PR #156, opened 2017-11-09
(BOLT-45) Support JSON format

adreyer:
This commit adds support for the --format flag and the json format
option. When specified the format is used for results of the action
printed to stdout. It does not effect logging to stderr.

Once the functions are moved into bolt it will be easier to clean up the relationship between results in bolt and execution results in puppet.

### DIFF
--- a/lib/bolt/cli.rb
+++ b/lib/bolt/cli.rb
@@ -6,6 +6,7 @@ require 'json'
 require 'bolt/node'
 require 'bolt/version'
 require 'bolt/executor'
+require 'bolt/outputter'
 require 'bolt/config'
 require 'io/console'
 
@@ -16,6 +17,12 @@ module Bolt
     def initialize(msg, error_code: 1)
       super(msg)
       @error_code = error_code
+    end
+
+    def to_json
+      { kind: "bolt/cli-error",
+        msg: message,
+        details: { error_code: @error_code } }.to_json
     end
   end
 
@@ -149,6 +156,17 @@ HELP
                 "Parameters to a task or plan") do |params|
           results[:task_options] = parse_params(params)
         end
+
+        results[:format] = 'human'
+        opts.on('--format FORMAT',
+                "Output format to use") do |format|
+          if %w[human json].include? format
+            results[:format] = format
+          else
+            raise ArgumentError "Unsupported format: #{format}"
+          end
+        end
+        results[:insecure] = false
         opts.on('-k', '--insecure',
                 "Whether to connect insecurely ") do |insecure|
           results[:insecure] = insecure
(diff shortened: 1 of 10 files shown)


lib/bolt/cli.rb
```diff
@@ -149,6 +156,17 @@ def create_option_parser(results)
                 "Parameters to a task or plan") do |params|
           results[:task_options] = parse_params(params)
         end
+
+        results[:format] = 'human'
```
MikaelSmith: Some consistency around spacing between default/option pairs would be nice. This is the only place with a newline, and I like the newline for readability.

lib/bolt/cli.rb
```diff
@@ -451,6 +456,10 @@ def file_stat(path)
       File.stat(path)
     end
 
+    def outputter
```
MikaelSmith: This function seems unnecessary and confusing when we're only going to use the outputter in a single function lifetime.

lib/bolt/result.rb
```diff
@@ -95,12 +119,15 @@ def to_h
     private
 
     def output_to_json_hash(output)
-      obj = JSON.parse(output)
-      if obj.is_a? Hash
-        obj
+      begin
+        obj = JSON.parse(output)
+        if obj.is_a? Hash
+          obj
```
MikaelSmith: This won't be returned, so the whole if-clause is pointless. Not sure this is doing what you intend because if `obj` is json but not a hash, we're still going to reach line 130 and return obj rather than wrap it in `_output`.

MikaelSmith: Should add a test covering that case after you decide how it should be handled.

MikaelSmith: Running `bolt command run 'echo "{"a": 1}"' -n node1` didn't do what I'd expect. Piped through `jq` it looks like 
```
{
  "items": [
    {
      "name": "node1",
      "status": "success",
      "result": {
        "stdout": "{a: 1}\n",
        "stderr": "",
        "exit_code": 0
      }
    }
  ],
  "node_count": 1,
  "elapsed_time": 0
}
```

I thought it would be parsed in some way.

MikaelSmith:
Now I'm trying `bolt task run echo message='{"a": 1}' -n node1 --format json --modulepath ../orchestration-control/site | jq` and get
```
{
  "items": [
    {
      "name": "node1",
      "status": "success",
      "result": {
        "message": "{\"a\": 1}"
      }
    }
  ],
  "node_count": 1,
  "elapsed_time": 0
}
```

Not sure what I'm missing there.

MikaelSmith:
Oh, probably because the echo task already creates an object. So that looks right.

PR #157, opened 2017-11-13
(BOLT-163) Configurable connection timeout

Iristyle:
Supercedes #81 

Needs a few more tests:

- [x] bad hostname
- [x] port that is open but cannot be connected to given specific protocol (i.e. try WinRM connection to SSH port)
- [x] port that is not open

Needs some extra info on why WinRM is now configured as such.

### DIFF
--- a/lib/bolt/cli.rb
+++ b/lib/bolt/cli.rb
@@ -144,9 +144,9 @@ HELP
                 "(Optional, defaults to 100)") do |concurrency|
           results[:concurrency] = concurrency
         end
-        opts.on('--timeout TIMEOUT', Integer,
+        opts.on('--connect-timeout TIMEOUT', Integer,
                 "Connection timeout (Optional)") do |timeout|
-          results[:timeout] = timeout
+          results[:connect_timeout] = timeout
         end
         opts.on('--modulepath MODULES',
                 "List of directories containing modules, " \
--- a/lib/bolt/cli.rb
+++ b/lib/bolt/cli.rb
@@ -144,6 +144,10 @@ HELP
                 "(Optional, defaults to 100)") do |concurrency|
           results[:concurrency] = concurrency
         end
+        opts.on('--timeout TIMEOUT', Integer,
+                "Connection timeout (Optional)") do |timeout|
+          results[:timeout] = timeout
+        end
         opts.on('--modulepath MODULES',
                 "List of directories containing modules, " \
                 "separated by #{File::PATH_SEPARATOR}") do |modulepath|
--- a/lib/bolt/config.rb
+++ b/lib/bolt/config.rb
@@ -23,7 +23,7 @@ module Bolt
     TRANSPORT_OPTIONS = %i[insecure password run_as sudo sudo_password key tty user connect_timeout].freeze
 
     TRANSPORT_DEFAULTS = {
-      connect_timeout: 1,
+      connect_timeout: 10,
       insecure: false,
       tty: false
     }.freeze
--- a/lib/bolt/config.rb
+++ b/lib/bolt/config.rb
@@ -94,6 +94,15 @@ module Bolt
         if data['ssh']['insecure']
           self[:transports][:ssh][:insecure] = data['ssh']['insecure']
         end
+        if data['ssh']['connect-timeout']
+          self[:transports][:ssh][:connect_timeout] = data['ssh']['connect-timeout']
+        end
+      end
+
+      if data['winrm']
+        if data['winrm']['connect-timeout']
+          self[:transports][:winrm][:connect_timeout] = data['winrm']['connect-timeout']
+        end
       end
       # if data['pcp']
       # end
--- a/lib/bolt/config.rb
+++ b/lib/bolt/config.rb
@@ -20,10 +20,10 @@ module Bolt
       log_destination: STDERR
     }.freeze
 
-    TRANSPORT_OPTIONS = %i[insecure password run_as sudo sudo_password key tty user timeout].freeze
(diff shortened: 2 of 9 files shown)


spec/bolt/node/ssh_spec.rb
```diff
@@ -98,6 +106,35 @@
         ssh.connect
       end
     end
+
+    it "takes < 2 seconds if connection fails to non-SSH port", ssh: true do
+      bogus_port = 5555
+      Process.spawn("nc -l #{bogus_port}")
```
Iristyle: This is gross.. probably want to try to use something cross-platform, like `TCPServer.new`.  Also, can remove `ssh: true`

spec/bolt/node/winrm_spec.rb
```diff
@@ -35,25 +35,26 @@ def stub_winrm_to_raise(klass, message)
...
-                        'CONNECT_ERROR',
-                        /Failed to connect to/) do
-        winrm.connect
+      # timeout will never occur given default 1 second connection timeout
+      Timeout.timeout(2) do
+        expect_node_error(Bolt::Node::ConnectError,
+                          'CONNECT_ERROR',
+                          /Failed to connect to/) do
+          winrm.connect
+        end
```
Iristyle: Should probably add a `winrm` test that validates that it gives up after the given `timeout` seconds, like the SSH tests above using netcat.

lib/bolt/cli.rb
```diff
@@ -140,6 +140,10 @@ def create_option_parser(results)
                 "(Optional, defaults to 100)") do |concurrency|
           results[:concurrency] = concurrency
         end
+        opts.on('-t', '--timeout TIMEOUT', Integer,
```
Iristyle: Should this be called `timeout` or `connection_timeout`?  I would expect we're able to eventually implement some sort of `execution_timeout` (on the WinRM side the functionality is already there, just not exposed)

joshcooper: 👍  to `connection_timeout` as it's fairly common to also expose `read_timeout`

MikaelSmith: Probably agree with `connection_timeout`. How often does this need to be used?

lib/bolt/node/winrm.rb
```diff
@@ -12,9 +12,17 @@ def initialize(host, port, user, password, shell: :powershell, **kwargs)
...
     def connect
-      @connection = ::WinRM::Connection.new(endpoint: @endpoint,
-                                            user: @user,
-                                            password: @password)
+      options = { endpoint: @endpoint,
+                  user: @user,
+                  password: @password,
+                  retry_limit: 1 }
+
+      if @timeout
```
MikaelSmith: Is this both connection and execution timeouts? Why are we changing from the default of `receive_timeout` being 10 seconds longer than `operation_timeout`?

MikaelSmith: Also, why are we changing the `retry_limit`?

Iristyle: I need to play a bit more with `receive_timeout` to see if it actually influences `operation_timeout` in a meaningful way. The initial defaults seemed a bit arbitrary based on observed behavior.

`retry_limit` is what causes the `timeout` to be completely ineffectual - i.e. what happens by default (with `retry_limit` of 3) is:

* fails to connect
* sleep for 10 seconds (default retry sleep)
* retry
* fails to connect
* sleep for 10 seconds
* retry
* fails to connect
* retry limit hit, fail

So for a connection failure, by default, it takes 20+ seconds before it fails, which doesn't make a lot of sense when the intent is to wait up to 1 second for a fail.

There are some notes in the commit about this, and I'll add more detail when this is finished.

lib/bolt/cli.rb
Iristyle: I think we should set low sensible defaults (1 second seems reasonable to me), making `connection_timeout` rarely used in cases where network connectivity is a problem.

nicklewis:
Updated the tests to not shell out to `nc`, and renamed the option to `--connect-timeout`.

I'm still having some trouble with WinRM in the case where the port is entirely unresponsive. The timeout doesn't seem to be getting hit there...

nicklewis:
Okay this seems to be working now. The default connect timeout wasn't being honored but is now, and I changed the way we apply the timeout to WinRM so that it actually works.

nicklewis:
Tests pass! I think this is ready to go.

lib/bolt/cli.rb
```diff
@@ -144,6 +144,10 @@ def create_option_parser(results)
                 "(Optional, defaults to 100)") do |concurrency|
           results[:concurrency] = concurrency
         end
+        opts.on('-t', '--connect-timeout TIMEOUT', Integer,
```
adreyer: perhaps we shouldn't use -t for this

lib/bolt/config.rb
```diff
@@ -20,9 +20,10 @@ module Bolt
       log_destination: STDERR
     }.freeze
 
-    TRANSPORT_OPTIONS = %i[insecure password run_as sudo sudo_password key tty user].freeze
+    TRANSPORT_OPTIONS = %i[insecure password run_as sudo sudo_password key tty user connect_timeout].freeze
```
adreyer: Should we load this from the config file too?

nicklewis:
Updated to remove the short flag `-t` and to allow `connect-timeout` to be set in the transport-specific sections of the config.

nicklewis:
Okay, tests are finally green. Hopefully they stay that way. I added some more helpful messaging around connection timeouts.

PR #158, opened 2017-11-15
Updated human format

jennymahmoudi:
Updated human format for tasks, commands, scripts and files

### DIFF
--- a/ux/bolt-CLI-spec.md
+++ b/ux/bolt-CLI-spec.md
@@ -86,47 +86,72 @@ PARAMETERS:
 
 
 
-**View output per node while task/command/script is running on single or multiple nodes.**
+**View output per node while task/command/script/file is running on single or multiple nodes.**
 
-**--human format (default)**
+**--human format (default) for commands, scripts, and files**
 
 ~~~
-$ bolt [task/command/script] run [options...]
-Starting [task/command/script]...
+$ bolt [command/script/file] run [options...]
+Starting [command/script/file]...
 Nodes: 3
 
 Started on node-1...
 Started on node-2...
 Started on node-3...
-Finished on node-1
-  status: [succeeded/failed/<task-specific staus>]
-    STDOUT:
-      [stdout output...]
-    STDERR:
-      [stderr output...]
-Finished on node-2
-  status: [succeeded/failed/<task-specific staus>]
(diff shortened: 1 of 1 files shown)


ux/bolt-CLI-spec.md
```diff
@@ -86,47 +86,72 @@ PARAMETERS:
...
 Duration: 27 sec
 ~~~
 
+
 - colors should match puppet job run output for started, finished, and errors.
 - For tasks, capitalization of the finished node results comes from module (eg. status vs. Status).
-- status will be succeeded or failed for command and scripts; task status may be something else specified by task.
 - STDOUT or STDERR labels will be printed only if the outout from that stream is non-empty.
-
+- file will only output STDERR; there will be no STDOUT
```
adreyer: There is not STDERR just

Finished on node-3 (in green)
or
Failed on node-3 (in red)
  Could not connect to node-3 (in red)

jennymahmoudi: Is the error message _error?  Is the file output a special case of tasks, commands, or should it have its own spec section?

jennymahmoudi:
@adreyer made the changes. ready for review/merge

ux/bolt-CLI-spec.md
```diff
@@ -86,49 +86,96 @@ PARAMETERS:
...
-  status: [succeeded/failed/<task-specific staus>]
-    STDOUT:
-      [stdout output...]
-    STDERR:
-      [stderr output...] 
+[Succeeded/Failed] on node-1
+  STDOUT:
+    [stdout output...]
+  STDERR:
+    [stderr output...]
```
MikaelSmith: Should we print
```
STDERR:

```
when there's no stderr output?

jennymahmoudi: No. There is a note in the spec: "STDOUT or STDERR labels will be printed only if the output from that stream is non-empty."

adreyer: oops I didn't implement that.
Should any message be printed if there is no output?

jennymahmoudi: Let's go with no message for now, unless we learn of a use case where one is required. The status will be printed regardless.

PR #162, opened 2017-11-18
(maint) Increase AppVeyor matrix to include <PS5

Iristyle:
- Previously, the matrix was not spread across multiple versions,
   which have some differing behaviors when it comes to the behavior
   of WinRM, powershell.exe exit codes, etc.

 - Increase the matrix to include both Windows 2012R2 testing (which
   includes PowerShell 4)

 - Add a failsafe in the event AppVeyor changes image formats and
   upgrades PowerShell past 4, invalidating our test matrix.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,4 +1,9 @@
 # See https://www.appveyor.com/docs/lang/ruby/
+image:
+# Windows Server 2012R2
+- Visual Studio 2013
+# Windows Server 2016
+- Visual Studio 2017
 version: 0.1.0.{build}-{branch}
 cache:
   - .bundle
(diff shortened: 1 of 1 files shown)


Iristyle:
I'm attempting this craziness because we uncovered a bug when executing in PS4 that didn't show up in testing against PS5 either locally (Nano) OR in AppVeyor (2012R2 and 2016 images all have WMF5)

appveyor.yml
```diff
@@ -14,7 +19,27 @@ environment:
...
+init:
+- ps: |
+    if (($ENV:APPVEYOR_BUILD_WORKER_IMAGE -eq 'Visual Studio 2013') -and ($PSVersionTable.PSVersion -gt [Version]'5.0.0'))
+    {
+      Get-Hotfix
+      Write-Output "Uninstalling WMF5.1...";
+      # uninstall the WMF5.1 package on 2012R2
+      # wusa /uninstall /kb:3191564
+      Write-Output "Uninstalling WMF5...";
+      Start-Process 'wusa.exe' -ArgumentList @('/uninstall', '/kb:3134758', '/norestart') -Wait
```
Iristyle: `-Wait`... forever it would seem.

MikaelSmith:
WIP?

Iristyle:
Yup @MikaelSmith ... this could be a total dead end.  Was trying to increase test coverage to include PS4 due to differences between it and PS5, but we may just have to rely on some Jenkins smoke testing.

Recently discovered that error / exit behavior in PS4 is different, and therefore nodes are reporting as successful even when they fail.  #153 fixes that particular problem FWIW.

adreyer:
Closed for now. We do need to find a way to test on ps4 eventually but this is probably not it.

Iristyle:
Yup, this is too hacky.  The solution might be to use custom Hyper-V images published to Rackspace Cloud or custom images published to Google Cloud.

https://www.appveyor.com/docs/build-environment/#build-clouds

Issue #163, opened 2017-11-20
Sudo password prompt


zyronix:
Hi there,

I just tried out the latest version of bolt and really great to see the sudo support! Currently it only works if you supply the sudo password through the commandline interface. I prefer to also have a password prompt for the sudo password (the same as it is now for the ssh password). I guess it would only require a couple of extra lines of code.

Current lines of code for password
```
        opts.on('-p', '--password [PASSWORD]',
                'Password to authenticate with (Optional).',
                'Omit the value to prompt for the password.') do |password|
          if password.nil?
            STDOUT.print "Please enter your password: "
            results[:password] = STDIN.noecho(&:gets).chomp
            STDOUT.puts
          else
            results[:password] = password
          end
end
```
But this is not there at the sudo_password field:

```
        opts.on('--sudo-password [PASSWORD]',
                'Password for privilege escalation') do |password|
          options[:sudo_password] = password
end
```

This would really be great if this can also be implemented :)

MikaelSmith:
That seems pretty straight-forward. I'll take a look at adding it.

MikaelSmith:
And hit the wrong button.

MikaelSmith:
Reproduced this, it just hangs if I pass `--sudo-password` without an argument and one is required.

MikaelSmith:
It also seems to behave poorly if your user doesn't have sudoer access. Just hangs after the successful password. I'll see if I can do anything about that.

zyronix:
Awesome, thanks and confirmed it to be working :D

nicklewis:
Sounds like this is fixed! Please reopen if I'm mistaken.

PR #164, opened 2017-11-21
(BOLT-226) Fix sudo-password so password is actually optional (#163)

MikaelSmith:
Prompts for the privelege escalation password if `sudo-password` is
passed on the CLI with no value.

### DIFF
--- a/lib/bolt/cli.rb
+++ b/lib/bolt/cli.rb
@@ -190,7 +190,13 @@ HELP
         end
         opts.on('--sudo-password [PASSWORD]',
                 'Password for privilege escalation') do |password|
-          options[:sudo_password] = password
+          if password.nil?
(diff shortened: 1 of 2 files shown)


MikaelSmith:
Accidentally pushed an extra commit, fixed.

lib/bolt/cli.rb
```diff
@@ -190,7 +190,13 @@ def create_option_parser(results)
         end
         opts.on('--sudo-password [PASSWORD]',
                 'Password for privilege escalation') do |password|
-          options[:sudo_password] = password
+          if password.nil?
+            STDOUT.print "Please enter your privelege escalation password: "
```
nicklewis: privilege

MikaelSmith: I managed to misspell that everywhere, fixed up.

PR #168, opened 2017-11-22
(BOLT-44) refactor result handling and implement human output

adreyer:
This commit implements the human output format for bolt and adds a
node_started event to outputters to support this.

It also refactors how results are passed simplifying the number of
result classes. Previously there were two sets of result classes one
that was used in the transport for control flow which was translated to
a result of the second type for display. Now there is only a single set
of result classes which are returned by the transport run_* functions
directly. There are no longer separate classes for success and failure
the success? method should be called to differentiate between the two.
Control flow in transport logic has been refactored to use
raise/rescue/ensure instead of result.then/result.ensure.

TODO:
- In order to maintain compatibility with ExecutionResults the to_h
  method of the result class has not been changed. The to_result method
  should be moved to to_h in the future when ExecutionResult can be
  updated
- The transport methods currently catch exceptions and generate "Result"
  instances from them. This is unecessary since Exceptions are caught in
  the executor as well. The transport methods should raise
  Bolt::Node::BaseError subclasses and let the executor handle them


lib/bolt/cli.rb
```diff
@@ -5,25 +5,20 @@
...
 
 module Bolt
-  class CLIError < RuntimeError
+  class CLIError < Bolt::Error
     attr_reader :error_code
 
     def initialize(msg, error_code: 1)
-      super(msg)
+      super(msg, "bolt/cli-error")
       @error_code = error_code
```
MikaelSmith: How does error_code get used? Should it be passed as the details argument to `super`?

lib/bolt/executor.rb
```diff
@@ -37,18 +37,18 @@ def on(nodes, callback = nil)
...
             rescue StandardError => ex
-              node.logger.error(ex)
-              Bolt::ExceptionResult.new(ex)
+              Bolt::Result.from_exception(ex)
             ensure
               node.disconnect
             end
           results[node] = result
-          @notifier.notify(callback, node, result) if callback
+          if callback
```
MikaelSmith: Why do we make this optional? I see `callback = nil` in args, but if we never exercised it I'm not sure why we want to preserve that.

lib/bolt/node/errors.rb
```diff
@@ -1,10 +1,10 @@
 module Bolt
   class Node
-    class BaseError < StandardError
+    class BaseError < Bolt::Error
       attr_reader :issue_code
 
       def initialize(message, issue_code)
-        super(message)
+        super(message, kind)
```
MikaelSmith: I'd opt for passing `super(message, kind, nil, issue_code)` personally to make it clear that `@issue_code` is handled by the parent. Minor quibble.

lib/bolt/node/errors.rb
```diff
@@ -18,5 +18,22 @@ def kind
...
+    class FileError < BaseError
+      def kind
+        'puppetlabs.tasks/task_file_error'
+      end
+    end
+
+    class EnvironmentVarError < BaseError
+      def initialize(var, val)
+        message = "Could not set environment variable '#{var}' to '#{val}'"
+        super(message, 'ENVVAR_ERROR')
```
MikaelSmith: How is this issue code helpful vs `puppetlabs.tasks/environment-var-error`?

lib/bolt/node/orch.rb
```diff
@@ -47,40 +47,29 @@ def _run_task(task, _input_method, arguments)
...
-      # Otherwise create a new result with the captured output
-      result_output = Bolt::Node::ResultOutput.new
-      result_output.stdout << task_result['stdout']
-      result_output.stderr << task_result['stderr']
-      if (task_result['exit_code']).zero?
-        Bolt::Node::Success.new(task_result['stdout'], result_output)
-      else
-        Bolt::Node::Failure.new(task_result['exit_code'], result_output)
-      end
+      Bolt::CommandResult.new(result.value['stdout'], result.value['stderr'], result.value['exit_code'])
```
MikaelSmith: We expect to use Bolt::CommandResult for scripts as well?

lib/bolt/node/ssh.rb
```diff
@@ -109,45 +108,55 @@ def execute(command, sudoable: false, **options)
       end
       session_channel.wait
 
-      if status[:exit_code].zero?
+      if result_output.exit_code.zero?
         @logger.debug { "Command returned successfully" }
-        Bolt::Node::Success.new(result_output.stdout.string, result_output)
       else
         @logger.info { "Command failed with exit code #{status[:exit_code]}" }
```
MikaelSmith: `status[:exit_code]` no longer exists, should be `result_output.exit_code`.

lib/bolt/node/errors.rb
MikaelSmith: I guess every BaseError has an issue code, so that makes sense.

lib/bolt/node/ssh.rb
```diff
@@ -109,45 +108,55 @@ def execute(command, sudoable: false, **options)
...
-      make_tempdir.then do |dir|
-        (yield dir).ensure do
-          execute("rm -rf '#{dir}'")
+      dir = make_tempdir
+      begin
+        yield dir
+      ensure
+        output =  execute("rm -rf '#{dir}'")
+        if output.exit_code != 0
+          logger.warn("Failed to clean up tempdir '#{dir}': #{result.stderr}")
```
MikaelSmith: Where does `result` come from? I think you mean `output`.

lib/bolt/node/ssh.rb
```diff
@@ -167,14 +176,20 @@ def make_wrapper_stringio(task_path, stdin)
...
+
     def with_task_wrapper(remote_task, dir, stdin)
       wrapper = make_wrapper_stringio(remote_task, stdin)
       command = "#{dir}/wrapper.sh"
-      _upload(wrapper, command).then do
-        execute("chmod u+x '#{command}'")
-      end.then do
-        yield command
-      end
+      # TODO: update with_remote script to support this
```
MikaelSmith: I'm not sure what this TODO means.

lib/bolt/node/winrm.rb
```diff
@@ -38,7 +38,7 @@ def disconnect
     end
 
     def shell_init
-      return Bolt::Node::Success.new if @shell_initialized
+      return nil if @shell_initialized
```
MikaelSmith: Should we merge BOLT-72 before this? Seems like they might overlap.

lib/bolt/node/winrm.rb
```diff
@@ -220,12 +221,18 @@ def process_from_extension(path)
     end
 
     def _upload(source, destination)
+      write_remote_file(source, destination)
+      Bolt::Result.new
+    # TODO: we should rely on the executor for this
```
MikaelSmith: When do you plan to address the TODO?

lib/bolt/node/winrm.rb
```diff
@@ -220,12 +221,18 @@ def process_from_extension(path)
...
+    rescue StandardError => ex
+      Bolt::Result.from_exception(ex)
+    end
+
+    def write_remote_file(source, destination)
       @logger.debug { "Uploading #{source} to #{destination}" }
       fs = ::WinRM::FS::FileManager.new(@connection)
+      # TODO: raise FileError here if this fails
```
MikaelSmith: That seems simple, still planning to add it?

lib/bolt/node/winrm.rb
```diff
@@ -236,46 +243,49 @@ def make_tempdir
...
-        dir = value
-        ext = File.extname(file)
-        ext = VALID_EXTENSIONS.include?(ext) ? ext : '.ps1'
-        dest = "#{dir}\\#{File.basename(file, '.*')}#{ext}"
-        Bolt::Node::Success.new
-      end.then do
-        _upload(file, dest)
-      end.then do
+      ext = File.extname(file)
+      ext = VALID_EXTENSIONS.include?(ext) ? ext : '.ps1'
```
MikaelSmith: Ah, this is how we support scripts without an extension on Windows.

lib/bolt/node/winrm.rb
```diff
@@ -236,46 +243,49 @@ def make_tempdir
...
 PS
-        result
       end
     end
 
     def _run_command(command)
-      execute(command)
+      output = execute(command)
+      Bolt::CommandResult.from_output(output)
+    # TODO: we should rely on the executor for this
```
MikaelSmith: Is this another ticket you'll file?

lib/bolt/outputter/human.rb
```diff
@@ -1,27 +1,80 @@
 module Bolt
   class Outputter
     class Human < Bolt::Outputter
+      COLORS = { red: "31",
```
MikaelSmith: Do these get translated on Windows? Or does this only work in *nix?

lib/bolt/outputter/json.rb
```diff
@@ -15,10 +15,17 @@ def print_head
...
+        when :node_result
+          print_result(node, event[:result])
+        end
+      end
+
       def print_result(node, result)
         item = {
           name: node.uri,
-          status: result.is_a?(Bolt::ErrorResult) ? 'failure' : 'success',
+          status: result.success? ? 'failure' : 'success',
```
MikaelSmith: This seems reversed... if `success?` is true, wouldn't we use `success`?

lib/bolt/result.rb
```diff
@@ -1,69 +1,71 @@
...
 
     def to_result
-      {
-        '_error' => {
-          'issue_code' => @issue_code,
-          'kind' => @kind,
-          'msg' => @message
-        }
-      }
+      # TODO: This should be to_h but we need to update the plan functions to
```
MikaelSmith: Is there a PUP ticket related to that? Or a BOLT ticket (now that we moved run_* to Bolt)?

MikaelSmith:
I didn't look at tests yet, going to leave that for after initial feedback is addressed.

lib/bolt/cli.rb
adreyer: It's the exit code for bolt

lib/bolt/node/winrm.rb
Iristyle: Yes, please merge #153 first

lib/bolt/cli.rb
MikaelSmith: Ah, ok.

lib/bolt/executor.rb
adreyer: That is how the boltlib functions call it
https://github.com/puppetlabs/bolt/blob/master/modules/boltlib/lib/puppet/functions/run_task.rb#L76

lib/bolt/node/errors.rb
adreyer: issue code is to help with reverse translation in the future. It should match the message template.

lib/bolt/node/orch.rb
adreyer: Yeah I considered renaming it. Any suggestions?

MikaelSmith: It seems ok, just surprised me initially.

lib/bolt/node/ssh.rb
adreyer: There are too many similar functions here. We should be able to refactor them into fewer more like the winrm transport.

lib/bolt/node/winrm.rb
adreyer: It looks like you did that?

lib/bolt/node/winrm.rb
adreyer: As part of the transport stabilization

lib/bolt/node/winrm.rb
adreyer: I'm not sure what the failure cases look like or what message to generate

lib/bolt/node/winrm.rb
MikaelSmith: Yes, done.

lib/bolt/outputter/human.rb
adreyer: presumably the latter

lib/bolt/result.rb
adreyer: unfortunately ExecutionResult has a from_bolt method that has to change.

MikaelSmith: That's do-able, just curious what the plan was to do it.

lib/bolt/outputter/human.rb
```diff
@@ -1,27 +1,80 @@
...
+        end
+
+        if result.message
+          @stream.puts(remove_trail(indent(2, result.message)))
+        end
+
+        if result.instance_of? Bolt::TaskResult
+          @stream.puts(indent(2, ::JSON.pretty_generate(result.value)))
+        elsif result.instance_of? Bolt::CommandResult
+          @stream.puts(indent(2, "STDOUT:"))
```
MikaelSmith: Should thees be printed even if stdout or stderr are empty?

adreyer: I think it's useful to at least have the stdout key printed to make it clear it was empty.

PR #169, opened 2017-11-22
(BOLT-66) Use Puppet type system to validate task arguments

MikaelSmith:
Example output
```
$ bolt task run echo message='{"a": 1}' foo=1 -n node1 --modulepath site
Error: 'new_Echo' expects one of:
  (Struct[{'message' => String, 'exit_code' => Optional[Integer[0, 255]]}] hash)
    rejected: parameter 'hash' unrecognized key 'foo'
  (String message, Integer[0, 255] exit_code?)
    rejected: parameter 'message' expects a String value, got Struct on node my-laptop
Task parameters did not match

$ bolt task run foo message='{"a": 1}' -n node1 --modulepath site
Could not find task 'foo'
```

### DIFF
--- a/lib/bolt/cli.rb
+++ b/lib/bolt/cli.rb
@@ -414,12 +414,13 @@ HELP
             when 'task'
               task_name = options[:object]
 
-              path, metadata = load_task_data(task_name, options[:modulepath])
-              input_method = metadata['input_method']
+              task_type = load_task_data(task_name, options[:task_options], options[:modulepath])
+              metadata = task_type._pcore_type
+              input_method = metadata['input_method'].value
 
               input_method ||= 'both'
               executor.run_task(
-                nodes, path, input_method, options[:task_options]
+                nodes, task_type.executable_path, input_method, options[:task_options]
               ) do |node, event|
                 outputter.print_event(node, event)
               end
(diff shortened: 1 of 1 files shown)


MikaelSmith:
I've noted some issues that still need to be worked through on the ticket. Error handling is not ideal, we end up restricting parameters to those that don't collide with other metadata keys, and tasks that don't declare parameters will error if any are supplied.

lib/bolt/cli.rb
```diff
@@ -485,38 +486,30 @@ def outputter
...
-            if data[:metadata_file]
-              JSON.parse(File.read(data[:metadata_file]))
-            else
-              {}
+          Puppet.initialize_settings(cli)
+          Puppet::Pal.in_tmp_environment('bolt', modulepath: [BOLTLIB_PATH] + modulepath) do |pal|
+            pal.with_script_compiler do |compiler|
+              # Update once Pal includes task_signature
+              #task_type = compiler.task_signature(name)
+              task_type = Puppet.lookup(:loaders).private_environment_loader.load(:type, name)
```
MikaelSmith: @mruzicka this call will get a type object. From it you can query `task_type.attributes['description']` for the description, and likewise for other metadata. It currently also includes all parameters. I'm looking to decouple these at some point.

lib/bolt/cli.rb
```diff
@@ -485,38 +486,30 @@ def outputter
...
-        env = Puppet::Node::Environment.create('bolt', modulepath)
-        Puppet.override(environments: Puppet::Environments::Static.new(env)) do
-          data = Puppet::InfoService::TaskInformationService.task_data(
-            env.name, module_name, name
-          )
-
-          file = data[:files].find { |f| File.basename(f, '.*') == file_name }
-          if file.nil?
-            raise Bolt::CLIError, "Failed to load task file for '#{name}'"
+        Dir.mktmpdir('bolt') do |dir|
```
MikaelSmith: Much of this may be unnecessary... would be interested in any suggestions @hlindberg. I think for type lookup, we primarily need the modulepath.

MikaelSmith:
This makes `bolt task run` really slow. Using a simple `echo` task:
- with current `master`: 0.37 seconds, almost as fast as `command run 'id -u -n'`
- this PR with no parameter metadata: 2.76 seconds
- this PR with parameter metadata: 7.97 seconds

That's *highly* concerning.

lib/bolt/cli.rb
```diff
@@ -480,38 +481,30 @@ def outputter
...
-              JSON.parse(File.read(data[:metadata_file]))
-            else
-              {}
+          Puppet.initialize_settings(cli)
+          Puppet::Pal.in_tmp_environment('bolt', modulepath: [BOLTLIB_PATH] + modulepath) do |pal|
+            pal.with_script_compiler do |compiler|
+              # Update once Pal includes task_signature (PUP-8202)
+              #task_type = compiler.task_signature(name)
+              task_type = Puppet.lookup(:loaders).private_environment_loader.load(:type, name)
+              task = compiler.create(task_type, args) if task_type
```
hlindberg: PUP-8171 adds API for typechecking plans. A similar method for task (getting a signature and then checking the args against it is yet to be added).

lib/bolt/cli.rb
hlindberg: I think a directory is required, you could have an empty tmp directory in bolt that is always used for this. Later you may for example want to have a default hiera.yaml there.

hlindberg: sorry, you can also do this `Puppet::Pal.in_tmp_environment` and just give a modulepath.

hlindberg:
on my machine the PAL unit tests (75 examples, including one that runs facter) takes roughly 4 seconds - so the 7sec time is extreme

ripienaar:
On my machines with the facter 2 that bolt includes I had to rm the ec2 fact because it’s crap in old facter. Instant speed up. Worth a try.

MikaelSmith:
Updated, now down to 1.6 seconds, which is barely tolerable. For some reason we still have to initialize settings, might be faster if we could eliminate that.

mruzicka:
Superseded by https://github.com/puppetlabs/bolt/pull/190

PR #170, opened 2017-11-22
(maint) Don't pretty-print plan results

adreyer:
Plan results may be any object and cannot always be pretty printed.


MikaelSmith:
Should we try and fallback, or was that never in the design?

PR #171, opened 2017-11-27
(BOLT-44) Don't print stdout or stderr unless they have content

adreyer:


lib/bolt/outputter/human.rb
```diff
@@ -59,10 +59,14 @@ def print_result(node, result)
         if result.instance_of? Bolt::TaskResult
           @stream.puts(indent(2, ::JSON.pretty_generate(result.value)))
         elsif result.instance_of? Bolt::CommandResult
-          @stream.puts(indent(2, "STDOUT:"))
-          @stream.puts(indent(4, result.stdout))
-          @stream.puts(indent(2, "STDERR:"))
-          @stream.puts(indent(4, result.stderr))
+          unless result.stdout.strip.empty?
```
Iristyle: Can `result.stdout` or `result.stderr` be nil?

adreyer: Not with the existing transports but I'll set that to be sure.

zreichert:
@adreyer  did you mean to fixup that commit?

PR #172, opened 2017-11-27
(BOLT-236) Allow private key path for ssh to be specified with --private-key

adreyer:


spec/bolt/node/ssh_spec.rb
```diff
@@ -106,7 +107,122 @@ def result_value(stdout = nil, stderr = nil, exit_code = 0)
     end
   end
 
-  context "when running_commands" do
+  context "when executing with private key" do
```
nicklewis: Is it important that we run all of these tests with both a private key and a password?

adreyer: probably not that important.

spec/bolt/node/ssh_spec.rb
```diff
@@ -13,6 +13,7 @@
   let(:user) { ENV['BOLT_SSH_USER'] || "vagrant" }
   let(:password) { ENV['BOLT_SSH_PASSWORD'] || "vagrant" }
   let(:port) { ENV['BOLT_SSH_PORT'] || 2224 }
+  let(:key) { ENV['BOLT_SSH_KEY'] || Dir[".vagrant/**/private_key"] }
```
nicklewis: This looks like it will pass an array as `:key` in the default case. That doesn't reflect how this will actually work in practice.

adreyer: if we don't pass an array here tests may break if users switch vagrant providers.

adreyer: I'm ok just testing the string option in travis but if you want I can shell out to `vagrant provider linux` to find a single private key.

nicklewis: I'm fine with this, knowing there's an actual reason it uses a glob.

PR #173, opened 2017-11-27
(maint) start ssh agent for travis tests to silence log messages

adreyer:


Iristyle:
Confirmed this removes all the fun messages like:

```
2017-11-27T17:22:16.961940  ERROR net.ssh.authentication.agent[1ff4668]: could not connect to ssh-agent: Agent not configured
```

PR #174, opened 2017-11-28
(BOLT-227) Catch wrong password and fail the node

adreyer:


adreyer:
moved to https://github.com/puppetlabs/bolt/pull/174

nicklewis:
Should we handle `sudo: sorry, you must have a tty to run sudo` and `Sorry, user foo is not allowed to execute [...]` here also or are those separate tickets?

adreyer:
I think those are separate. This is much worse since it hangs while those will error and appear at least in debug logs

PR #175, opened 2017-11-28
(BOLT-225) use scp instead of sftp to copy files

adreyer:
This commit changes the ssh transport to use scp instead of sftp for
file uploads. This allows users to upload files, scripts and tasks
regardelss of whether sshd on the target is configured to support sftp
or the user has sftp permissions.

### DIFF
--- a/bolt.gemspec
+++ b/bolt.gemspec
@@ -26,7 +26,7 @@ Gem::Specification.new do |spec|
 
   spec.add_dependency "addressable", '< 2.5.0'
   spec.add_dependency "concurrent-ruby", "~> 1.0"
-  spec.add_dependency "net-sftp", "~> 2.0"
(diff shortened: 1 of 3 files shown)


nicklewis:
If I try to use `bolt file upload` to upload to a directory that doesn't exist, I get this:

```
Started on a7drnckeorwu9bd.delivery.puppetlabs.net...
Ran on 0 nodes in 0.13 seconds
```

Compare to the previous behavior with SFTP:

```
Started on a7drnckeorwu9bd.delivery.puppetlabs.net...
Failed on a7drnckeorwu9bd.delivery.puppetlabs.net:
  Net::SFTP::StatusException open /tmp/foo/bar/baz (2, "no such file")
Ran on 1 node in 0.13 seconds
```

PR #179, opened 2017-12-05
(BOLT-156) Add powershell input method

mruzicka:
This PR adds support for the `powershell` input method to the `Bolt::WinRM` class.
The implementation injects a base64 encoded JSON blob containing the task parameter values to the PowerShell session which is then expanded in that session into powershell's `Hashtable` data structure which is then in turn used for splatting in the invocation of the task code.

### DIFF
--- a/lib/bolt/node/winrm.rb
+++ b/lib/bolt/node/winrm.rb
@@ -1,6 +1,7 @@
 require 'winrm'
 require 'winrm-fs'
 require 'bolt/result'
+require 'base64'
 
 module Bolt
   class WinRM < Node
@@ -47,6 +48,9 @@ $ENV:RUBYLIB = "${ENV:ProgramFiles}\\Puppet Labs\\Puppet\\puppet\\lib;" +
   "${ENV:ProgramFiles}\\Puppet Labs\\Puppet\\facter\\lib;" +
   "${ENV:ProgramFiles}\\Puppet Labs\\Puppet\\hiera\\lib;" +
   $ENV:RUBYLIB
+
+Add-Type -AssemblyName System.ServiceModel.Web, System.Runtime.Serialization
+$utf8 = [System.Text.Encoding]::UTF8
 
 function Invoke-Interpreter
 {
@@ -138,6 +142,183 @@ function Invoke-Interpreter
       }
       $process.Dispose()
     }
+  }
+}
+
+function Write-Stream {
+  PARAM(
+    [Parameter(Position=0)] $stream,
+    [Parameter(ValueFromPipeline=$true)] $string
+  )
+  PROCESS {
+    $bytes = $utf8.GetBytes($string)
+    $stream.Write( $bytes, 0, $bytes.Length )
+  }
+}
+
+function Convert-JsonToXml {
+  PARAM([Parameter(ValueFromPipeline=$true)] [string[]] $json)
+  BEGIN {
+    $mStream = New-Object System.IO.MemoryStream
+  }
+  PROCESS {
+    $json | Write-Stream -Stream $mStream
+  }
+  END {
+    $mStream.Position = 0
+    try {
+      $jsonReader = [System.Runtime.Serialization.Json.JsonReaderWriterFactory]::CreateJsonReader($mStream,[System.Xml.XmlDictionaryReaderQuotas]::Max)
+      $xml = New-Object Xml.XmlDocument
+      $xml.Load($jsonReader)
+      $xml
+    } finally {
+      $jsonReader.Close()
+      $mStream.Dispose()
+    }
+  }
+}
+
+Function ConvertFrom-Xml {
+  [CmdletBinding(DefaultParameterSetName="AutoType")]
+  PARAM(
+    [Parameter(ValueFromPipeline=$true,Mandatory=$true,Position=1)] [Xml.XmlNode] $xml,
+    [Parameter(Mandatory=$true,ParameterSetName="ManualType")] [Type] $Type,
+    [Switch] $ForceType
+  )
+  PROCESS{
+    if (Get-Member -InputObject $xml -Name root) {
+      return $xml.root.Objects | ConvertFrom-Xml
+    } elseif (Get-Member -InputObject $xml -Name Objects) {
+      return $xml.Objects | ConvertFrom-Xml
+    }
+    $propbag = @{}
+    foreach ($name in Get-Member -InputObject $xml -MemberType Properties | Where-Object{$_.Name -notmatch "^__|type"} | Select-Object -ExpandProperty name) {
+      Write-Debug "$Name Type: $($xml.$Name.type)" -Debug:$false
+      $propbag."$Name" = Convert-Properties $xml."$name"
+    }
+    if (!$Type -and $xml.HasAttribute("__type")) { $Type = $xml.__Type }
+    if ($ForceType -and $Type) {
+      try {
+        $output = New-Object $Type -Property $propbag
+      } catch {
+        $output = New-Object PSObject -Property $propbag
+        $output.PsTypeNames.Insert(0, $xml.__type)
+      }
+    } elseif ($propbag.Count -ne 0) {
+      $output = New-Object PSObject -Property $propbag
+      if ($Type) {
+        $output.PsTypeNames.Insert(0, $Type)
+      }
+    }
+    return $output
+  }
+}
+
+Function Convert-Properties {
+  PARAM($InputObject)
+  switch ($InputObject.type) {
+    "object" {
+      return (ConvertFrom-Xml -Xml $InputObject)
+    }
+    "string" {
+      $MightBeADate = $InputObject.get_InnerText() -as [DateTime]
+      ## Strings that are actually dates (*grumble* JSON is crap)
+      if ($MightBeADate -and $propbag."$Name" -eq $MightBeADate.ToString("G")) {
+        return $MightBeADate
+      } else {
+        return $InputObject.get_InnerText()
+      }
+    }
+    "number" {
+      $number = $InputObject.get_InnerText()
+      if ($number -eq ($number -as [int])) {
+        return $number -as [int]
+      } elseif ($number -eq ($number -as [double])) {
+        return $number -as [double]
+      } else {
+        return $number -as [decimal]
+      }
+    }
+    "boolean" {
+      return [bool]::parse($InputObject.get_InnerText())
+    }
+    "null" {
+      return $null
+    }
+    "array" {
+      [object[]]$Items = $(foreach( $item in $InputObject.GetEnumerator() ) {
+        Convert-Properties $item
+      })
+      return $Items
+    }
+    default {
+      return $InputObject
+    }
+  }
+}
+
+Function ConvertFrom-Json2 {
+  [CmdletBinding()]
+  PARAM(
+    [Parameter(ValueFromPipeline=$true,Mandatory=$true,Position=1)] [string] $InputObject,
+    [Parameter(Mandatory=$true)] [Type] $Type,
+    [Switch] $ForceType
+  )
+  PROCESS {
+    $null = $PSBoundParameters.Remove("InputObject")
+    [Xml.XmlElement]$xml = (Convert-JsonToXml $InputObject).Root
+    if ($xml) {
+      if ($xml.Objects) {
+        $xml.Objects.Item.GetEnumerator() | ConvertFrom-Xml @PSBoundParameters
+      } elseif ($xml.Item -and $xml.Item -isnot [System.Management.Automation.PSParameterizedProperty]) {
+        $xml.Item | ConvertFrom-Xml @PSBoundParameters
+      } else {
+        $xml | ConvertFrom-Xml @PSBoundParameters
+      }
+    } else {
+      Write-Error "Failed to parse JSON with JsonReader" -Debug:$false
+    }
+  }
+}
+
+function ConvertFrom-PSCustomObject
+{
+  PARAM([Parameter(ValueFromPipeline = $true)] $InputObject)
+  PROCESS {
+    if ($null -eq $InputObject) { return $null }
+
+    if ($InputObject -is [System.Collections.IEnumerable] -and $InputObject -isnot [string]) {
+      $collection = @(
+        foreach ($object in $InputObject) { ConvertFrom-PSCustomObject $object }
+      )
+
+      $collection
+    } elseif ($InputObject -is [System.Management.Automation.PSCustomObject]) {
+      $hash = @{}
+      foreach ($property in $InputObject.PSObject.Properties) {
+        $hash[$property.Name] = ConvertFrom-PSCustomObject $property.Value
+      }
+
+      $hash
+    } else {
+      $InputObject
+    }
+  }
+}
+
+function Get-ContentAsJson
+{
+  [CmdletBinding()]
+  PARAM(
+    [Parameter(Mandatory = $true)] $Text,
+    [Parameter(Mandatory = $false)] [Text.Encoding] $Encoding = [Text.Encoding]::UTF8
+  )
+
+  # using polyfill cmdlet on PS2, so pass type info
+  if ($PSVersionTable.PSVersion -lt [Version]'3.0') {
+    $Text | ConvertFrom-Json2 -Type PSObject | ConvertFrom-PSCustomObject
+  } else {
+    $Text | ConvertFrom-Json | ConvertFrom-PSCustomObject
   }
 }
 PS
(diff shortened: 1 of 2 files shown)


lib/bolt/node/winrm.rb
```diff
@@ -337,7 +518,23 @@ def _run_task(task, input_method, arguments)
           # NOTE: cannot redirect STDIN to a .ps1 script inside of PowerShell
           # must create new powershell.exe process like other interpreters
           # fortunately, using PS with stdin input_method should never happen
-          output = execute("try { &""#{remote_path}"" } catch { exit 1 }")
+          output = execute(<<-PS)
```
Iristyle: Since `stdin` is nil, that means we're not using the `stdin` input_method. Are the `environment` and `powershell` input methods mutually exclusive or can they be enabled concurrently as will be supported here?

lib/bolt/node/winrm.rb
```diff
@@ -337,7 +518,23 @@ def _run_task(task, input_method, arguments)
...
+& { # create a new scope
+  $_task_path = [Text.Encoding]::UTF8.GetString(
```
Iristyle: To support PS2, I believe `[Text.Encoding]` and `[Convert]` need to have the `System.` prefix, like `[System.Text.Encoding]` and `[System.Convert]`.  Since we have the JSON polyfill here to support PS2 (it's not necessary in PS3 which includes`ConvertTo-Json` already) that means we also need to make sure other PS2 compatibility fixes are in place.  PS2 testing is currently a blind spot.

We ran into this while building the PowerShell module. /cc @jpogran

lib/bolt/node/winrm.rb
```diff
@@ -337,7 +518,23 @@ def _run_task(task, input_method, arguments)
...
+    [Convert]::FromBase64String('#{Base64.encode64(remote_path)}')
+  )
+  $_task_args = Get-ContentAsJson (
```
Iristyle: Nit - `$_task_args` and `$_task_path` are not idiomatic PowerShell variable names, which tends to prefer camelCase rather than snake_case.

lib/bolt/node/winrm.rb
```diff
@@ -140,6 +144,183 @@ def shell_init
     }
   }
 }
+
+function Write-Stream {
```
Iristyle: In a separate commit we might want to think about moving this bootstrapping code to its own file.  I had a feeling it was going to grow over time, and the inline string it starting to get a little unwieldy.

spec/bolt/node/winrm_spec.rb
```diff
@@ -279,6 +279,20 @@ def stub_winrm_to_raise(klass, message)
     end
   end
 
+  it "passes arguments to the task in the powershell way", winrm: true do
+    contents = <<-PS
+      param ($Name, $Age)
+
+      Write-Host "$Name is $($Age+5) years old."
+    PS
+    arguments = { :Name => 'John Doe',
```
Iristyle: I'd suggest passing the args out of order from their definition to make sure you're not accidentally getting parameters matching by position.

spec/bolt/node/winrm_spec.rb
```diff
@@ -279,6 +279,20 @@ def stub_winrm_to_raise(klass, message)
     end
   end
 
+  it "passes arguments to the task in the powershell way", winrm: true do
+    contents = <<-PS
+      param ($Name, $Age)
```
Iristyle: Can we make this param block typed?

```powershell
[CmdletBinding()]
param (
  [String]
  $Name,
  [Int32]
  $Age
)
```

lib/bolt/node/winrm.rb
```diff
@@ -337,7 +518,23 @@ def _run_task(task, input_method, arguments)
...
           # must create new powershell.exe process like other interpreters
           # fortunately, using PS with stdin input_method should never happen
-          output = execute("try { &""#{remote_path}"" } catch { exit 1 }")
+          output = execute(<<-PS)
+& { # create a new scope
+  $_task_path = [Text.Encoding]::UTF8.GetString(
+    [Convert]::FromBase64String('#{Base64.encode64(remote_path)}')
+  )
+  $_task_args = Get-ContentAsJson (
+    [Text.Encoding]::UTF8.GetString(
```
Iristyle: Same notes here about `System.`

lib/bolt/node/winrm.rb
```diff
@@ -337,7 +518,23 @@ def _run_task(task, input_method, arguments)
           # NOTE: cannot redirect STDIN to a .ps1 script inside of PowerShell
           # must create new powershell.exe process like other interpreters
           # fortunately, using PS with stdin input_method should never happen
-          output = execute("try { &""#{remote_path}"" } catch { exit 1 }")
+          output = execute(<<-PS)
+& { # create a new scope
+  $_task_path = [Text.Encoding]::UTF8.GetString(
+    [Convert]::FromBase64String('#{Base64.encode64(remote_path)}')
```
Iristyle: When would the `remote_path` need to be base64 encoded / decoded? If this is a concern, we should probably have a test demonstrating the problem...

Also a better way to avoid escaping rules with the literal (assuming it's even necessary) would be to just use a here-string like:

```powershell
$taskPath = @'
#{remote_path}
@'
```

lib/bolt/node/winrm.rb
mruzicka: We should, this is actually a copypasta from https://github.com/puppetlabs/pxp-agent/blob/409c82cd4ffa16a65e1c231b444619774e520cd9/exe/PowershellShim-Helper.ps1 which is used by pxp-agent. If we move this to a separate file it could potentially be shared.

lib/bolt/node/winrm.rb
mruzicka: I'm going to do some testing around that, especially I want to know what happens if a script is passed the argument hash while it doesn't define the `param ()`, or defines a different set of params than those passed in the hash.

Iristyle: Yeah, definitely a useful scenario to think about adding tests for.

For the situation you describe, given extraneous parameters, PowerShell will explode like this when params are splatted (example taken from REPL):

```powershell
PS /Users/Iristyle/source/puppetlabs-dsc_lite> function Foo                                                        >> {                                                                                                               >> [CmdletBinding()]                                                                                               >> param (                                                                                                         >>   [String]                                                                                                      >>   $foo = 'bar'                                                                                                  >> )                                                                                                               >> Write-Host $foo                                                                                                 >> }
PS /Users/Iristyle/source/puppetlabs-dsc_lite> $badParam = @{ 'baz' = 1234 }
PS /Users/Iristyle/source/puppetlabs-dsc_lite> Foo @badParam
Foo : A parameter cannot be found that matches parameter name 'baz'.
At line:1 char:5
+ Foo @badParam
+     ~~~~~~~~~
    + CategoryInfo          : InvalidArgument: (:) [Foo], ParameterBindingException
    + FullyQualifiedErrorId : NamedParameterNotFound,Foo
```

For *missing* parameters that are not decorated with `[Parameter(Mandatory = $true)]`, standard PowerShell behavior applies - i.e. there are no exceptions thrown for parameter validation issues.

```
PS /Users/Iristyle/source/puppetlabs-dsc_lite> Foo
bar

PS /Users/Iristyle/source/puppetlabs-dsc_lite> Foo -foo hi
hi
```

lib/bolt/node/winrm.rb
mruzicka: One more note here: I originally used:
```powershell
Add-Type -Assembly System.Web.Extensions
```
then I could use the `JavaScriptSerializer` class from that assembly for parsing the JSON like this:
```powershell
$_task_args = (New-Object System.Web.Script.Serialization.JavaScriptSerializer).DeserializeObject(
   $utf8.GetString([Convert]::FromBase64String('#{Base64.encode64(JSON.dump(arguments))}'))
)
```
which worked on my development machine, but not on Windows Nano VM spun up by vagrant.
If we could use that class we could drop all of the code added here.
Could we declare the assembly as a requirement?

Iristyle: I don't think we can guarantee `System.Web.Extensions` is available on all target machines / PowerShell / .NET Framework versions unfortunately.

mruzicka: oh well ...

lib/bolt/node/winrm.rb
mruzicka: ok

lib/bolt/node/winrm.rb
mruzicka: How about the leading underscore? (I used it deliberately so as to make the variables unlikely to conflict with any variables in the task.)

spec/bolt/node/winrm_spec.rb
mruzicka: ok

spec/bolt/node/winrm_spec.rb
mruzicka: Did that even before reading this ... ;-)

lib/bolt/node/winrm.rb
mruzicka: Specifically: Would `$_taskPath` & `$_taskArg` be ok?

spec/bolt/node/winrm_spec.rb
mruzicka: OTH, this will mask the fact that the powershell input method provides the values of the correct types even without the need to specify the types in the `param ()` section.
When the types are specified in the `param ()` section the values will be coerced to the correct types irrespective of types of the values in the hash provided by the powershell input method.
So I'd rather keep them untyped.

lib/bolt/node/winrm.rb
Iristyle: Ah, I understand why you did this now.  Per my other comment at https://github.com/puppetlabs/bolt/pull/179#discussion_r155053986, I don't think you need `$_task_path` assigned to a variable (but double check my comment there).

If you want to prevent the child scope inheriting from the parent you can use `$private:foo = 'bar'` (the long form is `New-Variable -Name foo -Value 'bar' -Option private`)

```
PS /Users/Iristyle/source/bolt> echo $foo
PS /Users/Iristyle/source/bolt> cat ./foo.ps1
Write-Host $foo
Write-Host $script:foo
Write-Host $global:foo
PS /Users/Iristyle/source/bolt> echo $foo
PS /Users/Iristyle/source/bolt> $private:foo = 'bar'
PS /Users/Iristyle/source/bolt> echo $foo
bar
PS /Users/Iristyle/source/bolt> ./foo.ps1



PS /Users/Iristyle/source/bolt> & ./foo.ps1



```

lib/bolt/node/winrm.rb
mruzicka: What about character encoding: UTF8 vs. UTF16?
By using here-string, you lock yourself to the encoding assumed by powershell when parsing the source.
But it my not be a concern here.

spec/bolt/node/winrm_spec.rb
Iristyle: Yes, PowerShell will do a lot of automaticion coercion - you can see some of the details in the gist I put together (which was the basis for the tasks shim in PE) - https://gist.github.com/Iristyle/3b37ad44c93e3e00502d799344294d93

I can understand why'd you want to make sure JSON typed values (i.e. string, number, array, bool, null) all flow through to untyped params on the PS side as you suggest.

I can also understand where you'd want to make sure the coercion behaves correctly for typed params on the PS side (i.e. to make sure JSON types properly coerce to their expected corresponding PS types) - if for no other reason than to document in tests what coercions are allowed. The norm for PowerShell scripts for end users will have the `[Type]` attributes for all parameters from what I've seen - parameter typing is pretty idiomatic for PS.

lib/bolt/node/winrm.rb
mruzicka: $private:taskArgs is the winner ...

spec/bolt/node/winrm_spec.rb
mruzicka: I have added the types to the other two test, it makes a difference here:
https://github.com/puppetlabs/bolt/pull/179/files#diff-a0e64205eda75ed9e253f283c7f02df6R336
Without the typed parameter, the expected string was:
```
foo:  (String), bar: 30 (String)\r\n
```

mruzicka: I've updated the first test to use a regexp to match the output to account for the possible differences in the data types (`Int32` vs. `Int64`, `Decimal` vs. `Double`).

lib/bolt/node/winrm.rb
mruzicka: I more less reverted the original handling of the `remote_path`, except I fixed the quoting, so now we have:
```ruby
"try { & '#{remote_path}' } catch { exit 1 }"
```
instead of the original:
```ruby
"try { &""#{remote_path}"" } catch { exit 1 }"
```

mruzicka:
I think I've addressed all the comments and the checks are now passing too, yay!

lib/bolt/node/winrm.rb
```diff
@@ -333,15 +514,27 @@ def _run_task(task, input_method, arguments)
...
-          output = execute_process(path, args, stdin)
-        end
+        output =
+          if powershell_file?(remote_path) && stdin.nil?
+            # NOTE: cannot redirect STDIN to a .ps1 script inside of PowerShell
+            # must create new powershell.exe process like other interpreters
+            # fortunately, using PS with stdin input_method should never happen
+            if input_method == 'powershell'
+              execute(<<-PS)
+& { # create a new scope
```
Iristyle: What was the reason for using call operator `&` against the anonymous scriptblock?  I think you've covered the scope inheritance situation you were concerned about with the `$private:taskArgs` - but maybe I'm missing something?

lib/bolt/node/winrm.rb
mruzicka: I've made the `environment` & `powershell` methods mutually exclusive, so the parameters hash is only passed to the task when the `powershell` input method is specified.

lib/bolt/node/winrm.rb
mruzicka: True, removing.

lib/bolt/node/winrm.rb
```diff
@@ -333,15 +514,27 @@ def _run_task(task, input_method, arguments)
...
+  $private:taskArgs = Get-ContentAsJson (
+    $utf8.GetString([System.Convert]::FromBase64String('#{Base64.encode64(JSON.dump(arguments))}'))
+  )
+  try { & '#{remote_path}' @taskArgs } catch { exit 1 }
```
Iristyle: On Windows, `"` is a reserved character for filenames, but `'` is allowed, which will break this interpolation - https://msdn.microsoft.com/en-us/library/windows/desktop/aa365247%28v=vs.85%29.aspx?f=255&MSPPError=-2147217396

I'm responsible for the original `&""#{remote_path}""` in https://github.com/puppetlabs/bolt/commit/ea68173084041cf5fddf5f97a2febfce18834dda#diff-0ce90a36bcfd738e65f1aa3cc4892943R330 ... and it appears I did it that way because I wanted `"`, but had to escape as `""` given a Ruby interpolated string.

You mentioned a problem with the quoting here - what were the specifics?

mruzicka:
```
$ ruby -e 'str = "text"; puts ">>""#{str}""<<"'
>>text<<
```
i.e. it seems to me to be equivalent to:
```ruby
">>#{str}<<"
```

Further I was assuming that the string is passed to a running PowerShell session by some other means than a Windows cmd.exe command line, if that assumption is wrong, then the single quotes cannot be used for the reason you point out.

Iristyle: Yeah, you're totally right - I screwed up the initial implementation there.

Original intent was supposed to wrap the path with double quotes in the event it had spaces in it (though it shouldn't). I should have used `\"` instead of `""` (was probably confusing the Ruby interpolation rules with the shell escaping rules!)

```
[1] pry(main)> foo = 'c:\program files\file\test script.ps1'
=> "c:\\program files\\file\\test script.ps1"
[2] pry(main)> "& ""#{foo}"""
=> "& c:\\program files\\file\\test script.ps1"
[3] pry(main)> "& \"#{foo}\""
=> "& \"c:\\program files\\file\\test script.ps1\""
```

Good catch on my bug! I think we still do want to use double quotes rather than single to wrap the path given the Windows filename rules...

mruzicka: I misread your comment, sorry. I thought the problem was quoting on the windows command line ...
Now I see your concern is the possibility of having a `'` in a windows filename, in the case the most appropriate quoting would IMO be:
```ruby
%(try { & "#{remote_path}" } catch { exit 1 })
```

Iristyle:
A couple of additional things to track down / file tix for @adreyer:

* PS2 / PS4 testing ... do we need Beaker or can we use custom images with AppVeyor? Some info about images in Rackspace Hyper-V cloud / GCE - https://www.appveyor.com/docs/build-environment/ (cc @jcoconnor)
* Let's move bootstrapping code currently in `winrm.rb` to a separate PS1 file?
* Some docs about `powershell` input method only applying to tasks and not scripts (plausibly I can create a scenario with multiple parameter sets where PS cannot disambiguate how to call the script given only ordered parameters)

PR #180, opened 2017-12-05
(BOLT-241) Load config from file

adreyer:
This commit adds the ability to load config options from a file. This
required a small refactor to seperate config-like options from
command-options. Config like options are all flags and are passed around
with config but may not be read from the config file. The Command
options are those that are not prefaced with flags and are only exposed
to the execute function of the CLI not passed around with the config.

There is a further distinction between global config options and those
that are transport specific. Transport specific options are specified
for each transport in the config file but only map to a single flag for
all transports as command line flags. Most of these flags only work with
the ssh transport now.

### DIFF
--- a/lib/bolt/config.rb
+++ b/lib/bolt/config.rb
@@ -1,37 +1,140 @@
 require 'logger'
+require 'yaml'
 
 module Bolt
   Config = Struct.new(
     :concurrency,
     :format,
-    :insecure,
     :log_destination,
     :log_level,
-    :password,
-    :run_as,
-    :sudo,
-    :sudo_password,
+    :modulepath,
     :transport,
-    :key,
-    :tty,
-    :user
+    :transports
   ) do
+
     DEFAULTS = {
       concurrency: 100,
-      tty: false,
-      insecure: false,
       transport: 'ssh',
+      format: 'human',
       log_level: Logger::WARN,
       log_destination: STDERR
     }.freeze
 
+    TRANSPORT_OPTIONS = %i[insecure password run_as sudo sudo_password key tty user].freeze
+
+    TRANSPORT_DEFAULTS = {
+      insecure: false,
+      tty: false
+    }.freeze
+
+    TRANSPORTS = %i[ssh winrm pcp].freeze
+
     def initialize(**kwargs)
       super()
       DEFAULTS.merge(kwargs).each { |k, v| self[k] = v }
+
+      self[:transports] ||= {}
+      TRANSPORTS.each do |transport|
+        unless self[:transports][transport]
+          self[:transports][transport] = {}
+        end
+        TRANSPORT_DEFAULTS.each do |k, v|
+          unless self[:transports][transport][k]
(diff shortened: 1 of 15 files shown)


lib/bolt/config.rb
```diff
@@ -1,37 +1,140 @@
...
+      raise Bolt::CLIError, "Could not parse config file: #{path}"
+    rescue IOError, SystemCallError
+      raise Bolt::CLIError, "Could not read config file: #{path}"
     end
 
-    def escalate?
-      sudo || run_as
+    def update_from_file(data)
+      if data['modulepath']
+        self[:modulepath] = data['modulepath']
```
adreyer: This needs to be split

lib/bolt/config.rb
```diff
@@ -1,37 +1,143 @@
...
+          end
+        end
+      end
+    end
+
+    def default_path
+      path = ['.puppetlabs', 'bolt.yml']
+      root_path = '~'
+      if ENV['OS'] == 'Windows_NT'
+        unless root_path == ENV['USERPROFILE']
```
nicklewis: I'm not sure I understand what this code is doing.

adreyer: Attempting to discover the '~' equivalent on windows to load the config file. I have not tested it yet.
 It's copied from https://github.com/puppetlabs/leatherman/blob/master/file_util/src/file.cc#L115

joshcooper: ruby should expand `~` on Windows based on those environment variables, e.g. `C:\Users\Administrator`

adreyer: ah that's much better. i don't know why I thought it didn't

lib/bolt/config.rb
```diff
@@ -1,37 +1,140 @@
...
+    end
+
+    def read_config_file(path)
+      path_passed = path
+      path ||= default_path
+      path = File.expand_path(path)
+      # safe_load doesn't work with psych in ruby 2.0
+      # The user controls the configfile so this isn't a problem
+      # rubocop:disable YAMLLoad
+      File.open(path) { |f| YAML.load(f.read) }
```
Iristyle: Don't forget to specify UTF-8 encoding here

lib/bolt/config.rb
```diff
@@ -1,37 +1,142 @@
...
+    def read_config_file(path)
+      path_passed = path
+      path ||= default_path
+      path = File.expand_path(path)
+      # safe_load doesn't work with psych in ruby 2.0
+      # The user controls the configfile so this isn't a problem
+      # rubocop:disable YAMLLoad
+      File.open(path, "r:UTF-8") { |f| YAML.load(f.read) }
+    # rubocop thinks Errno::ENOENT shadows Psych::Exception
+    # rubocop:disable ShadowedException
```
adreyer: I don't understand why this is failing commenting out the Psych::Exception rescue prevents the failure. What am I missing.

PR #184, opened 2017-12-06
(MAINT) fix specs for win 2008/PS2 host

zreichert:
improves the regexes used in spec tests to match output provided by PS2

### DIFF
--- a/spec/bolt/node/winrm_spec.rb
+++ b/spec/bolt/node/winrm_spec.rb
@@ -295,7 +295,7 @@ SHELLWORDS
                   Name: 'John Doe' }
     with_tempfile_containing('task-params-test-winrm.ps1', contents) do |file|
       expect(winrm._run_task(file.path, 'powershell', arguments).message)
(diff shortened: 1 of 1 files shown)


spec/bolt/node/winrm_spec.rb
```diff
@@ -295,7 +295,7 @@ def stub_winrm_to_raise(klass, message)
                   Name: 'John Doe' }
     with_tempfile_containing('task-params-test-winrm.ps1', contents) do |file|
       expect(winrm._run_task(file.path, 'powershell', arguments).message)
-        .to match(/^Name: John Doe \(String\), Age: 30 \(Int\d+\), Height: 5.75 \((Double|Decimal)\)\r\n$/s)
+        .to match(/^Name: John Doe \(String\), ?Age: 30 \(Int\d+\), ?Height: 5.75 \((Double|Decimal)\)\r?\n?$/s)
```
Iristyle: I think in this case it's better to use a here-string in the test, rather than fuzzy matching.

See https://github.com/puppetlabs/bolt/pull/188/commits/94ef80533b8d7d8139251cf94d7200d87ddad42e

PR #185, opened 2017-12-07
(MAINT) add rake tasks for integration testing

zreichert:
- new rake tasks for integration testing
- new rake task to update submodules
- fixes default port used in winrm spec

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -35,3 +35,22 @@ task test: %w[spec rubocop]
 task :default do
   system "rake --tasks"
 end
+
+namespace :integration do
+  desc 'Run tests that require a host System Under Test configured with WinRM'
+  RSpec::Core::RakeTask.new(:winrm) do |t|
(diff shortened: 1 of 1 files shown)


spec/bolt/node/winrm_spec.rb
```diff
@@ -10,7 +10,7 @@
   include BoltSpec::Files
 
   let(:host) { ENV['BOLT_WINRM_HOST'] || 'localhost' }
-  let(:port) { ENV['BOLT_WINRM_PORT'] || 55985 }
+  let(:port) { ENV['BOLT_WINRM_PORT'] || 5985 }
```
nicklewis: I think this was 55985 because the port is forwarded in the Vagrantfile.

zreichert: Oops, removed that change.

PR #186, opened 2017-12-07
(BOLT-147) Allow users to specify tmpdir for tasks and scripts

adreyer:


lib/bolt/cli.rb
```diff
@@ -139,6 +139,10 @@ def create_option_parser(results)
                 "Private ssh key to authenticate with (Optional)") do |key|
           results[:key] = key
         end
+        opts.on('--tmpdir DIR',
+                "The directory to upload and execute temporary files on the target(optional)") do |tmpdir|
```
nicklewis: Looks like we capitalize `Optional`, with a space before `(`.

lib/bolt/node/ssh.rb
```diff
@@ -171,7 +171,9 @@ def write_remote_file(source, destination)
     end
 
     def make_tempdir
-      result = execute('mktemp -d')
+      command = 'mktemp -d'
+      command = "TMPDIR=#{@tmpdir} #{command}" if @tmpdir
```
nicklewis: This doesn't seem to actually change the resulting dir I get from `mktemp -d` on my mac. It works on EL 7.

adreyer: hmmm maybe I should just ditch mktemp when using a tmpdir

command = "mkdir #{@tmpdir}/#{random_dir}" ?

lib/bolt/node/ssh.rb
```diff
@@ -171,11 +171,19 @@ def write_remote_file(source, destination)
     end
 
     def make_tempdir
-      result = execute('mktemp -d')
+      tmppath = nil
+      if @tmpdir
+        tmppath = "#{@tmpdir}/#{(('a'..'z').to_a + ('A'..'Z').to_a).shuffle.first(12).join}"
```
joshcooper: Is this secure, maybe `SecureRandom.uuid` instead?

adreyer: no I was trying to decide if it mattered. I guess you could prevent the expected script from actually being run.

joshcooper: ah, I'm more concerned about TOCTOU race conditions if the `tmppath` is writable by less privileged users.

lib/bolt/node/ssh.rb
```diff
@@ -171,11 +171,19 @@ def write_remote_file(source, destination)
...
     def make_tempdir
-      result = execute('mktemp -d')
+      tmppath = nil
+      if @tmpdir
+        tmppath = "#{@tmpdir}/#{SecureRandom.uuid}"
+        command = "mkdir -m 700 #{tmppath}"
+      else
+        command = 'mktemp -d'
+      end
+      command = "TMPDIR=#{@tmpdir} #{command}" if @tmpdir
```
nicklewis: Is this line necessary now?

PR #189, opened 2017-12-09
(maint) Add empty .fixtures.yml for PSH 2.6.0

Iristyle:
- puppetlabs_spec_helper 2.6.0 release introduced a new bug that
 requires that a .fixtures.yml exists with a fixtures: section in
 it. There are 3 options to address this:

 * Land a fix in PSH and wait for a new gem release
 * Pin to 2.5.1 of PSH (done already in #186)
 * Add a dummy .fixtures.yml to make the PSH problem disappear

 There is a fix up for PSH, but it may be some time before a new gem
 is released, so an empty .fixtures.yml can be used in the interim.

### DIFF
--- a/.fixtures.yml
+++ b/.fixtures.yml
@@ -1,0 +1,1 @@
+fixtures:
--- a/bolt.gemspec
+++ b/bolt.gemspec
@@ -43,7 +43,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency "win32-service", "= 0.8.8"
 
   # there is a bug in puppetlabs_spec_helper for modules without fixtures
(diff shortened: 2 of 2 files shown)


Iristyle:
Since the 2.6.0 gem was yanked, but the gemspec was updated to pick it, this will fail until a 2.6.1 is cut.

When 2.6.1 is cut, the bogus .fixtures.yml should no longer be necessary.

adreyer:
the psh release was yanked

Iristyle:
Yup @adreyer - see https://github.com/puppetlabs/bolt/pull/189#issuecomment-350495541

PR #190, opened 2017-12-10
(BOLT-66) Use puppet_pal for task parameters validation

mruzicka:
Make use of the new Puppet PAL (Puppet As a Library) facilities for validation of the task parameters to validate the task parameters passed to the `bolt task run` command.
The `run_task` puppet function provided by bolt is updated too, as the `Task` puppet type it used to employ was removed. As a consequence there are changes to its signatures.


mruzicka:
Apparently the `run_task` function spec tests need updating.

modules/boltlib/lib/puppet/functions/run_task.rb
```diff
@@ -10,42 +10,29 @@
...
-    repeated_param 'TargetOrTargets', :targets
-  end
-
-  def run_task_type(task_type, targets, task_args = nil)
-    use_args = task_args.nil? ? {} : task_args
-    task_instance = call_function('new', task_type, use_args)
-    run_task_instance(task_instance, targets)
+    param 'TargetOrTargets', :targets
+    optional_param 'Hash[String[1], Any]', :task_args
   end
```
mruzicka: This commit reduces the `run_task` function signatures to:
```ruby
dispatch :run_named_task do
   param 'String[1]', :task_name
   param 'TargetOrTargets', :targets
   optional_param 'Hash[String[1], Any]', :task_args
end

dispatch :run_named_task do
   param 'Task', :task
   param 'TargetOrTargets', :targets
   optional_param 'Hash[String[1], Any]', :task_args
end
```
Does it make sense to provide the vararg signature shown below?
```ruby
dispatch :run_task_vararg do
   param 'Task', :task
   repeated_param 'TargetOrTargets', :targets
end
```
Note that this signature could only be used for running tasks which don't require any arguments.

modules/boltlib/lib/puppet/functions/run_task.rb
```diff
@@ -59,21 +46,27 @@ def run_task_instance(task, *targets)
       )
     end
 
+    use_args = task_args.nil? ? {} : task_args
+    # TODO: Is it OK to refer to the Puppet::Pal namespace here?
+    Puppet::Pal::TaskSignature.new(task).runnable_with?(use_args) do |mismatch|
```
mruzicka: Is it OK to refer the `Puppet::Pal` namespace from the implementation of `run_task` puppet function provided by the `boltlib' module?

modules/boltlib/lib/puppet/functions/run_task.rb
mruzicka: This commit reduces the `run_task` function signatures to:
```ruby
dispatch :run_named_task do
   param 'String[1]', :task_name
   param 'TargetOrTargets', :targets
   optional_param 'Hash[String[1], Any]', :task_args
end

dispatch :run_named_task do
   param 'Task', :task
   param 'TargetOrTargets', :targets
   optional_param 'Hash[String[1], Any]', :task_args
end
```
Does it make sense to provide the vararg signature shown below?
```ruby
dispatch :run_task_vararg do
   param 'Task', :task
   repeated_param 'TargetOrTargets', :targets
end
```
Note that this signature could only be used for running tasks which don't require any arguments.

modules/boltlib/lib/puppet/functions/run_task.rb
mruzicka: Is it OK to refer the `Puppet::Pal` namespace from the implementation of the `run_task` puppet function provided by the `boltlib' module?

modules/boltlib/lib/puppet/functions/run_task.rb
```diff
@@ -59,21 +46,27 @@ def run_task_instance(task, *targets)
...
+      raise Puppet::ParseError, mismatch
```
mruzicka: Is there a more appropriate error type than the `Puppet::ParseError` we could report when the specified task arguments don't match the task definition?

modules/boltlib/lib/puppet/functions/run_task.rb
thallgren: There's a method name `#task_signature` on `Puppet::Pal::ScriptCompiler` which would allow you to retrieve a `TaskSignature` instance using a `Task` name. It requires you to get hold of the that compiler though and that's not possible just yet (this is what blocked me).

thallgren: In general, yes. Puppet::Pal is supposed to be the Puppets external API (PAL is short for Puppet As a Library).

modules/boltlib/lib/puppet/functions/run_task.rb
thallgren: I think there should be only one signature. The top one. But it needs to also inject the script compiler to be able to resolve the task_name. The injection of the script_compiler is not ready yet. See https://tickets.puppetlabs.com/browse/PUP-8237.

modules/boltlib/lib/puppet/functions/run_task.rb
mruzicka: I actually used that method [in the bolt cli](https://github.com/puppetlabs/bolt/pull/190/files#diff-b4ccc59dee0e054966d665d51e08d67eR487).
I dint't use it here because:
- I didn't know how the get the hold of the compiler 😉
- I wanted to split the lookup and the parameter checking into two functions anyway

lib/bolt/cli.rb
```diff
@@ -477,38 +476,25 @@ def outputter
...
-
-          file = data[:files].find { |f| File.basename(f, '.*') == file_name }
-          if file.nil?
-            raise Bolt::CLIError, "Failed to load task file for '#{name}'"
+    def load_task(name, args, modulepath)
+      task_signature = nil
+      error = nil
+      Puppet.initialize_settings
+      Puppet::Pal.in_tmp_environment('bolt', modulepath: [BOLTLIB_PATH] + modulepath, facts: {}) do |pal|
+        pal.with_script_compiler do |compiler|
```
thallgren: The `#run_task` command should be called within script_compiler scope, or it will not find the loader or script_compiler. That's partly why I think this approach is wrong. If the loading of Task is done outside the function, then the compiler needs to be set up twice.

modules/boltlib/lib/puppet/functions/run_task.rb
```diff
@@ -10,45 +10,42 @@
...
 
-  dispatch :run_named_task do
-    param 'String[1]', :task_type
+  # this is used from 'bolt task run'
+  dispatch :run_named_task_raw do
+    param 'String[1]', :task_name
     param 'TargetOrTargets', :targets
     optional_param 'Hash[String[1], Any]', :task_args
+    block_param
   end
```
mruzicka: The function has these signatures now:
```ruby
dispatch :run_named_task do
  param 'String[1]', :task_name
  param 'TargetOrTargets', :targets
  optional_param 'Hash[String[1], Any]', :task_args
end

dispatch :run_named_task_raw do
  param 'String[1]', :task_name
  param 'TargetOrTargets', :targets
  optional_param 'Hash[String[1], Any]', :task_args
  block_param
end
```
note that they only differ in the `block_param` parameter - is this potentially ambiguous?

modules/boltlib/lib/puppet/functions/run_task.rb
```diff
@@ -10,45 +10,42 @@
...
     end
-    use_args = task_args.nil? ? {} : task_args
-    task_instance = call_function('new', task_type, use_args)
-    run_task_instance(task_instance, targets)
-  end
 
-  def run_task_instance(task, *targets)
-    unless Puppet[:tasks]
+    # TODO: use the compiler injection once PUP-8237 lands
+    task_signature = Puppet::Pal::ScriptCompiler.new(closure_scope.compiler).task_signature(task_name)
```
mruzicka: I'm surprised this works ...

modules/boltlib/lib/puppet/functions/run_task.rb
```diff
@@ -10,45 +10,38 @@
...
 
-  dispatch :run_named_task do
-    param 'String[1]', :task_type
+  # this is used from 'bolt task run'
+  dispatch :run_task_raw do
+    param 'String[1]', :task_name
     param 'TargetOrTargets', :targets
     optional_param 'Hash[String[1], Any]', :task_args
+    block_param
   end
```
mruzicka: The function has these signatures now:
```ruby
dispatch :run_task do
  param 'String[1]', :task_name
  param 'TargetOrTargets', :targets
  optional_param 'Hash[String[1], Any]', :task_args
end

dispatch :run_task_raw do
  param 'String[1]', :task_name
  param 'TargetOrTargets', :targets
  optional_param 'Hash[String[1], Any]', :task_args
  block_param
end
```
note that they only differ in the `block_param` parameter - is this potentially ambiguous?

modules/boltlib/lib/puppet/functions/run_task.rb
```diff
@@ -10,45 +10,38 @@
...
     end
-    use_args = task_args.nil? ? {} : task_args
-    task_instance = call_function('new', task_type, use_args)
-    run_task_instance(task_instance, targets)
-  end
 
-  def run_task_instance(task, *targets)
-    unless Puppet[:tasks]
+    # TODO: use the compiler injection once PUP-8237 lands
+    task_signature = Puppet::Pal::ScriptCompiler.new(closure_scope.compiler).task_signature(task_name)
```
mruzicka: I'm surprised this works ...

lib/bolt/cli.rb
```diff
@@ -477,48 +483,34 @@ def outputter
...
+      raise Bolt::CLIError, parse_error.message
     end
 
-    def run_plan(plan, args, modulepath)
+    def run_plan(plan, args)
       Dir.mktmpdir('bolt') do |dir|
         cli = []
         Puppet::Settings::REQUIRED_APP_SETTINGS.each do |setting|
           cli << "--#{setting}" << dir
         end
```
mruzicka: Should we perform the same kind of setup (i.e. creating a temporary directory, and setting the REQUIRED_APP_SETTINGS) in the run_task method too? Why is it even necessary?

modules/boltlib/lib/puppet/functions/run_task.rb
```diff
@@ -59,22 +52,26 @@ def run_task_instance(task, *targets)
...
 
       # TODO: separate handling of default since it's platform specific
-      input_method = task._pcore_type['input_method'].value
+      input_method = task.input_method
 
-      arguments = task.task_args
-      Puppet::Pops::Types::ExecutionResult.from_bolt(
-        executor.run_task(executor.from_uris(hosts), task.executable_path, input_method, arguments)
-      )
+      executor.run_task(executor.from_uris(hosts), task.executable, input_method, use_args, &block)
```
thallgren: I might be wrong here but I was under the impression that we eventually wanted the `executor.run_task` to return an `ExecutionResult` so that it didn't need to get wrapped here. If so, that means that the handling of that `ExecutionResult` must be done above this function.

Regardless of that, I think it less ideal to have the function use different return types depending on how it is called. I would like this function to always return an `ExecutionResult`. It's in line with the intention to let a `run_plan` return an aggregated `ExecutionResult`. In order to support this, Bolt must be able to present an `ExecutionResult` to the user (same goes for `run_script`, `run_command`, and `file_upload`.

lib/bolt/cli.rb
```diff
@@ -409,15 +409,7 @@ def execute(options)
...
-              task_name = options[:object]
-
-              path, metadata = load_task_data(task_name, @config[:modulepath])
-              input_method = metadata['input_method']
-
-              input_method ||= 'both'
-              executor.run_task(
-                nodes, path, input_method, options[:task_options]
-              ) do |node, event|
+              execute_task(executor, options) do |node, event|
```
thallgren: If the execute_task returns an `ExecutionResult`, the passing of a block would be unnecessary. Instead, the `ExecutionResult` could be presented to the user in some standardized way (iterating over the result of each node, etc.).

mruzicka: As discussed in HipChat, the block is used to report progress, rather than return the result.

lib/bolt/cli.rb
```diff
@@ -477,48 +483,34 @@ def outputter
...
-          [file, metadata]
         end
-      rescue Puppet::Module::Task::TaskNotFound
-        raise  Bolt::CLIError,
-               "Could not find task '#{name}' in module '#{module_name}'"
-      rescue Puppet::Module::MissingModule
-        # Generate message so we don't expose "bolt environment"
-        raise  Bolt::CLIError, "Could not find module '#{module_name}'"
       end
+      raise Bolt::CLIError, parse_error.message
```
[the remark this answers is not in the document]
mruzicka: Ideally I'd raise an error directly in the block passed to the `with_script_compiler` but unfortunately that method rescues the error and throws a different one instead, so it is not possible to have the desired exception type propagate from the block. The current implementation is IMO the closest to the desired behavior.

modules/boltlib/lib/puppet/functions/run_task.rb
mruzicka: While making the executor use the pcore types for its arguments & return value is the cleanest solution, it unfortunately makes us pay the price of loading puppet_pal even when executing simple scripts.
Are we OK with it, @adreyer?

lib/bolt/cli.rb
```diff
@@ -441,12 +433,26 @@ def execute(options)
       raise e
     end
 
+    def with_bolt_executor(executor, &block)
+      Puppet.override(bolt_executor: executor, &block)
```
adreyer: Can we bind this block to the executor in bolt instead of having puppet aware of it?

lib/bolt/cli.rb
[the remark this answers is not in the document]
adreyer: Is there a ticket to fix this behavior and allow the script compiler to propagate errors.

modules/boltlib/lib/puppet/functions/run_task.rb
adreyer: I don't think we can unless we can move ExecutionResult out of puppet or decrease the performance penalty of loading puppet. All of the run_* functions should return similar values and loading puppet for commands, scripts and files should be as fast as possible.

thallgren: @joshcooper , FYI and off-topic in this PR but here's one reason to rewrite Bolt in Go. There is a performance penalty caused by Puppet load times.

lib/bolt/cli.rb
thallgren: @adreyer , I think it's desirable to be able to use this block from Puppet too. That would enable a run_plan to report progress from the tasks that it run and to have full control over the blocks. We have already discussed moving `Target` to Bolt. The block would be "puppet safe" if called with a `Target` instance and a `Hash[String,Data]` instead of the current Ruby `node` and a hash using symbols.

modules/boltlib/lib/puppet/functions/run_task.rb
joshcooper: @thallgren It seems premature to worry about puppet load times when you're making hundreds of network connections, performing SSH/WinRM+SSL authentication, and executing commands on remote systems. Do you know what percentage of time is taken for compilation and what the speedup would be?

lib/bolt/cli.rb
adreyer: That makes sense. The format is basically what the transport proposal suggests.

modules/boltlib/lib/puppet/functions/run_task.rb
thallgren: @joshcooper , I completely agree with that. But what about the case at hand? Where puppet is loaded just to run a task, a script, or a command?

lib/bolt/cli.rb
mruzicka: Could we move both `Target` & `ExecutionResult` to bolt then? The would allow for a clean implementation without performance penalty.

thallgren: Part of that work is in progress. I'm working on a prerequisite. See [PUP-8195](https://tickets.puppetlabs.com/browse/PUP-8195).

mruzicka: 👍
So can I just create & use plain old ruby classes or do I need to create something more specific?

thallgren:
I think that one of the changes we do when moving `Target` and `ExecutionResult` into Bolt is to transform them into regular Ruby classes. There's nothing stopping Puppet from then mapping them data types (like we do with, for instance the Ruby `URI`). That means that they can be used without Puppet present.

adreyer:
@thallgren agreed that seems like the best path to code reuse without a performance penalty

adreyer:
@thallgren I'm going to merge this in and we can clean up some of the calls with PUP-8190

PR #192, opened 2017-12-12
(BOLT-125) Support verified SSL for WinRM connections

nicklewis:
We now default to HTTPS connections for WinRM unless --insecure is
specified. This requires a CA certificate to be provided, which can be
done via the winrm ca-cert setting in the config file.

The WinRM connection now uses the appropriate default port depending on
whether it's connecting via HTTP or HTTPS (using ports 5985 and 5986
respectively).


nicklewis:
I'm not sure a great way to test this... I don't think we can easily enable SSL in the vagrant environment (the `winrm` command-line isn't available in nano server?), and I'm unclear what machinations we would need to go through in appveyor to generate an SSL cert, enable winrm SSL, and make that cert available to the tests. Even the winrm gem itself only tests using a cert fingerprint (rather than a CA cert) in appveyor.

It may be easier to cover this functionality at the acceptance level.

nicklewis:
I also don't know whether we actually want to default to secure when you're required to set up a config file to point at the CA cert.

nicklewis:
This is the script I used to setup SSL on a host for local testing:

```powershell
$ErrorActionPreference = "Stop"
$system = Get-WmiObject win32_computersystem
$fqdn = $system.DNSHostName.toLower()+".delivery.puppetlabs.net"
$c = New-SelfSignedCertificate -DnsName $fqdn -CertStoreLocation cert:\LocalMachine\My -KeyUsage CertSign
winrm create winrm/config/Listener?Address=*+Transport=HTTPS "@{Hostname=`"$fqdn`";CertificateThumbprint=`"$($c.ThumbPrint)`"}"
$out = [System.Convert]::ToBase64String($c.RawData, "InsertLineBreaks")
echo "-----BEGIN CERTIFICATE-----"
echo $out
echo "-----END CERTIFICATE-----"
```

spec/bolt/node/winrm_spec.rb
```diff
@@ -58,6 +46,18 @@ def stub_winrm_to_raise(klass, message)
       end
     end
 
+    it "raises Node::ConnectError if the node name can't be resolved" do
```
nicklewis: If I run this test first, it fails because it takes > 1 second to fail (11 seconds), implying it hit the timeout. However, the assertions pass, implying that the correct exception is raised.

If I don't run it first, it works just fine. That's cool.

nicklewis: I increased the default timeout to 20 seconds and this still fails after 11. That means it must not be the bolt timeout it's waiting for (a relief, since it's not raising a timeout error...), but some other DNS timeout. One mystery down, but that still leaves open the question of why reordering the test fixes it (as does including a `winrm.connect` at the beginning...).

nicklewis:
It appears the failing test on Ruby 2.0 may be due to a [ruby bug](https://bugs.ruby-lang.org/issues/12604) causing slow resolution for the first (maybe???) DNS request. The bug was surfaced by this change because I removed the `before(:each) { winrm.connect }` call for tests which were trying to create error conditions (and thus ignoring that connect call anyway).

Unfortunately, that means that the very first connection the tests now try to make is for the one test that expects a DNS request to fail in a specific amount of time...

The simplest thing here is probably just to reorder the tests, since I don't think there's much we can do about the actual bug.

Iristyle:
@nicklewis there are PowerShell equivalents of the `winrm` commands. Something like https://github.com/hhazeley/Azure-V2-Powershell-Scripts/blob/master/CustomScriptExtenstions/WinRMHTTPS-FC-Set.ps1 should work for Nano.

nicklewis:
@Iristyle I'll check that out, thanks. The `New-SelfSignedCertificate` command also doesn't seem to be available. We could probably just pregenerate a cert and stick it in the repo though, for the vagrant case. I'll have to see what we can do to get ahold of the cert content in appveyor.

Iristyle:
Yeah, pre-generating a well-known cert should be OK (there may already be something in the cert store on the image as well).  Looks like others ran into the same problem with `New-SelfSignedCertificate` missing - https://github.com/PlagueHO/LabBuilder/issues/86

nicklewis:
I added some certs in `resources/` and updated the Vagrantfile and appveyor to use them, so we now have a test of SSL! This should be good to go now.

lib/bolt/node/winrm.rb
```diff
@@ -9,40 +9,52 @@ def protocol
...
     end
 
-    def initialize(host, port, user, password, shell: :powershell, **kwargs)
-      super(host, port, user, password, **kwargs)
+    def port
+      default_port = @insecure ? 5985 : 5986
+      @port || default_port
+    end
 
-      @shell = shell
```
adreyer: We didn't expose a way to get shell into the config before?

nicklewis: Not as far as I can tell. The `shell` argument to the initializer is a separate arg from the config, not a part of it. Also, we immediately execute powershell code upon making the connection (`$PSVersionTable.PSVersion`), so I don't see how you could reasonably use another shell anyway.

spec/bolt/node/winrm_spec.rb
```diff
@@ -113,519 +114,531 @@ def stub_winrm_to_raise(klass, message)
     end
   end
 
-  it "executes a command on a host", winrm: true do
-    expect(winrm.execute(command).stdout.string).to eq("#{user}\r\n")
+  context "connecting over SSL" do
+    it "executes a command on a host", winrm: true do
```
MikaelSmith: How does file transfer work? I feel like we should also test that over SSL.

lib/bolt/config.rb
```diff
@@ -106,9 +107,15 @@ def update_from_file(data)
         if data['winrm']['connect-timeout']
           self[:transports][:winrm][:connect_timeout] = data['winrm']['connect-timeout']
         end
+        if data['winrm']['insecure']
```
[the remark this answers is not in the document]
adreyer: yeah, I think we need a map of what values go where.

lib/bolt/config.rb
```diff
@@ -20,10 +20,11 @@ module Bolt
       log_destination: STDERR
     }.freeze
 
-    TRANSPORT_OPTIONS = %i[insecure password run_as sudo sudo_password key tty tmpdir user connect_timeout].freeze
+    TRANSPORT_OPTIONS = %i[insecure password run_as sudo sudo_password
+                           key tty tmpdir user connect_timeout ca_cert].freeze
 
     TRANSPORT_DEFAULTS = {
-      connect_timeout: 10,
+      connect_timeout: 20,
```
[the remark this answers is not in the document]
nicklewis: Whoops, I changed that to debug the appveyor failure. It should be changed back.

lib/bolt/config.rb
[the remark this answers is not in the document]
nicklewis: Yeah I think if we have transport classes, they can be responsible for configuring themselves.

spec/bolt/node/winrm_spec.rb
nicklewis: Good call, I'll add a test for that.

nicklewis: Updated.

nicklewis:
Updated to address comments, should be good to go again

spec/bolt/node/winrm_spec.rb
```diff
@@ -11,11 +11,14 @@
 
   let(:host) { ENV['BOLT_WINRM_HOST'] || 'localhost' }
   let(:port) { ENV['BOLT_WINRM_PORT'] || 55985 }
+  let(:ssl_port) { ENV['BOLT_WINRM_SSL_PORT'] || 55986 }
```
Iristyle: I guess this is an automatic mapping (i.e. I don't see the `Vagrantfile` updated to map 5986 to 55986)?

PR #193, opened 2017-12-12
(BOLT-163) improve error message for config connect_timeout

zreichert:


nicklewis:
I'm impressed you managed to get 23 rubocop offenses in a 31 line change.

lib/bolt/config.rb
```diff
@@ -147,6 +147,13 @@ def validate
       unless %w[human json].include? self[:format]
         raise Bolt::CLIError, "Unsupported format: '#{self[:format]}'"
       end
+
+      self[:transports].each_value do |v|
+        timeout_value = v[:connect_timeout]
+        unless timeout_value.to_s =~ /^\d*$/ || timeout_value.nil?
```
nicklewis: If the config file is YAML (with typed data) and the CLI parser checks types, do we need to allow strings of digits here, rather than just enforcing `timeout_value.is_a?(Integer)`?

zreichert: I wasn't aware that the CLI parser was checking types.  If thats the case then yeah, `timeout_value.is_a?(Integer)` seems better.

nicklewis: Yep, [here's an example](https://github.com/puppetlabs/bolt/blob/master/lib/bolt/cli.rb#L151).

zreichert: Updated

lib/bolt/config.rb
```diff
@@ -147,6 +147,13 @@ def validate
       unless %w[human json].include? self[:format]
         raise Bolt::CLIError, "Unsupported format: '#{self[:format]}'"
       end
+
+      self[:transports].each_value do |v|
+        timeout_value = v[:connect_timeout]
+        unless timeout_value.is_a?(Integer) || timeout_value.nil?
+          raise Bolt::CLIError, "Unsupported connect-timeout value #{timeout_value}"
```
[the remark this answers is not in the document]
zreichert: I think my preference is for a better error message.  Since `.inspect` seems very close to what we just changed it from.

MikaelSmith:
Typo in the commit message.

lib/bolt/config.rb
```diff
@@ -147,6 +147,14 @@ def validate
       unless %w[human json].include? self[:format]
         raise Bolt::CLIError, "Unsupported format: '#{self[:format]}'"
       end
+
+      self[:transports].each_value do |v|
+        timeout_value = v[:connect_timeout]
+        unless timeout_value.is_a?(Integer) || timeout_value.nil?
+          error_msg = "connect-timeout value must be an Integer received #{timeout_value}:#{timeout_value.class}"
```
MikaelSmith: I'd like a separator between `Integer` and `received`, either comma or `but`.

zreichert: Fixed

PR #195, opened 2017-12-13
(BOLT-165) Format ExecutionResult returned from a plan

MikaelSmith:
Apply special formatting when an ExecutionResult is the final output from running a plan.

Example output would be something like
```
[
  {
    "node": "node1",
    "status": "finished",
    "result": {
      "_output": "foo\n"
    }
  },
  {
    "node": "node2",
    "status": "failed",
    "result": {
      "_error": "bar\n"
    }
  }
]
```

### DIFF
--- a/lib/bolt/outputter/human.rb
+++ b/lib/bolt/outputter/human.rb
@@ -78,7 +78,17 @@ module Bolt
       end
 
       def print_plan(result)
-        @stream.puts result
+        # If a hash or array, pretty-print as JSON
+        if result.is_a?(Hash) || result.is_a?(Array)
+          if result.empty?
+            # Avoids extra lines for an empty result
+            @stream.puts(result.to_json)
+          else
+            @stream.puts(::JSON.pretty_generate(result))
+          end
+        else
+          @stream.puts result.to_s
+        end
       end
 
       def fatal_error(e); end
--- a/lib/bolt/outputter/json.rb
+++ b/lib/bolt/outputter/json.rb
@@ -44,6 +44,7 @@ module Bolt
       end
 
       def print_plan(result)
+        # Ruby JSON patches most objects to have a to_json method.
         @stream.puts result.to_json
       end
 
--- a/spec/bolt/outputter/json_spec.rb
+++ b/spec/bolt/outputter/json_spec.rb
@@ -33,6 +33,28 @@ describe "Bolt::Outputter::JSON" do
     expect(parsed['items'][1]['status']).to eq('failure')
   end
 
+  it "formats ExecutionResult from a plan" do
+    result = [
+      { 'node' => 'node1', 'status' => 'finished', 'result' => { '_output' => 'yes' } },
+      { 'node' => 'node2', 'status' => 'failed', 'result' =>
+        { '_error' => { 'message' => 'The command failed with exit code 2',
+                        'kind' => 'puppetlabs.tasks/command-error',
+                        'issue_code' => 'COMMAND_ERROR',
+                        'partial_result' => { 'stdout' => 'no', 'stderr' => '', 'exit_code' => 2 },
+                        'details' => { 'exit_code' => 2 } } } }
+    ]
+    outputter.print_plan(result)
+
+    result_hash = JSON.parse(output.string)
+    expect(result_hash).to eq(result)
+  end
+
+  it "prints non-ExecutionResult from a plan" do
+    result = "some data"
+    outputter.print_plan(result)
+    expect(output.string.strip).to eq('"' + result + '"')
+  end
+
   it "handles fatal errors" do
     outputter.print_head
     outputter.print_result(Bolt::Node.from_uri('node1', config: config),
--- a/lib/bolt/cli.rb
+++ b/lib/bolt/cli.rb
@@ -503,6 +503,23 @@ HELP
       raise Bolt::CLIError, parse_error.message
     end
 
+    # Expects to be called with a configured Puppet compiler or error.instance? will fail
+    def unwrap_execution_result(result)
+      if result.instance_of? Puppet::Pops::Types::ExecutionResult
+        error = Puppet::Pops::Types::TypeFactory.error
+        result.iterator.map do |node, output|
+          if error.instance?(output)
+            # Get the original error hash used to initialize the Error type object.
+            { node: node, status: 'failed', result: { '_error' => output._pcore_init_hash } }
+          else
+            { node: node, status: 'finished', result: output }
+          end
+        end
+      else
+        result
+      end
+    end
+
     def run_plan(plan, args)
       Dir.mktmpdir('bolt') do |dir|
         cli = []
@@ -512,7 +529,10 @@ HELP
         Puppet.initialize_settings(cli)
         Puppet::Pal.in_tmp_environment('bolt', modulepath: [BOLTLIB_PATH] + @config[:modulepath], facts: {}) do |pal|
           pal.with_script_compiler do |compiler|
-            compiler.call_function('run_plan', plan, args)
+            result = compiler.call_function('run_plan', plan, args)
+            # Querying ExecutionResult for failures currently requires a script compiler.
+            # Convert from an ExecutionResult to structured output that we can print.
+            unwrap_execution_result(result)
           end
         end
       end
--- a/spec/bolt/cli_spec.rb
+++ b/spec/bolt/cli_spec.rb
@@ -789,7 +789,7 @@ NODES
           cli.config.modulepath = [File.join(__FILE__, '../../fixtures/modules')]
         end
 
-        it "runs a plan given a name" do
+        it "formats results of a passing task" do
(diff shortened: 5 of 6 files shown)


lib/bolt/outputter.rb
```diff
@@ -14,6 +14,18 @@ def self.for_format(format)
     def initialize(stream = $stdout)
       @stream = stream
     end
+
+    def unwrap_execution_result(result)
+      ok_hash = result.ok_nodes.iterator.map do |node, output|
```
MikaelSmith: I originally started with checking whether result is an ExecutionResult. I'd still rather do that, but had trouble getting spec tests to correctly load the class.

lib/bolt/outputter/human.rb
```diff
@@ -78,7 +78,13 @@ def print_summary(results, elapsed_time)
       end
 
       def print_plan(result)
-        @stream.puts result
+        if result_hash = unwrap_execution_result(result)
+          unless result_hash.empty?
```
MikaelSmith: Chose to avoid printing anything - rather than `[\n\n]\n` - for human format.

lib/bolt/outputter/json.rb
```diff
@@ -44,7 +44,11 @@ def print_summary(results, elapsed_time)
       end
 
       def print_plan(result)
-        @stream.puts result.to_json
+        if result_hash = unwrap_execution_result(result)
+          @stream.puts result_hash.to_json
+        else
+          @stream.puts result.to_json
```
adreyer: Won't this blow up for some objects? I think we need to fall back to "\"#{result.to_string}\"" or { "_output": "\"#{result.to_string}\""}

lib/bolt/outputter/human.rb
adreyer: I think there needs to be something here  `[ ]` seems better than that though

lib/bolt/outputter/human.rb
```diff
@@ -78,7 +78,13 @@ def print_summary(results, elapsed_time)
...
+            @stream.puts(::JSON.pretty_generate(result_hash))
+          end
+        else
+          @stream.puts result
```
adreyer: We should attempt to pretty print json here too probably. For example of the plan returns a Hash we converting it to json and pretty printing is probably better than just puts.

spec/bolt/outputter/json_spec.rb
```diff
@@ -33,6 +33,23 @@
     expect(parsed['items'][1]['status']).to eq('failure')
   end
 
+  it "formats ExecutionResult from a plan" do
```
adreyer: I think parsing the JSON results in more robust tests

lib/bolt/outputter.rb
```diff
@@ -14,6 +14,18 @@ def self.for_format(format)
...
+        {node: node, status: 'finished', result: output}
+      end
+      err_hash = result.error_nodes.iterator.map do |node, output|
```
adreyer: Does this map work properly? I thought the result was getting restructured sort of like: 

```
{'key': 'val', '_error': {'msg': ....} } 
to
{'msg': .... , 'value': { 'key': 'val' } } 
```

lib/bolt/outputter/human.rb
MikaelSmith: That implies an empty result?

lib/bolt/outputter/json.rb
MikaelSmith: It's what was originally there. It might blow up depending on what's returned.

MikaelSmith: Amazingly it works with a plain string, as shown by one of the spec tests.

lib/bolt/outputter.rb
MikaelSmith: This approach totally breaks if there are some errored nodes, but it's not even getting this far. ok_nodes seems to error.

lib/bolt/outputter/json.rb
adreyer: Yeah I think it's blowing up now

lib/bolt/outputter/human.rb
adreyer: Isn't it an empty result?

lib/bolt/outputter.rb
MikaelSmith: The way I'm trying to use this seems to be missing some setup. ok_nodes works fine within a plan, but not called here.

lib/bolt/outputter.rb
MikaelSmith: @thallgren calling `ok_nodes` or `error_nodes` - or using `TypeFactory.error.instance?` - all seem to cause an exception:
```
NoMethodError: undefined method `module_for_type' for nil:NilClass
  /Users/michaelsmith/puppetlabs/bolt/vendored/puppet/lib/puppet/pops/types/p_object_type_extension.rb:27:in `create'
  /Users/michaelsmith/puppetlabs/bolt/vendored/puppet/lib/puppet/pops/types/puppet_object.rb:18:in `_pcore_type'
  /Users/michaelsmith/puppetlabs/bolt/vendored/puppet/lib/puppet/pops/types/type_calculator.rb:512:in `infer_Object'
  /Users/michaelsmith/puppetlabs/bolt/vendored/puppet/lib/puppet/pops/visitor.rb:48:in `block in visit_this'
  /Users/michaelsmith/puppetlabs/bolt/vendored/puppet/lib/puppet/pops/visitor.rb:42:in `each'
  /Users/michaelsmith/puppetlabs/bolt/vendored/puppet/lib/puppet/pops/visitor.rb:42:in `visit_this'
  /Users/michaelsmith/puppetlabs/bolt/vendored/puppet/lib/puppet/pops/visitor.rb:82:in `visit_this_0'
  /Users/michaelsmith/puppetlabs/bolt/vendored/puppet/lib/puppet/pops/types/type_calculator.rb:252:in `infer'
  /Users/michaelsmith/puppetlabs/bolt/vendored/puppet/lib/puppet/pops/types/type_calculator.rb:271:in `infer_set'
  /Users/michaelsmith/puppetlabs/bolt/vendored/puppet/lib/puppet/pops/types/iterable.rb:77:in `block in on'
  /Users/michaelsmith/puppetlabs/bolt/vendored/puppet/lib/puppet/pops/types/iterable.rb:77:in `map'
  /Users/michaelsmith/puppetlabs/bolt/vendored/puppet/lib/puppet/pops/types/iterable.rb:77:in `on'
  /Users/michaelsmith/puppetlabs/bolt/vendored/puppet/lib/puppet/pops/types/execution_result.rb:72:in `iterator'
  /Users/michaelsmith/puppetlabs/bolt/lib/bolt/outputter.rb:21:in `unwrap_execution_result'
...
```

They go away if I force `TypeFactory.error.instance?` to get called during the plan run, such as adding a "noop" line like `$result.error_nodes` in the middle (and throwing away the result). Have any suggestions about how I should be checking for failure in ExecutionResult safely?

MikaelSmith: Maybe it needs to happen within a `Puppet.override` block.

lib/bolt/outputter.rb
MikaelSmith: Won't currently unwrap an error, which is a separate `Error` type.

lib/bolt/outputter/json.rb
MikaelSmith: What was it failing on? I don't think it was related to the failing tests on the PR.

lib/bolt/outputter/human.rb
MikaelSmith: Implemented that in the case of a Hash.

MikaelSmith:
Updated, I think issues are addressed.

lib/bolt/outputter.rb
```diff
@@ -14,6 +14,20 @@ def self.for_format(format)
     def initialize(stream = $stdout)
       @stream = stream
     end
+
+    # Expects to be called with a configured Puppet compiler or error.instance? will fail
```
MikaelSmith: This is annoying, but not sure what to do about it, and it'll require work in Puppet to fix. /cc @thallgren

thallgren: @MikaelSmith , please create a ticket for this and assign it to me. It will be easy to fix.

MikaelSmith:
Fixed rubocop.

lib/bolt/outputter.rb
adreyer: Can't we just unwrap this inside run_plan instead of the outputter? It seems cleaner if we're converting from puppety stuff to bolty stuff to do it there instead of where we "output" the value anyway.

MikaelSmith: Ah, I'll do that. I guess we need to define what print_plan expects if not an ExecutionResult.

MikaelSmith: Done. That turned out a lot cleaner.

PR #196, opened 2017-12-15
(BOLT-258) Configure PCP transport

lucywyman:


lib/bolt/node.rb
```diff
@@ -49,6 +49,9 @@ def initialize(host, port = nil, user = nil, password = nil, uri: nil,
       @sudo_password = transport_conf[:sudo_password]
       @run_as = transport_conf[:run_as]
       @tmpdir = transport_conf[:tmpdir]
+      @service_url = transport_conf[:service_url]
```
lucywyman: I assume this will resolve to nil if this isn't in the config?

lucywyman:
I was also thinking it'd be appropriate to add a pcp integration test, in addition to the unit tests? I'm mostly not terribly confident my changes are doing everything I expect them to :P

adreyer:
@lucywyman I think it will be hard to do real integration tests for PCP without an oss orhcestrator. The current tests don't actually use orch they call the bolt task directly.

lib/bolt/config.rb
```diff
@@ -117,6 +119,21 @@ def update_from_file(data)
...
+        if data['pcp']['service-url']
+          self[:transports][:pcp][:service_url] = data['pcp']['service-url']
+        end
+        if data['pcp']['ca-cert']
+          self[:transports][:pcp][:ca_cert] = data['pcp']['ca-cert']
+        end
+        if data['pcp']['token-file']
+          self[:transports][:pcp][:token_file] = data['pcp']['token-file']
+        end
+        if data['pcp']['environment']
```
adreyer: This should be task-environment I think. Sorry if I didn't update the ticket

adreyer: Actually we should probably refer to it even more specifically in the code.
```
pcp:
  task-environment: my_environment
```
is good in the config file but if it's going to be visible to all the transports it should probably be 'orch_task_environment' as a variable.

lib/bolt/node.rb
adreyer: it should

lib/bolt/node/orch.rb
```diff
@@ -41,10 +41,20 @@ def task_name_from_path(path)
 
     def _run_task(task, _input_method, arguments)
       body = { task: task_name_from_path(task),
+               environment: @environment,
                params: arguments,
                scope: {
                  nodes: [@host]
                } }
+      if @service_url
+        body.merge(service_url: @service_url)
```
adreyer: `body[:service_url] = @service_url if @service_url`
is more idiomatic

spec/bolt/cli_spec.rb
```diff
@@ -911,6 +911,7 @@ def stub_config(file_content = nil, base = nil)
 
   describe 'configfile' do
     let(:configdir) { File.join(__dir__, '..', 'fixtures', 'configs') }
+    # TODO: why isn't this just a fixture?
```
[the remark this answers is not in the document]
adreyer: it can be

PR #197, opened 2017-12-15
(maint) Remove PSH 2.5.1 pin

Iristyle:
- puppetlabs_spec_helper 2.6.0 release introduced a new bug that
 requires that a .fixtures.yml exists with a fixtures: section in
 it. A temporary pin to 2.5.1 was made until the issue could be
 fixed, which it has now with a 2.6.1 gem release


Iristyle:
GitHub wasn't letting me reopen #189 for some reason with this updated... so, new PR

PR #198, opened 2017-12-15
(BOLT-155) List available tasks

MikaelSmith:
Lists available tasks, but does not yet include their descriptions.


MikaelSmith:
Also, the CLI parsing is starting to get unwieldy.

MikaelSmith:
Working on adding tests.

lib/bolt/cli.rb
```diff
@@ -483,6 +494,19 @@ def outputter
       @outputter ||= Bolt::Outputter.for_format(@config[:format])
     end
 
+    def list_tasks
+      Puppet.initialize_settings
+      Puppet::Pal.in_tmp_environment('bolt', modulepath: [BOLTLIB_PATH] + @config[:modulepath], facts: {}) do |pal|
```
adreyer: Can we make an in_bolt_compiler function so we don't forget to update this.

MikaelSmith: What do you mean forget to update this?

lib/bolt/cli.rb
```diff
@@ -86,8 +88,11 @@ class CLI
 Available options are:
 HELP
 
-    MODES = %w[command script task plan file].freeze
-    ACTIONS = %w[run upload download].freeze
+    COMMANDS = { 'command' => %w[run],
```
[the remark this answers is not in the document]
MikaelSmith: What do we call objects? The combination of MODES and ACTIONS?

lib/bolt/cli.rb
MikaelSmith: Added `in_bolt_compiler`.

lib/bolt/cli.rb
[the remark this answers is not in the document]
adreyer: No the "MODE" is called an OBJECT at least internally since it related to what the action is taken on

MikaelSmith:
I filed https://tickets.puppetlabs.com/browse/PUP-8259 to deal with better errors. We could augment that to provide a way to perform better error handling.

lib/bolt/cli.rb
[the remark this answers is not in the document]
MikaelSmith: object is the thing we act on, as in `<mode> <action> <object>`.

lib/bolt/cli.rb
```diff
@@ -483,21 +494,39 @@ def outputter
...
+    def list_tasks
+      in_bolt_compiler do |compiler|
+        tasks = compiler.list_tasks
+        tasks.map(&:name).sort.map do |task_name|
+          task_sig = compiler.task_signature(task_name)
+          [task_name, task_sig.task.description]
+        end
+      end
+    rescue Puppet::Error
+      raise Bolt::CLIError, "Failure while reading task metadata"
```
MikaelSmith: Updated to catch Puppet exceptions so we don't print them twice (and don't print a stack trace).

MikaelSmith:
I'd like to add an invalid task and test behavior with it present (in a separate fixture modulepath).

MikaelSmith:
Added a test with an invalid task.

lib/bolt/cli.rb
```diff
@@ -483,21 +494,39 @@ def outputter
...
+    end
+
+    def run_task(name, nodes, args, &block)
+      parse_error = nil
+      in_bolt_compiler do |compiler|
+        begin
+          return compiler.call_function('run_task', name, nodes, args, &block)
```
adreyer: Are we intentionally returning to prevent some PAL code from executing?

MikaelSmith: No, I just had C++ scoping in mind when I wrote this. Pushed a commit to fix it to return the result more cleanly.

PR #199, opened 2017-12-18
(BOLT-258) Configure pcp transport

lucywyman:
This is #196 re-opened. I messed up branch names, and cannot seem to fix deleting a branch

### DIFF
--- a/spec/bolt/cli_spec.rb
+++ b/spec/bolt/cli_spec.rb
@@ -922,6 +922,12 @@ NODES
         },
         'winrm' => {
           'connect-timeout' => 7
+        },
+        'pcp' => {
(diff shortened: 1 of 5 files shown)


spec/bolt/cli_spec.rb
```diff
@@ -911,6 +911,7 @@ def stub_config(file_content = nil, base = nil)
 
   describe 'configfile' do
     let(:configdir) { File.join(__dir__, '..', 'fixtures', 'configs') }
+    # TODO: why isn't this just a fixture?
```
adreyer: This can probably be removed or fixed

lucywyman: Fixed!

PR #200, opened 2017-12-18
(BOLT-262) Report skipped nodes as such

mruzicka:
Make sure the error reported when a task target node is skipped indicates the fact.

### DIFF
--- a/lib/bolt/node/orch.rb
+++ b/lib/bolt/node/orch.rb
@@ -58,6 +58,18 @@ module Bolt
 
       if state == 'finished'
         exit_code = 0
+      elsif state == 'skipped'
+        return Bolt::TaskResult.new(
+          JSON.dump(
+            '_error' => {
+              'kind' => 'puppetlabs.tasks/skipped-node',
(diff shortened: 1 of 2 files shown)


lib/bolt/node/orch.rb
```diff
@@ -53,6 +53,8 @@ def _run_task(task, _input_method, arguments)
 
       if state == 'finished'
         exit_code = 0
+      elsif state == 'skipped'
+        return Bolt::Result.new('Node was skipped')
```
adreyer: This should probably be a TaskResult with a skipped error. I have an example error here https://github.com/puppetlabs/task-modules/blob/master/site/canary/lib/puppet/functions/canary/skip.rb

mruzicka: ok

mruzicka:
@adreyer updated, while at it I've cleaned up the Bolt::Orch class spec tests so that it's clearer which values are actually used in the tests - those defined by the `let` blocks vs. those assigned to the instance variables.

PR #201, opened 2017-12-19
(BOLT-119) Support noop

MikaelSmith:


MikaelSmith:
I still want to add some more tests, but this seems to be working.

modules/boltlib/lib/puppet/functions/run_task.rb
```diff
@@ -58,6 +58,14 @@ def run_task_raw(task_name, targets, task_args = nil, &block)
     end || (raise Puppet::ParseError, 'Task parameters did not match')
     task = task_signature.task
 
+    if executor.noop
+      if task.supports_noop
```
adreyer: I think logic will have to move to bolt since the vague discussions around bolt run plan --noop is that it would just skip any tasks that don't support_noop rather than error. This is fine until we actually elaborate that though.

MikaelSmith: Yeah, I'm avoiding writing too many spec tests around the fine details of executor and run_task handling of noop. Sticking to macro tests.

MikaelSmith: Added integration tests and rebased.

MikaelSmith:
This will need an update after BOLT-155 is merged.

MikaelSmith:
Tests passed, ready to go.

PR #202, opened 2017-12-19
(BOLT-267) Add ExecutionResult and Target data types to Bolt

thallgren:
This commit adds the `ExecutionResult` and `Target` data types to `Bolt`
and makes them available to Puppet using the new
`Puppet::DataTypes#create_type` introduced by PUP-8195.

Note! The data types were previously present in Puppet core and this
commit must be synchronized with an upgrade to use a puppet version
where PUP-8265 (covering the removal) has been merged.

### DIFF
--- a/lib/bolt/cli.rb
+++ b/lib/bolt/cli.rb
@@ -401,6 +401,10 @@ HELP
                              end
       end
 
+      # ExecutionResult loaded here so that it can get puppet features if
+      # puppet is present
+      require 'bolt/execution_result'
+
       if options[:action] == 'show'
         if options[:mode] == 'task'
           if options[:object]
@@ -574,10 +578,9 @@ HELP
 
     # Expects to be called with a configured Puppet compiler or error.instance? will fail
     def unwrap_execution_result(result)
-      if result.instance_of? Puppet::Pops::Types::ExecutionResult
-        error = Puppet::Pops::Types::TypeFactory.error
+      if result.instance_of? Bolt::ExecutionResult
         result.iterator.map do |node, output|
-          if error.instance?(output)
+          if output.is_a?(Puppet::DataTypes::Error)
             # Get the original error hash used to initialize the Error type object.
             { node: node, status: 'failed', result: { '_error' => output._pcore_init_hash } }
           else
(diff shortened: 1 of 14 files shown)


ripienaar:
May I ask why this is being moved here? Execution result is generally useful outside of bolt for other integrations into the plan DSL. A bolt dependency to get plans working seems undesirable (if I understand the implication right)

thallgren:
The `run_plan` function has alredy moved into Bolt and with it, the plan DSL. See https://tickets.puppetlabs.com/browse/PUP-8073.

ripienaar:
Ah bummer. And bolt is still censoring the entire puppet or is this towards being installable in AIO ?

ripienaar:
Vendoring. Sorry can’t edit on mobile github

joshcooper:
@ripienaar all of the puppet feature branch changes have been merged back into puppet#master, and that will be released as puppet 5.4.0 in January-ish. So bolt will be able to stop vendoring puppet "real soon now".

ripienaar:
@joshcooper yes was really looking forward to 5.4 so I can realesse my plan integrations as well as my task runner. 

Looks like I can finish task runner but with the plan function and results type gone that’s a dead stop to anything that want to build on the plan DSL - arguably the most interesting aspect of the new features in Puppet. 

Anyway there also seem some disagreement between this and what Alex said on slack either that or it’s a case of can but won’t for a few months more :(

adreyer:
@joshcooper moving off of the feature branch is a necessary but not sufficient condition for undvendoring puppet. Bolt is still moving fast and will continue to require changes in puppet. We may not want to be blocked on puppet's release cycle. Users have expressed concern about installing the latest puppet vendoring removes this concern. I don't think we're quite done with vendoring.

MikaelSmith:
This needs a rebase.

thallgren:
Rebased.

MikaelSmith:
Looks like some rubocop work needed, as well as changes to how we reference ExecutionResult from the modules/boltlib spec tests.

MikaelSmith:
Apparently we've deleted `feature_puppet-tasks`, so `.gitmodules` will need an update.

Also, whatever code vendored/puppet points to isn't on puppetlabs/puppet, so I can't update to it.

MikaelSmith:
I'm actually confused how specs work but I'm not able to get submodules setup.

PR #203, opened 2017-12-19
(BOLT-157) show docs for tasks

lucywyman:
This depends on merging #198 first.

Built gem with these changes is installed to hf3a1dxx7lajzf0.delivery.puppetlabs.net, if you'd like to ssh in and verify it works (at least until EOD today, 12/19/2017).

### DIFF
--- a/lib/bolt/cli.rb
+++ b/lib/bolt/cli.rb
@@ -31,6 +31,7 @@ Available subcommands:
     bolt command run <command>       Run a command remotely
     bolt script run <script>         Upload a local script and run it remotely
     bolt task show                   Show list of available tasks
+    bolt task show <task>            Show documentation for task
     bolt task run <task> [params]    Run a Puppet task
     bolt plan show                   Show list of available plans
     bolt plan run <plan> [params]    Run a Puppet task plan
@@ -402,7 +403,14 @@ HELP
 
       if options[:action] == 'show'
         if options[:mode] == 'task'
-          outputter.print_table(list_tasks)
+          if options[:object]
+            outputter.print_task_info(get_task_info(options[:object]))
+          else
+            outputter.print_table(list_tasks)
+            outputter.print_message("Use `bolt task show <task-name>` to view
+                                  details and parameters for a specific
+                                  task.")
+          end
         elsif options[:mode] == 'plan'
           outputter.print_table(list_plans)
         end
@@ -536,6 +544,14 @@ HELP
       end
     rescue Puppet::Error
       raise Bolt::CLIError, "Failure while reading plans"
+    end
+
+    def get_task_info(task_name)
+      in_bolt_compiler do |compiler|
+        compiler.task_signature(task_name).task_hash
+      end
+    rescue Puppet::Error
+      raise Bolt::CLIError, "Failure while reading task metadata"
     end
 
     def run_task(name, nodes, args, &block)
--- a/lib/bolt/outputter/human.rb
+++ b/lib/bolt/outputter/human.rb
@@ -93,6 +93,33 @@ module Bolt
         )
       end
 
+      # @param [Hash] A hash representing the task
+      def print_task_info(task)
+        # Building lots of strings...
+        pretty_params = ""
+        task_info = ""
+        usage = "bolt task run --nodes, -n <node-name> #{task['name']}"
+
+        task['parameters'].each do |k, v|
+          pretty_params << "- #{k}: #{v['type']}"
+          pretty_params << "\n    #{v['description']}\n" if v['description']
+          usage << if !v['type'].to_s.include? "Optional"
+                     " #{k}=<value>"
+                   else
+                     " [#{k}=<value>]"
+                   end
+        end
+
+        usage << " [--noop]" if task['supports_noop']
+
+        task_info << task['name']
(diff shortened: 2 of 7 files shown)


lib/bolt/outputter/human.rb
```diff
@@ -93,6 +93,24 @@ def print_table(results)
...
+        task['parameters'].each do |k, v|
+          pretty_params << "- #{k}: #{v['type']}\n\t#{v['description']}\n"
+        end
+        task_info = <<-INFO
+#{task['name']} - #{task['description']}
+
+USAGE:
+#{task['usage']}
+
+PARAMETERS:
```
MikaelSmith: Should we avoid PARAMETERS if none are defined?

lib/bolt/outputter/human.rb
```diff
@@ -93,6 +93,21 @@ def print_table(results)
         )
       end
 
+      # @param [Hash] A hash representing the task
+      def print_task(task)
+        pretty_params = ""
+        task_info = ""
+        task['parameters'].each do |k, v|
+          pretty_params << "- #{k}: #{v['type']}\n\t#{v['description']}\n"
```
lucywyman: This feels gross but also works and may be the best way to build the string?

MikaelSmith: It seems fine. You could use heredoc, but I think that'd be less clear if we actually want tabs.

spec/bolt/outputter/human_spec.rb
```diff
@@ -40,6 +40,32 @@
...
+    outputter.print_task(task)
+    expect(output.string).to eq(<<-TASK_OUTPUT)
+cinnamon roll - A delicious sweet bun
+
+USAGE:
+Nourishment and happiness
+
+PARAMETERS:
+- icing: Cream cheese
+\tRich, tangy, sweet
```
lucywyman: Also gross. `\t` just seemed like the best way to make sure editors don't accidentally break this test (as mine is prone to do with tabs). But definitely tell me if it's unacceptable.

lib/bolt/cli.rb
```diff
@@ -515,6 +523,37 @@ def list_tasks
       raise Bolt::CLIError, "Failure while reading task metadata"
     end
 
+    def get_task_info(task_name)
+      task_info = {}
+      # TODO: should this formatting be done somewhere else? I think it makes
```
MikaelSmith: Seems ok.

adreyer: Long term maybe in puppet strings. I think it makes sense to generate the usage line in the human outputter and not display it with json output for now.

lib/bolt/outputter/human.rb
```diff
@@ -93,6 +93,21 @@ def print_table(results)
         )
       end
 
+      # @param [Hash] A hash representing the task
+      def print_task(task)
```
adreyer: I think its less than ideal that `print_task` prints the metadata while `print_plan` prints the results of a plan.
it should either be` print_task_info` or `print_plan_results` probably

lib/bolt/outputter/human.rb
adreyer: It will print an extra \n\t when there is no description which is probably wrong.

lib/bolt/cli.rb
```diff
@@ -390,7 +391,14 @@ def execute(options)
       end
 
       if options[:mode] == 'task' && options[:action] == 'show'
-        outputter.print_table(list_tasks)
+        if options[:object]
+          outputter.print_task(get_task_info(options[:object]))
+        else
+          outputter.print_table(list_tasks)
+          outputter.print_message("Use `bolt task show <task-name>` to view
```
adreyer: this `bolt task show <task>` elsewhere it should probably consistent

spec/bolt/outputter/human_spec.rb
adreyer: why not just use spaces?

lib/bolt/cli.rb
lucywyman: Not sure what you mean? Where else should I put the string?

MikaelSmith: Updated the comment for clarity.

lib/bolt/cli.rb
lucywyman: Should the json output just display hash returned from `compiler.task_signature(task_name).task_hash` ? Or should I edit that at all?

spec/bolt/cli_spec.rb
```diff
@@ -582,6 +582,45 @@ def stub_config(file_content = nil, base = nil)
...
+          options = {
+            mode: 'task',
+            action: 'show',
+            object: task_name
+          }
+          cli.execute(options)
+          expect(JSON.parse(@output.string)).to eq(
+            "name" => "sample::params",
+            "description" => "Task with parameters",
+            "executable" => "#{Dir.pwd}/spec/fixtures/modules/sample/tasks/params.sh",
```
lucywyman: This seems fragile. Is it better to remove the executable kv pair from the object here, remove it when we get the task info from the compiler, or to just leave it as is?

adreyer: Lets just not test that part of the output. I don't know how important showing the executable path is.

PR #205, opened 2017-12-19
(BOLT-270) Update ca-cert to cacert

MikaelSmith:
Switch PCP and WinRM transport config options from `ca-cert` to `cacert`
to be shorter and consistent with CURL and other Puppet tools.

### DIFF
--- a/lib/bolt/config.rb
+++ b/lib/bolt/config.rb
@@ -21,7 +21,7 @@ module Bolt
     }.freeze
 
     TRANSPORT_OPTIONS = %i[insecure password run_as sudo sudo_password
-                           key tty tmpdir user connect_timeout ca_cert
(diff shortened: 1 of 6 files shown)


lib/bolt/config.rb
```diff
@@ -115,17 +115,17 @@ def update_from_file(data)
         if data['winrm']['tmpdir']
           self[:transports][:winrm][:tmpdir] = data['winrm']['tmpdir']
         end
-        if data['winrm']['ca-cert']
-          self[:transports][:winrm][:ca_cert] = data['winrm']['ca-cert']
+        if data['winrm']['cacert']
+          self[:transports][:winrm][:ca_cert] = data['winrm']['cacert']
```
adreyer: can we change this to cacert instead of ca_cert in the config object too. ditto for orch

PR #208, opened 2017-12-20
(BOLT-157) Handle tasks that don't have parameters

lucywyman:


lib/bolt/outputter/human.rb
```diff
@@ -100,14 +100,16 @@ def print_task_info(task)
         task_info = ""
         usage = "bolt task run --nodes, -n <node-name> #{task['name']}"
 
-        task['parameters'].each do |k, v|
```
MikaelSmith: One more nit: the spec for BOLT-157 actually shows an empty line between running the command and printing the name (+description). I was overlooking the name while looking at the output, so I think that'd be good to add.

PR #209, opened 2017-12-20
(BOLT-157) Better messaging when task isn't found

lucywyman:


lib/bolt/cli.rb
```diff
@@ -549,6 +549,9 @@ def list_plans
     def get_task_info(task_name)
       in_bolt_compiler do |compiler|
         task = compiler.task_signature(task_name)
+        if task.nil?
+          raise "Could not find task #{task_name} in your modulepath"
```
MikaelSmith: This approach results in Puppet logging the error, because Puppet catches the exception and wraps it in a `Puppet::Error`.

I'd opt for
```
task = in_bolt_compiler do |compiler|
  compiler.task_signature(task_name)
end

raise Bolt::CLIError, "Could not find task #{task_name} in your modulepath" if task.nil?
task.task_hash
```

The difference is your change produces
```
Error: Could not find task foo in your modulepath on node localhost
Failure while reading task metadata
```
what I'm suggesting produces
```
Could not find task foo in your modulepath
```

MikaelSmith: It'd be nice to have a test around this too.

lucywyman: Can do! I wasn't entirely sure what we wanted the message to be, or if we still needed the 'Failure while reading task metadata' message. Thanks!

MikaelSmith: The code you just pushed doesn't really address my comment.

PR #210, opened 2017-12-20
(BOLT-274) Strip init name from bolt and task show

adreyer:


MikaelSmith:
Supposed to be BOLT-274?

lib/bolt/cli.rb
```diff
@@ -526,10 +526,19 @@ def in_bolt_compiler(opts = [])
       end
     end
 
+    # TODO: remove this when the PAL functions return the correct names
+    def fix_typed_name(typed_name)
+      name_parts = typed_name.name_parts
+      if name_parts[-1] == 'init'
```
MikaelSmith: What if the module is named 'init'?

MikaelSmith: I guess for now, we'll still get `init::init` and strip only the last one. So that's probably fine.

adreyer: I would not be surprised if there were bugs elsewhere though.

PR #211, opened 2017-12-20
(MAINT) remove bundle exec from commands inside Rakefile

zreichert:


acceptance/Rakefile
```diff
@@ -20,9 +20,10 @@ rototiller_task :host_config do |task|
     n_new << node
   end
   nodes_final = n_new.join('-')
-  generate = "bundle exec beaker-hostgenerator"
+  generate = "beaker-hostgenerator"
   generate += " #{nodes_final}"
   generate += " > hosts.yaml"
+  puts nodes_final
```
lucywyman: Spare puts?

zreichert: Crap

lucywyman:
I'm not familiar enough with bundlr to know: I'm assuming `bundle exec rake` is what makes it so we don't need to have `bundle exec` in the rakefile? Does it propogate?

PR #212, opened 2017-12-20
(BOLT-264) add run-as to ssh config

lucywyman:
Only note: if I do pass `--sudo`, it still asks for a password. It doesn't seem to actually do anything, but the prompt is misleading and annoying. It seems to be coming from https://github.com/puppetlabs/bolt/blob/master/lib/bolt/cli.rb#L197

```
$ bb task run exec command=whoami -n fgyhjd78bkyzjtv.delivery.puppetlabs.net -u lucy -p puppetlabs --sudo --tty
Please enter your privilege escalation password: 
Started on fgyhjd78bkyzjtv.delivery.puppetlabs.net...
Finished on fgyhjd78bkyzjtv.delivery.puppetlabs.net:
  lucy
  {
  }
Ran on 1 node in 2.32 seconds
```

### DIFF
--- a/lib/bolt/cli.rb
+++ b/lib/bolt/cli.rb
@@ -109,6 +109,7 @@ HELP
       }
       @config = Bolt::Config.new
       @parser = create_option_parser(@options)
+      @logger = Logger.new(STDERR)
     end
 
     def create_option_parser(results)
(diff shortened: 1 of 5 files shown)


adreyer:
If you pass `--run-as --sudo "password"` does it behave the same as `--run-as --sudo-password "password"`

lucywyman:
@adreyer Yes, `--sudo` has the same behavior as `--sudo-password`.

PR #215, opened 2017-12-21
(PUP-8271) bump puppet for init name fix

adreyer:


lucywyman:
```
lucy@ophelia:~/githubs/bolt ((c6a917a...))$ bb task show
apache                                 Allows you to perform apache service functions
echo                                   Foo bar
exec                                   Executes a command on the system
facter
```

Looks good to me!

PR #216, opened 2017-12-21
(maint) fix schema of Error object in plan results

adreyer:
I found this bug when I was working to pull in new puppet code yesterday


lucywyman:
LGTM

PR #217, opened 2017-12-22
(BOLT-264) Don't show warning when run-as is configured in file

lucywyman:
I realized for the warning message around having `--sudo-password` without `--run-as` I left out the case where run-as is configured in a file. This fixes that issue, and gets rid of warnings in the spec tests.

### DIFF
--- a/lib/bolt/config.rb
+++ b/lib/bolt/config.rb
@@ -154,7 +154,7 @@ module Bolt
     end
 
     def update_from_cli(options)
-      %i[concurrency transport format modulepath ssh['run_as']].each do |key|
+      %i[concurrency transport format modulepath].each do |key|
         self[key] = options[key] if options[key]
       end
 
@@ -169,6 +169,12 @@ module Bolt
         TRANSPORTS.each do |transport|
           self[:transports][transport][key] = options[key] if options[key]
         end
+      end
+
+      if options[:sudo_password] && self[:transports][:ssh][:run_as].nil?
+        logger = Logger.new(self[:log_destination])
+        logger.warn("'--sudo-password will not be used without specifying a" \
+                    "user to escalate to with --run-as")
       end
     end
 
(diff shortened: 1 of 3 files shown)


lib/bolt/config.rb
```diff
@@ -37,6 +37,7 @@ module Bolt
     def initialize(**kwargs)
       super()
       DEFAULTS.merge(kwargs).each { |k, v| self[k] = v }
+      @logger = Logger.new(self[:log_destination])
```
adreyer: I think this will always use the default log level

lucywyman: What log level should it use?

adreyer: We should wait to create the logger until we have read the log level from the CLI. If we need the logger before that we'll have to move logic around or recreate it.

lucywyman: Should I just comment those out with a note to hold off until we implement that?  Is there a better way to warn users in the meantime?

adreyer: I would add a update_logger method to this class then call it with the log level here before we start processing other options:
https://github.com/puppetlabs/bolt/blob/master/lib/bolt/cli.rb#L254

lucywyman: Ok, I think I did it correctly?

lib/bolt/config.rb
```diff
@@ -170,6 +170,13 @@ def update_from_cli(options)
           self[:transports][transport][key] = options[key] if options[key]
         end
       end
+
+      if options[:sudo_password] && self[:transports][:ssh][:run_as].nil?
+        logger = Logger.new(self[:log_destination])
+        logger.warn("--sudo-password only provides privilege escalation "\
```
adreyer: this message isn't super clear to me.
something like:
'--sudo-password will not be used without specifying a user to escalate to with --run-as'

lucywyman: Fixed

PR #218, opened 2017-12-27
(BOLT-261) Error when invalid task name is in plan

lucywyman:


lib/bolt/cli.rb
```diff
@@ -326,6 +326,16 @@ def read_arg_file(file)
       raise Bolt::CLIError, "Error attempting to read #{file}: #{err}"
     end
 
+    # This regex should be in one place so we don't have to keep track of it
+    def validate_task_name(name)
```
lucywyman: This doesn't seem like what Alex had in mind for regexp validation, but I wanted to use what I know to figure out where to validate first, and keep the regexp in one place.

spec/bolt/cli_spec.rb
```diff
@@ -974,6 +975,19 @@ def stub_config(config, file_content = nil)
...
+          options = {
+            nodes: node_names,
+            mode: 'plan',
+            action: 'run',
+            object: 'sample::upper_case',
+            task_options: { 'nodes' => nodes.join(',') }
+          }
+
+          expect { cli.execute(options) }.to raise_error "Exiting because of an error in Puppet code"
+          puts @output.string
```
lucywyman: Will remove once this is working.

spec/bolt/cli_spec.rb
```diff
@@ -974,6 +975,19 @@ def stub_config(config, file_content = nil)
...
+        it "plan with upper case task name errors" do
+          options = {
+            nodes: node_names,
+            mode: 'plan',
+            action: 'run',
+            object: 'sample::upper_case',
+            task_options: { 'nodes' => nodes.join(',') }
+          }
+
+          expect { cli.execute(options) }.to raise_error "Exiting because of an error in Puppet code"
```
lucywyman: I get the following here:
```
............................................Error: Sample::Echo is not a valid task name  at /home/lucy/githubs/bolt/spec/fixtures/modules/sample/plans/upper_case.pp:3 on node ophelia.delivery.puppetlabs.net
{"kind":"bolt/cli-error","msg":"Exiting because of an error in Puppet code","details":{}}
..............................................................................................................................

Finished in 4.31 seconds (files took 0.46992 seconds to load)
205 examples, 0 failures
```

How do I catch that error, and not the 'Exiting because of an error' line? Should I be testing that in https://github.com/puppetlabs/puppet/pull/6474 ?

MikaelSmith:
This is blocked on INVALID_TASK_NAME being added to Puppet, correct? i.e. https://github.com/puppetlabs/puppet/pull/6475

lucywyman:
Yep! Apologies, I think I left a comment on the ticket but not the PR :P

PR #219, opened 2017-12-29
(BOLT-233) Pass params over stdin if not using sudo

lucywyman:


lib/bolt/node/ssh.rb
```diff
@@ -290,18 +290,32 @@ def _run_task(task, input_method, arguments)
           "PT_#{env}='#{val}'"
         end.join(' ')
       end
+     
+      use_sudo = @sudo || @run_as
```
adreyer: Why do we still have @sudo? It shouldn't be set anymore? If it is accomplishing something we should probably move this logic to a function.

lib/bolt/node/ssh.rb
```diff
@@ -290,18 +290,32 @@ def _run_task(task, input_method, arguments)
...
-        command = if export_args.empty?
-                    "'#{remote_path}'"
-                  else
-                    "#{export_args} '#{remote_path}'"
-                  end
-        output = execute(command, sudoable: true)
-        Bolt::TaskResult.from_output(output)
+      if use_sudo
+        with_remote_task(task, stdin) do |remote_path|
+          command = if export_args.empty?
```
adreyer: If we start command outside of this block this might be a bit cleaner

```
command = export_args.empty? ? '' :  "#{export_args} "
if use_sudo
  <snip>
  command += remote_path
else
  <snip>
   command += remote_path
  <snip>
```

lib/bolt/node/ssh.rb
```diff
@@ -290,18 +290,32 @@ def _run_task(task, input_method, arguments)
...
+          Bolt::TaskResult.from_output(output)
+        end
+      else
+        with_remote_file(task) do |remote_path|
+          command = if export_args.empty?
+                      "'#{remote_path}'"
+                    else
+                      "#{export_args} '#{remote_path}'"
+                    end
+          output = execute(command, stdin: stdin, sudoable: true)
```
adreyer: this isn't sudoable right?

lib/bolt/node/ssh.rb
adreyer: I don't think this is super important but if you're unhappy with the duplication this minimizes it to one line.

lib/bolt/node/ssh.rb
```diff
@@ -290,18 +290,32 @@ def _run_task(task, input_method, arguments)
...
+          Bolt::TaskResult.from_output(output)
+        end
       end
-    # TODO: We should be able to rely on the excutor for this but it will mean
-    # a test refactor
+      # TODO: We should be able to rely on the excutor for this but it will mean
+      # a test refactor
```
adreyer: This indentation looks wrong The comment belongs on the rescue

lib/bolt/node/ssh.rb
lucywyman: Ack, you're right.

lib/bolt/node/ssh.rb
```diff
@@ -291,15 +291,21 @@ def _run_task(task, input_method, arguments)
...
+      command = export_args.empty? ? '' : export_args.to_s
+
+      if @run_as
+        with_remote_task(task, stdin) do |remote_path|
+          output = execute(command, sudoable: true)
+          command += remote_path
+        end
+      else
+        with_remote_file(task) do |remote_path|
+          output = execute(command, stdin: stdin, sudoable: true)
```
adreyer: This command isn't sudoable since it includes stdin

adreyer:
you can add a test to make sure task_wrapper isn't called if it's not sudoing

PR #220, opened 2017-12-29
(BOLT-252) change raketask to work with LAYOUT ENV

zreichert:


acceptance/Rakefile
```diff
@@ -2,29 +2,19 @@ require 'rototiller'
...
-    roles << 'bolt' if i.zero?
-    roles << if /win/ =~ node
-               'winrm'
-             else
-               'ssh'
-             end
-    node = node + roles.join(',') + '.'
-    n_new << node
+
+  task.add_env(name: 'LAYOUT', message: <<-MESSAGE)
```
adreyer: can we get a default or some examples?

adreyer:
You can add a test that makes sure the task_wrapper isn't called if it's not going to sudo

acceptance/Rakefile
zreichert: Sure

MikaelSmith:
It looks like the README should've been updated.

PR #221, opened 2018-01-02
(BOLT-283) Assert that run_task arguments are Data compatible

thallgren:
This commit adds add check that the user defined argument hash passed to
the `run_task` function is an instance of the `Data` data type.


thallgren:
The added unit test in boltlib will not run unless vendored/puppet includes the fix for PUP-8283.

thallgren:
@adreyer you can unblock by doing review and merge of [PUP-8283](/puppetlabs/puppet/pull/6479)

MikaelSmith:
PUP change merged, need to update vendored/puppet to use it.

MikaelSmith:
Updated in https://github.com/puppetlabs/bolt/pull/227.

PR #224, opened 2018-01-02
(BOLT-266) update how run_* actions are logged

lucywyman:

### DIFF
--- a/lib/bolt/cli.rb
+++ b/lib/bolt/cli.rb
@@ -1,7 +1,7 @@
 require 'uri'
 require 'optparse'
 require 'benchmark'
-require 'logger'
+require 'bolt/logger'
 require 'json'
 require 'bolt/node'
 require 'bolt/version'
@@ -109,7 +109,8 @@ HELP
       }
       @config = Bolt::Config.new
       @parser = create_option_parser(@options)
-      @logger = Logger.new(STDERR)
+      @logger = Logger.instance(STDERR)
+      @logger.level = Logger::NOTICE
     end
 
     def create_option_parser(results)
@@ -425,7 +426,11 @@ HELP
         nodes = executor.from_uris(options[:nodes])
 
         results = nil
+        mode = options[:mode]
+        action = options[:action]
+        object = options[:object]
         outputter.print_head
+        @logger.info("Starting bolt #{mode} #{action} #{object}")
 
         elapsed_time = Benchmark.realtime do
           results =
--- a/lib/bolt/executor.rb
+++ b/lib/bolt/executor.rb
@@ -1,4 +1,5 @@
-require 'logger'
+require 'bolt/logger'
+require 'json'
 require 'concurrent'
 require 'bolt/result'
 require 'bolt/config'
@@ -11,8 +12,8 @@ module Bolt
 
     def initialize(config = Bolt::Config.new, noop = nil)
       @config = config
-      @logger = Logger.new(config[:log_destination])
-      @logger.progname = 'executor'
+      @logger = Logger.instance(config[:log_destination])
+      # @logger.progname = 'executor'
       @logger.level = config[:log_level]
       @logger.formatter = Bolt::Formatter.new
       @noop = noop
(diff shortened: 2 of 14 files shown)


lib/bolt/cli.rb
```diff
@@ -425,7 +445,11 @@ def execute(options)
         nodes = executor.from_uris(options[:nodes])
 
         results = nil
+        mode = options[:mode]
+        action = options[:action]
+        object = options[:object]
         outputter.print_head
+        @logger.notice("Starting bolt #{mode} #{action} #{object}")
```
adreyer: This should be logged at info. The user just typed it in we don't need to log it by default

lib/bolt/cli.rb
```diff
@@ -460,6 +484,8 @@ def execute(options)
             end
         end
 
+        failures = count_failures(results)
+        @logger.notice { log_summary(mode, object, results.size, failures) }
```
adreyer: The human outputter prints this by default so it should be at info level to avoid duplication

lib/bolt/node/ssh.rb
```diff
@@ -127,14 +127,14 @@ def execute(command, sudoable: false, **options)
...
-            @logger.debug { "stdout: #{data}" }
+            @logger.info { "stdout: #{JSON.dump(data)}" }
           end
 
           channel.on_extended_data do |_, _, data|
             unless use_sudo && handled_sudo(channel, data)
               result_output.stderr << data
             end
-            @logger.debug { "stderr: #{data}" }
+            @logger.info { "stderr: #{JSON.dump(data)}" }
```
adreyer: These are subcommands and should continue to be logged at debug. For example chmoding a task before running it appears here

lib/bolt/node/ssh.rb
```diff
@@ -260,6 +260,7 @@ def _run_command(command)
     end
 
     def _run_script(script, arguments)
+      # This is redundant with the message in node.rb
       @logger.info { "Running script '#{script}'" }
```
adreyer: This should probably happen in the executor not each transport to avoid duplication

lib/bolt/node/winrm.rb
```diff
@@ -361,9 +361,9 @@ def execute(command, _ = {})
 
       output = @session.run(command) do |stdout, stderr|
         result_output.stdout << stdout
-        @logger.debug { "stdout: #{stdout}" }
```
adreyer: These are subcommands and should be logged at the debug level

lib/bolt/cli.rb
lucywyman: They're slightly different outputs, "Ran task exec on 2 nodes with 0 failures" and "Ran on 2 nodes in 0.72 seconds" -- agreed we don't need both, but not sure if one is preferred or if they should be combined or whatnot.

lib/bolt/node/ssh.rb
lucywyman: Should this still print the data as JSON?

lib/bolt/node/ssh.rb
lucywyman: It currently happens in both `node.rb` and `node/transport.rb`. Simplest solution is to remove the logs in `transport.rb`, should we do that and also move the `node.rb` log to the executor? 
https://github.com/puppetlabs/bolt/blob/master/lib/bolt/node.rb#L84
https://github.com/puppetlabs/bolt/blob/master/lib/bolt/node/ssh.rb#L263

lib/bolt/node/winrm.rb
lucywyman: JSON?

lib/bolt/cli.rb
```diff
@@ -425,7 +426,11 @@ def execute(options)
         nodes = executor.from_uris(options[:nodes])
 
         results = nil
+        mode = options[:mode]
+        action = options[:action]
+        object = options[:object]
         outputter.print_head
+        @logger.info("Starting bolt #{mode} #{action} #{object}")
```
lucywyman: Do we still need this?

lib/bolt/node.rb
```diff
@@ -76,12 +76,10 @@ def upload(source, destination)
     end
 
     def run_command(command)
-      @logger.info { "Running command: #{command}" }
```
lucywyman: Deduplicating

lib/bolt/executor.rb
```diff
@@ -68,34 +69,46 @@ def on(nodes, callback = nil)
     end
 
     def run_command(nodes, command)
+      @logger.info("Starting bolt command run #{command}")
       callback = block_given? ? Proc.new : nil
 
       on(nodes, callback) do |node|
-        node.run_command(command)
+        r = node.run_command(command)
+        @logger.info(JSON.dump(r.to_h))
```
adreyer: these should probably have some context. Got result for #{node}: JSON or something

adreyer: to_h is also probably wrong for these we use to_result for json output elsewhere. That should get cleaned up with BOLT-286

lib/bolt/executor.rb
```diff
@@ -68,34 +69,46 @@ def on(nodes, callback = nil)
     end
 
     def run_command(nodes, command)
+      @logger.info("Starting bolt command run #{command}")
```
adreyer: These need to include the node url.

lib/bolt/node/orch.rb
```diff
@@ -91,6 +91,7 @@ def unwrap_bolt_result(result)
         return result
       end
 
+      # Is there a reason the result isn't logged here like with other transports?
```
adreyer: Its retrievable from orchestrator. We could log the job-id for everything that goes through _run_task

lib/bolt/node/ssh.rb
adreyer: the run_ functions in node.rb are going away in BOLT-287 only logging this in the executor is probably best.

lib/bolt/node/winrm.rb
adreyer: no there is a good chance they're not JSON

modules/boltlib/lib/puppet/functions/run_task.rb
```diff
@@ -24,10 +26,25 @@
...
+           object,
+           node_count,
+           node_count == 1 ? '' : 's',
+           fail_count,
+           fail_count == 1 ? '' : 's')
+  end
+
   def run_task(task_name, targets, task_args = nil)
-    Bolt::ExecutionResult.from_bolt(
+    logger = Logger.new(STDOUT)
```
adreyer: We should never log to stdout. This needs access to a configured bolt logger

lib/bolt/cli.rb
adreyer: Where else would it happen?

modules/boltlib/lib/puppet/functions/run_task.rb
lucywyman: I'm not sure how to do that. Is there a way to pass a configured logger to that puppet function, or do I need to read the config in this file?

adreyer:
-> #233

PR #227, opened 2018-01-03
(BOLT-283) Assert that run_task arguments are Data compatible

MikaelSmith:
This commit adds add check that the user defined argument hash passed to
the `run_task` function is an instance of the `Data` data type.


MikaelSmith:
Updated vendored Puppet, replaces https://github.com/puppetlabs/bolt/pull/221.

PR #228, opened 2018-01-04
(maint) Remove use of System.ServiceModel.Web

MikaelSmith:
System.ServiceModel.Web was included as a relic of a previous iteration
of the powershell shim. It requires .NET Framework 3.5, which is newer
than we intended to support. Drop inclusion of that assembly, as it's
not used.


MikaelSmith:
I'd like to see verification that https://github.com/puppetlabs/pxp-agent/pull/692 works before merging this.

PR #230, opened 2018-01-04
(BOLT-293) Load Datatypes before running plans.

adreyer:
This ensures the datatypes we expose to plans are loaded before plans
are run. Otherwise there is a chance one will be returned by run_* but
it's methods will not be available.

### DIFF
--- a/spec/fixtures/modules/results/plans/test_methods.pp
+++ b/spec/fixtures/modules/results/plans/test_methods.pp
@@ -1,0 +1,15 @@
+plan results::test_methods(
+  String $target,
+  Optional[String] $fail = 'false'
+) {
+  if($fail) {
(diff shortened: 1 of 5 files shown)


adreyer:
I'm going to push some tests to this

spec/fixtures/modules/results/plans/test_methods.pp
```diff
@@ -0,0 +1,15 @@
+plan results::test_methods(
+  String $target,
+  Optional[String] $fail = 'false'
```
zreichert: Did you add $fail for a future test? And is the hashkey 'fail' an instrumentation for testing purposes?  Or does it serve a purpose other than testing?

PR #231, opened 2018-01-05
(BOLT-126) Support WinRM with Kerberos (kind of)

MikaelSmith:
Works unreliably on Linux, couldn't get to work on Windows or macOS yet.

### DIFF
--- a/lib/bolt/node/winrm.rb
+++ b/lib/bolt/node/winrm.rb
@@ -27,12 +27,20 @@ module Bolt
         transport = :ssl
       end
       endpoint = "#{scheme}://#{host}:#{port}/wsman"
+
+      if @kerberos
+        transport = :kerberos
+        realm = @kerberos
+      end
+
+      # Allow unspecified user/password for kerberos auth.
       options = { endpoint: endpoint,
-                  user: @user,
-                  password: @password,
+                  user: @user || '',
+                  password: @password || '',
                   retry_limit: 1,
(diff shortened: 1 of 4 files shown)


lib/bolt/node/winrm.rb
```diff
@@ -27,12 +27,20 @@ def connect
...
+      if @kerberos
+        transport = :kerberos
+        realm = @kerberos
+      end
+
+      # Allow unspecified user/password for kerberos auth.
       options = { endpoint: endpoint,
-                  user: @user,
-                  password: @password,
+                  user: @user || '',
```
MikaelSmith: The gem requires user/password, so I defaulted them to empty strings.

lib/bolt/node/winrm.rb
```diff
@@ -27,12 +27,20 @@ def connect
         transport = :ssl
       end
       endpoint = "#{scheme}://#{host}:#{port}/wsman"
+
+      if @kerberos
```
MikaelSmith: The result of making `kerberos` separate from `insecure`, with a default scheme of `http`, means you usually need to specify `--insecure` with `--kerberos`. We could change to kerberos implying HTTP... there's a chance that the kerberos "transport" can support HTTPS, but I haven't looked into the library enough to see if it'll do that.

MikaelSmith:
This so far has been pretty flaky. Closing until we can resume work on it.

PR #233, opened 2018-01-05
(BOLT-266) standardize logging and log actions from plans

adreyer:


lib/bolt/cli.rb
```diff
@@ -420,8 +420,10 @@ def execute(options)
       executor = Bolt::Executor.new(@config, options[:noop])
 
       if options[:mode] == 'plan'
+        executor = Bolt::Executor.new(@config, options[:noop], true)
```
MikaelSmith: Line 420 seems redundant.

lib/bolt/execution_result.rb
```diff
@@ -47,7 +47,7 @@ def empty
 
     def error_nodes
       result = {}
-      @result_hash.each_pair { |k, v| result[k] = v if v.is_a?(Error) }
+      @result_hash.each_pair { |k, v| result[k] = v if v.is_a?(Puppet::DataTypes::Error) }
```
MikaelSmith: Was this needed?

lib/bolt/config.rb
```diff
@@ -1,12 +1,11 @@
-require 'logger'
+require 'bolt/logger'
 require 'yaml'
 require 'bolt/cli'
 
 module Bolt
   Config = Struct.new(
     :concurrency,
     :format,
-    :log_destination,
```
MikaelSmith: You removed log_destination here but not log_level, and removed defaults for both later. Seems like log_level could also be removed.

lib/bolt/config.rb
```diff
@@ -187,6 +178,12 @@ def validate
         raise Bolt::CLIError, "Unsupported format: '#{self[:format]}'"
       end
 
+      if self[:transports][:ssh][:sudo_password] && self[:transports][:ssh][:run_as].nil?
+        logger = Logger.get_logger
+        logger.warn("'--sudo-password will not be used without specifying a " \
```
MikaelSmith: Should `--sudo-password` be wrapped in single-quotes? There's a starting single-quote but not ending.

modules/boltlib/lib/puppet/functions/file_upload.rb
```diff
@@ -18,6 +18,15 @@
     return_type 'ExecutionResult'
   end
 
+  def log_summary(object, node_count, fail_count)
```
MikaelSmith: This seems to be unused.

lib/bolt/execution_result.rb
adreyer: yes it fixed an unreported bug. I moved it to a separate commit. I'll write a test too

lib/bolt/config.rb
adreyer: No the default is handled elsewhere but it's still configurable. We never allowed log_destination to be configured.

lib/bolt/config.rb
adreyer: we don't do that a few lines below so I'll  just remove it

modules/boltlib/lib/puppet/functions/file_upload.rb
adreyer: yup

lib/bolt/cli.rb
adreyer: yup

lib/bolt/cli.rb
```diff
@@ -420,8 +420,10 @@ def execute(options)
       executor = Bolt::Executor.new(@config, options[:noop])
```
MikaelSmith: Still here.

lib/bolt/config.rb
adreyer: We still configure loglevel we just manage the default in the logger. log_destination is not configured yet and never has been.

lib/bolt/config.rb
adreyer: we don't quote a flag a few lines down so I removed it

PR #234, opened 2018-01-08
(MAINT) fix broken acceptance tests

zreichert:


acceptance/tests/plan_ssh.rb
```diff
@@ -21,8 +21,8 @@
                        "#{dir}/modules/test/plans/my_unix_plan.pp", <<-FILE)
     plan test::my_unix_plan($nodes) {
       $nodes_array = $nodes.split(',')
-      run_task(test::a_unix, $nodes_array)
-      run_task(test::b_unix, $nodes_array)
+      notice("${run_task(test::a_unix, $nodes_array)}")
```
MikaelSmith: We should still have some output, just formatted differently.

PR #235, opened 2018-01-08
(BOLT-295) Support running arbitrary scripts on Windows

MikaelSmith:
Support running arbitrary scripts on Windows as long as an extension
handler is registered and its extension added to PATHEXT.

Preserves support for running scripts without an extension as powershell
scripts, but will only do it if they have no extension (rather than when
they have an extension that doesn't match one of the built-in defaults:
.rb, .ps1, .pp).

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,15 +73,31 @@ ssh:
 
 `private-key`: The path to the private key file to use for SSH authentication.
 
-`connect-timeout`: Maximum amount of time to allow for an SSH connection to be established, in seconds
+`connect-timeout`: Maximum amount of time to allow for an SSH connection to be established, in seconds.
+
+`tmpdir`: The directory to store temporary files on the target node. (default: location used by `mktemp -d`, usually `/tmp`)
+
+`run-as`: Triggers privilege escalation for commands on the target node as the specified user. Currently only works via `sudo`.
 
 ### `winrm` transport configuration options
 
-`connect-timeout`: Maximum amount of time to allow for a WinRM connection to be established, in seconds
+`connect-timeout`: Maximum amount of time to allow for a WinRM connection to be established, in seconds.
+
+`insecure`: Whether to skip requiring SSL for connections. (default: false)
+
+`cacert`: The CA certificate used to authenticate SSL connections. (default: uses system CA certificates)
+
+`tmpdir`: The directory to store temporary files on the target node. (default: `[System.IO.Path]::GetTempPath()`)
 
 ### `pcp` transport configuration options
 
-There are currently no options for the PCP orchestrator transport
+`service-url`: The URL of the Orchestrator service, usually of the form `https://puppet:8143`. If not specified, will attempt to read local PE Client Tools configuration for the same setting from `orchestrator.conf`.
+
+`cacert`: The CA certificate used to authenticate the `service-url`. If not specified, will attempt to read local PE Client Tools configuration for the same setting from `orchestrator.conf`.
+
+`token-file`: The token certificate used to authorize requests to the `service-url`. If not specified, will attempt to read local PE Client Tools configuration for the same setting from `orchestrator.conf`. (default: `~/.puppetlabs/token`)
+
+`task-environment`: The environment from which Orchestrator will serve task implementations. (default: `production`)
 
 
 ## Usage examples
--- a/README.md
+++ b/README.md
@@ -89,6 +89,8 @@ ssh:
 
 `tmpdir`: The directory to store temporary files on the target node. (default: `[System.IO.Path]::GetTempPath()`)
 
+`extensions`: List of file extensions that will be accepted for scripts or tasks. Scripts with these file extensions will rely on the target node's file type association to run. For example, if Python is installed on the system, a `.py` script should run with `python.exe`. `.ps1`, `.rb`, and `.pp` are always allowed and run via hard-coded executables.
+
 ### `pcp` transport configuration options
 
 `service-url`: The URL of the Orchestrator service, usually of the form `https://puppet:8143`. If not specified, will attempt to read local PE Client Tools configuration for the same setting from `orchestrator.conf`.
--- a/lib/bolt/node/winrm.rb
+++ b/lib/bolt/node/winrm.rb
@@ -466,8 +466,6 @@ PS
 
     def with_remote_file(file)
       file_base = File.basename(file)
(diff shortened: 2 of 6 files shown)


MikaelSmith:
A manual test I've tried is: install Python, set user environment PATHEXT to include '.PY', and run an example 'file.py' as both a script and a task. I'm not sure how much we want to automate this, since it requires installing python.exe. It makes more sense to me as an acceptance test if anything.

lib/bolt/node/winrm.rb
```diff
@@ -462,11 +466,11 @@ def make_tempdir
     end
 
     def with_remote_file(file)
-      ext = File.extname(file)
-      ext = VALID_EXTENSIONS.include?(ext) ? ext : '.ps1'
-      file_base = File.basename(file, '.*')
+      file_base = File.basename(file)
+      # Default to powershell if no extension present
```
adreyer: Does this add value? Is there a real chance we'll get a powershell script without a .ps1 extension?

MikaelSmith: I'm not sure, it was there originally. Our integration tests rely on this behavior, but could be updated not to.

lib/bolt/node/winrm.rb
```diff
@@ -428,6 +426,12 @@ def process_from_extension(path)
           'puppet.bat',
           ['apply', "\"#{path}\""]
         ]
+      else
+        # Run the script via cmd, letting Windows extension handling determine how
+        [
+          'cmd.exe',
+          ['/c', "\"#{path}\""]
```
Iristyle: This may not work how you think if the default action associated with the file extension is something like `Edit` or `Open in Editor` rather than something like `Run` or `Execute`.

We intentionally chose not to go this route and to stick to a whitelist because you have no control over how the file associations are setup, and could end up popping a Window'd UI rather than actually running code through an appropriate interpreter.

lib/bolt/node/winrm.rb
Iristyle: I would hope you don't... also seems better to be explicit in the case where you've written code that's not PowerShell, it confuses the PS interpreter, and accidentally does bad things.

lib/bolt/node/winrm.rb
Iristyle: Also - for the general approach, I think this was getting kicked down the road until users could register their own extensions inside the Bolt config.

MikaelSmith: That is a concern, but I'm not sure how a whitelist helps with that. This also prevents Bolt from running tasks that would be valid in PE.

If we want it to be node-specific, Windows already has one: PATHEXT.

If we made it configurable, users would be able to specify other things to try that might misbehave as you say. I'm not sure how doing that via an extra config option is terribly different than doing it by selecting something with a specific file extension.

If it's not user configurable, then we simply prevent users from doing something they want to do.

MikaelSmith: Yes, [BOLT-146](https://tickets.puppetlabs.com/browse/BOLT-146). But that ticket is mostly targeting other more complicated scenarios.

I plan to look at BOLT-146 soon. If we don't believe this is desirable without BOLT-146, I can focus on that instead.

Iristyle: Yeah, totally understand that we want users to be able to have flexibility in what they run.... no disagreement there.

I'm just not sure this is the way to go about it. For instance, create an empty text file on disk ending in `.txt`. From a terminal run `cmd.exe /c foo.txt`

MikaelSmith: Ah, you're right about `PATHEXT`. I think I assumed it was a problem rather than how Bolt was treating files with extensions it didn't recognize (if it didn't recognize the extension, it would try to run it as a .ps1 file).

Iristyle: Yeah, I'm lukewarm on treating extensionless files as PowerShell, because extensionless files are foreign to Windows users / potentially confusing... so wouldn't be totally upset if extensions were made mandatory

MikaelSmith:
After discussion with my team, I plan to update this to include a new config option that allows you to white-list extensions. This seems like a useful low-cost addition until we can get to BOLT-146.

I'd like to have `.ps1`, `.rb`, `.py`, `.bat`, `.pp` all work out of the box. Should they be a default of that option, and if you want to set it yourself you have to include them? Or a baseline that the option adds to?

MikaelSmith:
Updated to add an extensions whitelist.

lib/bolt/node/winrm.rb
```diff
@@ -428,6 +431,12 @@ def process_from_extension(path)
           'puppet.bat',
           ['apply', "\"#{path}\""]
         ]
+      else
```
Iristyle: Do you want to consider adding `.py` to the list of defaults? Seems common enough?

MikaelSmith: Sure. I considered it, hadn't made a decision so I went with the conservative option.

README.md
```diff
@@ -73,15 +73,33 @@ ssh:
...
-`connect-timeout`: Maximum amount of time to allow for a WinRM connection to be established, in seconds
+`connect-timeout`: Maximum amount of time to allow for a WinRM connection to be established, in seconds.
+
+`insecure`: Whether to skip requiring SSL for connections. (default: false)
+
+`cacert`: The CA certificate used to authenticate SSL connections. (default: uses system CA certificates)
+
+`tmpdir`: The directory to store temporary files on the target node. (default: `[System.IO.Path]::GetTempPath()`)
+
+`extensions`: List of file extensions that will be accepted for scripts or tasks. `.ps1`, `.rb`, and `.pp` are always allowed.
```
Iristyle: Can you add a small note / example about how additional extensions are evaluated?  I think it would be helpful to Windows users to understand precisely what this means...

Iristyle:
👍  to this general direction over the previous PR iteration... Adds flexibility, without a lot of additional overhead.

My only remaining comment was about having a useful example in docs

README.md
MikaelSmith: Added.

lib/bolt/node/winrm.rb
MikaelSmith: I think I'll leave it out for now without more demand. It should be fairly clear what's needed to add it when it doesn't work.

PR #237, opened 2018-01-10
(BOLT-285) Overhaul error handling around Puppet code evaluation

MikaelSmith:
Overhauls error handling when calling Puppet functions to catch exceptions before they're passed to the PAL compiler and rethrow them outside the PAL API. This allows us to work with undecorated exceptions and provide simpler error messages. It also prevents Puppet from logging when we would rather display the message directly.

For errors running Bolt subcommands:
- When using human-format output, print errors to stdout and color them
red rather than printing to stderr.
- When using json-format output, errors will no longer be reprinted on
stderr.

Errors parsing Bolt configuration will continue to be output on stderr.

### DIFF
--- a/exe/bolt
+++ b/exe/bolt
@@ -10,6 +10,5 @@ begin
 rescue Bolt::CLIExit
   exit
 rescue Bolt::CLIError => e
-  warn e.message
   exit e.error_code
 end
--- a/spec/bolt/cli_spec.rb
+++ b/spec/bolt/cli_spec.rb
@@ -629,8 +629,7 @@ NODES
             mode: 'task',
             action: 'show'
           }
-          expect(Puppet).to receive(:err).with(/unexpected token at/)
-          expect { cli.execute(options) }.to raise_error "Failure while reading task metadata"
+          expect { cli.execute(options) }.to raise_error(/unexpected token at/)
         end
       end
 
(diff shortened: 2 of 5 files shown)


exe/bolt
```diff
@@ -10,6 +10,6 @@ begin
 rescue Bolt::CLIExit
   exit
 rescue Bolt::CLIError => e
-  warn e.message
+  warn($stderr.isatty ? "\033[1;31m#{e.message}\033[0m" : e.message)
```
MikaelSmith: This choice of coloring may be controversial.

adreyer: I think we should not color at this level since there will be no way to disable it.

But I also think its a mistake to let so much bubble up to this level instead of handling it with the outputter and the logger. Errors should be handled through them if bolt is running normally and this should be left for extreme cases.

spec/bolt/cli_spec.rb
```diff
@@ -728,7 +741,7 @@ def stub_config(config, file_content = nil)
             task_options: task_params
           }
           expect { cli.execute(options) }.to raise_error(
-            Bolt::CLIError, /Task not found: dne::task1/
+            Bolt::CLIError, /^Task not found: dne::task1/
```
adreyer: Is there something you think might wrap this message that wasn't intentional?

MikaelSmith: Ah, that may be unnecessary since it's just checking the exception.

MikaelSmith:
Simplified the PR. An outcome of not using Puppet's logging for many errors is that they're no longer colored.

MikaelSmith:
Yes, that's what I was agreeing with. The CLI tests assume the exception is thrown and then written, so if we start writing it through the outputter the tests will have a side-effect of writing a bunch to stdout; that just needs to be cleaned up.

MikaelSmith:
Updated. Execution errors will no longer be printed on stderr, which is a bit of a change.

PR #239, opened 2018-01-11
Revert "(BOLT-252) change raketask to work with LAYOUT ENV"

MikaelSmith:
This reverts commit b6ed2a3. Ended up being unused by CI.

### DIFF
--- a/acceptance/Rakefile
+++ b/acceptance/Rakefile
@@ -2,19 +2,29 @@ require 'rototiller'
 
 desc "Generate Beaker Host config"
 rototiller_task :host_config do |task|
-  task.add_env(name: 'LAYOUT', message: <<-MESSAGE)
-A host generator string is required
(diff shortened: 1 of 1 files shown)


MikaelSmith:
If LAYOUT's no longer used, we may instead want to go back to BOLT_CONTROLLER and BOLT_NODES.

zreichert:
@MikaelSmith Whatever we find most valuable.  If you find the fine grained control valuable then we can revert the PR that changed it.  I wanted to have CI use the Rakefile but I was not successful.

MikaelSmith:
I think the previous was a bit more usable. I don't see the flexibility of controlling the full LAYOUT as helpful.

MikaelSmith:
Switched to reverting the Rakefile change instead.

MikaelSmith:
poke @zreichert

PR #240, opened 2018-01-11
(BOLT-81) implement disable_ssh_agent option

tkishel:
Prior to this commit, bolt reported an error when Net-SSH
attempted to use an unconfigured ssh-agent.

Net-SSH offers a use_agent option which can be used to skip ssh-agent.

With this commit, bolt offers a disable-ssh-agent option
which can be used to used to specifiy the Net-SSH use_agent option.

PR #241, opened 2018-01-11
(BOLT-81) check for ssh_agent

tkishel:
Prior to this commit, bolt reported an error when Net-SSH
attempted to use an unconfigured ssh-agent.

Net-SSH offers a use_agent option which can be used to skip ssh-agent.

With this commit, bolt checks the environment to see if ssh-agent
is configured and its socket exists.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -33,6 +33,9 @@ Style/GuardClause:
 Style/MultilineBlockChain:
   Enabled: false
 
+Style/DoubleNegation:
+  Enabled: false
+
 # Disable nearly all Metrics checks. These seem better off left to judgement.
 
 Metrics/AbcSize:
--- a/lib/bolt/node/ssh.rb
+++ b/lib/bolt/node/ssh.rb
@@ -20,20 +20,18 @@ module Bolt
       @conf_run_as = @run_as
     end
 
-    # Mirroring:
-    # https://github.com/net-ssh/net-ssh/blob/master/lib/net/ssh/authentication/pageant.rb#L403
+    def protocol
+      'ssh'
+    end
+
     if !!File::ALT_SEPARATOR
       require 'ffi'
       module Win
         extend FFI::Library
         ffi_lib 'user32'
         ffi_convention :stdcall
-        attach_function :FindWindow, :FindWindowA, %i[string string], :int
+        attach_function :FindWindow, :FindWindowW, %i[buffer_in buffer_in], :int
       end
-    end
-
-    def protocol
-      'ssh'
     end
 
     def connect
@@ -54,12 +52,20 @@ module Bolt
                                   end
       options[:timeout] = @connect_timeout if @connect_timeout
 
-      if defined?(UNIXSocket) && UNIXSocket && ENV['SSH_AUTH_SOCK'].to_s.empty?
-        @logger.debug { "Disabling use_agent in net-ssh: ssh-agent is not available" }
-        options[:use_agent] = false
-      elsif !!File::ALT_SEPARATOR && Win.FindWindow('Pageant', 'Pageant').to_i == 0
-        @logger.debug { "Disabling use_agent in net-ssh: pageant process not running" }
-        options[:use_agent] = false
+      # Mirroring:
+      # https://github.com/net-ssh/net-ssh/blob/master/lib/net/ssh/authentication/agent.rb#L80
+      # https://github.com/net-ssh/net-ssh/blob/master/lib/net/ssh/authentication/pageant.rb#L403
+      if defined?(UNIXSocket) && UNIXSocket
+        if ENV['SSH_AUTH_SOCK'].to_s.empty?
+          @logger.debug { "Disabling use_agent in net-ssh: ssh-agent is not available" }
+          options[:use_agent] = false
+        end
+      elsif !!File::ALT_SEPARATOR
+        pageant_wide = 'Pageant'.encode('UTF-16LE')
+        if Win.FindWindow(pageant_wide, pageant_wide).to_i == 0
+          @logger.debug { "Disabling use_agent in net-ssh: pageant process not running" }
+          options[:use_agent] = false
+        end
       end
 
       @session = Net::SSH.start(@target.host, @user, options)
--- a/lib/bolt/node/ssh.rb
+++ b/lib/bolt/node/ssh.rb
@@ -20,6 +20,18 @@ module Bolt
       @conf_run_as = @run_as
     end
 
+    # Mirroring:
+    # https://github.com/net-ssh/net-ssh/blob/master/lib/net/ssh/authentication/pageant.rb#L403
+    if !!File::ALT_SEPARATOR
+      require 'ffi'
+      module Win
+        extend FFI::Library
+        ffi_lib 'user32'
+        ffi_convention :stdcall
+        attach_function :FindWindow, :FindWindowA, %i[string string], :int
+      end
+    end
+
     def protocol
       'ssh'
     end
@@ -41,6 +53,14 @@ module Bolt
                                     Net::SSH::Verifiers::Secure.new
                                   end
       options[:timeout] = @connect_timeout if @connect_timeout
+
+      if defined?(UNIXSocket) && UNIXSocket && ENV['SSH_AUTH_SOCK'].to_s.empty?
+        @logger.debug { "Disabling use_agent in net-ssh: ssh-agent is not available" }
+        options[:use_agent] = false
+      elsif !!File::ALT_SEPARATOR && Win.FindWindow('Pageant', 'Pageant').to_i == 0
+        @logger.debug { "Disabling use_agent in net-ssh: pageant process not running" }
+        options[:use_agent] = false
+      end
 
       @session = Net::SSH.start(@target.host, @user, options)
       @logger.debug { "Opened session" }


lib/bolt/node/ssh.rb
```diff
@@ -33,6 +33,8 @@ def connect
                                     Net::SSH::Verifiers::Secure.new
                                   end
       options[:timeout] = @connect_timeout if @connect_timeout
+      # Only use ssh-agent if it is configured and its socket exists.
+      options[:use_agent] = ENV['SSH_AUTH_SOCK'].to_s && File.socket?(ENV['SSH_AUTH_SOCK'])
```
adreyer: lets add
```
@logger.debug("Could not connect to ssh agent.")
```

joshcooper: Does checking the socket work on Windows? I’m not sure if putty’s ssh agent uses named pipes?

tkishel: I see now that we shouldn't do File.socket validation on the value of SSH_AUTH_SOCK. If the user specifies an invalid value for SSH_AUTH_SOCK, that should be net-ssh's responsibility.

net-ssh `use_agent` defaults to true, and when true, net-ssh agent `connect()` expects an agent to be configured. Since we don't allow configuration of `use_agent` within bolt, we could implement the following checks within bolt to avoid the "Agent not configured" or "pageant process not running" errors:

https://github.com/net-ssh/net-ssh/blob/master/lib/net/ssh/authentication/agent.rb#L85
https://github.com/net-ssh/net-ssh/blob/master/lib/net/ssh/authentication/pageant.rb#L403

Amended PR, with debugging. Thanks, all.

lib/bolt/node/ssh.rb
```diff
@@ -14,6 +14,11 @@ def self.initialize_transport(logger)
       }
     end
 
+    if Gem.win_platform?
+      require 'Win32API'
+      FindWindow = Win32API.new('user32', 'FindWindow', %w[P P], "L")
+    end
+
```
tkishel: `FindWindow = Win32API.new` threw an error when executed within connect().

.rubocop.yml
```diff
@@ -63,4 +63,4 @@ Metrics/ParameterLists:
   Enabled: false
 
 Metrics/PerceivedComplexity:
-  Enabled: false
+  Enabled: false
```
Iristyle: Shouldn't need to edit this file.

lib/bolt/node/ssh.rb
```diff
@@ -14,6 +14,16 @@ def self.initialize_transport(logger)
       }
     end
 
+    if Gem.win_platform?
+      require 'ffi'
+      module Win
+        extend FFI::Library
+        ffi_lib 'user32'
+        ffi_convention :stdcall
+        attach_function :FindWindow, :FindWindowA, %i[string string], :int
```
Iristyle: Should use `FindWindowW`, not `FindWindowA` ... which means using wide strings.

lib/bolt/node/ssh.rb
```diff
@@ -14,6 +14,16 @@ def self.initialize_transport(logger)
       }
     end
 
+    if Gem.win_platform?
```
Iristyle: Don't use `Gem.win_platform?` - instead use `!!File::ALT_SEPARATOR` to check for Windows like we do in https://github.com/puppetlabs/puppet/blob/ebd96213cab43bb2a8071b7ac0206c3ed0be8e58/lib/puppet/util/platform.rb#L10

This is compiled into Ruby itself and won't change.

lib/bolt/node/ssh.rb
```diff
@@ -34,6 +44,14 @@ def connect
                                   end
       options[:timeout] = @connect_timeout if @connect_timeout
 
+      if defined?(UNIXSocket) && UNIXSocket && ENV['SSH_AUTH_SOCK'].to_s.empty?
+        @logger.debug { "Disabling use_agent in net-ssh: ssh-agent is not available" }
+        options[:use_agent] = false
+      elsif Gem.win_platform? && Win.FindWindow('Pageant', 'Pageant').to_i == 0
```
Iristyle: FWIW - checking for a specific putty process (by looking for its Window) is probably not a great heuristic to use.  I'm sure it won't work with git bash for instance, or other tools that may be part of Cygwin.

It seems like there should be something more reliable that can identify a running ssh-agent on Windows.

Iristyle: Interestingly, posh-git does a similar thing (but looks for the process, not the window name, which might not work).

But it also does a number of additional things - https://github.com/dahlbyk/posh-git/blob/2ecf05a55b9cacb2e9bcc20321cccd983e25a428/src/GitUtils.ps1#L395-L415

adreyer: This is what net-ssh does I don't think there is much value in us doing better than net-ssh
https://github.com/net-ssh/net-ssh/blob/master/lib/net/ssh/authentication/pageant.rb#L403

.rubocop.yml
tkishel: True. I had disabled Style::SymbolArray then found I could use %i since bolt is Ruby 2.0+ Deleted instead of reverted but missed the newline. Will do.

lib/bolt/node/ssh.rb
tkishel: I initially tried FindWindowW:

```
attach_function :FindWindowW, :FindWindowW, %i[buffer_in buffer_in], :int
```
But it always returns 0, even when `FindWindowA` succeeds.

Any advice?

lib/bolt/node/ssh.rb
tkishel: Will do!

Though I found the same in bolt and pdk, and most importantly in:

https://github.com/net-ssh/net-ssh/blob/master/lib/net/ssh/authentication/agent.rb#L9

which I modeled.

lib/bolt/node/ssh.rb
tkishel: True, but I specifically modeled the code in net-ssh in an attempt to symmetrically unwind its `use_agent` setting.

I still prefer https://github.com/puppetlabs/bolt/pull/240 which would allow bolt users to set `use_agent`, rather than hard code the default of `use_agent=true` and import these implementation details from net-ssh into bolt.

lib/bolt/node/ssh.rb
Iristyle: You basically need to do an `.encode('UTF-16LE')` on the string to make sure its a wide string before passing. We implement some helpers in Puppet itself, but probably can't rely on that here - https://github.com/puppetlabs/puppet/blob/master/lib/puppet/util/windows/api_types.rb#L23-L34

Still, it seems getting a process list, rather than searching by window name is preferred? I'm not sure why `pageant` has a window?

lib/bolt/node/ssh.rb
Iristyle: Yeah, understand your point @adreyer ... haven't looked at net-ssh code too closely, but is there not a way to call to one of its helpers? I see `Net::SSH::Authentication:: Pageant::Socket.new` will raise if there's no Putty based ssh-agent. Of course, that won't help for others...

If we're relying on Net::SSH anyhow, seems like the best place to have a check is in that library that we can rely on?

lib/bolt/node/ssh.rb
tkishel: Using `encode` works, thank you! I'll push tomorrow and then look at searching the process list, but have been avoiding deviating from net-ssh's tests.

tkishel:
I cannot find this comment on GitHub, so replying via email ...

> Yeah, understand your point @adreyer ... haven't looked at net-ssh code too closely, but is there not a way to call to one of its helpers? I see Net::SSH::Authentication:: Pageant::Socket.new will raise if there's no Putty based ssh-agent. Of course, that won't help for others...
>
> If we're relying on Net::SSH anyhow, seems like the best place to have a check is in that library that we can rely on?

I've tried to rescue the Net::SSH:: Authentication::AgentNotAvailable
error within lib/bolt/node/ssh.rb, similar to what Nick Lewis
suggested in Jira. No success.

I've looked for something more 'public' in Net-SSH to use. Nothing
found, other than allowing the bolt user to set 'net-ssh::use_agent'
as per PR 240.

lib/bolt/node/ssh.rb
tkishel: Oh ...

I've tried to rescue the Net::SSH:: Authentication::AgentNotAvailable
error within lib/bolt/node/ssh.rb, similar to what Nick Lewis
suggested in Jira. No success, and might not capture the error message anyway.

I've looked for something more 'public' in Net-SSH to use. Nothing
found, other than allowing the bolt user to set 'net-ssh::use_agent'
as per PR 240.

lib/bolt/node/ssh.rb
mruzicka: FWIW, this is how putty itself checks for the `pageant`'s presence: https://github.com/github/putty/blob/49fb598b0e78d09d6a2a42679ee0649df482090e/windows/winpgntc.c#L18-L26
So using the `FindWindowW` seems appropriate.

MikaelSmith:
This needs a rebase now. It should be a pretty simple rebase.

PR #243, opened 2018-01-11
(BOLT-289) Move NodeURI to Target, use Target instead of Node

MikaelSmith:
Moves NodeURI functionality to Target. Uses Target instead of Node where
we only need target info.


lib/bolt/node.rb
```diff
@@ -53,15 +43,27 @@ def initialize(host, port = nil, user = nil, password = nil, uri: nil,
...
+    def host
+      @target.host
+    end
+
+    def port
+      @target.port
+    end
+
+    def uri
+      @target.uri.to_s.sub(%r{^/+}, '')
```
MikaelSmith: Choosing to not preserve the original uri string and instead try to generate it from an Addressable::URI has some interesting side-effects (if there's no protocol/scheme, it still adds a `//` prefix that we don't really want). I may go back to preserving the original uri string separately.

lib/bolt/cli.rb
```diff
@@ -425,7 +426,7 @@ def execute(options)
         execute_plan(executor, options)
       else
         executor = Bolt::Executor.new(@config, options[:noop])
-        nodes = executor.from_uris(options[:nodes])
+        targets = options[:nodes]
```
MikaelSmith: This part feels gross, but we use `--nodes` on the CLI so that name's going to be present somewhere. I'd really like to get ride of the `Node` class soon so it gets less confusing again.

lib/bolt/executor.rb
```diff
@@ -17,11 +18,12 @@ def initialize(config = Bolt::Config.new, noop = nil, plan_logging = false)
       @plan_logging = plan_logging
     end
 
-    def from_uris(nodes)
-      nodes.map do |node|
-        Bolt::Node.from_uri(node, config: @config)
+    def from_targets(targets)
```
MikaelSmith: This is now only used internally to - essentially - choose a transport for each node that's used in the `on` method.

lib/bolt/outputter/human.rb
```diff
@@ -25,24 +25,24 @@ def remove_trail(string)
...
 
-      def print_event(node, event)
+      def print_event(target, event)
         case event[:type]
         when :node_start
-          print_start(node)
+          print_start(target)
         when :node_result
-          print_result(node, event[:result])
+          print_result(event[:result])
```
MikaelSmith: Since result has a target, target is redundant. This creates a kind of asymmetry with start events that I don't much like.

MikaelSmith: Should we just make `:target` a part of the node_start event?

modules/boltlib/lib/puppet/datatypes/target.rb
```diff
@@ -1,7 +1,7 @@
 Puppet::DataTypes.create_type('Target') do
```
MikaelSmith: I don't think we have any tests that use this.

adreyer: no I don't think people were really using it before but we should probably create some minimal integration tests at least.

lib/bolt/executor.rb
```diff
@@ -64,15 +66,18 @@ def on(nodes, callback = nil)
...
       fc = result.select { |_, r| r.error }.length
       npl = result.length == 1 ? '' : 's'
       fpl = fc == 1 ? '' : 's'
       "Ran #{action} '#{object}' on #{result.length} node#{npl} with #{fc} failure#{fpl}"
     end
+    private :summary
 
-    def run_command(nodes, command)
+    def run_command(targets, command)
+      nodes = from_targets(targets)
```
adreyer: Is there a reason to call this in the run_ methods instead of in on?

lib/bolt/node.rb
```diff
@@ -53,15 +43,27 @@ def initialize(host, port = nil, user = nil, password = nil, uri: nil,
       @orch_task_environment = transport_conf[:orch_task_environment]
       @extensions = transport_conf[:extensions]
 
-      @logger = Logger.get_logger(progname: @host)
-      @transport_logger = Logger.get_logger(progname: @host, log_level: Logger::WARN)
+      @logger = Logger.get_logger(progname: host)
+      @transport_logger = Logger.get_logger(progname: host, log_level: Logger::WARN)
+    end
+
+    def host
```
adreyer: Do we use these enough to justify the top level methods for them? We'll have to translate `host
 -> target.host` and `port -> target.port` in the transports at some point anyway. I don't know if thats worth fixing now or later.

lib/bolt/node/ssh.rb
```diff
@@ -24,7 +24,7 @@ def connect
         non_interactive: true
       }
 
-      options[:port] = @port if @port
+      options[:port] = port if port
```
adreyer: regarding dropping the port/host/uri methods on Nod: this would `options[:port] = target.port if target.port`

lib/bolt/executor.rb
MikaelSmith: I felt it made the call to `on` slightly more readable. The alternative would pass targets, but then still get a `node` passed to the block.

MikaelSmith:
I'm a little stumped on why the host key verification test isn't throwing.

spec/integration/execution_result_spec.rb
```diff
@@ -3,21 +3,25 @@
...
-  let(:user) { conn_info('ssh')[:user] }
-  let(:password) { conn_info('ssh')[:password] }
+  let(:uri) { conn_uri('ssh', true) }
 
   after(:each) { Puppet.settings.send(:clear_everything_for_tests) }
 
   context 'when using CLI options' do
     let(:config_flags) { %W[--insecure --format json --modulepath #{modulepath}] }
 
+    it 'returns true on success' do
```
MikaelSmith: These integration tests could use something to suppress the logging output. With the change to log notices from a plan by default, we now get them showing up during test runs.

spec/bolt/node/ssh_spec.rb
```diff
@@ -9,21 +9,30 @@
...
   include BoltSpec::Files
 
+  def mk_config(conf)
+    Bolt::Config.new(transports: { ssh: conf })
+  end
+
   let(:hostname) { ENV['BOLT_SSH_HOST'] || "localhost" }
   let(:user) { ENV['BOLT_SSH_USER'] || "vagrant" }
   let(:password) { ENV['BOLT_SSH_PASSWORD'] || "vagrant" }
   let(:port) { ENV['BOLT_SSH_PORT'] || 2224 }
```
MikaelSmith: The failing test only passes in Travis if `port` is a string. I'm flummoxed.

MikaelSmith: In Travis, this test relies on checking for `localhost:22` rather than `localhost`, because the Travis image already has `localhost` (specifically, `[1]`) as an entry in `known_hosts`. Net::SSH will only check `localhost:22` if the port is a string; if it's an integer == 22, it checks `localhost`.

MikaelSmith:
Fixing rubocop.

PR #244, opened 2018-01-11
(BOLT-268) Switch from built-in Logger to Logging gem

nicklewis:
This commit swaps out the built-in Ruby Logger class for the Logging
gem, modeled after log4j.

This allows us much better control over which components log at which
levels, as well as allowing multiple log destinations (ie. both STDERR
and a file). It also includes built-in support for problems like
colorization that we would otherwise need to roll on our own.

### DIFF
--- a/lib/bolt/cli.rb
+++ b/lib/bolt/cli.rb
@@ -1,8 +1,9 @@
 require 'uri'
 require 'optparse'
 require 'benchmark'
-require 'bolt/logger'
 require 'json'
+require 'logging'
+require 'bolt/logger'
 require 'bolt/node'
 require 'bolt/version'
 require 'bolt/error'
@@ -103,12 +104,14 @@ HELP
     attr_accessor :options
 
     def initialize(argv)
+      Bolt::Logger.initialize_logging
       @argv    = argv
       @options = {
         nodes: []
       }
       @config = Bolt::Config.new
       @parser = create_option_parser(@options)
+      @logger = Logging.logger[self]
     end
 
     def create_option_parser(results)
@@ -248,8 +251,8 @@ HELP
 
       @config.load_file(options[:configfile])
       @config.update_from_cli(options)
-      Logger.configure(@config)
       @config.validate
+      Logging.logger[:root].level = @config[:log_level] || :notice
 
       # This section handles parsing non-flag options which are
       # mode specific rather then part of the config
@@ -393,7 +396,7 @@ HELP
         end
 
         Puppet::Util::Log.newdestination(:console)
-        Puppet[:log_level] = if @config[:log_level] == Logger::DEBUG
+        Puppet[:log_level] = if @config[:log_level] == :debug
                                'debug'
                              else
                                'notice'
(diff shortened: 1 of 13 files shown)


nicklewis:
Rubocop's a jerk.

lib/bolt/node/ssh.rb
```diff
@@ -19,8 +20,10 @@ def protocol
     end
 
     def connect
+      transport_logger = Logging.logger[Net::SSH]
+      transport_logger.level = :warn
       options = {
-        logger: @transport_logger,
```
MikaelSmith: What was transport_logger used for?

nicklewis: It's a separate logger that gets passed to `Net::SSH`, so that we can control the verbosity of its messages separate from the rest of the system.

MikaelSmith: Ah, but you got rid of the `@transport_logger` variable.

lib/bolt/cli.rb
```diff
@@ -248,8 +251,8 @@ def parse
 
       @config.load_file(options[:configfile])
       @config.update_from_cli(options)
-      Logger.configure(@config)
       @config.validate
+      Logging.logger[:root].level = @config[:log_level] || :notice
```
adreyer: By doing this we don't have to worry about loggers that may already be created in other class instances?

nicklewis: My understanding is: Because the only appender is defined on the root logger, logging to any other logger will do nothing directly. But because the other loggers are additive, they will also pass the event up to the root logger, which will then choose whether to log it based on this level.

But based on that understanding, it doesn't seem like the thing we do to set the net-ssh logging to `warn` should work (since it should still be passed up to the root logger, which will log it because it's at notice level)... so let me investigate further.

nicklewis: Okay, that's not quite true.

It turns out that the logger that initially receives the log method (`debug()`, `warn()`) decides whether it will be logged. When passing the event up its parent chain (which it will only do if the event was at an appropriate 
 level), it skips checking the level of its parent loggers and will log unconditionally. All of that is true *unless* a specific appender has been configured with a less verbose log level than its logger. However, the default log level for all appenders is `all`, meaning they will log any message given to them, deferring all decisions about level to the logger itself.

Changing the log level on a logger will automatically propagate the change to its children, assuming they haven't *explicitly* set a level.

The short version is that this line of code will affect all loggers that don't specify a log level, whether they've been created yet or not.

adreyer: cool

lib/bolt/node/ssh.rb
```diff
@@ -19,8 +20,10 @@ def protocol
     end
 
     def connect
+      transport_logger = Logging.logger[Net::SSH]
+      transport_logger.level = :warn
```
Sharpie: Hmm. Won't this just perpetually re-set the log level to `WARN` ever time a new connection is opened? What does one do if they need `DEBUG` logs for a connection issue?

nicklewis: Yes, it will. That was the previous behavior so I preserved it while switching to the new logging library. Now that we have better facilities in place to control log level for these sorts of things, I think we need to do a pass through and figure out the behavior of `--verbose` and `--debug`. It definitely seems like it makes sense for everything to log at debug if specified.

PR #246, opened 2018-01-12
(BOLT-222) run_* functions will raise an exception when they fail

adreyer:
Previously the run_* functions would return ExecutionResults containing
failures by default leaving it up to the plan author to check `ok`
before continting. Now run_* functions will check the ExecutionResult
and raise a RunFailure exception unless they are successful. The
'_abort' option can be passed to allow failing results to be returned

### DIFF
--- a/lib/bolt/cli.rb
+++ b/lib/bolt/cli.rb
@@ -13,11 +13,9 @@ require 'io/console'
 
 module Bolt
   class CLIError < Bolt::Error
-    attr_reader :error_code
-
-    def initialize(msg, error_code: 1)
+    def initialize(msg)
       super(msg, "bolt/cli-error")
-      @error_code = error_code
+      @error_code = error_code if error_code
     end
   end
 
@@ -465,7 +463,7 @@ HELP
 
         outputter.print_summary(results, elapsed_time)
       end
-    rescue Bolt::CLIError => e
+    rescue Bolt::Error => e
       outputter.fatal_error(e)
       raise e
     end
@@ -516,8 +514,9 @@ HELP
       @outputter ||= Bolt::Outputter.for_format(@config[:format])
     end
 
-    # Runs a block in a PAL script compiler configured for Bolt.
-    # Catches exceptions thrown by the block and re-raises them as Bolt::CLIError.
+    # Runs a block in a PAL script compiler configured for Bolt.  Catches
+    # exceptions thrown by the block and re-raises them ensuring they are
+    # Bolt::Errors since the script compiler block will squash all exceptions.
     def in_bolt_compiler(opts = [])
       Puppet.initialize_settings(opts)
       r = Puppet::Pal.in_tmp_environment('bolt', modulepath: [BOLTLIB_PATH] + @config[:modulepath], facts: {}) do |pal|
@@ -525,14 +524,26 @@ HELP
           begin
             yield compiler
           rescue Puppet::PreformattedError => err
-            err.cause
+            # Puppet sometimes rescues exceptions notes the location and reraises
+            # For now return the original error.
+            if err.cause
+              if err.cause.is_a? Bolt::Error
+                err.cause
+              else
+                Bolt::CLIError.new(err.cause.message)
+              end
+            else
+              Bolt::CLIError.new(err.message)
+            end
           rescue StandardError => err
-            err
+            Bolt::CLIError.new(err.message)
           end
         end
       end
 
-      raise Bolt::CLIError, r.message if r.is_a? StandardError
+      if r.is_a? StandardError
+        raise r
+      end
       r
     end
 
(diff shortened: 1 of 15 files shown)


lib/bolt/cli.rb
```diff
@@ -532,7 +530,10 @@ def in_bolt_compiler(opts = [])
         end
       end
 
-      raise Bolt::CLIError, r.message if r.is_a? StandardError
+      if r.is_a? StandardError
+        r = Bolt::CLIError.new(r.message) unless r.is_a? Bolt::Error
```
nicklewis: I don't quite follow this code. Should it just be another rescue inside the block above? It's not clear to me what `e.cause` returns for a `Puppet::PreformattedError`...

adreyer: There are two things going on.
I'm not sure what the goal of returning e.cause is. At least in some cases it's silently squashing all output into a nil return value.

I can move some of this logic up into the compiler block and rely on rescue instead of conditionals which seems better.

lib/bolt/cli.rb
```diff
@@ -525,14 +523,23 @@ def in_bolt_compiler(opts = [])
           begin
             yield compiler
           rescue Puppet::PreformattedError => err
-            err.cause
-          rescue StandardError => err
+            # In what cases does this help?
+            if err.cause && err.cause.message
+              Bolt::CLIError.new(err.cause.message)
```
adreyer: @MikaelSmith in what cases do we want to use the causes message? Its causing at least some errors to be squashed.

MikaelSmith: I tested some basic Puppet parsing errors, and errors thrown by Bolt functions. Specific errors I tried to unwrap were: `raise 'failure'`, `raise <Bolt::Error>`, and `raise <Ruby Error>`.

lib/bolt/cli.rb
```diff
@@ -516,23 +514,36 @@ def outputter
...
       r = Puppet::Pal.in_tmp_environment('bolt', modulepath: [BOLTLIB_PATH] + @config[:modulepath], facts: {}) do |pal|
         pal.with_script_compiler do |compiler|
           begin
             yield compiler
           rescue Puppet::PreformattedError => err
-            err.cause
+            # Puppet sometimes rescues exceptions notes the location and reraises
+            # For now return the original error.
+            if err.cause
+              if err.is_a? Bolt::Error
```
MikaelSmith: We throw Bolt::Errors that include a cause?

MikaelSmith: Actually, this is specifically catching `Puppet::PreformattedError`, how would that also be a `Bolt::Error`?

lib/bolt/cli.rb
```diff
@@ -13,11 +13,9 @@
 
 module Bolt
   class CLIError < Bolt::Error
-    attr_reader :error_code
-
-    def initialize(msg, error_code: 1)
+    def initialize(msg, error_code: nil)
```
MikaelSmith: I don't think we ever pass the `error_code` argument, could probably drop it.

lib/bolt/execution_result.rb
```diff
@@ -78,6 +78,23 @@ def values
       @result_hash.values
     end
 
+    # Expects to be called with a configured Puppet compiler or error.instance? will fail
```
MikaelSmith: Comment is no longer relevant.

lib/bolt/cli.rb
adreyer: oh thats a mistake. it's supposed to be err.cause.is_a?
We do throw Bolt::Errors with causes though

MikaelSmith: Is the line below also supposed to be `err.cause`?

lib/bolt/execution_result.rb
adreyer: I think it still suffers from the same problem since the execution result outlives the compiler. I intend to clean it up some when this class goes away.

modules/boltlib/lib/puppet/functions/run_command.rb
```diff
@@ -5,18 +5,22 @@
...
+
 Puppet::Functions.create_function(:run_command) do
   local_types do
     type 'TargetOrTargets = Variant[String[1], Target, Array[TargetOrTargets]]'
   end
 
   dispatch :run_command do
     param 'String[1]', :command
-    repeated_param 'TargetOrTargets', :targets
+    param 'TargetOrTargets', :targets
```
MikaelSmith: Ok, these changes all are more consistent with run_task. That makes sense.

adreyer: yeah that was my thinking. They were kind of forced since we can't accept both splatted options and targets.

modules/boltlib/lib/puppet/functions/run_script.rb
```diff
@@ -10,26 +10,16 @@
...
-    param 'Struct[arguments => Array[String]]', :arguments
-    return_type 'ExecutionResult'
-  end
-
   dispatch :run_script do
     scope_param
     param 'String[1]', :script
-    repeated_param 'TargetOrTargets', :targets
+    param 'TargetOrTargets', :targets
+    optional_param 'Hash[String[1], Any]', :options
```
MikaelSmith: It's not clear to me if this changes behavior. Before you needed to do
`run_script('script', $targets, arguments => ['-a', '-b'])`

The same would still work then, because you look for `arguments` in the options hash.

PR #248, opened 2018-01-17
(BOLT-303) Speed up tasks/plans when using winrm from Windows host

MikaelSmith:
An issue with OpenSSL on Windows results in a very slow call to
`random_bytes` unless done very early in the program. Make sure we
initialize early to avoid this issue if we're doing anything complicated
on Windows (like using Puppet).

### DIFF
--- a/lib/bolt/cli.rb
+++ b/lib/bolt/cli.rb
@@ -390,6 +390,14 @@ HELP
     def execute(options)
       if options[:mode] == 'plan' || options[:mode] == 'task'
         begin
+          if Gem.win_platform?
+            # Windows 'fix' for openssl behaving strangely. Prevents very slow operation
+            # of random_bytes later when establishing winrm connections from a Windows host.
+            # See https://github.com/rails/rails/issues/25805 for background.
+            require 'openssl'
+            OpenSSL::Random.random_bytes(1)
+          end
+
           require_relative '../../vendored/require_vendored'
         rescue LoadError
           raise Bolt::CLIError, "Puppet must be installed to execute tasks"


MikaelSmith:
It would be nice to understand this better, but it doesn't feel like an important thing to dig into right now.

glennsarti:
Some initial testing

Running `bolt task` connect time dropped alot

Connection times in seconds:
```
Task
---
16.0 - 15.3 - 15.2

to

3.9 - 3.9 - 3.8
```

lib/bolt/cli.rb
```diff
@@ -390,6 +390,14 @@ def handle_parser_errors
     def execute(options)
       if options[:mode] == 'plan' || options[:mode] == 'task'
         begin
+          if Gem.win_platform?
+            # Windows 'fix' for openssl behaving strangely. Prevents very slow operation
+            # of random_bytes later when establishing winrm connections from a Windows host.
+            # See https://github.com/rails/rails/issues/25805 for background.
+            require 'openssl'
```
glennsarti: Given this will 'pause' bolt for a few seconds on older rubies, perhaps emit a debug message saying "Warming up OpenSSL" or something to that effect

lib/bolt/cli.rb
```diff
@@ -390,6 +390,14 @@ def handle_parser_errors
     def execute(options)
       if options[:mode] == 'plan' || options[:mode] == 'task'
         begin
+          if Gem.win_platform?
```
glennsarti: I also tried this with `bolt command` by moving this to the beginning of the method.

resulted in speed up as well

Connection times in seconds
```
Command
----
8.9 - 8.9 - 9.0

to

3.9 - 4.4 - 3.8
```

lib/bolt/cli.rb
MikaelSmith: Why do you say it’ll take a few seconds on older Ruby? My finding was that the time to do this was negligible, and made later calls much faster.

glennsarti: Ruby 2.3.1

Without this PR
bolt-without

Note the runtime of ~10s

With this PR (But I've enabled it for `bolt command` just for this test)
bolt-with

Note the runtime is now ~5s BUT it takes about 4-5 seconds for any output to be shown.  This is in debug mode too.

MikaelSmith: I'm not able to reproduce that... I'm using Windows 10 Enterprise (x64), Ruby 2.3.3 from RubyInstaller. Where are you getting Ruby 2.3.1 from?

I'm running in Powershell and using Measure-Command. I get no increase runtime from this change. The command also runs in 5 seconds rather than 10 for me.

glennsarti: * Ruby 2.3.1 is from https://rubyinstaller.org/downloads/archives/

* Using PowerShell measure-command will of course show no change.  In my examples above the total runtime is always ~10s.  This PR shifts the delay earlier but the delay has to be incurred somewhere

* The VM I'm connecting to is a default-ish Windows installation with default WinRM settings.  Pooler has allowunecrypted and basic auth enabled.  When using bolt (with this PR) across the VPN into pooler, I was getting runtimes of 4.0 seconds.  Compared to pooler-to-pooler connection of 3.5-4.0 seconds. So that's reasonable.

* There are differences in my laptop and pooler

Laptop:
Intel64 Family 6 Model 60 Stepping 3 GenuineIntel ~2201 Mhz (Intel Speed Step will change the speed as well varied between 2.0 and 2.9GHZ)
8GB RAM
Windows 10 1709 (Fall Creators Update)

Pooler:
Intel64 Family 6 Model 62 Stepping 4 GenuineIntel ~2800 Mhz
4GB RAM
Windows 10 1703 (Creators Update)

Given that ruby is "mostly" single threaded, raw CPU speed will make a difference

---
But this doesn't change the UX of bolt seeming to do nothing for 3-5 seconds on startup.

MikaelSmith: I'm not able to reproduce this particular scenario. I've tried with Ruby 2.3.1 on a win-10-ent-x86_64 vmpooler image as well as a Windows 10 1709 image in VMWare Fusion on my laptop. In both cases I don't see any particularly slow behavior with `bolt command` with vanilla Bolt 0.14.0 and after adding this PR.

The only thing remotely similar I saw was that, after a fresh start of Windows 10, the first invocation of Bolt was slightly slower to start logging. Subsequent runs were normal.

glennsarti:
Note - The PR is mergable in its current state.  This is more a UX nice-ity

lib/bolt/cli.rb
nicklewis: Should this be `!!File::ALT_SEPARATOR` per https://github.com/puppetlabs/bolt/pull/241#discussion_r162177318?

MikaelSmith: I guess so? I don't understand @Iristyle's comment because Gem is shipped with Ruby now.

nicklewis: Hmm good point, this is probably fine then?

adreyer: it's also the logic that net:ssh is using

Iristyle: For Windows checking I prefer to not rely on anything external / versioned outside of core Ruby, especially when that code is designed to be mocked, when there's not a clear reason for it to be.

`FILE::ALT_SEPARATOR` is only compiled into Windows versions of Ruby and makes for a fast, simple and well understood (albeit a bit cryptic) check.

`Gem.win_platform?` at https://github.com/rubygems/rubygems/blob/master/lib/rubygems.rb#L1087-L1094 has indeed been around for a long time - originally added in 2007 at  https://github.com/rubygems/rubygems/commit/776d791c55e6d033bc7b450186821edbe4add579 ... but it's a bit more complex, and is also in place for the sake of tests pretending they're on Windows when they're not.

Either should work here, I just prefer not to use helpers on `Gem` or `Bundler` for this.

PR #249, opened 2018-01-17
(BOLT-222) Finalize catch_errors option name

adreyer:


modules/boltlib/lib/puppet/functions/run_task.rb
```diff
@@ -31,7 +31,7 @@ def run_task(task_name, targets, task_args = nil)
     r = Bolt::ExecutionResult.from_bolt(
       run_task_raw(task_name, targets, task_args)
     )
-    if !r.ok && task_args && task_args['_abort'] != false
+    if !r.ok && !task_args['_catch_errors']
```
MikaelSmith: task_args could still be nil.

PR #250, opened 2018-01-17
(BOLT-209) Return exitcode 2 when some nodes fail an operation

nicklewis:
Previously, as long as an operation was started successfully, bolt would
always exit 0, regardless of whether it actually completed successfully.
This meant that it was impossible to tell whether a task or command had
actually run on the nodes in question, only whether the task exists
or whether the command was properly interpreted by bolt.

For tasks, scripts, comamnds and file uploads, we now exit 2 if any node
fails. The behavior for plans is unchanged, as we still don't have a
solid definition of a plan "result". It can now be assumed that if bolt
exits 0, the operation completed successfully on every node.

### DIFF
--- a/lib/bolt/cli.rb
+++ b/lib/bolt/cli.rb
@@ -430,6 +430,7 @@ HELP
       if options[:mode] == 'plan'
         executor = Bolt::Executor.new(@config, options[:noop], true)
         execute_plan(executor, options)
+        code = 0
       else
         executor = Bolt::Executor.new(@config, options[:noop])
         targets = options[:nodes]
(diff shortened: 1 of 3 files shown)


lib/bolt/cli.rb
```diff
@@ -471,7 +472,10 @@ def execute(options)
         end
 
         outputter.print_summary(results, elapsed_time)
+        successful = results.values.all?(&:success?)
```
MikaelSmith: Why wasn't `results.ok` sufficient here?

MikaelSmith: Oh, because it's a hash, not an execution result yet. It might help to highlight what we expect `results` to be when we get here.

nicklewis: I expect that [BOLT-286](https://tickets.puppetlabs.com/browse/BOLT-286) will have an impact on this.

PR #251, opened 2018-01-18
(BOLT-290) decompose --insecure flag 

zreichert:


zreichert:
This will still need integration and acceptance test updates.

lib/bolt/cli.rb
```diff
@@ -181,9 +181,13 @@ def create_option_parser(results)
                 "Output format to use: human or json") do |format|
           results[:format] = format
         end
-        opts.on('-k', '--insecure',
-                "Whether to connect insecurely ") do |insecure|
-          results[:insecure] = insecure
+        opts.on('--[no-]host-key-check',
```
MikaelSmith: `--nodes` help still mentions `--insecure`. We should make it clear that we use `host-key-check` and `ssl` by default.

MikaelSmith: The README also mentions `insecure`, should be updated to reflect new options.

zreichert: Fixed

lib/bolt/cli.rb
```diff
@@ -181,9 +181,13 @@ def create_option_parser(results)
...
+                "Check host keys with SSH") do |host_key_check|
+          results[:host_key_check] = host_key_check
+        end
+        opts.on('--[no-]ssl',
+                "Use ssl with WINRM") do |ssl|
```
nicklewis: `SSL` should be capitalized, and it should be `WinRM`

lib/bolt/config.rb
```diff
@@ -18,17 +18,24 @@ module Bolt
...
                            token_file orch_task_environment service_url].freeze
 
     TRANSPORT_DEFAULTS = {
       connect_timeout: 10,
       orch_task_environment: 'production',
-      insecure: false,
       tty: false
     }.freeze
 
+    SHH_DEFAULTS = {
```
nicklewis: Should be `SSH_DEFAULTS`

nicklewis: This should maybe be refactored to something like:

```rb
TRANSPORT_SPECIFIC_DEFAULTS = {
  ssh: {
    host_key_check: true
  },
  winrm: {
    ssl: true
  }
}
```

Then the code below doesn't need to have an explicit switch based on transport.

lib/bolt/config.rb
```diff
@@ -167,9 +188,20 @@ def update_from_cli(options)
...
 
       TRANSPORT_OPTIONS.each do |key|
-        # TODO: We should eventually make these transport specific
         TRANSPORTS.each do |transport|
-          self[:transports][transport][key] = options[key] if options[key]
+          unless %i[ssl host_key_check].any? { |k| k == key }
+            self[:transports][transport][key] = options[key] if options[key]
+            next
+          end
+          if key == :ssl && transport == :winrm
```
nicklewis: I don't fully understand what this section is doing, and why we weren't already doing something like this for options like `:key` or `:cacert`?

MikaelSmith: We've generally been passing through options, even if they're unused by the transport. Eventually we'll want to refactor this a bit so transports can register their own expected options.

lib/bolt/cli.rb
zreichert: Fixed

lib/bolt/config.rb
zreichert: @nicklewis I'm not sure if what I did addresses your concern or not.  I think this could generally use a refactor though.  If my newest change works for you I'll fixup the last commit.

lib/bolt/config.rb
```diff
@@ -46,6 +54,14 @@ def initialize(**kwargs)
             self[:transports][transport][k] = v
           end
         end
+
+        TRANSPORT_SPECIFIC_DEFAULTS.each do |k,v|
```
nicklewis: This should be `TRANSPORT_SPECIFIC_DEFAULTS[transport].each`

zreichert: updated

lib/bolt/config.rb
```diff
@@ -100,8 +115,8 @@ def update_from_file(data)
         if data['ssh']['private-key']
           self[:transports][:ssh][:key] = data['ssh']['private-key']
         end
-        if data['ssh']['insecure']
-          self[:transports][:ssh][:insecure] = data['ssh']['insecure']
+        if data['ssh'].key?('host_key_check')
```
nicklewis: This should be hyphenated rather than underscored, to match `private-key`.

zreichert: fixed

README.md
```diff
@@ -57,7 +57,7 @@ modulepath: "~/.puppetlabs/bolt-code/site:~/.puppetlabs/bolt-code/modules"
 concurrency: 10
 format: human
 ssh:
-  insecure: true
+  host_key_check: false
```
nicklewis: This also should use hyphens.

zreichert: fixed

nicklewis:
Looks good to me. Did you want to squash?

spec/integration/ssh_spec.rb
```diff
@@ -56,7 +58,7 @@
       { 'format' => 'json',
         'modulepath' => modulepath,
         'ssh' => {
-          'insecure' => true
+          'host_key_check' => false
```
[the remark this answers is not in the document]
zreichert: Does this test not get run by Travis?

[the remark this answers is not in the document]
nicklewis: It should... It may be set up in such a way that it doesn't cause a host key issue in travis?

PR #252, opened 2018-01-19
(BOLT-311) Provide Ruby 2.0 fallback for exception causes

MikaelSmith:
Exception causes were added in Ruby 2.1. We want to continue using them.
Provide a fallback to inferior error handling for Ruby 2.0.


MikaelSmith:
Where should the known issue go? The README?

nicklewis:
We have a [known issues page](https://puppet.com/docs/bolt/0.x/bolt_known_issues.html)

PR #253, opened 2018-01-19
(BOLT-286) Convert ExecutionResult to ResultSet

adreyer:
This standardizes how results are handled in bolt and puppet. Previously
a hash Bolt::Result objects were generated by they executor when a
run_* action was taken from puppet. The puppet run_* functions then
convverted those into and ExecutionResult. Now Results are built into a
Bolt::ResultSet object in the executor. Both this and the bolt result
are now pcore datatypes and available directly in the plan.


spec/bolt/result_spec.rb
```diff
@@ -25,74 +25,49 @@
...
-    expected = { 'value' => { 'key' => 'val' },
-                 'error' => { 'msg' => 'oops' } }
-    result = Bolt::TaskResult.new(target, obj.to_json, '', 1)
-    expect(result.to_h).to eq(expected)
-  end
+  describe :for_task do
+    it 'parses json objects' do
+      obj = { "key" => "val" }
+      result = Bolt::Result.for_task(target, obj.to_json, '', 0)
+      expect(result.to_result).to eq(obj)
```
MikaelSmith: Should this be `generic_value`?

lib/bolt/result.rb
```diff
@@ -18,124 +18,122 @@ def self.from_exception(target, exception)
...
-  class TaskResult < CommandResult
-    attr_reader :value
 
-    def initialize(target, stdout, stderr, exit_code)
-      super(target, stdout, stderr, exit_code)
-      @value = parse_output(stdout)
-      @message = @value.delete('_output')
-      @error = @value.delete('_error')
+    # TODO: remove in favor of ok?
+    def success?
```
MikaelSmith: Having only one (`ok?`) seems fine.

lib/bolt/result.rb
```diff
@@ -18,124 +18,122 @@ def self.from_exception(target, exception)
...
-        unless obj.is_a? Hash
-          obj = nil
-        end
-      rescue JSON::ParserError
-        obj = nil
-      end
-      obj || { '_output' => output }
+    # Warning: This will fail outside of a compilation.
+    # Use error_hash inside bolt.
+    # Is it crazy for this to behave differently outside a compiler?
```
MikaelSmith: It seems necessary, somewhere.

lib/bolt/cli.rb
```diff
@@ -399,6 +399,9 @@ def execute(options)
           raise Bolt::CLIError, "Puppet must be installed to execute tasks"
         end
 
+        # Now that puppet is loaded we can include puppet mixins in data types
+        Bolt::ResultSet.include_iterable
```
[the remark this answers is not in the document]
MikaelSmith: Ah, this is preserving previous behavior. There's fallback for iterable when not using Puppet.

lib/bolt/result_set.rb
```diff
@@ -0,0 +1,115 @@
...
+  def names
+    @results.map { |r| r.target.name }
+  end
+
+  def ok
+    @results.all? { |r| r.ok? }
+  end
+  alias ok? ok
+
+  def errorset
```
MikaelSmith: This feels like a minor deviation from naming conventions. Would have expected `error_set` and `ok_set`.

lib/bolt/result_set.rb
```diff
@@ -0,0 +1,115 @@
...
+  def find(target_name)
+    result_hash[target_name]
+  end
+
+  def first
+    @results.first
+  end
+
+  # TODO: should this exist?
+  def [](node_uri)
```
MikaelSmith: Asking whether you should be able to look a value up by its name/uri? I feel like the one case that could be useful is when cross-referencing between several results.  Like: several nodes performed an action against load-balanced servers, I want to group the results of that action by the server they operated against.

I feel like I'm having to stretch for examples, but it doesn't feel like much work to enable some use-cases we can't envision right now.

MikaelSmith: I guess `find` already allows that. Are you suggesting that as better than `[]`?

lib/bolt/result_set.rb
```diff
@@ -0,0 +1,115 @@
...
+    find[node_uri]
+  end
+
+  # TODO: should this exist?
+  def value(node_uri)
+    self[node_uri]
+  end
+
+  # TODO delete me?
```
MikaelSmith: Anything we don't use and that's not exposed in the Puppet DataType seems like we can remove.

lib/bolt/result_set.rb
```diff
@@ -0,0 +1,115 @@
...
+  def values
+    @results
+  end
+
+  def eql?(other)
+    self.class == other.class && @results == other.results
+  end
+
+  # TODO: this loses target information
```
MikaelSmith: Target data seems like something we'd want when serializing, but I'm not sure what we're using the serialization for.

lib/bolt/target.rb
```diff
@@ -60,6 +60,12 @@ def port
       @uri_obj.port
     end
 
+    # name is currently just uri but should be be used instead to identify the
+    # Target ouside the transport or uri options.
```
MikaelSmith: That seems like hostname...

modules/boltlib/lib/puppet/functions/run_task.rb
```diff
@@ -16,22 +16,21 @@
     param 'String[1]', :task_name
     param 'TargetOrTargets', :targets
     optional_param 'Hash[String[1], Any]', :task_args
-    return_type 'ExecutionResult'
+    return_type 'ResultSet'
   end
 
   # this is used from 'bolt task run'
   dispatch :run_task_raw do
```
MikaelSmith: What is `run_task_raw` used for? Not sure what we do with the block.

lib/bolt/cli.rb
```diff
@@ -599,11 +604,7 @@ def run_plan(plan, args)
         end
         in_bolt_compiler(cli) do |compiler|
           result = compiler.call_function('run_plan', plan, args)
-          # Querying ExecutionResult for failures currently requires a script compiler.
-          # Convert from an ExecutionResult to structured output that we can print.
-          if result.instance_of? Bolt::ExecutionResult
-            result = result.unwrap
-          end
+          # TODO: Is there anything we still need to unwrap?
```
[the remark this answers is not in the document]
adreyer: We should be displaying the error_hash not the error datatype in bolt. I may not have fixed that properly

lib/bolt/executor.rb
```diff
@@ -57,22 +58,21 @@ def on(nodes, callback = nil)
...
-          result
         end
       }
       pool.shutdown
       pool.wait_for_termination
 
       @notifier.shutdown
 
-      results_to_hash(results)
+      Bolt::ResultSet.new( results.map { |r| r } )
```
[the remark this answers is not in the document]
adreyer: I'm not sure. it was doing this before with the hash.

lib/bolt/cli.rb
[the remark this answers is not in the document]
MikaelSmith: I guess that will be a ResultSet no matter what, which uses status_hash?

lib/bolt/result_set.rb
```diff
@@ -0,0 +1,94 @@
...
+
+    def first
+      @results.first
+    end
+
+    def eql?(other)
+      self.class == other.class && @results == other.results
+    end
+
+    def to_json
```
MikaelSmith: Getting an error
```
Plan aborted: run_task 'upgrade_pe::monolithic' failed on 1 nodes
bundler: failed to load command: bolt (/Users/michaelsmith/puppetlabs/orchestration-control/.bundle/gems/ruby/2.3.0/bin/bolt)
ArgumentError: wrong number of arguments (given 1, expected 0)
  /Users/michaelsmith/puppetlabs/orchestration-control/.bundle/gems/ruby/2.3.0/bundler/gems/bolt-f260b52f88e9/lib/bolt/result_set.rb:82:in `to_json'
  /Users/michaelsmith/.rbenv/versions/2.3.4/lib/ruby/2.3.0/json/common.rb:286:in `generate'
  /Users/michaelsmith/.rbenv/versions/2.3.4/lib/ruby/2.3.0/json/common.rb:286:in `pretty_generate'
  /Users/michaelsmith/puppetlabs/orchestration-control/.bundle/gems/ruby/2.3.0/bundler/gems/bolt-f260b52f88e9/lib/bolt/outputter/human.rb:142:in `fatal_error'
```

Plan I'm trying to run is
```
plan upgrade_pe::mono_cms(
  String[1]                                      $mono_master,
  String[1]                                      $build,
  Optional[Variant[String[1], Array[String[1]]]] $compile_masters = undef,
) {
  $mono = run_task('upgrade_pe::monolithic', $mono_master, build => $build)
  $mono.each |$result| {
    if $result.ok {
      notice("${result.target} completed:\n${result.message}")
    } else {
      notice("${result.target} errored with message ${result.error}")
    }
  }

  if ($mono.ok and $compile_masters) {
    $_compile_masters = $compile_masters.split(',')
    $cms = run_task('upgrade_pe::compile_master', $_compile_masters, master => $mono_master)
    $cms.each |$result| {
      if $result.ok {
        notice("${result.target} completed:\n${result.message}")
      } else {
        notice("${result.target} errored with message ${result.error}")
      }
    }
  }
}
```

adreyer: this was caused by to_json not accepting opts. it should be fixed now

spec/pal/result_set_spec.rb
```diff
@@ -0,0 +1,96 @@
+require 'spec_helper'
+require 'bolt_spec/files'
+require 'bolt_spec/pal'
+
+require 'bolt/pal'
+# TODO: clean this up
```
MikaelSmith: What's this TODO?

adreyer: the requires. This should get cleaned up with the PAL work

PR #256, opened 2018-01-22
(BOLT-286) Update acceptance tests for ResultSet

adreyer:


acceptance/tests/plan_winrm.rb
```diff
@@ -51,8 +51,8 @@
     message = "The plan was expected to notify but did not"
     assert_match(/Notice:/, result.stdout, message)
     winrm_nodes.each do |node|
-      message = "The plan was expceted to mention the host #{node.hostname} with _output"
-      assert_match(/#{node.hostname}"=>{"_output"=>/, result.stdout, message)
+      message = "The plan was expected to mention the host #{node.hostname} with _output"
+      assert_match(/#{node.hostname}","status":"success"/, result.stdout, message)
```
MikaelSmith: Why does this check differ from `plan_ssh.rb`?

adreyer: because there is a protocol on the url in this case

PR #257, opened 2018-01-22
(MAINT) remove ci namespace from acceptance rakefile

zreichert:
Since CI doesn't actually use the CI namespace I removed it from the Rakefile.


zreichert:
@MikaelSmith Updated

PR #259, opened 2018-01-23
(BOLT-292) Add support for _run_as option to run functions

MikaelSmith:
Adds support for passing `_run_as => 'foo'` to `run_command`, `run_task`, `run_script`, and `run_plan` as part of a plan to tell them to run as a different user. Only applies to the SSH transport.

When specified on `run_plan`, it overrides any config settings for the duration of that plan and applies to all run function calls during that plan.

When specified on other run functions, overrides any config setting and prior plan setting when running that command/script/task on the targets.

### DIFF
--- a/lib/bolt/node.rb
+++ b/lib/bolt/node.rb
@@ -49,6 +49,22 @@ module Bolt
     def uri
       @target.uri
     end
+
+    def upload(_source, _destination)
+      raise NotImplementedError, 'transports must implement upload(source, destination)'
+    end
+
+    def run_command(_command, _options = nil)
+      raise NotImplementedError, 'transports must implement run_command(command, options = nil)'
+    end
+
+    def run_script(_script, _arguments, _options = nil)
+      raise NotImplementedError, 'transports must implement run_script(script, arguments, options = nil)'
+    end
+
+    def run_task(_task, _input_method, _arguments, _options = nil)
+      raise NotImplementedError, 'transports must implement run_task(task, input_method, arguments, options = nil)'
+    end
   end
 end
 
--- a/lib/bolt/node.rb
+++ b/lib/bolt/node.rb
@@ -22,7 +22,7 @@ module Bolt
 
     def self.initialize_transport(_logger); end
 
-    attr_reader :logger, :user, :password, :connect_timeout, :target
+    attr_reader :logger, :user, :password, :connect_timeout, :target, :run_as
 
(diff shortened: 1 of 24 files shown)


MikaelSmith:
I've done some testing and believe this works. Working on adding tests now.

lib/bolt/node.rb
```diff
@@ -49,6 +49,22 @@ def initialize(target, config: Bolt::Config.new)
     def uri
       @target.uri
     end
+
+    def upload(_source, _destination)
+      raise 'Not implemented'
```
nicklewis: Should these raise a `NotImplementedError`?

lib/bolt/node/ssh.rb
```diff
@@ -253,16 +258,20 @@ def with_remote_task(task_file, stdin)
       end
     end
 
-    def run_command(command)
+    def run_command(command, options = {})
+      @run_as = options['_run_as'] || @conf_run_as
```
nicklewis: Should we just pass this value through `execute` (which already takes an `options` hash) rather than clobbering the instance variable?

MikaelSmith: It's all the extra things that use execute - with_remote_task, with_task_wrapper, with_remote_tempdir - that made that messy.

lib/bolt/node.rb
MikaelSmith: I'll do that, I was being lazy and it was mainly for documentation of function signature.

adreyer: We're getting rid of the base class soon anyway...

lib/bolt/node/ssh.rb
nicklewis: Ah that's fair. Hopefully the transport API changes will simplify some of this.

adreyer: this will be moot soon as well since the instance var won't exist in transports

nicklewis:
Running tasks and scripts with `_run_as` (or with `--run-as` from the CLI, for that matter) fails with permission denied. I assume that's because of [BOLT-218](https://tickets.puppetlabs.com/browse/BOLT-218)?

MikaelSmith:
Might be. I'm generally testing <user> -> root, were you trying the other way?

nicklewis:
Yeah, I was connecting as root and running as a different user.

PR #260, opened 2018-01-23
(BOLT-286) Mark some Result/ResultSet function return types Optional

nicklewis:
Currently, calling these functions under certain conditions can cause
Undef to be returned when the return type forbids it. That causes plans
using those functions in those ways to fail, even though they're doing
legitimate operations.

The cases covered are:
* calling error() on a Result which has no error
* calling message() on a Result which has no message
* calling first() on an empty ResultSet
* calling find() on a ResultSet with an unknown target name


nicklewis:
I'm not 100% sure whether it's correct to have `ResultSet#find` return undef for an unknown name.

adreyer:
It seems like the best option for now at least.

PR #261, opened 2018-01-24
(BOLT-286) Results and ResultSets should be displayed in json from plans

adreyer:
This favors to_json over to_s when displaying the result of a plan. As a
side effect this will cause strings to be quoted which we may want to
reconsider later.

### DIFF
--- a/lib/bolt/outputter/human.rb
+++ b/lib/bolt/outputter/human.rb
@@ -123,13 +123,15 @@ module Bolt
       end
 
       def print_plan(result)
-        # If a hash or array, pretty-print as JSON
-        if result.is_a?(Hash) || result.is_a?(Array)
-          if result.empty?
-            # Avoids extra lines for an empty result
+        # If the object has a json representation display it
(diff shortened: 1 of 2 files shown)


lib/bolt/outputter/human.rb
```diff
@@ -123,13 +123,15 @@ def print_task_info(task)
       end
 
       def print_plan(result)
-        # If a hash or array, pretty-print as JSON
-        if result.is_a?(Hash) || result.is_a?(Array)
-          if result.empty?
-            # Avoids extra lines for an empty result
+        # If the object has a json representation display it
+        if result.respond_to?(:to_json)
```
MikaelSmith: This will be true for almost everything, since the json library monkey-patches things.

We could special-case String. It's probably not important to address right now though.

adreyer: Yeah I was thinking long term we may want to have some convention like "_output" for the string summary of a plan.

PR #262, opened 2018-01-25
(BOLT-304) Add support for displaying plan parameters

mruzicka:
This PR adds support for displaying plan parameters info using:
```
bolt plan show <plan name>
```
A new function for printing the info - `print_plan_info` - is added to the outputter classes. The existing `print_plan` function is renamed to `print_plan_result` so as to distinguish it better from the newly added `print_plan_info`.

Appropriate tests are added for the new functionality.

Additionally Vagrant setup is updated to make the winrm over ssl tests pass when executing against the Vagrant environment. And since some of this setup is shared with AppVeyor, the `appveyor.yml` is updated too.


lib/bolt/outputter/human.rb
```diff
@@ -104,10 +104,10 @@ def print_task_info(task)
           task['parameters'].each do |k, v|
             pretty_params << "- #{k}: #{v['type']}\n"
             pretty_params << "    #{v['description']}\n" if v['description']
-            usage << if !v['type'].to_s.include? "Optional"
-                       " #{k}=<value>"
-                     else
+            usage << if v['type'].to_s.include? "Optional"
```
mruzicka: Is there a better way of testing for optional parameters?

lib/bolt/cli.rb
```diff
@@ -45,6 +46,7 @@ class CLI
 
 Available actions are:
     show                             Show list of available tasks
+    show <task>                      Show documentation for task
```
mruzicka: I'm not sure if this should be here or not.

lib/bolt/cli.rb
```diff
@@ -75,6 +77,7 @@ class CLI
 
 Available actions are:
     show                             Show list of available plans
+    show <plan>                      Show details for plan
```
mruzicka: I'm not sure if this should be here or not.

lib/bolt/cli.rb
mruzicka: I'm not sure if this should be here or not.

lib/bolt/cli.rb
mruzicka: I'm not sure if this should be here or not.

lib/bolt/outputter/human.rb
mruzicka: Is there a better way of testing for optional parameters?

adreyer: Using the compiler to test whether the undef is assignable to the type is more accurate than looking for Optional probably

lib/bolt/cli.rb
adreyer: I think this is better than a single line and definitely better than not mentioning it.

lib/bolt/outputter/human.rb
```diff
@@ -122,7 +122,34 @@ def print_task_info(task)
...
+        # Building lots of strings...
+        pretty_params = ""
+        plan_info = ""
+        usage = "bolt plan run #{plan['name']}"
+
+        if plan['parameters']
+          plan['parameters'].each do |p|
+            name, type = p.values_at('name', 'type')
+            pretty_params << "- #{name}: #{type}\n"
+            usage << if type.to_s.include? "Optional"
```
adreyer: Optional plan params are more complex that optional task params because a plan param can be optional either because it accepts undef or because it has a default.

Vagrantfile
```diff
@@ -3,9 +3,9 @@ echo "vagrant ALL=(ALL) ALL" > /etc/sudoers.d/vagrant
 SCRIPT
 
 windows_enable_winrm_ssl = <<SCRIPT
-$thumbprint = "B888D0548D394D9678131A36A4869E4C56F20E4F"
-Import-PfxCertificate -FilePath C:\\cert.pfx -CertStoreLocation cert:\\LocalMachine\\My -Password (ConvertTo-SecureString -String vagrant -Force -AsPlainText)
-New-WSManInstance -ResourceURI winrm/config/Listener -SelectorSet @{Address='*';Transport="HTTPS"} -ValueSet @{Hostname="localhost";CertificateThumbprint="$thumbprint"}
+$cert = Import-PfxCertificate -FilePath C:\\cert.pfx -CertStoreLocation cert:\\LocalMachine\\My -Password (ConvertTo-SecureString -String vagrant -Force -AsPlainText)
```
MikaelSmith: 💯

lib/bolt/outputter/human.rb
nicklewis: `Optional` is a little different for plans than for tasks. `Optional` means you can pass `undef` as a value, not that you don't need to provide a value. This is different from tasks, where we treat not providing a value and providing null as equivalent.

For this plan:

```puppet
plan foo(Optional[String] $param) {}
```

Trying to run it as `bolt plan run foo` will fail with `foo: expects a value for parameter 'param'`.

A parameter is *actually* optional if it has a default value.

```puppet
plan foo(String $param = 'foo') {}
```

This plan will run successfully with `bolt plan run foo`, since every parameter has a value.

nicklewis: Oh I just saw that this is for tasks and not for plans. So kindly pretend I left this comment below where it's actually relevant. :)

mruzicka:
I've updated the optional parameter logic as per @nicklewis's comment: The only optional parameters of a plan are those with default values specified.
I've also updated the check for the task's optional parameters to use:
```ruby
task_parameter_type.is_a?(Puppet::Pops::Types::POptionalType)
```

PR #263, opened 2018-01-25
(maint) Use Bolt::PAL for interactions with puppet

adreyer:
Previously the PAL class was only used for tests and was duplicated in
CLI. This converts CLI to interact with Puppet through the PAL clas
instead.

### DIFF
--- a/lib/bolt/cli.rb
+++ b/lib/bolt/cli.rb
@@ -11,6 +11,7 @@ require 'bolt/executor'
 require 'bolt/target'
 require 'bolt/outputter'
 require 'bolt/config'
+require 'bolt/pal'
 require 'io/console'
 
 module Bolt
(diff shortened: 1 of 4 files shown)


lib/bolt/pal.rb
```diff
@@ -1,15 +1,27 @@
...
     BOLTLIB_PATH = File.join(__FILE__, '../../../modules')
 
     def initialize(config)
-      # TODO: how should we manage state? Does noop go here?
+      # Nothing works without initialized this global state. Reinitializing
+      # is safe and in practice only happen in tests
+      self.class.load_puppet
+      self.class.configure_logging(config[:log_level])
+
+      # TODO: how should we manage executor state?
```
MikaelSmith: Not sure what this means.

MikaelSmith: Why not pass the executor in to run_task/run_plan? Seems as straight-forward as passing `noop` through.

spec/spec_helper.rb
```diff
@@ -4,6 +4,7 @@
 require 'bolt/logger'
 require 'logging'
 require 'rspec/logging_helper'
+require_relative '../vendored/require_vendored'
```
MikaelSmith: Why is this needed? Doesn't load_puppet for particular tests cover it?

MikaelSmith: Or... why did this work before?

adreyer: we may try to reset puppet's settings before any test has called load_puppet.

adreyer: I'm not sure how it worked before

lib/bolt/cli.rb
```diff
@@ -394,44 +395,33 @@ def handle_parser_errors
...
 
       if options[:mode] == 'plan'
         executor = Bolt::Executor.new(@config, options[:noop], true)
-        execute_plan(executor, options)
+        result = pal.run_plan(options[:object], options[:task_options], executor)
+        outputter.print_plan(result)
+        # An exception would have been raised if the plan failed
         code = 0
       else
+        # This executor isn't used for tasks
```
MikaelSmith: Comment is no longer correct.

spec/bolt/pal_spec.rb
```diff
@@ -0,0 +1,2 @@
+# Currently PAL is tested through the CLI execute method. New tests should be
+# added here as we add features  or fix bugs.
```
MikaelSmith: Extra space you could remove.

MikaelSmith:
Needs a rebase now.

PR #265, opened 2018-01-29
(BOLT-194) Gracefully handle Ctrl-C

mruzicka:
Adds code which installs a SIGINT handler just before any command/script/task/plan is executed and removes it afterwards.
When invoked, the handler logs an appropriate message at the info level and then makes bolt exit promptly possibly leaving behind processes running on nodes.

Note that this does not prevent bolt from printing a stacktrace when it receives SIGINT while not executing any command/script/task/plan.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -11,16 +11,21 @@ SCRIPT
 Vagrant.configure("2") do |config|
   config.vm.define :windows do |windows|
     windows.vm.box = "mwrock/WindowsNano"
-    windows.vm.network :forwarded_port, guest: 22, host: 2222, id: "ssh"
     windows.vm.guest = :windows
+    windows.vm.communicator = "winrm"
+    windows.vm.network :forwarded_port, guest: 22, host: 2222, id: "ssh", disabled: true
+    windows.vm.network :forwarded_port, guest: 5985, host: 25985, host_ip: "127.0.0.1", id: "winrm"
+    windows.vm.network :forwarded_port, guest: 5986, host: 25986, host_ip: "127.0.0.1", id: "winrm-ssl"
     windows.vm.provision "file", source: "resources/cert.pfx", destination: 'C:\cert.pfx'
     windows.vm.provision "shell", privileged: true, inline: windows_enable_winrm_ssl
-    windows.vm.communicator = "winrm"
+    windows.vm.provider "virtualbox" do |vb|
+      vb.gui = false
+    end
   end
 
   config.vm.define :linux do |linux|
     linux.vm.box = "bento/centos-6.7"
-    linux.vm.network :forwarded_port, guest: 22, host: 2224, id: "ssh"
+    linux.vm.network :forwarded_port, guest: 22, host: 20022, host_ip: "127.0.0.1", id: "ssh"
     linux.vm.provision "shell", inline: linux_provision
   end
 end
--- a/lib/bolt/cli.rb
+++ b/lib/bolt/cli.rb
@@ -400,6 +400,15 @@ HELP
     end
 
     def execute(options)
+      message = nil
+
+      handler = Signal.trap :INT do |signo|
+        @logger.info(
+          "Exiting after receiving SIG#{Signal.signame(signo)} signal." << (message ? ' ' << message : '')
+        )
+        exit!
+      end
+
       if options[:mode] == 'plan' || options[:mode] == 'task'
         pal = Bolt::PAL.new(@config)
       end
@@ -426,6 +435,7 @@ HELP
       end
 
       inventory = Bolt::Inventory.from_config(@config)
+      message = 'There may be processes left executing on some nodes.'
 
       if options[:mode] == 'plan'
         executor = Bolt::Executor.new(@config, options[:noop], true)
@@ -484,6 +494,9 @@ HELP
     rescue Bolt::Error => e
       outputter.fatal_error(e)
       raise e
+    ensure
+      # restore original signal handler
+      Signal.trap :INT, handler if handler
     end
 
     def validate_file(type, path)
--- a/spec/bolt/executor_spec.rb
+++ b/spec/bolt/executor_spec.rb
@@ -34,7 +34,7 @@ describe "Bolt::Executor" do
   let(:node_results) { mock_node_results }
 
   before(:each) do
-    allow(executor).to receive(:from_targets).with(targets).and_return(node_results.map(&:first))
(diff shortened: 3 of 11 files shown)


lib/bolt/executor.rb
```diff
@@ -40,6 +40,14 @@ def on(nodes, callback = nil)
       pool = Concurrent::FixedThreadPool.new(poolsize)
       @logger.debug { "Started with #{poolsize} thread(s)" }
 
+      handler = Signal.trap :INT do |signo|
+        @logger.info([
+          "Exiting after receiving SIG#{Signal.signame(signo)} signal.",
+          'There may be processes left executing on some nodes.'
```
mruzicka: There is quite likely a better wording for this message.

spec/bolt/executor_spec.rb
```diff
@@ -309,6 +309,28 @@ def mock_node_results(run_as = nil)
     end
   end
 
+  it "traps SIGINT" do
+    skip "hangs on Windows" if Gem.win_platform?
```
mruzicka: This test hangs at least in AppVeyor. I'm going to check it outside of AppVeyor tomorrow.

adreyer:
The handler won't work in here? https://github.com/puppetlabs/bolt/blob/master/lib/bolt/pal.rb#L50

It wouldn't surprise but I just want to make sure we tried.

spec/bolt/executor_spec.rb
mruzicka: The test does work on Windows but when the rspec is executed from a batch script, the batch script prints _Terminate batch job (Y/N)?_ message and waits for user input when the rspec finishes, in effect it hangs.
The problem is that when the SIGINT is sent it is delivered not only to the ruby process executing the rspec tests but rather to the entire process group including the batch script from which the ruby process is executed and that's why the script prints the message and waits for the user input and in effect hangs.

mruzicka:
It turns out it works even in the bolt/cli.rb, so I'm moving the signal handler there.

lib/bolt/executor.rb
```diff
@@ -69,6 +77,9 @@ def on(nodes, callback = nil)
 
       @notifier.shutdown
 
+      # restore original handler
+      Signal.trap :INT, handler
```
joshcooper: If you're trying to use bolt as a library, it seems odd to have signal handling here. I'd expect it to be handled by the application that is calling bolt (library) or the bolt cli itself.

mruzicka: Yes, that's where I'm moving the signal handler registration.
I was originally convinced that registering the handler there has no effect, but that's not the case.

nicklewis:
This message only appears with `--verbose`. The ticket does specify info level, but I wanted to double-check that it's indeed intended to only be a verbose message. @adreyer, @joshcooper (since you filed the initial ticket) Thoughts?

joshcooper:
I don't have an opinion about the log level. I'm more concerned about gracefully shutting down the thread pool, eg. do we need to call `pool.kill`?

lib/bolt/cli.rb
```diff
@@ -426,58 +426,72 @@ def execute(options)
...
-              pal.run_task(options[:object],
-                           targets,
-                           options[:task_options],
-                           executor) do |event|
-                outputter.print_event(event)
-              end
-            when 'file'
-              src = options[:object]
-              dest = options[:leftovers].first
+      handler = Signal.trap :INT do |signo|
```
adreyer: This should happen before puppet is loaded probably since that takes so long.

lib/bolt/cli.rb
```diff
@@ -426,58 +426,72 @@ def execute(options)
...
-            end
-        end
+          end
 
-        outputter.print_summary(results, elapsed_time)
-        code = results.ok ? 0 : 2
+          outputter.print_summary(results, elapsed_time)
+          code = results.ok ? 0 : 2
+        end
+      ensure
```
adreyer: I think this can just be an ensure on the method
```
ensure
  Signal.trap :INT, 'DEFAULT'
```

Vagrantfile
```diff
@@ -11,16 +11,18 @@ SCRIPT
...
     windows.vm.provision "shell", privileged: true, inline: windows_enable_winrm_ssl
-    windows.vm.communicator = "winrm"
+    windows.vm.provider "virtualbox" do |vb|
+      vb.gui = false
+    end
   end
 
   config.vm.define :linux do |linux|
     linux.vm.box = "bento/centos-6.7"
-    linux.vm.network :forwarded_port, guest: 22, host: 2224, id: "ssh"
```
adreyer: This will break if there are other VMs running. We should pick a less common value than 2222 and set it explicitly.

mruzicka: True, but the same is true for the winrm & winrm-ssl, should we pick some less common ports values for those too?

lib/bolt/cli.rb
nicklewis: What will that do when running tests through rspec which sets its own signal handlers?

nicklewis: I guess that question already sort of applies regardless, since we're setting up a signal handler that calls `exit!`...

lib/bolt/cli.rb
mruzicka: ok, I'll need to vary the message depending on when exactly the SIGINT is delivered though

lib/bolt/cli.rb
mruzicka: yeah, that should work

mruzicka:
If we want to perform a graceful shutdown, we need to be prepared to accept that it may take some time. And in that case we even might be better off rescuing the ruby's SignalException rather than using a signal handler.
@adreyer @joshcooper @nicklewis thoughts?

mruzicka:
@joshcooper If the concern is a graceful shutdown, then we can drop the signal handler and instead add a rescue clause like this:
```ruby
rescue SignalException
  pool.shutdown
  ...
end
```
at an appropriate place.
If the concern is rather making sure the pooler threads are stopped, then the current solution is ok, as all threads in ruby are killed when the main thread exits.

Vagrantfile
mruzicka: I've listed all the required port forwardings explicitly and without the `auto_correct` options, i.e. they are now static. Either the required ports are available or the provisioning fails.

adreyer:
I'm still getting backtraces when I try to kill a task run. See my comment on the ticket

mruzicka:
Rebased on latest master.

PR #267, opened 2018-01-29
(BOLT-104) Add inventory file support to bolt

adreyer:


lib/bolt/target.rb
```diff
@@ -21,7 +21,34 @@ def self.parse_urls(urls)
...
+      @password ||= t_conf[:password]
+      @port ||= t_conf[:port]
+
+      @options = t_conf.reject { |k, _| url_keys.include?(k) }.merge(@options)
+      @configured = true
+
+      self
+    end
+
+    def invalidate_conf
```
nicklewis: Is this used for anything?

lib/bolt/target.rb
```diff
@@ -21,7 +21,34 @@ def self.parse_urls(urls)
...
+      @password = @uri_obj.password
+      @port = @uri_obj.port
       @options = options || {}
+      @configured = false
+    end
+
+    def update_conf(conf)
+      @protocol = conf[:transport] unless protocol
+
+      t_conf = conf[:transports][protocol.to_sym]
```
nicklewis: It might be clearer to store a `@transport_conf` instance variable and write accessors that handle picking between the conf value and the assigned value. An `options` method could handle removing the url keys when requested, or we could continue to store those in their own variable. That would also simplify the `invalidate_conf` bit (though if it's actually not used, maybe that's a smaller win), since it could just unassign the `@transport_conf`.

lib/bolt/util.rb
```diff
@@ -0,0 +1,47 @@
+module Bolt
+  module Util
+    # CODEREVIEW I hate mixing in random modules
+    class << self
+      def read_config_file(path, default_paths = nil)
```
nicklewis: Extracting these makes some sense. This version isn't being used universally though, which it should.

lib/bolt/util.rb
```diff
@@ -0,0 +1,47 @@
+module Bolt
+  module Util
+    # CODEREVIEW I hate mixing in random modules
+    class << self
```
nicklewis: I think the common pattern for this is to define these as standard methods and then `extend self`. I guess all that accomplishes is the ability to mixin the module, which there's not a good reason for.

lib/bolt/cli.rb
```diff
@@ -215,6 +215,10 @@ def create_option_parser(results)
                 'Specify where to load the config file from') do |path|
           results[:configfile] = path
         end
+        opts.on('--inventoryfile INVENTORY_PATH',
```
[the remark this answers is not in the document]
adreyer: I don't like inventory since that might be specifying what source to use. 
I kind of prefer inventory-file but it doesn't align with configfile

lib/bolt/cli.rb
```diff
@@ -426,15 +425,18 @@ def execute(options)
...
+
       if options[:mode] == 'plan'
         executor = Bolt::Executor.new(@config, options[:noop], true)
-        result = pal.run_plan(options[:object], options[:task_options], executor)
+        result = pal.run_plan(options[:object], options[:task_options], executor, inventory)
         outputter.print_plan_result(result)
         # An exception would have been raised if the plan failed
         code = 0
       else
+        # This executor isn't used for tasks
```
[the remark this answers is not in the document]
adreyer: sorry thats inaccurate, from my last PR and added accidentally fixing merge conflicts

lib/bolt/cli.rb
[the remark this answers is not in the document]
nicklewis: I didn't realize we used `configfile`, whoops. Well, consistency is probably better...

lib/bolt/inventory.rb
```diff
@@ -0,0 +1,239 @@
...
+        @path << parent_group
+      end
+
+      def message
+        "#{@_message} for group at #{path}"
+      end
+    end
+
+    # Group is a specific implementation of Inventory based on structred data.
+    class Group
```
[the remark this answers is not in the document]
adreyer: not really

lib/bolt/inventory.rb
```diff
@@ -0,0 +1,239 @@
...
+          end
+        end
+        nil
+      end
+
+      def node_data(node_name)
+        # This pollutes with name
+        data_merge(
+          { 'config' => @config, 'groups' => [@name] },
+          @nodes.find { |n| n['name'] == node_name }
```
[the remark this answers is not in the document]
adreyer: it appears that would sneak something into the "groups" key which wasn't actually used outside the groups debugging method. Validating the schema of node or selecting only the keys we expect to handle(config) could fix that.

lib/bolt/inventory.rb
```diff
@@ -0,0 +1,239 @@
...
+      # This only handles config and groups for now.
+      # config hashes are deep merged prefering values in data1
+      # data1 is appended to the group array after data2
+      # Returns nil if both data1 and data2 are nil
+      def data_merge(data1, data2)
+        recursive_merge = proc do |_key, h1, h2|
+          if h1.is_a?(Hash) && h2.is_a?(Hash)
+            h1.merge(h2, &recursive_merge)
+          else
+            h2.nil? ? h1 : h2
```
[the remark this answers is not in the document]
adreyer: I remember adding it later but I can't remember why I'd want to prevent unsetting variables.

lib/bolt/inventory.rb
```diff
@@ -0,0 +1,239 @@
...
+      #   @groups.reverse.each_with_object(data) do |g, acc|
+      #     g_hash = g.node_hash
+      #     acc.deep_merge(g_hash)
+      #   end
+      # end
+    end
+
+    def self.from_config(config)
+      # CODEREVIEW should we accept .yml here?
+      default_paths = [File.expand_path(File.join('~', '.puppetlabs', 'bolt', 'inventory.yaml'))]
```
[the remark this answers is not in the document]
adreyer: We error if the inventory was intentionally set and absent but don't otherwise.

lib/bolt/target.rb
adreyer: no neither is @configured at this point. The idea was that "get_targets" would not search and update the targets configuration if it was already configured.

lib/bolt/target.rb
adreyer: I had something closer to that originally. I'll change it back.

invalidate_conf was intended not to actually remove the conf but to mark is as an 'expired configuration cache' so the target would get it's configuration updated. It's not used and should just be removed.

lib/bolt/inventory.rb
```diff
@@ -0,0 +1,239 @@
...
+          if (d = g.data_for(node_name))
+            data_merge(d, acc)
+          else
+            acc
+          end
+        end
+
+        # If the node is in this group add group data
+        if data || node_names.include?(node_name)
+          data_merge(node_data(node_name), data)
```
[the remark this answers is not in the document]
adreyer: no it means the child is always preferred. In fact there is bug here where the spec says the parents "node" definition should override the child group definition and thats not happening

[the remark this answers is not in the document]
nicklewis: Ah I think I got confused by the code above where we swap the arguments to `data_merge`, and thought the semantics were left wins. :+1:

lib/bolt/util.rb
adreyer: Is there a reason not to?
Should we just mix this in anywhere we need it to be more idiomatic?

nicklewis: I don’t think it makes sense to mix it in, since it’s basically semantically void. May as well just treat it like a static class.

lib/bolt/inventory.rb
```diff
@@ -0,0 +1,102 @@
...
+        @path << parent_group
+      end
+
+      def message
+        "#{@_message} for group at #{path}"
+      end
+    end
+
+    def self.default_paths
+      # CODEREVIEW should we accept .yml here?
```
adreyer: This is the only remaining question I have

lib/bolt/inventory.rb
```diff
@@ -0,0 +1,101 @@
...
+        "#{@_message} for group at #{path}"
+      end
+    end
+
+    def self.default_paths
+      [File.expand_path(File.join('~', '.puppetlabs', 'bolt', 'inventory.yaml'))]
+    end
+
+    def self.from_config(config)
+      data = Bolt::Util.read_config_file(config[:inventoryfile], default_paths)
```
nicklewis: We should pass `'inventory'` at the end here, or it will log `Could not read file file` if there's a problem.

lib/bolt/util.rb
```diff
@@ -0,0 +1,54 @@
...
+        end
+
+        path = File.expand_path(path)
+        # safe_load doesn't work with psych in ruby 2.0
+        # The user controls the configfile so this isn't a problem
+        # rubocop:disable YAMLLoad
+        File.open(path, "r:UTF-8") { |f| YAML.load(f.read) }
+      rescue Errno::ENOENT
+        if path_passed
+          raise Bolt::CLIError, "Could not read #{file_name} file: #{path}"
```
nicklewis: Actually per my inventory comment, I think it makes more sense to just remove the word "file" from this strings. Otherwise there's another issue where config files will say "Could not read config file file".

adreyer: for now I think thats true. Long term I think inventory file may be meaningful

PR #268, opened 2018-01-30
(maint) Update submodules to pull in fixes for listing tasks/plans

nicklewis:
Previously, listing tasks or plans with `bolt task show` or `bolt plan
show` would fail if any task or plan was invalid. This issue was fixed
in Puppet, and this commit updates the submodule to a version that
includes the fix. This required a couple of test updates to account for
the new and improved behavior.


spec/bolt/cli_spec.rb
```diff
@@ -747,12 +750,16 @@ def stub_config(config, file_content = nil)
...
-        it "task show displays an error" do
+        it "task show prints a warning but shows other valid tasks" do
           options = {
             mode: 'task',
             action: 'show'
           }
-          expect { cli.execute(options) }.to raise_error(/unexpected token at/)
+          cli.execute(options)
+          json = JSON.parse(output.string)
+          expect(json).to eq([['sample::ok', nil]])
```
zreichert: Where is the `sample::ok` task?  I don't see it in the fixtures.

nicklewis: I forgot to `git add` those! Will do.

nicklewis:
Updated with the new fixtures included

spec/bolt/cli_spec.rb
zreichert: 👍

PR #269, opened 2018-01-31
(BOLT-291) coerce parameter type Data to Any

zreichert:


lib/bolt/outputter.rb
```diff
@@ -14,6 +14,15 @@ def self.for_format(format)
     def initialize(stream = $stdout)
       @stream = stream
     end
+
+    def replace_undef(params)
```
nicklewis: This could use a comment explaining what it's doing and why. Also it might be more appropriately named `replace_data` or `replace_data_type`. Technically it will be Data when it was undefined, but it can also be set to that explicitly; also, `undef` has very specific meaning in puppet types and doesn't apply here.

lib/bolt/outputter.rb
```diff
@@ -14,6 +14,15 @@ def self.for_format(format)
...
+      params.map do |_, v|
+        t = v['type']
+        if defined?(t.name) && t.name == 'Data'
```
nicklewis: `defined?(t.name)` should be `t.respond_to?(:name)`

What case will it not respond to `name`?

zreichert: I think its because the spec tests insert a String.  I'll hunt those down and put a double in place.

zreichert:
@nicklewis I did't realize that we where just interpolating the Type when we built the message.  I simplified the new method in `outputter.rb` to call `to_s` rather than `name` it simplifies the testing.

PR #274, opened 2018-02-02
(BOLT-253) Use containers instead of vms for tests

lucywyman:
At least the unit and integration tests


nicklewis:
This is awesome.

Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM rastasheep/ubuntu-sshd
+
+RUN apt-get update && apt-get -y install sudo
+
+RUN useradd -m bolt && echo "bolt:bolt" | chpasswd && adduser bolt sudo
+
+CMD ["/usr/sbin/sshd", "-D"]
```
lucywyman: I'm not sure I actually need this.

spec/bolt/node/ssh_spec.rb
```diff
@@ -139,43 +138,14 @@ def result_value(stdout = nil, stderr = nil, exit_code = 0)
     end
   end
 
-  context "when executing with private key" do
```
MikaelSmith: Why remove these tests? They seem useful.

spec/bolt/node/ssh_spec.rb
```diff
@@ -14,10 +14,9 @@ def mk_config(conf)
   end
 
   let(:hostname) { ENV['BOLT_SSH_HOST'] || "localhost" }
-  let(:user) { ENV['BOLT_SSH_USER'] || "vagrant" }
-  let(:password) { ENV['BOLT_SSH_PASSWORD'] || "vagrant" }
-  let(:port) { ENV['BOLT_SSH_PORT'] || 20022 }
-  let(:key) { ENV['BOLT_SSH_KEY'] || Dir[".vagrant/**/private_key"] }
+  let(:user) { ENV['BOLT_SSH_USER'] || "root" }
```
MikaelSmith: Eventually I can see us wanting to test both
- login as root and sudo to non-root
- login as non-root and sudo to root

I have some concern about changing from non-root to root though, as I believe some tests around sudo specifically test escalating to root.

docker-compose.yml
```diff
@@ -0,0 +1,12 @@
...
+  ubuntu_node:
+    build: .
+    ports:
+      - "2224:22"
+
+  # This is the same as used in the Vagrantfile
+  centos_node:
+    image: jdeathe/centos-ssh:centos-6
+    ports:
+      - "2226:22"
```
MikaelSmith: Is this for tests that haven't been written? Doesn't look like anything interacts with 2 targets yet.

spec/bolt/node/ssh_spec.rb
lucywyman: M'k -- root is the default user for containers.  Should adding a new user with escalating privileges be part of this ticket, or should it be separate work?

docker-compose.yml
lucywyman: Yes, I added it because I wasn't sure which os we should use, and didn't see too much harm in it being there? But I also don't have specific plans for it, so can remove it.

spec/bolt/node/ssh_spec.rb
lucywyman: Fixed, though now the host key error test is failing and I cannot figure out why....

resources/private_key
```diff
@@ -0,0 +1,51 @@
+-----BEGIN RSA PRIVATE KEY-----
```
lucywyman: Uploading a private key feels like "I'm doing it wrong", but I'm not sure how else to be able to run the ssh key tests locally? Is there a way to generate this each time the tests are run and use that instead?

adreyer: This is probably ok otherwise you'd have to generate a key when bringing up the containers.
It should go in fixtures though.

spec/bolt/node/ssh_spec.rb
```diff
@@ -64,13 +64,14 @@ def result_value(stdout = nil, stderr = nil, exit_code = 0)
...
     end
 
-    it "rejects the connection if host key verification fails" do
-      expect_node_error(Bolt::Node::ConnectError,
-                        'HOST_KEY_ERROR',
-                        /Host key verification failed/) do
-        ssh.connect
-      end
-    end
+    # TODO: I'm not sure why this is failing
```
lucywyman: I'm not sure why adding the ssh keys caused this to fail.

spec/bolt/node/ssh_spec.rb
MikaelSmith: Creating non-default users could be part of the container creation or test. We can write the tests so they require root access, but I'd like to preserve tests that login to a non-root account and test that they can escalate correctly.

If you want to make that separate work, then make sure the current tests still correctly check that run-as works.

spec/bolt/node/ssh_spec.rb
MikaelSmith: Ah, that could be similar to what I found in https://github.com/puppetlabs/bolt/commit/ebd99efdc9f74bd89fd92701b3cb3b58fc0a7f03. This test explicitly expects the key not be in known hosts. Then the rest of the tests use `-k` (or `--no-host-key-check`).

spec/lib/bolt_spec/conn.rb
```diff
@@ -2,18 +2,22 @@ module BoltSpec
...
 
       {
         protocol: transport,
         host: ENV["BOLT_#{tu}_HOST"] || "localhost",
-        user: ENV["BOLT_#{tu}_USER"] || "vagrant",
-        password: ENV["BOLT_#{tu}_PASSWORD"] || "vagrant",
+        user: ENV["BOLT_#{tu}_USER"] || default_user,
+        password: ENV["BOLT_#{tu}_PASSWORD"] || default_password,
         port: ENV["BOLT_#{tu}_PORT"] || default_port,
         key: ENV["BOLT_#{tu}_KEY"] || Dir[".vagrant/**/private_key"]
```
MikaelSmith: Does this need to change?

lucywyman: I think I need to change the default ssh user to 'bolt', but since for now we plan to keep using vagrant for windows tests and docker for ssh tests I think this is the best solution? Alternatively we could add a vagrant user to the containers (shrug).

Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+
+# Add bolt user
+RUN useradd bolt && echo "bolt:bolt" | chpasswd && adduser bolt sudo
+
+# Add SSH key support
+RUN mkdir -p /home/bolt/.ssh/
+COPY spec/fixtures/keys/id_rsa.pub /home/bolt/.ssh/id_rsa.pub
+COPY spec/fixtures/keys/id_rsa.pub /home/bolt/.ssh/authorized_keys
+RUN touch /home/bolt/.ssh/known_hosts
+RUN echo "    IdentityFile /home/bolt/.ssh/id_rsa" >> /etc/ssh/ssh_config
```
MikaelSmith: I think the test below will fail if you remove this line. We expect to require insecure in tests.

PR #275, opened 2018-02-02
(TEST) Use GCE in test matrix

Iristyle:


nicklewis:
@Iristyle Are you doing anything with this PR or should it be closed?

PR #276, opened 2018-02-02
(BOLT-218) Upload files as the run-as user

nicklewis:
Previously, uploaded files would be owned by the user making the SSH
connection. For `file upload`, that was plainly incorrect. For tasks and
scripts, that caused an issue where the uploaded file might or might not be
executable by the user in question, depending on how the tempdir was created
and what the original file permissions were.

We now always chown the file to the run-as user, to ensure that the user can
actually execute the file.

We also now support the `_run_as` argument to the `file_upload()` plan
function, which has the same semantics as passing `--run-as` to `bolt file
upload` on the CLI.

### DIFF
--- a/lib/bolt/executor.rb
+++ b/lib/bolt/executor.rb
@@ -150,7 +150,7 @@ module Bolt
       r
     end
 
-    def file_upload(targets, source, destination)
+    def file_upload(targets, source, destination, options = {})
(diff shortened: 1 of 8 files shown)


nicklewis:
Going to add some tests for running tasks/scripts.

PR #279, opened 2018-02-05
(BOLT-123) Add fail_plan function to boltlib

adreyer:
This adds a fail_plan function to the boltlib module. This function
should be used to raise an error if a plan fails.

### DIFF
--- a/lib/bolt/error.rb
+++ b/lib/bolt/error.rb
@@ -42,4 +42,11 @@ module Bolt
       @error_code = 2
     end
   end
+
+  class PlanFailure < Error
+    def initialize(*args)
+      super(*args)
+      @error_code = 2
+    end
+  end
 end
--- a/spec/lib/bolt_spec/integration.rb
+++ b/spec/lib/bolt_spec/integration.rb
@@ -1,6 +1,6 @@
 module BoltSpec
   module Integration
-    def run_cli(arguments)
+    def run_cli(arguments, rescue_exec: false)
       cli = Bolt::CLI.new(arguments)
 
       # prevent tests from reading users config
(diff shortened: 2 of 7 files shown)


lib/bolt/error.rb
```diff
@@ -42,4 +42,12 @@ def initialize(resultset, action, object)
       @error_code = 2
     end
   end
+
+  class PlanFailure < Error
+    def initialize(*args)
+      super(*args)
+      # CODEREVIEW: what should this be?
+      @error_code = 3
```
lucywyman: I think the message probably matters more than the numeric code? We could also have 1xx be task failures, and 2xx be plan failures. But ultimately I think the message is more important.

adreyer: this eventually become the exit code so we probably don't want to stake out that much space. 

I think the question is should the exitcode be
1. we use this for general bolt failures
2. used for "runfailure" when a bolt execution failed automatically because a task/script/command/upload failed on some targets.
3. Some new number to differentiate explicit plan failures from run_* automatic failures.

MikaelSmith: I figured plan failure would be like other execution failures.  Otherwise 3 seems fine (and we should document that somewhere).

spec/lib/bolt_spec/integration.rb
```diff
@@ -11,12 +11,22 @@ def run_cli(arguments)
...
 
       opts = cli.parse
-      cli.execute(opts)
+
+      if rescue_exec
+        err_kls = error_support ? Bolt::Error : StandardError
+        begin
+          cli.execute(opts)
+        # rubocop:disable HandleExceptions
+        rescue err_kls
```
MikaelSmith: Why return output after this rather than the error object?

adreyer: because the only thing the error object is used for is the exit code. The output is where the cli actually did everything

PR #281, opened 2018-02-06
(maint) Pin FFI to 1.9.18

MikaelSmith:
FFI releases since 1.9.18 seem to be messed up. On Centos 7 they now try
to build libffi and require autoconf. Pin to 1.9.18 until that mess
settles.

### DIFF
--- a/bolt.gemspec
+++ b/bolt.gemspec
@@ -44,6 +44,9 @@ Gem::Specification.new do |spec|
   spec.add_dependency "win32-security", "= 0.2.5"
   spec.add_dependency "win32-service", "= 0.8.8"
 
+  # Pin FFI until its releases are fixed up
+  spec.add_dependency "ffi", "<= 1.9.18"
+
   # there is a bug in puppetlabs_spec_helper for modules without fixtures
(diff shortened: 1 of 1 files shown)


Iristyle:
@MikaelSmith were those fails against 1.9.21?

MikaelSmith:
Yes. 1.9.21 was requiring autoconf to build. If that's a new thing that's going to stick around, we can adjust our acceptance setup to install it, but I didn't know if it would be.

Iristyle:
I'm inclined to follow the same pinning for other projects (like Puppet itself) that consume FFI based on your additional feedback. 

I noticed the Windows gems roughly doubled in size from 1.9.18 to 1.9.21, and it looks like there is a new maintainer who is accidentally introducing breaking changes in .z releases. I think the only benefit right now is Ruby 2.5 support, so probably worth holding off across the board until we're confident that these problems have been corrected and/or Ruby 2.5 is being tested. The churn probably isn't worth it right now.

Thanks for the info!

Iristyle:
Filed Puppet PR at https://github.com/puppetlabs/puppet/pull/6621

PR #284, opened 2018-02-07
(BOLT-255) Replace Node classes with stateless Transports

nicklewis:
Previously, "transports" were implemented as Node classes, where each instance
of a Node represented a single target plus its connection. The executor would
connect and disconnect nodes, and would invoke methods on the Node objects to
run tasks, etc. That model makes it impossible to have multiple targets share a
single logical "transport" or connection.

We now instead have Transport classes that are responsible for managing the
lifecycle of any connections they need, with the executor passing targets
instead of nodes.

For the SSH and WinRM transports, we still follow a relatively similar model
with new SSH::Connection and WinRM::Connection classes that hold the connection
information and the target. Those classes are responsible for direct
interaction with a single target, where the transport classes handle the logic
for how to upload, stage and run scripts, etc.

For the Orch transport, the logic is all handled directly in the transport,
with a single client connection that's managed by the transport itself.


lib/bolt/executor.rb
```diff
@@ -26,39 +34,22 @@ def initialize(config = Bolt::Config.new, noop = nil, plan_logging = false)
...
-
-      nodes.each { |node|
+      targets.each { |target|
         pool.post do
           result =
             begin
-              @notifier.notify(callback, type: :node_start, target: node.target) if callback
-              node.connect
-              yield node
+              transport = @transports[target.protocol || 'ssh']
```
adreyer: I don't think we need this when running. We had to handle this default in inventory or we couldn't have loaded config options properly.

lib/bolt/executor.rb
```diff
@@ -26,39 +34,22 @@ def initialize(config = Bolt::Config.new, noop = nil, plan_logging = false)
...
+              yield transport, target
             rescue StandardError => ex
-              Bolt::Result.from_exception(node.target, ex)
-            ensure
-              begin
-                node.disconnect
-              rescue StandardError => ex
-                @logger.info("Failed to close connection to #{node.uri} : #{ex.message}")
-              end
+              Bolt::Result.from_exception(target, ex)
```
adreyer: Now that the diconnect isn't here this will only catch logging/notifying issues right?

lib/bolt/transport/orch.rb
```diff
@@ -0,0 +1,118 @@
...
+        content = Base64.encode64(content)
+        params = {
+          action: 'script',
+          content: content,
+          arguments: arguments
+        }
+        unwrap_bolt_result(target, run_task(target, BOLT_MOCK_FILE, 'stdin', params))
+      end
+
+      def run_task(target, task, _inputmethod, arguments, _options = {})
```
adreyer: oh yeah one of the previous transport standardization tickets was expected to update the run_task api to expect an object not a path. We should probably ticket that separately.

lib/bolt/transport/orch.rb
```diff
@@ -0,0 +1,118 @@
...
+        body = { task: task_name_from_path(task),
+                 environment: target.options[:orch_task_environment],
+                 noop: arguments['_noop'],
+                 params: arguments.reject { |k, _| k == '_noop' },
+                 scope: {
+                   nodes: [target.host]
+                 } }
+        # Should we handle errors here or let them propagate?
```
adreyer: Now is the time to answer this. I think we can leave it as is and remove this comment

lib/bolt/transport/ssh.rb
```diff
@@ -0,0 +1,144 @@
...
+          }
+        end
+      end
+
+      def with_connection(target)
+        conn = Connection.new(target)
+        conn.connect
+        yield conn
+      ensure
+        conn.disconnect if conn
```
adreyer: this will error sometimes after scp fails. Is that still handled properly?

lib/bolt/transport/ssh.rb
```diff
@@ -0,0 +1,144 @@
...
+      end
+
+      def running_as(connection, user)
+        target = connection.target
+        original_run_as = target.options[:run_as]
+        target.options[:run_as] = user
```
adreyer: I think it might be worth avoiding this by saving a "run_as" value on the connect or something. The less we mutate target objects the better.

lib/bolt/executor.rb
```diff
@@ -4,18 +4,26 @@
...
   class Executor
-    attr_reader :noop
+    attr_reader :noop, :transports
     attr_accessor :run_as
 
     def initialize(config = Bolt::Config.new, noop = nil, plan_logging = false)
       @config = config
       @logger = Logging.logger[self]
 
+      @transports = {
```
adreyer: previously we didn't initialize transports until we used them.

lib/bolt/executor.rb
adreyer: I guess if a user set nil as transport in config?

lib/bolt/executor.rb
nicklewis: Yeah, because we run everything wrapped in `with_exception_handling` which does exactly the same thing. So I think this isn't necessary anymore.

lib/bolt/executor.rb
nicklewis: This was really just here for tests, so we don't have to put a protocol on all the URIs or call `update_conf` every time.

lib/bolt/transport/ssh.rb
nicklewis: I expect that would cause a failure now. I'll update that.

lib/bolt/executor.rb
nicklewis: I suppose this does defeat the purpose of moving the require statements inside the initialize methods... I'll update.

MikaelSmith:
This has some conflicts to resolve now.

nicklewis:
Rebased!

PR #286, opened 2018-02-08
(BOLT-307) Add _catch_error option to run_plan

adreyer:
In order to add module unit tests I'll have to spread the "handles errors" ruby 2.0 logic there. I think it's better to just wait until we drop ruby 2.0 support to test at that level.

### DIFF
--- a/lib/bolt/error.rb
+++ b/lib/bolt/error.rb
@@ -25,20 +25,24 @@ module Bolt
     def to_json(opts = nil)
       to_h.to_json(opts)
     end
+
+    def to_puppet_error
+      Puppet::DataTypes::Error.from_asserted_hash(to_h)
+    end
   end
 
   class RunFailure < Error
-    attr_reader :resultset
+    attr_reader :result_set
 
-    def initialize(resultset, action, object)
+    def initialize(result_set, action, object)
       details = {
-        action: action,
-        object: object,
-        failed_targets: resultset.error_set.names
+        'action' => action,
+        'object' =>  object,
+        'result_set' => result_set
       }
-      message = "Plan aborted: #{action} '#{object}' failed on #{details[:failed_targets].length} nodes"
+      message = "Plan aborted: #{action} '#{object}' failed on #{result_set.error_set.length} nodes"
       super(message, 'bolt/run-failure', details)
-      @resultset = resultset
+      @result_set = result_set
       @error_code = 2
     end
   end
--- a/modules/boltlib/lib/puppet/functions/run_plan.rb
+++ b/modules/boltlib/lib/puppet/functions/run_plan.rb
@@ -8,6 +8,8 @@
 #   }
 #   run_plan('myplan', { x => 'testing' })
 #
+require 'bolt/error'
+
 Puppet::Functions.create_function(:run_plan, Puppet::Functions::InternalFunction) do
   dispatch :run_plan do
     scope_param
@@ -29,7 +31,7 @@ Puppet::Functions.create_function(:run_plan, Puppet::Functions::InternalFunction
       )
     end
 
-    use_args = named_args.reject { |k, _| k.start_with?('_') }
+    params = named_args.reject { |k, _| k.start_with?('_') }
 
     loaders = closure_scope.compiler.loaders
     # The perspective of the environment is wanted here (for now) to not have to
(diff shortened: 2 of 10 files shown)


lib/bolt/error.rb
```diff
@@ -25,18 +25,24 @@ def to_h
...
     def initialize(resultset, action, object)
       details = {
-        action: action,
-        object: object,
-        failed_targets: resultset.error_set.names
+        'action' => action,
+        'object' =>  object,
+        # CODEREVIEW these aren't target objects should we remove this?
+        'failed_targets' => resultset.error_set.names,
+        'resultset' => resultset
```
adreyer: This should be result_set to align with ok_set and error_set

lib/bolt/error.rb
```diff
@@ -25,20 +25,26 @@ def to_h
...
-    def initialize(resultset, action, object)
+    def initialize(result_set, action, object)
       details = {
-        action: action,
-        object: object,
-        failed_targets: resultset.error_set.names
+        'action' => action,
+        'object' =>  object,
+        # CODEREVIEW these aren't target objects should we remove this?
+        'failed_targets' => result_set.error_set.names,
```
MikaelSmith: Why not pass the target objects?

modules/boltlib/lib/puppet/functions/run_plan.rb
```diff
@@ -41,7 +43,20 @@ def run_plan(scope, plan_name, named_args = {})
         old_run_as = executor.run_as
         executor.run_as = run_as
       end
-      result = func.class.dispatcher.dispatchers[0].call_by_name_with_scope(scope, use_args, true)
+      if named_args['_catch_errors']
```
MikaelSmith: I think spaces before and after this if-block would help readability a bit.

modules/boltlib/lib/puppet/functions/run_plan.rb
```diff
@@ -41,7 +43,20 @@ def run_plan(scope, plan_name, named_args = {})
...
+        begin
+          result = func.class.dispatcher.dispatchers[0].call_by_name_with_scope(scope, use_args, true)
```
MikaelSmith: What? This is long... if it doesn't make sense for whatever it's calling to provide a more readable signature, I'd opt to put it in its own function that has a more readable name.

modules/boltlib/lib/puppet/functions/run_plan.rb
```diff
@@ -41,7 +43,20 @@ def run_plan(scope, plan_name, named_args = {})
...
+        rescue Puppet::PreformattedError => err
+          # CODEREVIEW: should this be ducktyped?
```
MikaelSmith: I'm not sure why that would be helpful. It would have to implement `to_puppet_error`? I'd at least want a more sensible interface if it's something we expect people to extend.

modules/boltlib/lib/puppet/functions/run_plan.rb
MikaelSmith: I guess this was here before. But since we're calling it twice, I'd opt for taking this opportunity to make it more readable.

A separate concern is that we could simplify this by putting the `_catch_errors` check in the rescue block and reraise if false, but I'm comfortable avoiding it completely as well.

spec/fixtures/modules/error/plans/run_fail.pp
```diff
@@ -0,0 +1,5 @@
+plan error::run_fail(
+  String $target
+) {
+ run_task('error::fail', $target)
```
MikaelSmith: It'd also be interesting to have a simple case that catches `fail_plan`.

MikaelSmith: Oh, that's what `error::catch_plan` does.

spec/lib/bolt_spec/integration.rb
```diff
@@ -29,7 +29,7 @@ def run_cli_json(arguments, **opts)
       output = run_cli(arguments, **opts)
 
       begin
-        result = JSON.parse(output)
+        result = JSON.parse(output, quirks_mode: true)
```
MikaelSmith: This is supposed to be the default in Ruby 2.0. Did it change later?

lib/bolt/error.rb
adreyer: this was here previously to avoid putting the entire result_set in details. I'm not sure saving a single method call is worth the duplication

spec/lib/bolt_spec/integration.rb
MikaelSmith: Oh, they might be different for parse than load. 🤷‍

modules/boltlib/lib/puppet/functions/run_plan.rb
```diff
@@ -41,7 +43,20 @@ def run_plan(scope, plan_name, named_args = {})
...
+          if err.respond_to?(:cause) && err.cause && err.cause.is_a?(Bolt::Error)
+            result = err.cause.to_puppet_error
+          else
+            raise err
+          end
+        end
+      else
+        result = func.class.dispatcher.dispatchers[0].call_by_name_with_scope(scope, use_args, true)
+      end
       if run_as
```
adreyer: This needs to be in an ensure now

modules/boltlib/lib/puppet/functions/run_plan.rb
adreyer: In the end I missed an ensure around the run_as here so I've moved everything into the same block. I think the backtraces might be worse but the line number is probably more useful anyway.

PR #288, opened 2018-02-08
(BOLT-318) Add support for targeting nodes by group

MikaelSmith:
Adds support for targeting nodes by group. A group name can be used in-place of a target URI on the CLI
```
bolt command run 'hostname' --nodes ssh://host1,group_name
```
or in commands that accept a list of targets
```
run_command('hostname', ['ssh://host1', 'group_name'])
```

Note that the top-level group in inventory is named `all` unless otherwise specified.

Bolt plan run functions will now also split comma-separated list of target URIs, allowing the following to work via a CLI argument similar to `--nodes`
```
plan targets(String $targets) { run_command('hostname', $targets) }
```

A new function has been added to help resolve target input to a list of Targets, `get_targets`. It backs the Bolt plan run functions, so it will split on comma/space and translate groups to targets in that group.

### DIFF
--- a/lib/bolt/cli.rb
+++ b/lib/bolt/cli.rb
@@ -136,7 +136,7 @@ HELP
                   '* protocol is `ssh` by default, may be `ssh` or `winrm`',
                   '* port defaults to `22` for SSH',
                   '* port defaults to `5985` or `5986` for WinRM, based on the --[no-]ssl setting') do |nodes|
-            results[:nodes] << get_arg_input(nodes).split(/[[:space:],]+/).reject(&:empty?)
+            results[:nodes] << get_arg_input(nodes)
           end
         end
         opts.on('-u', '--user USER',
@@ -259,6 +259,12 @@ HELP
       parser
     end
 
+    # Only call after @config has been initialized.
+    def inventory
+      Bolt::Inventory.from_config(@config)
+    end
+    private :inventory
+
     def parse
       if @argv.empty?
         options[:help] = true
@@ -308,6 +314,9 @@ HELP
       options[:leftovers] = remaining
 
       validate(options)
+
+      # After validation, initialize inventory and targets. Errors here are better to catch early.
+      options[:targets] = inventory.get_targets(options[:nodes]) if options[:nodes]
 
       options
     rescue Bolt::CLIError => e
@@ -434,7 +443,6 @@ HELP
         return 0
       end
 
-      inventory = Bolt::Inventory.from_config(@config)
       message = 'There may be processes left executing on some nodes.'
 
       if options[:mode] == 'plan'
@@ -445,7 +453,7 @@ HELP
         code = 0
       else
         executor = Bolt::Executor.new(@config, options[:noop])
-        targets = inventory.get_targets(options[:nodes])
+        targets = options[:targets]
 
         results = nil
         outputter.print_head
--- a/lib/bolt/inventory/group.rb
+++ b/lib/bolt/inventory/group.rb
@@ -101,9 +101,24 @@ module Bolt
         }
       end
 
+      # Returns all nodes contained within the group, which includes nodes from subgroups.
       def node_names
+        @groups.inject(local_node_names) do |acc, g|
+          acc.merge(g.node_names)
+        end
+      end
+
+      # Return a mapping of group names to group.
+      def collect_groups
+        @groups.inject(name => self) do |acc, g|
+          acc.merge(g.collect_groups)
+        end
+      end
+
+      def local_node_names
         @_node_names ||= Set.new(nodes.map { |n| n['name'] })
       end
+      private :local_node_names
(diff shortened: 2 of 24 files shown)


lib/bolt/target.rb
```diff
@@ -10,12 +10,8 @@ def self.from_asserted_hash(hash)
       new(hash['uri'], hash['options'])
     end
 
-    def self.from_uri(uri)
-      new(uri)
-    end
-
     def self.parse_urls(urls)
-      urls.split(/[[:space:],]+/).reject(&:empty?).uniq.map { |url| from_uri(url) }
+      urls.split(/[[:space:],]+/).reject(&:empty?).uniq.map { |url| new(url) }
```
adreyer: Do we need this method anymore?

modules/boltlib/lib/puppet/functions/get_targets.rb
```diff
@@ -0,0 +1,32 @@
...
+# - 'group1'
+# - 'host1,group1,winrm://host2:54321'
+# - ['host1', 'group1', 'winrm://host2:54321']
+#
+# Returns an array of Targets resolved from any target URIs and groups.
+
+require 'bolt/error'
+
+Puppet::Functions.create_function(:get_targets, Puppet::Functions::InternalFunction) do
+  local_types do
```
adreyer: I think this should be a public type now so it can be used in plan signatures. Maybe we should bikeshed the name first...

modules/boltlib/lib/puppet/functions/get_targets.rb
```diff
@@ -0,0 +1,32 @@
...
+# - a target URI
+# - an array of groups and/or target URIs
+# - a string that consists of a comma-separated list of groups and/or target URIs
+#
+# Examples of the above would be
+# - 'group1'
+# - 'host1,group1,winrm://host2:54321'
+# - ['host1', 'group1', 'winrm://host2:54321']
+#
+# Returns an array of Targets resolved from any target URIs and groups.
```
adreyer: We should clarify that it returns an array of unique targets probably

lib/bolt/cli.rb
```diff
@@ -309,6 +315,9 @@ def parse
 
       validate(options)
 
+      # After validation, initialize inventory and targets. Errors here are better to catch early.
+      options[:targets] = inventory.get_targets(options[:nodes]) if options[:nodes]
```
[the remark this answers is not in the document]
MikaelSmith: Partially it was to make it easier to test (see commit message). Otherwise it probably doesn't make much difference whether it's at the end of `parse` or beginning of `execute`.

lib/bolt/inventory/group.rb
```diff
@@ -30,29 +30,38 @@ def initialize(data)
...
+        if node_names.include?(@name)
+          raise ValidationError.new("Group #{@name} conflicts with node of the same name", @name)
+        end
         raise ValidationError.new("Group #{@name} is too deeply nested", @name) if depth > 1
 
         used_names << @name
 
         @nodes.each do |n|
-          raise ValidationError.new("node #{n['name']} does not have a name", @name) unless n['name']
+          raise ValidationError.new("node #{n['name']} does not have a name", n['name']) unless n['name']
```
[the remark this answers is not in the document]
MikaelSmith: Ah yes, haven't added something yet. Will do.

lib/bolt/target.rb
MikaelSmith: Hmm, I'll check tomorrow.

modules/boltlib/lib/puppet/functions/get_targets.rb
MikaelSmith: Yeah. Also, what's all the `Puppet::Functions::InternalFunction` stuff?

modules/boltlib/lib/puppet/functions/get_targets.rb
```diff
@@ -0,0 +1,32 @@
...
+    type 'TargetOrTargets = Variant[String[1], Target, Array[TargetOrTargets]]'
+  end
+
+  dispatch :get_targets do
+    param 'TargetOrTargets', :names
+    return_type 'Array[Target]'
+  end
+
+  def get_targets(names)
+    inventory = Puppet.lookup(:bolt_inventory) { nil }
```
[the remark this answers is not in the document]
MikaelSmith: Should I also guard on `Puppet[:tasks]`, or do we not care much?

lib/bolt/inventory/group.rb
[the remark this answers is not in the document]
MikaelSmith: That particular check doesn't work. URIs are funny.

I think `unless name == Addressable::URI.parse('//'+name).authority` will work though.

MikaelSmith:
Updated.

PR #289, opened 2018-02-09
(PUP-8441) Monkey Patch to_json onto Puppet::DataTypes::Error

adreyer:
This is a temporary hack until we have a better way to generate a JSON
object for Error objects returned from plans. Since our output is based
on to_json now we need Error to support to_json to handle nested
objects.

### DIFF
--- a/lib/bolt/pal.rb
+++ b/lib/bolt/pal.rb
@@ -37,6 +37,14 @@ module Bolt
 
       # Now that puppet is loaded we can include puppet mixins in data types
       Bolt::ResultSet.include_iterable
+
+      # TODO: This is a hack for PUP-8441 remove it once that is fixed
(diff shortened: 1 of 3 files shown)


MikaelSmith:
What triggered this failure previously? If a plan returns an Error?

MikaelSmith:
I'd like to see a test around this.

adreyer:
This fails if the plan returns an error. There was a previous fix for that case but it only works with a top level error not an error inside another object. That is tested by the two run_plan tests that return an error value. Is it worth testing the nested case?

MikaelSmith:
Ah... maybe, if only to make it easier to test when this is fixed in Puppet.

adreyer:
I'll add in integration test then. The fix sounds like it may involve some large changes to how bolt handles output.

PR #291, opened 2018-02-13
(BOLT-316) Batch tasks, etc when using the orch transport

nicklewis:
This pull request adds new methods batch_task, batch_script, batch_command and
batch_upload to transports. These methods are used to indicate to a transport
that it should perform an operation on multiple nodes as part of a single
batch. This is specifically useful for the orch transport, which can run many
nodes as part of a single orchestrator job. Previously, it would create one
orchestrator job for each node.

This PR adds a batches() method to the transport, which is called by
the executor to determine how to batch runs. This allows all the
concurrency constructs to be moved entirely into the executor, so that
transports don't have to care about concurrency at all, other than the
requirement that they be thread-safe.

The executor now has a batch_execute() method that takes a list of
targets, splits them by transport, batches them according to the
transport's batching implementation, and then calls batch_task, etc
on each batch in a thread, collecting the results into a single
ResultSet.

The default implementation of the batches() method will split the list
of targets into batches of 1 target. The corresponding batch_* methods
handle sending start and finish events, and defer to the run_* methods.
This means that transport authors who don't want batching still only
need to implement run_task, etc.

Transports which do have their own batching need to implement a custom
batches() method to decide how batches should be constructed, and need
to implement batch_task, etc appropriately, including sending start and
finish events, but don't need to worry about managing concurrency.

We now use a CachedThreadPool for the executor rather than a FixedThreadPool,
and keep it alive between invocations. The previous behavior was to queue up
all the work on the threadpool and then stop it and wait for everything to
finish. We now create promises for every expected result and block until those
have been delivered.

### DIFF
--- a/lib/bolt/transport/base.rb
+++ b/lib/bolt/transport/base.rb
@@ -1,30 +1,53 @@
-require 'concurrent'
 require 'logging'
 
 module Bolt
   module Transport
+    # This class provides the default behavior for Transports. A Transport is
+    # responsible for uploading files and running commands, scripts, and tasks
+    # on Targets.
+    #
+    # Bolt executes work on the Transport in "batches". To do that, it calls
+    # the batches() method, which is responsible for dividing the list of
+    # Targets into batches according to how it wants to handle them. It will
+    # then call Transport#batch_task, or the corresponding method for another
+    # operation, passing a list of Targets. The Transport returns a list of
+    # Bolt::Result objects, one per Target. Each batch is executed on a
+    # separate thread, controlled by the `concurrency` setting, so many batches
+    # may be running in parallel.
+    #
+    # The default batch implementation splits the list of Targets into batches
+    # of 1. It then calls run_task(), or a corresponding method for other
+    # operations, passing in the single Target.
+    #
+    # Most Transport implementations, like the SSH and WinRM transports, don't
+    # need to do their own batching, since they only operate on a single Target
+    # at a time. Those Transports can implement the run_task() and related
+    # methods, which will automatically handle running many Targets in
+    # parallel, and will handle publishing start and finish events for each
+    # Target.
+    #
+    # Transports that need their own batching, like the Orch transport, can
+    # instead override the batches() method to split Targets into sets that can
+    # be executed together, and override the batch_task() and related methods
+    # to execute a batch of nodes. In that case, those Transports should accept
+    # a block argument and call it with a :node_start event for each Target
+    # before executing, and a :node_result event for each Target after
+    # execution.
     class Base
       attr_reader :logger
 
-      def initialize(_config, executor = Concurrent.global_immediate_executor)
+      def initialize(_config)
         @logger = Logging.logger[self]
-        @executor = executor
       end
 
-      def on(targets, callback)
-        targets.map do |target|
-          Concurrent::Future.execute(executor: @executor) do
-            begin
-              callback.call(type: :node_start, target: target) if callback
-              result = yield target
-              @logger.debug("Result on #{target.uri}: #{JSON.dump(result.value)}")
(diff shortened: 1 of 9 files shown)


lib/bolt/transport/orch.rb
```diff
@@ -58,36 +64,122 @@ def run_script(target, script, arguments, _options = {})
...
+          body = build_request(env_targets, task, arguments)
+
+          # Make promises to hold the eventual results. The `set` operation is
+          # fast, so they can just execute in the thread that's creating them,
+          # using the :immediate executor.
+          result_ivars = Hash[env_targets.map { |target| [target, Concurrent::Promise.new(executor: :immediate)] }]
+
+          @executor.post do
+            env_targets.each do |target|
+              callback.call(type: :node_start, target: target)
```
adreyer: We get a real start event from orch. There should at least be a TODO to use that here.

nicklewis:
Fixed the issue with `$ERROR_INFO` not being actually turned into a `Bolt::Result` object.

spec/bolt/transport/orch_spec.rb
```diff
@@ -13,8 +13,8 @@
   end
 
   let(:targets) do
-    [Bolt::Target.new('node1').update_conf(Bolt::Config.new.transport_conf),
-     Bolt::Target.new('node2').update_conf(Bolt::Config.new.transport_conf)]
+    [Bolt::Target.new('pcp://node1').update_conf(Bolt::Config.new.transport_conf),
+     Bolt::Target.new('pcp://node2').update_conf(Bolt::Config.new.transport_conf)]
```
adreyer: perhaps one of each here?

nicklewis: good call, updated

lib/bolt/transport/orch.rb
```diff
@@ -58,36 +64,116 @@ def run_script(target, script, arguments, _options = {})
         unwrap_bolt_result(target, run_task(target, BOLT_MOCK_FILE, 'stdin', params))
       end
 
+      def build_request(targets, task, arguments)
+        { task: task_name_from_path(task),
+          # XXX What to do with multiple targets?
```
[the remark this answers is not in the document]
nicklewis: Squashed a fix into aa86012914169aa6ccb178630d1b354360665ab8 that introduced the grouping by env

lib/bolt/transport/orch.rb
```diff
@@ -58,36 +64,116 @@ def run_script(target, script, arguments, _options = {})
...
+        results = @client_lock.synchronize do
+          client.run_task(body)
+        end
+
+        process_run_results(targets, results).each do |result|
+          callback.call(type: :node_result, result: result)
+        end
+      end
+
       def run_task(target, task, _inputmethod, arguments, _options = {})
```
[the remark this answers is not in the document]
nicklewis: Pushed a commit removing all the single-target run_* functions.

lib/bolt/transport/base.rb
```diff
@@ -0,0 +1,158 @@
...
+    # responsible for uploading files and running commands, scripts, and tasks
+    # on Targets.
+    #
+    # Bolt executes work on the Transport in "batches". To do that, it calls
+    # the batches() method, which is responsible for dividing the list of
+    # Targets into batches according to how it wants to handle them. It will
+    # then call Transport#batch_task, or the corresponding method for another
+    # operation, passing a list of Targets. The Transport returns a list of
+    # Bolt::Result objects, one per Target. Each batch is executed on a
+    # dedicated thread, so many batches may be running in parallel.
```
adreyer: I would mention that the size of this thread pool is controlled by the concurrency option.

nicklewis: Updated to clarify that

PR #297, opened 2018-02-16
(BOLT-342) Expose url parts on Target Datatype

adreyer:


spec/pal/target_spec.rb
```diff
@@ -15,27 +15,38 @@
...
   it 'should expose uri' do
-    code = <<PUPPET
-$target = Target('pcp://example.com')
-$target.uri
-PUPPET
-    expect(peval(code, pal)).to eq('pcp://example.com')
+    expect(target('$target.uri')).to eq('pcp://user1:pass1@example.com:33')
   end
 
   it 'should expose name' do
```
MikaelSmith: Could use a test for `port`.

PR #303, opened 2018-02-20
(BOLT-340) Properly escape strings when constructing commands

nicklewis:
Previously, we weren't always safely interpolating filenames and variables
into SSH commands. We now construct commands as arrays, using Shellquote to
properly escape them.

When interpolating task parameters to set powershell environment variables,
we were naively using just single quotes to escape the values. That doesn't
work if the value itself contains single quotes. We now use a single-quoted
here string to ensure that we get the literal value as intended.

### DIFF
--- a/lib/bolt/transport/winrm.rb
+++ b/lib/bolt/transport/winrm.rb
@@ -84,7 +84,7 @@ catch
 
           if ENVIRONMENT_METHODS.include?(input_method)
             arguments.each do |(arg, val)|
-              cmd = "[Environment]::SetEnvironmentVariable('PT_#{arg}', '#{val}')"
+              cmd = "[Environment]::SetEnvironmentVariable('PT_#{arg}', @'\n#{val}\n'@)"
               result = conn.execute(cmd)
               if result.exit_code != 0
                 raise EnvironmentVarError(var, value)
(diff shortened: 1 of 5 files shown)


MikaelSmith:
Got a few test failures.

nicklewis:
Updated, they should be working now. 🤞

nicklewis:
Ugh, rubocop failure. Fixed that one too...

lib/bolt/transport/winrm.rb
```diff
@@ -84,7 +84,7 @@ def run_task(target, task, arguments, _options = {})
 
           if ENVIRONMENT_METHODS.include?(input_method)
             arguments.each do |(arg, val)|
-              cmd = "[Environment]::SetEnvironmentVariable('PT_#{arg}', '#{val}')"
+              cmd = "[Environment]::SetEnvironmentVariable('PT_#{arg}', @'\n#{val}\n'@)"
```
Iristyle:
```
PS /Users/Iristyle/source/puppetlabs-dsc/pkg> [Environment]::SetEnvironmentVariable('PT_foo', @'                   >> some val                                                                                                        >> asd;f                                                                                                           >> '@)
```

Yields


```
PS /Users/Iristyle/source/puppetlabs-dsc/pkg> $env:PT_foo
some val
asd;f
```

👍

MikaelSmith: Where do we want to document the limitations of this?

PR #304, opened 2018-02-20
(BOLT-333) Fix packaging, some rename cleanup

MikaelSmith:
Fixes gem packaging to include `bolt-modules`. Also remove an unused variable.

### DIFF
--- a/bolt.gemspec
+++ b/bolt.gemspec
@@ -17,8 +17,8 @@ Gem::Specification.new do |spec|
                        Dir['lib/**/*.rb'] +
                        Dir['vendored/*.rb'] +
                        Dir['vendored/*/lib/**/*.rb'] +
-                       Dir['modules/boltlib/lib/**/*.rb'] +
-                       Dir['modules/boltlib/types/**/*.pp']
+                       Dir['bolt-modules/boltlib/lib/**/*.rb'] +
+                       Dir['bolt-modules/boltlib/types/**/*.pp']
   spec.bindir        = "exe"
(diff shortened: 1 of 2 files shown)


bolt-modules/boltlib/spec/functions/file_upload_spec.rb
```diff
@@ -23,7 +23,7 @@
     let(:message) { 'uploaded' }
     let(:result) { Bolt::Result.new(target, message: message) }
     let(:result_set) { Bolt::ResultSet.new([result]) }
-    let(:module_root) { File.expand_path(fixtures('modules', 'test')) }
+    let(:module_root) { File.expand_path(fixtures('bolt-modules', 'test')) }
```
adreyer: I don't think we need this isn't fixtures/modules created by rspec puppet

MikaelSmith: I'm not exactly sure what the replacement needs to be. We use module_root in the next line.

adreyer: You should be able to just undo this change. a helper will create spec/fixtures/modules is the correct path.

MikaelSmith: Ah, this path is setup by rspec-puppet. Not sure how to access it yet.

MikaelSmith: Done.

PR #305, opened 2018-02-20
(BOLT-344) Restore displaying errors for failed connection on the CLI

MikaelSmith:
Work on batching execution introduced a regression where errors would be
caught but not yielded to the event block passed to the executor.
Therefore the CLI would no longer report connection failures.

Restore yielding transport exceptions as error results if a block is
passed.

### DIFF
--- a/lib/bolt/transport/base.rb
+++ b/lib/bolt/transport/base.rb
@@ -42,13 +42,13 @@ module Bolt
 
       def with_events(target, callback)
         callback.call(type: :node_start, target: target) if callback
-        result = yield
-        @logger.debug("Result on #{target.uri}: #{JSON.dump(result.value)}")
-        callback.call(type: :node_result, result: result) if callback
-        result
-      rescue StandardError => ex
-        result = Bolt::Result.from_exception(target, ex)
-        @logger.debug("Failure on #{target.uri}: #{JSON.dump(result.value)}")
+
+        result = begin
+          yield
+        rescue StandardError => ex
+          Bolt::Result.from_exception(target, ex)
+        end
+
         callback.call(type: :node_result, result: result) if callback
         result
       end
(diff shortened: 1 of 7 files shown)


lib/bolt/transport/base.rb
```diff
@@ -47,7 +47,10 @@ def with_events(target, callback)
         callback.call(type: :node_result, result: result) if callback
         result
       rescue StandardError => ex
-        Bolt::Result.from_exception(target, ex)
+        result = Bolt::Result.from_exception(target, ex)
+        @logger.debug("Failure on #{target.uri}: #{JSON.dump(result.value)}")
+        callback.call(type: :node_result, result: result) if callback
```
adreyer: this will double report if callback.call fails. Is it worth guarding against bugs with an extra begin/rescue around that?

spec/lib/bolt_spec/conn.rb
```diff
@@ -27,10 +27,11 @@ def conn_info(transport)
       }
     end
 
-    def conn_uri(transport, include_password = false)
+    def conn_uri(transport, include_password: false, override_port: false)
```
adreyer: override_port = nil is probably clearer

lib/bolt/transport/base.rb
nicklewis: If `callback.call` fails, who really knows what will be reported? That shouldn't really be able to fail aside from a bug in bolt though, so it doesn't seem too concerning.

MikaelSmith: I think tests will do a better job guarding against those bugs. I'd be tempted to change this control-flow a bit though: if there's an exception in the yield, we should catch it and wrap it in Bolt::Result. Does it really make sense to do the same for an exception in the callback?

MikaelSmith: I added a 2nd commit simplifying this block and intentionally only catching errors from the `yield`.

adreyer: I believe we create empty results somewhere else and a result with a callback traceback is propabably about as useful

MikaelSmith: I'm not sure what your answer implies. Not really familiar with where and why we create empty results.

PR #306, opened 2018-02-21
(BOLT-331) Add helper to generate an inventory from PuppetDB

nicklewis:
This adds the bolt-inventory-pdb command which can be used to populate a
"template" inventory file with nodes based on PuppetDB queries.

The input to the command is basically an inventory file, except groups
should have a "query" field instead of a "nodes" field. The command will
read the input, execute each of the queries against PuppetDB, and print
a complete inventory with the "nodes" field populated for each group.

This command supports both token and cert auth to PuppetDB, and will
read the puppetdb.conf from the `puppet query` command if it's present.

### DIFF
--- a/lib/bolt_ext/puppetdb_inventory.rb
+++ b/lib/bolt_ext/puppetdb_inventory.rb
@@ -1,0 +1,243 @@
+#!/usr/bin/env ruby
+
+require 'json'
+require 'httpclient'
+require 'optparse'
+require 'yaml'
+
+module Bolt
+  class PuppetDBInventory
+    class Client
+      def self.from_config(config)
+        uri = URI.parse(config['server_urls'].first)
+        uri.port ||= 8081
+
+        cacert = File.expand_path(config['cacert'])
+        token = config.token
+
+        cert = config['cert']
+        key = config['key']
+
+        new(uri, cacert, token: token, cert: cert, key: key)
+      end
+
+      def initialize(uri, cacert, token: nil, cert: nil, key: nil)
+        @uri = uri
+        @cacert = cacert
+        @token = token
+        @cert = cert
+        @key = key
+      end
+
+      def query_certnames(query)
+        return [] unless query
+
+        body = JSON.generate(query: query)
+
+        response = http_client.post("#{@uri}/pdb/query/v4", body: body, header: headers)
+        if response.code != 200
+          raise "Failed to query PuppetDB: #{response.body}"
+        else
+          results = JSON.parse(response.body)
+          if results.first && !results.first.key?('certname')
+            raise "Query results did not contain a 'certname' field: got #{results.first.keys.join(', ')}"
+          end
+          results.map { |result| result['certname'] }.uniq
+        end
+      end
+
+      def http_client
+        return @http if @http
+        @http = HTTPClient.new
+        @http.ssl_config.set_client_cert_file(@cert, @key)
+        @http.ssl_config.add_trust_ca(@cacert)
+
+        @http
+      end
+
+      def headers
+        headers = { 'Content-Type' => 'application/json' }
+        headers['X-Authentication'] = @token if @token
+        headers
+      end
+    end
+
+    class Config
+      DEFAULT_TOKEN = File.expand_path('~/.puppetlabs/token')
+      DEFAULT_CONFIG = File.expand_path('~/.puppetlabs/client-tools/puppetdb.conf')
+
+      def initialize(config_file, options)
+        @settings = load_config(config_file)
+        @settings.merge!(options)
+
+        expand_paths
+        validate
+      end
+
+      def load_config(filename)
+        if filename
+          if File.exist?(filename)
+            config = JSON.parse(File.read(filename))
+          else
+            raise "config file #{filename} does not exist"
+          end
+        elsif File.exist?(DEFAULT_CONFIG)
+          config = JSON.parse(File.read(DEFAULT_CONFIG))
+        end
+        config.fetch('puppetdb', {})
+      end
+
+      def token
+        return @token if @token
+        if @settings['token']
+          File.read(@settings['token'])
+        elsif File.exist?(DEFAULT_TOKEN)
+          File.read(DEFAULT_TOKEN)
+        end
+      end
+
+      def [](key)
+        @settings[key]
+      end
+
+      def expand_paths
+        %w[cacert cert key token].each do |file|
+          @settings[file] = File.expand_path(@settings[file]) if @settings[file]
+        end
+      end
+
+      def validate_file_exists(file)
+        if @settings[file] && !File.exist?(@settings[file])
+          raise "#{file} file #{@settings[file]} does not exist"
+        end
+      end
+
+      def validate
+        unless @settings['server_urls']
+          raise "server_urls must be specified in the config file or with --url"
+        end
+        unless @settings['cacert']
+          raise "cacert must be specified in the config file or with --cacert"
+        end
+
+        if (@settings['cert'] && !@settings['key']) ||
+           (!@settings['cert'] && @settings['key'])
+          raise "cert and key must be specified together"
+        end
+
+        validate_file_exists('cacert')
+        validate_file_exists('cert')
+        validate_file_exists('key')
+      end
+    end
+
+    class CLI
+      def initialize(args)
+        @args = args
+        @cli_opts = {}
+        @parser = build_parser
+      end
+
+      def build_parser
+        parser = OptionParser.new('') do |opts|
+          opts.on('--cacert CACERT', "Path to the CA certificate") do |cacert|
+            @cli_opts['cacert'] = cacert
+          end
+          opts.on('--cert CERT', "Path to the certificate") do |cert|
+            @cli_opts['cert'] = cert
+          end
+          opts.on('--key KEY', "Path to the private key") do |key|
+            @cli_opts['key'] = key
+          end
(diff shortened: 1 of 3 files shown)


lib/bolt_ext/puppetdb_inventory.rb
```diff
@@ -0,0 +1,191 @@
...
+      def run
+        @parser.permute(@args)
+
+        if @show_help
+          puts @parser.help
+          return 0
+        end
+
+        unless @inventory_file
+          raise "--file is a required option"
```
lucywyman: Is there a reason it's not `--inventoryfile` for consistency with bolt? I like `--file` more  since less typing.

lib/bolt_ext/puppetdb_inventory.rb
```diff
@@ -0,0 +1,191 @@
...
+        @puppetdb_client = Client.new(config)
+
+        unless File.readable?(@inventory_file)
+          raise "Can't read the inventory file #{@inventory_file}"
+        end
+
+        inventory = YAML.load_file(@inventory_file)
+        resolve_group(inventory)
+
+        puts inventory.to_yaml
```
lucywyman: spare puts (not sure if this is final or not)

MikaelSmith: Does the script write the result to stdout? It'd be nice if that were clear in the help output.

lib/bolt_ext/puppetdb_inventory.rb
```diff
@@ -0,0 +1,191 @@
...
+        headers['X-Authentication'] = @token if @token
+        headers
+      end
+    end
+
+    class Config
+      DEFAULT_TOKEN = File.expand_path('~/.puppetlabs/token')
+
+      def initialize(options)
+        @settings = load_config(File.expand_path('~/.puppetlabs/client-tools/puppetdb.conf'))
```
MikaelSmith: I think it'd make sense for this to be configurable as well, since the config could live in a different location.

lib/bolt_ext/puppetdb_inventory.rb
```diff
@@ -0,0 +1,191 @@
...
+            @inventory_file = File.expand_path(file)
+          end
+          opts.on('-h', '--help', "Display help") do |_|
+            @show_help = true
+          end
+        end
+        parser.banner = <<-BANNER
+  Usage: bolt-inventory-pdb --file <inventory-file> [--url <url>] [auth-options]
+
+  Populate the nodes in an inventory file based on PuppetDB queries.
```
MikaelSmith: We should note how the inventory file is supposed to be formed (i.e. we expect an inventory file that includes `query` keys in place of `nodes`).

lib/bolt_ext/puppetdb_inventory.rb
```diff
@@ -0,0 +1,191 @@
...
+        @settings.merge!(options)
+        validate
+      end
+
+      def load_config(filename)
+        if File.exist?(filename)
+          JSON.parse(File.read(filename)).fetch('puppetdb', {})
+        # Only try to use the default token if it exists, so we don't override
+        # cert/key with a non-existent token
+        elsif File.exist?(DEFAULT_TOKEN)
```
MikaelSmith: `puppetdb.conf` is not required to have token specified, it'll use DEFAULT_TOKEN if not included. We should handle that.

lib/bolt_ext/puppetdb_inventory.rb
```diff
@@ -0,0 +1,191 @@
...
+          puts @parser.help
+          return 0
+        end
+
+        unless @inventory_file
+          raise "--file is a required option"
+        end
+
+        config = Config.new(@cli_opts)
+        @puppetdb_client = Client.new(config)
```
MikaelSmith: This call should be `Client.from_config`. I also found it useful to add a `--trace` option that enables printing `e.backtrace` in the rescue at line 175 for debugging.

MikaelSmith: My patches to get it working
```
diff --git a/lib/bolt_ext/puppetdb_inventory.rb b/lib/bolt_ext/puppetdb_inventory.rb
index 66ce9e1..29bb86c 100755
--- a/lib/bolt_ext/puppetdb_inventory.rb
+++ b/lib/bolt_ext/puppetdb_inventory.rb
@@ -69,7 +69,8 @@ module Bolt

       def initialize(options)
-        @settings = load_config(File.expand_path('~/.puppetlabs/client-tools/puppetdb.conf'))
+        @settings =
+          default_token(load_config(File.expand_path('~/.puppetlabs/client-tools/puppetdb.conf')))
         @settings.merge!(options)
         validate
       end
@@ -77,12 +78,18 @@ module Bolt
       def load_config(filename)
         if File.exist?(filename)
           JSON.parse(File.read(filename)).fetch('puppetdb', {})
+        else
+          {}
+        end
+      end
+
+      def default_token(conf)
         # Only try to use the default token if it exists, so we don't override
         # cert/key with a non-existent token
-        elsif File.exist?(DEFAULT_TOKEN)
-          { 'token' => DEFAULT_TOKEN }
+        if File.exist?(DEFAULT_TOKEN)
+          { 'token' => DEFAULT_TOKEN }.merge(conf)
         else
-          {}
+          conf
         end
       end

@@ -134,6 +141,9 @@ module Bolt
           opts.on('--file FILE', "Path to the inventory file to use") do |file|
             @inventory_file = File.expand_path(file)
           end
+          opts.on('--trace', "Enable stack trace on errors") do |_|
+            @trace = true
+          end
           opts.on('-h', '--help', "Display help") do |_|
             @show_help = true
           end
@@ -160,7 +170,7 @@ module Bolt
         end

         config = Config.new(@cli_opts)
-        @puppetdb_client = Client.new(config)
+        @puppetdb_client = Client.from_config(config)

         unless File.readable?(@inventory_file)
           raise "Can't read the inventory file #{@inventory_file}"
@@ -174,6 +184,7 @@ module Bolt
         return 0
       rescue StandardError => e
         puts "Error: #{e}"
+        puts e.backtrace if e.backtrace && @trace
         return 1
       end
```

lib/bolt_ext/puppetdb_inventory.rb
nicklewis: Whoops, good point. That should probably just be merged into whatever the other result is.

lib/bolt_ext/puppetdb_inventory.rb
nicklewis: Doh, I refactored some of that for testability, but I guess I didn't actually update all the uses of it in the code. Unit tests only get so far. :P

lib/bolt_ext/puppetdb_inventory.rb
MikaelSmith: With stdout, I can't do an in-place update of my inventory file. Trying to do so ends up overwriting the file before it's read.

nicklewis: This `puts` was intended, to write to stdout. But I didn't realize that still wouldn't let you update it in place. I can add an option for output file.

nicklewis:
I pushed some updates in a commit that I'll squash in after it's reviewed.

• It works again!
• `--trace` will print backtraces on error
• `--configfile` specifies the config file to load
• the input file is now a positional argument rather than a flag
• `--output` specifies where to write the output, defaulting to STDOUT
• The default token is always used if present and another token wasn't specified (if you're using cert auth, supplying the token anyway won't hurt)

MikaelSmith:
Just need to fix up tests now.

nicklewis:
Updated the tests and appeased rubocop (which found a bug!)

lib/bolt_ext/puppetdb_inventory.rb
```diff
@@ -125,55 +149,83 @@ def build_parser
...
+                  "Path to the token file",
+                  "Default: #{Config::DEFAULT_TOKEN} if present") do |token|
             @cli_opts['token'] = token
           end
           opts.on('--url URL', "The URL of the PuppetDB server to connect to") do |url|
             @cli_opts['server_urls'] = [url]
           end
-          opts.on('--file FILE', "Path to the inventory file to use") do |file|
-            @inventory_file = File.expand_path(file)
+          opts.on('--configfile CONFIG',
```
MikaelSmith: Naming of configfile and token-file feel inconsistent :(

MikaelSmith: Based on puppetdb-cli and puppet-nogui, I believe this should be `--config`.

nicklewis: I went for consistency with bolt (which uses `--configfile`) over consistency with client tools or good taste. The `--token-file` internal inconsistency seems like a problem though.

I'm fine with whatever, really.

MikaelSmith: Maybe we should change bolt...

MikaelSmith: Turns out `--config` is actually accepted by Bolt, because it's unambiguous.

nicklewis: Changing Bolt seems like the best option. In that case, I'll go with `--config` and `--token-file` and we can deal with Bolt separately?

MikaelSmith: Yeah, let's do that. It's not consistent with anything else we do. Even Puppet uses `--config`.

PR #307, opened 2018-02-21
(BOLT-332) Store vars on target objects

lucywyman:


hlindberg:
See how `LookupContext` in puppet is implemented - it has a cache that is similar to vars on a target.

thallgren:
How about this:
Maintain a `:target_variables` hash in the global Puppet context. Let Bolt provide this hash  when it initializes an execution. The hash is keyed by targed and have values that are hashes. The `Target#set_var` will then essentially do `Puppet.lookup[:target_variables][self][var] = value` and `Target` would no longer hold onto the `@vars` instance variable.

If a `Target#vars` method that returns the full target is required, then the `set_var` method must perform a copy on write as Henrik suggest to prevent that `vars` returns a hash that is mutable. An alternative could be to not add `Target#vars` at all but instead provide a `get_var` method and  iterative methods such as `#vars_each_pair`, `#vars_each_key`, etc. that takes a lambda. I.e. hash for the target would not be accessible in Puppet code. The need for copy on write would then be eliminated.

hlindberg:
If done as thomas suggest with an iterator, or a get_var('var') , it is important that there is no ruby code that mutates the returned values.

lucywyman:
@hlindberg @thallgren Ok, I updated this to use the inventory as a global target_vars store, and created puppet functions `set_var` and `vars` to access + modify it.  

@MikaelSmith I do have one question around why the commented-out spec tests fail (with error `undefined method `set_var' for #<Bolt::Target:0x00000003bebc90>`),  but besides that I think this is ready.

spec/bolt/inventory_spec.rb
```diff
@@ -306,4 +192,23 @@ def targets(names)
       end
     end
   end
+
+  # TODO: Why do these fail?
```
MikaelSmith: The way you've implemented `set_var` and `vars` doesn't create methods on the Bolt::Target Ruby class, only Puppet functions that bind to the Target datatype. So you would have to run those calls through the Puppet parser to do the kind of tests you have.

I would focus on testing the new Puppet functions in boltlib, and the new functions on Bolt::Inventory here.

lib/bolt/cli.rb
```diff
@@ -313,6 +313,7 @@ def parse
       validate(options)
 
       # After validation, initialize inventory and targets. Errors here are better to catch early.
+      @logger.debug("Get targs #{inventory.method(:get_targets)}")
```
MikaelSmith: Why would we call this twice?

lib/bolt/inventory.rb
```diff
@@ -88,16 +92,37 @@ def groups_in(node_name)
...
+
     # Pass a target to get_targets for a public version of this
     # Should this reconfigure configured targets?
     def update_target(target)
-      inv_conf = config_for(target.host)
-      unless inv_conf
-        @logger.debug("Did not find #{target.host} in inventory")
-        inv_conf = {}
+      data = @groups.data_for(target.host)
+      processed = transform_data(data) || {}
```
adreyer: I don't think we want to symbolize keys in the vars. That makes them hard to interact with from puppet.

lib/bolt/inventory.rb
```diff
@@ -88,16 +92,37 @@ def groups_in(node_name)
...
       end
 
-      conf = Bolt::Util.deep_merge(@config.transport_conf, inv_conf)
+      unless processed['vars']
+        @logger.debug("Did not find any variables for #{target.host} in inventory")
+        processed['vars'] = {}
+      end
+
+      tvars = @target_vars[target.host] || {}
+      merged = Bolt::Util.deep_merge(tvars, processed['vars'])
```
adreyer: Won't this overwrite vars a user has set on the target from puppet code next time they call run_*?

spec/bolt/inventory_spec.rb
```diff
@@ -61,120 +61,6 @@ def targets(names)
     }
   }
 
-  describe :config_for do
```
adreyer: these specific search merge behavior tests need to go somewhere unless they were already dupes. Which is possible since I got more systematic about writing them at the end and may not have cleaned everything up

MikaelSmith: I suggested rewriting them as tests using `get_targets`, that verifies the final `Target` objects. We need that anyway because our tests don't match the `Target` contract very well right now.

lib/bolt/inventory.rb
lucywyman: We don't symbolize the keys for vars, but this in place of calling 'config_for' which was originally symbolizing the keys for `data['config']`

lib/bolt/cli.rb
lucywyman: Woops, leftover debugging line. Removed.

bolt-modules/boltlib/lib/puppet/functions/vars.rb
```diff
@@ -0,0 +1,20 @@
...
+  def vars(target)
+    unless Puppet[:tasks]
+      raise Puppet::ParseErrorWithIssue.from_issue_and_stack(
+        Puppet::Pops::Issues::TASK_OPERATION_NOT_SUPPORTED_WHEN_COMPILING, operation: 'set_var'
+      )
+    end
+
+    inventory = Puppet.lookup(:bolt_inventory) { nil }
+
+    inventory.vars(target)
```
thallgren: This will return a mutable hash which is a no-go. It can be resolved in one of three ways.
1. by having set_var perform a "copy on write", i.e. it modifies a cloned hash for the given target. By doing so, anyone holding on to the old hash will not get affected by the set_var call. The target hash should then always be frozen.
2. by letting vars perform a "copy on read", i.e. it returns a copy of the hash which should be frozen.
3. by removing the vars method and instead provide iterative methods to access its entries (as I suggested earlier).

lib/bolt/inventory.rb
```diff
@@ -142,5 +172,22 @@ def expand_targets(targets)
...
+      # data['groups']&.each do |group|
+      # group['nodes']&.each do |node|
+      if data['groups'] && data['nodes']
+        data['groups'].each do |group|
+          next unless group['vars']
+          data['nodes'].each do |node|
+            target = get_targets(node)[0]
+            # Instantiate empty vars hash in case no vars are defined
+            @target_vars[target.host] = @target_vars[target.host] || {}
+            group['vars'].map { |k, v| set_var(target, k, v) }
```
thallgren: This design would make "copy on write" inefficient since it means that each `set_var` call would create a new clone. Could be remedied by using a merge, i.e. `@target_vars[target.host] = @target_vars[target.host].merge(group['vars']).freeze`

bolt-modules/boltlib/lib/puppet/functions/vars.rb
adreyer: We already return mutable hashes/object all over the place. If we decide to change that it probably needs to be it's own effort.

thallgren: The hash in itself may be mutable as long as it is a private copy that cannot be changed by ruby code. This one is intended to be changed for each invocation of set_var. In essence, that's the same thing as introducing a mutable object to in Puppet. We can't do that.

adreyer: We've already done that with Result objects.

thallgren: In that case that was a mistake and the Result objects need to change. It doesn't mean that introducing new similar mistakes is acceptable.

thallgren: where does this happen with the `Result` object? I can't find it.

adreyer: Nothing is frozen about the value hash before it's returned.

thallgren: No, it's not frozen, but it very well could be (and it's a good idea to ensure that it is). There's no way it can be manipulated unless someone would change the error_hash from Ruby code. AFAICS, no one does.

lib/bolt/inventory.rb
```diff
@@ -52,6 +52,8 @@ def initialize(data, config = nil)
       @data = data ||= {}
       @groups = Group.new(data.merge('name' => 'all'))
       @group_lookup = {}
+      @target_vars = {}
+      init_vars(data)
```
lucywyman: todo don't do this, init in update_target

lib/bolt/inventory.rb
```diff
@@ -73,31 +74,56 @@ def get_targets(targets)
...
-    # Should this be a public method?
     def config_for(node_name)
       data = @groups.data_for(node_name)
       if data
         Bolt::Util.symbolize_keys(data['config'])
       end
     end
 
+    def set_var(target, key, value)
+      # TODO: If target_vars is frozen, how do we set values from plans?
```
adreyer: The idea isn't to freeze target_vars but to freeze the hash for each targets. When we want to change a value we merge and reassign.
```ruby
host = target.host
@target_vars[host] = @target_vars[host].merge(key => value).freeze
nil
```

lucywyman: Ah, I see

bolt-modules/boltlib/spec/functions/vars.rb
```diff
@@ -0,0 +1,18 @@
+require 'spec_helper'
+
+describe 'vars' do
+  let(:target) { mock('target') }
+
+  # Do these spec tests even make sense?
```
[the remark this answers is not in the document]
lucywyman: Should I just remove them?

lucywyman:
@hlindberg Still cleaning up the rspec unit tests -- this is my first foray into them in a long time :P Once those are done I think I addressed most of your concerns? Let me know if I should add any more info to the Yard docs, or if there are any other issues!

lucywyman:
Appveyor failure is a 'node connect error', definitely unrelated to this PR.

lib/bolt/inventory.rb
```diff
@@ -73,31 +74,53 @@ def get_targets(targets)
...
+      end
+      data
+    end
+
     # Pass a target to get_targets for a public version of this
     # Should this reconfigure configured targets?
     def update_target(target)
-      inv_conf = config_for(target.name)
-      unless inv_conf
+      data = @groups.data_for(target.host)
```
MikaelSmith: This should remain `target.name`, or we lose the changes in BOLT-337.

lib/bolt/inventory.rb
```diff
@@ -73,31 +74,53 @@ def get_targets(targets)
...
+        processed['config'] = {}
+      end
+
+      unless processed['vars']
+        @logger.debug("Did not find any variables for #{target.host} in inventory")
+        processed['vars'] = {}
       end
 
-      conf = Bolt::Util.deep_merge(@config.transport_conf, inv_conf)
+      set_vars_from_inv(target.name, processed) || {}
```
MikaelSmith: Minor nit: seems clearer to pass `processed['vars']` and have `set_vars_from_inv` be closer to `set_vars_from_hash`.

bolt-modules/boltlib/lib/puppet/functions/set_var.rb
```diff
@@ -0,0 +1,31 @@
+# Sets a variable { key => value } for a target
+#
+# This function takes 3 parameters:
+# * A Target object to set the variable for
+# * The key for the variable (String)
+# * The value of the variable (Any)
+#
+# For time being only accepts a single key-value pair
```
MikaelSmith: This sentence seems unnecessary. It's clear from the parameter.

bolt-modules/boltlib/lib/puppet/functions/set_var.rb
```diff
@@ -0,0 +1,31 @@
+# Sets a variable { key => value } for a target
+#
+# This function takes 3 parameters:
+# * A Target object to set the variable for
+# * The key for the variable (String)
+# * The value of the variable (Any)
```
MikaelSmith: We may expand this to `Any` eventually, but it should match the argument type `Data`.

lib/bolt/inventory/group.rb
```diff
@@ -8,7 +8,6 @@ class Group
       def initialize(data)
         @logger = Logging.logger[self]
         @name = data['name']
-
```
MikaelSmith: You didn't like the grouping? I found it easier to read.

lib/bolt/inventory/group.rb
```diff
@@ -107,6 +109,7 @@ def data_merge(data1, data2)
 
         {
           'config' => Bolt::Util.deep_merge(data1['config'], data2['config']),
+          'vars'   => data2['vars'].merge(data1['vars']),
```
MikaelSmith: We made an explicit choice to do a shallow merge here rather than deep merge? That seems sensible to me, might be worth a comment either here or in the vars function (so others can see how inventory collisions are resolved).

lib/bolt/inventory/group.rb
lucywyman: Ah, I think this was an accident. I'll add it back.

bolt-modules/boltlib/lib/puppet/functions/set_var.rb
```diff
@@ -0,0 +1,30 @@
+# Sets a variable { key => value } for a target
+#
+# This function takes 3 parameters:
+# * A Target object to set the variable for
+# * The key for the variable (String)
+# * The value of the variable (Data)
+#
+# Returns nil
```
hlindberg: The docs comment needs to be adjacent to the call to create_function, now there is a `require` between it and the call.

bolt-modules/boltlib/lib/puppet/functions/vars.rb
```diff
@@ -0,0 +1,28 @@
+# Returns a hash of the 'vars' (variables) assigned to a target through the
+# inventory file or `set_var` function
+#
+# Accepts no parameters
+#
+# Plan authors can call this function on a target to get the variable hash for that
+# target
```
hlindberg: same here - needs to move

bolt-modules/boltlib/lib/puppet/functions/set_var.rb
hlindberg: From the perspective of the puppet language the return is `undef` not `nil` (since `nil` becomes `undef` in puppet).

bolt-modules/boltlib/lib/puppet/functions/vars.rb
```diff
@@ -0,0 +1,28 @@
+# Returns a hash of the 'vars' (variables) assigned to a target through the
+# inventory file or `set_var` function
+#
+# Accepts no parameters
```
hlindberg: picky: missing a `.` at the end of the sentence.

bolt-modules/boltlib/lib/puppet/functions/vars.rb
hlindberg: Missing `.` at end of sentence.

lib/bolt/inventory/group.rb
MikaelSmith: You did it one line down :)

lib/bolt/inventory.rb
```diff
@@ -73,31 +74,53 @@ def get_targets(targets)
...
-    # Should this be a public method?
     def config_for(node_name)
       data = @groups.data_for(node_name)
       if data
         Bolt::Util.symbolize_keys(data['config'])
       end
     end
 
+    def set_var(target, key, value)
+      data = { 'vars' => { key => value } }
```
MikaelSmith: This needs to change to match the updated signature of `set_vars_from_inv`. I'd be in favor of changing the function name to `set_vars_from_hash` as well.

MikaelSmith: I think this is the cause of the test failures.

bolt-modules/boltlib/lib/puppet/functions/set_var.rb
lucywyman: This is true for most of the functions in this repo. Should I change them all in this PR? Does that break yarddocs?

MikaelSmith: That would be nice to fix as an additional maint commit.

bolt-modules/boltlib/lib/puppet/functions/set_var.rb
```diff
@@ -0,0 +1,29 @@
+require 'bolt/error'
+
+# Sets a variable { key => value } for a target
+#
+# This function takes 3 parameters:
+# * A Target object to set the variable for
+# * The key for the variable (String)
+# * The value of the variable (Data)
+#
+# Returns Undef
```
hlindberg: If you want to be precise say "Returns Undef value", or "Returns undef" (since `Undef` is the data type it is possible to misunderstand).
Also `.`

bolt-modules/boltlib/lib/puppet/functions/vars.rb
```diff
@@ -0,0 +1,27 @@
+require 'bolt/error'
+
+# Returns a hash of the 'vars' (variables) assigned to a target through the
+# inventory file or `set_var` function
+#
+# Accepts no parameters.
+#
+# Plan authors can call this function on a target to get the variable hash for that.
+# target
```
hlindberg: `.`

bolt-modules/boltlib/lib/puppet/functions/vars.rb
```diff
@@ -0,0 +1,27 @@
+require 'bolt/error'
+
+# Returns a hash of the 'vars' (variables) assigned to a target through the
+# inventory file or `set_var` function
```
hlindberg: `.`

bolt-modules/boltlib/lib/puppet/functions/set_var.rb
```diff
@@ -0,0 +1,29 @@
+require 'bolt/error'
+
+# Sets a variable { key => value } for a target
```
hlindberg: `.`

PR #310, opened 2018-02-23
(BOLT-336) Treat --nodes as the nodes parameter for run plan

mruzicka:
This change makes it possible to specify the `nodes` parameter of a plan by passing the `--nodes` option(s). (If more than one is passed then the values they specify are concatenated comma delimited.)

It is an error to specify both the `--nodes` option(s) and the `nodes` parameter. It is also an error to specify the `--nodes` option when running a plan which doesn't take the `nodes` parameter. The specified value *is* however passed to the plan as the `nodes` parameter, but that fails as that parameter is not expected by the plan.

While the use of the `--nodes` option in the way enabled by this commit is allowed it is not encouraged and so the option is not printed in the help messages related to `bolt plan` commands produced by `bolt help`. This necessitated creation of a subclass of the `OptionParser` used for producing of these messages with support for excluding selected options from those messages on demand.

### DIFF
--- a/lib/bolt/cli.rb
+++ b/lib/bolt/cli.rb
@@ -24,7 +24,8 @@ module Bolt
   class CLIExit < StandardError; end
 
   class CLI
-    BANNER = <<-HELP.freeze
+    class BoltOptionParser < OptionParser
+      BANNER = <<-HELP.freeze
 Usage: bolt <subcommand> <action> [options]
 
 Available subcommands:
@@ -39,9 +40,9 @@ Available subcommands:
     bolt file upload <src> <dest>    Upload a local file
 
 where [options] are:
-HELP
+      HELP
 
-    TASK_HELP = <<-HELP.freeze
+      TASK_HELP = <<-HELP.freeze
 Usage: bolt task <action> <task> [options] [parameters]
 
 Available actions are:
@@ -52,27 +53,27 @@ Available actions are:
 Parameters are of the form <parameter>=<value>.
 
 Available options are:
-HELP
+      HELP
 
-    COMMAND_HELP = <<-HELP.freeze
+      COMMAND_HELP = <<-HELP.freeze
 Usage: bolt command <action> <command> [options]
 
 Available actions are:
     run                              Run a command remotely
 
 Available options are:
-HELP
+      HELP
 
-    SCRIPT_HELP = <<-HELP.freeze
+      SCRIPT_HELP = <<-HELP.freeze
 Usage: bolt script <action> <script> [[arg1] ... [argN]] [options]
 
 Available actions are:
     run                              Upload a local script and run it remotely
 
 Available options are:
-HELP
+      HELP
 
-    PLAN_HELP = <<-HELP.freeze
+      PLAN_HELP = <<-HELP.freeze
 Usage: bolt plan <action> <plan> [options] [parameters]
 
 Available actions are:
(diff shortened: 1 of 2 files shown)


lib/bolt/cli.rb
```diff
@@ -83,186 +84,232 @@ class CLI
...
+      def initialize(options)
+        super()
+
+        @options = options
+
+        @nodes = define('-n', '--nodes NODES',
+                        'Node(s) to connect to in URI format [protocol://]host[:port] (Optional)',
+                        'Eg. --nodes bolt.puppet.com',
+                        'Eg. --nodes localhost,ssh://nix.com:2222,winrm://windows.puppet.com',
+                        "\n", # the help message won't print if replaced with an empty string
```
nicklewis: I'm not sure what this is trying to indicate.

mruzicka: I'll make the comment clearer.

lib/bolt/cli.rb
```diff
@@ -313,38 +363,16 @@ def parse
       validate(options)
 
       # After validation, initialize inventory and targets. Errors here are better to catch early.
-      options[:targets] = inventory.get_targets(options[:nodes]) if options[:nodes]
+      unless options[:action] == 'show' || options[:mode] == 'plan'
```
MikaelSmith: This is starting to get ugly, but I'm ok leaving refactoring it for later.

lib/bolt/cli.rb
```diff
@@ -443,13 +471,21 @@ def execute(options)
       message = 'There may be processes left executing on some nodes.'
 
       if options[:mode] == 'plan'
-        executor = Bolt::Executor.new(@config, options[:noop], true)
+        unless options[:nodes].empty?
+          if options[:task_options]['nodes']
+            raise Bolt::CLIError,
+                  "plan's nodes parameter may be specified using the --nodes option, but in that case " \
```
MikaelSmith: Based on formatting for other errors, I'd rework this to say: `A plan's 'nodes' parameter may be specified using the --nodes option, but in that case it must not be specified as a separate nodes=<value> parameter`.

lib/bolt/cli.rb
mruzicka: I was tempted to move it to the `execute` function, but held back because of the comment.

lib/bolt/cli.rb
```diff
@@ -443,13 +475,21 @@ def execute(options)
       message = 'There may be processes left executing on some nodes.'
 
       if options[:mode] == 'plan'
-        executor = Bolt::Executor.new(@config, options[:noop], true)
+        unless options[:nodes].empty?
+          if options[:task_options]['nodes']
+            raise Bolt::CLIError,
+                  "A plan's 'nodes' parameter may be specified using the --nodes option, but in that " \
+                  "case it must not be specified as a separate nodes=<value> parameter."
```
mruzicka: I have updated the message as requested, but I'm unsure whether we should somehow indicate that the 'nodes'  must not be specified in the JSON passed to the `--params` option either when already specified using the `--nodes` options.

mruzicka:
Rebased on latest master, while at it I've streamlined the options parsing logic slightly.

mruzicka:
I've added a handler for the `OptionParser::InvalidArgument` exception.

mruzicka:
Rebased on latest master.

PR #312, opened 2018-02-23
(BOLT-346) Support different inventory orchestrator config

adreyer:
The orchestrator transport will now batch based on service-url,
token-file in addition to task-environment. Instead of reading config on
initialization is used the config for the first target of each batch to
create a client.

### DIFF
--- a/lib/bolt/transport/orch.rb
+++ b/lib/bolt/transport/orch.rb
@@ -11,15 +11,18 @@ module Bolt
       CONF_FILE = File.expand_path('~/.puppetlabs/client-tools/orchestrator.conf')
       BOLT_MOCK_TASK = Struct.new(:name, :executable).new('bolt', 'bolt/tasks/init').freeze
 
-      def initialize(config)
-        super
-
+      def create_client(opts)
         client_keys = %i[service-url token-file cacert]
-        @client_opts = config.select { |k, _v| client_keys.include?(k) }
-      end
+        client_opts = opts.reduce({}) do |acc, (k, v)|
+          if client_keys.include?(k)
+            acc.merge(k.to_s => v)
+          else
+            acc
+          end
+        end
+        @logger.debug("Creating orchestrator client for #{client_opts}")
 
-      def create_client
-        OrchestratorClient.new(@client_opts, true)
+        OrchestratorClient.new(client_opts, true)
       end
 
       def build_request(targets, task, arguments)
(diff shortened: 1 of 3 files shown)


lib/bolt/transport/orch.rb
```diff
@@ -124,9 +131,13 @@ def run_task_job(targets, task, arguments)
         end
 
         begin
-          results = create_client.run_task(body)
+          results = create_client(targets.first.options).run_task(body)
```
MikaelSmith: This might eventually benefit from some connection pooling, similar to the other transports. Seems fine to leave that until later.

lib/bolt/transport/orch.rb
```diff
@@ -113,7 +116,11 @@ def batch_upload(targets, source, destination, _options = {}, &callback)
       end
 
       def batches(targets)
-        targets.group_by { |target| target.options[:orch_task_environment] }.values
+        targets.group_by do |target|
+          [target.options[:orch_task_environment],
+           target.options[:"service-url"],
+           target.options[:"token-file"]].join(' * ')
```
MikaelSmith: I don't think you need to join, group_by can handle arrays as keys.

MikaelSmith: Is the idea to squash out nil values?

MikaelSmith: I guess if it's worth keeping, I'd like a comment why it's useful over grouping by the tuple of those values.

adreyer: at one point I was logging it. I can remove it now

lib/bolt/transport/orch.rb
adreyer: That will require work in the client itself as well

lib/bolt/transport/orch.rb
```diff
@@ -124,9 +131,13 @@ def run_task_job(targets, task, arguments)
         end
 
         begin
-          results = create_client.run_task(body)
+          results = create_client(targets.first.options).run_task(body)
 
           process_run_results(targets, results)
+        rescue OrchestratorClient::ApiError => e
+          targets.map do |target|
+            Bolt::Result.new(target, error: e.data)
```
nicklewis: Should this be `Bolt::Result.from_exception`?

nicklewis: Oh, because we want to unwrap the exception data, I see.

PR #313, opened 2018-02-23
(BOLT-337) Nodes in inventory accept full URI syntax

MikaelSmith:
Node names in inventory now accept full URIs. To use config from a node
in inventory, you will need to exactly match the name in inventory when
specifying the node on the command-line or in a plan.

### DIFF
--- a/spec/bolt/inventory_spec.rb
+++ b/spec/bolt/inventory_spec.rb
@@ -52,7 +52,7 @@ describe Bolt::Inventory do
               'config' => {
                 'ssh' => { 'user' => 'someone' }
               } },
-            'node7', 'node8'
+            'node7', 'ssh://node8'
           ],
           'config' => { 'ssh' => {
             'insecure' => 'maybe'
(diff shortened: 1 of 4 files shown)


spec/bolt/inventory_spec.rb
```diff
@@ -290,13 +294,17 @@ def targets(names)
       end
 
       it 'should split a comma-separated list of target URI and group name' do
-        targets = inventory.get_targets('group1,a')
-        expect(targets).to eq(targets(%w[node4 node5 node6 node7 a]))
+        matched_nodes = %w[node4 node5 node6 node7 ssh://node8]
+        matched_nodes.each do |node|
+          expect_any_instance_of(Bolt::Inventory).to receive(:config_for).with(node)
```
MikaelSmith: I'd like to do some more cleanup and check the returned Targets rather than watch for `config_for`, but that overlaps with some larger cleanup around these tests that @lucywyman's working on in https://github.com/puppetlabs/bolt/pull/307.

PR #315, opened 2018-02-23
(BOLT-345) Provide summary of successful and failed nodes

MikaelSmith:
Provides a summary at the end of a run of which nodes succeeded and
failed that can be copied and re-run via `--nodes`.


MikaelSmith:
This makes me want a `--quiet` option more.

PR #319, opened 2018-02-27
(maint) Add license_finder to Bolt.

shrug:
Using PDK project as a model, this adds the license_finder
gem and a task to the test gemset. Also adds all current
dependencies to the "approved" list, and adds a travis check
to validate dependency licenses.

### DIFF
--- a/.dependency_decisions.yml
+++ b/.dependency_decisions.yml
@@ -1,0 +1,61 @@
+---
+- - :approve
+  - CFPropertyList
+  - :who: bradejr
+    :why: 
+    :versions: []
+    :when: 2018-02-26 23:47:22.727579000 Z
+- - :whitelist
+  - MIT
+  - :who: bradejr
+    :why: 
+    :versions: []
+    :when: 2018-02-26 23:48:16.400477000 Z
+- - :whitelist
+  - Apache 2.0
+  - :who: bradejr
+    :why: 
+    :versions: []
+    :when: 2018-02-26 23:48:48.021244000 Z
+- - :whitelist
+  - ruby
+  - :who: bradejr
+    :why: 
+    :versions: []
+    :when: 2018-02-26 23:49:17.135760000 Z
+- - :whitelist
+  - Ruby
+  - :who: bradejr
+    :why: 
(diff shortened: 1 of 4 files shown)


shrug:
This basically duplicates the setup in https://github.com/puppetlabs/pdk - Do we want to add this to Bolt? See any downside?

MikaelSmith:
I'm not against it. Rubocop seems to be complaining about a few things on your PR.

Rakefile
```diff
@@ -32,6 +32,11 @@ end
 desc "Run tests and style checker"
 task test: %w[spec rubocop]
 
+desc 'Check for unapproved licenses in dependencies'
+task(:license_finder) do
+  system('license_finder --decisions-file=.dependency_decisions.yml') || raise(StandardError, 'Unapproved license(s) found on dependencies')
```
mwaggett: Is there a way for the error to specify which dependencies/licenses are a problem?

.dependency_decisions.yml
```diff
@@ -0,0 +1,61 @@
+---
```
mwaggett: I assume this was generated by the license finder, but man, this is kind of hard to parse with human eyes..

Rakefile
nicklewis: The output of `license_finder` includes that information, so it will be printed anyway.

.travis.yml
```diff
@@ -4,6 +4,9 @@ cache: bundler
 rvm:
 - 2.4
 - 2.0
+env:
+  - CHECK = spec
+  - CHECK = license_manager
```
nicklewis: This should be `license_finder`

nicklewis: And I think this should just be added to the `bundle exec rake travisci rubocop` invocation. Currently it's trying to run the check from within the `bolt-modules/boltlib` directory rather than the top-level, and it's also running all the other tests before doing the license check (while also running them all separately).

nicklewis:
Whoops, I accidentally selected approved instead of request changes. :)

adreyer:
@shrug Is this still needed?

PR #323, opened 2018-02-27
(BOLT-352, BOLT-358) Fix config file and inventory file config bugs

MikaelSmith:
Moves task_environment to be specific to the pcp transport, and fixes specifying token-file and service-url in the Bolt config file (it worked correctly in the inventory file).

Fixes the inventory file config section so it uses structure and keys
that are consistent with the config file format. Previously it required
them to match the internal structure of the Bolt::Config object, which
led to a lot of inconsistency between inventory and config file formats.

So
```
config:
  transports:
    ssh:
      host_key_check: false
```
is now
```
config:
  ssh:
    host-key-check: false
```

### DIFF
--- a/lib/bolt/config.rb
+++ b/lib/bolt/config.rb
@@ -22,7 +22,7 @@ module Bolt
 
     TRANSPORT_OPTIONS = %i[host_key_check password run_as sudo_password extensions
                            ssl key tty tmpdir user connect_timeout cacert
-                           token-file task_environment service-url].freeze
+                           token-file task-environment service-url].freeze
 
     TRANSPORT_DEFAULTS = {
       connect_timeout: 10,
@@ -37,7 +37,7 @@ module Bolt
         ssl: true
       },
       pcp: {
-        task_environment: 'production'
+        :"task-environment" => 'production'
       }
     }.freeze
 
@@ -168,7 +168,7 @@ module Bolt
           self[:transports][:pcp][:"token-file"] = data['pcp']['token-file']
         end
         if data['pcp']['task-environment']
-          self[:transports][:pcp][:task_environment] = data['pcp']['task-environment']
+          self[:transports][:pcp][:"task-environment"] = data['pcp']['task-environment']
         end
       end
     end
@@ -192,7 +192,7 @@ module Bolt
 
       TRANSPORT_OPTIONS.each do |key|
         TRANSPORTS.each do |transport|
-          unless %i[ssl host_key_check task_environment].any? { |k| k == key }
+          unless %i[ssl host_key_check task-environment].any? { |k| k == key }
             self[:transports][transport][key] = options[key] if options[key]
             next
           end
--- a/lib/bolt/config.rb
+++ b/lib/bolt/config.rb
@@ -71,6 +71,15 @@ module Bolt
       end
     end
 
+    def deep_clone
+      copy = clone
+      copy[:transports] = self[:transports].clone
+      TRANSPORTS.each do |transport|
+        copy[:transports][transport] = self[:transports][transport].clone
+      end
+      copy
+    end
+
     def default_paths
       root_path = File.expand_path(File.join('~', '.puppetlabs'))
       [File.join(root_path, 'bolt.yaml'), File.join(root_path, 'bolt.yml')]
@@ -168,6 +177,7 @@ module Bolt
         end
       end
     end
+    private :update_from_file
 
     def load_file(path)
       data = Bolt::Util.read_config_file(path, default_paths, 'config')
@@ -201,6 +211,27 @@ module Bolt
             next
           end
         end
+      end
+    end
+
+    def update_from_inventory(data)
+      update_from_file(data)
+
+      if data['transport']
+        self[:transport] = data['transport']
+      end
+
+      # Add options that aren't allowed in a config file, but are allowed in inventory
+      %w[user password port].each do |opt|
+        (TRANSPORTS - [:pcp]).each do |transport|
+          if data[transport.to_s] && data[transport.to_s][opt]
+            self[:transports][transport][opt.to_sym] = data[transport.to_s][opt]
+          end
+        end
+      end
+
+      if data['ssh'] && data['ssh']['sudo-password']
+        self[:transports][:ssh][:sudo_password] = data['ssh']['sudo-password']
       end
     end
 
--- a/lib/bolt/config.rb
+++ b/lib/bolt/config.rb
@@ -232,6 +232,16 @@ module Bolt
                      "user to escalate to with --run-as")
       end
 
+      host_key = self[:transports][:ssh][:host_key_check]
+      unless !!host_key == host_key
+        raise Bolt::CLIError, 'host-key-check option must be a Boolean true or false'
+      end
+
+      ssl_flag = self[:transports][:winrm][:ssl]
+      unless !!ssl_flag == ssl_flag
+        raise Bolt::CLIError, 'ssl option must be a Boolean true or false'
+      end
+
       self[:transports].each_value do |v|
         timeout_value = v[:connect_timeout]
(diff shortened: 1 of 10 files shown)


MikaelSmith:
@adreyer can we turn the ad-hoc perf testing you've done into a benchmarking script? This may impact it.

lib/bolt/config.rb
```diff
@@ -159,7 +170,7 @@ def update_from_cli(options)
 
       TRANSPORT_OPTIONS.each do |key|
         TRANSPORTS.each do |transport|
-          unless %i[ssl host_key_check].any? { |k| k == key }
+          unless %i[ssl host_key_check task_environment].any? { |k| k == key }
```
adreyer: do we accept task environment from the cli?

lib/bolt/config.rb
```diff
@@ -127,19 +137,20 @@ def update_from_file(data)
...
         end
         if data['pcp']['cacert']
           self[:transports][:pcp][:cacert] = data['pcp']['cacert']
         end
         if data['pcp']['token-file']
-          self[:transports][:pcp][:token_file] = data['pcp']['token-file']
+          self[:transports][:pcp][:"token-file"] = data['pcp']['token-file']
         end
         if data['pcp']['task-environment']
-          self[:transports][:pcp][:orch_task_environment] = data['pcp']['task-environment']
```
adreyer: oh hmm I wasn't properly testing task environment.

This doesn't get '-' because it's not passed on the client?

lib/bolt/config.rb
```diff
@@ -65,6 +66,15 @@ def initialize(**kwargs)
       end
     end
 
+    def deep_clone
```
adreyer: this should work for now but it may cause a bug in the future by not cloning arrays or any hashes we add later. It may be better to have this be more generic if it's not too expensive

lib/bolt/config.rb
```diff
@@ -176,6 +187,27 @@ def update_from_cli(options)
       end
     end
 
+    def update_from_inventory(data)
+      update_from_file(data)
```
adreyer: do we want to copy all the non-transport config onto the target? I guess it doesn't do anything and isn't accessible from puppet code

lib/bolt/config.rb
MikaelSmith: Not that I could find.

lib/bolt/config.rb
MikaelSmith: Right now they're ignored, and it's a couple strings (so cost of cloning should be low). That said, I'd like to benchmark how this all affects searching a large inventory.

lib/bolt/config.rb
MikaelSmith: Sounds reasonable. I'm not quite sure what a generic implementation will look like yet...

lib/bolt/config.rb
MikaelSmith: I could make it consistent with the other two. `token-file` and `service-urls` aren't accessible on the CLI either.

It probably doesn't get a dash because it's not available in the Orchestrator client config.

lib/bolt/inventory.rb
```diff
@@ -97,8 +96,12 @@ def update_target(target)
         inv_conf = {}
       end
 
-      conf = Bolt::Util.deep_merge(@config.transport_conf, inv_conf)
-      target.update_conf(conf)
+      # Use Config object to ensure config section is treated consistently with config file
+      conf = @config.deep_clone
+      conf.update_from_inventory(inv_conf)
```
MikaelSmith: I haven't made this backward compatible. I'd be tempted to have Inventory error if it finds a `transports` property in `config` and require people to update their inventory files.

lib/bolt/config.rb
adreyer: "It probably doesn't get a dash because it's not available in the Orchestrator client config."

thats what I was getting at. i was more curious about the reasoning.

lib/bolt/inventory.rb
adreyer: ok

lib/bolt/inventory/group.rb
```diff
@@ -32,6 +32,14 @@ def initialize(data)
         @rest = data.reject { |k, _| %w[name nodes config groups].include? k }
       end
 
+      def check_deprecated_config(context, name, config)
```
MikaelSmith: Added validation that `config: transports` isn't used @adreyer @katelopresti 

Curious what you think about having Bolt link directly to the docs site.

lib/bolt/config.rb
MikaelSmith: Many of the deep_* implementations rely on monkey-patching Object, such as https://github.com/rubyworks/facets/blob/27d9b7f5b3a507da17c0f48c1a7a95079359470d/lib/core/facets/kernel/deep_clone.rb. https://github.com/rubyworks/facets/blob/27d9b7f5b3a507da17c0f48c1a7a95079359470d/work/rejected/kernel/deep_clone.rb is an interesting option that might not require monkey-patching.

MikaelSmith: Added a general deep_clone implementation.

lib/bolt/config.rb
```diff
@@ -127,19 +137,20 @@ def update_from_file(data)
...
         if data['pcp']['cacert']
           self[:transports][:pcp][:cacert] = data['pcp']['cacert']
         end
         if data['pcp']['token-file']
-          self[:transports][:pcp][:token_file] = data['pcp']['token-file']
+          self[:transports][:pcp][:"token-file"] = data['pcp']['token-file']
         end
         if data['pcp']['task-environment']
-          self[:transports][:pcp][:orch_task_environment] = data['pcp']['task-environment']
+          self[:transports][:pcp][:task_environment] = data['pcp']['task-environment']
```
lucywyman: Nit: Should this be task-environment for consistency?

MikaelSmith: It doesn't have to be (the others are written so to simplify https://github.com/puppetlabs/bolt/blob/master/lib/bolt/transport/orch.rb#L15-L25). I can change it easily enough.

lib/bolt/inventory.rb
```diff
@@ -73,20 +73,19 @@ def get_targets(targets)
...
-
     #### PRIVATE ####
     #
     # For debugging only now
     def groups_in(node_name)
       @groups.data_for(node_name)['groups'] || {}
     end
+    private :groups_in
+
+    def config_for(node_name)
```
lucywyman: I could be wrong but I don't believe this is being called anywhere. I think you can just remove it if you refactor the tests.

lib/bolt/config.rb
MikaelSmith: Done.

lib/bolt/util.rb
```diff
@@ -43,10 +43,37 @@ def deep_merge(hash1, hash2)
         hash1.merge(hash2, &recursive_merge)
       end
 
-      def symbolize_keys(hash)
-        hash.each_with_object({}) do |(k, v), acc|
-          k = k.to_sym
-          acc[k] = v.is_a?(Hash) ? symbolize_keys(v) : v
+      # Performs a deep_clone, using an identical copy if the cloned structure contains multiple
+      # references to the same object and prevents endless recursion.
+      # Credit to Jan Molic via https://github.com/rubyworks/facets
```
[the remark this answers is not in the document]
MikaelSmith: Done.

lib/bolt/inventory.rb
MikaelSmith: The `groups_in`? `config_for` is still used until your PR is merged.

lib/bolt/inventory/group.rb
lucywyman: +1 I found this helpful

lucywyman:
Re-reviewed, still looks good!

PR #325, opened 2018-02-28
(maint) bolt-pdb-inventory shouldn't error without config

adreyer:


MikaelSmith:
Wen would this be useful? And how did it fail before?

MikaelSmith:
Now Rubocop's complaining.

PR #328, opened 2018-03-01
(BOLT-363) Fix batching by protocol

MikaelSmith:
Previously, running a command across targets with different protocols
would try to reach the target on each protocol. Now correctly group by
protocols and only try to reach a particular target by its specified
protocol.

### DIFF
--- a/lib/bolt/executor.rb
+++ b/lib/bolt/executor.rb
@@ -58,9 +58,9 @@ module Bolt
     # transport, is yielded to the block in turn and the results all collected
     # into a single ResultSet.
     def batch_execute(targets)
-      promises = targets.group_by(&:protocol).flat_map do |protocol, _protocol_targets|
+      promises = targets.group_by(&:protocol).flat_map do |protocol, protocol_targets|
         transport = transport(protocol)
-        transport.batches(targets).flat_map do |batch|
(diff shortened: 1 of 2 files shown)


lib/bolt/executor.rb
```diff
@@ -58,9 +58,9 @@ def summary(action, object, result)
     # transport, is yielded to the block in turn and the results all collected
     # into a single ResultSet.
     def batch_execute(targets)
-      promises = targets.group_by(&:protocol).flat_map do |protocol, _protocol_targets|
+      promises = targets.group_by(&:protocol).flat_map do |protocol, protocol_targets|
         transport = transport(protocol)
-        transport.batches(targets).flat_map do |batch|
+        transport.batches(protocol_targets).flat_map do |batch|
           batch_promises = Hash[Array(batch).map { |target| [target, Concurrent::Promise.new(executor: :immediate)] }]
```
mruzicka: could be simplified using `each_with_object`:
```ruby
batch_promises = Array(batch).each_with_object({}) { |target, h| h[target] = Concurrent::Promise.new(executor: :immediate) }
```

MikaelSmith: Done.

PR #330, opened 2018-03-08
(BOLT-167) Query PuppetDB for targets with --query flag

nicklewis:
This adds a --query flag to specify a PQL query to use to determine targets.
The query will be executed against the PuppetDB instance configured in
bolt.yaml, falling back to ~/.puppetlabs/client-tools/puppetdb.conf if it
exists (thereby working automatically for users who have PE client tools
configured). In order to serve as targets, the query results must include a
'certname' field, which will be used as the target name. Targets are then
looked up in the inventory to find connection information.

This flag is mutually exclusive with the --nodes flag. Unlike --nodes, it
can only be specified once.


MikaelSmith:
What does the PuppetDB-specific config look like? Might be appropriate to add to https://github.com/puppetlabs/bolt/#configuring-bolt.

lib/bolt/cli.rb
```diff
@@ -129,6 +130,10 @@ def initialize(options)
                         '* port defaults to `5985` or `5986` for WinRM, based on the --[no-]ssl setting') do |nodes|
           @options[:nodes] << get_arg_input(nodes)
         end.extend(SwitchHider)
+        @query = define('-q', '--query QUERY',
```
MikaelSmith: Having query be specific to PuppetDB might bite us later. But it certainly seems like the best option for usability right now.

lib/bolt/cli.rb
```diff
@@ -412,8 +425,12 @@ def validate(options)
         end
       end
 
-      if options[:nodes].empty? && options[:mode] != 'plan' && options[:action] != 'show'
-        raise Bolt::CLIError, "Option '--nodes' must be specified"
+      if options[:mode] != 'plan' && options[:action] != 'show'
```
MikaelSmith: We didn't really address it in the ticket, so it's probably better to skip it. However I might expect `--query` to work similar to `--nodes` for a plan.

nicklewis:
Pushed a commit adding some config/usage docs.

lib/bolt/cli.rb
nicklewis: I agree with both of those points. I think I would prefer to keep it straightforward now and complicate it later if necessary.

lib/bolt/cli.rb
```diff
@@ -432,6 +449,18 @@ def handle_parser_errors
       raise Bolt::CLIError, "Unknown argument '#{e.args.first}'"
     end
 
+    def puppetdb_client
+      return @puppetdb_client if @puppetdb_client
+      puppetdb_config = Bolt::PuppetDB::Config.new(nil, config)
```
MikaelSmith: This seems to result in `Could not retrieve targets from PuppetDB: no implicit conversion of Bolt::Config into Hash`

lib/bolt/cli.rb
nicklewis: I was considering that, but decided that since the reason for `--nodes` for plans is just uniformity during the learning experience, it doesn't seem as obvious that `--query` should behave the same.

lib/bolt/cli.rb
MikaelSmith: It looks like you should be able to explicitly call `to_h` on a struct and get a hash.

nicklewis: Whoops, bad rebase. That should be `config.puppetdb`...

README.md
```diff
@@ -166,6 +166,30 @@ $ bolt command run hostname --nodes 'sshnode*.example.com'
...
 
+### Querying PuppetDB for targets
+
+In addition to the inventory file, you can query PuppetDB to generate lists of targets.
+
+First, add the necessary configuration to `~/.puppetlabs/bolt.yaml`.
+
+```yaml
+puppetdb:
+  server_urls: ["https://puppetdb.example.com:8081"]
```
MikaelSmith: This doesn't seem to work for me.

MikaelSmith: Also, I know it's consistent with puppetdb.conf, but it's inconsistent with other Bolt config. *sigh*

MikaelSmith: Might need to pull the `puppetdb` section out of config before passing it to the PuppetDB::Config object.

lib/bolt/cli.rb
MikaelSmith: Yeah, that makes sense.

README.md
nicklewis: Pushed a fix to actually pass the puppetdb section

MikaelSmith: Is that something we can test very well? I guess this is best tested against a running PuppetDB.

PR #332, opened 2018-03-09
(BOLT-356) Add pcp/local-validation setting

mruzicka:
Adds support for the `pcp/local-validation` configuration setting.
When set to `false` then, under the condition that a task is run on targets exclusively using the `pcp` transport, no validation of the specified task parameters against the local task definition is performed (the definition is not even loaded so the task doesn't even need to be installed locally).
The validation still takes place if the task is run on targets which use different transports (which may or may not include the `pcp` transport).

### DIFF
--- a/bolt-modules/boltlib/lib/puppet/functions/run_task.rb
+++ b/bolt-modules/boltlib/lib/puppet/functions/run_task.rb
@@ -41,12 +41,6 @@ Puppet::Functions.create_function(:run_task) do
       )
     end
 
-    # TODO: use the compiler injection once PUP-8237 lands
-    task_signature = Puppet::Pal::ScriptCompiler.new(closure_scope.compiler).task_signature(task_name)
-    if task_signature.nil?
-      raise with_stack(:UNKNOWN_TASK, Bolt::Error.unknown_task(task_name))
-    end
-
     executor = Puppet.lookup(:bolt_executor) { nil }
     inventory = Puppet.lookup(:bolt_inventory) { nil }
     unless executor && inventory && Puppet.features.bolt?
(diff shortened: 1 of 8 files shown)


mruzicka:
Fixed the boltlib tests.

lib/bolt/cli.rb
```diff
@@ -240,28 +244,32 @@ def initialize(options)
...
 
       def update
+        # show the --nodes and --query switches by default
+        @nodes.hide = @query.hide = false
+        # hide the --[no-]local-validation switch by default
+        @local_validation.hide = true
+
         # Update the banner according to the mode
         self.banner = case @options[:mode]
                       when 'plan'
```
adreyer: I think we need to expose local_validation here too.

lib/bolt/cli.rb
```diff
@@ -437,6 +445,11 @@ def validate(options)
         raise Bolt::CLIError,
               "Option '--noop' may only be specified when running a task"
       end
+
+      unless options[:local_validation].nil? || options[:mode] == 'task'
```
adreyer: We need to accept this for plans too

mruzicka:
I've turned the command line option into a configuration setting of the `pcp` transport. The setting is not completely encapsulated in the transport though as that would require initializing the `pcp` transport for every task run regardless if any targets of the task run actually need it.

bolt-modules/boltlib/lib/puppet/functions/run_task.rb
```diff
@@ -57,16 +51,33 @@ def run_task_raw(task_name, targets, task_args = nil, &block)
 
     use_args = task_args.reject { |k, _| k.start_with?('_') }
 
-    task_signature.runnable_with?(use_args) do |mismatch_message|
-      raise with_stack(:TYPE_MISMATCH, mismatch_message)
-    end || (raise with_stack(:TYPE_MISMATCH, 'Task parameters do not match'))
+    # Don't bother loading the local task definition if the --no-local-validation
+    # command line option has been specified and all targets use the PCP protocol
+    if executor.local_validation == false && targets.all? { |t| t.protocol == 'pcp' }
```
adreyer: I think we can change the check here to:
`if targets.all? { |t| t.options[:"local-validation"] == false}`

That simplifies the executor and makes this transport option more consistent even if there isn't a good use case for local-validation on only some nodes.

mruzicka:
@adreyer  Updated as per your suggestion.

PR #335, opened 2018-03-13
(BOLT-366) Convert Target options to use string keys

MikaelSmith:
**Freeze options hash:** Ensures Target hash won't change later if someone updates the options
hash.
**Rename vars/set_var spec files to match naming pattern:** We name tests for functions as `<function>_spec`. Update naming of new tests to match.
**Stringify Target options keys:** Uses strings for keys of the Target options hash so they are accessible from the Puppet language. Stringifies all transport-specific config. Also rewrites Config to delegate option filtering to Transports, and became the sole registry of Transport implementations.


spec/bolt/transport/winrm_spec.rb
```diff
@@ -130,25 +131,27 @@ def stub_winrm_to_raise(klass, message)
   end
 
   context "connecting over SSL", winrm: true do
+    let(:target) { make_target(host, ssl_port, ssl_config) }
```
MikaelSmith: Turned out this was no longer testing SSL because config passed to transports was unused. Corrected by passing an SSL config when constructing the Target.

lib/bolt/config.rb
```diff
@@ -251,7 +181,7 @@ def transport_conf
     end
 
     def validate
-      TRANSPORTS.each do |transport|
```
MikaelSmith: This should probably also delegate to transports.

spec/bolt/transport/ssh_spec.rb
```diff
@@ -430,7 +432,8 @@ def result_value(stdout = nil, stderr = nil, exit_code = 0)
 
     context "as non-root" do
       let(:config) {
-        mk_config(host_key_check: false, sudo_password: password, run_as: user, user: user, password: password)
+        mk_config('host-key-check' => false, 'sudo-password' => password, 'run_as' => user,
+                  user: user, password: password)
```
lucywyman: do 'user' and password also need to be 'user' => user ?

MikaelSmith: I pulled some shenanigans in mk_config to stringify keys. I could change them all if that's confusing though.

MikaelSmith:
Fixed rubocop issues.

spec/bolt/transport/ssh_spec.rb
lucywyman: Eh, I don't care too strongly. Just wasn't sure if it would accept both.

lib/bolt/target.rb
```diff
@@ -72,6 +73,11 @@ def protocol
       @uri_obj.scheme || @protocol
     end
 
+    # path is meant to be a temporary property that describes a file path on the target
```
adreyer: Should this be part of this PR since nothing's using it?

MikaelSmith: Oh yeah, I'll drop this out. Was a separate experiment.

MikaelSmith: Updated and rebased.

PR #336, opened 2018-03-13
(BOLT-356) Change local-validation default to false

MikaelSmith:
Simplifies setup for those using Bolt with Orchestrator.

### DIFF
--- a/lib/bolt/config.rb
+++ b/lib/bolt/config.rb
@@ -43,7 +43,7 @@ module Bolt
       },
       pcp: {
         :"task-environment" => 'production',
-        :"local-validation" => true
+        :"local-validation" => false
       },
       local: {}
     }.freeze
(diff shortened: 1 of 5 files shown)


lib/bolt/config.rb
```diff
@@ -43,7 +43,7 @@ module Bolt
       },
       pcp: {
         :"task-environment" => 'production',
-        :"local-validation" => true
+        :"local-validation" => false
```
adreyer: Do we need to set this here or can we just treat undef as false in the code?

MikaelSmith: Ah, we can do that.

MikaelSmith: Well, there's a validation check that asserts it's true or false that needs to be updated if we don't default to false.

PR #338, opened 2018-03-13
(BOLT-377) Use safe_load instead of load

MikaelSmith:
Use `safe_load` to avoid any possible concerns around loading arbitrary YAML.


MikaelSmith:
Ok. Control over environment variables is already a significant security violation, so this isn't really exploitable. We'll update it when we drop Ruby 2.0.

PR #340, opened 2018-03-14
(BOLT-379) Accept key-data in addition to the path of a key file

adreyer:
This allows private key data to be stored directly in config in addition
to the path to private key files.

### DIFF
--- a/lib/bolt/util.rb
+++ b/lib/bolt/util.rb
@@ -43,6 +43,21 @@ module Bolt
         hash1.merge(hash2, &recursive_merge)
       end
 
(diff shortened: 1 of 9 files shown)


lib/bolt/util.rb
```diff
@@ -43,6 +43,21 @@ def deep_merge(hash1, hash2)
         hash1.merge(hash2, &recursive_merge)
       end
 
+      # Accepts a Data object and returns a copy with all hash keys
```
MikaelSmith: +9000. We keep needing then not needing this. Generalized method makes so much sense.

PR #341, opened 2018-03-14
(BOLT-388) Avoid adding 'localhost' to inventory

MikaelSmith:
When added to inventory implicitly, it becomes part of the `all` group.
Instead default `transport` to `local` only if the name `localhost`
isn't found in inventory when creating the Target.

### DIFF
--- a/lib/bolt/inventory.rb
+++ b/lib/bolt/inventory.rb
@@ -56,7 +56,6 @@ module Bolt
       inventory = new(data, config)
       inventory.validate
       inventory.collect_groups
-      inventory.add_localhost
       inventory
     end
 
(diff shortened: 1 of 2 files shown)


lib/bolt/inventory.rb
```diff
@@ -119,10 +108,16 @@ def groups_in(node_name)
...
 
-      unless data['config']
+      unless data
         @logger.debug("Did not find #{target.name} in inventory")
+        data = {}
+        data['config'] = { 'transport' => 'local' } if target.name == 'localhost'
+      end
+
+      unless data['config']
+        @logger.debug("Did not find config for #{target.name} in inventory")
```
adreyer: I think this is overly verbose

MikaelSmith: Ok. Leave out the new debug statement?

MikaelSmith: Removed.

nicklewis:
An alternative would be to have like a `Target::localhost()` method that builds and returns a localhost target instance with its default config set. Then `expand_targets()` would call that (instead of `Target.new`) if the name is "localhost". Any config from the inventory would automatically override 
it in `update_target()`.

I'm not sure if that's any better, but it's different!

MikaelSmith:
I'm not sure that's a useful enough optimization. Also, we don't use the `local` transport if `localhost` is defined in inventory at all (even if it doesn't specify a different transport). So your solution would change the current behavior to require you to explicitly override the transport type for `localhost`.

PR #342, opened 2018-03-14
(BOLT-365) Update nodes usage

MikaelSmith:
Remove `Eg.` as inconsistent docs practice. Update the documentation and example.

New output
```
    -n, --nodes NODES                Node(s) to connect to in URI format or group name (Optional)
                                      --nodes localhost,node_group,ssh://nix.com:2222,winrm://windows.puppet.com

                                     * URI format is [protocol://]host[:port]
                                     * NODES can be a comma-separated list of URI format and/or group names
                                     * NODES can also be '@<file>' to read from a file or '-' to read from stdin
                                     * Windows nodes must specify protocol with winrm:// unless otherwise configured
                                     * protocol is `ssh` by default; may be ssh, winrm, pcp, local
                                     * port defaults to `22` for SSH
                                     * port defaults to `5985` or `5986` for WinRM, based on the --[no-]ssl setting
```

### DIFF
--- a/lib/bolt/cli.rb
+++ b/lib/bolt/cli.rb
@@ -116,20 +116,15 @@ Available options are:
         @options = options
 
         @nodes = define('-n', '--nodes NODES',
-                        'Node(s) to connect to in URI format or group name (Optional)',
-                        ' --nodes localhost,node_group,ssh://nix.com:2222,winrm://windows.puppet.com',
-                        # An empty line in a switch description causes the OptionParser#help
-                        # method to raise an exception. Specifying a line containing only a
-                        # newline is the closest one can get to the empty line without
-                        # triggering that exception.
-                        "\n",
-                        '* URI format is [protocol://]host[:port]',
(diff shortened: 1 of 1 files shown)


MikaelSmith:
@katelopresti thoughts on these updates?

adreyer:
```
  * NODES can either be comma-separated, '@<file>' to read
  * nodes from a file, or '-' to read from stdin
```
wrapping with * confuses me.

```
* Windows nodes must specify protocol with winrm://
* protocol is `ssh` by default; may be any TRANSPORT listed below
* port defaults to `22` for SSH
* port defaults to `5985` or `5986` for WinRM, based on the --[no-]ssl setting
```
especially the first line seems confusing without  "unless otherwise configured"

MikaelSmith:
Updated.

MikaelSmith:
Updated for text specified on the ticket. Also removed `Optional` because it was pretty much never helpful.

PR #343, opened 2018-03-14
(BOLT-389) Initial implementation of WinRM --[no-]ssl-verify

nmaludy:
Currently Bolt doesn't support ignoring SSL verification for use with self-signed certificats.

To support this i've introduced a new command line switch and config option under the WinRM transport called `ssl-verify` that can be used from the CLI with the `--[no-]ssl-verify` option.

In the interest of security i've continued to use the "secure by default" paradigm and make the default value of `--ssl-verify` be `true`

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -57,7 +57,7 @@ To run Bolt from source:
 
 To use `rubocop`, perform the bundle install with no exclusions
 
-    bundle install --path .bundle
+    bundle install --path .bundle --with test
     bundle exec rake rubocop
 
 ## Testing
--- a/README.md
+++ b/README.md
@@ -103,6 +103,8 @@ ssh:
 
 `ssl`: If false, skip requiring SSL for connections. (default: true)
 
+`ssl-verify`: If false, skip remote host SSL certificate verification (for self-signed certs). (default: true)
+
 `cacert`: The CA certificate used to authenticate SSL connections. (default: uses system CA certificates)
 
 `tmpdir`: The directory to store temporary files on the target node. (default: `[System.IO.Path]::GetTempPath()`)
--- a/lib/bolt/cli.rb
+++ b/lib/bolt/cli.rb
@@ -188,6 +188,10 @@ Available options are:
                'Use SSL with WinRM') do |ssl|
           @options[:ssl] = ssl
         end
+        define('--[no-]ssl-verify',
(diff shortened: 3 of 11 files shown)


MikaelSmith:
That looks great! Are you familiar with our ticketing system? I'd like to have a ticket in JIRA for this in the BOLT project.

nmaludy:
@MikaelSmith I've created a New Feature ticket in JIRA here: https://tickets.puppetlabs.com/browse/BOLT-389

MikaelSmith:
Thanks. Looks like there are still a few rubocop issues. Are you able to run it locally with `bundle exec rake rubocop`?

nmaludy:
@MikaelSmith I had to update the `CONTRIBUTING.md` to detail the steps for setting up rubocop, specifically i needed to run `bundle install --path .bundle --with test` after looking at the `Gemfile`

MikaelSmith:
Ah, they didn't take into account the previous bundle run. Thanks for fixing that.

nmaludy:
Here is an example of the command failing when trying to connect to a host with self-signed certs (notice no `--ssl-verify` is passed and it's defaulting to verify, also that a helpful error message is returned):

```shell
$ bolt file upload test.txt 'C:\' --nodes winrm://windowshost.domain.tld --user Administrator --password xxx
Started on windowshost.domain.tld...
Failed on windowshost.domain.tld:
  Failed to connect to https://windowshost.domain.tld:5986/wsman: SSL_connect returned=1 errno=0 state=error: certificate verify failed                                                                                                          
  Is the remote host using a self-signed SSL certificate? Use --no-ssl-verify to disable remote host SSL verification.      
Failed on 1 node: winrm://windowshost.domain.tld
```

Here is an example of the command working by passing in `--no-ssl-verify`:
```shell
$ bolt file upload test.txt 'C:\' --nodes winrm://windowshost.domain.tld --user Administrator --password xxx --no-ssl-verify
Started on windowshost.domain.tld...
Finished on windowshost.domain.tld:
  Uploaded 'test.txt' to 'windowshost.domain.tld:C:\'
Successful on 1 node: winrm://windowshost.domain.tld
Ran on 1 node in 2.69 seconds
```

MikaelSmith:
This would be nice to test against the vagrant Windows VM, but setting that up could take some work. I think I'm ok with the unit test level on this.

PR #346, opened 2018-03-14
(BOLT-260) Parse task parameters

adreyer:
Does this look like the correct place for this logic?
How should we handle remote orchestrator tasks?

### DIFF
--- a/lib/bolt/pal.rb
+++ b/lib/bolt/pal.rb
@@ -48,6 +48,10 @@ module Bolt
           _pcore_init_hash.to_json(opts)
         end
       end
+
+      unless Puppet.settings.global_defaults_initialized?
+        Puppet.initialize_settings
+      end
     end
 
     # Create a top-level alias for TargetSpec so that users don't have to
@@ -65,8 +69,7 @@ module Bolt
     # Runs a block in a PAL script compiler configured for Bolt.  Catches
     # exceptions thrown by the block and re-raises them ensuring they are
     # Bolt::Errors since the script compiler block will squash all exceptions.
-    def in_bolt_compiler(opts = [])
-      Puppet.initialize_settings(opts)
+    def in_bolt_compiler
       r = Puppet::Pal.in_tmp_environment('bolt', modulepath: full_modulepath(@config[:modulepath]), facts: {}) do |pal|
         pal.with_script_compiler do |compiler|
           add_target_spec(compiler)
@@ -108,8 +111,8 @@ module Bolt
 
     def in_plan_compiler(executor, inventory)
       with_bolt_executor(executor, inventory) do
-        with_puppet_settings do |opts|
-          in_bolt_compiler(opts) do |compiler|
+        with_puppet_settings do
+          in_bolt_compiler do |compiler|
             yield compiler
           end
         end
@@ -124,13 +127,16 @@ module Bolt
       end
     end
 
+    # TODO: PUP-8553 should replace this
     def with_puppet_settings
       Dir.mktmpdir('bolt') do |dir|
         cli = []
         Puppet::Settings::REQUIRED_APP_SETTINGS.each do |setting|
           cli << "--#{setting}" << dir
         end
-        yield cli
+        Puppet.settings.send(:clear_everything_for_tests)
+        Puppet.initialize_settings(cli)
+        yield
       end
     end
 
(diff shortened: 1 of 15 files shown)


MikaelSmith:
We could start querying Orchestrator for task metadata. Otherwise if there is no task metadata, we have to assume they're strings.

lib/bolt/pal.rb
```diff
@@ -143,6 +143,35 @@ def list_tasks
       end
     end
 
+    def parse_param(val, type, compiler)
```
MikaelSmith: This kind of seems like it should be in `run_task` and `run_plan`. Do we want to make the same logic work for plans?

MikaelSmith: Counterpoint: we'd end up doing parameter matching even when called from within a plan.

adreyer: Do we want to parse JSON when the functions are called from a plan?

lib/bolt/pal.rb
```diff
@@ -143,6 +143,35 @@ def list_tasks
...
+      # TODO: is there a constant for this in puppet?
+      any_string = compiler.evaluate_string('String[0]')
+      unless any_string.assignable?(type)
```
nicklewis: This is backward, should be `type.assignable?(any_string)` (ie. does the type accept an instance of string?)

lib/bolt/pal.rb
```diff
@@ -143,6 +143,35 @@ def list_tasks
       end
     end
 
+    def parse_param(val, type, compiler)
+      # TODO: is there a constant for this in puppet?
+      any_string = compiler.evaluate_string('String[0]')
```
nicklewis: `Puppet::Pops::Types::TypeFactory.string`

lib/bolt/pal.rb
MikaelSmith: This might be our best option for now. I'd allow for the task to not be found and not do any conversion in that case. We can then say you either need to setup modulepath if you want conversion, and we can work on improving it later if we think of a better way.

lib/bolt/pal.rb
adreyer: That won't work: String[1] will not accept all instances of String[0] an so is not assignable. I don't think there is actually an assignability test we can use that would work for something like:
Optional[Pattern[/<uuid>/]]

adreyer:
I've changed the behavior from what is descibed in the ticket since "assignable" is meant for type simplication and requires "any" instance of a type be assignable. This means there is no way to handle something like "Pattern[/<uuid/]".

The new behavior is to always attempt to parse the parameters string. 
Use the parsed value if it parses and the parsed value is an instance of the type in metadata.
Otherwise pass the original string on.

nicklewis:
This makes sense to me. My only complaint is that it doesn't solve my use case since it doesn't work for plans. :P I see why that's more difficult though. :/

nicklewis:
Actually, scratch that. If we're just doing unconditional JSON parsing irrespective of the specified types of the params, then this should be just as easy to do for plans. In that case, is there a reason not to?

adreyer:
@nicklewis I was planning on doing plans as part of this PR too. The plan_signature is a bit more complex so I put this up first while I worked on how to get the correct param types from the type signature.

We're not doing parsing irrespective of the type. We always try to parse but if the parsed value is not  an instance of the params type we throw it away. This doesn't suffer from the same issue as docs generation for plans because the "simplified" type is acceptable.

lib/bolt/pal.rb
```diff
@@ -143,6 +143,33 @@ def list_tasks
...
+      rescue JSON::ParserError
+        # This value may not be assignable in which case run_* will error
+        str
+      end
+    end
+
+    def parse_task_params(task_name, params, compiler)
+      sig = compiler.task_signature(task_name)
+      # CODEREVIEW: How should we handle the lack of a task signature?
+      # This will primarily be an issue for non-local orch tasks
```
MikaelSmith: The best option I've come up with is to fallback on querying Orchestrator if: no local task is found, and Orchestrator is configured. That doesn't seem ideal though.

lib/bolt/pal.rb
```diff
@@ -143,6 +143,33 @@ def list_tasks
       end
     end
 
+    def parse_param(str, type, compiler)
+      begin
+        parsed = JSON.parse(str, quirks_mode: true)
+        type.instance?(parsed) ? parsed : str
```
MikaelSmith: This seems reasonable, and like something we can reproduce with pcore-java.

lib/bolt/pal.rb
adreyer: I think interacting with the transport API it too rickety.
We should support a remote task store for using orchestrator and use that for validation.

lib/bolt/pal.rb
adreyer: We can reproduce it with pcore-java but that means that anything in puppet-task or the console  needs to be parsed server side.

MikaelSmith: In general this approach seems robust.

nicklewis:
This behavior feels a little unintuitive in practice... I have a plan with a param that requires an integer. If I specify an integer, it works. If I specify a float, it complains that I gave it a string.

Instead of trying to check whether the type accepts "strings" generically, could we instead check whether it will accept the actual string value provided? And if not, *then* proceed to parse it? That may be unintuitive in other ways though...

Realistically, this is probably just a game of whac-a-mole with various edge cases.

MikaelSmith:
I think Float is the primary edge-case I've run across, and it's not that bad to need to add a trailing '.0'. We could handle that case in-particular if we wanted to.

Checking that it accepts the actual string value provided was my original proposal, but it wasn't clear that we could recreate that behavior with pcore-java. Maybe it's worth revisiting that question.

adreyer:
@nicklewis that will prefer a string over anything else which seems wrong in many cases. I think the proper solution is to move validation and error generating into the parsing layer. I wanted to leave that at as separate work however since I'm concerned about edge cases and the orchestrator integration.

I have yet to run into a task that requires a float. Unfortunately as we ran into when using json schema in the task spec: JSON number parsing into languages other than JS is not very consistent.

PR #349, opened 2018-03-15
(BOLT-360) Add facts to targets

lucywyman:


bolt-modules/boltlib/lib/puppet/functions/add_facts.rb
```diff
@@ -0,0 +1,22 @@
...
+  def add_facts(target, facts)
+    unless Puppet[:tasks]
+      raise Puppet::ParseErrorWithIssue.from_issue_and_stack(
+        Puppet::Pops::Issues::TASK_OPERATION_NOT_SUPPORTED_WHEN_COMPILING, operation: 'add_facts'
+      )
+    end
+
+    inventory = Puppet.lookup(:bolt_inventory) { nil }
+
+    inventory.add_facts(target, facts)
```
MikaelSmith: We should test for `nil` inventory before using it. I think other functions raise an error then.

bolt-modules/boltlib/lib/puppet/functions/facts.rb
```diff
@@ -0,0 +1,22 @@
...
+  def facts(target)
+    unless Puppet[:tasks]
+      raise Puppet::ParseErrorWithIssue.from_issue_and_stack(
+        Puppet::Pops::Issues::TASK_OPERATION_NOT_SUPPORTED_WHEN_COMPILING, operation: 'facts'
+      )
+    end
+
+    inventory = Puppet.lookup(:bolt_inventory) { nil }
+
+    inventory.facts(target)
```
MikaelSmith: Also check for `nil` inventory here.

lib/bolt/inventory.rb
```diff
@@ -108,6 +109,15 @@ def vars(target)
       @target_vars[target.name]
     end
 
+    def add_facts(target, new_facts = {})
+      # TODO: Should this warn if facts is empty?
```
MikaelSmith: If `new_facts` is empty? I'm on the fence. I can imagine scenarios where you might update facts, but expect that some are empty and not want to have to separately filter them. But also that empty facts could be a warning of something going wrong.

I guess let's warn for now.

lib/bolt/inventory.rb
```diff
@@ -193,5 +201,13 @@ def set_vars_from_hash(target_name, data)
       end
     end
     private :set_vars_from_hash
+
+    def set_facts(tname, hash)
+      if hash
+        @target_facts[tname] = @target_facts[tname] || {}
```
MikaelSmith: Can be `@target_facts[tname] ||= {}`.

lib/bolt/inventory.rb
```diff
@@ -126,12 +136,10 @@ def update_target(target)
...
 
-      unless data['vars']
-        @logger.debug("Did not find any variables for #{target.name} in inventory")
-        data['vars'] = {}
-      end
-
-      set_vars_from_hash(target.name, data['vars'])
+      # These should only get set from the inventory if they have not yet
+      # been instantiated
+      set_vars_from_hash(target.name, data['vars']) unless @target_vars[target.name]
```
MikaelSmith: So this initializes them the 1st time from the inventory data, then all subsequent lookups will use whatever latest is. 👍

spec/integration/inventory_spec.rb
```diff
@@ -88,6 +91,21 @@
         expect(run_cli_json(var_plan)[0]['result']['stdout']).to eq(output)
       end
     end
+
+    context 'with facts set' do
+      let(:fact_plan) { ['plan', 'run', 'facts', "host=#{target}"] + config_flags }
+      # TODO: ensure deep merge works
```
MikaelSmith: TODO!

lib/bolt/inventory.rb
lucywyman: Yep :P

MikaelSmith: Is there a test that covers this particular fix?

MikaelSmith: Ah, I see it in spec/pal/vars_spec.rb.

lib/bolt/inventory.rb
```diff
@@ -108,6 +109,15 @@ def vars(target)
       @target_vars[target.name]
     end
 
+    def add_facts(target, new_facts = {})
+      @logger.warn("No facts to add")
```
MikaelSmith: This warning should only happen if `new_facts` is empty.

lucywyman: Ack, good catch!

MikaelSmith:
Seems this didn't need a rebase. Going to try to get CI working first though.

PR #350, opened 2018-03-15
(BOLT-389) Add a test for no-ssl-verify

nicklewis:


MikaelSmith:
Seems like it failed on a different test.

PR #351, opened 2018-03-20
(BOLT-361) Add puppetdb_facts function

lucywyman:


lib/bolt/puppetdb/client.rb
```diff
@@ -51,6 +51,34 @@ def query_certnames(query)
         end
       end
 
+      # This method expects an array of certnames to get facts for
+      def puppetdb_fact(certnames)
```
adreyer: facts_for_node?

lucywyman:
(party) This is working locally! Going to figure out a workaround for the `Failure/Error: raise Bolt::PuppetDBError, "server_urls must be specified"` for the spec tests, add some acceptance test asserts, and I think this is GTG

bolt-modules/boltlib/lib/puppet/functions/puppetdb_fact.rb
```diff
@@ -0,0 +1,27 @@
...
+
+  def puppetdb_fact(targets)
+    unless Puppet[:tasks]
+      raise Puppet::ParseErrorWithIssue.from_issue_and_stack(
+        Puppet::Pops::Issues::TASK_OPERATION_NOT_SUPPORTED_WHEN_COMPILING, operation: 'puppetdb_fact'
+      )
+    end
+
+    executor = Puppet.lookup(:bolt_executor) { nil }
+    executor.puppetdb_fact(targets)
```
MikaelSmith: I don't really like the idea of everything being hard-coded into the executor. However it does make sense for some things to be attached to its lifetime.

I'd probably be in favor of a caching mechanism to store and retrieve function-specific data in the executor. I'd probably enforce a namespace (the function name) and key for lookup.

MikaelSmith: I guess in this case we also need access to the global config. Does it make sense to give all functions access to that?

MikaelSmith: I guess this is a generalization problem we can address later when it becomes necessary.

MikaelSmith:
Will you be adding the wrapper plan that calls puppetdb_facts and adds facts to the Target objects?

bolt-modules/boltlib/lib/puppet/functions/puppetdb_fact.rb
```diff
@@ -0,0 +1,27 @@
...
+# Returns a hash of certname to facts objects for each matched Target.
+# This functions accepts a TargetSpec, and returns a hash of target certnames
+# and their corresponding facts hash.
+#
+# * If a node is not found in PuppetDB, it's included in the returned hash with empty facts hash.
+# * Otherwise the node is included in the hash with a value that is a hash of it's facts.
+#
+Puppet::Functions.create_function(:puppetdb_fact) do
+  dispatch :puppetdb_fact do
+    param 'Boltlib::TargetSpec', :targets
```
MikaelSmith: We should require this to be an array of strings (certnames), and handle turning TargetSpec into an array of strings in the wrapper plan.

MikaelSmith: Currently you accept a TargetSpec, but then expect it to be a list of certnames in the executor.

lib/bolt/puppetdb/client.rb
```diff
@@ -51,6 +51,34 @@ def query_certnames(query)
...
+        else
+          parsed = JSON.parse(response.body)
+        end
+
+        # Parse the data
+        parsed&.each do |f|
+          output[f['certname']][f['name']] = f['value']
+        end
+
+        # TODO: I don't think I need this explicit return?
```
MikaelSmith: Otherwise you'd return `parsed`. I think the explicit return is needed.

bolt-modules/boltlib/lib/puppet/functions/puppetdb_fact.rb
lucywyman: I actually initially had this live in the puppetdb client class, but the issue is that then every time `run_plan` is called it inits a connection to puppetdb whether we need it or not. I could refactor the pdb client to just init a 'dummy' connection, then actually initialize it if a function that needs it is called, but this seemed easier + more like "the right way". But would that actually be preferable?

I'm not sure I understand what 'namespace' means in that context? Would the caching just be a hash, and what data would it store? Or did you have something else in mind?

lucywyman:
Re: wrapper plan, yes! I entirely forgot. Thank you!

bolt-modules/boltlib/lib/puppet/functions/puppetdb_fact.rb
MikaelSmith: I meant something like
```
dbclient = executor.cache_lookup(:puppetdb_fact, :client)
unless dbclient
  dbconfig = Bolt::PuppetDB::Config.new(executor.config.puppetdb)
  dbclient = Bolt::PuppetDB::Client.from_config(dbconfig)
  executor.cache_store(:puppetdb_fact, :client, dbclient)
end

dbclient.facts_for_node(certnames)
```

But since we're not exposing config, I don't think we should do that until we actually need it. Hard-coding puppetdb as a thing the executor handles is fine for now.

lucywyman: Ah, ok, I see what you mean. M'k, that seems reasonable.

lucywyman: where "that" is waiting until...we need to limit access to parts of the configuration?

MikaelSmith: I don't think we should change what you have and add support for generalized caching on the executor until we have better use cases. I don't want to open up direct access to config from Bolt functions without a better reason.

bolt-modules/boltlib/lib/puppet/functions/puppetdb_fact.rb
lucywyman: Yep yep. Updated!

bolt-modules/boltlib/lib/puppet/functions/puppetdb_fact.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+require 'bolt/error'
+
+# Returns a hash of certname to facts objects for each matched Target.
+# This functions accepts a TargetSpec, and returns a hash of target certnames
```
MikaelSmith: Description needs an update.

bolt-modules/boltlib/spec/functions/puppetdb_fact.rb
```diff
@@ -0,0 +1,19 @@
...
+  let(:executor) { mock('bolt_executor') }
+  let(:inventory) { mock('inventory') }
+
+  around(:each) do |example|
+    Puppet[:tasks] = true
+    Puppet.features.stubs(:bolt?).returns(true)
+
+    Puppet.override(bolt_executor: executor, bolt_inventory: inventory) do
+      example.run
+    end
```
MikaelSmith: This could use actual spec tests. Which probably require mocking PuppetDB::Client or the executor. So mostly not useful...

modules/puppetdb_fact/plans/init.pp
```diff
@@ -0,0 +1,7 @@
+plan puppetdb_fact(TargetSpec $nodes) {
+  $certnames = get_targets($nodes).map |$target| { $target.uri }
```
MikaelSmith: I think you want `$target.host`. `uri` will frequently not be a valid certname because it can include protocol/port/etc.

bolt-modules/boltlib/spec/functions/puppetdb_fact.rb
lucywyman: I wrote a spec test and determined it wasn't really testing anything? But if you have ideas for a useful one let me know. Or I can add a simple one.

lucywyman:
Update: Spec tests for plans are blocked until Alex's PR lands.

bolt-modules/boltlib/spec/functions/puppetdb_fact.rb
MikaelSmith: I guess a simple one is useful to make sure there are no syntax errors.

modules/puppetdb_fact/plans/init.pp
```diff
@@ -0,0 +1,7 @@
+plan puppetdb_fact(TargetSpec $nodes) {
+  $certnames = get_targets($nodes).map |$target| { $target.host }
+  $pdb_facts = puppetdb_fact(get_targets($nodes))
+  $pdb_facts.each |$node, $node_facts| {
+    add_facts($node, $node_facts)
```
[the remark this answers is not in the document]
MikaelSmith: What I got working was
```
plan puppetdb_fact(TargetSpec $nodes) {
  $targets = get_targets($nodes)
  $certnames = $targets.map |$target| { $target.host }
  $pdb_facts = puppetdb_fact($certnames)
  $targets.each |$target| {
    add_facts($target, $pdb_facts[$target.host])
  }
  return $pdb_facts
}
```

PR #352, opened 2018-03-22
(BOLT-135) Support multiple task implementations

nicklewis:
This change adds logic to the existing transports to dynamically select the
appropriate task implementation based on the available features of a
target. Target now has a select_impl() method which, given a task and an
optional list of "additional" features, will return the path to the first
suitable executable. The "additional" features are passed by the transport
and account for any features which are directly *implied* to be present
based on the transport (ie. 'shell' for ssh and 'powershell' for winrm).

This allows a single task to have multiple implementations for different
platforms and for Bolt to choose the appropriate one for each target at
runtime.

A new set_feature() function in puppet language can be used to add and
remove features on a target.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,7 +1,7 @@
 [submodule "vendored/puppet"]
 	path = vendored/puppet
-	url = https://github.com/puppetlabs/puppet
-	branch = master
+	url = https://github.com/nicklewis/puppet
+  branch = PUP-8588-multiple-task-implementations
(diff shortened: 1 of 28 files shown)


nicklewis:
This depends on puppetlabs/puppet#6758

nicklewis:
I'm also going to poke at some multi-node integration and/or acceptance tests on this...

nicklewis:
Closing this until we decide exactly how we're talking this feature.

PR #353, opened 2018-03-23
(BOLT-407) Ensure TargetSpec is aliased while resolving types

MikaelSmith:
When resolving types to display as part of `bolt plan show`, ensure
`TargetSpec` is still an alias in the compiler so that types using that
name resolve to an existing type.

### DIFF
--- a/spec/integration/parsing_spec.rb
+++ b/spec/integration/parsing_spec.rb
@@ -30,6 +30,26 @@ describe "CLI parses input" do
                          "hash" => { "this" => "that" })
   end
 
+  it 'shows a plan with aliased type' do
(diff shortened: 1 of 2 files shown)


spec/integration/parsing_spec.rb
```diff
@@ -30,6 +30,26 @@
                          "hash" => { "this" => "that" })
   end
 
+  it 'shows a plan with aliased type' do
+    result = run_cli_json(%w[plan show parsing] + config_flags)
+    target_spec_type = "TargetSpec = Boltlib::TargetSpec = Variant[String[1], Object[{name => 'Target', attributes =>"\
```
MikaelSmith: This mess was filed as BOLT-413.

PR #355, opened 2018-03-24
(BOLT-359) Add minifact module to modules

mruzicka:
Adds the `minifact` module which provides several plans (and supporting tasks) for retrieving facts from nodes.

### DIFF
--- a/modules/minifact/plans/retrieve.pp
+++ b/modules/minifact/plans/retrieve.pp
@@ -1,0 +1,46 @@
+# A plan that retrieves facts from the specified nodes by running
+# an apropriate minifact::* task on each. Care is taken to run a
+# minifact::* task corresponding to the node's platfrom on each
+# node. If there is no variant of the task appropriate for some
+# node's platform the node is treated as if the task had failed.
+#
+# The $nodes parameter is a list of the nodes to retrieve the facts
+# from.
+plan minifact::retrieve(TargetSpec $nodes) {
+  $targets = get_targets($nodes)
+
+  # Build a mapping from the names of the tasks to run to the lists of
+  # targets to run the tasks on
+  $task_targets = $targets.minifact::group_by |$target| {
+    $target.protocol ? {
+      'ssh'   => 'minifact::bash',
+      'winrm' => 'minifact::powershell',
+      'pcp'   => 'minifact::ruby',
+      'local' => if (minifact::is_command_available('bash')) { 'minifact::bash' } else { undef },
+      default => undef,
+    }
+  }
+
+  # Return a single result set composed of results from the result sets
+  # returned by the individual task runs (and a result set synthesized
+  # for the unsupported targets)
+  ResultSet(
+    $task_targets.map |$task, $targets| {
+      if ($task == undef) {
+        # Synthesize a result set for the unsupported targets
+        ResultSet($targets.map |$target| {
+          Result($target, '_error' => {
+            'kind' => 'minifact/unsupported',
+            'msg'  => 'Target not supported by minifact.',
+          })
+        })
+      } else {
+        # Run the task on the targets
+        run_task($task, $targets, '_catch_errors' => true)
+      }
+    }.reduce([]) |$results, $result_set| {
+      # Collect the results from the individual result sets
+      $results + $result_set.results
+    }
+  )
+}
--- a/modules/minifact/tasks/powershell.ps1
+++ b/modules/minifact/tasks/powershell.ps1
@@ -1,0 +1,53 @@
+#!powershell.exe
+
+# Delegate to facter if available
+if (Get-Command facter -ErrorAction SilentlyContinue) {
+    facter --json
+} else {
+    if ([System.Environment]::OSVersion.Platform -gt 2) { # [System.PlatformID]::Win32NT
+@'
+{
+  "_error": {
+    "kind": "minfact/noname",
+    "msg": "Could not determine OS name"
+  }
+}
+'@
+    } else {
+        $release = [System.Environment]::OSVersion.Version.ToString() -replace '\.[^.]*\z'
+        $version = $release -replace '\.[^.]*\z'
+
+        # This fails for regular users unless explicitly enabled
+        $os = Get-CimInstance Win32_OperatingSystem -ErrorAction SilentlyContinue
+        $consumerrel = $os.ProductType -eq '1'
+
+        if ($version -eq '10.0') {
+            $release = if ($consumerrel) { '10' } else { '2016' }
+        } elseif ($version -eq '6.3') {
+            $release = if ($consumerrel) { '8.1' } else { '2012 R2' }
+        } elseif ($version -eq '6.2') {
+            $release = if ($consumerrel) { '8' } else { '2012' }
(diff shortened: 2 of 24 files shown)


mruzicka:
This is ready for review, although it will need to be updated to make use of #357.

modules/minifact/plans/retrieve.pp
```diff
@@ -0,0 +1,53 @@
...
+        })
+      } else {
+        # Run the task on the targets
+        run_task($task, $targets, '_catch_errors' => true)
+      }
+    }.reduce([]) |$results, $result_set| {
+      # Collect the results from the individual result sets
+      $results + $result_set.results
+    }
+  )
```
mruzicka: The order of the results in the result set may not match the order of the nodes in the plan's `nodes` parameter. Is Is this a concern? Should we reorder the results to match the order of the nodes?

adreyer: it's fine for now I think

modules/minifact/plans/retrieve.pp
```diff
@@ -0,0 +1,53 @@
...
+# node's platform the node is treated as if the task had failed.
+#
+# The $nodes parameter is a list of the nodes to retrieve the facts
+# from.
+plan minifact::retrieve(TargetSpec $nodes) {
+  $targets = get_targets($nodes)
+
+  # Build a mapping from the names of the tasks to run to the lists of
+  # targets to run the tasks on
+  $task_targets = $targets.reduce({}) |$task_targets, $target| {
```
MikaelSmith: This is kind of unreadable. We could simplify this by leveraging Ruby's `group_by` in a separate function. Then it would be
```
$task_targets = $targets.group_by |$target| {
  $target.protocol ? {
    'ssh'   => 'minifact::bash',
    'winrm' => 'minifact::powershell',
    'local' => if minifact::is_command_available('bash') { 'minifact::bash' } else { undef },
    default => undef,
  }
}
```

I guess we could also break it out in a separate Puppet function that re-implements `group_by` as you have here.

MikaelSmith: Also, there's no support for pcp (which could directly use facter, i.e. the facter_task). Was that a separate ticket?

mruzicka: The `group_by` function seems too general to include in the `minifact` module, it would be better suited for the `stdlib` module IMO. So I'm going to add the puppet function.

mruzicka: I'm going to add the pcp support through a ruby task which should run on all platforms supported by puppet agent.

mruzicka: It seems it's not possible to write a native puppet function accepting a block parameter so I ended up writing a ruby based function instead. Since it is namespaced, we don't risk a conflict with a globally defined `group_by` function (if/when added).

modules/minifact/tasks/ruby.rb
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env ruby # rubocop:disable Lint/ScriptPermission
+# frozen_string_literal: true
+
+# Delegate to facter
+exec('facter', '--json')
```
MikaelSmith: It'd probably make sense to bring in all of https://github.com/puppetlabs/puppetlabs-facter_task/blob/master/tasks/init.rb, which has already dealt with some issues around finding Facter.

MikaelSmith: Oh, the implementation is different because it's only returning a single fact.

MikaelSmith: Yeah, I don't know what's going on with that module then. Maybe no one's used it with structured facts?

MikaelSmith: Yeah, it works but doesn't return anything readable. Comes back as
```
{
  "status" : "{\n  architecture => \"x86_64\",\n  family => \"RedHat\",\n  hardware => \"x86_64\",\n  name => \"CentOS\",\n  release => {\n    full => \"7.2.1511\",\n    major => \"7\",\n    minor => \"2\"\n  },\n  selinux => {\n    config_mode => \"enforcing\",\n    config_policy => \"targeted\",\n    current_mode => \"enforcing\",\n    enabled => true,\n    enforced => true,\n    policy_version => \"28\"\n  }\n}"
}
```

mruzicka: My comment was unsubstantiated, the JSON parse is parsing the task's input parameters rather than the facter output, as I was thinking when I wrote the comment.
But still the value currently returned for structured facts is not very useful for machine processing.

modules/minifact/tasks/ruby.rb
```diff
@@ -0,0 +1,18 @@
...
+# frozen_string_literal: true
+
+def facter_executable
+  if ENV.key? 'Path'
+    ENV['Path'].split(';').each do |p|
+      if p =~ /Puppet\\bin\\?$/
+        return File.join(p, 'facter')
+      end
+    end
+    'C:\Program Files\Puppet Labs\Puppet\bin\facter'
```
MikaelSmith: I want to double check this works, but this all makes sense to me.

MikaelSmith: Works great 👍

mruzicka: It does at least on the version of Windows I'm testing it on:
```
C:\Users\Administrator>"C:\Program Files\Puppet Labs\Puppet\sys\ruby\bin\ruby" -e "exec('C:\Program Files\Puppet Labs\Puppet\bin\facter', '--json', 'os')"
{
  "os": {
    "architecture": "x64",
    "family": "windows",
    "hardware": "x86_64",
    "name": "windows",
    "release": {
      "full": "2012 R2",
      "major": "2012 R2"
    },
    "windows": {
      "system32": "C:\\Windows\\system32"
    }
  }
}
```

modules/minifact/tasks/powershell.ps1
```diff
@@ -0,0 +1,53 @@
+#!powershell.exe
+
+# Delegate to facter if available
+if (Get-Command facter -ErrorAction SilentlyContinue) {
+    facter --json
```
MikaelSmith: I think at some point we'll want to unify finding Facter on Windows between this and the Ruby version. Not sure how yet.

mruzicka: We can re-implement the same logic, but I don't see a way how we could share it.

adreyer: I think we can probably do away with the ruby one once we can mutliple implementations

adreyer:
Is michael's comment about the module name not matching the requirements in the ticket getting addressed? I think we can land code here before the module but we probably don't want the name to change.

mruzicka:
The question is: Is `facter_task` still a good name for the functionality?
Given that:
- the module doesn't rely on facter
- it provides a minimal set of one fact if facter is not available
- if facter is available the additional facts it provides is a bonus rather than something to rely on
- there is no support for retrieving individual facts

MikaelSmith:
Continued in https://github.com/puppetlabs/bolt/pull/371.

PR #356, opened 2018-03-26
(BOLT-349) Make acceptance Rakefile env vars valid bhg strings

lucywyman:
Jenkins needs to read an environment variable *before* running the rakefile that will tell it what targets are required for the job so that it can schedule getting the nodes and running the job. Unfortunately, this means that we need to have the env vars for fetching nodes be valid beaker-hostgenerator strings.

I removed the string manipulation from the Rakefile to reflect this, and also added a script to convert from the old-world to the new-world just to make things a little easier for us? I can remove that if no one finds it useful. The defaults have been updated to generate the same hosts file.

### DIFF
--- a/acceptance/README.md
+++ b/acceptance/README.md
@@ -1,159 +1,219 @@
 # Running Acceptance Tests
 
+### Table of Contents
+
+* [Description](#description)
+* [Setup](#setup)
+* [Environment variables](#environment-variables)
+   * [Required](#required)
+   * [Optional](#optional)
+   * [For Gem Testing](#for-gem-testing)
+   * [For Git Testing](#for-git-testing)
+   * [For Package Testing](#for-package-testing)
+* [Running Tests](#running-tests-on-the-vcloud)
+   * [Rake tasks](#rake-tasks)
+      * [On Specified Targets](#on-specified-targets)
+      * [With Specified Users](#with-specified-users)
+      * [Gem](#gem)
+      * [Git](#git)
+      * [Package](#package)
+   * [With Beaker](#beaker)
+      * [Hosts file](#hosts-file)
+      * [Run test suite with beaker](#run-test-suite-with-beaker)
+
+## Description
+
 This acceptance suite allows you to test bolt using the following methods:
 
 * gem: Install bolt via gem from a gemsource defined as an environment
-  variable (defaults to https://gems.rubygems.org).
-* git: Install bolt via the git repo from a branch/SHA defined as
-  environment variables (defaults to #master/HEAD).
+ variable (defaults to https://gems.rubygems.org).
+* git: Install bolt via the git repo from a branch or SHA defined as
+ environment variables (defaults to #master/HEAD).
+* package: Install bolt via the puppet repos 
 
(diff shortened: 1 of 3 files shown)


MikaelSmith:
Was something broken as a result of this?

lucywyman:
Is there anywhere we run the tests automatically that needs to be updated? I can't seem to find where these env vars would get set in CI.

MikaelSmith:
It appears we skip the Rakefile in CI and use beaker --hosts directly.

lucywyman:
(thumbsup)

MikaelSmith:
Is this necessary then?

lucywyman:
Well, it's necessary for us to have an environment variable to define what hosts are needed. We could have a separate `TEST_TARGETS` env var or somesuch and not mess with `BOLT_NODES` or `BOLT_CONTROLLER`, it just seemed easier to have one var get set than multiple. Plus having humans who needs to run the acceptance tests write bhg strings instead of comma-separated almost-bhg strings didn't seem like too big of a change. 

But Jenkins needs to be able to read it so it knows which resources to allocate for the job, so unfortunately yes :/

lucywyman:
@McdonaldSeanp  would know, but there might other ways for us to tell Jenkins what resources a job needs besides env vars?

MikaelSmith:
Ah, the new pipeline now uses the Rake tasks. This seems fine then.

acceptance/README.md
```diff
@@ -3,184 +3,219 @@
...
+### Required with no default
+
+**SSH_PASSWORD** (No default): Value to be used by the tests as the
+password for the `SSH_USER` for connecting to all hosts over SSH.
+
+
+**WINRM_PASSWORD** (No default): Value to be used by the tests as the
+passord for the `WINRM_USER` for connecting to all hosts over WinRM.
+
+### Required with default
```
MikaelSmith: `Required with default` is a weird thing to say.

MikaelSmith: Why not just required and optional?

lucywyman: Hm, I wasn't sure how to distinguish "This needs to be set, but it's already set for you" and "This doesn't need to be set", but maybe that's not a meaningful distinction?

lucywyman:
Well, we could do this all in the Rakefile if jenkins was allocating resources after running the rake task, but Jenkins reads these env vars before running the rake tasks or bhg command.

acceptance/README.md
MikaelSmith: It doesn't feel particularly meaningful.

lucywyman:
Apparently there have been some updates  to ci-job-configs that may not necessitate this PR.

lucywyman:
M'k, so Jenkins actually generates it's own hosts file before running the rake task, and doesn't depend on our environment variables to do that (as we previously thought). Which means the envrionment variables don't need to be beaker host generator strings! Yay!

The main change here is to *not* generate a new hosts file if Jenkins has already done so. We can tell this has happened if `BEAKER_HOSTS` is set, so just skip host file generation if that env var is set. 

I'm keeping some of the README changes as well, but reverted the host generator string bits.

Thanks for having patience on this PR!

PR #359, opened 2018-03-27
(maint) Stringify environment variables with local transport

nicklewis:
Open3.capture3() fails if the value of any provided environment variable
isn't a string, so now we stringifying them. This behavior matches what the
SSH transport does, by just using whatever the Ruby coercion to string
does, without any special handling for structured data.

This also fixes an issue with the local transport raising misleading errors,
effectively assuming that every exception was a tmpdir creation failure.

### DIFF
--- a/lib/bolt/transport/local.rb
+++ b/lib/bolt/transport/local.rb
@@ -32,8 +32,6 @@ module Bolt
             yield dir
           end
         end
-      rescue StandardError => e
-        raise Bolt::Node::FileError.new("Could not make tempdir: #{e.message}", 'TEMPDIR_ERROR')
       end
       private :in_tmpdir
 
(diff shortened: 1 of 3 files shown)


lib/bolt/transport/local.rb
```diff
@@ -32,8 +32,6 @@ def in_tmpdir(base)
             yield dir
           end
         end
-      rescue StandardError => e
-        raise Bolt::Node::FileError.new("Could not make tempdir: #{e.message}", 'TEMPDIR_ERROR')
```
MikaelSmith: I guess I was trying to be helpful, but it didn't work very well.

MikaelSmith: A test probably needs updated for this. You can run the local transport specific tests via `rake integration:bash`.

nicklewis: Yeah, the problem is that the block form of this method requires makes it hard to rescue *only* tmpdir-related errors... We could do something like wrap the `yield` in a begin/rescue that specially wraps any exception it catches and rethrows, then have the outer begin/rescue rethrow the inner exception. But that seems a little over the top, considering tmpdir errors are unlikely to happen in the first place.

PR #360, opened 2018-03-27
(PUP-8441) Handle serializing Puppet Error Datatype better

adreyer:
Currently none of the Datatypes we expect to support can contain
non-data attributes. Therefore to handle the Error Datatype we only have
to walk a single top level data object so we can do that and drop the
monkey patch. We don't restrict plans to returning only expected data
types. Essentially we previously supported PlanReturn as defined below
and we should either make that official or handle pcore transformation
explicitly.

type PlanReturn = Variant[Data, Error, Result, ResultSet, Target,
Array[PlanReturn], Hash[String, PlanReturn]]

### DIFF
--- a/lib/bolt/util.rb
+++ b/lib/bolt/util.rb
@@ -46,7 +46,7 @@ module Bolt
       end
 
       # Accepts a Data object and returns a copy with all hash keys
-      # modifed by block. use &:to_s to stringify keys or :to_sym to symbolize them
+      # modified by block. use &:to_s to stringify keys or &:to_sym to symbolize them
       def walk_keys(data, &block)
         if data.is_a? Hash
           data.each_with_object({}) do |(k, v), acc|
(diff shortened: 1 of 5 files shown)


adreyer:
I'll put up an example PR of what fully handling pcore objects as plan results would look like.

MikaelSmith:
This didn't require changes in Puppet?

adreyer:
@MikaelSmith it turns out for the specific case of Error we don't have to worry about objects inside objects since error only contains Data.

lib/bolt/util.rb
```diff
@@ -60,6 +60,20 @@ def walk_keys(data, &block)
         end
       end
 
+      # Accepts a Data object and returns a copy with all hash and array values
+      # Arrays and hashes including the inital object are modiefied before
```
MikaelSmith: Typos.

MikaelSmith:
This change should be transparent to users.

PR #363, opened 2018-03-28
(DOCS) add predocs to bolt repo

adreyer:


pre-docs/writing_plans.md
```diff
@@ -0,0 +1,395 @@
...
+	run_command("echo 'Vars for ${host}: ${$targetvars}'", $host)
+}
+```
+
+Or set `vars` in the inventory file using the `vars` key at the group level.
+
+```yaml
+groups:
+  - name: my_nodes
+	nodes:
```
jennaesq: Maybe remove whitespace?

MikaelSmith: Be nice to fix this. I don't see any other issues.

adreyer:
note I haven't formatted known_issues, new_features and release_notes.
I think we probably want to remove them from this repo

pre-docs/writing_plans.md
adreyer: This indentation is messed up on the live site too(though in a different way).
It will need to be fixed in a separate PR so it gets back into easyDITA

PR #364, opened 2018-03-29
(BOLT-435) Return empty hash rather than nil when not in inventory

MikaelSmith:
When a node is not in inventory or doesn't have facts/vars set, return
an empty hash rather than nil so that the plan doesn't error.

### DIFF
--- a/spec/fixtures/modules/facts/plans/emit.pp
+++ b/spec/fixtures/modules/facts/plans/emit.pp
@@ -1,0 +1,4 @@
+plan facts::emit(String $host) {
+  $target = get_targets($host)[0]
+  return "Facts for ${host}: ${facts($target)}"
+}
--- a/spec/fixtures/modules/vars/plans/emit.pp
+++ b/spec/fixtures/modules/vars/plans/emit.pp
@@ -1,5 +1,4 @@
-plan vars(String $host) {
+plan vars::emit(String $host) {
   $target = get_targets($host)[0]
-  $target.set_var('bugs', 'bunny')
   return "Vars for ${host}: ${$target.vars}"
 }
--- a/lib/bolt/inventory.rb
+++ b/lib/bolt/inventory.rb
@@ -93,7 +93,7 @@ module Bolt
     end
 
     def vars(target)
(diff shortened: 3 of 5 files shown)


spec/fixtures/modules/facts/plans/emit.pp
```diff
@@ -0,0 +1,4 @@
+plan facts::emit(String $host) {
+  $target = get_targets($host)[0]
+  run_command("echo 'Facts for ${host}: ${facts($target)}'", $host)
```
mruzicka: Could we return the string directly (without calling the `run_command`) like this:
```
"Facts for ${host}: ${facts($target)}"
```
?

spec/fixtures/modules/vars/plans/emit.pp
```diff
@@ -0,0 +1,4 @@
+plan vars::emit(String $host) {
+  $target = get_targets($host)[0]
+  run_command("echo 'Vars for ${host}: ${$target.vars}'", $host)
```
mruzicka: Same here.

spec/fixtures/modules/facts/plans/emit.pp
MikaelSmith: Yeah, we could.

MikaelSmith: I'll make all the facts and vars tests independent of transport and return values rather than echo them.

MikaelSmith: Updated.

spec/fixtures/modules/facts/plans/emit.pp
```diff
@@ -0,0 +1,4 @@
+plan facts::emit(String $host) {
+  $target = get_targets($host)[0]
+  return "Facts for ${host}: ${facts($target)}"
```
mruzicka: 😮 there is a return in puppet?

MikaelSmith: Yup, for functions. PUP-8608 intends to change plans to rely on it for returning anything.

PR #368, opened 2018-03-30
(PUP-8441) Serialize pcore values in plans

adreyer:
Currently plans are allowed to return arbitrary objects which may not
support to_json and therefore will error when we try to serialize them.
Worse, pcore objects tend to blow up when to_s is called on them outside
a compiler.

We should choose between supporting arbitrary pcore objects as the
return value for plans and handling serialization of them or limiting
the return to datatypes we expect. Currentlty I think the valid types
are Data, Error, Result, ResultSet, Target, and combinations thereof.

If we do decide on serializing pcore we need to choose between
displaying it as raw json like we do with the expected types of using
the rich data format.

I personally lean towards restricting the return types of plans. I think
it's adaquate for now and it's easier to expand the spec to support more
return types than restrict it in the future.

Notes about this implementation:

I'm recreating the ruby objects we care about so that we can handle them
more cleanly outside of pal. An alternate implementation would be to
keep the compiler arround for duration of a bolt execution and serialize
pcore more later as needed.

### DIFF
--- a/lib/bolt/pal.rb
+++ b/lib/bolt/pal.rb
@@ -2,6 +2,7 @@
 
 require 'bolt/executor'
 require 'bolt/error'
+require 'bolt/util'
 
 module Bolt
   class PAL
@@ -50,6 +51,12 @@ module Bolt
     # something has been, no pun intended, "bolted on".
     def add_target_spec(compiler)
       compiler.evaluate_string('type TargetSpec = Boltlib::TargetSpec')
+    end
+
+    def prime_types(compiler)
+      #result set references Result with references Target
+      compiler.evaluate_string('ResultSet')
+      compiler.evaluate_string('Error')
     end
 
     def full_modulepath(modulepath)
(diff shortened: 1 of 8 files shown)


adreyer:
This is known not to work as is. The PR is intended for discussion only

lib/bolt/pal.rb
```diff
@@ -228,8 +235,44 @@ def run_task(task_name, targets, params, executor, inventory, &eventblock)
...
     def run_plan(plan_name, params, executor = nil, inventory = nil)
       in_plan_compiler(executor, inventory) do |compiler|
-        r = compiler.call_function('run_plan', plan_name, params)
-        Bolt::PuppetError.convert_puppet_errors(r)
+        result = compiler.call_function('run_plan', plan_name, params)
+        strip_pcore(result, compiler)
+      end
+    end
+
+    # Depth first walk of pcore rich data to convert it into data and bolt ruby objects
```
MikaelSmith: I think limiting Bolt's support to Data makes sense for now. Does that mean we won't display other types nicely, or that they'll cause an error?

adreyer: I think we should error in run_plan to make it clear that you shouldn't write plans that return unexpected types.

PR #369, opened 2018-03-30
(maint) Add spec tests for the puppetdb_fact plan

MikaelSmith:
Note that it places expectations on the executor directly. The
`puppetdb_fact` method may move out of the executor directly, so we
don't add an explicit helper to BoltSpec::Plans.

### DIFF
--- a/lib/bolt/cli.rb
+++ b/lib/bolt/cli.rb
@@ -16,6 +16,7 @@ require 'bolt/puppetdb'
 require 'bolt/pal'
 require 'bolt/target'
 require 'bolt/version'
+require 'bolt/util/on_access'
 
 module Bolt
   class CLIError < Bolt::Error
@@ -449,8 +450,10 @@ Available options are:
 
     def puppetdb_client
       return @puppetdb_client if @puppetdb_client
-      puppetdb_config = Bolt::PuppetDB::Config.new(nil, config.puppetdb)
-      @puppetdb_client = Bolt::PuppetDB::Client.from_config(puppetdb_config)
+      @puppetdb_client = Bolt::Util::OnAccess.new do
+        puppetdb_config = Bolt::PuppetDB::Config.new(nil, config.puppetdb)
+        Bolt::PuppetDB::Client.from_config(puppetdb_config)
+      end
     end
 
     def query_puppetdb_nodes(query)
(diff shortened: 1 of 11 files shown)


modules/puppetdb_fact/spec/plans/init_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+require 'spec_helper'
+require 'bolt_spec/plans'
+require 'bolt/target'
+
+describe 'puppetdb_fact' do
+  include BoltSpec::Plans
+
+  let(:facts) {
+    {
+      'foo' => { 'fqdn' => 'foo' },
```
lucywyman: One nit: you may as well add another fact to one of these nodes just to make sure it's not just the first fact that gets returned.

MikaelSmith:
Updated and added a refactor @adreyer suggested.

lib/bolt/cli.rb
```diff
@@ -511,7 +511,7 @@ def execute(options)
           options[:task_options]['nodes'] = options[:nodes].join(',')
         end
         executor = Bolt::Executor.new(config, options[:noop], true)
-        result = pal.run_plan(options[:object], options[:task_options], executor, inventory)
+        result = pal.run_plan(options[:object], options[:task_options], executor, inventory, puppetdb_client)
```
MikaelSmith: @adreyer should we catch and log debug if constructing puppetdb_client fails? Or wrap it in a function that initializes it the first time it's used?

MikaelSmith: Made a wrapper to delay construction until we try to use the client.

MikaelSmith: I don't think we want to always fail if puppetdb isn't configured. This delays errors until use, which seems like the only viable UX.

lib/bolt/on_access.rb
```diff
@@ -0,0 +1,21 @@
...
+    def initialize(&block)
+      @constructor = block
+      @obj = nil
+    end
+
+    # If a method is called and we haven't constructed the object,
+    # construct it and pass through. Otherwise error.
+    # rubocop:disable Style/MethodMissing
+    def method_missing(method, *args, &block)
+      super unless @obj.nil?
```
MikaelSmith: This isn’t right, it’ll fail all calls after the first.

lib/bolt/cli.rb
adreyer: agreed

PR #371, opened 2018-03-30
(BOLT-359) Add facts module to modules

MikaelSmith:
Adds the `facts` module which provides several plans (and supporting
tasks) for retrieving facts from nodes.


lib/bolt/inventory.rb
```diff
@@ -36,6 +36,12 @@ def initialize(target)
       end
     end
 
+    class UnknownProtocolError < Bolt::Error
+      def initialize(uri)
+        super("Unknown protocol found in #{uri}", 'bolt.inventory/unknown-protocol')
```
adreyer: Could this come from the config too? Should it be bolt.config maybe?(nitpick)

modules/facts/plans/retrieve.pp
```diff
@@ -0,0 +1,46 @@
...
+  $targets = get_targets($nodes)
+
+  # Build a mapping from the names of the tasks to run to the lists of
+  # targets to run the tasks on
+  $task_targets = $targets.facts::group_by |$target| {
+    $target.protocol ? {
+      'ssh'   => 'facts::bash',
+      'winrm' => 'facts::powershell',
+      'pcp'   => 'facts::ruby',
+      'local' => if (facts::is_command_available('bash')) { 'facts::bash' } else { undef },
```
adreyer: If you have local in your plan and don't have bash there are probably bigger problems at this point.

modules/facts/plans/retrieve.pp
```diff
@@ -0,0 +1,46 @@
...
+    }
+  }
+
+  # Return a single result set composed of results from the result sets
+  # returned by the individual task runs (and a result set synthesized
+  # for the unsupported targets)
+  ResultSet(
+    $task_targets.map |$task, $targets| {
+      if ($task == undef) {
+        # Synthesize a result set for the unsupported targets
```
adreyer: Is there anything here except local transport on windows?

lib/bolt/inventory.rb
MikaelSmith: Yeah it could, have a suggestion how I should clean it up?

modules/facts/plans/retrieve.pp
MikaelSmith: That's probably true. Drop `is_command_available`?

modules/facts/plans/retrieve.pp
MikaelSmith: Nope, not at this point.

lib/bolt/inventory.rb
adreyer: `bolt.config/unknown-transport` seems more general and slightly clearer to the user.

modules/facts/plans/retrieve.pp
adreyer: it seems like it removes a lot of complexity we can just let the normal error bubble up.

lib/bolt/inventory.rb
MikaelSmith: Actually, this is currently only thrown if the target URI itself specifies an unknown protocol. I haven't dealt with unexpected input from config/inventory.

adreyer: ah so this could come from the --nodes or nodes listed in the inventory.
`bolt.target/unknown-transport` or just `bolt/unknown-transport` it's probably just fine as is.

MikaelSmith:
Fixed up.

MikaelSmith:
Updated.

PR #373, opened 2018-04-03
(BOLT-410) Allow descriptions for run_* functions and CLI

nicklewis:
The `run_task()`, `run_script()`, `run_command()`, and `file_upload()`
functions can now take a positional "description" argument after the target. If
supplied, the description will be logged at notice level when running that
step, and will also be set as the job description in orchestrator.

There is also now a `--description` argument which can be used to specify the
description for a one-off job from the CLI. That description will similarly be
logged and sent to orchestrator.

### DIFF
--- a/lib/bolt/executor.rb
+++ b/lib/bolt/executor.rb
@@ -98,6 +98,7 @@ module Bolt
     end
 
     def run_command(targets, command, options = {}, &callback)
+      @logger.notice(options['_description']) if options.key?('_description')
       @logger.info("Starting command run '#{command}' on #{targets.map(&:uri)}")
       notify = proc { |event| @notifier.notify(callback, event) if callback }
       options = { '_run_as' => run_as }.merge(options) if run_as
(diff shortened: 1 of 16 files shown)


lib/bolt/executor.rb
```diff
@@ -98,6 +98,7 @@ def batch_execute(targets)
     end
 
     def run_command(targets, command, options = {}, &callback)
+      @logger.notice(options['_description']) if options.key?('_description')
```
adreyer: I think this should just be worked into the next message and follow the verbosity of it.

adreyer:
The ticket in the commit message should be 410 not 140 too

nicklewis:
I'm working on updating the log messages now. This is what I have with descriptions:

image

And without descriptions:

image

That seem reasonable?

nicklewis:
I think it's actually more difficult to read if I replace the "finished" message with a generic one:

image

nicklewis:
Updated.

PR #375, opened 2018-04-03
(BOLT-359) Fixes for the facts module

MikaelSmith:
Adds tasks to the gem, adds fallback to kernel for bash facts (which adds support for macOS), and adds integration tests for Windows and Linux.

### DIFF
--- a/modules/facts/tasks/powershell.ps1
+++ b/modules/facts/tasks/powershell.ps1
@@ -3,17 +3,20 @@
 # Delegate to facter if available
 if (Get-Command facter -ErrorAction SilentlyContinue) {
     facter --json
-} else {
-    if ([System.Environment]::OSVersion.Platform -gt 2) { # [System.PlatformID]::Win32NT
-@'
+}
+else {
+    if ([System.Environment]::OSVersion.Platform -gt 2) {
+        # [System.PlatformID]::Win32NT
+        @'
 {
   "_error": {
-    "kind": "minfact/noname",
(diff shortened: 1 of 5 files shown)


MikaelSmith:
Yeah, we probably do want acceptance tests. I could move the integration tests there.

modules/facts/tasks/powershell.ps1
```diff
@@ -3,17 +3,20 @@
 # Delegate to facter if available
 if (Get-Command facter -ErrorAction SilentlyContinue) {
     facter --json
-} else {
-    if ([System.Environment]::OSVersion.Platform -gt 2) { # [System.PlatformID]::Win32NT
-@'
+}
+else {
+    if ([System.Environment]::OSVersion.Platform -gt 2) {
+        # [System.PlatformID]::Win32NT
```
mruzicka: This comment refers to the value "2" on the previous line so should be on that line. It is the value of the `[System.PlatformID]::Win32NT` constant. For some reason using that constant doesn't work on Windows Server Core but it does work on other versions of Windows.

MikaelSmith: This was mostly auto-reformatting by the Powershell extensions for Visual Studio Code. We can revert it. The only actual change was to the error kind.

mruzicka: No need to revert it, I'll put a PR up moving the comment before the condition and making it clearer.

PR #380, opened 2018-04-04
(BOLT-410) Add test that description is used in logging

MikaelSmith:
Add a test that description, when passed as an argument to a run action,
is included in logging.


spec/integration/logging_spec.rb
```diff
@@ -50,6 +50,13 @@
     expect(result[0]['result']['_output'].strip).to match(/hi there/)
   end
 
+  it 'logs description when included' do
```
MikaelSmith: I could make this replace the similar test above to avoid too much ssh integration testing.

PR #382, opened 2018-04-04
(BOLT-410) Pre-docs for description and other fixes

MikaelSmith:


adreyer:
can you make a note for kate regarding the reordering of the command options

PR #386, opened 2018-04-05
(BOLT-454) Return undef from plans without explicit return

mruzicka:
This disables implicit return values from plans. More precisely: Any plan which finishes without executing a `return` statement returns `undef`.
If a different return value is desired it must be returned using the `return` statement.

### DIFF
--- a/bolt-modules/boltlib/lib/puppet/functions/run_plan.rb
+++ b/bolt-modules/boltlib/lib/puppet/functions/run_plan.rb
@@ -47,7 +47,12 @@ Puppet::Functions.create_function(:run_plan, Puppet::Functions::InternalFunction
       end
 
       begin
-        result = func.class.dispatcher.dispatchers[0].call_by_name_with_scope(scope, params, true)
+        func.class.dispatcher.dispatchers[0].call_by_name_with_scope(scope, params, true)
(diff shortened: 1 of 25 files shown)


bolt-modules/boltlib/lib/puppet/functions/run_plan.rb
```diff
@@ -47,7 +47,12 @@ def run_plan(scope, plan_name, named_args = {})
       end
 
       begin
-        result = func.class.dispatcher.dispatchers[0].call_by_name_with_scope(scope, params, true)
+        func.class.dispatcher.dispatchers[0].call_by_name_with_scope(scope, params, true)
+        # If the plan returns using a return statement then we never get here.
+        # We only get here if the plan finishes without executing any return
+        # statement and since we don't want to leak any implicit return value
+        # from the plan we simply set the result to nil here.
+        result = nil
```
adreyer: Ah I see what you were saying about making it harder to validate types. We'll have to catch the return for that.

adreyer:
I think we need to add a message when a plan returns undef. I don't think there is a good reason to differentiate between "return undef" and no return value so something simple in the human outputter is probably adequate.

"Plan completed successfully with no result"

PR #389, opened 2018-04-05
 (BOLT-454, BOLT-456) Return undef from plans without explicit return and restrict plan return types

adreyer:


pre-docs/writing_plans.md
```diff
@@ -88,6 +88,24 @@ or a comma seperated string of target names.
...
 ```
 
+## Returning results from plans
+
+> Plans can optionally return a result that can be used from other plans or
+> saved externally.
+
+Plans, unlike functions, are primarily run for side effects but they can
+optionally return a result. To return a result from a plan use the `return`
+function. Any plan that does not call the `return` function will have and `undef`
```
MikaelSmith: `and` -> `an`

bolt-modules/boltlib/types/planresult.pp
```diff
@@ -0,0 +1,4 @@
+# A Planresult describes the supported return values of a plan. It
+# passed to get_targets() to return an Array[Target].
+#type Boltlib::PlanResult = Variant[Data, Error, Result, ResultSet, Target, Array[PlanResult], Hash[String, PlanResult]]
```
MikaelSmith: What happened here?

lib/bolt/pal.rb
```diff
@@ -43,12 +43,13 @@ def self.load_puppet
...
-    # Create a top-level alias for TargetSpec so that users don't have to
+    # Create a top-level alias for TargetSpec and PlanResult so that users don't have to
     # namespace it with Boltlib, which is just an implementation detail. This
-    # allows TargetSpec to feel like a built-in type in bolt, rather than
+    # allows them to feel like a built-in type in bolt, rather than
     # something has been, no pun intended, "bolted on".
-    def add_target_spec(compiler)
+    def alias_types(compiler)
       compiler.evaluate_string('type TargetSpec = Boltlib::TargetSpec')
+      compiler.evaluate_string('type PlanResult = Boltlib::PlanResult')
```
MikaelSmith: Do we expect people to use PlanResult directly?

pre-docs/writing_plans.md
```diff
@@ -88,6 +88,45 @@ or a comma seperated string of target names.
...
+- `Result`
+- `ResultSet`
+- `Error`
+- `Array` with only `PlanResult`
+- `Hash` with `String` keys and `PlanResult` values
+
+or
+
+```
+Variant[Data, String, Numeric, Boolean, Error, Result, ResultSet, Target, Array[Boltlib::PlanResult], Hash[String, Boltlib::PlanResult]]
```
MikaelSmith: Based on https://puppet.com/docs/puppet/5.3/lang_data_abstract.html#data, I would expect `Data` to match scalars.

lib/bolt/pal.rb
adreyer: In functions that run plans perhaps

PR #393, opened 2018-04-09
(maint) Increase winrm timeout in spec tests

MikaelSmith:
AppVeyor periodically times out connecting over WinRM during spec tests.
Increase the default timeout to avoid connection problems during
testing.

### DIFF
--- a/spec/bolt/transport/winrm_spec.rb
+++ b/spec/bolt/transport/winrm_spec.rb
@@ -15,6 +15,8 @@ describe Bolt::Transport::WinRM do
 
   def mk_config(conf)
     stringified = conf.each_with_object({}) { |(k, v), coll| coll[k.to_s] = v }
+    # The default of 10 seconds seems to be too short to always succeed in AppVeyor.
+    stringified['connect-timeout'] ||= 20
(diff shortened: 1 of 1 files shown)


spec/bolt/transport/winrm_spec.rb
```diff
@@ -14,8 +14,10 @@
   include BoltSpec::Task
 
   def mk_config(conf)
-    stringified = conf.each_with_object({}) { |(k, v), coll| coll[k.to_s] = v }
-    Bolt::Config.new(transport: 'winrm', transports: { winrm: stringified })
+    unstringified = conf.each_with_object({}) { |(k, v), coll| coll[k.to_s] = v }
```
nicklewis: Why is this *un*stringified now?

MikaelSmith: Oh, neither name actually makes much sense. `stringified` was confusing. I can try to come up with a new one.

MikaelSmith: Oh right, I forgot what i was doing. Turning keys into strings.

PR #394, opened 2018-04-09
 (maint) Fail travis when module tests fail

adreyer:


.travis.yml
```diff
@@ -13,14 +13,20 @@ script:
...
+  if ! bundle exec rake spec; then
+    exit 1
+  fi
   popd
 - |
+  status=0
   for i in $( ls modules ); do
     pushd modules/$i
-    bundle exec rake spec
+    if ! bundle exec rake spec; then
```
MikaelSmith: Interesting. I guess blocks run in an implicit `set -e`?

PR #395, opened 2018-04-09
(BOLT-439) Support inventory groups defined within other groups

nicklewis:
Previously, the code needed to support nested inventory groups was
present, but with an explicit check to disable it. This commit removes
that check and fixes a bug where variables were merged in reverse order
compared to other values.

Groups can now be defined within other groups. Variables/config prefer the
first "branch" (top-level group) that the node matches, and use the most
deeply nested (most specific) group within that branch. In other words,
we use depth-first lookup, and merge across branches.

A group is not allowed to be defined as a child of multiple parent
groups. This triggers the same error as redefining a group did
previously.

### DIFF
--- a/lib/bolt/inventory/group.rb
+++ b/lib/bolt/inventory/group.rb
@@ -52,7 +52,6 @@ module Bolt
         if node_names.include?(@name)
           raise ValidationError.new("Group #{@name} conflicts with node of the same name", @name)
         end
-        raise ValidationError.new("Group #{@name} is too deeply nested", @name) if depth > 1
 
         check_deprecated_config('Group', @name, @config)
 
@@ -129,7 +128,7 @@ module Bolt
           # Shallow merge instead of deep merge so that vars with a hash value
           # are assigned a new hash, rather than merging the existing value
           # with the value meant to replace it
-          'vars'   => data2['vars'].merge(data1['vars']),
+          'vars'   => data1['vars'].merge(data2['vars']),
           'facts'  => Bolt::Util.deep_merge(data1['facts'], data2['facts']),
           'groups' => data2['groups'] + data1['groups']
         }
--- a/spec/bolt/inventory/group_spec.rb
+++ b/spec/bolt/inventory/group_spec.rb
@@ -421,4 +421,132 @@ describe Bolt::Inventory::Group do
       expect { group.validate }.to raise_error(Bolt::Inventory::ValidationError, /conflicts with node/)
     end
   end
+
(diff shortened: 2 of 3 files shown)


MikaelSmith:
This is failing tests.

lib/bolt/inventory/group.rb
```diff
@@ -129,7 +128,7 @@ def data_merge(data1, data2)
           # Shallow merge instead of deep merge so that vars with a hash value
           # are assigned a new hash, rather than merging the existing value
           # with the value meant to replace it
-          'vars'   => data2['vars'].merge(data1['vars']),
+          'vars'   => data1['vars'].merge(data2['vars']),
```
MikaelSmith: Why was this reversed?

MikaelSmith: Ah, I see the commit references a bug here.

nicklewis: I should update the commit message to mention what the bug was...

MikaelSmith: This may have been intentionally reversed, so that `vars` from the 1st entry was the one used. How does it now behave when a node exists in multiple unrelated groups and declares vars in each?

MikaelSmith: Well, the tests there look reasonable and we didn't have tests before. So this behavior seems sane.

MikaelSmith: And the bug appears to be that we prefered group vars over node vars, which definitely doesn't seem right.

nicklewis: This method is used in a couple of places.

When we're reducing over the *list* of groups, we pass each group as the first argument, and the accumulator as the *second* argument, which means the first group that sets something wins (it goes into the accumulator and overrides later groups).

When we're traversing *down* the group hierarchy, we pass the parent group as the first argument and the aggregated value from the children as the second argument, so that children win over parents.

When we're combining node-specific data and group-specific data, we pass the group data as the first argument and the node data as the second argument, so that node-specific data wins.

I think that last one means that in this case:

```yaml
groups:
  - name: parent
    nodes:
      - node1:
          vars:
            foo: bar
  groups:
    - name: child
      nodes: node1
      vars:
        foo: baz
```

The node `node1` has vars `foo: bar`. That is, even though the node-specific value is set at the `parent` group level, it wins over the group value in the *more specific* `child` group.

So the total preference order is:
- depth-first search for node-specific values
- depth-first search for group-specific values

PR #396, opened 2018-04-10
(maint) Update ruby on targets for acceptance tests

lucywyman:
This ensures that either ruby 2.4 or a specified version are installed on the SUTs before running the acceptance tests.

### DIFF
--- a/acceptance/README.md
+++ b/acceptance/README.md
@@ -99,6 +99,8 @@ will default to `root`.
 tests as the user for connecting to all hosts using the `winrm` role.
 This user must already be present on the system.
 
+**RUBY_URL** (Default `https://artifactory.delivery.puppetlabs.net/artifactory/generic__buildsources/buildsources/ruby-2.4.4.tar.gz`): The url to download and install ruby from.
+
 ### For Gem Testing
 
 **GEM_VERSION** (Default `> 0.1.0`): When testing via gem install,
--- a/acceptance/lib/acceptance/install_ruby.rb
+++ b/acceptance/lib/acceptance/install_ruby.rb
@@ -1,0 +1,7 @@
+def install_ruby_from_source(host)
+  ENV['RUBY_URL'] ||= "https://artifactory.delivery.puppetlabs.net/artifactory/generic__buildsources/buildsources/ruby-2.4.4.tar.gz"
+  on(host, "mkdir ruby")
+  on(host, "curl -o ruby.tar.gz #{ENV['RUBY_URL']}")
+  on(host, "tar -xvf ruby.tar.gz")
+  on(host, "cd $(ls | grep 'ruby-') && ./configure && make && make install")
+end
--- a/acceptance/setup/common/pre-suite/010_install_ruby.rb
+++ b/acceptance/setup/common/pre-suite/010_install_ruby.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require 'install_ruby'
+
 test_name "Install Ruby" do
   step "Ensure Ruby is installed on Bolt controller" do
     result = nil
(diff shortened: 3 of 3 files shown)


MikaelSmith:
I'm not clear why this is needed.

MikaelSmith:
We previously addressed this by only running gem/git tests on platforms that include Ruby 2.3+. We were going to test packages on all supported platforms.

acceptance/lib/acceptance/install_ruby.rb
```diff
@@ -0,0 +1,6 @@
+def install_ruby_from_source(host, url = "https://cache.ruby-lang.org/pub/ruby/2.4/ruby-2.4.4.tar.gz")
```
MikaelSmith: We should mirror this instead of downloading from ruby-lang every time.

MikaelSmith: Which of course we didn't do for Windows.

lucywyman: Eh, still a good idea :P

acceptance/lib/acceptance/install_ruby.rb
```diff
@@ -0,0 +1,7 @@
+def install_ruby_from_source(host, version="2.4.4")
+  url = "https://artifactory.delivery.puppetlabs.net/artifactory/generic__buildsources/buildsources/ruby-#{version}.tar.gz"
```
MikaelSmith: I'd like to see this in an environment variable. It doesn't look like we have anything in Jenkins yet. Maybe just a way to override with an environment variable if folks want to run this outside our VPN.

acceptance/setup/common/pre-suite/010_install_ruby.rb
```diff
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require 'install_ruby'
```
MikaelSmith: I'm surprised this is sufficient to pickup `lib/acceptance`.

MikaelSmith:
Can you also link to where we're configuring the platforms that rely on this change? My feeling is these should only be used for smoke tests, and that can be a fairly small set of platforms. Testing with packages should pick up the full suite.

acceptance/setup/common/pre-suite/010_install_ruby.rb
lucywyman: lib/acceptance is already included as there's another library there :) So it's already set up.

acceptance/lib/acceptance/install_ruby.rb
```diff
@@ -0,0 +1,7 @@
+def install_ruby_from_source(host, version="2.4.4")
+  url = ENV['RUBY_URL'] || "https://artifactory.delivery.puppetlabs.net/artifactory/generic__buildsources/buildsources/ruby-#{version}.tar.gz"
```
MikaelSmith: Document in the acceptance/README as well.

acceptance/README.md
```diff
@@ -99,6 +99,8 @@ will default to `root`.
 tests as the user for connecting to all hosts using the `winrm` role.
 This user must already be present on the system.
 
+**RUBY_URL** (Default `https://artifactory.delivery.puppetlabs.net/artifactory/generic__buildsources/buildsources/ruby-2.4.4.tar.gz`): The url to download and install ruby from.
```
MikaelSmith: Can you include the ruby.org source as a suggestion? I think it was `https://cache.ruby-lang.org/pub/ruby/2.4/ruby-2.4.4.tar.gz`.

lucywyman: Yeah, should I include that this is a mirror of that?

MikaelSmith: I think that's probably clear. Maybe say something like
> If the default mirror is not accessible to you, use `https://cache.ruby-lang.org/pub/ruby/2.4/ruby-2.4.4.tar.gz`

MikaelSmith:
Only recently noticed we were doing a Ruby `make install`. I don't think this is the approach we want to take.

PR #398, opened 2018-04-10
(BOLT-426) Add pre-docs for installing

MikaelSmith:
Highlight packages as the primary install method, with installing from
gem available as a last option.

### DIFF
--- a/pre-docs/bolt_installing.md
+++ b/pre-docs/bolt_installing.md
@@ -1,67 +1,21 @@
 
 # Installing Bolt
 
-Install Bolt and any dependencies for your operating system, such as Ruby, a
-GNU Compiler Collection (GCC) compiler and the Bolt gem.
-
-Packaged versions of Bolt are available for select Linux distributions and
-versions: Debian 8 and 9, Enterprise Linux 6 and 7, SUSE Linux Enterprise
-Server 12 and Ubuntu 14.04 and 16.04.
-
-
-## Install Bolt on *nix
-
-Install Bolt and its dependencies on *nix platforms.
-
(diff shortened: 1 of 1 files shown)


pre-docs/bolt_installing.md
```diff
@@ -125,27 +72,49 @@ http://yum.puppet.com/puppet5/ Packages are named using the convention
...
+  ```
+  sudo rpm -Uvh https://yum.puppet.com/puppet5/puppet5-release-sles-12.noarch.rpm
+  sudo zypper install bolt
+  ```
+
+Get started with `bolt --help`.
+
+## Install Bolt as a gem
+
+Install Bolt and any dependencies for your operating system, such as Ruby, a
```
nicklewis: We should probably explicitly call out that it depends on Ruby 2.3 and therefore only very modern platforms are supported. Just so otherwise users don't try and fail to follow our instructions.

pre-docs/bolt_installing.md
```diff
@@ -125,27 +72,49 @@ http://yum.puppet.com/puppet5/ Packages are named using the convention
...
+
+- For Fedora 27
+  ```
+  dnf install -y ruby rubygem-json rubygem-ffi rubygem-bigdecimal rubygem-io-console
+  ```
+- For Debian 9 or Ubuntu 16.04
+  ```
+  apt-get install -y ruby ruby-ffi
+  ```
+- For Mac OS X
```
nicklewis: 10.13 / High Sierra only

pre-docs/bolt_installing.md
MikaelSmith: Updated.

PR #400, opened 2018-04-11
(BOLT-455) Use bolt logger from puppet

nicklewis:
This adds a stub 'logging' logdestination type to Puppet that passes log
messages through to the Logging logger used by bolt. This causes
Puppet's logs to be logged alongside bolt logs, with the same format.
This also means that Puppet logs properly get logged to files in
addition to the console.

This also fixes a bug where we weren't actually logging in Puppet,
because `with_puppet_settings` was resetting the log destinations to
empty. We now reconfigure logging every time we re-initialize settings,
so that we always log.

We now use the 'debug' log level for Puppet, causing it to "log"
*everything* (which really means pass it through to the Logging logger),
allowing the Logging logger to decide which messages to actually log.

### DIFF
--- a/lib/bolt/pal.rb
+++ b/lib/bolt/pal.rb
@@ -12,7 +12,7 @@ module Bolt
       # Nothing works without initialized this global state. Reinitializing
       # is safe and in practice only happen in tests
       self.class.load_puppet
-      self.class.configure_logging
+
       # This makes sure we don't accidentally create puppet dirs
       with_puppet_settings { |_| nil }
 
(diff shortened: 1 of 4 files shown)


lib/bolt/pal.rb
```diff
@@ -21,7 +22,42 @@ def initialize(config)
...
+      Puppet::Util::Log.newdestination(Logging.logger[self])
+      # Defer all log level decisions to the Logging library by telling Puppet
+      # to log everything
+      Puppet.settings[:log_level] = 'debug'
+    end
+
+    def self.define_external_logdest
+      return if defined?(Puppet::Util::Log::DestExternal)
+
+      Puppet::Util::Log.newdesttype :external do
```
MikaelSmith: Ah, hadn't found this approach. Nice.

lib/bolt/pal.rb
```diff
@@ -39,6 +42,8 @@ def self.load_puppet
         raise Bolt::CLIError, "Puppet must be installed to execute tasks"
       end
 
+      require 'bolt/pal/logging'
```
MikaelSmith: This is an odd pattern. Seems like it'd usually be done with a Module that's loaded at the top, then used here.

PR #405, opened 2018-04-19
(BOLT-372) Send information to orchestrator when running a plan

adreyer:
When a plan is run the orchestrator transport will send the plan_start
command the first time a task runs on a node through that orchestrator.
Every task will be run with the plan_task command instead of the task
command and before exiting bolt will send a plan_finish command with the
result to each orchestrator that was used during the plan.

### DIFF
--- a/lib/bolt/executor.rb
+++ b/lib/bolt/executor.rb
@@ -16,10 +16,10 @@ module Bolt
     attr_reader :noop, :transports
     attr_accessor :run_as
 
-    def initialize(config = Bolt::Config.new, noop = nil, plan_logging = false)
+    def initialize(config = Bolt::Config.new, noop = nil)
       @config = config
       @logger = Logging.logger[self]
-      @plan_logging = plan_logging
+      @plan_logging = false
 
       @transports = Bolt::TRANSPORTS.each_with_object({}) do |(key, val), coll|
         coll[key.to_s] = Concurrent::Delay.new { val.new }
@@ -34,6 +34,7 @@ module Bolt
 
     def transport(transport)
       impl = @transports[transport || 'ssh']
+      raise(Bolt::UnknownTransportError, transport) unless impl
       # If there was an error creating the transport, ensure it gets thrown
       impl.no_error!
       impl.value
@@ -183,6 +184,25 @@ module Bolt
         @notifier.shutdown
         results
       end
+    end
+
+    # Plan context doesn't make sense for most transports but it is tightly
+    # coupled with the orchestrator transport since the transport behaves
+    # differently when a plan is running. In order to limit how much this
+    # pollutes the transport API we only handle the orchestrator transport here.
+    # Since we callt this function without resolving targets this will result
+    # in the orchestrator transport always being initialized during plan runs.
+    # For now that's ok.
+    #
+    # In the future if other transports need this or if we want a plan stack
+    # we'll need to refactor.
+    def start_plan(plan_context)
+      transport('pcp').plan_context = plan_context
+      @plan_logging = true
+    end
+
+    def finish_plan(plan_result)
+      transport('pcp').finish_plan(plan_result)
     end
   end
 end
--- a/lib/bolt/transport/orch/connection.rb
+++ b/lib/bolt/transport/orch/connection.rb
@@ -1,0 +1,80 @@
+# frozen_string_literal: true
+
+module Bolt
+  module Transport
+    class Orch < Base
+      class Connection
+        attr_reader :logger, :key
+
+        CONTEXT_KEYS = Set.new(%i[plan_name description params]).freeze
+
+        def self.get_key(opts)
+          [
+            opts['service-url'],
+            opts['task-environment'],
+            opts['token-file']
+          ].join('-')
+        end
+
+        def initialize(opts, plan_context, logger)
+          @logger = logger
+          @key = self.class.get_key(opts)
+          client_keys = %w[service-url token-file cacert]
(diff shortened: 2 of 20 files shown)


adreyer:
It seems like we may want to test that the correct connection is used for each batch but any test for it seems more likely to test the specific implementation rather then the behavior.

lib/bolt/executor.rb
```diff
@@ -117,6 +118,11 @@ def with_node_logging(description, batch)
     end
     private :with_node_logging
 
+    def update_options(options)
```
lucywyman: `update_run_as` might be more descriptive unless we plan to add other options to this in the future

lib/bolt/executor.rb
```diff
@@ -184,5 +190,24 @@ def file_upload(targets, source, destination, options = {}, &callback)
...
+    # pollutes the transport API we only handle the orchestrator transport here.
+    # Since we callt this function without resolving targets this will result
+    # in the orchestrator transport always being initialized during plan runs.
+    # For now that's ok.
+    #
+    # In the future if other transports need this or if we want a plan stack
+    # we'll need to refactor.
+    def start_plan(plan_context)
+      transport('pcp').plan_context = plan_context
+      @plan_logging = true
```
lucywyman: Should this get set back to false in `finish_plan` so it's not still true for any other tasks that get run? Or does the executor get instantiated anew for each job run? I can't think of an exception where a plan then a task would run separately as part of one execution, but that might exist.

lib/bolt/executor.rb
adreyer: Ah yeah this is leftover from the original implementation when it was injecting the plan_context

lib/bolt/executor.rb
adreyer: finish_plan is called by the CLI so no more tasks will run.

lib/bolt/transport/orch/connection.rb
```diff
@@ -0,0 +1,79 @@
...
+          logger.debug("Creating orchestrator client for #{client_opts}")
+
+          @client = OrchestratorClient.new(client_opts, true)
+          @plan_job = start_plan(plan_context)
+
+          @environment = opts["task-environment"]
+        end
+
+        def start_plan(plan_context)
+          if plan_context
```
lucywyman: Should this throw an error if there's no plan context? Or does it just return nil?

spec/lib/bolt_spec/plans/mock_executor.rb
```diff
@@ -181,6 +181,10 @@ def run_task(targets, task, arguments, options)
         result
       end
 
+      def with_plan_context(_context)
```
lucywyman: Where does this get used?

lib/bolt/transport/orch/connection.rb
adreyer: I'm not sure how it would get called with no plan context and I don't know anywhere else we validate internal arguments like that.

spec/lib/bolt_spec/plans/mock_executor.rb
adreyer: ah yeah it doesn't anymore

lib/bolt/transport/orch/connection.rb
```diff
@@ -0,0 +1,79 @@
...
+              end
+              nil
+            end
+          end
+        end
+
+        def finish_plan(plan_result)
+          if @plan_job
+            @client.command.plan_finish(
+              result: plan_result.result,
```
MikaelSmith: I think this should be `plan_result.value`.

lucywyman:
@adreyer I'm still seeing unexpected output from this, but perhaps I'm doing something wrong?
```
[root@ni8g44xt80efsvs ~]# bolt plan run aggregate::count command="ls" nodes=kir1y6kpuat8snc.delivery.puppetlabs.net,ni8g44xt80efsvs.delivery.puppetlabs.net --transport pcp
Starting: command 'ls' on kir1y6kpuat8snc.delivery.puppetlabs.net, ni8g44xt80efsvs.delivery.puppetlabs.net
Finished: command 'ls' with 0 failures in 1.7 sec
Failed to finish plan on -production-: plan_job must be present
{
  "stdout": {
    "bin\nboot\ndev\netc\nhome\nlib\nlib64\nmedia\nmnt\nopt\nproc\nroot\nrun\nsbin\nsrv\nsys\ntmp\nusr\nvar\n": 2
  },
  "stderr": {
    "": 2
  },
  "exit_code": {
    "0": 2
  }
}
```

nicklewis:
This will need a change to `bolt-vanagon` to bump the orchestrator-client dependency.

lib/bolt/transport/orch/connection.rb
```diff
@@ -0,0 +1,79 @@
...
+              end
+              nil
+            end
+          end
+        end
+
+        def finish_plan(plan_result)
+          if @plan_job
+            @client.command.plan_finish(
+              result: plan_result.value,
```
nicklewis: Need to pass `plan_job` here. Also this fails if result is nil.

nicklewis: Hmm, if I fix those two issues locally, I get `Failed to finish plan on -production-: Some tasks for this plan job have not finished running`

adreyer: yeah me too, I'm looking into whats wrong with orch there.

regardless I think the solution for bolt is to not finish plans until irving z

PR #413, opened 2018-04-25
(BOLT-483) Task to apply a single resource

nicklewis:
This adds the apply module with an apply::resource task that will apply
a single resource using Puppet. This allows users to leverage the
great deal of existing native Puppet types from their plans, without
having to create a bespoke task for each one. This task functions
basically like running `puppet apply` against a single resource
declaration.


nicklewis:
Pushed this up to see what travis does with the new module...

PR #415, opened 2018-04-26
(BOLT-483) Add apply module to include resource task

nicklewis:


Puppetfile
```diff
@@ -7,6 +7,7 @@ moduledir File.join(File.dirname(__FILE__), 'modules')
 mod 'puppetlabs-package', '0.2.0'
 mod 'puppetlabs-service', '0.3.1'
 mod 'puppetlabs-puppet_conf', '0.2.0'
+mod 'puppetlabs-apply', '0.1.0'
```
MikaelSmith: Presumably this will work once we've actually released the module?

nicklewis: I did a release before putting up this PR. It should be working

PR #425, opened 2018-05-01
(BOLT-506) Add Boolean type to metatdata types for writing tasks docu…

donoghuc:
…mentation


MikaelSmith:
Thanks. I'd generally like to keep the title a little shorter so it doesn't wrap, but this seems fine.

PR #431, opened 2018-05-03
Bolt 485

donoghuc:
Handle case where canary run fails by failing plan and providing expected result format.


MikaelSmith:
I believe this has some spec tests at modules/canary/spec/plans we should add to.

PR #437, opened 2018-05-11
(BOLT-486) Plans accept targets via puppetdb

donoghuc:
When executing a plan, allow targets to be build from PDB query. Also ensure EITHER --nodes OR --query.

### DIFF
--- a/spec/bolt/cli_spec.rb
+++ b/spec/bolt/cli_spec.rb
@@ -592,12 +592,20 @@ bar
 
       it "accepts targets resulting from --query from puppetdb" do
         cli = Bolt::CLI.new(%w[plan run foo --query nodes{}])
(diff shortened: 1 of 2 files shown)


spec/bolt/cli_spec.rb
```diff
@@ -593,11 +593,19 @@ def stub_config(config, file_content = nil)
       it "accepts targets resulting from --query from puppetdb" do
         cli = Bolt::CLI.new(%w[plan run foo --query nodes{}])
         allow(cli).to receive(:query_puppetdb_nodes).and_return(%w[foo bar])
```
adreyer: Lets add .once to this allowance to make sure we're not querying puppetdb multiple times in these cases.

PR #440, opened 2018-05-14
(MAINT) Corrections based on Rubocop 0.56.0

donoghuc:
- passing splat keyword args as single hash
- Style/MethodMissing was split into two separate cops
- save navigation on map

### DIFF
--- a/spec/lib/bolt_spec/integration.rb
+++ b/spec/lib/bolt_spec/integration.rb
@@ -29,7 +29,7 @@ module BoltSpec
     end
 
     def run_cli_json(arguments, **opts)
-      output = run_cli(arguments, **opts)
+      output = run_cli(arguments, opts)
 
       begin
         result = JSON.parse(output, quirks_mode: true)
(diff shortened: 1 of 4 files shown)


spec/lib/bolt_spec/integration.rb
```diff
@@ -28,8 +28,9 @@ def run_cli(arguments, rescue_exec: false)
       output.string
     end
 
-    def run_cli_json(arguments, **opts)
-      output = run_cli(arguments, **opts)
+    def run_cli_json(arguments, *opts)
+      # require "pry"; binding.pry
```
adreyer: pry

spec/lib/bolt_spec/integration.rb
```diff
@@ -28,8 +28,9 @@ def run_cli(arguments, rescue_exec: false)
       output.string
     end
 
-    def run_cli_json(arguments, **opts)
-      output = run_cli(arguments, **opts)
+    def run_cli_json(arguments, *opts)
```
adreyer: We still want to double splat in the signature so we can accept nil or a hash
We don't need to double splat the opts we can just pass them on as a single hash

```
def run_cli_json(arguments, **opts)
  output = run_cli(arguments, opts)
```

donoghuc: Ahhh, I see. I was testing this one and I thought that when I did that it failed. I just re-ran locally and it passed. Makes sense.

PR #441, opened 2018-05-14
(BOLT-135) Support multiple task implementations

nicklewis:
This change adds logic to the existing transports to dynamically select the
appropriate task implementation based on the available features of a
target. Target now has a select_impl() method which, given a task and an
optional list of "additional" features, will return the path to the first
suitable executable. The "additional" features are passed by the transport
and account for any features which are directly implied to be present
based on the transport (ie. 'shell' for ssh and 'powershell' for winrm).

This allows a single task to have multiple implementations for different
platforms and for Bolt to choose the appropriate one for each target at
runtime.

A new set_feature() function in puppet language can be used to add and
remove features on a target.

### DIFF
--- a/lib/bolt/transport/orch.rb
+++ b/lib/bolt/transport/orch.rb
@@ -22,6 +22,8 @@ module Bolt
         %w[service-url cacert token-file task-environment local-validation]
       end
 
+      PROVIDED_FEATURES = ['puppet-agent'].freeze
+
       def self.validate(options)
         validation_flag = options['local-validation']
         unless !!validation_flag == validation_flag
--- a/pre-docs/writing_tasks.md
+++ b/pre-docs/writing_tasks.md
@@ -223,9 +223,6 @@ The task filename `init` is special: the task it defines is referenced using the
 module name only. For example, in the `puppetlabs-service` module, the task
 defined in `init.rb` is the `service` task.
 
-Task names must be unique. If there are two tasks with the same name but different file extensions in a
-module, the task runner won't load either of them.
-
 Each task or plan name segment must begin with a lowercase letter and:
 
 - Must start with a lowercase letter.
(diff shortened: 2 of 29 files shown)


nicklewis:
This depends on puppetlabs/puppet#6829.

nicklewis:
I pushed a commit that bumps the puppet submodule to my PR. That should be removed before we actually merge this.

lib/bolt/transport/orch.rb
```diff
@@ -22,6 +22,8 @@ def self.options
         %w[service-url cacert token-file task-environment local-validation]
       end
 
+      PROVIDED_FEATURES = ['puppet-agent'].freeze
```
adreyer: I think we probably *should* do validation of whether there is a suitable implementation if local-validation is true.

I'm not sure how useful that actually is though

.gitmodules
```diff
@@ -1,7 +1,7 @@
 [submodule "vendored/puppet"]
 	path = vendored/puppet
-	url = https://github.com/puppetlabs/puppet
-	branch = master
+	url = https://github.com/nicklewis/puppet
```
MikaelSmith: Can now pull from puppet#master.

nicklewis:
Updated the vendored puppet for real, now this should be good to go!

MikaelSmith:
Doesn't look like it is yet.

MikaelSmith:
Waiting on https://github.com/puppetlabs/puppet/pull/6855.

lib/bolt/transport/orch.rb
nicklewis: There's no way we *could* make use of it, other than for local validation. This can't work with orchestrator until orchestrator supports it.

nicklewis:
Updated to the latest puppet master branch and added docs on using `implementations`.

pre-docs/writing_tasks.md
```diff
@@ -234,6 +231,51 @@ Each task or plan name segment must begin with a lowercase letter and:
...
+```json
+{
+  "implementations": [
+    {"name": "sql.sh", "requirements": ["shell"]},
+    {"name": "sql.ps1", "requirements": ["powershell"]}
+  ]
+}
+```
+
+Each implementations has a `name` and a list of `requirements`. The
```
MikaelSmith: I think it's worth calling out that `name` can differ from the metadata file name, and that it's still restricted to valid task name patterns (I'm pretty sure Puppet will fail to find `my-impl.sh`).

MikaelSmith:
Bumping Puppet seems to have moved us to needing to include semantic_puppet ourselves. I think that needs to be added into the bolt-runtime artifact.

nicklewis:
Depends on https://github.com/puppetlabs/bolt-vanagon/pull/22

MikaelSmith:
Oh grr, you never updated bolt.gemspec.

PR #442, opened 2018-05-17
(BOLT-520) Enforce concurrency limit

donoghuc:
Initializing CachedThreadPool with max_threads does not cap concurrency as expected. Use ThreadPoolExecutor instead.

### DIFF
--- a/lib/bolt/executor.rb
+++ b/lib/bolt/executor.rb
@@ -27,7 +27,7 @@ module Bolt
 
       @noop = noop
       @run_as = nil
-      @pool = Concurrent::CachedThreadPool.new(max_threads: @config[:concurrency])
+      @pool = Concurrent::ThreadPoolExecutor.new(max_threads: @config[:concurrency])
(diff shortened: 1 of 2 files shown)


donoghuc:
Based on Nick's review (and Alex's original suggestion) was able to implement using promise instead of sleep to keep threads busy. IRB session demonstrates the idea. 
```
irb(main):001:0> require "concurrent"
irb(main):002:0> h = {promise: Concurrent::Promise.new{ "foo" }}
irb(main):003:0> t = Thread.new { puts h[:promise].value }
irb(main):004:0> t.status
=> "sleep"
irb(main):005:0> h[:promise].fulfilled?
=> false
irb(main):006:0> h[:promise].execute
*** foo
irb(main):007:0> t.status
=> false
irb(main):008:0> t.join
=> #<Thread:0x000056272af51430@(irb):3 dead>
irb(main):009:0> h[:promise].fulfilled?
=> true
irb(main):010:0> h[:promise].value
=> "foo"
```

PR #445, opened 2018-05-18
(BOLT-508) Stop using CLIError outside of Bolt::CLI

adreyer:
CLIError was being used as a generic error in many parts of bolt. This
limits the use of CLIError to the CLI class. In most cases error that
were CLIErrors become Bolt::ValidationError or generic Bolt::Errors.
    
CLIError is still probably overused in the CLI class.

### DIFF
--- a/lib/bolt/inventory.rb
+++ b/lib/bolt/inventory.rb
@@ -45,7 +45,7 @@ module Bolt
         begin
           data = YAML.safe_load(ENV[ENVIRONMENT_VAR])
         rescue Psych::Exception
-          raise Bolt::CLIError, "Could not parse inventory from $#{ENVIRONMENT_VAR}"
+          raise Bolt::Error.new("Could not parse inventory from $#{ENVIRONMENT_VAR}", 'bolt/parse-error')
         end
       else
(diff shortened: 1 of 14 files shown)


lib/bolt/inventory.rb
```diff
@@ -45,7 +45,7 @@ def self.from_config(config)
         begin
           data = YAML.safe_load(ENV[ENVIRONMENT_VAR])
         rescue Psych::Exception
-          raise Bolt::CLIError, "Could not parse inventory from $#{ENVIRONMENT_VAR}"
+          raise Bolt::Error, "Could not parse inventory from $#{ENVIRONMENT_VAR}"
```
MikaelSmith: Should this have an issue code?

adreyer: issue code is a puppet thing that we don't really make use of.
This should happen before we get into puppet anyway.

MikaelSmith: Oh, I meant kind, not issue code. I think this might straight-up fail right now, because it's only passing 1 argument but `Bolt::Error` requires 2.

adreyer: ah yeah that will be an argument error

PR #446, opened 2018-05-18
(BOLT-532) Restore message when plan returns no results

MikaelSmith:
A prior change introduced PlanResult, an object that wraps all plan
output. That broke the message when no results are returned; fix it.

### DIFF
--- a/lib/bolt/outputter/human.rb
+++ b/lib/bolt/outputter/human.rb
@@ -161,22 +161,12 @@ module Bolt
         @stream.puts(plan_info)
       end
 
-      # @param [Hash] A hash representing the plan result
-      def print_plan_result(result)
-        if result.nil?
+      # @param [Bolt::PlanResult] A PlanResult object
+      def print_plan_result(plan_result)
+        if plan_result.value.nil?
           @stream.puts("Plan completed successfully with no result")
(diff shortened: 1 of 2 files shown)


lib/bolt/outputter/human.rb
```diff
@@ -161,22 +161,12 @@ def print_plan_info(plan)
...
-        if result.nil?
+      # @param [Bolt::PlanResult] A PlanResult object
+      def print_plan_result(plan_result)
+        if plan_result.value.nil?
           @stream.puts("Plan completed successfully with no result")
-        # Otherwise if object has a json representation display it
-        elsif result.respond_to?(:to_json)
-          # Guard against to_json methods that don't accept options
-          # and don't print empty results on multiple lines
-          if result.method(:to_json).arity == 0 ||
```
MikaelSmith: Do we need this in PlanResult's `to_json`? I thought I remembered this code changing, but I don't see where in the git log.

MikaelSmith: @nicklewis @adreyer I think this is ok, but if you remember where we still need to do special things with turning plan results into json we can come back to this. I think if we do need something, it needs to go in the PlanResult's to_json method anyway.

PR #447, opened 2018-05-21
(BOLT-519) User-Agent header: Orchestrator client

donoghuc:
When orchestrator is used, set User-Agent header to specify bolt.


donoghuc:
See https://github.com/puppetlabs/orchestrator_client-ruby/pull/13 for changes in orchestrator_client-ruby.

MikaelSmith:
We'll need a corresponding version bump in https://github.com/puppetlabs/bolt-vanagon/blob/master/configs/components/rubygem-orchestrator_client.rb once the gem is released.

PR #451, opened 2018-05-24
(BOLT-536) Default input method for PowerShell

donoghuc:
When powershell task is run over winrm, parameters are passed by default as powershell named arguments.

### DIFF
--- a/lib/bolt/transport/winrm.rb
+++ b/lib/bolt/transport/winrm.rb
@@ -104,6 +104,7 @@ catch
         raise "No suitable implementation of #{task.name} for #{target.name}" unless executable
 
         input_method = task.input_method
+        input_method ||= powershell_file?(executable) ? 'powershell' : 'both'
         with_connection(target) do |conn|
           if STDIN_METHODS.include?(input_method)
             stdin = JSON.dump(arguments)
--- a/spec/bolt/transport/winrm_spec.rb
+++ b/spec/bolt/transport/winrm_spec.rb
@@ -485,20 +485,28 @@ PS
       end
     end
 
-    it "can run a task passing input on stdin and environment", winrm: true do
+    it "can run a task passing input on environment", winrm: true do
       contents = <<PS
 Write-Host "$env:PT_message_one $env:PT_message_two"
-$line = [Console]::In.ReadLine()
-Write-Host $line
 PS
       arguments = { message_one: 'Hello from task', message_two: 'Goodbye' }
-      with_task_containing('tasks-test-both-winrm', contents, 'both', '.ps1') do |task|
+      with_task_containing('tasks-test-both-winrm', contents, 'environment', '.ps1') do |task|
         expect(
           winrm.run_task(target, task, arguments).message
-        ).to eq([
-          "Hello from task Goodbye\r\n",
-          "{\"message_one\":\"Hello from task\",\"message_two\":\"Goodbye\"}\r\n"
-        ].join(''))
+        ).to eq("Hello from task Goodbye\r\n")
+      end
+    end
+
+    it "defaults to powershell input method when executing .ps1", winrm: true do
+      contents = <<PS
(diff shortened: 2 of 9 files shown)


donoghuc:
Vendored Puppet (via pal) parses the `input_method`. If `input_method` is not defined in task metadata JSON then the default is set to `both`. 

The *meaningful* `input_method` values are `both`, `stdin`, `environment`, and `powershell`. The implementation in this commit will not allow a user to specify `both` when using powershell. Is this a problem? Will a user ever need multiple input methods with powershell? 

Also I noticed that if you set `input_method` in task metadata to `foo` then there is no 'error checking' and in my solution the default is not honored (simple check is if default `both` is detected then set `input_method` to `powershell`). Is this something I should account for? Or should we implement some kind of validation on `input_method`?

donoghuc:
Updated https://github.com/puppetlabs/tasks-hands-on-lab/pull/46 05-writing-tasks example accordingly

lucywyman:
I might be misunderstanding this code, but I believe that `powershell` is equivalent to `both`.  That isn't entirely clear, so maybe having a `--powershell-params` flag or something would be clearer? But I doubt that's terribly high priority for us.

In my opinion we should have some type checking for command line args, especially here since it's an easy + limited set of possible arguments. But I'm not sure if there's a good reason for that not being added in the first place.

lib/bolt/transport/winrm.rb
```diff
@@ -99,6 +99,9 @@ def run_script(target, script, arguments, _options = {})
 
       def run_task(target, task, arguments, _options = {})
         input_method = task.input_method
+        if powershell_file?(task.executable)
+          input_method = 'powershell' if input_method == 'both'
```
MikaelSmith: This feels a little weird to me. I think it's valid to put `both` explicitly in the metadata. I feel like we should only be setting `input_method` to `powershell` if nothing was explicitly added in metadata.

donoghuc: I sensed the weirdness with this solution. I did it this way because I did not want to mess with the vendored puppet code. 

I am trying to figure out what changing the default input_method in puppet would look like. Could it be as simple as setting the input_method https://github.com/puppetlabs/puppet/blob/5b9f49b6f75a8453e496e2567bdc89a3a3129eb4/lib/puppet/pops/pcore.rb#L74 at provided link to something like "default" and then dealing with setting (and possibly validating) the input_method in the various transport implementations in bolt? 

What kind of impact would changing that default value have outside of the Bolt transport  implementations? It seems like the current default value of "both" is somewhat misleading given that there are at least three valid options (`stdin`, `environment`, `powershell`, + others in the future? ).

donoghuc:
This will fail CI until https://github.com/puppetlabs/puppet/pull/6860 is incorporated into vendored puppet. PUP-8898 will set default value of `input_method` to `nil` instead of `both`.

MikaelSmith:
PUP work is merged, so this could be updated.

spec/bolt/transport/winrm_spec.rb
```diff
@@ -485,20 +485,28 @@ def stub_winrm_to_raise(klass, message)
...
           winrm.run_task(target, task, arguments).message
-        ).to eq([
-          "Hello from task Goodbye\r\n",
-          "{\"message_one\":\"Hello from task\",\"message_two\":\"Goodbye\"}\r\n"
-        ].join(''))
+        ).to eq("Hello from task Goodbye\r\n")
+      end
+    end
+
+    it "defualts to powershell input method when executing .ps1", winrm: true do
```
MikaelSmith: `defualts` -> `defaults`

MikaelSmith:
I'd also lean towards squashing all the commits. They're a single change. In particular, it's nice to have tests pass at each commit (so you can git bisect) and with the way this is broken up they wouldn't.

PR #452, opened 2018-05-29
(BOLT-521) Allow users to specify the run-as-command for ssh

adreyer:
This adds a new config option run-as-command to the ssh transport. This
command can be used to override the sudo support either to use a
different command than sudo or to specify which sudo options to use. It
will not use the sudo-password option.

### DIFF
--- a/lib/bolt/transport/ssh.rb
+++ b/lib/bolt/transport/ssh.rb
@@ -9,7 +9,7 @@ module Bolt
   module Transport
     class SSH < Base
       def self.options
-        %w[port user password sudo-password private-key host-key-check connect-timeout tmpdir run-as tty]
+        %w[port user password sudo-password private-key host-key-check connect-timeout tmpdir run-as tty run-as-command]
       end
 
       PROVIDED_FEATURES = ['shell'].freeze
@@ -38,6 +38,11 @@ module Bolt
         unless timeout_value.is_a?(Integer) || timeout_value.nil?
           error_msg = "connect-timeout value must be an Integer, received #{timeout_value}:#{timeout_value.class}"
           raise Bolt::ValidationError, error_msg
+        end
+
+        run_as_cmd = options['run-as-command']
+        if run_as_cmd && (!run_as_cmd.is_a?(Array) || run_as_cmd.any? { |n| !n.is_a?(String) })
+          raise Bolt::ValidationError, "run-as-command must be an Array of Strings, received #{run_as_cmd}"
         end
       end
 
--- a/lib/bolt/transport/ssh/connection.rb
+++ b/lib/bolt/transport/ssh/connection.rb
@@ -184,14 +184,20 @@ module Bolt
         def execute(command, sudoable: false, **options)
           result_output = Bolt::Node::Output.new
           run_as = options[:run_as] || self.run_as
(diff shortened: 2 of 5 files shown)


lib/bolt/transport/ssh/connection.rb
```diff
@@ -184,10 +184,14 @@ def handled_sudo(channel, data)
         def execute(command, sudoable: false, **options)
           result_output = Bolt::Node::Output.new
           run_as = options[:run_as] || self.run_as
-          use_sudo = sudoable && run_as && @user != run_as
+          use_sudo = sudoable && run_as && @user != run_as && options['run-as-command'].nil?
 
           command_str = command.is_a?(String) ? command : Shellwords.shelljoin(command)
-          if use_sudo
+          if sudoable && run_as && @user != run_as &&
```
MikaelSmith: This seems like it could be its own variable, since we use it twice. Minor nit.

pre-docs/bolt_configuration_options.md
```diff
@@ -39,8 +39,12 @@ groups of nodes on the commandline and from plans.
 
 `connect-timeout`: How long Bolt should wait when establishing connections.
 
-`tmpdir`: The directory to upload and execute temporary files on the target.
+`run-as-command`: The command to escalate permissions. The user to escalate to
```
MikaelSmith: We should call out that this must be an array, allowing extra options to be supplied.

lib/bolt/transport/ssh.rb
```diff
@@ -9,7 +9,7 @@ module Bolt
   module Transport
     class SSH < Base
       def self.options
-        %w[port user password sudo-password private-key host-key-check connect-timeout tmpdir run-as]
+        %w[port user password sudo-password private-key host-key-check connect-timeout tmpdir run-as run-as-command]
```
MikaelSmith: We should add a section in `validate` that asserts `run-as-command` is an array of strings.

MikaelSmith:
Looks like just rubocop issues.

lib/bolt/transport/ssh.rb
```diff
@@ -39,6 +39,11 @@ def self.validate(options)
           error_msg = "connect-timeout value must be an Integer, received #{timeout_value}:#{timeout_value.class}"
           raise Bolt::ValidationError, error_msg
         end
+
+        run_as_cmd = options['run-as-command']
+        if run_as_cmd && (!run_as_cmd.is_a?(Array) || run_as_command.any? {|n| !n.is_a?(String)})
```
MikaelSmith: `run_as_command` isn't a variable, it's `run_as_cmd`.

spec/bolt/config_spec.rb
```diff
@@ -185,6 +185,18 @@
...
     end
 
+   it "does not accept a non-array for run-as-command" do
+      config = {
+        transports: {
+          ssh: { 'run-as-command' => 'sudo -n' }
+        }
+      }
+      expect {
+        Bolt::Config.new(config).validate
```
MikaelSmith: We should probably have a positive test as well. None of the tests caught the broken validation check.

MikaelSmith:
Apparently needs a rebase now. Not sure where that came from.

PR #454, opened 2018-05-30
(BOLT-135) Add semantic_puppet dependency to gemspec

MikaelSmith:
Without this, gem builds will fail when using tasks and plans because
Puppet won't find its dependency.


MikaelSmith:
`be rake test:git GIT_FORK=MikaelSmith/bolt GIT_BRANCH=add-semantic-to-gemspec` passes for me now.

PR #455, opened 2018-05-30
(BOLT-399) Add `facts` metadata task

MikaelSmith:
Adds `facts` metadata-only task that dispatches to specific
implementations based on features in transports. Use it to simplify
`facts` plans.


lucywyman:
This looks good to me! Passes local tests, and I don't see any missed "facts::implementation" calls.

PR #460, opened 2018-05-30
(BOLT-544) add secondary default for puppetdb.conf

donoghuc:
Try to load puppetdb config first from user supplied configfile, second from ~/.puppetlabs/client-tools, finally from /etc/puppetlabs/client-tools. This simplifies Bolt-PE integration out of the box.

### DIFF
--- a/lib/bolt/puppetdb/config.rb
+++ b/lib/bolt/puppetdb/config.rb
@@ -1,30 +1,35 @@
 # frozen_string_literal: true
 
 require 'json'
+require 'bolt/util'
 
 module Bolt
   module PuppetDB
     class Config
       DEFAULT_TOKEN = File.expand_path('~/.puppetlabs/token')
-      DEFAULT_CONFIG = File.expand_path('~/.puppetlabs/client-tools/puppetdb.conf')
+      DEFAULT_CONFIG = { user: File.expand_path('~/.puppetlabs/client-tools/puppetdb.conf'),
+                         global: '/etc/puppetlabs/client-tools/puppetdb.conf',
+                         win_global: 'C:/ProgramData/PuppetLabs/client-tools/puppetdb.conf' }.freeze
 
       def initialize(config_file, options)
         @settings = load_config(config_file)
         @settings.merge!(options)
-
         expand_paths
         validate
       end
 
       def load_config(filename)
+        global_path = Bolt::Util.windows? ? DEFAULT_CONFIG[:win_global] : DEFAULT_CONFIG[:global]
         if filename
           if File.exist?(filename)
             config = JSON.parse(File.read(filename))
           else
             raise Bolt::PuppetDBError, "config file #{filename} does not exist"
           end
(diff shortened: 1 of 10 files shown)


donoghuc:
Thinking of tests for this. Possibly mock out configs at the expected default locations?

MikaelSmith:
spec tests that mock filesystem lookup would probably cover it. Verify that if the user client-tools config doesn't exist, it tries to open the system client-tools config.

adreyer:
We should update the docs to mention global config.

lib/bolt/puppetdb/config.rb
```diff
@@ -6,7 +6,8 @@ module Bolt
   module PuppetDB
     class Config
       DEFAULT_TOKEN = File.expand_path('~/.puppetlabs/token')
-      DEFAULT_CONFIG = File.expand_path('~/.puppetlabs/client-tools/puppetdb.conf')
+      DEFAULT_CONFIG = { user: File.expand_path('~/.puppetlabs/client-tools/puppetdb.conf'),
+                         global: '/etc/puppetlabs/client-tools/puppetdb.conf' }.freeze
```
nicklewis: This location different between Windows and *nix. See [puppet-specifications](https://github.com/puppetlabs/puppet-specifications/blob/master/file_paths.md#puppet-agent-windows) for more detail.

nicklewis: Here's an example of how we check for Windows now:

https://github.com/puppetlabs/bolt/blob/15995af05ae4ac189d970b34a783ffeffb3b4927/lib/bolt/transport/ssh/connection.rb#L62

adreyer: Can we add that to Bolt::Util

donoghuc:
For this work a new method called `windows?` was added to Bolt::Util. This could replace the following two patterns for determining windows OS. Should I replace here, start a new ticket, or leave them alone? 

File::ALT_SEPARATOR
https://github.com/puppetlabs/bolt/blob/15995af05ae4ac189d970b34a783ffeffb3b4927/lib/bolt/transport/ssh/connection.rb#L62
https://github.com/puppetlabs/bolt/blob/15995af05ae4ac189d970b34a783ffeffb3b4927/lib/bolt/transport/ssh/connection.rb#L105

Gem.win_platform?
https://github.com/puppetlabs/bolt/blob/a53f2f872a82b2a5ba6a4d8961243f2cb7c1ce1f/spec/signal_helper.rb#L36
https://github.com/puppetlabs/bolt/blob/46cddd57a8b78e3d9f621f69fe01dbcda6dccd6f/lib/bolt/transport/local.rb#L23
https://github.com/puppetlabs/bolt/blob/dae704d9e565b9997e964b02caa906a80b0c2fb6/lib/bolt/pal.rb#L57

MikaelSmith:
As a refactor, I don't believe it needs a ticket. Otherwise, do whichever is simplest for you. Personally I'd find it easier to add a refactor commit to my existing PR.

PR #462, opened 2018-06-01
(BOLT-488) Submit data to Google Analytics for each command invocation

nicklewis:
Every time the user runs a command (`bolt task run`, `bolt plan show`),
we now submit a "screenview" event to Google Analytics, where the
"screen" is the command that was run. The information included is:

a) bolt version
b) user locale
c) the "screen" run
d) a unique, random user id

IP addresses are anonymized, and we do *not* include the name of the
task or plan run.

The unique ID is stored in ~/.puppetlabs/bolt/analytics.yaml. That
file also accepts the setting `disabled: true` which will disable the
collection of analytics.

To avoid polluting the data with development and test runs, we also use
a BOLT_DISABLE_ANALYTICS environment variable, which is automatically
set when running via bundler and while running the tests.

Analytics submission is performed in the background and errors are
ignored, both so as not to disrupt normal usage of bolt. There is a
grace period of 250ms at the end of the application to allow any
outstanding event submissions to complete without introducing a
noticeable hang.

The complete data submitted is logged at debug level so that users can
be aware of the information we're collecting.

### DIFF
--- a/acceptance/setup/common/pre-suite/050_build_bolt_inventory.rb
+++ b/acceptance/setup/common/pre-suite/050_build_bolt_inventory.rb
@@ -32,4 +32,6 @@ test_name "build bolt inventory file" do
 
   on bolt, "mkdir -p #{bolt_confdir}"
   create_remote_file(bolt, "#{bolt_confdir}/inventory.yaml", inventory.to_yaml)
+
+  create_remote_file(bolt, "#{bolt_confdir}/analytics.yaml", { 'disabled' => true }.to_yaml)
 end
--- a/lib/bolt/analytics.rb
+++ b/lib/bolt/analytics.rb
@@ -1,0 +1,143 @@
+# frozen_string_literal: true
+
+require 'bolt/version'
+require 'httpclient'
+require 'json'
+require 'locale'
+require 'logging'
+require 'securerandom'
+
+module Bolt
+  module Analytics
+    PROTOCOL_VERSION = 1
+    APPLICATION_NAME = 'bolt'
+    TRACKING_ID = 'UA-120367942-1'
+    TRACKING_URL = 'https://google-analytics.com/collect'
+
+    def self.build_client
+      logger = Logging.logger[self]
+
+      config_file = File.expand_path('~/.puppetlabs/bolt/analytics.yaml')
+      config = load_config(config_file)
+
+      if config['disabled'] || ENV['BOLT_DISABLE_ANALYTICS']
+        logger.debug "Analytics opt-out is set, analytics will be disabled"
+        NoopClient.new
+      else
+        unless config.key?('user-id')
+          config['user-id'] = SecureRandom.uuid
+          write_config(config_file, config)
+        end
+
+        Client.new(config['user-id'])
+      end
+    rescue StandardError => e
+      logger.debug "Failed to initialize analytics client, analytics will be disabled: #{e}"
+      NoopClient.new
+    end
+
+    def self.load_config(filename)
+      if File.exist?(filename)
+        YAML.load_file(filename)
+      else
+        {}
+      end
+    end
+
+    def self.write_config(filename, config)
+      FileUtils.mkdir_p(File.dirname(filename))
+      File.write(filename, config.to_yaml)
+    end
+
+    class Client
+      attr_reader :user_id
+
+      def initialize(user_id)
+        @logger = Logging.logger[self]
+        @http = HTTPClient.new
+        @user_id = user_id
+        @executor = Concurrent.global_io_executor
+      end
+
+      def screen_view(screen)
+        screen_view_params = {
+          # Type
+          t: 'screenview',
+          # Screen Name
+          cd: screen
+        }
+
+        submit(base_params.merge(screen_view_params))
+      end
+
+      def event(category, action)
+        event_params = {
+          # Type
+          t: 'event',
+          # Event Category
+          ec: category,
+          # Event Action
+          ea: action
+        }
+
+        submit(base_params.merge(event_params))
+      end
+
+      def submit(params)
+        # Handle analytics submission in the background to avoid blocking the
+        # app or polluting the log with errors
+        Concurrent::Future.execute(executor: @executor) do
+          @logger.debug "Submitting analytics: #{JSON.pretty_generate(params)}"
+          @http.post(TRACKING_URL, params)
+          @logger.debug "Completed analytics submission"
+        end
+      end
+
+      # These parameters have terrible names. See this page for complete documentation:
+      # https://developers.google.com/analytics/devguides/collection/protocol/v1/parameters
+      def base_params
+        {
+          v: PROTOCOL_VERSION,
+          # Client ID
+          cid: @user_id,
+          # Tracking ID
+          tid: TRACKING_ID,
+          # Application Name
+          an: APPLICATION_NAME,
+          # Application Version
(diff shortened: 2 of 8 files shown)


nicklewis:
This is still using a placeholder project id while I wait for the real one to be created.

MikaelSmith:
This is failing some rubocop rules.

lib/bolt/cli.rb
```diff
@@ -493,6 +495,15 @@ def execute(options)
         exit!
       end
 
+      @analytics = Bolt::Analytics.build_client
```
adreyer: This won't catch help, right. Should it?

lib/bolt/analytics.rb
```diff
@@ -0,0 +1,141 @@
...
+    ProtocolVersion = 1
+    ApplicationName = 'bolt'
+    TrackingID = 'UA-fakefakefake-1'
+    TrackingURL = 'https://google-analytics.com/collect'
+
+    def self.build_client
+      logger = Logging.logger[self]
+
+      config_file = File.expand_path('~/.puppetlabs/bolt/analytics.yaml')
+      config = load_config(config_file)
```
adreyer: doesn't this need to be load_config(nil, [config_file]) to avoid erroring when the file is absent

lib/bolt/analytics.rb
```diff
@@ -0,0 +1,141 @@
...
+
+      if config['disabled'] || ENV['BOLT_DISABLE_ANALYTICS']
+        logger.debug "Analytics opt-out is set, analytics will be disabled"
+        NoopClient.new
+      else
+        if !config.key?('user-id')
+          config['user-id'] = SecureRandom.uuid
+          write_config(config_file, config)
```
adreyer: Should this make the directory first?

lib/bolt/analytics.rb
```diff
@@ -0,0 +1,141 @@
...
+          # Tracking ID
+          tid: TrackingID,
+          # Application Name
+          an: ApplicationName,
+          # Application Version
+          av: Bolt::VERSION,
+          # Anonymize IPs
+          aip: true,
+          # User locale
+          ul: Locale.current.to_rfc
```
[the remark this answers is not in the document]
nicklewis: I don't believe so.

```rb
# * Returns: an Array of the current locales (Locale::Tag's class).
def current
  unless Thread.current[:current_languages]
    loc = driver_module.locales
    Thread.current[:current_languages] = loc ? loc : Locale::TagList.new([default])
  end
  Thread.current[:current_languages]
end
```

It seems it will return a default if no other locale is specified.

lib/bolt/cli.rb
nicklewis: I left that out because it was sort of unclear what that means. We have both `bolt help task` and `bolt task --help`. Would those both just be "help" or would we want to try to distinguish what they were asking for help about?

lib/bolt/analytics.rb
nicklewis: This is using a `load_config` method from the Analytics class, not the standard `read_config_file` util method. The reason is mostly that we don't want to swallow exceptions. If the config file *doesn't exist*, this `load_config` method will return a default empty config. But if the config file can't be read for some *other* reason (permissions, bad YAML, etc) it will be handled by the `rescue` in this method, which will err on the side of caution and disable analytics.

lib/bolt/analytics.rb
nicklewis: Good call.

nicklewis:
Updated to fix the rubocop issues and to create the directory before writing the analytics config.

nicklewis:
Added a docs page about what information is collected and how to opt-out, with a link from the installation instructions. @katelopresti

acceptance/setup/common/pre-suite/050_build_bolt_inventory.rb
```diff
@@ -32,4 +32,6 @@
 
   on bolt, "mkdir -p #{bolt_confdir}"
   create_remote_file(bolt, "#{bolt_confdir}/inventory.yaml", inventory.to_yaml)
+
+  create_remote_file(bolt, "#{bolt_confdir}/analytics.yaml", {disabled: true}.to_yaml)
```
MikaelSmith: rubocop

spec/bolt/analytics_spec.rb
```diff
@@ -0,0 +1,96 @@
...
+require 'spec_helper'
+require 'bolt/analytics'
+
+describe Bolt::Analytics do
+  let(:default_config) { {} }
+
+  before :each do
+    # We use a hard override to disable analytics for tests, but that obviously
+    # interferes with these tests...
+    ENV.delete('BOLT_DISABLE_ANALYTICS')
```
MikaelSmith: Is this reverted after the test? Seems like it might need to be `:around`.

MikaelSmith: Oh, we reapply it before each test.

nicklewis: The spec_helper sets it in a `before :each`, so it should be reset for future tests.

nicklewis:
Updated. I think I've appeased rubocop now.

lib/bolt/analytics.rb
```diff
@@ -0,0 +1,143 @@
...
+require 'json'
+require 'locale'
+require 'logging'
+require 'securerandom'
+
+module Bolt
+  module Analytics
+    PROTOCOL_VERSION = 1
+    APPLICATION_NAME = 'bolt'
+    TRACKING_ID = 'UA-fakefakefake-1'
```
MikaelSmith: Still waiting on a TRACKING_ID?

lib/bolt/analytics.rb
```diff
@@ -0,0 +1,143 @@
...
+      end
+
+      # If the user is running a very fast command, there may not be time for
+      # analytics submission to complete before the command is finished. In
+      # that case, we give a little buffer for any stragglers to finish up.
+      # 250ms strikes a balance between accomodating slower networks while not
+      # introducing a noticeable "hang".
+      def finish
+        @executor.shutdown
+        @executor.wait_for_termination(0.25)
```
MikaelSmith: I could see spinning off another process to handle submissions so we can shut Bolt down immediately, but that's a far more complicated architecture that's probably not warranted yet.

lib/bolt/analytics.rb
nicklewis: Yeah. The ticket has been set "in progress", so I'm hopeful.

lib/bolt/analytics.rb
nicklewis: Right now we only make a single request, at the very start, so this is only an issue if you're doing something *very* fast. I can only make it cause a problem if I'm running something over the `local` transport.

In the future, if we're submitting events, they might happen toward the end, but they'll also reuse the existing HTTP connection and should be *significantly* faster.

lib/bolt/cli.rb
```diff
@@ -6,6 +6,8 @@
 require 'io/console'
 require 'logging'
 require 'optparse'
+require 'securerandom'
```
MikaelSmith: Why did you bring in securerandom?

nicklewis: Whoops, was generating the uuid here in a prior iteration. Will remove this.

nicklewis:
I removed the stray `require` and also fixed a bug that was causing data to be sent during tests.

PR #464, opened 2018-06-04
(BOLT-533) Move facts module tests to bolt spec tests

lucywyman:
This PR moves the spec tests from the facts module included with bolt to live with the 'general' spec tests. This is so that we can maintain some spec testing while rewriting most of the spec tests to be beaker acceptance tests for the actual puppetlabs-facts module that we're replacing the local facts module with.

### DIFF
--- a/spec/modules/facts/functions/group_by_spec.rb
+++ b/spec/modules/facts/functions/group_by_spec.rb
@@ -1,0 +1,22 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+describe 'facts::group_by' do
+  it 'delegates to the native group_by function' do
+    collection = mock('Puppet::Pops::Types::Iterable').extend(Puppet::Pops::Types::Iterable)
+    return_value = {}
+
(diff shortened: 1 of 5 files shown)


spec/modules/facts/functions/group_by_spec.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+describe 'facts::group_by' do
```
adreyer: I think we can safely test this in the module since it doesn't actually run tasks

spec/modules/facts/plans/info_spec.rb
```diff
@@ -0,0 +1,97 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+require 'bolt_spec/plans'
+
+describe 'facts::info' do
```
adreyer: We should be able to move this to the module tests since it mocks out the task running. 
There may be some work to do to expose the plan mock framework for those tests though

PR #465, opened 2018-06-04
(BOLT-534) Fix sudo hanging when user shell is bash

MikaelSmith:
When run-as is requested without a password but the password is required we should fail. However when attempting that against targets where the user's login shell is bash, a 2nd sudo attempt would hang. Fix the issue causing us to attempt a 2nd sudo in the first place, and close the connection after when no password is available to prevent future issues.

### DIFF
--- a/spec/bolt/transport/ssh_spec.rb
+++ b/spec/bolt/transport/ssh_spec.rb
@@ -22,6 +22,8 @@ describe Bolt::Transport::SSH do
   let(:hostname) { ENV['BOLT_SSH_HOST'] || "localhost" }
   let(:user) { ENV['BOLT_SSH_USER'] || "bolt" }
   let(:password) { ENV['BOLT_SSH_PASSWORD'] || "bolt" }
+  let(:bash_user) { 'test' }
+  let(:bash_password) { 'test' }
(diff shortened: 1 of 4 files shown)


spec/bolt/transport/ssh_spec.rb
```diff
@@ -515,9 +516,24 @@ def result_value(stdout = nil, stderr = nil, exit_code = 0)
...
+                         "Sudo password for user #{user} was not provided for #{hostname}:#{port}")
+      end
+    end
+
+    context "as bash user with no password" do
+      let(:config) {
+        mk_config('host-key-check' => false, 'run-as' => 'root', user: bash_user, password: bash_password)
+      }
+
+      it "returns a failed result when a temporary directory is created", ssh: true do
```
nicklewis: This test doesn't seem to fail when I revert the corresponding code changes.

nicklewis: Or was this just intended as added coverage for an existing case, rather than a test of the new behavior?

nicklewis: I spoke too soon. It does fail (well, hangs) if I revert *both* the `chown` change and the sudo change. I had previously only reverted one.

PR #467, opened 2018-06-05
(BOLT-489) Report user OS and version to google analytics

nicklewis:
This adds a new custom dimension for operating system. We use Facter to
compute the value, and we do so asynchronously after creating the
analytics client. The overhead to load facter and compute OS is ~100ms,
which should be negligible for all but the speediest Bolt invocations.

### DIFF
--- a/bolt-modules/boltlib/spec/spec_helper.rb
+++ b/bolt-modules/boltlib/spec/spec_helper.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require_relative '../../../vendored/require_vendored.rb'
+require 'puppet_pal'
 
 # Ensure tasks are enabled when rspec-puppet sets up an environment
 # so we get task loaders.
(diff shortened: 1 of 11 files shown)


lib/bolt/analytics.rb
```diff
@@ -109,10 +110,21 @@ def base_params
...
+          # Custom Dimension 1 (Operating System)
+          cd1: @os.value
         }
       end
 
+      def compute_os
+        Concurrent::Future.execute(executor: @executor) do
+          require_relative '../../vendored/require_vendored'
+          require 'facter'
+          os = Facter.value('os')
```
MikaelSmith: There goes my plan of stubbing out Facter :(

nicklewis:
This is difficult to test, because the custom dimension isn't available in the real-time data...

bolt-modules/boltlib/spec/spec_helper.rb
```diff
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require_relative '../../../vendored/require_vendored.rb'
+require 'puppet_pal'
```
MikaelSmith: 👍

PR #468, opened 2018-06-05
(BOLT-533) externalize facts module

lucywyman:


lucywyman:
Depends on puppetlabs-facts 0.1.0 being published

PR #470, opened 2018-06-06
(BOLT-490) Collect events for each transport used

nicklewis:
We now send one event per-transport when a transport is first
initialized. This allows us to understand how commonly each transport is
being used and whether they're being used together.

### DIFF
--- a/lib/bolt/executor.rb
+++ b/lib/bolt/executor.rb
@@ -5,6 +5,8 @@ require 'English'
 require 'json'
 require 'concurrent'
 require 'logging'
+require 'set'
+require 'bolt/analytics'
 require 'bolt/result'
 require 'bolt/config'
 require 'bolt/notifier'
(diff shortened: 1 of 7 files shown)


nicklewis:
This currently also sends the number of nodes associated with the transport. If you're running a plan, that number will be however many were being run in the *first* command/script/task invocation of the plan. Does that seem reasonable?

MikaelSmith:
Some tests are failing.

nicklewis:
Tests are 🍏now

spec/bolt/executor_spec.rb
```diff
@@ -8,7 +8,8 @@
   include BoltSpec::Task
 
   let(:config) { Bolt::Config.new(concurrency: 1) }
-  let(:executor) { Bolt::Executor.new(config) }
+  let(:analytics) { Bolt::Analytics::NoopClient.new }
```
MikaelSmith: This is no longer necessary, but doesn't hurt.

lib/bolt/executor.rb
```diff
@@ -50,6 +52,7 @@ def transport(transport)
     def batch_execute(targets)
       promises = targets.group_by(&:protocol).flat_map do |protocol, protocol_targets|
         transport = transport(protocol)
+        report_transport(transport, protocol_targets.count)
```
MikaelSmith: This appears to be reporting every time we use a transport.

lib/bolt/executor.rb
```diff
@@ -16,13 +17,17 @@ class Executor
...
+    def initialize(config = Bolt::Config.new, analytics = Bolt::Analytics::NoopClient.new, noop = nil)
       @config = config
+      @analytics = analytics
       @logger = Logging.logger[self]
       @plan_logging = false
 
       @transports = Bolt::TRANSPORTS.each_with_object({}) do |(key, val), coll|
-        coll[key.to_s] = Concurrent::Delay.new { val.new }
+        coll[key.to_s] = Concurrent::Delay.new do
+          report_transport(key.to_s)
```
MikaelSmith: Turns out Orch is initialized every time. So we'll need to use something else if we only want to send it on initialization.

nicklewis:
Updated to really really only send once (now with a test for the multiple jobs on one executor case), and only for transports that are actually used. I also restored the behavior of sending the node count.

PR #472, opened 2018-06-08
(BOLT-558) Update doc strings for Bolt core Puppet functions

MikaelSmith:


bolt-modules/boltlib/lib/puppet/functions/run_plan.rb
```diff
@@ -2,21 +2,18 @@
...
+  # @param plan_name The plan to run.
+  # @param named_args Arguments to the plan. Can also include additional options: '_catch_errors', '_run_as'.
+  # @return The result of running the plan. Undef if plan does not explicitly return results.
+  # @example Run a plan
+  #   run_plan('canary', 'command' => 'false', 'nodes' => $targets, '_catch_errors' => true)
   dispatch :run_plan do
     scope_param
     param 'String', :plan_name
     optional_param 'Hash', :named_args
+    return_type 'PlanResult'
```
adreyer: Is this Optional?

MikaelSmith: I don't think it needs to be, but it was going to be messy to get spec tests working because they type wasn't defined there so I removed it again.

PR #473, opened 2018-06-08
(BOLT-559) Add `rake docs` and `rake json` to generate Bolt docs

MikaelSmith:
`rake docs` generates docs for Bolt, it's core functions, and packaged
modules.

`rake json` generates a JSON file containing docs for Bolt's core
functions.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -4,6 +4,10 @@ require "bundler/gem_tasks"
 require "rspec/core/rake_task"
 require "rubocop/rake_task"
 
+require_relative 'vendored/require_vendored'
+require "puppet-strings/tasks/generate"
+require "fileutils"
+
 desc "Run all RSpec tests"
 RSpec::Core::RakeTask.new(:spec)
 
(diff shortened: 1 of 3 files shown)


MikaelSmith:
This is the best I could come up with for now. I didn't see a clear way to have puppet strings generate both the markdown docs and the JSON file in the same rake task. Adding the json argument to puppet strings' rake task stops outputting markdown, and it has an explicit check against trying to do both.

I also didn't include the packaged modules in the JSON output because it wasn't capturing the plan docs, making the output mostly useless for those modules.

Rakefile
```diff
@@ -33,6 +33,24 @@ task :default do
...
+desc "Generate docs for Bolt and its modules"
+task :docs do
+  require_relative 'vendored/require_vendored'
+  require "puppet-strings/tasks/generate"
+  module_paths = %w[bolt-modules/boltlib modules/aggregate modules/canary modules/puppetdb_fact].join(' ')
+  Rake::Task['strings:generate'].invoke(nil, nil, nil, nil, nil, nil, module_paths)
+end
+
+desc "Generate a JSON file containing docs for Bolt's core Puppet functions"
+task :json do
```
lucywyman: This also seems to require that the `doc` directory doesn't already exist. Do we want to remove it first, or put the json in a different directory?

MikaelSmith:
Removed modules directory from markdown output.

lucywyman:
@MikaelSmith Is there a reason they don't want users to generate markdown and json? Just curious.

Rakefile
```diff
@@ -33,6 +33,23 @@ task :default do
   system "rake --tasks"
 end
 
+desc "Generate docs for Bolt and its core Puppet functions"
+task :docs do
+  require_relative 'vendored/require_vendored'
+  require "puppet-strings/tasks/generate"
+  Rake::Task['strings:generate'].invoke(nil, nil, nil, nil, nil, nil, 'bolt-modules/boltlib')
```
MikaelSmith: This actually generates generic yardocs. I'll switch it to generating markdown.

adreyer:
I don't think there is a good reason to commit the JSON

PR #477, opened 2018-06-12
(BOLT-491) Collect number of targeted nodes for each bolt command

nicklewis:
We now include the number of targeted nodes with the screen_view hit
that we submit each time a bolt command is run. This counts the
*resolved* targets, after groups and PuppetDB queries have been taken
into consideration. This does not account for the number of nodes run as
part of a plan, unless those nodes were passed using --nodes or --query.

### DIFF
--- a/lib/bolt/analytics.rb
+++ b/lib/bolt/analytics.rb
@@ -17,7 +17,8 @@ module Bolt
     CUSTOM_DIMENSIONS = {
       operating_system: :cd1,
       inventory_nodes: :cd2,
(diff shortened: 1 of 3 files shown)


nicklewis:
Putting this up for discussion. It doesn't account for nodes targeted by a plan that *weren't* passed as the `nodes` parameter. It's less obvious a) how to work that into the existing code, b) how to submit that as a data point, unless we move the screen_view to the end of execution, in which case it is likely to block the program from completing.

PR #478, opened 2018-06-12
(BOLT-557) Pass _task param to tasks

lucywyman:


lib/bolt/executor.rb
```diff
@@ -169,6 +169,7 @@ def run_task(targets, task, arguments, options = {}, &callback)
       description = options.fetch('_description', "task #{task.name}")
       log_action(description, targets) do
         notify = proc { |event| @notifier.notify(callback, event) if callback }
+        options = { '_task' => task.name }.merge(options)
```
adreyer: What happens when this is sent to orchestrator?
this should probably just set the value after the merge on the next line. We don't want to favor a `_task` key passed  in from elsewhere

pre-docs/sharing_code_among_tasks.md
```diff
@@ -0,0 +1,41 @@
+
+# Sharing Code Among Tasks
```
adreyer: This probably belongs in the exisiting writing tasks page

pre-docs/sharing_code_among_tasks.md
```diff
@@ -0,0 +1,41 @@
...
+
+Tasks can use libraries which live alongside the task code, which one or many tasks can access. Tasks can relatively include
```
adreyer: I don't think this is true yet. This feature only allows multiple tasks to use the same executable

lib/bolt/executor.rb
lucywyman: Like so?
```
        options = { '_run_as' => run_as }.merge(options) if run_as
        options = options.merge('_task' => task.name)
```

Gemfile
```diff
@@ -20,6 +20,7 @@ group(:test) do
 end
 
 group(:development) do
+  gem "puppet-strings", "~> 2.0"
```
adreyer: How did this get in here?

lib/bolt/executor.rb
```diff
@@ -170,6 +170,7 @@ def run_task(targets, task, arguments, options = {}, &callback)
       log_action(description, targets) do
         notify = proc { |event| @notifier.notify(callback, event) if callback }
         options = { '_run_as' => run_as }.merge(options) if run_as
+        options = options.merge('_task' => task.name)
```
adreyer: There isn't a good reason to merge and create a new hash here. We can just options['_task'] = task.name

pre-docs/writing_tasks.md
```diff
@@ -276,6 +276,48 @@ The following features are defined by default:
 * `shell`: present if the target has a posix shell
 * `powershell`: present if the target has powershell
 
+### Tasks with multiple files
```
adreyer: Isn't this "multiple tasks with one file?

I don't see how require_relative worksbecause the other file won't be available on the target?
Where does the `run_task` method below come from?

This feature is more about having one init.rb task and using it in the `implementations` section of `hello.json` and `goodbye.json`

spec/bolt/executor_spec.rb
```diff
@@ -177,7 +178,7 @@ def mock_node_results(_run_as = nil)
       node_results.each do |target, result|
         expect(ssh)
           .to receive(:run_task)
-          .with(target, task_type(task), task_arguments, {})
+          .with(target, task_type(task), task_arguments, _task)
```
adreyer: I think just rewrite the _task options here

lucywyman: What's wrong with having it as a variable?

lib/bolt/executor.rb
adreyer: I think just `options['_task'] = task.name` is more readable and performant

spec/bolt/executor_spec.rb
adreyer: You're not using it consistently and it's weirdly named. I would expect `_task` to be `service::restart` not an opts hash containing only the `_task` options

pre-docs/writing_tasks.md
```diff
@@ -276,6 +276,80 @@ The following features are defined by default:
...
+For example, the following creates 2 tasks `service::stop` and `service::start`, which can live in the executable but appear as 2 separate tasks.
+
+`service/tasks/init.rb`
+```ruby
+#!/usr/bin/env ruby
+
+require 'json'
+require 'open3'
+
+def execute(*cmd)
```
adreyer: I think ditch the error handling and JSON output to keep this shorter
```ruby
#!/usr/bin/env ruby
params = JSON.parse(STDIN.read)
action = params['action'] || params['_task']
if ['start',  'stop'].include?(action)
  `systemctl #{params['_task']} #{params['service']}
end
```

lib/bolt/executor.rb
```diff
@@ -170,6 +170,7 @@ def run_task(targets, task, arguments, options = {}, &callback)
       log_action(description, targets) do
         notify = proc { |event| @notifier.notify(callback, event) if callback }
         options = { '_run_as' => run_as }.merge(options) if run_as
+        options['_task'] = task.name
```
nicklewis: I think we want to add this to `arguments`, not to `options`. The options hash is for parameters that affect *how* the task is run, not for parameters that are passed *to* the task. Right now, this seems like it just passes `_task` to the transport, which ignores it.

PR #481, opened 2018-06-15
(test) DITA source files output to markdown

katelopresti:
This file is an example of DITA-to-markdown output created in the XML editor.


MikaelSmith:
That seems pretty good.

PR #486, opened 2018-06-22
(BOLT-580)  Bolt server POC

lucywyman:


lucywyman:
To use these changes:
```
gem build bolt.gemspec
gem install bolt-0.21.0.gem
./temp/server-client.rb
```

lib/bolt.rb
```diff
@@ -2,4 +2,5 @@
 
 module Bolt
   require 'bolt/executor'
+  require 'bolt/orch_executor'
```
lucywyman: This is just for messing with run_task and isolating it a bit. I expect this will live in the executor

exe/bolt-server
```diff
@@ -0,0 +1,36 @@
...
+
+command = ARGV.shift
+subcommands[command].order!
+
+# TODO can this logic be moved to the orch client?
+if command == 'start'
+  puts "Starting bolt server..."
+  server = File.join(File.dirname(__FILE__), "..", "lib", "bolt_ext", "server.rb")
+  pid = Process.spawn("ruby #{server}")
+  Process.detach(pid)
```
adreyer: It may be easier to debug if you don't detach

adreyer:
- we should not detach from the process or track a pid. That limits how bolt-server can be managed.
- The current executor signature is run_task(targets, task, arguments, options) It seems like we can keep that and have a different implementation of the Task object that is easier to set up in the server context
- pal.run_task -> puppet function run_task -> executor.run_task what are you considering changing?

lucywyman:
* M'k
* M'k
* The change would be to have `cli.rb` call executor.run_task directly, but I didn't follow the whole path to see that it was invoked. That's fine as is, I was mostly worried about the implementations diverging / code duplication.

adreyer:
We may want a version of run_task that accepts only one target and doesn't run in a different thread I guess.

adreyer:
CLI.rb needs to go through PAL to know what the task is. The Puppet function is needed for calls from plans too. Since the server isn't doing any of the task spec stuff it seems like duplication is minimal

adreyer:
I think there is a big question about where this code should live and how it should be packaged? How can we avoid the bloat of bundled puppet in the transport service gem.

MikaelSmith:
Still some rubocop issues.

exe/bolt-server
```diff
@@ -0,0 +1,19 @@
...
+  define("--port", "specify a port") do |port|
+    options[:port] = port
+  end
+  define("--cert", "pass a certificate") do |cert|
+    options[:cert] = cert
+  end
+end.parse!
+
+server = File.join(File.dirname(__FILE__), "..", "lib", "bolt_ext", "server.rb")
+system("ruby #{server}")
```
MikaelSmith: Is this done to essentially remove `optparse` from the Ruby interpreter? We'd already required `bolt_ext/server.rb`, so it seems funny to fork for it.

MikaelSmith: Also don't see where the `options` are used at all.

lucywyman: They aren't yet -- I spent a good chunk yesterday trying to get them working but having sinatra take command line arguments is not hard but non-trivial (which I think is what we want, unless we want to write a config file each time before starting the server?)

lucywyman: I can just remove the optparse chunk.

adreyer: I think this file is just part of the POC. we should probably ticket moving this to something more normal for this. For example using json data instead of optparse.

lib/bolt/task.rb
```diff
@@ -0,0 +1,26 @@
...
+    end
+
+    def initialize(task_name, files, metadata = nil)
+      unless Bolt::Task.task_name?(task_name)
+        raise Bolt::Error, _("Task names must start with a lowercase letter " \
+                             "and be composed of only lowercase letters, numbers, and underscores")
+      end
+
+      @name = task_name
+      @metadata = metadata if metadata
```
adreyer: should we validate this if we're validating the name?

lib/bolt_ext/server.rb
```diff
@@ -0,0 +1,27 @@
...
+
+set :server, :puma
+
+post '/ssh/run_task' do
+  body = JSON.parse(request.body.read)
+  target = [Bolt::Target.new(body['target']['hostname'], body['target']['options'])]
+  task = Bolt::Task.new(body['task']['name'], body['task']['file'], body['task']['metadata'])
+  parameters = body['parameters']
+
+  @analytics = Bolt::Analytics.build_client
```
adreyer: Should we be using the analytics client? It seems like we should let something upstream handle analytics using the analytics service.

lib/bolt_ext/server.rb
```diff
@@ -0,0 +1,27 @@
...
+  executor = Bolt::OrchExecutor.new(Bolt::Config.new, @analytics, false)
+
+  @r = ''
```
adreyer: We probably shouldn't be setting this to empty. It seems like it can only serve to mask bugs.

We do probably need some error handling.

lib/bolt_ext/server.rb
```diff
@@ -0,0 +1,31 @@
...
+  post '/ssh/run_task' do
+    body = JSON.parse(request.body.read)
+    target = [Bolt::Target.new(body['target']['hostname'], body['target']['options'])]
+    task = Bolt::Task.new(body['task']['name'], body['task']['file'], body['task']['metadata'])
+    parameters = body['parameters']
+
+    @analytics = Bolt::Analytics.build_client
+    executor = Bolt::OrchExecutor.new(Bolt::Config.new, @analytics, false)
+
+    @r = ''
```
adreyer: I don't think we should be setting an empty body. This seems like it will only serve to mask errors.(side note are instance variables on the Base per request?)

We probably need to add some explicit error handling instead.

temp/server-client.rb
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/env ruby
```
MikaelSmith: `temp` is a weird place to put this. If it's something we're going to use, I've been leaning towards a `libexec` directory.

lib/bolt_ext/server.rb
```diff
@@ -0,0 +1,31 @@
...
+      if event[:type] == :node_result
+        @r = event[:result].to_json
+      end
+    end
+    response.body = @r
+    response.status = 200
+  end
+end
+
+TransportAPI.run!
```
MikaelSmith: Maybe this should happen in `exe/bolt_server`? Seems weird to start a server when you require a file.

temp/server-client.rb
lucywyman: Makes sense -- it's just an example client for proving the concept, I think if it lives anywhere it'll be in docs

lib/bolt_ext/server.rb
lucywyman: Sure, and it could be an option passed on startup? Do we want to collect bolt analytics for these jobs? If so, do we need to pass the analytics client to the executor to get analytics for the jobs, or can the client handle analytics on it's own?

lib/bolt_ext/server.rb
lucywyman: Created https://tickets.puppetlabs.com/browse/BOLT-657

lib/bolt/task.rb
lucywyman: Created https://tickets.puppetlabs.com/browse/BOLT-658 -- I couldn't find a function that exists  to do this already, and we should probably validate the rest of the data we recieve as well.

lib/bolt_ext/server.rb
```diff
@@ -0,0 +1,29 @@
...
+class TransportAPI < Sinatra::Base
+  set :server, :puma
+
+  post '/ssh/run_task' do
+    body = JSON.parse(request.body.read)
+    target = [Bolt::Target.new(body['target']['hostname'], body['target']['options'])]
+    task = Bolt::Task.new(body['task']['name'], body['task']['file'], body['task']['metadata'])
+    parameters = body['parameters']
+
+    @analytics = Bolt::Analytics.build_client
```
adreyer: We should remove this line and just use the noop client

lib/bolt.rb
MikaelSmith: Is work to refactor the executor ticketed somewhere?

lib/bolt/task.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+require 'json'
+require 'bolt/error'
+
+module Bolt
+  class Task
```
MikaelSmith: I used a struct at https://github.com/puppetlabs/bolt/blob/master/lib/bolt/applicator.rb#L7 for a similar purpose (though skipped validation because I controlled the task). It'd be nice to unify these eventually.

lucywyman: I think for now it makes more sense to use a struct. I based this off of `Puppet::Module::Task` but since we're not loading the task from a module I don't think we'll need any of that functionality.

lib/bolt/transport/ssh.rb
```diff
@@ -115,8 +115,14 @@ def run_script(target, script, arguments, options = {})
       end
 
       def run_task(target, task, arguments, options = {})
-        executable = target.select_impl(task, PROVIDED_FEATURES)
-        raise "No suitable implementation of #{task.name} for #{target.name}" unless executable
+        if task.input_method == 'api'
```
lucywyman: Is there a better way to use the file passed on the API than this?

MikaelSmith: According to https://github.com/net-ssh/net-scp it can upload from an in-memory buffer with `scp.upload! StringIO.new("some data to upload"), "/remote/path"`.

MikaelSmith: A temporary file seems like the only short-term solution for WinRM though. We'd have to modify winrm-fs otherwise.

MikaelSmith: This isn't what `task.input_method` is normally used for. Right now it precludes passing the input_method for the task (part of its metadata) via the API.

temp/server-client.rb
MikaelSmith: I'd rather it in something like `examples/` then.

lib/bolt/transport/ssh.rb
MikaelSmith: Probably easiest to put the file content in something called `file_content`, then key off its existence (i.e. `task.respond_to? :file_content`.

lib/bolt_ext/server.rb
```diff
@@ -0,0 +1,37 @@
...
+class TransportAPI < Sinatra::Base
+  set :server, :puma
+
+  post '/ssh/run_task' do
+    body = JSON.parse(request.body.read)
+    keys = %w[user password port ssh-key-content]
+    opts = body['target'].select { |k, _| keys.include? k }
+    opts.merge(body['target']['options'])
+    target = [Bolt::Target.new(body['target']['hostname'], opts)]
+    Task = Struct.new(:name,
```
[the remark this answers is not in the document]
lucywyman: Should we have it defined somewhere more general? Since now we're using it in 2 places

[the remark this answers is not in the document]
MikaelSmith: Sure. That'd be easy enough to add as `lib/bolt/task`. Would look similar to config.rb.

lib/bolt/executor.rb
```diff
@@ -8,6 +8,7 @@
 require 'set'
 require 'bolt/analytics'
 require 'bolt/result'
+require 'bolt/task'
```
MikaelSmith: The executor doesn't use this, I'm tempted to leave it out of the requires.

lucywyman: And include 'bolt/task' directly in files that use it?

lib/bolt/transport/ssh.rb
```diff
@@ -138,7 +140,12 @@ def run_task(target, task, arguments, options = {})
             end
 
             conn.with_remote_tempdir do |dir|
-              remote_task_path = conn.write_remote_executable(dir, executable)
+              if task.respond_to? :file_content
```
MikaelSmith: This probably breaks `apply`. Apply would pass in an object that responds to `file_content`, but `file_content` returns nil. So you should also check that `file_content` is not nil.

MikaelSmith: Eventually it'd probably make sense to wrap the Puppet anon classes we get for tasks in our own task object so we can have a consistent API.

lib/bolt/transport/ssh.rb
```diff
@@ -115,8 +115,12 @@ def run_script(target, script, arguments, options = {})
       end
 
       def run_task(target, task, arguments, options = {})
-        executable = target.select_impl(task, PROVIDED_FEATURES)
-        raise "No suitable implementation of #{task.name} for #{target.name}" unless executable
+        unless task.respond_to? :file_content
```
MikaelSmith: This section is kind of hard to reason about. I'd also like to unify these checks and the section below.

I'd lean towards
```
if task.respond_to?(:file_content) and task.file_content
  filename = task.name.split("::").last
  write_executable = lambda { |conn, dir| conn.write_executable_from_content(dir, task.file_content, filename) }
else
  executable = target.select_impl(task, PROVIDED_FEATURES)
  raise "No suitable implementation of #{task.name} for #{target.name}" unless executable
  write_executable = lambda { |conn, dir| conn.write_remote_executable(dir, executable) }
end
...
remote_task_path = write_executable(conn, dir)
```

lib/bolt/transport/ssh/connection.rb
```diff
@@ -309,6 +309,13 @@ def write_remote_executable(dir, file, filename = nil)
           remote_path
         end
 
+        def write_executable_from_content(dest, content, filename = nil)
```
MikaelSmith: filename is required.  Letting it be `nil` when that's going to make us try to upload content to a directory name seems bad.

lib/bolt/transport/ssh/connection.rb
```diff
@@ -309,6 +309,13 @@ def write_remote_executable(dir, file, filename = nil)
...
+          remote_path = "#{dest}/#{filename}"
```
MikaelSmith: `File.join(dest, filename)` is probably more portable.

temp/encoded-file
```diff
@@ -0,0 +1,4 @@
...
+IyEvb3B0L3B1cHBldGxhYnMvcHVwcGV0L2Jpbi9ydWJ5CgpyZXF1aXJlICdqc29uJwoKYXJncyA9
```
MikaelSmith: Lines 2-4 appear to be junk when I try to decode them.

lib/bolt/transport/ssh/connection.rb
lucywyman: Yeah -- opted for consistency over portability, but I like File.join better and don't think it matters much.

temp/encoded-file
lucywyman: I think I may have appended instead of rewritten....

lib/bolt/transport/ssh/connection.rb
MikaelSmith: Oh, didn't notice above. I'd probably change both.

lib/bolt/transport/ssh.rb
```diff
@@ -138,7 +144,15 @@ def run_task(target, task, arguments, options = {})
             end
 
             conn.with_remote_tempdir do |dir|
-              remote_task_path = conn.write_remote_executable(dir, executable)
+              if from_api?(task)
+                filename = task.name.split("::").last
+                remote_task_path = conn.write_executable_from_content(dir, task.file_content, filename)
+              else
+                executable = target.select_impl(task, PROVIDED_FEATURES)
+                raise "No suitable implementation of #{task.name} for #{target.name}" unless executable
```
MikaelSmith: Before we had some notice earlier that the executable couldn't be identified. Although that doesn't look like it makes much difference in practice (it'll now happen after we try to establish a connection).

PR #487, opened 2018-06-22
(BOLT-583) Do not default to localhost on windows

donoghuc:
When bolt is invoked from a windows OS with the target of localhost, do not set the default transport to 'local'


pre-docs/bolt_options.md
```diff
@@ -97,6 +97,8 @@ transport for each node. To override the default transport, specify the
 protocol on a per-host basis, such as `bolt command run facter --nodes
 win1,ssh://linux --transport winrm`
 
+If ```localhost``` is passed to ```--nodes``` when invoking Bolt on a non-windows OS the `local` transport is used automatically. This behaviour can be avoided by prepending the target with the desired transport (ex: ```ssh://localhost```).
```
MikaelSmith: Most of these examples could use single backticks rather than triple. I'm surprised it formats well.

PR #488, opened 2018-06-25
(BOLT-567) Add command to generate catalogs

adreyer:
This adds a POC new command to generate catalogs and JSON AST from
manifests. The code is kept separate from most of bolt since it's a
different process and we should expect it to be completely rewritten.

### DIFF
--- a/exe/bolt_catalog
+++ b/exe/bolt_catalog
@@ -1,0 +1,34 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+
+require 'bolt'
+require 'bolt/catalog'
+require 'json'
+
+# This accepts a catalog request on stdin:
+# { "code_ast": "JSON serialized Puppet AST",
+#   "code_string": "String of code, ignored if AST is provided,
+#   "modulepath": "Array of directories to use as the modulepath for catalog compilation.
+#   "target": {
+#   "name": "the name of the node usually fqdn fro url",
+#   "facts": "Hash of facts to use for the node",
+#   "variables": "Hash of variables to use for compilation"
+#   }
+# }
+
+command = ARGV[0]
+if command == "parse"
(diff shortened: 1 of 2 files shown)


exe/bolt_catalog
```diff
@@ -0,0 +1,33 @@
...
+
+command = ARGV[0]
+if command == "parse"
+  code = File.open(ARGV[1], &:read)
+  puts Bolt::Catalog.new.generate_ast(code)
+elsif command == "compile"
+  request = if ARGV[1]
+              File.open(ARGV[1]) { |fh| JSON.parse(fh) }
+            else
+              JSON.parse(STDIN)
```
MikaelSmith: This didn't work, maybe `STDIN.read`.

adreyer: annoying it was load but rubocop didn't like it

MikaelSmith: Seems rubocop wants you to be explicit about reading from a stream, so `JSON.parse(STDIN.read)` rather than `JSON.load(STDIN)`: https://github.com/rubocop-hq/rubocop/blob/master/manual/cops_security.md#securityjsonload. I'm having trouble identifying why it's a real security concern though.

exe/bolt_catalog
```diff
@@ -0,0 +1,33 @@
...
+#   "target": {
+#   "name": "the name of the node usually fqdn fro url",
+#   "facts": "Hash of facts to use for the node",
+#   "variables": "Hash of variables to use for compilation"
+#   }
+
+command = ARGV[0]
+if command == "parse"
+  code = File.open(ARGV[1], &:read)
+  puts Bolt::Catalog.new.generate_ast(code)
```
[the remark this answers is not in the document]
MikaelSmith: It seems to work as a string to round-trip back in for the compile step.

PR #491, opened 2018-06-26
(BOLT-594) Experimental 'apply' keyword

MikaelSmith:
Create an apply_executor that Puppet calls back to when evaluating an
apply expression. Compile the apply AST into a catalog.

Depends on BOLT-567.


lib/bolt/executor.rb
```diff
@@ -12,17 +12,19 @@
...
 require 'bolt/puppetdb'
+require 'open3'
 
 module Bolt
   class Executor
     attr_reader :noop, :transports
     attr_accessor :run_as, :plan_logging
 
-    def initialize(config = Bolt::Config.new, analytics = Bolt::Analytics::NoopClient.new, noop = nil)
+    def initialize(config = Bolt::Config.new, analytics = Bolt::Analytics::NoopClient.new, noop = nil, inventory)
```
MikaelSmith: This isn't ideal. It starts to feel weird to have both inventory and executor in Puppet.lookup and also need to have a new class that wraps both that we use for an apply executor.

lib/bolt/applicator.rb
```diff
@@ -0,0 +1,53 @@
...
+  Task = Struct.new(:name, :implementations, :input_method)
+
+  class Applicator
+    def initialize(inventory, executor)
+      @inventory = inventory
+      @executor = executor
+    end
+
+    def apply(args, apply_body, _scope)
+      raise(ArgumentError, 'apply requires a TargetSpec') if args.empty?
```
MikaelSmith: This will probably never be hit, the grammar requires at least one argument.

I'd like to narrow this down to two arguments sometime soon. Right now it's open-ended.

.gitmodules
```diff
@@ -1,7 +1,7 @@
 [submodule "vendored/puppet"]
 	path = vendored/puppet
 	url = https://github.com/puppetlabs/puppet
-	branch = master
+	branch = apply-grammar
```
MikaelSmith: @nicklewis pointed out that this grammar currently allows resource declarations in the arguments that would not normally be allowed in a plan. Need to fix that.

MikaelSmith: Fixed.

lib/bolt/applicator.rb
adreyer: Do we want to do that in the grammar on bolt?
I can see things like no arguments implicitly runs locally or adding a description argument to align with the other functions.

MikaelSmith: Ok, that could be worth allowing zero arguments, I'll need to modify the grammer.

I'm ok adding more, but I think it would still be a friendly experience if we updated the grammar to reflect the number of options we expect to support.

lib/bolt/applicator/apply_catalog.rb
```diff
@@ -0,0 +1,61 @@
+#! /opt/puppetlabs/puppet/bin/ruby
```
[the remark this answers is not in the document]
MikaelSmith: Have a suggestion where? I was trying to decide where it would be innocuous but still packaged with the gem?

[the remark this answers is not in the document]
MikaelSmith: Maybe `libexec` would make sense for it and bolt_catalog. Both things we don't think people should use directloy.

PR #493, opened 2018-06-26
(BOLT-568) Add task to apply a catalog

nicklewis:
This task accepts a catalog in JSON format, then loads Puppet and uses
it to apply the catalog. The result is a JSON representation of the
report generated by the catalog application. This skips pluginsync and
doesn't submit a report to the master.

### DIFF
--- a/bolt-modules/boltlib/tasks/apply_catalog.rb
+++ b/bolt-modules/boltlib/tasks/apply_catalog.rb
@@ -1,0 +1,56 @@
+#!/opt/puppetlabs/puppet/bin/ruby
+# frozen_string_literal: true
+
+require 'json'
+require 'puppet'
+require 'puppet/configurer'
+
+Puppet.initialize_settings([])
+run_mode = Puppet::Util::RunMode[:user]
+Puppet.settings.initialize_app_defaults(Puppet::Settings.app_defaults_for_run_mode(run_mode))
+
+Puppet::ApplicationSupport.push_application_context(run_mode, :local)
+
+# Avoid extraneous output
+Puppet[:summarize] = false
+
+# Don't interfere with the normal agent
+Puppet[:report] = false
+Puppet[:statefile] = Tempfile.new('statefile')
+Puppet[:graph] = false
+
+# Make sure to apply the catalog
+Puppet[:use_cached_catalog] = false
+Puppet[:noop] = false
+Puppet[:strict_environment_mode] = false
+
+# The whole catalog
+Puppet[:tags] = nil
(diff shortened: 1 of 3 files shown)


bolt-modules/boltlib/tasks/apply_catalog.rb
```diff
@@ -0,0 +1,57 @@
+#!/opt/puppetlabs/puppet/bin/ruby
+# frozen_string_literal: true
+
```
adreyer: Do we want to put this somewhere else so it doesn't show up as a public task?

nicklewis: That seems probably good, but I wasn't sure how best to accomplish it. If the intent is for boltlib to be "internal", we could redact it from the `bolt task show` output.

Or we could have this task live somewhere else and only be added to the modulepath temporarily while we're executing an `apply` block. That seems a bit more heavyweight.

MikaelSmith: Redacting from `bolt task show` makes sense to me.

adreyer: It seems like we could bypass the run_task puppet function and searching the modulepath and just create a "task object" to hand to the executor directly?

MikaelSmith: I'll give that a try.

nicklewis:
Updated to no longer delete `/dev/null`.

MikaelSmith:
I've rolled `apply_catalog.rb` into https://github.com/puppetlabs/bolt/pull/491. If you want to keep this commit I can merge it, or close and carry on in the other PR.

bolt-modules/boltlib/tasks/apply_catalog.rb
```diff
@@ -0,0 +1,56 @@
...
+Puppet.settings.initialize_app_defaults(Puppet::Settings.app_defaults_for_run_mode(run_mode))
+
+Puppet::ApplicationSupport.push_application_context(run_mode, :local)
+
+# Avoid extraneous output
+Puppet[:summarize] = false
+
+# Don't interfere with the normal agent
+Puppet[:report] = false
+Puppet[:statefile] = Tempfile.new('statefile')
```
MikaelSmith: I don't think this worked for me, I had to create a Tempfile, then call `path` on it.

MikaelSmith:
Subsumed by https://github.com/puppetlabs/bolt/pull/491.

PR #497, opened 2018-06-29
(BOLT-630) Only pass expected params for powershell input to function

MikaelSmith:
When using powershell input method, only pass expected parameters to the
function. When it receives unexpected parameters the function call will
fail, and we don't want that to happen due to metaparameters that aren't
used.

This restriction requires that functions explicitly mention all
parameters they wish to use, rather than using a collector array for any
remaining parameters.

### DIFF
--- a/lib/bolt/transport/winrm.rb
+++ b/lib/bolt/transport/winrm.rb
@@ -128,9 +128,12 @@ catch
                 # fortunately, using PS with stdin input_method should never happen
                 if input_method == 'powershell'
                   conn.execute(<<-PS)
-$private:taskArgs = Get-ContentAsJson (
+$private:tempArgs = Get-ContentAsJson (
   $utf8.GetString([System.Convert]::FromBase64String('#{Base64.encode64(JSON.dump(arguments))}'))
 )
+$allowedArgs = (Get-Command "#{remote_path}").Parameters.Keys
+$private:taskArgs = @{}
+$private:tempArgs.Keys | ? { $allowedArgs -contains $_ } | % { $private:taskArgs[$_] = $private:tempArgs[$_] }
 try { & "#{remote_path}" @taskArgs } catch { Write-Error $_.Exception; exit 1 }
               PS
                 else
--- a/spec/fixtures/modules/task_param/tasks/win.ps1
+++ b/spec/fixtures/modules/task_param/tasks/win.ps1
@@ -1,0 +1,8 @@
+[CmdletBinding()]
+param(
+  [Parameter(Mandatory = $False)]
+  [String]
+  $_task
+)
+
+Write-Output "Running task $_task"
--- a/spec/bolt/transport/winrm_spec.rb
+++ b/spec/bolt/transport/winrm_spec.rb
@@ -401,22 +401,18 @@ Height: $Height ($(if ($Height -ne $null) { $Height.GetType().Name } else { 'nul
       end
     end
 
-    it "fails when the powershell input method is used to pass unexpected parameters to a task", winrm: true do
+    it "ignores unexpected parameters when the powershell input method is used", winrm: true do
       contents = <<-PS
         param (
           [Parameter()]
           [String]$foo
         )
 
-        $bar = $env:PT_bar
-        Write-Host `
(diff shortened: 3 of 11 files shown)


MikaelSmith:
@Iristyle have any particular tests you think we should add?

nicklewis:
This makes sense to me.

lib/bolt/transport/winrm.rb
```diff
@@ -128,9 +128,12 @@ def run_task(target, task, arguments, _options = {})
                 # fortunately, using PS with stdin input_method should never happen
                 if input_method == 'powershell'
                   conn.execute(<<-PS)
-$private:taskArgs = Get-ContentAsJson (
+$private:tempArgs = Get-ContentAsJson (
   $utf8.GetString([System.Convert]::FromBase64String('#{Base64.encode64(JSON.dump(arguments))}'))
 )
+$allowedArgs = (Get-Command "#{remote_path}").Parameters.Keys
+$private:taskArgs = @{}
+$private:tempArgs.Keys | ? { $_ -in $allowedArgs } | % { $private:taskArgs[$_] = $private:tempArgs[$_] }
```
Iristyle: Just double checked the `-in` operator. It's PS3 only, so we might want to change `$_ -in $allowedArgs` to `$allowedArgs -contains $_`

MikaelSmith: Makes sense.

spec/fixtures/modules/task_param/tasks/win.ps1
```diff
@@ -0,0 +1,8 @@
+[CmdletBinding()]
+param(
+  [Parameter(Mandatory = $False)]
+  [String]
+  $_task
```
Iristyle: One question I have in general about task metadata in the context of PowerShell.... would it make sense to pass in a single object containing key-value pairs, instead any one of x arbitrary parameters?

That would make it a lot easier to scaffold task templates from the PDK perspective, and reduces the need for future modifications to script prototypes, once a script accepts the metadata object.

MikaelSmith: How do you envision people using them? A metadata hash, used like `meta['_task']`? We can talk about that a bit, but I'd like to get this fix out this week.

Iristyle:
Yes, something like that. I think it makes things cleaner for implementers.

But it also has the cons of introducing another difference between
Powershell and other input methods and a slightly more magical / less
self-documenting PowerShell script. I think those trade-offs are worth it
though.

spec/fixtures/modules/task_param/tasks/win.ps1
Iristyle: Did you file a ticket for this at all @MikaelSmith ?

MikaelSmith: I'll create a ticket, but without further input I'm not really in favor of this idea (and thus it probably won't go anywhere) for a couple reasons:
- it would require a change to the task specification, that affects how tasks use `_noop` in existing PE releases
- we'd want to name it something like `_meta` anyway to avoid collisions with valid user parameter names
- I'm not convinced it's that much of an improvement when they still need to go lookup what the metaparameter syntax is to be able to scaffold it, which means it doesn't feel worth the effort

Iristyle: * I'm not sure how that would impact `_noop` if there are no PowerShell consumers of this concept yet?
* Only one thing to worry about colliding is better than more than one (agree that the risk of collision is probably low given I'm not aware of anyone who uses params in PowerShell that start with `_`)
* Never having to update PDK templates when new params are added is a long-term maintenance win (especially if the template comments a line like `Write-Debug $_meta | Select *` which will give you all the props)
* I think a single self-descriptive object is much better UX than receiving any number of arbitrary params (only passed when a script explicitly asks for them) - I'd wager that has the benefit of providing better flexibility for task authors running against multiple Bolt / task versions as well
* To expand on above, if a script is designed to run local, but can also receive task metadata, it's cleaner to have a single arg exposed (for things like tab completion)
* Generally speaking, introducing parameter objects is a good refactor

PR #498, opened 2018-07-02
(BOLT-578) Report bundled content to GA

donoghuc:
Include the name of the task/plan being run if that name refers to a piece of built-in content. The task/plan name should not be included if it isn't a piece of bundled content, as the name may be sensitive.

### DIFF
--- a/lib/bolt/cli.rb
+++ b/lib/bolt/cli.rb
@@ -262,7 +262,7 @@ module Bolt
                          params: params }
         plan_context[:description] = options[:description] if options[:description]
 
-        executor = Bolt::Executor.new(config, @analytics, options[:noop])
+        executor = Bolt::Executor.new(config, @analytics, options[:noop], bundled_content: bundled_content)
         executor.start_plan(plan_context)
         result = pal.run_plan(options[:object], options[:task_options], executor, inventory, puppetdb_client)
 
@@ -271,7 +271,7 @@ module Bolt
         outputter.print_plan_result(result)
         code = result.ok? ? 0 : 1
       else
-        executor = Bolt::Executor.new(config, @analytics, options[:noop])
+        executor = Bolt::Executor.new(config, @analytics, options[:noop], bundled_content: bundled_content)
         targets = options[:targets]
 
         results = nil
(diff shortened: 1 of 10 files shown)


donoghuc:
Plan to move logic such that tasks/plans not directly called via CLI will be captured (possibly in run_task?).

lib/bolt/executor.rb
```diff
@@ -127,6 +146,13 @@ def report_function_call(function)
       @analytics&.event('Plan', 'call_function', function)
     end
 
+    def report_bundled_content(mode, name)
+      if BUNDLED_CONTENT.include?(name)
+        screen = "bundled_#{mode}_content"
+        @analytics&.screen_view(screen, bundled_content: name)
```
MikaelSmith: This makes more sense as an event, similar to calling a function or using a transport. Could be something like `event('Bundled Content', mode, name)`.

MikaelSmith:
It looks like somehow when you updated you pulled in all the changes from https://github.com/puppetlabs/bolt/commit/f16c05f68d6cc5bcb81eedd0db2fb705fb395c0c.

MikaelSmith:
A `git reset --hard upstream/master && git cherry-pick 44ace3194511679343bf22c7a523d75f04b58e50` would probably fix that.

donoghuc:
Yes that was a mistake. I will clean that up.

lib/bolt/cli.rb
```diff
@@ -353,5 +357,16 @@ def file_stat(path)
     def outputter
       @outputter ||= Bolt::Outputter.for_format(config[:format], config[:color], config[:trace])
     end
+
+    def load_bundled_content
+      default_content = Bolt::PAL.new(Bolt::Config.new)
```
MikaelSmith: It looks like this should avoid picking up the default modulepath and just use the vendored modules. It also doesn't appear to have any adverse impact on performance. So this approach seems ok.

I'd rename this to `bundled_content` and just have it called directly when constructing the Executor. No need to have separate checks beforehand.

donoghuc: Oh right, task/plan are mutually exclusive on CLI. Will clean that up.

PR #500, opened 2018-07-03
(BOLT-633) Set modulepath, enable puppetdb functions during apply

MikaelSmith:
Sets modulepath to Bolt's modulepath. Enables Bolt's puppetdb functions
when compiling a manifest block. Other functions continue to be
disabled.


MikaelSmith:
Added tests.

PR #502, opened 2018-07-06
(maint) Fix acceptance helpers so Rototiller and beaker subcommands work

MikaelSmith:
Rototiller expects defaults to be declared in the Rakefile, while beaker
subcommands don't use the Rakefile. The end of the acceptance README
provides instructions for running with beaker subcommands that didn't
work unless you explicitly set all the optional environment variables
you needed. Refactor so that both paths share defaults.

### DIFF
--- a/acceptance/README.md
+++ b/acceptance/README.md
@@ -288,11 +288,15 @@ Example of running test suite using gem install using an existing hosts file:
     ```
     bundle exec beaker provision
     ```
-1. Run tests with specified environment variables
+1. Run pre-suite and tests with specified environment variables
     ```
     SSH_USER=root           \
     SSH_PASSWORD='S3@ret3'       \
     WINRM_USER=Administator      \
     WINRM_PASSWORD='S3@ret3'      \
     bundle exec beaker exec -t ./tests
+    ```
+1. Re-run tests
+    ```
+    bundle exec beaker exec ./tests
     ```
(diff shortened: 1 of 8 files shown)


donoghuc:
This is really helpful! Allows easy config for installing from a github source on configurable set of targets. If you choose to install from a github branch and persist the targets is it possible to re-run the tests without installing bolt every time? 
```
# select targets (add keyfile: path in hosts.yaml if using vmpooler)
bundle exec beaker-hostgenerator ubuntu1604-64bolt,ssh.-centos7-64ssh. > hosts.yaml
``` 
```
# init with a config (in this case git)
bundle exec beaker init -h hosts.yaml -o config/git/options.rb
```
```
# get nodes
bundle exec beaker provision
```
```
# install from source and preserve hosts for re-running tests
GIT_FORK=MikaelSmith/bolt \
GIT_BRANCH=fix-acceptance-helpers \
BEAKER_PRESERVE_HOSTS=always \
bundle exec beaker exec -t ./tests
```
Is it possible to re-run tests without trying to re-install bolt? 
```
# if you try to re-run tests, error because "bolt" directory already installed
bundle exec beaker exec -t ./tests --preserve-hosts=always --no-provision
```

MikaelSmith:
You shouldn't need to set BEAKER_PRESERVE_HOSTS. It looks like using the `-t` option makes it re-run the pre-suite, you can drop that option.

PR #503, opened 2018-07-09
(BOLT-650) Add bolt as a service

lucywyman:
This adds the necessary infra to deploy the pe-bolt-server as a service, including service init files, defaults, and the unit file. We also need to check if the `$HOME` environment variable exists and set paths more explicitly if it doesn't, since when the service starts that env var won't be avialable.

To test this:
* Copy exe/bolt-server to /opt/puppetlabs/puppet/bin
* Copy ext/systemd/pe-bolt-server.service file to /lib/systemd/system/ 
    * Create symlink to /etc/systemd/system/multi-user.target.wants/
* Copy ext/systemd/pe-bolt-server.logrotate to /etc/logrotate.d/pe-bolt-server
* Copy ext/debian/pe-bolt-server.default file to /etc/default/pe-bolt-server
* Copy ext/debian/pe-bolt-server.init file to /etc/init.d/pe-bolt-server
* `systemctl daemon-reload`
* `service pe-bolt-server start`

```
sudo cp exe/bolt-server /opt/puppetlabs/puppet/bin
sudo cp ext/systemd/pe-bolt-server.service /lib/systemd/system/pe-bolt-server
sudo cp ext/debian/pe-bolt-server.default /etc/default/pe-bolt-server
sudo cp ext/debian/pe-bolt-server.init /etc/init.d/pe-bolt-server
systemctl daemon-reload
service pe-bolt-server start
```

### DIFF
--- a/ext/debian/pe-bolt-server.default
+++ b/ext/debian/pe-bolt-server.default
@@ -1,0 +1,4 @@
+# Defaults for pe-bolt-server - sourced by /etc/init.d/pe-bolt-server
+
+# Startup options
+#PE_BOLT_SERVER_OPTIONS=--loglevel=debug
--- a/ext/debian/pe-bolt-server.init
+++ b/ext/debian/pe-bolt-server.init
@@ -1,0 +1,114 @@
+#! /bin/sh
+### BEGIN INIT INFO
+# Provides:          pe-bolt-server
+# Required-Start:    $network $named $remote_fs $syslog
+# Required-Stop:     $network $named $remote_fs $syslog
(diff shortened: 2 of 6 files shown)


lucywyman:
I can't replicate the travis error, `unable to resolve type 'uintptr_t'` -- seems like it's an ffi issue?

MikaelSmith:
A addressed that with https://github.com/puppetlabs/bolt/commit/25239f15b20b1887ea15b71dbf5c6e3f8acbc388, although it appears to be inconsistent.

PR #504, opened 2018-07-09
(BOLT-672, BOLT-673) Add plan variables and trusted data

MikaelSmith:
Include plan variables and trusted data when compiling a manifest block.

### DIFF
--- a/lib/bolt/applicator.rb
+++ b/lib/bolt/applicator.rb
@@ -23,13 +23,16 @@ module Bolt
     end
 
     def compile(target, ast, plan_vars)
+      trusted = Puppet::Context::TrustedInformation.new('local', target.host, {})
+
       catalog_input = {
         code_ast: ast,
         modulepath: [],
         target: {
           name: target.host,
           facts: @inventory.facts(target),
-          variables: @inventory.vars(target).merge(plan_vars)
+          variables: @inventory.vars(target).merge(plan_vars),
+          trusted: trusted.to_h
         }
       }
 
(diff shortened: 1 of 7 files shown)


lib/bolt/applicator.rb
```diff
@@ -24,16 +24,28 @@ def apply(args, apply_body, _scope)
         params = args[1]
       end
 
+      # collect plan vars and merge them over target vars
+      plan_vars = scope.to_hash
+      %w[trusted server_facts facts].each { |k| plan_vars.delete(k) }
```
lucywyman: Not sure I follow how they get merged into target vars?

lucywyman: Ugh, because I didn't read like 3 more lines. nvm.

lucywyman:
This looks fine to me (beyond the merge conflict) but I'm not sure how to test it. I'll give writing a plan that uses it a shot, but if you have one on hand you've been using let me know!

MikaelSmith:
Updated.

lib/bolt/applicator.rb
```diff
@@ -22,14 +22,22 @@ def catalog_apply_task
       Task.new('apply_catalog', [impl], 'stdin')
     end
 
-    def compile(target, ast)
+    def compile(target, ast, plan_vars)
+      trusted = {
+        'certname' => target.host,
+        'hostname' => target.host,
+        'domain' => target.host,
```
nicklewis: Using host as domain seems wrong... Should we try to cut it out of the hostname or maybe omit it altogether?

MikaelSmith: I think I'll try using https://github.com/puppetlabs/puppet/blob/master/lib/puppet/context/trusted_information.rb to generate this.

PR #505, opened 2018-07-09
(BOLT-634) Fix modulepath on Windows

MikaelSmith:
Windows CLI uses a different modulepath, and needs a non-Cygwin
directory to install the facts module.

Skip the apply winrm test if the controller node is a winrm target and
has its own Ruby installed, as Bolt will fail to run apply in Puppet
Agent's Ruby. That will be fixed separately.

### DIFF
--- a/acceptance/lib/acceptance/bolt_command_helper.rb
+++ b/acceptance/lib/acceptance/bolt_command_helper.rb
@@ -13,17 +13,36 @@ module Acceptance
     # @option flags [nil] '--no-ssl' specify nil to use
     # @param [Hash] opts the options hash for this method
     def bolt_command_on(host, command, flags = {}, opts = {})
-      platform = host['platform']
       bolt_command = command.dup
       flags.each { |k, v| bolt_command << " #{k} #{v}" }
 
-      if platform =~ /windows/
+      case host['platform']
+      when /windows/
         execute_powershell_script_on(host, bolt_command, opts)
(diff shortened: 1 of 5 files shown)


MikaelSmith:
Test still fails with packages.

donoghuc:
Do we want the cygwin path or the "windows" path?
```
[1] pry(#<Beaker::TestCase>)> modulepath(File.join(dir, 'modules'))
=> "\"C:/Users/Administrator/.puppetlabs/bolt/modules;C:/cygwin64/tmp/apply_winrm.QsQGsc/modules\""
[2] pry(#<Beaker::TestCase>)> bolt['platform']
=> "windows-10ent-64"
[3] pry(#<Beaker::TestCase>)> on(bolt, 'cygpath -m $(printenv USERPROFILE)').stdout.chomp
=> "C:/Users/Administrator"
```
`cygpath -m` seems to exclude the cygwin part. 
`cygpath -u` would include it.
```
-bash-4.3$ cygpath -u $(printenv USERPROFILE)
/cygdrive/c/Users/Administrator
```

MikaelSmith:
Want to exclude cygwin path, cygwin won't require it and Bolt won't recognize it.

MikaelSmith:
@donoghuc the last commit fixes running from packages for me.

MikaelSmith:
YOLO

PR #506, opened 2018-07-10
add simple bash completion

scoopex:


resources/bolt_bash_completion.sh
```diff
@@ -0,0 +1,62 @@
+#####################################################################
+###
+### Bash autocompletion for bolt
+### include this file in your .bash_profile or .bashrc to use completion
+### "source bolt_bash_completion.sh"
+<<<<<<< HEAD
```
lucywyman: It looks like you've still got some git diffs here.

scoopex: i'm sorry :-) Fixed it :-)

lucywyman: You might need to make sure you push to your branch - I'm not  seeing it fixed

lucywyman:
@scoopex Can you squash the 2 commits into 1 commit? And perhaps add more details to your commit message about what you're adding, how to use it, and why? Doesn't need to be long, just a line or 2 :)

```
git rebase -i HEAD~2
```
Then change the "pick" on the second line to "f". Let us know if you need help or want us to do it for ya!

PR #507, opened 2018-07-10
(maint) Use platform-specific separator when updating PATH

MikaelSmith:
When invoking the bolt manifest block compiler, we update PATH to put
the current version of Ruby at its head. On Windows that update used the
wrong path list separator; fix it.


MikaelSmith:
This may actually be part of BOLT-634. Rolling into https://github.com/puppetlabs/bolt/pull/505.

PR #508, opened 2018-07-10
(BOLT-684) Ensure we can operate on a ResultSet returned by apply

MikaelSmith:
Previously if only an `apply` was used, nothing would refer to ResultSet
and the type would not be loaded. Ensure the type is always loaded so we
can use it as the result of an `apply`.

### DIFF
--- a/spec/fixtures/apply/basic/plans/init.pp
+++ b/spec/fixtures/apply/basic/plans/init.pp
@@ -7,9 +7,11 @@ plan basic(TargetSpec $nodes) {
   return apply($nodes) {
     file { '/root/test/':
       ensure => directory,
-    } -> file { "/root/test/hello.txt":
-      ensure => file,
+    } -> file { '/root/test/hello.txt':
+      ensure  => file,
       content => "hi there I'm ${$facts['os']['family']}\n",
     }
+  }.map |$r| {
+    $r['resources']
   }
 }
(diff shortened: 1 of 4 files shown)


nicklewis:
An alternate solution to this is to add `return_type 'Boltlib::PlanResult'` to the `run_plan()` function definition. That will cause `Boltlib::PlanResult` to be loaded, which is an alias that mentions `Result`, `ResultSet`, and `Target` (our three datatypes). That will ensure that whenever we run a plan, those types will always be loaded.

spec/fixtures/apply/basic/plans/init.pp
```diff
@@ -7,9 +7,11 @@
   return apply($nodes) {
     file { '/root/test/':
       ensure => directory,
-    } -> file { "/root/test/hello.txt":
-      ensure => file,
+    } -> file { '/root/test/hello.txt':
+      ensure  => file,
       content => "hi there I'm ${$facts['os']['family']}\n",
     }
+  }.map |$r| {
```
donoghuc: What is the purpose of this? It seems like there is no key `$r['resources']` so `null` is returned. A key `resources` is nested under `metrics`. `$r['metrics']['resources']`

MikaelSmith: That's surprising. What we should get back from `apply` (in these tests) is the catalog produced from compiling the manifest block. It iterates over the result of each node, and expects to access the `resources` property of each catalog.

Running the plan directly won't work, it has to be run with apply_catalog.rb overridden by `spec/fixtures/apply/apply_catalog.sh`.

donoghuc: Ahhh, I see. I was running as a plan outside of the test.

PR #509, opened 2018-07-10
(BOLT-635) Parallelize multi-target apply

MikaelSmith:
Refactors the executor to expose `queue_execute` and `await_results` as async components. `apply` uses them to parallelize compiling and executing manifest blocks across multiple targets, using separate thread pools for compilation (a local CPU-intensive operation) and execution. Compilation thread pool is limited by a new `compile-concurrency` config option.

### DIFF
--- a/lib/bolt/config.rb
+++ b/lib/bolt/config.rb
@@ -2,6 +2,7 @@
 
 require 'yaml'
 require 'logging'
+require 'concurrent'
 require 'bolt/cli'
 require 'bolt/transport/ssh'
 require 'bolt/transport/winrm'
@@ -25,6 +26,7 @@ module Bolt
 
   Config = Struct.new(
     :concurrency,
+    :'compile-concurrency',
     :format,
     :trace,
     :inventoryfile,
@@ -39,6 +41,7 @@ module Bolt
 
     DEFAULTS = {
       concurrency: 100,
+      'compile-concurrency': Concurrent.processor_count,
       transport: 'ssh',
       format: 'human',
       modulepath: [],
@@ -132,7 +135,7 @@ module Bolt
         self[:modulepath] = data['modulepath'].split(File::PATH_SEPARATOR)
       end
 
-      %w[inventoryfile concurrency format puppetdb hiera-config color transport].each do |key|
+      %w[inventoryfile concurrency compile-concurrency format puppetdb hiera-config color transport].each do |key|
         if data.key?(key)
           self[key.to_sym] = data[key]
         end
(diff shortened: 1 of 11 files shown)


MikaelSmith:
Added a test for the compile concurrency implementation.

spec/bolt/applicator_spec.rb
```diff
@@ -5,39 +5,87 @@
 require 'bolt/target'
 
 describe Bolt::Applicator do
-  let(:inventory) { nil }
-  let(:applicator) { Bolt::Applicator.new(inventory, nil, :mod, :pdb) }
+  let(:uri) { 'foobar' }
```
lucywyman: I assume this is leftover?

lib/bolt/config.rb
```diff
@@ -39,6 +41,7 @@ def initialize(transport, uri = nil)
 
     DEFAULTS = {
       concurrency: 100,
+      'compile-concurrency': Concurrent.processor_count,
```
lucywyman: Should we add a check to `validate` to ensure the value is less than or equal to this? We can error or warn + use processor count if it's too high

donoghuc:
In order to test this out I did some benchmarking. 
**Setup**
1. Install puppet agent on 10 `rastasheep/ubuntu-sshd` docker nodes
2. write following plan: 
```
  return apply($nodes) {
    notify { "hello ": }
  }
}
```
First I measure using the execution times that bolt reports. I ran the plan above against the 10 targets 3 times each with parallel and sequential apply implementation. 
```
seconds       mean      stddev
parallel:     5.2         0.67
sequential:   11.4        0.18
```
Next I used a benchmarking tool that @MikaelSmith found. 
https://github.com/Gabriel439/bench
**Parallel**
```
time                 6.110 s    (5.002 s .. 6.929 s)
                     0.995 R²   (0.993 R² .. 1.000 R²)
mean                 6.282 s    (6.086 s .. 6.463 s)
std dev              227.2 ms   (192.4 ms .. 243.6 ms)
variance introduced by outliers: 19% (moderately inflated)
```
**Sequential**
```
time                 24.19 s    (22.79 s .. 25.45 s)
                     1.000 R²   (0.998 R² .. 1.000 R²)
mean                 24.15 s    (24.03 s .. 24.38 s)
std dev              221.6 ms   (9.449 ms .. 280.7 ms)
variance introduced by outliers: 19% (moderately inflated)
```
Having not really dug in to how the benchmarking tool executes it seems like we can safely say running in parallel is at least twice as fast as sequential.

lib/bolt/config.rb
lucywyman: :woman_shrugging:  I don't expect users will try to run with compile-concurrency of 0, but didn't realize there was value to setting it above the cpu count. Prolly fine as is

MikaelSmith: Added validation anyway.

PR #511, opened 2018-07-11
(BOLT-631) Support Hiera config for plan apply

donoghuc:
When compiling catalogs for apply bolt should configure and use hiera. This data should by default come from the boltdir not the global or user specific puppet path.

### DIFF
--- a/lib/bolt/applicator.rb
+++ b/lib/bolt/applicator.rb
@@ -7,11 +7,12 @@ module Bolt
   Task = Struct.new(:name, :implementations, :input_method)
 
   class Applicator
-    def initialize(inventory, executor, modulepath, pdb_config)
+    def initialize(inventory, executor, modulepath, pdb_config, hiera_config)
       @inventory = inventory
       @executor = executor
       @modulepath = modulepath
       @pdb_config = pdb_config
+      @hiera_config = hiera_config ? validate_hiera_config(hiera_config) : nil
     end
 
     private def libexec
@@ -31,6 +32,7 @@ module Bolt
         code_ast: ast,
         modulepath: @modulepath,
         pdb_config: @pdb_config,
+        hiera_config: @hiera_config,
         target: {
           name: target.host,
           facts: @inventory.facts(target),
@@ -48,6 +50,16 @@ module Bolt
 
       raise ApplyError.new(target.to_s, err) unless stat.success?
       JSON.parse(out)
+    end
+
+    def validate_hiera_config(hiera_config)
+      if File.exist?(File.path(hiera_config))
+        data = File.open(File.path(hiera_config), "r:UTF-8") { |f| YAML.safe_load(f.read) }
+        unless data['version'] == 5
+          raise ApplyError.new("All Targets", "Hiera v5 is required.")
+        end
+        hiera_config
+      end
     end
 
     def apply(args, apply_body, scope)
--- a/lib/bolt/catalog.rb
+++ b/lib/bolt/catalog.rb
@@ -55,7 +55,7 @@ end
 
 module Bolt
   class Catalog
-    def with_puppet_settings
+    def with_puppet_settings(hiera_config)
       Dir.mktmpdir('bolt') do |dir|
         cli = []
         Puppet::Settings::REQUIRED_APP_SETTINGS.each do |setting|
@@ -63,6 +63,7 @@ module Bolt
         end
         Puppet.settings.send(:clear_everything_for_tests)
         Puppet.initialize_settings(cli)
+        Puppet.settings[:hiera_config] = hiera_config
         # self.class.configure_logging
         yield
       end
@@ -101,7 +102,7 @@ module Bolt
         Bolt::PuppetDB::Client.from_config(pdb_config)
       end
 
-      with_puppet_settings do
+      with_puppet_settings(request['hiera_config']) do
         Puppet[:code] = ''
         Puppet[:node_name_value] = target['name']
         Puppet::Pal.in_tmp_environment(
--- a/lib/bolt/config.rb
+++ b/lib/bolt/config.rb
@@ -31,6 +31,7 @@ module Bolt
     :log,
     :modulepath,
     :puppetdb,
+    :'hiera-config',
     :color,
     :transport,
     :transports
@@ -131,7 +132,7 @@ module Bolt
         self[:modulepath] = data['modulepath'].split(File::PATH_SEPARATOR)
       end
 
-      %w[inventoryfile concurrency format puppetdb color transport].each do |key|
+      %w[inventoryfile concurrency format puppetdb hiera-config color transport].each do |key|
         if data.key?(key)
           self[key.to_sym] = data[key]
         end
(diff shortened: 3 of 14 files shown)


donoghuc:
My testing strategy so far: 
Hiera data is configured in ~/.puppetlabs/bolt/
```
# /home/cas/.puppetlabs/bolt/hiera.yaml
---
version: 5

defaults:  # Used for any hierarchy level that omits these keys.
  datadir: /home/cas/.puppetlabs/bolt/hiera_data  # This path is relative to the environment -- <ENVIRONMENT>/data
  data_hash: yaml_data  # Use the built-in YAML backend.

hierarchy:
  - name: "Common"
    path: "common.yaml"
```
```
# /home/cas/.puppetlabs/bolt/hiera_data/common.yaml
---
cas_data: "cas"
```
This is the test plan
```
  return apply($nodes) {
    notify { "hello ${hiera('cas_data')}": }
  }
}
```

Invocation and output
```
cas@cas-ThinkPad-T460p:~/working_dir/bolt$ bolt plan run test -n x6x9wezjmz9dbih.delivery.puppetlabs.net
Starting: apply catalog on x6x9wezjmz9dbih.delivery.puppetlabs.net
Finished: apply catalog with 0 failures in 1.31 sec
[
  {
    "node": "x6x9wezjmz9dbih.delivery.puppetlabs.net",
    "status": "success",
    "result": {
       *****ABBREVIATED******
      "logs": [
        {
          "level": "notice",
          "message": "hello cas",   <- got the hiera data
       *****ABBREVIATED******
```
This first iteration simply provides the path to the hiera.yaml supplied in either the default location(s) or specified in a configfile. I noticed in the ticket specification that there is mention of a default datadir if one is not provided. Does this imply that we should be parsing the hiera config and looking for a datadir?

lib/bolt/catalog.rb
```diff
@@ -92,6 +93,16 @@ def generate_ast(code)
       end
     end
 
+    def validate_hiera_config(hiera_config)
```
donoghuc: Is this the appropriate place to check for hiera 5 version? Error looks like: 
```
{
  "kind": "bolt/apply-error",
  "msg": "Apply failed to compile for Target('x6x9wezjmz9dbih.delivery.puppetlabs.net', {\"connect-timeout\"=>10, \"tty\"=>false, \"host-key-check\"=>false, \"private-key\"=>\"~/.ssh/id_rsa-acceptance\"}): /home/cas/working_dir/bolt/lib/bolt/catalog.rb:100:in `validate_hiera_config': Hiera 5 required. Version 8 detected. (Bolt::FileError)\n\tfrom /home/cas/working_dir/bolt/lib/bolt/catalog.rb:115:in `compile_catalog'\n\tfrom /home/cas/working_dir/bolt/libexec/bolt_catalog:32:in `<main>'\n",
  "details": {
  }
}
```
I am holding off on writing test for this because I feel like it might be incorrect.

MikaelSmith: I think we could and should check earlier, maybe at the beginning of the `apply` call after testing arguments.

donoghuc: Is `bolt/file-error` ok? Moved to `apply` call after testing args and message looks like:
```
{
  "kind": "bolt/file-error",
  "msg": "Hiera 5 required. Version 8 detected.",
  "details": {
    "path": "/home/cas/.puppetlabs/bolt/hiera.yaml"
  }
}
```

MikaelSmith: Your test specifies `version: 8` explicitly?

I'd make it a `bolt/apply-error`, I think that makes it clear what it's related to.

donoghuc: I will change to `bolt/apply-error` and write a test for it. I was just manually testing with setting `version: 8` in my hiera.yaml. I had version interpolated in the error message string. Thanks @MikaelSmith  !

MikaelSmith: I think the most likely scenario is that `version` won't be set at all, which implies version 3.

donoghuc: Ok, I will leave out the version interpolation in the error message. Current implementation will throw the error when `version` is not set. This should guard against the hiera 3 default.

lib/bolt/catalog.rb
```diff
@@ -92,6 +93,16 @@ def generate_ast(code)
...
+      if File.exist?(File.path(hiera_config))
+        data = Bolt::Util.read_config_file(hiera_config, nil, 'hiera_config')
```
MikaelSmith: This is overly specific. `read_config_file` has a specific use case: at takes an explicitly specified path (that may be nil), and a list of fallbacks. If the specified path is not nil, then it will error if not found; for the fallbacks, it'll only log if not found.

This could simply be `data = File.open(path, "r:UTF-8") { |f| YAML.safe_load(f.read) }`.

We should also consider where and whether we should error if `hiera_config` is explicitly configured in bolt.yaml but not found.

MikaelSmith: I guess on the plus side, `read_config_file` gives us consistent logging.

lib/bolt/config.rb
```diff
@@ -31,6 +31,7 @@ def initialize(transport, uri = nil)
     :log,
     :modulepath,
     :puppetdb,
+    :hiera_config,
```
MikaelSmith: We lean towards hyphens rather than underscore for config options. This would be `:'hiera_config'`.

MikaelSmith: I believe the ticket specification was wrong on how to format the option.

lib/bolt/applicator.rb
```diff
@@ -50,11 +52,23 @@ def compile(target, ast, plan_vars)
...
+        hiera_config
+      end
+    end
+
     def apply(args, apply_body, scope)
       raise(ArgumentError, 'apply requires a TargetSpec') if args.empty?
       type0 = Puppet.lookup(:pal_script_compiler).type('TargetSpec')
       Puppet::Pal.assert_type(type0, args[0], 'apply targets')
 
+      @hiera_config = validate_hiera_config(@hiera_config)
```
donoghuc: Would it make sense to do the validation upon assignment in initialization?

lib/bolt/config.rb
```diff
@@ -248,6 +254,7 @@ def update(options)
     def load_file(path)
       data = Bolt::Util.read_config_file(path, [default_config], 'config')
       update_from_file(data) if data
+      validate_hiera_conf(data ? data['hiera-config'] : nil)
```
donoghuc: Idea here is if `hiera-config` is specified in `configfile` or exists in default location then make sure it can be parsed (with the expected bolt read_config_file logging). 

Validation that hiera v5 is executed only if there is an apply block in a plan. That way we do not error out if a hiera.yaml file is laying around somewhere but the `apply` is not used in a plan.

lib/bolt/applicator.rb
MikaelSmith: I think we should do it as early as possible.

lib/bolt/config.rb
MikaelSmith: That makes sense.

lib/bolt/applicator.rb
MikaelSmith: I think it makes sense to do when the Applicator is initialized. It may trigger if you're not using apply, but the hiera config is entirely opt-in, so if you're doing something with Puppet I think we should always say you have an invalid hiera config. Later we may use it for things besides apply.

MikaelSmith:
I want to hold off briefly to see if we can get the config refactor merged first.

PR #514, opened 2018-07-12
(BOLT-699) Simplify boltdir and config loading

nicklewis:
Previously, the config class had functionality whereby defaults could be
set by passing a hash to the initialize method. This was only used for
tests, and complicated the config class by adding another level to the
"hierarchy". It also interacted in confusing ways with settings whose
defaults were already somewhat dynamic (anything depending on boltdir).

We also previously built the config object up incrementally, instantiating
it with optional default overrides, and then later parsing a config file,
and then overriding CLI options. This meant several intermediate steps
before getting to the final value of the config.

The config object now takes a config data hash (which is the result of
loading the config file) as well as an optional hash of overrides (which
are the result of processing the CLI arguments). This allows the entire
final state of config to be computed immediately, and it also provides a
way for tests to override settings by the same mechanism that users do.

This also introduces a Boltdir class that captures the particular behavior of
the Boltdir, separate from the config class. The config takes a Boltdir during
initialization and uses the Boltdir to determine some of its defaults.

We now accept only one of --configfile or --boltdir. If --boltdir is used,
then the config file will be read from <boltdir>/bolt.yaml. If
--configfile is specified, then boltdir will be set to the parent dir of
the config file. The default behavior when neither is specified remains the
same: boltdir will be found in the current directory hierarchy, or it will
fall back to ~/.puppetlabs/bolt.


lib/bolt/bolt_option_parser.rb
```diff
@@ -189,7 +189,7 @@ def initialize(options)
       end
       define('--boltdir FILEPATH',
              'Specify what Boltdir to load config from (default: autodiscovered from current working dir)') do |path|
-        @options[:configfile] = path
+        @options[:boltdir] = path
       end
       define('--configfile FILEPATH',
              'Specify where to load config from (default: ~/.puppetlabs/bolt/bolt.yaml)') do |path|
```
MikaelSmith: What if we had `--configfile` directly set `:boltdir`? I'm in favor of fewer options to handle later.

lib/bolt/cli.rb
```diff
@@ -99,6 +95,15 @@ def parse
...
       validate(options)
 
+      @config = if options[:configfile]
+                  Bolt::Config.from_file(options[:configfile], options)
+                else
+                  boltdir = options[:boltdir] ?  Bolt::Boltdir.new(options[:boltdir]) : Bolt::Boltdir.find_boltdir(Dir.pwd)
+                  Bolt::Config.from_boltdir(boltdir, options)
+                end
+
+      Bolt::Logger.configure(config.log, config.color)
```
MikaelSmith: I almost want an `assert(@pal.nil?)` here to make sure we don't accidentally call `pal` before `@config` is finished. But tests will probably catch something that obvious.

lib/bolt/config.rb
```diff
@@ -69,35 +53,54 @@ def initialize(transport, uri = nil)
       local: {}
     }.freeze
 
-    BOLTDIR_NAME = 'Boltdir'
+    def self.default
+      new(Bolt::Boltdir.new('.'), {})
```
MikaelSmith: Having this default to current working directory seems a little odd. Is that used in some way?

lib/bolt/pal.rb
```diff
@@ -119,7 +119,7 @@ def with_bolt_executor(executor, inventory, pdb_client = nil, &block)
         bolt_executor: executor,
         bolt_inventory: inventory,
         bolt_pdb_client: pdb_client,
-        apply_executor: Applicator.new(inventory, executor, full_modulepath(@config[:modulepath]), @config.puppetdb)
+        apply_executor: Applicator.new(inventory, executor, full_modulepath(@modulepath), pdb_client)
```
MikaelSmith: Shouldn't this pass just `@modulepath`?

spec/bolt/logger_spec.rb
```diff
@@ -42,17 +42,16 @@ def initialize(*args)
...
-        log: {
-          'file:/bolt.log' => {
-          },
-          'file:/debug.log' => {
-            level: :debug,
-            append: false
-          }
+    let(:appenders) {
+      {
+        'console' => {},
```
MikaelSmith: Why does this need to exist now?

spec/bolt/transport/ssh_spec.rb
```diff
@@ -14,9 +14,11 @@
   include BoltSpec::Files
   include BoltSpec::Task
 
+  let(:boltdir) { Bolt::Boltdir.new('') }
```
MikaelSmith: What's the difference between `Bolt::Boltdir.new('')` and `Bolt::Boltdir.new('.')`?

lib/bolt/cli.rb
nicklewis: Yeah I think I'm comfortable letting tests handle that. Any of our tests that run the CLI should find it.

spec/bolt/logger_spec.rb
nicklewis: In the CLI class, the argument passed to `Logging.configure` comes from the Config object, which always has a default for `console`. The tests were previously configuring a Config object with some files and then passing it in. Now we're bypassing the Config object, so this was changed to explicitly include the `console` section.

But it seems like if I just remove it and update the one test that is _counting_ the number of appenders, it works. I'll do that.

spec/bolt/transport/ssh_spec.rb
nicklewis: One creates an empty base path and one creates a base path of `.`

In practice, I don't think there's a difference, because empty string + "/foo" is the same as "./foo"

lib/bolt/bolt_option_parser.rb
MikaelSmith: As I read more, this wouldn't really make sense. configfile can have a different name, and the logic around loading them is a little different.

lib/bolt/boltdir.rb
```diff
@@ -0,0 +1,38 @@
+require 'pathname'
+
+module Bolt
+  class Boltdir
+    BOLTDIR_NAME = 'Boltdir'
```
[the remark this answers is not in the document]
nicklewis: Fixed this by adding `# frozen_string_literal: true`

spec/bolt/transport/ssh_spec.rb
MikaelSmith: I'd hope for either an explanation around the expectations, or to use a consistent default to mean "we don't care".

nicklewis:
Pushed a commit to replace all the `Boltdir.new('')` with `Boltdir.new('.')` for consistency.

MikaelSmith:
👍 other than wanting a ticket for the bug fix.

PR #515, opened 2018-07-12
(BOLT-648, BOLT-649) Support _catch_errors, _run_as

MikaelSmith:
Add support for `_catch_errors` and `_run_as` to `apply`. Previously
`_catch_errors=true` was the default behavior, now it will default to
false and allow it to be changed to true.

### DIFF
--- a/spec/bolt/applicator_spec.rb
+++ b/spec/bolt/applicator_spec.rb
@@ -134,7 +134,7 @@ describe Bolt::Applicator do
     it 'captures compile errors in a result set' do
       expect(applicator).to receive(:compile).and_raise('Something weird happened')
 
-      resultset = applicator.apply([uri], :body, {})
+      resultset = applicator.apply([uri, '_catch_errors' => true], :body, {})
       expect(resultset).to be_a(Bolt::ResultSet)
       expect(resultset.count).to eq(1)
       expect(resultset.first.ok).to be(false)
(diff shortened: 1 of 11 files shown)


MikaelSmith:
Relies on work in https://github.com/puppetlabs/bolt/pull/508 and is going to have some conflicts with https://github.com/puppetlabs/bolt/pull/509 and https://github.com/puppetlabs/bolt/pull/512.

spec/bolt/applicator_spec.rb
```diff
@@ -73,8 +73,9 @@
       end
 
       targets = [Bolt::Target.new('node1'), Bolt::Target.new('node2'), Bolt::Target.new('node3')]
-      results = targets.map { |target| Bolt::Result.new(target) }
-      allow_any_instance_of(Bolt::Transport::SSH).to receive(:batch_task).and_return(*results)
+      allow_any_instance_of(Bolt::Transport::SSH).to receive(:batch_task) do |_, batch|
```
MikaelSmith: This test was passing before because the error when we returned the result for the wrong target was ignored. Now without `_catch_errors=true`, it gets thrown.

MikaelSmith:
The new acceptance test appears to have some issues, working on it.

MikaelSmith:
Got it working. There were some surprises around tty with the facts plan and sudo on macOS.

PR #110, opened 2017-10-13
(BOLT-200) Prevent WinRM stream reading deadlock

Iristyle:
- Previously, the stdout stream was read / parsed synchronously prior
   to draining the stderr stream, which could lead to a deadlock if
   too much output was produced on stderr

 - The correct way to drain stdout / stderr is by responding to events
   and sending them directly to the host. Add a few ScriptBlocks to
   handle OutputDataReceived and ErrorDataReceived from the process.

 - Note for output to not appear correctly on stderr, the code had
   to stop using the Write-Error cmdlet and had to move to using
   $Host.UI.WriteErrorLine

   This has the side effect of not appending the extra newlines
   that were introduced as part of
   9fff248

PR #519, opened 2018-07-17
(BOLT-698) Add tests for output ordering

MikaelSmith:
Ensure many lines our output in the right order. Acceptance test
currently expected to fail.


lib/bolt/transport/winrm/connection.rb
```diff
@@ -408,30 +315,21 @@ def execute(command)
...
-  Path = "#{path}"
-  Arguments = $quoted_array -Join ' '
-  Timeout = #{timeout}
-  #{stdin.nil? ? '' : "StdinInput = @'\n" + stdin + "\n'@"}
-}
-
-# winrm gem checks $? prior to using $LASTEXITCODE
-# making it necessary to exit with the desired code to propagate status properly
-exit $(Invoke-Interpreter @invokeArgs)
+#{exec_cmd}
```
MikaelSmith: @Iristyle can you tell me why this is a bad idea? It seems to work for all our current test cases, and addresses the issue.

MikaelSmith:
apply_winrm acceptance test is failing, working on it.

MikaelSmith:
Needed to correctly quote stdin. All tests should now pass.

lib/bolt/transport/winrm/connection.rb
```diff
@@ -408,30 +315,21 @@ def execute(command)
...
             "'" + arg.gsub("'", "''") + "'"
-          end.join(',')
-
+          end.join(' ')
+
+          exec_cmd =
+            if stdin.nil?
+              "& #{path} #{quoted_args}"
+            else
+              "@'\n#{stdin}\n'@ | & #{path} #{quoted_args}"
```
Iristyle: Unfortunately this isn't how `stdin` works on Windows. That's a PowerShell pipe operator, not a stdin redirection operator.

MikaelSmith: I don't think I understand where that matters. This will always be a non-powershell script or task, and the examples I've tried all read happily read this from stdin.

MikaelSmith: I'll give it a run through with .bat, .py, and .rb as some examples to try.

MikaelSmith: This may be a problem for powershell tasks that use input_method=stdin. Need to explore that.

Iristyle: Apparently this *will* redirect in some circumstances (but not to a PowerShell script).

For instance, given this batch file:

```batch
@echo off

ECHO Called %~nx0
ECHO From %~dp0
setlocal
for /F "tokens=*" %%a in ('more') do (
  echo #%%a
)
```

The output is

```
C:\Users\Administrator> 'foo' | C:\source\bolt\foo.bat
Called foo.bat
From C:\source\bolt\
#foo
```


There are some tricks you can use to work around PowerShell scripts treating that as pipeline input - https://stackoverflow.com/questions/38667584/getting-stdin-into-the-powershell-stream

Iristyle: Also - I'm not sure that this method will work properly depending on system codepage, encoding of the passed string, etc. That would be worth investigating by calling a Python or Ruby script.

Iristyle:
Original implementation was done in #67 and #110

lib/bolt/transport/winrm/connection.rb
MikaelSmith: I've added some additional testing with powershell scripts, they seem to read stdin happily via either `$Input` or `[Console]::In.ReadLine()`. I think the way we're invoking the script here results in a powershell subprocess that gets passed data on the pipe as stdin.

I'm not exactly sure how to test the codepage concerns you mention, and don't think we had any prior testing around that. I can mess around with UTF-8 to a Ruby script. Were you also imagining trying that with a different code page setup on the target?

MikaelSmith: I think I've dealt with encoding now too.

spec/integration/winrm_spec.rb
```diff
@@ -35,14 +33,24 @@
       expect(result['_error']['msg']).to eq('The command failed with exit code 1')
     end
 
-    it 'runs a task', :reset_puppet_settings do
-      result = run_one_node(%W[task run #{stdin_task} message=somemessage] + config_flags)
-      expect(result['_output'].strip).to match(/STDIN: {"messa/)
+    it 'runs a task reading from stdin', :reset_puppet_settings do
+      result = run_one_node(%w[task run sample::winstdin message=❄somemessage] + config_flags)
```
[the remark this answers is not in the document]
MikaelSmith: @Iristyle these tests wouldn't have worked before either. They work locally with Nano Server, since it defaults to a UTF-8 encoding/codepage. I'm having trouble finding the right way to reset it in AppVeyor so I can have a consistent target codepage for testing.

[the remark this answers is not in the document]
MikaelSmith: Something funny's happening with this output. Not sure how to fix it.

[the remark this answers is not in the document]
MikaelSmith: I can't get these tests to work with Windows Server 2012 R2 with either the old or new code. The unicode doesn't round-trip through the script execution correctly.

lib/bolt/transport/winrm/connection.rb
```diff
@@ -408,30 +315,22 @@ def execute(command)
...
-  Arguments = $quoted_array -Join ' '
-  Timeout = #{timeout}
-  #{stdin.nil? ? '' : "StdinInput = @'\n" + stdin + "\n'@"}
-}
-
-# winrm gem checks $? prior to using $LASTEXITCODE
-# making it necessary to exit with the desired code to propagate status properly
-exit $(Invoke-Interpreter @invokeArgs)
+$OutputEncoding = [Console]::OutputEncoding
+#{exec_cmd}
```
MikaelSmith: @Iristyle just noticed we always execute a subprocess because https://github.com/puppetlabs/bolt/blob/0.21.4/lib/bolt/transport/winrm.rb#L158-L182.

MikaelSmith: Given that, I think this PR is fine to merge.

PR #520, opened 2018-07-17
(BOLT-705) Bundled content for plan/task only

donoghuc:
Only generate a list of bundled plan/task content when running a plan or task. Building list of bundled content is expensive and not required when not running a plan or task.

### DIFF
--- a/lib/bolt/cli.rb
+++ b/lib/bolt/cli.rb
@@ -385,14 +385,16 @@ module Bolt
     end
 
     def bundled_content
-      default_content = Bolt::PAL.new([], nil)
-      plans = default_content.list_plans.each_with_object([]) do |iter, col|
-        col << iter&.first
+      if %w[plan task].include?(options[:subcommand])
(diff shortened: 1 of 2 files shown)


donoghuc:
I will add a spec test for this. Wanted to make sure this reflects what we discussed in hipchat room.

lib/bolt/cli.rb
```diff
@@ -385,14 +385,16 @@ def outputter
     end
 
     def bundled_content
-      default_content = Bolt::PAL.new([], nil)
-      plans = default_content.list_plans.each_with_object([]) do |iter, col|
-        col << iter&.first
-      end
-      tasks = default_content.list_tasks.each_with_object([]) do |iter, col|
-        col << iter&.first
+      if %w[plan task].include?(options[:subcommand])
```
MikaelSmith: I probably would've implemented this slightly differently (accessing options from random methods feels weird to me, so we could've just skipped calling bundled_content in the first place unless a plan or task). But it works and makes sense.

PR #521, opened 2018-07-18
(BOLT-523) Add command to install a Puppetfile in Boltdir

nicklewis:
This introduces a `bolt puppetfile install` command which will install
modules from a Puppetfile located in a Boltdir. Modules are installed to
the first location in the modulepath for that Boltdir (excluding the
directories automatically added by Bolt).


lib/bolt/cli.rb
```diff
@@ -6,6 +6,7 @@
 require 'io/console'
 require 'logging'
 require 'optparse'
+require 'r10k/action/puppetfile/install'
```
MikaelSmith: We should make sure this doesn't accidentally load Puppet. It may be worth restricting it to the install workflow where it's used.

adreyer: r10k doesn't depend on puppet so it shouldn't

pre-docs/bolt_installing_modules.md
```diff
@@ -0,0 +1,26 @@
...
+Bolt uses a Puppetfile to specify which modules to install. The Puppetfile
+should be in the Boltdir, which can either be a project-specific directory
+named `Boltdir` at the root of your project, or a global directory in
+`$HOME/.puppetlabs/bolt`.
+
+The Puppetfile contains a list of modules and versions to install. If the
+modules have dependencies on other modules, those dependencies will need to be
+listed in the Puppetfile as well.
+
+By default, modules will be downloaded from the Puppet Forge. Modules can be
```
adreyer: Are we intentionally avoiding linking to the full Puppetfile docs after these examples?

pre-docs/bolt_installing_modules.md
```diff
@@ -0,0 +1,26 @@
...
+installed from git repositories instead by setting the `git` and `ref`
+properties.
+
+```
+mod 'puppetlabs/package', '0.2.0'
+mod 'puppetlabs/service', '0.3.1'
+mod 'puppetlabs/puppetlabs-facter_task', git: 'git@github.com:puppetlabs/puppetlabs-facter_task.git', ref: 'master'
```
adreyer: Do we want an example of a local module or should we tell users to put local modules in Boltdir/site? We don't support modulepaths relative to boltdir yet do we?

pre-docs/running_tasks_and_plans_with_bolt.md
```diff
@@ -20,8 +20,8 @@ tasks based on the results of previous tasks.
 
 Tasks and plans are packaged in Puppet modules, so you can install them as you
 would any module from the Forge. Install from the command line with the puppet
-module install command, or install and manage them with a Puppetfile and Code
-Manager. For more details, see the installing modules documentation.
+module install command, or install and manage them with a Puppetfile. For more
```
adreyer: Install them from the commandline with the bolt puppetfile command

pre-docs/bolt_installing_modules.md
nicklewis: I'll add a note about that. I don't think we support relative modulepaths, but I'm also not 100% sure what would happen if you tried.

PR #524, opened 2018-07-19
(BOLT-613) Move plan spec helpers to lib/bolt_spec

MikaelSmith:
Move plan spec helpers to `lib/bolt_spec` so they will be shipped as
part of the Bolt gem and available to modules to test plans with
`require 'bolt_spec/plans'`.


MikaelSmith:
I've tested this out with `puppetlabs-facts`. Update the Gemfile, delete `spec/lib`, and tests still work.

MikaelSmith:
Updated.

PR #525, opened 2018-07-23
(BOLT-691) Fix running task with tty echoing input

MikaelSmith:
When running a task with tty and using sudo to execute as a non-root
user, if stdin isn't consumed (like in the `facts` task) but is sent
over SSH it appears to be echoed back. Disable echo mode when using tty
to ensure it's not echoed.

Note that when using `--tty`, Puppet prints errors on stdout instead of
stderr.

### DIFF
--- a/acceptance/tests/apply_nonroot.rb
+++ b/acceptance/tests/apply_nonroot.rb
@@ -22,16 +22,6 @@ test_name "bolt plan run with should apply manifest block on remote hosts via ss
     end
   end
 
-  step 'disable requiretty for root user' do
-    create_remote_file(ssh_nodes, "/etc/sudoers.d/#{user}", <<-FILE)
-Defaults:root !requiretty
-FILE
-
-    teardown do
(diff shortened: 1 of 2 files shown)


acceptance/tests/apply_nonroot.rb
```diff
@@ -63,7 +54,7 @@
       result = json.select { |n| n['node'] == host }
       assert_equal('failure', result[0]['status'],
                    "The task did not fail on #{host}")
-      assert_match(/Permission denied/, result[0]['result']['_error']['msg'])
```
MikaelSmith: This behavior is weird, going to look into it a bit more. Presumably Puppet is returning a different exit code based on `--tty`.

MikaelSmith: Message is in `_output` because it appears on stdout rather than stderr with `--tty`. We should probably discourage using `--tty` generally.

PR #526, opened 2018-07-24
(BOLT-709) Error 'apply' when report identifies failure, and provide human-readable summary

MikaelSmith:
Ideas for how to display apply failure output.

### DIFF
--- a/lib/bolt/error.rb
+++ b/lib/bolt/error.rb
@@ -63,6 +63,17 @@ module Bolt
     end
   end
 
+  class ApplyFailure < RunFailure
+    def initialize(result_set)
+      super(result_set, 'apply', 'catalog')
+      @kind = 'bolt/apply-failure'
+    end
+
+    def to_s
+      result_set.select(&:error_hash).map { |result| result.error_hash['msg'] }.join("\n")
+    end
+  end
+
   class PlanFailure < Error
     def initialize(*args)
       super(*args)
--- a/lib/bolt/outputter/human.rb
+++ b/lib/bolt/outputter/human.rb
@@ -165,6 +165,8 @@ module Bolt
       def print_plan_result(plan_result)
         if plan_result.value.nil?
           @stream.puts("Plan completed successfully with no result")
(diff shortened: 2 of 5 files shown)


MikaelSmith:
One example of what output would look like with
```
plan foo::simple (TargetSpec $nodes) {
  return apply($nodes) {
    file { '/tmp/does/not/exist': ensure => directory }
    notify { 'hello': }
    alert('now')
  }
}
```
screen shot 2018-07-24 at 11 56 19 am

MikaelSmith:
Alternative where we only emit error logs as part of an error result.

MikaelSmith:
TODO: look at puppetlabs-apply for ideas of how we might want to reduce the report to a simpler form for printing.

lib/bolt/error.rb
```diff
@@ -50,19 +50,32 @@ def initialize(msg)
   class RunFailure < Error
     attr_reader :result_set
 
-    def initialize(result_set, action, object)
+    def generate_msg(result_set, action, object)
```
adreyer: I think we can just override `msg` and store the and grab the action/object/result_set from details

lib/bolt/outputter/human.rb
```diff
@@ -165,6 +165,8 @@ def print_plan_info(plan)
       def print_plan_result(plan_result)
         if plan_result.value.nil?
           @stream.puts("Plan completed successfully with no result")
+        elsif plan_result.value.is_a? Bolt::ApplyFailure
```
adreyer: Lets create a ticket to revisit human output for plans that error. I suspect there is something standard we can do here.

MikaelSmith: Added BOLT-716.

lib/bolt/error.rb
MikaelSmith: Overrode `to_s`, as that seems to be what everything else - inspect, message, msg - uses.

nicklewis:
Tried this out in one of my own use cases and the failure output is :+1:. The compilation failure is better too.

PR #532, opened 2018-07-25
(BOLT-640) Build and sync plugin tarball to agents

nicklewis:
When using the apply() keyword, we now copy the contents of the `lib`
and `files` directories of every module to the agent. This allows native
types and providers to be used, as well as for puppet:// file sources to
resolve correctly.

This is accomplished by packing those files into a gzipped tarball,
which is then passed as a parameter to the apply_catalog internal
task. That tarball is then written to disk on the agent, unpacked, and
set as the modulepath for the catalog application.

### DIFF
--- a/lib/bolt/applicator.rb
+++ b/lib/bolt/applicator.rb
@@ -1,9 +1,12 @@
 # frozen_string_literal: true
 
+require 'base64'
+require 'concurrent'
+require 'find'
 require 'json'
 require 'logging'
+require 'minitar'
 require 'open3'
-require 'concurrent'
 require 'bolt/util/puppet_log_level'
 
 module Bolt
@@ -19,6 +22,7 @@ module Bolt
 
       @pool = Concurrent::ThreadPoolExecutor.new(max_threads: max_compiles)
       @logger = Logging.logger[self]
+      @plugin_tarball = Concurrent::Delay.new { build_plugin_tarball }
     end
 
     private def libexec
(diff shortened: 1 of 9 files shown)


nicklewis:
This still needs lots of tests, but I wanted to get it up for an early look.

MikaelSmith:
I think this would be best tested by modifying the existing apply acceptance tests to include a file from source and custom type/provider.

lib/bolt/applicator.rb
```diff
@@ -188,5 +192,41 @@ def apply(args, apply_body, scope)
...
+
+    def plugins
+      @plugin_tarball.value or raise Bolt::Error.new("Failed to pack module plugins: #{@plugin_tarball.reason}", 'bolt/plugin-error')
+    end
+
+    def build_plugin_tarball
+      start_time = Time.now
+      plugins = Tempfile.new('plugins')
+      sio = StringIO.new
+      output = Minitar::Output.new(Zlib::GzipWriter.new(sio))
```
MikaelSmith: This could get more complicated with Puppet 6, which appears to stop shipping minitar.

MikaelSmith:
Tests passed with rake test:git GIT_BRANCH=BOLT-640-sync-plugins-for-apply GIT_FORK=nicklewis/bolt. Going to try with a Windows controller as well.

libexec/apply_catalog.rb
```diff
@@ -39,7 +40,14 @@
 Puppet[:prerun_command] = nil
 Puppet[:postrun_command] = nil
 
-env = Puppet.lookup(:environments).get('production')
+Puppet[:default_file_terminus] = :file_server
+
+moduledir = Dir.mktmpdir
+plugins = Tempfile.new('plugins.tar.gz')
+File.binwrite(plugins, Base64.decode64(args['plugins']))
+Puppet::ModuleTool::Tar.instance.unpack(plugins, moduledir, Etc.getlogin)
```
adreyer: I thought weren't going to pass this as a task param. I guess this is fine until it causes problems. We probably need to make sure we don't use environment variables for the catalog_apply_task though.

MikaelSmith: Oh yeah, this could be a file upload, which would probably be faster for ssh.

MikaelSmith: Tempted to leave that as an improvement just to get this out faster. But I can experiment with it a bit.

adreyer: yeah I think we'll want to do something else there but lets ticket it later.

PR #534, opened 2018-07-26
(maint) Add server gemspec

MikaelSmith:
Make Gemfile depend on both gemspecs. Update gemspec to include
dependencies particular to each use-case.


MikaelSmith:
Adding this might mess with our gem release process. I'll look into how that works.

MikaelSmith:
Not going this route quite yet.

PR #535, opened 2018-07-27
(BOLT-642) Support inventory function in apply

donoghuc:
Serialize information about targets in inventory. Pass this data
to bolt_catalog and load into puppet so that get_targets etc can
be used in an apply block.

### DIFF
--- a/lib/bolt/catalog.rb
+++ b/lib/bolt/catalog.rb
@@ -51,6 +51,19 @@ module Bolt
       end
     end
 
+    def setup_inventory(inventory)
+      config = Bolt::Config.default
+      config.overwrite_transport_data(inventory['config']['transport'],
+                                      Bolt::Util.symbolize_top_level_keys(inventory['config']['transports']))
+
+      bolt_inventory = Bolt::Inventory.new(inventory['data'],
+                                           config,
+                                           Bolt::Util.symbolize_top_level_keys(inventory['target_hash']))
+
+      bolt_inventory.collect_groups
+      bolt_inventory
+    end
+
     def compile_catalog(request)
       pal_main = request['code_ast'] || request['code_string']
       target = request['target']
@@ -69,7 +82,10 @@ module Bolt
           node = Puppet.lookup(:pal_current_node)
           setup_node(node, target["trusted"])
 
-          Puppet.override(pal_main: pal_main, bolt_pdb_client: pdb_client) do
+          Puppet.override(pal_main: pal_main,
+                          bolt_pdb_client: pdb_client,
+                          bolt_inventory:
+                          setup_inventory(request['inventory'])) do
             compile_node(node)
           end
         end
--- a/lib/bolt/inventory.rb
+++ b/lib/bolt/inventory.rb
@@ -53,16 +53,16 @@ module Bolt
       inventory
     end
 
-    def initialize(data, config = nil)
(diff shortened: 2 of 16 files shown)


donoghuc:
I do not think I am on the right track with this. Pushed what I have so far to solicit help/feedback.  
Test plan:
```
plan inventory_test(TargetSpec $nodes) {
  $target = get_targets('all')
  
  $r = apply($nodes) {
  	alert($nodes)
  	alert("${get_targets('all')}")
  }
  return $target
}
```
Current output
```
bolt plan run inventory_test -n handsStarting: apply catalog on 0.0.0.0:32768, 0.0.0.0:32769
0.0.0.0:32768: Scope(Class[main]): hands
0.0.0.0:32768: Scope(Class[main]): [Target('all', {"connect-timeout"=>10, "tty"=>false, "host-key-check"=>true})]
0.0.0.0:32769: Scope(Class[main]): hands
0.0.0.0:32769: Scope(Class[main]): [Target('all', {"connect-timeout"=>10, "tty"=>false, "host-key-check"=>true})]
Finished: apply catalog with 0 failures in 2.32 sec
[
  "Target('0.0.0.0:20022', {\"connect-timeout\"=>10, \"tty\"=>false, \"host-key-check\"=>false})",
  "Target('0.0.0.0:32768', {\"connect-timeout\"=>10, \"tty\"=>false, \"host-key-check\"=>false})",
  "Target('0.0.0.0:32769', {\"connect-timeout\"=>10, \"tty\"=>false, \"host-key-check\"=>false})"
]
```
inventoryfile:
```
groups:
  - name: spec
    nodes: 
      - 0.0.0.0:20022
    config:
      transport: ssh
      ssh:
        host-key-check: false
        user: root
        password: root
  - name: hands
    nodes: 
      - 0.0.0.0:32768
      - 0.0.0.0:32769
    config:
      transport: ssh
      ssh:
        host-key-check: false
        user: root
        password: root
```

lib/bolt/catalog.rb
```diff
@@ -56,6 +56,7 @@ def compile_catalog(request)
       target = request['target']
 
       pdb_client = Bolt::PuppetDB::Client.new(Bolt::PuppetDB::Config.new(request['pdb_config']))
+      bolt_inventory = Bolt::Inventory.new(request[:inventory])
```
donoghuc: I realize this should be `request['inventory'] but it still does not help...

donoghuc:
Further manual testing revealed that it will be important to pass along config data as well. Specifically transport specific data such as "host-key-check". I think I see a path forward, I am just working out the data flow specifics.

lib/bolt/inventory.rb
```diff
@@ -60,9 +60,9 @@ def initialize(data, config = nil)
       @data = data ||= {}
       @groups = Group.new(data.merge('name' => 'all'))
       @group_lookup = {}
-      @target_vars = {}
-      @target_facts = {}
-      @target_features = {}
+      @target_vars = data['target_vars'] ||= {}
```
MikaelSmith: Passing these in data now means someone can add them in an inventory file. We should pass them as a separate hash.

lib/bolt/catalog.rb
```diff
@@ -51,6 +51,17 @@ def generate_ast(code)
       end
     end
 
+    def setup_inventory(inventory)
```
MikaelSmith: This whole function feels messy. I don't have a better idea at the moment though.

PR #536, opened 2018-07-27
(BOLT-721) Get service running

MikaelSmith:
Update the service paths to be consistent with packaged location and
binary name. Set GEM_HOME in service scripts so services use installed
gem location.

Also merge `master`.


lucywyman:
@MikaelSmith I think this might be forked from master? I'lll update the transport-service branch

MikaelSmith:
Thought I was clear in the description that it also included a merge of everything on master.

lucywyman:
:woman_facepalming: Yep, saw the list of commits and assumed. :+1:

ext/systemd/pe-bolt-server.service
```diff
@@ -6,9 +6,10 @@ After=syslog.target network.target
 Type=simple
 EnvironmentFile=-/etc/sysconfig/pe-bolt-server-service
 EnvironmentFile=-/etc/default/pe-bolt-server-service
-ExecStart=/opt/puppetlabs/puppet/bin/pe-bolt-server $PE_BOLT_SERVER_OPTIONS --foreground
+Environment=GEM_HOME=/opt/puppetlabs/server/apps/bolt-server/lib/ruby/gems/2.4.0
+ExecStart=/opt/puppetlabs/server/apps/bolt-server/bin/bolt-server $PE_BOLT_SERVER_OPTIONS --foreground
```
lucywyman: Do we want to change the var to BOLT_SERVER_OPTIONS? Minor, but also less confusing. I don't feel strongly either way.

MikaelSmith: I'm tempted to change 'pe bolt server' to 'bolt server' everywhere, but as long as the service is named pe-bolt-server I think the environment variable should stay that way too.

PR #539, opened 2018-07-31
(BOLT-191) Support uploading directories

MikaelSmith:
Implements uploading directories - i.e. files in a directory recursively - with SSH and WinRM.

### DIFF
--- a/lib/bolt/cli.rb
+++ b/lib/bolt/cli.rb
@@ -298,7 +298,7 @@ module Bolt
               if dest.nil?
                 raise Bolt::CLIError, "A destination path must be specified"
               end
-              validate_file('source file', src)
+              validate_file('source file', src, true)
               executor.upload_file(targets, src, dest, executor_opts) do |event|
                 outputter.print_event(event)
               end
@@ -392,7 +392,7 @@ module Bolt
       @pal ||= Bolt::PAL.new(config.modulepath, config.hiera_config, config.compile_concurrency)
     end
 
-    def validate_file(type, path)
+    def validate_file(type, path, allow_dir = false)
       if path.nil?
         raise Bolt::CLIError, "A #{type} must be specified"
       end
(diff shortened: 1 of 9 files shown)


MikaelSmith:
Doesn't support PCP, I imagine that will need an update to bolt_shim to deserialize a tarball containing multiple files.

MikaelSmith:
There are a couple cases where I just want to upload files and know where they went, but not care about the specific directory. Uploading a temp directory and returning the path could be useful. Bolt uses this internally a number of places.

lib/bolt/cli.rb
```diff
@@ -389,7 +389,7 @@ def pal
       @pal ||= Bolt::PAL.new(config.modulepath, config.hiera_config, config.compile_concurrency)
     end
 
-    def validate_file(type, path)
+    def validate_file(type, path, allow_dir = false)
```
adreyer: If we're using allow_dir do we need to recurse through the directory to make sure it's descendants are readable?

MikaelSmith: That makes sense to me.

adreyer:
We need to add cp -r support for the local transport and minitar support for orchestrator

MikaelSmith:
For orchestrator, it makes sense to me to update `bolt_shim::upload` to accept a new format that notes it's a tarball and where it should be expanded.

MikaelSmith:
Orchestrator supports depends on https://github.com/puppetlabs/puppetlabs-bolt_shim/pull/5. Without it, the fallback behavior will be to write the path out as a tar-gzipped file.

MikaelSmith:
This seems functional, but I think it needs some more testing, particularly with Orchestrator.

lib/bolt/transport/orch.rb
```diff
@@ -114,14 +118,50 @@ def batch_script(targets, script, arguments, options = {}, &callback)
...
+        output&.close
+      end
+
       def batch_upload(targets, source, destination, options = {}, &callback)
-        content = File.open(source, &:read)
+        stat = File.stat(source)
+        content = if stat.directory?
+                    pack(source)
+                  else
+                    File.open(source, &:read)
```
adreyer: is this different than File.read(source)

MikaelSmith: Really doesn't seem like it.

PR #540, opened 2018-07-31
(BOLT-641) Add apply_prep function for ensuring agent and gathering facts

MikaelSmith:
Adds a function that operates on a list of targets to ensure `puppet-agent` is installed, add the feature, and gather facts (including custom facts).


bolt-modules/boltlib/lib/puppet/functions/puppetize.rb
```diff
@@ -0,0 +1,42 @@
+# frozen_string_literal: true
+
+require 'fileutils'
+
+Puppet::Functions.create_function(:puppetize) do
```
adreyer: We should bikeshed over what to call this something more descriptive like apply_prep seems better to me.

My expectation is that eventually plan authors won't include this on their own it will be configured by the plan user in bolt.

bolt-modules/boltlib/lib/puppet/functions/puppetize.rb
```diff
@@ -0,0 +1,49 @@
...
+    unless applicator && executor && inventory && Puppet.features.bolt?
+      raise Puppet::ParseErrorWithIssue.from_issue_and_stack(
+        Puppet::Pops::Issues::TASK_MISSING_BOLT, action: _('puppetize')
+      )
+    end
+
+    targets = inventory.get_targets(target_spec)
+
+    # Ensure Puppet is installed
+    versions = call_function(:run_task, 'puppet_agent::version', targets)
```
MikaelSmith: If done in the `facts` plan, we can reuse this task to determine whether to sync over custom facts. We would need to rewrite (or have the applicator provide) a custom function to collect all the module code into a temp moduledir.

I also think we'd want https://github.com/puppetlabs/bolt/pull/539 along with my suggestion to be able to upload directly to a temporary directory. Without that it's much harder to handle the platform variations when trying to upload files and use them.

The fact task would also need an option to collect custom facts, with some way to pass in paths for Facter `--custom-dir` and `--external-dir`.

MikaelSmith: This function would be re-usable with Orchestrator if we pulled the custom_facts task into a module. I've initially avoided that because it's so closely tied to how we send the custom facts data and that may change.

MikaelSmith:
Currently depends on https://github.com/puppetlabs/puppetlabs-puppet_agent/pull/306.

MikaelSmith:
I think this is ready for some review. I'm working on integration tests to provide more coverage, but what's here should be decent.

spec/integration/apply_full_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+    let(:password) { conn_info('ssh')[:password] }
+    let(:apply_task) { 'apply_catalog.sh' }
+    let(:tflags) { %W[--no-host-key-check --run-as root --sudo-password #{password}] }
+
+    after(:each) do
+      uninstall = '/opt/puppetlabs/bin/puppet resource package puppet-agent ensure=absent'
+      run_cli_json(%W[command run #{uninstall}] + config_flags)
+    end
+
+    it 'installs puppet' do
```
MikaelSmith: Right now this is pretty simple, but it's something we can probably build on. I think for this PR its sufficient.

nicklewis:
Should we have a `report_function_call` in the `apply_prep` function? If we're expecting that to be internal, we can omit it, but if we expect people to use it in their plans we should include it.

MikaelSmith:
That would probably cover BOLT-639.

nicklewis:
That would approximate it, but it wouldn't cover the potentially common case of applying snippets of very basic puppet code that don't need any facts. Not sure if we care about that or not.

bolt-modules/boltlib/lib/puppet/functions/apply_prep.rb
```diff
@@ -0,0 +1,49 @@
...
+    # Gather facts, including custom facts
+    plugins = applicator.build_plugin_tarball do |mod|
+      search_dirs = []
+      search_dirs << mod.plugins if mod.plugins?
+      search_dirs << mod.pluginfacts if mod.pluginfacts?
+      search_dirs
+    end
+
+    task = applicator.custom_facts_task
+    results = executor.run_task(targets, task, 'plugins' => plugins)
```
nicklewis: I think this needs to be a `ResultSet`. And should this function return a `ResultSet`?

nicklewis:
We may want to add `without_default_logging` here and have our own `log_action` call with a more meaningful message.

bolt-modules/boltlib/lib/puppet/functions/apply_prep.rb
nicklewis: Actually, I don't understand why this isn't becoming a `ResultSet`...

nicklewis: Update: it is. The local changes I was playing with were broken. :+1:

MikaelSmith:
Did all the things.

bolt-modules/boltlib/lib/puppet/functions/apply_prep.rb
```diff
@@ -0,0 +1,62 @@
...
+          search_dirs << mod.plugins if mod.plugins?
+          search_dirs << mod.pluginfacts if mod.pluginfacts?
+          search_dirs
+        end
+
+        task = applicator.custom_facts_task
+        results = executor.run_task(targets, task, 'plugins' => plugins)
+        raise Bolt::RunFailure.new(results, 'run_task', task.name) unless results.ok?
+
+        results.each do |result|
```
MikaelSmith: What should be the return value for `apply_prep`? Currently it's returning the results of the facts task as a side-effect of updating facts in inventory.

adreyer: undef seems best since we want to deprecate it

nicklewis: I was thinking potentially just the original `TargetSpec` that was passed in. Then you could do like `apply($targets.apply_prep) { ... }`

MikaelSmith: Yeah, that makes sense to me.

PR #541, opened 2018-08-01
(BOLT-629) Validate inventory file content more carefully

nicklewis:
Previously, we didn't properly validate whether certain nested data
structures were the shape we expected. Specifically, we didn't check
whether the nodes or groups entries were arrays, or whether the
vars/facts/features/config sections were the correct type. This led to
confusing ruby error messages when we tried to treat them as though they
were correct. We now explicitly check these types and raise useful
errors accordingly.

### DIFF
--- a/lib/bolt/inventory/group.rb
+++ b/lib/bolt/inventory/group.rb
@@ -9,30 +9,54 @@ module Bolt
 
       def initialize(data)
         @logger = Logging.logger[self]
-        @name = data['name']
-        @nodes = {}
 
-        data['nodes']&.each do |n|
-          n = { 'name' => n } if n.is_a? String
-          if @nodes.include? n['name']
-            @logger.warn("Ignoring duplicate node in #{@name}: #{n}")
+        unless data.is_a?(Hash)
+          raise ValidationError.new("Expected group to be a Hash, not #{data.class}", nil)
+        end
+
+        if data.key?('name')
+          if data['name'].is_a?(String)
+            @name = data['name']
           else
-            @nodes[n['name']] = n
+            raise ValidationError.new("Group name must be a String, not #{data['name'].inspect}", nil)
+          end
+        else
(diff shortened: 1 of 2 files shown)


lib/bolt/inventory/group.rb
```diff
@@ -9,30 +9,47 @@ class Group
...
       def initialize(data)
         @logger = Logging.logger[self]
-        @name = data['name']
-        @nodes = {}
 
-        data['nodes']&.each do |n|
-          n = { 'name' => n } if n.is_a? String
-          if @nodes.include? n['name']
-            @logger.warn("Ignoring duplicate node in #{@name}: #{n}")
+        if data.key?('name')
```
adreyer: Do we need to make sure this is a hash?

nicklewis: I'm not sure! Without full schema validation, there will definitely be some scenarios that aren't covered. I was trying to think of cases where users might inadvertently enter the wrong thing (like comma-separated string for `nodes` instead of an array). So I was unclear whether the `groups` list could reasonably be misinterpreted.

I can certainly add a check for it though.

nicklewis:
Updated to check that the group entry is an object

lib/bolt/inventory/group.rb
```diff
@@ -9,30 +9,51 @@ class Group
...
+        @facts = fetch_value(data, 'facts', Hash)
+        @features = fetch_value(data, 'features', Array)
+        @config = fetch_value(data, 'config', Hash)
+
+        nodes = fetch_value(data, 'nodes', Array)
+        groups = fetch_value(data, 'groups', Array)
+
+        @nodes = {}
+        nodes.each do |node|
+          node = { 'name' => node } if node.is_a? String
```
adreyer: I think we need to assert each node is a Hash after this assignment

PR #543, opened 2018-08-01
(BOLT-653) Add configuration to the server

lucywyman:
This reads a config file from `/etc/puppetlabs/bolt-server/conf.d/bolt-server.conf`, reads the 'port' value if there is one and starts the sinatra server on that port


lib/bolt_ext/server.rb
```diff
@@ -4,9 +4,14 @@
 require 'bolt'
 require 'bolt/task'
 require 'json'
+require 'bolt_ext/server-config.rb'
 
 class TransportAPI < Sinatra::Base
+  config = TransportConfig.new
```
adreyer: How are we going to test this?

spec/bolt_ext/server_spec.rb
```diff
@@ -1,11 +1,10 @@
 # frozen_string_literal: true
 
-ENV['RACK_ENV'] = 'test'
-
 require 'bolt_spec/conn'
 require 'bolt_ext/server'
 require 'json'
 require 'rack/test'
+require 'spec_helper'
```
MikaelSmith: This should probably be first.

exe/bolt-server
```diff
@@ -1,6 +1,4 @@
-#!/usr/bin/env ruby
-# frozen_string_literal: true
+#!/opt/puppetlabs/puppet/bin/ruby
 
-require 'bolt_ext/server.rb'
-
-TransportAPI.run!
+# Used in https://gist.github.com/rennex/f379b8a53c8f2791d5838b4622ed6dd8
+`rackup --env production -D /opt/puppetlabs/server/apps/bolt-server/config.ru`
```
MikaelSmith: This is all very context-specific. I'd lean towards putting this all directly in the service file.

exe/bolt-server
```diff
@@ -1,6 +1,4 @@
-#!/usr/bin/env ruby
-# frozen_string_literal: true
+#!/opt/puppetlabs/puppet/bin/ruby
```
MikaelSmith: I think we should stick with looking up ruby via env. The gem install (or our service config) should take care of making sure it finds the right ruby.

MikaelSmith: Philosophically, I don't think this file should know about specific install paths.

lib/bolt_ext/server-config.rb
```diff
@@ -0,0 +1,45 @@
...
+  attr_accessor :port
+
+  def initialize(global = nil, local = nil)
+    # TODO: is 8144 a reasonable default port? It doesn't seem to be in use per
+    # https://confluence.puppetlabs.com/display/ENG/Choosing+a+port+for+services,
+    # and is close to other orch/pcp service ports
+    @port = 8144
+
+    global_path = global || '/etc/puppetlabs/bolt-server/conf.d/bolt-server.conf'
+    local_path = local || File.join(ENV['HOME'].to_s, ".puppetlabs", "bolt-server.conf")
```
MikaelSmith: HOME may not always be set.

MikaelSmith: I guess failure is acceptable in that case.

lib/bolt_ext/server-config.rb
```diff
@@ -0,0 +1,45 @@
...
+
+    load_config(global_path)
+    load_config(local_path)
+  end
+
+  def load_config(path = nil)
+    begin
+      conf_path = path || '/etc/puppetlabs/bolt-server/conf.d/bolt-server.conf'
```
MikaelSmith: How does this trigger reading the path/file? And it seems redundant to allow path to be nil when we only call this from initialize.

lib/bolt_ext/server-config.rb
```diff
@@ -0,0 +1,45 @@
...
+      raise("Hocon data in '#{conf_path}' failed to load.\n Error: '#{e.message}'")
+    end
+
+    unless parsed_hocon.nil?
+      @port = parsed_hocon['port'] if parsed_hocon.key?('port')
+    end
+  end
+
+  def validate
+    # TODO: should we do a check to ensure the port isn't already in use?
```
MikaelSmith: That'll error quickly enough already.

lib/bolt_ext/server-config.rb
lucywyman: Should it fail, or warn and just load the global path?

spec/bolt_ext/server_spec.rb
lucywyman: Oh, ha, I just alphabetized with abandon :sweat_smile:

lib/bolt_ext/server-config.rb
MikaelSmith: warn and just load global path

lib/bolt_ext/server-config.rb
```diff
@@ -0,0 +1,38 @@
...
+    global_path = global || '/etc/puppetlabs/bolt-server/conf.d/bolt-server.conf'
+    local_path = local || File.join(ENV['HOME'].to_s, ".puppetlabs", "bolt-server.conf")
+
+    load_config(global_path)
+    load_config(local_path)
+  end
+
+  def load_config(path)
+    begin
+      parsed_hocon = Hocon.load(path)['bolt-server']
```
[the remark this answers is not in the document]
lucywyman: Yeah, that's why I didn't catch an error :shrug:

ext/debian/pe-bolt-server.init
```diff
@@ -8,8 +8,9 @@
 # Default-Stop:      0 1 6
 ### END INIT INFO
 
-export GEM_HOME=/opt/puppetlabs/server/apps/bolt-server/lib/ruby/gems/2.4.0
-exec=/opt/puppetlabs/server/apps/bolt-server/bin/bolt-server
+prefix='/opt/puppetlabs/server/apps/bolt-server'
+export GEM_HOME="${prefix}/lib/ruby/gems/2.4.0"
+exec="${prefix}/bin/rackup --env production -D ${prefix}/config.ru"
```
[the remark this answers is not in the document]
lucywyman: And have bolt-server call rackup?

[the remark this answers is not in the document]
MikaelSmith: Yeah.

exe/bolt-server
```diff
@@ -1,6 +1,5 @@
 #!/usr/bin/env ruby
-# frozen_string_literal: true
 
-require 'bolt_ext/server.rb'
-
-TransportAPI.run!
+# ARGV[0] is the path to config.ru
+# The rackup file that configures the bolt-server
+`rackup --env production -D #{ARGV[0]}`
```
MikaelSmith: This would be cleaner as `exec 'rackup', '--env', 'production', '-D', ARGV[0]`. That will avoid an extra process that's not needed.

ext/debian/pe-bolt-server.init
```diff
@@ -8,8 +8,9 @@
 # Default-Stop:      0 1 6
 ### END INIT INFO
 
-export GEM_HOME=/opt/puppetlabs/server/apps/bolt-server/lib/ruby/gems/2.4.0
-exec=/opt/puppetlabs/server/apps/bolt-server/bin/bolt-server
+prefix='/opt/puppetlabs/server/apps/bolt-server'
+export GEM_HOME="${prefix}/lib/ruby/gems/2.4.0"
+exec="${prefix}/bin/bolt-server ${prefix}/config.ru"
```
MikaelSmith: I think the service files will all need to add `#{prefix}/bin` to PATH so bolt-server can find rackup.

PR #545, opened 2018-08-02
(BOLT-641) Use consistent HOME dir when testing Bolt

MikaelSmith:
Previously we used two different HOME locations when running Bolt on
Windows. This was to provide path to modules that we could put on the
command-line that both Cygwin and Bolt would understand. However with
the move to using Bolt to install from a Puppetfile to Boltdir, we need
to use a single location for Boltdir. Update test helpers to use the
Windows paths to the default Boltdir.

Also, always ensure the default_boltdir exists before we try to copy to
it.

### DIFF
--- a/bolt-modules/boltlib/spec/functions/apply_prep_spec.rb
+++ b/bolt-modules/boltlib/spec/functions/apply_prep_spec.rb
@@ -76,6 +76,13 @@ describe 'apply_prep' do
       end
     end
 
+    it 'fails if version task is not found' do
+      Puppet::Pal::ScriptCompiler.any_instance.expects(:task_signature).with('puppet_agent::version')
+      is_expected.to run.with_params(hostnames).and_raise_error(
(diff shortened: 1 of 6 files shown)


MikaelSmith:
Verified testing with a Windows controller - `BOLT_CONTROLLER=windows10ent-64` - works again.

bolt-modules/boltlib/spec/functions/apply_prep_spec.rb
```diff
@@ -76,6 +76,13 @@
       end
     end
 
+    it 'fails if version task is not found' do
```
adreyer: We expect this to happen because users may have used their own puppet_agent package?

MikaelSmith: Yes. It's a flaw we'll need to work to address.

PR #546, opened 2018-08-03
(BOLT-735) Avoid Etc.getlogin

MikaelSmith:
According to Ruby's docs at
https://ruby-doc.org/stdlib-2.4.0/libdoc/etc/rdoc/Etc.html#method-c-getlogin
we should avoid `getlogin` for security-related purposes and try
`getpwuid` if it fails. `getlogin` returns `nil` in containers under
Docker for Mac. Use `getwpuid.name` instead to get the current user name
when deciding what user should own the files that we untar.


MikaelSmith:
Not sure there's much we can do to automatically test this other than make sure it continues to work the places it worked before.

PR #549, opened 2018-08-06
(BOLT-717) Add BoltSpec::Run public helper

adreyer:
This adds a new set of BoltSpec spec helpers that can be used to run
tasks, commands, and plans from tests without handling cli flags or
running in a separate process.

### DIFF
--- a/lib/bolt/catalog.rb
+++ b/lib/bolt/catalog.rb
@@ -56,12 +56,9 @@ module Bolt
       config.overwrite_transport_data(inventory['config']['transport'],
                                       Bolt::Util.symbolize_top_level_keys(inventory['config']['transports']))
 
-      bolt_inventory = Bolt::Inventory.new(inventory['data'],
-                                           config,
-                                           Bolt::Util.symbolize_top_level_keys(inventory['target_hash']))
-
-      bolt_inventory.collect_groups
-      bolt_inventory
+      Bolt::Inventory.new(inventory['data'],
+                          config,
+                          Bolt::Util.symbolize_top_level_keys(inventory['target_hash']))
     end
 
     def compile_catalog(request)
--- a/lib/bolt/inventory.rb
+++ b/lib/bolt/inventory.rb
@@ -49,7 +49,6 @@ module Bolt
 
       inventory = new(data, config)
       inventory.validate
-      inventory.collect_groups
       inventory
     end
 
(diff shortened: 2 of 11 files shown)


lib/bolt/inventory.rb
```diff
@@ -63,6 +62,7 @@ def initialize(data, config = nil, target_vars: {}, target_facts: {}, target_fea
       @target_vars = target_vars
       @target_facts = target_facts
       @target_features = target_features
+      collect_groups
```
donoghuc: Should we also remove `collect_groups` call from `setup_inventory` in catalog.rb? Used to build up new inventory from JSON passed to bolt_catalog.  https://github.com/puppetlabs/bolt/blob/a21403650140b54a9b28953170c97e0246065be4/lib/bolt/catalog.rb#L54-L65

adreyer: yup fixed

lib/bolt/catalog.rb
```diff
@@ -60,7 +60,6 @@ def setup_inventory(inventory)
                                            config,
                                            Bolt::Util.symbolize_top_level_keys(inventory['target_hash']))
 
-      bolt_inventory.collect_groups
       bolt_inventory
```
MikaelSmith: Could get rid of the temporary variable.

lib/bolt/pal.rb
```diff
@@ -97,6 +97,8 @@ def in_bolt_compiler
           alias_types(compiler)
           begin
             yield compiler
+          rescue Bolt::Error => err
```
MikaelSmith: How did this come up?

lib/bolt_spec/run.rb
```diff
@@ -0,0 +1,93 @@
+# frozen_string_literal: true
+
+# shorthand for requiring the parts of bolt we need
+require 'bolt/cli'
+require 'bolt/util'
+
+# This is intended to provide a relatively stable method of executing bolt in process from tests.
+# Currently it provdides run_task, run_plan and run_command helpers.
```
MikaelSmith: Typo in `provdides`.

lib/bolt/pal.rb
MikaelSmith: Running tasks directly can result in a Bolt::Error, we should be throwing it. Previously this primarily handled plans and our external task validation made sure we wouldn't run into an error.

MikaelSmith:
Merge when passing.

PR #550, opened 2018-08-06
(BOLT-745) Include boltlib in system loader

MikaelSmith:
Makes the intention clearer.

### DIFF
--- a/lib/bolt/catalog/loaders.rb
+++ b/lib/bolt/catalog/loaders.rb
@@ -1,0 +1,19 @@
+# frozen_string_literal: true
+
+module Bolt
+  class Catalog
+    class BoltLoaders < Puppet::Pops::Loaders
+      def create_puppet_system_loader
+        parent = super
+
+        Puppet::Pops::Loader::ModuleLoaders::FileBased.new(
+          parent,
+          self,
+          'boltlib',
+          File.join(__dir__, '../../../bolt-modules/boltlib'),
+          'boltlib_system'
+        )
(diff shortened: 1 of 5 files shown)


lib/bolt/catalog/loaders.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+module Puppet
```
adreyer: Can we make the file match the class path by moving the class or the file?

MikaelSmith: This is following the pattern you established in catalog/compiler.rb

MikaelSmith: So not sure what you're asking me to do.

adreyer: I expect the compiler class to go away when we switch to PAL. It's in Puppet because it was a pain to namespace stuff otherwise.
If this would require excesssive namespacing it should probably go int puppet/pops/bolt_loaders otherwise we can just move it to Bolt::Catalog::Loaders

MikaelSmith: I'll try it out with Bolt::Catalog::Loaders.

MikaelSmith: `Bolt::Catalog` is a class, so it looks a little funny, but it works.

MikaelSmith: Updated.

PR #551, opened 2018-08-07
(BOLT-653) Ensure rack server starts without rackup on PATH

MikaelSmith:
Ensure we start the rack server. This is safer than exec because with
exec we need to make sure we get the rackup application that points to
the correct Ruby.


MikaelSmith:
I think this avoids all the pathing issues, and ensures we run with the right Ruby interpreter.

PR #552, opened 2018-08-07
(maint) merge in master

lucywyman:


MikaelSmith:
I haven't seen that failure before.

MikaelSmith:
The failing tests seem to be returning more results than expected. I'd suggest dropping a puts to view the full results, and maybe also `@log_output.read`.

MikaelSmith:
Oh duh, new call to `Tasks.new`.

lucywyman:
Not in master I guess? https://github.com/puppetlabs/bolt/blob/master/lib/bolt/applicator.rb#L44-L52 https://github.com/puppetlabs/bolt/blob/transport-service/lib/bolt/applicator.rb#L35

PR #555, opened 2018-08-08
(BOLT-757) Include custom facts and features in apply

MikaelSmith:
Ensure external facts are copied to targets during apply, that the apply
sets up Facter to find them, and that the lib dirs of modules are
available when resolving `puppet/feature` with the autoloader.

Also make it easier to debug `apply_catalog` by accepting a file name on
the CLI as an alternative to passing args via stdin.


libexec/apply_catalog.rb
```diff
@@ -50,6 +50,13 @@
   end
 
   env = Puppet.lookup(:environments).get('production').override_with(modulepath: [moduledir])
+  # Needed to ensure features are loaded
+  env.each_plugin_directory do |dir|
+    $LOAD_PATH << dir unless $LOAD_PATH.include?(dir)
```
adreyer: This would have broken providers with non-relative requires too right?

MikaelSmith: Probably.

MikaelSmith: Extended the tests to include require from puppet_x.

MikaelSmith: But acceptance tests may not be working yet.

MikaelSmith: Now passes.

MikaelSmith: Tested with both Windows and Debian controllers.

PR #556, opened 2018-08-08
(BOLT-654) Add SSL configuration for bolt-server

lucywyman:
This adds 3 keys to the bolt-server configuration:
* `ssl-cert`: the path to the ssl cert for bolt-server to use
* `ssl-key`: the path to the ssl key for bolt-server to use
* `ssl-ca-cert`: the path to the ssl ca certificat for bolt-server to use

It validates that each of these keys is present and a readable file, then configures Puma to bind to the specified port over SSL using these credentials.

### DIFF
--- a/config.ru
+++ b/config.ru
@@ -28,4 +28,13 @@ configure do
   # end
 end
 
-Rack::Handler::Puma.run(TransportAPI, Port: config.port)
+host = "0.0.0.0"
+bind_addr = "ssl://#{host}?"
+bind_addr << "cert=#{config.ssl_cert}"
+bind_addr << "&key=#{config.ssl_key}"
+bind_addr << "&ca=#{config.ssl_ca_cert}"
+bind_addr << "&verify_mode=force_peer"
+
+Rack::Handler::Puma.run(TransportAPI,
+                        Port: config.port,
+                        Host: bind_addr)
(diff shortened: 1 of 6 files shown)


config.ru
```diff
@@ -28,4 +28,12 @@ configure do
   # end
 end
 
-Rack::Handler::Puma.run(TransportAPI, Port: config.port)
+host = "0.0.0.0"
+bind_addr = "ssl://#{host}?"
```
MikaelSmith: These don't go in the configure block?

MikaelSmith:
Tests are failing in Travis. Overall looks good though.

ext/config/bolt-server.conf
```diff
@@ -0,0 +1,6 @@
+bolt-server: {
+    port: 8144
+    ssl-cert: cert.pem
```
adreyer: Are these useful defaults?

config.ru
lucywyman: Yeah :(  https://github.com/puma/puma#binding-tcp--sockets for that, and https://github.com/puma/puma/blob/master/lib/puma/dsl.rb for all config options (of which ssl is not there).

ext/config/bolt-server.conf
adreyer: yeah I was wondering they should be commented out rather then pointing to certs we dont' expect to exist. If orch does this it's fine for consistency.

lucywyman: :+1: https://github.com/puppetlabs/orchestrator/blob/johnson/resources/ext/config/conf.d/orchestrator.conf is what I'm copying just for full disclosure!

PR #561, opened 2018-08-08
(BOLT-767) Add logging messages for plans

nmaludy:
Current plans do not have a `Starting: plan xxx` and `Finished: plan xxx in SS.ss sec`.

This PR adds those logging messages into plan executions.

A timestamp is returned from the `log_start_plan()` function so that nested plans produce accurate time measurements.

### DIFF
--- a/lib/bolt/executor.rb
+++ b/lib/bolt/executor.rb
@@ -134,16 +134,20 @@ module Bolt
       results
     end
 
-    def log_start_plan(plan_name)
+    def log_plan(plan_name)
       log_method = @plan_logging ? :notice : :info
       @logger.send(log_method, "Starting: plan #{plan_name}")
-      Time.now
-    end
+      start_time = Time.now
 
-    def log_finish_plan(plan_name, start_time)
-      log_method = @plan_logging ? :notice : :info
-      duration = Time.now - start_time
-      @logger.send(log_method, "Finished: plan #{plan_name} in #{duration.round(2)} sec")
+      results = nil
+      begin
+        results = yield
+      ensure
+        duration = Time.now - start_time
+        @logger.send(log_method, "Finished: plan #{plan_name} in #{duration.round(2)} sec")
+      end
+
+      results
     end
 
     def report_transport(transport, count)
--- a/lib/bolt/executor.rb
+++ b/lib/bolt/executor.rb
@@ -134,6 +134,18 @@ module Bolt
       results
     end
 
+    def log_start_plan(plan_name)
+      log_method = @plan_logging ? :notice : :info
+      @logger.send(log_method, "Starting: plan #{plan_name}")
+      Time.now
+    end
+
+    def log_finish_plan(plan_name, start_time)
+      log_method = @plan_logging ? :notice : :info
+      duration = Time.now - start_time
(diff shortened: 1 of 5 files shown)


lib/bolt/executor.rb
```diff
@@ -134,6 +134,18 @@ def log_action(description, targets)
       results
     end
 
+    def log_start_plan(plan_name)
```
MikaelSmith: A pattern we used in `log_action` was to take a block and do the logging around yielding to that block. You could still use an `ensure` section to make sure the finished message is logged.

I'd be in favor of that here. And if need be splitting some of the run_plan function out into other functions.

MikaelSmith: Otherwise this seems to follow the semantics we've defined for other logging, so 👍

nmaludy: Great idea!

I saw the other functions `start_plan` and `finish_plan` and flipped the coin.  I'll change to block semantics and resubmit. 

Thanks for the feedback!

MikaelSmith: Yeah, those could've followed a similar pattern. I may fix that up at some point.

MikaelSmith:
The output of this for a simple plan is
```
Starting: plan profiles::simple
Starting: install puppet and gather facts on k9i68z96wg1gl6t.delivery.puppetlabs.net
Finished: install puppet and gather facts with 0 failures in 21.47 sec
Starting: apply catalog on k9i68z96wg1gl6t.delivery.puppetlabs.net
Finished: apply catalog with 0 failures in 5.71 sec
Finished: plan profiles::simple in 27.18 sec
```

I do think that adds something even for the top-level plan you asked it to run: documents the name of the plan (if you're not invoking Bolt directly) and provides execution time for the whole thing. We get runtime with - for example - `bolt task run`:
```
Started on k9i68z96wg1gl6t.delivery.puppetlabs.net...
Finished on k9i68z96wg1gl6t.delivery.puppetlabs.net:
  {
    "version": "5.5.4",
    "source": "/opt/puppetlabs/puppet/VERSION"
  }
Successful on 1 node: k9i68z96wg1gl6t.delivery.puppetlabs.net
Ran on 1 node in 1.23 seconds
```

MikaelSmith:
I do kind of want indentation now. But that's better to tackle separately.

MikaelSmith:
The one thing that printing for the top-level plan prohibits is suppressing all logging in a plan with `without_default_logging`. I lean towards putting the top-level messages at a lower logging level (so you have to use --verbose). That could be accomplished by passing `named_args['_bolt_api_call']` to `log_plan` as an additional argument force `:info` level when true.

MikaelSmith:
I do understand that. I'm partial to adding a `--quiet` option eventually. But I think I'm comfortable with the additional logging as-is.

nmaludy:
@MikaelSmith I also agree about indentation! I've been trying to think more about how we can beautify the CLI output to make the end-user experience much more pleasant.

In reality, the user is going to be interacting with the CLI quite a bit and the output should be a pleasure to read. 

Some thoughts i had:

* Indentation might be nice for different plan layers. This may become too hard to read if you get down very deep into tested plans (make it an option maybe).

* Separate the human-readable lines with some whitespace or  visual cues (repeated characters) to break up the dense output

* Other random thoughts :)

PR #564, opened 2018-08-09
(BOLT-768) Fix acceptance tests without Beaker puppet install

MikaelSmith:
beaker-puppet's install helpers also setup puppet to be found when
running the command. We're not using that so we can test apply. However
that means any use of puppet needs to use the full path, which is easy
to do because we're only using it on non-Windows tests.

Also remove test that's adequately covered by winrm transport tests and
relied on Puppet's ruby.


MikaelSmith:
Passed my acceptance runs.

PR #569, opened 2018-08-13
(PE-24271) Run pe-bolt-server service as service user

lucywyman:
This adds the pe-bolt-server user and group as the owners of the pe-bolt-server service


lucywyman:
Currently testing that this works

lucywyman:
I tested on Ubuntu 16.04 and it worked. Planning to try centos next.

MikaelSmith:
Since we dropped Ubuntu 14.04 as a server platform, we'll never actually use the debian-specific init script.

lucywyman:
Ohhh, I see.

PR #570, opened 2018-08-14
(BOLT-772) Allow failure when removing Bolt Apply temp directory

MikaelSmith:
Some providers keep handle open to files in their module after running.
On Windows this prevents deleting the temporary modulepath we create
when applying a catalog. Allow this to fail and return a message over
stderr (which will be logged at debug level).

### DIFF
--- a/libexec/apply_catalog.rb
+++ b/libexec/apply_catalog.rb
@@ -1,6 +1,7 @@
 #! /opt/puppetlabs/puppet/bin/ruby
 # frozen_string_literal: true
 
+require 'fileutils'
 require 'json'
 require 'puppet'
 require 'puppet/configurer'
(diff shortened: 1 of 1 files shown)


MikaelSmith:
When we start caching module code on targets, this will probably be a non-issue.

libexec/apply_catalog.rb
```diff
@@ -42,39 +43,43 @@
...
+  catalog.environment_instance = env
+
+  configurer = Puppet::Configurer.new
+  configurer.run(catalog: catalog, report: report, pluginsync: false)
+end
+
+puts JSON.pretty_generate(report.to_data_hash)
+exit_code = report.exit_status != 1
+
+begin
```
adreyer: Should this be in a finally?

MikaelSmith: That would make sense...

libexec/apply_catalog.rb
```diff
@@ -43,7 +44,8 @@
 Puppet[:default_file_terminus] = :file_server
 
 exit_code = 0
-Dir.mktmpdir do |moduledir|
+begin
+  moduledir = Dir.mktmpdir
```
adreyer: Should this be outside the begin so we get the 'proper' error when Dirmktmpdir fails?

MikaelSmith: Well, we shouldn't try to remove it if we don't create it in the first place.

PR #572, opened 2018-08-15
(BOLT-722) Don't load ssh config if using bolt via api

lucywyman:
Ensure we don't read from the users system SSH config or the global Bolt config when running bolt via the API

### DIFF
--- a/lib/bolt/executor.rb
+++ b/lib/bolt/executor.rb
@@ -18,14 +18,18 @@ module Bolt
     attr_reader :noop, :transports
     attr_accessor :run_as
 
+    # FIXME: There must be a better way
+    # https://makandracards.com/makandra/36011-ruby-do-not-mix-optional-and-keyword-arguments
     def initialize(concurrency = 1,
                    analytics = Bolt::Analytics::NoopClient.new,
                    noop = nil,
-                   bundled_content: nil)
+                   bundled_content: nil,
+                   load_config: true)
       @analytics = analytics
       @bundled_content = bundled_content
       @logger = Logging.logger[self]
       @plan_logging = false
+      @load_config = load_config
 
       @transports = Bolt::TRANSPORTS.each_with_object({}) do |(key, val), coll|
         coll[key.to_s] = Concurrent::Delay.new do
@@ -212,6 +216,7 @@ module Bolt
       log_action(description, targets) do
         notify = proc { |event| @notifier.notify(callback, event) if callback }
         options = { '_run_as' => run_as }.merge(options) if run_as
+        options = options.merge('_load_config' => @load_config)
         arguments['_task'] = task.name
 
         results = batch_execute(targets) do |transport, batch|
--- a/spec/bolt/transport/ssh_spec.rb
+++ b/spec/bolt/transport/ssh_spec.rb
@@ -31,6 +31,7 @@ describe Bolt::Transport::SSH do
   let(:command) { "pwd" }
   let(:config) { mk_config(user: user, password: password) }
   let(:no_host_key_check) { mk_config('host-key-check' => false, user: user, password: password) }
+  let(:no_user_config) { mk_config('host-key-check' => false, user: nil, password: password) }
   let(:ssh) { Bolt::Transport::SSH.new }
   let(:echo_script) { <<BASH }
 for var in "$@"
@@ -141,6 +142,24 @@ BASH
         }.to raise_error(Bolt::Node::ConnectError)
         expect(Time.now - exec_time).to be > 2
       end
+    end
+
+    it "uses Net::SSH config when no user is specified" do
+      expect(Net::SSH::Config)
+        .to receive(:for)
+        .at_least(:once)
+        .with(hostname, any_args)
+        .and_return(user: user)
+
+      ssh.with_connection(make_target(conf: no_user_config)) {}
+    end
+
+    it "doesn't read system config if load_config is false" do
+      allow(Etc).to receive(:getlogin).and_return('bolt')
+      expect(Net::SSH::Config).not_to receive(:for)
+
+      config_user = ssh.with_connection(make_target(conf: no_user_config), false, &:user)
+      expect(config_user).to be('bolt')
     end
   end
 
--- a/lib/bolt/transport/ssh.rb
+++ b/lib/bolt/transport/ssh.rb
@@ -60,8 +60,8 @@ module Bolt
         end
       end
 
-      def with_connection(target)
-        conn = Connection.new(target)
+      def with_connection(target, load_config = true)
+        conn = Connection.new(target, load_config)
         conn.connect
         yield conn
       ensure
@@ -125,7 +125,7 @@ module Bolt
 
       def run_task(target, task, arguments, options = {})
         input_method = task.input_method || "both"
-        with_connection(target) do |conn|
+        with_connection(target, options.fetch('_load_config', true)) do |conn|
           conn.running_as(options['_run_as']) do
             stdin, output = nil
 
--- a/lib/bolt/transport/ssh/connection.rb
+++ b/lib/bolt/transport/ssh/connection.rb
@@ -54,10 +54,12 @@ module Bolt
         attr_reader :logger, :user, :target
         attr_writer :run_as
 
-        def initialize(target)
+        def initialize(target, load_config = true)
           @target = target
+          @load_config = load_config
 
-          @user = @target.user || Net::SSH::Config.for(target.host)[:user] || Etc.getlogin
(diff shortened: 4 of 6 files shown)


acceptance/setup/pbs_package/pre-suite/072_install_puppet_agent.rb
```diff
@@ -0,0 +1,24 @@
...
+test_name "Install puppet agent" do
+  step "on the bolt controller" do
+    install_puppetlabs_release_repo_on(bolt, "puppet6-nightly")
+    bolt.install_package('puppet-agent')
+    result = on(bolt, "puppet agent --version")
+    assert_match(/6.0/, result.output)
+  end
+
+  step "restart puppet-agent service to generate certs" do
+    on bolt, puppet("agent -t"), :acceptable_exit_codes => [0,1]
```
lucywyman: I'm really struggling to get the certificates working here. This is the output from `ls -R`:
```
    /etc/puppetlabs/puppet/ssl/ca/private:
    ca.pass
    
    /etc/puppetlabs/puppet/ssl/ca/requests:
    
    /etc/puppetlabs/puppet/ssl/ca/signed:
    
    /etc/puppetlabs/puppet/ssl/certificate_requests:
    gpdwbp3mpywyt9b.delivery.puppetlabs.net.pem
    
    /etc/puppetlabs/puppet/ssl/certs:
    ca.pem
    
    /etc/puppetlabs/puppet/ssl/private:
    
    /etc/puppetlabs/puppet/ssl/private_keys:
    gpdwbp3mpywyt9b.delivery.puppetlabs.net.pem
    
    /etc/puppetlabs/puppet/ssl/public_keys:
    gpdwbp3mpywyt9b.delivery.puppetlabs.net.pem
```

But when I go to sign the certificate it says there's no request.

lucywyman:
To manually test:
1. SSH into slice box with provided credentials
2. Run `curl -X POST -H "Content-Type: application/json" -d @echo.json -E /etc/puppetlabs/bolt-server/ssl/10-32-167-150.rfc1918.puppetlabs.net.cert.pem --key /etc/puppetlabs/bolt-server/ssl/10-32-167-150.rfc1918.puppetlabs.net.key.pem -k https://0.0.0.0:8144/ssh/run_task | jq` to run the 'echo' task
3. Feel free to add your own json files to run tasks. Just be sure to base64 encode task file content and follow the structure of `echo.json` - we haven't yet updated the API to reflect the docs.
4. The SSH configs we read are `Net::SSH::Config.for`, which reads "default_files" which are:
```
githubs/bolt/vendor/bundle/ruby/2.3.0/gems/net-ssh-4.2.0/lib/net/ssh/config.rb
42:      @@default_files = %w(~/.ssh/config /etc/ssh_config /etc/ssh/ssh_config)
```
5. Modify these files and send a new request to pe-bolt-server to ensure they are not getting read.

config.ru
```diff
@@ -30,9 +30,9 @@ end
 
 host = "0.0.0.0"
 bind_addr = "ssl://#{host}?"
-bind_addr << "cert=#{config.ssl_cert}"
-bind_addr << "&key=#{config.ssl_key}"
-bind_addr << "&ca=#{config.ssl_ca_cert}"
+bind_addr << "cert=#{config.ssl_cert}" if config.ssl_cert
```
MikaelSmith: What happens if we don't pass in certs?

lib/bolt/executor.rb
```diff
@@ -18,14 +18,18 @@ class Executor
     attr_reader :noop, :transports
     attr_accessor :run_as
 
+    # FIXME: There must be a better way
```
MikaelSmith: Probably make them all keyword arguments.

lib/bolt/executor.rb
```diff
@@ -212,6 +216,7 @@ def run_task(targets, task, arguments, options = {}, &callback)
       log_action(description, targets) do
         notify = proc { |event| @notifier.notify(callback, event) if callback }
         options = { '_run_as' => run_as }.merge(options) if run_as
+        options = { '_load_config' => @load_config }.merge(options) if @load_config
```
MikaelSmith: https://github.com/puppetlabs/bolt/blob/master/lib/bolt/applicator.rb#L198 uses `batch_task` directly. If you make `_load_config` an opt-in check, then we also need to pass it there or `apply` won't use SSH config. Alternatively we could make it an opt-in by default (true unless passed false). I prefer opt-in by default, because I think it makes sense in most use-cases.

config.ru
lucywyman: I think this is leftover from when I wasn't sure if we'd need to pass in a CA cert. Puma SSL doesn't accept self-signed certs with no CA though, so we do need to provide all these. A validation error is raised well before we get here. I'll remove this bit.

lib/bolt/executor.rb
lucywyman: Since this requires changes unrelated to this pr in other parts of bolt I figured it'd be a different, maybe maint PR? But I can make those changes here - we only invoke the executor in a handful of places.

MikaelSmith: A separate maint PR, if you feel like addressing it, is fine.

lib/bolt/executor.rb
lucywyman: In `with_connection` it does default to true. Will that get overwritten with 'nil' if it's passed?  I think my false assumption here was that if passed 'nil' it'd set the default, but does it actually get nil which is falsey?

MikaelSmith: I think it gets nil.

MikaelSmith: I might've been misreading the whole thing. I think what's here is correct, but probably has more checks than it needs.

lucywyman: (shrug) Better safe than sorry.

lib/bolt/executor.rb
```diff
@@ -18,14 +18,18 @@ class Executor
...
                    analytics = Bolt::Analytics::NoopClient.new,
                    noop = nil,
-                   bundled_content: nil)
+                   bundled_content: nil,
+                   load_config: true)
       @analytics = analytics
       @bundled_content = bundled_content
       @logger = Logging.logger[self]
       @plan_logging = false
+      @load_config = load_config || true
```
MikaelSmith: Will this ever be false? `false || true` will be `true`. I think this can just be `@load_config = load_config`.

lib/bolt/executor.rb
```diff
@@ -212,6 +216,7 @@ def run_task(targets, task, arguments, options = {}, &callback)
       log_action(description, targets) do
         notify = proc { |event| @notifier.notify(callback, event) if callback }
         options = { '_run_as' => run_as }.merge(options) if run_as
+        options = { '_load_config' => @load_config }.merge(options)
```
MikaelSmith: This should be `options = options.merge { '_load_config' => @load_config }` to prevent the caller overriding the executor.

lib/bolt/executor.rb
```diff
@@ -18,14 +18,18 @@ class Executor
...
     attr_accessor :run_as
 
+    # FIXME: There must be a better way
+    # https://makandracards.com/makandra/36011-ruby-do-not-mix-optional-and-keyword-arguments
     def initialize(concurrency = 1,
                    analytics = Bolt::Analytics::NoopClient.new,
                    noop = nil,
-                   bundled_content: nil)
+                   bundled_content: nil,
+                   load_config: true)
```
MikaelSmith: Is there a test we can add around running a task with executor configured with `load_config: false`? Maybe expect that some Net::SSH setup isn't called?

lucywyman: +1. I added a unit test to ssh_spec to verify this works! It doesn't ensure the var gets passed from executor, but it does ensure that the Net::SSH config isn't used.

spec/bolt/transport/ssh_spec.rb
```diff
@@ -142,6 +142,18 @@ def result_value(stdout = nil, stderr = nil, exit_code = 0)
         expect(Time.now - exec_time).to be > 2
       end
     end
+
+    it "doesn't read system config if load_config is specified" do
+      no_user_conf = mk_config(user: nil, password: password, 'host-key-check' => false)
+      allow(Net::SSH::Config)
```
MikaelSmith: Why is this part of the test?

MikaelSmith: Should this be `expect`, asserting that we do use `Net::SSH::Config` by default? Then another test that sets up an executor that doesn't load it with an `expect(...).never`?

PR #576, opened 2018-08-16
(PE-24721) Correctly daemonize

MikaelSmith:
Pass daemonize option to the correct handler. Also ensure pidfile is configured. Update services to match.

Put pidfile and logs in subdirectories so pe-bolt-server can be made owner of them.

### DIFF
--- a/config.ru
+++ b/config.ru
@@ -35,6 +35,10 @@ bind_addr << "&key=#{config.ssl_key}"
 bind_addr << "&ca=#{config.ssl_ca_cert}"
 bind_addr << "&verify_mode=force_peer"
 
+# This doesn't seem ideal, but it behaves well. I couldn't figure out how to get Puma or Rack to do
+# it for me. An alternative might be rewriting this to use 'puma -C config.rb'.
+Process.daemon
 Rack::Handler::Puma.run(TransportAPI,
+                        pidfile: '/var/run/puppetlabs/bolt-server/bolt-server.pid',
                         Port: config.port,
                         Host: bind_addr)
--- a/exe/bolt-server
+++ b/exe/bolt-server
@@ -4,4 +4,4 @@
 # ARGV[0] is the path to config.ru, the rackup file that configures the bolt-server
 
 require 'rack'
-Rack::Server.start(environment: 'production', daemonize: true, config: ARGV[0])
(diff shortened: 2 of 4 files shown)


exe/bolt-server
```diff
@@ -4,4 +4,4 @@
 # ARGV[0] is the path to config.ru, the rackup file that configures the bolt-server
 
 require 'rack'
-Rack::Server.start(environment: 'production', daemonize: true, config: ARGV[0])
```
MikaelSmith: These args seemed to be completely ignored, I'm not sure why.

config.ru
```diff
@@ -35,6 +35,10 @@ bind_addr << "&key=#{config.ssl_key}"
 bind_addr << "&ca=#{config.ssl_ca_cert}"
 bind_addr << "&verify_mode=force_peer"
 
+# This doesn't seem ideal, but it behaves well. I couldn't figure out how to get Puma or Rack to do
+# it for me. An alternative might be rewriting this to use 'puma -C config.rb'.
+Process.daemon
 Rack::Handler::Puma.run(TransportAPI,
+                        pidfile: '/var/run/puppetlabs/bolt-server/bolt-server.pid',
```
MikaelSmith: This works, even though it's not documented anywhere I can find.

ext/redhat/pe-bolt-server.init
```diff
@@ -15,14 +15,14 @@
 prefix='/opt/puppetlabs/server/apps/bolt-server'
 export GEM_HOME="${prefix}/lib/ruby/gems/2.4.0"
 exec="${prefix}/bin/bolt-server"
-prog="pe-bolt-server"
+prog="bolt-server"
```
MikaelSmith: `prog` was only used for lockfile/piddir/logdir, so I changed it to be consistent with packaging.

ext/systemd/pe-bolt-server.service
```diff
@@ -3,7 +3,8 @@ Description=PE Bolt Server
 After=syslog.target network.target
 
 [Service]
-Type=simple
+Type=fork
```
MikaelSmith: The init service script expects a daemonized service that writes a pidfile. Updated systemd to expect the same thing for consistency.

PR #580, opened 2018-08-20
(BOLT-788) Update bolt server to accept documented API

lucywyman:
The bolt-server expected API and the documented API have gotten out of sync.
This updates the bolt server to expect the structure in the documented api, specifically moving the 'options' keys out a level and removing 'options' as a key.

### DIFF
--- a/spec/bolt_ext/server_spec.rb
+++ b/spec/bolt_ext/server_spec.rb
@@ -25,22 +25,22 @@ echo $PT_message
 TASK
 
       body = {
-        'task' => {
-          'name' => 'echo',
-          'metadata' => {
-            'description' => 'Echo a message',
-            'parameters' => { 'message' => 'Hello world!' }
+        'task': {
+          'name': 'echo',
+          'metadata': {
+            'description': 'Echo a message',
+            'parameters': { 'message': 'Default message' }
           },
-          'file_content' => Base64.encode64(impl)
+          'file_content': Base64.encode64(impl)
         },
-        'target' => {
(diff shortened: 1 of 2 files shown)


spec/bolt_ext/server_spec.rb
```diff
@@ -38,7 +38,7 @@ def app
           'user' => target[:user],
           'password' => target[:password],
           'port' => target[:port],
-          'options' => { "host-key-check" => "false" }
```
donoghuc: Is travis still failing with this fix?

lucywyman: Yes, though it's passing locally for me....

spec/bolt_ext/server_spec.rb
```diff
@@ -25,22 +25,22 @@ def app
 TASK
 
       body = {
-        'task' => {
-          'name' => 'echo',
-          'metadata' => {
-            'description' => 'Echo a message',
-            'parameters' => { 'message' => 'Hello world!' }
+        'task': {
```
MikaelSmith: That seems weird, why do the keys need to be symbols?

lucywyman: They don't, it was just an aesthetic change.

lucywyman: I think the travis failures were transient? My last commit just added a print statement to figure out why it was passing locally but not in travis, and then travis started passing :woman_shrugging:

PR #581, opened 2018-08-20
(maint) Fork -> forking typo in service

lucywyman:


MikaelSmith:
It seems that at least on CentOS 7 the incorrect `Type` gets ignored and the service works because the `PIDFile` is correct. This fix gets rid of the warning and the service still works.

PR #582, opened 2018-08-20
(BOLT-685) Avoid setting `level=` on log level repeatedly

MikaelSmith:
I believe there's a race condition that can be triggered in the Logging
library when creating many loggers and setting `level=` concurrently.
Filed as https://github.com/TwP/logging/issues/200.

This fix avoids the race by setting up the shared transport logger once
and passing it to connection instances.

### DIFF
--- a/lib/bolt/transport/winrm.rb
+++ b/lib/bolt/transport/winrm.rb
@@ -37,10 +37,13 @@ module Bolt
         super
         require 'winrm'
         require 'winrm-fs'
+
+        @transport_logger = Logging.logger[::WinRM]
+        @transport_logger.level = :warn
       end
 
       def with_connection(target)
-        conn = Connection.new(target)
+        conn = Connection.new(target, @transport_logger)
         conn.connect
         yield conn
       ensure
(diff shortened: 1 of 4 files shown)


MikaelSmith:
My method for testing this was to raise concurrency to 300, get an SSH target, and add a line in `Executor#run_task` to set `targets` to an array of 300+ of the target passed in. Trying to pass the same target multiple times at a higher level gets deduplicated in Inventory.

lib/bolt/transport/winrm.rb
```diff
@@ -37,10 +37,13 @@ def initialize
         super
         require 'winrm'
         require 'winrm-fs'
+
+        @transport_logger = Logging.logger[::WinRM]
+        @transport_logger.level = :warn
```
adreyer: Do we want to set warn here since we're not trying to silence net::ssh?

MikaelSmith: Any idea why it was set in the first place? I'm not changing the behavior, just where we set it up.

adreyer: not really. I guess we can leave it for now

PR #585, opened 2018-08-21
(BOLT-794) Unwrap sensitive data

nmaludy:
### Problem

Currently, when passing a `Sensitive` string to a task or script, the data is passed as a `Sensitive` class instead of a string, resulting in an error. For more details, see:

https://tickets.puppetlabs.com/projects/BOLT/issues/BOLT-794

### Solution

We need to "unwrap" the underlying string from the `Sensitive` wrapper before invoking the task or script.

We also need to modify our type check in `run_task()` to compare against `RichData` instead of `Data` because `Data` does not contain `Sensitive` in its type alias.

### TODO

* [x] Add unit tests 
* [x] Produce a helpful error message about the type mis-match problems


nmaludy:
Added helpful error message for future issues / type mis-matches when calling `run_task`:

Example (this was a forced error, couldn't actually figure out how to generate a bad type that wasn't accepted by this alias, **suggestions welcome!**):
```
{
  "kind": "bolt/pal-error",
  "msg": "Task parameters is not of type RichData (Variant[Scalar,SemVerRange,Binary,Sensitive,Type,TypeSet,URI,Object,Undef,Default,Hash[RichDataKey,RichData],Array[RichData]]) it is of type: Struct[{'key' => String, 'decrypt' => Boolean[true], 'api_key' => Sensitive[String]}] (file: /path/to/bolt/plans/xxx.pp, line: 55)",
  "details": {
  }
}
```

bolt-modules/boltlib/lib/puppet/functions/run_task.rb
```diff
@@ -119,8 +119,15 @@ def run_task_raw(task_name, targets, description = nil, task_args = nil, &block)
       task = task_signature.task
     end
 
-    unless Puppet::Pops::Types::TypeFactory.data.instance?(use_args)
-      raise with_stack(:TYPE_NOT_DATA, 'Task parameters is not of type Data')
+    unless Puppet::Pops::Types::TypeFactory.rich_data.instance?(use_args)
```
MikaelSmith: I'm not familiar enough with the semantics of RichData yet. Tasks require input that is JSON-serializable, which is why we initially restricted it to Data. If we're accepting more, that seems to imply some sort of transform that needs to be applied when passing the arguments to the task. Is that transform always clear?

MikaelSmith:
This looks like it's probably a good idea, but I want to better understand RichData in general to see the ramifications. It looks like the intention is primarily to accept sensitive data and keep it from being logged.

bolt-modules/boltlib/lib/puppet/functions/run_task.rb
adreyer: I'm concerned that we aren't careful enough with how we treat data downstream of the run functions yet and it may be best if the plan author has to explicitly unwrap sensitive data until we are.

nmaludy: How do you then invoke a task whose parameter is of type `Sensitive[String]` from a plan?

Tasks will need to then modify their parameter types to be `String`, otherwise type checks will fail earlier on in the `run_task` function.

nmaludy:
@MikaelSmith yes, the only reason i switched to RichData was to allow for Sensitive types to be accepted

bolt-modules/boltlib/lib/puppet/functions/run_task.rb
adreyer: We need to consider carefully how the two sensitives interact.

I think this is roughly what they mean.
1. When a parameter is marked sensitive Bolt should not log it anywhere inside "run task".
2. When a value in a plan is marked sensitive Bolt should not log it anywhere during the plan execution

How do they interact though?
It seems obvious that a sensitive value should be unwrapped automatically when given to a sensitive param. The task has opted into treating the parameter sensitively.
Should a sensitive value be unwrapped automatically into a non-sensitive param? The task may leak it.

bolt-modules/boltlib/lib/puppet/functions/run_task.rb
```diff
@@ -117,10 +118,26 @@ def run_task_raw(task_name, targets, description = nil, task_args = nil, &block)
...
+        if task.parameters and task.parameters[k]['sensitive']
+          use_args[k] = Puppet::Pops::Types::PSensitiveType::Sensitive.new(v)
+        end
+      end
+
     end
 
-    unless Puppet::Pops::Types::TypeFactory.data.instance?(use_args)
-      raise with_stack(:TYPE_NOT_DATA, 'Task parameters is not of type Data')
+    unless Puppet::Pops::Types::TypeFactory.rich_data.instance?(use_args)
```
adreyer: If we put this check inside the 124 loop and check each param instead of the entire params hash we can support sensitive from plans without opening up tasks to the entire world of richdata which will probably require more work.

adreyer: Ah yeah. we should move this check to that branch and have a per param check in the "local code" branch to account for sensitive.

nmaludy:
OK i think this is good to go, except for the fact that some transports (winrm and ssh) both log full commands and full outputs when running with `--debug`.

Few options i see:

1. Pass in the "structured" data all the way down into the transports so it is still wrapped with `Sensitive` up until the moment before we send it over SSH / WinRM.
2. Remove the logging code within the transports that exposes the sensitive data
3. Leave it and let people see it if they pass `--debug`

nmaludy:
Talking with @MikaelSmith and he liked option "3. Leave it and let people see it if they pass --debug"

If this is the choice then everything is good to go.

bolt-modules/boltlib/types/argsspec.pp
```diff
@@ -0,0 +1,10 @@
+# SensitiveData represent the acceptable types to pass as arguments into the run_xxx()
```
MikaelSmith: Looks like the name may have changed to ArgsSpec. Can you update the comment?

pre-docs/writing_plans.md
```diff
@@ -290,6 +290,26 @@ $r.each |$result| {
...
+
+If a parameter is declared with `"sensitive": true` within the task's metadata, then the task
+can be run normally. Bolt will automatically ensure that this parameter is not logged using
+the metadata for that parameter.
+
+```
+run_task('task_with_secrets', ..., password => '$ecret!')
+```
+
+Alternatively paramters can have a `Sensitive` type defined within the task metadata.
```
MikaelSmith: Typo here, `paramters` -> `parameters`.

pre-docs/writing_tasks.md
```diff
@@ -602,6 +602,54 @@ For example, the following code in a metadata file describes a `provider` parame
...
+    "password": {
+      "description": "The password",
+      "type": "String[1]",
+      "sensitive": true
+    }
+  }
+}
+```
+
+2) Wrap the parameter with `Sensitive` in the `type` property within the JSON metadata:
```
MikaelSmith: Is there a good reason to prefer this option over the 1st? The 1st is the only one that will work in PE, and I'd prefer to suggest only one way to do things.

nmaludy: The usecase i see for this is:

A task author doesn't mark a parameter as `"sensitive": "true"` but the plan author doesn't want the parameter logged by Bolt, so wrapping it in a `Sensitive(xxx)` would prevent this from happening.

Maybe i should add that usecase to the documentation....

nmaludy: Ahh sorry, didn't see this was in the `Writing Tasks` document... maybe i should remove this option and also remove the Sensitive type from the list in the `Common task data types` section?

Or maybe a note about option 1 being the only option supported by PE? 

Thoughts?

bolt-modules/boltlib/types/argsspec.pp
nmaludy: Fixed, thanks!

pre-docs/writing_plans.md
nmaludy: Fixed, thanks!

pre-docs/writing_tasks.md
MikaelSmith: Oh right, I didn't put together that it was listed in `Common task data types`. I'm partial to removing the option and the entry from `Common task data types`, but I may be forgetting something. @adreyer or @katelopresti do you remember why it was added? It seems like it was added when docs were reorganized.

MikaelSmith: I think this is ok to merge, we can clean up the docs later.

PR #589, opened 2018-08-22
(BOLT-723) Add winrm run_task endpoint

lucywyman:
This adds a new `/winrm/run_task` endpoint to pe-bolt-server which runs a task on a node over the winrm transport. It currently writes a `.ps1` file to the bolt controller system, and uses that as an executable to run the task over winrm. It also sets the `input_method` to 'powershell' in order for the parameters key to work.

### DIFF
--- a/lib/bolt/transport/winrm.rb
+++ b/lib/bolt/transport/winrm.rb
@@ -103,8 +103,18 @@ catch
       end
 
       def run_task(target, task, arguments, _options = {})
-        executable = target.select_impl(task, PROVIDED_FEATURES)
-        raise "No suitable implementation of #{task.name} for #{target.name}" unless executable
+        if from_api?(task)
+          # TODO: Remove as part of BOLT-664
+          dir = Dir.mktmpdir
+          executable = File.join(dir, task.file['filename'])
+          File.open(executable, 'w') { |f|
+            f.write(Base64.decode64(task.file['file_content']))
+          }
+          task.input_method = powershell_file?(executable) ? 'powershell' : 'both'
+        else
+          executable = target.select_impl(task, PROVIDED_FEATURES)
+          raise "No suitable implementation of #{task.name} for #{target.name}" unless executable
+        end
 
         input_method = task.input_method
         input_method ||= powershell_file?(executable) ? 'powershell' : 'both'
@@ -147,6 +157,10 @@ try { & "#{remote_path}" @taskArgs } catch { Write-Error $_.Exception; exit 1 }
                 path, args = *process_from_extension(remote_path)
                 conn.execute_process(path, args, stdin)
               end
+
+            if from_api?(task)
+              FileUtils.remove_entry dir
+            end
             Bolt::Result.for_task(target, output.stdout.string,
                                   output.stderr.string,
                                   output.exit_code)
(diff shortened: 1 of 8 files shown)


lucywyman:
Updated based on changes in https://github.com/puppetlabs/bolt/pull/590. Will wait until that's merged to remove the 'Do no merge'

lib/bolt/transport/winrm.rb
```diff
@@ -100,8 +100,17 @@ def run_script(target, script, arguments, _options = {})
       end
 
       def run_task(target, task, arguments, _options = {})
-        executable = target.select_impl(task, PROVIDED_FEATURES)
-        raise "No suitable implementation of #{task.name} for #{target.name}" unless executable
+        if from_api?(task)
+          # TODO: Remove as part of BOLT-664
+          executable = File.join(__dir__, task.file['filename'])
```
MikaelSmith: That's definitely not a great place to put it. I'd rather use mktmpdir to ensure we don't accidentally overwrite part of Bolt.

lib/bolt_ext/server.rb
```diff
@@ -1,11 +1,15 @@
 # frozen_string_literal: true
 
 require 'sinatra'
+# require 'sinatra/base'
```
MikaelSmith: What's all the extra Sinatra stuff here?

lucywyman: I was hoping to have time to figure out why I can't use 'namespace', and perhaps use it. It seems appropriate here and would be cleaner, but isn't worth sinking too much time into. If it's not done when the API docs get merged I'll remove the comments.

lib/bolt/transport/winrm.rb
```diff
@@ -100,8 +100,19 @@ def run_script(target, script, arguments, _options = {})
       end
 
       def run_task(target, task, arguments, _options = {})
-        executable = target.select_impl(task, PROVIDED_FEATURES)
-        raise "No suitable implementation of #{task.name} for #{target.name}" unless executable
+        if from_api?(task)
+          # TODO: Remove as part of BOLT-664
+          Dir.mktmpdir do |dir|
```
MikaelSmith: The tmp directory will be deleted after exiting the block. That doesn't seem like it'll work. I think you want to use the non-block variant (then delete it at the end).

lib/bolt_ext/server.rb
```diff
@@ -28,8 +28,30 @@ class TransportAPI < Sinatra::Base
...
+
+  post '/winrm/run_task' do
+    content_type :json
+
+    body = JSON.parse(request.body.read)
+    keys = %w[user password port connect-timeout ssl ssl-verify tmpdir cacert extensions]
+    opts = body['target'].select { |k, _| keys.include? k }
+    opts['protocol'] = 'winrm'
+    target = [Bolt::Target.new(body['target']['hostname'], opts)]
+    task = Bolt::Task.new(body['task'])
```
MikaelSmith: In `examples/`, we pass `{ "task": { "name": "...", "file_content": "..." } }`. That doesn't seem to match the API we've documented. Do we still need to pass `name`? The example will also be wrong about where `file_content` goes.

MikaelSmith: Nevermind, `name` is called out in the docs, I was overlooking the `file_content` restructuring. I think the examples are the only thing that's wrong. It's worth either updating or deleting them.

lucywyman: Updated examples -  I had completely forgotten it was even there.

PR #590, opened 2018-08-22
(BOLT-723) Document changes to the bolt-server API

lucywyman:
This commit documents 2 changes to the bolt-server API:
* The addition of a `/winrm/run_task` endpoint, which is very similar to the `/ssh/run_task` endpoint but whose target object has different options
* The addition of a `file` JSON object in place of `file_contents` in the task object.

The latter change is because we need to know the filetype to run tasks on windows hosts. Long term we want to get this data directly from puppetserver, but aren't sure we can make this change in the time frame we have. For now we still need the file_content, in addition to the filetype, and will accept other keys provided by the [puppetserver task api](https://github.com/puppetlabs/puppetserver/blob/master/documentation/puppet-api/v3/task_detail.markdown).

### DIFF
--- a/developer-docs/bolt_server.md
+++ b/developer-docs/bolt_server.md
@@ -34,6 +34,68 @@ Each API endpoint accepts a request as described below. The request body must be
 - `task`: [Task Object](#task-object), *required* - Task to run on target.
 - `parameters`: Object, *optional* - JSON formatted parameters to be provided to task.
 
+For example, the following runs the 'echo' task on localhost:
+```
+{
+  "target": {
+    "hostname": "localhost",
+    "user": "marauder",
+    "password": "I solemnly swear that I am up to no good",
+    "host-key-check": "false"
+  },
+  "task": {
+    "name": "echo",
+    "metadata": {
+      "description": "Echo a message",
+      "parameters": {
+        "message": "Default string"
+      }
+    },
+    "file": {
+      "filename": "echo.rb",
+      "file_content": "IyEvdXNyL2Jpbi9lbnYgYmFzaAplY2hvICRQVF9tZXNzYWdlCg==\n"
+    }
+  },
+  "parameters": {
+    "message": "Hello world"
+  }
+}
+```
+
+## POST /winrm/run_task
+- `target`: [WinRM Target Object](#winrm-target-object), *required* - Target information to run task on.
+- `task`: [Task Object](#task-object), *required* - Task to run on target.
+- `parameters`: Object, *optional* - JSON formatted parameters to be provided to task.
+
+For example, the following runs 'echo' task on localhost:
+```
+{
+  "target": {
+    "hostname": "localhost",
+    "user": "chamber-of-secrets",
+    "password": "parseltongue",
+    "host-key-check": "false"
+  },
+  "task": {
+    "name": "echo",
+    "metadata": {
+      "description": "Echo a message",
+      "parameters": {
+        "message": "Default string"
+      }
+    },
+    "file": {
+      "filename": "echo.rb",
+      "file_content": "IyEvdXNyL2Jpbi9lbnYgYmFzaAplY2hvICRQVF9tZXNzYWdlCg==\n"
+    }
+  },
+  "parameters": {
+    "message": "Hello world"
+  }
+}
+
+```
+
 ### SSH Target Object
 The Target be a JSON object. The following keys are available:
 - `hostname`: String, *required* - Target identifier.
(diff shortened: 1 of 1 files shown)


developer-docs/bolt_server.md
```diff
@@ -47,11 +109,79 @@ The Target be a JSON object. The following keys are available:
...
 
+### WinRM Target Object
+The Target be a JSON object. The following keys are available:
+- `hostname`: String, *required* - Target identifier.
+- `password`: String, *optional* - Password for WinRM transport authentication.
+- `port`: Integer, *optional* - Connection port (Default: `5986`, or `5985` if `ssl: false`.)
+- `user`: String, *optional* - Login user (Default: `root`).
+- `connect-timeout`: Integer, *optional* - How long Bolt should wait when establishing connections.
+- `tmpdir`: String, *optional* - The directory to upload and execute temporary files on the target.
+- `ssl`: Boolean, *optional* - When true, Bolt will use normal http connections for WinRM (Default: `true`).
```
ALawley: Doesn't it use https when true and http when false?

developer-docs/bolt_server.md
```diff
@@ -47,11 +109,79 @@ The Target be a JSON object. The following keys are available:
...
+- `description`: String, *optional* - The task description from it's metadata
+- `parameters`: Object, *optional* - A JSON object whose keys are parameter names, and whose values are JSON objects with 2 keys:
+    - `description`: String, *optional* - The parameter description
+    - `type`: String, *optional* - The type the parameter should accept
+
+#### File
+**NOTE**: We plan to eventually get this information directly from puppetserver
+
+The task file and it's metadata. This is a JSON object that includes the following keys:
+- `filename`: String, *required* - The name of the file the task is in
```
jonathannewman: Wondering what requirements sanitization there are around this.  For example, I don't think this should have any path specification in it for security reasons?

developer-docs/bolt_server.md
```diff
@@ -47,11 +109,79 @@ The Target be a JSON object. The following keys are available:
...
 - `file_content`: String, *required* - Task's base64 encoded file content.
+- `sha256`: String, *optional* - The SHA of the task file
+- `size_bytes`: String, *optional* - The size of the file
+- `uri`: Object, *optional* - A JSON object with the following keys:
```
ALawley: I'm a little confused by this. Is this for once bolt-server is getting task information directly from puppetserver? Is this currently supported at all?

developer-docs/bolt_server.md
lucywyman: @katelopresti I copied this from https://puppet.com/docs/bolt/0.x/bolt_configuration_options.html#winrm-transport-configuration-options looks like we need to update it?

developer-docs/bolt_server.md
lucywyman: We use this to determine how to interpret the file on windows system (`ps1`, `py`, etc.), and to create a file on the node running the task. Given that context are there security concerns? We're not reading from the controller filesystem.

developer-docs/bolt_server.md
lucywyman: No, not currently supported. I can remove it given that, but left them in this PR in case we did want to make any of them required (ie. check for file size or verify the sha).

ALawley: Maybe just denote that they aren't yet supported?

developer-docs/bolt_server.md
jonathannewman: Is the intent this would be used to write the file to the target?  If my filename was `/etc/passwd` for example, will it overwrite the file on the target system?  Or perhaps `/bin/bash` or `../../../../bin/bash`.  I'm just suggesting that perhaps it should be only a simple filename and not anything else allowed.

developer-docs/bolt_server.md
MikaelSmith: If Bolt's not using them, I think I'd remove them.

adreyer: We can remove these keys from the docs but I think we need to work towards a standard format for task details which similar to what puppet-server provides. Bolt-server shouldn't error if these keys are present.

developer-docs/bolt_server.md
adreyer: We will eventually need to support relative names for tasks in sub directories. As long as bolt only uses the last part I don't think the API should restrict this.

lucywyman: I guess it seems like you're only shooting yourself in the foot by overwriting existing files? You already have permission to remote into the machine you're running the task on as the user you're running the task as, so I'm not sure what the threat model is. We could see about raising a warning if the filepath already exists on the system?

jonathannewman: It may just not be existing files.  If I can write arbitrary files to unexpected location on a given system, I may be able to compromise other parts of the system that are expected to be controlled.   This is a pretty normal exploit vector to protect against.

developer-docs/bolt_server.md
lucywyman: M'k, removed!

developer-docs/bolt_server.md
MikaelSmith: This isn't a security concern because you can only overwrite locations the user you're using to SSH lets you, and since you're including the SSH credentials you must already have them.

I agree its good practice to ensure the filename doesn't allow relative path traversal though.

PR #592, opened 2018-08-22
(BOLT-786) Change bolt-server port to 62658

lucywyman:
This sets the bolt-server port higher since it's an internal tool

### DIFF
--- a/developer-docs/bolt_server.md
+++ b/developer-docs/bolt_server.md
@@ -11,7 +11,7 @@ Bolt server can be configured by defining content in HOCON format at one of the 
 **Local Config**: `~/.puppetlabs/bolt-server.conf`
 
 **Options**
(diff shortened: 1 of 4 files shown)


lucywyman:
I've also updated https://confluence.puppetlabs.com/display/ENG/Choosing+a+port+for+services

nicklewis:
Because this is in the dynamic port range, I'm concerned about it possibly conflicting with a port that might be open for some outbound connection. Not sure whether that's a realistic concern or not?

MikaelSmith:
On CentOS 7 and Ubuntu 18.04, the dynamic range appears to be 32768-60999. So, using that and PE's traditional ActiveMQ setting (61613), maybe we should use 62658.

MikaelSmith:
Found by catting out `/proc/sys/net/ipv4/ip_local_port`.

PR #593, opened 2018-08-22
(BOLT-655) Add loglevel and logfile

MikaelSmith:
Adds logging. Sets Bolt logging based on loglevel (default 'notice'). Logs all messages to console, or to logfile if configured.

Also refactors to use the Puma CLI and restrict hosting to localhost.


ext/redhat/pe-bolt-server.init
```diff
@@ -37,7 +37,7 @@ fi
 start() {
     echo -n $"Starting PE Bolt Server: "
     mkdir -p $piddir $logdir
-    daemon $daemonopts --user=$owner $exec "${prefix}/config.ru" ${PE_BOLT_SERVER_OPTIONS}
+    daemon $daemonopts --user=$owner $exec -C "${prefix}/puma_config.rb" -d --pidfile $pidfile ${PE_BOLT_SERVER_OPTIONS}
```
MikaelSmith: These changes require https://github.com/puppetlabs/bolt-vanagon/pull/45.

lucywyman: Is this just one less tool in the chain? Are there other benefits? Not against it, just curious.

ext/tmpfiles/pe-bolt-server.conf
```diff
@@ -1 +0,0 @@
-d /var/run/puppetlabs/bolt-server 0755 pe-bolt-server pe-bolt-server -
```
lucywyman: Why don't we need this anymore?

lib/bolt_ext/server.rb
```diff
@@ -6,6 +6,10 @@
 require 'json'
 
 class TransportAPI < Sinatra::Base
+  get '/' do
```
lucywyman: Is this for testing?

lucywyman: Ah, yes it is :)

spec/fixtures/configs/global-bolt-server.conf
```diff
@@ -1,6 +1,10 @@
 bolt-server: {
+    host: 10.0.0.1
     port: 12345
     ssl-cert: "spec/fixtures/ssl/cert.pem"
     ssl-key: "spec/fixtures/ssl/key.pem"
     ssl-ca-cert: "spec/fixtures/ssl/ca.pem"    
+
```
lucywyman: blank line?

ext/redhat/pe-bolt-server.init
lucywyman: Ah, commit on puppetlabs/bolt-vanagon#45 explained. :+1:

MikaelSmith: Using Puma's CLI and a puma config file gives us access to a bunch of options that aren't there using rackup. Puma adds its own layer of config that much overrides rackup. So it's exposing puma config on the CLI, and letting us use its helpers. So daemonization is now `-d --pidfile ...`.

ext/tmpfiles/pe-bolt-server.conf
MikaelSmith: On systemd, I switched back to running the server foreground (simple mode) rather than having it fork. So no more need for a pidfile.

lucywyman: https://github.com/puppetlabs/bolt-vanagon/pull/45 explained

lib/bolt_ext/server.rb
MikaelSmith: It also gives us a simple API to hit to see that the server's running. It's usually nice when the root path returns 200.

spec/fixtures/configs/global-bolt-server.conf
MikaelSmith: Separated server and logging config.

spec/fixtures/configs/local-bolt-server.conf
```diff
@@ -1,6 +1,10 @@
 bolt-server: {
+    host: 0.0.0.0
     port: 6789
     ssl-cert: "spec/fixtures/ssl/cert.pem"
     ssl-key: "spec/fixtures/ssl/key.pem"
     ssl-ca-cert: "spec/fixtures/ssl/ca.pem"    
+
+    loglevel: info
```
lucywyman: Unrelated to this PR, but it seems like we don't actually support 'verbose' as the docs state: https://puppet.com/docs/bolt/0.x/bolt_configuration_options.html#log-file-configuration-options but support `Logging.init :debug, :info, :notice, :warn, :error, :fatal, :any`.

PR #594, opened 2018-08-22
(BOLT-805) No host key verify w/ host-key-check=false

donoghuc:
This commit changes the netssh verifier from `lenient` to `null` when `host-key-check` is false. The lenient verifier will raise an exception (HostKeyMismatch) when a target is present in `~/.ssh/known_hosts`, but the key given does not match any known for that target. The `null` verifier does no host-key checking at all.


donoghuc:
With this addition the experimental pipeline for puppetlabs-puppet_agent module task acceptance tests are consistently passing.

PR #596, opened 2018-08-23
(BOLT-655) Include umask on pe-bolt-server to lock down logfile

MikaelSmith:
Set a umask to prevent bolt-server's logfile being world-readable when
running as a service.

### DIFF
--- a/ext/redhat/pe-bolt-server.init
+++ b/ext/redhat/pe-bolt-server.init
@@ -10,6 +10,9 @@
 # Source function library.
 . /etc/rc.d/init.d/functions
 
+#set default privileges to -rw-r-----
+umask 027
+
 [ -f /etc/sysconfig/pe-bolt-server ] && . /etc/sysconfig/pe-bolt-server
 
 prefix='/opt/puppetlabs/server/apps/bolt-server'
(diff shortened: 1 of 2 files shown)


ext/systemd/pe-bolt-server.service
```diff
@@ -10,6 +10,8 @@ EnvironmentFile=-/etc/default/pe-bolt-server-service
 Environment=GEM_HOME=/opt/puppetlabs/server/apps/bolt-server/lib/ruby/gems/2.4.0
 ExecStart=/opt/puppetlabs/server/apps/bolt-server/bin/bolt-server -C /opt/puppetlabs/server/apps/bolt-server/puma_config.rb
 Restart=always
+#set default privileges to -rw-r-----
```
lucywyman: I think this is `-rwxr-x---` ?

ext/redhat/pe-bolt-server.init
```diff
@@ -10,6 +10,9 @@
 # Source function library.
 . /etc/rc.d/init.d/functions
 
+#set default privileges to -rw-r-----
```
lucywyman: Same here

ext/systemd/pe-bolt-server.service
lucywyman: Ah, that must just be for directories though? So since files aren't executable they don't get the executable bit?

MikaelSmith: Correct.

PR #602, opened 2018-08-24
(maint) Hard-code GEM_PATH for bolt-server

MikaelSmith:
GEM_HOME appears to be appended to GEM_PATH. If a different version of
Bolt is installed in puppet-agent, that impacts how bolt-server
functions. Set GEM_PATH instead to ensure we use the bolt-server
specific gems first.

This ties bolt-server's service config more closely to the puppet-agent
Ruby setup, since we have to include its GEM_PATH as well.

### DIFF
--- a/ext/redhat/pe-bolt-server.init
+++ b/ext/redhat/pe-bolt-server.init
@@ -16,7 +16,7 @@ umask 027
 [ -f /etc/sysconfig/pe-bolt-server ] && . /etc/sysconfig/pe-bolt-server
 
 prefix='/opt/puppetlabs/server/apps/bolt-server'
-export GEM_HOME="${prefix}/lib/ruby/gems/2.4.0"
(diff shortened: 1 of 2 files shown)


ext/redhat/pe-bolt-server.init
```diff
@@ -16,7 +16,7 @@ umask 027
 [ -f /etc/sysconfig/pe-bolt-server ] && . /etc/sysconfig/pe-bolt-server
 
 prefix='/opt/puppetlabs/server/apps/bolt-server'
-export GEM_HOME="${prefix}/lib/ruby/gems/2.4.0"
+export GEM_PATH="${prefix}/lib/ruby/gems/2.4.0:/opt/puppetlabs/puppet/lib/ruby/gems/2.4.0:/opt/puppetlabs/puppet/lib/ruby/vendor_gems"
```
lucywyman: Why not `GEM_PATH="${prefix}/lib/ruby/gems/2.4.0:$GEM_PATH"`?

MikaelSmith: GEM_PATH likely isn't set. And I'm pretty sure systemd doesn't let you do that.

MikaelSmith:
Going to move this change to packaging.

PR #603, opened 2018-08-24
(maint) Move service config, drop pe-bolt-server reference

MikaelSmith:


lib/bolt/puppetdb/config.rb
```diff
@@ -12,8 +12,8 @@ class Config
                            global: '/etc/puppetlabs/client-tools/puppetdb.conf',
                            win_global: 'C:/ProgramData/PuppetLabs/client-tools/puppetdb.conf' }.freeze
       else
-        DEFAULT_TOKEN = "/etc/puppetlabs/client-tools/pe-bolt-server/token"
-        DEFAULT_CONFIG = { user: "/etc/puppetlabs/puppet/puppetdb.conf",
+        DEFAULT_TOKEN = Bolt::Util.windows? ? 'nul' : '/dev/null'
```
lucywyman: 'nul' or 'null'?

MikaelSmith: https://stackoverflow.com/a/313118

PR #610, opened 2018-08-27
(BOLT-664) Upload file from memory over WinRM

donoghuc:
This commit uses a modified winrm-fs library to upload file contents received by the bolt-server API from memory instead of writing to a temporary file. The modifications to winrm-fs will be included in Bolt by extending relevant winrm-fs classes to avoid being blocked on changes getting accepted to winrm-fs. The winrm-fs work is currently being tracked with https://github.com/donoghuc/winrm-fs/tree/BOLT-664.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -6,11 +6,11 @@ source ENV['GEM_SOURCE'] || 'https://rubygems.org'
 ENV['BOLT_DISABLE_ANALYTICS'] = 'true'
 
 gemspec
-gem  "winrm-fs", git: "https://github.com/donoghuc/winrm-fs.git", branch: "BOLT-664"
 gem "hocon"
 gem "puma"
 gem "rack"
 gem "sinatra"
+gem "winrm-fs", git: "https://github.com/donoghuc/winrm-fs.git", branch: "BOLT-664"
 
 # Required to pick up plan specs in the rake spec task
 gem "puppetlabs_spec_helper",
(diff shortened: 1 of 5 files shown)


Gemfile
```diff
@@ -6,6 +6,7 @@ source ENV['GEM_SOURCE'] || 'https://rubygems.org'
 ENV['BOLT_DISABLE_ANALYTICS'] = 'true'
 
 gemspec
+gem  "winrm-fs", git: "https://github.com/donoghuc/winrm-fs.git", branch: "BOLT-664"
```
MikaelSmith: Let's try a PR to winrm-fs. It looks like it wasn't quite as easy to separate out, and I don't think we're pressed for time right now.

lib/bolt/transport/winrm/connection.rb
```diff
@@ -355,16 +355,23 @@ def make_tempdir
         end
 
         def with_remote_file(file)
-          ext = File.extname(file)
+          if file.is_a?(Hash)
```
MikaelSmith: Can we make this handling somewhat consistent between the SSH and WinRM implementations?

donoghuc:
If we controlled the winrm-fs repo I would propose this as a solution for uploading content from memory to a remote file using winrm. I will work on integrating the changes tracked with https://github.com/donoghuc/winrm-fs/tree/BOLT-664 in to bolt. 

Once we "finalize" the additions we would like to see in winrm-fs I will raise a PR for that project and see if we can get our changes in.

donoghuc:
Update to bolt-vanagon https://github.com/puppetlabs/bolt-vanagon/pull/50

PR #611, opened 2018-08-27
(BOLT-754) Use rails-auth to enforce a whitelist

MikaelSmith:
Allow configuring a whitelist as an array of certnames. Use rails-auth to enforce the whitelist of services allowed to make requests via bolt-server.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -9,6 +9,7 @@ gemspec
 gem "hocon"
 gem "puma"
 gem "rack"
+gem "rails-auth"
 gem "sinatra"
 
 # Required to pick up plan specs in the rake spec task
--- a/lib/bolt_ext/server_acl.rb
+++ b/lib/bolt_ext/server_acl.rb
@@ -1,0 +1,37 @@
+# frozen_string_literal: true
+
+require 'rails/auth/rack'
+
+class TransportACL < Rails::Auth::ErrorPage::Middleware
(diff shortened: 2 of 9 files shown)


MikaelSmith:
Blocked by https://github.com/puppetlabs/bolt-vanagon/pull/48.

MikaelSmith:
This should be accompanied by an acceptance test verifying we can't make requests with a cert not in the whitelist, as well as a config change to add to the whitelist.

lucywyman:
It seems like there might be a possibility of adding more robust spec tests: https://github.com/square/rails-auth/wiki/RSpec-Support#acl-specs ? It'd be nice to see failures before merge.

MikaelSmith:
I can probably refactor a bit to make this more testable.

nicklewis:
This works, but it currently just returns a stack trace when denying the request. Is there a separate ticket to improve error handling or should that be done as part of this PR?

nicklewis:
Ahh that makes sense. I see if I set the env to production, I get a generic message. It definitely would be nice for that to be json though.

MikaelSmith:
Refactored a bit, added error response middleware (which returns `403 Access denied`), and added tests.

lucywyman:
Heh, I think that's just rubocop

lucywyman:
Working as expected when built into a package and installed.

PR #612, opened 2018-08-28
(BOLT-657) Catch sinatra html errors and return json

lucywyman:
By default Sinatra returns HTML when it errors. This can't be parsed into JSON, and makes handling errors for api users difficult and hard to trace. This catches those errors and returns json instead

### DIFF
--- a/lib/bolt_ext/server.rb
+++ b/lib/bolt_ext/server.rb
@@ -6,8 +6,15 @@ require 'bolt/task'
 require 'json'
 
 class TransportAPI < Sinatra::Base
+  # This disables Sinatra's error page generation
+  set :show_exceptions, false
+
   get '/' do
     200
+  end
+
+  get '/500_error' do
+    raise 'Unexpected error'
   end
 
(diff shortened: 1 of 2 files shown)


lib/bolt_ext/server.rb
```diff
@@ -6,10 +6,17 @@
...
 
 class TransportAPI < Sinatra::Base
+  # This disables Sinatra's error page generation
+  set :show_exceptions, false
+
   get '/' do
     200
   end
 
+  get '/500_error' do
```
MikaelSmith: This is specifically for testing?

lucywyman: Yeah - I'm not married to the idea of adding it. It wasn't clear if there was a better way to test the 500 error with rspec, and I didn't want to sink more time into it. I'm not especially worried about the 500 error breaking since it's formatted just like the 404 though, so I'd be fine with removing this.

MikaelSmith: I think it's fine.

PR #615, opened 2018-08-29
(BOLT-750) Remove apply module and apply::resource task

nicklewis:
This module was an experimental inclusion to provide basic functionality
to reuse existing Puppet resource types. Since we've now introduced the
apply() keyword as a general, supported solution to this problem, the
apply module is no longer necessary. We're choosing to immediately
remove this rather than add a deprecation based on the fact that we have
no known users of the feature.

### DIFF
--- a/spec/bolt/cli_spec.rb
+++ b/spec/bolt/cli_spec.rb
@@ -929,28 +929,15 @@ bar
           }
           cli.execute(options)
           json = JSON.parse(output.string)
-          expect(json).to eq(
-            [
-              ["apply::resource", "Apply a single Puppet resource"],
-              ["facts", "Gather system facts"],
-              ["facts::bash", "Gather system facts using bash"],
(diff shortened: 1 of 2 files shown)


spec/bolt/cli_spec.rb
```diff
@@ -931,7 +931,6 @@ def stub_config(file_content = {})
           json = JSON.parse(output.string)
           expect(json).to eq(
             [
-              ["apply::resource", "Apply a single Puppet resource"],
```
adreyer: Can we make this an include rather than an eq?  That wont' solve this problem but will make it easier to add new content. I don't think the point of this test is to make sure we have bundled the things we're supposed to have bundled because it doesn't execute at the right point in the pipeline

nicklewis:
Updated the test to just check for `package` and `service`, which are much less likely to be removed.

PR #616, opened 2018-08-29
(BOLT-794) Adjustments to sensitive handling

nmaludy:
### Changes

* Plans can no longer pass in `Sensitive` typed arguments
* `run_task()` validates input parameters against the `Data` type once again
* The `Boltlib::ArgsSpec` type has been removed, since `Sensitive` is no longer accepted as an argument type.
* The `Sensitive` type is no longer listed in the `writing_tasks` documentation for acceptable parameter data types

### Remains the same

* Task authors can still specify `"sensitive": true` on parameters and they will be wrapped with `Sensitive` under the hood and not logged (except in `--debug` mode).

### DIFF
--- a/pre-docs/writing_plans.md
+++ b/pre-docs/writing_plans.md
@@ -292,7 +292,7 @@ $r.each |$result| {
 
 ## Passing sensitive data to tasks
 
-Tasks can define parameters to be `Sensitive`, meaning the parameter values will not be logged by Bolt.
+Tasks can define parameters to be `sensitive`, meaning the parameter values will not be logged by Bolt.
 
 If a parameter is declared with `"sensitive": true` within the task's metadata, then the task
 can be run normally. Bolt will automatically ensure that this parameter is not logged using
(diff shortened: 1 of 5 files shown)


pre-docs/writing_tasks.md
```diff
@@ -628,25 +624,10 @@ There are two options for defining a task parameter as `Sensitive`:
...
-```
-
-When a parameter is `Sensitive` the only expectation, currently, is that the parameter
-will not be logged. No other precautions are taken.
+When a parameter is `sensitive` the only expectation, currently, is that the parameter
+will not be logged (except when passing `--debug`). No other precautions are taken.
 
-It is up to the task author to implement their tasks in such a way that `Sensitive`
+It is up to the task author to implement their tasks in such a way that `sensitive`
 parameters are not exposed or logged. Please be mindful when coding your tasks!
```
MikaelSmith: Can you also remove line 706/687, the entry about `Sensitive`? I think that was added accidentally, and doesn't work.

pre-docs/writing_plans.md
```diff
@@ -292,7 +292,7 @@ $r.each |$result| {
 
 ## Passing sensitive data to tasks
 
-Tasks can define parameters to be `Sensitive`, meaning the parameter values will not be logged by Bolt.
+Tasks can define parameters to be `sensitive`, meaning the parameter values will not be logged by Bolt.
```
MikaelSmith: Should we note somewhere that you need to explicitly unwrap `Sensitive` types before passing to `run_task`?

nicklewis:
This works exactly how I was thinking of. :+1:

pre-docs/writing_plans.md
nmaludy: Added a sub-section below this.

adreyer:
Can you squash this into a single commit

PR #619, opened 2018-08-30
(BOLT-855) Make Deferred work in 'apply'

MikaelSmith:
Makes the `Deferred` type work in `apply` statements. This type allows
Puppet functions to be invoked when the catalog is being applied on the
agent rather than during compilation.

### DIFF
--- a/libexec/apply_catalog.rb
+++ b/libexec/apply_catalog.rb
@@ -6,6 +6,7 @@ require 'json'
 require 'puppet'
 require 'puppet/configurer'
 require 'puppet/module_tool/tar'
+require 'securerandom'
 require 'tempfile'
 
 args = JSON.parse(ARGV[0] ? File.read(ARGV[0]) : STDIN.read)
@@ -58,7 +59,7 @@ begin
   end
 
   # Ensure custom facts are available for provider suitability tests
-  Puppet::Node::Facts.indirection.find('puppetversion', environment: env)
+  facts = Puppet::Node::Facts.indirection.find(SecureRandom.uuid, environment: env)
 
   report = if Puppet::Util::Package.versioncmp(Puppet.version, '5.0.0') > 0
              Puppet::Transaction::Report.new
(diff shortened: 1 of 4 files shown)


MikaelSmith:
This probably depends on https://github.com/puppetlabs/puppet/pull/7022, and will require a Puppet 6 package on the target (can use Puppet 6 nightlies for testing).

Example: https://github.com/puppetlabs/puppet/pull/7022.

libexec/apply_catalog.rb
```diff
@@ -67,9 +67,14 @@
            end
 
   Puppet.override(current_environment: env, loaders: Puppet::Pops::Loaders.new(env)) do
-    catalog = Puppet::Resource::Catalog.from_data_hash(args['catalog']).to_ral
+    catalog = Puppet::Resource::Catalog.from_data_hash(args['catalog'])
     catalog.environment = env.name.to_s
     catalog.environment_instance = env
+    if defined?(Puppet::Pops::Evaluator::DeferredResolver)
+      # Only available in Puppet 6
+      Puppet::Pops::Evaluator::DeferredResolver.resolve_and_replace(Facter.to_hash, catalog)
```
MikaelSmith: This may not be the best way to collect facts. We should probably use the facts terminus.

libexec/apply_catalog.rb
```diff
@@ -58,7 +58,7 @@
   end
 
   # Ensure custom facts are available for provider suitability tests
-  Puppet::Node::Facts.indirection.find('puppetversion', environment: env)
+  facts = Puppet::Node::Facts.indirection.find('me', environment: env)
```
MikaelSmith: Is there a more sensible name to put here? It doesn't appear to matter.

adreyer: are we concerned that the node name would find the wrong the facts? Should it be a random string?

MikaelSmith: Probably should. There are also changes on the Puppet side, so this may need updates to work with Puppet 6.

MikaelSmith:
BOLT-855.

MikaelSmith:
Now fixed up and ready to go.

libexec/apply_catalog.rb
MikaelSmith: I don't think it'll find the wrong facts after BOLT-883.

spec/integration/apply_spec.rb
```diff
@@ -84,6 +84,19 @@ def root_config
...
         expect(resources).to include('Notify[hello world]')
       end
+
+      it 'applies the deferred type' do
+        result = run_cli_json(%w[plan run basic::defer] + config_flags)
+        expect(result).not_to include('kind')
+        expect(result[0]['status']).to eq('success')
+        resources = result[0]['result']['report']['resource_statuses']
+
+        local_pid = resources['Notify[local pid]']['events'][0]['desired_value'][/(\d+)/, 1]
```
adreyer: `expect(resources['Notify[local pid]']['events'][0]['desired_value']).to match(/(\d+)/)`

seems clearer.

MikaelSmith: The test is specifically testing that local pid and remote pid are different.

PR #620, opened 2018-08-30
(BOLT-751) Rename file_upload function to upload_file

nicklewis:
This function was inconsistent with the other plan functions, including
run_\* and set_\*, which are all \<verb\>\_\<object\> rather than \<object\>\_\<verb\>.
This change standardizes the plan functions so they should be easier to
remember.

This leaves behind a stub file_upload implementation that issues a deprecation
warning and calls upload_file in its place.


nicklewis:
It will be _much_ easier to review this by looking at each commit separately.

PR #622, opened 2018-08-30
(BOLT-794) Remove requirement that Puppet is initialized for transports

MikaelSmith:
Transports must be usable even when Puppet is not initialized. Only
check for Sensitive types if they're loaded, and add an argument to
acceptance tests for `script run` to ensure that it's not used when
Puppet isn't loaded.

### DIFF
--- a/lib/bolt/transport/base.rb
+++ b/lib/bolt/transport/base.rb
@@ -196,6 +196,9 @@ module Bolt
       # This works on deeply nested data structures composed of Hashes, Arrays, and
       # and plain-old data types (int, string, etc).
       def unwrap_sensitive_args(arguments)
+        # Skip this if Puppet isn't loaded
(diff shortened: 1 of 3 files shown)


lib/bolt/transport/base.rb
```diff
@@ -196,6 +196,9 @@ def upload(*_args)
       # This works on deeply nested data structures composed of Hashes, Arrays, and
       # and plain-old data types (int, string, etc).
       def unwrap_sensitive_args(arguments)
+        # Skip this if Puppet isn't loaded
+        return arguments unless defined?(Puppet::Pops::Types::PSensitiveType::Sensitive)
```
MikaelSmith: It might make sense to refactor this later so that we use a Bolt-native type to hide sensitive values.

PR #623, opened 2018-09-05
(maint) Add run-as option to developer docs

lucywyman:
This adds the 'run-as' option example to bolt-server's developer docs


donoghuc:
Also just notice that `tmpdir` is listed twice in https://github.com/puppetlabs/bolt/blob/master/developer-docs/bolt_server.md#winrm-target-object

lucywyman:
I played around with adding "Only available for SSH / WinRM" targets to the transport specific options, but had a hard time finding a way to do it that wasn't clunky + way too many words for too little information. I thought about adding little '*' next to transport-specific options, but I think that'll just be confusing. Ultimately it seems like having the options under different headings is probably enough?

PR #624, opened 2018-09-05
(BOLT-658) Validate bolt-transport-api requests

donoghuc:
Validate JSON requests for ssh and winrm run_task endpoints. This commit adds the gem dependency json-schema. This library provides JSON validation against defined schemas as well as detailed error messages about malformed or prohibited JSON requests.

### DIFF
--- a/lib/bolt_ext/schemas/ssh-run_task.json
+++ b/lib/bolt_ext/schemas/ssh-run_task.json
@@ -1,0 +1,71 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "title": "SSH run_task request",
+  "description": "POST ssh/run_task request schema",
+  "type": "object",
+  "properties": {
+    "target": {
+      "type": "object",
+      "description": "Target information to run task on",
+      "properties": {
+        "hostname": {
+          "type": "string",
+          "description": "Target identifier"
+        },
+        "user": {
+          "type": "string",
+          "description": "Login user"
+        },
+        "password": {
+          "type": "string",
+          "description": "Password for SSH transport authentication"
+        },
+        "private-key-content": {
+          "type": "string",
+          "description": "Contents of private key for SSH"
+        },
+        "port": {
+          "type": "integer",
+          "description": "Connection port"
+        },
+        "connect-timeout": {
+          "type": "integer",
+          "description": "How long Bolt should wait when establishing connections"
+        },
+        "run-as-command": {
+          "type": "array",
+          "description": "Command elevate permissions",
+          "items": { "type": "string" }
+        },
+        "run-as": {
+          "type": "string",
+          "description": "A different user to run commands as after login"
+        },
+        "tmpdir": {
+          "type": "string",
+          "description": "The directory to upload and execute temporary files on the target"
+        },
+        "host-key-check": {
+          "type": "boolean",
+          "description": "Whether to perform host key validation when connecting over SSH"
+        },
+        "sudo-password": {
+          "type": "string",
+          "description": "Password to use when changing users via run-as"
+        }
+      },
+      "oneOf": [
+        { "required": ["password"] },
+        { "required": ["private-key-content"] }
+      ],
+      "required": ["hostname", "user"],
+      "additionalProperties": false
+    },
+    "task": { "$ref": "file:lib/bolt_ext/schemas/task.json"},
+    "parameters": {
+      "type": "object",
+      "description": "JSON formatted parameters to be provided to task"
+    }  
+  },
+  "required": ["target", "task"]
+}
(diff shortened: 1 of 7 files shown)


lib/bolt_ext/server.rb
```diff
@@ -71,3 +77,19 @@ class TransportAPI < Sinatra::Base
     [500, "500: Unknown error: #{e.message}"]
   end
 end
+
+class ValidateRequest
```
donoghuc: Where should this live? Similarly where should the schemas directory live?

lib/bolt_ext/schemas/ssh-run_task.json
```diff
@@ -0,0 +1,71 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
```
donoghuc: I saw these defined in the puppet repo (https://github.com/puppetlabs/puppet/blob/6597624ccfc79ce8f8f0632defd5587d1c026b4b/api/schemas/catalog.json#L2), I think that `json-schema` gem assumes draft-04 by default. Will this cause a problem if there is no internet connection to json-schema.org?

lib/bolt_ext/schemas/winrm-run_task.json
```diff
@@ -21,7 +21,7 @@
           "description": "Password for SSH transport authentication"
         },
         "port": {
-          "type": "integer",
+          "type": ["integer", "string"],
```
donoghuc: Appveyor runs were failing for https://github.com/puppetlabs/bolt/blob/450c17ce617ddfa9c539b5a45786dc3ab74ee574/spec/bolt_ext/server_spec.rb#L159-L178 I am assuming because `JSON.parse` was not preserving the `integer` type for `port`. I am unable to replicate this locally running against vagrant target. Presumably before the JSON validation work `port` being a string did not cause a problem with the test. 

What is a good way to deal with this? Should we coerce all expected integer types after calling JSON.parse and before schema validation? Should we allow both string or integer for port, connection-timeout, etc?

donoghuc:
Bolt-vanagon PR to add json-schema gem: https://github.com/puppetlabs/bolt-vanagon/pull/51

lib/bolt_ext/schemas/ssh-run_task.json
nicklewis: It appears the `$schema` here is just parsed to determine which of the built-in validators to use, so no network required.

lib/bolt_ext/server.rb
lucywyman: I think here is fine - we can always pull it out if it gets too unwieldy

lib/bolt_ext/schemas/ssh-run_task.json
```diff
@@ -0,0 +1,71 @@
...
+        "password": {
+          "type": "string",
+          "description": "Password for SSH transport authentication"
+        },
+        "private-key-content": {
+          "type": "string",
+          "description": "Contents of private key for SSH"
+        },
+        "port": {
+          "type": "integer",
```
lucywyman: I think this can be `["integer", "string"]` like for winrm

lib/bolt_ext/server.rb
```diff
@@ -71,3 +77,19 @@ class TransportAPI < Sinatra::Base
...
+  def initialize(endpoint, request)
+    case endpoint
```
lucywyman: It might be easier to just have the schema uri be from the endpoint that's passed in, like:
`@schema = "lib/bolt_ext/schemas/#{endpoint}.json"`

Then you don't need the case statement.

lucywyman: Actually, this seems like it probably doesn't need to be a whole class? I think having:
`schema_error = JSON::Validator.fully_validate("lib/bolt_ext/schemas/ssh-run-task.json", body)` would work.

lib/bolt_ext/schemas/ssh-run_task.json
nicklewis: It seems better to be strict. Maybe we should change winrm to just integer?

donoghuc: I agree that strict is better. Added some logic for the winrm transport to get it through appveyor. Could add that same checking to expected integer fields in ssh as well.

lib/bolt_ext/server.rb
```diff
@@ -48,6 +51,22 @@ class TransportAPI < Sinatra::Base
     content_type :json
 
     body = JSON.parse(request.body.read)
+    # Some windows ruby interpreters will not preserve integer type when parsing JSON
```
nicklewis: We should figure out what's actually going on here...

donoghuc: The environment variable for BOLT_WINRM_PORT port was a string in https://github.com/puppetlabs/bolt/blob/0c211fd33b8e0e9defe01c2df6e30e00eb187def/spec/lib/bolt_spec/conn.rb#L27
(facepalm)

Issue #102, opened 2017-10-12
Streaming stdout and stderr


philtech-oh:
When running a script or command remotely, Bolt should stream the output (stdout and stderr) back in near real time.  For a script that takes a long time to complete, it is frustrating to wait for script outpuy.

Iristyle:
Thanks - we have already discussed this a bit internally.  This may have to be limited to SSH transport only to start.  WinRM streaming can be a bit tricky given the nature of the SOAP based protocol.

genebean:
FWIW, [fabric](http://www.fabfile.org/) does this really well. It prefixes all the output with the name of the host it ran on and a timestamp. It also returns the command that was run as part of the output (I think) which makes keeping track of where you are in the process much easier. 

All in all, this combination of feedback makes it easy to track whats going on and, when output is being run in parallel, the name prefixing lets you tell one from the other. In the event that you need to see a single system's output after the fact for debugging you can easily filter the text from your console by hostname.

MikaelSmith:
If you're using SSH or WinRM, this may already be visible via `--debug`. Try `bolt command run 'for ((i=10; i>=1; i--)) do sleep 1; echo $i; done' -n some_nodes --debug`

krorvikrl:
I'm +1 on this. It is useful for pipelines that generate a bit of output, for instance.

By the way, we've seen cases where bolt seems to buffer output for a while, then simply returning no output, and exiting nonzero failing the relevant pipeline. Even if the job is successful on the remote node. Is this due to overflowing some buffer?

In either case, this is an interesting feature for us :)

genebean:
@MikaelSmith You are correct that with `--debug` that bolt will stream output but the usefulness is lacking in my opinion. The difference between bolt's `--debug` and normal output of fabric is that with fabric you know where the output is coming from. 

I generated the output below using each tool and the sample command you listed above. In the fabric one you may notice that each host is grouped together... that is coincidence as I have seen the hosts' output all mixed together many times which is why the prefixing of each piece of output is very important to me.

**Bolt Output**

```bash
$ bolt command run 'for ((i=10; i>=1; i--)) do sleep 1; echo $i; done' -n some-server-prod-6.example.com,some-server3-prod.example.com,some-server5-prod.example.com --debug

Did not find config for some-server-prod-6.example.com in inventory
Did not find config for some-server3-prod.example.com in inventory
Did not find config for some-server5-prod.example.com in inventory
Started with 100 max thread(s)
[...]

PR #625, opened 2018-09-06
(TBD) Provides a '--stream' option for streaming output

MikaelSmith:
Adds a `stream` transport option for streaming output that's implemented
on SSH and WinRM. Augments logging to include target name, and when
`stream` is enabled outputs stdout and stderr from executing commands to
a special `stream` log level (that can be used to direct it to a
specific file).

Fxes #102.

### DIFF
--- a/lib/bolt/bolt_option_parser.rb
+++ b/lib/bolt/bolt_option_parser.rb
@@ -251,6 +251,9 @@ Usage: bolt apply <manifest.pp> [options]
       define('--tmpdir DIR', 'The directory to upload and execute temporary files on the target') do |tmpdir|
         @options[:tmpdir] = tmpdir
       end
+      define('--[no-]stream', 'Request results be streamed for transports that support streaming output') do |stream|
+        @options[:stream] = stream
+      end
 
       separator 'Display:'
       define('--format FORMAT', 'Output format to use: human or json') do |format|
--- a/lib/bolt/config.rb
+++ b/lib/bolt/config.rb
@@ -34,7 +34,7 @@ module Bolt
 
     TRANSPORT_OPTIONS = %i[password run-as sudo-password extensions
                            private-key tty tmpdir user connect-timeout
-                           cacert token-file service-url].freeze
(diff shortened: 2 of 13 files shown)


MikaelSmith:
This is proposal for how we could address https://github.com/puppetlabs/bolt/issues/102.

MikaelSmith:
I haven't looked into what could be done for the local transport. PCP would require changes to Orchestrator/pxp-agent that are not currently planned.

MikaelSmith:
TODO:
- update the Docker transport
- investigate whether we can stream Local transport
- don't repeat yourself (Bolt output)

lib/bolt/bolt_option_parser.rb
```diff
@@ -251,6 +251,9 @@ def initialize(options)
       define('--tmpdir DIR', 'The directory to upload and execute temporary files on the target') do |tmpdir|
         @options[:tmpdir] = tmpdir
       end
+      define('--[no-]stream', 'Request results be streamed by on nodes that support it') do |stream|
```
nicklewis: "by on nodes"

adreyer:
There was some offline discussion about whether this is best solved with a special flag, reorganizing the verbosity levels or if we have to create a new format. As written this just bumps the log level of a certain class of debug messages to always print.

- It double prints output first streaming to stderr then normally to stdout.
- It doesn't work well with plans or other places where you may want some actions to be verbose and others quieter.

The user story here seems like:

I'm doing something with a long running task and script and I want to make sure bolt isn't just frozen

Is this really better than passing `--debug`? Should we just move the "stream" output to a higher level so it prints with `--verbose`? Is it a big problem for the story that this double prints?

adreyer:
Closing this since it's not moving forward. The ticket remains open

genebean:
I am not sure the best way to verbalize why one way is better than another to solve this but I can say that again today I found it very disconcerting to not be able to just watch the live output of what was being done by bolt. I think the idea of moving streaming to the --verbose level could make good sense here.

PR #629, opened 2018-09-10
(TBD) Bolt Service Dockerfile

davejohnston:
This wraps bolt-server in a container.  It uses a simple shell script to
install and configure defaults.

Long term this should be built in the same way as rpm,dpkg packages are
built to avoid issues such as systemctl not running in most containers.  We could also switch to a lighter alpine container, to try and get the size down.

To build this requires docker and golang

    ```
    cd bolt-service
    make
    ```

To run

    ```
    make run
    ```

See the README.md for more details


bolt-service/README.md
```diff
@@ -0,0 +1,81 @@
...
+## Building
+Run make to build the project, this generates test certs and creates a docker
+image called puppet/bolt-service:latest
+
+```
+make
+```
+
+## Running the service
+You can run the service by creating a new container.  N.B by default the Service
```
lucywyman: Out of curiosity what does `N.B` mean?

bolt-service/README.md
```diff
@@ -0,0 +1,81 @@
...
+listens on port 8144, so you must ensure it is exposed.
```
lucywyman: We changed it! It listens on `62658` (call 1-800-6-BOLT if you ever need to remember it :smile: )

bolt-service/scripts/setup.sh
```diff
@@ -0,0 +1,42 @@
+#!/bin/bash
+
+BOLT_SERVER_PACKAGE=pe-bolt-server_0.22.0.7.g3257f59-1bionic_amd64.deb
```
lucywyman: Does this get updated somehow? Should we read it from a file instead, so the file can be updated as we release new versions? I can also talk to RE about making a 'latest' directory, though they likely have higher priorities for the next few weeks.

lucywyman:
@davejohnston  Those errors seem to just be rubocop - you might try giving it a run and either fixing the issues or turning rubocop off where it isn't applicable.

bolt-service/scripts/setup.sh
```diff
@@ -0,0 +1,42 @@
...
+
+# Install Puppet-Agent and Bolt Server
+apt update -y
+apt install -y curl vim
```
lucywyman: What does this need vim for? Is vi not there already?

bolt-service/scripts/setup.sh
```diff
@@ -0,0 +1,42 @@
...
+mkdir /tmp/bolt && pushd /tmp/bolt
+# Setup Puppet Agent Repo
+curl -O http://nightlies.puppet.com/apt/puppet6-nightly-release-bionic.deb && apt install -y ./puppet6-nightly-release-bionic.deb && rm ./puppet6-nightly-release-bionic.deb && apt update -y
+
+# Install PE Package
+curl -O https://artifactory.delivery.puppetlabs.net/artifactory/debian__local/pool/development/pe-bolt-server/3257f5940e3d4cebfa60ff9a06540f3b689020a6/ubuntu-18.04/${BOLT_SERVER_PACKAGE}
```
lucywyman: This also seems very hard coded - is there a way to build this URL so that we don't have to update every release?

bolt-service/scripts/setup.sh
```diff
@@ -0,0 +1,42 @@
...
+cd /etc/puppetlabs/bolt-server/ssl
+cp /tmp/certs/localhost.crt /tmp/certs/localhost.key /tmp/certs/bolt-server-ca.crt /etc/puppetlabs/bolt-server/ssl/
+
+# Create SSL Config for Bolt Service
+cat <<EOF > /etc/puppetlabs/bolt-server/conf.d/bolt-server.conf
+bolt-server: {
+  ssl-cert: "/etc/puppetlabs/bolt-server/ssl/localhost.crt"
+  ssl-key: "/etc/puppetlabs/bolt-server/ssl/localhost.key"
+  ssl-ca-cert: "/etc/puppetlabs/bolt-server/ssl/bolt-server-ca.crt"
+  port: 8144
```
lucywyman: 62658

bolt-service/test/request.json
```diff
@@ -0,0 +1,23 @@
+{
+  "target": {
+    "hostname": "test.delivery.puppetlabs.net",
+    "user": "root",
+    "password": "Q*****!",
+    "options": {
```
lucywyman: We've also updated the API: https://github.com/puppetlabs/bolt/blob/master/developer-docs/bolt_server.md#post-sshrun_task - no more 'options' key

bolt-service/scripts/setup.sh
```diff
@@ -0,0 +1,42 @@
...
+  host: 0.0.0.0
```
lucywyman: If this is only taking requests from localhost this should be 127.0.0.1

bolt-service/scripts/setup.sh
davejohnston: Ideally we wouldn't use the package at all, systemctl doesn't actually run on most containers, so we get errors in the output due to service scripts.   It would be great to install the pe-bolt-server from source.   I'm happy to make changes if someone can give me guidance - so ideally if we build a container here it would reflect current codeline.   Is there a way to gem install from source?

bolt-service/test/test.sh
```diff
@@ -0,0 +1 @@
+curl --insecure -X POST -H "Content-Type: application/json" -d @request.json -E ../certs/127.0.0.1.crt --key ../certs/127.0.0.1.key -k https://localhost:8144/ssh/run_task
```
lucywyman: I don't think you need the `--insecure` here, and might try sending the CA as well instead of `-k`

bolt-service/scripts/setup.sh
davejohnston: Yeah we can take out vim

lucywyman:
Thanks for putting this up @davejohnston ! I think it'd be really helpful for testing, especially for non-bolt members.

MikaelSmith:
This can target the master branch now.

bolt-service/README.md
lucywyman: https://en.wikipedia.org/wiki/Nota_bene Ah so cute!

bolt-service/Makefile
```diff
@@ -0,0 +1,25 @@
+all: image
```
lucywyman: @davejohnston  You probably know better than I - can this be a docker-compose file? If not why not?

MikaelSmith:
Is this something we want to get working?

The pe-bolt-server packages aren't shipped publicly. I'd lean towards using alpine with its ruby, build and install the bolt gem, and add extra gems from the Gemfile.

I also think pe-bolt-server's usage has changed. It's generally `bolt-server -e production -C puma_config.rb`.

lucywyman:
@MikaelSmith Yes! At least in the mid-term, possibly the long term, this is something Discovery plans to use.

Agreed on both those changes. I've talked to Dave - feel free to push any changes you're interested in working on, but don't feel obligated since nothing is ticketed. I plan to wait until after Puppetize to work a little more on this, then sit down again with @davejohnston  and see if this is meeting their needs.

davejohnston:
@MikaelSmith @lucywyman if you guys can get this into an Alpine container that would be great.  container size would be a concern for us, so keeping it to a small container around 60MB would be great.    I used the ubuntu container as I only had instructions to install the service using the package.

PR #632, opened 2018-09-13
(BOLT-766) Add 'site' to default modulepath

nicklewis:
Previously, the default modulepath was just `modules`. That made it
difficult to use a Puppetfile with `bolt puppetfile install` while also
using some custom module content, as every custom module needed to be
added to Puppetfile with `local: true` simply in order to be preserved
when updating.

Now, we use `modules:site` as the default modulepath, which provides
users a location to install modules via the Puppetfile (`modules`) as
well as a location for custom content that should not be managed with
the Puppetfile (`site`). `bolt puppetfile install` will always install
to the first element of the modulepath, so we break the typical Puppet
convention and put `modules` *first*.

### DIFF
--- a/lib/bolt/boltdir.rb
+++ b/lib/bolt/boltdir.rb
@@ -25,7 +25,7 @@ module Bolt
       @path = Pathname.new(path).expand_path
       @config_file = @path + 'bolt.yaml'
       @inventory_file = @path + 'inventory.yaml'
-      @modulepath = [(@path + 'modules').to_s]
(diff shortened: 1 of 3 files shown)


nicklewis:
I'm not sure if there's a specific reason to put `site` before `modules` in Puppet other than that it logically makes sense that your site-specific content should override any upstream content. It's important in Bolt that it comes _after_ `modules` though, so that `modules` can be managed with a Puppetfile and `site` can be preserved.

PR #634, opened 2018-09-14
(maint) Add note to install modules with gem install

lucywyman:
**What this changes** This adds a note to the gem install docs and the apply manifest docs about installing bolt's required puppet modules
**Why** `apply_prep` specifically doesn't work without modules installed, and it's difficult to find documentation on why or know that it's related to installing from a gem.

### DIFF
--- a/pre-docs/applying_manifest_blocks.md
+++ b/pre-docs/applying_manifest_blocks.md
@@ -8,6 +8,10 @@ Within a plan, you can use Bolt to apply blocks of Puppet code (manifest blocks)
 
 [Writing plans](writing_plans.md)
 
+**Note:** If you installed bolt as a ruby gem, you'll want to:
+1. Create a Puppetfile with the contents of [bolt's Puppetfile](../Puppetfile) at `~/.puppetlabs/bolt/Puppetfile`
+2. Run `bolt puppetfile install`
+
 ## Create a sample manifest for nginx on Linux
 
 Create a manifest to set up a web server with [nginx](https://nginx.org) and run it as a plan. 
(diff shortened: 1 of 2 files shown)


pre-docs/applying_manifest_blocks.md
```diff
@@ -8,6 +8,8 @@ Within a plan, you can use Bolt to apply blocks of Puppet code (manifest blocks)
 
 [Writing plans](writing_plans.md)
 
+**Note:** If you installed bolt as a ruby gem, you'll need to `cd` into the bolt directory and run `bolt puppetfile install` to install required puppet modules
```
nicklewis: I don't think this actually works, since `bolt puppetfile install` assumes the Puppetfile is inside `Boltdir`... You would need to just regular `r10k puppetfile install`.

pre-docs/bolt_installing.md
```diff
@@ -201,6 +201,12 @@ If Ruby is already included in your operating system, verify that it is version
     ```
     bolt --help
     ```
+4. To install bolt's Puppet module dependencies:
+
+    ```
+    cd bolt
+    bolt puppetfile install
```
MikaelSmith: I'm not sure this will work. Bolt's puppetfile includes an absolute module path. Although from my testing we may be ignoring that.

MikaelSmith: Oh, this uses `~/.puppetlabs/bolt/Puppetfile`. And I don't see any way to override it. So I think the instructions need to say: add the modules in https://github.com/puppetlabs/bolt/blob/master/Puppetfile to `~/.puppetlabs/bolt/Puppetfile`.

PR #636, opened 2018-09-14
(BOLT-656) Add single-threaded executor for bolt-server

MikaelSmith:


lib/bolt/executor.rb
```diff
@@ -40,7 +40,7 @@ def initialize(concurrency = 1,
 
       @noop = noop
       @run_as = nil
-      @pool = Concurrent::ThreadPoolExecutor.new(max_threads: concurrency)
+      @pool = concurrency > 0 ? Concurrent::ThreadPoolExecutor.new(max_threads: concurrency) : nil
```
nicklewis: I think if no concurrency is configured, we should use `Concurrent.global_immediate_executor`, which works like a normal Concurrent::Executor, but executes in the current thread.

Alternately, we could support passing *either* a Concurrent::Executor *or* a concurrency level here and let the caller decide. That would be useful if we wanted multiple Bolt::Executor instances backed by a single Concurrent::Executor

MikaelSmith: I also wanted to avoid the promises, although maybe that's not much overhead.

MikaelSmith: Alright, switched to using an ImmediateExecutor. I didn't see any negative impact in memory use from that (when running 1000 tasks).

lib/bolt_ext/server.rb
```diff
@@ -10,10 +10,24 @@ class TransportAPI < Sinatra::Base
...
+  def initialize(app = nil)
+    @executor = Bolt::Executor.new(0, load_config: false)
+    super(app)
+  end
+
   get '/' do
     200
   end
 
+  get '/admin/gc' do
```
nicklewis: Exposing this via the API concerns me... Could we gate this behind `RACK_ENV` for dev purposes only? I'm not quite as worried about the `/gc_stats` endpoint.

MikaelSmith: Yeah, I can add that.

PR #637, opened 2018-09-17
(maint) Update JSON schema paths

donoghuc:
Previously the paths for validating JSON schema templates only worked when running from Rspec tests as they were relative to the root directory of Bolt. This commit handles the paths accordingly.

### DIFF
--- a/lib/bolt_ext/server.rb
+++ b/lib/bolt_ext/server.rb
@@ -10,6 +10,17 @@ class TransportAPI < Sinatra::Base
   # This disables Sinatra's error page generation
   set :show_exceptions, false
 
+  def initialize
+    super()
+    @schemas = {
+      "ssh-run_task" => JSON.parse(File.read(File.join(__dir__, 'schemas', 'ssh-run_task.json'))),
+      "winrm-run_task" => JSON.parse(File.read(File.join(__dir__, 'schemas', 'ssh-run_task.json')))
+    }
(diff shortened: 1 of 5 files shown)


donoghuc:
the schemas directory is not available when installing the pe-bolt-server package. Is this something we have to explicitly add in packaging? 
```
root@j4v2varc0vcgmcr:/opt/puppetlabs/server/apps/bolt-server/lib/ruby/gems/bolt-0.22.0/lib/bolt_ext# ls
puppetdb_inventory.rb  server_acl.rb  server_config.rb  server.rb
```

MikaelSmith:
Yes. Need to add it to bolt.gemspec.

lib/bolt_ext/server.rb
```diff
@@ -22,7 +22,8 @@ class TransportAPI < Sinatra::Base
     content_type :json
 
     body = JSON.parse(request.body.read)
-    schema = "lib/bolt_ext/schemas/ssh-run_task.json"
+    schema = File.join(__dir__, 'schemas', 'ssh-run_task.json')
+    JSON::Validator.add_schema(shared_schema)
```
MikaelSmith: Is this thread-safe? Maybe we should just do this on startup.

MikaelSmith: I'd like to see this load all the schemas into memory at startup (the `initialize` function), add the shared schema once, then use the preloaded ssh and winrm schemas to validate the body.

PR #639, opened 2018-09-17
(BOLT-840) Include "vendored" Puppet types

nicklewis:
In Puppet 6, many types and providers have been removed from core into
their own modules. Those modules are included in the puppet-agent package
so the out-of-the-box experience remains the same. In order to preserve
that same experience when using the apply() functionality, we now vendor
the same modules in Bolt.

Because we vendor the externalized "core" Puppet modules, we need to avoid
syncing them to agents. These plugins are included out-of-the-box in
puppet-agent 5 (where they are core) and 6 (as modules). That means it's
unnecessary to copy any of this content to agents. Moreover, because the
vendored modules are only intended to be used by Puppet 6 agents, they can
cause problems when syncing to Puppet 5 agents, where they can supersede
the built-in plugins with potentially backward-incompatible code.

### DIFF
--- a/Puppetfile
+++ b/Puppetfile
@@ -12,6 +12,18 @@ mod 'puppet_agent',
     git: 'https://github.com/puppetlabs/puppetlabs-puppet_agent',
     ref: '319ce44a65e73bcf2712ad17be01f9636f0673c9'
 
+# Core types and providers for Puppet 6
+mod 'puppetlabs-augeas_core', '1.0.1'
+mod 'puppetlabs-host_core', '1.0.1'
+mod 'puppetlabs-scheduled_task', '1.0.0'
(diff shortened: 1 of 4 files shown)


MikaelSmith:
We're still susceptible to incompatible type definitions in the newer module code, but hopefully that's not broken often.

nicklewis:
@MikaelSmith That's true. If you have a version of Bolt that's newer than your agent, you may have code that Bolt believes is valid that doesn't work on the agent. I'm not too worried about that; if your agent is old, you can't use new features. The flipside is that an older version of Bolt may believe some parameter isn't valid when it actually is. In that case, you can update Bolt.

In either case, you could install the newer module version in Bolt *explicitly*, and then Bolt would copy it over and would work.

PR #641, opened 2018-09-18
(BOLT-844) Add Puppet 6 PATH for Ruby on Windows

MikaelSmith:
Ensure ruby tasks continue to work on Windows when upgrading to Puppet 6.


MikaelSmith:
Testing is covered by Orchestrator tests with Puppet 6, but we're going to need a way to test Puppet 5 and Puppet 6 in acceptance. That pretty much means we need to install and uninstall Puppet versions in tests.

adreyer:
We should add tests but it's probably best to get this merged first.

PR #643, opened 2018-09-18
(BOLT-646) Have bolt_catalog use PAL

lucywyman:
**What this changes** This has the bolt catalog use the pal catalog compiler instead of having it's own bolt catalog compiler. It removes unnecessary files, and adds unit tests for bolt catalog
**Why** The PAL catalog compiler will provider greater stability, and removing old files improves organization of the repo.


lucywyman:
Relies on changes in https://github.com/puppetlabs/puppet/pull/7102

lib/bolt/catalog.rb
```diff
@@ -69,33 +57,32 @@ def setup_inventory(inventory)
...
-          end
-
-          setup_node(node, target["trusted"])
-
+        Puppet::Pal.in_tmp_environment('bolt_catalog',
+                                       modulepath: request["modulepath"] || [],
+                                       facts: target["facts"] || {},
+                                       variables: target["variables"] || {}
+        ) do |pal|
           Puppet.override(pal_main: pal_main,
```
adreyer: I don't think we need to set this anymore(I could be wrong)

lib/bolt/catalog.rb
```diff
@@ -69,33 +57,32 @@ def setup_inventory(inventory)
...
                           bolt_pdb_client: pdb_client,
                           bolt_inventory:
-                          setup_inventory(request['inventory'])) do
-            compile_node(node)
-          end
+                          setup_inventory(request['inventory'])
+           ) do
+             pal.with_catalog_compiler(
+               facts: target["facts"] || {},
+               variables: target["variables"] || {}
```
adreyer: Do we need to set these twice?

lib/bolt/catalog.rb
```diff
@@ -69,33 +57,32 @@ def setup_inventory(inventory)
     def compile_catalog(request)
       pal_main = request['code_ast'] || request['code_string']
       target = request['target']
-
       pdb_client = Bolt::PuppetDB::Client.new(Bolt::PuppetDB::Config.new(request['pdb_config']))
 
       with_puppet_settings(request['hiera_config']) do
         Puppet[:code] = ''
```
adreyer: I don't think we need to set this anymore

lib/bolt/catalog.rb
```diff
@@ -28,30 +32,15 @@ def with_puppet_settings(hiera_config)
...
-    end
-
     def generate_ast(code)
       with_puppet_settings do
-        Puppet::Pal.in_tmp_environment("bolt_parse") do |_pal|
-          node = Puppet.lookup(:pal_current_node)
-          compiler = Puppet::Parser::BoltCompiler.new(node)
-          compiler.dump_ast(compiler.parse_string(code))
+        Puppet::Pal.in_tmp_environment("bolt_parse") do |pal|
+          pal.with_catalog_compiler do |compiler|
```
adreyer: I think you can maybe set trusted here with something like

`Puppet.lookup(:pal_current_node).trusted_data = trusted_data`

nicklewis:
I tried this against a plan that uses `include apache`, and the only difference in the resulting catalog is that `Stage[main]` now has the `"class"` tag, which seems technically more correct.

PR #645, opened 2018-09-19
(BOLT-854) Properly escape stdin when using --run-as

nicklewis:
When using --run-as, we create a wrapper script around the task and
embed the stdin directly into the script inside a heredoc. Previously,
it was possible for escape sequences within that heredoc to corrupt the
stdin, resulting in incorrect task behavior. We now properly set the
heredoc to ignore all escape sequences, ensuring the task receives
the stdin exactly as intended.

Because we always send stdin as a single line of JSON, it's impossible
for it to escape the heredoc by including a line starting with EOF (our
heredoc marker).

### DIFF
--- a/lib/bolt/transport/ssh.rb
+++ b/lib/bolt/transport/ssh.rb
@@ -174,7 +174,7 @@ module Bolt
       def make_wrapper_stringio(task_path, stdin)
         StringIO.new(<<-SCRIPT)
 #!/bin/sh
-'#{task_path}' <<EOF
+'#{task_path}' <<'EOF'
 #{stdin}
 EOF
 SCRIPT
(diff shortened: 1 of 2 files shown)


lib/bolt/transport/ssh.rb
```diff
@@ -174,7 +174,7 @@ def run_task(target, task, arguments, options = {})
       def make_wrapper_stringio(task_path, stdin)
         StringIO.new(<<-SCRIPT)
 #!/bin/sh
-'#{task_path}' <<EOF
+'#{task_path}' <<'EOF'
```
MikaelSmith: Can we add a test around this?

MikaelSmith:
This seems like it should have a ticket.

PR #650, opened 2018-09-20
(maint) Do no restrict keys in bolt-server task metadata schema

donoghuc:
This commit changes the schema for Task in bolt server to allow additional properties in the task metadata field. It explicitly adds a definition for `input_method` as there are only two acceptable values.

### DIFF
--- a/lib/bolt_ext/schemas/task.json
+++ b/lib/bolt_ext/schemas/task.json
@@ -4,7 +4,6 @@
   "title": "Task",
   "description": "Task schema for bolt-server",
   "type": "object",
-  "description": "The task is a JSON object which includes the following keys",
   "properties": {
     "name": {
       "type": "string",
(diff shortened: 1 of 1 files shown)


lib/bolt_ext/schemas/task.json
```diff
@@ -31,9 +31,13 @@
               "description": "The type the parameter should accept"
             }
           }
+        },
+        "input_method": {
+          "type": "string",
+          "enum": ["stdin", "environment"],
```
MikaelSmith: There are other options. For Windows, `powershell` is valid.

MikaelSmith: https://github.com/puppetlabs/puppet-specifications/tree/master/tasks#input-methods

donoghuc: OK, so maybe we just leave out `input_method` and remove the additionalProperties bit? Looks like you did that as part of #651. @adreyer Asked earlier if we should be validating anything in metadata? For example if the task description is an integer would that really matter?

PR #652, opened 2018-09-20
(maint) Add Puppetfile code inline to docs

lucywyman:
**What this changes** Instead of linking to the Puppetfile in the bolt module, this should include the Puppetfile code inline to be copy-pasted
**Why** Since we don't expect to update the Puppetfile that often, having a better user experience of copy-pasting rather than having to go to the bolt repo outweighs the risk of having the Puppetfile content in 3 places.

### DIFF
--- a/pre-docs/applying_manifest_blocks.md
+++ b/pre-docs/applying_manifest_blocks.md
@@ -9,7 +9,26 @@ Within a plan, you can use Bolt to apply blocks of Puppet code (manifest blocks)
 [Writing plans](writing_plans.md)
 
 **Note:** If you installed bolt as a ruby gem, you'll want to:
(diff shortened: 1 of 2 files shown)


pre-docs/applying_manifest_blocks.md
```diff
@@ -9,7 +9,26 @@ Within a plan, you can use Bolt to apply blocks of Puppet code (manifest blocks)
 [Writing plans](writing_plans.md)
 
 **Note:** If you installed bolt as a ruby gem, you'll want to:
-1. Create a Puppetfile with the contents of [bolt's Puppetfile](../Puppetfile) at `~/.puppetlabs/bolt/Puppetfile`
+1. Create a Puppetfile with the contents of [bolt's Puppetfile](../Puppetfile) at `~/.puppetlabs/bolt/Puppetfile`:
```
adreyer: I don't think putting this inline is a good idea. I highly suspect it will result in us forgetting to update it.

side-note. The local modules should be left out of this.

PR #654, opened 2018-09-24
(BOLT-758) Support different input_methods for implementations

MikaelSmith:
Update for task spec revision 3: support input_method in individual implementations.

Depends on https://github.com/puppetlabs/bolt/pull/651.

### DIFF
--- a/lib/bolt/task.rb
+++ b/lib/bolt/task.rb
@@ -54,9 +54,12 @@ module Bolt
                files.first.dup
              end
 
-      unless (inmethod = metadata['input_method']).nil?
-        impl['input_method'] = inmethod
-      end
+      inmethod = impl['input_method'] || metadata['input_method']
+      impl['input_method'] = inmethod unless inmethod.nil?
+
+      mfiles = impl.fetch('files', []) + metadata.fetch('files', [])
+      impl['files'] = mfiles.map { |file| { 'name' => file, 'path' => file_map[file] } } unless mfiles.empty?
+
       impl
     end
   end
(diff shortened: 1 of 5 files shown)


lib/bolt/task.rb
```diff
@@ -1,23 +1,66 @@
...
+               impl
+             else
+               files.first.dup
+             end
+
+      unless (inmethod = impl['input_method'] || metadata['input_method']).nil?
+        impl['input_method'] = inmethod
+      end
+      unless (mfiles = impl.fetch('files', []) + metadata.fetch('files', [])).empty?
+        impl['files'] = mfiles.map { |filename| { 'name' => filename, 'path' => file_map[filename] } }
```
MikaelSmith: I added this for BOLT-755 as the most straightforward way to capture the list of files.

lib/bolt/task.rb
```diff
@@ -54,9 +54,12 @@ def select_implementation(target, additional_features = [])
                files.first.dup
              end
 
-      unless (inmethod = metadata['input_method']).nil?
+      unless (inmethod = impl['input_method'] || metadata['input_method']).nil?
```
adreyer:
```inmethod = impl['input_method'] || metadata['input_method']
impl['input_method'] = inmethod unless inmethod.nil?
```
is easier to read IMO
but what happens if the key is present but nil? ie
`impl['input_method'] ||= metadata['input_method']`

MikaelSmith: I didn't want to add a nil entry if neither had it. But this is probably never printede...

PR #658, opened 2018-09-25
(BOLT-816) Enable control repos as Boltdirs

nicklewis:
For existing Puppet users, the control repo is a source of Puppet content
including modules and hiera data. It's helpful to be able to use Bolt from
within that context and pick up all of the content that's deployed in the
Puppet environment. However, doing that currently requires creating a
Boltdir with a bolt.yaml that refers to the hiera config and modulepath
using relative paths like ../modules.

We now look for a bolt.yaml in the current directory (if there is no
Boltdir) and treat the local directory as the Boltdir if one is found. This
allows the defaults for hiera config and modulepath to work when running
Bolt from a standard control repo.

This also fixes an issue where hiera.yaml files containing symbols wouldn't be
loaded. Since many existing hiera.yaml configs do contain symbols, fixing that
is an important part of enabling Bolt usage from a control repo.

### DIFF
--- a/lib/bolt/boltdir.rb
+++ b/lib/bolt/boltdir.rb
@@ -12,13 +12,21 @@ module Bolt
       Boltdir.new(File.join('~', '.puppetlabs', 'bolt'))
     end
 
+    # Search recursively up the directory hierarchy for the Boltdir. Look for a
+    # directory called Boltdir or a file called bolt.yaml (for a control repo
+    # type Boltdir). Otherwise, repeat the check on each directory up the
+    # hierarchy, falling back to the default if we reach the root.
     def self.find_boltdir(dir)
-      local_boltdir = Pathname.new(dir).ascend do |path|
-        boltdir = path + BOLTDIR_NAME
-        break new(boltdir) if boltdir.directory?
+      dir = Pathname.new(dir)
+      if (dir + BOLTDIR_NAME).directory?
+        new(dir + BOLTDIR_NAME)
+      elsif (dir + 'bolt.yaml').file?
+        new(dir)
+      elsif dir.root?
+        default_boltdir
+      else
+        find_boltdir(dir.parent)
       end
-
-      local_boltdir || default_boltdir
     end
 
     def initialize(path)
--- a/lib/bolt/applicator.rb
+++ b/lib/bolt/applicator.rb
@@ -105,7 +105,7 @@ module Bolt
 
     def validate_hiera_config(hiera_config)
       if File.exist?(File.path(hiera_config))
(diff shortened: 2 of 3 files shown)


lib/bolt/boltdir.rb
```diff
@@ -12,13 +12,21 @@ def self.default_boltdir
...
     def self.find_boltdir(dir)
-      local_boltdir = Pathname.new(dir).ascend do |path|
-        boltdir = path + BOLTDIR_NAME
-        break new(boltdir) if boltdir.directory?
+      dir = Pathname.new(dir)
+      if (dir + BOLTDIR_NAME).directory?
+        new(dir + BOLTDIR_NAME)
+      elsif (dir + 'bolt.yaml').file?
+        new(dir)
+      elsif dir.root?
```
adreyer: lets make sure we have a good windows test for finding the default

lib/bolt/boltdir.rb
```diff
@@ -12,13 +12,21 @@ def self.default_boltdir
       Boltdir.new(File.join('~', '.puppetlabs', 'bolt'))
     end
 
+    # Search recursively up the directory hierarchy for the Boltdir. Look for a
+    # directory called Boltdir or a file called bolt.yaml (for a control repo
+    # type Boltdir). Otherwise, repeat the check on each directory up the
+    # hierarchy, falling back to the default if we reach the root.
     def self.find_boltdir(dir)
-      local_boltdir = Pathname.new(dir).ascend do |path|
```
MikaelSmith: Why'd you switch from `ascend`?

nicklewis: It took me a while to wrap my brain around the logic and this was just where I landed. 😄 And it seemed a bit cleaner with multiple conditionals.

The ascend version would be something like this:

```rb
def self.find_boltdir(dir)
  local_boltdir = Pathname.new(dir).ascend do |path|
    if (path + BOLTDIR_NAME).directory?
      break new(path + BOLTDIR_NAME)
    elsif (path + 'bolt.yaml').file?
      break new(path)
    end
  end
                                                     
  local_boltdir || default_boltdir
end
```

Or this:

```rb
def self.find_boltdir(dir)
  Pathname.new(dir).ascend do |path|
    if (path + BOLTDIR_NAME).directory?
      return new(path + BOLTDIR_NAME)
    elsif (path + 'bolt.yaml').file?
      return new(path)
    end
  end
                                                     
  default_boltdir
end
```

lib/bolt/boltdir.rb
nicklewis: Added an explicit test of that.

PR #660, opened 2018-09-25
(BOLT-748) Unvendor Puppet

MikaelSmith:
Bolt no longer vendors gem dependencies. Previously we had vendored
Puppet to allow for more flexibility in development. We're now working
towards Bolt 1.0 and a plan for stability; vendoring Puppet is no longer
worth the cost in complexity and bugs.

Unvendering will fix BOLT-681, as well as simplify using `bolt_spec` by
giving Bolt an explicit dependency on Puppet versions. It may also help
to make it easier to build tools that interact with Bolt since its
dependencies are now visible.


MikaelSmith:
Depends on https://github.com/puppetlabs/bolt-vanagon/pull/53 and a released version of Puppet 6.0.1.

PR #661, opened 2018-09-25
(WIP)(BOLT-841) Add modulepath to task details

lucywyman:
**What** This adds the path of the module that a task is found in to the task output, for both human and json outputs
**Why** It's not always clear where a task is coming from - this gives the user visibility into where tasks are defined


lib/bolt/outputter/human.rb
```diff
@@ -136,6 +136,7 @@ def print_task_info(task)
         task_info << "\n\n"
         task_info << "USAGE:\n#{usage}\n\n"
         task_info << "PARAMETERS:\n#{pretty_params}\n" unless pretty_params.empty?
+        task_info << "MODULE:\n#{task['module']}"
```
MikaelSmith: If we don't get this into Puppet 6.0.1, we could derive this from the path of the task's first file.

lucywyman: Oh true. I forgot that the first file is guaranteed to be the task file.

MikaelSmith: Well, guaranteed to be *a* task file.

lucywyman: Right, but other files in that array can be from other modules, but since the first one is the task itself it'll be from the task module.

lib/bolt/outputter/human.rb
```diff
@@ -8,6 +8,15 @@ class Human < Bolt::Outputter
                  green: "32",
                  yellow: "33" }.freeze
 
+      BUILTIN_MODULES = %w[
```
MikaelSmith: We've previously addressed this with https://github.com/puppetlabs/bolt/blob/master/lib/bolt/cli.rb#L416. We should make sure we don't have 2 methods.

MikaelSmith: There are more built-in modules than are included in this list. I think we should use the previous method.

lib/bolt/outputter/human.rb
```diff
@@ -136,6 +145,10 @@ def print_task_info(task)
         task_info << "\n\n"
         task_info << "USAGE:\n#{usage}\n\n"
         task_info << "PARAMETERS:\n#{pretty_params}\n" unless pretty_params.empty?
+        path = task['files'][0]['path'].chomp("/tasks/#{task['files'][0]['name']}")
+        builtin = path =~ BUILTIN_REGEX
+        task_info << "MODULE:\n#{path}"
+        task_info << " (builtin module)" unless builtin.nil?
```
MikaelSmith: We should probably use a hyphen: `built-in module`.

lib/bolt/outputter/human.rb
```diff
@@ -8,6 +8,15 @@ class Human < Bolt::Outputter
...
                  yellow: "33" }.freeze
 
+      BUILTIN_MODULES = %w[
+        puppet_conf
+        package
+        service
+        puppet_agent
+        facts
+      ].freeze
+      BUILTIN_REGEX = %r{\/bolt(-\d+\.\d+\.\d+)?\/modules\/(?:#{Regexp.union(BUILTIN_MODULES).source})}
```
adreyer: I think we should just find the whole path to this directory and use that rather than this regex. We can also skip looking for specific modules and just consider anything in the directory builtin.

lib/bolt/outputter/human.rb
lucywyman: It looks like that only includes the name of the task, not the path? I'm worried that people will make their own tasks with the same name, and those will accidentally be caught and labelled as builtin.

lib/bolt/outputter/human.rb
lucywyman: Is that directory always the same though, no matter what platform you install on or whether you use `gem install` or package? I had modules in both `~/.puppetlabs/bolt/modules` and then when I installed from a package on a new system the modules were at "/opt/puppetlabs/bolt/lib/ruby/gems/2.4.0/gems/bolt-0.23.0/modules/facts"

MikaelSmith: The partial path is probably pretty good: is it in `gems/bolt-<version>/modules`. The rest can depend on platform and whether it's package or gem installed.

lib/bolt/outputter/human.rb
MikaelSmith: Yeah, that'd be less helpful.

adreyer: Yeah we definitely support that. I think we shouldn't care about the task names at all and just call anything installed in the builtin moduledir a builtin

lib/bolt/outputter/human.rb
adreyer: It will always be the same relative to the rest of bolt right so we should be able to find it absolutely. We rely on that to load it.

lib/bolt/outputter/human.rb
```diff
@@ -136,6 +138,14 @@ def print_task_info(task)
         task_info << "\n\n"
         task_info << "USAGE:\n#{usage}\n\n"
         task_info << "PARAMETERS:\n#{pretty_params}\n" unless pretty_params.empty?
+        task_info << "MODULE:\n"
+
+        path = task['files'][0]['path'].chomp("/tasks/#{task['files'][0]['name']}")
+        task_info << if path.include?(Bolt::PAL::MODULES_PATH)
```
MikaelSmith: Something like `File.realdirpath(path) == File.realdirpath(Bolt::PAL::MODULES_PATH)` might be more robust.

spec/bolt/outputter/human_spec.rb
```diff
@@ -178,6 +191,8 @@
 - foo: Bar
 - baz: Bar
 
+MODULE:
```
MikaelSmith: Can we add a test for built-in modules?

lib/bolt/outputter/human.rb
adreyer: I'd prefer to avoid another assumption that the task is in the top level of the tasks directory. `.start_with?` ?

MikaelSmith: Sure, can drop the chomp and use `start_with?`.

MikaelSmith: Presumably Ruby will be consistent about constructing absolute paths, so we probably don't need extra normalization.

lucywyman:
This also won't modify the JSON output - assuming that's ok ?

MikaelSmith:
That's a good question... I'd lean towards keeping human and json content comparable.

PR #662, opened 2018-09-25
(BOLT-834) Test compatibility with puppet5 and puppet6

donoghuc:
Test that apply works with puppet 5 and 6. Also test that running a ruby task with puppet's ruby works.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -6,6 +6,10 @@ cache:
 matrix:
   fast_finish: true
 
+configuration:
+  - Agentfull
+  - Agentless
+
 environment:
   BOLT_WINRM_USER: roddypiper
   BOLT_WINRM_HOST: localhost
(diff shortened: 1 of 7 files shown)


donoghuc:
Depends on https://github.com/puppetlabs/puppetlabs-puppet_agent/pull/325. Once that gets in I can update the Puppetfile to track the appropriate ref. I have run through acceptance tests on all platforms from CI for the apply_ssh and apply_winrm tests. Note that the osx11 tests are still skipped. I could potentially add some logic to skip ONLY the puppet6 specific tests.

donoghuc:
Keeping the test at the integration level adds as little as 20 seconds and as much as a minute based on the limited data I've seen.

MikaelSmith:
Can be retargeted at puppet_agent#master.

spec/integration/apply_spec.rb
```diff
@@ -38,6 +59,55 @@ def root_config
...
+            expect(result['result']).to eq('ruby' => 'Hi')
+          end
+        end
+      end
+
+      it 'runs an apply plan' do
+        with_tempfile_containing('inventory', YAML.dump(agent_version_inventory), '.yaml') do |inv|
+          results = run_cli_json(%W[plan run basic::notify --nodes agent_targets
+                                    --modulepath #{modulepath} --inventoryfile #{inv.path}])
+          results.each do |result|
```
donoghuc: This was passing earier today on travis and was passing locally. In order to reproduce locally I rebased to upstream/master and destroyed my Gemfile.lock/re-installed. 

The error message (for both puppet5 and 6) for the provided task is: 
```
      "result"=>
       {"_error"=>
         {"kind"=>"puppetlabs.tasks/task-error",
          "issue_code"=>"TASK_ERROR",
          "msg"=>
           "The task failed with exit code 1:\n/opt/puppetlabs/puppet/lib/ruby/v
endor_ruby/puppet/configurer.rb:54:in `rescue in rescue in init_storage': Cannot
 remove /tmp/d20181003-792-xynxz5/state/state.yaml: No such file or directory @ 
unlink_internal - /tmp/d20181003-792-xynxz5/state/state.yaml (Puppet::Error)\n\t
from /opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/configurer.rb:50:in `res
cue in init_storage'\n\tfrom /opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/
configurer.rb:45:in `init_storage'\n\tfrom /opt/puppetlabs/puppet/lib/ruby/vendo
r_ruby/puppet/configurer.rb:205:in `run'\n\tfrom /tmp/09555e1f-b723-4189-8264-d4
1c049c312a/apply_catalog.rb:81:in `block in <main>'\n\tfrom /opt/puppetlabs/pupp
et/lib/ruby/vendor_ruby/puppet/context.rb:65:in `override'\n\tfrom /opt/puppetla
bs/puppet/lib/ruby/vendor_ruby/puppet.rb:260:in `override'\n\tfrom /tmp/09555e1f-b723-4189-8264-d41c049c312a/apply_catalog.rb:69:in `<main>'\n",
          "details"=>{"exit_code"=>1}}}},
```

docker-compose.yml
```diff
@@ -4,3 +4,11 @@ services:
     build: .
     ports:
       - "20022:22"
+  puppet_5_node:
```
MikaelSmith: I think this compose file can be setup so it only builds a single image, then starts 3 instances of it.

MikaelSmith: https://stackoverflow.com/a/35662191/2048059 is one possibility of how to do it.

spec/integration/apply_spec.rb
```diff
@@ -38,6 +59,55 @@ def root_config
       run_command(uninstall, uri, config: config_data, inventory: inventory_data)
     end
 
+    context "when running against puppet 5 or puppet 6" do
```
MikaelSmith: We should also clean up (uninstall) packages so we can re-run tests. Or document that docker images need to be recreated if you want to re-run tests (which seems like a bad experience).

MikaelSmith: I guess testing install is not really an important part. We could make this part of building the images; or just document that tests will install to these agents initially, then re-use the installed agent.

spec/integration/apply_spec.rb
donoghuc: Issue appears to only arises when running apply plan as non-root user.

spec/integration/apply_spec.rb
```diff
@@ -38,6 +59,55 @@ def root_config
...
+      it 'installs puppet agent' do
```
MikaelSmith: This would be better as a `before(:all)`.

MikaelSmith:
This can be rebased now.

spec/integration/apply_spec.rb
```diff
@@ -103,4 +173,98 @@ def root_config
...
+      { 'modulepath' => File.join(__dir__, '../fixtures/apply'),
+        'winrm' => {
+          'ssl' => false,
+          'ssl-verify' => false,
+          'user' => conn_info('winrm')[:user],
+          'password' => conn_info('winrm')[:password]
+        } }
+    end
+
+    context "when running against puppet 5" do
```
MikaelSmith: Are you relying on the puppet_agent::install task doing an upgrade here?

MikaelSmith: Also, with Windows tests. I think we can probably leave out the acceptance tests. These integration tests should be enough to catch incompatibilities in Puppet.

donoghuc: From manual testing on vmpooler instances when installing from an MSI or DMG the install task can handle moving back and forth between puppet 5 and 6 collections. It is only the linux package managers that needed special handling for the upgrade (and will fail to downgrade) across collections.

appveyor.yml
```diff
@@ -17,7 +17,7 @@ environment:
     - RUBY_VERSION: 25
 
 install:
-  - set PATH=C:\Ruby%RUBY_VERSION%\bin;%PATH%
+  - set PATH="C:\Program Files\Puppet Labs\Puppet\bin";C:\Ruby%RUBY_VERSION%\bin;%PATH%
```
donoghuc: Still getting the following error:
```
{"kind"=>"bolt/apply-failure", "msg"=>"Found a Ruby without Puppet present, please install Puppet or remove Ruby from $env:Path to enable 'apply'", "details"=>{"action"=>"apply", "object"=>"catalog", "result_set"=>[{"node"=>"winrm://roddypiper@localhost:5985", "status"=>"failure", "result"=>{"report"=>{"_output"=>""}, "_error"=>{"msg"=>"Found a Ruby without Puppet present, please install Puppet or remove Ruby from $env:Path to enable 'apply'", "kind"=>"bolt/apply-error"}}}]}}
```

appveyor.yml
```diff
@@ -23,47 +27,98 @@ install:
...
-            Invoke-WebRequest -Uri 'https://curl.haxx.se/ca/cacert.pem' -UseBasicParsing -OutFile $CACertFile | Out-Null
+for:
+  -
+    matrix:
+      exclude:
+        - configuration: Agentless
+        - configuration: Agentfull
+          RUBY_VERSION: 23
+        - configuration: Agentfull
+          RUBY_VERSION: 24
```
MikaelSmith: To avoid massive pile-ups in AppVeyor, I think we should drop Ruby 2.4 support. Ruby 2.3 and 2.5 seems sufficient.

donoghuc: Absolutely. Working on that now.

PR #664, opened 2018-09-27
(BOLT-749) Remove support for legacy bolt.yaml location

adreyer:
~/.puppetlabs/bolt.yaml(yml) has been deprecated for a few releases and
is removed with this commit.

### DIFF
--- a/lib/bolt/config.rb
+++ b/lib/bolt/config.rb
@@ -60,17 +60,7 @@ module Bolt
     end
 
     def self.from_boltdir(boltdir, overrides = {})
-      # *Optionally* load the boltdir config file, and fall back to the legacy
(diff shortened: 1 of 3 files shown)


lib/bolt/config.rb
```diff
@@ -60,17 +60,7 @@ def self.default
     end
 
     def self.from_boltdir(boltdir, overrides = {})
-      # *Optionally* load the boltdir config file, and fall back to the legacy
-      # config if that isn't found. Because logging is built in to the
-      # legacy_conf method, we don't want to look that up unless we need it.
-      configs = if boltdir.config_file.exist?
```
nicklewis: I'm so glad we can get rid of this

PR #667, opened 2018-09-28
(BOLT-863) Apply should return an ApplyResult not a Result

adreyer:
This adds a new datatype ApplyResult that is returned in the ResultSet
of an apply statement. This is intended to provide a more futureproof
return value so we can add more data and helper methods in the future.


lib/bolt/applicator.rb
```diff
@@ -201,7 +196,17 @@ def apply(args, apply_body, scope)
               raise future.reason if future.rejected?
               results = transport.batch_task(batch, catalog_apply_task, arguments, options, &notify)
               Array(results).map do |result|
-                identify_resource_failures(provide_puppet_missing_errors(result))
+                if (puppet_missing = create_puppet_missing_error(result))
```
adreyer: Should this be a class method on the ApplyResult
`ApplyResult.from_task_result(result)`
?

MikaelSmith: That seems reasonable.

MikaelSmith:
The refactor looks useful, and rubocop. Otherwise this makes sense.

spec/bolt/applicator_spec.rb
```diff
@@ -89,51 +89,6 @@
     )
   end
 
-  describe '#provide_puppet_missing_errors' do
```
MikaelSmith: It looks like these tests were completely removed.

bolt-modules/boltlib/lib/puppet/datatypes/resultset.rb
```diff
@@ -3,14 +3,14 @@
 Puppet::DataTypes.create_type('ResultSet') do
   interface <<-PUPPET
     attributes => {
-      'results' => Array[Result],
+      'results' => Variant[Array[Result], Array[ApplyResult]],
```
nicklewis: Is there ever a time we might want to mix Results and ApplyResults? Perhaps if you were trying to return a ResultSet that contained the results of a set of operations that were taken on a single node?

MikaelSmith: I agree, we should let ResultSet contain a mix of Result and ApplyResult, i.e. `Array[Variant[Result, ApplyResult]]`.

adreyer: I think that makes it harder to iterate on since they don't have similar methods. My thought was we could change this as needed but I guess which direction is a more breaking change?

spec/bolt/applicator_spec.rb
adreyer: oh I forgot to add the file they were moved to

bolt-modules/boltlib/lib/puppet/datatypes/resultset.rb
MikaelSmith: I think enough methods are shared, and you can use a case statement to differentiate types (I think). But going from this to mixed result types in a resultset is the non-breaking change.

adreyer: harder to iterate over the ResultSet

nicklewis: I think it's reasonable to keep it this way and extend it later. That will be a backward-compatible change.

PR #669, opened 2018-09-28
(BOLT-413) Don't expand type alias names in bolt plan show

nicklewis:
Previously, using `bolt plan show <plan-name>` for a plan with a
parameter of type TargetSpec would display the parameter type as

    TargetSpec = Boltlib::TargetSpec = Variant[String[1], Object[{name => 'Target', attributes => {'uri' => String[1], 'options' => {type => Hash[String[1], Data], value => {}}}, functions => {'host' => Callable[[0, 0], String[1]], 'name' => Callable[[0, 0], String[1]], 'password' => Callable[[0, 0], Optional[String[1]]], 'port' => Callable[[0, 0], Optional[Integer]], 'protocol' => Callable[[0, 0], Optional[String[1]]], 'user' => Callable[[0, 0], Optional[String[1]]]}}], Array[Boltlib::TargetSpec]]

Which is unhelpful.

We now explicitly check for alias types and just display the type name,
rather than the full string representation.

### DIFF
--- a/lib/bolt/pal.rb
+++ b/lib/bolt/pal.rb
@@ -241,7 +241,12 @@ module Bolt
     def plan_hash(plan_name, plan)
       elements = plan.params_type.elements || []
       parameters = elements.each_with_object({}) do |param, acc|
-        acc[param.name] = { 'type' => param.value_type }
+        type = if param.value_type.is_a?(Puppet::Pops::Types::PTypeAliasType)
+                 param.value_type.name
+               else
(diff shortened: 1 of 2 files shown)


MikaelSmith:
This didn't actually have anything to do with the change to how we pass tasks from PAL, but I'm glad we found a solution.

MikaelSmith:
spec/integration/parsing_spec.rb:41 has a test you missed that already covered this.

PR #671, opened 2018-10-01
(maint) Update acceptance test for apply() result

donoghuc:


acceptance/tests/apply_ssh.rb
```diff
@@ -160,9 +160,9 @@
         # Verify that node succeeded
         host = node.hostname
         result = json.select { |n| n['node'] == host }
-        assert_equal('failure', result[0]['status'],
-                     "The task did not fail on #{host}")
-        assert_match(/Permission denied/, result[0]['result']['_error']['msg'])
+        assert_equal('success', result[0]['status'],
```
adreyer: There is an actual bug here. We shouldn't change this

PR #673, opened 2018-10-01
(BOLT-883) Redirect all Puppet input and output to temp dir

MikaelSmith:
Bolt Apply still picked up settings from `puppet.conf`, which can lead
to confusing failures. Update `apply_catalog` to override required app
settings - similar to how Bolt sets up Puppet - to point to a temp
directory. The result should be that we ignore any config on the system,
write any logs and runtime state to that temporary directory, and read
environments from that location (i.e. the only available environment
will be the built-in `production`).

### DIFF
--- a/libexec/apply_catalog.rb
+++ b/libexec/apply_catalog.rb
@@ -10,7 +10,14 @@ require 'tempfile'
 
 args = JSON.parse(ARGV[0] ? File.read(ARGV[0]) : STDIN.read)
 
-Puppet.initialize_settings([])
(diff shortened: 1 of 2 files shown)


MikaelSmith:
I'm trying to figure out how to make better tests for this. I can lay down a puppet.conf with a different environment and make sure we don't read it in integration tests.

MikaelSmith:
@adreyer is working on some tests in this area, so going to hold off adding them for now.

PR #674, opened 2018-10-01
(maint) Add integration tests for apply

adreyer:
This adds some basic integration tests for the apply statement. It moves
apply_spec to apply_compile spec since it confusingly mocks out actually
catalog compilation.


spec/integration/apply_compile_spec.rb
```diff
@@ -0,0 +1,244 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+require 'bolt_spec/conn'
+require 'bolt_spec/files'
+require 'bolt_spec/integration'
+require 'bolt/catalog'
+require 'bolt/task'
+
+describe "passes parsed AST to the apply_catalog task" do
```
adreyer: IMO these tests are less full integration tests and more targeted integration tests at the applicator->compiler boundary. We should probably mock out all of run_task in them and move them into spec/bolt/applicator_spec.rb.

I'm going to leave that for a later PR so we can get some real integration tests for apply in quickly

PR #679, opened 2018-10-03
(BOLT-879) Add note to docs about installing kerberos

lucywyman:
**What this changes** Adds a note to the docs that the user needs to install the kerberos gem on their own
**Why** The license for kerberos is incompatible with other components we distribute, so we need to remove it from our packaging and users who need it need to install it on their own.

### DIFF
--- a/pre-docs/bolt_installing.md
+++ b/pre-docs/bolt_installing.md
@@ -209,4 +209,10 @@ To disable the collection of analytics data add the following line to `~/.puppe
 ```
 disabled: true
 ```
+## A Note on Kerberos over SSH
+
+We support using Kerberos over SSH for authentication, but due to license incompatibility with other components we are distributing you'll need to install the gem yourself. You can do this with
+```
+gem install net-ssh-krb
+```
 
--- a/lib/bolt/transport/ssh.rb
+++ b/lib/bolt/transport/ssh.rb
@@ -55,7 +55,8 @@ module Bolt
           require 'net/ssh/krb'
         rescue LoadError
           logger.debug {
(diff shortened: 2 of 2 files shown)


lucywyman:
It looks like the logic around the gem not being installed is already there, and we've already removed it from our packaging. Let me know if there's any other work we need to do that I'm missing.

README.md
```diff
@@ -43,6 +43,13 @@ Additionally the Bolt project includes:
     * [Writing Plans](https://puppet.com/docs/bolt/latest/writing_plans.html)
 * [Applying Manifest Blocks in Plans](https://puppet.com/docs/bolt/latest/applying_manifest_blocks.html)
 
+## A Note on Kerberos
+
+We support using [Kerberos]() for authentication, but due to license incompatibility with other components we are distributing you'll need to install the gem yourself. You can do this with
```
MikaelSmith: I don't think I'd include this in the README at all. I think it should go in the install docs as a note at the end.

pre-docs/bolt_installing.md
```diff
@@ -209,4 +209,10 @@ To disable the collection of analytics data add the following line to `~/.puppe
 ```
 disabled: true
 ```
+## A Note on Kerberos
+
+We support using [Kerberos]() for authentication, but due to license incompatibility with other components we are distributing you'll need to install the gem yourself. You can do this with
```
MikaelSmith: Why is `Kerberos` a link?

MikaelSmith: Also, this needs to be more specific: `We support using Kerberos over SSH for authentication,`.

PR #680, opened 2018-10-03
(BOLT-883) Override settings more directly

MikaelSmith:
Without this, we get errors like
```
The task failed with exit code 1:
/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/configurer.rb:54:in `rescue in rescue in init_storage': Cannot remove /tmp/d20181003-855-156fn3k/state/state.yaml: No such file or directory @ unlink_internal - /tmp/d20181003-855-156fn3k/state/state.yaml (Puppet::Error)
	from /opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/configurer.rb:50:in `rescue in init_storage'
	from /opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/configurer.rb:45:in `init_storage'
	from /opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/configurer.rb:205:in `run'
	from /tmp/c33af27e-1acd-495c-a495-304fd058ded3/apply_catalog.rb:81:in `block in <main>'
	from /opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/context.rb:65:in `override'
	from /opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet.rb:260:in `override'
	from /tmp/c33af27e-1acd-495c-a495-304fd058ded3/apply_catalog.rb:69:in `<main>'
```

Ensure directories are created from main settings, similar to how `puppet apply` performs setup.

### DIFF
--- a/libexec/apply_catalog.rb
+++ b/libexec/apply_catalog.rb
@@ -12,40 +12,31 @@ require 'tempfile'
 args = JSON.parse(ARGV[0] ? File.read(ARGV[0]) : STDIN.read)
 
 # Create temporary directories for all core Puppet settings so we don't clobber
-# existing state or read from puppet.conf
+# existing state or read from puppet.conf. Also create a temporary modulepath.
+# Additionally include rundir, which gets its own initialization.
 puppet_root = Dir.mktmpdir
 moduledir = File.join(puppet_root, 'modules')
 Dir.mkdir(moduledir)
-cli = Puppet::Settings::REQUIRED_APP_SETTINGS.flat_map { |setting| ["--#{setting}", puppet_root] }
+cli = (Puppet::Settings::REQUIRED_APP_SETTINGS + [:rundir]).flat_map do |setting|
+  ["--#{setting}", File.join(puppet_root, setting.to_s.chomp('dir'))]
+end
 cli << '--modulepath' << moduledir
 Puppet.initialize_settings(cli)
-run_mode = Puppet::Util::RunMode[:user]
-Puppet.settings.initialize_app_defaults(Puppet::Settings.app_defaults_for_run_mode(run_mode))
-
-Puppet::ApplicationSupport.push_application_context(run_mode)
 
 # Avoid extraneous output
-Puppet[:summarize] = false
 Puppet[:report] = false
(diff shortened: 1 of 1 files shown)


MikaelSmith:
Shouldn't that happen with https://github.com/puppetlabs/puppet/blob/master/lib/puppet/util/storage.rb#L45?

nicklewis:
Ah yeah it should. That does make this a bit more mysterious.

libexec/apply_catalog.rb
```diff
@@ -16,6 +16,7 @@
 puppet_root = Dir.mktmpdir
 moduledir = File.join(puppet_root, 'modules')
 Dir.mkdir(moduledir)
+Dir.mkdir(File.join(puppet_root, 'state'))
 cli = Puppet::Settings::REQUIRED_APP_SETTINGS.flat_map { |setting| ["--#{setting}", puppet_root] }
 cli << '--modulepath' << moduledir
 Puppet.initialize_settings(cli)
```
MikaelSmith: Should we include `Puppet.settings.use :main`?

MikaelSmith:
Ok, I think this is better now.

MikaelSmith:
Something deeper's going wrong here.

MikaelSmith:
I've tested this with https://github.com/puppetlabs/bolt/pull/662, which gives me more confidence it'll work.

libexec/apply_catalog.rb
```diff
@@ -13,34 +13,22 @@
 
 # Create temporary directories for all core Puppet settings so we don't clobber
 # existing state or read from puppet.conf
+Puppet.initialize_settings
 puppet_root = Dir.mktmpdir
+Puppet::Settings::REQUIRED_APP_SETTINGS.each do |setting|
+  Puppet[setting] = File.join(puppet_root, setting.to_s.chomp('dir'))
+end
```
adreyer: I think we should explicitly call puppet.settings.use( ??? ) to make it easier to understand and avoid regressions.

Not sure exactly where that should happen. before line 25 maybe?

PR #681, opened 2018-10-03
(BOLT-866) Log results per-node at info level and improve Apply logging

adreyer:


lib/bolt/applicator.rb
```diff
@@ -134,29 +134,43 @@ def apply(args, apply_body, scope)
...
+          end
+        end
+
+        log_proc = proc do |event|
+          if event[:type] == :node_result
+            result = event[:result]
+            # If the catalog finished there will be a message.
+            if result.message
+              status = result.ok? ? "Finished" : "Failed"
+              log.info("#{status} apply on #{result.target.host}: #{result.message}")
```
adreyer: This is a deviation from the other logging since it only logs a human readable result at "info". run_* functions all log the actual result json at info. Long term I think we need to support "format" for these and log human readable to stderr and json to disk or when format: json is specified.

PR #682, opened 2018-10-03
(docs) Minor fixes to docs around `bolt.yml` and using `gem` from packages

MikaelSmith:


pre-docs/bolt_installing.md
```diff
@@ -216,3 +216,6 @@ We support using Kerberos over SSH for authentication, but due to license incomp
 gem install net-ssh-krb
 ```
 
+If using `puppet-bolt` packages, the `gem` program can be found
```
adreyer: I think this needs it's own section in installing modules

MikaelSmith: Something similar already exists in https://github.com/puppetlabs/bolt/blob/master/pre-docs/applying_manifest_blocks.md#using-hiera-data-in-a-manifest-block. This seems most relevant when installing Bolt, and we could link back to this section from the others.

MikaelSmith: @katelopresti do you have thoughts on where this should live?

MikaelSmith:
Updated.

pre-docs/applying_manifest_blocks.md
```diff
@@ -168,9 +168,7 @@ Manifest block compilation can access Hiera data that you add to your Bolt confi
 
 Following the Hiera 5 convention, the default data dir is relative to `hiera.yaml` at `$BOLTDIR/data`. For config file examples, see https://puppet.com/docs/puppet/5.5/hiera_config_yaml_5.html.
 
-If a custom data provider is used (such as `hiera-eyaml`, which allows you to encrypt your data) the gem dependencies must be available to Bolt. If using `puppet-bolt` packages and installing `hiera-eyaml`
```
MikaelSmith: @adreyer so it looks like we should just revert to the previous language here. And you wanted essentially the same thing in the bolt_installing_modules dco.

adreyer: Yeah that sounds correct. We should target docs at users who installed with the package.

I'm ok only actually having the specific commands in one place. Installing modules is where i would put it but I'm fine with it just being on the installing bolt page and linking there from elsehwere.

pre-docs/applying_manifest_blocks.md
```diff
@@ -168,9 +168,7 @@ Manifest block compilation can access Hiera data that you add to your Bolt confi
...
-- on Windows, run `"C:/Program Files/Puppet Labs/Bolt/bin/gem.bat" install hiera-eyaml`
-- on other platforms, run `/opt/puppetlabs/bolt/bin/gem install hiera-eyaml`
+If a custom data provider is used (such as `hiera-eyaml`, which allows you to encrypt your data) the gem dependencies must be available to Bolt (for example `gem install hiera-eyaml`). See [Installing Gems with Bolt Packages](bolt_installing.md#installing-gems-with-bolt-packages).
```
adreyer: do we really want the example?

MikaelSmith: I can remove it. I guess it's better to point people to the full command only.

PR #683, opened 2018-10-03
(BOLT-900) Allow deprecations in functions used in a manifest block

MikaelSmith:
We want Bolt to be able to compile most Puppet 4+ manifests, so we need to allow deprecated functions. TODO: this should be configurable.

### DIFF
--- a/lib/bolt/catalog.rb
+++ b/lib/bolt/catalog.rb
@@ -53,6 +53,7 @@ module Bolt
       pal_main = request['code_ast'] || request['code_string']
       target = request['target']
       pdb_client = Bolt::PuppetDB::Client.new(Bolt::PuppetDB::Config.new(request['pdb_config']))
+      options = request['puppet_config'] || {}
 
       with_puppet_settings(request['hiera_config']) do
         Puppet[:rich_data] = true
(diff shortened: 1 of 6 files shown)


lib/bolt/catalog.rb
```diff
@@ -65,6 +65,9 @@ def compile_catalog(request)
                           bolt_inventory: setup_inventory(request['inventory'])) do
             Puppet.lookup(:pal_current_node).trusted_data = target['trusted']
             pal.with_catalog_compiler do |compiler|
+              # Override undesirable settings in the CatalogCompiler. We want Bolt to be able
+              # to compile most Puppet 4+ manifests, so we need to allow deprecated functions.
+              Puppet[:strict] = :warning
```
nicklewis: Will this revert back to `:error` at the end of this block?

MikaelSmith:
I want to add tests, and proposed making these things configurable.

lib/bolt/catalog.rb
MikaelSmith: I don't think so. We use `PAL#with_catalog_compiler`, which calls https://github.com/puppetlabs/puppet/blob/master/lib/puppet/parser/catalog_compiler.rb#L15-L16 first. I don't think it resets afterwards.

PR #684, opened 2018-10-03
(BOLT-898) Include agent facts in apply_prep

MikaelSmith:
Make custom_facts more like apply_catalog facts gathering. Use a
temporary directory for isolation, and ensure agent facts like
clientcert are included.


MikaelSmith:
Ah, quite possibly.

PR #687, opened 2018-10-04
Test using 2 Windows Server 2016 docker containers

MikaelSmith:


appveyor.yml
```diff
@@ -49,6 +49,9 @@ before_test:
       ($cert = Import-PfxCertificate -FilePath resources/cert.pfx -CertStoreLocation cert:\\LocalMachine\\My -Password (ConvertTo-SecureString -String bolt -Force -AsPlainText)) | Format-List
       New-WSManInstance -ResourceURI winrm/config/Listener -SelectorSet @{Address='*';Transport='HTTPS'} -ValueSet @{Hostname='localhost';CertificateThumbprint=$cert.Thumbprint} | Format-List
 
+      docker run -d -p 35985:5985 microsoft/windowsservercore
```
MikaelSmith: Networking with docker-compose on Windows was unreliable, but we don't need it. So created two individual instances.

MikaelSmith:
I don't think this is done yet, not sure if we can interact with the containers.

appveyor.yml
donoghuc: Do you have a windows test environment with bolt and docker configured to test this out? If so did you do that on a vmpooler instance?

Vagrantfile
```diff
@@ -46,6 +46,28 @@ Vagrant.configure('2') do |config|
...
+    windows.vm.communicator = "winrm"
+    windows.vm.network :forwarded_port, guest: 5985, host: 35985, host_ip: '127.0.0.1', id: 'winrm'
+    windows.vm.provision 'shell', privileged: true, inline: 'slmgr /rearm'
+    windows.vm.provider "virtualbox" do |vb|
+      vb.gui = false
+    end
+  end
+
+  config.vm.define :windows6 do |windows|
+    windows.vm.box = "jacqinthebox/windowsserver2016core"
```
donoghuc: I am connected to the ethernet at the office and downloading this image has been in progress for almost an hour. Estimated 5% done with projected 32 hours left. I guess I will just let it go over-night and test it out tomorrow.

PR #689, opened 2018-10-05
Sync DITA output with 0.25.0 Markdown files.

katelopresti:


.gitignore
```diff
@@ -42,6 +43,12 @@ modules/puppet_agent
 modules/puppet_conf
 modules/service
 
+<<<<<<< HEAD
+# OS generated files #
+######################
+.DS_Store
```
MikaelSmith: This is better to put in `~/.gitignore_global`, since it's OS-specific and will show up pretty much everywhere.

PR #692, opened 2018-10-08
(BOLT-914) Clean up Bolt-as-a-library includes

MikaelSmith:
When loading Bolt as a library, the main entry point is `bolt/executor`.
It previously required Config and CLI, both of which are unnecessary for
the Executor. Clean up config so `bolt` and `bolt/executor` don't bring
in unused files, especially the CLI (which loads r10k and is not as
careful about side-effects).


MikaelSmith:
Ran through acceptance tests as well.

MikaelSmith:
Tested that it fixes https://github.com/puppetlabs/puppetlabs-reboot/pull/171 as well.

spec/integration/apply_spec.rb
```diff
@@ -54,7 +54,7 @@ def root_config
 
         # Includes agent facts from apply_prep
         agent_facts = report['resource_statuses']['Notify[agent facts]']['events'][0]['desired_value'].split("\n")
-        expect(agent_facts[0]).to match(/^\w+\./)
+        expect(agent_facts[0]).to match(/^\w+/)
```
MikaelSmith: This changed when I rebuilt my docker container. Not quite sure why, but might have to do with what my host network configuration looks like when it's rebuilt.

adreyer:
can we try to run the puppet_agent package tests with this commit

MikaelSmith:
Tested both https://github.com/puppetlabs/bolt/pull/662 and https://github.com/puppetlabs/puppetlabs-puppet_agent/tree/master/task_spec.

PR #693, opened 2018-10-08
(BOLT-755) Add shareable task code

MikaelSmith:
Add shareable task code in the form of the `files` metadata property. Implemented specifically for the SSH, WinRM, and Local transports; the PCP transport doesn't use it because Bolt only sends the name of the task, and Orchestrator handles the rest.

### DIFF
--- a/lib/bolt/task.rb
+++ b/lib/bolt/task.rb
@@ -58,19 +58,17 @@ module Bolt
       impl['input_method'] = inmethod unless inmethod.nil?
 
       mfiles = impl.fetch('files', []) + metadata.fetch('files', [])
-      unless mfiles.empty?
-        dirnames, filenames = mfiles.partition { |file| file.end_with?('/') }
-        impl['files'] = filenames.map do |file|
-          path = file_map[file]
-          raise "No file found for reference #{file}" if path.nil?
-          { 'name' => file, 'path' => path }
-        end
+      dirnames, filenames = mfiles.partition { |file| file.end_with?('/') }
+      impl['files'] = filenames.map do |file|
+        path = file_map[file]
+        raise "No file found for reference #{file}" if path.nil?
+        { 'name' => file, 'path' => path }
+      end
 
-        unless dirnames.empty?
-          files.each do |file|
-            if dirnames.any? { |dirname| file['name'].start_with?(dirname) }
-              impl['files'] << file
-            end
+      unless dirnames.empty?
+        files.each do |file|
+          if dirnames.any? { |dirname| file['name'].start_with?(dirname) }
+            impl['files'] << file
           end
         end
       end
--- a/lib/bolt/task.rb
+++ b/lib/bolt/task.rb
@@ -58,7 +58,22 @@ module Bolt
       impl['input_method'] = inmethod unless inmethod.nil?
 
       mfiles = impl.fetch('files', []) + metadata.fetch('files', [])
-      impl['files'] = mfiles.map { |file| { 'name' => file, 'path' => file_map[file] } } unless mfiles.empty?
+      unless mfiles.empty?
+        dirnames, filenames = mfiles.partition { |file| file.end_with?('/') }
+        impl['files'] = filenames.map do |file|
+          path = file_map[file]
+          raise "No file found for reference #{file}" if path.nil?
+          { 'name' => file, 'path' => path }
+        end
+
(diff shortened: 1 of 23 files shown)


lib/bolt/transport/winrm.rb
```diff
@@ -116,32 +116,44 @@ def run_task(target, task, arguments, _options = {})
...
-
           conn.with_remote_tempdir do |dir|
             remote_task_path = if from_api?(task)
                                  conn.write_executable_from_content(dir, file_content, executable)
                                else
                                  conn.write_remote_executable(dir, executable)
                                end
+
+            if extra_files
+              conn.mkdirs(extra_files.map { |file| File.join(dir, '_installdir', File.dirname(file['name'])) })
```
adreyer: I'm always surprised this works when running from windows.

lib/bolt/transport/local.rb
```diff
@@ -83,14 +83,24 @@ def run_task(target, task, arguments, _options = {})
...
-
-        # unpack any Sensitive data, write it to a separate variable because
-        # we log 'arguments' below
-        unwrapped_arguments = unwrap_sensitive_args(arguments)
-        stdin = STDIN_METHODS.include?(input_method) ? JSON.dump(unwrapped_arguments) : nil
-        env = ENVIRONMENT_METHODS.include?(input_method) ? envify_params(unwrapped_arguments) : nil
+        extra_files = implementation['files']
 
         with_tmpscript(executable, target.options['tmpdir']) do |script, dir|
+          if extra_files
```
adreyer: can we create an installdir variable first. I think that would make this slightly more readable.

MikaelSmith:
Those tests mostly exercise Puppet. They make sense as integration tests; I'll work on em.

lib/bolt/task.rb
```diff
@@ -58,7 +58,22 @@ def select_implementation(target, additional_features = [])
       impl['input_method'] = inmethod unless inmethod.nil?
 
       mfiles = impl.fetch('files', []) + metadata.fetch('files', [])
-      impl['files'] = mfiles.map { |file| { 'name' => file, 'path' => file_map[file] } } unless mfiles.empty?
+      unless mfiles.empty?
```
adreyer: does this check matter? It seems like we can rid of one level indentation otherwise.

lib/bolt/transport/local.rb
```diff
@@ -83,14 +83,24 @@ def run_task(target, task, arguments, _options = {})
...
+            FileUtils.mkdir_p(extra_files.map { |file| File.join(dir, '_installdir', File.dirname(file['name'])) })
+            arguments['_installdir'] = File.join(dir, '_installdir')
```
adreyer: I don't think we're copying the task itself into the correct location. it looks like it will be outside the installdir only

lib/bolt/task.rb
MikaelSmith: It affects other tests, and consumers will need to change slightly. But I initially did go without it.

lib/bolt/transport/local.rb
MikaelSmith: Oh, is the task supposed to be in installdir? I don't think we defined that.

MikaelSmith: Pretty sure pxp-agent doesn't put it in installdir either.

adreyer: that was the original design but I'm not sure it's clear from the spec.

spec/bolt/transport/local_spec.rb
```diff
@@ -259,6 +260,48 @@ def result_value(stdout = nil, stderr = nil, exit_code = 0)
       end
     end
 
+    context "when files are provided", ssh: true do
```
adreyer: should this be bash: true? now?

pre-docs/writing_tasks.md
```diff
@@ -274,6 +274,120 @@ myservice/tasks/stop.json
...
+#### Task
+
+```python
+#!/usr/bin/env python
+import sys
+import os
+import json
+
+params = json.load(sys.stdin)
+sys.path.append(os.path.join(params['_installdir'], 'python_helpers/files'))
```
adreyer: does this need to be 
os.path.join(params['_installdir'], 'python_helpers', 'files')

PR #702, opened 2018-10-11
(maint) add Dockerfile for bolt-server

adreyer:
This adds a Dockerfile to run boltserver. It also moves the testing
containers to spec/containers to make the build process easier.


adreyer:
@davejohnston I think this is closer to what you need. I'm going to make some tweaks to how bolt loads config though

developer-docs/bolt-server.md
```diff
@@ -174,7 +174,64 @@ For example:
...
+export BOLT_ROOT=https://localhost:62658
+```
+
+you can now make a curl request to bolt which should have an empty response
+
+```
+curl -v --cacert $BOLT_CACERT --cert $BOLT_CERT --key $BOLT_KEY $BOLT_ROOT
+```
+
+## Run from a Package on A VM
```
adreyer: I don't think this section is particulairly useful anymore. If you want a vm you probably want the rest of PE as well. Should I delete it?

developer-docs/bolt-server.md
```diff
@@ -174,7 +174,64 @@ For example:
 If the task runs the response will have status 200.
 The response will be a standard bolt Result JSON object.
 
-## Install from Source
+## Running bolt-server
```
lucywyman: Not sure we need this heading

developer-docs/bolt-server.md
```diff
@@ -174,7 +174,64 @@ For example:
...
+## Running in a container
+
+from your checkout of bolt build and run the container with
+
+```
+docker run --rm -p 62658:62658 `docker build -q ./`
+```
+
+```
+export BOLT_CACERT=spec/fixtures/ssl/ca.pem
```
lucywyman: Looks like these got copy-pasted twice?

developer-docs/bolt-server.md
lucywyman: Or include a link to the PE docs for setting up a test environment.

developer-docs/bolt-server.md
```diff
@@ -223,12 +280,30 @@ bolt-server: {
 
 7. Start bolt-server
 `service pe-bolt-server start`
-8. Build a request
-Save the following JSON to `~/request.json`
+
+8. setup your environment for running commands with
```
lucywyman: s/setup/Setup/ for consistency with other numbered items.

lib/bolt_ext/server_config.rb
```diff
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require 'hocon'
+require 'bolt/error'
```
lucywyman: Looks like this isn't used in this file?

developer-docs/bolt-server.md
adreyer: it is for ValidationError

lib/bolt_ext/server_config.rb
MikaelSmith: It is for `Bolt::ValidationError`.

Gemfile
```diff
@@ -6,14 +6,17 @@ source ENV['GEM_SOURCE'] || 'https://rubygems.org'
...
 gemspec
-gem "hocon"
-gem "json-schema"
-gem "puma"
-gem "rack"
-gem "rails-auth"
-gem "sinatra"
+
+# Bolt server gems are managed here not in the gemspec
+gem "hocon", '>= 1.2.5'
```
MikaelSmith: was considering putting this in a gem group to make that clearer.

developer-docs/bolt-server.md
```diff
@@ -174,61 +174,61 @@ For example:
...
+you can now make a curl request to bolt which should have an empty response
+
 ```
-/etc/puppetlabs/puppet/ssl/certs/$HOSTNAME.pem
-/etc/puppetlabs/puppet/ssl/private_keys/$HOSTNAME.pem
+curl -v --cacert $BOLT_CACERT --cert $BOLT_CERT --key $BOLT_KEY $BOLT_ROOT
 ```
-3. Download and install bolt-server. The latest release can be found at http://builds.delivery.puppetlabs.net/pe-bolt-server
+
+## Running in a container
```
MikaelSmith: We should document where the certs are mounted in the container, in case people want to provide their own.

adreyer: I don't think that should block merging this in. It's solely intended for a handful of dev/test usecases. I don't think anything I'd document now would be stable over the next month so it's best to leave it up to the users to come up with new usecases and improve the bolt-server code, container, and docs as needed.

Dockerfile
```diff
@@ -1,25 +1,12 @@
...
+ENV BOLT_SERVER_CONF /bolt-server/config/docker.conf
 
-# Add test user without authorized key and different login shell
-RUN useradd test && echo "test:test" | chpasswd && adduser test sudo
-RUN echo test | chsh -s /bin/bash test
-RUN mkdir -p /home/test/
-RUN chown -R test:sudo /home/test
-
-CMD ["/usr/sbin/sshd", "-D"]
+ENTRYPOINT bundle exec puma -C puma_config.rb
```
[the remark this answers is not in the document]
adreyer: What difference does it make. This is really intended for development so setting the production environment is kind of a red flag.

Gemfile
adreyer: I've never found gem groups to be that useful. Unless we're super concerned about bloat in some test matrix cells I wouldn't bother.

PR #703, opened 2018-10-11
(BOLT-380) Make Bolt compatible with net-ssh 4 and 5

MikaelSmith:
We plan to package Bolt with net-ssh 5, primarily for ed25519 support.
However many things that might be used alongside bolt-as-a-gem - such as
Beaker - still pin to net-ssh 4. Make it compatible with both.


MikaelSmith:
Yeah, that should always be true. I'm not sure I want to add a test matrix that shows it still works with net-ssh 4...

donoghuc:
:+1: when rubocop is.

PR #705, opened 2018-10-11
(maint) remove unsupported ruby versions from the travis matrix

adreyer:


MikaelSmith:
We should drop some from appveyor too.

PR #706, opened 2018-10-12
(BOLT-920) Add known issue for net-ssh with OpenSSH 7.8

MikaelSmith:
This is an issue with net-ssh - https://github.com/net-ssh/net-ssh/issues/633 - that we don't think we can immediately address.


MikaelSmith:
Added a brief note on key conversion.

PR #707, opened 2018-10-15
(BOLT-614) Add docs around multi-implementation task names

lucywyman:
**What this changes** Add documentation around task implementations having distinct names and individual metadata
**Why** So that tasks can be uploaded to the forge, and for a better user experience with puppet enterprise

### DIFF
--- a/pre-docs/writing_tasks.md
+++ b/pre-docs/writing_tasks.md
@@ -188,23 +188,37 @@ A task can also have multiple implementations, with metadata that explains when 
 
 ```
 - tasks
-  - sql.sh
-  - sql.ps1
+  - sql_linux.sh
+  - sql_linux.json
+  - sql_windows.ps1
+  - sql_windows.json
   - sql.json
 ```
 
-This task has two executables \(sql.sh and sql.ps1\) with a metadata file. The metadata file contains an implementations section:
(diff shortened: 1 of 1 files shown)


pre-docs/writing_tasks.md
```diff
@@ -188,23 +188,36 @@ A task can also have multiple implementations, with metadata that explains when
 
 ```
 - tasks
-  - sql.sh
-  - sql.ps1
+  - sql-linux.sh
```
adreyer: I don't think these are valid task names '-' and won't be backwards compatible.

pre-docs/writing_tasks.md
```diff
@@ -188,23 +188,36 @@ A task can also have multiple implementations, with metadata that explains when
...
-  - sql.ps1
+  - sql_linux.sh
+  - sql_linux.json
+  - sql_windows.ps1
+  - sql_windows.json
   - sql.json
 ```
 
-This task has two executables \(sql.sh and sql.ps1\) with a metadata file. The metadata file contains an implementations section:
+This task has two executables \(sql_linux.sh and sql_windows.ps1\) with a metadata file. The executables have distinct names, so that they can be uploaded to the forge. Each implementation can also optionally have it's own metadata, in addition to the task metadata. This provides information about the implementations, which can be seen in the Puppet Enterprise Tasks section and help others determine which implementation to use.
```
adreyer: - I think it would be best to explain the underlying reason "so that they can be used with older task runners such as Puppet Enterprise before 2019".  
- Each implementation must have it's own metadata to be uploaded to forge. now that private has landed soemthing like:  "Task authors should provide metadata for each implementation either documenting how to use it directly or marking it private to hide it from UI lists.

PR #708, opened 2018-10-15
(BOLT-916) Download task files from master and support shareable code

adreyer:
This updates bolt server so that it can download and cache task files
from puppet-server if supplied with puppet-server file-data instead of a
single task file. This enables support for shareable task code tasks in
the transport API.

PR #711, opened 2018-10-16
(BOLT-931) Put task executable in installdir when present

MikaelSmith:
When using an installdir for extra files, put the task executable there under its own module path to simplify writing and testing tasks that use other files as libraries. Updating according to clarification of the tasks specification.


MikaelSmith:
This is going to conflict with https://github.com/puppetlabs/bolt/pull/708.

PR #713, opened 2018-10-17
(WIP)(maint) Remove user from run-as-command command

lucywyman:
**What this changes** We appended the user to the configured 'run-as-command' in the ssh transport. This removes that to allow greater flexibility for the run-as-command.
**Why** This made it difficult to construct a full run-as-command for users since they needed to anticipate the user being added to the command being run. It also made it impossible to run any command that didn't specify the user.

### DIFF
--- a/lib/bolt/transport/ssh/connection.rb
+++ b/lib/bolt/transport/ssh/connection.rb
@@ -221,8 +221,9 @@ module Bolt
         def execute(command, sudoable: false, **options)
           result_output = Bolt::Node::Output.new
           run_as = options[:run_as] || self.run_as
-          escalate = sudoable && run_as && @user != run_as
-          use_sudo = escalate && @target.options['run-as-command'].nil?
+          run_as_cmd = @target.options['run-as-command']
+          escalate = sudoable && (run_as || run_as_cmd) && @user != run_as
+          use_sudo = escalate && run_as_cmd.nil?
 
           command_str = command.is_a?(String) ? command : Shellwords.shelljoin(command)
           if escalate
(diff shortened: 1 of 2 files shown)


lib/bolt/transport/ssh/connection.rb
```diff
@@ -232,8 +233,7 @@ def execute(command, sudoable: false, **options)
               sudo_str = Shellwords.shelljoin(sudo_flags)
               command_str = "#{sudo_str} #{command_str}"
             else
-              run_as_str = Shellwords.shelljoin(@target.options['run-as-command'] + [run_as])
-              command_str = "#{run_as_str} #{command_str}"
+              command_str = "#{run_as_cmd} #{command_str}"
```
adreyer: This is a breaking change because it prevents using run-as-command with 'run-as'

lucywyman: Yes, but users will just need to update their configs to add the user to run-as-command. It'll be a pain to communicate but we can add an error message about it and I think that's doable, and will allow blocked users to use bolt

nicklewis: You can currently pass `_run_as` to individual tasks inside a plan, and can have a plan which 
runs as several users, which this change wouldn't allow.

lucywyman: Does that us 'run-as-command'?

lucywyman:
This is turning out to be a larger lift than anticipated, and we aren't ready to refactor how run-as and run-as-command interact, particularly using _run_as from plans. See the conversation above.

PR #719, opened 2018-10-18
(BOLT-916) Remove ability accept file content from boltserver

adreyer:
Remove support for sending a "file" with the task in boltserver now that
we support downloading and caching files.

### DIFF
--- a/lib/bolt/task/puppet_server.rb
+++ b/lib/bolt/task/puppet_server.rb
@@ -11,18 +11,16 @@ module Bolt
         super(task_data)
       end
 
-      # TODO: currently this both converts filename -> name and fetches files to generate the path
-      # In the future we should not generate paths until after the implementation is selected to avoid
-      # unecessary fetching of files
+      # puppetserver file entries have 'filename' rather then 'name'
       def update_file_data(task_data)
-        if task_data['files']
-          task_data['files'] = task_data['files'].map do |f|
-            { 'name' => f['filename'],
-              'path' => @file_cache.update_file(f) }
-          end
-        end
-
+        task_data['files'].each { |f| f['name'] = f['filename'] }
         task_data
+      end
+
+      # Compute local path and download files from puppetserver as needed
+      def file_path(file_name)
+        file = file_map[file_name]
+        file['path'] ||= @file_cache.update_file(file)
       end
     end
   end
--- a/lib/bolt/transport/ssh.rb
+++ b/lib/bolt/transport/ssh.rb
@@ -122,19 +122,10 @@ module Bolt
       end
 
       def run_task(target, task, arguments, options = {})
-        if from_api?(task)
-          executable = task.file['filename']
-          file_content = Base64.decode64(task.file['file_content'])
-          input_method = task.metadata['input_method']
-          extra_files = []
-          write_executable = proc { |conn, dir| conn.write_executable_from_content(dir, file_content, executable) }
-        else
-          implementation = task.select_implementation(target, PROVIDED_FEATURES)
-          executable = implementation['path']
-          input_method = implementation['input_method']
-          extra_files = implementation['files']
-          write_executable = proc { |conn, dir| conn.write_remote_executable(dir, executable) }
-        end
+        implementation = task.select_implementation(target, PROVIDED_FEATURES)
+        executable = implementation['path']
+        input_method = implementation['input_method']
+        extra_files = implementation['files']
         input_method ||= 'both'
 
         # unpack any Sensitive data
@@ -158,7 +149,7 @@ module Bolt
                 end
               end
 
-              remote_task_path = write_executable.call(conn, task_dir)
+              remote_task_path = conn.write_remote_executable(task_dir, executable)
 
               if STDIN_METHODS.include?(input_method)
                 stdin = JSON.dump(arguments)
--- a/lib/bolt/transport/winrm.rb
+++ b/lib/bolt/transport/winrm.rb
@@ -108,19 +108,10 @@ catch
       end
 
       def run_task(target, task, arguments, _options = {})
-        if from_api?(task)
-          executable = task.file['filename']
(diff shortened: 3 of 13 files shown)


lib/bolt/task/puppet_server.rb
```diff
@@ -11,19 +11,17 @@ def initialize(task_data, file_cache)
...
+      # puppetserver file entries have 'filename' rather then 'name'
       def update_file_data(task_data)
-        if task_data['files']
-          task_data['files'] = task_data['files'].map do |f|
-            { 'name' => f['filename'],
-              'path' => @file_cache.update_file(f) }
-          end
-        end
-
+        task_data['files'] = task_data['files'].map { |f| f['name'] = f['filename'] }
```
MikaelSmith: Could use `map!`.

lib/bolt/task/puppet_server.rb
```diff
@@ -11,19 +11,17 @@ def initialize(task_data, file_cache)
         super(task_data)
       end
 
-      # TODO: currently this both converts filename -> name and fetches files to generate the path
-      # In the future we should not generate paths until after the implementation is selected to avoid
-      # unecessary fetching of files
+      # puppetserver file entries have 'filename' rather then 'name'
```
MikaelSmith: That's inconvenient.

lib/bolt/task/puppet_server.rb
```diff
@@ -11,19 +11,17 @@ def initialize(task_data, file_cache)
...
-          end
-        end
-
+        task_data['files'] = task_data['files'].map { |f| f['name'] = f['filename'] }
         task_data
       end
+
+      # Compute local path and download files from puppetserver as needed
+      def file_path(file_name)
+        file = file_map[file_name]
```
MikaelSmith: Do we need to look up the file if `file['path']` is already set? Could use a `begin` block to wrap everything.

lib/bolt/transport/ssh.rb
```diff
@@ -122,19 +122,11 @@ def run_script(target, script, arguments, options = {})
...
-          executable = implementation['path']
-          input_method = implementation['input_method']
-          extra_files = implementation['files']
-          write_executable = proc { |conn, dir| conn.write_remote_executable(dir, executable) }
-        end
+        implementation = task.select_implementation(target, PROVIDED_FEATURES)
+        executable = implementation['path']
+        input_method = implementation['input_method']
+        extra_files = implementation['files']
+        write_executable = proc { |conn, dir| conn.write_remote_executable(dir, executable) }
```
MikaelSmith: The proc isn't really needed anymore, could move the call back to where it's used.

lib/bolt/transport/winrm.rb
```diff
@@ -108,19 +108,11 @@ def run_script(target, script, arguments, _options = {})
...
-          executable = implementation['path']
-          input_method = implementation['input_method']
-          extra_files = implementation['files']
-          write_executable = proc { |conn, dir| conn.write_remote_executable(dir, executable) }
-        end
+        implementation = task.select_implementation(target, PROVIDED_FEATURES)
+        executable = implementation['path']
+        input_method = implementation['input_method']
+        extra_files = implementation['files']
+        write_executable = proc { |conn, dir| conn.write_remote_executable(dir, executable) }
```
MikaelSmith: Don't need the proc anymore.

lib/bolt/task/puppet_server.rb
adreyer: Yeah I discovered that a few months ago when working on cross platform tasks in puppetserver. It's a public API so we should probably change puppet.

lib/bolt/task/puppet_server.rb
adreyer: yeah I'm mutating them anyway

lib/bolt/task/puppet_server.rb
adreyer: We don't call update_file on the next line if path is set. I'm not sure what you're suggesting, this cuts out a line of code:
```
file_map[file_name]['path'] ||= @file_cache.update_file(file)
```

lib/bolt/task/puppet_server.rb
adreyer: I think each is actually what I want.

lib/bolt_server/file_cache.rb
```diff
@@ -19,38 +27,56 @@ def tmppath
...
-      uri = "#{@config.file_server_uri}#{path}"
+    def ssl_cert
+      @ssl_cert ||= File.read(@config.ssl_cert)
+    end
 
+    def ssl_key
+      @ssl_key ||= File.read(@config.ssl_key)
+    end
+
+    def client
```
adreyer: I 'm not sure what http library we'd want to use but I think this should keep the connection open for 2s by default.

MikaelSmith:
Some tests for pe-bolt-server in Orchestrator need to be fixed before this is merged.

MikaelSmith:
Fix for winrm up at https://github.com/puppetlabs/bolt/pull/720.

MikaelSmith:
I'm working on the Orchestrator test fix. The winrm fix is in.

MikaelSmith:
We should merge https://github.com/puppetlabs/orchestrator/pull/930 before this.

spec/lib/bolt_spec/file_cache.rb
```diff
@@ -0,0 +1,33 @@
...
+
+      def setup
+        self
+      end
+
+      def update_file(file_data)
+        parts = file_data['filename'].split('/')
+        modulename = 'sample'
+        if parts.size == 1
+          File.join(@moduledir, modulename, 'tasks', parts[0])
```
MikaelSmith: Not sure this will work on windows because it doesn't include an extension.

MikaelSmith: Nevermind, misread what it does.

PR #720, opened 2018-10-19
(BOLT-916) Update bolt-server winrm endpoint to support files

MikaelSmith:
Update the bolt-server winrm endpoint to be consistent with the ssh
endpoint for handling the task and target parameters.


klawhorn:
LGTM! Thank ya for the speed o' fixin. 

giphy 31

PR #721, opened 2018-10-19
(BOLT-928) Add daily purge process to bolt-server's file cache

MikaelSmith:
Add a purge process - run on startup and every day after - to remove
tasks from the file cache that haven't been used in over a week.

### DIFF
--- a/lib/bolt_server/file_cache.rb
+++ b/lib/bolt_server/file_cache.rb
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 
+require 'concurrent'
 require 'digest'
 require 'fileutils'
 require 'net/http'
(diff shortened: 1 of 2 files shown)


lib/bolt_server/file_cache.rb
```diff
@@ -95,6 +110,20 @@ def serial_execute(&block)
       promise.value
     end
 
+    # Create a cache dir if necessary and update it's last write time. Returns the dir.
+    # Acquires @cache_dir_mutex to ensure we don't try to purge the directory at the same time.
+    # Uses the directory mtime because it's simpler to ensure the directory exists and update
+    # mtime in a single place than with a file in a directory that may not exist.
+    def create_cache_dir(sha)
+      file_dir = File.join(@cache_dir, sha)
+      @cache_dir_mutex.synchronize do
```
adreyer: This could be very slightly faster if we used a read lock. Multiple processes can touch the file at once as long as we're not expiring.

MikaelSmith: Added a commit switching to read-write lock.

lib/bolt_server/file_cache.rb
```diff
@@ -95,6 +110,20 @@ def serial_execute(&block)
...
 
+    # Create a cache dir if necessary and update it's last write time. Returns the dir.
+    # Acquires @cache_dir_mutex to ensure we don't try to purge the directory at the same time.
+    # Uses the directory mtime because it's simpler to ensure the directory exists and update
+    # mtime in a single place than with a file in a directory that may not exist.
+    def create_cache_dir(sha)
+      file_dir = File.join(@cache_dir, sha)
+      @cache_dir_mutex.with_read_lock do
+        # mkdir_p doesn't error if the file exists
+        FileUtils.mkdir_p(file_dir, mode: 0o750)
```
lucywyman: Out of curiosity why does this need the 'o'?

donoghuc: I think Octal. For example hex is `0x`.

PR #723, opened 2018-10-19
(BOLT-858) Add `bolt apply` to apply a manifest directly to targets

nicklewis:
Previously, users could write a plan that used the `apply_prep` and
`apply` functions to apply a snippet of manifest code to targets. This is a
convenient and common workflow for both applying simple snippets of
manifest code to targets as well as for testing out local changes to code.
The overhead of wrapping the manifest code in a one-off plan with the
necessary ceremony to accept parameters, etc is onerous for that basic use
case.

This introduces a `bolt apply <manifest>` command that can be used to take
a snippet of ordinary puppet code, not wrapped in a plan, and apply it
remotely to targets. This allows to a smooth workflow for making local
changes to Puppet code, testing them on nodes through Bolt, and then
pushing them up to a central control repo.

This also adds a `-e` flag to accept a snippet of code directly on the CLI.


nicklewis:
Still need to add some tests

lib/bolt/bolt_option_parser.rb
```diff
@@ -270,6 +281,8 @@ def initialize(options)
     def update
       # show the --nodes and --query switches by default
       @nodes.hide = @query.hide = false
+      # Don't show the --execute switch excpet for `apply`
```
lucywyman: s/excpet/except/

lib/bolt/cli.rb
```diff
@@ -183,6 +187,10 @@ def validate(options)
         raise Bolt::CLIError,
               "Option '--noop' may only be specified when running a task"
       end
+
```
lucywyman: We should also add a check that `options[:object]` is `-e` or `--execute` - I don't think this will be common, but puppet users might try other puppet apply flags with this command.

nicklewis: `options[:code]` is set to the _value_ of the `-e/--execute` flag in the hook in `BoltOptionParser`. It will only be set if they used that flag.

lib/bolt/bolt_option_parser.rb
nicklewis:
```suggestion
      # Don't show the --execute switch except for `apply`
```

acceptance/tests/bolt_apply.rb
```diff
@@ -0,0 +1,46 @@
...
+  targets -= select_hosts(platform: [/osx-10.11/])
+
+  skip_test('no applicable nodes to test on') if targets.empty?
+
+  bolt_command = "bolt apply -e 'notify { \"hello world\": }' --nodes #{targets.join(',')}"
+  flags = {
+    '--format' => 'json'
+  }
+
+  step "execute `bolt apply` with json output" do
```
lucywyman: Should this also include a test around applying a manifest file?

MikaelSmith: The integration tests may be enough.

lib/bolt/cli.rb
```diff
@@ -183,6 +189,10 @@ def validate(options)
...
+      if options[:subcommand] == 'apply' && (options[:object] && options[:code])
```
MikaelSmith: Parens for grouping are unnecessary. But I guess they kind of make sense for what's being tested.

lib/bolt/cli.rb
```diff
@@ -362,6 +379,32 @@ def run_plan(plan_name, plan_arguments, nodes, options)
       result.ok? ? 0 : 1
     end
 
+    def apply_manifest(code, targets, filename = nil)
+      require 'bolt/catalog'
+      begin
+        ast = Puppet::Pops::Parser::EvaluatingParser.new.parse_string(code, filename)
+        ast_data = Puppet::Pops::Serialization::ToDataConverter.convert(ast, rich_data: true, symbol_to_string: true)
+      rescue Puppet::Error => e
+        raise Bolt::PAL::PALError, "Failed to parse manifest: #{e}"
```
MikaelSmith: Is this really a PALError? We're not using PAL here.

lib/bolt/cli.rb
```diff
@@ -362,6 +379,32 @@ def run_plan(plan_name, plan_arguments, nodes, options)
       result.ok? ? 0 : 1
     end
 
+    def apply_manifest(code, targets, filename = nil)
+      require 'bolt/catalog'
```
MikaelSmith: You don't seem to use anything from `Bolt::Catalog`. Is this just triggering `Bolt::PAL.load_puppet`?

MikaelSmith:
Also needs a rebase.

lib/bolt/cli.rb
nicklewis: Whoops, I was using`Bolt::Catalog#parse_manifest` but changed it and left that in.

lib/bolt/cli.rb
nicklewis: I used it because it's the closest thing we have to "an error from puppet". We have `Bolt::PuppetError` but that's meant to be literally a `Puppet::DataTypes::Error` object turned into a Bolt exception...

lib/bolt/cli.rb
nicklewis: Yeah I was sort of trying to logically divide this into "this validation step applies to the apply command" vs. "this is the case we're actually validating"

lib/bolt/bolt_option_parser.rb
```diff
@@ -107,6 +107,13 @@ def self.examples(cmd, desc)
 Available options are:
     HELP
 
+    APPLY_HELP = <<-HELP
```
MikaelSmith: `bolt apply` doesn't show up as an available subcommand.

lib/bolt/cli.rb
```diff
@@ -362,6 +379,31 @@ def run_plan(plan_name, plan_arguments, nodes, options)
       result.ok? ? 0 : 1
     end
 
+    def apply_manifest(code, targets, filename = nil)
+      begin
+        ast = Puppet::Pops::Parser::EvaluatingParser.new.parse_string(code, filename)
```
MikaelSmith: It looks like we still need something to pull this into scope. When I run `bolt apply test.pp -n localhost` I get
```
NameError: uninitialized constant Bolt::CLI::Puppet
  /Users/michaelsmith/puppetlabs/bolt/lib/bolt/cli.rb:386:in `rescue in apply_manifest'
  /Users/michaelsmith/puppetlabs/bolt/lib/bolt/cli.rb:383:in `apply_manifest'
  /Users/michaelsmith/puppetlabs/bolt/lib/bolt/cli.rb:275:in `execute'
  /Users/michaelsmith/puppetlabs/bolt/exe/bolt:10:in `<top (required)>'
  /Users/michaelsmith/puppetlabs/bolt/.bundle/gems/ruby/2.5.0/bin/bolt:23:in `load'
  /Users/michaelsmith/puppetlabs/bolt/.bundle/gems/ruby/2.5.0/bin/bolt:23:in `<top (required)>'
```

MikaelSmith: Probably because we call this before `pal`.

nicklewis: Whoops so that `require 'bolt/catalog'` I just removed was necessary after all. :(

MikaelSmith: Or something like it. Calling `pal` before the block also works.

MikaelSmith: Maybe worth throwing this into a helper in Bolt::PAL?

nicklewis: "load-bearing", shall we say

PR #724, opened 2018-10-22
[FM-7504] - Put a conditional around Winrm version

HelenCampbell:
Due to a new version of Winrm being released that has a higher Ruby dependency (>=2.2.0)
installations of Bolt are failing on Puppet 4, considering that Puppet 4 ships with Ruby 1.9.0.

This is an attempt to fix this issue by pinning to the last version of winrm that will work with older Ruby,
while allowing newer versions to install the latest 2.x release.

### DIFF
--- a/bolt.gemspec
+++ b/bolt.gemspec
@@ -41,7 +41,18 @@ Gem::Specification.new do |spec|
   spec.add_dependency "puppet", [">= 6.0.1", "< 7"]
   spec.add_dependency "r10k", "~> 2.6"
   spec.add_dependency "terminal-table", "~> 1.8"
-  spec.add_dependency "winrm", "~> 2.0"
+
+  # Winrm 2.3.x requires Ruby 2.2.0 or greater, so require an earlier version for lesser rubies
+  begin
+    if RUBY_VERSION < "2.2.0"
+      spec.add_dependency "winrm", "= 2.2.3"
+    else
+      spec.add_dependency "winrm", "~> 2.0"
+    end
+  rescue
+    exit(1)
+  end
+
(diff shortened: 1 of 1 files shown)


HelenCampbell:
https://rubygems.org/gems/winrm/versions/2.3.0

HelenCampbell:
@lucywyman Would you or someone in your team mind reviewing this for me please? This is causing failures on our CI pipelines for our task modules that are tested on Puppet 4 - https://jenkins-master-prod-1.delivery.puppetlabs.net/view/modules/view/cross-platform/view/exec/view/master/job/forge-module_puppetlabs-exec_intn-sys_full-pe-lts-master/

lucywyman:
@HelenCampbell  We technically only support ruby 2.3+ https://github.com/puppetlabs/bolt/blob/master/README.md#supported-platforms - I tried installing the gemfile with ruby 1.9, and it doesn't know about `__dir__`, which we use throughout the gem and package install processes. I can try to remove instances where it isn't necessary (like installing our local gemfile), but I'm not sure I'll be able to get all the way to a successful install on 1.9. Even if we did, I think a lot of parts of bolt would be pretty unusable. Is there CI that depends on this, or are users raising the issue?

HelenCampbell:
Hey @lucywyman , thank you for having a look! I had issues too with the local setup of Ruby 1.9.0. We must have successfully set it up in CI without realising Bolts Ruby requirement in the first place as it's been running green until the Winrm release.
That being said I'm happy enough to close this PR and will skip the tests when Ruby is < 2.3.
And to confirm, we are just seeing this in CI, no users have raised the issue.
Thanks again!

PR #725, opened 2018-10-22
(maint) Fix typo in exception handler

donoghuc:
When a `Bolt::error` is rescued there was a typo in the `is_a?` method where the `?` was ommitted. This was causing acceptance tests failures in ORCH.


donoghuc:
Running acceptance tests locally with changes to `file_cache.rb` from https://github.com/puppetlabs/bolt/pull/721.

PR #726, opened 2018-10-23
(maint) Fix typo in options hash

donoghuc:
The `:executor` option in the concurrent promise instantiation had a typo that resulted in non-parallel download resulting in non-serial execution of `download_file`.


donoghuc:
This was easy to overlook when only a single task is downloaded. It also got passed some manual testing when the cache was being populated as failed attempts (for fixing a different, LOUDER bug) succeeded in eventually downloading the files and thus masking the error.

PR #727, opened 2018-10-23
(maint) Ensure puppet6 is installed in spec test

donoghuc:
When testing deferred type ensure puppet 6 is installed.


donoghuc:
The `puppet_agent::install` task should point to the latest puppet release (thus be greater than puppet 6). For a brief window the latest puppet was resolving to puppet 5 which appears to be an issue with internal packaging that has been resolved. I think this is safe to close given that the expected behavior of `puppet_agent::install` with no parameters is to get the latest puppet.

PR #730, opened 2018-10-24
(BOLT-944) Only print selected options for a target

lucywyman:
**What this changes** Modify the `to_s` method of `Bolt::Target` to only print the user, host, protocol, and port of the target
**Why** So that sensitive and unnecessary information is not displayed when printing targets

### DIFF
--- a/spec/bolt/target_spec.rb
+++ b/spec/bolt/target_spec.rb
@@ -102,6 +102,16 @@ describe Bolt::Target do
     expect(uri.port).to be_nil
   end
 
+  it "does not print password when converted to string" do
+    opts = { 'user' => 'person',
+             'password' => 'secret',
+             'host' => 'machine',
+             'protocol' => 'ssh' }
+    target = Bolt::Target.new('example.com', opts)
+    expect(target.to_s).to eq("Target('example.com', "\
+                              "#{opts.reject { |k, _| k == 'password' }})")
+  end
+
   describe "with winrm" do
     it "accepts 'winrm://host:port'" do
       uri = Bolt::Target.new('winrm://neptune:55985')
--- a/spec/fixtures/apply/basic/plans/inventory_lookup.pp
+++ b/spec/fixtures/apply/basic/plans/inventory_lookup.pp
@@ -11,9 +11,9 @@ plan basic::inventory_lookup(TargetSpec $nodes) {
     notify { "Target 1 Facts: ${$t[1].facts}": }
     # Vars from target
     notify { "Target 1 Vars: ${$t[1].vars}": }
-    # Prove config is respected (tty set to 11 in config)
-    notify { "Target 0 Config: ${t[0]}": }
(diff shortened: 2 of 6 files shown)


spec/bolt/target_spec.rb
```diff
@@ -102,6 +102,14 @@
     expect(uri.port).to be_nil
   end
 
+  it "does not print password when converted to string" do
+    opts = { 'user' => 'person',
+             'host' => 'machine',
```
donoghuc: Should this hash have `password' in it?

lucywyman: :woman_facepalming:  yes

spec/integration/apply_compile_spec.rb
```diff
@@ -234,7 +234,7 @@ def get_notifies(result)
           expect(notify[0]['title']).to eq("Num Targets: 3")
           expect(notify[1]['title']).to eq("Target 1 Facts: {operatingsystem => Ubuntu, added => fact}")
           expect(notify[2]['title']).to eq("Target 1 Vars: {environment => production, features => [puppet-agent]}")
-          res = "Target 0 Config: Target('foo', {\"connect-timeout\"=>11, \"tty\"=>false, \"host-key-check\"=>false})"
+          res = "Target 0 Config: Target('foo', {})"
```
donoghuc: The purpose of this particular test is to prove that inventory options are respected when querying inventory in an apply block. Will need to modify the plan to return the options instead of implicitly calling `to_s` with `#{}`.

lucywyman: Are there other opts we should include in the list? Before making the changes here.

donoghuc: Would it be a pain to have it configurable? Make optional arg with default value of empty hash. The default (empty hash) could just return the `uri` then let users supply a hash of options. Maybe a special case for "all" in the case someone wants to quickly debug something?

donoghuc: Probably not worth making configurable. In a plan you can just access the `options` attribute.
`notify { "Target 0 Config: ${t[0].options}": }`

spec/fixtures/apply/basic/plans/inventory_lookup.pp
```diff
@@ -11,9 +11,9 @@
     notify { "Target 1 Facts: ${$t[1].facts}": }
     # Vars from target
     notify { "Target 1 Vars: ${$t[1].vars}": }
-    # Prove config is respected (tty set to 11 in config)
-    notify { "Target 0 Config: ${t[0]}": }
+    # Prove config is respected (port set to 12345)
```
donoghuc: Where is port set to 12345?

spec/bolt/target_spec.rb
donoghuc: I still do not see password?

PR #735, opened 2018-10-29
(maint) Rubocop update with new cops

lucywyman:
**What this changes** Unalign ruby hashes so that `=>` is only one space away from key and value, not aligned with other `=>`
**Why** Rubocop updated to include a cop around this which is causing travis to fail. I don't think it's a cop worth skipping, so updated bolt to conform


adreyer:
I really hate mandatory hash alinging it results in a lot of useless whitespace changes.

lucywyman:
You'll like this cop then :)  Sorry, the cop itself is called 'AlignHash' but it aligns based on the key, not the separator.

MikaelSmith:
Do we care about enforcing this? We could also just disable the cop.

lucywyman:
@MikaelSmith I don't care strongly, but think having consistent hash layout is good and Alex seems to not like aligning the rockets, so I think we should keep the cop as is.

PR #737, opened 2018-10-30
Support device modules in apply statements

adreyer:
The following PR to the panos module shows how device tasks can be updated to use the remote transport.
https://github.com/puppetlabs/puppetlabs-panos/pull/70

### DIFF
--- a/lib/bolt/config.rb
+++ b/lib/bolt/config.rb
@@ -10,6 +10,7 @@ require 'bolt/transport/winrm'
 require 'bolt/transport/orch'
 require 'bolt/transport/local'
 require 'bolt/transport/docker'
+require 'bolt/transport/remote'
 
 module Bolt
   TRANSPORTS = {
@@ -17,7 +18,8 @@ module Bolt
     winrm: Bolt::Transport::WinRM,
     pcp: Bolt::Transport::Orch,
     local: Bolt::Transport::Local,
-    docker: Bolt::Transport::Docker
+    docker: Bolt::Transport::Docker,
+    remote: Bolt::Transport::Remote,
   }.freeze
 
   class UnknownTransportError < Bolt::Error
@@ -41,6 +43,7 @@ module Bolt
       'tty' => false
     }.freeze
 
+    # TODO: move these to the transport themselves
     TRANSPORT_SPECIFIC_DEFAULTS = {
       ssh: {
         'host-key-check' => true
@@ -53,7 +56,10 @@ module Bolt
         'task-environment' => 'production'
       },
       local: {},
-      docker: {}
+      docker: {},
+      remote: {
+        'run-on' => 'localhost'
+      }
     }.freeze
 
     def self.default
--- a/lib/bolt/inventory.rb
+++ b/lib/bolt/inventory.rb
@@ -207,6 +207,7 @@ module Bolt
     private :resolve_name
 
     def expand_targets(targets)
+      # TODO: inventory may not be set in this case?
       if targets.is_a? Bolt::Target
         targets
       elsif targets.is_a? Array
--- a/lib/bolt/inventory.rb
+++ b/lib/bolt/inventory.rb
@@ -136,6 +136,17 @@ module Bolt
       }
     end
 
+    # TODO: This does two things because the applicator bypasses run_task
+    # It would probably be cleaner to give the exectutor access to inventory
+    # and handle this there.
+    def run_on_target(target, params)
+      if target.remote?
+        [get_target(target.run_on || 'localhost'), params.merge('_target' => target.hash)]
+      else
+        [target, params]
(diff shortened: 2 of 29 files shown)


lib/bolt/inventory.rb
```diff
@@ -136,6 +136,17 @@ def data_hash
       }
     end
 
+    # TODO: This does two things because the applicator bypasses run_task
+    # It would probably be cleaner to give the exectutor access to inventory
+    # and handle this there.
+    def run_on_target(target, params)
+      if target.remote?
```
lucywyman: Is this true for all non-local targets, or just `:remote` transport targets?

lib/bolt/target.rb
```diff
@@ -110,5 +131,9 @@ def user
     def password
       Addressable::URI.unencode_component(@uri_obj.password) || @password
     end
+
+    def remote?
```
lucywyman: `device?`?

lib/bolt/node/errors.rb
```diff
@@ -7,7 +7,8 @@ class Node
     class BaseError < Bolt::Error
       attr_reader :issue_code
 
-      def initialize(message, issue_code)
+      # TODO can we just drop issue code here?
```
lucywyman: +1

lib/bolt/target.rb
```diff
@@ -63,6 +63,15 @@ def parse(string)
     end
     private :parse
 
+    def run_on_target
+      if remote?
+        raise "Target was creates without inventory? Not get_targets?" unless @inventory
```
lucywyman: s/creates/created/

I'm not sure I can parse the error message...

lib/bolt/transport/base.rb
```diff
@@ -68,6 +69,33 @@ def with_events(target, callback)
...
+      end
+
+      def select_implementation(target, task)
+        impl =task.select_implementation(target.run_on_target, provided_features)
+        impl['input_method'] ||= default_input_method
+        impl
+      end
+
+      def add_target_param(target, task, params)
+        # TODO: How should metadata impact this?
```
lucywyman: Metadata seems to be more "controllable" (and visible) to the user, so I think "prioritizing" that makes the most sense:
```
if !target.remote? && task.remote?
    raise Bolt::Node::RemoteError.new('Remote tasks can only be run on remote targets')
end
```
Raising an error if the target is remotable and the task doesn't specify it could be confusing if users don't know where the 'remotable' target is coming from.

adreyer: on the other hand running a non-remote task against a remote target seems incredibly dangerous.

lib/bolt/config.rb
```diff
@@ -9,13 +9,15 @@
...
+require 'bolt/transport/remote'
 
 module Bolt
   TRANSPORTS = {
     ssh: Bolt::Transport::SSH,
     winrm: Bolt::Transport::WinRM,
     pcp: Bolt::Transport::Orch,
-    local: Bolt::Transport::Local
+    local: Bolt::Transport::Local,
+    remote: Bolt::Transport::Remote,
```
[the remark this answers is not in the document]
adreyer: This is a larger issue(or two closely related issues) I think we need to discuss and/or research.

Are all remote targets devices(ie physical devices or vms with limited control APIs)?
No, remote tasks should be used for cloud resources, notification APIs etc.

Do they use the puppet-device framework?
Usually but not always. I would expect most ruby implementations to use the puppet-device framework but they could very will be written without that especially to enable non-ruby implementations.

My general hypothesis is that module authors especially ruby authors will want to think of the modules they write as "device" modules since they use the device framework and renaming the framework is probably not worth the effort. On the other hand, I think bolt users, and eventually PE GUI admins should not think of these things as devices but rather remote targets that rely on a proxy since it's weird to call an aws account a device.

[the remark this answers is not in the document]
lucywyman: Naming is hard :) This seems to echo us not knowing how to refer to ephemeral, cloudy | container "nodes". 

Definitely worth more discussion.

lib/bolt/transport/base.rb
lucywyman: I'm not sure I understand what a task be "remotable" is enough to comment. For right now it just seems like "the user said in the metadata it can be run on remote targets", and if they *don't* specify that they still might be able to run on remote targets. 

I think we can raise a warning in that case, but limiting the user to only tasks that are remote seems like it makes using remote tasks much harder.

adreyer: I think the general statement is that a remoteable task is one that does something meaningful with the `_target` metaparam. 

IE the service task is not remoteable. If I ran it targeting an aws account it would stop a service on the run-on node rather then do anything to the real device.

PR #739, opened 2018-10-30
(BOLT-917) Return JSON from bolt_server for errors

lucywyman:
**What this changes** Generate JSON for body of 400 and 500 error responses from bolt_server
**Why** so that PE tools that consume the errors can parse them.

### DIFF
--- a/lib/bolt_server/transport_app.rb
+++ b/lib/bolt_server/transport_app.rb
@@ -2,6 +2,7 @@
 
 require 'sinatra'
 require 'bolt'
+require 'bolt/error'
 require 'bolt/target'
 require 'bolt/task/puppet_server'
 require 'bolt_server/file_cache'
@@ -37,6 +38,15 @@ module BoltServer
       result
     end
 
+    def validate_schema(schema, body)
+      schema_error = JSON::Validator.fully_validate(schema, body)
+      if schema_error.any?
+        Bolt::Error.new("There was an error validating the request body.",
+                        'boltserver/schema-error',
+                        schema_error)
+      end
+    end
+
     get '/' do
       200
     end
@@ -60,8 +70,8 @@ module BoltServer
       content_type :json
 
       body = JSON.parse(request.body.read)
-      schema_error = JSON::Validator.fully_validate(@schemas["ssh-run_task"], body)
-      return [400, schema_error.join] if schema_error.any?
+      error = validate_schema(@schemas["ssh-run_task"], body)
+      return [400, error.to_json] unless error.nil?
 
       opts = body['target']
       if opts['private-key-content']
@@ -85,8 +95,8 @@ module BoltServer
       content_type :json
 
       body = JSON.parse(request.body.read)
-      schema_error = JSON::Validator.fully_validate(@schemas["winrm-run_task"], body)
-      return [400, schema_error.join] if schema_error.any?
+      error = validate_schema(@schemas["winrm-run_task"], body)
+      return [400, error.to_json] unless error.nil?
 
       opts = body['target'].merge('protocol' => 'winrm')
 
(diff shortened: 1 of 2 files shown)


lib/bolt_server/transport_app.rb
```diff
@@ -54,7 +54,8 @@ def initialize(config)
 
       body = JSON.parse(request.body.read)
       schema_error = JSON::Validator.fully_validate(@schemas["ssh-run_task"], body)
-      return [400, schema_error.join] if schema_error.any?
+      # Do we want to construct an error hash, or just this?
```
MikaelSmith: schema_error is an array. I think we should at least return the array rather than joining it.

MikaelSmith:
Just quoting the string responses doesn't seem like the intention of the ticket. @adreyer can you clarify what problem it's supposed to solve?

MikaelSmith:
I suspect the intention is to use an error format similar to what our other services use, like https://github.com/puppetlabs/bolt/blob/1.2.0/lib/bolt/error.rb#L20-L22.

lucywyman:
Can do! I left a comment on the ticket asking and should probably have brought it up when we groomed this.

lib/bolt_server/transport_app.rb
```diff
@@ -54,7 +54,8 @@ def initialize(config)
...
+      return [400, JSON.generate(schema_error.join)] if schema_error.any?
```
MikaelSmith: I think the schema errors should be included as an array item under `details`.

lib/bolt_server/transport_app.rb
```diff
@@ -54,7 +54,13 @@ def initialize(config)
 
       body = JSON.parse(request.body.read)
       schema_error = JSON::Validator.fully_validate(@schemas["ssh-run_task"], body)
-      return [400, schema_error.join] if schema_error.any?
+      if schema_error.any?
+        msg = "There was an error validating the request body. Please see for expected schema"
```
donoghuc: I don't think you need the `please see for expected schema'. See the details? 

Also in general would it be worth just having a `make_error(kind, msg, details)` method that returns error hash?

lib/bolt_server/transport_app.rb
```diff
@@ -78,7 +84,13 @@ def initialize(config)
 
       body = JSON.parse(request.body.read)
       schema_error = JSON::Validator.fully_validate(@schemas["winrm-run_task"], body)
-      return [400, schema_error.join] if schema_error.any?
+      if schema_error.any?
+        msg = "There was an error validating the request body. Please see for expected schema"
```
MikaelSmith: Was this supposed to be different from line 58?

lucywyman: Good catch - fixed to match!

lib/bolt_server/transport_app.rb
```diff
@@ -54,7 +54,13 @@ def initialize(config)
 
       body = JSON.parse(request.body.read)
       schema_error = JSON::Validator.fully_validate(@schemas["ssh-run_task"], body)
-      return [400, schema_error.join] if schema_error.any?
+      if schema_error.any?
+        msg = "There was an error validating the request body."
```
MikaelSmith: This would make sense as a helper, to avoid having inconsistent responses in the future.

lib/bolt_server/transport_app.rb
```diff
@@ -53,8 +63,8 @@ def initialize(config)
       content_type :json
 
       body = JSON.parse(request.body.read)
-      schema_error = JSON::Validator.fully_validate(@schemas["ssh-run_task"], body)
-      return [400, schema_error.join] if schema_error.any?
+      error = validate_schema(@schemas["ssh-run_task"], body)
+      return [400, JSON.generate(error)] unless error.nil?
```
lucywyman: I think this return has to be outside of the helper in order to return from the sinatra endpoint instead of the helper function. I could be wrong though - lemme know if there's a better way to do this.

lib/bolt_server/transport_app.rb
lucywyman: Yeah, I'm not sure why I didn't use `Bolt::Error`....refactored!

MikaelSmith:
Seems to need a rebase.

PR #740, opened 2018-10-30
(BOLT-925) Scrub stack trace from bolt server errors

donoghuc:
Remove stack traces from bolt-server responses by removing them from the result status hash in bolt server. This should preserve the trace in log messages. If a stack trace is included in the result of a task output (for example a task that throws and exception during it's execution) it will still be included in the response.

### DIFF
--- a/lib/bolt_server/transport_app.rb
+++ b/lib/bolt_server/transport_app.rb
@@ -30,6 +30,13 @@ module BoltServer
       super(nil)
     end
 
+    def scrub_stack_trace(result)
+      if result.dig(:result, '_error', 'details', 'stack_trace')
+        result[:result]['_error']['details'].reject! { |k| k == 'stack_trace' }
+      end
+      result
+    end
+
     get '/' do
       200
     end
@@ -70,7 +77,8 @@ module BoltServer
 
       # Since this will only be on one node we can just return the first result
       results = @executor.run_task(target, task, parameters)
-      [200, results.first.to_json]
+      result = scrub_stack_trace(results.first.status_hash)
+      [200, result.to_json]
     end
 
     post '/winrm/run_task' do
@@ -90,7 +98,8 @@ module BoltServer
 
       # Since this will only be on one node we can just return the first result
       results = @executor.run_task(target, task, parameters)
-      [200, results.first.to_json]
+      result = scrub_stack_trace(results.first.status_hash)
+      [200, result.to_json]
     end
 
     error 404 do
--- a/spec/bolt_server/transport_app_spec.rb
+++ b/spec/bolt_server/transport_app_spec.rb
@@ -133,6 +133,37 @@ describe "BoltServer::TransportApp" do
       expect(result).to include('status' => 'success')
       expect(result['result']['_output']).to match(/got passed the message: Hello!/)
     end
+
+    context 'with Bolt::Error that contains a stack trace' do
+      let(:error_result) do
+        ex = RuntimeError.new("oops")
+        ex.set_backtrace('/path/to/bolt/code.rb:42')
+        Bolt::Result.from_exception(conn_target('ssh'), ex)
+      end
+
+      it 'scrubs stack trace from result' do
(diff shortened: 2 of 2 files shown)


lib/bolt_server/transport_app.rb
```diff
@@ -30,6 +30,13 @@ def initialize(config)
       super(nil)
     end
 
+    def scrub_stack_trace(result)
```
MikaelSmith: This name doesn't imply anything about jsonifying the result, I'd rather leave that separate.

donoghuc:
When a stack trace is reported for a result object it is removed from the response but is preserved in the logs. However if a task has an error that results in a stack trace it is included in the response. 
Task:
```
#!/usr/bin/env ruby
raise "ouch"
```

Result 
```
{"node":"jk248fme0gpv2zk.delivery.puppetlabs.net","status":"failure","result":{"_output":"","_error":{"kind":"puppetlabs.tasks/task-error","issue_code":"TASK_ERROR","msg":"The task failed with exit code 1:\n/tmp/6bdf4211-9a41-4ff7-aa14-5059969d3012/echo.sh:2:in `<main>': ouch (RuntimeError)\n","details":{"exit_code":1}}}}
```
Removing stack traces from task errors seems difficult because they appear to be captured from stdout.

MikaelSmith:
I think those are fine to leave. Stack traces produced by a task run are a valid part of the result.

lib/bolt_server/transport_app.rb
```diff
@@ -30,6 +30,13 @@ def initialize(config)
...
+      if result[:result]['_error'] && result[:result]['_error']['details']['stack_trace']
```
donoghuc: Seems like it is it safe to assume that every `result[:result]['_error'] has a `details` key. If a result does not have an `_error` key the `&&` condition will short circuit and avoid a `undefined method `[]' for nil:NilClass` problem for nested keys.

donoghuc:
Finding it hard to get a stack trace to clean in an automated test. I guess I do not see a test anywhere for existence of `[:result]['_error']['details']['stack_trace']`.

lib/bolt_server/transport_app.rb
adreyer: `if result.dig(:result, '_error', 'details', 'stack_trace')`

donoghuc: :man_facepalming: Of course. That was silly of me.

spec/bolt_server/transport_app_spec.rb
```diff
@@ -133,6 +133,37 @@ def file_data(file)
...
+          'parameters': { "message": "Hello!" }
+        }
+
+        post path, JSON.generate(body), 'CONTENT_TYPE' => 'text/json'
+        expect(last_response).to be_ok
+        expect(last_response.status).to eq(200)
+        result = JSON.parse(last_response.body)
+        expect(result).to include('status' => 'failure')
+        expect(result['result']['_error']).to include('msg' => 'oops', 'details' => { 'class' => 'RuntimeError' })
+        expect(result['result']['_error']).not_to include('stack_trace' => '/path/to/bolt/code.rb:42')
```
MikaelSmith: You can also say `.not_to include('stack_trace')`, which seems clearer about the intention.

spec/bolt_server/transport_app_spec.rb
```diff
@@ -212,6 +243,36 @@ def file_data(file)
...
+          'parameters': { "input": "Hello!" }
+        }
+
+        post path, JSON.generate(body), 'CONTENT_TYPE' => 'text/json'
+        expect(last_response).to be_ok
+        expect(last_response.status).to eq(200)
+        result = JSON.parse(last_response.body)
+        expect(result).to include('status' => 'failure')
+        expect(result['result']['_error']).to include('msg' => 'oops', 'details' => { 'class' => 'RuntimeError' })
+        expect(result['result']['_error']).not_to include('stack_trace' => '/path/to/bolt/code.rb:42')
```
MikaelSmith: Same here.

PR #743, opened 2018-11-01
(BOLT-956) Add wait_until_available function

MikaelSmith:
Add the `wait_until_available` function to wait for all targets to be available for connections. Accepts a description, wait_time in seconds (default 120), and retry_interval in seconds (default 1).

### DIFF
--- a/bolt-modules/boltlib/lib/puppet/functions/run_command.rb
+++ b/bolt-modules/boltlib/lib/puppet/functions/run_command.rb
@@ -26,7 +26,7 @@ Puppet::Functions.create_function(:run_command) do
   # @param options Additional options: '_catch_errors', '_run_as'.
   # @return A list of results, one entry per target.
   # @example Run a command on targets
-  #   run_command('hostname', $targets, '_catch_errors' => true)
+  #   run_command('hostname', $targets, 'Get hostname')
   dispatch :run_command_with_description do
     param 'String[1]', :command
     param 'Boltlib::TargetSpec', :targets
--- a/lib/bolt/error.rb
+++ b/lib/bolt/error.rb
@@ -50,13 +50,14 @@ module Bolt
   class RunFailure < Bolt::Error
     attr_reader :result_set
 
-    def initialize(result_set, action, object)
+    def initialize(result_set, action, object = nil)
       details = {
         'action' => action,
         'object' => object,
         'result_set' => result_set
       }
-      message = "Plan aborted: #{action} '#{object}' failed on #{result_set.error_set.length} nodes"
+      object_msg = " '#{object}'" if object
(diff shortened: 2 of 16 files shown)


MikaelSmith:
Needs tests.

bolt-modules/boltlib/lib/puppet/functions/run_command.rb
```diff
@@ -26,7 +26,7 @@
   # @param options Additional options: '_catch_errors', '_run_as'.
   # @return A list of results, one entry per target.
   # @example Run a command on targets
-  #   run_command('hostname', $targets, '_catch_errors' => true)
+  #   run_command('hostname', $targets, 'Get hostname')
```
lucywyman: Not sure I follow why this changed?

spec/bolt/executor_spec.rb
```diff
@@ -310,6 +310,44 @@ def mock_node_results(_run_as = nil)
...
+      results = executor.wait_until_available(targets)
+      results.each do |result|
+        expect(result).to be_instance_of(Bolt::Result)
+      end
+    end
+
+    it 'errors after timeout' do
+      allow(ssh).to receive(:connected?).and_return(false)
+
+      results = executor.wait_until_available(targets, wait_time: 0, retry_interval: 0)
```
lucywyman: Might also be worth adding a test to ensure that retry_interval and wait_time are respected. wait_time would be easy to test, retry_interval might be trickier...you could maybe count how many times `Time.now` is called?

MikaelSmith: I can work on it. Getting the `wait` tests reliable has proved tricky.

lib/bolt/error.rb
```diff
@@ -50,13 +50,14 @@ def initialize(msg)
...
 
-    def initialize(result_set, action, object)
+    def initialize(result_set, action, object = nil)
       details = {
         'action' => action,
         'object' => object,
         'result_set' => result_set
       }
-      message = "Plan aborted: #{action} '#{object}' failed on #{result_set.error_set.length} nodes"
+      object_msg = " '#{object}'" if object
```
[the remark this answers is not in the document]
MikaelSmith: It allows `object` to be optional.

bolt-modules/boltlib/lib/puppet/functions/run_command.rb
MikaelSmith: The example didn't map to the function, which takes a 3rd `description` argument before options.

spec/bolt/executor_spec.rb
MikaelSmith: Added a new test and improved existing ones around `Time.now`.

PR #744, opened 2018-11-01
(BOLT-942) Allow target to be added to group during plan execution

donoghuc:
This commit adds an `add_to_group` method callable from `Target` objects in plans. The method can be called from a `Target` object with the argument of the group to add the `Target` to. When a group does not exist the target will not be added to ANY group. If a `Target` of the same name exists in the requested group it will be overridden by the added `Target`.

### DIFF
--- a/bolt-modules/boltlib/lib/puppet/functions/add_to_group.rb
+++ b/bolt-modules/boltlib/lib/puppet/functions/add_to_group.rb
@@ -1,0 +1,36 @@
+# frozen_string_literal: true
+
+require 'bolt/error'
+
+# Adds a target to specified inventory group.
+Puppet::Functions.create_function(:add_to_group) do
+  # @param targets A pattern or array of patterns identifying a set of targets.
+  # @param The name of the group to add targets to.
+  # @example Add new Target to group.
+  #   Target.new('foo@example.com', 'password' => 'secret').add_to_group('group1')
+  # @example Add new target to group by name.
+  #   add_to_group('bolt:bolt@web.com', 'group1')
+  # @example Add an array of targets to group by name.
+  #   add_to_group(['host1', 'group1', 'winrm://host2:54321'], 'group1')
+  # @example Add a comma separated list list of targets to group by name.
+  #   add_to_group('foo,bar,baz', 'group1')
+  dispatch :add_to_group do
+    param 'Boltlib::TargetSpec', :targets
+    param 'String[1]', :group
+  end
+
+  def add_to_group(targets, group)
+    inventory = Puppet.lookup(:bolt_inventory) { nil }
+
+    unless inventory && Puppet.features.bolt?
+      raise Puppet::ParseErrorWithIssue.from_issue_and_stack(
+        Puppet::Pops::Issues::TASK_MISSING_BOLT, action: _('process targets through inventory')
+      )
+    end
+
+    executor = Puppet.lookup(:bolt_executor) { nil }
+    executor&.report_function_call('add_to_group')
+
+    inventory.add_to_group(inventory.get_targets(targets), group)
+  end
+end
--- a/lib/bolt/inventory.rb
+++ b/lib/bolt/inventory.rb
@@ -93,6 +93,19 @@ module Bolt
       targets.map { |t| update_target(t) }
     end
 
+    def add_to_group(targets, desired_group)
+      if group_names.include?(desired_group)
+        targets.each do |target|
+          if group_names.include?(target.name)
+            raise ValidationError.new("Group #{target.name} conflicts with node of the same name", target.name)
+          end
+          add_node(@groups, target, desired_group)
+        end
+      else
+        raise ValidationError.new("Group #{desired_group} does not exist in inventory", nil)
+      end
+    end
+
     def set_var(target, key, value)
       data = { key => value }
       set_vars_from_hash(target.name, data)
@@ -243,5 +256,40 @@ module Bolt
       end
     end
     private :set_facts
+
+    def add_node(current_group, target, desired_group, track = { 'all' => nil })
+      if current_group.name == desired_group
+        # Group to add to is found
+        t_name = target.name
+        # Add target to nodes hash
+        current_group.nodes[t_name] = { 'name' => t_name }.merge(target.options)
+        # Inherit facts, vars, and features from hierarchy
+        current_group_data = { facts: current_group.facts, vars: current_group.vars, features: current_group.features }
+        data = inherit_data(track, current_group.name, current_group_data)
+        set_facts(t_name, @target_facts[t_name] ? data[:facts].merge(@target_facts[t_name]) : data[:facts])
+        set_vars_from_hash(t_name, @target_vars[t_name] ? data[:vars].merge(@target_vars[t_name]) : data[:vars])
+        data[:features].each do |feature|
+          set_feature(target, feature)
+        end
+        return true
+      end
+      # Recurse on children Groups if not desired_group
+      current_group.groups.each do |child_group|
+        track[child_group.name] = current_group
+        add_node(child_group, target, desired_group, track)
+      end
+    end
+    private :add_node
+
+    def inherit_data(track, name, data)
+      unless track[name].nil?
+        data[:facts] = track[name].facts.merge(data[:facts])
+        data[:vars] = track[name].vars.merge(data[:vars])
+        data[:features].concat(track[name].features)
+        inherit_data(track, track[name].name, data)
+      end
+      data
+    end
+    private :inherit_data
   end
 end
--- a/spec/fixtures/modules/add_group/plans/init.pp
+++ b/spec/fixtures/modules/add_group/plans/init.pp
@@ -1,0 +1,30 @@
+plan add_group (TargetSpec $nodes) {
+  $new_target = Target.new(
+    "0.0.0.0:20024",
+    "host-key-check" => false,
+    "user" => 'bolt',
+    'password' => 'bolt'
+  )
+  $t = get_targets('add_me')
+  # Add new facts/var, specifically one that is novel, the other that should override existing.
+  $new_target.add_facts({'plan_context' => 'keep', 'override_parent' => 'keep'})
+  $new_target.set_var('plan_context','keep')
+  $new_target.set_var('override_parent', 'keep')
+  # Add from Target object
+  $new_target.add_to_group('add_me')
+  # Add from string
+  add_to_group('bar_1', 'bar')
+  # Add to default "all" group
+  add_to_group('add_to_all', 'all')
+  $add_me_targets = get_targets('add_me')
+
+  $result = { 'addme_group' => $add_me_targets,
+			  'existing_facts' => $add_me_targets[0].facts,
+			  'existing_vars' => $add_me_targets[0].vars,
+			  'added_facts' => $add_me_targets[1].facts,
+			  'added_vars' => $add_me_targets[1].vars,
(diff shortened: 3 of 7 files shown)


donoghuc:
So far I have manually tested against our docker spec images with the following:
**inventoryfile**
```
---
groups:
  - name: foo
    nodes:
      - name: 'localhost'
    config:
      transport: local
    groups:
      - name: one
        nodes:
          - name: "bolt:bolt@0.0.0.0:20023"
        config:
          transport: ssh
  - name: two
    nodes:
      - name: "127.0.0.1"
    config:
      transport: local
```
**add_to_group plan**
```
plan add_to_group (TargetSpec $nodes) {
  $new_target = Target.new(
    "0.0.0.0:20024",
    "host-key-check" => false,
    "user" => 'bolt',
    'password' => 'bolt'
  )
  $new_target.add_facts({'one' => 'two'})
  $new_target.add_to_group('one')
  run_command('whoami', 'one')
  return get_targets('one')[1].facts
}
```
**invocation and results**
```
cas@cas-ThinkPad-T460p:~/working_dir/bolt$ bolt plan run add_to_group -n 0.0.0.0:20022 --no-host-key-check --user bolt --password bolt
Loading modules from /home/cas/working_dir/bolt/bolt-modules:/home/cas/working_dir/bolt/Boltdir/modules:/home/cas/working_dir/bolt/Boltdir/site:/home/cas/working_dir/bolt/modules
Starting: plan add_to_group
Starting: command 'whoami' on bolt:bolt@0.0.0.0:20023, 0.0.0.0:20024
Running command 'whoami' on ["bolt:bolt@0.0.0.0:20023"]
Running command 'whoami' on ["0.0.0.0:20024"]
{"node":"0.0.0.0:20024","status":"success","result":{"stdout":"bolt\n","stderr":"","exit_code":0}}
{"node":"bolt:bolt@0.0.0.0:20023","status":"success","result":{"stdout":"bolt\n","stderr":"","exit_code":0}}
Finished: command 'whoami' with 0 failures in 0.2 sec
Finished: plan add_to_group in 0.21 sec
{
  "one": "two"
}

```
I am looking at adding integration tests in inventory_spec.rb but I am really curious about why we have an inventoryfile that is in JSON with a yaml extension?   https://github.com/puppetlabs/bolt/blob/2ecf7d19300a4ee3a7081b4900d5454ac349b7b9/spec/integration/inventory_spec.rb#L58
Will I need to have JSON there or yaml if I want to create some groups?

bolt-modules/boltlib/lib/puppet/functions/add_to_group.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+
+require 'bolt/error'
+
+# Adds a target to specified inventory group.
+Puppet::Functions.create_function(:add_to_group) do
+  # @param target A Target object from implicitly passed by calling method.
+  # @param The name of the group to add target to.
+  # @example Add new target to group.
+  #   Target.new('foo@example.com', 'password' => 'secret').add_to_group('group1')
```
adreyer: There should probably be a string example too
```
add_to_group('foo@example.com', 'group1')
```

lib/bolt/inventory.rb
```diff
@@ -243,5 +249,16 @@ def set_facts(tname, hash)
       end
     end
     private :set_facts
+
+    def add_node(grp, target, group_name)
+      grp.groups.each do |g|
+        add_node(g, target, group_name) if g.groups.any?
+        if g.name == group_name
+          g.nodes[target.uri.to_s] = { 'name' => target.uri.to_s }.merge(target.options)
```
adreyer: This should probably use `target.name` rather then `target.uri.to_s`

lib/bolt/inventory.rb
```diff
@@ -93,6 +93,12 @@ def get_targets(targets)
       targets.map { |t| update_target(t) }
     end
 
+    def add_to_group(target, group_name)
+      if group_names.include?(group_name)
```
adreyer: do we need an error raising else here?

lib/bolt/inventory.rb
```diff
@@ -243,5 +249,16 @@ def set_facts(tname, hash)
       end
     end
     private :set_facts
+
+    def add_node(grp, target, group_name)
+      grp.groups.each do |g|
+        add_node(g, target, group_name) if g.groups.any?
+        if g.name == group_name
+          g.nodes[target.uri.to_s] = { 'name' => target.uri.to_s }.merge(target.options)
+          break
```
adreyer: this control flow only stops the actual level in which the group was found. 
If we use `return true` instead we could short circuit more.

adreyer:
@donoghuc yaml is a superset of json.

spec/integration/inventory_spec.rb
```diff
@@ -148,6 +148,82 @@ def fact_plan(name = 'facts')
...
+            },
+            vars: {
+              exclude: 'dont_inherit'
+            }
+          }
+        ]
+      }
+    end
+
+    it 'sets a variable on the target' do
```
donoghuc: I intend to clean this test setup up. I wanted to get something in to help with review.

bolt-modules/boltlib/lib/puppet/functions/add_to_group.rb
```diff
@@ -0,0 +1,32 @@
...
+  # @example Add new target to group by name.
+  #   add_to_group('bolt:bolt@web.com', 'group1')
+  dispatch :add_to_group do
+    param 'Boltlib::TargetSpec', :target
```
MikaelSmith: This could legally be multiple targets. And I think we should allow that. So the name should be `targets`, use `get_targets` to convert it to an array of targets, and pass that to Inventory.

lib/bolt/inventory.rb
```diff
@@ -243,5 +250,33 @@ def set_facts(tname, hash)
       end
     end
     private :set_facts
+
+    def add_node(grp, target, group_name, track = { 'all' => nil })
+      grp.groups.each do |g|
+        # track the parent data to inherit facts and vars from the group hierarchy
+        track[g.name] = { name: grp.name, data: grp }
```
MikaelSmith: This could use a little spacing to denote grouping. At least after `next ...`.

lib/bolt/inventory.rb
MikaelSmith: Should it create a group if the name is unknown?

donoghuc: According to ticket description is should error. I forgot about this comment/requirement. I implemented it to raise an error and am working on a test for that now.

lib/bolt/inventory.rb
```diff
@@ -243,5 +253,34 @@ def set_facts(tname, hash)
...
+        set_vars_from_hash(t_name, @target_vars[t_name] ? data[:vars].merge(@target_vars[t_name]) : data[:vars])
+        return true
+      end
+    end
+    private :add_node
+
+    def inherit_data(track, name, acc = { facts: {}, vars: {} })
+      if track[name]
+        acc[:facts].merge!(track[name][:data].facts)
+        acc[:vars].merge!(track[name][:data].vars)
```
donoghuc: I also need to update `@target_features` here...

spec/fixtures/modules/add_group/plans/init.pp
```diff
@@ -0,0 +1,26 @@
...
+    'password' => 'bolt'
+  )
+  # Add new facts/var, specifically one that is novel, the other that should override existing.
+  $new_target.add_facts({'plan_context' => 'keep', 'override_parent' => 'keep'})
+  $new_target.set_var('plan_context','keep')
+  $new_target.set_var('override_parent', 'keep')
+  # Add from Target object
+  $new_target.add_to_group('add_me')
+  # Add from string
+  add_to_group('bar_1', 'bar')
```
lucywyman: Is there any verification that this worked?

bolt-modules/boltlib/lib/puppet/functions/add_to_group.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+
+require 'bolt/error'
+
+# Adds a target to specified inventory group.
+Puppet::Functions.create_function(:add_to_group) do
```
lucywyman: This should also be added to the plan function documentation https://github.com/puppetlabs/bolt/blob/master/pre-docs/plan_functions.md

lib/bolt/inventory.rb
adreyer: I think creating a group should be a separate command since we wouldn't know where to put the new group.

donoghuc:
The other approach I have tried for this is to update the `@data` hash with the new target, then assign `@groups = Group.new(@data, @config,...)` followed by an invocation of `collect_groups`. I was having trouble maintaining the `target_{facts, vars, features}` hierarchy with this approach though. I think that if we wanted to write out an updated inventory from targets added/modified in a plan updating the `@data` hash will be important.

spec/fixtures/modules/add_group/plans/init.pp
donoghuc: Yep last expect under `'computes facts and vars based on group hierarchy'`.

bolt-modules/boltlib/lib/puppet/functions/add_to_group.rb
donoghuc: I think that will be automatically generated?

lib/bolt/inventory.rb
```diff
@@ -243,5 +253,39 @@ def set_facts(tname, hash)
       end
     end
     private :set_facts
+
+    def add_node(grp, target, group_name, track = { 'all' => nil })
```
adreyer: I'd like to rename the variables to make this more readable. Something like

grp -> current_group
group_name -> desired_group
g -> child_group

track_item.name -> parent_name

lib/bolt/inventory.rb
```diff
@@ -243,5 +253,39 @@ def set_facts(tname, hash)
...
+      grp.groups.each do |g|
+        track[g.name] = { name: grp.name, data: grp }
+        add_node(g, target, group_name, track) if g.groups.any?
+        next unless g.name == group_name
```
adreyer: It seems like this will fail if you try to add a node to "all" (or the top group)?

I suspect we need to refactor this to check if they current group is the desired group outside the look and then recurse on all children. I also think that will be more readable than next unless.

lib/bolt/inventory.rb
```diff
@@ -243,5 +253,39 @@ def set_facts(tname, hash)
...
+        data[:features].each do |feature|
+          set_feature(target, feature)
+        end
+        return true
+      end
+    end
+    private :add_node
+
+    def inherit_data(track, name, acc = { facts: {}, vars: {}, features: [] })
+      if track[name]
```
adreyer: This is just checking for nil right? `name.nil?` seems easier to understand. perhaps `inherit_data(track, track[name][:name], acc) unless track[name][:name].nil?`

adreyer:
We need to make sure that when we add nodes to a group we dont' add a node whose name conflicts with an existing group name since we validate that normally:
https://github.com/puppetlabs/bolt/blob/master/lib/bolt/inventory/group.rb#L85

bolt-modules/boltlib/lib/puppet/functions/add_to_group.rb
MikaelSmith: Can you change the param symbol to `targets`?

PR #745, opened 2018-11-07
(BOLT-965) bolt-server Dockerfile to use alpine

lucywyman:
**What this changes** This updates the pe-bolt-server dockerfile to use the alpine linux base image instead of the ruby base image. It then installs ruby and dependencies, and installs and starts pe-bolt-server just as the previous dockerfile did.
**Why** This is slimmed-down container that only includes what we need for bolt-server, and is more consistent with the image layering we use throughout Puppet.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,10 +1,38 @@
-FROM ruby:2.5.1
-RUN apt-get update -qq && apt-get install -y ruby-dev
+# Install gems
+FROM alpine:3.8 as build
+
+RUN \
+apk --no-cache add build-base ruby-dev ruby-bundler ruby-json ruby-bigdecimal git openssl-dev && \
+echo 'gem: --no-document' > /etc/gemrc && \
+bundle config --global silence_root_warning 1
 
 RUN mkdir /bolt-server
-WORKDIR /bolt-server
+# Gemfile requires gemspec which requires bolt/version which requires bolt
 ADD . /bolt-server
-RUN bundle install
+WORKDIR /bolt-server
+RUN bundle install --no-cache --path vendor/bundle
+
+# Final image
+FROM alpine:3.8
+ARG bolt_version=no-version
+LABEL org.label-schema.maintainer="Puppet Bolt Team <team-direct-change-bolt@puppet.com>" \
+      org.label-schema.vendor="Puppet" \
+      org.label-schema.url="https://github.com/puppetlabs/bolt" \
+      org.label-schema.name="PE Bolt Server" \
+      org.label-schema.license="Apache-2.0" \
+      org.label-schema.version=${bolt_version} \
+      org.label-schema.vcs-url="https://github.com/puppetlabs/bolt" \
+      # Same with these
+      #org.label-schema.vcs-ref="b75674e1fbf52f7821f7900ab22a19f1a10cafdb" \
+      #org.label-schema.build-date="2018-05-09T20:10:01Z" \
+      #org.label-schema.schema-version="1.0" \
+      org.label-schema.dockerfile="/Dockerfile"
+
+RUN \
+apk --no-cache add ruby openssl ruby-bundler ruby-json ruby-bigdecimal
+
+COPY --from=build /bolt-server /bolt-server
+WORKDIR /bolt-server
 
 EXPOSE 62658
 ENV BOLT_SERVER_CONF /bolt-server/config/docker.conf
--- a/.dockerignore
+++ b/.dockerignore
@@ -2,3 +2,4 @@
 .git/
 # If we start shipping this we may reevaluate
 Gemfile.lock
+vendor/bundle


Dockerfile
```diff
@@ -1,5 +1,24 @@
...
-RUN apt-get update -qq && apt-get install -y ruby-dev
+FROM alpine:3.8
+
+LABEL org.label-schema.maintainer="Puppet Bolt Team <team-direct-change-bolt@puppet.com>" \
+      org.label-schema.vendor="Puppet" \
+      org.label-schema.url="https://github.com/puppetlabs/bolt" \
+      org.label-schema.name="PE Bolt Server" \
+      org.label-schema.license="Apache-2.0" \
+      # Is there a way to set this?
+      #org.label-schema.version=$PUPPET_VERSION \
```
lucywyman: Solution to the version? I'm not sure what hooks to set up to have it auto-update when we update the version.

Dockerfile
```diff
@@ -1,5 +1,24 @@
...
+      org.label-schema.vcs-url="https://github.com/puppetlabs/bolt" \
+      # Same with these
+      #org.label-schema.vcs-ref="b75674e1fbf52f7821f7900ab22a19f1a10cafdb" \
+      #org.label-schema.build-date="2018-05-09T20:10:01Z" \
+      #org.label-schema.schema-version="1.0" \
+      org.label-schema.dockerfile="/Dockerfile"
+
+RUN \
+apk update && apk upgrade && \
```
MikaelSmith: I don't think you need these when using `--no-cache`.

Dockerfile
```diff
@@ -1,5 +1,24 @@
...
+apk --no-cache add build-base ruby-dev git && \
```
MikaelSmith: I think we should probably avoid having these in the final image. How about a multi-stage build that only copies over the installed gems? That should help with target size. I think we can have a final image < 100MB.

Dockerfile
```diff
@@ -1,5 +1,24 @@
...
+gem install bundler --no-ri --no-rdoc && \
```
MikaelSmith: apk has a `ruby-bundler` package that might be preferable.

Dockerfile
```diff
@@ -1,5 +1,24 @@
...
+bundle config --global silence_root_warning 1
 
 RUN mkdir /bolt-server
 WORKDIR /bolt-server
```
MikaelSmith: I saw a recommendation to use `RUN echo 'gem: --no-document' > /etc/gemrc` to avoid installing gem documentation. Might make sense to slim this down. Same with using `bundle install --no-cache`.

adreyer:
@lucywyman I don't think we should use alpine as the base of the ssh targets.(re your question)

Dockerfile
lucywyman: Would that only copy the gems so that we don't need bundler, or also copy in ruby? I'm not opposed to it, but I'm not familiar with that pattern around containers.

MikaelSmith: It can. See https://docs.docker.com/develop/develop-images/multistage-build/. The stages all go in a single Dockerfile, and are built the same way.

I'd suggest 3 stages: a base runtime image with ruby and openssl, a build image that installs all the gems (and builds their extensions), and the final image that copies over the installed gems (so we can avoid all the build tools and dependencies).

Dockerfile
MikaelSmith: I'm not sure. Probably worth asking a wider audience.

MikaelSmith: Updating it may just have to be part of the tag-and-release job.

adreyer: It seems like this only matters for the containers we publish.
Can we use an ARG and pass it when we build the container for publishing? Then we can ensure we use the same version here as in the tag.

adreyer: This is what I'm doing
```yaml 
script:
    - export BOLT_VERSION=`git describe`
    - export BOLT_TAG=$REGISTRY_HOST/$DOCKER_REGISTRY/bolt-server:$BOLT_VERSION
    - docker login -u TOKEN -p $DOCKER_TOKEN $REGISTRY_HOST
    - docker build --tag $BOLT_TAG --build-arg bolt_version=$BOLT_VERSION ./
    - docker push $BOLT_TAG
```
so it should work if you add

```Dockerfile
ARG bolt_version=no-version
org.label-schema.version=${bolt_version}
````

File: Gemfile [hunk 6]
```diff
@@ -8,7 +8,9 @@ ENV['BOLT_DISABLE_ANALYTICS'] = 'true'
 gemspec
 
 # Bolt server gems are managed here not in the gemspec
+gem "bigdecimal", '~> 1.3.5'
```
MikaelSmith: Alpine provides packages for these, ruby-bigdecimal and ruby-json. Not sure whether it's better to install packages are add gem dependencies for things that are commonly in core ruby.

Dockerfile
```diff
@@ -1,10 +1,38 @@
...
+      org.label-schema.version=${bolt_version} \
+      org.label-schema.vcs-url="https://github.com/puppetlabs/bolt" \
+      # Same with these
+      #org.label-schema.vcs-ref="b75674e1fbf52f7821f7900ab22a19f1a10cafdb" \
+      #org.label-schema.build-date="2018-05-09T20:10:01Z" \
+      #org.label-schema.schema-version="1.0" \
+      org.label-schema.dockerfile="/Dockerfile"
+
+RUN \
+apk --no-cache add ruby-dev openssl-dev ruby-bundler
```
MikaelSmith: I don't think we need the `dev` versions of ruby or openssl here.

donoghuc:
Tested against `ssh` and `winrm` tasks. Both worked as expected. Also looks like image size went from `1.03GB` to `122MB`. I looked briefly at building with https://github.com/docker-library/ruby/blob/9af33e632f173e90b4a7aba7644f2080b574e54f/2.5/alpine3.8/Dockerfile but it did not seem to be any smaller image size.

Dockerfile
lucywyman: It didn't work with regular SSL - puma would complain that SSL bindings weren't available :woman_shrugging:

lucywyman: Let me know if it works with ssl on your machine and I can investigate further.

Gemfile
MikaelSmith: Switching to using `ruby-bigdecimal` and `ruby-json`, and using non-dev versions of ruby and openssl gets my image down to 91.5MB. So I think it's worth going that route.

Dockerfile
```diff
@@ -1,10 +1,38 @@
...
+FROM alpine:3.8
+ARG bolt_version=no-version
+LABEL org.label-schema.maintainer="Puppet Bolt Team <team-direct-change-bolt@puppet.com>" \
+      org.label-schema.vendor="Puppet" \
+      org.label-schema.url="https://github.com/puppetlabs/bolt" \
+      org.label-schema.name="PE Bolt Server" \
+      org.label-schema.license="Apache-2.0" \
+      org.label-schema.version=${bolt_version} \
+      org.label-schema.vcs-url="https://github.com/puppetlabs/bolt" \
+      # Same with these
```
donoghuc: What does `Same with these` mean?

lucywyman: Oh, I was also not sure how to programmatically set them. @adreyer  Are there similar variables that get set that I can put here?

PR #746, opened 2018-11-07
(BOLT-966) Configure Bolt-Server from environment variables

donoghuc:
This commit allows select configuration options to be set with environment variables. Environment variable options take precedence over options defined in `bolt-server.conf`.

### DIFF
--- a/developer-docs/bolt-server.md
+++ b/developer-docs/bolt-server.md
@@ -22,6 +22,18 @@ Bolt server can be configured by defining content in HOCON format at one of the 
 - `whitelist`: Array, *optional* - A list of hosts which can connect to pe-bolt-server.
 - `concurrency`: Integer, *optional* - The maximum number of server threads (default `100`).
 
+**Environmnet Variable Options**
+The following configuration options can be set with environment variables. 
+- `BOLT_SSL_CERT`
+- `BOLT_SSL_KEY`
+- `BOLT_SSL_CA_CERT`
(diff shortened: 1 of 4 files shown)


lib/bolt_server/config.rb
```diff
@@ -53,8 +56,18 @@ def load_config(path)
       raise "Could not find bolt-server config at #{path}" if parsed_hocon.nil?
 
       parsed_hocon = parsed_hocon.select { |key, _| CONFIG_KEYS.include?(key) }
-      @data = @data.merge(parsed_hocon)
 
+      ENV_KEYS.each do |key|
```
adreyer: I think this should probably be a different method from loading config from a file.

developer-docs/bolt-server.md
```diff
@@ -22,6 +22,18 @@ Bolt server can be configured by defining content in HOCON format at one of the
 - `whitelist`: Array, *optional* - A list of hosts which can connect to pe-bolt-server.
 - `concurrency`: Integer, *optional* - The maximum number of server threads (default `100`).
 
+**Environmnet Variable Options**
+The following configuration options can be set with environment variables. 
+- `SSL_CERT`
```
MikaelSmith: I feel like these should have a BOLT_ prefix to avoid accidentally being overridden.

donoghuc: That is a great idea. Will revise.

PR #747, opened 2018-11-08
(BOLT-962) Add Docker transport

MikaelSmith:
Add a transport for invoking `docker exec` against a local or remote Docker runtime. A remote target can be configured using
```
config:
  docker:
    url: tcp://example.com:5422
    options: {}
```
The options hash accepts keys according to https://github.com/swipely/docker-api. Settings can also be supplied via DOCKER_* environment variables as noted in the `docker-api` README as well.

Also includes a test refactor to share examples testing POSIX targets. Would like to do WinRM but those tests are largely unique.


MikaelSmith:
I plan to add a few integration tests, but this is largely ready.

spec/bolt/transport/docker_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+    let(:transport_conf) { { 'url' => 'tcp://localhost:55555' } }
+
+    it 'uses the url' do
+      expect {
+        docker.with_connection(target) {}
+      }.to raise_error(Bolt::Node::ConnectError, /Connection refused .* 127.0.0.1:55555/)
+    end
+  end
+
+  context 'when options are specified', docker: true do
```
adreyer: can we just set docker true on the describe?

spec/integration/shareable_task_spec.rb
```diff
@@ -82,6 +82,23 @@
     include_examples "invalid metadata"
   end
 
+  describe 'over docker', docker: true do
```
adreyer: we may want to move this to the transport shared examples

MikaelSmith: I guess there's not much reason for it to use the CLI.

spec/bolt/transport/docker_spec.rb
MikaelSmith: Yeah, looks like it.

lib/bolt/target.rb
```diff
@@ -47,8 +47,8 @@ def update_conf(conf)
       @password = t_conf['password']
       @port = t_conf['port']
 
-      url_keys = %w[user password port]
-      @options = t_conf.reject { |k, _| url_keys.include?(k) }.merge(@options)
+      # Preserve everything in options so we can easily create copies of a Target.
```
MikaelSmith: This was necessary to make the Local transport play nice when changing tmpdir. I needed to create a copy of the target without tmpdir. I think the change makes sense on its own.

spec/integration/shareable_task_spec.rb
MikaelSmith: Actually these tests cover several aspects of shareable tasks: that Puppet finds all the relevant files, and that we put them in the right place. We don't tend to use fixtures in the unit tests section. So I think this separation still makes sense.

MikaelSmith: But then the cli_spec makes me a liar :)

adreyer: but we can test those aspects separately. Once with a single transport for integration and then at the transport level.

MikaelSmith: Updated.

lib/bolt/transport/docker/connection.rb
```diff
@@ -0,0 +1,111 @@
...
+    class Docker < Base
+      class Connection
+        def initialize(target)
+          @target = target
+          @logger = Logging.logger[target.host]
+        end
+
+        def connect
+          # Explicitly create the new Connection to avoid relying on global state in the Docker module.
+          url = @target.options['url'] || ::Docker.url
```
adreyer: I'm concerned url vs uri is going to end up being confusing here. "docker-url" "docker-api" ???

MikaelSmith: Maybe. It'll already be
```
config:
  docker:
    url: ...
```

MikaelSmith: What part do you feel will be confusing?

adreyer: `target.uri` vs `target.options['url']`.  Maybe `service-uri` to align with the pcp transport?

MikaelSmith: Ok, that seems reasonable. `target.options['options']` is also probably weird.

MikaelSmith:
bolt packaging needs the new gems.

lucywyman:
Works as expected using file upload, run task, run plan, and run script commands on a local ubuntu vm. I'm trying to test on a windows machine but it's taking forever...

MikaelSmith:
It's quite likely it won't work on Windows. Not sure the API's stable there yet, but folks reported some issues with things based on the docker-api gem.

pre-docs/bolt_configuration_options.md
```diff
@@ -82,6 +82,14 @@ ssh:
 
 `tmpdir`: The directory to copy and execute temporary files.
 
+## Docker transport configuration options
+
+`tmpdir`: The directory to upload and execute temporary files on the target.
+
+`url`: URL of the Docker host used for API requests. Defaults to local via a unix socket at `unix:///var/docker.sock`.
```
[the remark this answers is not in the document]
MikaelSmith: Updated.

PR #749, opened 2018-11-08
(BOLT-697) Push to PCR for successful builds

adreyer:


.travis.yml
```diff
@@ -40,6 +40,21 @@ script:
 - bundle exec rake puppetserver
 after_script:
 - docker-compose -f spec/docker-compose.yml logs
+deploy:
```
donoghuc: Does this need to be nested under `jobs:` as a multi stage build?

adreyer: My understanding is that this is a "phase" not a "stage"
https://docs.travis-ci.com/user/job-lifecycle/#the-job-lifecycle

.travis.yml
```diff
@@ -40,6 +40,21 @@ script:
...
+  provider: script
+  on:
+    # deploy the master branch or tag pushes
+    condition: $TRAVIS_BRANCH =~ ^master|\d+\.\d+\.\d+$
```
MikaelSmith: I think
```
on:
    repo: puppetlabs/bolt
    branch: master
```
may be sufficient. I don't think TRAVIS_BRANCH will be set to a tag name if the tag is pushed.

MikaelSmith: https://github.com/travis-ci/travis-ci/issues/7780 has some discussion of ways to do master and tags.

.travis.yml
```diff
@@ -40,6 +40,21 @@ script:
...
+  env:
+    - REGISTRY_HOST=pcr-internal.puppet.net
+    - DOCKER_REGISTRY=pe-and-platform
+    - secure: "KF0rqrducC1oom9XpWsOsO/opdvYn9oXmYoavJbqJ3FcwkjmBD3//vGyDaguZ62p8BwUxK5xX99/HnrPbBC/KxBPIoUZsAPoVp9FK4wOlu0UkrvXLMJKeQ8XTxihtEy1kpXfp6+G98R+a1GoQYsZ4u6h9+cOMM2Ml7UesqP9PfSeN9hxJEHkKsaiAoYR8QWiXSBp00mvN11grp+sWDxCOB5wPvc1E5qhCZcaYDPYAIDiGBENCo7YW/GBqIjrV3ZNeNxbsFUT9mP6WVnh+X206u98ajEs1JYYAKEdntkb3DF4t1rSzov0t3YS2C1L+1PntZ5QhrKlW4DoO92p0qv3TitUA4F+6+5eaeGMk9FFL88Zc+Qr240u8Or1K+RDxm7HzerqW5CUYQ6omvdPzQRJ1qYQMDO3rT/Hgxst5hc5x9wDyosPCyndaJezwEg8iAz6C9xpXSmmGILZuZCVf0xSVT0PlgJtwM+zEYLITFgdlm03Xb+aYeWXqMBtvsnZNZQu+kjpjyJ+CxCt/+UA+row4/wgA/rcnf4UBiRGNrd1vd7ar03Jb8btO1ti2iEuNzRPkP0ZX9yOSt5/pmRCrSnIJCEIwCWFnowFMNvlz70oKPkCkcpXHLUY2g4DJVXzmTfXu9XQZzXPd9YSpRHga5YsaxxYO6obaoROr6GTlvVtfD8="
+  script:
+    - export BOLT_VERSION=`git describe`
+    - export BOLT_TAG=$REGISTRY_HOST/$DOCKER_REGISTRY/bolt-server:$BOLT_VERSION
+    - docker login -u TOKEN -p $DOCKER_TOKEN $REGISTRY_HOST
+    - docker build --tag $BOLT_TAG --build-arg bolt_version=$BOLT_VERSION ./
+    - docker push $BOLT_TAG
```
MikaelSmith: Do we also want to maintain a `latest` that points to the last pushed tag?

.travis.yml
adreyer: Somewhere else in discussion of how to do tags it suggested that TRAVIS_BRANCH is set to the tag.

PR #752, opened 2018-11-12
(Docs) Copyedit request for Bolt

katelopresti:
Updated content after review by @kayayarai


pre-docs/installing_tasks_from_the_forge.md
```diff
@@ -52,7 +52,7 @@ This list of packaged modules is available in a Puppetfile in the [Bolt reposito
 
 -   [canary](https://github.com/puppetlabs/bolt/tree/master/modules/canary): Run action against a small number of nodes and only if it succeeds will it run on the rest.
 
--   [puppetdb\_fact](https://github.com/puppetlabs/bolt/tree/master/modules/puppetdb_fact): Collect facts for the specified nodes from the configured PuppetDB connection and stores the collected facts on the Targets.
+-   [puppetdb\_fact](https://github.com/puppetlabs/bolt/tree/master/modules/puppetdb_fact): Collect facts for the specified nodes from the configured PuppetDB connection and stores the collected facts on the targets.
```
MikaelSmith: This change is misleading, maybe because the original statement wasn't clear enough. It seems to say that facts will be stored on the remote machines (targets). What it was actually saying is that the Target data types in a Bolt Plan will be updated with those facts. We should probably find a clearer way to say that.

katelopresti: Agreed. I thought it was talking about nodes. I'll edit the file now.

PR #758, opened 2018-11-21
(BOLT-644) Add analytics for size of applied manifests

MikaelSmith:
For apply actions - via `bolt apply` or the `apply` function in a Plan -
capture events counting the size of the code being run (in number of
statements), and number of resources produced by those manifests.

### DIFF
--- a/lib/bolt/applicator.rb
+++ b/lib/bolt/applicator.rb
@@ -20,6 +20,7 @@ module Bolt
       @plugin_dirs = plugin_dirs
       @pdb_client = pdb_client
       @hiera_config = hiera_config ? validate_hiera_config(hiera_config) : nil
+      @analytics = executor&.analytics
 
       @pool = Concurrent::ThreadPoolExecutor.new(max_threads: max_compiles)
       @logger = Logging.logger[self]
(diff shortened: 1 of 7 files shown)


lib/bolt/applicator.rb
```diff
@@ -134,12 +135,26 @@ def apply(args, apply_body, scope)
...
+        count_statements(ast.body)
+      when Puppet::Pops::Model::BlockExpression
+        ast.statements.count
+      else
+        1
+      end
     end
 
-    def apply_ast(ast, targets, options, plan_vars = {})
+    def apply_ast(raw_ast, targets, options, plan_vars = {})
```
adreyer: It seems like this will report analytics for bolt apply as well as apply statements. I'm not sure if thats intentional. We may want to track those separately but I'm not sure.

MikaelSmith: That was intentional. I could change it if we do want to track them separately.

lib/bolt_spec/plans/mock_executor.rb
```diff
@@ -224,6 +224,8 @@ def without_default_logging
       def report_function_call(_function); end
 
       def report_bundled_content(_mode, _name); end
+
+      def analytics; end
```
MikaelSmith: Added here to mirror the accessor in the executor.

lib/bolt/applicator.rb
```diff
@@ -134,12 +135,26 @@ def apply(args, apply_body, scope)
...
+      @analytics&.event('Apply', 'ast', 'statements', count_statements(raw_ast))
```
nicklewis: This should be like `@analytics&.event('Apply', 'statements', count_statements(raw_ast))`. The "label" is the value that we want to be able to filter/group by.

PR #760, opened 2018-11-27
(docs) Document support for Kerberos as a known issue (BOLT-980)

MikaelSmith:
Add a known issue to explain our current lack of support for Kerberos because we had to remove it and can't support the obvious gem.

### DIFF
--- a/pre-docs/bolt_installing.md
+++ b/pre-docs/bolt_installing.md
@@ -241,10 +241,3 @@ To disable the collection of analytics data add the following line to `~/.puppe
 ```
 disabled: true
 ```
-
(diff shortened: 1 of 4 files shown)


pre-docs/bolt_installing.md
```diff
@@ -241,10 +241,3 @@ To disable the collection of analytics data add the following line to `~/.puppe
 ```
 disabled: true
 ```
-
-## Using Kerberos over SSH
```
lucywyman: I think this is also linked from [pre-docs/index.md] - does that automatically get updated?

PR #762, opened 2018-11-27
(maint) Refresh DOCKER_TOKEN to fix docker push to registry

MikaelSmith:
Update docker login to be more secure, and correctly encrypt DOCKER_TOKEN.


MikaelSmith:
I think this is good, can ignore the AppVeyor failure.

lucywyman:
Not sure if there's a way to test besides just merging? But it's already not working so :woman_shrugging:

PR #764, opened 2018-11-27
(BOLT-982) Create plan executor

MikaelSmith:


lib/bolt_server/plan_app.rb
```diff
@@ -0,0 +1,105 @@
+# frozen_string_literal: true
```
MikaelSmith: This probably isn't where we want it to live if it's going to be called the Plan Executor. Would that be `lib/plan_executor/app.rb`?

lib/bolt_server/plan_app.rb
```diff
@@ -0,0 +1,105 @@
...
+      @worker = Concurrent::SingleThreadExecutor.new
+
+      # Create a basic executor, leave concurrency up to Orchestrator.
+      @executor = Bolt::Executor.new(0)
+      # Use an empty inventory until we figure out where this data comes from.
+      @inventory = Bolt::Inventory.new(nil)
+      # TODO: what should max compiles be set to for apply?
+      @pal = Bolt::PAL.new(modulepath, nil)
+
+      @last_result = Concurrent::Promise.new(executor: @worker) { nil }
```
lucywyman: I'm not sure I understand what this promise is for, or if we still need it? Isn't this a part of the `/plan/run` handler?

MikaelSmith: This was to make sure I didn't have to check whether @last_result was set before accessing it.

spec/integration/apply_spec.rb
```diff
@@ -125,6 +125,7 @@ def agent_version_inventory
       end
 
       it 'succeeds when run twice' do
+        puts config_flags
```
MikaelSmith: This seems like something accidentally added.

spec/plan_executor/app_spec.rb
```diff
@@ -76,45 +73,47 @@ def app
...
+    post '/plan/run', json_req, header
     expect(last_response).to be_ok
     expect(last_response.status).to eq(200)
     status = JSON.parse(last_response.body)
     expect(status).to eq('status' => 'running')
 
     get '/plan/result'
     expect(last_response).to be_ok
-    expect(last_response.body).to be_empty
+    puts JSON.parse(last_response.body)
```
MikaelSmith: Should be removed?

MikaelSmith:
Looks good. Going to squash and merge when it passes CI.

PR #765, opened 2018-11-28
(BOLT-644) Send a single event for apply with metrics

MikaelSmith:
Send a single event for each `apply` that includes metrics for that
apply as custom dimensions: statement count, mean and standard deviation
of resource counts across targets. This provides more flexibility in how
to use the data than sending event values.

### DIFF
--- a/lib/bolt/applicator.rb
+++ b/lib/bolt/applicator.rb
@@ -20,7 +20,6 @@ module Bolt
       @plugin_dirs = plugin_dirs
       @pdb_client = pdb_client
       @hiera_config = hiera_config ? validate_hiera_config(hiera_config) : nil
-      @analytics = executor&.analytics
 
       @pool = Concurrent::ThreadPoolExecutor.new(max_threads: max_compiles)
       @logger = Logging.logger[self]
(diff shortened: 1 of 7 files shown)


MikaelSmith:
I still need a few fixes for tests. I've pushed data up to the dev analytics site. @nicklewis mind taking a look at this?

lib/bolt/executor.rb
```diff
@@ -165,20 +165,32 @@ def log_plan(plan_name)
...
       end
     end
 
+    def report_apply(statement_count, resource_counts)
+      sum = resource_counts.inject(0) { |accum, i| accum + i }
+      mean = sum / resource_counts.length.to_f
+      variance = resource_counts.inject(0) { |accum, i| accum + (i - mean)**2 } / resource_counts.length.to_f
+      std_dev = Math.sqrt(variance)
+
+      data = { statement_count: statement_count, resource_mean: mean, resource_stddev: std_dev }
```
nicklewis: I'm not sure what std dev will really tell us, since it's difficult for us to correlate it to the mean. We could potentially send coefficient of variation instead, but it's still unclear what questions that would help us answer...

MikaelSmith:
Ready for review again.

lib/bolt/applicator.rb
```diff
@@ -188,6 +184,10 @@ def apply_ast(raw_ast, targets, options, plan_vars = {})
         @executor.await_results(result_promises)
       end
 
+      # Allow for report to exclude event metrics (apply_result doesn't require it to be present)
+      resource_counts = r.ok_set.map { |result| result.event_metrics&.fetch('total') }.compact
```
nicklewis: `fetch` will raise a `KeyError` if the key doesn't exist. Is that intended?

MikaelSmith: I think the result produces an error if event_metrics exists but doesn't include 'total'.

PR #767, opened 2018-11-29
(BOLT-791) Support remote tasks

adreyer:
This adds a remote transport to support revision 4 of the task spec. The
remote transport will proxy a task to the node defined as run-on its
config.

I still want to add a few more tests and update the predocs

### DIFF
--- a/bolt.gemspec
+++ b/bolt.gemspec
@@ -40,6 +40,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency "net-ssh", ">= 4.0"
   spec.add_dependency "orchestrator_client", "~> 0.3.1"
   spec.add_dependency "puppet", [">= 6.0.1", "< 7"]
+  spec.add_dependency "puppet-resource_api"
   spec.add_dependency "r10k", "~> 2.6"
   spec.add_dependency "terminal-table", "~> 1.8"
   spec.add_dependency "winrm", "~> 2.0"
--- a/lib/bolt/config.rb
+++ b/lib/bolt/config.rb
@@ -10,6 +10,7 @@ require 'bolt/transport/winrm'
 require 'bolt/transport/orch'
 require 'bolt/transport/local'
 require 'bolt/transport/docker'
+require 'bolt/transport/remote'
 
 module Bolt
   TRANSPORTS = {
@@ -17,7 +18,8 @@ module Bolt
     winrm: Bolt::Transport::WinRM,
     pcp: Bolt::Transport::Orch,
     local: Bolt::Transport::Local,
-    docker: Bolt::Transport::Docker
+    docker: Bolt::Transport::Docker,
+    remote: Bolt::Transport::Remote
   }.freeze
 
   class UnknownTransportError < Bolt::Error
@@ -36,16 +38,15 @@ module Bolt
                            private-key tty tmpdir user connect-timeout
                            cacert token-file service-url].freeze
 
-    TRANSPORT_DEFAULTS = {
-      'connect-timeout' => 10,
-      'tty' => false
-    }.freeze
-
+    # TODO: move these to the transport themselves
     TRANSPORT_SPECIFIC_DEFAULTS = {
       ssh: {
-        'host-key-check' => true
+        'connect-timeout' => 10,
+        'host-key-check' => true,
+        'tty' => false
       },
       winrm: {
+        'connect-timeout' => 10,
         'ssl' => true,
         'ssl-verify' => true
       },
@@ -53,7 +54,10 @@ module Bolt
         'task-environment' => 'production'
       },
       local: {},
-      docker: {}
+      docker: {},
+      remote: {
+        'run-on' => 'localhost'
+      }
     }.freeze
 
     def self.default
@@ -88,7 +92,7 @@ module Bolt
 
       @transports = {}
       TRANSPORTS.each_key do |transport|
-        @transports[transport] = TRANSPORT_DEFAULTS.merge(TRANSPORT_SPECIFIC_DEFAULTS[transport])
+        @transports[transport] = TRANSPORT_SPECIFIC_DEFAULTS[transport].dup
       end
 
       update_from_file(config_data)
@@ -158,7 +162,7 @@ module Bolt
 
       TRANSPORTS.each do |key, impl|
         if data[key.to_s]
-          selected = data[key.to_s].select { |k| impl.options.include?(k) }
+          selected = impl.filter_options(data[key.to_s])
           @transports[key].merge!(selected)
         end
       end
--- a/lib/bolt/node/errors.rb
+++ b/lib/bolt/node/errors.rb
@@ -7,7 +7,8 @@ module Bolt
     class BaseError < Bolt::Error
       attr_reader :issue_code
 
-      def initialize(message, issue_code)
+      # TODO: can we just drop issue code here?
+      def initialize(message, issue_code = nil)
         super(message, kind, nil, issue_code)
       end
 
@@ -31,6 +32,12 @@ module Bolt
     class FileError < BaseError
       def kind
         'puppetlabs.tasks/task_file_error'
+      end
+    end
+
+    class RemoteError < BaseError
+      def kind
+        'puppetlabs.tasks/remote-task-error'
       end
     end
 
--- a/lib/bolt/target.rb
+++ b/lib/bolt/target.rb
@@ -6,7 +6,9 @@ require 'bolt/error'
 module Bolt
   class Target
     attr_reader :uri, :options
-    attr_writer :inventory
+    # CODEREVIEW: this feels wrong. The altertative is threading inventory through the
+    # executor to the RemoteTransport
+    attr_accessor :inventory
 
     PRINT_OPTS ||= %w[host user port protocol].freeze
 
@@ -41,7 +43,7 @@ module Bolt
     def update_conf(conf)
       @protocol = conf[:transport]
 
-      t_conf = conf[:transports][protocol.to_sym]
+      t_conf = conf[:transports][transport.to_sym] || {}
       # Override url methods
       @user = t_conf['user']
       @password = t_conf['password']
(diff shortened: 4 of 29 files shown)


lib/bolt/node/errors.rb
```diff
@@ -7,7 +7,8 @@ class Node
     class BaseError < Bolt::Error
       attr_reader :issue_code
 
-      def initialize(message, issue_code)
+      # TODO: can we just drop issue code here?
```
lucywyman: +1

lib/bolt/target.rb
```diff
@@ -85,6 +87,18 @@ def to_s
       "Target('#{@uri}', #{opts})"
     end
 
+    def to_h
+      options.merge(
+        'name' => name,
+        'uri' => uri,
+        'protocol' => protocol,
+        'user' => user,
+        'password' => password,
```
lucywyman: We redacted password in `to_s`, should we also redact it here?

lib/bolt/transport/base.rb
```diff
@@ -68,6 +68,20 @@ def with_events(target, callback)
         result
       end
 
+      def provided_features
+        []
+      end
+
+      def default_input_method
```
lucywyman: Out of curiosity, is there a reason we prefer making these methods instead of constants?

lib/bolt/transport/remote.rb
```diff
@@ -0,0 +1,54 @@
+# frozen_string_literal: true
+
+require 'bolt/task/remote'
+require 'bolt/transport/base'
+
+module Bolt
+  module Transport
+    class Remote < Base
+      def self.options
+        # TODO: We should accept arbitrary options here
```
lucywyman: Arbitrary as in configured? Is that something that's ticketed or should be ticketed?

lib/bolt/transport/base.rb
adreyer: we want one clear thing to override. methods are better than constants for that.

lib/bolt/target.rb
adreyer: no that would break serialization.

lib/bolt/transport/remote.rb
adreyer: no we can't really meaningfully restrict it in the general case. This still needs to be ticketed.

adreyer: I may need to do this sooner to support the slack task

bolt.gemspec
```diff
@@ -40,6 +40,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency "net-ssh", ">= 4.0"
   spec.add_dependency "orchestrator_client", "~> 0.3.1"
   spec.add_dependency "puppet", [">= 6.0.1", "< 7"]
+  spec.add_dependency "puppet-resource_api"
```
MikaelSmith: Why? A mention in the commit would help.

lib/bolt/config.rb
```diff
@@ -41,6 +43,7 @@ class Config
       'tty' => false
     }.freeze
 
+    # TODO: move these to the transport themselves
```
MikaelSmith: 👍

lib/bolt/transport/remote.rb
```diff
@@ -0,0 +1,54 @@
...
+
+      # Cannot batch because arugments differ
+      def run_task(target, task, arguments, options = {})
+        proxy_target = get_proxy(target)
+        transport = @executor.transport(proxy_target.protocol)
+        arguments = arguments.merge('_target' => target.to_h.reject { |_, v| v.nil? })
+
+        remote_task = Bolt::Task::Remote.new(task.to_h)
+
+        # TODO: should we have to handle this logic
```
MikaelSmith: Does this really make sense as a transport? We're having to break a lot of the abstraction to make it work.

lib/bolt/target.rb
```diff
@@ -95,10 +109,18 @@ def name
...
 
+    def remote?
+      @uri_obj.scheme == 'remote' || @protocol == 'remote'
+    end
+
     def port
       @uri_obj.port || @port
     end
 
+    def transport
```
MikaelSmith: I think this could use a discussion of the difference between transport and protocol. Not immediately clear to me what they mean semantically.

lib/bolt/node/errors.rb
MikaelSmith: Might technically be a breaking change, `fail_plan` accepts an optional `issue_code`.

pre-docs/inventory_file.md
```diff
@@ -195,10 +195,28 @@ nodes:
...
+option. This is useful if the target is an http API as in the following example.
+
+```yaml
+nodes:
+  - host1.example.com
+  - name: https://user1:secret@remote.example.com
+    config:
+      transport: remote
+      remote:
+        run-on: host1.example.com
```
MikaelSmith: This defaults to using `host1` over SSH? Would be useful to expand the example somehow to clarify how you get config for host1.example.com.

bolt.gemspec
adreyer: it's needed to use the panos module and is part of the puppet 6 package

lib/bolt/target.rb
adreyer: where should that discussion go?

protocol is part of a url/transport is a bolt thing that is usually the same

lib/bolt/transport/remote.rb
adreyer: I had a previous implementation in the executor/inventory more directly but it was a lot more code and made the inventory syntax less natural. I'm open to suggestions if you want to talk about it.

spec/bolt/transport/local_spec.rb
```diff
@@ -3,13 +3,16 @@
 require 'spec_helper'
 require 'bolt/transport/local'
 require 'bolt/target'
+require 'bolt/inventory'
```
MikaelSmith: Why does this include `bolt/inventory`?

lib/bolt/transport/base.rb
```diff
@@ -44,7 +44,12 @@ class Base
 
       # Returns options this transport supports
       def self.options
-        raise NotImplementedError, "self.options() must be implemented by the transport class"
+        raise NotImplementedError,
+              "self.options() or self.filter_options(unfiltered) must be implemented by the transport class"
+      end
+
+      def self.filter_options(unfiltered)
```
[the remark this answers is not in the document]
adreyer: ugh I didn't even realize that. Suggestions?

pre-docs/bolt_configuration_options.md
```diff
@@ -92,6 +92,15 @@ ssh:
 
 `service-options`: A hash of options to configure the Docker connection. Only necessary if using a non-default URL. See https://github.com/swipely/docker-api for supported options.
 
+## Remote transport configuration options
+
+*The remote transport is a new feature and currently experimental. It's configuration options and behavior may change between y releases*
+
+`run-on`: The proxy target the task should execute on. Default is `localhost`
+
+`conn-info`: A hash of connection info that will be passed to the device as `_target`.
```
adreyer: this is incorrect

pre-docs/writing_tasks.md
```diff
@@ -361,6 +361,71 @@ Successful on 1 node: localhost
...
+```ruby
+#!/usr/bin/env ruby
+# modules/slack/tasks/message.rb
+
+require 'json'
+require 'net/http'
+
+params = JSON.parse(STDIN.read)
+# the slack API token is passed in from inventory
+token = params['_target']['token']
```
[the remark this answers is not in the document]
adreyer: I think for now the tasks will have to. For tasks that leverage RsAPI we'll be able to add some validation later. I intentionally left this out of this rev of the spec because I have questions about how we can enable without requiring a lot of copy-paste in metadata.


https://github.com/puppetlabs/puppet-specifications/pull/133

pre-docs/writing_tasks.md
```diff
@@ -361,6 +361,71 @@ Successful on 1 node: localhost
...
+In order to use this you'll have to add a slack as a remote target in your
+inventory file.
+
+```yaml
+---
+nodes:
+  - name: my_slack
+    config:
+      transport: remote
+      token: <slack API token goes here>
```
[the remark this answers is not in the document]
adreyer: no this is incorrect. it must be under remote.

lib/bolt/target.rb
```diff
@@ -6,7 +6,9 @@
 module Bolt
   class Target
     attr_reader :uri, :options
-    attr_writer :inventory
+    # CODEREVIEW: this feels wrong. The altertative is threading inventory through the
```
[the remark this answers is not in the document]
adreyer: I think another PR since it will probably require some test updates.

pre-docs/writing_tasks.md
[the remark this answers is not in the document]
MikaelSmith: The task example itself could error if token ended up being nil. But I'm ok ignoring it initially too.

lib/bolt/node/errors.rb
lucywyman: Remove the todo if we're keeping this?

spec/bolt/transport/shared_examples.rb
```diff
@@ -327,6 +330,67 @@ def posix_context
     end
   end
 
+  context 'when used by the remote transport' do
+    let(:remote_target) do
+      # TODO: remove the config here. it is a workaround for ????
```
MikaelSmith: I don't know what this comment is saying.

adreyer: BOLT-943 means that we can't add a target with options to the inventory so we have to set config here.

pre-docs/inventory_file.md
```diff
@@ -225,10 +225,30 @@ nodes:
           run-as: root
 ```
 
--   **[Generating inventory files](inventory_file_generating.md)**
```
melamos: Just FYI, this link in docs proper is auto-generated based on how the publication is set up, so I can't remove it.

PR #769, opened 2018-11-29
(maint) Fix an occasional test failure on Windows

MikaelSmith:
Bolt tests periodically fail when run on Windows in
`spec/bolt/catalog_spec.rb:81`. This test compiles a small catalog. The
test would fail when it encountered recursion resolving the
`vendormoduledir` fact.

We shouldn't be looking up that fact during compilation anyway because
Bolt does its own vendoring. Override the setting so we don't attempt to
set it (and avoid looking up unnecessary facts).


MikaelSmith:
I don't think this does what I want yet. It doesn't prevent the default from being evaluated.

MikaelSmith:
Ok, this seems to work now.

MikaelSmith:
Failing fact was actually `env_windows_installdir`, set by puppet-agent .bat wrappers. I'll update the commit.

PR #770, opened 2018-12-03
(BOLT-510) Add node aliases to inventory

MikaelSmith:
An `alias` property can be set on a node in inventory as an alternate name whereby you may refer to that node. `alias` may be a string or an array of strings; each string is restricted to alphanumeric characters identical to group names. Any of the listed aliases may refer to that node. The alias can be referred to when looking up a node via `get_targets`.

The alias can be referred to when listing nodes in a group in inventory by referring to only the alias, as in
```
groups:
  - name: group1
    nodes:
      - mytarget
    config:
      transport: ssh
  - name group2
    nodes:
      - alias: mytarget
        name: localhost:2002
```

It's an error for an alias to match the name of a group or node name, or for an alias to refer to multiple different nodes. Node names continue to be unique, i.e. multiple node entries that use the same name refer to the same node and aggregate configuration among them, including aliases.


lib/bolt/inventory.rb
```diff
@@ -27,7 +27,11 @@ def add_parent(parent_group)
       end
 
       def message
-        "#{@_message} for group at #{path}"
+        if path
```
adreyer: should this be `unless path.empty?`

lib/bolt/inventory.rb
```diff
@@ -194,27 +200,32 @@ def update_target(target)
...
         # Try to wildcard match nodes in inventory
         # Ignore case because hostnames are generally case-insensitive
         regexp = Regexp.new("^#{Regexp.escape(target).gsub('\*', '.*?')}$", Regexp::IGNORECASE)
 
         nodes = []
         @groups.node_names.each do |node|
-          if node =~ regexp
-            nodes << node
-          end
+          nodes << node if node =~ regexp
```
adreyer: can this just be `@groups.node_names.select { |name| name =~ regexp }`

lib/bolt/inventory.rb
```diff
@@ -194,27 +200,32 @@ def update_target(target)
...
         end
 
-        raise(WildcardError, target) if nodes.empty?
-        nodes
-      else
-        [target]
+        @groups.node_aliases.each do |target_alias, node|
```
adreyer: Is there a concern with a node getting added more than once at this point?

MikaelSmith: No, that's handled by calling `uniq` in `get_targets`.

lib/bolt/inventory.rb
MikaelSmith: Yep.

lib/bolt/inventory.rb
MikaelSmith: Yeah. Next was only slightly more complicated as well.

MikaelSmith:
Updated.

adreyer:
@MikaelSmith should the last commit have (WIP) in the message?

MikaelSmith:
Oh, nope.

MikaelSmith:
Fixed.

PR #774, opened 2018-12-04
(BOLT-510) Add docs for the 'alias' inventory property

MikaelSmith:


pre-docs/inventory_file.md
```diff
@@ -123,7 +146,7 @@ The inventory file uses the following objects.
 
     A map of fact names and values. values may include arrays or nested maps.
 
--   **Features**
+-   **Feature**
```
melamos: I can't figure out why you changed this to singular, because we refer to features (plural) everywhere else on this page.

pre-docs/inventory_file.md
```diff
@@ -58,7 +58,7 @@ groups:
 
 ```
 
-## Override a user for a specific node
+### Override a user for a specific node
```
melamos: This is another one of those heading level issues that I can't easily change. Is it okay to leave this at the same level for now? I think at some point I'll have to fix this limitation in docs...

pre-docs/inventory_file.md
MikaelSmith: It was specifically related to how this section describes things. We refer to the `features` entry as `Array[Feature]`, and this section describes a single feature.

pre-docs/inventory_file.md
MikaelSmith: Ok, leave it how it was. It just seemed more like an example of using inventory rather than a completely new section.

PR #775, opened 2018-12-04
(BOLT-608) Add `local` transport for Windows OS

donoghuc:
Add support for local transport on windows OS.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -24,7 +24,7 @@ end
 
 desc "Run RSpec tests for TravisCI that don't require WinRM"
 RSpec::Core::RakeTask.new(:travisci) do |t|
-  t.rspec_opts = '--tag ~winrm --tag ~appveyor_agents --tag ~puppetserver'
+  t.rspec_opts = '--tag ~winrm --tag ~appveyor_agents --tag ~puppetserver --tag ~windows'
 end
 
 desc "Run RSpec tests that require pupperserver"
@@ -125,6 +125,11 @@ namespace :integration do
   desc 'Run tests that require Bash on the local host'
   RSpec::Core::RakeTask.new(:bash) do |t|
     t.rspec_opts = '--tag bash'
+  end
+
+  desc 'Run tests that require windows OS on the local host'
+  RSpec::Core::RakeTask.new(:windows) do |t|
+    t.rspec_opts = '--tag windows'
   end
 
   desc 'Run tests that require Puppet Agents configured with Appveyor'
--- a/lib/bolt/inventory.rb
+++ b/lib/bolt/inventory.rb
@@ -174,7 +174,7 @@ module Bolt
         data['config'] = {}
       end
 
-      unless Bolt::Util.windows? || data['config']['transport']
+      unless data['config']['transport']
         data['config']['transport'] = 'local' if target.name == 'localhost'
       end
 
--- a/lib/bolt/transport/local.rb
+++ b/lib/bolt/transport/local.rb
@@ -4,6 +4,7 @@ require 'json'
 require 'fileutils'
 require 'tmpdir'
 require 'bolt/transport/base'
+require 'bolt/transport/powershell'
 require 'bolt/util'
 
 module Bolt
@@ -21,12 +22,7 @@ module Bolt
 
       def initialize
         super
-
-        if Bolt::Util.windows?
-          raise NotImplementedError, "The local transport is not yet implemented on Windows"
-        else
-          @conn = Shell.new
-        end
+        @conn = Shell.new
       end
 
       def in_tmpdir(base)
@@ -73,12 +69,24 @@ module Bolt
 
           # unpack any Sensitive data AFTER we log
           arguments = unwrap_sensitive_args(arguments)
-          if arguments.empty?
-            # We will always provide separated arguments, so work-around Open3's handling of a single
-            # argument as the entire command string for script paths containing spaces.
-            arguments = ['']
+          if Bolt::Util.windows?
+            if Powershell.powershell_file?(file)
+              command = Powershell.run_script(arguments, file)
+              output = @conn.execute(command, dir: dir, env: "powershell.exe")
+            else
+              path, args = *Powershell.process_from_extension(file)
+              args += Powershell.escape_arguments(arguments)
+              command = args.unshift(path).join(' ')
+              output = @conn.execute(command, dir: dir)
+            end
+          else
+            if arguments.empty?
+              # We will always provide separated arguments, so work-around Open3's handling of a single
+              # argument as the entire command string for script paths containing spaces.
+              arguments = ['']
+            end
+            output = @conn.execute(file, *arguments, dir: dir)
           end
-          output = @conn.execute(file, *arguments, dir: dir)
           Bolt::Result.for_command(target, output.stdout.string, output.stderr.string, output.exit_code)
         end
       end
@@ -88,6 +96,7 @@ module Bolt
         executable = implementation['path']
         input_method = implementation['input_method']
         extra_files = implementation['files']
+        input_method ||= Powershell.powershell_file?(executable) ? 'powershell' : 'both'
 
         in_tmpdir(target.options['tmpdir']) do |dir|
           if extra_files.empty?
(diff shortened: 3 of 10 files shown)


lib/bolt/transport/local.rb
```diff
@@ -65,27 +60,298 @@ def run_command(target, command, _options = {})
...
+          if arg =~ / /
+            "\"#{arg}\""
+          else
+            arg
+          end
+        end
+      end
+
+      def shell_init
+        command = <<-PS
```
MikaelSmith: I'm guessing this is copied from WinRM. Can we pull shared knowledge around how to run things in powershell out into a separate class that the WinRM and Local transports can share?

Rakefile
```diff
@@ -127,6 +127,11 @@ namespace :integration do
     t.rspec_opts = '--tag bash'
   end
 
+  desc 'Run tests that require windows local host'
+  RSpec::Core::RakeTask.new(:bash) do |t|
```
MikaelSmith: This should have a different name. I'd opt for `:powershell`.

spec/integration/local_spec.rb
```diff
@@ -41,4 +41,22 @@
       expect(result[0]['_error']).to be
     end
   end
+
+  context 'when using CLI options on POSIX OS', windows: true do
```
MikaelSmith: I like the naming to be closer to the context (i.e. powershell vs bash).

lib/bolt/transport/local.rb
nicklewis: That would also be useful if we eventually support powershell over SSH. Like a transport vs. runtime separation

donoghuc: Will absolutely refactor to use a shared class for powershell code. I was just trying to get something working to write tests against. Having tests will help me with the refactor.

spec/integration/local_spec.rb
```diff
@@ -41,4 +51,68 @@
...
+      end
+    end
+
+    it 'runs a task reading from stdin', :reset_puppet_settings do
+      result = run_one_node(%w[task run sample::winstdin message=µsomemessage] + config_flags)
+      output = result['_output'].strip
+      output = output.force_encoding(Encoding.default_external).encode!(''.encoding) if RUBY_VERSION =~ /^2\.3\./
+      expect(output).to match(/STDIN: {"message":"µsomemessage"/)
+    end
+
```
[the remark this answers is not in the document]
donoghuc: I think this is a bug outside of this work. For example from a windows machine: 
```
C:\Users\Administrator\bolt>bundle exec bolt task run sample::winstdin message=µsomemessage --modulepath .\spec\fixtures\modules --transport winrm --user Administrator --password *** --no-ssl -n fwviuuk1typ4xz4.delivery.puppetlabs.net
bundle exec bolt task run sample::winstdin message=µsomemessage --modulepath .\spec\fixtures\modules --transport winrm --user Administrator --password Qu@lity! --no-ssl -n fwviuuk1typ4xz4.delivery.puppetlabs.net
Started on fwviuuk1typ4xz4.delivery.puppetlabs.net...
Finished on fwviuuk1typ4xz4.delivery.puppetlabs.net:
  STDIN: {"message":"-¦somemessage","_task":"sample::winstdin"}
  {
  }
Successful on 1 node: fwviuuk1typ4xz4.delivery.puppetlabs.net
Ran on 1 node in 2.36 seconds
```
I think it is a problem outside of bolt: 
```
PS C:\Users\Administrator\bolt> echo "echo µsomemessage;" | powershell -noprofile -noninteractive -command '$input | iex'
echo "echo µsomemessage;" | powershell -noprofile -noninteractive -command '$input | iex'
??somemessage
```
```
PS C:\Users\Administrator\bolt> echo "echo somemessage;" | powershell -noprofile -noninteractive -command '$input | iex'
echo "echo somemessage;" | powershell -noprofile -noninteractive -command '$input | iex'
somemessage
```

Rakefile
```diff
@@ -24,7 +24,7 @@ end
 
 desc "Run RSpec tests for TravisCI that don't require WinRM"
 RSpec::Core::RakeTask.new(:travisci) do |t|
-  t.rspec_opts = '--tag ~winrm --tag ~appveyor_agents --tag ~puppetserver'
+  t.rspec_opts = '--tag ~winrm --tag ~appveyor_agents --tag ~puppetserver --tag ~powershell'
```
adreyer: Are you running the powershell tests on linux?

donoghuc:
I have a testing environment set up at `jbwztop5z8no6nc.delivery.puppetlabs.net` feel free to use it. See example below. 

```
C:\Users\Administrator\bolt>bundle exec bolt script run test.rb one -n localhost
bundle exec bolt script run test.rb one -n localhost
Started on localhost...
Finished on localhost:
  STDOUT:
    Ruby script
    one
Ran on 1 node in 0.81 seconds
```

lib/bolt/transport/winrm.rb
```diff
@@ -78,30 +75,11 @@ def run_script(target, script, arguments, _options = {})
...
-{
-  Invoke-Command @invokeArgs
-}
-catch
-{
-  Write-Error $_.Exception
-  exit 1
-}
-          PS
+            if Bolt::Transport::Powershell.powershell_file?(remote_path)
```
adreyer: I think include the module so you can just call these methods

Rakefile
MikaelSmith: My suggestion was the opposite. We might support powershell on linux, and there's probably nothing windows-specific about it. https://github.com/puppetlabs/bolt/pull/775#discussion_r238873814 specifically.

donoghuc: I have not been running the powershell tests on my laptop. In fact I had not even really considered that as an option for linux. Will explore how this should work more in depth.

The following work with PR as is.
```
cas@cas-ThinkPad-T460p:~/working_dir/bolt$ bolt command run "powershell -noprofile -noninteractive -command 'Write-host hi'" -n localhost
Started on localhost...
Finished on localhost:
  STDOUT:
    hi
Ran on 1 node in 0.89 seconds
```
```
cas@cas-ThinkPad-T460p:~/working_dir/bolt$ cat argv.ps1 
#!/usr/bin/powershell
Write-Host $args
cas@cas-ThinkPad-T460p:~/working_dir/bolt$ bolt script run argv.ps1 one -n localhost
Started on localhost...
Finished on localhost:
  STDOUT:
    one
Ran on 1 node in 1.21 seconds
```

tphoney:
Looking forward to seeing this. +1 to all involved.

spec/integration/local_spec.rb
```diff
@@ -4,22 +4,22 @@
...
 
-  let(:whoami) { "whoami" }
   let(:modulepath) { File.join(__dir__, '../fixtures/modules') }
-  let(:stdin_task) { "sample::stdin" }
   let(:uri) { 'localhost,local://foo' }
   let(:user) { ENV['USER'] }
 
   after(:each) { Puppet.settings.send(:clear_everything_for_tests) }
 
-  context 'when using CLI options' do
```
MikaelSmith: Are there common tests we can run on all platforms?

lib/bolt/transport/local.rb
```diff
@@ -73,21 +69,34 @@ def run_script(target, script, arguments, _options = {})
 
           # unpack any Sensitive data AFTER we log
           arguments = unwrap_sensitive_args(arguments)
-          if arguments.empty?
-            # We will always provide separated arguments, so work-around Open3's handling of a single
-            # argument as the entire command string for script paths containing spaces.
-            arguments = ['']
+          if Bolt::Util.windows?
```
MikaelSmith: Seems promising.

lib/bolt/transport/winrm.rb
donoghuc: How about just changing `Bolt::Transport::Powershell.` to `Powershell.`? Seems like this solution cuts down on characters but helps disambiguate methods like `run_script`.

lib/bolt/transport/local.rb
```diff
@@ -111,13 +120,46 @@ def run_task(target, task, arguments, _options = {})
           # unpack any Sensitive data, write it to a separate variable because
           # we log 'arguments' below
           unwrapped_arguments = unwrap_sensitive_args(arguments)
-          stdin = STDIN_METHODS.include?(input_method) ? JSON.dump(unwrapped_arguments) : nil
-          env = ENVIRONMENT_METHODS.include?(input_method) ? envify_params(unwrapped_arguments) : nil
 
+          if Bolt::Util.windows?
+            # WINDOWS
+            stdin = if STDIN_METHODS.include?(input_method)
```
MikaelSmith: `stdin` is constructed identically here. Would be nice not to duplicate it.

lib/bolt/transport/local.rb
```diff
@@ -111,13 +120,46 @@ def run_task(target, task, arguments, _options = {})
...
+              end
+          else
+            # POSIX
+            stdin = STDIN_METHODS.include?(input_method) ? JSON.dump(unwrapped_arguments) : nil
+            env = ENVIRONMENT_METHODS.include?(input_method) ? envify_params(unwrapped_arguments) : nil
+
+            output = @conn.execute(script, stdin: stdin, env: env, dir: dir)
+          end
           # log the arguments with sensitive data redacted, do NOT log unwrapped_arguments
           logger.debug("Running '#{script}' with #{arguments}")
```
MikaelSmith: This logging should happen before the task is executed.

spec/integration/local_spec.rb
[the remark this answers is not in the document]
MikaelSmith: What version of Ruby are you using?

MikaelSmith:
Needs a rebase.

lib/bolt/inventory.rb
```diff
@@ -174,7 +174,7 @@ def update_target(target)
         data['config'] = {}
       end
 
-      unless Bolt::Util.windows? || data['config']['transport']
+      unless data['config']['transport']
         data['config']['transport'] = 'local' if target.name == 'localhost'
```
MikaelSmith: Whole thing could be collapsed to `data['config']['transport'] ||= 'local' if target.name == 'localhost'`.

lib/bolt/transport/local.rb
```diff
@@ -108,16 +121,43 @@ def run_task(target, task, arguments, _options = {})
...
 
-          output = @conn.execute(script, stdin: stdin, env: env, dir: dir)
+            output =
+              if Powershell.powershell_file?(script) && stdin.nil?
+                command = Powershell.run_ps_task(arguments, script, input_method)
+                command = environment_params + Powershell.shell_init + command
+                if input_method == 'powershell'
+                  @conn.execute(command, dir: dir, env: "powershell.exe")
+                else
+                  @conn.execute(command, dir: dir, stdin: stdin, env: "powershell.exe")
```
MikaelSmith: This feels like a weird use of `env`. `env` was supposed to be about setting up environment variables. However it looks like the way we order that in shell.rb doesn't work for Windows.

PR #777, opened 2018-12-05
(BOLT-983) Add plan executor configuration file

lucywyman:
**What this changes** This adds a new `config.rb` to the plan executor to load and manage configuration for the app. It also adds a `plan_config.rb` which reads config and starts the puma server. Lastly it updates documentation in the `developer-docs` subdir to be up to date.
**Why** So that we can configure and run the plan executor puma application.


lucywyman:
Still updating the docs, but I think this is about done.

lib/plan_executor/config.rb
```diff
@@ -0,0 +1,119 @@
+# frozen_string_literal: true
+
+require 'hocon'
+require 'bolt/error'
+
+module PlanExecutor
```
adreyer: This seems like a lot of copy/paste can we make a base class?

plan_config.rb
```diff
@@ -0,0 +1,47 @@
+# frozen_string_literal: true
```
adreyer: lets create a config directory for this and the other config.rb script

lib/plan_executor/config.rb
```diff
@@ -0,0 +1,119 @@
...
+                   'ssl-ca-cert', 'ssl-cipher-suites', 'loglevel', 'logfile',
+                   'whitelist', 'workers', 'cache-dir',
+                   'file-server-conn-timeout', 'file-server-uri'].freeze
+
+    ENV_KEYS = ['ssl-cert', 'ssl-key', 'ssl-ca-cert', 'loglevel',
+                'workers', 'file-server-conn-timeout', 'file-server-uri'].freeze
+
+    DEFAULTS = {
+      'host' => '127.0.0.1',
+      'port' => 67526,
```
nicklewis: This isn't a legal port on many systems (it's > 65535). We should probably just make it adjacent to the bolt-server port for simplicity.

lib/bolt_server/base_config.rb
```diff
@@ -0,0 +1,140 @@
+# frozen_string_literal: true
+
+require 'hocon'
+require 'bolt/error'
+
+module BoltServer
+  class BaseConfig
+    def self.base_config_keys
```
lucywyman: I think there's a better way to do this, but I wasn't able to turn up how.
What I want is to set a 'default' here that's available in for the child subclass to modify, and when the function is called the child class modified version is called.

lib/bolt_server/base_config.rb
```diff
@@ -0,0 +1,140 @@
...
+
+      unless ssl_cipher_suites.is_a?(Array)
+        raise Bolt::ValidationError, "Configured 'ssl-cipher-suites' must be an array of cipher suite names"
+      end
+
+      unless whitelist.nil? || whitelist.is_a?(Array)
+        raise Bolt::ValidationError, "Configured 'whitelist' must be an array of names"
+      end
+
+      if self.class.config_keys.include?('concurrency')
```
lucywyman: Would it be better to break this into a "natural numbers" list of configurations? I don't think that's any shorter and like having the example of service-specific validation, but having 2 in a row does seem redundant.

lib/bolt_server/base_config.rb
nicklewis: You can use `super` from the child class to call the parent implementation.

```rb
# parent
def config_keys
  [ ... ]
end

# child
def config_keys
  super + ['additional_key']
end
```

nicklewis: I guess it's a little more complicated because that's defined as a class method. But I think just making it an instance method would be fine.

plan_config.rb
adreyer: I think both this and the other config file should be in a directory. Is that going to be difficult for packaging?

lib/bolt_server/config.rb
```diff
@@ -1,118 +1,31 @@
...
-        @data[key]
-      end
-    end
-
-    def initialize(config = nil)
-      @data = DEFAULTS.clone
-      @data = @data.merge(config.select { |key, _| CONFIG_KEYS.include?(key) }) if config
-      @config_path = nil
+  class Config < BoltServer::BaseConfig
+    def self.config_keys
```
adreyer: This will be simpler if it's not a class method and just users super probably. True for the other shared functions too.

```
def config_keys
  super << 'concurrency'
end
```

lib/bolt_server/base_config.rb
lucywyman: I wasn't sure how to create the accessor methods with config keys as an instance method:
```
    config_keys.each do |key|
      define_method(key.tr('-', '_').to_sym) do
        @data[key]
      end
    end
```

With that I get `undefined method 'config_keys' for PlanExecutor::Config:Class`

plan_config.rb
MikaelSmith: Shouldn't be, we already copy them somewhere else.

lib/service_config/bolt_server_config.rb
```diff
@@ -0,0 +1,25 @@
...
+    def config_keys
+      super + %w[concurrency]
+    end
+
+    def env_keys
+      super + %w[concurrency]
+    end
+
+    def defaults
+      parent_defaults = super
```
adreyer: `super.merge( 'port' => 62658, 'concurrency' => 100, 'cache-dir' => "/opt/puppetlabs/server/data/bolt-server/cache")`

puma_config.rb
```diff
@@ -8,7 +8,7 @@
 
 require 'bolt_server/transport_app'
 require 'bolt_server/acl'
-require 'bolt_server/config'
+require 'service_config/bolt_server_config'
```
adreyer: is this right?

lib/service_config/bolt_server_config.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+require 'hocon'
+require 'service_config/base_config'
+require 'bolt/error'
+
+module BoltServer
```
adreyer: this is not a standard file path for this module

adreyer: I was suggesting a new directory for the puma config scripts not these configuration classes.

lucywyman: Suggestions?

adreyer: ruby files in lib should always contain classes in standard format. So `BoltServer::Config` should be in `lib/bolt_server/config`.

However we shouldn't clutter up the top of the repo with a bunch of different puma config scripts. so we can move `plan_config.rb` and `puma_config.rb` to some directory. `config/plan.rb` `config/transport.rb` for example.

lucywyman: Ah, sorry - that wasn't clear but makes a lot more sense.

config/puma_config.rb
```diff
@@ -8,7 +8,7 @@
 
 require 'bolt_server/transport_app'
 require 'bolt_server/acl'
-require 'bolt_server/config'
+require 'service_config/bolt_server_config'
```
nicklewis: Looks like this should still be `bolt_server/config`?

developer-docs/bolt-api-servers.md
```diff
@@ -1,16 +1,22 @@
...
+**Global Bolt Server Config**: `/etc/puppetlabs/bolt-server/conf.d/bolt-server.conf`
 
-**Local Config**: `~/.puppetlabs/bolt-server.conf`
+**Local Bolt Server Config**: `~/.puppetlabs/bolt-server.conf`
 
-**Options**
+
+**Global Plan Executor Config**: `/etc/puppetlabs/plan-executor/conf.d/plan-executor.conf`
+
+**Local Plan Executor Config**: `~/.puppetlabs/plan-executor.conf`
```
nicklewis: We don't actually look up a local config file (for either service), so those should be removed.

config/puma_config.rb
nicklewis: Also, this should have a more specific name than `puma_config.rb` now, since there's also a `plan_config.rb`. Maybe `transport_server_config.rb` or something? Given we have to make a corresponding bolt-vanagon change either way, this seems like the best time to rename it.

lib/bolt_server/base_config.rb
```diff
@@ -0,0 +1,131 @@
...
+        'loglevel' => 'notice',
+        'file-server-conn-timeout' => 120 }
+    end
+
+    def initialize(config = nil)
+      @data = defaults
+      @data = @data.merge(config.select { |key, _| config_keys.include?(key) }) if config
+      @config_path = nil
+      config_keys.each do |key|
+        self.class.send(:define_method, key.tr('-', '_').to_sym) { @data[key] }
```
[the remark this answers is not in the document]
adreyer: Are these methods even that useful? Can we just remove them?

lib/bolt_server/base_config.rb
```diff
@@ -0,0 +1,131 @@
...
+      end
+    end
+
+    def load_file_config(path)
+      @config_path = path
+      begin
+        # This lets us get the actual config values without needing to
+        # know the service name
+        parsed_hocon = Hocon.load(path).values[0]
```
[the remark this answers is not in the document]
lucywyman: The path? Like, construct the path from service names?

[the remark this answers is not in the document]
nicklewis: The name of the section in the config file, rather than assuming there is exactly one and taking the first.

lib/bolt_server/base_config.rb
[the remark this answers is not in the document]
lucywyman: Since we already have `@data` and have a getter for it I decided to just use that, but if there's a better way to access those variables I'm open to it.

lucywyman:
I kept the last commit separate in case the changes are too much and we want to break it out into a separate PR.

nicklewis:
I like the change to `@data` :+1:

PR #781, opened 2018-12-06
(maint) Bump Rubocop

donoghuc:
Version 0.61.0 contained a bug so we pinned to 0.60.0 with https://github.com/puppetlabs/bolt/pull/778 PR. The 0.61.1 tag fixes the issue.


MikaelSmith:
A revert would've been pretty bullet-proof.

PR #782, opened 2018-12-07
(BOLT-969) Exclude cache tmp directory from purge

donoghuc:
Previously when purging the temporary directory for downloading task files (weekly or on startup) the `tmppath` directory itself was deleted resulting in errors downloading files. Instead of deleting the temporary directory during a purge delete only the contents of the temporary directory.

### DIFF
--- a/developer-docs/bolt-server.md
+++ b/developer-docs/bolt-server.md
@@ -250,9 +250,9 @@ Where `<TARGET>` is either:
 * A vmpooler VM. To use this, replace `<TARGET>` above with the hostname.
 * One of the containers brought up by the `docker-compose` in the `spec` directory. To use these, you'll want to:
 
-    * get the IP of the **bolt-server container** (not the target container):
+    * get the IP that the **bolt-server container** believes it is hosted on (for example the IP of the developer laptop hosting the bolt-server container):
       ```
-      docker inspect -f '{{range .NetworkSettings.Networks}}{{.IPAddress}}{{end}}' bolt_boltserver_1
+      bolt command run "/sbin/ip route" -n docker://bolt_boltserver_1 | awk '/default/ { print $3 }'
       # Should return an IP such as 172.20.0.1
       ```
     * Append the port of one of the 3 containers to that IP: `20022` (for an ubuntu node with no agent), `20023` (for a puppet 5 agent), or `20024` (for a puppet 6 agent). It's also helpful to include the protocol (`ssh`), user (`bolt`), and password (`bolt`) in the URI.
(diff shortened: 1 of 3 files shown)


lib/bolt_server/file_cache.rb
```diff
@@ -166,8 +166,11 @@ def expire(purge_ttl)
       expired_time = Time.now - purge_ttl
       @cache_dir_mutex.with_write_lock do
         Dir.glob(File.join(@cache_dir, '*')).select { |f| File.directory?(f) }.each do |dir|
-          if (mtime = File.mtime(dir)) < expired_time
-            @logger.debug("Removing #{dir}, last used at #{mtime}")
+          next unless (mtime = File.mtime(dir)) < expired_time
+          @logger.debug("Removing #{dir}, last used at #{mtime}")
+          if dir == tmppath
+            FileUtils.rm_rf(Dir.glob("#{tmppath}/*"))
```
donoghuc: It was not immediately clear to me if it is possible that anything would be left in the tmp dir to delete? I see we use a `Tempfile` i'm not sure exactly how those are handled.

developer-docs/bolt-server.md
```diff
@@ -250,9 +250,9 @@ Where `<TARGET>` is either:
 * A vmpooler VM. To use this, replace `<TARGET>` above with the hostname.
 * One of the containers brought up by the `docker-compose` in the `spec` directory. To use these, you'll want to:
 
-    * get the IP of the **bolt-server container** (not the target container):
+    * get the IP that the **bolt-server container** believes it is hosted on (for example the IP of the developer laptop hosting the bolt-server container):
```
donoghuc: @lucywyman Does this make sense to you?

lucywyman: I think so. If you can find a way to work in that it's the laptop IP on the bridge network between the host and container, I found that useful, but it also assume a lot of knowledge about container networking that I don't think we need to explain there.

lib/bolt_server/file_cache.rb
adreyer: I think it should be fine wihtout this. I think this causes a race since the tmpdir operations don't use the mutex so an in progress download would get deleted.

donoghuc: Chose to skip deleting anything in `tmppath` given that tmpdir operations don't use the mutex and should always be cleaned.

lib/bolt_server/file_cache.rb
```diff
@@ -166,7 +166,8 @@ def expire(purge_ttl)
       expired_time = Time.now - purge_ttl
       @cache_dir_mutex.with_write_lock do
         Dir.glob(File.join(@cache_dir, '*')).select { |f| File.directory?(f) }.each do |dir|
-          if (mtime = File.mtime(dir)) < expired_time
+          next unless (mtime = File.mtime(dir)) < expired_time
```
adreyer: this seems more readable as a single conditional statement
```
if (mtime = File.mtime(dir)) < expired_time &&  dir != tmppath
    @logger.debug("Removing #{dir}, last used at #{mtime}")
     FileUtils.remove_dir(dir)
end
```

PR #786, opened 2018-12-10
(BOLT-645) Support using device modules in apply statements

adreyer:
This updates the custom_facts and device_catalog tasks so that they
support running with device modules.

### DIFF
--- a/lib/bolt/applicator.rb
+++ b/lib/bolt/applicator.rb
@@ -43,25 +43,38 @@ module Bolt
       @custom_facts_task ||= begin
         path = File.join(libexec, 'custom_facts.rb')
         file = { 'name' => 'custom_facts.rb', 'path' => path }
-        metadata = { 'supports_noop' => true, 'input_method' => 'stdin' }
+        metadata = { 'supports_noop' => true, 'input_method' => 'stdin',
+                     'implementations' => [
+                       { 'name' => 'custom_facts.rb' },
+                       { 'name' => 'custom_facts.rb', 'remote' => true }
+                     ] }
         Bolt::Task.new(name: 'apply_helpers::custom_facts', files: [file], metadata: metadata)
       end
     end
 
     def catalog_apply_task
       @catalog_apply_task ||= begin
-        path = File.join(libexec, 'apply_catalog.rb')
-        file = { 'name' => 'apply_catalog.rb', 'path' => path }
-        metadata = { 'supports_noop' => true, 'input_method' => 'stdin' }
-        Bolt::Task.new(name: 'apply_helpers::apply_catalog', files: [file], metadata: metadata)
-      end
+                                path = File.join(libexec, 'apply_catalog.rb')
+                                file = { 'name' => 'apply_catalog.rb', 'path' => path }
+                                metadata = { 'supports_noop' => true, 'input_method' => 'stdin',
+                                             'implementations' => [
+                                               { 'name' => 'apply_catalog.rb' },
+                                               { 'name' => 'apply_catalog.rb', 'remote' => true }
+                                             ] }
+                                Bolt::Task.new(name: 'apply_helpers::apply_catalog', files: [file], metadata: metadata)
+                              end
     end
 
     def query_resources_task
       @query_resources_task ||= begin
         path = File.join(libexec, 'query_resources.rb')
         file = { 'name' => 'query_resources.rb', 'path' => path }
-        metadata = { 'supports_noop' => true, 'input_method' => 'stdin' }
+        metadata = { 'supports_noop' => true, 'input_method' => 'stdin',
+                     'implementations' => [
+                       { 'name' => 'query_resources.rb' },
+                       { 'name' => 'query_resources.rb', 'remote' => true }
+                     ] }
+
         Bolt::Task.new(name: 'apply_helpers::query_resources', files: [file], metadata: metadata)
       end
     end
--- a/libexec/apply_catalog.rb
+++ b/libexec/apply_catalog.rb
@@ -48,6 +48,31 @@ begin
     $LOAD_PATH << dir unless $LOAD_PATH.include?(dir)
   end
 
+  if (conn_info = args['_target'])
+    unless (type = conn_info['remote-transport'])
+      puts "Cannot execute a catalog for a remote target without knowing it's the remote-transport type."
+      exit 1
+    end
+
+    begin
+      require 'puppet/resource_api/transport'
+    rescue LoadError
+      msg = "Could not load 'puppet/resource_api/transport', puppet-resource_api "\
+            "gem version 1.8.0 or greater is required on the proxy target"
+      puts msg
+      exit 1
+    end
+
+    # Transport.connect will modify this hash!
+    transport_conn_info = conn_info.each_with_object({}) { |(k, v), h| h[k.to_sym] = v }
+
+    transport = Puppet::ResourceApi::Transport.connect(type, transport_conn_info)
+    Puppet::ResourceApi::Transport.inject_device(type, transport)
+
+    Puppet[:facts_terminus] = :network_device
+    Puppet[:certname] = conn_info['uri']
+  end
+
   # Ensure custom facts are available for provider suitability tests
   facts = Puppet::Node::Facts.indirection.find(SecureRandom.uuid, environment: env)
 
@@ -57,8 +82,11 @@ begin
              Puppet::Transaction::Report.new('apply')
            end
 
-  Puppet.override(current_environment: env,
-                  loaders: Puppet::Pops::Loaders.new(env)) do
+  overrides = { current_environment: env,
+                loaders: Puppet::Pops::Loaders.new(env) }
+  overrides[:network_device] = true if args['_target']
+
+  Puppet.override(overrides) do
     catalog = Puppet::Resource::Catalog.from_data_hash(args['catalog'])
     catalog.environment = env.name.to_s
     catalog.environment_instance = env
--- a/libexec/custom_facts.rb
+++ b/libexec/custom_facts.rb
@@ -29,7 +29,33 @@ Dir.mktmpdir do |puppet_root|
     $LOAD_PATH << dir unless $LOAD_PATH.include?(dir)
   end
 
+  if (conn_info = args['_target'])
+    unless (type = conn_info['remote-transport'])
+      puts "Cannot collect facts for a remote target without knowing the remote-transport type."
+      exit 1
+    end
+
+    begin
+      require 'puppet/resource_api/transport'
+    rescue LoadError
+      msg = "Could not load 'puppet/resource_api/transport', puppet-resource_api "\
+            "gem version 1.8.0 or greater is required on the proxy target"
+      puts msg
+      exit 1
+    end
+
+    # Transport.connect will modify this hash!
+    transport_conn_info = conn_info.each_with_object({}) { |(k, v), h| h[k.to_sym] = v }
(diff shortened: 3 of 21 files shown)


adreyer:
I need to clean up some tests and we probably don't want to release until we handle devices/remote_transports better.

libexec/custom_facts.rb
```diff
@@ -29,7 +30,33 @@
...
+    # TODO: validate_target
+    special_keys = %w[type debug uri]
+    connection = conn_info.reject { |k, _| special_keys.include?(k) }
+    device = OpenStruct.new(connection)
+    device.url = conn_info['uri']
+    device.provider = conn_info['type']
+    device.options = { debug: true } if conn_info['debug']
+    Puppet[:facts_terminus] = :network_device
+    Puppet[:certname] = device.name
+    Puppet::Util::NetworkDevice.init(device)
```
adreyer: This doesn't work well with modules that subclass the `NetworkDevice::Simple` from resource API because it means tasks that work like the panos PR are more flexible in terms of config format while apply only works when the name is set to the devices url. 

Solutions:
1. Do nothing. if you want to use apply you have to rely on the url, all `NetworkDevice` modules should document only this format. When the Transport API is added to to resource_api we can use more flexible config for them.
2. Check for subclasses of `NetworkDevice::Simple` and bypass `NetworkDevice.init` for them.

I think there is a related question of how should we differentiate "network_device" modules from "transport" based remote modules.

adreyer:
This work will be difficult to use without improvements in the resource_api https://github.com/puppetlabs/puppet-specifications/pull/133

libexec/apply_catalog.rb
```diff
@@ -48,7 +49,26 @@
...
+    # TODO: validate_target
+    special_keys = %w[type debug uri]
+    connection = conn_info.reject { |k, _| special_keys.include?(k) }
+    device = OpenStruct.new(connection)
+    device.url = conn_info['uri']
+    device.provider = conn_info['type']
+    device.options = { debug: true } if conn_info['debug']
+    Puppet[:facts_terminus] = :network_device
+    Puppet[:certname] = device.name
+    Puppet::Util::NetworkDevice.init(device)
```
adreyer: @DavidS I need a way to do this with a transport.

DavidS: now I get it, bolt can't use the transport *alone*, because it needs to hook up the rest of puppet :man_facepalming: that took me way too long.

I believe you should be able to say
```
device = OpenStruct.new
device.url = connection
```

and have everything work out with the backwards compatibility layers we're building in for plain puppet.

I don't think that's a particular appealing long-term solution.

Would the following make sense?

```
transport = Puppet::ResourceApi::Transport.connect(conn_info['type'], connection)
wrapper = Puppet::ResourceApi::Transport::Wrapper.new(transport) # needs the wrapper to be modified, easy
Puppet::Util::NetworkDevice.current = wrapper # needs puppet to be modified (or use instance_variable_set)
```

da-ar: OK having chatted to @DavidS we could update NetworkDevice to be:
`Puppet::Util::NetworkDevice.init_transport(transport)` which then wraps the transport and sets current.

adreyer: `Puppet::Util::NetworkDevice.init_transport(Puppet::ResourceApi::Transport.connect(conn_info['type'], connection))`
,
`Puppet::Util::NetworkDevice.init_transport(conn_info['type'], connection)`
`Puppet::ResourceApi::Transport.set_device(conn_info['type'], connection)`


Code in puppet-resource_api instead of puppet will prevent this from being blocked on a puppet release.

DavidS: I'm happy to not have to wait for a puppet release and just force-feed from the outside

If we want, we can also monkey-patch an `init_transport()` function into `Puppet::Util::NetworkDevice` for puppet versions that don't carry it natively.

adreyer: We try not to monkey patch puppet from bolt. I think it's ok to require a new agent version on the proxy. In most cases it will be localhost and bolt should just use the agent that ships with it anyway.

adreyer: But I'd like to merge this in and release it without having to wait for puppet release if possible

adreyer: Bolt no longer vendors puppet we now just have a gem dependency on it

DavidS: In that case bolt needs to pull in the `puppet-resource_api` gem on its own anyways, and we can release the rsapi gem whenever we need it.

libexec/apply_catalog.rb
```diff
@@ -62,7 +62,12 @@
 
     transport = Puppet::ResourceApi::Transport.connect(conn_info['type'], transport_conn_info)
     transport_wrapper = Puppet::ResourceApi::Transport::Wrapper.new(conn_info['type'], transport)
-    Puppet::Util::NetworkDevice.instance_variable_set(:@current, transport_wrapper)
+    if  Puppet::Util::NetworkDevice.respond_to?(:set_transport)
+      Puppet::Util::NetworkDevice.set_transport(conn_info['type'], transport_wrapper)
```
DavidS: either this takes a `Device`, then it should be called `set_device`, or if it is called `set_transport`, it should take a `Transport`.

adreyer: hmm it seems to have ended up taking a transport but being called set_device. Oh well.

libexec/custom_facts.rb
```diff
@@ -29,7 +30,28 @@
     $LOAD_PATH << dir unless $LOAD_PATH.include?(dir)
   end
 
+  if (conn_info = args['_target'])
+    unless (type = conn_info['remote-transport'])
+      puts "Cannot collect facts for a remote target without knowing the remote-transport type."
+      exit 1
+    end
+
+    require 'puppet/resource_api/transport'
```
adreyer: Put a guard around this

spec/fixtures/apply/device_test/lib/puppet/provider/fake_device/fake_device.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
```
adreyer: delete this file

adreyer:
[] update get_resources to work with devices
[] make sure resource_api 1.8.0 is in bolt vanagon

lib/bolt/applicator.rb
```diff
@@ -43,25 +43,38 @@ def custom_facts_task
       @custom_facts_task ||= begin
         path = File.join(libexec, 'custom_facts.rb')
         file = { 'name' => 'custom_facts.rb', 'path' => path }
-        metadata = { 'supports_noop' => true, 'input_method' => 'stdin' }
+        metadata = { 'supports_noop' => true, 'input_method' => 'stdin',
+                     'implementations' => [
+                       { 'name' => 'custom_facts.rb' },
```
donoghuc: Is it necessary to have both entries? From the task spec it seems like not https://github.com/puppetlabs/puppet-specifications/tree/master/tasks#options-2 
```
remote: This implementation is remote. Set remote on specific implementations if the task supports both normal and remote execution.(rev 4)
```

spec/fixtures/apply/device_test/lib/puppet/type/fake_device.rb
```diff
@@ -0,0 +1,31 @@
...
+
+require 'puppet/resource_api'
+
+Puppet::ResourceApi.register_type(
+  name: 'fake_device',
+  docs: 'fake device for testing puppet-device',
+  attributes: {
+    ensure: {
+      type: 'Enum[present, absent]',
+      desc: 'Whether this apt key should be present or absent on the target system.',
```
donoghuc: apt key?

spec/fixtures/modules/facts_test/plans/emit.pp
```diff
@@ -1,4 +1,4 @@
-plan facts::emit(String $host) {
+plan facts_test::emit(String $host) {
```
donoghuc: Nice. That name needed to be changed.

spec/integration/device_spec.rb
```diff
@@ -0,0 +1,113 @@
...
+    let(:device_inventory) do
+      device_group = { 'name' => 'device_targets',
+                       'nodes' => [
+                         { 'name' => 'fake_device1',
+                           'config' => {
+                             'transport' => 'remote',
+                             'remote' => {
+                               'remote-transport' => 'fake',
+                               'run-on' => 'puppet_6',
+                               'path' => device_path
```
donoghuc: This seems like an important pattern to document? Specifically passing data here and then accessing it according to the schema.You mentioned we should wait until we have a good example of a real world module to work on documentation.

spec/integration/device_spec.rb
```diff
@@ -0,0 +1,113 @@
...
+    end
+
+    after(:all) do
+      uninstall('puppet_6', inventory: agent_version_inventory)
+    end
+
+    context "when running against puppet 6" do
+      before(:all) do
+        install('puppet_6', inventory: agent_version_inventory)
+        cmd = 'yes | /opt/puppetlabs/puppet/bin/gem uninstall '\
```
donoghuc: Could add a TODO note that we can remove this once puppet-agent gets new resource API.

spec/integration/device_spec.rb
```diff
@@ -0,0 +1,113 @@
...
+              '-i /opt/puppetlabs/puppet/lib/ruby/vendor_gems puppet-resource_api'
```
donoghuc: It looks like the RSAPI is included as a vendored gem in puppet agent. It might be worth adding a comment or a TODO about removing this setup when the new version is shipped?

lib/bolt/applicator.rb
adreyer: ```Set remote on specific implementations if the task supports both normal and remote execution.```

Is about a task with a remote implemenation and a local implementation. This is a single implementation used for both.

spec/fixtures/apply/device_test/lib/puppet/type/fake_device.rb
adreyer: This was copied from panos. I can update the desc.

pre-docs/applying_manifest_blocks.md
```diff
@@ -276,9 +276,30 @@ Create a manifest that sets up a web server with IIS and run it as a plan.
...
+option and any other connnection info under the `remote` section of config.
+
+```
+---
+nodes:
+  - name: "https://username:password@panos-device.example.com"
+    config:
+      transport: remote
+      remote:
+        address: panos-device-example.com
```
donoghuc: For panos 1.0.0 they changed `address` to `host`. It can be a hostname or an IP. Specifying user and password in the URI is ok. I pretty much tried every combination of the config options in the snipped below from an inventoryfile. 
```yaml
  - name: http://admin:admin@xo9xrk9blv97nbs.delivery.puppetlabs.net
    alias: panos-8
    config:
      transport: remote
      remote:
        remote-transport: panos
        run-on: puppet-6
        host: xo9xrk9blv97nbs.delivery.puppetlabs.net
        # host: 10.32.115.95
        # user: admin
        # password: admin
        # apikey: LUFRPT14MW5xOEo1R09KVlBZNnpnemh0VHRBOWl6TGM9bXcwM3JHUGVhRlNiY0dCR0srNERUQT09
        ssl: false
```

donoghuc: Note: Setting `host` does not make sense as it will always get overwritten by the host in the node name.

PR #787, opened 2018-12-10
(BOLT-791) Don't allow remote targets to run on other remote targets.

adreyer:
Remote targets should always point to a non-remote target with run-on.
This prevents multiple levels of lookups in loops in trying to determine
the correct proxy and makes behavior clearer when run-on is a group.

### DIFF
--- a/lib/bolt/transport/remote.rb
+++ b/lib/bolt/transport/remote.rb
@@ -26,10 +26,15 @@ module Bolt
       end
 
       def get_proxy(target)
-        # TODO: This needs to have access to the inventory
         inventory = target.inventory
         raise "Target was created without inventory? Not get_targets?" unless inventory
-        inventory.get_targets(target.options['run-on'] || 'localhost').first
+        proxy = inventory.get_targets(target.options['run-on'] || 'localhost').first
+
+        if proxy.transport == 'remote'
+          msg = "#{proxy.name} is not a valid run-on target for #{target.name} since is also remote."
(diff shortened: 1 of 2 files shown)


lib/bolt/transport/remote.rb
```diff
@@ -26,10 +26,15 @@ def initialize(executor)
       end
 
       def get_proxy(target)
-        # TODO: This needs to have access to the inventory
         inventory = target.inventory
         raise "Target was created without inventory? Not get_targets?" unless inventory
-        inventory.get_targets(target.options['run-on'] || 'localhost').first
+        proxy = inventory.get_targets(target.options['run-on'] || 'localhost').first
+
+        if proxy.transport == 'remote'
```
lucywyman: This implies you can't have 2 "levels" of remotes:
```
  - name: vmpooler
    nodes:
     - xlrw1e4t6mrduyo.delivery.puppetlabs.net
    config:
      transport: remote
      remote:
        run-on: other_group

  - name: other_group
    nodes:
      - abcdef.example.com
    config:
      transport: remote
      remote:
        run-on: localhost:20022
```

Was this already invalid? It works for me locally on you PR without this patch.
```

MikaelSmith: That's kind of wild. Did it do what you'd expect? What would you expect?

adreyer: Nothing prevented it before but I don't see a good reason to support it. Remote targets can't run tasks so they shouldn't be used as a proxy targets.

PR #789, opened 2018-12-11
(maint) agent was not correctly uninstalled after tests

adreyer:
This is an issue because BOLT-997 means the the agent task won't work if facts are collected by facter instead of the shell facts task.

### DIFF
--- a/spec/integration/apply_spec.rb
+++ b/spec/integration/apply_spec.rb
@@ -52,11 +52,9 @@ describe "apply" do
 
     after(:all) do
       # TODO: Extract into test helper if needed in more files
-      uri = conn_uri('ssh')
-      inventory_data = conn_inventory
-      config_data = root_config
+      ssh_node = conn_uri('ssh', include_password: true)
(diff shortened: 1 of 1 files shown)


spec/integration/apply_spec.rb
```diff
@@ -52,11 +52,8 @@ def agent_version_inventory
 
     after(:all) do
       # TODO: Extract into test helper if needed in more files
-      uri = conn_uri('ssh')
-      inventory_data = conn_inventory
-      config_data = root_config
       uninstall = '/opt/puppetlabs/bin/puppet resource package puppet-agent ensure=absent'
-      run_command(uninstall, uri, config: config_data, inventory: inventory_data)
+      run_command(uninstall, 'agent_targets', config: root_config, inventory: agent_version_inventory)
```
MikaelSmith: I think line ~154 means we want to uninstall from everything.

adreyer: that would explain the failure in the facts module

PR #790, opened 2018-12-12
(BOLT-608) Set provided features for local transport  

donoghuc:
When using the local transport set `provided_features` to `shell` or `powershell` to reflect running on a windows or POSIX system. 

Run the local transport unit tests on windows bolt controllers.

### DIFF
--- a/spec/bolt/transport/shared_examples.rb
+++ b/spec/bolt/transport/shared_examples.rb
@@ -18,6 +18,7 @@ def posix_context
     stderr_command: ['ssh -V', /OpenSSH/],
     destination_dir: '/tmp',
     supported_req: 'shell',
+    extension: '.sh',
     unsupported_req: 'powershell',
     cat_cmd: 'cat',
     rm_cmd: 'rm -rf',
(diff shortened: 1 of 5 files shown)


spec/spec_helper.rb
```diff
@@ -36,6 +36,7 @@
   end
 
   config.filter_run_excluding appveyor_agents: true unless ENV['APPVEYOR_AGENTS']
+  config.filter_run_excluding windows: true unless ENV['BOLT_WINDOWS']
```
MikaelSmith: We generally do this via tags in the Rakefile, not separate environment variables.

MikaelSmith: Actually not sure why this is necessary at all.

spec/bolt/transport/shared_examples.rb
```diff
@@ -36,6 +37,25 @@ def posix_context
   }
 end
 
+def windows_context
```
MikaelSmith: Oh oops, I had a local copy of much of this stashed.

How does the stderr_command work?

donoghuc: I think ssh client sends that info to `stderr` with default https://en.wikibooks.org/wiki/OpenSSH/Print_version config. I guess Appveyor must have `openssh`, the win10 floaty box I developed tests on does. 
```
cas@cas-ThinkPad-T460p:~/working_dir/bolt$ irb
irb(main):001:0> require 'open3'
=> true
irb(main):002:0> Open3.capture3('ssh -V')
=> ["", "OpenSSH_7.2p2 Ubuntu-4ubuntu2.6, OpenSSL 1.0.2g  1 Mar 2016\n", #<Process::Status: pid 26948 exit 0>]
```

MikaelSmith: I'd expect this test to fail with the Nano Server VM.

spec/spec_helper.rb
MikaelSmith: Ah. I guess all the other tests run from macOS or Linux. Definitely a non-Windows view of development.

MikaelSmith:
I'm comfortable merging this as-is unless you want to try to come up with an alternative to OpenSSH.

PR #791, opened 2018-12-12
(docs) remove reference to unimplemented '_items' from example

adreyer:
_items was part of a pre rev1 task spec proposal and shouldn't appear in
documentation.


adreyer:
@melamos

PR #793, opened 2018-12-13
(docs) Include reccomendations for where files should go in a module

adreyer:


melamos:
Added to docs proper, with edits proposed in https://github.com/puppetlabs/bolt/pull/797

PR #794, opened 2018-12-13
(maint) Bump task helper modules

donoghuc:
The {ruby,python}_task_helper modules have been refactored to be structured under the `files` instead of the `lib` directory.


donoghuc:
Will update once `puppetlabs-reboot` has been released.

PR #795, opened 2018-12-13
(BOLT-1036) Include 'files' content from bundled modules

MikaelSmith:
Tasks can now make use of `files` content from other modules. Ensure the
bundled modules include the `files` directory, particularly for
`python_task_helper` and `ruby_task_helper`. Also include `locales` for
translations that might be used in plans.

We explicitly exclude other things to restrict this to plan and task
content. We'd rather manifests and other traditional Puppet content be
explicitly pulled in by users that want them.


MikaelSmith:
Acceptance tests in this area are a little tricky because they'd only apply to testing with packages, not testing from git.

PR #796, opened 2018-12-14
(BOLT-954) Support bolt apply in BoltSpec

donoghuc:
This commit adds the ability to use `bolt apply` in `BoltSpec`. An `apply_manifest` function has been added to `BoltSpec::Run` module to allow applying manifest code from a file or from a string with the `:execute` option.

### DIFF
--- a/lib/bolt_spec/run.rb
+++ b/lib/bolt_spec/run.rb
@@ -9,7 +9,6 @@ require 'bolt/puppetdb'
 require 'bolt/util'
 
 # This is intended to provide a relatively stable method of executing bolt in process from tests.
-# Currently it provides run_task, run_plan, run_script and run_command helpers.
 module BoltSpec
   module Run
     def run_task(task_name, targets, params = nil, config: nil, inventory: nil)
@@ -47,6 +46,27 @@ module BoltSpec
       end
       result = result.to_a
       Bolt::Util.walk_keys(result, &:to_s)
+    end
+
+    def apply_manifest(manifest, targets, execute: false, noop: false, config: nil, inventory: nil)
+      # The execute parameter is equivalent to the --execute option
+      if execute
+        code = manifest
+      else
+        begin
+          unless File.stat(manifest).readable?
+            raise BOLT::FileError.new("The manifest '#{manifest}' is unreadable", manifest)
+          end
+        rescue Errno::ENOENT
+          raise Bolt::FileError.new("The manifest '#{manifest}' does not exist", manifest)
+        end
+        code = File.read(File.expand_path(manifest))
+        filename = manifest
+      end
+      result = BoltRunner.with_runner(config, inventory) do |runner|
+        runner.apply_manifest(code, targets, filename, noop)
+      end
+      JSON.parse(result.to_json)
     end
 
     class BoltRunner
@@ -106,6 +126,20 @@ module BoltSpec
         executor = Bolt::Executor.new(config.concurrency, @analytics)
         targets = inventory.get_targets(targets)
         executor.run_script(targets, script, arguments, options)
+      end
+
+      def apply_manifest(code, targets, filename = nil, noop = false)
+        ast = pal.parse_manifest(code, filename)
+        executor = Bolt::Executor.new(config.concurrency, @analytics, noop)
+        targets = inventory.get_targets(targets)
+
+        pal.in_plan_compiler(executor, inventory, puppetdb_client) do |compiler|
+          compiler.call_function('apply_prep', targets)
+        end
+
+        pal.with_bolt_executor(executor, inventory, puppetdb_client) do
+          Puppet.lookup(:apply_executor).apply_ast(ast, targets, '_catch_errors' => true, '_noop' => noop)
+        end
       end
     end
   end
--- a/spec/bolt_spec/run_spec.rb
+++ b/spec/bolt_spec/run_spec.rb
@@ -3,6 +3,7 @@
 require 'spec_helper'
 require 'bolt_spec/conn'
 require 'bolt_spec/run'
+require 'bolt_spec/files'
 
 # In order to speed up tests there are only ssh versions of these specs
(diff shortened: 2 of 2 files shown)


spec/bolt_spec/run_spec.rb
```diff
@@ -81,4 +83,39 @@
       expect(result['value']['kind']).to eq('bolt/run-failure')
     end
   end
+
+  describe 'apply_manifest' do
```
donoghuc: These tests will require a different config (there is a `facts` module on the modulepath and it tries to install puppet agent with `bolt` user). Will need to update.

spec/bolt_spec/run_spec.rb
```diff
@@ -69,16 +69,70 @@
...
+
+  context 'with a target that has a puppet-agent installed' do
+    def run_root_config
+      { 'ssh' => root_config['ssh'] }
+    end
+
+    before(:all) do
+      result = run_task('puppet_agent::version', 'ssh', inventory: conn_inventory, config: run_root_config)
+      expect(result.first['status']).to eq('success')
+      unless result.first['result']['version']
```
donoghuc: I encountered a bug in the `puppet_agent::install`/`facts` tasks when running against the `rastasheep/ubuntu-sshd:16.04` container we build off of for `bolt-spec`.

When facter is not installed the `facts::bash` task determines the platform and release for 16.04 as expected with 
```
root@2b504359b3b0:/# grep "^NAME" /etc/os-release | cut -d'=' -f2 | sed "s/\"//g"
Ubuntu
root@2b504359b3b0:/# grep "^VERSION_ID" /etc/os-release | cut -d'=' -f2 | sed "s/\"//g"
16.04
```
The results of running facter are different than those using the shell code.
```
root@2b504359b3b0:/# /opt/puppetlabs/bin/facter --yaml os.name
os.name: Debian
root@2b504359b3b0:/# /opt/puppetlabs/bin/facter --yaml os.release.full
os.release.full: stretch/sid
```
This causes an error in the constructing the target download of the `.deb`. 
```
expected: http://apt.puppetlabs.com/puppet-release-xenial.deb
observed: http://apt.puppetlabs.com/puppet-release-.deb
```
Looks like this has been reported as well at https://tickets.puppetlabs.com/browse/BOLT-997

MikaelSmith: Returning Debian on Ubuntu sounds very wrong.

spec/lib/bolt_spec/conn.rb
```diff
@@ -56,5 +56,35 @@ def conn_inventory
       end
       { "groups" => groups }
     end
+
+    def root_config
+      { 'modulepath' => File.join(__dir__, '../fixtures/apply'),
```
MikaelSmith: We shouldn't hard-code this modulepath in the Conn module.

spec/bolt_spec/run_spec.rb
```diff
@@ -69,16 +69,70 @@
...
+        result = run_task('puppet_agent::install', 'ssh', inventory: conn_inventory, config: run_root_config)
```
MikaelSmith: I'd be comfortable running these tests on a single target rather than making sure we get puppet 5 and puppet 6.

MikaelSmith: Maybe that's what we're doing with the `ssh` target?

lib/bolt_spec/run.rb
```diff
@@ -49,6 +48,27 @@ def run_script(script, targets, arguments = nil, options = {}, config: nil, inve
       Bolt::Util.walk_keys(result, &:to_s)
     end
 
+    def apply_manifest(manifest, targets, execute: false, noop: false, config: nil, inventory: nil)
+      # The execute parameter is equivalent to the --execute option
+      if execute
```
MikaelSmith: From a testing perspective it seems weird to design around `bolt apply ...`. If you're testing something with Bolt, it's probably a plan. But from a usable API that, it seems like a useful feature to include.

lib/bolt_spec/run.rb
```diff
@@ -107,6 +127,23 @@ def run_script(script, targets, arguments = nil, options = {})
         targets = inventory.get_targets(targets)
         executor.run_script(targets, script, arguments, options)
       end
```
MikaelSmith: Is indentation messed up here?

lib/bolt_spec/run.rb
```diff
@@ -107,6 +127,23 @@ def run_script(script, targets, arguments = nil, options = {})
...
+
+      def apply_manifest(code, targets, filename = nil, noop = false)
+        ast = pal.parse_manifest(code, filename)
+
+        executor = Bolt::Executor.new(config.concurrency, @analytics, noop)
+        executor.start_plan(nil)
```
MikaelSmith: Why is this necessary for apply but nothing else?

spec/lib/bolt_spec/conn.rb
donoghuc: Yeah, given that we are not going to run against puppet 5 and puppet 6 targets I don't think it makes sense to share these configs in the `Conn` class.

spec/bolt_spec/run_spec.rb
donoghuc: Yeah, I was originally going to test against puppet 5 and puppet 6 but went with just testing against the single `ubuntu_node` `bolt-spec` image.

lib/bolt_spec/run.rb
donoghuc: Are you suggesting we do not implement the `execute` option? I agree that if you are not applying a manifest file the puppet code makes more sense in an apply block.

MikaelSmith: I think what's here is ok. I'm not sure what the better default is though. Thinking about this plan-centrically the default should be `execute: true`. But I'm not really close to the people planning to use this.

PR #797, opened 2018-12-14
Helper file location details

melamos:
PR-793 got a little confusing because it was merged at the same time as remote tasks, which was inserted -- I think -- in the wrong place on this page. Please confirm the location of this added helper file detail, as well as the formatting change and edit. 

This is already in docs, if you'd like to confirm there too: https://puppet.com/docs/bolt/1.x/writing_tasks.html

### DIFF
--- a/pre-docs/writing_tasks.md
+++ b/pre-docs/writing_tasks.md
@@ -361,18 +361,18 @@ Successful on 1 node: localhost
 Ran on 1 node in 0.12 seconds
 ```
 
-In general helper files should go in the `files` directory of a module to
-prevent them from being added to the puppet ruby loadpath or seen as tasks. However
-if a task helper can also be called as a task on it's own it belongs in the `tasks`
-directory. If it is ruby code that will be reused by types, providers or
-puppet functions it should go in the `lib` directory.
-
 
 
 
 ### Task Helpers
 
 To simplify writing tasks, Bolt includes [python\_task\_helper](https://github.com/puppetlabs/puppetlabs-python_task_helper) and [ruby\_task\_helper](https://github.com/puppetlabs/puppetlabs-ruby_task_helper). It also makes a useful demonstration of including code from another module.
+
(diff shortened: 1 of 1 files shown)


pre-docs/writing_tasks.md
```diff
@@ -361,19 +361,19 @@ Successful on 1 node: localhost
...
 
 
 
 ### Task Helpers
 
 To simplify writing tasks, Bolt includes [python\_task\_helper](https://github.com/puppetlabs/puppetlabs-python_task_helper) and [ruby\_task\_helper](https://github.com/puppetlabs/puppetlabs-ruby_task_helper). It also makes a useful demonstration of including code from another module.
 
+Follow these guidelines to determine where to store helper files:
+
+- `files` directory of a module -- Most helper files. This prevents the file from being treated as a task or added to the Puppet Ruby loadpath.
```
MikaelSmith: Looks like this didn't copy the long hyphen. Is that a reflection of something on the docs site?

MikaelSmith:
This was a set of general guidelines for where to put shared code. I think it still makes sense in the original location. The section you moved it to is specific to helpers we created, where the recommendation isn't directly actionable.

melamos:
Okay, sounds like you're okay with the edit, but I got it in the wrong spot. The concern I had with the original location in PR-793 was that it was nested at the bottom of the "Sharing task code" topic, after a bunch of codeblocks, where it seemed like it might be easily missed. 

If we leave it in the original location right before "Task helpers," how about giving it a heading, like: 
### Helper file location
?

pre-docs/writing_tasks.md
melamos: If I remember right, double-hyphens in Markdown get turned into em dashes on our docs site, so I think this is workable Markdown, if a little wonky. :)

MikaelSmith:
How about move it up around line 298, where we first introduce those locations.

Issue #798, opened 2018-12-17
Support specifying configuration file to load via environment variables.


icesmurf:
1) I'd like to be able to specify BOLT_CONFIG and have bolt load whatever is in my BOLT_CONFIG environment variable. 
2) I'd like -c command line parameter to override any environment variables.

This means I can have one install of bolt, and have different projects easily segregated, being able to configure shared options via config if needed.

I use direnv extensively for automation task segregation.

lucywyman:
Hi @icesmurf - does using a Boltdir not work for your use case? https://puppet.com/docs/bolt/1.x/configuring_bolt.html If you want to configure projects separately that's the best way to do it, but if it doesn't fit your use case we'd be interested to hear more details about how you're using Bolt.

adreyer:
What values are you going to store in you're configuration file?
What if any bolt information is going to be project specific?

icesmurf:
Hi Lucy, i want to be able to override the Boltdir locally if need be without impacting a particular repo, For example, if i've pulled down some code that contains a Boltdir, and it's using module/task xyz, then i'd like the ability to override that in a local config using an environment variable, Also for cases in testing, where i might want to force the test/job to use a specific known set of parameters/hosts.

lucywyman:
See @adreyer's comment on #799 - it might make more sense to have boltdir be configurable via env vars. It might also be useful to move this to #bolt channel on https://puppetcommunity.slack.com if you'd like to discuss in more real time :) Thanks for raising the issue!

icesmurf:
Yeah thinking about it i'm thinking boldir from an environment variable sounds more "boltish" and that'd satisfy my usecases just as well.

icesmurf:
https://tickets.puppetlabs.com/browse/BOLT-718 satisfies this i think.

PR #799, opened 2018-12-17
Add BOLT_CONFIG environment variable support.

icesmurf:
Simple change to add BOLT_CONFIG environment variable to provide an alternative to the command line options. #798


adreyer:
given the interaction between botldir and config/inventory/modules I think it makes more sense to allow BOLTDIR to be set with an environment variable. 

Would that solve your use case?

adreyer:
If that would solve your usecase it should be added: https://github.com/puppetlabs/bolt/blob/master/lib/bolt/boltdir.rb#L19

It is already ticketed https://tickets.puppetlabs.com/browse/BOLT-718 . So please add `(BOLT-718)` to the beginning of your commit message.

This will also need a test.

PR #801, opened 2018-12-17
(BOLT-1008) Add applicator to plan_executor service.

adreyer:
This implementation raises a not implemented error for all functionality
to prevent plans from compiling catalogs locally or sending plugins as
task parameters.


lib/plan_executor/applicator.rb
```diff
@@ -0,0 +1,30 @@
...
+      raise_not_implemented("apply")
+    end
+
+    def build_plugin_tarball
+      raise_not_implemented("build_plugin_tarball")
+    end
+
+    def custom_facts_task
+      raise_not_implemented('custom_facts_task')
+    end
```
MikaelSmith: Need query_resources_task as well now. Although I'm not sure why you specifically want to avoid those, it seems like we could inherit from `Bolt::Applicator` and only override `apply`.

adreyer: We don't want to upload the plugintarballs over orchestrator.

MikaelSmith: Still makes sense to error on `query_resources_task`, although `build_plugin_tarball` will error first.

PR #802, opened 2018-12-18
(BOLT-1047) Add short commands

lucywyman:
**What this changes** Adds shortened command syntax for description, modulepath, inventoryfile, transport, and verbose.
**Why** So they're faster to type.

### DIFF
--- a/lib/bolt/bolt_option_parser.rb
+++ b/lib/bolt/bolt_option_parser.rb
@@ -214,7 +214,7 @@ Usage: bolt apply <manifest.pp> [options]
              'Maximum number of simultaneous manifest block compiles (default: number of cores)') do |concurrency|
         @options[:'compile-concurrency'] = concurrency
       end
-      define('--modulepath MODULES',
+      define('-m', '--modulepath MODULES',
              "List of directories containing modules, separated by '#{File::PATH_SEPARATOR}'") do |modulepath|
         # When specified from the CLI, modulepath entries are relative to pwd
         @options[:modulepath] = modulepath.split(File::PATH_SEPARATOR).map do |moduledir|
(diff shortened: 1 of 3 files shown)


lib/bolt/bolt_option_parser.rb
```diff
@@ -150,7 +150,7 @@ def initialize(options)
       define('--noop', 'Execute a task that supports it in noop mode') do |_|
         @options[:noop] = true
       end
-      define('--description DESCRIPTION',
+      define('-d', '--description DESCRIPTION',
```
adreyer: Does this really get used?

lucywyman: That's fair - I'll remove it.

lucywyman: The other one I wasn't sure about was `-t`. Transport isn't very long and also probably doesn't get used very often. `-m` is the one I type the most and is longest, so it's the one I feel most strongly about, and `-i` is also long.

adreyer: I don't have strong opinions on those. -t seems like the least useful of them?

why do you use -m? Is it because you're running bolt from source or is it a situation a user would like find themselves in? IE not running near a boltdir or with the normal modulepath configured for your user?

lucywyman: It's usually because I don't want to modify the configured modulepath in a file, but just want to do one-off runs with a module that's not in my modulepath. Granted I keep my modulepath short because otherwise it really clutters up `bolt task show`, but like for testing a task or plan in the spec directory I don't want to add the spec fixtures to my modulepath but do want to run the task to make sure it works.

But it's not running bolt from source.

MikaelSmith:
I think I'm ok with these, and don't see anything they'd obviously conflict with in the future.

MikaelSmith:
Might be useful to have a ticket for @melamos to highlight the docs changes.

PR #808, opened 2018-12-20
(DOCS) fix require in ruby_task_helper example

adreyer:


pre-docs/writing_tasks.md
```diff
@@ -430,7 +430,7 @@ Create task and metadata in a new module at `~/.puppetlabs/bolt/site/mymodule/ta
 
 ```
 #!/usr/bin/env ruby
-require_relative '../files/task_helper.rb'
+require_relative '../../files/task_helper.rb'
```
adreyer: @melamos the second `'` is missing from this example on the docs site as well

MikaelSmith: This only worked when running from the same module. Needs to include the module name as well, like https://github.com/puppetlabs/puppetlabs-ruby_task_helper#usage shows.

melamos:
Docs updated

PR #809, opened 2018-12-20
(BOLT-1054) Create a Bolt standard library of plan functions

MikaelSmith:
Add a set of functions suitable to a plan standard library. Breaks them
into namespaced domains: collection (coll), control-flow (ctrl), file,
and system.


bolt-modules/ctrl/lib/puppet/functions/ctrl/do_until.rb
```diff
@@ -0,0 +1,18 @@
...
+  #     run_task('test', $target, _catch_errors => true).ok?
+  #   }
+  dispatch :do_until do
+    block_param
+    return_type 'Undef'
+  end
+
+  def do_until
+    until yield; end
+    nil
```
adreyer: It might be useful to return the value on this one since it could be anything truthy. It might also behave unexpectedly....

MikaelSmith: Ah, that could be useful.

bolt-modules/coll/lib/puppet/functions/coll/group_by.rb
```diff
@@ -0,0 +1,20 @@
...
+
+# Groups the collection by result of the block. Returns a hash where the keys are the evaluated result from the block
+# and the values are arrays of elements in the collection that correspond to the key.
+Puppet::Functions.create_function(:'coll::group_by') do
+  # @param collection A collection of things to group.
+  # @example Group targets by protocol, results in e.g. { ssh => [target1, target2], winrm => [target3]}
+  #   get_targets($nodes).group_by |$target| {
+  #     $target.protocol
+  #   }
+  dispatch :group_by do
```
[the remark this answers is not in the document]
MikaelSmith: https://github.com/puppetlabs/puppet/pull/7310

bolt-modules/file/lib/puppet/functions/file/read.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+# Read a file and return its contents.
+Puppet::Functions.create_function(:'file::read', Puppet::Functions::InternalFunction) do
```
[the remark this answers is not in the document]
MikaelSmith: Yes, but I probably won't get to that until January. So we should feel comfortable merging without that if we want to do a release before then.

bolt-modules/ctrl/lib/puppet/functions/ctrl/sleep.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+# Sleeps for specified number of seconds.
+Puppet::Functions.create_function(:'ctrl::sleep') do
+  # @param period Time to sleep (in seconds)
+  # @example Sleep for 5 seconds
+  #   ctrl::sleep(5)
+  dispatch :sleeper do
+    required_param 'Integer', :period
```
nicklewis: I think this could just be a `Numeric`

bolt-modules/file/lib/puppet/functions/file/read.rb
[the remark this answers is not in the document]
adreyer: I dont know if we'll release again this year but I didn't intend this to block merging

bolt-modules/ctrl/lib/puppet/functions/ctrl/do_until.rb
adreyer: I agree but am not sure. I'll ask for feedback in slack

MikaelSmith: Explore how `break` works in these, and whether it can be used to return a value. If we can use break to return a value, a loop might be a better alternative.

MikaelSmith: Explore loop-recur:
```
loop || {
  <do something>
  if true { break }
}
```
or
```
loop(0) |$updated_val| {
  if $updated_val > 2 { break $updated_val }
  $updated_val + 1
}
```

MikaelSmith: Break can't return a value. So I lean towards continuing with `do_while` and `do_until`.

MikaelSmith: `break` works, but has the interesting side-effect of causing `do_until` to return `"break"`.

PR #810, opened 2018-12-20
(WIP)(maint) Query remote WinRM host for path separator

Iristyle:
- When running Bolt over WinRM against Linux hosts, a temp file path to write
   scripts is created that hardcodes a path like:

   dir\\filename

 - This path is not friendly to Linux based hosts.

   Instead, query for the separator when initializing a Bolt WinRM connection
   using PowerShells [System.IO.Path]::DirectorySeparatorChar and save it
   locally. When constructing / joining remote paths, use this value.

### DIFF
--- a/lib/bolt/transport/winrm.rb
+++ b/lib/bolt/transport/winrm.rb
@@ -55,6 +55,7 @@ module Bolt
         begin
           conn&.disconnect
         rescue StandardError => ex
+          require 'pry'; binding.pry
           logger.info("Failed to close connection to #{target.uri} : #{ex.message}")
         end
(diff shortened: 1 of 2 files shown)


Iristyle:
Still diagnosing issues with the protocol handling between WinRM gem and OMI server. Further fixes may well land in those projects.

Iristyle:
I should probably clean this up so we can merge at least the path separator change. This is necessary for PowerShell support on non-Windows -- it's just that there's a bunch more work to make that work as well, and I haven't invested any time to figured out what yet.

PR #811, opened 2018-12-21
(BOLT-1056) Add warning about using shell one-liners

adreyer:
@melamos This is really painful when users run into it. I'm not sure how to make it as discoverable as possible.

### DIFF
--- a/pre-docs/running_bolt_commands.md
+++ b/pre-docs/running_bolt_commands.md
@@ -38,6 +38,14 @@ bolt command run "netstat -an | grep 'tcp.*LISTEN'" --nodes web5.mydomain.edu, w
 
     **Note:** When connecting to Bolt hosts over WinRM that have not configured SSL for port 5986, passing the `--no-ssl` switch is required to connect to the default WinRM port 5985.
 
+## Running commands with redirection or pipes
+
+When you run one-line commands that include redirection or pipes, pass `bash`
+or another shell as the command. Using a shall ensures that the one-liner is
+run as a single command and that it works correctly with `run-as`. For example,
+instead of `bolt command run "echo foo > /root/foo" --run-as root`, use `bolt
+command run "bash -c 'echo foo > /root/foo'" --run-as root`.
+
 
 ## Run a script on remote nodes
 


melamos:
@adreyer If we tag it as a note with the "important" attribute, it will at least get flashier styling. Alternatively, we could make it a separate topic, "Running commands with redirection or pipes"(?). This would surface it in the top-of-page table of contents, make it linkable, and improve SEO. I'm open to either solution, just depends how much emphasis you think it needs.

adreyer:
@melamos Lets make it a separate topic

pre-docs/running_bolt_commands.md
```diff
@@ -38,6 +38,13 @@ bolt command run "netstat -an | grep 'tcp.*LISTEN'" --nodes web5.mydomain.edu, w
 
     **Note:** When connecting to Bolt hosts over WinRM that have not configured SSL for port 5986, passing the `--no-ssl` switch is required to connect to the default WinRM port 5985.
 
+    **Warning:** When running shell one-liners that include redirection or
+    pipes it is important to pass `bash` or another shell as the command.
+    Failure to do this will cause the one-liner to run as separate commands and not work
+    correctly with `run-as`. For example do not run `bolt command run "echo foo
+    > /root/foo" --run-as root` instead pass the one-liner as single string to
+    `bash` with `bolt command run "bash 'echo foo > /root/foo'" --run-as root`.
```
MikaelSmith: Does that need to be `bash -c`?

pre-docs/running_bolt_commands.md
```diff
@@ -38,6 +38,13 @@ bolt command run "netstat -an | grep 'tcp.*LISTEN'" --nodes web5.mydomain.edu, w
 
     **Note:** When connecting to Bolt hosts over WinRM that have not configured SSL for port 5986, passing the `--no-ssl` switch is required to connect to the default WinRM port 5985.
 
+    **Warning:** When running shell one-liners that include redirection or
```
melamos:
```suggestion
## Running commands with redirection or pipes

When you run one-line commands that include redirection or pipes, pass `bash` or another shell as the command. Using a shall ensures that the one-liner is run as a single command and that it works correctly with `run-as`. For example, instead of `bolt command run "echo foo > /root/foo" --run-as root`, use `bolt command run "bash 'echo foo > /root/foo'" --run-as root`.
```

melamos:
Added to docs:
https://puppet.com/docs/bolt/1.x/running_bolt_commands.html#concept-325

PR #815, opened 2019-01-02
(maint) Correct `transports` reference in inventoryfile

lucywyman:
**What this changes** Replaces incorrect `transports` reference with correct transport-specific config option


melamos:
Moved to docs proper.

PR #817, opened 2019-01-03
Update links in writing_plans.md

melamos:
I can't add related info links to a separate doc set (i.e., from Bolt to Puppet). Is this rewrite using inline links acceptable?


melamos:
Moved to docs.

PR #820, opened 2019-01-04
(BOLT-1007) Create plan executor class for plan_executor service

lucywyman:
This adds a new executor class for the `plan_executor` service, which bypasses logic around batching and queueing and exclusively uses the orchestrator transport. It also includes a new api transport class based on the orchestrator transport, which doesn't call `start_plan` as that's handled by the orchestrator.

Giving the plan executor service it's own executor greatly simplifies the logic around running jobs through the service.

### DIFF
--- a/lib/bolt/transport/api/connection.rb
+++ b/lib/bolt/transport/api/connection.rb
@@ -1,0 +1,70 @@
+# frozen_string_literal: true
+
+# This is a copy of the orchestrator transport connection, but without the
+# 'start_plan' call in the init function which is handled by the orchestrator
+# client
+module Bolt
+  module Transport
+    class Api < Orch
+      class Connection
+        attr_reader :logger, :key
+
+        CONTEXT_KEYS = Set.new(%i[plan_name description params]).freeze
+
+        def self.get_key(opts)
+          [
+            opts['service-url'],
+            opts['task-environment'],
+            opts['token-file']
+          ].join('-')
+        end
+
+        def initialize(opts, logger)
+          @logger = logger
+          @key = self.class.get_key(opts)
+          client_keys = %w[service-url token-file cacert]
+          client_opts = client_keys.each_with_object({}) do |k, acc|
+            acc[k] = opts[k] if opts.include?(k)
+          end
+          client_opts['User-Agent'] = "Bolt/#{VERSION}"
+          logger.debug("Creating orchestrator client for #{client_opts}")
+
+          @client = OrchestratorClient.new(client_opts, true)
+          @environment = opts["task-environment"]
+        end
+
+        def finish_plan(plan_result)
+          if @plan_job
+            @client.command.plan_finish(
+              plan_job: @plan_job,
+              result: plan_result.value || '',
+              status: plan_result.status
+            )
+          end
+        end
+
+        def build_request(targets, task, arguments, description = nil)
+          body = { task: task.name,
+                   environment: @environment,
+                   noop: arguments['_noop'],
+                   params: arguments.reject { |k, _| k.start_with?('_') },
+                   scope: {
+                     nodes: targets.map(&:host)
+                   } }
+          body[:description] = description if description
+          body[:plan_job] = @plan_job if @plan_job
+          body
+        end
+
+        def run_task(targets, task, arguments, options)
+          body = build_request(targets, task, arguments, options['_description'])
+          @client.run_task(body)
+        end
+
+        def query_inventory(targets)
+          @client.post('inventory', nodes: targets.map(&:host))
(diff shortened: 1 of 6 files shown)


lib/bolt/transport/orch.rb
```diff
@@ -37,7 +37,8 @@ def self.validate(options); end
 
       def initialize(*args)
         @connections = {}
-        super
+        @from_api = args.select { |arg| arg[:from_api] == true }.any?
```
adreyer: We should probably avoid this pattern and subclass the transport. This adds complexity to the live bolt code for a prototype which is bad and long term it seems likely the normal transport and the orchestrator executor will diverge.

lib/plan_executor/executor.rb
```diff
@@ -0,0 +1,301 @@
...
+                   noop = nil,
+                   bundled_content: nil,
+                   load_config: true)
+      @bundled_content = bundled_content
+      @logger = Logging.logger[self]
+      @plan_logging = false
+      @load_config = load_config
+      @noop = noop
+      @run_as = nil
+      @pool = if concurrency > 0
```
adreyer: I don't think we need threading here because each puma thread has it's own executor(and instance of the app class) right?

adreyer: Assuming thats the case I think we can write a much much simpler executor implementation that just has an "orchestrator transport" and implements only the run_* functions which call the transport directly. All the threading and queueing is too support ssh/winrm etc.

lib/plan_executor/executor.rb
```diff
@@ -0,0 +1,218 @@
...
+        results = yield
+      ensure
+        duration = Time.now - start_time
+        @logger.send(log_method, "Finished: plan #{plan_name} in #{duration.round(2)} sec")
+      end
+
+      results
+    end
+
+    # CODEREVIEW Do we still need this?
```
adreyer: no because the orchestrator client doesn't implement it.

lib/plan_executor/executor.rb
```diff
@@ -0,0 +1,218 @@
...
+                       results = yield
+                     rescue StandardError => e
+                       @logger.warn(e)
+                       # CODEREVIEW how should we fail if there's an error?
+                       Array(Bolt::Result.from_exception(targets[0], e))
+                     end
+      Bolt::ResultSet.new(result_array)
+    end
+
+    def log_action(description, targets)
```
adreyer: these log levels are probably wrong for a service. I'm tempted to remove it and add back in service specific logging later. If it starts to cause problems its' probably worth just removing.

adreyer: In general "plan logging" should not be part of the logger but about sending log events to orchestrator in a plan service. run_* events are already handled by the task command already so removing this would only effect "run_plan"

lib/bolt/transport/api/connection.rb
```diff
@@ -0,0 +1,70 @@
+# frozen_string_literal: true
+
+# This is a copy of the orchestrator transport connection, but without the
+# 'start_plan' call in the init function which is handled by the orchestrator
+# client
+module Bolt
+  module Transport
+    class Api < Base
```
adreyer: This shouldn't work! Even if it does we need to extend the same base class in both declarations.

lib/plan_executor/executor.rb
```diff
@@ -0,0 +1,199 @@
...
+  class Executor
+    attr_reader :noop, :transport
+
+    def initialize(noop = nil,
+                   bundled_content: nil,
+                   load_config: true)
+      @bundled_content = bundled_content
+      @logger = Logging.logger[self]
+      @plan_logging = false
+      @load_config = load_config
```
adreyer: this is only used by the ssh transport so it should probably be left off.  I think it's just a weird way to test ssh.

lib/bolt/transport/api/connection.rb
lucywyman: What shouldn't work? I'm not sure what you mean by 'both declarations'? I was following the pattern in https://github.com/puppetlabs/bolt/blob/master/lib/bolt/transport/orch/connection.rb#L5 - let me know if you want to bluejeans about it.

lucywyman: Actually, separately, this relies on https://github.com/puppetlabs/bolt/blob/master/lib/bolt/transport/orch.rb#L60 to create new connections. I think it'd be more clear to explicitly create the connection (`Bolt::Transport::Orch::Connection.new`) and overload the method in API to create an API connection. But maybe it just adds unnecessary complexity and copy-pasting?

PR #821, opened 2019-01-04
(BOLT-938) Failover to backup puppetdb instance

adreyer:
This adds support for retrying puppetdb connections when there are more
one entry in server_urls.

This also adds two new containers to the test environment so that bolt
can be tested against a real puppetdb instance.


lib/bolt/puppetdb/client.rb
```diff
@@ -41,22 +45,35 @@ def facts_for_node(certnames)
 
       def make_query(query, path = nil)
         body = JSON.generate(query: query)
-        url = "#{@config.uri}/pdb/query/v4"
+        url = "#{uri}/pdb/query/v4"
         url += "/#{path}" if path
 
         begin
           response = http_client.post(url, body: body, header: headers)
+        rescue SocketError, OpenSSL::SSL::SSLError, SystemCallError, Net::ProtocolError, IOError => err
```
adreyer: These appear to be the only errors that pdb fails over for. I'm not sure why we wouldn't just fail over on any StandardError though.

adreyer:
The ssl-setup script is copied and very slightly modified from a PR against pdb's docker work which it turn copies it from the the pdb package https://github.com/puppetlabs/puppetdb/pull/2738.

I think we should look at using pupperware for this eventually but I have concerns about stability and speed for now.

lib/bolt/puppetdb/client.rb
lucywyman: The only reason I have for not failing over and StandardError is if there are a lot of server urls, and the issue is going to cause them all to fail, failing fast rather than having to wait and try each. That doesn't seem like it will be common though, and I think it'll be more annoying to expect failover for any errors and have that not happen.

spec/Dockerfile.puppetdb
```diff
@@ -0,0 +1,22 @@
+FROM puppet/puppetdb
+
+ARG hostname="bolt-puppetdb"
+
+#ENV PUPPETSERVER_HOSTNAME "$hostname"
+
```
lucywyman:
```suggestion
```

lucywyman: Just an extra newline

spec/Dockerfile.puppetdb
```diff
@@ -0,0 +1,22 @@
+FROM puppet/puppetdb
+
+ARG hostname="bolt-puppetdb"
+
+#ENV PUPPETSERVER_HOSTNAME "$hostname"
```
lucywyman: Is there a reason this is commented? I don't see harm in setting it and it looks like it'll be useful for https://github.com/puppetlabs/puppetdb/pull/2738#discussion_r239637958

spec/Dockerfile.puppetdb
```diff
@@ -0,0 +1,22 @@
...
+
+# Use our own certs so this doesn't have to wait for puppetserver startup
+COPY fixtures/ssl/ca.pem /etc/puppetlabs/puppet/ssl/certs/ca.pem
+COPY fixtures/ssl/cert.pem /etc/puppetlabs/puppet/ssl/certs/pdb.pem
+COPY fixtures/ssl/key.pem /etc/puppetlabs/puppet/ssl/private_keys/pdb.pem
+COPY fixtures/ssl/crl.pem /etc/puppetlabs/puppet/ssl/ca/ca_crl.pem
+
+COPY fixtures/puppetdb/ssl-setup.sh /opt/puppetlabs/bin/ssl-setup.sh
+
+RUN opt/puppetlabs/bin/ssl-setup.sh && chown -R puppetdb:puppetdb /etc/puppetlabs/puppetdb/ssl/
```
lucywyman:
```suggestion
RUN /opt/puppetlabs/bin/ssl-setup.sh && chown -R puppetdb:puppetdb /etc/puppetlabs/puppetdb/ssl/
```

spec/bolt/puppetdb/client_spec.rb
```diff
@@ -60,24 +61,65 @@
...
+      conf = pdb_conf
+      conf['server_urls'] = ['https://bad_host1.dne.com', 'https://bad_host2.dne.com']
+      client = Bolt::PuppetDB::Client.new(Bolt::PuppetDB::Config.new(conf))
+      msg = "Failed to connect to all PuppetDB server_urls: https://bad_host1.dne.com, https://bad_host2.dne.com."
+      expect { client.facts_for_node(%w[node1 node2]) }.to raise_error(Bolt::PuppetDBError, msg)
     end
 
-    it 'returns an empty list if no certnames are given' do
-      expect(client.facts_for_node([])).to eq({})
+    it 'should failover if the first server fails' do
```
lucywyman: It may also be worth testing that errors are caught and raise `PuppetDBFailoverError`

spec/Dockerfile.puppetdb
adreyer: It's set so that we can get a certificate from puppetserver. The whole reason we have our own dockerfile is because we don't want to get a certificate from puppetserver. I'll delete the line.

spec/bolt/puppetdb/client_spec.rb
adreyer: that feels like testing implementation not behavior to me.

.travis.yml
```diff
@@ -15,6 +15,7 @@ before_script:
 - eval `ssh-agent`
 - cat Gemfile.lock
 - bundle exec r10k puppetfile install
+- sudo apt-get install nmap
```
donoghuc: Don't forget to remove this.

Rakefile
```diff
@@ -14,22 +14,24 @@ RSpec::Core::RakeTask.new(:spec)
 
 desc "Run RSpec tests that don't require VM fixtures or a particular shell"
 RSpec::Core::RakeTask.new(:unit) do |t|
-  t.rspec_opts = '--tag ~ssh --tag ~docker --tag ~bash --tag ~winrm --tag ~appveyor_agents --tag ~puppetserver'
+  t.rspec_opts = '--tag ~ssh --tag ~docker --tag ~bash --tag ~winrm ' \
```
donoghuc: Another option is to replace all `--tag` with just `-t`. Though the line break is fine also if `--tag` is more readable.

Rakefile
```diff
@@ -14,22 +14,24 @@ RSpec::Core::RakeTask.new(:spec)
...
+         '--tag ~orchestrator --tag ~puppetserver --tag ~puppetdb'
 end
 
 desc "Run RSpec tests for TravisCI that don't require WinRM"
 RSpec::Core::RakeTask.new(:travisci) do |t|
-  t.rspec_opts = '--tag ~winrm --tag ~appveyor_agents --tag ~puppetserver --tag ~windows'
+  t.rspec_opts = '--tag ~winrm --tag ~appveyor_agents --tag ~puppetserver --tag ~puppetdb --tag ~windows'
 end
 
-desc "Run RSpec tests that require pupperserver"
```
donoghuc: pupperserver -> puppetserver

lib/bolt/puppetdb/client.rb
```diff
@@ -68,6 +85,23 @@ def http_client
         @http
       end
 
+      def reject_url
+        @bad_urls << @current_url if @current_url
+        @current_url = nil
+      end
+
+      def uri
+        @current_url ||= (Set.new(@config.server_urls) - @bad_urls).first
```
nicklewis: Sets aren't guaranteed to preserve insertion order. But you can also just subtract arrays, so `@config.server_urls - @bad_urls.to_a` would work.

nicklewis: In practice, it probably does preserve insertion order because it's backed by a Hash under the hood, which preserves insertion order. But the specified behavior doesn't make that guarantee, so I would prefer not to rely on it (implementations can change). Order is important, because generally `server_urls` represents primary and secondary.

Rakefile
adreyer: thats the old version

PR #822, opened 2019-01-04
(BOLT-1010) Convert Addressable::URI errors to Bolt errors

donoghuc:
Previously exceptions raised by the `Addressable::URI.parse` method would be surfaced. This commit converts `InvalidURIError`s in to `Bolt::ParseError`s.


donoghuc:
Now that I look at it there are kind of a lot of `:` in the message. 
```
cas@cas-ThinkPad-T460p:~/working_dir/bolt$ bolt command run 'echo test' -n invalid:uri
Error: Could not parse target URI: Invalid port number: "uri"
```

PR #823, opened 2019-01-04
(docs) Add missing parens to ctrl::do_until function example

donoghuc:
Previously the example used to generate function docs was missing parens.


donoghuc:
@melamos A slack user pointed out we forgot parenthesis in the `ctrl::do_until` std lib function. 

The markdown should now look like

## ctrl::do_until

Repeat the block until it returns a truthy value. Returns the value.


```
ctrl::do_until(Callable &$block)
```

*Returns:* `Any` 

* **&block** `Callable` 

**Example:** Run a task until it succeeds
```
ctrl::do_until() || {
  run_task('test', $target, _catch_errors => true).ok?
}
```

melamos:
Fixed in docs. Should appear in the 2p auto-build.

PR #828, opened 2019-01-07
(maint) Test bolt compatibility with ruby 2.6

donoghuc:
The latest version of the `ffi` gem adds support for ffi on windows.


donoghuc:
This is just a test do not merge.

PR #830, opened 2019-01-08
(maint) Update docs around `run_task` function

lucywyman:
**What this changes** This clarifies that `run_task` accepts a hash of task parameters, rather than configuration options. It also corrects that it runs a task, not a plan.
Additionally it adds a note about what changes need to be made for BOLT-952

**Why** The docs can be misread to accept config options.
Some of the work for BOLT-952 has been done, but it it's half-complete, so this clarifies what needs to change

### DIFF
--- a/pre-docs/plan_functions.md
+++ b/pre-docs/plan_functions.md
@@ -460,7 +460,7 @@ run_task(String[1] $task_name, Boltlib::TargetSpec $targets, Optional[Hash[Strin
 
 * **task_name** `String[1]` The task to run.
 * **targets** `Boltlib::TargetSpec` A pattern identifying zero or more targets. See [`get_targets`](#get_targets) for accepted patterns.
-* **task_args** `Optional[Hash[String[1], Any]]` Arguments to the plan. Can also include additional options: '_catch_errors', '_run_as'.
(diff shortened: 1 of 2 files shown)


pre-docs/plan_functions.md
```diff
@@ -460,7 +460,7 @@ run_task(String[1] $task_name, Boltlib::TargetSpec $targets, Optional[Hash[Strin
 
 * **task_name** `String[1]` The task to run.
 * **targets** `Boltlib::TargetSpec` A pattern identifying zero or more targets. See [`get_targets`](#get_targets) for accepted patterns.
-* **task_args** `Optional[Hash[String[1], Any]]` Arguments to the plan. Can also include additional options: '_catch_errors', '_run_as'.
+* **task_args** `Optional[Hash[String[1], Any]]` Arguments to the task. Can also include additional options: '_catch_errors', '_run_as'. Note: This is parameters for the task, not [configuration options](#bolt_configuration_options)
```
adreyer: I'm not sure this would have prevented the issue reported today. I think it may be more confusing to new users.

adreyer:
This is also an autogenerated file. I'm going to close this since there hasn't been any more discussion and it's not mergeable anyway.

PR #831, opened 2019-01-08
(maint) Only wrap tmpdir-related errors in FileError

MikaelSmith:
Previously other errors from invoking a task or script would also be
caught and wrapped in FileError. We should only do that for errors we
know are related to creating the tmpdir.

### DIFF
--- a/lib/bolt/transport/local.rb
+++ b/lib/bolt/transport/local.rb
@@ -36,11 +36,15 @@ module Bolt
 
       def in_tmpdir(base)
         args = base ? [nil, base] : []
-        Dir.mktmpdir(*args) do |dir|
-          yield dir
-        end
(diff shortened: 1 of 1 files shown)


MikaelSmith:
Should probably be a ticket even though it's very minor.

lib/bolt/transport/local.rb
```diff
@@ -36,11 +36,15 @@ def initialize
 
       def in_tmpdir(base)
         args = base ? [nil, base] : []
-        Dir.mktmpdir(*args) do |dir|
-          yield dir
-        end
-      rescue StandardError => e
```
MikaelSmith: I liked this syntax, but not sure what set of exceptions mktmpdir can throw.

PR #834, opened 2019-01-09
(BOLT-1009) Add plan_finish to plan executor app

lucywyman:
**What this changes** This adds a `finish_plan` call to the plan executor app once the plan finishes executing, which sends `command/plan_finish` post request to the orchestrator to mark the plan as finished. This also includes passing the plan job id to the orchestrator connection class `Bolt::Transport::Api::Connection` so that the requests to orchestrator that it constructs are in the correct plan context.

**Why** So that the plan is marked as 'finished' by the plan executor once plan execution finishes


lib/bolt/transport/api.rb
```diff
@@ -11,6 +11,12 @@ def initialize(*args)
         super
       end
 
+      def conn_plan_job(plan_id)
```
lucywyman: Another question: Do these connections get reused for multiple plan runs? Should the plan id be passed in to each `build_request` and `finish_plan` call so it doesn't get overwritten?

adreyer: yes the executor is shared when puma is threaded. You'll need to create a separate executor for each request.

lib/plan_executor/executor.rb
```diff
@@ -160,19 +160,8 @@ def wait_until(timeout, retry_interval)
...
-    # Since we callt this function without resolving targets this will result
-    # in the orchestrator transport always being initialized during plan runs.
-    # For now that's ok.
-    #
-    # In the future if other transports need this or if we want a plan stack
-    # we'll need to refactor.
-    def start_plan(plan_context)
-      @transport.plan_context = plan_context
-      @plan_logging = true
+    def conn_plan_job(plan_id)
```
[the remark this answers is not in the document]
lucywyman: Rubocop errored on 'set_plan_job', and since this isn't setting a var `plan_job=` also isn't what gets called with `executor.plan_job`. This was my compromise. I'll take another stab at using either of the other 2, or just turning off rubocops

lib/bolt/transport/api.rb
```diff
@@ -11,6 +11,12 @@ def initialize(*args)
...
+        @connections.each_value do |conn|
```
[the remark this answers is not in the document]
lucywyman: This is from the orch transport - does it make sense to rename to `@connection`, and for both orch and api or just the api transport?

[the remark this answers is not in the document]
adreyer: Just here. This doesn't accept connection info from the target so there will never be multiple orchestrators.

lib/bolt/transport/api.rb
```diff
@@ -7,21 +7,27 @@
...
+        @connection ||= Bolt::Transport::Api::Connection.new(conn_opts, logger, @plan_job)
       end
 
       def batches(targets)
-        targets.group_by { |target| Bolt::Transport::Api::Connection.get_key(target.options) }.values
+        targets
+      end
+
+      def finish_plan(result)
+        # CODEREVIEW should we check that result is a Bolt::PlanResult?
```
adreyer: That does not seem necessary since it should be coming from our code.

TODO: Verify that the plan result is generated by whatever call this.

lib/bolt/transport/api.rb
```diff
@@ -7,21 +7,27 @@
...
+        @connection.finish_plan(result)
+      rescue StandardError => e
+        @logger.debug("Failed to finish plan on #{@connection.key}: #{e.message}")
```
adreyer: This seems like a serious error in the case of a service since its means data is probably lost.

lib/bolt/transport/api/connection.rb
```diff
@@ -41,6 +41,9 @@ def finish_plan(plan_result)
               status: plan_result.status
             )
           end
+          # This will raise an error if for any reason another job is tried
```
adreyer: How would this happen? There is no attr_writer?

lib/plan_executor/app.rb
```diff
@@ -78,15 +69,24 @@ def validate_schema(schema, body)
       error = validate_schema(@schema, body)
       return [400, error.to_json] unless error.nil?
 
+      # Create a basic executor, leave concurrency up to Orchestrator.
+      # This is just a class variable for testing
+      executor = PlanExecutor::Executor.new(0, body['job_id'])
+      # Use an empty inventory until we figure out where this data comes from.
+      inventory = Bolt::Inventory.new(nil)
+      pal = Bolt::PAL.new(@modulepath, nil)
```
adreyer: Why did pal move here? Is it not threadsafe? I don't see any obvious mutable state in the PAL class.

adreyer:
I suspect there are some larger threadsafety issues with how we use PAL around https://github.com/puppetlabs/bolt/blob/master/lib/bolt/pal.rb#L166.

I think it's ok to consider those out of scope for this ticket at this point.

.rubocop.yml
```diff
@@ -81,3 +81,6 @@ Metrics/ParameterLists:
 
 Metrics/PerceivedComplexity:
   Enabled: false
+
+Naming/MemoizedInstanceVariableName:
```
adreyer: Do we still need this?

lib/bolt/transport/api.rb
adreyer: Logging it at error with the entire `result` seems better.

.rubocop.yml
lucywyman: Yes, I get 
```
lib/bolt/transport/api.rb:19:9: C: Naming/MemoizedInstanceVariableName: Memoized variable @connection does not match method name get_connection. Use @get_connection instead.
        @connection ||= Bolt::Transport::Api::Connection.new(conn_opts, logger, @plan_job)
```
if it's not. I can turn it off locally if it's a cop we want to catch other places. `get_connection` is called from `run_task_job` in the orchestrator transport, and it doesn't seem worth overloading the whole `run_task_job` in the api transport just to give `get_connection` a different name. I also tried renaming `@connection` to `@conn`, but the same error was raised.

adreyer: In general I don't think it's a super important cop. But this method has some smell to it which I think this cop was reasonably picking up on. The method accepts arguments but returns the memoized value which means it probably shouldn't be accepting arguments. IMO we need a larger refactor of this transport/connection probably ditching the "transport" all together and folding this logic directly into the executor/connection.  I think that refactor should probably remain out of scope for this ticket at this point.

If it's easy to fix I would change the function so that it's just a memoized reader for `connection`.

lib/bolt/transport/api.rb
adreyer: the error message? I think the error message should be updated to include both the error and the entire json format plan result.

lib/plan_executor/app.rb
adreyer: It makes this comment a lie. https://github.com/puppetlabs/bolt/blob/master/lib/bolt/pal.rb#L40

There are far bigger problems with how we're using PAL at this point though since the setting method is not threadsafe regardless of how many instances there are.

lib/bolt/transport/api.rb
```diff
@@ -7,21 +7,26 @@
...
+        @connection ||= Bolt::Transport::Api::Connection.new(conn_opts, logger, @plan_job)
       end
 
       def batches(targets)
-        targets.group_by { |target| Bolt::Transport::Api::Connection.get_key(target.options) }.values
+        targets
+      end
+
+      def finish_plan(result)
+        @connection.finish_plan(result)
```
nicklewis: This fails if the plan doesn't run any tasks, since `run_task_job` is the spot where `get_connection` is originally called. Otherwise, `@connection` is nil.

lib/plan_executor/app.rb
```diff
@@ -78,15 +69,24 @@ def validate_schema(schema, body)
       error = validate_schema(@schema, body)
       return [400, error.to_json] unless error.nil?
 
+      # Create a basic executor, leave concurrency up to Orchestrator.
+      # This is just a class variable for testing
+      executor = PlanExecutor::Executor.new(0, body['job_id'])
```
nicklewis: The first argument to `PlanExecutor::Executor.new` appears to be `noop`, so `0` isn't an appropriate value for that. Also, this is a local variable, not a class variable.

nicklewis:
It's unclear how the plan executor service ever gets any config to talk to orchestrator in this PR.

lucywyman:
@nicklewis @adreyer  Once appveyor and travis are passing I think this is ready for review. I plan to look over the tests tomorrow and at least beef up the unit testing a bit, as well as look at adding end-to-end testing.

lib/bolt/transport/api/connection.rb
```diff
@@ -1,13 +1,14 @@
...
-# This is a copy of the orchestrator transport connection, but without the
-# 'start_plan' call in the init function which is handled by the orchestrator
 # client
 module Bolt
   module Transport
-    class Api < Orch
+    # Remove transport but keep Connection class
+    # This is mostly for organization + testing
+    # Useful to capture all the connection info to orch, and how you build request and handle response
+    class Api
```
nicklewis: Is this class used now?

lib/bolt/transport/api.rb
```diff
@@ -2,26 +2,213 @@
...
 require 'bolt/transport/api/connection'
 
-# A copy of the orchestrator transport which uses the api connection class
-# in order to bypass calling 'start_plan'
 module Bolt
   module Transport
-    class Api < Orch
-      def initialize(*args)
-        super
+    class Api
```
nicklewis: I think we should move this out of the `Bolt::Transport` namespace, since it's not a transport in the typical sense. Maybe something under `plan_executor/executor`, since it's really just a part of the executor.

lib/bolt/transport/api.rb
```diff
@@ -2,26 +2,213 @@
...
+          end
+        end
+      end
+
+      # Unwraps any Sensitive data in an arguments Hash, so the plain-text is passed
+      # to the Task/Script.
+      #
+      # This works on deeply nested data structures composed of Hashes, Arrays, and
+      # and plain-old data types (int, string, etc).
+      def unwrap_sensitive_args(arguments)
```
nicklewis: It would be good to move this to a location where it can be shared. Maybe a class method on `Bolt::PAL`, since it does explicitly Puppet-y things.

lib/plan_executor/app.rb
```diff
@@ -27,19 +28,25 @@ def puppetdb_client
       end
     end
 
-    def initialize(modulepath, executor = nil)
-      @schema = JSON.parse(File.read(File.join(__dir__, 'schemas', 'run_plan.json')))
-      @worker = Concurrent::SingleThreadExecutor.new
+    def initialize(modulepath, conn_opts)
+      conn_opts['User-Agent'] = "Bolt/#{Bolt::VERSION}"
+      @client = OrchestratorClient.new(conn_opts, true)
```
nicklewis: This should be `false`, since we definitely _don't_ want to load the default config files.

lib/plan_executor/executor.rb
```diff
@@ -7,26 +7,29 @@
...
+      @transport = nil
+    end
+
+    def make_transport(plan_job, client)
+      @transport = Bolt::Transport::Api.new(plan_job, client, @logger)
     end
 
     # This handles running the job, catching errors, and turning the result
     # into a result set
+    # Rename to something more useful
```
nicklewis: Maybe something like `as_resultset`.

nicklewis:
With a couple of changes (`plan_job` to `job_id`, setting environment) I was able to use my `puppet-plan` tool prototype to run a plan that runs a task and get back the correct result!

lib/plan_executor/app.rb
```diff
@@ -77,16 +84,17 @@ def validate_schema(schema, body)
       body = JSON.parse(request.body.read)
       error = validate_schema(@schema, body)
       return [400, error.to_json] unless error.nil?
-
       name = body['plan_name']
       # Errors if plan is not found
       @pal.get_plan_info(name)
 
+      @executor.make_transport(body['plan_job'], @client)
```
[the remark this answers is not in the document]
lucywyman: Sounds good - I really debated which pattern to go with and ultimately figured someone would comment if there was a 'better path' :) Thanks!

lib/bolt/transport/api.rb
lucywyman: `PlanExecutor::OrchClient`?

lucywyman: I guess to me it seems like, even though it's only used by the executor it's not logically part of the executor - it's really the interface between the executor and orchestrator. I don't feel strongly about it though if you'd rather I make it a class under  the executor, `PlanExecutor::Executor::OrchClient` or somesuch.

lib/bolt/transport/api.rb
lucywyman: I'd kind of like to ticket this and do it later in the interest of time, but if you're ok with waiting until Monday or doing it yourself I'm not opposed to having it be part of this PR

nicklewis: Doing it later is fine.

lucywyman: BOLT-1102 :+1:

nicklewis:
We need to teach `orchestrator_client-ruby` to use certs (and to not automatically load a token), but I think that can be done separately from this PR. For now, you need to write a token to `/opt/puppetlabs/server/data/bolt-server/.puppetlabs/token`.

lucywyman:
@nicklewis I assume that means I also need to generate a token - I'm not sure what the best way to do that here is?

nicklewis:
@lucywyman For now I think dealing with generating and configuring a token can be left as a manual exercise when trying to set up a testing environment. We should prioritize extending [orchestrator_client-ruby](https://github.com/puppetlabs/orchestrator_client-ruby) with cert support.

nicklewis:
I think the missing require is the last thing needed to merge this.

PR #835, opened 2019-01-09
Update directory_structure.md

melamos:


melamos:
Just updated to match what's in docs proper.

PR #838, opened 2019-01-09
(BOLT-756) Only try to load puppetdb.conf if exists

donoghuc:
Previously an error message would occur if no puppetdb.conf was found.


donoghuc:
This avoids the `"Could not load puppetdb.conf from #{filepath}: #{e.message}"` message if any config files are missing.

PR #840, opened 2019-01-10
(docs) Track output of rake docs task as plan_functions.md

donoghuc:
Previously the `docs` rake task would output to `REFERENCE.md` and the output would be added to `pre-docs/plan_functions.md`. This commit changes the `docs` rake task to update the `plan_functions.md` file directly.


donoghuc:
Based on conversation with @melamos

melamos:
Strong +1 to limiting the source of truth about functions to a single file. Plus I won't have to solve my rake docs issue with this solution!

PR #841, opened 2019-01-14
(BOLT-1022) Allows hyphens in aliases and group names

nicklewis:
This allows valid hostnames like prod-master-1 to be used as aliases,
without having to translate the hyphens to underscores. This does not
allow an alias to _begin_ with a hyphen.


donoghuc:
Started to review by investigating Travis failure. See https://github.com/puppetlabs/bolt/pull/842

PR #842, opened 2019-01-15
(BOLT-938) Handle connect timeout error for puppetdb client failover

donoghuc:
This commit adds `HTTPClient::ConnectTimeoutError` to the list of handled exceptions for rolling with a puppetdb failover. Previously this would raise a PuppetDBerror and not result in checking for other acceptable puppetdb connections.

### DIFF
--- a/lib/bolt/puppetdb/client.rb
+++ b/lib/bolt/puppetdb/client.rb
@@ -50,10 +50,8 @@ module Bolt
 
         begin
           response = http_client.post(url, body: body, header: headers)
-        rescue SocketError, OpenSSL::SSL::SSLError, SystemCallError, Net::ProtocolError, IOError => err
-          raise Bolt::PuppetDBFailoverError, "Failed to query PuppetDB: #{err}"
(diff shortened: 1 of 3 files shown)


lib/bolt/puppetdb/client.rb
```diff
@@ -50,7 +50,8 @@ def make_query(query, path = nil)
 
         begin
           response = http_client.post(url, body: body, header: headers)
-        rescue SocketError, OpenSSL::SSL::SSLError, SystemCallError, Net::ProtocolError, IOError => err
+        rescue SocketError, OpenSSL::SSL::SSLError, SystemCallError,
```
adreyer: maybe we should just rescue StandardError for this? this is the kind of bug I was concerned about when I added this

I'm ok merging if someone thinks thats a bad idea.

donoghuc: Updated to raise ` Bolt::PuppetDBFailoverError` for *any* standard error with http response as discussed.

PR #843, opened 2019-01-16
(FM-7656) Update puppet_agent module ref

donoghuc:
The powershell implementation for the puppet_agent::install task now downloads the msi over https instead of http.


MikaelSmith:
This should probably have a ticket associated with it.

PR #844, opened 2019-01-17
Add parameter limit to function do_until() and update documentation; …

automaticserver:
…fix minor documentation error with or() function

### DIFF
--- a/bolt-modules/ctrl/lib/puppet/functions/ctrl/do_until.rb
+++ b/bolt-modules/ctrl/lib/puppet/functions/ctrl/do_until.rb
@@ -3,7 +3,7 @@
 # Repeat the block until it returns a truthy value. Returns the value.
 Puppet::Functions.create_function(:'ctrl::do_until') do
   # @example Run a task until it succeeds
-  #   ctrl::do_until() || {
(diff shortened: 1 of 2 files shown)


MikaelSmith:
This is an interesting feature. The one part I think we might question is whether we want more flexibility to expand later, and want to make it a hash of named parameters (with only one useful option at the moment). We'd also appreciate if you opened a BOLT feature ticket at tickets.puppet.com.

automaticserver:
I think the until() method for implementation inside ruby is a bad decision: it's lacking of a mechanism to return the results. It requires to return a boolean type. Because of puppet's scope concept, there is no second channel to return a value to a wider scope - which is not a bad design decision.

Because bolt requires to have access to results (that is a big value), there should be another implementation at all.

At the moment in case of a timeout, there wouldn't be a result available. Even in case of success you don't know how often something was done. On a later stage you need maybe the results instead just "true".

adreyer:
@automaticserver I think we can merge this in with the following changes:
1. Move `limit` to be an option rather than a positional argument.
2. Add a test.

do you want to clean this up or should we?

adreyer:
merged in 1331

PR #848, opened 2019-01-23
(maint) Pin to compatible puppetdb docker image

donoghuc:
The puppetdb docker image has had a major update that changes the way we use in for testin bolt. This commit pins to a compatible version while we sort out how to update tests to be compatible with new image.


donoghuc:
I would like to discuss how to properly utilize the new image with Nate Wolf but he is on PTO today.

PR #850, opened 2019-01-24
(maint) Update puppetdb container setup

donoghuc:
The latest and greatest puppetdb container requires some changes to how it is set up as it uses a stripped down alpine base image and builds puppetdb from source instead of a package. This commit changes the entrypoint to run only the ssl-setup script and then invoke the closure service.


spec/fixtures/puppetdb/ssl-setup.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/sh
```
nwolfe: Is there a reason to have your own copy of this file?

I think you could just expect this to be in the upstream puppetdb image at `/ssl-setup.sh`, and call that.

spec/Dockerfile.puppetdb
```diff
@@ -8,11 +8,12 @@ COPY fixtures/ssl/cert.pem /etc/puppetlabs/puppet/ssl/certs/pdb.pem
...
 
-RUN /opt/puppetlabs/bin/ssl-setup.sh && chown -R puppetdb:puppetdb /etc/puppetlabs/puppetdb/ssl/
-
-# The normal bootstrap should copy the cert in place
-# Skip the normal bootstrapping and just run puppet-server
- ENTRYPOINT ["/opt/puppetlabs/bin/puppetdb"]
- CMD ["foreground"]
+# Use our own entrypoint
+ENTRYPOINT ["/sbin/tini", "-g", "--", "/entrypoint.sh"]
+CMD ["services"]
```
nwolfe: I think you could just overwrite the `/docker-entrypoint.sh` file itself, and not have to redefine these `ENTRYPOINT` and `CWD` lines.

spec/fixtures/puppetdb/ssl-setup.sh
donoghuc: I think the only change was we hardcoded the value of `mycertname` to be `pdb` instead of using the `HOSTNAME` environment variable. I am testing out setting `HOSTNAME=pdb` in our dockerfile to see if we can avoid maintaining our own copy of ssl-setup.sh.

donoghuc: Just setting the environment variable appears to work. This will allow us to utilize the script that ships with the image.

nwolfe: Ah, didn't notice that difference.

A few thoughts come to mind:

* The hardcoded ssl fixture files are expected to be named after the hostname, since things will try and use `$HOSTNAME` or `puppet config print certname` to find them. If you have fixtures named `certs/puppetdb.pem` then ssl-setup.sh would not find them if it was using HOSTNAME=pdb since it'll look for `certs/pdb.pem`
* If you do overwrite HOSTNAME, limit the scope as much as possible by putting it in front of the command that needs it, like `HOSTNAME=pdb /ssl-setup.sh`, versus at a global/higher scope
* We can just change the real ssl-setup.sh file in puppetdb so this hostname can be passed in

donoghuc:
waiting for upstream enhancement to merge.

spec/fixtures/puppetdb/ssl-setup.sh
```diff
@@ -1,7 +1,10 @@
...
 
 ssl_command="puppetdb ssl-setup"
 
+# If CERTNAME environment variable is set use that, otherwise use HOSTNAME
+mycertname="${CERTNAME:-$HOSTNAME}"
```
donoghuc: This is the upstream change that we intend to make. For now it is just here for testing. Should be removed once available.

donoghuc:
Upstream change made it to `latest` tag of `puppet/puppetdb` on docker hub. This should be ready.

PR #852, opened 2019-01-25
(BOLT-836) Support lookup() in plans outside of apply blocks

MikaelSmith:
Makes simple lookups work from Bolt's hiera config. The inputs to lookup - certname, environment, facts, etc - may not be correct.

### DIFF
--- a/lib/bolt/pal.rb
+++ b/lib/bolt/pal.rb
@@ -172,6 +172,7 @@ module Bolt
         Puppet.settings.send(:clear_everything_for_tests)
         Puppet.initialize_settings(cli)
         Puppet::GettextConfig.create_default_text_domain
+        Puppet.settings[:hiera_config] = @hiera_config
         self.class.configure_logging
         yield
       end


MikaelSmith:
This should probably also have a lookup plan that reports what's accessible via hiera and how. Many things won't be, and we need something *like* `puppet lookup` to describe what's happening.

nmaludy:
+1

adreyer:
I'm fairly certain we don't want the same hierarchy at the plan level as we do for applies.

adreyer:
Is the desire to reuse lookup plugins from the plan level? I think we need to find a way to support them that is less "scope->hierarchy" focused.

nmaludy:
@adreyer our desire is to use hiera as a config file to populate variables and common parameters from existing hiera data instead of having to copy this data into inventory files. Also, we would like to use hiera-eyaml to encrypt/decrypt keys that we need during plan execution to pass in to tasks, commands and apply blocks, etc

adreyer:
There appears to the a bug where `$trusted` is set with nil values during plan execution. This is the only thing preventing hiera from blowing up with basically any hiera.yaml during plan execution. Other hierarchies are likely to blow up and fixing this will likely break anyone who starts using this. 

I think we need a separate hiera.yaml for the plan level at a minimum. I would consider adding plan_hiera.yaml  and fixing the $trusted bugs, then releasing this as experimental.

@nmaludy are you using `set_var` on nodes or just setting `$variablename` in the plan?

adreyer:
This discussion moved forward in slack and additional work is needed so I'm closing this PR.

PR #853, opened 2019-01-28
(BOLT-1009) Plan executor container

lucywyman:
This is branched off of #834 

**What this changes** This adds the docker infrastructure and documentation for running the plan-executor service locally and testing it with `curl` commands.
**Why** This provides a way to manually test the plan executor locally and verify that interactions that don't include the orchestrator work as expected. Note: this isn't connected to any orchestrator - see integration testing in the orchestrator repo for end-to-end plan executor testing.


lucywyman:
@nicklewis  Do we want to consider these unnecessary and close? We can always reopen, but I don't see them moving forward in the next week.

PR #854, opened 2019-01-28
(BOLT-830) Add script extension metadata for bolt_shim script task

donoghuc:
The bolt_shim::script task now expects the script extension to be provided when the target is has a windows OS. See https://github.com/puppetlabs/puppetlabs-bolt_shim/pull/8 for corresponding bolt_shim module PR.


donoghuc:
Is it worth adding the `extension` configuration option to pcp transport similar to winrm? This would be ignored for linux targets and only provide value to users who want to run non powershell, ruby, etc scripts on windows agents.

PR #855, opened 2019-01-29
(BOLT-1027) Print modulepath when showing tasks/plans

donoghuc:
This commit shows the modulepath (excluding the internal bolt modulepaths) when `bolt {task,plan} show` is invoked in human readable form.


donoghuc:
Should I modify this to include the modulepath with JSON output?

donoghuc:
Looks like there has only been one instance we have captured of json output whith plan/task show.
image

MikaelSmith:
I don't see a reason to leave it out of JSON output.

PR #856, opened 2019-01-30
(BOLT-153) Support SMB file transfer

joshcooper:
WinRM sucks for file transfer because it base64 encodes the payload,
compresses it as a zip file, transfers the contents via Powershell in
SOAP/XML messages, and decompresses on the remote side.

This commit allows bolt to upload files using SMB using the Rapid7 gem.
If anyone understands SMB, it's Rapid7/metasploit. The WinRM connection
lazily generates an SMBClient if a file is uploaded. The connection
will only login once, but connect to and disconnect from the share for
each uploaded file.

Before this commit transferring a 75MB jre takes 58.7 seconds to a local
Windows VM. With this commit it takes 6.2 seconds.

By default, WinRM connections continue to use WinRM for file transfers.
To enable SMB, add the following to the inventory file:

    winrm:
      file-protocol: smb

When uploading a file, the destination path should be specified as
either a drive letter or UNC path. If a drive letter is given, then bolt
will use the administrative share, usually `\\<host>\C$`. Note this
requires administrative privileges. To upload a file as a non-admin
user, enter the share as a UNC path. If bolt is running on a Windows
host:

    PS C:\> bolt file upload C:\path\to\file \\host\share\path\to\filebolt file upload C:\path\to\file \\host\share\path\to\file

If bolt is running on a non-Windows host, backslashes in the destination
path need to be escaped:

    $ bolt file upload /path/to/file \\\\host\\share\\path\\to\\file

Or forward slashes can be used (for drive letter or UNC paths):

    $ bolt file upload /path/to/file C:/path/to/file

### DIFF
--- a/bolt.gemspec
+++ b/bolt.gemspec
@@ -44,6 +44,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency "puppet", [">= 6.0.1", "< 7"]
   spec.add_dependency "puppet-resource_api"
   spec.add_dependency "r10k", "~> 2.6"
+  spec.add_dependency "ruby_smb", "~> 1.0"
   spec.add_dependency "terminal-table", "~> 1.8"
   spec.add_dependency "winrm", "~> 2.0"
   spec.add_dependency "winrm-fs", "~> 1.3"
--- a/lib/bolt/transport/winrm/connection.rb
+++ b/lib/bolt/transport/winrm/connection.rb
@@ -2,6 +2,7 @@
 
 require 'bolt/node/errors'
 require 'bolt/node/output'
+require 'ruby_smb'
 
 module Bolt
   module Transport
@@ -92,6 +93,7 @@ module Bolt
 
         def disconnect
           @session&.close
+          @client&.disconnect!
           @logger.debug { "Closed session" }
         end
 
@@ -140,8 +142,55 @@ module Bolt
         end
 
         def write_remote_file(source, destination)
+          if target.options['file-protocol'] == 'smb'
+            write_remote_file_smb(source, destination)
+          else
+            write_remote_file_winrm(source, destination)
+          end
+        end
+
+        def write_remote_file_winrm(source, destination)
           fs = ::WinRM::FS::FileManager.new(@connection)
           fs.upload(source, destination)
+        rescue StandardError => e
+          raise Bolt::Node::FileError.new(e.message, 'WRITE_ERROR')
+        end
+
+        def write_remote_file_smb(source, destination)
+          win_dest = destination.tr('/', '\\')
+          if (md = win_dest.match(/^([a-z]):\\(.*)/i))
+            # if drive, use admin share for that drive, so path is '\\host\C$'
+            path = "\\\\#{@target.host}\\#{md[1]}$"
+            dest = md[2]
+          elsif (md = win_dest.match(/^(\\\\[^\\]+\\[^\\]+)\\(.*)/))
+            # if unc, path is '\\host\share'
+            path = md[1]
+            dest = md[2]
+          else
+            raise ArgumentError, "Unknown destination '#{destination}'"
+          end
+
+          client = smb_client_login
+          tree = client.tree_connect(path)
+          begin
+            file = tree.open_file(filename: dest, write: true, disposition: ::RubySMB::Dispositions::FILE_OVERWRITE_IF)
+            begin
+              # `file` doesn't derive from IO, so can't use IO.copy_stream
+              File.open(source, 'rb') do |f|
+                pos = 0
+                while (buf = f.read(8 * 1024 * 1024))
+                  file.write(data: buf, offset: pos)
+                  pos += buf.length
+                end
+              end
+            ensure
+              file.close
+            end
+          ensure
+            tree.disconnect!
+          end
+        rescue ::RubySMB::Error::UnexpectedStatusCode => e
+          raise Bolt::Node::FileError.new("SMB Error: #{e.message}", 'WRITE_ERROR')
         rescue StandardError => e
           raise Bolt::Node::FileError.new(e.message, 'WRITE_ERROR')
         end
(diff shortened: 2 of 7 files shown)


joshcooper:
I got this mostly working over the holiday break, and added a last bit last night. Tests are currently failing, but wanted to get 👀 on this before going further.

bolt.gemspec
```diff
@@ -44,6 +44,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency "puppet", [">= 6.0.1", "< 7"]
   spec.add_dependency "puppet-resource_api"
   spec.add_dependency "r10k", "~> 2.6"
+  spec.add_dependency "ruby_smb", "~> 1.0"
```
MikaelSmith: BSD 3 clause license: https://github.com/rapid7/ruby_smb/blob/master/LICENSE.txt. That's probably acceptable.

joshcooper: Filed as PE-25855

lib/bolt/transport/winrm/connection.rb
```diff
@@ -140,12 +142,59 @@ def mkdirs(dirs)
         end
 
         def write_remote_file(source, destination)
+          if target.options['file-protocol'] == 'smb'
+            write_remote_file_smb(source, destination)
+          else
+            write_remote_file_winrm(source, destination)
+          end
+        end
```
joshcooper: Looking for feedback about the following questions:

* What's the best way to select smb over winrm for file copy operations? Should it be a setting in the `winrm` section? Should it be part of the protocol, eg `winrm+smb://`? If so, can it be configured globally?
* It'd be nice to configure some aspects of SMB, such whether to disallow SMBv1, whether to require SMB signing, and in the future options for SMBv3. Should that be a subsection within `winrm`? Or should it be at the same level as `winrm`, `ssh`, sections?
* Does bolt support `sftp` and `scp` for file transfers?

adreyer: part of the protocol would make this a separate transport which I don't think is ideal. I think it's best for this to just be part of the `winrm` transport enabled with a config option. If there are `smb` specific options those should just be more `winrm` configuration options. If you expect there to be more than a few they can be nested inside a `smb` key

lib/bolt/transport/winrm/connection.rb
```diff
@@ -183,6 +232,51 @@ def write_executable_from_content(dir, content, filename)
...
+              'CONNECT_ERROR'
+            )
+          end
+
+          @client
+        end
+
+        def smb_socket_connect
+          # It's lame that TCPSocket doesn't take a connect timeout
+          # Using Timeout.timeout is bad, but is done elsewhere...
```
Iristyle: Yeah, it would be really nice to get rid of `Timeout.timeout` if we can

Iristyle: The correct solution for that is detailed in https://stackoverflow.com/a/21014439

lib/bolt/transport/winrm/connection.rb
```diff
@@ -140,12 +142,59 @@ def mkdirs(dirs)
...
+            dest = md[2]
+          elsif (md = win_dest.match(/^(\\\\[^\\]+\\[^\\]+)\\(.*)/))
+            # if unc, path is '\\host\share'
+            path = md[1]
+            dest = md[2]
+          else
+            raise ArgumentError, "Unknown destination '#{destination}'"
+          end
+
+          client = smb_client_login
```
Iristyle: So even without caching / reusing the existing connection, this still provides significant benefit.

Did you happen to look at the impact of SMB connection reuse? (No worries if not, I believe there will be an effort to support better connection handling for winrm in the future, so this could come along for the ride)

joshcooper: Minor terminology nit, the SMB connection is reused, but the tree is not. I believe the tree connect/disconnect is similar to concept to mounting a remote filesystem, and in this PR we do that for each file upload.

It'd be possible to reuse the same tree across multiple uploads, but bolt would need to keep track of which destination paths map to which trees, and create them as needed. It might be useful if there large numbers of small files being uploaded in a plan, but I didn't test it. Does bolt support uploading a directory?

Iristyle: Thanks for the clarification - I completely missed the `@client` bit

Iristyle:
FYI @joshcooper it looks like you filed this as BOLT-153 already if you want to update the commit.

Do you want to work out the tests or do you want to hand off to me to wrap it up?

lib/bolt/transport/winrm/connection.rb
```diff
@@ -183,6 +232,51 @@ def write_executable_from_content(dir, content, filename)
...
+        STATUS_SUCCESS = 0
+        STATUS_LOGON_FAILURE = 0xC000006D
+
+        SMB_PORT = 445
+
+        def smb_client_login
+          return @client if @client
+
+          dispatcher = RubySMB::Dispatcher::Socket.new(smb_socket_connect)
+          @client = RubySMB::Client.new(dispatcher, smb1: true, smb2: true, username: @user, password: target.password)
```
joshcooper: Drop smb1, there's no good reason to support as everything since Vista supports SMBv2 (or v3)

lib/bolt/transport/winrm/connection.rb
```diff
@@ -183,6 +232,51 @@ def write_executable_from_content(dir, content, filename)
...
+          status = @client.login
```
joshcooper: Verify signing has been required by the server (by default), but allow downgrade (maybe).

lib/bolt/transport/winrm/connection.rb
```diff
@@ -140,12 +142,59 @@ def mkdirs(dirs)
         end
 
         def write_remote_file(source, destination)
+          if target.options['file-protocol'] == 'smb'
```
joshcooper: The `ruby_smb` gem only supports SMBv1/2, but not v3 (yet). However, only SMBv3 supports encryption. What should the behavior be when using `winrm+ssl`? I assume we should error here?

lib/bolt/transport/winrm/connection.rb
Iristyle: Agree - no reason for ancient SMB1 support.

lib/bolt/transport/winrm/connection.rb
Iristyle: More background on SMB2 vs SMB3 - https://blogs.msdn.microsoft.com/openspecification/2017/05/26/smb-2-and-smb-3-security-in-windows-10-the-anatomy-of-signing-and-cryptographic-keys/

Iristyle: Agree that user asks for encryption with `+ssl` that file transfers should also be considered encrypted. So I think we have to fallback to using the winrm channel for file transfers (perhaps with a warning?) under such circumstances.

https://docs.microsoft.com/en-us/windows-server/storage/file-server/smb-security is also a bit easier to understand than the detail presented in the above links.

adreyer:
it looks like this was superseded.

PR #859, opened 2019-02-05
(BOLT-952) Add developer docs description for task schema

donoghuc:
Improve documentation and the task.json schema. Also delete an unmaintained bit of documentation.


developer-docs/bolt-api-servers.md
```diff
@@ -152,6 +152,7 @@ object](https://github.com/puppetlabs/puppetserver/blob/master/documentation/pup
 from [puppetserver](https://github.com/puppetlabs/puppetserver), with an
 additional `file_content` key.
 
+See the [schema](../lib/bolt_ext/schemas/task.json)
```
donoghuc: ugh. I copy pasted these from above. They all point to the old location... Will wait for any other comments before fixing up to save some CI load.

developer-docs/bolt-api-servers.md
```diff
@@ -165,9 +166,18 @@ The metadata object is optional, and contains metadata about the task being run.
 - `parameters`: Object, *optional* - A JSON object whose keys are parameter names, and whose values are JSON objects with 2 keys:
     - `description`: String, *optional* - The parameter description.
     - `type`: String, *optional* - The type the parameter should accept.
+    - `sensitive`: Boolean, *optional* - Whether the task runner should treat the parameter value as sensitive
+    - `input_method`: String, *optional* - What input method should be used to pass params to task (stdin, environment, powershell)
 
 #### Files
-# TODO
+The files array contains details about the files the task needs as well as how to get them. Array items should be objects with the following keys:
```
lucywyman: I think include that it's required? Should we just add an italicized `*required*` next to that and name? Or is it obvious from examples + other docs that it's required?

developer-docs/bolt-api-servers.md
```diff
@@ -165,9 +166,18 @@ The metadata object is optional, and contains metadata about the task being run.
...
+- `uri`: Object, *optional* - Information on how to request task files
```
lucywyman: I think this is required, looking at the schema? https://github.com/puppetlabs/bolt/blob/master/lib/bolt_server/schemas/task.json#L69

developer-docs/bolt-api-servers.md
```diff
@@ -165,9 +166,18 @@ The metadata object is optional, and contains metadata about the task being run.
...
+    - `path`: String, *optional* - Relative URI for requesting task content
+    - `params`: Object, *optional* - Query parameters for locating task data
+        - `environment`: String, *required* - Environment task files are in
```
lucywyman: Is this required for the `uri` object? If so I think `params` is also required, since it contains environment

lib/bolt_server/schemas/task.json
```diff
@@ -50,19 +50,35 @@
...
+            "properties": {
+              "path": {
+                "description": "Relative URI for accessing task contents",
+                "type": "string"
+              },
+              "params": {
+                "description": "Map of query params when requesting task content",
+                "type": "object",
+                "properties": {
+                  "environment": {
```
lucywyman: Does this need to be in a `required:` list somewhere/?

developer-docs/bolt-api-servers.md
```diff
@@ -165,9 +166,18 @@ The metadata object is optional, and contains metadata about the task being run.
...
 #### Files
-# TODO
+The files array is required, and contains details about the files the task needs as well as how to get them. Array items should be objects with the following keys:
+- `uri`: Object, *required* - Information on how to request task files
+    - `path`: String, *required* - Relative URI for requesting task content
+    - `params`: Object, *required* - Query parameters for locating task data
+        - `environment`: String, *required* - Environment task files are in
+- `sha256`: String, *required* - Shasum of the file contents
+- `filename`: String, *required* - File name including extension
+- `size`: Number, *optional* - Size of file in Bytes
```
lucywyman: I forget - do we actually do anything with this parameter?

lucywyman: Eh it isn't hurting anything - the bigger question is do we want to add a size check somewhere if this is present? I don't see where in bolt server or puppetserver that check happens, but maybe I'm missing it.

lucywyman:
The travis failures also look legit:
```
 expected `#<Rack::MockResponse:0x0000000002b84208 @original_headers={"Content-Type"=>"application/json", "Conte...erty '#/task/files/0/uri' did not contain a required property of 'params' in schema file:task\"]}"]>.ok?` to return true, got false
```

PR #860, opened 2019-02-06
(BOLT-1107) Use internal orchestrator API for running plan tasks

nicklewis:
Previously, the plan executor service used the public orchestrator API
to run plan tasks and submit plan results. This required it to follow
the same rules as a user running plans; specifically, it needed to
authenticated with an RBAC token and was only allowed to operate on
plans that were started by the same user for whom its RBAC token was
issued.

Now the service uses the internal API with certificate authentication to
connect as the API user and operate on plans with complete authority.
Because the public orchestrator-client library doesn't support features
we *need* like certificate authentication and the internal command API,
and it *does* have some automatic behavior around config loading that
we *don't* want, we now use a JSONClient (from the httpclient library)
directly to talk to the orchestrator service from the plan executor
instead.

We now create the http client during app startup time and pass it
through when creating an executor / orch client. That simplifies mocking
the http client in tests and also ensures that if we were to run
multiple plans at once, they would share a single http client.

### DIFF
--- a/lib/plan_executor/orch_client.rb
+++ b/lib/plan_executor/orch_client.rb
@@ -1,30 +1,31 @@
 # frozen_string_literal: true
 
+require 'jsonclient'
+
 module PlanExecutor
   class OrchClient
-    attr_reader :plan_job, :client
+    attr_reader :plan_job, :http
 
     BOLT_COMMAND_TASK = Struct.new(:name).new('bolt_shim::command').freeze
     BOLT_SCRIPT_TASK = Struct.new(:name).new('bolt_shim::script').freeze
     BOLT_UPLOAD_TASK = Struct.new(:name).new('bolt_shim::upload').freeze
 
-    def initialize(plan_job, client, logger)
+    def initialize(plan_job, http_client, logger)
       @plan_job = plan_job
-      @client = client
       @logger = logger
-      @client_url = client.config['service-url']
-      logger.debug("Creating orchestrator client for #{@client_url}")
+      @http = http_client
       @environment = 'production'
     end
 
     def finish_plan(plan_result)
-      result = @client.command.plan_finish(
+      body = {
         plan_job: @plan_job,
         result: plan_result.value || '',
         status: plan_result.status
-      )
+      }
+      post_command('internal/plan_finish', body)
     rescue StandardError => e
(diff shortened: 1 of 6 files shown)


lib/plan_executor/orch_client.rb
```diff
@@ -1,30 +1,36 @@
...
         plan_job: @plan_job,
         result: plan_result.value || '',
         status: plan_result.status
-      )
+      }
+      response = @http.post('internal/plan_finish', body)
+      if response.status == 202
+        response.body
+      else
+        raise Bolt::Error.new(response.body['msg'], response.body['kind'], response.body['details'])
```
branan: Raising an error just to rescue it two lines down feels kinda gross to me?

Since a very similar pattern is used below, does it make sense to wrap this in its own function just like was done for `get`, to encapsulate this error handling? That'd clean up (or at least hide ;) ) this weirdness with an immediately-logged exception, without the need to make the two uses of `@http.post` behave totally differently

That means lines 25-31 would be replaced with a simple `post('internal/plan_finish', body)`

lib/plan_executor/orch_client.rb
```diff
@@ -55,7 +69,19 @@ def send_request(targets, task, arguments, options = {})
       body[:description] = description if description
       body[:plan_job] = @plan_job if @plan_job
 
-      @client.run_task(body)
+      response = @http.post('internal/plan_task', body)
+      if response.status != 202
+        raise Bolt::Error.new(response.body['msg'], response.body['kind'], response.body['details'])
+      end
+      url = response.body.dig('job', 'id')
```
branan: And I guess adding a `post` function would make the usage here something like `url = post('internal/plan_task', body).dig('job', 'id')`

lib/plan_executor/orch_client.rb
nicklewis: I think that makes sense. I definitely don't like that it's catching its own exception, but it was either that or duplicate the `@logger.error` messages in two cases. Pulling the exception raising into a `post` method more accurately represents the desired behavior here, which is to log _any_ error, whether from an internal http client exception or from the response being an error.

nicklewis:
Updated to pull out a `post_command` function. I went with the more specific name because that explains the check for HTTP 202, and justifies _not_ using the same function for `connected?` which expects a 200.

nicklewis:
Whoops I broke it. Pushed a fix.

PR #861, opened 2019-02-07
(BOLT-1118) Add command to show available modules

Sharpie:
This commit adds a `bolt module show` command that lists out each module path
in the Bolt environment, followed by the modules available and their versions
in a manner similar to `puppet module list`.

### DIFF
--- a/bolt.gemspec
+++ b/bolt.gemspec
@@ -21,6 +21,7 @@ Gem::Specification.new do |spec|
                        Dir['libexec/*'] +
                        Dir['bolt-modules/*/lib/**/*.rb'] +
                        Dir['bolt-modules/*/types/**/*.pp'] +
+                       Dir['modules/*/metadata.json'] +
                        Dir['modules/*/files/**/*'] +
                        Dir['modules/*/lib/**/*.rb'] +
                        Dir['modules/*/locales/**/*'] +
--- a/lib/bolt/bolt_option_parser.rb
+++ b/lib/bolt/bolt_option_parser.rb
@@ -32,6 +32,7 @@ Available subcommands:
   bolt plan run <plan> [params]    Run a Puppet task plan
   bolt apply <manifest>            Apply Puppet manifest code
   bolt puppetfile install          Install modules from a Puppetfile into a Boltdir
+  bolt puppetfile show-modules     List modules available to Bolt
 
 Run `bolt <subcommand> --help` to view specific examples.
 
@@ -101,6 +102,7 @@ Usage: bolt puppetfile <action> [options]
 
 Available actions are:
   install                          Install modules from a Puppetfile into a Boltdir
+  show-modules                     List modules available to Bolt
 
 Install modules into the local Boltdir
   bolt puppetfile install
--- a/lib/bolt/cli.rb
+++ b/lib/bolt/cli.rb
@@ -30,7 +30,7 @@ module Bolt
                  'task' => %w[show run],
                  'plan' => %w[show run],
                  'file' => %w[upload],
-                 'puppetfile' => %w[install],
+                 'puppetfile' => %w[install show-modules],
                  'apply' => %w[] }.freeze
 
     attr_reader :config, :options
@@ -259,6 +259,9 @@ module Bolt
           end
         end
         return 0
+      elsif options[:action] == 'show-modules'
+        list_modules
+        return 0
       end
 
       message = 'There may be processes left executing on some nodes.'
@@ -398,6 +401,10 @@ module Bolt
       outputter.print_apply_result(results)
 
       results.ok ? 0 : 1
+    end
+
+    def list_modules
+      outputter.print_module_list(pal.list_modules)
     end
 
     def install_puppetfile(puppetfile, modulepath)
--- a/lib/bolt/outputter/human.rb
+++ b/lib/bolt/outputter/human.rb
@@ -203,7 +203,12 @@ module Bolt
             @stream.puts('(no modules installed)')
           else
             module_info = modules.map do |m|
-              version = m[:version] || '???'
+              version = if m[:version].nil?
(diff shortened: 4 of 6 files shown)


donoghuc:
I think that we should exclude the modules we bundle with bolt. I guess that kind of hides some of the core puppet modules but the experience after packaging is not great. 
```
root@wwg6boekoxa3csf:~# bolt module show
/opt/puppetlabs/bolt/lib/ruby/gems/2.5.0/gems/bolt-1.10.0/bolt-modules
boltlib   ???
ctrl      ???
file      ???
system    ???

/opt/puppetlabs/bolt/lib/ruby/gems/2.5.0/gems/bolt-1.10.0/modules
aggregate            ???
augeas_core          ???
canary               ???
cron_core            ???
facts                ???
host_core            ???
mount_core           ???
package              ???
puppet_agent         ???
puppet_conf          ???
puppetdb_fact        ???
python_task_helper   ???
reboot               ???
ruby_task_helper     ???
scheduled_task       ???
selinux_core         ???
service              ???
sshkeys_core         ???
yumrepo_core         ???
zfs_core             ???
zone_core            ???
```

Sharpie:
> I think that we should exclude the modules we bundle with bolt.

I disagree. One of the reasons I added this was to make it possible to easily locate the core modules in order to troubleshoot issues with them.

> the experience after packaging is not great.

I also believe packaging of the core modules should retain `metadata.json` files. This has at least two benefits:

  - It lets users know which version they have bundled. This can be used to determine if a core module has a particular feature or is affected by a particular bug.

  - Having `metadata.json` with a populated `dependencies` field allows the Puppet module loader to optimize lookup of classes, functions, etc. This is important in larger environments as it means lookup times scale on the order of the dependency list instead of total modules installed in the environment.

donoghuc:
Shipping the modules with their metadata.json seems like a win. The reason I was hesitant to expose where the bundled modules are is that for `{task,plan} show` we have not shown the bundled module paths. I think this is mostly to encourage users to NOT add modules in the packaged location. 
```
root@wwg6boekoxa3csf:/opt/puppetlabs/bolt/lib/ruby/gems/2.5.0/gems# bolt task show
facts                    Gather system facts
package                  Manage and inspect the state of packages
puppet_agent::install    Install the Puppet agent package
puppet_agent::version    Get the version of the Puppet agent package installed. Returns nothing if none present.
puppet_conf              Inspect puppet agent configuration settings
reboot                   Reboots a machine
reboot::last_boot_time   Gets the last boot time of a Linux or Windows system
service                  Manage and inspect the state of services

MODULEPATH:
/root/.puppetlabs/bolt/modules:/root/.puppetlabs/bolt/site

Use `bolt task show <task-name>` to view details and parameters for a specific task.
root@wwg6boekoxa3csf:/opt/puppetlabs/bolt/lib/ruby/gems/2.5.0/gems# bolt task show facts

facts - Gather system facts

USAGE:
bolt task run --nodes <node-name> facts

MODULE:
built-in module
```

Sharpie:
How about adding a disclaimer to the bundled module paths, such as:

```
/opt/puppetlabs/bolt/lib/ruby/gems/2.5.0/gems/bolt-1.10.0/bolt-modules
(these modules are bundled with bolt --- modifications will be overwritten on upgrade)
boltlib   ???
ctrl      ???
file      ???
system    ???
```

And `"bundled_module": true` in the JSON output.

Sharpie:
Added a commit that implements the disclaimer for bundled module directories.  Let me know if it looks better or needs tweaking --- it's easy to revert or amend.

adreyer:
I really don't want people to see what that modulepath is. We've already run into issues with people trying to put modules in them

Sharpie:
If we've run into issues from people putting modules in the internal paths, then that makes it more critical to have a command that can list out the contents of those paths --- so we can verify things are as expected.

I think the solution to this problem is user education. PE has the same sort of situation with the `/opt/puppetlabs/puppet/modules` and it is mostly a non-issue since users know that isn't a spot to be adding modules.

Sharpie:
After thinking about this a bit more, I pushed an update that omits the internal module paths from the default output, but provides the info if `--format json` is requested.

Does this seem like a good compromise between keeping the paths from being an attractive nuisance, but making the information findable for those that need to troubleshoot Bolt and its built-in module content?

adreyer:
1. we bikeshedded and decide that this shouldn't clutter up the top level help text. The best we could come up with was `bolt puppetfile show-modules` 

2. I do think we should show the plan language and bundled modules. I just don't think we should display the path to where they're installed. IMO This is better:

We also need to start including metadata.json in the package. But I'm ok merging this in first.

```
Plan Language Modules
boltlib   ???
ctrl      ???
file      ???
system    ???

/Users/adreyer/src/bolt/Boltdir/modules
puppetlabs/apt          5.0.1
puppetlabs/concat       4.2.1
puppetlabs/panos        0.2.0
puppetlabs/postgresql   5.7.0
simple                  ???
puppetlabs/stdlib       4.25.1
puppetlabs/vcsrepo      2.4.0

/Users/adreyer/src/bolt/Boltdir/site
data     ???
devbox   ???
rem      ???

Packaged modules
aggregate            ???
augeas_core          ???
canary               ???
cron_core            ???
facts                ???
host_core            ???
mount_core           ???
package              ???
puppet_agent         ???
puppet_conf          ???
puppetdb_fact        ???
python_task_helper   ???
reboot               ???
ruby_task_helper     ???
scheduled_task       ???
selinux_core         ???
service              ???
sshkeys_core         ???
yumrepo_core         ???
zfs_core             ???
zone_core            ???
```

adreyer:
Bolt.gemspec should be updated to include `metadata.json` too.

Sharpie:
Sounds good to me. I've updated this PR with the requested changes.

I've filed [BOLT-1129](https://tickets.puppetlabs.com/browse/BOLT-1129) and PR #870 for adding the `metadata.json` files.

PR #862, opened 2019-02-07
(BOLT-920) Remove solved known issue ssh key format

donoghuc:
BOLT-920 has been resolved fixing this known issue.


donoghuc:
This still seems to be unresolved. Closing.

PR #865, opened 2019-02-12
(BOLT-146) Add configuration option to set task interpreter

donoghuc:
This commit adds a configuration option to the local transport to configure what interpreter to run a task with. If an interpreter is set for a task with a given extension then that will override the shebang in the task.

### DIFF
--- a/lib/bolt/config.rb
+++ b/lib/bolt/config.rb
@@ -38,30 +38,6 @@ module Bolt
                            private-key tty tmpdir user connect-timeout
                            cacert token-file service-url interpreters].freeze
 
-    # TODO: move these to the transport themselves
-    TRANSPORT_SPECIFIC_DEFAULTS = {
-      ssh: {
-        'connect-timeout' => 10,
-        'host-key-check' => true,
-        'tty' => false
-      },
-      winrm: {
-        'connect-timeout' => 10,
-        'ssl' => true,
-        'ssl-verify' => true
-      },
-      pcp: {
-        'task-environment' => 'production'
-      },
-      local: {
-        'interpreters' => { 'rb' => RbConfig.ruby }
-      },
-      docker: {},
-      remote: {
-        'run-on' => 'localhost'
-      }
-    }.freeze
-
     def self.default
       new(Bolt::Boltdir.new('.'), {})
     end
@@ -93,8 +69,9 @@ module Bolt
       @log = { 'console' => {} }
 
       @transports = {}
-      TRANSPORTS.each_key do |transport|
-        @transports[transport] = TRANSPORT_SPECIFIC_DEFAULTS[transport].dup
+
+      TRANSPORTS.each do |key, transport|
+        @transports[key] = transport.default_options
       end
 
       update_from_file(config_data)
--- a/lib/bolt/config.rb
+++ b/lib/bolt/config.rb
@@ -36,7 +36,7 @@ module Bolt
 
     TRANSPORT_OPTIONS = %i[password run-as sudo-password extensions
                            private-key tty tmpdir user connect-timeout
-                           cacert token-file service-url].freeze
+                           cacert token-file service-url interpreters].freeze
 
     # TODO: move these to the transport themselves
     TRANSPORT_SPECIFIC_DEFAULTS = {
@@ -53,7 +53,9 @@ module Bolt
       pcp: {
         'task-environment' => 'production'
       },
-      local: {},
+      local: {
+        'interpreters' => { 'rb' => RbConfig.ruby }
+      },
       docker: {},
       remote: {
         'run-on' => 'localhost'
@@ -114,6 +116,12 @@ module Bolt
       Bolt::Util.deep_clone(self)
     end
 
+    def normalize_interpreters(interpreters)
+      Bolt::Util.walk_keys(interpreters) do |key|
+        key.chars[0] == '.' ? key : '.' + key
+      end
+    end
+
     def normalize_log(target)
       return target if target == 'console'
       target = target[5..-1] if target.start_with?('file:')
@@ -168,7 +176,10 @@ module Bolt
       TRANSPORTS.each do |key, impl|
         if data[key.to_s]
           selected = impl.filter_options(data[key.to_s])
-          @transports[key].merge!(selected)
+          @transports[key] = Bolt::Util.deep_merge(@transports[key], selected)
+        end
+        if @transports[key]['interpreters']
+          @transports[key]['interpreters'] = normalize_interpreters(@transports[key]['interpreters'])
         end
       end
     end
--- a/lib/bolt/transport/base.rb
+++ b/lib/bolt/transport/base.rb
@@ -87,6 +87,10 @@ module Bolt
         impl
       end
 
+      def select_interpreter(executable, interpreters)
+        interpreters[Pathname(executable).extname] if interpreters
+      end
+
       def reject_transport_options(target, options)
         if target.options['run-as']
           options.reject { |k, _v| k == '_run_as' }
--- a/lib/bolt/transport/base.rb
+++ b/lib/bolt/transport/base.rb
@@ -48,6 +48,10 @@ module Bolt
               "self.options() or self.filter_options(unfiltered) must be implemented by the transport class"
       end
 
+      def self.default_options
+        {}
+      end
+
       def self.filter_options(unfiltered)
         unfiltered.select { |k| options.include?(k) }
       end
--- a/lib/bolt/transport/docker.rb
+++ b/lib/bolt/transport/docker.rb
@@ -8,7 +8,7 @@ module Bolt
   module Transport
     class Docker < Base
       def self.options
-        %w[service-url service-options tmpdir]
+        %w[service-url service-options tmpdir interpreters]
       end
 
       def provided_features
@@ -87,7 +87,7 @@ module Bolt
         arguments = unwrap_sensitive_args(arguments)
         with_connection(target) do |conn|
           execute_options = {}
-
+          execute_options[:interpreter] = select_interpreter(executable, target.options['interpreters'])
           conn.with_remote_tempdir do |dir|
             if extra_files.empty?
               task_dir = dir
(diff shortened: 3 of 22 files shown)


donoghuc:
This is a WIP PR to iron out some questions that came up when working on this. 

An example:
Task defined in `interpreter.py`
```
#!/home/cas/miniconda/pkgs/python-3.7.0-hc3d631a_0/bin/python
import platform
print(platform.python_version())
```
When run without overriding shebang in task the output is:
```
cas@cas-ThinkPad-T460p:~/working_dir/bolt/Boltdir$ bolt task run test::interpreter -n localhost
Started on localhost...
Finished on localhost:
  3.7.0
  {
  }
Ran on 1 node in 0.08 seconds
```
Now override the shebang in the inventory config. 
```
nodes:
  - name: localhost
    config:
      transport: local
      local:
        interpreters:
          py: /usr/bin/python
```
```
cas@cas-ThinkPad-T460p:~/working_dir/bolt/Boltdir$ bolt task run test::interpreter -n localhost
Started on localhost...
Finished on localhost:
  2.7.12
  {
  }
Ran on 1 node in 0.07 seconds
```
In the ticket description we have an example that looks like: 
```
config:
  interpreters:
    rb: /usr/bin/ruby
```
Which seems to indicate that we intend the `interpreters` key to be at the same level as transport config instead of nested under the transport key. I could be mistaken but the `interpreters` map seems to be transport specific? For example the map for `winrm` would likely be different than `ssh` or `docker`.

lib/bolt/config.rb
```diff
@@ -36,7 +36,7 @@ class Config
 
     TRANSPORT_OPTIONS = %i[password run-as sudo-password extensions
                            private-key tty tmpdir user connect-timeout
-                           cacert token-file service-url].freeze
+                           cacert token-file service-url interpreters].freeze
 
     # TODO: move these to the transport themselves
```
donoghuc: I intend to do this as soon as I have tests passing.

lib/bolt/transport/base.rb
```diff
@@ -87,6 +87,15 @@ def select_implementation(target, task)
         impl
       end
 
+      def select_interpreter(executable, interpreters)
```
donoghuc: I thought about adjusting for case here as well? For example  `.rb,rb,.RB,.rb` would all map to ruby interpreter specified? Is that too much "magic"? Is there a sane workflow that someone would choose to map `.rb` to one interpreter and `.RB` to another?

donoghuc: The ticket states "extension name to absolute path of an executable (exact match to extension)" which I interpret to mean do not try to ignore case.

lib/bolt/transport/docker.rb
```diff
@@ -112,7 +112,7 @@ def run_task(target, task, arguments, _options = {})
             end
 
             stdout, stderr, exitcode = conn.execute(remote_task_path, execute_options)
-            Bolt::Result.for_task(target, stdout.join, stderr.join, exitcode)
+            Bolt::Result.for_task(target, stdout.join, stderr.join.force_encoding("UTF-8"), exitcode)
```
donoghuc: I added this to get tests passing because Non-utf characters from sterr were causing https://github.com/puppetlabs/bolt/blob/e4fe443c90e151ed087cbbb23bbc76a401cba12b/lib/bolt/result.rb#L119-L121 to stack trace. Any suggestions for where to make sure these characters are serializeable?

pre-docs/bolt_configuration_options.md
```diff
@@ -52,6 +55,8 @@ ssh:
 
 `tmpdir`: The directory to upload and execute temporary files on the target.
 
+`interpreters`: A map of extension name to absolute path of an executable. This allows a user to override the shebang defined in a task executable.
```
donoghuc: Waiting for review before I paste this in the other relevant transports. Also will need to add a note that the `local` transport will default to bolt's ruby.

lib/bolt/transport/local.rb
```diff
@@ -146,25 +148,32 @@ def run_task(target, task, arguments, _options = {})
...
+              output =
                 if input_method == 'powershell'
-                  @conn.execute(command, dir: dir, env: "powershell.exe")
+                  @conn.execute(command, dir: dir, interpreter: interpreter)
                 else
-                  @conn.execute(command, dir: dir, stdin: stdin, env: "powershell.exe")
+                  @conn.execute(command, dir: dir, stdin: stdin, interpreter: interpreter)
                 end
+            end
+            unless output
```
donoghuc: In general this nested logic for dealing with windows, interpreter, etc is getting messy. I got it working to develop tests. I think it could use some re-organization.

lib/bolt/transport/base.rb
nicklewis: I think honoring case is the right thing to do here. But we should normalize the extensions (to either always include 
or never include a `.`)  when processing the config, so we only have to look up a single key. Because `File.extname()` return an extension containing a `.`, it probably makes sense to normalize the config option to always include a `.`

donoghuc:
Removed `DO NOT MERGE` label and the WIP from the PR title. I will remove the WIP from the commit message when I address reviewer feedback. This PR is ready for review. Sorry if those labels were misleading...

nmaludy:
Is there an advantage to using this over changing the shebang at the top of the script file?

donoghuc:
@nmaludy I think the main advantage is that the implementation will remain consistent between tasks targeting windows/non-windows targets. For now we assume that ssh targets are unix-like and winrm targets are windows. If we relied on modifying shebang for ssh targets and someday supported ssh on windows then we would likely have to account for both methods.

nmaludy:
@donoghuc that makes sense, i'm also wondering if the interpreter could be used to help select which implementation of a task to use?

adreyer:
@nmaludy I don't think we can imply features from this. Users may want to set an interpreter for all  `.py` tasks in their environment without worrying about which targets actually have that interpreter installed.

At this point the task spec does not include anything about altering the task file. There are cases where this will behave differently than altering the shebang but I'd rather make sure those cases are important before we start altering task files themselves.

lib/bolt/transport/docker.rb
nicklewis: What was the scenario in which you were seeing non-utf characters? I'm surprised this worked, because generally all `force_encoding` does is cause the string to _claim_ it's encoded in UTF-8. I would expect a `valid_encoding?` check to fail after doing that, because the bytes are still not actually valid UTF-8.

lib/bolt/transport/local.rb
```diff
@@ -146,25 +154,32 @@ def run_task(target, task, arguments, _options = {})
...
-                command = Powershell.run_ps_task(arguments, script, input_method)
-                command = environment_params + Powershell.shell_init + command
+            if Powershell.powershell_file?(script) && stdin.nil?
+              command = Powershell.run_ps_task(arguments, script, input_method)
+              command = environment_params + Powershell.shell_init + command
+              interpreter ||= 'powershell.exe'
+              output =
                 if input_method == 'powershell'
-                  @conn.execute(command, dir: dir, env: "powershell.exe")
+                  @conn.execute(command, dir: dir, interpreter: interpreter)
```
nicklewis: Hooray! I never liked that we used `env` for that before.

nicklewis:
This looks solid to me, but I would like to see the change to normalize extensions.

lib/bolt/transport/docker.rb
donoghuc: Running interpreter.py against `ubuntu_node` defined in bolt/spec. You can reproduce with:

`test.py`
```
#!/foo/bar
print('nope')
```
```
cas@cas-ThinkPad-T460p:~/working_dir/bolt$ bolt task run test::test -n ubuntu_node --transport docker
Started on ubuntu_node...
Failed on ubuntu_node:
<Encoding::UndefinedConversionError> "\xE2" from ASCII-8BIT to UTF-8
	/home/cas/working_dir/bolt/lib/bolt/result.rb:120:in `encode'
	/home/cas/working_dir/bolt/lib/bolt/result.rb:120:in `to_json'
	/home/cas/working_dir/bolt/lib/bolt/result.rb:120:in `to_json'
	/home/cas/working_dir/bolt/lib/bolt/executor.rb:205:in `with_node_logging'
	/home/cas/working_dir/bolt/lib/bolt/executor.rb:252:in `block (2 levels) in run_task'
	/home/cas/working_dir/bolt/lib/bolt/executor.rb:84:in `block (3 levels) in queue_execute'
	/home/cas/working_dir/bolt/.bundle/ruby/2.5.0/gems/concurrent-ruby-1.1.4/lib/concurrent/executor/ruby_thread_pool_executor.rb:348:in `run_task'
	/home/cas/working_dir/bolt/.bundle/ruby/2.5.0/gems/concurrent-ruby-1.1.4/lib/concurrent/executor/ruby_thread_pool_executor.rb:337:in `block (3 levels) in create_worker'
	/home/cas/working_dir/bolt/.bundle/ruby/2.5.0/gems/concurrent-ruby-1.1.4/lib/concurrent/executor/ruby_thread_pool_executor.rb:320:in `loop'
	/home/cas/working_dir/bolt/.bundle/ruby/2.5.0/gems/concurrent-ruby-1.1.4/lib/concurrent/executor/ruby_thread_pool_executor.rb:320:in `block (2 levels) in create_worker'
	/home/cas/working_dir/bolt/.bundle/ruby/2.5.0/gems/concurrent-ruby-1.1.4/lib/concurrent/executor/ruby_thread_pool_executor.rb:319:in `catch'
	/home/cas/working_dir/bolt/.bundle/ruby/2.5.0/gems/concurrent-ruby-1.1.4/lib/concurrent/executor/ruby_thread_pool_executor.rb:319:in `block in create_worker'
	/home/cas/working_dir/bolt/.bundle/ruby/2.5.0/gems/logging-2.2.2/lib/logging/diagnostic_context.rb:474:in `block in create_with_logging_context'
  The task failed with exit code 127:
  env: ‘/tmp/9dc6641e-c583-4f5e-a4ba-13bd4d1c423b/test/tasks/test.ps1’: No such file or directory
 
  {
  }
Failed on 1 node: ubuntu_node
Ran on 1 node in 1.06 seconds
```

lib/bolt/transport/local.rb
donoghuc: Yeah that was not great. For some reason at the time I was hesitant to make any changes to `shell.rb`.

donoghuc:
Updated pre-docs, added normalization to interpreters options as Nick requested and came up with a more robust solution for handling non-UTF values from executing on docker transport.

PR #869, opened 2019-02-15
(BOLT-1057) Standardize BoltSpec::Run helper method signatures

nicklewis:
Previously, these methods had confusing and inconsistent signatures
using a mix of optional parameters and keyword arguments. In some cases,
the optional parameters were themselves expects to be hashes, which
conflicted with the keyword arguments. That caused confusing behavior
when invoking the functions with only some arguments, as a hash intended
to be a positional argument could be interpreted as an _invalid_ keyword
arguments hash, or vice versa.

These functions now always require their `params` (or equivalent)
argument, leading to slightly more verbose but always clear
invocations. The functions which previously expected an options hash in
addition to their keyword arguments now expect the options hash to
be provided as a keyword argument named `options`.

### DIFF
--- a/lib/bolt_spec/run.rb
+++ b/lib/bolt_spec/run.rb
@@ -11,36 +11,35 @@ require 'bolt/util'
 # This is intended to provide a relatively stable method of executing bolt in process from tests.
 module BoltSpec
   module Run
-    def run_task(task_name, targets, params = nil, config: nil, inventory: nil)
+    def run_task(task_name, targets, params, config: nil, inventory: nil)
       result = BoltRunner.with_runner(config, inventory) do |runner|
-        runner.run_task(task_name, targets, params || {})
+        runner.run_task(task_name, targets, params)
       end
       result = result.to_a
       Bolt::Util.walk_keys(result, &:to_s)
     end
 
-    def run_plan(plan_name, params = nil, config: nil, inventory: nil)
+    def run_plan(plan_name, params, config: nil, inventory: nil)
       # Users copying code from run_task may forget that targets is not a parameter for run plan
-      params ||= {}
       raise ArgumentError, "params must be a hash" unless params.is_a?(Hash)
 
       result = BoltRunner.with_runner(config, inventory) do |runner|
-        runner.run_plan(plan_name, params || {})
+        runner.run_plan(plan_name, params)
       end
 
       { "status" => result.status,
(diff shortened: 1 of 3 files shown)


nicklewis:
This will require some corresponding changes to a few modules that use `BoltSpec`. I know of at least `puppetlabs/puppetlabs-package`, `puppetlabs/puppetlabs-service`, `puppetlabs/puppetlabs-facts`, `puppetlabs/puppetlabs-puppet_agent`, `puppetlabs/puppetlabs-reboot.

nicklewis:
It doesn't look like package or service need changes, because they only use `run_task` and they always pass the params argument. I've filed corresponding PRs for the three other modules:

https://github.com/puppetlabs/puppetlabs-puppet_agent/pull/389
https://github.com/puppetlabs/puppetlabs-facts/pull/22
https://github.com/puppetlabs/puppetlabs-reboot/pull/201

nicklewis:
For the most part, consumers of this code can support both old and new versions of BoltSpec by always passing the `params`/`arguments` option.

The only case where there isn't a fully compatible change available is anyone passing `options` positionally to `run_command`, since it's now a keyword argument. That's only necessary if you want to use `_description` or `_run_as` though, so it's unlikely to be currently used.

lib/bolt_spec/run.rb
```diff
@@ -11,36 +11,35 @@
 # This is intended to provide a relatively stable method of executing bolt in process from tests.
 module BoltSpec
   module Run
-    def run_task(task_name, targets, params = nil, config: nil, inventory: nil)
+    def run_task(task_name, targets, params, config: nil, inventory: nil)
```
adreyer: Calling out a breaking change for run_task without parameters.

lib/bolt_spec/run.rb
```diff
@@ -11,36 +11,35 @@
...
       result = BoltRunner.with_runner(config, inventory) do |runner|
-        runner.run_command(command, targets, params)
+        runner.run_command(command, targets, options)
       end
       result = result.to_a
       Bolt::Util.walk_keys(result, &:to_s)
     end
 
-    def run_script(script, targets, arguments = nil, options = {}, config: nil, inventory: nil)
+    def run_script(script, targets, arguments, options: {}, config: nil, inventory: nil)
```
adreyer: Calling out a breaking change for run_script without arguments.

PR #870, opened 2019-02-16
(BOLT-1129) Add metadata.json files to bundled modules

Sharpie:
This change set updates the Bolt gemspec to include `metadata.json` in gem packages. Templates for generating `metadata.json` files have been added to internal modules such as boltlib. The Rakefile has also been updated with logic for rendering these templates whenever they change, or the Bolt version number is updated. Template rendering has also been added as a dependency of the `rake build` task.

### DIFF
--- a/bolt.gemspec
+++ b/bolt.gemspec
@@ -19,6 +19,7 @@ Gem::Specification.new do |spec|
                        Dir['lib/**/*.rb'] +
                        Dir['lib/**/*.json'] +
                        Dir['libexec/*'] +
+                       Dir['bolt-modules/*/metadata.json'] +
                        Dir['bolt-modules/*/lib/**/*.rb'] +
                        Dir['bolt-modules/*/types/**/*.pp'] +
                        Dir['modules/*/files/**/*'] +
--- a/.gitignore
+++ b/.gitignore
@@ -36,6 +36,12 @@ Gemfile.local
 
 acceptance/hosts.yaml
 
+## Generated metadata.json for internal modules
+bolt-modules/*/metadata.json
+modules/aggregate/metadata.json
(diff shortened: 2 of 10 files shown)


Sharpie:
Flagging as "Do Not Merge" for now as the metadata templates are completely filled out. Descriptions are generic  and the dependency lists are not populated. Tossing this up for discussion to see if the approach is sound before completely filling out the templates.

If this PR is accepted, we will need to update puppetlabs/bolt-vanagon to use `rake build` instead of `gem build` to trigger auto-generation of the `metadata.json` files.

bolt.gemspec
```diff
@@ -19,8 +19,10 @@ Gem::Specification.new do |spec|
                        Dir['lib/**/*.rb'] +
                        Dir['lib/**/*.json'] +
                        Dir['libexec/*'] +
+                       Dir['bolt-modules/*/metadata.json'] +
                        Dir['bolt-modules/*/lib/**/*.rb'] +
                        Dir['bolt-modules/*/types/**/*.pp'] +
+                       Dir['modules/*/metadata.json'] +
```
adreyer: I think we should merge this change in separately since it's obviously useful.

Sharpie: Moved this one over to PR #861.

Sharpie:
To close the loop: we decided that things in `bolt-modules/` don't need `metadata.json` files because:

  - They have no dependencies for which lookups could be optimized.

  - `bolt-modules/` is at the front of the modulepath, so searches of the entire environment will encounter it first and terminate early.

The remainder of the bundled content under `modules/` is slated to be spun out to their own repos and are not worth optimizing at this time.

PR #871, opened 2019-02-19
(BOLT-153) Windows SMB file transfer support

Iristyle:
Updated version of #856 - adds / addresses a small number of new things

- Fixes `Vagrantfile` to port forward 445 to 2445 of host (else there is no way to test locally)
- Errors when SSL is specified in conjunction with SMB (as encryption is not yet available)
- Removes SMB1 support
- Uses `WindowsError::NTStatus` status objects rather than embedding Win32 error codes
- Improves error handling / messages when SMB port gets a connection refused error
- Adds smb-port configuration option (defaults to port 445)
- Adds tests against SMB transfers
- Adds recursive directory copy over SMB

From original PR courtesy @joshcooper:

WinRM sucks for file transfer because it base64 encodes the payload,
compresses it as a zip file, transfers the contents via Powershell in
SOAP/XML messages, and decompresses on the remote side.

This commit allows bolt to upload files using SMB using the Rapid7 gem.
If anyone understands SMB, it's Rapid7/metasploit. The WinRM connection
lazily generates an SMBClient if a file is uploaded. The connection
will only login once, but connect to and disconnect from the share for
each uploaded file.

Before this commit transferring a 75MB jre takes 58.7 seconds to a local
Windows VM. With this commit it takes 6.2 seconds.

By default, WinRM connections continue to use WinRM for file transfers.
To enable SMB, add the following to the inventory file:

    winrm:
      file-protocol: smb

When uploading a file, the destination path should be specified as
either a drive letter or UNC path. If a drive letter is given, then bolt
will use the administrative share, usually `\\<host>\C$`. Note this
requires administrative privileges. To upload a file as a non-admin
user, enter the share as a UNC path. If bolt is running on a Windows
host:

    PS C:\> bolt file upload C:\path\to\file \\host\share\path\to\filebolt file upload C:\path\to\file \\host\share\path\to\file

If bolt is running on a non-Windows host, backslashes in the destination
path need to be escaped:

    $ bolt file upload /path/to/file \\\\host\\share\\path\\to\\file

Or forward slashes can be used (for drive letter or UNC paths):

    $ bolt file upload /path/to/file C:/path/to/file

### DIFF
--- a/bolt.gemspec
+++ b/bolt.gemspec
@@ -45,6 +45,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency "puppet", [">= 6.0.1", "< 7"]
   spec.add_dependency "puppet-resource_api"
   spec.add_dependency "r10k", "~> 2.6"
+  spec.add_dependency "ruby_smb", "~> 1.0"
   spec.add_dependency "terminal-table", "~> 1.8"
   spec.add_dependency "winrm", "~> 2.0"
   spec.add_dependency "winrm-fs", "~> 1.3"
--- a/lib/bolt/config.rb
+++ b/lib/bolt/config.rb
@@ -36,7 +36,7 @@ module Bolt
 
     TRANSPORT_OPTIONS = %i[password run-as sudo-password extensions
                            private-key tty tmpdir user connect-timeout
-                           cacert token-file service-url interpreters].freeze
+                           cacert token-file service-url interpreters file-protocol].freeze
 
     def self.default
       new(Bolt::Boltdir.new('.'), {})
--- a/lib/bolt/config.rb
+++ b/lib/bolt/config.rb
@@ -36,7 +36,7 @@ module Bolt
 
     TRANSPORT_OPTIONS = %i[password run-as sudo-password extensions
                            private-key tty tmpdir user connect-timeout
-                           cacert token-file service-url interpreters file-protocol].freeze
+                           cacert token-file service-url interpreters file-protocol smb-port].freeze
 
     def self.default
       new(Bolt::Boltdir.new('.'), {})
--- a/lib/bolt/transport/winrm.rb
+++ b/lib/bolt/transport/winrm.rb
@@ -8,7 +8,10 @@ module Bolt
   module Transport
     class WinRM < Base
       def self.options
-        %w[port user password connect-timeout ssl ssl-verify tmpdir cacert extensions interpreters file-protocol]
+        %w[
+          port user password connect-timeout ssl ssl-verify tmpdir cacert
+          extensions interpreters file-protocol smb-port
+        ]
       end
 
       def self.default_options
--- a/lib/bolt/transport/winrm.rb
+++ b/lib/bolt/transport/winrm.rb
@@ -8,14 +8,15 @@ module Bolt
   module Transport
     class WinRM < Base
       def self.options
-        %w[port user password connect-timeout ssl ssl-verify tmpdir cacert extensions interpreters]
+        %w[port user password connect-timeout ssl ssl-verify tmpdir cacert extensions interpreters file-protocol]
       end
 
       def self.default_options
         {
           'connect-timeout' => 10,
           'ssl' => true,
-          'ssl-verify' => true
+          'ssl-verify' => true,
+          'file-protocol' => 'winrm'
         }
       end
 
@@ -32,6 +33,10 @@ module Bolt
         ssl_flag = options['ssl']
         unless !!ssl_flag == ssl_flag
           raise Bolt::ValidationError, 'ssl option must be a Boolean true or false'
+        end
+
+        if ssl_flag && (options['file-protocol'] == 'smb')
+          raise Bolt::ValidationError, 'SMB file transfers are not allowed with SSL enabled'
         end
 
         ssl_verify_flag = options['ssl-verify']
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -89,7 +89,7 @@ To exclude tests that rely on Vagrant, run:
 
 Windows includes additional tests that require a full Windows Server VM to run; we run them in AppVeyor. If you need to run the tests locally set `APPVEYOR_AGENTS=true`, re-run `vagrant up` to create a `Windows Server 2016 Core` VM, and run tests with
 
-    $ BOLT_WINRM_PORT=35985 BOLT_WINRM_USER=vagrant BOLT_WINRM_PASSWORD=vagrant bundle exec rake integration:appveyor_agents
+    $ BOLT_WINRM_PORT=35985 BOLT_WINRM_SMB_PORT=3445 BOLT_WINRM_USER=vagrant BOLT_WINRM_PASSWORD=vagrant bundle exec rake integration:appveyor_agents
 
 ### `rubocop` on Windows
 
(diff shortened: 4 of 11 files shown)


lib/bolt/config.rb
```diff
@@ -48,7 +48,9 @@ class Config
       winrm: {
         'connect-timeout' => 10,
         'ssl' => true,
-        'ssl-verify' => true
+        'ssl-verify' => true,
+        'file-protocol' => 'winrm',
+        'smb-port' => 445
```
Iristyle: @joshcooper / @adreyer I needed to add an `smb-port` configuration option for testing purposes. We had previously discussed the potential for multiple config options for SMB at https://github.com/puppetlabs/bolt/pull/856#discussion_r253982265 and where they should land in the options hierarchy.

If we intend to add more SMB options, is now the right time to break that out to a separate category?

adreyer: If we have defaults for these options putting them at the top level may be best. Otherwise we add complexity around nested merging that I don't think we're ready to handle yet. We can always migrate to a nested hash later if there are so many smb options that it becomes a problem.

Iristyle: 👍 	Thanks

I'd also prefer not to deal with adding nested structures and any resulting problems.

Iristyle:
Everything is good against the Nano box and in AppVeyor. However, I'm running into a snag testing the 2016 Vagrant machine locally now:


```
1) Bolt::Transport::WinRM with an open connection can upload a file to a host using smb
     Failure/Error: raise Bolt::Node::FileError.new(e.message, 'WRITE_ERROR')

     Bolt::Node::FileError:
       Read timeout expired when reading from the Socket (timeout=30)
     # ./lib/bolt/transport/winrm/connection.rb:183:in `rescue in write_remote_file_smb'
     # ./lib/bolt/transport/winrm/connection.rb:160:in `write_remote_file_smb'
     # ./lib/bolt/transport/winrm/connection.rb:146:in `write_remote_file'
diff --git a/CONTRIBUTING.md b/CONTRIBUTING.md
index aa65e13..1277532 100644
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -89,7 +89,7 @@ To exclude tests that rely on Vagrant, run:

     # ./lib/bolt/transport/winrm.rb:68:in `block in upload'
     # ./lib/bolt/transport/winrm.rb:57:in `with_connection'
     # ./lib/bolt/transport/winrm.rb:67:in `upload'
     # ./spec/bolt/transport/winrm_spec.rb:241:in `block (5 levels) in <top (required)>'
     # ./spec/lib/bolt_spec/files.rb:16:in `block in with_tempfile_containing'
     # ./spec/lib/bolt_spec/files.rb:13:in `with_tempfile_containing'
     # ./spec/bolt/transport/winrm_spec.rb:239:in `block (4 levels) in <top (required)>'
     # ------------------
     # --- Caused by: ---
     # RubySMB::Error::CommunicationError:
     #   Read timeout expired when reading from the Socket (timeout=30)
     #   ./.bundle/gems/ruby/2.4.0/gems/ruby_smb-1.0.5/lib/ruby_smb/dispatcher/socket.rb:65:in `recv_packet'

  2) Bolt::Transport::WinRM with an open connection can upload a directory to a host using smb
     Failure/Error: raise Bolt::Node::FileError.new(e.message, 'WRITE_ERROR')

     Bolt::Node::FileError:
       Read timeout expired when reading from the Socket (timeout=30)
     # ./lib/bolt/transport/winrm/connection.rb:183:in `rescue in write_remote_file_smb'
     # ./lib/bolt/transport/winrm/connection.rb:160:in `write_remote_file_smb'
     # ./lib/bolt/transport/winrm/connection.rb:146:in `write_remote_file'
     # ./lib/bolt/transport/winrm.rb:68:in `block in upload'
     # ./lib/bolt/transport/winrm.rb:57:in `with_connection'
     # ./lib/bolt/transport/winrm.rb:67:in `upload'
     # ./spec/bolt/transport/winrm_spec.rb:283:in `block (5 levels) in <top (required)>'
     # ./spec/bolt/transport/winrm_spec.rb:276:in `block (4 levels) in <top (required)>'
     # ------------------
     # --- Caused by: ---
     # RubySMB::Error::CommunicationError:
     #   Read timeout expired when reading from the Socket (timeout=30)
     #   ./.bundle/gems/ruby/2.4.0/gems/ruby_smb-1.0.5/lib/ruby_smb/dispatcher/socket.rb:65:in `recv_packet'
```

Also, it looks unrelated... but contributing.md mentions doing `BOLT_WINRM_PORT=35985 BOLT_WINRM_SMB_PORT=3445 BOLT_WINRM_USER=vagrant BOLT_WINRM_PASSWORD=vagrant bundle exec rake integration:appveyor_agents`. None of those tests pass due to a failure to find tasks. I suspect they may have moved at some point. Do you know anything about that @MikaelSmith?

bolt.gemspec
```diff
@@ -45,6 +45,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency "puppet", [">= 6.0.1", "< 7"]
   spec.add_dependency "puppet-resource_api"
   spec.add_dependency "r10k", "~> 2.6"
+  spec.add_dependency "ruby_smb", "~> 1.0"
```
donoghuc: Will have to add component to bolt-vanagon. I guess https://rubygems.org/gems/ruby_smb/versions/1.0.5 should be good for now? https://github.com/puppetlabs/bolt-vanagon/tree/master/configs/components

lib/bolt/transport/winrm.rb
```diff
@@ -8,14 +8,19 @@ module Bolt
...
       end
 
       def self.default_options
         {
           'connect-timeout' => 10,
           'ssl' => true,
-          'ssl-verify' => true
+          'ssl-verify' => true,
+          'file-protocol' => 'winrm',
+          'smb-port' => 445
```
donoghuc: I think we should keep this default with the winrm port defaults. https://github.com/puppetlabs/bolt/blob/71aff3e06e3a6232b14f76d99af91edbf6d53570/lib/bolt/transport/winrm/connection.rb#L29-L30

donoghuc: Oh, you do set that in `::WinRM::Connection`. You can just delete the `smb-port` default here I think.

bolt.gemspec
Iristyle: This has already been approved as part of https://tickets.puppetlabs.com/browse/RE-12026

Iristyle: PR up at https://github.com/puppetlabs/bolt-vanagon/pull/80

donoghuc:
Would we expect that smb on default port is configured on winrm targets in bolt acceptance suite? Specifically: ` windows2016-64`,`windows10ent-64`,`windows2008-64`,`windows2016-64`. This seems like a good think to test against multiple windows target types at https://github.com/puppetlabs/bolt/blob/master/acceptance/tests/file_winrm.rb

Iristyle:
I'm not sure I totally understand your question @donoghuc - but Windows hosts should generally have 445 open by default. I'm not sure there's any additional value beyond the tests I already added.

donoghuc:
I was just wondering if it was worth testing on variety of windows versions. Sounds like no.

donoghuc:
Bolt-vanagon PR merged and through CI. Updated bolt-server schema and validated smb would work. https://github.com/puppetlabs/bolt/pull/882

joshcooper:
💯

PR #874, opened 2019-02-21
(BOLT-159) Replace known issue 159 with BOLT-1130

Iristyle:
- A simple fix was put in place for the triple quotes problem, but a
   large problem with JSON handling still looms.

   BOLT-1130 has been created to describe / address that problem and
   the known issues have been updated accordingly.


melamos:
BOLT-1130 has been added to known issues.

PR #877, opened 2019-02-21
(BOLT-1053) Update ruby_task_helper to 0.3.0

donoghuc:
The 0.3.0 release includes a bug fix that symbolizes all parameter keys passed to the TaskHelper::task method by TaskHelper::run.


donoghuc:
Is there a convention around distinguishing between class and instance methods? Something maybe like `TaskHelper::class_method` vs `TaskHelper#instance_method`?

PR #879, opened 2019-02-25
(doc) add pcp to transport options list

adreyer:


melamos:
This has been moved to docs.

PR #880, opened 2019-02-26
(BOLT-1100) Add install note about loading powershell module

nicklewis:
On PS 2.0 and 3.0, the Bolt powershell module isn't automatically
loaded. This commit adds a note to the install docs, based on the
equivalent note from the pdk docs, explaining the steps to take to load
Bolt manually.


nicklewis:
This is based _very_ closely on [the same note from the 
PDK install docs](https://puppet.com/docs/pdk/1.x/pdk_install.html#task-3944).

pre-docs/bolt_installing.md
```diff
@@ -42,6 +42,21 @@ You must have the Chocolatey package manager installed.
...
 
+### Add the Bolt module to PowerShell
+
+PowerShell versions 2.0 and 3.0 cannot automatically discover and load the Bolt module, so you'll need to add it manually.
+
+To allow PowerShell to load Bolt, add the module to your PowerShell profile.
+
+1.  Add the following line to your PowerShell profile: 
+    ```
+    Import-Module -Name "$($env:ProgramFiles)\WindowsPowerShell\Modules\PuppetBolt"
```
Iristyle: You can avoid the double `$` like this:


```powershell
Import-Module -Name ${Env:ProgramFiles}\WindowsPowerShell\Modules\PuppetBolt
```

pre-docs/bolt_installing.md
```diff
@@ -42,6 +42,21 @@ You must have the Chocolatey package manager installed.
...
+    ```
+
+2.  Close and re-open the PowerShell window to re-source your profile.
```
Iristyle: This part is unnecessary.  You can simply dot-source it like this:

```powershell
. $PROFILE
```

If you really want to streamline the whole process, the first part of the instructions can append to the `$PROFILE` at the command line like:

```powershell
'Import-Module -Name ${Env:ProgramFiles}\WindowsPowerShell\Modules\PuppetBolt' | Out-File -Append $PROFILE
```


So putting it together:

```
Windows PowerShell
Copyright (C) 2009 Microsoft Corporation. All rights reserved.

PS C:\Users\Administrator> $PSVersionTable

CLRVersion                     2.0.50727.8800
BuildVersion                   6.0.6002.18111
PSVersion                      2.0
WSManStackVersion              2.0
PSCompatibleVersions           {1.0, 2.0}
PSRemotingProtocolVersion      2.1


PS C:\Users\Administrator> bolt
The term 'bolt' is not recognized as the name of a cmdlet, function, script file, or operable program. Check the spelli
ng of the name, or if a path was included, verify that the path is correct and try again.
At line:1 char:5
+ bolt <<<<
    + CategoryInfo          : ObjectNotFound: (bolt:String) [], CommandNotFoundException
    + FullyQualifiedErrorId : CommandNotFoundException

PS C:\Users\Administrator> type $PROFILE
Get-Content : Cannot find path 'C:\Users\Administrator\Documents\WindowsPowerShell\Microsoft.PowerShell_profile.ps1' because it does not exist.
At line:1 char:5
+ type <<<<  $PROFILE
    + CategoryInfo          : ObjectNotFound: (C:\Users\Admini...ell_profile.ps1:String) [Get-Content], ItemNotFoundException
    + FullyQualifiedErrorId : PathNotFound,Microsoft.PowerShell.Commands.GetContentCommand

PS C:\Users\Administrator> 'Import-Module -Name ${Env:ProgramFiles}\WindowsPowerShell\Modules\PuppetBolt' | Out-File -Ap
pend $PROFILE

PS C:\Users\Administrator> type $PROFILE
Import-Module -Name ${Env:ProgramFiles}\WindowsPowerShell\Modules\PuppetBolt

PS C:\Users\Administrator> . $PROFILE

PS C:\Users\Administrator> bolt --version
1.13.0
```

nicklewis: Will the `$PROFILE` be idempotent? If not, it seems like reopening the window is a bit safer.

Iristyle: I suppose you know the answer to that question already given it's an arbitrary script.

However, I'm not sure idempotency is much of a concern in this particular case (it's possible a user would get duplicate additions to `PATH` or `PSModulePath` for instance which is fairly innocuous), but I don't have really strong feelings one way or another. My main goal here was just to make it "simpler" for an end user by not needing to close / re-open.

nicklewis: I'm happy with this suggestion. It seems like the odds of re-sourcing the profile causing any problems are low and the natural fix is to just close and re-open the window if anything weird happens.

I was also a little concerned about giving users a command that will change their profile directly, but if they're unhappy with that then they can just... not run it.

nicklewis:
Updated with @Iristyle's suggestions.

nicklewis:
I tested these instructions out on a Windows 2008 host with PowerShell 2.0

pre-docs/bolt_installing.md
Iristyle: 👍 that was my line of thinking - optimize for the most straightforward set of steps for a user - those who run into an issue can easily close / reopen (and probably already know that they've edited their profile / are comfortable with that).

PR #884, opened 2019-02-28
(BOLT-1151) Load and show YAML plans

nicklewis:
This allows `bolt plan show` and `bolt plan show <plan>` to find and
display YAML plans. We create a new YamlPlanEvaluator class and pass it
to Puppet via the context (as yaml_plan_instantiator), which then calls
back into Bolt whenever a YAML plan is loaded. This allows the logic for
listing and finding plans to live in Puppet, but for all of the logic
about how to run them to live in Bolt.

The execution of YAML plans follows a similar model as Puppet plans: we
create a Puppet function which, when invoked, will run the plan. We
use a PlanWrapper class to encapsulate the plan data and make it behave
like a Puppet PlanDefinition object. This commit also includes an empty
definition of the `evaluate_block_with_bindings` method, which allows
our YamlPlanEvaluator to function as though it were a regular Puppet
evaluator.

### DIFF
--- a/lib/bolt/pal.rb
+++ b/lib/bolt/pal.rb
@@ -81,6 +81,7 @@ module Bolt
 
       require 'bolt/pal/logging'
       require 'bolt/pal/issues'
+      require 'bolt/pal/yaml_plan_evaluator'
 
       # Now that puppet is loaded we can include puppet mixins in data types
       Bolt::ResultSet.include_iterable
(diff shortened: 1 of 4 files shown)


nicklewis:
This depends on some corresponding work in Puppet. Or rather, it won't do anything until the change is made in Puppet.

I'm also still working on tests for this, but wanted some early eyes on it.

PR #886, opened 2019-03-01
(BOLT-1096) Update CONTRIBUTING.md

donoghuc:
Update installation instructions and refresh information about testing.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,5 +1,9 @@
 # Contributing
 
+## Bolt Community Slack Channel 
+
+Join the `#bolt` channel in the [puppet community slack instance](https://slack.puppet.com/) where Bolt developers and community members who use and contribute to Bolt discuss the tool. Another channel of interest is the `#office-hours` where once a week a Bolt developer leads a Q&A session about using Bolt.
+
 ## Issues
 
 Please submit new issues on the GitHub issue tracker: https://github.com/puppetlabs/bolt/issues
@@ -30,6 +34,8 @@ License Agreement before we can accept your pull request: https://cla.puppet.com
 
 ## Installing Bolt
 
+If you are interested in trying bolt out or using in production we recommend installing from a system package detailed in [installing bolt](https://puppet.com/docs/bolt/latest/bolt_installing.html). The following installation instructions are focused on developers who wish to contribute to Bolt.
+
 Depending on your development workflow, you can install Bolt one of three ways:
 
 * From [RubyGems](https://rubygems.org)
@@ -50,11 +56,16 @@ To use [Bundler](https://bundler.io), add this to your Gemfile:
 
     gem 'bolt'
 
+
+Alternatively there are rake tasks defined in bundler that allow you to install your local Bolt with any changes you made for development and testing.
+
+    bundle exec rake install
+    bundle exec rake install:local
+
 ### Run Bolt from source
 
 To run Bolt from source:
 
-    git submodule update --init --recursive
     bundle install --path .bundle --without test
     bundle exec bolt ...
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -73,7 +74,7 @@ To change external modules (to add a new module or bump the version), update the
 
 Some tests require a Windows VMs or Linux containers. For Linux tests (recommended, if you're not sure) run `docker-compose up -d --build` from `spec/` directory. For windows tests, execute `vagrant up` from the root of the bolt repo to bring these up with the provided Vagrantfile. Any tests that require this are tagged with `:winrm` or `:ssh` in rspec.
```
lucywyman: I think we can change this to `docker-compose -f spec/docker-compose.yml up -d --build` like we have in developer docs and remove the bit about `spec/`. It seems easy to miss.

CONTRIBUTING.md
```diff
@@ -30,6 +30,8 @@ License Agreement before we can accept your pull request: https://cla.puppet.com
 
 ## Installing Bolt
 
+If you are interested in trying bolt out or using in production we recommend installing from a system package detailed in [installing bolt](https://puppet.com/docs/bolt/latest/bolt_installing.html). The following installation instructions are focused on developers who wish to contribute to Bolt.
+
 Depending on your development workflow, you can install Bolt one of three ways:
 
 * From [RubyGems](https://rubygems.org)
```
lucywyman: Should we include `bundle exec rake install` and `install:local`? This seems like it's more about building the gem and installing than just installing the gem, and those both play more nicely with ruby environments and won't confuse users about installing bolt for 'real use' on their system with packages.

CONTRIBUTING.md
lucywyman:
```suggestion
Some tests require a Windows VM or Linux containers. For Linux tests (recommended, if you're not sure) run 
```
docker-compose -f spec/docker-compose.yml up -d --build
``
For windows tests, execute 
```
vagrant up
``` 
from the root of the bolt repo to bring these up with the provided Vagrantfile. Any tests that require this are tagged with `:winrm` or `:ssh` in rspec.
```

Comment (lucywyman) [reply CONTRIBUTING.md]: Hmm I don't think having `\`\`\`` in a suggestion parses correctly...

**Comment by lucywyman** [comment]
I think it'd be helpful to include a link to our community slack towards the top of the contributing doc (technically that's another ticket BOLT-1133 but while we're here?), and probably a link to the docs as well? Maybe a 'Contact' or 'Resources' section at the top?

File: CONTRIBUTING.md (hunk 2 above)
Comment (donoghuc) [reply CONTRIBUTING.md]: I dont really understand how to document your suggestion. I have used a workflow like `gem build bolt` and then `gem install bolt-[VERSION].gem`  are those rake tasks `install` and `install::local` built ins? They don't appear to be defined in Rakefile. Is that in our repo somewhere?

Comment (lucywyman) [reply CONTRIBUTING.md]: Are they not there for you? If I do `bundle exec rake -T` from the root of bolt I get:
```
...
rake install                      # Build and install bolt-1.13.0.gem int...
rake install:local                # Build and install bolt-1.13.0.gem int...
...
```
Those tasks will build the gem and either install to the system gem environment, or the local env (so `bundle exec bolt` will be the built version).

Comment (donoghuc) [reply CONTRIBUTING.md]: So they appear to be part of bundler https://github.com/bundler/bundler/blob/bca8a4bada7bbedab36bb04d6bdce19630d1b873/lib/bundler/gem_helper.rb#L50-L53
Will update that section. Thanks.

File: CONTRIBUTING.md [hunk 3]
```diff
@@ -50,11 +56,15 @@ To use [Bundler](https://bundler.io), add this to your Gemfile:
 
     gem 'bolt'
 
+Alternatively there are rake tasks defined in bunlder that allow you to install with
```
lucywyman:
```suggestion
Alternatively there are rake tasks defined in bundler that allow you to install your local bolt with any changes you made for development and testing.
```

PR #887, opened 2019-03-01
(BOLT-1155) Update wrapper script used for run-as over ssh

donoghuc:
The initial implementation of the `interpreters` configuration option for BOLT-146 failed to handle the case where a task both expected parameters to be passed over STDIN and be run with privilage escalation configured with the `run-as` option. Specifically the wrapper script that is used to invoke the task and pass the parameters needed to be updated to include the interpreter. This commit adds and update to the wrapper script to respect the interpreter while using the `run-as` feature.

### DIFF
--- a/lib/bolt/transport/ssh.rb
+++ b/lib/bolt/transport/ssh.rb
@@ -142,6 +142,7 @@ module Bolt
             stdin, output = nil
             command = []
             execute_options = {}
+            execute_options[:interpreter] = select_interpreter(executable, target.options['interpreters'])
 
             conn.with_remote_tempdir do |dir|
               if extra_files.empty?
@@ -167,7 +168,9 @@ module Bolt
               end
 
               if conn.run_as && stdin
-                wrapper = make_wrapper_stringio(remote_task_path, stdin)
+                # Inject interpreter in to wrapper script and remove from execute options
+                wrapper = make_wrapper_stringio(remote_task_path, stdin, execute_options[:interpreter])
+                execute_options.delete(:interpreter)
                 remote_wrapper_path = conn.write_remote_executable(dir, wrapper, 'wrapper.sh')
                 command << remote_wrapper_path
               else
@@ -177,7 +180,6 @@ module Bolt
               dir.chown(conn.run_as)
 
               execute_options[:sudoable] = true if conn.run_as
-              execute_options[:interpreter] = select_interpreter(executable, target.options['interpreters'])
               output = conn.execute(command, execute_options)
             end
             Bolt::Result.for_task(target, output.stdout.string,
@@ -187,13 +189,22 @@ module Bolt
         end
       end
 
-      def make_wrapper_stringio(task_path, stdin)
-        StringIO.new(<<-SCRIPT)
+      def make_wrapper_stringio(task_path, stdin, interpreter = nil)
+        if interpreter
+          StringIO.new(<<-SCRIPT)
+#!/bin/sh
+'#{interpreter}' '#{task_path}' <<'EOF'
+#{stdin}
+EOF
+SCRIPT
+        else
+          StringIO.new(<<-SCRIPT)
 #!/bin/sh
 '#{task_path}' <<'EOF'
 #{stdin}
 EOF
 SCRIPT
+        end
       end
 
       def connected?(target)
--- a/spec/integration/ssh_spec.rb
+++ b/spec/integration/ssh_spec.rb
@@ -137,6 +137,17 @@ describe "when runnning over the ssh transport", ssh: true do
       end
     end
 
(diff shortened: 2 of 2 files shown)


donoghuc:
Tested manually with following:
### inventory.yaml
```yaml
nodes:
  - name: 0.0.0.0:20022
    alias: override-ssh
    config:
      transport: ssh
      ssh:
        user: root
        password: root
        interpreters:
          py: /usr/bin/python3
```
### interperter.py
```python
#!/foo

import os
import fileinput
import json
import pwd

for line in fileinput.input():
    input_data = json.loads(line)
    if input_data['message']:
    	print(json.dumps({'env': os.environ['PT_message'], 'stdin': input_data['message']}))


print(os.getuid())
print(pwd.getpwuid(os.getuid()))
```
## run-as NOT set
```
cas@cas-ThinkPad-T460p:~/working_dir/bolt$ bolt task run test::interpreter message=hi -n override-ssh
Started on 0.0.0.0...
Finished on 0.0.0.0:
  {"env": "hi", "stdin": "hi"}
  0
  pwd.struct_passwd(pw_name='root', pw_passwd='x', pw_uid=0, pw_gid=0, pw_gecos='root', pw_dir='/root', pw_shell='/bin/bash')
  {
  }
Successful on 1 node: 0.0.0.0:20022
Ran on 1 node in 0.49 seconds
```
### run-as set
```
cas@cas-ThinkPad-T460p:~/working_dir/bolt$ bolt task run test::interpreter message=hi -n override-ssh --run-as bolt --sudo-password bolt
Started on 0.0.0.0...
Finished on 0.0.0.0:
  {"env": "hi", "stdin": "hi"}
  1000
  pwd.struct_passwd(pw_name='bolt', pw_passwd='x', pw_uid=1000, pw_gid=1000, pw_gecos='', pw_dir='/home/bolt', pw_shell='')
  {
  }
Successful on 1 node: 0.0.0.0:20022
Ran on 1 node in 0.80 seconds
```
I also tried escalation path vice versa (escalating from bolt to root).

lib/bolt/transport/ssh.rb
```diff
@@ -167,8 +168,9 @@ def run_task(target, task, arguments, options = {})
               end
 
               if conn.run_as && stdin
-                wrapper = make_wrapper_stringio(remote_task_path, stdin)
+                wrapper = make_wrapper_stringio(remote_task_path, stdin, execute_options[:interpreter])
                 remote_wrapper_path = conn.write_remote_executable(dir, wrapper, 'wrapper.sh')
+                execute_options.delete(:interpreter)
```
nicklewis: A comment would be helpful here, explaining that we don't need to use the interpreter to run the command since we've put it inside the script.

lib/bolt/transport/ssh.rb
```diff
@@ -187,10 +189,15 @@ def run_task(target, task, arguments, options = {})
         end
       end
 
-      def make_wrapper_stringio(task_path, stdin)
+      def make_wrapper_stringio(task_path, stdin, interpreter = nil)
         StringIO.new(<<-SCRIPT)
 #!/bin/sh
-'#{task_path}' <<'EOF'
+if test -z '#{interpreter}'; then
```
nicklewis: It would probably be safer to put this conditional outside the wrapper script and just generate two different possible scripts depending on whether interpreter was passed in.

lib/bolt/transport/ssh.rb
```diff
@@ -187,10 +189,15 @@ def run_task(target, task, arguments, options = {})
...
+  set -- '#{task_path}'
```
nicklewis: What does `set --` do here?

lib/bolt/transport/ssh.rb
donoghuc: Updated to implement the logic in ruby and simplified the wrapper script. Below is the previously proposed wrapper. 
```
      def make_wrapper_stringio(task_path, stdin, interpreter = nil)
        StringIO.new(<<-SCRIPT)
#!/bin/sh
if test -z '#{interpreter}'; then
  set -- '#{task_path}'
else
  set -- '#{interpreter}' '#{task_path}'
fi
"$@" <<'EOF'
#{stdin}
EOF
SCRIPT
      end
```

lib/bolt/transport/ssh.rb
donoghuc: `set --` would unset any positional parameters given to the script, then on the same line replaces them with value for `interpreter` and `task_path`. This allows an array like behavior when executing with `"$@"`. I think that as you mentioned before keeping this wrapper script as simple and straight forward will help keep in maintainable and easy to debug. Updated to simplify.

PR #888, opened 2019-03-01
(BOLT-1085) Document how nil values are passed to tasks

donoghuc:
This commit documents how parameter that have the value `nil` are passed to tasks that use the `environment` `input-method` as the string value `null`.

### DIFF
--- a/pre-docs/writing_plans.md
+++ b/pre-docs/writing_plans.md
@@ -78,6 +78,43 @@ bolt plan run mymodule::myplan --modulepath ./PATH/TO/MODULES load_balancer=lb.m
 
 **Related information**  
 
+It is important to consider that parameters that are passed to the `run_*` plan functions are serialized to JSON. In order to illustrate this consider the following plan. 
+```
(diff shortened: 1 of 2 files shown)


pre-docs/writing_plans.md
```diff
@@ -78,6 +78,53 @@ bolt plan run mymodule::myplan --modulepath ./PATH/TO/MODULES load_balancer=lb.m
...
+```puppet
+plan test::parameter_passing (
+  TargetSpec $nodes,
+  Optional[String[1]] $example_nul = undef,
+) {
+  return run_task('test::demo_undef_ruby', $nodes, example_nul => $example_nul)
+}
+```
+Note that the default value of `$example_nul` is `undef`. The plan calls the `test::demo_undef_ruby` with the `example_nul` parameter. The implementation of the `demo_undef_ruby.rb` task is as follows:
+```ruby
```
adreyer: I think I bash example would be clearer.

PR #889, opened 2019-03-06
(BOLT-1076) Update to latest orchestrator-client

donoghuc:
Update gemspec to include orchestrator_client gem with persistent http connection and better error message for malformed token file support.

### DIFF
--- a/bolt.gemspec
+++ b/bolt.gemspec
@@ -41,7 +41,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency "minitar", "~> 0.6"
   spec.add_dependency "net-scp", "~> 1.2"
   spec.add_dependency "net-ssh", ">= 4.0"
-  spec.add_dependency "orchestrator_client", "~> 0.3.1"
+  spec.add_dependency "orchestrator_client", "~> 0.4.1"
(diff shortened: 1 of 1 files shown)


donoghuc:
I think that the failure in Travis shows that we should try to handle the malformed `service-url` error better. 
As is currently implemented the following config:
```
pcp:
  service-url: "foo"
  cacert: /tmp/ca.pem
  task-environment: production
```
produces the error message below:
```
cas@cas-ThinkPad-T460p:~/working_dir/bolt$ bolt command run whoami -n k4ix901u3kj3iom.delivery.puppetlabs.net --transport pcp
Started on k4ix901u3kj3iom.delivery.puppetlabs.net...
Failed on k4ix901u3kj3iom.delivery.puppetlabs.net:
  bad URI(is not URI?): ://
Failed on 1 node: k4ix901u3kj3iom.delivery.puppetlabs.net
Ran on 1 node in 0.00 seconds
```

donoghuc:
The appveyor issue shows a bigger problem. If I build for windows in vanagon I get the same problem. 
There is a patch that claims to fix the windows issue: https://github.com/drbrain/net-http-persistent/pull/90/files
The last commit to that repo was 2016.

PR #892, opened 2019-03-07
(discuss) experiment using fork instead of capture for compiling

adreyer:
Initializing puppet for catalog compiles is very expensive. By calling
fork instead of capture we can still safely compile catalogs in their
own threads without having to reload puppet code. This also means we
won't have to serialize and deserialize the inventory. Unfortunately
fork is not available on windows so we cannot rely on this
functionality.

This doesn't currently work with the `bolt apply` commandline tool tests since the raw include statements don't work. It seems like the fork should be loading the code when parsing the AST but it isn't, I suspect there is some cache in puppet we need clear or it's an issue with a cache and our ad hoc environments.

### DIFF
--- a/lib/bolt/applicator.rb
+++ b/lib/bolt/applicator.rb
@@ -66,6 +66,54 @@ module Bolt
       end
     end
 
+    def fork_compile(catalog_input)
+      out_r, out_w = IO.pipe
+      err_r, err_w = IO.pipe
+
+      pid = fork do
+        begin
+          out_r.close
+          err_r.close
+          $stdout = out_w
+          $stderr = err_w
+
+          # This should be preloaded in most cases.
+          require 'bolt/catalog'
+
+          catalog_input = JSON.parse(catalog_input.to_json)
+          catalog = Bolt::Catalog.new.from_request(catalog_input)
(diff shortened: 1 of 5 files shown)


nicklewis:
This hangs if the catalog output is too large (probably >64k) because the pipe buffer fills up.

https://github.com/puppetlabs/puppet/blob/6b1a1307a0d23a9e24f727c476d4c738b54052ef/lib/puppet/util/execution.rb#L234-L255

lib/bolt/applicator.rb
```diff
@@ -66,6 +66,54 @@ def query_resources_task
...
+          Kernel.exit!(1)
+        end
+      end
+
+      _, status = Process.wait2(pid)
+
+      out_w.close
+      err_w.close
+
+      return [out_r.read, err_r.read, status]
```
Iristyle: This can deadlock if either `stdout` or `stderr` fills its buffer.

adreyer: I think the puppet solution nick points to above may be better. I think `Thread.new { out.read }` may not exit properly in out doesn't get an eof.

PR #894, opened 2019-03-08
(WIP)(BOLT-1052) Add run-as option to local transport

lucywyman:
This adds support for `run-as`, `sudo-password`, and `run-as-command` to the local transport. These can be configured in the config file, by command line options, or in the inventory file, and should behave exactly like the same options for SSH.


lucywyman:
I spent some time refactoring this over the weekend and want to finish that up before review. Closing this for now!

PR #895, opened 2019-03-08
(BOLT-1152) Run YAML plans

nicklewis:
This implements support for running YAML plans containing task and command
steps only. Every other kind of step is unsupported.

Variable interpolation is supported for targets and for parameter values,
with only exact variable matches (like `$foo`) being supported. Any other
value will be treated as a literal.

### DIFF
--- a/spec/bolt/pal/yaml_plan_evaluator_spec.rb
+++ b/spec/bolt/pal/yaml_plan_evaluator_spec.rb
@@ -1,0 +1,352 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+require 'bolt/pal'
+require 'bolt/pal/yaml_plan_evaluator'
+
+describe Bolt::PAL::YamlPlanEvaluator do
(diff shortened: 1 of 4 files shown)


nicklewis:
This is built on / depends on #884

spec/bolt/pal/yaml_plan_evaluator_spec.rb
```diff
@@ -0,0 +1,347 @@
...
+    end
+
+    it 'returns a puppet function wrapper' do
+      plan_body = '{}'
+      plan = described_class.create(loader, plan_name, 'test.yaml', plan_body)
+      expect(plan).to be_a(Puppet::Functions::Function)
+    end
+  end
+
+  describe "evaluating a plan" do
```
adreyer: I'd like to see a test of variable extraction at this level in addition to the run step level.

nicklewis:
I updated the test that runs a plan to use variable interpolation. I also added an integration test that runs a YAML plan via `run_cli`.

PR #897, opened 2019-03-13
 (BOLT-1052) Add run-as option to local transport 

lucywyman:
This adds support for `run-as`, `sudo-password`, and `run-as-command` to the local transport. These can be configured in the config file, by command line options, or in the inventory file, and should behave exactly like the same options for SSH.

I ended up keeping the SSH and local functionality mostly separate in order to take advantage of `popen3` and other system libraries for the local transport. I also pulled most of SSH's `sudo` tests into the shared examples so that local can use them.

### DIFF
--- a/lib/bolt/transport/local.rb
+++ b/lib/bolt/transport/local.rb
@@ -1,17 +1,10 @@
 # frozen_string_literal: true
 
-require 'json'
-require 'fileutils'
-require 'tmpdir'
-require 'bolt/transport/base'
-require 'bolt/transport/powershell'
-require 'bolt/util'
-
 module Bolt
   module Transport
-    class Local < Base
+    class Local < Sudoable
       def self.options
-        %w[tmpdir interpreters]
+        %w[tmpdir interpreters sudo-password run-as run-as-command]
       end
 
       def self.default_options
@@ -21,176 +14,17 @@ module Bolt
       end
 
       def provided_features
-        if Bolt::Util.windows?
-          ['powershell']
-        else
-          ['shell']
-        end
+        ['shell']
       end
 
-      def default_input_method(executable)
-        input_method ||= Powershell.powershell_file?(executable) ? 'powershell' : 'both'
-        input_method
+      def self.validate(options)
+        logger = Logging.logger[self]
+        validate_sudo_options(options, logger)
       end
 
-      def self.validate(_options); end
-
-      def initialize
-        super
-        @conn = Shell.new
-      end
-
-      def in_tmpdir(base)
-        args = base ? [nil, base] : []
-        dir = begin
-                Dir.mktmpdir(*args)
-              rescue StandardError => e
-                raise Bolt::Node::FileError.new("Could not make tempdir: #{e.message}", 'TEMPDIR_ERROR')
-              end
-
-        yield dir
-      ensure
-        FileUtils.remove_entry dir if dir
-      end
-      private :in_tmpdir
-
-      def copy_file(source, destination)
-        FileUtils.cp_r(source, destination, remove_destination: true)
-      rescue StandardError => e
-        raise Bolt::Node::FileError.new(e.message, 'WRITE_ERROR')
-      end
-
-      def with_tmpscript(script, base)
-        in_tmpdir(base) do |dir|
-          dest = File.join(dir, File.basename(script))
-          copy_file(script, dest)
-          File.chmod(0o750, dest)
-          yield dest, dir
-        end
-      end
-      private :with_tmpscript
-
-      def upload(target, source, destination, _options = {})
-        copy_file(source, destination)
-        Bolt::Result.for_upload(target, source, destination)
-      end
-
-      def run_command(target, command, _options = {})
-        in_tmpdir(target.options['tmpdir']) do |dir|
-          output = @conn.execute(command, dir: dir)
-          Bolt::Result.for_command(target,
-                                   output.stdout.string,
-                                   output.stderr.string,
-                                   output.exit_code,
-                                   'command', command)
-        end
-      end
-
-      def run_script(target, script, arguments, _options = {})
-        with_tmpscript(File.absolute_path(script), target.options['tmpdir']) do |file, dir|
-          logger.debug "Running '#{file}' with #{arguments}"
-
-          # unpack any Sensitive data AFTER we log
-          arguments = unwrap_sensitive_args(arguments)
-          if Bolt::Util.windows?
-            if Powershell.powershell_file?(file)
-              command = Powershell.run_script(arguments, file)
-              output = @conn.execute(command, dir: dir, env: "powershell.exe")
-            else
-              path, args = *Powershell.process_from_extension(file)
-              args += Powershell.escape_arguments(arguments)
-              command = args.unshift(path).join(' ')
-              output = @conn.execute(command, dir: dir)
-            end
-          else
-            if arguments.empty?
-              # We will always provide separated arguments, so work-around Open3's handling of a single
-              # argument as the entire command string for script paths containing spaces.
-              arguments = ['']
-            end
-            output = @conn.execute(file, *arguments, dir: dir)
-          end
-          Bolt::Result.for_command(target,
-                                   output.stdout.string,
-                                   output.stderr.string,
-                                   output.exit_code,
-                                   'script', script)
-        end
-      end
-
-      def run_task(target, task, arguments, _options = {})
-        implementation = select_implementation(target, task)
-        executable = implementation['path']
-        input_method = implementation['input_method']
-        extra_files = implementation['files']
-
-        in_tmpdir(target.options['tmpdir']) do |dir|
-          if extra_files.empty?
-            script = File.join(dir, File.basename(executable))
-          else
-            arguments['_installdir'] = dir
-            script_dest = File.join(dir, task.tasks_dir)
-            FileUtils.mkdir_p([script_dest] + extra_files.map { |file| File.join(dir, File.dirname(file['name'])) })
-
-            script = File.join(script_dest, File.basename(executable))
-            extra_files.each do |file|
-              dest = File.join(dir, file['name'])
-              copy_file(file['path'], dest)
-              File.chmod(0o750, dest)
-            end
-          end
-
-          copy_file(executable, script)
-          File.chmod(0o750, script)
-
-          interpreter = select_interpreter(script, target.options['interpreters'])
-          interpreter_debug = interpreter ? " using '#{interpreter}' interpreter" : nil
-          # log the arguments with sensitive data redacted, do NOT log unwrapped_arguments
-          logger.debug("Running '#{script}' with #{arguments}#{interpreter_debug}")
-          unwrapped_arguments = unwrap_sensitive_args(arguments)
-
-          stdin = STDIN_METHODS.include?(input_method) ? JSON.dump(unwrapped_arguments) : nil
-
-          if Bolt::Util.windows?
-            # WINDOWS
-            if ENVIRONMENT_METHODS.include?(input_method)
-              environment_params = envify_params(unwrapped_arguments).each_with_object([]) do |(arg, val), list|
-                list << Powershell.set_env(arg, val)
-              end
-              environment_params = environment_params.join("\n") + "\n"
-            else
-              environment_params = ""
-            end
-
-            if Powershell.powershell_file?(script) && stdin.nil?
-              command = Powershell.run_ps_task(arguments, script, input_method)
-              command = environment_params + Powershell.shell_init + command
-              interpreter ||= 'powershell.exe'
-              output =
-                if input_method == 'powershell'
-                  @conn.execute(command, dir: dir, interpreter: interpreter)
-                else
-                  @conn.execute(command, dir: dir, stdin: stdin, interpreter: interpreter)
-                end
-            end
-            unless output
-              if interpreter
-                env = ENVIRONMENT_METHODS.include?(input_method) ? envify_params(unwrapped_arguments) : nil
-                output = @conn.execute(script, stdin: stdin, env: env, dir: dir, interpreter: interpreter)
-              else
-                path, args = *Powershell.process_from_extension(script)
-                command = args.unshift(path).join(' ')
-                command = environment_params + Powershell.shell_init + command
-                output = @conn.execute(command, dir: dir, stdin: stdin, interpreter: 'powershell.exe')
-              end
-            end
-          else
-            # POSIX
-            env = ENVIRONMENT_METHODS.include?(input_method) ? envify_params(unwrapped_arguments) : nil
-            output = @conn.execute(script, stdin: stdin, env: env, dir: dir, interpreter: interpreter)
-          end
-          Bolt::Result.for_task(target, output.stdout.string, output.stderr.string, output.exit_code, task.name)
-        end
+      def with_connection(target, *_args)
+        conn = Shell.new(target)
+        yield conn
       end
 
       def connected?(_targets)
--- a/lib/bolt/transport/local/shell.rb
+++ b/lib/bolt/transport/local/shell.rb
@@ -1,26 +1,206 @@
 # frozen_string_literal: true
 
 require 'open3'
+require 'fileutils'
 require 'bolt/node/output'
+require 'bolt/util'
 
 module Bolt
   module Transport
-    class Local
-      class Shell
-        def execute(*command, options)
-          command.unshift(options[:interpreter]) if options[:interpreter]
-          command = [options[:env]] + command if options[:env]
+    class Local < Sudoable
+      class Shell < Sudoable::Connection
+        attr_accessor :user, :logger, :target
+        attr_writer :run_as
 
-          if options[:stdin]
-            stdout, stderr, rc = Open3.capture3(*command, stdin_data: options[:stdin], chdir: options[:dir])
+        CHUNK_SIZE = 4096
+
+        def initialize(target)
+          @target = target
+          # The familiar problem: Etc.getlogin is broken on osx
+          @user = ENV['USER'] || Etc.getlogin
+          @run_as = target.options['run-as']
+          @logger = Logging.logger[self]
+        end
+
+        # If prompted for sudo password, send password to stdin and return an
+        # empty string. Otherwise, check for sudo errors and raise Bolt error.
+        # If error is not sudo-related, return the stderr string to be added to
+        # node output
+        def handle_sudo(stdin, err, pid)
+          if err.include?(Sudoable.sudo_prompt)
+            # A wild sudo prompt has appeared!
+            if @target.options['sudo-password']
+              # Hopefully no one's sudo-password is > 64kb
+              stdin.write("#{@target.options['sudo-password']}\n")
+              ''
+            else
+              raise Bolt::Node::EscalateError.new(
+                "Sudo password for user #{@user} was not provided for localhost",
+                'NO_PASSWORD'
+              )
+            end
           else
-            stdout, stderr, rc = Open3.capture3(*command, chdir: options[:dir])
+            handle_sudo_errors(err, pid)
+          end
+        end
+
+        def handle_sudo_errors(err, pid)
+          if err =~ /^#{@user} is not in the sudoers file\./
+            @logger.debug { err }
+            raise Bolt::Node::EscalateError.new(
+              "User #{@user} does not have sudo permission on localhost",
+              'SUDO_DENIED'
+            )
+          elsif err =~ /^Sorry, try again\./
+            @logger.debug { err }
+            # CODEREVIEW can we kill a sudo process without sudo password?
+            Process.kill('TERM', pid)
+            raise Bolt::Node::EscalateError.new(
+              "Sudo password for user #{@user} not recognized on localhost",
+              'BAD_PASSWORD'
+            )
+          else
+            # No need to raise an error - just return the string
+            err
+          end
+        end
+
+        def copy_file(source, dest)
+          if source.is_a?(StringIO)
+            File.open("tempfile", "w") { |f| f.write(source.read) }
+            execute(['mv', 'tempfile', dest])
+          else
+            # Mimic the behavior of `cp --remove-destination`
+            # since the flag isn't supported on MacOS
+            result = execute(['rm', '-rf', dest])
+            if result.exit_code != 0
+              message = "Could not remove existing file #{dest}: #{result.stderr.string}"
+              raise Bolt::Node::FileError.new(message, 'REMOVE_ERROR')
+            end
+
+            result = execute(['cp', '-r', source, dest])
+            if result.exit_code != 0
+              message = "Could not copy file to #{dest}: #{result.stderr.string}"
+              raise Bolt::Node::FileError.new(message, 'COPY_ERROR')
+            end
+          end
+        end
+
+        def with_tmpscript(script)
+          with_tempdir do |dir|
+            dest = File.join(dir.to_s, File.basename(script))
+            copy_file(script, dest)
+            yield dest, dir
+          end
+        end
+
+        # See if there's a sudo prompt in the output
+        # If not, return the output
+        def check_sudo(out, inp, pid)
+          buffer = out.readpartial(CHUNK_SIZE)
+          # Split on newlines, including the newline
+          lines = buffer.split(/(?<=[\n])/)
+          # handle_sudo will return the line if it is not a sudo prompt or error
+          lines.map! { |line| handle_sudo(inp, line, pid) }
+          lines.join("")
+        end
+
+        def execute(command, sudoable: true, **options)
+          run_as = options[:run_as] || self.run_as
+          escalate = sudoable && run_as && @user != run_as
+          use_sudo = escalate && @target.options['run-as-command'].nil?
+
+          if options[:interpreter]
+            if command.is_a?(Array)
+              command.unshift(options[:interpreter])
+            else
+              command = [options[:interpreter], command]
+            end
           end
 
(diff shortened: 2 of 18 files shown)


lib/bolt/transport/helpers.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+require 'shellwords'
+
+def validate_sudo_options(options, logger)
```
adreyer: These should be in a module or class

lucywyman:
M'k - the tests that are failing in Travis are only failing in Travis, and it seems to be because the sudo prompt isn't getting printed to STDERR like we expect. I'm not sure how the tests with a correct password are passing, since we check for the prompt there as well, but I'm going to have to do some digging into why the second messages are getting lost

lib/bolt/transport/local.rb
```diff
@@ -33,155 +30,140 @@ def default_input_method(executable)
...
-          else
-            if arguments.empty?
-              # We will always provide separated arguments, so work-around Open3's handling of a single
-              # argument as the entire command string for script paths containing spaces.
-              arguments = ['']
+      def run_script(target, script, arguments, options = {})
+        logger.debug "Running '#{script}' with #{arguments}"
+        if Bolt::Util.windows?
+          with_connection(target) do |conn|
+            conn.running_as(options['_run_as']) do
```
adreyer: I wouldn't expect this to work on windows. We probably need to make sure there is clean error here.

lucywyman: Should we error if someone runs on winrm with `run_as`? Or just work around it and pretend it's not set?

adreyer: I think erroring is best. It may cause problems for users running `--run-as` but I think we can wait for wait for users to complain.

lib/bolt/transport/sudoable/connection_helper.rb
```diff
@@ -0,0 +1,52 @@
...
+          if result.exit_code != 0
+            raise Bolt::Node::FileError.new("Could not make tempdir: #{result.stderr.string}", 'TEMPDIR_ERROR')
+          end
+          path = tmppath || result.stdout.string.chomp
+          Sudoable::FileCommands.new(self, path)
+        end
+
+        # A helper to create and delete a tempdir on the remote system. Yields the
+        # directory name.
+        def with_tempdir
```
donoghuc: This delete does not get run with the elevated privileges. I have not had a chance to dig in, but it also seems random if the error is even surfaced for `script run`. Sometime I get 
```
cas@cas-ThinkPad-T460p:~/working_dir/bolt/Boltdir$ bolt script run test.sh -n localhost --run-as root --sudo-password $SECRET
Started on localhost...
Failed to clean up tempdir '/tmp/57a91557-a963-4787-92b1-bd8e13d451cf': rm: cannot remove '/tmp/57a91557-a963-4787-92b1-bd8e13d451cf': Operation not permitted

Finished on localhost:
  STDOUT:
    hi
Ran on 1 node in 0.06 seconds
```
and others the message does not surface. For `run task` it always fails
```
cas@cas-ThinkPad-T460p:~/working_dir/bolt/Boltdir$ bolt task run sample::noop -n localhost --run-as root --sudo-password $SECRET
Started on localhost...
Failed to clean up tempdir '/tmp/f294dbd1-3df6-453b-a772-f620b59faad1': rm: cannot remove '/tmp/f294dbd1-3df6-453b-a772-f620b59faad1': Operation not permitted

Failed on localhost:
  Broken pipe
Failed on 1 node: localhost
Ran on 1 node in 0.11 seconds
```

lib/bolt/transport/sudoable/file_commands.rb
```diff
@@ -0,0 +1,68 @@
...
+          if result.exit_code != 0
+            message = "Could not change owner of '#{@path}' to #{owner}: #{result.stderr.string}"
+            raise Bolt::Node::FileError.new(message, 'CHOWN_ERROR')
+          end
+
+          # File ownership successfully changed, record the new owner.
+          @owner = owner
+        end
+
+        def copy_file(destination)
```
donoghuc: I think this should be removed/refactored? It is kind of confusing having this be used by `Bolt::Transport::Local::Shell#copy_file`. It does not work for macos 
```
u1u186j1hflhufg:bolt root# bundle exec bolt task run facts -n localhost
Started on localhost...
Failed on localhost:
  Could not copy file to /tmp/cc0a2936-9081-408b-90d5-7deee60410fc/bash.sh: cp: illegal option -- -
  usage: cp [-R [-H | -L | -P]] [-fi | -n] [-apvXc] source_file target_file
         cp [-R [-H | -L | -P]] [-fi | -n] [-apvXc] source_file ... target_directory
Failed on 1 node: localhost
Ran on 1 node in 0.12 seconds
u1u186j1hflhufg:bolt root# uname -a
Darwin u1u186j1hflhufg 17.3.0 Darwin Kernel Version 17.3.0: Thu Nov  9 18:09:22 PST 2017; root:xnu-4570.31.3~1/RELEASE_X86_64 x86_64
```

Also is there a different approach for handling the `write_remote_file` and `copy_file` using `alias`? It seems kind of awkward (but maybe its just a common ruby thing that i'm just not used to seeing).

lib/bolt/transport/local/shell.rb
```diff
@@ -1,13 +1,94 @@
...
+            begin
+              FileUtils.cp_r(source, dest, remove_destination: true)
+            rescue StandardError => e
+              raise Bolt::Node::FileError.new(e.message, 'WRITE_ERROR')
+            end
+          elsif source.is_a?(StringIO)
+            File.open("tempfile", "w") { |f| f.write(source.read) }
+            execute(['mv', 'tempfile', dest])
+          else
+            result = execute(['cp', '-r', '--remove-destination', source, dest])
```
donoghuc: Still need an alternative to `--remove-destination`. 
```
b6jdy4w9ksktf9q:bolt osx$ bundle exec bolt task run facts -n localhost
Started on localhost...
Failed on localhost:
  Could not copy file to /tmp/401cbccd-b6f9-4482-a021-e02997e6c386/bash.sh: cp: illegal option -- -
  usage: cp [-R [-H | -L | -P]] [-fi | -n] [-apvXc] source_file target_file
         cp [-R [-H | -L | -P]] [-fi | -n] [-apvXc] source_file ... target_directory
Failed on 1 node: localhost
Ran on 1 node in 0.13 seconds
b6jdy4w9ksktf9q:bolt osx$ git log --oneline
903aeea (HEAD -> BOLT-1052, origin/BOLT-1052) (BOLT-1052) Add run-as option to local transport
```

lucywyman:
@adreyer @donoghuc  Tested and ready for another look over!

lucywyman:
On Monday ;)

donoghuc:
Still having issues with osx 10.13
```
b6jdy4w9ksktf9q:bolt osx$ bundle exec bolt task run facts -n localhost --run-as root --sudo-password *****
Started on localhost...
Failed to clean up tempdir '/tmp/005952d1-b4eb-4087-a358-2bf5f7820ca5': rm: /tmp/005952d1-b4eb-4087-a358-2bf5f7820ca5: Permission denied

Failed on localhost:
  Could not copy file to /tmp/005952d1-b4eb-4087-a358-2bf5f7820ca5/bash.sh:
Failed on 1 node: localhost
Ran on 1 node in 0.59 seconds
b6jdy4w9ksktf9q:bolt osx$ bundle exec bolt task run facts -n localhost
Started on localhost...
Failed on localhost:
  Could not copy file to /tmp/6389d7a6-6919-4cec-9dd9-155add8d4e1e/bash.sh: cp: illegal option -- -
  usage: cp [-R [-H | -L | -P]] [-fi | -n] [-apvXc] source_file target_file
         cp [-R [-H | -L | -P]] [-fi | -n] [-apvXc] source_file ... target_directory
Failed on 1 node: localhost
Ran on 1 node in 0.15 seconds
b6jdy4w9ksktf9q:bolt osx$ git log --oneline
ab0e39f (HEAD, origin/BOLT-1052) (BOLT-1052) Add run-as option to local transport
```
feel free to test on that node.

lucywyman:
@donoghuc :woman_facepalming:  Didn't include the change to remove `--remove-destination`. Seems to be working now:
```
b6jdy4w9ksktf9q:bolt osx$ bundle exec bolt task run facts -n localhost
Started on localhost...
Finished on localhost:
  {
    "os": {
      "name": "Darwin",
      "release": {
        "full": "17.3.0",
        "major": "17",
        "minor": "3"
      },
      "family": "Darwin"
    }
  }
Ran on 1 node in 0.17 seconds
```

donoghuc:
Did we decide not to error/warn when user tries to use run-as over local transport on windows os? 
```
PS C:\Users\Administrator\bolt> bundle exec bolt command run "echo hi" -n localhost --run-as foo --sudo-password bar
Started on localhost...
Finished on localhost:
  STDOUT:
    hi
Ran on 1 node in 0.02 seconds
```

lucywyman:
After rearchitecting there wasn't as convenient of a place to verify that the user hadn't passed in extra flags. I could add some logic for that to `Bolt::Transport::LocalWindows.validate` https://github.com/lucywyman/bolt/blob/BOLT-1052/lib/bolt/transport/local_windows.rb#L34

donoghuc:
I think that previously `Open3.capture3` used in `Bolt::Transport::Local::Shell#execute` would handle large amounts of data from the child process. I think we might need to figure out a similar solution for the new `Open3.popen3` method.

For example the following works for local transport and ssh transport on tip of bolt master but fails with this PR. 
```
b6jdy4w9ksktf9q:bolt osx$ bundle exec bolt command run 'echo `cat /dev/urandom | env LC_CTYPE=C tr -cd 'a-f0-9' | head -c 1000000`' -n localhostStarted on localhost...
Failed on localhost:
  The command failed with exit code 126
Failed on 1 node: localhost
Ran on 1 node in 1.82 seconds
```
The reason I think it is an issue with size is because something like the following works file against this PR
```
b6jdy4w9ksktf9q:bolt osx$ bundle exec bolt command run 'echo `cat /dev/urandom | env LC_CTYPE=C tr -cd 'a-f0-9' | head -c 10`' -n localhost
Started on localhost...
Finished on localhost:
  STDOUT:
    5388a120fa
Ran on 1 node in 0.11 seconds
```

lucywyman:
@casdonoghue I figured out why the tests were commented out. The local transport used to change directories to the temp dir even for running commands, not just put files there. It can't do that now because `popen3` runs as the system user, and so can't accept a `chdir` option for spawning the process to a directory that the user doesn't own (if you're using `run-as`). For SSH it looks like we just don't execute in the tmpdir, only put files there, so I think that's the right behavior and I just need to change the tests.

lucywyman:
@donoghuc @adreyer  Pending travis, this is ready. I was writing without blocking, I think to try to free up stderr for reading the sudo prompt but I didn't actually need the nonblock which was what was causing it to error on large input. Also fixed up the windows validation!

donoghuc:
It is still failing for me on latest commit? I think we will need to implement our own waiting/buffer management like in that blog post you shared. http://coldattic.info/post/63/
```
b6jdy4w9ksktf9q:bolt osx$ git log --oneline
d3d54b6 (HEAD, origin/BOLT-1052) (BOLT-1052) Add run-as option to local transport
b6jdy4w9ksktf9q:bolt osx$ bundle exec bolt command run 'echo `cat /dev/urandom | env LC_CTYPE=C tr -cd 'a-f0-9' | head -c 1000000`' -n localhost
Started on localhost...
Failed on localhost:
  The command failed with exit code 126
Failed on 1 node: localhost
Ran on 1 node in 1.67 seconds
b6jdy4w9ksktf9q:bolt osx$ 
```
What are you using for testing long stdout? I think i recall alex saying something about not using pipes, maybe it is a problem with my test? Though the same command (and even the next power of 10) works on upstream/master.

donoghuc:
Also somehow stderr is not being reported on osx. 
From master
```
b6jdy4w9ksktf9q:bolt osx$ bundle exec bolt command run 'echo `cat /dev/urandom | env LC_CTYPE=C tr -cd 'a-f0-9' | head -c 10 1>&2`' -n localhost --format json
{ "items": [
{"node":"localhost","status":"success","result":{"stdout":"\n","stderr":"3f371c9a41","exit_code":0}}
],
"node_count": 1, "elapsed_time": 0 }
b6jdy4w9ksktf9q:bolt osx$ git log --oneline
91135be (HEAD, upstream/master) Merge pull request #915 from da-ar/FM-7869
```
From 1052
```
b6jdy4w9ksktf9q:bolt osx$ bundle exec bolt command run 'echo `cat /dev/urandom | env LC_CTYPE=C tr -cd 'a-f0-9' | head -c 10 1>&2`' -n localhost --format json
{ "items": [
{"node":"localhost","status":"success","result":{"stdout":"\n","stderr":"","exit_code":0}}
],
"node_count": 1, "elapsed_time": 0 }
d3d54b6 (HEAD -> BOLT-1052, origin/BOLT-1052) (BOLT-1052) Add run-as option to local transport
```

lib/bolt/transport/local/shell.rb
```diff
@@ -1,26 +1,179 @@
...
+              # Read from out and err
+              ready_read&.each do |stream|
+                begin
+                  if use_sudo
+                    buffer = stream.readpartial(chunk_size)
+                    # Split on newlines, including the newline
+                    lines = buffer.split(/(?<=[\n])/)
+                    # handle_sudo will return the line if it is not a sudo prompt or error
+                    lines.map! { |line| handle_sudo(inp, line) }
+                    read_streams[stream] << lines.join("")
```
adreyer: It looks like with the current ssh implementation we don't add the lines if they're part of the sudo prompt

lib/bolt/transport/local/shell.rb
```diff
@@ -1,26 +1,179 @@
...
+                  else
+                    read_streams[stream] << stream.readpartial(chunk_size)
+                  end
+                rescue EOFError
+                  strings = { out => 'stdout', err => 'stderr' }
+                  @logger.debug "Read EOF on #{strings[stream]}"
```
adreyer: I don't think EOF should be logged it's seems like a normal condition.

lib/bolt/transport/local/shell.rb
```diff
@@ -1,26 +1,179 @@
...
+                end
+              end
+
+              if ready_write && index < in_buffer.length
+                to_print = in_buffer[index..index + chunk_size - 1]
```
adreyer: Since we're doing a nonblocking write can we just try to write  the entire string rather then chunking?

lib/bolt/transport/local/shell.rb
```diff
@@ -1,26 +1,179 @@
...
+                write_stream = []
+              end
+            end
+          rescue Errno::ECHILD
+            # Returned from waitpid if there is no child process
+            @logger.debug "No child process to wait for; Continuing"
+          end
+          result_output.stdout << read_streams[out]
+          result_output.stderr << read_streams[err]
+          # If the process exited before we processed input or output, !$ will have the exitstatus
```
adreyer: Is this threadsafe

lib/bolt/transport/local/shell.rb
```diff
@@ -1,26 +1,179 @@
...
+              end
+
+              if index >= in_buffer.length && !write_stream.empty?
+                inp.close
+                write_stream = []
+              end
+            end
+          rescue Errno::ECHILD
+            # Returned from waitpid if there is no child process
+            @logger.debug "No child process to wait for; Continuing"
```
adreyer: I don't think we need to log this if we expect it to happen when the process exits while we're reading. We should verify that is what causes it though.

lib/bolt/transport/local/shell.rb
```diff
@@ -1,26 +1,179 @@
...
+          end
+          result_output.stdout << read_streams[out]
```
adreyer: It looks like the loop that makes sure the streams are empty after the process exits got removed.

lib/bolt/transport/local/shell.rb
```diff
@@ -1,26 +1,179 @@
...
+
+            until Process.waitpid(t.pid, Process::WNOHANG)
+              timeout = 0.1
+              # See if we can read from out or err, or write to in
+              ready_read, ready_write, = select(read_streams.keys, write_stream, nil, timeout)
+
+              # Read from out and err
+              ready_read&.each do |stream|
+                begin
+                  if use_sudo
```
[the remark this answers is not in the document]
adreyer: I guess we do in the preexisting code anyway.

lib/bolt/transport/local/shell.rb
lucywyman: This doesn't - `handle sudo` returns an empty string if it consumes the sudo prompt. https://github.com/puppetlabs/bolt/pull/897/files#diff-a96c17ec64c4c2d8d0e546f3f96feeb5R559 verifies that it doesn't print anything besides expected stderr.

lib/bolt/transport/local/shell.rb
lucywyman: Was trying to think of clever ways around the Exception cop. Should I just disable it ?

lib/bolt/transport/local/shell.rb
lucywyman: I didn't see any issues with data not getting read, including for non-1024 multiple sized strings. I can add it back to be safe though

lib/bolt/transport/local/shell.rb
lucywyman: I had trouble finding more information on what raises this, besides that there's no child process to wait on. It'd be nice to check before the `until`, but `active?` and `status` both showed the thread was still running.

lucywyman:
Ugh, I didn't realize we were going to merge BOLT-1125. Was there an issue with blocking it on this, or branching off this PR?

lib/bolt/transport/local/shell.rb
lucywyman: It appears not? Commenting out the chunking bits causes it to hang on large input.

lib/bolt/transport/sudoable/connection_helper.rb
```diff
@@ -0,0 +1,76 @@
+# frozen_string_literal: true
+
+require 'bolt/transport/sudoable/file_commands'
+
+module Bolt
+  module Transport
+    class Sudoable < Base
+      class ConnectionHelper
```
adreyer: Nitpick I think helpers are more idiomatically mixin modules. This seems like it just a base class so we can just call it Connection

lib/bolt/transport/sudoable/file_commands.rb
```diff
@@ -0,0 +1,59 @@
+# frozen_string_literal: true
+
+module Bolt
+  module Transport
+    class Sudoable < Base
+      class FileCommands
```
adreyer: I think this object is easier to understand if we call it "a Tmpdir" rather then "a FileCommands"

lib/bolt/transport/sudoable.rb
```diff
@@ -0,0 +1,163 @@
...
+                                  output.stderr.string,
+                                  output.exit_code,
+                                  task.name)
+          end
+        end
+      end
+
+      def make_wrapper_stringio(task_path, stdin, interpreter = nil)
+        if interpreter
+          StringIO.new(<<-SCRIPT)
```
adreyer: Since we've switched to requiring ruby 2.3  I think we can rely on the tilde heredoc that is indentable. 

https://infinum.co/the-capsized-eight/multiline-strings-ruby-2-3-0-the-squiggly-heredoc

lib/bolt/transport/ssh/connection.rb
```diff
@@ -308,51 +240,17 @@ def write_remote_file(source, destination)
...
-        end
-
-        def write_remote_executable(dir, file, filename = nil)
-          filename ||= File.basename(file)
-          remote_path = File.join(dir.to_s, filename)
-          write_remote_file(file, remote_path)
-          make_executable(remote_path)
-          remote_path
-        end
+        # We use `copy_file` for compatibility with the local transport in
```
adreyer: I think we should settle on one name. IMO aliases are for supporting public APIs otherwise they just make things more confusing.

lib/bolt/transport/local/shell.rb
```diff
@@ -1,26 +1,185 @@
...
+          read_streams.each do |stream, string|
+            begin
+              loop { string << stream.read_nonblock(chunk_size) }
+            rescue Errno::EAGAIN, EOFError
+            end
+          end
+          result_output.stdout << read_streams[out]
+          result_output.stderr << read_streams[err]
+          # If the process exited before we processed input or output, !$ will have the exitstatus
+          result_output.exit_code = t.value&.exitstatus || status[1].exitstatus
```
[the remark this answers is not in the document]
lucywyman: I think what might be happening is that `t.value` is actually out of scope, not nil? The `Errno:ECHILD` happens when there's no child process to watch. But if I print `t.value.exitstatus` right above that line, it has a value. I added it to the rescue block which seems to work, but that won't always get called and `t.value.exitstatus` won't always be defined before Process.waitpid.

lib/bolt/transport/local/shell.rb
nicklewis: I think this can just be `in_buffer[index, chunk_size]`

lib/bolt/transport/local/shell.rb
```diff
@@ -1,26 +1,191 @@
...
+                    read_streams[stream] << lines.join("")
+                    checked_sudo += 1
+                  else
+                    read_streams[stream] << stream.readpartial(chunk_size)
+                  end
+                rescue EOFError
+                end
+              end
+
+              if ready_write && index < in_buffer.length
```
nicklewis: If there was a read stream ready and the write stream was not ready, `ready_write` will be an empty array, but this condition will be true, causing it to attempt to write when the stream is not ready.

lib/bolt/transport/local/shell.rb
```diff
@@ -1,26 +1,195 @@
...
+        def handle_sudo_errors(err, pid)
+          if err =~ /^#{@user} is not in the sudoers file\./
+            @logger.debug { err }
+            raise Bolt::Node::EscalateError.new(
+              "User #{@user} does not have sudo permission on localhost",
+              'SUDO_DENIED'
+            )
+          elsif err =~ /^Sorry, try again\./
+            @logger.debug { err }
+            Process.kill('TERM', pid)
```
adreyer: This doesn't work the sudo process is owned by root so it can't be killed from here instead it raises `Errno::EPERM`

spec/bolt/transport/local_spec.rb
```diff
@@ -29,10 +34,169 @@
...
+              break
+            end
+          end
+          TASK
+      end
+
+      it "runs with large input and captures large output" do
+        with_task_containing('slow_and_steady', ruby_task, 'stdin', '.rb') do |task|
+          result = runner.run_task(target, task, arguments).value
+          expect(result['input'].bytesize).to eq(file_size * 1024)
```
adreyer: input is nil which causes an error here. Looking at the code that is what I would expect.

lucywyman: I was seeing `input` because we pass params in as JSON but don't parse them in the task, so the return is also json which bolt makes the result object. I'm not sure why you wouldn't expect `input`?

lib/bolt/transport/local/shell.rb
lucywyman: Even if the user hasn't authenticated yet?

This raises the question: I'm not sure (obviously) how to test this, or how to kill a process owned by sudo if the user can't authenticate....

.rspec-sudo
```diff
@@ -0,0 +1,2 @@
+--require spec_helper
+--color
```
donoghuc: Is this tag meant to guard against the case where there is not a `travis:travis` `group:user` the local_spec test expects? An alternate pattern could be to guard with an environment variable. So for example add `filter_run_excluding` to spec_helper [like this](https://github.com/puppetlabs/bolt/blob/3955ff37c7bb24911a1799cbb0469107aabd4f88/spec/spec_helper.rb#L41) then add the environment variable only in .travis [like this](https://github.com/puppetlabs/bolt/blob/3955ff37c7bb24911a1799cbb0469107aabd4f88/appveyor.yml#L104)

spec/bolt/transport/local_spec.rb
adreyer: ah that makes sense. But that won't differentiate from the edge case where stdin gets printed by sudo when the task exits without reading it.

It also looks like this task will fail to print the last chunk if the eof isn't exactly on a boundary.

lib/bolt/transport/local/shell.rb
lucywyman: Still not addressed

spec/bolt/transport/local_spec.rb
lucywyman: Updated to 1011kb, so it's not a multiple of 4kb that get read in the loop

.rspec
```diff
@@ -1,2 +1,3 @@
 --require spec_helper
---color
\ No newline at end of file
+--color
+--tag ~sudo
```
donoghuc: Excluding sudo tag in `.rspec` should not be necessary with the addition `filter_run_excluding` in  `spec_helper`.

lucywyman: Woops good call - fixed!

spec/bolt/transport/local_spec.rb
adreyer: That sounds like the edge case where when a task doesn't consume stdin and is wrapped in sudo it gets printed to stdout by sudo.

PR #898, opened 2019-03-13
(BOLT-1170) Fixing crashes caused by large file uploads during apply() blocks

nmaludy:
### Background

Details are in the JIRA ticket: https://tickets.puppetlabs.com/browse/BOLT-1170

### Potential solution

Instead of base64 encoding the plugins tarball and catalog, i've written these out as temporary files and attached them to the `Bolt::Task` that `apply()` executes on the target hosts. Transports then take care of uploading all `files`  (extra files) associated with a task, putting them in the `_installdir` automatically for us. Finally `libexec/apply_catalog.rb` reads the plugins tarball and catalog from the filesystem at `_installdir` instead of from STDIN arguments.

### Extent of my testing 

I've tested this out practically using the same scenario where it was failing previously and no issues occurred on Linux (SSH transport) or Windows hosts (WinRM transport). Actually the Windows hosts executed faster because the WinRM file upload was _much_ quicker than having the data base64 encoded (actually makes it larger) and added onto the command line.

### TODO
* [ ] Check if this works with PXP / Orchestrator
* [ ] Check if we should be using this "out-of-band" transfer for the two other places where we transfer plugin tarballs: (
 - [ ] bolt-modules/boltlib/lib/puppet/functions/apply_prep.rb
 - [ ] bolt-modules/boltlib/lib/puppet/functions/get_resources.rb
* [ ] Potential further optimization. Instead of tar'ing up the whole `files/` directory for every module, we could inspect the catalog for `puppet:///` URIs and only include the files referenced by the resources in the catalog. Question? The only place i know if this being used is in a 'file' resource in the 'source' parameter, are there other places where this URI type is valid?

### DIFF
--- a/lib/bolt/applicator.rb
+++ b/lib/bolt/applicator.rb
@@ -25,7 +25,15 @@ module Bolt
       @pool = Concurrent::ThreadPoolExecutor.new(max_threads: max_compiles)
       @logger = Logging.logger[self]
       @plugin_tarball = Concurrent::Delay.new do
-        build_plugin_tarball do |mod|
+        # do NOT encoded into base64 here to save space and allow us to upload
+        # the tarball directly instead of passing it over stdin
+        #
+        # TODO: Potential optimization: only include the files/ that are needed
+        #       by the catalog. This should be anything that is declared with a
+        #       Puppet URI (puppet://).
+        #       Question? The only place i know if this being used is in a 'file' resource
+        #       in the 'source' parameter, are there other places where this URI type is valid?
+        build_plugin_tarball(base64: false) do |mod|
           search_dirs = []
           search_dirs << mod.plugins if mod.plugins?
           search_dirs << mod.pluginfacts if mod.pluginfacts?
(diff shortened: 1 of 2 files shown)


adreyer:
It will definitely not work with orchestrator since the "task" files used are ones deployed to the master. We'll have to continue base64 encoding these there.

I think is a good optimization otherwise. I feel like we're getting to the point that we need to extract the set up a puppet `environment` logic into some shared code.

lib/bolt/applicator.rb
```diff
@@ -48,12 +56,20 @@ def custom_facts_task
...
-    def catalog_apply_task
+    def catalog_apply_task(catalog_path, plugins_path)
       @catalog_apply_task ||= begin
         path = File.join(libexec, 'apply_catalog.rb')
-        file = { 'name' => 'apply_catalog.rb', 'path' => path }
-        metadata = { 'supports_noop' => true, 'input_method' => 'stdin' }
-        Bolt::Task.new(name: 'apply_helpers::apply_catalog', files: [file], metadata: metadata)
+        files = [
+          { 'name' => 'apply_catalog.rb', 'path' => path },
+          { 'name' => 'catalog.json', 'path' => catalog_path },
```
adreyer: I think we should keep the catalog as a parameter to the task rather then a file since it will be different for every invocation.

nmaludy: might be smart to compress the catalog since it's JSON by default, eventually we may run into a spot where the catalog is large (>10MB) and that will cause the same failures

PR #901, opened 2019-03-15
(BOLT-1173) Evaluate Puppet code expressions in YAML plans

nicklewis:
Previously, the only Puppet code that was allowed in a plan was a simple
variable lookup of the form `$foo`. No function calls, string
interpolation, or even array indexing were supported.

We now support more complex Puppet code expressions for every argument of a
step except for name. The specific behavior depends on how the value was
entered into the YAML document, as follows:

* Single-quoted strings are treated as literals
* Double-quoted strings are treated as double-quoted Puppet strings,
 with any `${...}` sections being interpolated
* Folded and "literal" strings (> and |) are treated as blocks of Puppet
 code
* Bare strings are treated as Puppet code if they start with a $, and as
 literal strings otherwise

We now use a customized YAML visitor so that we can preserve the
information about how strings were defined in the document. This visitor
inherits from the standard Ruby visitor and overrides the behavior for
"scalar" nodes. Rather than always converting them directly to strings, we
define classes that capture the various behaviors we want for each type of
string. Otherwise, our loader works akin to the standard
`YAML.safe_load` method.


nicklewis:
I _strongly_ suggest looking at the two commits separately, as the first one implements the behavior and the second is just a code reorganization.

lib/bolt/pal/yaml_plan/evaluator.rb
```diff
@@ -0,0 +1,111 @@
...
+          case value
+          when Array
+            value.map { |element| evaluate_code_blocks(scope, element) }
+          when Hash
+            value.each_with_object({}) do |(k, v), o|
+              key = k.is_a?(EvaluableString) ? k.value : k
+              o[key] = evaluate_code_blocks(scope, v)
+            end
+          when EvaluableString
+            value.evaluate(scope, @evaluator)
```
adreyer: I think we should have a local scope for each of these.

nicklewis: Pushed up a fix for this.

lib/bolt/pal/yaml_plan.rb
```diff
@@ -91,7 +93,9 @@ class CodeLiteral < EvaluableString
         def evaluate(scope, evaluator)
           parse_result = evaluator.parse_string(@value)
 
-          evaluator.evaluate(scope, parse_result)
+          scope.with_local_scope({}) do
+            evaluator.evaluate(scope, parse_result)
```
adreyer: I feel like parse_result could be a method and evaluate could live in the base class.

PR #902, opened 2019-03-15
(BOLT-1172) Support an eval step for arbitrary code

nicklewis:
This step just returns the result of whatever expression is provided for
the `eval` key. This will allow a standard way for code to be executed for
side effects or to generate a value to use later.


nicklewis:
This is based on #901

lib/bolt/pal/yaml_plan/evaluator.rb
```diff
@@ -0,0 +1,117 @@
...
+          description = step['description']
+          raise "Can't run a command without specifying a target" unless target
+
+          args = [command, target]
+          args << description if description
+          scope.call_function('run_command', args)
+        end
+
+        def eval_step(_scope, step)
+          step['eval']
```
adreyer: In order to bring this up to the level of the other steps I think we should add some logging.

nicklewis: What sort of logging do you think we should have here?

It's not entirely straightforward to log like "starting eval/finished eval" because of the current execution model. The code is actually eval-ed _before_ we ever call `eval_step`, since `dispatch_step` evaluates all the nested code blocks before dispatching.

We could log something like "eval-ed {code block} with result {foo}" though.

nicklewis: Actually, by that point we will have thrown out the original code and replaced it with its result.

The behavior of this is fairly unintuitive and really might need to be handled inside the `eval_step` method. It's really weird that `eval: notice("hello world")` doesn't work. Because it's a bare string and doesn't start with a `$`, it's not treated as code. That seems fine in most cases, but very wrong with `eval`.

I think there's a middle ground between the current behavior where `eval` is _just_ a way to assign the value of an expression to a variable, and the possibility where `eval` _always_ treats its argument as code, resulting in double evaluation. Because we have the marker class `BareString`, we could hold off on evaluating the `eval` value until we get into the `eval_step` method, and reinterpret any `BareString` as a `CodeLiteral` instead, so that it will always be evaluated.

That behavior is probably more intuitive, but also quite difficult to explain in detail...

adreyer: well you've at least convinced me to wait to discuss this until we can do some user testing.

PR #905, opened 2019-03-18
(maint) Remove redundant freeze

nicklewis:
Rubocop 0.66.0 now identifies that `ENV[]` will always return frozen
strings and therefore our freeze is redundant.


Iristyle:
I was just about to put up the same PR... 👍

PR #906, opened 2019-03-19
(BOLT-1119) Improve code loading startup times (redux)

Iristyle:
Builds on #900 and refactors some of the way `concurrent` is handled and takes on the next set of changes to improve Bolt startup time.


lib/bolt.rb
```diff
@@ -1,5 +1,4 @@
 # frozen_string_literal: true
 
 module Bolt
-  require 'bolt/executor'
```
Iristyle: This is probably not a good idea... maybe require something else?  Or find a way to make the executor load cheaper?

lib/bolt.rb
```diff
@@ -1,5 +1,18 @@
...
+require 'logging'
+
+# re-open Ruby logging gem
+module Logging
+  # the logging gem always sets itself up to initialize little-plugger
+  # https://github.com/TwP/logging/commit/5aeeffaaa9fe483c2258a23d3b9e92adfafb3b2e
+  class << self
+    # little-plugger calls Gem.find_files, incurring an expensive gem scan
+    # monkey patch Logging to remove the extended method
+    def initialize_plugins; end
```
Iristyle: This monkey patch is the ugliest part of this PR, but given this disables arbitrary code loading on the client, for plugins that we don't use, I'm OK with overriding the base method in this case.

Iristyle: `LittlePlugger` also allows configuration, but only to the default search path - see https://github.com/TwP/little-plugger/blob/master/lib/little-plugger.rb#L302... that config is hardcoded by the `logging` gem with no ability for a user override https://github.com/TwP/logging/commit/5aeeffaaa9fe483c2258a23d3b9e92adfafb3b2e#diff-35a0271f245bc1e1e062519a373c2512R22

(In testing, even if I manually established a path, I couldn't prevent a heavy scan)

lib/bolt.rb
```diff
@@ -1,5 +1,18 @@
 # frozen_string_literal: true
 
+require 'logging'
+
+# re-open Ruby logging gem
+module Logging
```
adreyer: can we move this into it's own file?

donoghuc:
Tested on ubuntu and linux. Seems good. PCP etc work as expected. Only thing of note is that that `scripts/server_client` is now broken for testing bolt server. Though bolt server seems to be fine because I can still use curl
```
cas@cas-ThinkPad-T460p:~/working_dir/bolt$ curl -X POST -H "Content-Type: application/json" -d @developer-docs/examples/ssh-echo.json --cacert $BOLT_CACERT --cert $BOLT_CERT --key $BOLT_KEY $BOLT_ROOT/ssh/run_task
{"node":"172.20.0.1","status":"success","result":{"_output":"24a404cf0afa got passed the message: hey\n"}}
```
Example output for server_client
```
cas@cas-ThinkPad-T460p:~/working_dir/bolt$ bundle exec scripts/server_client.rb sample::echo ssh://bolt:bolt@172.18.0.1:20022 '{"message": "hey"}'
bundler: failed to load command: scripts/server_client.rb (scripts/server_client.rb)
NameError: uninitialized constant BoltSpec::BoltServer::OpenSSL
Did you mean?  Open3
  /home/cas/working_dir/bolt/spec/lib/bolt_spec/bolt_server.rb:57:in `make_client'
  /home/cas/working_dir/bolt/spec/lib/bolt_spec/bolt_server.rb:66:in `puppet_server_get'
  /home/cas/working_dir/bolt/spec/lib/bolt_spec/bolt_server.rb:75:in `get_task_data'
  /home/cas/working_dir/bolt/spec/lib/bolt_spec/bolt_server.rb:84:in `build_request'
  scripts/server_client.rb:46:in `run_task'
  scripts/server_client.rb:65:in `<top (required)>'
```

Iristyle:
Thanks for the `BoltSpec` catch @donoghuc ... would be great if we could rely on specs to find issues like that if possible?

I've updated PR with what I hope will fix that particular problem.

Iristyle:
@adreyer I found a couple of problems with requires in a single file and addressed in fa76d55dc678fd6163724ddab552797a4936a6ce. There could be more, but need to re-run test suite with the right flags to find others (FWIW, I did grep all the code for each of the commits in this PR in an effort to ferret out the transitives).

donoghuc:
Verified with https://github.com/puppetlabs/bolt/pull/906/commits/d541f8825c77aa3ac872ab6f29a01a6775ef354d (latest commit when i went to re-review) and verified that the changes address the issue described above and no regressions were detected. @Iristyle do you have an example of a failing test invocation? I could see if it fails similarly for me.

Iristyle:
The diff for the repush looks good to me @donoghuc 

```patch
--- a/lib/logging_extensions/logging.rb
+++ b/lib/logging_extensions/logging.rb
@@ -1,5 +1,3 @@
-# frozen_string_literal: true
-
 require 'logging'

 module LoggingExtensions
```

Merging

PR #907, opened 2019-03-19
(BOLT-1171) Expose step results as variables

nicklewis:
Previously, there was no way for one step to rely on the result of another
step. Now we expose the result of each named step as a variable with the
same name as the step. This allows parameters for one step to be computed
by using the result of a previous step. The result for a step _without_ a
name will be discarded.

We also now validate that step names are legal variable names and that there
are no duplicate names between parameters and steps, as Puppet variables can
only be defined once.

This also includes a performance improvement to the YAML plan loader as well as
an improvement to how errors are displayed when listing plans.

### DIFF
--- a/lib/bolt/pal/yaml_plan.rb
+++ b/lib/bolt/pal/yaml_plan.rb
@@ -35,6 +35,42 @@ module Bolt
           stringified_step['name'] = stringify(stringified_step['name']) if stringified_step.key?('name')
           stringified_step
         end.freeze
+
+        validate
+      end
+
+      VAR_NAME_PATTERN = /\A[a-z_][a-z0-9_]*\z/.freeze
+
+      def validate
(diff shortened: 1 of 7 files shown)


lib/bolt/pal/yaml_plan.rb
```diff
@@ -35,6 +35,36 @@ def initialize(name, plan)
...
+
+      def validate
+        unless @body.is_a?(Array)
+          raise Bolt::Error.new("Plan must specify an array of steps", "bolt/invalid-plan")
+        end
+
+        used_names = Set.new
+
+        # Parameters come in a hash, so they must be unique
+        used_names.merge @parameters.map(&:name)
```
adreyer: we should validate parameter names too.

nicklewis:
Updated to validate parameter names.

PR #910, opened 2019-03-20
(BOLT-1187) Accept a `return` expression for YAML plans

nicklewis:
Previously, YAML plans always returned `undef`. Now they accept a top-level
`return` key which can be either a value or an expression which will be
evaluated and returned after the plan finishes. Plans which don't specify a
`return` will still return `undef`.


nicklewis:
This is built on top of #907.

PR #911, opened 2019-03-21
(BOLT-1175) Support running plans from YAML plans

nicklewis:
This adds a `plan` step which can be used to run another plan.

This also fixes an issue with scope leaking between plan invocations.
Previously, we were running plans in the scope in which they were called,
causing variables in the surrounding scope to be visible to the plan. We now
run plans from the scope in which they're defined, which is global scope. That
causes every plan invocation to have a fresh scope without any stray variables
polluting it.

### DIFF
--- a/lib/bolt/pal/yaml_plan/evaluator.rb
+++ b/lib/bolt/pal/yaml_plan/evaluator.rb
@@ -11,7 +11,7 @@ module Bolt
           @evaluator = Puppet::Pops::Parser::EvaluatingParser.new
         end
 
-        STEP_KEYS = %w[task command eval script source].freeze
+        STEP_KEYS = %w[task command eval script source plan].freeze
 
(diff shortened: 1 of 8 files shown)


bolt-modules/boltlib/lib/puppet/functions/run_plan.rb
```diff
@@ -66,7 +66,9 @@ def run_plan(scope, plan_name, named_args = {})
         # If the plan does not throw :return by calling the return function it's result is
         # undef/nil
         result = catch(:return) do
-          func.class.dispatcher.dispatchers[0].call_by_name_with_scope(scope, params, true)
+          scope.with_global_scope do |global_scope|
```
adreyer: can you move this commit to a separate PR against master and reference https://tickets.puppetlabs.com/browse/BOLT-1190

PR #913, opened 2019-03-21
(BOLT-1017) Warn when unexpected keys are present in inventory

adreyer:


lib/bolt/inventory/group.rb
```diff
@@ -10,6 +10,11 @@ class Group
       # Regex used to validate group names and target aliases.
       NAME_REGEX = /\A[a-z0-9_][a-z0-9_-]*\Z/.freeze
 
+      DATA_KEYS = %w[name config facts vars features].freeze
+      NODE_KEYS = DATA_KEYS + ['alias']
+      GROUP_KEYS = DATA_KEYS + %w[groups nodes]
+      CONFIG_KEYS = Bolt::TRANSPORTS.keys.map(&:to_s) + ['transport']
```
nicklewis: Do we not allow settings like `user: <...>` outside of a transport?

adreyer: nope

PR #915, opened 2019-03-22
(FM-7869) Move PuppetServer file_cache to Task

da-ar:
Still WIP, posting for early feedback. Will require further testing.

Changes to the API below require that the `name` identifier is provided

```
{
  "target": {
   "remote-transport": "panos",
   "name": "daves_panos",
   "host": "qod65wqq7p4hxpz.delivery.puppetlabs.net",
   "user": "admin",
   "password": "admin",
   "ssl": false
  },
  "task":
    {
    "metadata": {
        "description": "Retrieve a PAN-OS apikey",
        "files": [
            "panos/lib/puppet/util/task_helper.rb",
            "panos/lib/puppet/transport/panos.rb",
            "panos/lib/puppet/transport/schema/panos.rb"
        ],
        "parameters": {},
        "puppet_task_version": 1,
        "remote": true,
        "supports_noop": false
    },
    "name": "panos::apikey",
    "files": [
        {
            "filename": "apikey.rb",
            "sha256": "cbe78202efafd23bbd7ca60a14efcc72e400c71e16b3eda0856c3cbfde7a8601",
            "size_bytes": 524,
            "uri": {
                "path": "/puppet/v3/file_content/tasks/panos/apikey.rb",
                "params": {
                    "environment": "production"
                }
            }
        },
        {
            "filename": "panos/lib/puppet/util/task_helper.rb",
            "sha256": "c6f3c3d50918212e0372afcbf77cdf2d28c873b3b365edbfce1407e1aaa379b9",
            "size_bytes": 914,
            "uri": {
                "path": "/puppet/v3/file_content/plugins/puppet/util/task_helper.rb",
                "params": {
                    "environment": "production"
                }
            }
        },
        {
            "filename": "panos/lib/puppet/transport/panos.rb",
            "sha256": "ff865c99b9996e7beea42f3dd9fe2b261db8ce78359a02f5683e237a7c0478e9",
            "size_bytes": 15002,
            "uri": {
                "path": "/puppet/v3/file_content/plugins/puppet/transport/panos.rb",
                "params": {
                    "environment": "production"
                }
            }
        },
        {
            "filename": "panos/lib/puppet/transport/schema/panos.rb",
            "sha256": "d8f8479cc9e89b57c79bb4e6c1fd6bf89163204ca6c62dbd58f8b45359755eb3",
            "size_bytes": 3112,
            "uri": {
                "path": "/puppet/v3/file_content/plugins/puppet/transport/schema/panos.rb",
                "params": {
                    "environment": "production"
                }
            }
        }
    ]
},
  "parameters": {}
}
```


lib/bolt/transport/remote.rb
```diff
@@ -45,9 +45,9 @@ def get_proxy(target)
       def run_task(target, task, arguments, options = {})
         proxy_target = get_proxy(target)
         transport = @executor.transport(proxy_target.protocol)
-        arguments = arguments.merge('_target' => target.to_h.reject { |_, v| v.nil? })
+        arguments = arguments.merge('_target' => target.options)
 
-        remote_task = Bolt::Task::Remote.new(task.to_h)
+        remote_task = Bolt::Task::Remote.new(task.to_h.each_with_object({}){|(k,v), h| h[k.to_s] = v}, task.file_cache)
```
adreyer: This should probably be a class method on `Bolt::Task::Remote.from_task(task)`

lib/bolt/transport/remote.rb
```diff
@@ -45,9 +45,9 @@ def get_proxy(target)
       def run_task(target, task, arguments, options = {})
         proxy_target = get_proxy(target)
         transport = @executor.transport(proxy_target.protocol)
-        arguments = arguments.merge('_target' => target.to_h.reject { |_, v| v.nil? })
+        arguments = arguments.merge('_target' => target.options)
```
adreyer: I'm pretty sure this will prevent port etc from getting extracted from uris.

DavidS: Yeah, we're in a bind here. `target.to_h` (https://github.com/puppetlabs/bolt/blob/b4dcbcb2bfd21bc7b6a7374305a26b475d370e0e/lib/bolt/target.rb#L92-L102) overwrites `host` from the contents of `name`. This is especially awkward since we just recommended folks to use `host` as key in the transport schema (and released panos 1.0 with that) while https://github.com/puppetlabs/bolt/blob/b4dcbcb2bfd21bc7b6a7374305a26b475d370e0e/lib/bolt_server/transport_app.rb#L103 is a thing, and we can't just `s/hostname/host` to make that work for transports, that might not even have a `host` attribute

Another approach would be to say
```
target = [Bolt::Target.new(body['target']['host'] || body['target']['name'], opts)]
```
and ask the orchestrator to pass down the nodename/certname as `name`.

I wouldn't bother you with this if I weren't aware that all those options are unsatisfactory.

adreyer: This is just the way bolt works for now. You have to include the `host` in the name and there is no way to set it separately. We expect to fix this with version 2 of inventory but you'll need to fix it in ACE directly for now.

Are there modules that will accept but not require a `host` key in connection info. If thats not the case the second option seems like it will work.

adreyer: I expect we'll have inventory v2 functionality available as an opt-in within the next month so perhaps do something temporary now and fix it before ACE goes live.

DavidS: Cisco IOS uses `address` instead of host (but we can change that when porting to Transports), F5 will only provide and expect a single URI.

Longer term, cloud modules definitely have no host at all.

We'll continue exploring workarounds at the ACE API level then.

PR #917, opened 2019-03-25
(BOLT-1206) Allow configuration of BoltSpec via Rspec

donoghuc:
This commit facilitates setting inventory and config for BoltSpec::Run methods using memoized helper methods from RSpec. BoltSpec will attempt to look up a `bolt_inventory` or `bolt_config` method from RSpec and use that if `:config` or `:inventory` is not set in the call to run_*.

### DIFF
--- a/lib/bolt_spec/run.rb
+++ b/lib/bolt_spec/run.rb
@@ -12,6 +12,14 @@ require 'bolt/util'
 module BoltSpec
   module Run
     def run_task(task_name, targets, params, config: nil, inventory: nil)
+      if config.nil? && defined?(bolt_config)
+        config = bolt_config
+      end
+
+      if inventory.nil? && defined?(bolt_inventory)
+        inventory = bolt_inventory
+      end
+
       result = BoltRunner.with_runner(config, inventory) do |runner|
         runner.run_task(task_name, targets, params)
       end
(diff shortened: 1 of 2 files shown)


donoghuc:
@lucywyman you cant use memoized helpers in `before`. Regardless of this change. http://www.betterspecs.org/#let

lib/bolt_spec/run.rb
```diff
@@ -82,8 +82,24 @@ class BoltRunner
       # still be loaded
       def self.with_runner(config_data, inventory_data)
         Dir.mktmpdir do |boltdir_path|
-          config = Bolt::Config.new(Bolt::Boltdir.new(boltdir_path), config_data || {})
-          inventory = Bolt::Inventory.new(inventory_data || {}, config)
+          boltdir = Bolt::Boltdir.new(boltdir_path)
+
+          config = if config_data
+                     Bolt::Config.new(boltdir, config_data)
+                   elsif defined?(RSpec.current_example.example_group_instance.bolt_config)
```
adreyer: I don't think we should rely on RSpec being imported. IMO if any `bolt_config` or `bolt_inventory` is in scope we should just them.

donoghuc: I added this to support an RSpec test that looks something like:
```ruby
require 'bolt_spec/run'

describe "my_test" do
  include BoltSpec::Run

  let(:bolt_config) { {'modulepath' => 'foo' } }

  it 'should return' do
    result = run_task('foo', 'bar')
  end
end
```
Are you suggesting implementing a more generic lookup for `bolt_config` to replace the RSpec specific one, or simply adding a new conditional on `defined?(bolt_config)`?

adreyer: something more generic, We want it to work whether rspec is being used or not.

PR #918, opened 2019-03-25
(BOLT-1201) Use old Psych.parse signature if needed

lucywyman:
**What this changes** This checks the `Psych.parse` signature available in the installed ruby stdlib to see if `legacy_filename` is available. If it is, we can assume passing filename using the keyword parameter. If not, use the old Psych.parse signature and just pass a filename

**Why** So that we can successfully load plans on ruby < 2.6

### DIFF
--- a/lib/bolt/pal/yaml_plan/loader.rb
+++ b/lib/bolt/pal/yaml_plan/loader.rb
@@ -43,7 +43,13 @@ module Bolt
         end
 
         def self.parse_plan(yaml_string, source_ref)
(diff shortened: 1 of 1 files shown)


lucywyman:
@adreyer 2 reasons:
1. It raises a warning if you use legacy
2. We'll *have* to change it if they remove legacy support in a future version and we still need to support 2.5

I don't think it'd get surfaced, and the support removal seems unlikely, but both cases are avoidable.

PR #920, opened 2019-03-26
(BOLT-1204) Document installing bolt on el 8 platforms

donoghuc:
This commit documents installing puppet-bolt on el 8 platforms. Bolt is build and tested on redhat 8, this should also cover centos 8 when that is available.

### DIFF
--- a/pre-docs/bolt_installing.md
+++ b/pre-docs/bolt_installing.md
@@ -194,6 +194,14 @@ The Puppet repository for the YUM package management system is [http://yum.puppe
         sudo yum install puppet-bolt
         ```
 
+    -   Enterprise Linux 8
+
+        ```
+        sudo rpm -Uvh https://yum.puppet.com/puppet6/puppet6-release-el-8.noarch.rpm
+        sudo yum install puppet-bolt
+        ```
+
+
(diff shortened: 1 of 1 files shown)


pre-docs/bolt_installing.md
```diff
@@ -194,6 +194,14 @@ The Puppet repository for the YUM package management system is [http://yum.puppe
         sudo yum install puppet-bolt
         ```
 
+    -   Enterprise Linux 8
+
+        ```
+        sudo rpm -Uvh https://yum.puppet.com/puppet6/puppet6-release-el-8.noarch.rpm
+        sudo yum install puppet-bolt
```
nicklewis: Should we use `dnf` instead of `yum`? They're both symlinks to the same thing.

donoghuc: For all el examples? Or just 8?

nicklewis: Only for 8

nicklewis: It looks like maybe it actually is `yum` on EL 8. There's apparently been some convergence (yum 4 is based on dnf??). It's quite unclear to me which one is the "right" command, so leaving it this way seems fine.

PR #921, opened 2019-03-26
(maint) Update agent module in puppetfile to 2.1.0

McdonaldSeanp:
The puppetlabs-puppet_agent module version 2.1.0 has been released, so we
should be able to install the actual version rather than a git reference


donoghuc:
We should probably have acceptance tests for the new task parameters. I would also like to keep puppet_agent version consistent with the acceptance test setup suite, see: https://github.com/puppetlabs/bolt/pull/863/commits/687f54cab49842f5ae15bccd289354b0eac81721

donoghuc:
@McdonaldSeanp We ended up bumping the `puppet_agent` version in https://github.com/puppetlabs/bolt/pull/965 Thanks!

PR #923, opened 2019-03-26
(BOLT-1132) Add tests for environment variables passing with the local

adreyer:
transport


Gemfile
```diff
@@ -15,6 +15,9 @@ gem "rack", '>= 2.0.5'
 gem "rails-auth", '>= 2.1.4'
 gem "sinatra", '>= 2.0.4'
 
+gem 'aws-sdk', '~> 3'
```
donoghuc: Looks like these were committed on accident?

PR #925, opened 2019-03-27
(BOLT-1193) Collect data about YAML plan usage

nicklewis:
We now submit a "Plan -> yaml" event when running a YAML plan, which
includes the number of steps in the plan and whether the return value of
the plan is a constant value or a Puppet expression.

### DIFF
--- a/lib/bolt/executor.rb
+++ b/lib/bolt/executor.rb
@@ -201,6 +201,22 @@ module Bolt
       @analytics&.event('Apply', 'ast', data)
     end
 
+    def report_yaml_plan(plan)
+      steps = plan.steps.count
+      return_type = case plan.return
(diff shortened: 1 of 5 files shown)


lib/bolt/executor.rb
```diff
@@ -201,6 +201,21 @@ def report_apply(statement_count, resource_counts)
...
 
+    def report_yaml_plan(plan)
+      steps = plan.steps.count
+      return_type = case plan.return
+                    when Bolt::PAL::YamlPlan::EvaluableString
+                      'expression'
+                    when nil
+                      nil
+                    else
+                      'value'
```
adreyer: I don't think we're ever going to see this.

PR #929, opened 2019-03-28
(BOLT-1218) Improve YAML plan documentation

nicklewis:
This commit reworks the docs for YAML plans to flow in the same way that
users will likely adopt features. Previously, it was structured purely
as reference documentation, and assumed a base of Puppet knowledge. This
commit fleshes out some of the information about how to do basic
operations in Puppet like array/hash indexing, function invocation, and
blocks.

This also adds a clear warning that YAML plans are an experimental
feature subject to change.

### DIFF
--- a/pre-docs/writing_yaml_plans.md
+++ b/pre-docs/writing_yaml_plans.md
@@ -1,14 +1,20 @@
 # Writing plans in YAML
 
-Plans written in YAML consist of a list of steps executed in order, which allows for workflows to be defined. Steps can contain embedded Puppet code expressions to add logic.
+**YAML plans are an experimental feature and may experience breaking changes in y releases**
 
-## Naming plans
+YAML plans run a list of steps in order, which allows you to define simple workflows. Steps can contain embedded Puppet code expressions to add logic where necessary.
 
-Plan files should be named `planname.yaml`. Note that the file extension must be `.yaml` and not `.yml`.
+## Defining plans
 
-See [Naming plans](writing_plans.md) for details about where plan files should be located.
+### Plan naming
 
-## Plan structure
+Plan names are based on the module that contains the plan, and the path to the plan in the module.
+
+Plans files should be named `planname.yaml`. Note that the file extension must be `.yaml` and not `.yml`.
+
+See [Naming plans](writing_plans.md) for more details about where plan files should be located.
+
+### Plan structure
 
 A plan is a YAML document containing a map with several keys:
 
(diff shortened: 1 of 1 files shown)


pre-docs/writing_yaml_plans.md
```diff
@@ -165,91 +142,214 @@ steps:
...
 
 steps:
-  - task: echo
-    target: |
-      puppetdb_query('inventory[certname] { facts.osfamily = "Windows" }')
-    description: "Echo the message on the nodes matching a PuppetDB query"
-    parameters:
-      message: $message
+  - name: double_count
+    eval: $count * 2
```
adreyer: I think this should be used in the next step to emphasize why

pre-docs/writing_yaml_plans.md
```diff
@@ -165,91 +142,214 @@ steps:
...
+```yaml
+parameters:
+  numbers:
+    type: Array[Integer]
+
+steps:
+  - task: sum
+    description: "add up the numbers > 5"
+    parameters:
+      indexes: $numbers.filter |$num| { $num > 5 }
```
adreyer: I think we need to explain why this returns an array as opposed to a string briefly here

pre-docs/writing_yaml_plans.md
```diff
@@ -165,91 +142,214 @@ steps:
...
+
+## Advanced Puppet code expressions
+
+You can use a Puppet code expression as the value of any field of a step except the `name`. This allows for complex values to be computed.
+
+Immediately before running a step, any expressions embedded in the step will be evaluated. Each plan parameter and the values of every previous named step will be available in scope.
+
+This lets you take advantage of the power of Puppet language in the places it's necessary, while keeping the rest of your plan simple.
+
 YAML strings are interpreted differently depending on how they're specified.
```
adreyer: I think "handled differently when bolt is running the plan" is better than interpreted here. 

I also think this deserves it's own heading

pre-docs/writing_yaml_plans.md
```diff
@@ -165,91 +142,214 @@ steps:
...
+    target: $nodes
+
+return: $hostnames.map |$hostname_result| { $hostname_result['stdout'] }
+```
+
+## Advanced Puppet code expressions
+
+You can use a Puppet code expression as the value of any field of a step except the `name`. This allows for complex values to be computed.
+
+Immediately before running a step, any expressions embedded in the step will be evaluated. Each plan parameter and the values of every previous named step will be available in scope.
```
adreyer: I think this is a good point to stress the difference between loading and executing.

something like:
Bolt loads the yaml data structure. As it executes each step it will evaluate embedded expressions before running the step.

adreyer:
Overall I think we'll need to flesh out some of the puppet syntax documentation for this to succeed but this is a great first iteration.

nicklewis:
Pushed up changes based on feedback

PR #931, opened 2019-03-29
(BOLT-1150) Add support for YAML plans

nicklewis:
This adds the ability to load and run plans written in YAML.

YAML plans consist of a series of steps (tasks, commands, etc), each of which
will be run in order, making it easy to define a simple workflow. They can
accept parameters and can contain embedded Puppet code expressions to add logic
where necessary.


nicklewis:
This depends on https://github.com/puppetlabs/bolt-vanagon/pull/88

PR #932, opened 2019-04-01
(maint) Docs typo: http -> https

lucywyman:
**What this changes** Corrects "ssl: When true, Bolt will use normal http connections for WinRM. Default is true" to "ssl: When true, Bolt will use https connections for WinRM. Default is true"


melamos:
Moved to docs.

PR #933, opened 2019-04-01
(maint) Activate bolt_server integration test

donoghuc:
The app_integration.rb spec appeared to have been filtered out because it did not match the pattern `--pattern spec/\*\*\{,/\*/\*\*\}/\*_spec.rb`.

### DIFF
--- a/spec/bolt_server/app_integration_spec.rb
+++ b/spec/bolt_server/app_integration_spec.rb
@@ -52,15 +52,19 @@ describe "BoltServer::TransportApp", puppetserver: true do
 
     it 'runs a shareable task' do
       body = build_request('shareable',
-                           conn_target('ssh', include_password: true),
-                           "message": "Hello!")
+                           conn_target('ssh', include_password: true))
 
       post(path, JSON.generate(body), 'CONTENT_TYPE' => 'text/json')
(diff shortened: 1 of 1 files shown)


spec/bolt_server/app_integration_spec.rb
```diff
@@ -60,7 +60,12 @@ def app
       expect(last_response.status).to eq(200)
       result = JSON.parse(last_response.body)
       expect(result).to include('status' => 'success')
-      expect(result['result']['_output'].chomp).to match(/\w+ got passed the message: Hello!/)
+      files = result['result']['_output'].split("\n").map(&:strip).sort
```
donoghuc: This appears to have been a copy paste error, the expectation defined in the test would match the `echo` task, not the `shareable` task. We were not seeing this fail in CI because rspec filtered it out.

spec/bolt_server/app_integration_spec.rb
```diff
@@ -60,7 +60,12 @@ def app
       expect(last_response.status).to eq(200)
```
lucywyman: I think we can also delete the `"message": "Hello!"` line since shareable doesn't have parameters

PR #935, opened 2019-04-02
(BOLT-1188) Support new version 2 inventory file

adreyer:


lib/bolt/inventory/group2.rb
```diff
@@ -0,0 +1,313 @@
...
+
+        @nodes = {}
+        @aliases = {}
+        nodes.reject { |node| node.is_a?(String) }.each do |node|
+          unless node.is_a?(Hash)
+            raise ValidationError.new("Node entry must be a String or Hash, not #{node.class}", @name)
+          end
+
+          node['name'] ||= node['uri']
+          name = node['name']
```
lucywyman: Nit: rename to `node_name` or something to avoid conflict with the group name `@name`

lib/bolt/inventory/inventory2.rb
```diff
@@ -0,0 +1,264 @@
...
+
+module Bolt
+  class Inventory
+    class Inventory2
+      def initialize(data, config = nil, target_vars: {}, target_facts: {}, target_features: {})
+        @logger = Logging.logger[self]
+        # Config is saved to add config options to targets
+        @config = config || Bolt::Config.default
+        @data = data ||= {}
+        @groups = Group2.new(data.merge('name' => 'all'))
```
lucywyman: Is the only diff `Group` to `Group2`?  Would `super`, then resetting `@groups` to Group2 work?

adreyer: If we hadn't decided to copy/paste that would have worked but processing all the groups is really expensive so we wouldn't want to do it twice. We could have created a load_groups method to capture this difference though.

As it is `super` would just call the base `object` initialize and won't work.

lib/bolt/inventory/group2.rb
```diff
@@ -0,0 +1,312 @@
...
+  class Inventory
+    class Group2 < Group
+      attr_accessor :name, :nodes, :aliases, :name_or_alias, :groups, :config, :rest, :facts, :vars, :features
+
+      # THESE are duplicates with the old groups for now.
+      # Regex used to validate group names and target aliases.
+      NAME_REGEX = /\A[a-z0-9_][a-z0-9_-]*\Z/.freeze
+
+      DATA_KEYS = %w[name config facts vars features].freeze
+      NODE_KEYS = DATA_KEYS + ['alias']
```
lucywyman: No `uri`? It seems like this is only used for validating the yaml, and getting a `Found unexpected key(s) uri` seems wrong. Does postponing hostname validation rely on uri not being here?

lucywyman:
I get `Target myhost does not have a host` with the following:

lib/bolt/transport/ssh/connection.rb
```diff
@@ -68,6 +68,8 @@ def initialize(target, transport_logger, load_config = true)
           require 'net/ssh'
           require 'net/ssh/proxy/jump'
 
+          raise Bolt::ValidationError, "Target #{target.name} does not have a host" unless target.host
+
```
lucywyman: Phew ok finally figured it out - we need to add `host` to the options hash of each transport https://github.com/puppetlabs/bolt/blob/master/lib/bolt/transport/ssh.rb#L12

lib/bolt/inventory.rb
```diff
@@ -53,11 +54,23 @@ def self.from_config(config)
...
-      inventory = new(data, config)
+      inventory = create_version(data, config)
       inventory.validate
       inventory
     end
 
+    def self.create_version(data, config)
+      version = (data || {}).delete('version') { 1 }
+      case version
+      when 1
```
trevor-vaughan: This should probably follow semantic versioning unless you want *every* update to be breaking.

It would make sense to be able to fix typos and/or add fields without having to bump the version number.

adreyer: We currently only intend to version this file for breaking changes. We've added many features to version `1` without worrying about incrementing or noting the minor version number. This aligns with other file formats in the puppet world.

adreyer: The problem we're trying to solve is to allow one version of bolt to support multiple different incompatible versions of the inventory file format. By specifying the `version` bolt knows which implementation to use to load the file. That is a different problem than semver is intended to solve which is choosing a compatible version of a library to install when resolving dependencies.

lib/bolt/inventory/inventory2.rb
```diff
@@ -0,0 +1,264 @@
...
+      private :inherit_data
+
+      def create_target(target_name)
+        data = @groups.data_for(target_name) || {}
+        name_opt = {}
+        name_opt['name'] = data['name'] if data['name']
+
+        # If there is no name then this node was only referred to as a string.
+        uri = data['uri']
+        uri ||= target_name unless data['name']
```
lucywyman: I think this needs to get set to host if URI isn't set and 'name' is defined, otherwise `target.uri` returns nil from the plan.

lib/bolt/target.rb
```diff
@@ -4,10 +4,10 @@
 
 module Bolt
   class Target
-    attr_reader :uri, :options
+    attr_reader :options
     # CODEREVIEW: this feels wrong. The altertative is threading inventory through the
     # executor to the RemoteTransport
-    attr_accessor :inventory
+    attr_accessor :uri, :inventory
```
lucywyman: If uri can be nil, I think we need to make a getter method that will return the `host` and connection information if `target.uri` is called.

adreyer: Why? IMO the `uri` exists to provide a concise method to pass a lot of connection information in a single string from the CLI.

* None of our transports rely on connecting with the URI directly. 
* It seems better to be explicit about when a real uri was passed vs trying to build one there is too much weirdness around uris and escaping.
* Users should be encouraged to use *name* when stringifying a node. That is why we added a name method previously even though it could only be the URI.

lucywyman:
I'm not sure how to get around this, but https://github.com/puppetlabs/bolt/blob/master/bolt-modules/boltlib/lib/puppet/datatypes/target.rb#L10 means that if I call `target.host` before trying to run anything on the host I get 
```
{
  "kind": "bolt/pal-error",
  "msg": "value returned from function 'host' has wrong type, expects a String[1] value, got String",
  "details": {
  }
}
```
instead of the validation error. I don't think this is a blocking issue, but if it's easy to raise the validation error instead I think that's a better user experience. `uri` and `name` don't have the same issue, which makes this a little more confusing if you don't know how the target datatype is implemented.

adreyer:
@lucy I think we just need to update the datatype.  What was in your inventory that caused an empty string for host?

lib/bolt/target.rb
lucywyman: Ok - those are good points.

lucywyman:
```
version: 2
groups:
  - name: vms
    nodes: 
      - name: myhostname
```
where `myhostname` is an actual fqdn.

adreyer:
hmm I get undef not empty string with just a name. I don't think we need to handle empty string but I do want to make sure there is no easy way to create a target with an emptystring uri.

lib/bolt/target.rb
```diff
@@ -40,6 +41,14 @@ def initialize(uri, options = nil)
...
         @protocol = @options['protocol']
       end
+
+      if @options['host']
+        @host = @options['host']
+      end
+
+      # WARNING: name should never be updated
+      @name = @options['name'] || @uri
+      # TODO: this should error unless name is set
```
lucywyman: Do you want to leave this in?

lib/bolt/target.rb
```diff
@@ -77,8 +89,17 @@ def features
       end
     end
 
+    # TODO: WHAT does equality mean here?
+    # should we just compare names? is there something else that is meaninful?
```
lucywyman: What you've got here makes sense to me.

spec/integration/inventory2_spec.rb
```diff
@@ -0,0 +1,442 @@
...
+      }],
+      config: {
+        ssh: { 'host-key-check' => false },
+        winrm: { ssl: false, 'ssl-verify' => false }
+      },
+      vars: {
+        daffy: "duck"
+      },
+      facts: {
+        scooby: 'doo',
```
lucywyman: Whoa, really mixing your cartoons here.

lib/bolt/target.rb
adreyer: Yeah I think this is a question to answer later.

lib/bolt/target.rb
adreyer: I'm removing it. I don't think it matters until we update `Target` in puppet.

PR #936, opened 2019-04-03
(BOLT-1026) Add hiera-eyaml package to bolt

donoghuc:
This commit adds hiera-eyaml to the gemspec and includes a test to ensure the eyaml can be parsed/decoded in an apply block.


spec/integration/apply_compile_spec.rb
```diff
@@ -191,6 +191,21 @@ def get_notifies(result)
           'hiera-config' => File.join(__dir__, '../fixtures/apply/hiera_invalid.yaml').to_s
         }
       }
+      let(:eyaml_config) {
```
donoghuc: I could not figure out how to reliably add a relative path in a hiera config file, so instead of committing one like the other tests, i'm constructing one dynamically. Specifically, the paths to the `pkcs7` key pair.

donoghuc:
Added WIP label because hiera-eyaml and dependencies are included in puppet runtime. When puppet-runtime bumped from hiera-eyaml 2.1.0 to 3.0.0 it appears that a dependency was replaced. I am sorting that out with puppet agent maintainers, merging this is blocked until https://github.com/puppetlabs/puppet-runtime/pull/154 is merged and promoted to bolt-vanagon.

PR #939, opened 2019-04-04
(maint) Simplify Executor initialize signature

nicklewis:
The Executor currently takes both optional args and keyword args. It also takes
a couple of its arguments (`bundled_content` and `load_config`) purely so it
can pass them along to other parts of the system that need them.

This PR reworks the keyword args to be set more directly on the objects that
need them, rather than passing them through the Executor.

We now set `bundled_content` directly on the analytics client, since that's the only place it's needed.

We set `load-config` directly on Target objects via a default on the transport,
and override it explicitly in bolt-server. This carries the setting with the
Target, rather than leaking an implementation detail of the SSH transport into
the Executor. This specifically does _not_ allow `load-config` to be set in the
Bolt config or inventory file.

### DIFF
--- a/lib/bolt/transport/ssh/connection.rb
+++ b/lib/bolt/transport/ssh/connection.rb
@@ -14,15 +14,15 @@ module Bolt
         attr_reader :logger, :user, :target
         attr_writer :run_as
 
-        def initialize(target, transport_logger, load_config = true)
+        def initialize(target, transport_logger)
           # lazy-load expensive gem code
           require 'net/ssh'
           require 'net/ssh/proxy/jump'
 
           @target = target
-          @load_config = load_config
+          @load_config = target.options['load-config']
 
-          ssh_user = load_config ? Net::SSH::Config.for(target.host)[:user] : nil
+          ssh_user = @load_config ? Net::SSH::Config.for(target.host)[:user] : nil
           @user = @target.user || ssh_user || Etc.getlogin
           @run_as = nil
 
--- a/spec/bolt/applicator_spec.rb
+++ b/spec/bolt/applicator_spec.rb
@@ -51,7 +51,7 @@ describe Bolt::Applicator do
         config: {
           transport: 'ssh',
           transports: {
-            ssh: { 'connect-timeout' => 10, 'host-key-check' => true, tty: false },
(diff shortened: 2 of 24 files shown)


lib/bolt/transport/ssh/connection.rb
```diff
@@ -63,15 +63,15 @@ def delete
...
           # lazy-load expensive gem code
           require 'net/ssh'
           require 'net/ssh/proxy/jump'
 
           @target = target
-          @load_config = load_config
+          @load_config = target.options['load-config']
 
-          ssh_user = load_config ? Net::SSH::Config.for(target.host)[:user] : nil
+          ssh_user = @load_config ? Net::SSH::Config.for(target.host)[:user] : nil
```
adreyer: Can we handle this when we set `load_config`? can we handle the default in config instead of the initializer?

spec/bolt/applicator_spec.rb
```diff
@@ -51,7 +51,7 @@
         config: {
           transport: 'ssh',
           transports: {
-            ssh: { 'connect-timeout' => 10, 'host-key-check' => true, tty: false },
+            ssh: { 'connect-timeout' => 10, 'host-key-check' => true, tty: false, 'load-config' => true },
```
adreyer: Why are we loading config for tests?

nicklewis: This is the "expected" value for the input to catalog compilation. It expects `load-config => true` because that's the default.

lib/bolt/transport/ssh/connection.rb
nicklewis: Do you mean load the SSH config when we create the Target instance? We don't currently have any transport-specific code in the Target class, and this would mean adding some.

nicklewis:
I believe I have burned the appropriate incense and made the appropriate offerings to satisfy rubocop.

lib/bolt/transport/ssh/connection.rb
adreyer: I meant either 
just compute this a few lines earlier and make @load_config the computed value.
or
set the default as a normal config default.

spec/bolt/applicator_spec.rb
adreyer: oh right. I remember hating this last time I had to change it.

PR #941, opened 2019-04-08
(BOLT-1209) Refactor AppVeyor commands to PowerShell script / verify WinRM connection in PowerShell

Iristyle:
- There's a bunch of redundant and hard to read code in appveyor.yml
   that is moved verbatim to a separate file ci.ps1, then is refactored

 - Break it into reusable functions (that can also be used in local dev):

   * Set-CACert - Sets `SSL_CERT_FILE` environment variable to the latest CA bundel from curl.haxx.se
   * Install-Puppetfile - Installs the CA bundle, sets env var and installs the Puppetfile
   * New-RandomPassword - generates random password
   * New-LocalAdmin - creates a new user with name / password and adds them to RDP group
   * Install-Certificate - installs a cert in the machine store given its password
   * Grant-WinRMHttpsAccess - enables winrm SSL access given a cert thumbprint
   * Set-WinRMHostConfiguration - CI helper that wraps up installing the bolt cert and configuring winrm to use it
   * Set-ActiveRubyFromPuppet - ensures Puppets Ruby is earlier in `PATH` than System Ruby

- Logic in `appveyor.yml` now runs through some very high level functions that are more descriptive

- Adds new function `Test-WinRMConfiguration` - uses the `New-PSSession` cmdlet to establish a WinRM SSL connection back to `localhost` given a set of credentials. By default waits 15 times at 1 second intervals before throwing a terminating error.

- Retries acquiring CA bundle from curl.haxx.se

- Retries installing the `Puppetfile`

The hope here is that if WinRM needs some time to become available, the pre-suite checks should verify that / fail fast if necessary.

### DIFF
--- a/ci.ps1
+++ b/ci.ps1
@@ -21,13 +21,21 @@ function Install-Puppetfile
   bundle exec r10k puppetfile install
 }
 
-function Set-WinRMHostConfiguration
+function New-RandomPassword
 {
   Add-Type -AssemblyName System.Web
-  $ENV:BOLT_WINRM_PASSWORD = "&aA4" + [System.Web.Security.Membership]::GeneratePassword(10, 3)
-  ($user = New-LocalUser -Name $ENV:BOLT_WINRM_USER -Password (ConvertTo-SecureString -String $ENV:BOLT_WINRM_PASSWORD -Force -AsPlainText)) | Format-List
+  "&aA4" + [System.Web.Security.Membership]::GeneratePassword(10, 3)
+}
+
+function New-LocalAdmin($userName, $password)
+{
+  ($user = New-LocalUser -Name $userName -Password (ConvertTo-SecureString -String $password -Force -AsPlainText)) | Format-List
   Add-LocalGroupMember -Group 'Remote Management Users' -Member $user
   Add-LocalGroupMember -Group Administrators -Member $user
+}
+
+function Set-WinRMHostConfiguration
+{
   # configure WinRM to use resources/cert.pfx for SSL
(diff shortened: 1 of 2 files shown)


ci.ps1
```diff
@@ -0,0 +1,149 @@
...
+      }
+
+      if (New-PSSession @sessionArgs) { return $true }
+    }
+  }
+
+  Invoke-ScriptBlockWithRetry @retryArgs
+}
+
+# Ensure Puppet Ruby 5 / 6 takes precedence over system Ruby
```
donoghuc: IIRC these paths were to ensure that puppet ruby was picked up for the `basic::ruby_task`. Now that we are able to configure `implementations` for winrm transport we should be able to handle this in 
```ruby
puppet_5_config = config
puppet_5_config['winrm']['implementations'] = { '.rb' =>  "#{ENV['ProgramFiles'}\Puppet Labs\Puppet\sys\ruby\bin" }
with_tempfile_containing('bolt', YAML.dump(puppet_5_config), '.yaml') do |conf|
```
in https://github.com/puppetlabs/bolt/blob/b062fa3faed7fa8600ef17275a46933b37a12bc6/spec/integration/apply_spec.rb#L278-L279

Iristyle: I'd say let's leave that for a follow up because this PR only touches `appveyor.yml`, which makes it simpler to vet / merge.

donoghuc: That sounds good. Once we see this merged and green I can circle back on that.

PR #943, opened 2019-04-09
(BOLT-1205, BOLT-1200) Add default local config only for localhost

lucywyman:
This ensures that 'magic' local config only gets merged with the provided config when the hostname is localhost, regardless of the transport used. The 'magic' config options for localhost are to add the `puppet-agent` feature to the localhost target, and to set the default ruby interpreter to be Bolt's ruby.

### DIFF
--- a/lib/bolt/inventory.rb
+++ b/lib/bolt/inventory.rb
@@ -177,6 +177,21 @@ module Bolt
     end
     private :groups_in
 
+    # TODO: Possibly refactor this once inventory v2 is more stable
+    def self.localhost_defaults(data)
+      defaults = {
+        'config' => {
+          'transport' => 'local',
+          'local' => { 'interpreters' => { '.rb' => RbConfig.ruby } }
+        },
+        'features' => ['puppet-agent']
+      }
+      data = Bolt::Util.deep_merge(defaults, data)
+      # If features is an empty array deep_merge won't add the puppet-agent
+      data['features'] << 'puppet-agent' if data['features'].empty?
+      data
+    end
+
     # Pass a target to get_targets for a public version of this
     # Should this reconfigure configured targets?
     def update_target(target)
@@ -188,10 +203,7 @@ module Bolt
         data['config'] = {}
       end
 
-      unless data['config']['transport']
-        data['config']['transport'] = 'local' if target.name == 'localhost'
-      end
-
+      data = self.class.localhost_defaults(data) if target.name == 'localhost'
       # These should only get set from the inventory if they have not yet
       # been instantiated
       set_vars_from_hash(target.name, data['vars']) unless @target_vars[target.name]
--- a/lib/bolt/inventory/inventory2.rb
+++ b/lib/bolt/inventory/inventory2.rb
@@ -122,10 +122,7 @@ module Bolt
           data['config'] = {}
         end
 
-        unless data['config']['transport']
-          data['config']['transport'] = 'local' if target.name == 'localhost'
-        end
-
+        data = Bolt::Inventory.localhost_defaults(data) if target.name == 'localhost'
         # These should only get set from the inventory if they have not yet
         # been instantiated
         set_vars_from_hash(target.name, data['vars']) unless @target_vars[target.name]
--- a/lib/bolt/task.rb
+++ b/lib/bolt/task.rb
@@ -72,9 +72,9 @@ module Bolt
 
     # Returns a hash of implementation name, path to executable, input method (if defined),
     # and any additional files (name and path)
-    def select_implementation(target, additional_features = [])
+    def select_implementation(target, provided_features = [])
       impl = if (impls = implementations)
-               available_features = target.features + additional_features
+               available_features = target.features + provided_features
                impl = impls.find do |imp|
                  remote_impl = imp['remote']
                  remote_impl = metadata['remote'] if remote_impl.nil?
--- a/spec/bolt/inventory/inventory2_spec.rb
+++ b/spec/bolt/inventory/inventory2_spec.rb
@@ -665,4 +665,63 @@ describe Bolt::Inventory::Inventory2 do
       end
     end
   end
+
+  context 'with localhost' do
+    context 'with no additional config' do
+      let(:data) {
+        { 'nodes' => ['localhost'] }
+      }
+
+      let(:inventory) { Bolt::Inventory.new(data) }
+
+      it 'adds magic config options' do
+        target = get_target(inventory, 'localhost')
+        expect(target.protocol).to eq('local')
+        expect(target.options['interpreters']).to include('.rb' => RbConfig.ruby)
+        expect(target.features).to include('puppet-agent')
+      end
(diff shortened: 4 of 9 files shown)


lib/bolt/inventory.rb
```diff
@@ -188,8 +188,15 @@ def update_target(target)
         data['config'] = {}
       end
 
-      unless data['config']['transport']
-        data['config']['transport'] = 'local' if target.name == 'localhost'
+      if target.name == 'localhost'
```
adreyer: Can we pull this into a separate method.

lib/bolt/task.rb
```diff
@@ -72,9 +72,9 @@ def implementations
 
     # Returns a hash of implementation name, path to executable, input method (if defined),
     # and any additional files (name and path)
-    def select_implementation(target, additional_features = [])
+    def select_implementation(target, provided_features = [])
```
lucywyman: This is just for consistency + searchability with the transports

lib/bolt/inventory.rb
```diff
@@ -188,8 +188,15 @@ def update_target(target)
...
+        unless data['config']['transport']
+          data['config']['transport'] = 'local'
+        end
+        t = data['config']['transport']
```
adreyer: I don't think we should update data on other transports after seeing how this works it definitely feels wrong.

spec/bolt/inventory/inventory2_spec.rb
```diff
@@ -665,4 +665,45 @@ def common_data(transport)
       end
     end
   end
+
+  context 'with localhost' do
```
adreyer: We should have a test to ensure this works without anything named localhost too probably.

lib/bolt/inventory/inventory2.rb
```diff
@@ -122,8 +122,15 @@ def update_target(target)
           data['config'] = {}
         end
 
-        unless data['config']['transport']
-          data['config']['transport'] = 'local' if target.name == 'localhost'
+        if target.name == 'localhost'
+          unless data['config']['transport']
+            data['config']['transport'] = 'local'
+          end
+          t = data['config']['transport']
```
donoghuc: Should `t` just be hard coded to `local`? We only want to set interpreter for `local`.

lucywyman: In a convo with Alex we decided to apply this to localhost on all transports, but I think given comments above we're reversing that.

donoghuc: Oh, that is what alex said. Sorry, didn't see that review first.

lucywyman: Np - I also assumed it was only on the local transport, and you totally weren't in the convo where we decided otherwise! I should've updated the ticket.

donoghuc:
I think you just need to remove https://github.com/puppetlabs/bolt/blob/2c75b2850a229850439cde382e172eddaec89ec8/spec/bolt/applicator_spec.rb#L60 and it should be good to go.

spec/bolt/inventory/inventory2_spec.rb
adreyer: I'm still not sure where we're testing that the special localhost target works even if the user has no node named localhost in their inventory(which is the more common case)

lib/bolt/inventory.rb
```diff
@@ -177,6 +177,22 @@ def groups_in(node_name)
     end
     private :groups_in
 
+    def self.magic_local_config(target, data)
+      if target.name == 'localhost'
```
adreyer: nitpick. I think move this conditional out to the caller and call this "localhost_defaults"

lib/bolt/inventory.rb
```diff
@@ -177,6 +177,22 @@ def groups_in(node_name)
...
+        unless data['config']['transport']
+          data['config']['transport'] = 'local'
+        end
+        if data['config']['transport'] == 'local'
```
adreyer: I think we can safely set these fields regardless of what the transport is. It doesn't matter with the way we use this now but less conditionals is usually better than more.

lib/bolt/inventory.rb
adreyer: can we use deep merge?

spec/integration/local_spec.rb
```diff
@@ -76,10 +82,13 @@
       end
 
       it 'exposes environment variables during apply' do
-        result = run_cli_json(%w[plan run env_var::get_var] + config_flags)
-        expect(result).not_to include('kind')
-        event = result.first['result']['report']['resource_statuses']['Notify[gettingvar]']['events'].first
-        expect(event).to include('message' => "defined 'message' as 'testing this'")
+        # Uses system ruby instead of bolt's ruby
```
adreyer: It seems like we either need to move this test out of the block where `uri` is set to two nodes or actually make assertions about all the results.

lucywyman: I think the test is actually testing whether environment variables are available in an apply block. The comment was misleading (it's what I thought was happening and I forgot to update). In reality I think because it *isn't* getting set to Bolt's ruby because the hostname isn't localhost, it's failing. I think we can safely test statement on just `localhost` and re-set the URI for that test, or we can move it out and assert that it errors. I don't see too much point in asserting an error though, since we're not actually testing that here and have already verified that the interpreter configuration is picked up correctly in the `inventory_spec` test.

lib/bolt/inventory.rb
```diff
@@ -177,6 +177,17 @@ def groups_in(node_name)
     end
     private :groups_in
 
+    def self.localhost_defaults(data)
+      defaults = { 'config' => {
+        'transport' => 'local',
+        'local' => { 'interpreters' => { '.rb' => RbConfig.ruby } }
+      } }
+      data = Bolt::Util.deep_merge(defaults, data)
+      data['features'] ||= []
```
donoghuc: Might as well throw `{ 'features' => ['puppet-agent'] }` in `defaults` hash also?

lucywyman: Hmm, I think the issue is that sometimes features is set to an empty array in `data` which overrides the `puppet-agent`. I'll see what I can do!

lucywyman: I think this is a little better:
```
      defaults = { 
        'config' => {
          'transport' => 'local',
          'local' => { 'interpreters' => { '.rb' => RbConfig.ruby } } 
        },
        'features' => ['puppet-agent']
      }   
      data = Bolt::Util.deep_merge(defaults, data)
      # If features is an empty array deep_merge won't add the puppet-agent
      data['features'] << 'puppet-agent' if data['features'].empty?
      data
```

adreyer: long term this is where we should just use the logic from Group:
https://github.com/puppetlabs/bolt/blob/80bf29f16b226686a27dd8ce6df6c5cd53d01cdf/lib/bolt/inventory/group.rb#L217

PR #946, opened 2019-04-10
(BOLT-843) Support retrying on failed nodes.

adreyer:
Bolt will now write a <boltdir>/.last_result.json file whenever a
command fails. Users can use this file to target nodes from the last
failure with the `--retry` flag.

In order to avoid spreading `nodes` to a new file this also adds a
`target` key to the status hash of Results.

This still needs predocs.

### DIFF
--- a/lib/bolt/cli.rb
+++ b/lib/bolt/cli.rb
@@ -15,6 +15,7 @@ require 'bolt/config'
 require 'bolt/error'
 require 'bolt/executor'
 require 'bolt/inventory'
+require 'bolt/rerun'
 require 'bolt/logger'
 require 'bolt/outputter'
 require 'bolt/puppetdb'
@@ -117,22 +118,32 @@ module Bolt
 
       # After validation, initialize inventory and targets. Errors here are better to catch early.
       unless options[:subcommand] == 'puppetfile' || options[:action] == 'show'
-        if options[:query]
-          if options[:nodes].any?
-            raise Bolt::CLIError, "Only one of '--nodes' or '--query' may be specified"
-          end
-          nodes = query_puppetdb_nodes(options[:query])
-          options[:targets] = inventory.get_targets(nodes)
-          options[:nodes] = nodes if options[:subcommand] == 'plan'
-        else
-          options[:targets] = inventory.get_targets(options[:nodes])
-        end
+        options[:targets] = get_targets(options)
       end
 
       options
     rescue Bolt::Error => e
       warn e.message
       raise e
+    end
+
+    def get_targets(options)
+      target_opts =  options.keys.select { |opt| %i[query rerun].include?(opt) }
+      target_opts << :nodes unless options[:nodes].empty?
+      if target_opts.length > 1
+        raise Bolt::CLIError, "Only one of '--nodes', '--rerun', or '--query' may be specified"
+      end
+
+      if options[:query]
+        nodes = query_puppetdb_nodes(options[:query])
+        options[:nodes] = nodes if options[:subcommand] == 'plan'
+      elsif options[:rerun]
+        nodes = rerun.get_targets(options[:rerun])
+        options[:nodes] = nodes if options[:subcommand] == 'plan'
+      else
+        nodes = options[:nodes]
+      end
+      inventory.get_targets(nodes)
     end
 
     def validate(options)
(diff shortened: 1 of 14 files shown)


lib/bolt/last_failure.rb
```diff
@@ -0,0 +1,42 @@
...
+                 when 'success'
+                   data.select { |result| result['status'] == 'success' }
+                 else
+                   raise Bolt::CLIError, "Unexpected option #{filter} for '--retry'"
+                 end
+      filtered.map { |result| result['target'] }
+    end
+
+    def write(result_set)
+      data = result_set.map { |result| result.status_hash.select { |k, _| %i[target status].include? k } }
```
adreyer: This will result in the password for nodes where the password is in the name being persisted to disk. The only way this seems to result in new information being persisted to disk is if:
1. The password was included in `--nodes` which was probably already persisted by the shell history unless the nodes were passed on `stdin`
2. Node uris are built inside a plan and contain the password.

Options:
1. We can have a config option to disable writing this file. This may be a good idea anyway
2. We can formalize the concept of a "safe" name in the target and save that with the knowledge that those nodes can't be rerun.
3. do nothing.

adreyer: I've update the PR with 2 so the remaining question is do we need to do 1 in addition.

lib/bolt/cli.rb
```diff
@@ -380,6 +392,7 @@ def run_plan(plan_name, plan_arguments, nodes, options)
 
       # If a non-bolt exception bubbles up the plan won't get finished
       executor.finish_plan(result)
+      handle_last_failure(result)
```
donoghuc: I dont think this handles plan results. 
```
From: /home/cas/working_dir/bolt/lib/bolt/cli.rb @ line 484 Bolt::CLI#handle_last_failure:

    483: def handle_last_failure(result)
 => 484:   if result.is_a?(Bolt::ResultSet) && !result.ok
    485:     last_failure.write(result)
    486:   end
    487: end

[1] pry(#<Bolt::CLI>)> result.ok
NoMethodError: undefined method `ok' for #<Bolt::PlanResult:0x0000559228c54f08>
Did you mean?  ok?
from (pry):2:in `handle_last_failure'
[2] pry(#<Bolt::CLI>)> result.ok?
=> false
[3] pry(#<Bolt::CLI>)> result.is_a?(Bolt::ResultSet)
=> false
```

spec/bolt/transport/ssh_spec.rb
```diff
@@ -2,6 +2,7 @@
 
 require 'spec_helper'
 require 'net/ssh'
+require 'net/ssh/proxy/jump'
```
donoghuc: I guess I missed adding this in when I looked for tests that were not standalone after refactoring loading code.

spec/bolt/last_failure_spec.rb
```diff
@@ -0,0 +1,209 @@
...
+
+    it 'does not update the file when a plan fails with nil' do
+      expect(pal).to receive(:run_plan)
+        .and_return(Bolt::PlanResult.new('oops', 'failure'))
+      run_cli(%w[plan run whoami])
+
+      expect(read_last_failure).to eq(['original result'])
+    end
+
+    it 'does not update the file when a plan fails with nil' do
```
donoghuc: Copy paste error? This looks more like `does not update the file when a plan succeeds`.

spec/bolt/last_failure_spec.rb
```diff
@@ -0,0 +1,217 @@
...
+      expect(pal).to receive(:run_plan)
+        .and_return(Bolt::PlanResult.new(nil, 'failure'))
+      run_cli(%w[plan run whoami])
+
+      expect(read_last_failure).to eq(['original result'])
+    end
+
+    it 'does not update the file when a plan succeeds with a result set' do
```
donoghuc: I think `success_set` should also be wrapped in a plan result. 
```ruby
    it 'does not update the file when a plan succeeds with a result set' do
      pr = Bolt::PlanResult.new(success_set, 'success')
      expect(pal).to receive(:run_plan)
        .and_return(pr)
      run_cli(%w[plan run whoami])
      expect(read_last_failure).to eq(['original result'])
    end
```

donoghuc: I think `success_set` should also be wrapped in a plan result. 
```ruby
    it 'does not update the file when a plan succeeds with a result set' do
      pr = Bolt::PlanResult.new(success_set, 'success')
      expect(pal).to receive(:run_plan)
        .and_return(pr)
      run_cli(%w[plan run whoami])
      expect(read_last_failure).to eq(['original result'])
    end
```

PR #947, opened 2019-04-10
(maint) Speed up PuppetDB tests which timed out

nicklewis:
Previously, these tests were using bad_host{1,2}.dne.com as their
example "bad" hosts. However, those hosts happen to actually resolve to
servers somewhere, which blackhole our connections, causing them to
block for 60 seconds before finally timing out. Since we attempt three
total connections to these hosts, that adds 180 unnecessary seconds to
the test runtime. In general, it's also bad practice for our tests to
attempt to connect to random machines on the internet.

We now use example.com, which is specifically intended for this kind of
use case and simply fails to resolve. This causes the tests to fail fast
and cuts off three minutes from their runtime.

### DIFF
--- a/spec/bolt/puppetdb/client_spec.rb
+++ b/spec/bolt/puppetdb/client_spec.rb
@@ -108,15 +108,15 @@ describe Bolt::PuppetDB::Client do
 
     it 'should fail after all servers fail' do
       conf = pdb_conf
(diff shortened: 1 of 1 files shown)


nicklewis:
I accidentally changed my git author/email config, whoops. I just reset this commit and updated.

PR #948, opened 2019-04-11
(BOLT-1242) Don't check for sudo prompt when stderr is EOF

lucywyman:
**What this changes** If the readstream had reached EOF then select will still return the stream as available for reading, per http://man7.org/linux/man-pages/man2/select.2.html. We want to catch an EOF error if the stream can't be read and return an empty string to the node output.
**Why** This breaks when the user is not prompted for a password, for instance if the user doesn't have a password. Instead it should escalate the user and continue executing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,9 +17,6 @@ before_script:
 - cat Gemfile.lock
 - bundle exec r10k puppetfile install
   # Add users to test sudo on localhost
-- sudo groupadd bolt
-- sudo useradd -g bolt bolt
-- echo 'bolt:bolt' | sudo chpasswd
 - echo 'travis:travis' | sudo chpasswd
   # Undo travis sudoers config
 - sudo sh -c "echo 'Defaults authenticate' >> /etc/sudoers"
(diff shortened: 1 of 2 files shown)


lib/bolt/transport/local/shell.rb
```diff
@@ -103,6 +103,10 @@ def check_sudo(out, inp, pid)
           # handle_sudo will return the line if it is not a sudo prompt or error
           lines.map! { |line| handle_sudo(inp, line, pid) }
           lines.join("")
+        # If err has reached EOF, no password prompt is expected
```
adreyer: This could be out too right?

lucywyman: Yeah - I was thinking the sudo prompt only prints to err, but we run this check for both so it could be out that reached EOF. Good catch!

.travis.yml
```diff
@@ -17,9 +17,6 @@ before_script:
 - cat Gemfile.lock
 - bundle exec r10k puppetfile install
   # Add users to test sudo on localhost
-- sudo groupadd bolt
```
lucywyman: These are just leftover from previous testing and aren't used

PR #949, opened 2019-04-11
(BOLT-1205) Update documentation to clarify default ruby interpreter

donoghuc:


donoghuc:
Maybe we should just remove the reference to setting the default interpreter and have section about what to expect when your target name is `localhost`. Does that already exist? 

Either way, BOLT-1205 makes the claim that ruby tasks default to Bolt's ruby for `local` transport inaccurate.

lucywyman:
Ah, like a 'default localhost config' ? I think people are more likely to look at the specific config setting than the 'local transport' settings, so I think this change is good. Might be worth adding in the future if we add more default local configs.

melamos:
Added to docs.

donoghuc:
There is going to be a merge conflict with BOLT-843, so lets get that merged first and I will sort this one out.

melamos:
@lucywyman or @donoghuc, does the `localhost` update to writing_tasks.md apply to PE? The rest of that cross-platform task content is shared into PE 2019.0+.

lucywyman:
Yes, I think so.

melamos:
Updated in Bolt and PE 2019.0+ docs.

PR #950, opened 2019-04-11
(BOLT-126) Add WinRM Kerberos support

seanmil:
This is based on the work in #231, overhauled for the current version of Bolt.

### DIFF
--- a/lib/bolt/transport/winrm/connection.rb
+++ b/lib/bolt/transport/winrm/connection.rb
@@ -15,7 +15,7 @@ module Bolt
           raise Bolt::ValidationError, "Target #{target.name} does not have a host" unless target.host
           @target = target
 
-          default_port = target.options['ssl'] ? HTTPS_PORT : HTTP_PORT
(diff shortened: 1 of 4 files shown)


Iristyle:
Blocking on merge / release of WinRM gem based on bugfix at https://github.com/WinRb/WinRM/pull/302

MikaelSmith:
I anticipate this will need some docs around setting it up as well. Who's going to take on writing them?

lib/bolt/transport/winrm/connection.rb
```diff
@@ -31,20 +31,30 @@ def initialize(target, transport_logger)
         HTTPS_PORT = 5986
 
         def connect
-          if target.options['ssl']
+          if target.options['realm']
+            scheme = 'http'
+            transport = :kerberos
```
Iristyle: Will have to verify for sure, but Kerberos should actually work fine over HTTP or HTTPS, and we'll want to allow for both.

Iristyle:
Closing in favor of #999 - thanks a ton for the initial work on this @seanmil !

PR #951, opened 2019-04-11
(BOLT-1196) Update inaccuracies in bolt-inventory-pdb docs

adreyer:
This feature will probably be deprecated with v2 inventory work so I
don't think it's work adding a lot of documentation.


lucywyman:
@melamos  For review

melamos:
Added to docs.

PR #952, opened 2019-04-12
(BOLT-1242) Add acceptance test for validating local tasks

lucywyman:
**What this changes** This adds an acceptance test for running a task on the local transport which has `stdin` as an `input_method` but does not read from stdin. It also adds a test for running a task on the local transport as a user who doesn't require a password to escalate privileges and uses `_run_as` from a task.
**Why** Particularly the user-with-no-password case is difficult to test in Travis, since running a single test as a different user and setting up the user infrastructure in Travis is non-trivial. Especially since this is relatively low-risk we're adding a beaker acceptance test for it

### DIFF
--- a/acceptance/tests/task_local.rb
+++ b/acceptance/tests/task_local.rb
@@ -1,0 +1,105 @@
+# frozen_string_literal: true
+
+require 'bolt_command_helper'
+
+test_name "bolt tasks should run on the local transport" do
+  extend Acceptance::BoltCommandHelper
+
+  # From spec/bolt/transport/shared_examples
+  def posix_context
+    {
+      tmpdir: '/tmp',
+      supported_req: 'shell',
+      extension: '.sh',
+      env_task: "#!/bin/sh\nprintenv PT_message_one\nprintenv PT_message_two",
+      stdin_task: "#!/bin/sh\ncat"
+    }
+  end
+
+  def windows_context
+    {
+      tmpdir: 'C:/mytmp',
+      supported_req: 'powershell',
+      extension: '.ps1',
+      env_task: "Write-Output \"${env:PT_message_one}\n${env:PT_message_two}\"",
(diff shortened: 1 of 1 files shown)


donoghuc:
I think the fastest/most reliable way to validate this is to spin up experimental pipeline. That will allow us to easily iterate test in actual environtment. I put up https://github.com/puppetlabs/ci-job-configs/pull/5746 and intend to get experimental pipeline set up.

acceptance/tests/task_local.rb
```diff
@@ -0,0 +1,105 @@
...
+      assert_match(/Successful on 1 node: localhost/, result.stdout, "Unexpected output from the command:\n#{result.cmd}")
+      assert_match(/beep\n  boop/, result.stdout, "Unexpected output from the command:\n#{result.cmd}")
+      assert(result.exit_code == 0, "#{result.cmd} exited #{result.exit_code} with stderr: #{result.stderr}")
+    end
+  end
+
+  unless Bolt::Util.windows?
+    step 'as a user without a password' do
+      step 'create user who can escalate without password' do
+        on(bolt, "/opt/puppetlabs/bolt/bin/puppet resource user #{user} ensure=present")
```
donoghuc: I would just use beaker for this. `apply_manifest_on` maybe.

donoghuc: https://jenkins-platform.delivery.puppetlabs.net/job/experimental_auto_bolt_vanagon-integration-component_maint-no-password-bug_project-master/RMM_COMPONENT_TO_TEST_NAME=bolt,SLAVE_LABEL=beaker,TEST_TARGET=osx1013-64bolt%252Cssh.-centos7-64ssh.-windows2008-64winrm./1/testReport/junit/(root)/tests/task_local_rb/

lucywyman:
Closing in favor of #971

PR #954, opened 2019-04-15
(BOLT-1194) Validate simple plans

donoghuc:
When a plan is loaded the following validation occurs.
1. All required keys for all steps are present
2. No unexpected keys are allowed in plan steps or parameters
3. All puppet code is verified to be syntactically correct

### DIFF
--- a/lib/bolt/pal/yaml_plan.rb
+++ b/lib/bolt/pal/yaml_plan.rb
@@ -3,6 +3,40 @@
 module Bolt
   class PAL
     class YamlPlan
+      PLAN_KEYS = Set['parameters', 'steps', 'return', 'version']
+      PARAMETER_KEYS = Set['type', 'default']
+      COMMON_STEP_KEYS = %w[name description target].freeze
+      STEP_KEYS = {
+        'command' => {
+          'allowed_keys' => Set['command'].merge(COMMON_STEP_KEYS),
+          'required_keys' => Set['target']
+        },
+        'script' => {
+          'allowed_keys' => Set['script', 'parameters', 'arguments'].merge(COMMON_STEP_KEYS),
+          'required_keys' => Set['target']
+        },
+        'task' => {
+          'allowed_keys' => Set['task', 'parameters'].merge(COMMON_STEP_KEYS),
+          'required_keys' => Set['target']
+        },
+        'plan' => {
+          'allowed_keys' => Set['plan', 'parameters'].merge(COMMON_STEP_KEYS),
+          'required_keys' => Set.new
+        },
+        'source' => {
+          'allowed_keys' => Set['source', 'destination'].merge(COMMON_STEP_KEYS),
+          'required_keys' => Set['target', 'source', 'destination']
+        },
+        'destination' => {
+          'allowed_keys' => Set['source', 'destination'].merge(COMMON_STEP_KEYS),
+          'required_keys' => Set['target', 'source', 'destination']
+        },
+        'eval' => {
+          'allowed_keys' => Set['eval', 'name', 'description'],
+          'required_keys' => Set.new
+        }
+      }.freeze
+
       Parameter = Struct.new(:name, :value, :type_expr) do
         def captures_rest
           false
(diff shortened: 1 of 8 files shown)


lib/bolt/pal/yaml_plan.rb
```diff
@@ -3,6 +3,35 @@
...
+        },
+        'task' => {
+          'allowed_keys' => Set['task', 'parameters'].merge(COMMON_STEP_KEYS),
+          'required_keys' => Set['target']
+        },
+        'plan' => {
+          'allowed_keys' => Set['plan', 'parameters'].merge(COMMON_STEP_KEYS),
+          'required_keys' => Set.new
+        },
+        'source' => {
```
donoghuc: The upload step is the only step that does not have a single key that can identify it (must have both `source` and `destination`). As WIP i ignore this and identify it on `source`. Will need to update the logic.

lib/bolt/pal/yaml_plan.rb
```diff
@@ -103,6 +139,66 @@ def return_type
...
+                                  "bolt/invalid-plan")
+          else
+            raise Bolt::Error.new("No action specified for step", "bolt/invalid-plan")
+          end
+        end
+
+        # For validated step action, ensure only valid keys
+        unless STEP_KEYS[action.first]['allowed_keys'].superset?(step_keys)
+          illegal_keys = step_keys - STEP_KEYS[action.first]['allowed_keys']
+          raise Bolt::Error.new("The '#{action.first}' step does not support: '#{illegal_keys.to_a.join(',')}' key(s)",
```
donoghuc: I need to figure out consistent error reporting. Specifically how we point to *where* the error is found.  
The first thing that popped in to my mind was identifying steps by number (because steps are not required to have `name`).
It would be most useful as a user if we could point to a line number instead of a step number/name, but I do not see how that will be preserved with existing yaml parsing.

lib/bolt/pal/yaml_plan.rb
adreyer: we could keep this inner hash separate and have both source and destination point to it? That may be more complex in other ways but it seems like we should give the same error regardless of which special key is missing.

donoghuc: Went with having source and destination map to the a copy of a hash with the same info. The only added complexity I can tell it adds validating only a single action is specified per step.

lib/bolt/pal/yaml_plan.rb
donoghuc: Preserving line numbers is not worth it just for reporting validation. We can revisit this if there is demand. For now we will *always* reference step number and *attempt* to reference step name.

spec/fixtures/modules/yaml/plans/upload.yaml
```diff
@@ -7,9 +7,8 @@ steps:
     source: 'yaml/test.sh'
     destination: '/tmp/test_upload.sh'
     target: $nodes
-    arguments: ["foo", "bar", "baz"]
   - name: cleanup
-    eval: rm '/tmp/test_upload.sh'
+    command: rm '/tmp/test_upload.sh'
```
donoghuc: I changed this with the following assumptions: 
1. Upload step does accept arguments
2. Eval does not accept targets

I think these were just copy paste errors?

lib/bolt/pal/yaml_plan.rb
```diff
@@ -60,15 +99,25 @@ def validate
         end
 
         @steps.each do |step|
+          validate_step_keys(step, step_number)
+
+          begin
+            step.each { |k, v| validate_puppet_code(k, v) }
+          rescue Bolt::Error => e
```
donoghuc: I still need to think more about this. I want to be able to report `step_number` and `step['name']` in error message *in addition to* the key that does not parse with it's associated puppet parser error message. Instead of carrying the `step_number` and `step['name']` through every recursive call as a static parameters in the `validate_puppet_code`  method, I catch the puppet error, convert it to bolt error with the info I want and then can catch it and report here. I guess if I stick with this I should raise  a more specific error in `validate_puppet_code`.

adreyer:
it looks like I still get an error if parameters is nil.

lib/bolt/pal/yaml_plan/evaluator.rb
```diff
@@ -17,12 +17,9 @@ def initialize(analytics = Bolt::Analytics::NoopClient.new)
         def dispatch_step(scope, step)
           step = evaluate_code_blocks(scope, step)
 
-          step_type, *extra_keys = STEP_KEYS.select { |key| step.key?(key) }
-          if !step_type || extra_keys.any?
-            unsupported_step(scope, step)
-          end
+          step_type = STEP_KEYS.select { |key| step.key?(key) }
```
adreyer: `.select.first` => `.find`

donoghuc: Updated. I remember thinking that looked clunky. Never got around to investigating. Thanks for enlightening me.

PR #955, opened 2019-04-15
(BOLT-901) Rework output based on an event stream

nicklewis:
This pull request overhauls the way that user-facing output and log output are generated. Previously, the logger and the outputter split the responsibility for generating output that was meant for the user. When running a task/command/etc with human output, the outputter was responsible for printing progress information. But when running a plan, the outputter was only responsible for printing the final result, with progress information coming from the logger.

This made the overall output a patchwork of different functionality that
all happened to come together in a coherent result. Making any changes
to the output without breaking some cases was difficult.

The outputter and the logger now have a clear division of responsibility:
* The outputter is responsible for printing info meant for the user: progress information and results
* The logger is responsible for printing warnings and errors messages,  and for emitting _log-structured_ info to the log file

We accomplish this through an event stream emitted by the executor. We emit events for the start and end of plans, steps, and individual nodes. The individual outputter is now responsible for tracking state to decide whether "plan logging" is enabled and whether verbose mode is turned on, and is responsible for choosing which events to output.

"Verbose" is now defined as an output-only concept, and has no effect on
the logger. "Verbose" means that node-level results will be printed when using the human outputter. We now have both a --verbose and a --no-verbose option to turn verbosity on or off. Plans do default to no-verbose; every other subcommand is verbose by default.

Logging is now accomplished by an "outputter" that attaches to the event stream and converts some events to log messages.

We now include log information for `apply` results. Previously, we only ever printed a summary of the events of an `apply`, and didn't provide any information about what actually changed or
failed. We now print any log messages associated with resources (explicitly omitting system messages, those where the source is "Puppet") at or above notice level. We also print those logs when the result of a plan execution contains an apply result. This will help users know what specifically changed or failed when using either the `bolt apply` command or `apply` from a plan.


lib/bolt/outputter/logger.rb
```diff
@@ -0,0 +1,53 @@
...
+
+      def log_step_finish(description:, result:, duration:, **_kwargs)
+        failures = result.error_set.length
+        plural = failures == 1 ? '' : 's'
+        @logger.info("Finished: #{description} with #{failures} failure#{plural} in #{duration.round(2)} sec")
+      end
+
+      def log_plan_start(event)
+        plan = event[:plan]
+        @logger.notice("Starting: plan #{plan}")
```
donoghuc: When the console log level is `notice` start plan and end plan gets printed twice. Is that something we care about trying to avoid? Or would it be expected behavior? (other place it is printed) https://github.com/puppetlabs/bolt/pull/955/files#diff-505c7bb20923cd2e1fabc36fa57b2a51R151

donoghuc: This is expected behavior. Note that it does not occur with the default console log level of `warn`.

PR #956, opened 2019-04-16
(BOLT-1231) Switch V2 inventory to use 'targets' instead of 'nodes.

adreyer:
Moving forward bolt will standardize on calling targets targets instead
of nodes to avoid confusion about remote targets. This commit causes the
version 2 inventory file to look in `targets` instead of `nodes` for the
members of a group.

### DIFF
--- a/lib/bolt/inventory/inventory2.rb
+++ b/lib/bolt/inventory/inventory2.rb
@@ -5,6 +5,13 @@ require 'bolt/inventory/group2'
 module Bolt
   class Inventory
     class Inventory2
+      # This uses "targets" in the message instead of "nodes"
+      class WildcardError < Bolt::Error
+        def initialize(target)
+          super("Found 0 targets matching wildcard pattern #{target}", 'bolt.inventory/wildcard-error')
+        end
+      end
+
       def initialize(data, config = nil, target_vars: {}, target_facts: {}, target_features: {})
         @logger = Logging.logger[self]
         # Config is saved to add config options to targets
@@ -16,7 +23,7 @@ module Bolt
         @target_facts = target_facts
         @target_features = target_features
 
-        @groups.resolve_aliases(@groups.node_aliases, @groups.node_names)
+        @groups.resolve_aliases(@groups.target_aliases, @groups.target_names)
         collect_groups
       end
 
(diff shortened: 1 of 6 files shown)


lib/bolt/inventory/inventory2.rb
```diff
@@ -208,12 +217,15 @@ def set_facts(tname, hash)
...
 
-      def add_node(current_group, target, desired_group, track = { 'all' => nil })
+      def add_target(current_group, target, desired_group, track = { 'all' => nil })
         if current_group.name == desired_group
           # Group to add to is found
           t_name = target.name
-          # Add target to nodes hash
-          current_group.nodes[t_name] = { 'name' => t_name }.merge(target.options)
+          # Add target to targets hash
+          target_hash = { 'name' => t_name }.merge(target.options)
```
adreyer: This change fixes a bug previously introduced where adding a new target to a group using just a string name caused the URI to not be set.

spec/integration/inventory2_spec.rb
```diff
@@ -239,11 +240,7 @@ def fact_plan(name = 'facts_test')
           },
           {
             name: 'bar',
-            nodes: [
-              {
-                name: 'bar_1'
-              }
-            ],
+            targets: [],
```
adreyer: The test adds a string node `bar_1` to the `bar` group.  By having it already present in the group it seems like it wasn't really testing what was expected.

donoghuc: I think I just did a bad job describing/writing that test. What that **should** be testing is *This should do nothing if the user adds the target to a group it is already a part of.* [BOLT-942](https://tickets.puppetlabs.com/browse/BOLT-942). So, given `bar_1` in group `bar` adding a target `bar_1` does not override, or duplicate `bar_1`. Though I guess the test as is written does not really prove that the `bar_1` defined in inventory has been overridden. 

This test could certainly be improved. (I see an unused variable in the plan, plus the confusing/wrong descriptions). I can work on that as a maint PR against the inventory v1 integration test and then port those changes over as separate work.

donoghuc: https://github.com/puppetlabs/bolt/pull/958

adreyer:
I think we need to add a concept of a "short name" or "safe name" to targets themselves to solve those issues. I'll ticket that separately

PR #957, opened 2019-04-17
(BOLT-1195) Convert a YAML plan to a Puppet plan

lucywyman:
This creates a new command `bolt plan convert <path-to-plan>` which accepts either a relative or absolute path a YAML plan and prints to stdout the equivalent Puppet plan.


lucywyman:
This relies on https://github.com/puppetlabs/puppet/pull/7485

donoghuc:
Would be helpful to (temporarily) commit your Gemfile pointing to the change in puppet.

lucywyman:
Travis failures are just rubocop on the Gemfile :P

lib/bolt/transpiler.rb
```diff
@@ -0,0 +1,125 @@
...
+      end
+
+      # TODO: probably a better way to do this?
+      basename = File.basename(@filename, '.yaml')
+      fake_typed_name = if basename == 'init'
+                          @module
+                        else
+                          [@module, File.basename(@filename, '.yaml')].join('::')
+                        end
+      Bolt::PAL::YamlPlan.new(fake_typed_name, result).freeze
```
donoghuc: After https://github.com/puppetlabs/bolt/pull/954 I think some of the validation in the `Transpiler` code will be redundant. (for example random step keys, etc).

lib/bolt/transpiler.rb
```diff
@@ -0,0 +1,125 @@
...
+
+        target = step['target']
+        description = step.fetch('description', nil)
+        parameters = step.fetch('parameters', {})
+        if step_type == 'script' && step.key?('arguments')
+          parameters['arguments'] = step['arguments']
+        end
+
+        case step_type
+        when 'command', 'task', 'script', 'plan'
```
donoghuc: In general I think an erb template might make this easier than constructing one giant string.

lib/bolt/transpiler.rb
lucywyman: YAY awesome. I kind of assumed *most* of the checking would be done there (like that targets is in every step).

lib/bolt/error.rb
```diff
@@ -127,4 +127,10 @@ def initialize(msg, path)
       super(msg, 'bolt/file-error', { "path" => path })
     end
   end
+
+  class ConvertError < Bolt::Error
```
adreyer: In general we've been adding these into the classes that interact with them rather then further expanding this file.

lib/bolt/transpiler.rb
```diff
@@ -0,0 +1,125 @@
...
+        last = i + 1 == plan_object.parameters.length ? "\n" : ","
+        # This encodes strangely if we << directly to plan_string
+        plan_string << last
+      end
+      plan_string << ") {\n"
+
+      plan_object.steps&.each do |step|
+        step.transform_values! { |v| plan_object.stringify(v) }
+        plan_string << "$#{step['name']} = " if step['name']
+        step_type, *extra_keys = Bolt::PAL::YamlPlan::Evaluator::STEP_KEYS.select { |key| step.key?(key) }
```
adreyer: The fact that this has to reach so deep into another namespace is a good sign this code belongs in that namespace.

spec/bolt/transpiler_spec.rb
```diff
@@ -0,0 +1,56 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+require 'bolt/transpiler'
+
+describe Bolt::YamlTranspiler do
+  let(:basepath) { File.join(File.dirname(__FILE__), '..', 'fixtures', 'modules', 'yaml', 'plans') }
+  let(:plan) { described_class.transpile(plan_path) }
```
adreyer: It can result in confusing errors to declare a let without declaring it's dependencies.

lib/bolt/pal/yaml_plan.rb
```diff
@@ -1,10 +1,14 @@
...
+require 'puppet_pal'
+require 'bolt/pal/yaml_plan/parameter'
+require 'bolt/pal/yaml_plan/step'
+
 module Bolt
   class PAL
     class YamlPlan
       PLAN_KEYS = Set['parameters', 'steps', 'return', 'version']
-      PARAMETER_KEYS = Set['type', 'default']
+      PARAMETER_KEYS = Set['type', 'default', 'description']
```
donoghuc: I did not realize this was an allowable key. I put up a separate PR https://github.com/puppetlabs/bolt/pull/968 to get in to release without being blocked on this PR.

lib/bolt/pal/yaml_plan.rb
```diff
@@ -36,23 +39,16 @@ class YamlPlan
...
-        end
-      end
-
       attr_reader :name, :parameters, :steps, :return
 
       def initialize(name, plan)
         # Top-level plan keys aren't allowed to be Puppet code, so force them
         # all to strings.
-        plan = Bolt::Util.walk_keys(plan) { |key| stringify(key) }
+        plan = Bolt::Util.walk_keys(plan) { |key| self.class.stringify(key) }
```
adreyer: this is an anti-pattern since it won't be available on subclasses. It's also another hint that the transpiler should be more closely related to the yaml plan.

lib/bolt/pal/yaml_plan.rb
```diff
@@ -117,23 +110,23 @@ def validate
           validate_step_keys(step, step_number)
 
           begin
-            step.each { |k, v| validate_puppet_code(k, v) }
+            step.body.each { |k, v| validate_puppet_code(k, v) }
```
adreyer: This seems like this validation should move to the step now that it's it's own object.

lib/bolt/transpiler.rb
```diff
@@ -0,0 +1,68 @@
...
+module Bolt
+  class YamlTranspiler
+    class ConvertError < Bolt::Error
+      def initialize(msg, plan_path)
+        super(msg, 'bolt/convert-error', { "plan_path" => plan_path })
+      end
+    end
+
+    def self.transpile(plan_path)
+      @plan_path = plan_path
```
adreyer: It's almost always wrong to set an instance variable from a class method.

lib/bolt/transpiler.rb
```diff
@@ -0,0 +1,68 @@
...
+      @module = Bolt::Util.module_name(plan_path)
```
adreyer: This should probably expand the plan_path first. It currently fails if you try to run it from inside a module.

adreyer:
it looks like puppet/pal is not being set up properly
```
bundle exec bolt plan convert yaml/plans/foo.yaml
bundler: failed to load command: bolt (/Users/adreyer/src/bolt/.bundle/ruby/2.5.0/bin/bolt)
NameError: uninitialized constant Bolt::PAL::YamlPlan::Puppet
Did you mean?  Bolt::PuppetDB
  /Users/adreyer/src/bolt/lib/bolt/pal/yaml_plan.rb:77:in `block in initialize'
  /Users/adreyer/src/bolt/lib/bolt/pal/yaml_plan.rb:68:in `each'
  /Users/adreyer/src/bolt/lib/bolt/pal/yaml_plan.rb:68:in `map'
  /Users/adreyer/src/bolt/lib/bolt/pal/yaml_plan.rb:68:in `initialize'
  /Users/adreyer/src/bolt/lib/bolt/transpiler.rb:59:in `new'
  /Users/adreyer/src/bolt/lib/bolt/transpiler.rb:59:in `parse_plan'
  /Users/adreyer/src/bolt/lib/bolt/transpiler.rb:22:in `transpile'
  /Users/adreyer/src/bolt/lib/bolt/cli.rb:275:in `execute'
  /Users/adreyer/src/bolt/exe/bolt:10:in `<top (required)>'
  /Users/adreyer/src/bolt/.bundle/ruby/2.5.0/bin/bolt:23:in `load'
  /Users/adreyer/src/bolt/.bundle/ruby/2.5.0/bin/bolt:23:in `<top (required)>'
```

lucywyman:
Re: Templating: I see a few negatives to templating without any gain over building a string. 1. It's less readable, especially if like me you're not familiar with it 2. It's trickier to support inline logic, like "only add a parenthesis here if there are no parameters otherwise add a newline" all on one line is very long, 3. It adds a new tool to learn (granted it's ruby-ish, but it's still adding complexity). I didn't think it was worth exploring further, but if there are advantages I'm not considering I'm happy to listen.

nicklewis:
It looks like strings are always being emitted in single quotes. We should be checking whether it's an `EvaluableString` and switching based on that. Strings which were double-quoted in the YAML should be double-quoted in the output.

donoghuc:
WIthout having looked at code and just playing with this I see a couple things. 

1. No `convert` description in bolt help
```
cas@cas-ThinkPad-T460p:~/working_dir/bolt$ bolt
Usage: bolt <subcommand> <action> [options]

Available subcommands:
  bolt command run <command>       Run a command remotely
  bolt file upload <src> <dest>    Upload a local file
  bolt script run <script>         Upload a local script and run it remotely
  bolt task show                   Show list of available tasks
  bolt task show <task>            Show documentation for task
  bolt task run <task> [params]    Run a Puppet task
  bolt plan show                   Show list of available plans
  bolt plan show <plan>            Show details for plan
  bolt plan run <plan> [params]    Run a Puppet task plan
  bolt apply <manifest>            Apply Puppet manifest code
  bolt puppetfile install          Install modules from a Puppetfile into a Boltdir
  bolt puppetfile show-modules     List modules available to Bolt
```
2. Having a step description can result in double quoting in arg list
```
parameters:
  nodes:
    type: TargetSpec
  foo: 
    type: String[1]
    description: foo
    default: bar
steps:
  - name: run_task
    task: sample
    target: $nodes
    description: hi
    parameters:
      message: hello world
return: $run_task
```
```
cas@cas-ThinkPad-T460p:~/working_dir/bolt$ bolt plan convert Boltdir/site/yaml_plans/plans/test_convert.yaml
# WARNING: This is an autogenerated plan. It may not behave as expected.
plan yaml_plans::test_convert(
  TargetSpec $nodes,
  String[1] $foo = 'bar'
) {
$run_task = run_task('sample', $nodes, ''hi'', {'message' => 'hello world'})

return $run_task
}
The converted puppet plan contains invalid puppet code: Syntax error at 'hi' (line: 6, column: 42)
```
3. Eval can be used to have duplicate variable assignment that is not caught. 
```
parameters:
  nodes:
    type: TargetSpec
  foo: 
    type: String[1]
    description: foo
    default: bar
steps:
  - name: run_task
    task: sample
    target: $nodes
    parameters:
      message: hello world
  - eval: |
      $run_task = $foo
    name: eval_step

return: $run_task
```
```
cas@cas-ThinkPad-T460p:~/working_dir/bolt$ bolt plan convert Boltdir/site/yaml_plans/plans/test_convert.yaml
# WARNING: This is an autogenerated plan. It may not behave as expected.
plan yaml_plans::test_convert(
  TargetSpec $nodes,
  String[1] $foo = 'bar'
) {
$run_task = run_task('sample', $nodes, {'message' => 'hello world'})
$eval_step = $run_task = $foo


return $run_task
}
```

donoghuc:
Also I think that the warnings/errors about invalid puppet are getting printed to both stdout and stderr.

lucywyman:
@donoghuc:
1. I made adding the help text part of #977 for docs review
2. Woops! Removed the extra quotes.
3. I'm not sure what to do about that...I run the plan through the evaluating parser once it's converted, perhaps there's another puppet evaluator I need to use?
4. Investinating...

lib/bolt/pal/yaml_plan/transpiler.rb
```diff
@@ -0,0 +1,83 @@
...
+        end
+
+        def validate_path
+          unless File.extname(@filename) == ".yaml"
+            raise ConvertError.new("You can only convert plans written in yaml", @plan_path)
+          end
+        end
+
+        def validate_plan(plan)
+          Puppet::Pops::Parser::EvaluatingParser.new.parse_string(plan)
```
nicklewis: We should also be validating the AST.

```rb
parser = Puppet::Pops::Parser::EvaluatingParser.new
ast = parse.parse_string(plan)
parser.validate(ast)
```

That will only work if `Puppet[:tasks]` is set, so we need to make sure this whole conversion process is run inside a Bolt compiler environment (via a call to `in_bolt_compiler` or `in_plan_compiler`).

nicklewis:
Conflicting variable assignments are explicitly one of the things we _can't_ detect, and are indeed a 
key reason we emit the warning about how the plan may not be correct.

lib/bolt/pal/yaml_plan/transpiler.rb
lucywyman: I set `Puppet[:tasks]` manually here https://github.com/puppetlabs/bolt/pull/957/files#diff-a0f737621da3fb6ead05b549796d7406R301 - is there other context that `in_plan_compiler` will provide that's useful? Or is it more clear to use that? It just seemed easier to set `Puppet[:tasks]` for converting the plan manually rather than wrapping it in a block.

nicklewis: I think that's fine.

lib/bolt/util.rb
```diff
@@ -35,6 +35,73 @@ def read_config_file(path, default_paths = nil, file_name = 'file')
...
+        end
+
+        # The plan or task name is the rest of the path
+        [modulename, parts[1].split(fs)].flatten.join('::')
+      end
+
+      def to_code(string)
+        case string
+        when Bolt::PAL::YamlPlan::DoubleQuotedString
+          "\"#{string.value}\""
```
nicklewis: I think this line should be `string.value.inspect`. That will surround the string with double quotes _and_ it will escape any embedded quotes, etc.

lib/bolt/util.rb
```diff
@@ -35,6 +35,73 @@ def read_config_file(path, default_paths = nil, file_name = 'file')
...
+          formatted = String.new("{")
+          string.each do |k, v|
+            formatted << "#{to_code(k)} => #{to_code(v)}, "
+          end
+          formatted.chomp!(", ")
+          formatted << "}"
+          formatted
+        when Array
+          formatted = String.new("[")
+          string.each_with_index do |str, i|
```
nicklewis: I think you can do `formatted << string.map { |str| to_code(str) }.join(', ')`

lucywyman:
@donoghuc I think I've addressed all your comments - using `with() ||` seems to work for me for the eval code block use case. @nicklewis I think I'm going to head out for the night, but will make those changes first thing tomorrow!

donoghuc:
I thought I remembered talking about printing the error/warning when we detect invalid puppet code to stderr only (thought i dont see that in the ticket)? I see that it is exiting with exit code 1 but it seems like the error is still being printed to stdout.

lib/bolt/pal/yaml_plan/step.rb
```diff
@@ -0,0 +1,190 @@
...
+          if @type == 'script' && body.key?('arguments')
+            parameters['arguments'] = body['arguments']
+          end
+
+          case @type
+          when 'command', 'task', 'script', 'plan'
+            result << "run_#{@type}(#{Bolt::Util.to_code(body[@type])}"
+            result << ", #{Bolt::Util.to_code(@target)}" if @target
+            result << ", #{Bolt::Util.to_code(description)}" if description
+            result << ", #{Bolt::Util.to_code(parameters)}" unless parameters.empty?
```
donoghuc: I think we are missing a validation here that parameters is in fact an array. If it is not calling `.empty?` will stack trace.

donoghuc:
Do we need to give parameters the `with()` treatment like for `eval`?
```
parameters:
  nodes:
    type: TargetSpec
  x:
    type: Integer
  y:
    type: Integer
steps:
  - name: run_task
    task: sample::echo
    description: "foo"
    target: localhost
    parameters:
      message: |
        $z = $x + $y
        $z + 1
return: $run_task
```
```
cas@cas-ThinkPad-T460p:~/working_dir/bolt$ bolt plan run yaml_plans::test_convert x=1 y=1 -t localhost
Starting: plan yaml_plans::test_convert
Starting: foo on localhost
Finished: foo with 0 failures in 0.01 sec
Finished: plan yaml_plans::test_convert in 0.01 sec
Finished on localhost:
  cas-ThinkPad-T460p got passed the message: 3
  {
  }
Ran on 1 node
```
```
cas@cas-ThinkPad-T460p:~/working_dir/bolt$ bolt plan convert Boltdir/site/yaml_plans/plans/test_convert.yaml
# WARNING: This is an autogenerated plan. It may not behave as expected.
plan yaml_plans::test_convert(
  TargetSpec $nodes,
  Integer $x,
  Integer $y
) {
  $run_task = run_task('sample::echo', 'localhost', "foo", {'message' => $z = $x + $y
$z + 1
})

  return $run_task
}
The converted puppet plan contains invalid puppet code: Syntax error at 'z' (line: 8, column: 1)

```

lib/bolt/pal/yaml_plan/step.rb
lucywyman: Do you have a plan where this happens? Parameters should get set to an empty hash by default, and we have a check in the yaml plan that it's a hash...

donoghuc: I think somehow we are just not validating that parameters is a hash. 
```
parameters:
  nodes:
    type: TargetSpec
steps:
  - name: run_task
    task: sample::echo
    description: "foo"
    target: localhost
    parameters: foo
```
```
cas@cas-ThinkPad-T460p:~/working_dir/bolt$ bolt plan convert Boltdir/site/yaml_plans/plans/test_convert.yaml 
bundler: failed to load command: bolt (/home/cas/working_dir/bolt/.bundle/ruby/2.5.0/bin/bolt)
NoMethodError: undefined method `empty?' for #<Bolt::PAL::YamlPlan::BareString:0x000055fe834b08e8>
  /home/cas/working_dir/bolt/lib/bolt/pal/yaml_plan/step.rb:69:in `transpile'
  /home/cas/working_dir/bolt/lib/bolt/pal/yaml_plan/transpiler.rb:41:in `block in transpile'
  /home/cas/working_dir/bolt/lib/bolt/pal/yaml_plan/transpiler.rb:39:in `each'
  /home/cas/working_dir/bolt/lib/bolt/pal/yaml_plan/transpiler.rb:39:in `transpile'
  /home/cas/working_dir/bolt/lib/bolt/pal.rb:303:in `convert_plan'
  /home/cas/working_dir/bolt/lib/bolt/cli.rb:465:in `convert_plan'
  /home/cas/working_dir/bolt/lib/bolt/cli.rb:283:in `execute'
  /home/cas/working_dir/bolt/exe/bolt:10:in `<top (required)>'
  /home/cas/working_dir/bolt/.bundle/ruby/2.5.0/bin/bolt:23:in `load'
  /home/cas/working_dir/bolt/.bundle/ruby/2.5.0/bin/bolt:23:in `<top (required)>'
```

lib/bolt/pal/yaml_plan/step.rb
```diff
@@ -0,0 +1,190 @@
...
+          parameters = body.fetch('parameters', {})
+          if @type == 'script' && body.key?('arguments')
+            parameters['arguments'] = body['arguments']
+          end
+
+          case @type
+          when 'command', 'task', 'script', 'plan'
+            result << "run_#{@type}(#{Bolt::Util.to_code(body[@type])}"
+            result << ", #{Bolt::Util.to_code(@target)}" if @target
+            result << ", #{Bolt::Util.to_code(description)}" if description
```
donoghuc: I dont think `run_plan` accepts a description. 
```
steps:
  - name: run_plan
    plan: sample::single_task
    parameters:
      nodes: localhost
    description: foo

return: run_plan
```
```
cas@cas-ThinkPad-T460p:~/working_dir/bolt/Boltdir$ bolt plan convert site/yaml_plans/plans/nested/nested_yaml.yaml
# WARNING: This is an autogenerated plan. It may not behave as expected.
plan yaml_plans::nested::nested_yaml() {
  $run_plan = run_plan('sample::single_task', 'foo', {'nodes' => 'localhost'})

  return 'run_plan'
}
```
```
cas@cas-ThinkPad-T460p:~/working_dir/bolt/Boltdir$ bolt plan run yaml_plans::nested::nested_yaml_pp
Starting: plan yaml_plans::nested::nested_yaml_pp
Finished: plan yaml_plans::nested::nested_yaml_pp in 0.0 sec
{
  "kind": "bolt/pal-error",
  "msg": "'run_plan' expects between 1 and 2 arguments, got 3",
  "details": {
  }
}
```

lib/bolt/pal/yaml_plan.rb
```diff
@@ -67,82 +37,39 @@ def initialize(name, plan)
...
-
-            if used_names.include?(step['name'])
-              error_message = "Duplicate step name or parameter detected: #{step['name'].inspect}"
-              raise Bolt::Error.new(step_err_msg(step_number, step['name'], error_message), "bolt/invalid-plan")
-            end
-
-            used_names << step['name']
-          end
-          step_number += 1
+      def duplicate_check(used_names, name, index)
```
donoghuc: This does not seem to be detecting duplicate step names.

lucywyman: Ugh, that might actuallly be a hard fix. Do you have the plan that's failing?

donoghuc: Should be any plan with duplicate step names. Is it not failing for you?

lucywyman:
Re: Code blocks, I made https://tickets.puppetlabs.com/browse/BOLT-1285 - I'm going to resolve to comment just since it's kind of long.

lib/bolt/pal/yaml_plan/step.rb
lucywyman: Oh, I see, we're validating *plan* parameters not step parameters.

lib/bolt/pal/yaml_plan/transpiler.rb
```diff
@@ -0,0 +1,93 @@
...
+          plan_string << "\n  return #{Bolt::Util.to_code(plan_object.return)}\n" if plan_object.return
+          plan_string << "}"
+          # We always print the plan, even if there's an error
+          puts plan_string
+          validate_plan(plan_string)
+          plan_string
+        end
+
+        def parse_plan
+          file_contents = File.read(@plan_path)
```
donoghuc: This will stack trace if path doesn't exists.

lucywyman: haha this function is 100% begin..rescues now....

lib/bolt/cli.rb
```diff
@@ -279,6 +279,11 @@ def execute(options)
         return 0
       end
```
donoghuc: We should not load  inventory for plan convert. We should add this logic to https://github.com/puppetlabs/bolt/blob/d1aed20d51b3d6a85a312dd2c33cebe76adcfe18/lib/bolt/cli.rb#L122-L124 (maybe something a bit cleaner than just adding another `or` condition). 

Similarly we could be sneaky for now (before we start collecting analytics on this) and avoid loading inventory for analytics (and building analytics client all together) my moving the logic defined in line 282-286 to https://github.com/puppetlabs/bolt/blob/10e9bfc5a8fa8bb315418e99fbb70934f901c126/lib/bolt/cli.rb#L246

lib/bolt/pal/yaml_plan/transpiler.rb
```diff
@@ -0,0 +1,89 @@
...
+          validate_plan(plan_string)
+          plan_string
+        end
+
+        # Save me from all these rescue statements...
+        def parse_plan
+          begin
+            file_contents = File.read(@plan_path)
+          rescue Error => e
+            raise ConvertError, "Could not read file at #{@plan_path}: #{e.message}"
```
donoghuc: This should be `ConvertError.new` but it seems to still raise an error and stack trace here...

lib/bolt/cli.rb
lucywyman: This should be part of the acceptance criteria in the future.

lib/bolt/pal/yaml_plan/transpiler.rb
donoghuc: Actually I think that in this context `Error` resolves to a `Bolt::Error`. I think you want something like:
```ruby
          begin
            file_contents = File.read(@plan_path)
          rescue Errno::ENOENT
            msg = "Could not read yaml plan file: #{@plan_path}"
            raise Bolt::FileError.new(msg, @plan_path)
          end
```

nicklewis: This syntax will instantiate `ConvertError` with the string as the message.

lib/bolt/pal/yaml_plan/parameter.rb
```diff
@@ -0,0 +1,62 @@
...
+          definition_keys = definition.keys.to_set
+          unless PARAMETER_KEYS.superset?(definition_keys)
+            invalid_keys = definition_keys - PARAMETER_KEYS
+            raise Bolt::Error.new("Plan parameter #{param.inspect} contains illegal key(s)" \
+                                  " #{invalid_keys.to_a.inspect}",
+                                  "bolt/invalid-plan")
+          end
+        end
+
+        def captures_rest
```
donoghuc: What does this do?

lucywyman: I'm not sure - it was already part of the Parameter struct. @nicklewis might know

nicklewis: This is necessary to let our fake `Parameter` class masquerade as a real Puppet parameter. I think `captures_rest` is basically equivalent to a varargs type parameter in a Puppet function. Plans don't actually use positional arguments, so it's not a meaningful concept for them.

PR #958, opened 2019-04-17
(maint) Fix inventory add_to_group integration test

donoghuc:
Previously the test to verify that `add_to_group` does not overwrite existing targets in inventory was mislabled and did not actually test the functionality. This commit adds more descriptive names. It updates the tests such that when a target exists in inventory and has an associated var, adding a target with the same name from a plan using the `add_to_group` function does not overwrite or try to duplicate the target.

### DIFF
--- a/spec/fixtures/modules/add_group/plans/init.pp
+++ b/spec/fixtures/modules/add_group/plans/init.pp
@@ -5,26 +5,29 @@ plan add_group (TargetSpec $nodes) {
     "user" => 'bolt',
     'password' => 'bolt'
   )
-  $t = get_targets('add_me')
+
   # Add new facts/var, specifically one that is novel, the other that should override existing.
   $new_target.add_facts({'plan_context' => 'keep', 'override_parent' => 'keep'})
   $new_target.set_var('plan_context','keep')
   $new_target.set_var('override_parent', 'keep')
   # Add from Target object
   $new_target.add_to_group('add_me')
-  # Add from string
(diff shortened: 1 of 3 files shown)


spec/fixtures/modules/add_group/plans/init.pp
```diff
@@ -23,7 +23,8 @@
 			  'existing_vars' => $add_me_targets[0].vars,
         'added_facts' => $add_me_targets[1].facts,
 			  'added_vars' => $add_me_targets[1].vars,
-			  'target_from_string' => get_targets('bar'),
+        'target_not_overwritten' => get_targets('bar')[0].vars['bar_1_var'],
```
adreyer: indentation looks weird here

spec/integration/inventory2_spec.rb
```diff
@@ -282,7 +287,10 @@ def fact_plan(name = 'facts_test')
       expect(result['existing_vars']).to eq(expected_hash_pre)
       expect(result['added_facts']).to eq(expected_hash_post)
       expect(result['added_vars']).to eq(expected_hash_post)
-      expect(result['target_from_string']).to eq(["Target('bar_1', {})"])
+      expect(result['target_not_overwritten']).to eq("dont_overwrite")
+      # TODO, Given bar_1 is defined by just a name, printing the target gets an empty string.
```
adreyer: The alternative would be to give this a uri above.

donoghuc: Cool. That works for me.

PR #959, opened 2019-04-17
(BOLT-1232) Accept --targets as well as --nodes

adreyer:


lib/bolt/cli.rb
```diff
@@ -127,23 +128,24 @@ def parse
       raise e
     end
 
-    def get_targets(options)
-      target_opts =  options.keys.select { |opt| %i[query rerun].include?(opt) }
-      target_opts << :nodes unless options[:nodes].empty?
+    def update_targets(options)
+      target_opts = options.keys.select { |opt| %i[query rerun nodes targets].include?(opt) }
+      target_string = "--nodes', '--targets', '--rerun', or '--query'"
```
donoghuc:
```suggestion
      target_string = "'--nodes', '--targets', '--rerun', or '--query'"
```

PR #961, opened 2019-04-18
(maint) Fix typo in yaml plan docs

lucywyman:
We defined 2 parameters `frontends`, based on the descriptions the second should be `backends`.


lucywyman:
@melamos  for :eyes:

melamos:
Fixed in docs.

PR #962, opened 2019-04-19
(feat) wrap docker run_command in bash

tphoney:


adreyer:
The solution for this is going to be to add a new `shell_command` option to the docker transport. When set all commands on the target will be wrapped in the `shell_command`

lib/bolt/transport/docker.rb
```diff
@@ -8,7 +8,7 @@ module Bolt
   module Transport
     class Docker < Base
       def self.options
-        %w[host service-url service-options tmpdir interpreters]
+        %w[host service-url service-options tmpdir interpreters shell_command tty]
```
adreyer: Note that `service-url` has a hyphen, this probably should too for consistency in the config file.

adreyer:
The new config options should be added to https://github.com/puppetlabs/bolt/blob/master/pre-docs/bolt_configuration_options.md#docker-transport-configuration-options

florindragos:
> Can you update the commit message to include the ticket number.

@adreyer Done

PR #966, opened 2019-04-22
(BOLT-1109) Support puppetdb lookups in inventory2

adreyer:
This adds a primative plugin concept to Bolt. Currenlty this is closely
tailored to the needs of the 'puppetdb' plugin which is a bit magic
since it uses prexisting puppetdb configuration.

The key parts of the plugin framework are `self.setup` which returns a new
plugin library with plugins loaded and `by_name` which allows us to
fetch the puppetdb plugin. We should expect to refactor basically
everything else around how we load plugins as we develop more hooks and
external plugins.

This is currently based on the BOLT-1232 PR since it also needs to add pre-docs

### DIFF
--- a/lib/bolt/inventory/group2.rb
+++ b/lib/bolt/inventory/group2.rb
@@ -13,7 +13,7 @@ module Bolt
 
       DATA_KEYS = %w[name config facts vars features].freeze
       NODE_KEYS = DATA_KEYS + %w[alias uri]
-      GROUP_KEYS = DATA_KEYS + %w[groups targets]
+      GROUP_KEYS = DATA_KEYS + %w[groups targets target-lookups]
       CONFIG_KEYS = Bolt::TRANSPORTS.keys.map(&:to_s) + ['transport']
 
       def initialize(data)
@@ -36,6 +36,8 @@ module Bolt
         @features = fetch_value(data, 'features', Array)
         @config = fetch_value(data, 'config', Hash)
 
+        @target_lookups = fetch_value(data, 'target-lookups', Array)
+
         unless (unexpected_keys = @config.keys - CONFIG_KEYS).empty?
           msg = "Found unexpected key(s) #{unexpected_keys.join(', ')} in config for group #{@name}"
           @logger.warn(msg)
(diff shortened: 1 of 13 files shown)


lib/bolt/inventory/group2.rb
```diff
@@ -115,6 +77,79 @@ def target_data(target_name)
...
+        @target_lookups.each do |lookup|
+          unless lookup.is_a?(Hash)
+            raise ValidationError("target-lookup is not a hash: #{lookup}", @name)
+          end
+          unless lookup['plugin']
+            raise ValidationError("target-lookup does not specify a plugin: #{lookup}", @name)
+          end
+
+          unless (plugin = plugins.by_name(lookup['plugin']))
+            raise ValidateionError("target-lookup specifies an unkown plugin: '#{lookup['plugin']}'", @name)
```
donoghuc:
```suggestion
            raise ValidationError("target-lookup specifies an unknown plugin: '#{lookup['plugin']}'", @name)
```

lib/bolt/inventory/group2.rb
```diff
@@ -115,6 +77,79 @@ def target_data(target_name)
...
+        end
+      end
+
+      def lookup_targets(plugins)
+        @target_lookups.each do |lookup|
+          unless lookup.is_a?(Hash)
+            raise ValidationError("target-lookup is not a hash: #{lookup}", @name)
+          end
+          unless lookup['plugin']
+            raise ValidationError("target-lookup does not specify a plugin: #{lookup}", @name)
```
donoghuc:
```suggestion
            raise ValidationError.new("target-lookup does not specify a plugin: #{lookup}", @name)
```

lib/bolt/inventory/group2.rb
```diff
@@ -115,6 +77,79 @@ def target_data(target_name)
...
+            raise ValidationError.new(alias_conflict(alia, found, target['name']), @name)
+          end
+          @aliases[alia] = target['name']
+        end
+      end
+
+      def lookup_targets(plugins)
+        @target_lookups.each do |lookup|
+          unless lookup.is_a?(Hash)
+            raise ValidationError("target-lookup is not a hash: #{lookup}", @name)
```
donoghuc:
```suggestion
            raise ValidationError.new("target-lookup is not a hash: #{lookup}", @name)
```

pre-docs/inventory_version2.md
```diff
@@ -59,6 +59,37 @@ targets:
...
+lookup the targets in the node. The lookup method is eventually expected to be
+pluggable but for now Bolt only ships a single hardcoded plugin `puppetdb` that
+can use a puppetdb query to populate the targets.
+
+
+```
+groups:
+  - name: windows
+    target-lookups:
+      plugin: puppetdb
```
donoghuc: plugin needs to be array, for example
```
version: 2
groups:
  - name: puppet_db
    target-lookups:
       - plugin: puppetdb
         query: "nodes[certname] {}"
```

donoghuc:
Should we fail hard and not allow *any* tasks to be run when there is a problem loading a plugin? 
For example:
```yaml
version: 2
groups:
  - name: puppet_db
    target-lookups:
       - plugin: puppetdb
         query: inventory[foo]{facts.os.release.major = "7"}
    config: 
      transport: pcp
```
```
cas@cas-ThinkPad-T460p:~/working_dir/bolt$ bolt task show
Failed to query PuppetDB: Can't extract unknown 'inventory' field 'foo'. Acceptable fields are 'certname', 'timestamp', 'environment', 'facts', and 'trusted'
```

donoghuc:
Still digging in to this but I'm seeing some weird behavior nesting inventory:
```yaml
version: 2
groups:
  - name: puppet_db
    target-lookups:
       - plugin: puppetdb
    config: 
      transport: pcp
    groups: 
      - name: foo
        targets: 
          - name: localhost
        config:
          transport: local
        target-lookups:
           - plugin: puppetdb
```
The result reports it ran on all three targets but it appears to have just run three times on localhost?
```
cas@cas-ThinkPad-T460p:~/working_dir/bolt$ bolt command run whoami -n puppet_db
Started on c8jy9ikuuyf0g1r.delivery.puppetlabs.net...
Started on ...
Started on p7b9u0qbudxqvnh.delivery.puppetlabs.net...
Finished on :
  STDOUT:
    cas
Finished on c8jy9ikuuyf0g1r.delivery.puppetlabs.net:
  STDOUT:
    cas
Finished on p7b9u0qbudxqvnh.delivery.puppetlabs.net:
  STDOUT:
    cas
Successful on 3 nodes: c8jy9ikuuyf0g1r.delivery.puppetlabs.net,p7b9u0qbudxqvnh.delivery.puppetlabs.net,localhost
Ran on 3 nodes in 0.00 seconds
```

lib/bolt/inventory/group2.rb
```diff
@@ -115,6 +77,79 @@ def target_data(target_name)
...
+        end
+
+        raise ValidationError.new("Node #{target} does not have a name", @name) unless target['name']
+        @targets[target['name']] = target
+
+        unless (unexpected_keys = target.keys - NODE_KEYS).empty?
+          msg = "Found unexpected key(s) #{unexpected_keys.join(', ')} in target #{target['name']}"
+          @logger.warn(msg)
+        end
+        config_keys = target['config']&.keys || []
```
donoghuc: I'm don't think it was introduced here, but perhaps we should validate that `target['config']` is either a `Hash` or `nil`. I accidentally did `config: local` and got a stack trace. 
```
NoMethodError: undefined method `keys' for "local":String
  /home/cas/working_dir/bolt/lib/bolt/inventory/group2.rb:106:in `add_target'
  /home/cas/working_dir/bolt/lib/bolt/inventory/group2.rb:59:in `block in initialize'
```

donoghuc: This likely needs to be addressed in inventory v1 as well. Do not want to block this work on that. Filed https://tickets.puppetlabs.com/browse/BOLT-1274

PR #967, opened 2019-04-22
(BOLT-1262) Do not let file upload hang with sudoable

donoghuc:
In the case where a user specified with `run-as` tried to upload a file to a destination owned by a user with different privilages an command prompt is raised which causes the command to hang indefinitely. By adding the `-f` flag to the `mv` command an appropriate permission error is raised and the operation does not hang.

### DIFF
--- a/lib/bolt/transport/sudoable.rb
+++ b/lib/bolt/transport/sudoable.rb
@@ -44,7 +44,7 @@ module Bolt
               conn.copy_file(source, tmpfile)
               # pass over file ownership if we're using run-as to be a different user
               dir.chown(conn.run_as)
-              result = conn.execute(['mv', tmpfile, destination], sudoable: true)
+              result = conn.execute(['mv', '-f', tmpfile, destination], sudoable: true)
               if result.exit_code != 0
                 message = "Could not move temporary file '#{tmpfile}' to #{destination}: #{result.stderr.string}"
                 raise Bolt::Node::FileError.new(message, 'MV_ERROR')


donoghuc:
Example prompt that was hanging:
```
sfoec4k5bqh4pah:transport root# sudo -u local_user mv /tmp/19f68c73-36fe-4cee-876c-861681529355/local_file.txt /tmp/local_file.txt
override rw-------  root/wheel for /tmp/local_file.txt? (y/n [n]) 
```
Instead of hanging you now get something like:
```
Failed on localhost:
  Could not move temporary file '/tmp/bc8445e1-b3e0-46aa-975e-df9d311a47e0/local_file.txt' to /tmp/local_file.txt: mv: rename /tmp/bc8445e1-b3e0-46aa-975e-df9d311a47e0/local_file.txt to /tmp/local_file.txt: Permission denied
Failed on 1 node: localhost
```

donoghuc:
Filed https://tickets.puppetlabs.com/browse/BOLT-1262
Will clean up once tests run/reviewer feedback

donoghuc:
Updated https://tickets.puppetlabs.com/browse/BOLT-1262 with another bug I found. I would also like to search for other places we use `mv`. For example this line https://github.com/puppetlabs/bolt/blob/55de205f60b5ca69c1812cec18ef2bec8db47f1b/lib/bolt/transport/local/shell.rb#L70 though, i guess in that case we are moving to a fresh directory.

lib/bolt/transport/local/shell.rb
```diff
@@ -67,8 +67,9 @@ def handle_sudo_errors(err, pid)
 
         def copy_file(source, dest)
           if source.is_a?(StringIO)
-            File.open("tempfile", "w") { |f| f.write(source.read) }
-            execute(['mv', 'tempfile', dest])
+            File.open(dest, 'w') { |f| f.write(source.read) }
```
donoghuc: I think I see why this was written to a tempfile and then copied before. For the case where the less privileged user needs to do something as root. The case I was running in to was the opposite (root running as a different user). Will need to figure out a solution that works for both.

lib/bolt/transport/local/shell.rb
```diff
@@ -67,8 +67,7 @@ def handle_sudo_errors(err, pid)
 
         def copy_file(source, dest)
           if source.is_a?(StringIO)
-            File.open("tempfile", "w") { |f| f.write(source.read) }
-            execute(['mv', 'tempfile', dest])
+            execute(['echo', source.read.to_s, '>', dest.to_s])
```
donoghuc: Getting the escaping and interpolation just right here will require a bit more work than just a `to_s` method.

donoghuc: Actually I'm thinking redirection just wont work due to https://tickets.puppetlabs.com/browse/BOLT-1056

PR #974, opened 2019-04-30
(BOLT-1243) Error with helpful message when no CLI command given

donoghuc:
Previously an missing command argument would result in a stack trace. This commit validates that a command is given when invoking `bolt command run` on the CLI is a string with at least one character. Note this validation is modeled after the plan function `run_command` which validates the command argument is of type `String[1]`.

### DIFF
--- a/spec/bolt/cli_spec.rb
+++ b/spec/bolt/cli_spec.rb
@@ -179,17 +179,17 @@ describe "Bolt::CLI" do
       let(:targets) { [target, Bolt::Target.new('bar')] }
 
       it "accepts a single node" do
-        cli = Bolt::CLI.new(%w[command run --nodes foo])
+        cli = Bolt::CLI.new(%w[command run uptime --nodes foo])
         expect(cli.parse).to include(targets: [target])
       end
 
       it "accepts multiple nodes" do
-        cli = Bolt::CLI.new(%w[command run --nodes foo,bar])
+        cli = Bolt::CLI.new(%w[command run uptime --nodes foo,bar])
         expect(cli.parse).to include(targets: targets)
       end
 
       it "accepts multiple nodes across multiple declarations" do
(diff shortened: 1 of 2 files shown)


spec/bolt/cli_spec.rb
```diff
@@ -2108,7 +2122,7 @@ def stub_config(file_content = {})
       after(:each) { ENV.delete('BOLT_INVENTORY') }
 
       it 'errors when BOLT_INVENTORY is set' do
-        cli = Bolt::CLI.new(%W[command run --inventoryfile #{File.join(inventorydir, 'invalid.yml')} --nodes foo])
+        cli = Bolt::CLI.new(%W[command run id --inventoryfile #{File.join(inventorydir, 'invalid.yml')} --nodes foo])
```
lucywyman: Haha, one of these things is not like the others :)

spec/bolt/cli_spec.rb
```diff
@@ -1989,31 +2003,31 @@ def stub_config(file_content = {})
...
+        cli = Bolt::CLI.new(%W[command run uptime --configfile #{conf.path} --nodes foo])
         cli.parse
         expect(cli.config.transports[:ssh]['host-key-check']).to eq(false)
       end
     end
 
     it 'reads run-as for ssh' do
       with_tempfile_containing('conf', YAML.dump(complete_config)) do |conf|
-        cli = Bolt::CLI.new(%W[command run --configfile #{conf.path} --nodes foo --password bar --no-host-key-check])
+        cli = Bolt::CLI.new(%W[command run r --configfile #{conf.path} --nodes foo --password bar --no-host-key-check])
```
lucywyman: Is this an actual command?

donoghuc: It doesn't need to be an actual command. The reason I used a single character is to avoid line length rubocop

lucywyman: Ah, makes sense :+1:

PR #975, opened 2019-04-30
(BOLT-1223) Rename result type to result action, expose in plan

donoghuc:
This commit renames the type attribute of a result object to action (BOLT-1125). It also exposes it as an action method for a result object in the plan language.

### DIFF
--- a/bolt-modules/boltlib/lib/puppet/datatypes/applyresult.rb
+++ b/bolt-modules/boltlib/lib/puppet/datatypes/applyresult.rb
@@ -10,6 +10,8 @@ Puppet::DataTypes.create_type('ApplyResult') do
       error => Callable[[], Optional[Error]],
       ok => Callable[[], Boolean],
       message => Callable[[], Optional[String]],
+      action => Callable[[], String],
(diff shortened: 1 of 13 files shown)


donoghuc:
I think what you are referring to is an event `type` https://github.com/puppetlabs/bolt/blob/b062fa3faed7fa8600ef17275a46933b37a12bc6/lib/bolt/transport/base.rb#L35-L38 (like `:node_start`, or `:node_result`). 

The change in the first commit for this PR is simply renaming the a result `type` attribute to `action`. Basically in CLI terms `action` should map to `subcommand` (task, command, etc). Too many things called `type`!

lucywyman:
:woman_facepalming: Yep, definitely got that confused. I saw it was iterating over a 'result' array and totally assumed it was making result objects.

nicklewis:
One of the commits references `BOLT-1125` which appears to be the wrong ticket.

donoghuc:
BOLT-1125 is the ticket that describes adding the `type` key to results. https://tickets.puppetlabs.com/browse/BOLT-1125

PR #976, opened 2019-05-01
(ORCH-2302) Update plan executor to be thread safe

McdonaldSeanp:
This commit updates the ruby app.rb code to fix code meant to limit concurrent
calls to the bolt PAL to a single thread.

The code has been updated to work with the newer version of the
concurrent-ruby gem, which has consolidated to a single Promisises class for
everything. Additionally, the call to "pal.get_plan_info" has also been
wrapped in a concurrency promise so that it will also be forced to use the
SingleThreadExecutor.

### DIFF
--- a/lib/plan_executor/app.rb
+++ b/lib/plan_executor/app.rb
@@ -92,9 +92,15 @@ module PlanExecutor
       error = validate_schema(@schema, body)
       return [400, error.to_json] unless error.nil?
       name = body['plan_name']
-      # Errors if plan is not found
-      @pal.get_plan_info(name)
-
+      # We need to wrap all calls to @pal (not just plan_run) in a future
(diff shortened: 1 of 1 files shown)


lib/plan_executor/app.rb
```diff
@@ -92,9 +92,10 @@ def validate_schema(schema, body)
       error = validate_schema(@schema, body)
       return [400, error.to_json] unless error.nil?
       name = body['plan_name']
-      # Errors if plan is not found
-      @pal.get_plan_info(name)
-
+      result = Concurrent::Future.execute(executor: @worker) do
```
caseywilliams: I think it's worth leaving the comment about why this is done -- otherwise looks fine to me

McdonaldSeanp: Updated

PR #973, opened 2019-04-30
(BOLT-1227) Document OpenSSH configuration file options

donoghuc:
Add documentation about where OpenSSH config files are read from and what values are respected. Also document dependency on ssh-agent for authentication with password protected private keys.

PR #978, opened 2019-05-01
(BOLT-1227) Enumerate OpenSSH configuration options available with bolt

donoghuc:
This commit adds a list of configuration options that the net-ssh gem will respect (and wont be overridden by bolt).


donoghuc:
Continuation of https://github.com/puppetlabs/bolt/pull/973#discussion_r279819964

I *think* this is the complete list. Basically i just dug through https://github.com/net-ssh/net-ssh/blob/master/lib/net/ssh/config.rb and found the options that were not overriden or ignored in how we net-ssh with bolt.

PR #980, opened 2019-05-02
(BOLT-1154) Remove unnecessary guard conditions in plan functions

donoghuc:
BOLT-1131 introduced a guard condition that prevents select plan functions from bein executed in an apply block. Specifically if `Puppet[:tasks]` is not truthy then an error is raised. This commit removes redundant conditionals to ensure the plan functions are runnable by the plan.

### DIFF
--- a/bolt-modules/boltlib/lib/puppet/functions/add_facts.rb
+++ b/bolt-modules/boltlib/lib/puppet/functions/add_facts.rb
@@ -23,9 +23,9 @@ Puppet::Functions.create_function(:add_facts) do
         .from_issue_and_stack(Bolt::PAL::Issues::PLAN_OPERATION_NOT_SUPPORTED_WHEN_COMPILING, action: 'add_facts')
     end
 
-    inventory = Puppet.lookup(:bolt_inventory) { nil }
-    executor = Puppet.lookup(:bolt_executor) { nil }
-    executor&.report_function_call('add_facts')
+    inventory = Puppet.lookup(:bolt_inventory)
+    executor = Puppet.lookup(:bolt_executor)
+    executor.report_function_call('add_facts')
 
     inventory.add_facts(target, facts)
   end
--- a/bolt-modules/boltlib/lib/puppet/functions/add_facts.rb
+++ b/bolt-modules/boltlib/lib/puppet/functions/add_facts.rb
@@ -24,13 +24,6 @@ Puppet::Functions.create_function(:add_facts) do
     end
 
     inventory = Puppet.lookup(:bolt_inventory) { nil }
-
-    unless inventory
-      raise Puppet::ParseErrorWithIssue.from_issue_and_stack(
-        Puppet::Pops::Issues::TASK_MISSING_BOLT, action: _('add facts')
-      )
-    end
-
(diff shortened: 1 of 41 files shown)


donoghuc:
In some cases we would check for `Puppet.features.bolt?`. Is that something we should keep? It seems that it used to be common to gate on existence of `Puppet.lookup(:bolt_inventory)` is that a check that pre-dated `Pupet.features.bolt? that can be replaced? Is there a case where `Puppet[:tasks]` would be true and `Puppet.lookup(:bolt_inventory or :bolt_executor)` would be false?

nicklewis:
I don't think we need to be checking for `Puppet.features.bolt?`. That just checks that `require 'bolt'` works, which it must.

These checks would be more relevant if we were shipping boltlib outside Bolt, but we're not. So it _must_ be the case that they're being called from Bolt. In that case, the only meaningful check is for `Puppet[:tasks]` to determine whether or not we're running as _plan_ code or as _apply_ code.

We can also probably remove the default blocks for the calls to `Puppet.lookup`, since we explicitly expect that they can never fail.

lib/bolt/catalog.rb
```diff
@@ -74,7 +73,8 @@ def compile_catalog(request)
                                        facts: target["facts"] || {},
                                        variables: target["variables"] || {}) do |pal|
           Puppet.override(bolt_pdb_client: pdb_client,
-                          bolt_inventory: setup_inventory(request['inventory'])) do
+                          bolt_inventory: setup_inventory(request['inventory']),
+                          bolt_executor: nil) do
```
donoghuc: @nicklewis For plan functions that are allowed in apply blocks, the default `nil` block handled the case where we look up the executor for analytics. Should we handle that by overriding in catalog compilation? Or should we keep the default `nil` blocks in the various plan functions?

nicklewis: If there's a case where it's _legal_ for it to be `nil`, then we should have a default block and appropriately handle it being `nil`. But for cases where that _isn't_ supposed to be possible, we should just let it fail rather than having a default and then failing explicitly.

bolt-modules/boltlib/lib/puppet/functions/add_facts.rb
```diff
@@ -23,15 +23,8 @@ def add_facts(target, facts)
...
-    unless inventory
-      raise Puppet::ParseErrorWithIssue.from_issue_and_stack(
-        Puppet::Pops::Issues::TASK_MISSING_BOLT, action: _('add facts')
-      )
-    end
-
-    executor = Puppet.lookup(:bolt_executor) { nil }
+    inventory = Puppet.lookup(:bolt_inventory)
+    executor = Puppet.lookup(:bolt_executor)
     executor&.report_function_call('add_facts')
```
nicklewis: We shouldn't use the `&.` syntax here, since we expect that `executor` will never be `nil`. For the functions that _can_ be called from apply / without an executor, `&.` is still appropriate.

donoghuc: Ahh, yeah good call. Updated.

PR #981, opened 2019-05-06
(maint) Exclude files by default in gitignore, dockerignore

donoghuc:
Instead of maintaining a list of what to exclude in config file for git and docker, exclude everything by default and maintain a list of what to include. Also support having an untracked `Boltdir` in project root directory by adding it to `.gitignore` and `.rubocop` configs.


donoghuc:
I can easily maintain https://github.com/puppetlabs/bolt/pull/981/commits/588f7681ec90ca5c4523e184cb44a6e768b3f5a3  locally, seeing if there is any interest from others. I think that the bolt-server one could be worth doing either way.

lucywyman:
I think this is a good idea, though I haven't been annoyed enough by the previous workflow to care too much one way or the other :woman_shrugging:

PR #983, opened 2019-05-08
(BOLT-1299) Properly detect multi-line evals

nicklewis:
Previously, we were looking for a newline character to decide that the
content of an eval was multiple lines and should be wrapped in `with()`.
However, single line strings created using the mutli-line syntax in YAML
(which is necessary to be interpreted as code blocks) end in a newline.
We now check for more precisely what we means, which is that the string
contains multiple lines, not just that it might end with a newline.

### DIFF
--- a/lib/bolt/pal/yaml_plan/step.rb
+++ b/lib/bolt/pal/yaml_plan/step.rb
@@ -77,7 +77,7 @@ module Bolt
             # We have to do a little extra parsing here, since we only need
             # with() for eval blocks
             code = Bolt::Util.to_code(body['eval'])
-            if @name && code.include?("\n")
+            if @name && code.lines.count > 1
               # A little indented niceness
               indented = code.gsub(/\n/, "\n    ").chomp("  ")
               result << "with() || {\n    #{indented}}"


MikaelSmith:
This sounds like it's related to a bug. Shouldn't it have a ticket?

nicklewis:
Hmm yes it should, for the sake of a release note.

donoghuc:
Is this the expected behavior for folded string as well?  (`>`)

For example (yaml plan snippet):
```
  - eval: >
      $x = $foo + 1
      $x * 2
    name: eval_step
    description: foo
  - eval: |
      $foo
    name: eval
```
Previously:
```
  $eval_step = with() || {
    $x = $foo + 1 $x * 2
  }
  $eval = with() || {
    $foo
  }
```
With this change:
```
  $eval_step = $x = $foo + 1 $x * 2

  $eval = $foo
```

nicklewis:
Hmm, that's annoying because I believe there's no way for us to tell if a folded string was _written_ as multiple lines. We could choose to always wrap folded strings since they _could_ have been written as 
 multiple lines, or we could choose (as this PR implicitly does) to never wrap folded strings since they're always _technically_ one line. Note the former requires building a new subclass of 
`EvaluableString` though.

donoghuc:
I think it is reasonable to expect users to choose the correct block scalar. I think this is an excellent SO for helping decide. https://stackoverflow.com/a/21699210 .

PR #984, opened 2019-05-08
(BOLT-1268) Only load inventory when necessary

nicklewis:
Previously, we loaded the inventory for all commands, regardless of
whether the command actually needed the inventory. This meant that an
invalid inventory would cause commands like `bolt task show` to fail,
even though they shouldn't care about the inventory whatsoever.

We were loading the inventory in order to figure out the number of nodes
and groups in use to include in the analytics report. We now only
include that information if the command is using targets.

### DIFF
--- a/lib/bolt/cli.rb
+++ b/lib/bolt/cli.rb
@@ -261,11 +261,19 @@ module Bolt
         screen += '_object'
       end
 
-      @analytics.screen_view(screen,
-                             output_format: config.format,
-                             target_nodes: options.fetch(:targets, []).count,
-                             inventory_nodes: inventory.node_names.count,
-                             inventory_groups: inventory.group_names.count)
+      screen_view_fields = {
+        output_format: config.format
+      }
+
+      # Only include target and inventory info for commands that take a targets
+      # list. This avoids loading inventory for commands that don't need it.
+      if options.key?(:targets)
+        screen_view_fields.merge!(target_nodes: options[:targets].count,
+                                  inventory_nodes: inventory.node_names.count,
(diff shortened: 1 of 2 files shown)


donoghuc:
The root cause of the travis failures appears to be that the `puppet_agent::install` task is getting the latest puppet 5 version. Seems like the change is that in the install shell task we default `collection` to `puppet` resulting in downloading http://apt.puppet.com/puppet-release-xenial.deb which points to puppet 5.5.14. The test runs https://github.com/puppetlabs/bolt/blob/master/spec/fixtures/apply/basic/plans/defer.pp a plan to test `Deferred` which is a puppet 6 feature.

lib/bolt/cli.rb
```diff
@@ -261,11 +261,18 @@ def execute(options)
...
 
-      @analytics.screen_view(screen,
-                             output_format: config.format,
-                             target_nodes: options.fetch(:targets, []).count,
-                             inventory_nodes: inventory.node_names.count,
-                             inventory_groups: inventory.group_names.count)
+      screen_view_fields = {
+        output_format: config.format,
+      }
+
```
donoghuc: The unrelated failure in travis has been resolved (see https://github.com/puppetlabs/bolt/pull/985 conversation for details). There are a couple rubocop violations. I dont think we would likely regress on this but perhaps a test would make sense?

nicklewis: Hmm I can look into how to test this.

nicklewis:
Updated with a couple of tests

nicklewis:
Rubocop _should_ be happy with this update

PR #989, opened 2019-05-15
(maint) Update puppet_agent module to 2.1.2

donoghuc:
The location of public release packages will change on May 14.
As such, the following changes aim to prepare for the changeover
to ensure there are no breakages with downstream dependencies.


donoghuc:
Seems like http://apt.puppet.com/puppet-release-xenial.deb is pointing to puppet5 collection. I think RE can fix that. Will re-kick once confirmed.

donoghuc:
I guess this should get a ticket. I'm guessing people using apply prep will run in to this on affected platforms.

donoghuc:
https://tickets.puppetlabs.com/browse/RE-12397

PR #990, opened 2019-05-15
(maint) Ensure testing deferred in apply has compatible agent version

donoghuc:
For the second time in two weeks a puppet 5 agent was installed for "latest". The test failure message for using an apply with the `deferred` feature does not make it obvious that an incompatible agent is used. This commit updates the test to use the generalized `BoltSpec::PuppetAgent` install method and ensure a `deferred` compatible agent is installed. It also modifies the install method such that it will default to install the latest collection and optionally install from a specified collection.

### DIFF
--- a/spec/integration/apply_spec.rb
+++ b/spec/integration/apply_spec.rb
@@ -160,11 +160,12 @@ describe "apply" do
 
     context "with a puppet_agent installed" do
       before(:all) do
-        # TODO: Extract into test helper if needed in more files
-        uri = conn_uri('ssh')
-        inventory_data = conn_inventory
-        config_data = root_config
-        run_task('puppet_agent::install', uri, {}, config: config_data, inventory: inventory_data)
+        # Deferred must use puppet >= 6
+        target = conn_uri('ssh')
+        install(target, inventory: conn_inventory)
+        result = run_task('puppet_agent::version', target, {}, config: root_config, inventory: conn_inventory)
+        major_version = result.first['result']['version'].split('.').first.to_i
+        expect(major_version).to be >= 6
       end
 
       context "apply() function" do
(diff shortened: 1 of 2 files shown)


donoghuc:
https://tickets.puppetlabs.com/browse/RE-12397

spec/lib/bolt_spec/puppet_agent.rb
```diff
@@ -35,10 +35,18 @@ def install(target, collection: 'puppet6', inventory: nil)
         }
       }
       inventory ||= {}
-      result = run_task('puppet_agent::install', target,
-                        { 'collection' => collection },
-                        config: config,
-                        inventory: inventory)
+      # Task will get latest collection without collection specified
+      result = if collection
```
nicklewis: The only conditional part of this is the params, right? So this can be simplified a bit

```
task_params = if collection
                { 'collection' => collection }
              else
                {}
              end

run_task('puppet_agent::install', target, task_params, config: config, inventory: inventory)
```

spec/integration/apply_spec.rb
```diff
@@ -160,11 +160,12 @@ def agent_version_inventory
...
-        # TODO: Extract into test helper if needed in more files
-        uri = conn_uri('ssh')
-        inventory_data = conn_inventory
-        config_data = root_config
-        run_task('puppet_agent::install', uri, {}, config: config_data, inventory: inventory_data)
+        # Deferred must use puppet >= 6
+        target = conn_uri('ssh')
+        install(target, inventory: conn_inventory)
+        result = run_task('puppet_agent::version', target, {}, config: root_config, inventory: conn_inventory)
+        major_version = result.first['result']['version'].split[0].to_i
```
nicklewis: What is this splitting _on_?

PR #991, opened 2019-05-15
(maint) Git based acceptance tests on fedora28

donoghuc:
Install appropriate dependencies for installing bolt as a gem on fedora28

### DIFF
--- a/acceptance/setup/common/pre-suite/010_install_ruby.rb
+++ b/acceptance/setup/common/pre-suite/010_install_ruby.rb
@@ -25,11 +25,23 @@ PS
       install_package(bolt, 'ruby')
       install_package(bolt, 'ruby-ffi')
       result = on(bolt, 'ruby --version')
-    when /el-|centos|fedora/
+    when /el-|centos/
       # install system ruby packages
       install_package(bolt, 'ruby')
       install_package(bolt, 'rubygem-json')
       install_package(bolt, 'rubygem-ffi')
+      install_package(bolt, 'rubygem-bigdecimal')
+      install_package(bolt, 'rubygem-io-console')
+      result = on(bolt, 'ruby --version')
+    when /fedora/
+      # install system ruby packages
+      install_package(bolt, 'ruby')
+      install_package(bolt, 'ruby-devel')
+      install_package(bolt, 'libffi')
+      install_package(bolt, 'libffi-devel')
+      install_package(bolt, 'redhat-rpm-config')
+      on(bolt, "dnf group install -y 'C Development Tools and Libraries'")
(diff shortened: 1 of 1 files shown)


acceptance/setup/common/pre-suite/010_install_ruby.rb
```diff
@@ -25,14 +25,29 @@
...
+    when /fedora/
+      # install system ruby packages
+      install_package(bolt, 'ruby')
+      install_package(bolt, 'ruby-devel')
+      install_package(bolt, 'make')
+      install_package(bolt, 'make-devel')
+      install_package(bolt, 'libffi')
+      install_package(bolt, 'libffi-devel')
+      install_package(bolt, 'redhat-rpm-config')
+      on(bolt, "dnf group install -y 'C Development Tools and Libraries'")
```
donoghuc: I tried sooo many combinations. This works locally with 
```
 2014  bundle exec beaker-hostgenerator fedora28-64bolt%2Cssh.-ubuntu1604-64ssh.-windows10ent-64winrm. > hosts.yaml
 2015  bundle exec beaker init -h hosts.yaml -o config/git/options.rb
 2016  bundle exec beaker provision
 2017  BEAKER_debug=true bundle exec beaker exec -t ./tests
```
I would not be suprised if some of those are redundant...

nicklewis: This looks like a fairly reasonable set of dependencies. I would expect `make-devel` probably isn't necessary and that `make` would be covered by `C Development Tools and Libraries`, but who knows.

donoghuc: Yeah, Sorry it was frustrating how much effort it took to finally get the right combination of dependencies. I was so excited I got it working finally I didnt want to go back and prune out unneeded deps.  I removed make, make-devel, and rubygem-ffi and its still working for me locally.

acceptance/setup/common/pre-suite/010_install_ruby.rb
```diff
@@ -25,14 +25,26 @@
...
       # install system ruby packages
       install_package(bolt, 'ruby')
       install_package(bolt, 'rubygem-json')
       install_package(bolt, 'rubygem-ffi')
       install_package(bolt, 'rubygem-bigdecimal')
       install_package(bolt, 'rubygem-io-console')
       result = on(bolt, 'ruby --version')
+    when /fedora/
+      # install system ruby packages
+      install_package(bolt, 'ruby')
```
lucywyman: ugh that's so annoying that it doesn't take an array

PR #992, opened 2019-05-16
(BOLT-1264) Set config from PuppetDB facts

lucywyman:
This allows user who are using the PuppetDB inventory plugin to set config
values based on facts gathered from PuppetDB. It works by collecting a list
of facts and a list of certnames that we need those facts for, then makes a
single puppetdb query to the `fact-contents` endpoint for the facts. It then
iterates back through the config to fill in the right values for the right
nodes.

### DIFF
--- a/lib/bolt/plugin/puppetdb.rb
+++ b/lib/bolt/plugin/puppetdb.rb
@@ -3,8 +3,19 @@
 module Bolt
   class Plugin
     class Puppetdb
+      class FactLookupError < Bolt::Error
+        def initialize(fact, err = nil)
+          m = String.new("Fact lookup '#{fact}' contains an invalid factname")
+          m << ": #{err}" unless err.nil?
+          super(m, 'bolt.plugin/fact-lookup-error')
+        end
+      end
+
+      TARGET_OPTS = %w[uri name config].freeze
+
       def initialize(pdb_client)
         @puppetdb_client = pdb_client
+        @logger = Logging.logger[self]
       end
 
       def name
@@ -15,9 +26,61 @@ module Bolt
         ['lookup_targets']
       end
 
+      def warn_missing_fact(certname, fact)
+        @logger.warn("Could not find fact #{fact} for node #{certname}")
+      end
+
+      def fact_path(raw_fact)
+        fact_path = raw_fact.split(".")
+        if fact_path[0] == 'facts'
+          fact_path.drop(1)
+        elsif fact_path == ['certname']
+          fact_path
+        else
+          raise FactLookupError.new(raw_fact, "fact lookups must start with 'facts.'")
+        end
+      end
+
       def lookup_targets(opts)
-        nodes = @puppetdb_client.query_certnames(opts['query'])
-        nodes.map { |certname| { 'uri' => certname } }
+        targets = @puppetdb_client.query_certnames(opts['query'])
+        facts = []
+
+        target_opts = opts.select { |k, _| TARGET_OPTS.include?(k) }
+        Bolt::Util.walk_vals(target_opts) do |value|
+          # This is done in parts instead of in place so that we only need to
+          # make one puppetDB query
+          if value.is_a?(String)
+            facts << fact_path(value)
+          end
+          value
+        end
+
+        facts.uniq!
+        # Returns {'mycertname' => [{'path' => ['nested', 'fact'], 'value' => val'}], ... }
+        fact_values = @puppetdb_client.fact_values(targets, facts)
+
+        targets.map do |certname|
+          target_data = fact_values[certname]
+          target = resolve_facts(target_opts, certname, target_data) || {}
+          target['uri'] = certname unless target['uri'] || target['name']
+
+          target
+        end
+      end
+
+      def resolve_facts(config, certname, target_data)
+        Bolt::Util.walk_vals(config) do |value|
+          if value.is_a?(String)
+            data = target_data&.detect { |d| d['path'] == fact_path(value) }
+            warn_missing_fact(certname, value) if data.nil?
+            # If there's no fact data this will be nil
+            data&.fetch('value', nil)
+          elsif value.is_a?(Array) || value.is_a?(Hash)
+            value
+          else
+            raise FactLookupError.new(value, "fact lookups must be a string")
+          end
+        end
       end
     end
   end
--- a/lib/bolt/puppetdb/client.rb
+++ b/lib/bolt/puppetdb/client.rb
@@ -42,6 +42,22 @@ module Bolt
         end
       end
 
+      def fact_values(certnames = [], facts = [])
+        return {} if certnames.empty? || facts.empty?
+
+        certnames.uniq!
+        name_query = certnames.map { |c| ["=", "certname", c] }
+        name_query.insert(0, "or")
+
+        facts_query = facts.map { |f| ["=", "path", f] }
+        facts_query.insert(0, "or")
+
+        query = ['and', name_query, facts_query]
+        result = make_query(query, 'fact-contents')
+        result.map! { |h| h.delete_if { |k, _v| %w[environment name].include?(k) } }
+        result.group_by { |c| c['certname'] }
+      end
+
       def make_query(query, path = nil)
         body = JSON.generate(query: query)
         url = "#{uri}/pdb/query/v4"
--- a/spec/bolt/puppetdb/client_spec.rb
+++ b/spec/bolt/puppetdb/client_spec.rb
@@ -78,11 +78,34 @@ describe Bolt::PuppetDB::Client do
       },
         'node2' => {
           'foo' => 'bar',
-          'name' => 'node2'
+          '1' => 'the loneliest number',
+          'name' => {
+            'node' => {
+              'kit' => 'kat'
+            }
+          }
         } }
     end
 
     let(:client) { pdb_client }
+
+    # Hash formatting is hard, so do it in the examples
+    let(:expected_node1_foo) do
+      [{ "certname" => "node1", "path" => ["foo"], "value" => "bar" }]
+    end
+
+    let(:expected_node2_foo) do
+      [{ "certname" => "node2", "path" => ["foo"], "value" => "bar" }]
+    end
+
+    let(:expected_node2_all) do
+      [{ "certname" => "node2",
+         "path" => ["foo"],
+         "value" => "bar" },
+       { "certname" => "node2",
+         "path" => %w[name node kit],
+         "value" => "kat" }]
+    end
 
     before(:all) do
       push_facts(facts_hash)
@@ -97,8 +120,25 @@ describe Bolt::PuppetDB::Client do
       expect(facts).to eq(facts_hash)
     end
 
+    it 'should get fact values' do
+      values = client.fact_values(%w[node1], [['foo']])
+      expect(values).to eq('node1' => expected_node1_foo)
+    end
+
+    it 'should get fact values for multiple nodes' do
+      values = client.fact_values(%w[node1 node2], [['foo']])
+      expect(values).to eq('node1' => expected_node1_foo,
+                           'node2' => expected_node2_foo)
+    end
+
+    it 'should get fact values for multiple facts' do
+      values = client.fact_values(%w[node1 node2], [['foo'], %w[name node kit]])
+      expect(values).to eq('node1' => expected_node1_foo,
+                           'node2' => expected_node2_all)
+    end
+
     it 'should get certnames' do
-      certnames = client.query_certnames("inventory { facts.name = 'node2' }")
+      certnames = client.query_certnames("inventory { facts.name.node.kit = 'kat' }")
       expect(certnames).to eq(['node2'])
     end
 
@@ -120,6 +160,14 @@ describe Bolt::PuppetDB::Client do
       client = Bolt::PuppetDB::Client.new(Bolt::PuppetDB::Config.new(conf))
       facts = client.facts_for_node(%w[node1 node2])
       expect(facts).to eq(facts_hash)
+    end
+  end
+
+  describe "#fact_values" do
+    it 'should not make a request to pdb if there are no nodes' do
+      expect(client).to receive(:http_client).never
+      facts = client.fact_values([])
+      expect(facts).to eq({})
     end
   end
 end
--- a/spec/integration/inventory2_spec.rb
+++ b/spec/integration/inventory2_spec.rb
@@ -453,18 +453,20 @@ describe 'running with an inventory file', reset_puppet_settings: true do
 
   # TODO: these tests require docker so they may as well require ssh for now
   context 'with pdb lookups', ssh: true, puppetdb: true do
+    let(:shell_cmd) { 'whoami' }
+    let(:ssh_config) { {} }
+    let(:addtl_inv) { {} }
+    let(:facts) { {} }
+
     before(:each) do
       allow_any_instance_of(Bolt::CLI).to receive(:puppetdb_client).and_return(pdb_client)
 
       push_facts(facts)
     end
 
-    after(:each) { clear_facts(facts) }
-
-    let(:facts) {
-      { conn[:host] => { 'fact1' => true },
-        'noanode.example.com' => { 'fact1' => false } }
-    }
+    after(:each) do
+      clear_facts(facts)
+    end
 
     let(:inventory) do
       {
@@ -472,8 +474,11 @@ describe 'running with an inventory file', reset_puppet_settings: true do
         "target-lookups" => [
           {
             plugin: 'puppetdb',
-            query: 'inventory { facts.fact1 = true }'
-          }
+            query: 'inventory { facts.fact1 = true }',
+            config: {
+              ssh: ssh_config
+            }
+          }.merge(addtl_inv)
         ],
         config: {
           transport: conn[:protocol],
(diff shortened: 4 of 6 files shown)


lib/bolt/plugin/puppetdb.rb
```diff
@@ -16,8 +16,35 @@ def hooks
...
+          target = { 'uri' => certname }
+          # Produces [{certname => 'foo', 'value' => 'bar'}, {...}]
+          target_data = fact_values.select { |data| data['certname'] == certname }
+          target['config'] = resolve_config(opts['config'], target_data)
+          target
+        end.compact
+      end
+
+      def resolve_config(config, target_data)
+        Bolt::Util.walk_vals(config) do |value|
```
lucywyman: Is this too expensive to do twice? I'm not sure if there's a clever way to avoid it.

donoghuc:
Have not dug in, but with a query like:
```yaml
version: 2
groups:
  - name: puppet_db
    target-lookups:
       - plugin: puppetdb
    config: 
      transport: pcp
```
I get 
```
cas@cas-ThinkPad-T460p:~/working_dir/bolt/Boltdir$ bolt command run whoami -n puppet_db
Failed to query PuppetDB: Malformed JSON for query: ["and", ["or", ["=", "certname", "b30uljjp3ivqejm.delivery.puppetlabs.net"], ["=", "certname", "rvlut1mdk1rr11n.delivery.puppetlabs.net"]], ]
```

lucywyman:
@donoghuc I need to add docs, but I think you just put the fact name not the whole query.

lib/bolt/plugin/puppetdb.rb
```diff
@@ -16,8 +16,40 @@ def hooks
...
+              facts << value.split(".")
+            end
+            value
+          end
+          fact_values = @puppetdb_client.fact_values(targets, facts)
+        end
+
+        targets.map do |certname|
+          target = { 'uri' => certname }
+          if opts['config']
```
donoghuc: The ticket description is kind of confusing. The title `Allow puppetdb plugin to set host and name independently` and first paragraph of description
```
The puppetdb plugin will only accept a query that returns a certname and then set that certname to be the targets uri. For some users the node is not accessible through the certname so they need to lookup an IP address and set the name and certname of the target separately.
```
Make is seem like you should be able to set `name` or `uri` in addition to `config` like in the terraform plugin. This condition looks like you can only set config values from facts.

lib/bolt/plugin/puppetdb.rb
```diff
@@ -15,9 +16,77 @@ def hooks
         ['lookup_targets']
       end
 
+      def warn_missing_fact(certname, fact)
+        @logger.warn("Could not find fact #{fact} for node #{certname}")
+      end
+
+      def fact_name(raw_fact)
+        fact = raw_fact.split(".")
+        fact.drop(1) if fact[0] == 'facts'
```
adreyer: We should verify that this starts with "facts"(or certname) and error if it isn't.

lib/bolt/plugin/puppetdb.rb
```diff
@@ -15,9 +16,77 @@ def hooks
...
+      end
+
       def lookup_targets(opts)
-        nodes = @puppetdb_client.query_certnames(opts['query'])
-        nodes.map { |certname| { 'uri' => certname } }
+        targets = @puppetdb_client.query_certnames(opts['query'])
+        facts = []
+        facts << fact_name(opts['name']) if opts.key?('name')
```
adreyer: We had discussed not making config special but allowing all of the "target data" keys here. I think a generic solution will be easier than treating config special.

```ruby
target_opts = opts.select {|k, _| ['uri', 'name', 'config', (maybe other options)].includes?(k) }
Bolt::Util.walk_vals(target_opts) do |value|
  ...
end
```

lib/bolt/plugin/puppetdb.rb
```diff
@@ -15,9 +16,77 @@ def hooks
         ['lookup_targets']
       end
 
+      def warn_missing_fact(certname, fact)
+        @logger.warn("Could not find fact #{fact} for node #{certname}")
+      end
+
+      def fact_name(raw_fact)
```
adreyer: This returns a `facts_path` not a name right?

lib/bolt/plugin/puppetdb.rb
```diff
@@ -15,9 +16,77 @@ def hooks
...
+            end
+            value
+          end
+        end
+
+        facts.uniq!
+        # Returns {'mycertname' => [{'path' => ['nested', 'fact'], 'value' => val'}], ... }
+        fact_values = facts.empty? ? {} : @puppetdb_client.fact_values(targets, facts)
+
+        targets.map do |certname|
```
adreyer: Can we verify that we have the correct certnames and then just loop over the the map described above?

lib/bolt/plugin/puppetdb.rb
```diff
@@ -15,9 +16,77 @@ def hooks
...
+          target_data = fact_values[certname]
+
+          target = {}
```
adreyer: If we build up the target generically first we can then have a simpler check at the end.

target['uri] = certname unless target['uri'] || target['name']

lib/bolt/plugin/puppetdb.rb
```diff
@@ -15,9 +16,77 @@ def hooks
...
+          @logger.warn("No facts available for #{certname}")
+          return nil
+        end
+
+        Bolt::Util.walk_vals(config) do |value|
+          if value.is_a?(String)
+            data = target_data.detect { |d| d['path'] == fact_name(value) }
+
+            if data.nil?
+              warn_missing_fact(certname, value)
```
adreyer: This should probably include the fact path.

spec/bolt/puppetdb/client_spec.rb
```diff
@@ -84,6 +84,29 @@ def facts_hash
...
+         "path" => ["foo"],
+         "value" => "bar" },
+       { "certname" => "node1",
+         "path" => ["name"],
+         "value" => "node1" }]
+    end
+
+    let(:node2_foo) { [{ "certname" => "node2", "path" => ["foo"], "value" => "bar" }] }
+
+    let(:node2_all) do
```
adreyer: We should probably have a test with nested facts.

spec/integration/inventory2_spec.rb
```diff
@@ -453,27 +453,34 @@ def fact_plan(name = 'facts_test')
 
   # TODO: these tests require docker so they may as well require ssh for now
   context 'with pdb lookups', ssh: true, puppetdb: true do
+    let(:shell_cmd) { 'whoami' }
+    # This is an actual user - password is also 'test'
+    let(:user) { 'test' }
```
adreyer: Where is this user created/managed?

lib/bolt/puppetdb/client.rb
```diff
@@ -42,6 +42,22 @@ def facts_for_node(certnames)
...
+        return [] if certnames.empty? || certnames.nil?
+
+        certnames.uniq!
+        name_query = certnames.map { |c| ["=", "certname", c] }
+        name_query.insert(0, "or")
+
+        facts_query = facts.map { |f| ["=", "path", f] } unless facts.empty?
+        facts_query.insert(0, "or") unless facts.empty?
+
+        query = "[\"and\", #{name_query}, #{facts_query}]"
```
adreyer: We should continue to operate on data structures and then `.to_json` the data instead of building a string.

lib/bolt/plugin/puppetdb.rb
```diff
@@ -15,9 +16,77 @@ def hooks
...
+              next
+            end
+            data['value']
+          else
+            value
```
adreyer: We should probably treat integers and unexpected strings the same.

ie if we error on `"config": {"transport": "foo"}` we should error on `"config": {"transport": 1}`

lib/bolt/plugin/puppetdb.rb
lucywyman: Ah so not even support if there is not 'facts'?

lib/bolt/plugin/puppetdb.rb
lucywyman: Should we also apply this pattern to the terraform plugin? This PR is very much based on that - I can put up a separate pr as part of this ticket for that?

lib/bolt/plugin/puppetdb.rb
lucywyman: The value here will be what the user input, so `facts.myfact` as opposed to the fact path which is `['myfact']`. I think the former will make more sense to the user - I'm not sure how to rename the var to make that more clear?

spec/integration/inventory2_spec.rb
lucywyman: https://github.com/puppetlabs/bolt/blob/master/spec/Dockerfile#L20

lib/bolt/plugin/puppetdb.rb
lucywyman: If a certname doesn't have any of the facts in the query it won't be returned, so I think iterating over just the certnames is safer. I refactored a bit to solve what I think the issue was?

lib/bolt/plugin/puppetdb.rb
lucywyman: I don't think we do any inventory content validation here - I'm not sure how we'd distinguish an actual integer (like for concurrency) from something that wasn't supposed to be? I think that happens at the transport level?

lib/bolt/plugin/puppetdb.rb
adreyer: I don't think we have anything analagous to $facts with the terraform plugin.

spec/integration/inventory2_spec.rb
adreyer: we should probably add it to the helper libraries then

lib/bolt/plugin/puppetdb.rb
adreyer: We do for strings. The only way to hard code a string value is to do it outside of the `lookup-targets` otherwise it will error after `facts.` change is made.

string: error if the string references data we don't understand
hash: descend into the hash
array: descend into the array.
number: use the number... This seems wrong we should either use unhandled strings(which will make adding support for new data harder) or error on numbers. The latter seems best.

lib/bolt/plugin/puppetdb.rb
```diff
@@ -15,9 +24,66 @@ def hooks
         ['lookup_targets']
       end
 
+      def warn_missing_fact(certname, fact)
+        @logger.warn("Could not find fact #{fact} for node #{certname}")
+      end
+
+      def fact_path(raw_fact)
+        fact = raw_fact.split(".")
+        raise FactLookupError, raw_fact unless fact[0] == 'facts'
```
adreyer: We should handle `certname` specially here since the `inventory` data does.

lib/bolt/plugin/puppetdb.rb
```diff
@@ -15,9 +24,66 @@ def hooks
...
+
+        facts.uniq!
+        # Returns {'mycertname' => [{'path' => ['nested', 'fact'], 'value' => val'}], ... }
+        fact_values = facts.empty? ? {} : @puppetdb_client.fact_values(targets, facts)
+
+        targets.map do |certname|
+          target_data = fact_values[certname]
+
+          target = {}
+          target_opts.each do |k, v|
```
adreyer: I may be missing but I think we just want to pass `target_opts` directly to `resolve_facts` and let it `walk_vals` rather than looping over them.

lib/bolt/plugin/puppetdb.rb
```diff
@@ -15,9 +24,66 @@ def hooks
...
+            target[k] = resolve_facts(v, certname, target_data)
+          end
+          target['uri'] = certname unless target['uri'] || target['name']
+
+          target
+        end.compact
```
adreyer: What is the `compact` handling?

spec/bolt/puppetdb/client_spec.rb
```diff
@@ -78,12 +78,30 @@ def facts_hash
...
+              '2' => 'kitkat'
+            }
+          }
         } }
     end
 
     let(:client) { pdb_client }
 
+    # Hash formatting is hard, so do it in the examples
+    let(:node1_foo) { [{ "certname" => "node1", "path" => ["foo"], "value" => "bar" }] }
```
adreyer: can we add `expected` to this name to make it clearer what it's for.

lucywyman:
@adreyer I think I addressed all of your comments!

donoghuc:
I am getting a warning about `no facts available...` when I'm not trying to set any config from facts. 
```yaml
version: 2
groups:
  - name: test
    target-lookups:
      - plugin: puppetdb
    config:
      transport: pcp
```
```
cas@cas-ThinkPad-T460p:~/working_dir/bolt/Boltdir$ bolt command run whoami -n test
No facts available for d10z2p30nhcsxnw.delivery.puppetlabs.net
No facts available for byc7mlez1cf0jvf.delivery.puppetlabs.net
Started on d10z2p30nhcsxnw.delivery.puppetlabs.net...
Started on byc7mlez1cf0jvf.delivery.puppetlabs.net...
Finished on d10z2p30nhcsxnw.delivery.puppetlabs.net:
  STDOUT:
    root
Finished on byc7mlez1cf0jvf.delivery.puppetlabs.net:
  STDOUT:
    root
Successful on 2 nodes: d10z2p30nhcsxnw.delivery.puppetlabs.net,byc7mlez1cf0jvf.delivery.puppetlabs.net
Ran on 2 nodes in 1.98 seconds
```

lucywyman:
@donoghuc Fixed

spec/integration/inventory2_spec.rb
```diff
@@ -483,17 +488,119 @@ def fact_plan(name = 'facts_test')
...
+
+        it 'falls back to default' do
+          result = run_cli_json(run_command)
+          expect(result).not_to include('kind')
+          expect(result['items'][0]['result']['stdout']).to eq("#{conn[:user]}\n")
+        end
+      end
+
+      context 'using certname fact' do
+        # TODO
```
donoghuc: TODO in a later PR?

spec/integration/inventory2_spec.rb
```diff
@@ -483,17 +488,119 @@ def fact_plan(name = 'facts_test')
...
+      end
+
+      context 'on a non-queried node' do
+        # Are curly braces : rspec :: parens : lisps?
```
donoghuc: I think you need the "double" parens here, the outer set is the block, the inner set is a hash.

donoghuc:
Played around with it against PE setup. Seems to work well. 
Example inventory
```yaml
version: 2
groups:
  - name: test
    targets:
      - uri: vd6fu4hoop2idtm.delivery.puppetlabs.net
    target-lookups:
      - plugin: puppetdb
        name: facts.ipaddress
        uri: facts.ipaddress
        config: 
          ssh: 
            user: facts.identity.user
            host: facts.networking.hostname
            # this is a boolean fact
            host-key-check: facts.selinux
    config:
      transport: ssh
      ssh:
        private-key: ~/.ssh/id_rsa-acceptance
    vars:
      custom: hi 
```

spec/integration/inventory2_spec.rb
lucywyman: Wait is this something I actually need to change? The comment was just meant to be cheeky.

spec/integration/inventory2_spec.rb
lucywyman: Oh no, I test this elsewhere - I'll remove

spec/integration/inventory2_spec.rb
donoghuc: I couldn't tell if you were actually asking about the curly braces. It looks right if you want that memoized helper to return a hash.

lib/bolt/puppetdb/client.rb
```diff
@@ -42,6 +42,22 @@ def facts_for_node(certnames)
...
+        return [] if certnames.empty? || certnames.nil?
+
+        certnames.uniq!
+        name_query = certnames.map { |c| ["=", "certname", c] }
+        name_query.insert(0, "or")
+
+        facts_query = facts.map { |f| ["=", "path", f] } unless facts.empty?
+        facts_query.insert(0, "or") unless facts.empty?
+
+        query = ['and', name_query, facts_query]
```
adreyer: two unlesses in a row suggest this should be a single conditional

What happens if facts is empty? Why is facts an optional argument?

donoghuc:
I think something is wrong with logging missing facts. For example, something like 
```yaml
      - name: windows-nodes
        target-lookups:
          - plugin: puppetdb
            query: inventory[certname]{facts.os.family = "windows"}
            name: facts.foo
            config:
              winrm:
                host: facts.bar
```
wont raise any waring. But something like 
```
    groups:
      - name: windows-nodes
        target-lookups:
          - plugin: puppetdb
            query: inventory[certname]{facts.os.family = "windows"}
            name: facts.foo
            config:
              winrm:
                host: facts.ipaddress
```
will
```
cas@cas-ThinkPad-T460p:~/working_dir/bolt/Boltdir$ bolt command run whoami -t windows-nodes 
Could not find fact facts.foo for node o3ju5xmklq1ptcc.delivery.puppetlabs.net
Started on o3ju5xmklq1ptcc.delivery.puppetlabs.net...
Finished on o3ju5xmklq1ptcc.delivery.puppetlabs.net:
  STDOUT:
    o3ju5xmklq1ptcc\administrator
Successful on 1 node: o3ju5xmklq1ptcc.delivery.puppetlabs.net
Ran on 1 node in 1.30 seconds
```
So it looks like the warning is only raised if there is at least one fact that is found.

lucywyman:
@donoghuc Fixed up and added a test for it!

PR #993, opened 2019-05-17
(BOLT-1264) Document puppetdb inventory plugin config

lucywyman:
This allows user who are using the PuppetDB inventory plugin to set config
values based on facts gathered from PuppetDB. This adds documentation including an example on how the feature works.

### DIFF
--- a/pre-docs/inventory_version2.md
+++ b/pre-docs/inventory_version2.md
@@ -61,7 +61,7 @@ targets:
 
 ### `target-lookups` and dynamic inventory
 
-`target-lookups` is a new key at the group level that allows you to dynamically
+`target-lookups` is a key at the group level that allows you to dynamically
 lookup the targets in the node. `target-lookups` contains an array of target
 lookup objects. Each `target-lookups` entry must include a `plugin` key that
 defines which plugin should be used for the lookup. The rest of the keys are
@@ -75,8 +75,16 @@ currently Bolt only ships builtin lookup plugins. The plugins are:
 
 #### PuppetDB
 
-The PuppetDB plugin takes a `query` field, which is either a string containing
-a PQL query or an array containing a PuppetDB AST format query.
+The PuppetDB plugin takes a `query` field, which is either a string
+containing a
+[PQL](https://puppet.com/docs/puppetdb/latest/api/query/v4/pql.html) query or
+an array containing a [PuppetDB
+AST](https://puppet.com/docs/puppetdb/latest/api/query/v4/ast.html) format
+query. The query is used to determine which targets should be included in the
+group. If `name` or `uri` is not specified with a [fact lookup](#fact-lookups)
+then the [certname] for each target in the query result will be used as the
+`uri` for the new target. Read the [Migrating to Version
+2](#migrating-to-version-2) section for more details on `uri` and `name` keys.
 
 ```
 groups:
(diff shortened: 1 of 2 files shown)


pre-docs/inventory_version2.md
```diff
@@ -90,6 +94,10 @@ groups:
     target-lookups:
       - plugin: puppetdb
         query: "inventory[certname] { facts.osfamily = 'RedHat' }"
+        config:
+          user: identity.user
```
donoghuc: I think these got stale. Here is one i've been using
```yaml
version: 2
groups:
  - name: test
    target-lookups:
      - plugin: puppetdb
        name: facts.ipaddress
        config:
          ssh:
            user: facts.identity.user
            host: facts.networking.hostname
    config:
      transport: ssh
      ssh:
        private-key: ~/.ssh/id_rsa-acceptance
```

lucywyman:
Yeah - I was going to wait until the #992 got merged, then update these to avoid having to keep them up to date continuously

pre-docs/inventory_version2.md
lucywyman: Updated!

pre-docs/inventory_version2.md
```diff
@@ -90,8 +95,14 @@ groups:
     target-lookups:
       - plugin: puppetdb
         query: "inventory[certname] { facts.osfamily = 'RedHat' }"
+        config:
```
adreyer: This example won't work for two reasons:

1. the config is not scoped to a transport so I think it will be ignored.
2. host won't be used because uri will be magically set.

I think we should leave the simple example where config is not set and then add a new more complex one that uses an ip address as the host.  We should also explain the magic and what values are available, (facts and certname)

```
name: certname
config:
   ssh:
     host: facts.networking.interfaces.en0.ipaddress
     user: facts.identity.user
```

donoghuc: I have tested the example i linked in https://github.com/puppetlabs/bolt/pull/993#discussion_r286259187 against PE nodes. I use `facts.networking.hostname` for the `host` and  `facts.ipaddress` for the target `name`. It seems like you are suggesting the opposite. I am wondering if there is something i have set in OpenSSH config that is making this magically work for me.

adreyer: that will work as long as the hostname is resolvable. The goal with my example is to connect to a node with an ip address without relying on dns.

pre-docs/inventory_version2.md
```diff
@@ -75,8 +75,14 @@ currently Bolt only ships builtin lookup plugins. The plugins are:
 
 #### PuppetDB
 
-The PuppetDB plugin takes a `query` field, which is either a string containing
-a PQL query or an array containing a PuppetDB AST format query.
+The PuppetDB plugin takes a `query` field, which is either a string
```
adreyer: I think this needs to be more referency and may need more examples. I don't think I could use it based on this description.

perhaps breaking it up into two sections one about making a query to select the nodes and a second about setting "config/name/uri" based on facts and certname.

lucywyman:
@adreyer Updated! I think the examples still aren't great - let me know if you have ideas on more thorough examples.

pre-docs/inventory_version2.md
```diff
@@ -75,8 +75,12 @@ currently Bolt only ships builtin lookup plugins. The plugins are:
...
+containing a
+[PQL](https://puppet.com/docs/puppetdb/latest/api/query/v4/pql.html) query or
+an array containing a [PuppetDB
+AST](https://puppet.com/docs/puppetdb/latest/api/query/v4/ast.html) format
+query.
```
adreyer: This is used to determine which targets should be included in the group. If no fact lookup is specified the certname for each item in the query result will used as the uri for the new target.

pre-docs/inventory_version2.md
```diff
@@ -96,6 +100,48 @@ groups:
 
 Make sure you have [configured PuppetDB](./bolt_connect_puppetdb.md)
 
+##### Fact Lookups
+
+The PuppetDB plugin also accepts fact lookups as values for `name`, `uri`,
+and values in the `config` hash which will be resolved to the facts for the
+nodes in the PuppetDB query. The fact lookups accept [PQL dot
```
adreyer: If more specific config is required the puppetdb plugin can be used to lookup configuration values for the `name`, `uri`, and `config` inventory options for each target. The fact lookup values can be either `certname` to reference the certaname of the target or a pql dot notation `facts` string such as `facts.os.family` to reference fact value.(link to dot notation docs)

note: If the `name` or `uri` values are set to a lookup the puppetdb inventory plugin will not set the `uri` to the certname of the target.

I'm not sure exactly what to do here but:
* dot notation is required for both structured and unstructured facts.
* I think listing the top level things that can be referenced `certname` and `facts` is clearer than  listing the cofig keys twice with each top level reference.
* I think it needs to mention the magic around when `uri` is set or unset.

lucywyman: I tried linking back up to the other docs that explain it.

adreyer:
I'm going to merge so that there are least some docs ready.

lucywyman:
@melamos FYI

PR #994, opened 2019-05-17
(BOLT-1269) Add prompt plugin

donoghuc:
This commit adds a prompt plugin. When data is loaded from inventory, values that the key '_plugin' points to will be used to look up available plugins by name. For the prompt plugin, a new instance of the prompt class is instantiated with the data in the same level of '_plugin' key in the hash. The `inventory_config_lookup` hook returns a concurrent delay which when executed prompts the user for a value.

### DIFF
--- a/lib/bolt/inventory/group2.rb
+++ b/lib/bolt/inventory/group2.rb
@@ -16,11 +16,16 @@ module Bolt
       GROUP_KEYS = DATA_KEYS + %w[groups targets target-lookups]
       CONFIG_KEYS = Bolt::TRANSPORTS.keys.map(&:to_s) + ['transport']
 
-      def initialize(data)
+      def initialize(data, plugins)
         @logger = Logging.logger[self]
-
         raise ValidationError.new("Expected group to be a Hash, not #{data.class}", nil) unless data.is_a?(Hash)
+        raise ValidationError.new("Cannot set group with plugin", nil) if data.key?('_plugin')
         raise ValidationError.new("Group does not have a name", nil) unless data.key?('name')
+        @plugins = plugins
+
+        %w[name vars features facts].each do |key|
+          validate_config_plugin(data[key], key, nil)
+        end
 
         @name = data['name']
         raise ValidationError.new("Group name must be a String, not #{@name.inspect}", nil) unless @name.is_a?(String)
@@ -34,7 +39,8 @@ module Bolt
         @vars = fetch_value(data, 'vars', Hash)
         @facts = fetch_value(data, 'facts', Hash)
         @features = fetch_value(data, 'features', Array)
-        @config = fetch_value(data, 'config', Hash)
+
+        @config = config_only_plugin(fetch_value(data, 'config', Hash))
 
         @target_lookups = fetch_value(data, 'target-lookups', Array)
 
@@ -60,8 +66,40 @@ module Bolt
           end
         end
 
-        @groups = groups.map { |g| Group2.new(g) }
+        @groups = groups.map { |g| Group2.new(g, plugins) }
       end
+
+      def validate_config_plugin(data, key, group_name = nil)
+        if data.is_a?(Hash) && data.include?('_plugin')
+          if group_name
+            raise ValidationError.new("Cannot set target #{key.inspect} with plugin", group_name)
+          else
+            raise ValidationError.new("Cannot set group #{key.inspect} with plugin", nil)
+          end
+        end
+        if data.is_a? Hash
+          data.each do |_k, v|
+            validate_config_plugin(v, key, group_name)
+          end
+        elsif data.is_a? Array
+          data.map { |v| validate_config_plugin(v, key, group_name) }
+        end
+      end
+      private :validate_config_plugin
+
+      def config_only_plugin(data)
+        Bolt::Util.walk_vals(data) do |value|
+          if value.is_a?(Hash) && value.include?('_plugin')
+            unless (plugin = @plugins.by_name(value['_plugin']))
+              raise ValidationError.new("unkown plugin: #{value['_plugin'].inspect}", nil)
+            end
+            plugin.new.inventory_config_lookup(value)
+          else
+            value
+          end
+        end
+      end
+      private :config_only_plugin
 
       def target_data(target_name)
         if (data = @targets[target_name])
@@ -82,6 +120,11 @@ module Bolt
         # they be handled?
         unless target.is_a?(Hash)
           raise ValidationError.new("Node entry must be a String or Hash, not #{target.class}", @name)
+        end
+        raise ValidationError.new("Cannot set target with plugin", @name) if target.key?('_plugin')
+        target.each do |k, v|
+          next if k == 'config'
+          validate_config_plugin(v, k, @name)
         end
 
         target['name'] ||= target['uri']
@@ -113,6 +156,7 @@ module Bolt
           @logger.warn(msg)
         end
 
+        target['config'] = config_only_plugin(target['config'])
         unless target.include?('alias')
           return
         end
@@ -145,7 +189,6 @@ module Bolt
 
           unless (plugin = plugins.by_name(lookup['plugin']))
             raise ValidationError.new("target-lookup specifies an unkown plugin: '#{lookup['plugin']}'", @name)
-
           end
 
           targets = plugin.lookup_targets(lookup)
--- a/lib/bolt/inventory/inventory2.rb
+++ b/lib/bolt/inventory/inventory2.rb
@@ -102,13 +102,13 @@ module Bolt
 
       def data_hash
         {
-          data: @data,
+          data: {},
           target_hash: {
-            target_vars: @target_vars,
-            target_facts: @target_facts,
-            target_features: @target_features
+            target_vars: {},
+            target_facts: {},
+            target_features: {}
           },
-          config: @config.transport_data_get
+          config: { transports: {} }
         }
       end
 
--- a/lib/bolt/inventory/inventory2.rb
+++ b/lib/bolt/inventory/inventory2.rb
@@ -16,13 +16,12 @@ module Bolt
         @logger = Logging.logger[self]
         # Config is saved to add config options to targets
         @config = config || Bolt::Config.default
-        @data = data ||= {}
-        @groups = Group2.new(data.merge('name' => 'all'))
+        @data = data || {}
+        @groups = Group2.new(@data.merge('name' => 'all'), plugins)
         @group_lookup = {}
         @target_vars = target_vars
         @target_facts = target_facts
         @target_features = target_features
-
         @groups.lookup_targets(plugins)
         @groups.resolve_aliases(@groups.target_aliases, @groups.target_names)
         collect_groups
@@ -121,6 +120,18 @@ module Bolt
       end
       private :groups_in
 
+      # Look for _plugins
+      def config_plugin(data)
+        Bolt::Util.walk_vals(data) do |val|
+          if val.is_a?(Concurrent::Delay)
+            val.value
+          else
+            val
+          end
+        end
+      end
+      private :config_plugin
+
       # Pass a target to get_targets for a public version of this
       # Should this reconfigure configured targets?
       def update_target(target)
@@ -138,6 +149,7 @@ module Bolt
         set_vars_from_hash(target.name, data['vars']) unless @target_vars[target.name]
         set_facts(target.name, data['facts']) unless @target_facts[target.name]
         data['features']&.each { |feature| set_feature(target, feature) } unless @target_features[target.name]
+        data['config'] = config_plugin(data['config'])
 
         # Use Config object to ensure config section is treated consistently with config file
         conf = @config.deep_clone
--- a/lib/bolt/plugin/prompt.rb
+++ b/lib/bolt/plugin/prompt.rb
@@ -1,0 +1,32 @@
+# frozen_string_literal: true
+
+require 'concurrent/delay'
+module Bolt
+  class Plugin
+    class Prompt
+      def initialize
+        # Might not need this
+        @logger = Logging.logger[self]
+      end
+
+      def self.name
+        'prompt'
+      end
+
+      def hooks
+        ['inventory_config_lookup']
+      end
+
+      def inventory_config_lookup(opts)
+        raise Bolt::ValidationError, "Prompt requires a 'message'" unless opts['message']
+        # Return a delay to only be evaluated when needed
+        Concurrent::Delay.new do
+          STDOUT.print "#{opts['message']}:"
+          value = STDIN.noecho(&:gets).chomp
+          STDOUT.puts
+          value
+        end
+      end
+    end
+  end
+end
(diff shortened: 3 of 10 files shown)


donoghuc:
Just wanted to push what I have been playing with for early feedback. So many things wrong with it currently...

donoghuc:
Question about behaviour: 
With the following inventory:
```yaml
version: 2
groups:
  - name: outter
    targets:
      - uri: s5xx09est930z86.delivery.puppetlabs.net
        name: pooler
        config:
          transport: ssh
          ssh:
            user: root
            password: ****
    groups:
      - name: inner
        targets:
          - name: localhost
    config:
      transport: local
      local:
        user:
          _plugin: prompt
          message: password please
```
When targeting `pooler`, prompted once (to inherit outter group config)
```
cas@cas-ThinkPad-T460p:~/working_dir/bolt$ bolt command run whoami -t pooler
password please:
Started on s5xx09est930z86.delivery.puppetlabs.net...
Finished on s5xx09est930z86.delivery.puppetlabs.net:
  STDOUT:
    root
Successful on 1 node: pooler
Ran on 1 node in 0.34 seconds
```
Similarly with `inner`
```
cas@cas-ThinkPad-T460p:~/working_dir/bolt$ bolt command run whoami -t inner
password please:
Started on ...
Finished on :
  STDOUT:
    cas
Ran on 1 node in 0.00 seconds
```
When 'outter` is targeted you are prompted twice. 
```
cas@cas-ThinkPad-T460p:~/working_dir/bolt$ bolt command run whoami -t outter
password please:
password please:
Started on s5xx09est930z86.delivery.puppetlabs.net...
Started on ...
Finished on :
  STDOUT:
    cas
Finished on s5xx09est930z86.delivery.puppetlabs.net:
  STDOUT:
    root
Successful on 2 nodes: pooler,localhost
Ran on 2 nodes in 0.33 seconds
```

lib/bolt/inventory/inventory2.rb
```diff
@@ -121,6 +121,25 @@ def groups_in(target_name)
       end
       private :groups_in
 
+      # Look for _plugins
+      def config_plugin(data)
+        if data.is_a? Hash
```
adreyer: I think you can use `Bolt::Util.walk_vals` for this

lib/bolt/plugin/prompt.rb
```diff
@@ -0,0 +1,27 @@
...
+        # Might not need this
+        @logger = Logging.logger[self]
+      end
+
+      def name
+        'prompt'
+      end
+
+      def hooks
+        ['lookup_value']
```
adreyer: If this hook is only going to be used inside a "config" block I think we should make it "inventory_config_lookup".

lib/bolt/inventory/inventory2.rb
```diff
@@ -138,6 +157,7 @@ def update_target(target)
         set_vars_from_hash(target.name, data['vars']) unless @target_vars[target.name]
         set_facts(target.name, data['facts']) unless @target_facts[target.name]
         data['features']&.each { |feature| set_feature(target, feature) } unless @target_features[target.name]
+        data['config'] = config_plugin(data['config'])
```
adreyer: I think there may be a bug here/`@groups.data_for()` where we deep merged config values from two sources one value is a hash of raw config data and the other is a plugin_lookup hash. Where the `_plugin` will get added to the raw data causing it to be evaluated as a plugin and throw away the raw data regardless of precedence order.

lib/bolt/inventory/inventory2.rb
```diff
@@ -17,12 +17,19 @@ def initialize(data, config = nil, plugins: nil, target_vars: {}, target_facts:
...
-        @groups = Group2.new(data.merge('name' => 'all'))
+        @data = Bolt::Util.walk_vals(data) do |value|
+          if value.is_a?(Hash) && value.include?('_plugin')
+            # How do we reliably instantiate new plugin object by name?
+            Object.const_get("Bolt::Plugin::#{value['_plugin'].capitalize}").new.inventory_config_lookup(value)
+          else
+            value
+          end
+        end
+        @groups = Group2.new(@data.merge('name' => 'all'))
```
donoghuc: In this case the init method has a parameter called `data` and an instance variable `@data`. When I iterate over the instance variable and convert `{'_plugin' => promt, ...}` I want to instantiate `@groups` with the modified `@data` version instead of `data` version. Do we intentionally *not* want to initialize `Group2` with the instance variable version?

lib/bolt/inventory/inventory2.rb
```diff
@@ -121,6 +128,19 @@ def groups_in(target_name)
       end
       private :groups_in
 
+      # Look for _plugins
+      def config_plugin(data)
+        Bolt::Util.walk_vals(data) do |value|
+          if value.is_a?(Concurrent::Promise)
+            value.execute
```
donoghuc: Will likely choose *not* to call the block parameter `value`...

lib/bolt/inventory/inventory2.rb
```diff
@@ -17,12 +17,19 @@ def initialize(data, config = nil, plugins: nil, target_vars: {}, target_facts:
         # Config is saved to add config options to targets
         @config = config || Bolt::Config.default
         @data = data ||= {}
-        @groups = Group2.new(data.merge('name' => 'all'))
+        @data = Bolt::Util.walk_vals(data) do |value|
+          if value.is_a?(Hash) && value.include?('_plugin')
+            # How do we reliably instantiate new plugin object by name?
```
donoghuc: I realize the way it has been implemented in this WIP is not correct. Would like to discuss how to do this.

spec/integration/inventory2_spec.rb
```diff
@@ -496,4 +496,50 @@ def fact_plan(name = 'facts_test')
...
+              password: { _plugin: 'prompt', message: 'password please' },
+              user: conn[:user],
+              host: conn[:host],
+              port: conn[:port],
+              'host-key-check': false
+            }
+          }
+        },
+                  {
+                    name: 'target-2',
```
donoghuc: Tired of fighting rubocop on this. For some reason it insists this is the correct indentation.

lib/bolt/inventory/inventory2.rb
```diff
@@ -110,13 +110,13 @@ def features(target)
 
       def data_hash
         {
-          data: @data,
+          data: {},
           target_hash: {
-            target_vars: @target_vars,
-            target_facts: @target_facts,
```
donoghuc: This should ensure that get_targets always returns an empty array. Previously the catalog data was serializing the object instance of `Concurrent::Delay` in a way that would not be usable within the apply block. https://github.com/puppetlabs/bolt/blob/71d37234dc8c4635c147c5fdd6c650649cf7dec7/lib/bolt/applicator.rb#L104
Example plan:
```
plan test(TargetSpec $nodes){
  apply($nodes){
  	notice(get_targets('all'))
  }
}
```
Inventory v1
```yaml
nodes:
  - name: s0b58ex4u43wq6y.delivery.puppetlabs.net
    alias: linux
    config:
      transport: ssh
      ssh:
        host-key-check: false
        user: root
```
```
cas@cas-ThinkPad-T460p:~/working_dir/bolt/Boltdir$ bolt plan run test -t linux
Starting: plan test
Starting: plan test
Starting: apply catalog on s0b58ex4u43wq6y.delivery.puppetlabs.net
s0b58ex4u43wq6y.delivery.puppetlabs.net: Scope(Class[main]): [Target('s0b58ex4u43wq6y.delivery.puppetlabs.net', {"user"=>"root"})]
Finished: apply catalog with 0 failures in 4.41 sec
Finished: plan test in 4.41 sec
Finished: plan test in 4.41 sec
```
inventory v2
```yaml
version: 2
targets: 
  - uri: s0b58ex4u43wq6y.delivery.puppetlabs.net
    alias: linux
    config:
      transport: ssh
      ssh:
        host-key-check: false
        user: root
```
```
cas@cas-ThinkPad-T460p:~/working_dir/bolt/Boltdir$ bolt plan run test -t linux
Starting: plan test
Starting: plan test
Starting: apply catalog on s0b58ex4u43wq6y.delivery.puppetlabs.net
s0b58ex4u43wq6y.delivery.puppetlabs.net: Scope(Class[main]): []
Finished: apply catalog with 0 failures in 4.4 sec
Finished: plan test in 4.41 sec
Finished: plan test in 4.41 sec
```

lib/bolt/inventory/inventory2.rb
donoghuc: I think with the updated implementation that converts every instance of a `{'_plugin' => ... }` hash to a concurrent delay instance before instantiating a `Group2` object we avoid this issue with `deep_merge`. https://github.com/puppetlabs/bolt/pull/994/commits/21fd90e2a2748cf6c638eba380c82faa47bc35d3#diff-8f5017c5ca9a2d12a104805c2a404110R23

lib/bolt/plugin/prompt.rb
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+
+require 'concurrent'
```
donoghuc: This is causing a regression in our integration test for lazily loading gems (https://github.com/puppetlabs/bolt/blob/71d37234dc8c4635c147c5fdd6c650649cf7dec7/spec/integration/cli_spec.rb#L30). Eventually will we need a solution for loading gem libraries specific to the plugin being used? Also it looks like I may be able to just require `concurrent/delay`.

adreyer: Long term I think we will lazily require/load this file so plain requires should be fine.

spec/integration/inventory2_spec.rb
donoghuc: finally won the battle and got reasonable indentation.

lib/bolt/plugin/prompt.rb
donoghuc: cool. that makes life easy. For now i'm removing concurrent-ruby from https://github.com/puppetlabs/bolt/pull/994/commits/b9ce42bb0c8f2e41fcd2156420b94b9eb79700ee#diff-fb52e06d3eb62217257cc824e4323d80L30

lib/bolt/inventory/group2.rb
```diff
@@ -118,13 +117,14 @@ def add_target(target)
         end
 
         aliases = target['alias']
-        aliases = [aliases] if aliases.is_a?(String)
+        aliases = [aliases] if aliases.is_a?(String) || aliases.is_a?(Concurrent::Delay)
```
donoghuc: Example
```yaml
version: 2
targets: 
  - uri: j6gppqkp45z18uo.delivery.puppetlabs.net
    alias: 
      - _plugin: prompt
        message: alias name
      - foo
    config:
      transport: ssh
      ssh:
        host-key-check: false
        user: 
          _plugin: prompt
          message: ssh user
```
alias name: prompt-me
ssh user: root
```
cas@cas-ThinkPad-T460p:~/working_dir/bolt$ bolt command run whoami -t prompt-me
alias name:
ssh user:
Started on j6gppqkp45z18uo.delivery.puppetlabs.net...
Finished on j6gppqkp45z18uo.delivery.puppetlabs.net:
  STDOUT:
    root
Successful on 1 node: j6gppqkp45z18uo.delivery.puppetlabs.net
Ran on 1 node in 0.43 seconds
```

donoghuc: This seems like a bad idea. I will continue working on preventing this outside of config.

lib/bolt/inventory/inventory2.rb
```diff
@@ -199,6 +212,41 @@ def expand_targets(targets)
       end
       private :expand_targets
 
+      def resolve_plugins(plugins, data)
+        if data.is_a? Hash
+          data.each_with_object({}) do |(k, v), acc|
+            v = case k
+                when 'config'
+                  config_only_plugin(plugins, v)
+                when 'target-lookups'
```
donoghuc: Anything under the `config` key can be safely converted to concurrent delay *unless* it is nested under a `target-lookups` key. The `config_only_plugin` method should only be called on the value pointed to by a `config` key that is not nested under a `target-lookups` key.

pre-docs/inventory_version2.md
```diff
@@ -229,6 +229,26 @@ google_compute_instance.app.1:
   self_link = https://www.googleapis.com/compute/v1/projects/cloud-app1/zones/us-west1-a/instances/app-1
   zone = us-west1-a
 ```
+#### Prompt
```
adreyer: I think we need two sections since we have two types of plugins now.

adreyer:
trying to set `transport` to a prompt value causes an error. I don't think this is super valuable but I do think we need to support all of `config` for consistency

donoghuc:
Interesting "bug" when using `out::message`.
```yaml
version: 2
groups:
  - name: plugin
    targets: 
      - uri: wo70nknfq0x75on.delivery.puppetlabs.net
        alias:
          - foo
        config:
          transport: ssh
          ssh:
            host-key-check: false
            user: 
              _plugin: prompt
              message: ssh user
```
```
  out::message("I will be on same line as prompt")
  run_command('whoami', get_targets('all'))
}
```
```
cas@cas-ThinkPad-T460p:~/working_dir/bolt$ bolt plan run test -t localhost
Starting: plan test
Starting: plan test
ssh user:I will be on same line as prompt

Starting: command 'whoami' on wo70nknfq0x75on.delivery.puppetlabs.net
Finished: command 'whoami' with 0 failures in 0.41 sec
Finished: plan test in 4.05 sec
Finished: plan test in 4.05 sec
```
Not quite sure how we could get around that one. Note that is still works as expected, it just looks kind of weird.

lib/bolt/inventory/group2.rb
```diff
@@ -60,9 +66,41 @@ def initialize(data)
...
         end
 
-        @groups = groups.map { |g| Group2.new(g) }
+        @groups = groups.map { |g| Group2.new(g, plugins) }
       end
 
+      def validate_config_plugin(data, key, group_name = nil)
+        if data.is_a?(Hash) && data.include?('_plugin')
+          if group_name
+            raise ValidationError.new("Cannot set target #{key.inspect} with config plugin", group_name)
```
adreyer: I think we should generalize these messages to something like "No plugin type  available for setting #{key.inspect} on a target"

pre-docs/inventory_version2.md
```diff
@@ -104,7 +109,7 @@ groups:
 
 Make sure you have [configured PuppetDB](./bolt_connect_puppetdb.md)
 
-##### Fact Lookups
+###### PuppetDB Fact Lookups
```
melamos: This heading level is toooo deep; it won't show up properly on the docs site. Can we get rid of this heading entirely and make the discussion of PuppetDB fact lookups part of the section about the PuppetDB plugin?

pre-docs/inventory_version2.md
```diff
@@ -234,6 +241,33 @@ google_compute_instance.app.1:
   zone = us-west1-a
 ```
 
+#### `data-lookup` plugins
```
adreyer: I think we should jsut call these config-lookup plugins until we can use them outside config.

PR #996, opened 2019-05-20
(BOLT-1324) Install Microsoft OMI server in Travis

Iristyle:
This PR provides the basis for a few additional features:

* More robust PowerShell on Linux support (changes required to underlying winrm gem)
* Linux -> Linux PowerShell over WinRM with Kerberos authentication (serving as a representation of the same Kerb support on Windows) - PR #950 started this effort
* PowerShell supporting using SSH as the transport
* Kerberos support over SSH (the next PR adding KDC will build on this)

Details about the implementation:

 - Update the docker-compose file to provision a new Ubuntu 18.04 node
   with Microsoft OMI server (currently 1.6.0-0 from Oct 16 2018).

   OMI server will setup a WinRM endpoint on Linux on both ports 5985
   and 5986 which Bolt can communicate with.

 - To allow it to execute PowerShell requires installing the PowerShell
   package (currently 6.2.0) and the PSRP plugin for OMI server
   (currently 1.4.2-2).

 - For now, configure NTLM authentication using the NTLM auth file and
   add the standard bolt:bolt user:password used throughout tests.

   Like other test fixtures, expose port 25895 / 25986 for HTTP / HTTPS
   so that existing tests can be leveraged

 - NOTE: Currently the winrm gem does not fully support winrm to Linux
   execution at feature parity with winrm to Windows, so therefore
   only the most basic of tests will pass - those that verify SSL
   connectivity over WinRM.

 - Added a new `omi` tag that will be applied to relevant tests until
   it can be merged with `winrm`. The `puppetserver` Rake task which
   leverages the compose stack sets this flag as well.

 - NOTE: OMI Server also does not support Basic authentication over
   HTTP, only Basic over HTTPS. However, the current certificate is
   auto-generated, so the only enabled test is the one that disables SSL
   verification and removes the CA cert from the connection config.

 - NOTE: There is no $ENV:UserName on Linux, but both Windows and Linux
   can both use the .NET API `[Environment]::UserName`

 - Generate a new SSL certificate given the CA private key that also
   includes a few new hostnames - omiserver, omi, spec_omiserver_1 and
   spec_omi_1

   Furthermore it includes the IP 127.0.0.1

 - Include a new helper script and openssl metadata files to easily
   regenerate certs in the future

 - Update specs to run a few more winrm based PowerShell specs against
   OMI server, this time given the root CA as trusted

 - The AppVeyor and Windows vagrant cert.pfx were generated from a
   different root CA than all the other cert fixtures used in tests.

   This created a slight problem when testing WinRM over HTTPS when
   tests run in both Linux and Windows as it would create divergent
   testing resources.

 - Regenerate a new PFX based on the SSL private key / cert pair stored
   in spec/fixtures/ssl, and remove the old information.

   Add the openssl invocation to produce the PFX file in the
   createcerts.sh helper script

 - Update Vagrant and AppVeyor to use the new cert paths, noting that
   the PowerShell API has to use CN=boltserver, rather than CN=localhost
   to not cause an error. The pfx has a subject alt name of localhost


Iristyle:
The additional container build and spec run adds somewhere between 20 and 30 seconds to the overall Travis build.

Iristyle:
After provisioning with`docker-compose -f spec/docker-compose.yml up --build`, another easy interactive test to verify is to use bolt to query the Linux host:

```
bundle exec bolt command run "\$PSVersionTable" --nodes winrm://localhost:25986 --ssl --user bolt --password bolt --no-ssl-verify
```

Output should be 

```
Started on localhost...
Finished on localhost:
  STDOUT:

    PSVersion                      6.2.0
    PSEdition                      Core
    GitCommitId                    6.2.0
    OS                             Linux 4.9.125-linuxkit #1 SMP Fri Sep 7 08:20:28 UTC 2018
    Platform                       Unix
    PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0…}
    PSRemotingProtocolVersion      2.3
    WSManStackVersion              3.0

Successful on 1 node: winrm://localhost:25986
Ran on 1 node in 2.02 seconds
```

PR #998, opened 2019-05-20
(maint) Include source samples directly from files

nicklewis:
Previously, we had all the source code samples copy-pasted into the README
files, which was a burden to keep in sync with the actual code on disk. We
now use the include_relative function to load the content directly from the
files, ensuring it's always in sync.

This PR also adds the layout html from our jekyll theme into the repo
and modifies it to set page title/time/difficulty based on jekyll front
matter.

This also adds an explicit link for the favicon, which seems to be necessary
for GH pages. And it renames the README.md to index.md files to ensure they are
used as the index for their respective directories, which was broken by the
addition of the front matter.


nicklewis:
Ughhh it looks like it didn't properly detect the renames, so it will definitely be easier to review this by commit. Or just by looking at the site, which shouldn't actually have been significantly changed by this PR.

docs/_layouts/default.html
```diff
@@ -0,0 +1,46 @@
...
+      <script>
+        window.dataLayer = window.dataLayer || [];
+        function gtag(){dataLayer.push(arguments);}
+        gtag('js', new Date());
+        gtag('config', '{{ site.google_analytics }}');
+      </script>
+    {% endif %}
+    <meta charset="UTF-8">
+
+{% seo %}
```
lucywyman: Ooo what does this tag do?

PR #999, opened 2019-05-21
(BOLT-1472) Setup Kerberos in Container Tests

Iristyle:
* builds on ~~#996~~
* blocked by ~~#1087~~
* follow-on dev PR setup at #1093 

 - Ensure that Basic auth with credentials can be used over HTTP(S)
   against the OMI server setup

   This basically verifies that the NTLM authentication file is properly
   setup for use with OMI server and contains bolt:$BOLT_PASSWORD
   where $BOLT_PASSWORD is setup during container build

 - These checks are performed with both omicli *and* pwsh. The scripts
   are separated so that the pwsh tests can be performed from another
   container used for development

 - The current docker-compose hostname: omiserver is sufficient for the
   container to correctly self-identify

 - This establishes an Active Directory server in the Docker compose
   stack, which is one of the easier ways on Linux to establish a KDC
   (Key Distribution Center). While originally inspired by a simpler
   KDC setup at gcavalcante8808/docker-krb5-server
   it became necessary to setup full AD to be able to interoperate with
   OMI server.

   Samba implements not just the AD on Linux, but also sets up DNS SRV
   records for critical addresses like _ldap._tcp.bolt.test

   The compose stack is configured with a new BOLT.TEST network so that
   strict control over the subnet can be setup to give the Samba DNS a
   static IP address, which is critically necessary for it behave
   correctly. Since Kerberos relies heavily on network identity / DNS
   names, it's important to configure the network properly. It's also
   necessary to add the kernel capability SYS_ADMIN through Compose.

 - Samba is setup to own the realm BOLT.TEST and the short name BOLT,
   the provisioning of which happens in samba-ad-config.sh

   The smb.conf in use is instructed to use the Docker DNS of 127.0.0.11
   which is configured in docker-compose.yml to redirect back to the
   address of the Samba container.

   DNS -> 127.0.0.11 -> 172.22.0.100

 - Setup TravisCI to register samba-ad.bolt.test as an entry in
   /etc/hosts so that Bolt may acquire a ticket for the realm BOLT.TEST.
   It's also necessary to specify the correct krb5.conf settings via the
   KRB5_CONFIG env var, pointing it to /tmp/krb5.conf which is generated
   by helper script kerberos-client-config.sh

 - kerberos-client-config.sh takes a few environment variables:

   KRB5_REALM - realm name, typically BOLT.TEST
   KRB5_KDC - KDC DNS name, typically samba-ad.bolt.test
   KRB5_ADMINSERVER (optional) - KDC Admin DNS name, same as above
   KRB5_CONFIG (optional) - path to config, default to /etc/krb5.conf

   The script is currently used in Travis to dynamically produce the
   necessary config file, but will also be used in the future to
   produce the same file for OMI server

 - Add a single test under the :kerberos tag that performs a regex
   match against the `klist` output to make sure that a single ticket
   has been acquired, so that the test may be run in TravisCI and
   locally on OSX dev environments.

   NOTE: OSX can only acquire tickets, and will not be able to
   participate in future testing because of its Kerberos library.

   NOTE: Windows has a very different klist command line tool with
   wildly different output, and Windows is not tested in this PR

 - To configure Docker on OSX to communicate with the Samba KDC to
   acquire tickets requires several setup steps.

   * Adding an entry to /etc/hosts like:

     127.0.0.1 samba-ad.bolt.test

   * export KRB5_CONF=$(pwd)/spec/fixtures/kdc/krb5.osx.conf, the
     contents of which are setup to match sensible defaults:

     [realms]
     BOLT.TEST = {
       kdc = tcp/samba-ad.bolt.test
       admin_server = tcp/samba-ad.bolt.test
     }

     [domain_realm]
     bolt.test = BOLT.TEST
     .bolt.test = BOLT.TEST

     Alternatively, those settings may be more permanently added to
     /etc/krb5.conf if desired

   * NOTE: tcp/ is necessary Kerberos will try to use UDP by default
     which is not available to OSX Docker by default

 - To acquire the ticket granting ticket, use the command:

    KRB5_TRACE=/dev/stdout kinit Administrator@BOLT.TEST

    The current password is "B0ltrules!" as specified in docker-compose
    with the variable SMB_ADMIN_PASSWORD. KRB5_TRACE is not necessary
    but provides additional debug output that can be useful

    To destroy the ticket, use the command:

    kdestroy --credential=krbtgt/BOLT.TEST@BOLT.TEST

 - Install additional packages to support joining to the running Samba
   domain controller - time synchronization, Kerberos client libraries
   and various Samba tools

 - For the omiserver container to join the Samba domain, it must be on
   the same network, responding to the same DNS. Configure the Docker
   network accordingly to ensure that the containers are on the same
   subnet, and that the DNS alias samba-ad.bolt.test DNS resolves properly.

 - This does a basic domain join, configuring user lookup with sssd,
   which includes the following steps from the OMI server container:

   * Verify DNS resolution works for _ldap._tcp.bolt.test
   * Grab the Kerberos TGT (ticket granting ticket) for the domain
     Administrator account
   * Emit the nsswitch.conf for debug purposes
   * Configure realmd service with the specified domain and updating
     the smb.conf with config values for:
     - workgroup
     - realm
     - kerberos method = secrets and keytabs
   * Execute a realm join
   * Execute a net ads join
     - theoretically only one join should be necessary
   * Configure sssd with the given domain / realm
   * Emit relevant sssd info from /etc/pam.d config files
   * Start the sssd service
   * Verify realm discovered properly
   * Use `net ads info` to emit domain information
   * Verify sssd service running
   * Verify domain admin lookup succeeds with getent passwd

 - This has the added side effect of enabling OMI server SPNEGO based
   authentication to work properly, which is demonstrated by using both
   omicli *and* pwsh remoting

 - Now that there is a Samba server available to issue Kerberos tickets,
   configure OMI to use it, also adding the HTTP service SPN

 - Setup is based on instructions from OMI documentation at
   Microsoft/omi:Unix/doc/setup-kerberos-omi.md@master#ubuntu-linux-1604-lts-and-1404-lts-1610-yakkety-and-1704-zesty

 - Unfortunately, there is a bug in the winrm gem preventing it from
   properly interoperating with OMI server when Kerberos authentication
   is enabled. Therefore tests are still marked as pending.

   This will be fixed in tickets.puppetlabs.com/browse/BOLT-1476

### DIFF
--- a/developer-docs/kerberos.md
+++ b/developer-docs/kerberos.md
@@ -81,6 +81,8 @@ winrm:
 
 In Bolt testing atop Docker containers, [Samba server](https://www.samba.org/) is setup on Linux to approximate an Active Directory setup, which also includes DNS and LDAP support. Given Kerberos has very strict requirements around computer identity (via DNS), Docker user-defined networks with custom DNS and subnets are easier to setup than running in an arbitrary network environment. This lends itself well to an automated and reproducable ephemeral Kerberos environment.
 
+[OMI Server](https://github.com/microsoft/omi) requires the additional Active Directory-like features beyond just a KDC to enable Kerberos based authentication. OMI provides a PowerShell WinRM endpoint on Linux that is intended to behave like the equivalent Windows endpoint.
+
 This environment is intended to support multiple environments:
 
 * [TravisCI automated testing](#travisCI-automated-testing)
@@ -88,15 +90,15 @@ This environment is intended to support multiple environments:
 
 #### Container Setup
 
-The current `spec/docker-compose.yml` supports a number of containers, many of which are intended to be built when started. For Kerberos, build / start just the relevant container:
+The current `spec/docker-compose.yml` supports a number of containers, many of which are intended to be built when started. For Kerberos, build / start just the relevant containers:
 
-`docker-compose -f spec/docker-compose.yml up -d --build samba-ad`
+`docker-compose -f spec/docker-compose.yml up -d --build samba-ad omiserver`
 
 ##### Samba AD (KDC)
 
-A Kerberos server is provided by running an Alpine container with Samba as an Active Directory domain controller. The Kerberos realm is `BOLT.TEST`, Active Directory domain is `BOLT.TEST` (short name `BOLT`) and DNS suffix is `bolt.test`
+A Kerberos server is provided by running an Alpine container with Samba as an [Active Directory domain controller](https://wiki.samba.org/index.php/Setting_up_Samba_as_an_Active_Directory_Domain_Controller). The Kerberos realm is `BOLT.TEST`, Active Directory domain is `BOLT.TEST` (short name `BOLT`) and DNS suffix is `bolt.test`
 
-This container provides DNS and LDAP support, and a variety of [other services](https://wiki.samba.org/index.php/Samba_AD_DC_Port_Usage) including:
+This container provides DNS and LDAP support, but does not contain NTP as that is already provided in a Docker environment. It also hosts a variety of [other services](https://wiki.samba.org/index.php/Samba_AD_DC_Port_Usage) including:
 
 ###### External Ports / Services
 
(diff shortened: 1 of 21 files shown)


developer-docs/kerberos.md
```diff
@@ -77,4 +77,140 @@ winrm:
 - `bolt command run 'whoami' --targets winrm://dc.domain.com` to connect over HTTPS (`--no-ssl-verify` may be required if the target uses a self-signed certificate)
 - `bolt command run 'whoami' --targets winrm://dc.domain.com --no-ssl` to connect over HTTP
 
-In the future, this testing will be automated.
+### Testing with Docker
+
+In Bolt testing atop Docker containers, [Samba server](https://www.samba.org/) is setup on Linux to approximate an Active Directory setup, which also includes DNS and LDAP support. Given Kerberos has very strict requirements around computer identity (via DNS), Docker user-defined networks with custom DNS and subnets are easier to setup than running in an arbitrary network environment. This lends itself well to an automated and reproducable ephemerable Kerberos environment.
```
```suggestion
In Bolt testing atop Docker containers, [Samba server](https://www.samba.org/) is setup on Linux to approximate an Active Directory setup, which also includes DNS and LDAP support. Given Kerberos has very strict requirements around computer identity (via DNS), Docker user-defined networks with custom DNS and subnets are easier to setup than running in an arbitrary network environment. This lends itself well to an automated and reproducable ephemerable Kerberos environment.
```
```suggestion
In Bolt testing atop Docker containers, [Samba server](https://www.samba.org/) is setup on Linux to approximate an Active Directory setup, which also includes DNS and LDAP support. Given Kerberos has very strict requirements around computer identity (via DNS), Docker user-defined networks with custom DNS and subnets are easier to setup than running in an arbitrary network environment. This lends itself well to an automated and reproducable ephemeral Kerberos environment.
```

developer-docs/kerberos.md
```diff
@@ -77,4 +77,140 @@ winrm:
...
+* [`realm`](https://www.systutorials.com/docs/linux/man/8-realm/) - manages enrollment in Kerberos realms and Active Directory domains
+* [`net`](https://www.samba.org/samba/docs/current/man-html/net.8.html) - designed to work like the `net` tool on Windows
+* [`pwsh`](https://docs.microsoft.com/en-us/powershell/scripting/install/installing-powershell-core-on-linux?view=powershell-6) - PowerShell 6
+* [`omicli`](https://github.com/microsoft/omi/blob/master/Unix/cli/examples.txt) - client tool used to verify basic OMI server functionality
+
+#### TravisCI automated testing
+
+At present, Travis setup will:
+
+* Ensure that it can refer to itself as `samba-ad.bolt.test` and `omiserver.bolt.test` (the same name that the containesr refer to themselves inside the Docker UDN)
```
```suggestion
* Ensure that it can refer to itself as `samba-ad.bolt.test` and `omiserver.bolt.test` (the same name that the containers refer to themselves inside the Docker UDN)
```

developer-docs/kerberos.md
```diff
@@ -77,4 +77,140 @@ winrm:
...
+* Install the Kerberos client package
+* Configure the Kerberos client with the approriate server (`samba-ad.bolt.test`) for the realm `BOLT.TEST`
+
+Automated tests (in `spec/bolt/transport/winrm_spec.rb`) simplify verify that the correct TGT can be acquired from the Samba AD using `kinit` using the domain administrator account `Administrator@BOLT.TEST`.
```
```suggestion
Automated tests (in `spec/bolt/transport/winrm_spec.rb`) verify that the correct TGT (ticket granting ticket) can be acquired from the Samba AD using `kinit` using the domain administrator account `Administrator@BOLT.TEST`.
```

spec/Dockerfile.omiserver
```diff
@@ -3,12 +3,22 @@
 FROM ubuntu:18.04
 
 ARG BOLT_PASSWORD=bolt
+# NOTE: Only designed to be set at build time, not at run time!
+ENV BOLT_PASSWORD=${BOLT_PASSWORD}
+ENV KRB5_REALM=
```
lucywyman: Should these be getting set? Do they need to be here if not?

spec/bolt/transport/winrm_spec.rb
```diff
@@ -223,13 +227,19 @@ def stub_winrm_to_raise(klass, message)
       make_target(host_: 'omiserver.bolt.test', port_: 45986, conf: conf)
     end
 
+    # verifies the local setup has already acquired the right Kerberos ticket/
```
lucywyman:
```suggestion
    # verifies the local setup has already acquired the right Kerberos ticket
```

spec/Dockerfile.omiserver
Iristyle: They are passed in through the `docker-compose.yml` declaration of the `omiserver` service.

PR #1004, opened 2019-05-23
(BOLT-1337) Make ResultSet.filter return ResultSet

reidmv:
This should be possible/easy:

    $reboot_targets = $patch_results.ok_set.filter_set |$result| {
      $result['reboot_flag'] == true
    }.targets

This commit makes it work by adding a `.filter_set` method which acts the same as `.filter`, but returns a ResultSet datatype.

The existing `.filter` method was insufficient because it returned a Tuple, not a ResultSet.

### DIFF
--- a/pre-docs/writing_plans.md
+++ b/pre-docs/writing_plans.md
@@ -278,6 +278,7 @@ A `ResultSet` has the following methods:
 -   `error_set()`: A `ResultSet`containing only the results of failed nodes.
 
 -   `ok_set()`: A `ResultSet` containing only the successful results.
+-   `filter_set(block)`: Filters a `ResultSet` with the given block and returns a `ResultSet` object (where [Puppet's filter function](https://puppet.com/docs/puppet/6.4/function.html#filter) returns an array or hash)
 
 -   `targets()`: An array of all the `Target` objects from every `Result`in the set.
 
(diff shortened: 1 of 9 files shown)


adreyer:
When we originally discussed there I was a bit concerned this would be a breaking change. 

Since `ResultSet` is largely array like maybe it's not a problem. The main case I think would be if you were passing to result to something that had a type check.

lucywyman:
Failure is just rubocop

reidmv:
Pushing rubocop fix

adreyer:
after some discussion the bolt team has decided that this would be a breaking change. Our recommendation is to rename the new `filter` function to `filter_set`. 

Docs should be added https://github.com/puppetlabs/bolt/blob/master/pre-docs/writing_plans.md#handling-plan-function-results but we can handle that if you don't want to.

reidmv:
@adreyer Updated the PR to `.filter_set`. Would definitely appreciate you guys handling the docs. :)

pre-docs/writing_plans.md
```diff
@@ -278,6 +278,7 @@ A `ResultSet` has the following methods:
 -   `error_set()`: A `ResultSet`containing only the results of failed nodes.
 
 -   `ok_set()`: A `ResultSet` containing only the successful results.
+-   `filter_set(block)`: Filters a `ResultSet` with the given block and returns a `ResultSet` object (where [Puppet's Iterable](https://puppet.com/docs/puppet/6.4/puppet-6.4/content/documents/functions_iterative.html) `filter` returns an array)
```
donoghuc: This could be the wrong docs, but I think technically filter can return any type? https://puppet.com/docs/puppet/6.4/function.html#filter

pre-docs/writing_plans.md
```diff
@@ -351,6 +352,13 @@ $r = run_command('whoami', 'localhost,local://0.0.0.0')
 $r.to_data.each |$result_hash| { notice($result_hash['result']['stdout']) }
 ```
 
+You can also use `filter_set` to filter a ResultSet and apply a ResultSet function to the output:
```
donoghuc: Maybe call out that `.targets` is a method available on a `ResultSet` (something you would not get with an array).

PR #1005, opened 2019-05-24
(BOLT-1325) Add out::message function to print a message

nicklewis:
Previously, plans used the built-in Puppet logging functions
(info/notice/etc) to print output for the user. This doesn't work well
now that the default log level for the console is warn. It also doesn't
match our expected semantics around when logging should be used vs. when
the outputter should be used.

We have a new out::message function that can be used to send output to
the user via the outputter. This function takes a message and directs it
to the console if using the human outputter. Currently, the message is
ignored when using the JSON outputter.

### DIFF
--- a/bolt-modules/out/lib/puppet/functions/out/message.rb
+++ b/bolt-modules/out/lib/puppet/functions/out/message.rb
@@ -1,0 +1,29 @@
+# frozen_string_literal: true
+
+# Output a message for the user.
+#
+# This will print a message to stdout when using the human output format.
+#
+# **NOTE:** Not available in apply block
+Puppet::Functions.create_function(:'out::message') do
+  # Output a message.
+  # @param message The message to output.
+  # @example Print a message
+  #   out::message('Something went wrong')
+  dispatch :output_message do
+    param 'String', :message
+    return_type 'Undef'
+  end
+
+  def output_message(message)
+    unless Puppet[:tasks]
+      raise Puppet::ParseErrorWithIssue
+        .from_issue_and_stack(Bolt::PAL::Issues::PLAN_OPERATION_NOT_SUPPORTED_WHEN_COMPILING, action: 'out::message')
+    end
+
+    executor = Puppet.lookup(:bolt_executor)
+    executor.publish_event(type: :message, message: message)
+
+    nil
+  end
+end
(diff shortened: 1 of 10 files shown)


bolt-modules/out/lib/puppet/functions/out/message.rb
```diff
@@ -0,0 +1,33 @@
...
+Puppet::Functions.create_function(:'out::message') do
+  COLORS = %w[red green yellow].freeze
+
+  dispatch :output_message do
+    param 'String', :message
+    optional_param 'Hash[String[1], Any]', :options
+  end
+
+  def output_message(message, options = {})
+    executor = Puppet.lookup(:bolt_executor) { nil }
```
donoghuc: Instead of ignoring `out::message` in an apply block should we instead raise an error like the other functions that are not allowed in an apply block? 
```
    unless Puppet[:tasks]
        .from_issue_and_stack(Bolt::PAL::Issues::PLAN_OPERATION_NOT_SUPPORTED_WHEN_COMPILING, action: 'out::message')
    end
```

bolt-modules/out/lib/puppet/functions/out/message.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+Puppet::Functions.create_function(:'out::message') do
+  COLORS = %w[red green yellow].freeze
```
donoghuc: Still need puppet strings docs.

bolt-modules/out/lib/puppet/functions/out/message.rb
nicklewis: Good call, will do

nicklewis:
Updated this to remove the color feature, since it's not clear that's the right abstraction. We can revisit that functionality later.

spec/bolt/outputter/human_spec.rb
```diff
@@ -285,4 +285,15 @@
     outputter.fatal_error(Bolt::CLIError.new("oops"))
     expect(output.string).to eq("oops\n")
   end
+
+  it "handles message events" do
+    outputter.print_message_event(message: "hello world")
+    expect(output.string).to eq("hello world\n")
+  end
+
+  it "colorizes message events" do
```
donoghuc: I think removing the color is a good idea. Need to update the test.

nicklewis: Ugh I missed that test.

PR #1010, opened 2019-05-30
(tmp) DO NOT MERGE

donoghuc:
make sure that the new travis check that ensures plan_functions.md is updated.

### DIFF
--- a/bolt-modules/boltlib/lib/puppet/functions/add_facts.rb
+++ b/bolt-modules/boltlib/lib/puppet/functions/add_facts.rb
@@ -6,7 +6,7 @@ require 'bolt/error'
 #
 # **NOTE:** Not available in apply block
 Puppet::Functions.create_function(:add_facts) do
-  # @param target A target.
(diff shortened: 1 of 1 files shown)


donoghuc:
Fails as expected and shows the diff. 
```
diff --git a/pre-docs/plan_functions.md b/pre-docs/plan_functions.md
index 04acbe6..7c838cf 100644
--- a/pre-docs/plan_functions.md
+++ b/pre-docs/plan_functions.md
@@ -14,7 +14,7 @@ add_facts(Target $target, Hash $facts)
 
 *Returns:* `Hash[String, Data]` The target's new facts.
 
-* **target** `Target` A target.
+* **target** `Target` A target. oops forgot to run docs rake task
 * **facts** `Hash` A hash of fact names to values that may include structured facts.
 
 **Example:** Adding facts to a target
The command "if ! bundle exec rake docs; then
  false
elif ! git diff --exit-code; then
  false
fi
" exited with 1.
```

PR #1011, opened 2019-05-30
(maint) Temporarily pin cri gem so that r10k works for test setup

donoghuc:
It appears that the new cri release (2.15.7) introduces a breaking change to r10k which is used to install modules for integration tests. This commit temporarily pins cri to a compatible version.


donoghuc:
Note that `bolt puppetfile install` appears to be unaffected by this. This change is only necessary for CI testing.

donoghuc:
oh fun. looks like a new rubocop today too.

donoghuc:
r10k work ticketed with https://tickets.puppetlabs.com/browse/RK-342

PR #982, opened 2019-05-07
(BOLT-1265) Add terraform target-lookup plugin

nicklewis:
This adds a plugin that can load terraform state and map resource
properties to target parameters. This allows a terraform project to be
used to dynamically determine the targets to use when running Bolt.

PR #1012, opened 2019-05-30
(BOLT-1341) Support Terraform statefile version 4

nicklewis:
Terraform 0.12 introduced version 4 of the statefile format, which is
significantly different from the previous version. In particular, resources are
organized differently, named differently, and their parameters are broken into
nested hashes and arrays instead of alays being a flat hash.

This adds support for the new statefile format by checking the "version" field
and determining how to process the resources. We still accept exactly the same
options in the plugin and the user doesn't need to be aware of which format
their statefile is in.

We do some standardization of the resources into a list, but keep their
parameters in whatever form they came from the statefile. Then at lookup time,
we check for both the dotted form of the key as well as doing the nested
lookup. Only one of those should match at a time, so this prevents us from
having to keep track of which format the data is in.

### DIFF
--- a/lib/bolt/plugin/terraform.rb
+++ b/lib/bolt/plugin/terraform.rb
@@ -24,29 +24,20 @@ module Bolt
       def lookup_targets(opts)
         state = load_statefile(opts)
 
-        resources = state.fetch('modules', {}).flat_map do |mod|
-          mod.fetch('resources', {}).map do |name, resource|
-            [name, resource.dig('primary', 'attributes')]
-          end
-        end
+        resources = extract_resources(state)
 
         regex = Regexp.new(opts['resource_type'])
 
         resources.select do |name, _resource|
           name.match?(regex)
         end.map do |name, resource|
-          unless resource.key?(opts['uri'])
-            warn_missing_property(name, opts['uri'])
-            next
-          end
+          uri = lookup(name, resource, opts['uri'])
+          next unless uri
 
-          target = { 'uri' => resource[opts['uri']] }
+          target = { 'uri' => uri }
           if opts.key?('name')
-            if resource.key?(opts['name'])
-              target['name'] = resource[opts['name']]
-            else
-              warn_missing_property(name, opts['name'])
-            end
+            real_name = lookup(name, resource, opts['name'])
+            target['name'] = real_name if real_name
           end
           if opts.key?('config')
             target['config'] = resolve_config(name, resource, opts['config'])
@@ -65,15 +56,57 @@ module Bolt
         raise Bolt::FileError.new("Could not load Terraform state file #{filename}: #{e}", filename)
       end
 
+      # Format the list of resources into a list of [name, attribute map]
+      # pairs. This method handles both version 4 and earlier statefiles, doing
+      # the appropriate munging based on the shape of the data.
+      def extract_resources(state)
+        if state['version'] >= 4
+          state.fetch('resources', []).flat_map do |resource_set|
+            prefix = "#{resource_set['type']}.#{resource_set['name']}"
+            resource_set['instances'].map do |resource|
+              instance_name = prefix
+              instance_name += ".#{resource['index_key']}" if resource['index_key']
+
+              [instance_name, resource['attributes']]
+            end
+          end
+        else
+          state.fetch('modules', {}).flat_map do |mod|
+            mod.fetch('resources', {}).map do |name, resource|
+              [name, resource.dig('primary', 'attributes')]
+            end
+          end
+        end
+      end
+
+      # Look up a nested value from the resource attributes. The key is of the
+      # form `foo.bar.0.baz`. For terraform statefile version 3, this will
+      # exactly correspond to a key in the resource. In version 4, it will
+      # correspond to a nested hash entry at {foo: {bar: [{baz: <value>}]}}
+      # For simplicity's sake, we just check both.
+      def lookup(name, resource, path)
+        segments = path.split('.').map do |segment|
+          begin
+            Integer(segment)
+          rescue ArgumentError
+            segment
+          end
+        end
+
+        value = resource[path] || resource.dig(*segments)
+        unless value
+          warn_missing_property(name, path)
+        end
+        value
+      end
+
+      # Walk the "template" config mapping provided in the plugin config and
+      # replace all values with the corresponding value from the resource
+      # parameters.
       def resolve_config(name, resource, config_template)
         Bolt::Util.walk_vals(config_template) do |value|
           if value.is_a?(String)
-            if resource.key?(value)
-              resource[value]
-            else
-              warn_missing_property(name, value)
-              nil
-            end
+            lookup(name, resource, value)
           else
             value
           end
--- a/lib/bolt/plugin/terraform.rb
+++ b/lib/bolt/plugin/terraform.rb
@@ -31,10 +31,12 @@ module Bolt
         resources.select do |name, _resource|
           name.match?(regex)
         end.map do |name, resource|
-          uri = lookup(name, resource, opts['uri'])
-          next unless uri
+          target = {}
 
-          target = { 'uri' => uri }
+          if opts.key?('uri')
+            uri = lookup(name, resource, opts['uri'])
+            target['uri'] = uri if uri
+          end
           if opts.key?('name')
             real_name = lookup(name, resource, opts['name'])
             target['name'] = real_name if real_name
--- a/pre-docs/inventory_version2.md
+++ b/pre-docs/inventory_version2.md
@@ -156,10 +156,12 @@ The Terraform plugin accepts several fields:
 
 `dir`: The directory from which to load Terraform state  
 `resource_type`: The Terraform resources to match, as a regular expression  
-`uri`: The property of the Terraform resource to use as the target URI  
+`uri`: The property of the Terraform resource to use as the target URI (optional)  
 `statefile`: The name of the Terraform state file to load within `dir` (optional, defaults to `terraform.tfstate`)  
 `name`: The property of the Terraform resource to use as the target name (optional)  
 `config`: A Bolt config map where each value is the Terraform property to use for that config setting
+
+One of `uri` or `name` is required. If only `uri` is set, then the value of `uri` will be used as the `name`.
 
 ```
 groups:
@@ -175,7 +177,9 @@ groups:
         uri: public_ip
 ```
 
-The resource and property names correspond to the output of `terraform show`. 
+Multiple resources with the same name are identified <resource>.0, <resource>.1, etc.
+
+The path to nested properties must be separated with `.`: for example, `network_interface.0.access_config.0.nat_ip`.
 
 For example, the following truncated output creates two targets, named `34.83.150.52` and `34.83.16.240`. These targets are created by matching the resources `google_compute_instance.web.0` and `google_compute_instance.web.1`. The `uri` for each target is the value of their `network_interface.0.access_config.0.nat_ip` property, which corresponds to the externally routable IP address in Google Cloud.
 
--- a/spec/bolt/plugin/terraform_spec.rb
+++ b/spec/bolt/plugin/terraform_spec.rb
@@ -54,6 +54,14 @@ describe Bolt::Plugin::Terraform do
                                          { 'uri' => ip1, 'name' => 'test-instance-1' })
     end
 
+    it 'sets only name if uri is not specified' do
+      opts.delete('uri')
+      targets = subject.lookup_targets(opts.merge('name' => 'id'))
+
+      expect(targets).to contain_exactly({ 'name' => 'test-instance-0' },
+                                         { 'name' => 'test-instance-1' })
+    end
+
     it 'builds a config map from the inventory' do
       config_template = { 'ssh' => { 'user' => 'metadata.sshUser' } }
       targets = subject.lookup_targets(opts.merge('config' => config_template))
--- a/spec/bolt/plugin/terraform_spec.rb
+++ b/spec/bolt/plugin/terraform_spec.rb
@@ -6,9 +6,8 @@ require 'bolt/plugin/terraform'
 # rubocop:disable Style/BracesAroundHashParameters
 describe Bolt::Plugin::Terraform do
   let(:terraform_dir) { File.expand_path(File.join(__dir__, '../../fixtures/terraform')) }
-  let(:resource_type) { 'google_compute_instance' }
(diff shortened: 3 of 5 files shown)


lib/bolt/plugin/terraform.rb
```diff
@@ -65,15 +56,57 @@ def load_statefile(opts)
...
+          state.fetch('modules', {}).flat_map do |mod|
+            mod.fetch('resources', {}).map do |name, resource|
+              [name, resource.dig('primary', 'attributes')]
+            end
+          end
+        end
+      end
+
+      # Look up a nested value from the resource attributes. The key is of the
+      # form `foo.bar.0.baz`. For terraform statefile version 3, this will
```
donoghuc: This seems reasonable. I'm not sure how many array elements might end up nested under keys (hopefully not so many that it is hard to calculate index by "eye"). It was kind of nice in version 3 that the index were part of the path. We should add a note about this in the docs as my impression is that this is the most noticeable change for how we are resolving values.

lib/bolt/plugin/terraform.rb
```diff
@@ -24,29 +24,20 @@ def warn_missing_property(name, property)
...
 
         resources.select do |name, _resource|
           name.match?(regex)
         end.map do |name, resource|
-          unless resource.key?(opts['uri'])
-            warn_missing_property(name, opts['uri'])
-            next
-          end
+          uri = lookup(name, resource, opts['uri'])
+          next unless uri
```
donoghuc: I dont think this was introduced with this work, but would always setting a `uri` here result in overriding potential config like `host` for ssh config? So specifically If I had 
```yaml
version: 2
groups:
  - name: terraform-nodes
    groups:
      - name: terraform-group-1
        target-lookups:
          - plugin: terraform
            dir: /home/cas/working_dir/BOLT-1265-tf-plugin
            resource_type: aws_instance.xenial.[0-1]
            name: id
            config:
              ssh:
                host: public_ip
```
I would get the following stack trace
```
cas@cas-ThinkPad-T460p:~/working_dir/bolt$ bolt command run whoami -t i-09ebed1560c80c46e
bundler: failed to load command: bolt (/home/cas/working_dir/bolt/.bundle/ruby/2.5.0/bin/bolt)
NoMethodError: undefined method `split' for nil:NilClass
  /home/cas/working_dir/bolt/lib/bolt/plugin/terraform.rb:88:in `lookup'
  /home/cas/working_dir/bolt/lib/bolt/plugin/terraform.rb:34:in `block in lookup_targets'
```
With the following implementation of `lookup_targets` I can set the `name` only and reference it by id
```ruby
      def lookup_targets(opts)
        state = load_statefile(opts)

        resources = extract_resources(state)

        regex = Regexp.new(opts['resource_type'])

        resources.select do |name, _resource|
          name.match?(regex)
        end.map do |name, resource|
          target = if opts.key?('name')
              real_name = lookup(name, resource, opts['name'])
              { 'name' => real_name } if real_name
            else
              uri = lookup(name, resource, opts['uri'])
              { 'uri' => uri } if uri
            end

          next unless target

          if opts.key?('config')
            target['config'] = resolve_config(name, resource, opts['config'])
          end
          target
        end.compact
      end
```
```
cas@cas-ThinkPad-T460p:~/working_dir/bolt$ bolt command run whoami -t i-09ebed1560c80c46e
Started on 34.222.221.208...
Finished on 34.222.221.208:
  STDOUT:
    ubuntu
Successful on 1 node: i-09ebed1560c80c46e
Ran on 1 node in 0.52 seconds
```

donoghuc:
I want to include an example `tfstate` file generated from the HCL manifest I was using before (https://github.com/puppetlabs/bolt/pull/982#issuecomment-492302536) as an example statefule for some EC2 nodes. 
```
{
  "version": 4,
  "terraform_version": "0.12.0",
  "serial": 19,
  "lineage": "5bfd22a0-11e8-5a6d-527e-db91b2616614",
  "outputs": {
    "public_ips": {
      "value": [
        [
          "52.33.75.202",
          "35.163.156.17"
        ]
      ],
      "type": [
        "tuple",
        [
          [
            "tuple",
            [
              "string",
              "string"
            ]
          ]
        ]
      ]
    }
  },
  "resources": [
    {
      "mode": "managed",
      "type": "aws_instance",
      "name": "xenial",
      "each": "list",
      "provider": "provider.aws",
      "instances": [
        {
          "index_key": 0,
          "schema_version": 1,
          "attributes": {
            "ami": "ami-076e276d85f524150",
            "arn": "arn:aws:ec2:us-west-2:130685871274:instance/i-08dce15e22923f5d9",
            "associate_public_ip_address": true,
            "availability_zone": "us-west-2c",
            "cpu_core_count": 1,
            "cpu_threads_per_core": 1,
            "credit_specification": [],
            "disable_api_termination": false,
            "ebs_block_device": [],
            "ebs_optimized": false,
            "ephemeral_block_device": [],
            "get_password_data": false,
            "host_id": null,
            "iam_instance_profile": "",
            "id": "i-08dce15e22923f5d9",
            "instance_initiated_shutdown_behavior": null,
            "instance_state": "running",
            "instance_type": "t1.micro",
            "ipv6_address_count": 0,
            "ipv6_addresses": [],
            "key_name": "aws_key",
            "monitoring": false,
            "network_interface": [],
            "network_interface_id": null,
            "password_data": "",
            "placement_group": "",
            "primary_network_interface_id": "eni-07e822a3f0fe0af8e",
            "private_dns": "ip-172-31-15-143.us-west-2.compute.internal",
            "private_ip": "172.31.15.143",
            "public_dns": "ec2-52-33-75-202.us-west-2.compute.amazonaws.com",
            "public_ip": "52.33.75.202",
            "root_block_device": [
              {
                "delete_on_termination": true,
                "iops": 100,
                "volume_id": "vol-046064ab8523a7cf6",
                "volume_size": 8,
                "volume_type": "gp2"
              }
            ],
            "security_groups": [
              "allow_ssh"
            ],
            "source_dest_check": true,
            "subnet_id": "subnet-9fb3a6c5",
            "tags": null,
            "tenancy": "default",
            "timeouts": null,
            "user_data": null,
            "user_data_base64": null,
            "volume_tags": {},
            "vpc_security_group_ids": [
              "sg-08ad4499b36979194"
            ]
          }
        },
        {
          "index_key": 1,
          "schema_version": 1,
          "attributes": {
            "ami": "ami-076e276d85f524150",
            "arn": "arn:aws:ec2:us-west-2:130685871274:instance/i-0ab9c7f69a3a33ca6",
            "associate_public_ip_address": true,
            "availability_zone": "us-west-2c",
            "cpu_core_count": 1,
            "cpu_threads_per_core": 1,
            "credit_specification": [],
            "disable_api_termination": false,
            "ebs_block_device": [],
            "ebs_optimized": false,
            "ephemeral_block_device": [],
            "get_password_data": false,
            "host_id": null,
            "iam_instance_profile": "",
            "id": "i-0ab9c7f69a3a33ca6",
            "instance_initiated_shutdown_behavior": null,
            "instance_state": "running",
            "instance_type": "t1.micro",
            "ipv6_address_count": 0,
            "ipv6_addresses": [],
            "key_name": "aws_key",
            "monitoring": false,
            "network_interface": [],
            "network_interface_id": null,
            "password_data": "",
            "placement_group": "",
            "primary_network_interface_id": "eni-00449f3c890fa20f9",
            "private_dns": "ip-172-31-9-192.us-west-2.compute.internal",
            "private_ip": "172.31.9.192",
            "public_dns": "ec2-35-163-156-17.us-west-2.compute.amazonaws.com",
            "public_ip": "35.163.156.17",
            "root_block_device": [
              {
                "delete_on_termination": true,
                "iops": 100,
                "volume_id": "vol-0d0658a3d57e430a6",
                "volume_size": 8,
                "volume_type": "gp2"
              }
            ],
            "security_groups": [
              "allow_ssh"
            ],
            "source_dest_check": true,
            "subnet_id": "subnet-9fb3a6c5",
            "tags": null,
            "tenancy": "default",
            "timeouts": null,
            "user_data": null,
            "user_data_base64": null,
            "volume_tags": {},
            "vpc_security_group_ids": [
              "sg-08ad4499b36979194"
            ]
          }
        }
      ]
    },
    {
      "mode": "managed",
      "type": "aws_key_pair",
      "name": "example",
      "provider": "provider.aws",
      "instances": [
        {
          "schema_version": 1,
          "attributes": {
            "fingerprint": "1f:15:89:ad:16:1d:b0:d8:d9:e8:bd:af:e7:ed:60:7e",
            "id": "aws_key",
            "key_name": "aws_key",
            "key_name_prefix": null,
            "public_key": "ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAACAQCdV9Fmfw/eKkkbjasq+L86sxxT4DUlvtIJSswJSXCC079H8mBLs8/Q9yL3gjM/P8XaqC69GrNffItji2WWVurXLcxBOz6TA0eM5114gpHnE/JHRIpSPkie3J7Zv2R+07bhwzeed85SYuPcofWxxS9cf2/k/GixYCN7GV5ySvmuf3xumb3s9WAocOI5BaeJ0SIGdxBcQf33foceD7qkO+PAqixAZE2tOj1TnbzXJayFKQtfn8FfHbiMlOD0VhKEujdSVFMrlzi+rskGprwFzj987hkFjW9Y5Tz6fLBLHa0sKMK8Odd/MF4M4jG41qRX0GJVLLjLFf88eyWPN8HZgcSuj669UBBD5SO4JslNqu9sUmi7Pop50W9Yw6XOpQh6jSaEwoZPZ/lsjVgqFLKMehiyRSm29+Hy8H2e/vSaZeNLziHicVpwU8djeGnJY9pHssxplGA1UwmSP++2gTXZSO8XIOL87bDtB9GuoEX8PfDPQxSEOn75buAJYcmIt+TIh++cH/Cuc+lx6+5l4lUKt6T+xkvoMAmJ9P0Aw+KHp71EG9K3pS33+ATomASbipYKkVCnx0RoD4Y+2IlI8Ny4NKCKzujBJZxGNJvGrIkNF2v+6D5WPhu0ysDSKV3AvdyJ3N00YkZ+FauMB7hbFm9hNhsq0Bb0fj3z1GvOuMyOZcx0hQ== cas@cas-ThinkPad-T460p"
          }
        }
      ]
    },
    {
      "mode": "managed",
      "type": "aws_security_group",
      "name": "allow_ssh",
      "provider": "provider.aws",
      "instances": [
        {
          "schema_version": 1,
          "attributes": {
            "arn": "arn:aws:ec2:us-west-2:130685871274:security-group/sg-08ad4499b36979194",
            "description": "Allow all inbound traffic",
            "egress": [
              {
                "cidr_blocks": [
                  "192.69.65.0/24"
                ],
                "description": "",
                "from_port": 0,
                "ipv6_cidr_blocks": [],
                "prefix_list_ids": [],
                "protocol": "-1",
                "security_groups": [],
                "self": false,
                "to_port": 0
              }
            ],
            "id": "sg-08ad4499b36979194",
            "ingress": [
              {
                "cidr_blocks": [
                  "192.69.65.0/24"
                ],
                "description": "",
                "from_port": 22,
                "ipv6_cidr_blocks": [],
                "prefix_list_ids": [],
                "protocol": "tcp",
                "security_groups": [],
                "self": false,
                "to_port": 22
              }
            ],
            "name": "allow_ssh",
            "name_prefix": null,
            "owner_id": "130685871274",
            "revoke_rules_on_delete": false,
            "tags": null,
            "timeouts": null,
            "vpc_id": "vpc-0357fe7b"
          }
        }
      ]
    }
  ]
}

```

lib/bolt/plugin/terraform.rb
nicklewis: Good call. I will update this to allow just name.

nicklewis:
I updated this to check for name and uri separately, and to allow the calling inventory code to make the determination about whether the generated targets are valid.

This means that targets can be created with just a name or with a name and host/port/etc, rather than with a uri. This change will _also_ cause failures for any terraform resources which match the `resource_type` selector but don't have the fields that are mapped to uri/name, where such resources used to be ignored. Since that doesn't seem like a valid use case anyway, I don't think the change in behavior is significant. If anything, this makes it somewhat _more_ correct.

donoghuc:
>  This change will also cause failures for any terraform resources which match the resource_type selector but don't have the fields that are mapped to uri/name, where such resources used to be ignored.

 

I am seeing a warning here, not a failure. Are we expecting these to fail if they do not match? For example:
1. `name` does not match and `uri` is not set, therefore the expected error with group validation is raised. 
```yaml
groups:
  - name: terraform-nodes
    groups:
      - name: terraform-group-1
        target-lookups:
          - plugin: terraform
            dir: /home/cas/working_dir/BOLT-1265-tf-plugin
            resource_type: aws_instance.xenial.[0-1]
            name: foo
            config:
              ssh:
                host: public_ip
```
```
cas@cas-ThinkPad-T460p:~/working_dir/bolt$ bolt command run whoami -t terraform-nodes
Could not find property foo of terraform resource aws_instance.xenial.0
Could not find property foo of terraform resource aws_instance.xenial.1
No name or uri for target: {"config"=>{"ssh"=>{"host"=>"34.220.193.216"}}, "name"=>nil} for group at ["terraform-group-1"]
```
2. `uri` is set but doesn't map to anything, there is enough information to get by with just the warning. 
```yaml
groups:
  - name: terraform-nodes
    groups:
      - name: terraform-group-1
        target-lookups:
          - plugin: terraform
            dir: /home/cas/working_dir/BOLT-1265-tf-plugin
            resource_type: aws_instance.xenial.[0-1]
            uri: foo
            name: id
            config:
              ssh:
                host: public_ip
```

```
cas@cas-ThinkPad-T460p:~/working_dir/bolt$ bolt command run whoami -t terraform-nodes
Could not find property foo of terraform resource aws_instance.xenial.0
Could not find property foo of terraform resource aws_instance.xenial.1
Started on 34.213.101.90...
Started on 34.220.193.216...
Finished on 34.213.101.90:
  STDOUT:
    ubuntu
Finished on 34.220.193.216:
  STDOUT:
    ubuntu
Successful on 2 nodes: i-0a98454920b302828,i-01f15c738ca27c417
Ran on 2 nodes in 0.53 seconds
```

nicklewis:
That looks like the correct behavior to me. It will be a warning if it can't find the property, and then a failure if that caused the target to be invalid. But if you try to set uri and it can't be found _but_ there's a name, then the target will just have a name and no uri.

nicklewis:
I pushed up a docs change to clarify that `name` _or_ `uri` is required, not both. Also to clarify the way properties are identified.

PR #1013, opened 2019-05-30
(BOLT-1358) Fix local_windows transport calling profile

McdonaldSeanp:
Previous to this commit the local_windows transport would call the powershell
profile if the default value for the 'interpreter' variable was ever used. in
this scenario the default interpreter needs to include PS_ARGS defined in the
powershell transport.

This commit adds PS_ARGS to the default of interpreter

### DIFF
--- a/lib/bolt/transport/local_windows.rb
+++ b/lib/bolt/transport/local_windows.rb
@@ -63,7 +63,8 @@ module Bolt
       private :with_tmpscript
 
       def execute(*command, options)
-        command.unshift(options[:interpreter]) if options[:interpreter]
+        # Interpreter can be an array or string. It will be appended to the command array.
+        command.unshift(options[:interpreter]).flatten! if options[:interpreter]
         command = [options[:env]] + command if options[:env]
 
         if options[:stdin]
@@ -106,7 +107,8 @@ module Bolt
           arguments = unwrap_sensitive_args(arguments)
           if Powershell.powershell_file?(file)
             command = Powershell.run_script(arguments, file)
-            output = execute(command, dir: dir, env: "powershell.exe")
+            interpreter = ['powershell.exe', *Powershell.ps_args]
+            output = execute(command, dir: dir, interpreter: interpreter)
           else
             path, args = *Powershell.process_from_extension(file)
             args += Powershell.escape_arguments(arguments)
@@ -166,7 +168,7 @@ module Bolt
           if Powershell.powershell_file?(script) && stdin.nil?
             command = Powershell.run_ps_task(arguments, script, input_method)
             command = environment_params + Powershell.shell_init + command
-            interpreter ||= 'powershell.exe'
+            interpreter ||= ['powershell.exe', *Powershell.ps_args]
             output =
               if input_method == 'powershell'
                 execute(command, dir: dir, interpreter: interpreter)
--- a/acceptance/config/gem/options.rb
+++ b/acceptance/config/gem/options.rb
@@ -6,6 +6,7 @@
     'setup/gem/pre-suite/020_install.rb',
     'setup/common/pre-suite/030_set_password.rb',
     'setup/common/pre-suite/031_add_local_nix_user.rb',
+    'setup/common/pre-suite/032_configure_windows_profile.rb',
     'setup/common/pre-suite/050_build_bolt_inventory.rb',
     'setup/common/pre-suite/071_install_modules.rb'
   ],
--- a/acceptance/config/git/options.rb
+++ b/acceptance/config/git/options.rb
@@ -7,6 +7,7 @@
     'setup/git/pre-suite/020_install.rb',
     'setup/common/pre-suite/030_set_password.rb',
     'setup/common/pre-suite/031_add_local_nix_user.rb',
+    'setup/common/pre-suite/032_configure_windows_profile.rb',
     'setup/common/pre-suite/050_build_bolt_inventory.rb',
     'setup/common/pre-suite/071_install_modules.rb'
   ],
--- a/acceptance/config/package/options.rb
+++ b/acceptance/config/package/options.rb
@@ -5,6 +5,7 @@
     'setup/package/pre-suite/020_install.rb',
     'setup/common/pre-suite/030_set_password.rb',
     'setup/common/pre-suite/031_add_local_nix_user.rb',
+    'setup/common/pre-suite/032_configure_windows_profile.rb',
     'setup/common/pre-suite/050_build_bolt_inventory.rb'
   ],
   load_path: './lib/acceptance'
--- a/acceptance/lib/acceptance/bolt_setup_helper.rb
+++ b/acceptance/lib/acceptance/bolt_setup_helper.rb
@@ -46,6 +46,10 @@ module Acceptance
       ENV['LOCAL_USER'] || 'local_user'
     end
 
+    def profile_tracker
+      ENV['PROFILE_TRACKER'] || "C:/ProfileDir/profile_tracker.txt"
+    end
+
     def local_user_homedir
       case bolt['platform']
       when /osx/
--- a/acceptance/setup/common/pre-suite/032_configure_windows_profile.rb
+++ b/acceptance/setup/common/pre-suite/032_configure_windows_profile.rb
@@ -1,0 +1,27 @@
+# frozen_string_literal: true
+
+require 'bolt_setup_helper'
+
+test_name "Configure Windows Profile" do
+  extend Acceptance::BoltSetupHelper
+
+  step "Configure a Windows Profile that contains will write to a file every time it is loaded" do
+    if bolt['platform'] =~ /windows/
+      execute_powershell_script_on(bolt, <<-PS)
+$profileTracker = #{profile_tracker.inspect}
+$BaseDir = Split-Path $profileTracker
+if (!(Test-Path -Path $BaseDir ))
+{ New-Item -Type Directory -Path $BaseDir -Force }
+if (!(Test-Path -Path $PROFILE.AllUsersAllHosts))
+{ New-Item -Type File -Path $PROFILE.AllUsersAllHosts -Force }
+$ProfileCode = [string]::Format('"Profile Loaded" | Out-File {0} -Append', $profileTracker)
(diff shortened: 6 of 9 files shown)


lib/bolt/transport/local_windows.rb
```diff
@@ -166,7 +166,7 @@ def run_task(target, task, arguments, options = {})
           if Powershell.powershell_file?(script) && stdin.nil?
             command = Powershell.run_ps_task(arguments, script, input_method)
             command = environment_params + Powershell.shell_init + command
-            interpreter ||= 'powershell.exe'
+            interpreter ||= ['powershell.exe', *Bolt::Transport::Powershell.singleton_class::PS_ARGS]
```
donoghuc: I think we can just make `PS_ARGS` a method in the `Bolt::Transport::Powershell` module. It is meant to make powershell specific code/templates available where needed. I think having to access `PS_ARGS` with `singleton_class` is not ideal. I think we can take this over and assign a ticket to it. Thanks @McdonaldSeanp

adreyer:
@McdonaldSeanp can you file a bolt ticket for this since it will result in a behavioral change.

donoghuc:
https://tickets.puppetlabs.com/browse/BOLT-1358

nicklewis:
This change looks good. Is there a reasonable way to test it? I fear that any way to test this very specific case would wind up being too implementation-specific...

donoghuc:
I guess we could try to add something like the following to CI. This is how I tested locally. 
```
PS C:\Users\Administrator\bolt> echo $profile
C:\Users\Administrator\Documents\WindowsPowerShell\Microsoft.PowerShell_profile.ps1
PS C:\Users\Administrator\bolt> type $profile
Write-Host 'hi from profile'
PS C:\Users\Administrator\bolt> type .\Boltdir\site-modules\test\tasks\init.ps1
Write-Host 'hi'
PS C:\Users\Administrator\bolt> bundle exec bolt task run test -t localhost
Started on localhost...
Finished on localhost:
  hi from profile
  hi
  {
  }
Ran on 1 node in 0.77 seconds
PS C:\Users\Administrator\bolt> git branch
* (HEAD detached at origin/master)
  fix_loading_profile_windows_local
  master
PS C:\Users\Administrator\bolt> git checkout fix_loading_profile_windows_local
Previous HEAD position was 7350519b Merge pull request #1012 from nicklewis/BOLT-1341-support-terraform-version-4-statefiles
Switched to branch 'fix_loading_profile_windows_local'
PS C:\Users\Administrator\bolt> bundle exec bolt task run test -t localhost
Started on localhost...
Finished on localhost:
  hi
  {
  }
Ran on 1 node in 0.58 seconds
```

donoghuc:
@McdonaldSeanp `run script` also appears to load profiles on windows using local transport  to execute powershell scripts. I assume that is not the expected behavior? The template for that is https://github.com/puppetlabs/bolt/blob/7350519b0a28cf184a237c149cbfc5ee7013681c/lib/bolt/transport/powershell.rb#L94-L114 . 

Also, I have been trying to replicate this using the winrm transport. By running something like 
```powershell
if (!(Test-Path -Path $PROFILE.AllUsersCurrentHost))
{ New-Item -Type File -Path $PROFILE.AllUsersCurrentHost -Force }
Set-Content -Path $PROFILE.AllUsersCurrentHost -value "Write-Host 'hi'"
```
I can test for profile being loaded using local transport based on `hi` being printed. For example
```
PS C:\Users\Administrator\bolt\Boltdir> bundle exec bolt script run test.ps1 -t localhost
Started on localhost...
Finished on localhost:
  STDOUT:
    hi
    powershell script
Ran on 1 node in 0.41 seconds
PS C:\Users\Administrator\bolt\Boltdir> type test.ps1
Write-Host 'powershell script'
```
But the profile does not seem to be loaded (or at least I cant detect `hi`) using the winrm transport (with scripts or tasks even when I delete all of `PS_ARGS`). My guess is that the profile I am setting in that script is not picked up when executing remotely. Do you know of a profile that I could set that would be loaded for any powershell script invocation?

McdonaldSeanp:
@donoghuc I don't know enough about the winrm transport in bolt to know how it interacts with powershell and how it would react to "Write-host".

My guess at your repro problem in winrm is the output of "write-host" could disappear regardless of whether or not powershell.exe was called with or without -NoProfile. I know there's alot of powershell logic and pipelining in what actually gets called on the host. "write-host" will _not_ propagate any output down any pipelines, it sends the output directly to the console. Since there's not technically a "console" in winrm maybe the output from write-host in your profile just gets eaten.

I would try and see if using something other than "write-host" in the test profile as a test for if the profile was executed will work?

maybe write to a file with https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.utility/out-file?view=powershell-6 in the test profile and then check if that file exists?

McdonaldSeanp:
Also yes, we do _not_ want to load the profile in `run script`

ci.ps1
```diff
@@ -147,3 +147,23 @@ function Set-ActiveRubyFromPuppet
...
+  Write-Host $ENV:windir
+  $allProfiles = Join-Path -Path $ENV:WINDIR -ChildPath 'System32\WindowsPowerShell\v1.0\profile.ps1'
+  Write-host $allProfiles
+  if (!(Test-Path -Path $allProfiles))
+  {
+    New-Item -Type File -Path $allProfiles -Force
+  }
+  $BaseDir = Split-Path $profileTracker
+  New-Item -ItemType Directory -Path $BaseDir
+  $ProfileCode = [string]::Format('"Profile Loaded" | Out-File {0} -Append', $profileTracker)
```
donoghuc: $Profile evaluates to null when executed from this script. Additionally setting it like this results in 
```
Successfully established WinRM connection with roddypiper after 1 attempt(s)
C:\ProfileDir\profile_tracker.txt
PSVersion                      5.1.14409.1012                                                                                                                                                                                        
PSEdition                      Desktop                                                                                                                                                                                               
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}                                                                                                                                                                               
BuildVersion                   10.0.14409.1012                                                                                                                                                                                       
CLRVersion                     4.0.30319.42000                                                                                                                                                                                       
WSManStackVersion              3.0                                                                                                                                                                                                   
PSRemotingProtocolVersion      2.3                                                                                                                                                                                                   
C:\windows
C:\windows\System32\WindowsPowerShell\v1.0\profile.ps1
LastWriteTime : 6/10/2019 8:10:37 PM
Length        : 0
Name          : profile.ps1
PSPath            : Microsoft.PowerShell.Core\FileSystem::C:\ProfileDir
PSParentPath      : Microsoft.PowerShell.Core\FileSystem::C:\
PSChildName       : ProfileDir
PSDrive           : C
PSProvider        : Microsoft.PowerShell.Core\FileSystem
PSIsContainer     : True
Name              : ProfileDir
FullName          : C:\ProfileDir
Parent            : 
Exists            : True
Root              : C:\
Extension         : 
CreationTime      : 6/10/2019 8:10:37 PM
CreationTimeUtc   : 6/10/2019 8:10:37 PM
LastAccessTime    : 6/10/2019 8:10:37 PM
LastAccessTimeUtc : 6/10/2019 8:10:37 PM
LastWriteTime     : 6/10/2019 8:10:37 PM
LastWriteTimeUtc  : 6/10/2019 8:10:37 PM
Attributes        : Directory
Mode              : d-----
BaseName          : ProfileDir
Target            : {}
LinkType          : 
"Profile Loaded" | Out-File C:\ProfileDir\profile_tracker.txt -Append
bundle exec rake appveyor
..........................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................*........*.........................................................................................................................................................................C:\ProfileDir\profile_tracker.txt
No such file or directory @ rb_sysopen - C:\windows\System32\WindowsPowerShell\v1.0\profile.ps1
.
.......................................................................................................................
Pending: (Failures listed here are expected and do not affect your suite's status)
```
Which makes it seem like the file is not persisted (or at least connot be read in rspec context). Either way it does not seem possible to create a profile that will get loaded with this approach. I think we may been do do this in acceptance tests.

donoghuc:
Got a clean beaker run locally with 
```
cas@cas-ThinkPad-T460p:~/working_dir/bolt/acceptance$ bundle exec beaker-hostgenerator windows10ent-64bolt%2Cwinrm.-debian9-64ssh.-windows2016-64winrm. > hosts.yaml
cas@cas-ThinkPad-T460p:~/working_dir/bolt/acceptance$ bundle exec beaker init -h hosts.yaml -o config/git/options.rb
cas@cas-ThinkPad-T460p:~/working_dir/bolt/acceptance$ bundle exec beaker provision
cas@cas-ThinkPad-T460p:~/working_dir/bolt/acceptance$ GIT_FORK=McdonaldSeanp/bolt.git GIT_BRANCH=fix_loading_profile_windows_local bundle exec beaker exec -t ./tests
```
```
/tests/show/plan_show.rb passed in 35.05 seconds
      Test Suite: tests @ 2019-06-10 17:26:30 -0700

       Total Suite Time: 581.87 seconds
      Average Test Time: 30.62 seconds
              Attempted: 19
                 Passed: 16
                 Failed: 0
                Errored: 0
                Skipped: 3
                Pending: 0
                  Total: 19

      - Specific Test Case Status -
        
Failed Tests Cases:
Errored Tests Cases:
Skipped Tests Cases:
  Test Case ./tests/apply_plan_winrm.rb 
  Test Case ./tests/command_local.rb 
  Test Case ./tests/file_local.rb 
Pending Tests Cases:
```
I gave up on fighting appveyor trying to set a profile that would be loaded. I don't fully understand why `$Profile` evaluates to `null` when executing powershell in appveyor. I started investigating writing profile code to the default location for `Profile.AllUsersAllHosts` but when I ran in to not easily being able to verify that the profile was being loaded I decided to just implement the tests at the acceptance level. 

I spent some time trying to force loading profiles when running scripts or tasks over winrm (for example by removing `-NoProfile` etc from `PS_ARGS`. It kind of makes me think that something in the winrm library is effectively setting `-NoProfile` when using the `:powershell` session shell option (which is good).

PR #1018, opened 2019-06-04
(BOLT-1327) Support `proxy` setting for bolt puppetfile install

nicklewis:
This exposes the r10k proxy setting in a new section of bolt.yaml called
`puppetfile`. This proxy will be used for Git HTTPS and Puppet Forge
operations.

### DIFF
--- a/pre-docs/bolt_configuration_options.md
+++ b/pre-docs/bolt_configuration_options.md
@@ -40,6 +40,8 @@ interpreters:
 
 `inventoryfile`: The path to a structured data inventory file used to refer to groups of nodes on the commandline and from plans. The default path for the inventory file is `inventory.yaml` inside the Bolt project directory.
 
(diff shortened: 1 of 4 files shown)


pre-docs/bolt_configuration_options.md
```diff
@@ -40,6 +40,8 @@ interpreters:
 
 `inventoryfile`: The path to a structured data inventory file used to refer to groups of nodes on the commandline and from plans. The default path for the inventory file is `inventory.yaml` inside the Bolt project directory.
 
+`puppetfile`: A map containing options for the `bolt puppetfile install` command. The only allowed key is `proxy` which specifies the HTTP proxy to use for Git and Puppet Forge operations.
```
adreyer: Do we want to constrain what we accept here or just rely on documentation.

nicklewis: I updated it to only pull out the supported key.

PR #1019, opened 2019-06-05
(website) Add more Puppet-branded styling

nicklewis:
This modifies the styling to fit more closely with Puppet brand guidelines,
and generally makes it more aesthetically pleasing.


nicklewis:
See this PR running at https://nicklewis.github.io/bolt/

@lucywyman I took your PR and mostly just fixed up a couple things around the way the CSS is built.

I'm not sure how I feel about the centered section headers, but aside from that one nitpick this seems like a big improvement.

Sigler:
Really loving how approachable this is. It is shaping up really nicely and looks lovely. I have a deep appreciation for the fact that we're still exploring how Bolt fits into our broader branding, look, and feel.

At the risk of being a bit disruptive, I'd love the opportunity to throw some of the design system (PDS) color scheme at this. A big part of the PDS initiative is to begin reducing the number of distinct designs we have to manage across the org and get them to a common place. This will not only reduce the number of distinct decisions everyone has to make but also provides us a common reference frame from which to consider the look of our open source projects. 

That's a long winded way of saying I'd like to make some text bolder, add some margin between the nav items, and make the links blue. I'm available to provide some design and heck, some CSS assistance.

We all have a ton on our plates so please let me know if this feels heavy handed or misguided. I'm still getting my sea legs and looking for some PDS wins that set us up for stronger cross-functional collaboration.

docs/_layouts/default.html
```diff
@@ -24,12 +24,11 @@
...
         <li>
-          <a href="{{ '/' | relative_url }}">
-            <img src="{{ '/assets/img/bolt-logo.png' | relative_url }}" class="bolt-logo" />
-          </a>
-        </li>
-        <li>
-          <span class="nav-header">Hands-on Lab</span>
+          <span class="nav-header">
+            <a href="{{ '/' | relative_url }}">
+              <img src="{{ '/assets/images/bolt-logo.png' | relative_url }}" class="bolt-logo"> Bolt Hands-on Lab
```
steveax: What does this markup render to? If possible, might want to add an alt attribute (I think this is purely decorative so, `alt=""` will make assistive technology ignore it.)

docs/_sass/modified-cayman.scss
```diff
@@ -0,0 +1,383 @@
...
+  table {
+    display: block;
+    width: 100%;
+    overflow: auto;
+    word-break: normal;
+    word-break: keep-all; // For Firefox to horizontally scroll wider tables.
+    -webkit-overflow-scrolling: touch;
+
+    th {
+      font-weight: bold;
```
steveax: This will I think get you "faux bold" (ie: the browser bolding the 400 weight.) Have a look at [the console's font mixins](https://github.com/puppetlabs/pe-base-styles/#a-note-about-typography)

docs/_layouts/default.html
nicklewis: Added this

PR #1021, opened 2019-06-05
(BOLT-1324) Fix port collision for local test setup

donoghuc:
When the OMI docker container was introduced, it was configured to map local port 2598{5,6} to 598{5,6}. This is the same port used the vagrant windows nano image. While this is not a problem in Travis or Appveyor, it prevents local testing using both the windows nano vagrant image and the docker OMI container. This commit removes the port conflict.


lucywyman:
All the affected rake tasks succeed for me locally with both Windows VM and omi server running! :+1:

PR #1023, opened 2019-06-07
 (BOLT-1379) (docs) 'file upload' can handle directories

caseywilliams:
Update the string docs for the upload_file boltlib function to mention
that the function is able to upload directories as well as files.

### DIFF
--- a/bolt-modules/boltlib/lib/puppet/functions/upload_file.rb
+++ b/bolt-modules/boltlib/lib/puppet/functions/upload_file.rb
@@ -7,9 +7,9 @@ require 'bolt/error'
 #
 # **NOTE:** Not available in apply block
 Puppet::Functions.create_function(:upload_file, Puppet::Functions::InternalFunction) do
-  # Upload a file.
-  # @param source A source path, either an absolute path or a modulename/filename selector for a file in
-  #               <moduleroot>/files.
+  # Upload a file or directory.
+  # @param source A source path, either an absolute path or a modulename/filename selector for a
+  #               file or directory in <moduleroot>/files.
   # @param destination An absolute path on the target(s).
   # @param targets A pattern identifying zero or more targets. See {get_targets} for accepted patterns.
   # @param options Additional options: '_catch_errors', '_run_as'.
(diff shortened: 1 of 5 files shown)


caseywilliams:
Appveyor failure looks unrelated (I hope)?

donoghuc:
We generate https://github.com/puppetlabs/bolt/blob/master/pre-docs/plan_functions.md with a rake task based on puppet-strings documentation in plan function definitions. You can run `bundle exec rake docs` and commit the markdown. 

I was looking for places in the docs where we may clarify this, but I think it is missing. Maybe we could add it to:

https://github.com/puppetlabs/bolt/blob/master/pre-docs/bolt_command_reference.md#bolt-command-reference
https://github.com/puppetlabs/bolt/blob/df2bffe56a3fa3b144bcb121c3fea6750bd55fa2/pre-docs/running_bolt_commands.md#upload-files-to-remote-nodes

adreyer:
unfortunately this needs a ticket or it won't get copy pasted into the real docs

caseywilliams:
Ok, thanks both - I've created a JIRA ticket and regenerated the markdown, and added mentions of directories in several other places.

caseywilliams:
I'm not sure I understand why Travis is failing

donoghuc:
@casey it is unrelated to this PR. We have a regression in the puppet-resource_api gem. We are working on a fix for now. Thanks!

lib/bolt/executor.rb
```diff
@@ -267,12 +267,12 @@ def run_task(targets, task, arguments, options = {})
     end
 
     def upload_file(targets, source, destination, options = {})
-      description = options.fetch('_description', "file upload from #{source} to #{destination}")
+      description = options.fetch('_description', "file or directory upload from #{source} to #{destination}")
```
donoghuc: I think this logging/outputting is intended more to map to the action `file upload` rather than describe to the user the behavior of the action. I would lean toward leaving that part out and instead just focus on documenting what to expect with `file upload`.

caseywilliams: Ok, removed the executor changes

PR #1025, opened 2019-06-10
(BOLT-1342) Remove excluded options and separators from help text

beechtom:
This commit removes the SwitchHider class and filters excluded switches from the OptionParser stack. Previously, switches were extended with the SwitchHider class and their visiblity was toggled depending on whether or not an option was valid for a specified command. This commit removes the SwitchHider class and instead removes a switch entirely from the stack, resulting in the options not being shown in the help text or being considered valid commands.

### DIFF
--- a/lib/bolt/bolt_option_parser.rb
+++ b/lib/bolt/bolt_option_parser.rb
@@ -6,6 +6,76 @@ require 'optparse'
 
 module Bolt
   class BoltOptionParser < OptionParser
+    OPTIONS = { inventory: %w[nodes targets query rerun description],
+                authentication: %w[user password private-key host-key-check ssl ssl-verify],
+                escalation: %w[run-as sudo-password],
+                run_context: %w[concurrency inventoryfile save-rerun],
+                global_config_setters: %w[modulepath boltdir configfile],
+                transports: %w[transport connect-timeout tty],
+                display: %w[format color verbose trace],
+                global: %w[help version debug] }.freeze
+
+    ACTION_OPTS = OPTIONS.values.flatten.freeze
+
+    def get_help_text(subcommand, action = nil)
+      case subcommand
+      when 'apply'
+        { flags: ACTION_OPTS + %w[noop execute compile-concurrency],
+          banner: APPLY_HELP }
+      when 'command'
+        { flags: ACTION_OPTS,
+          banner: COMMAND_HELP }
+      when 'file'
+        { flags: ACTION_OPTS + %w[tmpdir],
+          banner: FILE_HELP }
+      when 'plan'
+        case action
+        when 'convert'
+          { flags: OPTIONS[:global] + OPTIONS[:global_config_setters],
+            banner: PLAN_CONVERT_HELP }
+        when 'show'
+          { flags: OPTIONS[:global] + OPTIONS[:global_config_setters],
+            banner: PLAN_SHOW_HELP }
+        when 'run'
+          { flags: ACTION_OPTS + %w[params compile-concurrency tmpdir],
+            banner: PLAN_RUN_HELP }
+        else
+          { flags: ACTION_OPTS + %w[params compile-concurrency tmpdir],
+            banner: PLAN_HELP }
+        end
+      when 'puppetfile'
+        case action
+        when 'install'
+          { flags: OPTIONS[:global] + OPTIONS[:global_config_setters],
+            banner: PUPPETFILE_INSTALL_HELP }
+        when 'show-modules'
+          { flags: OPTIONS[:global] + OPTIONS[:global_config_setters],
+            banner: PUPPETFILE_SHOWMODULES_HELP }
+        else
+          { flags: OPTIONS[:global] + OPTIONS[:global_config_setters],
+            banner: PUPPETFILE_HELP }
+        end
+      when 'script'
+        { flags: ACTION_OPTS + %w[tmpdir],
+          banner: SCRIPT_HELP }
+      when 'task'
+        case action
+        when 'show'
+          { flags: OPTIONS[:global] + OPTIONS[:global_config_setters],
+            banner: TASK_SHOW_HELP }
+        when 'run'
+          { flags: ACTION_OPTS + %w[params tmpdir],
+            banner: TASK_RUN_HELP }
+        else
+          { flags: ACTION_OPTS + %w[params tmpdir],
+            banner: TASK_HELP }
+        end
+      else
+        { flags: OPTIONS[:global],
+          banner: BANNER }
+      end
+    end
+
     def self.examples(cmd, desc)
       <<-EXAMP
 #{desc} a Windows host via WinRM, providing for the password
@@ -18,7 +88,7 @@ EXAMP
     end
 
     BANNER = <<-HELP
-Usage: bolt <subcommand> <action> [options]
+Usage: bolt <subcommand> <action>
 
 Available subcommands:
   bolt command run <command>       Run a command remotely
@@ -37,16 +107,35 @@ Available subcommands:
 
 Run `bolt <subcommand> --help` to view specific examples.
 
-where [options] are:
+Available options are:
     HELP
 
     TASK_HELP = <<-HELP
-Usage: bolt task <action> <task> [options] [parameters]
+Usage: bolt task <action> <task> [parameters]
 
 Available actions are:
   show                             Show list of available tasks
   show <task>                      Show documentation for task
-  run                              Run a Puppet task
+  run <task>                       Run a Puppet task
+
+Parameters are of the form <parameter>=<value>.
+
+#{examples('task run facts', 'run facter on')}
+Available options are:
+    HELP
+
+    TASK_SHOW_HELP = <<-HELP
+Usage: bolt task show <task>
+
+Available actions are:
+  show                             Show list of available tasks
+  show <task>                      Show documentation for task
+
+Available options are:
+    HELP
+
+    TASK_RUN_HELP = <<-HELP
+Usage: bolt task run <task> [parameters]
 
 Parameters are of the form <parameter>=<value>.
 
@@ -55,7 +144,7 @@ Available options are:
     HELP
 
     COMMAND_HELP = <<-HELP
-Usage: bolt command <action> <command> [options]
+Usage: bolt command <action> <command>
 
 Available actions are:
   run                              Run a command remotely
@@ -65,7 +154,7 @@ Available options are:
     HELP
 
     SCRIPT_HELP = <<-HELP
-Usage: bolt script <action> <script> [[arg1] ... [argN]] [options]
+Usage: bolt script <action> <script> [[arg1] ... [argN]]
 
 Available actions are:
   run                              Upload a local script and run it remotely
@@ -75,7 +164,7 @@ Available options are:
     HELP
 
     PLAN_HELP = <<-HELP
-Usage: bolt plan <action> <plan> [options] [parameters]
+Usage: bolt plan <action> <plan> [parameters]
 
 Available actions are:
   convert <plan_path>              Convert a YAML plan to a Puppet plan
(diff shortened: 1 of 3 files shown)


lib/bolt/bolt_option_parser.rb
```diff
@@ -6,6 +6,16 @@
 
 module Bolt
   class BoltOptionParser < OptionParser
+    EXCLUDES = { 'apply' => %w[params],
```
beechtom: Not sure if this is the best way to exclude options from being shown, but it feels a bit cleaner than separate methods that toggle the hide variable. This approach lets you add an option to a command's array, and it will then be removed from the help text.

lib/bolt/bolt_option_parser.rb
```diff
@@ -136,53 +146,53 @@ def initialize(options)
...
-      end
-      @execute = define('-e', '--execute CODE',
-                        "Puppet manifest code to apply to the targets") do |code|
+      end.extend(SwitchHider)
+      define('-e', '--execute CODE',
+             "Puppet manifest code to apply to the targets") do |code|
         @options[:code] = code
       end.extend(SwitchHider)
 
       separator 'Authentication:'
```
beechtom: One side affect of removing options is it leaves the separators in place. Not a huge issue for most commands, but for one like puppetfile you'll have a bunch of separators in the help text with no options under them.

lib/bolt/bolt_option_parser.rb
```diff
@@ -6,6 +6,16 @@
...
   class BoltOptionParser < OptionParser
+    EXCLUDES = { 'apply' => %w[params],
+                 'command' => %w[execute tmpdir boltdir modulepath compile-concurrency params],
+                 'file' => %w[execute boltdir modulepath compile-concurrency params],
+                 'plan' => %w[execute],
+                 'puppetfile' => %w[nodes targets query rerun user params password private-key host-key-check
+                                    ssl ssl-verify run-as sudo-password concurrency compile-concurrency modulepath
+                                    boltdir inventoryfile save-rerun trannsport connect-timeout tty tmpdir execute],
+                 'script' => %w[execute boltdir modulepath compile-concurrency params],
+                 'task' => %w[execute compile-concurrency] }.freeze
```
beechtom: Not sure if this is the best approach to hide options from the help text, but it feels cleaner than writing separate methods for each command to toggle the hide variable. This lets you just add an option that should be hidden to a command's array.

lib/bolt/bolt_option_parser.rb
beechtom: One side affect of hiding options is that the separators are left behind. This doesn't happen with most of the commands, but for one like 'puppetfile' you'll have several separators with no options listed under them.

Should these be removed or should some logic be added to turn them off when all associated options are hidden?

nicklewis: I think we can get rid of the `SwitchHider` class entirely. It's not really what we want, since it only causes unsupported flags to not be _showed_, but they will still be _accepted_ and then we rely on later validation to ensure we fail if they're not welcome.

Because we now have an explicit filter step, we could potentially just _delete_ the excluded flags rather than extend them all with the `SwitchHider`. In that case, we could keep track of the `top` stack and, if we've deleted every flag up to a specific separator, also delete the separator.

nicklewis:
The subject line of your commit message should be written in imperative form, ie. "Only include valid options for specified command in help text".

The current phrasing "Help text only lists options for specific command" is ambiguous; it could be describing the _problem_ the commit is solving or it could be describing the _new_ behavior.

lib/bolt/bolt_option_parser.rb
```diff
@@ -6,6 +6,16 @@
 
 module Bolt
   class BoltOptionParser < OptionParser
+    EXCLUDES = { 'apply' => %w[params],
+                 'command' => %w[execute tmpdir boltdir modulepath compile-concurrency params],
+                 'file' => %w[execute boltdir modulepath compile-concurrency params],
+                 'plan' => %w[execute],
+                 'puppetfile' => %w[nodes targets query rerun user params password private-key host-key-check
+                                    ssl ssl-verify run-as sudo-password concurrency compile-concurrency modulepath
+                                    boltdir inventoryfile save-rerun trannsport connect-timeout tty tmpdir execute],
```
donoghuc: Typo in `trannsport` results in transport options shown in help.

beechtom:
Added tests to cli_spec for some missing help commands. Are there any other tests we'd like to see?

lib/bolt/bolt_option_parser.rb
```diff
@@ -298,15 +296,24 @@ def initialize(options)
...
 
-    def hide_target_opts(toggle = true)
-      @nodes.hide = @query.hide = @rerun.hide = @targets.hide = toggle
+    def remove_excluded_opts
+      # Remove any options that are not available for the specified subcommand
+      top.list.delete_if do |opt|
+        opt.respond_to?(:switch_name) && EXCLUDES.key?(@options[:subcommand]) &&
+          EXCLUDES[@options[:subcommand]].include?(opt.switch_name)
+      end
+      # Remove any separators if all options of that type have been removed
```
donoghuc: I think you could avoid iterating over all options twice (well I guess 4 times because currently  `remove_exclude_opts` is called twice see above comment) with something like
```ruby
      top.list.reject!.with_index do |opt, index| 
        (opt.respond_to?(:switch_name) && EXCLUDES.key?(@options[:subcommand]) &&
        EXCLUDES[@options[:subcommand]].include?(opt.switch_name)) || 
        (opt.is_a?(String) && top.list[index + 1].is_a?(String))
      end
```
Obviously that is super messy and we cram quite a bit of logic in there, but you could explore something like that.

lib/bolt/bolt_option_parser.rb
```diff
@@ -298,15 +296,24 @@ def initialize(options)
       update
```
donoghuc: Is this necessary in both the initialize method and at https://github.com/puppetlabs/bolt/blob/dfe4f2916b7f0952aee6bfb2c9750ed8787a2d4e/lib/bolt/cli.rb#L64 ?

lib/bolt/bolt_option_parser.rb
beechtom: I tried something like that previously but it leaves behind separators unfortunately. Since the switches are left in place until the iteration is complete, there's never an instance where the separators are next to each other in `top.list`. If another method is used that removes switches as we go, then that causes some switches to be skipped since they're shifted during the iteration.

lib/bolt/bolt_option_parser.rb
beechtom: Removing either instance of `update` results in some failed tests. `parser.update` causes several to fail while the one in the initialize method only causes `./spec/bolt/cli_spec.rb:101` to fail.

lib/bolt/bolt_option_parser.rb
donoghuc: Oh, yeah I guess modifying lists as you are iterating on them is bad anyway. That was horrible advice. My bad.

lib/bolt/bolt_option_parser.rb
```diff
@@ -6,6 +6,31 @@
 
 module Bolt
   class BoltOptionParser < OptionParser
+    INCLUDE_OPTS = { 'apply' => %w[nodes targets query rerun noop description execute user password private-key
```
lucywyman: I think it'd be useful to bundle these into the categories they already have in the help text ('authentication', 'run_context', etc.) and then maybe have a 'common_opts' or something the bundles those together as well, since most of these share most of their options.

lib/bolt/bolt_option_parser.rb
```diff
@@ -298,15 +311,22 @@ def initialize(options)
       update
     end
 
-    def hide_target_opts(toggle = true)
-      @nodes.hide = @query.hide = @rerun.hide = @targets.hide = toggle
+    def remove_excluded_opts
```
lucywyman: Is this `include_opts` now? Despite it's verbosity I stand by my preference for the `include` over `exclude`

lib/bolt/bolt_option_parser.rb
adreyer: agreed. I think this can be fine grained based on specific characteristics or more focussed on what we need right now. 

I think there is a large group of options that are used by all "targetable actions"
targetting options: nodes/target/query/rerun
transport options: user/password/private-key/transport/...
inventoryfile/save-rerun

spec/bolt/cli_spec.rb
```diff
@@ -161,6 +161,24 @@ def stub_config(file_content = {})
...
           }.to output(/Available actions are:.*upload/m).to_stdout
         end
+
+        it 'accepts puppetfile' do
+          cli = Bolt::CLI.new(%w[help puppetfile])
+          expect {
+            expect {
+              cli.parse
+            }.to raise_error(Bolt::CLIExit)
+          }.to output(/Available actions are:.*install.*show-modules/m).to_stdout
```
lucywyman: This should also have a check that some options aren't included in the output. `--private-key` was the option that raised the ticket, so that might be a good one to make sure isn't in the output.

```
}.to output(/Available actions are:.*install.*show-modules/m).to_stdout
.and_not_to include(/--private-key/)
```

lib/bolt/bolt_option_parser.rb
```diff
@@ -6,6 +6,21 @@
 
 module Bolt
   class BoltOptionParser < OptionParser
+    INCLUDE_OPTS = { 'apply' => [:common_opts, 'noop', 'execute', 'compile-concurrency'],
```
adreyer: We should just build these lists up by combining the lists defined in OPTIONS rather then using magic keyword symbols that get looked up later.

ie define `OPTIONS` then look up sets of options from it to build the options for each command. `'apply' => OPTIONS[:run_opts] + ['noop', 'execute', 'compile-concurrency']`

It may be easier to do this if you build this list in a method rather then with constants.

calling the `common_opts` run_opts or `action_opts` is probably clearer.

lib/bolt/bolt_option_parser.rb
```diff
@@ -6,6 +6,21 @@
...
+                     'command' => [:common_opts],
+                     'file' => [:common_opts],
+                     'plan' => [:common_opts, 'params', 'compile-concurrency'],
+                     'puppetfile' => [:display, 'configfile'],
```
adreyer: `puppetfile` should support `modulepath`  probably.

lucywyman:
Note to self: we shouldn't include `[options]` in the subcommand help banner

lib/bolt/bolt_option_parser.rb
```diff
@@ -46,7 +117,26 @@ def self.examples(cmd, desc)
...
+  run <task>                       Run a Puppet task
+
+Parameters are of the form <parameter>=<value>.
+
+#{examples('task run facts', 'run facter on')}
+Available options are:
+    HELP
+
+    TASK_SHOW_HELP = <<-HELP
+Usage: bolt task show <task> [options]
```
lucywyman: I think we should remove `[options]` from all the help texts

lib/bolt/cli.rb
```diff
@@ -51,7 +51,7 @@ def inventory
     end
     private :inventory
 
-    def help?(parser, remaining)
+    def help?(_parser, remaining)
```
lucywyman: You can remove the `_parser` option since it's not used in this function anymore

lib/bolt/cli.rb
```diff
@@ -73,16 +77,18 @@ def parse
...
         raise Bolt::CLIExit
       end
 
-      # This section handles parsing non-flag options which are
-      # subcommand specific rather then part of the config
-      actions = COMMANDS[options[:subcommand]]
-      if actions && !actions.empty?
-        options[:action] = remaining.shift
-      end
+      # # This section handles parsing non-flag options which are
```
lucywyman: This can get deleted

spec/bolt/cli_spec.rb
```diff
@@ -161,6 +161,33 @@ def stub_config(file_content = {})
             }.to raise_error(Bolt::CLIExit)
           }.to output(/Available actions are:.*upload/m).to_stdout
         end
+
```
lucywyman: We should also add a test that `bolt help subcommand action` help text doesn't include invalid options or usage text. I think `plan show` and `plan run` would be good, to ensure that options and help text are present when expected and not present when not expected.

lib/bolt/bolt_option_parser.rb
```diff
@@ -6,6 +6,77 @@
...
+            banner: TASK_SHOW_HELP }
+        when 'run'
+          { flags: ACTION_OPTS + ['params'],
+            banner: TASK_RUN_HELP }
+        else
+          { flags: ACTION_OPTS + ['params'],
+            banner: TASK_HELP }
+        end
+      else
+        { flags: ACTION_OPTS + ['noop', 'execute', 'compile-concurrency', 'params'],
```
lucywyman: Agree this should include just global flags.

lucywyman:
I think this just needs tests for the `bolt help subcommand action` case, and it looks like one stray `[options]` on line 239: https://github.com/beechtom/bolt/blob/parser/lib/bolt/bolt_option_parser.rb#L239 - then this looks good to me!

lib/bolt/bolt_option_parser.rb
```diff
@@ -6,6 +6,76 @@
...
+
+    ACTION_OPTS = OPTIONS.values.flatten.freeze
+
+    def get_help_text(subcommand, action = nil)
+      case subcommand
+      when 'apply'
+        { flags: ACTION_OPTS + ['noop', 'execute', 'compile-concurrency'],
+          banner: APPLY_HELP }
+      when 'command'
+        { flags: ACTION_OPTS,
```
donoghuc: I dont think run command uses a `tmpdir`. Not a huge deal, i think it will just be ignored. Other than that looks awesome!

PR #1026, opened 2019-06-10
(BOLT-1316) Add catch_errors plan language function

lucywyman:
This adds a new Bolt plan language function `catch_errors`, which accepts an optional Array of error kinds to catch and a block of Bolt plan code to catch errors for. If an error matching the kinds (or all kinds, if none are specified) is raised in the block it is returned from the block, otherwise the result of the block is returned.

### DIFF
--- a/bolt-modules/boltlib/lib/puppet/functions/catch_errors.rb
+++ b/bolt-modules/boltlib/lib/puppet/functions/catch_errors.rb
@@ -1,0 +1,57 @@
+# frozen_string_literal: true
+
+# Catches errors in a given block and returns them. This will return the
+# output of the block if no errors are raised. Accepts an optional list of
+# error kinds to catch.
+#
+# **NOTE:** Not available in apply block
+Puppet::Functions.create_function(:catch_errors) do
+  # @param error_types An array of error types to catch
+  # @param block The block of steps to catch errors on
+  # @return Undef If an error is raised in the block then the error will be
+  # returned, otherwise the result will be returned
+  # @example Catch errors for a block
+  #   catch_errors() || {
+  #     run_command("whoami", $nodes)
+  #     run_command("adduser ryan", $nodes)
+  #   }
+  # @example Catch parse errors for a block of code
+  #   catch_errors(['bolt/parse-error']) || {
+  #    run_plan('canary', $nodes)
+  #    run_plan('other_plan)
+  #    apply($nodes) || {
+  #      notify { "Hello": }
+  #    }
+  #   }
+  dispatch :catch_errors do
+    optional_param 'Array[String[1]]', :error_types
+    block_param 'Callable[0, 0]', :block
+  end
+
+  def catch_errors(error_types = nil)
+    unless Puppet[:tasks]
+      raise Puppet::ParseErrorWithIssue
+        .from_issue_and_stack(Bolt::PAL::Issues::PLAN_OPERATION_NOT_SUPPORTED_WHEN_COMPILING,
+                              action: self.class.name)
+    end
+
+    executor = Puppet.lookup(:bolt_executor)
+    executor.report_function_call(self.class.name)
+
+    begin
+      yield
+    rescue Puppet::PreformattedError => e
+      if e.cause.is_a?(Bolt::Error)
+        if error_types.nil?
+          return e.cause.to_puppet_error
+        elsif error_types.include?(e.cause.to_h['kind'])
+          return e.cause.to_puppet_error
+        else
+          raise e
+        end
+      else
+        raise e
+      end
+    end
+  end
+end
--- a/spec/fixtures/modules/catch_errors/plans/break.pp
+++ b/spec/fixtures/modules/catch_errors/plans/break.pp
@@ -1,0 +1,13 @@
+plan catch_errors::break(
+  TargetSpec $nodes,
+  Array $list
+) {
+  $list.each |$elem| {
+    catch_errors() || {
+      notice($elem)
+      break()
+      notice("Out of bounds")
+    }
+  }
+  return "Break the chain"
+}
--- a/spec/fixtures/modules/error/tasks/typed.sh
+++ b/spec/fixtures/modules/error/tasks/typed.sh
@@ -1,0 +1,7 @@
+#!/usr/bin/env bash
+
+if [ "$PT_fail" = true ]; then
+  exit 1
+else
+  echo "{\"tag\": \"you're it\"}"
+fi
--- a/spec/integration/catch_errors_spec.rb
+++ b/spec/integration/catch_errors_spec.rb
@@ -1,0 +1,122 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+require 'bolt_spec/config'
+require 'bolt_spec/conn'
+require 'bolt_spec/files'
+require 'bolt_spec/integration'
+require 'bolt/util'
+
+describe "catch_errors" do
+  include BoltSpec::Integration
+  include BoltSpec::Config
+  include BoltSpec::Conn
+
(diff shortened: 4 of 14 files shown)


bolt-modules/boltlib/lib/puppet/functions/catch_errors.rb
```diff
@@ -0,0 +1,50 @@
...
+    # If it's a task that failed, that should be returned
+    # If it's a plan, plan failure, etc.
+    begin
+      yield
+    rescue => e
+      if e.is_a?(Puppet::PreformattedError)
+        # Produces a ResultSet of errors
+        errs = e.cause.to_h['details']['result_set'].error_set
+        unless error_types.nil?
+          errs.filter_set { |r| error_types.include?(r.value['_error']['kind']) }
```
nicklewis: This builds a new ResultSet which is then being thrown away.

It's also unclear what happens if no errors match the specified types. It seems like if there are any errors that don't match `error_types`, this should raise the error rather than filtering _out_ those error types.

bolt-modules/boltlib/lib/puppet/functions/catch_errors.rb
```diff
@@ -0,0 +1,50 @@
...
+      .from_issue_and_stack(Bolt::PAL::Issues::PLAN_OPERATION_NOT_SUPPORTED_WHEN_COMPILING,
+                            action: 'catch_errors')
+    end
+
+    # If it's a task that failed, that should be returned
+    # If it's a plan, plan failure, etc.
+    begin
+      yield
+    rescue => e
+      if e.is_a?(Puppet::PreformattedError)
```
nicklewis: This is a bit different than the logic we use for `_catch_error` when running a plan, and I can't quite wrap my brain around how it ought to work.

What is the current return value (not including this function) of `run_plan('foo', _catch_errors => true)` if `foo` 
runs a task which errors? It seems like that should be the same as the return value of this function.

bolt-modules/boltlib/lib/puppet/functions/catch_errors.rb
```diff
@@ -0,0 +1,54 @@
...
+      catch(:break) do
+        break
+      end
+      yield
+    rescue Puppet::PreformattedError => e
+      if e.cause.is_a?(Bolt::Error)
+        if error_types.nil?
+          return e.cause.to_puppet_error
+        else
+          if error_types.include?(e.cause.to_h['kind'])
```
nicklewis: If `error_types` defaulted to `[]` instead of nil, this logic could be simplified quite a bit.

donoghuc:
Would we expect an error like the following to be handled by this?
```
plan test() {
  $x = catch_errors(['bolt/pal-error']) || { 
    $y = 0/0
    out::message('will not get here')
  }
  out::message("Should get here")
  notice($x)
}
```
```
cas@cas-ThinkPad-T460p:~/working_dir/bolt$ bolt plan run test
Starting: plan test
Starting: plan test
Finished: plan test in 0.0 sec
Finished: plan test in 0.0 sec
{
  "kind": "bolt/pal-error",
  "msg": "divided by 0",
  "details": {
  }
}
```
Note I get the same result (uncaught error) when leaving the `error_types` array empty.

bolt-modules/boltlib/lib/puppet/functions/catch_errors.rb
```diff
@@ -0,0 +1,52 @@
...
+  dispatch :catch_errors do
+    optional_param 'Optional[Array[String[1]]]', :error_types
+    block_param 'Callable[0, 0]', :block
+  end
+
+  def catch_errors(error_types = nil)
+    unless Puppet[:tasks]
+      raise Puppet::ParseErrorWithIssue
+        .from_issue_and_stack(Bolt::PAL::Issues::PLAN_OPERATION_NOT_SUPPORTED_WHEN_COMPILING,
+                              action: 'catch_errors')
```
donoghuc: Instead of setting `action` explicitly with the string `catch_errors` I think you can use `self.class.name` (could be a good time to go through all the plan functions and do that). I think you could trivially get analytics working for this with 
```
```
Also, along the same lines, it might be nice to add some tests similar to https://github.com/puppetlabs/bolt/blob/master/bolt-modules/boltlib/spec/functions/wait_until_available_spec.rb

spec/fixtures/modules/catch_errors/plans/break.pp
```diff
@@ -0,0 +1,13 @@
...
+  Array $list
+) {
+  $list.each |$elem| {
+    catch_errors() || {
+      notice($elem)
+      break()
+      notice("Out of bounds")
+    }
+  }
+  return run_task('sample::echo', $nodes, message => "Break the chain")
```
donoghuc: Why does this need to run the result of running a task? Could it just return some data? Or even just trigger another `notice`?

bolt-modules/boltlib/lib/puppet/functions/catch_errors.rb
lucywyman: I put up changing the other `action: ` statements in other functions in another PR - that seems like separate maintenance work.

spec/fixtures/modules/error/tasks/typed.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
```
donoghuc: Are you planning on making this a cross platform task, or planning on restricting the integration tests to only ssh targets?

donoghuc: I thin you could get away with having a single ruby task if you target `localhost`.

bolt-modules/boltlib/lib/puppet/functions/catch_errors.rb
```diff
@@ -0,0 +1,55 @@
+# frozen_string_literal: true
+
+# Catches errors in a given block and returns them. This will return the
+# output of the block if no errors are raised. Accepts an optional list of
+# error kinds to catch.
```
donoghuc: This needs a note calling out it is not available in apply block. See https://github.com/puppetlabs/bolt/blob/ea5b61a8daf363d679ed21225564dfb4fab6ad79/bolt-modules/boltlib/lib/puppet/functions/without_default_logging.rb#L9

lucywyman: Fixed

spec/integration/catch_errors_spec.rb
```diff
@@ -0,0 +1,116 @@
...
+
+describe "catch_errors" do
+  include BoltSpec::Integration
+  include BoltSpec::Config
+  include BoltSpec::Conn
+
+  after(:each) { Puppet.settings.send(:clear_everything_for_tests) }
+
+  let(:modulepath) { [fixture_path('modules'), fixture_path('apply')].join(File::PATH_SEPARATOR) }
+  let(:target) do
```
donoghuc: I think that the tests cover the right cases. I get that our current integration test setup can safely assume that the bolt controller being Linux signals that SSH targets are available (If we are calling OMI linux) and that if it is windows winrm ones are. However I think we should get to a place where the bolt controller OS is not used to decide what transport to use to act on targets (with the obvious caveat of local transport). So in this case I think you could lay down some inventory data (which it looks like we might allready have handy https://github.com/puppetlabs/bolt/blob/ea5b61a8daf363d679ed21225564dfb4fab6ad79/spec/lib/bolt_spec/conn.rb#L54-L63)  and run tests against "all", then iterate over the results. This approach would also avoid having both winrm and ssh specific flags in the `config_flags` helper.

donoghuc: Though I guess the ssh nodes would still fail on appveyor and vice versa with travis. So I guess that leaves just using local transport and a ruby task?

PR #1030, opened 2019-06-11
(BOLT-1317) Add rake tasks for building gem for shipping pipeline

donoghuc:
This commit adds the necessary rake tasks, gem dependency and build metadata to build and stage the bolt gem for the release pipeline. This allows the gem to be shipped to public repos with the same job that ships the bolt system packages.


donoghuc:
Corresponding CJC PR https://github.com/puppetlabs/ci-job-configs/pull/5945

PR #1031, opened 2019-06-11
(BOLT-1382) Fix batch err in wait_until_available

reidmv:
Previously, wait_until_available() would only check connectivity for a
batch at a time, and if any target in the batch was unavailable it would
report the whole batch as unavailable.

This commit modifies the error handling such that if a batch comes back
as having any unavailable targets in it, the results are partitioned so
that individual, target-by-target availability reporting is performed
correctly.

### DIFF
--- a/spec/bolt/executor_spec.rb
+++ b/spec/bolt/executor_spec.rb
@@ -310,6 +310,23 @@ describe "Bolt::Executor" do
         expect(result.error_hash['kind']).to eq('puppetlabs.tasks/exception-error')
       end
     end
+
+    context 'with batched execution with more than one target' do
+      let(:pcp) { executor.transport('pcp') }
+      let(:target_1) { Bolt::Target.new('pcp://node1').update_conf(Bolt::Config.default.transport_conf) }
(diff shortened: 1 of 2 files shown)


spec/bolt/executor_spec.rb
```diff
@@ -310,6 +310,25 @@ def mock_node_results(_run_as = nil)
...
+      let(:pcp) { executor.transport('pcp') }
+      let(:target_1) { Bolt::Target.new('pcp://node1').update_conf(Bolt::Config.default.transport_conf) }
+      let(:target_2) { Bolt::Target.new('pcp://node2').update_conf(Bolt::Config.default.transport_conf) }
+      let(:targets) { [target_1, target_2] }
+      
+      it 'partitions failures and successes by batch' do
+        allow(pcp).to receive(:batch_connected?).with(targets).and_return(false)
+        allow(pcp).to receive(:batch_connected?).with([target_1]).and_return(false)
+        allow(pcp).to receive(:batch_connected?).with([target_2]).and_return(true)
+        expect(executor).to receive(:wait_now).and_return(Time.now - 1, Time.now, Time.now + 1)
```
donoghuc: OOps, meant to delete these two expectations. Will do once I verify CI is green.

PR #1034, opened 2019-06-12
Add `submit_command` to the Bolt PuppetDB client

austb:
Still working on updating/adding tests, but I thought I'd get this opened and solicit some feedback.

### DIFF
--- a/lib/bolt/puppetdb/client.rb
+++ b/lib/bolt/puppetdb/client.rb
@@ -3,6 +3,7 @@
 require 'json'
 require 'logging'
 require 'uri'
+require 'time'
 
 module Bolt
   module PuppetDB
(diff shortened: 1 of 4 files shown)


austb:
The `submit_command` function will allow a Bolt plan to submit any one of [PuppetDB's commands ](https://puppet.com/docs/puppetdb/6.3/api/command/v1/commands.html). Currently the only imagined use is a plan that gathers and submits facts for remote servers. But it could theoretically be used to submit catalogs and reports as well.

austb:
Also, from Slack

>  It’s one indirector’s worth less hacky than the official way to gather and push facts to PuppetDB

adreyer:
@austb we consider this ruby client private(like the other stateful objects like inventory and executor). This makes it easier for the Bolt Enterprise team to create their own implementations of these objects as needed.

That means the functions that interact with it probably also need to be part of Bolt's plan language rather than standalone modules. We can discuss this more but it probably makes sense to have a single PR that updates the client and adds the "push facts" function if thats what we decide to do.

adreyer:
This looks generally fine but I think we should wait to merge until we have a full feature to release.

PR #1035, opened 2019-06-12
(BOLT-1356) Add docstrings for canary plan

adreyer:
This adds docstrings for the canary plan to help with strings
development. Currenlty it appears as those strings only handles the
@param statements

### DIFF
--- a/modules/canary/plans/init.pp
+++ b/modules/canary/plans/init.pp
@@ -1,3 +1,30 @@
+# @summary
+#   Run a task, command or script on canary nodes before running it on all nodes.
+#
+# This plan accepts a action and a $nodes parameter. The action can be the name
+# of a task, a script or a command to run. It will run the action on a canary
+# group of nodes and only continue to the rest of the nodes if it succeeds on
+# all canaries. This returns a ResultSet object with a Result for every node.
+# Any skipped nodes will have a 'canary/skipped-node' error kind.
+#
+# @param task
+#  The name of the task to run. Mutually exclusive with command and script.
+# @param command
+#   The command to run. Mutually exclusive with task and script.
+# @param script
+#   The script to run. Mutually exclusive with task and command.
+# @param nodes
+#   The target to run on.
+# @param params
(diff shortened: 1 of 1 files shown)


lucywyman:
Is there an example of how these might be displayed? Or is this *just* for forge work?

adreyer:
You can run strings in the canary module to see the pages it generates or run it with `--format json` to see raw data.

modules/canary/plans/init.pp
```diff
@@ -1,3 +1,30 @@
+# @summary
+#   Run a task, command or script on canary nodes before running it on all nodes.
+#
+# This plan accepts a task/command/script and a $nodes parameter. It will run
+# the task on a canary group of of nodes and only continue to the rest of the
```
lucywyman:
```suggestion
# the task on a canary group of nodes and only continue to the rest of the
```

modules/canary/plans/init.pp
```diff
@@ -1,3 +1,30 @@
+# @summary
+#   Run a task, command or script on canary nodes before running it on all nodes.
+#
+# This plan accepts a task/command/script and a $nodes parameter. It will run
```
lucywyman:
```suggestion
# This plan accepts a task/command/script to run, and a node list as parameters. It will run
```

modules/canary/plans/init.pp
```diff
@@ -1,3 +1,30 @@
...
+# @param task
+#  The name of the task to run. Mutually exclusive with command and script.
+# @param command
+#   The command to run. Mutually exclusive with task and script.
+# @param script
+#   The script to run. Mutually exclusive with task and command.
+# @param nodes
+#   The target to run on.
+# @param params
+#   The parameters to use for the task.
```
lucywyman:
```suggestion
#   The parameters to use for the action.
```

adreyer: Users should not generally set parameters for non-task actions.

PR #1036, opened 2019-06-14
(maint) Fix typos in puppet strings docs for catch_errors

lucywyman:
A newline in the `@return` docstring cut off the string when compiled in CI, and the return type for `catch_errors` was not specified. This also rewords the `Catching Errors` section of the docs for clarity.

### DIFF
--- a/bolt-modules/boltlib/lib/puppet/functions/catch_errors.rb
+++ b/bolt-modules/boltlib/lib/puppet/functions/catch_errors.rb
@@ -8,8 +8,8 @@
 Puppet::Functions.create_function(:catch_errors) do
   # @param error_types An array of error types to catch
   # @param block The block of steps to catch errors on
-  # @return Undef If an error is raised in the block then the error will be
-  # returned, otherwise the result will be returned
+  # @return If an error is raised in the block then the error will be returned,
+  #         otherwise the result will be returned
   # @example Catch errors for a block
   #   catch_errors() || {
   #     run_command("whoami", $nodes)
@@ -26,6 +26,7 @@ Puppet::Functions.create_function(:catch_errors) do
   dispatch :catch_errors do
     optional_param 'Array[String[1]]', :error_types
     block_param 'Callable[0, 0]', :block
+    return_type 'Any'
   end
 
   def catch_errors(error_types = nil)
--- a/pre-docs/writing_plans.md
+++ b/pre-docs/writing_plans.md
@@ -189,7 +189,7 @@ fail_plan($errorobject)
 
 ### Catching Errors in a Plan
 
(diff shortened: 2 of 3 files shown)


melamos:
Thanks for the quick fix. This is added to docs for 1.23.0.

pre-docs/writing_plans.md
```diff
@@ -189,7 +189,7 @@ fail_plan($errorobject)
 
 ### Catching Errors in a Plan
 
-Bolt includes a `catch_errors` function which executes a block of code and returns the error if an error is raised or the result of the block if no errors are raised. When you call `run_plan` with `_catch_errors`, use a `catch_errors` block, or call the `error` method on a result, you may get an error.
+Bolt includes a `catch_errors` function which executes a block of code and returns the error if an error is raised or the result of the block if no errors are raised. You may get an `Error` object returned if you call a function with `_catch_errors`, use a `catch_errors` block, or call the `Error` function.
```
donoghuc: Not quite sure what "use a `cat_errors` block, or call the `Error` function", previously the instruction was "when you do x, use catch error", not it read "here is some info, use catch error".

bolt-modules/boltlib/lib/puppet/functions/catch_errors.rb
```diff
@@ -8,8 +8,7 @@
 Puppet::Functions.create_function(:catch_errors) do
   # @param error_types An array of error types to catch
   # @param block The block of steps to catch errors on
-  # @return Undef If an error is raised in the block then the error will be
-  # returned, otherwise the result will be returned
+  # @return if an error is raised in the block then the error will be returned, otherwise the result will be returned
```
donoghuc: Thats good to know. Looks like if we wanted/needed longer than 120 line length we just need some special indentation like https://github.com/puppetlabs/bolt/blob/de5fbd26dcf0be04e8e6c31e71f4201bb8027cf8/bolt-modules/boltlib/lib/puppet/functions/run_script.rb#L9-L10

donoghuc: capitalize `if` ?

lucywyman: Huh - I thought the other function I looked at had it lower cased but turns out I just looked wrong :woman_shrugging:

pre-docs/writing_plans.md
lucywyman: Well, the other 2 concepts are introduced earlier in the doc. So this introduces the 3rd concept, then says "If you use any of these concepts you'll get an Error object. Here's what an Error object is"

donoghuc: ohhhh, I see, you are enumerating the cases you would expect an error object to be returned. That makes sense.

PR #1038, opened 2019-06-14
(BOLT-1361) Improve plans documentation regarding parameter passing

donoghuc:
Previously the example provided for passing parameters to tasks from a plan was confusing as it made it appear that a Target object could be passed to a task via standard parameter passing. This commit updates the example to illustrate how parameters are serialized to JSON when passed to tasks and therefore to not expect objects to be passed directly without serialization.

### DIFF
--- a/pre-docs/writing_plans.md
+++ b/pre-docs/writing_plans.md
@@ -58,26 +58,28 @@ The following example shows node parameters specified as data type `TargetSpec`.
 
 This allows the user to pass, for each parameter, either a simple node name or a URI that describes the protocol to use, the hostname, username and password.
 
-The plan then calls the `run_task` function, specifying which nodes the tasks should be run on.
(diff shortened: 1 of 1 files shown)


lucywyman:
The example looks good. Might want other eyes on the words, but I don't have suggestions for improvements :woman_shrugging:

lucywyman:
Pinging @MFredette  for review

pre-docs/writing_plans.md
```diff
@@ -58,26 +58,28 @@ The following example shows node parameters specified as data type `TargetSpec`.
 
 This allows the user to pass, for each parameter, either a simple node name or a URI that describes the protocol to use, the hostname, username and password.
 
-The plan then calls the `run_task` function, specifying which nodes the tasks should be run on.
+The plan then calls the `run_task` function, specifying which nodes the tasks should be run on. Note how the `Target` names are collected and stored in `$webserver_names` by iterating over the list of `Target` objects returned by `get_targets`. It is important to understand that task parameters are serialized to JSON format, so extracting the names into an Array of Strings ensures that the `webservers` parameter is in a format that can be converted to JSON.
```
melamos:
```suggestion
The plan then calls the `run_task` function, specifying which nodes the tasks should be run on. The `Target` names are collected and stored in `$webserver_names` by iterating over the list of `Target` objects returned by `get_targets`. 

Task parameters are serialized to JSON format, so extracting the names into an array of strings ensures that the `webservers` parameter is in a format that can be converted to JSON.
```

PR #1039, opened 2019-06-14
(BOLT-1197) Add $nodes run_plan() positional arg 2

reidmv:
To provide consistency of special treatment for a plan parameter named
$nodes, allow the run_plan() function to be invoked with $nodes as the
second positional argument, so that it can be used the same way
run_task() is. E.g.

    run_task('util::hello', $targets,
      message => 'hello, beautiful world!',
    )

    run_plan('reboot', $targets,
      message => 'what a great day to reboot!',
    )

This matches the way run_task() works when a nodes/targets argument is
defined.

Previously, a plan needed to be invoked as follows, which resulted in
inconsistency in how to run "actions" against "targets", in tasks vs.
plans.

    # OLD WAY; NO LONGER NECESSARY (but still works)
    run_task('util::hello', $targets,
      message => 'hello, beautiful world!',
    )

    run_plan('reboot',
      nodes   => $targets,
      message => 'what a great day to reboot',
    )

### DIFF
--- a/bolt-modules/boltlib/spec/fixtures/modules/test/plans/run_me_nodes.pp
+++ b/bolt-modules/boltlib/spec/fixtures/modules/test/plans/run_me_nodes.pp
@@ -1,0 +1,6 @@
+plan test::run_me_nodes(
+  BoltLib::TargetSpec $nodes,
+  Optional[Integer]   $x = undef,
+) {
+  return $nodes
+}
--- a/bolt-modules/boltlib/spec/fixtures/modules/test/plans/run_me_nodes.pp
+++ b/bolt-modules/boltlib/spec/fixtures/modules/test/plans/run_me_nodes.pp
@@ -1,6 +1,5 @@
 plan test::run_me_nodes(
   BoltLib::TargetSpec $nodes,
-  Optional[Integer]   $x = undef,
 ) {
   return $nodes
 }
--- a/bolt-modules/boltlib/spec/functions/run_plan_spec.rb
+++ b/bolt-modules/boltlib/spec/functions/run_plan_spec.rb
@@ -48,6 +48,10 @@ describe 'run_plan' do
 
         is_expected.to run.with_params('test::run_me', '_run_as' => 'bar').and_return('worked2')
       end
+
+      it 'run_plan(name, nodes, hash) where nodes is the "nodes" parameter to the plan' do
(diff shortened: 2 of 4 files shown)


bolt-modules/boltlib/spec/functions/run_plan_spec.rb
```diff
@@ -78,6 +82,11 @@
         is_expected.to run.with_params('test::run_me_int', 'x' => 'should not work')
                           .and_raise_error(/expects an Integer value/)
       end
+
+      it 'fails with argument error if given nodes positional argument and nodes named argument' do
```
lucywyman: It might also be worth adding a test that just `nodes =>` also works? I don't see any fixtures that use that...

lucywyman: Also one that a plan that doesn't have nodes fails if nodes are specified (per Alex's comment)

reidmv: Added a test that calling `run_plan(name, targets)` on a plan without a `nodes` argument fails.

Re: testing that `run_plan(name, nodes => targets)` works, does that exercise anything that this existing test doesn't?

```
run_plan(name, hash) where hash is mapping argname to value
```

Easy enough to add if so.

bolt-modules/boltlib/spec/fixtures/modules/test/plans/run_me_nodes.pp
```diff
@@ -0,0 +1,6 @@
+plan test::run_me_nodes(
+  BoltLib::TargetSpec $nodes,
+  Optional[Integer]   $x = undef,
```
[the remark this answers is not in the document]
reidmv: Removed.

reidmv:
Rebased on master. Saw some seemingly unrelated spec failures previously...

bolt-modules/boltlib/spec/functions/run_plan_spec.rb
lucywyman: Ah I don't think so - I naively searched for `nodes =>` in the spec instead of a generic params passing test :) LGTM!

donoghuc:
Thinking that https://github.com/puppetlabs/bolt/pull/1044 will solve that travis failure.

PR #1042, opened 2019-06-18
(BOLT-1403) Separate bolt-server JSON schemas for targets

caseywilliams:
Splits the SSH and WinRM target JSON schemas into their own dedicated
files, so that other bolt-server request schemas besides the existing
task schema can reuse them.

This is intended to support the future addition of new endpoints in bolt-server that allow for running commands and scripts and uploading files, in addition to running tasks.


caseywilliams:
Now that I'm looking further ahead into the "allow for other actions besides run_task" epic, I'm seeing that there's probably a better way to do this -- I think I'll close this PR for now and open a new one with a different approach.

PR #1045, opened 2019-06-19
(BOLT-1270) Add pkcs7 encryption plugin

adreyer:
This adds support for hiera-eyaml compatible encryption in the inventory
file as well as a `bolt secret` command that allows users to interact
with encyrpted values.

### DIFF
--- a/lib/bolt/bolt_option_parser.rb
+++ b/lib/bolt/bolt_option_parser.rb
@@ -58,6 +58,9 @@ module Bolt
       when 'script'
         { flags: ACTION_OPTS + %w[tmpdir],
           banner: SCRIPT_HELP }
+      when 'secret'
+        { flags: OPTIONS[:global] + OPTIONS[:global_config_setters],
+          banner: SECRET_HELP }
       when 'task'
         case action
         when 'show'
@@ -104,6 +107,9 @@ Available subcommands:
   bolt apply <manifest>            Apply Puppet manifest code
   bolt puppetfile install          Install modules from a Puppetfile into a Boltdir
   bolt puppetfile show-modules     List modules available to Bolt
+  bolt secret createkeys           Create new encryption keys
+  bolt secret encrypt <plaintext>  Encrypt a value
+  bolt secret decrypt <encrypted>  Decrypt a value
 
 Run `bolt <subcommand> --help` to view specific examples.
 
@@ -249,6 +255,16 @@ Usage: bolt apply <manifest.pp>
 
 Available options are:
     HELP
+
+    SECRET_HELP = <<~SECRET_HELP
+      Manage secrets for inventory and hiera data.
+
+      Available actions are:
+        createkeys               Create new encryption keys
+        encrypt                  Encrypt a value
+        decrypt                  Decrypt a value
+      Available options are:
+    SECRET_HELP
 
     def initialize(options)
       super()
--- a/lib/bolt/plugin/pkcs7.rb
+++ b/lib/bolt/plugin/pkcs7.rb
@@ -1,0 +1,100 @@
+# frozen_string_literal: true
+
+require 'bolt/secret/base'
+require 'fileutils'
+
+module Bolt
+  class Plugin
+    class Pkcs7 < Bolt::Secret::Base
+      def self.validate_config(config)
+        known_keys = ['private-key', 'public-key', 'keysize']
+        known_keys.each do |key|
+          unless key.is_a? String
+            raise Bolt::ValidationError, "Invalid config for pkcs7 plugin: '#{key}' is not a String"
+          end
+        end
+
+        config.keys.each do |key|
+          unless known_keys.include?(key)
+            raise Bolt::ValidationError, "Unpexpected key in pkcs7 plugin config: #{key}"
+          end
+        end
+      end
+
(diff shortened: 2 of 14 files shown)


lib/bolt/bolt_option_parser.rb
```diff
@@ -87,6 +90,7 @@ def self.examples(cmd, desc)
 EXAMP
     end
 
+    # TODO: document
```
lucywyman: :x:

lib/bolt/encryptors.rb
```diff
@@ -0,0 +1,44 @@
...
+
+      raw = Base64.decode64(match[:encoded])
+      [raw, match[:plugin]]
+    end
+
+    def execute(action, object, _options)
+      enc = @encryptors['PKCS7']
+      case action
+      # TODO: should this be createkeys for eyaml compatibility
+      when 'create-keys'
```
lucywyman: Unless there's a compelling reason I'm missing I think `createkeys` everywhere makes sense. It makes it easier to search for all the places that's referred to, and seems standard, and we're already using it in most cases of `create*keys`.

adreyer: hmm I thought I fixed this, I'm not sure how the integration tests are passing

adreyer: ah this file is no longer used.

lib/bolt/plugin/pkcs7.rb
```diff
@@ -0,0 +1,76 @@
...
+        cert.version = 2
+        cert.not_before = Time.now
+        cert.not_after = Time.now + 50 * 365 * 24 * 60 * 60
+        cert.public_key = key.public_key
+        cert.sign(key, OpenSSL::Digest.new('SHA512'))
+
+        private_keydir = File.dirname(private_key_path)
+        FileUtils.mkdir_p(private_keydir) unless File.exist?(private_keydir)
+        FileUtils.touch(private_key_path)
+        FileUtils.chmod(0o600, private_key_path)
```
adreyer: This appears to not work on windows. What should we do?

donoghuc: https://github.com/puppetlabs/bolt/blob/ea5b61a8daf363d679ed21225564dfb4fab6ad79/lib/bolt/transport/local_windows.rb#L59 maybe?

pre-docs/inventory_version2.md
```diff
@@ -260,11 +259,67 @@ targets:
...
+
+```
+---
+plugins:
+  pkcs7:
+    private_key: ~/bolt_private_key.pem
+```
+
+`keysize`: They size of the key to generate with `bolt secret createkeys`: default: `2048`
+`private_key`: The path to the private key file. default: `keys/private_key.pkcs7.pem`
```
donoghuc: If you use underscores here to separate `private` and `keys` instead of `private-key`, the config is silently ignored. Should we do some validation of pkcs7 options? Or some logging in createkeys? both? https://github.com/puppetlabs/bolt/pull/1045/files#diff-1c5970042cde05522a17f74b1ee5c988R20

lib/bolt/plugin/pkcs7.rb
```diff
@@ -0,0 +1,76 @@
...
+        OpenSSL::PKCS7.encrypt([public_key], plaintext, cipher, OpenSSL::PKCS7::BINARY).to_der
+      end
+
+      def decrypt_value(ciphertext)
+        pkcs7 = OpenSSL::PKCS7.new(ciphertext)
+        pkcs7.decrypt(private_key, public_key)
+      end
+
+      def secret_createkeys
+        key = OpenSSL::PKey::RSA.new(keysize)
```
donoghuc: Logging might be nice here. Also would users expect keys to be overwritten if they exist?

pre-docs/inventory_version2.md
adreyer: I fixed the docs and added validation

lib/bolt/plugin/pkcs7.rb
adreyer: Should we only log if we're replacing file? Should it refuse to replace files without confirmation or a flag?

pre-docs/inventory_version2.md
```diff
@@ -260,11 +259,67 @@ targets:
...
+Before using the pkcs7 plugin you need to create encryption keys. You can
+create these keys automatically with `bolt secret createkeys` or reuse existing
+hiera-eyaml pkcs7 keys with bolt secret. You can then encrypt values with `bolt
+secret encrypt <plaintext>` command and copy the result into your inventory
+file. If you need to inspect an encrypted value from the inventory you can
+decrypt it with `bolt secret decrypt <encrypted-value>`.
+
+##### Configuration
+
+By default keys are store in the `keys` directory of bolt project repo. If
```
donoghuc:
```suggestion
By default keys are stored in the `keys` directory of bolt project repo. If
```

donoghuc: Also, I know you did not introduce this problem, but the options `keysize`, `private-key` etc appear to be renered on the same line in markdown. Not sure it that matters for easyditta given nobody complained about it for the other plugins.

PR #1046, opened 2019-06-19
(BOLT-1302) Enable apply acceptance tests for fedora 30

donoghuc:
With the release of puppet agent 6.5.0 for fedora 30 the puppet_agent::install task should have an agent to download which will allow apply acceptance tests to work on fedora 30.


lucywyman:
Passes for me!

PR #1047, opened 2019-06-19
Support SSH config's StrictHostKeyChecking option

MikaelSmith:
When `host-key-check` is not configured for a target, check `StrictHostKeyChecking` from SSH config. If it's not set, then use the default of always (strict) checking host keys. If it's set to yes/true, do the same; Bolt's usage model wouldn't work well with prompting for whether to add the key to your known_hosts file. If it's no/false, accept new keys (based on your known_hosts file) to match SSH's behavior.


lucywyman:
@MikaelSmith Would you be up for rebasing this? We prioritized it for this week, and plan to reach out to the net-ssh maintainer about releasing this week (though it seems like we can merge this before that goes out).

MikaelSmith:
I think there's a test case that fails without the updated gem.

MikaelSmith:
Requires a new release of net-ssh.

PR #1049, opened 2019-06-19
(BOLT-1399) Return original parameter hash when erroring

beechtom:
This commit returns the original hash of expected parameters for printing when a script or task stubs raise an error. Previously, the original hash of parameters was split into separate 'arguments' and 'options' hashes which were then incorrectly reassembled into a single hash whenever the stubs errored.

### DIFF
--- a/lib/bolt_spec/plans/action_stubs/script_stub.rb
+++ b/lib/bolt_spec/plans/action_stubs/script_stub.rb
@@ -31,7 +31,7 @@ module BoltSpec
       end
 
       def parameters
-        @invocation[:arguments] + @invocation[:options] if @invocation.include[:arguments]
+        @invocation[:params]
       end
 
       def result_for(target, stdout: '', stderr: '')
@@ -41,6 +41,7 @@ module BoltSpec
       # Public methods
 
       def with_params(params)
+        @invocation[:params] = params
         @invocation[:arguments] = params['arguments']
         @invocation[:options] = params.select { |k, _v| k.start_with?('_') }
         self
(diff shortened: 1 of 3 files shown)


lib/bolt_spec/plans/action_stubs/script_stub.rb
```diff
@@ -31,7 +31,9 @@ def call(targets, script, arguments, options)
       end
 
       def parameters
-        @invocation[:arguments] + @invocation[:options] if @invocation.include[:arguments]
+        # For scripts, arguments are an array while options are a hash
+        # Merging the two requires adding an arguments key to hold the array of arguments for error printing
+        @invocation[:options].merge('arguments' => @invocation[:arguments]) if @invocation.include?(:arguments)
```
donoghuc: Given these seem to only be generated for helpful error messages, maybe we should standardize by either returning a hash or an interpolated string for all the different action stubs.

beechtom: The three other action stubs are already returning hashes, so this change aligns the script stub with the others.

It looks like the `parameters` method is just rebuilding the original hash passed to `with_params`. Would it make more sense to just store this original hash and return it instead of rebuilding it? That's essentially what the command and upload stubs do.

donoghuc: Yeah, If we go the hash route i would suggest having a hash that always has the same two keys. Something like `{options: @invocation[:options], arguments: @invocation[:arguments]}` then the caller would know what to do with that hash.

spec/bolt_spec/plan_spec.rb
```diff
@@ -35,6 +35,11 @@
   it 'fails when not stubbed' do
     expect { run_plan(plan_name, 'nodes' => targets) }.to raise_error(RuntimeError, /Unexpected call to/)
   end
+
+  it 'prints expected parameters when erroring' do
+    expect_action.not_be_called
+    expect { run_plan(plan_name, 'nodes' => targets) }.to raise_error(RuntimeError, /with parameters/)
```
lucywyman: This should also include a test that the parameters are part of the error message. It's hard to do for the reused test cases, but I think putting a specific test to look for the correct parameters  in the `context 'with tasks'` section would be useful.

PR #1050, opened 2019-06-20
(BOLT-1409) Deprecate target-lookups to standardize plugin behavior

adreyer:
This removes the target-lookups section of inventory v2. In place of it
users should put target-lookup entries directly in the targets list.

### DIFF
--- a/lib/bolt/inventory/group2.rb
+++ b/lib/bolt/inventory/group2.rb
@@ -13,7 +13,7 @@ module Bolt
 
       DATA_KEYS = %w[name config facts vars features].freeze
       NODE_KEYS = DATA_KEYS + %w[alias uri]
-      GROUP_KEYS = DATA_KEYS + %w[groups targets target-lookups]
+      GROUP_KEYS = DATA_KEYS + %w[groups targets]
       CONFIG_KEYS = Bolt::TRANSPORTS.keys.map(&:to_s) + ['transport']
 
       def initialize(data, plugins)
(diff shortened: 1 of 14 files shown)


lib/bolt/inventory/group2.rb
```diff
@@ -31,6 +31,12 @@ def initialize(data, plugins)
         raise ValidationError.new("Group name must be a String, not #{@name.inspect}", nil) unless @name.is_a?(String)
         raise ValidationError.new("Invalid group name #{@name}", @name) unless @name =~ NAME_REGEX
 
+        # DEPRECATION : remove this before finalization
+        if data.key?('target-lookups')
+          msg = "'target-lookups' are no longer a separate key. Merge 'target-lookups' and 'targets' lists and replace 'plugin' with '_plugin'." # rubocop:disable Metrics/LineLength
```
donoghuc: nit, omit period at the end of this message to handle case where we have group `@name` otherwise we get:
```
'target-lookups' are no longer a separate key. Merge 'target-lookups' and 'targets' lists and replace 'plugin' with '_plugin'. for group at ["lookup-nodes"]
```

pre-docs/inventory_version2.md
```diff
@@ -127,7 +152,7 @@ fact](https://puppet.com/docs/facter/latest/core_facts.html#identity):
 version: 2
 groups:
   - name: dynamic_config
-    target-lookups:
+    targets:
       - plugin: puppetdb
```
donoghuc:
```suggestion
      - _plugin: puppetdb
```

pre-docs/inventory_version2.md
```diff
@@ -60,28 +60,53 @@ targets:
 
 ## Plugins and Dynamic Inventory
 
-There are two types of  plugins. The target-lookups plugin type allows dynamic target discovery when running Bolt. The config-lookup plugin type allows dynamically looking up values not explicitly stored in an inventoryfile.
+Inventory Plugins can be used to dynamically load information into the inventory file.
+To use a plugin replace a static value in the inventory file with an Object
+containing a `_plugin` key and any plugin specific options that are required.
+The location in which you do this replacement will determine how the plugin
+behaves. Currenlty plugins are only supported for `inventory_targets`, in the
```
donoghuc:
```suggestion
behaves. Currently plugins are only supported for `inventory_targets`, in the
```

pre-docs/inventory_version2.md
donoghuc: Pretty much all the converted examples should be `_plugin`.

PR #1052, opened 2019-06-24
(maint) Pin postgres docker image to 9.6.13

donoghuc:
Changes in 9.6.14 introduce erros in the migration scripts for puppetdb which is blocking CI. Unpin once resolved.

### DIFF
--- a/spec/docker-compose.yml
+++ b/spec/docker-compose.yml
@@ -20,7 +20,7 @@ services:
       - "20024:22"
 
   postgres:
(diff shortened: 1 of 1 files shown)


donoghuc:
In order to track this work there may be a patch introduced in puppetdb with https://github.com/puppetlabs/puppetdb/pull/3010

Or an update to upstream postres addressing https://www.postgresql.org/message-id/15865-17940eacc8f8b081%40postgresql.org

PR #1053, opened 2019-06-24
(BOLT-1403) Use sinatra route params in bolt-server

caseywilliams:
This commit updates the bolt-server routes so that WinRM and SSH
transport behaviors of request responses can be reused for multiple,
future actions in addition to the existing run_task action.

Supporting changes:

- The routes for the SSH and WinRM variants of run_task in the
  bolt-server sinatra app have been updated to use a route parameter,
  :action, to call an action-specific handler method after validating
  the generic transport details.
- The JSON schemas for the bolt-server task action have been refactored
  so that there are two points of validation. First, at the
  /{ssh,winrm}/:action level, a dedicated transport schema is checked
  for either SSH or WinRM. Then, at the action level, a separate schema
  for the action itself (run_task, etc.) is checked.

(This is a second attempt at #1042)


lib/bolt_server/transport_app.rb
```diff
@@ -67,14 +88,18 @@ def validate_schema(schema, body)
...
+
+      content_type :json
       body = JSON.parse(request.body.read)
-      error = validate_schema(@schemas["ssh-run_task"], body)
+
+      error = validate_schema(@schemas["transport-ssh"], body)
       return [400, error.to_json] unless error.nil?
 
-      opts = body['target']
+      opts = body['target'].clone
```
donoghuc: I'm trying to understand why this is different. Were we mutating the body of the request before, and now that is getting a schema validation error?

donoghuc: Specifically, without making a copy we get 
```
{
  "kind": "boltserver/schema-error",
  "msg": "There was an error validating the request body.",
  "details": [
    "The property '#/target' of type object did not match one or more of the required schemas. The schema specific errors were:\n\n- anyOf #0:\n    - The property '#/target' contains additional properties [\"load-config\"] outside of the schema when none are allowed\n- anyOf #1:\n    - The property '#/target' contains additional properties [\"host-key-check\", \"load-config\"] outside of the schema when none are allowed"
  ]
}
```
with a request like
```
=> {"task"=>
  {"metadata"=>{},
   "name"=>"sample::echo",
   "files"=>
    [{"filename"=>"echo.sh",
      "sha256"=>
       "c5abefbdecee006bd65ef6f625e73f0ebdd1ef3f1b8802f22a1b9644a516ce40",
      "size_bytes"=>64,
      "uri"=>
       {"path"=>"/puppet/v3/file_content/tasks/sample/echo.sh",
        "params"=>{"environment"=>"production"}}}]},
 "target"=>
  {"hostname"=>"172.18.0.1",
   "user"=>"bolt",
   "password"=>"bolt",
   "port"=>20022,
   "host-key-check"=>false},
 "parameters"=>{"message"=>"hey"}}
```
It could be that `load-config` needs to be added as an allowable option for an ssh-target. I dont quite get why its complaining about `host-key-check`, but then again those schema error messages are kind of rough.

caseywilliams: Right, we were mutating the body before -- this was after the only schema check there was, so it didn't matter. Now there's a second check after this point, which fails if it has the mutations.

caseywilliams: Oh huh, maybe I'm not familiar enough with what this should look like - if it sounds like it makes sense to add `load-config` to the schema I can definitely do that

donoghuc: Given that these are things that we control after the request is made I'm not sure how valuable adding it to the schema is. If it turns out it makes our lives easier, great, if not we can skip it. I just want to make sure that the `private-key-content` is correctly munged and that we dont load ssh config on host. Will try to test that out.

caseywilliams: Ok, I stepped through this part in the tests with pry and it looks like the body does get mutated when `opts` is changed (unless it's cloned). That said, is `load-config` something we should allow in the schema anyway? Would it make sense for a client to send a true value for this?

donoghuc: I was just testing this out also. Confirmed that using the cloned value results in desired behavior (no ssh config is loaded and private-key-content is munged as expected). 

We should never load config for a target when executed via pe-bolt-server. I'm not really sure what to do about the schema validation here. Honestly those error messages are horrible, but I guess it could be helpful in preventing regressions?

caseywilliams: What if I raise a 400 error preemptively when the request contains a true value for `target['load-config']` instead of letting schema error bubble up? (And if so, are there other body params I should disallow?)

donoghuc: So, with fresh eyes this morning i'm starting to think that omitting the additional options (`protocol`, `load-config`, etc) from the schema makes sense. We do not expect those to be configurable from the request. Instead we set those to known-good values. I think that what you have here works well. If you are happy with it I would say go ahead and merge. Sorry for the distraction.

PR #1054, opened 2019-06-24
(BOLT-1408) Add task plugin

adreyer:
This adds a task plugin that will run an arbitray task on localhost and
use the result. This currenlty supports the inventory_config and
inventory_target hooks.

Since this plugin is only used to bootstrap inventory at this point it
creates a one off Inventory instance to use. This approach may have to
be reconsidered if the task plugin is extended to post-inventory hooks
in the future. For simplicity it creates it's own PAL and Executor
instances too.

### DIFF
--- a/lib/bolt/plugin/task.rb
+++ b/lib/bolt/plugin/task.rb
@@ -1,0 +1,91 @@
+# frozen_string_literal: true
+
+module Bolt
+  class Plugin
+    class Task
+      def hooks
+        %w[inventory_targets inventory_config]
+      end
+
+      def name
+        'task'
+      end
+
+      # This creates it's own PAL so we don't have to pass a promise around
+      #
+      def initialize(config)
+        @config = config
+      end
+
+      attr_reader :config
+
+      def pal
+        # Hiera config should not be used yet.
+        @pal ||= Bolt::PAL.new(config.modulepath, config.hiera_config)
+      end
+
+      def executor
+        # Analytics should be handled at a higher level so create a new executor.
+        @executor ||= Bolt::Executor.new
+      end
+
+      def inventory
+        @inventory ||= Bolt::Inventory.new({}, config)
+      end
+
+      def run_task(opts)
+        result = pal.run_task(opts['task'],
+                              'localhost',
+                              opts['parameters'] || {},
+                              executor,
+                              inventory).first
+
+        raise Bolt::Error.new(result.error_hash['msg'], result.error_hash['kind']) if result.error_hash
+        result
+      end
+
+      def validate_options(opts)
+        raise Bolt::ValidationError, "Task plugin requires that the 'task' is specified" unless opts['task']
+
+        task = pal.task_signature(opts['task'])
+
+        raise Bolt::ValidationError, "Could not find task #{opts['task']}" unless task
+
+        errors = []
+        unless task.runnable_with?(opts['parameters'] || {}) { |msg| errors << msg }
+          # This relies on runnable with printing a partial message before the first real error
+          raise Bolt::ValidationError, "Invalid parameters for #{errors.join("\n")}"
+        end
+      end
+      alias validate_inventory_config validate_options
+
+      def inventory_config(opts)
+        result = run_task(opts)
+
+        unless result.value.include?('config')
+          raise Bolt::ValidationError, "Task result did not return 'config': #{result.value}"
+        end
+
+        result['config']
+      end
+
+      def inventory_targets(opts)
+        raise Bolt::ValidationError, "Task plugin requires that the 'task' is specified" unless opts['task']
+
+        result = run_task(opts)
+
+        targets = result['targets']
+        unless targets.is_a?(Array)
+          raise Bolt::ValidationError, "Task result did not return a targets array: #{result.value}"
+        end
+
+        unless targets.all? { |t| t.is_a?(Hash) }
+          msg = "All targets returned by an inventory targets task must be hashes, got: #{targets}"
+          raise Bolt::ValidationError, msg
+        end
+
+        targets
+      end
+    end
+  end
+end
--- a/pre-docs/inventory_version2.md
+++ b/pre-docs/inventory_version2.md
@@ -103,6 +103,93 @@ The Following plugins an be used for config:
 * `prompt`: Prompt a user for a configuration value
 * `pkcs7`: decrypt a pkcs7 encrypted value from the inventory file.
 
+#### Task
+
+The Task plugin will run a task on `localhost` to lookup configuration
+information or target lists for the inventory.
+
+For both use cases the plugin accepts two keys:
+* `task`: The task to run.
+* `parameters`: The parameters to pass to the task.
+
+For example this will run the `my_json_file::targets` task to look up targets
+and the `my_db::secret_lookup` task to look up the ssh password.
+
+```
+---
+version: 2
+targets:
+  - _plugin: task
+    task: my_json_file::targets
+    parameters:
+      # These parameters are specific to the task
+      file: /etc/targets/data.json
+      environment: production
+      app: my_app
+config:
+  ssh:
+    password:
+      _plugin: task
+      task: my_db::secret_lookup
+      parameters:
+        # These parameters are task specific
+        key: ssh_password
+```
+
+##### Inventory Config Tasks
+
+To look up configuration information for the inventory return a `config` key in
+the task result containing the data that will be used in place where the plugin entry is.
+The value of config can be any type of data that is appropriate for the
+specific location in config. For example `host-key-check` for ssh must be a
+boolean, `password` must be a string and `run-as-command` must be an array of
+strings. This result would be appropriate for an entire `ssh` section of config.
+
+```
+{
+  "config": {
+    "host-key-check": true,
+    "password": "hunter2",
+    "run-as-command": [ "sudo", "-k", "-S", "-E", "-u", "user", "-p", "password"]
+  }
+}
+```
+
+This task looks up a password value from a secret database and returns it.
+
+```python
+#!/usr/bin/env python
+import json, sys
+from my_secret import Client
+
+params = json.load(sys.stdin)
+
+client = Client
+secret = client.get_secret(data['key'])
+# secret can be any value that can be dumped to json.
+json.dump({'config': secret}, sys.stdout)
+```
+
+##### Inventory Target Tasks
+
+To look up a list of targets for inventory return a hash or JSON object that
+includes a targets key with an array value from task. Each item in this list
+should be a Target JSON object matching the Object format you would use in a targets
+section of the inventory file. Bare string targets are not valid.
+
+This task reads in a JSON file and looks up a value.
+
+```python
+#!/usr/bin/env python
+import json, sys
+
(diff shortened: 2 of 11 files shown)


pre-docs/inventory_version2.md
```diff
@@ -180,6 +180,77 @@ groups:
...
+
+```python
+#!/usr/bin/env python
+import json, sys
+
+params = json.load(sys.stdin)
+with open(params['file']) as fh:
+  data = json.load(fh)
+targets = data[params['environment']][params['app']]
+json.dump({'targets': targets}, sys.stdout)
```
lucywyman: I think you have the examples and text switched.

lib/bolt/plugin/task.rb
```diff
@@ -0,0 +1,71 @@
...
+        result
+      end
+
+      def validate_options(opts); end
+
+      def inventory_config(opts)
+        result = run_task(opts)
+
+        unless result.value.include?('config')
+          raise Bolt::Inventory::ValidatonError, "Task result did not return 'config': #{result.value}"
```
lucywyman:
```suggestion
          raise Bolt::Inventory::ValidationError, "Task result did not return 'config': #{result.value}"
```

lib/bolt/plugin/task.rb
```diff
@@ -0,0 +1,71 @@
...
+        end
+
+        result['config']
+      end
+
+      def inventory_targets(opts)
+        result = run_task(opts)
+
+        unless result['targets'].is_a?(Array)
+          raise Bolt::Inventory::ValidatonError, "Task did not return 'targets': #{result.value}"
```
lucywyman:
```suggestion
          raise Bolt::Inventory::ValidationError, "Task did not return 'targets': #{result.value}"
```

lib/bolt/plugin/task.rb
lucywyman: That error class also has a second argument for `offending_group` which should be optional.

lib/bolt/plugin/task.rb
lucywyman: I think we should have separate errors for when `'targets'` isn't in the result object (like this error message), and when the `targets` value is not an array.

pre-docs/inventory_version2.md
```diff
@@ -180,6 +180,77 @@ groups:
             hostname: facts.networking.interfaces.en0.ipaddress
 ```
 
+#### Task
+
+The Task plugin will run a task on `localhost` to lookup configuration
+information or target lists for inventory..
```
```suggestion
information or target lists for inventory.
```

pre-docs/inventory_version2.md
```diff
@@ -180,6 +180,77 @@ groups:
             hostname: facts.networking.interfaces.en0.ipaddress
 ```
 
+#### Task
+
+The Task plugin will run a task on `localhost` to lookup configuration
```
```suggestion
The Task plugin will run a task on `localhost` to lookup connection configuration
```

pre-docs/inventory_version2.md
```diff
@@ -180,6 +180,77 @@ groups:
...
+information or target lists for inventory..
+
+For both use cases the plugin accepts two keys:
+* `task`: The task to run.
+* `parameters`: The parameters to pass to the task.
+
+example:
```
```suggestion
For example, this would equivalent to running `bolt task run my_json_file::targets -n localhost file=/etc/targets/data.json environment=production app=my_app`:
```

pre-docs/inventory_version2.md
```diff
@@ -180,6 +180,77 @@ groups:
...
+      _plugin: task
+      task: my_db::secret_lookup
+      parameters:
+        # These parameters are task specific
+        key: ssh_password
+```
+
+##### Inventory Config Tasks
+
+To lookup configuration information for the inventory return a `config` key in
```
```suggestion
To look up configuration information for the inventory return a `config` key in
```

lucywyman: "To look up connection configuration return a hash or JSON object with a `config` key from the task. The hash should be the only thing output by the task. The value of `config` can be any form of data that is appropriate for the
specific location in config. For example `host-key-check` should be a boolean value, `run-as` should be a string, and `run-as-command` should be an array, so a task that sets these should return:
```
{
"config": {
  "host-key-check": true,
  "run-as": "sudo-user",
  "run-as-command": [ "sudo", "-k", "-S", "-E", "-u", "user", "-p", "password"] 
}
}
```

pre-docs/inventory_version2.md
```diff
@@ -180,6 +180,77 @@ groups:
...
+the task containing the data that should be using in place of the plugin entry.
+The value of config can be any form of data that is appropriate for the
+specific location in config. where it is used.
+
```
lucywyman: For example, this task reads in a JSON file, gets a secret key, and returns <output>.

pre-docs/inventory_version2.md
```diff
@@ -180,6 +180,77 @@ groups:
...
+params = json.load(sys.stdin)
+with open(params['file']) as fh:
+  data = json.load(fh)
+targets = data[params['environment']][params['app']]
+json.dump({'targets': targets}, sys.stdout)
+```
+
+##### Inventory Target Tasks
+
+To lookup a list of targets for inventory return a `targets` list from task.
```
lucywyman: To look up a list of targets for inventory return a hash or JSON object that includes a  `targets` key with an array value from task. Each item in this list should be a `Target` json object matching the format you would use in a `targets` section of the inventory file. Bare string targets are not valid.

spec/integration/plugin/task_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+      plan_dir = File.join(boltdir, 'modules', 'passw', 'plans')
+      FileUtils.mkdir_p(plan_dir)
+      File.write(File.join(plan_dir, 'init.pp'), plan)
+      output = run_cli(['plan', 'run', 'passw', '--boltdir', boltdir])
+
+      expect(output.strip).to eq('"ssshhh"')
+    end
+  end
+
+  it 'supports a target lookup' do
```
lucywyman: This should also test a failure case, ie. returning a list of bare strings.

spec/integration/plugin/task_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+  end
+
+  it 'supports a config lookup' do
+    inventory = { 'version' => 2,
+                  'targets' => [
+                    { 'uri' => 'node1',
+                      'config' => {
+                        'ssh' => {
+                          'user' => 'me',
+                          'password' => {
```
lucywyman: It might also be worth adding a test that a returned array loads correctly (ie for `run-as-command`). Might also be more trouble than it's worth.

lib/bolt/plugin/task.rb
```diff
@@ -0,0 +1,71 @@
...
+                              opts['parameters'] || {},
+                              executor,
+                              inventory).first
+
+        # TODO: what kind of error is this?
+        raise result.error_hash['msg'] if result.error_hash
+        result
+      end
+
+      def validate_options(opts); end
```
lucywyman: Are we supporting this as part of this ticket? I don't see how this can be used yet.

lib/bolt/plugin/task.rb
adreyer: I think making the error message clearer is probably adequate since we print the result as part of this.

lib/bolt/plugin/task.rb
adreyer: This was hard to understand but I think making the error message clearer is probably adequate since we print the result as part of this.

lib/bolt/plugin/task.rb
adreyer: updated

lucywyman:
I think you need a `rake docs` run

lib/bolt/plugin/task.rb
```diff
@@ -0,0 +1,91 @@
...
+
+        raise Bolt::ValidationError, "Could not find task #{opts['task']}" unless task
+
+        errors = []
+        unless task.runnable_with?(opts['parameters'] || {}) { |msg| errors << msg }
+          # This relies on runnable with printing a partial message before the first real error
+          raise Bolt::ValidationError, "Invalid parameters for #{errors.join("\n")}"
+        end
+      end
+      alias validate_inventory_config validate_options
```
lucywyman: Is there a reason not to just change `validate_inventory_config` to `validate_options`? Or some other common name for both validate methods?

pre-docs/inventory_version2.md
```diff
@@ -180,6 +180,93 @@ groups:
             hostname: facts.networking.interfaces.en0.ipaddress
 ```
 
+#### Task
+
+The Task plugin will run a task on `localhost` to lookup configuration
+information or target lists for the inventory..
```
```suggestion
information or target lists for the inventory.
```

pre-docs/inventory_version2.md
```diff
@@ -180,6 +180,93 @@ groups:
...
+
+For both use cases the plugin accepts two keys:
+* `task`: The task to run.
+* `parameters`: The parameters to pass to the task.
+
+For example this will run the `my_json_file::targets` task to lookup targets
```
```suggestion
For example this will run the `my_json_file::targets` task to look up targets
```

pre-docs/inventory_version2.md
```diff
@@ -180,6 +180,93 @@ groups:
...
+and the `my_db::secret_lookup` task to lookup the ssh password.
```
```suggestion
and the `my_db::secret_lookup` task to look up the ssh password.
```

pre-docs/inventory_version2.md
```diff
@@ -180,6 +180,93 @@ groups:
             hostname: facts.networking.interfaces.en0.ipaddress
 ```
 
+#### Task
```
lucywyman: It might be worth making a bit more noise about this, since users can write their own inventory plugins now. Maybe a level-3 heading of "Writing Inventory Plugins as Tasks" or something?

pre-docs/inventory_version2.md
```diff
@@ -180,6 +180,93 @@ groups:
...
+      _plugin: task
+      task: my_db::secret_lookup
+      parameters:
+        # These parameters are task specific
+        key: ssh_password
+```
+
+##### Inventory Config Tasks
+
+To lookup configuration information for the inventory return a `config` key in
```
```suggestion
To look up configuration information for the inventory return a `config` key in
```

pre-docs/inventory_version2.md
```diff
@@ -180,6 +180,93 @@ groups:
...
+the task result containing the data that should be using in place of the plugin entry.
```
```suggestion
the task result containing the data that will be used in place where the plugin entry is.
```

pre-docs/inventory_version2.md
```diff
@@ -180,6 +180,93 @@ groups:
...
+The value of config can be any type of data that is appropriate for the
+specific location in config. For example `host-key-check` for ssh must be a
+boolean, `password` must be a string and `run-as-command must be an array of
```
```suggestion
boolean, `password` must be a string and `run-as-command` must be an array of
```

lib/bolt/plugin/task.rb
adreyer: Yes it's generally useful for validating any likely hook.

pre-docs/inventory_version2.md
adreyer: This is not writing your own plugin. You are always using the `task` plugin. I can move it to the top though?

PR #1055, opened 2019-06-25
(maint) Specify bolt_server spec helper default port as integer

donoghuc:
Previously the target2request method would set the port for an ssh target as the string '22'. The transport app schema requires this value to be an integer. This commit updates the method to set the default ssh port to the integer value.

### DIFF
--- a/spec/lib/bolt_spec/bolt_server.rb
+++ b/spec/lib/bolt_spec/bolt_server.rb
@@ -39,7 +39,7 @@ module BoltSpec
         elsif target.options['private-key']
           req['private-key-content'] = File.read(target.options['private-key'])
         end
-        req['port'] ||= '22'
+        req['port'] ||= 22
         req['user'] ||= 'root'
       when 'winrm'
         req['ssl'] = false


donoghuc:
Noticed then when reviewing https://github.com/puppetlabs/bolt/pull/1053. 
```
cas@cas-ThinkPad-T460p:~/working_dir/bolt$ bundle exec scripts/server_client.rb sample::echo 'ssh://root:****@f32gbe3lv9s3erz.delivery.puppetlabs.net' '{"message": "hey"}'
{
  "kind": "boltserver/schema-error",
  "msg": "There was an error validating the request body.",
  "details": [
    "The property '#/target/port' of type string did not match the following type: integer in schema fd716135-94d5-5356-9213-f5c4225bc92a"
  ]
}
```
```
cas@cas-ThinkPad-T460p:~/working_dir/bolt$ git diff
diff --git a/spec/lib/bolt_spec/bolt_server.rb b/spec/lib/bolt_spec/bolt_server.rb
index 2a0dc255..a1f79d58 100644
--- a/spec/lib/bolt_spec/bolt_server.rb
+++ b/spec/lib/bolt_spec/bolt_server.rb
@@ -39,7 +39,7 @@ module BoltSpec
         elsif target.options['private-key']
           req['private-key-content'] = File.read(target.options['private-key'])
         end
-        req['port'] ||= '22'
+        req['port'] ||= 22
         req['user'] ||= 'root'
       when 'winrm'
         req['ssl'] = false
```
```
cas@cas-ThinkPad-T460p:~/working_dir/bolt$ bundle exec scripts/server_client.rb sample::echo 'ssh://root:****@f32gbe3lv9s3erz.delivery.puppetlabs.net' '{"message": "hey"}'
{
  "node": "f32gbe3lv9s3erz.delivery.puppetlabs.net",
  "target": "f32gbe3lv9s3erz.delivery.puppetlabs.net",
  "action": "task",
  "object": "sample::echo",
  "status": "success",
  "result": {
    "_output": "f32gbe3lv9s3erz got passed the message: hey\n"
  }
}
```

PR #1057, opened 2019-06-27
(BOLT-1318) Warn when case-insensitive modulepath exists

beechtom:
This commit adds a warning that displays when a specified modulepath does not exist but a case-insensitive version does. The case-insensitive version of the modulepath is displayed with the warning. Previously, if the modulepath's case was incorrect then Bolt would not be able to locate the modules.

The reason for this change is that some Bolt users may be expecting case-insensitivity when specifying a modulepath, but not know that Bolt is case-sensitive.

### DIFF
--- a/lib/bolt/config.rb
+++ b/lib/bolt/config.rb
@@ -286,5 +286,27 @@ module Bolt
         impl.validate(@transports[transport])
       end
     end
+
+    # Check if there is a case-insensitive match to the path
+    def check_path_case(type, paths)
+      return if paths.nil?
+      matches = matching_paths(paths)
+
+      if matches.any?
+        msg = "WARNING: Bolt is case sensitive when specifying a #{type}. Did you mean:\n"
+        matches.each { |path| msg += "         #{path}\n" }
+        @logger.warn msg
+      end
+    end
+
+    def matching_paths(paths)
+      [*paths].map { |p| Dir.glob([p, casefold(p)]) }.flatten.uniq.reject { |p| [*paths].include?(p) }
+    end
+
+    def casefold(path)
+      path.chars.map do |l|
+        l =~ /[A-Za-z]/ ? "[#{l.upcase}#{l.downcase}]" : l
+      end.join
+    end
   end
 end
--- a/lib/bolt/bolt_option_parser.rb
+++ b/lib/bolt/bolt_option_parser.rb
@@ -366,7 +366,8 @@ Available options are:
         @options[:'compile-concurrency'] = concurrency
       end
       define('-m', '--modulepath MODULES',
-             "List of directories containing modules, separated by '#{File::PATH_SEPARATOR}'") do |modulepath|
+             "List of directories containing modules, separated by '#{File::PATH_SEPARATOR}'",
+             'Directories are case-sensitive') do |modulepath|
         # When specified from the CLI, modulepath entries are relative to pwd
         @options[:modulepath] = modulepath.split(File::PATH_SEPARATOR).map do |moduledir|
           File.expand_path(moduledir)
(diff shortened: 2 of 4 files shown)


lib/bolt/util.rb
```diff
@@ -202,6 +202,29 @@ def validate_file(type, path, allow_dir = false)
...
+        return if paths.nil?
+        matches = matching_paths(paths)
+        warn_path_case(type, matches) if matches.any?
+      end
+
+      def matching_paths(paths)
+        [*paths].map { |p| Dir.glob([p, casefold(p)]) }.flatten.uniq.reject { |p| [*paths].include?(p) }
+      end
+
+      def warn_path_case(type, paths)
```
lucywyman: I think this can just be inline in `check_path_case`

lib/bolt/util.rb
```diff
@@ -202,6 +202,29 @@ def validate_file(type, path, allow_dir = false)
...
+        msg = "WARNING: Bolt is case sensitive when specifying a #{type}. Did you mean:\n",
+              paths.join("\n"), "\n"
+        puts msg
```
lucywyman: This should be `logger.warn`

lib/bolt/config.rb
```diff
@@ -247,6 +247,8 @@ def modulepath
     end
 
     def validate
+      Bolt::Util.check_path_case('modulepath', @modulepath)
```
lucywyman: Do we want to do this for other configured paths as well, ie. boltdir and inventroyfile?

lib/bolt/util.rb
```diff
@@ -202,6 +202,29 @@ def validate_file(type, path, allow_dir = false)
...
       end
 
+      # Check if there is a case-insensitive match to the path
+      def check_path_case(type, paths)
+        return if paths.nil?
+        matches = matching_paths(paths)
+        warn_path_case(type, matches) if matches.any?
+      end
+
+      def matching_paths(paths)
```
lucywyman: This can also probably just be in check_path_case, unless there's somewhere else you want to use it?

beechtom: I had that originally, but appveyor kept failing the test I wrote when it was originally expecting output, so I changed the test to expect possible paths instead. I'd like to get it into `warn_path_case` if I can, but am not sure how to get the test working with my original test.

lib/bolt/config.rb
beechtom: Oddly enough boltdir and inventoryfile don't _seem_ to be case-sensitive.

lib/bolt/config.rb
```diff
@@ -286,5 +286,27 @@ def validate
...
+        @logger.warn msg
+      end
+    end
+
+    def matching_paths(paths)
+      [*paths].map { |p| Dir.glob([p, casefold(p)]) }.flatten.uniq.reject { |p| [*paths].include?(p) }
+    end
+
+    def casefold(path)
+      path.split(//).map do |l|
```
adreyer: would `path.chars` work here it's a little bit more readable I think.

lib/bolt/config.rb
```diff
@@ -286,5 +286,27 @@ def validate
...
       end
     end
+
+    # Check if there is a case-insensitive match to the path
+    def check_path_case(type, paths)
+      return if paths.nil?
+      matches = matching_paths(paths)
+
+      if matches.any?
+        msg = "WARNING: Bolt is case sensitive when specifying a #{type}. Did you mean:\n"
```
lucywyman: This spaces things a little weirdly. I think the colon is enough separation without the newline and the tab before the path. So just "WARNING: Bolt is case sensitive when specifying a #{type}. Did you mean: #{path}"

PR #1059, opened 2019-06-27
(BOLT-1403) Make bolt-server specs less task-specific

caseywilliams:
Refactors transport_app_spec tests in a way that should allow for easier
additions of other actions (like run_command, upload_file, run_script)
to the bolt server API alongside run_task.

app_integration_spec tests remain the same, except for a renamed helper
method.

### DIFF
--- a/lib/bolt_server/transport_app.rb
+++ b/lib/bolt_server/transport_app.rb
@@ -62,11 +62,7 @@ module BoltServer
 
       task = Bolt::Task::PuppetServer.new(body['task'], @file_cache)
       parameters = body['parameters'] || {}
-      results = @executor.run_task(target, task, parameters)
-
-      # Since this will only be on one node we can just return the first result
-      result = scrub_stack_trace(results.first.status_hash)
-      [200, result.to_json]
+      @executor.run_task(target, task, parameters)
     end
 
     get '/' do
@@ -105,10 +101,13 @@ module BoltServer
         opts.delete('private-key-content')
       end
       opts['load-config'] = false
-
       target = [Bolt::Target.new(body['target']['hostname'], opts)]
 
-      method(params[:action]).call(target, body)
+      results = method(params[:action]).call(target, body)
+
+      # Since this will only be on one node we can just return the first result
+      result = scrub_stack_trace(results.first.status_hash)
+      [200, result.to_json]
     end
 
     post '/winrm/:action' do
@@ -121,10 +120,13 @@ module BoltServer
       return [400, error.to_json] unless error.nil?
 
       opts = body['target'].clone.merge('protocol' => 'winrm')
-
       target = [Bolt::Target.new(body['target']['hostname'], opts)]
 
-      method(params[:action]).call(target, body)
+      results = method(params[:action]).call(target, body)
+
+      # Since this will only be on one node we can just return the first result
+      result = scrub_stack_trace(results.first.status_hash)
+      [200, result.to_json]
     end
 
     error 404 do
--- a/spec/bolt_server/transport_app_spec.rb
+++ b/spec/bolt_server/transport_app_spec.rb
@@ -38,288 +38,259 @@ describe "BoltServer::TransportApp" do
     expect(last_response.status).to eq(200)
   end
 
-  context 'with ssh target', ssh: true do
-    let(:target) { conn_info('ssh') }
-    let(:path) { '/ssh/run_task' }
-    let(:echo_task) {
-      {
-        'name': 'sample::echo',
-        'metadata': {
-          'description': 'Echo a message',
-          'parameters': { 'message': 'Default message' }
-        },
-        files: [{
-          filename: "echo.sh",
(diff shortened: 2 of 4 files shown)


lib/bolt_server/transport_app.rb
```diff
@@ -62,11 +62,7 @@ def run_task(target, body)
 
       task = Bolt::Task::PuppetServer.new(body['task'], @file_cache)
       parameters = body['parameters'] || {}
-      results = @executor.run_task(target, task, parameters)
-
-      # Since this will only be on one node we can just return the first result
-      result = scrub_stack_trace(results.first.status_hash)
-      [200, result.to_json]
+      @executor.run_task(target, task, parameters)
```
caseywilliams: Moving the HTTP response details to the caller here seemed like a simple way to make `run_task` more focused on just the task operations, but open to changing this

spec/bolt_server/transport_app_spec.rb
```diff
@@ -38,288 +38,255 @@ def file_data(file)
...
-        'task': echo_task,
-        'target': {
-          'password': 'foo',
-          'private-key-content': 'content'
-        },
-        'parameters': { "message": "Hello!" }
-      }
-
-      post path, JSON.generate(body), 'CONTENT_TYPE' => 'text/json'
+  context 'when raising errors' do
```
caseywilliams: I moved these app-wide test cases further up, above the transport-specific and action-specific tests. I also removed the task body content passed to this 404 test case, since they should never reach validation anyway .

spec/bolt_server/transport_app_spec.rb
```diff
@@ -38,288 +38,255 @@ def file_data(file)
...
-      result = last_response.body
-      expect(result).to match(%r{The property '#/target' of type object did not match any of the required schemas})
+      expect(last_response.status).to eq(500)
+      result = JSON.parse(last_response.body)
+      expect(result['msg']).to eq('500: Unknown error: Unexpected error')
+      expect(result['kind']).to eq('boltserver/server-error')
+    end
+  end
+
+  describe 'transport routes' do
```
caseywilliams: My goal with moving these test cases to a "transport routes" section was to unit test the `/ssh/:action` and `/winrm/:action` logic without making the test content specific to one action. These tests no longer use the `:ssh` or `:winrm` tags because the functions they call have been mocked. FWIW, there are actual echo task executions for ssh and winrm further down, but if we'd rather actually run these in the docker/vagrant environment, I can take this back to the drawing board.

spec/bolt_server/transport_app_spec.rb
```diff
@@ -38,288 +38,255 @@ def file_data(file)
...
-          'description': 'Echo a message',
-          'parameters': { 'message': 'Default message' }
-        },
-        'file': {
-          'file_content': Base64.encode64("#!/usr/bin/env bash\necho $PT_message"),
-          'filename': "echo.sh"
-        }
-      }
-    }
+  describe 'action endpoint' do
```
caseywilliams: My intent is to use this section for testing of other endpoints like run_command, run_script, or upload_file.

caseywilliams:
:worried: I'm not sure I understand why there's a connection failure in appveyor -- the same test seems to be passing for me locally with vagrant

spec/bolt_server/transport_app_spec.rb
```diff
@@ -38,288 +38,259 @@ def file_data(file)
...
 
-    it 'returns non-html 500 when the request times out' do
-      get '/500_error'
-      expect(last_response).not_to be_ok
-      expect(last_response.status).to eq(500)
-      r = JSON.parse(last_response.body)
-      expect(r['msg']).to eq('500: Unknown error: Unexpected error')
-      expect(r['kind']).to eq('boltserver/server-error')
+    describe 'run_task' do
+      it 'runs a simple echo task over SSH' do
```
donoghuc: There are no available ssh targets on appveyor. You will need to filter this with something like `ssh: true` (similar to the :winrm, below). https://github.com/puppetlabs/bolt/blob/45056d869f7377e196e6b0aeb177d3515ed3d63f/Rakefile#L26-L30

caseywilliams: Gah, :woman_facepalming: Thank you @donoghuc!

caseywilliams:
Hmm, now Appveyor is passing, but Travis is failing on docs generation -- I'm not sure what's up, since it was passing before

donoghuc:
Uhh, my best guess is the newest yard gem release changes how puppet-strings works? Put up https://github.com/puppetlabs/bolt/pull/1061 but I want to verify that we actually want it. Sorry!

spec/bolt_server/transport_app_spec.rb
McdonaldSeanp: Since there are executions of actual executions of ssh/winrm further down I think there's a low level of risk in your approach here. 👍

PR #1061, opened 2019-06-28
(docs) Pull in updates to auto-generated docs for plan functions

donoghuc:
Update to yard gem appears to have changed the way puppet-strings generates the docs. This commit updates the plan reference based on those changes.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -88,7 +88,8 @@ task :docs do
   #       text: example body
   @functions = funcs.map do |func|
     func['text'] = func['docstring']['text']
-
+    # TODO: remove once puppet-strings releases fix for :'file::read'
+    func['name'] = func['name'][1..-1].gsub("'", '') if func['name'].start_with?(':')
     overloads = func['docstring']['tags'].select { |tag| tag['tag_name'] == 'overload' }
     sig_tags = overloads.map { |overload| overload['docstring']['tags'] }
     sig_tags = [func['docstring']['tags']] if sig_tags.empty?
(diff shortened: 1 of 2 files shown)


donoghuc:
Put this up just to see how it renders in markdown. Pinning `yard` to `0.9.19` seems to change the output. Still evaluating if we want to pull in these changes.

pre-docs/plan_functions.md
```diff
@@ -1,6 +1,177 @@
 # Bolt Functions
 
 
+## :'ctrl::do_until'
```
donoghuc: Oh, boo it prepends non-core functions with a `:` which I think changes the order. Perhaps we should pin yard?

lucywyman: I'm fine with pinning yard if that fixes it. I also don't think it's a huge issue if it's not quick to fix. I'm more urgent about getting travis passing.

Rakefile
```diff
@@ -88,7 +88,8 @@ task :docs do
   #       text: example body
   @functions = funcs.map do |func|
     func['text'] = func['docstring']['text']
-
+    # TODO: remove once puppet-strings releases fix for :'file::read'
+    func['name'] = func['name'][1..-1].gsub("'", '') if func['name'].start_with?(':')
```
donoghuc: If this had not affected ordering I would be more inclined to do it. so i think pinning makes sense. https://github.com/puppetlabs/bolt/pull/1062

donoghuc:
Chose to pin yard for now. Will follow up on a fix with puppet-strings.

PR #1063, opened 2019-06-28
(BOLT-1398) Add inventory show command

beechtom:
This adds support for an `inventory show` command that lists target names for targets which match the targetting option and are in the inventory.

### DIFF
--- a/lib/bolt/bolt_option_parser.rb
+++ b/lib/bolt/bolt_option_parser.rb
@@ -28,6 +28,9 @@ module Bolt
       when 'file'
         { flags: ACTION_OPTS + %w[tmpdir],
           banner: FILE_HELP }
+      when 'inventory'
+        { flags: OPTIONS[:inventory] + OPTIONS[:global] + %w[format inventoryfile boltdir configfile],
+          banner: INVENTORY_HELP }
       when 'plan'
         case action
         when 'convert'
@@ -110,6 +113,7 @@ Available subcommands:
   bolt secret createkeys           Create new encryption keys
   bolt secret encrypt <plaintext>  Encrypt a value
   bolt secret decrypt <encrypted>  Decrypt a value
+  bolt inventory show              Show the list of targets an action would run on
 
 Run `bolt <subcommand> --help` to view specific examples.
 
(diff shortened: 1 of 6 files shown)


beechtom:
Still need to write tests, but wanted to make sure this is the behavior we want before doing them.

lib/bolt/bolt_option_parser.rb
```diff
@@ -28,6 +28,9 @@ def get_help_text(subcommand, action = nil)
       when 'file'
         { flags: ACTION_OPTS + %w[tmpdir],
           banner: FILE_HELP }
+      when 'inventory'
+        { flags: OPTIONS[:inventory] + OPTIONS[:global],
```
beechtom: Add `format` as an option. Does it make sense to have all display options available?

adreyer: I think all the display options make sense. 
`--boltdir` and `--configfile` should probably be supported too. I think it makes sense to move them to a separate group from `modulepath`

lib/bolt/bolt_option_parser.rb
```diff
@@ -263,9 +267,19 @@ def self.examples(cmd, desc)
...
         decrypt                  Decrypt a value
+
       Available options are:
     SECRET_HELP
 
+    INVENTORY_HELP = <<~INVENTORY_HELP
+      Usage: bolt inventory <action>
+
+      Available actions are:
+        show                     Show list of targets in an inventory group
```
adreyer: `Show the list of targets an action would run on` or something? Since it's supports `query` and `rerun` I think that makes the most sense.

lib/bolt/outputter/human.rb
```diff
@@ -311,6 +311,11 @@ def print_module_list(module_list)
         end
       end
 
+      def print_targets(options)
+        targets = options[:targets].map(&:name)
+        puts targets.join("\n"), "\nNUMBER OF TARGETS: #{targets.count}"
```
adreyer: bikeshed: How do we feel about this capitalization?

I think it makes sense to do something that makes this stand apart from target names but just having spaces should do that and we have color to utilize as well?

beechtom: Ya, I'm not fully happy with the capitalization. Latest commit switches to just the number in green text, which looks a bit nicer.

adreyer: I think "#{targets.count} targets" in green is better. Just the number is a bit cryptic.

lib/bolt/outputter/human.rb
```diff
@@ -311,6 +311,12 @@ def print_module_list(module_list)
         end
       end
 
+      def print_targets(options)
+        targets = options[:targets].map(&:name)
+        @stream.puts targets.join("\n")
+        @stream.puts colorize(:green, "#{targets.count} targets")
```
lucywyman: I don't think we care too much about this, but it'd be nice if we did '1 target' when there's exactly one target. I think '0 targets' reads better than '0 target', so I'd leave that be.

PR #1064, opened 2019-07-01
(BOLT-1445, BOLT-1455) Update log level docs

donoghuc:
This commit updates the log level documentation in the writing plan docs to include updated information about how to log messages from plan execution. It also adds a section to the common debugging page with some prescriptive recommendations for handling plan logging.

### DIFF
--- a/docs/debugging.md
+++ b/docs/debugging.md
@@ -42,6 +42,10 @@ By default, Bolt tries to connect over the standard SSH port 22. If you need to 
 
 Make sure you've specified the `winrm` protocol for the target. You can either include it in the name of the target (`winrm://hostname.example.com`), pass `--transport winrm` on the CLI, or set the transport in your Bolt config or inventory. By default, Bolt will try to connect over SSH.
 
+## Puppet log functions are not logging to the console
+
+The default log level for the console is `warn`. When a `notice` function is used in a plan, it will not be printed to the console. When you have messages you want to be printed to the console regardless of log level you should use the `out::message` plan function. The [`out::message`](../pre-docs/plan_functions.md#outmessage) function is not available for use in an apply block and only accepts string values. If you need to send a message that is not a String value or is in an apply block you can use the `warning` Puppet log function. The `notice` Puppet log function could be used when you only wish to see the output in the console when executing your plan with the `--debug` flag which will set the console log level to `debug` for that run. See the docs for configuring the [Bolt's log level](https://puppet.com/docs/bolt/latest/bolt_configuration_options.html#log-file-configuration-options) for more information about how to configure the log levels.
+
 ## I still need help
 
 Visit the **#bolt** channel in the [Puppet Community Slack](https://slack.puppet.com) and you will find a whole community of people waiting to help!
--- a/docs/debugging.md
+++ b/docs/debugging.md
@@ -44,7 +44,11 @@ Make sure you've specified the `winrm` protocol for the target. You can either i
 
 ## Puppet log functions are not logging to the console
 
-The default log level for the console is `warn`. When a `notice` function is used in a plan, it will not be printed to the console. When you have messages you want to be printed to the console regardless of log level you should use the `out::message` plan function. The [`out::message`](../pre-docs/plan_functions.md#outmessage) function is not available for use in an apply block and only accepts string values. If you need to send a message that is not a String value or is in an apply block you can use the `warning` Puppet log function. The `notice` Puppet log function could be used when you only wish to see the output in the console when executing your plan with the `--debug` flag which will set the console log level to `debug` for that run. See the docs for configuring the [Bolt's log level](https://puppet.com/docs/bolt/latest/bolt_configuration_options.html#log-file-configuration-options) for more information about how to configure the log levels.
+The default log level for the console is `warn`. When a `notice` function is used in a plan, it will not be printed to the console. When you have messages you want to be printed to the console regardless of log level you should use the `out::message` plan function. The [`out::message`](../pre-docs/plan_functions.md#outmessage) function is not available for use in an apply block and only accepts string values.
+
+If you need to send a message that is not a String value or is in an apply block you can use the `warning` Puppet log function. The `notice` Puppet log function could be used when you only wish to see the output in the console when executing your plan with the `--debug` flag which will set the console log level to `debug` for that run.
+
+See the docs for configuring the [Bolt's log level](https://puppet.com/docs/bolt/latest/bolt_configuration_options.html#log-file-configuration-options) for more information about how to configure the log levels.
 
 ## I still need help
 
--- a/lib/bolt/outputter/json.rb
+++ b/lib/bolt/outputter/json.rb
@@ -21,6 +21,8 @@ module Bolt
         case event[:type]
         when :node_result
           print_result(event[:result])
+        when :message
+          print_message_event(event)
         end
       end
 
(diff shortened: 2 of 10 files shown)


docs/debugging.md
```diff
@@ -42,6 +42,10 @@ By default, Bolt tries to connect over the standard SSH port 22. If you need to
 
 Make sure you've specified the `winrm` protocol for the target. You can either include it in the name of the target (`winrm://hostname.example.com`), pass `--transport winrm` on the CLI, or set the transport in your Bolt config or inventory. By default, Bolt will try to connect over SSH.
 
+## Plan functions are not logging to the console
+
+The default log level for the console is `warn`. When a `notice` function is used in a plan, it will not be printed to the console. You can see these messages by running the plan with the `--debug` parameter or by configuring the (console log level)(https://puppet.com/docs/bolt/latest/bolt_configuration_options.html#log-file-configuration-options).  You could also use the `warning` log function instead of `notice`. Another option is to use the `out::message` plan function which will print string messages to the console, regardless of log level.
```
adreyer: Should we be more prescriptive here?

If you have messages you always want to be printed use the `out::message` plan function instead of `notice`. If you are temporarily debugging a plan use `warning` or `out::message` to print messages. To permanently add debug messages to a plan use `notice` and pass `--debug` when running the plan.

donoghuc: Yeah, being prescriptive here makes sense. I think i could also add a note https://puppet.com/docs/bolt/latest/writing_plans.html#puppet-log-functions describing what to expect between logger and outputter.

adreyer: Yeah I think rewriting that section to be primarily out::message based makes sense.


also puppetdb should probably have it's own section

docs/debugging.md
```diff
@@ -42,6 +42,10 @@ By default, Bolt tries to connect over the standard SSH port 22. If you need to
...
 
+## Puppet log functions are not logging to the console
+
+The default log level for the console is `warn`. When a `notice` function is used in a plan, it will not be printed to the console. When you have messages you want to be printed to the console regardless of log level you should use the `out::message` plan function. The `out::message` function is not available for use in an apply block and only accepts string values. If you need to send a message that is not a String value or is in an apply block you can use the `warning` Puppet log function. The `notice` Puppet log function could be used when you only wish to see the output in the console when executing your plan with the `--debug` flag which will set the console log level to `debug` for that run. See the docs for configuring the [Bolt's log level](https://puppet.com/docs/bolt/latest/bolt_configuration_options.html#log-file-configuration-options) for more information about how to configure the log levels.
```
lucywyman: I'm not sure where to slot this in, but it'd be nice to link to the `out::message` function docs so people can see an example. I think you just have (<docs link>) in parens or something after first introducing it.

donoghuc: updated

pre-docs/writing_plans.md
```diff
@@ -494,18 +494,36 @@ plan run_with_facts(TargetSpec $nodes) {
...
 [Connecting Bolt to PuppetDB](bolt_connect_puppetdb.md)
 
 ## Plan logging
 
-Set up log files to record certain events that occur when you run plans.
+Plan information can be captured in log files or printed to a terminal session. This section details the methods for collecting information about plan runs.
+
+### Outputting message to the terminal
+
+The plan function [`out::message`](plan_functions.md#outmessage) can be used to print message strings to `STDOUT`. These message will always be printed regardless of the log level and will not be logged to the log file.
```
adreyer: it looks like we're not printing it with format=json yet. 
https://github.com/puppetlabs/bolt/blob/master/lib/bolt/outputter/json.rb#L109
we should really fix that.

lucywyman: Created https://tickets.puppetlabs.com/browse/BOLT-1455 - feel free to add more detail!

donoghuc:
@lucy we should probably update https://github.com/puppetlabs/bolt/blob/c049c3609ac2d3af88b9f772cdfac93d4950485d/bolt-modules/out/lib/puppet/functions/out/message.rb#L5 and rebuild the plan function docs

lucywyman:
@donoghuc Great point I completely spaced that! Updated

lib/bolt/outputter/json.rb
```diff
@@ -21,6 +21,8 @@ def handle_event(event)
         case event[:type]
         when :node_result
           print_result(event[:result])
```
donoghuc: I think that maybe subscribing the json outputter prints intermediate node results. (disclaimer this is the last comment I am writing before a long vacation, so I could be way off).

docs/debugging.md
```diff
@@ -42,6 +42,10 @@ By default, Bolt tries to connect over the standard SSH port 22. If you need to
...
 
+## Puppet log functions are not logging to the console
+
+The default log level for the console is `warn`. When a `notice` function is used in a plan, it will not be printed to the console. When you have messages you want to be printed to the console regardless of log level you should use the `out::message` plan function. The [`out::message`](../pre-docs/plan_functions.md#outmessage) function is not available for use in an apply block and only accepts string values. If you need to send a message that is not a String value or is in an apply block you can use the `warning` Puppet log function. The `notice` Puppet log function could be used when you only wish to see the output in the console when executing your plan with the `--debug` flag which will set the console log level to `debug` for that run. See the docs for configuring the [Bolt's log level](https://puppet.com/docs/bolt/latest/bolt_configuration_options.html#log-file-configuration-options) for more information about how to configure the log levels.
```
beechtom:
```suggestion
The default log level for the console is `warn`. When a `notice` function is used in a plan, it will not be printed to the console. When you have messages you want to be printed to the console regardless of log level you should use the `out::message` plan function. The [`out::message`](../pre-docs/plan_functions.md#outmessage) function is not available for use in an apply block and only accepts string values. 

If you need to send a message that is not a String value or is in an apply block you can use the `warning` Puppet log function. The `notice` Puppet log function could be used when you only wish to see the output in the console when executing your plan with the `--debug` flag which will set the console log level to `debug` for that run. 

See the docs for configuring [Bolt's log level](https://puppet.com/docs/bolt/latest/bolt_configuration_options.html#log-file-configuration-options) for more information about how to configure the log levels.
```

pre-docs/writing_plans.md
```diff
@@ -494,18 +494,36 @@ plan run_with_facts(TargetSpec $nodes) {
 }
 ```
 
-**Related information**  
+### Collect general data from PuppetDB
+
+You can use the `puppetdb_query` function in plans to make direct queries to PuppetDB. For example you can discover nodes from PuppetDB and then run tasks on them. You'll have to configure the [puppetdb client](bolt_connect_puppetdb.md)before running it. You can learn how to [structure pql queries here](https://puppet.com/docs/puppetdb/latest/api/query/tutorial-pql.html), and find [pql reference and examples here](https://puppet.com/docs/puppetdb/latest/api/query/v4/pql.html)
```
beechtom:
```suggestion
You can use the `puppetdb_query` function in plans to make direct queries to PuppetDB. For example you can discover nodes from PuppetDB and then run tasks on them. You'll have to configure the [puppetdb client](bolt_connect_puppetdb.md) before running it. You can learn how to [structure pql queries here](https://puppet.com/docs/puppetdb/latest/api/query/tutorial-pql.html), and find [pql reference and examples here](https://puppet.com/docs/puppetdb/latest/api/query/v4/pql.html)
```

Issue #666, opened 2018-09-28
SSH disconnect hangs when using `ProxyCommand`


t-8ch:
# Environment

* Ubuntu 16.04
* Ruby: 2.5.1p57 (from ubuntu package)
* Bolt: 0.24.0 (from ubuntu package)

# Observation

When trying to run a task on Server A the output stop after printing `Started on $HOSTNAME...`, no output is printed and bolt does not return to the shell.
The same command works on Server B, it prints `Finished on $HOSTNAME`, the stdout of the task and the final report.
The only difference between Server A and B is the fact, that Server A uses the follwoing SSH config
```
ProxyCommand ssh $PROXY netcat -w 120 %h %p
User root
```

The same behaviour as Server A can be reproduced on more server using a `ProxyCommand`.

Executing bolt with `--debug` show that the command is actually executed:
```
# [..]
Started on $HOSTNAME
Running task $TASK
Opened session
Executing: mkdir -m 700 /tmp/9268c9c1-4087-4e89-a56a-859e0c280884
Command returned successfully
Executing: chmod u\+x /tmp/9268c9c1-4087-4e89-a56a-859e0c280884/init.rb
Command returned successfully
Executing: /tmp/9268c9c1-4087-4e89-a56a-859e0c280884/init.rb
stdout: {"status":"running","enabled":"true"}
Command returned successfully
Executing: rm -rf /tmp/9268c9c1-4087-4e89-a56a-859e0c280884
Command returned successfully
# never finishes
```

Browsing the source a little bit, it seems that the closing of the underlying SSH connection/session is the culprit.

# Expected behaviour

Server A works like Server B

TheMeier:
@MikaelSmith any update here? Is there a corresponding Jira issue?

donoghuc:
Hi @TheMeier I have filed https://tickets.puppetlabs.com/browse/BOLT-1423 . Are you running in to a similar issue? Are you setting [proxyjump](https://puppet.com/docs/bolt/latest/bolt_configuration_options.html#ssh-transport-configuration-options) in Bolt's ssh configuration options in addition to `ProxyConfig` in you OpenSSH configuration options?

TheMeier:
@donoghuc we have a propriatary jump host wich does not work with `ProxyJump`. We use a config like this on the agent
```
jump.host.
  ControlMaster auto
  ControlPath ~/.ssh/connections/%r_%h_%p
  ControlPersist no
  ProxyCommand none

[...]

PR #1065, opened 2019-07-01
(BOLT-1423) Ensure ssh session is closed after disconnect-timeout

donoghuc:
In certain ssh configurations the socket.close method for net-ssh https://github.com/net-ssh/net-ssh/blob/dd13dd44d68b7fa82d4ca9a3bbe18e30c855f1d2/lib/net/ssh/transport/session.rb#L130 can hang. This commit implements the use of a fallback method defined in the net-ssh session class called `shutdown!` after a configurable timeout period to avoid hanging indefinitely when net-ssh cannot close a socket.

### DIFF
--- a/lib/bolt/transport/ssh.rb
+++ b/lib/bolt/transport/ssh.rb
@@ -10,7 +10,7 @@ module Bolt
     class SSH < Sudoable
       def self.options
         %w[host port user password sudo-password private-key host-key-check
-           connect-timeout tmpdir run-as tty run-as-command proxyjump interpreters]
+           connect-timeout disconnect-timeout tmpdir run-as tty run-as-command proxyjump interpreters]
       end
 
       def self.default_options
@@ -18,7 +18,8 @@ module Bolt
           'connect-timeout' => 10,
           'host-key-check' => true,
           'tty' => false,
-          'load-config' => true
+          'load-config' => true,
+          'disconnect-timeout' => 5
         }
       end
 
@@ -42,10 +43,12 @@ module Bolt
           end
         end
 
-        timeout_value = options['connect-timeout']
-        unless timeout_value.is_a?(Integer) || timeout_value.nil?
-          error_msg = "connect-timeout value must be an Integer, received #{timeout_value}:#{timeout_value.class}"
-          raise Bolt::ValidationError, error_msg
+        %w[connect-timeout disconnect-timeout].each do |timeout|
+          timeout_value = options[timeout]
+          unless timeout_value.is_a?(Integer) || timeout_value.nil?
+            error_msg = "#{timeout} value must be an Integer, received #{timeout_value}:#{timeout_value.class}"
+            raise Bolt::ValidationError, error_msg
+          end
         end
       end
 
--- a/lib/bolt/transport/ssh/connection.rb
+++ b/lib/bolt/transport/ssh/connection.rb
@@ -130,7 +130,11 @@ module Bolt
 
         def disconnect
           if @session && !@session.closed?
-            @session.close
+            begin
+              Timeout.timeout(@target.options['disconnect-timeout']) { @session.close }
+            rescue Timeout::Error
+              @session.shutdown!
+            end
             @logger.debug { "Closed session" }
           end
         end
--- a/lib/bolt_server/schemas/partials/target-ssh.json
+++ b/lib/bolt_server/schemas/partials/target-ssh.json
@@ -28,6 +28,10 @@
       "type": "integer",
       "description": "How long Bolt should wait when establishing connections"
     },
+    "disconnect-timeout": {
+      "type": "integer",
+      "description": "How long Bolt should wait before forcing a disconnect"
+    },
     "run-as-command": {
       "type": "array",
       "description": "Command elevate permissions",
--- a/pre-docs/bolt_configuration_options.md
+++ b/pre-docs/bolt_configuration_options.md
@@ -54,6 +54,8 @@ to avoid writing them to disk.
 
 `connect-timeout`: How long Bolt should wait when establishing connections.
 
+`disconnect-timeout`: How long Bolt should wait to force-close an ssh connection.
+
 `host-key-check`: Whether to perform host key validation when connecting over SSH. Default is `true`.
 
 `password`: Login password.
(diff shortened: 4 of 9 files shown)


lib/bolt/transport/ssh/connection.rb
```diff
@@ -130,7 +130,11 @@ def connect
 
         def disconnect
           if @session && !@session.closed?
-            @session.close
+            begin
+              Timeout.timeout(@target['connect-timeout']) { @session.close }
```
donoghuc: I am having a hard time reproducing the case described in https://github.com/puppetlabs/bolt/issues/666. Further conversation in slack revealed that there are other configuration details besides the simple `ProxyCommand`. Instead there is a combination of `ForceCommand` and `Command*`. I think that between the unsupported net-ssh `Control*` settings and the proprietary/custom jump host used by the reporter we get in to a situation where net-ssh hangs forever trying to close a socket. Here is the config I am using to try to repro this:
## local ~/.ssh/config
```
Host p71tjeuwlftc81l.delivery.puppetlabs.net
  IdentityFile ~/.ssh/id_rsa-acceptance
  User root
  StrictHostKeyChecking no
  HostName p71tjeuwlftc81l
  Port 22
  ProxyCommand ssh -q root@q0an0ruursqn43r.delivery.puppetlabs.net
  ControlMaster auto
  ControlPath ~/.ssh/sockets/%r@%h-%p
  ControlPersist 600
```
# Proxy ~/.ssh/config
```
Host p71tjeuwlftc81l.delivery.puppetlabs.net
  IdentityFile ~/.ssh/id_rsa-acceptance
  User root
  StrictHostKeyChecking no
  HostName p71tjeuwlftc81l
  Port 22
  ControlMaster auto
  ControlPath ~/.ssh/sockets/%r@%h-%p
  ControlPersist 600
```
# Proxy sshd config
```
root@q0an0ruursqn43r:~# cat /etc/ssh/sshd_config 
# File is managed by Puppet
ForceCommand /bin/test.sh  
AcceptEnv LANG LC_*
ChallengeResponseAuthentication no
GSSAPIAuthentication no
PermitRootLogin yes
PrintMotd no
Subsystem sftp /usr/lib/openssh/sftp-server
UseDNS no
UsePAM yes
X11Forwarding yes
```
## test.sh
```
#!/bin/bash
nc p71tjeuwlftc81l 22
```
Using the `ForceCommand` to execute a shell script mirrors what the reporter's system does. However even with this configuration I am unable to reproduce the hanging behavior. 

I put up this PR based on some comments in the source code of net-ssh and arbitrarily used the same `connect-timeout` for `disconnect-timeout`. If we can reproduce this or verify this patch works then perhaps we can introduce another setting and write some unit tests for this.

TheMeier:
This patch fixes the problems i have in #666

lib/bolt/transport/ssh/connection.rb
donoghuc: Re-examining this work shows that I had introduced a (pretty serious) bug. I should have used `@target.options['connect-timeout']` by omitting  `.options` a `NoMethodError` was raised here as a result of trying to use a `[]` accessor on a `Target` object. This was caught in the blanket `StandardError` handling at https://github.com/puppetlabs/bolt/blob/d739a203bc4a7e5afc4c2b650517e1c50fa153ab/lib/bolt/transport/ssh.rb#L74 . So I think this resulted in the `shutdown!` method never being called.

donoghuc:
I investigated the effects of using the `shutdown!` method when using the openssh configuration options to use compression. Given each session is re-initialized for each run_* plan step I think it is safe to not reset the compressor/decompressor state in https://github.com/net-ssh/net-ssh/blob/8f38e64cabe44f2c5ca33c1b859446967162e04e/lib/net/ssh/transport/state.rb#L167 . I observed the only consequence is a warning from zlib `zlib(finalizer): the stream was freed prematurely.` which is logged as expected. I confirmed a plan like 
```puppet
plan test (TargetSpec $nodes, Integer $data_size = 10) {
  $result_1 = run_command("printf \"%0.sa\" {1..${data_size}}", $nodes).first.value['stdout']
  notice($result_1.length)
  $check_quality_1 = $result_1.filter |$l| {$l == 'a'}
  notice($check_quality_1.length)
  $result_2 = run_command("printf \"%0.sa\" {1..${data_size}}", $nodes).first.value['stdout']
  notice($result_2.length)
  $check_quality_2 = $result_2.filter |$l| {$l == 'a'}
  notice($check_quality_2.length)
}
```
did not drop/corrupt any data with very large output.

lib/bolt/transport/ssh.rb
```diff
@@ -42,10 +43,12 @@ def self.validate(options)
           end
         end
 
-        timeout_value = options['connect-timeout']
-        unless timeout_value.is_a?(Integer) || timeout_value.nil?
-          error_msg = "connect-timeout value must be an Integer, received #{timeout_value}:#{timeout_value.class}"
-          raise Bolt::ValidationError, error_msg
+        %w[connect-timeout disconnect-timeout].each do |timeout|
+          timeout_value = options[timeout]
+          unless timeout_value.is_a?(Integer) || timeout_value.nil?
```
donoghuc: Enforcing that this value is an integer effectively makes 1 second the shortest possible timeout. We could allow floats here to allow even shorter timeouts in the case where the user knows that the connection will always hang otherwise.

lib/bolt_server/schemas/partials/target-ssh.json
```diff
@@ -28,6 +28,10 @@
       "type": "integer",
       "description": "How long Bolt should wait when establishing connections"
     },
+    "disconnect-timeout": {
+      "type": "integer",
+      "description": "How long Bolt should wait when establishing connections"
```
nicklewis: Need to actually change the description

pre-docs/bolt_configuration_options.md
```diff
@@ -54,6 +54,8 @@ to avoid writing them to disk.
 
 `connect-timeout`: How long Bolt should wait when establishing connections.
 
+`diconnect-timeout`: How long Bolt should wait to force-close an ssh connection.
```
nicklewis:
```suggestion
`disconnect-timeout`: How long Bolt should wait to force-close an SSH connection.
```

lib/bolt/transport/ssh.rb
donoghuc: Further conversation on slack revealed that 1 second is a reasonable minimum for now. We can wait and see if this needs to be updated.

PR #1068, opened 2019-07-02
(maint) Remove Hipchat notifications

glennsarti:
Hipchat is no longer used for notifications.  This commit removes this from
Travis.


lucywyman:
Thanks Glenn!

PR #1070, opened 2019-07-02
(BOLT-1410) Report plugin usage analytics for bundled plugins

adreyer:
Should have integration tests for analytics?

### DIFF
--- a/lib/bolt/inventory/group2.rb
+++ b/lib/bolt/inventory/group2.rb
@@ -98,18 +98,27 @@ module Bolt
       def config_only_plugin(data)
         Bolt::Util.walk_vals(data) do |value|
           if value.is_a?(Hash) && value.include?('_plugin')
-            unless (plugin = @plugins.by_name(value['_plugin']))
-              raise ValidationError.new("Config lookup specifies an unknown plugin: #{value['_plugin'].inspect}", @name)
+            plugin_name = value['_plugin']
+            begin
+              hook = @plugins.get_hook(plugin_name, :inventory_config)
+            rescue Bolt::Plugin::PluginError => e
+              raise ValidationError.new(e.message, @name)
             end
-            unless plugin.hooks.include?('inventory_config')
-              raise ValidationError.new("#{plugin.name} does not support inventory_config.", @name)
+
+            begin
+              validate_proc = @plugins.get_hook(plugin_name, :validate_inventory_config)
+            rescue Bolt::Plugin::PluginError
+              validate_proc = proc { |*args| }
             end
-            plugin.validate_inventory_config(value) if plugin.respond_to?(:validate_inventory_config)
+
+            validate_proc.call(value)
+
             Concurrent::Delay.new do
               begin
-                plugin.inventory_config(value)
+                hook.call(value)
               rescue StandardError => e
-                raise Bolt::Plugin::PluginError.new(e.message, plugin, "inventory_targets in #{@name}")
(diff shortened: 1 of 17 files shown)


lib/bolt/inventory/group2.rb
```diff
@@ -98,18 +98,27 @@ def validate_config_plugin(data, key, group_name = nil)
...
+            begin
+              hook = @plugins.get_hook(plugin_name, :inventory_config)
+            rescue Bolt::Plugin::PluginError => e
+              raise ValidationError.new(e.message, @name)
             end
-            unless plugin.hooks.include?('inventory_config')
-              raise ValidationError.new("#{plugin.name} does not support inventory_config.", @name)
+
+            begin
+              validate = @plugins.get_hook(plugin_name, :validate_inventory_config)
```
lucywyman: Should this be `hook` instead of `validate` to avoid conflicting with the method of the same name?

lib/bolt/inventory/group2.rb
```diff
@@ -98,18 +98,27 @@ def validate_config_plugin(data, key, group_name = nil)
...
+            rescue Bolt::Plugin::PluginError
+              validate = proc { |*args| }
```
lucywyman: Is there a reason not to move `validate.call` into the `begin` and just have an empty rescue (besides rubocop)?

spec/bolt/cli_spec.rb
```diff
@@ -770,34 +770,40 @@ def stub_config(file_content = {})
...
-        expect(cli.bundled_content).not_to be_empty
+        expect(cli.bundled_content['Task']).not_to be_empty
       end
 
       it "calculates bundled content for a plan" do
         cli = Bolt::CLI.new(%w[plan run foo --nodes bar])
         cli.parse
-        expect(cli.bundled_content).not_to be_empty
+        expect(cli.bundled_content['Plan']).not_to be_empty
+        expect(cli.bundled_content['Task']).not_to be_empty
```
lucywyman: Not a big deal but this check seems redundant with the test above?

lib/bolt/plugin.rb
```diff
@@ -27,19 +41,32 @@ def self.setup(config, pdb_client)
...
+      raise PluginError::UnsupportedHook.new(plugin_name, hook) unless plugin.respond_to?(hook)
+      @analytics.report_bundled_content("Plugin #{hook}", plugin_name)
+
+      plugin.method(hook)
     end
 
     def by_name(plugin_name)
+      unless @reported.include?(plugin_name)
+        @reported << plugin_name
+        @analytics.report_bundled_content('Plugin', plugin_name)
```
lucywyman: Why do we submit this twice? The version with the hook seems like it has a superset of this information. Will this show up as 2 plugin uses in analytics?

lib/bolt/inventory/group2.rb
adreyer: yes errors from validation should not be squashed

spec/bolt/cli_spec.rb
adreyer: when executing a plan both task and plan data are needed. Given the current code path it shouldn't be different but it could if we optimized it. Checks are cheap.

lib/bolt/plugin.rb
adreyer: this was a bug left over from the initial implementation.

PR #1071, opened 2019-07-03
(BOLT-1425) Expose orchestrator-client_ruby settings in bolt conf

donoghuc:
Previously the `job-poll-{interval,timeout}` settings for the ruby orchestrator client were only configurable in `orchestrator.conf`. This commit exposes these settings in bolt configuration. If these are set in bolt config, they will override the settings in orchestrator.conf which is the desired behaviour because *only* the ruby implementation of the orchestrator client should use these settings.

### DIFF
--- a/lib/bolt/transport/orch.rb
+++ b/lib/bolt/transport/orch.rb
@@ -22,7 +22,7 @@ module Bolt
       attr_writer :plan_context
 
       def self.options
-        %w[host service-url cacert token-file task-environment]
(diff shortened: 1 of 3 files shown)


donoghuc:
Leaving up my PE setup with bolt_shim installed etc with following master:
```
pcp:
  service-url: "https://afsv8pe6jzrdefl.delivery.puppetlabs.net:8143"
  cacert: /tmp/ca.pem
  task-environment: production
```

PR #1073, opened 2019-07-03
(BOLT-1328) Add AWS EC2 target-lookup plugin

beechtom:
This adds a plugin that can lookup AWS EC2 instances in a specified region, allowing Bolt to dynamically determine the targets to use when running a command. Instances can also be filtered by tags and other instance properties.

### DIFF
--- a/lib/bolt/plugin/aws.rb
+++ b/lib/bolt/plugin/aws.rb
@@ -1,0 +1,103 @@
+# frozen_string_literal: true
+
+require 'json'
+
+module Bolt
+  class Plugin
+    class Aws
+      class EC2
+        attr_accessor :client
+        attr_reader :config
+
+        def initialize(config)
+          require 'aws-sdk-ec2'
+          @config = config
+          @logger = Logging.logger[self]
+        end
+
+        def name
+          'aws::ec2'
+        end
+
+        def hooks
+          %w[inventory_targets]
+        end
+
+        def config_client(opts)
+          return client if client
+
+          options = {}
+
+          if opts.key?('region')
+            options[:region] = opts['region']
+          end
+          if opts.key?('profile')
+            options[:profile] = opts['profile']
+          end
+          if config.aws&.key?('credentials')
+            creds = File.expand_path(config.aws['credentials'])
+            if File.exist?(creds)
+              options[:credentials] = ::Aws::SharedCredentials.new(path: creds)
+            else
+              raise Bolt::ValidationError, "Cannot load credentials file #{config.aws['credentials']}"
+            end
+          end
+
+          ::Aws::EC2::Client.new(options)
+        end
+
+        def inventory_targets(opts)
+          client = config_client(opts)
+          resource = ::Aws::EC2::Resource.new(client: client)
+
+          # Retrieve a list of EC2 instances and create a list of targets
+          # Note: It doesn't seem possible to filter stubbed responses...
+          resource.instances(filters: opts['filters']).map do |instance|
+            next unless instance.state.name == 'running'
+            target = {}
+
+            if opts.key?('uri')
+              uri = lookup(instance, opts['uri'])
+              target['uri'] = uri if uri
+            end
+            if opts.key?('name')
+              real_name = lookup(instance, opts['name'])
+              target['name'] = real_name if real_name
+            end
+            if opts.key?('config')
+              target['config'] = resolve_config(instance, opts['config'])
+            end
+
+            target if target['uri'] || target['name']
+          end.compact
+        end
+
+        # Look for an instance attribute specified in the inventory file
+        def lookup(instance, attribute)
+          value = instance.data.respond_to?(attribute) ? instance.data[attribute] : nil
+          unless value
+            warn_missing_attribute(instance, attribute)
+          end
+          value
+        end
+
+        def warn_missing_attribute(instance, attribute)
+          @logger.warn("Could not find attribute #{attribute} of instance #{instance.instance_id}")
+        end
+
+        # Walk the "template" config mapping provided in the plugin config and
+        # replace all values with the corresponding value from the resource
+        # parameters.
+        def resolve_config(name, config_template)
+          Bolt::Util.walk_vals(config_template) do |value|
+            if value.is_a?(String)
+              lookup(name, value)
+            else
+              value
+            end
+          end
+        end
+      end
+    end
+  end
+end
--- a/pre-docs/inventory_version2.md
+++ b/pre-docs/inventory_version2.md
@@ -83,6 +83,7 @@ The following plugins can be used for targets
 
 * `puppetdb` - Query PuppetDB to populate the targets.
 * `terraform` - Load a Terraform state file to populate the targets.
+* `aws::ec2` - Load running AWS EC2 instances to populate the targets.
 
 #### Config plugins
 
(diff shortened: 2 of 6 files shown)


lib/bolt/plugin/aws.rb
```diff
@@ -0,0 +1,107 @@
...
+        def config_client(opts)
+          options = {}
+
+          if opts.key?('region')
+            options[:region] = opts['region']
+          end
+          if opts.key?('profile')
+            options[:profile] = opts['profile']
+          end
+          if @config.aws && @config.aws.key?('credentials')
```
lucywyman: I would throw an error here in an else statement stating that you need to configure credentials to connect to aws.

lib/bolt/plugin/aws.rb
```diff
@@ -0,0 +1,107 @@
...
+            if File.exist?(creds)
+              options[:credentials] = ::Aws::SharedCredentials.new(path: creds)
+            end
+          end
+
+          ::Aws::EC2::Client.new(options)
+        end
+
+        def inventory_targets(opts)
+          @client ||= config_client(opts)
```
lucywyman: I would move this to the top of `config_client` as `return @client if @client`, and rename that function to just `client`. Then if we need the client elsewhere we won't  accidentally instantiate 2 clients, and the function itself knows if it's already been instantiated.

lib/bolt/plugin/aws.rb
beechtom: If credentials aren't provided in the config file, the aws sdk searches in `ENV['ACCESS_KEY_ID']` and `ENV['SECRET_ACCESS_KEY']`, followed by the default credentials location `~/.aws/credentials`. If no credentials are found, then the aws sdk throws a descriptive error that credentials are missing.

Would we want to check these default locations and throw our own error if none of them hold credentials, or require the user to specify credentials in the config? My original thought was the config file would allow for an alternate to the default location for the credentials.

lucywyman: In that case I'd catch AWS error and return a `Bolt::ValidationError` stating that you need to configure credentials.

beechtom: It looks like the AWS error is caught and a `Bolt::PluginError` is raised at `bolt/inventory/group2.rb:213`.

```
213:     raise Bolt::Plugin::PluginError.new(e.message, plugin, "inventory_targets in #{@name}")
```

lucywyman: Also! I think you can shorten this to just be `@config&.aws.key?('credentials')`

lib/bolt/plugin/aws.rb
```diff
@@ -0,0 +1,112 @@
...
+            target if target['uri'] || target['name']
+          end.compact
+        end
+
+        # Look for an instance property specified in the inventory file
+        # All instance properties are prefixed with 'instance.'
+        # If a property is not found, then a warning is displayed
+        # If a property is not an instance property, it's left as-is
+        def lookup(instance, property, default = nil)
+          if default.nil? || property.start_with?("instance.")
```
lucywyman: I remember you asking about this and I'm not sure why past-me said to keep `instance`, but I think you can have users just put the data name they want without `instance.` and not need this. It adds risk of `instance` being misspelled or something, and just makes the user type more unnecessarily. Sorry if I misled you!

lib/bolt/plugin/aws.rb
```diff
@@ -0,0 +1,112 @@
...
+
+          options = {}
+
+          if opts.key?('region')
+            options[:region] = opts['region']
+          end
+          if opts.key?('profile')
+            options[:profile] = opts['profile']
+          end
+          if config.aws&.key?('credentials')
```
lucywyman:
```suggestion
          if config&.aws.key?('credentials')
```

lib/bolt/plugin/aws.rb
```diff
@@ -0,0 +1,112 @@
...
+            creds = File.expand_path(config.aws['credentials'])
+            if File.exist?(creds)
+              options[:credentials] = ::Aws::SharedCredentials.new(path: creds)
+            else
+              @logger.warn("Cannot load credentials file #{config.aws['credentials']}")
```
lucywyman: I think if you specified an alternate credentials file and we don't find it we should error instead of warn. While it could theoretically fall back to a valid credentials file, the user (presumably) intended to use the credential file specified. Rather than behaving in a way the user didn't intend we should error and let the user either remove the aws config if they are not using it, or correct it if it's incorrect.

lib/bolt/plugin/aws.rb
```diff
@@ -0,0 +1,112 @@
...
+          resource = ::Aws::EC2::Resource.new(client: client)
+
+          # Retrieve a list of EC2 instances and create a list of targets
+          # Note: It doesn't seem possible to filter stubbed responses...
+          resource.instances(filters: opts['filters']).map do |instance|
+            next unless instance.state.name == 'running'
+            target = {}
+
+            if opts.key?('uri')
+              uri = lookup(instance, opts['uri'])
```
beechtom: Would we ever want to lookup a default property for either `uri` or `name` if one wasn't provided in the inventory? For example, if the user did not specify a uri, then lookup would default to `public_ip_address`?

On one hand this simplifies things if the user doesn't know what properties are available / want less set up to get the plugin working... on the other hand it seems kind of like 'magic'.

lib/bolt/plugin/aws.rb
```diff
@@ -0,0 +1,112 @@
...
+        end
+
+        def hooks
+          %w[inventory_targets]
+        end
+
+        def config_client(opts)
+          return client if client
+
+          options = {}
```
lucywyman: I think you can shorten this a bit, since we're not doing any checking or munging of the `region` or `profile` keys:
```
opts.transform_keys! &:to_s
```
then just use `opts` instead of `options` for the credentials.

lucywyman: To explain some ruby trickery: `!` will modify an object inplace, instead of creating a new one.
`&:` creates a new process that applies the following function to whatever it's operating on. Under the hood it calls `to_proc`, and is similar to doing `opts.transform_keys! { |k| k.to_sym }`

lib/bolt/plugin/aws.rb
```diff
@@ -0,0 +1,112 @@
...
+        end
+
+        # Look for an instance property specified in the inventory file
+        # All instance properties are prefixed with 'instance.'
+        # If a property is not found, then a warning is displayed
+        # If a property is not an instance property, it's left as-is
+        def lookup(instance, property, default = nil)
+          if default.nil? || property.start_with?("instance.")
+            instance_property = property.delete_prefix("instance.")
+            if instance.data.respond_to?(instance_property)
```
lucywyman: I think you can make this:
```
unless instance.data[instance_property]
  warn_missing_property
  default
end
instance.data[instance_property]
```

pre-docs/inventory_version2.md
```diff
@@ -68,7 +68,7 @@ behaves. Currently plugins are only supported for `inventory_targets`, in the
 `targets` section of inventory, and `inventory_config`, in the config section of
 inventory. Most plugins only work in one location or the other.
 
-#### Target plugins
+### Target plugins
```
lucywyman: I think we want to keep these at a heading 4

pre-docs/inventory_version2.md
```diff
@@ -352,7 +353,76 @@ google_compute_instance.app.1:
   zone = us-west1-a
 ```
 
-#### Prompt plugin
+### AWS EC2
+
+The AWS EC2 plugin supports looking up running AWS EC2 instances in a specified region. It supports several fields:
```
```suggestion
The AWS EC2 plugin supports looking up running AWS EC2 instances. It supports several fields:
```

pre-docs/inventory_version2.md
```diff
@@ -352,7 +353,76 @@ google_compute_instance.app.1:
...
+
+- `profile`: The [named profile](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-profiles.html) to use when loading from AWS `config` and `credentials` files. (optional, defaults to `default`)
+- `region`: The region to look up EC2 instances from. (optional)
```
lucywyman: I think this might actually be required by the `aws-sdk`?

lib/bolt/plugin/aws.rb
beechtom: Since `opts` might contain keys other than `region`, `profile`, or `credentials`, passing the original hash with these other keys will cause the aws sdk to raise an error.

pre-docs/inventory_version2.md
beechtom: Setting a region is definitely required, but it can be through an environment variable or through the credentials file. Do we want to require it being set in the inventory file only?

spec/bolt/plugin/aws_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+    expect(targets).to contain_exactly({ 'name' => name1 },
+                                       'name' => name2)
+  end
+
+  it 'returns nothing if neither name nor uri are specified' do
+    targets = plugin.inventory_targets({})
+    expect(targets).to be_empty
+  end
+
+  it 'builds a config map from the inventory' do
```
lucywyman: This feature is looking awesome!! I think adding a test that a validation error is raised if we don't find a credentials file would be good. So far everything is working perfectly as expected!

beechtom: Updated.

Issue #1060, opened 2019-06-28
Unable to use Windows Docker Daemon with Bolt Docker transport


glennsarti:
## What actually happens

When running a task on a Windows image running in Windows Docker Daemon, Bolt errors with:

```
Running task run '#<struct Bolt::Task name="puppet_agent::install", file=nil, files=[{"name"=>"install_shell.sh", "path"=>"C:/Source/puppet-strings/spec/fixtures/modules/puppet_agent/tasks/install_shell.sh"}, {"name"=>"install_powershell.ps1", "path"=>"C:/Source/puppet-strings/spec/fixtures/modules/puppet_agent/tasks/install_powershell.ps1"}, {"name"=>"facts/tasks/bash.sh", "path"=>"C:/Source/puppet-strings/spec/fixtures/modules/facts/tasks/bash.sh"}], metadata={"description"=>"Install the Puppet agent package", "parameters"=>{"version"=>{"description"=>"The version of puppet-agent to install", "type"=>"Optional[String]"}, "collection"=>{"description"=>"The Puppet collection to install from (defaults to puppet, which maps to the latest collection released)", "type"=>"Optional[Enum[puppet5, puppet6, puppet, puppet5-nightly, puppet6-nightly, puppet-nightly]]"}, "yum_source"=>{"description"=>"The source location to find yum repos (defaults to yum.puppet.com)", "type"=>"Optional[String]"}, "apt_source"=>{"description"=>"The source location to find apt repos (defaults to apt.puppet.com)", "type"=>"Optional[String]"}, "mac_source"=>{"description"=>"The source location to find mac packages (defaults to downloads.puppet.com)", "type"=>"Optional[String]"}, "windows_source"=>{"description"=>"The source location to find windows packages (defaults to downloads.puppet.com)", "type"=>"Optional[String]"}, "install_options"=>{"description"=>"optional install arguments to the windows installer (defaults to REINSTALLMODE=\"amus\")", "type"=>"Optional[String]"}}, "implementations"=>[{"name"=>"install_shell.sh", "requirements"=>["shell"], "files"=>["facts/tasks/bash.sh"], "input_method"=>"environment"}, {"name"=>"install_powershell.ps1", "requirements"=>["powershell"]}]}>' on 95c42917e56e53ec5d22a8a9ec7b14d92618a97c65cfcdacf136b578f9785c39
Started on 95c42917e56e53ec5d22a8a9ec7b14d92618a97c65cfcdacf136b578f9785c39...
[...]

PR #1075, opened 2019-07-05
(GH-1060) Use Docker CLI instead of docker-api gem

glennsarti:
LCOL = Linux Containers on Linux
LCOW via HyperV Isolation = [Link](https://docs.microsoft.com/en-us/virtualization/windowscontainers/deploy-containers/linux-containers#linux-containers-with-hyper-v-isolation)
WCOW = Windows Containers on Windows

This PR is only about being able to use the docker transport, on Windows, with Linux containers. While not explicitly tested with LCOW using Hyper-V isolation (or even LCOW2), this should also support this scenario.

This PR does NOT address Windows containers on Windows (WCOW). That will require a lot of additional work as there are currently hardcoded assumptions in both the code and tests.

~~This is currently not test on Appveyor. Additional work would be needed to see if this is possible.  This PR was focused on making sure that the new implementation still worked for LCOL~~

Update: It is not possible to test this on AppVeyor as the workers do not support LCOW.

---

Previously in BOLT-962 the docker transport was added however the use of the
docker-api gem meant that this could not be used on Windows. In particular
the docker-api gem relies on Unix Domain Sockets for communication with Docker
which are not available on Windows platforms. Without this change Bolt users
would not be able to use the docker transport to either LCOW or Docker Desktop
based linux containers from a Windows Host.

This commit removes the use of the docker-api gem as the number of operations
needed are small enough that the plain docker CLI can be used instead.
* Use `docker ps` and `docker inspect` during the "connection" stage to
  establish that a container exists and is running, and use the continer ID for
  future operations instead of the name.
* Use --interactive when redirecting STDIN
* Standardises on UTF8 output with LF line-endings
* Uses `docker cp` to copy files into the container
* When a user specifies a `service-url`, this is translated into the DOCKER_HOST
  environment variable.  This environment variable is used during the docker CLI
  commands.  The default service-url is platform dependant and is handled by the
  CLI itself.
* Updates tests as some of the error messages needed to change.
* Removed assumptions that the spec tests could only be executed on a
  non-Windows platform.  This allows Windows developers to modify Bolt and be
  able to run the tests locally.

### DIFF
--- a/lib/bolt/transport/docker/connection.rb
+++ b/lib/bolt/transport/docker/connection.rb
@@ -8,21 +8,24 @@ module Bolt
     class Docker < Base
       class Connection
         def initialize(target)
-          # lazy-load expensive gem code
-          require 'docker'
-
           raise Bolt::ValidationError, "Target #{target.name} does not have a host" unless target.host
-
           @target = target
           @logger = Logging.logger[target.host]
+          @docker_host = @target.options['service-url']
         end
 
         def connect
-          # Explicitly create the new Connection to avoid relying on global state in the Docker module.
-          url = @target.options['service-url'] || ::Docker.url
-          options = ::Docker.options.merge(@target.options['service-options'] || {})
-          @container = ::Docker::Container.get(@target.host, {}, ::Docker::Connection.new(url, options))
+          # We don't actually have a connection, but we do need to
+          # check that the container exists and is running.
+          output = execute_local_docker_json_command('ps')
+          index = output.find_index { |item| item["ID"] == @target.host || item["Names"] == @target.host }
+          raise "Could not find a container with name or ID matching '#{@target.host}'" if index.nil?
+          # Now find the indepth container information
+          output = execute_local_docker_json_command('inspect', [output[index]["ID"]])
+          # Store the container information for later
+          @container_info = output[0]
           @logger.debug { "Opened session" }
+          true
         rescue StandardError => e
           raise Bolt::Node::ConnectError.new(
             "Failed to connect to #{@target.uri}: #{e.message}",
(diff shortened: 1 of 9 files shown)


glennsarti:
CI failure is due to the puppet-strings release, and not do to do with this PR

glennsarti:
Bouncing PR to clear unrelated CI failure.

glennsarti:
Those are weird failures in yaml_plan_spec

```
{"kind"=>"bolt/apply-prep",                                                                                                          "msg"=>"puppet_agent::version could not be found",
 "details"=>{}}
```

glennsarti:
... and we're green

beechtom:
Everything's working as expected.

Since `service-options` is no longer needed/used in the config options, should it be removed entirely? Currently both the docs and `docker.rb` reference it.

https://github.com/puppetlabs/bolt/blob/master/pre-docs/bolt_configuration_options.md#docker-transport-configuration-options

https://github.com/puppetlabs/bolt/blob/master/lib/bolt/transport/docker.rb#L25-L29

glennsarti:
@beechtom Yeah, RE `service-options` is interesting.  I could _probably_ be removed. Although removing it is kinda/sorta breaking change?

Should there be a deprecation warning?

beechtom:
Since the Docker transport is experimental breaking changes are expected. Cleaning up `service-options` since it's not being used should be fine.

lib/bolt/transport/docker/connection.rb
```diff
@@ -8,60 +8,85 @@ module Transport
...
+          command_options = []
+          # Need to be interactive if redirecting STDIN
+          command_options << '--interactive' unless options[:stdin].nil?
+          command_options.concat(envs) unless envs.empty?
+          command_options << container_id
+          command_options.concat(command)
+
+          @logger.debug { "Executing: exec #{command_options}" }
+
+          result = execute_local_docker_command('exec', command_options, options[:stdin])
```
Iristyle: Would be better to use destructuring assignment here for readability like is done below:

```ruby
stdout_str, stderr_str, status = execute_local_docker_command('exec', command_options, options[:stdin])
```

lib/bolt/transport/docker/connection.rb
```diff
@@ -8,60 +8,85 @@ module Transport
...
             @logger.info { "Command failed with exit code #{result[2]}" }
           end
-          result[0] = result[0].join.force_encoding('UTF-8')
-          result[1] = result[1].join.force_encoding('UTF-8')
+          result[0] = result[0].force_encoding('UTF-8')
+          result[1] = result[1].force_encoding('UTF-8')
+          # Normalise line endings
+          result[0].gsub!("\r\n", "\n")
+          result[1].gsub!("\r\n", "\n")
           result
```
Iristyle: And then instead of `result`, rebuild the array like

```ruby
[stdout_str, stderr_str, status]
```

lib/bolt/transport/docker/connection.rb
```diff
@@ -8,60 +8,85 @@ module Transport
...
+          # The actual result is the exitstatus not the process object
+          result[2] = result[2].nil? ? -32768 : result[2].exitstatus
           if result[2] == 0
             @logger.debug { "Command returned successfully" }
           else
             @logger.info { "Command failed with exit code #{result[2]}" }
           end
-          result[0] = result[0].join.force_encoding('UTF-8')
-          result[1] = result[1].join.force_encoding('UTF-8')
+          result[0] = result[0].force_encoding('UTF-8')
```
Iristyle: `force_encoding` already mutates the original string, and prefer to use `Encoding::UTF_8` constant

```
irb(main):003:0> a = 'foo'
=> "foo"
irb(main):004:0> a.encoding
=> #<Encoding:UTF-8>
irb(main):005:0> a.force_encoding(Encoding::ASCII)
=> "foo"
irb(main):006:0> a.encoding
=> #<Encoding:US-ASCII>
```

lib/bolt/transport/docker/connection.rb
glennsarti: That'll teach for me copy-pasta the original code :-).  Will do.

glennsarti:
@beechtom  I'll remove `service-options`

One thing to consider:

> Since the Docker transport is experimental breaking changes are expected

That's been experimental for over a year now. Is there a timeframe as to whether this will not be experimental or deleted?

lib/bolt/transport/docker/connection.rb
glennsarti: Done.  Thanks for that!

lib/bolt/transport/docker/connection.rb
glennsarti: Done.

lib/bolt/transport/docker/connection.rb
glennsarti: Done.

glennsarti:
`service-options` has been removed from code and documentation.

PR #1077, opened 2019-07-09
(maint) Update integration test name

glennsarti:
The test name `when running over local` is ambiguous because the test is tagged
with bash.  This commit updates the test name to better reflect what is actually
being tested.

### DIFF
--- a/spec/integration/inventory2_spec.rb
+++ b/spec/integration/inventory2_spec.rb
@@ -324,7 +324,7 @@ describe 'running with an inventory file', reset_puppet_settings: true do
     include_examples 'basic inventory'
   end
 
-  context 'when running over remote', bash: true do
+  context 'when running over remote with bash shell', bash: true do
     let(:inventory) do
       { version: 2,
         targets: [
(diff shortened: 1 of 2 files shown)


lucywyman:
Hey @glennsarti - if you're interested, I think having these work for powershell as well would be pretty easy. For the `remote` one you'd just have to add a super small powershell version of the task to https://github.com/puppetlabs/bolt/tree/master/spec/fixtures/modules/remote/tasks. The local one I think would need it's own powershell-version of the test, but that's also just copy-paste and change the `shell_cmd`. I'm not sure if you have time to make those changes, or care about them, but I think having these run on Windows as well would be good. Otherwise I'll just merge - this is definitely an improvement!

glennsarti:
> would be pretty easy. 
So far in my experience, unpicking the bash assumptions can be a little tricky so I would definitely need to be a separate PR and work item

> I'm not sure if you have time to make those changes, or care about them,
I don't have time right now (I'm working on LCOW and WCOW support for bolt already_, but I **always**  care about test coverage on Windows platforms

PR #1079, opened 2019-07-09
(maint) Document actual default SSH username

nicklewis:


lucywyman:
Ping @MFredette - how do we get this into production?

PR #1081, opened 2019-07-10
(maint) Use 64bit Ruby for testing

glennsarti:
Previously Appveyor testing was only using the 32bit runtime, however as bolt
is 64bit only on Windows, this should really be 64bit.  This commit updates the
RUBY_VERSION to express that it should be 64bit.


glennsarti:
@Iristyle Light is green, trap is clean.

PR #1087, opened 2019-07-12
(BOLT-126) Add WinRM Kerberos support from Linux

Iristyle:
* #999, except without the not-quite-working-yet Docker container setup!
* supersedes #950 

 - This PR is considered experimental due to a number of current
   limitations:

   * Works only with MIT Kerberos from a Linux node
   * Does not work with Heimdal on OSX
     - gssapi gem support for Heimdal is not well vetted
     - OSX doesn't export Kerberos IOV functions needed for MS DCE RPC
   * Does not work from a Windows node as winrm / gssapi gems only
     support MIT Kerberos, and Windows has its own APIs
   * Has been manually tested in a simple AD environment that has a
     CentOS host domain joined to Windows Active Directory

   Note that Kerberos is an authentication method, not a transport, so
   can be used with or without SSL just like other authentication.

 - Automated tests have been scaffolded, to be completed in a future PR
   that enables a Kerberos authentication test setup through Docker

- Manual verification includes the following steps:

   * Acquire ticket from KDC (in this case Active Directory)

```
[centos@bolt-dev-centos7 bolt]$ echo '*****!' | kinit -C Administrator@bolt.puppet
Password for Administrator@bolt.puppet:

[centos@bolt-dev-centos7 bolt]$ klist
Ticket cache: KEYRING:persistent:1000:1000
Default principal: Administrator@BOLT.PUPPET

Valid starting       Expires              Service principal
07/12/2019 18:18:04  07/13/2019 04:18:04  krbtgt/BOLT.PUPPET@BOLT.PUPPET
	renew until 07/19/2019 18:18:04
```

   * Show that bolt.yaml contains the default winrm realm configuration

```
[centos@bolt-dev-centos7 bolt]$ cat ~/.puppetlabs/bolt/bolt.yaml
---
winrm:
  realm: BOLT.PUPPET
```

   * Ask Bolt to run a command over WinRM over HTTPS using the AD domain as realm

```
[centos@bolt-dev-centos7 bolt]$ bundle exec bolt command run 'whoami' --targets winrm://bolt-dc.bolt.puppet --no-ssl-verify
Started on bolt-dc.bolt.puppet...
Finished on bolt-dc.bolt.puppet:
  STDOUT:
    bolt\administrator
Successful on 1 node: winrm://bolt-dc.bolt.puppet
Ran on 1 node in 0.82 seconds
```

   * Verify Kerberos tickets (note the new `HTTP` service ticket)

```
[centos@bolt-dev-centos7 bolt]$ klist
Ticket cache: KEYRING:persistent:1000:1000
Default principal: Administrator@BOLT.PUPPET

Valid starting       Expires              Service principal
07/12/2019 18:18:20  07/13/2019 04:18:04  HTTP/bolt-dc.bolt.puppet@BOLT.PUPPET
	renew until 07/19/2019 18:18:04
07/12/2019 18:18:04  07/13/2019 04:18:04  krbtgt/BOLT.PUPPET@BOLT.PUPPET
	renew until 07/19/2019 18:18:04
```

   * Ask Bolt to run a command over WinRM over HTTP using the AD domain as realm

```
[centos@bolt-dev-centos7 bolt]$ bundle exec bolt command run 'whoami' --targets winrm://bolt-dc.bolt.puppet --no-ssl
Started on bolt-dc.bolt.puppet...
Finished on bolt-dc.bolt.puppet:
  STDOUT:
    bolt\administrator
Successful on 1 node: winrm://bolt-dc.bolt.puppet
Ran on 1 node in 0.74 seconds
```

   * Show that bolt.yaml contains an invalid realm

```
[centos@bolt-dev-centos7 bolt]$ cat ~/.puppetlabs/bolt/bolt.yaml
---
winrm:
  realm: foo
```

   * Verify an incorrectly specified realm fails to connnect over HTTP

```
[centos@bolt-dev-centos7 bolt]$ bundle exec bolt command run 'whoami' --targets winrm://bolt-dc.bolt.puppet --no-ssl
Started on bolt-dc.bolt.puppet...
Failed on bolt-dc.bolt.puppet:
  Failed to connect to http://bolt-dc.bolt.puppet:5985/wsman: gss_init_sec_context did not return GSS_S_COMPLETE: Unspecified GSS failure.  Minor code may provide more information
  Server not found in Kerberos database
Failed on 1 node: winrm://bolt-dc.bolt.puppet
Ran on 1 node in 0.22 seconds
```

   * Verify an incorrectly specified realm fails to connnect over HTTPS

```
[centos@bolt-dev-centos7 bolt]$ bundle exec bolt command run 'whoami' --targets winrm://bolt-dc.bolt.puppet --no-ssl-verify
Started on bolt-dc.bolt.puppet...
Failed on bolt-dc.bolt.puppet:
  Failed to connect to bolt-dc.bolt.puppet:5986/wsman: gss_init_sec_context did not return GSS_S_COMPLETE: Unspecified GSS failure.  Minor code may provide more information
  Server not found in Kerberos database
Failed on 1 node: winrm://bolt-dc.bolt.puppet
Ran on 1 node in 0.16 seconds
```

 - This PR also relies on version 2.3.2 of the winrm gem which includes
   a critical Kerberos encryption bug that was fixed in
   WinRb/WinRM#302

 - There is a separate PR which will add a Docker container setup to
   verify Kerberos in an automated fashion against TravisCI using
   Samba and OMI server. At present, there is a problem properly
   establishing a connection between the WinRM gem and OMI server, which
   appears to be a bug in the WinRM gem protocol handling.

 - Windows support is https://tickets.puppet.com/browse/BOLT-1323

 - OSX support is https://tickets.puppet.com/browse/BOLT-1471

 - Includes basic experimental docs

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -20,24 +20,25 @@ desc "Run RSpec tests that don't require VM fixtures or a particular shell"
 RSpec::Core::RakeTask.new(:unit) do |t|
   t.rspec_opts = '--tag ~ssh --tag ~docker --tag ~bash --tag ~winrm ' \
                  '--tag ~appveyor_agents --tag ~puppetserver --tag ~puppetdb ' \
-                 '--tag ~omi'
+                 '--tag ~omi --tag ~kerberos'
 end
 
 desc "Run RSpec tests for AppVeyor that don't require SSH, Bash, Appveyor Puppet Agents, or orchestrator"
 RSpec::Core::RakeTask.new(:appveyor) do |t|
   t.rspec_opts = '--tag ~ssh --tag ~docker --tag ~bash --tag ~appveyor_agents ' \
-         '--tag ~orchestrator --tag ~puppetserver --tag ~puppetdb --tag ~omi'
+         '--tag ~orchestrator --tag ~puppetserver --tag ~puppetdb --tag ~omi ' \
+         '--tag ~kerberos'
 end
 
 desc "Run RSpec tests for TravisCI that don't require WinRM"
 RSpec::Core::RakeTask.new(:travisci) do |t|
   t.rspec_opts = '--tag ~winrm --tag ~appveyor_agents --tag ~puppetserver --tag ~puppetdb ' \
-  '--tag ~omi --tag ~windows'
+  '--tag ~omi --tag ~windows --tag ~kerberos'
 end
 
 desc "Run RSpec tests that require slow to start puppet containers"
 RSpec::Core::RakeTask.new(:puppetserver) do |t|
-  t.rspec_opts = '--tag puppetserver --tag puppetdb --tag omi'
+  t.rspec_opts = '--tag puppetserver --tag puppetdb --tag omi --tag kerberos'
 end
 
 RuboCop::RakeTask.new(:rubocop) do |t|
--- a/developer-docs/kerberos.md
+++ b/developer-docs/kerberos.md
@@ -1,0 +1,80 @@
+# Bolt + Kerberos
+
+### Overview
+
+Bolt supports using Kerberos as an authentication mechanism, in lieu of a username / password. An excellent background on key Kerberos concepts is available in [Designing an Authentication System: a Dialogue in Four Scenes](https://web.mit.edu/kerberos/www/dialogue.html).
+
+While Linux **does** support a standalone Kerberos KDC (Key Distribution Center), the [`winrm`](https://github.com/WinRb/WinRM) gem that provides connectivity support from Bolt has only been written for and tested against Active Directory.
+
+In the future, it's possible other transports like SSH may be able to authenticate and authorize with just a KDC, but for now, the primary use case for Kerberos is in conjunction with WinRM.
+
+### Kerberos Implementations
+
+There are three primary implementations of the Kerberos protocol widely available:
+
+* [Heimdal](#heimdal)
+* [MIT Kerberos](#mit-kerberos)
+* [Microsoft Kerberos](#microsoft-kerberos)
+
+#### Heimdal
+
+[Heimdal](https://www.h5l.org/) is an open source implementation that is shipped on Mac OSX and is the default library used for the Samba server packages on Linux.
+
+Even though Heimdal provided interoperability with [Microsoft DCE/RPC](https://en.wikipedia.org/wiki/DCE/RPC) first with the addition of [IOV message wrapping extension functions](https://web.mit.edu/kerberos/krb5-latest/doc/appdev/gssapi.html#iov-message-wrapping), these APIs are not exported in the OSX libraries that ship with the operating system.
+
+The [gssapi](https://github.com/zenchild/gssapi) gem that Bolt relies on for [Generic Security Services](https://en.wikipedia.org/wiki/Generic_Security_Services_Application_Program_Interface) defaults to MIT Kerberos, but also supports Heimdal via a programmatic opt-in by adding `require 'gssapi/heimdal'` prior to `require 'gssapi'` (for instance, to the code in `winrm/connection.rb`). Unfortunately the support for Heimdal is not well tested, the paths to the Heimdal libraries are currently hard-coded, and the library doesn't account for semantic differences in the Heimdal and MIT implementations of Kerberos APIs, which can cause segfaults.
+
+Samba also has optional experimental support for compiling against MIT Kerberos instead of the default Heimdal.
+
+#### MIT Kerberos
+
+[MIT Kerberos](https://web.mit.edu/kerberos/) is the default implementation used in Linux, and there is some general information on the [client installation instructions page](https://web.mit.edu/Kerberos/www/krb5-latest/doc/admin/install_clients.html).
+
+##### RHEL
+
+RHEL typically installs with the following packages
+
+> yum install krb5-workstation krb5-libs krb5-auth-dialog
+
+RedHat maintains docs about [Configuring a Kerberos 5 client](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/6/html/managing_smart_cards/configuring_a_kerberos_5_client)
+
+##### Ubuntu
+
+Ubuntu typically installs with the following packages
+
+> apt-get install krb5-user libpam-krb5 libpam-ccreds auth-client-config
+
+Canonical maintains docs about running a [Kerberos Linux Client](https://help.ubuntu.com/lts/serverguide/kerberos.html#kerberos-linux-client)
+
+#### Microsoft Kerberos
+
+Microsoft Kerberos is built-in to Windows and most easily consumed transparently by domain joining computers to Active Directory.
+
+Microsoft provides some guidance around [SSPI/Kerberos Interopability with GSSAPI](https://docs.microsoft.com/en-us/windows/win32/secauthn/sspi-kerberos-interoperability-with-gssapi) that covers how to map gssapi calls to their equivalent Windows APIs.
+
+### Kerberos Tooling
+
+On Linux and OSX, regardless of the package in use, client configuration is typically stored in `/etc/krb5.conf` (overridable with the `KRB5_CONFIG` environment variable). Windows Active Directory does not use such a config file.
+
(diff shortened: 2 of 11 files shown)


.travis.yml
```diff
@@ -57,7 +57,11 @@ script:
   fi
 - bundle exec rake puppetserver
 after_script:
-- docker-compose -f spec/docker-compose.yml logs
+- |
+  sync
+  export line_sep=$(printf '%*s' 80 | tr ' ' '#')
+  docker ps --all -q | xargs -L 1 -I % sh -c 'echo "\n${line_sep}\nContainer Logs - $(docker inspect % --format={{.Name}})\n${line_sep}\n"; docker logs %'
```
Iristyle: The downside is no colorized output... the upside is that the logs are no longer streamed together, hence readable!

Rakefile
```diff
@@ -37,7 +37,7 @@ end
 
 desc "Run RSpec tests that require slow to start puppet containers"
 RSpec::Core::RakeTask.new(:puppetserver) do |t|
-  t.rspec_opts = '--tag puppetserver --tag puppetdb --tag omi'
+  t.rspec_opts = '--tag puppetserver --tag puppetdb --tag omi --tag kerberos'
```
lucywyman: I think this also needs `--tag ~kerberos` added to the `unit`, `travisci`, and `appveyor` rake tasks

.travis.yml
```diff
@@ -57,7 +57,11 @@ script:
...
+  sleep 3s
```
lucywyman: What's the sleep for?

spec/docker-compose.yml
```diff
@@ -56,5 +56,5 @@ services:
       args:
         BOLT_PASSWORD: bolt
     ports:
-      - "24985:5985"
-      - "24986:5986"
+      - "45985:5985"
```
lucywyman: Purely out of curiosity is there a reason for changing these ports?

.travis.yml
Iristyle: See commit message at https://github.com/puppetlabs/bolt/pull/1087/commits/ac356d9ceb1fa56beecf1529a83ee3789e240267

spec/docker-compose.yml
Iristyle: Noted in commit message at https://github.com/puppetlabs/bolt/pull/1087/commits/c4c8a2783fd7831de9a0edfe3a6339816f8b5449 ... its to keep synced up around 5985 / 5986, which are the winrm ports you'd typically be grepping for.

Rakefile
Iristyle: Thank you - good catch! Totally missed in the logs that those tests were running multiple times in Travis. Updating!

Iristyle: And now that AppVeyor is running again, on the repush we'll get validation that tests aren't running in that environment either.

.travis.yml
lucywyman: Do we plan to leave these in, or was it just for debugging this PR?

lucywyman: It looks like it only takes 3.6 seconds, so that seems fine to leave in as long as it doesn't fail and cause travis to fail.

developer-docs/kerberos.md
```diff
@@ -0,0 +1,73 @@
+# Bolt + Kerberos
+
+### Overview
+
+Bolt supports using Kerberos as an authentication mechanism, in lieu of a username / password. An excellent background on key Kerberos concepts is available in [Designing an Authentication System: a Dialogue in Four Scenes](https://web.mit.edu/kerberos/www/dialogue.html).
+
+While Linux **does** support a standalone Kerberos KDC (Key Distribution Center), the [`winrm`](https://github.com/WinRb/WinRM) gem that provides connectivity support from Bolt has really only been written for / tested against Active Directory.
```
```suggestion
While Linux **does** support a standalone Kerberos KDC (Key Distribution Center), the [`winrm`](https://github.com/WinRb/WinRM) gem that provides connectivity support from Bolt has only been written for and tested against Active Directory.
```

developer-docs/kerberos.md
```diff
@@ -0,0 +1,73 @@
...
+
+#### Heimdal
+
+[Heimdal](https://www.h5l.org/) is an open source implementation that is shipped on Mac OSX and is the default library used for the Samba server packages on Linux.
+
+Even though Heimdal provided interoperability with [Microsoft DCE/RPC](https://en.wikipedia.org/wiki/DCE/RPC) first with the addition of [IOV message wrapping extension functions](https://web.mit.edu/kerberos/krb5-latest/doc/appdev/gssapi.html#iov-message-wrapping), these APIs are not exported in the OSX libraries that ship with the operating system.
+
+The [gssapi](https://github.com/zenchild/gssapi) gem that Bolt relies on for [Generic Security Services](https://en.wikipedia.org/wiki/Generic_Security_Services_Application_Program_Interface) defaults to MIT Kerberos, but also supports Heimdal via a programmatic opt-in by adding `require 'gssapi/heimdal'` prior to `require 'gssapi'` (for instance, to the code in `winrm/connection.rb`). Unfortunately the support for Heimdal is not well tested, the paths to the Heimdal libraries are currently hard-coded, and the library doesn't account for semantic differences in the Heimdal and MIT implementations of Kerberos APIs, which can cause segfaults.
```
lucywyman: Is the TL;DR here that Heimdal isn't supported?

developer-docs/kerberos.md
```diff
@@ -0,0 +1,73 @@
...
+* `kinit` for acquiring tickets (not used on Windows)
+* `klist` for listing tickets
+* `kdestroy` for destroying tickets (not used on Windows)
+
+### Usage
+
+Manual verification of Bolt can be performed from a Linux node that is domain joined to Active Directory using the following steps:
+
+- `kinit -C Administrator@domain.com` to acquire a TGT (ticket granting ticket)
+- `bolt command run 'whoami' --targets winrm://dc.domain.com --realm DOMAIN.COM --no-ssl-verify` to connect over HTTPS
```
lucywyman: Is `--no-ssl-verify` necessary? If not I'd remove it here.

developer-docs/kerberos.md
Iristyle: It's a bit subtle and I tried explaining in as few words as possible.

`gssapi` supports Heimdal... under specific circumstances.

But Bolt doesn't opt-in under the circumstances where it might work anyhow (though could if we wanted to test it, etc).

developer-docs/kerberos.md
Iristyle: It's likely this could be needed on Linux, but shouldn't be needed when the client is Windows.

I may note that it's optional based on environment.

Iristyle:
The commit to add `--realm` flag back is at https://github.com/Iristyle/bolt/commit/34d664171e5ab6e657ffad64413552800efc4c42

developer-docs/kerberos.md
lucywyman: I see so you could patch Bolt to work with heimdal? But not use it out of the box?

Iristyle: Heimdal is a cross-platform library... and `gssapi` loads it differently based on platform:
https://github.com/zenchild/gssapi/blob/826c91b8481407b0db36cb97fe88f49ac3e21315/lib/gssapi/lib_gssapi_loader.rb#L46-L57

I only tested on OSX... to even get it to load, I had to:

* put a `require 'gssapi/heimdal'` before other `gssapi` requires in the Bolt source
* patch the above load path (since I was trying to use system Heimdal, not a separately installed Heimdal)
* patch parts of the `gssapi` gem itself to prevent two different double-free segfaults

Even then, OSX system Heimdal doesn't export the `IOV` methods we care about... so it was a dead end basically.

I didn't try:

* using Heimdal on Linux
* using a different Heimdal binary on OSX

Plausibly those work without patches, but I didn't try because:

* Linux MIT Kerberos is the most prevalent (except inside Samba server, go figure)
* I filed a separate ticket to figure out how we'll address the OSX situation... we could compile + ship MIT or Heimdal binaries with the OSX packages, for instance

PR #1088, opened 2019-07-15
(BOLT-1315) Collect analytics about Boltdir usage

beechtom:
Adds a custom dimension to the screen view for Boltdir usage. When the Boltdir is set up, it will log its type as either Boltdir, bolt.yaml, the default directory, missing default directory, or a Boltdir specified as a command line option.

### DIFF
--- a/lib/bolt/analytics.rb
+++ b/lib/bolt/analytics.rb
@@ -21,7 +21,8 @@ module Bolt
       statement_count: :cd6,
       resource_mean: :cd7,
       plan_steps: :cd8,
-      return_type: :cd9
+      return_type: :cd9,
+      boltdir_type: :cd11
     }.freeze
 
     def self.build_client
--- a/lib/bolt/boltdir.rb
+++ b/lib/bolt/boltdir.rb
@@ -6,10 +6,10 @@ module Bolt
   class Boltdir
     BOLTDIR_NAME = 'Boltdir'
 
-    attr_reader :path, :config_file, :inventory_file, :modulepath, :hiera_config, :puppetfile, :rerunfile
+    attr_reader :path, :config_file, :inventory_file, :modulepath, :hiera_config, :puppetfile, :rerunfile, :type
 
     def self.default_boltdir
-      Boltdir.new(File.join('~', '.puppetlabs', 'bolt'))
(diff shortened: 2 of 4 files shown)


beechtom:
Still need to write tests - I'm thinking boltdir_specs might be the best location to make sure the type was set correctly?

lib/bolt/boltdir.rb
```diff
@@ -37,6 +38,17 @@ def initialize(path)
       @hiera_config = @path + 'hiera.yaml'
       @puppetfile = @path + 'Puppetfile'
       @rerunfile = @path + '.rerun.json'
+      @type = compute_type
+    end
+
+    def compute_type
```
adreyer: This misses the case where `--boltdir` was passed. 

Instead of recomputing this I think `find_boltdir` or whatever creates the instance should pass the "type" as an argument.

beechtom: Updated to pass the type to the initialize method.

lib/bolt/analytics.rb
```diff
@@ -21,7 +21,8 @@ module Analytics
       statement_count: :cd6,
       resource_mean: :cd7,
       plan_steps: :cd8,
-      return_type: :cd9
+      return_type: :cd9,
+      boltdir_type: :cd10
```
adreyer: I already set up cd10 for "inventory_version in google analytics. I've added cd11 for Boltdir Type now

lib/bolt/boltdir.rb
```diff
@@ -6,10 +6,10 @@ module Bolt
   class Boltdir
     BOLTDIR_NAME = 'Boltdir'
 
-    attr_reader :path, :config_file, :inventory_file, :modulepath, :hiera_config, :puppetfile, :rerunfile
+    attr_reader :path, :config_file, :inventory_file, :modulepath, :hiera_config, :puppetfile, :rerunfile, :type
 
     def self.default_boltdir
-      Boltdir.new(File.join('~', '.puppetlabs', 'bolt'))
+      Boltdir.new(File.join('~', '.puppetlabs', 'bolt'), 'default')
```
adreyer: Lets call this "homedir" or "user" so we can change it.

beechtom: Updated and set type to `user`.

PR #1089, opened 2019-07-15
(fix) Change read timeout on docker transport to large number

hajee:
The default timeout of a command using a docker transport is 60 seconds. See this example:

```
$ bolt command run 'sleep 300' --targets a8ad33ff68069d48fdacc80172c8e4e79214f55523aaab143e1a7238e670a73a -i inventory.yaml
Started on a8ad33ff68069d48fdacc80172c8e4e79214f55523aaab143e1a7238e670a73a...
Failed on a8ad33ff68069d48fdacc80172c8e4e79214f55523aaab143e1a7238e670a73a:
  read timeout reached
Failed on 1 node: a8ad33ff68069d48fdacc80172c8e4e79214f55523aaab143e1a7238e670a73a
Ran on 1 node in 62.26 seconds
```

This is to short for some longer running commands (Like a puppet apply) The timeout seems ro be default value provided by the docker-api gem. You can override this value .

This PR allows uses a large number (24 hours) to override this default timeout.

With this change, the example looks like this:

```
 $ bolt command run 'sleep 300' --targets a8ad33ff68069d48fdacc80172c8e4e79214f55523aaab143e1a7238e670a73a -i inventory.yaml
Started on a8ad33ff68069d48fdacc80172c8e4e79214f55523aaab143e1a7238e670a73a...
Finished on a8ad33ff68069d48fdacc80172c8e4e79214f55523aaab143e1a7238e670a73a:
Successful on 1 node: a8ad33ff68069d48fdacc80172c8e4e79214f55523aaab143e1a7238e670a73a
Ran on 1 node in 299.95 seconds
```

### DIFF
--- a/lib/bolt/transport/docker.rb
+++ b/lib/bolt/transport/docker.rb
@@ -60,6 +60,7 @@ module Bolt
         if target.options['tty']
           options[:Tty] = true
         end
+        options[:wait] = 24 * 60 * 60  # Choose 24 hour as timeout. But needs to be big
         if target.options['shell-command'] && !target.options['shell-command'].empty?
           # escape any double quotes in command
           command = command.gsub('"', '\"')


lucywyman:
Hello! Thanks for submitting this. I'm curious about why the docker operation is timing out, and if there's a better way than mis-using the connection timeout setting. I plan to noodle on it and talk it over with the team tomorrow, but this seems like a good workaround if we can't think of a better solution!

Expect to hear from us by about this time tomorrow!

tphoney:
@lucywyman from speaking with @hajee docker instance that is being connected to is under high load. So the connection is made successfully, but the command takes a long time to run. From memory it is timing out trying to install a puppet agent, using the puppetagent module.

hajee:
The issue is more likely to occur on high loaded systems, but we can also have commands that always take longer than a minute and so always will timeout. For example when creating a database node.  This is also demonstrated by the sleep command. 

Hope this clarification helps.

hajee:
An other option would be using the `service-options` in the inventory, like this:

```
      docker:
        shell-command: bash -lc
        service-options:
          wait: 300
```
Then the code to use this would become:

```ruby
      def run_command(target, command, options = {})
        options[:wait] = target.options.dig('service-options','wait')
```

adreyer:
#1069 is switching the docker transport to shell out to the docker cli. I'm not sure how that interacts with this.

hajee:
@adreyer I guess it doesn't work with that PR. It works by providing an option to the docker-api. So if you remove that gem, this will surely not work. 

The question is, if the new solution has a time limit for running commands and if we need/can adjust this time.

nicklewis:
We don't support timeouts for other transports, which in practice means we will wait indefinitely for the operation to complete. It seems reasonable to replicate that behavior in this case, which I suppose we would achieve by hardcoding an arbitrarily _long_ timeout. Then if we _do_ add an operation-level timeout (as opposed to our current connection timeouts), we can add it to each of the transports at once.

hajee:
I think is best to change the current PR to implement the arbitrarily long timeout and remove any influence you have over the setting through cli options or yaml file. You all agree?

nicklewis:
This is great, thank you for the PR!

PR #1090, opened 2019-07-17
(maint) Fixes from puppet-strings 2.3.0 release

beechtom:
- Unpin yard gem from 0.9.19
Previously, yard was pinned due to changes in how dynamic symbols are handled by yard 0.9.20, causing formatting issues for plan functions such as out::message. The puppet-strings gem was updated to fix this formatting issue.

- Delete data_types field from generated docs
As of puppet-strings 2.3.0, an additional field is generated for the docs, causing the docs rake task to fail. This commit deletes the data_types field.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -68,6 +68,7 @@ task :docs do
                                      'bolt-modules/system'])
   json = JSON.parse(File.read(tmpfile))
   funcs = json.delete('puppet_functions')
+  json.delete('data_types')
   json.each { |k, v| raise "Expected #{k} to be empty, found #{v}" unless v.empty? }
 
   # @functions will be a list of function descriptions, structured as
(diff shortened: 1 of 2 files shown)


Rakefile
```diff
@@ -68,6 +68,8 @@ task :docs do
                                      'bolt-modules/system'])
   json = JSON.parse(File.read(tmpfile))
   funcs = json.delete('puppet_functions')
+  # TODO: With puppet-strings >= 2.3.0 data type docs are generated - do we want/need to do anything with this?
```
Iristyle: Seems reasonable for now to me.

adreyer: This is fine for now. We do have reference docs for the datatypes I think we should ticket switching those out for the strings docs separately.

PR #1093, opened 2019-07-19
(BOLT-1475) Kerberos Development Container

Iristyle:
Builds on top of #999
Blocks #1101 

This adds a development environment useful for testing / debugging *any* problem between the winrm / gssapi gem and OMI server. 

Note that the final commit currently clones source from https://github.com/Iristyle/omi/tree/debugging-winrm-gem-kerberos which has extra debug / trace statements.

There are at least 2 known problems for interop between the winrm gem and OMI server that this should help diagnosing:

* [BOLT-1476](https://tickets.puppetlabs.com/browse/BOLT-1476) Kerberos auth fails somewhere after OMI server authorizes the connection. Then the next payload OMI returns is encrypted, but cannot be decrypted by the winrm gem. It's unclear if the wrong cipher is being used or if there is another problem?

* There are problems running tasks + higher level Bolt constructs against OMI (basic commands generally work)

* OMI server, even against the PowerShell 6 client, seems to close connections unexpectedly

### DIFF
--- a/developer-docs/kerberos.md
+++ b/developer-docs/kerberos.md
@@ -214,3 +214,34 @@ Once DNS and the Kerberos client are properly configured, `kinit` can be used to
 To remove the ticket, use:
 
 > kdestroy --credential=krbtgt/BOLT.TEST@BOLT.TEST
+
+### Advanced: Debugging Bolt + OMI
+
+#### Building OMI server from source
+
+When debugging interoperability problems between the winrm gem and OMI server, it may be necessary to build OMI server from source, rather than consuming packages. This makes it easy to modify OMI server code, rebuild and start using it immediately. When starting the containers, add the build arg `BUILD_OMI=true` like:
+
+> docker-compose -f spec/docker-compose.yml build --build-arg BUILD_OMI=true samba-ad omiserver
+> docker-compose -f spec/docker-compose.yml up -d --build samba-ad omiserver
+
+This will:
+
+* install vim
+* install all necessary dev / build tooling
+* clone source from https://github.com/Microsoft/omi to `/tmp/omi`
+* write the script `/build-omi.sh` inside the container
+* increase sssd log output
+* tail sssd logs and OMI messages in addition to OMI logs
+
+##### `/build-omi.sh`
+
+Run this script at any point after making changes to the source in `/tmp/omi` to build and redeploy OMI server. This will also make sure that the OMI `loglevel` config setting is set to `VERBOSE`
+
+##### OMI source notes
+
+In some cases, OMI server source has to be modified to increase log output, in addition to the `loglevel` change thats already been made. One such example is [Unix/sock/sock.c](https://github.com/microsoft/omi/blob/master/Unix/sock/sock.c#L39-L41), where values must be uncommented like:
+
+```
+# define ENABLE_TRACING 1
+# define TRACING_LEVEL 4
+```
--- a/developer-docs/kerberos.md
+++ b/developer-docs/kerberos.md
@@ -245,3 +245,21 @@ In some cases, OMI server source has to be modified to increase log output, in a
 # define ENABLE_TRACING 1
 # define TRACING_LEVEL 4
 ```
+
+#### Bolt Development Environment
+
+In addition to building OMI from source, it can be useful to run the Bolt source from a Linux agent to iterate on Bolt itself. To really vet Kerberos, this should be performed on a separate container image from the Samba Active Directory controller or the OMI server. This new dev container is defined in `spec/docker-compose-dev.yml` and can be built in addition to the other relevant containers by running `docker-compose`:
+
+> docker-compose -f spec/docker-compose.yml -f spec/docker-compose-dev.yml build --build-arg BUILD_OMI=true samba-ad omiserver linuxdev
+> docker-compose -f spec/docker-compose.yml -f spec/docker-compose-dev.yml up samba-ad omiserver linuxdev
+
+This will:
+
+* Provision Ubuntu 18.04
+* Join the Samba domain just like OMI server
+* Verify pwsh can use all authentication mechanisms against OMI
+* Expose port 422 on the Docker host for SSH access
+* Install vim, git, rbenv / ruby 2.55 / bundler
+* Clone the Bolt source to ~/bolt
+* Install all gems with bundler including pry-byebug and pry-stackexplorer
+* Provide a test script `/bolt-kerberos-test.sh` that can be used for simple test reproductions using Kerberos, noting that some failures that result are currently expected given incompatibilities between winrm gem and OMI server
--- a/spec/Dockerfile.linuxdev
+++ b/spec/Dockerfile.linuxdev
@@ -1,0 +1,86 @@
+FROM ubuntu:18.04
+
+ARG BOLT_PASSWORD=bolt
+ARG RUBY_VERSION=2.5.5
+# NOTE: Only designed to be set at build time, not at run time!
+ENV BOLT_PASSWORD=${BOLT_PASSWORD}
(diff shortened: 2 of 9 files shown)


spec/Dockerfile.linuxdev
```diff
@@ -0,0 +1,91 @@
...
+EXPOSE 22
+
+ADD fixtures/linuxdev/docker-entrypoint.sh /
+ADD fixtures/linuxdev/bolt-kerberos-test.sh /
+ADD fixtures/samba-ad/kerberos-client-config.sh /
+ADD fixtures/samba-ad/krb5.conf.tmpl /
+ADD fixtures/omiserver/domain-join.sh /
+ADD fixtures/omiserver/realmd.conf.tmpl /
+ADD fixtures/omiserver/smb.conf.tmpl /
+ADD fixtures/omiserver/user.map.tmpl /
```
lucywyman: This doesn't seem to exist? I'm not sure if it's necessary or not - I commented it out locally.

Iristyle:
Ohhh thanks! That file belongs in a commit in the final PR that we're not
merging.

Will fix up shortly.

Iristyle:
@donoghuc had mentioned that it might be good to move the dev files elsewhere as there is already a pattern around where dev containers go.

But I don't recall what the recommendation was since I only see another Dockerfile / docker-compose in the root of the project.

spec/Dockerfile.linuxdev
```diff
@@ -0,0 +1,90 @@
...
+    git \
+    ssh \
+    dnsutils \
+    gettext-base \
+    gss-ntlmssp \
+    libgssapi-krb5-2 \
+    ntp \
+    krb5-config \
+    krb5-user \
+    # TODO: not sure which of these 3 are needed for SSH + Kerberos
```
Iristyle: Aw boo I left a TODO in here... I was originally going to setup SSH to support Kerberos, but didn't get to doing that part. And I don't think this PR should be held up on that aspiration right now.

developer-docs/kerberos.md
```diff
@@ -214,3 +214,52 @@ Once DNS and the Kerberos client are properly configured, `kinit` can be used to
...
 > kdestroy --credential=krbtgt/BOLT.TEST@BOLT.TEST
+
+### Advanced: Debugging Bolt + OMI
+
+#### Building OMI server from source
+
+When debugging interoperability problems between the winrm gem and OMI server, it may be necessary to build OMI server from source, rather than consuming packages. This makes it easy to modify OMI server code, rebuild and start using it immediately. When starting the containers, add the build arg `BUILD_OMI=true` like:
+
+> docker-compose -f spec/docker-compose.yml build --build-arg BUILD_OMI=true samba-ad omiserver
+> docker-compose -f spec/docker-compose.yml up --build samba-ad omiserver
```
lucywyman: Is having this second line intentional?

developer-docs/kerberos.md
```diff
@@ -214,3 +214,52 @@ Once DNS and the Kerberos client are properly configured, `kinit` can be used to
...
+##### `/build-omi.sh`
+
+Run this script at any point after making changes to the source in `/tmp/omi` to build and redeploy OMI server. This will also make sure that the OMI `loglevel` config setting is set to `VERBOSE`
+
+##### OMI source notes
+
+In some cases, OMI server source has to be modified to increase log output, in addition to the `loglevel` change thats already been made. One such example is [Unix/sock/sock.c](https://github.com/microsoft/omi/blob/master/Unix/sock/sock.c#L39-L41, where values must be uncommented like:
+
+```
+#define ENABLE_TRACING 1
```
```suggestion
# define ENABLE_TRACING 1
```

developer-docs/kerberos.md
```diff
@@ -214,3 +214,52 @@ Once DNS and the Kerberos client are properly configured, `kinit` can be used to
...
+* increase sssd log output
+* tail sssd logs and OMI messages in addition to OMI logs
+
+##### `/build-omi.sh`
+
+Run this script at any point after making changes to the source in `/tmp/omi` to build and redeploy OMI server. This will also make sure that the OMI `loglevel` config setting is set to `VERBOSE`
+
+##### OMI source notes
+
+In some cases, OMI server source has to be modified to increase log output, in addition to the `loglevel` change thats already been made. One such example is [Unix/sock/sock.c](https://github.com/microsoft/omi/blob/master/Unix/sock/sock.c#L39-L41, where values must be uncommented like:
```
```suggestion
In some cases, OMI server source has to be modified to increase log output, in addition to the `loglevel` change thats already been made. One such example is [Unix/sock/sock.c](https://github.com/microsoft/omi/blob/master/Unix/sock/sock.c#L39-L41), where values must be uncommented like:
```

developer-docs/kerberos.md
lucywyman: I know it's wrong in the source, but it looks weird here :P

developer-docs/kerberos.md
lucywyman: Ok woops I see now! Do we also want to add `-d`?

spec/docker-compose-dev.yml
```diff
@@ -0,0 +1,26 @@
...
+    ports:
+      - "422:22"
+    environment:
+      KRB5_REALM: ${KRB5_REALM:-BOLT.TEST}
+      KRB5_KDC: ${KRB5_KDC:-samba-ad.bolt.test}
+      SMB_DOMAIN: ${SMB_DOMAIN:-BOLT}
+      SMB_ADMIN: Administrator
+      SMB_ADMIN_PASSWORD: ${SMB_ADMIN_PASSWORD:-B0ltrules!}
+    depends_on:
+      - samba-ad
```
lucywyman: I think since this isn't defined in this file it doesn't know about it. Is there a reason not to have this in the main `docker-compose`?

developer-docs/kerberos.md
Iristyle: Yeah, I thought the same.. but opted to copy the original source.

I'm fine either way.

spec/docker-compose-dev.yml
Iristyle: The deal there is that I didn’t want to build that container unless it was explicitly asked for since it’s a dev container.. hence the separate file.

Dockers `depends_on` doesn’t actually guarantee anything useful in terms of orchestrating containers - it only agrees to start `samba-ad` first. Since our instructions already start the other compose file, before starting this one, then that requirement is already met.


IOW, fine to remove it I think.

Iristyle: Actually on second thought... can you verify you actually ran the commands as documented?

> docker-compose -f spec/docker-compose.yml -f spec/docker-compose-dev.yml build --build-arg BUILD_OMI=true samba-ad omiserver linuxdev
> docker-compose -f spec/docker-compose.yml -f spec/docker-compose-dev.yml up samba-ad omiserver linuxdev

Multiple YAML files should get merged prior to running anything.

If you tried to run just `docker-compose-dev.yml` by itself I could see why things might fail.

spec/fixtures/linuxdev/bolt-kerberos-test.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+set -e
+
+mkdir -p ~/.puppetlabs/bolt
+rm ~/.puppetlabs/bolt/bolt.yaml
```
lucywyman: I think this needs to be `rm -f` in case the file doesn't exist.

spec/Dockerfile.linuxdev
Iristyle: Just to clarify on this - I'll remove the unneeded packages and the TODO from here prior to merge. I'll move that as a placeholder into the next PR as a reminder to either:

- investigate / add to that PR
- ditch for now

PR #1098, opened 2019-07-23
 (BOLT-1492) Note behavior change for get_targets 

lucywyman:
Add note that `get_targets` behaves different within apply blocks when
used with version 2 inventory as opposed to version 1. Include docs on
how to convert a plan using the version 2 inventory to version 1
inventory behavior.

### DIFF
--- a/bolt-modules/boltlib/lib/puppet/functions/get_targets.rb
+++ b/bolt-modules/boltlib/lib/puppet/functions/get_targets.rb
@@ -3,6 +3,10 @@
 require 'bolt/error'
 
 # Parses common ways of referring to targets and returns an array of Targets.
+#
+# **NOTE:** Calling `get_targets` inside an `apply` block with a
+# version 2 inventory creates a new Target object.
+# `get_targets('all')` returns an empty array.
 Puppet::Functions.create_function(:get_targets) do
   # @param names A pattern or array of patterns identifying a set of targets.
   # @return A list of unique Targets resolved from any target URIs and groups.
--- a/pre-docs/inventory_version2.md
+++ b/pre-docs/inventory_version2.md
@@ -39,6 +39,43 @@ string to a target defined elsewhere in the inventory. If no target has a name
 or alias matching the string bolt will create a new target with the string as
 it's uri.
 
+### Migrating Plans
+
+In addition to inventory file changes, inventory functions such as `get_targets` might not work as expected when called from an [apply block](applying_manifest_blocks.md) using inventory version 2. `get_targets` returns an empty array when called with `'all'` as an argument. Otherwise, it creates a new Target object for that host. If you want to have the same behavior as the version 1 inventory, you can extract information outside of an apply block into a variable, and use that variable inside the apply block.
+
+For example, the following plan:
+```
+plan setup_lb (
+    TargetSpec $pool,
+    TargetSpec $lb
+) {
+  apply_prep([$pool, $lb])
+
+  apply($lb) {
+    class { 'profile::lb':
+      members => get_targets($pool).map |$targ| { $targ.host }
+    }
+  }
+}
+```
+
+Would need to be converted to:
+```
+plan setup_lb (
+   TargetSpec $pool,
+   TargetSpec $lb
+) {
+   apply_prep([$pool, $lb])
+
+   $members = get_targets($pool).map |$targ| { $targ.host }
+   apply($lb) {
+       class { 'profile::lb':
+          members => $members
+       }
+   }
+}
+```
+
 
 ## Creating a node with a human readable name and ip address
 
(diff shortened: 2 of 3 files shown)


pre-docs/inventory_version2.md
```diff
@@ -39,6 +39,7 @@ string to a target defined elsewhere in the inventory. If no target has a name
 or alias matching the string bolt will create a new target with the string as
 it's uri.
 
+**Note:** In addition to inventory file changes, inventory functions such as `get_targets` cannot be called from an [apply block](applying_manifest_blocks.md) using inventory V2
```
adreyer: I think make this a whole new section. "Updating plans" With an example of extracting information at the plan level into a variable. 

```
plan setup_lb (
   TargetSpec $pool_members,
   TargetSpec $lb
) {
   apply_prep([$pool_members, $lb])

   $member_hosts = get_targets($pool_members).map |$targ| { target.host }
   apply($lb) {
       class { 'profile::lb':
          members => $member_hosts
       }
   }
}
```

adreyer: you could also extract an ip address from the facts of the pool members.

adreyer:
@MFredette does this need a ticket?

lucywyman:
For the work of adding inventory function support within apply blocks with inventory v2, yes. For docs I don't know.

bolt-modules/boltlib/lib/puppet/functions/get_targets.rb
```diff
@@ -3,6 +3,8 @@
 require 'bolt/error'
 
 # Parses common ways of referring to targets and returns an array of Targets.
+#
+# **NOTE:** Not available in apply block if using Inventory version 2
```
donoghuc: In every other case where we say "not available in apply block" an error is raised when the offending function is called in an apply block. In this case we just have empty data. See https://github.com/puppetlabs/bolt/blob/600252ce36bf624b7a53c9365e66e23b67829cd4/lib/bolt/inventory/inventory2.rb#L106-L116 . Unfortunately I think there is a bug. For example `get_targets('all')` returns an empty set of targets. But `get_targets('foo')` raises a random error. 
```
ab8omb37vvw9i8f.delivery.puppetlabs.net: Evaluation Error: Error while evaluating a Function Call, undefined method `[]' for nil:NilClass
```
I think for now the desired behavior is to always return an empty result list? It is certainly not to raise that exception.

donoghuc: https://github.com/puppetlabs/bolt/pull/1105 is a proposed fix. Still needs tests if we decide this is the desired behavior.

donoghuc: Updated with tests.

pre-docs/inventory_version2.md
```diff
@@ -39,6 +39,43 @@ string to a target defined elsewhere in the inventory. If no target has a name
 or alias matching the string bolt will create a new target with the string as
 it's uri.
 
+### Migrating Plans
+
+In addition to inventory file changes, inventory functions such as `get_targets` may not work as expected when called from an [apply block](applying_manifest_blocks.md) using inventory version 2. `get_targets` will return an empty array when called with `'all'` as an argument, otherwise it will create a new Target object for that host. If you want to have the same behavior as the version 1 inventory, you can extract information outside of an apply block into a variable, and use that variable inside the apply block.
```
cwcowellshah: Per Puppet style guide, recommend changing `may` to `might` since `may` can feel ambiguous to non-native English speakers (or even native speakers):
> ... might not work as expected when called from an ...

cwcowellshah: Recommend present tense and breaking up run-on:
> `get_targets` returns an empty array when called with `'all'` as an argument. Otherwise, it creates a new Target object for that host.

pre-docs/plan_functions.md
```diff
@@ -345,6 +345,10 @@ get_resources('target1,target2', [Package, File[/etc/puppetlabs]])
 
 Parses common ways of referring to targets and returns an array of Targets.
 
+**NOTE:** When using get_targets inside an apply block with a v2
```
cwcowellshah: The text above uses `version 2`, and this text uses `v2`. Can we standardize on one approach? Not sure which is preferred.

cwcowellshah: Double-check me on this, but we probably want to backtick these two terms:

>When using `get_targets` inside an `apply` block...

pre-docs/plan_functions.md
```diff
@@ -345,6 +345,10 @@ get_resources('target1,target2', [Package, File[/etc/puppetlabs]])
...
+inventory, a new Target object will be created. In the case that
+get_targets('all') is called, it will return an empty array.
```
cwcowellshah: Prefer active voice, standardizing on "version 2" instead of "v2", and backticking any words that are actually code:

>**NOTE:** Using `get_targets` inside an `apply` block with a version 2 inventory creates a new Target object. `get_targets('all')` returns an empty array.

PR #1100, opened 2019-07-24
(maint) Add a CODEOWNERS file

lucywyman:
This adds a CODEOWNERS file to Bolt, which will automatically request
reviews from the Bolt team members for PRs. It will also request review
from our docs person when the docs are updated.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,0 +1,4 @@
+# Default to Bolt team
+* @puppetlabs/bolt
+
+/pre-docs @MFredette @puppetlabs/bolt


CODEOWNERS
```diff
@@ -0,0 +1,4 @@
+# Default to Bolt team
+* @puppetlabs/bolt
+
+/pre-docs @MFredette @puppetlabs/bolt
```
adreyer: Is this still accurate?

lucywyman: I think it will change to Chris soon, but for now I think? @MFredette ?

PR #1102, opened 2019-07-26
(BOLT-1433) Load resource types for plan execution

donoghuc:
With this commit resource types are loaded when a plan is run. This allows resource types to be referenced without running in to `Evaluation Error: Resource type not found...` for use with plan functions such as `get_resources`.

### DIFF
--- a/bolt-modules/boltlib/lib/puppet/functions/get_resources.rb
+++ b/bolt-modules/boltlib/lib/puppet/functions/get_resources.rb
@@ -6,7 +6,10 @@ require 'bolt/task'
 # The results are returned as a list of hashes representing each resource.
 #
 # Requires the Puppet Agent be installed on the target, which can be accomplished with apply_prep
-# or by directly running the puppet_agent::install task.
+# or by directly running the puppet_agent::install task. In order to be able to reference types without
+# string quoting (for example `get_resources($target, Package)` instead of `get_resources($target, 'Package')`)
+# run the command `bolt puppetfile generate-types` to generate type references in `$Boldir/.resource_types`.
+#
 #
 # **NOTE:** Not available in apply block
 Puppet::Functions.create_function(:get_resources) do
@@ -16,7 +19,7 @@ Puppet::Functions.create_function(:get_resources) do
   #   get_resources('target1,target2', [Package, File[/etc/puppetlabs]])
   dispatch :get_resources do
     param 'Boltlib::TargetSpec', :targets
-    param 'Variant[String, Resource, Array[Variant[String, Resource]]]', :resources
+    param 'Variant[String, Type[Resource], Array[Variant[String, Type[Resource]]]]', :resources
   end
 
   def script_compiler
@@ -44,6 +47,10 @@ Puppet::Functions.create_function(:get_resources) do
     inventory = Puppet.lookup(:bolt_inventory)
 
     resources = [resources].flatten
+
+    # Stringify resource types to pass to task
+    resources.map! { |r| r.is_a?(String) ? r : r.to_s }
+
     resources.each do |resource|
       if resource !~ /^\w+$/ && resource !~ /^\w+\[.+\]$/
         raise Bolt::Error.new("#{resource} is not a valid resource type or type instance name", 'bolt/get-resources')
--- a/lib/bolt/pal.rb
+++ b/lib/bolt/pal.rb
@@ -39,7 +39,7 @@ module Bolt
 
     attr_reader :modulepath
 
-    def initialize(modulepath, hiera_config, max_compiles = Etc.nprocessors)
+    def initialize(modulepath, hiera_config, resource_types, max_compiles = Etc.nprocessors)
       # Nothing works without initialized this global state. Reinitializing
       # is safe and in practice only happens in tests
       self.class.load_puppet
@@ -48,6 +48,7 @@ module Bolt
       @modulepath = [BOLTLIB_PATH, *modulepath, MODULES_PATH]
       @hiera_config = hiera_config
       @max_compiles = max_compiles
+      @resource_types = resource_types
 
       @logger = Logging.logger[self]
       if modulepath && !modulepath.empty?
@@ -110,6 +111,21 @@ module Bolt
       compiler.evaluate_string('type PlanResult = Boltlib::PlanResult')
     end
 
+    # Register all resource types defined in $Boltdir/.resource_types as well as
+    # the built in types registered with the runtime_3_init method.
+    def register_resource_types(loaders)
+      static_loader = loaders.static_loader
+      static_loader.runtime_3_init
+      if File.directory?(@resource_types)
+        Dir.children(@resource_types).each do |resource_pp|
+          type_name_from_file = File.basename(resource_pp, '.pp').capitalize
+          typed_name = Puppet::Pops::Loader::TypedName.new(:type, type_name_from_file)
+          resource_type = Puppet::Pops::Types::TypeFactory.resource(type_name_from_file)
+          loaders.static_loader.set_entry(typed_name, resource_type)
+        end
+      end
+    end
+
     # Runs a block in a PAL script compiler configured for Bolt.  Catches
     # exceptions thrown by the block and re-raises them ensuring they are
     # Bolt::Errors since the script compiler block will squash all exceptions.
@@ -119,6 +135,7 @@ module Bolt
       r = Puppet::Pal.in_tmp_environment('bolt', modulepath: @modulepath, facts: {}) do |pal|
         pal.with_script_compiler do |compiler|
           alias_types(compiler)
+          register_resource_types(Puppet.lookup(:loaders)) if @resource_types
           begin
             Puppet.override(yaml_plan_instantiator: Bolt::PAL::YamlPlan::Loader) do
               yield compiler
@@ -348,6 +365,16 @@ module Bolt
 
           [path, values]
         end.to_h
+      end
+    end
+
+    def generate_types
+      require 'puppet/face/generate'
+      in_bolt_compiler do
+        generator = Puppet::Generate::Type
+        inputs = generator.find_inputs(:pcore)
+        FileUtils.mkdir_p(@resource_types)
+        generator.generate(inputs, @resource_types, true)
       end
     end
 
--- a/pre-docs/bolt_command_reference.md
+++ b/pre-docs/bolt_command_reference.md
@@ -328,6 +328,22 @@ List modules available to Bolt.
 | `--boltdir FILEPATH` | Specify what Boltdir to load config from. | Autodiscovered from current working directory. |
 | `--configfile FILEPATH` | Specify where to load config from. | `~/.puppetlabs/bolt/bolt.yaml` |
 
+## `puppetfile generate-types`
+
+Generate type references to register in Plans
+
+### Usage
+
+`bolt puppetfile generate-types`
+
+### Options
+
+| Option | Description | Default |
+|--------|-------------|---------|
+| **Run Context** |
+| `-m`, `--modulepath FILEPATHS` | List of directories containing modules, separated by `:`. <br> Directories are case-sensitive. |
+| `--boltdir FILEPATH` | Specify what Boltdir to load config from. | Autodiscovered from current working directory. |
+| `--configfile FILEPATH` | Specify where to load config from. | `~/.puppetlabs/bolt/bolt.yaml` |
 
 ## `script run`
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,12 +10,15 @@
 
   When using the interpreters setting on the WinRM transport, a task would fail to execute if the path to the specified interpreter contains a space. Interpreter paths with spaces in the name are now supported.
 
+* **Resource Types were not registered while running plans** ([#1140](https://github.com/puppetlabs/bolt/issues/1140))
+
+  When referencing a resource type in a plan (for example `get_resources($nodes, Package)`) the plan would fail because the `Package` type is not found. Now all built in types and types on the modulepath can be generated with an invocation of `puppetfile generate-types` which will be registered for plan execution.
+  
 #### New features
 
 * **Azure inventory plugin** ([#1148](https://github.com/puppetlabs/bolt/issues/1148))
 
   A new [module based plugin](https://github.com/puppetlabs/puppetlabs-azure_inventory) allows generation of Bolt targets from Azure VMs.
-
 
 ## 1.31.1
 
--- a/bolt-modules/boltlib/spec/functions/run_plan_spec.rb
+++ b/bolt-modules/boltlib/spec/functions/run_plan_spec.rb
@@ -8,12 +8,13 @@ describe 'run_plan' do
   include PuppetlabsSpec::Fixtures
   let(:executor) { Bolt::Executor.new }
   let(:tasks_enabled) { true }
+  let(:inventory) { Bolt::Inventory.new({}) }
 
   around(:each) do |example|
     Puppet[:tasks] = tasks_enabled
     executor.stubs(:noop).returns(false)
 
-    Puppet.override(bolt_executor: executor) do
+    Puppet.override(bolt_executor: executor, bolt_inventory: inventory) do
       example.run
     end
   end
--- a/lib/bolt/bolt_option_parser.rb
+++ b/lib/bolt/bolt_option_parser.rb
@@ -54,6 +54,9 @@ module Bolt
         when 'show-modules'
           { flags: OPTIONS[:global] + OPTIONS[:global_config_setters],
             banner: PUPPETFILE_SHOWMODULES_HELP }
+        when 'generate-types'
+          { flags: OPTIONS[:global] + OPTIONS[:global_config_setters],
+            banner: PUPPETFILE_GENERATETYPES_HELP }
         else
           { flags: OPTIONS[:global] + OPTIONS[:global_config_setters],
             banner: PUPPETFILE_HELP }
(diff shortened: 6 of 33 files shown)


bolt-modules/boltlib/lib/puppet/functions/get_resources.rb
```diff
@@ -16,7 +16,7 @@
   #   get_resources('target1,target2', [Package, File[/etc/puppetlabs]])
   dispatch :get_resources do
     param 'Boltlib::TargetSpec', :targets
-    param 'Variant[String, Resource, Array[Variant[String, Resource]]]', :resources
+    param 'Variant[String, Type[Resource], Array[Variant[String, Type[Resource]]]]', :resources
```
donoghuc: Without adding `Type[]` this was complaining.

bolt-modules/boltlib/lib/puppet/functions/run_plan.rb
```diff
@@ -67,6 +67,9 @@ def run_plan(scope, plan_name, named_args = {})
     # The perspective of the environment is wanted here (for now) to not have to
     # require modules to have dependencies defined in meta data.
     loader = loaders.private_environment_loader
+    # Load resource types (full path commented below, TODO: delete)
+    # closure_scope.compiler.loaders.static_loader.runtime_3_init
+    loaders.static_loader.runtime_3_init
```
donoghuc: Unsure if this will load resource types on the module path as well. Still need to test.

donoghuc:
Previously the following plan:
```
plan test(TargetSpec $nodes) {
  $x = get_resources($nodes, Package)
  notice($x.results)
}
```
would error with: 
```
cas@cas-ThinkPad-T460p:~/working_dir/bolt$ bolt plan run test -t resource
Starting: plan test
Starting: plan test
Finished: plan test in 0.0 sec
Finished: plan test in 0.0 sec
{
  "kind": "bolt/pal-error",
  "msg": "Evaluation Error: Resource type not found: Package (file: /home/cas/working_dir/bolt/Boltdir/site/test/plans/init.pp, line: 247, column: 30)",
  "details": {
  }
}
```
By initializing the static loader it appears that referencing resource types is allowed. I have been investigating the other side effects of this but I wanted to put up a minimal example for early review or guidance.

bolt-modules/boltlib/lib/puppet/functions/run_plan.rb
adreyer: Does this belong in our PAL class?

nicklewis: I would expect this to probably go next to `alias_types`. https://github.com/puppetlabs/bolt/blob/master/lib/bolt/pal.rb#L109

donoghuc: I ended up doing it here because I did not see a way around using some hackery next to `alias_types` (something like: `compiler.instance_variable_get(:@internal_compiler).loaders.static_loader.runtime_3_init`)

HOWEVER Could the solution for this to be better handled in puppet? Specifically by **not** treating `Puppet[:tasks]]` special? 
https://github.com/puppetlabs/puppet/blob/904023ffc59bc6771c0c7abf2a1d2c4acf941b09/lib/puppet/pops/loaders.rb#L383-L397

donoghuc:
It seems like in order to have resource types available in plans we will need to use the `StaticLoader` for built-in resource types and the `Runtime3Loader` for resource types defined on the modulepath. These loaders have been explicitly disabled when using `Puppet[:tasks]` (more context on why the https://github.com/puppetlabs/puppet/pull/6240 ). An example of an issue due to loading manifest code is described with https://tickets.puppetlabs.com/browse/BOLT-100

Gemfile
```diff
@@ -15,6 +15,7 @@ gem "rack", '>= 2.0.5'
 gem "rails-auth", '>= 2.1.4'
 gem "sinatra", '>= 2.0.4'
 
+gem "puppet", git: "https://github.com/donoghuc/puppet.git", branch: 'BOLT-1433'
```
donoghuc: This fork contains a revert for disabling loaders for `Puppet[:tasks]`. I am trying to understand the type name vs task name conflict concern. For example the following plan seems to work just fine even though both the package task as well as the package type are both used in the same plan. 
```
plan test(TargetSpec $nodes) {
  # built in type
  $x = get_resources($nodes, Package)
  # loaded core type
  $y = get_resources($nodes, Cron)
  # task name same as built in type 
  $z = run_task(package, $nodes, action => status, name => vim)
  notice([$x.results, $y.results, $z.results])
```

donoghuc: After discussing in person I think that the concern about name collisions between tasks and types is no longer an issue. I am still looking in to the root cause of https://tickets.puppetlabs.com/browse/BOLT-100

lib/bolt/pal.rb
```diff
@@ -144,7 +144,8 @@ def with_bolt_executor(executor, inventory, pdb_client = nil, applicator = nil,
           pdb_client,
           @hiera_config,
           @max_compiles
-        )
+        ),
+        squelch_parse_errors: true
```
donoghuc: Following the stack trace in the ticket description at https://tickets.puppetlabs.com/browse/BOLT-100 I think that the `squelch_parse_errors` override option might have prevented it. I have a suspicion that while parsing some manifest code that assumed there was a `Catalog` which is not present in script mode (https://tickets.puppetlabs.com/browse/BOLT-1433?focusedCommentId=673243&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-673243). I am still working to reproduce this though.

donoghuc: Steps to try to repro:
With the following manifest at `/test/manifests/init.pp` 
```puppet
class test {
  # execute 'apt-get update'
  exec { 'apt-update':                    # exec resource named 'apt-update'
    command => '/usr/bin/apt-get update'  # command this resource will run
  }

  # install apache2 package
  package { 'apache2':
    require => Exec['apt-update'],        # require 'apt-update' before installing
    ensure => installed,
  }

  # ensure apache2 service is running
  service { 'apache2':
    ensure => running,
  }

  # install mysql-server package
  package { 'mysql-server':
    require => Exec['apt-update'],        # require 'apt-update' before installing
    ensure => installed,
  }

  # ensure mysql service is running
  service { 'mysql':
    ensure => running,
  }

  # install php5 package
  package { 'php7.0':
    require => Exec['apt-update'],        # require 'apt-update' before installing
    ensure => installed,
  }

  # ensure info.php file exists
  file { '/var/www/html/info.php':
    ensure => file,
    content => '<?php  phpinfo(); ?>',    # phpinfo code
    require => Package['apache2'],        # require 'apache2' package before creating
  } 
}
```
I can run the following plan that successfully sets up the lamp stack both with and without the squelch_parse_errors override. 
```puppet
 plan test(TargetSpec $nodes){
  $nodes.get_resources('User')
  run_task('facts', $nodes)
  apply($nodes){
    include test
  }
  return $nodes.get_resources(['user', 'File[/tmp]'])
}
```
So far the only somewhat promising difference I've been able to come up with between using the override vs not is that without it I can trigger an error that results in an exception with
```
From: /home/cas/working_dir/bolt/bolt-modules/boltlib/lib/puppet/functions/run_plan.rb @ line 73 run_plan#run_plan:

    68:     # require modules to have dependencies defined in meta data.
    69:     loader = loaders.private_environment_loader
    70:     # Load resource types (full path commented below, TODO: delete)
    71:     # closure_scope.compiler.loaders.static_loader.runtime_3_init
    72:     # loaders.static_loader.runtime_3_init
 => 73:     require 'pry'; binding.pry
    74:     # TODO: Why would we not have a private_environment_loader?
    75:     unless loader && (func = loader.load(:plan, plan_name))
    76:       raise Puppet::ParseErrorWithIssue.from_issue_and_stack(
    77:         Puppet::Pops::Issues.issue(:UNKNOWN_PLAN) { Bolt::Error.unknown_plan(plan_name) }
    78:       )

[1] pry(#<#<Class:0x0000565028400af0>>)> closure_scope.environment.known_resource_types.loader.import_all
A 'define' expression is only available when compiling a catalog (file: /home/cas/working_dir/bolt/Boltdir/modules/concat/manifests/init.pp, line: 53, column: 1)
A Resource Statement is only available when compiling a catalog (file: /home/cas/working_dir/bolt/Boltdir/modules/concat/manifests/init.pp, line: 94, column: 5)
A Resource Statement is only available when compiling a catalog (file: /home/cas/working_dir/bolt/Boltdir/modules/concat/manifests/init.pp, line: 116, column: 7)
A Resource Statement is only available when compiling a catalog (file: /home/cas/working_dir/bolt/Boltdir/modules/concat/manifests/init.pp, line: 124, column: 5)
```
Which can be avoided (and shows the warnings below pointing to the init.pp) when the `squelch_parse_error` symbol is set. 
```
 Host Class Definition is only available when compiling a catalog (file: /home/cas/working_dir/bolt/Boltdir/site/test/manifests/init.pp, line: 1, column: 1)
A Resource Statement is only available when compiling a catalog (file: /home/cas/working_dir/bolt/Boltdir/site/test/manifests/init.pp, line: 3, column: 2)
A Resource Statement is only available when compiling a catalog (file: /home/cas/working_dir/bolt/Boltdir/site/test/manifests/init.pp, line: 8, column: 2)
A Resource Statement is only available when compiling a catalog (file: /home/cas/working_dir/bolt/Boltdir/site/test/manifests/init.pp, line: 14, column: 2)
A Resource Statement is only available when compiling a catalog (file: /home/cas/working_dir/bolt/Boltdir/site/test/manifests/init.pp, line: 19, column: 2)
A Resource Statement is only available when compiling a catalog (file: /home/cas/working_dir/bolt/Boltdir/site/test/manifests/init.pp, line: 25, column: 2)
A Resource Statement is only available when compiling a catalog (file: /home/cas/working_dir/bolt/Boltdir/site/test/manifests/init.pp, line: 30, column: 2)
A Resource Statement is only available when compiling a catalog (file: /home/cas/working_dir/bolt/Boltdir/site/test/manifests/init.pp, line: 36, column: 2)
```

lib/bolt/pal.rb
```diff
@@ -107,6 +107,18 @@ def in_bolt_compiler
       r = Puppet::Pal.in_tmp_environment('bolt', modulepath: @modulepath, facts: {}) do |pal|
         pal.with_script_compiler do |compiler|
           alias_types(compiler)
+          
+          internal_compiler = compiler.instance_variable_get(:@internal_compiler)
+          loaders = internal_compiler.loaders
+          # # Load built in types
+          loaders.static_loader.runtime_3_init
```
donoghuc: This does successfully load built in types.

lib/bolt/pal.rb
```diff
@@ -107,6 +107,18 @@ def in_bolt_compiler
...
+          # Load types from modulepath
+          puppet_system_loader = internal_compiler.loaders.puppet_system_loader
```
donoghuc: This bit here attempts to set up the runtime3 loader, but it does not actually end up resolving the types.

lucywyman:
Closes #1140

pre-docs/bolt_command_reference.md
```diff
@@ -328,6 +328,22 @@ List modules available to Bolt.
 | `--boltdir FILEPATH` | Specify what Boltdir to load config from. | Autodiscovered from current working directory. |
 | `--configfile FILEPATH` | Specify where to load config from. | `~/.puppetlabs/bolt/bolt.yaml` |
 
+## `puppetfile generate-types`
```
nicklewis: What's the reason for making this a separate command and not running it automatically as part of `bolt puppetfile install`?

bolt-modules/boltlib/lib/puppet/functions/run_plan.rb
```diff
@@ -75,6 +75,21 @@ def run_plan(scope, plan_name, named_args = {})
...
+    boltdir_path = Puppet.lookup(:bolt_inventory).config.boltdir.path
+    static_loader = loaders.static_loader
+    static_loader.runtime_3_init
+    resource_dir = File.join(boltdir_path, '.resource_types')
+
+    if File.directory?(resource_dir)
+      Dir.entries(resource_dir).each do |resource_pp|
+        unless resource_pp.start_with?('.')
+          type_name = resource_pp[0..-4].capitalize
+          loaders.static_loader.send(:create_resource_type_reference, type_name)
```
nicklewis: I'm not sure whether it's better to forcibly use the private method here, which ensures we match the implementation in Puppet, or to copy the implementation, which uses the public API. I think that would be:

```rb
loaders.set_entry(TypedName.new(:type, type_name), Types::TypeFactory.resource(type_name))
```

My slight preference may be the latter, but I'm really on the fence.

bolt-modules/boltlib/lib/puppet/functions/run_plan.rb
```diff
@@ -75,6 +75,21 @@ def run_plan(scope, plan_name, named_args = {})
       )
     end
 
+    # Check for `$BOLTDIR/.resource_types` and create reference based on file name
```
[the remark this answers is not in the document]
donoghuc: I would like to do it in run_plan for two reasons
1. I can get the Boltdir path (we are already going to be looking up inventory for GH-1125 in run_plan)
2. We only need to register the types for plan execution, it seems wasteful to do it any time we need to do it in pal.

pre-docs/bolt_command_reference.md
donoghuc: We talked about it in the meeting about this. I think it is so you do not *have* to have a puppetfile at all? Like maybe you are developing a new module and writing some types, you can just generate-types every time you want to refresh?

bolt-modules/boltlib/lib/puppet/functions/run_plan.rb
donoghuc: I think using the public API is better. updated. Thanks.

bolt-modules/boltlib/lib/puppet/functions/run_plan.rb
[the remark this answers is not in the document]
nicklewis: On the flipside, in `run_plan` this will happen every time we run a plan, not just for the outer plan. It doesn't really take any time either way, though.

It just seems like it makes more sense to do it in PAL because it's part of setting up our execution environment rather than specifically part of executing a plan.

pre-docs/bolt_command_reference.md
nicklewis: That's a good reason to have a command for it. Is there a reason not to also do it in `bolt puppetfile install` though?

bolt-modules/boltlib/lib/puppet/functions/run_plan.rb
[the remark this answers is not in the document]
donoghuc: Cool. I can look in to how to get a hold of the `boldir` path there. Honestly I did not really try too hard before, I just had been working on GH-1125 and knew I could easily get it in `run_plan`.

pre-docs/bolt_command_reference.md
donoghuc: I think that makes sense. Will update to do it automatically with puppetfile install.

bolt-modules/boltlib/lib/puppet/functions/run_plan.rb
donoghuc: Update, we do want it in PAL class. We do not want to do it in puppet because of the risk of using the 3x loaders.

lib/bolt/pal.rb
donoghuc: Should not need this setting using the generate face approach.

bolt-modules/boltlib/lib/puppet/functions/run_plan.rb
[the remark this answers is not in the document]
donoghuc: Updated to do this in PAL.

donoghuc:
@puppetlabs/skeletor A review of this PR from one of your team would be nice. The idea is to get away from the 3x loaders for bolt/script compilation. We basically just use the `generate` face to generate the types (storing them under `$Boltdir/.resource_types` and then register them like the `builtin` types. This simply allows the types to be referenced in a plan (the minimum we currently need for plans to be aware of types). There is a comment above about how `codemanager` will run `generate resources` though the first iteration of plans in PE might not have this functionality. Are there any concerns about this approach?

lib/bolt/pal.rb
```diff
@@ -351,6 +368,17 @@ def list_modules
       end
     end
 
+    def generate_types(config)
+      require 'puppet/face/generate'
+      in_bolt_compiler do
+        generator = Puppet::Generate::Type
+        inputs = generator.find_inputs(:pcore)
+        outputdir = File.join(config.boltdir.path, '.resource_types')
```
nicklewis: Should this just be `config.boltdir.resource_types`?

donoghuc: Actually, now that pal has an attribute `@resource_types` I dont think this method even needs an argument.

bolt-modules/boltlib/lib/puppet/functions/run_plan.rb
```diff
@@ -75,6 +75,21 @@ def run_plan(scope, plan_name, named_args = {})
...
+    boltdir_path = Puppet.lookup(:bolt_inventory).config.boltdir.path
+    static_loader = loaders.static_loader
+    static_loader.runtime_3_init
+    resource_dir = File.join(boltdir_path, '.resource_types')
+
+    if File.directory?(resource_dir)
+      Dir.entries(resource_dir).each do |resource_pp|
```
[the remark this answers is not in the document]
donoghuc: I guess that is not available in ruby 2.3 https://github.com/puppetlabs/bolt/pull/1268

PR #1104, opened 2019-07-29
(maint) Split help text banners into their own file

beechtom:


lucywyman:
This seems like unnecessary complication? I'm not sure what we gain by doing this.

beechtom:
I was mostly just trying to clean up the option parser and make it a bit more readable since it was getting pretty lengthy from all of the additional banners that were added. Would there be a better approach / is it not really a concern?

PR #1105, opened 2019-07-29
(BOLT-1492) Fix error in get_targets with inventory v2 in apply

donoghuc:
Until get_targets is supported for inventory v2 in apply blocks empty data should be passed to bolt_catalog. There is still the expectation that config data (namely transport information) is passed along in order for a net new target to be returned when get_targets is used within an apply block with inventory v2. This commit sends along the expected serialized config data while still excluding any existing target information in the inventory.

### DIFF
--- a/spec/fixtures/apply/basic/plans/inventory_2_lookup.pp
+++ b/spec/fixtures/apply/basic/plans/inventory_2_lookup.pp
@@ -1,0 +1,11 @@
+plan basic::inventory_2_lookup(TargetSpec $nodes) {
+  # TODO: Provide inventory data to bolt_catalog to support get_targets in apply
+  return apply($nodes) {
+    # get_targets should return empty list for 'all'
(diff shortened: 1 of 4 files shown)


spec/fixtures/apply/basic/plans/inventory_2_lookup.pp
```diff
@@ -0,0 +1,11 @@
+plan basic::inventory_2_lookup(TargetSpec $nodes) {
+  # TODO: Provide inventory data to bolt_catalog to support get_targets in apply
+  return apply($nodes) {
+    # get_targets should return empty list for 'all'
+  	$t_all = get_targets('all')
```
lucywyman: This indentation seems weird

donoghuc: I had tab size set to 4 for some reason in my text editor. Updated.

PR #1106, opened 2019-07-30
(BOLT-1286) Document Yaml plan limitations

donoghuc:
This commit adds some examples of the sharp edges and quirks around converting plans between Yaml and Puppet.

### DIFF
--- a/pre-docs/writing_yaml_plans.md
+++ b/pre-docs/writing_yaml_plans.md
@@ -441,7 +441,7 @@ bolt plan convert path/to/my/plan.yaml
 
 This takes the path (relative or absolute) to the YAML plan to be converted and prints the converted Puppet plan to stdout.
 
-**NOTE:** Converting a YAML plan might result in a Puppet plan which is syntactically correct, but behaviorally different from its YAML plan. Always manually verify a converted Puppet plan's functionality. If you convert a YAML plan to puppet and it does *not* have the same behavior as the YAML plan, please [file an issue in our git repo](https://github.com/puppetlabs/bolt/issues).
+**NOTE:** Converting a YAML plan might result in a Puppet plan which is syntactically correct, but behaviorally different from its YAML plan. Always manually verify a converted Puppet plan's functionality. There are some constructs that do not translate from YAML plans to Puppet plans. These are [listed](#yaml-plan-constructs-that-cannot-be-translated-to-puppet-plans) below. If you convert a YAML plan to Puppet and it does *not* have the same behavior as the YAML plan, please [file an issue in our git repo](https://github.com/puppetlabs/bolt/issues).
 
 For example, with this YAML plan:
 ```yaml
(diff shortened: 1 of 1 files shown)


lucywyman:
I asked if Chris is the right person for review - if he is I'll update the CODEOWNERS

pre-docs/writing_yaml_plans.md
```diff
@@ -441,7 +441,7 @@ bolt plan convert path/to/my/plan.yaml
 
 This takes the path (relative or absolute) to the YAML plan to be converted and prints the converted Puppet plan to stdout.
 
-**NOTE:** Converting a YAML plan might result in a Puppet plan which is syntactically correct, but behaviorally different from its YAML plan. Always manually verify a converted Puppet plan's functionality. If you convert a YAML plan to puppet and it does *not* have the same behavior as the YAML plan, please [file an issue in our git repo](https://github.com/puppetlabs/bolt/issues).
+**NOTE:** Converting a YAML plan might result in a Puppet plan which is syntactically correct, but behaviorally different from its YAML plan. Always manually verify a converted Puppet plan's functionality. There are some cases where constructs do not translate from YAML plans to Puppet plans which are [listed](#yaml-plan-constructs-which-cannot-be-translated-to-puppet-plans) below. If you convert a YAML plan to puppet and it does *not* have the same behavior as the YAML plan, please [file an issue in our git repo](https://github.com/puppetlabs/bolt/issues).
```
cwcowellshah: capitalization fix: `puppet` to `Puppet` in the sentence that begins `If you convert a YAML plan to puppet...`

cwcowellshah: For clarity, I recommend changing:

>There are some cases where constructs do not translate from YAML plans to Puppet plans which are [listed](#yaml-plan-constructs-which-cannot-be-translated-to-puppet-plans) below.

to:

>There are some constructs that do not translate from YAML plans to Puppet plans. These are [listed](#yaml-plan-constructs-which-cannot-be-translated-to-puppet-plans) below.

pre-docs/writing_yaml_plans.md
```diff
@@ -473,3 +473,134 @@ plan mymodule::yamlplan(
 }
 ```
 
+## YAML plan constructs which cannot be translated to Puppet plans
```
cwcowellshah: Super nit-picky: because this is a defining clause it should be `YAML plan constructs that cannot...`

pre-docs/writing_yaml_plans.md
```diff
@@ -473,3 +473,134 @@ plan mymodule::yamlplan(
...
+
+This section points out some of the quirks and limitations associated with converting a plan expressed as YAML to a Puppet plan. In some cases it is impossible to accurately translate from YAML to Puppet, in others code that is generated from the conversion is syntactically correct but not idiomatic Puppet.
```
cwcowellshah: To avoid the run-on sentence feel, I'd break this long sentence into two:
> In some cases it is impossible to accurately translate from YAML to Puppet. In others, code that is generated from the conversion is syntactically correct but not idiomatic Puppet.

pre-docs/writing_yaml_plans.md
```diff
@@ -473,3 +473,134 @@ plan mymodule::yamlplan(
 }
 ```
 
+## YAML plan constructs which cannot be translated to Puppet plans
+
+This section points out some of the quirks and limitations associated with converting a plan expressed as YAML to a Puppet plan. In some cases it is impossible to accurately translate from YAML to Puppet, in others code that is generated from the conversion is syntactically correct but not idiomatic Puppet. 
+
+### Named eval step
+
+The eval step allows snippets of Puppet code to be expressed in YAML plans. When converting a multi-line eval step to Puppet and storing the resulting value in a variable the `with` lambda is used. For example, the following YAML plan:
```
cwcowellshah: Add comma and prefer active voice:

>When converting a multi-line eval step to Puppet and storing the resulting value in a variable, use the `with` lambda.

pre-docs/writing_yaml_plans.md
```diff
@@ -473,3 +473,134 @@ plan mymodule::yamlplan(
...
+) {
+  $eval_step = with() || {
+    $x = $foo + 1
+    $x * 2
+  }
+
+  return $eval_step
+}
+```
+If the author were writing the code as a Puppet plan they would likely not employ the use of the lambda. In this example the converted Puppet code is 'correct' but is not particularly natural or readable.
```
cwcowellshah: Shorten for clarity:

>>>If the author were writing the code as a Puppet plan, they would likely not use the lambda.

cwcowellshah: The quotation marks make it look like the correct code is actually incorrect. Recommend removing the quotations and shortening slightly:

>In this example the converted Puppet code is correct, but not natural or readable.

pre-docs/writing_yaml_plans.md
```diff
@@ -473,3 +473,134 @@ plan mymodule::yamlplan(
...
+  }
+
+  return $eval_step
+}
+```
+If the author were writing the code as a Puppet plan they would likely not employ the use of the lambda. In this example the converted Puppet code is 'correct' but is not particularly natural or readable. 
+
+### Resource step variable interpolation
+
+When applying puppet resources in a resource step variable interpolation will behave differently between YAML and Puppet plans. In order to illustrate an important distinction consider the following yaml plan:
```
cwcowellshah: `puppet` should be `Puppet`

pre-docs/writing_yaml_plans.md
```diff
@@ -473,3 +473,134 @@ plan mymodule::yamlplan(
...
+        content: $facts['os']['family']
+        ensure: present
+  - name: file_contents
+    description: Read contents of file managed with file resource
+    eval: >
+      file::read('/tmp/foo')
+      
+return: $file_contents
+```
+This plan performs apply prep on a localhost target, next it uses a Puppet `file` resource to write the OS family discovered from the Puppet `$facts` hash to a temporary file. Finally it reads the value written to the file and returns it. Running `bolt plan convert` on this plan produces the following puppet code:
```
cwcowellshah: Several minor fixes:

>This plan performs `apply prep` on a localhost target. Then it uses a Puppet `file` resource to write the OS family discovered from the Puppet `$facts` hash to a temporary file. Finally it reads the value written to the file and returns it. Running `bolt plan convert` on this plan produces this Puppet code:

pre-docs/writing_yaml_plans.md
```diff
@@ -473,3 +473,134 @@ plan mymodule::yamlplan(
...
+      content => $facts['os']['family'],
+      ensure => 'present',
+    }
+  }
+  $file_contents = file::read('/tmp/foo')
+
+  return $file_contents
+}
+```
+In this case the Puppet plan produced from the Yaml plan works as expected where the Yaml version fails. The cause of the failure is that the `$facts` variable is attempted to be resolved as a plan variable instead of being evaluated as part of compiling the manifest code in an apply block.
```
cwcowellshah: Minor wordsmithing and capitalization fix:

>This Puppet plan works as expected, whereas the YAML plan it was converted from fails. The failure stems from the `$facts` variable being resolved as a plan variable, instead of being evaluated as part of compiling the manifest code in an `apply` block.

pre-docs/writing_yaml_plans.md
```diff
@@ -473,3 +473,134 @@ plan mymodule::yamlplan(
...
+
+The next example also has a somewhat related interpolation issue. 
+
+## Dependency order
+
+The resources in a resource list are applied in order. It is still possible to explicitly set dependencies but it is important to note how to reference them. Consider the following Yaml plan:
```
cwcowellshah: `Yaml` should be `YAML`

pre-docs/writing_yaml_plans.md
```diff
@@ -473,3 +473,134 @@ plan mymodule::yamlplan(
...
+          before: File['/etc/ssh/sshd_config']
+      - title: /etc/ssh/sshd_config
+        type: file
+        parameters:
+          ensure: file
+          mode: '0600'
+          content: ''
+          require: Package['openssh-server']
+```
+Executing this plan will fail during catalog compilation because of how the resources referenced in the `before` and `require` parameters are parsed. Specifically the error will be at `Could not find resource 'File['/etc/ssh/sshd_config']' in parameter 'before'`. The solution for this is to not quote the resource titles. See the corrected plan example below.
```
cwcowellshah: Recommend changing `see the corrected plan example below.` to `Here is the corrected plan example:` This is because when I first read this I wasn't sure whether "below" meant way farther down in the document, or immediately below.

pre-docs/writing_yaml_plans.md
```diff
@@ -473,3 +473,134 @@ plan mymodule::yamlplan(
...
+          before: File[/etc/ssh/sshd_config]
+      - title: /etc/ssh/sshd_config
+        type: file
+        parameters:
+          ensure: file
+          mode: '0600'
+          content: ''
+          require: Package[openssh-server]
+```
+It should be noted that in general resources should be declared in order and this particular use case is uncommon. It is meant to illustrate a case where parameter parsing can lead to some non-intuitive results.
```
cwcowellshah: Recommend shortening:

>Note that in general, resources should be declared in order. This example is unusual because it is meant to illustrate a case where parameter parsing can lead to non-intuitive results.

pre-docs/writing_yaml_plans.md
MFredette: Not nitpicky to be grammatically correct.

PR #1107, opened 2019-07-30
(BOLT-1412) Only process task_options for tasks/plans

donoghuc:
When running a script there could be positional script arguments that contain `=` characters. Previously, regardless of the subcommand the cli parser would try to extract task_options by splitting arguments on the `=` character. This commit updates the parser to only attepmt to built a task_options hash if a task or plan is being run.


lucywyman:
Works locally for me!

PR #1109, opened 2019-07-31
(BOLT-1329) Delete wrapper script with sudoable connections

donoghuc:
This commit adds a line to delete the wrapper script that could contain sensitive parameters from the temporary directory once it it loaded. This is safe to do because the wrapper script is loaded only once. Deleting it reduces the time the sensitive parameters are stored on disk.

### DIFF
--- a/lib/bolt/transport/sudoable.rb
+++ b/lib/bolt/transport/sudoable.rb
@@ -120,9 +120,11 @@ module Bolt
 
               if conn.run_as && stdin
                 # Inject interpreter in to wrapper script and remove from execute options
-                wrapper = make_wrapper_stringio(remote_task_path, stdin, execute_options[:interpreter])
+                wrapper_name = 'wrapper.sh'
+                wrapper_path = File.join(dir.to_s, wrapper_name)
+                wrapper = make_wrapper_stringio(remote_task_path, stdin, wrapper_path, execute_options[:interpreter])
                 execute_options.delete(:interpreter)
-                remote_wrapper_path = conn.write_executable(dir, wrapper, 'wrapper.sh')
+                remote_wrapper_path = conn.write_executable(dir, wrapper, wrapper_name)
                 command << remote_wrapper_path
               else
                 command << remote_task_path
@@ -141,10 +143,11 @@ module Bolt
         end
       end
 
-      def make_wrapper_stringio(task_path, stdin, interpreter = nil)
+      def make_wrapper_stringio(task_path, stdin, wrapper_path, interpreter = nil)
         if interpreter
           StringIO.new(<<~SCRIPT)
             #!/bin/sh
+            rm '#{wrapper_path}'
             '#{interpreter}' '#{task_path}' <<'EOF'
             #{stdin}
             EOF
@@ -152,6 +155,7 @@ module Bolt
         else
           StringIO.new(<<~SCRIPT)
             #!/bin/sh
+            rm '#{wrapper_path}'
             '#{task_path}' <<'EOF'
             #{stdin}
             EOF
--- a/lib/bolt/transport/sudoable.rb
+++ b/lib/bolt/transport/sudoable.rb
@@ -120,12 +120,17 @@ module Bolt
 
               if conn.run_as && stdin
                 # Inject interpreter in to wrapper script and remove from execute options
-                wrapper_name = 'wrapper.sh'
-                wrapper_path = File.join(dir.to_s, wrapper_name)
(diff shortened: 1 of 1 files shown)


donoghuc:
I did some manual testing by setting breakpoint right before wrapper is executed and tested that after execution the file is deleted. Not sure about automated tests, open to ideas.

lib/bolt/transport/sudoable.rb
```diff
@@ -120,7 +120,8 @@ def run_task(target, task, arguments, options = {})
 
               if conn.run_as && stdin
                 # Inject interpreter in to wrapper script and remove from execute options
-                wrapper = make_wrapper_stringio(remote_task_path, stdin, execute_options[:interpreter])
+                wrapper_path = File.join(dir.to_s, 'wrapper.sh')
```
lucywyman: I'd be inclined to pull `'wrapper.sh'` into a var, since we use it twice. It seems silly to make the path twice, but I'd rather do that than manually call `copy_file` and `make_executable`.

donoghuc:
I think that even with huge stdin (even up to 150 M)  this is OK, tested with following plan
```puppet
plan test(TargetSpec $nodes) {
  $long = file::read('/home/cas/working_dir/bolt/foo')
  notice($long.length)
  $result = run_task('sample::stdin', $nodes, message => $long).results[0].to_data['result']['message']
  notice($result.length)
}
```
```
cas@cas-ThinkPad-T460p:~/working_dir/bolt$ bolt plan run test -t ixg143vm7hctjdc.delivery.puppetlabs.net --run-as root --sudo-password Qu@lity!
Starting: plan test
Starting: plan test
160000000
Starting: task sample::stdin on ixg143vm7hctjdc.delivery.puppetlabs.net
Finished: task sample::stdin with 0 failures in 11.92 sec
160000000
Finished: plan test in 12.03 sec
Finished: plan test in 12.03 sec
```

lucywyman:
Agree - I used https://github.com/puppetlabs/bolt/blob/master/spec/bolt/transport/local_spec.rb#L118 with ~200 megs and the task executed as expected.

donoghuc:
Further testing with total lines (vs just a really long line tested above) in the script also supports that this is safe to do. I generate a script with 
```
File.open('t.sh', 'w') do |f|
   f.write("rm '/home/cas/working_dir/bolt/t.sh'\n")
  (1..10000000).each { |i| f.write("echo #{i}\n")}
end
```
And can verify that it both deletes the source as well as echos every line. 
```
cas@cas-ThinkPad-T460p:~/working_dir/bolt$ chmod +x t.sh
cas@cas-ThinkPad-T460p:~/working_dir/bolt$ ./t.sh | tail
9999991
9999992
9999993
9999994
9999995
9999996
9999997
9999998
9999999
10000000
cas@cas-ThinkPad-T460p:~/working_dir/bolt$ stat t.sh
stat: cannot stat 't.sh': No such file or directory
```

lib/bolt/transport/sudoable.rb
donoghuc: updated

donoghuc:
Based on the discussion on https://stackoverflow.com/questions/2028874/what-happens-to-an-open-file-handle-on-linux-if-the-pointed-file-gets-moved-del it seems like this should be safe to do.

lib/bolt/transport/sudoable.rb
```diff
@@ -120,12 +120,17 @@ def run_task(target, task, arguments, options = {})
...
-                execute_options.delete(:interpreter)
-                remote_wrapper_path = conn.write_executable(dir, wrapper, wrapper_name)
-                command << remote_wrapper_path
+                # wrapper_name = 'wrapper.sh'
+                # wrapper_path = File.join(dir.to_s, wrapper_name)
+                # wrapper = make_wrapper_stringio(remote_task_path, stdin, wrapper_path, execute_options[:interpreter])
+                # execute_options.delete(:interpreter)
+                # remote_wrapper_path = conn.write_executable(dir, wrapper, wrapper_name)
+                build_command = if execute_options[:interpreter]
+                                  "sh -c '(cat <<EOF\n#{stdin}\nEOF\n) | #{execute_options[:interpreter]} #{remote_task_path}'"
```
donoghuc: Oh right, I dont think this will ever work with big input.
```
    Errno::E2BIG:
2646       Argument list too long
```

donoghuc:
Based on discussion this has sparked a new ticket has been filed https://tickets.puppetlabs.com/browse/BOLT-1502 to cover working on a way to completely circumvent using a wrapper script.

PR #1111, opened 2019-08-01
(BOLT-1424) Add Vault plugin

beechtom:
This adds support for looking up secrets from a Vault server to populate
fields in the inventory config. Secrets can be looked up by specifying
the path of a KV engine and the address of the server, and authenticating
through a token or userpass.

### DIFF
--- a/lib/bolt/plugin/vault.rb
+++ b/lib/bolt/plugin/vault.rb
@@ -25,7 +25,7 @@ module Bolt
 
       # Make sure no unexpected keys are in the config
       def validate_config(config)
-        known_keys = %w[server_url auth]
+        known_keys = %w[server_url auth cacert]
 
         config.each do |key, _v|
           next if known_keys.include?(key)
@@ -36,12 +36,12 @@ module Bolt
       # Make sure no unexpected keys are in the inventory config and
       # that required keys are present
       def validate_options(opts)
-        known_keys = %w[_plugin server_url auth path field version]
+        known_keys = %w[_plugin server_url auth path field version cacert]
         required_keys = %w[path]
 
         opts.each do |key, _v|
           next if known_keys.include?(key)
-          raise Bolt::ValidationError, "Unpexpected key in inventory config: #{key}"
+          raise Bolt::ValidationError, "Unexpected key in inventory config: #{key}"
         end
 
         required_keys.each do |key|
@@ -71,7 +71,7 @@ module Bolt
           TOKEN_HEADER => token(opts)
         }
 
-        response = request(:Get, uri(opts), nil, header)
+        response = request(:Get, uri(opts), opts, nil, header)
 
         parse_response(response, opts)
       end
@@ -82,12 +82,33 @@ module Bolt
 
         # Handle the different versions of the API
         if opts['version'] == 2
-          opts['path'] = opts['path'].split('/').insert(1, 'data').join('/')
+          mount, store = opts['path'].split('/', 2)
+          opts['path'] = [mount, 'data', store].join('/')
         end
 
         path ||= opts['path']
 
         URI.parse(File.join(url, "v1", path))
+      end
+
+      # Configure the http/s client
+      def client(uri, opts)
+        client = Net::HTTP.new(uri.host, uri.port)
+
+        if uri.scheme == 'https'
+          cacert = opts['cacert'] || config['cacert'] || ENV['VAULT_CACERT']
+
+          unless cacert
+            raise Bolt::ValidationError, "Expected cacert to be set when using https"
+          end
+
+          client.use_ssl = true
+          client.ssl_version = :TLSv1_2
+          client.ca_file = cacert
+          client.verify_mode = OpenSSL::SSL::VERIFY_PEER
+        end
+
+        client
       end
 
       # Auth token to vault server
@@ -99,12 +120,12 @@ module Bolt
         end
       end
 
-      def request(verb, uri, data, header = {})
+      def request(verb, uri, opts, data, header = {})
         # Add on any header options
         header = DEFAULT_HEADER.merge(header)
 
         # Create the HTTP request
-        http = Net::HTTP.new(uri.host, uri.port)
+        client = client(uri, opts)
         request = Net::HTTP.const_get(verb).new(uri.request_uri, header)
 
         # Attach any data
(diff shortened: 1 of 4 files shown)


pre-docs/inventory_version2.md
```diff
@@ -541,6 +541,80 @@ plugins:
 * `private-key`: The path to the private key file. default: `keys/private_key.pkcs7.pem`
 * `public-key`: The path to the public key file. default: `keys/public_key.pkcs7.pem`
 
+#### Vault plugin
+
+
+This plugin allows config values to be set by accessing secrets on a Vault server. It supports several fields:
```
cwcowellshah: Would it be helpful to have a link out to some external resource explaining what Vault is? I've never heard of it, but this sounds like a great feature that I'd want to take advantage of. Or maybe we should just rely on users to engage their own google-fu to find out more?

pre-docs/inventory_version2.md
```diff
@@ -541,6 +541,80 @@ plugins:
...
+- `_plugin`: The value of `_plugin` must be `vault`
+- `server_url`: The URL of the Vault server (optional, defaults to `ENV['VAULT_ADDR']`)
+- `auth`: The method for authorizing with the Vault server and any necessary parameters (optional, defaults to `ENV['VAULT_TOKEN']`)
+- `path`: The path to the secrets engine (required)
+- `field`: The specific secret being used (optional, defaults to a Ruby hash of all secrets at the `path`)
+
+
+**Authentication Methods**
+
+Vault requires a token to assign an identity and set of policies to a user before accessing secrets. Currently, the Vault plugin offers 2 authentication methods:
```
cwcowellshah: Do we need `Currently` here? Maybe there's a good reason for it.

pre-docs/inventory_version2.md
```diff
@@ -541,6 +541,80 @@ plugins:
...
+
+
+`token`
+
+Authenticate using a valid token. This method requires the following fields:
```
cwcowellshah: Do we need `valid` here?

pre-docs/inventory_version2.md
```diff
@@ -541,6 +541,80 @@ plugins:
...
+`userpass`
+
+Request a token by logging into the Vault server with a username and password. This method requires the following fields:
+
+- `method`: The value of `method` must be `userpass`
+- `user`: The username
+- `pass`: The password
+
+
+Each of the available fields for the Vault plugin can be used in the inventory configuration:
```
cwcowellshah: Would it be equally clear if we shortened and changed to active voice, like this:

`You can add any Vault plugin field to the inventory configuration:`

pre-docs/inventory_version2.md
```diff
@@ -541,6 +541,80 @@ plugins:
...
+        auth:
+          method: userpass
+          user: bolt
+          pass: bolt
+        path: secrets/bolt
+        field: xxxxx-xxxxx
+```
+
+
+You can also set configuration in the config file - typically `bolt.yaml` - under the `plugins` field. Bolt will prefer any fields set in the inventory file. The available fields for the config file are:
```
cwcowellshah: Since dashes are sometimes problematic (use one or two? Hyphen or dash? Em-dash or en-dash?) I recommend using parentheses in cases like this where either punctuation style would work equally well:

`You can also set configuration in the config file (typically **bolt.yaml**) under the **plugins** field....

cwcowellshah: I totally know what you mean, but I wonder if it would be clearer if we used a different construction instead of using `prefer`? Like maybe this wordier-but-unambiguous wording:

`If a field set in both the inventory file and the config file, Bolt will use the value set in the inventory file.`  

But if `prefer` is commonly used in this way as a term of art, disregard this comment.

pre-docs/inventory_version2.md
beechtom: I'm not sure. I'll ask about adding a link out during standup tomorrow. Thanks for the suggestions, Chris!

lib/bolt/plugin/vault.rb
```diff
@@ -0,0 +1,185 @@
...
+        parse_response(response, opts)
+      end
+
+      # Request uri - built up from Vault server url and secrets path
+      def uri(opts, path = nil)
+        url = opts['server_url'] || config['server_url'] || ENV['VAULT_ADDR']
+
+        # Handle the different versions of the API
+        if opts['version'] == 2
+          opts['path'] = opts['path'].split('/').insert(1, 'data').join('/')
```
nicklewis: This could also be something like:

```rb
mount, key = opts['path'].split('/', 2)
opts['path'] = [mount, 'data', key].join('/')
```

`split('/', 2)` will split it into exactly two parts, preserving slashes in the second part. Might be clearer than insert, but I'll leave that up to you.

beechtom:
Updated with https support.

lib/bolt/plugin/vault.rb
```diff
@@ -0,0 +1,206 @@
...
+
+      # Default header for all requests, including auth methods
+      DEFAULT_HEADER = {
+        "Content-Type" => "application/json",
+        "Accept" => "application/json"
+      }.freeze
+
+      # Make sure no unexpected keys are in the config
+      def validate_config(config)
+        known_keys = %w[server_url auth]
```
nicklewis: This needs to include `cacert`.

lib/bolt/plugin/vault.rb
```diff
@@ -0,0 +1,206 @@
...
+
+      # Make sure no unexpected keys are in the inventory config and
+      # that required keys are present
+      def validate_options(opts)
+        known_keys = %w[_plugin server_url auth path field version]
+        required_keys = %w[path]
+
+        opts.each do |key, _v|
+          next if known_keys.include?(key)
+          raise Bolt::ValidationError, "Unpexpected key in inventory config: #{key}"
```
nicklewis:
```suggestion
          raise Bolt::ValidationError, "Unexpected key in inventory config: #{key}"
```

lib/bolt/plugin/vault.rb
```diff
@@ -0,0 +1,206 @@
...
+        config.each do |key, _v|
+          next if known_keys.include?(key)
+          raise Bolt::ValidationError, "Unexpected key in Vault plugin config: #{key}"
+        end
+      end
+
+      # Make sure no unexpected keys are in the inventory config and
+      # that required keys are present
+      def validate_options(opts)
+        known_keys = %w[_plugin server_url auth path field version]
```
nicklewis: This also needs to include `cacert`.

beechtom:
Updated with correct validation.

PR #1113, opened 2019-08-02
(docs) Documents gotcha for enterprise win users with HOMEDRIVE set

cbrwflo:
The command `bolt puppetfile install` fails in certain situations where the
cachedir is inaccessible; this change calls out the ENV variables responsible.

The $HOMEDRIVE is set in some Windows enterprise environments (for specifying DFS mounts), and the bolt docs currently do not call this out.

### DIFF
--- a/pre-docs/bolt_installing_modules.md
+++ b/pre-docs/bolt_installing_modules.md
@@ -84,4 +84,4 @@ For more details about specifying modules in a Puppetfile, see the [Puppetfile d
      directory. To override this location, update the modulepath setting in the
      Bolt config file.
 
-
+NOTE: Bolt uses r10k to download modules, and caches them in `~/.r10k/cache` before unzipping them. If you're a Windows user in an enterprise environment, ensure `HOME` or `HOMEDRIVE + HOMEPATH` is set sensibly.


pre-docs/bolt_installing_modules.md
```diff
@@ -84,4 +84,4 @@ For more details about specifying modules in a Puppetfile, see the [Puppetfile d
      directory. To override this location, update the modulepath setting in the
      Bolt config file.
 
-
+NOTE: Bolt uses r10k to download modules, and caches them in `~/.r10k/cache` before unzipping them. If you're a Windows user in an enterprise environment, ensure `HOME` or `HOMEDRIVE + HOMEPATH` is set sensibly.
```
lucywyman: Thank you so much for contributing - this will save people a lot of debugging time! One small note: I'm not sure what `sensibly` means in this context? Can you elaborate, or be more specific a bit?

cbrwflo: This is pretty specific to having multiple mounts and windows, so probably sensibly===on the same mount.

The issue I ran into (during a Bolt workshop) was that my HOMEDRIVE was set to a different (inaccessible) mount, than my CWD was

lucywyman: Hmm, maybe "ensure `HOME` or `HOMEDRIVE + HOMEPATH` is accessible from you Bolt controller"?  It's also probably fine as is - I'll let our docs guru @cwcowellshah weigh in

nicklewis: I read "sensibly" here to mean "they actually point at your HOME", which makes sense to me.

PR #1115, opened 2019-08-05
(BOLT-1422) Update command reference docs

beechtom:
This updates the command reference documentation to show available
options for each subcommand-action. Previously, the documentation had a
list of subcommand-actions and a list of options, making it difficult to
tell which options applied to which subcommand-actions.

This also adds the 'secret' subcommand to the command reference.

### DIFF
--- a/pre-docs/bolt_command_reference.md
+++ b/pre-docs/bolt_command_reference.md
@@ -2,108 +2,497 @@
 
 Review the subcommands, actions, and options that are available for Bolt.
 
-## Common Bolt commands
+- [apply](#apply)
+- [command run](#command-run)
+- [file upload](#file-upload)
+- [inventory show](#inventory-show)
+- [plan convert](#plan-convert)
+- [plan run](#plan-run)
+- [plan show](#plan-show)
+- [puppetfile install](#puppetfile-install)
+- [puppetfile show-modules](#puppetfile-show-modules)
+- [script run](#script-run)
+- [secret createkeys](#secret-createkeys)
+- [secret decrypt](#secret-decrypt)
+- [secret encrypt](#secret-encrypt)
+- [task run](#task-run)
+- [task show](#task-show)
 
-Bolt commands use the syntax: `bolt <subcommand> <action> [options]`
 
-|Command|Description|Arguments|
-|-------|-----------|---------|
-| `bolt command run` `<COMMAND>` | Runs a command on remote nodes. |- The command, single quoted if it contains spaces or special characters.<br>- The nodes on which to run the command.
-| `bolt script run` | Runs a script in any language that will run on the remote system. |- A path to a local script to run on the remote nodes.<br>- Optionally, arguments to pass to the script.<br>- The nodes on which to run the script.
-| `bolt task run` | Runs a task on a remote system, passing any specified parameters. | - The task name, in the format `modulename::taskname`.<br>- The nodes on which to run the task.
-| `bolt plan run` | Runs a task plan. | - The plan name, in the format `modulename::planname`.<br>- The nodes on which to run the plan.
-| `bolt apply` | Applies a Puppet manifest file. | - The path to the manifest file.<br>- The nodes on which to run the plan.
-| `bolt file upload` | Uploads a local file or directory to a remote node. | - The path to the source file or directory.<br>- The path to the remote location.<br>- The nodes on which to upload the file or directory.
-| `bolt task show` | Lists all the tasks on the modulepath that have not been marked `private`. Will note whether a task supports no-operation mode. | - Adding a specific task name displays details and parameters for the task.<br>- Optionally, the name of a task you want details for: `bolt task show <TASK NAME>`
-| `bolt plan show` | Lists the plans that are installed on the current module path. | - Adding a specific plan name displays details and parameters for the plan.
-| `bolt plan convert` | Converts a YAML plan to a Puppet plan | - The path (relative or absolute) to the YAML plan to be converted.
-| `bolt puppetfile install` | Installs the modules listed in the Puppetfile to the current `Boltdir`. | |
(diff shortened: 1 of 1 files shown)


pre-docs/bolt_command_reference.md
```diff
@@ -1,109 +1,498 @@
...
+### Usage
+
+`bolt plan convert <PLAN>`
+
+### Options
+
+| Option | Description | Default |
+|--------|-------------|---------|
+| **Run Context** |
+| `-m`, `--modulepath FILEPATHS` | List of directories containing modules, separated by `:`. <br> Directories are case-sensitive. |
```
donoghuc: Not related to this PR, but do the `global_config_setters` actually do anything for plan convert?

donoghuc: https://github.com/puppetlabs/bolt/blob/52645942d5812a4f70a79f7d9a675acce44d0ca9/lib/bolt/bolt_option_parser.rb#L13

pre-docs/bolt_command_reference.md
```diff
@@ -1,109 +1,498 @@
...
-|`--user`, `-u`|User to authenticate as.|
-|`--password`, `-p`|Password to authenticate with. Pass this flag without any value to securely prompt for the password.|
-| `--private-key` |Private ssh key to authenticate with|
-| `--host-key-check, --no-host-key-check` | Do not require verification of new hosts in the `known_hosts` file.
+| Option | Description | Default |
+|--------|-------------|---------|
+| `-n`, `--nodes NODES` | Alias for `--targets`. |
+| `-t`, `--targets TARGETS` | Identifies the targets of command. |
+| `-q`, `--query QUERY` | Query PuppetDB to determine the targets. <br> Enter a comma-separated list of target URIs or group names. Or read a target list from an input file `@<file>` or stdin `-`. |
+| `--rerun FILTER` | Retry on nodes from the last run. <br> `all` runs on all targets from the last run. <br> `failure` runs on all targets that failed in the last run. <br> `success` runs on all targets that succeded in the last run. |
```
donoghuc: s/succeded/succeeded/g I would make the suggestion in the UI, but it looks like the typo got copied around. Will likely be easiest to just do it in your text editor.

pre-docs/bolt_command_reference.md
```diff
@@ -1,109 +1,498 @@
...
+| `--noop` | Execute a task that suports it in noop mode. |
```
donoghuc:
```suggestion
| `--noop` | Execute a task that supports it in noop mode. |
``

**Comment by lucywyman** [comment]
I'm not sure exactly how this would work, but this seems like it can be generated from the existing help text. Should we explore doing that instead?

File: pre-docs/bolt_command_reference.md [hunk 4]
```diff
@@ -1,109 +1,498 @@
-# Bolt command reference
+# Bolt Command Reference
```
cwcowellshah: Puppet corp style guide recommends sentence case for headers (minus any punctuation at the end). So the original version of this is preferred.

pre-docs/bolt_command_reference.md
```diff
@@ -1,109 +1,498 @@
...
-| `bolt file upload` | Uploads a local file or directory to a remote node. | - The path to the source file or directory.<br>- The path to the remote location.<br>- The nodes on which to upload the file or directory.
-| `bolt task show` | Lists all the tasks on the modulepath that have not been marked `private`. Will note whether a task supports no-operation mode. | - Adding a specific task name displays details and parameters for the task.<br>- Optionally, the name of a task you want details for: `bolt task show <TASK NAME>`
-| `bolt plan show` | Lists the plans that are installed on the current module path. | - Adding a specific plan name displays details and parameters for the plan.
-| `bolt plan convert` | Converts a YAML plan to a Puppet plan | - The path (relative or absolute) to the YAML plan to be converted.
-| `bolt puppetfile install` | Installs the modules listed in the Puppetfile to the current `Boltdir`. | |
-| `bolt puppetfile show-modules` | Lists the modules, and their versions, that are installed in the current `Boltdir`. | |
-| `bolt inventory show` | Show the list of targets an action would run on. | - The inventory targets to show. |
+## Global Options
```
cwcowellshah: lower case `options` is preferred

pre-docs/bolt_command_reference.md
```diff
@@ -1,109 +1,498 @@
...
 
-## Command options
+The following options are available for all subcommands and actions.
```
cwcowellshah: Nit-picky, but we could slightly shorten this, and it feels like it wants a colon:

`These options are available for all subcommands and actions:`

pre-docs/bolt_command_reference.md
```diff
@@ -1,109 +1,498 @@
...
+|--------|-------------|---------|
+| **Run Context** |
+| `-m`, `--modulepath FILEPATHS` | List of directories containing modules, separated by `:`. <br> Directories are case-sensitive. |
+| `--boltdir FILEPATH` | Specify what Boltdir to load config from. | Autodiscovered from current working directory. |
+| `--configfile FILEPATH` | Specify where to load config from. | `~/.puppetlabs/bolt/bolt.yaml` |
+
+
+## `puppetfile install`
+
+Install modules from a Puppetfile into a Boltdir.
```
cwcowellshah: A few lines below, `Boltdir` has backticks. Are they deliberately left off in this line?

Come to think of it, there are lots of references to Boltdir in this file. I wonder if it makes the most sense to use backticks when referring to an actual directory named `Boltdir`, and no backticks when referring to the concept of a Boltdir. Maybe like this?

"Move the file to your Boltdir"
vs.
"Move the file to `Boltdir`"

If you're cool with that sort of change, we should do a search & replace to find all references in this file.

beechtom: Good catch. I removed backticks from all instances since it's always referring to the concept of a Boltdir.

PR #1121, opened 2019-08-06
(maint) Ensure that rspec unit tag does not rely on ssh/winrm targets

donoghuc:
Limit the catch_errors integration tests to ssh targets. Also note in the CONTRIBUTING.md that bundled modules are required for some tests.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -104,6 +104,11 @@ Some tests require a Windows VM or Linux containers. For Linux tests (recommende
 
 Additional tests may run in a local environment and require certain shell capabilities. For example tests that require a Bash-like environment are tagged with `:bash` in rspec.
 
+Some tests will also require that the [bundled modules](#contributing-to-bundled-modules) described above are installed. Ensure the modules are installed for testing with the following command:
+```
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -104,6 +104,8 @@ Some tests require a Windows VM or Linux containers. For Linux tests (recommende
 
 Additional tests may run in a local environment and require certain shell capabilities. For example tests that require a Bash-like environment are tagged with `:bash` in rspec.
 
+Some tests will also require that the [bundled modules](#contributing-to-bundled-modules) described above are installed. Ensure the modules are installed for testing with `bundle exec r10k puppetfile install .`.
```
lucywyman:
```suggestion
Some tests will also require that the [bundled modules](#contributing-to-bundled-modules) described above are installed. Ensure the modules are installed for testing with
```
bundle exec r10k puppetfile install
```
```

lucywyman: Ugh this formatting is weird. Basically move it into a newline :p

donoghuc: Ha, yeah that is a weird one. Updated.

PR #1124, opened 2019-08-06
(BOLT-1400) Add `run_command` endpoint to bolt server

smcclellan:
In cases where PXP Agent is not installed on a target node
and thus cannot run a command, SSH or WinRM should be used.
This commit adds the `POST /run_command` endpoint which will
delegate either to SSH or to WinRM accordingly.

### DIFF
--- a/spec/bolt_server/app_integration_spec.rb
+++ b/spec/bolt_server/app_integration_spec.rb
@@ -19,52 +19,85 @@ describe "BoltServer::TransportApp", puppetserver: true do
   end
 
   context 'with ssh target', ssh: true do
-    let(:path) { '/ssh/run_task' }
+    describe "run_task" do
+      let(:path) { '/ssh/run_task' }
 
-    it 'runs an echo task with a password' do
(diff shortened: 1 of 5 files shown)


spec/bolt_server/app_integration_spec.rb
```diff
@@ -19,52 +19,85 @@ def app
...
-      expect(files[2]).to match(%r{^310 .*/results/lib/puppet/functions/results/make_result.rb$})
-      expect(files[3]).to match(%r{^43 .*/error/tasks/fail.sh$})
+        post(path, JSON.generate(body), 'CONTENT_TYPE' => 'text/json')
+        expect(last_response).to be_ok
+        expect(last_response.status).to eq(200)
+        result = JSON.parse(last_response.body)
+        expect(result).to include('status' => 'failure')
+        expect(result['result']["exit_code"]).to eq(127)
+        expect(result['result']["stderr"]).to match(/not-a-command/)
+      end
```
smcclellan: Any other test cases I should include here?

PR #1127, opened 2019-08-06
(BOLT-1451) Implicitly call apply_prep for yaml resources step

donoghuc:
This commit implements calling apply_prep on the targets defined in a resources step of a yaml plan. Currently apply prep is implicitly called on all the targets listed in the step for each resources step. The transpiler has also been updated to reflect the change.

### DIFF
--- a/lib/bolt/pal/yaml_plan/step/resources.rb
+++ b/lib/bolt/pal/yaml_plan/step/resources.rb
@@ -65,6 +65,13 @@ module Bolt
 
           def transpile
             code = StringIO.new
+
+            code.print "  "
+            fn = 'apply_prep'
+            args = [@target]
+            code << function_call(fn, args)
+            code.print "\n"
+
             code.print "  "
             code.print "$#{@name} = " if @name
 
--- a/pre-docs/writing_yaml_plans.md
+++ b/pre-docs/writing_yaml_plans.md
@@ -175,7 +175,7 @@ steps:
 
 #### Resources step
 
-Use a `resources` step to apply a list of Puppet resources. A resource defines the _desired state_ for part of a target. Bolt will ensure each resource is in its desired state.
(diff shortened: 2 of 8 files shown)


donoghuc:
Fixes https://github.com/puppetlabs/bolt/issues/1128

pre-docs/writing_yaml_plans.md
```diff
@@ -175,7 +175,7 @@ steps:
 
 #### Resources step
 
-Use a `resources` step to apply a list of Puppet resources. A resource defines the _desired state_ for part of a target. Bolt will ensure each resource is in its desired state.
+Use a `resources` step to apply a list of Puppet resources. A resource defines the _desired state_ for part of a target. Bolt will ensure each resource is in its desired state. For each `resources` step the `apply_prep` plan function will be executed against the targets specified with the `targets` field. For more information about `apply_prep` see the [How manifest blocks are applied](./applying_manifest_blocks.md#how-manifest-blocks-are-applied) reference.
```
cwcowellshah:
```suggestion
Use a `resources` step to apply a list of Puppet resources. A resource defines the _desired state_ for part of a target. Bolt will ensure each resource is in its desired state. For each `resources` step, Bolt executes the `apply_prep` plan function against the targets specified with the `targets` field. For more information about `apply_prep` see the [How manifest blocks are applied](./applying_manifest_blocks.md#how-manifest-blocks-are-applied) reference.
```

lib/bolt/pal/yaml_plan/step/resources.rb
```diff
@@ -65,6 +65,7 @@ def body
 
           def transpile
             code = StringIO.new
+            code.print "  apply_prep(#{Bolt::Util.to_code(@target)})\n"
```
nicklewis: This should use the `Step#function_call` helper.

donoghuc: Updated. That never even crossed my mind. Thanks.

Issue #1129, opened 2019-08-07
Do not rely on wrapper script for run-as with stdin task input


donoghuc:
## Describe the Bug
Currently tasks run with the run-as option rely on laying down a wrapper script to direct task input data over stdin. There are several issues with this approach, the first is that the wrapper script can contain sensitive data in clear text. A ticket describing an idea to minimize the time the wrapper script lives on disk has been filed under https://tickets.puppetlabs.com/browse/BOLT-1329

The second issue described has to do with file permissions of the wrapper script when run on the local transport which is ticketed under https://tickets.puppetlabs.com/browse/BOLT-1283

The root issue is that there’s only one stdin, and both the password for sudo and the parameters of the task need to be on stdin. But the password should only be on stdin if sudo asks for it. But we don’t know how to conclusively determine that sudo doesn’t need the password.
So we skirt around that by wrapping the task executable in a script that also passes its stdin, so in the case where sudo doesn’t need the password, it will just start executing the script, which will pass stdin to the task.

## Expected Behavior
Instead of relying on the wrapper script we should implement a solution that does not involve one. The solution should be able to safely pass the input data (even if it is very large) to the task over stdin even if it is ambiguous whether a sudo password is required. The solution should avoid problems with passing the data over the bash CLI such as `Argument list too long` and exposing the sensitive data in the process list or in the shell history.

This could be accomplished by running something like  `sudo 'echo "okay done with sudo, please pass stdin now"; /path/to/actual/task`, and then wait to see the “done with sudo” message before passing the task parameters on stdin.

PR #1141, opened 2019-08-09
(GH-1159) Do not use wrapper script for tasks that require data on stdin 

donoghuc:
Previously when running a task that required parameters to be passed on stdin using the run-as option for the ssh or local transport a wrapper script was used on the target. The data to be sent on stdin to the task executable was written to the wrapper script which was directed to stdin. This allowed reserving stdin to *only* pass the privilage escalation password. Writing input data to disk is undesirable given sensitive parameters could be written to disk as well as permissions issues on the wrapper when executing bolt as root with the run-as option set to a less privileged user for the local transport. This commit updates the local and ssh transports to not use a wrapper script. This is accomplished by generating a random string and echoing it before executing a task executable, when the string is detected that indicates the privilege escalation has succeeded and it is safe to pass task data on stdin.

### DIFF
--- a/acceptance/tests/task_local.rb
+++ b/acceptance/tests/task_local.rb
@@ -38,9 +38,6 @@ test_name "bolt task run should execute tasks on localhost via local transport" 
       #!/bin/sh
       echo "$PT_greetings from $(whoami)"
       FILE
-      # TODO: Use input_method: both (default) once bug BOLT-1283 is fixed
-      conf = { 'input_method' => 'environment' }
-      create_remote_file(bolt, "#{dir}/modules/test/tasks/whoami_nix.json", conf.to_json)
     end
 
     step "execute `bolt task run` on localhost via local transport" do
--- a/lib/bolt/transport/local/shell.rb
+++ b/lib/bolt/transport/local/shell.rb
@@ -131,7 +131,7 @@ module Bolt
             else
               sudo_str = Shellwords.shelljoin(@target.options['run-as-command'] + [run_as])
             end
-            command_str = build_sudoable_command_str(command_str, sudo_str, @sudo_id, options[:stdin])
+            command_str = build_sudoable_command_str(command_str, sudo_str, @sudo_id, options)
           end
 
           command_arr = options[:environment].nil? ? [command_str] : [options[:environment], command_str]
--- a/lib/bolt/transport/local/shell.rb
+++ b/lib/bolt/transport/local/shell.rb
@@ -20,17 +20,18 @@ module Bolt
           @user = ENV['USER'] || Etc.getlogin
           @run_as = target.options['run-as']
           @logger = Logging.logger[self]
+          @sudo_id = SecureRandom.uuid
         end
 
         # If prompted for sudo password, send password to stdin and return an
         # empty string. Otherwise, check for sudo errors and raise Bolt error.
+        # If sudo_id is detected, that means the task needs to have stdin written.
         # If error is not sudo-related, return the stderr string to be added to
         # node output
-        def handle_sudo(stdin, err, pid)
+        def handle_sudo(stdin, err, pid, sudo_stdin)
           if err.include?(Sudoable.sudo_prompt)
             # A wild sudo prompt has appeared!
             if @target.options['sudo-password']
-              # Hopefully no one's sudo-password is > 64kb
               stdin.write("#{@target.options['sudo-password']}\n")
               ''
             else
@@ -39,6 +40,12 @@ module Bolt
                 'NO_PASSWORD'
               )
             end
+          elsif err =~ /^#{@sudo_id}/
+            if sudo_stdin
+              stdin.write("#{sudo_stdin}\n")
+              stdin.close
+            end
+            ''
           else
             handle_sudo_errors(err, pid)
           end
@@ -96,12 +103,12 @@ module Bolt
 
         # See if there's a sudo prompt in the output
         # If not, return the output
-        def check_sudo(out, inp, pid)
+        def check_sudo(out, inp, pid, stdin)
           buffer = out.readpartial(CHUNK_SIZE)
           # Split on newlines, including the newline
           lines = buffer.split(/(?<=[\n])/)
           # handle_sudo will return the line if it is not a sudo prompt or error
-          lines.map! { |line| handle_sudo(inp, line, pid) }
+          lines.map! { |line| handle_sudo(inp, line, pid, stdin) }
           lines.join("")
         # If stream has reached EOF, no password prompt is expected
         # return an empty string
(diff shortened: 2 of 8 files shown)


lib/bolt/transport/local/shell.rb
```diff
@@ -20,17 +20,19 @@ def initialize(target)
           @user = ENV['USER'] || Etc.getlogin
           @run_as = target.options['run-as']
           @logger = Logging.logger[self]
+          @sudo_id = SecureRandom.uuid
+          @send_stdin = true
```
donoghuc: I was not able to trigger data to be sent twice, but I left this in to guard against it. Can we safely remove it?

lib/bolt/transport/local/shell.rb
```diff
@@ -20,17 +20,19 @@ def initialize(target)
...
         # empty string. Otherwise, check for sudo errors and raise Bolt error.
+        # If sudo_id is detected, that means the task needs to have stdin written.
         # If error is not sudo-related, return the stderr string to be added to
         # node output
-        def handle_sudo(stdin, err, pid)
+        def handle_sudo(stdin, err, pid, sudo_stdin)
           if err.include?(Sudoable.sudo_prompt)
             # A wild sudo prompt has appeared!
             if @target.options['sudo-password']
-              # Hopefully no one's sudo-password is > 64kb
```
donoghuc: Based on my testing I think that a `write` here will handle any data size.

acceptance/tests/task_local.rb
```diff
@@ -38,9 +38,6 @@
       #!/bin/sh
       echo "$PT_greetings from $(whoami)"
       FILE
-      # TODO: Use input_method: both (default) once bug BOLT-1283 is fixed
```
donoghuc: The bug described in https://tickets.puppetlabs.com/browse/BOLT-1283 is solved by dropping the wrapper script entirely.

lib/bolt/transport/local/shell.rb
```diff
@@ -213,6 +223,15 @@ def execute(command, sudoable: true, **options)
           result_output.exit_code = t.value.exitstatus
           result_output
         end
+
+        def build_command(sudo_str, command_str, intput_stdin)
```
MikaelSmith: What's `intput`?

lib/bolt/transport/local/shell.rb
```diff
@@ -213,6 +223,15 @@ def execute(command, sudoable: true, **options)
...
+          if intput_stdin
+            "#{sudo_str} sh -c 'echo #{@sudo_id} 1>&2; #{command_str}'"
```
MikaelSmith: Why are these different for local vs ssh?

lib/bolt/transport/local/shell.rb
donoghuc: :rofl:

lib/bolt/transport/local/shell.rb
donoghuc: Oops. I worked on the ssh transport first, then when I got to the local I realized I needed the string on stderr. I think they can be the same and I can move this to a shared method in `Sudoable::Connection`.

lib/bolt/transport/local/shell.rb
lucywyman: I think this can only happen if `sudo randomchars` is in stdout twice, which would only happen if that *happened* to be output by a script that was running. That could happen before or after the actual sudo check though, so I don't think we gain anything from checking if it's been run already - if `sudo randomchars` is on stdout *not* from us, behavior is unexpected.

donoghuc: Also once stdin is sent we close that channel. Just to clarify we send and look for the `sudo_id` on stderr.

lib/bolt/transport/sudoable/connection.rb
```diff
@@ -70,6 +70,15 @@ def execute(*_args)
           message = "#{self.class.name} must implement #{method} to execute commands"
           raise NotImplementedError, message
         end
+
+        def build_command(sudo_str, command_str, sudo_id, input_stdin)
```
nicklewis: This is a fairly generic name to be on a mixin. It's clear in the context of this module, but once it's included in another class, it's not so obvious what its function is. It might be helpful to include the word `sudo` in there somewhere.

It might also make sense to split out the part that generates `sh -c '...'` from the part that puts `sudo_str` in front of it. Since the latter is just string concatenation, maybe the only thing this function needs to do is generate the `sh` invocation and let the caller handle prepending the run-as command.

lib/bolt/transport/sudoable.rb
```diff
@@ -118,16 +118,9 @@ def run_task(target, task, arguments, options = {})
...
-                # Inject interpreter in to wrapper script and remove from execute options
-                wrapper = make_wrapper_stringio(remote_task_path, stdin, execute_options[:interpreter])
-                execute_options.delete(:interpreter)
-                remote_wrapper_path = conn.write_executable(dir, wrapper, 'wrapper.sh')
-                command << remote_wrapper_path
-              else
-                command << remote_task_path
-                execute_options[:stdin] = stdin
-              end
+              command << remote_task_path
```
nicklewis: I think the `command` variable can go away entirely. It appears that now it starts out as an empty array and we put `remote_task_path` in it and nothing else.

lib/bolt/transport/ssh/connection.rb
```diff
@@ -135,7 +137,7 @@ def disconnect
           end
         end
 
-        def handled_sudo(channel, data)
+        def handled_sudo(channel, data, stdin = nil)
```
nicklewis: Is there a case where we _don't_ pass `stdin`? Should this be a required argument?

donoghuc: There is no case where we do not pass this param. It should be required.

lib/bolt/transport/sudoable/connection.rb
donoghuc: I added more descriptive names and separated out appending the command to echo the sudo_id before the command. I left both methods in the Sudoable::Connection class because the same logic is used for SSH::connection and Local::Shell.

donoghuc:
@nicklewis I punted on trying to get stdin to work with tty. This is ready for another review.

lib/bolt/transport/ssh/connection.rb
```diff
@@ -228,8 +231,8 @@ def execute(command, sudoable: false, **options)
               channel.on_request("exit-status") do |_, data|
                 result_output.exit_code = data.read_long
               end
-
-              if options[:stdin]
+              # A wrapper is used to direct stdin when elevating privilage or using tty
+              if options[:stdin] && !use_sudo && !options[:wrapper]
                 channel.send_data(options[:stdin])
```
donoghuc: I have a local change to do `"option[:stdin]\n"` but it does not seem to make a difference. Seems like it may be safer to do? Either way I will standardize with `handled_sudo`.

lib/bolt/transport/sudoable/connection.rb
```diff
@@ -70,6 +70,33 @@ def execute(*_args)
...
         end
+
+        def prepend_sudo_success(sudo_id, command_str)
+          "sh -c 'echo #{sudo_id} 1>&2; #{command_str}'"
+        end
+
+        # A helper to build up a single string that contains all of the options
+        # for privilage escalation.
+        def build_sudoable_command_str(command_str, sudo_str, sudo_id, options)
+          if options[:stdin] && !options[:wrapper]
```
nicklewis: This could definitely use a comment explaining why it works this way.

lib/bolt/transport/ssh/connection.rb
nicklewis: I think it only makes a difference when using a tty.

lib/bolt/transport/sudoable/connection.rb
donoghuc: Added some more detail.

donoghuc:
I agree that we could improve how me manage state. I added some tests to cover the case where a task expects large input for stdin and as environment variables when using sudo and tty.

donoghuc:
fixes https://github.com/puppetlabs/bolt/issues/1129

Issue #1131, opened 2019-08-07
Support remote state for the terraform inventory plugin


donoghuc:
## Use Case
I want to use a remote state storage solution with the Terraform inventory plugin. Currently the only supported option is a local statefile.

## Describe the Solution You Would Like
Configure the plugin to query remote state backend(s)

donoghuc:
Released in Bolt 1.29.0

PR #1142, opened 2019-08-09
(GH-1131) Add remote state support for Terraform plugin

beechtom:
This adds support for loading a remote Terraform state file using the
Terraform CLI. Previously, state could only be read from local
state files.

Additionally, inventory cofig options are validated to ensure only known
keys are used and that all required keys are present.

### DIFF
--- a/lib/bolt/plugin/terraform.rb
+++ b/lib/bolt/plugin/terraform.rb
@@ -5,6 +5,10 @@ require 'json'
 module Bolt
   class Plugin
     class Terraform
+      KNOWN_KEYS = Set['_plugin', 'dir', 'resource_type', 'uri', 'name', 'statefile',
+                       'config', 'backend']
+      REQ_KEYS = Set['dir', 'resource_type']
+
       def initialize
         @logger = Logging.logger[self]
       end
@@ -21,7 +25,25 @@ module Bolt
         @logger.warn("Could not find property #{property} of terraform resource #{name}")
       end
 
+      # Make sure no unexpected keys are in the inventory config and
+      # that required keys are present
+      def validate_options(opts)
+        opt_keys = opts.keys.to_set
+
+        unless KNOWN_KEYS.superset?(opt_keys)
+          keys = opt_keys - KNOWN_KEYS
+          raise Bolt::ValidationError, "Unexpected key(s) in inventory config: #{keys.to_a.inspect}"
+        end
+
+        unless opt_keys.superset?(REQ_KEYS)
+          keys = REQ_KEYS - opt_keys
+          raise Bolt::ValidationError, "Expected key(s) in inventory config: #{keys.to_a.inspect}"
+        end
+      end
+
       def inventory_targets(opts)
+        validate_options(opts)
+
         state = load_statefile(opts)
 
         resources = extract_resources(state)
@@ -49,11 +71,32 @@ module Bolt
       end
 
       def load_statefile(opts)
+        statefile = if opts['backend'] == 'remote'
+                      load_remote_statefile(opts)
+                    else
+                      load_local_statefile(opts)
+                    end
+
+        JSON.parse(statefile)
+      end
+
+      # Uses the Terraform CLI to pull remote state files
+      def load_remote_statefile(opts)
+        stdout_str, stderr_str, = Open3.capture3('terraform state pull', chdir: opts['dir'])
+
+        unless stderr_str.empty?
+          err = stdout_str.split("\n").first
+          msg = "Could not pull Terraform remote state file for #{opts['dir']}: #{err}"
+          raise Bolt::Error.new(msg, 'bolt/terraform-state-error')
+        end
+
+        stdout_str
+      end
+
+      def load_local_statefile(opts)
         dir = opts['dir']
         filename = opts.fetch('statefile', 'terraform.tfstate')
-        statefile = File.expand_path(File.join(dir, filename))
-
-        JSON.parse(File.read(statefile))
+        File.read(File.expand_path(File.join(dir, filename)))
       rescue StandardError => e
         raise Bolt::FileError.new("Could not load Terraform state file #{filename}: #{e}", filename)
       end
--- a/lib/bolt/plugin/terraform.rb
+++ b/lib/bolt/plugin/terraform.rb
@@ -111,8 +111,10 @@ module Bolt
             resource_set['instances'].map do |resource|
               instance_name = prefix
               instance_name += ".#{resource['index_key']}" if resource['index_key']
-
-              [instance_name, resource['attributes']]
+              # When using `terraform state pull` with terraform >= 0.12 version 3 statefiles
+              # Will be converted to version 4. When converted attributes is converted to attributes_flat
+              attributes = resource['attributes'] || resource['attributes_flat']
+              [instance_name, attributes]
             end
           end
         else
--- a/lib/bolt/plugin/terraform.rb
+++ b/lib/bolt/plugin/terraform.rb
@@ -82,11 +82,22 @@ module Bolt
 
       # Uses the Terraform CLI to pull remote state files
       def load_remote_statefile(opts)
-        stdout_str, stderr_str, = Open3.capture3('terraform state pull', chdir: opts['dir'])
+        dir = File.expand_path(opts['dir'])
 
-        unless stderr_str.empty?
-          err = stdout_str.split("\n").first
-          msg = "Could not pull Terraform remote state file for #{opts['dir']}: #{err}"
+        begin
+          stdout_str, stderr_str, status = Open3.capture3('terraform state pull', chdir: dir)
+        rescue Errno::ENOENT
+          reason = if File.directory?(dir)
+                     "Could not find executable 'terraform'"
+                   else
+                     "Could not find directory '#{dir}'"
+                   end
+          raise Bolt::Error.new(reason, 'FILE_ERROR')
+        end
+
+        unless status.success?
+          err = stdout_str + stderr_str
+          msg = "Could not pull Terraform remote state file for #{opts['dir']}:\n#{err}"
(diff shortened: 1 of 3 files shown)


lib/bolt/plugin/terraform.rb
```diff
@@ -49,11 +68,37 @@ def inventory_targets(opts)
...
+
+      # Uses the Terraform CLI to pull remote state files
+      def load_remote_statefile(opts)
         dir = opts['dir']
-        filename = opts.fetch('statefile', 'terraform.tfstate')
-        statefile = File.expand_path(File.join(dir, filename))
+        cwd = Dir.pwd
+
+        Dir.chdir(dir)
+        stdout_str, stderr_str, = Open3.capture3('terraform state pull', binmode: true)
```
donoghuc: Check out the `chdir` option for capture3 https://www.rubydoc.info/stdlib/core/Process.spawn . You can do something like: `Open3.capture3('terraform state pull', chdir: opts['dir'])`. Also, what is the `:binmode` option used for here?

pre-docs/inventory_version2.md
```diff
@@ -312,12 +312,13 @@ groups:
 The Terraform plugin supports looking up target object from a Terraform state
 file. It accepts several fields:
 
-`dir`: The directory from which to load Terraform state
+`dir`: The directory containing either a local Terraform state file or Terraform configuration to read remote state from
 `resource_type`: The Terraform resources to match, as a regular expression
 `uri`: The property of the Terraform resource to use as the target URI (optional)
-`statefile`: The name of the Terraform state file to load within `dir` (optional, defaults to `terraform.tfstate`)
+`statefile`: The name of the local Terraform state file to load within `dir` (optional, defaults to `terraform.tfstate`)
```
donoghuc: Can we note that this is *always* relative to the `dir` and that absolute paths will not work.

lib/bolt/plugin/terraform.rb
```diff
@@ -21,7 +21,26 @@ def warn_missing_property(name, property)
         @logger.warn("Could not find property #{property} of terraform resource #{name}")
       end
 
+      # Make sure no unexpected keys are in the inventory config and
+      # that required keys are present
+      def validate_options(opts)
```
donoghuc: Could also use `Set` logic here to check keys instead of iterating over lists. We do that for yaml validation. https://github.com/puppetlabs/bolt/blob/c049c3609ac2d3af88b9f772cdfac93d4950485d/lib/bolt/pal/yaml_plan.rb#L34-L38

donoghuc:
This might not be an issue for "real" users, but i was thinking it might be nice to expose an option for `terraform-executable`, that way when testing between pre/post 0.12 you could set which cli executable to use instead of modifying `PATH`.

lib/bolt/plugin/terraform.rb
beechtom: Updated.

I had put in `binmode` based on Glenn's use of `capture3` in the docker CLI pull: https://github.com/puppetlabs/bolt/pull/1075/files#diff-7bd8cc88b6b442a98afb7f1cef570a6aR173

I'm not sure if there would be a concern with Windows compatibility / line ending translations here without binary mode being set.

lucywyman:
Relates to #1131

beechtom:
If we do have a `terraform-executable` field, would we want this to be undocumented since it'd be primarily for testing?

donoghuc:
I think we would *only* build in the `terraform-executable` option if we thought users would want it. We depend on ruby being on the path to execute bolt_catalog https://github.com/puppetlabs/bolt/blob/da7ab6e82a979eba1c2fc13955467c2aba3c40ca/lib/bolt/applicator.rb#L104 . Maybe we just expect the user to have their path configured to use the version they want if they have multiple.

lib/bolt/plugin/terraform.rb
```diff
@@ -49,11 +71,32 @@ def inventory_targets(opts)
       end
 
       def load_statefile(opts)
+        statefile = if opts['backend'] == 'remote'
```
nicklewis: Do we need the `load_local_statefile` path at all or can we just always use `terraform state pull`, which should load from the local state if that's what's in use?

nicklewis:
I would prefer to err on the side of waiting for someone to request a `terraform-executable` field.

lib/bolt/plugin/terraform.rb
donoghuc: Confirmed that terraform pull appears to work even when a remote backend is not used with both 0.11 and 0.12. Could there be a case where the user specifies a `statefile` that the CLI would not be configured to "pull" from?

beechtom: As far as I can tell, `terraform state pull` will only read a local file when it's named `terraform.tfstate`. So if a user had a state file with a different name then the plugin couldn't read from it. Not sure if this is a concern considering terraform itself won't read from these files.

donoghuc: I'm also investigating another bug. Basically if you have a statefile that is version 3 (presumably generated by terrafomr < 0.12) and execute `terraform state pull` with terraform > 0.12 you get a `version: 4` state file that does not actually work with our extract_resources method. So unless I find an easy fix for that case, this could be another reason to still support parsing a statefile without using the CLI.

lib/bolt/plugin/terraform.rb
```diff
@@ -111,8 +111,10 @@ def extract_resources(state)
             resource_set['instances'].map do |resource|
               instance_name = prefix
               instance_name += ".#{resource['index_key']}" if resource['index_key']
-
-              [instance_name, resource['attributes']]
+              # When using `terraform state pull` with terraform >= 0.12 version 3 statefiles
+              # Will be converted to version 4. When converted attributes is converted to attributes_flat
+              attributes = resource['attributes'] || resource['attributes_flat']
+              [instance_name, attributes]
```
donoghuc: Snippet from version 3 statefile generated with terraform 0.11:
```
            "resources": {
                "aws_instance.xenial.0": {
                    "type": "aws_instance",
                    "depends_on": [],
                    "primary": {
                        "id": "i-08594d1edf25dca11",
                        "attributes": {
```
Snippet from result of `terraform state pull` using terraform 0.12:
```
  "resources": [
    {
      "mode": "managed",
      "type": "aws_instance",
      "name": "xenial",
      "each": "list",
      "provider": "provider.aws",
      "instances": [
        {
          "index_key": 0,
          "schema_version": 1,
          "attributes_flat": {
            "ami": "ami-076e276d85f524150",
```

lib/bolt/plugin/terraform.rb
```diff
@@ -49,11 +71,32 @@ def inventory_targets(opts)
...
+                    else
+                      load_local_statefile(opts)
+                    end
+
+        JSON.parse(statefile)
+      end
+
+      # Uses the Terraform CLI to pull remote state files
+      def load_remote_statefile(opts)
+        stdout_str, stderr_str, = Open3.capture3('terraform state pull', chdir: opts['dir'])
```
donoghuc: Would it be worth wrapping this in a begin/rescue block and raising a bolt error if terraform executable is not on PATH? Also is there anything useful we would want to do with the exit code?

beechtom: Added a begin/rescue block - good catch. Also modified the error after the block to read the status code instead of `stderr`.

lib/bolt/plugin/terraform.rb
```diff
@@ -82,9 +82,18 @@ def load_statefile(opts)
 
       # Uses the Terraform CLI to pull remote state files
       def load_remote_statefile(opts)
-        stdout_str, stderr_str, = Open3.capture3('terraform state pull', chdir: opts['dir'])
+        dir = File.expand_path(opts['dir'])
+
+        begin
+          stdout_str, _stderr_str, status = Open3.capture3('terraform state pull', chdir: dir)
+        rescue StandardError => e
```
donoghuc: I think we can rescue an `Errno::ENOENT` and assume it is due to no terraform executable on the path. The error message can be something to that effect. 

I am not sure what kinds of status's or errors that `terraform state pull` can give. I was just using the "local" backend:
```
terraform {
  backend "local" {
    path = "test/terraform_local.tfstate"
  }
}
```
Did you have an s3 backend set up? What happens if there is a connection or auth errror?

beechtom: I do. If there's a connection or auth error then terraform raises an error that we catch after the rescue block and also writes to `stderr`.

beechtom: Rescuing `Errno::ENOENT` works fine. The only issue with saying it's due to no terraform executable on the path is that the same kind of error is raised when the path to the `dir` does not exist. Would we want to validate the `dir` before attempting to load the state file in that case?

lib/bolt/plugin/terraform.rb
```diff
@@ -49,11 +71,41 @@ def inventory_targets(opts)
...
+          raise Bolt::Error.new(
+            "Failed to load remote state file: #{e.message}",
+            'FILE_ERROR'
+          )
+        end
+
+        unless status.success?
+          err = stdout_str.split("\n").first
```
beechtom: Depending on any errors that terraform raises, it might print to one or both of `stdout` and `stderr`. Since these errors can be raised for a couple reasons (terraform hasn't been initialized, missing files, etc.) would it make sense to display all of the text that terraform prints to `stdout` and `stderr`, or just have a generic message saying there was an issue with the terraform configuration and to sort that out separately?

donoghuc: You could grep the dir name in the error message. But i guess if the dirname was terraform that might be confusing. In the rescue block you could just do `File.directory?(dir)` and change the message based on that.

donoghuc: I guess that only addresses the first part of the question. I think surfacing the error message terraform displays is important. We just want to be as helpful as possible.

PR #1143, opened 2019-08-10
Add homepage for developer blog on website

beechtom:
This adds a homepage that lists all developer blog posts in a single
location with short excerpts. It also lists posts in reverse
chronological order in the sidebar so new posts will be visible.

### DIFF
--- a/docs/_layouts/default.html
+++ b/docs/_layouts/default.html
@@ -52,11 +52,12 @@
           <li>
             <span class="nav-header">Tips &amp; Tricks</span>
             <ul>
-              <li><a href="{{ '/debugging.html' | relative_url }}">Debugging common Bolt errors</a></li>
+              <li><a href="{{ '/debugging' | relative_url }}">Debugging common Bolt errors</a></li>
             </ul>
           </li>
           <li>
-            <span class="nav-header">Bolt Developer Updates</span>
+            <span class="nav-header">
+              <a href="{{ '/blog' | relative_url }}" class="center-vertical">Bolt Developer Updates</a></span>
             <ul>
               {% for post in site.posts limit: 3 %}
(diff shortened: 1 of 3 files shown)


docs/_layouts/blog.html
```diff
@@ -0,0 +1,16 @@
+---
```
nicklewis: Why is this a layout and not just `docs/blog.html`?

docs/_layouts/default.html
```diff
@@ -56,9 +56,10 @@
             </ul>
           </li>
           <li>
-            <span class="nav-header">Bolt Developer Updates</span>
+            <span class="nav-header">
+              <a href="{{ '/blog' | relative_url }}" class="center-vertical">Bolt Developer Updates</a></span>
             <ul>
-              {% for post in site.posts limit: 3 %}
+              {% for post in site.posts reversed limit: 3 %}
```
nicklewis: I don't think we want `reversed`. It seems like `posts` already sort from newest to oldest.

docs/_layouts/blog.html
beechtom: Updated.

docs/_layouts/default.html
```diff
@@ -56,7 +56,8 @@
             </ul>
           </li>
           <li>
-            <span class="nav-header">Bolt Developer Updates</span>
+            <span class="nav-header">
+              <a href="{{ '/blog.html' | relative_url }}" class="center-vertical">Bolt Developer Updates</a></span>
```
nicklewis: The `.html` part doesn't seem to be necessary. I'm not sure I understand why though...

beechtom: Weird. Updated and also removed `.html` from the link to the debugging page while I was at it.

PR #1144, opened 2019-08-12
Toggle target count in inventory show with verbose logging

beechtom:
This allows users to toggle the target count when using the `inventory
show` command and printing in human format. Previously, the target count
would always display, making it more challenging to use the command's
output in scripts. The target count will only display when verbose
logging is set.

### DIFF
--- a/lib/bolt/outputter/human.rb
+++ b/lib/bolt/outputter/human.rb
@@ -315,7 +315,7 @@ module Bolt
         targets = options[:targets].map(&:name)
         count = "#{targets.count} target#{'s' unless targets.count == 1}"
         @stream.puts targets.join("\n")
-        @stream.puts colorize(:green, count)
+        @stream.puts colorize(:green, count) if @verbose
       end
 
       # @param [Bolt::ResultSet] apply_result A ResultSet object representing the result of a `bolt apply`


beechtom:
This PR is in response to a user who requested the ability to have the target count turned off. The `inventory show` command has verbose logging set by default in the cli. Would it make sense to have it not set for this command, or just require users who want the count disabled to use the `--no-verbose` option?

lucywyman:
I think if they want to use the output in scripts they can use `--format json` and get the `targets` list there.

beechtom:
It was a specific request to be able to use both the json and human format in scripts: https://tickets.puppetlabs.com/browse/BOLT-1384

nicklewis:
This command defaults to verbose, so this doesn't change the behavior unless you explicitly specify `--no-verbose`. That part seems safe.

The concern I have with this is that it sort of inverts the typical meaning we have for verbose. Specifically, verbose usually means show target-specific info and no-verbose means only show aggregate info. In this case, no-verbose means _don't_ show the aggregate info. The consistent (although not necessarily correct) behavior seems like it would be for verbose to show the list of targets and for no-verbose to _only_ show the count.

I wonder whether including the count is actually helpful in the first place...

lucywyman:
I guess that's a separate question. I want to know more about Corey's use case before making this change just for him. It's hard for me to know if / how people are using it without asking.

beechtom:
Should we close this for now and look at the possibility of a `--format oneline` option?

lucywyman:
I don't think we need to close this, but I also don't see us designing or prioritizing a `--oneline` option soon.

adreyer:
I think we're likely to break parsing of this command in the future so it's probably not a good idea to encourage people to parse it like this.

PR #1150, opened 2019-08-14
Add Atom feed to website

beechtom:
This adds support for an Atom feed for the Bolt Developer Updates blog  using the
jekyll-feed plugin.


nicklewis:
This is awesome! The commit subject says RSS but the body says Atom. Looks like it is in fact an Atom feed.

Issue #1148, opened 2019-08-13
Add Azure inventory plugin


beechtom:
## Use Case
Users should be able to run Bolt commands on Azure virtual machines and virtual machine scale sets without manually adding them to the inventory.

## Describe the Solution You Would Like
Bolt should have an inventory plugin that gets targets from a list of virtual machines and scale sets on an Azure account. 

Users should be able to filter VMs and scale sets based on the following:

- Resource group
- Scale set
- Location
- Tags

The `name` and `uri` for a target should be set to the FQDN and public IP address, if available.

```
targets:
   - _plugin: azure
     location: eastus
     resource-group: my-group
     scale-set: my-set
     tags:
        key: value
```

It should be possible to set configuration for authorizing with Azure in the following locations:

- inventory config
- Bolt config
- environment variables (e.g. `ENV['AZURE_CLIENT_ID']`)
- a credentials file at `~/.azure/credentials`

```
plugins:
   azure:
      tenant-id: xxxx-xxxx
      client-id: xxxx-xxxx
      client-secret: xxxx-xxxx
      subscription-id: xxxx-xxxx
```

## Questions

- Should the `name` and `uri` always be set to the IP address / FQDN by default, or can these be overridden? This can be tricky since values in Azure's REST API are nested and might be difficult for the user to specify. Is there a need to specify other properties?
- Should VMs and scale sets be targeted separately? The above inventory config would load all VMs and scale sets that match the filters.
- If a credentials file is desired, should it be a `.ini` file? This is the format that the Azure SDKs use.

PR #1156, opened 2019-08-15
(GH-1148) Add Azure inventory plugin

beechtom:
This adds support for looking up virtual machines and virtual machine
scale sets on Azure. VMs can be filtered by location, resource group,
and tags.

### DIFF
--- a/lib/bolt/plugin/azure.rb
+++ b/lib/bolt/plugin/azure.rb
@@ -1,0 +1,199 @@
+# frozen_string_literal: true
+
+require 'json'
+
+module Bolt
+  class Plugin
+    class Azure
+      class AzureHTTPError < Bolt::Error
+        def initialize(response)
+          err = JSON.parse(response.body).dig('error', 'message')
+          m = String.new("#{response.code} \"#{response.msg}\"")
+          m += ": #{err}" if err
+          super(m, 'bolt.plugin/azure-http-error')
+        end
+      end
+
+      attr_reader :config
+
+      CONFIG_KEYS = Set['tenant_id', 'client_id', 'client_secret', 'subscription_id', 'profile']
+      OPTS_KEYS = Set['_plugin', 'location', 'tags', 'resource_group', 'scale_set'] + CONFIG_KEYS
+
+      def initialize(config)
+        validate_config(config)
+        @config = config
+        @logger = Logging.logger[self]
+      end
+
+      def name
+        'azure'
+      end
+
+      def hooks
+        %w[inventory_targets]
+      end
+
+      def validate_config(config)
+        keys = config.keys.to_set
+
+        unless CONFIG_KEYS.superset?(keys)
+          keys -= CONFIG_KEYS
+          raise Bolt::ValidationError, "Unexpected key(s) in plugin config: #{keys.to_a.inspect}"
+        end
+      end
+
+      def validate_options(opts)
+        keys = opts.keys.to_set
+
+        unless OPTS_KEYS.superset?(keys)
+          keys -= OPTS_KEYS
+          raise Bolt::ValidationError, "Unexpected key(s) in inventory config: #{keys.to_a.inspect}"
+        end
+      end
+
+      def inventory_targets(opts)
+        validate_options(opts)
+        creds = credentials(opts)
+        token = token(creds)
+        instances = instances(token, creds, opts)
+
+        # Filter by location
+        if opts['location']
+          instances.select! do |instance|
+            instance['location'] == opts['location']
+          end
+        end
+
+        # Filter by tags - tags are ANDed
+        if opts['tags']
+          instances.select! do |instance|
+            instance['tags'] && opts['tags'] <= instance['tags']
+          end
+        end
+
+        # Map the VMs to a list of targets
+        # name is set to the FQDN while uri is set to the public ip address
+        # Any VMs that have neither of these set are dropped
+        instances.map do |instance|
+          target = {
+            'name' => instance.dig('properties', 'dnsSettings', 'fqdn'),
+            'uri' => instance.dig('properties', 'ipAddress')
+          }.compact
+
+          target unless target.empty?
+        end.compact
+      end
+
+      # Hash of required credentials for authorizing with the Azure REST API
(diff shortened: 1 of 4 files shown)


beechtom:
Still writing tests and will need editing once the plugin refactoring is finished.

lib/bolt/plugin/azure.rb
```diff
@@ -0,0 +1,200 @@
...
+      def inventory_targets(opts)
+        validate_options(opts)
+        creds = credentials(opts)
+        token = token(creds)
+        instances = instances(token, creds, opts)
+
+        # Filter by location
+        if opts['location']
+          instances.select! do |instance|
+            instance['location'].eql? opts['location']
```
nicklewis: Is there a reason this is `eql?` instead of `==`?

pre-docs/inventory_version2.md
```diff
@@ -464,6 +464,67 @@ plugins:
...
+
+- `_plugin`: The value of `_plugin` must be `azure`
+- `resource_group`: The resource group to filter by (optional)
+- `scale_set`: The scale set to filter by (optional, requires that `resource_group` also be set)
+- `location`: The location to filter by (optional)
+- `tags`: A list of one or more tags to filter by (optional, tags are `name: value` pairs and are ANDed together)
+
+Accessing Azure resources requires credentials for signing all API requests. Each credential can be set in the inventory config, Bolt config, or as an environment variable, and are searched in this order until a value is found.
+
+- `tenant_id`: The Azure AD tenant ID (optional, defaults to `ENV['AZURE_TENANT_ID']`)
```
nicklewis: These don't have to be specified where you _invoke_ the plugin, but you have to have a value for all of them set _somewhere_. Does that mean they're optional or not?

pre-docs/inventory_version2.md
```diff
@@ -464,6 +464,67 @@ plugins:
...
 
+#### Azure
+
+The Azure plugin supports looking up virtual machines and virtual machine scale sets. It supports several fields:
+
+- `_plugin`: The value of `_plugin` must be `azure`
+- `resource_group`: The resource group to filter by (optional)
+- `scale_set`: The scale set to filter by (optional, requires that `resource_group` also be set)
+- `location`: The location to filter by (optional)
+- `tags`: A list of one or more tags to filter by (optional, tags are `name: value` pairs and are ANDed together)
```
nicklewis: Might be clearer to say instances must match all the listed tags.

lib/bolt/plugin/azure.rb
```diff
@@ -0,0 +1,200 @@
...
+                "providers/Microsoft.Network/publicIPAddresses?api-version=2019-06-01"
+              end
+
+        header = {
+          'Authorization' => "#{token['token_type']} #{token['access_token']}"
+        }
+
+        instances = []
+
+        loop do
```
nicklewis: This could just be `while url`, then the `break` at the end is unnecessary.

lib/bolt/plugin/azure.rb
```diff
@@ -0,0 +1,200 @@
...
+          # Update the URI and make the next request
+          uri = URI.parse(url)
+          result = request(:Get, uri, nil, header)
+
+          # Add the VMs to the list of instances
+          instances += result['value']
```
nicklewis: Prefer `<<` over `+=`. `<<` inserts into the array, `+=` makes a copy.

lib/bolt/plugin/azure.rb
```diff
@@ -0,0 +1,200 @@
...
+        request(:Post, uri, data)
+      end
+
+      def request(verb, uri, data, header = {})
+        # Create the client
+        client = Net::HTTP.new(uri.host, uri.port)
+
+        # Azure REST API always uses SSL
+        client.use_ssl = true
+        client.verify_mode = OpenSSL::SSL::VERIFY_NONE
```
nicklewis: Is there a reason we're using unverified SSL?

lib/bolt/plugin/azure.rb
```diff
@@ -0,0 +1,200 @@
...
+      # https://docs.microsoft.com/en-us/azure/active-directory/develop/v1-oauth2-client-creds-grant-flow
+      def token(creds)
+        data = {
+          grant_type: 'client_credentials',
+          client_id: creds['client_id'],
+          client_secret: creds['client_secret'],
+          resource: 'https://management.azure.com'
+        }
+
+        uri = URI.parse("https://login.microsoftonline.com/#{creds['tenant_id']}/oauth2/token")
```
nicklewis: Using an oauth workflow seems strange here. Does this require the user to create a new "application" in azure?

pre-docs/inventory_version2.md
beechtom: The credentials themselves aren't optional, but don't necessarily have to be set in the inventory/config. I can remove the 'optional' part if it makes it clearer.

lib/bolt/plugin/azure.rb
beechtom: Yep. It's a simple process, but creating a client application is the only way to obtain tokens to authorize with the REST API. Process is covered in the docs for the API: https://docs.microsoft.com/en-us/rest/api/azure/

lib/bolt/plugin/azure.rb
beechtom: Oops. Just a force of habit from grading/writing projects in C... Updated. :)

lib/bolt/plugin/azure.rb
```diff
@@ -0,0 +1,199 @@
...
+
+        instances = []
+
+        while url do
+          # Update the URI and make the next request
+          uri = URI.parse(url)
+          result = request(:Get, uri, nil, header)
+
+          # Add the VMs to the list of instances
+          instances << result['value']
```
nicklewis: Sorry! I was mistaken. This should be `concat` (because `result['value']` is an array). Then the `flatten` call is unnecessary.

nicklewis:
Closing in favor of https://github.com/puppetlabs/puppetlabs-azure_inventory/pull/1

PR #1157, opened 2019-08-16
(GH-1123) Allow users to configure apply_prep

### LINKED ISSUE #1123: Allow users to configure how puppet libraries are managed for apply_prep
# Problem
Currently the `apply_prep` method calls the `puppet_agent::install` task to install the agent. Instead the behavior for this should be configurable via plugins.

### Test Cases

- [X] Configure the `task` plugin to use the `puppet_agent::install` plugin with different arguments.
- [x] Configure the `task` plugin to use the `bootstrap` task to install the agent.
- [x] ~~Use run_as root with the `puppet_agent::install` task.~~ #1191

### New `plugin_hooks` section in bolt.yaml and inventory
plugin_ hook allows users to configure what hooks are used on a per node basis. This section is an object where the keys are hooknames and the value specifies and configures the plugin to use. This key should be set on a target or a group. It should use the same merging strategy(shallow merge) as `vars` does. 

inventory example
```
---
targets:
   - name: foo
      plugin_hooks:
        puppet_library:
          plugin: install_agent
```

config example
```
---
modulepath: [~/modules]
plugin_hooks:
  puppet_library
    plugin: install_agent
```

### puppet_library plugin hook
When the puppet library plugin hook is configured the apply_prep function will run the plugins hook rather then using the builtin logic.

### Task plugin hook for `puppet_library`

In order to let users write their own task or use an existing task we should add a `puppet_library` hook handler to the task plugin.

```
plugin_hooks:
  puppet_library:
    plugin: task
    task: 'bootstrap'
    parameters:
      master: 'puppet.example.com'
      cacert_content: <cert>
``` 

### Expose existing behavior as a plugin

The existing behavior should be exposed as the `install_agent` plugin written in ruby. It should take no configuration for now.

## Questions
- Should bolt always use the `puppet_agent::version` task before running the `puppet_libraries` hook? Otherwise all tasks that would be used would need to optionally check for an agent which the examples don't do currently. Should we expose a second hook?
- Once one library hook is successful should we call the hook again for that target?

lucywyman:
Closes #1123

This commit allows users to configure how to install the puppet agent on
remote systems when preparing nodes to have Puppet code applied. It adds
a new `plugin_hooks` configuration which sets which plugin to use when a
specific hook is called. This is how users can specify using either the
`install_agent` builtin plugin to install the agent, or the `task`
plugin so they can write their own task to install the agent. The agent
in installed by calling the `puppet_library` hook included in this
commit and defined by the plugin. Lastly worth noting: the
`install_agent` plugin exposes the old behavior of the `apply_prep`
function as a plugin.


lib/bolt/plugin/task.rb
```diff
@@ -33,9 +33,9 @@ def inventory
         @inventory ||= Bolt::Inventory.new({}, config)
       end
 
-      def run_task(opts)
+      def run_task(opts, target = 'localhost')
         result = pal.run_task(opts['task'],
-                              'localhost',
+                              target,
```
adreyer: I'm concerned that a Target object bound to one Inventory instance is being used with a different Inventory instance here.

bolt-modules/boltlib/lib/puppet/functions/apply_prep.rb
```diff
@@ -70,12 +74,28 @@ def apply_prep(target_spec)
 
           unless need_install.empty?
             need_install_targets = need_install.map(&:target)
-            run_task(executor, need_install_targets, 'puppet_agent::install')
-            # Service task works best when targets have puppet-agent feature
+
+            plugin_targets, builtin_targets = need_install_targets.partition do |t|
```
adreyer: We should just use the the `InstallAgent` Plugin as a default in config rather then treating these nodes specially.

bolt-modules/boltlib/lib/puppet/functions/apply_prep.rb
```diff
@@ -70,12 +74,18 @@ def apply_prep(target_spec)
...
           unless need_install.empty?
             need_install_targets = need_install.map(&:target)
-            run_task(executor, need_install_targets, 'puppet_agent::install')
-            # Service task works best when targets have puppet-agent feature
+
+            need_install_targets.each do |t|
+              plugin_name = t.plugin_hooks&.fetch('puppet_library')&.fetch('plugin') || 'install_agent'
+              hook = inventory.plugins.get_hook(plugin_name, :puppet_library)
+              begin
+                hook.call(t, t.plugin_hooks['puppet_library'])
```
adreyer: Am I missing something or will this will be a serious performance regression since we'll no longer parallelize the agent installation.

pre-docs/bolt_configuration_options.md
```diff
@@ -225,3 +225,23 @@ The `puppetfile` section configures how modules are retrieved when running `bolt
 `proxy`: The HTTP proxy to use for Git and Puppet Forge operations
 
 `forge`: A subsection which can have its own `proxy` setting to set an HTTP proxy for only Puppet Forge operations, and a `baseurl` setting to specify a different Forge host
+
+## Plugin hooks configuration options
```
adreyer: maybe mention that `puppet_library` is the only configurable hook now though we'll likely to forget to keep that up to date.

I think most of the specifics about how this hook works should be in Install Libaries section.

bolt-modules/boltlib/lib/puppet/functions/apply_prep.rb
```diff
@@ -69,13 +72,36 @@ def apply_prep(target_spec)
           end
 
           unless need_install.empty?
+            # lazy-load expensive gem code
+            require 'concurrent'
+            pool = Concurrent::ThreadPoolExecutor.new(max_threads: executor.concurrency)
```
adreyer: I don't think we need to limit the concurrency here assuming we run these through the correct executor instance.

pre-docs/bolt_configuration_options.md
```diff
@@ -225,3 +225,25 @@ The `puppetfile` section configures how modules are retrieved when running `bolt
...
+For now, the only configurable plugin hook is the `puppet_library` plugin hook.
+
+```
+plugin_hooks:
+  puppet_library:
+    plugin: install_agent
+```
+
+```
+plugin_hooks:
```
adreyer: We usually prefer hyphens in the inventory file. That may be a bit confusing since the plugin names are `_`

bolt-modules/boltlib/lib/puppet/functions/apply_prep.rb
```diff
@@ -70,12 +73,23 @@ def apply_prep(target_spec)
...
             need_install_targets = need_install.map(&:target)
-            run_task(executor, need_install_targets, 'puppet_agent::install')
-            # Service task works best when targets have puppet-agent feature
+            # lazy-load expensive gem code
+            require 'concurrent'
+            pool = Concurrent::ThreadPoolExecutor.new
+
+            futures = need_install_targets.map do |t|
+              opts = t.plugin_hooks&.fetch('puppet_library')
+              inventory.plugins.get_hook(opts['plugin'], :load_task).call(opts)
```
adreyer: if this fails any previously started tasks are orphaned. either we need to do this(and get the `puppet_library`) in a separate loop from creating the future or we need to wrap it in a `begin`/`rescue` and make a failed result for each of these.

adreyer: We also probably need to catch the error here and either ignore it, assuming the plugin doesn't have to prep anything in serial.  Otherwise this generates a confusing error about the `load_task` hook which the user has never heard about.

pre-docs/inventory_version2.md
```diff
@@ -230,6 +230,35 @@ targets = data[params['environment']][params['app']]
 json.dump({'targets': targets}, sys.stdout)
 ```
 
+##### Install Library Tasks
+
+To install a library on the remote system (ie. the Puppet agent). This task
+will be executed when `apply_prep()` is called from a plan to provision
+targets.
+
+For example, this task installs the Puppet agent from a custom repository, then
```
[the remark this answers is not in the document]
adreyer: You can also use an existing task for the Task plugin. A common use case is going to be passing a specific argument to the `puppet::agent` task to control what version to install for example. We might want a PE example with the bootstrap task too. 

I think we need to document what the inventory file that configures a task plugin looks like. We probably also should document the `puppet_agent` default plugin.

adreyer:
>Use run_as root with the puppet_agent::install task.

I'm not sure how to do that with this PR. I think we may need to add a new paramater `run_as` to the `puppet_library` hook of the task plugin

bolt-modules/boltlib/lib/puppet/functions/apply_prep.rb
```diff
@@ -70,12 +73,47 @@ def apply_prep(target_spec)
...
+            futures = ok_hooks.map do |hash|
+              Concurrent::Future.execute(executor: pool) do
+                hash['hook'].call(hash['target'],
+                                  hash['target'].plugin_hooks&.fetch('puppet_library'),
+                                  self)
+              end
+            end
+
+            results = futures.zip(ok_hooks).map do |f, hash|
+              if f.value
```
adreyer: I think you can just use "`.from exception` here.  `f.value || Bolt::Result.from_exception(hash['target'], f.reason)`

lucywyman:
Verified this works with the bootstrap task:
```
targets:
 - uri: target.uri
plugin_hooks:
  puppet_library:
    plugin: task
    task: bootstrap
    parameters:
      master: master.fqdn
```

pre-docs/bolt_configuration_options.md
```diff
@@ -225,3 +225,25 @@ The `puppetfile` section configures how modules are retrieved when running `bolt
 `proxy`: The HTTP proxy to use for Git and Puppet Forge operations
 
 `forge`: A subsection which can have its own `proxy` setting to set an HTTP proxy for only Puppet Forge operations, and a `baseurl` setting to specify a different Forge host
+
+## Plugin hooks configuration options
+
+The `plugin_hooks` section allows you configure what plugins are used for a hook on a per target basis. This section is a hash where the keys are hooknames and the value specifies and configures the plugin to use. Currently there are two plugins you'll want to use for this: your own custom task using the `task` plugin, or the `install_agent` plugin which runs the `puppet_agent::install` task.
```
cwcowellshah:
```suggestion
The `plugin_hooks` section allows you to configure what plugins a specific hook should use for a specific target. This section is a hash where keys are hook names, and values specify and configure the plugin to use. Currently there are two possible plugins: the `task` plugin runs your own custom task, and the `install_agent` plugin runs the `puppet_agent::install` task.
```

pre-docs/bolt_configuration_options.md
```diff
@@ -225,3 +225,25 @@ The `puppetfile` section configures how modules are retrieved when running `bolt
...
+
+For now, the only configurable plugin hook is the `puppet_library` plugin hook.
```
cwcowellshah:
```suggestion
For now, the only configurable plugin hook is `puppet_library`.
```

pre-docs/inventory_version2.md
```diff
@@ -230,6 +230,77 @@ targets = data[params['environment']][params['app']]
 json.dump({'targets': targets}, sys.stdout)
 ```
 
+##### Install Library Tasks
```
cwcowellshah:
```suggestion
##### Install library tasks
```

pre-docs/inventory_version2.md
```diff
@@ -230,6 +230,77 @@ targets = data[params['environment']][params['app']]
...
+
+To install a library on the remote system (ie. the Puppet agent). This task
```
cwcowellshah:
```suggestion
To install a library on the remote system (the Puppet agent). This task
```

pre-docs/plan_functions.md
```diff
@@ -59,14 +59,17 @@ add_to_group('foo,bar,baz', 'group1')
 
 ## apply_prep
 
-Installs the puppet-agent package on targets if needed then collects facts, including any custom
-facts found in Bolt's modulepath.
+Installs the puppet-agent package on targets if needed then collects facts,
```
cwcowellshah:
```suggestion
Installs the puppet-agent package on targets if needed, then collects facts,
```

pre-docs/plan_functions.md
```diff
@@ -59,14 +59,17 @@ add_to_group('foo,bar,baz', 'group1')
...
+including any custom facts found in Bolt's modulepath. The package is
+installed using either the configured plugin or the `task` plugin with the
+`puppet_agent::install` task
```
cwcowellshah:
```suggestion
`puppet_agent::install` task.
```

pre-docs/plan_functions.md
```diff
@@ -59,14 +59,17 @@ add_to_group('foo,bar,baz', 'group1')
...
 
 Agent detection will be skipped if the target includes the 'puppet-agent' feature, either as a
 property of its transport (PCP) or by explicitly setting it as a feature in Bolt's inventory.
 
-If no agent is detected on the target using the 'puppet_agent::version' task, it's installed
-using 'puppet_agent::install' and the puppet service is stopped/disabled using the 'service' task.
+If no agent is detected on the target using the 'puppet_agent::version' task,
```
cwcowellshah:
```suggestion
If Bolt does not detect an agent on the target using the 'puppet_agent::version' task,
```

pre-docs/plan_functions.md
```diff
@@ -59,14 +59,17 @@ add_to_group('foo,bar,baz', 'group1')
...
+it's installed using either the configured plugin or the
```
cwcowellshah:
```suggestion
it will install the agent using either the configured plugin or the
```

bolt-modules/boltlib/lib/puppet/functions/apply_prep.rb
```diff
@@ -61,21 +83,47 @@ def apply_prep(target_spec)
...
+              Concurrent::Future.execute(executor: pool) do
+                hash['hook_proc'].call
+              end
+            end
+
+            results = futures.zip(ok_hooks).map do |f, hash|
+              f.value || Bolt::Result.from_exception(hash['target'], f.reason)
+            end
+            set = Bolt::ResultSet.new(results + hook_errors)
+            raise Bolt::RunFailure.new(set.error_set, 'apply_prep') unless set.ok
```
adreyer: In the facts stuff we make the "action" in the run_failure `run_task`. I think this is better but we should put a TODO on line 141 so we dont' forget to standardize when we do the tasks work.

bolt-modules/boltlib/lib/puppet/functions/apply_prep.rb
```diff
@@ -24,14 +27,31 @@ def script_compiler
...
+
+  # rubocop:disable Naming/AccessorMethodName
+  def set_agent_feature(target)
+    inventory.set_feature(target, 'puppet-agent')
+  end
+  # rubocop:enable Naming/AccessorMethodName
+
+  def run_task(targets, tasksig, args = {})
+    errors = []
+    unless tasksig.runnable_with?(args || {}) { |msg| errors << msg }
```
adreyer: This is puppety and should probably happen outside of the concurrent section for clarity.

PR #1166, opened 2019-08-20
(GH-1060) Port over unused docker-api option tty, remove unused option wait

donoghuc:
When moving to the docker cli instead of the gem the tty option was not ported over, this commit adds the -tty flag when a the tty transport option is requested. The wait option was added as a workaround which was specific to the docker-api gem and was added after the original PR was raised, this commit removes it as it does not appear to be an issue with the CLI.

### DIFF
--- a/lib/bolt/transport/docker/connection.rb
+++ b/lib/bolt/transport/docker/connection.rb
@@ -51,6 +51,7 @@ module Bolt
           command_options = []
           # Need to be interactive if redirecting STDIN
           command_options << '--interactive' unless options[:stdin].nil?
+          command_options << '--tty' if options[:tty]
           command_options.concat(envs) unless envs.empty?
           command_options << container_id
           command_options.concat(command)
--- a/lib/bolt/transport/docker.rb
+++ b/lib/bolt/transport/docker.rb
@@ -51,10 +51,8 @@ module Bolt
       end
 
       def run_command(target, command, options = {})
-        if target.options['tty']
-          options[:Tty] = true
-        end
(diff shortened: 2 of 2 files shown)


donoghuc:
PR that introduced tty: https://github.com/puppetlabs/bolt/pull/962
Might not work from windows host, but I guess it never did?

lib/bolt/transport/docker/connection.rb
```diff
@@ -51,6 +51,7 @@ def execute(*command, options)
           command_options = []
           # Need to be interactive if redirecting STDIN
           command_options << '--interactive' unless options[:stdin].nil?
+          command_options << '--tty' if options[:tty]
```
Iristyle: @glennsarti you want to validate this on Windows?

IIRC this won't work

glennsarti: Nope.  `--tty` works fine, and for me is needed for cmd.exe

```
C:\Source> docker run --interactive ruby:2.5 bash

bash: line 1: $'\r': command not found

bash: line 2: $'\r': command not found

bash: line 3: $'\r': command not found
asdads
bash: line 4: $'asdads\r': command not found


C:\Source> docker run --interactive --tty ruby:2.5 bash
root@b528a0cd9c55:/# Hello
bash: Hello: command not found
root@b528a0cd9c55:/# ^C
root@b528a0cd9c55:/# exit
exit
C:\Source>
```

glennsarti: The `\r` is me pressing enter

PR #1170, opened 2019-08-22
Add developer update for 2019-08-22

nicklewis:


docs/_posts/2019-08-22-keep-on-pluggin.md
```diff
@@ -0,0 +1,17 @@
...
+
+We currently have a slew of large plugin-related features in flight, which means this week's release has just a couple of bug fixes and improvements.
+
+**Coming soon**
+
+We're reworking the [API for interacting with Target objects](https://github.com/puppetlabs/bolt/issues/1125) from within a plan to make it easier to dynamically create Targets and to change their config during a plan run.
+
+One of our most requested features, the ability to [control how `apply_prep` installs Puppet on targets](https://github.com/puppetlabs/bolt/issues/1123), is also in progress now.
+
+Also on the plugin front, we have both an [Azure inventory integration](https://github.com/puppetlabs/bolt/issues/1148) in the works as well as the ability to [add new plugins via Puppet modules](https://github.com/puppetlabs/bolt/issues/1133).
```
donoghuc:
```suggestion
On the plugin front, we have both an [Azure inventory integration](https://github.com/puppetlabs/bolt/issues/1148) in the works as well as the ability to [add new plugins via Puppet modules](https://github.com/puppetlabs/bolt/issues/1133).
```

donoghuc: Would suggest dropping the `also` either here or in the point above.

docs/_posts/2019-08-22-keep-on-pluggin.md
```diff
@@ -0,0 +1,17 @@
+---
+title: Keep on pluggin'
```
donoghuc: haha, nice

docs/_posts/2019-08-22-keep-on-pluggin.md
nicklewis: Removed the "also" above. I like it here because it points out that the `apply_prep` work is a plugin too (which helps support the title of the post!).

Issue #1175, opened 2019-08-27
Populate `$targets` plan parameter with target list


lucywyman:
## Describe the Solution You Would Like
We should populate a `$targets` plan parameter with the target list passed to bolt on the CLI, similar to how `$nodes` is autopopulated now. This should populate the `plan_arguments['targets']`value if `targets` is a plan parameter with `nodes` in `cli.rb`, just like how `$nodes` is populated (https://github.com/puppetlabs/bolt/blob/master/lib/bolt/cli.rb#L409-L417)

If both `$nodes` and `$targets` are plan parameters, populate neither and raise a warning. This should only populate if `$targets` has type `TargetSpec` or `Optional[TargetSpec]`.

This should include integration testing ([spec/integration/plan_spec.rb](spec/integration/plan_spec.rb)), updated documentation ([pre-docs/writing_plans.md#L414](pre-docs/writing_plans.md#L414)), and a CHANGELOG note

## Describe Alternatives You've Considered
Not doing this.

PR #1189, opened 2019-08-29
(GH-1125) Clear API for Target

### LINKED ISSUE #1125: Clear API for the `Target` Datatype
## Problem

The Target object in the plan language has a few well defined methods which are primarily around reading information. The current format of `Target.new` is not compatible with target objects defined in the inventory and there is no good way to change connection information for a target during a plan or create targets programmatically. 

## Impact

If you use the experimental v2 inventory file and use `Target.new` in your plans you may need to update them to support the new `Target` attributes.

### Changes to the `Target` object

New attributes for a target to support the use of `Target.new` from plans in a way that accepts the same data as a target declaration in the inventory file.

- `uri` Any data in the uri cannot be overridden by other config
- `name` If no name is specified the uri will be used as the name
- `config` May set the transport and/or config options for transports.
- `vars`
- `facts`
- `features`

In addition a new `display_name` method will be added to target. If a name was specified the name will be the display name. If the name is inferred from the uri then the display_name will exclude a password if specified in the URI.

The options attribute will no longer be available. 

### Examples

To create a new node with a name and host add it to a group for it's connection info

```
$target = Target.new(uri => 'my_node.example.com')
$target.add_to_group('my_config')
```

To create a node with a different name and uri. 
```
# Set the host with the URI
$target = Target.new(name => 'my_node', uri => '123.456.789.101')
# Or explicitly set the host on the transport.
$target = Target.new(name => 'my_node', config => { ssh => { host => '123.456.789.101' } } } )
```

To create new targets from the results of a target-lookups task

```
$target_data = run_task('module::my_lookup', 'localhost').first['targets']
$targets = $target_data.map |$data| { Target.new($data) }
```

### How data is tracked and merged

Conceptually there is a singleton target object for each unique target name. That object is stateful and tracks the connection information and other known data about the target. When a user accesses the data on a target object they will get the same result regardless of how they access the object `get_targets`, `Target.new()`, a `Target` parameter etc. 

#### Target and Group data are separate

In order to allow consistent data behavior target and group data will be treated separately in inventory version 2. Target level data will be computed once when the inventory is loaded and updated with the various setter functions on the target. IE `$target.add_facts`, `$target.set_var`, `target.set_config` etc. When a getter function is called ie `$target.vars` or connection information is accessed ie `run_task` the Target data for the target will be merged with the group data(where target level takes precedence).

#### Merge strategy

- `config` - deep merge of hashes. This allows `ssh` information to come from multiple locations
- `facts` - deep merge of hashes. This allows structured data to be set in mutiple locations and works with the `add_facts` function.
- `features` - All features are concatenated into a single feature set.
- `vars` - Shallow merge. The entire vars hash is merged but objects inside vars are not merged.

### Notes

- It is expected that users will not normally call `Target.new` with a target name that matches an existing target. Calling `Target.new` will overwrite any preexisting target level configuration but not change the groups the target is in.

### `get_target($TargetSpec $spec)`

This function resolves a `TargetSpec` similarly to `get_targets` but only returns a single `Target` object. If the targetspec resolves to more than one `Target` the function will error. 

To create a new target from a URI you can use `get_target` instead of `target.new`
```
get_target('ssh://me:secret@example.com')
```

To get a single object to operate on
```
$targ = get_target('my_target')
$targ.add_facts({<some facts>})
```

### `set_config(Target $target, Variant[String, Array] key_or_key_path, Any $value)`

Set a value at the target level of config.

To change the transport for a target
```
$targ.set_config('transport', 'pcp')
```

To set a password for a target
```
$targ.set_config(['ssh', 'password'], 'secret')
```

To override all of the ssh info for a target
```
$target.set_config('ssh', { user => 'me', password => 'secret' })
$target.add_to_group('no_hostkey')
```

order of groups should not matter so this is equivalent:
```
$target.add_to_group('no_hostkey')
$target.set_config('ssh', { user => 'me', password => 'secret' })
```

## Release Plan

In order to avoid breaking changes the new `Target` datatype should only be used if bolt is run with a `version: 2` inventory file.  The behavior of the ruby `Target` object is less important.

donoghuc:
With this commit the API for interacting with `Target` objects in the Plan language is updated to more closely match the way targets are defined in inventory V2. The previous API was focused more on reading data and setting a limited set of `Target` attributes.

This work enables new `Target` objects to be defined in a plan and for the configuration of `Targets` to easily be modified. The fundamental idea for this implementation is that a target is a single unique entity which tracks the connection information and other data associated with a `Target`.

 In order for `Target` data to predictably be resolved it will be kept separate from `Group` and `config` data. Data set directly on the `Target` will be given the highest priority. `Target` level data can be assigned in an inventory file, when using `Target.new` in a plan or with the various `set_*` plan functions. `Target` data is computed by merging `Target` data with `Group` and `config` data each time a target is modified (for example with a `set_*` method or when it is added to a group).

An important new plan function `set_config` has been added to enable setting a particular config value on a `Target`. Similarly a `get_target` method has been introduced to return a single `Target` to unambiguously set `Target` data. In addition to the updated interface this work introduces a pattern whereby targets created during a plan are automatically added to the inventory (in the `all` group). Similarly when a plan parameters has the type `TargetSpec` the parameter’s `Target` values will be added to inventory automatically if not already defined in an inventoryfile.

closes #1125

### DIFF
--- a/bolt-modules/boltlib/lib/puppet/datatypes/target.rb
+++ b/bolt-modules/boltlib/lib/puppet/datatypes/target.rb
@@ -1,22 +1,55 @@
 # frozen_string_literal: true
 
 Puppet::DataTypes.create_type('Target') do
-  interface <<-PUPPET
-    attributes => {
-      uri => String[1],
-      options => { type => Hash[String[1], Data], value => {} }
-    },
-    functions => {
-      name => Callable[[], String[1]],
-      host => Callable[[], Optional[String]],
-      password => Callable[[], Optional[String[1]]],
-      port => Callable[[], Optional[Integer]],
-      protocol => Callable[[], Optional[String[1]]],
-      user => Callable[[], Optional[String[1]]],
-    }
-  PUPPET
-
+  begin
+    inventory = Puppet.lookup(:bolt_inventory)
+    inventory_version = inventory.version
+    if inventory_version != 1
+      target_implementation_class = inventory.target_implementation_class
+    end
+  rescue Puppet::Context::UndefinedBindingError
+    inventory_version = 1
+  end
   load_file('bolt/target')
 
-  implementation_class Bolt::Target
+  if inventory_version == 1
+    interface <<-PUPPET
+      attributes => {
+        uri => String[1],
+        options => { type => Hash[String[1], Data], value => {} }
+      },
+      functions => {
+        name => Callable[[], String[1]],
+        host => Callable[[], Optional[String]],
+        password => Callable[[], Optional[String[1]]],
+        port => Callable[[], Optional[Integer]],
+        protocol => Callable[[], Optional[String[1]]],
+        user => Callable[[], Optional[String[1]]],
+      }
+    PUPPET
+    implementation_class Bolt::Target
+  else
+    interface <<-PUPPET
+      attributes => {
+        uri => { type => Optional[String[1]], kind => given_or_derived },
+        name => { type => Optional[String[1]] , kind => given_or_derived },
+        target_alias => { type => Optional[Variant[String[1], Array[String[1]]]], kind => given_or_derived },
+        config => { type => Optional[Hash[String[1], Data]], kind => given_or_derived },
+        vars => { type => Optional[Hash[String[1], Data]], kind => given_or_derived  },
+        facts => { type => Optional[Hash[String[1], Data]], kind => given_or_derived  },
+        features => { type => Optional[Array[String[1]]], kind => given_or_derived },
+        plugin_hooks => { type => Optional[Hash[String[1], Data]], kind => given_or_derived  }
+      },
+      functions => {
+        safe_name => Callable[[], String[1]],
+        host => Callable[[], Optional[String]],
+        password => Callable[[], Optional[String[1]]],
+        port => Callable[[], Optional[Integer]],
+        protocol => Callable[[], Optional[String[1]]],
+        user => Callable[[], Optional[String[1]]],
+      }
+    PUPPET
+
+    implementation_class target_implementation_class
+  end
 end
--- a/bolt-modules/boltlib/lib/puppet/functions/get_target.rb
+++ b/bolt-modules/boltlib/lib/puppet/functions/get_target.rb
@@ -1,0 +1,36 @@
+# frozen_string_literal: true
+
+require 'bolt/error'
+
+# Get a single target from inventory if it exists, otherwise create a new Target.
+#
+# **NOTE:** Calling `get_target` inside an `apply` block with a
+# version 2 inventory creates a new Target object.
+# `get_target('all')` returns an empty array.
+# **NOTE:** Only compatible with inventory v2
+Puppet::Functions.create_function(:get_target) do
+  # @param name A Target name.
+  # @return A single target, either new or from inventory.
+  # @example Create a new Target from a URI
+  #   get_target('winrm://host2:54321')
+  # @example Get an existing Target from inventory
+  #   get_target('existing-target')
+  dispatch :get_target do
+    param 'Boltlib::TargetSpec', :name
+    return_type 'Target'
+  end
+
+  def get_target(name)
+    inventory = Puppet.lookup(:bolt_inventory)
+    # Bolt executor not expected when invoked from apply block
+    executor = Puppet.lookup(:bolt_executor) { nil }
+    executor&.report_function_call(self.class.name)
+
+    unless inventory.version > 1
+      raise Puppet::ParseErrorWithIssue
+        .from_issue_and_stack(Bolt::PAL::Issues::UNSUPPORTED_INVENTORY_VERSION, action: 'get_target')
+    end
+
+    inventory.get_target(name)
+  end
+end
--- a/bolt-modules/boltlib/lib/puppet/functions/run_plan.rb
+++ b/bolt-modules/boltlib/lib/puppet/functions/run_plan.rb
@@ -83,6 +83,22 @@ Puppet::Functions.create_function(:run_plan, Puppet::Functions::InternalFunction
       executor.report_yaml_plan(closure.model.body)
     end
 
+    # If a TargetSpec parameter is passed, ensure it is in inventory
+    inventory = Puppet.lookup(:bolt_inventory)
+
+    if inventory.version > 1
+      param_types = closure.parameters.each_with_object({}) do |param, param_acc|
+        param_acc[param.name] = extract_parameter_types(param.type_expr).flatten
+      end
+      params.each do |param, value|
+        # Note the safe lookup operator is needed to handle case where a parameter is passed to a
+        # plan that the plan is not expecting
+        if param_types[param]&.include?('TargetSpec') || param_types[param]&.include?('Boltlib::TargetSpec')
+          inventory.get_targets(value)
+        end
+      end
+    end
+
     # wrap plan execution in logging messages
     executor.log_plan(plan_name) do
       result = nil
@@ -114,6 +130,32 @@ Puppet::Functions.create_function(:run_plan, Puppet::Functions::InternalFunction
       end
 
       result
+    end
+  end
+
+  # Recursively examine the type_expr to build a list of types
+  def extract_parameter_types(type_expr)
+    # No type
+    if type_expr.nil?
+      []
+    # Multiple types to extract (ex. Variant[TargetSpec, String])
+    elsif defined?(type_expr.keys)
+      type_expr.keys.flat_map { |param| extract_parameter_types(param) }
+    # Store cased value
+    elsif defined?(type_expr.cased_value)
+      [type_expr.cased_value]
+    # Type alias, able to resolve alias
+    elsif defined?(type_expr.resolved_type.name)
+      [type_expr.resolved_type.name]
+    # Nested type alias, recurse
+    elsif defined?(type_expr.type)
+      extract_parameter_types(type_expr.type)
+    # Array conatins alias types
+    elsif defined?(type_expr.types)
+      type_expr.types.flat_map { |param| extract_parameter_types(param) }
+    # Each element can be handled by a resolver above
+    elsif defined?(type_expr.element_type)
+      extract_parameter_types(type_expr.element_type)
     end
   end
 end
--- a/bolt-modules/boltlib/lib/puppet/functions/set_var.rb
+++ b/bolt-modules/boltlib/lib/puppet/functions/set_var.rb
@@ -9,13 +9,14 @@ Puppet::Functions.create_function(:set_var) do
   # @param target The Target object to set the variable for. See {get_targets}.
   # @param key The key for the variable.
   # @param value The value of the variable.
-  # @return [Undef]
+  # @return The target with the updated feature
   # @example Set a variable on a target
   #   $target.set_var('ephemeral', true)
   dispatch :set_var do
     param 'Target', :target
     param 'String', :key
     param 'Data', :value
+    return_type 'Target'
   end
 
   def set_var(target, key, value)
@@ -28,6 +29,9 @@ Puppet::Functions.create_function(:set_var) do
     executor = Puppet.lookup(:bolt_executor)
     executor.report_function_call(self.class.name)
 
-    inventory.set_var(target, key, value)
+    var_hash = { key => value }
+    inventory.set_var(target, var_hash)
+
+    target
   end
 end
--- a/bolt-modules/boltlib/spec/functions/add_facts_spec.rb
+++ b/bolt-modules/boltlib/spec/functions/add_facts_spec.rb
@@ -14,6 +14,7 @@ describe 'add_facts' do
   around(:each) do |example|
     Puppet[:tasks] = tasks_enabled
     Puppet.override(bolt_executor: executor, bolt_inventory: inventory) do
+      inventory.stubs(:version).returns(1)
       example.run
     end
   end
--- a/bolt-modules/boltlib/spec/functions/get_target_spec.rb
+++ b/bolt-modules/boltlib/spec/functions/get_target_spec.rb
@@ -1,0 +1,71 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+require 'bolt/executor'
+require 'bolt/target'
+
+describe 'get_target' do
+  include PuppetlabsSpec::Fixtures
+  let(:executor) { Bolt::Executor.new }
+  let(:inventory) { mock('inventory') }
+  let(:tasks_enabled) { true }
+
+  around(:each) do |example|
+    Puppet[:tasks] = tasks_enabled
+    Puppet.override(bolt_executor: executor, bolt_inventory: inventory) do
+      inventory.stubs(:version).returns(2)
+      inventory.stubs(:target_implementation_class).returns(Bolt::Target2)
+      example.run
+    end
+  end
+
+  context 'with inventory v1' do
+    it 'fails and reports that set_config is not available with inventory v1' do
+      inventory.stubs(:version).returns(1)
+      is_expected.to run
+        .with_params('foo').and_raise_error(/Plan language function 'get_target' cannot be used/)
+    end
+  end
+
+  context 'with inventory v2' do
+    let(:hostname) { 'foo.example.com ' }
+    let(:target) { Bolt::Target2.new(nil, hostname) }
+    let(:groupname) { 'all' }
+
+    it 'with given uri' do
+      inventory.expects(:get_target).with(hostname).returns(target)
+
+      is_expected.to run.with_params(hostname).and_return(target)
+    end
+
+    it 'with given Target' do
+      inventory.expects(:get_target).with(target).returns(target)
+
+      is_expected.to run.with_params(target).and_return(target)
+    end
+
+    it 'with given Target in array' do
+      inventory.expects(:get_target).with([target]).returns(target)
+
+      is_expected.to run.with_params([target]).and_return(target)
+    end
+
+    it 'errors when multiple targets returned' do
+      inventory.expects(:get_target).with(groupname).returns([target, target])
+
+      is_expected.to run.with_params(groupname)
+                        .and_raise_error(Puppet::Pops::Types::TypeAssertionError)
+    end
+
+    it 'errors on unknown types' do
+      is_expected.to run.with_params(mock('anything')).and_raise_error(ArgumentError)
+    end
+
+    it 'reports the call to analytics' do
+      inventory.expects(:get_target).with(hostname).returns(target)
+      executor.expects(:report_function_call).with('get_target')
+
+      is_expected.to run.with_params(hostname).and_return(target)
+    end
+  end
+end
--- a/bolt-modules/boltlib/spec/functions/set_config_spec.rb
+++ b/bolt-modules/boltlib/spec/functions/set_config_spec.rb
@@ -1,0 +1,64 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+require 'bolt/executor'
+require 'bolt/target'
+
+describe 'set_config' do
+  include PuppetlabsSpec::Fixtures
+  let(:executor) { Bolt::Executor.new }
+  let(:inventory) { mock('inventory') }
+  let(:target) { Bolt::Target2.new(nil, 'example') }
+  let(:tasks_enabled) { true }
+
+  around(:each) do |example|
+    Puppet[:tasks] = tasks_enabled
+    Puppet.override(bolt_executor: executor, bolt_inventory: inventory) do
+      inventory.stubs(:version).returns(2)
+      inventory.stubs(:target_implementation_class).returns(Bolt::Target2)
+      example.run
+    end
+  end
+
+  context 'with inventory v1' do
+    let(:target) { Bolt::Target.new('target1') }
+    it 'fails and reports that set_config is not available with inventoy v1' do
+      inventory.stubs(:version).returns(1)
+      is_expected.to run
+        .with_params(target, 'ssh', 'name' => 'foo')
+        .and_raise_error(/Plan language function 'set_config' cannot be used/)
+    end
+  end
+
+  context 'without tasks enabled' do
+    let(:tasks_enabled) { false }
+
+    it 'fails and reports that set_config is not available' do
+      is_expected.to run
+        .with_params(target, 'a', 'b').and_raise_error(/Plan language function 'set_config' cannot be used/)
+    end
+  end
+
+  it 'should set a config on a target' do
+    inventory.expects(:set_config).with(target, 'a', 'b').returns(target)
+    is_expected.to run.with_params(target, 'a', 'b').and_return(target)
+  end
+
+  it 'should sets nested config on a target' do
+    inventory.expects(:set_config).with(target, %w[a b], 'c').returns(target)
+    is_expected.to run.with_params(target, %w[a b], 'c').and_return(target)
+  end
+
+  it 'errors when passed invalid data types' do
+    is_expected.to run.with_params(target, 1, 'one')
+                      .and_raise_error(ArgumentError,
+                                       /'set_config' parameter 'key_or_key_path' expects a value of type String/)
+  end
+
+  it 'reports the call to analytics' do
+    executor.expects(:report_function_call).with('set_config')
+    inventory.expects(:set_config).with(target, 'a', 'b').returns(target)
+
+    is_expected.to run.with_params(target, 'a', 'b').and_return(target)
+  end
+end
--- a/lib/bolt/inventory/group2.rb
+++ b/lib/bolt/inventory/group2.rb
@@ -1,19 +1,20 @@
 # frozen_string_literal: true
 
 require 'bolt/inventory/group'
+require 'bolt/inventory/inventory2'
 
 module Bolt
   class Inventory
     class Group2
       attr_accessor :name, :targets, :aliases, :name_or_alias, :groups,
-                    :config, :rest, :facts, :vars, :features, :plugin_hooks
+                    :config, :facts, :vars, :features, :plugin_hooks
 
       # THESE are duplicates with the old groups for now.
       # Regex used to validate group names and target aliases.
       NAME_REGEX = /\A[a-z0-9_][a-z0-9_-]*\Z/.freeze
 
       DATA_KEYS = %w[name config facts vars features plugin_hooks].freeze
-      NODE_KEYS = DATA_KEYS + %w[alias uri]
+      TARGET_KEYS = DATA_KEYS + %w[alias uri]
       GROUP_KEYS = DATA_KEYS + %w[groups targets]
       CONFIG_KEYS = Bolt::TRANSPORTS.keys.map(&:to_s) + ['transport']
 
@@ -59,8 +60,10 @@ module Bolt
         groups = fetch_value(data, 'groups', Array)
 
         @targets = {}
+        # @target_objects = {}
         @aliases = {}
         @name_or_alias = []
+
         targets.each do |target|
           # If target is a string, it can refer to either a target name or
           # alias. Which can't be determined until all groups have been
@@ -152,6 +155,7 @@ module Bolt
         unless target.is_a?(Hash)
           raise ValidationError.new("Node entry must be a Hash, not #{target.class}", @name)
         end
+
         # This check prevents plugins from returning plugins
         raise ValidationError.new("Cannot set target with plugin", @name) if target.key?('_plugin')
         target.each do |k, v|
@@ -159,56 +163,57 @@ module Bolt
           validate_config_plugin(v, k, @name)
         end
 
-        target['name'] ||= target['uri']
+        t_name = target['name'] || target['uri']
 
-        if target['name'].nil? || target['name'].empty?
+        if t_name.nil? || t_name.empty?
           raise ValidationError.new("No name or uri for target: #{target}", @name)
         end
 
-        if @targets.include?(target['name'])
+        unless t_name.ascii_only?
+          raise ValidationError.new("Target name must be ASCII characters: #{target}", @name)
+        end
+
+        if @targets.include?(t_name)
           @logger.warn("Ignoring duplicate target in #{@name}: #{target}")
           return
         end
 
-        raise ValidationError.new("Node #{target} does not have a name", @name) unless target['name']
-        @targets[target['name']] = target
-
-        unless (unexpected_keys = target.keys - NODE_KEYS).empty?
-          msg = "Found unexpected key(s) #{unexpected_keys.join(', ')} in target #{target['name']}"
+        unless (unexpected_keys = target.keys - TARGET_KEYS).empty?
+          msg = "Found unexpected key(s) #{unexpected_keys.join(', ')} in target #{t_name}"
           @logger.warn(msg)
         end
 
         unless target['config'].nil? || target['config'].is_a?(Hash)
-          raise ValidationError.new("Invalid configuration for target: #{target['name']}", @name)
+          raise ValidationError.new("Invalid configuration for target: #{t_name}", @name)
         end
 
         config_keys = target['config']&.keys || []
         unless (unexpected_keys = config_keys - CONFIG_KEYS).empty?
-          msg = "Found unexpected key(s) #{unexpected_keys.join(', ')} in config for target #{target['name']}"
+          msg = "Found unexpected key(s) #{unexpected_keys.join(', ')} in config for target #{t_name}"
           @logger.warn(msg)
         end
 
         target['config'] = config_only_plugin(target['config'])
 
-        unless target.include?('alias')
-          return
-        end
-
-        aliases = target['alias']
-        aliases = [aliases] if aliases.is_a?(String)
-        unless aliases.is_a?(Array)
-          msg = "Alias entry on #{target['name']} must be a String or Array, not #{aliases.class}"
-          raise ValidationError.new(msg, @name)
-        end
-
-        aliases.each do |alia|
-          raise ValidationError.new("Invalid alias #{alia}", @name) unless alia =~ NAME_REGEX
-
-          if (found = @aliases[alia])
-            raise ValidationError.new(alias_conflict(alia, found, target['name']), @name)
+        if target.include?('alias')
+          aliases = target['alias']
+          aliases = [aliases] if aliases.is_a?(String)
+          unless aliases.is_a?(Array)
+            msg = "Alias entry on #{t_name} must be a String or Array, not #{aliases.class}"
+            raise ValidationError.new(msg, @name)
           end
-          @aliases[alia] = target['name']
+
+          aliases.each do |alia|
+            raise ValidationError.new("Invalid alias #{alia}", @name) unless alia =~ NAME_REGEX
+
+            if (found = @aliases[alia])
+              raise ValidationError.new(alias_conflict(alia, found, t_name), @name)
+            end
+            @aliases[alia] = t_name
+          end
         end
+
+        @targets[t_name] = target
       end
 
       def lookup_targets(lookup)
@@ -307,21 +312,22 @@ module Bolt
 
         # Collect target names and aliases into a list used to validate that subgroups don't conflict.
         # Used names validate that previously used group names don't conflict with new target names/aliases.
-        @targets.each_key do |n|
+        @targets.each do |t_name, t_data|
           # Require targets to be parseable as a Target.
           begin
-            Target.new(n)
+            # Catch malformed URI here
+            Bolt::Inventory::Inventory2.parse_uri(t_data['uri'])
           rescue Bolt::ParseError => e
             @logger.debug(e)
-            raise ValidationError.new("Invalid target name #{n}", @name)
+            raise ValidationError.new("Invalid target uri #{t_data['uri']}", @name)
           end
 
-          raise ValidationError.new(group_target_conflict(n), @name) if used_names.include?(n)
-          if aliased.include?(n)
-            raise ValidationError.new(alias_target_conflict(n), @name)
+          raise ValidationError.new(group_target_conflict(t_name), @name) if used_names.include?(t_name)
+          if aliased.include?(t_name)
+            raise ValidationError.new(alias_target_conflict(t_name), @name)
           end
 
-          target_names << n
+          target_names << t_name
         end
 
         @aliases.each do |n, target|
--- a/lib/bolt/inventory/inventory2.rb
+++ b/lib/bolt/inventory/inventory2.rb
@@ -5,6 +5,7 @@ require 'bolt/inventory/group2'
 module Bolt
   class Inventory
     class Inventory2
+      attr_reader :targets, :plugins, :config
       # This uses "targets" in the message instead of "nodes"
       class WildcardError < Bolt::Error
         def initialize(target)
@@ -12,11 +13,7 @@ module Bolt
         end
       end
 
-      attr_reader :plugins, :config
-
-      def initialize(data, config = nil, plugins: nil, target_vars: {},
-                     target_facts: {}, target_features: {},
-                     target_plugin_hooks: {})
+      def initialize(data, config = nil, plugins: nil)
         @logger = Logging.logger[self]
         # Config is saved to add config options to targets
         @config = config || Bolt::Config.default
@@ -24,10 +21,8 @@ module Bolt
         @groups = Group2.new(@data.merge('name' => 'all'), plugins)
         @plugins = plugins
         @group_lookup = {}
-        @target_vars = target_vars
-        @target_facts = target_facts
-        @target_features = target_features
-        @target_plugin_hooks = target_plugin_hooks
+        # The targets hash is the canonical source for all targets in inventory
+        @targets = {}
         @groups.resolve_aliases(@groups.target_aliases, @groups.target_names)
         collect_groups
       end
@@ -38,6 +33,10 @@ module Bolt
 
       def version
         2
+      end
+
+      def target_implementation_class
+        Bolt::Target2
       end
 
       def collect_groups
@@ -56,13 +55,15 @@ module Bolt
       alias node_names target_names
 
       def get_targets(targets)
-        targets = expand_targets(targets)
-        targets = if targets.is_a? Array
-                    targets.flatten.uniq(&:name)
-                  else
-                    [targets]
-                  end
-        targets.map { |t| update_target(t) }
+        flat_target_list(targets).map { |t| update_target(t) }
+      end
+
+      def get_target(target)
+        target_array = flat_target_list(target)
+        if target_array.count > 1
+          raise ValidationError.new("'#{target}' refers to #{target_array.count} targets", nil)
+        end
+        get_targets(target_array.first).first
       end
 
       def add_to_group(targets, desired_group)
@@ -76,41 +77,6 @@ module Bolt
         else
           raise ValidationError.new("Group #{desired_group} does not exist in inventory", nil)
         end
-      end
-
-      def set_var(target, key, value)
-        data = { key => value }
-        set_vars_from_hash(target.name, data)
-      end
-
-      def vars(target)
-        @target_vars[target.name] || {}
-      end
-
-      def add_facts(target, new_facts = {})
-        @logger.warn("No facts to add") if new_facts.empty?
-        set_facts(target.name, new_facts)
-      end
-
-      def facts(target)
-        @target_facts[target.name] || {}
-      end
-
-      def set_feature(target, feature, value = true)
-        @target_features[target.name] ||= Set.new
-        if value
-          @target_features[target.name] << feature
-        else
-          @target_features[target.name].delete(feature)
-        end
-      end
-
-      def features(target)
-        @target_features[target.name] || Set.new
-      end
-
-      def plugin_hooks(target)
-        @target_plugin_hooks[target.name] || {}
       end
 
       def data_hash
@@ -147,8 +113,13 @@ module Bolt
       private :config_plugin
 
       # Pass a target to get_targets for a public version of this
-      # Should this reconfigure configured targets?
       def update_target(target)
+        # Ensure all targets in inventory are included in the all group.
+        unless @groups.target_names.include?(target.name)
+          add_to_group([target], 'all')
+        end
+
+        # Get merged data between targets and groups
         data = @groups.data_for(target.name)
         data ||= {}
 
@@ -157,23 +128,24 @@ module Bolt
           data['config'] = {}
         end
 
-        data = Bolt::Inventory.localhost_defaults(data) if target.name == 'localhost'
-        # These should only get set from the inventory if they have not yet
-        # been instantiated
-        set_vars_from_hash(target.name, data['vars']) unless @target_vars[target.name]
-        set_facts(target.name, data['facts']) unless @target_facts[target.name]
-        data['features']&.each { |feature| set_feature(target, feature) } unless @target_features[target.name]
-        unless @target_plugin_hooks[target.name]
-          set_plugin_hooks(target.name, @config.plugin_hooks.merge(data['plugin_hooks'] || {}))
+        # Add defaults for special 'localhost' target (currently just config and features)
+        if target.name == 'localhost'
+          data = Bolt::Inventory.localhost_defaults(data)
         end
+
+        # Data from inventory
         data['config'] = config_plugin(data['config'])
+        # Data from set_config (make sure to resolve plugins)
+        resolved_target_config = config_plugin(@targets[target.name]['config'] || {})
+        data['config'] = Bolt::Util.deep_merge(data['config'], resolved_target_config)
 
         # Use Config object to ensure config section is treated consistently with config file
         conf = @config.deep_clone
         conf.update_from_inventory(data['config'])
         conf.validate
 
-        target.update_conf(conf.transport_conf)
+        # Recompute the target cached state with the merged data
+        update_target_state(target, conf.transport_conf, data)
 
         unless target.transport.nil? || Bolt::TRANSPORTS.include?(target.transport.to_sym)
           raise Bolt::UnknownTransportError.new(target.transport, target.uri)
@@ -182,6 +154,17 @@ module Bolt
         target
       end
       private :update_target
+
+      # This algorithm for getting a flat list of targets is used several times.
+      def flat_target_list(targets)
+        target_array = expand_targets(targets)
+        if target_array.is_a? Array
+          target_array.flatten.uniq(&:name)
+        else
+          [target_array]
+        end
+      end
+      private :flat_target_list
 
       # If target is a group name, expand it to the members of that group.
       # Else match against targets in inventory by name or alias.
@@ -209,8 +192,7 @@ module Bolt
       private :resolve_name
 
       def expand_targets(targets)
-        if targets.is_a? Bolt::Target
-          targets.inventory = self
+        if targets.is_a? Bolt::Target2
           targets
         elsif targets.is_a? Array
           targets.map { |tish| expand_targets(tish) }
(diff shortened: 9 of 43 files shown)


bolt-modules/boltlib/spec/functions/add_facts_spec.rb
```diff
@@ -14,6 +14,7 @@
   around(:each) do |example|
     Puppet[:tasks] = tasks_enabled
     Puppet.override(bolt_executor: executor, bolt_inventory: inventory) do
+      inventory.stubs(:version).returns(1)
```
donoghuc: For now most stdlib tests are restricted to V1 target/inventory. Shall I duplicate/add conditions for v2? or just replace all together?

donoghuc:
I put this up for early review because I think that all of the pieces for the ticket at least have a minimal implementation. I am currently working on converting some of the plans/inventoryfiles I have written to tests. As well as looking for places to add validation and optimizations. There are some questions commented in the code. I can go through  and translate those to PR comments.

donoghuc:
Can config plugins be used with `set_config`? 
```
$target.set_config('ssh', 'password' => { '_plugin' => 'prompt', 'message' => 'password plz'}
```

bolt-modules/boltlib/lib/puppet/functions/run_plan.rb
```diff
@@ -85,6 +85,29 @@ def run_plan(scope, plan_name, named_args = {})
...
+                                    type_acc << param_type.cased_value
+                                  end
+                                # Only a single key, may or may not have a type specified
+                                else
+                                  param.type_expr ? [param.type_expr.cased_value] : []
+                                end
+      end
+
+      params.each do |param, value|
+        if parameter_names_to_type[param].include?('TargetSpec')
```
adreyer: Does this handle aliases?

bolt-modules/boltlib/spec/functions/add_facts_spec.rb
adreyer: I wonder if we're better off testing these in `spec/pal`

lib/bolt/inventory/inventory2.rb
```diff
@@ -196,8 +196,7 @@ def resolve_name(target)
       private :resolve_name
 
       def expand_targets(targets)
-        if targets.is_a? Bolt::Target
-          targets.inventory = self
+        if targets.is_a? Bolt::Target2
```
adreyer: `kind_of(Bolt::Target)` is probably safer.

bolt-modules/boltlib/lib/puppet/functions/run_plan.rb
donoghuc: Ah, good question. When I was writing tests (pushed after your initial review) for this in `boltlib` I realized I needed `BoltLib::TargetSpec` so updated it to match `/TargetSpec/` and left a `CODEREVIEW` question. I think I for sure need look in to this more.

lib/bolt/inventory/inventory2.rb
donoghuc: I can dig in to the differences. Rubocop prefers `is_a?`
```
lib/bolt/inventory/inventory2.rb:197:20: C: Style/ClassCheck: Prefer Object#is_a? over Object#kind_of?.
        if targets.kind_of?(Bolt::Target2)
```

nicklewis: There is no difference.

lib/bolt/inventory/inventory2.rb
```diff
@@ -183,6 +159,17 @@ def update_target(target)
       end
       private :update_target
 
+      # This algorithm for getting a flat list of targets is used several times.
+      def flat_target_list(targets)
+        target_array = expand_targets(targets)
+        if target_array.is_a? Array
+          target_array.flatten.uniq(&:name)
```
lucywyman: Does `expand_targets` ever return an unflattened array?

lib/bolt/inventory/inventory2.rb
```diff
@@ -219,69 +205,51 @@ def expand_targets(targets)
...
-        end
-      end
-      private :set_vars_from_hash
-
-      def set_facts(tname, hash)
-        if hash
-          @target_facts[tname] ||= {}
-          @target_facts[tname] = Bolt::Util.deep_merge(@target_facts[tname], hash).freeze
-        end
+      # When a new target is added durring plan execution, ensure
```
lucywyman:
```suggestion
      # When a new target is added during plan execution, ensure
```

lib/bolt/inventory/inventory2.rb
```diff
@@ -219,69 +205,51 @@ def expand_targets(targets)
...
+      # It is added to the all group
```
lucywyman:
```suggestion
      # it is added to the all group
```

lib/bolt/inventory/inventory2.rb
```diff
@@ -56,13 +54,24 @@ def target_names
...
-        targets.map { |t| update_target(t) }
+        flat_target_list(targets).map { |t| update_target(t) }
+      end
+
+      def get_target(target)
+        # Used to track if new target was created
+        existing_targets = @targets.keys
+        target_array = flat_target_list(target)
+        if target_array.count > 1
+          matches = target_array.map(&:name)
```
lucywyman: `safe_name`?

lib/bolt/inventory/inventory2.rb
```diff
@@ -56,13 +54,24 @@ def target_names
...
+          raise ValidationError.new("'#{target}' refers to #{matches.count} targets", nil)
```
lucywyman: This seems like it can just be `target_array.count`, since `matches` content isn't used.

lib/bolt/inventory/inventory2.rb
```diff
@@ -56,13 +54,24 @@ def target_names
...
+        end
+        single_target = target_array.first
+        # add_to_group expect array
```
lucywyman: You could theoretically just use `target_array` instead then, since it is already known to be exactly `[single_target]`

lib/bolt/inventory/inventory2.rb
```diff
@@ -78,41 +87,6 @@ def add_to_group(targets, desired_group)
...
-        @target_features[target.name] || Set.new
-      end
-
-      def plugin_hooks(target)
-        @target_plugin_hooks[target.name] || {}
-      end
-
       def data_hash
         {
           data: {},
```
lucywyman: Are we leaving this be for now to be worked on later? Should it return the `@target` hash?

lib/bolt/inventory/inventory2.rb
```diff
@@ -157,23 +130,26 @@ def update_target(target)
...
-        # been instantiated
-        set_vars_from_hash(target.name, data['vars']) unless @target_vars[target.name]
-        set_facts(target.name, data['facts']) unless @target_facts[target.name]
-        data['features']&.each { |feature| set_feature(target, feature) } unless @target_features[target.name]
-        unless @target_plugin_hooks[target.name]
-          set_plugin_hooks(target.name, @config.plugin_hooks.merge(data['plugin_hooks'] || {}))
+        # Add defaults for special 'localhost' target (currently just config and features)
+        if target.name == 'localhost'
+          data = Bolt::Inventory.localhost_defaults(data)
+          data['features'].each do |feature|
```
lucywyman:
```suggestion
          data['features']&.each do |feature|
```

lucywyman: Actually I think the only time this is nil is if it's *explicitly* set to nil. In that case I think we can remove line 126 here, since it doesn't seem like the group data will ever be nil, since all nodes belong to at least the 'all' group.

lib/bolt/inventory/inventory2.rb
```diff
@@ -147,7 +121,6 @@ def config_plugin(data)
       private :config_plugin
 
       # Pass a target to get_targets for a public version of this
-      # Should this reconfigure configured targets?
       def update_target(target)
         data = @groups.data_for(target.name)
         data ||= {}
```
lucywyman: I'm not sure when this data would be nil, since targets should always at least be part of the 'all' group.

lib/bolt/inventory/inventory2.rb
lucywyman: I don't follow why this only happens for localhost?

lib/bolt/inventory/inventory2.rb
```diff
@@ -304,17 +272,141 @@ def inherit_data(track, name, data)
...
+      private :build_config_hash
+
+      def update_conf(target, conf)
+        @targets[target.name]['protocol'] = conf[:transport]
+        t_conf = conf[:transports][target.transport.to_sym] || {}
+        @targets[target.name]['user'] = t_conf['user']
+        @targets[target.name]['password'] = t_conf['password']
+        @targets[target.name]['port'] = t_conf['port']
+        @targets[target.name]['host'] = t_conf['host']
+        @targets[target.name]['options'] = @targets[target.name]['options'].merge(t_conf)
```
lucywyman: This seems like it should be `config` now? I'm not sure how `options` is different from `config` for `Bolt::Target2`. If they're different there's no harm leaving them, but it might be helpful to rename options

lib/bolt/inventory/inventory2.rb
```diff
@@ -304,17 +272,141 @@ def inherit_data(track, name, data)
...
+                              end
+        end
+
+        # Set all this for easy merge later
+        data['config'] ||= {}
+        data['vars'] ||= {}
+        data['facts'] ||= {}
+        data['features'] = data['features'] ? Set.new(data['features']) : Set.new
+        data['groups'] ||= []
+        data['options'] ||= {}
```
lucywyman: Do we still need this?

bolt-modules/boltlib/lib/puppet/datatypes/target.rb
```diff
@@ -1,22 +1,51 @@
...
+        host => Callable[[], Optional[String]],
+        password => Callable[[], Optional[String[1]]],
+        port => Callable[[], Optional[Integer]],
+        protocol => Callable[[], Optional[String[1]]],
+        user => Callable[[], Optional[String[1]]],
+      }
+    PUPPET
+    implementation_class Bolt::Target
+  else
+    # Is the `kind` appropriate for these?
```
lucywyman: Based on http://puppet-on-the-edge.blogspot.com/2018/11/ I think yes.

bolt-modules/boltlib/lib/puppet/functions/run_plan.rb
```diff
@@ -85,6 +85,31 @@ def run_plan(scope, plan_name, named_args = {})
       executor.report_yaml_plan(closure.model.body)
     end
 
+    # If a TargetSpec parameter is passed, ensure it is in inventory
+    inventory = Puppet.lookup(:bolt_inventory)
+    # Is it possible that inventory is not here?
```
lucywyman: Given we don't handle it being nil in any other function, I don't think so. Inventory is created when the CLI starts (unless it's just `--help`), and then loaded using `with_bolt_executor` in `Bolt::Pal`, which all plans use as part of executing the plan (via `in_plan_compiler`). I don't see any way this doesn't get set, at least currently.

lib/bolt/target.rb
```diff
@@ -1,8 +1,143 @@
...
+
+    def user
+      unencode(@inventory.targets[@name]['uri_obj']&.user) || @inventory.targets[@name]['user']
+    end
+
+    def password
+      unencode(@inventory.targets[@name]['uri_obj']&.password) || @inventory.targets[@name]['password']
+    end
+
+    def options
```
lucywyman: Should this have an alias to `.config`? It seems weird the config is the only object attribute that doesn't have a getter.

donoghuc: I think we can expose 'target_config` which would preserve the nested hash structure to allow something like:
```
Target.new('name' => 'copy', 'config' => get_target('blueprint').config)
```

lib/bolt/inventory/inventory2.rb
donoghuc: Basically config is boiled down to a flat options hash for easy use in transports. The outstanding question I have about options is that with the current implementation options are not pruned when a target's config is updated. So for example if you change protocol, any overlapping options are updated as expected, but ones that are not used by the updated transport persist. So for example if I were to update from ssh to winrm, the `disconnect-timeout` option would be persisted in options.

lib/bolt/inventory/inventory2.rb
donoghuc: Because `localhost` has some "magic" features set that are defined in `Bolt::Inventory.localhost_defaults`.

lib/bolt/inventory/inventory2.rb
donoghuc: This can be nil. 
As a trivial example, with an empty v2 inventory. Run `bolt command run whoami -t localhost`, the group `data_for` method will return nil.

lib/bolt/inventory/inventory2.rb
donoghuc: So this is used to reconstruct inventory to support target functions in apply blocks. We intentionally set this to empty data for now (the config stuff is required or get_targets will blow up). But basically the idea is that if you use get_targets in an apply block you wont get anything from inventory v2, but you can still get new v1 targets. I think we are going to drop support for get_targets etc in apply blocks in the future.

lib/bolt/inventory/inventory2.rb
donoghuc: I was thinking `single_target` was more readable than `target_array.first` everywhere.

lib/bolt/inventory/inventory2.rb
donoghuc: Ah, yeah, I originally was going to print out the names that matched. Removed.

bolt-modules/boltlib/lib/puppet/datatypes/target.rb
```diff
@@ -1,22 +1,51 @@
...
+        safe_name => Callable[[], String[1]],
+        host => Callable[[], Optional[String]],
+        password => Callable[[], Optional[String[1]]],
+        port => Callable[[], Optional[Integer]],
+        protocol => Callable[[], Optional[String[1]]],
+        user => Callable[[], Optional[String[1]]],
+      }
+    PUPPET
+
+    implementation_class Bolt::Target2
```
branan: In PE we need to be able to control Target::new() to disallow setting connection information, since that is expected to come from the Console inventory.

If this was `implementation_class inventory.target_class()`, we could override the Target class to our subclass without the need to modify this file further.

branan: /cc @McdonaldSeanp @adreyer

bolt-modules/boltlib/lib/puppet/functions/get_target.rb
```diff
@@ -0,0 +1,36 @@
...
+    # Bolt executor not expected when invoked from apply block
+    executor = Puppet.lookup(:bolt_executor) { nil }
+    executor&.report_function_call(self.class.name)
+
+    unless inventory.version > 1
+      raise Puppet::ParseErrorWithIssue
+        .from_issue_and_stack(Bolt::PAL::Issues::UNSUPPORTED_INVENTORY_VERSION, action: 'get_target')
+    end
+
+    inventory.get_target(name)
```
lucywyman: Is this not just a special case of `get_targets`? Should we just call that and get the first element?

donoghuc: `get_targets` wont add a target to inventory. The desire for `get_target` called in a plan with a target name that does not exist in inventory to be part of the `all` group makes this a special case. 
Also unrelated, i noticed a pretty bad bug in this method. I invoke `existing_targets = @targets.key` before resolving the targets with `target_array = flat_target_list(target)` in the next line which can leave `@targets` un-populated with targets from inventory... Will push an update soon.

bolt-modules/boltlib/lib/puppet/functions/run_plan.rb
```diff
@@ -85,6 +85,31 @@ def run_plan(scope, plan_name, named_args = {})
...
+                                # Only a single key, may or may not have a type specified
+                                else
+                                  param.type_expr ? [param.type_expr.cased_value] : []
+                                end
+      end
+
+      params.each do |param, value|
+        # CODEREVIEW: is it safe to match on with i =~ /TargetSpec/ it handles the BoltLib::TargetSpec
+        # but would I guess could match on any type werhe cased_value matches TargetSpec
+        if parameter_names_to_type[param]&.any? { |i| i =~ /TargetSpec/ }
```
lucywyman: I can't think of a reasonable type that would falsely match this (hopefully no one is doing `Variant[TargetSpec, Integer]`). I'm inclined to wait until it's a problem. With that said it might be safer to just match `TargetSpec` and `Optional[TargetSpec]`, since I expect that will be 99% of cases.

donoghuc: I added some ugly logic to resolve type aliases. I think that I can just check for two conditions, `Boltlib::TargetSpec` and `TargetSpec`.

bolt-modules/boltlib/lib/puppet/datatypes/target.rb
donoghuc: Updated.

donoghuc:
@lucywyman , I refactored this based on a discussion with @adreyer to *only* set target attributes either when loading inventoryfile (from target level), when calling Target.new, or the set_* functions when running a plan. I still need to clean up some of the tests but I wanted to get this up asap so you can see the updated approach.

lib/bolt/target.rb
donoghuc: Exposed config

bolt-modules/boltlib/lib/puppet/functions/get_target.rb
donoghuc: Actually, I think I can use `get_targets` here in addition to the new `ensure_in_all_group` method which was implemented after this. Thanks.

bolt-modules/boltlib/lib/puppet/functions/set_var.rb
```diff
@@ -28,6 +28,12 @@ def set_var(target, key, value)
     executor = Puppet.lookup(:bolt_executor)
     executor.report_function_call(self.class.name)
 
-    inventory.set_var(target, key, value)
+    # Inventory v2 expects a hash, should we just update v1 as well?
```
lucywyman: I don't see a reason not to :woman_shrugging:

lib/bolt/inventory/group2.rb
```diff
@@ -307,21 +313,22 @@ def validate(used_names = Set.new, target_names = Set.new, aliased = {}, depth =
 
         # Collect target names and aliases into a list used to validate that subgroups don't conflict.
         # Used names validate that previously used group names don't conflict with new target names/aliases.
-        @targets.each_key do |n|
+        @targets.each do |t_name, t_data|
           # Require targets to be parseable as a Target.
           begin
-            Target.new(n)
+            # For now just validate that uri can parse, other validation?
```
lucywyman: If there is it seems like it should go in `parse_uri`, but I think that validation is ok

lib/bolt/inventory/group2.rb
```diff
@@ -1,19 +1,20 @@
...
-                    :config, :rest, :facts, :vars, :features, :plugin_hooks
+                    :config, :facts, :vars, :features, :plugin_hooks
 
       # THESE are duplicates with the old groups for now.
       # Regex used to validate group names and target aliases.
       NAME_REGEX = /\A[a-z0-9_][a-z0-9_-]*\Z/.freeze
 
       DATA_KEYS = %w[name config facts vars features plugin_hooks].freeze
-      NODE_KEYS = DATA_KEYS + %w[alias uri]
+      TARGET_KEYS = DATA_KEYS + %w[alias uri safe_name]
```
adreyer: I don't think we want people to set this. The only reason it exists is when the URI isn't safe right and there is no explicit name?

donoghuc: You are right. I was using this because we assign a name from URI when adding targets to groups (which was complicating my algorithm for when a safe name should not be the name), I figured if a user added `safe_name` in inventory it would just silently overwritten.  I see how that would be confusing to to the user. I will figure out a different way.

bolt-modules/boltlib/lib/puppet/functions/set_var.rb
donoghuc: updated

bolt-modules/boltlib/lib/puppet/functions/run_plan.rb
donoghuc: Updated to handle type aliases

lib/bolt/inventory/inventory2.rb
```diff
@@ -219,102 +197,202 @@ def expand_targets(targets)
...
+        target = Target2.new(nil, data['name'])
+        target.inventory = self
+        data['self'] = target
+        @targets[data['name']] = data
+        target
       end
-      private :set_facts
+      private :create_target
+
+      def create_target_from_plan(data)
```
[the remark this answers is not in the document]
donoghuc: Yep, this is a known issue. The safe_name method will make this possible to fix. I spoke with Alex and he agrees that updating that is outside the scope of this ticket. (Having the empty host is not newly introduced with this refactor).

lib/bolt/target.rb
```diff
@@ -1,8 +1,147 @@
...
+    end
+
+    # Not married to this method, I was just exceeding 120 char line length
+    def unencode(component)
+      Addressable::URI.unencode_component(component)
+    end
+    private :unencode
+  end
+
+  # TODO: Move this? Remove pulling in addressable/uri, etc
```
lucywyman: I think it's fine to leave it here?

lib/bolt/inventory/inventory2.rb
```diff
@@ -219,102 +197,202 @@ def expand_targets(targets)
...
+        end
+        # Recurse on children Groups if not desired_group
+        current_group.groups.each do |child_group|
+          add_target(child_group, target, desired_group)
         end
       end
-      private :set_vars_from_hash
+      private :add_target
+
+      # CODEREVEW: For now these methods live in inventory, should it be a subclass or some other module?
```
lucywyman: Since these are both reading from and writing to state on the `Inventory` class it seems like managing another class or module would only complicate things (either having to pass data back and forth, or making `@groups` readable and `@targets` writable, plus manage the new class' lifecycle). It makes sense (to me at least?) for these methods to be operating on the Inventory.

lib/bolt/inventory/inventory2.rb
```diff
@@ -219,102 +197,202 @@ def expand_targets(targets)
...
+      private :create_target
+
+      def create_target_from_plan(data)
+        t_name = data['name'] || data['uri']
 
-      def set_plugin_hooks(tname, hash)
-        if hash
-          @target_plugin_hooks[tname] ||= {}
-          @target_plugin_hooks[tname].merge!(hash)
+        # If target allready exists, delete old and replace with new
```
lucywyman:
```suggestion
        # If target already exists, delete old and replace with new
```

lib/bolt/inventory/inventory2.rb
```diff
@@ -219,102 +197,202 @@ def expand_targets(targets)
...
+        # Otherwise add to new to all group
```
lucywyman:
```suggestion
        # Otherwise add target to all group
```

lib/bolt/inventory/inventory2.rb
```diff
@@ -219,102 +197,202 @@ def expand_targets(targets)
...
-          end
-          hook_data = @config.plugin_hooks.merge(data[:plugin_hooks])
-          hash = if @target_plugin_hooks[t_name]
-                   hook_data.merge(@target_plugin_hooks[t_name])
-                 else
-                   hook_data
-                 end
-          set_plugin_hooks(t_name, hash)
-          return true
+      # CODEREVIEW: This is helpful in group2 for validation. Where is a good place for it?
```
lucywyman: Honestly `Bolt::Util` seems like a good place for it? It's a helper method, and since it's not really part of the groups or inventory's functionality I think util makes the most sense.

pre-docs/inventory_version2.md
```diff
@@ -76,6 +76,93 @@ plan setup_lb (
 }
 ```
 
+## Creating targets in plans
+
+When using inventory v2 a new and improved API for interacting with `Target`s in inventory is available. Two new plan functions have been added: `get_target` which allows retrieving a single `Target` from inventory and `set_config` which allows setting config on a specific `Target`. The updated API also provides a way to instantiate new `Target`s with data that more closely resembles how targets are declared in an inventory file.
```
lucywyman: I'm not sure if Chris will want this to change, but I like that `Target` is caps and preformatted, since it indicates that it's an object.

pre-docs/inventory_version2.md
```diff
@@ -76,6 +76,93 @@ plan setup_lb (
...
+
+For example, consider the following `new_targets` plan:
+```
+plan new_targets(){
+  $new_target = get_target('ssh://user:secret@1.2.3.4:2222')
+  $new_target.set_config(['ssh', 'host-key-check'], false)
+}
+```
+In the `new_targets` plan the `get_target` function returns a `Target` identified with the name `ssh://user:secret@1.2.3.4:2222`. If a `Target` with that name does not exist in inventory a new `Target` is instantiated with the `uri` and `name` attributes set to `ssh://user:secret@1.2.3.4:2222` and is added to the `all` group in inventory (where it inherits and configuration for the `all` group. If the `Target` with that name does exist, then it is simply returned. The `set_config` method is used to set a transport specific setting specified by the array of keys to that setting that matches the keys in the structured hash found in an inventoryfile under the `config` key. This illustrates how a new `Target` can be created from a URI and configuration options that are not able to be set in URI parts can be modified.
```
```suggestion
In the `new_targets` plan the `get_target` function returns a `Target` identified with the name `ssh://user:secret@1.2.3.4:2222`. If a `Target` with that name does not exist in inventory a new `Target` is instantiated with the `uri` and `name` attributes set to `ssh://user:secret@1.2.3.4:2222` and is added to the `all` group in inventory (where it inherits and configuration for the `all` group). If the `Target` with that name does exist, then it is simply returned. The `set_config` method is used to set a transport specific setting specified by the array of keys to that setting that matches the keys in the structured hash found in an inventoryfile under the `config` key. This illustrates how a new `Target` can be created from a URI and configuration options that are not able to be set in URI parts can be modified. 
```

pre-docs/inventory_version2.md
```diff
@@ -76,6 +76,93 @@ plan setup_lb (
...
+plan new_target_alternate(){
+  $config = { 'transport' => 'ssh',
+              'ssh' => {
+                'user' => 'user',
+                'password' => 'secret',
+                'host' => '1.2.3.4'
+                'port' => 2222,
+                'host-key-check' => false
+                }}
+  $new_target = Target.new('name' => 'new_target', 'config' => $config)
```
lucywyman: 2 things: 1. I'd add a line explaining what this plan does (just like "This plan creates a new target using `Target.new` and passing in a config hash") 2. I'd add another example that also passes in features or vars or facts

pre-docs/inventory_version2.md
```diff
@@ -76,6 +76,93 @@ plan setup_lb (
...
+}
+```
+In the `new_target_alternate` plan a new `Target` is created from a hash and added to the `all` group in inventory. If a `Target` with name `new_target` had already existed in inventory, that `Target` would be destroyed and the new `Target` would take its place.
```
lucywyman: I'd make this a note so it's a bit more prominent

pre-docs/inventory_version2.md
```diff
@@ -76,6 +76,93 @@ plan setup_lb (
...
+```
+The `auto_add` plan returns all of the targets in the `all` group, if the value of `$nodes` resolves to a `String` that does not match a `Target` name, a group name, a `Target` alias or a target regex, a new `Target` is created and added to the `all` group. 
+
+## `Target` reference
+
+A target object can be instantiated with `Target.new` from a plan with either a `String` representing the `Target` `name` and `uri` or a hash with the following structure:
+
+- `uri`: `String`, Target URI (will be used as the `Target` name if a name is not specified)
+- `name`: `String`, The name of the target
+- `target_alias`, `Variant[String, Array[String]]`, The alias to refer to a target by
```
```suggestion
- `target_alias`:  `Variant[String, Array[String]]`, The alias to refer to a target by
```

pre-docs/inventory_version2.md
```diff
@@ -76,6 +76,93 @@ plan setup_lb (
...
+
+For example:
+```
+plan auto_add(TargetSpec $nodes) {
+  return get_targets('all')
+}
+```
+The `auto_add` plan returns all of the targets in the `all` group, if the value of `$nodes` resolves to a `String` that does not match a `Target` name, a group name, a `Target` alias or a target regex, a new `Target` is created and added to the `all` group. 
+
+## `Target` reference
```
lucywyman: In a better docs world I'd say this might deserve it's own page, and get linked from the inventory v2?

bolt-modules/boltlib/spec/functions/get_target_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+    let(:target) { Bolt::Target2.new(nil, hostname) }
+    let(:groupname) { 'all' }
+
+    it 'with given uri' do
+      inventory.expects(:get_target).with(hostname).returns(target)
+
+      is_expected.to run.with_params(hostname).and_return(target)
+    end
+
+    it 'with given Target' do
```
lucywyman: May also be worth adding a test with an array

bolt-modules/boltlib/spec/functions/set_config_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+    let(:tasks_enabled) { false }
+
+    it 'fails and reports that set_var is not available' do
+      is_expected.to run
+        .with_params(target, 'a', 'b').and_raise_error(/Plan language function 'set_config' cannot be used/)
+    end
+  end
+
+  it 'should set a config on a target' do
+    inventory.expects(:set_config).with(target, 'a', 'b').returns(target)
```
lucywyman: This seems like it should also verify that nested hashes are set, and that the alternate syntax works (so `(target, ['a', 'b'], 'secret'`)

bolt-modules/boltlib/spec/functions/set_config_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+  end
+
+  it 'reports the call to analytics' do
+    executor.expects(:report_function_call).with('set_config')
+    inventory.expects(:set_config).with(target, 'a', 'b').returns(target)
+
+    is_expected.to run.with_params(target, 'a', 'b').and_return(target)
+  end
+
+  context 'without tasks enabled' do
```
lucywyman: I think you already test this above?

lib/bolt/inventory/group2.rb
```diff
@@ -190,25 +200,28 @@ def add_target(target)
...
-            raise ValidationError.new(alias_conflict(alia, found, target['name']), @name)
+            if (found = @aliases[alia])
+              raise ValidationError.new(alias_conflict(alia, found, target['name']), @name)
+            end
+            @aliases[alia] = target['name']
           end
-          @aliases[alia] = target['name']
         end
+
+        raise ValidationError.new("Node #{target} does not have a name", @name) unless target['name']
```
lucywyman: This seems like it will already be raised on line 173?  Are `if target['name'].nil?` and `unless target['name']` different?

lib/bolt/inventory/group2.rb
```diff
@@ -190,25 +200,28 @@ def add_target(target)
...
+        t_name = target['name']
```
lucywyman: I think this line can just be inline on 224:
```
@targets[target['name']] = target
```

lib/bolt/inventory/inventory2.rb
```diff
@@ -157,23 +125,23 @@ def update_target(target)
...
+          data = Bolt::Inventory.localhost_defaults(data)
         end
+
+        # Data from inventory
         data['config'] = config_plugin(data['config'])
+        # Data from set_config (make sure to resolve plugins)
+        resolved_target_config = config_plugin(@targets[target.name]['config'] || {})
+        data['config'] = Bolt::Util.deep_merge(data['config'], resolved_target_config)
 
-        # Use Config object to ensure config section is treated consistently with config file
```
lucywyman: Is there a reason for removing this?

lib/bolt/inventory/inventory2.rb
```diff
@@ -219,102 +197,202 @@ def expand_targets(targets)
...
-          hook_data = @config.plugin_hooks.merge(data[:plugin_hooks])
-          hash = if @target_plugin_hooks[t_name]
-                   hook_data.merge(@target_plugin_hooks[t_name])
-                 else
-                   hook_data
-                 end
-          set_plugin_hooks(t_name, hash)
-          return true
+      # CODEREVIEW: This is helpful in group2 for validation. Where is a good place for it?
+      # Also is loading the addressable code here OK?
```
lucywyman: I think so - it's ~2MB (which I think is big but not *that* big???), given we pulled it in in init for Target1 there at least won't be a performance hit.

lib/bolt/inventory/inventory2.rb
```diff
@@ -219,102 +197,202 @@ def expand_targets(targets)
...
+        @targets[target.name]['vars'] = @targets[target.name]['vars'].merge(var_hash)
+        update_target(target)
+      end
+
+      def vars(target)
+        @targets[target.name]['cached_state']['vars'] || {}
+      end
+
+      def add_facts(target, new_facts = {})
+        # CODEREVIEW: Do we want to warn for set methods?
```
lucywyman: I don't think there's a need to. The scenario where someone unintentionally has an empty hash where they intended to have data seem unlikely, and the puppet function doesn't allow this to be nil.

lib/bolt/inventory/inventory2.rb
```diff
@@ -219,102 +197,202 @@ def expand_targets(targets)
...
-        # If there is no name then this target was only referred to as a string.
-        uri = data['uri']
-        uri ||= target_name unless data['name']
+      def target_config(target)
+        @targets[target.name]['cached_state']['config'] || {}
+      end
 
-        Target.new(uri, name_opt)
+      def build_config_hash(key_or_key_path, value)
+        # https://stackoverflow.com/questions/5095077/ruby-convert-array-to-nested-hash
```
lucywyman: :smile:

spec/bolt/inventory/inventory2_spec.rb
```diff
@@ -387,14 +391,15 @@ def get_target(inventory, name, alia = nil)
 
         it 'should return config for the target' do
           target = get_target(inventory, 'target3')
+          # Should name continue to be duplicated in options? Seems uneccesary
```
lucywyman: It seems like the purpose of options is to capture connection configuration for the target, which I don't think includes `name`. So I think we can leave it out.

spec/bolt/inventory/inventory2_spec.rb
```diff
@@ -682,6 +687,329 @@ def common_data(transport)
...
+          names = ['pcp://a', 'winrm://b', 'c']
+          expect {
+            inventory.get_target(names)
+          }.to raise_error(Bolt::Inventory::ValidationError, "'#{names}' refers to 3 targets")
+        end
+
+        it 'should fail for unknown protocols' do
+          expect {
+            inventory.get_target('z://foo')
+          }.to raise_error(Bolt::UnknownTransportError, %r{Unknown transport z found for z://foo})
```
lucywyman: Should this also verify an error is raised with an invalid URI?

spec/bolt/inventory/inventory2_spec.rb
```diff
@@ -682,6 +687,329 @@ def common_data(transport)
...
+        end
+      end
+
+      context 'non-empty inventory' do
+        let(:inventory) {
```
lucywyman:
```suggestion
        let(:inventory) { Bolt::Inventory::Inventory2.new(data) }
```
And remove the 2 lines below

spec/bolt/inventory/inventory2_spec.rb
```diff
@@ -682,6 +687,329 @@ def common_data(transport)
         end
       end
     end
+
+    describe 'get_target' do
```
lucywyman: You could have
```
let(:inventory) { Bolt::Inventory::Inventory2.new(data) }
```
under the `describe`, then just set data where it's special (including `{}` for the empty inventory).

spec/bolt/inventory/inventory2_spec.rb
```diff
@@ -682,6 +687,329 @@ def common_data(transport)
...
+      end
+
+      context 'with data in the group' do
+        let(:inventory) { Bolt::Inventory::Inventory2.new(data) }
+
+        it 'should use value from lowest target definition' do
+          expect(inventory.get_target('target4').user).to eq('me')
+        end
+
+        it 'should use values from the lowest group' do
```
lucywyman: I think this is technically leftmost?

spec/integration/inventory2_spec.rb
```diff
@@ -190,6 +190,30 @@ def fact_plan(name = 'facts_test')
     end
   end
 
+  context 'when creating targets from a plan with empty inventory', ssh: true do
```
lucywyman: This seems like it could also test `set_config`. I think it'd also be useful to add a test that creating a node and adding it to a group picks up all the right data

spec/bolt/inventory/inventory2_spec.rb
```diff
@@ -682,6 +687,329 @@ def common_data(transport)
...
+        let(:inventory) { Bolt::Inventory::Inventory2.new(data) }
+
+        it 'should return group config for string targets' do
+          target = inventory.get_target('target1')
+          expect(target.options).to include('host-key-check' => false)
+          expect(target.user).to eq('you')
+          expect(target.config).to eq("ssh" => { "user" => "you", "host-key-check" => false })
+        end
+
+        it 'should return group config for array targets' do
```
lucywyman:
```suggestion
        it 'should return group config for hash targets' do
```

spec/bolt/inventory/inventory2_spec.rb
```diff
@@ -682,6 +687,329 @@ def common_data(transport)
...
+        }
+        let(:conf) { Bolt::Config.default }
+        let(:inventory) { Bolt::Inventory::Inventory2.new(data, conf) }
+
+        it 'should not modify existing config' do
+          inventory.get_target('ssh://target')
+          expect(conf.transport).to eq('ssh')
+          expect(conf.transports[:ssh]['load-config']).to be true
+          expect(conf.transports[:winrm]['ssl']).to be true
+          expect(conf.transports[:winrm]['ssl-verify']).to be true
```
lucywyman: One check for winrm seems like enough

spec/bolt/inventory/inventory2_spec.rb
```diff
@@ -743,6 +1071,123 @@ def common_data(transport)
...
+                       "file-protocol" => "winrm",
+                       "connect-timeout" => 10 }
+        expect(target.options).to include(winrm_conf)
+        expected_config['transport'] = 'winrm'
+        expected_config['winrm'] = { 'winrm' => { 'password' => 'winrmpass' } }
+      end
+    end
+  end
+
+  describe 'set_config' do
```
lucywyman: I don't follow why there's 2 `describe 'set_config'` here? I think you just need one

lucywyman:
You might've already caught this but it also needs a CHANGELOG update

bolt-modules/boltlib/lib/puppet/functions/run_plan.rb
```diff
@@ -85,6 +85,22 @@ def run_plan(scope, plan_name, named_args = {})
       executor.report_yaml_plan(closure.model.body)
     end
 
+    # If a TargetSpec parameter is passed, ensure it is in inventory
+    inventory = Puppet.lookup(:bolt_inventory)
+
+    if inventory.version > 1
```
lucywyman: Can we pull this algorithm into it's own function and use it in `run_task` as well, for #1175

bolt-modules/boltlib/lib/puppet/datatypes/target.rb
```diff
@@ -1,22 +1,55 @@
...
+        user => Callable[[], Optional[String[1]]],
+      }
+    PUPPET
+    implementation_class Bolt::Target
+  else
+    interface <<-PUPPET
+      attributes => {
+        uri => { type => Optional[String[1]], kind => given_or_derived },
+        name => { type => Optional[String[1]] , kind => given_or_derived },
+        target_alias => { type => Optional[Variant[String[1], Array[String[1]]]], kind => given_or_derived },
```
lucywyman: I can't find / don't remember the discussion where we decided to call this `target_alias`? I assumed it was `alias`, and the error message is readable but not pretty:
```
lucy ~/githubs/bolt (GH-1125) $ be bolt plan run test --params @data.json
Starting: plan test
Finished: plan test in 0.01 sec
{
  "kind": "bolt/pal-error",
  "msg": "'new_Target' expects one of:\n  (Struct[{'uri' => Optional[String[1]], 'name' => Optional[String[1]], 'target_alias' => Optional[Variant[String[1], Array[String[1]]]], 'config' => Optional[Hash[String[1], Data]], 'vars' => Optional[Hash[String[1], Data]], 'facts' => Optional[Hash[String[1], Data]], 'features' => Optional[Array[String[1]]], 'plugin_hooks' => Optional[Hash[String[1], Data]]}] hash)\n    rejected: parameter 'hash' unrecognized key 'alias'\n  (String[1] uri?, String[1] name?, Variant[String[1], Array[String[1]]] target_alias?, Hash[String[1], Data] config?, Hash[String[1], Data] facts?, Hash[String[1], Data] vars?, Array[String[1]] features?, Hash[String[1], Data] plugin_hooks?)\n    rejected: parameter 'uri' expects a value of type Undef or String, got Struct",
  "details": {
  }
}
```

Is there a reason not to just make this `alias`?

donoghuc: alias is a reserved word in ruby. I dont think there is much we can do about the error message.

lib/bolt/inventory/inventory2.rb
donoghuc: I dont think it belongs in util. There is some specific logic to ensure that the `Uri` object that is returned is compatible with how we use it in inventory (for example, how we append `//` conditionally and forbid empty strings). Maybe it should go in `Bolt::Inventory` becuase both `Inventory2 and `Group2` are derived from that?

lucywyman: Inventory2 and Group2 are just nested classes not child classes, so they wouldn't inherit the method. If we don't want to use it elsewhere it makes the most sense in Inventory2 - it isn't used in Inventory, and conceptually fits with being an inventory2 helper method.

lib/bolt/inventory/group2.rb
donoghuc: yeah, that seems safe to remove.

lib/bolt/inventory/group2.rb
donoghuc: This might make more sense with the following two lines
```
        t_name = target['name']
        target.delete('name') if name_from_uri
        @targets[t_name] = target
```
if we set the name from the URI (in line 166) which will be the target name and is needed to keep track of targets in current group. When we later call `create_target` we need to set a safe_name. If we already have artificially assigned the name as the target URI, the algorithm in `creat_target` that decides when we need a safe name will be fooled. So the idea is to save the name, delete `target['name']` from the data (because we artificially populated it) and then assign to the group's `@targets` variable.

lib/bolt/inventory/inventory2.rb
donoghuc: Nope, throughout development on this and the refactors i've done I had a ton of comment notes to myself. I think I deleted this one thinking it was one of mine. Adding it back.

spec/bolt/inventory/inventory2_spec.rb
donoghuc: Oops. this comment is from an earlier iteration. Name is not stored in options. My bad. deleted.

spec/bolt/inventory/inventory2_spec.rb
donoghuc: We have a test in group2 to test that invalid URI's are forbidden. Were you suggesting a new test here with an invalid uri? (`z://foo`) is valid.

spec/bolt/inventory/inventory2_spec.rb
donoghuc: This was copy pasted from line 299. Updated in both places.

donoghuc:
How should we handle target features? Internally we store features in a set datastructure and rely on some set functionality. With this PR the Target2 class has a `features` method that returns a set which does not work well for plans. Should we reserve that *only* for use in plans (and have it return an array) and use a separate method internally? Something like `feature_set`?

spec/bolt/inventory/inventory2_spec.rb
donoghuc: This is replicated from above. Seems harmless to explicitly ensure things are not overridden.

spec/bolt/inventory/inventory2_spec.rb
donoghuc: The other contexts are not nested though? Again, this is replicated from above. Personally I find the memo definitions being in "close" proximity to where they are used can be really helpful. Though I see how it is kind of a mess in this file. If you feel strongly about it I can try a bigger refactor.

bolt-modules/boltlib/spec/functions/set_config_spec.rb
donoghuc: yeah, copy paste problems. deleted.

bolt-modules/boltlib/spec/functions/set_config_spec.rb
donoghuc: Updated.

spec/bolt/inventory/inventory2_spec.rb
lucywyman: I was suggesting adding a new test here, either directly testing `parse_uri` or exercising the `parse_uri` call in `inventory2` in addition to `group2`. I don't think it's worth holding up this PR for though.

lib/bolt/target.rb
donoghuc: @puppetlabs/bolt Open question

lib/bolt/inventory/inventory2.rb
donoghuc: @puppetlabs/bolt Open question

lib/bolt/inventory/inventory2.rb
donoghuc: @puppetlabs/bolt Open question

pre-docs/inventory_version2.md
donoghuc: @puppetlabs/bolt Open question

pre-docs/inventory_version2.md
donoghuc: @puppetlabs/bolt Open question

lib/bolt/inventory/inventory2.rb
donoghuc: @puppetlabs/bolt Open question

bolt-modules/boltlib/spec/functions/add_facts_spec.rb
donoghuc: @puppetlabs/bolt Open question

bolt-modules/boltlib/lib/puppet/functions/run_plan.rb
```diff
@@ -118,4 +134,31 @@ def run_plan(scope, plan_name, named_args = {})
       result
     end
   end
+
+  # Recursively examine the type_expr to build a list of types
+  def extract_parameter_types(type_expr, acc = [])
```
nicklewis: This is a little weird since it's both using an accumulator _and_ calling `map`. Do you expect the recursive calls to return their values and be aggregated by the caller or do you expect them to add to `acc`?

I think that if the `acc << value` lines were replaced by `value` and the `map` calls were replaced by `flat_map`, there wouldn't be a need for an accumulator at all.

bolt-modules/boltlib/lib/puppet/functions/run_plan.rb
```diff
@@ -85,6 +85,22 @@ def run_plan(scope, plan_name, named_args = {})
...
+
+    if inventory.version > 1
+      param_types = closure.parameters.each_with_object({}) do |param, param_acc|
+        param_acc[param.name] = extract_parameter_types(param.type_expr).flatten
+      end
+      params.each do |param, value|
+        # Note the safe lookup operator is needed to handle case where a parameter is passed to a
+        # plan that the plan is not expecting
+        if param_types[param]&.include?('TargetSpec') || param_types[param]&.include?('Boltlib::TargetSpec')
+          inventory.ensure_in_all_group(value)
```
nicklewis: Is this different than `inventory.get_targets(value)` other than doing a bit less computation?

pre-docs/inventory_version2.md
lucywyman: Oh I think we can just ping @cwcowellshah on this - was just stating that the formatting may seem weird, but I like it as it is.

lib/bolt/inventory/group2.rb
```diff
@@ -152,28 +155,35 @@ def add_target(target)
...
         # This check prevents plugins from returning plugins
         raise ValidationError.new("Cannot set target with plugin", @name) if target.key?('_plugin')
         target.each do |k, v|
           next if k == 'config'
           validate_config_plugin(v, k, @name)
         end
 
-        target['name'] ||= target['uri']
+        unless target['name']
+          target['name'] = target['uri']
```
nicklewis: Is there a reason we can't just use a variable here instead of putting the `name` field on the target? Then we don't have to delete it below. I may be missing somewhere that we use the whole target hash though...

nicklewis: We do in fact put it in `t_name`, but only at the end of the function. Can we do that at the top of the function and refer to `t_name` everywhere?

lib/bolt/inventory/inventory2.rb
```diff
@@ -219,102 +198,199 @@ def expand_targets(targets)
...
+      # This is effectively the init method for Target2
+      def create_target(target_name, target_hash = nil)
+        # Prefer target hash, then data from inventoryfile, allow for uri only with empty hash
+        data = target_hash || @groups.target_collect(target_name) || {}
+        data = { 'uri' => target_name } if data['uri'].nil? && data['name'].nil?
+        data['uri_obj'] = Bolt::Inventory::Inventory2.parse_uri(data['uri'])
+
+        if data['uri'] && data['name'].nil?
+          data['name'] = data['uri']
+          data['safe_name'] = data['uri_obj'].omit(:password).to_str.sub(%r{//}, '')
```
nicklewis: Why do we remove the `//`?

nicklewis: Is this for the case where we have a bare leading `//`? If so, then this should be `%r{^//}`

bolt-modules/boltlib/lib/puppet/functions/run_plan.rb
donoghuc: The main difference is that get_targets will add a target to inventory, but it will not add it to the all group. Effectively an inventory's `@target` hash will have the new target, but that does not necessarily add the target to the all group. My understanding was that we only add a target to all with `Target.new`, `get_target` or if a plan has a `TargeSpec` parameter. This is based off the "riddle" i posted https://puppet.slack.com/archives/CF466J51Q/p1566581349473200

donoghuc:
```
cas donoghue 10:29 AM
With an empty inventory:
version: 2
a plan like:
plan test(TargetSpec $nodes){
  notice(get_targets('all')
}
And an invocation:
bolt plan run test -t localhost
Will localhost be in the targets array?
Nick Lewis:boltlogo: 10:31 AM
are you the sphinx??
:rolling_on_the_floor_laughing:
1

I’m going with no, but I’m sure not confident about it!
David Kramer:boltlogo: 10:32 AM
BoltTeasers™
Nick Lewis:boltlogo: 10:33 AM
I would potentially think it will not be in the inventory, but if you call $nodes.get_targets first then it will be in the inventory for get_targets('all')
But I also think that is not the actual behavior
cas donoghue 10:37 AM
I think this relates to Target.new in a plan. It is pretty easy to add a new target in inventory with get_target, but Target.new does not yet have an inventory attribute. It would be pretty easy to add $nodes to inventory but if they are not in inventory, why would Target.new be?
Nick Lewis:boltlogo: 10:40 AM
I distinctly remember having discussions about this, but I don’t remember the conclusion.
I think calling Target.new should add the target to the inventory.
Calling it from a plan, specifically.
```
I guess I just read the frist answer ("i'm going with no") and rolled with it. Reading it again it seems like we (at least in that thread) did not come to a conclusion other than me implementing it the way I did.

donoghuc: Though this example would be in inventory. Sorry wrong link. I don't remember when we talked about get_targets. Maybe I made it up.

lib/bolt/inventory/inventory2.rb
donoghuc: Yes, it is to remove the `//` we append in `parse_uri` when we add it to initialized the addressable URI object with an empty scheme to ensure we parse the hostname correctly. I updated to only remove the `//` when it is the first two chars. Good catch.

lib/bolt/inventory/group2.rb
donoghuc: Oh, yeah, that makes way more sense. That is a much better approach. Updated.

bolt-modules/boltlib/lib/puppet/functions/run_plan.rb
donoghuc: Wow, yeah that makes way more sense. Thanks, this method just became more and more complex I think I got so focused on making it correct that I forgot about making it readable. Udpated to not use acc.

bolt-modules/boltlib/spec/functions/add_facts_spec.rb
adreyer: Lets ticket this separately or add it as a requirement to the merge inventory versions work

bolt-modules/boltlib/lib/puppet/functions/run_plan.rb
nicklewis: I see. I think the context around that question _at the time_ was whether targets provided on the CLI are automatically added to the inventory at all (irrespective of the "all" group). Since we later (IIRC) decided that arguments of type `TargetSpec` should be automatically added to the inventory, the behavior of that example changes.

bolt-modules/boltlib/spec/functions/add_facts_spec.rb
donoghuc: https://github.com/puppetlabs/bolt/issues/1257

spec/integration/inventory2_spec.rb
```diff
@@ -605,7 +605,7 @@ def fact_plan(name = 'facts_test')
           result = run_cli_json(run_command)
           expect(result).not_to include('kind')
           # If puppetdb config loaded this would be fake
-          expect(result['items'][0]['result']['stdout']).to eq("#{conn[:system_user]}\n")
+          expect(result['items'][0]['result']['stdout']).to eq("#{conn[:user]}\n")
```
donoghuc: This is an interesting side effect of adding any target to the all group with get_targets. In this case the inventoryfile looks like:
```
---
version: 2
targets:
- _plugin: puppetdb
  query: inventory { facts.fact1 = true }
  config:
    ssh: {}
config:
  transport: ssh
  ssh:
    host: localhost
    user: bolt
    port: 20022
    host-key-check: false
```
previously `run command whoami -t localhost` would use the local transport (the magic of localhost defaults) because there is no `localhost` target in inventory, however now that it is added to the all group it picks up the transport config on the `all` group and the configured ssh transport is used.

bolt-modules/boltlib/lib/puppet/functions/run_plan.rb
donoghuc: Updated

spec/integration/inventory2_spec.rb
adreyer: ugh that seems like a problem

adreyer: I think this is ok if it was previously the case when `--transport` was passed or transport `bolt.yaml`

donoghuc: does not appear to be the case:
```
cas@cas-ThinkPad-T460p:~/working_dir/bolt/Boltdir$ cat bolt.yaml
---
transport: winrm
cas@cas-ThinkPad-T460p:~/working_dir/bolt/Boltdir$ bolt command run whoami -t localhost
Started on localhost...
Finished on localhost:
  STDOUT:
    cas
Ran on 1 node in 0.0 sec
cas@cas-ThinkPad-T460p:~/working_dir/bolt/Boltdir$ bolt command run whoami -t localhost --transport winrm
Started on localhost...
Finished on localhost:
  STDOUT:
    cas
Ran on 1 node in 0.0 sec
```
Do we need to make `localhost` even more special?

donoghuc: (note this is based on master, not the PR branch), though that would be the same with empty inventory 2.

nicklewis:
`Target2` needs a `target_alias` method, since it has that defined on the datatype. Otherwise `String($target)` doesn't work.

But if I add that method, then `String($target)` returns something like this:
```
Target({'uri' => 'bzinxhsh79apkel.delivery.puppetlabs.net', 'name' => 'bzinxhsh79apkel.delivery.puppetlabs.net', 'target_alias' => [], 'config' => {'transport' => 'ssh', 'ssh' => {'user' => 'someone', 'password' => 'something', 'sudo-password' => 'something'}}, 'facts' => {}, 'vars' => {}, 'features' => [], 'plugin_hooks' => {}})
```

Which seems wrong.

nicklewis:
I'm not sure if there's a way to make `String($target)` _not_ default to the Puppet-y representation of the Target object... You can call it with `String($target, '%s')` which will use `to_s`, but otherwise it will pull off all the attributes and print those.

donoghuc:
Updated with a `target_alias` method. What is the concern with `String($target)` printing all the attributes? Confusion between when the `to_s` method is used vs the `String` method? 
Or is the concern that it would print sensitive information?

nicklewis:
The main problem is that I didn't realize that `"${target}"` has a different (more correct) behavior than `String($target)`. So I was concerned there was no simple way to get the desired output. There is, so not a problem!

You can get sensitive information out by calling `String($target)`, but that's not a big deal. If you have access to the target, you can get that info anyway.

PR #1192, opened 2019-08-30
(maint) Include hours and minutes in duration display

nicklewis:
Previously, we displayed durations as only seconds. That meant that
lengthy operations would be reported as, for instance, 123 seconds. We
now break these times down into hours and minutes where applicable,
written as:

1 hr, 2 min, 3 sec
2 min, 3 sec
3.76 sec

When the duration is under a minute, we include two decimal places for
the seconds display. Otherwise, we strip the sub-second duration and
round to the nearest second.


nicklewis:
This matches the behavior of `puppet job` / `puppet task`.

Issue #1167, opened 2019-08-20
apply() blocks ignore run-as when set by a plan


nicklewis:
If a plan contains an `apply()` block and is called by _another_ plan passing `_run_as`, the `apply()` block in the called plan will ignore `_run_as` when applying the catalog.

Example code:

```puppet
plan a() {
  run_plan(b, _run_as => 'someone')
}

plan b() {
  apply('group') { }
}
```

In this case, running plan `a` will cause the `apply()` block to be evaluated as whatever the default user for the target is, rather than as "someone".

This misbehavior _only_ happens for `_run_as` as a parameter to `run_plan`. Passing `--run-as` on the CLI or setting `run-as` in the inventory both work as expected.

This appears to happen because the applicator bypasses the executor and calls the transport directly, causing it to ignore the `run_as` variable on the executor.

nicklewis:
Closed by #1196

PR #1196, opened 2019-09-04
(GH-1167) Honor run_plan(..., _run_as => ...) from apply()

nicklewis:
Previously, apply() blocks would ignore the `_run_as` argument passed to
their containing plan.

    plan a() {
      run_plan(b, _run_as => 'someone')
    }
    plan b() {
      apply(...) { }
    }

In this case, the catalog compiled for the `apply()` block in plan b
would be executed as whatever user bolt connected to the user as instead
of the "someone" user specified via `_run_as`. This happened because the
Applicator class failed to replicate the full executor behavior of
running a task.

The Applicator now honors `run_as` if it's set on the executor, passing
it through to the transport's `batch_task` method.

This already worked fine for `run_as` set on a Target in the inventory
or for the `--run-as` flag.

### DIFF
--- a/lib/bolt/applicator.rb
+++ b/lib/bolt/applicator.rb
@@ -195,6 +195,7 @@ module Bolt
               arguments = {
                 'catalog' => Puppet::Pops::Types::PSensitiveType::Sensitive.new(catalog),
                 'plugins' => Puppet::Pops::Types::PSensitiveType::Sensitive.new(plugins),
+                '_task' => catalog_apply_task.name,
                 '_noop' => options['_noop']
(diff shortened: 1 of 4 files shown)


nicklewis:
I just played around with calling `run_task` a bit. The problem is that if we wrap it in `without_default_logging`, then we don't get the results logged. It's unclear exactly why, but I think there's some trouble with that function because these blocks run concurrently but `without_default_logging` is set at the executor level. So when one finishes and should be logged, another one turns `without_default_logging` on concurrently...

PR #1197, opened 2019-09-04
(docs) Fix puppet_library plugin hook reference

donoghuc:
Previously the documentation suggests specifying parameters with `params`, this commit updates it to use the correct `parameters` as well as fixing a reference to `nodes` in an inventory 2 file.


donoghuc:
@lucywyman I found another `params` key in `pre-docs/bolt_configuration_options.md`. Updated it, I think that should be all of them.

Issue #1205, opened 2019-09-06
Task parameters with `type` in the name are filtered out with powershell v2


donoghuc:
## Describe the Bug
When running a powershell task against a target using a powershell 2 interpreter, parameters with the string `type` in the name were filtered out and not passed to the task. 

## Expected Behavior
Parameters, for example `footype` should be passed to powershell tasks. 

## Additional Context
This will also need to be fixed in pxp agent in addition to the winrm/local windows transports.

Iristyle:
The equivalent pxp-agent fix is located at https://github.com/puppetlabs/pxp-agent/pull/772

PR #1206, opened 2019-09-06
(GH-1205) Allow powershell task parameters with `type` in the name

donoghuc:
closes https://github.com/puppetlabs/bolt/issues/1205
Previously, when parsing xml in the `Get-ContentAsJson` implementation a regex used to filter out xml elements with whos `MemberType` `Properties` had a name that was preceeded with `__` or contained the string `type`. Filtering on the string `type` caused task parameters that contain that string to be ignored, preventing users from being able to effectively pass powershell task paramters with the `type` in the name. Unfortunately, with the current implementation powershell parameters with the name `type` will be clobbered by a `Property` called type. However this commit updates the regex to *only* filter out the property that exactly matches `type`. Note that this work allows powershell task parameters with `type` in the name, but not the name `type` and only affects powershell < 2.

### DIFF
--- a/lib/bolt/transport/powershell.rb
+++ b/lib/bolt/transport/powershell.rb
@@ -193,7 +193,7 @@ PROCESS{
     return $xml.Objects | ConvertFrom-Xml
   }
   $propbag = @{}
-  foreach ($name in Get-Member -InputObject $xml -MemberType Properties | Where-Object{$_.Name -notmatch "^__|type"} | Select-Object -ExpandProperty name) {
+  foreach ($name in Get-Member -InputObject $xml -MemberType Properties | Where-Object{$_.Name -notmatch "^(__.*|type)$"} | Select-Object -ExpandProperty name) {
     Write-Debug "$Name Type: $($xml.$Name.type)" -Debug:$false
     $propbag."$Name" = Convert-Properties $xml."$name"
   }
--- a/spec/fixtures/complex_params/input.json
+++ b/spec/fixtures/complex_params/input.json
@@ -23,5 +23,6 @@
   "argtimespan": "12:34:56",
   "argguid": "74be6039-e777-45fe-aa56-3d356443e096",
   "argregex": "^http:\\/\\/",
-  "argswitch": true
+  "argswitch": true,
+  "types": "foo"
 }
--- a/spec/fixtures/complex_params/output
+++ b/spec/fixtures/complex_params/output
@@ -45,3 +45,5 @@ hello all
 True
 * ArgTimeSpan (timespan):
 12:34:56
+* types (string):
+foo
--- a/spec/fixtures/modules/sample/tasks/complex_params.ps1
+++ b/spec/fixtures/modules/sample/tasks/complex_params.ps1
@@ -46,7 +46,12 @@ param(
 
   [Parameter(Mandatory = $False)]
   [Switch]
-  $ArgSwitch
+  $ArgSwitch,
+
+  # Parameters may have the string `type` in them, but not have the name `type`
+  [Parameter(Mandatory = $False)]
+  [String]
+  $types
 )
 
 function ConvertTo-String


donoghuc:
Where should we document this?

lib/bolt/transport/powershell.rb
```diff
@@ -195,7 +195,7 @@ def shell_init
   $propbag = @{}
   foreach ($name in Get-Member -InputObject $xml -MemberType Properties | Where-Object{$_.Name -notmatch "^__|\\btype\\b"} | Select-Object -ExpandProperty name) {
     Write-Debug "$Name Type: $($xml.$Name.type)" -Debug:$false
-    $propbag."$Name" = Convert-Properties $xml."$name"
+    if ("$($xml.$Name)" -eq "System.Xml.XmlElement") { $propbag."$Name" = Convert-Properties $xml."$name" }
```
donoghuc: I want to do something like `if ( $xml.$Name -is System.Xml.XmlElement )` but that seems to be wrong. Converting to the string representation is just a hack to show the idea.

lib/bolt/transport/powershell.rb
```diff
@@ -195,7 +195,7 @@ def shell_init
   $propbag = @{}
   foreach ($name in Get-Member -InputObject $xml -MemberType Properties | Where-Object{$_.Name -notmatch "^__|\\btype\\b"} | Select-Object -ExpandProperty name) {
```
donoghuc: doh, i need to delete the regex part.

Iristyle: Can we just change the regex from `"^__|type"` to `"^(__|type)"`?

lib/bolt/transport/powershell.rb
Iristyle: You want to make sure the object instance is a `System.Xml.XmlElement`? Is that necessary?

You'd do that like:

```powershell
if ($xml -is System.Xml.XmlElement)
```

lib/bolt/transport/powershell.rb
donoghuc: I think we want to avoid the regex because it seems like there is always a key with the `name` `type` and it's value `object` (not sure how to dig deeper in to what *kind* of object). However, each of the `$xml.$Name` objects we want always have the type `[Xml.XmlElement]`. So the idea is to filter out the parameters prepended with underscores and then add only those key/value to `propbag` when the value is an xml element.

donoghuc: Actually, `type` is still lost with this approach. Somewhere the key `type` gets overridden. @Iristyle Do you see a way around that or should we just update the regex and then document that powershell parameters cannot be named `type` with powershell 2?

Iristyle: I'm not sure I'm totally following, but I don't think you want to remove the omission of `type` here.

lib/bolt/transport/powershell.rb
Iristyle: I forgot to put the brackets on earlier that you need to qualify a .NET type in PowerShell. Comment ^ updated

lib/bolt/transport/powershell.rb
donoghuc: So, any parameter that is *not* called `type` is a System.Xml.XmlElement. For some reason parameters with the name type are 

If you modify the following debug line 
https://github.com/puppetlabs/bolt/blob/49cec2cd552519f278bc843aaa763766c5a3905c/lib/bolt/transport/powershell.rb#L197 to print out
`"$Name Type: $($xml.$Name)"`
The type parameter appears like:
`type Type: object type`
Where any other parameter (for example `types`) appears as expected:
`types Type: System.Xml.XmlElement`
What is curious is that when there is no parameter named `type` there is always an entry for:
`type Type: object`
So my thought is, that somehow the `type` parameter (when specified in powershell sig) is getting overwritten/merged with the inherent `$xml.$Name` `type` that is  always present.

Iristyle:
FYI - some of this PS code originated in https://github.com/puppetlabs/puppetlabs-iis/pull/33/files#diff-296a373840c79ef86386de692316379bR343

lib/bolt/transport/powershell.rb
donoghuc: Thanks, `[Xml.XmlElement]` did the trick, but of course we ditched that idea :rofl:

lib/bolt/transport/powershell.rb
donoghuc: With this implementation parameters with the name `type` will be clobbered. However, updating the regex allows parameters that contain the string `type`.

Iristyle:
Equivalent pxp-agent fix in https://github.com/puppetlabs/pxp-agent/pull/772

PR #1217, opened 2019-09-12
(maint) Stub publish_event in mock-executor

donoghuc:
This commit updates the mock executor to allow users to mock out an event collector for capturing messages from `out::message` when stubbing out plans with BoltSpec::Plans.

### DIFF
--- a/lib/bolt_spec/plans.rb
+++ b/lib/bolt_spec/plans.rb
@@ -63,6 +63,8 @@ require 'bolt/pal'
 # - allow_upload(file), expect_upload(file): expect the identified source file
 # - allow_apply_prep: allows `apply_prep` to be invoked in the plan but does not allow modifiers
 # - allow_apply: allows `apply` to be invoked in the plan but does not allow modifiers
+# - allow_out_message, expect_out_message: expect a message to be passed to out::message (only modifiers are
+#   be_called_times(n), with_params(params), and not_be_called)
 #
 # Stub modifiers:
 # - be_called_times(n): if allowed, fail if the action is called more than 'n' times
@@ -127,6 +129,12 @@ require 'bolt/pal'
 #         Bolt::ResultSet.new(targets.map { |targ| Bolt::Result.new(targ, {'result_key' => 10'})})
 #       end
 #       expect(run_plan('my_plan', { 'param1' => 10 })).to eq(10)
+#     end
+
+#     it 'expects multiple messages to out::message' do
+#       expect_out_message.be_called_times(2).with_params(message)
+#       result = run_plan(plan_name, 'messages' => [message, message])
+#       expect(result).to be_ok
 #     end
 #   end
 #
@@ -219,6 +227,15 @@ module BoltSpec
     def allow_get_resources
       allow_task('apply_helpers::query_resources')
       nil
+    end
+
+    def allow_out_message
+      executor.stub_out_message.add_stub
+    end
+    alias allow_any_out_message allow_out_message
+
+    def expect_out_message
+      allow_out_message.expect_call
     end
 
     # Example helpers to mock other run functions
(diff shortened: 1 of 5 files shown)


jacobisaliveandwell:
Why not actually print the message?
```
      def publish_event(opts)
        if opts[:type] == :message
          puts opts[:message]
        end
      end
```

donoghuc:
The `publish_event` method is used for more than just printing with `out::message`. It will behave differently depending on the outputter (for example with human outputter the message will go to STDOUT vs STDERR with the json outputter). I should note also that subbing out the `publish_event` with a no-op will also fix other plan functions that require it to be there (for example, without this fix `wait_until_available` would not work either).

jacobisaliveandwell:
Is that `type` check not sufficient to not clobber those requirements? I found this very useful while debugging a plan.

donoghuc:
@jacobisaliveandwell I updated the PR to allow users to define an event collector to inspect what is published with `publish_event`. The following test shows an example of how that could be used https://github.com/puppetlabs/bolt/pull/1217/commits/e8f1f5d708700bd5c520e2092aa0c713fc635b98#diff-76f684696b6c03f64574b63da305b617R225-R235 . Does that work for you?

lib/bolt_spec/plans.rb
```diff
@@ -1,9 +1,11 @@
 # frozen_string_literal: true
 
 require 'bolt_spec/plans/mock_executor'
+require 'bolt_spec/event_collector'
```
donoghuc: I realize this is currently failing outside of bolt/spec tests. If this approach is approved i think we will want to ship it under `/lib/bolt_spec`?

lib/bolt_spec/plans.rb
```diff
@@ -164,6 +166,11 @@ def inventory
       @inventory ||= Bolt::Inventory.new({})
     end
 
+    # Override in your tests
```
adreyer: I don't think we should expose this to test authors. I think we should expose a way to make assertions about what `out::message` calls occur in a plan.

One option for that would be to implement our own event publisher and write allow/expect methods for it.
Another option would be to add an executor method that `out::message` calls to the executor and use that for testing.

jacobisaliveandwell:
@donoghuc unfortunately I'm not versed enough to comment beyond saying I was using `out::message` for extremely simplistic debugging; just wanted to print something for a few minutes and then remove it. I tried `puppet-debugger`, but this was a much lighter way to iterate quickly. So, if there was an equivalent `out::debug` i'd probably be just as happy.

As it is, I patched using my above snippet and was able to wrap up my work in an hour, when I had previously spent an entire day on the problem.

adreyer:
@jacobisaliveandwell The `out` functions are intended to be used during production plan runs and having it print during successful tests would pollute test output.

I suggest you use something specific to debugging tests like
```
Puppet::Functions.create_function(:debug::puts) do
  dispatch :putter do
    param 'Any', message 
  end
  def putter(message)
    puts messsage
  end
end
```

I usually use this pry function for tests.

```
require 'pry'

Puppet::Functions.create_function(:pry_bind, Puppet::Functions::InternalFunction) do
  def pry_bind(*args)
    # rubocop:disable Debugger
    binding.pry
    # rubocop:enable Debugger
  end 
end
```

lib/bolt_spec/plans.rb
donoghuc: not relevant anymore

jacobisaliveandwell:
@adreyer Could you elaborate on your suggested solution? I have no idea where'd you put that function, how it'd be called, how it plugs into the testing system, etc.

Disclaimer: I don't come from a ruby background, and most of what I see seems "magical" leaving me completely confused.

PR #1219, opened 2019-09-12
(maint) Fix verbose logging and catch_errors

kris-bosland:
Fixes 'no info with --verbose' to 'no info without --verbose'
Adds missing '_catch_errors' line

### DIFF
--- a/docs/_lab/10-deploying-an-application/index.md
+++ b/docs/_lab/10-deploying-an-application/index.md
@@ -316,7 +316,7 @@ Bolt logs every action on each node which results in 5 messages for each node.
 By default these messages are logged at the `notice` level and can make it hard
 to see the more useful `notice` messages that the plan logs directly. To make the
 automatic messages log at `info` instead so they won't appear on the terminal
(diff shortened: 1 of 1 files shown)


docs/_lab/10-deploying-an-application/index.md
```diff
@@ -316,7 +316,7 @@ Bolt logs every action on each node which results in 5 messages for each node.
 By default these messages are logged at the `notice` level and can make it hard
 to see the more useful `notice` messages that the plan logs directly. To make the
 automatic messages log at `info` instead so they won't appear on the terminal
-with the `--verbose` flag the plan wraps the loop in a
```
donoghuc: I think we need to revisit this entire paragraph. Will talk with the team about it. Thanks for the fix.

lucywyman: Hey @kris-bosland - thanks so much for submitting this!! Would you mind making this line "...instead so they only appear on the terminal with the `--verbose` flag the plan wraps the loop in a..." so we avoid the double negative? Other than that I think this is GTG

Issue #1120, opened 2019-08-06
Unexpected behavior of upload_file() when uploading directories


desolat:
The following behavior of `upload_file()` is rather suprising to most users: when the target directory does not exist, it copies the contents of the source directory (which is what I was expecting) to the automatically created target directory. If the target folder already exists, an additional folder named after the target folder gets created in the target directory and the source directory contents get copied there!

Looking at the [code](https://github.com/puppetlabs/bolt/blob/master/lib/bolt/transport/sudoable.rb#L38-L56) this originates from the usage of the corelibs [mv](https://www.gnu.org/software/coreutils/manual/html_node/Target-directory.html#Target-directory) command. But this implementation detail should be hidden from the user. 

I'd suggest the following:
It should be defined and documented if the command silently overwrites (files and also directories) or does not and then fails. Currently, for files it already does silent overwriting (`mv -f`) but does not for directories. I'd prefer fail-if-exists as default and an option for overwriting. The `-t` and `-T` options may help to accomplish this but maybe the coreutils are not the right tool for that aim.

adreyer:
`-T` is only in gnu mv not bsd so we can't rely on it. 

Failing if the file already exists would be a pretty big breaking change and while its safer I don't think it's the most common use case.

It makes sense to change the directory case to overwrite by default to match the file behavior. I think we could expose this change with a feature flag for the rest on the 1.x bolt series.

MikaelSmith:
How is overwriting instead of copying-into less of a breaking change than failing if copying a directory and the target exists?

adreyer:
That specifically was about files not directories. The goal seems to be to have the behavior the same. Having directories error if they already exist and files just overwrite seems barely better than the current behavior.

adreyer:
I suppose that would be different behavior anyway.

`upload_file('../file.txt', '~/foo')` will copy into foo if it's a directory and replace foo if it's a file.
`upload_file('../file/', '~/foo')` would always replace in that case.

lucywyman:
To do:
- Check `upload_file` behavior on other transports
- Possibly have `rsync` be an option for the SSh transport (similar to `smb` option for `winrm`)

@desolat Just fyi, we're likely not going to prioritize this soon, but we'll leave it open for now.

gucki:
I tried to upload a directory and the current functionality seems totally broken if the target directory already exists. The name of the source directory is always used for a subfolder in the target directory.

[...]

PR #1220, opened 2019-09-13
(ORCH-2340) Add `upload_file` endpoint to bolt-server

smcclellan:
In cases where PXP Agent is not installed on a target node
and thus cannot upload a file, SSH or WinRM should be used.
This commit adds the `POST /upload_file` endpoint which will
delegate either to SSH or to WinRM accordingly.

### DIFF
--- a/developer-docs/bolt-api-servers.md
+++ b/developer-docs/bolt-api-servers.md
@@ -176,6 +176,78 @@ For example, the following runs "echo 'hi'" on localhost:
 }
 ```
 
+### POST /ssh/upload_file
+- `target`: [SSH Target Object](#ssh-target-object), *required* - Target information to run task on.
+- `files`: Object, *required* - Which file(s) to upload, and where.
+    - `relative_path`: String, *required* - The destination for the file.
+    - `uri`: Object, *required* - The location of where to find the file.
+        - `path`: String, *required* - The endpoint to retrieve the file.
+        - `params`: Object, *required* - The parameters to supply the endpoint.
+    - `sha256`: String, *required* - The SHA256 value for the file.
+    - `kind`: String, *required* - Whether the file is a `file` or `directory`.
+- `job_id`: Integer, *required* - An identifier for the job. If this matches an existing ID, cached files may be returned.
+- `destination`: String, *required* - Where to put the files on the target machine.
+
+For example, the following uploads file 'abc' on linux_target.net:
+```
+{
+  "target": {
+    "hostname": "linux_target.net",
+    "user": "marauder",
+    "password": "I solemnly swear that I am up to no good",
+    "host-key-check": false,
+    "run-as": "george_weasley"
+  },
+  "destination": "/home/bolt/root",
+  "job_id": 1,
+  "files" : [{
+    "relative_path": "file.sh"
+    "uri": {
+      "path": "/puppet/v3/file_content/modules/some_module/file.sh",
+      "params": {}
+    },
+    "sha256": "SHA256VALUE",
+    "kind": "file"
+  }]
+}
+```
+
+### POST /winrm/upload_file
+- `target`: [WinRM Target Object](#winrm-target-object), *required* - Target information to run task on.
+- `files`: Object, *required* - Which file(s) to upload, and where.
+    - `relative_path`: String, *required* - The destination for the file.
+    - `uri`: Object, *required* - The location of where to find the file.
+        - `path`: String, *required* - The endpoint to retrieve the file.
+        - `params`: Object, *required* - The parameters to supply the endpoint.
+    - `sha256`: String, *required* - The SHA256 value for the file.
+    - `kind`: String, *required* - Whether the file is a `file` or `directory`.
+- `job_id`: Integer, *required* - An identifier for the job. If this matches an existing ID, cached files may be returned.
+- `destination`: String, *required* - Where to put the files on the target machine.
+
+For example, the following uploads file 'abc' on windows_target.net:
+```
+{
+  "target": {
+    "hostname": "windows_target.net",
+    "user": "Administrator",
+    "password": "Secret",
+    "ssl": false,
+    "ssl-verify": false
+  },
+  "destination": "C:\\Users\\bolt\\root",
+  "job_id": 1,
+  "files" : [{
+    "relative_path": "file.exe"
+    "uri": {
+      "path": "/puppet/v3/file_content/modules/some_module/file.exe",
+      "params": {}
+    },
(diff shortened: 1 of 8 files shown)


lib/bolt_server/transport_app.rb
```diff
@@ -73,6 +73,38 @@ def run_command(target, body)
...
+        if kind == 'file'
+          # The parent should already be created by `directory` entries,
+          # but this is to be on the safe side.
+          parent = File.dirname(path)
+          FileUtils.mkdir_p(parent)
+          @file_cache.download_file(path, sha256, uri)
+        elsif kind == 'directory'
+          # Create directory in cache so we can move files in.
+          FileUtils.mkdir_p(path)
+        else
```
speedofdark: does this need to worry about the symlink support that @McdonaldSeanp just added?

spec/bolt_server/app_integration_spec.rb
```diff
@@ -99,5 +99,40 @@ def app
         expect(result['result']["stderr"]).to match(/not-a-command/)
       end
     end
+
+    describe "upload_file" do
```
smcclellan: Any other test cases I should add here?

spec/fixtures/modules/upload_file/files/subdir/sub-file.sh
```diff
@@ -0,0 +1 @@
+Sub-content!
```
speedofdark: Another approach would be to embed some heredocs to create whatever filesystem objects you might need?

spec/bolt_server/app_integration_spec.rb
```diff
@@ -99,5 +99,40 @@ def app
...
+      let(:path) { '/ssh/upload_file' }
```
speedofdark: do we need to worry about a discrete test for WinRM?

developer-docs/bolt-api-servers.md
```diff
@@ -176,6 +176,78 @@ For example, the following runs "echo 'hi'" on localhost:
...
+}
+```
+
+### POST /winrm/upload_file
+- `target`: [WinRM Target Object](#winrm-target-object), *required* - Target information to run task on.
+- `files`: Object, *required* - Which file(s) to upload, and where.
+    - `relative_path`: String, *required* - The destination for the file.
+    - `uri`: Object, *required* - The location of where to find the file.
+        - `path`: String, *required* - The endpoint to retrieve the file.
+        - `params`: Object, *required* - The parameters to supply the endpoint.
```
speedofdark: do we need to define/declare any semantics around this Object?

spec/bolt_server/app_integration_spec.rb
smcclellan: If the directory already exists, this weird behavior happens: https://github.com/puppetlabs/bolt/issues/1120

developer-docs/bolt-api-servers.md
smcclellan: I will add that it requires string keys and values.

lib/bolt_server/transport_app.rb
smcclellan: Per @McdonaldSeanp, the links will appear as normal files and directories here, so I'm under the assumption that link support is not needed.

spec/fixtures/modules/upload_file/files/subdir/sub-file.sh
smcclellan: That seems about equivalent to using fixtures here, to me. If anything, on-disk fixtures are easier to find/debug.

spec/bolt_server/app_integration_spec.rb
speedofdark: So does your implementation mirror this behaviour?

smcclellan: Yes, because we want the implementation of this to mirror Bolt's behavior as closely as possible. When that issue gets fixed, it will be fixed here too. (It's inside `sudoable.rb`.)

spec/bolt_server/app_integration_spec.rb
smcclellan: I'm not sure if we're set up for that, as the other endpoints aren't tested that way.

speedofdark: So does anywhere you've looked test WinRM vs SSH in this type of context?

smcclellan: Maybe that's better handled on the orchestrator acceptance/integration side? @McdonaldSeanp ideas here?

smcclellan: The docker tests in this file won't be compatible with Windows, so testing likely needs a more full stack.

McdonaldSeanp: No, we don't need discrete unit testing for WinRM. There are no differences in this level of the code between ssh/winrm as far as the execution goes.

Acceptance testing for bolt will cover that the bolt level code (the level below this puma API) that we call here works over WinRM, and acceptance testing for orchestrator will test the integration of everything works for WinRM hosts.

lib/bolt_server/transport_app.rb
```diff
@@ -73,6 +73,38 @@ def run_command(target, body)
...
+        uri = file['uri']
+        sha256 = file['sha256']
+        kind = file['kind']
+        path = File.join(cache_dir, relative_path)
+        if kind == 'file'
+          # The parent should already be created by `directory` entries,
+          # but this is to be on the safe side.
+          parent = File.dirname(path)
+          FileUtils.mkdir_p(parent)
+          @file_cache.download_file(path, sha256, uri)
```
McdonaldSeanp: This all looks great. It looks like for file downloads we might be able to take advantage of file checks from the file_cache object/class to see if the file has already been downloaded: https://github.com/puppetlabs/bolt/blob/master/lib/bolt_server/file_cache.rb#L106, which could save a bunch of requests to puppetserver as this endpoint will be called multiple times for the same job.

smcclellan: Doesn't `download_file` already run `check_file` as a first step? So if it's already in the cache, it won't hit the endpoint?

smcclellan:
```
    def download_file(file_path, sha, uri)
      if check_file(file_path, sha)
        @logger.debug("File was downloaded while queued: #{file_path}")
        return file_path
      end
      ...
```

McdonaldSeanp: Oh, hehe, yes it does do that, nevermind :)

PR #1221, opened 2019-09-14
(ORCH-2341) Add check_node_connections endpoint to bolt_server

caseywilliams:
Adds a check_node_connections endpoint for use in the Puppet Enterprise orchestrator's implementation of wait_until_available. Individual commit messages have details.

### DIFF
--- a/spec/bolt_server/transport_app_spec.rb
+++ b/spec/bolt_server/transport_app_spec.rb
@@ -151,6 +151,29 @@ describe "BoltServer::TransportApp" do
         expect(last_response).to be_ok
         expect(last_response.status).to eq(200)
       end
+
+      it 'expects either a single target or a set of targets, but not both' do
+        single_target = {
+          hostname: target[:host],
+          user: target[:user],
+          password: target[:password],
+          port: target[:port]
+        }
+        body = { target: single_target }
+
+        post(path, JSON.generate(body), 'CONTENT_TYPE' => 'text/json')
+        expect(last_response.status).to eq(200)
+
+        body = { targets: [single_target] }
+
+        post(path, JSON.generate(body), 'CONTENT_TYPE' => 'text/json')
+        expect(last_response.status).to eq(200)
+
+        body = { target: single_target, targets: single_target }
+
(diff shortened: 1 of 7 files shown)


caseywilliams:
I'd like to add one more spec tests to confirm endpoints that use `target` will not accept `targets` - I'll unblock once I do that.

spec/bolt_server/transport_app_spec.rb
```diff
@@ -222,26 +245,74 @@ def file_data(file)
...
+        expect(last_response.status).to eq(200)
+
+        body = { targets: [single_target] }
+
+        post(path, JSON.generate(body), 'CONTENT_TYPE' => 'text/json')
+        expect(last_response.status).to eq(200)
+
+        body = { target: single_target, targets: single_target }
+
+        post(path, JSON.generate(body), 'CONTENT_TYPE' => 'text/json')
```
donoghuc: With regards to https://github.com/puppetlabs/bolt/pull/1221/commits/58adf03e13caa3f69f5c864faa899d757652be40 , I think its pretty bad that we regressed on that so bad. (for example with a typo in the schema you would get)
```
cas@cas-ThinkPad-T460p:~/working_dir/bation/json" -d @request.json --cacert $BOLT_CACERT --cert $BOLT_CERT --key $BOLT_KEY $BOLT_ROOT/winrm/run_task
{"kind":"boltserver/server-error","msg":"500: Unknown error: undefined method `status_hash' for 400:Integer","details":{}}
```
I see that we are expecting a status 400 (where it would have been 500 before), maybe we should also do some regex matching on the body of the response `expect(last_response.body).to match(/type object matched more than one of the required schemas/)` to ensure we are surfacing the message from the schema validation. Or possibly we need to just test schema validation error handling somewhere more generally.

caseywilliams: I've added one test each for SSH and WinRM as you describe, here:
https://github.com/puppetlabs/bolt/pull/1221/files#diff-3afac83384bfe4aaafb92f6c7c6d5225R373
Anything else you'd like to see in this PR?

donoghuc: Looks great. Thanks!

PR #1222, opened 2019-09-14
Add a newline to the site_content on lesson 11

ClashTheBunny:
When `curl`ing from the command line, it is easy to have your output
swallowed up by a fancy prompt.  It's probably cleaner anyhow.


lucywyman:
Thanks for submitting! Looks good to me

PR #1225, opened 2019-09-17
(ORCH-2341) Fix aggregate results reporting in bolt_server

caseywilliams:
Fixes a bug I introduced yesterday in #1221:

Previously, the conversion to JSON of bolt_server responses that use
`targets` instead of `target` was broken - it did not correctly format
the results as a parsable array.

This change fixes that, and instead of returning just the result list,
aggregate endpoints now return their own status hash, with two entries:

- 'status': will be 'success' only if all targets succeeded. If at least
  one target failed, this will also be set to 'failure', even though the
  request itself was successful and returns 200.
- 'result': contains the array with individual status hashes for each target.

### DIFF
--- a/developer-docs/bolt-api-servers.md
+++ b/developer-docs/bolt-api-servers.md
@@ -248,6 +248,56 @@ For example, the following uploads file 'abc' on windows_target.net:
 }
 ```
 
+### POST /ssh/check_node_connections
+- `targets`: An array of [SSH Target Objects](#ssh-target-object), *required* - A set of targets to check once for connectivity over SSH.
+
+Example request for a single connectivity check on two nodes over SSH:
+```
+{
+  "targets": [
+      {
+        "hostname": "first.ssh_node.net",
+        "user": "marauder",
+        "password": "I solemnly swear that I am up to no good",
+        "host-key-check": false,
+        "run-as": "george_weasley"
+      }, {
+        "hostname": "second.ssh_node.net",
+        "user": "marauder",
+        "password": "I solemnly swear that I am up to no good",
+        "host-key-check": false,
+        "run-as": "fred_weasley"
+      }
+  ]
+}
+```
+
+This returns a JSON object of the shape:
+```
+{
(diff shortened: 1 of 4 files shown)


caseywilliams:
It looks like we may need https://github.com/puppetlabs/bolt/pull/1227 to fix the travis failure

caseywilliams:
Sorry, I haven't updated the docs here. I'll unblock this once I do that.

caseywilliams:
Ok, docs added!

developer-docs/bolt-api-servers.md
```diff
@@ -248,6 +248,56 @@ For example, the following uploads file 'abc' on windows_target.net:
...
+      }
+  ]
+}
+```
+
+This returns a JSON object of the shape:
+```
+{
+    "status": "success",
+    "results": [
```
caseywilliams: Other status hashes for single-target actions like run_task use "result" here, not "results" - does it make more sense to use "result" here for consistency?

donoghuc: Oh, Yeah I think so.

developer-docs/bolt-api-servers.md
```diff
@@ -248,6 +248,56 @@ For example, the following uploads file 'abc' on windows_target.net:
...
+      }
+  ]
+}
+```
+
+This returns a JSON object of the shape:
+```
+{
+    "status": "success",
+    "result": [
```
caseywilliams: Would it be more helpful to return an object with node names as keys and result hashes as values here?

developer-docs/bolt-api-servers.md
caseywilliams: thanks, updated to use `result`

caseywilliams:
(force-pushed to update commit message)

developer-docs/bolt-api-servers.md
branan: I think this is fine - it looks like a ResultSet

Issue #1158, opened 2019-08-16
`interpreters` with spaces fail with the winrm transport


nicklewis:
When using the `interpreters` setting on the WinRM transport, a task will fail to execute if the path to the specified interpreter contains a space. This happens because we fail to quote the executable and only quote the arguments.

This doesn't appear to be an issue with the SSH transport as we use `Shellwords.shelljoin` to join each piece into a single command string, which handles properly quoting each element.

PR #1230, opened 2019-09-18
(GH-1158) Quote paths for interpreter and task when using winrm

donoghuc:
Previously when running a task over winrm where a space was contained in a user defined `interpreter` path or the path to the temporary task executable the task could fail due to not quoting the paths to handle the spaces. This commit updates the winrm transport to quote both paths when `interpreters` is configured. Note that for the `interpreter` path, if the user has specified a quoted path, no additional quoting will be added.

closes https://github.com/puppetlabs/bolt/issues/1158

### DIFF
--- a/lib/bolt/transport/powershell.rb
+++ b/lib/bolt/transport/powershell.rb
@@ -52,16 +52,23 @@ module Bolt
           "[Environment]::SetEnvironmentVariable('#{arg}', @'\n#{val}\n'@)"
         end
 
-        def execute_process(path, arguments, stdin = nil)
-          quoted_args = arguments.map do |arg|
-            "'" + arg.gsub("'", "''") + "'"
-          end.join(' ')
+        def quote_string(string)
+          "'" + string.gsub("'", "''") + "'"
+        end
 
+        def execute_process(path, arguments, stdin = nil)
+          quoted_args = arguments.map { |arg| quote_string(arg) }.join(' ')
+
+          quoted_path = if path =~ /^'.*'$/ || path =~ /^".*"$/
+                          path
+                        else
+                          quote_string(path)
+                        end
           exec_cmd =
             if stdin.nil?
-              "& #{path} #{quoted_args}"
+              "& #{quoted_path} #{quoted_args}"
             else
-              "@'\n#{stdin}\n'@ | & #{path} #{quoted_args}"
+              "@'\n#{stdin}\n'@ | & #{quoted_path} #{quoted_args}"
             end
           <<-PS
 $OutputEncoding = [Console]::OutputEncoding
--- a/lib/bolt/transport/winrm.rb
+++ b/lib/bolt/transport/winrm.rb
@@ -163,8 +163,7 @@ module Bolt
               if Powershell.powershell_file?(remote_task_path) && stdin.nil?
                 conn.execute(Powershell.run_ps_task(arguments, remote_task_path, input_method))
               else
-                interpreter = select_interpreter(remote_task_path, target.options['interpreters'])
-                if interpreter
+                if (interpreter = select_interpreter(remote_task_path, target.options['interpreters']))
                   path = interpreter
                   args = [remote_task_path]
                 else
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,14 @@
 
 #### Bug fixes
 
-* A plan apply() incorrectly returns successful if the report is unparseable([#1241](https://github.com/puppetlabs/bolt/issues/1241))
+* **A plan apply() incorrectly returns successful if the report is unparseable** ([#1241](https://github.com/puppetlabs/bolt/issues/1241))
+
+  Unexpect results for the result of an `apply` are now treated as errors.
+
+* **`interpreters` with spaces fail with the winrm transport** ([#1158](https://github.com/puppetlabs/bolt/issues/1158))
+
+  When using the interpreters setting on the WinRM transport, a task would fail to execute if the path to the specified interpreter contains a space. Interpreter paths with spaces in the name are now supported.
+
 
 ## 1.31.1
 
--- a/spec/bolt/transport/winrm_spec.rb
+++ b/spec/bolt/transport/winrm_spec.rb
@@ -1096,6 +1096,67 @@ OUTPUT
         end
       end
 
+      context "with intperpreter path containing spaces" do
+        let(:config) { mk_config(ssl: false, interpreters: interpreter, user: user, password: password) }
+        context "with non-quoted interpreter" do
+          let(:interpreter) { { 'py' => 'C:/Program Files/Miniconda3/python.exe' } }
+          it "quotes interpreter and task executable", winrm: true do
+            expect(Bolt::Transport::Powershell)
+              .to receive(:make_tempdir)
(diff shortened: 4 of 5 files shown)


donoghuc:
If anybody (on the LAN) wants a windows node with a python interpreter containing a space in the path checkoutout this inventoryfile
```yaml
version: 2
targets:
  - name: win-space-python
    uri: uqvno7cdc341ndn.delivery.puppetlabs.net
    config:
      transport: winrm
      winrm:
        password: ****
        user: Administrator
        ssl: false
        ssl-verify: false
        tmpdir: C:/Program Files
        interpreters:
          py: C:\Program Files\Miniconda3\python.exe
```
Tested with 
```python
#!/foo

import os
import fileinput
import json

for line in fileinput.input():
    input_data = json.loads(line)
    if input_data['message']:
    	print(json.dumps({'env': os.environ['PT_message'], 'stdin': input_data['message']}))
```
```
cas@cas-ThinkPad-T460p:~/working_dir/bolt$ bolt task run test::interpreter message=hi -t win-space-python
Started on uqvno7cdc341ndn.delivery.puppetlabs.net...
Finished on uqvno7cdc341ndn.delivery.puppetlabs.net:
  {
    "env": "hi",
    "stdin": "hi"
  }
Successful on 1 node: win-space-python
Ran on 1 node in 1.77 sec
```
Also with the workaround users might be using:
```yaml
        interpreters:
          py: "'C:\\Program Files\\Miniconda3\\python.exe'"
```

lib/bolt/transport/winrm.rb
```diff
@@ -162,7 +162,11 @@ def run_task(target, task, arguments, _options = {})
                 interpreter = select_interpreter(remote_task_path, target.options['interpreters'])
                 if interpreter
                   path = interpreter
-                  args = [remote_task_path]
+                  # Quote path if not quoted
+                  unless path =~ /^'.*'$/ || path =~ /^".*"$/
+                    path = "\"#{path}\""
```
donoghuc: Should we prefer single quotes here?

glennsarti:
This has potentially bad assumption.  Interpreters can be either an array OR a string. Blindly adding quotes is going to cause issues eventually.

e.g.

https://github.com/puppetlabs/bolt/blob/0d7ad83987aa4b5a0fc5284a72e171e2aee0291b/lib/bolt/transport/sudoable/connection.rb#L94-L106

https://github.com/puppetlabs/bolt/blob/0d7ad83987aa4b5a0fc5284a72e171e2aee0291b/lib/bolt/transport/local_windows.rb#L66

and what I have to do for local docker support

``` ruby
describe Bolt::Transport::Docker, docker_wcow: true do
  # Windows Containers
  include BoltSpec::Conn
  include BoltSpec::Transport

  let(:default_transport_conf) {
    {
      'interpreters' => {
        # Unlike a linux based container, commands like echo are not binaries, they require a shell. Because Windows
        # Containers could be using either cmd.exe, powershell.exe, or even, pwsh.exe as the shell we can't assume
        # anything and instead need to be very specific about which interpreter we're going to use
        '.bat' => ['cmd.exe', '/c'],
        '.ps1' => ['powershell.exe', '-NoProfile', '-NonInteractive', '-NoLogo', '-ExecutionPolicy', 'Bypass', '-File']
      }
    }
  }

  context 'using PowerShell' do
    include_examples 'docker transport', 'docker-windows', windows_powershell_container_context
  end

  context 'using cmd.exe' do
    include_examples 'docker transport', 'docker-windows', windows_cmd_container_context
  end
end
```

And slightly less authorative ... bolt server JSON spec just uses a hash
https://github.com/puppetlabs/bolt/blob/master/lib/bolt_server/schemas/partials/target-winrm.json#L47-L50

donoghuc:
The winrm implementation does not currently handle array values for interpreters. Are you concerned that users will pass in an array value for the interpreters hash value? In that case i would expect it to blow up regardless of quoting. Would you like to see array values explicitly raise a validation error? 

Are you suggesting that in the future we need to support an array value for interpreters?

lib/bolt/transport/winrm.rb
lucywyman: Based on #1231 it seems like yes, since escaping characters in single quotes seems easier (based on reading some StackOverflow posts). @Iristyle thoughts?

glennsarti:
> The winrm implementation does not currently handle array values for interpreters. Are you concerned that users will pass in an array value for the interpreters hash value?

Then there's a bigger issue here. Some transports support arrays, some do not. Documentation doesn't suggest there are differences (and in fact was quite hard to find at all https://github.com/puppetlabs/bolt/blob/master/pre-docs/bolt_configuration_options.md#global-configuration-options) 

> Would you like to see array values explicitly raise a validation error?

This would at least be better than some obtuse error, yes.

> Are you suggesting that in the future we need to support an array value for interpreters?

1. IMO, There should be consistency for this option given it's global.  If only some "things" support array and string, and some "things" expect only a string, it's setting us up for problems.

2. Not having the ability to add arbitrary command line args may bite us in the future e.g. Both PowerShell and cmd.exe require command line args in order to "execute" something.

lib/bolt/transport/winrm.rb
nicklewis: I was thinking the place to change this would be `Powershell#execute_process`, which is what currently handles escaping the arguments. So we ought to change it to just do the same thing to the command.

lib/bolt/transport/powershell.rb
```diff
@@ -57,11 +57,16 @@ def execute_process(path, arguments, stdin = nil)
             "'" + arg.gsub("'", "''") + "'"
           end.join(' ')
 
+          quoted_path = if path =~ /^'.*'$/ || path =~ /^".*"$/
+                          path
+                        else
+                          "\"#{path}\""
```
nicklewis: Should we copy the logic from above that we use for arguments?

```rb
"'" + arg.gsub("'", "''") + "'"
```

donoghuc: I tried that, it does not work. The literal string does not execute the task.

donoghuc:
```
PS C:\program files\Miniconda3> '"C:/Program Files/Miniconda3/python.exe"'
"C:/Program Files/Miniconda3/python.exe"
```

glennsarti: You'd need to use the `&` command to force it to execute.  Whether this should be used in this case is debatable

```
C:\Source> & '"C:/Program Files/Miniconda3/python.exe"'
& : The term '"C:/Program Files/Miniconda3/python.exe"' is not recognized as the name of a cmdlet, function, script file, or operable program. Check the spelling of the name, or if
a path was included, verify that the path is correct and try again.
At line:1 char:3
+ & '"C:/Program Files/Miniconda3/python.exe"'
+   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : ObjectNotFound: ("C:/Program Files/Miniconda3/python.exe":String) [], CommandNotFoundException
    + FullyQualifiedErrorId : CommandNotFoundException

C:\Source> '"C:/Program Files/Miniconda3/python.exe"'
"C:/Program Files/Miniconda3/python.exe"
```

Issue #1164, opened 2019-08-20
WinRM: Misleading error message when cacert file cannot be found


donoghuc:
## Describe the Bug
If a cacert file is specified in bolt.yaml and the specified file cannot be found, rather than returning an obvious/helpful error, such as "Configured cacert file cannot be found.", Bolt will attempt to connect and return the following error:
```
  {
    "node": "host.example.com",
    "target": "host.example.com",
    "action": null,
    "object": null,
    "status": "failure",
    "result": {
      "_error": {
        "kind": "puppetlabs.tasks/connect-error",
        "msg": "Failed to connect to http://host.example.com:5985/wsman: system lib",
        "details": {
        },
        "issue_code": "CONNECT_ERROR"
      }
    }
  }
]
```

## Expected Behavior
When cacert is specified and the file cannot be found:
- When --no-ssl is specified, ignore the configuration item (silently succeed)
- When --ssl is specified, fail with an error that makes it clear that failure is due to misconfiguration (file cannot be found/read)

donoghuc:
Solved with https://github.com/puppetlabs/bolt/pull/1247

PR #1232, opened 2019-09-19
(GH-1164) Only try to read cacert for winrm when needed

donoghuc:
This commit adds logic to the winrm transport to not try to read the `cacert` file when not using `ssl`. This avoids confusing errors when the `cacert` is not readable but `ssl` is not being used.

### DIFF
--- a/lib/bolt/transport/winrm/connection.rb
+++ b/lib/bolt/transport/winrm/connection.rb
@@ -41,7 +41,7 @@ module Bolt
 
           transport = :kerberos if target.options['realm']
           endpoint = "#{scheme}://#{target.host}:#{@port}/wsman"
-          cacert = target.options['cacert'] && target.options['ssl'] ? target.options['cacert'] : nil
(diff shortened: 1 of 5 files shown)


donoghuc:
closes GH-1164

lib/bolt/transport/winrm/connection.rb
```diff
@@ -41,7 +41,7 @@ def connect
 
           transport = :kerberos if target.options['realm']
           endpoint = "#{scheme}://#{target.host}:#{@port}/wsman"
-          cacert = target.options['cacert'] && target.options['ssl'] ? target.options['cacert'] : nil
+          cacert = target.options['cacert'] && target.options['ssl'] ? File.expand_path(target.options['cacert']) : nil
```
donoghuc: I went ahead and added this commit on this PR because it build on the new logic here. See https://github.com/puppetlabs/bolt/pull/1233 for similar fix for orch transport.

Issue #1174, opened 2019-08-26
Configuration option(s) pointing to files do not expand `~`


donoghuc:
## Describe the Bug
When the `token-file` for pcp/puppet-db is something like `~/.puppetlabs/token`, the home directory is not expanded. This can manifest in `file not found` errors. 

## Expected Behavior
The following config should work without needing fully qualified paths.
```
puppetdb:
  server_urls: ["https://master:8081"]
  cacert: "~/.puppetlabs/client-tools/ca.pem"
  token: "~/.puppetlabs/token"
pcp:
  service-url: "https://master:8143"
  cacert: "~/.puppetlabs/client-tools/ca.pem"
  token-file: "~/.puppetlabs/token"
  task-environment: dev
```
Instead (currently) the following error will occur:
```
% bolt task run service action=status name=puppet --transport pcp -t foo
Started on foo...
Failed on foo:
  token-file '~/.puppetlabs/token' is unreadable
Failed on 1 node: foo
Ran on 1 node in 0.18 seconds
```

donoghuc:
The `pcp`  tansport options `cacert` and `token-file` will now expand paths as well as the `cacert` option for the `winrm` transport. Similarly the orchestrator_client gem will also expand the paths 
 which has been released in 0.4.3 https://rubygems.org/gems/orchestrator_client/versions/0.4.3

PR #1233, opened 2019-09-19
(GH-1174) Expand file paths for orchestrator client config

donoghuc:
Previously when the `token-file` or `cacert` file paths were used to construct a new orchestrator client instance, the paths were not expanded. Other paths in bolt expand and there are examples in the wild where paths need to be expanded (for example `token-file: ~/.puppetlabs/token`). This commit updates the orch transport to expand file paths before creeting an Orchestrator client object.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 
   When using the winrm transport do not try to read the cacert file when not using ssl. This avoids confusing errors when the cacert is not readable but ssl is not being used.
 
+* **Some configuration options would not support file path expansion** ([#1174](https://github.com/puppetlabs/bolt/issues/1174))
+
+  The `token-file` and `cacert` file paths for the `pcp` transport now support file expansion. Similarly the `cacert` for the `winrm` transport has been updated to support file expansion.
+
 ## 1.30.1
 
 #### Bug fixes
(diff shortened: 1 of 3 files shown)


nicklewis:
Should this be fixed in `orchestrator_client-ruby`?

donoghuc:
yeah, I guess it should just happen in orch client. 

I've gone through everywhere I can think of to make sure paths are expanded. 
```
puppetdb:
  server_urls: ["https://sv2e3sw3550ruy3.delivery.puppetlabs.net:8081"]
  cacert: ~/working_dir/bolt/Boltdir/ca.pem
  token: ~/working_dir/bolt/Boltdir/token
pcp:
  service-url: "https://sv2e3sw3550ruy3.delivery.puppetlabs.net:8143"
  cacert: ~/working_dir/bolt/Boltdir/ca.pem
  task-environment: production
  job-poll-timeout: 2
  token-file: ~/working_dir/bolt/Boltdir/token
plugins:
  pkcs7:
    private-key: ../Boltdir/keys/foo/priv
    public-key: ../Boltdir/keys/foo/pub
    keysize: 4096

hiera-config: ~/working_dir/bolt/spec/fixtures/apply/hiera_datadir.yaml
```
Plugins should be fine because they are relative to the `Boltdir`. I have verified for ssh private key. Still working on winrm cacert. I will close this when I put up fix in orch client.

lucywyman:
Closes #1174 

I think this also needs a CHANGELOG update? I think it's also testable, but if it turns out to be hard to test then manual testing seems fine.

donoghuc:
PR to fix in orchestrator_client-ruby https://github.com/puppetlabs/orchestrator_client-ruby/pull/25

PR #1235, opened 2019-09-20
(maint) Remove installation instruction for deprecated platform

donoghuc:
As of 1.30.1 bolt is no-longer built and shipped for ubuntu 1404. Remove platform from install page.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@
 
 ## 1.30.1
 
+#### Deprecations and removals
+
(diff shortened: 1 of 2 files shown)


donoghuc:
I know last release we make a deprecation warning that 1404 would be removed in next release. Should we have another entry in 1.30.1 that there is no 1404 packages?

lucywyman:
I think we should have another changelog entry that in 1.30.1 that packages are no longer available.

cwcowellshah:
I agree that we'd want a changelog entry explaining that Ubuntu 14.04 support is gone.

I'll add a line to the "Deprecations" section of the docs saying the same thing.

Otherwise, I'm good with merging.

PR #1236, opened 2019-09-20
(ORCH-2317) Update upload_file in bolt-server to work with single files

McdonaldSeanp:
Prior to this commit, the bolt-server implementation of upload_file would
always attempt to download with a source location of cache_dir/job_id (i.e. a
top-level directory). This breaks downloads for single files, since a single
file download request could have a destination with a fully qualified
filename, in which case the download would make the destination filename in to
a directory and download the single file in to that directory.

This commit updates upload_file in transport_app.rb to special case the single
file scenario, so that the upload source is `cache_dir/job_id/relative_path`
(i.e. a fully qualified path to the single file).

### DIFF
--- a/spec/bolt_server/app_integration_spec.rb
+++ b/spec/bolt_server/app_integration_spec.rb
@@ -133,6 +133,38 @@ describe "BoltServer::TransportApp", puppetserver: true do
           post('/ssh/run_command', JSON.generate(body), 'CONTENT_TYPE' => 'text/json')
         end
       end
+
+      it 'copies a single file' do
+        job_id = Time.now.usec
+        body = build_upload_request(job_id, conn_target('ssh', include_password: true))
+        body['files'] = body['files'].select { |file_entry| file_entry['relative_path'] == 'test-file.sh' }
+        body['destination'] = '/home/bolt/single_file_test.sh'
+        destination = body['destination']
+
+        begin
+          post(path, JSON.generate(body), 'CONTENT_TYPE' => 'text/json')
+          expect(last_response).to be_ok
+          expect(last_response.status).to eq(200)
+          result = JSON.parse(last_response.body)
+          expect(result).to include('status' => 'success')
+          expect(result).to include('action' => 'upload')
+          expect(result['result']['_output'])
+            .to match(%r{Uploaded .*cache/#{job_id}/test-file.sh' to 'localhost:#{destination}'})
+
+          # Inspect results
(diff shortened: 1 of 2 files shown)


spec/lib/bolt_spec/bolt_server.rb
```diff
@@ -127,6 +127,28 @@ def build_upload_request(job_id, target)
       }
     end
 
+    def build_single_file_upload_request(job_id, target)
```
smcclellan: Could we just delete all but one 'file' entry in `files` for the other request? That'd save some code duplication.

McdonaldSeanp: I'll take a look at removing the duplication

spec/bolt_server/app_integration_spec.rb
```diff
@@ -133,6 +133,37 @@ def app
...
+            .to match(%r{Uploaded .*cache/#{job_id}/test-file.sh' to 'localhost:#{destination}'})
+
+          # Inspect results
+          body = build_command_request("ls #{destination}/*",
+                                       conn_target('ssh', include_password: true))
+
+          post('/ssh/run_command', JSON.generate(body), 'CONTENT_TYPE' => 'text/json')
+          result = JSON.parse(last_response.body)
+          expect(result['result']['stdout']).to match(/test-file.sh/)
+          expect(result['result']['stdout']).to match(/sub-file.sh/)
```
smcclellan: I imagine this will fail?

McdonaldSeanp: Yeah, I just removed that.

spec/bolt_server/app_integration_spec.rb
```diff
@@ -133,6 +133,40 @@ def app
           post('/ssh/run_command', JSON.generate(body), 'CONTENT_TYPE' => 'text/json')
         end
       end
+
+      it 'copies a single file' do
+        job_id = Time.now.usec
+        body = build_upload_request(job_id, conn_target('ssh', include_password: true))
+        body['files'] = body['files'].select { |file_entry| file_entry['relative_path'] == 'test_file.sh' }
+        puts "DEBUG DEBUG DEBUG"
```
speedofdark: Once you pull the debug, LGTM.

speedofdark: (and fix Travis :) )

Issue #1207, opened 2019-09-08
Unable to call run_task() in noop mode from a plan


nmaludy:
## Describe the Bug
From a plan i'm trying to call a `run_task()` and have it execute in noop mode. No matter what value i pass to `run_task()` the task never seems to be receiving a value for the `_noop` parameter.

## Expected Behavior
When i pass in the `_noop` option into `run_task()` i would like the `_noop` parameter to be passed into the task.

## Steps to Reproduce
1. Create a task metadata file:
```json
{
  "description": "Tests noop",
  "supports_noop": true
}
```

2. Create a task shell script:
```sh
#!/bin/bash

echo "PT__noop = $PT__noop"
if [[ -n "$PT__noop" && "$PT__noop" == "true" ]]; then
  echo "noop mode"
else
  echo "regular mode"
fi
exit 0
```

3. Create a plan:
```puppet
plan patching::noop_test (
  TargetSpec $nodes,
) {
  return run_task('patching::noop_test', $nodes,
                  _noop => true)
}
```

4. Execute the plan and check the output: 
```sh
bolt plan run patching::noop_test --nodes xxx
.....
  PT__noop = true
  noop mode
  {
  }
```

## Environment

 - Version:
```sh
$ bolt --version
1.30.0
```

 - Platform
```sh
$ cat /etc/redhat-release 
CentOS Linux release 7.6.1810 (Core)
```

## Additional Context
I think this is caused by the following line of code: https://github.com/puppetlabs/bolt/blob/master/bolt-modules/boltlib/lib/puppet/functions/run_task.rb#L68

It seems that this line is splitting all of the options passed into `run_task()` that start with a `_` into another hash. The `_noop` parameter is never added back into the `use_args[]` hash and therefore lost when executing the task.

lucywyman:
Hi Nick!

I'm running into this too trying to use `_catch_errors`. I'll follow up with the team tomorrow - I'm not sure what the line is *intended* to do, but I don't think this behavior is desired.

donoghuc:
This showed up in the stream but just wanted to comment on it here. https://github.com/puppetlabs/bolt/pull/1237#issuecomment-539137427 There is a code path in the CLI where `noop` *would* be passed to plan run. This ticket should maybe include addressing if we want to support that or not.

PR #1237, opened 2019-09-20
(maint) Standardize params vs. options and use symbols

nicklewis:
Previously, we split out params and options into separate hashes
based on whether the keys started with _, but then left the keys that
way. That meant that the internals of Bolt were littered with code that
included leading underscores and thus had some evidence of where those
options originated from. As the leading underscore is an implementation
detail of the Puppet functions, we now consistently strip them out when
exiting the Puppet language and standardize them to symbols.

### DIFF
--- a/bolt-modules/boltlib/lib/puppet/functions/upload_file.rb
+++ b/bolt-modules/boltlib/lib/puppet/functions/upload_file.rb
@@ -47,18 +47,19 @@ Puppet::Functions.create_function(:upload_file, Puppet::Functions::InternalFunct
     return_type 'ResultSet'
   end
 
-  def upload_file(scope, source, destination, targets, options = nil)
+  def upload_file(scope, source, destination, targets, options = {})
     upload_file_with_description(scope, source, destination, targets, nil, options)
   end
 
-  def upload_file_with_description(scope, source, destination, targets, description = nil, options = nil)
+  def upload_file_with_description(scope, source, destination, targets, description = nil, options = {})
     unless Puppet[:tasks]
       raise Puppet::ParseErrorWithIssue
         .from_issue_and_stack(Bolt::PAL::Issues::PLAN_OPERATION_NOT_SUPPORTED_WHEN_COMPILING, action: 'upload_file')
     end
 
-    options ||= {}
-    options = options.merge('_description' => description) if description
+    options = options.select { |opt| opt.start_with?('_') }.map { |k, v| [k.sub(/^_/, '').to_sym, v] }.to_h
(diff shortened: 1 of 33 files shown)


spec/bolt/transport/local_spec.rb
```diff
@@ -131,7 +131,7 @@
     end
 
     context 'with slow input' do
-      let(:file_size) { 100 }
+      let(:file_size) { 10 }
```
nicklewis: This is an unrelated change but it cuts this test from 20 seconds to 2 seconds and I don't _think_ it has an impact on the behavior of the test. I couldn't quite tell what problem it was meant to spot though, so I can't be sure.

adreyer:
Note how plugins use options.

https://github.com/puppetlabs/bolt/blob/master/lib/bolt/plugin/module.rb#L167

bolt-modules/boltlib/lib/puppet/functions/upload_file.rb
```diff
@@ -47,18 +47,19 @@
...
-  def upload_file_with_description(scope, source, destination, targets, description = nil, options = nil)
+  def upload_file_with_description(scope, source, destination, targets, description = nil, options = {})
     unless Puppet[:tasks]
       raise Puppet::ParseErrorWithIssue
         .from_issue_and_stack(Bolt::PAL::Issues::PLAN_OPERATION_NOT_SUPPORTED_WHEN_COMPILING, action: 'upload_file')
     end
 
-    options ||= {}
-    options = options.merge('_description' => description) if description
+    options = options.select { |opt| opt.start_with?('_') }.map { |k, v| [k.gsub(/^_/, '').to_sym, v] }.to_h
```
donoghuc: Why do you use `gsub` here? Looks like for `run_plan` and this you use it, but for the others its just `sub`? Are there multiple underscores to replace for only these two?

spec/bolt/transport/local_spec.rb
donoghuc: This seems totally reasonable.

bolt-modules/boltlib/lib/puppet/functions/upload_file.rb
nicklewis: I think I was just inconsistent. I'll standardize them all the `sub`.

nicklewis:
Updated to use `sub` everywhere and rebased on master. I think the noop thing is sort of related to #1207 so might be better to address there. They're both about the fact that we have half-implemented noop support for plans.

PR #1238, opened 2019-09-20
(ORCH-2317) Update transport_app upload_file to use thread safe download

McdonaldSeanp:
The upload_file api in transport_app.rb (for bolt-server) uses
`@file_cache.download_file` to download files to a cachedir before actually
uploading them to hosts.

This is broken when more than one thread is in use, because download_file
should be wrapped in `@file_cache.serial_execute` so downloads only happen one
at a time (the download function itself is not thread safe).

This commit updates upload_file to use `serial_execute`


caseywilliams:
I'm not sure I see the link between this and the Travis failure

PR #1246, opened 2019-09-25
(ORCH-2339) Add `run_script` endpoint to bolt server

smcclellan:
In cases where PXP Agent is not installed on a target node and thus
cannot run a script, SSH or WinRM should be used. This commit adds the
`POST /run_script` endpoint which will delegate either to SSH or to
WinRM accordingly.

### DIFF
--- a/developer-docs/bolt-api-servers.md
+++ b/developer-docs/bolt-api-servers.md
@@ -248,6 +248,70 @@ For example, the following uploads file 'abc' on windows_target.net:
 }
 ```
 
+### POST /ssh/run_script
+- `target`: [SSH Target Object](#ssh-target-object), *required* - Target information to run script on.
+- `script`: Object, *required* - The script being executed.
+    - `filename`: String, *required* - The destination for the script on the target.
+    - `uri`: Object, *required* - The location of where to find the script.
+        - `path`: String, *required* - The endpoint to retrieve the script.
+        - `params`: Object, *required* - The parameters to supply the endpoint.
+    - `sha256`: String, *required* - The SHA256 value for the script.
+- `arguments`: String, *optional* - Which arguments to pass to the script.
+
+For example, the following runs script 'file.sh' on linux_target.net:
+```
+{
+  "target": {
+    "hostname": "linux_target.net",
+    "user": "marauder",
+    "password": "I solemnly swear that I am up to no good",
+    "host-key-check": false,
+    "run-as": "george_weasley"
+  },
+  "script" : {
+    "filename": "file.sh",
+    "uri": {
+      "path": "/puppet/v3/file_content/modules/some_module/file.sh",
+      "params": {}
+    },
+    "sha256": "SHA256VALUE"
+  },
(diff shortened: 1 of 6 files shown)


spec/bolt_server/app_integration_spec.rb
```diff
@@ -135,6 +135,38 @@ def app
       end
     end
 
+    describe "run_script" do
```
smcclellan: It seems a little odd to me that we're only testing these endpoints in `app_integration_spec.rb`, rather than `transport_app_spec.rb`, which seems like a natural place for testing changes to that file.

developer-docs/bolt-api-servers.md
```diff
@@ -248,6 +248,70 @@ For example, the following uploads file 'abc' on windows_target.net:
 }
 ```
 
+### POST /ssh/run_script
+- `target`: [SSH Target Object](#ssh-target-object), *required* - Target information to run task on.
```
speedofdark: run *script* on

developer-docs/bolt-api-servers.md
```diff
@@ -248,6 +248,70 @@ For example, the following uploads file 'abc' on windows_target.net:
...
+      "params": {}
+    },
+    "sha256": "SHA256VALUE"
+  },
+  "arguments": "--test",
+}
+```
+
+### POST /winrm/run_script
+- `target`: [WinRM Target Object](#winrm-target-object), *required* - Target information to run task on.
```
speedofdark: run *script* on

spec/bolt_server/app_integration_spec.rb
McdonaldSeanp: I _think_ that this is just a poorly named file. I'm pretty sure the use of the word "integration" is supposed to identify that this test actually sets up the entire service and runs it with some example nodes. `transport_app_spec.rb` seems to only have tests that mock any underlying calls out to nodes, whereas `app_integration_spec.rb` appears to be actually running things on example nodes.

McdonaldSeanp:
Travis errors appear to be rubocop, once those are fixed I am 👍 to merge

spec/bolt_server/app_integration_spec.rb
caseywilliams: When I initially refactored the transport app for SSH and WinRM to accept multiple actions (instead of just run-task), I did the minimum here to make sure we could add new actions quickly - I did not try to reorganize the tests. There just hasn't really been any effort in any direction - reoganization is welcome as far as I'm concerned. I don't think it makes sense right now either.

McdonaldSeanp: FWIW I think rename the file might make things make a little more sense, but I don't know if we have a ton of time to do much more than that right now with the timeline we are on.

smcclellan: I was mostly commenting here to make a note, no rush in making any immediate changes as far as I'm concerned.

Issue #1241, opened 2019-09-23
A plan apply() incorrectly returns successful if the report is unparseable.


jpartlow:
## Bug

If a Puppet provider puts() extraneous output during the apply, it can render the report unparseable (invalid JSON). However, Bolt's apply() function ends up returning ok() true, and with the report() pointing to the String output of the run rather than the actual parsed report.

## Expected

Bolt should fail the apply and let you know that the report was unparseable.

## Reproduction

The puppetlabs/puppet-enterprise-modules pe_postgresql modules has a buggy [pe_postgresql_psql provider](https://github.com/puppetlabs/puppet-enterprise-modules/blob/2019.2.0.0/modules/pe_postgresql/lib/puppet/provider/pe_postgresql_psql/ruby.rb#L22) that puts() status output when the unless parameter's execution fails.

You can reproduce this with this plan (but you'll need the puppet_enterprise and pe_postgresql modules from puppetlabs/puppet-enterprise-modules in Bolt's modulepath). Also bear in mind that we'll be patching pe_postgresql_psql to remove the puts(), so you'll want 2019.2.0.0 or less...or build/modify a provider with some puts() calls :)

```
plan enterprise_tasks::test(
  Variant[String,Target] $master,
) {
  $master_target = get_targets($master)[0]
  set_feature($master_target, 'puppet-agent', true)
  run_plan('facts', 'nodes' => $master_target)

  $result = apply($master_target, '_catch_errors' => true) {
    include puppet_enterprise
    puppet_enterprise::psql { 'with a broken unless':
      command  => 'select 1',
      db       => 'bar',
      unless   => 'select * from foo',
    }
  }.first()
  err($result.report())
  err($result.ok())
}
```

Output:

```
[root@wuh7cxdfusfnult ~]# /opt/puppetlabs/installer/bin/bolt --boltdir /opt/puppetlabs/installer/share/Boltdir/ plan run enterprise_tasks::test master=localhost
Starting: plan enterprise_tasks::test
Starting: plan facts
Starting: task facts on localhost
Finished: task facts with 0 failures in 1.96 sec
Finished: plan facts in 1.98 sec
Starting: apply catalog on localhost
{_output => pid 9468 exit 2
{
  "host": "wuh7cxdfusfnult.delivery.puppetlabs.net",
  "time": "2019-09-21T00:35:11.170783419+00:00",
  "configuration_version": 1569026107,
  "transaction_uuid": null,
  "report_format": 10,
[...]

PR #1248, opened 2019-09-25
(GH-1241) Handle unexpect results from the apply task

adreyer:
Previously ApplyResult would create difficult to use objects when it got
unexpected reports back from the apply task. It will now treat those
results as errors. This may result in an apply erroring when a
misbehaved provider prints to stdout even when the application otherwise
appears successful.

@jpartlow do you want to test this with your use case.

### DIFF
--- a/lib/bolt/apply_result.rb
+++ b/lib/bolt/apply_result.rb
@@ -46,6 +46,25 @@ module Bolt
       end
     end
 
+    def self.invalid_report_error(result)
+      # These are the keys ApplyResult methods rely on.
+      expected_report_keys = %w[metrics resource_statuses status]
+      missing_keys = expected_report_keys.reject { |k| result.value.include?(k) }
+
+      unless missing_keys.empty?
+        if result['_output']
+          # rubocop:disable Metrics/LineLength
+          msg = "Report result contains an '_output' key. Catalog application may have printed extraneous output to stdout: #{result['_output']}"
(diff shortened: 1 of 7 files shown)


adreyer:
This still needs a changelog entry but I figured I'd wait for the release.

lib/bolt/apply_result.rb
```diff
@@ -46,13 +46,36 @@ def self.resource_error(result)
...
+      expected_report_keys = %w[metrics resource_statuses status]
+      missing_keys = expected_report_keys.reject { |k| result.value.include?(k) }
+
+      unless missing_keys.empty?
+        if result['_output']
+          # rubocop:disable Metrics/LineLength
+          msg = "Report result contains an '_output' key. Catalog application may have printed extraneous output to stdout: #{result['output']}"
+          # rubocop:enable Metrics/LineLength
+        else
+          msg = "Report did not contain all expected keys missing: #{missing_keys.join(' ,')}"
```
nicklewis:
```suggestion
          msg = "Report did not contain all expected keys, missing: #{missing_keys.join(' ,')}"
```

lucywyman:
Closes #1241

donoghuc:
Remember to add a CHANGELOG commit.

jpartlow:
@adreyer I think I can test this today, thanks.

PR #1249, opened 2019-09-25
(maint) Properly set the `wait` argument on PuppetDB commands

nicklewis:
Previously, this code was using an invalid hash lookup instead of a hash
assignment. However, it was also trying to set this parameter in the
POST body instead of the URL where it belongs. We now correctly set it
as a query param.

### DIFF
--- a/spec/lib/bolt_spec/puppetdb.rb
+++ b/spec/lib/bolt_spec/puppetdb.rb
@@ -20,9 +20,9 @@ module BoltSpec
     def make_command(command:, version:, payload:, wait: nil)
       client = pdb_client
       url = "#{client.uri}/pdb/cmd/v1"
+      url += "?secondsToWaitForCompletion=#{wait}" if wait
 
       body = { command: command, version: version, payload: payload }
-      body['secondsToWaitForCompletion' => wait] if wait
       body = JSON.generate(body)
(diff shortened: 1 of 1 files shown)


spec/lib/bolt_spec/puppetdb.rb
```diff
@@ -22,7 +22,7 @@ def make_command(command:, version:, payload:, wait: nil)
       url = "#{client.uri}/pdb/cmd/v1"
 
       body = { command: command, version: version, payload: payload }
-      body['secondsToWaitForCompletion' => wait] if wait
```
donoghuc: Unfortunately I think it only works in the url 
```
    def make_command(command:, version:, payload:, wait: nil)
      client = pdb_client
      url = if wait 
              "#{client.uri}/pdb/cmd/v1"
            else
              "#{client.uri}/pdb/cmd/v1?secondsToWaitForCompletion=#{wait}"
            end

      body = { command: command, version: version, payload: payload }
      body = JSON.generate(body)

      headers = { "Content-Type" => "application/json" }

      response = client.http_client.post(url, body: body, header: headers)
      response
    end
```

nicklewis:
Updated to set it as a query param

PR #1253, opened 2019-09-26
(maint) Update pdb container setup for new ssl setup

donoghuc:
The ssl-setup.sh script is no longer used by the puppetdb container. This commit removes the curtom entrypoint for setting up docker-compose stack for testing to use the entrypoint defined by the base image. It also updates the destination for the ssl setup to the expected location.

### DIFF
--- a/spec/Dockerfile.puppetdb
+++ b/spec/Dockerfile.puppetdb
@@ -1,13 +1,7 @@
 FROM puppet/puppetdb
 
-ARG hostname="bolt-puppetdb"
-
 # Use our own certs so this doesn't have to wait for puppetserver startup
-COPY fixtures/ssl/ca.pem /etc/puppetlabs/puppet/ssl/certs/ca.pem
-COPY fixtures/ssl/cert.pem /etc/puppetlabs/puppet/ssl/certs/pdb.pem
-COPY fixtures/ssl/key.pem /etc/puppetlabs/puppet/ssl/private_keys/pdb.pem
-COPY fixtures/ssl/crl.pem /etc/puppetlabs/puppet/ssl/ca/ca_crl.pem
-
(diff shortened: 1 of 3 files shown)


donoghuc:
lol, i think we need ssl-setup.sh https://github.com/puppetlabs/puppetdb/commit/fbb997751fc33281c0ac8b8a642410eba03c5cf3#diff-11e47e7f4c33c3faf0670e2268088a2d

spec/Dockerfile.puppetdb
```diff
@@ -1,13 +1,9 @@
-FROM puppet/puppetdb:6.5.0
+FROM puppet/puppetdb
 
 ARG hostname="bolt-puppetdb"
```
Iristyle: This can be supplied in `docker-compose.yml`

spec/fixtures/puppetdb/docker-entrypoint.sh
```diff
@@ -1,8 +1,8 @@
 #!/bin/sh
```
Iristyle: This entire file can be trashed

donoghuc: :fire:

spec/Dockerfile.puppetdb
donoghuc: Oh, added to docker-compose.yaml. need to delete i here.

Iristyle:
This was needed due to changes from https://github.com/puppetlabs/puppetdb/pull/3082 / https://github.com/puppetlabs/puppetdb/pull/3086.

https://github.com/puppetlabs/puppetdb/pull/3087 had to merge / ship with some minor improvements to make this PR possible.

PR #1255, opened 2019-09-26
(maint) Add github action to publish preview docs on PR

lucywyman:
This adds a github action which will build and update the bolt docs
preview site when a PR is opened, synchronized, or reopened.


MikaelSmith:
Where will the preview site be visible?

lucywyman:
I *think* the same place they were viewable before? https://docs-preview.webteam.puppet.com/docs/bolt/dev/  but...now that I say that I think that site is password protected, which won't work for external folks.

nicklewis:
How will that work if we have multiple PRs open at the same time?

lucywyman:
@nicklewis I'll copy-paste Heston's answer directly: "I think that would be fairly difficult to achieve ATM because the web team has to set up each new directory in the s3 to point to something in Drupal - it’s a fairly manual process. We’re hoping that once our site has migrated off of Drupal (which is happening soonish) we’ll be able to build the site locally - that will probably mean more versatility. For now I think all PRs will have to push to the same staging environment on Drupal"

adreyer:
Can we clone all of this from somewhere else instead of committing 2k+ files?

.github/workflows/stage.yaml
```diff
@@ -0,0 +1,25 @@
...
+    - name: Build HTML from source
+      run: .github/build_s3.sh
+    - name: Install s3-cli
+      uses: actions/aws/cli@master  
+    - name: Upload docs to the bucket
+      run: |
+        aws s3 sync html s3://puppet-docs-website-transfer/bolt-dev --delete
+        aws s3 ls --recursive s3://puppet-docs-website-transfer/bolt-dev
+      env:
+       AWS_ACCESS_KEY_ID: ${{ secrets.S3_ACCESS_KEY_ID }}
```
adreyer: I don't know enough about GHA but are these really secret? What prevents someone from creating a PR that adds a step that sends these secrets somewhere?

I assume these creds are locked down enough that they only allow someone to push to an s3 bucket we don't really care about. However, IMO if they're not secret we shouldn't encrypt them someone may see that and put an important secret here too.

MikaelSmith: The workflow that's executed will be whatever's on the primary (master) branch, not the one in the PR. The only way to test workflows is to run them in your own fork (which won't have that secret) before PRing back.

MikaelSmith: Now if you have a workflow that executes code from the PR - like building and running that code - then could it read the secret and print it? Maybe. Although I *think* GitHub actions actually filters for the secret value and covers it in asterisks.

PR #1259, opened 2019-09-26
(maint) Add github action to publish preview docs on PR

lucywyman:
This adds a github action which will build and update the bolt docs
preview site when a PR is opened, synchronized, or reopened.


lucywyman:
Woo! Got this working up to where I need the AWS credentials (I don't have them locally, so can't use them on my fork) https://github.com/lucywyman/bolt/commit/f88a73f68512a9e26849809ee06ff54fe98f651e/checks . 

I'm still concerned about the security issue Alex brought up. The github action that gets run seems to be the one in the PR, not on master (similar to how .travis.yaml works), which seems dangerous, and I'm not sure what that allows users to do with our secrets. I'll investigate more later today.

PR #1260, opened 2019-09-26
(maint) Eagerly evaluate OS in analytics client

nicklewis:
Previously, we spun evaluation of the client OS into another thread to
avoid slowing down the main execution thread by needing to require
facter. This caused a race condition in facter between the thread
compting the OS and the main thread loading PAL which also needs to
evaluate some facts. As it turns out, by the time the analytics client
is instantiated we've already required facter anyway, so we don't have
to pay that cost. Evaluating the OS fact is negligible (~25ms), so we
now do it in the main thread to avoid the race condition.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,11 @@
+## 1.31.1
+
+#### Bug fixes
+
+* **Spurious plan failures and warnings on startup**
+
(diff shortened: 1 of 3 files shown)


donoghuc:
Can you add a CHANGELOG commit?

PR #1261, opened 2019-09-27
(packaging) Ship azure_inventory plugin and update bundled modules

donoghuc:
This commit pulls in the azure_inventory module. In addition other bundled modules are updated to the latest version.

### DIFF
--- a/Puppetfile
+++ b/Puppetfile
@@ -12,7 +12,7 @@ mod 'puppetlabs-puppet_agent', '2.2.0'
 # Core types and providers for Puppet 6
 mod 'puppetlabs-augeas_core', '1.0.4'
 mod 'puppetlabs-host_core', '1.0.2'
-mod 'puppetlabs-scheduled_task', '1.0.1'
+mod 'puppetlabs-scheduled_task', '2.0.0'
 mod 'puppetlabs-sshkeys_core', '1.0.2'
 mod 'puppetlabs-zfs_core', '1.0.2'
 mod 'puppetlabs-cron_core', '1.0.2'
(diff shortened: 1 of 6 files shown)


Puppetfile
```diff
@@ -23,11 +23,14 @@ mod 'puppetlabs-zone_core', '1.0.2'
...
 # Useful additional modules
 mod 'puppetlabs-package', '0.6.0'
-mod 'puppetlabs-puppet_conf', '0.3.1'
+mod 'puppetlabs-puppet_conf', '0.4.0'
 mod 'puppetlabs-python_task_helper', '0.3.0'
 mod 'puppetlabs-reboot', '2.2.0'
 mod 'puppetlabs-ruby_task_helper', '0.4.0'
 
+# Plugin modules
+#mod 'puppetlabs-azure_inventory', '0.2.0'
```
donoghuc: @nicklewis We need to figure out why the push to forge job is failing. https://jenkins-master-prod-1.delivery.puppetlabs.net/job/forge-module_puppetlabs-azure_inventory_init-tag_push-forge/3/console

donoghuc:
Another packaging PR https://github.com/puppetlabs/puppet-runtime/pull/229

Puppetfile
donoghuc: Alex pointed out that it is not that important to publish this module to the forge. Pulling from git tag for now.

lib/bolt/cli.rb
```diff
@@ -544,7 +544,7 @@ def bundled_content
       # We only need to enumerate bundled content when running a task or plan
       content = { 'Plan' => [],
                   'Task' => [],
-                  'Plugin' => %w[puppetdb pkcs7 prompt terraform task] }
+                  'Plugin' => Bolt::Plugin::BUILTIN_PLUGINS }
```
donoghuc: Do we also want to append the 'azure_inventory` module here?

donoghuc: Yes, eventually we will want to move most of these to modules (notably *not* pdb or prompt).

Puppetfile
```diff
@@ -23,11 +23,16 @@ mod 'puppetlabs-zone_core', '1.0.2'
...
 # Useful additional modules
 mod 'puppetlabs-package', '0.6.0'
-mod 'puppetlabs-puppet_conf', '0.3.1'
+mod 'puppetlabs-puppet_conf', '0.4.0'
 mod 'puppetlabs-python_task_helper', '0.3.0'
 mod 'puppetlabs-reboot', '2.2.0'
 mod 'puppetlabs-ruby_task_helper', '0.4.0'
 
+# Plugin modules
+mod 'puppetlabs-azure_inventory',
```
nicklewis: I pushed 0.2.0 up manually, so we should be able to pull it from the forge now.

donoghuc: nice, Updated.

PR #1263, opened 2019-09-28
Friendly deprecation error message

kreeuwijk:
Changing the `encrypted-value` deprecation message to something that's more descriptive and user-friendly.

### DIFF
--- a/lib/bolt/secret/base.rb
+++ b/lib/bolt/secret/base.rb
@@ -36,7 +36,9 @@ module Bolt
       def validate_resolve_reference(opts)
         # TODO: Remove deprecation warning
         if opts.include?('encrypted-value')
-          raise Bolt::ValidationError, "Parsing error: The 'encrypted-value' key is deprecated and can no longer be used. In your Bolt config files, please use 'encrypted_value' instead."
+          msg = "Parsing error: The 'encrypted-value' key is deprecated and can no longer be used. " \
+                "In your Bolt config files, please use 'encrypted_value' instead."
+          raise Bolt::ValidationError, msg
         end
         decode(opts['encrypted_value'])
       end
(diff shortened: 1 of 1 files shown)


lib/bolt/secret/base.rb
```diff
@@ -36,7 +36,7 @@ def secret_decrypt(opts)
       def validate_resolve_reference(opts)
         # TODO: Remove deprecation warning
         if opts.include?('encrypted-value')
-          raise Bolt::ValidationError, "The 'encrypted-value' key is deprecated migrate to to 'encrypted_value'"
+          raise Bolt::ValidationError, "Inventory file parsing error: The 'encrypted-value' key in the inventory file is deprecated and can no longer be used. Please change the name of the key to 'encrypted_value' instead, and all will be fine."
```
donoghuc: It is not really an inventory error. Would `Plugin error` help make this easier to understand?

kreeuwijk: From the user's perspective, it's not clear what a `Plugin error` is....
The problem is that the inventory file contains a key that's deprecated and needs to be changed. The error message should reflect that, if it is to be of help to the user.

adreyer: these values will be used from bolt.yaml too so it's not just inventory.

kreeuwijk: Ah I see, ok that warrants some changes

lib/bolt/secret/base.rb
```diff
@@ -36,7 +36,7 @@ def secret_decrypt(opts)
       def validate_resolve_reference(opts)
         # TODO: Remove deprecation warning
         if opts.include?('encrypted-value')
-          raise Bolt::ValidationError, "The 'encrypted-value' key is deprecated migrate to to 'encrypted_value'"
+          raise Bolt::ValidationError, "Parsing error: The 'encrypted-value' key is deprecated and can no longer be used. In your Bolt config files, please use 'encrypted_value' instead."
```
nicklewis: To satisfy rubocop:

```suggestion
          msg = "Parsing error: The 'encrypted-value' key is deprecated and can no longer be used. " \
                "In your Bolt config files, please use 'encrypted_value' instead."
          raise Bolt::ValidationError, msg
```

PR #1266, opened 2019-09-30
(maint) Properly raise FileError when hiera-config doesn't exist

nicklewis:


lib/bolt/config.rb
```diff
@@ -277,7 +277,7 @@ def validate
       end
 
       if @hiera_config && !(File.file?(@hiera_config) && File.readable?(@hiera_config))
-        raise Bolt::FileError, "Could not read hiera-config file #{@hiera_config}", @hiera_config
+        raise Bolt::FileError.new("Could not read hiera-config file #{@hiera_config}", @hiera_config)
```
donoghuc: Can we just use `Bolt::Util#validate_file` for this? https://github.com/puppetlabs/bolt/blob/2293724e1e8226682ad07b4de05108e365db2c2f/lib/bolt/util.rb#L193

nicklewis: Good call, updated.

Issue #1140, opened 2019-08-09
get_resources does not work with resource objects.


donoghuc:
## Describe the Bug
We document `get_resources(localhost, Package)` as an example of how to use `get_resources` and the function itself is defined to accept `Resource` objects in addition to strings. However `Resource Types` are not loaded during a plan so unless you quote the resource name and pass it as a string the plan will error.

`Evaluation Error: Resource type not found: Package[...]`

## Expected Behavior
Users should be able to use core types like Package and File or any custom types available on the module path with get_resources which may change how these objects behave more generally in the plan language.## Steps to Reproduce

donoghuc:
Solved with https://github.com/puppetlabs/bolt/pull/1102

PR #1268, opened 2019-10-01
(GH-1140) Maintain ruby 2.3 compatibility by not using Dir.children

donoghuc:
Ruby 2.3 does not support the `children` method for the `Dir` class. This commit updates the code to explicitly filter the `.` and `..` directories when loading resource types using the `Dir#entries` method.

### DIFF
--- a/lib/bolt/pal.rb
+++ b/lib/bolt/pal.rb
@@ -117,7 +117,8 @@ module Bolt
       static_loader = loaders.static_loader
       static_loader.runtime_3_init
       if File.directory?(@resource_types)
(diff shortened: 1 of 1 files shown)


donoghuc:
https://github.com/puppetlabs/bolt/pull/1102#discussion_r328311340

lucywyman:
Worth noting this is *slightly* different from `.childeren` in that it won't match dotfiles (ie. `.foo.pp`), but I assume we don't care about that for loading resource types. Otherwise works locally for me!

Issue #1243, opened 2019-09-24
Use Target `safe_name` when referring to a target


donoghuc:
## Describe the Solution You Would Like
When a plan report is printed it uses the `safe_name` everywhere it encounters a `Target` object. In addition any place a `Target` needs to be referenced, the safe_name should be displayed. The `to_s` method for a Target should return the `safe_name`. This could include implementing a `safe_name` method for Target objects. That way we can refactor every place that calls `target.host` to reference a target with `to_s`.

donoghuc:
resolved with https://github.com/puppetlabs/bolt/pull/1271

PR #1271, opened 2019-10-02
(GH-1243) Refer to Targets by their safe name

donoghuc:
This commit updates all logging and outputters to refer to targets by their `safe_name`. For an `Target2` object this is the uri minus the password. For a `Target` object it is the host.

### DIFF
--- a/lib/bolt/applicator.rb
+++ b/lib/bolt/applicator.rb
@@ -81,7 +81,7 @@ module Bolt
     end
 
     def compile(target, ast, plan_vars)
-      trusted = Puppet::Context::TrustedInformation.new('local', target.host, {})
+      trusted = Puppet::Context::TrustedInformation.new('local', target.name, {})
       facts = @inventory.facts(target).merge('bolt' => true)
 
       catalog_input = {
@@ -90,7 +90,7 @@ module Bolt
         pdb_config: @pdb_client.config.to_hash,
         hiera_config: @hiera_config,
         target: {
-          name: target.host,
+          name: target.name,
           facts: facts,
           variables: @inventory.vars(target).merge(plan_vars),
           trusted: trusted.to_h
--- a/lib/bolt/outputter/human.rb
+++ b/lib/bolt/outputter/human.rb
@@ -73,14 +73,14 @@ module Bolt
       end
 
       def print_start(target)
-        @stream.puts(colorize(:green, "Started on #{target.host}..."))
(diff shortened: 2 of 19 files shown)


donoghuc:
https://github.com/puppetlabs/bolt/issues/1243

lib/bolt/outputter/human.rb
```diff
@@ -312,7 +312,7 @@ def print_module_list(module_list)
       end
 
       def print_targets(options)
-        targets = options[:targets].map(&:name)
+        targets = options[:targets].map(&:safe_name)
```
donoghuc: I'm assuming this change applies to  `bolt inventory show`? It makes more sense for `Target2`. Let me know if we want this or not. I'm leaning toward not doing it because it could.

lib/bolt/applicator.rb
```diff
@@ -90,7 +90,7 @@ def compile(target, ast, plan_vars)
         pdb_config: @pdb_client.config.to_hash,
         hiera_config: @hiera_config,
         target: {
-          name: target.host,
+          name: target.safe_name,
```
adreyer: Since this and the trusted info is actually used in code I think it needs to be the real name.

lib/bolt/outputter/human.rb
adreyer: I think this should be the "string" representation of the target which I believe is the `name`

lib/bolt/outputter/json.rb
```diff
@@ -90,7 +90,7 @@ def print_puppetfile_result(success, puppetfile, moduledir)
       end
 
       def print_targets(options)
-        targets = options[:targets].map(&:name)
+        targets = options[:targets].map(&:safe_name)
```
adreyer: This definitely feels wrong since we otherwise stringify the target as the name.

lib/bolt/outputter/human.rb
adreyer: I feel more strongly about the json output than this.

adreyer:
There are a bunch of messages the the transports that use target.uri. I think we should switch most of those to target.safe_name. Even for the debug messages that seems good enough for now.

donoghuc:
Cool. I will look in to that.

adreyer:
one more uri call https://github.com/puppetlabs/bolt/blob/master/lib/bolt/executor.rb#L221

PR #1277, opened 2019-10-03
(docs) Regenerate plan reference with unformatted function names

donoghuc:
This commit updates the erb template for the rake docs task to format the plan runction names with unformatted names.


cwcowellshah:
LGTM

Issue #1152, opened 2019-08-15
Bolt missing error message context on error in module manifest


zonk1:
## Describe the Bug
I've made an error in json manifest for task with implementations. My requirements stanza contained string instead of an array. Bolt issued an error message which doesn't help too much to find an error.

```
$ bolt task run --nodes mediapi upgrade::upgrade
Started on mediapi...
Failed on mediapi:
  value must be enumerable
Failed on 1 node: mediapi
Ran on 1 node in 0.11 seconds
```
Running with `--debug` doesn't help.

## Expected Behavior
Error messages for errors in config files should point to the place where the error is to allow to locate the problem even in larger code base or when testing bigger change. I would expect ideally filename and line.

```
Failed on mediapi:
      site-modules/upgrade/tasks/upgrade.json:4: "requirements": "raspbian": value must be enumerable
```

Or even `"requirements" should be an array`. But even showing just a field name and value as a context with the exception message could be a huge time saver.

## Steps to Reproduce
Create a module with manifest like below, where `features` is mistakenly set to just string instead of an array. 
```
$ cat site-modules/upgrade/tasks/upgrade.json
{
        "implementations": [
                {"name": "upgrade_debian.sh", "requirements": ["debian"]},
                {"name": "upgrade_raspbian.sh", "requirements": "raspbian"}
        ]
}
```

Run `bolt task run --nodes mediapi upgrade::upgrade`

Attached is a minimal set of config files that displays this behavior.
[bolt-error-message.tar.gz](https://github.com/puppetlabs/bolt/files/3506793/bolt-error-message.tar.gz)

## Environment
 - Version 1.28.0
 - Platform Debian 10

donoghuc:
The underlying error handing has been fixed in puppet when loading task metadata. https://tickets.puppetlabs.com/browse/PUP-10095 tracks that work. It should be released in the next 6.4.z series as well as the next 6.x release.

PR #1278, opened 2019-10-03
(GH-1152) Improve error messages when task metadata cannot be loaded

donoghuc:
Raise more informative error message to replace the cryptic message that was surfaced when `requirements` field in a task's `implementations` metadata was not an array.

Closes https://github.com/puppetlabs/bolt/issues/1152

### DIFF
--- a/lib/bolt/task.rb
+++ b/lib/bolt/task.rb
@@ -77,6 +77,10 @@ module Bolt
                impl = impls.find do |imp|
                  remote_impl = imp['remote']
                  remote_impl = metadata['remote'] if remote_impl.nil?
+                 unless imp['requirements'].is_a?(Array) || imp['requirements'].nil?
+                   msg = "The task metadata 'requirements' key expects an Array, got #{imp['requirements'].class}"
+                   raise ValidationError, msg
+                 end
                  Set.new(imp['requirements']).subset?(available_features) && !!remote_impl == @remote
                end
                raise NoImplementationError.new(target, self) unless impl
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,12 +12,16 @@
 
 * **`apply_prep` error when using Inventory Version 2** ([#1303](https://github.com/puppetlabs/bolt/pull/1303))
 
-  When the `plugin_hooks` key was not set for a target/group in Inventory v2 the `apply_prep` function would not work. Bolt now uses the default `plugin_hooks` and honors the `plugin_hooks` from bolt config when using Inventory v2. 
+  When the `plugin_hooks` key was not set for a target/group in Inventory v2 the `apply_prep` function would not work. Bolt now uses the default `plugin_hooks` and honors the `plugin_hooks` from bolt config when using Inventory v2.
 
 * **Better error output when parsing malformed `yaml` files** ([#1296](https://github.com/puppetlabs/bolt/issues/1296))
 
   Previously when parsing `yaml` config files a generic error message was surfaced with no information about where in the file the problem occurred. Now an error message that contains the path to the file as well as the line and column in the file where the error originated from.
 
+* **More detailed error message when task metadata has wrong form** ([#1152](https://github.com/puppetlabs/bolt/issues/1152))
+
+  Raise more informative error message to replace the cryptic message that was surfaced when `requirements` field in a task's `implementations` metadata was not an array.
+  
 ## 1.33.0
 
 ### Bug fixes
--- a/spec/bolt/task_spec.rb
+++ b/spec/bolt/task_spec.rb
@@ -84,5 +84,16 @@ describe Bolt::Task do
         expect { task.select_implementation(target) }.to raise_error('No suitable implementation of foo for example')
       }
     end
+
+    context 'requirements is not an array' do
+      let(:implementations) {
+        [{ 'name' => 'foo.rb', 'requirements' => 'foo' }]
+      }
+
+      it {
+        expect { task.select_implementation(target) }
+          .to raise_error(Bolt::ValidationError, "The task metadata 'requirements' key expects an Array, got String")
+      }
+    end
   end
 end


bolt-modules/boltlib/lib/puppet/functions/run_task.rb
```diff
@@ -75,7 +75,13 @@ def run_task_with_description(task_name, targets, description, task_args = nil)
       task = Bolt::Task.new(name: task_name, files: [{ 'name' => '', 'path' => '' }])
     else
       # TODO: use the compiler injection once PUP-8237 lands
-      task_signature = Puppet::Pal::ScriptCompiler.new(closure_scope.compiler).task_signature(task_name)
+      # Avaliable with Puppet.lookup https://github.com/puppetlabs/puppet/pull/6440/files
+      script_compiler = Puppet.lookup(:pal_script_compiler)
```
donoghuc: According to https://github.com/puppetlabs/puppet/pull/6440/files this should be available. This was the PR referenced in the `TODO`. However I was not able to get it to work with something like:
```
              Puppet::Functions.create_function(:'a::myscriptcompilerfunc', Puppet::Functions::InternalFunction) do
                dispatch :myscriptcompilerfunc do
                  script_compiler_param
                  param 'String',:name
                end
                def myscriptcompilerfunc(script_compiler, name)
                  script_compiler.is_a?(Puppet::Pal::ScriptCompiler) ? name : 'no go'
                end
              end
```
Is this something we want to do? It would require some alternate mocking in the spec tests where we use the lookup.

bolt-modules/boltlib/lib/puppet/functions/get_resources.rb
```diff
@@ -23,11 +23,15 @@
...
 
   def script_compiler
-    @script_compiler ||= Puppet::Pal::ScriptCompiler.new(closure_scope.compiler)
+    @script_compiler ||= Puppet.lookup(:pal_script_compiler)
   end
 
   def run_task(executor, targets, name, args = {})
-    tasksig = script_compiler.task_signature(name)
+    begin
+      tasksig = script_compiler.task_signature(name)
```
donoghuc: Should we put a method somewhere that accepts a compiler and a task name and then does this? It feels weird doing this same begin/rescue everywhere, but on the other hand where would the method live that makes sense all the places this is used?

donoghuc:
I would like some early review if this is the direction we want to go with this before I spend any more time on writing tests/stubbing out the boltlib spec tests.

bolt-modules/boltlib/lib/puppet/functions/get_resources.rb
lucywyman: I think having a wrapper in the Bolt::PAL class to do this makes sense. Something like:
```
def self.task_signature(name)
  script_compiler = Puppet.lookup(:pal_script_compiler)
  begin
    tasksig = script_compiler.task_signature(name)
  rescue
    ...
  end
  ...
```

I haven't looked at the other functions but it seems like this is the only place we use the script_compiler, so we could remove that entirely from at least this function.

lucywyman: I'm also not sure how looking up the script compiler is different from using the compiler configured for Bolt?
https://github.com/puppetlabs/bolt/blob/master/lib/bolt/pal.rb#L136-L137

donoghuc: So if task_signature (with is already an instance method in pal) were to be a class method, boltlib functions would require bolt/pal? Is that right?

Gemfile
```diff
@@ -11,6 +11,7 @@ gemspec
 gem "hocon", '>= 1.2.5'
 gem "json-schema", '>= 2.8.0'
 gem "puma", '>= 3.12.0'
+gem "puppet", git: "https://github.com/donoghuc/puppet.git", branch: 'GH-1152'
```
donoghuc: Instead of trying to catch and munge erros from loading task metadata, some additional error handling was implemented upstream. https://github.com/puppetlabs/puppet/pull/7769

donoghuc: https://tickets.puppetlabs.com/browse/PUP-10095

bolt-modules/boltlib/lib/puppet/functions/run_task.rb
donoghuc: I'm gong to wait on this. If the parameter was injected and we did not need to lookup I would do it, but it looks like maybe it is only injected for internal puppet functions?

bolt-modules/boltlib/lib/puppet/functions/get_resources.rb
donoghuc: Should be unnecessary given upstream validation.

lucywyman:
Closes #1152

Gemfile
lucywyman: Heh, should've merged that right when you put it up and I think it could've made it into the release tomorrow :) As it is, I think we can get the gem from the puppetlabs master branch until it's released (which I don't think is even scheduled, since 6.4.4 hasn't gone out yet)

donoghuc: The nice thing about this is that nothing is *dependent* on the change. Whenever we pick up a version of puppet with the improved messaging it will "just work". So all I need to do is delete this line and we are good. ( i guess I need a changelog too...)

donoghuc:
It was passing tests, just needed to rebase for conflict in changelog.

lib/bolt/task.rb
```diff
@@ -77,6 +77,10 @@ def select_implementation(target, provided_features = [])
                impl = impls.find do |imp|
                  remote_impl = imp['remote']
                  remote_impl = metadata['remote'] if remote_impl.nil?
+                 unless imp['requirements'].is_a?(Array) || imp['requirements'].nil?
```
lucywyman: Should this check also be done in Puppet? It seems like doing validation in one place would make this less confusing. This check seems like it would fit in the `find_implementations` method https://github.com/puppetlabs/puppet/blob/master/lib/puppet/module/task.rb#L133

donoghuc: I had not thought of that. I guess that makes sense. If we do it in puppet I dont think we should do it in bolt (at least not have the tests set up the way they are because they will break as soon as we pull in a puppet version with the update). I'm not opposed to it.

lucywyman: Yeah that sentence was ambiguous - 'also' = 'in addition to the other checks that are part of this issue', so we would only have the validation in Puppet. I think using the same pattern for both the `files` key check and the `requirements` makes it easier to find and understand how we validate metadata.

donoghuc: Cool. Will raise new PR in puppet and close this.

donoghuc:
Will likely close in favor of https://github.com/puppetlabs/puppet/pull/7783

Issue #1162, opened 2019-08-20
Standardize specifying file paths in bolt config


donoghuc:
## Use Case
Currently, most config settings are relative to the directory bolt is invoked from (inventoryfile, private-key, etc) while some paths such as modulepath or keys for the pkcs7 plugin are relative to the boltdir. For consistency, relative paths should be standardized.

## Describe the Solution You Would Like
 When configuration settings point to a file path it should be assumed to be relative to the bolt project directory (Boltdir) unless an absolute path is specified. Do not support the behavior where the user expect the path to be relative to the directory where bolt was invoked. 

Make sure to consider when `future` is available when implementing this.

PR #1280, opened 2019-10-03
Fix bug in pkcs7 plugin resolving home directories

nmaludy:
Previously the `pkcs7` plugin was not able to resolve a path like so:
```yaml
plugins:
  pkcs7:
    public-key: ~/.puppetlabs/bolt/keys/public_key.pkcs7.pem
    private-key: ~/.puppetlabs/bolt/keys/private_key.pkcs7.pem
```

When running it would produce an error:

```
Error executing plugin pkcs7 from resolve_reference in dev: No such file or directory @ rb_sysopen - /home/nmaludy/src/puppet-xxx/~/.puppetlabs/bolt/keys/private_key.pkcs7.pem
```

That's definitely not what we want.

Looking at the way we resolve paths in `config.rb` shows that we were using `File.expand_path()` where in `pkcs7.rb` we were using `File.absolute_path()`.

The fix was to simply make it `File.expand_path()` lookups and everything seems happy.

**NEED HELP  HERE**
On the testing side... i'm not sure how to create a spec test that can utilize a home directory to store these temporary keys for testing?

Any other thoughts here on the best way to test this?

### DIFF
--- a/lib/bolt/plugin/pkcs7.rb
+++ b/lib/bolt/plugin/pkcs7.rb
@@ -39,7 +39,7 @@ module Bolt
 
       def private_key_path
         path = @options['private-key'] || 'keys/private_key.pkcs7.pem'
-        path = File.absolute_path(path, boltdir)
+        path = File.expand_path(path, boltdir)
         @logger.debug("Using private-key: #{path}")
         path
       end
@@ -50,7 +50,7 @@ module Bolt
 
       def public_key_path
         path = @options['public-key'] || 'keys/public_key.pkcs7.pem'
-        path = File.absolute_path(path, boltdir)
+        path = File.expand_path(path, boltdir)
         @logger.debug("Using public-key: #{path}")
         path
       end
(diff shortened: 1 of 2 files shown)


donoghuc:
@nmaludy We would like to standardize these paths to be relative to the boltdir. https://github.com/puppetlabs/bolt/issues/1162 .

nicklewis:
I don't think it's super important that tests try to actually read the files. It's probably sufficient to just test the `private_key_path` and `public_key_path` methods and ensure that the paths are what is expected.

nmaludy:
@donoghuc i don't mind that, but in this case you're basically specifying an absolute path `~/.puppetlabs/bolt/keys/private_key.pkcs7.pem` but bolt isn't resolving it properly.

nicklewis:
I think this change is sensible and doesn't conflict with #1162.

nmaludy:
if a relative path is specified, i think it will still resolve to being relative to the boltdir (all the tests passed on my box)

donoghuc:
Oh, reading the full issue is important. I missed the `unless absolute path is specified`. I just went through the other week and made sure we were expanding paths (https://github.com/puppetlabs/bolt/pull/1233) but at the time I had convinced myself that we did not want absolute paths. My bad.

nmaludy:
@nicklewis i wrote a unit test instead of an integration test for this.

Also, going to bump the build because it looks like the Travis failure is random.

donoghuc:
@nmaludy I think its rubocop error.

Issue #1245, opened 2019-09-24
Standardize JSON and Puppet language Result objects


adreyer:
## Use Case
It can be confusing that the JSON representation of a Result or ResultSet do not behave the same way as the puppet objects. They should be harmonized as part of the 2.0 release of bolt.

## Describe the Solution You Would Like
* Add a new `status` attribute to a the Result object that is a string.
* Remove `node` from Result status_hash and attributes in favor of `target` (read: the JSON representation of the Result)
* Rename `result`  to `value` in the Result JSON representation

## Describe Alternatives You've Considered
These were decided to be the least breaking changes to harmonize these object.

## Additional Context
We should also add a key to the config `future` which will be in front of this feature (so when the flag is true, this is on). The config key should be globally accessible, so we can read it from anywhere without having to pass config around.

donoghuc:
Closed with https://github.com/puppetlabs/bolt/pull/1281

PR #1281, opened 2019-10-03
(GH-1245) Standardize JSON and Puppet Result objects

donoghuc:
This commit standardizes the Result object interface in plans with the JSON representation. This involved exposing a new `status` method on the Puppet Type, removing the `node` key from the status hash (in favor of `target`), and renaming `result` to `value`. This change is able to be toggled on and off with the newly introduced `$future` global variable.

### DIFF
--- a/bolt-modules/boltlib/lib/puppet/datatypes/result.rb
+++ b/bolt-modules/boltlib/lib/puppet/datatypes/result.rb
@@ -10,6 +10,7 @@ Puppet::DataTypes.create_type('Result') do
       error => Callable[[], Optional[Error]],
       message => Callable[[], Optional[String]],
       action => Callable[[], String],
+      status => Callable[[], String],
       to_data => Callable[[], Hash],
       ok => Callable[[], Boolean],
       '[]' => Callable[[String[1]], Data]
--- a/lib/bolt/config.rb
+++ b/lib/bolt/config.rb
@@ -33,7 +33,7 @@ module Bolt
   class Config
     attr_accessor :concurrency, :format, :trace, :log, :puppetdb, :color, :save_rerun,
                   :transport, :transports, :inventoryfile, :compile_concurrency, :boltdir,
-                  :puppetfile_config, :plugins, :plugin_hooks
+                  :puppetfile_config, :plugins, :plugin_hooks, :future
     attr_writer :modulepath
 
     TRANSPORT_OPTIONS = %i[password run-as sudo-password extensions
@@ -163,6 +163,8 @@ module Bolt
       # Plugins are only settable from config not inventory so we can overwrite
       @plugins = data['plugins'] if data.key?('plugins')
       @plugin_hooks.merge!(data['plugin_hooks']) if data.key?('plugin_hooks')
+
+      @future = data['future'] == true
 
       %w[concurrency format puppetdb color transport].each do |key|
         send("#{key}=", data[key]) if data.key?(key)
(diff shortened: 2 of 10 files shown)


nmaludy:
:+1: 

This is the #1 complain i hear from my engineers when starting up with bolt plans. Their immediate reaction is to do a `out::message($results)` and they see the data they want in the JSON, but the API to access that data is different.

Huge bonus here for usability and ease of adoption!

bolt-modules/boltlib/lib/puppet/datatypes/result.rb
```diff
@@ -1,20 +1,40 @@
...
-      error => Callable[[], Optional[Error]],
-      message => Callable[[], Optional[String]],
-      action => Callable[[], String],
-      to_data => Callable[[], Hash],
-      ok => Callable[[], Boolean],
-      '[]' => Callable[[String[1]], Data]
-    }
-  PUPPET
+  # rubocop:disable Style/GlobalVars
+  if $future
```
adreyer: Is there any breaking change here?

lib/bolt/config.rb
```diff
@@ -155,6 +155,13 @@ def update_from_file(data)
         @puppetfile_config = data['puppetfile'].select { |k, _| PUPPETFILE_OPTIONS.include?(k) }
       end
 
+      # Set global $future if defined
+      if data.key?('future')
```
adreyer: I think after parsing all the config we should set this probably in the CLI class rather than setting it as a side effect of reading the config file.

bolt-modules/boltlib/lib/puppet/datatypes/result.rb
donoghuc: Oh, I guess not. I think I got too excited about `$future`. Will update.

donoghuc:
closes GH-1245

lib/bolt/config.rb
```diff
@@ -164,6 +164,8 @@ def update_from_file(data)
       @plugins = data['plugins'] if data.key?('plugins')
       @plugin_hooks.merge!(data['plugin_hooks']) if data.key?('plugin_hooks')
 
+      @future = data['future'] == true
```
adreyer: Is this a rubocopism?

spec/integration/result_set_spec.rb
```diff
@@ -49,6 +49,19 @@ def run_plan(plan, params)
       end
     end
 
+    context 'with $future flag' do
+      it 'exposes `status` method for result and renames `result` to `value`' do
+        with_tempfile_containing('conf', { future: true }.to_json) do |conf|
+          params = { nodes: uri }.to_json
+          additional_flags = config_flags + ['--configfile', conf.path.to_s]
+          result = run_cli(['plan', 'run', 'results::test_result', "--params", params] + additional_flags)
```
adreyer: We should update run_cli to unset `$future`

lib/bolt/config.rb
donoghuc: No, my thought was that `future` should only have two possible values:  `true` and `false. The only way it is `true` is if we set `future: true` in `bolt.yaml`. Should we instead *only* set the `$future` global when `data['future']` is true?

donoghuc: Should we *only* set $future if `future: true`? Instead of always setting `$future` to either `true` or `false`?

donoghuc:
Closes https://github.com/puppetlabs/bolt/issues/1245

PR #1282, opened 2019-10-03
(DOCS) move predocs to documentation

adreyer:


adreyer:
oh this will probably break the rakes doc task

Issue #1178, opened 2019-08-28
Make ResultSet array indexable


lucywyman:
## Use Case
As a plan author I want to access Result objects in a ResultSet by array indexing

## Describe the Solution You Would Like
```
$result = run_plan('module::plan', $nodes)
return $result[0].value
```
ResultSet should be able to be accessed like an array, and should behave similar to typical array accessing.

donoghuc:
closed by https://github.com/puppetlabs/bolt/pull/1283

PR #1283, opened 2019-10-03
(GH-1178) Make ResultSet array indexible

donoghuc:
This commit allows the `ResultSet` object to be indexed like an array.


lucywyman:
Closes #1178 

Travis looks to be rubocop errors

bolt-modules/boltlib/lib/puppet/datatypes/resultset.rb
```diff
@@ -17,6 +17,7 @@
       ok_set => Callable[[], ResultSet],
       targets => Callable[[], Array[Target]],
       to_data => Callable[[], Array[Hash]],
+      '[]' => Callable[[Integer], Data]
```
lucywyman: I think this has the same return type as `find`, which is `Optional[Variant[Result, ApplyResult]]`

donoghuc: I'm having a hard time finding reference material for how to specify an optional parameter with the `Callable` type. Looking at https://puppet.com/docs/puppet/latest/lang_data_abstract.html#the-callable-data-type and https://docs.google.com/document/d/1D1yXXDIw6a67o_hLeqdHWDq2bM41C4eToeJK9tzUc98/edit?usp=sharing but not comming up with the right syntax.

lucywyman:
I think this also needs CHANGELOG + docs

bolt-modules/boltlib/lib/puppet/datatypes/resultset.rb
nicklewis: Can you use `Variant[Callable[[], Data], Callable[[Integer], Data]]`?

donoghuc: ah, yeah. Thanks Nick!

CHANGELOG.md
```diff
@@ -2,6 +2,14 @@
 
 ## 1.32.0
 
+### New features
+
+* **The `ResultSet` type is now indexible.** ([#1178](https://github.com/puppetlabs/bolt/issues/1178))
```
```suggestion
* **The `ResultSet` type is now indexable.** ([#1178](https://github.com/puppetlabs/bolt/issues/1178))
```

lib/bolt/result_set.rb
```diff
@@ -106,5 +106,13 @@ def to_s
     def ==(other)
       eql?(other)
     end
+
+    def [](from, up_to = nil)
+      if up_to
+        @results[from, up_to]
```
lucywyman: While this is the right syntax for Puppet, it looks like for Ruby this doesn't support negative indexing (which the Puppet one does support :P). I think making this `@results[from..up_to]` will work as the Puppet one does, which is what I think users will expect.

Example:
```
out::message("${$result[1, -1]}")
```

donoghuc: wow, good catch. Thanks!

CHANGELOG.md
```diff
@@ -2,6 +2,14 @@
 
 ## 1.32.0
 
+### New features
+
+* **The `ResultSet` type is now indexable.** ([#1178](https://github.com/puppetlabs/bolt/issues/1178))
+
+  When working with `ResultSet` types in plans the bracket `[]` operator can be used to get `Results` by index.
+
+## 1.32.0
```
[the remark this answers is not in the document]
donoghuc: Oh, oops. I will update than when I resolve merge conflict in other PR.

Issue #1256, opened 2019-09-26
Log file uploads


nicklewis:
## Use Case
Currently, there's no good way to diagnose which files are uploaded by Bolt as part of running a task. They're uploaded to a random directory and deleted after Bolt runs, so you can't look at them on the filesystem. There are also not log messages associated with file uploads for every transport, so you can't check whether Bolt uploaded the right things. This can also lead users to the appearance that Bolt isn't doing the right thing, when it did in fact upload the correct files.

## Describe the Solution You Would Like
Each transport should issue  debug message for each file that is uploaded, just as it does for each command that is run.

## Describe Alternatives You've Considered
Another way to handle this uncertainty about what Bolt is doing would be to add an option to preserve the tempdir after Bolt runs, so that users can log in and inspect it. That would likely require a new CLI flag and would also cause trouble for cases where Bolt creates multiple tempdirs. It also would require Bolt have some clear way to tell users where the tempdir is.

donoghuc:
Closed by https://github.com/puppetlabs/bolt/pull/1287

PR #1287, opened 2019-10-07
(GH-1256) Add debug logging for transport file transfer

donoghuc:
With this commit a debug log message is produced when a transport uploads a file to a target. Note this excludes the orch transport.


lib/bolt/transport/winrm/connection.rb
```diff
@@ -156,6 +156,7 @@ def write_remote_file(source, destination)
         end
 
         def write_remote_file_winrm(source, destination)
+          @logger.debug { "Uploading #{source}, to #{destination}" }
```
lucywyman: It seems like there's a few other methods in this file that need this - I think:
```
lucy ~/githubs/bolt (master) $ ag -r "def write" lib/bolt/transport/winrm/connection.rb 
150:        def write_remote_file(source, destination)
158:        def write_remote_file_winrm(source, destination)
165:        def write_remote_file_smb(source, destination)
218:        def write_remote_executable(dir, file, filename = nil)
226:        def write_executable_from_content(dir, content, filename)
280:        def write_remote_file_smb_recursive(tree, source, dest)
```

lib/bolt/transport/docker/connection.rb
```diff
@@ -79,6 +79,7 @@ def execute(*command, options)
         end
 
         def write_remote_file(source, destination)
+          @logger.debug { "Uploading #{source}, to #{destination}" }
```
lucywyman: Same here:
```
81:        def write_remote_file(source, destination)
88:        def write_remote_directory(source, destination)
126:        def write_remote_executable(dir, file, filename = nil)
```

lib/bolt/transport/ssh/connection.rb
```diff
@@ -261,6 +261,7 @@ def execute(command, sudoable: false, **options)
         end
 
         def copy_file(source, destination)
+          @logger.debug { "Uploading #{source}, to #{destination}" }
```
lucywyman: Also `269:        def write_executable_from_content(dest, content, filename)`

lucywyman:
Closes #1256 

I think this also needs changelog - I wonder if we should add a 'Bolt NEXT' changelog right after shipping the last Bolt, maybe even as part of CI...

lib/bolt/transport/ssh/connection.rb
donoghuc: Should this just be `writting wrapper script to #{destination}`? I kind of held back from logging this because it is something we do behind the scenes.

lib/bolt/transport/winrm/connection.rb
donoghuc: oh, I think write_remote_file should cover all those. I'm not sure what `write_executable_from_content` is used for (actually for both ssh and winrm i dont think that method is used).

lib/bolt/transport/ssh/connection.rb
donoghuc: unless I can finid what those methods are for i'm going to delete them.

donoghuc:
I'm not sure what to do about changelog. I've been waiting until something is basically approved or in my mind the PR is fully ready to go. I have been adding them as a separate commit because there will almost certainly be merge conflicts.

lib/bolt/transport/ssh/connection.rb
donoghuc: I think net-scp is smart enough to know that the source is a stringIO. Though i guess I wont want to log that.

CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # Changelog
 
+## BOLT NEXT
+
+### New features
+
+* **Debug logging for file transfer** ([#1256](https://github.com/puppetlabs/bolt/issues/1256))
+
+  When a file is transferred to a target the source and destination are logged at debug level.
```
K8med: @donoghuc -  was this not being logged before or was it being logged at a level other than debug? Can you clarify in the text and re-push?

PR #1295, opened 2019-10-08
(maint) Delete signal handling tests

donoghuc:
These signal handling tests are no longer compatible with rspec-mocks (3.9.0). They provide little value given they test something that is unlikely to regress.


donoghuc:
https://github.com/rspec/rspec-mocks/commit/24442456174015b6ab7a786e78abbf894d0bbc19#diff-5e732c165cb40508878a32c8c6bd7ad5 commit that broke tests.

nicklewis:
So long, farewell

PR #1297, opened 2019-10-08
(GH-1296) Added YAML parsing error details to the bolt logging

### LINKED ISSUE #1296: Provide helpful error output when parsing YAML file fails
## Use Case
As a new user, you often make mistakes when writing `bolt.yaml` and `inventory.yaml` files.

Currently when you make an error in one of these files, or they don't conform to the right schema an error is produced:

```
Could not parse inventory file: inventory.yaml
```

As a new user, you're not sure where the problem is so you're left guessing.

This was very evident at the Bolt Workshop before Puppetize PDX 2019

Couple ideas here:
* Provide some linting capability for YAML syntax bugs. Errors with what's broken or line-numbers where the problem is would be super helpful.
* Provide some schema validation and helpful errors when the schema of the file doesn't match the valid schema defined by Bolt.

nmaludy:
Closes #1296 

**TODO**
- [x] Tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@
 
   When the `plugin_hooks` key was not set for a target/group in Inventory v2 the `apply_prep` function would not work. Bolt now uses the default `plugin_hooks` and honors the `plugin_hooks` from bolt config when using Inventory v2. 
 
+* **Better error output when parsing malformed `yaml` files** ([#1296](https://github.com/puppetlabs/bolt/issues/1296))
+
+  Previously when parsing `yaml` config files a generic error message was surfaced with no information about where in the file the problem occurred. Now an error message that contains the path to the file as well as the line and column in the file where the error originated from.
+
 ## 1.33.0
 
 ### Bug fixes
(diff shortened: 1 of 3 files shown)


adreyer:
`(<unknown>): did not find expected node content while parsing a flow node at line 4 column 4`

Some of these errors are really cryptic. I wonder if we're better off using the `err.column` and `err.line` only.

nmaludy:
Testing out the IO errors for an inventory file with bad owner / read permissions results in:
```
Could not read inventory file: /home/nmaludy/storage/home/src/git/nmaludy/bolt/bolt/workshop/inventory.yaml
error: Permission denied @ rb_sysopen - /home/nmaludy/storage/home/src/git/nmaludy/bolt/bolt/workshop/inventory.yaml
```

nmaludy:
Parsing an inventory with the following in it:

```yaml
---
version: 2
config:
  transport: winrm
    ssl-verify: false
    ssl: true
```

results in the following error:

```
Could not parse inventory file: /home/nmaludy/storage/home/src/git/nmaludy/bolt/bolt/workshop/inventory.yaml
error: (<unknown>): mapping values are not allowed in this context at line 5 column 15
```

nmaludy:
Cool, i think this is good to go then (as long as the CI passes)

PR #1298, opened 2019-10-08
python 2.4 json module does not work with this code, but the simplejson module does

sirinek:
The example code does not work on RHEL5. The Python 2.4 json module does not work correctly, but the simplejson module does.

Later versions of python are okay.


nicklewis:
Can you use the simplejson module on the other versions of python? ie. can we just use that unconditionally here?

PR #1301, opened 2019-10-10
(maint) Load plugin modules from builtin modulepath

donoghuc:
Now that we are shipping bundled module plugins we need to load them from the default modulepath.

### DIFF
--- a/lib/bolt/plugin.rb
+++ b/lib/bolt/plugin.rb
@@ -143,11 +143,12 @@ module Bolt
       @analytics = analytics
       @plugin_context = PluginContext.new(config, pal)
       @plugins = {}
+      @pal = pal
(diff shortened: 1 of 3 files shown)


lib/bolt/plugin.rb
```diff
@@ -143,11 +143,12 @@ def initialize(config, pal, analytics)
       @analytics = analytics
       @plugin_context = PluginContext.new(config, pal)
       @plugins = {}
+      @pal = pal
```
donoghuc: Maybe i'm reading this wrong but it looks like we meant to have `pal` be an attribute anyway https://github.com/puppetlabs/bolt/pull/1301/files#diff-e55bc985f91060535fd013d69f316af6R139

PR #1302, opened 2019-10-10
(packaging) Finalize changelog for 1.33.0

donoghuc:


CHANGELOG.md
```diff
@@ -10,7 +10,7 @@
 
 ### New features
 
-* **Only print `safe_name` when referencing `Target`s** ([#1271](https://github.com/puppetlabs/bolt/pull/1271))
+* **Only print `safe_name` when referencing `Target`s** ([#1243](https://github.com/puppetlabs/bolt/issues/1243))
```
cwcowellshah: Recommend rewording to avoid confusion around the scope of the word "only":

```suggestion
* **When referring to `Target`s in log or output, use the `safe_name`** ([#1243](https://github.com/puppetlabs/bolt/issues/1243))
```

CHANGELOG.md
```diff
@@ -10,7 +10,7 @@
...
 
   When using Inventory v2 a Target's safe name is the `uri` minus the password component (unless the target `name` is explicitly defined in which case the `name` is used). For Inventory v1, the safe name is the `host`.
```
cwcowellshah:
```suggestion
  When using inventory version 2, a `Target`'s safe name is the `uri` minus the password (unless the `Target`'s has an explicitly defined `name`, in which case the safe name is the value of `name`). For inventory version 1, the safe name is the value of `host`. 
```

cwcowellshah:
I'm not able to get this branch and push suggested text changes back up to this PR right now, so I'll add a wording suggestion comment here instead:

Recommend changing this text:
>When working with `ResultSet` types in plans the bracket `[]` operator can be used to get Results by index.`

to use active voice, like this:
>When wording with `ResultSet` types in plans, use the bracket `[]` operator to get Results by index.

cwcowellshah:
Recommend changing this changelog entry header:

>**Debug logging for file transfer**

To this, to avoid confusion about whether this change has debugged the logging, or has added a new logging message at the "debug" level:

>**Log file transfer details at `debug` level.

PR #1303, opened 2019-10-11
(maint) Fix apply prep for inventory v2

donoghuc:
Previously the default plugin_hooks from config were not being set on a Target2 object. This was causing failures in apply prep https://github.com/puppetlabs/bolt/blob/345131e31c02a3047c1607ce3b6df0a484976542/bolt-modules/boltlib/lib/puppet/functions/apply_prep.rb#L103 which assumes that `plugin_hooks` will have a value. This commit updates the inventory 2 update_target_state to consider the `plugin_hooks` set in config. The merge order in order of precedence is target level in inventory, group level in inventory, specified in bolt.yaml, default.

### DIFF
--- a/bolt-modules/boltlib/spec/functions/apply_prep_spec.rb
+++ b/bolt-modules/boltlib/spec/functions/apply_prep_spec.rb
@@ -208,6 +208,48 @@ describe 'apply_prep' do
         expect(inventory.facts(target)).to eq(fact)
       end
     end
+
+    context 'with default plugin inventory v2' do
+      let(:hostname) { 'agentless' }
+      let(:data) {
+        {
+          'version' => 2,
+          'targets' => [{ 'uri' => hostname }]
+        }
+      }
+
+      let(:config) { Bolt::Config.new(Bolt::Boltdir.new('.'), {}) }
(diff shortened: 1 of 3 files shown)


bolt-modules/boltlib/spec/functions/apply_prep_spec.rb
```diff
@@ -208,6 +208,48 @@
...
+          'targets' => [{ 'uri' => hostname }]
+        }
+      }
+
+      let(:config) { Bolt::Config.new(Bolt::Boltdir.new('.'), {}) }
+      let(:pal) { nil }
+      let(:plugins) { Bolt::Plugin.new(config, pal, Bolt::Analytics::NoopClient.new) }
+      let(:inventory) { Bolt::Inventory.create_version(data, config, plugins) }
+      let(:target) { inventory.get_target(hostname) }
+      let(:targets) { inventory.get_targets(hostname) }
```
lucywyman: I don't think this gets used anywhere?

donoghuc: Yes, targets is used in the before each block.

PR #1304, opened 2019-10-11
(Docs) Remove duplicate entry in ditamap

alexjfisher:
I'm assuming this file is used in building https://puppet.com/docs/bolt/latest/writing_tasks_and_plans.html where `Running tasks` currently appears twice under the `Tasks and plans` menu.


donoghuc:
Added do not merge while I sort out jenkins job for building preview docs based on PR.

donoghuc:
Automated build does not delete files. We will need to delete that other one separately. This will prevent that from re-generating in the future though.

PR #1311, opened 2019-10-14
(GH-1207) Run task function should respect _noop option

### LINKED ISSUE #1207: Unable to call run_task() in noop mode from a plan

lucywyman:
When running a task from a plan, the `run_task` function should respect
the `_noop` option. Now when the `run_task` function is called either
with an executor in noop mode or with the `_noop` option it will run in
noop.

Closes #1207

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,10 @@
 
   Previously when parsing `yaml` config files a generic error message was surfaced with no information about where in the file the problem occurred. Now an error message that contains the path to the file as well as the line and column in the file where the error originated from.
 
+* **Run task function should respect _noop option** ([#1311](https://github.com/puppetlabs/bolt/pull/1311))
+
+  Previously when calling the `run_task` function from a plan with the `_noop` metaparameter, `_noop` would not be passed to the task. Now the parameter is correctly passed to the task.
+
 ## 1.33.0
 
 ### Bug fixes
--- a/bolt-modules/boltlib/lib/puppet/functions/run_task.rb
+++ b/bolt-modules/boltlib/lib/puppet/functions/run_task.rb
@@ -12,7 +12,7 @@ Puppet::Functions.create_function(:run_task) do
   # Run a task.
   # @param task_name The task to run.
   # @param targets A pattern identifying zero or more targets. See {get_targets} for accepted patterns.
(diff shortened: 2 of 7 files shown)


donoghuc:
I linked to a kind of ambiguous commend https://github.com/puppetlabs/bolt/issues/1207#issuecomment-539149226 trying to convey the resolution for https://github.com/puppetlabs/bolt/pull/1237#pullrequestreview-296465282 . We thought that cleaning up that dead code path (or actually implementing it fully) would be resolved with this ticket.

lucywyman:
I removed adding `[:noop]` to the plan run parameters if you pass in `--noop`, since it sounds like supporting that for plans is an entirely separate bit of work.

Issue #1294, opened 2019-10-08
Extract `install_agent` plugin to `puppet_agent`


adreyer:
Since the `puppet_agent` module now has the ability to stop the puppet service we should replace the `install_agent` plugin with a `puppet_agent` module plugin.

donoghuc:
Closed with https://github.com/puppetlabs/bolt/pull/1312 shipped in 1.35.0!

PR #1312, opened 2019-10-14
(GH-1294) Use the puppet_agent module instead of a built in plugin.

adreyer:


lucywyman:
@adreyer Looks to be missing a `task_plugin::apply_prep` fixture?

adreyer:
yeah I'm just refactoring the task_spec tests so they look like the module_spec ones and don't need any special files.

adreyer:
Why don't we install the puppetfile in the repo?

CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 ## Bolt Next
 
+### Breaking Changes
```
K8med: Note: I haven't seen a "Breaking changes" header before, I'm checking that it's standard? Will we continue to use it as needed?

adreyer: hmm I think deprecations may be the standard header

PR #1316, opened 2019-10-15
(docs) Fix absolute links and correct puppetdb config location

lucywyman:
This fixes some links which where hardcoded to in-repo markdown docs to
be relative so they link to the production site pages. This also
corrects a misprint that the only config location for puppetdb
connection is `~/.puppetlabs/bolt`, when it is actually the Bolt config
file you're currently using (regardless of location).


lucywyman:
Weird it looks like @K8med and @cwcowellshah didn't get pinged on this?

documentation/using_plugins.md
```diff
@@ -126,7 +126,14 @@ json.dump({'value': secret}, sys.stdout)
...
+The PuppetDB plugin supports looking up targets from PuppetDB. It takes a `query` field, which is
+either a string containing a [PQL](https://puppet.com/docs/puppetdb/latest/api/query/v4/pql.html)
+query or an array containing a [PuppetDB
+AST](https://puppet.com/docs/puppetdb/latest/api/query/v4/ast.html) format query. The query is used
+to determine which targets should be included in the group. If `name` or `uri` is not specified with
+a fact lookup then the `[certname]` for each target in the query result will be used as the `uri`
```
donoghuc: Seems like removing the link to `fact lookup` is correct as there is no header for that. Do we need to add a section for that?

lucywyman: @adreyer @K8med ^^  I don't think so, 'fact lookup' feels like it's relatively self explanatory in context? But that could also be because I know what we mean :stuck_out_tongue:

documentation/inventory_file_v2.md
```diff
@@ -178,751 +178,6 @@ targets:
...
-plugins:
-  vault:
-    server_url: https://127.0.0.1:8200
-    cacert: /path/to/cert
-    auth:
-      method: token
-      token: <XXXX-XXXX>
-```
-
-## Inventory config
```
adreyer: The rest of this is duped from inventory v1 not the plugin docs right?

documentation/using_plugins.md
adreyer: Since we don't describe it until lower I think something more could be said at this point.

documentation/inventory_file_v2.md
lucywyman: https://puppet.com/docs/bolt/latest/inventory_file.html#inventory-config Yes - should I still have a section in this doc that links to that? That seems useful - I'll update

documentation/using_plugins.md
lucywyman: Ok, added another header level and linked this to that.

lucywyman:
https://docs-preview.webteam.puppet.com/docs/bolt/latest/using_plugins.html#puppetdb
https://docs-preview.webteam.puppet.com/docs/bolt/latest/plan_functions.html

documentation/inventory_file_v2.md
```diff
@@ -176,753 +176,13 @@ targets:
     uri: 192.168.100.179
 ```
 
-### Plugins and dynamic inventory
+## Plugins and dynamic inventory
 
-Inventory plugins can be used to dynamically load information into the inventory file.
+Inventory plugins can be used to dynamically load information into the inventory file. See
+[Using Plugins](using_plugins.md) for more information on using Bolt's builtin plugins, and
```
cwcowellshah:
```suggestion
[Using Plugins](using_plugins.md) for more information on using Bolt's built-in plugins, and
```

Per the official Puppet dictionary (Merriam Webster's 11th Collegiate).

documentation/using_plugins.md
```diff
@@ -126,7 +126,14 @@ json.dump({'value': secret}, sys.stdout)
...
+The PuppetDB plugin supports looking up targets from PuppetDB. It takes a `query` field, which is
+either a string containing a [PQL](https://puppet.com/docs/puppetdb/latest/api/query/v4/pql.html)
+query or an array containing a [PuppetDB
+AST](https://puppet.com/docs/puppetdb/latest/api/query/v4/ast.html) format query. The query is used
```
cwcowellshah:
```suggestion
AST](https://puppet.com/docs/puppetdb/latest/api/query/v4/ast.html)-format query. The query is used
```

cwcowellshah: Recommend `The query determines which targets...`

documentation/using_plugins.md
```diff
@@ -126,7 +126,14 @@ json.dump({'value': secret}, sys.stdout)
...
+to determine which targets should be included in the group. If `name` or `uri` is not specified with
```
cwcowellshah:
```suggestion
which targets to include in the group. If `name` or `uri` is not specified with
```

PR #1326, opened 2019-10-16
(GH-1310) Remove generated docs, update docs rake task

### LINKED ISSUE #1310: Build auto-generated plan function reference in docs CI
Currently when a change to the puppet strings docs in plan function source code is committed the author must run the [docs rake task](https://github.com/puppetlabs/bolt/blob/03f05e755fd39e60bf20539a0fdaab7f5f69e32a/Rakefile#L59) and commit that update. Checking in the update to the markdown is enforced with a [check](https://github.com/puppetlabs/bolt/blob/03f05e755fd39e60bf20539a0fdaab7f5f69e32a/.travis.yml#L69-L74) in travis CI. Instead of requiring authors of commits that change plan function reference source to run the rake task and check that docs code in, generating the docs should be part of the docs publishing job in jenkins.

To do:
- [x] Remove the file from `documentation` directory
- [x] Add the file to gitignore 
- [x] Add generating the file to the [CI script](https://github.com/puppetlabs/ci-job-configs/blob/master/resources/scripts/publish-docs.sh)
- [x] Update travis to only check that the doc builds cleanly

lucywyman:
This PR:
- Removes the generated doc `plan_functions.md` and adds it to
`.gitignore`
- Removes the git diff check for the plan_functions page from travis
- Updates the pattern for Strings to find files in to just be the
`bolt-modules` directory, so that warnings about `lib/*.rb` don't
clutter the output.

Closes #1310


.travis.yml
```diff
@@ -67,7 +67,7 @@ script:
     false
   fi
 - |
-  if ! bundle exec rake docs; then
+  if ! bundle exec rake docs_test; then
```
adreyer: I think we can just call the regular docs task and skip the git diff check

Rakefile
```diff
@@ -60,7 +60,7 @@ desc "Generate markdown docs for Bolt's core Puppet functions"
 task :docs do
   FileUtils.mkdir_p 'tmp'
   tmpfile = 'tmp/boltlib.json'
-  PuppetStrings.generate(PuppetStrings::DEFAULT_SEARCH_PATTERNS,
```
donoghuc: Why this change? https://github.com/puppetlabs/puppet-strings/blob/7ab471857f957836759ea0e666568d60fa25b954/lib/puppet-strings.rb#L4-L11 seem ok?

lucywyman: It includes `lib/**/*.rb` which results in a lot of warnings for files that we don't intend to write docs strings for. It just seems unnecessary when we already have to specify the actual path we want to build docs from.

PR #1329, opened 2019-10-17
(maint) ignore files created for running bolt from source

adreyer:


.gitignore
```diff
@@ -34,6 +34,11 @@ Gemfile.local
 .vagrant/
 .byebug_history
 
+# Running Bolt from source
+.rerun.json
+Boltdir/
+bolt.yaml
```
donoghuc: I wanted to do a similar thing https://github.com/puppetlabs/bolt/pull/981/files a while back but there was no interest at the time. I think this would be helpful and would *love" to see it ignored in rubocop as well.

PR #1331, opened 2019-10-17
(GH-1270) Support limit for `do_until`

### LINKED ISSUE #1270: Add parameter limit to function do_until()
## Use Case
Add a max-loops option to do_until so it will terminate given a maximum number of iterations reached.

See - https://github.com/puppetlabs/bolt/pull/844

adreyer:
This adds tests to #844 

Closes #1270

### DIFF
--- a/bolt-modules/ctrl/spec/functions/ctrl/do_until_spec.rb
+++ b/bolt-modules/ctrl/spec/functions/ctrl/do_until_spec.rb
@@ -3,13 +3,32 @@
 require 'spec_helper'
 
 describe 'ctrl::do_until' do
-  it {
-    count = 0
-    seq = ['something', false, false]
+  let(:seq) { [false, 'truthy', false, false] }
+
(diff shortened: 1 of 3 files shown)


adreyer:
waiting for release to add a changelog

donoghuc:
Changelog entry is un-blocked. we have a tag.

bolt-modules/ctrl/spec/functions/ctrl/do_until_spec.rb
```diff
@@ -3,13 +3,32 @@
 require 'spec_helper'
 
 describe 'ctrl::do_until' do
-  it {
-    count = 0
-    seq = ['something', false, false]
+  let(:seq) { [false, 'truthy', false, false] }
```
lucywyman: Should 'truthy' be the last element here, assuming it's ordered? I don't think the 'exits early with limit 2' test will actually test exiting early with this order.

Issue #1322, opened 2019-10-16
Make it easier to optionally set module based plugins either bolt config or inventory


donoghuc:
## Use Case
As a plugin consumer and author I want to be able to set parameters to be used in the plugin implementations either in `bolt.yaml` or `inventory.yaml`. Currently this requires handling merging the top level parameters passed to the task implementations with the keys nested under `_config`. This also prevents using metadata to require a parameter when it could be provided in either `bolt.yaml` or `inventory.yaml`. As a plugin author I do not want to have to handle munging that data and dealing with precedence. 

## Describe the Solution You Would Like

* If a bolt_plugin.json file has a config specification it should continue to behave as it does today for now. The config should be validated against it and passed as `_config`.
*  If the bolt_plugin.json file does not have a config specification should be the intersection of the task parameters for each hook. Each config value should be treated as optional since it could be passed as an option to the plugin instead
* If the bolt_plugin.json file does not have a config section when the plugin runs the options passed to the specific invocation of the hook should be merged into the config values. IE the config values are providing defaults for the task parameters. The `_config` hash should not be passed.

The expectation of this work is that `_config` and setting `config` in `bolt_plugin.yaml` will be deprecated.

lucywyman:
If no config is defined in `bolt_plugin.json` then we accept any config that would work for all of the hook tasks.

PR #1333, opened 2019-10-17
(GH-1322) Validate plugin config against task metadata

donoghuc:
Previously when a config section existed in a `bolt_plugin.json` specification any config that was set in `bolt.yaml` would be passed to the task as under the key `_config`. Tasks that expect values to be set in either `bolt.yaml` or `inventory.yaml` would have to handle munging the values gathered from either file. This commit begins to phase out the config section in `bolt_plugin.json`. If there is not a config section in that file config values are validated against the intersection of the parameter metadata for the tasks that implement each hook. Values set in `bolt.yaml` are merged with those in inventory.yaml and passed as parameters. No `_config` is passed along.

Closes https://github.com/puppetlabs/bolt/issues/1322

### DIFF
--- a/lib/bolt/plugin/module.rb
+++ b/lib/bolt/plugin/module.rb
@@ -35,9 +35,18 @@ module Bolt
       # This method interacts with the module on disk so it's separate from initialize
       def setup
         @data = load_data
-        @config_schema = process_schema(@data['config'] || {})
-
         @hook_map = find_hooks(@data['hooks'] || {})
+        # If there is a config section in bolt_plugin.json, validate against that and send
+        # validated values nested under `_config` key. Otherwise validate againsts the intersection
+        # of all task schemas.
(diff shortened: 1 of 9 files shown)


lib/bolt/plugin/module.rb
```diff
@@ -150,15 +160,29 @@ def process_params(task, opts)
...
         metaparams['_boltdir'] = @context.boltdir
 
         validate_params(task, params)
         [params, metaparams]
       end
 
+      def extract_task_parameter_schema
+        @hook_map.each_with_object({}) do |(_hook, task), acc|
+          if (schema = task['task'].metadata['parameters'])
+            acc.merge!(schema)
```
donoghuc: This needs to be the *intersection* of types. Also, while i'm here I might as well wrap in `Optional`.

Issue #1314, opened 2019-10-14
Auto-generate CLI docs


adreyer:
The cli docs (documentation/bolt_command_reference.md) should be auto-generated by the option parser.

PR #1334, opened 2019-10-18
(GH-1314) Auto-generate command reference docs

beechtom:
This commit adds a rake task to automatically generate the command reference docs
from the option parser.


beechtom:
What do you mean by deleted file?

donoghuc:
documentation/bolt_command_reference.md should be deleted and only generated when the docs shipping pipelines run. Sorry my preview had shown that as a collapsed diff and I assumed it had been deleted (which is what we expect).

Rakefile
```diff
@@ -56,6 +57,41 @@ def format_links(text)
   text.gsub(/{([^}]+)}/, '[`\1`](#\1)')
 end
 
+desc "Generate markdown docs for Bolt's command line options"
+task :reference_docs do
```
donoghuc: In order to keep the rake file organized I think we should make a docs namespace similar to the integration namespace. Then create a task that runs everything under that namespace for use in travis or building docs locally to test out changes. 
So maybe a namespace called `docs`, then nest two tasks `function_reference` and `cli_reference` under those, you could make a task like
```
desc "generate docs"
task generate_docs: [:docs, 'plan_function', 'cli_reference']
```

Rakefile
```diff
@@ -56,6 +57,44 @@ def format_links(text)
...
+        {
+          short: switch.short.first,
+          long: switch.long.first,
+          arg: switch.arg,
+          desc: switch.desc.map { |d| d.gsub("<", "&lt;") }.join("<br>")
+        }
+      end
+
+      @commands[command] = {
+        banner: help_text[:banner].gsub("<", "&lt;").gsub("\n", "<br>"),
```
donoghuc: Not sure why we need to sub the newlines? Maybe just sub the last `\n` with `<br>` i think it is causing only a single line from the banner to show up.

beechtom: That was to make sure when there is a single newline between lines that it displays correctly in the markdown doc. I'm not sure how that translates to the static site, so I'll remove the sub and do a preview of the site to make sure formatting isn't wonky.

documentation/bolt_command_reference.md.erb
```diff
@@ -0,0 +1,18 @@
...
+```plain
+<%= command[1][:banner] -%>
+```
+
+| Option | Description |
+| ------ | ----------- |
+<% for option in command[1][:options] -%>
+| <% if option[:short] %>`<%= option[:short] %>`, <% end -%>`<%= option[:long] -%><%= option[:arg] %>` | <%= option[:desc] %> |
+<% end %>
+<% end %>
```
donoghuc:
```suggestion
# Bolt command reference

These subcommands, actions, and options are available for Bolt.


<% @commands.each do |key, val| %>
## <%= key %>

\```plain
<%= val[:banner] -%>
\```

| Option | Description |
| ------ | ----------- |
<% val[:options].each do |option|  -%>
| <% if option[:short] %>`<%= option[:short] %>`, <% end -%>`<%= option[:long] -%><%= option[:arg] %>` | <%= option[:desc] %> |
<% end %>
<% end %>
```

donoghuc: The iteration in the templates using indexing does not feel very ruby-like to me. I think you can use something like the suggestion.

beechtom: Updated. We should probably update `reference.md.erb` to be more ruby-like as well.

PR #1336, opened 2019-10-18
    enhance bash completion

scoopex:
Enhancements: 

    - added automatic completion of groups and nodes
      (bolt group show, bolt inventory show --nodes all)
    - added some unimplemented completions

It would be a good idea to add "resources/bolt_bash_completion.sh" to the packaging process (i.e. dpkg). I haven't discovered packaging configuration in this repo.

Executing "bolt group show" or "bolt inventory show --nodes all" tends to be slowish on my machine.
It might be a good idea to accelerate this.


resources/bolt_bash_completion.sh
```diff
@@ -30,15 +39,30 @@ _bolt()
...
+            next="--boltdir --configfile -i --inventoryfile"
+         elif [ "$prevprev" == "inventory" ];then
+            next="-n --nodes -q --query --description --boltdir --configfile -i --inventoryfile"
+         fi
+      ;;
+      --nodes|-n|-t|--targets)
+         # executing "bolt group show" or "bolt inventory show --nodes all" tends to be slowish
+         # it might be a good idea to accelerate this
+         groups="$(bolt group show|grep -v -P '\d+ groups'|tr '\n' ' ')"
+         nodes="$(bolt inventory show --nodes all|grep -v -P '\d+ targets'|tr '\n' ' ')"
```
donoghuc: With the introduction of dynamic inventory in bolt inventory version 2 there will be some issues with this auto completion. You noticed that performance is not great (i'm assuming with inventory v1) because we essentially have to load all of bolt config and build up an inventory. For inventory v2 we will have to resolve all of dynamic inventory which could be even more costly. 

Another problem is that many of the inventory plugins allow/expect environment variables. When running the auto completion the inheritance of the environment variables may not work as expected. For example:
```
cas@cas-ThinkPad-T460p:~/working_dir/bolt$ PATH=$PATH:~/working_dir/dynamic-inventory-demo/terraform bolt command run "whoami" --nodes 
Could              executing          -h                 plugin
--debug            find               --help             resolve_reference
Error              --format           in                 terraform
executable         from               not                terraform-group:
```
When I try to auto-complete the PATH environment variable that tells the terraform plugin where to find the version of terraform I want to use is not passed to the invocation of bolt that generates the inventory and results in a difficult to debug list of "next" options. (the error message is `Error executing plugin terraform from resolve_reference in terraform-group: Could not find executable 'terraform'`). 

Perhaps one way around that could be to run just `"$(bolt inventory show --nodes all)"` and based on the return code there either raise the error message or at least not try to process it further. This would apply to the `group show` as well. 

The additional support for keys that have been added recently is great though! Thanks for the contribution.

donoghuc: More context for inventory v2 and plugin implications:
https://puppet.com/docs/bolt/latest/inventory_file_v2.html
https://puppet.com/docs/bolt/latest/using_plugins.html

scoopex: Slow lookups to the inventory will be not welcome by the users of bolt or the bolt command completion.

If you cannot speedup the access dramatically to the inventory you should think about adding a cache (invalidation by comparing the creation date of the cache- and inventory-file or by defining a maximum cache age).

Should i add a feature toggle, to disable lookups?

I haven't expected that bolt will output messages like that.
A good practice would be to just output warning- and error-messages on STDERR. 
This helps to deal with situations like that but would also be good for other scenarios.

donoghuc: We have been talking about how we want to handle caching, I think we will go ahead and merge this now given how useful it is (even though there is much room for improvement with `--targets` autocompletion). Directing errors/warnings to stderr is a good suggestion. Thanks for working on this!

Issue #1292, opened 2019-10-08
Extract vault plugin into a module


adreyer:
Extract the vault plugin into a module that ships with bolt.

donoghuc:
closed with https://github.com/puppetlabs/bolt/pull/1339

PR #1339, opened 2019-10-21
(GH-1292) Migrate Vault ruby plugin to be in a module

lucywyman:
This moves the Vault plugin from being a ruby plugin that lives in Bolt
to being a ruby task which lives in the puppetlabs-vault module.


Puppetfile
```diff
@@ -30,6 +30,7 @@ mod 'puppetlabs-ruby_task_helper', '0.4.0'
 
 # Plugin modules
 mod 'puppetlabs-azure_inventory', '0.2.0'
+mod 'puppetlabs-vault', '0.2.0'
```
donoghuc: Should this be 0.1.0 ?

donoghuc:
Ahh, boo merge conflict in puppetfile.

Issue #1147, opened 2019-08-13
Use resolve_reference plugins in more places


adreyer:
## Use Cases
* I want to use a plugin to load all inventory data
* I want to use the `pkcs7` or `prompt` plugin to load the `password` for the `vault` plugin
* I want to encrypt `vars` on a target or group.

## Out of scope
* A `resolve_reference[s]` function that calls a lookup plugin directly from a plan.

## Questions
* Do we need to limit the depth of a plugin returning a plugin?

## Prerequisites
This ticket assumes that `inventory_config` and `inventory_target` lookup hooks have been consolidated into a single `resolve_reference` hook.

## Where lookup plugins can be used
Any value in `inventory.yaml`
Items under `plugin_hooks` or `plugins` in bolt.yaml. 

## Where lookup plugins cannot be used
As a key in a yaml object anywhere in an inventoryfile/bolt.yaml
For `bolt.yaml`, only allow under an item in `plugin_hooks` and `plugins`. Disallow any other places.

## When plugins are evaluated
* When `bolt.yaml` is evaluated all references are resolved.
* If `bolt.yaml` config for one plugin references another that should happen before the first plugin is initialized. A helpful error should be raised if there is a dependency loop.
* A reference at the top level of inventory(the `all` group) is evaluated when inventory is loaded
* A reference at the top level of `groups` is evaluated when inventory is loaded
* A reference at the top level of `targets` is evaluated when inventory is loaded.
* A reference for a `name` or `uri` key should be evaluated when inventory is loaded
* A reference in `config`, `facts`, `vars` or `features` should be evaluated when it would be used by a `target`, specifically any time a `get_target(s)` call gets the target (lazily, but in a single step).
* Merging two references should merge the results of resolving the references.
* A `_plugin` object inside a second `_plugin` object should be evaluated immediately before the first `_plugin` object.
* A `_plugin` object returned from another reference should be evaluated following the above rules. ie if lookup at the top level of targets returns a `_plugin` inside a targets `config` it will be lazily loaded but if it just returns a `_plugin` object it would be immediately evaluated.

## Associated Changes
* The `targets`, `groups` and `features` array, should be flattened to generalize the existing behavior of a plugin that returns a list of targets or a single target.

lucywyman:
Closed by #1340

PR #1340, opened 2019-10-21
(GH-1147) Allow resolve_reference plugins everywhere in inventory

nicklewis:
Previously, resolve_reference plugins could only be used in the
inventory in select ways. Namely, they could only be used to lookup
targets or to set config on a group or target. They couldn't be used to
set any other attributes of a group (vars, facts, etc) or to set the
arguments to another plugin.

We now allow every value in the inventory to be specified using a plugin
reference, including the inventory itself. Keys are never allowed to be
plugin references.

Some references will be resolved as soon as the inventory is loaded.
For a group, these are `name`, `targets`, `groups`. For each target, we
will immediately resolve its `name`, uri`, and `alias`.

The other fields (`vars`, `facts`, `features`, `config`, `plugin_hooks`)
are termed "group/target data" and are resolved only when looking up a
target with that data.

We no longer use `Concurrent::Delay` objects to defer plugin evaluation.
Instead, the group tracks whether its data has been resolved and will do
it on demand if necessary. Similarly, the group tracks which of its
targets have been resolved and will resolve them as needed.

References are now allowed to return data containing more references,
which will be recursively re-evaluated until all the references are
gone. The exceptions to this rule are the `targets` and `groups` keys,
which will only be re-evaluated until the top-level object is no longer
a reference. This allows them to return references for their "data"
which will still be lazily resolved.

We now use an internal `Bolt::Inventory::Target` object to track the
current state of a target in the inventory and reconcile it with the
data set by its groups and config. Previously, this was a combination of
a hash on the inventory as well as the `Bolt::Target2` class. The
`Bolt::Inventory::Target` is now the canonical representation of the
current state of a target, tracking both the data set on that target
individually as well as querying the data from its group.

Instead of computing and caching the "final" merged data for the target,
we now only store the target-specific data and the cached group data
separately and compute results when needed. This allows us to invalidate
the group cache less frequently, as it now needs only be done when the
target is added to a new group and not whenever any value is set on the
target.

The `Bolt::Target2` class is now a thin, stateless wrapper around the
inventory and defers to the internal `Bolt::Inventory::Target` object
for all of its logic. This decouples the way we expose targets to plans
from the way that we store their state internally.

This update to the way we store target data simplifies the evaluation
patterns for the references in a group. Specifically, we query the
target-level data in each group exactly once, when we initially reify
the target (add it to the inventory via `get_targets`). After that, the
data is no longer needed as it is stored in its own space on the target.
We query the group-level data only when the target is first reified or
when its list of groups changes.

### DIFF
--- a/lib/bolt/inventory/group2.rb
+++ b/lib/bolt/inventory/group2.rb
@@ -2,153 +2,173 @@
 
 require 'bolt/inventory/group'
 require 'bolt/inventory/inventory2'
+require 'bolt/inventory/target'
 
 module Bolt
   class Inventory
     class Group2
-      attr_accessor :name, :targets, :aliases, :name_or_alias, :groups,
-                    :config, :facts, :vars, :features, :plugin_hooks
+      attr_accessor :name, :groups
 
       # THESE are duplicates with the old groups for now.
       # Regex used to validate group names and target aliases.
       NAME_REGEX = /\A[a-z0-9_][a-z0-9_-]*\Z/.freeze
 
-      DATA_KEYS = %w[name config facts vars features plugin_hooks].freeze
-      TARGET_KEYS = DATA_KEYS + %w[alias uri]
-      GROUP_KEYS = DATA_KEYS + %w[groups targets]
+      DATA_KEYS = %w[config facts vars features plugin_hooks].freeze
+      TARGET_KEYS = DATA_KEYS + %w[name alias uri]
+      GROUP_KEYS = DATA_KEYS + %w[name groups targets]
       CONFIG_KEYS = Bolt::TRANSPORTS.keys.map(&:to_s) + ['transport']
 
-      def initialize(data, plugins)
+      def initialize(input, plugins)
         @logger = Logging.logger[self]
-        raise ValidationError.new("Expected group to be a Hash, not #{data.class}", nil) unless data.is_a?(Hash)
-        raise ValidationError.new("Cannot set group with plugin", nil) if data.key?('_plugin')
-        raise ValidationError.new("Group does not have a name", nil) unless data.key?('name')
         @plugins = plugins
 
-        %w[name vars features facts plugin_hooks].each do |key|
-          validate_config_plugin(data[key], key, nil)
-        end
+        raise ValidationError.new("Expected group to be a Hash, not #{input.class}", nil) unless input.is_a?(Hash)
 
-        @name = data['name']
+        input = resolve_top_level_reference(input) if reference?(input)
+
+        raise ValidationError.new("Group does not have a name", nil) unless input.key?('name')
+
+        @name = resolve_references(input['name'])
+
         raise ValidationError.new("Group name must be a String, not #{@name.inspect}", nil) unless @name.is_a?(String)
         raise ValidationError.new("Invalid group name #{@name}", @name) unless @name =~ NAME_REGEX
 
-        # DEPRECATION : remove this before finalization
-        if data.key?('target-lookups')
-          msg = "'target-lookups' are no longer a separate key. Merge 'target-lookups' and 'targets' lists and replace 'plugin' with '_plugin'" # rubocop:disable Metrics/LineLength
-          raise ValidationError.new(msg, @name)
+        validate_group_input(input)
+
+        @input = input
+
+        validate_data_keys(@input)
+
+        targets = input.fetch('targets', [])
+        if reference?(targets)
+          targets = resolve_top_level_reference(targets)
+        elsif targets.is_a?(Array)
+          targets = targets.map { |target| resolve_top_level_reference(target) }
+        else
+          raise ValidationError.new("Targets list must be an Array, not #{targets.class}", @name)
         end
 
-        unless (unexpected_keys = data.keys - GROUP_KEYS).empty?
-          msg = "Found unexpected key(s) #{unexpected_keys.join(', ')} in group #{@name}"
-          @logger.warn(msg)
-        end
-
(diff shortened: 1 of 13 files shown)


lib/bolt/inventory/group2.rb
```diff
@@ -2,166 +2,183 @@
...
 
-        # This check prevents plugins from returning plugins
-        raise ValidationError.new("Cannot set target with plugin", @name) if target.key?('_plugin')
-        target.each do |k, v|
-          next if k == 'config'
-          validate_config_plugin(v, k, @name)
-        end
+        target['name'] = resolve_references(target['name']) if reference?(target['name'])
+        target['uri'] = resolve_references(target['uri']) if reference?(target['uri'])
+        target['alias'] = resolve_references(target['alias']) if reference?(target['alias'])
```
adreyer: couldn't this contain a reference?

`['mynode', {_plugin: myalias}]`

lib/bolt/inventory/group2.rb
```diff
@@ -253,37 +247,30 @@ def data_merge(data1, data2)
...
+        @string_targets.each do |string_target|
+          # If this is the name of a target defined elsewhere, then insert the
+          # target into this group as just a name. Otherwise, add a new target
+          # with the string as the URI.
+          if known_targets.include?(string_target)
+            @unresolved_targets[string_target] = { 'name' => string_target }
+          # If this is an alias for an existing target, then add it to this group
+          elsif (canonical_name = aliases[string_target])
+            if local_targets.include?(canonical_name)
+              @logger.warn("Ignoring duplicate target in #{@name}: #{canonical_name}")
```
adreyer: Do we need this warning message? If there is no data on the target nothing meaningful is being ignored right?

lib/bolt/inventory/group2.rb
```diff
@@ -355,81 +363,92 @@ def validate(used_names = Set.new, target_names = Set.new, aliased = {}, depth =
...
       end
 
-      # The data functions below expect and return nil or a hash of the schema
-      # {'config' => Hash, 'vars' => Hash, 'facts' => Hash, 'features' => Array,
-      #  'plugin_hooks' => Hash, 'groups' => Array}
-      def data_for(target_name)
-        data_merge(group_collect(target_name), target_collect(target_name))
+      def resolve_data_keys(data, target = nil)
+        result = {
+          'config' => resolve_references(data.fetch('config', {})),
```
adreyer: Is this functionally different from `result = resolve_references(data)` at or `empty_data.merge(data)` this point?

lib/bolt/inventory/group2.rb
```diff
@@ -253,37 +247,30 @@ def data_merge(data1, data2)
...
+          # with the string as the URI.
+          if known_targets.include?(string_target)
+            @unresolved_targets[string_target] = { 'name' => string_target }
+          # If this is an alias for an existing target, then add it to this group
+          elsif (canonical_name = aliases[string_target])
+            if local_targets.include?(canonical_name)
+              @logger.warn("Ignoring duplicate target in #{@name}: #{canonical_name}")
             else
-              @targets[target_name] = { 'name' => target_name }
+              @unresolved_targets[canonical_name] = { 'name' => canonical_name }
```
adreyer: Can this go directly into `@resolved_targets`?

lib/bolt/inventory/group2.rb
nicklewis: You're right, it could. I think we can actually just remove the `if reference?` checks now since `resolve_references` has that check embedded.

lib/bolt/inventory/group2.rb
nicklewis: It was already there, so I left it. I don't think it's especially helpful though.

lib/bolt/inventory/group2.rb
nicklewis: We pass this `@input` on the group, which is the entire initial hash including the _unresolved_ data from `targets` and `groups`. If we just did `resolve_references(@data)`, that would resolve the targets and groups again.

It could make sense to do something like `@unresolved_data = input.select { ... }` with just the `DATA_KEYS`, instead of having `@input`. That doesn't extend well to targets though, since we don't have an object to store data separate from name/uri/alias...

lib/bolt/inventory/group2.rb
nicklewis: I think the rationale was that putting it in `@unresolved_targets` ensures it will have all the data keys added to it. That might not matter though, may have just been to quiet a failing test. I'll check it out.

nicklewis: Ah I see what the issue was. `data_merge` doesn't gracefully handle missing keys, so it needs to either be added to `@unresolved_targets` or added to `@resolved_targets` with all the missing keys populated.

lib/bolt/inventory/group2.rb
nicklewis: I updated these to check for the presence of the key rather than whether it's a reference.

donoghuc:
Enabling lookups under `plugins` in `bolt.yaml` will be a separate PR/commit? 

I think I remember discussing on slack that resolving plugins from Targets dynamically created in plans (something like:
```
plan test(){
  $conf = {'ssh' => {'user' => {'_plugin' => 'prompt', 'message' => 'user plz'}}}
  $x = Target.new('uri' => 'qw7a3o43k94ga0h.delivery.puppetlabs.net', 'config' => $conf)
  run_command('whoami', $x)
}
```
is out of scope for this ticket (and may not ever be supported)?

Issue #1288, opened 2019-10-07
Optional values in plans cause warning when reference in an apply() block


nmaludy:
## Describe the Bug

When referencing a `Optional[]` parameter from a plan in an `apply()` block a warning is emitted:

```
Unknown variable: 'patching_dir'. (file: /home/nmaludy/bolt/modules/patching/plans/deploy_scripts.pp, line: 19, column: 28)
Unknown variable: 'bin_dir'. (file: /home/nmaludy/bolt/modules/patching/plans/deploy_scripts.pp, line: 20, column: 28)
Unknown variable: 'owner'. (file: /home/nmaludy/bolt/modules/patching/plans/deploy_scripts.pp, line: 21, column: 28)
Unknown variable: 'group'. (file: /home/nmaludy/bolt/modules/patching/plans/deploy_scripts.pp, line: 22, column: 28)
Unknown variable: 'mode'. (file: /home/nmaludy/bolt/modules/patching/plans/deploy_scripts.pp, line: 23, column: 28)
```

## Expected Behavior
These variables should be passed to the Puppet scope with a value of `undef`

## Steps to Reproduce
Execute the following plan: 

https://github.com/EncoreTechnologies/puppet-patching/blob/master/plans/deploy_scripts.pp

```puppet
# @example CLI deploy a pre patching script
#   bolt plan run patching::deploy_scripts scripts='{"pre_patch.sh": {"source": "puppet:///modules/test/patching/pre_patch.sh"}}'
#
# @example CLI deploy a pre and post patching script
#   bolt plan run patching::deploy_scripts scripts='{"pre_patch.sh": {"source": "puppet:///modules/test/patching/pre_patch.sh"}, "post_patch.sh": {"source": "puppet:///modules/test/patching/post_patch.sh"}}'
plan patching::deploy_scripts(
  TargetSpec $nodes,
  Hash $scripts,
  Optional[String] $patching_dir = undef,
  Optional[String] $bin_dir      = undef,
  Optional[String] $log_dir      = undef,
  Optional[String] $owner        = undef,
  Optional[String] $group        = undef,
  Optional[String] $mode         = undef,
) {
  $targets = run_plan('patching::get_targets', nodes => $nodes)
  return apply($targets) {
    include patching::params
    class { 'patching':
      scripts      => $scripts,
      patching_dir => pick($patching_dir, $patching::params::patching_dir),
      bin_dir      => pick($bin_dir, $patching::params::bin_dir),
[...]

PR #1342, opened 2019-10-22
(GH-1288) Pass parameter references explicitly set to undef to apply

donoghuc:
Previously plan parameters that were explicitly set to `undef` were not passed to apply scope which resulted in warnings about undefined variable references. This commit passes variables explicitly set to `undef` to apply to avoid the errors.

Closes https://github.com/puppetlabs/bolt/issues/1288

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,12 @@
 
   When there is no `config` section in `bolt_plugin.json` configuration options in `bolt.yaml` are validated against the intersection of the parameters schema specified for each task implementation of the plugin's hooks and the values are passed to the task at run time merged with options set in `inventory.yaml`.
 
+### Bug fixes
+
+* **Optional plan parameters referenced in `apply` blocks issued warning** ([#1288](https://github.com/puppetlabs/bolt/issues/1288))
+
+    Previously plan parameters that were explicitly set to `undef` (optional parameters) that were referenced in an `apply` block resulted in a warning message when applying puppet code. The warning is no longer issued when optional parameters are referenced.
+
 ## 1.34.0
 
 ### New features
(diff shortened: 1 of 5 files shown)


lib/bolt/applicator.rb
```diff
@@ -157,6 +156,10 @@ def apply(args, apply_body, scope)
       plan_vars = scope.to_hash
       %w[trusted server_facts facts].each { |k| plan_vars.delete(k) }
 
+      # Get variables explicitly set to undef from plan scope
+      undef_symbols = scope.effective_symtable(false).instance_variable_get(:@symbols).select { |_sym, val| val.nil? }
```
nicklewis: This doesn't seem to work if the `apply()` is inside another scope, since it only uses the most local symtable.

```puppet
plan test(
  TargetSpec $nodes,
  Optional[String] $myvar = undef
) {
  1.each |$x| {
    $nodes.apply_prep
    apply($nodes) {
      notice("My var is ${myvar}")
    }
  }
}
```

donoghuc: ahh, do you think `effective_symtables` is the way to go? I think I can collect `@symbols` from by following the `@parent` attributes for that. Though I wonder what the impact of modifying the `to_hash` method to not exclude `undef` variable references would be.

CHANGELOG.md
```diff
@@ -12,6 +12,12 @@
 
   When there is no `config` section in `bolt_plugin.json` configuration options in `bolt.yaml` are validated against the intersection of the parameters schema specified for each task implementation of the plugin's hooks and the values are passed to the task at run time merged with options set in `inventory.yaml`.
 
+### Bug fixes
+
+* **Optional plan parameters referenced in `apply` blocks issued warning** ([#1288](https://github.com/puppetlabs/bolt/issues/1288))
+
+    Previously plan parameters that were explicitly set to `undef` (optional parameters) that were referenced in an `apply` block resulted in a warning message when applying puppet code. The warning is no longer issued when optional parameters are referenced.
```
```suggestion
    Previously plan parameters that were explicitly set to `undef` (optional parameters) that were referenced in an `apply` block resulted in a warning message when applying puppet code. The warning is no longer issued when explicitly `undef` parameters are referenced.
```

PR #1344, opened 2019-10-23
Loosening of minitar dependency

cardil:
Puppet has moved to using minitar `~> 0.9`, but pdk and bolt still have minitar as `~> 0.6`. That's a problem as users can't use them all in module testing.

Loosening constraint to accept any version before 1.0


cardil:
Closing this as for the same as: https://github.com/puppetlabs/pdk/pull/787

PR #1346, opened 2019-10-24
(maint) Install modules from bolt's Puppetfile for acceptance tests

donoghuc:
Previously for component acceptance tests once bolt is cloned from git and a gem is built, a puppetfile was created in the default moduledir containing a reference to module versions independed from the modules shipped with bolt. We had to remember to manually update both the bolt Puppetfile as well as the acceptance test setup script when updating module references. With this commit, we use the Puppetfile from the bolt repo as the sole source of modules to be used in component acceptance tests.


lucywyman:
@donoghuc Looks like this needs a rebase?

donoghuc:
I'm working on re-doing this based on alex's comment.

donoghuc:
Closing to build in https://github.com/puppetlabs/bolt/pull/1354

PR #1348, opened 2019-10-24
(GH-1200) Add --detail option to show resolved inventory information

### LINKED ISSUE #1200: It's hard to debug dynamic inventory.
## Use Case
It can be hard to debug dynamic inventory files. For example it's easy for a user to accidentally refer to dynamic information outside of a plugin block and other than writing a specific plan there is no way to see what is going on.

A command that resolves dynamic information in the inventory file and prints the results.
It probably should support both resolving the entire file or the data for a specific target. Should it show information only for a specific group or unmerged info for a target?

beechtom:
This commit adds support for a `--detail` flag that shows resolved
inventory information when using the `inventory show` subcommand.

Closes #1200

### DIFF
--- a/lib/bolt/cli.rb
+++ b/lib/bolt/cli.rb
@@ -307,7 +307,11 @@ module Bolt
             list_plans
           end
         elsif options[:subcommand] == 'inventory'
-          list_targets
+          if options[:detail]
+            show_targets
+          else
+            list_targets
+          end
         elsif options[:subcommand] == 'group'
           list_groups
         end
@@ -413,7 +417,12 @@ module Bolt
 
     def list_targets
       update_targets(options)
-      outputter.print_targets(options)
+      outputter.print_targets(options[:targets])
+    end
+
+    def show_targets
+      update_targets(options)
+      outputter.print_target_info(options[:targets])
     end
 
     def list_groups
--- a/lib/bolt/outputter/human.rb
+++ b/lib/bolt/outputter/human.rb
@@ -311,11 +311,20 @@ module Bolt
         end
       end
 
-      def print_targets(options)
-        targets = options[:targets].map(&:name)
+      def print_targets(targets)
         count = "#{targets.count} target#{'s' unless targets.count == 1}"
-        @stream.puts targets.join("\n")
+        @stream.puts targets.map(&:name).join("\n")
         @stream.puts colorize(:green, count)
+      end
+
+      def print_target_info(targets)
+        data = targets.map(&:detail).map do |target|
+          target.transform_values do |value|
+            value.nil? ? "nil" : value
+          end
+        end
+
+        @stream.puts data.to_yaml
       end
 
       def print_groups(groups)
--- a/lib/bolt/outputter/json.rb
+++ b/lib/bolt/outputter/json.rb
@@ -89,11 +89,18 @@ module Bolt
                        "moduledir": moduledir }.to_json)
       end
 
-      def print_targets(options)
-        targets = options[:targets].map(&:name)
-        count = targets.count
-        @stream.puts({ "targets": targets,
-                       "count": count }.to_json)
+      def print_targets(targets)
+        @stream.puts ::JSON.pretty_generate(
+          "targets": targets.map(&:name),
+          "count": targets.count
+        )
+      end
+
+      def print_target_info(targets)
+        @stream.puts ::JSON.pretty_generate(
+          "targets": targets.map(&:detail),
+          "count": targets.count
+        )
       end
 
       def print_groups(groups)
--- a/lib/bolt/target.rb
+++ b/lib/bolt/target.rb
@@ -80,6 +80,22 @@ module Bolt
       )
     end
 
+    def detail
+      {
+        'name' => name,
+        'uri' => uri,
+        'alias' => @target_alias,
+        'config' => config.merge(
+          'transport' => transport,
+          transport => options
+        ),
+        'vars' => vars,
+        'features' => features,
+        'facts' => facts,
+        'plugin_hooks' => Bolt::Util.deep_clone(plugin_hooks)
+      }
+    end
+
     def inventory_target
       @inventory.targets[@name]
     end
@@ -226,6 +242,14 @@ module Bolt
       end
     end
 
+    def vars
+      @inventory.vars(self)
+    end
+
+    def facts
+      @inventory.facts(self)
+    end
+
     # TODO: WHAT does equality mean here?
     # should we just compare names? is there something else that is meaninful?
     def eql?(other)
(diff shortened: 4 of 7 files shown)


lib/bolt/outputter/json.rb
```diff
@@ -89,13 +89,16 @@ def print_puppetfile_result(success, puppetfile, moduledir)
...
-      def print_targets(options)
-        targets = options[:targets].map(&:name)
+      def print_targets(targets)
         count = targets.count
         @stream.puts({ "targets": targets,
                        "count": count }.to_json)
       end
 
+      def print_target_config(targets)
+        @stream.puts targets.to_json
```
adreyer: This format should probably be a superset of the print targets format.

lib/bolt/target.rb
```diff
@@ -80,6 +80,17 @@ def to_h
       )
     end
 
+    def target_config_hash
+      {
```
adreyer: I think we should print the entire data structure for the target rather then the "compacted config"

so something like
```
{ name: foo
  uri: foo@example.com
  config: { transport: ssh, ssh: { }, winrm{ } }.
  vars: { } 
  etc
}
```

beechtom: Looking at the target data structure, it would make sense to include `name`, `uri`, `alias`, `config`, `vars`, `facts`, and `features`. There's also `plugin_hooks`, though I'm not sure if that should be included?

For fields that are emtpy (i.e. a target doesn't have a `uri`) do we want to remove that from the displayed data, or show the user that it's empty/unset?

lucywyman: I think we want to show that it's empty or unset. I think plugin_hooks is also useful for debugging -  I can slightly see how it could be confusing if people think it's plugins that generated the target or something? But that seems unlikely.

lib/bolt/cli.rb
```diff
@@ -413,7 +417,14 @@ def list_plans
 
     def list_targets
       update_targets(options)
-      outputter.print_targets(options)
+      targets = options[:targets].map(&:name)
```
lucywyman: Why did this get moved here? It seems like if we ever want to call `print_targets` elsewhere we'll want it to have the same format.

lib/bolt/cli.rb
```diff
@@ -413,7 +417,14 @@ def list_plans
...
+      outputter.print_targets(targets)
+    end
+
+    def show_targets
+      update_targets(options)
+      targets = inventory.targets.values.map(&:to_h)
```
lucywyman: This line also seems like it belongs in the outputter

beechtom: Updated.

lib/bolt/cli.rb
beechtom: Updated.

lib/bolt/outputter/human.rb
```diff
@@ -311,13 +311,16 @@ def print_module_list(module_list)
...
+      def print_targets(targets)
         count = "#{targets.count} target#{'s' unless targets.count == 1}"
-        @stream.puts targets.join("\n")
+        @stream.puts targets.map(&:name).join("\n")
         @stream.puts colorize(:green, count)
       end
 
+      def print_target_info(targets)
+        @stream.puts targets.map(&:detail).to_yaml
+      end
```
lucywyman: The Psych `to_yaml` does a super annoying thing where it tries to create and refence anchors for common data. This means that the output of this, if you have multiple targets, is:
```yaml
- name: hostname1
  uri: hostname1
  alias: 
  config:
    transport: ssh
    ssh:
      connect-timeout: 10
      tty: false
      load-config: true
      disconnect-timeout: 5
      host-key-check: false
  vars: {}
  features: []
  facts: {}
  plugin_hooks:
    puppet_library: &1
      plugin: install_agent
- name: hostname2
  uri: hostname2
  alias: 
  config:
    transport: ssh
    ssh:
      connect-timeout: 10
      tty: false
      load-config: true
      disconnect-timeout: 5
      host-key-check: false
  vars: {}
  features: []
  facts: {}
  plugin_hooks:
    puppet_library: *1
- name: hostname3
  uri: hostname3
  alias: 
  config:
    transport: ssh
    ssh:
      connect-timeout: 10
      tty: false
      load-config: true
      disconnect-timeout: 5
      host-key-check: false
  vars: {}
  features: []
  facts: {}
  plugin_hooks:
    puppet_library: *1
```

If you directly copy paste this into `inventory.yaml`, you get:
```
NoMethodError: undefined method `line' for #<Psych::BadAlias: Unknown alias: 1>
```

In looking into Psych a bit I'm not sure there's a way to disable or configure aliasing. Is there a workaround for this?

lucywyman: The second comment here perhaps? https://stackoverflow.com/questions/3981128/ruby-yaml-write-without-aliases

beechtom: Fixed with a deep clone of the plugins hash.

lib/bolt/target.rb
```diff
@@ -80,6 +80,22 @@ def to_h
       )
     end
 
+    def detail
+      {
+        'name' => name,
+        'uri' => uri,
+        'alias' => @target_alias,
```
lucywyman: I think we should actually remove this if it's not set. Right now it's `nil` if it's not set, which results in:
```
...
alias:
config:
  transport: ssh
...
```
Which raises this error: `Alias entry on <hostname> must be a String or Array, not NilClass for group at ["all"]`. If we set it to an empty string if nil, then we get `Invalid alias for group at ["all"]`. We could theoretically set it to the name if it's unset, but that seems unnecessarily verbose. I think the best thing to do is to set `alias` in this hash after creating it if it has a value, otherwise omit it.

beechtom: Updated to replace all `nil` values with a stringified `"nil"`

spec/bolt/cli_spec.rb
```diff
@@ -2100,6 +2100,12 @@ def stub_config(file_content = {})
       cli.execute(cli.parse)
     end
 
+    it 'lists targets with resolved configuration' do
+      cli = Bolt::CLI.new(%w[inventory show -t localhost --detail])
```
lucywyman: Can we also verify that the output of this is what we expect? Last comment and then I think this is GtG!

lib/bolt/target.rb
donoghuc: @target_alias is not an attribute for either Target or Target2. I think that for Target2 we can use the `target_alias` method, and we will need to introduce one for the `Target` object.

lib/bolt/target.rb
```diff
@@ -258,6 +282,22 @@ def to_h
       )
     end
 
+    def detail
+      {
+        'name' => name,
+        'uri' => uri,
```
donoghuc: wont name always be the same as uri for Target? I think we should only print uri

lucywyman: No, you can set name and URI separately:
```
targets:
  - uri: dhix35pjb8hiya9.delivery.puppetlabs.net
    name: myhost
```

`bolt command run hostname -n myhost`

donoghuc: This is for inventory v1 target v1. I kept the URI for v2.

PR #1349, opened 2019-10-24
(maint) Maintain ruby 2.3 compatibility by not using Dir.children

beechtom:
Ruby 2.3 does not support the `Dir#children` method. This commit updates
the `Bolt::Module` class to use `Dir#entries` and filter the `.` and
`..` directories instead.


donoghuc:
Fun. This was introduced a while ago. We just never hit that code path bc this is the first time we have used a module based plugin in acceptance tests. This exact problem happened a while ago. https://github.com/puppetlabs/bolt/pull/1268 short of testing with ruby 2.3 in Travis I'm not sure a great way to detect old ruby problems before acceptance tests.

PR #1352, opened 2019-10-24
(maint) Do not exclude bolt_plugin.json in gemspec

donoghuc:
Now that modules need to be shiped with a bolt_plugin.json, do not exclude that file when building bolt packages.


donoghuc:
Should we also ship `modules/*/manifests/*.pp`?

PR #1353, opened 2019-10-24
(maint) Run rubocop before tests in Travis

beechtom:
This commit updates the build script for Travis to run rubocop before
any tests. This is helpful to catch any failures caused by rubocop early
on before running tests, which can take a significant amount of time.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,8 +39,8 @@ before_script:
 - sudo sh -c "echo 'Defaults authenticate' >> /etc/sudoers"
 - sudo sh -c "echo 'travis  ALL=(ALL) PASSWD:ALL' >> /etc/sudoers"
 script:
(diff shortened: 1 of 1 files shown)


donoghuc:
I'm a big +1 on this. I hate when we wait 15 mins for travis to run just to find out there is a rubocop error that is blocking merge.

lucywyman:
Also +1. FWIW, I have a pre-push hook that runs rubocop so I don't have to worry about it failing in travis:
```
# bolt/.git/hooks/pre-push
!#/bin/sh
bundle exec rake rubocop
```

It does take a bit to run, but is slightly less annoying than having it fail in Travis :woman_shrugging:

lucywyman:
With this change I'm not sure it's actually that useful :smile:

PR #1354, opened 2019-10-25
(packaging) Ship Puppetfile with bolt gem

donoghuc:
With this commit the Puppetfile that defines module versions shipped with bolt system packages is included in the gem artifact. This enables consumers of the gem to be able to install the module versions shipped with the corresponding system package.


lucywyman:
Works locally for me :+1:

Issue #1269, opened 2019-10-01
Standardize CLI flags


adreyer:
## Use Case
having flags like `--password` that takes an optional argument limits how we can accept arguments and the quality of error messages we can give. Many of the flags may no longer make sense now that bolt supports an inventory file. 

## Describe the Solution You Would Like
* Replace flags with optional arguments like `--password` a `--password` flag that requires an argument. The `prompt` plugin should be used when password prompt is needed.
* Consider what CLI flags still make sense in light of inventory improvements. Any flag must either require a parameter or not accept one. 
* Send `prompts` to stderr instead of stdout to avoid breaking the JSON outputter (update the prompt plugin)
* Any breaking changes should have deprecation warnings in the 1.x series
* Note: Cannot gate this with `future` flag in a meaningful way, so instead we need to just provide deprecation warning for the things we intend to get rid of. 

* Add a `--password-prompt` and `--sudo-password-prompt` prompt for data on stderr.
* For now, warn and advise to use `--*-prompt` or prompt plugin, make password required in 2.0.

donoghuc:
Closed by https://github.com/puppetlabs/bolt/pull/1355

PR #1355, opened 2019-10-25
(GH-1269) Require argument for --password and --sudo-password options

beechtom:
This commit removes the optional argument for the `--password` and
`--sudo-password` options and replaces it with a required option.
Previously, if a password was omitted as an argument, the user would be
prompted to enter one.

This also updates the `prompt` plugin to print to `stderr` instead of
`stdout`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,19 @@
 # Changelog
 
+## Bolt NEXT
+
+### Deprecation
+
+* **Replace optional arguments for `--password` and `--sudo-password` options with required arguments** ([#1269](https://github.com/puppetlabs/bolt/issues/1269))
+
+  The `--password` and `--sudo-password` options now require a password as an argument. Previously, if the password was omitted the user would be prompted to enter one. If you wish to be prompted for a password, use the `prompt` plugin.
+
+### New features
+
+* **`prompt` messages print to `stderr`** ([#1269](https://github.com/puppetlabs/bolt/issues/1269))
+
+  The prompt plugin now prints messages to `stderr` instead of `stdout`.
+
 ## 1.35.0
 
 ### Deprecation
--- a/lib/bolt/bolt_option_parser.rb
+++ b/lib/bolt/bolt_option_parser.rb
@@ -363,14 +363,14 @@ module Bolt
         @options[:user] = user
       end
       define('-p', '--password [PASSWORD]',
-             'Password to authenticate with. Omit the value to prompt for the password.') do |password|
-        if password.nil?
-          STDOUT.print "Please enter your password: "
-          @options[:password] = STDIN.noecho(&:gets).chomp
-          STDOUT.puts
-        else
-          @options[:password] = password
+             'Password to authenticate with') do |password|
+        # TODO: Remove deprecation message
+        unless password
+          msg = "Optional parameter for --password is deprecated and no longer prompts for password. " \
+                "Use the prompt plugin instead to prompt for passwords."
+          raise Bolt::CLIError, msg
         end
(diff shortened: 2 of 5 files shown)


beechtom:
Optparse raises the following error when parsing a command missing a required param

```
Option '--password' needs a parameter
```

Because optparse checks for params before executing the option's block, we would either have to catch that specific type of error or parse the command before optparse.

lib/bolt/bolt_option_parser.rb
```diff
@@ -389,15 +383,9 @@ def initialize(options)
       define('--run-as USER', 'User to run as using privilege escalation') do |user|
         @options[:'run-as'] = user
       end
-      define('--sudo-password [PASSWORD]',
```
donoghuc: I think you can just keep this "optional" with 
```
      define('-p', '--password [PASSWORD]',
             'Password to authenticate with') do |password|
          if password.nil?
            raise Bolt::CLIError,
                  "DEPRECATION MESSAGE: NO MORE PASSWORD PROMt"
          else
            @options[:password] = password
          end
      end
```
and then raise cli error with deprecation message.

lib/bolt/plugin/prompt.rb
```diff
@@ -19,9 +19,9 @@ def validate_resolve_reference(opts)
       end
 
       def resolve_reference(opts)
-        STDOUT.print "#{opts['message']}:"
+        STDERR.print "#{opts['message']}:"
```
donoghuc: Should we only print to STDERR when `future` is set?

beechtom: Do we have access to the config from plugins?

donoghuc: From the ruby ones for sure.

lib/bolt/bolt_option_parser.rb
beechtom: Updated.

lib/bolt/plugin/prompt.rb
beechtom: Updated.

spec/bolt/cli_spec.rb
```diff
@@ -380,14 +380,6 @@ def stub_config(file_content = {})
         cli = Bolt::CLI.new(%w[command run uptime --password opensesame --nodes foo])
         expect(cli.parse).to include(password: 'opensesame')
       end
-
-      it "prompts the user for password if not specified" do
```
donoghuc: Maybe we could update these to use $future and then check on stderr? that way when we deprecate stdout we can not have to delete tests.

beechtom: These tests are for the optional parameters in the CLI, so there's no checking on stderr. Since those changes aren't gated by future I'm not sure there's a need to do this?

donoghuc: Oh, my bad :man_facepalming: I thought they were for the prompt plugin. Maybe we can do that there?

CHANGELOG.md
```diff
@@ -4,12 +4,20 @@
 
 ### Deprecation
 
+* **Replace optional arguments for `--password` and `--sudo-password` options with required arguments** ([#1269](https://github.com/puppetlabs/bolt/issues/1269))
```
donoghuc: I think these will need to go in to a new section called `## BOLT NEXT` instead of under 1.35.0. Other than that LGTM

beechtom: Updated. Whoops.

beechtom: Whoops. Updated!

Issue #1345, opened 2019-10-24
Acceptance tests should use bolt's internal puppetfile


adreyer:
It looks like the acceptance tests preempt bolt's bundled content by managing modules separately acceptance/setup/common/pre-suite/071_install_modules.rb. This is both unrealistic and requires us to maintain two sources of truth. Instead of installing these modules as a presuite each install method should manage these modules if needed from bolts internal Puppetfile.

donoghuc:
Oh, oops I thought you wanted that as just a maint commit. https://github.com/puppetlabs/bolt/pull/1346 I raised that PR before I saw this. Can fix it up with ref to this issue.

PR #1356, opened 2019-10-25
(GH-1345) Install bundled modules on gem path in acceptance test

donoghuc:
Previously we maintained a subset of modules needed for acceptance tests by statically defining module versions in acceptance test setup. This is prone to error because when modules are updated in the Puppetfile, there is a risk the acceptance test will not be updated. Also, the statically defined modules were installed into the default boltdir modulepath, not the modules dir in the gem path. Now that we ship the Puppetfile with the bolt gem, we can use it to simply install the modules in the expected bolt gem path with r10k to mirror modules shipped in system packages.

### DIFF
--- a/Puppetfile
+++ b/Puppetfile
@@ -22,7 +22,7 @@ mod 'puppetlabs-yumrepo_core', '1.0.3'
 mod 'puppetlabs-zone_core', '1.0.2'
 
 # Useful additional modules
-mod 'puppetlabs-package', '0.6.0'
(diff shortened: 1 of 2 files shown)


donoghuc:
Do not merge, failing osx bolt controller.

donoghuc:
Fixed up OSX. This should be G2G https://jenkins-master-prod-1.delivery.puppetlabs.net/view/bolt/job/platform_bolt_vanagon-integration-component_master_project-master/781/

PR #1357, opened 2019-10-25
(maint) Allow empty hiera config, plus docs fixes

lucywyman:
This includes a few small improvements:
1. When using the [Boltdir
repository](https://github.com/puppetlabs/Boltdir) we provide an empty
hiera config. If left in place Bolt will try to load this which results
in a traceback when we try to access data from the file. This PR adds a
check which will ignore the hiera config if there's no data in it.
2. When using `--configfile` as a command line argument the parent
directory of the specified file is used as the Boltdir to load other
Bolt config, such as the inventory. This PR adds documentation to the
flag letting users know we do this.
3. Lastly, we had some plain URLs in the `using_plugins.md` file, which
in github markdown will render as links but in Dita don't get rendered
as links. This makes them have explicit links syntax, so they render as
links on the website.

### DIFF
--- a/lib/bolt/applicator.rb
+++ b/lib/bolt/applicator.rb
@@ -132,7 +132,9 @@ module Bolt
     def validate_hiera_config(hiera_config)
       if File.exist?(File.path(hiera_config))
         data = File.open(File.path(hiera_config), "r:UTF-8") { |f| YAML.safe_load(f.read, [Symbol]) }
-        unless data['version'] == 5
+        if data.nil?
+          return nil
+        elsif data['version'] != 5
           raise Bolt::ParseError, "Hiera v5 is required, found v#{data['version'] || 3} in #{hiera_config}"
         end
         hiera_config
--- a/lib/bolt/bolt_option_parser.rb
+++ b/lib/bolt/bolt_option_parser.rb
@@ -422,7 +422,8 @@ module Bolt
         @options[:boltdir] = path
       end
       define('--configfile FILEPATH',
-             'Specify where to load config from (default: ~/.puppetlabs/bolt/bolt.yaml)') do |path|
+             'Specify where to load config from (default: ~/.puppetlabs/bolt/bolt.yaml). ' \
+             'Directory containing bolt.yaml will be used as the Boltdir.') do |path|
         @options[:configfile] = path
       end
       define('-i', '--inventoryfile FILEPATH',
--- a/spec/integration/apply_spec.rb
+++ b/spec/integration/apply_spec.rb
@@ -15,6 +15,7 @@ describe "apply" do
   include BoltSpec::Run
 
   let(:modulepath) { File.join(__dir__, '../fixtures/apply') }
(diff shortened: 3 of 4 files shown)


lib/bolt/bolt_option_parser.rb
```diff
@@ -422,7 +422,8 @@ def initialize(options)
         @options[:boltdir] = path
       end
       define('--configfile FILEPATH',
-             'Specify where to load config from (default: ~/.puppetlabs/bolt/bolt.yaml)') do |path|
+             'Specify where to load config from (default: ~/.puppetlabs/bolt/bolt.yaml). ' \
+             'Parent directory will be used as the Boltdir.') do |path|
```
donoghuc: I'm not sure I understand what `parent directory` means here? For example if `bolt.yaml` is in `/tmp/bolt.yaml`, to me I would think the `parent` directory would be `/`? If you were to use `--configfile /tmp/bolt.yaml` you would expect the boltdir to be `/tmp`, not `/`.

lucywyman: "Containing directory"? "File's directory"?

donoghuc:
```suggestion
Directory containing bolt.yaml will be used as the Boltdir.
```

lib/bolt/applicator.rb
```diff
@@ -132,6 +132,7 @@ def compile(target, ast, plan_vars)
     def validate_hiera_config(hiera_config)
       if File.exist?(File.path(hiera_config))
         data = File.open(File.path(hiera_config), "r:UTF-8") { |f| YAML.safe_load(f.read, [Symbol]) }
+        raise Bolt::ParseError, "No hiera data found in #{hiera_config}" unless data
```
donoghuc: We allow an empty `bolt.yaml` and `inventory.yaml`. I see the `hiera.yaml` as a similar config file which I think we should allow to be empty. maybe we should just handle that case gracefully? Also If we decide to keep the error, I think hiera config is separate from hiera data (hiera config tells puppet/bolt how to look up hiera data), so saying `no hiera data found` should probably say `invalid hiera configuration`. But again, I think that empty should be valid.

spec/integration/apply_spec.rb
```diff
@@ -15,6 +15,7 @@
   include BoltSpec::Run
 
   let(:modulepath) { File.join(__dir__, '../fixtures/apply') }
+  let(:hiera_config) { File.join(__dir__, '../fixtures/configs/hiera.yaml') }
```
donoghuc: I think this should be `../fixtures/configs/empty.yaml`. I dont see a hiera.yaml there. Which makes me wonder why it does not error?

donoghuc: OH, nevermind. i see the empty file. I think you can just use `empty.yaml` so we dont have two empty files there. Seems that has been what we have been doing for empty configs other places.

PR #1364, opened 2019-10-29
(GH-1162) Expand configured paths relative to Boltdir

### LINKED ISSUE #1162: Standardize specifying file paths in bolt config

lucywyman:
There are a number of config options for Bolt that accept filepaths.
Previously we expanded the path relative to where Bolt was run in some
cases, and relative to the Boltdir in others. This standardizes all
configurable paths to be expanded relative to the Boltdir if they are
defined in a config file, and relative to the current working directory
if they are specified on the CLI. This change doesn't affect absolute
paths.  This provides consistent and predictable behavior for users. New
path expansions are gated on the `future` option being set to `true`.

Closes #1162

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,12 @@
 
   When a CLI option that can be overridden in inventory is provided and inventory is loaded from a file or from the BOLT_INVENTORY environment variable an issue is warned about which options could be overridden.
 
+### Bug fixes
+
+* **Standardize configured paths to be relative to Boltdir** ([#1162](https://github.com/puppetlabs/bolt/issues/1162))
+
+  Previously we expanded some configured paths relative to the directory Bolt was run from, and other paths were expanded relative to the Boltdir. This standardizes all configured paths, including the modulepath, to be relative to the Boltdir. This only applies to file-based config, not command line flags, which are expanded relative to CWD. This fix is gated on the `future` config option, and will be available by default in Bolt 2.0
+
 ## 1.35.0
 
 ### Deprecation
--- a/documentation/bolt_configuration_options.md
+++ b/documentation/bolt_configuration_options.md
@@ -45,7 +45,8 @@ ssh:
 -   `host-key-check`: Whether to perform host key validation when connecting over SSH. Default is `true`.
 -   `password`: Login password.
 -   `port`: Connection port. Default is `22`.
--   `private-key`: The path to the private key file to use for SSH authentication.
+-   `private-key`: Either the path to the private key file to use for SSH authentication, or a hash
+    with key `key-data` and the contents of the private key.
 -   `proxyjump`: A jump host to proxy SSH connections through, and an optional user to connect with, for example: jump.example.com or user1@jump.example.com.
 -   `run-as`: A different user to run commands as after login.
 -   `run-as-command`: The command to elevate permissions. Bolt appends the user and command strings to the configured run as a command before running it on the target. This command must not require an interactive password prompt, and the `sudo-password` option is ignored when `run-as-command` is specified. The run-as command must be specified as an array.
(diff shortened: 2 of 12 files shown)


documentation/bolt_configuration_options.md
```diff
@@ -54,6 +55,22 @@ ssh:
...
+targets:
+  - name: host1.example.net
+    config:
+      transport: ssh
+      ssh:
+        host-key-check: true
+        port: 22
+        run-as-command: ['sudo', '-k', '-n']
+        private-key:
+          key-data: |
```
lucywyman: I had forgotten about this option and realized it wasn't documented. It's a bit hard to describe in words, so I added an example of that as well as the other non-string option `run-as-command` for clarity.

lucywyman:
This is currently waiting on https://github.com/puppetlabs/puppetlabs-terraform/pull/3 and https://github.com/puppetlabs/puppetlabs-aws_inventory/pull/2 to be merged and released so we can update the Puppetfile here.

CHANGELOG.md
```diff
@@ -14,6 +14,12 @@
 
   The prompt plugin now prints messages to `stderr` instead of `stdout`.
 
+### Bug fixes
```
lucywyman: This might be a feature, though we probably should have just done this to begin with :woman_shrugging:

lib/bolt/cli.rb
```diff
@@ -244,7 +244,7 @@ def handle_parser_errors
 
     def puppetdb_client
       return @puppetdb_client if @puppetdb_client
-      puppetdb_config = Bolt::PuppetDB::Config.load_config(nil, config.puppetdb)
+      puppetdb_config = Bolt::PuppetDB::Config.load_config(nil, config.puppetdb, config.boltdir)
```
lucywyman: We check for `$future` in PuppetDB::Config, so this won't change behavior if you don't have `future` set.

lib/bolt/config.rb
```diff
@@ -173,6 +173,13 @@ def update_from_file(data)
       TRANSPORTS.each do |key, impl|
         if data[key.to_s]
           selected = impl.filter_options(data[key.to_s])
+          if @future
+            to_expand = %w[tmpdir private-key cacert token-file] & selected.keys
```
donoghuc: I dont think tmpdir should be expanded. It is a path on a target, so it should not care about being relative to boltdir.

nicklewis:
I think we should still treat file paths specified on the CLI as relative to cwd.

donoghuc:
I got terraform and aws_inventory releases out. I think once the log file path is implemented this should be g2g.

lucywyman:
Good catch - updated!

lib/bolt/config.rb
```diff
@@ -110,7 +110,7 @@ def normalize_interpreters(interpreters)
     def normalize_log(target)
       return target if target == 'console'
       target = target[5..-1] if target.start_with?('file:')
-      'file:' + File.expand_path(target)
+      'file:' + File.expand_path(target, @boltdir.path)
```
donoghuc: We should probably get this on @future? Is that possible

lucywyman: Updated - also  I think you accidentally pinged a random github user named 'future' :D

lib/bolt/config.rb
```diff
@@ -110,7 +110,11 @@ def normalize_interpreters(interpreters)
     def normalize_log(target)
       return target if target == 'console'
       target = target[5..-1] if target.start_with?('file:')
-      'file:' + File.expand_path(target)
+      if @future
```
donoghuc: I think normalize_log, ends up being called before @future is defined https://github.com/puppetlabs/bolt/pull/1364/files#diff-9637f8192166af0313bc199f0706d20bR140 . So @future will always be nil and this will never be relative to boltdir.

donoghuc: I think evaluating future in the first line of update_from_file is safe

lucywyman: Actually tested it this time :sweat_smile:

donoghuc: yeah, its always the seeminly simplest changes that dont end up working like you expect! Thanks.

Issue #1285, opened 2019-10-04
Bolt Project Scaffolding/Init


highb:
## Use Case
I'm a user trying to get started on adding Bolt to my project. Having used the pdk previously, I want to do something equivalent to `pdk new` in my project directory to add Bolt to it.

## Describe the Solution You Would Like
I type `bolt new_local_project` `bolt new_embedded_project` `bolt new_user_project` or something similar and get the directory structure described in the docs here https://puppet.com/docs/bolt/latest/bolt_project_directories.html#local-project-directory

## Describe Alternatives You've Considered
Making a bunch of directories myself and hopefully not making any typos.

## Solution
We should create a new command `bolt project init` which creates a `bolt.yaml`.

adreyer:
In order to match the other commands `bolt project init <optional-path>` makes sense where path defaults to `./`

Since this is targetted at the new user it should not create a lot of extraneous files just the minimum that would be useful.

highb:
Makes sense to me. 😄

PR #1367, opened 2019-10-29
(GH-1285) Add project init command

beechtom:
This adds a `project init` subcommand that creates a new Bolt project. When
specifying a filepath using `bolt project init [filepath]`, the project
directory will be created at the filepath. Otherwise, the project
directory is created in the current working directory.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,10 @@
 
   The prompt plugin now prints messages to `stderr` instead of `stdout`.
 
+* **New `project init` subcommand** ([#1285](https://github.com/puppetlabs/bolt/issues/1285))
+
+  The CLI now provides a subcommand `bolt project init` that will create a new file `bolt.yaml` in the current working directory, making the directory a [Bolt project directory](https://puppet.com/docs/bolt/latest/bolt_project_directories.html#local-project-directory)
+
 ## 1.35.0
 
 ### Deprecation
--- a/lib/bolt/cli.rb
+++ b/lib/bolt/cli.rb
@@ -37,6 +37,7 @@ module Bolt
                  'secret' => %w[encrypt decrypt createkeys],
                  'inventory' => %w[show],
                  'group' => %w[show],
+                 'project' => %w[init],
                  'apply' => %w[] }.freeze
 
     attr_reader :config, :options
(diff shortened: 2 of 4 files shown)


lib/bolt/cli.rb
```diff
@@ -504,6 +508,14 @@ def generate_types
       0
     end
 
+    def initialize_project
+      path = File.expand_path(options[:object] || Dir.pwd)
+      FileUtils.mkdir_p(path)
+      ok = FileUtils.touch(File.join(path, 'bolt.yaml'))
```
lucywyman: I think we typically only save this to a var if we want to pass `ok` to the outputter. Since we don't need to do that here this can just be `FileUtils.touch(File.join(path, 'bolt.yaml')) ? 0 : 1`.

lib/bolt/cli.rb
```diff
@@ -504,6 +508,12 @@ def generate_types
       0
     end
 
+    def initialize_project
+      path = File.expand_path(options[:object] || Dir.pwd)
+      FileUtils.mkdir_p(path)
+      FileUtils.touch(File.join(path, 'bolt.yaml')) ? 0 : 1
```
lucywyman: Ugh I think my update comment got lost: I think we should print something here, which means using the `ok` pattern you had but adding a `outputter.print_message("Successfully create Bolt project directory at #{path}")` before returning.

CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # Changelog
 
+## Bolt NEXT
+
+### New features
+
+* **New `project init` subcommand** ([#1285](https://github.com/puppetlabs/bolt/issues/1285))
+
+  The CLI now provides a subcommand `bolt project init` that will create a new Bolt project.
```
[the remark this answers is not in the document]
lucywyman: Wait I still think we should add this? Not sure if it got resolved on accident or the change got lost? Or if there's a reason not to add more details?

lucywyman:
I think the changelog still needs an update (unless there's a reason not to?) otherwise lgtm

beechtom:
Updated. I ended up rebasing which overwrote the suggested change you made. All fixed now!

PR #1369, opened 2019-10-30
(GH-1200) Avoide yaml anchors with inventory show --detail

donoghuc:
Previously nested hashes appear to have been converted to yaml anchors when using the --detail flag with inventory show. It looks like a similar problem was encountered with plugin_hooks. This commit updates the `Target` `detail` hash to make deep clones for any value that may be a nested hash.


donoghuc:
Repro:
```
version: 2
groups:
  - name: foo
    targets:
      - bar
      - baz
    facts:
      foo:
        bar: baz
```
```
cas@cas-ThinkPad-T460p:~/working_dir/bolt$ bolt inventory show -t all --detail
---
- name: bar
  uri: bar
  alias: nil
  config:
    transport: ssh
    ssh:
      connect-timeout: 10
      tty: false
      load-config: true
      disconnect-timeout: 5
      host-key-check: false
  vars: {}
  features: []
  facts:
    foo: &1
      bar: baz
  plugin_hooks:
    puppet_library:
      plugin: puppet_agent
      stop_service: true
- name: baz
  uri: baz
  alias: nil
  config:
    transport: ssh
    ssh:
      connect-timeout: 10
      tty: false
      load-config: true
      disconnect-timeout: 5
      host-key-check: false
  vars: {}
  features: []
  facts:
    foo: *1
  plugin_hooks:
    puppet_library:
      plugin: puppet_agent
      stop_service: true
```

donoghuc:
All this cloning does seem a bit over the top. We are sure we dont want to just 
```
@stream.puts ::JSON.pretty_generate(data)
```
comparison:
```
cas@cas-ThinkPad-T460p:~/working_dir/bolt$ bolt inventory show -t all --detail
[
  {
    "name": "bar",
    "uri": "bar",
    "alias": "nil",
    "config": {
      "transport": "ssh",
      "ssh": {
        "connect-timeout": 10,
        "tty": false,
        "load-config": true,
        "disconnect-timeout": 5,
        "host-key-check": false
      }
    },
    "vars": {
    },
    "features": [

    ],
    "facts": {
      "foo": {
        "bar": "baz"
      }
    },
    "plugin_hooks": {
      "puppet_library": {
        "plugin": "puppet_agent",
        "stop_service": true
      }
    }
  },
  {
    "name": "baz",
    "uri": "baz",
    "alias": "nil",
    "config": {
      "transport": "ssh",
      "ssh": {
        "connect-timeout": 10,
        "tty": false,
        "load-config": true,
        "disconnect-timeout": 5,
        "host-key-check": false
      }
    },
    "vars": {
    },
    "features": [

    ],
    "facts": {
      "foo": {
        "bar": "baz"
      }
    },
    "plugin_hooks": {
      "puppet_library": {
        "plugin": "puppet_agent",
        "stop_service": true
      }
    }
  }
]
```
```
cas@cas-ThinkPad-T460p:~/working_dir/bolt$ bolt inventory show -t all --detail
---
- name: bar
  uri: bar
  alias: nil
  config:
    transport: ssh
    ssh:
      connect-timeout: 10
      tty: false
      load-config: true
      disconnect-timeout: 5
      host-key-check: false
  vars: {}
  features: []
  facts:
    foo:
      bar: baz
  plugin_hooks:
    puppet_library:
      plugin: puppet_agent
      stop_service: true
- name: baz
  uri: baz
  alias: nil
  config:
    transport: ssh
    ssh:
      connect-timeout: 10
      tty: false
      load-config: true
      disconnect-timeout: 5
      host-key-check: false
  vars: {}
  features: []
  facts:
    foo:
      bar: baz
  plugin_hooks:
    puppet_library:
      plugin: puppet_agent
      stop_service: true
```

nicklewis:
Would it make more sense to do the deep clone of the object we're rendering rather than of the data we're using to generate that object? That is, call `Bolt::Util.deep_clone(data).to_yaml` in the outputter.

lib/bolt/target.rb
```diff
@@ -291,10 +291,11 @@ def detail
           'transport' => transport,
           transport => options
         },
-        'vars' => vars,
-        'facts' => facts,
+        # Dup frozen hashes to deep_cline in outputter
+        'vars' => vars.dup,
```
nicklewis: I'm not sure I follow why `dup` is necessary here...

donoghuc: deep_merge was throwing "can not modify frozen hash". We are just going with JSON.pretty_generate here anyway, so this will be removed.

nicklewis: That's... puzzling. `deep_merge` shouldn't need to modify anything...

nicklewis: Oh wait it's `deep_clone`. That makes more sense.

lib/bolt/target.rb
```diff
@@ -286,15 +287,16 @@ def detail
       {
         'name' => name,
         'uri' => uri,
-        'alias' => @target_alias,
+        'alias' => target_alias,
         'config' => {
           'transport' => transport,
           transport => options
         },
+        # Dup frozen hashes to deep_cline in outputter
```
donoghuc: TODO: delte comment

lib/bolt/target.rb
```diff
@@ -65,7 +65,7 @@ def safe_name
     end
 
     def target_alias
-      inventory_target.target_alias
```
donoghuc: I thought this was added in the inital GH-1200 implementation, but i guess not. I dont see a method or an instance variable access for `target_alias` on Inventory::Target. So what is it used for?

lib/bolt/target.rb
```diff
@@ -286,15 +287,16 @@ def detail
       {
         'name' => name,
         'uri' => uri,
```
donoghuc: I dont see a reason to print URI for `Target`. we do not expose that in inventory v1

spec/bolt/inventory_spec.rb
```diff
@@ -748,4 +748,37 @@ def common_data(transport)
...
+                   'facts' => { 'foo' => 'bar' } }] }
+    }
+
+    let(:inventory) { Bolt::Inventory.new(data) }
+    let(:target) { get_target(inventory, 'foo') }
+    let(:expected_data) {
+      { 'name' => 'foo',
+        'alias' => %w[bar baz],
+        'config' =>
+                          { 'transport' => 'ssh',
```
donoghuc: This is what rubocop wants for spacing. Is it worth trying to find a way around it???

spec/bolt/inventory_spec.rb
```diff
@@ -748,4 +748,37 @@ def common_data(transport)
...
+                            { 'connect-timeout' => 10,
+                              'tty' => false,
+                              'load-config' => true,
+                              'disconnect-timeout' => 100 } },
+        'vars' => {},
+        'facts' => { 'foo' => 'bar' },
+        'features' => [],
+        'plugin_hooks' =>
+                          { 'puppet_library' => { 'plugin' => 'puppet_agent', 'stop_service' => true } } }
+    }
```
beechtom:
```suggestion
          { 'name' => 'foo',
        'alias' => %w[bar baz],
        'config' => {
          'transport' => 'ssh',
          'ssh' => {
            'connect-timeout' => 10,
            'tty' => false,
            'load-config' => true,
            'disconnect-timeout' => 100
          }
        },
        'vars' => {},
        'facts' => { 'foo' => 'bar' },
        'features' => [],
        'plugin_hooks' => {
          'puppet_library' => {
            'plugin' => 'puppet_agent',
            'stop_service' => true
          }
        } }
    }
```

Probably best to match the formatting for other hashes in the file. This doesn't cause rubocop to complain.

beechtom:
```suggestion
    let(:expected_data) {
      { 'name' => 'foo',
        'alias' => %w[bar baz],
        'config' => {
          'transport' => 'ssh',
          'ssh' => {
            'connect-timeout' => 10,
            'tty' => false,
            'load-config' => true,
            'disconnect-timeout' => 100
          }
        },
        'vars' => {},
        'facts' => { 'foo' => 'bar' },
        'features' => [],
        'plugin_hooks' => {
          'puppet_library' => {
            'plugin' => 'puppet_agent',
            'stop_service' => true
          }
        } }
    }
```

Probably best to match the formatting for other hashes in the file. This doesn't raise any rubocop errors for me.

beechtom: Probably best to have the formatting match the rest of the hashes in the file. This doesn't raise any rubocop errors for me.

```
    let(:expected_data) {
      { 'name' => 'foo',
        'alias' => %w[bar baz],
        'config' => {
          'transport' => 'ssh',
          'ssh' => {
            'connect-timeout' => 10,
            'tty' => false,
            'load-config' => true,
            'disconnect-timeout' => 100
          }
        },
        'vars' => {},
        'facts' => { 'foo' => 'bar' },
        'features' => [],
        'plugin_hooks' => {
          'puppet_library' => {
            'plugin' => 'puppet_agent',
            'stop_service' => true
          }
        } }
    }
```

donoghuc: nice! will update, how did you do that? do you have an automated way of doing that?

beechtom: I just put the opening brace on the same line as the key and the contents on the next line. Seems weird that rubocop has different ways of formatting nested hashes.

lib/bolt/target.rb
donoghuc: @nicklewis Do you think it is OK to write this method like this? https://github.com/puppetlabs/bolt/pull/1340/files#diff-f0928cc9831d370263135e3cdb532fcbR68 It almost looks like maybe we just forgot a attr reader on Inventory::Target? It does not seem to be as simple as just exposing the @target_alias instance variable because it is currently only returning an empty array...

lib/bolt/target.rb
lucywyman: The name and URI can be different. For example if I have:
```
targets:
  - uri: dhix35pjb8hiya9.delivery.puppetlabs.net
    name: myhost
```

and run `bolt inventory show -n myhost --detail` I would want to know what the actual host it's using is. There's no way to find that without the URI.

spec/bolt/inventory_spec.rb
lucywyman: I don't think so - doesn't seem worth the time.

lib/bolt/target.rb
donoghuc: I only exclude URI for inventory v1 Target (v1). It is still included for v2.

lib/bolt/inventory/group2.rb
```diff
@@ -150,6 +150,7 @@ def target_data(target_name)
           resolved_data = resolve_data_keys(target, target_name).merge(
             'name' => target['name'],
             'uri' => target['uri'],
+            'alias' => target['alias'],
```
donoghuc: Do we need alias or target_alias in https://github.com/puppetlabs/bolt/blob/6741379142ab9485f5c7f8c0f9cb4d50e2201408/lib/bolt/inventory/group2.rb#L221-L239 ?

lib/bolt/inventory/target.rb
```diff
@@ -34,7 +34,8 @@ def initialize(target_data, inventory)
         @features = target_data['features'] || Set.new
         @options = target_data['options'] || {}
         @plugin_hooks = target_data['plugin_hooks'] || {}
-        @target_alias = target_data['target_alias'] || []
+
```
donoghuc: target_alias comes from plan, alias from inventory, handle either.

lib/bolt/inventory/group2.rb
beechtom: Looks like adding `alias` to the hash is causing a test to fail at `spec/bolt/inventory/group2_spec.rb:50` since the test isn't expecting the hash to have that value.

donoghuc: Its much worse than that unfortunately. For example try this plan out:
```
plan alias_broke(){
  Target.new('uri' => 'localhost', 'target_alias' => 'foo')
  notice(get_target('foo'))
  run_command('whoami', 'foo')
}
```
I think I have a fix, updating tests now.

lib/bolt/target.rb
donoghuc: Yes, we need to fix target_alias in general.

lib/bolt/inventory/group2.rb
```diff
@@ -41,7 +41,7 @@ def initialize(input, plugins)
 
         @unresolved_targets = {}
         @resolved_targets = {}
-        @targets = Set.new
```
donoghuc: This was not used. The commented out unused instance variable is removed in a different PR.

donoghuc:
@lucywyman The change you requested does not apply to both inventory v1 and v2. Can you take another look?

Issue #1191, opened 2019-08-30
Run apply_prep task with configured run_as


lucywyman:
## Use Case
Users should be able to run the `puppet_agent::install` task as root during apply_prep

## Describe the Solution You Would Like
Should this be a new parameter to the `puppet-library -> plugin` config? A metaparameter to the task?

Proposed in #1123 
```
plugin_hooks:
  puppet_library:
    plugin: task
    task: 'bootstrap'
    _run_as: root
    parameters:
      master: 'puppet.example.com'
      cacert_content: <cert>
```

## Specifics

* If the transport does not support `_run_as` apply_prep should error.

adreyer:
I think we should make `_run_as` a global/meta option for the puppet_library hook. That will align better with it not being an acceptable parameter for other task plugin hooks and making it always available for task based plugins that implement a puppet_library hook.

ie:
```plugin_hooks:
  puppet_library:
    plugin: task
    task: 'bootstrap'
    _run_as: root
    parameters:
      master: 'puppet.example.com'
      cacert_content: <cert>
```

PR #1370, opened 2019-10-31
(maint) Support testing with v2 inventory in more places

adreyer:
I think these are useful improvements but they don't really fit GH-1191

### DIFF
--- a/lib/bolt_spec/run.rb
+++ b/lib/bolt_spec/run.rb
@@ -5,6 +5,7 @@ require 'bolt/config'
 require 'bolt/executor'
 require 'bolt/inventory'
 require 'bolt/pal'
+require 'bolt/plugin'
 require 'bolt/puppetdb'
 require 'bolt/util'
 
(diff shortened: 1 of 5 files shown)


lib/bolt_spec/run.rb
```diff
@@ -130,20 +130,31 @@ class BoltRunner
...
-        @inventory = inventory
+      def initialize(config_data, inventory_data, boltdir_path)
+        @config_data = config_data || {}
+        @inventory_data = inventory_data || {}
+        @boltdir_path = boltdir_path
         @analytics = Bolt::Analytics::NoopClient.new
       end
 
+      def config
+        @config = Bolt::Config.new(Bolt::Boltdir.new(@boltdir_path), @config_data)
```
donoghuc: Does this need to be recreated every time?

lib/bolt_spec/run.rb
```diff
@@ -130,20 +130,31 @@ class BoltRunner
...
+      end
+
+      def inventory
+        @inventory ||= Bolt::Inventory.create_version(@inventory_data, config, plugins)
+      end
+
+      def plugins
+        @plugins ||= Bolt::Plugin.setup(config, pal, puppetdb_client, @analytics)
```
donoghuc: Need to `require 'bolt/plugin` for this constant.

donoghuc:
Seems like `BoltSpec::Run#apply_manifest` was broken (for inventory v1) before this change. I guess it is fairly expensive to integration test that (along with the other run methods). But this appears to work with inventory v1 and v2.

lib/bolt_spec/run.rb
adreyer: no that was a typo.

PR #1374, opened 2019-11-01
(GH-1365) Add resolve_references plan function

### LINKED ISSUE #1365: expose resolve_reference function in plan language
## Use Case
I want to be able to look up values with the resolve_reference hook during plan execution. For example, when target systems change during plan execution (ie `terraform_apply`) I want to be able to use `resolve_reference` to look up new targets. 

expose `resolve_reference` as a plan function. 

## Additional Context
In the case where resolve_reference is used to generate target data it should be up to the plan author to use the Target API to create targets based on the data returned from resolve_reference. (ie no magic adding to inventory).

## Solution
`resolve_references` function which accepts structured data and resolves all the references in that object. Returns the resolved data.

beechtom:
This adds support for a `resolve_references` plan function that accepts
a hash of target data, resolves all references in the hash, and returns
the resolved target data.

Closes #1365

### DIFF
--- a/bolt-modules/boltlib/lib/puppet/functions/resolve_references.rb
+++ b/bolt-modules/boltlib/lib/puppet/functions/resolve_references.rb
@@ -1,0 +1,38 @@
+# frozen_string_literal: true
+
+require 'bolt/error'
+
+# Evaulates all _plugin references in a hash and returns the resolved reference data.
+Puppet::Functions.create_function(:resolve_references) do
+  # Resolve references.
+  # @param references A hash of reference data to resolve.
+  # @return A hash of resolved reference data.
+  # @example Resolve a hash of reference data
+  #   $references = {
+  #     "targets" => [
+  #       "_plugin" => "terraform",
+  #       "dir" => "path/to/terraform/project",
+  #       "resource_type" => "aws_instance.web",
+  #       "uri" => "public_ip"
+  #     ]
+  #   }
+  #
+  #   resolve_references($references)
+  dispatch :resolve_references do
+    param 'Data', :references
+    return_type 'Data'
+  end
+
+  def resolve_references(references)
+    unless Puppet[:tasks]
+      raise Puppet::ParseErrorWithIssue
+        .from_issue_and_stack(
+          Bolt::PAL::Issues::PLAN_OPERATION_NOT_SUPPORTED_WHEN_COMPILING,
+          action: 'resolve_references'
+        )
+    end
+
+    plugins = Puppet.lookup(:bolt_inventory).plugins
+    plugins.resolve_references(references)
+  end
+end
--- a/bolt-modules/boltlib/spec/functions/resolve_references_spec.rb
+++ b/bolt-modules/boltlib/spec/functions/resolve_references_spec.rb
@@ -1,0 +1,61 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+require 'bolt/plugin'
+
+describe 'resolve_references' do
+  include PuppetlabsSpec::Fixtures
+  let(:boltdir) { Bolt::Boltdir.new('./spec/fixtures') }
+  let(:config) { Bolt::Config.new(boltdir, {}) }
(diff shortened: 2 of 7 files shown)


bolt-modules/boltlib/lib/puppet/functions/resolve_references.rb
```diff
@@ -0,0 +1,38 @@
...
+    unless Puppet[:tasks]
+      raise Puppet::ParseErrorWithIssue
+        .from_issue_and_stack(Bolt::PAL::Issues::PLAN_OPERATION_NOT_SUPPORTED_WHEN_COMPILING, action: 'resolve_references')
+    end
+
+    references = references.merge('name' => 'all') unless references['name']
+    inventory = Puppet.lookup(:bolt_inventory)
+    group = Bolt::Inventory::Group2.new(references, inventory.plugins)
+
+    group.resolve_references(references)
```
nicklewis: This should maybe be a method on `Bolt::Plugin` instead of `Group2`. It's a bit strange to have to create a `Group2` instance for this...

lib/bolt/plugin.rb
```diff
@@ -208,6 +208,76 @@ def by_name(plugin_name)
...
+    end
+
+    # Evaluates a single reference. The value returned may be another
+    # reference.
+    def resolve_single_reference(reference)
+      plugin_name = reference['_plugin']
+      begin
+        hook = get_hook(plugin_name, :resolve_reference)
+      rescue Bolt::Plugin::PluginError => e
+        raise ValidationError.new(e.message, @name)
```
nicklewis: This fails because `ValidationError` here refers to `Bolt::ValidationError` from `bolt/error.rb`, where previously it referred to `Bolt::Inventory::ValidationError` from `bolt/inventory.rb`. Also `@name` isn't a thing here.

beechtom: Updated to only pass in `e.message`.

bolt-modules/boltlib/spec/functions/resolve_references_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+describe 'resolve_references' do
+  include PuppetlabsSpec::Fixtures
+  let(:boltdir) { Bolt::Boltdir.new('./spec/fixtures') }
+  let(:config) { Bolt::Config.new(boltdir, {}) }
+  let(:pal) { Bolt::PAL.new(config.modulepath, config.hiera_config, config.boltdir.resource_types) }
+  let(:pdb_client) { Bolt::PuppetDB::Client.new({}) }
+  let(:analytics) { Bolt::Analytics::NoopClient.new }
+  let(:plugins) { Bolt::Plugin.setup(config, pal, pdb_client, analytics) }
+  let(:inventory) { Bolt::Inventory.new({}, plugins: plugins) }
+  let(:tasks_enabled) { true }
```
beechtom: I'm not particularly happy with doing all this setup, but I'm not sure how else to get around it... I tried setting it up similar to (`run_plan_spec`)[https://github.com/puppetlabs/bolt/blob/master/bolt-modules/boltlib/spec/functions/run_plan_spec.rb#L117-L120] but was getting errors about not being able to locate the `Task` plugin or not being able to locate the task `test::references` in `fixtures`.

nicklewis:
This should be rebased rather than merging master into it.

bolt-modules/boltlib/lib/puppet/functions/resolve_references.rb
```diff
@@ -0,0 +1,38 @@
...
+  #       "_plugin" => "terraform",
+  #       "dir" => "path/to/terraform/project",
+  #       "resource_type" => "aws_instance.web",
+  #       "uri" => "public_ip"
+  #     ]
+  #   }
+  #
+  #   resolve_references($references)
+  dispatch :resolve_references do
+    param 'Hash[String[1], Any]', :references
```
nicklewis: I think the param and return types should both be `Data`. `Data` means the subset of Puppet data types which are legal JSON (and therefore what we can pass to and receive from a task/plugin).

beechtom: Updated!

Issue #1375, opened 2019-11-05
Deprecate `--nodes` from cli


donoghuc:
## Use Case
Favor `--targets` over `--nodes`. 

## Describe the Solution You Would Like
Just a deprecation warning for now. 

## Additional Context
Will consider deprecating `$nodes` in plans later.

Issue #1376, opened 2019-11-05
Update bundled content to use $targets instead of $nodes


lucywyman:
## Describe the Change You Would Like
Update bundled content to use $targets instead of $nodes
Go through docs + hands-on lab + other examples and change $nodes to $targets

beechtom:
Blocked by #1175

PR #1380, opened 2019-11-05
(GH-1375) Add deprecation warning for `--nodes`

beechtom:
This adds a deprecation warning for `--nodes` to let users know the CLI
option has been deprecated in favor of `--targets`.

Closes https://github.com/puppetlabs/bolt/issues/1375

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@
 
   The `--password` and `--sudo-password` options now require a password as an argument. Previously, if the password was omitted the user would be prompted to enter one. To continue to be prompted for a password, use the `prompt` plugin.
 
+* **Favor `--targets` over `--nodes`** ([#1375](https://github.com/puppetlabs/bolt/issues/1375))
+
+  The `--nodes` command line option has been deprecated in favor of `--targets`. When using `--nodes`, a deprecation warning will be displayed.
+
 ### New features
 
 * **`prompt` messages print to `stderr`** ([#1269](https://github.com/puppetlabs/bolt/issues/1269))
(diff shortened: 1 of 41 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # Changelog
 
+## Bolt NEXT
+
```
donoghuc: Given we had merged the 1.36.0 changelog finalization last week before we found the ship stopper bug, Bolt Next *is* 1.36.0. You can either add it under that section, or change 1.36.0 to Bolt Next.

spec/integration/logging_spec.rb
```diff
@@ -21,7 +21,7 @@
   let(:lines) { @log_output.readlines }
 
   let(:config_flags) {
-    %W[--nodes #{uri} --no-host-key-check --format json --modulepath #{modulepath} --password #{password}]
+    %W[--targets #{uri} --no-host-key-check --format json --modulepath #{modulepath} --password #{password}]
```
donoghuc: I wonder if updating all spec tests to use this flag now would be part of this ticket? I maybe it should be part of https://github.com/puppetlabs/bolt/issues/1376 or its own separate thing. Either way I think we will need to do it eventually.

spec/bolt/outputter/human_spec.rb
```diff
@@ -91,6 +91,7 @@
     TABLE
   end
 
+  # TODO: Update when --nodes is deprecated
```
donoghuc: I think we should update the outputter now with 
```
USAGE:
bolt task run --targets <target-name> 
```

beechtom: Updated!

Issue #1350, opened 2019-10-24
Use resolve_reference in bolt.yaml


donoghuc:
## Use Case
I want to use plugins to look up values in bolt.yaml. 

## Describe the Solution You Would Like
In `bolt.yaml` plugins can be used under `plugin_hooks` and `plugins` (disallow anywhere else). 

- When bolt.yaml is evaluated all references are resolved.
- If bolt.yaml config for one plugin references another that should happen before the first plugin is initialized. A helpful error should be raised if there is a dependency loop.

## Additional Context
This work was split out from https://github.com/puppetlabs/bolt/issues/1147 , see that ticket for more details.

PR #1381, opened 2019-11-05
(GH-1350) Configure plugins using other plugins

nicklewis:
Previously, the `plugins` section of `bolt.yaml` was required to be 
 completely static. Now, plugin can use the same `_plugin` hashes as in
 `inventory.yaml` to define their configuration. For example, this allows 
 secrets for plugins to be encrypted or read from vault.

 We now instantiate (and thus validate) all plugins that are configured in
 `bolt.yaml` at the time the plugin system is initialized.

 We also now support `_plugin` lookups under the `plugin_hooks` key for
 similar reasons.
Closes GH-1350

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,11 +17,17 @@
   Bolt now accepts the `_run_as` metaparameter for puppet_library hooks. `_run_as` specifies which user the library install task will be executed as.
 
 * **Added `--password-prompt` and `--sudo-password-prompt` to CLI flags** ([#1269](https://github.com/puppetlabs/bolt/issues/1269))
+
   Two new flags have been added to support users who would like to set a `password` or `sudo-password` from a prompt without using a plugin. A deprecation message will appear when a value is not supplied for `--password` or `--sudo-password`.
 
 * **Subcommand `project migrate` new to the CLI** ([#1377](https://github.com/puppetlabs/bolt/issues/1377))
 
   The CLI now provides the subcommand `project migrate` which migrates Bolt projects to the latest version. When migrating a project the [inventory file](https://puppet.com/docs/bolt/latest/inventory_file.html) will be changed from `v1` to `v2`. Changes are made in place and will not preserve comments or formatting.
+
+* **Plugin support in `bolt.yml`** ([#1381](https://github.com/puppetlabs/bolt/pull/1381))
+
+  Plugin configuration can now be set by looking up data from other plugins. For example, the password for one plugin can be queried from another plugin.
+
 
 ## Bug fixes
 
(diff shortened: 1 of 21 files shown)


nicklewis:
This currently depends on the `resolve_references` refactor from #1374. I'll rebase once that's merged.

lib/bolt/cli.rb
```diff
@@ -507,6 +507,7 @@ def run_plan(plan_name, plan_arguments, nodes, options)
     end
 
     def apply_manifest(code, targets, filename = nil, noop = false)
+      Puppet[:tasks] = false
```
donoghuc: Was this a bug?

nicklewis: Not per se. `:tasks` defaults to false. When we run `bolt apply`, we just leave it alone and it works fine. When we run other things, we call `Puppet::Pal#with_script_compiler`, which sets `:tasks` to true.

But since we now use PAL / run tasks while setting up plugins, we can wind up with `:tasks` set to true when running `bolt apply`. So we have to make sure to explicitly turn it off.

At the same time, it may be a bug that `Puppet::Pal#with_script_compiler` doesn't restore `:tasks` to whatever its value was previously. `Puppet::Pal#with_catalog_compiler` sets it to false and then _does_ restore it afterward.

nicklewis: Opened puppetlabs/puppet#7822 to fix the upstream puppet issue

nicklewis:
Pushed a new commit to handle setting `plugin_hooks` with a top-level plugin reference. We now evaluate the plugins _first_ and then merge them over top of the defaults rather than vice versa.

Also squashed in a change to check whether `plugins` is set with a top-level plugin reference and fail.

donoghuc:
@nicklewis Can you add a CHANGELOG update for this issue?

nicklewis:
Rebased and updated the changelog

CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # Changelog
 
+## Bolt Next
+
+### New features
+
+* **Plugin support in `bolt.yml`** ([#1381](https://github.com/puppetlabs/bolt/pull/1381))
```
donoghuc: I think the 1.38.0 changelog PR got merged before we could review it. Would you mind adding this section under it and deleting Bolt Next?

donoghuc: Also, could you add an empty line after https://github.com/puppetlabs/bolt/pull/1381/commits/c257b05985abeb0492f4f69aac8ff3547e6e2066#diff-4ac32a78649ca5bdd8e0ba38b7006a1eR27

nicklewis: Updated

donoghuc: Thanks!

PR #1385, opened 2019-11-06
Feature: LXD transport

dionysius:
I've written a transport package for lxd very similiar to docker in source code and test code. Motivation is a potential transport package for kubernetes (using kubelet) but first learning the transport library with a known environment. I have almost no experience with ruby and just a little with bolt so reviewing it might be advised.

The lxd transport works the same way as docker: The node uri is the name of the container. Keep in mind that lxd container names are hostnames and thus the naming of the containers are a bit less flexible. But that's nothing this transport module cares, just a note ahead when playing around with it.

~~The only thing needed to run the tests is a linux with lxd installed and `lxc launch ubuntu:bionic ubuntunode`~~ I've added `spec/lxd-compose.sh` since I obviously can't add this container to docker-compose. I don't know how you want to add this step into the [testing requirements](https://github.com/puppetlabs/bolt/blob/master/CONTRIBUTING.md#testing)

To run the tests: `bundle exec rake integration:lxd`

On my machine the same tests fail for [docker](https://gist.github.com/dionysius/f9fec249454d6a44349dc3fc2954126b) and [lxd](https://gist.github.com/dionysius/6a2629d2e5702be7865043d3fa507ea0), assuming these are edge cases.

I also don't know which transport option I should use to define the lxd remote. There are no service-uri's in LXD. Each remote is configured in a file and per user, respectively using `lxc remote ...` and thus all remotes have just a name specified locally. I didn't want to introduce a new param so I reused it, but feel free to decide how you want it.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -19,14 +19,14 @@ RSpec::Core::RakeTask.new(:spec)
 
 desc "Run RSpec tests that don't require VM fixtures or a particular shell"
 RSpec::Core::RakeTask.new(:unit) do |t|
-  t.rspec_opts = '--tag ~ssh --tag ~docker --tag ~bash --tag ~winrm ' \
+  t.rspec_opts = '--tag ~ssh --tag ~docker --tag ~lxd --tag ~bash --tag ~winrm ' \
                  '--tag ~appveyor_agents --tag ~puppetserver --tag ~puppetdb ' \
                  '--tag ~omi --tag ~kerberos'
 end
 
 desc "Run RSpec tests for AppVeyor that don't require SSH, Bash, Appveyor Puppet Agents, or orchestrator"
 RSpec::Core::RakeTask.new(:appveyor) do |t|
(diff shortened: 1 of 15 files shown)


adreyer:
The connection options aren't really shared between transports so I think it's better to make a newer more specific option than `server-url`

adreyer:
This has conflicts and hasn't seen any movement recently so I'm closing it to declutter.

dionysius:
Well, I got no answer how this is going/should go forward... The feature is complete except some help from your side to make useful integration tests using actual lxd containers - and beside the one comment, I'm sure there are more wishes to this PR once it got reviewed thoroughly. A lot of code is very similiar to docker so it might be advisable to plan and take effort to rewrite/extract common parts out. Otherwise, it is on the same technical level as the docker transport - which seems to be used for testing (only?). Before I intend to invest time into that, I need some sort of feedback how this PR is received. It makes no sense to spend effort into something that no-one wants/needs. I'm sure it is nice to have additional transports. And I'll might offer a kubernetes transport PR with kubectl when the time comes in my other project.

PR #1388, opened 2019-11-06
(GH-1191) Support _run_as metaparameter for puppet_library plugins

### LINKED ISSUE #1191: Run apply_prep task with configured run_as

adreyer:
Closes #1191

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Changelog
 
+## Bolt Next
+
+### New features
+
+* **Support `_run_as` parameter for puppet_library hook** ([#1191](https://github.com/puppetlabs/bolt/issues/1191))
+
+  Bolt now accepts the `_run_as` metaparameter for puppet_library hooks. `_run_as` specifies which user the library install task will be executed as.
+
 ## Bolt 1.37.0
 
 ## New features
@@ -43,7 +51,15 @@
 * **Bolt issues a warning when inventory overrides a CLI option** ([#1341](https://github.com/puppetlabs/bolt/issues/1341))
 
   Bolt issues a warning when an option is set both on the CLI and in the inventory, whether the inventory loads from a file or from the `bolt_inventory` environment variable.
-  
+<<<<<<< HEAD
+
+=======
+
+* **New `resolve_references` plan function** ([#1365](https://github.com/puppetlabs/bolt/issues/1365))
+
+  The new plan function, `resolve_references`, accepts a hash of structured data and returns a hash of structured data with all plugin references resolved.
+
+>>>>>>> (GH-1191) Support _run_as metaparameter for puppet_library plugins
 ### Bug fixes
 
 * **Some configured paths were relative to Boltdir and some were relative to the current working directory** ([#1162](https://github.com/puppetlabs/bolt/issues/1162))
--- a/lib/bolt/plugin/module.rb
+++ b/lib/bolt/plugin/module.rb
@@ -158,20 +158,21 @@ module Bolt
         # opts are passed directly from inventory but all of the _ options are
         # handled previously. That may not always be the case so filter them
         # out now.
-        _meta, params = opts.partition { |key, _val| key.start_with?('_') }.map(&:to_h)
-        metaparams = {}
+        meta, params = opts.partition { |key, _val| key.start_with?('_') }.map(&:to_h)
+
         # Send config with `_config` when config is defined in bolt_plugin.json
         # Otherwise, merge config with params
         # TODO: remove @send_config when deprecated
         if @send_config
-          metaparams['_config'] = config if config?
+          validate_params(task, params)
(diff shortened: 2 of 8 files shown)


lucywyman:
Rubocop failure. I think this also needs a changelog entry

CHANGELOG.md
```diff
@@ -25,7 +25,11 @@
 * **Bolt issues a warning when inventory overrides a CLI option** ([#1341](https://github.com/puppetlabs/bolt/issues/1341))
 
   Bolt issues a warning when an option is set both on the CLI and in the inventory, whether the inventory loads from a file or from the `bolt_inventory` environment variable.
-  
+
+* **Support `_run_as` parameter for puppet_library hook** ([#1191](https://github.com/puppetlabs/bolt/issues/1191))
+
+  Bolt now accepts the `_run_as` metaparameter that allows users to specify which user the library install task should be executed as.
```
K8med: Clarity edit: Bolt now accepts the `_run_as` metaparameter. `_run_as` specifies which user should execute the library install task.
But want to make sure I captured the information correctly.

lib/bolt/plugin/module.rb
```diff
@@ -158,20 +158,20 @@ def process_params(task, opts)
...
-        metaparams = {}
+        meta, params = opts.partition { |key, _val| key.start_with?('_') }.map(&:to_h)
+        validate_params(task, params)
+
         # Send config with `_config` when config is defined in bolt_plugin.json
         # Otherwise, merge config with params
         # TODO: remove @send_config when deprecated
         if @send_config
-          metaparams['_config'] = config if config?
+          params['_config'] = config if config?
```
lucywyman: I think that `metaparams` used to collect all of the `_*` params, but now we're doing that with the `partition` command. What this is doing is adding `_config` and `_boltdir` to the task params, which I think is part of what's causing the travis failures.

CHANGELOG.md
```diff
@@ -25,7 +25,11 @@
 * **Bolt issues a warning when inventory overrides a CLI option** ([#1341](https://github.com/puppetlabs/bolt/issues/1341))
 
   Bolt issues a warning when an option is set both on the CLI and in the inventory, whether the inventory loads from a file or from the `bolt_inventory` environment variable.
-  
+
+* **Support `_run_as` parameter for puppet_library hook** ([#1191](https://github.com/puppetlabs/bolt/issues/1191))
```
lucywyman: This seems like really it's adding `_run_as` support to all task-based plugins, including the puppet_library one.

CHANGELOG.md
lucywyman: This seems like really it's adding _run_as support to all task-based plugins, including the puppet_library one.

lib/bolt/plugin/module.rb
```diff
@@ -199,6 +199,7 @@ def extract_task_parameter_schema
       end
 
       def run_task(task, opts)
+        opts = opts.reject{ |key, _val| key.start_with?('_') }
```
lucywyman: Purely out of curiosity is there a reason not to use `reject!` here?

nicklewis: It's generally bad form to modify arguments passed into a function. If the caller wanted to pass `opts` to `run_task` and then use it for something else after, they would find it modified.

lib/bolt/plugin/module.rb
```diff
@@ -158,20 +158,20 @@ def process_params(task, opts)
         # opts are passed directly from inventory but all of the _ options are
         # handled previously. That may not always be the case so filter them
         # out now.
-        _meta, params = opts.partition { |key, _val| key.start_with?('_') }.map(&:to_h)
-        metaparams = {}
+        meta, params = opts.partition { |key, _val| key.start_with?('_') }.map(&:to_h)
+        validate_params(task, params)
```
lucywyman: This validation needs to happen after merging with config, so that keys configured in `config` instead of the inventory are included in validation

lucywyman: Also needs a rubocop run

adreyer: shouldn't those keys have been validated when the config was loaded? We definitely cannot validate after `_config` is added.

lucywyman: We merge config from bolt.yaml with config from inventory if no config is defined in `bolt_plugin.json`, so we need to validate the merged config+inventory in that case

PR #1389, opened 2019-11-06
(DO NOT MERGE) (GH-1325) Update modules to ship Ruby plugin helper

### LINKED ISSUE #1325: Standardize templated data in target plugins
## Use Case
The Target plugins are kind of confusing because they contain templated data intermixed with options for the plugin.

* Users don't understand what options are template options vs other plugin options
* Users try to use static values in templated object
* Users try to use dynamic options outside the plugin(bad examples?)
* plugins in templates are not evaluated in the expected order.

## Solution
This code should live in a new 'plugin_helper' module. 

We should add a `target_mapping` key at the same level as `_plugin`, under which dynamic `name`, `uri`,  `config` etc. keys are nested.
```
groups:
  - name: cloud-webs
    targets:
      - _plugin: terraform
        dir: /path/to/terraform/project1
        resource_type: google_compute_instance.web
        target_mapping:
          uri: network_interface.0.access_config.0.nat_ip
          config:
            ssh:
              user: path.to.user
    config:
      ssh:
        password: static
        user:
        _plugin: prompt
        message: "Enter user:"
```
## Questions
- Should we allow static data in the `mappings` key? No, values under `mapping` can only be loaded dynamically from the plugin
- Should we use dot indexing? Not for keys, but we should support it for values (ie the TF plugin).
  - Don't need to worry about escaping dots for this.
  - When parsing dot-indexed values, the helper should treat numbers as indices if the object is an array, otherwise it should treat them as an integer.

lucywyman:
This updates the Puppetfile to include the new ruby plugin helper, which
provides helper methods for writing Bolt plugins in ruby. It also
updates the AWS inventory plugin to the newer version, which uses the
plugin helper

Closes #1325

### DIFF
--- a/Puppetfile
+++ b/Puppetfile
@@ -27,12 +27,13 @@ mod 'puppetlabs-puppet_conf', '0.4.0'
 mod 'puppetlabs-python_task_helper', '0.3.0'
 mod 'puppetlabs-reboot', '2.2.0'
 mod 'puppetlabs-ruby_task_helper', '0.4.0'
+mod 'puppetlabs-ruby_plugin_helper', '0.1.0'
 
 # Plugin modules
 mod 'puppetlabs-azure_inventory', '0.2.0'
 mod 'puppetlabs-terraform', '0.2.0'
 mod 'puppetlabs-vault', '0.2.2'
-mod 'puppetlabs-aws_inventory', '0.2.0'
+mod 'puppetlabs-aws_inventory', '0.3.0'
 
 # If we don't list these modules explicitly, r10k will purge them
 mod 'canary', local: true


Puppetfile
```diff
@@ -27,12 +27,13 @@ mod 'puppetlabs-puppet_conf', '0.4.0'
...
 mod 'puppetlabs-reboot', '2.2.0'
 mod 'puppetlabs-ruby_task_helper', '0.4.0'
+mod 'puppetlabs-ruby_plugin_helper', '0.1.0'
 
 # Plugin modules
 mod 'puppetlabs-azure_inventory', '0.2.0'
 mod 'puppetlabs-terraform', '0.2.0'
 mod 'puppetlabs-vault', '0.2.2'
-mod 'puppetlabs-aws_inventory', '0.2.0'
+mod 'puppetlabs-aws_inventory', '0.3.0'
```
donoghuc: I think it would make sense to release all the `inventory` modules with `target_mappings` at the same time/release. That we we can definitively say: these require `target_mappings`.

PR #1390, opened 2019-11-06
(GH-1377) Add project migrate command

### LINKED ISSUE #1377: Command to convert inventory v1 to v2
## Use Case
I have a v1 inventory and I want to automatically transcribe it to v2.  

Add a new CLI command `bolt project migrate`. This loads your inventory file and upgrades it to v2 in place. This should emit yaml, You will loose comments and we will have to make sure anchors/references are not included (as they cannot be loaded with YAML#safe_load). This should just operate on loaded data structure from the target inventoryfile instead of trying to create an inventory object. We need to rename `name` to `uri` and `nodes` to `targets`. This needs to also add `version: 2` to inventory, if `version: 2` is already present, do nothing. 

Requiring a path to an inventoryfile. 

## Additional Context
We assume you are version controlling your project dir and can test the outcome of the migration and commit on green.

beechtom:
This adds support for a `project migrate` command that migrates a Bolt
project to the newest version. Specifically, the inventory file for the
project will be migrated from an earlier version (`v1`) to the latest
version (`v2`). Changes to the inventory file are made in place and
comments/formatting are not preserved during migration.

Closes #1377

### DIFF
--- a/lib/bolt/cli.rb
+++ b/lib/bolt/cli.rb
@@ -37,7 +37,7 @@ module Bolt
                  'secret' => %w[encrypt decrypt createkeys],
                  'inventory' => %w[show],
                  'group' => %w[show],
-                 'project' => %w[init],
+                 'project' => %w[init migrate],
                  'apply' => %w[] }.freeze
 
     attr_reader :config, :options
@@ -368,7 +368,11 @@ module Bolt
 
       case options[:subcommand]
       when 'project'
-        code = initialize_project
+        if options[:action] == 'init'
+          code = initialize_project
+        elsif options[:action] == 'migrate'
+          code = migrate_project
+        end
       when 'plan'
         code = run_plan(options[:object], options[:task_options], options[:target_args], options)
       when 'puppetfile'
@@ -558,6 +562,52 @@ module Bolt
       outputter.print_message result
 
       ok ? 0 : 1
+    end
+
+    def migrate_project
+      if inventory.version == 2
+        ok = true
+      else
+        inventory_file = config.inventoryfile || config.boltdir.inventory_file
+
+        begin
+          Bolt::Util.file_stat(inventory_file)
+        rescue Errno::ENOENT
+          raise Bolt::FileError.new("The inventory file '#{inventory_file}' does not exist", inventory_file)
+        end
+
+        inv = YAML.safe_load(File.open(inventory_file))
+        migrate_group(inv)
+
+        ok = File.write(inventory_file, { 'version' => 2 }.merge(inv).to_yaml)
+      end
+
+      result = if ok
+                 "Successfully migrated Bolt project to latest version"
+               else
+                 "Could not migrate Bolt project to latest version"
+               end
+      outputter.print_message result
+
+      ok ? 0 : 1
+    end
+
+    # Walks an inventory hash and replaces all 'nodes' keys with 'targets' keys
+    # and all 'name' keys nested in a 'targets' hash with 'uri' keys. Data is
+    # modified in place.
+    def migrate_group(group)
+      if group.key?('nodes')
+        targets = group['nodes'].map do |target|
+          target['uri'] = target.delete('name') if target.is_a?(Hash)
+          target
+        end
+        group.delete('nodes')
+        group['targets'] = targets
+      end
+      (group['groups'] || []).each do |subgroup|
+        migrate_group(subgroup)
+      end
+      nil
     end
 
     def install_puppetfile(config, puppetfile, modulepath)
--- a/spec/bolt/cli_spec.rb
+++ b/spec/bolt/cli_spec.rb
@@ -2152,21 +2152,73 @@ describe "Bolt::CLI" do
   end
 
   describe 'project' do
-    it 'init creates a new project at the specified path' do
-      Dir.mktmpdir do |dir|
-        file = File.join(dir, 'bolt.yaml')
-        cli = Bolt::CLI.new(%W[project init #{dir}])
-        cli.execute(cli.parse)
-        expect(File.file?(file)).to be
+    context 'migrate' do
+      let(:inventory_v1) do
+        {
+          "name" => "all",
+          "groups" => [{
+            "name" => "group1",
+            "nodes" => [{
+              "name" => "target1",
+              "facts" => {
+                "name" => "foo"
+              }
+            }]
+          }]
+        }
+      end
+      let(:inventory_v2) do
+        {
+          "version" => 2,
+          "name" => "all",
+          "groups" => [{
+            "name" => "group1",
+            "targets" => [{
+              "uri" => "target1",
+              "facts" => {
+                "name" => "foo"
+              }
+            }]
+          }]
+        }
+      end
+
+      it 'migrates inventory v1 to inventory v2' do
+        with_tempfile_containing('inventory', YAML.dump(inventory_v1)) do |file|
+          cli = Bolt::CLI.new(%W[project migrate --inventoryfile #{file.path}])
+          cli.execute(cli.parse)
+          expect(YAML.load_file(file)).to eq(inventory_v2)
+        end
+      end
+
+      it 'does nothing when using inventory v2' do
+        inventory_v1['version'] = 2
+        with_tempfile_containing('inventory', YAML.dump(inventory_v1)) do |file|
+          contents = File.read(file)
+          cli = Bolt::CLI.new(%W[project migrate --inventoryfile #{file.path}])
+          cli.execute(cli.parse)
+          expect(File.read(file)).to eq(contents)
+        end
       end
     end
 
-    it 'init creates a new project in the current working directory' do
-      Dir.mktmpdir do |dir|
-        file = File.join(dir, 'bolt.yaml')
-        cli = Bolt::CLI.new(%w[project init])
-        Dir.chdir(dir) { cli.execute(cli.parse) }
-        expect(File.file?(file)).to be
+    context 'init' do
+      it 'init creates a new project at the specified path' do
+        Dir.mktmpdir do |dir|
+          file = File.join(dir, 'bolt.yaml')
+          cli = Bolt::CLI.new(%W[project init #{dir}])
+          cli.execute(cli.parse)
+          expect(File.file?(file)).to be
+        end
+      end
+
+      it 'init creates a new project in the current working directory' do
+        Dir.mktmpdir do |dir|
+          file = File.join(dir, 'bolt.yaml')
+          cli = Bolt::CLI.new(%w[project init])
+          Dir.chdir(dir) { cli.execute(cli.parse) }
+          expect(File.file?(file)).to be
+        end
       end
     end
   end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@
 
 * **Add `--password-prompt` and `--sudo-password-prompt` CLI flags** ([#1269](https://github.com/puppetlabs/bolt/issues/1269))
   Two new flags have been added to support the case where the user would like to set a `password` or `sudo-password` from a prompt without using a plugin. Deprecations messages have been added when a value is not supplied for `--password` or `--sudo-password`.
+  
+* **Subcommand `project migrate` new to the CLI** ([#1377](https://github.com/puppetlabs/bolt/issues/1377))
+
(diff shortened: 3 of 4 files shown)


lib/bolt/cli.rb
```diff
@@ -559,6 +563,40 @@ def initialize_project
       ok ? 0 : 1
     end
 
+    def migrate_project
+      if inventory.version == 2
+        ok = true
+      else
+        inventoryfile = config.inventoryfile || config.boltdir.inventory_file
```
donoghuc: While i dont think you necessarily need to worry about inventory from BOLT_INVENTORYenvironment variable, I recently found it slightly more complicated that you might initially think to find the actual source of inventory data... https://github.com/puppetlabs/bolt/blob/ab4a96649e8e6a74393527ea8b3b5a9be40a83f8/lib/bolt/cli.rb#L269-L282 maybe we should share that logic?

lib/bolt/cli.rb
```diff
@@ -559,6 +563,40 @@ def initialize_project
...
+        inv = YAML.load_file(inventoryfile)
```
donoghuc: I think you want safe_load

lib/bolt/cli.rb
```diff
@@ -559,6 +563,40 @@ def initialize_project
...
+
+        inv = Bolt::Util.walk_pairs(inv) do |k, v|
```
donoghuc: I think you can just use `Bolt::Util#walk_keys`. 
```
        inv = Bolt::Util.walk_keys(inv) do |k|
          case k
          when k == 'nodes'
            k = 'target'
          when k == 'name'
            k = 'uri'
          else
            k
          end
        end
```

lib/bolt/cli.rb
```diff
@@ -559,6 +563,40 @@ def initialize_project
...
+            v = v.map do |target|
+              target['uri'] = target.delete('name') if target.is_a? Hash
+              target
+            end
+          end
+
+          { k => v }
+        end
+
+        inv['version'] = 2
```
donoghuc: Looks like this will print `version:2` at the very bottom of the file. If you did something like
```
 ok = File.write(inventoryfile, {'version' => 2}.merge(inv).to_yaml)
```
I think it would print at the top.

spec/bolt/cli_spec.rb
```diff
@@ -2113,21 +2113,79 @@ def stub_config(file_content = {})
...
+        lambda { |inventory|
+          file = Tempfile.new('inventory.yaml')
+          file.write(inventory.to_yaml)
+          file.rewind
+          file.path
+        }
+      end
+
+      it 'migrates inventory v1 to inventory v2' do
+        file = inventoryfile[inventory_v1]
```
donoghuc: Check out https://github.com/puppetlabs/bolt/blob/bc499bedef615a1aad96fc34aa31fc78598bd6d5/spec/lib/bolt_spec/files.rb#L7

lib/bolt/cli.rb
beechtom: That would update every name key to uri, instead of just the name key for targets. Names for groups/facts/etc would be changed as well, which we don't want .

donoghuc: ahhh, yeah. forgot about that.

donoghuc: I guess the reason I brought that up was more that I just don't think we need a util method that walks pairs. I think we can do it with an existing one. What about something like:
```ruby
        inv = Bolt::Util.walk_vals(inv) do |v|
          if v.is_a?(Hash) && (grps = v['groups'])
            grps.map do |grp|
              grp.map do |key, val|
                if key == 'nodes'
                  key == 'targets'
                  val = val.map do |target|
                    target['uri'] = target.delete('name') if target.is_a?(Hash)
                    target
                  end
                end
              end
            end
          end
          v
        end
```
it is nasty but I think we could make it better maybe?

beechtom: I had something similar to that prior to opening the PR, but also felt it was messy. I don't really have a preference for one approach v another, so if we want to remove `walk_pairs` I can revert back to what I had previously.

lib/bolt/cli.rb
beechtom: It seems like the inventory is sourced the same way, this just has some extra validation thrown in with the file stat? If we want this to be reusable would it make sense to have it in `config`?

lib/bolt/cli.rb
lucywyman: @donoghuc Is your concern adding more functions to Util, or overcomplicating this traversal since we can filter what we walk here based on `Groups`?  When adding `walk_vals` we only used it in one place, so I don't see a problem with adding a function to `walk_hashes` and it could potentially be useful elsewhere where context for traversing objects is necessary (like it is here). 

If the concern is wanting to filter the top-level keys we traverse, I think we could write a more fine-tuned traversal method and include it here instead. Given the algorithm above I don't think that's your concern?

Worth noting the above algorithm misses nodes in the top-level `nodes` key (that are part of the `all` group)

lucywyman: @beechtom This seems to work:
```
        inv = Bolt::Util.walk_vals(inv) do |v| 
          if v.is_a?(Hash)
            v['targets'] = v.delete('nodes') if v.has_key?('nodes')
            if v.has_key?('targets')
              v['targets'] = v['targets'].map do |target|
                target['uri'] = target.delete('name') if target.is_a? Hash
                target
              end 
            end 
          end 
          v   
        end 
```

lib/bolt/cli.rb
```diff
@@ -559,6 +563,38 @@ def initialize_project
...
     end
 
+    def migrate_project
+      if inventory.version == 2
+        ok = true
+      else
+        inventoryfile = config.inventoryfile || config.boltdir.inventory_file
+        inv = YAML.safe_load(File.open(inventoryfile))
+
+        inv = Bolt::Util.walk_pairs(inv) do |k, v|
```
nicklewis: It would be helpful to have comments explicitly documenting the set of transformations we make here.

lib/bolt/util.rb
```diff
@@ -118,6 +118,21 @@ def map_vals(hash)
         end
       end
 
+      # Accepts a Data object and returns a copy with all hash keys and values
+      # modified by block.
+      def walk_pairs(data, &block)
```
nicklewis: I can't get an intuitive sense of what this function is doing. I think the problem I have with it is that it requires that a particular key will always have the same meaning no matter where it's found in the nesting.

lib/bolt/cli.rb
nicklewis: It might be clearer to do this with explicit recursion than with a helper function.

```rb
def migrate_group(group)
  if group.key?('nodes')
    targets = group['nodes'].map do |target|
      target['uri'] = target.delete('name') if target.is_a?(Hash)
      target
    end
    group.delete('nodes')
    group['targets'] = targets
  end
  (group['groups'] || {}).map { |name, subgroup| migrate_group(subgroup) }
end
```

lib/bolt/cli.rb
nicklewis: It might be clearer to do this with explicit recursion than with a helper function.

```rb
def migrate_group(group)
  if group.key?('nodes')
    targets = group['nodes'].map do |target|
      target['uri'] = target.delete('name') if target.is_a?(Hash)
      target
    end
    group.delete('nodes')
    group['targets'] = targets
  end
  (group['groups'] || {}).map { |name, subgroup| migrate_group(subgroup) }
end
```

beechtom: Updated to use `walk_vals`.

donoghuc: Does the recursive algo that Nick suggest work? I kind of like that one for readability.

beechtom: I'm probably not using it correctly, but I can't get it to work.

lib/bolt/cli.rb
donoghuc: For some reason i think I remember finding a case where the config.boldir.inventoryfile is different from the array in config.default_inventoryfile. Also you will for sure want to verify that the inventory file exists. If you do `bolt project migrate` currently with a non-existent inventoryfile you get a ENOENT stacktrace.

beechtom: Got it. Would you recommend I pull this functionality out into another method or something else?

donoghuc: I think a new method on the CLI method seems reasonable, this is the second time we want to know *where* inventory will be loaded from without loading it. I based the algo I posted on https://github.com/puppetlabs/bolt/blob/b12168d670853030657156b7242585e5bf5cd5fa/lib/bolt/inventory.rb#L46-L61

beechtom: I'm also really curious what that case would be since it seems like `config.default_inventoryfile` shouldn't be different from `config.boltdir.inventory_file` since the former just returns `[@boltdir.inventory_file]`.

donoghuc: I must have made that up.

beechtom: Added an `inventory_source` method to the CLI.

lib/bolt/cli.rb
```diff
@@ -559,6 +550,41 @@ def initialize_project
       ok ? 0 : 1
     end
 
+    def migrate_project
+      if inventory.version == 2
+        ok = true
+      else
+        inventory_file = inventory_source
```
donoghuc: Inventory source can return the constant `Bolt::Inventory::ENVIRONMENT_VAR` which points to the `BOLT_INVENTORY` environment variable. That allows you to specify inventory data (not a path to an inventory file). So I think we would want to either skip the migrate, or perhaps load the inventory data, check for v2, and print the conversion?

lib/bolt/cli.rb
donoghuc: Seems to work for me with 
```ruby
    def migrate_group(group)
      if group.key?('nodes')
        targets = group['nodes'].map do |target|
          target['uri'] = target.delete('name') if target.is_a?(Hash)
          target
        end
        group.delete('nodes')
        group['targets'] = targets
      end
      (group['groups'] || []).map do |subgroup|
        migrate_group(subgroup)
      end
    end
```
then simply replace the walk_vals logic with 
`migrate_groups(inv)`

lib/bolt/cli.rb
lucywyman: I think printing the conversion there makes sense - I can see users having docker workflows with inventory exported in a non-inventory file, and having a way to convert and copy-paste seems valuable. I think it'd also be fine to skip migration in that case for now and ticket it - we may want to define that behavior a bit more as a team.

beechtom: Removed the `inventory_source` method and do error handling in place. I agree with Lucy to skip migration for the environment var for now and ticket it separately.

lib/bolt/cli.rb
beechtom: Updated.

lib/bolt/cli.rb
```diff
@@ -559,6 +563,50 @@ def initialize_project
...
+      if group.key?('nodes')
+        targets = group['nodes'].map do |target|
+          target['uri'] = target.delete('name') if target.is_a?(Hash)
+          target
+        end
+        group.delete('nodes')
+        group['targets'] = targets
+      end
+      (group['groups'] || []).map do |subgroup|
+        migrate_group(subgroup)
```
donoghuc: Maybe `map` is confusing given the data structure is modified in place. We could use each, and explicitly return `nil` to avoid confusion on expected return value. We should also update the comment about mutating the data. 
```ruby
    def migrate_group(group)
      if group.key?('nodes')
        targets = group['nodes'].map do |target|
          target['uri'] = target.delete('name') if target.is_a?(Hash)
          target
        end
        group.delete('nodes')
        group['targets'] = targets
      end
      (group['groups'] || []).each { |subgroup| migrate_group(subgroup)} 
      nil
    end
```

lib/bolt/cli.rb
```diff
@@ -559,6 +563,50 @@ def initialize_project
...
     end
 
+    def migrate_project
+      if inventory.version == 2
+        ok = true
+      else
+        inventory_file = config.inventoryfile || config.boltdir.inventory_file
+
+        begin
+          Bolt::Util.file_stat(inventory_file)
```
donoghuc: I guess it makes sense to error both case where the default inventory file is not present in a Boltdir and when the specified inventoryfile is not present. For now project migrate only acts on inventory so without an inventoryfile it seems reasonable to error out.

lib/bolt/cli.rb
beechtom: Updated.

PR #1393, opened 2019-11-08
(GH-1269) Warn when --{sudo}-password is empty

donoghuc:
The initial implementation of GH-1269 prematurely forbit optional parameters for cli opts that could expect a prompt. This PR re-allows them and issues a warning when there is no value provided for `--password` or `--sudo-password`. This is accompished by adding a new instance variable to `BoltOptionParser` that is available with an attr reader to collect a hash of warnings when parsing options and surface them once the logger is configured.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Changelog
 
+## Bolt 1.36.1
+
+## Bug fixes
+
+* **Allow optional `--password` and `--sudo-password` parameters** ([#1269](https://github.com/puppetlabs/bolt/issues/1269))
+
+  Optional parameters for `--password` and `--sudo-password` were prematurely removed. The previous behavior of prompting for a password when an argument is not specified for `--password` or `--sudo-password` has been added back. Arguments will be required in a future version.
+
 ## Bolt 1.36.0
 
 ### Deprecation
@@ -29,10 +37,6 @@
 * **Bolt issues a warning when inventory overrides a CLI option** ([#1341](https://github.com/puppetlabs/bolt/issues/1341))
 
   Bolt issues a warning when an option is set both on the CLI and in the inventory, whether the inventory loads from a file or from the `bolt_inventory` environment variable.
-
-* **New `resolve_references` plan function** ([#1365](https://github.com/puppetlabs/bolt/issues/1365))
-
-  The new plan function, `resolve_references`, accepts a hash of structured data and returns a hash of structured data with all plugin references resolved.
   
 ### Bug fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,12 @@
 # Changelog
 
-## Bolt 1.36.1
+## Bolt 1.37.0
+
+## New features
+
+* **New `resolve_references` plan function** ([#1365](https://github.com/puppetlabs/bolt/issues/1365))
+
+  The new plan function, `resolve_references`, accepts a hash of structured data and returns a hash of structured data with all plugin references resolved.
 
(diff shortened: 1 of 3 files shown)


lib/bolt/bolt_option_parser.rb
```diff
@@ -338,10 +338,12 @@ def self.examples(cmd, desc)
       Available options are:
     PROJECT_INIT_HELP
 
+    attr_reader :warnings
     def initialize(options)
       super()
 
       @options = options
+      @warnings = {}
```
donoghuc: I made this a hash where key is the option and value is the message thinking we might want to know what option is associated with what message in the cli. But i dont actually use the keys currently, should it just be an array?

lib/bolt/bolt_option_parser.rb
```diff
@@ -396,13 +398,16 @@ def initialize(options)
       end
       define('-p', '--password [PASSWORD]',
              'Password to authenticate with') do |password|
-        # TODO: Remove deprecation message
-        unless password
-          msg = "Optional parameter for --password is deprecated and no longer prompts for password. " \
+        if password.nil?
+          msg = "Optional parameter for --password is deprecated and will no longer prompt for password. " \
```
lucywyman:
```suggestion
          msg = "Optional parameter for --password is deprecated.  " \
```

lib/bolt/bolt_option_parser.rb
```diff
@@ -424,12 +429,16 @@ def initialize(options)
       define('--sudo-password [PASSWORD]',
              'Password for privilege escalation') do |password|
         # TODO: Remove deprecation message
-        unless password
-          msg = "Optional parameter for --sudo-password is deprecated and no longer prompts for password. " \
+        if password.nil?
+          msg = "Optional parameter for --sudo-password is deprecated and will longer prompt for password. " \
```
lucywyman:
```suggestion
          msg = "Optional parameter for --sudo-password is deprecated. " \
```

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## Bolt 1.36.1
+
```
```suggestion
### Bug fixes
```

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
...
+* **Allow optional `--password` and `--sudo-password` parameters** ([#1269](https://github.com/puppetlabs/bolt/issues/1269))
+
+  Optional parameters for `--password` and `--sudo-password` were prematurely removed. The previous behaviour of whereby when an argument is not specified for `--password` or `--sudo-password` the user will be prompted for a value. Arguments will be required in a future version.
```
```suggestion
  Optional parameters for `--password` and `--sudo-password` were prematurely removed. The previous behavior of prompting for a password when an argument is not specified for `--password` or `--sudo-password` has been added back. Arguments will be required in a future version.
```

lib/bolt/bolt_option_parser.rb
lucywyman: It seems like for now we typically include the option in the message itself, so an array seems sufficient. Do you have a use case in mind for passing the associated option?

nicklewis: Perhaps an array of hashes?

```rb
@warnings = []
# ...
@warnings << { option: 'sudo-password', msg: 'Optional parameter for --sudo-password is deprecated and will no longer prompt for password.' }
```

lib/bolt/bolt_option_parser.rb
beechtom:
```suggestion
          msg = "Optional parameter for --password is deprecated. " \
```

PR #1394, opened 2019-11-08
(GH-1173) Allow task parameters to have a "default"

### LINKED ISSUE #1173: Allow tasks to specify defaults for parameters
## Use
New users of my tasks should be able to omit parameters, and have a sensible default value applied.

I'd like it if the metadata.json file supported an optional "default" attribute on each parameter, the value of which would be used if the user omits it when invoking the task.  For instance:
```
  "parameters": {
    "port_number": {
      "description": "Port the remote service uses for HTTPS",
      "type": "String[1]",
      "default": "443"
    }
```

I mark these parameters as optional in the metadata.json, and then my script itself silently supplies hardcoded defaults if the parameter is missing.

## Additional Context
This was originally filed in Jira as https://tickets.puppetlabs.com/browse/BOLT-305

nicklewis:
This adds a new "default" field to the parameter specification in task
metadata. If a parameter is not specified and has a default value defined, the
default value will be used in its place. The default will be validated against
the parameter type specification and rejected if invalid.

Fixes #1173.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 ## Bolt Next
 
+### New features
+
+* **Task metadata can now specify parameter defaults** ([#1394](https://github.com/puppetlabs/bolt/pull/1394))
+
+  Parameter defaults can be set in the task metadata file and will be used if no value is supplied for the parameter.
+
 ### Bug fixes
 
 * **`bolt inventory show --detail` now displays all target aliases** ([#1379](https://github.com/puppetlabs/bolt/issues/1379))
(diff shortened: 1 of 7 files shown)


nicklewis:
This will _not_ cause bolt_server to honor parameter defaults, because it bypasses the `run_task` function and calls the executor directly. I think that's desirable for now, since we probably want to handle that separately.

nicklewis:
This should not be merged until the [corresponding puppet-specifications PR](https://github.com/puppetlabs/puppet-specifications/pull/145) is merged.

PR #1399, opened 2019-11-11
Break travis tests into a matrix

nicklewis:
This PR contains a number of test speed improvements, the most significant of which is breaking the travis tests into a matrix job.

Previously, we ran a set of steps in a single job in travis:
* Rubocop
* Most rspec tests
* Rspec tests for embedded modules
* Docs validation
* Expensive rspec tests that use external services

This took ~17-18 min on average and resulted in a huge amount of output
to sort through to see what failed.

We now run each of those steps in a separate cell of a matrix job. That
makes it easier to quickly see what failed and it also cuts a lot of
time off the test runs.

This adds a `travis` namespace to the Rakefile, with separate
`travis:<check>` jobs for each cell in the matrix. That gives us an easy
ways to define new steps and to manage what those steps should do, while
separating the set of jobs that travis care about from the set of jobs
that developers care about while working locally.

We also now run `rubocop` and `docker-compose build` with `--parallel`,
speeding each of those steps up a bit.


Rakefile
```diff
@@ -57,6 +57,43 @@ def format_links(text)
...
+        sh 'rake spec' do |ok, _|
+          success = false unless ok
+        end
+      end
+    end
+    raise "Module tests failed" unless success
+  end
+  task docs: :generate_docs
+  task :integration do
+    sh "docker-compose -f spec/docker-compose.yml build --parallel"
```
Iristyle: 👍 on `--parallel`. I never realized that existed!

I *believe* the reason that I built `samba-ad` first was because it needs to be up and running for the Travis CI runner to be able to connect as a Kerberos client - you can look at the commit here - https://github.com/puppetlabs/bolt/commit/4637ffc8ff4e2443d9fc13a9f5de531965b6de9e#diff-354f30a63fb0907d4ad57269548329e3R30 for when that was originally introduced.

I don't think you'll need to update docs for any changes you've made here, but you might want to double check https://github.com/puppetlabs/bolt/blob/master/developer-docs/kerberos.md

lib/bolt/pal.rb
```diff
@@ -242,6 +242,7 @@ def list_modulepath
     end
 
     def parse_params(type, object_name, params)
+      return params if params.empty?
```
donoghuc: With the change you added in CLI is it possible to have an empty params hash? It seems like by setting 'options[:params_parsed]' we don't actually parse_params? Maybe this is possible in testing though. Also would a test asserting that pal.parse_params is not invoked when it is not necessary appropriate?

spec/integration/transport/winrm_spec.rb
```diff
@@ -180,14 +179,7 @@ def stub_winrm_to_raise(klass, message)
       target.options.delete('cacert')
       target.options['ssl-verify'] = false
 
-      expect(winrm.run_command(omi_target, command)['stdout']).to eq("#{user}\r\n")
```
donoghuc: Is there any value in keeping the kerberos tests/infra definitions around for local testing, but not run in CI?

Rakefile
```diff
@@ -34,12 +34,12 @@ end
 desc "Run RSpec tests for TravisCI that don't require WinRM"
 RSpec::Core::RakeTask.new(:travisci) do |t|
   t.rspec_opts = '--tag ~winrm --tag ~appveyor_agents --tag ~puppetserver --tag ~puppetdb ' \
-  '--tag ~omi --tag ~windows --tag ~kerberos'
+  '--tag ~omi --tag ~windows --tag ~kerberos --tag ~expensive'
```
donoghuc: I love the tag name! :money_mouth_face: :money_with_wings:

lib/bolt/pal.rb
nicklewis: Whoops, I think I fixed this bug in two places. I was going to have this method return early then decided to make the CLI handle it instead. So this line can probably just be removed.

spec/integration/transport/winrm_spec.rb
nicklewis: I figured I would leave them to document the current state of the system at least.

This test in particular was removed because it's... just wrong. It disables SSL and then tries to connect to the HTTPS port and expects that that will work. I'm unclear how it worked before unless OMI silently downgrades??

donoghuc:
What if docs and rubocop were in the same cell? Would that end up being more efficient due to loading ruby etc?

nicklewis:
I was thinking about that. It would be more efficient in terms of resource usage, but they're not our resources and it doesn't decrease the _time_ taken. It's still probably a good improvement, although it does mean having two different checks in the same cell, so you still have to click in to find out which one had the problem. Not a big deal though.

spec/integration/transport/winrm_spec.rb
```diff
@@ -211,10 +203,11 @@ def stub_winrm_to_raise(klass, message)
     end
   end
 
-  context "authenticating with Kerberos", kerberos: true do
+  context "authenticating with Kerberos", kerberos: true, omi: true do
     before(:all) do
       # disable all tests on Windows for now
       skip('Windows Active Directory tickets are different') if Bolt::Util.windows?
+      skip("WinRM gem and OMI server have a protocol negotiation bug")
```
Iristyle: One of the 3 tests was working previously... but now they're all skipped. Not sure I understand this change.

nicklewis: The only test that ran before just ensured that running `kinit` locally had worked as a precondition for the subsequent tests, which were marked pending. So I just skipped the whole section instead.

PR #1400, opened 2019-11-11
(maint) Delete puts debuggng line in test

donoghuc:
It appears that some print debugging got merged in when new tests were introduced. This commit removes the noise.


donoghuc:
@adreyer or @lucywyman was this intentional?

PR #1411, opened 2019-11-12
(GH-1175) Populate $nodes and $targets with --targets and --nodes

donoghuc:
When `--nodes` or `--targets` is provided on CLI, "magically" pass that arg to `$nodes` or `$targets` if there is a plan parameter with name `$targets` XOR `$nodes` of type `TargetSpec` AND there is not a plan argument specified on the CLI with name `nodes` or `targets`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@
 
   Bolt now includes a plugin to look up data from a YAML file.
 
+* **Pass value of `--targets` or `--nodes` to `TargetSpec $target` plan parameter** ([#1175](https://github.com/puppetlabs/bolt/issues/1175))
+
+  Bolt now passes the value of `--targets` or `--nodes` to plans with a `TargetSpec $targets` parameter.
+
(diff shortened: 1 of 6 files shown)


donoghuc:
Looks like I forgot to check in some fixtures. Also need to look in to some legit integration failures.

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@
 
 ### New features
 
+* **Pass value of `--targets` or `--nodes` to `TargetSpec $target` plan parameter** ([#1175](https://github.com/puppetlabs/bolt/issues/1175))
+
+  Bolt now passes the value of `--targets` or `--nodes` to plans with a `TargetSpec $targets` parameter.
```
```suggestion
  Bolt now passes the value of `--targets` or `--nodes` to plans with a `TargetSpec $targets` parameter.
```

beechtom:
>

beechtom:
Closes #1175

Issue #1379, opened 2019-11-05
`Alias` key in `bolt inventory show --detail` only shows alias from rightmost target definition


lucywyman:
## Describe the Bug
If a target is defined multiple times with different aliases in an inventoryfile, the aliases are merged and the target can be run using any alias (https://github.com/puppetlabs/bolt/blob/master/lib/bolt/inventory/group2.rb#L416). However when printing the target, regardless of which alias is used, only the alias defined in the rightmost definition will be printed.

```yaml
---
version: 2
groups:
  - name: foo 
    targets:
      - name: localhost
        alias:
          - printed
          - boop
  - name: bar 
    targets:
      - name: localhost
        alias: also_printed
```

## Expected Behavior
The `--detail` output should list aliases from all groups a target is a member of.

## Steps to Reproduce
For example with this inventory, `bolt inventory show --detail` will output the same result for `-n all`, `-n localhost`, `-n printed`, `-n boop`, and `-n also_printed`, which is:

```
  "targets": [
    {
      "name": "localhost",
      "uri": null,
      "alias": "also_printed",
...
```

donoghuc:
Closed by https://github.com/puppetlabs/bolt/pull/1412

PR #1412, opened 2019-11-12
(GH-1379) Show all aliases for a target in inventory show --detail

beechtom:
This commit fixes a bug where `inventory show --detail` only showed the
rightmost alias for a target. Now, all aliases for a given target will
be displayed in the output.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Bug fixes
 
-* **`bolt inventory show -- detail` now displays all target aliases** ([#1379](https://github.com/puppetlabs/bolt/issues/1379))
+* **`bolt inventory show --detail` now displays all target aliases** ([#1379](https://github.com/puppetlabs/bolt/issues/1379))
 
   Bolt now displays aliases from all groups, where a target is a member, in the output for `bolt inventory show --detail`. Previously, only the rightmost alias appeared in the output.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Changelog
 
+## Bolt Next
+
+### Bug fixes
+
+* **`bolt inventory show -- detail` now displays all target aliases** ([#1379](https://github.com/puppetlabs/bolt/issues/1379))
+
+  Bolt now displays aliases from all groups, where a target is a member, in the output for `bolt inventory show --detail`. Previously, only the rightmost alias appeared in the output.
+
 ## Bolt 1.38.0
 
 ### New features
@@ -38,7 +46,6 @@
 * **Results from `file::exists` and `file::readable` errored** ([#1415](https://github.com/puppetlabs/bolt/pull/1415))
 
   The `file::exists` and `file::readable` functions no longer error when the file path is specified relative to a module and the file doesn't exist.
-
 
 ## Bolt 1.37.0
 
--- a/lib/bolt/inventory/group2.rb
+++ b/lib/bolt/inventory/group2.rb
@@ -153,6 +153,10 @@ module Bolt
         end
       end
 
+      def clear_alia(target_name)
+        @aliases.reject! { |_alias, name| name == target_name }
+      end
+
       def data_merge(data1, data2)
         if data2.nil? || data1.nil?
           return data2 || data1
@@ -162,7 +166,8 @@ module Bolt
           'config' => Bolt::Util.deep_merge(data1['config'], data2['config']),
           'name' => data1['name'] || data2['name'],
           'uri' => data1['uri'] || data2['uri'],
-          'alias' => data1['alias'] || data2['alias'],
+          # Collect all aliases across all groups for each target uri
+          'alias' => [*data1['alias'], *data2['alias']],
           # Shallow merge instead of deep merge so that vars with a hash value
           # are assigned a new hash, rather than merging the existing value
           # with the value meant to replace it
--- a/spec/bolt/inventory_spec.rb
+++ b/spec/bolt/inventory_spec.rb
@@ -751,11 +751,20 @@ describe Bolt::Inventory do
 
   context 'when using inventory show' do
     let(:data) {
-      { 'nodes' => [{
-        'name' => 'foo',
-        'alias' => %w[bar baz],
-        'config' => { 'ssh' => { 'disconnect-timeout' => 100 } },
-        'facts' => { 'foo' => 'bar' }
+      { 'groups' => [{
+        'name' => 'group1',
+        'nodes' => [{
+          'name' => 'foo',
+          'alias' => %w[bar]
+        }]
+      }, {
+        'name' => 'group2',
+        'nodes' => [{
+          'name' => 'foo',
+          'alias' => %w[baz],
+          'config' => { 'ssh' => { 'disconnect-timeout' => 100 } },
+          'facts' => { 'foo' => 'bar' }
+        }]
       }] }
     }
 
--- a/lib/bolt/inventory/inventory2.rb
+++ b/lib/bolt/inventory/inventory2.rb
@@ -173,10 +173,12 @@ module Bolt
         existing_target = @targets.key?(new_target.name)
         @targets[new_target.name] = new_target
 
-        unless existing_target
+        if existing_target
+          clear_alia_from_group(@groups, new_target.name)
+        else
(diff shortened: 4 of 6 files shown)


CHANGELOG.md
```diff
@@ -15,6 +15,12 @@
 
   The CLI now provides the subcommand `project migrate` which migrates Bolt projects to the latest version. When migrating a project, the [inventory file](https://puppet.com/docs/bolt/latest/inventory_file.html) will be changed from `v1` to `v2`. Changes are made in place and will not preserve comments or formatting.
 
+## Bug fixes
+
+* **Display all target aliases in `bolt inventory show --detail`** ([#1379](https://github.com/puppetlabs/bolt/issues/1379))
+
+  Bolt will now display aliases from all groups that a target is a member of in the output for `bolt inventory show --detail`. Previously, only the rightmost alias was shown in the output.
```
K8med: Bolt now displays aliases from all groups, where a target is a member, in the output for `bolt inventory show --detail`. Previously, only the rightmost alias appeared in the output.
```suggestion
  Bolt will now display aliases from all groups that a target is a member of in the output for `bolt inventory show --detail`. Previously, only the rightmost alias was shown in the output.
```

CHANGELOG.md
```diff
@@ -15,6 +15,12 @@
 
   The CLI now provides the subcommand `project migrate` which migrates Bolt projects to the latest version. When migrating a project, the [inventory file](https://puppet.com/docs/bolt/latest/inventory_file.html) will be changed from `v1` to `v2`. Changes are made in place and will not preserve comments or formatting.
 
+## Bug fixes
+
+* **Display all target aliases in `bolt inventory show --detail`** ([#1379](https://github.com/puppetlabs/bolt/issues/1379))
```
K8med: `bolt inventory show -- detail` now displays all target aliases

lib/bolt/target.rb
```diff
@@ -66,7 +66,7 @@ def safe_name
     end
 
     def target_alias
-      inventory_target.target_alias
+      inventory_target.target_aliases
```
donoghuc: I think the intent of the `Inventory::Target` is to handle aliases in the instance variable https://github.com/puppetlabs/bolt/blob/17d88e7c3ae1eece83327be5dcf7a3541c41e2b6/lib/bolt/inventory/target.rb#L39 I implemented the `target_aliases` method on v1 inventory because we do not track that per target. I think we need to figure out how to make storing aliases on instance variables (or if that is impossible we need to delete them).

lib/bolt/inventory/group.rb
```diff
@@ -52,13 +52,7 @@ def initialize(data)
             raise ValidationError.new("Node entry must be a String or Hash, not #{node.class}", @name)
           end
 
-          if @nodes.include?(node['name'])
```
donoghuc: I'm not sure I understand why inventory and group v1 need to change? I think the bug in the ticket only occurs with inventory v2.

lib/bolt/inventory/group.rb
```diff
@@ -75,9 +69,9 @@ def initialize(data)
             @logger.warn(msg)
           end
 
-          next unless node.include?('alias')
-
-          aliases = node['alias']
+          # Add aliases before checking whether target is already in group, otherwise they will
```
donoghuc: I dont think this is what we want. If the target is already in the group, all the config for that target should be ignored I think (including an alias entry).

spec/bolt/inventory_spec.rb
```diff
@@ -756,14 +756,17 @@ def common_data(transport)
...
         'config' => { 'ssh' => { 'disconnect-timeout' => 100 } },
         'facts' => { 'foo' => 'bar' }
+      }, {
+        'name' => 'foo',
+        'alias' => %w[bak]
       }] }
     }
 
     let(:inventory) { Bolt::Inventory.new(data) }
-    let(:target) { get_target(inventory, 'foo') }
```
donoghuc: Does `get_target` not work for this?

beechtom: No - I'm not entirely sure what's happening differently between `get_target` and `Inventory#get_targets`, but when a target shows up in another group it gets ignored and its alias doesn't get added to all of the groups that it's a part of. I can dig down further to try to figure out what the issue is.

lib/bolt/inventory/group2.rb
```diff
@@ -162,7 +162,8 @@ def data_merge(data1, data2)
           'config' => Bolt::Util.deep_merge(data1['config'], data2['config']),
           'name' => data1['name'] || data2['name'],
           'uri' => data1['uri'] || data2['uri'],
-          'alias' => data1['alias'] || data2['alias'],
+          # Collect all aliases across all groups for each target uri
+          'alias' => [*data1['alias'], *data2['alias']].sort,
```
donoghuc: I think we can use the set notation for union here for arrays `'alias' => data1['alias'] | data2['alias']`. Why do you sort? Is that just for testing? The aliases are guaranteed to be unique which is good, but I think this edge case is very strange. But i guess it makes sense, specifically if you re-create a target with target.new you destroy the target information and re-create. It seems that the only thing that is preserved is the alias. I think this is OK but it is a weird use case. Also luckily it appears to be the same behavior in a plan. 
```puppet
plan test(){
  Target.new('name' => 'foo', 'target_alias' => ['printed', 'boop'])
  out::message("${get_target('boop')}")
  out::message("${get_target('printed')}")
  Target.new('name' => 'foo', 'target_alias' => 'also_printed')
  out::message("${get_target('boop')}")
  out::message("${get_target('printed')}")
  out::message("${get_target('also_printed')}")
}
```
```
$ bolt plan run test
Starting: plan test
Starting: plan test
foo
foo
foo
foo
foo
Finished: plan test in 0.01 sec
Finished: plan test in 0.01 sec
```

beechtom: I don't think the set notation for union alone will work here, since `data1['alias']` and `data2['alias']` could be `nil`. We would need to do something like `data1['alias'].to_a | data2['alias'].to_a` for it to work -- not sure if this or the splat makes more sense.

The sort was just personal preference to have the aliases easy to read through, but I'm realizing we don't do that anywhere else, so it probably doesn't make sense.

beechtom: Oh, neat. Looks like we need to make sure both values are arrays before doing the set union for it to work, since either value can be `nil`.

I added the sort just out of a personal preference to have things ordered nicely... but I'm realizing we dont' do that anywhere else, so I'll remove it.

donoghuc: I'm starting to think this is a mistake. Perhaps with `Target.new`, we should clean up old aliases in inventory v2.

donoghuc: I think the technically correct thing to do would be to handle it like for example `config`. In the case:
```yaml
version: 2
groups:
  - name: foo 
    targets:
      - name: targ
        alias:
          - printed
          - boop
        config:
          transport: local
  - name: bar 
    targets:
      - name: targ
        alias: [also_printed]
        config:
          transport: winrm
```
I would expect alias to be [printed, boop] and config to be transport local.

beechtom: Updated to remove all existing alia when using `Target.new`.

CHANGELOG.md
```diff
@@ -33,6 +33,9 @@
 
   The `file::exists` and `file::readable` functions will no longer error when the file path is specified relative to a module and the file doesn't exist.
 
+* **`bolt inventory show -- detail` now displays all target aliases** ([#1379](https://github.com/puppetlabs/bolt/issues/1379))
```
donoghuc: Can you rebase and put this in a new `Bolt Next` section?

beechtom: Updated!

CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # Changelog
 
+## Bolt Next
+
+### Bug fixes
+
+* **`bolt inventory show -- detail` now displays all target aliases** ([#1379](https://github.com/puppetlabs/bolt/issues/1379))
```
donoghuc:
```suggestion
* **`bolt inventory show --detail` now displays all target aliases** ([#1379](https://github.com/puppetlabs/bolt/issues/1379))
```

Issue #1358, opened 2019-10-25
Add YAML plugin


adreyer:
## Use Case
It's useful to load structured data from multiple files into inventory.yaml. This allows the file to be split up or for user specific credentials to be stored outside the bolt project.

This plugin can also be used in a plugin tutorial to clarify how reference objects are just references to data that lives somewhere else.

## Describe the Solution You Would Like
A simple resolve_reference plugin that reads a yaml file and returns the results.

```
_plugin: yaml
path: my_data.yaml
```

## Additional Context

I started this in https://github.com/puppetlabs/puppetlabs-yaml

* Do we want to call the plugin `yaml` or is that too generic and conflict prone?
* Should we support using the task outside of the plugin context? What does that mean for boltdir? Currently it is relative to cwd in this case but I'm concerned that won't be reliable.
* As part of this ticket we should use the yaml plugin to more extensively document how references are resolved.
* We should rename the `path` parameter to `filepath` or `file` to make ti clear that it's not a key path in the yaml file.

PR #1417, opened 2019-11-14
(GH-1358) Track YAML plugin usage

donoghuc:
We forgot to iniclude yaml in the BUNDLED_CONTENT list for bundled plugins. This commit adds it.


donoghuc:
Sent data to Bolt Development google analytics, appears to work as expected. 
image

PR #1422, opened 2019-11-15
Check that names of targets returned from plugins are strings

beechtom:
This adds a check to ensure that the name of a target returned from a
plugin is a String before adding it to a group. Since plugins using the new
ruby plugin helper can populate target data with any type of data, setting 
the `name` or `uri` of a target as anything other than a String would raise 
an obscure error.

### DIFF
--- a/spec/bolt/inventory/group2_spec.rb
+++ b/spec/bolt/inventory/group2_spec.rb
@@ -210,6 +210,19 @@ describe Bolt::Inventory::Group2 do
     end
   end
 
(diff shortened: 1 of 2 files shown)


spec/bolt/inventory/group2_spec.rb
```diff
@@ -210,6 +210,19 @@
     end
   end
 
+  context 'where a target name or uri is not a string' do
```
nicklewis: This is just name

Issue #1408, opened 2019-11-12
Convert puppetdb plugin to use same logic as ruby plugin helper


lucywyman:
## Use Case
To ensure consistent behavior across plugins we should copy-paste the `apply_mapping`  and `required_config` functions into the PuppetDB plugin for it to use.

PR #1428, opened 2019-11-19
(GH-1408) Update puppetdb plugin to use target_mapping field

beechtom:
This updates the `puppetdb` inventory plugin to accept a
`target_mapping` field that is used to populate target configuration
values with facts loaded from PuppetDB. This is to match the behavior of
other inventory plugins that use the Ruby plugin helper.

> **Do not merge** until other inventory plugins have been updated to use the Ruby plugin helper

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@
 
   `bolt plan show` now uses Puppet Strings to parse plan documentation to show plan and parameter descriptions and parameter defaults.
 
+* **Added `target_mapping` field in `puppetdb` inventory plugin** ([#1408](https://github.com/puppetlabs/bolt/pull/1408))
+
+  The `puppetdb` inventory plugin has a new `target_mapping` field that accepts a hash of target configuration options and the facts to populate them with.
+
 ## Bolt 1.39.0
 
 ### New features
--- a/lib/bolt/plugin/puppetdb.rb
+++ b/lib/bolt/plugin/puppetdb.rb
@@ -11,7 +11,8 @@ module Bolt
         end
       end
 
-      TARGET_OPTS = %w[uri name config].freeze
+      TEMPLATE_OPTS = %w[uri name config].freeze
+      PLUGIN_OPTS = %w[_plugin query target_mapping].freeze
 
       def initialize(pdb_client)
         @puppetdb_client = pdb_client
@@ -45,8 +46,19 @@ module Bolt
         targets = @puppetdb_client.query_certnames(opts['query'])
         facts = []
 
-        target_opts = opts.select { |k, _| TARGET_OPTS.include?(k) }
-        Bolt::Util.walk_vals(target_opts) do |value|
+        template = opts.delete('target_mapping') || {}
+
+        keys = Set.new(TEMPLATE_OPTS) & opts.keys
+        unless keys.empty?
+          raise Bolt::ValidationError, "PuppetDB plugin expects keys #{keys.to_a} to be set under 'target_mapping'"
+        end
+
+        keys = Set.new(opts.keys) - PLUGIN_OPTS
+        unless keys.empty?
+          raise Bolt::ValidationError, "Unknown keys in PuppetDB plugin: #{keys.to_a}"
+        end
+
+        Bolt::Util.walk_vals(template) do |value|
           # This is done in parts instead of in place so that we only need to
           # make one puppetDB query
           if value.is_a?(String)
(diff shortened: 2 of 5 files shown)


lib/bolt/plugin/puppetdb.rb
```diff
@@ -45,8 +45,14 @@ def resolve_reference(opts)
         targets = @puppetdb_client.query_certnames(opts['query'])
         facts = []
 
-        target_opts = opts.select { |k, _| TARGET_OPTS.include?(k) }
-        Bolt::Util.walk_vals(target_opts) do |value|
+        template = opts.delete('target_mapping') || {}
+
+        keys = Set.new(TEMPLATE_OPTS) & opts.keys
+        unless keys.empty?
+          raise Bolt::ValidationError, "PuppetDB plugin expects keys #{keys} to be set under 'target_mapping'"
```
lucywyman: This is great! In addition to this, while we're here, erroring if any key besides `query` or `target_mapping` is present would also be useful. I think the main cause of the error will be if one of those options is misspelled, and especially for target_mapping running without your specified configuration could be dangerous (ie. running as root on accident)

beechtom: Updated!

lib/bolt/plugin/puppetdb.rb
```diff
@@ -45,8 +46,19 @@ def resolve_reference(opts)
...
```
lucywyman: @beechtom I don't particularly mind this but I suspect you might: printing sets in Ruby isn't the greatest:
```
lucy ~/githubs/bolt ((e07533e...)) $ be bolt command run 'hostname' -t all
Error executing plugin puppetdb from resolve_reference in puppetdb: PuppetDB plugin expects keys #<Set: {"name", "config"}> to be set under 'target_mapping'
```

You can `to_a` for slightly better printing:
```
lucy ~/githubs/bolt ((e07533e...)) $ be bolt command run 'hostname' -t all
Error executing plugin puppetdb from resolve_reference in puppetdb: PuppetDB plugin expects keys ["name", "config"] to be set under 'target_mapping'
```

lib/bolt/plugin/puppetdb.rb
```diff
@@ -45,8 +46,19 @@ def resolve_reference(opts)
...
+        end
+
+        keys = Set.new(opts.keys) - PLUGIN_OPTS
+        unless keys.empty?
+          raise Bolt::ValidationError, "Unknown keys in PuppetDB plugin: #{keys}"
```
lucywyman: Same `Set` printing issue here.

beechtom:
Updated.

spec/bolt/plugin/puppetdb_spec.rb
```diff
@@ -52,14 +52,39 @@
...
+    context "with unknown keys" do
+      let(:opts) do
+        { "query" => '',
+          "target_mapping" => {},
+          "bad_key" => '' }
+      end
+
+      it "raises a validation error" do
+        expect { plugin.resolve_reference(opts) }
+          .to raise_error(/Unknown keys in PuppetDB plugin:.*bad_key/m)
```
lucywyman: FWIW I don't think you need the `.*` here since the output should be static and doesn't look to be very long:
```
.to raise_error(/Unknown keys in PuppetDB plugin: \["bad_key"\]/)
```

spec/bolt/plugin/puppetdb_spec.rb
```diff
@@ -52,14 +52,39 @@
...
+
+    context "with misplaced config keys" do
+      let(:opts) do
+        { "query" => '',
+          "name" => 'facts.name_fact' }
+      end
+
+      it "raises a validation error" do
+        expect { plugin.resolve_reference(opts) }
+          .to raise_error(/PuppetDB plugin expects keys.*name/m)
```
lucywyman: Same thing with .* here

beechtom:
Updated.

CHANGELOG.md
```diff
@@ -8,6 +8,10 @@
 
   Parameter defaults can be set in the task metadata file and will be used if no value is supplied for the parameter.
 
+* **Added `target_mapping` field in `puppetdb` inventory plugin** ([#1408](https://github.com/puppetlabs/bolt/pull/1408))
```
lucywyman: I think your changelog might need a rebase + move. After that this is GtG!

beechtom:
Updated.

PR #1434, opened 2019-11-21
(GH-1418) Add remove_from_group plan function

### LINKED ISSUE #1418: Create `remove_from_group` function
## Use Case
Example: I have a group that I define into a variable $hosts using the get_targets function. I want to be able to do something that could attempt to ping, nslookup, whatever on the members of $hosts, and remove the non-responsive hosts from that group.

It could be something such as:
```
run_command(
  'uptime',
  $unresponsives,
  '_remove_from_group' => true,
)
``` 
^ that would remove the hosts that fail the command from the group.

I haven't considered many alternatives, maybe managing an array of hosts from the group within the plan.

## Additional Context
Will upload as I gather examples.

## Solution
* We should add a `remove_from_group` function that takes a single target and a group name and removes the target from the group
* If the target is not in the group it should succeed
* If the group is `'all'` it should fail
* If you try to remove a target from a group that does not exist it should fail
* The function should remove the target from all children of the group. It should not remove from the parent of the group. If you remove from a child group and the target is not independently defined in the parent group, then it should also not be available in the parent group (ie don't move a target into the parent group).

beechtom:
This adds a new `remove_from_group` plan function that allows you to
remove a target from an inventory group during plan execution. The
removed target is removed from all child groups and all parent groups,
except when the parent group explicitly defines the target.

Closes #1418

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,10 @@
 
   Parameter defaults can be set in the task metadata file and will be used if no value is supplied for the parameter.
 
+* **New `remove_from_group` plan function** ([#1418](https://github.com/puppetlabs/bolt/issues/1418))
+
+  The new plan function, `remove_from_group`, allows you to remove a target from an inventory group during plan execution.
+
 ### Bug fixes
 
 * **`bolt inventory show --detail` did not display all target aliases** ([#1379](https://github.com/puppetlabs/bolt/issues/1379))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@
 
   `bolt plan show` now uses Puppet Strings to parse plan documentation to show plan and parameter descriptions and parameter defaults.
 
+* **New `remove_from_group` plan function** ([#1418](https://github.com/puppetlabs/bolt/issues/1418))
+
+  The new plan function, `remove_from_group`, allows you to remove a target from an inventory group during plan execution.
+
 * **Added `target_mapping` field in `puppetdb` inventory plugin** ([#1408](https://github.com/puppetlabs/bolt/pull/1408))
 
   The `puppetdb` inventory plugin has a new `target_mapping` field that accepts a hash of target configuration options and the facts to populate them with.
@@ -19,10 +23,6 @@
 * **Task metadata can now specify parameter defaults** ([#1394](https://github.com/puppetlabs/bolt/pull/1394))
 
   Parameter defaults can be set in the task metadata file and will be used if no value is supplied for the parameter.
-
-* **New `remove_from_group` plan function** ([#1418](https://github.com/puppetlabs/bolt/issues/1418))
-
-  The new plan function, `remove_from_group`, allows you to remove a target from an inventory group during plan execution.
 
 ### Bug fixes
 
--- a/bolt-modules/boltlib/lib/puppet/functions/remove_from_group.rb
+++ b/bolt-modules/boltlib/lib/puppet/functions/remove_from_group.rb
@@ -14,7 +14,7 @@ Puppet::Functions.create_function(:remove_from_group) do
   # @example Remove Target from group.
   #   remove_from_group('foo@example.com', 'group1')
   # @example Remove failing Targets from the rest of a plan
-  #   $result = run_command(uptime, my_group)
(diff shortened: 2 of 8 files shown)


adreyer:
@branan @McdonaldSeanp this is a new plan language

branan:
@McdonaldSeanp I think we just need to add `remove_from_group` to our inventory2 stubs?

CHANGELOG.md
```diff
@@ -8,6 +8,10 @@
 
   Parameter defaults can be set in the task metadata file and will be used if no value is supplied for the parameter.
 
+* **New `remove_from_group` plan function** ([#1418](https://github.com/puppetlabs/bolt/issues/1418))
+
+  The new plan function, `remove_from_group`, allows you to remove a target from an inventory group during plan execution.
+
```
K8med: Should we comment on any hickups this may present? This copy is nice and straightforward, but if there are any unforeseen issues we might want to comment on what they are and how to avoid them?

lib/bolt/inventory/inventory2.rb
```diff
@@ -136,6 +136,18 @@ def expand_targets(targets)
       end
       private :expand_targets
 
+      def remove_target(current_group, target, desired_group)
+        if current_group.name == desired_group
+          current_group.remove_target(target)
+        end
+        current_group.groups.each do |child_group|
+          # If target was in current group, remove it from all child groups
+          desired_group = child_group.name if current_group.name == desired_group
```
nicklewis: This looks like it will only work for the first child, since after that we've overridden `desired_group`. Should we have a separate variable for this?

bolt-modules/boltlib/lib/puppet/functions/remove_from_group.rb
```diff
@@ -0,0 +1,38 @@
...
+# not been explicitly defined. A target cannot be removed from the 'all' group.
+#
+# **NOTE:** Not available in apply block
+Puppet::Functions.create_function(:remove_from_group) do
+  # @param target A pattern identifying a single target.
+  # @param group The name of the group to remove the target from.
+  # @example Remove Target from group.
+  #   remove_from_group('foo@example.com', 'group1')
+  # @example Remove failing Targets from the rest of a plan
+  #   $result = run_command(uptime, my_group)
```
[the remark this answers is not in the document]
beechtom: Added `_catch_errors`.

lib/bolt/inventory/group2.rb
```diff
@@ -142,6 +142,10 @@ def add_target_definition(target)
         @unresolved_targets[t_name] = target
       end
 
+      def remove_target(target)
+        @resolved_targets.delete(target.name)
```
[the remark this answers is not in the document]
beechtom: Makes sense to me - made sure to remove from `@unresolved_targets` as well.

lib/bolt/inventory/inventory2.rb
beechtom: Ah, good catch. Updated.

lib/bolt/inventory/inventory2.rb
```diff
@@ -136,6 +136,21 @@ def expand_targets(targets)
       end
       private :expand_targets
 
+      def remove_target(current_group, target, desired_group)
```
nicklewis: We need to call `target.invalidate_group_cache!` to ensure the config, etc for the target is updated next time it's used. Same as we do for `add_to_group`.

That also implies we need a test somewhere that the resolved values for a target correctly change after removing it from a group.

nicklewis: We also need whatever the equivalent of that is for inventory v1.

lib/bolt/inventory.rb
```diff
@@ -345,13 +346,18 @@ def set_plugin_hooks(tname, hash)
...
     private :set_plugin_hooks
 
-    def remove_node(current_group, target, desired_group)
+    def remove_node(current_group, target, desired_group, track = { 'all' => nil })
+      # Remove the target from the group
       if current_group.name == desired_group
         current_group.nodes.delete(target.name)
+      # If the target remains in the group, add the group data to the cache
+      else
+        add_group_data_to_cache(current_group, target, track)
```
nicklewis: We should only do this if the target is actually in the group. Otherwise this winds up adding the config from _every_ group to the target.

PR #1440, opened 2019-11-22
(GH-1427) Use environment variable for puppetdb config path in Windows

### LINKED ISSUE #1427: Hardcoded default Windows paths should not be used.
Windows can redirect folder/have a different system drive than `C:\ProgramData`

e.g.

https://github.com/puppetlabs/bolt/blob/871e92d52dca54eedb3d5f87900cc054dcb9f7fd/lib/bolt/puppetdb/config.rb#L13

https://github.com/puppetlabs/bolt/blob/871e92d52dca54eedb3d5f87900cc054dcb9f7fd/lib/bolt/puppetdb/config.rb#L18

Instead Bolt should be using either Environment Variables (e.g. `ALLUSERSPROFILE`) or API calls to get the known shell folders

https://docs.microsoft.com/en-us/previous-versions/windows/desktop/legacy/bb776911(v%3Dvs.85)

https://docs.microsoft.com/en-au/windows/win32/shell/knownfolderid?redirectedfrom=MSDN

Also, Bolt should be using the standard Windows locations for user configuration. storing files in dot-folder (`~/.puppetlabs`) causes issues in (typically enterprise) situations which use folder redirection, roaming profiles, or Remote Desktop services.

e.g. https://github.com/puppetlabs/bolt/blob/871e92d52dca54eedb3d5f87900cc054dcb9f7fd/lib/bolt/puppetdb/config.rb#L10-L11

Application data should be stored in `%LOCALAPPDATA%` for settings that should not roam across computers, and `%APPDATA%` for settings that should

beechtom:
This updates how Bolt locates the puppetdb config file on Windows.
Previously, the path was hardcoded as `C:\ProgramData\`, which could
cause issues if the user redirected the folder or had a different system
drive. Bolt will now use the `win32-dir` constant
`Dir::COMMON_APPDATA` to get the known folder.

Fixes #1427

### DIFF
--- a/lib/bolt/puppetdb/config.rb
+++ b/lib/bolt/puppetdb/config.rb
@@ -9,19 +9,22 @@ module Bolt
       if !ENV['HOME'].nil?
         DEFAULT_TOKEN = File.expand_path('~/.puppetlabs/token')
         DEFAULT_CONFIG = { user: File.expand_path('~/.puppetlabs/client-tools/puppetdb.conf'),
-                           global: '/etc/puppetlabs/client-tools/puppetdb.conf',
-                           win_global: 'C:/ProgramData/PuppetLabs/client-tools/puppetdb.conf' }.freeze
+                           global: '/etc/puppetlabs/client-tools/puppetdb.conf' }.freeze
       else
         DEFAULT_TOKEN = Bolt::Util.windows? ? 'nul' : '/dev/null'
         DEFAULT_CONFIG = { user: '/etc/puppetlabs/puppet/puppetdb.conf',
-                           global: '/etc/puppetlabs/puppet/puppetdb.conf',
-                           win_global: 'C:/ProgramData/PuppetLabs/client-tools/puppetdb.conf' }.freeze
+                           global: '/etc/puppetlabs/puppet/puppetdb.conf' }.freeze
 
+      end
+
+      def self.default_windows_config
+        require 'win32/dir'
+        File.expand_path(File.join(Dir::COMMON_APPDATA, 'PuppetLabs/client-tools/puppetdb.conf'))
       end
 
       def self.load_config(filename, options, boltdir_path = nil)
         config = {}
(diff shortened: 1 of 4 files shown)


nicklewis:
If I just try this `File.expand_path` in irb on a Windows host, it doesn't seem to expand anything. Is there some other setup I need to do (which I assume Bolt is doing)?

nicklewis:
I also just tried it through Bolt and it doesn't seem to work for me.

lib/bolt/puppetdb/config.rb
```diff
@@ -10,12 +10,16 @@ class Config
         DEFAULT_TOKEN = File.expand_path('~/.puppetlabs/token')
         DEFAULT_CONFIG = { user: File.expand_path('~/.puppetlabs/client-tools/puppetdb.conf'),
                            global: '/etc/puppetlabs/client-tools/puppetdb.conf',
-                           win_global: 'C:/ProgramData/PuppetLabs/client-tools/puppetdb.conf' }.freeze
+                           win_global: File.expand_path(
+                             '%CSIDL_COMMON_APPDATA%/PuppetLabs/client-tools/puppetdb.conf'
```
nicklewis: With the win32-dir gem, we can use `File.join(Dir::COMMON_APPDATA, 'PuppetLabs/client-tools/puppetdb.conf')`

lib/bolt/puppetdb/config.rb
```diff
@@ -2,6 +2,7 @@
 
 require 'json'
 require 'bolt/util'
+require 'win32/dir'
```
nicklewis: I think this require (and the `Dir::COMMON_APPDATA` reference) will only work when running on Windows. So we probably need to guard that, which might mean reorganizing the data structures below a bit.

nicklewis:
Annoyingly, this doesn't show a merge conflict but it probably ought to. This needs to be rebased for the CHANGELOG update (merging it now will put the entry under `1.40.0`).

beechtom:
Updated `CHANGELOG.md`

Issue #1409, opened 2019-11-12
Expose descriptions for plans and plan parameters with `bolt plan show`


adreyer:
## Use Case
Bolt displays descriptions from tasks metadata but does not have comparable support for plans. This makes it harder for users to understand how to run plans without looking at the source code.

## Describe the Solution You Would Like

When `bolt puppetfile install` should generate json data with strings that can be leveraged by `bolt plan show`.

This data should be in the same format as the current `bolt plan show` data but include descriptions of the parameters and plan itself. It should include parameter defaults as a string of puppet code.

## Describe Alternatives You've Considered
* Generating strings data when `bolt plan show` is run. This is considered to be too slow.

## Additional Context

* YAML plan support is out of scope.
* `bolt plan show --format=human` should match `bolt task show --format=human` whenever possible.
* data should live in `.strings_data/` each module should have it's own json file.
* If `.strings_data` is not present bolt should generate it when `bolt plan show` 
* Make sure the strings gem is packaged.
* Do not include the source code for the plan in the data file. 
* We should submit a PR to strings to automatically proccess plans.
* Whether the data is further restructured is an open question for the implementer. `bolt plan show --format json` should match the current json format.

PR #1442, opened 2019-11-22
(GH-1409) Show plan info using puppet strings

nicklewis:
Previously, the info we showed with `bolt plan show <plan>` was very basic
and only included some of the information that Puppet could load from the
plan object. We now use puppet strings to parse the docstrings for the plan
so we can show description for the plan and its parameters as well as
parameter defaults.

We also now show the same data for YAML plans, though not through Puppet 
strings.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Changelog
 
+## Bolt Next
+
+### New features
+
+* **`bolt plan show` shows plan and parameter descriptions** ([#1442](https://github.com/puppetlabs/bolt/pull/1442))
+
+  `bolt plan show` now uses Puppet Strings to parse plan documentation to show plan and parameter descriptions and parameter defaults.
+
 ## Bolt 1.39.0
 
 ### New features
--- a/lib/bolt/outputter/human.rb
+++ b/lib/bolt/outputter/human.rb
@@ -257,15 +257,18 @@ module Bolt
         plan_info = +""
         usage = +"bolt plan run #{plan['name']}"
 
-        plan['parameters']&.each do |name, p|
+        plan['parameters'].each do |name, p|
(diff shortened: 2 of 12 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # Changelog
 
+## Bolt Next
+
+### New features
+
+* **`bolt plan show` shows plan and parameter description** ([#1442](https://github.com/puppetlabs/bolt/pull/1442))
```
```suggestion
* **`bolt plan show` shows plan and parameter descriptions** ([#1442](https://github.com/puppetlabs/bolt/pull/1442))
```

lib/bolt/outputter/human.rb
```diff
@@ -259,10 +259,13 @@ def print_plan_info(plan)
...
+          pretty_params << "    Default: #{p['default_value']}\n" unless p['default_value'].nil?
+          pretty_params << "    #{p['description']}\n" if p['description']
           usage << (p.include?('default_value') ? " [#{name}=<value>]" : " #{name}=<value>")
         end
 
         plan_info << "\n#{plan['name']}"
+        plan_info << " - #{plan['description']}" if plan['description']
         plan_info << "\n\n"
         plan_info << "USAGE:\n#{usage}\n\n"
         plan_info << "PARAMETERS:\n#{pretty_params}\n" if plan['parameters']
```
beechtom: Since `pal` returns an empty hash, this value will always exist. To be consistent with the behavior of `task show` we should probably check if the value is empty or set it to `nil` if there aren't any parameters.

nicklewis: Updated to check for `empty?` (and changed the `&.` to `.` above)

beechtom:
Closed #1409

beechtom:
Closes #1409

PR #1447, opened 2019-11-26
(GH-1409) Add puppet-strings dependency

nicklewis:
This is now needed for `bolt plan show`.


nicklewis:
This needs https://github.com/puppetlabs/puppet-runtime/pull/254 merged first

PR #1452, opened 2019-11-29
(PE-27794) add a puma status app

tkishel:
Adds a status app, which can be removed when puma improves security in its status app.

### DIFF
--- a/lib/bolt_server/config.rb
+++ b/lib/bolt_server/config.rb
@@ -21,6 +21,7 @@ module BoltServer
     def defaults
       super.merge(
         'port' => 62658,
+        'status-port' => 62657,
         'concurrency' => 100,
         'cache-dir' => "/opt/puppetlabs/server/data/bolt-server/cache",
         'file-server-conn-timeout' => 120
--- a/config/transport_service_config.rb
+++ b/config/transport_service_config.rb
@@ -47,3 +47,24 @@ unless config['whitelist'].nil?
 end
 
 app impl
+
+#### Metrics/Status
+
+control_token = 'none'
+control_url_https = bind_addr.sub(config['port'].to_s, config['status-port'].to_s).sub('force_peer', 'none')
+
+# Start either the default Puma Control/Status app (if it supports our 'auth_actions' feature)
+# or start our custom Puma Status app via our custom 'stats' app/plugin (shipped in the Bolt gem).
+
+if defined? auth_actions
+  control_actions = 'gc-stats,stats'
+  activate_control_app control_url_https, auth_token: control_token, auth_actions: control_actions
+else
+  begin
+    plugin 'stats'
+    stats_url control_url_https
+    stats_token control_token
+  rescue UnknownPlugin
+    puts '* Status endpoint disabled'
+  end
+end
--- a/lib/bolt_server/base_config.rb
+++ b/lib/bolt_server/base_config.rb
@@ -6,8 +6,9 @@ require 'bolt/error'
 module BoltServer
   class BaseConfig
     def config_keys
-      %w[host port ssl-cert ssl-key ssl-ca-cert
-         ssl-cipher-suites loglevel logfile whitelist]
+      %w[host port status-port
+         ssl-cert ssl-key ssl-ca-cert ssl-cipher-suites
+         loglevel logfile whitelist]
     end
 
     def env_keys
--- a/lib/puma/README.md
+++ b/lib/puma/README.md
@@ -1,0 +1,5 @@
+#### puma-stats
+
+Add the plugin from: https://github.com/tkishel/puma-stats
+Remove if/after https://github.com/puma/puma/pull/2083 is merged.
+Requires Puma 4.3.x or newer.
--- a/lib/puma/plugin/stats.rb
+++ b/lib/puma/plugin/stats.rb
@@ -1,0 +1,43 @@
+# frozen_string_literal: true
+
+require 'puma/stats/dsl'
+
+Puma::Plugin.create do
+  def start(launcher)
+    str = launcher.options[:stats_url] || 'tcp://0.0.0.0:51209'
+
+    require 'puma/stats/app'
+    require 'puma/minissl/context_builder'
+    require 'puma/util'
+
+    app = Puma::Stats::App.new launcher
+    uri = URI.parse str
+
+    stats = Puma::Server.new app, launcher.events
+    stats.min_threads = 0
+    stats.max_threads = 1
+
+    case uri.scheme
+    when 'ssl'
+      optional_token = launcher.options[:stats_token] ? "with auth token: #{launcher.options[:stats_token]}" : ''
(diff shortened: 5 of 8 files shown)


tkishel:
Depends upon https://github.com/puppetlabs/pe-bolt-vanagon/pull/39

lib/bolt_server/config.rb
```diff
@@ -21,6 +21,8 @@ def int_keys
     def defaults
       super.merge(
         'port' => 62658,
+        'status-port' => 62659,
+        'status-token' => "bolt",
```
donoghuc: Is the token sensitive? Would we lay down one with pe? Can we use the ssl-{cert,key,ca-cert} for auth?

tkishel: Puma 4.3.0 implements that in the built-in Status app. We could too, if we model:

https://github.com/puma/puma/pull/2046
https://github.com/puma/puma/pull/2052

But I kept the logic of this app identical to the app in the version of puma we currently ship.

donoghuc: I think we need to update our puma stack anyway. If we got on 4.3.0 we would still need the puma-stats gem?

tkishel: Yes, because the Status app ...

https://github.com/puma/puma/blob/master/lib/puma/app/status.rb

... includes both Control and Status, which, in principle, I think is inappropriate: the security requirements for querying status (metrics) are significantly different that the same for controlling (restarting / shutting down) the service.

tkishel: Note that the metrics from Puppet Server do not require a client certificate/whitelist:

```
[root@pe-201921-replica-ha ~]# curl -k https://pe-201921-master.puppetdebug.vlan:8140/status/v1/services?level=debug 
{"puppet-profiler":{"service_version":"6.7.1","service_status_version":1,"detail_level":"debug","state" 
...
```

```
[root@pe-201921-master ~]# grep -rl pe-201921-replica-ha /etc/puppetlabs | wc -l
0
```

But I may be able to implement `ssl://` in the plugin ...

donoghuc: OK. I ticketed updating puma https://tickets.puppetlabs.com/browse/PE-27840

tkishel: As a POC, I added the `MiniSSL::ContextBuilder` class from the current version of Puma to the plugin (I could simply remove it if we upgrade to Puma 4.3+) and can serve status via:

```
if config['status-host'] && config['status-port'] && config['status-token']
  plugin 'stats'
  bind_addr = +"ssl://#{config['status-host']}:#{config['status-port']}?"
  bind_addr << "cert=#{config['ssl-cert']}"
  bind_addr << "&key=#{config['ssl-key']}"
  bind_addr << "&ca=#{config['ssl-ca-cert']}"
  bind_addr << "&verify_mode=none"
  bind_addr << "&ssl_cipher_filter=#{config['ssl-cipher-suites'].join(':')}"
  stats_url bind_addr
  stats_token config['status-token']
end
```

And access that via:
 
```
[root@pe-201921-master ~]# curl https://pe-201921-master.puppetdebug.vlan:44634/puma-stats?token=ace
curl: (60) Peer's certificate issuer has been marked as not trusted by the user.

[root@pe-201921-master ~]# curl --insecure https://pe-201921-master.puppetdebug.vlan:44634/puma-stats?token=ace
{ "backlog": 0, "running": 0, "pool_capacity": 10, "max_threads": 10 }

[root@pe-201921-master ~]# curl --cacert /etc/puppetlabs/puppet/ssl/certs/ca.pem https://pe-201921-master.puppetdebug.vlan:44634/puma-stats?token=ace
{ "backlog": 0, "running": 0, "pool_capacity": 10, "max_threads": 10 }  

[root@pe-201921-master ~]# curl --cacert /etc/puppetlabs/puppet/ssl/certs/ca.pem https://pe-201921-master.puppetdebug.vlan:44634/puma-stats?token=zzz
Invalid stats auth token
```

Is there a direction to move forward that you prefer/recommend?

tkishel:
Resulting in:

```
ACE Server: Running on Primary Master, https://pe-201921-master.puppetdebug.vlan:44634/puma-stats
Activity Service: Running on Primary Master, https://pe-201921-master.puppetdebug.vlan:4433/activity-api
Bolt Server: Running on Primary Master, https://pe-201921-master.puppetdebug.vlan:62659/puma-stats
Classifier: Running on Primary Master, https://pe-201921-master.puppetdebug.vlan:4433/classifier-api
Orchestrator: Running on Primary Master, https://pe-201921-master.puppetdebug.vlan:8143/orchestrator
PCP Broker: Running on Primary Master, wss://pe-201921-master.puppetdebug.vlan:8142/pcp
PCP Broker v2: Running on Primary Master, wss://pe-201921-master.puppetdebug.vlan:8142/pcp2
Puppet Server: Running on Primary Master, https://pe-201921-master.puppetdebug.vlan:8140/
PuppetDB: Running on Primary Master, https://pe-201921-master.puppetdebug.vlan:8081/pdb
RBAC: Running on Primary Master, https://pe-201921-master.puppetdebug.vlan:4433/rbac-api
2019-12-03 00:42:14 +0000
11 of 11 services are fully operational.
```

The connection is encrypted, but the client does not need to provide a certificate, same as with our other services. The token remains, to be parallel with the default Control/Status app.

lib/bolt_server/config.rb
donoghuc: I will defer to @branan or @McdonaldSeanp on that. I don't fully understand the impact of having to manage a token for this.

tkishel: Okay. Note that not specifying a token in `config/transport_service_config.rb` or `conf.d/bolt-server.conf` will disable the token requirement :)

branan:
If we're gonna maintain this puma stats thing, I'd prefer to pull it into our namespace and ensure it's a known gem to RE, etc.

But really, I think I'd prefer we get our version of Puma up-to-date, then try to upstream the ability to enable status without control. That way we're not maintaining anything :sunglasses:

tkishel:
I hear you ...

I've submitted a feature and pr to the puma repository to implement separate tokens for control and status actions in the (sigh) Control Server implemented as the Status app. If it is accepted, I can easily modify the associated PRs in our repositories to match.

https://github.com/puma/puma/issues/2081
https://github.com/puma/puma/pull/2082

That said, the puma-stats plugin/gem is a trivial fork of another (metrics) plugin that I would happily move into our namespace, and if we updated our Puma to 4.3.x or newer, would be almost identical to the implementation of the Status app (minus the control actions). An argument for implementing our own status code would be if we wanted to extend it beyond what puma offers.

tkishel:
Updated POC. Smaller footprint. Requires Puma 4.3.x

tkishel:
Closing in favor of the simpler https://github.com/puppetlabs/bolt/pull/1503

PR #1454, opened 2019-12-02
Added detailed example of puppetdb_query() usage

nmaludy:
Based on a conversation in Slack, i've added a more detailed example of how to use `puppetdb_query()` so new users understand that the plan function just returns the raw query results and needs to transform them into Targets to use them in `run_*()` functions.

### DIFF
--- a/documentation/bolt_connect_puppetdb.md
+++ b/documentation/bolt_connect_puppetdb.md
@@ -76,6 +76,20 @@ You can test your configuration with the following plan, which returns a list of
 
 ```
 plan pdb_test {
-  return(puppetdb_query("nodes[certname] {}"))
+  # query PuppetDB for a list of node certnames
+  # this returns an array of objects, each object containing a "certname" parameter:
+  # [ {"certname": "node1"}, {"certname": "node2"} ]
+  $query_results = puppetdb_query("nodes[certname] {}")
+  
+  # since puppetdb_query() returns the JSON results from the API call, we need to transform this
+  # data into Targets to use it in one of the run_*() functions.
+  # extract the "certname" values, so now we have an array of hostnames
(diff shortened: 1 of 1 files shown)


lucywyman:
I'll wait for @K8med to approve before merging.

lucywyman:
Update: @nmaludy mentioned wanting to adjust the example, so waiting on that.

nmaludy:
@lucywyman @K8med i've updated the documentation with another detailed example of using the `resolve_references()` function. I put these into a new section called "Practical Usage" in that same document. Thoughts on this?

documentation/bolt_connect_puppetdb.md
```diff
@@ -79,3 +79,49 @@ plan pdb_test {
   return(puppetdb_query("nodes[certname] {}"))
 }
 ```
+
+## Practical Usage
+
+In practice it is common to extract inventory from PuppetDB dynamically and use them
```
K8med:
```suggestion
In practice, it is common to extract inventory from PuppetDB dynamically to use
```

documentation/bolt_connect_puppetdb.md
```diff
@@ -79,3 +79,49 @@ plan pdb_test {
...
+in a plan. The following is an example of doing that using the `puppetdb_query()` function
```
K8med:
```suggestion
in a plan. The following is an example using the `puppetdb_query()` function
```

documentation/bolt_connect_puppetdb.md
```diff
@@ -79,3 +79,49 @@ plan pdb_test {
...
+directly. This method works, but requires data munging to be effective.
```
K8med:
```suggestion
directly. This method works but requires data munging to be effective.
```

PR #1460, opened 2019-12-03
(GH-1244) Serialize plan variables as pcore

### LINKED ISSUE #1244: Consistent Support for Plan Language objects from within `apply` blocks
## Use Case
Currently plan language objects are not correctly serialized/deserialized around `apply` blocks.

* `Target` objects and `Data` containing `Target` objects that are assigned to variables should work and be available inside apply blocks. 
* The inventory may (and likely should) be pruned to exclude targets that are not in scope.
* The `get_target(s)` functions should raise a clear error when called inside an apply block.
* `Target.new` should raise a clear error when called inside an apply block.
* `Error` objects assigned to a variable should be accessible inside apply blocks.
* `Result`, `ApplyResult` and `ResultSet` objects should either be available or raise a clear error when certain methods/attributes are accessed (for example: Result should work, but the `target` attribute should not work (performance)). The current behavior of converting them `Data` is a bug. 

* Removing support for these objects entirely. This makes it harder for users to understand apply blocks.
* Supporting `get_target(s)`. This method has side effects(adding new targets to the `all` group) and prevents us from optimizing how much of the inventory is loaded in each compiler.

## Additional Context
We should at least spike on making Pcore serialization/serialization work. This may be difficult for Target objects. Experiment with that and come up with a proposal for the first step. Perhaps we should serialize targets and inventory and pass them separately. 

Should other Pcore types like `Pattern` work across the boundary? Can we just use Pcore serialization with explicit tests for the Plan language objects?

Should new serialization/deserialization be feature flagged with the `version: 2` inventory? Yes, flag this with `$future`

lucywyman:
There are 4 custom types in the Bolt plan language that we want to make
available inside apply blocks: Target, Result, ResultSet, ApplyResult.
For each of the \*Result\* objects this is accomplished by serializing the
plan variables to the apply block as pcore instead of as json. We also
needed to add pcore_init methods to the implementation classes of the
objects, because the pcore deserializer will by default call
`initialize` on the class and pass in the values of the attributes of
the type, which doesn't match the signature of the actual
implementations. Using `_pcore_init_hash` and `_pcore_init_from_hash`
lets us define what data gets serialized and how it gets deserialized
for those types.

Target objects are trickier for a few reasons:
1. They must have an associate inventory
2. They should not be modified or created inside an apply block

Instead of trying to create the right guard rails around using Target
objects in apply blocks and pass around an inventory, it's easier to
think of (and implement) targets-in-apply as their own class, an
ApplyTarget. When creating the catalog compiler we set the inventory to
a bare string, then look for that in the Target object. When a Target
object is instantiated while the inventory is in 'apply mode', it will
create an ApplyTarget instead of a Target. For now this is only ever
done by the deserializer when target objects are deserialized. This
protects the inventory within apply blocks, and clearly separates the
concept of Targets in plans, and Targets in apply blocks.

Closes #1244

### DIFF
--- a/bolt-modules/boltlib/lib/puppet/datatypes/target.rb
+++ b/bolt-modules/boltlib/lib/puppet/datatypes/target.rb
@@ -3,6 +3,7 @@
 Puppet::DataTypes.create_type('Target') do
   begin
     inventory = Puppet.lookup(:bolt_inventory)
+
     inventory_version = inventory.version
     if inventory_version != 1
       target_implementation_class = inventory.target_implementation_class
@@ -10,7 +11,7 @@ Puppet::DataTypes.create_type('Target') do
   rescue Puppet::Context::UndefinedBindingError
     inventory_version = 1
   end
-  load_file('bolt/target')
+  require 'bolt/target'
 
   if inventory_version == 1
     interface <<-PUPPET
@@ -33,15 +34,15 @@ Puppet::DataTypes.create_type('Target') do
       attributes => {
         uri => { type => Optional[String[1]], kind => given_or_derived },
         name => { type => Optional[String[1]] , kind => given_or_derived },
+        safe_name => { type =>  Optional[String[1]], kind => given_or_derived },
         target_alias => { type => Optional[Variant[String[1], Array[String[1]]]], kind => given_or_derived },
         config => { type => Optional[Hash[String[1], Data]], kind => given_or_derived },
-        vars => { type => Optional[Hash[String[1], Data]], kind => given_or_derived  },
-        facts => { type => Optional[Hash[String[1], Data]], kind => given_or_derived  },
+        vars => { type => Optional[Hash[String[1], Data]], kind => given_or_derived },
+        facts => { type => Optional[Hash[String[1], Data]], kind => given_or_derived },
         features => { type => Optional[Array[String[1]]], kind => given_or_derived },
-        plugin_hooks => { type => Optional[Hash[String[1], Data]], kind => given_or_derived  }
+        plugin_hooks => { type => Optional[Hash[String[1], Data]], kind => given_or_derived }
       },
       functions => {
-        safe_name => Callable[[], String[1]],
         host => Callable[[], Optional[String]],
         password => Callable[[], Optional[String[1]]],
         port => Callable[[], Optional[Integer]],
--- a/lib/bolt/applicator.rb
+++ b/lib/bolt/applicator.rb
@@ -8,6 +8,7 @@ require 'open3'
 require 'bolt/error'
 require 'bolt/task'
 require 'bolt/apply_result'
+require 'bolt/apply_target'
 require 'bolt/util/puppet_log_level'
 
 module Bolt
@@ -129,6 +130,49 @@ module Bolt
       JSON.parse(out)
     end
 
+    def future_compile(target, catalog_input)
+      trusted = Puppet::Context::TrustedInformation.new('local', target.name, {})
+      catalog_input[:target] = {
+        name: target.name,
+        facts: @inventory.facts(target).merge('bolt' => true),
+        variables: @inventory.vars(target),
+        trusted: trusted.to_h
+      }
+      # rubocop:disable Style/GlobalVars
+      catalog_input[:future] = $future
+      # rubocop:enable Style/GlobalVars
+
+      bolt_catalog_exe = File.join(libexec, 'bolt_catalog')
+      old_path = ENV['PATH']
+      ENV['PATH'] = "#{RbConfig::CONFIG['bindir']}#{File::PATH_SEPARATOR}#{old_path}"
+      out, err, stat = Open3.capture3('ruby', bolt_catalog_exe, 'compile', stdin_data: catalog_input.to_json)
+      ENV['PATH'] = old_path
+
+      # stderr may contain formatted logs from Puppet's logger or other errors.
+      # Print them in order, but handle them separately. Anything not a formatted log is assumed
+      # to be an error message.
+      logs = err.lines.map do |l|
+        begin
+          JSON.parse(l)
+        rescue StandardError
+          l
+        end
+      end
+      logs.each do |log|
+        if log.is_a?(String)
+          @logger.error(log.chomp)
+        else
+          log.map { |k, v| [k.to_sym, v] }.each do |level, msg|
+            bolt_level = Bolt::Util::PuppetLogLevel::MAPPING[level]
+            @logger.send(bolt_level, "#{target.name}: #{msg.chomp}")
+          end
+        end
+      end
+
+      raise(ApplyError, target.name) unless stat.success?
+      JSON.parse(out)
+    end
+
     def validate_hiera_config(hiera_config)
       if File.exist?(File.path(hiera_config))
         data = File.open(File.path(hiera_config), "r:UTF-8") { |f| YAML.safe_load(f.read, [Symbol]) }
@@ -155,7 +199,6 @@ module Bolt
         options = args[1].map { |k, v| [k.sub(/^_/, '').to_sym, v] }.to_h
       end
 
-      # collect plan vars and merge them over target vars
       plan_vars = scope.to_hash(true, true)
       %w[trusted server_facts facts].each { |k| plan_vars.delete(k) }
 
@@ -179,11 +222,33 @@ module Bolt
     def apply_ast(raw_ast, targets, options, plan_vars = {})
       ast = Puppet::Pops::Serialization::ToDataConverter.convert(raw_ast, rich_data: true, symbol_to_string: true)
 
+      # rubocop:disable Style/GlobalVars
+      if $future
+        # Serialize as pcore for *Result* objects
+        plan_vars = Puppet::Pops::Serialization::ToDataConverter.convert(plan_vars,
+                                                                         rich_data: true,
+                                                                         symbol_as_string: true,
+                                                                         type_by_reference: true,
+                                                                         local_reference: false)
+        scope = {
+          code_ast: ast,
+          modulepath: @modulepath,
+          pdb_config: @pdb_client.config.to_hash,
+          hiera_config: @hiera_config,
+          plan_vars: plan_vars,
+          # This data isn't available on the target config hash
+          config: @inventory.config.transport_data_get
+        }
+      end
+      # rubocop:enable Style/GlobalVars
+
       r = @executor.log_action('apply catalog', targets) do
         futures = targets.map do |target|
           Concurrent::Future.execute(executor: @pool) do
             @executor.with_node_logging("Compiling manifest block", [target]) do
-              compile(target, ast, plan_vars)
+              # rubocop:disable Style/GlobalVars
+              $future ? future_compile(target, scope) : compile(target, ast, plan_vars)
+              # rubocop:enable Style/GlobalVars
             end
           end
         end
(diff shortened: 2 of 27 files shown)


.rubocop.yml
```diff
@@ -95,3 +95,7 @@ Lint/SuppressedException:
 # Enforce LF line endings, even when on Windows
 Layout/EndOfLine:
   EnforcedStyle: lf
+
+Lint/AmbiguousOperator:
```
adreyer: Can we just fix this by using parens?

background: I think optional method parens are one of ruby's worst design decisions.

lib/bolt/apply_target.rb
```diff
@@ -0,0 +1,25 @@
...
+module Bolt
+  class ApplyTarget
+    ATTRIBUTES ||= %i[uri name target_alias config vars facts features
+                      plugin_hooks safe_name].freeze
+    COMPUTED ||= %i[host password port protocol user].freeze
+
+    attr_reader *ATTRIBUTES
+    attr_accessor *COMPUTED
+
+    def self._pcore_init_from_hash(init_hash); end
```
adreyer: This should probably error in case it does end up being called somewhere.

.rubocop.yml
lucywyman: I don't think so? I'm not sure how rubocop expects the splat to be called, but this is what I get using parens:
```
be rubocop lib/bolt/apply_target.rb
Inspecting 1 file
W

Offenses:

lib/bolt/apply_target.rb:9:17: W: Lint/AmbiguousOperator: Ambiguous splat operator. Parenthesize the method arguments if it's surely a splat operator, or add a whitespace to the right of the * if it should be a multiplication.
    attr_reader *(ATTRIBUTES)
                ^
lib/bolt/apply_target.rb:9:18: C: Style/RedundantParentheses: Don't use parentheses around a constant.
    attr_reader *(ATTRIBUTES)
                 ^^^^^^^^^^^^
lib/bolt/apply_target.rb:10:19: W: Lint/AmbiguousOperator: Ambiguous splat operator. Parenthesize the method arguments if it's surely a splat operator, or add a whitespace to the right of the * if it should be a multiplication.
    attr_accessor *COMPUTED
```

This will at least only disable it for that file. I can see if there's a recommended syntax, or try turning the constants into class methods or something?

nicklewis:
```rb
    attr_reader(*ATTRIBUTES)
    attr_accessor(*COMPUTED)
```

lucywyman: https://github.com/rubocop-hq/rubocop/issues/2000

lib/bolt/apply_target.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+module Bolt
+  class ApplyTarget
+    ATTRIBUTES ||= %i[uri name target_alias config vars facts features
+                      plugin_hooks safe_name].freeze
+    COMPUTED ||= %i[host password port protocol user].freeze
+
+    attr_reader (*ATTRIBUTES)
+    attr_accessor (*COMPUTED)
```
nicklewis: No space between the method name and the `()`

lib/bolt/apply_target.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+module Bolt
+  class ApplyTarget
+    ATTRIBUTES ||= %i[uri name target_alias config vars facts features
```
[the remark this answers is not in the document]
nicklewis: It looks like this may be because we call `load_file` in the datatype. We can just use `require` there instead.

bolt-modules/boltlib/lib/puppet/datatypes/target.rb
```diff
@@ -4,16 +4,16 @@
   begin
     inventory = Puppet.lookup(:bolt_inventory)
 
-    inventory_version = inventory.version
-    if inventory_version != 1
+    inventory_name = inventory.name
```
nicklewis: I'm not sure what this buys us, since it looks like the only place we check for `v2` is in the CLI (where `apply` is known to not be possible) and it doesn't look like we check for `apply` anywhere. I was thinking of the version as the API version that the inventory conforms to (1 or 2) rather than an identifier about which inventory it is.

lib/bolt/applicator.rb
```diff
@@ -179,11 +230,37 @@ def count_statements(ast)
     def apply_ast(raw_ast, targets, options, plan_vars = {})
       ast = Puppet::Pops::Serialization::ToDataConverter.convert(raw_ast, rich_data: true, symbol_to_string: true)
 
+      # rubocop:disable Style/GlobalVars
+      if $future
+        %w[result resultset applyresult].each do |type|
+          Puppet.lookup(:loaders).private_environment_loader.load(:type, type)
```
nicklewis: I don't _think_ this should be necessary. If we have any instances of these types to serialize, doesn't that imply the type must have been loaded?

Is there a specific case where omitting this causes a problem?

lib/bolt/applicator.rb
```diff
@@ -129,6 +130,49 @@ def compile(target, ast, plan_vars)
       JSON.parse(out)
     end
 
+    def future_compile(target, catalog_input)
```
nicklewis: This differs from the `compile` method only in how it builds the `catalog_input`. Would it make sense to have like `build_catalog_input` / `build_future_catalog_input` and selectively call _those_ but then pass the result to a single `compile` method?

lib/bolt/applicator.rb
```diff
@@ -155,13 +199,20 @@ def apply(args, apply_body, scope)
...
-      # collect plan vars and merge them over target vars
       plan_vars = scope.to_hash(true, true)
       %w[trusted server_facts facts].each { |k| plan_vars.delete(k) }
 
       targets = @inventory.get_targets(args[0])
 
-      apply_ast(apply_body, targets, options, plan_vars)
+      # rubocop:disable Style/GlobalVars
+      if $future
+        Puppet.override(bolt_inventory: 'apply') do
```
nicklewis: What does this override do?

lib/bolt/catalog.rb
```diff
@@ -65,23 +68,45 @@ def compile_catalog(request)
...
+                end
           Puppet.override(bolt_pdb_client: pdb_client,
-                          bolt_inventory: setup_inventory(request['inventory'])) do
+                          bolt_inventory: inv) do
             Puppet.lookup(:pal_current_node).trusted_data = target['trusted']
             pal.with_catalog_compiler do |compiler|
+              if request['future']
+                # This needs to happen inside the catalog compiler so loaders are initialized for loading
+                vars = Puppet::Pops::Serialization::FromDataConverter.convert(request['plan_vars'])
+                pal.send(:add_variables, compiler.send(:topscope), vars.merge(target['vars']))
```
nicklewis: We call it `target['vars']` in future but `target['variables']` in the old version. I see that that works around the fact that we would otherwise be setting the target variables earlier (`Puppet::Pal.in_tmp_environment`) and this avoids a conflict. There must be a clearer way to handle that though.

lib/bolt/catalog.rb
```diff
@@ -65,23 +68,45 @@ def compile_catalog(request)
...
               # to compile most Puppet 4+ manifests, so we default to allowing deprecated functions.
               Puppet[:strict] = options['strict'] || :warning
               Puppet[:strict_variables] = options['strict_variables'] || false
               ast = Puppet::Pops::Serialization::FromDataConverter.convert(pal_main)
-              compiler.evaluate(ast)
+              # CODEREVIEW: Is there a better way to catch this?
+              begin
+                compiler.evaluate(ast)
+              rescue Puppet::PreformattedError => e
+                if e.message =~ /Initializer for class Bolt::ApplyTarget does not match the attributes of Target/
```
nicklewis: We can make a `self.from_asserted_args` method on `ApplyTarget` that _does_ match the attributes of `Target` (so it will be called instead of `initialize`) and have it raise the error we want. We will also want `self.from_asserted_hash` to handle the hash case.

lib/bolt/result.rb
```diff
@@ -73,6 +73,24 @@ def self.from_asserted_args(target, value)
       new(target, value: value)
     end
 
+    def self._pcore_init_from_hash
+      raise "Result shouldn't be instantiated from a pcore_init class method. How did this get called?"
+    end
+
+    def _pcore_init_from_hash(init_hash)
+      opts = init_hash.reject { |k, _v| k == 'target' }
+      initialize(init_hash['target'], opts.transform_keys(&:to_sym))
```
nicklewis: We can't use `transform_keys` in Ruby 2.3. Use `Bolt::Util.walk_keys(opts, &:to_sym)` instead. This is also the case on `ApplyResult`.

lib/bolt/apply_result.rb
```diff
@@ -85,6 +85,21 @@ def self.from_task_result(result)
       end
     end
 
+    def self._pcore_init_from_hash
+      raise "ApplyResult shouldn't be instantiated from a pcore_init class method. How did this get called?"
+    end
+
+    def _pcore_init_from_hash(init_hash)
```
nicklewis: I think we can just inherit the `Result` version of this method?

lib/bolt/apply_inventory.rb
```diff
@@ -0,0 +1,89 @@
+# frozen_string_literal: true
+
+require 'bolt/boltdir'
+require 'bolt/config'
+require 'bolt/error'
+
+module Bolt
+  class ApplyInventory
```
nicklewis: I like this class a lot. :+1:

lib/bolt/apply_inventory.rb
```diff
@@ -0,0 +1,89 @@
...
+      raise InvalidFunctionCall, 'get_target'
+    end
+
+    # rubocop:disable Naming/AccessorMethodName
+    def set_var(*_params)
+      raise InvalidFunctionCall, 'set_var'
+    end
+
+    def set_feature(*_params)
+      raise InvalidFunctionCall, 'set_features'
```
nicklewis:
```suggestion
      raise InvalidFunctionCall, 'set_feature'
```

lib/bolt/apply_inventory.rb
```diff
@@ -0,0 +1,89 @@
...
+      @config_hash = config_hash
+      @targets = {}
+    end
+
+    def create_apply_target(target)
+      @targets[target.name] = target
+    end
+
+    def validate
+      @groups.validate
```
nicklewis: I assume this method isn't necessary, given that its current implementation will error (this class has no instance variable called `@groups`).

bolt-modules/boltlib/lib/puppet/datatypes/target.rb
lucywyman: Hm that's a good point - now that we have an apply inventory object with the target_implementation_class method this isn't necessary.

lib/bolt/catalog.rb
lucywyman: I thought about that, but was worried that that would dredge up the conversation about why `from_asserted_*` doesn't work...I think that's a more reliable way of raising this error though.

lib/bolt/applicator.rb
lucywyman: My thinking here was that we'll shortly be removing the `compile` method wholesale, so keeping them separate rather than abstracting parts out would make the 2.0 transition a bit easier.

nicklewis:
Actually, it also needs a changelog entry.

lucywyman:
Updated - I also didn't see that we mentioned this in the documentation at all, but should this be documented or have an example somewhere?? CC @hestonhoffman

nicklewis:
I think this mostly just makes it work the way it intuitively should, so I'm not sure what there is to document about it.

PR #1461, opened 2019-12-03
(GH-1446) Use second positional argument of run_plan function to spec…

### LINKED ISSUE #1446: Populate a plan's `$targets` parameter with the second positional argument of the `run_plan` plan function
## Use Case
The `run_plan` plan function currently populate's a plan's `$nodes` parameter when passing targets as the second positional argument. With bundled content using a `$targets` parameter instead, we should have `run_plan` populate this parameter with targets passed as the second positional argument.

* If a plan has a `$targets` parameter and is invoked with `run_plan('myplan', $targets)` it should populate the targets parameter
* If a plan has a `$targets` parameter and is invoked with `run_plan('myplan', $targets, 'targets'=>$targets)` then the invocation should error.
* If a plan has a `$nodes` parameter and is invoked with `run_plan('myplan', $targets)` it should not raise a warning
* If a plan has both a `$nodes` and `$targets` and is invoked with `run_plan('myplan', $targets)` it should populate the nodes parameter and raise a warning that this will error in 2.0.
* If a plan has both a `$nodes` and `$targets` AND `$future` is true and is run with `run_plan('myplan, $targets)` then the invocation should error.

## Additional Context
* The CLI currently looks at a plan's parameters and will populate either `$nodes` or `$targets` if one is present, or populate neither if both are present. This behavior is not acceptable for the plan language API.
* Blocks #1376

beechtom:
…ify $targets parameter

This adds support for using the second positional argument of the
`run_plan` plan function to specify a plan's `TargetSpec $targets`
parameter.

When a plan has both a `$nodes` and `$targets` parameter, the
`run_plan` plan function will do the following.

 * If `future` is set, the plan will fail when there is a `$nodes`
    and `$targets` parameter and the second positional argument is used.
 * If `future` is unset, the second positional argument will specify
    the `$nodes` parameter. A deprecation warning will also be logged.

 When a plan has only one of `$nodes` or `$targets` as parameters, then
 the second positional argument will specify that parameter. The plan
 will also fail if the second positional argument is used to specify one
 of these parameters and the arguments hash specifies the same
 parameter.

Closes #1446

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Changelog
 
+## Bolt Next
+
+### New features
+
+* **`run_plan` plan function will specify a plan's `$targets` parameter using the second positional argument** ([#1446](https://github.com/puppetlabs/bolt/issues/1446))
+
+  When running a plan with a `$targets` parameter with the `run_plan` plan function, the second positional argument can be used to specify the `$targets` parameter. If a plan has a `$nodes` parameter, the second positional argument will only specify the `$nodes` parameter.
+
 ## Bolt 1.42.0
 
 ### New features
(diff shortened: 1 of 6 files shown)


beechtom:
Rebased on master.

beechtom:
Rebased on master.

PR #1466, opened 2019-12-05
(feat) Add script-dir option to ssh transport

logicminds:
* In some cases you need to be able to predict where bolt
    will write the script file to. Previously, this was dictated
    by the value of tmpdir and a SecureRandom id.

  * This adds a new script-dir option that allows the user to supply
    the directory where the script will be placed.

    ```
    ---
    ssh:
       script-dir: 'bolt_script_dir'
       tmpdir: /home/vagrant
    ```
    would end up creating the script file under
      /home/vagrant/bolt_script_dir

    By default the previous behavior of creating a unique directory
    per script run is preserved.

### DIFF
--- a/lib/bolt/transport/sudoable/connection.rb
+++ b/lib/bolt/transport/sudoable/connection.rb
@@ -7,6 +7,7 @@ module Bolt
     class Sudoable < Base
       class Connection
         attr_accessor :target
+
         def initialize(target)
           @target = target
           @run_as = nil
@@ -38,7 +39,8 @@ module Bolt
 
         def make_tempdir
           tmpdir = @target.options.fetch('tmpdir', '/tmp')
-          tmppath = "#{tmpdir}/#{SecureRandom.uuid}"
+          script_dir = @target.options.fetch('script-dir', SecureRandom.uuid)
+          tmppath = File.join(tmpdir, script_dir)
           command = ['mkdir', '-m', 700, tmppath]
 
           result = execute(command)
--- a/spec/integration/transport/ssh_spec.rb
+++ b/spec/integration/transport/ssh_spec.rb
@@ -470,14 +470,21 @@ describe Bolt::Transport::SSH do
     end
   end
 
-  context "user supplied script directory" do
+  context "with specific tempdir using script-dir option" do
+    let(:script_dir) { "123456" }
     let(:config) do
-      mk_config("host-key-check" => false, "sudo-password" => password, "run-as" => "root",
-                user: user, password: password, "script-dir" => "123456", interpreters: { sh: "/bin/sh" })
+      mk_config("host-key-check" => false, "sudo-password" => password,
+                "run-as" => "root", user: user, password: password,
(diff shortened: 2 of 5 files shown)


spec/integration/transport/ssh_spec.rb
```diff
@@ -469,4 +469,15 @@ def make_target(host_: hostname, port_: port, conf: config)
...
   end
+
+  context "user supplied script directory" do
+    let(:config) do
+      mk_config("host-key-check" => false, "sudo-password" => password, "run-as" => "root",
+                user: user, password: password, "script-dir" => "123456", interpreters: { sh: "/bin/sh" })
+    end
+    let(:target) { make_target }
+    it "executes a command on a host", ssh: true do
+      expect(ssh.run_command(target, "pwd")["stdout"]).to eq("/home/bolt\n")
```
lucywyman: `run_command` doesn't actually create a tempdir to execute in (https://github.com/puppetlabs/bolt/blob/master/lib/bolt/transport/sudoable.rb#L20 - would have `conn.with_tempdir` if it used a tempdir). You can make a temporary script that prints the working directory and use that to verify that the correct directory is used:

```
  context "user supplied script directory" do
    let(:script_dir) { "123456" }
    let(:config) do
      mk_config("host-key-check" => false, "sudo-password" => password,
                "run-as" => "root", user: user, password: password,
                "script-dir" => script_dir, interpreters: { sh: "/bin/sh" })
    end 
    let(:target) { make_target(conf: config) }

    it "executes a command on a host", ssh: true do
      cmd = 'cd $( dirname $0) && pwd'
      with_tempfile_containing('dir', cmd, '.sh') do |script|
        result = ssh.run_script(target, script.path, nil)
        expect(result.value['stdout']).to eq("/tmp/123456\n")
      end 
    end 
  end 
```

lucywyman: You'll also want to pass your config to the `make_target` call, like above.

lib/bolt/transport/sudoable/connection.rb
```diff
@@ -38,7 +39,8 @@ def make_executable(path)
 
         def make_tempdir
           tmpdir = @target.options.fetch('tmpdir', '/tmp')
-          tmppath = "#{tmpdir}/#{SecureRandom.uuid}"
+          script_dir = @target.options.fetch('script-dir', SecureRandom.uuid)
```
lucywyman: I think we should also add some validation to this method: https://github.com/puppetlabs/bolt/blob/master/lib/bolt/transport/ssh.rb#L29 to validate that the option isn't empty. It should error there if it's an empty string.

logicminds:
@lucywyman pushing those changes would be awesome.  The bolt test setup is overwhelming.  I should have called this out that I wasn't really verifying the directory was being generated.  Instead I was just verifying that no error was returned. I did smoke test this on my own system using the puppet_agent::version task but I couldn't verify the directory was being created.  

Would be helpful to know what conditions the script-dir and tmpdir options are actually used under.

PR #1468, opened 2019-12-05
(maint) Shrink Bolt logo on docs homepage

nicklewis:
max-width isn't an img attribute, but width is.


nicklewis:
I don't know whether the problem is that attributes simply don't make it onto the website or that max-width is invalid so it was omitted, so I'm not sure if this will actually fix anything. I will check it out in the docs preview if I can make that work.

Setting an explicit width is not ideal for scaling, but on the other hand if it works at all then it's an improvement.

Issue #1425, opened 2019-11-18
Being prompted twice for `password` and `sudo-password` when they are the same value is annoying


lucywyman:
## Use Case
Users often want to use the same value for `password` and `sudo-password`, and want to use the `prompt` plugin to get the value. Right now that means using the `prompt` plugin in 2 places, and having to enter the same value twice, which is annoying.

## Describe the Solution You Would Like
I have a few idea for how we could solve this:
1. Provide a config option to use the `password` value for `sudo-password`, if sudo-password is needed.
2. Create a new plugin or sigil that will load another config value as the value for that option:
```
password:
  _plugin: prompt
  message: Please enter your password:
sudo-password:
  _plugin: reuse
  config_option: password
```
3. Add a `name` attribute to the `prompt` plugin and only prompt once for each named prompt. This would mean that a prompt could be used throughout the inventory.

## Additional Context
Slack convos: [1](https://puppetcommunity.slack.com/archives/C7GP57T50/p1574081783016800), [2](https://puppetcommunity.slack.com/archives/C7GP57T50/p1574089414019700), [3](https://puppetcommunity.slack.com/archives/C7GP57T50/p1574092911025100)

## Solution
The `sudo-password` should default to the value for `password` in all cases. This should be gated on future.

adreyer:
I don't think it's possible to create a new plugin that references other values without opening up a lot of unwanted complexity to the plugin API.

I do think we could allow named references as a general extension to the resolve_reference functionality. Given that references are lazy and can be recursive it would be easy for users to create broken inventories that are difficult to debug but it should be possible. 

I'm open to suggestions but the best way I can think to implement this is probably a new `_reference`(bikeshed) special key. 

```yaml
- config:
   password:
     _plugin: prompt
     _reference: password
    message: "enter passworf
  sudopassword:
    _reference: password
```

 - A better interface for this would allow users to define references outside any section of the inventory and then use them anywhere. 
- It makes sense intuitively for references to be scopeable to a section but I suspect this will make the error cases too.

nmaludy:
It would be really nice to reference other config values, without it we have a lot of copy/paste in our inventory files:

```
groups:
  - name: windows
    config:
      transport: winrm
      winrm:
        user: patching_svc@domain.tld
        password: xxx
        ssl: true
        ssl-verify: false
    groups:
      - name: windows_servers_a
        vars:
          patching_order: 1
        targets:
          - _plugin: wsus_inventory
[...]

PR #1469, opened 2019-12-06
(GH-1425) Set default value for sudo-password to value for password

beechtom:
This sets the default value for `sudo-password` to the same value as
`password`. This feature is gated by `future`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@
 
   Bolt packages are now available for Debian 10.
 
+* **SSH transport sets `sudo-password` to the same value as `password` by default** ([#1425](https://github.com/puppetlabs/bolt/issues/1425))
+
+  If `sudo-password` is not set when using `run-as`, Bolt will set the value of `sudo-password` to match the value of `password`. This behavior is gated on the future config option, and will be available by default in Bolt 2.0.
+  
 ### Bug fixes
 
 * **Default PuppetDB config lookup used hardcoded path in Windows** ([#1427](https://github.com/puppetlabs/bolt/pull/1427))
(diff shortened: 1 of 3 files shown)


lib/bolt/transport/ssh/connection.rb
```diff
@@ -148,10 +148,15 @@ def disconnect
 
         def handled_sudo(channel, data, stdin)
           if data.lines.include?(Sudoable.sudo_prompt)
+            # rubocop:disable Style/GlobalVars
             if target.options['sudo-password']
               channel.send_data("#{target.options['sudo-password']}\n")
               channel.wait
               return true
+            elsif target.options['password'] && $future
```
nicklewis: It might make sense to set this in the initializer, like `@sudo_password = target.options['sudo-password'] || target.options['password']` (gating that second part on `$future`). Then here we can just check in one place for whether we have `@sudo_password`.

beechtom: Updated.

CHANGELOG.md
```diff
@@ -8,6 +8,10 @@
 
   Bolt's help text has been reformatted to be more consistent with the formatting in other Puppet tools.
 
+* **SSH transport sets `sudo-password` to the same value as `password` by default** ([#1425](https://github.com/puppetlabs/bolt/issues/1425))
+
+  If `sudo-password` is not set when using `run-as`, Bolt will set the value of `sudo-password` to match the value of `password`. It is gated on the future config option, and will be available by default in Bolt 2.0.
```
```suggestion
  If `sudo-password` is not set when using `run-as`, Bolt will set the value of `sudo-password` to match the value of `password`. It is gated on the future configuration option and will be available by default in Bolt 2.0.
```

K8med: It's not clear what the "it" is in the beginning of the second sentence. Does "It is gated on the future configuration option" mean `sudo-password` will be gated on the future configuration option? If possible can we identify the "it"? Thanks!

beechtom: Updated to say "this behavior" instead.

K8med: Perfect.

PR #1472, opened 2019-12-06
(GH-1439) Add sudo-executable transport config option

### LINKED ISSUE #1439: Support dzdo with a password for privilege escalation
## Use Case
dzdo appears to support the flags we use to escalate with a sudo password.

A `sudo-executable` option that changes just the command not any of the sudo flags?

## Additional Context

We need a dzdo user to test this behavior for us before we release  it.

## Solution
Add a `sudo-executable` configuration option to any transports that support `run-as-command` which will replace `sudo` in the escalation command with whatever the user provides. This should be marked as experimental in the docs.

beechtom:
This adds support for a `sudo-executable` transport config
option. When using `run-as`, the `sudo-executable` option will specify
the executable to use when running as another user. It will run the
executable with the same options as sudo.

This feature is experimental.

Closes #1439

### DIFF
--- a/documentation/bolt_configuration_options.md
+++ b/documentation/bolt_configuration_options.md
@@ -50,6 +50,7 @@ ssh:
 -   `proxyjump`: A jump host to proxy SSH connections through, and an optional user to connect with, for example: jump.example.com or user1@jump.example.com.
 -   `run-as`: A different user to run commands as after login.
 -   `run-as-command`: The command to elevate permissions. Bolt appends the user and command strings to the configured run as a command before running it on the target. This command must not require an interactive password prompt, and the `sudo-password` option is ignored when `run-as-command` is specified. The run-as command must be specified as an array.
+-   `sudo-executable`: The executable to use when escalating to the configured `run-as` user. This is useful when you want to escalate using the configured `sudo-password`, since `run-as-command` does not use `sudo-password` or support prompting. The command executed on the target is `<sudo-executable> -S -u <user> -p custom_bolt_prompt <command>`. **Note:** This option is experimental.
 -   `sudo-password`: Password to use when changing users via `run-as`.
 -   `tmpdir`: The directory to upload and execute temporary files on the target.
 -   `script-dir`: The subdirectory of the tmpdir to use in place of a randomized subdirectory for
@@ -161,6 +162,7 @@ When using the SSH transport, Bolt also interacts with the ssh-agent for SSH key
 
 -   `run-as`: A different user to run commands as after login.
 -   `run-as-command`: The command to elevate permissions. Bolt appends the user and command strings to the configured run as a command before running it on the target. This command must not require an interactive password prompt, and the `sudo-password` option is ignored when `run-as-command` is specified. The run-as command must be specified as an array.
+-   `sudo-executable`: The executable to use when escalating to the configured `run-as` user. This is useful when you want to escalate using the configured `sudo-password`, since `run-as-command` does not use `sudo-password` or support prompting. The command executed on the target is `<sudo-executable> -S -u <user> -p custom_bolt_prompt <command>`. **Note:** This option is experimental.
 -   `sudo-password`: Password to use when changing users via `run-as`.
 -   `tmpdir`: The directory to copy and execute temporary files.
 
--- a/lib/bolt/transport/local/shell.rb
+++ b/lib/bolt/transport/local/shell.rb
@@ -126,7 +126,8 @@ module Bolt
 
           if escalate
             if use_sudo
-              sudo_flags = ["sudo", "-k", "-S", "-u", run_as, "-p", Sudoable.sudo_prompt]
+              sudo_exec = target.options['sudo-executable'] || "sudo"
+              sudo_flags = [sudo_exec, "-k", "-S", "-u", run_as, "-p", Sudoable.sudo_prompt]
               sudo_flags += ["-E"] if options[:environment]
               sudo_str = Shellwords.shelljoin(sudo_flags)
             else
--- a/lib/bolt/transport/ssh/connection.rb
+++ b/lib/bolt/transport/ssh/connection.rb
@@ -194,10 +194,10 @@ module Bolt
           use_sudo = escalate && @target.options['run-as-command'].nil?
 
           command_str = inject_interpreter(options[:interpreter], command)
-
           if escalate
             if use_sudo
-              sudo_flags = ["sudo", "-S", "-u", run_as, "-p", Sudoable.sudo_prompt]
+              sudo_exec = target.options['sudo-executable'] || "sudo"
+              sudo_flags = [sudo_exec, "-S", "-u", run_as, "-p", Sudoable.sudo_prompt]
               sudo_flags += ["-E"] if options[:environment]
               sudo_str = Shellwords.shelljoin(sudo_flags)
             else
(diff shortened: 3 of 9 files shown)


lib/bolt/transport/ssh/connection.rb
```diff
@@ -194,10 +194,10 @@ def execute(command, sudoable: false, **options)
           use_sudo = escalate && @target.options['run-as-command'].nil?
 
           command_str = inject_interpreter(options[:interpreter], command)
-
           if escalate
             if use_sudo
-              sudo_flags = ["sudo", "-S", "-u", run_as, "-p", Sudoable.sudo_prompt]
+              sudo_exec = target.options['sudo-executable'] || "sudo"
+              sudo_flags = [sudo_exec, "-S", "-u", run_as, "-p", Sudoable.sudo_prompt]
```
lucywyman: I *think* we can add a test to `spec/integration/transport/ssh_spec.rb` along the lines of:
```
  context "with sudo as non-root", sudo: true do
    let(:config) {
      mk_config('host-key-check' => false, 'sudo-executable' => 'fake',
                'run-as' => user, user: bash_user, password: bash_password)
    }   
    let(:target) { make_target }

    it 'uses the correct executable' do
      expect_any_instance_of(Net::SSH::Connection::Channel).to receive(:exec)
        .with("fake -S -u bolt -p \\[sudo\\]\\ Bolt\\ needs\\ to\\ run\\ as\\ another\\ user,\\ password:\\  whoami", any_args)

      ssh.run_command(target, 'whoami')
    end 
  end 
```

This currently hangs for me though, I'm not quite sure how to handle the block param to `exec`.

lucywyman: Ok, finally figured out what was hanging :P 
```
context "with sudo as non-root", sudo: true do
    let(:config) {
      mk_config('host-key-check' => false, 'sudo-executable' => 'fake',
                'run-as' => user, user: bash_user, password: bash_password)
    }   
    let(:target) { make_target }

    it 'uses the correct executable' do
      allow_any_instance_of(Net::SSH::Connection::Channel).to receive(:wait).and_return('')
      expect_any_instance_of(Net::SSH::Connection::Channel).to receive(:exec)
        .with("fake -S -u bolt -p \\[sudo\\]\\ Bolt\\ needs\\ to\\ run\\ as\\ another\\ user,\\ password:\\  whoami")

      ssh.run_command(target, 'whoami')
    end 
  end 
```

documentation/bolt_configuration_options.md
```diff
@@ -50,6 +50,7 @@ ssh:
 -   `proxyjump`: A jump host to proxy SSH connections through, and an optional user to connect with, for example: jump.example.com or user1@jump.example.com.
 -   `run-as`: A different user to run commands as after login.
 -   `run-as-command`: The command to elevate permissions. Bolt appends the user and command strings to the configured run as a command before running it on the target. This command must not require an interactive password prompt, and the `sudo-password` option is ignored when `run-as-command` is specified. The run-as command must be specified as an array.
+-   `sudo-executable`: The executable to use to run as another user when using `run-as`. **Note:** This option is experimental.
```
lucywyman: I think it'd be useful to explain, either here or somewhere else in the docs, that this is in order to support using the executable with a password. You could use `run-as-command` for *most* of the functionality, except escalating with a password.

lucywyman: `run-as` uses `sudo -S -u #{user} -p custom_prompt` then looks for the prompt to pass in the password if needed. `run-as-command` replaces that whole string with whatever you give it, and does not support using a password over stdin. `sudo-executable` only replaces the `sudo` part of the command, so would run `your_cmd -S -u #{user} -p custom_prompt`.

documentation/bolt_configuration_options.md
```diff
@@ -50,6 +50,7 @@ ssh:
 -   `proxyjump`: A jump host to proxy SSH connections through, and an optional user to connect with, for example: jump.example.com or user1@jump.example.com.
 -   `run-as`: A different user to run commands as after login.
 -   `run-as-command`: The command to elevate permissions. Bolt appends the user and command strings to the configured run as a command before running it on the target. This command must not require an interactive password prompt, and the `sudo-password` option is ignored when `run-as-command` is specified. The run-as command must be specified as an array.
+-   `sudo-executable`: The executable to use to run as another user when using `run-as`. This is used to replace the executable used when escalating with a password is desired, as the `run-as-command` option does not support escalating with a password. The command run on the target is `$sudo-executable -S -u $user -p $sudo-password`. **Note:** This option is experimental.
```
lucywyman:
```suggestion
-   `sudo-executable`: The executable to use when escalating to the configured `run-as` user. This is useful when you want to escalate using the configured `sudo-password`, since `run-as-command` does not use `sudo-password` or support prompting. The command executed on the target is `<sudo-executable> -S -u <user> -p custom_bolt_prompt <command>`. **Note:** This option is experimental.
```

documentation/bolt_configuration_options.md
```diff
@@ -158,6 +159,7 @@ When using the SSH transport, Bolt also interacts with the ssh-agent for SSH key
 
 -   `run-as`: A different user to run commands as after login.
 -   `run-as-command`: The command to elevate permissions. Bolt appends the user and command strings to the configured run as a command before running it on the target. This command must not require an interactive password prompt, and the `sudo-password` option is ignored when `run-as-command` is specified. The run-as command must be specified as an array.
+-   `sudo-executable`: The executable to use to run as another user when using `run-as`. This is used to replace the executable used when escalating with a password is desired, as the `run-as-command` option does not support escalating with a password. The command run on the target is `$sudo-executable -S -u $user -p $sudo-password`. **Note:** This option is experimental.
```
lucywyman: Same update above.

lib/bolt/transport/local/shell.rb
```diff
@@ -126,7 +126,8 @@ def execute(command, sudoable: true, **options)
 
           if escalate
             if use_sudo
-              sudo_flags = ["sudo", "-k", "-S", "-u", run_as, "-p", Sudoable.sudo_prompt]
+              sudo_exec = target.options['sudo-executable'] || "sudo"
```
lucywyman:
```suggestion
              sudo_exec = Shellwords.shellescape(target.options['sudo-executable']) || "sudo"
```

spec/integration/transport/ssh_spec.rb
```diff
@@ -337,6 +337,23 @@ def make_target(host_: hostname, port_: port, conf: config)
...
+    let(:config) {
+      mk_config('host-key-check' => false, 'sudo-executable' => 'fake',
+                'run-as' => user, user: bash_user, password: bash_password)
+    }
+    let(:target) { make_target }
+
+    it 'uses the correct executable' do
+      allow_any_instance_of(Net::SSH::Connection::Channel).to receive(:wait).and_return('')
+      expect_any_instance_of(Net::SSH::Connection::Channel).to receive(:exec)
+        .with("fake -S -u bolt -p \\[sudo\\]\\ Bolt\\ needs\\ to\\ run\\ as\\ another\\ "\
```
lucywyman: Since we have this test I think verifying that the executable gets escaped would be good. `sudo -n` seems like a good thing to test this with, and basically verify that it errors with `  STDERR:
    sh: 1: sudo\ -n: not found
` instead of `sudo: a password is required`

lib/bolt/transport/local/shell.rb
beechtom: `Shellwords.join` seems to already escape each element of the array, so I'm not sure if this is needed?

lucywyman: Even better!

PR #1473, opened 2019-12-09
(maint) Search plan subdirectories for plan filename

lucywyman:
We now have to discover the language a plan is written in based on it's
filename, which requires constructing the file path from the plan name
(i.e. `foo::bar::baz`) and testing if the file exists with either `.pp`
or `.yaml` extensions. This search falsely assumed that the plan path
could have at most 2 values: the module name and the plan filename.
Instead, it can have the module name, and a path relative to the `plans`
directory. This modifies the logic to correctly load nested plans.

Still needs testing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,10 @@
 
   Bolt will now lookup the default PuppetDB config at `%COMMON_APPDATA%\PuppetLabs\client-tools\puppetdb.conf` instead of the hardcoded path `C:\ProgramData\PuppetLabs\client-tools\puppetdb.conf`.
 
+* **Bolt could not find plans in subdirectories of `plans` directory**
+
+  Bolt will now correctly search for subdir paths under the `plans` directory for plan names when determining if the plan is a Puppet plan or yaml plan.
+
 ## Bolt 1.41.0
 
 ### New features
--- a/lib/bolt/pal.rb
+++ b/lib/bolt/pal.rb
@@ -312,8 +312,10 @@ module Bolt
 
       # If it's a Puppet language plan, use strings to extract data. The only
       # way to tell is to check which filename exists in the module.
-      plan_file = plan_name.split('::', 2)[1] || 'init'
-      pp_path = File.join(mod, 'plans', "#{plan_file}.pp")
+      plan_subpath = File.join(plan_name.split('::').drop(1))
+      plan_subpath = 'init' if plan_subpath.empty?
+
+      pp_path = File.join(mod, 'plans', "#{plan_subpath}.pp")
       if File.exist?(pp_path)
         require 'puppet-strings'
         require 'puppet-strings/yard'
@@ -346,7 +348,7 @@ module Bolt
 
       # If it's a YAML plan, fall back to limited data
       else
-        yaml_path = File.join(mod, 'plans', "#{plan_file}.yaml")
+        yaml_path = File.join(mod, 'plans', "#{plan_subpath}.yaml")
         plan_content = File.read(yaml_path)
         plan = Bolt::PAL::YamlPlan::Loader.from_string(plan_name, plan_content, yaml_path)
 
--- a/spec/fixtures/modules/sample/plans/subdir/command.pp
+++ b/spec/fixtures/modules/sample/plans/subdir/command.pp
@@ -1,0 +1,5 @@
+plan sample::subdir::command (
+  TargetSpec $targets
+) {
+  return run_command('echo From subdir', $targets)
+}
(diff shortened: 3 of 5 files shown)


lib/bolt/pal.rb
```diff
@@ -312,8 +312,16 @@ def get_plan_info(plan_name)
...
-      plan_file = plan_name.split('::', 2)[1] || 'init'
-      pp_path = File.join(mod, 'plans', "#{plan_file}.pp")
+      #
+      # It's easier to make 2 here than try to keep plan_dir clean
+      pp_plan = plan_name.split('::')[1..-1]
+      yaml_plan = plan_name.split('::')[1..-1]
+      if pp_plan.empty?
+        pp_plan << 'init.pp'
+      else
+        pp_plan[-1] = pp_plan[-1] << '.pp'
```
beechtom:
```suggestion
        pp_plan[-1] << '.pp'
```

I think the reassignment here is unnecessary. Appending to just the last element seems to have the same effect.

lib/bolt/pal.rb
```diff
@@ -346,7 +354,13 @@ def get_plan_info(plan_name)
 
       # If it's a YAML plan, fall back to limited data
       else
-        yaml_path = File.join(mod, 'plans', "#{plan_file}.yaml")
+        if yaml_plan.empty?
+          yaml_plan << 'init.yaml'
+        else
+          yaml_plan[-1] = yaml_plan[-1] << '.yaml'
```
beechtom:
```suggestion
          yaml_plan[-1] << '.yaml'
```

Same here.

lib/bolt/pal.rb
```diff
@@ -312,8 +312,16 @@ def get_plan_info(plan_name)
 
       # If it's a Puppet language plan, use strings to extract data. The only
       # way to tell is to check which filename exists in the module.
-      plan_file = plan_name.split('::', 2)[1] || 'init'
-      pp_path = File.join(mod, 'plans', "#{plan_file}.pp")
+      #
+      # It's easier to make 2 here than try to keep plan_dir clean
+      pp_plan = plan_name.split('::')[1..-1]
```
nicklewis: We could split and then `File.join` the segments, in which case we could just interpolate the filename along with the file extension where we need it. That would remove a lot of the duplication.

lib/bolt/pal.rb
```diff
@@ -312,8 +312,12 @@ def get_plan_info(plan_name)
 
       # If it's a Puppet language plan, use strings to extract data. The only
       # way to tell is to check which filename exists in the module.
-      plan_file = plan_name.split('::', 2)[1] || 'init'
-      pp_path = File.join(mod, 'plans', "#{plan_file}.pp")
+      #
+      # It's easier to make 2 here than try to keep plan_dir clean
+      plan_subpath = File.join(plan_name.split('::')[1..-1])
+      plan_subpath << 'init' if plan_name.split('::').length == 1
```
nicklewis: I think this is just `plan_subpath = ['init'] if plan_subpath.empty?`

lib/bolt/pal.rb
```diff
@@ -312,8 +312,12 @@ def get_plan_info(plan_name)
 
       # If it's a Puppet language plan, use strings to extract data. The only
       # way to tell is to check which filename exists in the module.
-      plan_file = plan_name.split('::', 2)[1] || 'init'
-      pp_path = File.join(mod, 'plans', "#{plan_file}.pp")
+      #
+      # It's easier to make 2 here than try to keep plan_dir clean
```
nicklewis: This comment doesn't apply anymore

lib/bolt/pal.rb
```diff
@@ -312,8 +312,12 @@ def get_plan_info(plan_name)
...
+      plan_subpath = File.join(plan_name.split('::')[1..-1])
```
nicklewis: `plan_name.split('::')[1..-1]` is also `plan_name.split('::').drop(1)`. I don't have a preference though.

CHANGELOG.md
```diff
@@ -22,6 +22,10 @@
 
   Bolt will now lookup the default PuppetDB config at `%COMMON_APPDATA%\PuppetLabs\client-tools\puppetdb.conf` instead of the hardcoded path `C:\ProgramData\PuppetLabs\client-tools\puppetdb.conf`.
 
+* **Bolt could not find plans in subdirectories of `plans` directory**
+
+  Bolt will now correctly search for subdir paths under the `plans` directory for plan names when determining if the plan is a Puppet plan or yaml plan.
```
```suggestion
  Bolt now searches for subdir paths, under the `plans` directory, for plan names when determining if the plan is a Puppet plan or YAML plan.
```

PR #1483, opened 2019-12-12
(maint) Update README

beechtom:


beechtom:
You can see the non-pretty-diff version [here](https://github.com/beechtom/bolt/tree/maint-readme-update).

Issue #1438, opened 2019-11-22
Support cascading `bolt.yaml` config overrides


op-ct:
## Use Case

From general to specific use cases:

1. As a **system administrator**: I want to configure system-wide defaults for bolt (#759) 
2. As a **user**: I want to configure my private data and personal preferences as _additional_ defaults for bolt
3. As a **bolt project maintainer**: I want to configure project-relevant settings in _addition to_ any of the user and system-wide defaults.

As a **user running bolt**, I want bolt to combine the configurations above, keeping the most specific use case's setting for each top-level option. (#608)

## Describe the Solution You Would Like

A set of cascading `bolt.yaml` config files in system -> user -> project order, with top-level settings in each file overriding the same setting in previous files(s):

![20191121_bolt_inherited_config_files](https://user-images.githubusercontent.com/8979062/69586182-be9ee800-0faf-11ea-95f0-e697043ac5a4.png)

adreyer:
The example provided does not just merge top level settings. The ssh config is merged at a secondary level.

It's unclear to me how this is different from #608 .

op-ct:
@adreyer:  The original example only merged top-level settings of the _config files,_ then followed through to show the final runtime behavior after CLI options.  I see now that was TMI to concisely describe the merging issue, so I've updated the example to just focus on _just_ that.

This issue is a superset of #608 and #759.  Those issues described different parts of a config file elephant; this is the whole thing.

adreyer:
I'm going to close this as a dupe of #608.

PR #1484, opened 2019-12-12
(GH-608) Add initial bolt config inheritance

### LINKED ISSUE #608: Bolt should use a configuration inheritance model similar to the one used by 'git'
When running `bolt` I expected the configuration options to be stacked in a method similar to those of `git` since some options are repository specific and some are user specific and should not be in a shared repository.

I did not expect the local configuration to override my global configuration completely (though it is documented at https://puppet.com/docs/bolt/0.x/configuring_bolt.html and I simply missed it).

Config inheritance order (from lowest to highest)
On non-Windows systems, the default config paths searched are:

1. `/etc/puppetlabs/bolt/bolt.yaml`
2. `~/.puppetlabs/etc/bolt/bolt.yaml`
3. `<boltdir>/bolt.yaml`

On Windows systems, the default config paths searched are:

1. `C:\Program Files\Puppet Labs\Bolt\bolt.yaml`
2. `%USERPROFILE%/.puppetlabs/etc/bolt/bolt.yaml`
3. `<boltdir>/bolt.yaml`

TODO:
- [x] Verify where Puppet looks for equivalent files on Windows
   ```
   Global: C:\ProgramData\PuppetLabs\bolt\etc
   Local: C:\Users\<user>\.puppetlabs\etc\bolt
   ```
- [x] Spike on how to merge configs. Any config options that's a hash should be evaluated for how it should be merged, and an RFC should be proposed to the Bolt team.
   https://gist.github.com/beechtom/f7b758893796d61a23c8ab868cf8b062
- [x] This should include a debug log message about what configs are being loaded and override precedence.

### LINKED ISSUE #759: Support global configuration file
We want to use `bolt` on our Jumphosts which we share within our team. For this, it is essential for us to use a shared configuration so everybody is using the same configuration (especially for Analytics).

In my opinion the configuration folder should be located at `/etc/puppetlabs/bolt/`

op-ct:
This patch adds support for bolt config files that specify system-wide
default settings (#759, #1438) and default settings for individual users
(#608, #1438).

The config files are read in a well-defined order, from most general to
most specific.  Settings defined in earlier config files are inherited
unless overridden by later configurations (including project directory
config files).

On non-Windows systems, the default config paths searched are:

1. `/etc/puppetlabs/bolt/bolt.yaml`
2. `~/.puppetlabs/etc/bolt/bolt.yaml`
3. `~/.puppetlabs/bolt.yaml`

On Windows systems, the default config paths searched are:

1. `C:\Program Files\Puppet Labs\Bolt\bolt.yaml`
2. `%USERPROFILE%/.puppetlabs/etc/bolt/bolt.yaml`
3. `%USERPROFILE%/.puppetlabs/bolt.yaml`

Closes #608 and closes #759.

### DIFF
--- a/lib/bolt/config.rb
+++ b/lib/bolt/config.rb
@@ -42,6 +42,12 @@ module Bolt
 
     PUPPETFILE_OPTIONS = %w[proxy forge].freeze
 
+    DEFAULT_CONFIG_PATHS = [
+      Bolt::Util.windows? ? 'C:\\Program Files\\Puppet Labs\\Bolt\\bolt.yaml' : '/etc/puppetlabs/bolt/bolt.yaml',
+      File.expand_path('.puppetlabs/etc/bolt/bolt.yaml', Bolt::Util.windows? ? ENV['USERPROFILE'] : '~'),
+      File.expand_path('.puppetlabs/bolt.yaml', Bolt::Util.windows? ? ENV['USERPROFILE'] : '~')
+    ].freeze
+
     def self.default
       new(Bolt::Boltdir.new('.'), {})
     end
@@ -81,6 +87,16 @@ module Bolt
       TRANSPORTS.each do |key, transport|
         @transports[key] = transport.default_options
       end
+
+      default_config_data = {}
+      DEFAULT_CONFIG_PATHS.each do |default_config_file|
+        begin
+          default_config_data.merge!(Bolt::Util.read_config_file(default_config_file, [], 'default config') || {})
+        rescue Bolt::FileError => e
+          @logger.debug(e.message)
+        end
+      end
+      config_data = default_config_data.merge(config_data)
 
       update_from_file(config_data)
       apply_overrides(overrides)
--- a/lib/bolt/config.rb
+++ b/lib/bolt/config.rb
@@ -12,6 +12,7 @@ require 'bolt/transport/local_windows'
 require 'bolt/transport/docker'
 require 'bolt/transport/remote'
 require 'bolt/util'
+require 'win32/dir' if Bolt::Util.windows?
 
 module Bolt
   TRANSPORTS = {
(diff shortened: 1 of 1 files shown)


op-ct:
* This change caused a variety of colorful spec test failures on my local system, and I expect they will fail here, too—mostly debug message expectations and File read mocks.
* For some reason, _all_ tasks failed during my account's Travis CI run with a "[fatal: No names found, cannot describe anything.](https://travis-ci.com/op-ct/bolt/jobs/266690015)" message.
* I couldn't decide which of the two user paths is better; № 3 seemed more "natural," № 2 is more consistent with the system-wide path and the user configs for Puppet.

trevor-vaughan:
@alexjfisher Thoughts? This seems to fix my general issue of Bolt not working like all of the other apps that I use.

lib/bolt/config.rb
```diff
@@ -42,6 +42,12 @@ class Config
 
     PUPPETFILE_OPTIONS = %w[proxy forge].freeze
 
+    DEFAULT_CONFIG_PATHS = [
+      Bolt::Util.windows? ? 'C:\\Program Files\\Puppet Labs\\Bolt\\bolt.yaml' : '/etc/puppetlabs/bolt/bolt.yaml',
+      File.expand_path('.puppetlabs/etc/bolt/bolt.yaml', Bolt::Util.windows? ? ENV['USERPROFILE'] : '~'),
```
adreyer: We should not include this path.

lib/bolt/config.rb
```diff
@@ -82,6 +88,16 @@ def initialize(boltdir, config_data, overrides = {})
...
 
+      default_config_data = {}
+      DEFAULT_CONFIG_PATHS.each do |default_config_file|
+        begin
+          default_config_data.merge!(Bolt::Util.read_config_file(default_config_file, [], 'default config') || {})
+        rescue Bolt::FileError => e
+          @logger.debug(e.message)
+        end
+      end
+      config_data = default_config_data.merge(config_data)
```
adreyer: This merge means that you cannot set some ssh config at one level and some at another similiarly all plugins would have to be configured at the same level. Is this desirable or should we do a deeper merge. Especially for plugin hooks arbitrarily deep merging is wrong but maybe one level of merge?

lib/bolt/config.rb
```diff
@@ -82,6 +88,16 @@ def initialize(boltdir, config_data, overrides = {})
         @transports[key] = transport.default_options
       end
 
+      default_config_data = {}
```
adreyer: As a rule we don't read file in `initialize` since it makes it impossible to avoid system config during tests and boltspec functionality.

alexjfisher:
> @alexjfisher Thoughts? This seems to fix my general issue of Bolt not working like all of the other apps that I use.

Did you mean me?

lib/bolt/config.rb
adreyer: I think this would have to be a class method that get's called by the other class methods before new.

lib/bolt/config.rb
```diff
@@ -42,6 +42,12 @@ class Config
 
     PUPPETFILE_OPTIONS = %w[proxy forge].freeze
 
+    DEFAULT_CONFIG_PATHS = [
+      Bolt::Util.windows? ? 'C:\\Program Files\\Puppet Labs\\Bolt\\bolt.yaml' : '/etc/puppetlabs/bolt/bolt.yaml',
+      File.expand_path('.puppetlabs/etc/bolt/bolt.yaml', Bolt::Util.windows? ? ENV['USERPROFILE'] : '~'),
+      File.expand_path('.puppetlabs/bolt.yaml', Bolt::Util.windows? ? ENV['USERPROFILE'] : '~')
```
adreyer: This path was used by bolt in the 0.x series. I think its the correct choice for this but we should feature flag acutally merging in default config data based on whether the Boltdir or specified file has `future` set. We can then expose this as the only config behavior for Bolt 2.0.

lib/bolt/config.rb
```diff
@@ -42,6 +42,12 @@ class Config
 
     PUPPETFILE_OPTIONS = %w[proxy forge].freeze
 
+    DEFAULT_CONFIG_PATHS = [
+      Bolt::Util.windows? ? 'C:\\Program Files\\Puppet Labs\\Bolt\\bolt.yaml' : '/etc/puppetlabs/bolt/bolt.yaml',
```
beechtom: Default paths in Windows should be specified similar to how they are in `Bolt::PuppetDB::Config`: https://github.com/puppetlabs/bolt/blob/master/lib/bolt/puppetdb/config.rb#L20-L23

Hardcoding the path to `C:\Program Files` can cause issues for users who redirect the folder or use a different system drive.

lib/bolt/config.rb
trevor-vaughan: In general, I'm used to a deep merge of SSH settings. Thinking about the `~/.ssh/config` files.

adreyer:
So I think the main issues are
1. These files should only be used if `future` is set in the Boltdir.
2. Merging needs to be consistent.
3. The initialize method should not load anything.
4. Long term we need to know where data came from to help users debug.(I think we can merge without this)

Currently the config has two methods that "merge" data
1. update_from_file which merges in arbitrary data with the correct precedence but has some weirdness around `future`
2. apply_overrides which handles CLI flags.

I think the simplest way to accomplish this is:

1. accept an array of config defaults in initialize. Call `update_from_file` on each config_default hash before calling it on the `config_data`.
2 Create a new class method to handle loading config_default that gets called from `from_boltdir` and `from_file`.

```
def self.load_defaults(boltdir, config_data, overrides)
  if config_data['future'] == true
    # return an array of config data hashes from each of the default files
    defaults = load_config_defaults
  else
     defaults = []
  end
  new(boltdir, config_data, overrides, defaults)
end
```

lib/bolt/config.rb
op-ct: @adreyer & @trevor-vaughan: I also wondered about that, but it's still not clear what the optimal balance of behaviors (predictable + desirable - undesirable - future-keys) should be.  

Some possibilities:

|  Strategy | Effected  | Comments |
| --- | --- | --- |
| Shallow merge | All settings, top-level |  Simplest, safest, probably inconvenient-est |
| Shallow merge  | All settings, 1-deep  | Allows multiple transport opts and plugins |
| Deep merge | All settings | ⛔️ Not appropriate for `plugin_hooks` |
| Deep merge | All settings - undesirable | Shallow (or 1-d )for keys like `plugin_hooks` |
| Deep merge | Only desired settings | e.g., just transport settings  |
  
For the moment, I've left it as shallow top-level everything.

op-ct: @adreyer & @trevor-vaughan: I also wondered about that, but it's still not clear what the optimal balance of behaviors (predictable + desirable - undesirable - future-keys) should be.

Some possibilities:

|  Strategy | Effected  | Comments |
| --- | --- | --- |
| Shallow merge | All settings, top-level |  Simplest, safest, probably inconvenient-est |
| Shallow merge  | All settings, 1-deep  | Allows multiple transport opts and plugins |
| Deep merge | All settings | ⛔️ Not appropriate for `plugin_hooks` |
| Deep merge | All settings - undesirable | Shallow (or 1-d )for keys like `plugin_hooks` |
| Deep merge | Only desired settings | e.g., just transport settings  |
  
For the moment, I've left it as shallow top-level everything because it's easy and safe.

nicklewis:
This is superseded by #1557

PR #1486, opened 2019-12-16
(bug-fix) Gate key change in JSON formatted output on future

lucywyman:
We introduced a breaking change to JSON formatted output without gating
it on future. This restores the previous keys in the JSON output if
`future` is not configured.


lucywyman:
Yeah that was exactly my thinking - we'll be removing `$future` in the first week or two of the new year, so I optimized for ease of deletion :)

PR #1494, opened 2019-12-17
(maint) Remove plan_executor POC code

donoghuc:
Given the plan executor was implemented in Clojure/JRuby in the orchestrator repo, this POC code is no longer used. This commit removes it.


lucywyman:
I think this can also probably remove `./spec/fixtures/plan_executor` and `./config/plan_executor_config.rb`

Issue #1491, opened 2019-12-16
Add deprecation warning for bolt-inventory-pdb


lucywyman:
## Use Case
We are dropping support for inventory version 1 with Bolt 2.0, and don't plan to support the `bolt-inventory-pdb` command moving forward. Users can use the PuppetDB inventory plugin instead, and use `bolt project migrate` to migrate an existing (static) inventory to V2.

Issue #1464, opened 2019-12-04
Update inventory file and plugins docs


beechtom:
## Describe the Change You Would Like
Currently, most of the information for how to write an inventory file is located on the [Inventory file v1](https://puppet.com/docs/bolt/latest/inventory_file.html) page. The [Inventory file v2](https://puppet.com/docs/bolt/latest/inventory_file_v2.html) page does not have similar information that shows examples of a version 2 inventory with a list of available fields. Additionally, the v2 page has information about creating and managing targets in a plan, which is not related to the inventory file itself.

The Inventory File v2 page should be rewritten to include the following content:
- [x] List of available fields
- [x] Examples of a v2 file
- [x] A simple example of using a plugin, with a link to a plugins reference page
- [x] How to migrate from a v1 file to a v2 file (e.g. manually changing fields or using `project migrate`
- [x] How reference plugins work

For now, this should be 1 page, and can be broken up during PR review

## Questions

* Should the v1 page also be updated, or should this be left as-is? We can mark the V1 page as deprecated (maybe rename to Legacy inventory file)

PR #1496, opened 2019-12-17
(GH-1491) Add deprecation message for bolt-inventory-pdb

beechtom:
This adds a deprecation message to the `bolt-inventory-pdb` command
warning users about support for the command and inventory file v1 being
dropped in the future.

This change should be shipped with the planned changes to the inventory docs: #1464

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Bolt Next
 
+### Deprecations
+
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Bolt Next
 
+### Deprecations
+
+* **WARNING**: Support for the `bolt-inventory-pdb` command will be dropped in the near future. Users can use the `puppetdb` plugin with a v2 inventory file to lookup targets from PuppetDB.
```
```suggestion
* **Support for the `bolt-inventory-pdb` command will be dropped in Bolt 2.0**. Users can use the [puppetdb inventory plugin](https://puppet.com/docs/bolt/latest/using_plugins.html#puppetdb) with a v2 inventory file to lookup targets from PuppetDB.
```

PR #1497, opened 2019-12-17
(GH-1480) Add deprecation message for v1 inventory files

### LINKED ISSUE #1480: Add deprecation warning about Inventory v1
We should add a deprecation warning when we detect users are using inventory version 1, stating that support for v1 will be dropped and that the v2 inventory will be the default. Include the command to convert v1 to v2.

This change should be shipped with the planned changes to the inventory docs: #1464

beechtom:
This adds a deprecation message for when a v1 inventory file is loaded
and how to migrate to a v2 inventory file.

This change should be shipped with the planned changes to the inventory docs: #1464

Closes #1480

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
 
 * **Support for the `bolt-inventory-pdb` command will be dropped in Bolt 2.0.** Users can use the [puppetdb inventory plugin](https://puppet.com/docs/bolt/latest/using_plugins.html#puppetdb) with a v2 inventory file to lookup targets from PuppetDB.
 
+* **Support for the v1 inventory files will be dropped in Bolt 2.0.** Inventory files [can be migrated](https://puppet.com/docs/bolt/latest/migrating_inventory_files.html) automatically using the `bolt project migrate` command.
+
 ### New features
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Bolt Next
 
+### Deprecations
+
+* **Support for the v1 inventory files will be dropped in Bolt 2.0.** Inventory files [can be migrated](https://puppet.com/docs/bolt/latest/migrating_inventory_files.html) manually or automatically using the `bolt project migrate` command.
```
```suggestion
* **Support for the v1 inventory files will be dropped in Bolt 2.0.** Inventory files [can be migrated](https://puppet.com/docs/bolt/latest/migrating_inventory_files.html) automatically using the `bolt project migrate` command.
```

spec/bolt/applicator_spec.rb
```diff
@@ -7,11 +7,14 @@
 require 'bolt/pal'
 require 'bolt/puppetdb'
 require 'bolt/target'
+require 'bolt/plugin'
```
lucywyman: I think these changes are here by accident

PR #1501, opened 2019-12-18
(GH-1481) Add deprecation message for Target.new(<uri>, <options>)

### LINKED ISSUE #1481: Print deprecation warning for Target.new(<uri>, <options>)
With 2.0 Bolt will only support `Target.new` from plans with a uri specified - the options parameter will no longer be available. We should include a deprecation warning when users try to call Target.new with an options parameter.

beechtom:
This adds a deprecation message for when `Target.new(<uri>, <options>)`
is used in a plan. It also notifies the user to use `Target.new(<uri>)`
instead.

Closes #1481

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,8 @@
 
 * **Support for the v1 inventory files will be dropped in Bolt 2.0.** Inventory files [can be migrated](https://puppet.com/docs/bolt/latest/migrating_inventory_files.html) automatically using the `bolt project migrate` command.
 
+* **Support for plan method `Target.new(<uri>, <options>)` will be dropped in Bolt 2.0.** The plan method `Target.new(<options>)`, where `options` is a hash of target data and config, should be used instead. See [the docs](https://puppet.com/docs/bolt/latest/writing_plans.html#creating-target-objects) for more information and examples.
+
 ### New features
 
 * **Packages for Fedora 31** ([#1373](https://github.com/puppetlabs/bolt/issues/1373))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,16 @@
   deprecated in version 1.35 in favor of the `puppet_agent` plugin, and is now removed. The plugins
   have the exact same behavior.
 
+* **Support for plan method `Target.new(<uri>, <options>)` will be dropped in Bolt 2.0.** Use 
+  `Target.new(<config>)`, where `config` is a hash with the same structure used to define targets in 
+  the inventory V2 file. See [the docs](https://puppet.com/docs/bolt/latest/writing_plans.html#creating-target-objects) 
+  for more information and examples.
+
+* **Support for `options` key in the hash parameter for `Target.new()` plan function will be dropped in Bolt 2.0.** Use 
+  `Target.new(<config>)`, where `config` is a hash with the same structure used to define targets in 
+  the inventory V2 file. See [the docs](https://puppet.com/docs/bolt/latest/writing_plans.html#creating-target-objects) 
+  for more information and examples.
+
 ### Bug fixes
 
 * **SSH commands will run from the home directory of the run-as user, not the connected user** ([#1518](https://github.com/puppetlabs/bolt/pull/1518))
@@ -23,8 +33,6 @@
 * **Support for the `bolt-inventory-pdb` command will be dropped in Bolt 2.0.** Users can use the [puppetdb inventory plugin](https://puppet.com/docs/bolt/latest/using_plugins.html#puppetdb) with a v2 inventory file to lookup targets from PuppetDB.
 
 * **Support for the v1 inventory files will be dropped in Bolt 2.0.** Inventory files [can be migrated](https://puppet.com/docs/bolt/latest/migrating_inventory_files.html) automatically using the `bolt project migrate` command.
-
-* **Support for plan method `Target.new(<uri>, <options>)` will be dropped in Bolt 2.0.** The plan method `Target.new(<options>)`, where `options` is a hash of target data and config, should be used instead. See [the docs](https://puppet.com/docs/bolt/latest/writing_plans.html#creating-target-objects) for more information and examples.
 
 ### New features
 
--- a/lib/bolt/inventory.rb
+++ b/lib/bolt/inventory.rb
@@ -72,7 +72,8 @@ module Bolt
       end
     end
 
-    attr_reader :plugins, :config
+    attr_reader :plugins, :config, :logger
+    attr_accessor :hash_deprecation_issued, :args_deprecation_issued
 
     def initialize(data, config = nil, plugins: nil, target_vars: {},
                    target_facts: {}, target_features: {}, target_plugin_hooks: {})
--- a/lib/bolt/inventory.rb
+++ b/lib/bolt/inventory.rb
@@ -435,5 +435,17 @@ module Bolt
       @target_plugin_hooks.delete(target.name)
     end
     private :invalidate_group_cache!
+
+    # Logs deprecation warning for Target.new(<uri>, <options>)
+    def target_deprecation
+      return if @deprecation_issued
+
+      @deprecation_issued = true
+      msg = <<~MSG
+        Deprecation Warning: Starting with Bolt 2.0, 'Target.new(<uri>, <options>)' will no
+        longer be supported. Use 'Target.new(<options>)' instead.
+      MSG
+      @logger.warn(msg)
+    end
   end
 end
--- a/lib/bolt/inventory.rb
+++ b/lib/bolt/inventory.rb
@@ -72,8 +72,7 @@ module Bolt
       end
     end
 
-    attr_reader :plugins, :config, :logger
-    attr_accessor :hash_deprecation_issued, :args_deprecation_issued
+    attr_reader :plugins, :config
 
     def initialize(data, config = nil, plugins: nil, target_vars: {},
                    target_facts: {}, target_features: {}, target_plugin_hooks: {})
@@ -435,17 +434,5 @@ module Bolt
       @target_plugin_hooks.delete(target.name)
     end
     private :invalidate_group_cache!
-
-    # Logs deprecation warning for Target.new(<uri>, <options>)
-    def target_deprecation
-      return if @deprecation_issued
-
-      @deprecation_issued = true
-      msg = <<~MSG
-        Deprecation Warning: Starting with Bolt 2.0, 'Target.new(<uri>, <options>)' will no
-        longer be supported. Use 'Target.new(<options>)' instead.
-      MSG
-      @logger.warn(msg)
-    end
   end
 end
--- a/lib/bolt/target.rb
+++ b/lib/bolt/target.rb
@@ -157,34 +157,32 @@ module Bolt
 
     # Satisfies the Puppet datatypes API
     def self.from_asserted_hash(hash)
-      inv = Puppet.lookup(:bolt_inventory)
-
-      if hash['uri'] && hash['options'] && inv.hash_deprecation_issued.nil?
-        inv.hash_deprecation_issued = true
+      if hash['uri'] && hash['options']
+        logger = Logging.logger[self]
         msg = <<~MSG
-          Deprecation Warning: Starting with Bolt 2.0, using 'Target.new' with an 'options' hash key will no
-          will no longer be supported. Use 'Target.new(<config>)', where 'config' is a hash with the same
-          structure used to define targets in the inventory V2 file. For more information see 
-          https://puppet.com/docs/bolt/latest/writing_plans.html#creating-target-objects
+          #{Puppet::Pops::PuppetStack.top_of_stack.join(':')}
+            Deprecation Warning: Starting with Bolt 2.0, using 'Target.new' with an 'options' hash key will no
+            will no longer be supported. Use 'Target.new(<config>)', where 'config' is a hash with the same
+            structure used to define targets in the inventory V2 file. For more information see
+            https://puppet.com/docs/bolt/latest/writing_plans.html#creating-target-objects
         MSG
-        inv.logger.warn(msg)
+        logger.warn(msg)
       end
-      
+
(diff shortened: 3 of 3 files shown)


lib/bolt/target.rb
```diff
@@ -157,6 +157,7 @@ class Target
 
     # Satisfies the Puppet datatypes API
     def self.from_asserted_hash(hash)
+      Puppet.lookup(:bolt_inventory).deprecation
```
donoghuc: I think you will need a `from_asserted_args` to capture what you want. This will only warn on `Target.new({'uri'=>'foo','options'=>{...}})` not `Target.new('foo', {...})`.

beechtom: I'm not sure I follow. Why would I add a method only to add a deprecation method to it? My misunderstanding is probably coming from the lack of documentation around Target.new(<uri>, <options>)..

donoghuc: I think the only way to throw the message you want is to add the `from_asserted_args` class method. Otherwise you will never be able determine if a new target is instantiated with `Target.new('foo', {...})`. Try this plan:
```
  Target.new('foo')
  Target.new('bar', {'host-key-check' => 'false'})
  Target.new({'uri' => 'baz', 'options' => {'host-key-check' => 'false'}})
```
The one you want to add a deprecation warning for is the second one (and maybe the third one?) currently i think it will only warn on the second one.

lucywyman: I *think* puppet will actually just call `.new` if `from_asserted_args` isn't defined https://github.com/puppetlabs/puppet/blob/master/lib/puppet/pops/types/p_object_type.rb#L520 Puppet uses `from_asserted_hash` and `from_asserted_args` to create datatypes by passing arguments from your `Target.new` call to the datatypes implementation class instantiator. Target is a datatype and `target.rb` has it's implementation classes. `from_asserted_hash` is called when `Target.new({'uri' => 'foo', 'options' => { ... }})` is called (like Cas said), but if you pass the attributes as arguments instead of as a hash then theoretically `from_asserted_args` is called with the arguments you passed it. The part that's confusing is that this API isn't really documented (that I can find...), this is pretty much from https://github.com/puppetlabs/puppet/blob/master/lib/puppet/pops/types/p_object_type.rb

lucywyman: Confirmed that `from_asserted_args` isn't called if it's not defined, it just calls `.new` with the arguments. So 
```
plan test::foo() {
  return Target.new('localhost', { 'name' => 'name' })
}
```
with a V1 inventory just calls `initialize`. I think you just need to add the deprecation call there as well

beechtom: I think if the intent is for only `Target.new('foo')` to be valid it makes sense to add a deprecation warning for both the second and third options. Updated to add `from_asserted_args`.

donoghuc: I dont know what to think of warning on the third one. I see `Target.new({...})` as being basically the whole point of the target (2) API. Specifically I think the *only* reason to use Target.new is to be able to set each part of the target you care about in the initialization.

lucywyman: `Target.new({'uri' => 'foo'})` is still valid as long as it doesn't have the options hash. That will still work with a V2 inventory. `Target.new({'uri' => 'foo', 'options' => { ... }})` should raise the warning and `Target.new('uri', { ... })` should also raise the warning.

donoghuc: Yeah that makes sense to me ^^. Whole sale warning on Target.new({..}) is bad, but if there is `uri`, and `options` keys that makes sense to warn on.

lucywyman: Yeah, I don't think we should warn on the third one - that will still work with Inventory v2. The intention here is to make sure that plans are compatible with Inventory v2, so any target calls that behave the same between inventory versions 1 and 2 should not raise a warning.

beechtom: Updated to warn on `Target.new(uri, {})` and `Target.new('uri' => ..., 'options' => ...)`.

lucywyman: @donoghuc Sorry I thought we might actually accept an `options` key in that hash, but confirmed we don't :) So yeah, `Target.new({ 'uri' => 'foo'})` should not warn, `Target.new({ 'uri' => 'foo', 'options' => { ... })` should.

donoghuc: wait really? https://github.com/puppetlabs/bolt/blob/a67b8013dd9121276217a8ab61ef675957262800/lib/bolt/target.rb#L160 looks like options is fine?

lucywyman: Sorry, I meant in Inventory 2! I thought we might still support `options` key in inventory 2, but confirmed we don't. that's why I deleted my earlier comment :P Either way seems like Tom implemented the right behavior.

donoghuc: ahh, that makes sense! yeah good riddence `options`. :)

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Bolt NEXT
 
+### Deprecations
+
+* **WARNING**: Support for plan method `Target.new(<uri>, <options>)` will be dropped in the near future. The plan method `Target.new(<uri>)` should be used instead.
```
lucywyman: I think we want users to use an options hash, like `Target.new('uri' => <uri>, 'config' => { 'ssh' => { 'port' => '23' } })`

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## Bolt Next
+
+### Deprecations
+
+* **WARNING**: Support for plan method `Target.new(<uri>, <options>)` will be dropped in the near future. The plan method `Target.new(<options>)`, where `options` is a hash of target data and config, should be used instead.
```
lucywyman: Ok, I think the best approach is to wait until #1522 is merged and do "Use `Target.new(<options>)`, where `options` is a hash of target data and config, instead. See [the docs](link-to-writing-plans-page) for more information and examples.

lib/bolt/target.rb
```diff
@@ -157,9 +157,15 @@ class Target
 
     # Satisfies the Puppet datatypes API
     def self.from_asserted_hash(hash)
+      Puppet.lookup(:bolt_inventory).target_deprecation if hash['uri'] && hash['options']
```
lucywyman: I think this actually needs to raise a different error message. This is raised when
```
Target.new({ 'uri' => 'foo', 'options' => { 'port' => '22' } })
```
Which is still not going to be valid in Bolt 2.0 because we don't support the 'options' key anymore. This needs a different message saying that the 'options' hash key has been deprecated in favor of 'config'.

CHANGELOG.md
```diff
@@ -8,6 +8,8 @@
 
 * **Support for the v1 inventory files will be dropped in Bolt 2.0.** Inventory files [can be migrated](https://puppet.com/docs/bolt/latest/migrating_inventory_files.html) automatically using the `bolt project migrate` command.
 
+* **Support for plan method `Target.new(<uri>, <options>)` will be dropped in Bolt 2.0.** The plan method `Target.new(<options>)`, where `options` is a hash of target data and config, should be used instead. See [the docs](https://puppet.com/docs/bolt/latest/writing_plans.html#creating-target-objects) for more information and examples.
```
```suggestion
* **Support for plan method `Target.new(<uri>, <options>)` will be dropped in Bolt 2.0.** Use `Target.new(<options>)`, where `options` is a hash with the same structure used to define targets in the inventory V2 file. See [the docs](https://puppet.com/docs/bolt/latest/writing_plans.html#creating-target-objects) for more information and examples.
```

lib/bolt/inventory.rb
```diff
@@ -434,5 +434,17 @@ def invalidate_group_cache!(target)
...
     private :invalidate_group_cache!
+
+    # Logs deprecation warning for Target.new(<uri>, <options>)
+    def target_deprecation
+      return if @deprecation_issued
+
+      @deprecation_issued = true
+      msg = <<~MSG
+        Deprecation Warning: Starting with Bolt 2.0, 'Target.new(<uri>, <options>)' will no
+        longer be supported. Use 'Target.new(<options>)' instead.
```
lucywyman:
```suggestion
        longer be supported. Use `Target.new(<options>)`, where `options` is a hash with the same structure used to define targets in the inventory V2 file. See https://puppet.com/docs/bolt/latest/writing_plans.html#creating-target-objects for more information.
```

lib/bolt/inventory.rb
```diff
@@ -434,5 +434,17 @@ def invalidate_group_cache!(target)
       @target_plugin_hooks.delete(target.name)
     end
     private :invalidate_group_cache!
+
+    # Logs deprecation warning for Target.new(<uri>, <options>)
+    def target_deprecation
```
lucywyman: The inventory here is actually just keeping track of whether we've already printed a deprecation message or not. I think it makes more sense to have the deprecation messages be defined where they're called, since the message is distinct for each case, and to track if they've been printed as class variables on the inventory class. `args_deprecation_issued` and `hash_deprecation_issued` or something, that are set to false in the inventory initialization and are in `attr_accessor` to be read and reset in the target class.

lib/bolt/inventory.rb
nicklewis: We should check `Puppet::Pops::PuppetStack.top_of_stack` to give the file and line number where the warning happened. If we do that, it might be reasonable to issue the warning each time it happens.

CHANGELOG.md
```diff
@@ -4,17 +4,15 @@
...
+* **Support for plan method `Target.new(<uri>, <options>)` will be dropped in Bolt 2.0.** Use `Target.new(<config>)`, where `config` is a hash with the same structure used to define targets in the inventory V2 file. See [the docs](https://puppet.com/docs/bolt/latest/writing_plans.html#creating-target-objects) for more information and examples.
 
 ### Bug fixes
 
 * **SSH commands will run from the home directory of the run-as user, not the connected user** ([#1518](https://github.com/puppetlabs/bolt/pull/1518))
 
-  Connecting via SSH and then switching users will now run as though it had
-  connected as the new user in the first place, using that user's home
-  directory as the working directory.
+  Connecting via SSH and then switching users will now run as though it had connected as the new user in the first place, using that user's home directory as the working directory.
```
lucywyman: We (Heston, Alex, and I at least) prefer these stay broken over multiple lines. It makes git diffs easier to read and editing easier.

CHANGELOG.md
```diff
@@ -24,6 +22,8 @@
 
 * **Support for the v1 inventory files will be dropped in Bolt 2.0.** Inventory files [can be migrated](https://puppet.com/docs/bolt/latest/migrating_inventory_files.html) automatically using the `bolt project migrate` command.
 
+* **Support for plan method `Target.new(<uri>, <options>)` will be dropped in Bolt 2.0.** Use `Target.new(<options>)`, where `options` is a hash with the same structure used to define targets in the inventory V2 file. See [the docs](https://puppet.com/docs/bolt/latest/writing_plans.html#creating-target-objects) for more information and examples.
```
lucywyman: This is a duplicate

CHANGELOG.md
```diff
@@ -4,17 +4,15 @@
 
 ### Deprecations and removals
 
-* **The install_agent plugin has been officially removed.** The `install_agent` plugin was
-  deprecated in version 1.35 in favor of the `puppet_agent` plugin, and is now removed. The plugins
-  have the exact same behavior.
+* **The install_agent plugin has been officially removed.** The `install_agent` plugin was deprecated in version 1.35 in favor of the `puppet_agent` plugin, and is now removed. The plugins have the exact same behavior.
+
+* **Support for plan method `Target.new(<uri>, <options>)` will be dropped in Bolt 2.0.** Use `Target.new(<config>)`, where `config` is a hash with the same structure used to define targets in the inventory V2 file. See [the docs](https://puppet.com/docs/bolt/latest/writing_plans.html#creating-target-objects) for more information and examples.
```
lucywyman: This still needs an additional deprecation notice for the `options` hash key for `Target.new()`.

CHANGELOG.md
```diff
@@ -8,14 +8,24 @@
...
+
 ### Bug fixes
 
 * **SSH commands will run from the home directory of the run-as user, not the connected user** ([#1518](https://github.com/puppetlabs/bolt/pull/1518))
 
   Connecting via SSH and then switching users will now run as though it had
   connected as the new user in the first place, using that user's home
   directory as the working directory.
-
+
```
lucywyman: Woops, extra space

lib/bolt/target.rb
```diff
@@ -157,9 +157,39 @@ class Target
 
     # Satisfies the Puppet datatypes API
     def self.from_asserted_hash(hash)
+      inv = Puppet.lookup(:bolt_inventory)
```
lucywyman: We should be able to instantiate a logger here and use that instead of having to pull in the inventory.
```
logger = Logging.logger[self]
...
logger.warn("...")
```

lib/bolt/inventory.rb
```diff
@@ -72,7 +72,7 @@ def self.create_version(data, config, plugins)
       end
     end
 
-    attr_reader :plugins, :config
+    attr_reader :plugins, :config, :logger
```
lucywyman: We can remove this if we just make a logger in the target itself.

CHANGELOG.md
```diff
@@ -8,6 +8,16 @@
   deprecated in version 1.35 in favor of the `puppet_agent` plugin, and is now removed. The plugins
   have the exact same behavior.
 
+* **Support for plan method `Target.new(<uri>, <options>)` will be dropped in Bolt 2.0.** Use 
+  `Target.new(<config>)`, where `config` is a hash with the same structure used to define targets in 
+  the inventory V2 file. See [the docs](https://puppet.com/docs/bolt/latest/writing_plans.html#creating-target-objects) 
+  for more information and examples.
+
+* **Support for plan method `Target.new` with an `options` hash key will be dropped in Bolt 2.0.** Use
```
```suggestion
* **Support for `options` key in the hash parameter for `Target.new()` plan function will be dropped ** Use 
```

PR #1502, opened 2019-12-18
(GH-1504) Do not try to expand a file path when ssh private-key key-data is specified

donoghuc:
When attempting to expand paths in config, only do so for the `ssh` transport setting `private-key` when it is not a `Hash` (indicating that `key-data` is being used.).

Closes https://github.com/puppetlabs/bolt/issues/1504

### DIFF
--- a/lib/bolt/config.rb
+++ b/lib/bolt/config.rb
@@ -223,7 +223,7 @@ module Bolt
           if @future
             to_expand = %w[private-key cacert token-file] & selected.keys
             to_expand.each do |opt|
-              selected[opt] = File.expand_path(selected[opt], @boltdir.path) if opt.is_a?(String)
+              selected[opt] = File.expand_path(selected[opt], @boltdir.path) if selected[opt].is_a?(String)
             end
           end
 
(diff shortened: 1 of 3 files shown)


donoghuc:
TODO: Make GH issue, write test, fix commit message.

lib/bolt/config.rb
```diff
@@ -223,7 +223,7 @@ def update_transports(data)
           if @future
             to_expand = %w[private-key cacert token-file] & selected.keys
             to_expand.each do |opt|
-              selected[opt] = File.expand_path(selected[opt], @boltdir.path) if opt.is_a?(String)
+              selected[opt] = File.expand_path(selected[opt], @boltdir.path) if selected[opt].is_a?(String)
             end
```
donoghuc: Is it possible for opt not to be a string? I couldn't tell if this was a typo or if there is actually a case where it is not a string.

lucywyman: I don't think so, because those are the config keys not the values. It looks like this was introduced in https://github.com/puppetlabs/bolt/commit/871e92d52dca54eedb3d5f87900cc054dcb9f7fd and is definitely a bug. Thanks Cas!

PR #1503, opened 2019-12-18
(PE-27794) add a puma status endpoint

tkishel:
With this commit, a puma status endpoint is added to our app.
(We already implement the gc_stats endpoint.)

Since this endpoint is served by our app, there will always be one
thread in use in the result: that thread is the status request itself.

This is not true with the default puma status endpoint,
as it runs within a separate server ... as per puma/lib/puma/runner.rb.

### DIFF
--- a/lib/bolt_server/transport_app.rb
+++ b/lib/bolt_server/transport_app.rb
@@ -188,6 +188,11 @@ module BoltServer
       [200, GC.stat.to_json]
     end
 
+    get '/admin/status' do
+      stats = Puma.stats
+      [200, stats.is_a?(Hash) ? stats.to_json : stats]
+    end
+
     get '/500_error' do
       raise 'Unexpected error'
     end


lib/bolt_server/transport_app.rb
```diff
@@ -188,6 +188,11 @@ def run_script(target, body)
       [200, GC.stat.to_json]
     end
 
+    get '/admin/status' do
+      stats = Puma.stats
```
donoghuc: This seems much simpler! I'm not sure I follow what you mean by the threading part... Also tests are not liking this constant. Might need a require or to fully qualify this.

tkishel: The default status app is started here:

https://github.com/puma/puma/blob/master/lib/puma/runner.rb#L49

Its "out of band" as in it has its own thread pool (from ace or bolt) ... limited to 1 thread.

donoghuc:
Is this not possible to test?

lib/bolt_server/transport_app.rb
tkishel: I noticed that we don't test the other (`/admin/gc_stat`) stat endpoint:

https://github.com/puppetlabs/bolt/blob/master/lib/bolt_server/transport_app.rb#L187

And `stats` is tested in puma:

https://github.com/puma/puma/blob/master/test/test_app_status.rb

Rather than a `require` I think I would have to mock `Puma.stats` resulting in a test that just checks this:

```
stats.is_a?(Hash) ? stats.to_json : stats
```

Possibly we do not have to test `/admin/status`, since we don't test `/admin/gc_stat` ? 

Here is a real test in Vagrant:

```
[root@pe-201921-master-ha ~]# curl https://$(puppet config print server):62658/admin/status --cacert /etc/puppetlabs/puppet/ssl/certs/ca.pem --cert /etc/puppetlabs/puppet/ssl/certs/$(puppet config print certname).pem --key /etc/puppetlabs/puppet/ssl/private_keys/$(puppet config print certname).pem
{ "backlog": 0, "running": 1, "pool_capacity": 99, "max_threads": 100 }
```

tkishel:
This - was - my spec test:

```
  it 'responds to status endpoint with Puma.stats' do
    get '/admin/stats'
    expect(last_response).to be_ok
    result = JSON.parse(last_response.body)
    expect(result).to have_key('max_threads')
  end
```

tkishel:
> Is this not possible to test?

`Puma.stats` is valid in:

```
[root@pe-201921-master-ha ~]# grep -A2 -B1 Puma /opt/puppetlabs/server/apps/bolt-server/lib/ruby/gems/bolt-1.33.0/lib/bolt_server/transport_app.rb
    get '/admin/status' do
      stats = Puma.stats
      [200, stats.is_a?(Hash) ? stats.to_json : stats]
    end
```

```
[root@pe-201921-master-ha ~]# curl https://$(puppet config print server):62658/admin/status --cacert /etc/puppetlabs/puppet/ssl/certs/ca.pem --cert /etc/puppetlabs/puppet/ssl/certs/$(puppet config print certname).pem --key /etc/puppetlabs/puppet/ssl/private_keys/$(puppet config print certname).pem
{ "backlog": 0, "running": 1, "pool_capacity": 99, "max_threads": 100 }
```

But I have not been able to resolve this error in the spec test:

```
uninitialized constant BoltServer::TransportApp::Puma
```

🤷‍♂️

donoghuc:
I saw the original spec test failure, but then it started passing in Travis? What did you change? Is it failing locally for you?

donoghuc:
Oh I see you marked the test as pending.

donoghuc:
I think we can just delete the test. Our tests cover the sinatra app. Not the puma server. So there is no puma instance https://github.com/puma/puma/blob/1bec245b631baf7e02f18ce2412dc9cca331b6cc/lib/puma.rb#L23 which is where the stats method is failing in the tests.

tkishel:
That's what I was thinking ... since `Puma.stats` is tested in Puma, and we chose not to test `/admin/gc_stat` which similarly calls the unqualified `GC.stat`.

donoghuc:
Even if Puma resolves correctly there I think (which maybe we should `require 'puma'` in the transport_app, it will still fail. GC works because that is just a ruby thing. So i think just delete the spec test. YOLO.

tkishel:
YOLO! Will delete that commit now ...

PR #1505, opened 2019-12-19
(GH-1500) Support .yml for bolt config and inventory

### LINKED ISSUE #1500: Support .yml extension for bolt.yaml and inventory.yaml
## Use Case
We should support both .yml and .yaml extensions for Bolt configuration files `bolt.yaml` and `inventory.yaml`. If both exist `.yaml` should take precedence.

https://tickets.puppetlabs.com/browse/BOLT-259

lucywyman:
This adds support for the `.yml` extension for bolt config and inventory
files. Bolt will recognize either `bolt.yaml` or `bolt.yml` as the root
of a local Boltdir, with `bolt.yml` following the same behavior as
`bolt.yaml` with respect to project directory type precedence. If both
`.yaml` and `.yml` files are present for config or inventory, `.yaml` is
preferred. Any yaml files generated by bolt still have the `.yaml`
extension.

Closes #1500

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,14 @@
 
   Bolt no longer attempts to expand a `private-key` configuration `Hash` when `key-data` is being used in conjunction with the `future` setting.
 
+### New features
+
+* **Support .yml extension for Bolt config and inventory** ([#1500](https://github.com/puppetlabs/bolt/issues/1500))
+
+  Bolt will now read config and inventory from `bolt.yml` and `inventory.yml` in the Boltdir. Bolt
+  will also treat `bolt.yml` as the root of a local-type project directory. If both `.yaml` and `.yml`
+  extensions are present, `.yaml` is preferred.
+
 ## Bolt 1.43.0
 
 ### New features
--- a/documentation/bolt_project_directories.md
+++ b/documentation/bolt_project_directories.md
@@ -10,7 +10,8 @@ There are three types of project directories that you can use depending on how y
 
 ### Local project directory
 
-Bolt treats a directory containing a `bolt.yaml` file as a project directory. Use this type of directory to track and share management code in a dedicated repository.
+Bolt treats a directory containing a `bolt.yaml` or `bolt.yml` file as a project directory. Use this
+type of directory to track and share management code in a dedicated repository.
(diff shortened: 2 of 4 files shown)


donoghuc:
Maybe to `hiera.yaml` => `hiera.yml` to? Though I feel like we talked about this a long time ago and decided we wanted to just follow the suggested `yaml` https://yaml.org/faq.html

lucywyman:
I didn't because for Puppet hiera config has to be `.yaml` (https://puppet.com/docs/puppet/6.10/hiera_config_yaml_5.html, https://github.com/puppetlabs/puppet/blob/master/lib/puppet/pops/lookup/hiera_config.rb#L42). I thought about making `analytics.yaml` and YAML plans also support it, but this came up because someone asked for it in slack and I figured I'd wait until someone also asked for those files to make those changes.

beechtom:
I think adding support for `analytics.yml` makes sense, if only because it seems weird to fix this issue for config and inventory, but not for analytics collection.

donoghuc:
Maybe the fact that it does not work for hiera is an indication that we should not allow it in bolt? I'm thinking this is more work that it is worth? But i'm not super strongly opposed.

lucywyman:
I think Hiera doesn't support both for the same reason we didn't: because `.yaml` is preferred and no one has asked for `.yml`. This came up because several users are accustomed to three-letter extensions (possibly from when Windows applications would only load / recognize three-letter extensions), and while this doesn't prevent people from manually specifying a `.yml` config and inventory I think it's a cheap quality of life improvement. See [this comment](https://puppetcommunity.slack.com/archives/C7GP57T50/p1576688952133600).

CHANGELOG.md
```diff
@@ -8,6 +8,14 @@
 
   Bolt no longer attempts to expand a `private-key` configuration `Hash` when `key-data` is being used in conjunction with the `future` setting.
 
+### New features
+
+* **Support .yml extension for bolt config and inventory** ([#1500](https://github.com/puppetlabs/bolt/issues/1500))
```
```suggestion
* **Support .yml extension for Bolt config and inventory** ([#1500](https://github.com/puppetlabs/bolt/issues/1500))
```

PR #1511, opened 2020-01-06
(GH-1159) Rename bolt-server docker artifacts

### LINKED ISSUE #1159: Change Dockerfile name in root of repo
## Describe the Bug
The Dockerfile at the root of the Bolt repo is used for bolt-server. Users who stumble on the Bolt repo and expect that file to build an image that includes Bolt don't expect a testing container used for bolt-server. See https://github.com/puppetlabs/tasks-playground/issues/12

## To do
We should investigate if this container is even being used, and if it is how. Hopefully we can just remove this, or move it to somewhere that makes more sense.

lucywyman:
Having the bolt-server Dockerfile at the root of the Bolt repo is
confusing for users who are doing the hands-on lab and find this
Dockerfile by mistake instead of the hands-on lab one. This renames
Dockerfile to Dockerfile.bolt-server and docker-compose to
docker-compose-bolt-server in order to clarify that they're used for
bolt-server and not bolt.

Closes #1159


nicklewis:
The commit subject is now incorrect (though I see the body has been updated)

PR #1512, opened 2020-01-07
(GH-1338) Support node definitions when applying manifests

### LINKED ISSUE #1338: Support node statements with `bolt apply`
## Use Case
Main manifests are based on node statements which allow a single manifest to describe the entirety of a users infrastructure. These manifests can include `node` declarations which match the node name to core manifest. code.  Since bolt ignores `node` declarations `bolt apply manifests/site.pp --nodes all` ignores nodes statements and does not support these main manifests.

## Desired Solution

`node` statements should be supported in `bolt apply`

## Questions

* Should node statements be supported in `apply` blocks?
* How many users still rely on `node` declarations?
* Can we issue a warning when a node statement is encountered in manifest code?
* Do we need to support applying an entire directory to fully capture the masterless puppet use case.

## Work Around

Use a case statement based on the `$trusted.certname` fact rather than a `node declaration`

## Solution
We should support this behavior, `bolt apply <file.pp>` should behave the same as `puppet apply <file.pp>`. We should not support applying an entire directory. Supporting this will require that if we can't find a node we should not fail so that node definitions aren't required.

## Things to think about while implementing
* How does Puppet fail if it can't find a node definition? How does that bubble up, and can we fail gracefully?

nicklewis:
This adds support for node definitions from both `bolt apply` and within
apply blocks in a plan. Node definitions are still not allowed directly
in plans outside an apply block however, as there is no meaningful
behavior for them.

Supporting node definitions in apply blocks (vs `bolt apply`) depends on
https://github.com/puppetlabs/puppet/pull/7909 and the corresponding tests are
marked `pending` accordingly.

Closed #1338

### DIFF
--- a/spec/integration/apply_compile_spec.rb
+++ b/spec/integration/apply_compile_spec.rb
@@ -111,6 +111,22 @@ describe "passes parsed AST to the apply_catalog task" do
       expect(warn.count).to eq(1)
     end
 
+    it 'evaluates a node definition matching the node name' do
+      pending "puppet does not allow node definitions in plans"
+      result = run_cli_json(%w[plan run basic::node_definition] + config_flags)
+      report = result[0]['result']['report']
+      warn = report['catalog']['resources'].select { |r| r['type'] == 'Warn' }
+      expect(warn.count).to eq(1)
+    end
+
+    it 'evaluates a default node definition if none matches the node name' do
+      pending "puppet does not allow node definitions in plans"
+      result = run_cli_json(%w[plan run basic::node_default] + config_flags)
+      report = result[0]['result']['report']
+      warn = report['catalog']['resources'].select { |r| r['type'] == 'Warn' }
+      expect(warn.count).to eq(1)
+    end
+
     it 'logs messages emitted during compilation' do
       result = run_cli_json(%w[plan run basic::error] + config_flags)
       expect(result[0]['status']).to eq('success')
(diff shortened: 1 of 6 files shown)


spec/integration/apply_compile_spec.rb
```diff
@@ -111,6 +111,24 @@ def get_notifies(result, future = false)
       expect(warn.count).to eq(1)
     end
 
+    it 'evaluates a node definition matching the node name' do
+      pending "puppet does not allow node definitions in plans"
+      result = run_cli_json(%w[plan run basic::node_definition] + config_flags)
```
lucywyman: I think we need to add the `basic/plans/node_definition.pp` fixture too for this to pass once Puppet changes are available.

spec/integration/apply_compile_spec.rb
```diff
@@ -111,6 +111,24 @@ def get_notifies(result, future = false)
...
+      report = result[0]['result']['report']
+      warn = report['catalog']['resources'].select { |r| r['type'] == 'Warn' }
+      expect(warn.count).to eq(1)
+      expect(warn[0]['parameters']['message']).to eq('named node definition')
+    end
+
+    it 'evaluates a default node definition if none matches the node name' do
+      pending "puppet does not allow node definitions in plans"
+      result = run_cli_json(%w[plan run basic::node_default] + config_flags)
```
lucywyman: Same here.

spec/integration/apply_compile_spec.rb
nicklewis: Hah thanks I guess I forgot to commit those files but the test is pending so it was fine anyway 🙃

PR #1513, opened 2020-01-07
(GH-1208) Ensure Puppet agent version with apply_prep

### LINKED ISSUE #1208: Ensure puppet agent version with apply_prep()
## Use Case
From what I can tell, it looks like the `apply_prep()` function will only install puppet agent if it is not currently installed on the target. If it is not installed, it will installed the version specified (via plugin_hooks). However, it will not ensure the version specified is installed if there already is any version installed.

If an exact puppet agent version was specified for the `puppet_agent::install` task, I would like the `apply_prep()` function to ensure that version is installed on the targets.

There is currently a workaround, and that is to call `puppet_agent::install` task with the `run_task` function; specifying the desired parameters. However, it appears that it is still necessary to run `apply_prep` afterwards, so that all the node's facts can be gathered.

## Solution
We should remove the puppet_agent version check from the `apply_prep` function and just use the install task.
* The behavior of the install task should be that if the agent is installed and version 'any' is passed.
* If the agent is installed and the version doesn't match then it should try to install the specified version.

beechtom:
This updates the `apply_prep` plan function to ensure the specified
version of the Puppet agent is installed on the target. Previously, the
`apply_prep` function would only install the agent if the target did not
have any version of the agent installed and did not have the
`puppet-agent` feature set. Now, `apply_prep` will attempt to install
the agent on any target that does not have the `puppet-agent` feature
set, even if the install would cause an upgrade or downgrade to an
already-installed agent.

The following PR in puppetlabs/puppetlabs-puppet_agent should be merged in first:
https://github.com/puppetlabs/puppetlabs-puppet_agent/pull/463

Closes #1208

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,12 @@
   depends on features that aren't in this version of Bolt, so Bolt will now
   print a warning if it sees such keys.
 
+* **`apply_prep` plan function ensures Puppet agent version** ([#1208](https://github.com/puppetlabs/bolt/issues/1208))
+
+  The `apply_prep` plan function now attempts to install the specified version of the Puppet agent on a target 
+  even when a version of the agent is already installed. If the specified version of the agent cannot be installed, 
+  then `apply_prep` will error.
+
 ## Bolt 1.47.0
 
 ### Deprecations and removals
--- a/bolt-modules/boltlib/lib/puppet/functions/apply_prep.rb
+++ b/bolt-modules/boltlib/lib/puppet/functions/apply_prep.rb
@@ -7,12 +7,8 @@ require 'bolt/task'
 # installed using either the configured plugin or the `task` plugin with the
 # `puppet_agent::install` task.
 #
-# Agent detection will be skipped if the target includes the 'puppet-agent' feature, either as a
+# Agent installation will be skipped if the target includes the 'puppet-agent' feature, either as a
 # property of its transport (PCP) or by explicitly setting it as a feature in Bolt's inventory.
-#
-# If Bolt does not detect an agent on the target using the 'puppet_agent::version' task,
-# it will install the agent using either the configured plugin or the
-# task plugin.
 #
 # **NOTE:** Not available in apply block
 Puppet::Functions.create_function(:apply_prep) do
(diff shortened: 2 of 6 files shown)


bolt-modules/boltlib/lib/puppet/functions/apply_prep.rb
```diff
@@ -56,8 +56,7 @@ def run_task(targets, task, args = {}, options = {})
 
   # Returns true if the target has the puppet-agent feature defined, either from inventory or transport.
   def agent?(target, executor, inventory)
-    inventory.features(target).include?('puppet-agent') ||
```
lucywyman: We shouldn't remove this line - if users have the `puppet-agent` feature defined in their inventory for certain targets this will still attempt to install the agent, when it should skip them and not try to install the agent.

nicklewis:
Does this require a corresponding `puppetlabs/puppetlabs-puppet_agent` change to function properly or does that already do the right thing?

beechtom:
It depends. Currently, the `puppet_agent::install` task doesn't behave the same across platforms. Some platforms allow for upgrading and downgrading between major versions, others only between minor versions, and yet others only allow for upgrading.

I've reworked the bash version of `puppet_agent::install` to allow for similar behavior across platforms and can open a PR against that repo. I don't think it's necessarily part of this ticket, though, which was focused on removing the version check for `apply_prep`.

nicklewis:
This will need to bump the version of `puppetlabs-puppet_agent` once https://github.com/puppetlabs/puppetlabs-puppet_agent/pull/463 is merged and released.

beechtom:
Confirmed that we can handle release and bump to 3.0.0 for the `puppet_agent` module.

- [ ] Review and merge task documentation changes puppetlabs/puppetlabs-puppet_agent#465
- [ ] Review and merge changelog edits for v3.0.0 of `puppet_agent` puppetlabs/puppetlabs-puppet_agent#466
- [ ] Release v3.0.0 of `puppet_agent` to the forge
- [ ] Bump the version in Bolt's Puppetfile

CHANGELOG.md
```diff
@@ -9,6 +9,12 @@
   Human formatted results no longer show empty strings or JSON objects. When a result only has an `_options` key, and the value
   is an empty string or whitespace, a message will be displayed saying the action completed successfully with no result.
 
+* **`apply_prep` plan function ensures Puppet agent version** ([#1208](https://github.com/puppetlabs/bolt/issues/1208))
+
+  The `apply_prep` plan function now attempts to install the specified version of the Puppet agent on a target 
+  even when a version of the agent is already installed. If the specified version of the agent cannot be installed, 
+  then `apply_prep` will do nothing.
```
```suggestion
  then `apply_prep` will error.
```

spec/bolt_spec/run_spec.rb
```diff
@@ -107,18 +107,25 @@ def root_config
       } }
     end
 
+    def agent_inventory
+      { 'groups' => conn_inventory['groups'],
```
lucywyman: So the reason these tests are failing is that `apply_manifest` calls `apply_prep`, which used to see that an agent was installed and do nothing. Now it tries to install the agent, but doesn't know to run try as `root` instead of as `bolt`.

This change is modifying the inventory for the entire test to use an inventory that says that there's currently an agent installed on the targets, even when there's not (like after we uninstall the agent on line 120). I think a better approach to this (and kind of the intended approach) is to set the agent feature in the inventory after running the `puppet_agent::install` task on line 121, then remove the feature after uninstalling the agent on line 128.

lucywyman: You additionally shouldn't have to use a modified inventory for any of the tasks since those can't check for target features.

lucywyman:
Changelog needs a rebase

spec/bolt_spec/run_spec.rb
```diff
@@ -123,35 +123,43 @@ def root_config
 
     describe 'apply_manifest' do
       it 'should apply a manifest file' do
+        bolt_inventory['features'] = ['puppet-agent']
```
lucywyman: Sorry, I realize my line numbers were probably off after you deleted your variables! I meant for this to go in the `before(:all) do` and `after(:all) do` blocks.

beechtom: You can't use a `let` declaration in a `before(:all)` or `after(:all)` block, so there's no way to modify the `bolt_inventory` object used by the spec tests unless you do so inside the examples.

PR #1514, opened 2020-01-07
(GH-1431) Add support for trusted external facts

### LINKED ISSUE #1431: Support "external" facts in apply blocks.
How should bolt expose external facts when compiling apply blocks

## TODO
* Investigate if you had a manifest used from puppetserver that relied on external facts, is there a way to make that same manifest work with Bolt?
* The outcome of this is likely that we need to expose more ways to control trusted data

lucywyman:
This adds support for the new trusted external facts command. It adds a
configuration option `trusted-external-command` where users can
configure a path to their trusted external facts executable. We pass that
value through to the Puppet settings config of the same name. This
allows users to access trusted external facts in `$trusted['external']`.

Closes #1431

### DIFF
--- a/documentation/bolt_configuration_options.md
+++ b/documentation/bolt_configuration_options.md
@@ -25,6 +25,7 @@ ssh:
 -   `concurrency`: The number of threads to use when executing on remote targets. Default is `100`.
 -   `format`: The format to use when printing results. Options are `human` and `json`. Default is `human`.
 -   `hiera-config`: Specify the path to your Hiera config. The default path is `hiera.yaml` inside the [Bolt project directory](bolt_project_directories.md#).
+-   `trusted-external-command`: Specify the path to an executable on the Bolt controller that can produce [external trusted facts](https://puppet.com/docs/puppet/latest/release_notes_puppet.html#puppet-new-features-x.10.1). **Note:** external trusted facts are experimental in both Puppet and Bolt, and this API may change or be removed.
 -   `interpreters`: A map of an extension name to the absolute path of an executable, enabling you to override the shebang defined in a task executable. The extension can optionally be specified with the `.` character (`.py` and `py` both map to a task executable `task.py`) and the extension is case sensitive. The transports that support interpreter configuration are `docker`, `local`, `ssh`, and `winrm`. When a target's name is `localhost`, Ruby tasks run with the Bolt Ruby interpreter by default. This example demonstrates configuring Python tasks to run with a `python3` interpreter:
     ```yaml
     interpreters:
--- a/lib/bolt/applicator.rb
+++ b/lib/bolt/applicator.rb
@@ -82,7 +82,11 @@ module Bolt
     end
 
     def compile(target, ast, plan_vars)
-      trusted = Puppet::Context::TrustedInformation.new('local', target.name, {})
+      # This simplified Puppet node object is what .local uses to determine the
+      # certname of the target
+      node = Puppet::Node.from_data_hash('name' => target.name,
+                                         'parameters' => { 'clientcert' => target.name })
+      trusted = Puppet::Context::TrustedInformation.local(node)
       facts = @inventory.facts(target).merge('bolt' => true)
 
       catalog_input = {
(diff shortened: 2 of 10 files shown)


lib/bolt/applicator.rb
```diff
@@ -185,6 +187,13 @@ def validate_hiera_config(hiera_config)
       end
     end
 
+    def get_trusted_external(certname)
+      Puppet[:trusted_external_command] = @trusted_external
+      # Use .send to call private method
+      external_facts = Puppet::Context::TrustedInformation.send(:retrieve_trusted_external, certname)
```
adreyer: Can we call `.local` instead of `.new` to avoid the private method call?

lib/bolt/applicator.rb
```diff
@@ -185,6 +187,14 @@ def validate_hiera_config(hiera_config)
       end
     end
 
+    def get_trusted_external(certname)
+      Puppet[:trusted_external_command] = @trusted_external
```
adreyer: I'm slightly concerned about changing puppet settings here. Maybe we should do this in PAL?

lucywyman: We already set some Puppet settings in the catalog.rb - we could also pass the trusted external command through and set this there since it's executed in a separate process. Is your concern that this will pollute an existing Puppet environment?

lucywyman: Updated - I gave it it's own class function in PAL to avoid making a new Pal, and because I didn't think there were any existing functions that seemed right for it to belong in.

lib/bolt/pal.rb
```diff
@@ -67,6 +68,10 @@ def self.configure_logging
       Puppet.settings[:log_level] = 'debug'
     end
 
+    def self.with_trusted_external_command(trusted_external_command)
```
nicklewis: Should we do this in `with_puppet_settings` instead? The purpose of that method is to manage the lifecycle of our Puppet settings.

documentation/bolt_configuration_options.md
```diff
@@ -25,6 +25,7 @@ ssh:
 -   `concurrency`: The number of threads to use when executing on remote targets. Default is `100`.
 -   `format`: The format to use when printing results. Options are `human` and `json`. Default is `human`.
 -   `hiera-config`: Specify the path to your Hiera config. The default path is `hiera.yaml` inside the [Bolt project directory](bolt_project_directories.md#).
+-   `trusted-external-command`: Specify the path to an executable that can produce [external trusted facts](https://puppet.com/docs/puppet/latest/release_notes_puppet.html#puppet-new-features-x.10.1).
```
adreyer: After talking to the coremmunity team I think we should document this as experimental for now.

lucywyman: Updated here, in the CHANGELOG, and commit message.

lucywyman: Updated here, in the CHANGELOG, and in the commit message.

documentation/bolt_configuration_options.md
```diff
@@ -25,6 +25,7 @@ ssh:
 -   `concurrency`: The number of threads to use when executing on remote targets. Default is `100`.
 -   `format`: The format to use when printing results. Options are `human` and `json`. Default is `human`.
 -   `hiera-config`: Specify the path to your Hiera config. The default path is `hiera.yaml` inside the [Bolt project directory](bolt_project_directories.md#).
+-   `trusted-external-command`: Specify the path to an executable on the Bolt controller that can produce [external trusted facts](https://puppet.com/docs/puppet/latest/release_notes_puppet.html#puppet-new-features-x.10.1). **Note:** This option is experimental.
```
adreyer:
```suggestion
-   `trusted-external-command`: Specify the path to an executable on the Bolt controller that can produce [external trusted facts](https://puppet.com/docs/puppet/latest/release_notes_puppet.html#puppet-new-features-x.10.1). **Note:** external trusted facts are experimental in both puppet and bolt and this API may change or be removed.
```

lucywyman: Updated

PR #1515, opened 2020-01-08
(PE-28010) Calculate plan metadata in bolt-server

smcclellan:
This adds an endpoint, `/plans`, to bolt-server that calculates
plan metadata. In order to handle environments in puppet, a new PAL
class is needed, and since bolt runs standalone in PE, this environment
code should live in bolt for now.

If the query parameter `environment` is supplied, this environment-aware
PAL will be used.

### DIFF
--- a/lib/bolt_server/transport_app.rb
+++ b/lib/bolt_server/transport_app.rb
@@ -10,6 +10,10 @@ require 'bolt/task/puppet_server'
 require 'json'
 require 'json-schema'
 
+# These are only needed for the `/plans` endpoint.
+require 'puppet'
+require 'bolt_server/pe/pal'
+
 module BoltServer
   class TransportApp < Sinatra::Base
     # This disables Sinatra's error page generation
@@ -44,6 +48,9 @@ module BoltServer
       @executor = Bolt::Executor.new(0)
 
       @file_cache = BoltServer::FileCache.new(@config).setup
+
+      # This is needed until the PAL is threadsafe.
+      @pal_mutex = Mutex.new
 
       super(nil)
     end
@@ -271,6 +278,20 @@ module BoltServer
 
       aggregate = body['target'].nil?
       [200, result_set_to_status_hash(result_set, aggregate: aggregate).to_json]
+    end
+
+    get '/plans' do
+      @pal_mutex.synchronize do
+        environment = params['environment']
+        if environment.nil?
+          [400, '`environment` is a required argument']
+        else
+          pal = BoltServer::PE::PAL.new({}, params['environment'] || 'production')
+          plans = pal.list_plans.flatten
+          plan_info = plans.each_with_object({}) { |plan_name, acc| acc[plan_name] = pal.get_plan_info(plan_name) }
+          [200, plan_info.to_json]
+        end
+      end
     end
 
     error 404 do
--- a/lib/bolt_server/pe/pal.rb
+++ b/lib/bolt_server/pe/pal.rb
@@ -1,0 +1,90 @@
+# frozen_string_literal: true
+
+require 'bolt/pal'
+require 'bolt/util'
+
+module BoltServer
+  module PE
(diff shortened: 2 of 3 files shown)


lib/bolt_server/pe/pe_pal.rb
```diff
@@ -0,0 +1,135 @@
+require 'bolt/pal'
+require 'bolt/util'
+
+module PEBolt
```
adreyer: We try fairly hard to put modules and files in predictable paths.

so `BoltServer::PE::PE_PAL` or something.

lib/bolt_server/pe/pe_pal.rb
```diff
@@ -0,0 +1,135 @@
...
+        # so that we can have full control over all the entries in modulepath. In the future
+        # it's likely we will need to preceed _both_ Bolt::PAL::BOLTLIB_PATH _and_
+        # Bolt::PAL::MODULES_PATH which would be more complex if we tried to use @modulepath since
+        # we need to interject a path between *modulepath_dirs and Bolt::PAL::MODULES_PATH
+        @original_modulepath = modulepath_dirs
+        @modulepath = [PE_BOLTLIB_PATH, Bolt::PAL::BOLTLIB_PATH, *modulepath_dirs, Bolt::PAL::MODULES_PATH]
+      end
+    end
+
+    # There's a chance that some plan functions could conflict with built-in
```
adreyer: Do we actually need this in order to load metadata?

McdonaldSeanp: Ah, yeah we can definitely delete the next part

smcclellan: Is the value here on keeping pe_pal.rb identical with orchestrator's, or only putting the necessary parts in here?

lib/bolt_server/transport_app.rb
```diff
@@ -45,9 +45,30 @@ def initialize(config)
...
+      require 'bolt_server/pe/pe_pal'
+
       super(nil)
     end
 
+    get '/plans' do
+      @pal_mutex.synchronize do
+        environment = params['environment']
+        if environment.nil?
+          pal = Bolt::PAL.new(['/opt/puppetlabs/puppet/modules'], nil, nil)
```
donoghuc: I think we should make `envirnoment` required and error if it is not specified. I don't think we should implement this with Bolt's pal (also the modulepath for in the init method does not make sense).

lib/bolt_server/pe/pe_pal.rb
McdonaldSeanp: IMO just the necessary parts.

McdonaldSeanp:
@smcclellan looks like there's some rubocop failures

lib/bolt_server/transport_app.rb
```diff
@@ -45,9 +45,30 @@ def initialize(config)
 
       @file_cache = BoltServer::FileCache.new(@config).setup
 
+      # This is needed until the PAL is threadsafe.
+      @pal_mutex = Mutex.new
+      # These are only needed for the `/plans` endpoint.
+      require 'puppet'
```
donoghuc: Why are these requires in the init method?

lib/bolt_server/transport_app.rb
```diff
@@ -45,9 +45,30 @@ def initialize(config)
...
+      require 'bolt/pal'
+      require 'bolt_server/pe/pe_pal'
+
       super(nil)
     end
 
+    get '/plans' do
```
donoghuc: nit: I think this get route should be defined down with the others in the file and not with the other supporting methods.

lib/bolt_server/transport_app.rb
smcclellan: Moved.

smcclellan:
Added a stub-heavy unit test for the endpoint.

PR #1518, opened 2020-01-08
(GH-1319) Execute in run-as user's home dir when using run-as with SSH

### LINKED ISSUE #1319: bolt plan apply{} blocks fail when a user's homedir is on NFS with root squash enabled.
## Describe the Bug
I've created a relatively simple blan with an apply block to create a user & group on a server.  When I attempt to run bolt plan run <plan name> against a node I receive a permission denied on my home directory.
error snippet:

The task failed with exit code 1:
/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/module_tool/tar/gnu.rb:8:in `chdir': Permission denied @ dir_chdir - /opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/module_tool/tar/gnu.rb:8:in `chdir': Permission denied @ dir_chdir - /home/mlanghor (Errno::EACCES)
        from /opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/module_tool/tar/gnu.rb:8:in `unpack'
        from /tmp/a03a2d62-e790-4715-ac49-87a646bc809b/apply_catalog.rb:42:in `block in <main>'
        from /opt/puppetlabs/puppet/lib/ruby/2.4.0/tempfile.rb:295:in `open'
        from /tmp/a03a2d62-e790-4715-ac49-87a646bc809b/apply_catalog.rb:40:in `<main>'

On the remote server, my directory /home/mlanghor is an NFS mounted home directory, on which the NFS server has root squash enabled.  Therefore the root user cannot see or access files in my home directory.

I get similar issues with Puppet's exec resources if/when I run puppet agent when I'm in my home directory, though without exec resources it will run ok.

My full bolt command line I was attempting to run:  bolt plan run caiso_unix_utils::add_springboot_workspace nodes=server1 user=blah uid=20000 port_prefix=400 --modulepath=./ --run-as root --sudo-password

## Expected Behavior
when I run the bolt plan with --run-as root, it should not use my home directory and should execute successfully.

## Steps to Reproduce
Have a remote system using an NFS mount for user's home, with root squash enabled (so root cannot access your home directory).  Also homedir is set to 0750 permissions

## Environment
 - 1.33.0
 - CentOS 7.6 as a intiating client
 - RHEL 7.6 as the target node.

## Solution
* When running with a temporary directory over SSH or local transport (calling `with_tempdir`) we should cd into the specified `tmpdir` before executing.

nicklewis:
Previously, connecting as one user and running as a different user via 
run-as would cause the operation to execute in the home directory of the 
connected user instead of the run-as user. This difference doesn't usually
matter because the common run-as case involves escalating to root, in which
case root can read the directory anyway. However, when connecting as root
and then *dropping* to a less-privileged user, this would leave the process
in a situation where it might not have read access to its working
directory. This problem also arises when using NFS with the standard root
squash configuration, whereby root still can't read the home directories of
non-root users.

We now prepend a `cd` command with no argument before every command we 
execute when using run-as. This form of `cd` implicitly changes to the 
value of $HOME. We also now run with `sudo -H` which instructs sudo to 
reset $HOME to the home directory of the user we're switching to.

This issue caused particular problems for the libexec tasks used to apply
Puppet code. Those tasks use Puppet::ModuleTool::Tar which in some cases
will use `Dir.chdir` with a block while extracting files. This form of
`Dir.chdir` will change to a temporary directory to do its work, and then
will attempt to change back to the directory it was in previously. In cases
where that directory is unreadable by the current user, it would fail to
switch back and the task would error. We now ensure that those tasks will
be running from a current working directory where they are allowed read
access.

Also fixed an unrelated issue with defaulting sudo-password to the value of
password. It will now use the password value even if the password was set via
the URI, not just with the --password flag or the `password` field in the
inventory.

Fixes #1319

### DIFF
--- a/lib/bolt/transport/sudoable/connection.rb
+++ b/lib/bolt/transport/sudoable/connection.rb
@@ -77,8 +77,13 @@ module Bolt
         # a random string is echoed to stderr indicating that the stdin is available
         # for task input data because the sudo password has already either been
         # provided on stdin or was not needed.
-        def prepend_sudo_success(sudo_id, command_str)
+        def prepend_sudo_success(sudo_id, command_str, reset_cwd)
+          command_str = "cd && #{command_str}" if reset_cwd
           "sh -c 'echo #{sudo_id} 1>&2; #{command_str}'"
+        end
+
+        def prepend_chdir(command_str)
+          "sh -c 'cd && #{command_str}'"
         end
 
         # A helper to build up a single string that contains all of the options for
@@ -87,7 +92,9 @@ module Bolt
         # using the wrapper or when the task does not require stdin data.
         def build_sudoable_command_str(command_str, sudo_str, sudo_id, options)
           if options[:stdin] && !options[:wrapper]
-            "#{sudo_str} #{prepend_sudo_success(sudo_id, command_str)}"
+            "#{sudo_str} #{prepend_sudo_success(sudo_id, command_str, options[:reset_cwd])}"
+          elsif options[:reset_cwd]
+            "#{sudo_str} #{prepend_chdir(command_str)}"
           else
             "#{sudo_str} #{command_str}"
           end
--- a/spec/integration/apply_compile_spec.rb
+++ b/spec/integration/apply_compile_spec.rb
@@ -376,7 +376,7 @@ describe "passes parsed AST to the apply_catalog task" do
       it 'serializes Error objects in apply blocks' do
         result = run_cli_json(%w[plan run puppet_types::error] + config_flags)
         notify = get_notifies(result, true)
-        expect(notify[0]['title']).to eq("ApplyResult resource: The command failed with exit code 1")
+        expect(notify[0]['title']).to eq("ApplyResult resource: The command failed with exit code 127")
       end
 
       context 'when calling invalid functions in apply' do
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,15 @@
 # Changelog
 
+## Bolt NEXT
+
+### Bug fixes
+
+* **SSH commands will run from the home directory of the run-as user, not the connected user** ([#1518](https://github.com/puppetlabs/bolt/pull/1518))
+
+  Connecting via SSH and then switching users will now run as though it had
+  connected as the new user in the first place, using that user's home
+  directory as the working directory.
+
 ## Bolt 1.45.0
 
 ### Deprecations
--- a/lib/bolt/transport/ssh/connection.rb
+++ b/lib/bolt/transport/ssh/connection.rb
@@ -197,13 +197,13 @@ module Bolt
           if escalate
             if use_sudo
               sudo_exec = target.options['sudo-executable'] || "sudo"
-              sudo_flags = [sudo_exec, "-S", "-u", run_as, "-p", Sudoable.sudo_prompt]
+              sudo_flags = [sudo_exec, "-S", "-H", "-u", run_as, "-p", Sudoable.sudo_prompt]
               sudo_flags += ["-E"] if options[:environment]
               sudo_str = Shellwords.shelljoin(sudo_flags)
             else
               sudo_str = Shellwords.shelljoin(@target.options['run-as-command'] + [run_as])
             end
-            command_str = build_sudoable_command_str(command_str, sudo_str, @sudo_id, options)
+            command_str = build_sudoable_command_str(command_str, sudo_str, @sudo_id, options.merge(reset_cwd: true))
           end
 
           # Including the environment declarations in the shelljoin will escape
--- a/lib/bolt/transport/ssh/connection.rb
+++ b/lib/bolt/transport/ssh/connection.rb
@@ -36,7 +36,7 @@ module Bolt
 
           @sudo_password = @target.options['sudo-password']
           # rubocop:disable Style/GlobalVars
-          @sudo_password ||= @target.options['password'] if $future
+          @sudo_password ||= @target.password if $future
           # rubocop:enable Style/GlobalVars
 
           if target.options['private-key']&.instance_of?(String)
--- a/spec/integration/transport/ssh_spec.rb
+++ b/spec/integration/transport/ssh_spec.rb
@@ -320,6 +320,13 @@ describe Bolt::Transport::SSH do
       ssh.run_command(target, "rm #{dest}", sudoable: true, run_as: 'root')
     end
 
+    it "runs from the run-as user's home directory" do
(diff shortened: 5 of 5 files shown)


nicklewis:
I'm not sure whether this behavior is desirable for the local transport or not. Users may be depending on those commands running from their cwd.

lib/bolt/transport/sudoable/connection.rb
```diff
@@ -78,7 +78,11 @@ def execute(*_args)
         # for task input data because the sudo password has already either been
         # provided on stdin or was not needed.
         def prepend_sudo_success(sudo_id, command_str)
-          "sh -c 'echo #{sudo_id} 1>&2; #{command_str}'"
+          "sh -c 'echo #{sudo_id} 1>&2; cd && #{command_str}'"
+        end
+
+        def prepend_chdir(command_str)
+          "sh -c 'cd && #{command_str}'"
```
nicklewis: This will change the way we interpret commands, by now executing them in a shell. I'm not sure if that's desirable or not...

nicklewis: I'm not happy with this change. But I also don't really see how to accomplish this another way...

nicklewis: Oh this is how this already works. Perfect.

nicklewis:
Updated to only chdir with the SSH transport

lib/bolt/transport/sudoable/connection.rb
lucywyman: This does slightly change behavior - one side effect is that `&&` is respected now over SSH when using run as. A few examples of how behavior differs now:
```
lucy ~/githubs/bolt ((8e632fb...)) $ bolt command run "whoami && whoami" --run-as ubuntu --sudo-password puppet -t thirsty-soap.delivery.puppetlabs.net
CLI arguments ["run-as", "sudo-password"] may be overridden by Inventory: /home/lucy/githubs/bolt/inventory.yaml
Started on thirsty-soap.delivery.puppetlabs.net...
Finished on thirsty-soap.delivery.puppetlabs.net:
  STDOUT:
    ubuntu
    root
Successful on 1 target: thirsty-soap.delivery.puppetlabs.net
Ran on 1 target in 0.57 sec

lucy ~/githubs/bolt ((8e632fb...)) $ be bolt command run "whoami && whoami" --run-as ubuntu --sudo-password puppet -t thirsty-soap.delivery.puppetlabs.net
CLI arguments ["run-as", "sudo-password"] may be overridden by Inventory: /home/lucy/githubs/bolt/inventory.yaml
Started on thirsty-soap.delivery.puppetlabs.net...
Finished on thirsty-soap.delivery.puppetlabs.net:
  STDOUT:
    ubuntu
    ubuntu
Successful on 1 target: thirsty-soap.delivery.puppetlabs.net
Ran on 1 target in 0.57 sec

lucy ~/githubs/bolt ((8e632fb...)) $ be bolt command run "cd && pwd" --run-as ubuntu --sudo-password puppet -t thirsty-soap.delivery.puppetlabs.net
CLI arguments ["run-as", "sudo-password"] may be overridden by Inventory: /home/lucy/githubs/bolt/inventory.yaml
Started on thirsty-soap.delivery.puppetlabs.net...
Finished on thirsty-soap.delivery.puppetlabs.net:
  STDOUT:
    /home/ubuntu
Successful on 1 target: thirsty-soap.delivery.puppetlabs.net
Ran on 1 target in 0.62 sec

lucy ~/githubs/bolt ((8e632fb...)) $ bolt command run "cd && pwd" --run-as ubuntu --sudo-password puppet -t thirsty-soap.delivery.puppetlabs.net
CLI arguments ["run-as", "sudo-password"] may be overridden by Inventory: /home/lucy/githubs/bolt/inventory.yaml
Started on thirsty-soap.delivery.puppetlabs.net...
Failed on thirsty-soap.delivery.puppetlabs.net:
  STDERR:
    sudo: cd: command not found
Failed on 1 target: thirsty-soap.delivery.puppetlabs.net
Ran on 1 target in 0.56 sec
```

lucywyman: I think supporting `&&` is actually generally desirable since people are surprised it doesn't work. I wonder if supporting it just for `run-as` use cases will be annoying and confusing though, since it will work sometimes and not others with no explanation:
```
lucy ~/githubs/bolt ((8e632fb...)) $ be bolt command run "cd && pwd" --run-as ubuntu --sudo-password puppet -t thirsty-soap.delivery.puppetlabs.net
CLI arguments ["run-as", "sudo-password"] may be overridden by Inventory: /home/lucy/githubs/bolt/inventory.yaml
Started on thirsty-soap.delivery.puppetlabs.net...
Finished on thirsty-soap.delivery.puppetlabs.net:
  STDOUT:
    /home/ubuntu
Successful on 1 target: thirsty-soap.delivery.puppetlabs.net
Ran on 1 target in 0.59 sec

lucy ~/githubs/bolt ((8e632fb...)) $ be bolt command run "cd && pwd" -u ubuntu -p password -t thirsty-soap.delivery.puppetlabs.netCLI arguments ["user", "password"] may be overridden by Inventory: /home/lucy/githubs/bolt/inventory.yaml
Started on thirsty-soap.delivery.puppetlabs.net...
all authorization methods failed (tried none, publickey, password)
Failed on thirsty-soap.delivery.puppetlabs.net:
  Authentication failed for user ubuntu@thirsty-soap.delivery.puppetlabs.net
Failed on 1 target: thirsty-soap.delivery.puppetlabs.net
Ran on 1 target in 2.06 sec
```

nicklewis: It would _seem_ that we already support `&&` when not using run-as, so I think this is _just_ a bug fix in the run-as case and otherwise not a behavior change?

spec/integration/apply_compile_spec.rb
```diff
@@ -376,7 +376,7 @@ def get_notifies(result, future = false)
       it 'serializes Error objects in apply blocks' do
         result = run_cli_json(%w[plan run puppet_types::error] + config_flags)
         notify = get_notifies(result, true)
-        expect(notify[0]['title']).to eq("ApplyResult resource: The command failed with exit code 1")
+        expect(notify[0]['title']).to eq("ApplyResult resource: The command failed with exit code 127")
```
lucywyman: I don't understand why this changed?

nicklewis: It changed because we now wrap the command in `sh -c`. If you instruct `sudo` to run a non-existent command, it will fail with exit code 1. But if you wrap it in `sh`, then `sudo` successfully runs the shell, which _then_ can't find the command and exits with the standard 127.

PR #1519, opened 2020-01-09
(PE-28010) Exclude built in bolt modules in PE::PAL

donoghuc:
Explicitly remove the modulepath to the modules shipped in the bolt gem when calculating plan metadata for environments as these will not be available for puppet-server.


adreyer:
has there been any discussion about shipping these modules with PE? It came up in slack as one more thing for users to think about when installing a bolt project into PE.

donoghuc:
@adreyer Yeah that came up today as a direct result of this. This will be high on the list to talk over with new product owner.

PR #1521, opened 2020-01-10
(PE-27909) Split out plan metadata endpoints

McdonaldSeanp:
This commit updates the plan metadata endpoint in bolt-server so that the
/plans API does the following:

GET /plans returns _only_ the list of plan names, no associated metadata

GET /plans/:module_name/:plan_name returns the metadata for only one plan

GET /plans_metadata does what the original /plans API did: fetches all plans
and all associated metadata with each plan

These changes to the API are required because fetching all plans and all
metadata was less performant than we originally experienced, so the additions
of the less complex APIs (/plans and /plans/:module/:plan) allows downstream
consumers to only get the exact data they need in a more performant way

### DIFF
--- a/lib/bolt_server/transport_app.rb
+++ b/lib/bolt_server/transport_app.rb
@@ -180,6 +180,26 @@ module BoltServer
       [@executor.run_script(target, file_location, body['arguments'])]
     end
 
+    def in_pe_pal_env(environment)
+      if environment.nil?
+        [400, '`environment` is a required argument']
+      else
+        @pal_mutex.synchronize do
+          begin
+            pal = BoltServer::PE::PAL.new({}, environment)
+            yield pal
+          rescue Puppet::Environments::EnvironmentNotFound
+            [400, {
+              "class" => 'bolt/unknown-environment',
+              "message" => "Environment #{environment} not found"
+            }.to_json]
+          rescue Bolt::Error => e
+            [400, e.to_json]
+          end
+        end
+      end
+    end
+
     get '/' do
       200
     end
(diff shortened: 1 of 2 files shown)


lib/bolt_server/transport_app.rb
```diff
@@ -280,17 +295,25 @@ def make_winrm_target(target_hash)
...
-          plan_info = plans.each_with_object({}) { |plan_name, acc| acc[plan_name] = pal.get_plan_info(plan_name) }
-          [200, plan_info.to_json]
-        end
+      in_pe_pal_env(params['environment']) do |pal|
+        plans = pal.list_plans.flatten.map { |plan| { 'name' => plan } }
+        [200, plans.to_json]
+      end
+    end
+
+    get '/plans_metadata' do
```
smcclellan: I wonder if this and the `/plans` endpoint can be merged with a `?detail` argument defaulting to `true` or something?

McdonaldSeanp: Oh, yeah that might be a good idea

smcclellan: Actually `metadata` might be a better parameter name.

McdonaldSeanp:
@smcclellan updated

spec/bolt_server/transport_app_spec.rb
```diff
@@ -69,17 +70,52 @@ def file_data(file)
...
+        # Actually creating the EnvironmentNotFound error with puppet is difficult to do without
+        # puppet actually loaded with settings, so just stub out the error type
+        stub_const("Puppet::Environments::EnvironmentNotFound", StandardError)
+        expect(BoltServer::PE::PAL).to receive(:new).and_raise(Puppet::Environments::EnvironmentNotFound)
+        get(path)
+        expect(last_response.status).to eq(404)
+      end
+    end
+
+    describe '/plans?metadata=true' do
```
smcclellan: Can you merge this with `describe '/plans'`?

spec/bolt_server/transport_app_spec.rb
```diff
@@ -69,17 +70,52 @@ def file_data(file)
         )
     end
 
+    describe '/plans/:module_name/:plan_name' do
+      let(:path) { "/plans/foo/bar?environment=production" }
+      let(:fake_pal) { instance_double('BoltServer::PE::PAL') }
+
+      it 'returns properly from /plans' do
```
smcclellan: This string may need updating.

lib/bolt_server/transport_app.rb
```diff
@@ -180,6 +180,21 @@ def run_script(target, body)
...
+    def in_pe_pal_env(environment)
+      if environment.nil?
+        [400, '`environment` is a required argument']
+      else
+        @pal_mutex.synchronize do
+          begin
+            pal = BoltServer::PE::PAL.new({}, environment)
+            yield pal
+          rescue Puppet::Environments::EnvironmentNotFound
+            [404, "Environment #{environment} not found"]
```
smcclellan: I wasn't sure when I added these, but do we need to i18n these strings?

McdonaldSeanp: Since bolt-server isn't technically documented as public API I don't think so. Nothing from a customer perspective should be routing directly to bolt-server: everything should come via orchestrator. So I think we can leave these as normal strings and let orch take care of i18n

spec/bolt_server/transport_app_spec.rb
McdonaldSeanp: yeah

lib/bolt_server/transport_app.rb
```diff
@@ -280,16 +295,25 @@ def make_winrm_target(target_hash)
...
     end
 
+    get '/plans/:module_name/:plan_name' do
+      in_pe_pal_env(params['environment']) do |pal|
+        plan_info = pal.get_plan_info("#{params[:module_name]}::#{params[:plan_name]}")
+        [200, plan_info.to_json]
+      end
+    end
+
     get '/plans' do
```
smcclellan: Can we add `@param` docs to these methods to indicate which params are used (and their defaults).

smcclellan:
Looks great, happy to merge once these bits are addressed.

McdonaldSeanp:
@smcclellan updated again :)

McdonaldSeanp:
It's difficult to prioritize API docs since we may get rid of standalone bolt-server soon. But if we think it's important enough we can make a ticket to add those.

PR #1523, opened 2020-01-10
Clarify that node definitions only work in `bolt apply`

nicklewis:
Node definitions in `apply()` blocks depends on a change in Puppet which is
not yet released and so is not yet functional.

Also moved that changelog entry into Bolt NEXT since it hasn't actually been
released yet. It was added to Bolt 1.44.0 as part of an automatic merge
resolution gone awry.


lucywyman:
Sorry, this (appropriately) needs a CHANGELOG rebase since I merged Tom's PR.

nicklewis:
Rebased!

PR #1526, opened 2020-01-13
(PE-27909) Return plan metadata for `init` plan

donoghuc:
Previously when the plan metadata is requested for the special `init.pp` plan (with the same name as its parent module) the metadata endpoint would not trim the `init` from the name before calling in to get_plan_info. With this commit the endpoint has been updated to handle plans referenced only by `[module name]::init`.

### DIFF
--- a/lib/bolt_server/transport_app.rb
+++ b/lib/bolt_server/transport_app.rb
@@ -305,7 +305,13 @@ module BoltServer
     # @param environment [String] the environment to fetch the plan from
     get '/plans/:module_name/:plan_name' do
       in_pe_pal_env(params['environment']) do |pal|
-        plan_info = pal.get_plan_info("#{params[:module_name]}::#{params[:plan_name]}")
+        # Handle case where plan name is simply module name with special `init.pp` plan
+        plan_name = if params[:plan_name] == 'init'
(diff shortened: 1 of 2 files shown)


lib/bolt_server/transport_app.rb
```diff
@@ -305,7 +305,13 @@ def make_winrm_target(target_hash)
     # @param environment [String] the environment to fetch the plan from
     get '/plans/:module_name/:plan_name' do
       in_pe_pal_env(params['environment']) do |pal|
-        plan_info = pal.get_plan_info("#{params[:module_name]}::#{params[:plan_name]}")
+        # Handle case where plan name is simply module name with special `init.pp` plan
+        plan_name = if params[:plan_name] == 'init'
```
lucywyman: Maybe this is straightforward enough not to test, but it also probably couldn't hurt? https://github.com/puppetlabs/bolt/blob/master/spec/bolt_server/transport_app_spec.rb#L73

donoghuc: Yeah, for sure. Found this bug working on orchestrator-client and pushed a quick fix to show Sean the problem. Added a test :)

PR #1530, opened 2020-01-14
(GH-1315) Auto-generate configuration options docs

### LINKED ISSUE #1315: Config options docs should be autogenerated
The configuration options page (documentation/bolt_configuration_options.md) should be auto-generated rather than manually maintained.

beechtom:
This adds a new `docs:config_reference` rake task to auto-generate
documentation for Bolt's configuration options. It also removes the list
of available transport configuration options from the `Bolt::Config`
class to the respective transport classes.

Closes #1315


.rubocop.yml
```diff
@@ -95,3 +95,7 @@ Lint/SuppressedException:
 # Enforce LF line endings, even when on Windows
 Layout/EndOfLine:
   EnforcedStyle: lf
+
+Layout/HashAlignment:
+  EnforcedHashRocketStyle: [key, table]
```
beechtom: I added the table style as an acceptable layout so the options hashes were easier to read.

documentation/bolt_configuration_reference.md.erb
```diff
@@ -0,0 +1,155 @@
...
+<% @transports[:options].each do |transport, options| %>
+### <%= transport %>
+
+| Option | Description | Default |
+| ------ | ----------- | ------- |
+<% options.each do |option, desc| -%>
+| `<%= option %>` | <%= desc %> | <%= @transports[:defaults][transport][option].to_s || "" %> |
+<% end %>
+
+<% if transport == 'ssh' %>
```
beechtom: This just ensures that the OpenSSH options are listed right after ssh. Alternatively I could modify the options hash so ssh comes last and just list this at the end. Or if we don't really care about them being next to each other I can remove this check altogether and put it at the end of the page.

documentation/bolt_configuration_reference.md.erb
```diff
@@ -0,0 +1,155 @@
+# Bolt configuration options
+
+## Global configuration options
+
+Global configuration options can be set in the configuration file.
```
lucywyman:
```suggestion
Global configuration options can be set in the [Bolt configuration file](https://puppet.com/docs/bolt/latest/configuring_bolt.html).
```

documentation/bolt_configuration_reference.md.erb
```diff
@@ -0,0 +1,155 @@
...
+
+| Option | Description | Default |
+| ------ | ----------- | ------- |
+<% @global[:options].each do |option, desc| -%>
+| `<%= option %>` | <%= desc %> | <%= @global[:defaults][option].to_s || "" %> |
+<% end %>
+
+## Log file configuration options
+
+You can configure how the results of plan runs are captured in a log file. Different log files can be configured by specifying the log file's location using either `console` or a path to the log file, such as `~/.puppetlabs/bolt/debug.log`.
```
lucywyman:
```suggestion
You can configure how the results of plan runs are logged, both to a file and to the console. Different log files can be configured by specifying the log file's location using either `console` or a path to the log file, such as `~/.puppetlabs/bolt/debug.log`.
```

documentation/bolt_configuration_reference.md.erb
```diff
@@ -0,0 +1,155 @@
...
+log:
+  console:
+    level: error
+  ~/.puppetlabs/bolt/debug.log:
+    level: debug
+```
+
+## Plugin hooks configuration options
+
+The `plugin_hooks` section allows you to configure which plugins a specific hook should use for a specific target. This section is a hash where keys are hook names, and values specify and configure the plugin that that hook should use. There are two possible plugins: the `install_agent` plugin runs the `puppet_agent::install` task, and the `task` plugin runs your own custom task.
```
lucywyman:
```suggestion
The `plugin_hooks` section allows you to configure which plugins a specific hook should use. This section is a hash where keys are hook names, and values specify and configure the plugin for the hook to use.
```

documentation/bolt_configuration_reference.md.erb
```diff
@@ -0,0 +1,155 @@
...
+
+For now, the only configurable plugin hook is `puppet_library`.
```
lucywyman:
```suggestion
For now, the only configurable plugin hook is `puppet_library`.  There are two possible plugins for this hook: the `install_agent` plugin which runs the `puppet_agent::install` task, and the `task` plugin runs your own custom task.
```

documentation/bolt_configuration_reference.md.erb
```diff
@@ -0,0 +1,155 @@
...
+
+The default is to use the `puppet_agent` plugin with the agent service stopped:
```
lucywyman:
```suggestion
The default is to use the `install_agent` plugin with the agent service stopped:
```

documentation/bolt_configuration_reference.md.erb
```diff
@@ -0,0 +1,155 @@
...
+
+```yaml
+plugin_hooks:
+  puppet_library:
+    plugin: puppet_agent
```
lucywyman:
```suggestion
    plugin: install_agent
```

Rakefile
```diff
@@ -96,6 +96,30 @@ namespace :travis do
 end
 
 namespace :docs do
+  desc "Generate markdown docs for Bolt's transport configuration options"
+  task :config_reference do
+    @transports = { options: {}, defaults: {} }
+    @global = {}
```
lucywyman: I think `@transports` is the only var that gets calculated here. Everything else can be set in it's definition for readability and brevity.
```
@global = { options: Bolt::Config::OPTIONS,
defaults: Bolt::Config::DEFAULT_OPTIONS
}
...
```

documentation/bolt_configuration_reference.md.erb
```diff
@@ -0,0 +1,155 @@
...
+```
+
+```
+Host *.example.net
+  UserKnownHostsFile=~/.ssh/known_hosts
+  User root
+  Port 444
+```
+
+In this example, the SSH connection is configured to use the user and known hosts file defined in OpenSSH config and the port defined inBolt config.
```
lucywyman:
```suggestion
In this example, the SSH connection is configured to use the user and known hosts file defined in OpenSSH config and the port defined in Bolt's config.
```

lib/bolt/logger.rb
```diff
@@ -4,6 +4,18 @@
 
 module Bolt
   module Logger
+    OPTIONS = {
```
lucywyman: I don't think this is used in this file, we can move it to `config.rb`

PR #1533, opened 2020-01-15
(maint) Exclude path to module for bolt_server plan metadata

donoghuc:
This commit excludes the path to a module on disk when requesting plan metadata from bolt server. We do not want use this data for anything and do not want to expose these paths as they should not ever be modified by a human operator. This commit also fixes the `init` bug fixed for the /plans/:module/:plan endpoint (https://github.com/puppetlabs/bolt/commit/5fb7505c2154f6f46e466345e93879b86581a4f6#diff-5455f0d1ba7ece350e45ec171e8fe914) in the /plans endpoint.

### DIFF
--- a/lib/bolt_server/transport_app.rb
+++ b/lib/bolt_server/transport_app.rb
@@ -200,6 +200,19 @@ module BoltServer
       end
     end
 
+    def pe_plan_info(pal, module_name, plan_name)
+      # Handle case where plan name is simply module name with special `init.pp` plan
+      plan_name = if plan_name == 'init' || plan_name.nil?
+                    module_name
+                  else
+                    "#{module_name}::#{plan_name}"
+                  end
+      plan_info = pal.get_plan_info(plan_name)
+      # Path to module is meaningless in PE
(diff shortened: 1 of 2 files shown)


donoghuc:
With latest pe fetching from environment deployed with codemanager: 
```
[root@heavy-editing ~]# curl --silent -v --cacert $BOLT_CACERT --cert $BOLT_CERT --key $BOLT_KEY $BOLT_ROOT/plans/zvaaegrplgttqswkeftl/yavotgdjcxuujjnezfvl?environment=test 2>/dev/null
{"name":"zvaaegrplgttqswkeftl::yavotgdjcxuujjnezfvl","description":"Random description 5e4ce715af110232716bf4dd0ae8d42a48a1d2358f34a3956f","parameters":{"fubabaz":{"type":"String","description":"parameter decsription"},"my_arr":{"type":"Array","description":"parameter decsription"},"another_node":{"type":"TargetSpec","description":"parameter decsription"},"foobar":{"type":"String","description":"parameter decsription"}}}
```

PR #1540, opened 2020-01-15
(GH-1151) Move tests to GitHub Actions

### LINKED ISSUE #1151: Consolidate automated testing
## Use Case
Currently, PR testing is automated with Travis and Appveyor, once those tests have passed code can be merged to master. Once a PR is merged acceptance tests are run with Jenkins. Travis CI allows using the docker-compose stack for setting up networking between docker containers for complex integration testing. Appveyor is used to test windows features. Jenkins acceptance tests are used to test packages built after code is merged in to master against the templated VMs in vmpooler. Maintaining the configuration for three different platforms for testing adds complexity. We should investigate a solution for consolidating the testing platforms. 

Perhaps a combination of Github Actions and Azure could provide all of the features we need to be able to consolidate tests.

Last step is to evaluate the existing Github action and move off of Appveyor and Travis.

## Steps

- [ ] Create workflow for rubocop checks and docs checks
- [ ] Create workflow for travis unit/integration tests
- [ ] Create workflow for appveyor agentful/agentless tests

beechtom:
This consolidates all tests run on TravisCI and Appveyor in GitHub
Actions.

Closes #1151


beechtom:
It looks like the checks won't run on `puppetlabs/bolt` until the PR is actually merged in, which is different behavior than when I was creating a branch on my fork. You can see the passing checks here: https://github.com/beechtom/bolt/commit/e54b7e8f0f2944cdd293676f619065d71abab519/checks?check_suite_id=403161558

It would be helpful to get this merged in to make sure the checks run on any pushes made from a forked repo. Once we confirm things work as expected I'll work on getting the unit and integration tests pulled into GitHub Actions.

nicklewis:
It seems like `on: [push]` should be `on: [push, pull_request]`?

beechtom:
Looks like that was it. :)

beechtom:
The appveyor check will always fail since I removed `appveyor.yml` in this PR. Once this is merged in we should delete the appveyor project so it doesn't mark other PRs as failing.

.travis.yml
```diff
@@ -1,45 +0,0 @@
...
-  sync
-  export line_sep=$(printf '%*s' 80 | tr ' ' '#')
-  docker ps --all -q | xargs -L 1 -I % sh -c 'echo "\n${line_sep}\nContainer Logs - $(docker inspect % --format={{.Name}})\n${line_sep}\n"; docker logs %'
-  sleep 3s
-deploy:
-  provider: script
-  on:
-    repo: puppetlabs/bolt
-    # deploy the master branch or tag pushes
-    condition: $TRAVIS_BRANCH == "master" || -n $TRAVIS_TAG
```
donoghuc: Let's make sure to keep pushing the bolt server container to pcr internal onmerge to master. We use it in orch integration tests

.github/workflows/lint.yaml
```diff
@@ -0,0 +1,48 @@
...
+      - name: Checkout repository
+        uses: actions/checkout@v1
+      - name: Setup Ruby
+        uses: actions/setup-ruby@v1
+        with:
+          ruby-version: '2.5.x'
+      - name: Install rubocop
+        run: |
+          gem install rubocop
+          which rubocop
```
lucywyman: Do we need the `which rubocop`? More just curious, we don't need to remove it.

beechtom: Whoops, nope. That was from some testing to see if I could cache rubocop. I'll remove it.

developer-docs/kerberos.md
```diff
@@ -162,9 +162,9 @@ Useful tooling on the instance includes:
 * [`pwsh`](https://docs.microsoft.com/en-us/powershell/scripting/install/installing-powershell-core-on-linux?view=powershell-6) - PowerShell 6
 * [`omicli`](https://github.com/microsoft/omi/blob/master/Unix/cli/examples.txt) - client tool used to verify basic OMI server functionality
 
-#### TravisCI automated testing
+#### GitHub Actions automated testing
 
-At present, Travis setup will:
```
lucywyman: Grepping for Travis on this branch, it looks like this also needs to change in `developer-docs/releasing-bolt.md`.

.github/workflows/linux.yaml
```diff
@@ -0,0 +1,103 @@
...
+        if: steps.cache.outputs.cache-hit != 'true'
+        run: bundle install --jobs 4 --retry 3
+      - name: Pre-test setup
+        run: |
+          sudo curl -L https://github.com/docker/compose/releases/download/1.23.0/docker-compose-`uname -s`-`uname -m` -o /usr/local/bin/docker-compose
+          sudo chmod +x /usr/local/bin/docker-compose
+          echo 'runner:runner' | sudo chpasswd
+          sudo sh -c "echo 'Defaults authenticate' >> /etc/sudoers"
+          sudo sh -c "echo 'runner  ALL=(ALL) PASSWD:ALL' >> /etc/sudoers"
+          docker-compose -f spec/docker-compose.yml build --parallel ubuntu_node puppet_5_node puppet_6_node
```
lucywyman: Do we need to have the docker infrastructure up to run the unit tests? At least locally they work without the docker infra up.

developer-docs/kerberos.md
lucywyman: Also in doc/index.html and doc/file.README.html

developer-docs/kerberos.md
```diff
@@ -176,7 +176,7 @@ Despite having a correctly setup environment where OMI server can authenticate a
 
 #### Local development
 
-To run tests locally on Linux or OSX requires that the local Kerberos client be configured in the same way that Travis is, which includes:
+To run tests locally on Linux or OSX requires that the local Kerberos client be configured in the same way that GitHub Actions is, which includes:
```
```suggestion
To run tests locally on Linux or OSX requires that the local Kerberos client be configured in the same way that CI is, which includes:
```

developer-docs/kerberos.md
```diff
@@ -162,9 +162,9 @@ Useful tooling on the instance includes:
 * [`pwsh`](https://docs.microsoft.com/en-us/powershell/scripting/install/installing-powershell-core-on-linux?view=powershell-6) - PowerShell 6
 * [`omicli`](https://github.com/microsoft/omi/blob/master/Unix/cli/examples.txt) - client tool used to verify basic OMI server functionality
 
-#### TravisCI automated testing
+#### GitHub Actions automated testing
 
-At present, Travis setup will:
+At present, GitHub Actions setup will:
```
```suggestion
At present, CI setup will:
```

.github/workflows/linux.yaml
```diff
@@ -0,0 +1,103 @@
...
+          key: ${{ runner.os }}-gems-${{ hashFiles('**/Gemfile') }}
+      - name: Install gems
+        if: steps.cache.outputs.cache-hit != 'true'
+        run: bundle install --jobs 4 --retry 3
+      - name: Pre-test setup
+        run: |
+          sudo curl -L https://github.com/docker/compose/releases/download/1.23.0/docker-compose-`uname -s`-`uname -m` -o /usr/local/bin/docker-compose
+          sudo chmod +x /usr/local/bin/docker-compose
+          echo 'runner:runner' | sudo chpasswd
+          sudo sh -c "echo 'Defaults authenticate' >> /etc/sudoers"
```
lucywyman: I think we're fine to leave these, but I'm a little curious if they're necessary in the action containers. Basically the `travis` user in TravisCI automatically doesn't need authentication to use `sudo`, and for some tests we *need* the user we're escalating as to ask for a password. If 'runner' already asks for a password when using sudo, and if the user already has a predictable password, then we can get rid of some of these steps.

lucywyman: That applies to this line and the next one.

beechtom: The `runner` user in GitHub Actions uses passwordless sudo just like Travis, so we'd need to keep these steps in.

developer-docs/kerberos.md
beechtom: I'm not sure what `doc/index.html` and `doc/file.README.html` are?

.github/workflows/linux.yaml
```diff
@@ -0,0 +1,103 @@
+name: Linux
+
+on: [push, pull_request]
+
+env:
+  REGISTRY_HOST: pcr-internal.puppet.net
+  DOCKER_REGISTRY: pe-and-platform
+  BOLT_SUDO_USER: true
```
lucywyman: Like Cas pointed out, we need to make sure we're still pushing a container to the PCR registry on merges to master and new tags. The way this worked in Travis was that we would encrypt the PCR password using this workflow: https://docs.travis-ci.com/user/encryption-keys/#usage, which encrypts a value and makes it available as an environment variable in the CI runner.

It looks like for GH actions we'll need to add the password as a secret to this repo, then add that as the DOCKER_TOKEN environment variable and run the deploy.sh script. https://help.github.com/en/actions/automating-your-workflow-with-github-actions/creating-and-using-encrypted-secrets#using-encrypted-secrets-in-a-workflow

Since we only want to deploy the container on merge to master or a new tag, I think it might make the most sens to make a separate `deploy.yaml` (or maybe have it in this file? https://github.community/t5/GitHub-Actions/Depend-on-another-workflow/td-p/32317 that might be more trouble than it's worth).

developer-docs/kerberos.md
lucywyman: Oh, ha, they're leftover docs generated when I was testing a different PR. Nevermind!

beechtom: `developer-docs/releasing-bolt.md` references the Travis tests for another repo and shouldn't be changed.

beechtom: I'm not sure what `doc/index.html` and `doc/file.README.html` are?

.github/workflows/linux.yaml
beechtom: Pulled it out. Let's see what happens. :)

.github/workflows/deploy.yaml
```diff
@@ -0,0 +1,21 @@
...
+
+on:
+  pull_request:
+    types: [closed]
+    branches: [master]
+
+env:
+  REGISTRY_HOST: pcr-internal.puppet.net
+  DOCKER_REGISTRY: pe-and-platform
+  DOCKER_TOKEN: ${{ secrets.DOCKER_TOKEN }}
```
beechtom: TODO: Get a token for pcr-internal and save it as a github secret.

.github/workflows/linux.yaml
beechtom: That appears to be a no: https://github.com/puppetlabs/bolt/commit/9249c5ab652703ae7766fb01c2c2d32808715924/checks?check_suite_id=417505454

.github/workflows/linux.yaml
beechtom: Created a `deploy` workflow that triggers on `pull_request:closed` and runs the script if the PR was merged. Working on getting an access token to the docker registry and then I'll get that stored as a secret.

lucywyman: I think merge into the master branch might be a better trigger? We don't typically push directly to master but if we did we'd want that to deploy too. Plus we also want new tags to deploy as well.

.github/workflows/linux.yaml
lucywyman: Ah, ok, I see my confusion. There's a `unit` rake task that *doesn't* require the docker infra. The `travis:unit` (and thus `github:unit`) actually runs all but the most expensive tests, including integration level tests that do rely on the docker infra.

I'm not sure if "cheap" and "expensive" is a better differentiator? At least, that's *why* they're separate jobs. I wonder, since these rake tasks are only used by CI, if it makes sense to just combine them into 'integration tests' and invoke the 'ci' then the 'puppetserver' tasks in order. The reason we do the expensive ones last is so we can fail faster, but I'm not sure we need entirely separate tasks for that?

.github/workflows/linux.yaml
beechtom: Unfortunately there's no `merge` trigger for actions. The workaround is to use `pull_request:closed` and check that the PR was merged. I can set it up to trigger on `push` to master, though.

beechtom: As for running when a new tag is created... it looks like the `create` trigger covers that, but also gets triggered when new branches of the repo are created. We might be able to get away with this trigger for now, though, since Jenkins pushes to master after tagging, which would trigger the workflow.

.github/workflows/deploy.yaml
beechtom: Done :)

.github/workflows/linux.yaml
lucywyman: Ah ok I see - `if: github.event.pull_request.merged`. Yeah, I'm ok with waiting until it's a problem.

beechtom: Doing a little more digging around into other people's workflows, it looks like a push trigger to master should cover merging PRs and direct commits.

.github/workflows/linux.yaml
```diff
@@ -0,0 +1,103 @@
+name: Linux
+
+on: [push, pull_request]
+
+env:
+  REGISTRY_HOST: pcr-internal.puppet.net
```
lucywyman: I think we can remove these env vars now (not `BOLT_SUDO`, but the docker ones)

.github/workflows/lint.yaml
```diff
@@ -0,0 +1,46 @@
+name: Linting
```
lucywyman: Looking at how these workflows are displayed in the Actions view,  I think it makes sense to pull 'docs' and 'modules' into their own workflows so that when individual jobs fail it's easier to identify quickly what might've gone wrong. I think that also makes sense organizationally, since the modules spec isn't particularly related to Linux for example, and generating the docs isn't related to linting the Ruby code in the repo. I don't feel really strongly about this if there's a good reason to group them together, but I'm not sure what that reason would be.

beechtom: Created separate 'docs' and 'modules' workflows. Since the modules are tested on both Linux and Windows, there are two jobs for the 'modules' workflow... that makes me wonder if instead of 'Linux' and 'Windows' workflows we would have 'Fast*' and 'Slow*' workflows with Linux and Windows jobs...

* naming TBD

spec/integration/puppetfile_spec.rb
```diff
@@ -8,17 +8,28 @@
...
-  let(:root) { Dir.mktmpdir }
-  let(:boltdir) { File.join(root, 'Boltdir') }
-  let(:module_source) { Dir.mktmpdir }
+  let(:boltdir) { File.join(@root, 'Boltdir') }
+
+  # Usually temp dirs would just be created under the system temp dir,
+  # but this doesn't work on Windows images on GitHub Actions. Creating a
+  # temp dir in the current working dir gets around this issue.
+  around(:each) do |example|
+    Dir.mktmpdir(nil, Dir.pwd) do |root|
```
lucywyman: Can this just be `let(:root) { Dir.mktmpdir(nil, Dir.pwd) }` instead of being an around(:each)? mktmpdir doesn't require a block

spec/integration/transport/local_spec.rb
```diff
@@ -69,7 +69,8 @@
           expect(runner.upload(target, file.path, dest, run_as: user).message).to match(/Uploaded/)
           expect(runner.run_command(target, "cat #{dest}", run_as: user)['stdout']).to eq(contents)
           expect(runner.run_command(target, "stat -c %U #{dest}", run_as:  user)['stdout'].chomp).to eq(user)
-          expect(runner.run_command(target, "stat -c %G #{dest}", run_as:  user)['stdout'].chomp).to eq(user)
+          expect(runner.run_command(target, "stat -c %G #{dest}", run_as:  user)['stdout'].chomp)
+            .to eq(user) | eq('docker')
```
lucywyman: What's the condition where this is docker?

beechtom: The runner's primary group is `docker` and it can't be changed.

spec/integration/puppetfile_spec.rb
beechtom: It can, I believe it just requires an `after(:each)` to clean up the tmpdir. Using `mktmpdir` with a block will clean up the tmpdirs automatically.

spec/integration/transport/local_spec.rb
lucywyman: Will it ever be 'runner' then?

beechtom: No. But if these tests are run locally, then the group should match the user.

lucywyman: I don't think these tests will get executed locally, since you'd need to set the 'BOLT_SUDO_USER' env var in order to run tests tagged 'sudo', which I don't think we document anywhere or need to do unless we're developing something specific locally. Even if you do set that to true locally, you'd need to have the 'runner' user be the one executing the rspec tests, and I don't think we set up that infra anywhere? Basically, I think we can just check for the 'docker' group.

beechtom: Ah, got it 👍

PR #1541, opened 2020-01-16
(maint) Remove install_agent plugin from Bolt

lucywyman:
The `install_agent` plugin was originally a plugin builtin to Bolt which
configured how the agent was installed when `apply_prep` was run.
Specifically it used the `puppet_agent::install` task and stopped the
puppet-agent service once installed.

Once Bolt was able to load plugins from modules it made more sense for
the plugin to live alongside the task it used. In Bolt 1.35 we moved the
plugin to be in the puppetlabs-puppet_agent module and renamed it to be
the `puppet_agent` plugin. When we did this we neglected to remove the
`install_agent` plugin that was builtin to Bolt. This cleans up the last
step by removing the builtin `install_agent` plugin. This will break any
users who are still using the `install_agent` plugin, however inventory
plugins are still considered experimental.


CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # Changelog
 
+## Bolt Next
+
+### Deprecations
```
```suggestion
### Deprecations and removals
```

Just to be more consistent with previous changelog entries.

PR #1542, opened 2020-01-16
(GH-1493) Validate metadata when running or showing tasks

### LINKED ISSUE #1493: Warn when user has additional properties in task metadata
## Use Case
Users occasionally accidentally include extra keys at the top level of their task metadata, specifically if the `"parameters"` key is omitted. Users don't have a good way of knowing what isn't right, as we don't raise a warning or error when invalid keys are present.

We should raise a helpful error when invalid metadata keys are present. Setting `additional_properties: false` in the task schema may be the best way to do this, provided it raises a good error.

This seems like it could be intentional in some cases, given JSON doesn't have comments (though...do we want to allow people to have invalid metadata as comments?). Warning is a user-friendly way of notifying users when their metadata is incorrect, while still allowing Bolt to execute.

## Solution
* We should spike on if we can print a warning when tasks are loaded, which should happen only for tasks that are being run or 'show info'd.

nicklewis:
We now validate metadata during the construction of a Bolt::Task object, 
which means a warning will be logged when running or showing a task if that
task contains metadata fields that are unknown to Bolt.

In some cases, an unknown metadata field may signal a typo or that certain
keys (such as parameter specifications) are in the wrong place in the file.
In others, it may mean that new features have been added to the task spec
that the current version of Bolt doesn't understand and may not do the
right thing with. In both cases, it's helpful to alert the user to a
potential problem.

This also includes a refactor of the Bolt::Task class as well as a change to
make `bolt task show` use an instance of the Bolt::Task class (which allows
this warning to appear in that case).

Closes #1493


lib/bolt/task.rb
```diff
@@ -8,30 +8,33 @@ def initialize(target, task)
...
+
+    attr_reader :name, :files, :metadata, :remote
+
+    # name [String] name of the task
+    # files [Array<Hash>] where each entry includes `name` and `path`
+    # metadata [Hash] task metadata
+    def initialize(name, metadata = {}, files = [], remote = false)
+      @name = name
+      @metadata = metadata || {}
+      @files = files || []
```
beechtom: I don't think the conditional assignment is needed since the parameters have defaults.

lib/bolt/task.rb
```diff
@@ -8,30 +8,33 @@ def initialize(target, task)
...
-    :files,
-    :metadata
-  ) do
-    attr_reader :remote
-
-    def initialize(task, remote: false)
-      super(nil, nil, [], {})
-
+  class Task
+    METADATA_KEYS = %w[description extensions files implementations
```
beechtom: What is `extensions`? It's not listed in the [docs](https://puppet.com/docs/bolt/latest/writing_tasks.html#task-metadata-reference).

lib/bolt/task.rb
```diff
@@ -118,5 +121,33 @@ def select_implementation(target, provided_features = [])
...
+        metadata: @metadata
+      }
+    end
+
+    def validate_metadata
+      unknown_keys = metadata.keys - METADATA_KEYS
+
+      if unknown_keys.any?
+        msg = "Metadata for task '#{@name}' contains unknown keys: #{unknown_keys.join(', ')}."
+        msg += " This could be a typo in the task metadata or may result in incorrect behavior"
```
beechtom:
```suggestion
        msg += " This could be a typo in the task metadata or may result in incorrect behavior."
```

lib/bolt/task.rb
nicklewis: That's mostly for the `from_task_signature` path, where it will pass `nil` if the keys aren't present in the hash. I can add these defaults there instead.

lib/bolt/task.rb
nicklewis: It's reserved in the task spec as a place to put metadata that's specific to some task runner. It's not really used, but there might be a task or two out there that has some data under `extensions`.

https://github.com/puppetlabs/puppet-specifications/blob/master/tasks/README.md#options

lib/bolt/plugin.rb
```diff
@@ -83,7 +83,7 @@ def get_validated_task(task_name, params = nil)
           raise Bolt::Error.unknown_task(task_name) unless tasksig
 
           Bolt::Task::Run.validate_params(tasksig, params) if params
```
lucywyman: Do we still need this line (or function) with these changes?

lib/bolt/task.rb
```diff
@@ -118,5 +121,33 @@ def select_implementation(target, provided_features = [])
...
+
+    def to_h
+      {
+        name: @name,
+        files: @files,
+        metadata: @metadata
+      }
+    end
+
+    def validate_metadata
```
lucywyman: I think it'd be useful to have a test to verify that this warns with extra keys present - https://github.com/puppetlabs/bolt/blob/master/spec/integration/parsing_spec.rb seems like a good place for that? The `sample::params` task already has an extra top-level `anything` key, if we want to just use that.

lucywyman:
Changelog needs a rebase :P

nicklewis:
Rebased

lib/bolt/plugin.rb
nicklewis: Yes, this line validates that the parameters provided are acceptable by the task. It might make sense in the future to combine the logic on `Bolt::Task::Run` into `Bolt::Task` so this would be like:

```rb
task = Bolt::Task.from_task_signature(tasksig)
task.validate_params(params) if params
task
```

But for now I don't think it fits with this PR.

PR #1546, opened 2020-01-21
(maint) Bump required ruby version to 2.5

nicklewis:
Ruby 2.3 has been EOL since March 2019 and Ruby 2.4 will be EOL in March 2020.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,5 +1,5 @@
 AllCops:
-  TargetRubyVersion: 2.3
+  TargetRubyVersion: 2.5
   Exclude:
     - 'Boltdir/**/*'
     - 'vendor/**/*'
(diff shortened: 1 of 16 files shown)


nicklewis:
This shouldn't be merged until after we've done our final 1.x release.

donoghuc:
I think we will need to make a similar change in CJC. https://github.com/puppetlabs/ci-job-configs/blob/8bd8123384fad9eb8830969d845d01341bf42c26/jenkii/platform/projects/bolt.yaml#L116 Specifically the gem build and stage job will build the gem with 2.4.1. 

I'm not sure exactly what will happen when module CI is installing the Bolt gem for `BoltSpec::Run`. I think quite a bit of that CI is on an older than 2.5 ruby. I realize that is not really a concern here, but it is something to look out for.

nicklewis:
This may need to coordinate with https://github.com/puppetlabs/ci-job-configs/pull/6692. I'm not actually sure which jobs that change affects. We may not use RVM Ruby for running Bolt anywhere but only for setup / packaging steps.

nicklewis:
Rebased and should be good now.

PR #1550, opened 2020-01-22
(bug-fix) Prefer Etc.getpwuid.name to Etc.getlogin when available

lucywyman:
There are 3 tasks defined in `libexec` that are used internally by Bolt
to perform tasks on remote hosts (when you have a hammer...). Each of
these tasks takes a 'plugins' argument which is the contents of a
tarball of plugins from modules that are available to Bolt, in order to
load custom facts. The 'plugins' argument contents are written to a tar
file, then unpacked by `Puppet::ModuleTool::Tar` into a tempdir. The
last step of the unpacking is to `chown -R` the untarred directory to
the user who is executing the task.

Previously we figured out which user to `chown` the directory to using
`Etc.getlogin`, because it works on both *nix and Windows platforms.
However when any task was run as a different user than the one Bolt
connected to the host with (using `run-as`), `Etc.getlogin` would return
the user Bolt had *connected* as, rather than the user the task was
*executing* as (the `run-as` user). This is acceptable in cases where the
user is escalating to root because the plugins would be unpacked as the
non-root user then `chown -R non-root .` by root, which would succeed, and
root can still access the files if they are owned by another user.
When `run-as` is de-escalating privileges this causes a permissions
error, because the task tries to run `chown -R root .` as a non-root
user.

To fix the issue we now use the more accurate `Etc.getpwuid.name` on
platforms where it's available, and fall back to `Etc.getlogin` on
platforms where `Etc.getpwuid` returns nil. Since `Etc.getpwuid` returns
nil on Windows and the WinRM transport doesn't support `run-as`,
*and* Windows is tightly coupled with the WinRM transport, this isn't
expected to cause issues for users executing on Windows. However, if we ever
support `run-as` for Windows hosts, we may need a
more robust solution.

This bug was missed by our acceptance tests because they intended to
verify that a file in the apply block couldn't be created, and verified
this by checking that `Permission denied` was in the output and that the
file wasn't created. Both of these were true, but the *reason* they were
true was because of an issue described in #1319 instead of because the
apply block executed as the correct user. This PR modifies
the acceptance tests to verify that a file is successfully created by
the `run_as` user by verifying that it's owned by the expected user when
created. This ensures that the apply block succeeds as expected and
that further unintended errors don't slip through in the future.

### DIFF
--- a/acceptance/files/example_apply/plans/init.pp
+++ b/acceptance/files/example_apply/plans/init.pp
@@ -1,12 +1,12 @@
 plan example_apply (
-  TargetSpec $nodes,
+  TargetSpec $targets,
   String $filepath,
   Boolean $noop = false,
   Optional[String] $run_as = undef,
 ) {
-  $nodes.apply_prep
+  $targets.apply_prep
 
-  return apply($nodes, _noop => $noop, _run_as => $run_as, _catch_errors => true) {
+  return apply($targets, _noop => $noop, _run_as => $run_as, _catch_errors => true) {
     file { $filepath:
       ensure => directory,
     } -> file { "${filepath}/hello.txt":
--- a/acceptance/tests/apply_plan_ssh.rb
+++ b/acceptance/tests/apply_plan_ssh.rb
@@ -25,10 +25,11 @@ test_name "bolt plan run should apply manifest block on remote hosts via ssh" do
     scp_to(bolt, File.join(fixtures, 'example_apply'), "#{dir}/modules/example_apply")
   end
 
-  bolt_command = "bolt plan run example_apply filepath=#{filepath} nodes=#{targets}"
+  bolt_command = "bolt plan run example_apply filepath=#{filepath}"
   flags = {
     '--modulepath' => modulepath(File.join(dir, 'modules')),
-    '--format' => 'json'
+    '--format' => 'json',
+    '-t' => targets
   }
 
   teardown do
@@ -94,7 +95,7 @@ test_name "bolt plan run should apply manifest block on remote hosts via ssh" do
   end
 
   step "puppet service should be stopped" do
-    service_command = "bolt plan run example_apply::puppet_status -n #{targets}"
+    service_command = "bolt plan run example_apply::puppet_status"
     result = bolt_command_on(bolt, service_command, flags)
 
     assert_equal(0, result.exit_code,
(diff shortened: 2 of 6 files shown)


lucywyman:
To test this locally:
```
cd acceptance && bundle update
export GIT_FORK="lucywyman/bolt"
export GIT_BRANCH="maint-fix-acceptance"
export BEAKER_TESTS="./tests/apply_plan_ssh.rb,./tests/apply_plan_winrm.rb"
bundle exec rake test:git
```

acceptance/tests/apply_plan_ssh.rb
```diff
@@ -160,15 +169,20 @@
...
-        assert_match(/Permission denied/, result[0]['result']['_error']['msg'])
-
-        # Verify that files were not created on the target
-        on(node, "cat #{restricted_filepath}/hello.txt", acceptable_exit_codes: [1])
+        result = json.select { |n| n['node'] == host }.first
+        assert_equal('success', result['status'],
+                     "The task failed on #{host}")
+
+        stat = if node['platform'] =~ /osx/
+                 "stat -f \"%Su\" #{filepath}/hello.txt"
```
nicklewis: Are these escaped quotes necessary here? It looks like, when we interpolate this inside of more quotes, this just results in `%Su` being the only _unquoted_ part of the command.

lucywyman: Oo good catch

acceptance/files/example_apply/plans/init.pp
```diff
@@ -1,12 +1,12 @@
 plan example_apply (
-  TargetSpec $nodes,
+  TargetSpec $targets,
```
nicklewis: `apply_plan_winrm.rb` needs to be changed accordingly

acceptance/tests/apply_plan_winrm.rb
```diff
@@ -21,10 +21,11 @@
     scp_to(bolt, File.join(fixtures, 'example_apply'), "#{dir}/modules/example_apply")
   end
 
-  bolt_command = "bolt plan run example_apply filepath=#{filepath} nodes=winrm_nodes"
+  bolt_command = "bolt plan run example_apply filepath=#{filepath}"
   flags = {
     '--modulepath' => modulepath(File.join(dir, 'modules')),
-    '--format' => 'json'
+    '--format' => 'json',
+    '-t' => winrm_nodes.to_s
```
nicklewis: Why do we use the `winrm_nodes` variable here and not the string `"winrm_nodes"` as we were using before?

lucywyman: Oh my gosh I thought I was going insane. Didn't even think about winrm_nodes vs "winrm_nodes" :woman_facepalming:

PR #1551, opened 2020-01-22
(spike) Hide private plans from UI

donoghuc:
This commit is a spike for hiding plans with a @private yard tag from bolt plan show.

Questions:
1. Is Yard parsing every plan too much overhead just to get at that tag?
2. Is @private the correct tag name
3. Should we also ensure `@private true` or should just the existence of @private hide the plan.

### DIFF
--- a/lib/bolt/pal.rb
+++ b/lib/bolt/pal.rb
@@ -291,17 +291,35 @@ module Bolt
     end
 
     def list_plans
+      require 'puppet-strings'
+      require 'puppet-strings/yard'
       in_bolt_compiler do |compiler|
         errors = []
         plans = compiler.list_plans(nil, errors).map { |plan| [plan.name] }.sort
         errors.each do |error|
           @logger.warn(error.details['original_error'])
         end
-        plans
+        plans.reject { |plan| private_plan?(plan.first) }
       end
     end
 
-    def get_plan_info(plan_name)
+    def private_plan?(plan_name)
+      pp_path = plan_location(plan_name)[:path]
+      if File.exist?(pp_path)
+        PuppetStrings::Yard.setup!
+        YARD::Logger.instance.level = :error
+        YARD.parse(pp_path)
+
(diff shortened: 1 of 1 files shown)


lib/bolt/pal.rb
```diff
@@ -291,17 +291,35 @@ def get_task_info(task_name)
...
+    def private_plan?(plan_name)
+      pp_path = plan_location(plan_name)[:path]
+      if File.exist?(pp_path)
+        PuppetStrings::Yard.setup!
+        YARD::Logger.instance.level = :error
+        YARD.parse(pp_path)
+
+        plan = YARD::Registry.at("puppet_plans::#{plan_name}")
+
+        private_plan = plan.tag(:private) && plan.tag(:private).text == 'true'
```
nicklewis: It might make sense to just add this to `get_plan_info` rather than make a bespoke method for it.

donoghuc: Thats a good point. I guess by the time we have already done the Yard parsing, avoiding a little bit more data munging is not saving much energy.

donoghuc:
I think this approach will slow down https://github.com/puppetlabs/bolt/blob/9190aa37d2fd0704423442770b166c41b3415586/lib/bolt_server/transport_app.rb#L330-L348 too much.

lib/bolt/pal.rb
```diff
@@ -297,7 +297,7 @@ def list_plans
         errors.each do |error|
           @logger.warn(error.details['original_error'])
         end
-        plans
+        plans.reject { |plan| get_plan_info(plan.first)['private'] }
```
robdaemon: A full parse of the plan here on listing the plans will be expensive from both a computation and IO perspective.

adreyer:
we use `@api private` for puppet classes. The UI does not currently interact with that though.

donoghuc:
The real solution for this should probably involve caching the parsed Yard data. Closing this for now

PR #1554, opened 2020-01-23
(GH-1544) Remove empty strings and objects from results in human output

### LINKED ISSUE #1544: Human format should not print an empty object when `_output` is present.
If a task result contains only an `_output` key, for example if the task printed a string, the human formatter prints the message followed by an empty json object. It should instead not print this empty object if the original result had an `_output` key.

beechtom:
This changes the human outputter so that it does not display empty
strings or empty JSON objects. Previously, if a result only had an
`_output` key, the human outputter would dislay the value of `_output`,
even if it was an empty string or whitespace. If the only key present
was `_output`, the outputter would also display an empty JSON object
after the value of `_output`.

Now, if the result only contains an `_option` key and the value for that
key is an empty string or whitespace, the human outputter will display a
message stating the action completed successfully with no result.

Closes #1544

### DIFF
--- a/lib/bolt/outputter/human.rb
+++ b/lib/bolt/outputter/human.rb
@@ -96,12 +96,15 @@ module Bolt
           end
         end
 
-        if result.message
-          @stream.puts(remove_trail(indent(2, result.message)))
-        end
+        # Only print results if there's something other than empty string and hash
+        if result.value.empty? || (result.value.keys == ['_output'] && !result.message?)
+          @stream.puts(indent(2, "#{result.action.capitalize} completed successfully with no result"))
+        else
+          # Only print messages that have something other than whitespace
+          if result.message?
+            @stream.puts(remove_trail(indent(2, result.message)))
+          end
 
-        # There is more information to output
-        if result.generic_value
           # Use special handling if the result looks like a command or script result
           if result.generic_value.keys == %w[stdout stderr exit_code]
             unless result['stdout'].strip.empty?
@@ -112,7 +115,7 @@ module Bolt
               @stream.puts(indent(2, "STDERR:"))
               @stream.puts(indent(4, result['stderr']))
             end
-          else
+          elsif result.generic_value.any?
             @stream.puts(indent(2, ::JSON.pretty_generate(result.generic_value)))
           end
         end
--- a/lib/bolt/result.rb
+++ b/lib/bolt/result.rb
@@ -95,7 +95,6 @@ module Bolt
       @value = value || {}
       @action = action
       @object = object
-      @value_set = !value.nil?
       if error && !error.is_a?(Hash)
         raise "TODO: how did we get a string error"
       end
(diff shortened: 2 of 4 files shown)


lib/bolt/outputter/human.rb
```diff
@@ -101,7 +101,7 @@ def print_result(result)
         end
 
         # There is more information to output
-        if result.generic_value
+        unless result.generic_value.nil? || result.generic_value.empty?
```
nicklewis: I think we need to always output _something_. So if there's no `_error` or `_output`, we should print the generic value even if it's empty.

lib/bolt/outputter/human.rb
```diff
@@ -112,9 +113,11 @@ def print_result(result)
               @stream.puts(indent(2, "STDERR:"))
               @stream.puts(indent(4, result['stderr']))
             end
-          else
+          elsif result.generic_value.any?
             @stream.puts(indent(2, ::JSON.pretty_generate(result.generic_value)))
           end
+        else
+          @stream.puts(indent(2, "#{result.action.capitalize} completed successfully with no result"))
```
beechtom: Since `#print_result` is used by more than just tasks, I felt it was best to use the action from the result instead of printing "Task completed successfully..."

lib/bolt/result.rb
```diff
@@ -107,6 +107,10 @@ def message
       @value['_output']
     end
 
+    def message?
```
beechtom: I'm not really sure how I feel about this... since `Bolt::Result#message` can return either `nil` or a string (including just whitespace or empty strings), this simplifies checking if there's a 'useful' message in the result.

lib/bolt/result.rb
```diff
@@ -121,9 +125,7 @@ def status_hash
     end
 
     def generic_value
-      if @value_set
```
beechtom: This method was only used in one place, so it seemed weird to return a `nil` value, especially since we might no longer display empty JSON objects.

nicklewis: We can get rid of the `@value_set` variable then

lib/bolt/outputter/human.rb
```diff
@@ -96,12 +96,15 @@ def print_result(result)
           end
         end
 
-        if result.message
-          @stream.puts(remove_trail(indent(2, result.message)))
-        end
+        # Only print results if there's something other than empty string and hash
+        if !result.message? && result.generic_value.empty?
```
nicklewis: I actually think we just want to check if `result.value.empty?`, because we don't want to print this message if there was an `error_hash` _or_ in the `ApplyResult` case above.

nicklewis: Actually I guess it's more like `result.value.empty? || (result.value.keys == ['_output'] && !result.message?)` which is a little wonky. That is, if the value is completely empty _or_ if the only value is an empty message.

PR #1556, opened 2020-01-23
(GH-1433) Expose show_diff option for applying Puppet code

### LINKED ISSUE #1433: Expose puppet agent configuration options during apply block execution
## Use Case
This specifically came up regarding how file diffs are displayed but there are probably other puppet.conf settings we may want to expose.

an apply_settings section of `bolt.yaml`? 

exposing options to the `apply` statement.

## Solution
* Figure out if configuring `show-diff` (?) for the agent in apply blocks will even change the logs in the report.
* If `show-diff` does modify the logs, then figure out if this is something we should just set, or if we should expose it to the user to be configured.
* Are there other puppet config settings that would make sense to set in apply blocks - what is the list that would be useful? Come up with a recommended list of settings, and then we will design an interface for those separately.

lucywyman:
Bolt creates it's own Puppet context when applying Puppet code, which
doesn't inherit config from the system it's running on. Thus, when
applying Puppet code using Bolt users are limited to the default Puppet
settings.

This makes the `show_diff` [Puppet configuration
setting](https://puppet.com/docs/puppet/latest/configuration.html#showdiff)
to be configurable in Bolt's configuration file. When set, Bolt will set
the correspondent setting in it's Puppet context when applying Puppet
code, both from apply blocks in plans and from standalone manifests.

Additionally, this fixes a missing option in the help text, where `noop`
was not listed as a flag for `bolt task run`.

Closes #1433

### DIFF
--- a/lib/bolt/bolt_option_parser.rb
+++ b/lib/bolt/bolt_option_parser.rb
@@ -127,7 +127,7 @@ module Bolt
       when 'task'
         case action
         when 'run'
-          { flags: ACTION_OPTS + %w[params tmpdir],
+          { flags: ACTION_OPTS + %w[params tmpdir noop],
             banner: TASK_RUN_HELP }
         when 'show'
           { flags: OPTIONS[:global] + OPTIONS[:global_config_setters],
@@ -627,7 +627,7 @@ module Bolt
              "'success' nodes that succeeded in the last run.") do |rerun|
         @options[:rerun] = rerun
       end
-      define('--noop', 'Execute a task that supports it in noop mode') do |_|
+      define('--noop', 'See what changes Bolt will make without actually executing the changes') do |_|
         @options[:noop] = true
       end
       define('--description DESCRIPTION',
--- a/lib/bolt/config.rb
+++ b/lib/bolt/config.rb
@@ -34,10 +34,12 @@ module Bolt
   class Config
     attr_accessor :concurrency, :format, :trace, :log, :puppetdb, :color, :save_rerun,
                   :transport, :transports, :inventoryfile, :compile_concurrency, :boltdir,
-                  :puppetfile_config, :plugins, :plugin_hooks, :future, :trusted_external
+                  :puppetfile_config, :plugins, :plugin_hooks, :future, :trusted_external,
+                  :apply_settings
     attr_writer :modulepath
 
     OPTIONS = {
+      "apply_settings"           => "A map of Puppet settings to use when applying Puppet code",
       "color"                    => "Whether to use colored output when printing messages to the console.",
       "compile-concurrency"      => "The maximum number of simultaneous manifest block compiles.",
       "concurrency"              => "The number of threads to use when executing on remote targets.",
@@ -101,6 +103,16 @@ module Bolt
       "level"  => "`warn` for console, `notice` for file"
     }.freeze
 
+    APPLY_SETTINGS = {
+      "show_diff" => "Whether to log and report a contextual diff when files are being replaced. "\
+                     "See [Puppet documentation](https://puppet.com/docs/puppet/latest/configuration.html#showdiff) "\
+                     "for details"
+    }.freeze
+
+    DEFAULT_APPLY_SETTINGS = {
+      "show_diff" => false
+    }.freeze
+
     def self.default
       new(Bolt::Boltdir.new('.'), {})
     end
@@ -131,6 +143,7 @@ module Bolt
       @puppetfile_config = {}
       @plugins = {}
       @plugin_hooks = {}
+      @apply_settings = {}
 
       # add an entry for the default console logger
       @log = { 'console' => {} }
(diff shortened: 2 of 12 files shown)


lib/bolt/config.rb
```diff
@@ -101,6 +102,14 @@ class Config
       "level"  => "`warn` for console, `notice` for file"
     }.freeze
 
+    APPLY_SETTINGS = {
```
beechtom: Since `apply` is a new top-level field in the config file, it should be added as an option in `Bolt::Config::OPTIONS`.

lib/bolt/config.rb
```diff
@@ -101,6 +102,14 @@ class Config
...
+      "show_diff" => "Whether to log and report a contextual diff when files are being replaced. See https://puppet.com/docs/puppet/latest/configuration.html#showdiff"
```
beechtom:
```suggestion
      "show_diff" => "Whether to log and report a contextual diff when files are being replaced. See the [documenation](https://puppet.com/docs/puppet/latest/configuration.html#showdiff)."
```

Not using the URL will look a bit nicer when this is rendered as markdown.

lib/bolt/config.rb
```diff
@@ -224,14 +234,21 @@ def update_from_file(data)
       @trusted_external = if data.key?('trusted-external-command')
                             File.expand_path(data['trusted-external-command'], @boltdir.path)
                           end
+
+      if data.key?('apply_settings')
+        @apply_settings, extra_opts = data['apply_settings'].partition do |k, _|
+          APPLY_SETTINGS.keys.include?(k)
+        end.map(&:to_h)
+        unless extra_opts.empty?
+          @logger.warn("Apply settings #{extra_opts.keys.join(', ')} are not respected by puppet code applied by Bolt")
```
beechtom:
```suggestion
          @logger.warn("Apply settings #{extra_opts.keys.join(', ')} are not respected by Puppet code applied by Bolt")
```

lib/bolt/config.rb
```diff
@@ -224,14 +237,21 @@ def update_from_file(data)
       @trusted_external = if data.key?('trusted-external-command')
                             File.expand_path(data['trusted-external-command'], @boltdir.path)
                           end
+
+      if data.key?('apply_settings')
+        @apply_settings, extra_opts = data['apply_settings'].partition do |k, _|
+          APPLY_SETTINGS.keys.include?(k)
+        end.map(&:to_h)
+        unless extra_opts.empty?
+          @logger.warn("Apply settings #{extra_opts.keys.join(', ')} are not valid in Bolt")
```
beechtom: This warning will never get displayed since the logger isn't configured until after the config.

beechtom: This warning will never get displayed since the logger isn't configured until after the config.

nicklewis: For the `puppetfile` section, we just ignore any unknown settings. I'm not sure whether that's the right thing or not, but we should probably be consistent between the two of them, whether the behavior is silence, warning, or error.

lucywyman: I mostly added it because I imagine we'll be changing which settings are supported relatively frequently in the next few weeks, and thought people might just try to add all their preferred Puppet settings without checking which are supported and which aren't. I don't know of any settings that would have harmful effects if they weren't set as intended though, it might be easiest to remove this and add it if it's a problem.

lucywyman: With the exception of noop, which is already supported.

nicklewis: I think _most_ Puppet settings will either have no effect or have a harmful effect. So I do like the idea of not just supporting everything.

lib/bolt/bolt_option_parser.rb
```diff
@@ -627,7 +627,7 @@ def initialize(options)
              "'success' nodes that succeeded in the last run.") do |rerun|
         @options[:rerun] = rerun
       end
-      define('--noop', 'Execute a task that supports it in noop mode') do |_|
+      define('--noop', 'Execute a task that supports noop or bolt apply in noop mode') do |_|
```
nicklewis: This is an unwieldy sentence. It could be read "a task that supports bolt apply", and since we only show this flag for `bolt task run` and `bolt apply` anyway, mentioning `bolt apply` in the text feels redundant. It also isn't helpful if you don't already know what `noop` means. It might be better to say something generic like "Simulate the result without making any changes".

lucywyman: I ended up using a modified version of the `puppet apply` noop help text

spec/bolt/config_spec.rb
```diff
@@ -240,6 +240,15 @@
       }
       expect { Bolt::Config.new(boltdir, config) }.not_to raise_error
     end
+
+    it "warns when extra keys are present in 'apply_settings'" do
```
beechtom: This test is failing since we're not warning on invalid keys.

lucywyman: Oh ha, I made this change locally and it looks like it got caught by my local gitignore. Thanks!

nicklewis:
It looks like actually _no_ output appears for an apply block by default, so I think it's fine that the diff is the same.

PR #1557, opened 2020-01-24
(GH-608) Add support for multiple configuration files

### LINKED ISSUE #608: Bolt should use a configuration inheritance model similar to the one used by 'git'

beechtom:
This change adds support for two new configuration files. In order from
highest precedence to lowest, the configuration files are:

- Project configuration

  `<boltdir>/bolt.yaml`

- User configuration

  `~/.puppetlabs/etc/bolt/bolt.yaml`

- System-wide configuration

  *nix: `/etc/puppetlabs/bolt/bolt.yaml`

  Windows: `%PROGRAMDATA%\PuppetLabs\bolt\etc\bolt.yaml`

When multiple configuration files are found, Bolt will employ the
following merge strategy:

- Shallow merge options that accept hashes

- Overwrite options that do not accept hashes

- Deep merge transport configuration options (e.g. `ssh`, `winrm`)

- Shallow merge configuration for individual plugins configured under
the `plugins` option.

Closes #608

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,11 @@
 
   The `prompt` plugin now requires a `message` option.
 
+* **Support for system-wide and user-level configuration files** ([#608](https://github.com/puppetlabs/bolt/issues/608))
+
+  Bolt now supports system-wide and user configuration files, in addition to the existing project configuration file.
+  File precedence and merge strategy can be found in the [Bolt configuration docs](https://puppet.com/docs/bolt/latest/configuring_bolt.html).
+
 ## Bolt 1.47.0
 
 ### Deprecations and removals
--- a/documentation/configuring_bolt.md
+++ b/documentation/configuring_bolt.md
@@ -21,13 +21,13 @@ You can place `bolt.yaml` configuration files at the following paths, from highe
   
   The `bolt.yaml` file in a project directory applies settings to that project specifically. Matching settings at the project level override those at the user and system-wide levels.
 
-  `boltdir/bolt.yaml` or `<MY_PROJECT_NAME>/bolt.yaml`
+  `Boltdir/bolt.yaml` or `<MY_PROJECT_NAME>/bolt.yaml`
 
   > **Note:** The project configuration file is loaded from the [Bolt project directory](bolt_project_directories.md). The default project directory is `~/.puppetlabs/bolt/`.
 
 - **User**
   
-  The `bolt.yaml` file at the user level applies settings only to that user. Matching settings at the user level are overridden by project-level settings, but take precedent over system-wide settings.  
+  The `bolt.yaml` file at the user level applies settings only to that user. Matching settings at the user level are overridden by project-level settings, but take precedence over system-wide settings.  
 
   `~/.puppetlabs/etc/bolt/bolt.yaml`
 
--- a/documentation/configuring_bolt.md
+++ b/documentation/configuring_bolt.md
@@ -1,22 +1,156 @@
 # Configuring Bolt
 
-Create a configuration file to store and automate the command-line flags you use every time you run Bolt.
+You can configure Bolt's options and features at a project level, a user level, or a system-wide level. 
 
-Configuration for Bolt is loaded from the [Bolt project directory](bolt_project_directories.md#). The default project directory is `~/.puppetlabs/bolt/`. Configure global options (i.e. the modulepath) at the top level of `<boltdir>/bolt.yaml`, and configure transport-specific options for each transport.
+Some options and features you can configure include:
+- global options
+- transports 
+- plugins
+- PuppetDB connections
+- log files 
 
-Bolt config uses the following precedence, from highest precedence (cannot be overridden) to lowest:
-- Target URI (i.e. ssh://user:password@hostname:port)
-- [Inventory file](inventory_file.md) options
-- Command line flags
-- Config file options
-- SSH config file options (`~/.ssh/config`, if using SSH)
+For a complete list of Bolt settings, see the [configuration reference](bolt_configuration_reference.md).
 
--   **[Project directories](bolt_project_directories.md#)**  
- Bolt runs in the context of a project directory or a `Boltdir`. This directory contains all of the configuration, code, and data loaded by Bolt.
--   **[Bolt configuration options](bolt_configuration_options.md)**  
-Your Bolt configuration file can contain global and transport options.
--   **[Using Bolt with Puppet Enterprise](bolt_configure_orchestrator.md)**  
-If you're a Puppet Enterprise (PE) customer, you can configure Bolt to use the PE orchestrator and perform actions on managed targets. Pairing PE with Bolt enables role-based access control, logging, and visual reports in the PE console.
--   **[Connecting Bolt to PuppetDB](bolt_connect_puppetdb.md)**  
-Configure Bolt to connect to PuppetDB.
+## Configuration files
 
+To configure Bolt, create a Bolt configuration file named `bolt.yaml` in one of the locations listed below. If a configuration file is not located at the expected path, or the file is empty, Bolt does not load the file. 
+
+You can place `bolt.yaml` configuration files at the following paths, from highest precedence to lowest:
+
+- **Project**
+  
+  The `bolt.yaml` file in a project directory applies settings to that project specifically. Matching settings at the project level override those at the user and system-wide levels.
+
+  `boltdir/bolt.yaml` or `<MY_PROJECT_NAME>/bolt.yaml`
+
+  > **Note:** The project configuration file is loaded from the [Bolt project directory](bolt_project_directories.md). The default project directory is `~/.puppetlabs/bolt/`.
+
+- **User**
+  
+  The `bolt.yaml` file at the user level applies settings only to that user. Matching settings at the user level are overridden by project-level settings, but take precedent over system-wide settings.  
+
+  `~/.puppetlabs/etc/bolt/bolt.yaml`
+
+- **System-wide**
+
+  Settings in a system-wide config file apply to all users running Bolt, regardless of the Bolt project directory. However, matching settings at the project or user level override system-wide settings.
+  
+  The `bolt.yaml` file at the system-wide level applies settings only to that user.
+
+  \*nix: `/etc/puppetlabs/bolt/bolt.yaml`
+
+  Windows: `%PROGRAMDATA%\PuppetLabs\bolt\etc\bolt.yaml`
+
+## Configuration precedence
+
+Bolt uses the following precedence when interpolating configuration settings, from highest precedence to lowest:
+
+  - Target URI (i.e. ssh://user:password@hostname:port)
+  - [Inventory file](inventory_file.md) options
+  - [Command line flags](bolt_command_reference.md)
+  - Project configuration file
+  - User configuration file
+  - System-wide configuration file
+  - SSH configuration file options (e.g. `~/.ssh/config`)
+
+## Merge strategy
+
+When merging configurations, Bolt's strategy is to shallow merge any options that accept hashes and to overwrite any options that do not accept hashes. There are two exceptions to this strategy:
+
+- [Transport configuration](bolt_configuration_reference.md#transport-configuration-options) (e.g. `ssh`, `winrm`) is deep-merged.
+
+- [Plugin configuration](using_plugins.md#configuring-plugins) is shallow-merged for _each individual plugin_.
+
+### Transport configuration merge strategy
+
+Transport configuration is deep merged. 
+
+For example, given this SSH configuration in a project configuration file:
+
+```yaml
+ssh:
+  user: bolt
+  password: bolt
+  host-key-check: false
+```
+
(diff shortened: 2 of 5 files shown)


beechtom:
This currently isn't gated on `future` since I believe we were intending for this to be a 2.0 feature. If that's changed I can update it to be future-only.

lib/bolt/config.rb
```diff
@@ -387,5 +423,95 @@ def casefold(path)
         l =~ /[A-Za-z]/ ? "[#{l.upcase}#{l.downcase}]" : l
       end.join
     end
+
+    def config_loaded
+      msg = <<~MSG
+        CONFIGURATION FILES
```
nicklewis: Okay I understand the concern about verbosity. I don't think we need to explain _how_ config is merged here, only that we're loading it from multiple places. The idea is to address users' confusion about where values that aren't in their `bolt.yaml` came from (that is, _discovering_ that there are multiple config files at play). Understanding how the final merged config was actually generated is something to consult the documentation for.

I was imagining something like:

```rb
@logger.debug("Loaded config from [...], [...]")
```

CHANGELOG.md
```diff
@@ -1,5 +1,14 @@
 # Changelog
 
+## Bolt Next
+
+### New features
+
+* **Support for multiple configuration files** ([#608](https://github.com/puppetlabs/bolt/issues/608))
```
```suggestion
* **Support for user-level and system-wide configuration files** ([#608](https://github.com/puppetlabs/bolt/issues/608))
```

lucywyman: This is nit-picky, but we don't support merging arbitrary configuration files - we still load config from specific places, now there's just more than one config file that gets loaded :P

beechtom: Not not-picky at all! :) Updated.

lib/bolt/config.rb
```diff
@@ -107,17 +102,37 @@ def self.default
...
-      new(boltdir, data, overrides)
+      data = []
+      data << { filepath: system_path, data: system_data } unless system_data.empty?
+      data << { filepath: user_path, data: user_data } unless user_data.empty?
+
+      data
     end
 
-    def initialize(boltdir, config_data, overrides = {})
+    def initialize(boltdir, config_data, overrides = {}, defaults = [])
```
nicklewis: It might make more sense to allow `config_data` to be either a hash or an array of hashes and apply the defaults that way. It feels a bit strange to make one of the config files "special".

lib/bolt/config.rb
```diff
@@ -239,6 +263,18 @@ def update_from_file(data)
     end
     private :update_from_file
 
+    # Shallow merge config for individual plugins
+    def update_plugins(data)
```
nicklewis: I believe this is equivalent to:

```rb
@plugins.merge!(data) do |key,val1,val2|
  v1.merge(v2)
end
```

The block will only be invoked for keys that exist in both hashes and is used to determine the merge strategy.

lib/bolt/config.rb
```diff
@@ -141,6 +157,11 @@ def initialize(boltdir, config_data, overrides = {})
         @transports[key] = transport.default_options
       end
 
+      defaults.each do |default|
+        @config_files << default[:filepath]
+        update_from_file(default[:data])
```
nicklewis: It might make sense to do the merging at the hash level and then call `update_from_file` once rather than calling for each level. That way we could have a single place that just implements the merge behavior. I'm not confident on that one though.

beechtom: I'm not sure either. I mostly based this off of Alex's comment: https://github.com/puppetlabs/bolt/pull/1484#issuecomment-565624774

lib/bolt/config.rb
beechtom: Oh, I didn't realize merge accepted a block with both values. Ruby you fancy.

lib/bolt/config.rb
nicklewis: If we merged all the data together, we could also have the 0th layer be the defaults. Then we could avoid all the `if data.key?` checks when assigning the instance variables. It might be worth a spike at least, to see if it simplifies things.

I was imagining something like:

```rb
merged_config = {
  # ... defaults here
}

config_datas.each do |data|
  merged_config.merge!(data) do |key, val1, val2|
    case key
    when <something we want to shallow or deep merge>
      # do the merge
    else
      # use the newer value
      val2
    end
  end
end
```

And then it will be very clear how data is merged.

nicklewis: We could then even transform the "overrides" into a hash and merge that on top too, if it uses the same merge strategy.

We might also be able to get rid of a bunch of instance variables if we just make the accessor methods read from the merged config hash.

lib/bolt/config.rb
```diff
@@ -107,17 +102,37 @@ def self.default
...
+    def self.load_defaults
+      # Lazy-load expensive gem code
+      require 'win32/dir' if Bolt::Util.windows?
+
+      system_path = if Bolt::Util.windows?
+                      Pathname.new(File.join(Dir::COMMON_APPDATA, 'PuppetLabs', 'bolt', 'etc', 'bolt.yaml'))
+                    else
+                      Pathname.new(File.join('/etc', 'puppetlabs', 'bolt', 'bolt.yaml'))
+                    end
+      user_path = Pathname.new(File.join('~', '.puppetlabs', 'etc', 'bolt', 'bolt.yaml'))
```
lucywyman: I think we might need to expand this path? Data isn't getting loaded from this file locally for me...

lib/bolt/config.rb
beechtom: Updated to merge the config hashes prior to calling `update_from_file`. It's definitely possible to have a default hash instead of the instance variables, but it'll take a bit more work that I think could be part of a larger effort to refactor the class.

lib/bolt/config.rb
```diff
@@ -214,6 +268,10 @@ def update_from_file(data)
         update_logs(data['log'])
       end
 
+      if data['plugins'].is_a?(Hash)
```
nicklewis: Are any of these changes in `update_from_file` still needed? Since all we've changed now is the input data, it seems like this method should behave the same as it did before.

lib/bolt/config.rb
```diff
@@ -154,12 +181,39 @@ def initialize(boltdir, config_data, overrides = {})
...
+        acc.merge!(config[:data]) do |key, val1, val2|
+          case key
+          # Plugin config is shallow merged for each plugin
+          when 'plugins'
+            val1.merge!(val2) { |_, v1, v2| v1.merge(v2) }
+          # Transports are deep merged
+          when 'ssh', 'winrm', 'local', 'remote', 'docker', 'pcp'
+            Bolt::Util.deep_merge(val1, val2)
+          # Hash values are shallow mergeed
+          when 'puppetdb', 'puppetfile_config', 'plugin_hooks', 'apply_settings', 'log'
```
nicklewis: The setting in the config file is called `puppetfile`. The variable is `@puppetfile_config` because we already have a `Config#puppetfile` method

nicklewis: But [the gist](https://gist.github.com/beechtom/f7b758893796d61a23c8ab868cf8b062) said we were going to overwrite puppetfile settings, not merge them. So in that case, this typo is making it behave correctly.

lib/bolt/config.rb
```diff
@@ -154,12 +181,39 @@ def initialize(boltdir, config_data, overrides = {})
...
       validate
     end
 
+    # Merge configuration
+    # Precedence from highest to lowest is: project, user-level, system-wide
+    def merge_config_data(config_data)
+      config_data.each_with_object({}) do |config, acc|
+        @config_files << config[:filepath] if config[:filepath]
+
+        acc.merge!(config[:data]) do |key, val1, val2|
```
nicklewis: We should change these `merge!` calls to just `merge`. That requires getting rid of `each_with_object`, but that's straightforward if the `@config_files << [...]` is removed from this method.

lib/bolt/config.rb
```diff
@@ -154,12 +181,39 @@ def initialize(boltdir, config_data, overrides = {})
...
+          case key
+          # Plugin config is shallow merged for each plugin
+          when 'plugins'
+            val1.merge!(val2) { |_, v1, v2| v1.merge(v2) }
+          # Transports are deep merged
+          when 'ssh', 'winrm', 'local', 'remote', 'docker', 'pcp'
```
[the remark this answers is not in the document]
beechtom: I tried that, but I was getting a syntax error:
`SyntaxError ((irb):1: syntax error, unexpected '\n', expecting '.' or &. or :: or '[')`

Not really sure what the issue there might be. Everything I found by Googling was from 2013 or earlier and showed exactly that syntax. :/

lib/bolt/config.rb
beechtom: Whoops. As Bob Ross would say - a happy little accident.

lib/bolt/config.rb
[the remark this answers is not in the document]
nicklewis: Can you try it and copy-paste exactly the code you're using?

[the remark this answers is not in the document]
beechtom: Well, now it's working... it's very possible that there was a bug in my code that caused issues with the case statement. I'll push a commit with the changes.

lib/bolt/config.rb
```diff
@@ -254,6 +307,18 @@ def update_from_file(data)
     end
     private :update_from_file
 
+    # Shallow merge config for individual plugins
+    def update_plugins(data)
```
nicklewis: This method is now unused.

lib/bolt/config.rb
```diff
@@ -402,5 +467,12 @@ def casefold(path)
         l =~ /[A-Za-z]/ ? "[#{l.upcase}#{l.downcase}]" : l
       end.join
     end
+
+    def config_loaded
+      msg = <<~MSG
+        Loaded configuration from: '#{[*@config_files, boltdir.config_file].join("', '")}'
```
nicklewis: This can just be `config_files` now, since that will contain the Boltdir config path as well.

nicklewis: It also might make more sense for this to just be inline in the CLI class where it's used instead of being a method on the `Config` object.

documentation/configuring_bolt.md
```diff
@@ -1,22 +1,156 @@
...
+
+To configure Bolt, create a Bolt configuration file named `bolt.yaml` in one of the locations listed below. If a configuration file is not located at the expected path, or the file is empty, Bolt does not load the file. 
+
+You can place `bolt.yaml` configuration files at the following paths, from highest precedence to lowest:
+
+- **Project**
+  
+  The `bolt.yaml` file in a project directory applies settings to that project specifically. Matching settings at the project level override those at the user and system-wide levels.
+
+  `boltdir/bolt.yaml` or `<MY_PROJECT_NAME>/bolt.yaml`
```
nicklewis:
```suggestion
  `Boltdir/bolt.yaml` or `<MY_PROJECT_NAME>/bolt.yaml`
```

documentation/configuring_bolt.md
```diff
@@ -1,22 +1,156 @@
...
+
+  > **Note:** The project configuration file is loaded from the [Bolt project directory](bolt_project_directories.md). The default project directory is `~/.puppetlabs/bolt/`.
+
+- **User**
+  
+  The `bolt.yaml` file at the user level applies settings only to that user. Matching settings at the user level are overridden by project-level settings, but take precedent over system-wide settings.
```
nicklewis:
```suggestion
  The `bolt.yaml` file at the user level applies settings only to that user. Matching settings at the user level are overridden by project-level settings, but take precedence over system-wide settings.  
```

hestonhoffman: Good catch ;)

Issue #1564, opened 2020-01-29
Add environment variable inventory plugin


beechtom:
## Use Case
It's useful to load data from environment variables into an inventory or config file. This allows users to keep secrets and other information secure, particularly in a CI environment where the `prompt` plugin can't be used and the `vault` and `pkcs7` plugins might not be preferable due to requiring access to unencrypted data or private key files.

## Describe the Solution You Would Like
A simple `resolve_reference` task that reads an environment variable and returns the value.

```
_plugin: env_plugin
var: MY_VARIABLE
```

## Additional Context
I've started the plugin at https://github.com/puppetlabs/puppetlabs-env_plugin - we can move this to be a builtin Bolt plugin.

- Should the plugin be named `env_plugin` or something more generic? `env_var`
- Currently, the task just returns the raw value in the env variable. Should it attempt to convert to JSON when possible, or coerce in general? The output should just be treated as a string

Issue #1568, opened 2020-01-29
Prompt plugin should require a message


beechtom:
## Describe the Bug
The `prompt` plugin can be used without a `message` option.

## Expected Behavior
The `prompt` plugin should raise an error when a `message` option is not provided.

PR #1569, opened 2020-01-29
(GH-1564) Add env_var builtin plugin

beechtom:
This adds an `env_var` builtin plugin that looks up data from an
environment variable.

### DIFF
--- a/lib/bolt/plugin/env_var.rb
+++ b/lib/bolt/plugin/env_var.rb
@@ -1,0 +1,30 @@
+# frozen_string_literal: true
+
+module Bolt
+  class Plugin
+    class EnvVar
+      def initialize(*_args); end
+
+      def name
+        'env_var'
+      end
+
+      def hooks
+        %i[resolve_reference validate_resolve_reference]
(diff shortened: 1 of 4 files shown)


lib/bolt/plugin/env_var.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+
+require 'concurrent/delay'
```
nicklewis: This is unused

beechtom: The perils of copy/paste. Updated.

nicklewis: The real joy is it's also unused in the place it was copy/pasted from :)

beechtom: I'll go ahead and remove it in #1568 then 👍

nicklewis:
If we _did_ add an option to parse the value as JSON/YAML, I think this plugin could entirely supplant the `$BOLT_INVENTORY` feature.

PR #1572, opened 2020-01-30
(PE-28244) Ignore bolt.yaml in environment root when no modulepath is specified

donoghuc:
This commit ports over the work done in orchestrator to support reading modulepath from bolt.yaml present in an environment root (https://github.com/puppetlabs/orchestrator/pull/1266/files). Without it a bolt.yaml without `modulepath` set would cause pe-bolt-server to error when hitting the /plans/* endpoints.

### DIFF
--- a/lib/bolt_server/pe/pal.rb
+++ b/lib/bolt_server/pe/pal.rb
@@ -51,7 +51,10 @@ module BoltServer
         basemodulepath = plan_executor_config['basemodulepath'] || "#{codedir}/modules:/opt/puppetlabs/puppet/modules"
 
         with_pe_pal_init_settings(codedir, environmentpath, basemodulepath) do
+          modulepath_dirs = []
+          modulepath_setting_from_bolt = nil
           environment = Puppet.lookup(:environments).get!(environment_name)
+          path_to_env = environment.configuration.path_to_env
 
           # In the instance where the environment is "production" but no production dir
           # exists, the lookup will succeed, but the configuration will be mostly empty.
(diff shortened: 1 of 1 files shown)


donoghuc:
Adding do not merge until I get this in a build w/code manager and a bolt.yaml in an env root.

lib/bolt_server/pe/pal.rb
```diff
@@ -64,17 +67,35 @@ def initialize(plan_executor_config, environment_name, hiera_config = nil, max_c
           # environment: _technically_ that environment always exists, but if the dir
           # isn't there it won't find the module and fail with "plan not found" rather
           # than "environment doesn't exist"
-          if environment.configuration.path_to_env
-            bolt_config = File.join(environment.configuration.path_to_env, 'bolt.yaml')
+          if path_to_env
+            # This array is ordered: the first entry found that has a modulepath setting
+            # is what will be chosen
+            ['bolt.yaml'].each do |yaml_file|
```
nicklewis: This array contains only one element.

branan: I think this is an artifact of us trying to add `bolt.yml` at one point. I'd say we shouldn't copy it into this repo, and should probably clean it up in Orchestrator.

McdonaldSeanp: yeah, this was an artifact of when we were going to support both `bolt.yaml` and `bolt.yml`

We ended up taking that out but leaving in the rest of the processing code for when we put it back in.

lib/bolt_server/pe/pal.rb
```diff
@@ -64,17 +67,28 @@ def initialize(plan_executor_config, environment_name, hiera_config = nil, max_c
...
+          # we will use that modulepath rather than the one loaded through puppet. modulepath will
+          # be the _only_ setting that will work from bolt.yaml in plans in PE.
+          if modulepath_setting_from_bolt
+            modulepath_setting_from_bolt.split(File::PATH_SEPARATOR).each do |path|
+              if Pathname.new(path).absolute? && File.exist?(path)
+                modulepath_dirs << path
+              elsif File.exist?(File.join(path_to_env, path))
+                modulepath_dirs << File.join(path_to_env, path)
+              end
+            end
```
robdaemon: What if this ends up being an empty array? Say, none of the module paths in the bolt.yaml exist.

robdaemon:
Code-wise this looks 👍 to me, but it's the first thing in Bolt I've ever reviewed

lib/bolt_server/pe/pal.rb
donoghuc: `modulepath_settings_from_bolt` will be `nil` and `modulepath_dirs` will be `environment.modulepath`

donoghuc: I think that just the modules on the basemodulepath will be available.        https://github.com/puppetlabs/bolt/pull/1572/commits/93f7fc6472192ce7e1ceae9ad416a32867a362a4#diff-bef67b4dd1c8e5fa42a6a70e7cd5d182R88

robdaemon: Yeah, that's what it looks like - is that desirable? Maybe we should log some warnings or something for the missing paths? This seems like it would be difficult to debug if something goes wrong without some sort of logging.

donoghuc: Using bolt-server to list plans is kind of strange given that it does not run plans. I removed all logging in BoltServer::PE::PAL bc i thought logging plan stuff to bolt-server logs would be confusing.

PR #1573, opened 2020-01-31
(maint) Refactor/rename Bolt::Util#read_config_file

donoghuc:
All config (inventory.yaml, bolt.yaml, etc...) should have a hash type (or be empty). It is possible to have valid YAML that is not a hash type. This commit re-names the `Bolt::Util#read_config_file` method to `read_yaml_hash`. This change also removes some of the complexity in the method by no longer accepting a list of "default" paths to check. Previously if a list of defaults was passed in actually finding a file to read became optional. Now a second util method `read_optional_yaml_hash` has been added which will return the results of `read_yaml_hash` if the file exists and an empty hash otherwise. This allows the caller to choose the appropriate method based on if the file is expected to exist or not.

### DIFF
--- a/lib/bolt/util.rb
+++ b/lib/bolt/util.rb
@@ -3,38 +3,30 @@
 module Bolt
   module Util
     class << self
-      def read_config_file(path, default_paths = nil, file_name = 'file')
+      def read_yaml_hash(path, file_name)
         require 'yaml'
 
         logger = Logging.logger[self]
-        path_passed = path
-        if path.nil? && default_paths
-          found_default = default_paths.select { |p| File.exist?(p) }
-          if found_default.size > 1
-            logger.warn "Found #{file_name}s at #{found_default.join(', ')}, using the first"
-          end
-          # Use first found, fall back to first default and try to load even if it didn't exist
-          path = found_default.first || default_paths.first
-        end
-
         path = File.expand_path(path)
-        content = File.open(path, "r:UTF-8") { |f| YAML.safe_load(f.read) }
+        content = File.open(path, "r:UTF-8") { |f| YAML.safe_load(f.read) } || {}
+        unless content.is_a?(Hash)
+          msg = "Invalid content for #{file_name} file: #{path} should be a Hash or empty, not #{content.class}"
+          raise Bolt::FileError.new(msg, path)
+        end
         logger.debug("Loaded #{file_name} from #{path}")
         content
       rescue Errno::ENOENT
-        msg = "Could not read #{file_name} file: #{path}"
-        if path_passed
-          raise Bolt::FileError.new(msg, path)
-        else
-          logger.debug(msg)
-          nil
-        end
+        raise Bolt::FileError.new("Could not read #{file_name} file: #{path}", path)
       rescue Psych::Exception => e
         raise Bolt::FileError.new("Could not parse #{file_name} file: #{path}\n"\
                                   "Error at line #{e.line} column #{e.column}", path)
       rescue IOError, SystemCallError => e
         raise Bolt::FileError.new("Could not read #{file_name} file: #{path}\n"\
                                   "error: #{e}", path)
+      end
+
+      def read_optional_yaml_hash(path, file_name)
+        File.exist?(path) ? read_yaml_hash(path, file_name) : {}
       end
 
       # Accepts a path with either 'plans' or 'tasks' in it and determines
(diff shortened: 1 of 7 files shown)


donoghuc:
```yaml
# bolt.yaml
- foo
```
without proposed change
```
cas@cas-ThinkPad-T460p:~/working_dir/bolt$ bolt command run whoami -t localhost
bundler: failed to load command: bolt (/home/cas/working_dir/bolt/vendor/bundle/ruby/2.4.0/bin/bolt)
TypeError: no implicit conversion of String into Integer
  /home/cas/working_dir/bolt/lib/bolt/config.rb:211:in `[]'
  /home/cas/working_dir/bolt/lib/bolt/config.rb:211:in `update_from_file'
  /home/cas/working_dir/bolt/lib/bolt/config.rb:157:in `initialize'
  /home/cas/working_dir/bolt/lib/bolt/config.rb:122:in `new'
  /home/cas/working_dir/bolt/lib/bolt/config.rb:122:in `from_boltdir'
  /home/cas/working_dir/bolt/lib/bolt/cli.rb:122:in `parse'
  /home/cas/working_dir/bolt/exe/bolt:9:in `<top (required)>'
  /home/cas/working_dir/bolt/vendor/bundle/ruby/2.4.0/bin/bolt:22:in `load'
  /home/cas/working_dir/bolt/vendor/bundle/ruby/2.4.0/bin/bolt:22:in `<top (required)>'
```
with proposed change
```
cas@cas-ThinkPad-T460p:~/working_dir/bolt$ bolt command run whoami -t localhost
Invalid type: Array. /home/cas/working_dir/bolt/Boltdir/bolt.yaml must contain a Hash or be empty.
```

donoghuc:
Is this something we could do? Or should the caller be responsible for ensuring the config is a hash?

nicklewis:
This seems like a fine change, given that a) this method doesn't _currently_ let you know whether no config file existed or one did but contained null, b) every caller of this method adds an `|| {}` somewhere.

nicklewis:
Hmm actually #1557 will rely on the current behavior...

donoghuc:
The one failing cell looks like https://tickets.puppetlabs.com/browse/BOLT-1209

lib/bolt/util.rb
```diff
@@ -3,27 +3,23 @@
 module Bolt
   module Util
     class << self
-      def read_config_file(path, default_paths = nil, file_name = 'file')
+      # If file exists it is parsed and validated to be a valid yaml hash or empty. In the case it
+      # is empty an empty hash is returned. If the file does not exist: when'required' is true, a
+      # validation error is raised, when `required` is false the function returns `nil`.
+      def read_yaml_hash(path, required, file_name)
```
nicklewis: It would be clearer if this were two methods so that the expected behavior will be clear at the call site without having to know what the second boolean argument means. That could be as simple as:

```rb
def read_optional_yaml_hash(path, file_name)
  FIle.exist?(path) ? read_yaml_hash(path, file_name) : {}
end
```

donoghuc: Ah yeah that is much cleaner. Updated.

lib/bolt/util.rb
```diff
@@ -3,32 +3,19 @@
...
-          end
-          # Use first found, fall back to first default and try to load even if it didn't exist
-          path = found_default.first || default_paths.first
-        end
-
         path = File.expand_path(path)
-        content = File.open(path, "r:UTF-8") { |f| YAML.safe_load(f.read) }
+        content = File.open(path, "r:UTF-8") { |f| YAML.safe_load(f.read) } || {}
+        unless content.is_a?(Hash)
+          raise Bolt::ValidationError, "Invalid type: #{content.class}. #{path} must contain a Hash or be empty."
```
nicklewis: This message doesn't match the format of our other errors well. I also think it should be a 
`Bolt::FileError`. Maybe something like:

```rb
raise Bolt::FileError.new("Invalid content for #{file_name} file: #{path} should be a Hash or empty, not #{content.class}", path)
```

donoghuc: Oh, dang i just realized i had not even tested that! updated on your suggestion and added a test.

donoghuc: https://github.com/puppetlabs/bolt/pull/1573/commits/5a6ed8e93a076cddf4dedbc86b2679fe736c6b07#diff-82fc2c6cdf3061c5c7e36886addcb9b5R95-R106

PR #1575, opened 2020-02-03
Two minor fixes to the inventory-2 file documentation

thallgren:
1. The `facts` referred to as "top-level" resided in `group2` are moved
 to top-level.

2. `aliases` cannot be declared in group level, so it's probably
 `features` that are intended.

### DIFF
--- a/documentation/inventory_file_v2.md
+++ b/documentation/inventory_file_v2.md
@@ -100,7 +100,7 @@ targets:
 
 When searching for a target's configuration data, Bolt matches a target's URI with its name. Bolt uses depth-first search and uses the first value it finds.
 
-The `config` values for a target object, such as `host`, `transport`, and `port`, take precedent and override any `config` values at the group level. Bolt merges non-`config` data in the target object, such as `alias`, `facts`, and `vars`, with data in the group object.
+The `config` values for a target object, such as `host`, `transport`, and `port`, take precedent and override any `config` values at the group level. Bolt merges non-`config` data in the target object, such as `features`, `facts`, and `vars`, with data in the group object.
 
 ```yaml
 version: 2
@@ -129,8 +129,8 @@ groups:
     config: 
       ssh: 
         password: password
-    facts: 
-      operatingsystem: CentOS
+facts: 
+  operatingsystem: CentOS
 
 ```
 
--- a/.gitignore
+++ b/.gitignore
@@ -23,6 +23,7 @@
 /.bundle/
 /vendor/bundle
 /lib/bundler/man/
+/vendored/
 
 Gemfile.lock
 Gemfile.local


documentation/inventory_file_v2.md
```diff
@@ -129,8 +131,6 @@ groups:
     config: 
       ssh: 
         password: password
-    facts:
```
lucywyman: I think the may actually just need to be unindented, so it's at the same level as `groups` and `version`, based on the "top-level" in the next paragraph. Maybe "There is a fact on the 'all' group" would be better phrasing? @hestonhoffman  would know

adreyer:
What group you "reference" should not have any impact on what facts are used. All data from all groups is collected separate from turning a group name into a list of targets. This is the behavior I replicate if I return the facts from a plan.

Since the operatingsystem fact is only set in one place it should always be picked up for mytarget.

thallgren:
I'm don't get it. The sample query is `bolt inventory show -t group1 --detail`. Will that query show facts declared in `group2`?

adreyer:
It should there appears to be a bug.

The "targetting" part of that is intended only to decide which targets should be included in the result not where they get facts from. This is the same behavior as the targetting options for the actual bolt actions.

documentation/inventory_file_v2.md
hestonhoffman: Yeah agreed - that `fact` should be at the top level. Let's replace that sentence with "There is a second fact, `operatingsystem: CentOS`, in the `all` group."

adreyer: That seems like it would no longer exemplify that "Inventory files are not context-aware." in the following paragraph.

thallgren:
I still don't get it. Why have a hierarchy if the things you query in a specific branch of that hierarchy sees everything declared in all branches? It just doesn't make any sense to me.

nicklewis:
The inventory file is meant to express to Bolt the things that are _true_ of a target (what the SSH port is, which password to use, etc). Bolt considers those things to _always_ be true regardless of how you ask for the target. If you sometimes want to connect to a node over SSH and sometimes want to connect over WinRM, then to Bolt those are two different targets and you would sometimes choose one and sometimes choose the other.

The hierarchy exists so you can model the way that you think about your infrastructure so that you can have _non-overlapping_ properties declared in separate parts of the hierarchy. For instance, I might have a `linux_nodes` group that lists target `foo.example.com` (among others) and sets the transport to SSH. But I might have that same target also in a separate `webapp1` group that sets some facts. No matter whether I target `linux_nodes` or `webapp1`, it will still be the case that I connect to `foo.example.com` over SSH and the facts I've declared are still true.

thallgren:
Aha, so I guess that when overlapping facts are used, the inventory will give priority to the chosen group? If the answer to that is yes, then using a non-overlapping strategy will fail if a decision is ever made based on the presence/absence of a property (a property that is present in one group and absent in another will be perceived as always present). Is it safe to assume that this never happens with features, vars, or facts?

nicklewis:
The behavior is that the bottom left-most node in the tree wins. So child groups override their parents and groups that are specified earlier in the file override groups at the same hierarchy level specified later in the file. The final values will be the same no matter which group is being targeted. The first step is that the entire inventory is loaded and resolved, and _then_ the targets you asked for are pulled out. So the targeting process has no effect on the contents of the inventory.

In normal use, conflicts between groups at the same level of the hierarchy shouldn't happen much. If you're trying to tell Bolt something _meaningful_, then there aren't many real world cases where you would _meaningfully_ have conflicting values.

thallgren:
I'm trying to understand the exact semantics so that I can use this properly. Their degree of meaningfulness is not my concern right now. This is my current understanding:

1. A conflict will occur if the same property is declared in more than one group irrespective of on what level either of the groups are in the hierarchy.
2. If a property is defined in some group in the file, then there is no way for another group to say that this property is absent.

Is that correct?

thallgren:
@adreyer if targets in `group1` sees all facts declared in `group2`, them presumably targets in `group2` sees all facts in `group1`, then why doesn't both targets have the `operatingsystem: CentOS` fact in your query result?

nicklewis:
Yes those statements are both correct.

adreyer:
@thallgren `mytarget` is in both groups so it gets facts from both groups. `myothertarget` is only in `group1` so it does not pick up facts from group 2.

thallgren:
In that case, neither of the two statements that I made above are correct. There will be no override if two adjacent groups declare the same property with different values (unless they also declare the same target), and one group may have properties absent that another group has declared as present .

Another confusion is that in the example, there are two target declarations with the same name. Shouldn't one of the groups just a target reference (string target in name for) to the other target? Or would that change the semantics so that only properties from the group where the target is declared would be picked up by the reference? What happens if the two targets have different uri's?

thallgren:
To answer my own questions here:

If the uri of `mytarget` in `group2` has a different uri than a target with the same name in `group1`, then the uri from `group2` silently overrides the declaration in `group1`.
If `group2` doesn't declare `mytarget` but instead just references it, the target declaration still picks up facts from `group2`.

The reason I ask a lot of questions and want clarifications is that we're considering using the inventory as the format for a federated inventory service spanning multiple PE-installations. There are a couple of things in the current design that prevents this:

1. A group / target name must be globally unique within the file (and all other groups and targets brought in by plugins).
   We need names to be unique within the group where they are declared only (more like a file system). This will require that a group can be uniquely identified using a qualified key.
2. The same target can be declared in many places.
   We need to reference the same target from multiple groups, but it must be declared just once. Furthermore, the declaration would only inherit properties where it is declared (from group and parent groups). Never from groups where it is referenced.
3. A target declares its own aliases.
   We would need an option to declare aliases in groups where a target is referenced (to avoid name conflicts in that group). E.g. "in this particular group I want target x to be named y". Not very different from symbolic links.

adreyer:
I think this is a valuable conversation because I believe there is a disconnect in how we think about inventory after reading the docs and how you're thinking about it. I'm not sure a PR is the best place for that conversation and now that I know why you're looking at this I suspect what you're talking about doesn't align well with the general conversations we've been having about federation. I think we should set up some time to talk about this in person.

I will say that people are successfully using the inventory file format to federate across pe instances. This works naturally if you can rely on certnames being unique. If not you need to enforce a naming scheme that that separates the certname in a pe instance from the ID(name in inventory) used to refer to the target at the federated later.

There is however a neede fix in this PR regarding aliases at the group level. do you want to remove the rest of this PR or should we just put up a new one?

thallgren:
PR updated. I moved the facts to top-level since the documentation states that "CentOS" is a "top-level fact".

adreyer:
I think "top level" is the typo since the next paragraph uses that example to show that the facts associated with the node are not based on the group context from the target.

adreyer:
superceded by #1606

PR #1576, opened 2020-02-03
(maint) Remove 'deploy' workflow

beechtom:
This removes the `deploy` workflow, as Puppet Container Registry (PCR)
has been taken offline. Since the workflow pushes a Docker container to
PCR, it will now fail.

This workflow should be modified and added back once a replacement registry is determined.


beechtom:
Deleted!

donoghuc:
I'm hoping we can just change the secrets to https://github.com/puppetlabs/pe-puppetserver/blob/2a7a30970fc50dfe5638bbd864dd160e18010cad/docker/distelli-manifest.yml#L5 and leave everyting else the same. Trying to track those down. Might as well just leave this up for now.

donoghuc:
I *think* https://github.com/puppetlabs/bolt/pull/1578 should cover this.

donoghuc:
Put up https://github.com/puppetlabs/bolt/pull/1583 bc this has merge conflict

PR #1578, opened 2020-02-03
Push bolt server to artifactory

donoghuc:


donoghuc:
This is the marketplace action i want to use for docker login https://github.com/Azure/docker-login

PR #1579, opened 2020-02-04
(maint) Support arbitrary URI schemes for remote targets in v2 inventory

nicklewis:
In both the v1 and v2 inventory formats, we use the URI scheme as the
transport for the target. In v1 however, that was only true if the
target did not also separately have transport specified as "remote". In
that case, the URI scheme was used as simply a URI scheme (called
`protocol` when running a task) allowing it to therefore be arbitrary.
With the v2 inventory, we didn't respect that distinction and would a) fail
if the URI scheme did not correspond to a valid transport, and b)
override the "remote" transport specification with the one derived from
the scheme, preventing the use of remote targets with specific URI schemes.

We now vary the behavior of protocol and transport for v2 targets based
on whether or not they are remote. For remote targets, `transport` will
always be "remote" and `protocol` will be the URI scheme if specified.
For ordinary targets, the URI scheme will be used to set the `transport`
if specified and there will be no `protocol`.

On the Target datatype that is exposed to Puppet, we continue to
implement the `protocol` method for backward compatibility, returning
the value of `transport` for non-remote targets (which have no
protocol).

### DIFF
--- a/lib/bolt/inventory/target.rb
+++ b/lib/bolt/inventory/target.rb
@@ -143,12 +143,26 @@ module Bolt
         @uri_obj.port || transport_config['port']
       end
 
+      # For remote targets, protocol is the value of the URI scheme. For
+      # non-remote targets, there is no protocol.
       def protocol
-        transport
+        if remote?
+          @uri_obj.scheme
+        end
       end
 
+      # For remote targets, the transport is always 'remote'. Otherwise, it
+      # will be either the URI scheme or set explicitly.
       def transport
-        @uri_obj.scheme || transport_config_cache['transport']
+        if remote?
+          'remote'
+        else
+          @uri_obj.scheme || transport_config_cache['transport']
+        end
+      end
+
+      def remote?
+        @uri_obj.scheme == 'remote' || transport_config_cache['transport'] == 'remote'
       end
 
       def user
--- a/lib/bolt/target.rb
+++ b/lib/bolt/target.rb
@@ -118,11 +118,11 @@ module Bolt
     end
 
     def transport
-      inventory_target.protocol
+      inventory_target.transport
     end
 
     def protocol
-      inventory_target.protocol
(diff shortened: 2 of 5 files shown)


lib/bolt/inventory/target.rb
```diff
@@ -143,12 +143,28 @@ def port
         @uri_obj.port || transport_config['port']
       end
 
+      # For remote targets, protocol is the value of the URI scheme. For
+      # non-remote targets, there is no protocol.
       def protocol
-        transport
+        if remote?
+          @uri_obj.scheme
```
nicklewis: This would be `@uri_obj.scheme || transport_config_cache['protocol']` if we were to expose a `protocol` config option. Then it would mirror the `transport` method and reinforce the idea that the URI scheme gets assigned to _either_ the protocol _or_ the transport depending on whether the target is remote.

beechtom:
I think this makes sense. Just needs a quick fix to make rubocop happy.

nicklewis:
I tried to standardize a bit on terminology, though it's fairly arbitrary.

• **transport**: The Bolt transport that will be used. `remote` for remote targets
• **scheme**: The first portion of the URI, when parsing a URI
• **protocol**: An arbitrary field derived from the scheme for remote targets and passed to the proxy target when running tasks. Unset for non-remote targets.

nicklewis:
I originally had this up as just a PoC so didn't have a commit message, whoops. I just pushed up a fleshed out version.

nicklewis:
Hmmmm it looks like we need to modify the `transport_config_cache` validation to take this new behavior into account.

nicklewis:
Ah no, that validation is fine. I was testing this alongside #1580 and had to do a bit of work to make the test setup the target properly. But as long as it's created the way it's supposed to be, this works great. I will hold off on adding any tests in _this_ PR for it as it will be implicitly tested by #1580.

Issue #1567, opened 2020-01-29
Remove deprecations for Bolt 2.0


lucywyman:
## TODO
- [x] Create a 2.0 branch
- [x] Remove inventory v1 and target v1 objects. This includes updating spec tests and removing documentation. The `Bolt::Inventory` class should remain as an inventory factory. #1580 
- [x] Remove `future` logic. Update spec tests, and ensure documentation is correct with new behavior. Bolt should warn if the future config is set, and do nothing. #1590 
- [x] Remove deprecated features (anything that currently has a deprecation warning). #1590 
- [x] Update acceptance tests to pass with new Bolt. This includes reviewing the acceptance tests to determine what we need to keep. Anything that isn't platform or packaging dependent should be tested at the integration level. #1595 
- [x] Bump minimum ruby version to 2.5 #1546
- [x] Add warning message that v1 inventory file needs to be upgraded to v2
- [x] Remove plugin config (bolt_plugin.json) #1598

PR #1580, opened 2020-02-04
(GH-1567) Remove inventory, group, and target v1

beechtom:
This commit removes the inventory, group, and target v1 classes and
renames the v2 classes. The classes have been renamed as follows:

- `Bolt::Inventory::Inventory2` => `Bolt::Inventory::Inventory`
- `Bolt::Inventory::Group2` => `Bolt::Inventory::Group`
- `Bolt::Target2` => `Bolt::Target`

Additional changes include the following:

- The `Bolt::Inventory` class now acts as an inventory factory, with
several methods for generating v2 inventory objects.
- The `Bolt::CLI#migrate_project` method now parses an inventory file
for `nodes` keys instead of checking the `version` field.
- Added a `Bolt::Target#from_hash` method to create a target from a
config hash, add it to the inventory, and return a `Bolt::Target`
object.


lib/bolt/task.rb
```diff
@@ -82,7 +82,7 @@ def implementations
     # and any additional files (name and path)
     def select_implementation(target, provided_features = [])
       impl = if (impls = implementations)
-               available_features = target.feature_set + provided_features
+               available_features = target.features.to_set + provided_features
```
nicklewis: Why was this change necessary?

beechtom: It wasn't. This was early on in my spec-fixing adventure, before I realized stubs needed to be updated. This is the part that I needed to fix: https://github.com/puppetlabs/bolt/blob/master/spec/bolt/task_spec.rb#L22

It's in the update.

spec/integration/transport/winrm_spec.rb
```diff
@@ -47,7 +50,14 @@ def mk_config(conf)
...
-    Bolt::Target.new("#{host_}:#{port_}").update_conf(conf.transport_conf)
+    t = inventory.get_target("#{host_}:#{port_}")
+    update_target(t, conf.transports[conf.transport.to_sym])
+    t
+  end
+
+  def update_target(targ, conf)
+    transport_config = targ.options.merge(conf)
+    targ.inventory_target.set_config(targ.transport, transport_config)
   end
```
beechtom: I'd like to look at using `Bolt::Target#from_hash` here instead... creating a config object to create a target, and using `set_config` to modify settings seems unnecessary, since all we care about is having a single target with the appropriate config set.

lib/bolt/cli.rb
```diff
@@ -611,6 +604,18 @@ def migrate_project
...
+    # This will be needed to properly migrate inventory files since v2 files
+    # no longer require a 'version' key
+    def collect_vals(data, key)
+      case data
+      when Array
+        data.flat_map { |d| collect_vals(d, key) }
+      when Hash
+        [data['nodes']] + collect_vals(data.values, key).compact
+      end
+    end
```
beechtom: Since v1 files more than likely won't have a `version` field set, and since v2 files are becoming the default, `migrate_project` can't rely on checking the `version` field for migrating. The best solution I could come up with was checking if any `nodes` keys are present and then migrating if there are.

I'm positive there's a better way to check a nested hash for a key than this, but this is the best I could come up with.

lib/bolt_server/transport_app.rb
```diff
@@ -291,7 +300,16 @@ def make_winrm_target(target_hash)
       }
 
       opts = target_hash.clone.merge(overrides)
-      Bolt::Target.new(target_hash['hostname'], opts)
+
+      data = {
+        'uri' => target_hash['hostname'],
+        'config' => {
+          'winrm' => opts
```
nicklewis: Should we have `'transport' => 'winrm'` here like we have for ssh above?

spec/fixtures/modules/results/lib/puppet/functions/results/make_apply_result.rb
```diff
@@ -8,7 +8,9 @@
   end
 
   def create(uri, value)
-    target = Bolt::Target.new(uri)
+    inventory = Puppet.lookup(:bolt_inventory)
+    inventory.get_target(uri)
+    target = Bolt::Target.new(uri, inventory)
```
nicklewis: `inventory.get_target(uri)` should already be returning a `Bolt::Target` instance

spec/integration/transport/docker_spec.rb
```diff
@@ -48,23 +53,24 @@
...
         docker.with_connection(target) {}
       }.to raise_error(Bolt::Node::ConnectError, /Could not find a container with name or ID matching/)
     end
   end
 
   context 'when there is no host in the target' do
-    let(:target) { Bolt::Target.new(nil, "name" => "hostless") }
+    # Directly create an inventory target, since Inventory#get_target doesn't allow
+    # for passing config and would set the host as the name passed to it
+    let(:target) { Bolt::Inventory::Target.new({ 'name' => 'hostless' }, inventory) }
```
nicklewis: Should this use `Bolt::Target.from_hash` now?

spec/integration/transport/ssh_spec.rb
```diff
@@ -38,15 +38,19 @@ def mk_config(conf)
...
   let(:task_input_size) { 100000 }
   let(:big_task_input) { "f" * task_input_size }
   let(:stdin_task) { "#!/bin/sh\ngrep data" }
   let(:env_task) { "#!/bin/sh\necho $PT_data" }
+  let(:inventory) { Bolt::Inventory.empty }
+  let(:transport_conf) { {} }
 
   def make_target(host_: hostname, port_: port, conf: config)
-    Bolt::Target.new("#{host_}:#{port_}", transport_conf).update_conf(conf.transport_conf)
+    t = inventory.get_target("#{host_}:#{port_}")
```
nicklewis: Ditto about `Bolt::Target.from_hash`

spec/lib/bolt_spec/conn.rb
```diff
@@ -48,13 +49,16 @@ def conn_uri(transport, include_password: false, override_port: nil)
     end
 
     def conn_target(transport, include_password: false, options: nil)
-      Bolt::Target.new(conn_uri(transport, include_password: include_password), options)
+      inventory = Bolt::Inventory.empty
+      target = inventory.get_target(conn_uri(transport, include_password: include_password))
+      inventory.set_config(target, transport, options) if options
```
nicklewis: Another spot for `Bolt::Target.from_hash`?

spec/pal/features_spec.rb
```diff
@@ -22,8 +25,9 @@
       'vars' => { 'pb' => 'jelly', 'mac' => 'cheese' }
     }
   }
-  let(:inventory) { Bolt::Inventory.new(data) }
   let(:pal) { Bolt::PAL.new(modulepath, nil, nil) }
+  let(:plugins) { Bolt::Plugin.setup(config, nil, nil, analytics) }
+  let(:inventory) { Bolt::Inventory::Inventory.new(data, plugins: plugins) }
```
nicklewis: Can this just be `Bolt::Inventory.empty`?

spec/shared_examples/transport.rb
```diff
@@ -62,8 +62,15 @@ def mk_config(conf)
   conf_object
 end
 
-def make_target(target_: host_and_port, conf: config)
-  Bolt::Target.new(target_, transport_conf).update_conf(conf.transport_conf)
+def make_target
+  target = inventory.get_target(host_and_port)
```
nicklewis: `Bolt::Target.from_hash`?

nicklewis:
I'm not too attached to the comments I left in the `spec` directory. There are potentially 
opportunities to simplify the tests with some of the new helpers but we could probably do that until the end of time.

lib/bolt_server/transport_app.rb
beechtom: For some reason it breaks the tests. Not really sure why...

lib/bolt/cli.rb
```diff
@@ -584,21 +584,14 @@ def initialize_project
...
-        end
-
-        inv = YAML.safe_load(File.open(inventory_file))
-        migrate_group(inv)
+      inventory_file = config.inventoryfile || config.boltdir.inventory_file
+      data = Bolt::Util.read_yaml_hash(inventory_file, 'inventory')
 
-        ok = File.write(inventory_file, { 'version' => 2 }.merge(inv).to_yaml)
+      if collect_vals(data, 'nodes').compact.any?
+        migrate_group(data)
```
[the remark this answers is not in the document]
beechtom: 🤦‍♂

lib/bolt_server/transport_app.rb
```diff
@@ -286,12 +295,23 @@ def make_ssh_target(target_hash)
...
 
     def make_winrm_target(target_hash)
-      overrides = {
-        'protocol' => 'winrm'
+      defaults = {
+        'ssl' => false,
+        'ssl-verify' => false
+      }
+
+      opts = defaults.merge(target_hash.clone)
```
donoghuc: Why do you have to clone the target_hash?

lib/bolt_server/transport_app.rb
beechtom: Fixed by setting defaults to `ssl` and `ssl-verify` to `false`.

lib/bolt_server/transport_app.rb
beechtom: I just copied what was in `make_ssh_target` already.

donoghuc: I think that one is cloned bc `overrides` is merged over it.

donoghuc: But i dont think it is needed there either.

beechtom: Removed `clone`.

lib/bolt/cli.rb
```diff
@@ -624,9 +618,9 @@ def migrate_group(group)
         group['targets'] = targets
       end
       (group['groups'] || []).each do |subgroup|
-        migrate_group(subgroup)
+        migrated = migrate_group(subgroup)
```
nicklewis: This should be `||=`. `migrated` is a latch that ought to only ever be switched in one direction. If _at any point_ we find a `nodes` key, then we should set `migrated` to true, but we should under no circumstances ever set it back to false.

Currently, this will report no changes in this case:

```yaml
nodes:
  - foo.example.com
  - bar.example.com
groups:
  - name: somegroup
```

Admittedly, there's not a lot of reason to have a leaf group without a `nodes` key, but it's technically allowed.

lib/bolt/cli.rb
```diff
@@ -584,22 +584,14 @@ def initialize_project
...
-        end
+      inventory_file = config.inventoryfile || config.default_inventoryfile
+      data = Bolt::Util.read_yaml_hash(inventory_file, 'inventory')
 
-        inv = YAML.safe_load(File.open(inventory_file))
-        migrate_group(inv)
-
-        ok = File.write(inventory_file, { 'version' => 2 }.merge(inv).to_yaml)
-      end
+      ok = if migrate_group(data)
```
nicklewis: We should report that there was nothing to do if `migrate_group` returns false rather than saying we migrated it when we didn't.

CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # Changelog
 
+## Bolt Next
```
adreyer: I think it's safe to call this Bolt 2.0.0

bolt-modules/boltlib/lib/puppet/datatypes/target.rb
```diff
@@ -3,54 +3,33 @@
 Puppet::DataTypes.create_type('Target') do
   begin
     inventory = Puppet.lookup(:bolt_inventory)
-
-    inventory_version = inventory.version
-    if inventory_version != 1
-      target_implementation_class = inventory.target_implementation_class
-    end
+    target_implementation_class = inventory.target_implementation_class
```
adreyer: I think we can just get rid of this logic and hardcode the implementation class.

adreyer: nm this is used to choose the special apply class.

beechtom: We need to keep it since 2.0 has `Bolt::ApplyInventory` and `Bolt::ApplyTarget`.

PR #1583, opened 2020-02-04
(maint) Do not ship bolt-server image with GH actions

donoghuc:
Now that we are shipping bolt-server to artifactory we cannot safely use GH actions (https://help.github.com/en/actions/automating-your-workflow-with-github-actions/about-self-hosted-runners#self-hosted-runner-security-with-public-repositories). Intead we will do this in Jenkins.


donoghuc:
We should delete `DOCKER_*` secrets from this repo.

PR #1586, opened 2020-02-05
(maint) Support Kerberos authentication with SSH

lucywyman:
The net-ssh-krb gem was previously licensed with an incompatible license
for use in Bolt, in addition to a dependency version conflict. The gem
was recently relicensed and updated, so can now be shipped out of the
box with Bolt and documented. This allows users to use Kerberos
authentication when connecting to targets over SSH by specifying
`gssapi-with-mic` as a `PreferredAuthenticationMethod` in their OpenSSH
config (typically `~/.ssh/config`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@
 
 ### New features
 
+* **Add Kerberos support for SSH transport**
+
+  Users can now authenticate with Kerberos when using the SSH transport.
+
 * **Warning when task metadata has unknown keys** ([#1542](https://github.com/puppetlabs/bolt/pull/1542))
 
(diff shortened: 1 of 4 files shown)


donoghuc:
Dont forget puppet-runtime :) I think we could use a broad gem update there it has been a while. Especially now that we appear to be having problems builidng native extensions for the older version of ffi we are shipping.

lucywyman:
Oh my gosh I totally forgot puppet-runtime. Thanks Cas!

lucywyman:
This is in puppet-runtime now - PR is ready for review

nicklewis:
@donoghuc That code shouldn't be necessary now, no

lucywyman:
Removed

PR #1587, opened 2020-02-05
(GH-1585) Don't show apply result hash in human output

### LINKED ISSUE #1585: Bolt outputs JSON report from plan, even when human format is specified
## Describe the Bug
Bolt (since 1.47) is printing the result of a plan in JSON format to the standard output in the console, even when human format is specified.

## Expected Behavior
I would expect the result report is not output when human format is specified.

## Steps to Reproduce
1. Consider the following plan:
    ```puppet
    plan test(
      TargetSpec $targets
    ) {
      return apply($targets) {
        notice('Hello World')
      }
    }
    ```
2. If you run this plan with `bolt plan run test --format=human --targets=localhost` then the standard human formatted text will be output (expected). But then after that, it will still output the JSON of the full ResultSet report (not expected).

## Environment
 - Version: 1.47+
 - Platform: Any

beechtom:
This fixes a regression introduced in #1544 that would show an apply
result hash as a JSON object when using human output. Now, if a plan
returns only an apply result, human output will display a metrics
message instead of the apply result hash.

Fixes #1585

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,14 @@
 # Changelog
 
+## Bolt Next
+
+### Bug fixes
+
+* **Remove apply result hash from human output** ([#1585](https://github.com/puppetlabs/bolt/issues/1585))
+
+  Apply result hashes will no longer be displayed when using human output. Instead, a metrics message
+  will be shown.
+
 ## Bolt 1.48.0
 
 ### New features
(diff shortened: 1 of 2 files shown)


nicklewis:
Would this be better handled by overriding the `generic_result` method on `ApplyResult` to return nothing?

beechtom:
That seems preferable. Updated. 👍

PR #1588, opened 2020-02-06
(maint) Don't include OpenSSH config section in table of contents

nicklewis:
This was at the same level as the actual transports, causing it to show
up as an entry in the table of contents under "Transport configuration
options".


nicklewis:
image

PR #1590, opened 2020-02-06
(GH-1567) Remove deprecated features and future logic

beechtom:
This removes features deprecated in Bolt 2.0 and removes all `future` logic, making any future behavior the default.

Rebased off #1580


spec/bolt/plugin/prompt_spec.rb
```diff
@@ -8,30 +8,12 @@
...
-    val = subject.resolve_reference(prompt_data)
-    expect(val).to eq(password)
-  end
-
-  it 'concurrent delay prompts for data on STDERR when executed with $future flag set' do
-    # rubocop:disable Style/GlobalVars
-    $future = true
-    # rubocop:enable Style/GlobalVars
-
+  it 'concurrent delay prompts for data on STDERR when executed' do
```
nicklewis: It doesn't concurrent delay anything anymore

lib/bolt/config.rb
```diff
@@ -260,7 +254,10 @@ def update_logs(logs)
     end
 
     def update_from_file(data)
-      @future = data['future'] == true
+      if data['future']
+        msg = "Configuration option 'future' no longer exposes future behavior."
```
nicklewis: THE FUTURE IS NOW

bolt-modules/boltlib/lib/puppet/functions/get_targets.rb
```diff
@@ -3,9 +3,9 @@
 require 'bolt/error'
 
 # Parses common ways of referring to targets and returns an array of Targets.
-# `get_targets('all')` returns an empty array.
 #
-# **NOTE:** Not available in apply block when `future` is true
+# **NOTE:** Calling `get_target('all')` returns an empty array.
```
nicklewis: This should be `get_targets` but I don't think it's actually true??

Also we should leave the notes about "Not available in apply blocks" and just remove the "when `future` is true" bit, for each of these functions.

documentation/bolt.ditamap
```diff
@@ -32,7 +32,6 @@
     <topicref href="bolt_configuration_reference.md" format="markdown"/>
     <topicref href="bolt_configure_orchestrator.md" format="markdown"/>
     <topicref href="bolt_connect_puppetdb.md" format="markdown"/>
-    <topicref href="inventory_file.md" format="markdown"/>
```
nicklewis: Should we delete this file altogether?

lib/bolt/catalog.rb
```diff
@@ -55,41 +55,33 @@ def generate_ast(code, filename = nil)
 
     def setup_inventory(inventory)
```
nicklewis: This method is now unused.

lib/bolt/applicator.rb
```diff
@@ -232,35 +176,32 @@ def count_statements(ast)
...
-      end
-      # rubocop:enable Style/GlobalVars
+      # Serialize as pcore for *Result* objects
+      plan_vars = Puppet::Pops::Serialization::ToDataConverter.convert(plan_vars,
+                                                                       rich_data: true,
+                                                                       symbol_as_string: true,
+                                                                       type_by_reference: true,
+                                                                       local_reference: false)
+
+      pdb_client_hash = @pdb_client.respond_to?(:config) ? @pdb_client.config.to_hash : {}
```
nicklewis: Under what circumstance does this not respond to `config`?

beechtom: There's a [test](https://github.com/puppetlabs/bolt/blob/master/spec/bolt_spec/plan_spec.rb#L198-L203) that uses a `BoltSpec::Plans::MockPuppetDBClient` client defined [here](https://github.com/puppetlabs/bolt/blob/master/lib/bolt_spec/plans.rb#L188). When that gets passed to the applicator it errors since `config` returns `nil`. I tried allowing instances of that class to accept `config` and return an empty hash, but the test still failed. I also tried defining a `config` method that returned an empty hash, but it caused a ton of other tests to fail.

Not really sure what the best approach is here... I made this change to at least get the test passing / review started.

documentation/bolt.ditamap
beechtom: Wasn't sure if we wanted to keep it for a potential 'archive' or not, but I guess we decided we're not supporting old docs? Removed.

lib/bolt/bolt_option_parser.rb
```diff
@@ -599,17 +599,11 @@ def initialize(options)
       @warnings = []
 
       separator "\nINVENTORY OPTIONS"
-      define('-n', '--nodes NODES',
```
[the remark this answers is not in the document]
beechtom: TIL optparse allows multiple short-form arguments. I dislike it a little less.

lib/bolt/applicator.rb
nicklewis: I think the way to go is to give `MockPuppetDBClient` an initialize method that takes a config, add an `attr_reader :config`, and pass `Bolt::PuppetDB::Config.new({})` when we construct it in `lib/bolt_spec/plans.rb`. Then it ought to behave the way the rest of the code expects.

Defining `config` to return an empty hash _should_ have worked (since `Hash#to_hash` is defined), so if this approach (adding a proper config attribute) causes problems with other tests, it's probably an underlying issue with the test.

bolt-modules/boltlib/lib/puppet/functions/add_facts.rb
```diff
@@ -8,7 +8,7 @@
 Puppet::Functions.create_function(:add_facts) do
   # @param target A target.
   # @param facts A hash of fact names to values that may include structured facts.
-  # @return The target's new facts or a `Target` object if the `future` flag is set to true
+  # @return A `Target` object.
```
nicklewis: The `return_type` needs to be updated.

nicklewis:
Looks like `add_facts` tests are failing because the inventory call is stubbed to return an incorrect result. We probably don't need to stub anything there in the first place.

nicklewis:
Quite a few of our function tests are checking the implementation of the function rather than its observable behaviors...

PR #1595, opened 2020-02-11
(GH-1567) Fix acceptance tests

beechtom:


acceptance/tests/apply_plan_ssh.rb
```diff
@@ -36,30 +36,6 @@
...
-      host = node.hostname
-      result = json.select { |n| n['node'] == host }
-      assert_equal('success', result[0]['status'],
-                   "The task did not succeed on #{host}")
-
-      # Verify that files were not created on the target
-      on(node, "cat #{filepath}/hello.txt", acceptable_exit_codes: [1])
-    end
-  end
-
```
beechtom: This test seemed redundant since we already test that a plan with `apply` works and we test that `noop` works at the integration level.

acceptance/tests/apply_plan_ssh.rb
```diff
@@ -94,32 +70,6 @@
     end
   end
 
-  step "puppet service should be stopped" do
```
nicklewis: Is this to ensure that `apply_prep` doesn't start the Puppet service? Why was this bit removed?

beechtom: That was a mistake. Added back in.

PR #1602, opened 2020-02-12
(GH-1567) Error when using a v1 inventory

beechtom:
This adds an error when Bolt detects that a v1 inventory file is being
used. A v1 inventory file is identified by the use of a `nodes` key at
the group level, which is deprecated in v2 inventory files.

### DIFF
--- a/lib/bolt/inventory/group.rb
+++ b/lib/bolt/inventory/group.rb
@@ -239,6 +239,15 @@ module Bolt
           raise ValidationError.new(msg, @name)
         end
 
+        if input.key?('nodes')
+          msg = <<~MSG.chomp
+                Found 'nodes' key in group #{@name}. This looks like a v1 inventory file, which is
+                no longer supported by Bolt. Migrate to a v2 inventory file automatically using
+                'bolt project migrate'.
+                MSG
+          raise ValidationError.new(msg, nil)
+        end
+
         unless (unexpected_keys = input.keys - GROUP_KEYS).empty?
           msg = "Found unexpected key(s) #{unexpected_keys.join(', ')} in group #{@name}"
           @logger.warn(msg)
--- a/spec/bolt/inventory/group_spec.rb
+++ b/spec/bolt/inventory/group_spec.rb
@@ -751,6 +751,12 @@ describe Bolt::Inventory::Group do
         Bolt::Inventory::Group.new({ 'name' => 'foo', 'unexpected' => 1 }, plugins)
       end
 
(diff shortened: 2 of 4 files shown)


lib/bolt/inventory/group.rb
```diff
@@ -239,7 +239,16 @@ def validate_group_input(input)
           raise ValidationError.new(msg, @name)
         end
 
-        unless (unexpected_keys = input.keys - GROUP_KEYS).empty?
+        if input.key?('nodes')
+          msg = <<~MSG
+                Found deprecated key nodes in group #{@name}. Nodes keys are only available in v1 inventory
```
nicklewis: Maybe something like:

```
Found 'nodes' key in group #{@name}. This looks like a v1 inventory file, which is no longer supported by Bolt. Migrate to a v2 inventory file automatically using 'bolt project migrate'.
```

lib/bolt/inventory/group.rb
```diff
@@ -239,7 +239,16 @@ def validate_group_input(input)
...
+                files, which are no longer supported by Bolt. Migrate to a v2 inventory file automatically
+                using 'bolt project migrate'.
+                MSG
+          @logger.warn(msg)
```
nicklewis: We should raise an exception here rather than just ignore the key.

lib/bolt/inventory/group.rb
```diff
@@ -239,6 +239,15 @@ def validate_group_input(input)
           raise ValidationError.new(msg, @name)
         end
 
+        if input.key?('nodes')
+          msg = <<~MSG
+                Found 'nodes' key in group #{@name}. This looks like a v1 inventory file, which is
+                no longer supported by Bolt. Migrate to a v2 inventory file automatically using
+                'bolt project migrate'.
+                MSG
+          raise ValidationError.new(msg, @name)
```
nicklewis: `ValidationError` leads to some weird behavior here because it appends "for group at [<path to group>]" to the message. So I got this:

```
Found 'nodes' key in group all. This looks like a v1 inventory file, which is
no longer supported by Bolt. Migrate to a v2 inventory file automatically using
'bolt project migrate'.
 for group at ["all"]
```

That can be avoided by passing `nil` instead of `@name`. That's reasonable because this is a problem with the _whole inventory file itself_, not just this group. Or at least that's how I'll rationalize it to myself.

Also, this message has a trailing newline which winds up being printed. We should add a `.chomp` after `<<~MSG`.

Issue #1596, opened 2020-02-12
it would be 'nice to have' ability to filter plans or tasks from the bolt show command


abuxton:
## Use Case
When you start to develop larger projects with more active module dependencies the number of tasks exposed by `bolt task show` becomes rather verbose. 

## Describe the Solution You Would Like
Provide a `--filter` option followed by a string and use substring matching. Tasks should only be filtered by name.

Bolt should raise a helpful error if the string contains invalid task name characters.

## Describe Alternatives You've Considered
currently parse through Grep, but that means documenting Grep example and Get-ent for Powershell users unified approach from the CLI would be prefered.

## Additional Context
same functionality for `bolt plan show` would be an obvious consideration

beechtom:
Closed by #1603

PR #1603, opened 2020-02-13
(GH-1596) Filter tasks and plans by substring

beechtom:
This adds support for a `--filter FILTER` option when using the `task show` and
`plan show` commands. When using the `--filter` option, the list of
tasks and plans displayed by the outputter will be filtered to tasks and
plans whose names include the provided substring.

### DIFF
--- a/lib/bolt/bolt_option_parser.rb
+++ b/lib/bolt/bolt_option_parser.rb
@@ -67,7 +67,7 @@ module Bolt
           { flags: ACTION_OPTS + %w[params compile-concurrency tmpdir],
             banner: PLAN_RUN_HELP }
         when 'show'
-          { flags: OPTIONS[:global] + OPTIONS[:global_config_setters],
+          { flags: OPTIONS[:global] + OPTIONS[:global_config_setters] + %w[filter format],
             banner: PLAN_SHOW_HELP }
         else
           { flags: OPTIONS[:global],
@@ -130,7 +130,7 @@ module Bolt
           { flags: ACTION_OPTS + %w[params tmpdir noop],
             banner: TASK_RUN_HELP }
         when 'show'
-          { flags: OPTIONS[:global] + OPTIONS[:global_config_setters],
+          { flags: OPTIONS[:global] + OPTIONS[:global_config_setters] + %w[filter format],
             banner: TASK_SHOW_HELP }
         else
           { flags: OPTIONS[:global],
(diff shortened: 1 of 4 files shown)


lib/bolt/cli.rb
```diff
@@ -453,15 +453,15 @@ def show_task(task_name)
     end
 
     def list_tasks
-      outputter.print_tasks(pal.list_tasks, pal.list_modulepath)
+      outputter.print_tasks(pal.list_tasks, pal.list_modulepath, options[:filter])
```
nicklewis: It seems better to apply the filter _here_ (and ditto in `list_plans`) so that each outputter doesn't have to support it separately (especially since I can't imagine the behavior differing between outputters).

lib/bolt/bolt_option_parser.rb
```diff
@@ -761,6 +761,9 @@ def initialize(options)
       end
 
       separator "\nDISPLAY OPTIONS"
+      define('--filter FILTER', 'Filter tasks and plans by a matching substring') do |filter|
```
nicklewis: We should validate that the filter only contains legal task/plan characters. Something like `[a-z0-9_:]`, either `filter =~ /[a-z0-9_:]+/` or `filter !~ /^[a-z0-9_:]/`.

lib/bolt/bolt_option_parser.rb
```diff
@@ -761,6 +761,14 @@ def initialize(options)
...
+        unless /[a-z0-9_:]+/.match(filter)
+          msg = "Illegal characters in filter string '#{filter}'. Filters must match the regular "\
```
nicklewis: Maybe `Filter must match a legal task or plan name`?

lib/bolt/bolt_option_parser.rb
```diff
@@ -761,6 +761,14 @@ def initialize(options)
       end
 
       separator "\nDISPLAY OPTIONS"
+      define('--filter FILTER', 'Filter tasks and plans by a matching substring') do |filter|
+        unless /[a-z0-9_:]+/.match(filter)
```
nicklewis: Whoops my bad, we actually want to anchor this, `/^[a-z0-9_:]+$/` otherwise it matches as long as there are _any_ of those characters in the string.

beechtom: Gotta love regex. :)

PR #1604, opened 2020-02-13
(GH-1598) Disallow config in bolt_plugin.json

### LINKED ISSUE #1598: Remove `config` section from `bolt_plugin.json` and `_config` metaparameter
We now infer this config from task parameters and pass config values as regular params. We should raise a helpful error if `config` is found in `bolt_plugin.json`

beechtom:
This removes support for a `config` field in a plugin's
`bolt_plugin.json`. Plugins now infer config from task parameters and
pass config values as task parameters.

Blocked by https://github.com/puppetlabs/puppetlabs-puppet_agent/pull/469
Closes #1598

### DIFF
--- a/lib/bolt/plugin/module.rb
+++ b/lib/bolt/plugin/module.rb
@@ -6,7 +6,7 @@ module Bolt
   class Plugin
     class Module
       class InvalidPluginData < Bolt::Plugin::PluginError
-        def initialize(plugin, msg)
+        def initialize(msg, plugin)
           msg = "Invalid Plugin Data for #{plugin}: #{msg}"
           super(msg, 'bolt/invalid-plugin-data')
         end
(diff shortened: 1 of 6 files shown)


beechtom:
A handful of tests will fail until https://github.com/puppetlabs/puppetlabs-puppet_agent/pull/469 is merged in, since the `puppet_agent` plugin has `config` set.

lib/bolt/plugin/module.rb
```diff
@@ -36,17 +36,17 @@ def initialize(mod:, context:, config:, **_opts)
...
-                   @send_config = true
-                   @data['config']
-                 else
-                   extract_task_parameter_schema
-                 end
-        @config_schema = process_schema(schema)
+
+        if @data['config']
+          msg = <<~MSG.chomp
+            Found deprecated key 'config' in bolt_plugin.json. Config for a plugin is inferred
```
nicklewis:
```suggestion
            Found unsupported key 'config' in bolt_plugin.json. Config for a plugin is inferred
```

lib/bolt/plugin/module.rb
```diff
@@ -159,17 +155,9 @@ def process_params(task, opts)
         # handled previously. That may not always be the case so filter them
         # out now.
         meta, params = opts.partition { |key, _val| key.start_with?('_') }.map(&:to_h)
+        params = @config ? config.merge(params) : params
```
nicklewis: I _believe_ `config` will always be set. If not (perhaps in tests?) we should set a default in the constructor so that this can always just be `config.merge(params)`

lib/bolt/plugin/module.rb
```diff
@@ -36,17 +36,17 @@ def initialize(mod:, context:, config:, **_opts)
...
-                   extract_task_parameter_schema
-                 end
-        @config_schema = process_schema(schema)
+
+        if @data['config']
+          msg = <<~MSG.chomp
+            Found deprecated key 'config' in bolt_plugin.json. Config for a plugin is inferred
+            from task parameters, with config values passed as parameters.
+          MSG
+          raise InvalidPluginData.new(msg, name)
```
nicklewis: It turns out every single use of this error class has the arguments in the reverse order, leading to messages like this:

```
Invalid Plugin Data for Found deprecated key 'config' in bolt_plugin.json. Config for a plugin is inferred
from task parameters, with config values passed as parameters.: test
```

Probably easier to swap the arguments in the initialize method than to update every caller. That's consistent with our other error classes which take `msg` as the first argument.

lib/bolt/plugin/module.rb
beechtom: I think it should be safe to just have it be `config.merge(params)` since `config` is a required param in the initializer, so if there ever isn't config passed it will error before reaching this point.

Issue #1600, opened 2020-02-12
`_boltdir` metaparameter is passed to agent install plugins.


adreyer:
the `_boltdir` metaparameter is passed to agent install plugins which is not expected to be useful since they generally execute on a remote target.

beechtom:
Closed by #1605

PR #1605, opened 2020-02-13
(GH-1600) Don't send _boltdir metaparam to agent installation plugins

adreyer:


lib/bolt/plugin/module.rb
```diff
@@ -170,7 +170,10 @@ def process_params(task, opts)
           params = @config ? config.merge(params) : params
           validate_params(task, params)
         end
-        params['_boltdir'] = @context.boltdir.to_s
+
+        if local
+          params['_boltdir'] = @context.boltdir.to_s
```
nicklewis: Logically it seems like this should be set on `meta` rather than `params`. And that alone should be enough to fix the bug, since the `puppet_library` method already ignores the metaparams (except for `_run_as` which it pulls out explicitly).

adreyer: good catch.

Issue #1528, opened 2020-01-13
Improve human error message when apply fails.


adreyer:
## Describe the Bug
it's hard to tell what errors came from where when an apply fails because the error messages may not reference what step of application failed or what target the failure occurred on. 

in this case the first message is tolerable but there is no way at the default log level to understand which target puppet was not installed on.
```
Finished: plan test::trusted_external in 2.86 sec
Execution of '/Users/adreyer/src/bolt/trust.sh localhost' returned 1: 
Puppet is not installed on the target, please install it to enable 'apply'
```

## Expected Behavior
Each error in the human output for an apply failure should state which part of applying code failed(ie catalog compilation, external fact gathering, etc.) and which target the failure occurred on.

## Solution
When creating the ApplyResult object we should include the target safe_name and apply phase (compilation, application, etc.) in each failure message.

PR #1607, opened 2020-02-14
(GH-1528) Provide a better error message when apply compilation fails

nicklewis:


nicklewis:
Added a changelog entry

PR #1611, opened 2020-02-15
(GH-1597) Add write_file plan function

### LINKED ISSUE #1597: Plan language function to write a file on remote targets.
## Use Case
I want to render a template in a plan and write the result to a file on targets.

`write_file($targets, "file content", dest, opts)`

- `opts` should be the same metaparameters that `file upload` takes

## Questions

How should the mode be set? We should not worry about this in this iteration. Write the file in a tmpdir with the `file::write` function and upload with `upload::file`.  Document that users should manage file permissions outside this function if the contents are sensitive.

beechtom:
This adds a new `write_file` plan function that writes content to a
specified file on the given targets. Internally, the function uses the
`file::write` function to write content to a tempfile on `localhost` and
then uses the `upload_file` function to upload the tempfile to the
destination on the given targets.

Closes #1597

### DIFF
--- a/bolt-modules/boltlib/lib/puppet/functions/write_file.rb
+++ b/bolt-modules/boltlib/lib/puppet/functions/write_file.rb
@@ -1,0 +1,48 @@
+# frozen_string_literal: true
+
+require 'tempfile'
+
+# Write contents to a file on the given set of targets.
+#
+# > **Note:** Not available in apply block
+Puppet::Functions.create_function(:write_file) do
+  # @param targets A pattern identifying zero or more targets. See {get_targets} for accepted patterns.
+  # @param content File content to write.
+  # @param destination An absolute path on the target(s).
+  # @option options [Boolean] _catch_errors Whether to catch raised errors.
+  # @option options [String] _run_as User to run as using privilege escalation.
+  # @return A list of results, one entry per target.
+  # @example Write a file to a target
+  #   $content = 'Hello, world!'
+  #   write_file($targets, $content, '/Users/me/hello.txt')
+  dispatch :write_file do
+    required_param 'String', :content
+    required_param 'String[1]', :destination
+    required_param 'Boltlib::TargetSpec', :targets
+    optional_param 'Hash[String[1], Any]', :options
+    return_type 'ResultSet'
+  end
+
+  def write_file(content, destination, target_spec, options = {})
+    unless Puppet[:tasks]
+      raise Puppet::ParseErrorWithIssue
+        .from_issue_and_stack(Bolt::PAL::Issues::PLAN_OPERATION_NOT_SUPPORTED_WHEN_COMPILING,
+                              action: 'write_file')
+    end
+
+    executor = Puppet.lookup(:bolt_executor)
+    executor.report_function_call(self.class.name)
+
+    inventory = Puppet.lookup(:bolt_inventory)
+    targets = inventory.get_targets(target_spec)
+
+    executor.log_action("write file #{destination}", targets) do
+      executor.without_default_logging do
+        Tempfile.create do |tmp|
+          call_function('file::write', tmp.path, content)
+          call_function('upload_file', tmp.path, destination, targets, options)
+        end
+      end
+    end
+  end
+end
--- a/spec/integration/modules/write_file_spec.rb
+++ b/spec/integration/modules/write_file_spec.rb
@@ -1,0 +1,81 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+require 'bolt_spec/conn'
+require 'bolt_spec/files'
+require 'bolt_spec/integration'
+require 'bolt_spec/run'
+
+describe 'running a plan with write_file' do
+  include BoltSpec::Conn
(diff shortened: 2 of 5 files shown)


spec/integration/modules/write_file_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+        result = run_command("#{print_cmd} #{filename}", transport.to_s, inventory: inventory)
+
+        expect(result.size).to eq(1)
+        data = result.first
+        expect(data['status']).to eq('success')
+        expect(data['value']['stdout']).to match(/#{content}/)
+      end
+    end
+
+    it 'reports multiple function calls to analytics' do
```
beechtom: I originally had this in the unit tests, but kept getting an error that `file::write` and `upload_file` weren't functions. I'm not entirely sure why this was happening, but I'm thinking the environment the unit tests run in don't load all the functions?

spec/integration/modules/write_file_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+  end
+
+  shared_examples 'when writing a file from a plan' do
+    # Delete the uploaded file after each test
+    after(:each) do
+      result = run_command("#{remove_cmd} #{filename}", transport.to_s, inventory: inventory)
+      expect(result.first['status']).to eq('success')
+    end
+
+    it 'writes a file to the destination' do
```
beechtom: I'm not sure if we needed any other tests for this function besides this one, since pretty much all of the logic is contained in `file::write` and `upload_file` and they have their own tests.

bolt-modules/boltlib/lib/puppet/functions/write_file.rb
```diff
@@ -0,0 +1,44 @@
...
+  # @param content File content to write.
+  # @param destination An absolute path on the target(s).
+  # @option options [Boolean] _catch_errors Whether to catch raised errors.
+  # @option options [String] _run_as User to run as using privilege escalation.
+  # @return A list of results, one entry per target.
+  # @example Write a file to a target
+  #   $content = 'Hello, world!'
+  #   write_file($targets, $content, '/Users/me/hello.txt')
+  dispatch :write_file do
+    required_param 'Boltlib::TargetSpec', :targets
```
nicklewis: For consistency with the `run_*` functions, we might want `$targets` to come last.

bolt-modules/boltlib/lib/puppet/functions/write_file.rb
```diff
@@ -0,0 +1,44 @@
...
+  def write_file(targets, content, destination, options = {})
+    unless Puppet[:tasks]
+      raise Puppet::ParseErrorWithIssue
+        .from_issue_and_stack(Bolt::PAL::Issues::PLAN_OPERATION_NOT_SUPPORTED_WHEN_COMPILING,
+                              action: 'write_file')
+    end
+
+    Puppet.lookup(:bolt_executor) {}&.report_function_call(self.class.name)
+
+    Tempfile.create do |tmp|
```
nicklewis: We should follow the `apply_prep` pattern and use `log_action`+`without_default_logging` here so we don't get log messages like `Starting: file upload from /var/folders/n1/mt59jhjd6l7b4122zbby3pk00000gn/T/20200218-26367-6x9har to /tmp/test.txt on dumb-formality.delivery.puppetlabs.net`.

nicklewis: Maybe something like `log_action("write file #{destination}", targets)`

bolt-modules/boltlib/lib/puppet/functions/write_file.rb
```diff
@@ -0,0 +1,49 @@
...
+    unless Puppet[:tasks]
+      raise Puppet::ParseErrorWithIssue
+        .from_issue_and_stack(Bolt::PAL::Issues::PLAN_OPERATION_NOT_SUPPORTED_WHEN_COMPILING,
+                              action: 'write_file')
+    end
+
+    executor = Puppet.lookup(:bolt_executor)
+    executor.report_function_call(self.class.name)
+
+    executor.log_action("write file #{destination}", targets) do
```
nicklewis: We need to call `get_targets` on `targets`. Otherwise the logging fails since it expects `targets` to be an array. It probably makes sense to call that once and pass it to `log_action` and `call_function('upload_file', [...])`

bolt-modules/boltlib/lib/puppet/functions/write_file.rb
```diff
@@ -0,0 +1,49 @@
+# frozen_string_literal: true
+
+require 'tempfile'
+
+# Write contents to a file on the given set of targets.
+#
+# > **Note:** Not available in apply block
+#
+# > **Note:** Files are granted 0600 permissions. If a file contains sensitive content
```
nicklewis: What causes this to happen?

beechtom: Removed. I had put that when reading about `Tempfile#create`, but forgot permissions could be different when the file is written to another system.

Issue #1614, opened 2020-02-18
BoltSpec::Run modifies the inventory.


adreyer:
## Describe the Bug
When a user passes an inventory hash to BoltSpec::Run functions the version number is deleted from the original hash. This means that a version 2 inventory hash cannot be used to run two consecutive actions.  

## Expected Behavior
No objects passed to the `run_*` functions should be modified.

## Workaround

deep copy the hash before passing it to a run function or add the `version` key back in before each call.

glennsarti:
Hit this today in puppet-strings testing 😢

glennsarti:
Also affects `upload_file` too  😢😢😢😢😢

PR #1620, opened 2020-02-19
(GH-1614) Deep copy config and inventory for the BoltSpec Runner

adreyer:
Inventory can be huge when read from a file and so the Inventory deletes
the version key to avoid copying the whole inventory. This will cause
BoltSpec to copy the inventory before creating an Inventory instance to
avoid mutating the callers inventory hash.

### DIFF
--- a/lib/bolt_spec/run.rb
+++ b/lib/bolt_spec/run.rb
@@ -131,7 +131,7 @@ module BoltSpec
       # still be loaded
       def self.with_runner(config_data, inventory_data)
         Dir.mktmpdir do |boltdir_path|
-          runner = new(config_data, inventory_data, boltdir_path)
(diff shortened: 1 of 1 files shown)


nicklewis:
Is there a reason we shouldn't just leave the `version` key on the inventory in `create_version`?

glennsarti:
Shouldn't this fix also make it into Bolt 1.x branch too?  Seems an egregious error that should be backported.

adreyer:
@glennsarti no, we're not doing any backports.
We chose to invest in making the upgrade process simple rather than long term support for 1.0 Bolt is still relatively young and we want to push users to upgrade to prevent fractures in the ecosystem.

@nicklewis I think it's safer for us to copy at this level and it will allow us to optimize for the more common CLI use case later. I think we'll soon crystalize this as part of the ruby API work.

PR #1625, opened 2020-02-20
(maint) Load inventory more efficiently

nicklewis:
The `resolve_string_targets` method was inefficient since it recomputed
the complete set of target names in the group for each target. We now
only compute the set once and keep it updated as we evaluate targets.
This saves ~0.5ms per target, which adds up when for users with
thousands or more targets.

### DIFF
--- a/lib/bolt/inventory/group.rb
+++ b/lib/bolt/inventory/group.rb
@@ -189,6 +189,9 @@ module Bolt
       end
 
       def resolve_string_targets(aliases, known_targets)
+        # Use a single copy of this because recomputing it for every target is
+        # prohibitively expensive for large groups
+        cached_local_targets = local_targets
         @string_targets.each do |string_target|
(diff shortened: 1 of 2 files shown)


lib/bolt/inventory/group.rb
```diff
@@ -197,20 +200,21 @@ def resolve_string_targets(aliases, known_targets)
             @unresolved_targets[string_target] = { 'name' => string_target }
           # If this is an alias for an existing target, then add it to this group
           elsif (canonical_name = aliases[string_target])
-            if local_targets.include?(canonical_name)
+            if cached_local_targets.include?(canonical_name)
               @logger.warn("Ignoring duplicate target in #{@name}: #{canonical_name}")
```
adreyer: Is this warning going to show up incorrectly now?

adreyer:
needs a changelog but it looks fine otherwise.

PR #1627, opened 2020-02-21
(maint) Validate inventory more efficiently

beechtom:
This updates the inventory to only validate groups once all targets have
been expanded. Previously, when calling `get_targets` and creating each
target, the inventory would add that target to the necessary group and
then validate all groups with `@groups.validate`. This resulted in large
inventories taking an exponentially long time to load. Now, groups are
only validated once all targets have been created and added to groups.

### DIFF
--- a/lib/bolt/inventory/inventory.rb
+++ b/lib/bolt/inventory/inventory.rb
@@ -168,7 +168,6 @@ module Bolt
       def add_target(current_group, target, desired_group)
         if current_group.name == desired_group
           current_group.add_target(target)
-          @groups.validate
           target.invalidate_group_cache!
           return true
         end
@@ -200,6 +199,8 @@ module Bolt
         # If target already exists, delete old and replace with new, otherwise add to new to all group
         new_target = Bolt::Inventory::Target.new(data, self)
         existing_target = @targets.key?(new_target.name)
+
+        validate_target_from_hash(new_target)
         @targets[new_target.name] = new_target
 
         if existing_target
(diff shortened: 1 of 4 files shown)


lucywyman:
I think this just means that we need to move `add_to_group(..., all)` to the top of https://github.com/puppetlabs/bolt/blob/master/lib/bolt/inventory/inventory.rb#L198? Or abstract the group name check to it's own function and do that before checking if a target of the same name exists and removing it.

nicklewis:
We need to do the validation _before_ we add anything to the inventory, otherwise if you use `run_plan(..., _catch_errors => true)` to run a plan that makes invalid modifications to the inventory, the calling plan will be left with an invalid inventory to deal with. So I think we need to copy the relevant pieces of validation from `Group#validate` into `create_target_from_hash`.

lib/bolt/inventory/inventory.rb
```diff
@@ -209,16 +210,46 @@ def create_target_from_hash(data)
...
+        target.target_alias.each do |alia|
+          if groups.include?(alia)
+            raise ValidationError.new("Alias #{alia} conflicts with group of the same name", nil)
+          end
+
+          if targets.include?(alia)
+            raise ValidationError.new("Alias #{alia} conflicts with target of the same name", nil)
+          end
+
+          next unless aliases[alia] && aliases[alia] != target.name
```
nicklewis: This should just be another `if aliases[alia] && aliases[alia] != target.name`

lib/bolt/inventory/inventory.rb
```diff
@@ -209,16 +210,46 @@ def create_target_from_hash(data)
         end
 
         if (aliases = new_target.target_alias)
+          @groups.insert_alia(new_target.name, aliases)
+        end
+
+        new_target
+      end
+
+      def validate_target_from_hash(target, _existing)
```
nicklewis: Looks like `_existing` is unused, so should it just be removed?

lib/bolt/inventory/inventory.rb
beechtom: Rubocop is forcing this with our current cops. I'll get that updated.

lib/bolt/inventory/inventory.rb
```diff
@@ -209,16 +210,47 @@ def create_target_from_hash(data)
         end
 
         if (aliases = new_target.target_alias)
+          @groups.insert_alia(new_target.name, aliases)
```
nicklewis: The validation accepts a String or Array, so this should be `Array(aliases)` to ensure we pass an Array.

nicklewis: This variable also seems unnecessary. This seems fine:

```rb
if new_target.target_alias
  @groups.insert_alia(new_target.name, Array(new_target.target_alias))
end
```

lib/bolt/inventory/inventory.rb
```diff
@@ -209,16 +210,43 @@ def create_target_from_hash(data)
...
+        targets = target_names
+
+        # Make sure there are no group name conflicts
+        if groups.include?(target.name)
+          raise ValidationError.new("Target name #{target.name} conflicts with group of the same name", nil)
+        end
+
+        # Validate any aliases
+        if (aliases = target.target_alias)
           aliases = [aliases] if aliases.is_a?(String)
```
nicklewis: This was previously a helpful step because we reused this variable to pass to `insert_alia`. That's not the case anymore, so this conversion isn't helpful. I think we can remove this line and just check `unless aliases.is_a?(Array) || aliases.is_a?(String)` on the line below.

spec/integration/inventory_spec.rb
```diff
@@ -326,8 +326,7 @@ def fact_plan(name = 'facts_test')
       plan = ['plan', 'run', 'add_group::x_fail_group_name_exists', '--targets', 'add_me'] + config_flags
       result = run_cli_json(plan)
       expect(result['kind']).to eq('bolt.inventory/validation-error')
-      expect(result['msg']).to match(/Group foo conflicts with target of the same name for group/)
-      expect(result['details']).to eq("path" => ["foo"])
+      expect(result['msg']).to match(/Target name foo conflicts with group of the same name/)
```
nicklewis: Heyyyy we even improved this error message

spec/bolt/inventory/inventory_spec.rb
```diff
@@ -1324,4 +1324,59 @@ def common_data(transport)
       expect(target.detail).to eq(expected_data)
     end
   end
+
+  describe 'creating a target from a hash' do
```
nicklewis: These tests are great.

lib/bolt/inventory/inventory.rb
```diff
@@ -208,17 +209,43 @@ def create_target_from_hash(data)
...
             raise ValidationError.new(msg, @name)
           end
-
-          @groups.insert_alia(new_target.name, aliases)
         end
 
-        new_target
+        # Make sure there are no conflicts with the new target aliases
+        aliases = @groups.target_aliases
+        target.target_alias.each do |alia|
```
nicklewis: This is another place where we need to handle that `target_alias` may be a String. Since we've validated it at this point, this line could just be `Array(target.target_alias).each` since we've validated it by this point.

We should also not reuse the `aliases` variable name for two different contexts. Perhaps call the second one `used_aliases` or `known_aliases` something.

nicklewis:
This cuts `bolt inventory show --target all` with 25k targets from (projected) ~2 hours to ~2 minutes.

PR #1628, opened 2020-02-21
(GH-1612) Add --puppetfile CLI option

### LINKED ISSUE #1612: Ability to pass puppetfile
## Use Case
We our automating our system using Bolt and Terraform and have a need to pass the Puppetfile directly in a container and install into a particular directory. 

An additional flag to allow for path to Puppetfile during the `bolt puppetfile install` command

We have considered moving the file to the `.bolt` directory but also run outside of containers and do not want to have a race condition where more than one job is trying to do that with different dependencies.

## Solution
Add a `--puppetfile` flag to the `bolt puppetfile install` command which accepts a path (relative or absolute) to a Puppetfile. The modules will be intsalled to whatever the current Boltdir is.

beechtom:
This adds support for a `--puppetfile` CLI option which accepts a
relative or absolute path to a Puppetfile to use when installing
modules with `bolt puppetfile install`. When using this flag, modules
will be installed to the current Boltdir. If a relative path is passed
to the option, it will be relative to the current Boltdir.

Closes #1612

### DIFF
--- a/lib/bolt/config.rb
+++ b/lib/bolt/config.rb
@@ -302,6 +302,8 @@ module Bolt
         send("#{key}=", options[key]) if options.key?(key)
       end
 
+      @puppetfile = options[:puppetfile] if options.key?(:puppetfile)
+
       @save_rerun = options[:'save-rerun'] if options.key?(:'save-rerun')
 
       if options[:debug]
@@ -386,7 +388,7 @@ module Bolt
     end
 
     def puppetfile
-      @boltdir.puppetfile
+      @puppetfile || @boltdir.puppetfile
     end
 
     def modulepath
--- a/spec/bolt/cli_spec.rb
+++ b/spec/bolt/cli_spec.rb
@@ -517,6 +517,16 @@ describe "Bolt::CLI" do
       end
     end
 
+    describe "puppetfile" do
+      let(:puppetfile) { File.expand_path('/path/to/Puppetfile') }
+      let(:cli) { Bolt::CLI.new(%W[puppetfile install --puppetfile #{puppetfile}]) }
+
+      it 'uses a specified Puppetfile' do
+        cli.parse
+        expect(cli.config.puppetfile.to_s).to eq(puppetfile)
+      end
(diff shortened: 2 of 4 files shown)


spec/bolt/cli_spec.rb
```diff
@@ -1396,9 +1406,9 @@ def stub_config(file_content = {})
 
             expect { cli.execute(options) }.to raise_error(
               Bolt::PAL::PALError,
-              /Task sample::params:\n(?x:
+              %r{Task sample::params:\n(?x:
                )\s*has no parameter named 'foo'\n(?x:
-               )\s*has no parameter named 'bar'/
+               )\s*has no parameter named 'bar'}
```
beechtom: This fixes an issue with syntax highlighting in VS.

lib/bolt/config.rb
```diff
@@ -302,6 +302,10 @@ def apply_overrides(options)
         send("#{key}=", options[key]) if options.key?(key)
       end
 
+      if options.key?(:puppetfile)
+        @puppetfile = Pathname.new(File.expand_path(options[:puppetfile], @boltdir.path))
```
nicklewis: This should be expanded in the option parser relative to cwd (exactly what we do for `inventoryfile` and `modulepath`). It might be good to standardize those other settings to also return `Pathname` objects so we don't have different behavior depending on whether they came from a Boltdir or not.

beechtom: Updated to expand from cwd and have the parser return a `Pathname` object for `inventoryfile`. It looks like PAL expects a string for `modulepath`, so left that one as-is.

spec/bolt/config_spec.rb
```diff
@@ -307,6 +307,16 @@
       expect(config.inventoryfile)
         .to eq(File.expand_path('targets.yml', boltdir.path))
     end
+
+    it "expands puppetfile relative to boltdir" do
```
nicklewis: This isn't accurate anymore

PR #1631, opened 2020-02-25
(GH-1630) Ability to mock out plans in spec tests

### LINKED ISSUE #1630: Ability to mock out the run_plan() in BoltSpec
## Use Case
I have a plan that calls another plan with the `run_plan()` function: https://github.com/EncoreTechnologies/puppet-patching/blob/master/plans/check_puppet.pp#L77-L84

I am trying to write spec tests for these plans, but am not able to mock out the `run_plan` function.

`expect_plan` and `allow_plan` action stubs similar to `expect_task` and `allow_task`.

## Additional Context
I've tried investigating writing these myself, but it looks like the `run_plan` function is implemented differently than the other `run_*` commands which use the executor. The ActionStubs spec class seems to mock out the executor, whereas the `run_plan` function looks like it delegates to a `loader`, then i got lost.

nmaludy:
Closes #1630 

This is some initial work on the ability to mock out `run_plan()`

**Changes**
- Based on a suggestion from Nick Lewis i moved a part of the `run_plan` function into the `Bolt::Executor` so it could be easily mocked like the other functions.

**TODOs**
- [x] Add documentation to `bolt_spec/plans.rb`
- [x] How do we catch and test for errors in plans (`fail_plan`) ?
- [x] How do we catch "any" plan is this even possible? 
- [x] Test using the `return` with a block
- [x] Flag for enabling / disabling allowing all plan executions or not
- [x] Fix normal spec tests (looks like several CLI tests are broken)
- [x] Implemented erroring from a plan (`fail_plan()`)
- [x] Raise exception when `return_for_targets` is called for plan stubs


lib/bolt_spec/plans/action_stubs/plan_stub.rb
```diff
@@ -0,0 +1,47 @@
...
+          return false
+        end
+
+        true
+      end
+
+      def call(_scope, plan, params)
+        @calls += 1
+        if @return_block
+          check_plan_result(@return_block.call(plan: plan, params: params), plan)
```
nicklewis: `check_plan_result` isn't defined anywhere.

More generally, the `ActionStub` base class is pretty specific to an action that gets performed on a set of targets and returns a `ResultSet`. It may be more trouble than it's worth to try to adapt that to also work for plans, so we might want to just implement `PlanStub` on its own without inheriting behavior from `ActionStub`.

nicklewis:
This is pretty close and the general approach seems great. I wrote a test that uses it, but the API provided `ActionStub` proves to be a bit of a minefield trying to use it for a plan, as mentioned in my other comment.

adreyer:
I think the behavior of plan mocking needs to differ from command/task etc mocking. With those actions it's important that our unit tests never actually execute them. Plans on the other hand can reasonably be executed as long as all of their commands/tasks etc are mocked out. For example a user should be able to write a plan and make assertions about the tasks run then refactor that plan to use subplans and still run the same unit tests.


- Plans are sometimes public APIs in which case mocking them out during testing makes sense.
- Plans are sometimes internal implementation details and we should be able to refactor them without breaking tests.
- Tasks/Scripts/etc should never be actually executed during unit tests but it may make sense to execute plans.
- The `allow_any_*` method is not adequate by itself for plans since sometimes we want to actually execute them. 
- For backwards compatibility subplans should by default be executed.
- There should probably be some flag to manage plan execution that is similar to allow/expect... `execute_any_plan` would execute any plan that doesn't match a mock. `execute_no_plan`(`no_execute_any_plan`?) would unset that behavior.  `execute_plan('my_plans::subplan') would enable execution for a specific plan.

lib/bolt_spec/plans/action_stubs.rb
```diff
@@ -55,7 +57,10 @@ def assert_called(object)
           end
           message = "Expected #{object} to be called #{times} times"
           message += " with targets #{@invocation[:targets]}" if @invocation[:targets]
-          message += " with parameters #{parameters}" if parameters
+          if parameters
+            parameters_str = JSON.parse(parameters.to_json)
```
adreyer: This seems fine but probably should have a comment. We probably should discuss whether to add better `.inspects` to the plan language datatype classes.

bolt-modules/boltlib/lib/puppet/functions/run_plan.rb
```diff
@@ -117,7 +117,7 @@ def run_inner_plan(scope, plan_name, targets, args = {})
         # undef/nil
         result = catch(:return) do
           scope.with_global_scope do |global_scope|
-            closure.call_by_name_with_scope(global_scope, params, true)
+            executor.run_plan(global_scope, closure, params)
```
adreyer: @McdonaldSeanp I think this will help you longterm too but heads up.

lib/bolt_spec/plans/action_stubs/plan_stub.rb
nmaludy: `check_plan_result` is now defined in action_stubs.rb

lib/bolt_spec/plans/action_stubs.rb
nmaludy: Added a comment here.

nmaludy:
@adreyer I just added the `execute_any_plan` / `execute_no_plan` flags. There are some examples in the spec tests if you want to take a look!

nmaludy:
Just need to add documentation to `BoltSpec::Plans`, but the code is feature complete!

nmaludy:
OK, documentation added to `BoltSpec::Plans`, release notes were added to my latest commit... I think this one is good for another review (no more outstanding TODOs from my end)!

bolt-modules/boltlib/lib/puppet/functions/run_plan.rb
McdonaldSeanp: I finally got around to taking a look. Thanks for the heads up, we'll make sure `run_plan` is available in PE's executor before we pull in the bolt update with this.

Long term I think you're right that this should be helpful, because the PE executor can intercept `run_plan` calls now, which allows us to make it more stable from the PE side.

lib/bolt_spec/plans/action_stubs.rb
```diff
@@ -127,7 +153,10 @@ def return(&block)
       def return_for_targets(data)
         data.each_with_object(@data) do |(target, result), hsh|
           raise "Mocked results must be hashes: #{target}: #{result}" unless result.is_a? Hash
-          hsh[target] = result_for(Bolt::Target.new(target), Bolt::Util.walk_keys(result, &:to_sym))
+          # set the inventory from the BoltSpec::Plans, otherwise if we try to convert
+          # this target to a string, it will fail to string conversion because the
+          # inventory is nil
+          hsh[target] = result_for(Bolt::Target.new(target, @inventory), Bolt::Util.walk_keys(result, &:to_sym))
```
lucywyman, line 159: Does the plan stub of `result_for` use the target parameter? I'm not sure in what cases this breaks if the inventory is nil...

lib/bolt_spec/plans/mock_executor.rb
```diff
@@ -28,6 +29,13 @@ def initialize(modulepath)
         MOCKED_ACTIONS.each { |action| instance_variable_set(:"@#{action}_doubles", {}) }
         @stub_out_message = nil
         @transport_features = ['puppet-agent']
+        @executor_real = Bolt::Executor.new
```
lucywyman, line 32: https://media.giphy.com/media/KwshQCykFmTlK/giphy.gif

lib/bolt_spec/plans/action_stubs.rb
lucywyman: This would also mean we don't have to pass inventory here.

lib/bolt_spec/plans.rb
```diff
@@ -40,22 +40,44 @@
...
+#  build complex workflows. To support this we offer running in two different
+#  modes:
+#    execute_any_plan (default) - This mode will execute any plan that is encountered
+#      without having to be stubbed/mocked. This default mode allows for plan control
+#      flow to behave as normal. If you choose to stub/mock out a sub-plan in this mode
+#      that will be honored and the sub-plan will not be executed. We will use the modifiers
+#      on the stub to check for the conditions specified (example: be_called_times(3))
+#
+#    execute_no_plan - This mode will not execute a plans that it encounters. Instead, when
+#      an plan is encountered it will throw an error unless the plan is mocked out. This
```
beechtom, line 55:
```suggestion
#      a plan is encountered it will throw an error unless the plan is mocked out. This
```

lib/bolt_spec/plans/action_stubs.rb
nmaludy: no, it doesn't use the target parameter, it just silently ignores it: https://github.com/puppetlabs/bolt/pull/1631/files#diff-075d987aae32893fba7412f7e47a1b2dR33

nmaludy: @lucywyman do you want me to change something here?

nmaludy: We need to pass the inventory here, because otherwise if there is an error and this target is printed out, it will fail to convert the `target` to a string and throw a different exception (PAL Error).

PR #1633, opened 2020-02-25
Inventory performance improvements

nicklewis:
This makes a couple more changes to improve the performance of loading the inventory.

**Don't copy entire config to merge transport data**

Previously, we made a deep copy of and validated the entire Bolt::Config 
object for every Target. That's quite expensive and creates a huge amount
of GC overhead, especially given that we're cloning complex active objects
like loggers and hitting the filesystem to expand paths while validating.
The only reason we did this copying was because the logic that knew how to
merge two transport configs together happened to live on the Bolt::Config
object and happened to make in-place modifications to the config.

That method is now abstracted to a class method on Bolt::Config so it can
be used by the Target object without copying the entire config. We now only
make a clone of the transport config hash, which is a much simpler ordeal.

**Add contains_target? method in place of enumerating all targets**

In several places in inventory, we use the `local_targets` method to check
whether a given target is contained by the group. That method computes the
list of targets every time, which is prohibitively expensive for large
inventories. We now use a `contains_target?` method which checks for the
presence of the target without making a list.

### DIFF
--- a/lib/bolt/inventory/target.rb
+++ b/lib/bolt/inventory/target.rb
@@ -109,14 +109,16 @@ module Bolt
       def transport_config_cache
         if @transport_config_cache.nil?
           merged_config = Bolt::Util.deep_merge(group_cache['config'], @config)
-          # Use the base config to ensure we handle the config validation and
-          # munging correctly
-          config = @inventory.config.deep_clone
-          config.update_from_inventory(merged_config)
-          config.validate
+          base_config = @inventory.config
+          transport_data = Bolt::Util.deep_clone(base_config.transports)
+          Bolt::Config.update_transport_hash(base_config.boltdir.path, transport_data, merged_config)
+          transport = merged_config['transport'] || base_config.transport
+          Bolt::TRANSPORTS.each do |name, impl|
+            impl.validate(transport_data[name])
+          end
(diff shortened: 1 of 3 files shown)


lib/bolt/inventory/target.rb
```diff
@@ -109,14 +109,19 @@ def invalidate_config_cache!
...
         if @transport_config_cache.nil?
           merged_config = Bolt::Util.deep_merge(group_cache['config'], @config)
-          # Use the base config to ensure we handle the config validation and
-          # munging correctly
-          config = @inventory.config.deep_clone
-          config.update_from_inventory(merged_config)
-          config.validate
+          base_config = @inventory.config
+          transport_data = Bolt::Util.deep_clone(base_config.transports)
+          Bolt::Config.update_transport_hash(base_config.boltdir.path, transport_data, merged_config)
```
nicklewis: This change is a bit weird, but it will be superseded by work that @beechtom is doing around how we handle transport config, so it seems worth the performance improvement for now.

nicklewis:
When applied on top of #1627, this cuts `bolt inventory show --target all` with 25k nodes from ~2 minutes to ~5 seconds.

PR #1636, opened 2020-02-25
(docs) Reorganize writing tasks doc

nicklewis:
The order of these sections was unnatural and didn't map well to either
their importance or to how users might adopt them.

* Move 'secure coding practices' and 'writing remote tasks' to the end, as those are specialized topics.
* Move 'defining parameters'
* Move returning errors under structured input and output
* Move task metadata basics before no-op, since no-op relies on metadata
* Rename more detailed task metadata docs to task metadata reference
* Remove 'specifying parameters' as that is about running tasks, not writing them

### DIFF
--- a/documentation/writing_tasks.md
+++ b/documentation/writing_tasks.md
@@ -40,113 +40,6 @@ rescue Puppet::Error => e
 end
 ```
 
-## Secure coding practices for tasks
-
-Use secure coding practices when you write tasks and help protect your system.
-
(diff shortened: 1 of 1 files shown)


hestonhoffman:
Still working on editing this, but I'm also going to have to figure out what to do with the PE versions of this page. More on that later

hestonhoffman:
I think this is good to merge in. I'm going to iterate on it a little further, but I'll open a separate PR for that. This is already a huge improvement.

PR #1637, opened 2020-02-26
(maint) Ignore markdown files in root for CI

beechtom:
This updates the GitHub workflows to ignore markdown files at the root
of the repository, which were not caught by the glob pattern in the
`paths-ignore` field.

### DIFF
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -3,10 +3,10 @@ name: Docs
 on:
   push:
     branches: [master]
-    paths-ignore: ['**/*.md']
+    paths-ignore: ['**.md']
   pull_request:
     type: [opened, reopened, edited]
-    paths-ignore: ['**/*.md']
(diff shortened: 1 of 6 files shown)


.github/workflows/docs.yaml
```diff
@@ -3,10 +3,10 @@ name: Docs
 on:
   push:
     branches: [master]
-    paths-ignore: ['**/*.md']
+    paths-ignore: ['*.md', '**/*.md']
```
nicklewis: Reading [the docs](https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet), I think we want `**.md` in place of both of these.

beechtom: Updated. Hopefully that will work... the docs seem to say that `**/*.md` should have caught all markdown files in the repo. :/

PR #1638, opened 2020-02-27
(maint) Bump puppetlabs-reboot module

lucywyman:
The latest puppetlabs-reboot version updates the deprecated `nodes`
parameter to `targets`. This is a breaking change for users who
explicitly set the parameter, either in the `run_plan` plan function or
on the CLI. Users will need to update the `nodes` parameter to be
`targets` instead. Calling the plan using `-t` or with
`run_plan('reboot', $mytargets)` will both behave the same and do not
need to be updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@
   The `puppetfile install` command now supports a `--puppetfile` option that accepts a relative or absolute path
   to a Puppetfile.
 
-* **Update reboot plan parameter `nodes` to `targets`** ([puppetlabs-reboot#223](https://github.com/puppetlabs/reboot/pulls/223)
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -13,6 +13,12 @@
   The `puppetfile install` command now supports a `--puppetfile` option that accepts a relative or absolute path
   to a Puppetfile.
 
+* **Update reboot plan parameter `nodes` to `targets`** ([puppetlabs-reboot#223](https://github.com/puppetlabs/reboot/pulls/223)
```
```suggestion
* **Update reboot plan parameter `nodes` to `targets`** ([puppetlabs-reboot#223](https://github.com/puppetlabs/reboot/pulls/223))
```

PR #1644, opened 2020-02-29
(GH-1591) Expose plugins in config file transport config

### LINKED ISSUE #1591: using a plugin to set transport config in bolt.yaml causes cryptic error
> If I put the same settings[vault plugin lookup for ssh key] in the bolt.yaml file, I get a strange error: “undefined method `match’ for #<Hash:0x0000000003dff608>”

## Solution
Raise a helpful error message when a `_plugins` key is somewhere in the config that it can't be used. We should also make plugins available in transport config in the config file.

Additional validation of the config would be helpful.

beechtom:
This adds support for using plugins under transport config settings in a
config file. Plugins references under transport config are resolved
immediately after plugins are configured and before inventory is loaded.

This also includes a refactoring of the `Bolt::Config` class and the
introduction of new transport configuration classes (e.g.
`Bolt::Config::SSH`, `Bolt::Config::WinRM`). Each transport config
object handles merging and validating transport config options set at
the config, group, and target levels, while the `Bolt::Config` class
will now only load config files, pass data along to the appropriate
classes, and validate Bolt-specific config.

Previously, transport configuration was validated and handled completely
by the `Bolt::Config` class. When resolving transport configuration for
specific targets, this required cloning the entire config object,
passing the group- and target-specific data to the config object,
re-validating the config, and pulling the transport-specific config out
of the config object. This approach was expensive and created
unnecessary duplicate objects, which slowed down the loading and
validating of large inventories.

Closes #1591 

## TODO

- [x] Update broken transport integration tests
- [x] Pass transport config to inventory instead of config
- [x] Type validation from `OPTIONS` hash on each transport config object
- [x] ~~Look at including `defaults` and `advisory_defaults` to `OPTIONS` hash~~
- [x] Resolve transport config plugin references from inventory
- [x] Update `Bolt::Config::Transport#merge` to accept multiple hashes
- [x] Add changelog entry


lib/bolt/config/transport.rb
```diff
@@ -0,0 +1,133 @@
...
+
+      def [](key)
+        @config[key]
+      end
+
+      def to_h
+        @config
+      end
+
+      def fetch(*args)
```
beechtom: This is used by the docker and sudoable transports, which currently have `target.options.fetch`. I can either update `Bolt::Inventory::Target#options` to return the transport config hash, or update the transports to `target.options.to_h.fetch`. Not sure which would be preferable, so I put this in for now.

lib/bolt/config/transport.rb
```diff
@@ -0,0 +1,133 @@
...
+        @config.fetch(*args)
+      end
+
+      def include?(args)
+        @config.include?(args)
+      end
```
beechtom: This is used by a handful of spec tests. Similar to `#fetch` above, the tests call `target.options.include?`. I could either update `#options` to return the config hash or just call `targets.options.to_h.include?`.

lib/bolt/config/transport.rb
```diff
@@ -0,0 +1,133 @@
...
+
+      def config=(data)
+        assert_hash(data)
+        @config = data
+        validate
+      end
+
+      def merge(*data)
```
beechtom: This method currently only accepts hashes when merging transport config. I thought it might be nice to have it also accept `self.class`, but the only place merging happens is when merging the group cache config and target config, which are both hashes.

lib/bolt/config/transport.rb
```diff
@@ -0,0 +1,133 @@
...
+      end
+
+      # Checks whether a config option contains a plugin reference, which
+      # should always be a valid input
+      private def reference?(value)
+        value.is_a?(Hash) && value.key?('_plugin')
+      end
+
+      # Generic type validation for config options
+      private def validate_type(type, *opts)
```
beechtom: TODO: Remove - superseded by `#assert_type`

lib/bolt/transport/local_windows.rb
```diff
@@ -98,14 +70,12 @@ def execute(*command, options)
         result_output
       end
 
-      def upload(target, source, destination, options = {})
-        self.class.validate(options)
```
beechtom: None of the other transports validate options in `#upload` or the `#run_*` methods. It looks like this is only here so we can issue a warning when the transport receives the `run-as` option, either through config or as an option in a plan function.

Issuing a warning from the transport config class is a little tricky, since the Logger is configured after the class. We could add a list of warnings in the transport config that are printed later (similar to what we do for the option parser).

Adding a specific warning for instances when the `_run_as` option is used in a plan function should be easy enough.

lib/bolt/config.rb
```diff
@@ -210,169 +237,122 @@ def merge_config_data(config_data)
...
-      end
+        if (v = acc[name][:level])
+          unless Bolt::Logger.valid_level?(v)
+            raise Bolt::ValidationError,
+                  "level of log #{name} must be one of #{Bolt::Logger.levels.join(', ')}; received #{v}"
+          end
+          unless v.is_a?(String) || v.is_a?(Symbol)
+            raise Bolt::ValidationError,
+                  "level of log #{name} must be a String or Symbol, received #{v.class} #{v.inspect}"
+          end
```
beechtom: TODO: Remove - It looks like this is redundant, since `#valid_level?` will already return false if the level isn't a valid string or symbol.

lib/bolt/bolt_option_parser.rb
```diff
@@ -720,7 +720,7 @@ def initialize(options)
       define('--puppetfile FILEPATH',
              'Specify a Puppetfile to use when installing modules. (default: ~/.puppetlabs/bolt/Puppetfile)',
              'Modules are installed in the current Boltdir.') do |path|
-        @options[:puppetfile] = Pathname.new(File.expand_path(path))
+        @options[:puppetfile_path] = Pathname.new(File.expand_path(path))
```
beechtom: This is needed to get around the hash slicing used in `Bolt::Config`, since puppetfile config is under the `puppetfile` key.

lib/bolt/inventory/target.rb
```diff
@@ -176,13 +167,19 @@ def password
       end
 
       def options
-        transport_config.dup
+        transport_config
       end
```
beechtom: I'm not sure if this wants to return the `Bolt::Config::Transport` object or just its config hash. I'm leaning towards the former..

documentation/bolt_configuration_reference.md.erb
```diff
@@ -108,10 +108,10 @@ Transport configuration options can be set in both the configuration file and in
...
 ### <%= transport %>
 
-| Option | Description | Default |
-| ------ | ----------- | ------- |
-<% options.each do |option, desc| -%>
-| `<%= option %>` | <%= desc %> | <%= @transports[:defaults][transport][option].to_s || "" %> |
+| Option | Description | Type | Default |
+| ------ | ----------- | ---- | ------- |
+<% options.each do |option, data| -%>
+| `<%= option %>` | <%= data[:desc] %> | <%= data[:type] %> | <%= @transports[:defaults][transport][option].to_s || "" %> |
```
beechtom: While I do like having the option's type listed, it does look a little weird when types like `Boolean` are missing.

lib/bolt/config/transport.rb
```diff
@@ -0,0 +1,133 @@
...
+          value = config[opt]
+          unless !!value == value || reference?(value)
+            raise Bolt::ValidationError,
+                  "#{opt} must be a Boolean true or false, received #{value.class} #{value.inspect}"
+          end
+        end
+      end
+
+      # Default validation? No validation!
+      private def validate; end
```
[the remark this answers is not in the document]
nicklewis: Hmmm what do we do about the case where the _entire_ value is a reference? We could add some logic around that to `validate` but then the transport-specific validation would still run... Currently all of those should work since they check whether the keys they're validating are present first, but some different validation in the future could fail. There's also the fact that we currently will merge the default config values _into_ the reference, which is bad.

Maybe we should just check in the `initialize` method whether the whole value is a reference and skip validating in that case?

There's a similar issue for transport-specific validation where individual values could be references and the transports don't check for that currently. I'm a bit unclear how to handle that though...

lib/bolt/inventory/inventory.rb
```diff
@@ -6,24 +6,31 @@
...
-        @config = config || Bolt::Config.default
-        @data = data || {}
-        @groups = Group.new(@data.merge('name' => 'all'), plugins)
-        @plugins = plugins
+      # TODO: Pass transport config instead of config object
+      def initialize(data, transport, transports, plugins)
+        @logger       = Logging.logger[self]
+        @data         = data || {}
+        @transport    = transport
+        @config       = transports
```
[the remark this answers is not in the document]
nicklewis: In that case, this could also just be passed as a single argument.

lib/bolt/config.rb
```diff
@@ -32,12 +24,16 @@ def initialize(transport, uri = nil)
...
-                  :apply_settings
-    attr_writer :modulepath
-    attr_reader :config_files, :warnings
+    attr_reader :config_files, :warnings, :data, :transports, :boltdir
+
+    TRANSPORT_CONFIG = {
+      'ssh'    => Bolt::Config::SSH,
+      'winrm'  => Bolt::Config::WinRM,
+      'pcp'    => Bolt::Config::Orch,
+      'local'  => Bolt::Util.windows? ? Bolt::Config::LocalWindows : Bolt::Config::Local,
```
nicklewis: Would it be better for the differences in behavior to be implemented on a single `Bolt::Config::Local` class? Maybe the class ought to override `self.options` to provide dynamic behavior at runtime rather than branching at load time. That would allow us to test both versions of it (by stubbing `Bolt::Util.windows?`) on a single platform.

lib/bolt/inventory/target.rb
nicklewis: That seems right to me. If we ever wanted higher-level transport-specific logic, it could live on the config class.

lib/bolt/config.rb
nicklewis: Ah I see that part of the complication there is that `Local` derives from `Sudoable` whereas `LocalWindows` doesn't. With my other suggestion to remove the `Sudoable` config class, that issue goes away.

We might also want to consider a similar refactor of the actual local transport as well, though not as part of this PR. That may be informed by our plan to separate "transport" from "shell" though.

lib/bolt/config.rb
```diff
@@ -210,169 +237,122 @@ def merge_config_data(config_data)
...
-      Bolt::Util.walk_keys(interpreters) do |key|
-        key.chars[0] == '.' ? key : '.' + key
+    # Recursively searches a data structure for plugin references
+    private def references?(input)
+      reference = false
+      if input.is_a?(Hash)
+        reference = input.key?('_plugin')
+        input.each_value { |v| reference ||= references?(v) }
+      elsif input.is_a?(Array)
+        input.each { |v| reference ||= references?(v) }
```
nicklewis:
```suggestion
    private def references?(input)
      if input.is_a?(Hash)
        input.key?('_plugin') || input.values.any? { |v| references?(v) }
      elsif input.is_a?(Array)
        input.any? { |v| references?(v) }
      else
        false
```

lib/bolt/config/transport.rb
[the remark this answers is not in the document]
beechtom: Updated to skip validation if the `input` to the config object is a reference. It still merges with the defaults, but since the reference will be resolved before any config options are accessed, that should be fine. Once the reference is resolved in `Bolt::Inventory` it will be remerged with the defaults and validated.

lib/bolt/config/transport.rb
```diff
@@ -0,0 +1,133 @@
...
+require 'bolt/util'
+
+module Bolt
+  class Config
+    class Transport
+      attr_reader :config
+
+      def initialize(data = {}, boltdir = nil)
+        assert_hash(data)
+        @config = Bolt::Util.deep_merge(defaults, filter(data))
```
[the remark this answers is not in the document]
beechtom: Updated to preserve the original input in `@input` and have `#merge` also accept objects of the same class. Merging will only merge `@input` and the data passed to `#merge` and then merge _that_ with the defaults and store it in `@config`.

I've left the`#to_h` method for now since I'm not sure how else we would retrieve the full config to pass to the applicator besides accessing individual options and building a hash with them e.g. `{ 'port' => config['port'], ... }`

lib/bolt/config.rb
beechtom: Great idea :) Updated to use a single `Bolt::Config::Transport::Local` class.

beechtom:
It looks like there may be some sort of error happening during the `puppet_agent::install` task. I'm not sure if it's transient/network-related or what, but the same commit was passing all tests earlier:
https://github.com/puppetlabs/bolt/actions/runs/50312993

And now it's failing:
https://github.com/puppetlabs/bolt/actions/runs/50347208

beechtom:
Failing integration tests will be fixed by https://github.com/puppetlabs/bolt/pull/1655

lib/bolt/config/transport.rb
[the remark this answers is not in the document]
nicklewis: There still remains the issue that some parameters can't be specified as plugin references because they will either be flagged as invalid or just cause an error. Since `assert_type` skips validating options whose values are references, the subsequent validation in each transport config class can be confronted with potentially invalid data.

For instance:

```yaml
ssh:
  private-key:
    _plugin: yaml
    filepath: key_data.yaml
```

This fails because it expects the value of `private-key` to be a hash containing a `key-data` key, which it isn't... yet.

This will affect:
* SSH: `private-key`, `interpreters` and `run-as-command`
* Docker: `interpreters`
* Local: `interpreters` and `run-as-command`
* Orch: `cacert` and `token-file`
* WinRM: `cacert` and `interpreters`

Admittedly, none of those seem like options you would especially _want_ to use a plugin for (except maybe `private-key`, but just set `key-data` instead of `private-key` and it will work), but it still ought to be accepted.

Given the explicit behavior of `super` here is to skip validation of anything that is a reference, I think it's not unreasonable to require the child class double-check the types of anything it wants to apply its own validation to.

lib/bolt/config/transport/base.rb
```diff
@@ -0,0 +1,125 @@
...
+          assert_hash_or_config(data)
+          @input   = reference?(data) ? data : filter(data)
+          @config  = Bolt::Util.deep_merge(defaults, @input)
+          @boltdir = boltdir
+          validate
+        end
+
+        def [](key)
+          @config[key]
+        end
```
nicklewis: Let's add `dig` as well, mostly for the sake of completeness. Should be simple:

```rb
def dig(*keys)
  @config.dig(*keys)
end
```

There are a few places in at least the tests where that would be helpful.

lib/bolt/config/transport.rb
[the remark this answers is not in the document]
beechtom: Updated to skip validation for each of these options if the value is a reference. Once the plugin references are resolved and the input is overwritten the transport config will be revalidated in full.

lib/bolt/config/transport/base.rb
```diff
@@ -56,6 +59,27 @@ def merge(*data)
           self.class.new(merged, @boltdir)
         end
 
+        # Resolve any references in the input data, then remerge it with the defaults
+        # and validate all values
+        def resolve(plugins)
+          unless plugins.is_a? Bolt::Plugin
```
nicklewis: I don't think this type check is necessary.

lib/bolt/config/transport/base.rb
```diff
@@ -11,39 +11,42 @@ class Base
...
           @config.fetch(*args)
         end
 
         def include?(args)
+          assert_resolved
           @config.include?(args)
         end
 
         def dig(*keys)
+          assert_resolved
```
nicklewis: Perhaps we could have a `resolved_config` method that asserts the data has been resolved+validated and then return `@config`. So this method, for instance, would just be:

```rb
def dig(*keys)
  resolved_config.dig(*keys)
end
```

We can also then be clear (by looking for `@config`) that we never use the unresolved config anywhere.

lib/bolt/config/transport/base.rb
```diff
@@ -11,39 +11,42 @@ class Base
 
         def initialize(data = {}, boltdir = nil)
           assert_hash_or_config(data)
-          @input   = reference?(data) ? data : filter(data)
-          @config  = Bolt::Util.deep_merge(defaults, @input)
-          @boltdir = boltdir
-          validate
+          @input    = data
+          @resolved = !Bolt::Util.references?(input)
+          @config   = Bolt::Util.deep_merge(defaults, filter(input))
```
nicklewis: Should we even do this merge if the input is unresolved? That would help ensure that it's impossible to _use_ unresolved data.

beechtom: Definitely went back and forth on whether or not to do that. Updated to default to `defaults` if the input is unresolved.

beechtom:
@puppetlabs/skeletor This PR changes how inventory is initialized, which affects the bolt server client script. Relevant changes are here: https://github.com/puppetlabs/bolt/pull/1644/commits/9a920759d411c16cdd8d810e95c2ba7df7e173eb

Issue #1535, opened 2020-01-15
Use Windows container environment instead of test host


lucywyman:
## Use Case
Instead of running tests that interact only with the test host, Windows testing should set up and use a container-based testing environment for mimicking remote hosts. This should be part of the GH actions testing workflow.

We should also figure out how to get Azure credentials, and document how to provision and run the tests in a Windows environment.

Depends on #1151

PR #1648, opened 2020-03-03
(GH-1535) Use container infrastructure for WinRM testing

lucywyman:
This modifies the infrastructure used to test WinRM connections to test
against a running container as opposed to connecting back to the virtual
machine itself. This changes how we provision the Github Actions
environment, now using docker-compose to bring up two Windows server
2019 containers - one with the Puppet Agent ruby taking precedence, and
one with Windows ruby taking precedence. The containers have the same
username and password as Linux container infrastructure, and connect
over winrm without SSL.

WinRM has 5 authentication methods by default, with the default non-SSL
authentication method being 'negotiate'. Negotiate determine whether to
use Kerberos or NTLM for authentication, preferring Kerberos. Previous
Bolt testing setups seem to have fallen back to using NTLM, or otherwise
been configured to allow user-password authentication (possibly through
Group Policies). However the default for the Windows Server 2019
container is to attempt Kerberos, which fails. As such the WinRM
connection must specify the `basic` auth method in order to use
user-pasword authentication between the the GH Action environment and
the containers. This is possible using the WinRM ruby gem, but not
something we want to expose to users. As such we specify the appropriate
settings when connecting to WinRM, wrapped in an environment variable
set when testing Bolt in CI.

### DIFF
--- a/.github/workflows/linux.yaml
+++ b/.github/workflows/linux.yaml
@@ -38,12 +38,10 @@ jobs:
         run: bundle install --jobs 4 --retry 3
       - name: Pre-test setup
         run: |
-          sudo curl -L https://github.com/docker/compose/releases/download/1.23.0/docker-compose-`uname -s`-`uname -m` -o /usr/local/bin/docker-compose
-          sudo chmod +x /usr/local/bin/docker-compose
           echo 'runner:runner' | sudo chpasswd
           sudo sh -c "echo 'Defaults authenticate' >> /etc/sudoers"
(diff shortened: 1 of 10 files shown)


.github/workflows/linux.yaml
```diff
@@ -38,12 +38,10 @@ jobs:
         run: bundle install --jobs 4 --retry 3
       - name: Pre-test setup
         run: |
-          sudo curl -L https://github.com/docker/compose/releases/download/1.23.0/docker-compose-`uname -s`-`uname -m` -o /usr/local/bin/docker-compose
```
lucywyman: Learned that docker and docker-compose are already installed in the GH action virtual envs.

PR #1649, opened 2020-03-03
(GH-1433) Only pass apply_settings task param when set

donoghuc:
Previously the apply_setting task parameter to the apply_catalog task would always be passed even if there were no apply_settings. This is incompatible with the puppetlabs-apply_helpers 0.1.0 module. Specifically if the apply_settings task parameter is passed to the task the task metadata forbids that parameter. The 0.2.0 module supports the apply_settings argument. With this commit we maintain compatibility with both versions.


donoghuc:
Users should just upgrade. Document this intead

PR #1656, opened 2020-03-06
(GH-1574) Initialize a Bolt project with modules

### LINKED ISSUE #1574: Get started fast with using any existing module
## Use Case

As a content consumer, I want a one-line command to run that lets me quickly and easily get started with using ANY module.

Today, using a module of any sophistication commonly requires pulling in other dependent modules. This means that to get started with using these kinds of modules, it is necessary to create a Bolt project, enumerate the target module AND its dependencies in a Puppetfile, then run `bolt puppetfile install`. You can't get started using your original target module until after you've done all of this preparatory work.

A project init command which allows you to specify a seed module to create a new project directory around.

```
bolt project init [--module=my-forgemodule] <project-name>
```

The command should create a new Bolt project directory <project-name>. For each `--module` specified, the new project should list that module in the Puppetfile along with a point-in-time evaluated set of other module dependencies, and perform work equivalent to `bolt puppetfile install`. When `bolt project init` completes, the project directory should be immediately usable. Subject to additional user-specified `bolt.yaml` or `inventory.yaml` configuration.

This command should leave an empty `bolt.yaml` in the directory so that bolt will detect it as a project.

#### What should the command do if a Puppetfile already exists? 
It should error and do nothing

#### What should the behavior be if a `bolt.yaml` exists? 
We should create a Puppetfile, and not interact with the `bolt.yaml`.

The command should not change existing artifacts, and if it is about to it should error.

beechtom:
This adds support for initializing a Bolt project with a set of
specified modules and their dependencies. A new CLI option `--modules`
accepts a comma-separated list of module names (e.g. `puppetlabs-apt`).

Bolt uses the `puppetfile-resolver` gem to parse the list of modules
provided at the CLI and then resolves a dependency graph for the
modules. If any of the specified modules or their dependencies are
missing, the command errors and the project directory is not created. If
the dependency graph is resolved successfully, the list of modules and
dependencies are written to a Puppetfile in the new project directory
and `bolt puppetfile install` is immediately run against this
Puppetfile.

### TODO

- [x] Update `puppet-runtime` with `puppetfile-resolver` and `monitillo` dependencies https://github.com/puppetlabs/puppet-runtime/pull/308

Closes #1574

### DIFF
--- a/lib/bolt/cli.rb
+++ b/lib/bolt/cli.rb
@@ -564,19 +564,116 @@ module Bolt
       0
     end
 
+    # Initializes a specified directory as a Bolt project and installs any modules
+    # specified by the user, along with their dependencies
     def initialize_project
-      path = File.expand_path(options[:object] || Dir.pwd)
-      FileUtils.mkdir_p(path)
-      ok = FileUtils.touch(File.join(path, 'bolt.yaml'))
+      boltdir    = Pathname.new(File.expand_path(options[:object] || Dir.pwd))
+      config     = boltdir + 'bolt.yaml'
+      puppetfile = boltdir + 'Puppetfile'
+      modulepath = [boltdir + 'modules']
 
-      result = if ok
-                 "Successfully created Bolt project directory at #{path}"
-               else
-                 "Could not create Bolt project directory at #{path}"
-               end
-      outputter.print_message result
+      # If modules were specified, first check if there is already a Puppetfile at the project
+      # directory, erroring if there is. If there is no Puppetfile, generate the Puppetfile
+      # content by resolving the specified modules and all their dependencies.
+      # We generate the Puppetfile first so that any errors in resolving modules and their
+      # dependencies are caught early and do not create a project directory.
+      if options[:modules]
+        if puppetfile.exist?
+          raise Bolt::CLIError,
+                "Found existing Puppetfile at #{puppetfile}, unable to initialize project with "\
+                "#{options[:modules].join(', ')}"
+        else
+          puppetfile_specs = resolve_puppetfile_specs
+        end
+      end
 
-      ok ? 0 : 1
+      # Warn the user if the project directory already exists. We don't error here since users
+      # might not have installed any modules yet.
+      if config.exist?
+        @logger.warn "Found existing project directory at #{boltdir}"
+      end
+
+      # Create the project directory
+      FileUtils.mkdir_p(boltdir)
+
+      # Bless the project directory as a boltdir
+      if FileUtils.touch(config)
+        outputter.print_message "Successfully created Bolt project at #{boltdir}"
+      else
+        raise Bolt::FileError.new("Could not create Bolt project directory at #{boltdir}", nil)
+      end
+
+      # Write the generated Puppetfile to the fancy new boltdir
+      if puppetfile_specs
+        File.write(puppetfile, puppetfile_specs.join("\n"))
+        outputter.print_message "Successfully created Puppetfile at #{puppetfile}"
+        # Install the modules from our shiny new Puppetfile
+        if install_puppetfile({}, puppetfile, modulepath)
+          outputter.print_message "Successfully installed #{options[:modules].join(', ')}"
+        else
+          raise Bolt::CLIError, "Could not install #{options[:modules].join(', ')}"
+        end
+      end
+
+      0
+    end
+
+    # Resolves Puppetfile specs from user-specified modules and dependencies resolved
+    # by the puppetfile-resolver gem.
+    def resolve_puppetfile_specs
+      require 'puppetfile-resolver'
+
+      # Build the document model from the module names, defaulting to the latest version of each module
+      model = PuppetfileResolver::Puppetfile::Document.new('')
+      options[:modules].each do |mod_name|
+        model.add_module(
+          PuppetfileResolver::Puppetfile::ForgeModule.new(mod_name).tap { |mod| mod.version = :latest }
+        )
+      end
+
+      # Make sure the Puppetfile model is valid
+      unless model.valid?
+        raise Bolt::ValidationError,
+              "Unable to resolve dependencies for #{options[:modules].join(', ')}"
+      end
+
+      # Create the resolver using the Puppetfile model. nil disables Puppet version restrictions.
+      resolver = PuppetfileResolver::Resolver.new(model, nil)
+
+      # Configure and resolve the dependency graph
+      result = resolver.resolve(
+        cache:                 nil,
+        ui:                    nil,
+        module_paths:          [],
+        allow_missing_modules: true
+      )
+
+      # Validate that the modules exist
+      missing_graph = result.specifications.select do |_name, spec|
+        spec.instance_of? PuppetfileResolver::Models::MissingModuleSpecification
+      end
+
+      if missing_graph.any?
+        titles = model.modules.each_with_object({}) do |mod, acc|
+          acc[mod.name] = mod.title
+        end
+
+        names = titles.values_at(*missing_graph.keys)
+        plural = names.count == 1 ? '' : 's'
+
+        raise Bolt::ValidationError,
+              "Unknown module name#{plural} #{names.join(', ')}"
+      end
+
+      # Filter the dependency graph to only include module specifications
+      spec_graph = result.specifications.select do |_name, spec|
+        spec.instance_of? PuppetfileResolver::Models::ModuleSpecification
+      end
+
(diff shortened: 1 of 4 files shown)


beechtom:
**Note:** This feature won't work until lingua-pupuli/puppetfile-resolver#8 is merged in and a new gem is released.

lib/bolt/cli.rb
```diff
@@ -564,19 +564,122 @@ def generate_types
...
+      require 'puppetfile-resolver'
+
+      # Convert module names to module specifications. The Puppetfile resolver requires
+      # that a version be specified, so we'll default to the latest.
+      modules = options[:modules].map do |mod|
+        "mod '#{mod}', :latest"
+      end
+
+      # Builds a model of the Puppetfile content that is used to build the dependency graph
+      model = PuppetfileResolver::Puppetfile::Parser::R10KEval.parse(modules.join("\n"))
```
glennsarti: you can just put the require in here `require 'puppetfile-resolver/puppetfile/parser/r10k_eval'`

This is the same as my docs and example CLI script.

spec/bolt/cli_spec.rb
```diff
@@ -2302,14 +2311,71 @@ def stub_config(file_content = {})
...
+        end
+      end
+
+      context 'with modules' do
+        it 'creates a Puppetfile and installs modules with dependencies' do
+          Dir.mktmpdir do |dir|
+            puppetfile = File.join(dir, 'Puppetfile')
+            modulepath = File.join(dir, 'modules')
+
+            cli = Bolt::CLI.new(%W[project init #{dir} --modules puppetlabs-apt])
```
glennsarti: If you're worried about test times, `puppetlabs-stdlib` is one of the worst modules due to the sheer number of releases, so it can take some time query the Forge.

lib/bolt/cli.rb
```diff
@@ -564,19 +564,122 @@ def generate_types
...
+
+      # Make sure the Puppetfile model is valid
+      unless model.valid?
+        plural = options[:modules].count == 1 ? '' : 's'
+        raise Bolt::ValidationError,
+              "Unable to resolve dependencies for module#{plural} #{options[:modules].join(', ')}"
+      end
```
glennsarti: @beechtom You don't really need any of the parsing stuff.  The parser is designed to take a text string and turn it into a document model.  However, you already know the module names and versions, so you can construct the model automatically.

e.g.

```
model = PuppetfileResolver::Puppetfile::Document.new('')
options[:modules].each do |mod_name|
  model.add_module(
    PuppetfileResolver::Puppetfile::ForgeModule.new(mod_name).tap { |mod| mod.version = :latest }
)
end
```

lib/bolt/cli.rb
beechtom: Ah, didn't spot that / realize the parser could be required separately. Thanks!

spec/bolt/cli_spec.rb
beechtom: Good to know. I'll switch that out with another one.

lib/bolt/cli.rb
glennsarti: It's one of the main reasons why I abstracted out the model, so you can do this kinda stuff.

beechtom: Fantastic, thanks! This is _much_ better than creating a string.

beechtom:
https://tickets.puppetlabs.com/browse/RE-13299

```
Molinillo is approved for use!
```

lib/bolt/cli.rb
```diff
@@ -564,19 +564,121 @@ def generate_types
...
+      # Bless the project directory as a boltdir
+      if FileUtils.touch(config)
+        outputter.print_message "Successfully created Bolt project at #{boltdir}"
+      else
+        raise Bolt::FileError.new("Could not create Bolt project directory at #{boltdir}", nil)
+      end
+
+      # Write the generated Puppetfile to the fancy new boltdir
+      if puppetfile_specs
+        if File.write(puppetfile, puppetfile_specs.join("\n"))
```
nicklewis: It looks like `File.write` returns an integer, so this will never be false. I expect that if it writing the file fails, it will raise an error. It's probably fine to just let that fail, since we've done some preliminary validation already (and so failing at this point is indeed exceptional).

lib/bolt/cli.rb
```diff
@@ -564,19 +564,121 @@ def generate_types
...
+    # Initializes a specified directory as a Bolt project and installs any modules
+    # specified by the user, along with their dependencies
     def initialize_project
-      path = File.expand_path(options[:object] || Dir.pwd)
-      FileUtils.mkdir_p(path)
-      ok = FileUtils.touch(File.join(path, 'bolt.yaml'))
+      boltdir    = Pathname.new(File.expand_path(options[:object] || Dir.pwd))
+      config     = boltdir + 'bolt.yaml'
+      puppetfile = boltdir + 'Puppetfile'
+      modulepath = [boltdir + 'modules']
```
nicklewis: What consideration do we need to give to the case where a user has a user-level or global `bolt.yaml` that specifies a different `modulepath`?

lib/bolt/cli.rb
```diff
@@ -564,19 +564,121 @@ def generate_types
...
+      plural     = options[:modules]&.count == 1 ? '' : 's'
```
nicklewis: I love to rephrase messages to avoid the need for pluralization. Is that possible in this case?

lib/bolt/cli.rb
```diff
@@ -564,19 +564,121 @@ def generate_types
...
+      missing_graph = result.specifications.select do |_name, spec|
+        spec.instance_of? PuppetfileResolver::Models::MissingModuleSpecification
+      end
+
+      if missing_graph.any?
+        titles = model.modules.each_with_object({}) do |mod, acc|
+          acc[mod.name] = mod.title
+        end
+
+        names  = titles.slice(*missing_graph.keys).values
```
nicklewis: I believe this is equivalent to `names = titles.values_at(*missing_graph.keys)`.

lib/bolt/cli.rb
```diff
@@ -564,19 +564,121 @@ def generate_types
...
+        plural = names.count == 1 ? '' : 's'
+
+        raise Bolt::ValidationError,
+              "Unknown module name#{plural} #{names.join(', ')}"
+      end
+
+      # Filter the dependency graph to only include module specifications
```
nicklewis: What else does it contain?

beechtom: Specifically a `PuppetfileResolver::Models::PuppetDependency` object: https://github.com/lingua-pupuli/puppetfile-resolver/blob/master/lib/puppetfile-resolver/models/puppet_dependency.rb

lib/bolt/cli.rb
beechtom: Removing most of the plurals. The only one I'm not sure how to reword is `Unknown module(s) ...`...

lib/bolt/cli.rb
beechtom: I think it makes the most sense to ignore system-wide and user-level config and have the `project init` command have no concept of the different config levels. 

It does raise an interesting question of what happens with users who have a global or user-level `modulepath` and expect to be able to use the modules in their new project... I think either warning the user that they might need to modify the modulepath in the project config or setting that in the new `bolt.yaml` would be helpful. I lean towards the former.

lib/bolt/cli.rb
glennsarti: Yup, Puppet Modules depend on other modules and Puppet versions.

lib/bolt/cli.rb
glennsarti: Not sure if this is an issue, but File.write will use CRLF line endings on Windows.  Not sure if you actually expect LF on all platforms?

glennsarti:
> This exhibits some strange but not necessarily unanticipated behavior when asked to install modules with conflicting dependencies. In that case, it will search backward to find compatible versions, which may result in quite old modules being installed.

Probably need to be more careful with wording here.  A conflicting dependency means it couldn't be resolved.  In the scenario you suggested, it's just old modules depending on other old modules.

> If we're worried about it, we could check whether we resolved to the latest version of the explicitly requested modules and warn if not. I don't think it's a big deal though.

You can warn but it's a tough UX decision.  What if the user needed that old version? what if an old version _is_ the latest?

beechtom:
Release note added to the end of the commit: https://github.com/puppetlabs/bolt/pull/1656/commits/c3c4af211324fdb7a8a37a6e84089f9d72005e12

PR #1658, opened 2020-03-06
Enable to use basic_auth_only option from winrm gem

knorx:
This pull request adds an option for the winrm transport named `basic-auth-only`, which is by default set to false, but if set to true will force basic auth by passing the corresponding `basic_auth_only` option to the winrm gem.
If there is an easier solution to my issue than to force basic auth, I would gladly welcome any advice.

I set up winRM to use with basicAuth on a machine at our site. I was able to connect properly using plain winRM powershell commands where I can pass the auth method via `-Authentication Basic`. For bolt this was not possible, the connection failed with an `AUTH_ERROR` all the time.
I looked into the code and saw that bolt offers no option to enforce basic_auth, while the winrm gem offers this option via `basic_auth_only`.

### DIFF
--- a/lib/bolt/transport/winrm.rb
+++ b/lib/bolt/transport/winrm.rb
@@ -63,7 +63,7 @@ module Bolt
         unless !!ssl_flag == ssl_flag
           raise Bolt::ValidationError, 'ssl option must be a Boolean true or false'
         end
-        
+
         basic_auth_only_flag = options['basic-auth-only']
         unless !!basic_auth_only_flag == basic_auth_only_flag
           raise Bolt::ValidationError, 'basic-auth-only option must be a Boolean true or false'
--- a/lib/bolt/transport/winrm.rb
+++ b/lib/bolt/transport/winrm.rb
@@ -7,48 +7,6 @@ require 'bolt/transport/powershell'
 module Bolt
   module Transport
     class WinRM < Base
-      OPTIONS = {
-        "cacert"          => "The path to the CA certificate.",
-        "connect-timeout" => "How long Bolt should wait when establishing connections.",
-        "extensions"      => "List of file extensions that are accepted for scripts or tasks. "\
-                              "Scripts with these file extensions rely on the target's file type "\
-                              "association to run. For example, if Python is installed on the system, "\
-                              "a `.py` script runs with `python.exe`. The extensions `.ps1`, `.rb`, and "\
-                              "`.pp` are always allowed and run via hard-coded executables.",
-        "file-protocol"   => "Which file transfer protocol to use. Either `winrm` or `smb`. Using `smb` is "\
-                              "recommended for large file transfers.",
-        "host"            => "Host name.",
-        "interpreters"    => "A map of an extension name to the absolute path of an executable, "\
-                              "enabling you to override the shebang defined in a task executable. The "\
-                              "extension can optionally be specified with the `.` character (`.py` and "\
-                              "`py` both map to a task executable `task.py`) and the extension is case "\
-                              "sensitive. When a target's name is `localhost`, Ruby tasks run with the "\
-                              "Bolt Ruby interpreter by default.",
-        "password"        => "Login password. **Required unless using Kerberos.**",
-        "port"            => "Connection port.",
-        "realm"           => "Kerberos realm (Active Directory domain) to authenticate against.",
-        "smb-port"        => "With file-protocol set to smb, this is the port to establish a connection on.",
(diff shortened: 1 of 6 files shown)


lib/bolt/transport/winrm.rb
```diff
@@ -31,7 +31,8 @@ class WinRM < Base
         "ssl"             => "When true, Bolt uses secure https connections for WinRM.",
         "ssl-verify"      => "When true, verifies the targets certificate matches the cacert.",
         "tmpdir"          => "The directory to upload and execute temporary files on the target.",
-        "user"            => "Login user. **Required unless using Kerberos.**"
+        "user"            => "Login user. **Required unless using Kerberos.**",
+        "basic-auth-only" => "Force basic authentication."
```
beechtom: `basic-auth-only` can be added to the `Bolt::Config::Transport::WinRM::OPTIONS` hash with the `type` set to `TrueClass`: https://github.com/puppetlabs/bolt/blob/master/lib/bolt/config/transport/winrm.rb#L10-L50

lib/bolt/transport/winrm.rb
```diff
@@ -62,6 +64,11 @@ def self.validate(options)
           raise Bolt::ValidationError, 'ssl option must be a Boolean true or false'
         end
 
+        basic_auth_only_flag = options['basic-auth-only']
+        unless !!basic_auth_only_flag == basic_auth_only_flag
+          raise Bolt::ValidationError, 'basic-auth-only option must be a Boolean true or false'
+        end
+
```
beechtom: Validation that the option is a Boolean value will be unnecessary since the config transport class automatically validates against a type set in the `OPTIONS` hash.

lib/bolt/transport/winrm.rb
```diff
@@ -43,7 +44,8 @@ def self.default_options
           'connect-timeout' => 10,
           'ssl' => true,
           'ssl-verify' => true,
-          'file-protocol' => 'winrm'
+          'file-protocol' => 'winrm',
+          'basic-auth-only' => false
```
beechtom: This can be added to the `Bolt::Config::Transport::WinRM::DEFAULTS` hash: https://github.com/puppetlabs/bolt/blob/master/lib/bolt/config/transport/winrm.rb#L52-L57

spec/bolt/inventory/inventory_spec.rb
```diff
@@ -629,6 +643,7 @@ def common_data(transport)
           expect(conf.transports[:ssh]['host-key-check']).to be nil
           expect(conf.transports[:winrm]['ssl']).to be true
           expect(conf.transports[:winrm]['ssl-verify']).to be true
+          expect(conf.transports[:winrm]['basic-auth-only']).to be false
```
beechtom:
```suggestion
          expect(conf.transports['winrm']['basic-auth-only']).to be false
```

lib/bolt/transport/winrm/connection.rb
```diff
@@ -50,6 +50,7 @@ def connect
                       retry_limit: 1,
                       transport: transport,
                       ca_trust_path: cacert,
+                      basic_auth_only: target.options['basic-auth-only'],
```
lucywyman: In order for this to use `basic` authentication we'll also need to set the `transport = :plaintext` when this option is specified https://github.com/WinRb/WinRM#plaintext

lucywyman:
@knorx If you're ok with me pushing to your branch I'd be happy to rebase this work and get it cleaned up for merge

lib/bolt/transport/winrm/connection.rb
knorx: I don't think this is right. I can use Basic Auth and do this using SSL as well. I configured our winrm server to only allow encrypted traffic. With the hacked bolt setting basic_auth_only to true, the connection works. So I would expect that it works as well with ssl transport. But I have only basic ruby knowledge and thus cannot deduce this from the code: https://github.com/WinRb/WinRM/blob/2a9a2ff55c5bbd903a019d63b1d134ac32ead4c7/lib/winrm/http/transport.rb#L263

lucywyman: Ah, ok, yeah. I was thinking this could be used for both SSL and non-SSL transports, but I'm fine with limiting it to use with SSL to avoid plaintext until someone asks for it.

beechtom:
Tests are failing since they set `basic-auth-only => true` and `ssl => false` further up in the options hash: https://github.com/knorx/bolt/blob/basic-auth-only/spec/bolt/inventory/inventory_spec.rb#L609-L626

PR #1659, opened 2020-03-06
(PE-28365) Make bolt-server compatible with bolt 2.0

donoghuc:
With the old inventory support dropped in bolt 2.0 we need to port bolt-server to the new inventory. This commit simply requires bolt/inventory in the transport app to take advantage of the Bolt::Inventory.empty class method. This also fixes up the debugging script and docs to run bolt-server for 2.0 compatibility.


donoghuc:
@smcclellan we need to not rely on this pattern https://github.com/puppetlabs/bolt/blob/e6a86058f0a11cdee3ef787199b7fddc9919e8e8/lib/bolt_server/transport_app.rb#L271-L272 anymore in bolt-server. I updated the test env so you can debug while you work on a fix.

donoghuc:
also sorry bolt team, i meant to raise a DRAFT>

donoghuc:
actually @smcclellan I think all we had to do was `require 'bolt/inventory'`

scripts/server_client.rb
```diff
@@ -27,7 +28,9 @@ class Client
   include BoltSpec::BoltServer
 
   def initialize
-    @inventory = Bolt::Inventory.new(conn_inventory.merge(easy_config))
+    config = Bolt::Config.default
```
lucywyman: The default for this is in the inventory init method, you should only have to initialize and pass in plugins.

donoghuc: @lucy I think that config is required for `Bolt::Plugin#setup`

donoghuc: @lucywyman ^^ (sorry other lucy :grimacing:  )

PR #1660, opened 2020-03-07
(GH-1555) Automatically generate changelog from commit notes

### LINKED ISSUE #1555: (SPIKE) Create directory for CHANGELOG entries
## Use Case
Currently, updating our CHANGELOG is a painful process. We need to add a new `## Bolt Next` heading, either while resolving the latest CHANGELOG before release or right after release. At best this results in regularly having to rebase PRs, at worst it results in features and bug fixes being listed under the wrong version number in the CHANGELOG.

We should try to create a Github action that uses information in the PR and identifies what the changelog message is and where it should exist in the changelog. The action will add it to the CHANGELOG when the PR is merged.

A first iteration of this could be writing a script that gets run manually before release that gets commits since the last tag, updates the changelog, the user manually adds the new version heading and pushes to master for Jenkins to tag and release.

Being a professional dog walker

beechtom:
This adds a rake task that executes a script to generate changelog
entries from commit messages. To execute the rake task, run `rake
'changelog[VERSION]'`, where `VERSION` is the new tag.

The script filters commits listed by the `git log` command and parses
them based on valid `!` labels. The list of valid labels are `feature`,
`bug`, `removal`, and `deprecation`. The label is followed by the
formatted changelog entry.

Closes #1555 

### Example

A commit message with a changelog entry would look like this:

```
<commit message>

!feature

* **Added a rake task to generate changelog entries** ([#1555](https://github.com/puppetlabs/bolt/issues/1555))

  Added a rake task that generates changelog entries from commit messages.
```

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -37,6 +37,19 @@ License Agreement before we can accept your pull request: https://cla.puppet.com
 1. Add the puppetlabs repo as an upstream - `git remote add upstream git@github.com:puppetlabs/bolt`
 1. Make a new branch off of master - `git checkout -b mybranchname`
 1. Commit your changes and add a useful commit message, including what specifically you changed and why - `git commit`
+    * If your changes are user-facing, add a release note to the end of a commit message. Release notes should begin
+      with a label indicating what kind of change you are making. Valid labels include `!feature`, `!bug`, `!deprecation`,
+      and `!removal`. 
+      
+      Release notes should follow this format:
+
+      ```
+      !label
+
+      * **Descriptive title of changes** ([#issue_number](issue_url))
+
+        Descriptive summary of changes.
+      ```
 1. Push your changes to your branch on your fork - `git push origin mybranchname`
 1. Open a PR against master at https://github.com/puppetlabs/bolt
 
--- a/developer-docs/releasing-bolt.md
+++ b/developer-docs/releasing-bolt.md
@@ -1,9 +1,9 @@
 # Releasing Bolt 
 
-Hello, fearless reader! This document details the release process for Bolt. Our current release cadence is weekly for both the bolt gem and bolt system package, usually on Wednesdays. Here's how we do it: 
+Hello, fearless reader! This document details the release process for Bolt. Our current release cadence is weekly for both the bolt gem and bolt system package, usually on Mondays. Here's how we do it: 
 
-1. Ensure the [CHANGELOG](https://github.com/puppetlabs/bolt/blob/master/CHANGELOG.md) has been updated to include all of the relevant changes included in the release under the new version header and has been merged in to the master branch.
+1. Generate the changelog using the rake task `rake 'changelog[VERSION]'`, where `VERSION` is the new tag, and merge the changes into `master`.
 2. Build and stage bolt packages with the [init_bolt-release job](https://jenkins-master-prod-1.delivery.puppetlabs.net/view/bolt/job/platform_bolt-vanagon_bolt-release-init_bolt-release/) using all defaults except the new tag for Bolt. The version specified will be the new tag in the github repo.
 3. Once step 2 is done and you are ready to push the packages to public repositories, run the [release job](https://jenkins-master-prod-1.delivery.puppetlabs.net/view/bolt/job/platform_ship-bolt_stage-foss-artifacts-all-repos/) with all default parameters except the new tag. This will also kick off the [docs publishing job](https://jenkins-master-prod-1.delivery.puppetlabs.net/view/bolt/job/platform_ship-bolt_publish_docs/) which will build and publish the docs based on the new tag. Note that this job can be run against any REF as a stand-alone pipeline in the case where a change is needed outside of a tagged version. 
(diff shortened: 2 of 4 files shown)


nicklewis:
Should this be a rake task instead of a standalone script?

beechtom:
Added a rake task.

developer-docs/releasing-bolt.md
```diff
@@ -1,9 +1,9 @@
 # Releasing Bolt 
 
-Hello, fearless reader! This document details the release process for Bolt. Our current release cadence is weekly for both the bolt gem and bolt system package, usually on Wednesdays. Here's how we do it: 
+Hello, fearless reader! This document details the release process for Bolt. Our current release cadence is weekly for both the bolt gem and bolt system package, usually on Mondays. Here's how we do it: 
 
-1. Ensure the [CHANGELOG](https://github.com/puppetlabs/bolt/blob/master/CHANGELOG.md) has been updated to include all of the relevant changes included in the release under the new version header and has been merged in to the master branch.
+1. Generate the changelog using the [`generate_changelog`](https://github.com/puppetlabs/bolt/blob/master/scripts/generate_changelog.rb) script. The changelog script accepts a single argument that is the new tag for Bolt `./scripts/generate_changelog.rb X.Y.Z`.
```
lucywyman: Pointing to the rake task might seem a bit more professional, since we have it.

CONTRIBUTING.md
```diff
@@ -37,6 +37,14 @@ License Agreement before we can accept your pull request: https://cla.puppet.com
 1. Add the puppetlabs repo as an upstream - `git remote add upstream git@github.com:puppetlabs/bolt`
 1. Make a new branch off of master - `git checkout -b mybranchname`
 1. Commit your changes and add a useful commit message, including what specifically you changed and why - `git commit`
+1. Add a release note with a label using `git notes add`. Valid labels are `@feature`, `@bug`, `@deprecation`, and `@removal`. Notes should be formatted like so:
+    ```
```
lucywyman: This may be out of scope for this PR, but could we add a template note that gets copied with `-C`? https://git-scm.com/docs/git-notes#Documentation/git-notes.txt--Cltobjectgt

scripts/generate_changelog.rb
```diff
@@ -0,0 +1,123 @@
...
+  exit 1
+end
+
+unless File.file?(changelog)
+  warn "Could not find changelog at #{changelog}"
+  exit 1
+end
+
+# Request the latest version of Bolt from GitHub
+uri    = URI('https://api.github.com/repos/puppetlabs/bolt/tags')
```
lucywyman: This just gets run once before putting up a pre-release PR, right? I think running `git tag` locally will work just as well, assuming whoever does the update has an up-to-date checkout, which I think they'd need to anyway.

nicklewis: Specifically I think we want `git describe --abbrev=0`

scripts/generate_changelog.rb
```diff
@@ -0,0 +1,123 @@
...
+# a valid LABEL (e.g. @feature, @bug)
+stdout_str, stderr_str, status = Open3.capture3(
+  "git log master...#{latest} "\
+  "--pretty=format:'#{FORMAT}' "\
+  "--grep='@(#{entries.keys.join('|')})' -E"
+)
+
+if status.success?
+  # Create an array of notes, tossing the first one which will always be blank.
+  # TODO: There's probably a better way to do this so we don't get a
```
nicklewis: Try `git log -z`. That will separate the entries with a `"\x00"` character instead of a newline, so you don't need to define your own marker (since a commit message won't have a null byte ordinarily).

nicklewis: You should also be able to get rid of the `%n` part of the pretty format

CONTRIBUTING.md
```diff
@@ -37,6 +37,19 @@ License Agreement before we can accept your pull request: https://cla.puppet.com
...
 1. Make a new branch off of master - `git checkout -b mybranchname`
 1. Commit your changes and add a useful commit message, including what specifically you changed and why - `git commit`
+    * If your changes are user-facing, add a release note to the end of a commit message. Release notes should begin
+      with a label indicating what kind of change you are making. Valid labels include `!feature`, `!bug`, `!deprecation`,
+      and `!removal`. 
+      
+      Release notes should follow this format:
+
+      ```
+      !label
```
lucywyman: I'd feel better about this if we had a workflow that checked for a release note, and failed if none was provided. `!no-release-note` would be valid for commits that don't require release.

scripts/generate_changelog.rb
```diff
@@ -0,0 +1,99 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+
+require 'open3'
+
+# Parses individual notes by searching for a !label, normalizing it to a
```
lucywyman:
```suggestion
# Parses individual commits by searching for a !label, normalizing it to a
```

CONTRIBUTING.md
beechtom: I think that gets tricky when you have a PR with more than one commit, as it would require that you add a label to _every_ commit message.

PR #1672, opened 2020-03-13
(maint) Include facts necessary for "warn" type

donoghuc:
The provider for the warn type in the `basic` module used for testing `apply` is confined to agents with the `another` fact. Bolt spec tests currently only test that an apply block containing the warn type will compile. If the apply block is actually attempted to be run on an agent the warn type will not actually raise the warning because the `another` fact is not bundled in the plugin targball that is shipped to the agent. This commit includes the fact so that the warn type would raise the warning if applied to an agent.


donoghuc:
As we look at adding acceptance tests in orch for apply in PE plans we will be looking to the basic module for test content https://tickets.puppetlabs.com/browse/PE-28597

Having this fact will allow us to test not only compilation but also application.

PR #1676, opened 2020-03-19
(GH-1671) Handle cases where loading paths with ~ fails

### LINKED ISSUE #1671: Don't assume users have a homedir
There are a number of paths in Bolt that assume the user running Bolt has a homedir, which isn't always the case when service accounts or other automation is running Bolt. This is especially problematic for directories that can't be configured by the user, such as the user-level config file `Pathname.new(File.expand_path(File.join('~', '.puppetlabs', 'etc', 'bolt', 'bolt.yaml')))`.

Additional info:
On Windows, Ruby tries HOME, HOMEDRIVE, and USERPROFILE from the environment in that order, before then trying to query the system for something home-y from the SHGetSpecialFolderLocation API (CSIDL_PROFILE or CSIDL_PERSONAL.) 

Questions:
* Should we validate that `~` is valid before trying to load and fall back to something else? Or should we just catch the `expand_path` error and not load that config?

lucywyman:
By default Bolt has 3 configuration file locations that get merged
together: project level, user level, and global. Project level config is
configurable (yo dawg...) using the `--configfile` CLI flag, but the 
user level and global configs are hard coded to
`~/.puppetlabs/etc/bolt/bolt.yaml` and `/etc/puppetlabs/bolt/bolt.yaml`
respectively. When `getlogin` returns nil loading the user level config
file will fail and error out trying to load `~`. This change catches the 
error and doesn't try to load user level config.

We also try to load a hard coded `~` path for analytics, at
`~/.puppetlabs/bolt/analytics.yaml`. Because loading the homedir will
typically only fail if getlogin isn't set when running Bolt through the 
Orchestrator or in other automated environments, we opt to disable
analytics if the config file can't be loaded.

This doesn't account for other places where paths use the homedir as
part of the default path since other paths are configurable, or already
fallback to another path if `ENV['HOME']` isn't set.

This additionally extends the connect timeout for WinRM apply
integration tests to 30 seconds, since they were consistently timing out 
in Github Actions.

Closes #1671

!bug

* **Handle cases where loading hardcoded homedir paths fail** ([#1671](https://github.com/puppetlabs/bolt/issues/1671))

  Bolt's user level config and analytics config paths are hardcoded and 
  include `~`, which errors out when getlogin fails to return a user. We
  now skip loading user level config if loading the file fails, and 
  disable analytics if loading the analytics config fails.

### DIFF
--- a/spec/integration/apply_spec.rb
+++ b/spec/integration/apply_spec.rb
@@ -431,6 +431,7 @@ describe "apply", expensive: true do
       { 'modulepath' => File.join(__dir__, '../fixtures/apply'),
         'winrm' => {
           'ssl' => false,
+          'connect-timeout' => 30,
           'ssl-verify' => false,
           'user' => conn_info('winrm')[:user],
           'password' => conn_info('winrm')[:password]
--- a/lib/bolt/analytics.rb
+++ b/lib/bolt/analytics.rb
@@ -28,8 +28,12 @@ module Bolt
 
     def self.build_client
       logger = Logging.logger[self]
-      config_file = File.expand_path('~/.puppetlabs/bolt/analytics.yaml')
-      config = load_config(config_file, logger)
+      begin
+        config_file = File.expand_path('~/.puppetlabs/bolt/analytics.yaml')
+        config = load_config(config_file, logger)
+      rescue ArgumentError
+        config = { 'disabled' => true }
+      end
 
       if config['disabled'] || ENV['BOLT_DISABLE_ANALYTICS']
         logger.debug "Analytics opt-out is set, analytics will be disabled"
--- a/lib/bolt/config.rb
+++ b/lib/bolt/config.rb
@@ -137,10 +137,15 @@ module Bolt
                     else
                       Pathname.new(File.join('/etc', 'puppetlabs', 'bolt', 'bolt.yaml'))
                     end
-      user_path = Pathname.new(File.expand_path(File.join('~', '.puppetlabs', 'etc', 'bolt', 'bolt.yaml')))
+      user_path = begin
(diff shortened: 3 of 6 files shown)


nicklewis:
What happens with PuppetDB config and orchestrator client config if you have no homedir? It looks like those reference `~` and `Dir.home` respectively.

nicklewis:
It also looks like we reference `~` when building the analytics client (since the analytics config lives there).

lucywyman:
I opted to leave in other paths that include `~`, namely:
* PuppetDB default token and config paths, `~/.puppetlabs/token` and `~/.puppetlabs/client-tools/puppetdb.conf`
* The default BoltDir `~/.puppetlabs/bolt`

Since these are configurable by the user. I'm open to setting fallback defaults for these

lucywyman:
@nicklewis The orch client already falls back to a `/etc` path https://github.com/puppetlabs/bolt/blob/master/lib/bolt/transport/orch.rb#L13 and the PuppetDB is configurable (though if we want to set a global default we can).

The analytics is trickier - should that fall back to `/etc/puppetlabs/bolt/analytics.yaml`?

nicklewis:
Oh I missed the key part where we already handle HOME being unset for PuppetDB/orchestrator 
client! :)

Maybe the fallback is to just disable analytics. That would actually potentially eliminate a lot of the automated users we see in our analytics, which could be a good thing or a bad thing...

spec/integration/apply_spec.rb
```diff
@@ -431,6 +431,7 @@ def config
       { 'modulepath' => File.join(__dir__, '../fixtures/apply'),
         'winrm' => {
           'ssl' => false,
+          'connect-timeout' => 30,
```
beechtom: 🎉

PR #1677, opened 2020-03-19
(GH-1403) Support Hiera lookups in plans

### LINKED ISSUE #1403: Support Hiera lookup() in plans outside of apply blocks
This is a duplication of a closed issue on the old ticket management:
https://tickets.puppetlabs.com/browse/BOLT-836

## Feature Request
It would be handy to support the hiera lookup() function within a plan.
## Open Question
What goes into the hierarchy for a plan? Without facter how are those values set? 
### Lookup different values to use the same plan in different environments or contexts

As a plan author i need to store data about how to execute the plan that may be environment or deployment specific. These settings would be looked up from the "perspective" of the bolt controller node. Example: may need different API keys, credentials, etc to use for API calls. Example 2: may be deploying in a different manner based on a setting in a config that is easy to switch.

### Use hiera to lookup task parameters for different targets

As a plan author i need to perform actions differently depending on certain facts and/or attributes of a node. These facts may be gather by the `facts` task. From these facts/attributes i then need to lookup() data to determine how the action should be run.  These facts would be looked up from the "perspective" of specific targets. Example: I'm deploying a software package against a variety of hosts, some may be 32-bit some 64-bit. I need to deploy different versions of software to the 32-bit systems. I would run a 'facts' task to determine architecture, then choose which software package to copy and install. The data hierarchy would be used to uniquely specify which software package to utilize.
## Learning from spike

* currently just exposing hiera allows backends to work without interpolation as long as all interpolation is based on $facts and $trusted. The fact that those are set during plan execution feels wrong and may be a bug.
* We probably need the ability for users to specify a separate hiearchy for the plan level so we don't limit interpolation options.
* There is value in exposing backends so that hiera-eyaml can be used but this problem may be solved by supporting eyaml in inventory.

# Questions for commentors
* What levels of hierarchy would you use at the plan level?
* Would there be any fact like interpolations in the hierarchy? What values would be used?
* Do you expect values from plan level data to be automatically set anywhere with data bindings like classes in manifests?

# The Plan
Users are interested in looking up hiera data outside of apply blocks in plans. There should be one context for the entirety of the plan, and it should always be empty. For now.
## Use Case
For now, this exists to give people the lookup function they seem to expect. Our goal is to learn more about how people want to use this functionality.
## Requirements
- [x] If the hierarchy references interpolations we should error
  **Note:** Requires changes in Puppet to remove `facts`, etc. from the `ScriptCompiler` scope. See https://github.com/beechtom/puppet/commit/6a8a7f5f0092e9f82c1669f2f2e54d5edb3c9439
- [x] We should use the config from the `hiera-config` option for now
- [x] You can use backends
- [x] We should collect analytics only when the plan-level function is called
  **Note:** Not sure if this is feasible currently. Wrapping the `lookup` function is difficult and there doesn't seem to be anything we can inject into Puppet.
- [ ] Document as experimental
  **Question:** Where is the best location for this documentation? 'Writing Plans'?

beechtom:
This adds support for the `lookup` plan function outside of apply
blocks.

When looking up values with Hiera, the `lookup` function will use the
Hiera config file specified by the user, defaulting to
`<boltdir>/hiera.yaml`. Any Hiera config file that includes
interpolations will cause the plan to error and return a helpful
message. Outside of the restrictions on interpolations, the
`lookup` function behaves similarly to how it would in apply block and
supports builtin and custom backends, default values and lambdas, and
merge strategies.

This also adds support for a `--hiera-config` option for the `bolt plan run`
command that accepts either an absolute or relative path to a Hiera
config file. The path is expanded relative to the boltdir.

Closes #1403 

!feature

* **Support `lookup` plan function outside of apply blocks**
([#1403](#1403))

  Plans can now use the `lookup` plan function outside of apply blocks
  to look up data with Hiera. The `lookup` function will use the Hiera
  config file specified in the Bolt config. Interpolations are not
  available outside of apply blocks and will cause a plan to error.

!feature

* **Add `--hiera-config` option for `bolt plan run` command**
([#1403](#1403))

  The `bolt plan run` command now supports a `--hiera-config` option
  that accepts an absolute or relative path to a Hiera config file.


Gemfile
```diff
@@ -14,6 +14,7 @@ gem "puma", '>= 3.12.0'
 gem "rack", '>= 2.0.5'
 gem "rails-auth", '>= 2.1.4'
 gem "sinatra", '>= 2.0.4'
+gem "puppet", github: 'beechtom/puppet', branch: 'GH-1403-hiera-lookup'
```
beechtom: #### TODO
Remove the commit that makes this change when/if the changes are made to pal.

lib/bolt/config.rb
```diff
@@ -334,7 +334,7 @@ def validate
         raise Bolt::ValidationError, "Unsupported format: '#{format}'"
       end
 
-      Bolt::Util.validate_file('hiera-config', @data['hiera_config']) if @data['hiera_config']
+      Bolt::Util.validate_file('hiera-config', @data['hiera-config']) if @data['hiera-config']
```
nicklewis: Does this imply we were previously never validating hiera-config?

beechtom: I believe so! Since this PR will be blocked for a while, it might be worth getting this in as a bug fix separately.

donoghuc:
Sorry for drive by review, i was just looking through how https://github.com/puppetlabs/puppet/pull/8063 might impact apply in PE plans and noticed that we wont need https://github.com/puppetlabs/bolt/blob/4a5a69296f07c37870638a6d1a47f378e179a2f1/lib/bolt/applicator.rb#L158 once that lands in puppet. 

Also there are already pkcs7 keys in fixtures https://github.com/puppetlabs/bolt/tree/master/spec/fixtures/keys

lib/bolt/pal.rb
```diff
@@ -145,6 +145,13 @@ def in_bolt_compiler
             end
           rescue Bolt::Error => e
             e
+          rescue Puppet::DataBinding::LookupError => e
```
lucywyman, line 148: It'd be nice if we also caught `Unknown variable $trusted / $server_facts / $settings::` here - while I highly doubt anyone is using them (and if they are they're surely misusing them) it'd be crazy-making to have a working plan suddenly start failing with no clue as to why.

beechtom: https://github.com/puppetlabs/bolt/pull/1677/files#diff-a0f737621da3fb6ead05b549796d7406R156-R159

Not really sure what the best phrasing for the message is, but that will catch any errors thrown by plans that were previously accessing those variables.

nicklewis:
Just need to change the puppet dependency / remove the Gemfile hack and this looks ready to go.

lib/bolt/pal.rb
```diff
@@ -145,8 +145,22 @@ def in_bolt_compiler
...
+            if e.issue_code == :HIERA_UNDEFINED_VARIABLE
+              message = "Interpolations are not supported in lookups outside of an apply block: #{e.message}"
+              PALError.new(message)
+            else
+              PALError.from_preformatted_error(e)
+            end
           rescue Puppet::PreformattedError => e
-            PALError.from_preformatted_error(e)
+            if e.issue_code == :UNKNOWN_VARIABLE &&
+               (match = /(?<var>facts|trusted|server_facts|settings)/.match(e.message))
```
nicklewis, line 157: This currently matches variables that contain these variable names as substrings. Even more confusingly, the substring it matches is the only part it prints in the error message.

```
plan test() {
  return $factses
}
```

```
{
  "kind": "bolt/pal-error",
  "msg": "Evaluation Error: Variable 'facts' is not available in the current scope unless explicitly defined. (file: /Users/nicklewis/projects/bolt/Boltdir/site/test/plans/init.pp, line: 2, column: 10)",
  "details": {
  }
}
```

It looks like the variable name may be available as `e.arguments[:name]`? I'm trying to trace the code paths to figure out _how_ that becomes the case, but it's... complex. But if that's reliable, then we can avoid using a regex entirely.

beechtom: That's helpful. Updated to use the argument from the error for now.

PR #1679, opened 2020-03-20
(GH-1267) Support project level Puppet content

### LINKED ISSUE #1267: Support project level plans
## Use Case
It's painful for new users to have to create three levels of directory before `site-modules/<module-name>/plans` before writing their first plan.

The project should implicitly be a module with it's own plans, tasks, manifests etc.  Users should think of projects as ready to use applications while modules are libraries meant to be called from a project. 

## Requirements
* For now users should opt into this experimental behavior by adding `project_name` to bolt.yaml.
   * `project_name` should not be merged from global config
   *  Project names must be valid module names. Existing core modules like `boltlib` and `out` should be considered reserved project names.
* When a project is named it should be loaded as a module that shadows any similarly named module on the module path.
* The boltdir directory name does not need to match the `project_name`. This is not a critical requirement for the first iteration.

lucywyman:
Users can now load puppet content from their Bolt project directory as
if it was a module. This adds the `name` configuration option to
`project.yaml` to specify a name for the project. If `project.yaml` is
present the project will be loaded as a module. The project is
namespaced either to the set `name` if present, or the name of the
project's directory. If the name conflicts with the boltlib modules the
project directory will not be loaded. This allows users to develop Bolt
content more easily by placing a `tasks` or `plans` directories directly
(ha) in their Bolt project and load.

Closes #1267

!feature

* **Experimental: Support project level Puppet content** ([#1267](#1267))

  Users can now load Puppet content from the root of the Bolt project directory, such as `<boltdir>/tasks`. Users must opt-in to this experimental feature by creating a `project.yaml` in their project directory.


documentation/directory_structure.md
```diff
@@ -56,6 +56,6 @@ To create a standalone module, run `pdk new module` outside of the project direc
 Follow these tips for managing standalone modules:
 
 -   Add `modules/*` to `.gitignore` of your project to prevent accidentally committing standalone modules.
--   When you run tasks and plans within a project directory, the modulepath (`modules/` and `site-modules/`) is searched for modules containing Bolt content. If a module is found in `modules`, tasks and plans from the version of the module in `site-modules` are ignored. Remove a module from `site-modules` if you convert it to a standalone module.
+-   When you run tasks and plans within a project directory the modulepath is searched in order for modules containing Bolt content. The Bolt project directory itself is loaded as a module, and the default modulepath is `<project directory>/modules:<project directory>/site-modules`. So if you have a module in both the `modules` and `site-modules` directories, the version in `modules` will actually be used.
```
beechtom:
```suggestion
-   When you run tasks and plans within a project directory the modulepath is searched in order for modules containing Bolt content. The Bolt project directory itself is loaded as a module, and the default modulepath is `<project directory>/modules:<project directory>/site-modules`. So if you have a module in both the `modules` and `site-modules` directories, the version in `modules` will be used.
```

lib/bolt/cli.rb
```diff
@@ -752,12 +752,14 @@ def install_puppetfile(config, puppetfile, modulepath)
     end
 
     def pal
+      proj_path = config.project_name ? config.boltdir.path.to_s : nil
```
beechtom: I think it might be better to have this logic in `Bolt::Config`, where you can call `config.project_path` and it returns either `@boltdir.path.to_s` if `project_name` is set or `nil`.

lib/bolt/config.rb
```diff
@@ -168,6 +182,7 @@ def initialize(boltdir, config_data, overrides = {})
         'log'                 => { 'console' => {} },
         'plugin_hooks'        => {},
         'plugins'             => {},
+        'project_name'        => false,
```
beechtom: I don't think this is strictly necessary, since `#project_name` will return `nil` if the setting was never set.

spec/fixtures/project_dir/bolt.yaml
```diff
@@ -0,0 +1,2 @@
+modulepath: modules:../modules
+project_name: true
```
nicklewis: I assumed that the value of `project_name` would be the name to use for the project.

lucywyman: Yeah I misunderstood the original issue - I'm reworking the Puppet side now and will update shortly.

lib/bolt/pal.rb
```diff
@@ -52,12 +52,22 @@ def initialize(modulepath, hiera_config, resource_types, max_compiles = Etc.npro
...
       @resource_types = resource_types
+      @project = project
 
       @logger = Logging.logger[self]
       if modulepath && !modulepath.empty?
         @logger.info("Loading modules from #{@modulepath.join(File::PATH_SEPARATOR)}")
       end
 
+      # If the project name is the same as one of the builtin modules raise a warning
+      bolt_mods = Dir.entries(BOLTLIB_PATH)
```
beechtom:
```suggestion
      bolt_mods = Dir.children(BOLTLIB_PATH)
```

`entries` includes `.` and `..` while `children` does not.

lib/bolt/config.rb
```diff
@@ -50,6 +50,9 @@ class Config
                                     "OS-specific PATH separator.",
       "plugin_hooks"             => "Which plugins a specific hook should use.",
       "plugins"                  => "A map of plugins and their configuration data.",
+      "project_name"             => "The namespace for project artifacts. Setting this also opts the user into "\
```
beechtom: Any chance the option could be `project-name` instead? Most of our config options use hyphens and I think it would be nice to maintain consistency as much as we can.

documentation/directory_structure.md
```diff
@@ -2,7 +2,7 @@
 
 Puppet tasks, plans, functions, classes and types must exist inside a Puppet module in order for Bolt to load them. Bolt loads modules by searching for module directories on the modulepath.
 
-By default, the `modulepath` includes the `modules/` and `site-modules` directories in the [Bolt project directory](bolt_project_directories.md#).
+By default, the modulepath includes the `modules/` and `site-modules` directories in the [Bolt project directory](bolt_project_directories.md#). The Bolt project directory itself is also loaded as a module, namespaced to the name of the directory.
```
beechtom:
```suggestion
By default, the modulepath includes the `modules/` and `site-modules/` directories in the [Bolt project directory](bolt_project_directories.md#). The Bolt project directory itself is also loaded as a module, namespaced to the name specified in the `bolt.yaml`.
```

documentation/bolt_running_tasks.md
```diff
@@ -68,8 +68,17 @@ bolt task run mymodule::mytask --targets app1.myorg.com --params $(@{load_balanc
 bolt plan run mymodule::myplan --targets app1.myorg.com --params $(@{load_balancers=@("lb1.myorg.com","lb2.myorg.com")} | ConvertTo-Json)
 ```
 
-## Specifying the module path
+## Specifying the modulepath
 
-In order for Bolt to find a task or plan, the task or plan must be in a module on the `modulepath`. By default, the `modulepath` includes `modules/` and `site-modules/` directories inside the Bolt project directory.
+In order for Bolt to find a task or plan, the task or plan must be in a module on the modulepath.
+The Bolt project directory (Boltdir) is considered a module on the modulepath, namespaced to the name of the
+directory. The default modulepath is `<Boltdir>/modules/:<Boltdir>/site-modules/`.
```
beechtom:
```suggestion
The Bolt project directory (Boltdir) is considered a module on the modulepath, namespaced to the name 
specified in the `bolt.yaml`. The default modulepath is `<Boltdir>/modules/:<Boltdir>/site-modules/`.
```

documentation/bolt_running_plans.md
```diff
@@ -68,9 +68,17 @@ bolt task run mymodule::mytask --targets app1.myorg.com --params $(@{load_balanc
 bolt plan run mymodule::myplan --targets app1.myorg.com --params $(@{load_balancers=@("lb1.myorg.com","lb2.myorg.com")} | ConvertTo-Json)
 ```
 
-## Specifying the module path
+## Specifying the modulepath
 
-In order for Bolt to find a task or plan, the task or plan must be in a module on the `modulepath`. By default, the `modulepath` includes `modules/` and `site-modules/` directories inside the Bolt project directory.
+In order for Bolt to find a task or plan, the task or plan must be in a module on the modulepath.
+The Bolt project directory (Boltdir) is considered a module on the modulepath, namespaced to the name of the
+directory. The default modulepath is `<Boltdir>/modules/:<Boltdir>/site-modules/`.
```
beechtom:
```suggestion
The Bolt project directory (Boltdir) is considered a module on the modulepath, namespaced to the name specified 
in the `bolt.yaml`. The default modulepath is `<Boltdir>/modules/:<Boltdir>/site-modules/`.
```

Not sure what the best phrasing would be here...

beechtom:
```
Traceback (most recent call last):
	...
/Users/thomasbeech/.rvm/gems/ruby-2.6.3/bundler/gems/puppet-d9f660ed81b5/lib/puppet/pops/loaders.rb:257:in `add_loader_by_name': Internal Error: Attempt to redefine loader named 'facts' (Puppet::ParseError)
```

It looks like the project does not shadow existing non-core modules since the project loader gets loaded first and then the module is loaded and attempts to replace the project loader, which raises an error.

Module loaders will get configured here: https://github.com/puppetlabs/puppet/blob/master/lib/puppet/pops/loaders.rb#L419
And will eventually error when creating the loader and adding it to the loaders here:
https://github.com/puppetlabs/puppet/blob/master/lib/puppet/pops/loader/module_loaders.rb#L141

hestonhoffman:
@lucywyman I might just open a separate PR for the docs once this is merged in.

lib/bolt/config.rb
lucywyman: We haven't been very consistent about it, but we tend to use underscores for config that can't be set with CLI flags (`apply_settings`, `plugin_hooks`, etc.) and hyphens for config that can be set on the CLI. I think the question is more whether we expect project-name to ever be configurable on the CLI.

beechtom:
This correctly shadows existing module content but doesn't shadow an entire module. I'm not sure which behavior we want, though shadowing the entire module seems like it would be less confusing?

For example, naming the project `facts` results in the following list of plans. `facts` is shadowed by `plans/init.pp` and `facts::test` is a new plan, but `facts::info` at `module/facts/plans/info.pp`is still visible and runnable :
```
...
canary
facts
facts::info
facts::test
haproxy::install
...
```

It looks like the issue is the `private_environment_loader` loads the module data and adds it to its `@module_resolver`, so it still sees all of the content for the shadowed module. I think skipping the step where the module data is added to the resolver when the module name matches the project name should get around this, though I'm not sure if that would have any bigger implications.

https://github.com/puppetlabs/puppet/blob/master/lib/puppet/pops/loaders.rb#L421-L426

beechtom:
A few failing tests, but otherwise LGTM. Modules are being shadowed correctly and project-level content is working. This is awesome! 👍

nicklewis:
Would it make sense (not necessarily as part of this PR) to rename `Bolt::Boltdir` to `Bolt::Project` and unify the two concepts?

beechtom:
I like that idea @nicklewis  👍 I think `boltdir` can sometimes be a little confusing since we use it to refer to both the project directory as well as the actual `Boltdir` directory.

documentation/bolt_running_plans.md
```diff
@@ -70,7 +70,15 @@ bolt plan run mymodule::myplan --targets app1.myorg.com --params $(@{load_balanc
 
 ## Specifying the module path
```
hestonhoffman, line 71:
```suggestion
## Specifying the modulepath
```

documentation/bolt_running_plans.md
```diff
@@ -70,7 +70,15 @@ bolt plan run mymodule::myplan --targets app1.myorg.com --params $(@{load_balanc
...
 
-In order for Bolt to find a task or plan, the task or plan must be in a module on the `modulepath`. By default, the `modulepath` includes `modules/` and `site-modules/` directories inside the Bolt project directory.
+In order for Bolt to find a task or plan, the task or plan must be in a module on the modulepath.
+The [Bolt project directory](#bolt_project_directories.md) is also loaded as a module, namespaced to the `project-name`
+specified in `bolt.yaml`. The default modulepath is `<Boltdir>/modules/:<Boltdir>/site-modules/`.
 
-If you are developing a new plan, you can specify `--modulepath <PARENT_DIR_OF/MODULE>` to tell Bolt where to load the module. For example, if your module is in `~/src/modules/my_module/`, run Bolt with `--modulepath ~/src/module`. If you often use the same `modulepath`, you can set `modulepath` in `bolt.yaml`.
+If you are developing a new task or plan you can create a `<Boltdir>/tasks/` or `<Boltdir>/plans/`
+directory to develop your content in, then run Bolt from the root of your Bolt project directory to
+test the task or plan.
```
hestonhoffman, line 79: Does `<Boltdir>` here signify the name of the project directory? If so, I think it would cut down on a bit of confusion if we used something like `<PROJECT_NAME>` throughout instead. Unless I'm missing something, and this only applies to embedded project directories? If that's the case, let's remove the `<>`.

documentation/directory_structure.md
```diff
@@ -2,7 +2,7 @@
 
 Puppet tasks, plans, functions, classes and types must exist inside a Puppet module in order for Bolt to load them. Bolt loads modules by searching for module directories on the modulepath.
 
-By default, the `modulepath` includes the `modules/` and `site-modules` directories in the [Bolt project directory](bolt_project_directories.md#).
+By default, the modulepath includes the `modules/` and `site-modules` directories in the [Bolt project directory](bolt_project_directories.md#). The Bolt project directory itself is also loaded as a module, namespaced to the `project-name` specified in `bolt.yaml`.
```
hestonhoffman, line 5:
```suggestion
By default, the modulepath includes the `modules` and `site-modules` directories in the [Bolt project directory](bolt_project_directories.md#). The Bolt project directory itself is also loaded as a module, namespaced to the `project-name` specified in `bolt.yaml`.
```

documentation/directory_structure.md
```diff
@@ -37,13 +37,12 @@ A typical module for use with Bolt may contain these directories:
 
 ### Where to put module code
 
-Modules can either be written directly in `site-modules/` or be installed from the Puppet Forge or a code repository into `modules/`.
+Modules can be developed directly in the Bolt project directory or be installed from the Puppet
+Forge into `modules/`.
```
hestonhoffman, line 41:
```suggestion
Forge into the `modules` directory.
```

documentation/directory_structure.md
```diff
@@ -37,13 +37,12 @@ A typical module for use with Bolt may contain these directories:
...
 
 ## Modules for projects
 
-Modules developed to support a particular project can be developed directly in the `site-modules` directory of the Bolt project directory.
-
-Create a new directory for each module inside `site-modules` that matches the modules name or use `pdk new module` to create a skeleton structure.
+Modules developed to support a particular project can be developed directly in the Bolt project directory. You can use `pdk new module` to create a skeleton structure, and `bolt project init` inside a directory to make it a Bolt project directory.
```
hestonhoffman, line 45:
```suggestion
Modules developed to support a particular project can be developed directly in the Bolt project directory. You can use `pdk new module` to create a skeleton structure, and `bolt project init` inside a directory to make it a Bolt project directory.

**Note**:  To use the `pdk` command, you must [install the Puppet Development Kit](https://puppet.com/docs/pdk/1.x/pdk_install.html) 
```

documentation/directory_structure.md
```diff
@@ -37,13 +37,12 @@ A typical module for use with Bolt may contain these directories:
 
 ### Where to put module code
 
-Modules can either be written directly in `site-modules/` or be installed from the Puppet Forge or a code repository into `modules/`.
+Modules can be developed directly in the Bolt project directory or be installed from the Puppet
```
hestonhoffman, line 40:
```suggestion
You can develop modules directly in the Bolt project directory or install them from the Puppet
```

lucywyman:
Ok, the failure now is only because Puppet hasn't released yet. This should be good to go. I updated the opt-in condition to be if `project.yaml` exists, and for `name` config setting to be used if present otherwise use the directory name as the project namespace.

lib/bolt/pal.rb
```diff
@@ -52,12 +52,21 @@ def initialize(modulepath, hiera_config, resource_types, max_compiles = Etc.npro
...
+      @project = project
 
       @logger = Logging.logger[self]
       if modulepath && !modulepath.empty?
         @logger.info("Loading modules from #{@modulepath.join(File::PATH_SEPARATOR)}")
       end
 
+      # If the project name is the same as one of the builtin modules raise a warning
+      bolt_mods = Dir.children(BOLTLIB_PATH)
+      if @project && bolt_mods.include?(@project&.name)
```
beechtom, line 64:
```suggestion
      if @project && bolt_mods.include?(@project.name)
```

With the change to `Bolt::Project#name`, this will always return a value and also pick up directory names that conflict with core modules.

lib/bolt/project.rb
```diff
@@ -45,20 +46,34 @@ def initialize(path, type = 'option')
...
       path == other.path
     end
     alias == eql?
 
+    def file?
+      File.exist?(@project_file)
+    end
+
+    def name
+      @data['name']
```
beechtom, line 74:
```suggestion
      @data['name'] || File.basename(@path)
```

Puppet is expecting a project to have a name: https://github.com/puppetlabs/puppet/blob/master/lib/puppet/node/environment.rb#L302-L303

If `name` isn't set in `project.yaml`, then it should pick up the name of the directory itself. Otherwise Puppet sees the project having a name of `nil`, which isn't a valid module:

```
        Invalid module name ''; module names must match either:
        An installed module name (ex. modulename) matching the expression /^[a-z][a-z0-9_]*$/ -or-
        A namespaced module name (ex. author-modulename) matching the expression /^[a-zA-Z0-9]+[-][a-z][a-z0-9_]*$/
```

lib/bolt/project.rb
```diff
@@ -45,20 +46,34 @@ def initialize(path, type = 'option')
...
+      { path: @path, name: name }
+    end
+
     def eql?(other)
       path == other.path
     end
     alias == eql?
 
+    def file?
+      File.exist?(@project_file)
```
nicklewis, line 70: Having renamed `boltdir` to `project` muddles this behavior a bit. You now _always_ have a project, but it may or may not be "a project" depending on whether it has `project.yaml`.

Maybe the logic then is that there is _always_ a project, but it simply may or may not be loaded as a module. In that case, we could rename this method to something that expresses exactly what we mean, ie. `load_as_module?`

lib/bolt/pal.rb
```diff
@@ -52,12 +52,21 @@ def initialize(modulepath, hiera_config, resource_types, max_compiles = Etc.npro
...
 
       @logger = Logging.logger[self]
       if modulepath && !modulepath.empty?
         @logger.info("Loading modules from #{@modulepath.join(File::PATH_SEPARATOR)}")
       end
 
+      # If the project name is the same as one of the builtin modules raise a warning
+      bolt_mods = Dir.children(BOLTLIB_PATH)
+      if @project && bolt_mods.include?(@project.name)
+        msg = "The project #{@project.name} will not load top-level content as it has a reserved name: #{bolt_mods}"
```
nicklewis, line 65: This message comes out a little weird.

```
The project boltlib will not load top-level content as it has a reserved name: ["file", "out", "ctrl", "system", "boltlib", "prompt"]
```

Not quite enough context is presented for _why_ that list of names is there.

lib/bolt/project.rb
```diff
@@ -45,20 +46,36 @@ def initialize(path, type = 'option')
...
     end
     alias == eql?
 
+    def load_as_module?
+      @project_file.file?
+    end
+
+    def name
+      # If the project is in mymod/Boltdir/project.yaml, use mymod as the modulename
+      namespace = type == 'embedded' ? @path.split[0].basename : @path.basename
```
nicklewis, line 75:
```suggestion
      namespace = type == 'embedded' ? @path.parent.basename : @path.basename
```

lib/bolt/pal.rb
```diff
@@ -136,33 +145,35 @@ def in_bolt_compiler
...
-              PALError.from_preformatted_error(e)
-            end
-          rescue Puppet::PreformattedError => e
-            if e.issue_code == :UNKNOWN_VARIABLE &&
-               %w[facts trusted server_facts settings].include?(e.arguments[:name])
-              message = "Evaluation Error: Variable '#{e.arguments[:name]}' is not available in the current scope "\
-                        "unless explicitly defined. (file: #{e.file}, line: #{e.line}, column: #{e.pos})"
-              PALError.new(message)
-            else
+        Puppet.override(bolt_project: @project) do
```
nicklewis, line 148: We should be able to include `yaml_plan_instantiator` in this same `override` call so we can avoid a level of nesting.

lib/bolt/pal.rb
lucywyman: Boltlib is especially weird because I can even say "builtin plan modules" since we alias boltlib functions and plans to nix the 'boltlib' prefix. Would just `"The project boltlib will not load top-level content as it conflicts with builtin Bolt modules: ["file", "out", "ctrl", "system", "boltlib", "prompt"]"` be better?

nicklewis: I guess I'm not sure it helps much to include all the options. That would be like... the natural name for your project conflicts with a built-in name, but make sure you don't accidentally pick one of these _other_ (very unrelated) built-in names when you change it.

"Project-level content will not be loaded because the  project name 'boltlib' conflicts with a built-in module of the same name" perhaps?

nicklewis: I'm also trying to think through cases where it makes sense to have a project whose name conflicts with a built-in module even though the project content won't be loaded, ie. why should this be a warning and not just an error. I guess it's mainly if you have a project without `name` set in project.yaml, but that happens to be in a directory that happens to be the same as a built-in module (unlikely as that is) and therefore the default name would conflict.

Other than that, the whole point of giving a project a name is to load its top-level content. You might not care about loading the top-level content if you're just developing the project to ship it as a module, but in that case it won't work as a module _either_ (since it will be shadowed by the built-in module of the same name).

nicklewis:
Hmm a couple more things about project names...

If the name is set explicitly in `project.yaml`, we should validate that it's a legal module name and fail if not.

If we're auto-detecting it from the directory name and the directory isn't a valid module name, we should just skip using the project. Possibly with a warning, but that might be more annoying to the user than it's worth. I expect that if you have your project in a directory that isn't a legal module name, you probably have no expectation that you can use it as a module and so we don't need to tell you that it won't work.

Last question is whether we want to properly auto-detect the project name from directory names like `puppetlabs-apache` (by splitting on `-` and taking the second half). I think that since we _are_ doing auto-detection, that would make sense.

lucywyman:
@nicklewis This is the current behavior for an invalidly named project directory, which I think is acceptable:
```
lucy ~/githubs/bolt (GH-1267) $ be bolt task show --boltdir ___/
        Invalid module name '___'; module names must match either:
        An installed module name (ex. modulename) matching the expression /^[a-z][a-z0-9_]*$/ -or-
        A namespaced module name (ex. author-modulename) matching the expression /^[a-zA-Z0-9]+[-][a-z][a-z0-9_]*$/
facts                                 Gather system facts
gcloud_inventory::resolve_reference   Generate targets from Google Cloud compute engine instances
package                               Manage and inspect the state of packages
puppet_agent::install                 Install the Puppet agent package
puppet_agent::version                 Get the version of the Puppet agent package installed. Returns nothing if none present.
puppet_conf                           Inspect puppet agent configuration settings
reboot                                Reboots a machine
reboot::last_boot_time                Gets the last boot time of a Linux or Windows system
service                               Manage and inspect the state of services
terraform::apply                      Apply an HCL manifest
terraform::destroy                    Destroy resources managed with Terraform
terraform::initialize                 Initialize a Terraform project directory
terraform::output                     JSON representation of Terraform outputs

MODULEPATH:
/home/lucy/githubs/bolt/___/modules:/home/lucy/githubs/bolt/___/site-modules:/home/lucy/githubs/bolt/___/site

Use `bolt task show <task-name>` to view details and parameters for a specific task.
```
For namespaced modules I think it's ok to get this in and do that separately, so we can get feedback on this at demos this week?

lib/bolt/project.rb
```diff
@@ -68,6 +87,15 @@ def plans
     end
 
     def validate
+      n = @data['name']
+      if n && !Puppet::Module.is_module_directory_name?(n) && !Puppet::Module.is_module_namespaced_name?(n)
+        raise Bolt::ValidationError, <<~ERROR_STRING
+        Invalid module name '#{n}'; module names must match either:
```
nicklewis, line 93: This error message isn't quite right for this use case. It shouldn't refer to "module" anywhere ("project" instead), and shouldn't refer to "installed" anything.

lib/bolt/pal.rb
lucywyman: This makes sense, I moved the validation to be in the project validation.

lib/bolt/project.rb
```diff
@@ -1,11 +1,13 @@
...
 require 'pathname'
+require 'bolt/pal'
 
 module Bolt
   class Project
     BOLTDIR_NAME = 'Boltdir'
     PROJECT_SETTINGS = {
+      "name"  => "The name of the project",
       "plans" => "An array of plan names that can be included in `bolt plan show` output",
       "tasks" => "An array of task names that can be included in `bolt task show` output"
```
hestonhoffman, line 12: I think we should rephrase slightly so that it's clear to users that these options whitelist plan and task names
```suggestion
      "plans" => "An array of plan names to whitelist. Whitelisted plans are included in `bolt plan show` output",
      "tasks" => "An array of task names to whitelist. Whitelisted tasks are included in `bolt task show` output"
```

lucywyman: Updateddddd

PR #1683, opened 2020-03-23
Add new 3rd party plugin list

logicminds:
* This adds a yaml file with a list of plugins that generates a
 markdown file of plugins via the `plugin_list` rake task.

 * The markdown is generated dynamically via data from the plugins.yaml
 file.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -13,7 +13,7 @@ AllCops:
 # modifier if/unless.
 Style/HashEachMethods:
   Enabled: true
-
+  
 Style/HashTransformKeys:
   Enabled: true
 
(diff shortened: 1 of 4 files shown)


logicminds:
The YAML file is required to have further automation of plugin lists. I wanted to create a bolt command to query this YAML file to see a list of plugins. ie. `bolt plugins show --available ` Having this data just in markdown format won't do.  Since it is YAML we can convert to markdown and other formats when needed.   As far as complexity, this is hardly complex.   Slurp data, format and output. 

The YAML is pretty straightforward and I would much rather fill this out than the markdown syntax. 

Moving resources is out of scope of this PR.  I didn't create the directory and it was suggested I put this data here.

adreyer:
I don't think that yaml file will be the source of truth for long. Bolt plugins will be searchable on forge eventually. Having dealt with the maintenance burden of `puppet module search` I don't really think that sort of `--available` functionality is worth it. In practice you have to go to forge or github to read the documentation regardless.

PR #1687, opened 2020-03-25
Fixed documentation for using the puppetdb plugin with certname fact

nmaludy:
When trying to implement some more complex inventory files using PuppetDB i ran into an issue when trying to use the `target_mapping` feature of the `puppetdb` plugin based on this example: https://puppet.com/docs/bolt/latest/using_plugins.html#example-usage-2

Running this as my inventory as shown in the example:
```yaml
version: 2
groups:
  - name: puppetdb_wsus
    targets:
      - _plugin: puppetdb
        query: "inventory[certname] { facts.osfamily = 'windows' }"
        target_mapping:
          name: certname
```

I kept getting an error: 
```
$ bolt inventory show --targets puppetdb_wsus
Could not find fact certname for node xxx
Could not find fact certname for node yyy
Could not find fact certname for node zzz
```

I dug into it a bit by querying the `fact-path` API in PuppetDB:
```bash
$ curl -s -k -X GET     --cacert /etc/puppetlabs/puppet/ssl/certs/ca.pem     --key /etc/puppetlabs/puppet/ssl/private_keys/$(hostname -f).pem     --cert /etc/puppetlabs/puppet/ssl/certs/$(hostname -f).pem 'https://127.0.0.1:8081/pdb/query/v4/fact-paths' | jq | grep -B4 -A3 certname
  {
    "type": "string",
    "path": [
      "trusted",
      "certname"
    ],
    "name": "trusted"
  },
```

Turns out that the path is actually `"trusted.certname"`.

My final inventory file that got the the results i wanted without errors was:

```yaml
version: 2
groups:
  - name: puppetdb_wsus
    targets:
      - _plugin: puppetdb
        query: "inventory[certname] { facts.osfamily = 'windows' }"
        target_mapping:
          name: facts.trusted.certname
```

### DIFF
--- a/documentation/using_plugins.md
+++ b/documentation/using_plugins.md
@@ -125,7 +125,7 @@ targets:
   - _plugin: puppetdb
     query: "inventory[certname] { facts.osfamily = 'RedHat' }"
     target_mapping:
-      name: certname
+      name: facts.trusted.certname
       config:
         ssh:
           hostname: facts.networking.interfaces.en0.ipaddress
--- a/documentation/using_plugins.md
+++ b/documentation/using_plugins.md
@@ -116,16 +116,25 @@ The following fields are available to the `puppetdb` plugin.
 | **`query`** | A string containing a [PQL query](https://puppet.com/docs/puppetdb/latest/api/query/v4/pql.html) or an array containing a [PuppetDB AST format query](https://puppet.com/docs/puppetdb/latest/api/query/v4/ast.html).<br> **Required.** | `String` | None |
 | `target_mapping` | A hash of target attributes (`name`, `uri`, `config`) to populate with fact lookup values. | `Hash` | None |
 
+#### Available fact paths
+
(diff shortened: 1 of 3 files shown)


nmaludy:
@adreyer it looks like at some point we thought `certname` was a valid path looking at this code? https://github.com/puppetlabs/bolt/blob/master/lib/bolt/plugin/puppetdb.rb#L34-L43

nmaludy:
I tried downgrading below PuppetDB 6.8.0 and something must different in the database and i couldn't get PuppetDB to start. As far back as 6.8.0 this query still fails and `certname` path doesn't exist, it is only `trusted.certname`.

nmaludy:
@adreyer i took your suggestion and added that hardcoded 'certname' fact path to the code. I've altered the documentation back to the example with `name: certname` and added a section about a reference of fact paths that can be used.

PR #1691, opened 2020-03-26
(GH-1685) Add data type functions reference docs

### LINKED ISSUE #1685: Attributes of bolt Datatypes should be discoverable from function reference page.

From a user perspective attributes and functions are interchangeable and it's confusing to have them documented on different pages. The functions pages should have 2 sections. Functions and Datatypes. The Datatypes section should include each datatype and the documentation for it's attributes and methods so that that all functions, attributes and methods re ctrl+fable. For now we should just write the datatypes documentation a template file rather than worrying about adding datatypes support to puppet strings.

beechtom:
This adds reference documentation for the various data type functions
available in plans.

> **Note:** This commit includes two options for the reference docs. The
first includes the data type functions in the existing `Bolt functions`
reference doc, while the second introduces a new `Bolt data types`
reference doc.

Closes #1685

!no-release-note

### DIFF
--- a/documentation/bolt_types_reference.md
+++ b/documentation/bolt_types_reference.md
@@ -1,0 +1,98 @@
+# Bolt data types
+
+This page lists custom data types used in Bolt plans and their functions.
+
+## `ApplyResult`
+
+An [apply action](applying_manifest_blocks.md#return-value-of-apply-action)
+returns an `ApplyResult`. An `ApplyResult` is part of a `ResultSet` object and
+contains information about the apply action. 
+
+You can access `ApplyResult` functions with dot notation, using the syntax: 
+`$apply_result.function`.
+
+The following functions are available to `ApplyResult` objects. 
+
+| Function | Type returned | Description |
+|---|---|---|
+| `action` | `String` | The action performed. `$apply_result.action` always returns the string `apply`. |
+| `error` | `Error` | An object constructed from the `_error` field of the result's value. |
+| `message` | `String` | The `_output` field of the result's value. |
+| `ok` | `Boolean` | Whether the result was successful. |
+| `report` | `Hash` | The Puppet report from the apply action. |
+| `target` | `Target` | The target the result is from. |
+| `to_data` | `Hash` | A serialized representation of `ApplyResult`. |
+
+## `Result`
+
+For each target that you execute an action on, Bolt returns a `Result` object and adds the 
+`Result` to a `ResultSet` object. A `Result` object contains information about the action 
+you executed on the target.
+
+You can access `Result` functions with dot notation, using the syntax: `$result.function`.
+
+The following functions are available to `Result` objects.
+
+| Function | Type returned | Description |
+|---|---|---|
+| `[]` | `Data` | Accesses the value hash directly and returns the value for the key. This function does not use dot notation. Call the function directly on the `Result`. For example, `$result[key]`. |
+| `action` | `String` | The type of result. For example, `task` or `command`. |
+| `error` | `Error` | An object constructed from the `_error` field of the result's value. |
+| `message` | `String` | The `_output` field of the result's value. |
+| `ok` | `Boolean` | Whether the result was successful. |
+| `status` | `String` | Either `success` if the result was successful or `failure`. |
+| `target` | `Target` | The target the result is from. |
+| `to_data` | `Hash` | A serialized representation of `Result`. |
+| `value` | `Hash` | The output or return of executing on the target. |
+
+## `ResultSet`
+
+For each target that you execute an action on, Bolt returns a `Result` object
+and adds the `Result` to a `ResultSet` object. In the case of 
+[apply actions](applying_manifest_blocks.md), Bolt returns a `ResultSet` 
+with one or more `ApplyResult` objects.
+
+You can access `ResultSet` functions with dot notation, using the syntax:
+`$result_set.function`.
+
+The following functions are available to `ResultSet` objects:
+
+| Function | Parameters (if applicable) | Type returned | Description |
+|---|---|---|---|
+| `[]` || `Variant[Result, ApplyResult, Array[Variant[Result, ApplyResult]]]` | The accessed results. This function does not use dot notation. Call the function directly on the `Result`. For example, `$result_set[0]`, `$result_set[0, 2]`. |
+| `count` || `Integer` | The number of results in the set. |
+| `empty` || `Boolean` | Whether the set is empty. |
+| `error_set` || `ResultSet` | The set of failing results. |
+| `filter_set(block)` | `block` | `ResultSet` | Filters a set of results by `block`. |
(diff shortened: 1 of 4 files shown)


documentation/plan_functions.md
```diff
@@ -0,0 +1,1646 @@
+# Bolt functions and types
```
lucywyman: I personally like keeping types and functions separate, following the Puppet documentation and avoiding a super long multi-purpose doc.

documentation/bolt_types_reference.md
```diff
@@ -0,0 +1,319 @@
+# Bolt data types
+
+## `ApplyResult`
```
lucywyman: Defer to @hestonhoffman on this, but I think we should dedup this documentation from https://puppet.com/docs/bolt/latest/writing_plans.html#handling-plan-function-results and link to it from there. Also need to update sentence here: https://puppet.com/docs/bolt/latest/applying_manifest_blocks.html#return-value-of-apply-action

documentation/bolt_types_reference.md
```diff
@@ -0,0 +1,319 @@
...
+### `to_data`
+
+**Returns**
+
+`Hash`: A serialized representation of `ApplyResult`.
+
+
+## `Result`
+
+A `Result` is returned from an action executed on a `Target` as part of a `ResultSet`
```
lucywyman:
```suggestion
A `Result` is returned from an action executed on a single target and contains information about the execution. 
`Result`s are aggregated to make a `ResultSet` object.
```

lucywyman: I'm not sure how to say "The return from all your functions is actually going to be ResultSet which is an array of Results" here without being to wordy...but anything else feels confusing.

documentation/bolt_types_reference.md
```diff
@@ -0,0 +1,319 @@
...
+
+**Returns**
+
+`Target`: The target the result is from.
+
+### `value`
+
+**Returns**
+
+`Hash`: The value of the result.
```
lucywyman: This doesn't feel helpful. Maybe "The result of executing the action on the target"

documentation/bolt_types_reference.md
```diff
@@ -0,0 +1,319 @@
...
+The `Target` object represents a target and its specific connection options.
+
+The following functions are available to `ResultSet` objects. They are accessed with
+dot notation `$result_set.function`.
+
+### `config`
+
+**Returns**
+
+`Hash[String, Data]`: The target-level configuration.
```
lucywyman: I think including a line about how this is only data from the inventory, though I think it's all inventory data that applies to a target (not just defined specifically for that target)

lucywyman: This in addition to the note below.

documentation/bolt_types_reference.md
```diff
@@ -0,0 +1,319 @@
+# Bolt data types
+
+## `ApplyResult`
+
+An `ApplyResult` is returned from an [apply action](https://puppet.com/docs/bolt/latest/applying_manifest_blocks.html#)
```
lucywyman: I think this can be relative, so just `[apply action](applying_manifest_blocks.md#)

documentation/bolt_types_reference.md
```diff
@@ -0,0 +1,319 @@
...
+
+**Returns**
+
+`Hash`: The value of the result.
+
+
+## `ResultSet`
+
+A `ResultSet` is returned from an execution function and contains one or more `Result`
+objects. An [apply action](https://puppet.com/docs/bolt/latest/applying_manifest_blocks.html#)
```
lucywyman: Relative link

documentation/bolt_types_reference.md
```diff
@@ -0,0 +1,319 @@
...
+
+**Returns**
+
+`Integer`: The target's connection port.
+
+### `protocol`
+
+**Returns**
+
+`String`: The protocol used to connect to the target.
```
lucywyman: Include that this is the same as the transport except for 'remote' transport

documentation/bolt_types_reference.md
```diff
@@ -0,0 +1,319 @@
...
+
+**Returns**
+
+`String`: The target's URI.
+
+### `user`
+
+**Returns**
+
+`String`: The user to run as on the target.
```
lucywyman:
```suggestion
`String`: The user to connect to the target.
```

documentation/bolt_types_reference.md
```diff
@@ -0,0 +1,319 @@
...
+
+### `vars`
```
lucywyman: https://puppet.com/docs/puppet/latest/lang_data.html I don't see the `Error` datatype documented in Puppet, but we should probably find a way to document it and include it since we reference it. Ideally adding it to the Puppet docs and linking would be the way to go, but if that's going to take a while including here might not be a bad idea?

documentation/plan_functions.md
hestonhoffman: Yeah, I prefer the option of a separate doc for types

documentation/bolt_types_reference.md
hestonhoffman: I removed the bits from that doc that referenced data types and linked to this doc. Also fixed that link in the second doc

documentation/bolt_types_reference.md
hestonhoffman: I'll open a ticket for this bit and add it to the Puppet docs

hestonhoffman:
I've opened a PR to your branch with my edits @beechtom

documentation/bolt_types_reference.md
```diff
@@ -0,0 +1,150 @@
+# Bolt data types
+
+This page lists the Bolt data types and describes how to use them.
```
lucywyman, line 3:
```suggestion
This page lists custom data types used in Bolt plans and their methods.
```

documentation/bolt_types_reference.md
```diff
@@ -0,0 +1,150 @@
...
+
+## `ApplyResult`
+
+An [apply action](applying_manifest_blocks.md#return-value-of-apply-action)
+returns an `ApplyResult`. An `ApplyResult` is part of a `ResultSet` object and
+contains information about the apply action. You can access `ApplyResult`
+functions with dot notation, using the syntax: `$apply_result.function`.
+
+For example, the following plan uses an apply action to install Docker as root,
```
lucywyman, line 12: At least to me this seems like it breaks the docs vs. reference distinction. I don't know if that's something formalized, but imo we should have examples and explanations in the docs proper, and this doc should assume that users already know how to use `apply` for example. Open to discussing it though.

documentation/bolt_types_reference.md
```diff
@@ -0,0 +1,150 @@
...
+}
+  $results.each |$result| {
+    notice($result.error)}
+}
+```
+
+The following functions are available to `ApplyResult` objects. 
+
+| Function | Type returned | Description |
+|---|---|---|
```
lucywyman, line 31: omg this is so much better

PR #1693, opened 2020-03-26
(PE-28373) Always return a result-like error result

donoghuc:
As part of the push to get orchestrator on 2.x it is expected that a bolt-server request will always expects a `status` and `value` key in the response. In most cases the Bolt::Result will have these keys. In the case where there is an error that is not target/action execution specific (in this case with a request schema validation error), return an error that has the `status` and `value` keys.

This commit also updates a stray `result` key to `value` in the scrub_stack_trace method.

### DIFF
--- a/lib/bolt_server/transport_app.rb
+++ b/lib/bolt_server/transport_app.rb
@@ -57,10 +57,17 @@ module BoltServer
     end
 
     def scrub_stack_trace(result)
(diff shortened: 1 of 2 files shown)


branan:
Is it worth waiting for tests on this? Do they even cover bolt-server?

donoghuc:
@branan I was going to see what fails and go from there

donoghuc:
Updated test

PR #1695, opened 2020-03-27
(maint) Add workflow for release note labels on commits

beechtom:
This adds a workflow that checks that at least one commit in a pull
request has a valid release note label. It uses the `rulesets` action to
match each commit's message against a regex, ensuring that at least one
commit matches the regex. This is used to ensure that release notes are
added to pull requests where needed and can be picked up by the
changelog rake task.

!no-release-note

### DIFF
--- a/.github/rulesets.yaml
+++ b/.github/rulesets.yaml
@@ -1,0 +1,24 @@
+rulesets:
+  commits:
+    message-matches:
+      range: once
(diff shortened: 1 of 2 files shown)


lucywyman:
https://github.com/lucywyman/bolt/pull/8 :raised_hands:

PR #1696, opened 2020-03-30
Bolt::Util.deep_clone fails with frozen hashes

DavidS:
The internet suggests replacing `clone` with `dup` to avoid copying the `frozen?` bit.

### DIFF
--- a/lib/bolt/util.rb
+++ b/lib/bolt/util.rb
@@ -170,7 +170,22 @@ module Bolt
         error_types << Java::JavaLang::CloneNotSupportedException if RUBY_PLATFORM == 'java'
 
         begin
-          cl = obj.clone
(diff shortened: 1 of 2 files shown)


lucywyman:
Thanks @DavidS! I only see the spec test changes in the PR, not sure if lib change is missing?

Also, it seems like the real issue is that we try to modify the cloned object, which inherits the frozen characteristic of the object it clones from. What are you trying to clone that's frozen? I wonder if using `clone(freeze: false)` is a better tact, so we preserve other object attributes?

DavidS:
* I only put in the failing test because I'm not sure how the solution should look like in bolt's context.
* See https://github.com/puppetlabs/puppet_litmus/commit/21dbbc749f7ea661bebab5482a7538eca0cf8ad1 to see where I found this.

PR #1701, opened 2020-03-31
Print group names for targets returned from bolt inventory show --detail

nmaludy:
Previously `bolt inventory show --detail` did not print out a list of group names that a target was a part of.

This addition adds that feature.

Before i write a bunch of tests, would like feedback on this. Not sure if there is a better way of getting a list of group names for a target? Am i on the right path?


adreyer:
This output was supposed to be directly copy/pastable into the inventory file or `Target.new()`. I think we need to hide it behind a flag `--show-groups` otherwise we're going to get bogged down in supporting `Target.new(name => foo, groups => [...])`

lucywyman:
I wonder if `bolt group show --detail` would make more sense, as that wouldn't gum up the inventory subcommand and would solve the use case of "I want to know what targets are in each group".

nicklewis:
I don't think the idea of this output being copy-pasteable really works out in practice.

```
    {
      "name": "cohesive-hooch.delivery.puppetlabs.net",
      "uri": "cohesive-hooch.delivery.puppetlabs.net",
      "alias": [

      ],
      "config": {
        "transport": "ssh",
        "ssh": {
          "connect-timeout": 10,
          "tty": false,
          "load-config": true,
          "disconnect-timeout": 5,
          "host-key-check": false,
          "user": "Administrator"
        }
      },
      "vars": {
      },
      "features": [

      ],
      "facts": {
      },
      "plugin_hooks": {
        "puppet_library": {
          "plugin": "puppet_agent",
          "stop_service": true
        }
      }
    }
```

For one thing, it's JSON which won't work in either inventory or `Target.new`. It also has all the config defaults merged in, which you wouldn't want to include.

I think it would make sense to merge this as it is, since it solves a real problem now, and we can add some option or formatter that will give you actually inventory-compatible (or valid Puppet code for `Target.new`) data+format if we think that's desirable.

nmaludy:
Let me know what the decision is on the approach and i can fix the tests to match.

Thanks!

beechtom:
@nmaludy We're going to move forward with this PR as-is and add `groups` as a field in the hash returned by `--detail`. Just needs a release note and a test fix. 👍

nmaludy:
@beechtom working on that now boss!

Issue #813, opened 2018-12-28
Support ssh connection to windows targets


ndelo:
Problem
Bolt transport make assumptions about what shell and commands are available on the the target based on the transport that is being used. 

Solution:
Command building logic should be split out of command execution logic in the transports. Each transport should probe what the login shell is on the other side of the connection and cache that information with the "connection" for the duration of a bolt execution. It should then build shell specific commands to run over the transports execute functionality.

We only need to support sh/bash for linux and we should assume the same LCD versions of unix commands are available. The code should be written in such a way that we can add csh support without a further refactor however.
We do not need to implement shell discovery functionality in the winrm transport but should for local, ssh, and docker.

Questions:
* How will transport provided features work? We should lazily evaluate provided features by connecting to the target when a user requests features or runs a task for the first time.
* If we know the login shell should we use the login shell when running commands with sudo? Yes, we should make this the default v2.0 behavior. 
* How do we detect powershell?

lucywyman:
Hi @ndelo! We do indeed assume that users using the ssh transport are connecting to unix hosts - https://github.com/puppetlabs/bolt/blob/master/lib/bolt/transport/ssh/connection.rb#L309 is the specific lines that will need to change. I'm not sure what the best approach to this will be though - maybe adding a check for the host os? Or having some kind of 'tag' on the hosts that are windows, and using that instead of the transport to determine commands like mkdir? Most of the team is off this week and early next week, so you might hear more next Wednesday or so!

MikaelSmith:
We'd like to separate the context of transport from the shell you're operating in on the target, but that'll be a bit of an undertaking. We've started it for powershell. I imagine it would look like

Transports: ssh, winrm, local
Shell: powershell, (linux) shell

ssh would default to shell, but could have a config to override that and use powershell on the target.
winrm would default to powershell and could be overridden to shell.

ndelo:
That would work for our use cases. I think Win OpenSSH defaults to CMD, but we’re testing with changing the default shell to PS.

In your thinking, would ‘shell’ be an available CLI option when executing bolt?

Sent from my iPhone

On Jan 2, 2019, at 3:57 PM, Michael Smith <notifications@github.com<mailto:notifications@github.com>> wrote:


We'd like to separate the context of transport from the shell you're operating in on the target, but that'll be a bit of an undertaking. We've started it for powershell. I imagine it would look like

Transports: ssh, winrm, local
Shell: powershell, (linux) shell

[...]

PR #1702, opened 2020-03-31
(GH-813) Abstract Bash shell implementation from local+ssh transports

nicklewis:
Previously, the similarities between the local and SSH transports were
represented by the Sudoable base transport class. This made the two
transports tightly coupled to bash. It also meant we had two separate
implementations of sudo support, one for each of the transports.

This commit introduces a conceptual split between the "connection"
portion of a transport and the "shell" that it talks to on the other
end. In this case, "local" and "SSH" are two implementations of how to
connect to a target, but the Bolt::Shell::Bash class contains all the logic of
how to actually get something done over the connection.

The SSH and local connection classes now are required to implement only
two methods: `execute` and `copy_file`. All other operations are built
in terms of those two primitives. All logic for how to run commands,
scripts, and tasks on a bash target, including how to invoke sudo, is
now contained in th Bolt::Shell::Bash class.

This represents the first step in eventually supporting running against
Powershell targets over SSH.


lib/bolt/shell.rb
```diff
@@ -0,0 +1,85 @@
...
+    end
+
+    def provided_features
+      []
+    end
+
+    def default_input_method(_executable)
+      'both'
+    end
+
```
lucywyman, line 36: I don't think these need to go in separate files, but I do feel like this class is really doing 2 separate things: defining the API for shell classes, and defining common helpers that shells can use. Might be worth at least a comment at this line distinguishing between the two?

lib/bolt/task.rb
```diff
@@ -9,6 +9,9 @@ def initialize(target, task)
   end
 
   class Task
+    STDIN_METHODS       = %w[both stdin].freeze
```
lucywyman, line 12: Is this going to eventually end up in the shell base class?

lib/bolt/transport/base.rb
```diff
@@ -2,6 +2,7 @@
 
 require 'logging'
 require 'bolt/result'
+require 'bolt/shell'
```
lucywyman, line 5: I don't think this is needed since we pull it in in `simple.rb`

lib/bolt/transport/simple.rb
```diff
@@ -0,0 +1,48 @@
...
+require 'bolt/result'
+require 'bolt/shell'
+require 'bolt/transport/base'
+
+module Bolt
+  module Transport
+    # A simple transport has a single connection per target and delegates its
+    # operation to a target-specific shell.
+    class Simple < Base
+      def with_connection(_target)
```
lucywyman, line 13: I wonder if this can also be generalized to
```
conn = Connection.new(target, @transport_logger)
        conn.connect
        yield conn
      ensure
        begin
          conn&.disconnect
        rescue StandardError => e
          logger.info("Failed to close connection to #{target.safe_name} : #{e.message}")
        end
      end
```
with connect and disconnect stubbed for transports that don't use it. That might be more effort than it's worth.

lib/bolt/transport/ssh/connection.rb
```diff
@@ -148,119 +142,88 @@ def disconnect
...
-              sudo_flags += ["-E"] if options[:environment]
-              sudo_str = Shellwords.shelljoin(sudo_flags)
-            else
-              sudo_str = Shellwords.shelljoin(@target.options['run-as-command'] + [run_as])
-            end
-            command_str = build_sudoable_command_str(command_str, sudo_str, @sudo_id, options.merge(reset_cwd: true))
-          end
-
+        def execute(command_str)
           # Including the environment declarations in the shelljoin will escape
```
lucywyman, line 146: I don't think this comment makes sense here anymore

lib/bolt/task.rb
nicklewis: It might make sense for it to eventually be a predicate method (maybe `accepts_stdin?`) on `Task` instead of constants that need to be interpreted by the transport.

I was sort of leaving it for a bit to see how the other transports/shells shake out.

lib/bolt/transport/simple.rb
nicklewis: I definitely considered that and I think it probably makes sense, but it's another one where I wanted to wait and see how the other transports turn out.

lib/bolt/transport/local/connection.rb
```diff
@@ -0,0 +1,61 @@
...
+require 'bolt/util'
+
+module Bolt
+  module Transport
+    class Local < Simple
+      class Connection
+        attr_accessor :user, :logger, :target
+
+        CHUNK_SIZE = 4096
+        PS_ARGS = %w[-NoProfile -NonInteractive -NoLogo -ExecutionPolicy Bypass].freeze
```
lucywyman, line 15: These don't look to get used anywhere in this PR (sorry I'm going one at a time :grimacing:)

lib/bolt/transport/ssh/connection.rb
```diff
@@ -148,119 +142,86 @@ def disconnect
...
-        def execute(command, sudoable: false, **options)
-          result_output = Bolt::Node::Output.new
-          run_as = options[:run_as] || self.run_as
-          escalate = sudoable && run_as && @user != run_as
-          use_sudo = escalate && @target.options['run-as-command'].nil?
+        def execute(command_str)
+          in_rd, in_wr = IO.pipe
+          out_rd, out_wr = IO.pipe
+          err_rd, err_wr = IO.pipe
+          th = Thread.new do
```
lucywyman, line 149: I'm a bit confused about why we're opening a thread here...this is only executed on one target right? What else does threading get us?

lib/bolt/transport/ssh/connection.rb
```diff
@@ -148,119 +142,86 @@ def disconnect
...
-          if result_output.exit_code == 0
-            @logger.debug { "Command returned successfully" }
-          else
-            @logger.info { "Command failed with exit code #{result_output.exit_code}" }
-          end
-          result_output
-        rescue StandardError
-          @logger.debug { "Command aborted" }
-          raise
+        def execute_simple(command)
```
lucywyman, line 211: I don't see where this gets used.

nicklewis: Whoops that was from earlier work, I'll remove it.

lib/bolt/transport/ssh/connection.rb
nicklewis: In order to have a single `sudo` implementation, we need both the local and SSH transports to present the same API. `Net::SSH` provides a blocking API that uses callbacks to send stdout and stderr, while the local transport uses `Open3.popen3` which returns `IO` objects for stdin, stdout and stderr, along with a `Process::Waiter` object that can be polled to wait for the command to finish and receive its exit code.

To allow the two transports to present the same API, we squish `Net::SSH` into the same paradigm as `Open3.popen3`. We create pipes for stdin, stdout and stderr and start a thread to internally wait for the command to finish. This thread effectively bridges between the pipe that `Bolt::Shell::Bash` is writing to and `Net::SSH` that doesn't use `IO` objects but just has a `send_data` method. Similarly, it bridges between the callbacks that `Net::SSH` uses for stdout/stderr and the pipes that the bash shell expects to read from.

Because the shell class and the connection class are both simultaneously reading from and writing to pipes to communicate with each other, they need to be running in separate threads.

lucywyman: :scream: That makes sense but also sounds terrible

lib/bolt/transport/local/connection.rb
```diff
@@ -0,0 +1,60 @@
...
+require 'bolt/node/output'
+require 'bolt/util'
+
+module Bolt
+  module Transport
+    class Local < Simple
+      class Connection
+        attr_accessor :user, :logger, :target
+
+        CHUNK_SIZE = 4096
```
lucywyman, line 14: I don't think this is used here anymore, it's in the bash class

spec/bolt/shell/bash_spec.rb
```diff
@@ -0,0 +1,190 @@
...
+      expect(connection).to receive(:execute).and_return(execute_result)
+
+      result = shell.execute('my cool command')
+      expect(result.stdout.string).to eq("hello world")
+      expect(result.stderr.string).to eq("some errors")
+      expect(result.exit_code).to eq(6)
+    end
+
+    it "runs a command with extremely long UTF-8 output" do
+      stdout = "hello ☃" * 60000
```
lucywyman, line 146: I added a similar test to the local integration tests, and it looks like this case doesn't actually break without this change? 
```
# spec/integration/transport/local_spec.rb
  it "long input" do
    stdin = "hello ☃" * 60000
    expect(runner.run_command(target, "echo #{stdin}").value['stdout'].strip).to eq(stdin)
  end 
```

I'm not sure how to reproduce the error that this change fixes.

nicklewis: Your example isn't passing a large value over stdin, it's running a large command line string.

spec/shared_examples/transport.rb
```diff
@@ -514,55 +514,6 @@ def set_config(target, config)
   end
 end
 
-# Shared failure tests for Transports
-#
-# Requires uploading files and making tempdir to be stubbed to throw Bolt::Node::FileError.
-shared_examples 'transport failures' do
-  context "when it can't upload a file" do
-    it 'returns an error result for upload' do
```
lucywyman, line 522: Were these just removed because they weren't really testing very much?

nicklewis: Yeah, it was testing that if you stub the method that uploads a file to raise an error, then `run_task`, `run_script`, etc will raise an error. Not especially meaningful.

A better version would be to replicate some particular way in which uploading a file can really fail and ensure that _that_ causes an error, but even that seems of rather dubious value.

PR #1703, opened 2020-03-31
Ensure library is loaded before referencing

logicminds:
* Previously some boltlib functions were using part of a bolt
    library from puppet without first loading that file.  This
    becomes a problem when Puppet is used in 3rd party tooling
    that parse Puppet code and also attempt to load the boltlib.

    This fixes the issue by requiring the bolt/pal/issues file
    so that it can be referenced without issue.


adreyer:
These functions don't really work outside of PAL anyway. I think this won't be a problem once we have better testing tools.

PR #1705, opened 2020-04-01
(GH-1686) Add transport and transport_config functions to Target type

### LINKED ISSUE #1686: target.config in plans is confusing
## Describe
Target objects in plans can only see attributes set in the inventory file. While this makes sense for attrs like `vars` and `features` that can only be set in the inventory, `config` (and, less consequentially, `plugin_hooks`) can be set in other places. This is particularly confusing when the config file data is not available through `target.config`.

We should at least consider merging config file data into the config attribute of inventory targets, or other solutions that make this less confusing.

## Solution
* We should add a `.transport` method to the bolt target object that exposes the transport 
* We should also add a `transport_config` method which exposes the transport_config instance variable, which includes config file values and defaults.
* We should also include pre docs around the use case of copying config into a temp variable, modifying config, and then restoring config later in the plan

## Additional Context
https://puppetcommunity.slack.com/archives/C7GP57T50/p1585081361160000

beechtom:
This adds two functions to the `Target` data type in plans:

- `transport`

  Returns the transport used to connect to the target.

 - `transport_config`

   Returns a hash of the merged transport config for the target,
   including defaults, config set in a `bolt.yaml`, and config set at
   the group and target level in an `inventory.yaml` or plan.

Closes #1686 

!feature

* **Added `transport` and `transport_config` functions to `Target` data
  type** ([#1686](#1686))

  The `Target` data type now supports a `transport` function, which
  returns the transport used to connect to the target, and a
  `transport_config` function, which returns a hash of merged
  configuration for the target's transport.

### DIFF
--- a/bolt-modules/boltlib/lib/puppet/datatypes/target.rb
+++ b/bolt-modules/boltlib/lib/puppet/datatypes/target.rb
@@ -27,7 +27,9 @@ Puppet::DataTypes.create_type('Target') do
       password => Callable[[], Optional[String[1]]],
       port => Callable[[], Optional[Integer]],
       protocol => Callable[[], Optional[String[1]]],
-      user => Callable[[], Optional[String[1]]],
+      transport => Callable[[], Optional[String[1]]],
+      transport_config => Callable[[], Optional[Hash[String[1], Data]]],
+      user => Callable[[], Optional[String[1]]]
     }
   PUPPET
 
--- a/bolt-modules/boltlib/lib/puppet/datatypes/target.rb
+++ b/bolt-modules/boltlib/lib/puppet/datatypes/target.rb
@@ -27,8 +27,8 @@ Puppet::DataTypes.create_type('Target') do
       password => Callable[[], Optional[String[1]]],
       port => Callable[[], Optional[Integer]],
       protocol => Callable[[], Optional[String[1]]],
-      transport => Callable[[], Optional[String[1]]],
-      transport_config => Callable[[], Optional[Hash[String[1], Data]]],
+      transport => Callable[[], String[1]],
+      transport_config => Callable[[], Hash[String[1], Data]],
       user => Callable[[], Optional[String[1]]]
     }
   PUPPET
--- a/documentation/bolt_types_reference.md
+++ b/documentation/bolt_types_reference.md
@@ -82,7 +82,7 @@ The following functions are available to `Target` objects:
 
 | Function | Type returned | Description | Note |
 |---|---|---|---|
-| `config` | `Hash[String, Data]` | The inventory configuration for the target. | This function does not return default configuration values or configuration set in a `bolt.yaml` file. It only returns the configuration set in an `inventory.yaml` file or set in a plan using the `Target.new` or `set_config()` functions. |
+| `config` | `Hash[String, Data]` | The inventory configuration for the target. | This function returns the configuration set directly on the target in `inventory.yaml` or set in a plan using `Target.new` or `set_config()`. It does not return default configuration values or configuration set in `bolt.yaml`.  |
 | `facts` | `Hash[String, Data]` | The target's facts. | This function does not lookup facts for a target and only returns the facts specified in an `inventory.yaml` file or set on a target during a plan run. |
 | `features` | `Array[String]` | The target's features. ||
 | `host` | `String` | The target's hostname. ||
--- a/documentation/bolt_types_reference.md
+++ b/documentation/bolt_types_reference.md
@@ -94,7 +94,7 @@ The following functions are available to `Target` objects:
 | `safe_name` | `String` | The target's safe name. Equivalent to `name` if a name was given, or the target's `uri` with any password omitted. ||
 | `target_alias` | `Variant[String, Array[String]]` | The target's aliases. ||
 | `transport` | `String` | The transport used to connect to the target. ||
-| `transport_config` | `Hash[String, Data]` | The merged configuration for the target's `transport`. | This function returns the merged configuration for a target's `transport`, including defaults, configuration set in a `bolt.yaml`, and configuration set in an `inventory.yaml` file or set in a plan using the `Target.new` or `set_config()` functions. |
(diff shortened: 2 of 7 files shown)


beechtom:
This PR is currently rebased off #1691

lib/bolt/target.rb
```diff
@@ -129,6 +129,10 @@ def transport
       inventory_target.transport
     end
 
+    def transport_config
```
lucywyman, line 132: This might also be a good opportunity to get rid of the `options` instance variable and alias, since those can't be defined in inventory or accessed by the target anymore.

beechtom:
Rebased off master now that #1691 is merged in.

lib/bolt/target.rb
```diff
@@ -129,6 +129,11 @@ def transport
...
+      inventory_target.transport_config.to_h
+    end
+    alias options transport_config
```
nicklewis, line 135: We only need the `options` method because all the transports use it, right? ie. the only consumers of it are within Bolt?

bolt-modules/boltlib/lib/puppet/datatypes/target.rb
```diff
@@ -27,7 +27,9 @@
       password => Callable[[], Optional[String[1]]],
       port => Callable[[], Optional[Integer]],
       protocol => Callable[[], Optional[String[1]]],
-      user => Callable[[], Optional[String[1]]],
+      transport => Callable[[], Optional[String[1]]],
+      transport_config => Callable[[], Optional[Hash[String[1], Data]]],
```
nicklewis, line 31: I don't think the results of `transport` or `transport_config` function should be `Optional`. A target always has to have a transport, right? And `transport_config` will always be a hash, just empty if there are no options.

As opposed to the other functions, where particularly `remote` targets might _not_ have any of host/port/user/password.

lib/bolt/target.rb
beechtom: Correct

bolt-modules/boltlib/lib/puppet/datatypes/target.rb
beechtom: Ah, that makes sense. I thought it was just a Puppet-ism. Will update. :)

documentation/bolt_types_reference.md
```diff
@@ -93,6 +93,8 @@ The following functions are available to `Target` objects:
 | `protocol` | `String` | The protocol used to connect to the target. | This is equivalent to the target's `transport`, except for targets using the `remote` transport. For example, a target with the URI `http://example.com` using the `remote` transport would return `http` for the `protocol`. |
 | `safe_name` | `String` | The target's safe name. Equivalent to `name` if a name was given, or the target's `uri` with any password omitted. ||
 | `target_alias` | `Variant[String, Array[String]]` | The target's aliases. ||
+| `transport` | `String` | The transport used to connect to the target. ||
+| `transport_config` | `Hash[String, Data]` | The merged configuration for the target's `transport`. | This function returns the merged configuration for a target's `transport`, including defaults, configuration set in a `bolt.yaml`, and configuration set in an `inventory.yaml` file or set in a plan using the `Target.new` or `set_config()` functions. |
```
lucywyman, line 97:
```suggestion
| `transport_config` | `Hash[String, Data]` | The merged configuration for the target's `transport`. | This function returns the merged configuration for a target's transport. This includes defaults, configuration set in a `bolt.yaml`, configuration set `inventory.yaml`, and configuration set in a plan using `set_config()`.|
```

lucywyman: I think we can avoid the details about which plan functions override all config here - hopefully anyone using Target.new or set_config will have read the relevant docs.

documentation/writing_plans.md
```diff
@@ -432,6 +432,39 @@ target objects, as well as add or remove objects from existing [inventory
...
+  # Store the target's original configuration
+  $original_config = $target.config['ssh']
+
+  # Modify the target's configuration
+  $config =  {
+    'user'     => 'bolt',
+    'password' => 'secret'
+  }
+
+  set_config($target, 'ssh', $config)
```
lucywyman, line 457: Hmmm...this same workflow works if I reset config using `set_config($target, 'ssh', {})` - I think something else is going on, trying to sort out what...

documentation/writing_plans.md
```diff
@@ -432,6 +432,39 @@ target objects, as well as add or remove objects from existing [inventory
...
+
+  ...
+
+  # Restore the target's original configuration
+  set_config($target, 'ssh', $original_config)
```
lucywyman, line 462: Hmmm...this same workflow works if I restore config using `set_config($target, 'ssh', {})` - I think something else is going on, trying to sort out what...

lucywyman: :facepalm: Definitely thought that top-level `config` in inventory got merged into this, but it's only config set on the target specifically.

documentation/bolt_types_reference.md
```diff
@@ -82,7 +82,7 @@ The following functions are available to `Target` objects:
 
 | Function | Type returned | Description | Note |
 |---|---|---|---|
-| `config` | `Hash[String, Data]` | The inventory configuration for the target. | This function does not return default configuration values or configuration set in a `bolt.yaml` file. It only returns the configuration set in an `inventory.yaml` file or the configuration set during a plan using the `Target.new` or `set_config()` functions. |
+| `config` | `Hash[String, Data]` | The inventory configuration for the target. | This function does not return default configuration values or configuration set in a `bolt.yaml` file. It only returns the configuration set in an `inventory.yaml` file or set in a plan using the `Target.new` or `set_config()` functions. |
```
lucywyman, line 85:
```suggestion
| `config` | `Hash[String, Data]` | The inventory configuration for the target. | This function returns the configuration set on the target in `inventory.yaml` or set in a plan using `Target.new` or `set_config()`. It does not return default configuration values or configuration set in `bolt.yaml`.  |
```

PR #1708, opened 2020-04-01
(GH-813) Abstract Powershell shell implementation

nicklewis:
This commit factors out the Powershell code from the WinRM and Windows
local transports to match the transport/shell split implemented for SSH
and Bash. That is, the transports are now merely responsible for
connecting to the target, executing commands and uploading files, while
the powershell shell class handles running tasks, etc.

We no longer have a dedicated local_windows transport, instead using the
standard local transport. The local transport will run any command it's
given via powershell. This matches the non-windows behavior where
commands are run via shell.

Due to bugs in passing stdin from ruby to powershell, we cannot simply
send the commands over stdin. Therefore, we write the commands to a file
and point powershell at that. This also avoids limitations around
command-line length if we were to instead run via `powershell.exe -Command`.

Because commands are now run via powershell, we can run into problems
with powershell inserting a byte order marker in content that we supply
to a task via stdin. For that reason, we now detect whether the
[Console] object is configured to use UTF-8 and, if so, we replace its
encoding a non-BOM version of UTF-8.

This also standardizes the newline behavior between the local and WinRM
transports. Previously, the WinRM transport would return \r\n while the
local transport would return \n. They now consistently return \r\n.

!feature

* **Commands run over local transport on Windows use powershell**
  ([#1708](#1708))

  Previously, the local transport on Windows would exec commands
  directly, meaning powershell constructs couldn't be used. These
  commands are now always executed through powershell, so powershell
  commands and script snippets can be run.

* **Commands and tasks on Windows now consistently return \r\n**
  ([#1708](#1708))

  The local transport on Windows was returning \n while WinRM returned
  \r\n. They are now consistent and always use \r\n.


lib/bolt/shell/powershell.rb
```diff
@@ -0,0 +1,263 @@
...
+      def mkdirs(dirs)
+        mkdir_command = "mkdir -Force #{dirs.uniq.sort.join(',')}"
+        result = execute(mkdir_command)
+        if result.exit_code != 0
+          message = "Could not create directories: #{result.stderr.string}"
+          raise Bolt::Node::FileError.new(message, 'MKDIR_ERROR')
+        end
+      end
+
+      def make_tempdir
```
lucywyman, line 122: Does it make sense to move the tempdir methods to a dedicated tmpdir class like the Bash shell has?

lib/bolt/shell/powershell.rb
```diff
@@ -0,0 +1,263 @@
...
+          raise Bolt::Node::FileError.new("Could not make tempdir: #{result.stderr.string}", 'TEMPDIR_ERROR')
+        end
+        result.stdout.string.chomp
+      end
+
+      def rmdir(dir)
+        execute(Snippets.rmdir(dir))
+      end
+
+      def with_remote_tempdir
```
lucywyman, line 135: This is `with_tempdir` in the bash shell (I *think* this is also used for local windows?)

lib/bolt/shell/powershell.rb
```diff
@@ -0,0 +1,263 @@
...
+      def make_tempdir
+        find_parent = target.options['tmpdir'] ? "\"#{target.options['tmpdir']}\"" : '[System.IO.Path]::GetTempPath()'
+        result = execute(Snippets.make_tempdir(find_parent))
+        if result.exit_code != 0
+          raise Bolt::Node::FileError.new("Could not make tempdir: #{result.stderr.string}", 'TEMPDIR_ERROR')
+        end
+        result.stdout.string.chomp
+      end
+
+      def rmdir(dir)
```
lucywyman, line 131: This is `Tmpdir.#delete` in bash shell - I honestly like rmdir more :shrug:, but I think it makes sense for them to be the same.

lib/bolt/shell/powershell.rb
```diff
@@ -0,0 +1,263 @@
...
+                                 output.stderr.string,
+                                 output.exit_code,
+                                 'command', command)
+      end
+
+      def run_script(script, arguments, _options = {})
+        # unpack any Sensitive data
+        arguments = unwrap_sensitive_args(arguments)
+        with_remote_tempdir do |dir|
+          remote_path = write_executable(dir, script)
```
lucywyman, line 171:
```suggestion
          path = write_executable(dir, script)
```

lib/bolt/shell/powershell.rb
```diff
@@ -0,0 +1,263 @@
...
+      end
+
+      def quote_string(string)
+        "'" + string.gsub("'", "''") + "'"
+      end
+
+      def write_executable(dir, file, filename = nil)
+        filename ||= File.basename(file)
+        validate_extensions(File.extname(filename))
+        remote_path = "#{dir}\\#{filename}"
```
lucywyman, line 85:
```suggestion
        path = "#{dir}\\#{filename}"
```

lib/bolt/shell/powershell.rb
```diff
@@ -0,0 +1,263 @@
...
+            # TODO: optimize upload of directories
+            arguments['_installdir'] = dir
+            task_dir = File.join(dir, task.tasks_dir)
+            mkdirs([task_dir] + extra_files.map { |file| File.join(dir, File.dirname(file['name'])) })
+            extra_files.each do |file|
+              conn.copy_file(file['path'], File.join(dir, file['name']))
+            end
+          end
+
+          remote_task_path = write_executable(task_dir, executable)
```
lucywyman, line 210:
```suggestion
          task_path = write_executable(task_dir, executable)
```

lib/bolt/shell/powershell.rb
```diff
@@ -0,0 +1,263 @@
...
+          output = execute([Snippets.shell_init, *env_assignments, command].join("\n"))
+
+          Bolt::Result.for_task(target, output.stdout.string,
+                                output.stderr.string,
+                                output.exit_code,
+                                task.name)
+        end
+      end
+
+      def execute(command)
```
lucywyman, line 245: This seems like it should be the `execute` method that's defined in `transport/winrm/connection.rb` - that definition of execute doesn't look to be called, but should move here to preserve logging

lib/bolt/transport/winrm/connection.rb
```diff
@@ -105,67 +100,48 @@ def disconnect
...
-        def shell_init
-          return nil if @shell_initialized
-          result = execute(Powershell.shell_init)
-          if result.exit_code != 0
-            raise BaseError.new("Could not initialize shell: #{result.stderr.string}", "SHELL_INIT_ERROR")
-          end
-          @shell_initialized = true
-        end
-
         def execute(command)
```
lucywyman, line 103: This should be moved to powershell.rb

lib/bolt/shell/powershell/snippets.rb
```diff
@@ -0,0 +1,273 @@
...
+              [Parameter(Position=0)] $stream,
+              [Parameter(ValueFromPipeline=$true)] $string
+            )
+            PROCESS {
+              $bytes = $utf8.GetBytes($string)
+              $stream.Write( $bytes, 0, $bytes.Length )
+            }
+            }
+
+            function Convert-JsonToXml {
```
lucywyman, line 102: Where do we use all these conversion functions we define?

lucywyman: nvm found it :P

lib/bolt/transport/local/connection.rb
```diff
@@ -44,7 +45,16 @@ def copy_file(source, dest)
         end
 
         def execute(command)
-          Open3.popen3(command)
+          if Bolt::Util.windows?
+            command += "\r\nif (!$?) { if($LASTEXITCODE) { exit $LASTEXITCODE } else { exit 1 } }"
+            script_file = Tempfile.new(['wrapper', '.ps1'])
```
lucywyman, line 50: Should this use the `Powershell::Tmpdir` (or tempdir methods if they stay in Powershell)? It seems weird to make an exceptional tempfile not in the configured tempdir...

lib/bolt/shell/powershell.rb
nicklewis: `Bash::Tmpdir` primarily exists so we have somewhere to keep track of who currently owns the directory, in support of `run-as`. I don't think powershell has a similar need.

lib/bolt/shell/powershell.rb
nicklewis: Good catch. I tried to standardize some of the method names but missed it.

lib/bolt/shell/powershell.rb
nicklewis: They're distinct because `rmdir` here is a helper method that could be called in any context, whereas `Tmpdir#delete` is operating on a specific object. I think consistency would be to have a `rmdir` method on the Bash shell and have `Tmpdir#delete` call it, but that seems sort of unnecessary.

lib/bolt/shell/powershell.rb
nicklewis: I'm not sure I follow. This `execute` method calls `execute` on the connection.

lib/bolt/transport/winrm/connection.rb
nicklewis: This method is responsible for executing a command over WinRM. The Powershell `execute` calls this method.

lib/bolt/transport/local/connection.rb
nicklewis: You're right, this should be in the tmpdir. The call in `copyfile` is okay I think, since that's a "local" tempfile, not a "remote" tempfile.

lib/bolt/shell/powershell.rb
lucywyman: Ahhh I'm not sure what I thought this was doing but it makes sense.

PR #1712, opened 2020-04-03
(GH-1688) Add BoltSpec helper to load bolt constructs

### LINKED ISSUE #1688: Make it easier to test functions that use Plan Datatypes with rspec-puppet
## Use Case
The Plan Datatypes live in the boltlib module that ships with the bolt gem which makes it hard to test functions that rely on them.

A helper from BoltSpec that can be used to wrap an rspec-puppet function test that creates the same context as a plan executor. You should be able to call the `allow_task` etc functions inside this wrapper.

1. A BoltSpec function that returns the path to `bolt-modules` inside the gem.
2. CI that publishes bolt-modules to forge

These will not create an executor or inventory so most functions that use plan language features would still fail.

lucywyman:
This creates a new helper module `BoltSpec::BoltContext` which provides
the `in_bolt_context` method. This method makes Bolt datatypes and
functions available to a block allowing code that references Bolt
constructs to be tested. The Plans module was already doing all of this
work, creating a mock executor, stubbing run_* functions, and setting
default config, inventory, and modulepath values. This moves all of that
logic into a common BoltContext module so that it can be generally
useful as opposed to being coupled with testing Bolt plans. This allows
things like Puppet functions that refer to the `TargetSpec` datatype to
be tested without having to wrap them in a Bolt plan for testing.

Closes #1688

!feature

* **Add Boltspec helper to load Bolt constructs** ([#1688](https://github.com/puppetlabs/bolt/issues/1688))

  A new helper function `in_bolt_context` can be used to wrap code that
  references Bolt constructs, such as the TargetSpec datatype.

### DIFF
--- a/bolt_spec_spec/spec/spec_helper.rb
+++ b/bolt_spec_spec/spec/spec_helper.rb
@@ -1,0 +1,3 @@
+# frozen_string_literal: true
+
+$LOAD_PATH.unshift File.join(__dir__, '..', '..', 'lib')
--- a/Rakefile
+++ b/Rakefile
@@ -45,6 +45,12 @@ RSpec::Core::RakeTask.new(:slow) do |t|
   t.rspec_opts = '--tag puppetserver --tag puppetdb --tag expensive'
 end
 
+task :bolt_spec do
+  Dir.chdir("#{__dir__}/bolt_spec_spec/") do
+    sh "rake spec"
+  end
+end
+
 RuboCop::RakeTask.new(:rubocop) do |t|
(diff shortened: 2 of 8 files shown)


bolt_spec_spec/spec/spec_helper.rb
```diff
@@ -0,0 +1,4 @@
+# frozen_string_literal: true
+
+ENV['RACK_ENV'] = 'test'
```
adreyer, line 3: do we need this for anything?

lucywyman: Woops nope

Issue #1713, opened 2020-04-05
Error when applying on local.


jaklaassen:
## Expected Behavior
On a non-trivial Puppet setup, my bolt apply runs are failing on localhost, with the following error:

[bolt_stacktrace.txt](https://github.com/puppetlabs/bolt/files/4434889/bolt_stacktrace.txt)

## Steps to Reproduce
1. Run `bolt apply`

## Environment
 - Version: 2.3.1
 - Platform: Ubuntu 18.04

## Additional Context
I believe I've tracked this down to incorrectly passing stdin to the local subshell. 

This fixed it for me, but I'm pretty new to Bolt so not sure if this is right: https://github.com/puppetlabs/bolt/pull/1714

lucywyman:
Closed by #1702

PR #1714, opened 2020-04-05
Fix a bug passing arguments to local shell transport.

jaklaassen:
!bug

* **Fix a bug passing arguments to local shell transport.** ([#1713](https://github.com/puppetlabs/bolt/issues/1713))

  Local shell transport was not correctly slicing stdin when writing to the subprocess.

### DIFF
--- a/lib/bolt/transport/local/shell.rb
+++ b/lib/bolt/transport/local/shell.rb
@@ -184,7 +184,7 @@ module Bolt
 
             begin
               if writable && index < in_buffer.length
-                to_print = in_buffer[index..-1]
+                to_print = in_buffer.byteslice(index..-1)
                 written = inp.write_nonblock to_print
                 index += written
 


jaklaassen:
This fixed the referenced issue for me, but not sure if this variable will always be a string?

lib/bolt/transport/local/shell.rb
```diff
@@ -184,7 +184,7 @@ def execute(command, sudoable: true, **options)
 
             begin
               if writable && index < in_buffer.length
```
nicklewis, line 186: Do we need to also change `length` here (and below) to `bytesize`?

Perhaps we ought to make a copy of `in_buffer` encoded as binary and then the existing algorithm should work.

nicklewis:
#1702 moves this code to another file and I wanted to make sure this fix doesn't get lost, so I introduced a fix in that PR. I opted to make a binary copy of the string. I'm going to close this PR in favor of the other. Thank you for doing the hard work of tracking down the problem!

Issue #1718, opened 2020-04-06
Human outputter does not print plan result for wait_until_available correctly


donoghuc:
## Describe the Bug
The result object created from wait_until_available does not have an `action`. The human outputter wants to print the action and fails when it attempts to capitalize it when printing a result. 

## Expected Behavior
Maybe wait_until_available should have an action? 

## Steps to Reproduce
```
plan test(){
    return wait_until_available('localhost')
}
```
```
NoMethodError: undefined method `capitalize' for nil:NilClass
  /home/cas/working_dir/bolt/lib/bolt/outputter/human.rb:101:in `print_result'
  /home/cas/working_dir/bolt/lib/bolt/outputter/human.rb:362:in `block in print_result_set'
  /home/cas/working_dir/bolt/lib/bolt/result_set.rb:41:in `block in each'
  /home/cas/working_dir/bolt/lib/bolt/result_set.rb:41:in `each'
  /home/cas/working_dir/bolt/lib/bolt/result_set.rb:41:in `each'
  /home/cas/working_dir/bolt/lib/bolt/outputter/human.rb:362:in `print_result_set'
  /home/cas/working_dir/bolt/lib/bolt/outputter/human.rb:355:in `print_plan_result'
  /home/cas/working_dir/bolt/lib/bolt/cli.rb:524:in `run_plan'
  /home/cas/working_dir/bolt/lib/bolt/cli.rb:373:in `execute'
  /home/cas/working_dir/bolt/exe/bolt:10:in `<top (required)>'
  /home/cas/working_dir/bolt/vendor/bundle/ruby/2.5.0/bin/bolt:23:in `load'
  /home/cas/working_dir/bolt/vendor/bundle/ruby/2.5.0/bin/bolt:23:in `<top (required)>'
```

nmaludy:
This can be closed, it is fixed in 2.5.0 as part of #1726

PR #1719, opened 2020-04-06
(GH-1718) Give wait_until_available Result an action

donoghuc:
When constructing a `Result` for `wait_until_available`, add an `action` key so that the human outputter can print print a plan result.


lucywyman:
@donoghuc It might be easiest to just open a new PR with just the PuppetDB client changes, since I think we're going to move forward with Nick Maludy's PR which should fix the wait_until_available action bit. I can also just push to this branch if you're busy. Let me know what you wanna do.

donoghuc:
Seems like there is no reason to change the client if there is not a real issue with http_client tls/ssl. I can just close this. :)

PR #1721, opened 2020-04-07
(maint) Bump checkout action to v2

beechtom:
This bumps the `checkout` action used in the CI workflows to `v2`.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -11,8 +11,11 @@ require "json"
 require "erb"
 
 # Needed for Vanagon component ship job
-require 'packaging'
-Pkg::Util::RakeUtils.load_packaging_tasks
+# Do not load in GitHub workflows
+unless ENV['GITHUB_WORKFLOW']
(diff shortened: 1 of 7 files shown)


Rakefile
```diff
@@ -11,8 +11,11 @@ require "json"
 require "erb"
 
 # Needed for Vanagon component ship job
-require 'packaging'
-Pkg::Util::RakeUtils.load_packaging_tasks
+# Do not load in GitHub workflows
+unless ENV['GITHUB_WORKFLOW']
```
beechtom, line 15: The `packaging` gem runs `git describe`, which causes the workflows to fail as the `checkout@v2` action does not pull in git tags. An alternative solution would be to pull in the tags in a separate step, but not requiring `packaging` for GitHub workflows seems more appropriate.

lucywyman: :shrug: Good thing our release pipeline isn't an action I guess

PR #1723, opened 2020-04-08
Add 3rd party plugins to plugins list

magec:
Adds two 3rd party plugins:
 
- [Consul Inventory Plugin](https://github.com/magec/puppetlabs-consul_inventory)
- [Hetzner Cloud Inventory Plugin](https://github.com/magec/puppetlabs-hcloud_inventory)


magec:
I saw this  http://docs.puppet.com/community/trivial_patch_exemption.html but the link is broken, I understand this patch fills into this.

PR #1727, opened 2020-04-08
(GH-1673) Add run_task_with plan function

### LINKED ISSUE #1673: Support per target task parameters in plans
## Use Case

Commonly, a large number of tasks need to be run against targets, but in such a way that a single invocation of `run_task()` is not feasible. For example, a task that requires some minor modification of parameter inputs on a target-by-target basis. Instead of one invocation of `run_task()`, a loop must be used to call `run_task()` once per node with customized inputs.

Because plans in Bolt lack any kind of parallelism features, the performance of one invocation of `run_task()` with many targets vs. many invocations of `run_task()` with one target is markedly different. A job that would take 30 seconds to run per node could draw out the time to execute a plan to 30 seconds * the number of nodes, which could easily become infeasibly long.

An example Bolt plan fragment is given below, showing what it can look like when `run_task()` needs to be called in a loop.

```puppet
['A', 'B'].each |$group| {
  getvar("compiler_${group.downcase()}_targets").each |$target| {
    run_task('peadm::agent_install', $target,
      server        => $master_target.peadm::target_name(),
      install_flags => [
        "main:certname=${target.peadm::target_name()}",
        "extension_requests:${pp_cluster}=${group}",
      ],
    )
  }
}
```

Bolt plans should provide some kind of parallelism feature for constructing and dispatching many tasks, but letting those tasks run in parallel and not blocking until all the dispatched tasks have completed.

Suggested implementation (per Slack conversation with @adreyer): a block-based syntax to `run_task()` which calls a block per-target to set task parameters. E.g.

```puppet
run_task(foo, $targets) |$t| {
   { param1 => 'foo', param2 => $t.name}
}
```

This provides parallelism for the use case of calling the _same_ task for a set of targets, which each need customized input parameters. While this doesn't provide unbounded parallelism power, it probably solves the 90% use case and has the strong benefit of being very simple but still decently powerful, and not dirty or ugly.

Implementation approaches could include:

* passing parameters as a map to `run_task()` or `run_task_per_target()` 

      run_task(foo, $targets, _target_params => {
         "t1" => { key => val1},
         "t2" => {key => val2},
      } )

This seems more verbose and could be confusing is the keys map to groups. 

# Questions

* Should this be `run_task` or a new function `run_task_with_per_target_parameters`? Is it easy for users to understand what the block does? Is there some other case we'd want to add a block to `run_task`? As part of this ticket poll in slack to determine what the least confusing and long name for this is.
* How long is PE likely to lack support for this behavior in plans it executes? Will require DB and UI updates, and updates to the pe-bolt-server executor.
* Should it be a block that returns the parameters or a parameter that is a map of targets to parameters? Block tests better with @reidmv (add more examples for further tests). Update 04/01/2020: a block is more elegant and likely better for the common case of using facts or vars on a target.
* Should we accept defaults too? Not if it's a new function. So no.
* Do we collect all params first? Yes.
* Should we add support for other `run_*`? Not plan maybe the others. Not as part of this ticket
* Do we need to add a YAML step for this? Not now.
* What should `catch_errors` do? Test what `catch_errors` does and see if it's reasonable...
* What happens if parameters are invalid for a single target? That one target should fail, other targets should continue executing.

beechtom:
This adds a new plan function `run_task_with` that runs a task on a set
of targets with target-specific parameters. It accepts a lambda that
returns a `Hash` of parameters to pass to the task run on a particular
target.

As an example, a task might expect a parameter with a value that is set
by a target's `facts`. The `run_task_with` function would have a lambda
that assigns this parameter with the appropriate fact:

```
$results = run_task_with('my_task', $targets) |$t| {
    { 'osfamily' => $t.facts['osfamily'] }
}
```

Similar to the `run_task` plan function, `run_task_with` also accepts
an optional `description` parameter, as well as three metaparameters:

- `_catch_errors`
- `_noop`
- `_run_as`

Closes #1673 

!feature

* **Added `run_task_with` plan function**
([#1673](https://github.com/puppetlabs/bolt/issues/1673))

  The new plan function `run_task_with` lets you run tasks on a set of
  targets with target-specific parameters. It accepts a lambda that
  returns a `Hash` of parameters for a particular target.

### DIFF
--- a/bolt-modules/boltlib/lib/puppet/functions/run_task_with.rb
+++ b/bolt-modules/boltlib/lib/puppet/functions/run_task_with.rb
@@ -1,0 +1,192 @@
+# frozen_string_literal: true
+
+require 'bolt/error'
+require 'bolt/pal'
+require 'bolt/task'
+
+# Runs a given instance of a `Task` with target-specific parameters on the given set of targets and
+# returns the result from each. This function differs from {run_task} by accepting a block that returns
+# a `Hash` of target-specific parameters that are passed to the task. This can be used to send parameters
+# based on a target's attributes, such as its `facts`, or to use conditional logic to determine the
+# parameters a task should receive for a specific target.
+#
+# This function does nothing if the list of targets is empty.
+#
+# > **Note:** Not available in apply block
+#
+# > **Note:** Not available to targets using the pcp transport
+Puppet::Functions.create_function(:run_task_with) do
+  # Run a task with target-specific parameters.
+  # @param task_name The task to run.
+  # @param targets A pattern identifying zero or more targets. See {get_targets} for accepted patterns.
+  # @param options A hash of additional options.
+  # @option options [Boolean] _catch_errors Whether to catch raised errors.
+  # @option options [Boolean] _noop Run the task in noop mode if available.
+  # @option options [String] _run_as User to run as using privilege escalation.
+  # @param block A block that returns a `Hash` of target-specific parameters for the task.
+  # @return A list of results, one entry per target.
+  # @example Run a task with target-specific parameters as root
+  #   run_task_with('my_task', $targets, '_run_as' => 'root') |$t| {
+  #     { 'param1' => $t.vars['var1'],
+  #       'param2' => $t.vars['var2'] }
+  #   }
+  dispatch :run_task_with do
+    param 'String[1]', :task_name
+    param 'Boltlib::TargetSpec', :targets
+    optional_param 'Hash[String[1], Any]', :options
+    required_block_param 'Callable[Target]', :block
+    return_type 'ResultSet'
+  end
+
+  # Run a task with target-specific parameters, logging the provided description.
+  # @param task_name The task to run.
+  # @param targets A pattern identifying zero or more targets. See {get_targets} for accepted patterns.
+  # @param description A description to be output when calling this function.
+  # @param options A hash of additional options.
+  # @option options [Boolean] _catch_errors Whether to catch raised errors.
+  # @option options [Boolean] _noop Run the task in noop mode if available.
+  # @option options [String] _run_as User to run as using privilege escalation.
+  # @param block A block that returns a `Hash` of target-specific parameters for the task.
+  # @return A list of results, one entry per target.
+  # @example Run a task with target-specific parameters and a description
+  #   run_task_with('my_task', $targets, 'Update system packages') |$t| {
+  #     { 'param1' => $t.vars['var1'],
+  #       'param2' => $t.vars['var2'] }
+  #   }
+  dispatch :run_task_with_with_description do
+    param 'String[1]', :task_name
+    param 'Boltlib::TargetSpec', :targets
+    param 'Optional[String]', :description
+    optional_param 'Hash[String[1], Any]', :options
+    required_block_param 'Callable[Target]', :block
+    return_type 'ResultSet'
+  end
+
+  def run_task_with(task_name, targets, options = {}, &block)
+    run_task_with_with_description(task_name, targets, nil, options, &block)
+  end
+
+  def run_task_with_with_description(task_name, targets, description, options = {})
+    unless Puppet[:tasks]
+      raise Puppet::ParseErrorWithIssue
+        .from_issue_and_stack(
+          Bolt::PAL::Issues::PLAN_OPERATION_NOT_SUPPORTED_WHEN_COMPILING,
+          action: 'run_task_with'
+        )
+    end
+
+    executor  = Puppet.lookup(:bolt_executor)
+    inventory = Puppet.lookup(:bolt_inventory)
+    error_set = []
+
+    # Report to analytics
+    executor.report_function_call(self.class.name)
+    executor.report_bundled_content('Task', task_name)
+
+    # Keep valid metaparameters, discarding everything else
+    options = options.select { |k, _v| k.start_with?('_') }
+                     .transform_keys { |k| k.sub(/^_/, '').to_sym }
+
+    options[:description] = description if description
+
+    # Get all the targets
+    targets = Array(inventory.get_targets(targets))
+
+    # If all targets use the 'pcp' transport, use a fake task instead of loading the local definition
+    # Otherwise, load the local task definition
+    if (pcp_only = targets.any? && targets.all? { |t| t.transport == 'pcp' })
+      task = Bolt::Task.new(task_name, {}, [{ 'name' => '', 'path' => '' }])
+    else
+      task_signature = Puppet::Pal::ScriptCompiler.new(closure_scope.compiler).task_signature(task_name)
+
+      if task_signature.nil?
+        raise Bolt::Error.unknown_task(task_name)
+      end
+
+      task = Bolt::Task.from_task_signature(task_signature)
+    end
+
+    # Map the targets to their specific parameters and merge with the defaults
+    target_mapping = targets.each_with_object({}) do |target, mapping|
+      params = yield(target)
+
+      # Parameters returned from the block should be a Hash. If they're not, create a failing
+      # Result for the target that will later be added to the ResultSet.
+      unless params.is_a?(Hash)
+        exception = with_stack(
+          :TYPE_MISMATCH,
+          "Block must return a Hash of parameters, received #{params.class}"
+        )
+        error_set << Bolt::Result.from_exception(target, exception, action: 'task')
+        next
+      end
+
+      # If parameters are mismatched, create a failing result for the target that will later
+      # be added to the ResultSet.
+      unless pcp_only
+        params = task.parameter_defaults.merge(params)
+
+        type_match = task_signature.runnable_with?(params) do |mismatch_message|
+          exception = with_stack(:TYPE_MISMATCH, mismatch_message)
+          error_set << Bolt::Result.from_exception(target, exception, action: 'task')
+        end
+
+        next unless type_match
+      end
+
+      # If there is a type mismatch between the type Data and the type of the params, create
+      # a failing result for the target that will later be added to the ResultSet.
+      unless Puppet::Pops::Types::TypeFactory.data.instance?(params)
+        params_t = Puppet::Pops::Types::TypeCalculator.infer_set(params)
+        desc = Puppet::Pops::Types::TypeMismatchDescriber.singleton.describe_mismatch(
+          'Task parameters are not of type Data. run_task_with()',
+          Puppet::Pops::Types::TypeFactory.data, params_t
+        )
+        exception = with_stack(:TYPE_NOT_DATA, desc)
+        error_set << Bolt::Result.from_exception(target, exception, action: 'task')
+        next
+      end
+
+      # Wrap parameters marked with '"sensitive": true' in the task metadata with a
+      # Sensitive wrapper type. This way it's not shown in logs.
+      if (param_spec = task.parameters)
+        params.each do |k, v|
+          if param_spec[k] && param_spec[k]['sensitive']
+            params[k] = Puppet::Pops::Types::PSensitiveType::Sensitive.new(v)
+          end
+        end
+      end
+
+      mapping[target] = task.parameter_defaults.merge(params)
+    end
+
+    # Add a noop parameter if the function was called with the noop metaparameter.
+    if options[:noop]
+      if task.supports_noop
+        target_mapping.each_value { |params| params['_noop'] = true }
+      else
+        raise with_stack(:TASK_NO_NOOP, 'Task does not support noop')
+      end
+    end
+
+    if targets.empty?
+      Bolt::ResultSet.new([])
+    else
+      # Combine the results from the task run with any failing results that were
+      # generated earlier when creating the target mapping
+      task_result = executor.run_task_with(target_mapping, task, options)
+      result = Bolt::ResultSet.new(task_result.results + error_set)
+
+      if !result.ok && !options[:catch_errors]
+        raise Bolt::RunFailure.new(result, 'run_task', task_name)
+      end
+
+      result
+    end
+  end
+
+  def with_stack(kind, msg)
+    issue = Puppet::Pops::Issues.issue(kind) { msg }
+    Puppet::ParseErrorWithIssue.from_issue_and_stack(issue)
+  end
+end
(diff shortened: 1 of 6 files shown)


lib/bolt/executor.rb
```diff
@@ -241,52 +245,73 @@ def with_node_logging(description, batch)
 
     def run_command(targets, command, options = {})
       description = options.fetch(:description, "command '#{command}'")
+      # TODO: Consider creating a run_command_* plan function that accepts a block
+      # that returns a target-specific command.
+      target_mapping = targets.product([command]).to_h
```
nicklewis, line 250: This use of `product` is a little mind-boggling but I think it's fine. Another option would be `targets.zip([command]*targets.length)` but I don't know that it's any clearer.

lib/bolt/transport/base.rb
```diff
@@ -164,8 +164,8 @@ def batch_connected?(targets)
       #
       # Transports may override this method, and the corresponding batch_*
       # methods, to implement their own batch processing.
-      def batches(targets)
-        targets.map { |target| [target] }
+      def batches(target_mapping)
+        target_mapping.keys.map { |target| [target] }
```
nicklewis, line 168: I think we want to pass the idea of target mapping all the way down to the transport's `batch_*` methods. So this method should return a list of single-element hashes which the executor will pass directly into `batch_*`. Then `batch_*` can pull them apart and call like `run_command(target, command, ...)`.

lib/bolt/transport/orch.rb
```diff
@@ -176,8 +176,12 @@ def batch_upload(targets, source, destination, options = {}, &callback)
         end
       end
 
-      def batches(targets)
-        targets.group_by { |target| Connection.get_key(target.options) }.values
+      def batches(target_mapping)
+        if target_mapping.values.uniq.length > 1
```
nicklewis, line 180: This could use a comment explaining the rationale.

bolt-modules/boltlib/lib/puppet/functions/run_task_with.rb
```diff
@@ -0,0 +1,179 @@
...
+    end
+
+    # If all targets use the 'pcp' transport, use a fake task instead of loading the local definition
+    # Otherwise, load the local task definition
+    if targets.any? && targets.all? { |t| t.transport == 'pcp' }
+      task = Bolt::Task.new(task_name, {}, [{ 'name' => '', 'path' => '' }])
+    else
+      task_signature = Puppet::Pal::ScriptCompiler.new(closure_scope.compiler).task_signature(task_name)
+
+      if task_signature.nil?
```
nicklewis, line 113: We should check whether the task exists before we compute the target parameters.

bolt-modules/boltlib/lib/puppet/functions/run_task_with.rb
```diff
@@ -0,0 +1,179 @@
...
+
+      # Retrieve the default parameters for the task and merge them with the
+      # target-specific parameters
+      default_params = task.parameter_defaults
+      target_mapping.transform_values! { |params| default_params.merge(params) }
+
+      target_mapping.each_value do |params|
+        task_signature.runnable_with?(params) do |mismatch_message|
+          raise with_stack(:TYPE_MISMATCH, mismatch_message)
+        end || (raise with_stack(:TYPE_MISMATCH, 'Task parameters do not match'))
```
nicklewis, line 127: This will abort the entire function if any target has invalid params, right? Is that the correct behavior or should it still try to run on targets that have valid params?

bolt-modules/boltlib/lib/puppet/functions/run_task_with.rb
```diff
@@ -0,0 +1,179 @@
...
+            params[k] = Puppet::Pops::Types::PSensitiveType::Sensitive.new(v)
+          end
+        end
+      end
+
+      # executor.noop is set when run task is called from the CLI
+      # options[:noop] is set when it's called from a plan
+      if executor.noop || options[:noop]
+        if task.supports_noop
+          params['_noop'] = true
```
nicklewis, line 157: What `params` is this modifying? Shouldn't this get folded into the params for each target?

lib/bolt/applicator.rb
```diff
@@ -206,7 +206,7 @@ def apply_ast(raw_ast, targets, options, plan_vars = {})
         end
 
         result_promises = targets.zip(futures).flat_map do |target, future|
-          @executor.queue_execute([target]) do |transport, batch|
+          @executor.queue_execute(target => nil) do |transport, batch|
```
nicklewis, line 209: Could this... whole thing be replaced with a call to `run_task_with`? Hypothetically

lib/bolt/executor.rb
beechtom: Fwiw, I also looked at `each_with_object` and `zip`. Using `product` seemed like the most concise way to accomplish what will hopefully be a temporary workaround.

bolt-modules/boltlib/lib/puppet/functions/run_task_with.rb
beechtom: It does get folded into the params for each target. This is in a loop that iterates over each parameter hash in `target_mapping`.

bolt-modules/boltlib/lib/puppet/functions/run_task_with.rb
beechtom: Correct.

Would the following approach be more correct?
- Partition targets based on valid/invalid params
- Run tasks on targets with valid params
- Create failing results for targets with invalid params and add them to the resultset

beechtom: Updated to create failing results using `Bolt::Result.from_exception` and add them to the resultset returned from the task run.

bolt-modules/boltlib/lib/puppet/functions/run_task_with.rb
beechtom: Updated to make it clearer that it updates the params mapped to each target.

bolt-modules/boltlib/lib/puppet/functions/run_task_with.rb
```diff
@@ -0,0 +1,195 @@
...
+      params = yield(target)
+
+      # Parameters returned from the block should be a Hash. If they're not, create a failing
+      # Result for the target that will later be added to the ResultSet.
+      unless params.is_a?(Hash)
+        exception = with_stack(
+          :TYPE_MISMATCH,
+          "Block must return a Hash of parameters, received #{params.class}"
+        )
+        error_set << Bolt::Result.from_exception(target, exception, action: 'task')
```
lucywyman, line 118: Hm, I think the common case here is going to be that users have defined a hash where the values are target attributes. If this error is raised it's likely that there's a typo in the block somewhere, and all targets will fail. I wonder if it's better to fail the plan here so that it doesn't need to check the output for every target just to fail all of them. Mostly I think the case where this fails for only a few targets will be rare, and that the output will be annoying for more than a handful of targets and will waste cycles:
```
lucy ~/githubs/bolt ((8add2fa...)) $ be bolt plan run test -t docker
Starting: plan test
Starting: install puppet and gather facts on 127.0.0.1:20022, 127.0.0.1:20023, 127.0.0.1:20024
Finished: install puppet and gather facts with 0 failures in 3.4 sec
Starting: description on 
Finished: description with 0 failures in 0.0 sec
Finished: plan test in 3.41 sec
Failed on 127.0.0.1:20022:
  Block must return a Hash of parameters, received String (file: /home/lucy/githubs/bolt/site-modules/test/plans/init.pp, line: 6)
Failed on 127.0.0.1:20023:
  Block must return a Hash of parameters, received String (file: /home/lucy/githubs/bolt/site-modules/test/plans/init.pp, line: 6)
Failed on 127.0.0.1:20024:
  Block must return a Hash of parameters, received String (file: /home/lucy/githubs/bolt/site-modules/test/plans/init.pp, line: 6)
Failed on 3 targets: 127.0.0.1:20022,127.0.0.1:20023,127.0.0.1:20024
Ran on 3 targets
```

bolt-modules/boltlib/lib/puppet/functions/run_task_with.rb
```diff
@@ -0,0 +1,195 @@
...
+
+        unless type_match
+          exception = with_stack(:TYPE_MISMATCH, 'Task parameters do not match')
+          error_set << Bolt::Result.from_exception(target, exception, action: 'task')
+          next
+        end
+      end
+
+      # If there is a type mismatch between the type Data and the type of the params, create
+      # a feailing result for the target that will later be added to the ResultSet.
```
lucywyman, line 141:
```suggestion
      # a failing result for the target that will later be added to the ResultSet.
```

bolt-modules/boltlib/lib/puppet/functions/run_task_with.rb
```diff
@@ -0,0 +1,195 @@
...
+      unless pcp_only
+        params = task.parameter_defaults.merge(params)
+
+        type_match = task_signature.runnable_with?(params) do |mismatch_message|
+          exception = with_stack(:TYPE_MISMATCH, mismatch_message)
+          error_set << Bolt::Result.from_exception(target, exception, action: 'task')
+          next
+        end
+
+        unless type_match
```
lucywyman, line 133: This actually just raises the same error twice with a different message, as `next` just escapes the inner `do |mismatch|` block, not the target_mapping loop. I'm not entirely sure what situation `mismatch_message` would not be set in, but you could check in the block if it's nil and set it to the default there, and remove this second check.

lib/bolt/executor.rb
```diff
@@ -101,12 +101,16 @@ def shutdown
     # groups targets by transport, then divides each group into batches as
     # defined by the transport. Yields each batch, along with the corresponding
     # transport, to the block in turn and returns an array of result promises.
-    def queue_execute(targets)
-      targets.group_by(&:transport).flat_map do |protocol, protocol_targets|
+    def queue_execute(target_mapping)
+      protocol_target_mapping = target_mapping.group_by { |target, _value| target.transport }
```
lucywyman, line 105: This seems like it's complicating what you really want, which is to pass around the target and associated parameters together. It's tempting to just add a `parameters` attribute to `Bolt::Target`, but we'd *really* have to make sure to clear it out after every task run. A safer option would be to have a `PlanTarget` object or something (bikeshed on name) that inherits from Bolt::Target and includes a `parameters` accessor. It would gets instantiated in `run_task_with`, and could be more easily passed around here without need to modify how other functions interact with the executor. You could maybe get away with a Struct instead of a whole class, if that seems easier.

nicklewis: I think it would be almost like a `WorkRequest` with an action, parameters and a target. I'm not sure we get a whole lot out of it right now though.

lucywyman: It wouldn't affect behavior, but the data we're passing around now feels sufficiently hard to express and reason about that I think it'd make the code more readable.

adreyer: Are we doing lookup on targets anywhere? At a minimum this feels more like a list of [target, params] than a hash of target -> params.

lib/bolt/executor.rb
```diff
@@ -240,52 +244,69 @@ def with_node_logging(description, batch)
...
+          with_node_logging("Running command '#{batch[:value]}'", batch[:targets]) do
+            transport.batch_command(batch[:targets], batch[:value], options, &method(:publish_event))
           end
         end
       end
     end
 
     def run_script(targets, script, arguments, options = {})
       description = options.fetch(:description, "script #{script}")
+      # TODO: Consider creating a run_script_* plan function that accepts a block
```
adreyer, line 264: this feels like a ticket not a TODO

bolt-modules/boltlib/lib/puppet/functions/run_task_with.rb
```diff
@@ -0,0 +1,195 @@
...
+      end
+    end
+
+    if targets.empty?
+      Bolt::ResultSet.new([])
+    else
+      # Combine the results from the task run with any failing results that were
+      # generated earlier when creating the target mapping
+      task_result = executor.run_task_with(target_mapping, task, options)
+      result = Bolt::ResultSet.new(task_result.results + error_set)
```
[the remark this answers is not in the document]
beechtom: I'm not quite sure I understand. They're added to the results from the task and are returned from the function as part of a `ResultSet`.

bolt-modules/boltlib/spec/functions/run_task_with_spec.rb
```diff
@@ -0,0 +1,437 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+require 'bolt/executor'
+require 'bolt/inventory'
+require 'bolt/result'
+require 'bolt/result_set'
+require 'bolt/target'
```
lucywyman, line 8: I don't think this or 'rspec/expectations' gets used

spec/bolt/executor_spec.rb
```diff
@@ -214,6 +214,77 @@ def mock_node_results
     end
   end
 
+  context 'running a task with per-target params' do
+    let(:target_mapping) do
+      targets.each_with_object({}) { |target, map| map[target] = task_arguments }
```
lucywyman, line 219: Since the expectation also iterates over each target, I think this can use a target attribute instead of a static string to be more like the actual use case for this function without any extra work.

lib/bolt/transport/orch.rb
```diff
@@ -210,6 +210,10 @@ def batch_task(targets, task, arguments, options = {}, &callback)
         end
       end
 
+      def batch_task_with(_targets, _task, _target_mapping, _options = {})
+        raise NotImplementedError, "pcp transport does not support batch_task_with()"
```
nicklewis, line 214: This will result in an unintuitive error message if users try to call `run_task_with()` when using this transport.

spec/bolt/executor_spec.rb
```diff
@@ -335,33 +406,6 @@ def mock_node_results
     end
   end
 
-  context 'prompting' do
```
nicklewis, line 338: Where did these go?

beechtom: Ugh. That's probably VS Code. It's been having weird caching issues the past few weeks when I switch branches.

nicklewis:
This looks good to me but has failing tests.

PR #1730, opened 2020-04-10
(GH-813) Add the ability to select powershell as the SSH login shell

nicklewis:
This adds a new `login-shell` option to the SSH transport which can be used
to tell Bolt which shell to expect. We support a number of bash-compatible
shells and powershell, which is experimental. Selecting any of the
supported non-powershell options will result in identical behavior.

!feature

* **Experimental support for interacting with Windows hosts via PowerShell over SSH**
  ([#813](https://github.com/puppetlabs/bolt/issues/813))

  The `login-shell: powershell` config setting can be set on a target to
  connect over SSH while running commands and tasks via PowerShell instead
  of Bash.


nicklewis:
There's some work to do around options, since currently all the target options are defined for the transport, but some of them (`run-as` related options, `extensions`) are shell-specific.

lib/bolt/config/transport/ssh.rb
```diff
@@ -70,7 +81,8 @@ class SSH < Base
           "connect-timeout"    => 10,
           "tty"                => false,
           "load-config"        => true,
-          "disconnect-timeout" => 5
+          "disconnect-timeout" => 5,
+          "login-shell"        => 'bash'
```
adreyer, line 85: Should the default be sh?

lib/bolt/config/transport/ssh.rb
```diff
@@ -7,6 +7,7 @@ module Bolt
   class Config
     module Transport
       class SSH < Base
+        LOGIN_SHELLS = %w[sh bash zsh dash ksh powershell].freeze
```
lucywyman, line 10: I cannot wait to see what the first PR to this is.

lib/bolt/transport/ssh/connection.rb
```diff
@@ -242,7 +242,11 @@ def net_ssh_verifier(verifier)
         end
 
         def shell
-          @shell ||= Bolt::Shell::Bash.new(target, self)
+          @shell ||= if target.options['login-shell'] == 'powershell'
+                       Bolt::Shell::Powershell.new(target, self)
+                     else
+                       Bolt::Shell::Bash.new(target, self)
```
lucywyman, line 248: I wonder if 'unix' makes more sense for this now? I was thinking of putting up a 'cleanup' PR - I think I missed a few "remote" var names in the first PR, and want to reassess unifying the transport classes that just define `with_connection`. Renaming this could be part of that.

spec/bolt/transport/ssh/connection_spec.rb
```diff
@@ -0,0 +1,128 @@
...
+    end
+  end
+
+  context "when connecting", ssh: true do
+    it "passes proxyjump options" do
+      inventory.set_config(target, 'ssh', 'proxyjump' => 'jump.example.com')
+
+      allow(Net::SSH).to receive(:start) do |_, _, options|
+        expect(options[:proxy]).to be_instance_of(Net::SSH::Proxy::Jump)
+      end
```
lucywyman, line 50: I think this needs to return a stubbed object so `transport.server_version.version` succeeds

nicklewis:
Fixed the tests

beechtom:
@hestonhoffman may have better insight on this, but I think it would be helpful to document how to use this feature, especially since there's some trickiness around making sure you have the right version of OpenSSH installed and have powershell as the default shell. Maybe somewhere on this page: https://puppet.com/docs/bolt/latest/bolt_examples.html ?

lib/bolt/shell/powershell.rb
```diff
@@ -243,6 +247,15 @@ def run_task(task, arguments, _options = {})
       end
 
       def execute(command)
+        if conn.max_command_length && command.length > conn.max_command_length
```
nicklewis, line 250: Or if you're really a fan of punctuation:

```rb
if conn.max_command_length&.< command.length
```

lib/bolt/transport/local/connection.rb
```diff
@@ -32,7 +32,7 @@ def copy_file(source, dest)
           if source.is_a?(StringIO)
             Tempfile.create(File.basename(dest)) do |f|
               f.write(source.read)
-              FileUtils.mv(t, dest)
+              FileUtils.mv(f, dest)
```
nicklewis, line 35: Apparently we no longer use `StringIO` with the local transport when using Bash, so this wasn't spotted by tests. We only use a wrapper script for sudo when using `tty` which is not available for the local transport.

PR #1733, opened 2020-04-10
(GH-1722) Use 'targets' parameter in YAML plan steps

### LINKED ISSUE #1722: Yaml steps use `target` while plan magic is based around `targets`
## Describe the Bug
This is a stumbling block for yaml plan authors

## Expected Behavior
Both use the same key. 

## Questions
- How painful will it be to switch the yaml steps?
- We should rename the step key to `targets`, support both `targets` and `target`
- We should raise a deprecation warning for `target` and drop support for it...sometime.
"You're using an experimental feature (YAML  plans) - `target` will be removed at some point"
- We should update all the function docs are updated - verify that we use `targets` everywhere
- Use targets, if targets isn't set use target, if target is set at all raise warning.
- Make sure any docs that use the target key are updated

beechtom:
This deprecates the `target` parameter for YAML plan steps in favor of a
new `targets` parameter, in line with the 'magic' parameter used in
plans written in Puppet.

YAML plans that have steps with a `target` parameter will raise a
helpful warning notifying the user that the parameter is deprecated and
to use `targets` instead. YAML plan steps will still accept a `target`
parameter, but will prefer a `targets` parameter.

Closes #1722

!deprecation

* **YAML plan step parameter `target` deprecated in favor of `targets`**
([#1722](https://github.com/puppetlabs/bolt/issues/1722))

  The `target` parameter for YAML plan steps has been deprecated in
  favor of `targets` and will be removed in a future release of Bolt.


lib/bolt/pal/yaml_plan/evaluator.rb
```diff
@@ -174,6 +176,17 @@ def evaluate_code_blocks(scope, value)
         def evaluate(value, _scope)
           value
         end
+
+        # TODO: Remove this method once 'target' parameter is removed
+        def targets(type, step)
+          if step.key?('target')
```
nicklewis, line 182: Not sure where to attach this comment, but it might be nicer to handle this warning in `evaluate_block_with_bindings` so that you only get the warning once even if multiple steps refer to `target`. We could just check `plan.steps.any? { |step| step.key?('target') }` and issue a generic warning.

lib/bolt/pal/yaml_plan/evaluator.rb
```diff
@@ -127,16 +129,22 @@ def generate_manifest(resources)
...
           apply_block = ast.body.body
           applicator = Puppet.lookup(:apply_executor)
-          applicator.apply([target], apply_block, scope)
+          applicator.apply([targets], apply_block, scope)
         end
 
         # This is the method that Puppet calls to evaluate the plan. The name
         # makes more sense for .pp plans.
         def evaluate_block_with_bindings(closure_scope, args_hash, plan)
+          unless plan.steps.any? { |step| step.body.key?('targets') }
```
nicklewis, line 142: This will skip the warning if any step is using `targets`. I think we actually want to check if any step _is_ using `target` (since some plan could have both a step with `target` and a step with `targets` and we still want to warn).

beechtom: Derp. Definitely a typo from being used to typing targets.

beechtom: Derp. Definitely a case of the Mondays.

PR #1748, opened 2020-04-15
(GH-1725) Target facts should take precedence in apply

### LINKED ISSUE #1725: Catalog compile fails when variable in plan scope has the same name as a Target fact
## Describe the Bug
If in the plan scope (outside of an apply block) a variable exists that happens to have the same name as a fact on the Target you are targeting for apply the apply block fails to compile. 

## Steps to Reproduce
```
plan test(TargetSpec $targets){
  $target = get_target($targets)
  $foo = 'bar'
  $target.add_facts('foo' => 'bar')
  return apply($target){
    notify {"$facts":}
  }
}
```

```
cas@cas-ThinkPad-T460p:~/working_dir/control-repo$ bolt plan run test --modulepath site-modules --targets localhost
The documented parameter 'foo' does not exist in plan signature
Starting: plan test
Finished: apply catalog with 1 failure in 1.91 sec
Finished: plan test in 1.91 sec
Failed on localhost:
  Apply failed to compile for localhost: Cannot reassign variable '$foo' on node localhost
Failed on 1 target: localhost
Ran on 1 target
```
## Additional Context
This line is where it fails https://github.com/puppetlabs/bolt/blob/master/lib/bolt/catalog.rb#L75
So it looks like the facts are in the same scope? https://github.com/puppetlabs/puppet/blob/f5d2b330391d706a864357972312ca5cadf8441f/lib/puppet/pal/pal_impl.rb#L373

lucywyman:
If a target has a fact that shares a name with a plan variable, a
target variable, or both, then the fact should take precedence when
passed in to apply. Previously, defining a fact with the same name as
another variable would cause a redefinition error in the apply block,
where now referencing the variable will refer to the fact value.

Closes #1725

!bug

* **Target facts with the same name as a plan or Target variable should
  not raise an error** ([#1725](https://github.com/puppetlabs/bolt/issues/1725))

  Previously, defining a fact with the same name as
  another variable would cause a redefinition error in the apply block,
  where now referencing the variable will refer to the fact value.

### DIFF
--- a/lib/bolt/catalog.rb
+++ b/lib/bolt/catalog.rb
@@ -70,9 +70,38 @@ module Bolt
                           bolt_inventory: inv) do
             Puppet.lookup(:pal_current_node).trusted_data = target['trusted']
             pal.with_catalog_compiler do |compiler|
-              # This needs to happen inside the catalog compiler so loaders are initialized for loading
-              vars = Puppet::Pops::Serialization::FromDataConverter.convert(request['plan_vars'])
-              pal.send(:add_variables, compiler.send(:topscope), target['variables'].merge(vars))
+              # Deserializing needs to happen inside the catalog compiler so
+              # loaders are initialized for loading
+              plan_vars = Puppet::Pops::Serialization::FromDataConverter.convert(request['plan_vars'])
+
+              # Facts will be set by the catalog compiler, so we need to ensure
+              # that any plan or target variables with the same name are not
+              # passed into the apply block to avoid a redefinition error.
+              # Filter out plan and target vars separately and raise a Puppet
+              # warning if there are any collisions for either. Puppet warning
+              # is the only way to log a message that will make it back to Bolt
+              # to be printed.
+              pv_collisions, pv_filtered = plan_vars.partition do |k, _|
+                target['facts'].keys.include?(k)
+              end.map(&:to_h)
+              unless pv_collisions.empty?
+                print_pv = pv_collisions.keys.map { |k| "$#{k}" }.join(', ')
+                plural = pv_collisions.keys.length == 1 ? '' : 's'
+                Puppet.warning("Plan variable#{plural} #{print_pv} will be overridden by fact#{plural} " \
+                               "of the same name in the apply block")
+              end
+
+              tv_collisions, tv_filtered = target['variables'].partition do |k, _|
+                target['facts'].keys.include?(k)
+              end.map(&:to_h)
+              unless tv_collisions.empty?
+                print_tv = tv_collisions.keys.map { |k| "$#{k}" }.join(', ')
+                plural = tv_collisions.keys.length == 1 ? '' : 's'
+                Puppet.warning("Target variable#{plural} #{print_tv} " \
+                               "will be overridden by fact#{plural} of the same name in the apply block")
+              end
+
+              pal.send(:add_variables, compiler.send(:topscope), tv_filtered.merge(pv_filtered))
 
               # Configure language strictness in the CatalogCompiler. We want Bolt to be able
               # to compile most Puppet 4+ manifests, so we default to allowing deprecated functions.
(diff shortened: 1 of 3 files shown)


lib/bolt/catalog.rb
```diff
@@ -71,8 +71,11 @@ def compile_catalog(request)
             Puppet.lookup(:pal_current_node).trusted_data = target['trusted']
             pal.with_catalog_compiler do |compiler|
               # This needs to happen inside the catalog compiler so loaders are initialized for loading
-              vars = Puppet::Pops::Serialization::FromDataConverter.convert(request['plan_vars'])
-              pal.send(:add_variables, compiler.send(:topscope), target['variables'].merge(vars))
+              plan_vars = Puppet::Pops::Serialization::FromDataConverter.convert(request['plan_vars'])
+              merged = target['variables'].merge(plan_vars)
+              merged.reject! { |k, _| target['facts'].keys.include?(k) }
```
nicklewis, line 76: A comment explaining this would be helpful. If I understand correctly, we exclude any variables that share the name of a fact, and then later Puppet will automatically assign variables for each fact. Is that correct or is the order reversed (Puppet has already assigned variables for the facts and we need to avoid assigning them here)?

nicklewis: The original issue also mentions that we want to warn here that a variable in the current scope will be shadowed by a target fact of the same name.

lucywyman: M'k, this should be logged and commented now

lib/bolt/catalog.rb
```diff
@@ -70,9 +70,38 @@ def compile_catalog(request)
...
+                               "of the same name in the apply block")
+              end
+
+              tv_collisions, tv_filtered = target['variables'].partition do |k, _|
+                target['facts'].keys.include?(k)
+              end.map(&:to_h)
+              print_tv = tv_collisions.keys.map { |k| "$#{k}" }.join(', ')
+              unless print_tv.empty?
+                plural = tv_collisions.keys.length == 1 ? '' : 's'
+                Puppet.warning("Target variable#{plural} #{print_tv} for target #{target['name']} " \
```
nicklewis, line 100: Bolt automatically prepends these log messages with the target name, so we don't need that part here.

lib/bolt/catalog.rb
```diff
@@ -70,9 +70,38 @@ def compile_catalog(request)
...
+              # passed into the apply block to avoid a redefinition error.
+              # Filter out plan and target vars separately and raise a Puppet
+              # warning if there are any collisions for either. Puppet warning
+              # is the only way to log a message that will make it back to Bolt
+              # to be printed.
+              pv_collisions, pv_filtered = plan_vars.partition do |k, _|
+                target['facts'].keys.include?(k)
+              end.map(&:to_h)
+              print_pv = pv_collisions.keys.map { |k| "$#{k}" }.join(', ')
+              unless print_pv.empty?
```
beechtom, line 88: A bit nitpicky, but it seems better to check if `pv_collisions` is empty and then create `print_pv` if there are any collisions.

lib/bolt/catalog.rb
```diff
@@ -70,9 +70,38 @@ def compile_catalog(request)
...
+                plural = pv_collisions.keys.length == 1 ? '' : 's'
+                Puppet.warning("Plan variable#{plural} #{print_pv} will be overridden by fact#{plural} " \
+                               "of the same name in the apply block")
+              end
+
+              tv_collisions, tv_filtered = target['variables'].partition do |k, _|
+                target['facts'].keys.include?(k)
+              end.map(&:to_h)
+              print_tv = tv_collisions.keys.map { |k| "$#{k}" }.join(', ')
+              unless print_tv.empty?
```
beechtom, line 98: Same as with `pv_collisions`.

PR #1754, opened 2020-04-21
(GH-1736,1738) Convert pkcs7 plugin to module, add --force option

### LINKED ISSUE #1736: Move PKCS7 plugin to a module
## Use Case
Provide a clear example of how users can write their own encryption plugins by extracting the pkcs7 plugin to a module

## Additional Context
* The base secret class should be deleted as part of this.
* The keys will be renamed to match the docuemented value rather then current implemented value since we cannot support `-` in tasks.

### LINKED ISSUE #1738: protect existing secret keys
## Use Case
Currently, `bolt secret createkeys` command overrides existing secret keys unconditionally and can lead to data loss if ran from wrong project directory.

command should fail if configured keys already exist. To provide existing option to override existing keys `--force`  should be used explicitly. This should turn into a `force` task parameter that gets passed to the createkeys plugin task. 

Tasks should return a standard error kind for this and the CLI should generate a helpful message telling the user to rerun with `--force` if they want to overwrite the existing keys.
The writing plugins docs should be updated to describe how to handle the force.

beechtom:
## Convert pkcs7 plugin to module

This removes the built-in `pkcs7` plugin and converts it to a
module-based plugin.

!feature

* **`pkcs7` plugin converted to module-based plugin**
  ([#1736](https://github.com/puppetlabs/bolt/issues/1736))

  The `pkcs7` plugin has been converted to a module-based plugin and
  includes the `pkcs7::secret_encrypt`, `pkcs7::secret_decrypt`, and
  `pkcs7::secret_createkeys` tasks.

## Add --force option for 'secret createkeys' command

This adds a `--force` CLI option for the `bolt secret createkeys`
command. Previously, use of the command would overwrite existing key
pairs, potentially leading to complete data loss. If the `--force`
option is used, the `bolt secret createkeys` command will pass a `force`
parameter with a value of `true` to the `secret_createkeys` method.

`secret_createkeys` plugins should have the `force` parameter present in
the task parameters file.

!feature

* **Require `--force` option to overwrite existing keys**
  ([#1738](https://github.com/puppetlabs/bolt/issues/1738))

  The `bolt secret createkeys` command now accepts an optional `--force`
  option to force secret plugins to overwrite existing keys. The default
  `pkcs7` secret plugin will now error when attempting to overwrite
  existing keys without the `--force` option set.

## Support default task parameters for plugins

This adds support for passing default task parameters to plugins.
Previously, Bolt would merge parameters set in a `bolt.yaml` and
`inventory.yaml`, but not the default task parameters.

This also updates parameter validation to ignore any unrecognized
parameters set in a `bolt.yaml` for a specific plugin as long as other
plugins in the same module recognize the parameter.

!feature

* **Support default task parameters in plugins**
  ([#1754](https://github.com/puppetlabs/bolt/pull/1754))

  Bolt now merges default task parameters for a plugin with parameters
  set in a `bolt.yaml` and `inventory.yaml` file.

Closes #1736 
Closes #1738 

### TODO

- [x] Review and merge https://github.com/puppetlabs/puppetlabs-pkcs7/pull/1
- [x] Release `pkcs7` module to the Forge
- [x] Update Bolt's Puppetfile


documentation/writing_plugins.md
```diff
@@ -34,27 +35,31 @@ A plugin can implement hooks as tasks. Bolt will search for tasks in a plugin mo
 
 In this case the plugin will implement two hooks: the `resolve_reference` hook that is explicitly defined and the `secret_decrypt` hook that is discovered from the task name. Bolt passes two metaparameters to all task hooks: `_config`, which contains the plugin configuration, and `_boltdir`, which contains the path to the current Boltdir.
 
-**`resolve_reference` tasks**
+### `resolve_reference` tasks
```
nicklewis, line 38: `###` will cause these to be included in the table of contents. Is that desired? It definitely renders quite poorly without this change at least...

lib/bolt/plugin/module.rb
```diff
@@ -155,7 +155,17 @@ def process_params(task, opts)
         # handled previously. That may not always be the case so filter them
         # out now.
         meta, params = opts.partition { |key, _val| key.start_with?('_') }.map(&:to_h)
-        params = config.merge(params)
+
+        # Reject parameters from config that are not accepted by the task and
+        # merge in parameter defaults
+        params = if task.parameters
+                   config.slice(*task.parameters.keys)
+                         .merge(task.parameter_defaults)
```
nicklewis, line 163: This order seem backward. `config` should override `parameter_defaults`, right?

lib/bolt/secret.rb
```diff
@@ -1,17 +1,37 @@
...
     def self.execute(plugins, outputter, options)
-      plugin = options[:plugin] || 'pkcs7'
+      name   = options[:plugin] || 'pkcs7'
+      plugin = plugins.by_name(name)
+
+      unless plugin
+        raise Bolt::Plugin::PluginError::Unknown, name
+      end
+
+      opts = plugin.config.slice(*KNOWN_KEYS[options[:action]])
```
nicklewis, line 21: I'm not sure I follow this. Shouldn't the plugin system already deal with pulling out the desired plugin config? It seems like it's `options` we ought to be pulling keys out of (and renaming) here.

nicklewis: Okay yeah it looks like this causes the CLI flags for `--private-key`, etc to be ignored.

documentation/writing_plugins.md
beechtom: Alternatively we can use `####`, but I think that's equivalent to just bolding... I'm not sure there's any harm in having this in the ToC, though, especially since it's fairly small for this page.

lib/bolt/plugin/module.rb
beechtom: Yeeep. Updated!

documentation/writing_plugins.md
hestonhoffman: Just a heads up - `####` would cause the docs build to break unless it was preceded with a `###` - Just a weird DITA-OT idiosyncrasy.

beechtom: Oh, good to know! 👍

nicklewis:
This changes the key names from using hyphens to underscores which would be a breaking change. We do document that bundled content may have breaking changes, but it's unclear whether a) this is bundled content since it wasn't a plugin before, b) those breaking changes extend to the config.

lib/bolt/plugin/module.rb
```diff
@@ -30,6 +30,19 @@ def initialize(mod:, context:, config:, **_opts)
...
+          %w[private-key public-key].each do |key|
+            next unless @config.key?(key)
+            fixed_key = key.gsub('-', '_')
+            @config[fixed_key] = @config.delete(key)
+
+            message = "pkcs7 key #{key} is deprecated and will be removed in a future version "\
+                      "of Bolt; use #{fixed_key} instead."
+            Logging.logger[self].warn(message)
+          end
+        end
```
beechtom, line 45: There are a few different places this can happen, including `Bolt::Plugin.add_module_plugin` and `Bolt::Plugin.config_for_plugin`. I stuck it here since it seemed to be the most-specific to the plugin itself before validating the config.

nicklewis: Can these be passed to a specific plugin invocation? What happens in that case?

beechtom: Ah, that will still error out. Updated to handle it in both locations.

documentation/writing_plugins.md
```diff
@@ -34,27 +35,31 @@ A plugin can implement hooks as tasks. Bolt will search for tasks in a plugin mo
...
 
-**`secret_createkeys` tasks**
+### `secret_createkeys` tasks
 
-Bolt passes no parameters other than the metaparameters to a `createkeys` task. It is expected to create the keys based on its `_config` and the `_boltdir` metaparameter.
+Bolt passes a single parameter `force` to a `secret_createkeys` task. When running the `bolt secret createkeys`
+command, the `--force` option can be used to set the `force` parameter's value to `true`.
 
-**`puppet_library` tasks**
+Bolt passes no additional parameters other than the metaparameters to a `secret_createkeys` task. It is expected
```
nicklewis, line 59: This part is now incorrect (and in direct conflict with the newly added bit just above it)

PR #1761, opened 2020-04-23
(GH-1755) Add `prompt` plan function

### LINKED ISSUE #1755: Support prompting for user feedback from an interactive plan
## Use Case
To let the user decide whether the plan should continue after a failure or canary deploy
To build a more guided UI experience from user answers.
Displaying discovered information and letting the user make a choice.

A `prompt` function that prints a message for the user and collects a response string from stdin.
Questions:
- Should we allow an empty string as the message for the prompt? Yes
- Should we allow users to submit an empty string to the prompt? Yes
- Should we error or warn if there is no tty? For now, we should error if stdin is not a tty

Currently, users can prompt by writing their own ruby function or using the `prompt` plugin and `resolve_references`. These options are clunky and tightly coupled with the Bolt CLI tool rather than the plan language as a whole. 

## Future improvements
* Default values
* Choices`prompt::choices`
* Yes/No
* non-stdin input.

beechtom:
This adds a `prompt` plan function that displays a user-specified prompt
message and accepts a response from `STDIN`. The function pauses
execution of a plan until a response is provided and returns the
response as a `String`, or a `Sensitive` type if the `sensitive => true`
option is set.

Closes #1755 

!feature

* **Add `prompt` plan function**
  ([#1755](https://github.com/puppetlabs/bolt/issues/1755))

  The new `prompt` plan function lets you pause plan execution and
  prompt the user for input.

### DIFF
--- a/bolt-modules/prompt/lib/puppet/functions/prompt.rb
+++ b/bolt-modules/prompt/lib/puppet/functions/prompt.rb
@@ -1,0 +1,43 @@
+# frozen_string_literal: true
+
+require 'bolt/error'
+
+# Display a prompt and wait for a response.
+#
+# > **Note:** Not available in apply block
+Puppet::Functions.create_function(:prompt) do
+  # @param prompt The prompt to display.
+  # @param options A hash of additional options.
+  # @option options [Boolean] sensitive Disable echo back and mark the response as sensitive.
+  # @return The response to the prompt.
+  # @example Prompt the user if plan execution should continue
+  #   $response = prompt('Continue executing plan? [Y\N]')
+  # @example Prompt the user for sensitive information
+  #   $password = prompt('Enter your password', 'sensitive' => true)
+  dispatch :prompt do
+    param 'String', :prompt
+    optional_param 'Hash[String[1], Any]', :options
+    return_type 'Variant[String, Sensitive]'
+  end
+
+  def prompt(prompt, options = {})
+    unless Puppet[:tasks]
+      raise Puppet::ParseErrorWithIssue
+        .from_issue_and_stack(Bolt::PAL::Issues::PLAN_OPERATION_NOT_SUPPORTED_WHEN_COMPILING,
+                              action: 'prompt')
+    end
+
+    options = options.transform_keys(&:to_sym)
+
+    executor = Puppet.lookup(:bolt_executor)
+    executor.report_function_call(self.class.name)
+
+    response = executor.prompt(prompt, options)
+
+    if options[:sensitive]
+      Puppet::Pops::Types::PSensitiveType::Sensitive.new(response)
+    else
+      response
+    end
+  end
+end
--- a/lib/bolt/executor.rb
+++ b/lib/bolt/executor.rb
@@ -326,6 +326,23 @@ module Bolt
       end
     end
 
+    def prompt(prompt, options)
+      unless STDIN.tty?
+        raise Bolt::Error.new('STDIN is not a tty, unable to prompt', 'bolt/no-tty-error')
+      end
+
+      STDERR.print("#{prompt}: ")
+      value = if options[:sensitive]
+                STDIN.noecho(&:gets).chomp
+              else
+                STDIN.gets.chomp
+              end
+
+      STDERR.puts if options[:sensitive]
+
+      value
+    end
+
     # Plan context doesn't make sense for most transports but it is tightly
     # coupled with the orchestrator transport since the transport behaves
     # differently when a plan is running. In order to limit how much this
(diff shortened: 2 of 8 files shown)


bolt-modules/prompt/lib/puppet/functions/prompt.rb
```diff
@@ -0,0 +1,47 @@
...
+  # @param prompt The prompt to display.
+  # @param options A hash of additional options.
+  # @option options [Boolean] sensitive Disable echo back and mark the response as sensitive.
+  # @return The response to the prompt.
+  # @example Prompt the user if plan execution should continue
+  #   $response = prompt('Continue executing plan? [Y\N]')
+  # @example Prompt the user for sensitive information
+  #   $password = prompt('Enter your password', 'sensitive' => true)
+  dispatch :prompt do
+    optional_param 'String[1]', :prompt
```
nicklewis, line 19: I think it's reasonable to make this required. This is the `prompt` function, so one would naturally expect that you must provide a `prompt`. We can make it `String` instead of `String[1]` though to allow for empty prompts. It's also odd that the default value (`''`) isn't valid based on this type specification.

lib/bolt/executor.rb
```diff
@@ -326,6 +326,18 @@ def wait_until(timeout, retry_interval)
       end
     end
 
+    def prompt(prompt = '', options = {})
```
nicklewis, line 329: At this point, at the very least `prompt` should be required if not `options` (I would prefer it, but we don't make it required anywhere else in this class so...)

bolt-modules/prompt/lib/puppet/functions/prompt.rb
```diff
@@ -0,0 +1,47 @@
...
+
+    unless STDIN.tty?
+      issue = Puppet::Pops::Issues.issue(:NO_TTY) { 'STDIN is not a tty, unable to prompt' }
+      raise Puppet::ParseErrorWithIssue.from_issue_and_stack(issue, action: 'prompt')
+    end
+
+    executor = Puppet.lookup(:bolt_executor)
+    executor.report_function_call(self.class.name)
+
+    response = executor.prompt(prompt, options)
```
nicklewis, line 39: Is the purpose of delegating to the `executor` here so that the function can fail in PE without having to modify the Puppet side?

If that's the case, then we should move the above `tty?` check into the executor method, otherwise that will always be the error you get in PE. I guess then it shouldn't be a `Puppet` error.

bolt-modules/prompt/lib/puppet/functions/prompt.rb
```diff
@@ -0,0 +1,47 @@
+# frozen_string_literal: true
+
+require 'bolt/error'
+
+# Display a prompt and wait for a response.
+#
+# > **Note:** Not available in apply block
+Puppet::Functions.create_function(:prompt) do
+  # Wait until targets are available.
```
nicklewis, line 9: Not quite.

lib/bolt/executor.rb
```diff
@@ -326,6 +326,18 @@ def wait_until(timeout, retry_interval)
...
+      STDERR.print("#{prompt}: ")
+      value = if options['sensitive']
+                STDIN.noecho(&:gets).chomp
+              else
+                STDIN.gets.chomp
+              end
+      STDERR.puts
```
nicklewis, line 336: This is only needed in the `sensitive` case. In the normal case, the user's newline will be echoed back to them.

bolt-modules/prompt/lib/puppet/functions/prompt.rb
beechtom: Heh, whoops. Definitely a leftover from a copy/paste. :P

bolt-modules/prompt/lib/puppet/functions/prompt.rb
beechtom: Didn't realize that `[1]` was a parameter to the type... time to look over the types docs. :)

bolt-modules/prompt/lib/puppet/functions/prompt.rb
```diff
@@ -0,0 +1,41 @@
...
+    unless Puppet[:tasks]
+      raise Puppet::ParseErrorWithIssue
+        .from_issue_and_stack(Bolt::PAL::Issues::PLAN_OPERATION_NOT_SUPPORTED_WHEN_COMPILING,
+                              action: 'prompt')
+    end
+
+    executor = Puppet.lookup(:bolt_executor)
+    executor.report_function_call(self.class.name)
+
+    response = executor.prompt(prompt, options)
```
nicklewis, line 33: We should keywordize options at the start of this function and particularly before passing them into the executor, as we do with other functions.

PR #1763, opened 2020-04-23
(maint) Pin net-ssh to 5.2.0

beechtom:
The recent release of net-ssh 6.0 has caused issues with corrupted
HMACs, resulting in some tests occasionally failing. This pins net-ssh
to the most recent working version.

Related issue: https://github.com/net-ssh/net-ssh/issues/728

!no-release-note

### DIFF
--- a/bolt.gemspec
+++ b/bolt.gemspec
@@ -44,7 +44,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency "logging", "~> 2.2"
   spec.add_dependency "minitar", "~> 0.6"
   spec.add_dependency "net-scp", "~> 1.2"
-  spec.add_dependency "net-ssh", ">= 4.0"
+  spec.add_dependency "net-ssh", [">= 4", "< 6"]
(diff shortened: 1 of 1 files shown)


bolt.gemspec
```diff
@@ -44,7 +44,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency "logging", "~> 2.2"
   spec.add_dependency "minitar", "~> 0.6"
   spec.add_dependency "net-scp", "~> 1.2"
-  spec.add_dependency "net-ssh", ">= 4.0"
+  spec.add_dependency "net-ssh", "~> 5.2"
```
donoghuc, line 47: I'm not sure if this is still a thing w/ beaker https://github.com/puppetlabs/bolt/blob/c57dd5663145358da4d919900e53dd789637bdfb/lib/bolt/transport/ssh/connection.rb#L60-L61 but we were at one point trying to keep compatabilty w/ netssh 4.

beechtom:
The corrupted hmac issue seems to have been fixed in net-ssh 6.0.1.

PR #1764, opened 2020-04-24
(GH-1756) Project authors can whitelist bolt * show output

### LINKED ISSUE #1756: I want to limit which parts of my project are displayed to users.
## Use Case
When building a project especially one that consumes modules from the forge `bolt task show` and `bolt plan show` can get cluttered with content that is not important to the projects users. In cases where the author of a project is different from the user of a project, there may be content that isn't relevant to users that shouldn't be listed.

A `project.yaml` file that allows the project author to whitelist public content that should be displayed with `bolt * show`. 

Using public and private at a module level. These methods are about the public API of a library rather than an application that will be handed to users.

## Questions
* Should the presence of `project.yaml` be enough to make a directory a project? Not yet
* Should `project_name` move to just `name` in `project.yaml`? Yes
* Bikeshed schema:
```
---
name: myproject
tasks:
  - facts
  - service::mytask
plans:
  - myplan::foo
```
* Do we need to provide options to expose all content? If there isn't a tasks key, show all available tasks. If there isn't a plans key, show all available plans.
* Should this just whitelist individual tasks/plans, or whole modules? For now, we should do individual content, eventually may be able to add `module::*` or something
* Does the whitelist inherently include builtin bolt content? No, that should also be listed.

This should be labelled as experimental.

lucywyman:
Projects that consume modules from the forge often include Bolt content
that isn't relevant to project users, which clutters `bolt [plan|task] show`
output and is confusing to project users. Project authors can now 
whitelist what Bolt content is shown to project users. This is
accomplished by creating a new `project.yaml` file at the root of the 
Bolt project which sets `tasks` or `plans` to lists of individual tasks
and plans that will be shown when users run `bolt * show` inside the 
project.

Closes #1756

!feature

* **Project authors can whitelist bolt * show output** ([#1756](https://github.com/puppetlabs/bolt/issues/1756))

  Project authors can now whitelist individual `bolt [plan|task] show`
  content in `project.yaml` using the `tasks` and `plans` settings.


documentation/bolt_configuration_reference.md.erb
```diff
@@ -80,6 +80,18 @@ plugin_hooks:
       message: "Which version of Puppet do you want to install?"
 ```
 
+## Project configuration options
+
+> **Note:** Project configuration is experimental
+
+Bolt projects are configured in `Boltdir/project.yaml`, and can define the following settings:
```
hestonhoffman, line 87:
```suggestion
Use `<PROJECT_NAME>/project.yaml` to whitelist the tasks or plans that are displayed by the `bolt [plan|task] show` command. You can define the following settings:
```

hestonhoffman: I know this narrows the focus a bit, but we can always go back and edit this if/when we expand the project.yaml functionality.

documentation/bolt_project_directories.md
```diff
@@ -89,6 +90,7 @@ The default paths for all Bolt configuration, code, and data are relative to the
 |[`bolt.yaml`](bolt_configuration_reference.md)|Contains configuration options for Bolt.|
 |`hiera.yaml`|Contains the Hiera config to use for target-specific data when using `apply`.|
 |[`inventory.yaml`](inventory_file_v2.md)|Contains a list of known targets and target specific data.|
+|[`project.yaml`](bolt_configuration_reference.md#project_configuration_options)|Contains configuration for the Bolt project.|
```
hestonhoffman, line 93:
```suggestion
|[`project.yaml`](bolt_configuration_reference.md#project_configuration_options)|Contains configuration for the Bolt project. The project.yaml file contains a whitelist of tasks and plans that you can use to limit the output from the `bolt [plan|task] show` command.|
```

lib/bolt/cli.rb
```diff
@@ -446,7 +447,13 @@ def show_task(task_name)
...
     def list_tasks
       tasks = pal.list_tasks
-      tasks.select! { |task| task.first.include?(options[:filter]) } if options[:filter]
+      if options[:filter] || config.project.tasks.any?
+        tasks.select! do |task|
+          filter = options[:filter] ? task.first.include?(options[:filter]) : true
+          in_project = config.project.tasks.any? ? config.project.tasks.include?(task.first) : true
+          filter && in_project
+        end
+      end
```
beechtom, line 456: I think just adding a second `select!` keeps things a little more readable and avoids the need for the ternary operators.

```
tasks.select! { |task, _| config.project.tasks.include?(task) } if config.project.tasks.any?
```

Even with a list of thousands of tasks a second iteration goes quickly.

lib/bolt/project.rb
```diff
@@ -3,19 +3,23 @@
 require 'pathname'
 
 module Bolt
-  class Boltdir
+  class Project
```
beechtom, line 6: Hooray! :D

lib/bolt/cli.rb
lucywyman: I thought traversing the list twice would be a lot slower, but the time diff with ~600 tasks is only 0.5 seconds, which is pretty negligible.

lib/bolt/cli.rb
```diff
@@ -456,7 +458,13 @@ def show_plan(plan_name)
...
     def list_plans
       plans = pal.list_plans
-      plans.select! { |plan| plan.first.include?(options[:filter]) } if options[:filter]
+      if options[:filter] || config.project.plans.any?
+        plans.select! do |plan|
+          filter = options[:filter] ? plan.first.include?(options[:filter]) : true
+          in_project = config.project.plans.any? ? config.project.plans.include?(plan.first) : true
+          filter && in_project
+        end
+      end
```
beechtom, line 467: Same here as with `list_tasks`

lucywyman: Oh my gosh my brain has clearly left for the weekend.

lib/bolt/cli.rb
```diff
@@ -447,6 +448,7 @@ def show_task(task_name)
     def list_tasks
       tasks = pal.list_tasks
       tasks.select! { |task| task.first.include?(options[:filter]) } if options[:filter]
+      tasks.select! { |task| config.project.tasks.include?(task.first) } if config.project.tasks.any?
```
nicklewis, line 451: This means there's no way\* to have a project that only includes tasks _or_ plans but not both, ie. there's no way to specify that you _want_ a whitelist but that it's empty.

\* There technically is a hacky way which is to make a whitelist that contains a task or plan that doesn't actually exist

lucywyman: Setting this to `tasks: []` should correctly show no tasks in `bolt task show`, while `tasks: ` throws an error.

lib/bolt/project.rb
```diff
@@ -0,0 +1,78 @@
...
+    def eql?(other)
+      path == other.path
+    end
+    alias == eql?
+
+    def tasks
+      @data['tasks']
+    end
+
+    def plans
```
nicklewis, line 66: It might make more sense to call these explicitly `task_whitelist` and `plan_whitelist`. That makes their intention clearer when we actually use them to filter.

PR #1767, opened 2020-04-25
(GH-1768) Add 'optional' and 'default' opts

op-ct:
This patch adds two (optional) opts for the `env_var` plugin:
* `default` provides a value to use when the environment variable `var` isn't set
* `optional` allows the env_var to return nil instead of failing when the environment variable `var` isn't set

### DIFF
--- a/documentation/using_plugins.md
+++ b/documentation/using_plugins.md
@@ -49,7 +49,9 @@ The following fields are available to the `env_var` plugin.
 
 | Key | Description | Type | Default |
 | --- | ----------- | ---- | ------- |
-| `var` | The name of the environment variable to read from. | `String` | None |
+| **`var`** | The name of the environment variable to read from. | `String` | None |
+| **`default_value`** | (Optional) A value to use if the environment variable `var` isn't set | `String` | None |
+| **`optional`** | (Optional) Unless `true`, env_var will raise an error when the environment variable `var` does not exist.  When `optional` is `true` and `var` does not exist, env_var returns `nil` | `Boolean` | `false` |
 
 #### Example usage
 
--- a/documentation/using_plugins.md
+++ b/documentation/using_plugins.md
@@ -51,7 +51,7 @@ The following fields are available to the `env_var` plugin.
 | --- | ----------- | ---- | ------- |
 | **`var`** | The name of the environment variable to read from. **Required.** | `String` | None |
 | **`default_value`** | (Optional) A value to use if the environment variable `var` isn't set | `String` | None |
-| **`optional`** | (Optional) Unless `true`, env_var will raise an error when the environment variable `var` does not exist.  When `optional` is `true` and `var` does not exist, env_var returns `nil` | `Boolean` | `false` |
+| **`optional`** | Unless `true`, env_var will raise an error when the environment variable `var` does not exist.  When `optional` is `true` and `var` does not exist, env_var returns `nil` | `Boolean` | `false` |
 
 #### Example usage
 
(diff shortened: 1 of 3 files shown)


documentation/using_plugins.md
```diff
@@ -49,7 +49,9 @@ The following fields are available to the `env_var` plugin.
 
 | Key | Description | Type | Default |
 | --- | ----------- | ---- | ------- |
-| `var` | The name of the environment variable to read from. | `String` | None |
+| **`var`** | The name of the environment variable to read from. | `String` | None |
```
beechtom, line 52:
```suggestion
| **`var`** | The name of the environment variable to read from. **Required.** | `String` | None |
```

documentation/using_plugins.md
```diff
@@ -49,7 +49,9 @@ The following fields are available to the `env_var` plugin.
...
+| **`default_value`** | (Optional) A value to use if the environment variable `var` isn't set | `String` | None |
```
beechtom, line 53:
```suggestion
| **`default`** | A value to use if the environment variable `var` isn't set | `String` | None |
```

documentation/using_plugins.md
```diff
@@ -49,7 +49,9 @@ The following fields are available to the `env_var` plugin.
...
+| **`optional`** | (Optional) Unless `true`, env_var will raise an error when the environment variable `var` does not exist.  When `optional` is `true` and `var` does not exist, env_var returns `nil` | `Boolean` | `false` |
```
beechtom, line 54:
```suggestion
| **`optional`** | Unless `true`, env_var will raise an error when the environment variable `var` does not exist.  When `optional` is `true` and `var` does not exist, env_var returns `nil` | `Boolean` | `false` |
```

lib/bolt/plugin/env_var.rb
```diff
@@ -17,13 +17,14 @@ def validate_resolve_reference(opts)
...
         end
+        return if opts['optional']
         unless ENV[opts['var']]
           raise Bolt::ValidationError, "env_var plugin requires that the var '#{opts['var']}' be set"
         end
       end
 
       def resolve_reference(opts)
-        ENV[opts['var']]
+        ENV[opts['var']] || opts['default_value']
```
beechtom, line 27:
```suggestion
        ENV[opts['var']] || opts['default']
```

spec/bolt/plugin/env_var_spec.rb
```diff
@@ -25,4 +25,32 @@
...
+    end
+
+    it 'returns undef when the var is not set' do
+      ENV.delete('BOLT_ENV_VAR')
+      expect { subject.validate_resolve_reference(env_var_data).to be_nil }
+    end
+  end
+
+  describe 'when default_value is provided' do
+    let(:env_var_data) { super().merge({ 'default_value' => 'DEFAULT_STRING' }) }
```
beechtom, line 44:
```suggestion
  describe 'when default is provided' do
    let(:env_var_data) { super().merge({ 'default' => 'DEFAULT_STRING' }) }
```

lib/bolt/plugin/env_var.rb
```diff
@@ -17,13 +17,14 @@ def validate_resolve_reference(opts)
         unless opts['var']
           raise Bolt::ValidationError, "env_var plugin requires that the 'var' is specified"
         end
+        return if opts['optional']
```
beechtom, line 20:
```suggestion
        return if opts['optional'] || opts['default']
```

PR #1770, opened 2020-04-27
(maint) Make inventory_config optional

donoghuc:
The code in the constructor for an `ApplyTarget` seems to suggest that config can be an empty hash. Similarly most instance variable declarations provide defaults. The `t_conf` variable assumes that the config hash has a `transport` config section. In practice the defaults for a bolt inventory/config ensure there is always data for config. In the case where a plan runner explicitly does not want to expose target connection information is it reasonable to just provide an empty hash?

### DIFF
--- a/lib/bolt/apply_target.rb
+++ b/lib/bolt/apply_target.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require 'addressable'
+
 module Bolt
   class ApplyTarget
     ATTRIBUTES = %i[uri name target_alias config vars facts features
@@ -48,7 +50,7 @@ module Bolt
       # Merge the config hash with inventory config
       config = Bolt::Util.deep_merge(config, @config || {})
       transport = config['transport'] || 'ssh'
-      t_conf = config['transports'][transport] || {}
+      t_conf = config.dig('transports', transport) || {}
       uri_obj = parse_uri(uri)
       @host = uri_obj.hostname || t_conf['host']
       @password = Addressable::URI.unencode_component(uri_obj.password) || t_conf['password']
--- a/lib/bolt/apply_inventory.rb
+++ b/lib/bolt/apply_inventory.rb
@@ -1,8 +1,7 @@
 # frozen_string_literal: true
 
-require 'bolt/project'
-require 'bolt/config'
 require 'bolt/error'
+require 'bolt/apply_target'
 
 module Bolt
   class ApplyInventory


lib/bolt/apply_target.rb
```diff
@@ -48,7 +50,7 @@ def initialize(target_hash, config)
       # Merge the config hash with inventory config
       config = Bolt::Util.deep_merge(config, @config || {})
```
donoghuc, line 51: This made me thing that empty hash should be valide. https://github.com/puppetlabs/bolt/blob/62e3e66468502763fc699b6170393f984d8f9ce3/lib/bolt/apply_inventory.rb#L18 Where does `@config` come from?

donoghuc:
@puppetlabs/bolt Should i just convert this to an issue? Is this something we could potentially do? Context: https://github.com/puppetlabs/puppetserver/pull/2329/files#diff-73104a90aa2b4d7f02119c3f544b30c6R47-R54

lib/bolt/apply_target.rb
nicklewis: It comes from the `instance_variable_set` loop above.

nicklewis:
There are parts of Bolt that assert that a Target has a transport and that it's a valid transport. I'm a little concerned about changing _only_ this line of code because I worry that will invalidate an assumption made elsewhere.

Starting from an explanation of what is needed and what the problem is would help.

lib/bolt/apply_target.rb
donoghuc: Sorry yeah that was a vague question. If i understand correctly `@config` is something set on a Target's config section (either in a plan with a target method or on a target's config section in inventory) and the `config_hash` is "global" config set on inventory.yaml/bolt.yaml (ie *not*) set on an individual Target object.

donoghuc:
I am working on updating the puppetserver compile endpoint to support compiling plans that contain Bolt objects (specifically types in the boltlib module). Part of this work includes making an ApplyInventory in the compile endpoint to support ApplyTargets in apply blocks. Given we block read/write of Target config in PE plans (this is all handled with pe inventory service) we do not seem to need to set any target config information in apply blocks. 

I needed to supply *some* data when construct a new ApplyInventory. I want to figure out what that data should be.

lib/bolt/apply_target.rb
```diff
@@ -48,7 +50,7 @@ def initialize(target_hash, config)
...
       transport = config['transport'] || 'ssh'
-      t_conf = config['transports'][transport] || {}
+      t_conf = config.dig('transports', transport) || {}
```
nicklewis, line 53: Since we don't store the value of `t_conf` anywhere, I think this change is fine. We just use it to check whether certain properties that aren't set in the URI are set on the transport. In this use case, they'll all just be nil and that will be that.

donoghuc:
Weird, in draft mode it was not showing the merge conflict. I was trying to figure out why tests were not triggering so i converted it from draft mode... I'm still working on this.

donoghuc:
Closing this for now, I expect we will work on something like it as part of https://tickets.puppetlabs.com/browse/PE-29095

Issue #1759, opened 2020-04-23
When a task outputs data that cannot be serialized to JSON the task will stacktrace


donoghuc:
## Describe the Bug
When a task result cannot be serialized to JSON in https://github.com/puppetlabs/bolt/blob/6851fc8b988c87606f13fc769b9357c5dddabb6f/lib/bolt/result.rb#L141-L143 there is a stacktrace:
```
cas@cas-ThinkPad-T460p:~/working_dir/bolt/Boltdir/site/test/tasks$ bolt task run test::big -t localhost
Started on localhost...
<JSON::GeneratorError> source sequence is illegal/malformed utf-8
	/home/cas/working_dir/bolt/lib/bolt/result.rb:142:in `to_json'
	/home/cas/working_dir/bolt/lib/bolt/result.rb:142:in `to_json'
	/home/cas/working_dir/bolt/lib/bolt/executor.rb:237:in `with_node_logging'
	/home/cas/working_dir/bolt/lib/bolt/executor.rb:274:in `block (2 levels) in run_task'
	/home/cas/working_dir/bolt/lib/bolt/executor.rb:115:in `block (3 levels) in queue_execute'
	/home/cas/working_dir/bolt/vendor/bundle/ruby/2.5.0/gems/concurrent-ruby-1.1.6/lib/concurrent-ruby/concurrent/executor/ruby_thread_pool_executor.rb:353:in `run_task'
	/home/cas/working_dir/bolt/vendor/bundle/ruby/2.5.0/gems/concurrent-ruby-1.1.6/lib/concurrent-ruby/concurrent/executor/ruby_thread_pool_executor.rb:342:in `block (3 levels) in create_worker'
	/home/cas/working_dir/bolt/vendor/bundle/ruby/2.5.0/gems/concurrent-ruby-1.1.6/lib/concurrent-ruby/concurrent/executor/ruby_thread_pool_executor.rb:325:in `loop'
	/home/cas/working_dir/bolt/vendor/bundle/ruby/2.5.0/gems/concurrent-ruby-1.1.6/lib/concurrent-ruby/concurrent/executor/ruby_thread_pool_executor.rb:325:in `block (2 levels) in create_worker'
	/home/cas/working_dir/bolt/vendor/bundle/ruby/2.5.0/gems/concurrent-ruby-1.1.6/lib/concurrent-ruby/concurrent/executor/ruby_thread_pool_executor.rb:324:in `catch'
[...]

PR #1778, opened 2020-04-28
(GH-1759) Handle non-UTF-8 results

nicklewis:
Previously, we treated non-UTF-8 task output as simply non-JSON output,
putting it in a hash under the `_output` key. This violated a key
constraint of the task specification (task output must always be UTF-8)
and caused problems when we tried to print the result as JSON (either as
output or for logging).

We now explicitly check for non-UTF-8 characters when interpreting the
result and we throw away stdout entirely and replace it with an error if
it is invalid.

Non-UTF-8 characters are generally allowed from commands and scripts.
However, they still can't be printed or converted to JSON. We now define a
"safe_value" that is the scrubbed version of the input and use that for
printing and JSON.

This commit removes the status_hash method and replaces it in all cases with
the to_data method which is otherwise identical except for having string
keys. The to_data method now returns the safe_value, as to_data is only
used to print the result, deal with it as JSON equivalent in Puppet, or
emit it as JSON. Each of those cases requires it to be UTF-8 only.

!bug

* **Task output that contains invalid UTF-8 is now rejected**
  ([#1759](https://github.com/puppetlabs/bolt/issues/1759))

  Tasks are defined as returning UTF-8, but Bolt didn't handle the
  non-UTF-8 case explicitly, leading to messy error messages and stack
  traces. The error should now be clear and meaningful.

### DIFF
--- a/spec/bolt/result_spec.rb
+++ b/spec/bolt/result_spec.rb
@@ -109,5 +109,25 @@ describe Bolt::Result do
       expect(result.value).to eq(obj)
       expect(result.error_hash).to eq(obj['_error'])
     end
+
+    it 'uses the unparsed value of stdout if it is not valid JSON' do
+      stdout = 'just some string'
+      result = Bolt::Result.for_task(target, stdout, '', 0, 'atask')
+      expect(result.value).to eq('_output' => 'just some string')
+    end
+
+    it 'generates an error for binary data' do
+      stdout = "\xFC].\xF9\xA8\x85f\xDF{\x11d\xD5\x8E\xC6\xA6"
+      result = Bolt::Result.for_task(target, stdout, '', 0, 'atask')
+      expect(result.value.keys).to eq(['_error'])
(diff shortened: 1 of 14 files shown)


nicklewis:
I also took the liberty of rewording a couple of error messages for the case where a task exits non-zero with no stdout.

spec/bolt/result_spec.rb
```diff
@@ -109,5 +109,25 @@
...
+
+    it 'generates an error for binary data' do
+      stdout = "\xFC].\xF9\xA8\x85f\xDF{\x11d\xD5\x8E\xC6\xA6"
+      result = Bolt::Result.for_task(target, stdout, '', 0, 'atask')
+      expect(result.value.keys).to eq(['_error'])
+      expect(result.error_hash['msg']).to match(/The task result contained invalid UTF-8/)
+    end
+
+    it 'generates an error for non-UTF-8 output' do
+      stdout = "☃".encode('utf-32')
```
beechtom, line 127: ⛄

PR #1784, opened 2020-05-01
(maint) Stop hacking PAL

donoghuc:
WIP. does this actually work?


lib/bolt/catalog.rb
```diff
@@ -69,40 +95,7 @@ def compile_catalog(request)
...
-              unless tv_collisions.empty?
-                print_tv = tv_collisions.keys.map { |k| "$#{k}" }.join(', ')
-                plural = tv_collisions.keys.length == 1 ? '' : 's'
-                Puppet.warning("Target variable#{plural} #{print_tv} " \
-                               "will be overridden by fact#{plural} of the same name in the apply block")
-              end
-
-              pal.send(:add_variables, compiler.send(:topscope), tv_filtered.merge(pv_filtered))
-
+            pal.with_catalog_compiler(variables: pv_filtered) do |compiler|
```
lucywyman, line 98: I think you want this to be `tv_filtered.merge(pv_filtered)` like the line above, though calling functions in an apply block on those datatypes still fails. I think running `spec/integration/apply_compile_spec:319` in your local setup will tell you if this is working or not, in case developing locally is faster than waiting for travis.

donoghuc: oh god. lol. doing too many things at once... Sorry for being lazy.

lib/bolt/catalog.rb
```diff
@@ -69,40 +95,7 @@ def compile_catalog(request)
           Puppet.override(bolt_pdb_client: pdb_client,
                           bolt_inventory: inv) do
             Puppet.lookup(:pal_current_node).trusted_data = target['trusted']
-            pal.with_catalog_compiler do |compiler|
-              # Deserializing needs to happen inside the catalog compiler so
-              # loaders are initialized for loading
-              plan_vars = Puppet::Pops::Serialization::FromDataConverter.convert(request['plan_vars'])
```
nicklewis, line 75: This is the line of code that makes `Target`, `Result`, etc objects work inside the apply block.

nicklewis: Just passing them into `with_catalog_compiler` doesn't cause them to be rehydrated into Puppet types. They're just passed along as regular hashes.

donoghuc: Oh. I thought the spec tests were passing. I was looking at the wrong cell

donoghuc:
This was not doing what I thought it was. I was looking at the wrong tests in GH actions. I explored a bit trying to see if there was an easy change in PAL, though i didnt really get anywhere. Here is a diff of what I tried in puppet: https://github.com/puppetlabs/puppet/compare/master...donoghuc:PUP-10476 (along with the following on top of this PR):

```diff
cas@cas-ThinkPad-T460p:~/working_dir/bolt$ git diff
diff --git a/Gemfile b/Gemfile
index 778d719a..d4451ba6 100644
--- a/Gemfile
+++ b/Gemfile
@@ -14,6 +14,7 @@ gem "puma", '>= 3.12.0'
 gem "rack", '>= 2.0.5'
 gem "rails-auth", '>= 2.1.4'
 gem "sinatra", '>= 2.0.4'
+gem 'puppet', git: 'https://github.com/donoghuc/puppet.git', branch: 'PUP-10476'
 
 group(:test) do
   gem "beaker-hostgenerator"
diff --git a/lib/bolt/catalog.rb b/lib/bolt/catalog.rb
index c10cb0a7..c7660667 100644
--- a/lib/bolt/catalog.rb
+++ b/lib/bolt/catalog.rb
@@ -89,13 +89,13 @@ module Bolt
         Puppet[:node_name_value] = target['name']
         env_conf = { modulepath: request['modulepath'] || [],
                      facts: target['facts'] || {} }
-        env_conf[:variables] = {}
         Puppet::Pal.in_tmp_environment('bolt_catalog', env_conf) do |pal|
           inv = Bolt::ApplyInventory.new(request['config'])
           Puppet.override(bolt_pdb_client: pdb_client,
                           bolt_inventory: inv) do
             Puppet.lookup(:pal_current_node).trusted_data = target['trusted']
-            pal.with_catalog_compiler(variables: tv_filtered.merge(pv_filtered)) do |compiler|
+            pal.with_catalog_compiler(variables: tv_filtered.merge(pv_filtered),
+                                      deserialize_pcore: true) do |compiler|
               # Configure language strictness in the CatalogCompiler. We want Bolt to be able
               # to compile most Puppet 4+ manifests, so we default to allowing deprecated functions.
               Puppet[:strict] = options['strict'] || :warning
```

nicklewis:
I think just having a public API on the PAL compiler to set variables would be reasonable.

PR #1786, opened 2020-05-05
(GH-1781) Add ResourceInstance data type

### LINKED ISSUE #1781: Resource Datatype Experiment
## Use Case
bolt has a limited ability to interact with the Puppet RAL today. `apply` allows a puppet catalog to be generated and returns a report with events and `get_resources` will discover resources on targets. We'd like to expand this to allow plan authors to build resource based logic directly into their plans to enable a discover, inspect, execute workflow for interacting with resources on remote systems. Building a resource datatype will allow this 

#### Attributes
- Target - Target object required
- Type - A type name not verified but required (doesn’t have to be a real type?)
- Title - required
- observed_state, known_state, state, the last known properties - optional hash default empty
- desired_state parameters and properties optional hash default empty
- Events - optional array default empty

##### Methods or functions
- Add event - start with catalog event schema.
- equality is target + type + title 
- set desired param/prop - single param/prop or hash for shallow merge
- overwrite desired state - hash 
- overwrite observed state - single prop or hash for shallow merge
- set observed property - hash 

### Target functions
- $target.resources() returns hash of resources with resource reference string as key ie ("File[/etc/puppetlabs]")
- `$target.set_resource(ResourceInstance or data hash)`
- `$target.set_resources(array of ResourceInstance or data hash)` creates resource or passes properties to overwrite observed state if a resource of the type and title already exists.

## Questions
- **What should this datatype be called?** 

  There is already an abstract Resource datatype we should avoid conflicting with for now.

- **What happens if you try to set a resource on target A when that resource is for target B?**

  ```
  $resource = ResourceInstance.new($targetB, 'file', '/etc/puppetlabs')
  $targetA.set_resource($resource)
  ```

  Error?

- **When setting a resource on a target with a data hash, should the `target` already be specified?**

  Maybe - this would keep it similar to `ResourceInstance.new`. But requiring the `target` to be present in the hash also seems a little weird when you are directly setting a resource on a target.

- **What should this type look like when returned from a plan? The resource reference? A hash of its attributes?**

  ```
  File[/etc/puppetlabs]
  ```

  ```
  {
    "target" => Target,
    "type" => "File",
    "title" => "/etc/puppetlabs",
    "state" => { ... }
    ...
  }
  ```

  Hash seems most useful?

- **When setting a resource that already exists on a target, should it only overwrite observed state, or also desired state and events? If completely overwriting all attributes, maybe just replace the `ResourceInstance`?**

beechtom:
This adds a new `ResourceInstance` data type to the plan language and a
corresponding `Bolt::ResourceInstance` implementation class.

The `ResourceInstance` data type stores observed state, desired state  and
events for a single resource on a target. Resources are identified by a
combination of their target, type, and title, and can have their observed
state, desired state, and events set and overwritten.

This data type does not modify or interact with resources on a target in
any way.

Closes #1781 

!feature

* **Added `ResourceInstance` data type**
  ([#1781](#1781))

  The new `ResourceInstance` data type is available for use in plans and
  can be used to store the observed state, desired state, and events for
  a target's resource.

!feature

* **Added `set_resources` plan function**
  ([#1781](#1781))

  The `set_resources` plan function can be used to set
  `ResourceInstance`s on a `Target`.

!feature

* **Added `resources` function to `Target` data type**
  ([#1781](#1781))

  `Target` objects have a new `resources` function that can be used to
  return a map of `ResourceInstance`s for the target.

### DIFF
--- a/bolt-modules/boltlib/lib/puppet/datatypes/resourceinstance.rb
+++ b/bolt-modules/boltlib/lib/puppet/datatypes/resourceinstance.rb
@@ -1,0 +1,27 @@
+# frozen_string_literal: true
+
+Puppet::DataTypes.create_type('ResourceInstance') do
+  interface <<-PUPPET
+    attributes => {
+      'target'        => Target,
+      'type'          => Variant[String[1], Type[Resource]],
+      'title'         => String[1],
+      'state'         => Optional[Hash[String[1], Data]],
+      'desired_state' => Optional[Hash[String[1], Data]],
+      'events'        => Optional[Array[Hash[String[1], Data]]]
+    },
+    functions => {
+      add_event               => Callable[[Hash[String[1], Data]], [Hash[String[1], Data]]],
+      set_state               => Callable[[Hash[String[1], Data]], Hash[String[1], Data]],
+      overwrite_state         => Callable[[Hash[String[1], Data]], Hash[String[1], Data]],
+      set_desired_state       => Callable[[Hash[String[1], Data]], Hash[String[1], Data]],
+      overwrite_desired_state => Callable[[Hash[String[1], Data]], Hash[String[1], Data]],
+      reference               => Callable[[], String]
+    }
+  PUPPET
+
+  load_file('bolt/resource_instance')
+  # Needed for Puppet to recognize Bolt::ResourceInstance as a Puppet object when deserializing
+  Bolt::ResourceInstance.include(Puppet::Pops::Types::PuppetObject)
+  implementation_class Bolt::ResourceInstance
+end
--- a/bolt-modules/boltlib/lib/puppet/datatypes/target.rb
+++ b/bolt-modules/boltlib/lib/puppet/datatypes/target.rb
@@ -20,7 +20,8 @@ Puppet::DataTypes.create_type('Target') do
       vars => { type => Optional[Hash[String[1], Data]], kind => given_or_derived },
       facts => { type => Optional[Hash[String[1], Data]], kind => given_or_derived },
       features => { type => Optional[Array[String[1]]], kind => given_or_derived },
-      plugin_hooks => { type => Optional[Hash[String[1], Data]], kind => given_or_derived }
+      plugin_hooks => { type => Optional[Hash[String[1], Data]], kind => given_or_derived },
+      resources => { type => Optional[Hash[String[1], ResourceInstance]], kind => given_or_derived }
     },
     functions => {
       host => Callable[[], Optional[String]],
--- a/bolt-modules/boltlib/lib/puppet/functions/set_resources.rb
+++ b/bolt-modules/boltlib/lib/puppet/functions/set_resources.rb
@@ -1,0 +1,113 @@
+# frozen_string_literal: true
+
+require 'bolt/error'
+
+# Sets one or more resources on a target.
+# For more information about resources, see: https://puppet.com/docs/puppet/latest/lang_resources.html
+#
+# > **Note:** Not available in apply block
+Puppet::Functions.create_function(:set_resources) do
+  # @param target The `Target` object to add resources to. See {get_targets}.
+  # @param resources The resources to set on the target.
+  # @return The added `ResourceInstance` objects.
+  # @example Add multiple resources to a target with an array of `ResourceInstance` objects.
+  #   $resource1 = ResourceInstance.new(
+  #     'target' => $target,
+  #     'type'   => 'file',
+  #     'title'  => '/etc/puppetlabs',
+  #     'state'  => { 'ensure' => 'present' }
+  #   )
+  #   $resource2 = ResourceInstance.new(
+  #     'target' => $target,
+  #     'type'   => 'package',
+  #     'title'  => 'openssl',
+  #     'state'  => { 'ensure' => 'installed' }
+  #   )
+  #   $target.set_resources([$resource1, $resource2])
+  # @example Add resources retrieved with [`get_resources`](#get_resources) to a target.
+  #   $target.apply_prep
+  #   $resources = $target.get_resources(Package).first['resources']
+  #   $target.set_resources($resources)
+  dispatch :set_resources do
+    param 'Target', :target
+    param 'Array[Variant[Hash, ResourceInstance]]', :resources
+    return_type 'Array[ResourceInstance]'
+  end
+
+  # @param target The `Target` object to add resources to. See {get_targets}.
+  # @param resource The resource to set on the target.
+  # @return The added `ResourceInstance` object.
+  # @example Add a single resource to a target with a resource data hash.
+  #   $resource = {
+  #     'type'  => 'file',
+  #     'title' => '/etc/puppetlabs',
+  #     'state' => { 'ensure' => 'present' }
+  #   }
+  #   $target.set_resources($resource)
+  dispatch :set_resource do
+    param 'Target', :target
+    param 'Variant[Hash, ResourceInstance]', :resource
+    return_type 'Array[ResourceInstance]'
+  end
+
+  def set_resources(target, resources)
+    unless Puppet[:tasks]
+      raise Puppet::ParseErrorWithIssue
+        .from_issue_and_stack(
+          Bolt::PAL::Issues::PLAN_OPERATION_NOT_SUPPORTED_WHEN_COMPILING,
+          action: 'set_resources'
+        )
+    end
+
+    inventory = Puppet.lookup(:bolt_inventory)
+    executor  = Puppet.lookup(:bolt_executor)
+    executor.report_function_call(self.class.name)
+
+    inventory_target = inventory.get_target(target)
+
+    resources.uniq.map do |resource|
+      if resource.is_a?(Hash)
+        # ResourceInstance expects a Target object, so either get a specified target from
+        # the inventory or use the target this function was called on.
+        resource_target = if resource.key?('target')
+                            inventory.get_target(resource['target'])
+                          else
+                            inventory_target
+                          end
+
+        # Observed state from get_resources() is under the 'parameters' key
+        resource_state = resource['state'] || resource['parameters']
+
+        init_hash = {
+          'target'        => resource_target,
+          'type'          => resource['type'],
+          'title'         => resource['title'],
+          'state'         => resource_state,
+          'desired_state' => resource['desired_state'],
+          'events'        => resource['events']
+        }
+
+        # Calling Bolt::ResourceInstance.new or Bolt::ResourceInstance.from_asserted_hash
+        # will not perform any validation on the parameters. Instead, we need to use the
+        # Puppet constructor to initialize the object, which will first validate the parameters
+        # and then call Bolt::ResourceInstance.from_asserted_hash internally. To do this we
+        # first need to get the Puppet datatype and then call the new function on that type.
+        type = Puppet::Pops::Types::TypeParser.singleton.parse('ResourceInstance')
+        resource = call_function('new', type, init_hash)
+      end
+
+      unless resource.target == inventory_target
+        file, line = Puppet::Pops::PuppetStack.top_of_stack
+        raise Bolt::ValidationError, "Cannot set resource #{resource.reference} for target "\
+                                     "#{resource.target} on target #{inventory_target}. "\
+                                     "#{Puppet::Util::Errors.error_location(file, line)}"
+      end
+
+      inventory_target.set_resource(resource)
+    end
+  end
+
+  def set_resource(target, resource)
+    set_resources(target, [resource])
+  end
+end
--- a/bolt-modules/boltlib/lib/puppet/functions/set_resources.rb
+++ b/bolt-modules/boltlib/lib/puppet/functions/set_resources.rb
@@ -2,7 +2,8 @@
 
 require 'bolt/error'
 
-# Sets one or more resources on a target.
+# Sets one or more ResourceInstances on a Target. This function does not apply or modify
+# resources on a target.
 #
 # For more information about resources see [the
 # documentation](https://puppet.com/docs/puppet/latest/lang_resources.html).
@@ -13,6 +14,7 @@ require 'bolt/error'
 #
 # > **Note:** Not available in apply block
 Puppet::Functions.create_function(:set_resources) do
+  # Set multiple resources
   # @param target The `Target` object to add resources to. See {get_targets}.
   # @param resources The resources to set on the target.
   # @return The added `ResourceInstance` objects.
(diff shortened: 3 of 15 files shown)


documentation/bolt_types_reference.md
```diff
@@ -91,7 +114,10 @@ The following functions are available to `Target` objects:
 | `plugin_hooks` | `Hash[String, Data]` | The target's `plugin_hooks` [configuration options](bolt_configuration_reference.md#plugin-hooks-configuration-options). ||
 | `port` | `Integer` | The target's connection port. ||
 | `protocol` | `String` | The protocol used to connect to the target. | This is equivalent to the target's `transport`, except for targets using the `remote` transport. For example, a target with the URI `http://example.com` using the `remote` transport would return `http` for the `protocol`. |
+| `resources` | `Hash[String, ResourceInstance]` | The target's resources. | This function does not lookup resources for a target and only returns resources set on a target during a plan run. |
```
hestonhoffman, line 117:
```suggestion
| `resources` | `Hash[String, ResourceInstance]` | The target's resources. | This function does not look up resources for a target and only returns resources set on a target during a plan run. |
```

(I know this is super nitpicky)

documentation/bolt_types_reference.md
```diff
@@ -91,7 +114,10 @@ The following functions are available to `Target` objects:
...
 | `safe_name` | `String` | The target's safe name. Equivalent to `name` if a name was given, or the target's `uri` with any password omitted. ||
+| `set_resource` | `ResourceInstance` | Sets a resource on a target. | If a target has an existing resource with the same type and title, then observed and desired state will be overwritten and events from both resources will be concatenated. |
+| `set_resources` | `Array[ResourceInstance]` | Set multiple resources on a target. | If a target has an existing resource with the same type and title, then observed and desired state will be overwritten and events from both resources will be concatenated. |
```
hestonhoffman, line 120:
```suggestion
| `set_resources` | `Array[ResourceInstance]` | Sets multiple resources on a target. | If a target has an existing resource with the same type and title, then observed and desired state will be overwritten and events from both resources will be concatenated. |
```

documentation/bolt_types_reference.md
```diff
@@ -23,6 +23,29 @@ The following functions are available to `ApplyResult` objects.
 | `target` | `Target` | The target the result is from. |
 | `to_data` | `Hash` | A serialized representation of `ApplyResult`. |
 
+## `ResourceInstance`
+
+`ResourceInstance` objects are used to store the observed and desired state of a target's resource and to track events for the resource. These objects do not modify or interact with a target's resources.
```
lucywyman, line 28: I think it'd be useful to document an example of using this in a plan. I think just in the resource type page would actually be good since this is experimental, but writing_plans.md would also be appropriate

bolt-modules/boltlib/lib/puppet/datatypes/resourceinstance.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+Puppet::DataTypes.create_type('ResourceInstance') do
+  interface <<-PUPPET
+    attributes => {
+      'target'        => Target,
+      'type'          => String[1],
```
lucywyman, line 7: Should this also support `Type[Resource]` like get_resources? https://github.com/puppetlabs/bolt/blob/master/bolt-modules/boltlib/lib/puppet/functions/get_resources.rb#L22

donoghuc:
How will this work in apply blocks? I dont see the pcore serialization methods. I doubt you could `set_*` anything and have it modify the inventory outside the apply block but it seems like you might be able to use `$target.resources` in an apply block?

lib/bolt/inventory/target.rb
```diff
@@ -36,12 +38,58 @@ def initialize(target_data, inventory)
...
+
+      # rubocop:disable Naming/AccessorMethodName
+      def set_resources(resources)
+        unless resources.is_a?(Array)
+          raise Bolt::ValidationError, "Resources must be an Array, got #{resources.class}"
+        end
+
+        resources.map do |resource|
+          set_resource(resource)
+        end.uniq
```
lucywyman, line 89: I think you can call `resources.uniq.map` instead to avoid extra cycles.

bolt-modules/boltlib/lib/puppet/functions/set_resources.rb
```diff
@@ -0,0 +1,98 @@
...
+      if resource.is_a?(Hash)
+        resource = create_resource_from_hash(inventory, inventory_target, resource)
+      end
+
+      unless resource.target == inventory_target
+        file, line = Puppet::Pops::PuppetStack.top_of_stack
+        raise Bolt::ValidationError, "Cannot set resource #{resource.reference} for target "\
+                                     "#{resource.target} on target #{inventory_target}. "\
+                                     "(#{file}, #{line})"
+      end
```
beechtom, line 77: I'm not sure if this is the best way to do this... raising a regular `Bolt::ValidationError` doesn't give the relevant line of the plan, which seems unhelpful. I was looking at using a `Puppet::Pops::Issue`, but it doesn't look like there's one that fits well here.

documentation/bolt_types_reference.md
beechtom: 👍 I'll write up some pre-docs for this once the API is a little more solidified.

lib/bolt/resource_instance.rb
```diff
@@ -0,0 +1,116 @@
...
+module Bolt
+  class ResourceInstance
+    attr_reader :target, :type, :title, :state, :desired_state
+    attr_accessor :events
+
+    def initialize(resource_hash)
+      @target        = resource_hash['target']
+      @type          = resource_hash['type']
+      @title         = resource_hash['title']
+      @state         = resource_hash['state'] || resource_hash['parameters'] || {}
```
beechtom, line 15: `resource_hash['parameters']` is here so you can create a `ResourceInstance` from the hash returned from `get_resources()`.

bolt-modules/boltlib/lib/puppet/datatypes/resourceinstance.rb
beechtom: Updated to include `Type[Resource]` as a type.

lib/bolt/resource_instance.rb
```diff
@@ -0,0 +1,117 @@
...
+      self.class.equal?(other.class) &&
+        target == other.target &&
+        type   == other.type &&
+        title  == other.title
+    end
+    alias == eql?
+
+    def to_s
+      reference
+    end
```
beechtom, line 62: Should this return the reference, or something else like the resource declaration? Do we want a separate method that will return the stringified resource declaration?

```
Type { 'title': attributes }
```

bolt-modules/boltlib/lib/puppet/functions/set_resources.rb
beechtom: I think `Puppet::Util::Errors.error_location` instead? `error_context` isn't a class method, so it raises an undefined method error. Am I missing something else?

nicklewis: Hmm yeah you're right. That doesn't save a whole lot then.

bolt-modules/boltlib/lib/puppet/datatypes/target.rb
```diff
@@ -20,7 +20,8 @@
       vars => { type => Optional[Hash[String[1], Data]], kind => given_or_derived },
       facts => { type => Optional[Hash[String[1], Data]], kind => given_or_derived },
       features => { type => Optional[Array[String[1]]], kind => given_or_derived },
-      plugin_hooks => { type => Optional[Hash[String[1], Data]], kind => given_or_derived }
+      plugin_hooks => { type => Optional[Hash[String[1], Data]], kind => given_or_derived },
```
lucywyman, line 23: It seems like most of these will never be nil, are they optional just in case?

bolt-modules/boltlib/lib/puppet/functions/set_resources.rb
```diff
@@ -0,0 +1,119 @@
+# frozen_string_literal: true
+
+require 'bolt/error'
+
+# Sets one or more resources on a target.
```
lucywyman, line 5: I think it's important to note that this doesn't apply the resource to the target.

bolt-modules/boltlib/lib/puppet/functions/set_resources.rb
```diff
@@ -0,0 +1,119 @@
...
+  # @param resource The resource to set on the target.
+  # @return The added `ResourceInstance` object.
+  # @example Add a single resource to a target with a resource data hash.
+  #   $resource = {
+  #     'type'  => 'file',
+  #     'title' => '/etc/puppetlabs',
+  #     'state' => { 'ensure' => 'present' }
+  #   }
+  #   $target.set_resources($resource)
+  dispatch :set_resource do
```
lucywyman, line 53: I don't follow why we have this? It isn't callable from a plan

bolt-modules/boltlib/lib/puppet/datatypes/target.rb
beechtom: I'm actually not sure -- the `Target` datatype is defined a little differently from the rest, so I just followed what was there. @donoghuc might have a better idea since it looks like he originally wrote it as optional?

bolt-modules/boltlib/lib/puppet/functions/set_resources.rb
beechtom: To allow you to pass a single resource or an array of resources without having a `Variant[]` type.

documentation/experimental_features.md
```diff
@@ -106,3 +115,174 @@ MODULEPATH:
...
+
+When setting resources using a hash of parameters, you can pass either a single hash or an
+array of hashes:
+
+> **Note:** When passing a data hash to `Target.set_resources`, the `target` parameter
+  is **optional**. If the `target` parameter is not specified, the function automatically
+  sets the target to the target the function is called on.
+
+> **Note:** If the `target` parameter is any target other than the one you are setting the
+  resource on, Bolt will raise an error.
```
hestonhoffman, line 157: Let's move these notes below the codeblock

documentation/experimental_features.md
```diff
@@ -106,3 +115,174 @@ MODULEPATH:
...
+  'desired_state' => Optional[Hash[String[1], Data]],
+  'events'        => Optional[Array[Hash[String[1], Data]]]
+}
+
+$target.set_resources(
+  Variant[Hash[String[1], Any], Array[Hash[String[1], Any]]] init_hash
+)
+```
+
+When setting resources using existing `ResourceInstance`s, you can pass either a single
```
hestonhoffman, line 174:
```suggestion
When setting resources using existing `ResourceInstance`s, you can pass either a single
```
This is more of a style thing, but I think the plural `ResourceInstance`s is going to look weird once it's rendered on the site. Maybe rephrase to "existing `ResourceInstance` objects"

documentation/experimental_features.md
```diff
@@ -106,3 +115,174 @@ MODULEPATH:
...
+`ResourceInstance` or an array of `ResourceInstance`s.
```
hestonhoffman, line 175:
```suggestion
`ResourceInstance` or an array of `ResourceInstance` objects.
```

documentation/experimental_features.md
```diff
@@ -106,3 +115,174 @@ MODULEPATH:
...
+
+$target.set_resources(
+  Variant[ResourceInstance, Array[ResourceInstance]] resource
+)
+```
+
+A target can only have a single instance of a given resource. If a duplicate resource is set
+on a target, the `state` and `desired_state` of the duplicate resource will be shallow merged
+with that of the existing resource, while any `events` for the duplicate resource will be
+added to the events for the existing resource.
```
hestonhoffman, line 191:
```suggestion
A target can only have a single instance of a given resource. If you set a duplicate resource
on a target, Bolt shallow merges the `state` and `desired_state` of the duplicate resource
with the `state` and `desired state` of the existing resource and adds any `events` for the 
duplicate resource to the existing resource.
```

documentation/experimental_features.md
```diff
@@ -106,3 +115,174 @@ MODULEPATH:
...
+A `ResourceInstance` is identified by its `target`, `type`, and `title`. As such, these
+three parameters _must_ be specified when creating a new `ResourceInstance` and are
+immutable. Since `Target.set_resources` will automatically set the `target` when passing
+a hash of parameters, the `target` parameter can be ommitted.
+
+The `state`, `desired_state`, and `events` parameters are optional when creating a
+`ResourceInstance`. If `state` and `desired_state` are not specified they will default to
+empty hashes, while `events` will default to an empty array. Each of these attributes can
+be modified during the lifetime of the `ResourceInstance` using [the data type's
+functions](bolt_types_reference.md#resourceinstance).
```
hestonhoffman, line 250:
```suggestion
The `state`, `desired_state`, and `events` parameters are optional when creating a
`ResourceInstance`. If you do not specify `state` and `desired_state`, they default to
empty hashes, while `events` defaults to an empty array. You can modify each of 
these attributes during the lifetime of the `ResourceInstance` using [the data type's
functions](bolt_types_reference.md#resourceinstance).
```

documentation/experimental_features.md
```diff
@@ -106,3 +115,174 @@ MODULEPATH:
...
+
+### Functions
+
+The `ResourceInstance` data type has several built-in functions. These range from accessing
+the object's attributes to modifying and overwriting state. A full list of the available
+functions can be found in the [data type reference 
+documentation](bolt_types_reference.md#resourceinstance).
```
hestonhoffman, line 257:
```suggestion
The `ResourceInstance` data type has several built-in functions. These range from accessing
the object's attributes to modifying and overwriting state. For a full list of the available
functions, see [Bolt data types](bolt_types_reference.md#resourceinstance).
```

lib/bolt/resource_instance.rb
```diff
@@ -0,0 +1,126 @@
...
+      @events        = resource_hash['events'] || []
+    end
+
+    # Creates a ResourceInstance from a data hash in a plan when calling
+    # ResourceInstance.new($resource_hash) or $target.set_resources($resource_hash)
+    def self.from_asserted_hash(resource_hash)
+      new(resource_hash)
+    end
+
+    # Creates a ResourceINstance from positional arguments in a plan when
```
lucywyman, line 45:
```suggestion
    # Creates a ResourceInstance from positional arguments in a plan when
```

PR #1788, opened 2020-05-05
(maint) Log task name with debug logging

donoghuc:
Previously the debug logging in the base transport class would attept to interpolate a task object into a string instead of using the Task#name method. This resulted in difficult to read debug logs. This commit updates the debug logging to use the task name.

!no-release-note


donoghuc:
Should `Task` just alias `name` with `to_s`?

nicklewis:
The name isn't a reasonable substitute for the whole task, so I would prefer to be explicit about it for now.

PR #1791, opened 2020-05-06
(maint) Expand filepaths passed on the CLI relative to cwd

lucywyman:
As part of the [#1364](https://github.com/puppetlabs/bolt/pull/1364) work to
standardize how we expand configured paths we expand all paths
configured in files relative to the project directory, and stated that
all paths set on the CLI would be expanded relative to the directory
Bolt was run from. However, only some CLI paths (inventoryfile and 
puppetfile) were expanded this way, while others (hiera-config and 
private-key) were expanded relative to the project
directory. This standardizes paths passed on the CLI to be expanded
relative to the cwd. This is accomplished by expanding the path in our 
CLI option handler, so that the config object only sees the path as
absolute and does not try to expand it relative to the project.

!bug

* **Expand filepaths passed on the CLI relative to the current working directory**
  Config options hiera-config and private-key are now expanded relative
  to the directory Bolt was run from when specified on the CLI, inline
  with other CLI options.

### DIFF
--- a/lib/bolt/bolt_option_parser.rb
+++ b/lib/bolt/bolt_option_parser.rb
@@ -655,8 +655,8 @@ module Bolt
         @options[:password] = STDIN.noecho(&:gets).chomp
         STDERR.puts
       end
-      define('--private-key KEY', 'Private ssh key to authenticate with') do |key|
-        @options[:'private-key'] = key
+      define('--private-key KEY', 'Path to private ssh key to authenticate with') do |key|
+        @options[:'private-key'] = File.expand_path(key)
       end
       define('--[no-]host-key-check', 'Check host keys with SSH') do |host_key_check|
         @options[:'host-key-check'] = host_key_check
@@ -718,7 +718,7 @@ module Bolt
       end
       define('--hiera-config FILEPATH',
              'Specify where to load Hiera config from (default: ~/.puppetlabs/bolt/hiera.yaml)') do |path|
-        @options[:'hiera-config'] = path
+        @options[:'hiera-config'] = File.expand_path(path)
       end
       define('-i', '--inventoryfile FILEPATH',
              'Specify where to load inventory from (default: ~/.puppetlabs/bolt/inventory.yaml)') do |path|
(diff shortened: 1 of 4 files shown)


lib/bolt/bolt_option_parser.rb
```diff
@@ -683,7 +683,7 @@ def initialize(options)
       end
       define('--sudo-executable EXEC', "Specify an executable for running as another user.",
              "This option is experimental.") do |exec|
-        @options[:'sudo-executable'] = exec
+        @options[:'sudo-executable'] = File.expand_path(exec)
```
nicklewis, line 686: We don't want to expand this one, we want to find it on the PATH.

lib/bolt/bolt_option_parser.rb
```diff
@@ -748,7 +748,7 @@ def initialize(options)
         @options[:tty] = tty
       end
       define('--tmpdir DIR', 'The directory to upload and execute temporary files on the target') do |tmpdir|
-        @options[:tmpdir] = tmpdir
+        @options[:tmpdir] = File.expand_path(tmpdir)
```
nicklewis, line 751: This path is used on the target, so expanding it locally doesn't make sense. Practically, it needs to be an absolute path but there's no easy way for us to test that (since absolute depends on whether the target is windows or *nix).

lucywyman: Well, that's what I get for testing on localhost :P

nicklewis: "Works with the local transport" is the new "works on my machine"

spec/integration/ssh_spec.rb
```diff
@@ -29,6 +29,13 @@
       expect(result['stdout'].strip).to eq(user)
     end
 
+    it 'expands options relative to current working directory' do
+      config_flags = %W[-t #{uri} --no-host-key-check -m #{modulepath} --private-key ./ssh/private --debug]
```
nicklewis, line 33: This seems like it's duplicating the test in `spec/bolt/cli_spec.rb`. Is there something it's testing beyond what's already covered in the other test?

lucywyman: :facepalm: Yup, removed

nicklewis:
There's a test failing now. It looks like the most recent pushed removed an `expand_path` in the test that was still necessary

PR #1796, opened 2020-05-06
Clarify project.yaml usage

hestonhoffman:


documentation/directory_structure.md
```diff
@@ -59,3 +61,49 @@ Follow these tips for managing standalone modules:
...
+
+Here is an example of a project using a simplified directory structure:
+```console
+.
+├── bolt.yaml
+├── plans
+│   └── myplan.yaml
+├── project.yaml
+└── tasks
+    └── mytask.yaml
```
beechtom, line 90: Tasks won't be a yaml file -- another extension like `.rb` or `.sh` would work better here.

documentation/directory_structure.md
```diff
@@ -59,3 +61,49 @@ Follow these tips for managing standalone modules:
...
+> **Note:** This is an experimental feature.
+
+To load your content from a Bolt project:
+1. Create a `project.yaml` file in the root directory of your Bolt project. 
+2. Place your Bolt plans and tasks in `plans` and `tasks` directories
+   respectively.
+
+If `project.yaml` exists at the root of a project directory, and your tasks and
+plans are located in their respective directories, Bolt loads the project as a 
+module namespaced to the name of the directory.
```
beechtom, line 80:
```suggestion
If `project.yaml` exists at the root of a project directory, Bolt loads the project as a module.
Tasks and plans are loaded from the `tasks/` and `plans/` directories and are namespaced
to the project name.
```

You don't need any tasks or plans for Bolt to load the project as a module. If the directories are missing or don't have any content, the project is loaded like any other module that doesn't have tasks or plans.

documentation/bolt_project_directories.md
```diff
@@ -90,8 +100,15 @@ The default paths for all Bolt configuration, code, and data are relative to the
...
 |[`modules/`](bolt_installing_modules.md#)|The directory where modules from the `Puppetfile` are installed. In most cases, do not edit these modules locally.|
 |[`site-modules`](bolt_installing_modules.md)|Local modules that are edited and versioned with the project directory.|
 |`data/`|The standard path to store static Hiera data files.|
+
+## Simplified directory structure
+
+If you're working on a new project and you want to use a simplified directory
+structure, you can load your plan and task content directly from your project as
+if it was a module. For more information, see
+[Directory structure](./directory_structure.md#simplified-directory-structure)
```
beechtom, line 114: I'm a little concerned that this could convey the wrong impression on what projects are meant for. A simplified directory structure is definitely a benefit of projects, but the intent is that project-level content should be application-specific, while any content being written as part of a reusable library should still go in a module.

documentation/bolt_project_directories.md
```diff
@@ -90,8 +100,15 @@ The default paths for all Bolt configuration, code, and data are relative to the
 |[`bolt.yaml`](bolt_configuration_reference.md)|Contains configuration options for Bolt.|
 |`hiera.yaml`|Contains the Hiera config to use for target-specific data when using `apply`.|
 |[`inventory.yaml`](inventory_file_v2.md)|Contains a list of known targets and target specific data.|
-|[`project.yaml`](bolt_configuration_reference.md#project_configuration_options)|Contains configuration for the Bolt project.  The project.yaml file contains a whitelist of tasks and plans that you can use to limit the output from the `bolt [plan|task] show` command.|
+|[`project.yaml`](bolt_configuration_reference.md#project_configuration_options)|Contains configuration for the Bolt project.  The `project.yaml` file contains a whitelist of tasks and plans that you can use to limit the output from the `bolt [plan|task] show` command.|
```
beechtom, line 103:
```suggestion
|[`project.yaml`](bolt_configuration_reference.md#project_configuration_options)|Contains configuration for the Bolt project.  The `project.yaml` file contains a whitelist of tasks and plans that you can use to limit the output from the `bolt task show` and `bolt plan show` commands. |
```

This won't render correctly with the brackets.

documentation/bolt_project_directories.md
```diff
@@ -4,6 +4,16 @@ Bolt runs in the context of a project directory or a `Boltdir`. This directory c
...
 
+## Project directory names
+
+Project directory names must match the expression: `[a-z][a-z0-9_]*`. In other words, they
+can contain only lowercase letters, numbers, and underscores, and begin with a 
+lowercase letter.
+
+Avoid giving your project the same name as another module in your modulepath.
+You cannot give your project the same name as a core Bolt module such as
+`boltlib`.
```
beechtom, line 15: I'm not sure that this needs to be explicitly stated, especially since we don't actually list the core Bolt modules anywhere. If a user happens to use a conflicting name, Bolt will display an error telling them as much. If we do want this in the docs I think it'd be helpful to list all of the core modules.

documentation/bolt_project_directories.md
```diff
@@ -4,6 +4,16 @@ Bolt runs in the context of a project directory or a `Boltdir`. This directory c
 
 The project directory structure makes it easy to share Bolt code by committing the project directory to Git. You can then check different repositories of Bolt code into different directories in order to manage various applications.
 
+## Project directory names
+
+Project directory names must match the expression: `[a-z][a-z0-9_]*`. In other words, they
+can contain only lowercase letters, numbers, and underscores, and begin with a 
+lowercase letter.
+
+Avoid giving your project the same name as another module in your modulepath.
```
beechtom, line 13: Would it be useful to say why users should avoid this? I'm not sure if there's a real use case for wanting to shadow an installed module, but since Bolt won't warn/break if you do it might be helpful to explain what would happen.

lucywyman: The use case for wanting to shadow an installed module is developing a module that's installed elsewhere on the modulepath. Agree it'd be better to say "Projects will take precedence over installed modules of the same name, and will error if they conflict with builtin Bolt modules".

documentation/bolt_project_directories.md
hestonhoffman: Yeah agreed. The use-case wasn't clear to me until I watched your demo TBH. I'm starting to think this doc needs its own page

hestonhoffman:
Closing this in favor of #1804

PR #1797, opened 2020-05-07
(GH-1779) Warn when Bolt is installed as a gem

### LINKED ISSUE #1779: Warn when running Bolt as a gem
## Use Case
When installed as a gem, Bolt doesn't include its bundled content. This causes unexpected behavior for users who should have installed the native package instead.

Also, it can cause problems for users who install bolt as gem and then later install the package, because the gem binary can shadow the package without the user realizing it.

When invoking the `bolt` command from a gem install, it should issue some warning that this is probably not what you meant to do and should suggest installing the native package instead.

We should detect this by ensuring the bundled content directory is populated and issuing a warning if the CLI is invoked when it's empty.
We should check `BOLT_GEM` environment variable and if set to anything other than true disable this warning for users who want to embed bolt's entire CLI in another project.

• Including the bundled content with the gem. This would require some changes to the gem build process and detracts from our message that you shouldn't use the gem install.
• Not shipping the `bolt` command with the gem. This could break any testing use cases for the gem install that _do_ require the command. It would also be unintuitive.

beechtom:
This adds a warning that Bolt may be installed as a gem when bundled
content is missing. Bolt checks if any modules other than `aggregate`,
`canary`, and `puppetdb_fact` are installed at `Bolt::PAL::MODULES_PATH`
to determine if it may be a gem install.

This warning can be disabled by setting the `BOLT_GEM` environment
variable.

Closes #1779 

!feature

* **Warn when Bolt is installed as a gem**
  ([#1779](https://github.com/puppetlabs/bolt/issues/1779))

  Bolt now issues a warning when it detects that it may have been
  installed as a gem. This warning can be disabled by setting the
  `BOLT_GEM` environment variable.

  To install Bolt reliably and with all of its dependencies, it should
  be [installed as a package](https://puppet.com/docs/bolt/latest/bolt_installing.html).

### DIFF
--- a/lib/bolt/cli.rb
+++ b/lib/bolt/cli.rb
@@ -804,6 +804,20 @@ module Bolt
     end
 
     def bundled_content
+      # If the bundled content directory is empty, Bolt is likely installed as a gem.
+      if ENV['BOLT_GEM'] != 'false' && gem_install?
+        msg = <<~MSG.chomp
+          Bolt may be installed as a gem. To use Bolt reliably and with all of its
+          dependencies, uninstall the 'bolt' gem and install Bolt as a package:
+          https://puppet.com/docs/bolt/latest/bolt_installing.html
+
+          If you meant to install Bolt as a gem and want to disable this warning,
+          set the BOLT_GEM environment variable to 'false'.
+        MSG
+
+        @logger.warn(msg)
+      end
+
       # We only need to enumerate bundled content when running a task or plan
       content = { 'Plan' => [],
                   'Task' => [],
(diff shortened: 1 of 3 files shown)


lib/bolt/cli.rb
```diff
@@ -804,6 +804,20 @@ def analytics
     end
 
     def bundled_content
+      # If the bundled content directory is empty, Bolt is likely installed as a gem.
+      if ENV['BOLT_GEM'] != 'false' && gem_install?
```
beechtom, line 808: The ticket said "when BOLT_GEM is anything but true", which seemed weird as we'd need to first check that it wasn't `nil`. Since a user will need to explicitly set this variable to disable the warning, I felt that requiring the value to be `false` made sense.

nicklewis: I think common practice with variables like these is to simply require they be set. That is, setting `BOLT_GEM` means "I know I'm running from a gem install and that's okay".

lib/bolt/cli.rb
```diff
@@ -827,5 +841,11 @@ def config_loaded
       MSG
       @logger.debug(msg)
     end
+
+    # Gem installs include the aggregate, canary, and puppetdb_fact modules, while
+    # package installs include modules listed in the Bolt repo Puppetfile
+    def gem_install?
```
nicklewis, line 847: This might better be named `incomplete_install?`, since it will also be the case for a fresh clone of Bolt.

beechtom: Updated!

PR #1798, opened 2020-05-07
(PUP-10476) Use updated PAL api for catalog compilation

donoghuc:
This commit updates the catalog compiler to use updates to PAL api to enable node definitions in plans and deserialized bolt types in apply blocks without relying on using public methods. This includes a refactor to clean up the compile_catalog method and make it more readable.

### DIFF
--- a/lib/bolt/catalog.rb
+++ b/lib/bolt/catalog.rb
@@ -19,7 +19,7 @@ module Bolt
       @log_level = log_level
     end
 
-    def with_puppet_settings(hiera_config = {})
+    def with_puppet_settings(overrides = {})
       Dir.mktmpdir('bolt') do |dir|
         cli = []
         Puppet::Settings::REQUIRED_APP_SETTINGS.each do |setting|
@@ -31,7 +31,9 @@ module Bolt
         Puppet.settings.override_default(:vendormoduledir, '')
 
         Puppet.initialize_settings(cli)
-        Puppet.settings[:hiera_config] = hiera_config
+        overrides.each do |setting, value|
+          Puppet.settings[setting] = value
+        end
 
         # Use a special logdest that serializes all log messages and their level to stderr.
         Puppet::Util::Log.newdestination(:stderr)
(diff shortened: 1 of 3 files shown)


nicklewis:
This makes sense to me.

donoghuc:
Depends on https://github.com/puppetlabs/puppet-runtime/pull/343

donoghuc:
Puppet runtime is now shipping puppet 6.16.0

lib/bolt/catalog.rb
```diff
@@ -54,89 +56,100 @@ def generate_ast(code, filename = nil)
...
+      topscope_vars = target_vars.merge(plan_vars)
+      env_conf = { modulepath: request['modulepath'],
+                   facts: target['facts'],
+                   variables: topscope_vars }
+
+      puppet_settings = {
+        node_name_value: target['name'],
+        # CODEREVIEW: Isnt this already the default?
+        # https://github.com/puppetlabs/puppet/blob/master/lib/puppet/parser/catalog_compiler.rb#L18
+        rich_data: true,
```
lucywyman, line 88: Yes, and Puppet's default for this is true as well, and we shouldn't be picking up Puppet config anywhere. I think it's safe to remove.

donoghuc: Awesome. Removed.

lib/bolt/catalog.rb
```diff
@@ -54,89 +56,97 @@ def generate_ast(code, filename = nil)
...
-                     facts: target['facts'] || {} }
-        env_conf[:variables] = {}
+      inv = Bolt::ApplyInventory.new(request['config'])
+      puppet_overrides = {
+        bolt_pdb_client: pdb_client,
+        bolt_inventory: inv,
+        bolt_project: bolt_project
+      }
+
+      # Facts will be set by the catalog compiler, so we need to ensure
```
lucywyman, line 69: I think this actually belongs above the definition of `shadow_vars`, since this describes exactly what that function is doing.

lucywyman: Or to above `plan_vars`, whichever you think makes more sense.

lucywyman: Or removed altogether, the current comment above the function definition does a pretty good job of getting the important point across

donoghuc: Maybe just the `target = request['target']` makes that a bit awkward. We need that `target` var there to access facts. The description i think is meant to descrive _how_ and _why_ topscope_vars is computed.

lucywyman: I just thought it might be more clear placed elsewhere, but I don't think it's worth holding this up for.

Issue #1795, opened 2020-05-06
Add JSON schema for common yaml files.


adreyer:
We should have json schema's in the bolt repo for common yaml file formats in Bolt. These should live in a top level `/schemas` directory. 

- [x] inventory.yaml
- [x] bolt.yaml
- [x] project.yaml

Issue #1771, opened 2020-04-27
Allow plugins to resolve keys in puppetdb config


nmaludy:
## Use Case
I'm trying to write a plugin that can resolve certificates for the PuppetDB plugin. This way we can support different laptop setups along with automatically generating the cert name given the hostname of the bolt control node (because the certs are usually: `/etc/puppetlabs/puppet/ssl/certs/<fqdn>.pem`).

## Describe the Solution You Would Like
Ability to do the following:
```
puppetdb:
  server_urls: ["https://puppetdb.domain.tld:8081"]
  cacert:
    _plugin: task
    task: inventory_utils::path_lookup
    parameters:
      paths:
        - ~/.puppetlabs/bolt/customer/ssl/certs/ca.pem
        - /etc/puppetlabs/puppet/ssl/certs/customer/ca.pem
        - /etc/puppetlabs/puppet/ssl/certs/ca.pem
  cert:
    _plugin: task
    task: inventory_utils::path_lookup
    parameters:
      paths:
        - ~/.puppetlabs/bolt/customer/ssl/certs/bolt.pem
        - /etc/puppetlabs/puppet/ssl/certs/customer/bolt.pem
        - /etc/puppetlabs/puppet/ssl/certs/bolt.pem
        - _plugin: task
          task: inventory_utils::erb_template
          parameters:
            template: '/etc/puppetlabs/puppet/ssl/certs/<%= Socket.gethostname %>.pem'
```

## Current behavior
When i try do that current i get the following:
```
Found unsupported key _plugin in config setting puppetdb. Plugins are only available in ssh, winrm, pcp, local, docker, remote, plugins, plugin_hooks.  
```

nmaludy:
Saw the core team picking this up and wanted to "brain dump" my thoughts where i left off.

- Added `puppetdb` to the list of allowed keys with a `_plugin`: https://github.com/puppetlabs/bolt/blob/master/lib/bolt/config.rb#L313
- Next time i ran `bolt` an exception was thrown:
```shell
Traceback (most recent call last):
        13: from /opt/puppetlabs/bolt/bin/bolt:23:in `<main>'
        12: from /opt/puppetlabs/bolt/bin/bolt:23:in `load'
        11: from /opt/puppetlabs/bolt/lib/ruby/gems/2.5.0/gems/bolt-2.6.0/exe/bolt:10:in `<top (required)>'
[...]

PR #1805, opened 2020-05-08
(GH-1771) Use plugins to set PuppetDB config

nicklewis:
Previously, PuppetDB config was not able to be set using plugins. This was
due to the fact that PuppetDB config is used to configure the PuppetDB
plugin which can then be used to configure other plugins. We already have
logic that handles incrementally resolving data in the
`plugins` section of the config, but that couldn't accomodate PuppetDB 
which was handled specially.

We used to construct the PuppetDB client and pass it to Bolt::Plugin during 
plugin setup, where it would be used to construct the PuppetDB plugin and
then other plugins would be resolved thereafter. This commit flips that
dependency so that Bolt::Plugin takes the unresolved PuppetDB config,
merges it with the rest of the plugin config, and uses that to construct
the PuppetDB plugin. The CLI then takes a reference to the PuppetDB client
from the PuppetDB plugin, rather than the plugin taking a reference to the
CLI's client. This allows us to still have a single PuppetDB client, but
for that client to be configurable using plugins.

!feature

  * **Use plugins to set PuppetDB config** ([#1771](https://github.com/puppetlabs/bolt/issues/1795))

    Plugin references can now be used to set configuration options for
    the PuppetDB client used by Bolt, in the `puppetdb` section of the
    config.

### DIFF
--- a/lib/bolt/plugin.rb
+++ b/lib/bolt/plugin.rb
@@ -159,6 +159,10 @@ module Bolt
       @unresolved_plugin_configs = config.plugins.dup
       # The puppetdb plugin config comes from the puppetdb section, not from
       # the plugins section
+      if @unresolved_plugin_configs.key?('puppetdb')
+        msg = "Configuration for the PuppetDB plugin must be in the 'puppetdb' config section, not 'plugins'"
+        raise Bolt::Error.new(msg, 'bolt/plugin-error')
+      end
       @unresolved_plugin_configs['puppetdb'] = config.puppetdb if config.puppetdb
       @plugin_hooks = DEFAULT_PLUGIN_HOOKS.dup
     end
--- a/lib/bolt/plugin.rb
+++ b/lib/bolt/plugin.rb
@@ -119,12 +119,8 @@ module Bolt
       end
     end
 
-    def self.setup(config, pal, pdb_client, analytics)
+    def self.setup(config, pal, analytics = Bolt::Analytics::NoopClient.new)
       plugins = new(config, pal, analytics)
-
-      # PDB is special because it needs the PDB client. Since it has no config,
-      # we can just add it first.
-      plugins.add_plugin(Bolt::Plugin::Puppetdb.new(pdb_client))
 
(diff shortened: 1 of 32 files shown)


lib/bolt/plugin.rb
```diff
@@ -119,13 +119,9 @@ def boltdir
       end
     end
 
-    def self.setup(config, pal, pdb_client, analytics)
+    def self.setup(config, pal, analytics = Bolt::Analytics::NoopClient.new)
```
lucywyman, line 122: If we set this default can we also remove passing the parameter when it's the noop client? I see it in `bolt_spec/bolt_context.rb`, `bolt/inventory.rb`, `bolt_spec/run.rb`, and a lot of tests. I know this PR is already big, but it's a nice simplification!

lib/bolt/plugin.rb
```diff
@@ -161,14 +157,16 @@ def initialize(config, pal, analytics)
       @unknown = Set.new
       @resolution_stack = []
       @unresolved_plugin_configs = config.plugins.dup
+      # The puppetdb plugin config comes from the puppetdb section, not from
+      # the plugins section
+      @unresolved_plugin_configs['puppetdb'] = config.puppetdb if config.puppetdb
```
lucywyman, line 162: I doubt many people will run into this but should we raise an error if `puppetdb` is set under the `plugins` key?

lib/bolt_spec/bolt_context.rb
```diff
@@ -147,7 +147,6 @@ def config
     def plugins
       @plugins ||= Bolt::Plugin.setup(config,
                                       pal,
-                                      nil,
                                       Bolt::Analytics::NoopClient.new)
```
lucywyman, line 150: This line can also be removed

nicklewis:
Updated to remove redundant passing of `NoopClient` (I did the same for `Bolt::Executor` in some places) and to error if `puppetdb` is set under `plugins`.

PR #1806, opened 2020-05-08
(GH-1780) Allow users to shell out to SSH

### LINKED ISSUE #1780: External SSH transport
## Use Case
I want to use SSH features (like key encryption algorithms) that aren't supported by Ruby libraries. In general users should use net::ssh instead of this implementation except where necessary.

Separate ssh connection implementation that shells out to system ssh and is controlled by an ssh config option. In general users should rely on their ssh config for configuring this transport rather than the inventory options.

`ssh-command`/`ssh-executable` is the config option to enable this and choose which ssh to use.

## Acceptable Limitations
For the initial pass at this:
• no password auth.
• invoke ssh separately for each command we run.
• run-as if it's easy otherwise disable.
• enumerate supported config options.
• experimental.
• no powershell support.
• not supported on windows.

lucywyman:
The SSH transport uses the net-ssh ruby library to establish an SSH
connection. While robust, the library is limited in it's OpenSSH
configuration and key exchange algorithm support. Users can now specify
an SSH command to run and Bolt will shell out to SSH, allowing users to
use all local OpenSSH configuration to connect.

This creates 2 new configuration options: `ssh-command` and
`copy-command`, which can be set in confg files, inventory files, or
specified on the CLI. Each option is an array which defines the command
to use when executing commands or copying files respectively.
`ssh-command` additionally enables using the 'external ssh' connection,
while `copy-command` does nothing if `ssh-command` is not set. Both
options accept either a string or array.

The external ssh connection does not support using Powershell.

The external ssh connection does not support the following configuration
options:
* connect-timeout
* disconnect-timeout
* host-key-check
* extensions
* load-config
* login-shell
* password
* proxyjump
* tty

These options should be specified in OpenSSH config.

Closes #1780

!feature

* **Allow users to shell out to SSH** ([#1780](https://github.com/puppetlabs/bolt/issues/1780))

  Users can now specify an SSH command, which Bolt will shell out to
  when using the SSH transport. This allows users to run SSH as if it were
  run locally without worrying about ruby library feature support.

### DIFF
--- a/documentation/bolt_configuration_reference.md.erb
+++ b/documentation/bolt_configuration_reference.md.erb
@@ -121,11 +121,19 @@ Transport configuration options can be set in both the configuration file and in
 <% @transports[:options].each do |transport, options| %>
 ### <%= transport %>
 
+<% if transport == 'ssh' %>
+| Option | Description | External SSH | Type | Default |
+| ------ | ----------- | :----------: | ---- | ------- |
+<% options.each do |option, data| -%>
+  | `<%= option %>` | <%= data[:desc] %> | <%= data[:external] == true ? '&#10004;' : '' %> | <%= data[:type] == TrueClass ? 'Boolean' : data[:type] %> | <%= @transports[:defaults][transport][option].to_s || "" %> |
+<% end %>
+<% else %>
 | Option | Description | Type | Default |
 | ------ | ----------- | ---- | ------- |
 <% options.each do |option, data| -%>
 | `<%= option %>` | <%= data[:desc] %> | <%= data[:type] == TrueClass ? 'Boolean' : data[:type] %> | <%= @transports[:defaults][transport][option].to_s || "" %> |
 <% end %>
+<%end%>
 
 <% if transport == 'ssh' %>
 #### OpenSSH
--- a/documentation/experimental_features.md
+++ b/documentation/experimental_features.md
@@ -281,4 +281,36 @@ $results = $targets.get_resources([Package, User])
 $results.each |$result| {
   $result.target.set_resources($result['resources'])
 }
+
+## External SSH transport
+
+Bolt's SSH transport uses the ruby library `net-ssh`, which is a pure ruby implementation of the
+SSH2 client protocol. While robust, the library lacks support for some features and algorithms that
+are available in native SSH. When you use the external SSH transport, Bolt uses the SSH executable
+you've specified instead of using `net-ssh`. Essentially, using the external SSH transport is the
+same as running SSH on your command line, but with Bolt managing the connections.
+
+To use the external SSH transport, set `ssh-command: <SSH>` in [bolt.yaml](configuring_bolt.md),
+where <SSH> is the SSH command to run. For example:
+
+```
+ssh:
+  ssh-command: 'ssh'
+```
+
+The value of `ssh-command` can be either a string or an array, and you can provide any flags to the
+command. Bolt will append Bolt-configuration settings to the command, as well as the specified
+target, when connecting. Not all Bolt configuration options are supported using the external SSH
+transport, but you can configure most options in your OpenSSH Config. See [bolt configuration
+reference](bolt_configuration_reference.md) for the list of supported Bolt SSH options.
+
+Bolt transports have two main functions: executing remotely, and copying files to the remote targets.
+`ssh-command` is what configures the remote execution, and `copy-command` configures the copy
+command. The default is `scp -r`, and `rsync` is not supported at this time.
+
+For example:
+```
+ssh:
+  ssh-command: 'ssh'
+  copy-command: 'scp -r -F ~/ssh-config/myconf'
 ```
--- a/lib/bolt/config/transport/ssh.rb
+++ b/lib/bolt/config/transport/ssh.rb
@@ -13,14 +13,20 @@ module Bolt
         #       in schemas/bolt-transport-definitions.json
         OPTIONS = {
           "cleanup"            => { type: TrueClass,
+                                    external: true,
                                     desc: "Whether to clean up temporary files created on targets." },
           "connect-timeout"    => { type: Integer,
                                     desc: "How long to wait when establishing connections." },
+          "copy-command"       => { external: true,
+                                    desc: "Command to use when copying files using ssh-command. "\
+                                          "Bolt runs `<copy-command> <src> <dest>`. **This option is experimental.**" },
           "disconnect-timeout" => { type: Integer,
                                     desc: "How long to wait before force-closing a connection." },
           "host"               => { type: String,
+                                    external: true,
                                     desc: "Host name." },
           "host-key-check"     => { type: TrueClass,
+                                    external: true,
                                     desc: "Whether to perform host key validation when connecting." },
           "extensions"         => { type: Array,
                                     desc: "List of file extensions that are accepted for scripts or tasks on Windows. "\
@@ -29,6 +35,7 @@ module Bolt
                                          "a `.py` script runs with `python.exe`. The extensions `.ps1`, `.rb`, and "\
                                          "`.pp` are always allowed and run via hard-coded executables." },
           "interpreters"       => { type: Hash,
+                                    external: true,
                                     desc: "A map of an extension name to the absolute path of an executable, "\
                                           "enabling you to override the shebang defined in a task executable. The "\
                                           "extension can optionally be specified with the `.` character (`.py` and "\
@@ -44,26 +51,36 @@ module Bolt
           "password"           => { type: String,
                                     desc: "Login password." },
           "port"               => { type: Integer,
+                                    external: true,
                                     desc: "Connection port." },
-          "private-key"        => { desc: "Either the path to the private key file to use for authentication, or a "\
+          "private-key"        => { external: true,
+                                    desc: "Either the path to the private key file to use for authentication, or a "\
                                           "hash with the key `key-data` and the contents of the private key." },
           "proxyjump"          => { type: String,
                                     desc: "A jump host to proxy connections through, and an optional user to "\
                                           "connect with." },
           "run-as"             => { type: String,
+                                    external: true,
                                     desc: "A different user to run commands as after login." },
           "run-as-command"     => { type: Array,
+                                    external: true,
                                     desc: "The command to elevate permissions. Bolt appends the user and command "\
                                           "strings to the configured `run-as-command` before running it on the "\
                                           "target. This command must not require an interactive password prompt, "\
                                           "and the `sudo-password` option is ignored when `run-as-command` is "\
                                           "specified. The `run-as-command` must be specified as an array." },
           "script-dir"         => { type: String,
+                                    external: true,
                                     desc: "The subdirectory of the tmpdir to use in place of a randomized "\
                                           "subdirectory for uploading and executing temporary files on the "\
                                           "target. It's expected that this directory already exists as a subdir "\
                                           "of tmpdir, which is either configured or defaults to `/tmp`." },
+          "ssh-command"        => { external: true,
+                                    desc: "Command and flags to use when SSHing. This enables the external "\
+                                          "SSH transport which shells out to the specified command. "\
+                                          "**This option is experimental.**" },
           "sudo-executable"    => { type: String,
+                                    external: true,
                                     desc: "The executable to use when escalating to the configured `run-as` "\
                                           "user. This is useful when you want to escalate using the configured "\
                                           "`sudo-password`, since `run-as-command` does not use `sudo-password` "\
@@ -71,14 +88,17 @@ module Bolt
                                           "`<sudo-executable> -S -u <user> -p custom_bolt_prompt <command>`. "\
                                           "**This option is experimental.**" },
           "sudo-password"      => { type: String,
+                                    external: true,
                                     desc: "Password to use when changing users via `run-as`." },
           "tmpdir"             => { type: String,
+                                    external: true,
                                     desc: "The directory to upload and execute temporary files on the target." },
           "tty"                => { type: TrueClass,
                                     desc: "Request a pseudo tty for the session. This option is generally "\
                                           "only used in conjunction with the `run-as` option when the sudoers "\
                                           "policy requires a `tty`." },
           "user"               => { type: String,
+                                    external: true,
                                     desc: "Login user." }
         }.freeze
 
@@ -103,6 +123,13 @@ module Bolt
 
             if key_opt.instance_of?(String)
               @config['private-key'] = File.expand_path(key_opt, @project)
+
+              # We have an explicit test for this to only warn if using net-ssh transport
+              Bolt::Util.validate_file('ssh key', @config['private-key']) if @config['ssh-command']
+            end
+
+            if key_opt.instance_of?(Hash) && @config['ssh-command']
+              raise Bolt::ValidationError, 'private-key must be a filepath when using ssh-command'
             end
           end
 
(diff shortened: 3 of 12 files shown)


lib/bolt/transport/ssh/exec_connection.rb
```diff
@@ -0,0 +1,123 @@
...
+              raise Bolt::ValidationError, 'private-key must be a filepath when using ssh-command'
+            end
+          end
+
+          _, err, stat = if source.is_a?(StringIO)
+                           Tempfile.create(File.basename(dest)) do |f|
+                             f.write(source.read)
+                             f.close
+                             cp_cmd << f
+                             cp_cmd << "#{userhost}:\"#{dest}\""
```
nicklewis, line 79: I think we need to shellescape `dest` rather than quoting it. Apparently that argument is interpreted by the shell on the target before the file is written. Who knew?

lib/bolt/transport/ssh/exec_connection.rb
```diff
@@ -0,0 +1,123 @@
...
+          Open3.popen3(build_ssh_command(command))
+        end
+
+        # This is used by the Bash shell to decide whether to `cd` before
+        # executing commands as a run-as user
+        def reset_cwd?
+          false
+        end
+
+        def validate(target)
```
nicklewis, line 106: Is there a reason not to have this in `Bolt::Config::Transport::SSH`? The first part could be gated on whether `ssh-command` is set and the second part looks like it should apply to both connection types.

lib/bolt/transport/ssh/exec_connection.rb
```diff
@@ -0,0 +1,123 @@
...
+        end
+
+        def execute(command)
+          Open3.popen3(build_ssh_command(command))
+        end
+
+        # This is used by the Bash shell to decide whether to `cd` before
+        # executing commands as a run-as user
+        def reset_cwd?
+          false
```
nicklewis, line 103: Why `false`? The existing connection type has `true`. It's really just there to distinguish SSH vs. local.

lib/bolt/transport/ssh/exec_connection.rb
```diff
@@ -0,0 +1,123 @@
...
+                    end
+          ssh_cmd += %W[-p #{@target.port}] if @target.port
+          ssh_cmd += %w[-o BatchMode=yes]
+          ssh_cmd += %w[-o StrictHostKeyChecking=no] if @target.transport_config['host-key-check'] == false
+          if (key = target.transport_config['private-key'])
+            if key.instance_of?(String)
+              ssh_cmd += %W[-i #{key}]
+            else
+              # This is simpler than writing out a tempfile for now
+              raise Bolt::ValidationError, 'private-key must be a filepath when using ssh-command'
```
nicklewis, line 52: This can be part of the validation on `Bolt::Config::Transport::SSH`.

lib/bolt/transport/ssh.rb
```diff
@@ -16,18 +16,21 @@ def initialize
...
-        @transport_logger = Logging.logger[Net::SSH]
-        @transport_logger.level = :warn
       end
 
       def with_connection(target)
-        conn = Connection.new(target, @transport_logger)
+        conn = if target.transport_config['ssh-command']
+                 ExecConnection.new(target)
+               else
+                 transport_logger = Logging.logger[Net::SSH]
```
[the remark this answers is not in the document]
lucywyman: Why does it need to be in the init method if we never need it? This just seemed like a more efficient and sensible place for it.

lib/bolt/transport/ssh/exec_connection.rb
```diff
@@ -0,0 +1,123 @@
...
+              'CONNECT_ERROR'
+            )
+          end
+        end
+
+        def shell
+          Bolt::Shell::Bash.new(@target, self)
+        end
+
+        def userhost
```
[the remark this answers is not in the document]
lucywyman: I think that could be confused with scp's 'dest', and we use this there.

lib/bolt/transport/ssh/exec_connection.rb
lucywyman: I don't think just shellescape will work. With that, scp sees `scp /local/file with spaces user@host:file with spaces`. It correctly interprets the first argument as '/local/file with spaces', but needs to have quotes around the destination otherwise it interprets the spaces as new arguments. It looks like there are other escapes we can do, but quotes seems just as reasonable as other options: https://stackoverflow.com/questions/19858176/how-do-i-escape-spaces-in-path-for-scp-copy-in-linux

lib/bolt/transport/ssh/exec_connection.rb
lucywyman: I thought I updated this? :shrug:

lib/bolt/transport/ssh.rb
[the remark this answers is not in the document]
nicklewis: `Logging.logger[Net::SSH]` gets a reference to a single logging object, so it seems strange for each connection to set the log level each time, but I don't have strong feelings about it.

But if we're doing it more than once rather than initializing it and passing it to each Connection, then it can 
be in the `Connection#initialize` method so this class doesn't need to know about it.

lib/bolt/transport/ssh/exec_connection.rb
```diff
@@ -0,0 +1,123 @@
...
+module Bolt
+  module Transport
+    class SSH < Simple
+      class ExecConnection
+        attr_reader :user, :target
+
+        def initialize(target)
+          @target = target
+          ssh_config = Net::SSH::Config.for(target.host)
+          @user = @target.user || ssh_config[:user] || Etc.getlogin
```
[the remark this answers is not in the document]
lucywyman: I think this actually needs to be set as we use it elsewhere in the codebase, like setting the owner for tmpdir https://github.com/puppetlabs/bolt/blob/master/lib/bolt/shell/bash/tmpdir.rb#L9 or setting error messages https://github.com/puppetlabs/bolt/blob/master/lib/bolt/shell/bash.rb

lib/bolt/transport/ssh.rb
```diff
@@ -16,18 +16,21 @@ def initialize
...
+                 transport_logger.level = :warn
+                 Connection.new(target, transport_logger)
+               end
         conn.connect
         yield conn
       ensure
         begin
-          conn&.disconnect
+          conn&.disconnect if conn.is_a?(Connection)
```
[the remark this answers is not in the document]
lucywyman: Is that necessary before we have control sockets? This conditional is a bit weird, but I'm concerned about defining an api for transports that doesn't always make sense. A disconnect method doesn't seem generally necessary for implementations of the SSH connection in the same way the connect method is.

lib/bolt/transport/ssh/exec_connection.rb
```diff
@@ -0,0 +1,123 @@
...
+            if key.instance_of?(String)
+              ssh_cmd += %W[-i #{key}]
+            else
+              # This is simpler than writing out a tempfile for now
+              raise Bolt::ValidationError, 'private-key must be a filepath when using ssh-command'
+            end
+          end
+          ssh_cmd << userhost
+          ssh_cmd << command
+          Shellwords.shelljoin(ssh_cmd)
```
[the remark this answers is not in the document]
lucywyman: Does using just an array escape special characters? Since a lot of this is user-input, including the command itself, this just seemed safer.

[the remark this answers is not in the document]
nicklewis: Using the array form will bypass the shell entirely, so it will be guaranteed to come through exactly as expected.

lib/bolt/transport/ssh/exec_connection.rb
nicklewis: Ah yes I see, you're not quoting it, you're escaped-quoting it. Then the options are either to double shellescape it (technically shellescape `dest` and then shellescape the entire argument) or to shellescape it once and use the array form of open3 rather than string.

lib/bolt/transport/ssh.rb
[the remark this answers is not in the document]
lucywyman: Oh, that's a good point, I didn't consider this would initialize a logger for every target. I'll revert it to how it was.

lib/bolt/transport/ssh.rb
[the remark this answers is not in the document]
nicklewis: I generally prefer to keep the differences encapsulated in the two connection classes rather than bubbling up to this level.

The `connect` method is a bit weird here, in that it doesn't actually _start_ a connection, it just tests whether a connection can be made. But the use of `with_connection` to implement `connected?` is part of the `Bolt::Transport::Simple` base class so it's not as obvious to change that.

The only consumer of the connections is this class itself, so I'm not really concerned about defining an API one way or the other. We can freely change it as it evolves.

lib/bolt/transport/ssh/exec_connection.rb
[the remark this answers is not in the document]
nicklewis: Ugh that's unfortunate. I wish we could use `ssh -G` to parse the config (rather than relying on `Net::SSH`) but it appears that's only available in newer SSH versions. :(

lib/bolt/transport/ssh/exec_connection.rb
```diff
@@ -0,0 +1,107 @@
...
+
+          if (key = target.transport_config['private-key'])
+            cmd += ['-i', key]
+          end
+          cmd
+        end
+
+        def build_ssh_command(command)
+          ssh_conf = @target.transport_config['ssh-command']
+          ssh_cmd = Array.new(ssh_conf)
```
nicklewis, line 60: This should be `Array(ssh_conf)` (ie. _convert_ to array). This will fail if `ssh-command` is a string since `Array.new(<string>)` doesn't work.

lib/bolt/transport/ssh.rb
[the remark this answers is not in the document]
lucywyman: That's true, I doubt this will set a precedent in stone if we ever make the API open, which is a huge if anyway. And if we did, it's easy to change. Updated!

lib/bolt/transport/ssh/exec_connection.rb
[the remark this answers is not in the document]
lucywyman: Yeah :(

lib/bolt/config/transport/ssh.rb
```diff
@@ -13,12 +13,18 @@ class SSH < Base
         #       in schemas/bolt-transport-definitions.json
         OPTIONS = {
           "cleanup"            => { type: TrueClass,
+                                    external: true,
                                     desc: "Whether to clean up temporary files created on targets." },
           "connect-timeout"    => { type: Integer,
                                     desc: "How long to wait when establishing connections." },
+          "copy-command"       => { type: Array,
```
beechtom, line 20: This and `ssh-command` should have schema definitions and be added as properties to the SSH definition: https://github.com/puppetlabs/bolt/blob/master/schemas/bolt-transport-definitions.json

nicklewis:
It's a little inconsistent in the code: is `ssh-command` allowed to be a string or no? The CLI accepts a string but converts it to an array (but also accepts an array, which is impossible on the CLI), the config class mandates that it must be an array, but the ExecConnection class handles it being a string _or_ an array (via the `Array()` call). I think it makes the most sense to allow it to be either, in which case the config class shouldn't have a type set for it, the CLI doesn't need to munge it, and the ExecConnection class should handle both forms (as it does now).

`copy-command` on the other hand is more consistent: it _must_ be an array (although again the CLI handles the impossible case that it is an array there).

lucywyman:
ssh-command and copy-command currently behave the same and should continue to. I had both accept a string and array on the CLI because powershell can accept arrays and I didn't see a reason to limit it. I initially wanted to try to enforce using arrays in configfiles because they seem to behave better with shellescaping, but that would definitely cause some inconsistent behavior and we're not shellescaping now anyway. 

I'll update this to allow both options to be a string or array in either cli or config.

schemas/bolt-transport-definitions.json
```diff
@@ -109,6 +111,11 @@
       "type": "integer",
       "min": 1
     },
+    "copy-command": {
+      "description": "Array of command and flags to use when copying files using ssh-command. Bolt runs `<copy-command> <src> <dest>`.",
+      "type": "array"
```
beechtom, line 116:
```suggestion
      "type": ["array", "string"],
      "items": {
        "type": "string"
      }
```

schemas/bolt-transport-definitions.json
```diff
@@ -231,6 +238,10 @@
       "type": "integer",
       "min": 0
     },
+    "ssh-command": {
+      "description": "Array of command and flags to use when SSHing. This enables the external SSH transport which shells out to the specified command.",
+      "type": "array"
```
beechtom, line 243:
```suggestion
      "type": ["array", "string"],
      "items": {
        "type": "string"
      }
```

lucywyman: I already hate this...

lib/bolt/config/transport/ssh.rb
```diff
@@ -13,12 +13,17 @@ class SSH < Base
...
                                     desc: "How long to wait when establishing connections." },
+          "copy-command"       => { external: true,
+                                    desc: "Command to use when copying files using ssh-command. "\
+                                          "Bolt runs `<copy-command> <src> <dest>`. **This option is experimental.**" },
           "disconnect-timeout" => { type: Integer,
                                     desc: "How long to wait before force-closing a connection." },
           "host"               => { type: String,
+                                    external: true,
                                     desc: "Host name." },
           "host-key-check"     => { type: TrueClass,
```
beechtom, line 28:
```suggestion
          "host-key-check"     => { type: TrueClass,
                                              external: true,
```

It looks like the external connection does support this option?

documentation/bolt_configuration_reference.md.erb
```diff
@@ -121,11 +121,19 @@ Transport configuration options can be set in both the configuration file and in
 <% @transports[:options].each do |transport, options| %>
 ### <%= transport %>
 
+<% if transport == 'ssh' %>
+| Option | Description | External SSH | Type | Default |
+| ------ | ----------- | ------------ | ---- | ------- |
```
beechtom, line 126:
```suggestion
| ------ | ----------- | :----------: | ---- | ------- |
```

documentation/bolt_configuration_reference.md.erb
```diff
@@ -121,11 +121,19 @@ Transport configuration options can be set in both the configuration file and in
...
+<% options.each do |option, data| -%>
+  | `<%= option %>` | <%= data[:desc] %> | <%= data[:external] == true ? 'X' : '' %> | <%= data[:type] == TrueClass ? 'Boolean' : data[:type] %> | <%= @transports[:defaults][transport][option].to_s || "" %> |
```
beechtom, line 128:
```suggestion
  | `<%= option %>` | <%= data[:desc] %> | <%= data[:external] == true ? '&#10004;' : '' %> | <%= data[:type] == TrueClass ? 'Boolean' : data[:type] %> | <%= @transports[:defaults][transport][option].to_s || "" %> |
```

An 'X' seems a little confusing since it's unclear if it means the option is or is not available. I think using a symbol that's more affirmative might be clearer?

documentation/bolt_configuration_reference.md.erb
lucywyman: @beechtom Is there a better name for the column? I don't love external ssh...

lucywyman: @hestonhoffman Might also have some ideas ^

beechtom: I don't mind external ssh... if anything, maybe a small blurb before the options saying that not all options are compatible when using external ssh?

documentation/experimental_features.md
```diff
@@ -106,3 +106,34 @@ MODULEPATH:
...
+The value of `ssh-command` can be either a string or array, and you can provide any flags to the
+command. Bolt will append bolt-configuration settings to the command, as well as the specified
+target, when connecting. Not all Bolt configuration options are supported using the external SSH
+transport, but you can configure most options in your OpenSSH Config. See [bolt configuration
+reference](bolt_configuration_reference.md) for the list of supported Bolt SSH options.
+
+Bolt transports have 2 main functions: executing remotely, and copying files to the remote targets.
+`ssh-command` is what configures the remote execution, and `copy-command` configures the copy
+command. The default is `scp -r`, and 'rsync' is not supported at this time.
+
```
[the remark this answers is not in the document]
lucywyman: +1 Updated based on the docs comments! Thanks so much Heston!

beechtom:
Needs a rebase, then ready to merge 👍

PR #1810, opened 2020-05-11
(maint) Add details to file function descriptions

lucywyman:
The file module includes several helpful functions for interacting with
files locally. Previously the documentation didn't clearly specify that
the functions only act on disk and cannot be run against remote targets.
The function documentation now clarifies what exactly each function is
running "under the hood", and states that the functions are limited to
localhost.

### DIFF
--- a/bolt-modules/file/lib/puppet/functions/file/read.rb
+++ b/bolt-modules/file/lib/puppet/functions/file/read.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
-# Read a file and return its contents.
+# Read a file on localhost and return its contents using ruby's `File.read`. This will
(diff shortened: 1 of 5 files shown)


bolt-modules/file/lib/puppet/functions/file/read.rb
```diff
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
-# Read a file and return its contents.
+# Read a file on localhost and return its contents using ruby's `File.read`. This will
+# only read files to the machine you run Bolt on.
```
beechtom, line 4:
```suggestion
# only read files on the machine you run Bolt on.
```

PR #1815, opened 2020-05-12
(maint) Distribute JSON schemas from Puppet Forge

beechtom:
We now have proxies set up to distribute the JSON schemas from
forgeapi.puppet.com. This updates the documentation for using the JSON
schemas from the Forge.

!no-release-note

### DIFF
--- a/schemas/README.md
+++ b/schemas/README.md
@@ -14,22 +14,19 @@ Bolt configuration files, including `bolt.yaml`, `inventory.yaml`, and
 
 ### Enabling schemas
 
-1. Download and save each of the JSON files in this directory.
+Open the [user or workspace settings file](https://code.visualstudio.com/docs/getstarted/settings) and add
+the following:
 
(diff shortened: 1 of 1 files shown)


schemas/bolt-config.schema.json
```diff
@@ -170,22 +170,22 @@
       "type": "string"
     },
     "docker": {
-      "$ref": "bolt-transport-definitions.json#/docker"
+      "$ref": "https://forgeapi.puppet.com/schemas/bolt-transport-definitions.json#/docker"
```
adreyer, line 173: Can we make this a relative ref somehow so that it works with a local schema if the file is local?

beechtom: It looks like just leaving that as `bolt-transport-definitions.json#...` works. I'll update the PR to remove the changes.

lucywyman:
Needs a rebase

PR #1818, opened 2020-05-13
(PE-28674) No longer look for modulepath in bolt.yaml for PE plans

donoghuc:
Attempting to use a modulepath from `bolt.yaml` in PE plans has exposed some issues that we are not ready to tackle yet. For example there is no interpolation for `$basemodulepath`. This commit prevents bolt-server from attempting to load a modulepath from bolt.yaml in an environment. This commit also updates `BoltServer::PE::PAL` to take advantage of the work done in the PAL we use in orchestrator to leverage threadsafe Puppet features.

### DIFF
--- a/lib/bolt_server/pe/pal.rb
+++ b/lib/bolt_server/pe/pal.rb
@@ -6,98 +6,53 @@ require 'bolt/util'
 module BoltServer
   module PE
     class PAL < Bolt::PAL
+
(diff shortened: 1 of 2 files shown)


donoghuc:
Draft for now as we are not on bolt latest yet so fully testing this in a pe stack is not easy.

donoghuc:
Actually I think i'm going to *not* try to use thread safe yet...

PR #1822, opened 2020-05-14
(GH-1816) Use bolt-project.yaml instead of project.yaml

### LINKED ISSUE #1816: Rename project.yaml to bolt-project.yaml
## Use Case
`project.yaml` is a generic name that doesn't give any indication that it's related to Bolt. We should give it a name that implies a relationship with Bolt.

Bolt should load `bolt-project.yaml` instead of `project.yaml` as the project config file. It should also treat a directory containing `bolt-project.yaml` as a project directory, even if that directory does not also contain `bolt.yaml`.

Warn and ignore `project.yaml` if one is detected.

beechtom:
This changes the project configuration file name to `bolt-project.yaml`.
Previously, the file name was `project.yaml`, which was too generic and
did not indicate any connection to Bolt.

If a user has an existing `project.yaml` file and no `bolt-project.yaml`
file, Bolt will issue a warning indicating that `project.yaml` has been
deprecated. Bolt will also no longer load project configuration from a
`project.yaml` file.

This also adds support for specifying a Bolt project directory with a
`bolt-project.yaml`. Previously, local project directories were
specified by the presence of a `bolt.yaml` file. Now, any directory
containing a `bolt-project.yaml` file, even if it does not contain a
`bolt.yaml` file, will be considered a project directory.

Closes #1816 

!feature

* **Specify a local project directory with a `bolt-project.yaml` file** (#1816)

  Directories containing a `bolt-project.yaml` file are now considered
  a [local project directory](https://puppet.com/docs/bolt/latest/bolt_project_directories.html#local-project-directory).

!deprecation

* **Project configuration file changed to `bolt-project.yaml`** (#1816)

  The project configuration file name `project.yaml` has been deprecated
  in favor of a more-specific `bolt-project.yaml`. Bolt will no longer
  load project configuration from a `project.yaml` file.


lib/bolt/project.rb
```diff
@@ -120,9 +120,16 @@ def validate
...
       %w[tasks plans].each do |conf|
         unless @data.fetch(conf, []).is_a?(Array)
-          raise Bolt::ValidationError, "'#{conf}' in project.yaml must be an array"
+          raise Bolt::ValidationError, "'#{conf}' in bolt-project.yaml must be an array"
         end
       end
     end
+
+    def check_deprecated_file
+      if (@path + 'project.yaml').file? && !load_as_module?
```
lucywyman, line 129: I think this should warn even if `bolt-project.yaml` is present.

schemas/README.md
```diff
@@ -23,7 +23,7 @@ Bolt configuration files, including `bolt.yaml`, `inventory.yaml`, and
     ```json
     "yaml.schemas": {
       "<path to bolt-project.schema.json>": [
-        "project.yaml"
+        "bolt-project.yaml"
```
lucywyman, line 26: This also needs to be updated on line 5 of this file.

documentation/bolt_project_directories.md
```diff
@@ -97,7 +113,7 @@ The default paths for all Bolt configuration, code, and data are relative to the
 |[`bolt.yaml`](bolt_configuration_reference.md)|Contains configuration options for Bolt.|
 |`hiera.yaml`|Contains the Hiera config to use for target-specific data when using `apply`.|
 |[`inventory.yaml`](inventory_file_v2.md)|Contains a list of known targets and target specific data.|
-|[`project.yaml`](bolt_configuration_reference.md#project_configuration_options)|Contains configuration for the Bolt project.  The `project.yaml` file contains a whitelist of tasks and plans that you can use to limit the output from the `bolt plan show` and `bolt task show` commands. This is an experimental feature. For more information, see [Bolt projects](./experimental_features.md#bolt-projects).|
+|[`bolt-project.yaml`](bolt_configuration_reference.md#project_configuration_options)|Contains configuration for the Bolt project.  The `bolt-project.yaml` file contains a whitelist of tasks and plans that you can use to limit the output from the `bolt plan show` and `bolt task show` commands. This is an experimental feature. For more information, see [Bolt projects](./experimental_features.md#bolt-projects).|
```
lucywyman, line 116: I think https://github.com/puppetlabs/bolt/blob/master/documentation/bolt_configuration_reference.md.erb#L87 also needs an update. Might also be a good chance to change that link to point to the experimental features page.

lib/bolt/project.rb
beechtom: My only concern with that is it would warn if a user had a `project.yaml` file in their directory that was being used for something else. Super narrow use case, but I figured if they have a `bolt-project.yaml` then they already know `project.yaml` isn't the right file?

lucywyman: Hm, I can see that, but I can also see a repo ending up with both and confusing new users about which is 'right'. If we errored I'd say we shouldn't, but since it's just a warning it's safer to raise it and let the user decide what to do.

beechtom: Fair enough. Updated. :)

Issue #1790, opened 2020-05-06
Add a way for a Plan parameter to hint that it should be treated securely by GUIs


kreeuwijk:
## Use Case
When consuming Bolt Plan content from a GUI (e.g. CD4PE custom deployment policies), it's difficult to have Bolt plan parameters treated securely, since you can't work with the `Sensitive` type as a parameter. Instead, you'll have to use a standard `String` type parameter, in order to be able to pass the parameter to Bolt at execution time. But a higher level GUI that exposes the parameter set of a plan, like CD4PE, is then unable to differentiate between parameter values than it can display in clear text, and parameter values that it should obfuscate to the user.

## Describe the Solution You Would Like
Some way of hinting to the GUI that is consuming Plan content, that certain Plan parameters should be treated securely within the calling application's GUI and logging system. A possible example is to introduce an extra Puppet Strings hint:
```
# @ param param1
# @ sensitive param1
# @ param param2
#
plan test_plan (
  String $param1
) {
  $_param1 = Sensitive($param1)
  out::message("Redacted content: ${_param1}")
}
```
As you can see, it's still up to the Plan author to convert the input to a natively secure variable (so that it can be redacted from Bolt's output going forward). The hint provides just enough info to a calling application that the value of `$param1` should be displayed as `*******` in the GUI.

## Describe Alternatives You've Considered
I've tried using the Sensitive type directly:
```
plan test_plan (
  Sensitive $param1
) {
  out::message("Redacted content: ${param1}")
}
```
but found that I have no way of passing the desired value for param1 to the plan, as the incoming value will be of type `String` and Bolt will then error on the type mismatch. A solution to this problem would be the most efficient & most clean way of supporting sensitive parameters though.

## Additional Context
Example of how CD4PE provides a GUI for Bolt plans:
![cd4pe_plan_ui](https://user-images.githubusercontent.com/26876487/81195720-84099480-8fbe-11ea-91cc-8e0bebea8746.png)
You can see how it would be desirable to natively support the handling of sensitive parameters.

lucywyman:
TODO: 
- [ ] See what makes sense with puppet-strings, and come up with an example (or multiple) to present at standup

reidmv:
[...]

PR #1823, opened 2020-05-14
(GH-1790) Wrap Sensitive plan parameters as Sensitive wrapper type

beechtom:
This updates the `run_plan` function to wrap parameters marked as
`Sensitive` in the Sensitive wrapper type. Previously, indicating a plan
parameter as `Sensitive` would cause a plan to fail, as the parameter is
passed as a `String` type.

Plan authors could get around this limitation
by accepting `String` parameters and then wrapping them as `Sensitive`,
but GUIs that expose plan parameters would still display these
parameters as strings. Now, Bolt will check the type expression for each
parameter on plan startup and wrap any parameters marked as `Sensitive`
as such.

Only parameters with a parameterized or non-parameterized `Sensitive`
type are automatically wrapped. Additionally, these parameters will only
be automatically wrapped for plans run via an API call, as plans run
from another plan can already receive `Sensitive` parameters.

!feature

* **Use `Sensitive` plan parameters with `bolt plan run`**
  ([#1790](#1790))

  Plans now support parameters with the `Sensitive` wrapper type when
  run with the `bolt plan run` command. Parameters marked as `Sensitive`
  will be automatically wrapped with the `Sensitive` wrapper type upon
  plan startup.

### DIFF
--- a/bolt-modules/boltlib/lib/puppet/functions/run_plan.rb
+++ b/bolt-modules/boltlib/lib/puppet/functions/run_plan.rb
@@ -109,6 +109,15 @@ Puppet::Functions.create_function(:run_plan, Puppet::Functions::InternalFunction
       end
     end
 
+    # Wrap Sensitive parameters for plans that are run from the CLI, as it's impossible to pass
+    # a Sensitive value that way. We don't do this for plans run from the run_plan function, as
+    # it can receive Sensitive values as arguments.
+    # This should only happen after expanding target params, otherwise things will blow up if
+    # the targets are wrapped as Sensitive. Hopefully nobody does that, though...
+    if options[:bolt_api_call]
+      params = wrap_sensitive_parameters(params, closure.parameters)
+    end
+
     # wrap plan execution in logging messages
     executor.log_plan(plan_name) do
       result = nil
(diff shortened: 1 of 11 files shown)


lib/bolt/pal.rb
```diff
@@ -390,6 +391,7 @@ def get_plan_info(plan_name)
                        param.type_expr
                      end
           params[name] = { 'type' => type_str }
+          params[name]['sensitive'] = true if param.types.first == 'Sensitive'
```
lucywyman, line 394: This looks to be `type_expr` for yaml plan parameters

bolt-modules/boltlib/lib/puppet/functions/run_plan.rb
```diff
@@ -97,6 +97,15 @@ def run_inner_plan(scope, plan_name, targets, args = {})
       param_acc[param.name] = extract_parameter_types(param.type_expr)&.flatten
```
lucywyman, line 97: extract_parameter_types only gets the rightmost parameter type, so for example `extract_parameter_types(Optional[TargetSpec])` will return just `TargetSpec`. This is fine when we're checking whether to populate a `$targets` since TargetSpec will always be the rightmost type, but for checking `Sensitive` it means the `Sensitive[String]` is a valid type but gets missed by this type checker.

I think the best thing to do would be to have extract_parameter_types traverse `l_expr` recursively in addition to `keys`, `types`, etc. then check whether any of  `param_types[param]` (which should be an array) is sensitive.

bolt-modules/boltlib/lib/puppet/functions/run_plan.rb
```diff
@@ -169,6 +178,16 @@ def extract_parameter_types(type_expr)
     end
   end
 
+  # Determine if the parameter is Sensitive and should be wrapped.
+  def sensitive?(type_expr)
+    # Parameterized Sensitive type (e.g. Sensitive[String])
+    (defined?(type_expr.left_expr) && type_expr.left_expr.cased_value == 'Sensitive') ||
```
lucywyman, line 184: This feels like the right level of complexity to handle for types. I think we should add a check where if all of these are false and `type_expr.to_s.include?('Sensitive')` is true we raise warning that we can't handle complex types that include sensitive (for example `Variant[Sensitive[String], Array]`).

beechtom: Playing around with this I realized that we probably _don't_ want to warn when using a complex type with `Sensitive`. 

Complex data types using `Sensitive` _are_ supported by `run_plan()`, just not by `bolt plan run`. Something like this works since the `outer` plan is able to wrap the parameter as `Sensitive` before passing it to the `inner` plan:

```
plan inner(
  Variant[Sensitive[String], Array[String]] $foo
) {
  return $foo.unwrap
}

plan outer(
  String $foo
) {
  $result = run_plan('inner', 'foo' => Sensitive($foo))
  return $result
}

$ bolt plan run outer foo=foo
```

I've updated the function to not automatically wrap a parameter if it's _already_ `Sensitive`, since I didn't take this into account originally. But I'm not really sure if there's a way for us to easily distinguish when a complex data type using `Sensitive` is and is not supported, or if it's desirable in general. Maybe it's more of a documentation issue?

beechtom: Playing around with this I realized that we probably _don't_ want to warn when using a complex type with `Sensitive`. 

Complex data types using `Sensitive` _are_ supported by `run_plan()`, just not by `bolt plan run`. Something like this works since the `outer` plan is able to wrap the parameter as `Sensitive` before passing it to the `inner` plan:

```
plan inner(
  Variant[Sensitive[String], Array[String]] $foo
) {
  return $foo.unwrap
}

plan outer(
  String $foo
) {
  $result = run_plan('inner', 'foo' => Sensitive($foo))
  return $result
}

$ bolt plan run outer foo=foo
```

I've updated the code to not automatically wrap a parameter that is _already_ a `Sensitive` type, since I didn't take that into account previously.

I'm not really sure if there's a way for us to easily distinguish when a complex data type using `Sensitive` is and is not supported, or if it's desirable in general. Maybe it's more of a documentation issue?

beechtom:
I'd still like to write a couple tests for this, but other than that this should have the behavior we want.

documentation/writing_plans.md
```diff
@@ -126,6 +126,25 @@ Similarly, parameters are made available to the task as environment variables wh
 
 - [Task metadata types](writing_tasks.md#)
 
+### Sensitive parameters
+
+Parameters that are sensitive and should not be displayed in logs can be masked using the `Sensitive` data type.
```
hestonhoffman, line 131:
```suggestion
Use the `Sensitive` data type to mask parameters that should not be displayed in logs.
```

documentation/writing_plans.md
```diff
@@ -126,6 +126,25 @@ Similarly, parameters are made available to the task as environment variables wh
...
+When passing a value to a `Sensitive` parameter, Bolt will automatically mask the value before the plan is run.
```
hestonhoffman, line 132:
```suggestion
When you pass a value to a `Sensitive` parameter, Bolt automatically masks the value before the plan is run.
```

documentation/writing_plans.md
```diff
@@ -126,6 +126,25 @@ Similarly, parameters are made available to the task as environment variables wh
...
+The unmasked value can be accessed by calling the `unwrap` function on the parameter.
```
hestonhoffman, line 133:
```suggestion
To access the unmasked value, call the `unwrap` function on the parameter.
```

documentation/writing_plans.md
```diff
@@ -126,6 +126,25 @@ Similarly, parameters are made available to the task as environment variables wh
 
 - [Task metadata types](writing_tasks.md#)
 
+### Sensitive parameters
+
+Use the `Sensitive` data type to mask parameters that should not be displayed in logs can be masked using the 
+`Sensitive` data type. When you pass a value to a `Sensitive` parameter, Bolt automatically masks the value
+before the plan is run. To accewss the unmasked value, call the `unwrap` function on the parameter.
```
hestonhoffman, line 133:
```suggestion
Use the `Sensitive` data type to mask parameters that should not be displayed in logs. 

When you pass a value to a `Sensitive` parameter, Bolt automatically masks the value
before the plan is run. 

To access the unmasked value, call the `unwrap` function on the parameter.
```

PR #1825, opened 2020-05-15
(GH-1789) Rescue EMFILE error with helpful message

### LINKED ISSUE #1789: Too Many Open Files Error
## Describe the Bug
Executed `bolt command run 'hostname'` against a set of 75 windows servers.  This failed on some, but not all, of the servers with the large message attached, followed by a series of

Failed on hostname.removed:
  Too many open files

and one error of 

Failed on hostname.removed
  Failed to connect to <hostname>: Too many open files @ rb_sysopen - /opt/puppetlabs/bolt/lib/ruby/gems/2.5.0/gems/winrm-2.3.4/lib/winrm/psrp/create_pipeline.xml.erb

Subsequent runs showed only the simple "Too many open files" failure on 2-3 targets - different ones each time - per run.
[too_many_files_error.txt](https://github.com/puppetlabs/bolt/files/4584799/too_many_files_error.txt)

## Expected Behavior
bolt executes the "hostname" command successfully on 75 or more remote windows servers without this error.

## Steps to Reproduce
1. create an inventory file with 75 windows hosts
2. execute bolt command run 'hostname' -t windows

## Environment
 Mac OS X Catalina 10.15.4
 ulimit -n = 256
 bolt 2.7.0

## Additional Context
Problem was resolved by adding ulimit -n 1024 to .bash_profile and launching a new terminal window.  I expect it wasn't actually resolved, but rather pushed out - I think it could be triggered again under the right condition.  Nick L/Lucy W suggested this fix.

Additional commentary from the discussion thread :

> Well, I at least understand what's raising the message, but not why. The WinRM gem seems to use 'Powershell Remoting Protocol' to encode messages to send over WinRM, and it does so by using a template that it reads into memory and populates. It's trying to do that when it fails saying too many files are open. It also sounds like "too many" is usually 1024, and that the fact the error is Errno::EMFILE indicates that it's too many files in the ruby process itself and not on the system.

> That problem could have been made worse by the recent addition of a lot more internal pipes.

> We might be using 5+ fds for each target

lucywyman:
Since splitting the shell implementation from the transport connecting
to the shell, we now open a new thread with input, output, and error
streams for every connection. With concurrency defaulting to 100 this
often means 300+ file descriptors are open in Bolt's ruby process when
running against many targets, which exceeds the common FD limit of 256.
We now set concurrency to 1/3 the file descriptor limit if the limit is
below 300, otherwise it's set to 100. We also warn the user that
concurrency is set low, and provide instructions on how to increase the
ulimit or concurrency. We also raise a helpful error message when the
EMFILE error is thrown instructing the user on how to increase their FD
limit. Lastly this adds a section in the known issues documentation about
the error, also with instructions on increasing the FD limit and on
dialing back concurrency.

Closes #1789

!feature

* **Lower default concurrency when ulimit is low** ([1789](#1789))

  Concurrency defaults to 1/3 the ulimit if ulimit is below 300, and
  warns if lowered concurrency is used.

### DIFF
--- a/documentation/bolt_known_issues.md
+++ b/documentation/bolt_known_issues.md
@@ -28,7 +28,10 @@ As a workaround, you can generate new keys with the ssh-keygen `-m PEM` flag. Fo
 
 ## Commands fail in remote Windows sessions
 
-Interactive tools fail when run in a remote PowerShell session. For example, using `--password` to prompt for a password when running Bolt triggers an error. As a workaround, consider putting the password in `bolt.yaml` or an inventory file, or passing the password on the command line. ([BOLT-1075](https://tickets.puppetlabs.com/browse/BOLT-1075))
+Interactive tools fail when run in a remote PowerShell session. For example, using
+`--password-prompt` to prompt for a password when running Bolt triggers an error. As a workaround,
+consider putting the password in `bolt.yaml` or an inventory file, or passing the password on the
+command line. ([BOLT-1075](https://tickets.puppetlabs.com/browse/BOLT-1075))
 
 ## Unable to authenticate with ed25519 keys over SSH transport on Windows
 
@@ -44,3 +47,23 @@ unsupported key type `ssh-ed25519'
 ## Limited Kerberos support
 
 Support for Kerberos over WinRM from a Linux host is currently experimental and requires the [MIT Kerberos library](https://web.mit.edu/Kerberos/www/krb5-latest/doc/admin/install_clients.html) to be installed. In the future, Bolt will support Kerberos when running on Windows ([BOLT-1323](https://tickets.puppet.com/browse/BOLT-1323)) and macOS ([BOLT-1471](https://tickets.puppet.com/browse/BOLT-1471)).
+
+## Errno::EMFILE Too many open files
+
+This error is raised when there are too many files open in Bolt's Ruby process. To see what
+your current limit is, run:
+
+```
+ulimit -n
+```
+
+To raise the limit, set the following in your shell configuration file (For example,
+`~/.bash_profile`):
+
+```
+ulimit -n 1024
+```
+
+You can also set Bolt's concurrency lower to have fewer file descriptors opened at once. The default
+is 100, and you can use `--concurrency` on the CLI, or set `concurrency: <CONCURRENCY>` in [Bolt
+config](configuring_bolt.md)
--- a/lib/bolt/config.rb
+++ b/lib/bolt/config.rb
@@ -66,7 +66,6 @@ module Bolt
 
     DEFAULT_OPTIONS = {
       "color" => true,
-      "concurrency" => 100,
       "compile-concurrency" => "Number of cores",
       "format" => "human",
       "hiera-config" => "Boltdir/hiera.yaml",
@@ -102,6 +101,8 @@ module Bolt
     DEFAULT_APPLY_SETTINGS = {
       "show_diff" => false
     }.freeze
+
+    DEFAULT_DEFAULT_CONCURRENCY = 100
 
     def self.default
       new(Bolt::Project.new('.'), {})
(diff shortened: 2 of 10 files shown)


lucywyman:
It seems like there's 2 tactics we can use to make this less likely to error
1. Be careful about how many file descriptors we have open
2. Try to handle the error by pausing execution until fewer FDs are being used

The second option seems difficult given we don't know what the file descriptor limit is, or how many targets are left to execute on. How long should we wait before resuming execution? Will this make Bolt unbearably slow? I think we *could* do this, but I'm not sure we *should*.

The first option is more promising. The most obvious thing we could do is check the users ulimit and set concurrency relative to that (maybe ulimit/5?). I'm wary of having the shell need to be aware of the transport and vice-versa, but we could have a `shell.execute_with_sudo` for transports that require writing to stdin, and have `shell.execute` run the command directly on the transport and only read stdout and err.

documentation/bolt_known_issues.md
```diff
@@ -44,3 +47,23 @@ unsupported key type `ssh-ed25519'
 ## Limited Kerberos support
 
 Support for Kerberos over WinRM from a Linux host is currently experimental and requires the [MIT Kerberos library](https://web.mit.edu/Kerberos/www/krb5-latest/doc/admin/install_clients.html) to be installed. In the future, Bolt will support Kerberos when running on Windows ([BOLT-1323](https://tickets.puppet.com/browse/BOLT-1323)) and macOS ([BOLT-1471](https://tickets.puppet.com/browse/BOLT-1471)).
+
+## Errno::EMFILE Too many open files
+
+This error is raised when there are too many files open in Bolt's ruby process. You can see what
+your current limit is by running
```
hestonhoffman, line 54:
```suggestion
This error is raised when there are too many files open in Bolt's Ruby process. To see what
your current limit is, run:
```

documentation/bolt_known_issues.md
```diff
@@ -44,3 +47,23 @@ unsupported key type `ssh-ed25519'
...
+
+```
+ulimit -n
+```
+
+and can raise the limit by setting the following in your shell configuration (ie.
+`~/.bash_profile`):
```
hestonhoffman, line 61:
```suggestion
To raise the limit, set the following in your shell configuration file (For example,
`~/.bash_profile`):
```

lib/bolt/config.rb
```diff
@@ -183,6 +184,18 @@ def initialize(project, config_data, overrides = {})
...
 
+      # If we need to lower concurrency and concurrency is not configured
+      ld_concurrency = loaded_data.map(&:keys).flatten.include?('concurrency')
+      if default_concurrency != DEFAULT_DEFAULT_CONCURRENCY &&
+         !ld_concurrency &&
+         !override_data.key?('concurrency')
+        concurrency_warning = { option: 'concurrency',
+                                msg: "Concurrency will default to #{default_concurrency} because ulimit "\
+                                "is low: #{Etc.sysconf(Etc::SC_OPEN_MAX)}. Set concurrency with "\
+                                "'--concurrency', or set your ulimit with 'ulimit -n <limit>'" }
```
beechtom, line 195: I wonder if we could send this to `info` instead of `warn` and update the docs to say concurrency goes _up to_ 100.. a warning for a default value seems a little weird to me, especially since we'll have a similar message to this when a low ulimit actually becomes a problem that the user needs to make adjustments for.

lucywyman: I think it's appropriate to warn here because it can cause unexpected behavior, particularly if a user typically runs bolt on a machine that has a higher ulimit - namely setting the concurrency lower can cause Bolt to run more slowly. I think of info messages as being useful but not necessary, things like what steps are being performed or where modules are being loaded from. This message feels slightly more important for users to be aware of and potentially take action on. It's not an error because nothing is wrong, but it's probably not a desirable state for Bolt to run in.

I do think I can update the default in the docs to be "100 or 1/3 of ulimit, whichever is lower"

PR #1828, opened 2020-05-19
(BOLT-1549) Add puppet data types to plugin tarbal

hajee:
Right now the plugin tarbal doesn't contain the puppet data types defined in the module directory `types` with extension `.pp`. As a result of that puppet code that requires these puppet data types not only for catalog compilation but also for run-time type checking, cannot run in a bolt apply block. 

This is related to  https://tickets.puppetlabs.com/browse/PUP-7197.

### DIFF
--- a/lib/bolt/applicator.rb
+++ b/lib/bolt/applicator.rb
@@ -34,6 +34,8 @@ module Bolt
           search_dirs << mod.plugins if mod.plugins?
           search_dirs << mod.pluginfacts if mod.pluginfacts?
           search_dirs << mod.files if mod.files?
+          type_files = "#{mod.path}/types"
+          search_dirs << type_files if File.exist?(type_files)
           search_dirs
         end
       end


lucywyman:
@hajee I think this can be solved without any changes by running `bolt puppetfile generate-types` before executing the plan. https://puppet.com/docs/bolt/latest/bolt_command_reference.html#puppetfile-generate-types

hajee:
@lucywyman This is not my experience. As far as I know, the generate types creates a puppet representation of the custom types available. Not of the data types.

lib/bolt/applicator.rb
```diff
@@ -34,6 +34,8 @@ def initialize(inventory, executor, modulepath, plugin_dirs, pdb_client, hiera_c
           search_dirs << mod.plugins if mod.plugins?
           search_dirs << mod.pluginfacts if mod.pluginfacts?
           search_dirs << mod.files if mod.files?
+          type_files = "#{mod.path}/types"
```
donoghuc, line 37: Do agents usually download these from puppetserver when applying a catalog?

hajee: See here:  https://tickets.puppetlabs.com/browse/PUP-7197.

hajee:
@lucywyman Thank you

PR #1829, opened 2020-05-19
(BOLT-1550) Fix couldn't find login name error

hajee:
* **Fix couldn't find login name when running non-interactive** ([BOLT-1550](https://tickets.puppetlabs.com/browse/BOLT-1550))

When running bolt in a non-interactive process, Bolt has issue’s finding the home directory of the current process. This fix, makes sure the home directory is found regardless of mode the process is running in.

!lbug

### DIFF
--- a/lib/bolt/project.rb
+++ b/lib/bolt/project.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require 'pathname'
+require 'etc'
 require 'bolt/pal'
 
 module Bolt
@@ -15,8 +16,16 @@ module Bolt
     attr_reader :path, :config_file, :inventory_file, :modulepath, :hiera_config,
                 :puppetfile, :rerunfile, :type, :resource_types
 
+    def self.home_dir
+      if Gem.win_platform?
+        Dir.home
+      else
+        Etc.getpwuid(Process.uid).dir
+      end
+    end
+
     def self.default_project
-      Project.new(File.join('~', '.puppetlabs', 'bolt'), 'user')
+      Project.new(File.join(home_dir, '.puppetlabs', 'bolt'), 'user')
     end
 
     # Search recursively up the directory hierarchy for the Project. Look for a


lucywyman:
Hi @hajee - thank you so much for putting this up! For the file path in `config.rb` we should already handle the case where `~` doesn't properly expand (see #1676 ).  I wonder if we should set a different default for the project directory if the user doesn't have a home dir? Or maybe just have an "empty project", since I don't think users would expect any project to be loaded.

hajee:
@lucywyman I really don't know enough about that part to say something sane about that. So I'm ready to follow your lead. 

Would you like me to remove the changes in the PR in the config.rb?

lucywyman:
I think the changes in config.rb are safe to remove. What error were you getting with the project directory path?

hajee:
I have removed the config changes. The error we got was:

When running a bolt plan command from a scheduled bolt task (e.g. a detached process) we get this error:

Error: Exited 1:
/opt/puppetlabs/bolt/lib/ruby/gems/2.5.0/gems/bolt-2.9.0/lib/bolt/project.rb:40:in `expand_path': couldn't find login name – expanding `~' (ArgumentError)
from /opt/puppetlabs/bolt/lib/ruby/gems/2.5.0/gems/bolt-2.9.0/lib/bolt/project.rb:40:in `expand_path'
from /opt/puppetlabs/bolt/lib/ruby/gems/2.5.0/gems/bolt-2.9.0/lib/bolt/project.rb:40:in `initialize'
from /opt/puppetlabs/bolt/lib/ruby/gems/2.5.0/gems/bolt-2.9.0/lib/bolt/project.rb:19:in `new'
from /opt/puppetlabs/bolt/lib/ruby/gems/2.5.0/gems/bolt-2.9.0/lib/bolt/project.rb:19:in `default_project'
from /opt/puppetlabs/bolt/lib/ruby/gems/2.5.0/gems/bolt-2.9.0/lib/bolt/project.rb:33:in `find_boltdir'
from /opt/puppetlabs/bolt/lib/ruby/gems/2.5.0/gems/bolt-2.9.0/lib/bolt/cli.rb:120:in `parse'
from /opt/puppetlabs/bolt/lib/ruby/gems/2.5.0/gems/bolt-2.9.0/exe/bolt:9:in `<top (required)>'
from /opt/puppetlabs/bolt/bin/bolt:23:in `load'
from /opt/puppetlabs/bolt/bin/bolt:23:in `<main>'

hajee:
@lucywyman What is the status of this PR?

lucywyman:
@hajee I think this solution means users on windows without a homedir will still have the same problem you ran into, and further if the user doesn't have a homedir I'm not sure it makes sense to try to load a different path. I think the best solution for this is to add a `begin...rescue...` around loading the path that will default it to this value: https://github.com/puppetlabs/bolt/blob/master/lib/bolt/config.rb#L139

If you'd like you can either make the changes yourself, or give me permissions to your branch and I can push them up (so that you still get credit in our changelog for the PR!), or I can just open a separate branch. Let me know which you prefer!

hajee:
@lucywyman Whatever is fastest. Since I cannot commit, I suspect you making your branch is fastest.

beechtom:
Closed in favor of #1849

PR #1839, opened 2020-05-21
(GH-1785) Warn when applying a manifest that only contains definitions

### LINKED ISSUE #1785: Bolt Apply should warn when run with a do nothing manifest
## Use Case
Often users will try to use `bolt apply` with a manifest that defines classes or defines without declaring anything. This is typically a problem with the bolt apply command, not an apply block.

Bolt should warn in this case. The warning should link to documentation about classification? We should make a specific section (page?) about the `bolt apply` command in the Bolt docs and make this a section in it. "How to apply puppet code with bolt" talk about declaring.

Should this be based on examining the AST, checking for an empty catalog, or both? Examining the AST can be done for all targets at once, whereas empty catalog checking has to be per target. We should check for if there's *only* a Definition in the AST and warn.

Should this be for just 'apply' or also for apply blocks? We should only do it for 'apply' unless it's unreasonably hard to limit it to just 'apply'.

beechtom:
This updates the `bolt apply` command to warn a user when they are
applying a manifest that only consists of definitions without declaring
anything.

This also updates the 'Applying manifest blocks' documentation to document
using the `bolt apply` command. It also reorganizes the structure of the
page to create separate sections for the `bolt apply` command and the
`apply` plan function.

Lastly, this adds support for the `--hiera-config` option to the 
`bolt apply` command.

Closes #1785

!feature

* **Warn when applying manifests that only contain definitions** (#1785)

  Applying a manifest that only contains definitions with the `bolt apply`
  command will now display a warning that no changes will be applied to
  the targets.

!feature

* **Support `--hiera-config` option when using `bolt apply`** (#1839)

  The `--hiera-config` option can now be used with the `bolt apply`
  command to specify the path to a Hiera configuration file.

### DIFF
--- a/documentation/applying_manifest_blocks.md
+++ b/documentation/applying_manifest_blocks.md
@@ -1,73 +1,202 @@
-# Applying manifest blocks
+# Applying Puppet manifests
 
-Within a plan, you can use Bolt to apply blocks of Puppet code (manifest blocks) to remote targets. 
+You can use Bolt to apply blocks of Puppet code, called manifest blocks, to remote targets.
 
-Similar to the `puppet apply` command, which applies a standalone Puppet manifest to a local system, the Bolt `apply` command uses manifest blocks to pass code to remote targets from the command line. You can create manifest blocks that use existing content from the Forge, or mix declarative resource configuration via manifest blocks with procedural orchestration and action in a plan. Most features of the Puppet language are available in a manifest block: classes, custom resource types, and functions. For information about what language features aren't supported, see [Manifest block limitations](applying_manifest_blocks.md#).
+You can create manifest blocks that use existing content from the Forge, or mix declarative resource
+configuration via manifest blocks with procedural orchestration and action in a plan. Most features
+of the Puppet language are available in a manifest block: classes, custom resource types, and
+functions. For information about what language features aren't supported, see [Manifest block
+limitations](#manifest-block-limitations).
 
-> 🔩 **Tip:** If you installed Bolt as a Ruby Gem, make sure you have installed the core modules required to use the `puppet apply` command. These modules are listed in the [Bolt GitHub repository](https://github.com/puppetlabs/bolt/blob/master/Puppetfile)and you can install them using a Puppetfile.
+> 🔩 **Tip:** If you installed Bolt as a Ruby Gem, make sure you have installed the core modules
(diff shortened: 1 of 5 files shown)


lib/bolt/cli.rb
```diff
@@ -562,6 +571,12 @@ def apply_manifest(code, targets, filename = nil, noop = false)
       results.ok ? 0 : 1
     end
 
+    def do_nothing_manifest?(ast)
+      ast.body.is_a?(Puppet::Pops::Model::Definition) ||
```
lucywyman, line 575: Hm, this seems a bit flawed. It doesn't catch some basic cases, like a completely empty manifest file or a class with no resources, while it *does* warn for some manifests that do define resources, such as a manifest that has node definitions:
```
# test.pp
node default {
file { '/root/foo.txt': ensure => present }
}
```
I think the main point of this ticket is to warn people who have neglected to declare their classes, which it does, in which case we may want to refine this function to recursively inspect NodeDefinitions and to include `Puppet::Pops::Model::Nop` as a check, though it's hard to be sure those are the only false cases.

Another approach would be to look for resource statements, either by stringifying the ast or walking it, neither of which is appealing...

nicklewis: I was thinking the purpose of the ticket is to warn people who are trying to apply a class via `bolt apply <modules>/manifests/<class>.pp` rather than by declaring it.

I don't think we should warn in the `NodeDefinition` case as node definitions are declared implicitly. If we're not warning there, then I also don't think we need to worry about the `BlockExpression` case as ordinary class/define manifests are not allowed to have more than a single definition (they may technically be allowed to have typedefs, but I wouldn't worry about that).

So that would make this simply `if ast.body.is_a?(Puppet::Pops::Model::HostClassDefinition) || ast.body.is_a?(Puppet::Pops::Model::ResourceTypeDefinition)`

nicklewis: I guess to put it another way, my thought for this ticket was that we want to identify a _usage_ error (you invoked `bolt apply` with what is effectively a bad argument) vs. a _logical_ error (you ran an apply that generated no resources).

documentation/applying_manifest_blocks.md
```diff
@@ -1,126 +1,265 @@
 # Applying manifest blocks
```
hestonhoffman, line 1:
```suggestion
# Applying Puppet manifests
```

PR #1840, opened 2020-05-22
(maint) Bump python-task-helper to 0.4.3

lucywyman:
We manually released a version of python-task-helper that contained a
virtual environment in it. We had to update the version of the helper to
republish to the forge. This bumps the version to the new module that
doesn't contain a venv.


beechtom:
Commit & PR title should be updated to 0.4.3

PR #1844, opened 2020-05-22
(GH-1836) Add PuppetObject interface to Bolt data types

### LINKED ISSUE #1836: Plans error on apply when targets share identical config
## Describe the Bug
When calling `apply` on multiple targets that share identical configuration, the plan will fail with an obscure error:

```
Apply failed to compile for TARGET: undefined method `[]' for #<Bolt::ApplyTarget:0x000000000297cfa0> on node TARGET
```

This is caused by an update to the applicator that sets `local_references: true` when serializing targets. When target objects are serialized, any targets that have similar config to another target will receive references to that config. For example, target B shares similar ssh cofiguration to target A:

```
targets:
  - name: targetA
    config:
      ssh:
        host-key-check: false
        run-as: root
        tty: true
  - name: targetB
    config:
      ssh:
        host-key-check: false
        run-as: root
        tty: true
```

When the targets are serialized, target B will receive a reference to target A's ssh configuration, resulting in a catalog that looks like this for target B:

```
...
"mygroup"=>	
  [{"__ptype"=>"Target",	
    "name"=>"targetB",	
    "config"=>{"transport"=>"ssh", "ssh"=>{"__ptype"=>"LocalRef", "__pvalue"=>"$['targetA'][0]['config']['ssh']"}},
...
```

When this is deserialized, Puppet will see the reference to target A, which has already been deserialized and is now an `ApplyTarget` object, and attempt to access its config with the `[]` method, causing the plan to fail.

## Expected Behavior
Targets should be deserialized without raising errors, even if they share similar config.

## Possible Solutions
- Have Bolt deserialize itself?
- Have Bolt replace references prior to deserialization?
- Update Puppet to not access the deserialized object and use the data hash instead?
- Something else?

beechtom:
This adds the `Puppet::Pops::Types::PuppetObject` interface to Bolt data
types and their associated classes. Previously, Bolt's data types were
missing this interface, which could lead to Puppet attempting to access
already-deserialized objects with the `[]` operator during the
deserialization process when local references are present. Since Bolt's
implementation classes do not implement the `[]` operator, and doing so
is undesirable, adding the `PuppetObject` interface is needed for
deserialization.

Closes #1836 

!bug

* **Add `PuppetObject` interface to Bolt data types**
  ([#1836](https://github.com/puppetlabs/bolt/issues/1836))

  Bolt data types would sometimes not be deserialized correctly when
  using `apply` blocks in plans. All Bolt data types now implement the
  `PuppetObject` interface so they can be deserialized correctly.

### DIFF
--- a/bolt-modules/boltlib/lib/puppet/datatypes/target.rb
+++ b/bolt-modules/boltlib/lib/puppet/datatypes/target.rb
@@ -8,8 +8,6 @@ Puppet::DataTypes.create_type('Target') do
     target_implementation_class = Bolt::Target
   end
 
-  require 'bolt/target'
-
   interface <<-PUPPET
     attributes => {
       uri => { type => Optional[String[1]], kind => given_or_derived },
@@ -34,5 +32,7 @@ Puppet::DataTypes.create_type('Target') do
     }
   PUPPET
 
+  # Needed for Puppet to recognize targets as Puppet objects when deserializing
+  target_implementation_class.include(Puppet::Pops::Types::PuppetObject)
   implementation_class target_implementation_class
 end
--- a/spec/bolt/catalog_spec.rb
+++ b/spec/bolt/catalog_spec.rb
@@ -13,6 +13,7 @@ describe Bolt::Catalog do
   let(:target) { inventory.get_target(uri) }
   let(:inventory) { Bolt::Inventory.empty }
   let(:executor) { Bolt::Executor.new }
+  let(:boltlib) { Bolt::PAL::BOLTLIB_PATH }
   let(:pdb_config) do
     Bolt::PuppetDB::Config.new('server_urls' => 'https://localhost:8081',
                                'cacert' => '/path/to/cacert',
@@ -53,25 +54,58 @@ describe Bolt::Catalog do
           'domain' => nil
         }
       },
-      'inventory' => {
-        'data' => {},
-        'target_hash' => {
-          'target_vars' => {},
-          'target_facts' => {},
-          'target_features' => {}
-        },
-        'config' => {
-          'transport' => "ssh",
-          'transports' => {
-            'ssh' => { 'connect-timeout' => 10, tty: false, "host-key-check" => true },
-            'winrm' => { 'connect-timeout' => 10, tty: false, ssl: true, "ssl-verify" => true },
-            'pcp' => { 'connect-timeout' => 10,
-                       'tty' => false,
-                       'task-environment' => 'production' },
-            'local' => { 'connect-timeout' => 10, tty: false }
-          }
+      'config' => {
+        'transport' => "ssh",
+        'transports' => {
+          'ssh' => { 'connect-timeout' => 10, tty: false, "host-key-check" => true },
+          'winrm' => { 'connect-timeout' => 10, tty: false, ssl: true, "ssl-verify" => true },
+          'pcp' => { 'connect-timeout' => 10,
+                     'tty' => false,
+                     'task-environment' => 'production' },
+          'local' => { 'connect-timeout' => 10, tty: false }
         }
       } }
+  end
+
+  let(:plan_vars) do
+    {
+      "t1" => [{
+        "__ptype" => "Target",
+        "uri" => "targeta.example.com",
+        "name" => "targetA",
+        "safe_name" => "targetA",
+        "target_alias" => [],
+        "config" => {
+          "ssh" => {
+            "host-key-check" => false,
+            "connect-timeout" => 100
+          }
+        },
+        "facts" => {},
+        "vars" => {},
+        "features" => [],
+        "plugin_hooks" => {},
+        "resources" => {}
+      }],
+      "t2" => [{
+        "__ptype" => "Target",
+        "uri" => "targetb.example.com",
+        "name" => "targetB",
+        "safe_name" => "targetB",
+        "target_alias" => [],
+        "config" => {
+          "ssh" => {
+            "__ptype" => "LocalRef",
+            "__pvalue" => "$['t1'][0]['config']['ssh']"
+          }
+        },
+        "facts" => {},
+        "vars" => {},
+        "features" => [],
+        "plugin_hooks" => {},
+        "resources" => {}
+      }]
+    }
   end
 
   it 'instantiates' do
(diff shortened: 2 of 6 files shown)


lib/bolt/apply_target.rb
```diff
@@ -30,6 +30,10 @@ def self.from_asserted_args(uri = nil,
     end
     # rubocop:enable Lint/UnusedMethodArgument
 
+    def self._pcore_type
+      Target
```
lucywyman, line 34: This should be `ApplyTarget`. It also seems like we don't need to define this if the datatype includes `Puppet::Pops::Types::PuppetObject`? Is there a reason to set both?

beechtom: I don't think this would be `ApplyTarget` since we don't define that as a Puppet type. The `Bolt::ApplyTarget` class is just an implementation class for the `Target` Puppet type.

`_pcore_type` is used by `ToDataConverter` here: https://github.com/puppetlabs/puppet/blob/master/lib/puppet/pops/serialization/to_data_converter.rb#L247 . It's probably not strictly necessary to have this since apply targets are never serialized.

lucywyman: It actually looks like this is already defined by the datatype creator? https://github.com/puppetlabs/puppet/blob/master/lib/puppet/datatypes.rb#L167 And...it seems like that overrides this local definition whether it's there or not? Which makes sense - I'm not sure where in this file `Target` would even be defined.

bolt-modules/boltlib/lib/puppet/datatypes/target.rb
```diff
@@ -34,5 +35,8 @@
     }
   PUPPET
 
+  # Needed for Puppet to recognize targets as Puppet objects when deserializing
+  Bolt::Target.include(Puppet::Pops::Types::PuppetObject)
```
nicklewis, line 39: This should be `target_implementation_class.include(Puppet::Pops::Types::PuppetObject)` so it stays unattached to any particular implementation class. We can also remove the `require 'bolt/target'` above.

lib/bolt/apply_target.rb
nicklewis: But `Target` here is a reference to `Bolt::Target`, not to the Puppet Target datatype. Something seems fishy.

nicklewis: @lucywyman That code you linked is really suspicious. It does look like it's meant to make exactly what we're doing here unnecessary. But it checks whether `implementation_class` already includes `PuppetObject`, and if so, it does some work that involves including `PuppetObject`. I wonder if the conditional is simply meant to be reversed?

nicklewis: I don't think the problem is only that the logic is backward... The alternate case calls `register_implementation` which looks to be the way that Puppet associates a Ruby type with a Puppet type if the Ruby type _doesn't_ have `_pcore_type`. In that case, it may just be that the `include PuppetObject` that Puppet does is redundant (since it's already guaranteed the class descends from `PuppetObject`). But if so, that means we _only_ need to include the class and it will handle the proper definition of `_pcore_type` for us.

spec/bolt/catalog_spec.rb
```diff
@@ -96,5 +161,9 @@
       expect(result['resources'].map { |r| r['type'] }).to include('Notify')
       expect(result['resources'].count).to eq(4)
     end
+
+    it 'resolves local references when deserializing Puppet types' do
+      expect { catalog.compile_catalog(reference_request) }.not_to raise_error(/undefined method `\[\]'/)
```
nicklewis, line 166: This test doesn't seem to fail when I run it _without_ the changes.

beechtom: That's interesting... it's failing for me when run on master.

nicklewis: Ah I was running against a local checkout of Puppet. Apparently it was hitting a different error instead, which technically meets this condition. Making this just `expect { ... }.not_to raise_error` might be less sensitive.

beechtom: Updated

lib/bolt/apply_target.rb
lucywyman: The reason I think the method I linked is what ends up being used is that when I call `_pcore_type` on this class I get a `Puppet::Pops::Types::PObjectType`, which...seems unlikely to be coming from this call, but seems very likely to be coming from that method.

All of this to say: I agree with you Nick that I think we only need to include PuppetObject on the datatypes themselves, and don't need this definition (which...seems like it will only be confusing)

spec/bolt/catalog_spec.rb
```diff
@@ -59,21 +60,85 @@
...
-          'transport' => "ssh",
-          'transports' => {
-            'ssh' => { 'connect-timeout' => 10, tty: false, "host-key-check" => true },
-            'winrm' => { 'connect-timeout' => 10, tty: false, ssl: true, "ssl-verify" => true },
-            'pcp' => { 'connect-timeout' => 10,
-                       'tty' => false,
-                       'task-environment' => 'production' },
-            'local' => { 'connect-timeout' => 10, tty: false }
-          }
+        }
```
lucywyman, line 63: Not to drag this out, but while we're updating this 'inventory' can be removed entirely from this, it looks like it's not used anywhere in `catalog.rb` or the test itself.

lucywyman:
I think the Gettext error is happening because we now load the type loader during catalog compilation, and puppet's gettext conflicts with...r10k's? https://tickets.puppetlabs.com/browse/BOLT-914 It doesn't seem to be an issue when actually running bolt, we may need to set something in the spec_helper or rakefile

libexec/bolt_catalog
```diff
@@ -1,8 +1,8 @@
 #!/usr/bin/env ruby
 # frozen_string_literal: true
 
-require 'bolt'
-require 'bolt/catalog'
+require_relative '../lib/bolt'
```
lucywyman, line 4: Do we still need this change?

spec/spec_helper.rb
```diff
@@ -16,6 +16,12 @@
 ENV['RACK_ENV'] = 'test'
 $LOAD_PATH.unshift File.join(__dir__, 'lib')
 
+# Disables internationalized strings, which shouldn't be needed for tests.
+# This gets around an issue where Puppet::Environment and GettextSetup in
+# r10k are fighting over the same text domain within the FastGettext domain.
+# https://github.com/voxpupuli/ra10ke/issues/39
+Puppet[:disable_i18n] = true
```
lucywyman, line 23: This makes me nervous, mostly because I don't know very much about it. Will users run into the gettext error in the wild? I can't reproduce it locally, but it's hard to be 100% sure. Is disabling gettext in Pal an option?

beechtom: Considering the error is caused by a conflict between r10k and `Puppet::Environment`, which is instantiated with `Puppet::Pal.in_tmp_environment`. I doubt this particular error will pop up for users since r10k is never required at the same time a `Puppet::Environment` is instantiated.

PR #1849, opened 2020-05-27
(GH-1829) Use system project directory if homedir isn't defined

lucywyman:
Bolt is sometimes run in processes where no homedir is defined, causing
the expansion of `~` to fail. This was previously taken care of in most
places in #1676 but missed the default project directory at
`~/.puppetlabs/bolt`. Now when we try to load the path if it fails it
will fall back to the system-wide configuration path at
`/etc/puppetlabs/bolt` as the default project directory.

!bug

* **Fall back to system-wide config path if homedir expansion fails**

  We now fall back to `/etc/puppetlabs/bolt` as the default project
  directory if expanding the homedir fails.

### DIFF
--- a/lib/bolt/config.rb
+++ b/lib/bolt/config.rb
@@ -115,7 +115,7 @@ module Bolt
         data: Bolt::Util.read_optional_yaml_hash(project.config_file, 'config')
       }
 
-      data = load_defaults.push(data).select { |config| config[:data]&.any? }
+      data = load_defaults(project).push(data).select { |config| config[:data]&.any? }
 
       new(project, data, overrides)
     end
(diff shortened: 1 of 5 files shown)


nicklewis:
Will this cause us to load `/etc/puppetlabs/bolt/bolt.yaml` twice and merge them together? Once 
as the project config and once as the system config. That should be fine, but it's definitely a bit weird.

lucywyman:
Oh, hm, that's true. I can ignore the system_path when it's the same as project.path.

lib/bolt/config.rb
```diff
@@ -473,18 +474,12 @@ def matching_paths(paths)
...
-      @default_concurrency ||= if !sc_open_max_available? || Etc.sysconf(Etc::SC_OPEN_MAX) >= 300
-                                 DEFAULT_DEFAULT_CONCURRENCY
-                               else
-                                 Etc.sysconf(Etc::SC_OPEN_MAX) / 3
-                               end
+      if Bolt::Util.windows? || Etc.sysconf(Etc::SC_OPEN_MAX) >= 300 || Etc.sysconf(Etc::SC_OPEN_MAX).nil?
+        DEFAULT_DEFAULT_CONCURRENCY
+      else
+        (Etc.sysconf(Etc::SC_OPEN_MAX) / 3).floor
+      end
```
beechtom, line 482: This needs a rebase so it doesn't overwrite these changes.

PR #1851, opened 2020-05-27
(GH-1843) Load analytics config from user-level config directory

### LINKED ISSUE #1843: load analytics config from a cleaner path.
## Use Case
currenlty analytics configuration lives in the `~/.puppetlabs/bolt` the default project. Now that we have user level global config at `~/.puppetlabs/etc/bolt/analytics.yaml` this is a better location for analytics config.

Bolt should create analytics files in the `etc` path rather then the default project
Bolt should fall back to checking the default project location.
Bolt should use the `etc` version of the file if both are present and issue a warning.
We should not yet issue a deprecation warning about the original path but we should update docs to point to the new path only.

# Questions
Do other tools rely on the existing location for the analytics.

beechtom:
This updates the analytics client to load its configuration from
`~/.puppetlabs/etc/bolt/analytics.yaml` by default instead of
`~/.puppetlabs/bolt/analytics.yaml`.

By default, Bolt will first look for the config file at the new
location and load it if it exists. If it does not exist, Bolt will fall
back to the old location and load it if it exists. If it does not exist
at either location, Bolt will write the config file to the new location.

If a user has a config file at both the old and new locations, it will
display a helpful warning that the old location is deprecated.

Closes #1843 

!feature

* **Analytics configuration loaded from user-level config directory** (#1843)

  Analytics configuration is now written to and loaded from
  `~/.puppetlabs/etc/bolt/analytics.yaml` by default. Bolt will fall
  back to loading analytics config from
  `~/.puppetlabs/bolt/analytics.yaml` when the file does not exist in
  the user-level config directory.

### DIFF
--- a/documentation/bolt_installing.md
+++ b/documentation/bolt_installing.md
@@ -366,7 +366,8 @@ Bolt collects data to help us understand how it's being used and make decisions 
 
 ### How can I opt out of Bolt data collection?
 
-To disable the collection of analytics data add the following line to `~/.puppetlabs/bolt/analytics.yaml`:
+To disable the collection of analytics data add the following line to `~/.puppetlabs/etc/bolt/analytics.yaml`:
+
 ```yaml
 disabled: true
 ```
--- a/lib/bolt/analytics.rb
+++ b/lib/bolt/analytics.rb
@@ -29,7 +29,7 @@ module Bolt
     def self.build_client
       logger = Logging.logger[self]
       begin
-        config_file = File.expand_path('~/.puppetlabs/bolt/analytics.yaml')
+        config_file = config_path(logger)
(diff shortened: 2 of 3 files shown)


nicklewis:
We create the analytics.yaml file in the first place. Should we therefore try to migrate to the new file path if it doesn't already exist?

beechtom:
I was thinking about that earlier and think it would be a slightly more elegant solution. We then wouldn't need to go through the whole dance of warning about the deprecated file before removing support for it entirely..

beechtom:
@nicklewis Updated it to move an existing analytics file to the new location and then always load from the new location. If this is the direction we want to go I'll squash the commits.

PR #1856, opened 2020-05-28
(GH-1842) Load config from bolt-project.yaml

### LINKED ISSUE #1842: Support loading config information from `bolt-project.yaml`
## Use Case
It is confusing to have two project level configuration files. Bolt should support loading most config from `bolt-project.yaml` in addition to `bolt.yaml`

- Load all bolt.yaml config from bolt-project.yaml except for transport configuration.
- when both `bolt.yaml` and `bolt-project.yaml` are present warn the user and ignore `bolt.yaml`
- document configuration under the experimental bolt project docs.

lucywyman:
This makes several changes to how Bolt loads config. We now load bolt
configuration options (eg. concurrency), excluding transport config
options, from `bolt-project.yaml` if that file is present in the project
directory and has non-project config keys (e.g. 'format'). If both
`bolt-project.yaml` and `bolt.yaml` are present in the root of the
project directory and `bolt-project.yaml` has config keys `bolt.yaml`
will be ignored, regardless of how bolt determines the project
directory. This means that users can specify `--configfile
path/to/bolt.yaml`, and if there's a `bolt-project.yaml` with config
next to it it will not be loaded. If only `bolt.yaml` is present, it
will be loaded normally including transport config. Users using
`bolt-project.yaml` should specify transport configuration in
`inventory.yaml`.

Closes #1842

!feature

* **Load config from bolt-project.yaml** ([#1842](#1842))

  Bolt configuration options, excluding transport config, can now be
  loaded from `bolt-project.yaml`. If both `bolt-project.yaml` and
  `bolt.yaml` are present in the project and `bolt-project.yaml` has
  bolt config keys (e.g. 'format'), `bolt.yaml` will be ignored.

### DIFF
--- a/documentation/bolt_configuration_reference.md.erb
+++ b/documentation/bolt_configuration_reference.md.erb
@@ -116,7 +116,14 @@ The `puppetfile` section of the configuration file configures how to retrieve mo
 
 ## Transport configuration options
 
-Transport configuration options can be set in both the configuration file and inventory file.
+Transport configuration options can be set in the inventory file. Options at
+the inventory's top level will apply to all targets.
+
+### The transport option
+
+<% @transport_option[:options].each do |option, desc| -%>
+| `<%= option %>` | <%= desc %> | <%= @transport_option[:defaults][option].to_s %> |
+<% end %>
 
 <% @transports[:options].each do |transport, options| %>
 ### <%= transport %>
--- a/documentation/experimental_features.md
+++ b/documentation/experimental_features.md
@@ -37,14 +37,10 @@ Before your begin, make sure you've [updated Bolt to version 2.8.0 or
 higher](./bolt_installing.md).
 
 To get started with a Bolt project:
-1. Create a Bolt project:
-   - To create a fresh project directory, use the command `bolt project init
-   <PROJECT_NAME>`
-   - To turn an existing directory into a Bolt project directory, run `Bolt
-     project init` from the root of the directory.   
-2. Create a `bolt-project.yaml` file in the root of your Bolt project directory.
-3. Develop your Bolt plans and tasks in `plans` and `tasks` directories
-   in the root of the project directory.
+1. Create a `bolt-project.yaml` file in the root of your Bolt project directory. This can be an
+   existing directory, or a new one you make.
+1. Develop your Bolt plans and tasks in `plans` and `tasks` directories
+   in the root of the project directory, next to `bolt-project.yaml`.
 
 If `bolt-project.yaml` exists at the root of a project directory, Bolt loads the
 project as a module. Bolt loads tasks and plans from the `tasks` and `plans`
@@ -53,7 +49,6 @@ directories and namespaces them to the project name.
 Here is an example of a project using a simplified directory structure:
 ```console
 .
-├── bolt.yaml
 ├── bolt-project.yaml
 ├── inventory.yaml
 ├── plans
@@ -62,7 +57,29 @@ Here is an example of a project using a simplified directory structure:
     └── mytask.yaml
 ```
 
-### Naming your project
+### Project configuration
+
+As of Bolt 2.13.0 `bolt-project.yaml` contains both project configuration and [Bolt
+configuration](bolt_configuration_reference.md), excluding [transport configuration
+options](bolt_configuration_reference.md#transport-configuration-options). If your project contains
+both `bolt.yaml` and `bolt-project.yaml` files and `bolt-project.yaml` contains valid bolt config
+data, `bolt.yaml` will be ignored and `bolt-project.yaml` will be preferred for loading Bolt
+configuration. If using `bolt-project.yaml`, set transport configuration in your [inventory.yaml
+file](inventory_file_v2.md).
+
+For example, this `bolt-project.yaml` configures logging and modulepath for Bolt:
+```yaml
+# bolt-project.yaml
+modulepath: ['modules','site-modules','/home/user/mymodules']
+
+log:
+  console:
+    level: notice
+  ~/.puppetlabs/bolt/debug.log
+    level: debug
+```
+
+#### Naming your project
 
 If you want to set a name for your project that is different from the name of
 the Bolt project directory, add a `name` key to `bolt-project.yaml` with the project
@@ -79,7 +96,7 @@ lowercase letter.
 
 > **Note:** Projects take precedence over installed modules of the same name.
 
-### Whitelisting plans and tasks
+#### Whitelisting plans and tasks
 
 To control what tasks and plans appear when your users run `bolt plan
 show` or `bolt task show`, add `tasks` and `plans`
--- a/lib/bolt/config.rb
+++ b/lib/bolt/config.rb
@@ -34,6 +34,13 @@ module Bolt
       'remote' => Bolt::Config::Transport::Remote
     }.freeze
 
+    TRANSPORT_OPTION = { 'transport' => 'The default transport to use when the '\
+                         'transport for a target is not specified in the URL.' }.freeze
+
+    DEFAULT_TRANSPORT_OPTION = { 'transport' => 'ssh' }.freeze
+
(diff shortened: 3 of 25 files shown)


lib/bolt/project.rb
```diff
@@ -47,9 +55,15 @@ def initialize(path, type = 'option')
       @resource_types = @path + '.resource_types'
       @type = type
 
-      @project_file = @path + 'bolt-project.yaml'
-      @data = Bolt::Util.read_optional_yaml_hash(File.expand_path(@project_file), 'project') || {}
-      validate if load_as_module?
+      yaml = Bolt::Util.read_optional_yaml_hash(File.expand_path(@project_file), 'project') || {}
```
donoghuc, line 58: I thought we made sure `read_optional_yaml_hash` always returned a hash? Is this `nil` somehow? Maybe its a bug?

lucywyman: I thought the exact same thing! Yes, there's a test that fails saying that it's nil. I wonder if the test mocks Bolt::Util and doesn't mock this function maybe?

lucywyman: Yep - found it and fixed.

donoghuc: What was the problem? I was thinking you wouldnt need the `|| {}`
```suggestion
yaml = Bolt::Util.read_optional_yaml_hash(File.expand_path(@project_file), 'project')
```

lib/bolt/config.rb
```diff
@@ -191,6 +195,7 @@ def initialize(project, config_data, overrides = {})
       ld_concurrency = loaded_data.map(&:keys).flatten.include?('concurrency')
       if default_concurrency != DEFAULT_DEFAULT_CONCURRENCY &&
          !ld_concurrency &&
+         !overrides[:from_default] &&
```
beechtom, line 198: I don't think that this is actually needed? The default config will never have its warnings emitted since it's replaced by the real config before that happens.

If we need this for some other reason I think it might be better to have it as a separate flag that's not part of the overrides hash, since that's technically supposed to be options passed from the CLI.

lib/bolt/project.rb
```diff
@@ -49,7 +49,15 @@ def self.find_boltdir(dir)
 
     def initialize(path, type = 'option')
       @path = Pathname.new(path).expand_path
-      @config_file = @path + 'bolt.yaml'
+      @project_file = @path + 'bolt-project.yaml'
+
+      @warnings = []
+      if (@path + 'bolt.yaml').file? && project_file?
+        msg = "Found bolt.yaml and bolt-project.yaml in project directory '#{@path}', ignoring bolt.yaml"
```
beechtom, line 56: I didn't think of this during planning, but I'm a little concerned that anyone who has opted into an experimental feature could have their projects immediately break when this feature is released. Even though it's probably a relatively small subset of users who will have both files, I imagine it will cause some headaches to not have config loaded... Would it make sense to only ignore bolt.yaml if bolt-project.yaml has config in it?

beechtom:
https://github.com/puppetlabs/bolt/blob/master/schemas/bolt-project.schema.json should also be updated to include config options.

nicklewis:
@lucywyman I think it would be great to integrate the `Bolt::Project` constructor refactor discussed on #1855 into this PR. This change makes that code more complex, so this seems like the right time to separate the file-handling logic from the object initialization.

documentation/experimental_features.md
```diff
@@ -37,14 +37,19 @@ Before your begin, make sure you've [updated Bolt to version 2.8.0 or
...
+   in the root of the project directory, next to `bolt-project.yaml`.
+
+As of [Bolt 2.12.0](https://github.com/puppetlabs/bolt/blob/master/CHANGELOG.md#bolt-2120-2020-06-01)
+`bolt-project.yaml` contains both project configuration and [Bolt
+configuration](bolt_configuration_reference.md), excluding [transport
+configuration options](bolt_configuration_reference.md#transport-configuration-options).
+If your project contains both `bolt.yaml` and `bolt-project.yaml` files, `bolt.yaml` will be ignored
+and `bolt-project.yaml` will be preferred for loading Bolt configuration. If using
+`bolt-project.yaml`, set transport configuration in your [inventory.yaml
+file](inventory_file_v2.md).
```
hestonhoffman, line 52: I think we should move this section down and put it under a subheading like "Setting Bolt config". Maybe under the "Whitelisting section" - depending on where the config settings would usually appear in a project file with a name and whitelist.

lib/bolt/project.rb
```diff
@@ -58,9 +74,26 @@ def initialize(path, type = 'option')
       @resource_types = @path + '.resource_types'
       @type = type
 
-      @project_file = @path + 'bolt-project.yaml'
-      @data = Bolt::Util.read_optional_yaml_hash(File.expand_path(@project_file), 'project') || {}
-      validate if load_as_module?
+      tc = Bolt::Config::TRANSPORT_CONFIG.keys & raw_data.keys
```
beechtom, line 77: I can't remember, but did we ever discuss the `transport` option? It seems a little weird that it's the only option that can be set in both `bolt-project.yaml` and `inventory.yaml`.

beechtom: From standup it looks like this was an oversight when we discussed this ticket. `transport` should only be supported in `inventory.yaml`.

lucywyman: Hm, that's true - I think it makes sense to only have that be settable in inventory.

documentation/bolt_configuration_reference.md.erb
```diff
@@ -116,7 +116,14 @@ The `puppetfile` section of the configuration file configures how to retrieve mo
 
 ## Transport configuration options
 
-Transport configuration options can be set in both the configuration file and inventory file.
+Transport configuration options can be set in the inventory file. Options at
+the inventory's top level will apply to all targets.
+
+### The transport option
```
beechtom, line 122: Definitely work for another ticket, but I think we need to look at creating separate ref pages for each config/inventory file. This is quickly getting confusing. 😬

lib/bolt/project.rb
```diff
@@ -78,7 +111,11 @@ def eql?(other)
     end
     alias == eql?
 
-    def load_as_module?
+    def config_from_project?
```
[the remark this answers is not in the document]
lucywyman: I originally called this multiple places, now that there's just one call we can inline it for now and make it a method if we need to later on

lib/bolt/config.rb
```diff
@@ -106,27 +113,31 @@ class Config
...
     end
 
     def self.from_file(configfile, overrides = {})
-      project = Bolt::Project.new(Pathname.new(configfile).expand_path.dirname)
-
-      data = {
-        filepath: project.config_file,
-        data: Bolt::Util.read_yaml_hash(configfile, 'config')
-      }
+      project = Bolt::Project.create_project(Pathname.new(configfile).expand_path.dirname)
```
nicklewis, line 134: Hmm this is a little strange. If I run with `--configfile bolt.yaml` but there's a `bolt-project.yaml` next to it, it will use that instead?

nicklewis:
One question about `--configfile` but otherwise this is looking good

lib/bolt/config.rb
lucywyman: Yes, I said as much in the commit message:
> This means that users can specify --configfile path/to/bolt.yaml, and if there's a bolt-project.yaml with config
next to it it will not be loaded.

lucywyman: I think the case where you want to do that is rare, and I think users who are using `bolt-project.yaml` and have config in it know what they're doing enough to work within that limitation. This is also pretty temporary (we'll hopefully have one config file soon), so I'm not Too Worried:tm:

nicklewis: :+1: That's what I get for skimming the commit message. I think we can deprecate `--configfile` altogether before long.

PR #1859, opened 2020-05-28
(GH-1847) Use task parameter default when parameter is `Undef`

### LINKED ISSUE #1847: Passing `undef` as the value of a task parameter should use the default value
## Use Case
I want to have an optional parameter passed directly through from a plan to a sub-task. If the value is not set, I want the default to be used in the task. 

For a plan with a subtask like:
```puppet
plan test::default(TargetSpec $targets, Optional[String] $foo = undef){
  return run_task(test, $targets, 'foo' => $foo)
}
```
and task metadata like:
```json
{
  "parameters": {
    "foo": {
      "type": "Optional[String]",
      "default": "bar",
      "description": "my desc"
    }
  }
}
```
When i do not supply a `foo` argument to the plan, I want the task to use the default `bar` (currently it is `null`/`undef`. 

The `Optional` in the task metadata might be a bit confounding. Currently if i try with:
```json
{
  "parameters": {
    "foo": {
      "type": "Optional[String]",
      "default": "bar",
      "description": "my desc"
    }
  }
}
```
I get a PAL error:
```json
{
  "kind": "bolt/pal-error",
  "msg": "Task test:\n parameter 'foo' expects a String value, got Undef (file: /opt/puppetlabs/puppet/modules/test/plans/default.pp, line: 2)",
  "details": {
  }
}
```

When a task specifies a default value we should replace `undef` with the specified default if is is explicitly passed. We do not expect tasks to treat `nil` different from an unspecified parameter so this is in line other task behavior.

beechtom:
This updates the `run_task` function to use a task parameter default
when a parameter is specified but has a value of `Undef`. Previously, if
a parameter was specified with a value of `Undef`, the default would not
be used and the `Undef` value would be passed to the task.

Closes #1847 

!feature

* **Use task parameter default when specified parameter is `Undef`**
  ([#1847](https://github.com/puppetlabs/issues/1847))

  Task parameters that are specified with a value of `Undef` will now
  use the default parameter value if one is defined in the task's
  metadata.

### DIFF
--- a/bolt-modules/boltlib/lib/puppet/functions/run_task.rb
+++ b/bolt-modules/boltlib/lib/puppet/functions/run_task.rb
@@ -88,8 +88,10 @@ Puppet::Functions.create_function(:run_task) do
 
       task = Bolt::Task.from_task_signature(task_signature)
 
-      # Set the default value for any params that have one and were not provided
-      params = task.parameter_defaults.merge(params)
+      # Set the default value for any params that have one and were not provided or are undef
+      params = task.parameter_defaults.merge(params) do |_, default, passed|
+        passed.nil? ? default : passed
+      end
 
       task_signature.runnable_with?(params) do |mismatch_message|
         raise with_stack(:TYPE_MISMATCH, mismatch_message)
--- a/bolt-modules/boltlib/lib/puppet/functions/run_task_with.rb
+++ b/bolt-modules/boltlib/lib/puppet/functions/run_task_with.rb
@@ -124,7 +124,10 @@ Puppet::Functions.create_function(:run_task_with) do
       # If parameters are mismatched, create a failing result for the target that will later
       # be added to the ResultSet.
       unless pcp_only
-        params = task.parameter_defaults.merge(params)
+        # Set the default value for any params that have one and were not provided or are undef
+        params = task.parameter_defaults.merge(params) do |_, default, passed|
+          passed.nil? ? default : passed
+        end
 
         type_match = task_signature.runnable_with?(params) do |mismatch_message|
           exception = with_stack(:TYPE_MISMATCH, mismatch_message)
@@ -157,7 +160,10 @@ Puppet::Functions.create_function(:run_task_with) do
         end
       end
 
-      mapping[target] = task.parameter_defaults.merge(params)
+      # Set the default value for any params that have one and were not provided or are undef
+      mapping[target] = task.parameter_defaults.merge(params) do |_, default, passed|
+        passed.nil? ? default : passed
+      end
     end
 
     # Add a noop parameter if the function was called with the noop metaparameter.
--- a/bolt-modules/boltlib/spec/fixtures/modules/test/tasks/undef.sh
+++ b/bolt-modules/boltlib/spec/fixtures/modules/test/tasks/undef.sh
@@ -1,0 +1,1 @@
+echo -n "$PT_undef_default"
--- a/documentation/writing_tasks.md
+++ b/documentation/writing_tasks.md
@@ -228,7 +228,9 @@ To define a parameter as sensitive within the JSON metadata, add the `"sensitive
 
 ### Setting default values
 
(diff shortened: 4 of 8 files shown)


bolt-modules/boltlib/lib/puppet/functions/run_task.rb
```diff
@@ -89,7 +89,8 @@ def run_task_with_description(task_name, targets, description, args = {})
       task = Bolt::Task.from_task_signature(task_signature)
 
       # Set the default value for any params that have one and were not provided
-      params = task.parameter_defaults.merge(params)
+      # If a parameter was given an undef value and has a default specified, use the default
+      params = task.parameter_defaults.merge(params) { |_, default, passed| passed || default }
```
nicklewis, line 93: This will cause a value of `false` to also be overridden by the default. I think we should just strip out any params with value `nil` at the start.

bolt-modules/boltlib/lib/puppet/functions/run_task_with.rb
```diff
@@ -124,7 +124,9 @@ def run_task_with_with_description(task_name, targets, description, options = {}
       # If parameters are mismatched, create a failing result for the target that will later
       # be added to the ResultSet.
       unless pcp_only
-        params = task.parameter_defaults.merge(params)
+        # Set the default value for any params that have one and were not provided
+        # If a parameter was given an undef value and has a default specified, use the default
+        params = task.parameter_defaults.merge(params) { |_, default, passed| passed || default }
```
nicklewis, line 129: I think if we move this line outside the `pcp_only` then we don't have to do the merge twice? That will result in our wrapping default values for sensitive parameters in `Sensitive` which we don't currently do. That seems more correct to me anyway though.

Are we modifying the hash yielded from the block now? That shouldn't commonly be a problem because you probably are constructing the hash specifically to return it to the block, but if you've built up a map of target -> params instead, then that variable will be modified after this function runs.

nicklewis: It appears we don't modify the hash in any case. The only time we _could_ modify it would be with `pcp_only == true` (since we don't make a copy by merging it with the defaults). But if `pcp_only == true`, we make a stub task and don't have any parameters so we will never wrap anything in Sensitive in that case (therefore making no modifications).

That also means my suggestion to move this line outside the conditional won't work, since `task.parameter_defaults` will be `nil` in the `pcp_only` case.

I do think the other suggestion of filtering `params` to only those with non-nil values early would reduce some duplication though.

bolt-modules/boltlib/spec/fixtures/modules/test/tasks/undef.sh
```diff
@@ -0,0 +1 @@
+echo -n "$PT_undef"
```
lucywyman, line 1: Since this doesn't actually work (the parameter name is wrong), I think we can either use an existing task instead, or should fix this and verify that the output is correct.

documentation/writing_tasks.md
```diff
@@ -228,7 +228,9 @@ To define a parameter as sensitive within the JSON metadata, add the `"sensitive
 
 ### Setting default values
 
-You can set a default value for a parameter which will be used if the parameter isn't specified. The default value must be valid according to the parameter's `type`.
+You can set a default value for a parameter which will be used if the parameter isn't specified or if the
+parameter is specified and has a value of `Undef`. The default value must be valid according to the parameter's
```
lucywyman, line 232:
```suggestion
parameter is specified and has a value of `Undef`. The default will be used even if the parameter type is optional, and the default value must be valid according to the parameter's
```

nicklewis:
This does have the somewhat strange behavior that you can now specify `undef` for an unexpected parameter and the task won't fail...

beechtom:
Reverted back to replacing `nil` values during the merge. It seems like the most straightforward way to accomplish what we want.

PR #1863, opened 2020-05-29
Allow boltdir to be found no matter the case

nmaludy:
In the current implementation Boltdir needs to match the case exactly on filesystems that care about case. Sticking with the Linux tradition i like to have my directories all lowercase and though it would be nice if Bolt was able to find `boltdir` or `Boltdir` whatever the user preferred.

This change allows `Boltdir` to be of any case and we will use the first `Boltdir` we find. 

I'm also sorting the `children` returned in the directory so the lexicographical "less" / "smallest" `Boltdir` variant will be chosen and this will be consistent across `bolt` runs. This is better than relying on whatever random ordering the filesystem will return the directory listings in.

### DIFF
--- a/lib/bolt/project.rb
+++ b/lib/bolt/project.rb
@@ -25,8 +25,10 @@ module Bolt
     # hierarchy, falling back to the default if we reach the root.
     def self.find_boltdir(dir)
       dir = Pathname.new(dir)
-      if (dir + BOLTDIR_NAME).directory?
-        new(dir + BOLTDIR_NAME, 'embedded')
+      # allows for any case of Boltdir, BoltDir, boltdir, etc
+      boltdirs = dir.children.select { |c| c.directory? && c.basename.to_s.casecmp?(BOLTDIR_NAME)  }
+      if !boltdirs.empty?
+        new(boltdirs.first, 'embedded')
       elsif (dir + 'bolt.yaml').file? || (dir + 'bolt-project.yaml').file?
         new(dir, 'local')
       elsif dir.root?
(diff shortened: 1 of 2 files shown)


lib/bolt/project.rb
```diff
@@ -25,8 +25,14 @@ def self.default_project
...
       dir = Pathname.new(dir)
-      if (dir + BOLTDIR_NAME).directory?
-        new(dir + BOLTDIR_NAME, 'embedded')
+      # allows for any case of Boltdir, BoltDir, boltdir, etc
+      # note: we're sorting the directories here so that we consistently pick the
+      #       same Boltdir between runs, and we always choose the lexicographical "less"
+      #       Boltdir name rather than whatever the first name we run into as returned by
+      #       the filesystem
+      boltdirs = dir.children.sort.select { |c| c.directory? && c.basename.to_s.casecmp?(BOLTDIR_NAME) }
+      if !boltdirs.empty?
```
adreyer, line 34: We should issue a warning and say which one we're using if more than one is found.

lib/bolt/project.rb
```diff
@@ -73,7 +79,7 @@ def load_as_module?
 
     def name
       # If the project is in mymod/Boltdir/bolt-project.yaml, use mymod as the project name
-      dirname = @path.basename.to_s == 'Boltdir' ? @path.parent.basename.to_s : @path.basename.to_s
+      dirname = @path.basename.to_s.casecmp?(BOLTDIR_NAME) ? @path.parent.basename.to_s : @path.basename.to_s
```
nicklewis, line 82: I think this check can just be `@type == 'embedded'`, which also better expresses the intent.

adreyer:
After thinking this over for a while I think supporting case insensitivity for boltdir is probably not the best path forward. I do think it's reasonable to warn if a `boltdir` is found however.

beechtom:
Closing since we do not plan on supporting case insensitivity for files/directories.

PR #1875, opened 2020-06-04
(GH-1874) Add resource plan function to lookup resource on a target

### LINKED ISSUE #1874: Add target.resource function to lookup resources on a target
## Use Case
I want to lookup a resource so that I can modify it

A new function on the target object 'resource' that takes a title and type as positional arguments and returns either the ResourceInstance object if found, or nil.

lucywyman:
This adds a `resource()` plan function that takes a target, type, and
title and returns either the resource object or nil. The function is a
plan function, so either `$target.resource('type', 'title')` or
`resource($target, 'type', 'title')` can be used. It creates a temporary
ResourceInstance object from the provided type and title and calls
reference directly on that, so that the user doesn't need to know
exactly how type and title are munged into the reference and if the
reference changes this shouldn't break.

Closes #1874

!feature

* **Add resource plan function to lookup a resource on a target** ([#1874](https://github.com/puppetlabs/bolt/issues/1874))

  This adds a `resource` plan function that can be used to lookup a
  resource on a target by 'type' and 'title'.


lucywyman:
Sorry, I know we have a lot of other stuff on the board, but I already had most of this done and just needed to add the test :P

bolt-modules/boltlib/lib/puppet/functions/resource.rb
```diff
@@ -0,0 +1,46 @@
...
+  # @param title The title of the resource
+  # @return The ResourceInstance if found, or Undef
+  # @example Get the openssl package resource
+  #   $target.apply_prep
+  #   $resources = $target.get_resources(Package).first['resources']
+  #   $target.set_resources($resources)
+  #   $openssl = $target.resource('Package', 'openssl')
+  dispatch :resource do
+    param 'Boltlib::TargetSpec', :target
+    param 'Variant[String[1], Type]', :type
```
beechtom, line 26:
```suggestion
    param 'Variant[String[1], Type[Resource]]', :type
```

bolt-modules/boltlib/lib/puppet/functions/resource.rb
```diff
@@ -0,0 +1,46 @@
...
+    param 'String[1]', :title
+    return_type 'Optional[ResourceInstance]'
+  end
+
+  def resource(target, type, title)
+    unless Puppet[:tasks]
```
beechtom, line 32:
```suggestion
    unless Puppet[:tasks]
```

It seems a little odd that you can use `$target.resources` in an apply, but not `$target.resource`.

bolt-modules/boltlib/lib/puppet/functions/resource.rb
```diff
@@ -0,0 +1,46 @@
...
+  # @param type The type of the resource
+  # @param title The title of the resource
+  # @return The ResourceInstance if found, or Undef
+  # @example Get the openssl package resource
+  #   $target.apply_prep
+  #   $resources = $target.get_resources(Package).first['resources']
+  #   $target.set_resources($resources)
+  #   $openssl = $target.resource('Package', 'openssl')
+  dispatch :resource do
+    param 'Boltlib::TargetSpec', :target
```
beechtom, line 25:
```suggestion
    param 'Target', :target
```

This can probably just be a Target object since one already has to exist to even retrieve resources. This would also be in line with the `set_resources` function, which expects a Target object.

lib/bolt/target.rb
```diff
@@ -80,6 +80,11 @@ def resources
       inventory_target.resources
     end
 
+    def resource(type, title)
+      rr = ResourceInstance.new({ 'type' => type, 'title' => title }).reference
```
beechtom, line 84: I feel like creating a new ResourceInstance here isn't the right approach. Maybe a class method that formats the type and title for you, which can be called from `ResourceInstance.reference` and `Target.resource`? Something like

```
def self.format_reference(type, title)
    "#{type}[#{title}]"
end

def reference
    self.class.format_reference(@type, @title)
end
```

bolt-modules/boltlib/lib/puppet/functions/resource.rb
lucywyman: This just limits how users can call the function. We call `get_target` on the parameter, so with `Boltlib::TargetSpec` users can do `resource('hostname', 'type', 'title')` where with just `Target` that will fail. I don't see a good reason to limit it.

bolt-modules/boltlib/lib/puppet/functions/resource.rb
lucywyman: In order to do that we'd need to make this a function on the target datatype itself. We can do that, but then we won't get analytics for it and `resource('target', 'type', 'title')` won't work. I'm not opposed to having the function be on the target - it's a small thing, but I didn't want to have to add a new "parameters" column to the datatype reference table just for this one function, and it seemed to make more sense as a plan function anyway.

bolt-modules/boltlib/lib/puppet/functions/resource.rb
lucywyman: Are most custom types a 'Resource' type? I don't think we enforce this when setting resources.

beechtom: We do enforce it for the type's `type` attribute: https://github.com/puppetlabs/bolt/blob/master/bolt-modules/boltlib/lib/puppet/datatypes/resourceinstance.rb#L7 . It looks like that doesn't really apply to `set_resources`, though, since it just expects a hash. Regardless, I believe custom types _should_ be a `Resource` type since they are created using the resource API.

bolt-modules/boltlib/lib/puppet/functions/resource.rb
nicklewis: For consistency with `facts()`, `vars()`, and `resources()`, I think it makes sense for this to be `Target`.

bolt-modules/boltlib/lib/puppet/functions/resource.rb
nicklewis: I'm not clear why this wouldn't work. What about an apply block requires it to be a function on the Target datatype?

bolt-modules/boltlib/lib/puppet/functions/resource.rb
nicklewis: We should support `Type[Resource]` but it would probably be better off as a second dispatch of this function rather than a `Variant`.

lib/bolt/target.rb
```diff
@@ -80,6 +80,11 @@ def resources
       inventory_target.resources
     end
 
+    def resource(type, title)
```
nicklewis, line 83: This logic shouldn't live on `Bolt::Target` as it's not part of the API of the `Target` datatype. It can either be implemented directly in the Puppet function (if all we're doing is calling `resources["<ref>"]`) or it should follow the pattern of similar functions, which is to call a method on the inventory that delegates to `Bolt::Inventory::Target`.

Alternately, we could just implement this as a function on the `Target` datatype.

bolt-modules/boltlib/lib/puppet/functions/resource.rb
lucywyman: I added that, though for now I don't know if it's safe to convert the string to Type in the `resource_from_string` function, since `set_resource` doesn't do that conversion as well.

bolt-modules/boltlib/lib/puppet/functions/resource.rb
lucywyman: I guess 'wouldn't work' is the wrong way of phrasing - I more thought it didn't make sense to have such a simple function if we could just do it on the datatype, but I guess we do that for `vars` etc. :P

bolt-modules/boltlib/lib/puppet/functions/resource.rb
```diff
@@ -0,0 +1,52 @@
...
+    inventory = Puppet.lookup(:bolt_inventory)
+    executor  = Puppet.lookup(:bolt_executor) { nil }
+    executor&.report_function_call(self.class.name)
+
+    inventory.resource(target, type, title)
+  end
+
+  def resource_from_string(target, type, title)
+    resource(target, type, title)
+  end
```
beechtom, line 51: It doesn't look like this is used anywhere.

documentation/experimental_features.md
```diff
@@ -216,6 +216,18 @@ ResourceInstance.new(
 )
 ```
 
+### Accessing `ResourceInstance` object on a Target
+
+You can get a specific `ResourceInstance` object stored on a Target in a plan using the `resource()`
+method:
```
beechtom, line 222:
```suggestion
function:
```

bolt-modules/boltlib/lib/puppet/functions/resource.rb
lucywyman: It supports passing the resource in as a string instead of a type

beechtom: Derp. Missed that it's a dispatch. 😅 Nevermind!

documentation/experimental_features.md
lucywyman: So pedantic. This is *technically* also a method on the target class.

documentation/experimental_features.md
```diff
@@ -216,6 +216,18 @@ ResourceInstance.new(
 )
 ```
 
+### Accessing `ResourceInstance` object on a Target
+
+You can get a specific `ResourceInstance` object stored on a Target in a plan using the `resource()`
```
hestonhoffman, line 221:
```suggestion
You can retrieve a specific `ResourceInstance` object stored on a Target in a plan using the `resource()`
```
Get is a little ambiguous here. Does "retrieve" make sense?

lucywyman: Sure, updated

lib/bolt/target.rb
```diff
@@ -86,6 +86,10 @@ def set_resource(resource)
     end
     # rubocop:enable Naming/AccessorMethodName
 
+    def resource(type, title)
```
nicklewis, line 89: I don't _think_ this version is needed, since we don't expose `resource()` as a function on the data type.

lib/bolt/resource_instance.rb
```diff
@@ -84,9 +84,13 @@ def to_json(opts = nil)
       to_hash.to_json(opts)
     end
 
-    def reference
+    def self.format_reference(type, title)
       "#{type}[#{title}]"
```
nicklewis, line 88: This should be `type.downcase.capitalize` to standardize the casing.

lib/bolt/target.rb
lucywyman: Having it here enables `$target.resource(type, title)` - we do this for set_resources, I think it makes sense to for `resource` as well

nicklewis: The difference is that the `set_resources` function _calls_ `Target#set_resource` directly. Nothing is calling this method. `$target.resource(type, title)` works because that's just how function invocation works in Puppet. It will call a function named `resource`, passing in `$target` as the first argument.

_If_ there were a `resource` function defined on the Target datatype, then it would call that 
instead, which would dispatch to this method. But since that's defined on the datatype, this method should never be invoked.

PR #1877, opened 2020-06-04
(GH-1873) Add `[]` function to ResourceInstance data type

### LINKED ISSUE #1873: Lookup resource state with bracket notation
The `[]` function on the resource_instance object should look up the key in the resource's state:
```
$resource['provider']
```
instead of 
```
$resource.state['provider']
```

It should follow the usual hash lookup conventions in Puppet.

beechtom:
This adds a `[]` function to the `ResourceInstance` data type that
directly accesses the `state` hash and returns the value for the
specified attribute. If the attribute is missing, the function will
return `nil`.

Closes #1873 

!feature

* **Added `[]` function to the `ResourceInstance` data type**
  ([#1873](https://github.com/puppetlabs/bolt/issues/1873))

  The `[]` function can be used to directly access the `state` hash for
  a `ResourceInstance` object and return the specified attribute.

### DIFF
--- a/bolt-modules/boltlib/lib/puppet/datatypes/resourceinstance.rb
+++ b/bolt-modules/boltlib/lib/puppet/datatypes/resourceinstance.rb
@@ -16,7 +16,8 @@ Puppet::DataTypes.create_type('ResourceInstance') do
       overwrite_state         => Callable[[Hash[String[1], Data]], Hash[String[1], Data]],
       set_desired_state       => Callable[[Hash[String[1], Data]], Hash[String[1], Data]],
       overwrite_desired_state => Callable[[Hash[String[1], Data]], Hash[String[1], Data]],
-      reference               => Callable[[], String]
+      reference               => Callable[[], String],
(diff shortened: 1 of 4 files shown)


bolt-modules/boltlib/lib/puppet/datatypes/resourceinstance.rb
```diff
@@ -16,7 +16,8 @@
       overwrite_state         => Callable[[Hash[String[1], Data]], Hash[String[1], Data]],
       set_desired_state       => Callable[[Hash[String[1], Data]], Hash[String[1], Data]],
       overwrite_desired_state => Callable[[Hash[String[1], Data]], Hash[String[1], Data]],
-      reference               => Callable[[], String]
+      reference               => Callable[[], String],
+      '[]'                    => Callable[[String[1]], Optional[Data]]
```
nicklewis, line 20: `Data` is an alias for a `Variant` that includes `Undef`, so this doesn't need to be `Optional`.

PR #1878, opened 2020-06-04
(GH-1872) Accept `resource_type` key in `set_resources` plan function

### LINKED ISSUE #1872: set_resource should accept either resource_type or type
## Use Case
I want to create a new resource on a target from an ApplyResult

Most of the information needed to make a new resource (target, type, title) is returned from an ApplyResult, however in the resource data in the result nests the type under the key 'resource_type' instead of 'type'. We should munge 'resource_type' to be the type if `type` isn't set in the hash. If both are set, prefer 'type' and don't warn.

We should also document creating a resource from an ApplyResult on the experimental features page:
```
plan myplan (
  TargetSpec $targets
) {
  $t = get_targets($targets)
  apply_prep($t)
  $ar = apply($t) {
    file { '/root/myfile':
      ensure => present
    }   
  }
  $ar.each |$result| {
    $result.report['resource_statuses'].values.each |$resource| {
      $target_obj = get_target($result.target)
      $target_obj.set_resources($resource)
    }
  }
  return $t.map |$target| { $target.resources }
}
```

beechtom:
This updates the `set_resources` plan function to accept a
`resource_type` key as part of the resource data hash, in addition to
`type`.

When creating a `ResourceInstance` object from a hash, a `type` key is
expected. However, when using resource data hashes returned from an
apply block, the resource's type is set under the `resource_type` key.
This made it impossible to directly use this hash to create a
`ResourceInstance` object, as the hash did not have the required `type`
key. The `set_resources` function now falls back to using the
`resource_type` key if the `type` key is not present.

It's important to note that calling `ResourceInstance.new` with a hash
that has a `resource_type` key instead of a `type` key will still raise
an exception.

To keep the behavior of `ResourceInstance.new` predictable, this also
removes the ability to use a `parameters` key instead of a `state` key
in the implementation class.

Lastly, this updates the `set_resources` function to have separate
dispatches instead of a `Variant` type for the `resources` parameter.

Closes #1872 

!feature

* **Accept `resource_type` key in resource data hash for `set_resources`
  plan function**
  ([#1872](#1872))

  The `set_resources` function now accepts resource data hashes that
  have a `resource_type` key instead of a `type` key. This allows users
  to set resources directly from reports from an apply block, which set
  a resource's type under the `resource_type` key.

### DIFF
--- a/bolt-modules/boltlib/lib/puppet/functions/set_resources.rb
+++ b/bolt-modules/boltlib/lib/puppet/functions/set_resources.rb
@@ -14,26 +14,61 @@ require 'bolt/error'
 #
 # > **Note:** Not available in apply block
 Puppet::Functions.create_function(:set_resources) do
-  # Set multiple resources
-  # @param target The `Target` object to add resources to. See {get_targets}.
-  # @param resources The resources to set on the target.
-  # @return The added `ResourceInstance` objects.
-  # @example Add multiple resources to a target with an array of `ResourceInstance` objects.
-  #   $resource1 = ResourceInstance.new(
+  # Set a single resource from a data hash.
+  # @param target The `Target` object to add a resource to. See {get_targets}.
+  # @param resource The resource data hash used to set a resource on the target.
+  # @return An array with the added `ResourceInstance` object.
+  # @example Add a resource to a target from a data hash.
+  #   $resource_hash = {
+  #     'type'  => File,
+  #     'title' => '/etc/puppetlabs',
+  #     'state' => { 'ensure' => 'present' }
+  #   }
+  #
+  #   $target.set_resources($resource_hash)
+  dispatch :set_single_resource_from_hash do
+    param 'Target', :target
+    param 'Hash', :resource
+    return_type 'Array[ResourceInstance]'
+  end
+
+  # Set a single resource from a `ResourceInstance` object
+  # @param target The `Target` object to add a resource to. See {get_targets}.
+  # @param resource The `ResourceInstance` object to set on the target.
+  # @return An array with the added `ResourceInstance` object.
+  # @example Add a resource to a target from a `ResourceInstance` object.
+  #   $resource_instance = ResourceInstance.new(
   #     'target' => $target,
-  #     'type'   => 'file',
+  #     'type'   => File,
   #     'title'  => '/etc/puppetlabs',
   #     'state'  => { 'ensure' => 'present' }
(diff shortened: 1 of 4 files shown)


nicklewis:
I like splitting this into separate dispatches, though I think the `Array` case needs to still be `Array[Variant[Hash, ResourceInstance]]` so we can accept a mixed array.

bolt-modules/boltlib/lib/puppet/functions/set_resources.rb
```diff
@@ -14,26 +14,61 @@
...
-  # Set multiple resources
-  # @param target The `Target` object to add resources to. See {get_targets}.
-  # @param resources The resources to set on the target.
-  # @return The added `ResourceInstance` objects.
-  # @example Add multiple resources to a target with an array of `ResourceInstance` objects.
-  #   $resource1 = ResourceInstance.new(
+  # Set a single resource from a data hash.
+  # @param target The `Target` object to add a resource to. See {get_targets}.
+  # @param resource The resource data hash used to set a resource on the target.
+  # @return The added `ResourceInstance` object.
```
lucywyman, line 20:
```suggestion
  # @return Array of the added `ResourceInstance` object.
```

bolt-modules/boltlib/lib/puppet/functions/set_resources.rb
```diff
@@ -14,26 +14,61 @@
...
+  dispatch :set_single_resource_from_hash do
+    param 'Target', :target
+    param 'Hash', :resource
+    return_type 'Array[ResourceInstance]'
+  end
+
+  # Set a single resource from a `ResourceInstance` object
+  # @param target The `Target` object to add a resource to. See {get_targets}.
+  # @param resource The `ResourceInstance` object to set on the target.
+  # @return The added `ResourceInstance` object.
```
lucywyman, line 38:
```suggestion
  # @return Array of the added `ResourceInstance` object.
```

bolt-modules/boltlib/lib/puppet/functions/set_resources.rb
lucywyman: This phrasing feels awkward, could definitely use improvement

lucywyman:
I think it'd also be useful to add an example of accessing a resource from the return of `set_resources` in the experimental features page, maybe to show the version for a package resource gotten using `get_resources` or something.

Issue #1862, opened 2020-05-29
Add option to enable deprecated algorithms in net-ssh


beechtom:
## Use Case
With the release of net-ssh 6.0, support for several weaker algorithms was deprecated and disabled by default. This can cause an issue for users who are attempting to connect to targets that are configured to only use deprecated algorithms, as it makes it impossible for Bolt to connect to them, even to update the SSH configuration to use stronger algorithms.

Users should be able to connect to targets using deprecated algorithms so they can continue to use Bolt, including using Bolt to update the target's SSH configuration.

## Describe the Solution You Would Like
Add a configuration option to the SSH transport that accepts a list of algorithms to enable in net-ssh. This option should also be configurable in PE.

## Questions
- Should Bolt display a warning when enabling a deprecated algorithm, or just have a warning in the documentation? net-ssh plans to remove support entirely for deprecated algorithms in 7.0, so it might be helpful to explicitly tell users this.

adreyer:
This is three settings which should override ssh config options.
ciphers
host-key algorithm
kex algorithm


Is there a case where net::ssh won't honor insecure but supported algorithms from ssh config? Discuss how to handle this if so.

adreyer:
add special string for `default-ciphers`.

beechtom:
net-ssh has a constant defining all supported algorithms: https://github.com/net-ssh/net-ssh/blob/92669095ab1689a140f7fcc46b16480f9b5c93ea/lib/net/ssh/transport/algorithms.rb#L66

Issue #1861, opened 2020-05-29
Provide a more detailed error message when net-ssh fails to settle on algorithms


beechtom:
## Use Case
With the release of net-ssh 6.0, support for several weaker algorithms was deprecated and disabled by default. When a user attempts to connect to a target over SSH and either the target or local host cannot settle on an algorithm, net-ssh will raise an error saying as much. However, this error can be unhelpful for several reasons:

- The algorithms may be supported in an older version of Bolt and suddenly stop working when upgrading to a newer version that uses net-ssh >= 6.0. We don't document algorithm support ourselves, so users may not know what they can use or what is deprecated.

- The net-ssh error lists available algorithms for both the server and the client, but filters out unsupported algorithms for the client. This can be frustrating to troubleshoot if you see a supported algorithm in the server's algorithms that you know is in the client's SSH configuration, but it doesn't show in the net-ssh error.

- The error doesn't explain that some algorithms are not supported and doesn't tell you where to get that info.

## Describe the Solution You Would Like
Bolt should rescue a `Net::SSH::Exception` that is raised from not settling on an algorithm and provide a more detailed message.

## Describe Alternatives You've Considered
Submit a PR to net-ssh that improves the error message.

## Additional Context
Several users have reported issues with algorithm support recently:

- https://github.com/puppetlabs/bolt/issues/733
- https://tickets.puppetlabs.com/browse/PE-29241

PR #1882, opened 2020-06-05
(GH-1862) Specify preferred  algorithms for SSH transport connections

beechtom:
This adds the following SSH transport configuration options:

- `encryption-algorithms`
- `host-key-algorithms`
- `kex-algorithms`
- `mac-algorithms`

Each of these options accepts an array of preferred algorithms to use when
establishing a connection using the SSH transport. When a `*-algorithms`
option is set, `net-ssh` is configured to only attempt to establish a
connection to the target using algorithms in the list.

Each `*-algorithms` option accepts a special `defaults` item, which will
automatically add all of default, supported algorithms for `net-ssh` to
the list. For example, the following config:

```yaml
ssh:
  encryption-algorithms:
    - defaults
```

Would expand to the following:

```yaml
ssh:
  encryption-algorithms:
    - aes-256-ctr
    - aes-192-ctr
    - aes-128-ctr
```

Starting with the release of `net-ssh` 6.0, several algorithms were
deprecated and disabled by default. This prevented users from using
these algorithms when establishing connections with targets. These
options allow users to enable these deprecated algorithms.

!feature

* **Specify preferred algorithms for SSH transport connections**
  ([#1862](https://github.com/puppetlabs/bolt/issues/1862))

  Users can now specify a list of preferred algorithms to use when
  establishing connections with targets using the SSH transport with the
  `encryption-algorithms`, `host-key-algorithms`, `kex-algorithms`, and
  `mac-algorithms` config options. Each option accepts an array of
  algorithms and overrides the default list of preferred algorithms.
  You can read more about these options in the [Bolt configuration
  reference](https://puppet.com/docs/bolt/latest/bolt_configuration_reference.html#ssh).

### DIFF
--- a/lib/bolt/config/transport/ssh.rb
+++ b/lib/bolt/config/transport/ssh.rb
@@ -12,94 +12,132 @@ module Bolt
         # NOTE: All transport configuration options should have a corresponding schema definition
         #       in schemas/bolt-transport-definitions.json
         OPTIONS = {
-          "cleanup"            => { type: TrueClass,
-                                    external: true,
-                                    desc: "Whether to clean up temporary files created on targets." },
-          "connect-timeout"    => { type: Integer,
-                                    desc: "How long to wait when establishing connections." },
-          "copy-command"       => { external: true,
-                                    desc: "Command to use when copying files using ssh-command. "\
-                                          "Bolt runs `<copy-command> <src> <dest>`. **This option is experimental.**" },
-          "disconnect-timeout" => { type: Integer,
-                                    desc: "How long to wait before force-closing a connection." },
-          "host"               => { type: String,
-                                    external: true,
-                                    desc: "Host name." },
-          "host-key-check"     => { type: TrueClass,
-                                    external: true,
-                                    desc: "Whether to perform host key validation when connecting." },
-          "extensions"         => { type: Array,
-                                    desc: "List of file extensions that are accepted for scripts or tasks on Windows. "\
-                                         "Scripts with these file extensions rely on the target's file type "\
-                                         "association to run. For example, if Python is installed on the system, "\
-                                         "a `.py` script runs with `python.exe`. The extensions `.ps1`, `.rb`, and "\
-                                         "`.pp` are always allowed and run via hard-coded executables." },
-          "interpreters"       => { type: Hash,
-                                    external: true,
-                                    desc: "A map of an extension name to the absolute path of an executable, "\
-                                          "enabling you to override the shebang defined in a task executable. The "\
-                                          "extension can optionally be specified with the `.` character (`.py` and "\
-                                          "`py` both map to a task executable `task.py`) and the extension is case "\
-                                          "sensitive. When a target's name is `localhost`, Ruby tasks run with the "\
-                                          "Bolt Ruby interpreter by default." },
-          "load-config"        => { type: TrueClass,
-                                    desc: "Whether to load system SSH configuration." },
-          "login-shell"        => { type: String,
-                                    desc: "Which login shell Bolt should expect on the target. "\
-                                          "Supported shells are #{LOGIN_SHELLS.join(', ')}. "\
-                                          "**This option is experimental.**" },
-          "password"           => { type: String,
-                                    desc: "Login password." },
-          "port"               => { type: Integer,
-                                    external: true,
-                                    desc: "Connection port." },
-          "private-key"        => { external: true,
-                                    desc: "Either the path to the private key file to use for authentication, or a "\
-                                          "hash with the key `key-data` and the contents of the private key." },
-          "proxyjump"          => { type: String,
-                                    desc: "A jump host to proxy connections through, and an optional user to "\
-                                          "connect with." },
-          "run-as"             => { type: String,
-                                    external: true,
-                                    desc: "A different user to run commands as after login." },
-          "run-as-command"     => { type: Array,
-                                    external: true,
-                                    desc: "The command to elevate permissions. Bolt appends the user and command "\
-                                          "strings to the configured `run-as-command` before running it on the "\
-                                          "target. This command must not require an interactive password prompt, "\
-                                          "and the `sudo-password` option is ignored when `run-as-command` is "\
-                                          "specified. The `run-as-command` must be specified as an array." },
-          "script-dir"         => { type: String,
-                                    external: true,
-                                    desc: "The subdirectory of the tmpdir to use in place of a randomized "\
-                                          "subdirectory for uploading and executing temporary files on the "\
-                                          "target. It's expected that this directory already exists as a subdir "\
-                                          "of tmpdir, which is either configured or defaults to `/tmp`." },
-          "ssh-command"        => { external: true,
-                                    desc: "Command and flags to use when SSHing. This enables the external "\
-                                          "SSH transport which shells out to the specified command. "\
-                                          "**This option is experimental.**" },
-          "sudo-executable"    => { type: String,
-                                    external: true,
-                                    desc: "The executable to use when escalating to the configured `run-as` "\
-                                          "user. This is useful when you want to escalate using the configured "\
-                                          "`sudo-password`, since `run-as-command` does not use `sudo-password` "\
-                                          "or support prompting. The command executed on the target is "\
-                                          "`<sudo-executable> -S -u <user> -p custom_bolt_prompt <command>`. "\
-                                          "**This option is experimental.**" },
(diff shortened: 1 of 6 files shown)


lucywyman:
I think these commits can be squashed.

lib/bolt/config/transport/ssh.rb
```diff
@@ -12,94 +12,130 @@ class SSH < Base
...
+                                             "All supported, non-deprecated algorithms are enabled by default when "\
+                                             "this option is not used. To enable support for default algorithms "\
+                                             "when using this option, add 'defaults' to the list of supported "\
+                                             "algorithms." },
+          "host"                  => { type: String,
+                                       external: true,
+                                       desc: "Host name." },
+          "host-key-algorithms"   => { type: Array,
+                                       desc: "List of host key algorithms to support when establishing a connection "\
+                                             "with a target. Supported algorithms are defined by net-ssh "\
```
lucywyman, line 40:
```suggestion
                                             "with a target. Supported algorithms are defined by the ruby net-ssh library "\
```

lib/bolt/transport/ssh/connection.rb
```diff
@@ -246,6 +268,19 @@ def net_ssh_verifier(verifier)
...
+        # out any unsupported algorithms.
+        def net_ssh_algorithms(type, algorithms)
+          if algorithms.include?('defaults')
+            defaults = Net::SSH::Transport::Algorithms::DEFAULT_ALGORITHMS[type]
+            algorithms += defaults
+          end
+
+          known = Net::SSH::Transport::Algorithms::ALGORITHMS[type]
+
+          algorithms & known
```
lucywyman, line 281: Until #1861 is in and released, it'd be nice if we warned here that certain specified algorithms aren't supported at all.
```
supported, unsupported = algorithms.partition { |a| known.include?(a) }
@logger.warn("These #{type.to_s} algorithms aren't supported by the net-ssh library: #{unsupported}")
supported
```

beechtom: My only issue with warning about unsupported algorithms is that it will warn for every single target. If we want #1861 in as an alternate to warning, then I think we can just block this ticket until that's ready.

beechtom: Added the work for #1861 to this PR. Felt the error message should also list disabled algorithms and how to enable them, so it's closely tied to the work for #1862 .

lucywyman:
It looks like we can add `Macs` to [this list](https://puppet.com/docs/bolt/latest/bolt_configuration_reference.html#openssh), but not KexAlgorithms

lib/bolt/transport/ssh/connection.rb
```diff
@@ -275,6 +312,32 @@ def validate_ssh_version
             raise "Powershell over SSH requires OpenSSH server >= 7.9, target is running #{match[1]}"
           end
         end
+
+        def handle_net_ssh_error(err)
+          match = /could not settle on (?<algo>encryption|hmac|host_key|kex)/.match(err.message)
+
+          if match
+            enabled = Net::SSH::Transport::Algorithms::DEFAULT_ALGORITHMS[match[:algo].to_sym]
```
lucywyman, line 320: I think 'available' and 'deprecated' would be better words for these. They aren't what's configured by the user, and I think 'enabled' indicates something on the user-side.

lucywyman:
I think all three commits can be squashed, this is all part of the same work.

schemas/bolt-transport-definitions.json
```diff
@@ -49,29 +49,33 @@
...
+      "encryption-algorithms": { "$ref": "#/definitions/encryption-algorithms" },
+      "host":                  { "$ref": "#/definitions/host" },
+      "host-key-algorithms":   { "$ref": "#/definitions/host-key-algorithms" },
+      "host-key-check":        { "$ref": "#/definitions/host-key-check" },
+      "extensions":            { "$ref": "#/definitions/extensions" },
+      "interpreters":          { "$ref": "#/definitions/interpreters" },
+      "kex-algorithms":        { "$ref": "#/definitions/kex-algorithms" },
+      "load-config":           { "$ref": "#/definitions/load-config" },
+      "login-shell":           { "$ref": "#/definitions/login-shell" },
+      "mac-algorithms":        { "$ref": "#/definitions/mac-algorithms" },
```
lucywyman, line 65:
```suggestion
      "hmac-algorithms":        { "$ref": "#/definitions/mac-algorithms" },
```

lib/bolt/config/transport/ssh.rb
```diff
@@ -12,94 +12,132 @@ class SSH < Base
...
+          "connect-timeout"       => { type: Integer,
+                                       desc: "How long to wait when establishing connections." },
+          "copy-command"          => { external: true,
+                                       desc: "Command to use when copying files using ssh-command. "\
+                                             "Bolt runs `<copy-command> <src> <dest>`. **This option is "\
+                                             "experimental.**" },
+          "disconnect-timeout"    => { type: Integer,
+                                       desc: "How long to wait before force-closing a connection." },
+          "encryption-algorithms" => { type: Array,
+                                       desc: "List of encryption algorithms to support when establishing a "\
```
lucywyman, line 27:
```suggestion
                                       desc: "List of encryption algorithms to use when establishing a "\
```

lib/bolt/config/transport/ssh.rb
```diff
@@ -12,94 +12,132 @@ class SSH < Base
...
+                                             "connection with a target. Supported algorithms are "\
+                                             "defined by the Ruby net-ssh library and can be viewed "\
+                                             "[here](https://github.com/net-ssh/net-ssh#supported-algorithms). "\
+                                             "All supported, non-deprecated algorithms are enabled by default when "\
```
lucywyman, line 31:
```suggestion
                                             "All supported, non-deprecated algorithms are available by default when "\
```

lib/bolt/config/transport/ssh.rb
```diff
@@ -12,94 +12,132 @@ class SSH < Base
...
+                                             "this option is not used. To enable support for default algorithms "\
```
lucywyman, line 32:
```suggestion
                                             "this option is not used. To use all default algorithms "\
```

lucywyman: 'reference'? 'specify'? might also work. Paging @hestonhoffman

lucywyman: Honestly overall this message is pretty good at communicating a weird situation. There's like...3 possible states for any algorithm to be in *in the net-ssh lib alone*, and then that's on top of whatever the user has configured locally.

lib/bolt/transport/ssh/connection.rb
```diff
@@ -275,6 +312,32 @@ def validate_ssh_version
...
+            disabled = Net::SSH::Transport::Algorithms::ALGORITHMS[match[:algo].to_sym] - enabled
+
+            message = "#{err.message}\n"\
+                      "Supported algorithms are defined by the Ruby net-ssh library. All supported, "\
+                      "non-deprecated algorithms are enabled by default. Disabled algorithms can be "\
+                      "enabled by setting the '#{match[:algo].gsub('_', '-')}-algorithms' config option.\n"\
+                      "Enabled algorithms: #{enabled.join(',')}\n"\
```
nicklewis, line 327: In practice, it's hard to follow how this information relates to the error when it occurs. Maybe it makes sense in the case where you *haven't* specified a list of preferred algorithms, but when you did, it's weird that it shows the list that you *could* have selected.

For context, with `encryption-algorithms: [ "blowfish-ctr" ]`
```
  could not settle on encryption_client algorithm
  Server encryption_client preferences: chacha20-poly1305@openssh.com,aes128-ctr,aes192-ctr,aes256-ctr,aes128-gcm@openssh.com,aes256-gcm@openssh.com,aes128-cbc,aes192-cbc,aes256-cbc,blowfish-cbc,cast128-cbc,3des-cbc
  Client encryption_client preferences: blowfish-ctr
  Supported algorithms are defined by the Ruby net-ssh library. All supported, non-deprecated algorithms are enabled by default. Disabled algorithms can be enabled by setting the 'encryption-algorithms' config option.
  Enabled algorithms: aes256-ctr,aes192-ctr,aes128-ctr
  Disabled algorithms: aes256-cbc,aes192-cbc,aes128-cbc,rijndael-cbc@lysator.liu.se,blowfish-ctr,blowfish-cbc,cast128-ctr,cast128-cbc,3des-ctr,3des-cbc,idea-cbc,none
```

PR #1883, opened 2020-06-05
(GH-1870) Fix low ulimit warning

### LINKED ISSUE #1870: Fix ulimit concurrency warning
## Describe the Bug
- The warning should only be displayed the first time the executor attempts to exceed the default concurrency during an invocation
- The calculation should be validated to ensure that we're correctly limiting the default concurrency
- The warning message should suggest changing ulimit or setting concurrency in the config and link to the docs.

lucywyman:
The low ulimit warning had a few issues:
1. It was displayed whether or not the concurrency limit was reached
   during the Bolt lifecycle
2. It miscalculated the concurrency at which users might cause the
   EMFILE error
3. It suggested setting concurrency on the CLI instead of in config

This PR rectifies each of those issues. We now only display the warning
once per Bolt run if a particular step causes the lowered concurrency
limit to be reached. The warning will only be displayed if the ulimit is
too low and the user hasn't configured the concurrency. This also sets
the default concurrency to `ulimit/7` to more accurately estimated how
many file descriptors each thread uses and try to prevent errors. Lastly
it updates the (now unprecedentedly long) warning to explain the issue,
suggest configuring ulimit or concurrency, and link to the known issues
document.

Closes #1870

!bug

* **Improve low ulimit warning** ([1870](https://github.com/puppetlabs/bolt/issues/1870))

  Users running Bolt with a low ulimit should only be warned if the
  number of targets they're running against may cause file limit issues.


lib/bolt/executor.rb
```diff
@@ -102,6 +110,15 @@ def shutdown
     # defined by the transport. Yields each batch, along with the corresponding
     # transport, to the block in turn and returns an array of result promises.
     def queue_execute(targets)
+      if @warn_concurrency && targets.length > @concurrency
+        @warn_concurrency = false
+        @logger.warn("The ulimit is low, which may cause file limit issues. We've configured concurrency to "\
+                     "'#{@concurrency}' to mitigate those issues, which may cause Bolt to run slow. "\
+                     "Configure ulimit using 'ulimit -n <limit>' in your shell configuration, or ignore this "\
+                     "warning by configuring Bolt's concurrency. "\
+                     "See https://puppet.com/docs/bolt/latest/bolt_known_issues.html for details.")
```
beechtom, line 119:
```suggestion
        @logger.warn("The ulimit is low, which may cause file limit issues. Default concurrency has "\
                     "'been set to  #{@concurrency}' to mitigate those issues, which may cause Bolt to run "\
                     "slower than usual. Disable this warning by setting the ulimit using 'ulimit -n <limit>' in  "\
                     "your shell configuration or by configuring Bolt's concurrency. "\
                     "See https://puppet.com/docs/bolt/latest/bolt_known_issues.html for details.")
```

I'm not sure if this reads better?

lib/bolt/config.rb
```diff
@@ -189,15 +189,13 @@ def initialize(project, config_data, overrides = {})
...
-                                "'--concurrency', or set your ulimit with 'ulimit -n <limit>'" }
-        @warnings << concurrency_warning
-      end
+      @modified_concurrency = if default_concurrency != DEFAULT_DEFAULT_CONCURRENCY &&
+                                 !ld_concurrency &&
+                                 !override_data.key?('concurrency')
+                                true
+                              else
+                                false
+                              end
```
beechtom, line 198: This can just directly assign the boolean expression instead of having an if/else.

PR #1887, opened 2020-06-09
(GH-1846) Exit with correct exit code when running on Windows with local transport

### LINKED ISSUE #1846: bolt --target localhost munges puppet's non-zero exit codes
## Describe the Bug
`bolt command run --targets localhost 'puppet ...'` does not correctly report exit codes that are not 0 or 1:

2020-05-27_11MS+0100_1140x327

## Expected Behavior

Notably the command in the screenshot above is returning exit code 2 and bolt should have correctly reported this.

## Steps to Reproduce

* get a win-2012r2-x86_64 machine from vmpooler
* Install latest bolt package 2.10.0
* Execute the following command: 
```
bolt command run --targets localhost 'puppet apply --trace --detailed-exitcodes "notify { foo: }"'
```
and it throws the following error

2020-05-27_12MS+0100_1237x390

Note how bolt reports exitcode 1, but the same command executed through the powershell has the expected `$LASTEXITCODE` of 2 ("something changed").

## Environment
 - Appveyor (win32) and its failing only on win32
 - Platform [windows]
 - bolt version 2.10.0

## Additional Context
In our testing we could not reproduce the behaviour with bolt 2.6.0. This is currently impacting our testing and we're rolling temporarily back to that version to unblock our pipelines.

Please let us know if you need more information. Thanks a lot for all the help.

lucywyman:
Powershell will [exit
1](PowerShell/PowerShell#11461) whenever the $LASTEXITCODE is greater
than 0, regardless of what the code is, whenever running a bare command.
Powershell exits with the correct exit code when executing files, or
when `exit $LASTEXITCODE` is appended to the command being run (as in
the [execute_process
snippet](https://github.com/puppetlabs/bolt/blob/0d44ce739d4dafcb6d27abd894e230894fdd50e0/lib/bolt/shell/powershell/snippets.rb#L19),
meaning this is only an issue when running commands through Bolt, and
not tasks or scripts. This appends a Powershell snippet to exit with
the correct exit code to commands running on Windows using the local
transport.

Closes #1846

!bug

* **Return correct exit code when running commands in powershell** ([1846](#1846))

  Bolt will now display the correct exit code when running commands in
  powershell that exit with code > 1.

### DIFF
--- a/lib/bolt/transport/local/connection.rb
+++ b/lib/bolt/transport/local/connection.rb
@@ -50,7 +50,8 @@ module Bolt
             # If it's already a powershell command then invoke it normally.
             # Otherwise, wrap it in powershell.exe.
             unless command.start_with?('powershell.exe')
-              command = ['powershell.exe', *Bolt::Shell::Powershell::PS_ARGS, '-Command', command]
+              cmd = Bolt::Shell::Powershell::Snippets.exit_with_code(command)
+              command = ['powershell.exe', *Bolt::Shell::Powershell::PS_ARGS, '-Command', cmd]
             end
           end
 
--- a/spec/integration/local_spec.rb
+++ b/spec/integration/local_spec.rb
@@ -33,6 +33,13 @@ describe "when running over the local transport" do
       expect(result[0]['_error']).to be
     end
 
+    it 'returns correct exit code', :reset_puppet_settings do
+      cmd = 'puppet apply --trace --detailed-exitcodes -e "notify {foo:}"'
+      result = run_failed_nodes(%W[command run #{cmd} -t #{uri}]).first
+      expect(result['_error']['msg']).to eq('The command failed with exit code 2')
+      expect(result['_error']['details']['exit_code']).to eq(2)
+    end
+
     it 'runs a ruby task using bolt ruby', :reset_puppet_settings do
       result = run_one_node(%w[task run sample::bolt_ruby message=somemessage] + config_flags)
       expect(result['env']).to match(/somemessage/)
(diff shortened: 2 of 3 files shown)


lib/bolt/transport/local/connection.rb
```diff
@@ -47,6 +47,7 @@ def copy_file(source, dest)
 
         def execute(command)
           if Bolt::Util.windows?
+            command = Bolt::Shell::Powershell::Snippets.exit_with_code(command)
```
nicklewis, line 50: This should be in the conditional below, so we only append the exit code stuff if we're goign to wrap it in 
`powershell.exe -Command`. Otherwise, it should already be a properly-formatted `powershell.exe` invocation.

spec/integration/local_spec.rb
```diff
@@ -33,6 +33,12 @@
       expect(result[0]['_error']).to be
     end
 
+    it 'returns correct exit code' do
+      result = run_failed_nodes(%W[command run exit(2) -t #{uri}]).first
```
nicklewis, line 37: This case already does the right thing without this change. `exit` is a keyword in powershell so it does the right thing (causes powershell.exe to exit 2), but the failure happens when you invoke another process in powershell and that process exits 2.

```
$ bolt command run 'exit(2)' -t localhost
The command failed with exit code 2
$ bolt command run 'powershell.exe -Command exit(2)' -t localhost
The command failed with exit code 2
$ bolt command run 'powershell.exe -Command "powershell.exe -Command exit(2)"' -t localhost
```

The first and second cases are actually identical, since we automatically wrap the bare `exit(2)` in a call to `powershell.exe -Command exit(2)`. But if you wrap _that_ in another powershell invocation, the outer invocation exits 1 because it throws away the exit code from the inner invocation.

PR #1889, opened 2020-06-10
(GH-1880) Raise error for empty apply block 

### LINKED ISSUE #1880: Empty apply block causes cryptic undefined method error
## Describe the Bug
When creating apply block without any content:
```
apply($targets, _run_as => root) {
}
```

Puppet bolt fails with 
```
Failed on target:
  Apply failed to compile for target: undefined method `locator' for nil:NilClass on node target
```

## Expected Behavior
No error - Puppet Bolt just checks if Puppet is installed

## Environment
 - Version [e.g. 1.27.0]: 2.11.1
 - Platform [e.g. Ubuntu 18.04]: Windows 10

lucywyman:
There's a [PR](puppetlabs/puppet#8193) to Puppet
to handle empty apply statements correctly. Until that's in, an empty
apply block will be considered unacceptable and an error will be raised,
since there's no use cases for empty apply blocks.

Closes #1880

!bug

* **Empty apply blocks now error correctly** ([#1880](#1880))

  This raises an appropriate error when an apply block is empty, instead
  of an undefined method error.

### DIFF
--- a/lib/bolt/catalog.rb
+++ b/lib/bolt/catalog.rb
@@ -1,10 +1,10 @@
 # frozen_string_literal: true
 
+require 'bolt/apply_inventory'
 require 'bolt/apply_target'
 require 'bolt/config'
 require 'bolt/error'
 require 'bolt/inventory'
-require 'bolt/apply_inventory'
 require 'bolt/pal'
 require 'bolt/puppetdb'
 require 'bolt/util'
--- a/lib/bolt/applicator.rb
+++ b/lib/bolt/applicator.rb
@@ -146,6 +146,7 @@ module Bolt
 
     def apply(args, apply_body, scope)
       raise(ArgumentError, 'apply requires a TargetSpec') if args.empty?
+      raise(ArgumentError, 'apply requires at least one statement in the apply block') if apply_body.nil?
       type0 = Puppet.lookup(:pal_script_compiler).type('TargetSpec')
       Puppet::Pal.assert_type(type0, args[0], 'apply targets')
 
--- a/spec/bolt/applicator_spec.rb
+++ b/spec/bolt/applicator_spec.rb
@@ -62,6 +62,11 @@ describe Bolt::Applicator do
     expect(applicator).to be
   end
 
+  it 'errors with an empty catalog' do
+    expect { applicator.apply([target], nil, nil) }
+      .to raise_error(ArgumentError, /apply requires at least one statement/)
+  end
+
   it 'passes catalog input' do
     expect(Open3).to receive(:capture3)
       .with('ruby', /bolt_catalog/, 'compile', stdin_data: input.merge(target_hash).to_json)


lib/bolt/catalog.rb
```diff
@@ -54,7 +54,7 @@ def generate_ast(code, filename = nil)
     end
 
     def compile_catalog(request)
-      pal_main = request['code_ast'] || request['code_string']
+      pal_main = request['code_ast'] || request['code_string'] || generate_ast('')
```
beechtom, line 57: It seems like this is really inefficient, since Bolt will need to generate the same AST for every target. 

I think it would be better to either catch the `nil` apply body early and generate the empty AST as part of `Bolt::Applicator.apply` or to cache the empty AST as part of `Bolt::Catalog.compile_catalog` so it only gets generated once.

The latter seems easier, but the former maybe seems a little more correct so the catalog input doesn't receive a `nil` value in the first place?

nicklewis:
I think we actually want to fix this in Puppet so that we will never get `nil` where we expect valid AST.

https://github.com/puppetlabs/puppet/blob/master/lib/puppet/pops/evaluator/evaluator_impl.rb#L879-L883

This code checks whether the apply body is _multiple_ statements and returns a block. Otherwise it tries to return just the first statement. But if there are _no_ statements, it returns `nil`. It should instead return an empty block when there are no statements. I think all we need to do is swap the conditional and check `if o.body.statements.count == 1`.

lucywyman:
@nicklewis I think the real question is whether an empty apply block is zero statements, or if it's one statement that's an empty string. An empty manifest file is read as an empty code string - granted that's just a side effect of Ruby's file reading, but I think having an empty file and an empty apply block conceptually behave the same will make them easier to reason about and prevent potentially weird differences in behavior.

lib/bolt/catalog.rb
lucywyman: I put this here because it ensures that any other tools that call into compile_catalog with a nil string behave the same with Bolt apply blocks. For example, Skeletor may need to write their own applicator that will use the same bolt_catalog task. I'm not sure how real that possibility is though.

lucywyman: I guess we can do both, that way this won't call generate_ast for every target but will get called if we ever end up here through other routes: https://github.com/puppetlabs/orchestrator/blob/master/src/puppetlabs/orchestrator/transports/ssh.clj#L87

donoghuc:
Can yall consider getting https://github.com/puppetlabs/bolt/pull/1798/files reviewed and merged and base this work on that refactor? I would like to get this fixed in puppetserver as well (perhaps the suggestion Nick had for improving this in puppet is a good route for that) and keeping the implementations similar will help maintain that.

lucywyman:
Relates to https://github.com/puppetlabs/puppet/pull/8193/files

@donoghuc Absolutely! I think we just weren't sure if that was ready to be reviewed and merged.

PR #1891, opened 2020-06-11
(GH-1881) Log STDERR when STDOUT from bolt_catalog is not JSON

### LINKED ISSUE #1881: Catalog compilation does not print a helpful error when there is a LoadError
## Describe the Bug
We only rescue StandardError in `bolt_catalog` which means load errors are swallowed. Instead it prints nothing to stdout which the applicator fails to parse. Helpful output may be printed to the stderr log by puppet but that is also discarded.

## Expected Behavior
We should print the logs that were captured from stderr when bolt_catalog does not return a json object. 

## Steps to Reproduce
call a ruby function that requires an uninstalled gem from manifest code.

lucywyman:
When compiling Bolt apply catalogs we only log the error message if the
error is a Puppet error or a ruby StandardError. If an error is raised
that isn't either of those types, it fails silently, and as there's
often not anything on stdout all that's raised is a JSON parse error.

Now, if the STDOUT of `bolt_catalog` is not valid JSON we log whatever
is on STDERR and include a helpful message for the ApplyError. This
should account for any case where `bolt_catalog` wasn't successful and
so didn't print JSON to STDOUT, and ensures we only print STDERR when
it's useful and not redundant with other error messages.

Closes #1881

!bug

* **Log error for unhandled catalog compilation errors** ([#1881](https://github.com/puppetlabs/bolt/issues/1881))

  We now log whatever is on STDERR when catalog compilation fails in a
  way that isn't already handled, and raise an ApplyError.

### DIFF
--- a/lib/bolt/applicator.rb
+++ b/lib/bolt/applicator.rb
@@ -106,6 +106,16 @@ module Bolt
       out, err, stat = Open3.capture3('ruby', bolt_catalog_exe, 'compile', stdin_data: catalog_input.to_json)
       ENV['PATH'] = old_path
 
+      # If bolt_catalog does not return valid JSON, we should print stderr to
+      # see what happened
+      print_logs = stat.success?
+      result = begin
+                 JSON.parse(out)
+               rescue JSON::ParserError
+                 print_logs = true
+                 { 'message' => "Something's gone terribly wrong! STDERR is logged." }
+               end
+
       # Any messages logged by Puppet will be on stderr as JSON hashes, so we
       # parse those and store them here. Any message on stderr that is not
       # properly JSON formatted is assumed to be an error message.  If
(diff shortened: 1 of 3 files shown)


lib/bolt/applicator.rb
```diff
@@ -119,17 +128,15 @@ def compile(target, catalog_input)
         { 'level' => 'err', 'message' => line }
       end
 
-      result = JSON.parse(out)
-      if stat.success?
+      if stat.success? || print_error
```
nicklewis, line 131: Since we have this variable anyway, we may as well use it as the sole condition variable. We can set `print_error = stat.success?` at the top and override it in the `rescue` if necessary when parsing results.

Maybe `print_logs` would be better because these aren't necessarily errors? That also makes this code read nicely: `if print_logs [... print the logs ...]`

spec/fixtures/apply/basic/lib/puppet/functions/load_error.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+require 'fake'
```
nicklewis, line 3: Can you add a comment indicating this line is here to raise a LoadError?

Issue #1888, opened 2020-06-10
Clean up bolt configuration reference


adreyer:
## Describe the Change You Would Like
We should have a separate configuration reference page for the different configuration files. 

- [x] bolt.yaml(soon to be deprecated)
- [x] bolt-project.yaml
- [x] inventory.yaml
- [x] bolt-defaults.yaml(soon)
- [x] Transport configuration options

* Options should be meaningfully groups as opposed to purely alphabetical. For example transport options together.
* We should try to change the phrasing of specific options to be more prescriptive
for example. connection-timeout. increase this above NNN if you see errors like ...
* These descriptions should be available from the json schema

hestonhoffman:
I'm going to jot down some thoughts here. I've been struggling with how to document experimental documentation in general, especially features like this, which replace existing functionality, and are so fundamental to the way people use Bolt.

I've been thinking a lot about where these pages (and docs on experimental features in general) should go. Regardless of what we choose, I think we need to add some kind of deprecation warning (or just a note that we're working on changing things) to the older doc/s, as well as a link to the experimental doc that describes the changes we're making. Similarly, any docs for experimental features outside of the experimental features page/section should also start with a brief explanation and a link to the experimental doc that explains the feature. 

Here are the options I see:
1. Put these alongside the existing configuration reference doc/s. This could potentially be quite confusing for users. If we do go this route, adding very clear introductions + links to the conceptual doc is very important. 
2. Add these docs as children to the experimental docs page - essentially turning it into an experimental section. I think this could work. I like the idea of keeping everything that's experimental in one place. I guess the argument for not doing this is that you'll have users who simply don't know to look there for reference docs.

PR #1893, opened 2020-06-12
(GH-1845) Add `bolt-defaults.yaml` configuration file

### LINKED ISSUE #1845: `bolt-defaults.yaml`
## Use Case
We currently support setting global(cross project) defaults at a system or user level with a `bolt.yaml` file. This is confusing since we project config currently lives in the project in a bolt.yaml file. We should rename the defaults file from `bolt.yaml` to `bolt-defaults.yaml` or `defaults.yaml`(since it's in a `bolt/`)

This file should *not* support setting project specific data like `modulepath`, `Puppetfile`, hieradata
CLI options like `color` should be exposed at the top level of the file.
the transport specific config options should be nested under a key called `inventory-config`
We should prefer the new file but fall back to the old one if not present.
We should issue a deprecation warning for the old file.

beechtom:
This adds support for a new config file named `bolt-defaults.yaml`. This
file is loaded from both the system-wide and user-level config
directories, and contains similar config options to a `bolt.yaml` file
(except for project-specific config such as `inventoryfile`).

When a user has a `bolt-defaults.yaml` file alongside a `bolt.yaml` file
in the system-wide or user-level directories, it is loaded instead of
the `bolt.yaml` file and warns the user that the `bolt.yaml` file has
been ignored.

If a user has a `bolt.yaml` file present in either the system-wide or
user-level directories, Bolt will issue a warning that the file is
deprecated and to use a `bolt-defaults.yaml` file instead.

Closes #1845 

!feature

* **Add `bolt-defaults.yaml` configuration file** ([#1845](#1845))

  Bolt now supports a new `bolt-defaults.yaml` configuration file in the
  [system-wide and user-level directories](https://puppet.com/docs/bolt/latest/configuring_bolt.html).
  This configuration file is intended to replace the `bolt.yaml`
  configuration file in the system-wide and user-level directories in
  a future version of Bolt. If a `bolt-defaults.yaml` file exists
  alongside a `bolt.yaml` file, Bolt will ignore the `bolt.yaml` file.

!deprecation

* **System-wide and user-level `bolt.yaml` is deprecated** ([#1845](#1845))

  The [system-wide and user-level](https://puppet.com/docs/bolt/latest/configuring_bolt.html)
  `bolt.yaml` files have been deprecated in favor of
  `bolt-defaults.yaml`.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -101,14 +101,14 @@ end
 namespace :docs do
   desc "Generate markdown docs for Bolt's transport configuration options"
   task :config_reference do
+    options = Bolt::Config::OPTIONS.sort.to_h
+
     @transports = { options: {}, defaults: {} }
-    @global = { options: Bolt::Config::OPTIONS, defaults: Bolt::Config::DEFAULT_OPTIONS }
+    @global = { options: options, defaults: Bolt::Config::DEFAULT_OPTIONS }
     @log = { options: Bolt::Config::LOG_OPTIONS, defaults: Bolt::Config::DEFAULT_LOG_OPTIONS }
     @puppetfile = { options: Bolt::Config::PUPPETFILE_OPTIONS }
     @apply = { options: Bolt::Config::APPLY_SETTINGS, defaults: Bolt::Config::DEFAULT_APPLY_SETTINGS }
     @project = { options: Bolt::Project::PROJECT_SETTINGS, defaults: {} }
-    @transport_option = { options: Bolt::Config::TRANSPORT_OPTION,
-                          defaults: Bolt::Config::DEFAULT_TRANSPORT_OPTION }
 
     Bolt::Config::TRANSPORT_CONFIG.each do |name, transport|
       @transports[:options][name] = transport::OPTIONS
@@ -117,6 +117,40 @@ namespace :docs do
 
     renderer = ERB.new(File.read('documentation/bolt_configuration_reference.md.erb'), nil, '-')
     File.write('documentation/bolt_configuration_reference.md', renderer.result)
+  end
+
+  desc 'Generate markdown docs for bolt-defaults.yaml'
+  task :defaults_reference do
+    @opts = {}
+
+    @opts[:options]    = Bolt::Config::BOLT_CONFIG.merge(Bolt::Config::DEFAULTS_CONFIG).sort.to_h
+    @opts[:defaults]   = Bolt::Config::DEFAULT_OPTIONS
+    @opts[:suboptions] = {}
+
+    @opts[:suboptions]['inventory-config'] = Bolt::Config::INVENTORY_CONFIG
+    @opts[:suboptions]['puppetfile']       = Bolt::Config::PUPPETFILE_OPTIONS
+    @opts[:suboptions]['puppetdb']         = Bolt::Config::PUPPETDB_OPTIONS
+
+    @transports = Bolt::Config::TRANSPORT_CONFIG.keys
+
+    renderer = ERB.new(File.read('documentation/bolt_defaults_reference.md.erb'), nil, '-')
+    File.write('documentation/bolt_defaults_reference.md', renderer.result)
+  end
+
+  task :transports_reference do
+    @opts       = {}
+    @transports = { options: {}, defaults: {} }
+
+    @opts[:options]  = Bolt::Config::INVENTORY_CONFIG
+    @opts[:defaults] = Bolt::Config::DEFAULT_OPTIONS
+
+    Bolt::Config::TRANSPORT_CONFIG.sort.each do |name, transport|
+      @transports[:options][name]  = transport::OPTIONS
+      @transports[:defaults][name] = transport::DEFAULTS
+    end
+
+    renderer = ERB.new(File.read('documentation/bolt_transports_reference.md.erb'), nil, '-')
+    File.write('documentation/bolt_transports_reference.md', renderer.result)
   end
 
   desc "Generate markdown docs for Bolt's command line options"
@@ -232,7 +266,7 @@ namespace :docs do
     File.write('documentation/plan_functions.md', renderer.result)
   end
 
-  task all: %i[cli_reference function_reference config_reference]
+  task all: %i[cli_reference function_reference config_reference defaults_reference transports_reference]
 end
 
 desc 'Generate all markdown docs'
--- a/documentation/bolt.ditamap
+++ b/documentation/bolt.ditamap
@@ -75,5 +75,7 @@
         <topicref href="plan_functions.md" format="markdown"/>
         <topicref href="bolt_types_reference.md" format="markdown" />
         <topicref href="bolt_configuration_reference.md" format="markdown"/>
+        <topicref href="bolt_defaults_reference.md" format="markdown"/>
+        <topicref href="bolt_transports_reference.md" format="markdown"/>
     </topichead>
 </map>
--- a/documentation/bolt_defaults_reference.md.erb
+++ b/documentation/bolt_defaults_reference.md.erb
@@ -1,0 +1,88 @@
+# `bolt-defaults.yaml` options
+
+This page lists the configuration options that are available in the `bolt-defaults.yaml` configuration file,
+which is supported in both the [system-wide and user-level configuration directories](configuring_bolt.md).
+
+<% @opts[:options].each do |option, desc| -%>
+## `<%= option %>`
+
+<%= desc %>
+
+<% if @opts[:defaults][option] -%>
+- **Default:** <%= @opts[:defaults][option] %>
+<% end -%>
+
+<% if @opts[:suboptions][option] -%>
+### Options
+
+This option accepts a map of additional options.
+
+<% if option == 'inventory-config' %>
+<% @opts[:suboptions][option].each do |option, _| -%>
+- `<%= option %>`
+<% end -%>
+
+A detailed description of each option, their default values, and any available sub-options
+can be viewed in [Transport configuration reference](bolt_transports_reference.md).
+<% else %>
+<% @opts[:suboptions][option].each do |option, desc| -%>
+#### `<%= option %>`
+
+<%= desc %>
+<% if @transports.include?(option) -%>
+This option accepts a map of additional options. You can view the available options in [Transport configuration reference](bolt_transports_reference.md#<%= option %>).
+<% end %>
+
+<% if @opts[:defaults][option] -%>
+- **Default:** <%= @opts[:defaults][option] %>
+<% end %>
+<% end -%>
+<% end %>
+<% end -%>
+<% end -%>
+
+## Example file
+
+```yaml
+color: true
+compile-concurrency: 5
+concurrency: 50
+format: human
+inventory-config:
+  transport: winrm
+  docker:
+    service-url: https://docker-host.example.com
+    shell-command: bash -lc
+  local:
+    cleanup: false
+    tmpdir: /tmp/bolt
+  pcp:
+    job-poll-interval: 30
+    job-poll-timeout: 60
+  remote:
+    run-on: proxy_target
+  ssh:
+    user: bolt
+    password: hunter2!
+  winrm:
+    user: bolt
+    password: hunter2!
+plugin_hooks:
+  puppet_library:
+    plugin: puppet_agent
+    stop_service: true
+plugins:
+  pkcs7:
+    private_key: ~/.ssh/my_key.pem
+    public_key: ~/.ssh/my_key.pub
+puppetdb:
+  server_urls: ["https://puppet.example.com:8081"]
+  cacert: /etc/puppetlabs/puppet/ssl/certs/ca.pem
+  token: ~/.puppetlabs/token
+puppetfile:
+  forge:
+    proxy: https://forge-proxy.org.com
+    baseurl: https://forge.org.com
+  proxy: https://forge-proxy.org.com
+save-rerun: true
+```
--- a/documentation/bolt_transports_reference.md.erb
+++ b/documentation/bolt_transports_reference.md.erb
@@ -1,0 +1,34 @@
+# Transport configuration options
+
+This page lists the configuration options that are available to each transport used by Bolt to connect
+to targets. These options can be set in multiple locations:
+
+- Under a `config` key in an [inventory file](inventory_file_v2.md).
+- As top-level keys in a `bolt.yaml` file.
+- Under the `inventory-config` key in a [`bolt-defaults.yaml` file](bolt_defaults_reference.md).
+
+<% @opts[:options].each do |option, desc| -%>
+## `<%= option %>`
+
+<%= desc %>
+
+<% if @transports[:options].key?(option) -%>
+<% @transports[:options][option].each do |t_option, data| -%>
+### `<%= t_option %>`
+
+<%= data[:desc] %>
+
+<% if data[:type] -%>
+- **Type:** <%= data[:type] == TrueClass ? 'Boolean' : data[:type] %>
+<% end -%>
+<% if @transports[:defaults][option][t_option] -%>
+- **Default:** <%= @transports[:defaults][option][t_option] %>
+<% end -%>
+<% if option == 'ssh' -%>
+- **External SSH Support:** <%= data[:external] ? 'Yes' : 'No' %>
+<% end %>
+<% end %>
+<% elsif @opts[:defaults].key?(option) -%>
+- **Default:** <%= @opts[:defaults][option] %>
+<% end %>
+<% end %>
--- a/documentation/configuring_bolt.md
+++ b/documentation/configuring_bolt.md
@@ -1,59 +1,190 @@
 # Configuring Bolt
 
 You can configure Bolt's options and features at a project level, a user level,
-or a system-wide level. 
+or a system-wide level. Unless your use case requires setting user-specific or
+system-wide configurations, configure Bolt at the project level. 
 
-Some options and features you can configure include:
-- global options
-- transports 
-- plugins
-- PuppetDB connections
-- log files 
+## Project-level configuration
 
-For a complete list of Bolt settings, see the [configuration
-reference](bolt_configuration_reference.md).
+Most of the time, you'll only need to set configuration at the project level. 
+You can set all configurable options in Bolt at the project level, and any options
+you set within a project only apply to that project. 
 
-## Configuration files
+Bolt loads project-level configuration files from the root of your [Bolt project
+directory](bolt_project_directories.md). If it can't find a project directory,
+Bolt uses the default project directory: `~/.puppetlabs/bolt/`.
 
-To configure Bolt, create a Bolt configuration file named `bolt.yaml` in one of
-the locations listed below. If a configuration file is not located at the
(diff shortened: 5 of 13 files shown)


documentation/bolt_defaults_reference.md.erb
```diff
@@ -0,0 +1,92 @@
...
+  server_urls: ["https://puppet.example.com:8081"]
+  cacert: /etc/puppetlabs/puppet/ssl/certs/ca.pem
+  token: ~/.puppetlabs/token
+puppetfile:
+  forge:
+    proxy: https://forge-proxy.org.com
+    baseurl: https://forge.org.com
+  proxy: https://forge-proxy.org.com
+save-rerun: true
+```
```
beechtom, line 92: Left out `inventory-config` for now, but wanted to see what thoughts were on having an example file that shows what having all options listed would look like. It's definitely verbose, but I think works better than having tons of smaller example files.

documentation/bolt_defaults_reference.md.erb
```diff
@@ -0,0 +1,92 @@
...
+### `<%= option %>`
+
+<%= desc %>
+
+<% if @top_level[:defaults][option] -%>
+- **Default:** <%= @top_level[:defaults][option] %>
+<% end %>
+<% end %>
+
+## Additional options
```
beechtom, line 26: Not thrilled with this header... 'Nested options', maybe?

documentation/bolt_defaults_reference.md.erb
```diff
@@ -0,0 +1,92 @@
+# `bolt-defaults.yaml` options
```
beechtom, line 1: Based on feedback in the community Slack, a list of options seems to be preferred over a table of options. Users find it easier to find what they need and like being able to link to individual options since the headers automatically receive anchors.

documentation/bolt_defaults_reference.md.erb
```diff
@@ -0,0 +1,92 @@
...
+#### `<%= option %>`
+
+<%= desc %>
+
+<% if @inventory[:defaults][option] -%>
+- **Default:** <%= @inventory[:defaults][option] %>
+<% end %>
+<% end %>
+
+<% @transports[:options].each do |transport, options| -%>
```
beechtom, line 45: This lists all of the options for each transport. I'm of two minds on having those options in this reference doc vs having a separate 'transports reference' page:

- Explicitly listing the transport options for each file that uses them means users don't have to jump back and forth between pages to get the info they need.
- This adds a lot of redundancy between reference pages and potentially adds unnecessary clutter.

This might be something better suited for #1888 , but I wanted to see what it would look like here for feedback/discussion.

Rakefile
```diff
@@ -101,14 +101,14 @@ end
 namespace :docs do
   desc "Generate markdown docs for Bolt's transport configuration options"
   task :config_reference do
+    options = Bolt::Config::OPTIONS.sort_by { |k, _| k }.to_h
```
nicklewis, line 104: I think `.sort_by { |k, _| k }` is equivalent to `sort`. Each element is an array and arrays are sorted by length and then by comparing each element in turn. So since the keys must be unique, sorting the hash will be equivalent to sorting by the first element of each array. It's a minor thing though.

beechtom: Oh, nice, it is.

documentation/bolt_defaults_reference.md.erb
lucywyman: I haven't looked deeply enough into this, but I feel like this kind of indicates that the distinction isn't necessarily meaningful to users? I think we can either distinguish transport configuration from other configuration (so this would be 'Transport configuration', or get more granular like the help text and have a heading for 'groups' of config.

lib/bolt/config.rb
```diff
@@ -34,50 +39,75 @@ class Config
...
+      "transport" => "The default transport to use when the transport for a target is not specified in the URI.",
+      "winrm"     => "A map of configuration options for the winrm transport."
+    }.freeze
 
-    # NOTE: All configuration options should have a corresponding schema property
-    #       in schemas/bolt-config.schema.json
-    OPTIONS = {
+    # Options that configure the project, such as paths to files used for a
+    # specific project. These settings are used in bolt.yaml and bolt-defaults.yaml.
+    PROJECT_CONFIG = {
```
lucywyman, line 81: Now that `bolt-project.yaml` is one file, I think we can move the `Bolt::Project::PROJECT_SETTINGS` values here as well and remove that constant from Bolt::Project

lib/bolt/config.rb
```diff
@@ -189,9 +305,13 @@ def initialize(project, config_data, overrides = {})
         'transport'           => 'ssh'
       }
 
-      loaded_data = config_data.map do |config|
-        @config_files.push(config[:filepath])
-        config[:data]
+      loaded_data = config_data.each_with_object([]) do |data, acc|
+        @warnings = @warnings.concat(data[:warnings]) if data[:warnings]
```
lucywyman, line 309: `concat` already modifies the object directly, so this shouldn't need to reassign `@warnings`

lucywyman: And since several files we load will always have this be empty for now (like project-level bolt.yaml) I think we can either not include `warnings` in those data hashes and keep the check here for nil, or make this `data[:warnings].empty?`.

lib/bolt/config.rb
beechtom: I think this change should be made as part of the config references doc changes, since it's currently used by that page to create a separate section for those options.

documentation/bolt_defaults_reference.md.erb
beechtom: Ya, the separate sections for sub-options felt a little strange. The current iteration lists the available sub-options under each option.

documentation/bolt_transports_reference.md.erb
```diff
@@ -0,0 +1,28 @@
+# Transport configuration reference
+
+This page lists the configuration options that are available to each transport used by Bolt to connect
+to targets. These options can be set in multiple locations:
+
+- Under a `config` key in an [inventory file](inventory_file_v2.md)
+- As top-level keys in a `bolt.yaml` file
+- Under the `inventory-config` key in a [`bolt-defaults.yaml` file](bolt_defaults_reference.md)
```
hestonhoffman, line 8:
```suggestion
- Under a `config` key in an [inventory file](inventory_file_v2.md).
- As top-level keys in a `bolt.yaml` file.
- Under the `inventory-config` key in a [`bolt-defaults.yaml` file](bolt_defaults_reference.md).
```

documentation/bolt_transports_reference.md.erb
```diff
@@ -0,0 +1,28 @@
...
+
+<% @transports[:options].each do |transport, options| -%>
+## `<%= transport %>`
+
```
hestonhoffman, line 12: Is there a description field we could add here?

documentation/bolt.ditamap
```diff
@@ -75,5 +75,7 @@
         <topicref href="plan_functions.md" format="markdown"/>
         <topicref href="bolt_types_reference.md" format="markdown" />
         <topicref href="bolt_configuration_reference.md" format="markdown"/>
+        <topicref href="bolt_defaults_reference.md" format="markdown"/>
```
lucywyman, line 78: I wonder if it makes sense to rename `Bolt configuration options` reference page to `bolt.yaml configuration options`. That'd be very slightly less confusing

documentation/bolt_defaults_reference.md.erb
```diff
@@ -0,0 +1,83 @@
...
+<% end -%>
+<% end -%>
+
+## Example file
+
+```yaml
+color: true
+compile-concurrency: 8
+concurrency: 100
+format: human
```
lucywyman, line 50: These are also just the defaults...is this just being generated somehow? If not I think it'd be nice to have them be non-defaults to make it a little more real, but if they are generated that's probably fine.

documentation/bolt.ditamap
beechtom: That page contains options for other files and will be part of the work for #1888

documentation/bolt_defaults_reference.md.erb
beechtom: This is manually created, but as part of the work for #1888 I'm looking at having the files auto-generated. That work has non-default values used. I'll update this to use non-defaults as well.

lucywyman: Discussed in slack - this will get fixed up as part of #1888

lib/bolt/config.rb
lucywyman: Hm, that makes me wonder if we should make the bolt.yaml configuration static so that we can update these without having to worry about legacy docs?

lib/bolt/config.rb
```diff
@@ -34,50 +39,84 @@ class Config
...
+      "concurrency"         => "100 or one-seventh of the ulimit, whichever is lower",
+      "format"              => "human",
+      "hiera-config"        => "Boltdir/hiera.yaml",
+      "inventoryfile"       => "Boltdir/inventory.yaml",
+      "modulepath"          => ["Boltdir/modules", "Boltdir/site-modules", "Boltdir/site"],
+      "save-rerun"          => true,
+      "transport"           => "ssh"
+    }.freeze
+
+    PUPPETDB_OPTIONS = {
```
lucywyman, line 113: I wonder if constants like this that are only defined for generating docs should be in a `Bolt::Config::Constants` class? Or `Bolt::Config::Options`?

nicklewis: Maybe it would make sense to keep a single hash defining all the metadata about the options (documentation _and_ default) and then have the other constants be arrays of which keys are used in which context? @beechtom think of all the `.slice`ing!

beechtom: Like this?

```ruby
    # Options that configure Bolt. These options are used in bolt.yaml and bolt-defaults.yaml.
    BOLT_CONFIG = {
      "color" => {
        desc: "Whether to use colored output when printing messages to the console.",
        type: TrueClass,
        exmp: false,
        def: true
      },
      "compile-concurrency" => {
        desc: "The maximum number of simultaneous manifest block compiles.",
        type: Integer,
        exmp: 4,
        def: "Number of cores."
      },
```

lib/bolt/config.rb
```diff
@@ -121,57 +160,143 @@ def self.from_project(project, overrides = {})
...
+    def self.user_path
+      Pathname.new(File.expand_path(File.join('~', '.puppetlabs', 'etc', 'bolt')))
+    rescue StandardError
+      nil
+    end
+
+    # Loads a 'bolt-defaults.yaml' file, which contains default configuration that applies to all
+    # projects. This file does not allow project-specific configuration such as 'hiera-config' and
+    # 'inventoryfile', and nests all default inventory configuration under an 'inventory-config' key.
+    def self.load_bolt_defaults_yaml(dir)
```
lucywyman, line 210: I like having this as a separate function because it might be nice to keep these warnings when we drop `bolt.yaml`, and will make the changes to dropping that minimal since loading each file is already a separate codepath.

lib/bolt/config.rb
```diff
@@ -189,9 +314,13 @@ def initialize(project, config_data, overrides = {})
         'transport'           => 'ssh'
       }
 
-      loaded_data = config_data.map do |config|
-        @config_files.push(config[:filepath])
-        config[:data]
+      loaded_data = config_data.each_with_object([]) do |data, acc|
+        @warnings.concat(data[:warnings]) if data[:warnings]&.any?
```
lucywyman, line 318: `data[:warnings]` should never be nil, right?

lib/bolt/project.rb
```diff
@@ -2,6 +2,7 @@
 
 require 'pathname'
 require 'bolt/pal'
+require 'bolt/config'
```
lucywyman, line 5: This circular dependency seems weird, but I don't think we can remove the project dependency from config and I'm not sure where else to put `system_path` that makes sense...

schemas/bolt-defaults.schema.json
```diff
@@ -0,0 +1,134 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
```
lucywyman, line 2: I wonder if we can generate these (not as part of this pr but eventually)

lib/bolt/project.rb
beechtom: `Bolt::Util`, maybe? `Bolt::Analytics` also uses the user path, so it could be nice to reuse it there as well.

schemas/bolt-defaults.schema.json
```diff
@@ -0,0 +1,134 @@
...
+  "title": "Bolt Project",
```
lucywyman, line 3:
```suggestion
  "title": "Bolt Defaults",
```

schemas/bolt-defaults.schema.json
```diff
@@ -0,0 +1,134 @@
...
+  "description": "Bolt Project bolt-project.yaml Schema",
```
lucywyman, line 4:
```suggestion
  "description": "Bolt Defaults bolt-defaults.yaml Schema",
```

lib/bolt/config.rb
```diff
@@ -121,57 +160,143 @@ def self.from_project(project, overrides = {})
...
+
+      { filepath: filepath, data: data, warnings: warnings }
+    end
+
+    # Loads a 'bolt.yaml' file, the legacy configuration file. There's no special munging needed
+    # here since Bolt::Config will just ignore any invalid keys.
+    def self.load_bolt_yaml(dir)
+      filepath = dir + BOLT_CONFIG_NAME
+      data     = Bolt::Util.read_optional_yaml_hash(filepath, 'config')
+      warnings = [msg: "Configuration file #{filepath} is deprecated and will be removed in a future version "\
```
lucywyman, line 260: This will always get raised whether `bolt.yaml` exists or not. It seems better to only raise it if `File.exist?(filepath)`

beechtom: Ah, good catch!

lib/bolt/project.rb
lucywyman: Yeah I noticed that too! If you have time to replace that in this PR I think that'd be great.

I think Util is more for Bolt-agnostic helper functions, and this feels Bolt-specific. I think Config is the right place for it, and in terms of ruby loading it's a nonissue, it just felt weird.

documentation/configuring_bolt.md
```diff
@@ -1,59 +1,190 @@
...
-expected path, or the file is empty, Bolt does not load the file. 
+> **Note:** Project configuration files are loaded from the [Bolt project
+> directory](bolt_project_directories.md). The default project directory is
+> `~/.puppetlabs/bolt/`.
 
-You can place `bolt.yaml` configuration files at the following paths, from
-highest precedence to lowest:
+The preferred method for setting project-level configuration is to use a
+combination of `bolt-project.yaml` and `inventory.yaml` files. This maintains
+a clear distinction between Bolt configuration and inventory configuration.
```
lucywyman, line 23: Note about deprecation?

documentation/configuring_bolt.md
```diff
@@ -1,59 +1,190 @@
...
+**Filepath:** `<project-directory>/bolt-project.yaml`
 
-  > **Note:** The project configuration file is loaded from the [Bolt project
-  > directory](bolt_project_directories.md). The default project directory is
-  > `~/.puppetlabs/bolt/`.
+The project configuration file supports options that configure how Bolt
+behaves, such as how many threads it can use when running commands on targets,
+and configures different components of the project, such as a list of plans
+and tasks that are visible to the user. Any directory containing a
+`bolt-project.yaml` file is automatically considered a [Project
```
lucywyman, line 37: I'm not sure how to convey this is only relevant if you're executing in that directory....maybe that's already obvious?

documentation/configuring_bolt.md
```diff
@@ -1,59 +1,190 @@
...
-  user.
+**Filepath:** `<project-directory>/inventory.yaml`
 
-  \*nix: `/etc/puppetlabs/bolt/bolt.yaml`
+The inventory file is a structured data file that contains groups of targets
+that you can run Bolt commands on, as well as configuration for the transports
+used to connect to the targets. Most projects will include an inventory file.
 
-  Windows: `%PROGRAMDATA%\PuppetLabs\bolt\etc\bolt.yaml`
+Inventory configuration can be set at multiple levels in an inventory file
```
lucywyman, line 66:
```suggestion
Inventory configuration can be set on any group, including the implicit 'all' group, in an inventory file
```

documentation/configuring_bolt.md
lucywyman: I like the note below. I guess my only concern right here is that people will read this far, stop reading, create a bolt.yaml because it seems easier (1 file instead of 2), then have it get dropped in a few months.

documentation/configuring_bolt.md
```diff
@@ -1,59 +1,190 @@
...
+options, including default inventory configuration options. Any directory
+containing a `bolt.yaml` file is automatically considered a [Project
+directory](bolt_project_directories.md).
+
+You can view a full list of the available options in [Bolt configuration
+options](bolt_configuration_reference.md).
+
+## User-level configuration
+
+Most configurable options in Bolt can be set at the user level. Configuration
```
lucywyman, line 98:
```suggestion
Configuration
```

lucywyman: That line feels more confusing than illuminating, and raises questions like "What is user level configuration" and "Wait, but what options aren't included"

documentation/configuring_bolt.md
```diff
@@ -1,59 +1,190 @@
...
+that should apply to all projects for a particular user should be set at this
+level. This might include paths to private keys, credentials for a plugin,
```
lucywyman, line 100: This and the next line are :100:

documentation/configuring_bolt.md
```diff
@@ -1,59 +1,190 @@
...
+
+The Bolt configuration file can be used to set all available configuration
+options, including project-specific configuration options.
+
+You can view a full list of the available options in [Bolt configuration
+options](bolt_configuration_reference.md).
+
+## System-wide configuration
+
+Most configurable options in Bolt can be set at the system level. Configuration
```
lucywyman, line 142:
```suggestion
Configuration
```

documentation/configuring_bolt.md
```diff
@@ -1,59 +1,190 @@
 # Configuring Bolt
 
 You can configure Bolt's options and features at a project level, a user level,
-or a system-wide level. 
+or a system-wide level.
 
-Some options and features you can configure include:
```
lucywyman, line 6: I really like the layout of this file - it's clear, well structured, and easy to follow

documentation/configuring_bolt.md
beechtom: Add targets to this?

```
Inventory configuration can be set on any group, including the implicit 'all' group, and any target in an inventory file
```

lucywyman: I like it

hestonhoffman:
Just a note that I have a PR up with edits: https://github.com/beechtom/bolt/pull/8

Issue #1895, opened 2020-06-13
Bolt PowerShell Cmdlets


jpogran:
## Use Case

There is a minimal [PowerShell function](https://github.com/puppetlabs/bolt-vanagon/blob/92a75061aaf455446c393a99ee21d38fb507e885/resources/files/windows/PuppetBolt/PuppetBolt.psm1#L8-L14) that detects the path to the bolt install, sets up environment variables and then executes the bolt command with ruby. It expects the user to know all arguments needed to pass to bolt to have bolt performed in intended operation. Any information about the various actions and parameters bolt has to be obtained from the bolt program. This would be fine, if the program returned the information quickly. Bolt is slower on Windows platforms than it is on other platforms, which means bolt is also slow to respond to any command, even help. This is a poor user experience.

Improving the PowerShell interface to bolt would accomplish increasing the speed of discover-ability of the bolt commands, speed up finding help on commands and parameters, and improve the speed of using the commands by presenting PowerShell native cmdlets. This doesn't change the speed of invoking the bolt command, but it does improves the main interaction point for the command line bolt user.

## Describe the Solution You Would Like

Bolt has several classes which store, parse, and validate the accepted bolt commands, actions, and parameters. These classes have all the data needed into order to build PowerShell cmdlets and their parameter sets.

A new rake task will be added that loads `bolt_options_parser` and creates a PowerShell module file containing all generated PowerShell cmdlets. This task can be run when building bolt, ensuring that the PowerShell cmdlets are always in sync with bolt commands.

The solution should provide a set of PowerShell cmdlets that adhere as closely to the standardized set of approved verb and noun pairings, while keeping the core concept of each bolt command. Each cmdlet should have the same parameters as the bolt counterpart, and be able to translate the bolt convention of hyphens in parameter names.

The built-in cmdlet help should be used, and be considered a primary required feature. Using the bolt help text in PowerShell comment based help will remove the need to invoke the bolt program to get help on a given bolt command. This will significantly speed up use of the help system.

The cmdlets should be able to translate the provided PowerShell command into a bolt command, which it then executes correctly. This means that the PowerShell cmdlets will have to know how to call bolt correctly with the given input. This may mean some duplicated mapping information.

Finally the existing bolt wrapper function should persist, as a fallback for the user to invoke the bolt commands without these new cmdlets.

## Describe Alternatives You've Considered

[...]

PR #1896, opened 2020-06-13
(GH-1895) Bolt PowerShell Cmdlets

jpogran:
- [x] Pull all commands, their actions, and their parameters from `bolt_options_parser`
- [x] Map all bolt commands to PowerShell noun convention (i.e. Verb-BoltNoun)
  - [x] Handle special cases like 'createkey' or 'show-modules' which technically are parameters but fall under the noun concept
- [x] Map all bolt actions to approved PowerShell verbs
  - [x] Take special care to choose verbs that match the type of action being performed
- [x] Convert all parameters to follow PowerShell standard convention (i.e. `ssh-command`)
  - [x] Store a mapping so that a changed parameter can be converted back
- [x] Pull help text from `bolt_options_parser`, ensuring both command level and parameter level help is stored.


Rakefile
```diff
@@ -425,3 +425,281 @@ end
...
+      actions << nil if actions.empty?
+      actions.each do |action|
+        help_text = parser.get_help_text(subcommand, action)
+        matches = help_text[:banner].match(/USAGE(?<usage>.+?)DESCRIPTION(?<desc>.+?)(EXAMPLES|\z)/m)
+        action.chomp unless action.nil?
+        if @hardcoded_cmdlets[action]
+          cmdlet_verb = @hardcoded_cmdlets[action]['verb']
+          cmdlet_noun = @hardcoded_cmdlets[action]['noun']
+        else
+          cmdlet_verb = (@pwsh_verbs[action] || action)
```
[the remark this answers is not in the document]
jpogran: I think that's a good point. I'll make it throw an error here

Rakefile
```diff
@@ -425,3 +425,281 @@ end
...
+  task :generate_module do
+    @pwsh_verbs = {
+      'apply'      => 'Invoke',
+      'convert'    => 'Convert',
+      'createkeys' => 'New',
+      'init'       => 'New',
+      'install'    => 'Install',
+      'encrypt'    => 'Protect',
+      'decrypt'    => 'Unprotect',
+      'migrate'    => 'Move',
```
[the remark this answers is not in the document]
jpogran: Good point.

In the conventions `migrate` is mapped to `move` because migrate implies movement of something. We aren't moving, we're upgrading/updating config files so it's better to use `update` instead. There isn't an approved `upgrade` verb, but we could use that anyway. They are more like guidelines than rules anyway.

.gitignore
```diff
@@ -78,8 +78,11 @@ documentation/plan_functions.md
 documentation/bolt_command_reference.md
 documentation/bolt_configuration_reference.md
 documentation/privilege_escalation.md
-documentation/bolt_defaults_reference.md
+documentation/bolt_pwsh_reference.md
+documentation/bolt_project_reference.md
```
lucywyman, line 82: This looks like a dup of the line below it.

documentation/templates/bolt_pwsh_reference.md.erb
```diff
@@ -0,0 +1,45 @@
+# Bolt PowerShell Command Reference
```
lucywyman, line 1: I think the generated markdown file needs to be added to the ditamap in order to be published: https://github.com/puppetlabs/bolt/blob/master/documentation/bolt.ditamap#L83 I think in the line below that would be a good place?

documentation/templates/bolt_pwsh_reference.md.erb
```diff
@@ -0,0 +1,45 @@
...
+
+The Bolt PowerShell module contains  PowerShell native cmdlets auto-generated from Puppet Bolt source code with correct PowerShell verb-noun semantics, familiar PowerShell parameter conventions, and identical Puppet Bolt help documentation.
```
lucywyman, line 3:
```suggestion
The Bolt PowerShell module contains PowerShell native cmdlets auto-generated from Bolt source code with correct PowerShell verb-noun semantics, familiar PowerShell parameter conventions, and identical Bolt help documentation.
```

.gitignore
jpogran: good catch

documentation/templates/bolt_pwsh_reference.md.erb
jpogran: there were more instances of 'Puppet Bolt', so i fixed all instead of adding this commit suggestion

documentation/templates/bolt_pwsh_reference.md.erb
jpogran: added to the ditamap

Rakefile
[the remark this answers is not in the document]
jpogran: Not sure I like the nested if conditionals with nil checks, but `apply` doesn't have an action so the prior lines don't catch it. Ruby for the nested ifs probably could be improved

jpogran:
Removed the generated markdown docs from dita because it should be located in the `generate_docs` task, but that would require refactoring the existing task in order not to duplicate code. Leaving that for a follow up task

documentation/templates/bolt_pwsh_reference.md.erb
```diff
@@ -0,0 +1,45 @@
+# Bolt PowerShell Command Reference
+
+The Bolt PowerShell module contains  PowerShell native cmdlets auto-generated from Bolt source code with correct PowerShell verb-noun semantics, familiar PowerShell parameter conventions, and identical Bolt help documentation.
```
lucywyman, line 3:
```suggestion
The Bolt PowerShell module contains PowerShell native cmdlets auto-generated from Bolt source code with correct PowerShell verb-noun semantics, familiar PowerShell parameter conventions, and identical Bolt help documentation.
```

Rakefile
```diff
@@ -425,3 +425,288 @@ end
...
+      'convert'    => 'Convert',
+      'createkeys' => 'New',
+      'init'       => 'New',
+      'install'    => 'Install',
+      'encrypt'    => 'Protect',
+      'decrypt'    => 'Unprotect',
+      'migrate'    => 'Update',
+      'run'        => 'Invoke',
+      'show'       => 'Get',
+      'upload'     => 'Send' # deploy? publish?
```
lucywyman, line 443: I have no idea what's typical for powershell, but I like 'send' the most of the three :shrug:

Rakefile
```diff
@@ -425,3 +425,288 @@ end
...
+
+    @commands = []
+    Bolt::CLI::COMMANDS.each do |subcommand, actions|
+      actions << nil if actions.empty?
+      actions.each do |action|
+        help_text = parser.get_help_text(subcommand, action)
+        matches = help_text[:banner].match(/USAGE(?<usage>.+?)DESCRIPTION(?<desc>.+?)(EXAMPLES|\z)/m)
+        action.chomp unless action.nil?
+
+        if action.nil? && subcommand == 'apply'
```
lucywyman, line 471: It seems like this is covered by the `else` clause in this statement?

Rakefile
```diff
@@ -425,3 +425,288 @@ end
...
+            mandatory:                  true,
+            position:                   0,
+            ruby_arg:                   'bare',
+            validate_not_null_or_empty: true
+          }
+        when 'project'
+          # bolt project init [directory] [options]
+          @pwsh_command[:options] << {
+            name:                       'directory',
+            help_msg:                   'The directory',
```
lucywyman, line 614:
```suggestion
            help_msg:                   'The directory to turn into a Bolt project',
```

Rakefile
```diff
@@ -425,3 +425,288 @@ end
...
+            mandatory:                  false,
+            type:                       'string',
+            switch:                     false,
+            position:                   0,
+            ruby_arg:                   'bare',
+            validate_not_null_or_empty: true
+          }
+        end
+
+        help_text[:flags].reject { |o| o =~ /verbose|debug|help/ }.map do |option|
```
lucywyman, line 624: Are these not valid in powershell?

Rakefile
```diff
@@ -425,3 +425,288 @@ end
...
+            ruby_short: ruby_param.short.first,
+            ruby_long:  ruby_param.long.first,
+            ruby_arg:   ruby_param.arg,
+            ruby_orig:  option
+          }
+
+          case ruby_param.class.to_s
+          when 'OptionParser::Switch::RequiredArgument'
+            # this isn't quite the same thing as a mandatory pwsh parameter
+            # pwsh_param[:mandatory] = true
```
lucywyman, line 645: Should mandatory be false in that case?

Rakefile
```diff
@@ -425,3 +425,288 @@ end
...
+          when 'concurrency'
+            pwsh_param[:validate_not_null_or_empty] = true
+          when 'compileconcurrency'
+            pwsh_param[:validate_not_null_or_empty] = true
+          when 'connecttimeout'
+            pwsh_param[:validate_not_null_or_empty] = true
+          when 'modulepath'
+            pwsh_param[:validate_not_null_or_empty] = true
+          when 'transport'
+            pwsh_param[:validate_set] = %w[winrm ssh pcp]
```
lucywyman, line 672: I think all the transports will work on Windows? https://github.com/puppetlabs/bolt/blob/master/lib/bolt/config/options.rb#L15 the keys of that hash are the complete set, if you want to compute it

Rakefile
```diff
@@ -425,3 +425,288 @@ end
...
+            pwsh_param[:parameter_set] = 'execute'
+            pwsh_param[:mandatory] = true
+            pwsh_param[:position] = 0
+          when 'params'
+            pwsh_param[:position] = 1
+            pwsh_param[:type] = nil
+          when 'modules'
+            pwsh_param[:type] = nil
+          when 'format'
+            pwsh_param[:validate_set] = %w[human json rainbow]
```
lucywyman, line 693: :rainbow:

lucywyman: Oh! Actually, we don't support rainbow on Windows. We specifically check the OS, not that we're in powershell, which may be a mistake since I think this won't work in powershell not just on Windows...

pwsh_module/command.tests.ps1
```diff
@@ -0,0 +1,242 @@
+BeforeAll {
```
lucywyman, line 1: Should we run these as part of CI? We can set up a GH Action that only runs them if the pwsh module is touched by a PR.

jpogran: Yes we should, but I didn't think about adding it in this PR. Do you think it should go in the existing `windows.yaml` file or a new file?

Rakefile
jpogran: Displaying ansi escape codes is a limitation of the terminal, both Windows and PowerShell 6/7 support them. We can investigate adding that baack if we detect the user is using Windows Terminal, or one of the Windows 10 conhosts in another ticket.

Rakefile
jpogran: My bad, I didn't look there. Thanks for pointing this out!

Rakefile
jpogran: In line 636 I set `mandatory:` to false in the intializer, so each param is set to false when created. Also all PowerShell parameters have `Mandatory` set to `$false` by default, so there isn't a need to be explicit in the pwsh code.

Rakefile
jpogran: These parameters are handled differently in powershell.

`Get-Help` is used instead of a `-help` parameter in each cmdlet.

Both `verbose` and `debug` are `CommonParameters` in cmdlets and are already provided by the cmdelt so I can't define them twice. If a user specifies either verbose or debug,PowerShell recognizes them and acts appropriately in pwsh code, and I pass them onto bolt in the final commandline.

Rakefile
jpogran: I have the link to the recommended verbs and nouns in the commit message, but I should probably include it here also

Rakefile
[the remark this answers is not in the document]
jpogran: Changed to migrate

Rakefile
jpogran: Added link

Rakefile
jpogran: I added some comments to the block. Does that help discoverability here?

Rakefile
jpogran: It now adds all transports

pwsh_module/command.tests.ps1
jpogran: I added another file since I realized we need to trigger based on path and the existing ones are for all pull requests or pushes. Now I have to figure out why pester is failing in github but passes locally

Rakefile
lucywyman: Awesome! I was mostly curious - comments are great!

Rakefile
jpogran: #1983 will add support for rainbow on windows, so I vote for keeping the enum with the full set here

Rakefile
jpogran: The `else` clause didn't cover `bolt apply` which is different than all the other commands because it has no action. I couldn't find an if/else or case statement that worked and looked nice, but I found if I made `apply` an action the original if statement works while keeping the overall intent *and* making sure it fails if something new is added we don't have a mapping for.

.gitignore
```diff
@@ -79,7 +79,10 @@ documentation/bolt_command_reference.md
 documentation/bolt_configuration_reference.md
 documentation/privilege_escalation.md
 documentation/bolt_defaults_reference.md
+documentation/bolt_pwsh_reference.md
 documentation/bolt_project_reference.md
 documentation/bolt_transports_reference.md
 tasks/
 plans/
+
+pwsh_module/pwsh_bolt.psm1
```
lucywyman, line 88: Should this also include `pwsh_module/autogenerated.tests.ps1`?

jpogran: I was initially troubled by it not being present to troubleshoot a CI failure, but realized we are running `bundle exec rake pwsh:generate_module` to get the psm1 file anyway so having the autogenerated tests present doesn't make much sense.

The only concern I have left is that we aren't tracking changes to either the psm1 or the tests, so we won't have a diff of what changes from one release to another

pwsh_module/command.tests.ps1
jpogran: Pester now works in CI because I have the scoping corrected

Rakefile
jpogran: I have left it with the if/elseif/else/end block because I need to account for `bolt apply` being different from all other commands in that it has no action. There is probably a more clever way of expressing it, but I think the code is readable enough for future use and it works

.gitignore
lucywyman: I think that's probably ok, since we can diff packages? It's a bit more cumbersome, but we're not committing the generated module anyway, so it doesn't make sense to commit the tests.

PR #1902, opened 2020-06-15
(GH-1867) Document how privilege escalation works in Bolt

### LINKED ISSUE #1867: Privilege Escalation documentation
## Create privilege escalation documentation

Privilege escalation is a critical capability for Bolt. By nature, the configuration of appropriate privilege escalation mechanisms is often carefully controlled and scrutinized, especially in large organizations, to prevent privilege exploits or breach of governance.

Bolt should provide clear documentation about how it uses privilege escalation, how to appropriately configure privilege escalation for Bolt, and relevant considerations and concerns with regards to Bolt's use of it.

See: the excellent [Ansible documentation](https://docs.ansible.com/ansible/latest/user_guide/become.html#privilege-escalation-must-be-general) on this topic. In particular, the [Privilege Escalation Must Be General](https://docs.ansible.com/ansible/latest/user_guide/become.html#privilege-escalation-must-be-general) header. Explaining that "You cannot limit privilege escalation permissions to certain commands [...]", and why, is particularly important for customers to know about Bolt's use of and requirements around privilege escalation.

lucywyman:
This adds a net new page to the docs titled `Escalating privilege with
Bolt` which details how Bolt builds the command to change users, how to
configure that command, and what limitations Bolt has with respect to
escalating privilege.

The page also includes a list of escalation-related configuration
options and command line flags, which are generated from Bolt itself the
same way the command line reference and configuration reference pages
are generated. The generation of the options on that page is done as
part of the `docs:config_reference` rake task. In order for this to
work, the run-as configuration options stored in
`Bolt::Config::Transport::SSH::OPTIONS` and
`Bolt::Config::Transport::Local::OPTIONS` had to be moved from those
constants to `Bolt::Config::Transport::Base::RUN_AS_OPTIONS` to be read
in the Rakefile. Those options are now merged into the other transport
options defined in those constants when the constant is instantiated.

!no-release-note

Closes #1867

### DIFF
--- a/lib/bolt/config/transport/local.rb
+++ b/lib/bolt/config/transport/local.rb
@@ -7,21 +7,13 @@ module Bolt
   class Config
     module Transport
       class Local < Base
-        OPTIONS = %w[
-          cleanup
-          interpreters
-          run-as
-          run-as-command
-          sudo-executable
-          sudo-password
-          tmpdir
-        ].freeze
-
         WINDOWS_OPTIONS = %w[
           cleanup
           interpreters
           tmpdir
         ].freeze
+
+        OPTIONS = WINDOWS_OPTIONS.dup.concat(RUN_AS_OPTIONS).sort.freeze
 
         DEFAULTS = {
           'cleanup' => true
--- a/lib/bolt/shell/bash.rb
+++ b/lib/bolt/shell/bash.rb
@@ -332,14 +332,14 @@ module Bolt
         end
 
         if escalate
-          if use_sudo
-            sudo_exec = target.options['sudo-executable'] || "sudo"
-            sudo_flags = [sudo_exec, "-S", "-H", "-u", run_as, "-p", sudo_prompt]
-            sudo_flags += ["-E"] if options[:environment]
-            sudo_str = Shellwords.shelljoin(sudo_flags)
-          else
-            sudo_str = Shellwords.shelljoin(@target.options['run-as-command'] + [run_as])
-          end
+          sudo_str = if use_sudo
+                       sudo_exec = target.options['sudo-executable'] || "sudo"
+                       sudo_flags = [sudo_exec, "-S", "-H", "-u", run_as, "-p", sudo_prompt]
+                       sudo_flags += ["-E"] if options[:environment]
+                       Shellwords.shelljoin(sudo_flags)
+                     else
+                       Shellwords.shelljoin(@target.options['run-as-command'] + [run_as])
+                     end
           command_str = build_sudoable_command_str(command_str, sudo_str, @sudo_id, options)
         end
 
--- a/.gitignore
+++ b/.gitignore
@@ -77,5 +77,6 @@ locales/
 documentation/plan_functions.md
 documentation/bolt_command_reference.md
 documentation/bolt_configuration_reference.md
+documentation/privilege_escalation.md
 tasks/
 plans/
--- a/Rakefile
+++ b/Rakefile
@@ -138,6 +138,25 @@ namespace :docs do
     File.write('documentation/bolt_configuration_reference.md', renderer.result)
   end
 
+  desc 'Generate privilege escalation document'
+  task :privilege_escalation do
+    @run_as = stringify_types(Bolt::Config::Transport::Options::TRANSPORT_OPTIONS
+      .slice(*Bolt::Config::Transport::Options::RUN_AS_OPTIONS))
+    parser = Bolt::BoltOptionParser.new({})
+    @run_as_options = Bolt::BoltOptionParser::OPTIONS[:escalation].map do |option|
+      switch = parser.top.long[option]
+
+      {
+        long: switch.long.first,
+        arg: switch.arg,
+        desc: switch.desc.map { |d| d.gsub("<", "&lt;") }.join("<br>")
+      }
+    end
+
+    renderer = ERB.new(File.read('documentation/templates/privilege_escalation.md.erb'), nil, '-')
+    File.write('documentation/privilege_escalation.md', renderer.result)
+  end
+
   desc 'Generate markdown docs for bolt-project.yaml'
   task :project_reference do
     @opts    = stringify_types(Bolt::Config::OPTIONS.slice(*Bolt::Config::BOLT_PROJECT_OPTIONS))
@@ -306,6 +325,7 @@ namespace :docs do
     cli_reference
     function_reference
     config_reference
+    privilege_escalation
     project_reference
     defaults_reference
     transports_reference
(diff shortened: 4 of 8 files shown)


lib/bolt/config/transport/base.rb
```diff
@@ -7,6 +7,30 @@ module Bolt
   class Config
     module Transport
       class Base
+        RUN_AS_OPTIONS = {
+          "run-as" => { type: String,
+                        external: true,
```
nicklewis, line 12: I don't love that the `external` key (which is specific the SSH transport) is here in the base class, but I don't have a good alternative.

lib/bolt/shell/bash.rb
```diff
@@ -332,14 +332,14 @@ def execute(command, sudoable: false, **options)
...
         if escalate
-          if use_sudo
-            sudo_exec = target.options['sudo-executable'] || "sudo"
-            sudo_flags = [sudo_exec, "-S", "-H", "-u", run_as, "-p", sudo_prompt]
-            sudo_flags += ["-E"] if options[:environment]
-            sudo_str = Shellwords.shelljoin(sudo_flags)
-          else
-            sudo_str = Shellwords.shelljoin(@target.options['run-as-command'] + [run_as])
-          end
+          sudo_str = if use_sudo
```
nicklewis, line 335: This is interesting. I would have expected rubocop to complain about the old way.

documentation/privilege_escalation.md.erb
```diff
@@ -0,0 +1,71 @@
...
+> Use the `run-as-command` option to specify your own sudo command and flags.
+
+You can replace this command wholesale by specifying `run-as-command`, which
+will prepend the following to the command being run:
+
+```
+<run-as-command> <run-as>
+```
+
+This enables using any executable and flags to change users. The only drawback
```
nicklewis, line 68: The other drawback might be that the run-as-command must take the user name as its last argument.

lib/bolt/shell/bash.rb
lucywyman: Same :shrug: I just prefer this, but if as a team we prefer the other way I can remove it.

nicklewis: I don't really have a preference either way. I usually write it like this because I assume rubocop will yell at me, so I just found it surprising that apparently it was fine with the old version.

lib/bolt/config/transport/base.rb
lucywyman: We could do something like merge `external => true` in for these in the Rakefile, but  I don't think that's better.

documentation/privilege_escalation.md.erb
```diff
@@ -0,0 +1,74 @@
...
+- **Type:** <%= data[:type] == TrueClass ? 'Boolean' : data[:type] %>
+<% end -%>
+<% end %>
+
+## How `run-as` works
+
+For all execution Bolt builds a command string to be run on the remote
+system. The command is built from many configuration options including
+interpreters, task input method, and `run-as` configuration. When `run-as` is
+set the following is prepended to the command being run:
```
hestonhoffman, line 48:
```suggestion
Each time Bolt executes, it builds a command string to run on the remote
system. The command is built from many configuration options, including
interpreters, task input methods, and the `run-as` configuration. When 
`run-as` is set, the following is prepended to the command:
```
Not sure about method/s. Is it methods, the method?

documentation/privilege_escalation.md.erb
```diff
@@ -0,0 +1,74 @@
...
+
+```
+<sudo-executable> -S -H -u <run-as> -p '[sudo] Bolt needs to run as another user, password:'
+```
+
+Where the default `sudo-executable` is `sudo`. If you're running a task that
```
hestonhoffman, line 54:
```suggestion
The default `sudo-executable` is `sudo`. If you're running a task that
```

documentation/privilege_escalation.md.erb
```diff
@@ -0,0 +1,74 @@
...
+reads parameters from environment variables `-E` will be appended.
```
hestonhoffman, line 55:
```suggestion
reads parameters from environment variables, `-E` will be appended.
```

documentation/privilege_escalation.md.erb
```diff
@@ -0,0 +1,74 @@
...
+<run-as-command> <run-as>
+```
+
+This enables using any executable and flags to change users. When setting
+`run-as-command` it's important to note that it's non-interactive, so if the
+system prompts for a password Bolt will error instead of supplying the password
+like we do for vanilla `run-as`. Bolt will also always append the specified
+`run-as` user to the specified `run-as-command`, which can be limiting for some
+escalating commands. Lastly, `run-as-command` will not be used if `run-as` is
+not set.
```
hestonhoffman, line 74:
```suggestion
This enables using any executable and flags to change users. Keep in mind that: 
- The `run-as-command` is non-interactive. If you set `run-as-command` and the
  system prompts for a password, Bolt will error instead of supplying the password. 
- Bolt always appends the specified `run-as` user to the specified `run-as-command`,
  which can be limiting for some escalating commands. 
- Bolt will not use the `run-as-command` option if `run-as` is not set.
```

documentation/privilege_escalation.md.erb
lucywyman: I think method, since it's one option?

documentation/privilege_escalation.md.erb
```diff
@@ -0,0 +1,77 @@
+# Escalating privilege with Bolt
+
+By default Bolt connects to, and executes on, remote systems as the same user.
```
hestonhoffman, line 3:
```suggestion
By default, Bolt connects to, and executes on, remote systems as the same user.
```

lib/bolt/config/transport/local.rb
```diff
@@ -7,22 +7,14 @@ module Bolt
...
-          tmpdir
-        ].freeze
-
         WINDOWS_OPTIONS = %w[
           cleanup
           interpreters
           tmpdir
         ].freeze
 
+        OPTIONS = WINDOWS_OPTIONS.dup.concat(RUN_AS_OPTIONS).freeze
```
beechtom, line 16: Concatenating the arrays has the side effect of listing the options out of order in the reference docs. I think there are a couple options here:

1. Sort the array after concatenating
1. Sorting the array before slicing the `OPTIONS` hash in the rake task
1. Don't concatenate the run-as options and just have the array in the base class that is only used by the rake task for the privilege escalation doc

I personally lean towards 3, just because it makes the code a bit easier to follow since you don't need to reference the `RUN_AS_OPTIONS` array when seeing what options are available and we don't group other options like this... but I don't feel super strongly about it.

PR #1911, opened 2020-06-17
(fun) Add rainbow output format

lucywyman:
This adds a new output format option `rainbow` which behaves like the
human outputt except that it will colorize success messages with rainbow
colors instead of green if the system supports it. This works by adding
a new outputter class `Bolt::Outputter::Rainbow` that inherits from the
`Bolt::Outputter::Human` class and overrides the `colorize` and
`print_summary` functions. The colorize function in the rainbow class
will color messages labelled as `:green` or `:rainbow` with rainbow
colors using an algorithm copied from the [lolcat
project](https://github.com/busyloop/lolcat) (which is licensed for free
modification and reuse). The `print_summary` function is overridden to
colorize success messages that aren't colorized in the Human class.

This option can be configured on the cli with `--format rainbow` or in
config with `format: rainbow`.

This option is not available on Windows, and will throw a validation error.

!feature

* **Add rainbow output format**

  This adds a new format option `rainbow` which will print success
  messages in rainbow colors. This option is not available on Windows.

### DIFF
--- a/bolt.gemspec
+++ b/bolt.gemspec
@@ -47,6 +47,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency "net-ssh", ">= 4.0"
   spec.add_dependency "net-ssh-krb", "~> 0.5"
   spec.add_dependency "orchestrator_client", "~> 0.4"
+  spec.add_dependency "paint", "~> 2.2"
   spec.add_dependency "puppet", [">= 6.16.0", "< 7"]
   spec.add_dependency "puppet-resource_api", ">= 1.8.1"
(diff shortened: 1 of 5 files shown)


lucywyman:
Shipping depends on https://github.com/puppetlabs/puppet-runtime/pull/347

nicklewis:
Does this work on windows?

lucywyman:
Hm, at least not on Windows 2012 in powershell 5: image

But it at least doesn't seem to break.

nicklewis:
Not broken is good enough for me. I just wanted to make sure it wouldn't have stray ANSI codes or anything.

Issue #1871, opened 2020-06-03
Require project name in bolt-project.yaml instead of defaulting to the directory name


lucywyman:
We currently attempt to infer a project name if it isn't set in `bolt-project.yaml`, which seems to cause more confusion than help. We should instead only load content from the bolt project if a `name` is set in bolt-project.yaml, and that content should be namespaced to the specified name. This should warn users with a bolt-project.yaml that doesn't have a name that project-level content won't be loaded.

PR #1913, opened 2020-06-18
(GH-1871) Don't infer project name from directory name

nicklewis:
Previously, we would infer the project name from the directory name if
no name was specified in the bolt-project.yaml. This behavior proved to
be confusing for users, so we now require `name` to be specified
in bolt-project.yaml and otherwise will warn and not load project
content.

We previously recognized `<author>-<name>` directory names as valid
project names (using only the `<name>` portion) as that is a common
repo naming scheme for modules. As the project name must now be
specified explicitly, we instead require it be simply the name portion.

!deprecation

  * **Project names must be explicitly specified** ([#1871](https://github.com/puppetlabs/bolt/issues/1871))

    Proejct names must now be specified in `bolt-project.yaml` in order
    for project-level content to be loaded, rather than the name being
    inferred by the name of the project directory.

### DIFF
--- a/lib/bolt/pal.rb
+++ b/lib/bolt/pal.rb
@@ -137,7 +137,9 @@ module Bolt
       # TODO: If we always call this inside a bolt_executor we can remove this here
       setup
       r = Puppet::Pal.in_tmp_environment('bolt', modulepath: @modulepath, facts: {}) do |pal|
-        Puppet.override(bolt_project: @project,
+        # Only load the project if it a) exists, b) has a name it can be loaded with
+        bolt_project = @project if @project&.name
+        Puppet.override(bolt_project: bolt_project,
                         yaml_plan_instantiator: Bolt::PAL::YamlPlan::Loader) do
           pal.with_script_compiler do |compiler|
             alias_types(compiler)
--- a/lib/bolt/project.rb
+++ b/lib/bolt/project.rb
@@ -116,10 +116,7 @@ module Bolt
     end
 
     def name
-      # If the project is in mymod/Boltdir/bolt-project.yaml, use mymod as the project name
-      dirname = @path.basename.to_s == 'Boltdir' ? @path.parent.basename.to_s : @path.basename.to_s
-      pname = @data['name'] || dirname
-      pname.include?('-') ? pname.split('-', 2)[1] : pname
+      @data['name']
     end
 
     def tasks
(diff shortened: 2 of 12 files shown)


spec/bolt/project_spec.rb
```diff
@@ -78,9 +78,9 @@
     describe "with namespaced project names" do
       let(:config) { { 'name' => 'puppetlabs-foo' } }
 
-      it "strips namespace and hyphen" do
-        project = Bolt::Project.new(config, pwd)
-        expect(project.name).to eq('foo')
+      it "raises an error" do
+        expect { Bolt::Project.new(config, pwd).validate }
+          .to raise_error(/Invalid project name 'puppetlabs-foo' in bolt-project.yaml/)
```
lucywyman, line 83: I think we also need to add `name:` to the bolt-project.yaml files in `spec/fixtures/projects`.

lucywyman:
This made me realize that in Puppet, for loading the project as a module, we assume that the project will *always* have a name so we always load it as a module. With this change, that means we load a module with name `''` if name isn't set.

```
lucy ~/githubs/bolt (GH-1871-require-project-name) $ be bolt plan run foo -t localhost
No project name is specified in bolt-project.yaml. Project-level content will not be available.
        Invalid module name ''; module names must match either:
        An installed module name (ex. modulename) matching the expression /^[a-z][a-z0-9_]*$/ -or-
        A namespaced module name (ex. author-modulename) matching the expression /^[a-zA-Z0-9]+[-][a-z][a-z0-9_]*$/
Could not find a plan named "foo". For a list of available plans, run "bolt plan show"
```

I think we need to only set `bolt_project` in PAL if the project has a name:
```
        overrides = { yaml_plan_instantiator: Bolt::PAL::YamlPlan::Loader }
        overrides[:bolt_project] = @project if @project&.name
        Puppet.override(overrides) do
```
And do the same in `compile_catalog`

lib/bolt/project.rb
```diff
@@ -130,36 +127,20 @@ def plans
...
-        Configure project name in <project_dir>/bolt-project.yaml
-        ERROR_STRING
-      # If the project name is the same as one of the built-in modules raise a warning
-      elsif Dir.children(Bolt::PAL::BOLTLIB_PATH).include?(name)
-        raise Bolt::ValidationError, "The project '#{name}' will not be loaded. The project name conflicts "\
-          "with a built-in Bolt module of the same name."
+      if name
+        name_regex = /^[a-z][a-z0-9_]*$/
+        if name !~ name_regex
+          raise Bolt::ValidationError, <<~ERROR_STRING
```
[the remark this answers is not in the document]
lucywyman: This is the error if I put `Bolt::Project.create_project(Dir.mktmpdir.to_s)` in `Bolt::Config.default`:

```
lucy ~/githubs/bolt (GH-1871-require-project-name) $ be bolt plan show
The project 'file' will not be loaded. The project name conflicts with a built-in Bolt module of the same name.
```

nicklewis:
Ah I see. Projects previously always had a name, now they will only have a name if there's a project file. But we also only want to load them if there's a project file. So I think we actually want to make the decision at the point where we currently call `project_file?`. Perhaps that ought to be renamed back to `load_as_module?` heh

spec/bolt/project_spec.rb
nicklewis: Whoops apparently I forgot to commit those files. Classic.

lucywyman:
It's not just if you have a project file though - because now projects will only have a name if you set `name` in the file, which is still optional. We only raise a warning if name isn't set that the project won't be loaded as a module. We should either warn and not set project in PAL / catalog if name isn't set, or error if name isn't set.

nicklewis:
Sorry yes, what I meant is that we _don't_ need to check if there's a project file, since we also need there to be a name, which requires a project file. So I will replace the current `project_file?` check.

lib/bolt/project.rb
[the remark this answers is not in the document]
nicklewis: This is interesting. It looks like the whole reason we need a default config is for the case where the config is invalid, since we use `config.format` and `config.color` to setup the outputter that we use to log the error. Because it's the default config, that means it disregards the output format the user wants. Maybe the right thing here is to not rely on the outputter at all for errors that happen this early in the lifecycle, so we can get rid of the default config entirely.

Separately, I do think it makes sense to have a `Bolt::Project.empty` or similar that will be used by the default config, rather than using the cwd. I will dig into that a bit more.

lucywyman:
This is weird - I pulled down your branch and can see the change in `bolt/lib/pal.rb`, but I don't see it in the UI diff here? Is it down or just me?

lib/bolt/pal.rb
```diff
@@ -137,7 +137,9 @@ def in_bolt_compiler
       # TODO: If we always call this inside a bolt_executor we can remove this here
       setup
       r = Puppet::Pal.in_tmp_environment('bolt', modulepath: @modulepath, facts: {}) do |pal|
-        Puppet.override(bolt_project: @project,
+        # Only load the project if it a) exists, b) has a name it can be loaded with
+        bolt_project = @project if @project&.name
```
lucywyman, line 141: I think we need to do this same thing in catalog.rb:
```
lucy ~/githubs/bolt ((5697f54...)) $ be bolt plan run bolt::foo -t localhost --verbose
No project name is specified in bolt-project.yaml. Project-level content will not be available.
Starting: plan bolt::foo
Starting: install puppet and gather facts on localhost
Finished: install puppet and gather facts with 0 failures in 2.29 sec
localhost:         Invalid module name ''; module names must match either:
        An installed module name (ex. modulename) matching the expression /^[a-z][a-z0-9_]*$/ -or-
        A namespaced module name (ex. author-modulename) matching the expression /^[a-zA-Z0-9]+[-][a-z][a-z0-9_]*$/
Started on localhost...
Finished on localhost:
  Notice: /Stage[main]/Main/Notify[foo]/message: defined 'message' as 'foo'
  changed: 1, failed: 0, unchanged: 0 skipped: 0, noop: 0
Finished: apply catalog with 0 failures in 3.89 sec
Finished: plan bolt::foo in 6.18 sec
```

```
bolt_project = Struct.new(:name, :path).new(project['name'], project['path']) unless project['name'].nil?
```

lucywyman:
It'd be great to have an integration test that ensures that a plan or task in a project without a defined name doesn't get loaded and warns there's no name, either by mocking out loading the project file or just laying down fixtures.

nicklewis:
It's a little unclear how to test that a task/plan doesn't get loaded, because the task/plan doesn't have a name if the project doesn't have a name. There's simply no way to refer to that content in the first place, so there's nothing we can _try_ to load and ensure it isn't loaded. I can add a test that PAL doesn't pass the project as `bolt_project` in the context if it has no name.

nicklewis:
Updated with some  new tests

lucywyman:
I don't see the `catalog.rb` updates?

nicklewis:
Whoops, I missed that comment today. I blame the weekend.

lib/bolt/pal.rb
nicklewis: Updated. I added the logic to the applicator to not include the project in the compilation 
 context if it doesn't have name.

PR #1921, opened 2020-06-22
Bump logging to notice level with verbose option

### LINKED ISSUE #1920: Bolt Apply log output
When I run Bolt Apply with a notice statement, the only way to view that notice statement is with `--debug`. For example
```test.pp
notice("This is a test: $::fqdn")
```
```
$ bolt apply test.pp -t target
```

That includes *a lot* of output.

I'd expect to see notice statements either by default, or at least when I run with `--verbose`.

MikaelSmith:
This seemed like the minimal change required to print notice statements for `bolt apply --verbose`.

Fixes #1920

### DIFF
--- a/lib/bolt/config.rb
+++ b/lib/bolt/config.rb
@@ -228,6 +228,11 @@ module Bolt
       # Set console log to debug if in debug mode
       if options[:debug]
         overrides['log'] = { 'console' => { 'level' => :debug } }
+      elsif options[:verbose]
+        configured_level = overrides.dig('log', 'console', 'level')
+        if configured_level.nil? || Bolt::Logger.lower_level?(:notice, configured_level)
+          overrides['log'] = { 'console' => { 'level' => :notice } }
+        end
       end
 
       if options[:puppetfile_path]
--- a/lib/bolt/logger.rb
+++ b/lib/bolt/logger.rb
@@ -93,6 +93,11 @@ module Bolt
       !Logging.level_num(level).nil?
     end
 
+    # Returns if level is lower severity than baseline
+    def self.lower_level?(level, baseline)
+      Logging.level_num(level) < Logging.level_num(baseline)
+    end
+
     def self.levels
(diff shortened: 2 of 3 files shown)


MikaelSmith:
For my example from #1920 
```
Starting: plan
Starting: install puppet and gather facts on target
Finished: install puppet and gather facts with 0 failures in 24.34 sec
Starting: apply catalog on target
target: Scope(Class[main]): This is a test: target.example.com
target: Scope(Class[main]): Reboot time: 02:00
Started on target...
Finished on target:
  changed: 0, failed: 0, unchanged: 0 skipped: 0, noop: 0
Finished: apply catalog with 0 failures in 19.68 sec
Successful on 1 target: target
Ran on 1 target in 44.15 sec
```

nicklewis:
We might want to lower the "Starting: plan" and "Finished: plan" messages to `info` in this case. Then there will be no messages logged at `notice` level in Bolt itself, which I think is ideal.

MikaelSmith:
There doesn't seem to be a `Finished: plan` in this case. I'm ambivalent on `Starting: plan`.

lib/bolt/config.rb
```diff
@@ -228,6 +228,11 @@ def normalize_overrides(options)
       # Set console log to debug if in debug mode
       if options[:debug]
         overrides['log'] = { 'console' => { 'level' => :debug } }
+      elsif options[:verbose]
```
MikaelSmith, line 231: Logging in Bolt still seems to be a little messy. I think this is more correct than what I had before, but made sure I wouldn't overwrite an existing level. And if console logging gets more options, both debug and verbose need to be fixed here.

MikaelSmith:
Sounds like you want to solve this differently based on @lucywyman's comment on the issue.

lucywyman:
I think so :/ It seemed to make more sense to just handle proper log levels on the CLI and not force users to set them in config. No pressure, I know you're busy, but if you put up a PR with that solution we'd gladly review!

PR #1922, opened 2020-06-22
(MAINT) Remove whitelist terminology from bolt documentation

npwalker:
This rewords our documentation to not use the phrase 'whitelist',
preferring just 'list' and rewording to describe what the list does
rather than using the term 'whitelist'.

!no-release-note

### DIFF
--- a/documentation/experimental_features.md
+++ b/documentation/experimental_features.md
@@ -28,10 +28,10 @@ where you're looking to share orchestration that is specific to your
 infrastructure, it's not always necessary, and can be cumbersome to get going
 quickly. 
 
-We also needed a way for content authors to whitelist the plans and tasks that
-they've created, so that when they share the content with other users, those
-users can run `bolt plan show` or `bolt task show` from the project directory
-and be presented with a list of only the content they need to see. 
+We also needed a way for content authors to display a subset of plans and tasks
+that they've created, so that when they share the content with other users,
+those users can run `bolt plan show` or `bolt task show` from the project
+directory and be presented with a list of only the content they need to see.
 
 ### Using Bolt projects
 
@@ -100,11 +100,11 @@ lowercase letter.
 
 > **Note:** Projects take precedence over installed modules of the same name.
 
-#### Whitelisting plans and tasks
+#### Limiting displayed plans and tasks
 
-To control what tasks and plans appear when your users run `bolt plan show` or
-`bolt task show`, add `tasks` and `plans` keys to your `bolt-project.yaml` and
-include an array of task and plan names. 
+To control what plans and tasks appear when your users run `bolt plan show` or
+`bolt task show`, add `plans` and `tasks` keys to your `bolt-project.yaml` and
+include an array of plan and task names.
 
 For example, if you wanted to surface a plan named `myproject::myplan`, and a
 task named `myproject::mytask`, you would use the following `bolt-project.yaml`
--- a/lib/bolt/project.rb
+++ b/lib/bolt/project.rb
@@ -9,8 +9,10 @@ module Bolt
     BOLTDIR_NAME = 'Boltdir'
     PROJECT_SETTINGS = {
       "name"  => "The name of the project",
-      "plans" => "An array of plan names to whitelist. Whitelisted plans are included in `bolt plan show` output",
-      "tasks" => "An array of task names to whitelist. Whitelisted plans are included in `bolt task show` output"
+      "plans" => "An array of plan names to show, if they exist in the project."\
+                 "These plans are included in `bolt plan show` output",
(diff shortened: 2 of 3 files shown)


lib/bolt/project.rb
```diff
@@ -8,8 +8,8 @@ class Project
     BOLTDIR_NAME = 'Boltdir'
     PROJECT_SETTINGS = {
       "name"  => "The name of the project",
-      "plans" => "An array of plan names to whitelist. Whitelisted plans are included in `bolt plan show` output",
-      "tasks" => "An array of task names to whitelist. Whitelisted plans are included in `bolt task show` output"
+      "plans" => "An array of plan names to show. These plans are included in `bolt plan show` output",
+      "tasks" => "An array of task names to show. These tasks are included in `bolt task show` output"
```
npwalker, line 12: I wonder if the plans and tasks keys should really be `displayed_plans` and `displayed_tasks` or something similar.  

From the key name you might think you're supposed to put all of your tasks/plans in that key.  Granted the description should help clear up that ambiguity.

nicklewis: Hmm this loses a subtle distinction which is that these aren't the names _to_ show, they're the names that _can_ be shown if they exist. If you don't have a task with that name, it won't show up even if it's in this list. Maybe "An array of task names that are part of the project. These tasks are included in `bolt task show` output"?

npwalker: "An array of plan names to show. These plans are included in `bolt plan show` output as long as they also exist inside the project"

or 

"An array of plan names to show, if they exist in the project. These plans are included in `bolt plan show` output"

hestonhoffman:
My understanding is that there's an ongoing conversation about this (and other similar) terminology at Puppet. I'd rather hold off on making this change until we've agreed on the terminology we're using company-wide, and we can make the changes across Puppet's documentation at the same time.

nicklewis:
I don't think a one-for-one substitute for the word whitelist is what we need here. Specifically, one of the problems with the term is its vagueness. Therefore, we are likely to standardize on a more specific term, and that more specific term will reflect the _most common_ usage of whitelist which is in an authorization context. But we're using the word in a filtering context here. Therefore, the replacement word we standardize on will probably not fit well here.

In fact, I think our use of the word whitelist here serves explicitly to rely on its vagueness to let us shift the burden of understanding onto our users. It's really not clear what a "whitelist of tasks" means, but nor is it technically incorrect, so we've deemed it good enough. We're better off rephrasing to make it clear what this list actually means.

hestonhoffman:
I'm not advocating for a find+replace. And I think `allowlist` is probably fine here, but there are probably a small set of contexts where the terminology is used - this may not the only place in the docs where the term is used to imply filtering - and I'm sure this is not the first time we've had to look at terminology across the org and figure out what term/s we need to replace it with.

I just think it makes sense to wait for the ongoing conversation to take place, and for tech pubs to have a chance to address this type of terminology all at once for the sake of consistency.

npwalker:
I don't think we should use allowlist here... I thought I was going to find and replace but it didn't really apply here.  I think we can just use the word list instead.  I pushed up a commit for that.  

@hestonhoffman I'm all for aligning with tech pubs on this but why not hash out what we as a team think we should use for this case and then that can be information for the broader discussion you have with tech pubs?  Granted we shouldn't spend a lot of time on this conversation but in this case the removal of whitelist seems like an improvement in clarity anyway and might not be directly related to the overarching conversation.

documentation/experimental_features.md
```diff
@@ -28,10 +28,10 @@ where you're looking to share orchestration that is specific to your
 infrastructure, it's not always necessary, and can be cumbersome to get going
 quickly. 
 
-We also needed a way for content authors to whitelist the plans and tasks that
-they've created, so that when they share the content with other users, those
-users can run `bolt plan show` or `bolt task show` from the project directory
-and be presented with a list of only the content they need to see. 
+We also need a way for content authors to display a subset of plans and tasks
```
hestonhoffman, line 31:
```suggestion
We also needed a way for content authors to display a subset of plans and tasks
```
The rest of this section is in past tense.

hestonhoffman:
That makes sense. I like the changes 👍

hestonhoffman:
Oh, I just took another look and there's a section below that contains "whitelist" in the heading, and describes the process in more detail - that's going to need an edit as well. I can do it tomorrow if nobody's got to it by then.

npwalker:
Yea good call, I didn't finish my ctrl+f all the way through.  https://github.com/puppetlabs/bolt/pull/1922/commits/278fc923f9a09c9513e3c0d6e922b91d947a8023

If we're good with these I can squash them to a single commit.

PR #1923, opened 2020-06-22
(GH-1888) Create separate reference docs for config files

### LINKED ISSUE #1888: Clean up bolt configuration reference

beechtom:
This adds separate reference docs for each of Bolt's configuration
files. Previously, configuration options were listed under a single
monolithic 'Bolt configuration options' page, which made it difficult to
know which options were available to which files. Reference pages are
now available for the following files:

- `bolt.yaml`
- `inventory.yaml`
- `bolt-project.yaml`
- `bolt-defaults.yaml`

This also adds a separate reference page for transport configuration
options, as these options can be set in multiple files.

Closes #1888 

!no-release-note


beechtom:
Changes are up on the preview site.

lib/bolt/config.rb
```diff
@@ -39,114 +39,320 @@ class Config
...
+              "the project directory as though it were a module.",
+        type: String,
+        exmp: "myproject"
+      },
+      "plans" => {
+        desc: "A list of plans to make visible in `bolt plan show` output. This option is used to limit "\
+              "the visibility of plans for users of the project. For example, project authors might want to "\
+              "limit the visibility of plans that are bundled with Bolt or plans that should only be run as "\
+              "part of another plan. When this option is not configured, all plans are visible. This "\
+              "option does not prevent users from running plans that are not part of the list of visible plans.",
```
beechtom, line 132: Will update this and the `tasks` description once #1922 is in.

hestonhoffman:
We probably want to add the new reference files to gitignore right?

documentation/bolt_inventory_reference.md
```diff
@@ -0,0 +1,376 @@
+# `inventory.yaml` options
+
+The [inventory file](inventory_file_v2.md) is used to store information about
+your targets and arrange them into groups. Both targets and groups can
+include data such as facts or vars and can configure the transports Bolt uses
+to establish connections with targets.
+
+## Options
```
hestonhoffman, line 8:
```suggestion
## Top-level options

The top level of your inventory file contains options that configure the implicit `all` group. These options apply to all of the groups and targets and targets in the inventory file. For more information on inventory files, such as precedence, see [Inventory files](inventory_file_v2.md)
```
The next heading is "Group objects", but obviously contains some identical options. We need something to make it clearer that all of these options are for the `all` group. I think top-level is fine, but I'm open to suggestions. I think adding the short intro paragraph is also helpful

documentation/bolt_inventory_reference.md
```diff
@@ -0,0 +1,376 @@
...
+- **Type:** Hash
+
+```yaml
+vars:
+  ssh_config: /etc/ssh_config
+```
+
+## Group objects
+
+The `groups` key is used to specify a list of groups, which contain a list of
```
hestonhoffman, line 105:
```suggestion
Use a `groups` key to specify a list of groups, which contain a list of
```

documentation/bolt_inventory_reference.md
```diff
@@ -0,0 +1,376 @@
...
+targets. Each item in the `groups` list is a map of data and configuration for
+the group. Group objects accept many of the same keys as the implicit `all`
+group.
+
+### `config`
+
+A map of configuration options for the group. Configuration set at this level
+applies to all groups and targets under the group. A detailed description of
+each option, their default values, and any available sub-options can be viewed
+in [Transport configuration reference](#bolt_transports_reference.md).
```
hestonhoffman, line 115:
```suggestion
applies to all groups and targets under the group. For a detailed description of
each option, their default values, and any available sub-options,
see [Transport configuration reference](#bolt_transports_reference.md).
```

documentation/bolt_inventory_reference.md
```diff
@@ -0,0 +1,376 @@
...
+  - uri: linux-2.example.com
+    alias:
+      - webserver
+```
+
+### `config`
+
+A map of configuration options for the target. A detailed description of each
+option, their default values, and any available sub-options can be viewed in
+[Transport configuration reference](#bolt_transports_reference.md).
```
hestonhoffman, line 250:
```suggestion
[Transport configuration reference](bolt_transports_reference.md).
```

documentation/templates/bolt_configuration_reference.md.erb
```diff
@@ -0,0 +1,61 @@
+# `bolt.yaml` options
```
hestonhoffman, line 1:
```suggestion
# DEPRECATED: `bolt.yaml` options
```
Let's see what this looks like in preview 😄  I've been doing some digging and some of our other products do this: https://puppet.com/docs/continuous-delivery/3.x/continuous_delivery_agent.html#continuous_delivery_agent

documentation/templates/bolt_defaults_reference.md.erb
```diff
@@ -0,0 +1,46 @@
...
+
+- **Type:** <%= data[:type] %>
+<% if data.key?(:def) -%>
+- **Default:** <%= data[:def] %>
+<% end -%>
+
+<% if @subopts.key?(option) -%>
+<% if option == 'inventory-config' %>
+A detailed description of each option, their default values, and any available sub-options
+can be viewed in [Transport configuration reference](bolt_transports_reference.md).
```
hestonhoffman, line 21:
```suggestion
For a detailed description of each option, their default values, and any available 
sub-options, see [Transport configuration reference](bolt_transports_reference.md).
```

documentation/bolt_inventory_reference.md
nicklewis: These are more like _fields_ than _options_ because this file is more akin to data than configuration.

documentation/bolt_inventory_reference.md
```diff
@@ -0,0 +1,376 @@
...
+
+A map of variables for the target.
+
+- **Type:** Hash
+
+```yaml
+targets:
+  - uri: linux.example.com
+    vars:
+      ssh_config: /etc/ssh_config
```
nicklewis, line 335: Missing ``` here causes the example file to be part of this code block and to not appear in the table of contents.

documentation/templates/bolt_configuration_reference.md.erb
beechtom: Updated the preview site. :)

documentation/bolt_inventory_reference.md
beechtom: Updated.

documentation/templates/bolt_configuration_reference.md.erb
hestonhoffman: I don't hate it

lucywyman:
The division of these into pages feels confusing: bolt-defaults and bolt-project also share config options, but those don't get a separate page? Plus I think users will want to navigate the reference based on what they want to configure, not where it's being configured.

I see 4 (kind of 5) categories of config:
| Config | bolt.yaml | bolt-default.yaml | bolt-project.yaml | inventory.yaml |
| --- | --- | --- | --- | --- |
| **Configuring Bolt behavior (Global)** | ✅ | ✅ | ✅ | ❌ |
| **Configuring connections (Transport)** | ✅ | ✅ under `inventory-config` key | ❌ | ✅ |
| **Configuring Bolt environment (Project)** | ✅ Except for `name`, `tasks`, `plans` | ❌ |✅ | ❌ |
| **Target data (Inventory)** | ❌ | ❌ | ❌ | ✅ |

hestonhoffman:
I guess the assumption we made is that people will read https://puppet.com/docs/bolt/latest/configuring_bolt.html first to figure out which files they need and where to put them, and they'll use the relevant config file when they're looking for a specific option. A table on that page is a great idea. The configuring Bolt page should cover the overlap between bolt-defaults and bolt-project. 

I think no matter what we do, this is going to be at least a little confusing as long as we're trying to document the old way and the new way at the same time. It will be drastically simpler once bolt.yaml is gone.

documentation/bolt.ditamap
```diff
@@ -77,6 +77,8 @@
         <topicref href="bolt_types_reference.md" format="markdown" />
         <topicref href="bolt_configuration_reference.md" format="markdown"/>
```
lucywyman, line 78: Might be nice to move this to the bottom of the list, since it's deprecated?

beechtom:
I do think that maintaining support for bolt.yaml is making this a bit more confusing than it otherwise would be, as it means we have multiple config files for a user to choose from at each level. However, once bolt.yaml is removed I think there is much less potential for confusion on where to look for settings, as there’s really only one file at each level that you can set config, and for a majority of users that will be the bolt-project.yaml (and inventory.yaml for the transport settings).

I’m open to moving the transports reference into each of the reference pages that can configure transports. My thought with separating it, though, is that there are a _ton_ of options and that they’re a bit of a special case as they can be configured in 2/3 of the config files and the inventory file. 

I do think splitting up config options by category puts us back in the same situation that this ticket is trying to fix, namely that we list a bunch of options on a page and then say that they can be set in _these_ files but not _those_ files. I think that paired with the changes to ‘Configuring Bolt’ it makes more sense to have reference pages for individual files, as we are explicitly telling users which files they should be using and when, as well as discouraging use of bolt.yaml altogether (I think we have somewhere around 7 deprecation notes for it?). With users being told to primarily use bolt-project.yaml and inventory.yaml, it makes sense to have a single source for them to reference to see what they can configure, instead of having to jump across all the different categories to see what’s available in that file.

lib/bolt/config.rb
```diff
@@ -39,114 +39,320 @@ class Config
...
+    #   :def     The **documented** default value. This is the value that is displayed
+    #            in the reference docs and is not used by Bolt to actually set a default
+    #            value.
+    #   :desc    The text description of the option that is displayed in documentation.
+    #   :exmp    An example value for the option. This is used to generate an example
+    #            configuration file in the reference docs.
+    #   :type    The option's expected type. If an option accepts multiple types, this is
+    #            an array of the accepted types. Any options that accept a Boolean value
+    #            should use the TrueClass type.
+    OPTIONS = {
```
lucywyman, line 54: This is awesome! The examples part is especially great. Organizationally, it might be nice if it was in a `Bolt::Config::Options` module so that it didn't take up so much real estate in the config class? Similar to the `Bolt::Shell::Powershell::Snippets` module.

beechtom: 👍 I like that idea.

beechtom:
Preview site is updated with changes to 'Configuring Bolt'.

documentation/configuring_bolt.md
```diff
@@ -1,8 +1,41 @@
...
-You can configure Bolt's options and features at a project level, a user level,
-or a system-wide level. Unless your use case requires setting user-specific or
-system-wide configurations, configure Bolt at the project level. 
+Bolt has many options and features that can be configured to suit your
+project's needs. In general, Bolt configuration falls into four categories:
+
+- **Bolt behavior**
+
+  Configure how Bolt itself runs, such as the format to use when displaying
+  output or how many threads to use when connecting to targets.
```
hestonhoffman, line 9:
```suggestion
- **Bolt behavior:** Configure how Bolt itself runs, such as the format to use when 
  displaying output or how many threads to use when connecting to targets.
```

hestonhoffman: Can we collapse the rest of the bulleted list like this? Mostly a style thing, but I think it'll look better in the output too.

documentation/configuring_bolt.md
```diff
@@ -1,8 +1,41 @@
...
+
+- **Projects**
+
+  Configure the Bolt project that you are running in, such as paths to the
```
hestonhoffman, line 13:
```suggestion
  Configure the Bolt project that you are running Bolt in, such as paths to the
```

documentation/configuring_bolt.md
```diff
@@ -1,8 +1,41 @@
...
+  inventory file or a Hiera configuration file.
```
hestonhoffman, line 14:
```suggestion
  inventory file or the path to a Hiera configuration file.
```

documentation/configuring_bolt.md
```diff
@@ -1,8 +1,41 @@
...
+  Group and configure the targets that you connect to and run commands on with
+  Bolt.
+
+Each type of configuration can be set in a different file. You can configure
+Bolt's options and features at a project level, a user level, or a system-wide
+level. Unless your use case requires setting user-specific or system-wide
+configurations, configure Bolt at the project level.
+
+| Type of Configuration | [`inventory.yaml`](bolt_inventory_reference.md) | [`bolt-project.yaml`](bolt_project_reference.md) | [`bolt-defaults.yaml`](bolt_defaults_reference.md) |
+| --- | :-: | :-: | :-: |
```
hestonhoffman, line 33: Good move leaving bolt.yaml out of the table. I think the less we highlight it, the better.

lib/bolt/config/transport/ssh.rb
```diff
@@ -12,132 +12,216 @@ class SSH < Base
...
+                  "non-deprecated algorithms are available by default when this option is not used. To "\
+                  "reference all default algorithms using this option, add 'defaults' to the list of "\
+                  "supported algorithms.",
+            exmp: %w[defaults diffie-hellman-group1-sha1]
+          },
+          "load-config" => {
+            type: TrueClass,
+            desc: "Whether to load system SSH configuration from '~/.ssh/config' and '/etc/ssh_config'.",
+            def:  true,
+            exmp: false
```
lucywyman, line 113: Tom gets team MVP status for updating *all* of these.

lib/bolt/config/options.rb
```diff
@@ -0,0 +1,320 @@
+# frozen_string_literal: true
+
+module Bolt
+  class Config
+    module Options
```
lucywyman, line 5: 🙌

Rakefile
```diff
@@ -99,57 +120,62 @@ namespace :ci do
 end
 
 namespace :docs do
-  desc "Generate markdown docs for Bolt's transport configuration options"
+  desc 'Generate markdown docs for bolt.yaml'
```
lucywyman, line 123: These feels like the ultimate meta-programming challenge....but I think having it be verbose is more clear and less brittle.

lib/bolt/config/options.rb
```diff
@@ -0,0 +1,320 @@
...
+                "value configuring the logfile output.",
+          type: Hash,
+          exmp: { "console" => { "level" => "info" },
+                  "~/logs/debug.log" => { "append" => false, "level" => "debug" } }
+        },
+        "modulepath" => {
+          desc: "An array of directories that Bolt loads content such as plans and tasks from. Read more "\
+                "about modules in [Module structure](module_structure.md).",
+          type: [Array, String],
+          def:  ["boltdir/modules", "boltdir/site-modules", "boltdir/site"],
```
lucywyman, line 81: I don't feel particularly strongly, but should we change this to project?

beechtom: I actually had that before I opened the PR. I was convinced in review we would want to change it to boltdir instead haha. I'll update.

lucywyman: Heh, damned if you do damned if you don't. @hestonhoffman Might have a more informed opinion.

lib/bolt/config/options.rb
beechtom: 🎉

Rakefile
```diff
@@ -99,57 +123,64 @@ namespace :ci do
...
+  desc 'Generate markdown docs for transports configuration reference'
   task :transports_reference do
-    @opts       = {}
-    @transports = { options: {}, defaults: {} }
+    @opts = stringify_types(Bolt::Config::INVENTORY_OPTIONS)
 
-    @opts[:options]  = Bolt::Config::INVENTORY_CONFIG
-    @opts[:defaults] = Bolt::Config::DEFAULT_OPTIONS
+    @subopts = Bolt::Config::TRANSPORT_CONFIG.sort.each_with_object({}) do |(name, transport), acc|
+      # Drop example from opts so we can generate a full YAML file example
```
lucywyman, line 171:
```suggestion
      # Only include suboption examples in the full example, not under individual suboption sections
```

lib/bolt/config/transport/base.rb
```diff
@@ -116,24 +119,21 @@ def self.options
...
-                      "#{opt} must be a Boolean true or false, received #{val.class} #{val.inspect}"
-              end
-            else
-              unless val.nil? || val.is_a?(type)
-                raise Bolt::ValidationError,
-                      "#{opt} must be a #{type}, received #{val.class} #{val.inspect}"
-              end
+            types = Array(TRANSPORT_OPTIONS.dig(opt, :type)).compact
+
+            next if val.nil? || types.empty? || types.include?(val.class)
```
lucywyman, line 127: It seems weird that we don't err when a value is nil? I see that's the same as the old behavior but I'm not sure what a valid use case is for having a config value be nil...

lib/bolt/config/transport/base.rb
```diff
@@ -116,24 +119,21 @@ def self.options
           assert_type
         end
 
-        # Validates that each option is the correct type. Types are loaded from the OPTIONS hash.
+        # Validates that each option is the correct type. Types are loaded from the TRANSPORT_OPTIONS hash.
         private def assert_type
           @config.each_pair do |opt, val|
```
lucywyman, line 124: This automated validation is so nice, it'd be nice to use it for general Bolt config as well (in another ticket!)

beechtom: That's the plan! :D

lib/bolt/config/transport/base.rb
beechtom: Ya, I really regret not commenting that when I wrote the class... there's a very particular edge case that it covers, but I can't remember what it was.

lib/bolt/config/transport/local.rb
```diff
@@ -7,51 +7,21 @@ module Bolt
...
+          cleanup
+          interpreters
+          run-as
+          run-as-command
+          sudo-executable
+          sudo-password
+          tmpdir
+        ].freeze
+
+        WINDOWS_OPTIONS = %w[
```
lucywyman, line 20: It seems useful to document this distinction as well? I think having `*nix` and `Windows` subheadings under `local` would make sense, or a `Windows: true` attribute for the ones that are supported (since it's a strict subset)

PR #1924, opened 2020-06-22
(maint) Update puppetserver-standalone to puppetserver in CI

beechtom:
This updates the Docker image used in CI tests from
`puppetserver-standalone` to `puppetsever`, as `puppetserver-standalone`
no longer exists.

!no-release-note

### DIFF
--- a/spec/Dockerfile.puppetserver
+++ b/spec/Dockerfile.puppetserver
@@ -1,8 +1,9 @@
-FROM puppet/puppetserver-standalone
+FROM puppet/puppetserver
 
 ARG hostname="boltserver"
 ENV PUPPETSERVER_HOSTNAME "$hostname"
-
+ENV PUPPET_STORECONFIGS false
+ENV PUPPET_REPORTS log
 
 # Use our own certs and disable the CA
 COPY fixtures/ssl/ca.pem /etc/puppetlabs/puppet/ssl/certs/ca.pem


spec/Dockerfile.puppetserver
```diff
@@ -1,8 +1,9 @@
-FROM puppet/puppetserver-standalone
+FROM puppet/puppetserver
 
 ARG hostname="boltserver"
 ENV PUPPETSERVER_HOSTNAME "$hostname"
-
+ENV PUPPET_STORCONFIGS false
+ENV PUPPET_REPORTS log
```
beechtom, line 6: See https://github.com/puppetlabs/puppetserver/commit/90893b09d1464519abd630bf3a58656ac0819ec5

spec/Dockerfile.puppetserver
```diff
@@ -1,8 +1,9 @@
-FROM puppet/puppetserver-standalone
+FROM puppet/puppetserver
 
 ARG hostname="boltserver"
 ENV PUPPETSERVER_HOSTNAME "$hostname"
-
+ENV PUPPET_STORCONFIGS false
```
lucywyman, line 5:
```suggestion
ENV PUPPET_STORECONFIGS false
```

lucywyman: It actually doesn't look like puppetserver honors either of these settings? https://puppet.com/docs/puppetserver/latest/configuration.html

nicklewis: The page is kind of misleading. I think it refers to the settings which affect the behavior of the clojure code in puppetserver. But puppetserver also runs puppet in jruby instances and those *do* honor the rest of puppet.conf.

beechtom: Updated with the correct env var name.

lucywyman: oic on the pr. :+1:

PR #1925, opened 2020-06-22
(GH-1904) Localhost default config overrides group-level config

### LINKED ISSUE #1904: 'localhost' target doesn't pickup special config if `transport: ssh` is at top level of inventory
## Describe the Bug
I haven't verified this yet but I wanted to make a note. Reid mentioned that the special target 'localhost' wasn't using Bolt's ruby interpreter, and thinks it's due to `transport: ssh` being at the top level of inventory. Per [the original issue](https://tickets.puppetlabs.com/browse/BOLT-1205) 'localhost' should pick up its special config regardless of transport.

## Solution

- Magic localhost config should be treated as target-level config that can't be overridden by the group, but can be overridden by a localhost target definition in the inventory
- Do merging in `target_collect`
- Re-evaluate whether localhost should be part of `all` for Bolt 3.x

lucywyman:
Previously, the default config for 'localhost' was merged as part of the
group data collected for that target, making it effectively group-level
data. Because all targets are considered part of the 'all' group
top-level config would override localhost's default config, which made
it difficult to set defaults for all targets except localhost. This
modifies how we load localhost defaults to now merge them as part of
initializing the `Bolt::Inventory::Target` object, meaning that any time
a target with name 'localhost' is created (either via the inventory or
by `Target.new` in a plan) it will pick up the localhost defaults at the
target level. This means that now the only way to override the defaults
is to set that config at the target-level.

Closes #1904

!bug

* **Consider localhost default config target-level instead of group-level** (

  Previously the 'localhost' special default config was merged at the
  group-level, meaning that group-level config in the inventory would
  override it. The config is now target-level, and must be overridden at
  the target-level in inventory.

### DIFF
--- a/lib/bolt/inventory/inventory.rb
+++ b/lib/bolt/inventory/inventory.rb
@@ -78,20 +78,6 @@ module Bolt
         target_array.first
       end
 
-      def self.localhost_defaults(data)
-        defaults = {
-          'config' => {
-            'transport' => 'local',
-            'local' => { 'interpreters' => { '.rb' => RbConfig.ruby } }
-          },
-          'features' => ['puppet-agent']
-        }
-        data = Bolt::Util.deep_merge(defaults, data)
-        # If features is an empty array deep_merge won't add the puppet-agent
-        data['features'] += ['puppet-agent'] if data['features'].empty?
-        data
-      end
-
       #### PRIVATE ####
       def group_data_for(target_name)
         @groups.group_collect(target_name)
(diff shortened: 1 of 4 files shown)


nicklewis:
Looks good! Commit message needs the actual link to the issue and there's the failing test, but otherwise 👍

nicklewis:
Integration test failing in a somewhat similar way. Except this time we're _incidentally_ overriding `localhost` instead of intentionally (since it just happens to be the hostname of the SSH target)

lucywyman:
Yeah, I'm actually not quite sure why that's failing...the plugin config should override the group-level config?

nicklewis:
The plugin config doesn't set transport though. So I think it executes with the default local config instead of the modified SSH config. Maybe just the plugin needs to return `transport: 'ssh'` in the config in `target_mapping`?

Issue #1934, opened 2020-06-24
Allow selective sync of plugins to target nodes


hajee:
## Use Case

When using a puppet apply block in a plan, bolt sync's all of the plugin files in all of the modules over to the target node. When the Bolt projects where small and the execution time of the puppet code was long, this was no issue. Since the integration with Puppet Enterprise, it is recommended that the Bolt dir is integrated with the control repo. In general control repo's for larger environments are quite large.

If we would like a plan to contain a very small fragment of puppet code, currently all plugins from all modules are synced to the targets. This is very time and resource consuming.

## Describe the Solution You Would Like

It would be great if the plan author can specify what modules need to be synced, and thereby controlling the amount of plugin data to be synced to the target.

lucywyman:
@hajee Are you wanting to use this with PE? Or just with Bolt?

hajee:
I think this can be useful for both. Right now we use bolt we might switch to the more integrated variant in PE.  Seeing the use case it is more beneficial for PE, but I can imagine the same issues happening using just bolt and having a large set of modules.

lucywyman:
Closed by #1935

PR #1935, opened 2020-06-24
(GH-1934) Add support for selective sync of plugins to target nodes

hajee:
Implement feature request #1934

### DIFF
--- a/bolt-modules/boltlib/lib/puppet/functions/apply_prep.rb
+++ b/bolt-modules/boltlib/lib/puppet/functions/apply_prep.rb
@@ -13,10 +13,13 @@ require 'bolt/task'
 # > **Note:** Not available in apply block
 Puppet::Functions.create_function(:apply_prep) do
   # @param targets A pattern or array of patterns identifying a set of targets.
+  # @param options Options hash.
+  # @option options [Array] _required_modules An array of modules to sync to the target.
   # @example Prepare targets by name.
   #   apply_prep('target1,target2')
   dispatch :apply_prep do
     param 'Boltlib::TargetSpec', :targets
+    optional_param 'Hash[String, Data]', :options
   end
 
   def script_compiler
@@ -60,17 +63,33 @@ Puppet::Functions.create_function(:apply_prep) do
     @executor ||= Puppet.lookup(:bolt_executor)
   end
 
-  def apply_prep(target_spec)
+  def apply_prep(target_spec, options = {})
     unless Puppet[:tasks]
       raise Puppet::ParseErrorWithIssue
         .from_issue_and_stack(Bolt::PAL::Issues::PLAN_OPERATION_NOT_SUPPORTED_WHEN_COMPILING, action: 'apply_prep')
     end
+
+    options = options.transform_keys { |k| k.sub(/^_/, '').to_sym }
 
     applicator = Puppet.lookup(:apply_executor)
 
     executor.report_function_call(self.class.name)
 
     targets = inventory.get_targets(target_spec)
+
+    required_modules = options[:required_modules].nil? ? nil : Array(options[:required_modules])
+    if required_modules&.any?
+      Puppet.debug("Syncing only required modules: #{required_modules.join(',')}.")
+    end
+
+    # Gather facts, including custom facts
+    plugins = applicator.build_plugin_tarball do |mod|
+      next unless required_modules.nil? || required_modules.include?(mod.name)
+      search_dirs = []
+      search_dirs << mod.plugins if mod.plugins?
+      search_dirs << mod.pluginfacts if mod.pluginfacts?
+      search_dirs
+    end
 
     executor.log_action('install puppet and gather facts', targets) do
       executor.without_default_logging do
@@ -107,14 +126,6 @@ Puppet::Functions.create_function(:apply_prep) do
           raise Bolt::RunFailure.new(set.error_set, 'apply_prep') unless set.ok
 
           need_install_targets.each { |target| set_agent_feature(target) }
-        end
-
-        # Gather facts, including custom facts
-        plugins = applicator.build_plugin_tarball do |mod|
-          search_dirs = []
-          search_dirs << mod.plugins if mod.plugins?
-          search_dirs << mod.pluginfacts if mod.pluginfacts?
-          search_dirs
         end
 
         task = applicator.custom_facts_task
--- a/lib/bolt/applicator.rb
+++ b/lib/bolt/applicator.rb
@@ -18,7 +18,6 @@ module Bolt
                    pdb_client, hiera_config, max_compiles, apply_settings)
       # lazy-load expensive gem code
       require 'concurrent'
-
       @inventory = inventory
       @executor = executor
       @modulepath = modulepath || []
@@ -30,17 +29,6 @@ module Bolt
 
       @pool = Concurrent::ThreadPoolExecutor.new(max_threads: max_compiles)
       @logger = Logging.logger[self]
-      @plugin_tarball = Concurrent::Delay.new do
-        build_plugin_tarball do |mod|
-          search_dirs = []
-          search_dirs << mod.plugins if mod.plugins?
-          search_dirs << mod.pluginfacts if mod.pluginfacts?
-          search_dirs << mod.files if mod.files?
-          type_files = "#{mod.path}/types"
-          search_dirs << type_files if File.exist?(type_files)
-          search_dirs
-        end
-      end
     end
 
     private def libexec
@@ -188,7 +176,6 @@ module Bolt
 
     def apply_ast(raw_ast, targets, options, plan_vars = {})
       ast = Puppet::Pops::Serialization::ToDataConverter.convert(raw_ast, rich_data: true, symbol_to_string: true)
-
       # Serialize as pcore for *Result* objects
       plan_vars = Puppet::Pops::Serialization::ToDataConverter.convert(plan_vars,
                                                                        rich_data: true,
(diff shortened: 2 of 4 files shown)


adreyer:
I don't think there is a good argument for allowing this to be set per target. It makes more sense as a global setting that is used for all targets.

hajee:
In our current use-case, we have multiple nodes with different roles and we run different bolt tasks on them. Each bolt task has a specific set of modules it needs. 

Setting it globally means all nodes would get the same set of modules. Although that would probably help a bit, this implementation gives you maximum control.

lib/bolt/applicator.rb
```diff
@@ -30,23 +29,16 @@ def initialize(inventory, executor, modulepath, plugin_dirs, project,
...
-        end
-      end
     end
 
     private def libexec
       @libexec ||= File.join(Gem::Specification.find_by_name('bolt').gem_dir, 'libexec')
     end
 
+    def selected_module?(mod)
+      @selected_modules.any? { |m| mod =~ /#{m}/ }
```
lucywyman, line 39:
```suggestion
      @selected_modules.include?(mod)
```

This will more accurately select the exact module, not ones that include the modulename. For example if I have module `nginx` and call `selected_module?('nginx-config')` this will return true.

hajee: The idea was to have an array of regular expressions, to make it easy to select multiple modules. So I put it there explicitly. For example, the augueas provides can be added like this:

```
['augeasproviders_.*']
```

Personally I prefer this above specifying all modules individually.

adreyer:
I'm concerned that scoping this to nodes and trying to track exactly what is managed on each node is going to quickly become unmaintainable for many users and introduces new performance issues on the bolt controller around building multiple plugin bundles. I this it works tolerably well in the case where you are applying roles to servers and can try to set which modules are needed for each role. But you're still left tracking the role assignment and module code separately from the plugin bundle mapping . What happens when a user wants to run a plan that goes outside of these bounds if a fact is simply not synced things may behave unexpectedly.

hajee:
@adreyer I understand your concern regarding maintainability. On the other hand, I guess that goes for all features. Always use it moderate and wise. Although you could add this to the inventory, I think it is best used in the plan by calling 

```
$target.set_config('selective_sync', ['stdlib'])
```

just before applying the puppet block. This way it keeps the selection close to the puppet code and keeps everything as neat as possible. Missing a module would be the same as missing a module in the Puppetfile. When the `set_config` is close to the puppet code, probably easy to spot and to fix.

I see this feature a bit like setting the `puppet-agent` feature. You allow the author of the plan, control if he/she wants to install an agent. If he/she tells bolt there is an agent and it is not there, you get an error.

I have been thinking about using a variable, but because it is functionally so close to a feature, I thought it best to use a config setting. 

Regarding performance: I'm not sure what the impact is on large sets of nodes. If this is an issue, we could always optimize this.

adreyer:
I was imagining this being set in the inventory file. If the plan author is expected to control this it seems much more natural as a parameter to `apply` and `apply_prep` themselves. Otherwise the plan author has to worry about aligning with preexisting global inventory state and cleaning up to avoid polluting it.

hajee:
Functionality that sounds like a great idea! I'll have a look at how we can implement this.

adreyer:
I think it's best if this feature starts off fairly simple and with the understanding that as a performance optimization it may be ignored/deprecated later.  For that reason I think we should name it something like "required_modules" and have it be an explicit list of modules. While a regex may be convienent it seems primarily to be be an issue for augeus providers.

`apply($targets, 'required_modules' => ['module_1', 'module_2']) { .... }`

Limiting this further by target may be a further optimization to this but there isn't really a good syntax for that in the plan language. We have per target parameters for `run_task` now but that won't translate to `apply` because it would require a second block. This seems like it's primarily a problem when you are applying a largeish but largely unique catalog on each node. In that case the cost of syncing module code is much less important.

hajee:
@adreyer @lucywyman The latest force commit contains the implementation as discussed

bolt-modules/boltlib/lib/puppet/functions/apply_prep.rb
```diff
@@ -111,6 +120,7 @@ def apply_prep(target_spec)
 
         # Gather facts, including custom facts
         plugins = applicator.build_plugin_tarball do |mod|
+          next if @required_modules.any? && !@required_modules.include?(mod)
```
beechtom, line 123:
```suggestion
          next if @required_modules.any? && !@required_modules.include?(mod.name)
```

lib/bolt/applicator.rb
```diff
@@ -272,8 +277,8 @@ def apply_ast(raw_ast, targets, options, plan_vars = {})
       r
     end
 
-    def plugins
-      @plugin_tarball.value ||
+    def plugins(for_target_node)
+      for_target_node.value ||
```
beechtom, line 281:
```suggestion
    def plugins(plugin_tarball)
      plugin_tarball.value ||
```

`for_target_node` seems ambiguous / misleading on what the parameter actually is.

bolt-modules/boltlib/lib/puppet/functions/apply_prep.rb
```diff
@@ -60,18 +62,25 @@ def executor
...
 
+    options = options.transform_keys { |k| k.sub(/^_/, '').to_sym }
+
     applicator = Puppet.lookup(:apply_executor)
 
     executor.report_function_call(self.class.name)
 
     targets = inventory.get_targets(target_spec)
 
+    @required_modules = Array(options[:required_modules])
```
lucywyman, line 79: This means that setting `apply_prep($targets, 'required_modules' => [])` will behave the same as `apply_prep($targets)`, rather than excluding all modules (which is probably what people will expect). This would make required_modules nil if the option isn't set, and convert whatever is passed to an array.

```suggestion
    @required_modules = options[:required_modules].nil? ? nil : Array(options[:required_modules])
```

bolt-modules/boltlib/lib/puppet/functions/apply_prep.rb
```diff
@@ -60,18 +62,25 @@ def executor
...
+    if @required_modules.any?
```
lucywyman, line 80:
```suggestion
    if @required_modules&.any?
```
With the changes above this may be nil, in which case the if will still correctly evaluate to false.

bolt-modules/boltlib/lib/puppet/functions/apply_prep.rb
```diff
@@ -111,6 +120,7 @@ def apply_prep(target_spec)
 
         # Gather facts, including custom facts
         plugins = applicator.build_plugin_tarball do |mod|
+          next if @required_modules.any? && !@required_modules.include?(mod.name)
```
lucywyman, line 123:
```suggestion
          next unless @required_modules.nil? || @required_modules.include?(mod.name)
```

bolt-modules/boltlib/lib/puppet/functions/apply_prep.rb
```diff
@@ -13,10 +13,12 @@
 # > **Note:** Not available in apply block
 Puppet::Functions.create_function(:apply_prep) do
   # @param targets A pattern or array of patterns identifying a set of targets.
+  # @param options Options hash. For now only supports `required_modules` as an array of modules to sync
```
beechtom, line 16:
```suggestion
  # @param options Options hash.
  # @option options [Array] _required_modules An array of modules to sync to the target.
```

This will let puppet-strings generate documentation similar to other Bolt functions with options hashes.

Other than this and the change Lucy mentioned above, this looks ready to go.

lib/bolt/applicator.rb
```diff
@@ -206,9 +193,13 @@ def apply_ast(raw_ast, targets, options, plan_vars = {})
         # This data isn't available on the target config hash
         config: @inventory.transport_data_get
       }
-
       description = options[:description] || 'apply catalog'
 
+      @required_modules = Array(options[:required_modules])
```
lucywyman, line 198:
```suggestion
      @required_modules = options[:required_modules].nil? ? nil : Array(options[:required_modules])
```
Same changes here

lib/bolt/applicator.rb
```diff
@@ -206,9 +193,13 @@ def apply_ast(raw_ast, targets, options, plan_vars = {})
...
+      if @required_modules.any?
```
lucywyman, line 199:
```suggestion
      if @required_modules&.any?
```

lib/bolt/applicator.rb
```diff
@@ -272,8 +277,8 @@ def apply_ast(raw_ast, targets, options, plan_vars = {})
       r
     end
 
-    def plugins
-      @plugin_tarball.value ||
+    def plugins(plugin_tarball)
+      plugin_tarball.value ||
         raise(Bolt::Error.new("Failed to pack module plugins: #{@plugin_tarball.reason}", 'bolt/plugin-error'))
```
lucywyman, line 282:
```suggestion
        raise(Bolt::Error.new("Failed to pack module plugins: #{plugin_tarball.reason}", 'bolt/plugin-error'))
```

lib/bolt/applicator.rb
```diff
@@ -235,9 +226,23 @@ def apply_ast(raw_ast, targets, options, plan_vars = {})
                   result
                 end
               else
+
+                plugin_tarball = Concurrent::Delay.new do
+                  build_plugin_tarball do |mod|
+                    next if @required_modules.any? && !@required_modules.include?(mod.name)
```
lucywyman, line 232:
```suggestion
                    next unless @required_modules.nil? || @required_modules.include?(mod.name)
```

lib/bolt/applicator.rb
```diff
@@ -235,9 +226,23 @@ def apply_ast(raw_ast, targets, options, plan_vars = {})
                   result
                 end
               else
+
+                plugin_tarball = Concurrent::Delay.new do
```
nicklewis, line 230: Will this be built once for each target because it's inside the `queue_execute` method?

hajee: Yes it was. I have changed it so it is calculated before the target loop.

lucywyman:
I think this is just about ready. My only comment is that in order for this to show up in our changelog, your commit needs to have a message after the `!feature` flag:

```
!feature

* **Add selective plugin sync for apply and apply_prep** ([1934](https://github.com/puppetlabs/bolt/issues/1934))
  A new option `_required_modules` for `apply` and `apply_prep` functions will limit which module artifacts are uploaded to each target. This is useful for improving performance when applying manifests that don't require custom content from modules in environments that contain a lot of modules.
```

hajee:
@lucywyman I can squash all commits and change the text. Is that what you want, or do you want to retain the editing history in the PR?

lucywyman:
Squash and edit would be great! I'm realizing we should update our contributing document with all this info - thanks for bearing with us!

hajee:
@nicklewis I force pushed these changes

PR #1940, opened 2020-06-24
(GH-1931) Add '--project' as an alias for '--boltdir'

### LINKED ISSUE #1931: Add --project alias for --boltdir CLI option
## Use Case
Since implementing Bolt projects as a feature, we have been updating the docs to favor the term 'project directory' over 'boltdir'. Both [Getting started with Bolt](https://puppet.com/docs/bolt/latest/getting_started_with_bolt.html) and [Project directories](https://puppet.com/docs/bolt/latest/bolt_project_directories.html) use 'project directory' over 'boltdir'. There is also potential for confusion between the concept of a boltdir and the `Boltdir` directory.

Since our docs have begun phasing out use of the term 'boltdir', we should move towards removing the `--boltdir` CLI option in favor of a `--project` option.

Add `--project` as an alias for the `--boltdir` CLI option.

## Other considerations?
- Should the `--boltdir` option be deprecated / removed in a future version of Bolt (i.e. 3.x)? We should not deprecate this immediately, but we should deprecate it closer to 3.x.

lucywyman:
This adds a `--project` CLI flag as an alias for `--boltdir`. If both
options are specified on the command line the rightmost option will be
used. `--project` behaves exactly like `--boltdir`. It also updates
documentation to use `--project` in place of `--boltdir`, since we
consistently refer to the concept as a 'project' in the documentation
and don't often use 'Boltdir'.

This additionally replaces `Boltdir` with `project` for all Bolt CLI flag descriptions.

Closes #1931

!feature

* **Add --project as an alias for the --boltdir CLI flag** ([1931](https://github.com/puppetlabs/bolt/issues/1931))

  New CLI flag `--project` can be used in place of `--boltdir`.

### DIFF
--- a/lib/bolt/bolt_option_parser.rb
+++ b/lib/bolt/bolt_option_parser.rb
@@ -10,7 +10,7 @@ module Bolt
                 authentication: %w[user password password-prompt private-key host-key-check ssl ssl-verify],
                 escalation: %w[run-as sudo-password sudo-password-prompt sudo-executable],
                 run_context: %w[concurrency inventoryfile save-rerun cleanup],
-                global_config_setters: %w[modulepath boltdir configfile],
+                global_config_setters: %w[modulepath project configfile],
                 transports: %w[transport connect-timeout tty ssh-command copy-command],
                 display: %w[format color verbose trace],
                 global: %w[help version debug] }.freeze
@@ -422,7 +422,7 @@ module Bolt
 
       ACTIONS
           generate-types        Generate type references to register in plans
-          install               Install modules from a Puppetfile into a Boltdir
+          install               Install modules from a Puppetfile into a project
           show-modules          List modules available to the Bolt project
     HELP
 
@@ -445,7 +445,7 @@ module Bolt
           bolt puppetfile install [options]
 
       DESCRIPTION
-          Install modules from a Puppetfile into a Boltdir
+          Install modules from a Puppetfile into a project
     HELP
 
     PUPPETFILE_SHOWMODULES_HELP = <<~HELP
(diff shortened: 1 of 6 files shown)


hestonhoffman:
Woah, that was quick!

lucywyman:
Haha I knew it'd be a small ticket :ticket:

lib/bolt/bolt_option_parser.rb
```diff
@@ -709,7 +709,7 @@ def initialize(options)
           File.expand_path(moduledir)
         end
       end
-      define('--boltdir FILEPATH',
+      define('--boltdir FILEPATH', '--project FILEPATH',
```
beechtom, line 712: If we're updating the documentation to reference `--project` over `--boltdir`, then I think `--project` should come first. Otherwise, the command reference doc does not show `--project`.

If we put `--project` first then it will need to be added to the `OPTIONS[:global_config_setters]` constant.

lib/bolt/bolt_option_parser.rb
```diff
@@ -709,7 +709,7 @@ def initialize(options)
...
              'Specify what Boltdir to load config from (default: autodiscovered from current working dir)') do |path|
```
beechtom, line 713:
```suggestion
             'Specify the project directory to load config from (default: autodiscovered from current working dir)') do |path|
```

lib/bolt/bolt_option_parser.rb
lucywyman: https://puppet.com/docs/bolt/latest/bolt_command_reference.html I think it will show both, but putting it first can't hurt.

PR #1942, opened 2020-06-25
(maint) Remove 'whitelist' from schemas

lucywyman:
This removes the term 'whitelist' from our json schemas.

!no-release-note

Realized too late I missed this on Nick's PR

### DIFF
--- a/schemas/bolt-project.schema.json
+++ b/schemas/bolt-project.schema.json
@@ -88,7 +88,7 @@
       "pattern": "^([a-zA-Z0-9]+[-])?[a-z][a-z0-9_]*$"
     },
     "plans": {
-      "description": "The list of whitelisted plans.",
+      "description": "A list of plan names to show in 'bolt plan show' output, if they exist. This option is used to limit the visibility of plans for users of the project. For example, project authors might want to limit the visibility of plans that are bundled with Bolt or plans that should only be run as part of another plan. When this option is not configured, all plans are visible. This option does not prevent users from running plans that are not part of this list.",
       "type": "array",
       "items": {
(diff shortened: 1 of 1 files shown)


schemas/bolt-project.schema.json
```diff
@@ -88,7 +88,7 @@
       "pattern": "^([a-zA-Z0-9]+[-])?[a-z][a-z0-9_]*$"
     },
     "plans": {
-      "description": "The list of whitelisted plans.",
+      "description": "The list of plans to show if they are available.",
```
beechtom, line 91:
```suggestion
      "description": "A list of plan names to show in 'bolt plan show' output, if they exist. This option is used to limit the visibility of plans for users of the project. For example, project authors might want to limit the visibility of plans that are bundled with Bolt or plans that should only be run as part of another plan. When this option is not configured, all plans are visible. This option does not prevent users from running plans that are not part of this list.",
```

schemas/bolt-project.schema.json
```diff
@@ -176,7 +176,7 @@
       "type": "boolean"
     },
     "tasks": {
-      "description": "The list of whitelisted tasks.",
+      "description": "The list of tasks to show if they are available.",
```
beechtom, line 179:
```suggestion
      "description": "A list of task names to show in 'bolt task show' output, if they exist. This option is used to limit the visibility of tasks for users of the project. For example, project authors might want to limit the visibility of tasks that are bundled with Bolt or plans that should only be run as part of a larger workflow. When this option is not configured, all tasks are visible. This option does not prevent users from running tasks that are not part of this list.",
```

Issue #1899, opened 2020-06-15
Ability to set environment variables for run_* functions


nmaludy:
## Use Case

As a bolt user i have a command/script that requires a password passed in on the command line. 

Example:
```shell
mongodump -u mongouser -p 'Abc123$' --gzip --archive=mongodump-$(date  --utc --iso-8601=seconds).gzip.archive
```

Ideally instead of embedding that password in the `command` or `arguments` string, i would like to set these as environment variables:

```shell
# set these secretly
export MONGO_PASSWORD='Abc123$'

# run the dump command
mongodump -u mongouser -p "${MONGO_PASSWORD}" --gzip --archive=mongodump-$(date  --utc --iso-8601=seconds).gzip.archive
```

Additionally, some commands that require passwords do not accept them on the CLI, for security reasons, and require the arguments be environment variables.

Example:
```shell
# these would be set secretly
export PGUSER='dbuser'
export PGPASSWORD='Abc123$'

# run the dump command
pg_dumpall --compress=9 -f pgdump-$(date  --utc --iso-8601=seconds).sql.gz
```

## Describe the Solution You Would Like
I would like the ability to set environment variables when calling the `run_command()` and `run_script()` functions (not sure if it makes sense for `run_task()`?). 

Example:
```puppet
run_command('mongodump -u mongouser -p "${MONGO_PASSWORD}" --gzip --archive=mongodump-$(date  --utc --iso-8601=seconds).gzip.archive',
            $targets,
            _environment => {
              MONGO_PASSWORD => $mongodb_password,
            })
```

Or the postgres example:
```puppet
run_command('pg_dumpall --compress=9 -f pgdump-$(date  --utc --iso-8601=seconds).sql.gz',
            $targets,
            _environment => {
              PGPASSWORD => $postgres_password,
              PGUSER => $postgres_username,
            })
```

### Additional Context

Maybe this deserves a different request, but also thinking about other "shell" things that people might want to set and `cwd` / `pwd` comes up ie. "what directory the command/script will be executed in". Example:

```
run_command('pg_dumpall --compress=9 -f pgdump-$(date  --utc --iso-8601=seconds).sql.gz',
            $targets,
            _env => {
              PGPASSWORD => $postgres_password,
              PGUSER => $postgres_username,
            },
            _cwd => '/mnt/backups')
```

beechtom:
- We should not do this for `run_task`, as tasks have a clear API and there are ways to do this already
[...]

PR #1943, opened 2020-06-25
(GH-1899) Accept env_vars for commands and scripts

nicklewis:
Previously, we had no way to specify environment variables when running
commands or scripts. There are many reasons a user might need to set
environment variables:

* To set up the execution context for a command
* To pass arguments to scripts that expect them as env vars
* To pass secrets to commands rather than passing them as args

We now support passing `_env_vars` to `run_command()` and `run_script()`
which will set the passed variables when running the command or script.

!feature

* **Set environment variables for commands and scripts** ([#1899](https://github.com/puppetlabs/bolt/issues/1899))

  The `run_command()` and `run_script()` plan functions now support an
  `_env_vars` argument which accepts a Hash of environment variable
  declarations to set when running the command/script.

### DIFF
--- a/bolt-modules/boltlib/lib/puppet/functions/run_command.rb
+++ b/bolt-modules/boltlib/lib/puppet/functions/run_command.rb
@@ -13,6 +13,7 @@ Puppet::Functions.create_function(:run_command) do
   # @param options A hash of additional options.
   # @option options [Boolean] _catch_errors Whether to catch raised errors.
   # @option options [String] _run_as User to run as using privilege escalation.
+  # @option options [Hash] _env_vars Map of environment variables to set
   # @return A list of results, one entry per target.
   # @example Run a command on targets
   #   run_command('hostname', $targets, '_catch_errors' => true)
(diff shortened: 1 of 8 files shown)


bolt-modules/boltlib/lib/puppet/functions/run_command.rb
```diff
@@ -13,6 +13,7 @@
   # @param options A hash of additional options.
   # @option options [Boolean] _catch_errors Whether to catch raised errors.
   # @option options [String] _run_as User to run as using privilege escalation.
+  # @option options [Hash] _env_vars Map of environment variables to set
```
beechtom, line 16: It looks like Puppet doesn't provide type validation for options and we don't provide it ourselves as part of the function. Is that something we'd want to do at some point?

nicklewis: We could probably do it by setting the type to a `Struct` with appropriately-typed keys. I wasn't sure what sort of validation to do beyond Hash though. Requiring the values be strings seems overly restrictive, since numbers and booleans seem common and should work fine.

beechtom: Ya, I'm not sure that it's all that important, especially since it doesn't break anything. 🤷

PR #1944, opened 2020-06-25
(GH-1894) Do not load projects from world-writable directories

### LINKED ISSUE #1894: Do not load config from world-writable directories
## Use Case
Bolt does not currently perform any permissions checks when loading config files. This present a security vulnerability, as a malicious user could (re)place a config file in a directory with world-writable permissions, altering how Bolt is configured and run. 

Before loading a config file, Bolt should check the permissions of the file's directory. If the directory is world-writable, Bolt should not load the config and issue a warning stating as such.

Some users may want to explicitly allow config to be loaded from world-writable directories. In these cases, Bolt should load a config file regardless of its permissions if an environment variable is set. This could be a single environment variable to enable loading all config files, or individual environment variables associated with each separate config file (system, user, project).

Do not add any restrictions and expect users to set their directory permissions appropriately for their security needs.

## Solution

- Bolt should not load anything from a world-writable project directory
- This behavior should be disabled with a `BOLT_PROJECT` environment variable
- If `BOLT_PROJECT` is set, use that directory as the Bolt project, regardless of which directory would normally be loaded
- `BOLT_PROJECT` takes precedence over the world writeable check
- Print a helpful error when a directory is world-writable

beechtom:
This updates the `Bolt::Project` class to raise an error if the
specified project directory is world-writable. This change is meant to
prevent malicious users from modifying project content that could alter
how Bolt runs and any secure information it might use.

Users who wish to override this behavior and force Bolt to load a
project from a world-writable directory can set the `BOLT_PROJECT`
environment variable to the project directory path.

Closes #1894 

!feature

* **Do not load projects from world-writable directories**
  ([#1894](https://github.com/puppetlabs/bolt/issues/1894))

  Bolt now raises an error when attempting to load a project from a
  world-writable directory. Users who wish to override this behavior and
  run a project from a world-writable directory should can set the
  `BOLT_PROJECT` environment variable to the project directory path.

### DIFF
--- a/documentation/bolt_project_directories.md
+++ b/documentation/bolt_project_directories.md
@@ -98,6 +98,8 @@ if you have a single set of Bolt code and data that you use across all projects.
 
 Bolt uses these methods, in order, to choose a project directory.
 
+1. **Environment variable:** You can specify a path to a project using the
+   `BOLT_PROJECT` environment variable.
 1. **Manually specified:** You can specify on the command line what directory
    Bolt to use with `--project <DIRECTORY_PATH>`. There is not an equivalent
    configuration setting because the project directory must be known in order to
@@ -108,6 +110,39 @@ Bolt uses these methods, in order, to choose a project directory.
 1. **User project directory:** If no directory is specified manually or found in
    a parent directory, the user project directory is used.
 
+## World-writable project directories
+
+On **Unix-like systems**, Bolt will not load a project from a world-writable
+directory by default, as loading from a world-writable directory presents a
+potential security risk. If you attempt to load a project from a
+world-writable directory, Bolt will not load any content and will raise an
+exception.
+
+If you wish to override this behavior and force Bolt to load a project from a
+world-writable directory, you can set the `BOLT_PROJECT` environment variable
+to the project directory path.
+
+For example, if you wanted to load a project named `my_project` from the
+world-writable directory at `~/project/`, you would set the `BOLT_PROJECT`
+environment variable as:
+
+```bash
+export BOLT_PROJECT='~/project/world_writable'
+```
+
+> **Note:** Exported environment variables expire at the end of the current
+> session. If you need a more permanent solution, add the `export` line to your
+> `~/.bashrc` or the relevant profile for the shell you're using.
+
+If you want to use a world-writable directory for a single Bolt execution, set the
+environment variable before the Bolt command:
+
+```bash
+BOLT_PROJECT='~/project/world_writable' bolt command run uptime -t target1
+```
+
+> **Note:** The `BOLT_PROJECT` environment variable takes precedence over the
+> `--configfile` CLI option. 
 
 ## Project directory structure
 
--- a/lib/bolt/cli.rb
+++ b/lib/bolt/cli.rb
@@ -111,7 +111,10 @@ module Bolt
 
       validate(options)
 
-      @config = if options[:configfile]
+      @config = if ENV['BOLT_PROJECT']
+                  project = Bolt::Project.create_project(ENV['BOLT_PROJECT'], 'environment')
+                  Bolt::Config.from_project(project, options)
+                elsif options[:configfile]
                   Bolt::Config.from_file(options[:configfile], options)
                 else
                   project = if options[:boltdir]
--- a/lib/bolt/project.rb
+++ b/lib/bolt/project.rb
@@ -31,6 +31,7 @@ module Bolt
     # hierarchy, falling back to the default if we reach the root.
     def self.find_boltdir(dir)
       dir = Pathname.new(dir)
+
       if (dir + BOLTDIR_NAME).directory?
         create_project(dir + BOLTDIR_NAME, 'embedded')
       elsif (dir + 'bolt.yaml').file? || (dir + 'bolt-project.yaml').file?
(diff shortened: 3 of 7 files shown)


documentation/bolt_project_directories.md
```diff
@@ -108,6 +108,27 @@ Bolt uses these methods, in order, to choose a project directory.
 1. **User project directory:** If no directory is specified manually or found in
    a parent directory, the user project directory is used.
 
+## World-writable project directories
+
+By default, Bolt will not load a project from a world-writable directory, as
+they present a potential security risk. If you attempt to load a project from
+a world-writable directory, Bolt will not load any content and will raise an
```
hestonhoffman, line 115:
```suggestion
By default, Bolt will not load a project from a world-writable directory, as
loading from a world-writable directory presents a potential security risk. 
If you attempt to load a project from a world-writable directory, Bolt will 
not load any content and will raise an
```

documentation/bolt_project_directories.md
```diff
@@ -108,6 +108,27 @@ Bolt uses these methods, in order, to choose a project directory.
...
+exception.
```
hestonhoffman, line 116:
```suggestion
By default, Bolt will not load a project from a world-writable directory, because
loading from a world-writable directory presents a potential security risk. 
If you attempt to load a project from a world-writable directory, Bolt will not load 
any content and will raise an exception.
```

documentation/bolt_project_directories.md
```diff
@@ -108,6 +108,27 @@ Bolt uses these methods, in order, to choose a project directory.
...
+
+If you wish to override this behavior and force Bolt to load a project from a
+world-writable directory, you can set the `BOLT_PROJECT` environment variable
+to the project directory path.
+
```
hestonhoffman, line 121:
```suggestion
to the project directory path.

For example, if you wanted to load a project named `my_project` from the world-writable directory
at `~/project/`, you would set the `BOLT_PROJECT` environment variable as:

```

lib/bolt/project.rb
```diff
@@ -44,13 +45,23 @@ def self.find_boltdir(dir)
 
     def self.create_project(path, type = 'option')
       fullpath = Pathname.new(path).expand_path
+
+      if ENV['BOLT_PROJECT'].nil? && fullpath.world_writable?
```
nicklewis, line 49: It doesn't seem quite right that we check for the env var here but we _use_ it elsewhere. Maybe we should pass `type='environment'` from the CLI and check _that_ here.

lib/bolt/project.rb
```diff
@@ -44,13 +45,23 @@ def self.find_boltdir(dir)
...
+        raise Bolt::Error.new(
+          "Project directory '#{fullpath}' is world-writable. Bolt will not load projects from "\
```
nicklewis, line 51: I think this needs to clarify a couple of a) that this isn't used for security reasons, b) exactly how to set `BOLT_PROJECT` to do what you want. Also "Bolt will not load projects from [...]" sounds like Bolt will _run_ but won't load the project, which isn't accurate. Maybe something like this:

```
"Project directory '#{fullpath}' is world-writable which poses a security risk. Set BOLT_PROJECT='${fullpath}' to force the use of this project directory."
```

documentation/bolt_project_directories.md
```diff
@@ -108,6 +108,27 @@ Bolt uses these methods, in order, to choose a project directory.
...
+```bash
+export BOLT_PROJECT='~/project/world_writable'
+```
+
```
hestonhoffman, line 125:
```suggestion

> **Note:** Exported environment variables expire at the end of the current session. If you need 
> a more permanent solution, add the `export` line to your `~/.bashrc` or the relevant profile for
> the shell you're using.

If you want to use a world-writable directory for a single Bolt execution, set the environment
variable before the Bolt command:   
```

lib/bolt/project.rb
beechtom: That makes wayyy more sense. 👍

beechtom:
@nicklewis It looks like `File.world_writable?` doesn't work as-is, but the [`win32/file` gem](https://github.com/chef/win32-file) redefines it so it works properly. Just ran it on Windows and confirmed that it works as expected.

lib/bolt/cli.rb
```diff
@@ -111,7 +111,10 @@ def parse
 
       validate(options)
 
-      @config = if options[:configfile]
+      @config = if !Bolt::Util.windows? && ENV['BOLT_PROJECT']
```
nicklewis, line 114: Is there a reason not to support using the environment variable on Windows? Ostensibly the reason for this feature is to support using world writable directories, but it doesn't _have_ to be used for that and it could still be helpful on Windows.

lib/bolt/cli.rb
```diff
@@ -111,7 +111,10 @@ def parse
...
+                  project = Bolt::Project.create_project(ENV['BOLT_PROJECT'], 'environment')
```
nicklewis, line 115: We should call `expand_path` to support `~` and relative paths

beechtom: `Bolt::Project.create_project` already calls `Pathname.new(path).expand_path`.

lib/bolt/cli.rb
beechtom: Heh, I was debating back and forth on that this morning and settled on not supporting it since I saw it as more closely tied with the world writable check. But I do think it would be generally helpful, so I'll remove the windows check.

PR #1948, opened 2020-06-26
(GH-1920) Add --log-level CLI option

### LINKED ISSUE #1920: Bolt Apply log output

beechtom:
This adds a new `--log-level` CLI option that accepts a log level. When
using this option, any existing configuration for the console's log
level will be overridden with the specified level. The `--log-level`
option cannot be used in conjunction with the `--debug` option.

This also adds a warning that the `--debug` option is deprecated and
that Bolt will set `--log-level debug` instead.

Closes #1920 
Supersedes #1921 

!feature

* **Added `--log-level` CLI option** ([#1920](https://github.com/puppetlabs/bolt/issues/1920))

  The new `--log-level` CLI option can be used to override the console's
  log level. It accepts the following log levels: `debug`, `info`,
  `notice`, `warn`, `error`, `fatal`, `any`.

### DIFF
--- a/lib/bolt/bolt_option_parser.rb
+++ b/lib/bolt/bolt_option_parser.rb
@@ -13,7 +13,7 @@ module Bolt
                 global_config_setters: %w[modulepath project configfile],
                 transports: %w[transport connect-timeout tty ssh-command copy-command],
                 display: %w[format color verbose trace],
-                global: %w[help version debug] }.freeze
+                global: %w[help version debug log-level] }.freeze
 
     ACTION_OPTS = OPTIONS.values.flatten.freeze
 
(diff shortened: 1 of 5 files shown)


lib/bolt/bolt_option_parser.rb
```diff
@@ -803,8 +803,15 @@ def initialize(options)
...
       define('--debug', 'Display debug logging') do |_|
         @options[:debug] = true
+        # We don't actually set '--log-level debug' here, but once the options are evaluated by
+        # the config class the end result is the same.
+        @warnings << { msg: "Command line option '--debug' is deprecated, setting '--log-level debug' instead." }
+      end
+      define('--log-level LEVEL',
+             "Set the log level for the console. Available options are",
+             "debug, info, notice, warn, error, fatal, any.") do |level|
+        @options[:log] = { 'console' => { 'level' => level } }
```
MikaelSmith, line 813: Is this before config is loaded? Want to make sure this doesn't override file logging from a config file.

beechtom: It happens before config is loaded, but all of our CLI options override logging from a config file. `log` specifically is merged, so it will only affect the config for `console`.

lib/bolt/bolt_option_parser.rb
```diff
@@ -803,8 +803,15 @@ def initialize(options)
       end
       define('--debug', 'Display debug logging') do |_|
         @options[:debug] = true
+        # We don't actually set '--log-level debug' here, but once the options are evaluated by
+        # the config class the end result is the same.
+        @warnings << { msg: "Command line option '--debug' is deprecated, setting '--log-level debug' instead." }
```
nicklewis, line 808: This phrasing is a little weird. The use of a flag in the message makes me assume it's going to be instructions, but it's a description. Maybe just "[...], set '--log-level debug' instead."

PR #1952, opened 2020-06-29
(GH-1938) Add native-ssh config option and warn_once logging utility

### LINKED ISSUE #1938: Pull external ssh transport into a separate transport class
## Use Case
Currently, the external SSH transport is folded into the `Bolt::Transport::SSH` class even though our docs describe it as a separate transport. This results in the SSH transport having a list of config options that may or may not be available depending on whether the `ssh-command` option is set. Because of this, the reference docs for the SSH transport are more complicated than the other transports, as we need to mark which options are supported when using the `ssh-command` option.

We should create an external SSH transport that is wholly separate from the existing SSH transport. This would still require users to explicitly opt-in to using the transport, while allowing us to set a reasonable default value for the SSH command and simplify our documentation for which options are available for the external SSH transport.

lucywyman:
This PR accomplishes 3 main tasks:
* It renames `external-ssh` to `native-ssh`, to make the purpose of the configuration option and documentation more intuitive and clear
* It adds a new `native-ssh` SSH configuration option which is used to enable shelling out to SSH rather than using the Ruby net-ssh library. Previously this option was toggled based on the presence of `ssh-command` configuration, which overloaded the option to both enable native ssh and provide the command that should be used to connect via SSH. This makes it so that if `native-ssh` is not set, `ssh-command` will be ignored
* Lastly it adds a new `warn_once` logging utility, which stores warn_once warnings in an instance variable on the `Bolt::Logger` class and verifies each warning type of warning is only printed once per Bolt run. Accessing the variable is protected by a mutex to avoid race conditions.

Closes #1938

### DIFF
--- a/lib/bolt/bolt_option_parser.rb
+++ b/lib/bolt/bolt_option_parser.rb
@@ -747,7 +747,7 @@ module Bolt
              "This option is experimental.") do |exec|
         @options[:'ssh-command'] = exec
       end
-      define('--copy-command EXEC', "Command to copy files to remote hosts if using external SSH. ",
+      define('--copy-command EXEC', "Command to copy files to remote hosts if using native SSH. ",
              "This option is experimental.") do |exec|
         @options[:'copy-command'] = exec
       end
--- a/lib/bolt/bolt_option_parser.rb
+++ b/lib/bolt/bolt_option_parser.rb
@@ -11,7 +11,7 @@ module Bolt
                 escalation: %w[run-as sudo-password sudo-password-prompt sudo-executable],
                 run_context: %w[concurrency inventoryfile save-rerun cleanup],
                 global_config_setters: %w[modulepath project configfile],
-                transports: %w[transport connect-timeout tty ssh-command copy-command],
+                transports: %w[transport connect-timeout tty native-ssh ssh-command copy-command],
                 display: %w[format color verbose trace],
                 global: %w[help version debug log-level] }.freeze
 
@@ -743,7 +743,11 @@ module Bolt
              "Specify a default transport: #{TRANSPORTS.keys.join(', ')}") do |t|
         @options[:transport] = t
       end
-      define('--ssh-command EXEC', "Executable to use instead of the net-ssh ruby library. ",
+      define('--[no-]native-ssh', 'Whether to shell out to native SSH or use the net-ssh Ruby library.',
+             'This option is experimental') do |bool|
+        @options[:'native-ssh'] = bool
+      end
+      define('--ssh-command EXEC', "Executable to use instead of the net-ssh Ruby library. ",
              "This option is experimental.") do |exec|
         @options[:'ssh-command'] = exec
       end
--- a/lib/bolt/logger.rb
+++ b/lib/bolt/logger.rb
@@ -15,6 +15,7 @@ module Bolt
       return if Logging.initialized?
 
       Logging.init :debug, :info, :notice, :warn, :error, :fatal, :any
+      @mutex = Mutex.new
 
       Logging.color_scheme(
         'bolt',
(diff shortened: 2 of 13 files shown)


lib/bolt/bolt_option_parser.rb
```diff
@@ -743,11 +743,15 @@ def initialize(options)
              "Specify a default transport: #{TRANSPORTS.keys.join(', ')}") do |t|
         @options[:transport] = t
       end
-      define('--ssh-command EXEC', "Executable to use instead of the net-ssh ruby library. ",
+      define('--[no-]native-ssh', 'Whether to shell out to native SSH or use the net-ssh Ruby library.',
+             'This option is experimental') do |bool|
+        @options[:'native-ssh'] = bool
+      end
```
beechtom, line 749: This option doesn't get picked up by the config class since `Bolt::Config#normalize_overrides` pulls out the relevant overrides using the `OPTIONS` hash: https://github.com/puppetlabs/bolt/blob/master/lib/bolt/config.rb#L250-L253

schemas/bolt-transport-definitions.json
```diff
@@ -199,6 +200,10 @@
       "type": "string",
       "enum": ["sh", "bash", "zsh", "dash", "ksh", "powershell"]
     },
+    "native-ssh": {
+      "description": "Whether to shell out directly to SSH or use the Ruby net-ssh library",
```
beechtom, line 204:
```suggestion
      "description": "Whether to shell out directly to SSH or use the Ruby net-ssh library.",
```

lib/bolt/bolt_option_parser.rb
lucywyman: It should - I modified `options` in the SSH config class to include ssh-command and native-ssh. They won't be documented because they aren't in OPTIONS.

lucywyman: Ah, I need to add 'native-ssh' there

schemas/bolt-transport-definitions.json
lucywyman: This was literally the first thing I did on this PR so I wouldn't forget

beechtom: Hopefully we can eventually have this generated so we can just forget about it :D

lib/bolt/logger/util.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+module Bolt
+  module Logger
+    module Util
+      # This should be fine to track once per logger instance
+      def warn_once(msg)
```
nicklewis, line 7: We should only warn once for a particular _kind_ of warning, not just the exact message. Like if multiple targets have the same config problem (maybe because you set it in the inventory), you should only see the warning one time. So we should have some sort of identifier here for "which" warning we're issuing (or we could take a formatter string and args to be interpolated).

On the same note, I'm unclear how having one `@warnings` per logger instance works with that. Will that cause us to still issue a warning once per target?

lucywyman: For this specifically, it will still only warn once ever - it would make it so we couldn't use warn_once in a transport connection class, but I didn't see any particular warnings there currently that we'd want to warn_once and this makes managing the lifecycle of the warn_once *much* easier. I think the best alternative is to have a warn_once singleton - would that be better?

lib/bolt/logger.rb
```diff
@@ -102,5 +102,17 @@ def self.levels
     def self.reset_logging
       Logging.reset
     end
+
+    def self.warn_once(type, msg)
+      @mutex ||= Mutex.new
```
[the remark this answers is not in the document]
lucywyman: Great point, I should have thought that through more.

PR #1953, opened 2020-06-29
(bug-fix) Find embedded projects specified on the CLI

lucywyman:
This modifies how projects specified with `--project` or `--boltdir` are
found. Previously the provided path would be used as the path to the
project, where bolt configuration was expected to be found. However with
embedded projects - those that store files in `Boltdir` - users would
specify the parent directory of `Boltdir` expect the project to be
loaded as if they were executing Bolt from that parent directory. This
modifies how we handle project directories first looks for a `Boltdir`
in the specified directory, and if none is present use the specified
directory as the project root.

!bug

* **Load projects with an embedded Boltdir when specified on the CLI**

  We now look for a `Boltdir` in the directory specified by `--project`
  or `--boltdir` and use that if present. If not, we use the specified
  path as the project root.

### DIFF
--- a/lib/bolt/cli.rb
+++ b/lib/bolt/cli.rb
@@ -115,7 +115,12 @@ module Bolt
                   Bolt::Config.from_file(options[:configfile], options)
                 else
                   project = if options[:boltdir]
-                              Bolt::Project.create_project(options[:boltdir])
+                              dir = Pathname.new(options[:boltdir])
+                              if (dir + Bolt::Project::BOLTDIR_NAME).directory?
+                                Bolt::Project.create_project(dir + Bolt::Project::BOLTDIR_NAME)
+                              else
+                                Bolt::Project.create_project(dir)
+                              end
                             else
                               Bolt::Project.find_boltdir(Dir.pwd)
                             end
--- a/spec/fixtures/projects/embedded/plans/init.pp
+++ b/spec/fixtures/projects/embedded/plans/init.pp
@@ -1,7 +1,0 @@
-plan embedded(
-  TargetSpec $targets
-) {
-  $target_object = get_target($targets)
-  return run_command("echo polo", $target_object)
-}
-
--- a/spec/integration/project_spec.rb
+++ b/spec/integration/project_spec.rb
@@ -1,14 +1,17 @@
 # frozen_string_literal: true
 
 require 'spec_helper'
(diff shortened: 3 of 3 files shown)


lib/bolt/cli.rb
```diff
@@ -114,12 +114,8 @@ def parse
...
                   Bolt::Config.from_file(options[:configfile], options)
                 else
-                  project = if options[:boltdir]
-                              Bolt::Project.create_project(options[:boltdir])
-                            else
-                              Bolt::Project.find_boltdir(Dir.pwd)
-                            end
-                  Bolt::Config.from_project(project, options)
+                  boltdirpath = options[:boltdir] || Dir.pwd
+                  Bolt::Config.from_project(Bolt::Project.find_boltdir(File.expand_path(boltdirpath)), options)
```
nicklewis, line 118: I think this will cause the user-level project to be used if you pass `--project` with a directory that doesn't exist / doesn't have a `bolt.yaml`/`bolt-project.yaml`, which is undesirable.

lucywyman: That's a good point - I updated to first look for a Boltdir, otherwise use the specified path.

lib/bolt/cli.rb
```diff
@@ -114,10 +114,11 @@ def parse
       @config = if options[:configfile]
                   Bolt::Config.from_file(options[:configfile], options)
                 else
-                  project = if options[:boltdir]
-                              Bolt::Project.create_project(options[:boltdir])
+                  boltdirpath = options[:boltdir] || Dir.pwd
+                  project = if (boltdirpath + Bolt::Project::BOLTDIR_NAME).directory?
+                              Bolt::Project.create_project(boltdirpath + Bolt::Project::BOLTDIR_NAME, 'embedded')
                             else
-                              Bolt::Project.find_boltdir(Dir.pwd)
```
nicklewis, line 120: This now removes the only use of the `find_boltdir` method. It will cause us to _never_ use the user- or system-level project.

I think the logic needs to be:
a) if `options[:boltdir]` is set and contains `Boltdir`, use `options[:boltdir] + Boltdir`
b) if `options[:boltdir]` is set and doesn't contain `Boltdir`, use `options[:boltdir]`
c) otherwise call `find_boltdir(Dir.pwd)`

PR #1954, opened 2020-06-29
(GH-1932) Add location of plan failures in error messages

### LINKED ISSUE #1932: Log helpful details when nested plan fails with bolt/pal-error
## Use Case

When a function has a failure in a Bolt plan, it can be very difficult to track down where in the plan the error came from. consider the following output:

```Finished: plan peadm::util::add_cert_extensions in 32.37 sec
Starting: plan peadm::util::add_cert_extensions
Starting: task peadm::trusted_facts on puppetmom01.newlab.sv.cloud.wellsfargo.net
Finished: task peadm::trusted_facts with 0 failures in 0.86 sec
Starting: task peadm::trusted_facts on puppetcm01.newlab.sv.cloud.wellsfargo.net
Finished: task peadm::trusted_facts with 0 failures in 1.43 sec
Starting: plan peadm::util::insert_csr_extension_requests
Starting: task peadm::read_file on puppetcm01.newlab.sv.cloud.wellsfargo.net
Finished: task peadm::read_file with 0 failures in 1.42 sec
Finished: plan peadm::util::insert_csr_extension_requests in 1.42 sec
Finished: plan peadm::util::add_cert_extensions in 3.72 sec
Finished: plan peadm::convert in 5 min, 59 sec
{
  "kind": "bolt/pal-error",
  "msg": "(<unknown>): did not find expected key while parsing a block mapping at line 2 column 1",
  "details": {
  }
}
```

Where did this error come from? It's pretty unclear, especially since Bolt kind of hides information about when in the log output the error actually occurred by printing a lot of green "Finished: ..." lines _after_ the error occured.

When this happens, the error message should include at a minimum the Puppet plan file name and line number where the error was raised. Maybe even more details if they are available. E.g.,

```
{
  "kind": "bolt/pal-error",
  "msg": "(<unknown>): did not find expected key while parsing a block mapping at line 2 column 1",
  "details": {
    "function": "parseyaml",
    "file": "peadm/plans/util/insert_csr_attributes.pp",
    "line": 11
  }
}
```

## Todo
- Investigate if this is happening because we're throwing away useful information from Puppet, or if Puppet just isn't providing us the information we want or need.
- This came up from the installer team using this and finding errors in plans that call plans
- This issue specifically seems to happen when a plan calls a plan which calls `parse_yaml` with invalid yaml, so specifically providing a file and line number would be really useful.
- We should try to make sure that when run_* functions error they provide info on where they were called from (updates to plan function errors basically)

beechtom:
This updates the error raised by a plan failure to include the location
of the plan failure, including the filepath, line, and column of the
error's source. This information is included in both the error message
and under the `details` key of the error hash. Previously, this
information was discarded by Bolt when rescuing an error raised by
Puppet, resulting in only the cause of the error to be included in the
message.

Closes #1932 

!feature

* **Add location of plan failures in error messages**
  ([#1923](https://github.com/puppetlabs/bolt/issues/1923))

  Errors raised by plan failures now include the location of the plan
  failure, including the filepath, line, and column.

### DIFF
--- a/lib/bolt/pal.rb
+++ b/lib/bolt/pal.rb
@@ -15,25 +15,36 @@ module Bolt
     # PALError is used to convert errors from executing puppet code into
     # Bolt::Errors
     class PALError < Bolt::Error
-      # Puppet sometimes rescues exceptions notes the location and reraises.
-      # Return the original error.
       def self.from_preformatted_error(err)
         if err.cause&.is_a? Bolt::Error
           err.cause
         else
-          from_error(err.cause || err)
+          from_error(err)
         end
       end
 
       # Generate a Bolt::Pal::PALError for non-bolt errors
       def self.from_error(err)
-        e = new(err.message)
+        # Use the original error message if available
+        message = err.cause ? err.cause.message : err.message
+
+        # Provide the location of an error if it came from a plan
+        details = if defined?(err.file) && err.file
+                    { file:   err.file,
+                      line:   err.line,
+                      column: err.pos }.compact
+                  else
+                    {}
+                  end
+
+        e = new(message, details)
+
         e.set_backtrace(err.backtrace)
         e
       end
 
-      def initialize(msg)
-        super(msg, 'bolt/pal-error')
+      def initialize(msg, details = {})
+        super(msg, 'bolt/pal-error', details)
       end
     end
 
@@ -159,8 +170,9 @@ module Bolt
               if e.issue_code == :UNKNOWN_VARIABLE &&
                  %w[facts trusted server_facts settings].include?(e.arguments[:name])
                 message = "Evaluation Error: Variable '#{e.arguments[:name]}' is not available in the current scope "\
-                  "unless explicitly defined. (file: #{e.file}, line: #{e.line}, column: #{e.pos})"
-                PALError.new(message)
+                          "unless explicitly defined."
+                details = { file: e.file, line: e.line, column: e.pos }
+                PALError.new(message, details)
               else
                 PALError.from_preformatted_error(e)
               end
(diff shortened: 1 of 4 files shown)


lib/bolt/pal.rb
```diff
@@ -15,25 +15,33 @@ class PAL
...
-          from_error(err.cause || err)
+          from_error(err)
         end
       end
 
       # Generate a Bolt::Pal::PALError for non-bolt errors
       def self.from_error(err)
-        e = new(err.message)
+        # Provide the location of an error if it came from a plan
+        details = if defined?(err.file)
```
lucywyman, line 29: https://github.com/puppetlabs/bolt/blob/master/lib/bolt/pal.rb#L162 It looks like we have file/line/column in this message already. I like this format better, so I think removing that part of the message would be good for consistency + deduplication.

lucywyman: More directly relevant to this line specifically, it looks like this will add these details if `err.file` is defined on the class, even if it and the line and position are nil:
```
lucy ~/githubs/bolt (1932/plan-errors) $ be bolt plan run bolt
Starting: plan bolt
Finished: plan bolt in 0.01 sec
{
  "kind": "bolt/pal-error",
  "msg": "Function lookup() did not find a value for the name 'abcdef'",
  "details": {
    "file": null,
    "line": null,
    "column": null
  }
}
```

It might be better to ensure this exists *and* isn't nil. It actually looks like these might be set separately, so we could theoretically just add each item to the hash if it isn't nil, but having a line and column without a file name seems pretty useless so I'm inclined to just not add the details if the file is nil.

beechtom: Updated. If these changes cover what we want I'll squash the commits. :)

lucywyman:
A squash seems good, then will merge!

PR #1956, opened 2020-06-29
(packaging) Make the paint gem required by the rainbow outputter optional

donoghuc:
Given we dont want to ship the paint gem with PE this commit makes the gem optional. The gem will only attempt to be loaded when the rainbow outputter is requested. The paint gem will ship with bolt packages and is included in the Gemfile, but is ommited from the gemspec.

### DIFF
--- a/.github/workflows/linux.yaml
+++ b/.github/workflows/linux.yaml
@@ -32,7 +32,7 @@ jobs:
         uses: actions/cache@v1
         with:
           path: vendor/bundle
-          key: ${{ runner.os }}-gems-${{ hashFiles('lib/bolt/version.rb') }}-${{ hashFiles('**/Gemfile') }}-${{ hashFiles('bolt.gemspec') }}
+          key: ${{ runner.os }}-gems-${{ hashFiles('lib/bolt/version.rb') }}-${{ hashFiles('Gemfile') }}-${{ hashFiles('bolt.gemspec') }}
       - name: Install gems
         if: steps.cache.outputs.cache-hit != 'true'
         run: bundle install --jobs 4 --retry 3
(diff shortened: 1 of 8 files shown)


lib/bolt/outputter.rb
```diff
@@ -9,6 +9,11 @@ def self.for_format(format, color, verbose, trace)
       when 'json'
         Bolt::Outputter::JSON.new(color, verbose, trace)
       when 'rainbow'
+        begin
+          require 'bolt/outputter/rainbow'
+        rescue LoadError
```
nicklewis, line 14: I think it would be better to move `require 'paint'` into the initialize method of the Rainbow outputter and guard it there. That would be a little more constrained than guarding the selection+loading of the entire outputter.

nicklewis: Specifically, it seems less prone to error to have the code fail when you try to _use_ it rather than fail when you try to _load_ it.

donoghuc: For sure. I can move it there.

.github/workflows/linux.yaml
```diff
@@ -32,7 +32,7 @@ jobs:
         uses: actions/cache@v1
         with:
           path: vendor/bundle
-          key: ${{ runner.os }}-gems-${{ hashFiles('lib/bolt/version.rb') }}-${{ hashFiles('**/Gemfile') }}-${{ hashFiles('bolt.gemspec') }}
+          key: ${{ runner.os }}-gems-${{ hashFiles('lib/bolt/version.rb') }}-${{ hashFiles('Gemfile') }}-${{ hashFiles('bolt.gemspec') }}
```
beechtom, line 35: This should be updated in each workflow that caches gems.

donoghuc: Doh! I think i got them all.... It is past 5pm if i missed any i'll update first thing in the am.

.github/workflows/modules.yaml
```diff
@@ -32,7 +32,7 @@ jobs:
         uses: actions/cache@v1
         with:
           path: vendor/bundle
-          key: ${{ runner.os }}-gems-${{ hashFiles('lib/bolt/version.rb') }}-${{ hashFiles('**/Gemfile') }}-${{ hashFiles('bolt.gemspec') }}
+          key: ${{ runner.os }}-gems-${{ hashFiles('lib/bolt/version.rb') }}-${{ hashFiles('Gemfile') }}-${{ hashFiles('bolt.gemspec') }}
```
lucywyman, line 35: Sorry there's another cache step on line 63 of this file that needs to be updated as well.

donoghuc: lol. knew i was likely to miss at least one. was halfway out the door last night. Thanks.

Issue #1900, opened 2020-06-15
JSON schemas flag plugin references as invalid


beechtom:
## Describe the Bug
Many of the options that can be set in `bolt.yaml`, `bolt-project.yaml`, and `inventory.yaml` expect a non-hash value but accept a plugin reference that will be resolved by Bolt and (hopefully) return a value with the expected type. The JSON schema correctly validates options based on their values' expected types, but does not support adding plugin references for options that accept them and will mark these plugin references as invalid.

## Expected Behavior
The JSON schemas should support having a plugin reference anywhere they are supported and not flag the plugin reference as invalid.

## Steps to Reproduce
1. Enable support for the schemas in VS Code
2. Create an inventoryfile with the following:
   ```yaml
   ssh:
     config:
       password:
         _plugin: vault
        path: secrets/bolt
        field: bolt-password
   ```

VS Code will highlight `_plugin` as invalid and display the error `Incorrect type. Expected "string".`

## Questions

- Can we generate these instead of hard-coding them?

PR #1959, opened 2020-07-01
(GH-1900) Generate JSON schemas with plugin references

beechtom:
This contains changes that enable generating the JSON schemas using a
rake task using data contained in the `Bolt::Config::Options` and
`Bolt::Config::Transport::TRANSPORT_OPTIONS` constants.

A new Rakefile `schemas/Rakefile` contains rake tasks that generate
each of the JSON schemas. These tasks can be run using the following
command:

```shell
$ rake -f schemas/Rakefile schemas
```

This also updates the `Bolt::Config::OPTIONS` and
`Bolt::Config::Transport::TRANSPORT_OPTIONS` constants to include
additional data for each option that is included in the JSON schemas.
The additional keys are direct equivalents to keys recognized by the
JSON Schema standard, Draft 07. The constants also include several
metadata keys such as `:_example` and `:_default`, which are used to
generate docuemntation for the configuration files.

This change also updates the JSON schemas to support plugin references.
The `:_plugin` metadata key can be set on an option to indicate that the
JSON schema should include a plugin reference as a valid value for the
option.

!bug

* **Support plugin references in JSON schemas**
  ([#1900](#1900))

  The JSON schemas no longer mark plugin references as invalid values
  when the option can accept a plugin reference. Previously, the schemas
  would mark any plugin reference as an invalid value.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -17,20 +17,18 @@ unless ENV['GITHUB_WORKFLOW']
   Pkg::Util::RakeUtils.load_packaging_tasks
 end
 
-def generate_yaml_file(data, subdata = {})
-  data.each_with_object({}) do |(option, option_data), acc|
-    if option_data.key?(:exmp)
-      acc[option] = option_data[:exmp]
-    elsif subdata.key?(option)
-      acc[option] = generate_yaml_file(subdata[option])
+def generate_yaml_file(data)
+  data.each_with_object({}) do |(option, definition), acc|
+    if definition.key?(:_example)
+      acc[option] = definition[:_example]
+    elsif definition.key?(:properties)
+      acc[option] = generate_yaml_file(definition[:properties])
     end
   end
 end
 
-def stringify_types(opts)
-  opts.each_value do |data|
-    next unless data.key?(:type)
-
+def stringify_types(data)
+  if data.key?(:type)
     types = Array(data[:type])
 
     if types.include?(TrueClass) || types.include?(FalseClass)
@@ -39,6 +37,14 @@ def stringify_types(opts)
 
     data[:type] = types.join(', ')
   end
+
+  if data.key?(:properties)
+    data[:properties] = data[:properties].transform_values do |d|
+      stringify_types(d)
+    end
+  end
+
+  data
 end
 
 desc "Run all RSpec tests"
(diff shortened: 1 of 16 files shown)


Rakefile
```diff
@@ -125,23 +131,33 @@ end
...
-    @subopts   = Bolt::Config::SUBOPTIONS.dup
-    @inventory = Bolt::Config::INVENTORY_OPTIONS
+    @opts      = Bolt::Config::OPTIONS.slice(*Bolt::Config::BOLT_OPTIONS)
+    @inventory = Bolt::Config::INVENTORY_OPTIONS.dup
 
-    @yaml           = generate_yaml_file(@opts, @subopts)
-    @inventory_yaml = generate_yaml_file(@inventory)
+    # Move sub-options for 'log' option up one level, as they're nested under
+    # 'console' and filepath
+    @opts['log'][:properties] = @opts['log'][:additionalProperties][:properties]
```
lucywyman, line 139: Should this be additive so that `console` is also documented? It makes sense with the example but is a bit confusing as text - though I'll defer to @hestonhoffman .

hestonhoffman: Ah, I see. So if we leave it like this, `console` is only documented in the description under log and in the examples. The examples do clear things up somewhat, but I think I agree with Lucy that it feels weird to skip a level in the hierarchy. However, we do something similar with `plugin_hooks`, so it's not unprecedented. Maybe if we just clear the language up a bit. Something like:

"A map of configuration for the logfile output. Under `log`, you can configure log options for your `console` and add configuration for individual log files, such as `~/.puppetlabs/bolt/debug.log`."

I think keeping it short and sweet and relying more on the example might be the best way to handle this one.

schemas/bolt-inventory.json
```diff
@@ -0,0 +1,228 @@
...
+    "config": {
+      "$ref": "#/definitions/config"
+    },
+    "facts": {
+      "$ref": "#/definitions/facts"
+    },
+    "features": {
+      "$ref": "#/definitions/features"
+    },
+    "groups": {
```
lucywyman, line 17: I'm a bit confused about why we're committing this file? It looks like the rake task generates `bolt-inventory.schema.json`.

beechtom: `bolt-inventory.json` is the base file that `bolt-inventory.schema.json` is built on. The inventory schemas is more complicated than the others to generate from scratch, since we don't have a list of available fields similar to the config class _and_ several of the fields can reference other fields. Since most of these fields realistically won't change or be added to, they're hardcoded while the inventory options are added in during the rake task.

beechtom:
The schemas should not be added to the gitignore, as the forge proxies point at the files on the main branch.

schemas/bolt-inventory.json
beechtom: Renamed the file to `bolt-inventory-base.json`.

lucywyman: Ugh I thought that was the case and grepping for `File.read` just failed. Gah, Monday.

lucywyman:
Ok I think this is ready to have the schemas committed and merge

beechtom:
Schemas are committed 👍

PR #1963, opened 2020-07-01
(bug) Validate 'inventory-config' option in bolt-defaults.yaml

beechtom:
This adds some validation checks for the `inventory-config` option in
the `bolt-defaults.yaml` file. Previously, setting this option to a
value other than a hash would result in an unhelpful error, as Bolt was
assuming that the value would always be a hash. Additionally, if the
value was a plugin reference, the plugin reference would be merged with
the top-level config, potentially resulting in config values being
overwritten and resulting in the entire config hash being considered a
plugin reference.

Now, Bolt will first check that the `inventory-config` option is a hash
and is not a plugin reference before continuing with loading config.

!bug

* **Validate `inventory-config` option in `bolt-defaults.yaml`**

  Bolt now checks that the `inventory-config` option in a
  `bolt-defaults.yaml` file is a hash and is not a plugin reference
  before merging configuration files. Previously, setting this value to
  a hash or plugin reference would raise an unhelpful error.

### DIFF
--- a/lib/bolt/config.rb
+++ b/lib/bolt/config.rb
@@ -139,9 +139,22 @@ module Bolt
         )
       end
 
-      # Move data under transport-config to top-level so it can be easily merged with
-      # config from other sources.
+      # Move data under inventory-config to top-level so it can be easily merged with
+      # config from other sources. Error early if inventory-config is not a hash or
+      # has a plugin reference.
       if data.key?('inventory-config')
+        unless data['inventory-config'].is_a?(Hash)
+          raise Bolt::ValidationError,
+                "Option 'inventory-config' must be of type Hash, received #{data['inventory-config']} "\
+                "#{data['inventory-config']} (file: #{filepath})"
+        end
+
+        if data['inventory-config'].key?('_plugin')
+          raise Bolt::ValidationError,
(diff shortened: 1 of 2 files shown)


lib/bolt/config.rb
```diff
@@ -139,9 +139,22 @@ def self.load_bolt_defaults_yaml(dir)
...
+      # config from other sources. Error early if inventory-config is not a hash or
+      # has a plugin reference.
       if data.key?('inventory-config')
+        unless data['inventory-config'].is_a?(Hash)
+          raise Bolt::ValidationError,
+                "Option 'inventory-config' must be of type Hash, received #{data['inventory-config']} "\
+                "#{data['inventory-config']} (file: #{filepath})"
+        end
+
+        if data['inventory-config'].key?('_plugin')
```
nicklewis, line 152: Why do we not support plugins here? I thought we did some work to allow transport config in `bolt.yaml` to support plugin references. Would this be a regression or is it just not supported at the _top_ level?

beechtom: It's just the top-level `inventory-config` key that can't be a plugin reference. The individual transports can be plugin references, though.

PR #1970, opened 2020-07-08
Updated bash completions with missing commands.

atika:
Hi.
I tried my best to add missing commands and correct some of them. 
For commands that need `--targets`, I have the idea to put the result in a cache file associated with the current session, like that is faster on the next usage of the `--targets`. I did not add an option to delete the file.
And I also added a test before to check if the previous keywords are a flag and not need completion like the `--targets` or `--rerun` flag to avoid losing the current command completion.

I tested on macOS ZSH and Bash, and on Ubuntu with Bash. It seems to works fine.

Let me know what you think about the modifications.

### DIFF
--- a/resources/bolt_bash_completion.sh
+++ b/resources/bolt_bash_completion.sh
@@ -6,74 +6,134 @@
 ###
 ### Marc Schoechlin ms-github@256bit.org
 
-_bolt()
-{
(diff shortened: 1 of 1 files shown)


nicklewis:
I updated the commit message and added a release note marker and merged it. Thank you for the updates!

PR #1971, opened 2020-07-08
(GH-1917) Don't create Bolt Config object before loading config

### LINKED ISSUE #1917: `Bolt::Project` instantiation errors (e.g. validation) will always have a traceback
`Bolt::Project` instantiation errors (e.g. validation) will always have a traceback. This happens because we create a 'default' `Bolt::Config` object in order to load default config values to create a `Bolt::Outputter` to use in case Bolt errors before we've loaded real config. `Bolt::Config.default` runs `Bolt::Project.create_project('.')`, which is sometimes where the actual project is, which, if it errors, does not get handled by Bolt and causes a traceback.

There's (at least) 2 related issues here:
1. We should probably have an unconfigured outputter object so we don't need to load `Bolt::Config.default` when we instantiate the CLI
2.`Bolt::Config.default` is useful in other places, and (probably) shouldn't load the cwd as a project. It could load an empty project that doesn't have a path, or that has a tempdir path (but then lifecycle issues), or it could just not have a project. 

## Steps to Reproduce
1. Create a `bolt-project.yaml` with an illegal name:
```
name: 123
```
2. Run `bolt --help`

## Environment
Any

## Solution

- Remove config.default from CLI instantiation
- Make the outputter in the CLI before logger is configured behave like outputter's `fatal_error` method. It should only rescue errors that occur before the logger is configure (Bolt::Logger.configure).

lucywyman:
Previously when we created a new instance of the `Bolt::CLI` object we
also loaded the default `Bolt::Config` object in order to create a
default outputter to print errors with. This caused any errors loading
config or the default project to be unhandled and backtrace. The error
was especially annoying because it used the default project path at
`~/.puppetlabs/bolt`, regardless of whether that was the actual project
that would be used during the Bolt run.

This commit moves the same behavior of the default outputter directly
into the CLI (namely printing any error messages to stdout, in red if
it's a tty) so that we no longer need to load the default project for
every invocation of Bolt. This makes it so any errors that occur when
loading config are printed, and their exit code returned, similar to
other errors raised in Bolt::CLI.

This also standardizes on calling the accessor method to get `config` in the CLI, so that it can easily be mocked for testing.

Closes #1917

!bug

* **Don't load default project for every Bolt invocation** ([1917](https://github.com/puppetlabs/bolt/issues/1917)

  We no longer load the default project at `~/.puppetlabs/bolt` for
  every Bolt invocation. When projects are loaded, the errors are now
  handled and don't backtrace.

### DIFF
--- a/lib/bolt/cli.rb
+++ b/lib/bolt/cli.rb
@@ -46,7 +46,6 @@ module Bolt
       Bolt::Logger.initialize_logging
       @logger = Logging.logger[self]
       @argv = argv
-      @config = Bolt::Config.default
       @options = {}
     end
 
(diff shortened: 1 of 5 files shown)


lib/bolt/cli.rb
```diff
@@ -163,8 +162,16 @@ def parse
...
+      # This needs to happen here to avoid a regression where error messages from
+      # loading inventory aren't printed
     rescue Bolt::Error => e
-      outputter.fatal_error(e)
+      if $stdout.isatty
+        # Print the error message in red, mimicking outputter.fatal_error
+        $stdout.puts("\033[31m#{e.message}\033[0m")
+      else
+        $stdout.puts(e.message)
+      end
```
beechtom, line 174: We shouldn't replace using the outputter here, as errors raised after the config is loaded will no longer provide the same information they did before this change. For example, when using JSON output, having a target/group name conflict will now only show this:

```
Group linux conflicts with target of the same name for group at ["linux", "all"]
```

When it used to show this:
```
{"kind":"bolt.inventory/validation-error","msg":"Group linux conflicts with target of the same name for group at [\"linux\", \"all\"]","details":{"path":["linux","all"]}}
```

We should instead have the outputter still handle all errors raised after the config is loaded and only print directly to stdout when it is not yet loaded.

lucywyman: Using the human outputter that doesn't work though - if there's an error in inventory the message gets dropped.

lucywyman: I get it we need two rescues. fuck me.

Issue #1957, opened 2020-06-30
Edits to project directories page


hestonhoffman:
There's been some confusion around this page, partly due to Bolt projects. I'll investigate and try to clear things up.

- Use neutral language to avoid confusion with Bolt projects, ie. "Directory structure"
- Prefer local directory structure and explain what happens if a bolt-project.yaml is not found - ie where Bolt would load content from.
- Make it clearer that bolt.yaml is deprecated
- Remove references to "Boltdir" outside its use in control repos.

Also, as part of this ticket, pull plugins out from under inventory

PR #1976, opened 2020-07-09
(GH-1957) Projects and project directories 

hestonhoffman:
- [x] Edits to Project directories page
- [x] Adds Bolt projects page with experimental unicode characters
- [x] Adds unicode for deprecated bolt.yaml stuff
- [x] Pulls plugins out from underneath inventory files in the hierarchy
- [x] Adds modules parent page

Also fixes a broken link in templates/bolt_transports_reference.md.erb


documentation/projects.md
```diff
@@ -0,0 +1,114 @@
...
+
+One of the first things you'll do when you start using Bolt is create a Bolt
+project. A Bolt project is a simple directory that serves as the launching point
+for Bolt. You run Bolt from your project directory, and the directory houses
+your Bolt configuration files and your Bolt content, such as plans and tasks.
+
+In addition to working with your local Bolt content, Bolt projects give you a
+way to share that content with other users in your organization. You can create
+orchestration that is specific to the infrastructure you're working with, and
+then commit the project directory to git for others to consume.
```
beechtom, line 15:
```suggestion
then commit the project directory to version control for others to consume.
```

documentation/projects.md
```diff
@@ -0,0 +1,114 @@
...
+Bolt treats a directory as a Bolt project as long as a `bolt-project.yaml` file
+exists at the root of the directory and contains a `name` key. Bolt loads tasks
+and plans from the `tasks` and `plans` directories and namespaces them to the
+project name.
+
+## Configuring a project
+
+The `bolt-project.yaml` file holds options to configure your Bolt project, as
+well as options to control how Bolt behaves when you run it from inside the
+project.
```
beechtom, line 56:
```suggestion
well as options to control how Bolt behaves when you run the project.
```

documentation/bolt_project_directories.md
```diff
@@ -1,114 +1,89 @@
...
 
-## How the project directory is chosen
+If Bolt can't find a directory based on `Boltdir` or `bolt-project.yaml`, it
+uses the default: `~/.puppetlabs/bolt-project.yaml`.
 
-Bolt uses these methods, in order, to choose a project directory.
+Bolt uses the following methods, in order, to choose a Bolt directory.
 
-1. **Environment variable:** You can specify a path to a project using the
-   `BOLT_PROJECT` environment variable.
```
beechtom, line 102:
```
1. **Environment variable:** You can specify a path to a project using the
   `BOLT_PROJECT` environment variable.
```
This should be added back in.

documentation/bolt_project_directories.md
```diff
@@ -1,114 +1,89 @@
...
+   with `--project <DIRECTORY_PATH>`. There is not an equivalent configuration
+   setting because the Bolt directory must be known in order to load
+   configuration.
+2. **Parent directory:** Bolt traverses parents of the current directory until
+   it finds a directory containing a `Boltdir`, or
    `bolt-project.yaml`, or it reaches the root of the file system.
-1. **User project directory:** If no directory is specified manually or found in
-   a parent directory, the user project directory is used.
+3. **Default project directory:** If no project directory is specified manually or found in
+   a parent directory, Bolt uses `~/.puppetlabs/bolt-project.yaml`.
```
beechtom, line 86:
```suggestion
   a parent directory, Bolt uses `~/.puppetlabs/bolt/` as the project directory.
```

documentation/bolt_project_directories.md
```diff
@@ -138,24 +113,24 @@ If you want to use a world-writable directory for a single Bolt execution, set t
...
-|[`bolt-project.yaml`](bolt_configuration_reference.md#project_configuration_options)|Contains configuration for the Bolt project.  The `bolt-project.yaml` file can contain a list of tasks and plans to limit the output from the `bolt plan show` and `bolt task show` commands. This is an experimental feature. For more information, see [Bolt projects](./experimental_features.md#bolt-projects).|
+|[`bolt-project.yaml`](bolt_configuration_reference.md#project_configuration_options)|Contains configuration options for Bolt and Bolt projects. For more information on Bolt projects, see [Bolt projects](./experimental_features.md#bolt-projects).|
 |[`Puppetfile`](bolt_installing_modules.md#)|Specifies which modules to install for the project.|
 |[`modules/`](bolt_installing_modules.md#)|The directory where modules from the `Puppetfile` are installed. In most cases, do not edit these modules locally.|
-|[`site-modules`](bolt_installing_modules.md)|Local modules that are edited and versioned with the project directory.|
+|[`site-modules`](bolt_installing_modules.md)|Local modules that are edited and versioned with the Bolt directory.|
```
beechtom, line 135:
```suggestion
|[`site-modules/`](bolt_installing_modules.md)|Local modules that are edited and versioned with the Bolt directory.|
```

documentation/bolt_project_directories.md
```diff
@@ -1,114 +1,89 @@
...
-
-### User project directory
+As long as your `bolt-project.yaml` file contains a `name` field, Bolt loads
+your local Bolt content from the top level of the `Boltdir`. Your `Boltdir` can
+also contain modules. For more information, see [Modules](modules.md).
 
-If Bolt can't find a project directory based on `Boltdir` or `bolt.yaml`, it
-uses `~/.puppetlabs/bolt` as the project directory. Use this type of directory
-if you have a single set of Bolt code and data that you use across all projects.
+## How Bolt chooses a Bolt directory
```
lucywyman, line 71:
```suggestion
## How Bolt chooses a project directory
```

documentation/projects.md
```diff
@@ -0,0 +1,114 @@
...
+project. A Bolt project is a simple directory that serves as the launching point
+for Bolt. You run Bolt from your project directory, and the directory houses
+your Bolt configuration files and your Bolt content, such as plans and tasks.
+
+In addition to working with your local Bolt content, Bolt projects give you a
+way to share that content with other users in your organization. You can create
+orchestration that is specific to the infrastructure you're working with, and
+then commit the project directory to git for others to consume.
+
+Here is an example of a typical project with a task, a plan, and an inventory file:
```
lucywyman, line 17: Should task, plan, and inventory link out to docs about what those are?

documentation/projects.md
lucywyman: Run 'with' a project? I'm not sure what good phrasing is there.

documentation/projects.md
hestonhoffman: Let's add a related info section with links to those docs

documentation/bolt_project_directories.md
hestonhoffman: Ah good catch - I think I missed this when I was resolving a conflict

documentation/projects.md
hestonhoffman: Hmmm. Is the expectation that user will not always run Bolt from inside their project directory? I guess I was making an assumption here that most users will - especially if they have more than one Bolt project. This makes me think that I should expand a little further on `--project <DIRECTORY_PATH>`.

beechtom: Ya, there isn't really an expectation that users will always run Bolt from inside the project directory. This is especially true for users who are only using the default project directory, as that will be loaded from anywhere.

nicklewis: I don't think we should emphasize `--project` though. In general, you should be inside the project directory _except_ for the special case where you're using the default project. I think "run the project" works here.

hestonhoffman: Okay, I'm going to go with "run a project" for now. Further up in the doc, I've said 
> "You run Bolt from your project directory, and the directory houses
> your Bolt configuration files and your Bolt content, such as plans and tasks." 

I think that sets the desired interaction for the user. There are more details for how Bolt loads projects, and more ways to configure the project location, in the Project directories doc, but we can always revisit if this causes confusion.

documentation/bolt_project_directories.md
```diff
@@ -1,114 +1,91 @@
...
 
-### Embedded project directory
+As long as your `bolt-project.yaml` file contains a `name` field, Bolt loads
+your local Bolt content from the top level of your directory. If you're
+developing a module for the Puppet Forge, you can use a Puppet module
+directory structure. For more information, see [Module
+structure](module_structure.md). 
+
+
+> 🔩 **Tip:** You can use an existing control repo as a Bolt directory by adding
```
beechtom, line 35:
```suggestion
> 🔩 **Tip:** You can use an existing Puppet control repo as a Bolt directory by adding
```

I think clarifying that this is for existing Puppet users might be helpful, since it can be a bit confusing what we're talking about if you've never used Puppet.

documentation/bolt_project_directories.md
```diff
@@ -1,114 +1,91 @@
...
+also contain modules. For more information, see [Modules](modules.md).
 
-If Bolt can't find a project directory based on `Boltdir` or `bolt.yaml`, it
-uses `~/.puppetlabs/bolt` as the project directory. Use this type of directory
-if you have a single set of Bolt code and data that you use across all projects.
+## How Bolt chooses a project directory
 
-## How the project directory is chosen
+If Bolt can't find a directory based on `Boltdir` or `bolt-project.yaml`, it
+uses the default: `~/.puppetlabs/bolt-project.yaml`.
```
beechtom, line 74:
```suggestion
uses the default: `~/.puppetlabs/bolt/`.
```

documentation/projects.md
```diff
@@ -0,0 +1,119 @@
...
+```yaml
+# bolt-project.yaml
+name: myproject
+modulepath: ['modules','site-modules','/home/user/mymodules']
+
+log:
+  console:
+    level: notice
+  ~/.puppetlabs/bolt/debug.log
+    level: debug
```
beechtom, line 75: We should either mention something along the lines of "if you want to configure log output" in the preceding paragraph, or remove this part of the config since it isn't referenced.

documentation/projects.md
```diff
@@ -0,0 +1,119 @@
...
+```
+
+For a list of all the available configuration options in `bolt-project.yaml`,
+see [`bolt-project.yaml` options](bolt_project_reference.md).
+
+### Limiting displayed plans and tasks
```
beechtom, line 81: A small blurb about why a project author would want to limit the displayed plans and tasks would be helpful. Something along the lines of this maybe?

```
Bolt is bundled with several plans and tasks that might not be used by the project.
You might also write plans and tasks that should only be used by other plans and
tasks, and not be run directly by users. Projects allow you to limit which plans and
tasks a user can see when running `bolt plan show` or `bolt task show`.
```

hestonhoffman: Good idea!

documentation/projects.md
hestonhoffman: Cool, I'll remove it from here. I think it's outside the scope of this doc.

PR #1978, opened 2020-07-09
(GH-1964) Add deprecation warning utility

### LINKED ISSUE #1964: Standardize Deprecation warnings
## Use Case
Bolt is moving rapidly and deprecates a feature every month or so we want to 

A standard function for deprecation warnings so that we ensure consistency and add usability features like silencing some warnings.

- This function should wrap "warn_once"
- This function should report an analytics event with a unique deprecation id for each feature.
- This function should allow the issueing of silent deprecations that only send analytics.

## Additional Context
When should we start sending warnings.
- Plan language deprecations should wait until near the major version release so people don't have to update their contents.
- Smaller deprecation should have ASAP since people don't necessarily upgrade frequently.
- Most deprecation warnings aren't seen by most users or are simple. (ie debug -> loglevel)
- experimental features should change as fast as possible so we can validate our ideas.

lucywyman:
Previously, we had no way to track how many users were experiencing
warnings for deprecated features, or which deprecated features were
being used. This commit adds a new class method `deprecation_warning` to
the `Bolt::Logger` class, which sends an analytics event with which type
of warning was issued then calls `warn_once` for the deprecation. This
also changes existing deprecation warnings to use the new deprecation
method, and audits existing warnings to determine if they should use
`warn_once` or `deprecation_warning`.

Closes #1964

!no-release-note

### DIFF
--- a/bolt-modules/boltlib/lib/puppet/datatypes/target.rb
+++ b/bolt-modules/boltlib/lib/puppet/datatypes/target.rb
@@ -1,8 +1,5 @@
 # frozen_string_literal: true
 
-# This was failing local test runs
-require 'bolt/target'
-
 Puppet::DataTypes.create_type('Target') do
   begin
     inventory = Puppet.lookup(:bolt_inventory)
--- a/bolt-modules/boltlib/lib/puppet/datatypes/target.rb
+++ b/bolt-modules/boltlib/lib/puppet/datatypes/target.rb
@@ -1,5 +1,8 @@
 # frozen_string_literal: true
 
+# This was failing local test runs
+require 'bolt/target'
+
 Puppet::DataTypes.create_type('Target') do
   begin
     inventory = Puppet.lookup(:bolt_inventory)
--- a/lib/bolt/cli.rb
+++ b/lib/bolt/cli.rb
@@ -131,6 +131,7 @@ module Bolt
                   end
 
         Bolt::Logger.configure(config.log, config.color)
+        Bolt::Logger.analytics = analytics
       rescue Bolt::Error => e
         if $stdout.isatty
           # Print the error message in red, mimicking outputter.fatal_error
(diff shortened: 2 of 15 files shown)


resources/bolt_bash_completion.sh
```diff
@@ -6,74 +6,134 @@
 ###
 ### Marc Schoechlin ms-github@256bit.org
 
-_bolt()
```
beechtom, line 9: I think you need a rebase since this looks like work from another PR?

lucywyman: I think this is just cause it's against main instead of master. Updated

lib/bolt/logger.rb
```diff
@@ -67,6 +68,10 @@ def self.configure(destinations, color)
       end
     end
 
+    def self.analytics(analytics = Bolt::Analytics::NoopClient.new)
```
nicklewis, line 71: I'm not clear what this method is supposed to do. Is it a getter or a setter or both?

lucywyman: It's meant to just be a setter, but calling it `set_analytics` makes rubocop unhappy. There's probably a better way.

nicklewis: In that case, it should be `def self.analytics=(analytics)`. No default value for the argument then.

lib/bolt/logger.rb
```diff
@@ -67,6 +67,10 @@ def self.configure(destinations, color)
       end
     end
 
+    def self.analytics=(analytics)
+      @analytics ||= analytics
```
nicklewis, line 71: This should just be straight assignment rather than `||=`.

lib/bolt/cli.rb
```diff
@@ -131,6 +131,7 @@ def parse
                   end
 
         Bolt::Logger.configure(config.log, config.color)
+        Bolt::Logger.analytics(analytics)
```
nicklewis, line 134: This needs to be updated to use assignment.

lucywyman: Oh no sorry I know still working on this - I pushed too soon.

bolt-modules/boltlib/lib/puppet/datatypes/target.rb
```diff
@@ -1,5 +1,8 @@
 # frozen_string_literal: true
 
+# This was failing local test runs
+require 'bolt/target'
```
nicklewis, line 4: We should fix the tests that aren't requiring this properly. The target class should come from the inventory which is injected via the puppet context. This type doesn't depend on `Bolt::Target`, which is important for use in PE.

lucywyman: I don't remember which test was failing - I'll just remove this for now.

PR #1979, opened 2020-07-09
(GH-1868) Add support for downloading files from targets

### LINKED ISSUE #1868: Add support for downloading files from targets
## Use Case
As a Bolt user there are times where i need to retrieve a file from a remote host/target and download it onto my Bolt control node. Examples:
- We operate RHEL servers as part of our infrastructure. The RHEL support tool `sosreport` is used on hosts to collect diagnostic info that you upload to your support case. I would love to automate this process with Bolt by initiating a command from my machine and it goes out to a list of hosts, runs `sosreport` downloads the bundles (tar.xz files) and then uploads them to the Red Hat support case.
- Similar use case for Graylog Support, they have a tool called `apollo` that generates support bundles that we would like to automate the collection and upload for.
- I also have a use case where i want to understand what the "current contents" of a file are and diff them between hosts. Example: I have a Production server and a Staging server that are exhibiting different behavior when writing to the database. I want to grab the database configs from both hosts and diff the configs to see if they are the same.
- We want to perform a Database / Application backup of a remote host, then download that backup so we can place it in the proper storage location. In this case it would be a `download` followed by an `upload` most likely, unless we were running Bolt from the storage server. Might also be an interesting use case for a `transfer_file` function. Although, the `download_file` and `upload_file` workflow is perfectly acceptable.

Would love to see a function `download_file()` that works similarly to `upload_file()` except in reverse.

The "hard" part here is answering the question "if i have multiple targets, where do i put the files". I'm open to suggestions on this, but my "gut" tells me to drop them into a directory tree where the directory names correspond to the names of the targets. Example:

```
$ bolt download /var/log/sosreport.tar.gz --targets host1,host2,host3 destination=./redhat-support-12345
.... # Bolt does its thing

$ tree ./redhat-support-12345/
./redhat-support-12345/
|-- host1
|   `-- sosreport.tar.gz
|-- host2
|   `-- sosreport.tar.gz
`-- host3
    `-- sosreport.tar.gz

3 directories, 3 files
```

Could also do something where the full directory path is included in the destination (maybe an option to turn this on/off):
```
$ bolt download /var/log/sosreport.tar.gz --targets host1,host2,host3 destination=./redhat-support-12345 full_path=true
.... # Bolt does its thing

$ tree ./redhat-support-12345/
./redhat-support-12345/
|-- host1
|   `-- var
|       `-- log
|           `-- sosreport.tar.gz
|-- host2
|   `-- var
|       `-- log
|           `-- sosreport.tar.gz
`-- host3
    `-- var
        `-- log
            `-- sosreport.tar.gz

9 directories, 3 files
```

I've thought about writing my own task that outputs the file contents as `base64` and then decode and write the data to a file. This seemed like a really slow and clunky option.

## Additional Context
Looks like Ansible supports this in the form of their [`fetch`](https://docs.ansible.com/ansible/latest/modules/fetch_module.html) and [`slurp`](https://docs.ansible.com/ansible/latest/modules/slurp_module.html) "modules"

beechtom:
This adds support for a `bolt file download` command that downloads a
file or directory from a target to a destination directory on the local
system. File downloads are supported on all transports _except for
`pcp`_.

The `bolt file download` command expects two arguments:

- **source:** The path to the file or directory to be downloaded from
the target
- **destination:** The path to the destination directory on the local
system

The destination directory accepts either an absolute path or a relative
path. If a relative path is given, it expands relative to the project's
`downloads` directory. For example, the default project's downloads
directory is located at `~/.puppetlabs/bolt/downloads/`.

When downloading content, Bolt creates a directory for each target
within the destination directory. The directory for each target has a
name matching the URL-encoded safe name for the target. All content
downloaded for the target is stored in this directory.

!feature

* **Add `bolt file download` CLI command** ([#1868](#1868))

  Users can now download files and directories from targets to the local
  system using the `bolt file download` CLI command. This command
  accepts a path to the file or directory to download from the targets
  and a path to a destination directory on the local system. The
  destination directory is expanded relative to the project downloads
  directory (i.e. `<project>/downloads/`).

---

This adds a new `download_file` plan function which can be used to
download a specified file or directory from a list of targets to the
local system.

Similar to the `bolt file download` CLI command, this plan function
accepts `source` and `destination` arguments. The `source` argument is
the path to the file or directory to download from the targets, while
the `destination` argument is a path to a directory on the local system
that the files should be saved to.

The path to the destination directory is expanded relative to the
project's `downloads` directory (i.e. `<project>/downloads/`).

The result for this plan function includes the `path` to the downloaded
file on the local system.

!feature

* **Add `download_file` plan function** (#1868)

  The `download_file` plan function can be used to download a file or
  directory from a list of targets to a destination directory on the
  local system. The result returned from this function includes the path
  to the downloaded file on the local system.

---

This adds a new YAML plan step for Bolt's file download functionality.
The step uses the following keys:

- `download`: The path to the file or directory to download from the
targets
- `destination`: The path to the destination directory on the local
system
- `targets`: The list of targets to download from

This also updates the YAML plan upload step to use an `upload` key in
place of a `source` key. Previously, the upload step used a `source`
key to identify the step as an upload step. This ambiguous key would
have been confusing with the addition of a download step, as it also
expects a source path.

The `source` key for YAML plan upload steps has been deprecated. Bolt
will raise a helpful warning when a plan step includes a `source` key.

!feature

* **Add YAML plan download step** ([#1868](#1868))

  YAML plans now support a download step which can be used to download a
  file or directory from a list of targets to a destination directory on
  the local system.

!deprecation

* **Deprecate `source` key for YAML plan upload step** ([#1868](#1868))

  The `source` key used in YAML plan upload steps has been deprecated in
  favor of a less ambiguous `upload` key.

---

This adds `allow_download` and `expect_download` stubs to BoltSpec.

!feature

* **Add `allow_download` and `expect_download` stubs to BoltSpec** ([#1868](#1868))

  Users can use the `allow_download` and `expect_download` stubs to test
  plans that contain calls to `download_file`.

Closes #1868


documentation/running_bolt_commands.md
```diff
@@ -128,6 +128,56 @@ best limited to small files.
...
+directory](bolt_project_directories.md), for example
+`~/.puppetlabs/bolt/downloads`.
+
+Each file or directory is saved to the destination directory under a
+directory with a name matching the URL-encoded name of the target it
+was downloaded from. The target directory names are URL-encoded to ensure
+that they are valid directory names.
+
+For example, running the following command will download the SSH daemon
+configuration file from each target:
```
hestonhoffman, line 156:
```suggestion
For example, the following command downloads the SSH daemon configuration file from two targets, `linux` and `ssh://example.com`:
```

documentation/running_bolt_commands.md
```diff
@@ -128,6 +128,56 @@ best limited to small files.
...
+
+```shell
+$ bolt file download /etc/ssh/sshd_config sshd_config --targets linux,ssh://example.com
+```
+
+After running this command, your project directory structure would look
+like:
```
hestonhoffman, line 163:
```suggestion
like this:
```

documentation/running_bolt_commands.md
```diff
@@ -128,6 +128,56 @@ best limited to small files.
...
+└── downloads/
+    └── sshd_config/
+        ├── linux/
+        │   └── sshd_config
+        └── ssh%3A%2F%2Fexample.com/
+            └── sshd_config
+```
+
+> 🔩 **Tip:** To avoid URL encoding the target's safe name, give the target a
+> simple, human-readable name in your inventory file.
```
hestonhoffman, line 179: This is a great tip!

documentation/running_bolt_commands.md
```diff
@@ -128,6 +128,56 @@ best limited to small files.
...
+To download a file or directory from a remote target, run the
+`bolt file download` command. Specify the remote path to the file or
+directory, the destination directory on your local system, and the targets.
+
+```shell
+$ bolt file download <SOURCE> <DESTINATION> --targets <TARGETS>
+```
+
+The `destination` can be either an absolute or relative path to a directory
+on your local system. If a relative path is given, the path is expanded
```
hestonhoffman, line 145: Just double-checking something: `downloads` is only used when you use a relative path? So if I gave an absolute path, I could specify any directory name and the result would not be nested under a `downloads` directory?

documentation/running_bolt_commands.md
```diff
@@ -128,6 +128,56 @@ best limited to small files.
...
+that they are valid directory names.
+
+For example, running the following command will download the SSH daemon
+configuration file from each target:
+
+```shell
+$ bolt file download /etc/ssh/sshd_config sshd_config --targets linux,ssh://example.com
+```
+
+After running this command, your project directory structure would look
```
hestonhoffman, line 162:
```suggestion
After running this command from the root of your project directory, your project directory structure would look
```

documentation/running_bolt_commands.md
beechtom: Correct

hestonhoffman: Okay cool, I think this bit is good then 👍

bolt-modules/boltlib/lib/puppet/functions/download_file.rb
```diff
@@ -0,0 +1,104 @@
...
+  # @param description A description to be output when calling this function.
+  # @param options A hash of additional options.
+  # @option options [Boolean] _catch_errors Whether to catch raised errors.
+  # @option options [String] _run_as User to run as using privilege escalation.
+  # @return A list of results, one entry per target, with the path to the downloaded file under the
+  #         `path` key.
+  # @example Download a file from multiple Linux targets to a destination directory
+  #   download_file('/etc/ssh/ssh_config', '~/Downloads', $targets, 'Downloading remote SSH config')
+  dispatch :download_file_with_description do
+    scope_param
```
lucywyman, line 57: I don't see where this gets used?

lib/bolt/transport/ssh/connection.rb
```diff
@@ -235,14 +235,22 @@ def execute(command_str)
...
           # Do not log wrapper script content
           @logger.debug { "Uploading #{source}, to #{destination}" } unless source.is_a?(StringIO)
           @session.scp.upload!(source, destination, recursive: true)
         rescue StandardError => e
           raise Bolt::Node::FileError.new(e.message, 'WRITE_ERROR')
         end
 
+        def download_file(source, destination)
+          # Do not log wrapper script content
+          @logger.debug { "Downloading #{source} to #{destination}" } unless destination.is_a?(StringIO)
```
nicklewis, line 248: This will never be a `StringIO`.

lib/bolt/pal/yaml_plan/evaluator.rb
```diff
@@ -140,8 +151,14 @@ def apply_manifest(scope, targets, manifest)
...
         def evaluate_block_with_bindings(closure_scope, args_hash, plan)
           if plan.steps.any? { |step| step.body.key?('target') }
-            msg = "The 'target' parameter for YAML plan steps is deprecated and will be removed "\
-                  "in a future version of Bolt. Use the 'targets' parameter instead."
+            msg = "The 'target' key for YAML plan steps is deprecated and will be removed "\
+                  "in a future version of Bolt. Use the 'targets' key instead."
+            @logger.warn(msg)
+          end
+
+          if plan.steps.any? { |step| step.body.key?('source') }
```
nicklewis, line 159: These should use the new `deprecation_warning` utility

lib/bolt/cli.rb
```diff
@@ -431,11 +431,22 @@ def execute(options)
...
+
               if dest.nil?
                 raise Bolt::CLIError, "A destination path must be specified"
               end
-              validate_file('source file', src, true)
-              executor.upload_file(targets, src, dest, executor_opts)
+
+              case options[:action]
+              when 'download'
+                dest = File.expand_path(dest, config.project.downloads)
```
nicklewis, line 444: This one should be relative to cwd.

bolt-modules/boltlib/lib/puppet/functions/download_file.rb
```diff
@@ -0,0 +1,104 @@
...
+    options = options.select { |opt| opt.start_with?('_') }.transform_keys { |k| k.sub(/^_/, '').to_sym }
+    options[:description] = description if description
+
+    executor = Puppet.lookup(:bolt_executor)
+    inventory = Puppet.lookup(:bolt_inventory)
+
+    # Paths should expand relative to the default downloads directory for the project
+    # e.g. ~/.puppetlabs/bolt/downloads/
+    project_downloads = Puppet.lookup(:bolt_project_data)&.downloads
+    destination = File.expand_path(destination, project_downloads)
```
nicklewis, line 85: We should validate that the destination doesn't have any directory traversal going on. It may be better to restrict it to a specific set of characters (like a download "key" rather than a path).

I'm on the fence about whether we should allow it to be nested. On the one hand, that seems like it opens up even more issues around directory traversal. But it might also be helpful if you have multiple downloads in a plan (`myplan/download1`, `myplan/download2`)... Maybe for now it's best to not allow that and we can open it up later.

bolt-modules/boltlib/lib/puppet/functions/download_file.rb
beechtom: Ah, whoops.

lib/bolt/bolt_option_parser.rb
```diff
@@ -218,10 +221,30 @@ def get_help_text(subcommand, action = nil)
           bolt file <action> [options]
 
       DESCRIPTION
-          Upload a local file or directory
+          Copy files and directories between the host and targets
```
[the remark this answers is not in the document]
beechtom: Controller seems to be the most common term we use.

lib/bolt/executor.rb
```diff
@@ -136,7 +136,7 @@ def queue_execute(targets)
               # by the CLI. Log a warning, as this is probably a problem with the transport.
               # If batch_* commands are used from the Base transport, then exceptions
               # normally shouldn't reach here.
-              @logger.warn(e)
+              @logger.warn(e) unless e.message =~ /pcp transport/
```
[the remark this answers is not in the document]
beechtom: Raising an error from `Bolt::Transport::Orch#batch_download` will be rescued here and log a warning with a backtrace:

```
<Bolt::Error> pcp transport does not support downloading files
	/Users/tom/git/bolt/lib/bolt/transport/orch.rb:188:in `batch_download'
	/Users/tom/git/bolt/lib/bolt/executor.rb:331:in `block (3 levels) in download_file'
	/Users/tom/git/bolt/lib/bolt/executor.rb:249:in `with_node_logging'
	/Users/tom/git/bolt/lib/bolt/executor.rb:330:in `block (2 levels) in download_file'
	/Users/tom/git/bolt/lib/bolt/executor.rb:128:in `block (3 levels) in queue_execute'
	/Users/tom/.rvm/gems/ruby-2.6.3/gems/concurrent-ruby-1.1.6/lib/concurrent-ruby/concurrent/executor/ruby_thread_pool_executor.rb:353:in `run_task'
	/Users/tom/.rvm/gems/ruby-2.6.3/gems/concurrent-ruby-1.1.6/lib/concurrent-ruby/concurrent/executor/ruby_thread_pool_executor.rb:342:in `block (3 levels) in create_worker'
	/Users/tom/.rvm/gems/ruby-2.6.3/gems/concurrent-ruby-1.1.6/lib/concurrent-ruby/concurrent/executor/ruby_thread_pool_executor.rb:325:in `loop'
	/Users/tom/.rvm/gems/ruby-2.6.3/gems/concurrent-ruby-1.1.6/lib/concurrent-ruby/concurrent/executor/ruby_thread_pool_executor.rb:325:in `block (2 levels) in create_worker'
	/Users/tom/.rvm/gems/ruby-2.6.3/gems/concurrent-ruby-1.1.6/lib/concurrent-ruby/concurrent/executor/ruby_thread_pool_executor.rb:324:in `catch'
	/Users/tom/.rvm/gems/ruby-2.6.3/gems/concurrent-ruby-1.1.6/lib/concurrent-ruby/concurrent/executor/ruby_thread_pool_executor.rb:324:in `block in create_worker'
	/Users/tom/.rvm/gems/ruby-2.6.3/gems/logging-2.3.0/lib/logging/diagnostic_context.rb:474:in `block in create_with_logging_context'
Started on fedora29...
Finished on fedora29:
  Downloaded '/etc/ssh/ssh_config' to '/Users/tom/git/bolt/ssh_config/fedora29'
  {
    "path": "/Users/tom/git/bolt/ssh_config/fedora29/ssh_config"
  }
Successful on 1 target: fedora29
Failed on 1 target: ubuntu1804
Ran on 2 targets in 0.73 sec
```

bolt-modules/boltlib/lib/puppet/functions/download_file.rb
beechtom: Just to make sure I understand, by no directory traversal you mean the destination is not something nested more than 1 level deep inside `downloads`? I.e. `foo` as a destination is valid, but `foo/bar` is not?

If that's the case, do we not want to permit absolute paths for a destination and force plans to always download to `<project>/downloads`?

lib/bolt/pal/yaml_plan/step/upload.rb
```diff
@@ -6,7 +6,7 @@ class YamlPlan
       class Step
         class Upload < Step
           def self.allowed_keys
-            super + Set['source', 'destination']
+            super + Set['source', 'destination', 'upload']
           end
 
           def self.required_keys
```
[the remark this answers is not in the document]
beechtom: No, https://github.com/puppetlabs/bolt/pull/1979/files#diff-ec96a76d58680201f59e7377e9d96cfaR96-R98 checks if the `upload` key is present and removes `source` as a required key if it is. This is similar to how we handle yaml plans that use the deprecated `target` key instead of `targets`.

I do think I got this a bit backwards by leaving `source` as the required key. Updated to have `upload` required instead.

beechtom:
Updated to delete existing contents in the destination directory before downloading.

bolt-modules/boltlib/lib/puppet/functions/download_file.rb
nicklewis: I was mostly thinking that it shouldn't contain characters like `..` to be able to "escape" the `downloads` directory. I'm not sure about the nested case.

I don't think absolute paths make a whole lot of sense though, given that we don't exactly download files to the specified directory (but instead to our own structure). It might be better to hold off on absolute paths for now, we can always add them later if needed.

beechtom: Ah, that makes sense. :) I'll add a check for that.

lib/bolt/executor.rb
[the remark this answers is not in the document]
nicklewis: Is there a clearer way to handle this? Regex match on the message of an exception to decide whether to log it seems brittle.

nicklewis:
I agree with your concern about deleting files from the CLI. It might be better to do that in the Puppet function rather than the executor. That way we also know that the deletion is coupled with whatever validation we do in that function, unlike the executor which can be called from many places. We need to be very careful that the destination isn't an empty string, etc.

lib/bolt/executor.rb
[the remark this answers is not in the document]
beechtom: Possibly raising a specific type of error that we can recognize and not log? The logic being that if we have a specific error type that indicates it's an intentional error then it doesn't need to be logged? Added a `Bolt::NotSupportedError` class for this, but open to other approaches.

bolt-modules/boltlib/lib/puppet/functions/download_file.rb
beechtom: Updated.

lib/bolt/executor.rb
```diff
@@ -133,10 +133,12 @@ def queue_execute(targets)
...
             result_promises.each do |target, promise|
               # If an exception happens while running, the result won't be logged
-              # by the CLI. Log a warning, as this is probably a problem with the transport.
+              # by the CLI. Unless the error is explicitly raised because a feature is not
+              # supported by the transport, log a warning as this is probably a problem
+              # with the transport.
               # If batch_* commands are used from the Base transport, then exceptions
               # normally shouldn't reach here.
-              @logger.warn(e)
+              @logger.warn(e) unless e.is_a?(Bolt::NotSupportedError)
```
nicklewis, line 141: Hmm actually I think the right thing to do is to make the Orchestrator transport return `Bolt::Result` objects that indicate the problem, rather than throwing an exception. That matches the behavior provided by the default implementation of the `batch_*` methods in `Bolt::Transport::Base#with_events`. Then this top level handler is really just for unhandled exceptions which probably _are_ a problem with the transport, as the comment above says.

That establishes/maintains a contract that the result of calling a `batch_*` method on a transport is a list of `Bolt::Result` objects.

bolt-modules/boltlib/lib/puppet/functions/download_file.rb
```diff
@@ -0,0 +1,117 @@
...
+    end
+
+    # Prevent path traversal so downloads can't be saved outside of the project downloads directory
+    if (destination.each_filename.to_a & %w[. ..]).any?
+      raise Bolt::ValidationError, "Destination must not include path traversal, received #{destination}"
+    end
+
+    # Paths expand relative to the default downloads directory for the project
+    # e.g. ~/.puppetlabs/bolt/downloads/
+    destination = Puppet.lookup(:bolt_project_data).downloads + destination
```
nicklewis, line 98: It might be good to have a test that a destination starting with `~` doesn't get expanded to the home dir (it should be appended literally to `downloads`). It actually looks like some tests currently do call it with that, but then don't verify what happens.

nicklewis:
`bolt file download 'C:\Users\Administrator' test -t <target>` results in a directory with the name `C:\Users\Administrator`. That definitely seems strange.

beechtom:
Updated to sub `\` with `/` in source paths. Ruby doesn't recognize `\` as a path separator on *nix, but does recognize `/` as a path separator on all systems.

nicklewis:
What happens then for paths that are _supposed_ to have a `\` on Linux where it's not a path separator?

documentation/running_bolt_commands.md
```diff
@@ -128,6 +128,55 @@ best limited to small files.
...
+directory, the destination directory on your local system, and the targets.
+
+```shell
+$ bolt file download <SOURCE> <DESTINATION> --targets <TARGETS>
+```
+
+The `destination` can be either an absolute or relative path to a directory
+on your local system. If a relative path is given, the path is expanded
+relative to the current working directory. If the destination directory does
+not exist, Bolt will automatically create it.
```
hestonhoffman, line 147:
```suggestion
on your local system. If you use a relative path, Bolt expands the path
relative to the current working directory. If the destination directory does
not exist, Bolt will automatically create it.
```

PR #1984, opened 2020-07-13
(maint) Don't expose Pathname object to Puppet

nicklewis:
Previously, when passing the Bolt project to Puppet to turn into a
module, we were exposing the "path" variable as a Pathname instance.
Puppet expects the path will always be a string and fails in at least
one case if it receives another value. Since we just use Pathname for
our own internal tracking, we now turn it into a string in the `to_h`
method for Puppet's consumption.

!bug

  * **Don't fail for tasks referring to project-level files**

    Tasks with a `files` key that refers to files that exist at the
    project level will now load properly rather than throwing an error.

### DIFF
--- a/lib/bolt/project.rb
+++ b/lib/bolt/project.rb
@@ -109,7 +109,7 @@ module Bolt
     # This API is used to prepend the project as a module to Puppet's internal
     # module_references list. CHANGE AT YOUR OWN RISK
     def to_h
-      { path: @path, name: name }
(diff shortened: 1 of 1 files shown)


lucywyman:
Should this include a test that would have caught the bug?

nicklewis:
I'm not sure. I think the only proper way to test it would be to test that every kind of task that works normally also works as a project-level task and I don't think there's much surface area for that to break aside from this one specific issue...

Issue #1973, opened 2020-07-08
Update known issues to link to GH issues


lucywyman:
## Describe the Change You Would Like
Known issues should link to open GH issues with detailed descriptions. If a known issue doesn't already have an open issue, add one. If it does but the issue not descriptive, update the description

PR #1990, opened 2020-07-13
(GH-1973) Add GH issue links to 'Known issues' doc

beechtom:
This adds links to GitHub issues for each of the issues listed on the
'Known issues' doc. It also updates several of the issues to provide
more context on the issue and clearer instructions on any applicable
workarounds.

!no-release-note

### DIFF
--- a/documentation/bolt_known_issues.md
+++ b/documentation/bolt_known_issues.md
@@ -1,39 +1,59 @@
 # Known issues
 
-Known issues for the Bolt 1.x release series.
-
 ## Tasks executed with PowerShell version 2.x or earlier cannot use parameters named `type`
 
-When executing PowerShell tasks on targets using a PowerShell interpreter
-version 2.x or earlier, you cannot use a task parameter with the name `type`.
-Bolt versions 1.30.0 and earlier contained a
-[bug](https://github.com/puppetlabs/bolt/issues/1205) that made parameters with
-the string `type` in their name (for example, `serverType`) incompatible.
-Starting with Bolt version 1.31.0, only PowerShell parameters with `type` as
-their complete name are incompatible. For PowerShell version 3 and later, any
-parameter names are permissible.
+When executing PowerShell tasks on targets using PowerShell version 2.x or
+earlier, you cannot use a task parameter with the name `type`. Because
+PowerShell version 2.x and earlier do not support `type` as a named argument,
+and PowerShell tasks convert parameters to named arguments, Bolt will filter out
+a `type` parameter before running the task.
+
+When running PowerShell tasks on targets using PowerShell version 3.0 or later,
+any parameter name is permissible.
+
+📖 **Related issues**
+
+- [#1988 - Tasks executed with PowerShell version 2.x or earlier cannot use
+  parameters named `type`](https://github.com/puppetlabs/bolt/issues/1988)
 
 ## JSON strings as command arguments might require additional escaping in PowerShell
 
 When passing complex arguments to tasks with `--params`, JSON strings (typically
 created with the `ConvertTo-Json` cmdlet) might require additional escaping. In
 some cases, you can use the PowerShell stop parsing symbol `--%` as a
-workaround. ([BOLT-1130](https://tickets.puppetlabs.com/browse/BOLT-1130))
+workaround.
+
+📖 **Related issues**
+
+- [#1985 - Bolt PowerShell wrapper should allow for the use of `Convert-ToJson`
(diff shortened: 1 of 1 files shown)


documentation/bolt_known_issues.md
```diff
@@ -1,39 +1,65 @@
 # Known issues
 
-Known issues for the Bolt 1.x release series.
-
 ## Tasks executed with PowerShell version 2.x or earlier cannot use parameters named `type`
 
-When executing PowerShell tasks on targets using a PowerShell interpreter
-version 2.x or earlier, you cannot use a task parameter with the name `type`.
+When executing PowerShell tasks on targets using PowerShell  version 2.x or
```
lucywyman, line 5:
```suggestion
When executing PowerShell tasks on targets using PowerShell version 2.x or
```

documentation/bolt_known_issues.md
```diff
@@ -42,14 +68,33 @@ unsupported key type `ssh-ed25519'
   * bcrypt_pbkdf (>= 1.0, < 2.0)
 ```
 
+A workaround is to use native SSH when you need to authenticate with
+ed25519 keys. When native SSH is enabled, Bolt will use a specified SSH
```
lucywyman, line 72: Does `ed25519` work if users manually install the gems in bolt's environment? That might be easier than using native ssh especially on Windows, where I'm not sure people will just know where ssh executable is installed.

documentation/bolt_known_issues.md
```diff
@@ -42,14 +68,33 @@ unsupported key type `ssh-ed25519'
...
+📖 **Related issues**
+
+- [#1987 - Unable to authenticate with ed25519 keys over SSH transport
+  on Windows](https://github.com/puppetlabs/bolt/issues/1987)
+
 ## Limited Kerberos support
 
-Support for Kerberos over WinRM from a Linux host is currently experimental and
-requires the [MIT Kerberos
+Authenticating with Kerberos over SSH is fully supported.
```
lucywyman, line 84: Could probably just retitle the issue "Limited Kerberos support over WinRM" ?

documentation/bolt_known_issues.md
beechtom: It looks like installing from Bolt's gem environment won't work since it runs into the same issues with building native extensions.

```
C:\Users\Administrator>"C:\Program Files\Puppet Labs\Bolt\bin\gem.bat" install --user-install ed25519
WARNING:  You don't have c:\users\administrator\.gem\ruby\2.5.0\bin in your PATH,
          gem executables will not run.
Building native extensions. This could take a while...
ERROR:  Error installing ed25519:
        ERROR: Failed to build gem native extension.

    current directory: C:/Users/Administrator/.gem/ruby/2.5.0/gems/ed25519-1.2.4/ext/ed25519_ref10
"C:/Program Files/Puppet Labs/Bolt/bin/ruby.exe" -r ./siteconf20200714-2460-14q9lrl.rb extconf.rb
creating Makefile
*** extconf.rb failed ***
Could not create Makefile due to some reason, probably lack of necessary
libraries and/or headers.  Check the mkmf.log file for more details.  You may
need configuration options.
```

Using native SSH shouldn't be difficult on Windows for any users who have OpenSSH installed, since the default commands Bolt uses (`ssh` and `scp -r`) are both valid.

documentation/bolt_known_issues.md
```diff
@@ -1,39 +1,65 @@
...
-
 ## Tasks executed with PowerShell version 2.x or earlier cannot use parameters named `type`
 
-When executing PowerShell tasks on targets using a PowerShell interpreter
-version 2.x or earlier, you cannot use a task parameter with the name `type`.
+When executing PowerShell tasks on targets using PowerShell version 2.x or
+earlier, you cannot use a task parameter with the name `type`. Since PowerShell
+version 2.x and earlier do not support `type` as a named argument, and
+PowerShell tasks convert parameters to named arguments, a `type` parameter will
+be filtered out before running the task.
```
hestonhoffman, line 9:
```suggestion
When executing PowerShell tasks on targets using PowerShell version 2.x or
earlier, you cannot use a task parameter with the name `type`. Because PowerShell
version 2.x and earlier do not support `type` as a named argument, and
PowerShell tasks convert parameters to named arguments, Bolt will filter out a `type` 
parameter before running the task.
```
Style tip: "Since" is fine grammatically, but we tend to avoid it because it's ambiguous for non native speakers.

beechtom: Good to know!

Issue #1994, opened 2020-07-14
Make http timeouts configurable


m0dular:
## Use Case
Currently there are some [hardcoded timeout values](https://github.com/nahi/httpclient/blob/ddd141b30848a8c4df2dfa84602f80b5b8a5f362/lib/httpclient/session.rb#L133-L139) in the HttpClient initialization used by Bolt.  This can cause timeouts when a `puppetdb_query()` calls takes longer than 60 seconds to return, among other situations.  The connect and read timeouts should be configurable in the `puppetdb` hash in a user's Bolt configuration, or whichever way makes the most sense.

### Solution
This should be 2 new settings under `puppetdb` config section: `connect-timeout` and `read-timeout` that correspond to connect-timeout and receive-timeout in HttpClient respectively. They take integers.

PR #2002, opened 2020-07-17
(GH-1994) Configure timeouts in PuppetDB client

beechtom:
This adds support for a `connect_timeout` and `read_timeout` option
in `puppetdb` config. The `connect_timeout` value is used to set the
connection timeout for the HTTP client that connects to PuppetDB, while
`read_timeout` is used to set the receive timeout for the HTTP client.
The value for each of these options must be a positive integer. If the
options are not set, both are set to the default values defined by the
`httpclient` Ruby library (currently 60 seconds).

!feature

* **Configure connection and read timeout length in PuppetDB client**
  ([#1994](#1994))

  Users can now configure the connection and read timeout length for the
  PuppetDB client with the `connect_timeout` and `read_timeout`
  options under the `puppetdb` config option.

### DIFF
--- a/documentation/bolt_connect_puppetdb.md
+++ b/documentation/bolt_connect_puppetdb.md
@@ -50,6 +50,9 @@ config](configuring_bolt.md) with the following values:
     the protocol `https` and the port, which is usually `8081`. For example,
     `https://my-master.example.com:8081`.
 -   `cacert`: The path to the ca certificate for PuppetDB.
+-   `connect_timeout`: How long to wait in seconds when establishing
+    connections with PuppetDB.
+-   `read_timeout`: How long to wait in seconds for a response from PuppetDB.
 
 If you are using certificate authentication also set:
 
--- a/spec/bolt/puppetdb/client_spec.rb
+++ b/spec/bolt/puppetdb/client_spec.rb
@@ -3,12 +3,58 @@
 require 'spec_helper'
 require 'bolt/puppetdb/client'
 require 'bolt_spec/puppetdb'
+require 'httpclient'
 
 describe Bolt::PuppetDB::Client do
   let(:uri) { 'https://puppetdb:8081' }
   let(:cacert) { File.expand_path('/path/to/cacert') }
-  let(:config) { double('config', server_urls: [uri], cacert: cacert, token: nil, cert: nil, key: nil) }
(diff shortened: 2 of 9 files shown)


nicklewis:
Why `server_url_timeout` rather than `connect_timeout`?

beechtom:
`puppetdb.conf` uses `server_url_timeout` to set the timeout for connecting to PuppetDB. Since we load `puppetdb.conf`, it seemed appropriate to keep the name of the setting consistent.

nicklewis:
Is that in the `pupppetdb.conf` used by the `puppet query` command? I don't see any reference to it in the source code. Or is it for `/etc/puppetlabs/puppet/puppetdb.conf` which is used by Puppet to talk to PuppetDB (and is _not_ the file we load in 
Bolt)?

beechtom:
Ah, I guess it'd be the former, but it looks like we do load the latter when `ENV['HOME']` is `nil`. Regardless, I'll just change it to `connect_timeout`.

lucywyman:
The commit message + PR description need to be updated with `connect_timeout` as well.

documentation/bolt_connect_puppetdb.md
```diff
@@ -50,6 +50,9 @@ config](configuring_bolt.md) with the following values:
     the protocol `https` and the port, which is usually `8081`. For example,
     `https://my-master.example.com:8081`.
 -   `cacert`: The path to the ca certificate for PuppetDB.
+-   `server_url_timeout`: How long to wait in seconds when establishing
```
lucywyman, line 53: This also needs to be updated

spec/bolt/puppetdb/client_spec.rb
```diff
@@ -3,12 +3,61 @@
...
+      token: nil,
+      cert: nil,
+      key: nil,
+      connect_timeout: nil,
+      read_timeout: nil
+    )
+  end
+
+  describe "#http_client" do
+    let(:http_client) { double('http_client') }
```
lucywyman, line 29: This is just a nifty little short-hand if you want to skip some of the allows:
```
    let(:ssl_config)  { double('ssl_config', set_client_cert_file: nil, add_trust_ca: nil) }
    let(:http_client) { double('httpclient', ssl_config: ssl_config) }
```

PR #2005, opened 2020-07-20
(bug) Do not re-rescue errors raised before config is loaded

beechtom:
This fixes a bug introduced in #1917 that caused errors raised before
config is loaded to be re-raised and rescued by `Bolt::CLI#parse` and
sent to the outputter. This resulted in a `NoMethodError` being raised,
as the outputter relies on the config being loaded when it is
instantiated.

This seperates the logic that was in the `#parse` method into three
separate methods. The `#parse_command` method handles parsing and
validating the command, the `#load_config` method loads the project and
config, and the `#finalize_setup` method finishes configuring Bolt and
issuing any warnings from the previous steps.

Splitting the logic into separate methods allows us to handle errors
raised before and after config is loaded differently.

!bug

* **Do not re-rescue errors raised before config is loaded**

  Errors raised before Bolt's configuration was loaded were raising a
  second, ambiguous `NoMethodError`. Bolt now correctly handles errors
  raised before configuration is loaded and will no longer trigger
  additional errors.

### DIFF
--- a/lib/bolt/cli.rb
+++ b/lib/bolt/cli.rb
@@ -75,72 +75,100 @@ module Bolt
     end
     private :help?
 
+    # Wrapper method that is called by the Bolt executable. Parses the command and
+    # then loads the project and config. Once config is loaded, it completes the
+    # setup process by configuring Bolt and issuing warnings.
+    #
+    # This separation is needed since the Bolt::Outputter class that normally handles
+    # printing errors relies on config being loaded. All setup that happens before
+    # config is loaded will have errors printed directly to stdout, while all errors
+    # raised after config is loaded are handled by the outputter.
     def parse
-      begin
-        parser = BoltOptionParser.new(options)
-        # This part aims to handle both `bolt <mode> --help` and `bolt help <mode>`.
-        remaining = handle_parser_errors { parser.permute(@argv) } unless @argv.empty?
-        if @argv.empty? || help?(remaining)
-          # Update the parser for the subcommand (or lack thereof)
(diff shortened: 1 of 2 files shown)


nicklewis:
These method names don't really help to understand why this was the cut point to divide the two steps. Also the name `parse_command_and_load_project` definitely screams that this method could itself be broken into two pieces.

It seems like we basically have:
a) parse the CLI
b) load config
c) validate and issue warnings

There are a couple of things in `finalize_setup` that could be moved to either `execute` or the CLI parsing step. Setting the 
default for `options[:verbose]` and calling `update_targets`. Maybe also `warn_inventory_overrides_cli`.

lib/bolt/cli.rb
```diff
@@ -75,72 +75,99 @@ def help?(remaining)
...
-                    project = if options[:boltdir]
-                                dir = Pathname.new(options[:boltdir])
-                                if (dir + Bolt::Project::BOLTDIR_NAME).directory?
-                                  Bolt::Project.create_project(dir + Bolt::Project::BOLTDIR_NAME)
-                                else
-                                  Bolt::Project.create_project(dir)
-                                end
+      validate(options)
+
+      # The parser is needed by #finalize_setup to print deprecation warnings
```
lucywyman, line 134: It might be better to collect these into a `deprecation_warnings` instance variable instead of passing the whole parser object around.

lucywyman:
I updated the PR comment to match the commit message. This LGTM!

PR #2009, opened 2020-07-21
(maint) Configure new default cops for Rubocop 0.88

beechtom:
This adds configuration for several new default cops introduced in
Rubocop 0.88. It also enables the `Style/BisectedAttrAccessor` cop,
which was previously disabled due to a bug.

!no-release-note

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -90,13 +90,25 @@ Style/RedundantRegexpEscape:
 Style/AccessorGrouping:
   Enabled: true
 
-# Disabled due to a bug in Rubocop that raises an exception when accessors
-# are assigned using a splat operator
-# https://github.com/rubocop-hq/rubocop/issues/8257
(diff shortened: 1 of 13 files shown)


.rubocop.yml
```diff
@@ -90,15 +90,27 @@ Style/RedundantRegexpEscape:
...
 Style/RedundantAssignment:
   Enabled: true
 
+Style/ArrayCoercion:
+  Enabled: true
+
+Style/CaseLikeIf:
+  Enabled: true
+
+Style/HashAsLastArrayItem:
```
lucywyman, line 105: I wouldn't mind disabling this one, though I don't feel super strongly. Mostly we use this syntax quite a bit in plan language examples, but at the same time the braces do make the code more readable.

beechtom: Disabled 👍

PR #2010, opened 2020-07-21
(docs) Correct Powershell version command

lucywyman:
This corrects a line on our installing Bolt documentation page for
checking the user's Powershell version from `$PSTableVersion` to
`$PSVersionTable`.

!no-release-note

### DIFF
--- a/documentation/bolt_installing.md
+++ b/documentation/bolt_installing.md
@@ -84,8 +84,8 @@ choco uninstall puppet-bolt
 
 PowerShell versions 2.0 and 3.0 cannot automatically discover and load the Bolt
 module, so you'll need to add it manually. Unless your system dates from 2013 or
-earlier, this situation probably does not apply to you. To confirm your version,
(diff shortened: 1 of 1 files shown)


documentation/bolt_installing.md
```diff
@@ -85,7 +85,7 @@ choco uninstall puppet-bolt
 PowerShell versions 2.0 and 3.0 cannot automatically discover and load the Bolt
 module, so you'll need to add it manually. Unless your system dates from 2013 or
 earlier, this situation probably does not apply to you. To confirm your version,
-run `echo $PSTableVersion` in PowerShell.
+run `echo $PSVersionTable` in PowerShell.
```
beechtom, line 88:
```suggestion
run `echo $PSVersionTable` in PowerShell.
```

It looks like you don't need to actually echo `$PSVersionTable` since it's an automatic variable that will print the info on its own.

Issue #1835, opened 2020-05-20
Don't fail when looking up data outside an apply block from a hiera.yaml that includes interpolations


magnuslarsen:
Hi! Thanks for including the `lookup()` function, for outside apply blocks!

I am however stopped by this error, upon using it:
```json
{
  "kind": "bolt/pal-error",
  "msg": "Interpolations are not supported in lookups outside of an apply block: Undefined variable 'domain' (file: /home/user/git/control-repo/hiera.yaml, line: 10)",
  "details": {
  }
}
```
It's not uncommon to use variable expansion in Hiera config, so I think it should be supported fully, or at very least be ignored, so a lookup works, if a yaml file was specified later in the hierarchy without variable expansion (e.g. `common.yaml`), and containing the key in the `lookup('key')`

Tested with Bolt 2.10.0

lucywyman:
A few options:
- automatically make plan params available to interpolations. Nested plans?
- Provide a means for specifying interpolation variables (config? CLI?)
- Lookup from special env vars

- This can't be in the same hierarchy as within apply blocks because `certname` doesn't have meaning outside of a per-target block, ie. apply
- Use case: I want to use data in dev/prod environment and want to use hiera to look it up because that's what I do elsewhere
- For the above use case specifically, just *having* a separate hierarchy (from apply) may be sufficient

First iteration:
- Create a new hierarchy separate from the apply hierarchy
  - Hierarchies are currently per-project
  - This should be a separate key in `hiera.yaml` called `plan-hierarchy` which is data to be used in the plan. This should share defaults such as `version` and `datadir` with the general `hierarchy`
- Getting started with Hiera document that describes using Hiera with Bolt, if you don't use Hiera or Puppet.

benridley:
@lucywyman Does this addition have a corresponding fix in regular Puppet? We've noticed after adding a plan_hierarchy to our hiera.yaml that our Puppet Server will no longer compile, saying hiera.yaml has an unknown key. We're happy to plan an upgrade to our PE installation, but just want to confirm this is in? Cheers!

benridley:
Ahh all good, I found it in https://puppet.com/docs/pe/2019.8/release_notes_pe.html. 

Time to upgrade :)

PR #2020, opened 2020-07-23
(GH-1835) Add tests for plan_hierarchy lookups

lucywyman:
Previously, Bolt would use the same Hiera hierarchy for lookups both
inside apply blocks and outside apply blocks, and simply removed facts
and vars from the PAL ScriptCompiler to prevent users from using
interpolations in their hierarchies for lookups outside apply blocks.
This meant that hierarchies that used interpolations would error even if
the user wasn't looking up data that included an interpolation. To
mitigate this, we added a separate hierarchy under the key
`plan_hierarchy` in puppetlabs/puppet#8232. This PR adds integration
level tests to verify that the plan hierarchy is loaded and used
correctly, and marks them as pending since this feature won't be
available in Bolt until Puppet 6.18 ships.

!no-release-note

### DIFF
--- a/spec/integration/lookup_spec.rb
+++ b/spec/integration/lookup_spec.rb
@@ -5,8 +5,8 @@ require 'bolt_spec/config'
 require 'bolt_spec/integration'
 
 describe "lookup() in plans" do
-  include BoltSpec::Integration
   include BoltSpec::Config
+  include BoltSpec::Integration
 
   after(:each) { Puppet.settings.send(:clear_everything_for_tests) }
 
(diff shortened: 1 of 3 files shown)


spec/integration/lookup_spec.rb
```diff
@@ -134,4 +134,32 @@
...
+    end
+  end
+
+  context 'with invalid plan_hierarchy' do
+    let(:hiera_config) { File.join(boltdir, 'plan_hiera_interpolations.yaml') }
+    let(:plan)         { 'test::plan_lookup' }
+    let(:uri)          { 'localhost' }
+
+    it 'raises a validation error' do
+      pending('Until Puppet 6.18 is released')
```
donoghuc, line 157: I think using pending without having the rest of the test as a "block" argument to it will result in the "end" of that step (so the expected error code wouldn't run). I thought the use of pending was to put the code you *expect* to work in the block, each time the test is run, if that code fails the test can still "pass" as pending, however as soon as the expectations in the block are "true" the test fails and tells you that your thing is no longer pending.

lucywyman: No, this will correctly pend them. This is the output if I use my Puppet PR:
```
lucy ~/githubs/bolt (GH-1835) $ be rspec spec/integration/lookup_spec.rb 
Run options: exclude {:sudo=>true, :windows=>true, :windows_agents=>true}
..............FF

Failures:

  1) lookup() in plans with plan_hiera uses plan_hierarchy outside apply block, and hierarchy in apply block FIXED
     Expected pending 'Until Puppet 6.18 is released' to fail. No error was raised.
     # ./spec/integration/lookup_spec.rb:143

  2) lookup() in plans with invalid plan_hierarchy raises a validation error FIXED
     Expected pending 'Until Puppet 6.18 is released' to fail. No error was raised.
     # ./spec/integration/lookup_spec.rb:156

Finished in 5.1 seconds (files took 1.56 seconds to load)
16 examples, 2 failures

Failed examples:

rspec ./spec/integration/lookup_spec.rb:143 # lookup() in plans with plan_hiera uses plan_hierarchy outside apply block, and hierarchy in apply block
rspec ./spec/integration/lookup_spec.rb:156 # lookup() in plans with invalid plan_hierarchy raises a validation error
```

donoghuc: ok, so when puppet is release you will need to update the tests. I was thinking of a use case where you write the tests that *should* pass (when the thing that is pending is released) and when they start passing (pending starts "failing") you simply remove the pending. I did not realize the code after a `pending()` executed.

lucywyman: Yup, these will start failing when Puppet is released and we'll need to remove the `pending`.

donoghuc: Ahhh, I see. I totally missed that the *desired* outcome is the Error. Ignore me!

PR #2021, opened 2020-07-23
(maint) Group bolt server gems

jpogran:
This groups together the bolt server gems in the Gemfile to allow a user to omit these gems if they want. This decreases the amount installed and speeds up the time it takes to bundle install this project on windows.

!no-release-note

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -10,13 +10,15 @@ ENV['BOLT_GEM'] = 'true'
 
 gemspec
 
-# Bolt server gems are managed here not in the gemspec
-gem "hocon", '>= 1.2.5'
-gem "json-schema", '>= 2.8.0'
-gem "puma", '>= 3.12.0'
-gem "rack", '>= 2.0.5'
-gem "rails-auth", '>= 2.1.4'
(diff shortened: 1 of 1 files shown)


donoghuc:
Will we need to update this invocation? https://github.com/puppetlabs/bolt/blob/main/Dockerfile.bolt-server#L13

jpogran:
I thought groups were optin and ignored otherwise?

donoghuc:
Right, so without opting in the bolt-server container wont have the web stack code.

jpogran:
I only suggested this because I thought that the default behavior is install everything in the gemfile, unless you set `bundle config without server test packaging`.  That's how I've been using bundle with puppet modules up until this time, they commonly have groups based on what you need for tests to run versus normal running.

FWIW, this is what docs say https://bundler.io/guides/groups.html#how-to-manage-groups-of-gems.

Or am I misunderstanding you?

donoghuc:
Ahh, i did not understand the default was to include all groups. This seems correct for the dockerfile. Perhaps we could get this performance benefit for some of the jobs in github actions that dont rely on those gems.

PR #2029, opened 2020-07-27
(GH-1980) Add --env-var cli flag

### LINKED ISSUE #1980: Ability to pass env-vars for run_command and run_script from the CLI
## Use Case
Now that `run_command()` and `run_script()` support the new `_env_vars` option in #1899 it would be great if we could set these from the CLI.

Similar to use cases in #1899 , want to pass in secrets / credentials to commands and scripts. In order to use that feature, we would currently have to wrap the `run_script` and `run_task` functions in a plan which is not ideal for integrations that use the CLI, such as https://github.com/StackStorm-Exchange/stackstorm-bolt

Three solutions come to mind:

#### 1. Provide a `--env-vars ENV_VARS` CLI option

`--env-vars` would be a new CLI option and take a JSON string of environment variables.

```
bolt command run --targets localhost --env-vars '{"NICKS_VAR": "Hello Word"}' 'echo $NICKS_VAR'
```

Downside here is that the environment variables would be exposed and visible when the CLI command. Since we're trying to hide secrets/creds, i'm not a huge fan of this option.

#### 2. Read environment variables from "special" environment variables when Bolt is run

We could designate some special environment variable prefix, in this example i'm going to use `BOLT_ENV_VAR_`. This prefix will be stripped and the rest of the variable name will become the name of the variable when executing the command/script.

Example
```
export BOLT_ENV_VAR_NICKS_VAR="Hello World"
bolt command run --targets localhost 'echo $NICKS_VAR'
```

#### 3. Pass JSON to a new BOLT_ENV_VARS environment variable

We could read BOLT_ENV_VARS and expect this to be a JSON hash. This hash would then set the environment variables in the context when running the command/script.

Example
```
export BOLT_ENV_VARS='{"NICKS_VAR": "Hello World"}'
bolt command run --targets localhost 'echo $NICKS_VAR'
```

lucywyman:
This adds the `--env-var` cli flag, which accepts a string that contains
an equals sign and parses it into a key-value pair to set as an
environment variable on the target. This flag is only available when
running commands or scripts, and can be specified multiple times with
different environment variables. If the same key is specified multiple
times the last instance will take precedence. This doesn't interact with
`_env_var` specified in plans since the flag can only be used when
running commands and scripts.

Closes #1980

!feature

* **Specify remote environment variables using --env-var** ([1980](https://github.com/puppetlabs/bolt/issue/1980))

  Users can now set environment variables on targets when running
  commands and scripts using the CLI flag `--env-var`.

### DIFF
--- a/lib/bolt/bolt_option_parser.rb
+++ b/lib/bolt/bolt_option_parser.rb
@@ -25,7 +25,7 @@ module Bolt
       when 'command'
         case action
         when 'run'
-          { flags: ACTION_OPTS,
+          { flags: ACTION_OPTS + %w[env-var],
             banner: COMMAND_RUN_HELP }
         else
           { flags: OPTIONS[:global],
@@ -106,7 +106,7 @@ module Bolt
       when 'script'
         case action
         when 'run'
-          { flags: ACTION_OPTS + %w[tmpdir],
+          { flags: ACTION_OPTS + %w[tmpdir env-var],
             banner: SCRIPT_RUN_HELP }
         else
           { flags: OPTIONS[:global],
@@ -759,6 +759,15 @@ module Bolt
       end
       define('--[no-]save-rerun', 'Whether to update the rerun file after this command.') do |save|
         @options[:'save-rerun'] = save
+      end
+
+      separator "\nREMOTE ENVIRONMENT OPTIONS"
+      define('--env-var ENVIRONMENT_VARIABLES', 'Environment variables to set on the target') do |envvar|
+        unless envvar.include?('=')
+          raise Bolt::CLIError, "Environment variables must be specified using 'myenvvar=key' format"
+        end
+        @options[:env_vars] ||= {}
+        @options[:env_vars].store(*envvar.split('=', 2))
       end
 
       separator "\nTRANSPORT OPTIONS"
--- a/modules/secure_env_vars/plans/init.pp
+++ b/modules/secure_env_vars/plans/init.pp
@@ -1,0 +1,20 @@
+plan secure_env_vars(
+  TargetSpec $targets,
+  Optional[String] $command = undef,
+  Optional[String] $script = undef
+) {
+  $env_vars = parsejson(system::env('BOLT_ENV_VARS'))
+  unless type($command) == Undef or type($script) == Undef {
+      fail_plan('Cannot specify both script and command for secure_env_vars')
+  }
+
+  return if $command {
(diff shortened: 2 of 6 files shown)


lib/bolt/bolt_option_parser.rb
```diff
@@ -761,6 +761,15 @@ def initialize(options)
         @options[:'save-rerun'] = save
       end
 
+      separator "\nREMOTE ENVIRONMENT OPTIONS"
+      define('--env-var ENVIRONMENT_VARIABLES', 'Environment variables to set on the target') do |envvar|
+        unless envvar.include?('=')
+          raise Bolt::CLIError, "Environment variables must be specified using 'myenvvar=key' format"
+        end
+        @options[:env_vars] ||= {}
+        @options[:env_vars].merge!(Hash[*envvar.split('=', 2)])
```
nicklewis, line 770: I think `@options[:env_vars].store(*envvar.split('=', 2))` would work here and avoid the conversion to `Hash`.

modules/secure_env_vars/plans/init.pp
```diff
@@ -0,0 +1,20 @@
...
+  }
+
+  return if $command {
+           run_command($command, $targets, '_env_vars' => $env_vars)
+         }
+         elsif $script {
+           run_script($script, $targets, '_env_vars' => $env_vars)
+         }
+         else {
+           fail_plan('Cannot specify both script and command for secure_env_vars')
```
nicklewis, line 18: `Must specify either script or command for secure_env_vars`

jpogran:
FWIW, works on windows:

```
➜ be bolt command run 'echo $env:Foo;$env:BAR' -t localhost --env-var FOO=BAR --env-var BAR=woot
[DEPRECATION] Struct layout is already defined for class Windows::ServiceStructs::SERVICE_STATUS_PROCESS. Redefinition as in C:/Users/james/src/puppetlabs/bolt/.bundle/gems/ruby/2.5.0/gems/win32-service-0.8.8/lib/win32/windows/structs.rb:67:in `<class:SERVICE_STATUS_PROCESS>' will be disallowed in ffi-2.0.
Started on localhost...
Finished on localhost:
  STDOUT:
    BAR
    woot
Ran on 1 target in 0.35 sec
```

However, the pwsh module building logic won't accept multiple `env-var` parameters. Will have to think on how to accept that. Maybe accept what I do for tasks with `-params @{ FOO = 'bar'; BAZ= 'woot' }` and unwrap it to multiple env-var statements before invoking

Issue #1982, opened 2020-07-10
Bolt hangs/crashes when loosing winrm service


bishopbm1:
## Describe the Bug
When running commands on windows if the WinRM Service gets restarted/flaps/or goes down for any reason the following error occurs and bolt just crashes and hangs until user input to exit out of the command.

Error:
```
# bolt task run --modulepath /opt/encore/puppet/ --targets winrm://1.2.3.4:5986 --user testuser --password testpassword --no-ssl-verify service action='restart' name='WinRM' force='true'
Fact file /etc/facter/facts.d/encore.json was parsed but returned an empty data set
Started on winrm://1.2.3.4:5986...
#<Thread:0x00007f6c68993870@/opt/puppetlabs/bolt/lib/ruby/gems/2.5.0/gems/logging-2.2.2/lib/logging/diagnostic_context.rb:471 run> terminated with exception (report_on_exception is true):
Traceback (most recent call last):
  14: from /opt/puppetlabs/bolt/lib/ruby/gems/2.5.0/gems/logging-2.2.2/lib/logging/diagnostic_context.rb:474:in `block in create_with_logging_context'
  13: from /opt/puppetlabs/bolt/lib/ruby/gems/2.5.0/gems/bolt-2.16.0/lib/bolt/transport/winrm/connection.rb:114:in `block in execute'
  12: from /opt/puppetlabs/bolt/lib/ruby/gems/2.5.0/gems/winrm-2.3.4/lib/winrm/shells/base.rb:79:in `run'
  11: from /opt/puppetlabs/bolt/lib/ruby/gems/2.5.0/gems/winrm-2.3.4/lib/winrm/shells/base.rb:131:in `with_command_shell'
  10: from /opt/puppetlabs/bolt/lib/ruby/gems/2.5.0/gems/winrm-2.3.4/lib/winrm/shells/base.rb:80:in `block in run'
   9: from /opt/puppetlabs/bolt/lib/ruby/gems/2.5.0/gems/winrm-2.3.4/lib/winrm/psrp/receive_response_reader.rb:58:in `read_output'
   8: from /opt/puppetlabs/bolt/lib/ruby/gems/2.5.0/gems/winrm-2.3.4/lib/winrm/wsmv/receive_response_reader.rb:73:in `with_output'
[...]

PR #2030, opened 2020-07-27
(GH-1982) Handle WinRM connection loss

nicklewis:
Previously, losing connection to WinRM (for instance, due to the
service being stopped or restarted) would cause a stacktrace and a
deadlock. This happened because we manage the output from the command in
a thread which wasn't handling exceptions. The stack trace came from the
exception bubbling up, with the default behavior of Thread being to
print all exceptions that escape the thread. The deadlock came from our
failure to close the out/err streams that the shell was waiting on.

We now ensure the out/err streams are always closed and we disabled the
"report_on_exception" behavior to avoid an unnecessary stack trace, as
our own code will handle the exception at the next level up.


nicklewis:
Release noted!

PR #2035, opened 2020-07-28
(GH-1997) Add tests for JSON schemas

### LINKED ISSUE #1997: Add tests for schemas
## Use Case
We should provide some tests for the JSON schemas to ensure that they are generated correctly, validate files correctly, and that Bolt provides the necessary information for each config option to generate a definition in the schema. We can check whether the schema validates files correctly by using the `json_schemer` gem to validate the schema against some fixtures. Potential tests include:

- Check that config options at a minimum provide a `description` and `type` key
- Check that the schema rake task completes successfully
- ~Validate a valid file against the generated schemas for each of the config files and inventory file~
- ~Validate an invalid file against the generated schemas for each of the config files and inventory file~

To accomplish this we should commit the generated schemas to the repo and add a GH action that generates the schemas and verifies that it matches what's committed on the PR. We should verify that the schema generation completes successfully and that each config option has a description and type.

beechtom:
This adds tests to ensure that the config options hashes include
`:type` and `:description` keys, which are required to correctly
generate Bolt's JSON schemas.

It also includes a workflow for generating the JSON schemas and
checking if there were any changes from the schemas that were committed.
If there are changes, the workflow fails and notifies the user that they
should run the `schemas:all` Rake task and commit the changes.

Closes #1997 

!no-release-note

### DIFF
--- a/.github/workflows/schemas.yaml
+++ b/.github/workflows/schemas.yaml
@@ -1,0 +1,56 @@
+name: Schemas
+
+on:
+  push:
+    branches: [main]
+    paths:
+      - 'rakelib/schemas.rake'
+      - 'schemas/*.json'
+      - 'lib/bolt/config/options.rb'
+      - 'lib/bolt/config/transport/options.rb'
+  pull_request:
+    type: [opened, reopened, edited]
+    paths:
+      - 'rakelib/schemas.rake'
+      - 'schemas/*.json'
+      - 'lib/bolt/config/options.rb'
+      - 'lib/bolt/config/transport/options.rb'
+
+jobs:
+
+  generate:
+    name: Generate
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+      - name: Setup Ruby
+        uses: actions/setup-ruby@v1
+        with:
+          ruby-version: '2.5.x'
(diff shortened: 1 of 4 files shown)


spec/shared_examples/options.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+# This function is used to ensure that every config option
+# has a :type and :description key, as they are required for
+# generating the JSON schemas and validating config.
+def assert_type_description(definitions)
```
nicklewis, line 6: This should be in `spec/lib` rather than `shared_examples`. This directory is specifically for [shared examples](https://relishapp.com/rspec/rspec-core/docs/example-groups/shared-examples).

spec/lib/options.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+# This function is used to ensure that every config option
+# has a :type and :description key, as they are required for
+# generating the JSON schemas and validating config.
+def assert_type_description(definitions)
```
nicklewis, line 6: This should follow the same file structure as the rest of `spec/lib` (specifically it should be in `spec/lib/bolt_spec`) and it should be inside a `module` that is included by each test case that needs this method, rather than defining it globally.

Issue #2038, opened 2020-07-28
out::message step for yaml plans


adreyer:
## Use Case
Users should not have to use eval to print a message it should have it's own dedicated plan step.

## Describe the Solution You Would Like
The step key should be "message" and it should take no further arguments.
This should also be added to the plan converter for converting to puppet plans

PR #2040, opened 2020-07-29
(GH-2038) Add YAML plan message step

beechtom:
This adds a YAML plan message step that prints a message. Messages are
printed to `stdout` when using the `human` output format and `stderr`
when using the `json` output format.

!feature

* **YAML plan message step**
  ([#2038](https://github.com/puppetlabs/bolt/issues/2038))

  YAML plans now support a message step that prints a message.

### DIFF
--- a/documentation/writing_yaml_plans.md
+++ b/documentation/writing_yaml_plans.md
@@ -71,11 +71,28 @@ failed.
 
 Steps use these fields:
 
--   `name`: A unique name that can be used to refer to the result of the step
-    later
+-   `name`: (Optional) A unique name that can be used to refer to the result of
+    the step later.
 -   `description`: (Optional) An explanation of what the step is doing.
 
 Other available keys depend on the type of step.
+
+#### Message step
+
+Use a `message` step to print a message. This will print a message to `stdout`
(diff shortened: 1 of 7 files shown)


lib/bolt/pal/yaml_plan/step/message.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+
+module Bolt
+  class PAL
+    class YamlPlan
+      class Step
+        class Message < Step
+          def self.allowed_keys
+            Set['message']
```
nicklewis, line 9: I think we should allow `name` and `description` here just for consistency. They don't do _much_, but it seems simpler to accept them and let them do nothing than to make this step the exception.

documentation/writing_yaml_plans.md
```diff
@@ -71,12 +71,29 @@ failed.
 
 Steps use these fields:
 
--   `name`: A unique name that can be used to refer to the result of the step
-    later
+-   `name`: (Optional) A unique name that can be used to refer to the result of
+    the step later
```
hestonhoffman, line 75:
```suggestion
    the step later.
```

Issue #1981, opened 2020-07-10
Write an example/prototype of how to run a directory of external executable facts and return output that looks something like what facter would produce


npwalker:
## Use Case
I've spoken with customers that have scripts they want to run with bolt and make sense of the information the script produces.  The scripts do only reading of system information and can produce nested structures.  

This makes the information produced by the scripts much more like puppet facts than just an arbitrary script that performs an action on the node.  

## Describe the Solution You Would Like
Given a directory of [external executable facts](https://puppet.com/docs/puppet/latest/external_facts.html#executable-facts-on-unix) run each one and collate the information produced to look something like what facter would produce.  

This provides a clearer piece of guidance on how to structure information gathering "scripts" instead of one big script break each one down into an external executable fact and we'll output it in a nicer format for you  

This should be a plan similar to the facts plan that takes a path to a directory and sets facts based on running scripts in that directory.

## Describe Alternatives You've Considered
An alternative would be to make these external executable facts and get them submitted by the agent every run.  But for many facts that don't influence the behavior of a puppet run storing those in PuppetDB may be detrimental to your health ( or at least the perf of PuppetDB) 

If facter provided a way to run a group of facts separately from those that are submitted every agent run that would help but also requires facter.  The solution here is more advantageous in that it doesn't require facter.

beechtom:
Closed by https://github.com/puppetlabs/puppetlabs-facts/pull/41

PR #2047, opened 2020-07-30
(GH-1981) Add dir::children plan function

lucywyman:
This adds a new builtin Bolt module `dir` that contains utilities for
interacting with directories during a Bolt plan, analagous to Ruby's
`Dir` class. It currently has just one function, `children`, which lists
the contents of a given directory and returns an ordered list of bare
filenames. This is used in the `facts::external` plan to get the list of
scripts to run to generate facts from.

The function accepts either an absolute path, a path relative to the
root of the active Bolt project, or a module and subpath. All of these
are only accepted as strings. If none are found, an error is raised.

!feature

* **New dir::children plan function**

  New plan function `dir::children` returns an array containing all of
  the filenames in the given directory, similar to Ruby's `Dir.children()`

### DIFF
--- a/bolt-modules/dir/lib/puppet/functions/dir/children.rb
+++ b/bolt-modules/dir/lib/puppet/functions/dir/children.rb
@@ -1,0 +1,35 @@
+# frozen_string_literal: true
+
+require 'pathname'
+
+# Returns an array containing all of the filenames except for "." and ".." in the given directory.
+Puppet::Functions.create_function(:'dir::children', Puppet::Functions::InternalFunction) do
+  # @param dirname Absolute path or Puppet module name.
+  # @return Array of files in the given directory.
+  # @example List filenames from an absolute path.
+  #   dir::children('/home/user/subdir/')
+  # @example List filenames from a Puppet file path.
+  #   dir::children('puppet_agent')
+  dispatch :children do
+    scope_param
+    required_param 'String', :dirname
+    return_type 'Array'
+  end
+
+  def children(scope, dirname)
+    # Send Analytics Report
+    Puppet.lookup(:bolt_executor) {}&.report_function_call(self.class.name)
+    modname, subpath = dirname.split(File::SEPARATOR, 2)
+    mod_path = scope.compiler.environment.module(modname)&.path
+
+    full_mod_path = File.join(mod_path, subpath || '') if mod_path
+
+    # Expand relative to the project directory if path is relative
+    project = Puppet.lookup(:bolt_project_data)
+    pathname = Pathname.new(dirname)
+    full_dir = pathname.absolute? ? dirname : File.expand_path(File.join(project.path, dirname))
+
+    # Sort for testability
+    Dir.children(full_mod_path || full_dir).sort
+  end
+end
--- a/bolt-modules/dir/Rakefile
+++ b/bolt-modules/dir/Rakefile
@@ -1,0 +1,3 @@
+# frozen_string_literal: true
+
+require 'puppetlabs_spec_helper/rake_tasks'
--- a/bolt-modules/dir/spec/functions/dir/children_spec.rb
+++ b/bolt-modules/dir/spec/functions/dir/children_spec.rb
@@ -1,0 +1,38 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+require 'fileutils'
+
+describe 'dir::children' do
+  include PuppetlabsSpec::Fixtures
+  let(:path) { fixtures('modules', 'test') }
+
+  around(:each) do |example|
+    Puppet[:tasks] = true
+    project = Struct.new(:name, :path).new('default', File.expand_path(path))
+
+    Puppet.override(bolt_project_data: project) do
+      example.run
+    end
+  end
+
+  context 'finding an absolute path' do
+    let(:abs_path) { File.expand_path(File.join(path, 'facts.d')) }
+    it {
+      is_expected.to run.with_params(abs_path).and_return(%w[.hidden fact.py two.sh])
+    }
+  end
+
+  context 'finding a relative path' do
+    it {
+      # This is relative to the project directory
+      is_expected.to run.with_params('facts.d').and_return(%w[.hidden fact.py two.sh])
+    }
+  end
+
+  context 'finding a puppet module' do
+    it {
+      is_expected.to run.with_params('test/facts.d').and_return(%w[.hidden fact.py two.sh])
+    }
+  end
+end
--- a/bolt-modules/dir/spec/spec_helper.rb
+++ b/bolt-modules/dir/spec/spec_helper.rb
@@ -1,0 +1,8 @@
+# frozen_string_literal: true
+
+require 'puppet_pal'
+
+# Ensure tasks are enabled when rspec-puppet sets up an environment
+# so we get task loaders.
+Puppet[:tasks] = true
+require 'puppetlabs_spec_helper/module_spec_helper'
--- a/rakelib/tests.rake
+++ b/rakelib/tests.rake
@@ -84,7 +84,7 @@ begin
     task :modules do
       success = true
       # Test core modules
-      %w[boltlib ctrl file out prompt system].each do |mod|
+      %w[boltlib ctrl file dir out prompt system].each do |mod|
         Dir.chdir("#{__dir__}/../bolt-modules/#{mod}") do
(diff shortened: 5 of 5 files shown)


beechtom:
+1 for this! I was thinking something like this would also be helpful for when you download a directory during a plan run. :)

bolt-modules/dir/lib/puppet/functions/dir/entries.rb
```diff
@@ -0,0 +1,26 @@
...
+    Puppet.lookup(:bolt_executor) {}&.report_function_call(self.class.name)
+    found = Puppet::Parser::Files.find_file(dirname, scope.compiler.environment)
+    unless found && Puppet::FileSystem.exist?(found)
+      raise Puppet::ParseErrorWithIssue.from_issue_and_stack(
+        Puppet::Pops::Issues::NO_SUCH_FILE_OR_DIRECTORY, file: filename
+      )
+    end
+
+    # Sort for testability
+    Dir.entries(dirname).sort
```
beechtom, line 24: Maybe `Dir.children` instead? I'm not sure that having `.` and `..` is particularly useful.

bolt-modules/dir/lib/puppet/functions/dir/entries.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+# Enumerate files in a given directory
```
beechtom, line 3: This should have an example, similar to the other functions. Since the `dirname` can be an absolute path or Puppet file path, I think showing examples for each would be helpful.

beechtom, line 3:
```suggestion
# Enumerate the files in a given directory.
```

bolt-modules/dir/lib/puppet/functions/dir/entries.rb
```diff
@@ -0,0 +1,26 @@
...
+Puppet::Functions.create_function(:'dir::entries', Puppet::Functions::InternalFunction) do
+  # @param dirname Absolute path or Puppet file path.
+  # @return Array of files in the given directory
```
beechtom, line 6:
```suggestion
  # @return Array of files in the given directory.
```

bolt-modules/dir/lib/puppet/functions/dir/children.rb
```diff
@@ -0,0 +1,25 @@
...
+    return_type 'Array'
+  end
+
+  def children(scope, dirname)
+    # Send Analytics Report
+    Puppet.lookup(:bolt_executor) {}&.report_function_call(self.class.name)
+    puppet_mod = scope.compiler.environment.module(dirname)&.path
+
+    # Sort for testability
+    Dir.children(puppet_mod || dirname).sort
```
beechtom, line 23: While not strictly necessary, I do think it would be nice to rescue the Ruby error here and raise a Bolt error with a cleaner message. Currently if the directory doesn't exist the message will be similar to: `No such file or directory @ dir_initialize - foo"`

bolt-modules/dir/lib/puppet/functions/dir/children.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+# Returns an array containing all of the filenames except for "." and ".." in the given directory.
+Puppet::Functions.create_function(:'dir::children', Puppet::Functions::InternalFunction) do
+  # @param dirname Absolute path or Puppet module name.
```
beechtom, line 5: This function does accept a relative path, which will expand relative to the current working directory.
Would it make sense to have it expand relative to the project instead? I don't really feel strongly either way.

bolt-modules/dir/lib/puppet/functions/dir/children.rb
```diff
@@ -0,0 +1,25 @@
...
+  dispatch :children do
+    scope_param
+    required_param 'String', :dirname
+    return_type 'Array'
+  end
+
+  def children(scope, dirname)
+    # Send Analytics Report
+    Puppet.lookup(:bolt_executor) {}&.report_function_call(self.class.name)
+    puppet_mod = scope.compiler.environment.module(dirname)&.path
```
beechtom, line 20: +1 for being able to specify a module by name. I think this could allow a little more flexibility by either allowing you to specify a path within a module (e.g. `module/plans`, `module/files`) or limit it to loading children in the module's `files` directory.

bolt-modules/dir/lib/puppet/functions/dir/children.rb
lucywyman: I think that makes sense since this is a plan function, and expanding relative to projects makes it more predictable. I'm almost inclined to just fail if users try to use a relative path, but it'd be useful even for things like the faux-external-facts plan where you could just load `facts.d`.

bolt-modules/dir/lib/puppet/functions/dir/children.rb
lucywyman: Hm, I think if we do that we'll lose the line and file we get from Pal, and that doesn't seem worth recalculating:
```
lucy ~/githubs/bolt (GH-1981) $ be bolt plan run bolt -t localhost
Starting: plan bolt
Finished: plan bolt in 0.0 sec
{
  "kind": "dir/children",
  "msg": "Could not list files in fake: No such file or directory @ dir_initialize - /home/lucy/githubs/bolt/fake",
  "details": {
  }
}
lucy ~/githubs/bolt (GH-1981) $ be bolt plan run bolt -t localhost
Starting: plan bolt
Finished: plan bolt in 0.0 sec
{
  "kind": "bolt/pal-error",
  "msg": "No such file or directory @ dir_initialize - /home/lucy/githubs/bolt/fake",
  "details": {
    "file": "/home/lucy/githubs/bolt/plans/init.pp",
    "line": 4,
    "column": 10
  }
}
```

I can modify the error message and reraise it, but honestly "Could not list files in fake: No such file or directory @ dir_initialize - /home/lucy/githubs/bolt/fake" doesn't seem any better than just "No such file or directory @ dir_initialize - /home/lucy/githubs/bolt/fake"

bolt-modules/dir/lib/puppet/functions/dir/children.rb
```diff
@@ -0,0 +1,35 @@
...
+  def children(scope, dirname)
+    # Send Analytics Report
+    Puppet.lookup(:bolt_executor) {}&.report_function_call(self.class.name)
+    modname, subpath = dirname.split(File::SEPARATOR, 2)
+    mod_path = scope.compiler.environment.module(modname)&.path
+
+    full_mod_path = File.join(mod_path, subpath) if mod_path
+
+    # Expand relative to the project directory if path is relative
+    project = Puppet.lookup(:bolt_project)
```
beechtom, line 28: `:bolt_project` will be `nil` if the project isn't named. This should use `:bolt_project_data` instead.

See https://github.com/puppetlabs/bolt/blob/main/lib/bolt/pal.rb#L158 and https://github.com/puppetlabs/bolt/blob/main/bolt-modules/boltlib/lib/puppet/functions/download_file.rb#L99

bolt-modules/dir/lib/puppet/functions/dir/children.rb
```diff
@@ -0,0 +1,35 @@
...
+    return_type 'Array'
+  end
+
+  def children(scope, dirname)
+    # Send Analytics Report
+    Puppet.lookup(:bolt_executor) {}&.report_function_call(self.class.name)
+    modname, subpath = dirname.split(File::SEPARATOR, 2)
+    mod_path = scope.compiler.environment.module(modname)&.path
+
+    full_mod_path = File.join(mod_path, subpath) if mod_path
```
beechtom, line 25:
```suggestion
    full_mod_path = File.join(mod_path, subpath) if mod_path && subpath
```

This will fail if there is no subpath.

bolt-modules/dir/lib/puppet/functions/dir/children.rb
lucywyman: Ah, I missed when bolt_project_data was added.

lucywyman: Hm, this would require a change to Puppet but it seems like we should just add a 'load' key to the project struct, rather than having two separate variables.

PR #2048, opened 2020-07-30
(Docs) Clarify inventory name, alias, uri

hestonhoffman:
Feedback from community Slack suggests we should be clearer
about how `uri`, `alias`, and `name` work in inventory files.

I've moved the examples up in the doc and added some extra
clarification.

!no-release-note

### DIFF
--- a/documentation/inventory_file_v2.md
+++ b/documentation/inventory_file_v2.md
@@ -68,20 +68,22 @@ groups:
 
 ### Target object
 
-You can specify a target with the string representation of a URI, or as a hash
-with any of the following fields:
+Specify a target under the `targets` key. The `targets` key accepts an array
+where each element is either a string or a hash. 
+- For a string, use the string representation of the target's Universal Resource
+  Identifier (URI).
+- A hash must specify either the `uri` for a hash, or a `name`. If you don't
+  specify a `name`, Bolt uses the `uri` as the target's name. To make it easier to run Bolt commands
+  on the target, specify both a `uri` and a `name` and run your commands against
+  the `name`.
 
-| Key | Description | Type |
-| --- | ----------- | ---- |
-| `alias` | A unique alias to refer to the target. Optional. | `String` |
-| `config` | The configuration for the target. Optional. | `Hash` |
-| `facts` | The facts for the target. Optional. | `Hash` |
-| `features` | The features for the target. Optional. | `Array[String]`
-| `name` | A human-readable name for a target.<br> **Required** when specifying a target using a hash. Optional if using `uri`. | `String` |
-| `uri` | The URI of the target.<br> **Required** when specifying a target using a hash. Optional if using `name`. | `String` |
-| `vars` | The variables for the target. Optional. | `Hash` |
+Bolt uses the target's `uri` to establish a connection to the target. If you set
+a `name` and no `uri`, you must specify the hostname for the target using the
+`host` key in your transport configuration. If the transport you're using does
+not support hostname configuration, you must set a `uri`.
 
-An example of targets specified with the string representations of their URIs:
+An example of two targets specified with the string representations of
+their URIs:
 
 ```yaml
 targets:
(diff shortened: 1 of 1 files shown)


hestonhoffman:
Not 100% on that third example.

documentation/inventory_file_v2.md
```diff
@@ -68,18 +68,14 @@ groups:
 
 ### Target object
 
-You can specify a target with the string representation of a URI, or as a hash
-with any of the following fields:
+Specify a target under the `targets` key. You can specify a target's name in the following ways:
+- Use an array with the string representation of the target's Universal Resource Identifier (URI).
```
beechtom, line 72:
```suggestion
- Use the string representation of the target's Universal Resource Identifier (URI).
```

The `targets` key accepts an array where each element is either a string or hash.

documentation/inventory_file_v2.md
```diff
@@ -68,18 +68,14 @@ groups:
...
-| `config` | The configuration for the target. Optional. | `Hash` |
-| `facts` | The facts for the target. Optional. | `Hash` |
-| `features` | The features for the target. Optional. | `Array[String]`
-| `name` | A human-readable name for a target.<br> **Required** when specifying a target using a hash. Optional if using `uri`. | `String` |
-| `uri` | The URI of the target.<br> **Required** when specifying a target using a hash. Optional if using `name`. | `String` |
-| `vars` | The variables for the target. Optional. | `Hash` |
+If you don't specify a `name` for your target, Bolt sets the name to the `uri`.
+If you set a `name` and no `uri`, and the transport you're using supports
+hostname configuration, Bolt uses the `host` configuration. If the transport
+you're using does not support hostname configuration, you must set a `uri`.
```
beechtom, line 78: I think it'd be helpful to explain what it uses `host` configuration for. Bolt will use the target's `uri` to establish a connection to the target, and falls back to `host` if the `uri` is not specified.

documentation/inventory_file_v2.md
```diff
@@ -103,6 +99,28 @@ targets:
       transport: winrm
 ```
 
+An example of a target specified with `name`. In this case, Bolt uses the hostname of the target from the transport's `host` key.
```
beechtom, line 102: Same as above - it would be helpful to say what `host` is used for.

documentation/inventory_file_v2.md
heylix: Maybe explain the "lookup order".

AFAIK: If all keys are present, `uri` overwrites what is specified in `host` and `host` overwrites what is specified in `name` when trying to connect. `uri` > `host` > `name`. If one of those is not present, Bolt starts evaluating from left to right.

A proper and easy to understand sentence for this would be hard, but necessary, IMHO.

beechtom: Bolt actually does not use the target's `name` to establish connections, it's only used to reference the target from the inventory. I think maybe that's where some confusion lies, since you can 1) name a target and 2) specify the target's host. The `uri` key does both of those.

| | Name a target | Connect to a target |
| --- | --- | --- |
| **Higher precedence** | `name` | `uri` |
| **Lower precedence** | `uri` | `host` |

In the case of naming a target, either a `uri` or `name` is required, otherwise Bolt will error when you load inventory. In the case of connecting to a target, either `uri` or `host` is required, otherwise Bolt will error when you try connecting to the target.

hestonhoffman: I've added some language around `uri` and `name` to make this a little clearer. I also messed with the examples a bit.

documentation/inventory_file_v2.md
```diff
@@ -68,41 +68,85 @@ groups:
...
       transport: ssh
   - uri: target2.example.com
-    alias: target2
+    name: target2
+    alias: database
     config:
       transport: winrm
 ```
 
+The `targets` key accepts the following:
```
beechtom, line 139:
```suggestion
Targets specified with a hash accept the following fields:
```

PR #2055, opened 2020-08-03
(GH-2012) Print objects using out::message

### LINKED ISSUE #2012: Handle ouputting more data types with out::message
## Use Case
Users have to explicitly decide how to turn Puppet Objects into strings with `out::message` if they try to pass a non-string object the plan errors. This puts the burden of stringifying objects onto the plan author even though in many cases the outputters already know how to stringify these objects.

Users should be able to pass more objects to out::message, `PlanResult` objects should be passed to the outputter for stringification while other objects should be converted to data before being passed to the outputter. Since the outputter runs in a separate thread we should avoid passing pcore objects to the outputter.

## Questions
Should we use a different function?
Should we use pcore serialization?
Is `PlanResult` the correct set of datatypes?

## Solution
We should have special behavior for the `PlanResult` type, and everything else should be stringified. We should just send the `PlanResult` to the outputter and have the outputter determine how to output it.
What about things that are not `PlanResult`, like `Regex`? We should stringify them, and not to_data them.

## Notes
Because the outputter is in a separate thread it may not have access to the compiler + types, so we may need to use the same method we use to print resultsets.
We should try to reuse the outputter logic that prints objects, and try to preserve the integrity of the objects being printed where possible.
We'll handle a certain set of datatypes acceptable in a plan, and we can print them as data. Custom types will just be converted into strings.
This should print the stringified object as out::message implies the object will be stringified. We may also want to add a new function for printing data.

lucywyman:
This allows users to pass any data type to the `out::message()` plan
function and send it to the configured outputter for printing. This
adds a formatting function to the outputter that all messages pass
through, which formats known Bolt datatypes and classes nicely for
readable printing. Any unhandled types are stringified and passed to the
outputter. Printing is still left entirely to the outputter, but
formatting is shared among all outputter that inherit from the base
outputter class `Bolt::Outputter`. This ensures a consistent experience
across outputters.

Closes #2012

!feature

* **Print objects using out::message()** ([#2012](https://github.com/puppetlabs/bolt/issues/2012))

  Users can now print any valid data type using the plan function
  `out::message()`.

### DIFF
--- a/lib/bolt/outputter.rb
+++ b/lib/bolt/outputter.rb
@@ -21,6 +21,62 @@ module Bolt
       @trace = trace
       @stream = stream
     end
+
+    def indent(indent, string)
+      indent = ' ' * indent
+      string.gsub(/^/, indent.to_s)
+    end
+
+    def print_message_event(event)
+      print_message(stringify(event[:message]))
+    end
+
+    def print_message
+      raise NotImplementedError, "print_message() must be implemented by the outputter class"
+    end
+
+    def stringify(message)
+      formatted = format_message(message)
+      if formatted.is_a?(Hash) || formatted.is_a?(Array)
+        ::JSON.pretty_generate(formatted)
+      else
+        formatted
+      end
+    end
+
+    def format_message(message)
+      case message
+      when Array
+        message.map { |item| format_message(item) }
+      when Bolt::ApplyResult
+        format_apply_result(message)
+      when Bolt::Result, Bolt::ResultSet
+        # This is equivalent to to_s, but formattable
+        message.to_data
+      when Bolt::RunFailure
+        formatted_resultset = message.result_set.to_data
+        message.to_h.merge('result_set' => formatted_resultset)
+      when Hash
+        message.each_with_object({}) do |(k, v), h|
+          h[format_message(k)] = format_message(v)
+        end
+      when Integer, Float, NilClass
+        message
+      else
+        message.to_s
+      end
+    end
+
+    def format_apply_result(result)
+      logs = result.resource_logs&.map do |log|
+        # Omit low-level info/debug messages
+        next if %w[info debug].include?(log['level'])
+        indent(2, format_log(log))
+      end
+      hash = result.to_data
+      hash['logs'] = logs unless logs.empty?
+      hash
+    end
   end
 end
 
--- a/spec/bolt/outputter/outputter_spec.rb
+++ b/spec/bolt/outputter/outputter_spec.rb
@@ -1,0 +1,165 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+require 'bolt/outputter'
+require 'bolt/cli'
+
+describe "Bolt::Outputter" do
+  let(:output) { StringIO.new }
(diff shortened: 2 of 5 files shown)


spec/bolt/outputter/outputter_spec.rb
```diff
@@ -0,0 +1,181 @@
...
+     ],
+     [
+       "subarray"
+     ]
+   ]
+   ARRAY
+  end
+
+  it "formats hashes of complex objects" do
+  end
```
beechtom, line 175: ?

lucywyman: Woops! I'll add this later this morning.

lib/bolt/outputter.rb
```diff
@@ -21,6 +21,76 @@ def initialize(color, verbose, trace, stream = $stdout)
...
+        message.map { |item| format_message(item) }
+      when Bolt::ApplyResult
+        format_apply_result(message)
+      when Bolt::Result, Bolt::ResultSet
+        # This is equivalent to to_s, but formattable
+        message.to_data
+      when Bolt::RunFailure
+        formatted_resultset = message.result_set.to_data
+        message.to_h.merge('result_set' => formatted_resultset)
+      when Bolt::PlanResult
```
nicklewis, line 59: I don't think we need this case. A `Bolt::PlanResult` instance is only constructed within Bolt when the outermost plan run finishes, so it shouldn't be available to `out::message` under normal circumstances. Technically a Ruby function could construct one on its own, but that's a corner case and probably not worth having special handling for.

lucywyman: You're totally right - I thought the result of a `run_plan` would be this type, but it's just a ResultSet. Removed!

lib/bolt/outputter.rb
```diff
@@ -21,6 +21,74 @@ def initialize(color, verbose, trace, stream = $stdout)
...
+      when Hash
+        message.each_with_object({}) do |(k, v), h|
+          h[format_message(k)] = format_message(v)
+        end
+      else
+        message.to_s
+      end
+    end
+
+    def format_plan_result(plan_result)
```
nicklewis, line 68: This method should be unneeded now.

lib/bolt/outputter.rb
```diff
@@ -21,6 +21,60 @@ def initialize(color, verbose, trace, stream = $stdout)
...
+        format_apply_result(message)
+      when Bolt::Result, Bolt::ResultSet
+        # This is equivalent to to_s, but formattable
+        message.to_data
+      when Bolt::RunFailure
+        formatted_resultset = message.result_set.to_data
+        message.to_h.merge('result_set' => formatted_resultset)
+      when Hash
+        message.each_with_object({}) do |(k, v), h|
+          h[format_message(k)] = format_message(v)
```
nicklewis, line 61: This prints hashes in kind of a misleading way, because it will convert some non-string elements to strings.

```puppet
  $obj = {
    "foo" => 1,
    "bar" => 2
  }
  out::message($obj)
```

results in

```
{
  "foo": "1",
  "bar": "2"
}
```

nicklewis: I would expect the array case to have similar behavior.

Issue #2025, opened 2020-07-26
Error: undefined method `[]' for nil:NilClass when applying manifest


theq86:
## Describe the Bug

When an inventory variable, which is an array,  is used inside a plan, the following error occurs:
`undefined method `[]' for nil:NilClass`

When a scalar value is used, the plan finishes successfully.

## Expected Behavior

Both scalar values and arrays should be working inside a plan which applies a manifest block

## Steps to Reproduce

I added a plan and two inventory files below in the context section. One inventory file contains a variable which is an array. Running the plan with this inventory should reproduce the failure. The other inventory with the scalar variable is a working example. I just added this to emphasize, that the array var is somehow the reason why it fails.

## Environment
 - Version 2.14.0 and onwards (earlier versions work)
 - Platform macOS 10.15.4 and onwards (not tested with earlier versions)

## Additional Context

This bug may be somehow related to https://github.com/puppetlabs/bolt/issues/1836 because this bug also occured when I used array inventory vars. This bug was fixed and until 2.14.0 my plan worked flawlessly.

_Plan which can the bug be reproduced with_
```
plan reproduce::error (
  TargetSpec $targets,
) {
    $hosts = get_targets($targets)
    $hosts.each |$target| {
    $vars = $target.vars()
    $roles = $vars['roles']
    apply_prep($target)
    apply($target) {
      notify { 'foo':
        message => 'bar'
      }
    }
  }
}
```

Inventory file which causes problem
```
---
version: 2
targets:
  - alias: web
    vars:
      roles:
        - array
        - values
    uri: example.org
```

Inventory file with works
```
---
version: 2
targets:
  - alias: web
    vars:
      roles: single value
    uri: example.org
```

beechtom:
This seems to be a bug introduced by https://github.com/puppetlabs/bolt/pull/1798 , specifically these lines: https://github.com/puppetlabs/bolt/pull/1798/files#diff-2ba6f352b8c1a13c88bd62f87af4aec3R76-R82 .

Previously, plan variables were deserialized as part of PAL's catalog compiler ( https://github.com/puppetlabs/bolt/pull/1798/files#diff-2ba6f352b8c1a13c88bd62f87af4aec3L76-L107 ) before collisions with facts and target variables were detected. The change in this PR instead filters out variable collisions before creating the compiler, and sets them as variables when creating the temp environment. Once the catalog compiler is created, Puppet attempts to deserialize the variables and raises an error.

[...]

PR #2056, opened 2020-08-03
(GH-2025) Do not modify order of plan variables in catalog compilation

beechtom:
This fixes a bug in catalog compilation that was caused by modifying the
order of plan variables in the catalog request. This bug was caused by
the following conditions being met:

- A plan variable has the same name as a target variable
- The plan variable references the same complex data as a variable that
was listed earlier in the plan (for example, it references a target's
config)

When Puppet serializes data with local references, it replaces any
repeated, complex data with a local reference. Then, when this data is
deserialized, Puppet will examine the variables in order and resolve any
local references it comes across. If a local reference appears before
the data it references, Puppet will be unable to resolve the reference
and will raise an error.

In short, Puppet expects that any data that has
been serialized will only create local references for data that appeared
earlier, and that during deserialization any local references will only
reference data that has already been resolved/deserialized.

When Bolt prepares to compile a catalog request, it first shadows any
variable conflicts between target facts, plan variables, and target
variables. After shadowing conflicts, Bolt then merges plan variables
into the target variables. Because of the way Ruby merges two hashes,
any variables that exist in the target hash will be replaced by
similarly-named plan variables, and then any additional variables will
be appended to the end of the hash. This results in any plan variables
with local references being placed 'higher up' in order than any plan
variable it references.

Now, instead of simply merging target and plan variables, Bolt will
first remove any conflicting target variables from the variables hash,
then merge in the plan variables. This results in the order of the plan
variables being maintained.

!bug

* **Do not modify order of plan variables in catalog compilation**
  ([#2025](https://github.com/puppetlabs/bolt/issues/2025))

  Bolt will no longer error during catalog compilation when a plan
  variable shares the same name as a target variable. Previously, Bolt
  would modify the order plan variables were listed in a catalog if they
  shared the name of a target variable, causing catalog compilation to
  fail when deserializing variables.

### DIFF
--- a/lib/bolt/catalog.rb
+++ b/lib/bolt/catalog.rb
@@ -76,7 +76,15 @@ module Bolt
       target = request['target']
       plan_vars = shadow_vars('plan', request['plan_vars'], target['facts'])
       target_vars = shadow_vars('target', target['variables'], target['facts'])
-      topscope_vars = target_vars.merge(plan_vars)
+
+      # Merge plan vars with target vars, while maintaining the order of the plan
+      # vars. It's critical that the order of plan vars is not changed, as Puppet
+      # will deserialize the variables in the order they appear. Variables may
+      # contain local references to variables that appear earlier in a plan. If
+      # these variables are moved before the variable they reference, Puppet will
+      # be unable to deserialize the data and raise an error.
+      topscope_vars = target_vars.reject { |k, _v| plan_vars.key?(k) }.merge(plan_vars)
+
       env_conf = { modulepath: request['modulepath'],
                    facts: target['facts'],
                    variables: topscope_vars }
--- a/spec/bolt/catalog_spec.rb
+++ b/spec/bolt/catalog_spec.rb
@@ -45,7 +45,9 @@ describe Bolt::Catalog do
       'target' => {
         'name' => uri,
         'facts' => {},
-        'variables' => {},
+        'variables' => {
+          'roles' => %w[foo bar]
+        },
         'trusted' => {
           'authenticated' => 'local',
           'certname' => uri,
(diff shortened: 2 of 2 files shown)


lib/bolt/catalog.rb
```diff
@@ -76,7 +76,15 @@ def compile_catalog(request)
...
       target_vars = shadow_vars('target', target['variables'], target['facts'])
-      topscope_vars = target_vars.merge(plan_vars)
+
+      # Merge plan vars with target vars, while maintaining the order of the plan
+      # vars. It's critical that the order of plan vars is not changed, as Puppet
+      # will deserialize the variables in the order they appear. Variables may
+      # contain local references to variables that appear earlier in a plan. If
+      # these variables are moved before the variable they reference, Puppet will
+      # be unable to deserialize the data and raise an error.
+      topscope_vars = target_vars.reject { |k, _v| plan_vars.key?(k) }.merge(plan_vars)
```
nicklewis, line 86: Only one line of code? What a simple issue!

beechtom: image

lucywyman:
:+1: Confirmed the test fails "correctly" on main too:
```
       @@ -1,4 +1,5 @@
       -[["t1",
       +[["roles", {"__ptype"=>"LocalRef", "__pvalue"=>"$['t1'][0]['vars']['roles']"}],
       + ["t1",
          [{"__ptype"=>"Target",
            "config"=>{"ssh"=>{"connect-timeout"=>100, "host-key-check"=>false}},
            "facts"=>{},
       @@ -23,6 +24,5 @@
            "safe_name"=>"targetB",
            "target_alias"=>[],
            "uri"=>"targetb.example.com",
       -    "vars"=>{}}]],
       - ["roles", {"__ptype"=>"LocalRef", "__pvalue"=>"$['t1'][0]['vars']['roles']"}]]
       +    "vars"=>{}}]]]
```

PR #2058, opened 2020-08-04
(PE-29834) Do not rely on initializing Bolt::Logger for class methods

donoghuc:
Previously the mutex used for warn_once relied on calling the initialize_logging method which contains logging configuration specific to bolt. This commit move the mutex to the warn_once class method so that it can be used without having to have called the bolt-specific initialize_logging method.

### DIFF
--- a/lib/bolt/logger.rb
+++ b/lib/bolt/logger.rb
@@ -15,7 +15,6 @@ module Bolt
       return if Logging.initialized?
 
       Logging.init :debug, :info, :notice, :warn, :error, :fatal, :any
-      @mutex = Mutex.new
 
       Logging.color_scheme(
         'bolt',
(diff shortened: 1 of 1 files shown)


lib/bolt/logger.rb
```diff
@@ -105,6 +104,7 @@ def self.reset_logging
     end
 
     def self.warn_once(type, msg)
+      @mutex ||= Mutex.new
```
donoghuc, line 107: Honestly i'm having a hard time understanding what instance variables do when there is a module with only class methods. I'm not sure if this is safe or not?

nicklewis: This would be an instance variable of the module itself (which is an object). It will be shared between invocations of this method as you expect. It's not safe however, as this method could be called twice concurrently (the scenario the mutex is meant to guard against) and each caller could find `@mutex` uninitialized, create a Mutex and assign it.

I think the correct solution is probably to have the Mutex be a constant in the `Bolt::Logger` namespace that's initialized when the file is required.

donoghuc: That is what I was afraid of. Thanks for the info.

Issue #2004, opened 2020-07-18
bolt should have commands for generating task and plan skeletons 


npwalker:
## Use Case
Whenever, I go to make a new task I have to find a module that already has a task, copy the task and metadata file out and then start modifying.  I cannot remember what I need to make a task or a plan.  

## Describe the Solution You Would Like
`bolt task new` and `bolt plan new` or similar commands that lay down a skeleton task/plan in the right location.   

## Describe Alternatives You've Considered
I considered pdk but pdk requires a valid metadata.json because it expects to work with puppet modules.  It also creates a lot of extra baggage in the form of files and laying down tests, etc... 

pdk should also have a way to do this as well but it should be additional to whatever the commands in bolt do.  Bolt has the commands that provide the minimum to allow you to just make a simple task or plan and pdk should add things on top that help an advanced author test, lint, etc... the tasks / plans.  Ideally pdk would literally use bolt to lay down the same skeletons or they could share skeletons centrally somehow?

lucywyman:
Plans are easier than tasks, we should start with 'bolt new plan' as it's simpler to implement.

PR #2060, opened 2020-08-06
(GH-2003) 'bolt project init' now creates bolt-project.yaml

### LINKED ISSUE #2003: `bolt project init` should create a `bolt-project.yaml`
`bolt project init` should create a `bolt-project.yaml` with the `name` key filled out to the name of the current working directory.  

It should also no longer create a bolt.yaml which is deprecated.  

## Solution
- [x] This should change `bolt project init` to create `bolt-project.yaml` with `name: <current directory>`. If the current directory is not a valid project name, we should error and tell the user to change the dir name or specify a name.
- [x] This should change `bolt project init <name>` to create bolt-project.yaml with the specified name.
- [x] If bolt-project.yaml already exists we should not remove the existing file.
- [x] --modules should behave the same
- [x] If bolt.yaml exists we should error.

lucywyman:
This PR modernizes the behavior for the `bolt project init` command in a
few ways:

- Previously, `bolt project init` with no arguments would create an
  empty `bolt.yaml`file in the current working directory. It now creates
  a file `bolt-project.yaml` in the current working directory, with the
  content:
  ```
  ---
  name: <cwd_basename>
  ```
  where `cwd_basename` is the name of the directory.
- Previously the command took an optional argument `[directory]` with a
  path to where to create the project. The command no longer accepts the
  directory as an argument. It instead has an optional argument `[name]`
  which specifies the name of the project.
  ```
  $ bolt project init myproject
  $ cat bolt-project.yaml
  ---
  name: myproject
  ```
- If `bolt-project.yaml` or `bolt.yaml` exist, we warn and do not modify the file even if it
  has no content
- All module installation behavior is the same.

Closes #2003

!feature

* **'bolt project init' now creates bolt-project.yaml** ([#2003](https://github.com/puppetlabs/bolt/issues/2003))

  `bolt project init` now creates a `bolt-project.yaml` file instead of
  `bolt.yaml`. It additionally accepts a project name, and no longer
  accepts a directory to the project to create.

### DIFF
--- a/lib/bolt/cli.rb
+++ b/lib/bolt/cli.rb
@@ -770,8 +770,26 @@ module Bolt
     # Initializes a specified directory as a Bolt project and installs any modules
     # specified by the user, along with their dependencies
     def initialize_project
-      project    = Pathname.new(File.expand_path(options[:object] || Dir.pwd))
-      config     = project + 'bolt.yaml'
+      # Dir.pwd will return backslashes on Windows, but Pathname always uses
+      # forward slashes to concatenate paths. This results in paths like
+      # C:\User\Administrator/modules, which fail module install. This ensure
+      # forward slashes in the cwd path.
+      dir = File.expand_path(Dir.pwd)
+      name = options[:object] || File.basename(dir)
+      if name !~ Bolt::Module::MODULE_NAME_REGEX
+        if options[:object]
+          raise Bolt::ValidationError, "The provided project name '#{name}' is invalid; "\
+            "project name must begin with a lowercase letter and can include lowercase "\
+            "letters, numbers, and underscores."
+        else
+          raise Bolt::ValidationError, "The current directory name '#{name}' is an invalid "\
+            "project name. Please specify a name using 'bolt project init <name>'."
+        end
+      end
+
+      project    = Pathname.new(dir)
+      old_config = project + 'bolt.yaml'
+      config     = project + 'bolt-project.yaml'
       puppetfile = project + 'Puppetfile'
       modulepath = [project + 'modules']
 
@@ -792,18 +810,24 @@ module Bolt
 
       # Warn the user if the project directory already exists. We don't error here since users
       # might not have installed any modules yet.
+      # If both bolt.yaml and bolt-project.yaml exist, this will just warn
+      # about bolt-project.yaml and subsequent Bolt actions will warn about
+      # both files existing
       if config.exist?
-        @logger.warn "Found existing project directory at #{project}"
-      end
-
-      # Create the project directory
-      FileUtils.mkdir_p(project)
-
+        @logger.warn "Found existing project directory at #{project}. Skipping file creation."
+      # This won't get called if bolt-project.yaml exists
+      elsif old_config.exist?
+        @logger.warn "Found existing #{old_config.basename} at #{project}. "\
+          "#{old_config.basename} is deprecated, please rename to #{config.basename}."
       # Bless the project directory as a...wait for it...project
-      if FileUtils.touch(config)
-        outputter.print_message "Successfully created Bolt project at #{project}"
       else
-        raise Bolt::FileError.new("Could not create Bolt project directory at #{project}", nil)
+        begin
+          content = { 'name' => name }
+          File.write(config.to_path, content.to_yaml)
+          outputter.print_message "Successfully created Bolt project at #{project}"
+        rescue StandardError => e
+          raise Bolt::FileError.new("Could not create bolt-project.yaml at #{project}: #{e.message}", nil)
+        end
       end
 
       # Write the generated Puppetfile to the fancy new project
--- a/rakelib/pwsh.rake
+++ b/rakelib/pwsh.rake
@@ -200,11 +200,11 @@ namespace :pwsh do
             validate_not_null_or_empty: true
           }
         when 'project'
-          # bolt project init [directory] [options]
+          # bolt project init [name] [options]
           @pwsh_command[:options] << {
-            name:                       'Directory',
-            ruby_short:                 'd',
-            help_msg:                   'The directory to turn into a Bolt project',
+            name:                       'Name',
+            ruby_short:                 'n',
+            help_msg:                   'The name of the Bolt project to create',
             mandatory:                  false,
             type:                       'string',
             switch:                     false,
--- a/spec/bolt/cli_spec.rb
+++ b/spec/bolt/cli_spec.rb
@@ -2636,30 +2636,62 @@ describe "Bolt::CLI" do
     end
 
     context 'init' do
-      it 'creates a new project at the specified path' do
-        Dir.mktmpdir do |dir|
-          file = File.join(dir, 'bolt.yaml')
-          cli = Bolt::CLI.new(%W[project init #{dir}])
-          cli.execute(cli.parse)
-          expect(File.file?(file)).to be
-        end
-      end
+      it 'creates a new project in the current directory' do
+        Dir.mktmpdir do |tmpdir|
+          # The tmpdir doesn't have a valid modulename
+          dir = File.join(tmpdir, 'valid')
+          Dir.mkdir(dir)
+          allow(Dir).to receive(:pwd).and_return(dir)
+          file = File.join(dir, 'bolt-project.yaml')
 
-      it 'creates a new project in the current working directory' do
-        Dir.mktmpdir do |dir|
-          file = File.join(dir, 'bolt.yaml')
           cli = Bolt::CLI.new(%w[project init])
-          Dir.chdir(dir) { cli.execute(cli.parse) }
+          cli.execute(cli.parse)
+
           expect(File.file?(file)).to be
+          expect(File.read(file)).to eq("---\nname: valid\n")
         end
       end
 
-      it 'warns when a bolt.yaml already exists' do
-        Dir.mktmpdir do |dir|
-          config = File.join(dir, 'bolt.yaml')
-          cli    = Bolt::CLI.new(%W[project init #{dir}])
+      it 'creates a new project with a specified name' do
+        Dir.mktmpdir do |tmpdir|
+          dir = File.join(tmpdir, 'valid')
+          Dir.mkdir(dir)
+          allow(Dir).to receive(:pwd).and_return(dir)
+          file = File.join(dir, 'bolt-project.yaml')
 
+          cli = Bolt::CLI.new(%w[project init myproject])
+          cli.execute(cli.parse)
+
+          expect(File.file?(file)).to be
+          expect(File.read(file)).to eq("---\nname: myproject\n")
+        end
+      end
+
+      it 'errors if the directory name is invalid' do
+        Dir.mktmpdir do |dir|
+          allow(Dir).to receive(:pwd).and_return(dir)
+          cli = Bolt::CLI.new(%w[project init])
+          expect { cli.execute(cli.parse) }
+            .to raise_error(Bolt::ValidationError,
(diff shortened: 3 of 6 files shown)


lib/bolt/cli.rb
```diff
@@ -654,8 +654,21 @@ def generate_types
...
-      config     = project + 'bolt.yaml'
+      name       = options[:object] || File.basename(Dir.pwd)
+      name_regex = /^[a-z][a-z0-9_]*$/
+      if name !~ name_regex
+        if options[:object]
+          raise Bolt::ValidationError, "The provided project name '#{name}' is invalid; "\
+            "project name must match #{name_regex.inspect}"
+        else
+          raise Bolt::ValidationError, "The current directory name '#{name}' will be used "\
+            "as the project name, and is invalid. Project name must match #{name_regex.inspect}"
```
beechtom, line 665:
```suggestion
            "as the project name, and is invalid. Project name must begin with a lowercase "\
            "letter, and may only contain lowercase letters, digits, and underscores."
```

For users who aren't familiar with regex, this can be cryptic. I think adding a more verbose description of the requirements, similar to [Bolt projects documentation](https://puppet.com/docs/bolt/latest/projects.html#creating-a-bolt-project), would be helpful.

lib/bolt/cli.rb
```diff
@@ -654,8 +654,21 @@ def generate_types
     # Initializes a specified directory as a Bolt project and installs any modules
     # specified by the user, along with their dependencies
     def initialize_project
-      project    = Pathname.new(File.expand_path(options[:object] || Dir.pwd))
-      config     = project + 'bolt.yaml'
+      name       = options[:object] || File.basename(Dir.pwd)
+      name_regex = /^[a-z][a-z0-9_]*$/
```
beechtom, line 658:
```suggestion
      name_regex = /\A[a-z][a-z0-9_]*\Z/
```

lib/bolt/cli.rb
```diff
@@ -676,18 +689,27 @@ def initialize_project
...
 
-      # Create the project directory
-      FileUtils.mkdir_p(project)
-
       # Bless the project directory as a...wait for it...project
-      if FileUtils.touch(config)
-        outputter.print_message "Successfully created Bolt project at #{project}"
-      else
-        raise Bolt::FileError.new("Could not create Bolt project directory at #{project}", nil)
+      unless old_config.exist? || config.exist?
```
nicklewis, line 705: This can just be an `else` for the above `if/elsif`.

spec/bolt/cli_spec.rb
```diff
@@ -2475,44 +2475,108 @@ def stub_config(file_content = {})
...
+
+          cli = Bolt::CLI.new(%w[project init valid])
+          expect { cli.execute(cli.parse) }
+            .to raise_error(Bolt::Error, /Could not move #{old_config} to bolt-project.yaml/)
+        end
+      end
+
       context 'with modules' do
-        it 'creates a Puppetfile and installs modules with dependencies' do
+        it 'creates a Puppetfile and installs modules with dependencies', ssh: true do
```
lucywyman, line 2571: I'm still debugging why this is failing in CI, but it works manually on Windows so...I suspect it's a CI issue

lib/bolt/cli.rb
```diff
@@ -676,18 +689,27 @@ def initialize_project
...
+      # about bolt-project.yaml and subsequent Bolt actions will warn about
+      # both files existing
       if config.exist?
-        @logger.warn "Found existing project directory at #{project}"
+        @logger.warn "Found existing project directory at #{project}. Skipping file creation."
+      # This won't get called if bolt-project.yaml exists
+      elsif old_config.exist?
+        @logger.warn "Found existing #{old_config.basename} at #{project}. "\
+        "Attempting to update to #{config.basename}."
+        update_project_config(old_config)
```
nicklewis, line 701: This seems more like `bolt project migrate` type behavior... I don't have strong feelings about it though.

beechtom, line 701: This should error with a helpful message instead of attempting to convert the `bolt.yaml` file to `bolt-project.yaml`.

lucywyman: It definitely is, I wasn't sure about adding a whole other subcommand as part of this though. and it feels like it follows the principle of "I'm going to tell you this is wrong and how to fix it" instead of "I can tell this is wrong so I'm going to fix it if I can".

lib/bolt/cli.rb
```diff
@@ -654,8 +654,21 @@ def generate_types
...
+            "project name must match #{name_regex.inspect}"
+        else
+          raise Bolt::ValidationError, "The current directory name '#{name}' will be used "\
+            "as the project name, and is invalid. Project name must match #{name_regex.inspect}"
+        end
+      end
+
+      project    = Pathname.new(Dir.pwd)
+      old_config = project + 'bolt.yaml'
+      config     = project + 'bolt-project.yaml'
```
beechtom, line 671: Since we use `'bolt.yaml'` and `'bolt-project.yaml'` in several locations, I wonder if it makes sense to define those as a constant in `Bolt::Project`, similar to `Bolt::Project::BOLTDIR_NAME`: https://github.com/puppetlabs/bolt/blob/main/lib/bolt/project.rb#L9

lib/bolt/cli.rb
beechtom: Also, since this regex is also used in `Bolt::Project`, I think it might make sense to define it as a constant so it can be reused elsewhere. I made this change in #2004 for reference: https://github.com/puppetlabs/bolt/pull/2061/files#diff-5fb12f608b24f29bc6e72995332d2c24R11

lib/bolt/cli.rb
lucywyman: "Must start with a letter and can include numbers, letters, or underscores". I cribbed this directly from the project class, so...I'll update there too.

nicklewis: "Must not be able to moonlight as a strong password"

lib/bolt/cli.rb
lucywyman: Is it possible to open a separate PR for it if it's going to be used here too? I don't want to duplicate work, or for this to unnecessarily rely on #2004.

lib/bolt/cli.rb
lucywyman: I was using the principle of "fix it instead of telling the user it's wrong and that they should fix it" - is there a reason you think it should error instead?

lib/bolt/cli.rb
lucywyman: If we do I don't think it should be here - we use those a lot throughout the codebase. On the minus side, `Bolt::Project::BOLT_CONFIG` is way more verbose than `bolt.yaml`.

lib/bolt/cli.rb
beechtom: Ya, I'll throw something up real quick.

beechtom: https://github.com/puppetlabs/bolt/pull/2063 has the change

lib/bolt/cli.rb
nicklewis: My thinking is that what's "wrong" here isn't which file exists on disk, but the command the user typed. It's not that they have a bolt.yaml when they need a bolt-project.yaml, but that they tried to use `bolt project init` in a directory that is already a project. I'm on the fence about it though.

lib/bolt/cli.rb
nicklewis: I can't think of names to give those constants that wouldn't be confusing. `'bolt.yaml'` and `'bolt_project.yaml'` are at least obvious when you see them written.

lib/bolt/cli.rb
lucywyman: Ugh, I just realized we already have this in https://github.com/puppetlabs/bolt/blob/8d3e2f207daf162aeceec12621553ecdf9f5d48a/lib/bolt/module.rb#L6. I'm just going to modify this PR to use that instead.

lucywyman: And in a way that seems better, since the only reason we limit the project name is because it needs to be loaded as a module.

lib/bolt/cli.rb
beechtom: It's mostly because in planning we had decided that it should error if there's a `bolt.yaml`. I'm not strictly opposed to having it convert a `bolt.yaml` if it's able to. However, if it does convert it I think we might as well add the `name` key for the user. It seems a little strange to go through the effort of converting the file for them but then have them manually enter the name.

lib/bolt/cli.rb
```diff
@@ -652,8 +652,27 @@ def generate_types
     # Initializes a specified directory as a Bolt project and installs any modules
     # specified by the user, along with their dependencies
     def initialize_project
-      project    = Pathname.new(File.expand_path(options[:object] || Dir.pwd))
-      config     = project + 'bolt.yaml'
+      # Dir.pwd will return backslashes on Windows, but Pathname always uses
+      # forward slashes to concatenate paths. This results in paths like
+      # C:\User\Administrator/modules, which fail module install. This ensure
+      # forward slashes in the cwd path.
```
beechtom, line 658: Both `/` and `\` are valid path separators on Windows, and should not affect how the path is interpreted by any of the `File` methods, even if the two separators are used interchangeably. For example, the following shows that use of both path separators does not affect either `Pathname` or `File` methods:

```
path = Pathname.new('C:\Users\tom')
=> #<Pathname:C:\Users\tom>
bolt_path = path + 'bolt'
=> #<Pathname:C:\Users\tom/bolt>
bolt_path.exist?
=> true
File.exist?(bolt_path.to_s)
=> true
File.write(bolt_path + 'foo.txt', 'foo')
=> 3
```

Installing modules in the cwd with `bolt project init --modules puppetlabs/apache` also works on Windows with Bolt 2.21.0:

```
> New-BoltProject -Modules puppetlabs/apache
Successfully created Bolt project at C:\Users\tom/bolt
Successfully created Puppetfile at C:\Users\tom/bolt/Puppetfile
...
```

lib/bolt/cli.rb
lucywyman: Hm, I initially didn't add 'name' because I was thinking the way to do it would be to read in yaml, add the key, then write it out, which removes comments and reorders things. But I think we might be able to safely just append the name to the config file...that would override a name if it was already there, but that'd be a particularly weird case.

lib/bolt/cli.rb
```diff
@@ -652,8 +652,27 @@ def generate_types
...
+      dir = File.expand_path(Dir.pwd)
+      name = options[:object] || File.basename(dir)
+      if name !~ Bolt::Module::MODULE_NAME_REGEX
+        if options[:object]
+          raise Bolt::ValidationError, "The provided project name '#{name}' is invalid; "\
+            "project name must begin with a lowercase letter and can include lowercase "\
+            "letters, numbers, and underscores."
+        else
+          raise Bolt::ValidationError, "The current directory name '#{name}' will be used "\
```
nicklewis, line 667: We should advise the user to supply a name. This message seems to read like I should rename the directory in order to proceed.

rakelib/pwsh.rake
```diff
@@ -200,11 +200,11 @@ namespace :pwsh do
             validate_not_null_or_empty: true
           }
         when 'project'
-          # bolt project init [directory] [options]
+          # bolt project init [name] [options]
           @pwsh_command[:options] << {
-            name:                       'Directory',
-            ruby_short:                 'd',
-            help_msg:                   'The directory to turn into a Bolt project',
+            name:                       'name',
```
beechtom, line 205:
```suggestion
            name:                       'Name',
```

PR #2071, opened 2020-08-11
(Docs) Overriding a packaged module

hestonhoffman:
!no-release-note

### DIFF
--- a/documentation/bolt_installing_modules.md
+++ b/documentation/bolt_installing_modules.md
@@ -35,8 +35,7 @@ To install a module:
    2. Run the `bolt puppetfile install` command. Bolt installs modules to the
       first directory in the `modulepath` setting. By default, this is the
       `modules/` subdirectory inside the Bolt project directory. To override
-      this location, update the `modulepath` setting in the [Bolt config
-      file](bolt_configuration_reference.md).
+      this location, update the `modulepath` setting in your [project configuration file](bolt_project_reference.md).
 
 ## Create a new Bolt project and install a list of modules with dependencies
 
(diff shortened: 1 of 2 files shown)


documentation/bolt_installing_modules.md
```diff
@@ -131,18 +130,25 @@ these core modules.
...
 
-> 🔩 **Tip:** To override a packaged module with another version, download the
-> version you want and configure your module path to point to it.
+## Overriding a packaged module
+
+To override a packaged module with another version, download the
+version you need and configure your module path to point to it.
+
+1. Create a directory for your module. For example, `module-override`.
+2. Download the version of the module you need and place it in the `module-override` directory.
```
lucywyman, line 146: Users shouldn't have to modify the modulepath to override a packaged module, their own modulepath is already in front of the packaged one. Just installing the module from their Puppetfile into `modules` will override the packaged one.

hestonhoffman: Okay cool. I'll update with your suggestion

documentation/bolt_installing_modules.md
```diff
@@ -78,6 +77,11 @@ This list of packaged modules is available in a
 Bolt repository. The modules and supporting documentation are publicly available
 on the [Puppet Forge](https://forge.puppet.com/).
 
+If you need to use a specific version of a packaged module, you can override the
+packaged version by installing the module into your `modules/` directory.
```
lucywyman, line 81: I'm not sure if this is worth detailing, or if it's TMI, but modules installed anywhere on the modulepath will override packaged modules. Where I think that might get confusing is if the user *doesn't* have `modules/` on the modulepath (for example they've configured their modulepath and don't have `<project>/modules` there) then the installed module won't get loaded at all and won't override packaged modules.

PR #2073, opened 2020-08-11
(GH-2072) Add 'interval' option to 'ctrl::do_until' function

### LINKED ISSUE #2072: Add `interval` parameter to ctrl::do_until
## Use Case
I want to sleep between loops when hitting an API, so that the API status endpoint isn't overwhelmed with requests while waiting for the status to update. It'd be nice to have a sleep interval parameter instead of having to do:

```
ctrl::do_until() || {
  if run_task(...).ok() {
    true
  } else {
    ctrl::sleep(2)
    false
  }
}
```

* An optional parameter `interval` to ctrl::do_until

beechtom:
This adds an `interval` option to the `ctrl::do_until` plan function. It
accepts a numeric value specifying the number of seconds to sleep
before repeating the block again.

Closes #2072 

!feature

* **Set interval for repeating block in `ctrl::do_until` plan function**
  ([#2072](https://github.com/puppetlabs/bolt/issues/2072))

  The `ctrl::do_until` plan function now accepts an `interval` option.
  This option accepts a numeric value that specifies the number of
  seconds to wait before repeating the block.

### DIFF
--- a/bolt-modules/ctrl/spec/functions/ctrl/do_until_spec.rb
+++ b/bolt-modules/ctrl/spec/functions/ctrl/do_until_spec.rb
@@ -31,4 +31,25 @@ describe 'ctrl::do_until' do
     end.and_return("truthy"))
     expect(seq.length).to eq(1)
   end
+
+  it 'sleeps with an interval' do
+    Kernel.expects(:sleep).with(1).times(2)
+
+    is_expected.to(run.with_params('interval' => 1).with_lambda do
+      seq.pop
+    end.and_return('truthy'))
+
+    expect(seq.length).to eq(1)
+  end
+
+  it 'does not sleep if first value is truthy' do
+    seq << 'truthy'
+    Kernel.expects(:sleep).never
+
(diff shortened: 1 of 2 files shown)


bolt-modules/ctrl/spec/functions/ctrl/do_until_spec.rb
```diff
@@ -31,4 +31,15 @@
...
   end
+
+  it 'sleeps with an interval' do
+    now = Time.now
+
+    is_expected.to(run.with_params('interval' => 1, 'limit' => 3).with_lambda do
+      seq.pop
+    end.and_return('truthy'))
+
+    expect(Time.now - now).to be >= 1
```
nicklewis, line 42: Maybe stub sleep and check that it's called twice? I would like to know that it's called after each invocation. Also it would be good to test that it's _not_ called when the first item is truthy  (ie. it doesn't sleep _unless_ it has to repeat the loop)

PR #2074, opened 2020-08-12
(maint) Do not set facts for script compiler scope

beechtom:
This updates `Bolt::PAL#in_bolt_compiler` to set the `set_local_facts`
option for the script compiler to `false`. Setting this will disable
setting facts, trusted facts, and server facts in the script compiler's
scope.

Changes made to Puppet in puppetlabs/puppet#8063 updated the script
compiler to never set facts in the script compiler's scope. However,
this resulted in breaking changes to PAL's public API. The work in
puppetlabs/puppet#8268 reverses these changes and adds the
`set_local_facts` option to allow users to disable setting facts if they
wish. By default, `set_local_facts` is set to `true`.

!no-release-note

### DIFF
--- a/lib/bolt/pal.rb
+++ b/lib/bolt/pal.rb
@@ -151,7 +151,7 @@ module Bolt
         # Only load the project if it a) exists, b) has a name it can be loaded with
         Puppet.override(bolt_project: @project,
                         yaml_plan_instantiator: Bolt::PAL::YamlPlan::Loader) do
-          pal.with_script_compiler do |compiler|
+          pal.with_script_compiler(set_local_facts: false) do |compiler|
             alias_types(compiler)
             register_resource_types(Puppet.lookup(:loaders)) if @resource_types
             begin


lib/bolt/pal.rb
```diff
@@ -157,7 +157,7 @@ def in_bolt_compiler
         Puppet.override(bolt_project: bolt_project,
                         bolt_project_data: bolt_project_data,
                         yaml_plan_instantiator: Bolt::PAL::YamlPlan::Loader) do
-          pal.with_script_compiler do |compiler|
+          pal.with_script_compiler(disable_facts: true) do |compiler|
```
lucywyman, line 160: @puppetlabs/skeletor I think this will also need to update here: https://github.com/puppetlabs/orchestrator/blob/70e55aae51084cf08c32515a0cd98bb4eb6ad86c/ruby/bolt-plan-executor-lib/bolt_plan_executor/pe_pal.rb#L119 For context, we had to revert removing $facts / trusted facts / server facts from the script compiler in Puppet PAL because it broke Choria. For now at least we're just going to add a `disable_facts` keyword argument to `with_script_compiler` to tell Puppet to not load those in the Bolt context (see https://github.com/puppetlabs/puppet/pull/8268).

beechtom: Updated this to be `set_local_facts` instead of `disable_facts` per @nicklewis 's review: https://github.com/puppetlabs/puppet/pull/8268#discussion_r469429071

PR #2075, opened 2020-08-12
(bugfix) Don't remove Bolt modulepaths from the modulepath when listing

lucywyman:
When we print the modulepath for users when listing tasks or plans, we
previously would calculate the modulpath by subtracting the
Bolt-packaged modulepaths `boltlib` and `modules` from the full
modulepath. This erroneously removes `modules` from the modulepath if it
is listed in the user's path though, which is confusing. This modifies
the modulepath to be calculated from the original modulepath the user
supplied, so that any supplied modulepaths aren't incorrectly removed.

!bugfix

* **Include modules/ in displayed modulepath if it's in the user-configured modulepath**
    
  Bolt now includes `<bolt-installation-directory>/modules` in the displayed
  modulepath if the user has the path as a component of their configured
  modulepath. If installed as a package on *nix,
  bolt-installation-directory would be `/opt/puppetlabs/bolt/lib/ruby/gems/2.5.0/gems/bolt-x.y.z/`.

### DIFF
--- a/lib/bolt/pal.rb
+++ b/lib/bolt/pal.rb
@@ -48,7 +48,7 @@ module Bolt
       end
     end
 
-    attr_reader :modulepath
+    attr_reader :modulepath, :user_modulepath
 
     def initialize(modulepath, hiera_config, resource_types, max_compiles = Etc.nprocessors,
                    trusted_external = nil, apply_settings = {}, project = nil)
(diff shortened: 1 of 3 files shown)


lib/bolt/pal.rb
```diff
@@ -279,7 +279,7 @@ def list_tasks
     end
 
     def list_modulepath
-      @modulepath - [BOLTLIB_PATH, MODULES_PATH]
+      @original_modulepath
```
nicklewis, line 282: I think we can just make this an `attr_reader`. Maybe rename the variable to `@user_modulepath` to make its meaning clear though?

PR #2077, opened 2020-08-12
(bug) Return empty hash when analytics file is empty

beechtom:
When Bolt loads the anayltics config file, it uses the `YAML.load`
method. If the file is empty, or otherwise cannot be parsed as YAML,
this method returns `false`. This causes an error when Bolt checks the
loaded config for the `disabled` key, as the returned value is not a
hash.

This updates the `Bolt::Analytics.load_config` method to return an empty
hash when the analytics config file cannot be parsed as YAML.

!bug

* **Do not error when analytics configuration file is empty**

  Empty analytics configuration files would cause Bolt to raise a
  `NoMethodError`. Now, if an analytics configuration file is empty,
  Bolt will instead rewrite the file.

### DIFF
--- a/lib/bolt/analytics.rb
+++ b/lib/bolt/analytics.rb
@@ -72,7 +72,7 @@ module Bolt
 
     def self.load_config(filename, logger)
       if File.exist?(filename)
-        YAML.load_file(filename)
+        Bolt::Util.read_optional_yaml_hash(filename, 'analytics')
       else
         unless ENV['BOLT_DISABLE_ANALYTICS']
           logger.warn <<~ANALYTICS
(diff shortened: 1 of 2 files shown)


lib/bolt/analytics.rb
```diff
@@ -72,7 +72,9 @@ def self.config_path(logger)
 
     def self.load_config(filename, logger)
       if File.exist?(filename)
-        YAML.load_file(filename)
+        data = YAML.load_file(filename)
```
nicklewis, line 75: Can we use `Bolt::Util.read_optional_yaml_hash`?

beechtom: ... ya. Updated. :D

nicklewis: That will still raise an error if the file has _invalid_ content, but that seems okay to me.

PR #2080, opened 2020-08-12
(GH-2051) Always set `bolt_project` Puppet setting

### LINKED ISSUE #2051: Always add Bolt project to PAL temp environment
## Use Case
The `Bolt::Project` class includes helpful contextual information for the current project, including the path to the project. Currently, this class is added to PAL's temp environment when the project is a named project (see https://github.com/puppetlabs/bolt/blob/main/lib/bolt/pal.rb#L152). However, if the project is not named, PAL does not receive this class.

When the `download_file` function was created, it needed the project's path so files could be downloaded to the project's `downloads` directory. Since the `Bolt::Project` class was only sent to PAL when the project if the project was named, it was necessary to ensure this class is always sent to PAL (see https://github.com/puppetlabs/bolt/blob/main/lib/bolt/pal.rb#L158). This means that whenever a user is running a named project, PAL will receive duplicate data.

We should update PAL/Bolt so Bolt will _always_ send the `Bolt::Project` class to PAL when creating a temp environment. We should then update Puppet to create a contract with Bolt so Puppet knows whether the project should be loaded as a module. This would include Puppet calling a method such as `named_project?` to ask Bolt whether the project is named and should be loaded as a module. (see https://github.com/puppetlabs/puppet/blob/4a9558230596721049c4dbbc084c0bd89ab6d125/lib/puppet/node/environment.rb#L300)

## Solution
We should add a method to `Bolt::Project` called 'load_as_module?' which tells puppet whether to load the project as a module, and which currently will switch on whether the project has a name key in `bolt-project.yaml`.
Update Puppet.

lucywyman:
This adds a method `load_as_module?` to the `Bolt::Project` class to use
in Puppet to determine whether to load the project as a module.
Previously, `bolt_project` would only be set if the project was named,
which isn't necessary for a valid project, and Puppet would only load
the project as a module if the setting was set. Puppet can now use
`load_as_module?` to determine whether to load the project as a module,
which lets us always set the `bolt_project` Puppet setting. This lets us
use that setting to access the project from within Bolt plan functions
for things like getting the project path to set the destination to
download files to.

This PR also removes the now-unnecessary `bolt_project_data` Puppet
setting, which was an interim means of accessing the project path from
plan functions.

Closes #2051

!no-release-note

### DIFF
--- a/bolt.gemspec
+++ b/bolt.gemspec
@@ -48,7 +48,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency "net-ssh", ">= 4.0"
   spec.add_dependency "net-ssh-krb", "~> 0.5"
   spec.add_dependency "orchestrator_client", "~> 0.4"
-  spec.add_dependency "puppet", [">= 6.16.0", "< 6.18.0"]
+  spec.add_dependency "puppet", "6.18.0"
   spec.add_dependency "puppetfile-resolver", "~> 0.1.0"
   spec.add_dependency "puppet-resource_api", ">= 1.8.1"
   spec.add_dependency "puppet-strings", "~> 2.3"
--- a/lib/bolt/applicator.rb
+++ b/lib/bolt/applicator.rb
@@ -183,11 +183,10 @@ module Bolt
                                                                        type_by_reference: true,
                                                                        local_reference: true)
 
-      bolt_project = @project if @project&.name
       scope = {
         code_ast: ast,
         modulepath: @modulepath,
-        project: bolt_project.to_h,
+        project: @project.to_h,
         pdb_config: @pdb_client.config.to_hash,
         hiera_config: @hiera_config,
         plan_vars: plan_vars,
--- a/lib/bolt/pal.rb
+++ b/lib/bolt/pal.rb
@@ -149,13 +149,7 @@ module Bolt
       setup
       r = Puppet::Pal.in_tmp_environment('bolt', modulepath: @modulepath, facts: {}) do |pal|
         # Only load the project if it a) exists, b) has a name it can be loaded with
-        bolt_project = @project if @project&.name
-        # Puppet currently won't receive the project unless it is a named project. Since
-        # the download_file plan function needs access to the project path, add it to the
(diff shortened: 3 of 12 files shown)


lib/bolt/pal.rb
```diff
@@ -149,13 +149,8 @@ def in_bolt_compiler
       setup
       r = Puppet::Pal.in_tmp_environment('bolt', modulepath: @modulepath, facts: {}) do |pal|
         # Only load the project if it a) exists, b) has a name it can be loaded with
-        bolt_project = @project if @project&.name
-        # Puppet currently won't receive the project unless it is a named project. Since
-        # the download_file plan function needs access to the project path, add it to the
-        # context.
-        bolt_project_data = @project
+        bolt_project = @project
         Puppet.override(bolt_project: bolt_project,
```
beechtom, line 153:
```suggestion
        Puppet.override(bolt_project: @project,
```

lib/bolt/applicator.rb
```diff
@@ -183,7 +183,7 @@ def apply_ast(raw_ast, targets, options, plan_vars = {})
                                                                        type_by_reference: true,
                                                                        local_reference: true)
 
-      bolt_project = @project if @project&.name
+      bolt_project = @project
```
beechtom, line 186: This could just go directly in the `scope` hash as `@project.to_h`.

bolt.gemspec
```diff
@@ -48,7 +48,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency "net-ssh", ">= 4.0"
   spec.add_dependency "net-ssh-krb", "~> 0.5"
   spec.add_dependency "orchestrator_client", "~> 0.4"
-  spec.add_dependency "puppet", [">= 6.16.0", "< 6.18.0"]
+  spec.add_dependency "puppet", [">= 6.16.0", "< 6.19.0"]
```
beechtom, line 51: We should probably set the minimum version to 6.18.0, since Bolt will no longer work with < 6.18.0.

PR #2085, opened 2020-08-12
(docs) Remove bolt projects from experimental features page

beechtom:
This removes Bolt projects from the experimental features doc, which was
missed when projects were moved out of experimental.

!no-release-note


hestonhoffman:
🚢

Issue #2078, opened 2020-08-12
Create Bolt topics pages for CLI


beechtom:
## Use Case
Users should be able to view help pages for Bolt topics from the CLI. For example, I'm using Bolt for the first time and receive a warning or error that tells me about a topic that I'm unfamiliar with, such as a modulepath. I should be able to use a command such as `bolt help modulepath` to have Bolt list information about what a modulepath is. The help page could potentially link to Bolt documentation as well.

Having a `bolt help` command could help users more easily find answers to their questions on their own, as well as help surface other features and topics they may be interested in or need to learn about. For example, the `bolt help` command could accept a `--topics` option (or something similar) to display a full list of topics that the user can display help for.

This feature would also pair nicely with https://github.com/puppetlabs/bolt/issues/1465 (adding a glossary page), as the help text for each topic could be used in both places.

PR #2088, opened 2020-08-13
(GH-2078) Add Bolt guide pages

beechtom:
This adds a new CLI command, `bolt guide`, which displays information
about various Bolt topics. The guides are read from text files that are
saved in the `guides` directory, and are only loaded when a specific
guide is being requested.

To show a list of available guides, use the `bolt guide` command:

  ```
  $ bolt guide
  ```

To view the guide for a specific topic, use the `bolt guide <topic>`
command:

  ```
  $ bolt guide inventory
  ```

This also adds an analytics event for both known guides and unknown
guides. This allows Bolt to collect data about which guides users are
viewing and which guides users may expect to be available.

!feature

* **View information about Bolt topics with the `bolt guide` command**
  ([#2078](https://github.com/puppetlabs/bolt/issues/2078))

  The new CLI command `bolt guide` can be used to view information about
  different Bolt topics, such as inventory and projects. Users can view
  a list of available guides using the `bolt guide` command, or view a
  specific guide using the `bolt guide <topic>` command.


guides/inventory.txt
```diff
@@ -0,0 +1,45 @@
...
+              password: puppet
+        - name: windows
+          targets:
+            - win-1.example.org
+            - win-2.example.org
+          config:
+            transport: winrm
+            winrm:
+              user: bolt
+              password: puppet
```
beechtom, line 41: Not really sure how I feel about the 'examples' section... I'm leaning towards leaving it to the docs to show examples.

hestonhoffman: Yeah, I agree. I think the docs are a better place for examples

guides/project.txt
```diff
@@ -0,0 +1,39 @@
...
+
+    When you run Bolt, it runs in the context of a project. If the directory you
+    run Bolt from is not a project, then Bolt will attempt to find a project by
+    traversing the parent directories. If Bolt is unable to find a project, it
+    will run from the default project, located at '~/.puppetlabs/bolt'.
+
+    A directory is only considered a Bolt project when it has a project
+    configuration file named 'bolt-project.yaml'. Project data and content,
+    including inventory files, are not loaded by Bolt unless they are part of
+    a project.
```
hestonhoffman, line 18:
```suggestion
    A directory is only considered a Bolt project when it has a project
    configuration file named 'bolt-project.yaml'. Bolt doesn't load project data 
    and content, including inventory files, unless the data and content are 
    part of a project.
```

guides/inventory.txt
nicklewis: We could add `bolt example inventory` or something later if we decide it's needed. It doesn't seem to fit in the guide though.

guides/inventory.txt
```diff
@@ -0,0 +1,45 @@
...
+
+DOCUMENTATION
+    pup.pt/bolt-inventory
```
nicklewis, line 44: `https://` would be helpful here as it's not totally obvious that these are URLs...

lib/bolt/outputter/human.rb
```diff
@@ -286,6 +286,18 @@ def print_plans(plans, modulepath)
                         "details and parameters for a specific plan.")
       end
 
+      def print_topics(topics)
+        @stream.puts(topics.join("\n"))
```
nicklewis, line 290: A header like "Available topics are:" (similar to what you get if you enter an invalid topic) might be helpful here.

guides/inventory.txt
beechtom: They do kind of look like Bolt error kinds, don't they?

lib/bolt/bolt_option_parser.rb
```diff
@@ -61,6 +61,9 @@ def get_help_text(subcommand, action = nil)
           { flags: OPTIONS[:global],
             banner: GROUP_HELP }
         end
+      when 'guide'
+        { flags: OPTIONS[:global],
```
lucywyman, line 65: Hm,  I don't think the extra flags are doing any harm here, but it does seem like `--help` is the only flag you could *actually* use with this command. We might eventually want to separate those out.

jpogran:
This is a great idea, but we don't have to do a whole new powershell command in order to surface this, 'topic' based help is built into powershell and is returned by the get-help command. If name the files correctly and we ship them inside the MSI in the places PowerShell expects, it will pick them up and respond when `get-help` is called. We can leave the command accessible on Windows platforms, but don't have to add it to our generator. The format for the files is exactly like you already have, so we're halfway there.

We can leave the source as is, but in vanagon have the following created:

Source | MSI
-- | --
guides/inventory.txt | en-US/about_bolt_inventory.help.txt
guides/project.txt | en-US/about_bolt_project.help.txt

Which will look like this deployed:

```
PS C:\Program Files\WindowsPowerShell\Modules\PuppetBolt> ls -rec

    Directory: C:\Program Files\WindowsPowerShell\Modules\PuppetBolt

Mode                 LastWriteTime         Length Name
----                 -------------         ------ ----
d-----         8/17/2020   1:32 PM                en-US
-a----         8/10/2020   8:34 PM          12694 PuppetBolt.psd1
-a----         8/13/2020  12:07 PM          89817 PuppetBolt.psm1

    Directory: C:\Program Files\WindowsPowerShell\Modules\PuppetBolt\en-US

Mode                 LastWriteTime         Length Name
----                 -------------         ------ ----
-a----         8/17/2020   1:32 PM              0 about_bolt_inventory.help.txt
-a----         8/17/2020   1:32 PM              0 about_bolt_project.help.txt
```

Which then shows up as:

image


We can also add 'links' to the commands that are covered by these topics inside their comment based help:

```
.LINK
  about_bolt_inventory
```

lib/bolt/bolt_option_parser.rb
lucywyman: Actually `--format` is also valid for this, which *isn't* currently listed.

lib/bolt/outputter/rainbow.rb
```diff
@@ -86,6 +86,13 @@ def print_summary(results, elapsed_time = nil)
         total_msg << " in #{duration_to_string(elapsed_time)}" unless elapsed_time.nil?
         @stream.puts colorize(:rainbow, total_msg)
       end
+
```
lucywyman, line 89: Do we want to override `print_topics` too? And `print_message` if `bolt guide invalidtopic` should be rainbow too. I think it's also fine to just leave this as is.

spec/bolt/cli_spec.rb
```diff
@@ -89,6 +89,81 @@ def stub_config(file_content = {})
     end
   end
 
+  context 'guide' do
+    let(:guides_path) { File.expand_path(File.join(__dir__, '..', '..', 'guides')) }
+    let(:config)      { double('config', format: nil) }
+
+    let(:guides) do
+      Dir.children(guides_path).sort.each_with_object({}) do |child, acc|
```
lucywyman, line 97: This is the same logic used in `cli.rb`, so if the logic fails this will still pass. I think you want to hardcode the expected output in this case.

lib/bolt/bolt_option_parser.rb
lucywyman: Er, I think this still needs to be addressed.  Once it is I'm :+1:

beechtom: D'oh. Updated.

lib/bolt/outputter/human.rb
```diff
@@ -286,6 +286,19 @@ def print_plans(plans, modulepath)
...
       end
 
+      def print_topics(topics)
+        print_message("Available topics are:")
+        print_message(topics.join("\n"))
+        print_message("\nUse `bolt guide <topic>` to view a specific guide.")
+      end
+
+      def print_guide(filepath, _topic)
+        content = File.read(filepath)
```
nicklewis, line 296: Can we handle loading the file in the CLI so it's not duplicated across each outputter?

beechtom: Updated.

PR #2089, opened 2020-08-14
(GH-2081) Update bolt project migrate

### LINKED ISSUE #2081: Update bolt project migrate
## Use Case
We should update the bolt project migrate command to also migrate bolt.yaml to bolt-project.yaml and inventory.yaml. It'd be nice if for the inventory it gets written to a separate yaml file and then included using a plugin to the main inventory, to make it easily editable.

- `bolt project migrate`
- We should maintain the old `project migrate` behavior.
- We should error out if bolt-project.yaml already exists
- We should print a useful message encouraging users to add a name, and link to documentation about the updated project files

lucywyman:
This updates the `bolt project migrate` command to rename `bolt.yaml` to
`bolt-project.yaml` in addition to it's previous behavior of updating a
version 1 inventory to version 2. There are two changes that support
this addition:
- The `migrate_project` handler method moved to be in the
  `Bolt::Project` class itself. This is purely organizational to avoid
  overloading the `Bolt::CLI` class, and because the command is acting
  within the context of a project.
- Backup any files before modifying them. We now backup inventory and config
  files to `<project>/.bolt-bak/<filename>.<date>.bak` before rewriting
  them with updated content. This is to preserve any valuable formatting,
  and to ensure that the file is recoverable should something go wrong
  while writing out the new file.

When migrating a project, any transport configuration in `bolt.yaml` is
removed from that file and merged with the 'all' group transport
configuration in the inventory, with the inventory config taking
precedence. `bolt.yaml` is then renamed to `bolt-project.yaml`. If the
inventory file doesn't already exist and transport config needs to be
added, the inventory file is created. If no config needs to be moved,
the inventory isn't created.

If `bolt-project.yaml` already exists this will not overwrite it. If
`bolt.yaml` does not exist it will not create a `bolt-project.yaml`.

Closes #2081

!feature

* **bolt project migrate now updates project files** ([#2081](https://github.com/puppetlabs/bolt/issues/2081))

  The `bolt project migrate` command will now update bolt.yaml to
  bolt-project.yaml and move transport configuration to inventory.yaml.


lib/bolt/project.rb
```diff
@@ -167,5 +167,131 @@ def check_deprecated_file
...
+            raise Bolt::FileError.new("#{e.message}; unable to write inventory.", migrating_inv)
+          end
+          # rubocop:enable Lint/ShadowedException
+        end
+
+        outputter.print_message "Renaming bolt.yaml to bolt-project.yaml"
+        FileUtils.mv(config_file, project_file)
+        outputter.print_message "Successfully updated project. Please add a "\
+          "'name' key to bolt-project.yaml to use project-level tasks and plans. "\
+          "Learn more about projects by running 'bolt guide project'"
```
beechtom, line 249:
```suggestion
          "Learn more about projects by running 'bolt guide project'."
```

lib/bolt/project.rb
```diff
@@ -167,5 +167,131 @@ def check_deprecated_file
         Bolt::Logger.deprecation_warning('Using project.yaml instead of bolt-project.yaml', msg)
       end
     end
+
+    def migrate_project(configured_inventory_path, outputter)
+      migrating_inv = configured_inventory_path || inventory_file
+      inventory_1_to_2(migrating_inv, outputter) if migrating_inv.file?
+      bolt_yaml_to_bolt_project(migrating_inv, outputter)
+    end
```
beechtom, line 175: This needs to return a code (`0` or `1`), otherwise you'll get an error after migrating:

```
Traceback (most recent call last):
	5: from /Users/tom/.rvm/gems/ruby-2.6.3/bin/ruby_executable_hooks:24:in `<main>'
	4: from /Users/tom/.rvm/gems/ruby-2.6.3/bin/ruby_executable_hooks:24:in `eval'
	3: from /Users/tom/.rvm/gems/ruby-2.6.3/bin/bolt:23:in `<main>'
	2: from /Users/tom/.rvm/gems/ruby-2.6.3/bin/bolt:23:in `load'
	1: from /Users/tom/git/bolt/exe/bolt:11:in `<top (required)>'
/Users/tom/git/bolt/exe/bolt:11:in `exit': no implicit conversion from nil to integer (TypeError)
```

We'll likely want to have both `inventory_1_to_2` and `bolt_yaml_to_bolt_project` return a code indicating whether it was successful, and then use that to determine whether the migration overall was successful.

lib/bolt/project.rb
```diff
@@ -167,5 +167,131 @@ def check_deprecated_file
...
+          Bolt::Config::INVENTORY_OPTIONS.keys.include?(k)
+        end.map(&:to_h)
+
+        if transport_data.any?
+          outputter.print_message "Moving transportation configuration options "\
+            "'#{transport_data.keys}' from bolt.yaml to inventory.yaml"
+
+          if File.exist?(inventory_file)
+            inventory_data = Bolt::Util.read_yaml_hash(migrating_inv, 'inventory')
+            updated_inventory = transport_data.merge(inventory_data)
```
beechtom, line 225: This currently merges the transport config as top-level keys in the inventory file, instead of under the top-level `config` key.

lib/bolt/project.rb
```diff
@@ -167,5 +167,131 @@ def check_deprecated_file
...
+
+      else
+        config_data = Bolt::Util.read_optional_yaml_hash(config_file, 'config')
+        transport_data, project_data = config_data.partition do |k, _|
+          Bolt::Config::INVENTORY_OPTIONS.keys.include?(k)
+        end.map(&:to_h)
+
+        if transport_data.any?
+          outputter.print_message "Moving transportation configuration options "\
+            "'#{transport_data.keys}' from bolt.yaml to inventory.yaml"
```
beechtom, line 221:
```suggestion
            "'#{transport_data.keys.join("', '")}' from bolt.yaml to inventory.yaml"
```

This currently shows the keys in a quoted array, which looks a bit weird:

```
Moving transportation configuration options '["ssh"]' from bolt.yaml to inventory.yaml
```

lib/bolt/project.rb
```diff
@@ -167,5 +167,131 @@ def check_deprecated_file
...
+          "#{path}. Skipping project file update."
+
+      else
+        config_data = Bolt::Util.read_optional_yaml_hash(config_file, 'config')
+        transport_data, project_data = config_data.partition do |k, _|
+          Bolt::Config::INVENTORY_OPTIONS.keys.include?(k)
+        end.map(&:to_h)
+
+        if transport_data.any?
+          outputter.print_message "Moving transportation configuration options "\
```
beechtom, line 220: This message should probably come after the inventory file is backed up. Currently the order reads a bit weird since it sounds like the inventory is being modified and _then_ backed up.

lib/bolt/project.rb
```diff
@@ -167,5 +167,131 @@ def check_deprecated_file
...
+      end
+
+      date = Time.new.strftime("%Y%m%d_%H%M%S%L")
+      backup_dir = File.join(path, '.bolt-bak')
+      FileUtils.mkdir_p(backup_dir)
+
+      filename = File.basename(origin_path)
+      backup_path = File.join(backup_dir, "#{filename}.#{date}.bak")
+
+      outputter.print_message "Backing up #{filename} from #{origin_path} to #{backup_path}"
```
beechtom, line 190: Since this is used to backup the inventory in two places, you can get a message that the inventory was backed up twice. I think we should have a flag to tell if a file has already been backed up, or have some other way to not show this message multiple times.

lib/bolt/project.rb
```diff
@@ -167,5 +167,131 @@ def check_deprecated_file
...
+          outputter.print_message "Moving transportation configuration options "\
+            "'#{transport_data.keys}' from bolt.yaml to inventory.yaml"
+
+          if File.exist?(inventory_file)
+            inventory_data = Bolt::Util.read_yaml_hash(migrating_inv, 'inventory')
+            updated_inventory = transport_data.merge(inventory_data)
+            backup_file(migrating_inv, outputter)
+          else
+            FileUtils.touch(inventory_file) unless File.exist?(inventory_file)
+            updated_inventory = transport_data
```
beechtom, line 229: This writes an inventory with the transport keys as top-level keys.

lib/bolt/project.rb
```diff
@@ -167,5 +167,131 @@ def check_deprecated_file
...
       end
     end
+
+    def migrate_project(configured_inventory_path, outputter)
+      migrating_inv = configured_inventory_path || inventory_file
+      inventory_1_to_2(migrating_inv, outputter) if migrating_inv.file?
+      bolt_yaml_to_bolt_project(migrating_inv, outputter)
+    end
+
+    def backup_file(origin_path, outputter)
```
beechtom, line 177:
```suggestion
    private def backup_file(origin_path, outputter)
```

lucywyman: I was thinking this might be useful elsewhere, but for now it's probalby better to leave a comment that this doesn't strictly need to be private

lib/bolt/bolt_option_parser.rb
```diff
@@ -444,10 +444,7 @@ def get_help_text(subcommand, action = nil)
           bolt project migrate [options]
 
       DESCRIPTION
-          Migrate a Bolt project to the latest version.
-
-          Loads a Bolt project's inventory file and migrates it to the latest version. The
-          inventory file is modified in place and will not preserve comments or formatting.
+          Migrate a Bolt project to the latest versions and best practices.
```
beechtom, line 447: The phrasing here seems a bit off. @hestonhoffman your thoughts? Maybe something along the lines of this?

```
Migrate a Bolt project to use current best practices and the latest versions of configuration files.
```

hestonhoffman: Yeah, that sounds good

lib/bolt/project.rb
lucywyman: Hm, I wonder if giving more context in the message would be useful? Since it's backing up 2 different inventory files, it seems somewhat useful to tell the user we're backing up both times, and so users aren't suprised when there are 2 different invetory backups. Maybe an optional parameter to the method could be a reason for backing up, and the message would be `Backing up inventory.yaml from /home/user/project/inventory.yaml to /home/user/project/.bolt-bak before migrating inventory to version 2` and `Backing up inventory.yaml from /home/user/project/inventory.yaml to /home/user/project/.bolt-bak before moving transport config into inventory.yaml`

lib/bolt/project.rb
```diff
@@ -167,5 +168,132 @@ def check_deprecated_file
...
+      else
+        config_data = Bolt::Util.read_optional_yaml_hash(config_file, 'config')
+        transport_data, project_data = config_data.partition do |k, _|
+          Bolt::Config::INVENTORY_OPTIONS.keys.include?(k)
+        end.map(&:to_h)
+
+        if transport_data.any?
+          if File.exist?(inventory_file)
+            inventory_data = Bolt::Util.read_yaml_hash(migrating_inv, 'inventory')
+            updated_inventory = { 'config' => transport_data }.merge(inventory_data)
```
beechtom, line 224: We should deep merge `transport_data` with `inventory_data['config'] || {}` here, otherwise the config from the inventory will just override all of the config from `bolt.yaml`. For example, given these two files:

```yaml
# bolt.yaml
format: rainbow
ssh:
  user: bolt
  password: puppet
winrm:
  user: bolt
transport: docker
```
```yaml
# inventory.yaml
targets:
  - foo
  - bar
config:
  ssh:
    user: puppet
  transport: winrm
```

You will get the following inventory file:

```yaml
# inventory.yaml
---
config:
  ssh:
    user: puppet
  transport: winrm
targets:
- foo
- bar
```

Another benefit to merging the config sections separately is that the result can be assigned to `inventory_data['config']`, allowing us to preserve the order of the inventory file (at least those that weren't upgraded from v1).

lib/bolt/project.rb
beechtom: Since from the user's perspective the project is migrated in a single step, I'm not sure if there's a benefit to having backup files for each step of the migration process. I do think having those specific messages is an improvement, though.

I think ideally the `migrate_project` method would determine which files need to be backed up and then offload the work to actually migrating the files to the other methods... but that could also be difficult considering all the different conditions for determining whether a file needs to be migrated or not.

lib/bolt/project.rb
lucywyman: Oof, yeah, that would be bad :grimacing:

lib/bolt/project.rb
```diff
@@ -167,5 +168,133 @@ def check_deprecated_file
...
+      else
+        config_data = Bolt::Util.read_optional_yaml_hash(config_file, 'config')
+        transport_data, project_data = config_data.partition do |k, _|
+          Bolt::Config::INVENTORY_OPTIONS.keys.include?(k)
+        end.map(&:to_h)
+
+        if transport_data.any?
+          if File.exist?(inventory_file)
+            inventory_data = Bolt::Util.read_yaml_hash(migrating_inv, 'inventory')
+            merged = Bolt::Util.deep_merge(transport_data, inventory_data['config'])
```
beechtom, line 224:
```suggestion
            merged = Bolt::Util.deep_merge(transport_data, inventory_data['config'] || {})
```

lib/bolt/project.rb
```diff
@@ -167,5 +168,133 @@ def check_deprecated_file
...
+
+      date = Time.new.strftime("%Y%m%d_%H%M%S%L")
+      FileUtils.mkdir_p(backup_dir)
+
+      filename = File.basename(origin_path)
+      backup_path = File.join(backup_dir, "#{filename}.#{date}.bak")
+
+      outputter.print_message "Backing up #{filename} from #{origin_path} to #{backup_path}"
+
+      # No matter which error we rescue first, rubocop complains the order is wrong
```
nicklewis, line 194: That's because `Errno::EACCES` is redundant, since it _is_ a `StandardError`.

nicklewis:
I think the migration logic should live in its own class separate from the project. Otherwise, how will we handle migrating once we've dropped support for the version you're migrating from, if we have to be able to load the project as a `Bolt::Project` instance?

PR #2091, opened 2020-08-17
(PE-29929) Add /tasks endpoint to bolt-server

barriserloth:
This commit adds a new endpoint to list all tasks to bolt-server,
similar to the /plans endpoint that already exists. Tasks are
returned as an array where each element in the array is of the
form `{"name":"<TASK_NAME>"}`

### DIFF
--- a/lib/bolt_server/transport_app.rb
+++ b/lib/bolt_server/transport_app.rb
@@ -375,6 +375,22 @@ module BoltServer
       end
     end
 
+    # Fetches the list of tasks for an environment
+    #
+    # @param environment [String] the environment to fetch the list of tasks from
+    get '/tasks' do
+      in_pe_pal_env(params['environment']) do |pal|
+        tasks = pal.list_tasks
+        tasks_response = tasks.map { |task_name, _description| { 'name' => task_name } }.to_json
+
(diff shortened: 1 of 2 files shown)


lib/bolt_server/transport_app.rb
```diff
@@ -375,6 +375,27 @@ def make_winrm_target(target_hash)
...
+    # Fetches the list of tasks for an environment
+    #
+    # @param environment [String] the environment to fetch the list of tasks from
+    get '/tasks' do
+      in_pe_pal_env(params['environment']) do |pal|
+        tasks = pal.list_tasks
+        tasks_response = tasks.map { |task_name, _description|
+          {
+            'name' => task_name,
+            'environment' => [{ 'name' => params['environment'] }]
```
donoghuc, line 387: This just seems like it belongs in clojure code. I dont think addition of environment is necessary for bolt-server to send over the wire.

donoghuc:
Have a look at some of these tests https://github.com/puppetlabs/bolt/blob/295e5086262c51cdb0b0bbff28033e02becdca7c/spec/bolt_server/transport_app_spec.rb#L135-L180 Eventually we will want similar coverage for this endpoint (for now just focus on the case where we calculate the list, not individual task metadata).

Issue #2068, opened 2020-08-07
Bolt should always write a debug log


adreyer:
## Use Case
When a bolt action fails it may be hard to debug since you have to rerun the action, which may be dangerous or slow, to see the `--debug` output.

## Describe the Solution You Would Like
Unless disabled bolt should have a default logger that logs important debug information to a file at the top level of the project with every action.
This file should be overwritten with every action.
This logger must not log sensitive information.
The log format must be easily greppable.

As part of this work we should audit the debug message we output:
- Are they helpful? Level 1: "I want to know what Bolt is doing" Level 2: "I want to know how Bolt is doing what it's doing"
- Do they make sense?
- Does the message have sensitive information?
- Is the message logged at the right level?
- Ensure that for connection messages the target name is the name of the logger
- Is logging analytics calls helpful?

The logger should be disabled by setting the following configuration:
```
log:
  /path/to/logfile: 'disable'
```
The filename should be `bolt-debug.log`.
Duplicates https://github.com/puppetlabs/bolt/issues/2019

One outcome of this might be recommending new levels (i.e. trace, sensitive, etc.) for certain messages.

Issue #1995, opened 2020-07-15
Make it easier to discover where a project is loaded from


beechtom:
## Use Case
For users who are running projects outside of the project directory or have not created a project directory correctly (by not creating a `Boltdir`, `bolt.yaml`, or `bolt-project.yaml`), it can be difficult to know where a project is being loaded from. Bolt should make it easier for users to discover where a project is loaded from, and if possible, how it was loaded (i.e. from an environment variable, local project, embedded project, default, etc.)

While most users will likely run their projects from the project directory, or use the default project, there are users who do not do this or do not set up their project directories correctly. For new users this can be especially frustrating to troubleshoot, as Bolt does not provide much context on where a project is loaded from or why.

## Possible Solutions
There are several ways Bolt can provide additional context to users about where and how the project is being loaded.

- Log a debug message with the project's path when the project is loaded.

  ```
  Project loaded from '/Users/tom/bolt'
  ```

- Log a debug message with the project's type.

  ```
  Default project loaded from '/Users/tom/.puppetlabs/bolt'
  ```

- Log debug messages as part of `Bolt::Project#find_boltdir` that describe what the method is doing to find a project.

  ```
  Did not detect Boltdir, bolt.yaml, or bolt-project.yaml at '/Users/tom/foo/bar'; will not load as project
  Did not detect Boltdir, bolt.yaml, or bolt-project.yaml at '/Users/tom/foo'; will not load as project
  ...
  ```

- Add a CLI command that displays the path for the project and other loaded files

  ```
  $ bolt project show

  Default project loaded from '/Users/tom/.puppetlabs/bolt'

  config        /Users/tom/.puppetlabs/bolt/bolt-project.yaml
                /etc/puppetlabs/bolt/bolt.yaml
  inventory     /Users/tom/.puppetlabs/bolt/inventory.yaml
  ...
  ```

## Solution
Now that our logging levels are sorted we should log 'Project loaded from '/Users/tom/bolt'' at info level, and log all the paths we search (option 3 above) at debug level.

PR #2093, opened 2020-08-18
(GH-2068) Write debug log by default

nicklewis:
Bolt now writes a `bolt-debug.log` file in the project directory that includes
debug-level information from the _most recent_ Bolt run. The log is overwritten
each time Bolt runs.

This also reorganizes the log levels a bit to suit writing a debug log without
including sensitive information. We now have a trace level beyond debug which
is used to log messages that are only really interesting when trying to debug
Bolt itself. Debug is now used for messages that help the user debug their
usage of Bolt or the execution of their plan. Info is used to report _what_
Bolt is doing.

To suit this debug log, we now downgrade Puppet log messages rather than
mapping them directly to the corresponding Bolt log level. Puppet info messages
go to Bolt debug level, Puppet debug messages go to Bolt trace level. As Puppet
is primarily an implementation detail of Bolt, this helps avoid noise in the
log. Specifically, Puppet is chatty at debug level and will overwhelm the debug
log without this change.

!feature

  * **Write a default log file**
    ([#2068](https://github.com/puppetlabs/bolt/issues/2068))

    Bolt will now log activity at debug level to `bolt-debug.log` in the
   project directory. This log will be truncated each time Bolt runs.

### DIFF
--- a/lib/bolt/applicator.rb
+++ b/lib/bolt/applicator.rb
@@ -300,7 +300,7 @@ module Bolt
 
         files.each do |file|
           tar_path = Pathname.new(file).relative_path_from(parent)
-          @logger.debug("Packing plugin #{file} to #{tar_path}")
+          @logger.trace("Packing plugin #{file} to #{tar_path}")
           stat = File.stat(file)
           content = File.binread(file)
           output.tar.add_file_simple(
@@ -314,7 +314,7 @@ module Bolt
       end
 
       duration = Time.now - start_time
-      @logger.debug("Packed plugins in #{duration * 1000} ms")
+      @logger.trace("Packed plugins in #{duration * 1000} ms")
 
       output.close
       Base64.encode64(sio.string)
--- a/lib/bolt/applicator.rb
+++ b/lib/bolt/applicator.rb
@@ -27,7 +27,7 @@ module Bolt
       @hiera_config = hiera_config ? validate_hiera_config(hiera_config) : nil
       @apply_settings = apply_settings || {}
 
-      @pool = Concurrent::ThreadPoolExecutor.new(max_threads: max_compiles)
+      @pool = Concurrent::ThreadPoolExecutor.new(name: 'apply', max_threads: max_compiles)
       @logger = Logging.logger[self]
     end
 
@@ -217,6 +217,7 @@ module Bolt
       r = @executor.log_action(description, targets) do
         futures = targets.map do |target|
           Concurrent::Future.execute(executor: @pool) do
+            Thread.current[:name] ||= Thread.current.name
             @executor.with_node_logging("Compiling manifest block", [target]) do
               compile(target, scope)
             end
--- a/lib/bolt/bolt_option_parser.rb
+++ b/lib/bolt/bolt_option_parser.rb
@@ -863,7 +863,7 @@ module Bolt
       end
       define('--log-level LEVEL',
              "Set the log level for the console. Available options are",
-             "debug, info, notice, warn, error, fatal, any.") do |level|
+             "trace, debug, info, warn, error, fatal, any.") do |level|
         @options[:log] = { 'console' => { 'level' => level } }
       end
       define('--plugin PLUGIN', 'Select the plugin to use') do |plug|
--- a/lib/bolt/cli.rb
+++ b/lib/bolt/cli.rb
@@ -1056,7 +1056,7 @@ module Bolt
       msg = <<~MSG.chomp
         Loaded configuration from: '#{config.config_files.join("', '")}'
       MSG
-      @logger.debug(msg)
+      @logger.info(msg)
     end
 
     # Gem installs include the aggregate, canary, and puppetdb_fact modules, while
--- a/lib/bolt/logger.rb
+++ b/lib/bolt/logger.rb
@@ -80,7 +80,7 @@ module Bolt
 
     def self.default_layout
       Logging.layouts.pattern(
-        pattern: '%d %-6l %c: %m\n',
+        pattern: '%d %-6l [%T] [%c] %m\n',
         date_pattern: '%Y-%m-%dT%H:%M:%S.%6N'
       )
     end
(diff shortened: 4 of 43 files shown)


lib/bolt/logger.rb
```diff
@@ -81,7 +80,7 @@ def self.console_layout(color)
 
     def self.default_layout
       Logging.layouts.pattern(
-        pattern: '%d %-6l %c: %m\n',
+        pattern: '%d %-6l [%T] [%c] %m\n',
```
beechtom, line 83: I like this change a lot! It looks like the applicator still needs to name the current thread, though:

```
2020-08-19T09:02:50.431928 INFO   [] [Bolt::Executor] Compiling manifest block on ["server"]
```

nicklewis: Updated, and I set names for the TPEs so the threads will show up as `exec-worker-1` and `apply-worker-1`.

beechtom:
I think we might want to lower the log level when logging a result to debug: https://github.com/puppetlabs/bolt/blob/main/lib/bolt/executor.rb#L250

Currently, running with `--log-level info` will show duplicate results, which seems a bit off and also doesn't fit into info showing _what_ Bolt is doing.

nicklewis:
I was torn on that one. I'm not super concerned about duplicate messages on the console as I would like to stop log messages from appearing there anyway. But I think logging those messages at info level mirrors the behavior of the outputter: if you run a task with bolt, bolt will tell you the result of the task. It's a bit more subtle because that's _not_ the default behavior for the outputter when running a plan, but it is when running anything else, whereas the log behavior is consistent between plans and other actions.

lib/bolt/bolt_option_parser.rb
```diff
@@ -863,7 +863,7 @@ def initialize(options)
       end
       define('--log-level LEVEL',
              "Set the log level for the console. Available options are",
-             "debug, info, notice, warn, error, fatal, any.") do |level|
+             "trace, debug, info, warn, error, fatal, any.") do |level|
```
lucywyman, line 866: It seems like we should leave `notice` here since it can be configured in Bolt config files, and that's what sets the console config for messages from Puppet. For example, this will break plans that call `notice()` in their apply blocks.

lucywyman: Or, alternatively, we should remove notice from both the CLI and config options. https://github.com/puppetlabs/bolt/blob/main/lib/bolt/config/options.rb#L189

nicklewis: This was meant as a sort of soft deprecation of `notice`. No messages should be going to that level now, but if you've specified it as the level for a log in your config, we don't want that to break. But we don't want anybody to actually _set_ their log level to `notice` now, so I removed it from the documentation.

I will update the config in `options.rb` to remove `notice` and also add `trace`. I wasn't sure if that code was used for validation so I left `notice` for compatibility. But now I see it's just for docs/schema generation, so it can be safely updated.

lib/bolt/applicator.rb
```diff
@@ -217,6 +217,7 @@ def apply_ast(raw_ast, targets, options, plan_vars = {})
       r = @executor.log_action(description, targets) do
         futures = targets.map do |target|
           Concurrent::Future.execute(executor: @pool) do
+            Thread.current[:name] ||= Thread.current.name
```
lucywyman, line 220: What is this doing? I'm assuming the `:name` attribute is the same as the name returned from the name getter, though the docs don't explicitly say so.

lib/bolt/bolt_option_parser.rb
lucywyman: Should we have a deprecation warning if you try to pass `notice`? It may be annoying to silently have it go from working to not working, even on the CLI.

lib/bolt/cli.rb
```diff
@@ -1056,7 +1056,7 @@ def config_loaded
       msg = <<~MSG.chomp
         Loaded configuration from: '#{config.config_files.join("', '")}'
       MSG
-      @logger.debug(msg)
+      @logger.info(msg)
```
lucywyman, line 1059: @beechtom Does this at least partially solve https://github.com/puppetlabs/bolt/issues/1995 ?

lib/bolt/applicator.rb
nicklewis: `:name` is just a thread local variable, separate from actual thread "name". concurrent-ruby sets `thread.name` when it makes the worker threads, but the logging library looks up `thread[:name]` when you use `%T` in the log formatter. So this is copying the name set by concurrent-ruby to the location used by the logging library.

lib/bolt/bolt_option_parser.rb
nicklewis: Just did some more digging: It appears that `Thread#name` was added in Ruby 2.3 which is why `logging`, which _well_ predates that Ruby version, uses a `:name` variable instead. We might be able to get them to accept a PR to check the `name` attr.

I was on the fence about warning for `notice` at this point. We should definitely warn about it before we remove it, but I wanted to let it sit a bit and make sure the new levels actually work for people before committing to removing it for real.

lib/bolt/cli.rb
lucywyman: Hm, I actually wonder if this belongs at the same level as `Loading modules from`. It seems like you'd only want to know this if Bolt wasn't working as expected, which seems like debug level? But it also seems like more of a pain point than users not having their modulepath set correctly, which maybe warrants a higher level. Either way it's not logged by default - I don't feel super strongly.

nicklewis: I think ideally we would have info level logging about which project is being used so that if you're using a log file shared between projects, you can tell which project was being used. Then details about which config files, modulepath, inventory, etc could live at debug level. Since we don't have anything that is explicitly about the project, mentioning the config file at info level is a half measure.

lib/bolt/applicator.rb
lucywyman: Ha! TIL. Thanks for the explanation!

lib/bolt/cli.rb
beechtom: @lucywyman Pretty much what Nick said, but I do think logging this at info combined with it being written to the debug log by default is an improvement, since it will hopefully be a bit more discoverable. 🤞

PR #2097, opened 2020-08-19
(docs) Update run_* function calls in 'Writing plans' docs

beechtom:
This updates several `run_*` function calls in the example plans of the
'Writing plans' doc to use a consistent format for specifying plan,
task, and parameter names. Previously, the doc was inconsistent in how
it showed specifying plan, task, and parameter names, switching between
quoted and unquoted names.

This also fixes a `run_plan` function call that was specifying the
`**puppetdb_fact**` plan, instead of `puppetdb_fact`.

!no-release-note

### DIFF
--- a/documentation/writing_plans.md
+++ b/documentation/writing_plans.md
@@ -98,7 +98,7 @@ plan test::parameter_passing (
   TargetSpec $targets,
   Optional[String[1]] $example_nul = undef,
 ) {
-  return run_task('test::demo_undef_bash', $targets, example_nul => $example_nul)
+  return run_task('test::demo_undef_bash', $targets, 'example_nul' => $example_nul)
 }
 ```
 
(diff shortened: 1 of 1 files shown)


documentation/writing_plans.md
```diff
@@ -692,12 +692,12 @@ to run on the targets. You must configure the PuppetDB client before you run it.
 ```
 plan run_with_facts(TargetSpec $targets) {
   # This collects facts on targets and update the inventory
-  run_plan(**puppetdb_fact**, targets => $targets)
```
donoghuc, line 695: Wouldnt this be the only thing that errored?

donoghuc: The single quotes all over dont seem necessary?

beechtom: Huh, TIL you don't need to quote task/plan names and parameters.
I do think having the consistency is preferable, though. I'll update the commit message.

donoghuc: Consistency is good. Names like `'my_module::my_task'` require quotes.

PR #2100, opened 2020-08-19
(GH-2039) Add plan and param descriptions to converted plans

### LINKED ISSUE #2039: Add plan and parameter descriptions to plan documentation when converting YAML plans
## Use Case

YAML plans accept a top-level `description` key and parameter-level `description` keys that are used by `bolt plan show` to show plan documentation. Puppet plans use puppet-strings to pull similar documentation from comments at the top of the plan file. The `bolt plan convert` command should add plan and parameter descriptions from a YAML plan to the comments in the converted Puppet plan to preserve documentation.

To test: `bolt plan show` is the same for the YAML plan as the converted puppet plan.

lucywyman:
This modifies the plan transpiler to include plan description and plan
parameter descriptions in the converted plan output when converting from
a YAML plan to a Puppet plan. This update is simple as plan and
parameters descriptions are already available from the YamlPlan object,
so just adds them in the proper puppet-strings format to the Puppet
plan. The converted plan warning is part of the plan description by
default.

Closes #2039

!feature

* **Plan conversion maintains plan and parameter descriptions** ([2039](https://github.com/puppetlabs/bolt/issues/2039))

  Converting a YAML plan to a Puppet plan will now preserve plan and
  parameter descriptions, so that `plan show` output is the same for the
  YAML plan as the converted Puppet plan.


lib/bolt/pal/yaml_plan/transpiler.rb
```diff
@@ -21,10 +21,18 @@ def transpile(relative_path)
           validate_path
 
           plan_object = parse_plan
+          param_descriptions = plan_object.parameters.each_with_object([]) do |param, acc|
```
nicklewis, line 24: This can just be `map` rather than `each_with_object` with an accumulator, since the results are being joined at the end anyhow.

lucywyman: Updated

Issue #2082, opened 2020-08-12
Add `bolt modules install`


lucywyman:
## Use Case
- [x] Users should be able to configure a list of modules their project depends on under a `modules` key in `bolt-project.yaml` (and not `bolt-default.yaml`), and install those modules with `bolt module[s]/project install`.
  - If no Puppetfile is present the command will generate a Puppetfile based on modules defined in bolt-project.yaml then install that puppetfile.
  - If a Puppetfile is present and it contains all the modules in modules.yaml bolt will install that Puppetfile.
  - If a Puppetfile is present and it is missing declared modules, or it can not be simply parsed the user will be instructed to rerun after deleting or moving it or with the `--no-resolve/--force` flag. The dependencies will not be validated every run since the puppetfile is expected to pin each module explicitly.
  - Modules should be installed to `.modules/`. `.modules/` shouldn't be added to the modulepath, though users can add it manually to their modulepath.
  - Users should be able to copy-paste module definitions from module dependency definitions in metadata.json into this file and Puppetfile will be considered a machine-only file for most users.

To ticket:
- [x] `bolt puppetfile show-modules` -> `bolt module[s]/project show` . Deprecation can happen later.
- [x] `bolt generate-types` -> `bolt module[s]/project generate-types`
- [x] Add PowerShell cmdlet `Install-BoltModule` / `Install-BoltProject`
- [ ] Update docs. This should happen last so docs don't go live for a feature that is flagged.
 
## Questions:
- [x] `bolt module install` or `bolt modules install` or `bolt project install`? Do a user test
- `--no-resolve` or `--force`? **--force makes more sense**

beechtom:
Closed by #2106

PR #2106, opened 2020-08-21
(GH-2082) Add 'module install' command to install module dependencies

beechtom:
This adds a new Bolt command `bolt module install` that installs a list of
modules declared in the project's configuration file, as well as all of
their dependencies. This command will resolve all dependencies for the
declared modules, generate a Puppetfile with specifications for each
module, and then install the modules to the project's `.modules`
directory.

As part of this work, the project configuration file,
`bolt-project.yaml`, now accepts a new `modules` key. This key accepts
an array of module declarations, where each declaration is a map that
must include the module's name under the `name` key. The `name` key is
the only supported key at this time.

The installed modules can be found in a new `.modules` directory within
the project. This directory is not currently added to the modulepath,
though users can configure their modulepath to include it if they want.

Depending on the modules declared in the project configuration file, and
whether the project has an existing Puppetfile, this command will behave
differently:

- **No module declarations**

  If the project configuration file `bolt-project.yaml` does not include
  a `modules` key, the command will do nothing.

- **No existing Puppetfile**

  If there are module declarations, but no existing Puppetfile, Bolt
  will resolve the module declarations, generate a new Puppetfile, and
  install the modules.

- **Existing Puppetfile with matching module declarations**

  If there are module declarations, and an existing Puppetfile has
  specifications for each of the module declarations, then Bolt will not
  resolve the module declarations and instead just reinstall the
  Puppetfile.

- **Existing Puppetfile with missing module declarations**

  If there are module declarations, and an existing Puppetfile does not
  have specifications for each of the module declarations, then Bolt
  will raise an error indicating that the Puppetfile may not be managed
  by Bolt.

  If a user wishes to override this behavior, they can run the command
  with the `--force` option, which will force Bolt to resolve the
  modules and overwrite the Puppetfile with a new Puppetfile:

  ```shell
  $ bolt module install --force
  ```

This work introduces a few new classes:

- `Bolt::Puppetfile`
   This manages the logical contents of a Puppetfile and includes
   methods for parsing a Puppetfile, resolving module dependencies,
   and writing a Puppetfile.

- `Bolt::Puppetfile::Installer`
   This installs modules from a Puppetfile.

- `Bolt::Puppetfile::Module`
   This represents a module specification. It includes a comparator
   and some convenience methods.

Lastly, this refactors the `Bolt::CLI#install_puppetfile` and
`Bolt::CLI#initialize_project` methods to use these new classes.

---

**Testing this feature**

This feature is feature flagged. To enable the `bolt module install`
command, set the environment variable `BOLT_MODULE_FEATURE`.

!no-release-note

### DIFF
--- a/lib/bolt/bolt_option_parser.rb
+++ b/lib/bolt/bolt_option_parser.rb
@@ -64,6 +64,15 @@ module Bolt
       when 'guide'
         { flags: OPTIONS[:global] + %w[format],
           banner: GUIDE_HELP }
+      when 'module'
+        case action
+        when 'install'
+          { flags: OPTIONS[:global] + %w[configfile force project],
+            banner: MODULE_INSTALL_HELP }
+        else
+          { flags: OPTIONS[:global],
+            banner: MODULE_HELP }
+        end
       when 'plan'
         case action
         when 'convert'
@@ -339,6 +348,35 @@ module Bolt
 
       DESCRIPTION
           Show the list of targets an action would run on.
+    HELP
+
+    MODULE_HELP = <<~HELP
+      NAME
+          module
+      
+      USAGE
+          bolt module <action> [options]
+
+      DESCRIPTION
+          Install the project's modules
+
+      ACTIONS
+          install       Install the project's modules
+    HELP
+
+    MODULE_INSTALL_HELP = <<~HELP
+      NAME
+          install
+      
+      USAGE
+          bolt module install [options]
+
+      DESCRIPTION
+          Install the project's modules.
+
+          Module declarations are loaded from the project's configuration
+          file. Bolt will automatically resolve all module dependencies,
+          generate a Puppetfile, and install the modules.
     HELP
 
     PLAN_HELP = <<~HELP
@@ -864,9 +902,9 @@ module Bolt
       define('--modules MODULES',
              'A comma-separated list of modules to install from the Puppet Forge',
              'when initializing a project. Resolves and installs all dependencies.') do |modules|
-        @options[:modules] = modules.split(',')
+        @options[:modules] = modules.split(',').map { |mod| { 'name' => mod } }
       end
-      define('--force', 'Overwrite existing key pairs') do |_force|
+      define('--force', 'Force a destructive action') do |_force|
         @options[:force] = true
       end
 
--- a/lib/bolt/cli.rb
+++ b/lib/bolt/cli.rb
@@ -40,7 +40,8 @@ module Bolt
                  'group' => %w[show],
                  'project' => %w[init migrate],
                  'apply' => %w[],
-                 'guide' => %w[] }.freeze
+                 'guide' => %w[],
+                 'module' => %w[install] }.freeze
 
     attr_reader :config, :options
 
@@ -210,10 +211,14 @@ module Bolt
     end
 
     def validate(options)
-      unless COMMANDS.include?(options[:subcommand])
+      # Disables the 'module' subcommand unless the module feature flag is set.
+      commands = COMMANDS.dup
+      commands.delete('module') unless ENV['BOLT_MODULE_FEATURE']
+
+      unless commands.include?(options[:subcommand])
         raise Bolt::CLIError,
               "Expected subcommand '#{options[:subcommand]}' to be one of " \
-              "#{COMMANDS.keys.join(', ')}"
+              "#{commands.keys.join(', ')}"
       end
 
       actions = COMMANDS[options[:subcommand]]
@@ -353,7 +358,7 @@ module Bolt
       # Initialize inventory and targets. Errors here are better to catch early.
       # options[:target_args] will contain a string/array version of the targetting options this is passed to plans
       # options[:targets] will contain a resolved set of Target objects
-      unless %w[project puppetfile secret guide].include?(options[:subcommand]) ||
+      unless %w[guide module project puppetfile secret].include?(options[:subcommand]) ||
              %w[convert new show].include?(options[:action])
         update_targets(options)
       end
@@ -443,12 +448,17 @@ module Bolt
         when 'run'
           code = run_plan(options[:object], options[:task_options], options[:target_args], options)
         end
+      when 'module'
+        case options[:action]
+        when 'install'
+          code = install_project_modules
+        end
       when 'puppetfile'
         case options[:action]
         when 'generate-types'
           code = generate_types
         when 'install'
-          code = install_puppetfile(config.puppetfile_config, config.puppetfile, config.modulepath)
+          code = install_puppetfile(config.puppetfile_config, config.puppetfile, config.modulepath.first)
         end
       when 'secret'
         code = Bolt::Secret.execute(plugins, outputter, options)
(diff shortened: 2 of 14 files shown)


lib/bolt/bolt_option_parser.rb
```diff
@@ -64,6 +64,15 @@ def get_help_text(subcommand, action = nil)
       when 'guide'
         { flags: OPTIONS[:global] + %w[format],
           banner: GUIDE_HELP }
+      when 'module'
+        case action
+        when 'install'
```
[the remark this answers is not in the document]
lucywyman: `bolt project install-modules`? I'm tempted to ask in community slack but it may be hard to articulate all the nuances.

lib/bolt/puppetfile.rb
```diff
@@ -0,0 +1,231 @@
...
+# dependencies, and installing modules from a Puppetfile.
+#
+module Bolt
+  class Puppetfile
+    attr_reader :modulepath, :modules, :path
+
+    def initialize(path:, config:, modulepath:, modules: [])
+      @path       = path
+      @config     = config
+      @modulepath = modulepath
```
nicklewis, line 19: It looks like `modulepath` can just be an argument to `install`? In that case, it ought to be a single directory and not a modulepath, since we only care about a single directory.

lib/bolt/cli.rb
```diff
@@ -840,109 +852,92 @@ def initialize_project
...
+        outputter.print_message "Parsing existing Puppetfile at #{puppetfile.path}"
+        existing_modules = puppetfile.parse
+
+        unless existing_modules.superset?(puppetfile.modules)
+          missing_modules = puppetfile.modules - existing_modules
+
+          raise Bolt::Error.new(
+            "Puppetfile #{puppetfile.path} is missing specifications for modules: "\
+            "#{missing_modules.map(&:title).join(', ')}. This may not be a Puppetfile "\
+            "managed by Bolt. To foricbly overwrite the Puppetfile, run with the "\
```
nicklewis, line 901:
```suggestion
            "managed by Bolt. To forcibly overwrite the Puppetfile, run with the "\
```

lib/bolt/puppetfile.rb
```diff
@@ -0,0 +1,231 @@
...
+          "Cannot overwrite existing Puppetfile at #{@path}. To forcibly overwrite, "\
+          "run with the '--force' option.",
+          @path
+        )
+      end
+
+      content  = "# This Puppetfile is managed by Bolt. Do not edit.\n"
+      content += "#{@modules.map(&:to_spec).join("\n")}\n"
+
+      File.write(@path, content)
```
nicklewis, line 76:
```rb
File.open(@path, 'w') do |file|
  file.puts "# This Puppetfile is managed by Bolt. Do not edit."
  modules.each do |mod|
    file.puts mod.to_spec
  end
  file.puts
end
```

lucywyman:
Should we add docs about this to https://puppet.com/docs/bolt/latest/experimental_features.html ?

lib/bolt/bolt_option_parser.rb
```diff
@@ -64,6 +64,15 @@ def get_help_text(subcommand, action = nil)
       when 'guide'
         { flags: OPTIONS[:global] + %w[format],
           banner: GUIDE_HELP }
+      when 'module'
```
lucywyman, line 67: Was this intentionally left out of the help text on line 180 of this file?

beechtom: Yes. Otherwise, users would see it as an available command in the help text even though it's currently gated.

lib/bolt/puppetfile/module.rb
```diff
@@ -0,0 +1,95 @@
...
+      #
+      def title
+        "#{@owner}-#{@name}"
+      end
+
+      # Checks two modules for equality.
+      #
+      def eql?(other)
+        return false unless other.is_a?(Bolt::Puppetfile::Module)
+        @owner == other.owner && @name == other.name
```
nicklewis, line 55:
```suggestion
    self.class == other.class &&
      @owner == other.owner &&
      @name == other.name
```

lib/bolt/puppetfile/installer.rb
```diff
@@ -0,0 +1,63 @@
...
+  class Puppetfile
+    class Installer
+      def initialize(path, moduledir, config = {})
+        @path      = path
+        @moduledir = moduledir
+        @config    = config
+        validate
+      end
+
+      def install
```
nicklewis, line 19: The more I think about it, the more I think it makes sense for `path` and `moduledir` to be arguments to `install` instead of instance variables. If you consider the lifecycle of this object, it's possible it would be constructed early on during configuration, before we know where/what we want to install with it.

lib/bolt/puppetfile/module.rb
```diff
@@ -0,0 +1,95 @@
...
+          raise Bolt::ValidationError, "Module name must be a String."
+        end
+
+        owner, name = mod['name'].split(%r{[/-]}, 2)
+
+        unless owner && name
+          raise Bolt::ValidationError, "Module name #{mod['name']} must include both the owner and module name."
+        end
+
+        new(owner, name)
```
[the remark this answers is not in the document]
beechtom: No, puppetfile-resolver doesn't support version ranges, so we can't support the case of copy/pasting dependencies from metadata.json yet. It _does_ support explicit versions, as well as version shorthand (1.x, etc.), though.

lib/bolt/project.rb
```diff
@@ -159,6 +163,17 @@ def validate
           raise Bolt::ValidationError, "'#{conf}' in bolt-project.yaml must be an array"
         end
       end
+
+      if @data['modules']
+        unless @data['modules'].is_a?(Array)
+          raise Bolt::ValidationError, "modules in bolt-project.yaml must be an array"
```
lucywyman, line 169:
```suggestion
          raise Bolt::ValidationError, "'modules' in bolt-project.yaml must be an array"
```
I'm not sure how to convey the distinction between the modules key and the module declarations themselves.

lib/bolt/puppetfile.rb
```diff
@@ -0,0 +1,169 @@
+# frozen_string_literal: true
+
+require 'pathname'
```
lucywyman, line 3: I don't think this is needed.

lib/bolt/puppetfile.rb
```diff
@@ -0,0 +1,169 @@
...
+
+require 'bolt/error'
+require 'bolt/puppetfile/module'
+
+# This class manages the logical contents of a Puppetfile. It includes methods
+# for parsing a Puppetfile and its modules, resolving module dependencies,
+# and writing a Puppetfile.
+#
```
lucywyman, line 11:
```suggestion
```

lib/bolt/puppetfile.rb
```diff
@@ -0,0 +1,169 @@
...
+    attr_reader :modules
+
+    def initialize(modules = [])
+      @modules = Set.new
+      add_modules(modules)
+    end
+
+    # Loads a Puppetfile and parses its module specifications, returning a
+    # Bolt::Puppetfile object with the modules set.
+    #
```
lucywyman, line 23:
```suggestion
```

lib/bolt/puppetfile.rb
```diff
@@ -0,0 +1,169 @@
...
+    def self.parse(path)
+      require 'puppetfile-resolver'
+      require 'puppetfile-resolver/puppetfile/parser/r10k_eval'
+
+      unless File.exist?(path)
```
lucywyman, line 28: This is currently dead code - are we sure we're eventually going to need this?

lib/bolt/puppetfile/module.rb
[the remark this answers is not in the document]
lucywyman: In that case can we warn when there's an invalid version specified, or other invalid keys? It's most annoying to have `version_requirements` defined and silently ignored.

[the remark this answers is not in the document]
beechtom: We won't support versioning at all until we can (hopefully) get a change into puppetfile-resolver to support version ranges. Updated to add a warning if any module declarations have an unsupported key.

lib/bolt/puppetfile.rb
```diff
@@ -0,0 +1,160 @@
...
+    def resolve
+      require 'puppetfile-resolver'
+
+      # Build the document model from the modules.
+      model = PuppetfileResolver::Puppetfile::Document.new('')
+
+      @modules.each do |mod|
+        model.add_module(
+          PuppetfileResolver::Puppetfile::ForgeModule.new(mod.title).tap do |tap|
+            tap.version = mod.version || :latest
```
lucywyman, line 85: This can't currently be set at all, we should just use :latest to avoid confusion.

lib/bolt/cli.rb
```diff
@@ -837,109 +849,82 @@ def initialize_project
...
+
+      # If the Puppetfile exists, check if it includes specs for each declared
+      # module, erroring if there are any missing. Otherwise, resolve the
+      # module dependencies and write a new Puppetfile. Users can forcibly
+      # overwrite an existing Puppetfile with the '--force' option.
+      if puppetfile_path.exist? && !options[:force]
+        outputter.print_message "Parsing existing Puppetfile at #{puppetfile_path}"
+        existing = Bolt::Puppetfile.parse(puppetfile_path)
+
+        unless existing.modules.superset? puppetfile.modules
```
lucywyman, line 888: This is probably good enough for now, but doesn't account at all for version discrepancies. I think we'll need to think through the upgrading process once users can specify versions for modules - if the module is there but is an older version than specified, do we error or upgrade?

lucywyman:
I'm honestly impressed with how well this dealt with a version conflict:
```
lucy ~/githubs/bolt ((8de4bba...)) $ be bolt module install --project project/
Resolving module dependencies, this may take a moment
Puppetfile Resolver could not find compatible versions for possibility named "stdlib":
  In Puppetfile:
    spy-closure_compiler >= 0 was resolved to Forge spy-closure_compiler-1.0.0, which depends on
      puppetlabs-stdlib >= 5.0.0 < 9.0.0

    gbloquel-postgres_exporter >= 0 was resolved to Forge gbloquel-postgres_exporter-0.1.1, which depends on
      puppetlabs-stdlib >= 4.13.1 < 5.0.0
```

lib/bolt/cli.rb
beechtom: Ya, without versioning this is all the validation that is needed, but we'll eventually need a more robust method for checking that the module declarations match what's in the Puppetfile. That should be fairly simple to implement once it's needed.

lib/bolt/cli.rb
```diff
@@ -837,109 +849,82 @@ def initialize_project
...
+        outputter.print_message "Writing Puppetfile at #{puppetfile_path}"
+        puppetfile.write(puppetfile_path, force: true)
       end
 
-      # Map specification models to a Puppetfile specification
-      spec_graph.values.map do |spec|
-        "mod '#{spec.owner}-#{spec.name}', '#{spec.version}'"
-      end
-    end
+      outputter.print_message "Syncing modules from #{puppetfile_path} to #{moduledir}"
```
lucywyman, line 906: This is a little misleading, since if the list of modules in the project matches the list in Puppetfile it will do nothing (including not updating outdated module versions). I think we should only print this if we're actually installing or updating anything, and can probably skip the install step to save some cycles if the installed modules match the project list.

lucywyman: Oo, the plot thickens! This *does* update versions if `--force` is provided, even if the sets are equal.

Issue #2086, opened 2020-08-13
Mark task output as sensitive to avoid exposing it


nicklewis:
## Use Case

If the output of a task is sensitive (for instance, if the task retrieves a password or other secret from a remote system) and the task is run as a utility from a plan, there's no way for the user to prevent the output from being exposed. Bolt will print it to the console with `--verbose` and to the log by default. The former behavior can be prevented using `without_default_logging`, but not the latter. The Puppet Enterprise console will also allow the output to be viewed as part of the step which ran it.

Wrapping the output in `Sensitive(...)` in the plan doesn't prevent these issues, as the result will be printed/logged/stored _before_ it's marked sensitive. Therefore, the only way to prevent this behavior is for Bolt to either _never_ print task output or to know when output is sensitive.

## Describe the Solution You Would Like

- Task authors can include `_sensitive` at the same level as `_output` and `_error`, and that value will get wrapped as `Sensitive`. We should ensure that hashes can be wrapped in Sensitive, and only document making hashes sensitive.
- We should add a `sensitive` method to the Result object that returns the wrapped value.

## Describe Alternatives You've Considered

We could add a `_sensitive => true` option to the `run_task()` function to tell Bolt that _this_ invocation of the task should be considered sensitive. That would generalize easily to `run_command()` and `run_script()`, but it shifts some burden to the user of the task. Given the output being sensitive is likely an inherent property of the task, and not specific to the context in which it's being run, it seems more correct to track it as an attribute of the task directly, similar to sensitive task parameters.

Alternately, we could simply instruct users that tasks should not return sensitive output.

PR #2107, opened 2020-08-21
(GH-2086) Treat the _sensitive key as sensitive

nicklewis:
When constructing a Bolt::Result from the output of a task, we now look
for a _sensitive key and wrap it in Puppet's Sensitive type if found.
This prevents the data from being printed on the CLI.

!feature

  * **Allow task output to be treated as sensitive**
     ([#2086](https://github.com/puppetlabs/bolt/issues/2086))

    Tasks can now return a _sensitive key in their output which can
    contain an arbitrary value which will be treated as sensitive by
    Bolt. This means that it won't be printed to the console or log at
    any log level, and plans will need to use unwrap() to get the value.

### DIFF
--- a/documentation/writing_tasks.md
+++ b/documentation/writing_tasks.md
@@ -421,6 +421,26 @@ end
 puts result.to_json
 ```
 
+### Returning sensitive data
+
+To return secrets from a task, use the `_sensitive` key in the output. Bolt
+will treat the result as sensitive and won't allow it to be printed to the
+console or log.
+
+```ruby
+#!/opt/puppetlabs/puppet/bin/ruby
+
+require 'json'
+
+user_name = 'someone'
+# Generate a 10 letter password
+user_password = [*'a'..'z'].sample(10).join
+
+result = { user: user_name, _sensitive: { password: user_password } }
+
+puts result.to_json
+```
+
 ## Supporting no-op in tasks
 
 Tasks support no-operation functionality, also known as no-op mode. This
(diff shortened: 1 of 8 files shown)


documentation/writing_tasks.md
```diff
@@ -421,6 +421,26 @@ end
...
+```ruby
+#!/opt/puppetlabs/puppet/bin/ruby
+
+require 'json'
+
+user_name = 'someone'
+# Generate a 10 letter password
+user_password = [*'a'..'z'].sample(10).join
+
+result = { user: user_name, _sensitive: { password: user_password } }
```
lucywyman, line 439: Should we also document a plan that unwraps and uses a sensitive value?

spec/fixtures/apply/puppet_types/plans/sensitive_result.pp
```diff
@@ -0,0 +1,14 @@
...
+  TargetSpec $targets
+) {
+  $target = get_target($targets)
+  $target.apply_prep
+
+  $result = run_task('sensitive', $target).first
+
+  return apply($target) {
+    notify { "Result sensitive value: ${$result.sensitive.unwrap['password']}": }
+    notify { "Result target name: ${$result.target.name}": }
```
lucywyman, line 11: This notify doesn't get used.

documentation/writing_tasks.md
nicklewis: We do have a section on "Sensitive parameters" which shows unwrapping a `Sensitive` instance in another context and there's a note on the `sensitive()` function of the `Result` datatype that says to use `unwrap()`. Pulling sensitive data off a `Result` seems to be at a finer level of detail than the rest of our writing plans documentation.

lucywyman:
Do we need to update the puppet-specifications section to include this? https://github.com/puppetlabs/puppet-specifications/tree/master/tasks#output-handling

nicklewis:
Oops I made the change and forgot to commit it and open a PR! https://github.com/puppetlabs/puppet-specifications/pull/149

PR #2118, opened 2020-08-25
(GH-2118) Fix Invoke-BoltTask -params

jpogran:
The `-params` parameter was originally typed an `[object]` to allow both string `action=status name=spooler` and hashtable values `@{ action = 'status'; name = 'spooler' }`. This succeeded with hashtables but failed with strings because PowerShell parses `action=status name=spooler` into two groups of parameter-value pairs: `action=status` and `name=spooler`. PowerShell then (rightly) fails to bind the second set of params to any of the existing parameter sets and produces an error. The fix for this is to make `-params` have the `ValueFromRemainingArguments` attribute which will pull in all of the task params and store it in `-params`.

This fix however falls into a subtle trap in the switch statement in `Get-BoltCommandLine`. In PowerShell switch statements unwrap arrays and since we 'fixed' `-params` to now be an array, it takes `action=status name=spooler` and makes `--params action=status --params name=spooler`, which bolt (rightly) fails to process. To fix this an if/else statement is used. While this is more verbose than the switch statement, it ensures correct parsing of arrays, and presumably allows for more parameter types to be handled correctly in the future.

As a result of this investigation, I found/rediscovered the `$params` variable used in `Invoke-BoltCommandLine`. Since Invoke-BoltTask uses a parameter called `-params`, and parameters are variables, there are two 'uses' of the variable `$params`. This could cause a subtle bug where the value of `$params` is reassigned before it could be used, but we avoid this because we use `$PSBoundParameters` instead of individual parameters when building the commandline in `Get-BoltCommandLine`. To be safe, I renamed `$params` to be `$executionParams` to avoid any name clashes.

Acceptance Criteria:

To test:

Clone the repo, then build the cmdlets and import them:
```
PS > bundle exec rake pwsh:generate_module
PS > ipmo ./pwsh_module/puppetbolt.psd1 -for
```

The following should all work:

```
Invoke-BoltTask -t 'localhost' -name 'service' name=spooler action=status
Invoke-BoltTask -t 'localhost' 'service' name=spooler action=status
Invoke-BoltTask -t 'localhost' -name 'service' -params '{"name":"spooler","action":"status"}'
Invoke-BoltTask -t 'localhost' 'service' -params '{"name":"spooler","action":"status"}'
Invoke-BoltTask -t 'localhost' -name 'service' -params @{ name = "spooler"; action = "status" }
```

### DIFF
--- a/rakelib/pwsh.rake
+++ b/rakelib/pwsh.rake
@@ -322,7 +322,7 @@ namespace :pwsh do
     source = File.expand_path(File.join(__dir__, '..', 'guides'))
     dest   = File.expand_path(File.join(__dir__, '..', 'pwsh_module', 'en-US'))
 
-    FileUtils.mkdir(dest)
+    FileUtils.mkdir(dest) unless File.exist?(dest)
 
     Dir.children(source).each do |file|
       next if file !~ /\.txt\z/
--- a/rakelib/pwsh.rake
+++ b/rakelib/pwsh.rake
@@ -292,8 +292,10 @@ namespace :pwsh do
             pwsh_param[:mandatory] = true
             pwsh_param[:position] = 0
           when 'params'
+            pwsh_param[:mandatory] = false
             pwsh_param[:position] = 1
             pwsh_param[:type] = nil
+            pwsh_param[:value_from_remaining_arguments] = true
           when 'modules'
             pwsh_param[:type] = nil
           when 'format'
--- a/pwsh_module/command.tests.ps1
+++ b/pwsh_module/command.tests.ps1
@@ -4,7 +4,7 @@ BeforeAll {
   Mock -ModuleName 'PuppetBolt' -Verifiable -CommandName Invoke-BoltCommandLine -MockWith {
     return "bolt " + $params -join " "
   }
-  
+
   Mock Get-ItemProperty {
     return [PSCustomObject]@{
       RememberedInstallDir = 'C:/Program Files/Puppet Labs/Bolt'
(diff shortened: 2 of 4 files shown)


jpogran:
Example run:

```
PS > ipmo ./pwsh_module/puppetbolt.psd1 -for
PS > bundle exec rake pwsh:generate_module
PS > Invoke-BoltTask -t 'localhost' -name 'service' name=spooler action=status
Started on localhost...
Finished on localhost:
  {
    "status": "running",
    "enabled": "true"
  }
Ran on 1 target in 7.56 sec
PS > Invoke-BoltTask -t 'localhost' 'service' name=spooler action=status
Started on localhost...
Finished on localhost:
  {
    "status": "running",
    "enabled": "true"
  }
Ran on 1 target in 7.26 sec
PS > Invoke-BoltTask -t 'localhost' -name 'service' -params '{"name":"spooler","action":"status"}'
Started on localhost...
Finished on localhost:
  {
    "status": "running",
    "enabled": "true"
  }
Ran on 1 target in 8.12 sec
PS > Invoke-BoltTask -t 'localhost' 'service' -params '{"name":"spooler","action":"status"}'
Started on localhost...
Finished on localhost:
  {
    "status": "running",
    "enabled": "true"
  }
Ran on 1 target in 8.84 sec
PS > Invoke-BoltTask -t 'localhost' -name 'service' -params @{ name = "spooler"; action = "status" }
Started on localhost...
Finished on localhost:
  {
    "status": "running",
    "enabled": "true"
  }
Ran on 1 target in 7.9 sec
```

rakelib/pwsh.rake
```diff
@@ -322,7 +324,7 @@ namespace :pwsh do
     source = File.expand_path(File.join(__dir__, '..', 'guides'))
     dest   = File.expand_path(File.join(__dir__, '..', 'pwsh_module', 'en-US'))
 
-    FileUtils.mkdir(dest)
+    FileUtils.mkdir(dest) unless File.exist?(dest)
```
beechtom, line 327: 🎉  Thanks for picking this up!

PR #2119, opened 2020-08-25
(maint) Pin Puppet to 6.17.0

beechtom:
This temporarily pins Puppet to 6.17.0 until we can merge in
puppetlabs/bolt#2051 and puppetlabs/bolt#2074, as the Puppet 6.18.0
release includes changes that rely on these being merged. Without
pinning Puppet to 6.17.0, Bolt will fail when run from source.

!no-release-note

### DIFF
--- a/bolt.gemspec
+++ b/bolt.gemspec
@@ -48,7 +48,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency "net-ssh", ">= 4.0"
   spec.add_dependency "net-ssh-krb", "~> 0.5"
   spec.add_dependency "orchestrator_client", "~> 0.4"
-  spec.add_dependency "puppet", [">= 6.16.0", "< 7"]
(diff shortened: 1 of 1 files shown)


bolt.gemspec
```diff
@@ -48,7 +48,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency "net-ssh", ">= 4.0"
   spec.add_dependency "net-ssh-krb", "~> 0.5"
   spec.add_dependency "orchestrator_client", "~> 0.4"
-  spec.add_dependency "puppet", [">= 6.16.0", "< 7"]
+  spec.add_dependency "puppet", "6.17.0"
```
donoghuc, line 51: This will break vanagon packaging https://github.com/puppetlabs/puppet-runtime/blob/master/configs/components/rubygem-puppet.rb

beechtom: Updated

Issue #2120, opened 2020-08-25
Allow log files to be disabled


nicklewis:
## Use Case
There's presently no way for a project to remove/disable a log file that is defined in `bolt-defaults.yaml`. This happens simply because we define hashes of log files and merge them and have no mechanism to remove a key from the hash.

This is particularly important now that we always define a default log file.

This feature was initially defined as part of #2068 but was missed in the implementation of that ticket.

## Describe the Solution You Would Like

We should look for `<file>: disable` in the `log` section and remove the key after merging.

beechtom:
Closes by #2121

PR #2121, opened 2020-08-25
(GH-2120) Disable log files when set to "disable"

nicklewis:
We now allow users to set a log config to the string "disable" to indicate
that the file should be disabled if it was defined at a previous level of
the config hierarchy. This also allows users to disable the default
bolt-debug.log file.

This behavior actually worked already, but only incidentally as we happened
to skip over any logs whose config wasn't a Hash. We now explicitly check
for the string "disable" and otherwise fail if the value isn't a Hash.

!feature

  * **Disable log files by setting them to `disable`**
    ([#2120](https://github.com/puppetlabs/bolt/issues/2120))

    Log files can now be disabled if they were set at a previous level
   of the hierarchy. This also allows the default `bolt-debug.log` file
   to be disabled.

### DIFF
--- a/lib/bolt/config.rb
+++ b/lib/bolt/config.rb
@@ -338,9 +338,17 @@ module Bolt
 
     private def update_logs(logs)
       logs.each_with_object({}) do |(key, val), acc|
-        next unless val.is_a?(Hash)
+        # Remove any disabled logs
+        next if val == 'disable'
 
         name = normalize_log(key)
+
+        # But otherwise it has to be a Hash
+        unless val.is_a?(Hash)
+          raise Bolt::ValidationError,
+                "config of log #{name} must be a Hash, received #{val.class} #{val.inspect}"
+        end
+
         acc[name] = val.slice('append', 'level')
                        .transform_keys(&:to_sym)
 
--- a/lib/bolt/config/options.rb
+++ b/lib/bolt/config/options.rb
@@ -176,7 +176,8 @@ module Bolt
           description: "A map of configuration for the logfile output. Under `log`, you can configure log options "\
                        "for `console` and add configuration for individual log files, such as "\
                        "`~/.puppetlabs/bolt/debug.log`. Individual log files must be valid filepaths. If the log "\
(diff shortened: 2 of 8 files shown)


lib/bolt/config/options.rb
```diff
@@ -194,7 +195,7 @@ module Options
           },
           additionalProperties: {
             description: "Configuration for the logfile output.",
-            type: Hash,
+            type: [String, Hash],
```
beechtom, line 198:
```suggestion
            type: [String, Hash],
            enum: ["disable"],
```

This will restrict string values to only allow `disable` in the schema.

nicklewis: Ah perfect, I was looking for something like that but failed to find it.

lib/bolt/config.rb
```diff
@@ -338,9 +338,17 @@ def deep_clone
 
     private def update_logs(logs)
       logs.each_with_object({}) do |(key, val), acc|
-        next unless val.is_a?(Hash)
+        # Remove any disabled logs
+        next if val == 'disable'
```
beechtom, line 342: Do we want to allow users to disable `console` as well? The [schema](https://github.com/puppetlabs/bolt/pull/2121/files#diff-893a47fc3cbeb564a562c04c9348ae96R185) currently says that it only permits a hash.

nicklewis: I started to make that update when I was adding the `enum` property for other logs. It's complicated because `--log-level` turns into a log entry that will override `disable`. I'm not sure we want that.

Right now you _can't_ disable console, because `Bolt::Logger` will define it regardless of whether you specify it in your config or not. I want to move away from using the console as a log destination entirely, so I'm not super concerned about allowing it to be disabled now. The only thing you could really accomplish with that would be to silence warnings and errors, which seems unwanted anyway.

hestonhoffman:
Where will this show up in the docs? Just in the config references?

nicklewis:
Updated with a couple of mentions of `bolt-debug.log` in the documentation, both in the config file reference and on the projects page.

Issue #2117, opened 2020-08-24
Version 2.24.1 of this gem breaks litmus tests


mmarod:
## Describe the Bug
The litmus rake tasks fail with 2.24.1

## Expected Behavior
That the litmus rake tasks work

## Steps to Reproduce
```
$ pdk bundle exec rake 'litmus:provision[docker,litmusimage/ubuntu:16.04]'
pdk (INFO): Using Ruby 2.5.8
pdk (INFO): Using Puppet 6.17.0
rake aborted!litmusimage/ubuntu:16.04 using docker provisioner.[/]
Provisioning litmusimage/ubuntu:16.04 using docker provisioner.[-]NoMethodError: undefined method `trace' for #<Logging::Logger:0x00007f7fd37f5e90>
Did you mean?  trap
/Users/mmarod/.pdk/cache/ruby/2.5.0/gems/bolt-2.24.0/lib/bolt/pal/logging.rb:15:in `handle'
/opt/puppetlabs/pdk/private/puppet/ruby/2.5.0/gems/puppet-6.17.0/lib/puppet/util/log.rb:196:in `block in newmessage'
/opt/puppetlabs/pdk/private/puppet/ruby/2.5.0/gems/puppet-6.17.0/lib/puppet/util/log.rb:195:in `each'
/opt/puppetlabs/pdk/private/puppet/ruby/2.5.0/gems/puppet-6.17.0/lib/puppet/util/log.rb:195:in `newmessage'
/opt/puppetlabs/pdk/private/puppet/ruby/2.5.0/gems/puppet-6.17.0/lib/puppet/util/log.rb:315:in `initialize'
/opt/puppetlabs/pdk/private/puppet/ruby/2.5.0/gems/puppet-6.17.0/lib/puppet/util/log.rb:88:in `new'
/opt/puppetlabs/pdk/private/puppet/ruby/2.5.0/gems/puppet-6.17.0/lib/puppet/util/log.rb:88:in `create'
/opt/puppetlabs/pdk/private/puppet/ruby/2.5.0/gems/puppet-6.17.0/lib/puppet/util/logging.rb:277:in `block in setup_facter_logging!'
/Users/mmarod/.pdk/cache/ruby/2.5.0/gems/facter-4.0.35/lib/facter/framework/logging/logger.rb:66:in `debug'
[...]

PR #2122, opened 2020-08-26
(GH-2117) Initialize logger in BoltSpec::Run

beechtom:
This updates `BoltSpec::Run` to initialize Bolt's logger as part of
initializing the `BoltRunner`. Previously, the logger was not
initialized, which would cause BoltSpec to fail whenever Bolt attempted
to log a message to the `trace` level, which is not a level recognized
by default in the `Logging` class. This fix ensures that `BoltSpec::Run`
will always initialize the logger in the same manner as Bolt itself.

!bug

* **Initialize logger in `BoltSpec::Run`**
  ([#2117](https://github.com/puppetlabs/bolt/issues/2117))

  Bolt now initializes the logger when using `BoltSpec::Run` methods.
  Previously, the logger was not initialized with Bolt's custom log
  levels, causing `BoltSpec::Run` to raise an error when it encountered
  a message being logged to one of these custom levels.

### DIFF
--- a/lib/bolt_spec/run.rb
+++ b/lib/bolt_spec/run.rb
@@ -8,6 +8,7 @@ require 'bolt/pal'
 require 'bolt/plugin'
 require 'bolt/puppetdb'
 require 'bolt/util'
+require 'bolt/logger'
 
 # This is intended to provide a relatively stable method of executing bolt in process from tests.
 module BoltSpec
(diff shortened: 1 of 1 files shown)


nicklewis:
This should make the `RSpec.configure` blocks in the `spec_helper.rb` of each directory of `bolt-modules` unnecessary. Can you make that change and confirm that this solves it?

nicklewis:
Hmm it doesn't seem to. That's because we don't actually use `BoltSpec` to test those functions. It might be that what our functions do is different or it might be that this change isn't sufficient to fix litmus...

lucywyman:
This solved the Litmus error for me in the service module

Issue #2007, opened 2020-07-21
Add docs on how to use out::message for debugging


npwalker:
## Describe the Change You Would Like

I’m currently doing
```
   $show_databases = run_task('mysql::sql', $targets, "Show the new database is there", {'sql' => "SHOW DATABASES"})
-  return $show_databases
+  out::message($show_databases.first.message)
```

which results in

```
MacBook-Pro:automations nw$ bolt plan run automations::apache_plus_mysql_converted --targets="agent-no-pe-centos" --no-host-key-check --user=centos --run-as=root
CLI arguments ["user", "host-key-check", "run-as"] may be overridden by Inventory: /Users/nw/git_repos/automations/inventory.yaml
Starting: plan automations::apache_plus_mysql_converted
Starting: Show the new database is there on agent-no-pe-centos
Finished: Show the new database is there with 0 failures in 4.44 sec
Finished: plan automations::apache_plus_mysql_converted in 4.46 sec
{
  "kind": "bolt/pal-error",
  "msg": "'out::message' parameter 'message' expects a String value, got Undef",
  "details": {
    "file": "/Users/nw/git_repos/automations/plans/apache_plus_mysql_converted.pp",
    "line": 13,
    "column": 3
  }
}
```

but I think I also tried results.message as well which is a different problem

Cas recommends 

Remember run_* returns a ResultSet. For debugging with out::message I would encourage you to try to explicity wrap stuff in String (or just do interpolation).

```
$result_set = run_task(my_task, $targets)
out::message(String($result_set[0].value))
```

This makes sense but I couldn't find any reference on how to use out::message to output things for debugging purposes.  Having a few examples of outputting different useful parts of a result_set would probably allowed me to figure this out on my own.

adreyer:
Is there a reason to document `out::message` for debugging over the log functions?

npwalker:
Maybe debugging is the wrong word for what I was doing... I was actually trying to output in the plan what the output of the command I ran on the system was.

donoghuc:
Not that you would be necessarily debugging plans in PE console, but out::message is the way you can see that there.

lucywyman:
We should have a 'debugging plans' section on the writing plans pages (on both the yaml and puppet plan pages).
In the section we should document using `out::message()` to print strings (emphasizing that out::message can only print strings).

PR #2128, opened 2020-08-27
(GH-2007)(Docs) Debugging

hestonhoffman:
Updates the "writing plans" docs to give more information about
the different options for debugging plans.

I've also made some style changes re stdout, stderr, stdin. These are abbreviations, so the way we usually deal with them is by giving the full term the first time they're used, with the abbreviation in brackets, and then thereafter using the abbreviation. No need for backticks.

!no-release-note

closes GH-2007

### DIFF
--- a/documentation/writing_plans.md
+++ b/documentation/writing_plans.md
@@ -224,6 +224,86 @@ plan mymodule::myplan {
 }
 ```
 
+## Debugging plans
+
+By default, Bolt does not print verbose logs for each plan execution to stdout.
+However, you can use one of the following methods to investigate a plan
+execution:
+- Each time you run a Bolt command, Bolt prints a debug log to a
+  `bolt-debug.log` file in the root of your project directory.
+- You can use the `--verbose` CLI option for verbose logging to stdout.
+- You can print a message to stdout using the `out::message` function. 
+- You can adjust your log level for detailed information on how Bolt is
+  executing your plan.
+
+### Using `out::message` to debug a plan
+
+You can print a message, or a variable, to stdout using the `out::message`
+function. If the variable contains a valid plan result, Bolt formats the plan
+result using a JSON representation of the result object. If the object is not a
+plan result, Bolt prints the object as a string.
+
+To print a variable to stdout with `out::message`, use the following syntax:
+
+```puppet
+out::message($variable) 
+```
+
+For example, the following plan uses a `run_task` function to check targets for
+the MySQL package and assigns the result to a variable called `$check_mysql`.
+The plan uses `out::message($check_mysql)` to print the result set from
+`$check_mysql` to stdout.
+
+```puppet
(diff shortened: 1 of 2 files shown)


documentation/writing_plans.md
```diff
@@ -224,6 +224,80 @@ plan mymodule::myplan {
...
 
+## Debugging plans
+
+By default, Bolt does not print verbose logs for each plan execution to stdout.
+However, you can use one of the following methods to investigate a plan
+execution:
+- Each time you run a Bolt command, Bolt prints a debug log to a
+  `bolt-debug.log` file in the root of your project directory.
+- You can use the `--verbose` CLI option for verbose logging to stdout.
+- You can print a parameter to stdout using the `out::message` function.
```
lucywyman, line 235:
```suggestion
- You can print to stdout using the `out::message` function. 
```
Only since 'parameter' has specific meaning in plans, and you can print anything with out::message not just plan parameters

documentation/writing_plans.md
```diff
@@ -224,6 +224,80 @@ plan mymodule::myplan {
...
+- Each time you run a Bolt command, Bolt prints a debug log to a
+  `bolt-debug.log` file in the root of your project directory.
+- You can use the `--verbose` CLI option for verbose logging to stdout.
+- You can print a parameter to stdout using the `out::message` function. 
+- You can adjust your log level for detailed information on how Bolt is
+  executing your plan.
+
+### Using `out::message` to debug a plan
+
+You can print any plan parameter to stdout using the `out::message` function. If
```
lucywyman, line 241:
```suggestion
You can print anything to stdout using the `out::message` function. If
```

lucywyman: I mostly think this is useful because I frequently just print messages like "Got here" or other non-variable data to help debug plans.

lucywyman: Alternatively you might be looking for the word 'variable'?

documentation/writing_plans.md
```diff
@@ -224,6 +224,80 @@ plan mymodule::myplan {
...
+standard error (stderr), at the `debug` level. Use the `--log-level debug` CLI
+option or the [`log` configuration setting](bolt_project_reference.html#log).
+
+Each time you run a Bolt command, Bolt prints a debug level log to a
+`bolt-debug.log` file in the root of your project directory.
+
+```shell
+$ bolt task run mytask param1=foo param2=bar -t all --log-level debug
+```
+
```
lucywyman, line 300: I'd maybe add a line about "You can disable this behavior by specifying:
```
log:
  bolt-debug.log: disable
```
in bolt-project.yaml

documentation/writing_plans.md
hestonhoffman: Ah you're right - I think I meant variable.

PR #2136, opened 2020-08-28
add doc on how to use target name in hiera paths

jessereynolds:
This PR adds some additional points to the docs on using hiera with `bolt apply`, specifically how to use the target name as part of a hiera path.

### DIFF
--- a/documentation/applying_manifest_blocks.md
+++ b/documentation/applying_manifest_blocks.md
@@ -241,6 +241,13 @@ Following the Hiera 5 convention, the default data dir is relative to
 `hiera.yaml` at `$BOLTDIR/data`. For configfile examples, see [Configuring
 Hiera](https://puppet.com/docs/puppet/latest/hiera_config_yaml_5.html).
 
+The target name is made available to hiera via the `certname` key of the `$trusted` top level Puppet variable, ie `$trusted['certname']`. This can be interpolated in hierarchy paths within `$BOLTDIR/hiera.yaml` using `%{trusted.certname}`.
+
+```yaml
+    paths:
+      - "targets/%{trusted.certname}.yaml"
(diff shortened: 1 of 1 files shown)


lucywyman:
Thanks for putting this up Jesse! I think the reason we didn't include this is that we don't want to duplicate the Hiera docs, and so assumed Hiera knowledge in this doc and focused narrowly on using Hiera in the context of Bolt. There's a note in [this section of the Hiera docs](https://puppet.com/docs/puppet/6.18/hiera_merging.html#interpolate_puppet_variable) about `trusted.certname` - it still seems useful to include a note here that that's how you get a target's name, but it might be better to make it short and then link to this documentation?

documentation/applying_manifest_blocks.md
```diff
@@ -241,6 +241,13 @@ Following the Hiera 5 convention, the default data dir is relative to
 `hiera.yaml` at `$BOLTDIR/data`. For configfile examples, see [Configuring
 Hiera](https://puppet.com/docs/puppet/latest/hiera_config_yaml_5.html).
 
+The target name is made available to hiera via the `certname` key of the `$trusted` top level Puppet variable, ie `$trusted['certname']`. This can be interpolated in hierarchy paths within `$BOLTDIR/hiera.yaml` using `%{trusted.certname}`.
```
lucywyman, line 244:
```suggestion
> **Note**: The target name is made available to Hiera as `trusted.certname`. You can learn more about [interpolating variables into Hiera config here](https://puppet.com/docs/puppet/latest/hiera_merging.html#interpolate_puppet_variable).
```

hestonhoffman:
@jessereynolds What do you think about Lucy's suggestion above?

Issue #2112, opened 2020-08-24
Ensure tasks that print `_error` have a message


lucywyman:
## Describe the Bug
When a task prints JSON to stdout with an `_error` key we convert it to a Bolt error. However, we assume the value of the key will have a `msg` field and don't catch the stack trace when it doesn't. We should error when a task prints an invalid error object. We should also ensure that we inject kind and details.

tasks/mytask.rb
```
#!/usr/bin/env ruby

puts { '_error' => 'Fail' }.to_json
```

```
$ bolt task run module::mytask -t localhost
...
Failed on localhost:
bundler: failed to load command: bolt (/home/lucy/githubs/bolt/vendor/bundle/ruby/2.6.0/bin/bolt)
NoMethodError: undefined method `gsub' for nil:NilClass
  /home/lucy/githubs/bolt/lib/bolt/outputter.rb:27:in `indent'
  /home/lucy/githubs/bolt/lib/bolt/outputter/human.rb:82:in `print_result'
  /home/lucy/githubs/bolt/lib/bolt/outputter/human.rb:358:in `block in print_result_set'
  /home/lucy/githubs/bolt/lib/bolt/result_set.rb:41:in `block in each'
  /home/lucy/githubs/bolt/lib/bolt/result_set.rb:41:in `each'
  /home/lucy/githubs/bolt/lib/bolt/result_set.rb:41:in `each'
  /home/lucy/githubs/bolt/lib/bolt/outputter/human.rb:358:in `print_result_set'
  /home/lucy/githubs/bolt/lib/bolt/outputter/human.rb:349:in `print_plan_result'
  /home/lucy/githubs/bolt/lib/bolt/cli.rb:716:in `run_plan'
  /home/lucy/githubs/bolt/lib/bolt/cli.rb:437:in `execute'
  /home/lucy/githubs/bolt/exe/bolt:11:in `<top (required)>'
  /home/lucy/githubs/bolt/vendor/bundle/ruby/2.6.0/bin/bolt:23:in `load'
  /home/lucy/githubs/bolt/vendor/bundle/ruby/2.6.0/bin/bolt:23:in `<top (required)>'
```

PR #2137, opened 2020-08-28
(GH-2112) Ensure task error objects have correct format

beechtom:
This updates `Bolt::Result.for_task` to ensure that an error object
returned from a task has the correct format. Previously, any error
object returned from a task was accepted as-is, even if it did not have
the expected `msg`, `kind`, and `detail` keys. Now, a `msg` key is
required, and Bolt will raise a helpful error if the error object is
missing this key. If either the `kind` or `details` keys are missing
from the error object, Bolt will automatically add reasonable default
values of `bolt/error` and `{}` to the error object.

!bug

* **Ensure task error objects have correct format**
  ([#2112](https://github.com/puppetlabs/bolt/issues/2112))

  Bolt now ensures that error objects returned from a task have the
  correct format and include a `msg` key. Bolt will also automatically
  add `kind` and `details` keys if they are absent from the object, with
  default values of `bolt/error` and `{}`.

### DIFF
--- a/lib/bolt/result.rb
+++ b/lib/bolt/result.rb
@@ -66,9 +66,24 @@ module Bolt
                             'details' => { 'exit_code' => exit_code } }
       end
 
+      if value.key?('_error')
+        unless value['_error'].is_a?(Hash) && value['_error'].key?('msg')
+          value['_error'] = {
+            'msg'     => "Invalid error returned from task #{task}: #{value['_error'].inspect}. Error "\
+                         "must be an object with a msg key.",
+            'kind'    => 'bolt/invalid-task-error',
+            'details' => { 'original_error' => value['_error'] }
+          }
+        end
+
+        value['_error']['kind']    ||= 'bolt/error'
+        value['_error']['details'] ||= {}
+      end
+
       if value.key?('_sensitive')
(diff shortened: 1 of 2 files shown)


lib/bolt/result.rb
```diff
@@ -66,9 +66,23 @@ def self.for_task(target, stdout, stderr, exit_code, task)
                             'details' => { 'exit_code' => exit_code } }
       end
 
+      if value.key?('_error')
+        unless value['_error'].is_a?(Hash) && value['_error'].key?('msg')
+          raise Bolt::Error.new(
```
nicklewis, line 71: ....where does this error go? I guess it would probably be the fallback handler in the executor, which is meant to only handle transport bugs. Should we be inserting this as the `_error` key of the result instead of raising it? Maybe with something like `{"details" => { "original_error" => [the original _error] } }`

nicklewis: Ah looks like it's caught in `Bolt::Transport::Base#with_events` which calls `Bolt::Result.from_exception`, so that's not so bad. I do think it makes more sense to include this error in the original result though, so we preserve any _other_ result keys the user might care about.

PR #2138, opened 2020-08-28
(GH-2116) Handle file writing errors

### LINKED ISSUE #2116: Handle non-existent projects more gracefully
## Use Case

Currently, the behavior of Bolt when a project doesn't exist is relatively undefined. In some cases, we will create the project but only at the end of the run (as a side effect of writing `.rerun.json`) and other times we won't create it at all. Given that Bolt is _always_ running within the context of a project, even if that project happens to be the default one and not actually exist, we should settle once and for all the question of whether the project needs to exist.

We should assert that the project Bolt is using must exist. That means either creating it or failing in the cases where it might not already exist.

There are a number of methods whereby Bolt can determine its project.

**System-level projects**

The Bolt package should create the system-level project directory (`/etc/puppetlabs/bolt` or <common_appdata>/PuppetLabs/bolt/etc`). If Bolt decides to use the system-level project and it doesn't exist, it should fail. We shouldn't create the project at run time because we may not be running as an appropriate user to do so. Worth noting that on Windows common_appdata is per user (same for user-level directories).

**User-level projects**

If Bolt decides to use the user-level default project, it should create the directory.

**Running inside a project**

By definition, the project directory has to already exist for this case.

**Specifying a project with `--project`**

_Should we fail or create the project here?_

**Specify a project with `--configfile`**

This should work roughly the same as `--project`. _What about the case where the directory exists but the file doesn't?_

**Environment variable projects**

This should probably behave the same as `--project`.

We could instead declare that the Bolt project directory _doesn't_ need to exist. In that case, we need to decide what to do about files that we manage inside the project directory like `bolt-debug.log` and `.rerun.json`. Are those files simply not written if the project doesn't exist?

## Solution
- [x] If we don't find a project, create a user level project.
- [x] If we can't make the project, warn once gracefully and continue without a project.
- [x] If we can't write to the project, warn that we can't write files to the project and continue.
- [x] If we can't write any individual file we should warn that we couldn't write that file and not warn about not being able to write other files.

How should we handle log files we can't write to?
- [x] If the users has explicitly requested a log and we can't write to it, we should error. This means we need to check that the default logfile is writeable before merging in user-level config, so that we don't fail when the default logfile isn't writeable.

If the user points to a project that doesn't exist, we should error.

lucywyman:
This cleans up how we handle failures when Bolt fails to write a file to
the current project. We now attempt to create the project directory if
it is the default directory and does not exist. If that creation fails,
we warn and continue without a Bolt project - this effectively means
that Bolt has default configuration, and that logfiles and rerun files
will not be written. If any individual file can't be written to the
project, we warn once and continue. If multiple files can't be written,
we warn for the first file that isn't writeable and not for subsequent
files.

When users explicitly configure a logfile that is unwriteable we now
error.

When users explicitly configure a project that is unwriteable, we now
error.

Closes #2116

!bug

* **Handle project file writing errors more gracefully** ([#2116](https://github.com/puppetlabs/bolt/issues/2116))

  Bolt will now warn and continue executing when it fails to write files
  to the active project. Additionally, any user-specified file data that
  fails to be written to will error.

### DIFF
--- a/lib/bolt/analytics.rb
+++ b/lib/bolt/analytics.rb
@@ -93,6 +93,10 @@ module Bolt
     def self.write_config(filename, config)
       FileUtils.mkdir_p(File.dirname(filename))
       File.write(filename, config.to_yaml)
+    rescue StandardError => e
+      Bolt::Logger.warn_once('unwriteable_file', "Could not write analytics configuration to #{filename}.")
+      # This will get caught by build_client and create a NoopClient
+      raise e
     end
 
     class Client
--- a/lib/bolt/config.rb
+++ b/lib/bolt/config.rb
@@ -64,7 +64,7 @@ module Bolt
              end
 
       data = load_defaults(project).push(
-        filepath: project.config_file,
+        filepath: configfile,
         data: conf,
         logs: logs,
         deprecations: []
@@ -344,6 +344,14 @@ module Bolt
     end
 
     private def update_logs(logs)
+      begin
+        if logs['bolt-debug.log'] && logs['bolt-debug.log'] != 'disable'
+          FileUtils.touch(File.expand_path('bolt-debug.log', @project.path))
+        end
+      rescue StandardError
+        logs.delete('bolt-debug.log')
+      end
+
       logs.each_with_object({}) do |(key, val), acc|
         # Remove any disabled logs
         next if val == 'disable'
--- a/lib/bolt/project.rb
+++ b/lib/bolt/project.rb
@@ -50,6 +50,20 @@ module Bolt
     def self.create_project(path, type = 'option', logs = [])
       fullpath = Pathname.new(path).expand_path
 
+      if type == 'user'
+        begin
+          # This is already expanded if the type is user
+          FileUtils.mkdir_p(path)
+        rescue StandardError
+          logs << { warn: "Could not create default project at #{path}. Continuing without a writeable project. "\
+                    "Log and rerun files will not be written." }
+        end
+      end
+
+      if type == 'option' && !File.directory?(path)
+        raise Bolt::Error.new("Could not find project at #{path}", "bolt/project-error")
+      end
+
       if !Bolt::Util.windows? && type != 'environment' && fullpath.world_writable?
         raise Bolt::Error.new(
           "Project directory '#{fullpath}' is world-writable which poses a security risk. Set "\
--- a/lib/bolt/rerun.rb
+++ b/lib/bolt/rerun.rb
@@ -53,7 +53,7 @@ module Bolt
         end
       end
     rescue StandardError => e
-      @logger.warn("Failed to save result to #{@path}: #{e.message}")
+      Bolt::Logger.warn_once('unwriteable_file', "Failed to save result to #{@path}: #{e.message}")
     end
   end
 end
--- a/spec/bolt/analytics_spec.rb
+++ b/spec/bolt/analytics_spec.rb
@@ -85,7 +85,17 @@ describe Bolt::Analytics do
       subject.build_client
     end
 
-    it 'warns when user-level config and defaul project config both exist' do
+    it 'errors if new config cannot be written' do
+      allow(subject).to receive(:write_config).and_call_original
+      allow(File).to receive(:write)
+        .and_raise(Errno::EACCES, "Permission denied")
+
+      subject.build_client
+
+      expect(@log_output.readlines).to include(/Could not write analytics/)
+    end
+
+    it 'warns when user-level config and default project config both exist' do
       allow(File).to receive(:exist?).with(path).and_return(true)
       allow(File).to receive(:exist?).with(old_path).and_return(true)
 
--- a/spec/bolt/config_spec.rb
+++ b/spec/bolt/config_spec.rb
@@ -9,10 +9,13 @@ describe Bolt::Config do
   let(:user_path)     { Bolt::Config.user_path }
   let(:config_name)   { Bolt::Config::BOLT_CONFIG_NAME }
   let(:defaults_name) { Bolt::Config::BOLT_DEFAULTS_NAME }
+  let(:logfile)       { 'bolt.log' }
 
   around(:each) do |example|
     Dir.mktmpdir("foo") do |tmpdir|
-      @tmpdir = Pathname.new(File.join(tmpdir, "validprojectname")).to_s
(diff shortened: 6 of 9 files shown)


lib/bolt/config.rb
```diff
@@ -340,7 +340,16 @@ def deep_clone
     private def normalize_log(target)
       return target if target == 'console'
       target = target[5..-1] if target.start_with?('file:')
-      'file:' + File.expand_path(target, @project.path)
+      file = File.expand_path(target, @project.path)
+      begin
+        # It's easier to do this here and ignore the default log than to rewalk
+        # config and get the filenames.
+        FileUtils.touch(file) unless target == 'bolt-debug.log'
```
nicklewis, line 347: We already fail in `Bolt::Logger` if a log file isn't writeable. So I think what we actually want is to try to touch bolt-debug.log and _remove_ it from the list of logs if that fails. Then we can let `Bolt::Logger` handle the others later in the process.

lib/bolt/project.rb
```diff
@@ -50,6 +50,20 @@ def self.find_boltdir(dir, logs = [])
     def self.create_project(path, type = 'option', logs = [])
       fullpath = Pathname.new(path).expand_path
 
+      if %w[user system].include?(type)
```
nicklewis, line 53: We shouldn't try to create the system level project here. It's almost guaranteed to fail and it's not really desirable anyway, since a system level directory shouldn't be owned by an arbitrary user.

lib/bolt/analytics.rb
```diff
@@ -93,6 +93,9 @@ def self.load_config(filename, logger)
     def self.write_config(filename, config)
       FileUtils.mkdir_p(File.dirname(filename))
       File.write(filename, config.to_yaml)
+    rescue StandardError => e
+      Bolt::Logger.warn_once('unwriteable_file', "Could not write analytics configuration to #{filename}; "\
```
nicklewis, line 97: I think "continuing" is redundant, since this is a warning.

We should also disable analytics in this case, otherwise we'll generate a new uuid every run. So maybe in place of "continuing" we could say "disabling analytics".

lucywyman: It seems like the easiest way to do that is to have this warn then fail, and the failure will get caught by `build_client` and create a NoopClient and print a message.

lib/bolt/config.rb
```diff
@@ -344,6 +344,12 @@ def deep_clone
     end
 
     private def update_logs(logs)
+      begin
+        FileUtils.touch('bolt-debug.log')
```
nicklewis, line 348: This needs to be expanded relative to the project directory.

lib/bolt/config.rb
```diff
@@ -344,6 +344,12 @@ def deep_clone
...
+      rescue StandardError
+        logs.delete_if { |k, _| k == 'bolt-debug.log' }
```
nicklewis, line 350: `logs.delete('bolt-debug.log')` should work here

nicklewis (changes requested):

lib/bolt/config.rb
```diff
@@ -64,7 +64,7 @@ def self.from_file(configfile, overrides = {})
              end
 
       data = load_defaults(project).push(
-        filepath: project.config_file,
+        filepath: configfile,
```
nicklewis, line 67: If you pass `--configfile /path/to/bolt.yaml`, we will actually load `/path/to/bolt-project.yaml` if it exists. Will this cause us to instead report that we loaded `/path/to/bolt.yaml` when we didn't?

lucywyman: I don't think so, that logging happens above on line 62 and is only if the configfile *isn't* bolt-project.yaml:
```
lucy ~/githubs/bolt (GH-2116) $ be bolt plan show --configfile project/bolt.yaml --log-level debug
Analytics opt-out is set, analytics will be disabled
Loaded project from '/home/lucy/githubs/bolt/project'
bolt-project.yaml contains valid config keys, bolt.yaml will be ignored
Loaded configuration from /home/lucy/.puppetlabs/etc/bolt/bolt-defaults.yaml
Project-level configuration in bolt.yaml is deprecated if using bolt-project.yaml. Transport config should be set in inventory.yaml, all other config should be set in bolt-project.yaml.
Loading modules from /home/lucy/githubs/bolt/bolt-modules:/home/lucy/githubs/bolt/project/modules:/home/lucy/githubs/bolt/modules:/home/lucy/githubs/bolt/project/spec/fixtures/modules:/home/lucy/githubs/bolt/project/spec/fixtures/apply:/home/lucy/githubs/bolt/modules
aggregate::count
aggregate::nodes
aggregate::targets
canary
facts
facts::external
facts::info
puppet_agent::run
puppetdb_fact
reboot
secure_env_vars
terraform::apply
terraform::destroy
```

lucywyman: This is kind of hacky way to make this *slightly* more correct but not entirely. The `project.config_file` will still be wrong, but we just don't actually use it anywhere so it doesn't matter. Since we're removing it soon I'm not inclined to do the restructuring needed to make it correct.

nicklewis: Ah right I forgot we got rid of the `config_loaded` method, which was using this. :thumbsup:

lib/bolt/config.rb
```diff
@@ -344,6 +344,12 @@ def deep_clone
     end
 
     private def update_logs(logs)
+      begin
+        FileUtils.touch(File.expand_path('bolt-debug.log', @project.path))
```
nicklewis, line 348: Oh hmm, we do this before we handle the fact that the log might be disabled (and before we validate its config), so we will still create it even if the user disabled it.

It's a bit of a pain though because the loop below also calls `normalize_log`, which obscures the location of `bolt-debug.log` within the `log` map. It honestly might be easiest to just repeat the check here. We could also use that as an opportunity to verify that `bolt-debug.log` is indeed a defined log (there's no way it wouldn't be _now_, but that could change).

```rb
if logs['bolt-debug.log'] && logs['bolt-debug.log'] != 'disable'
  FileUtils.touch(File.expand_path('bolt-debug.log', @project.path))
end
```

nicklewis: Also this block could probably use a comment since its intent is non-obvious.

lucywyman: Oh dang, good point. I'll just do that.

nicklewis (approved):

PR #2141, opened 2020-08-31
(PE-29495) Serve task metadata from bolt-projects managed with code manager

donoghuc:
Add new endpoints to bolt-server capable of listing tasks and task metadata from bolt projects.

### DIFF
--- a/developer-docs/bolt-api-servers.md
+++ b/developer-docs/bolt-api-servers.md
@@ -25,6 +25,7 @@ Most options are shared by the bolt server and plan executor applications
 
 **Bolt Server Only Options**
 - `concurrency`: Integer, *optional* - The maximum number of server threads (default `100`).
+- `projects-dir`: String, *optional* - Path to bolt-projects dir managed by code manager when serving bolt-project content for team console.
 
 **Plan Executor Only Options**
 - `modulepath`: String, *required* - The path to modules to read plans from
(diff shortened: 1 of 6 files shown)


developer-docs/bolt-api-servers.md
```diff
@@ -404,6 +405,22 @@ The files array is required, and contains details about the files the task needs
...
+```
```
steveax, line 422: It looks like most of the examples show example payload as formatted json rather than a compressed curl command. That seems perhaps more useful?

donoghuc: This is a response.

lib/bolt_server/transport_app.rb
```diff
@@ -440,8 +476,22 @@ def make_winrm_target(target_hash)
...
+        # makes switching between puppetserver and bolt-server easier, which makes changes to switch
+        # to bolt-server smaller/simpler.
+        [200, tasks_response]
+      end
+    end
+
+    # Fetches the list of tasks for a bolt-project
+    #
+    # @param environment [String] the environment to fetch the list of tasks from
+    get '/bolt_project_tasks' do
```
donoghuc, line 492: I guess we are going with `project` prepended, not `bolt_project`. :upside_down_face:

lib/bolt_server/transport_app.rb
```diff
@@ -271,6 +287,10 @@ def pe_task_info(pal, module_name, task_name, environment)
       }
     end
 
+    def task_list(pal)
+      pal.list_tasks.map { |task_name, _description| { 'name' => task_name } }
```
donoghuc, line 291: This needs to be updated to filter tasks based on allowlist in bolt-project.yaml

donoghuc: Updated.

lib/bolt_server/transport_app.rb
```diff
@@ -191,20 +192,43 @@ def run_script(target, body)
...
+        project = Bolt::Project.create_project(project_dir)
+        bolt_config = Bolt::Config.from_project(project, {})
+        begin
+          pal = Bolt::PAL.new(bolt_config.modulepath, nil, nil, nil, nil, nil, bolt_config.project)
+        rescue StandardError => e
+          raise e.backtrace.to_s
         end
+        module_path = [BoltServer::PE::PAL::PE_BOLTLIB_PATH, Bolt::PAL::BOLTLIB_PATH, *bolt_config.modulepath]
+        # CODEREVIEW: I *think* this is the only thing we need to make different between bolt's PAL. Is it acceptable
+        # to hack this? Modulepath is currently a readable attribute, could we make it writeable?
```
McdonaldSeanp, line 223: This was one of the primary reasons PE::PAL exists, is there a reason not to update the PE::PAL implementation so that it can take project instead of environment?

donoghuc: PE::PAL is very environment focused. For bolt projects we want  both the PAL object as well as the project config in the overall context. Given there is only a single change we actually need it seemed reasonable to just use a *slightly* modified bolt PAL.

McdonaldSeanp: Yeah, I can understand where you're coming from since we don't need much of the functionality we've written for PE::PAL. 

I'm a bit wary of allowing modulepath as a writable attribute because we don't actually need it to be "writable" in an object sense: we just need to change it the one time, and "writable" attributes to me means they are intended to be changed during the lifetime of the object.

I wonder if there's a way we can change bolt's PAL such that we can tell it not to do the module_path munging and instead take the module_path param to the initializer wholesale.

If we did that, we could potentially upstream it to orch and make our code similar.

donoghuc: `Bolt::PAL::initialize` signature is already pretty busy. Will defer to @puppetlabs/bolt on that idea.

lib/bolt_server/transport_app.rb
```diff
@@ -191,20 +192,43 @@ def run_script(target, body)
...
+
+    def in_bolt_project(bolt_project)
+      return [400, '`project_ref` is a required argument'] if bolt_project.nil?
+      project_dir = File.join(@config['projects-dir'], bolt_project)
+      return [400, "`project_ref`: #{project_dir} does not exist"] unless Dir.exist?(project_dir)
+      @pal_mutex.synchronize do
+        project = Bolt::Project.create_project(project_dir)
+        bolt_config = Bolt::Config.from_project(project, {})
+        begin
+          pal = Bolt::PAL.new(bolt_config.modulepath, nil, nil, nil, nil, nil, bolt_config.project)
```
donoghuc, line 217: OOPS. need to remove this debugging...

lib/bolt_server/transport_app.rb
```diff
@@ -191,20 +192,39 @@ def run_script(target, body)
...
+
+    def in_bolt_project(bolt_project)
+      return [400, '`project_ref` is a required argument'] if bolt_project.nil?
+      project_dir = File.join(@config['projects-dir'], bolt_project)
+      return [400, "`project_ref`: #{project_dir} does not exist"] unless Dir.exist?(project_dir)
+      @pal_mutex.synchronize do
+        project = Bolt::Project.create_project(project_dir)
+        bolt_config = Bolt::Config.from_project(project, {})
+        pal = Bolt::PAL.new(bolt_config.modulepath, nil, nil, nil, nil, nil, bolt_config.project)
+        module_path = [BoltServer::PE::PAL::PE_BOLTLIB_PATH, Bolt::PAL::BOLTLIB_PATH, *bolt_config.modulepath]
```
donoghuc, line 217: This ignores modules shipped in bolt repo :(

PR #2142, opened 2020-08-31
(PE-29989) task list metadata docs

steveax:


donoghuc:
@puppetlabs/bolt will this break anything given none of the commit messages have `!no-release-note`?

nicklewis:
The release note check doesn't run for docs-only changes, so it should be fine

Issue #2108, opened 2020-08-21
Warn when project shadows a module with the same name


nicklewis:
## Use Case

A simple pattern for running tasks/plans from some module with an inventory is to create a project that simply references the module and includes an `inventory.yaml`. It's intuitive to give this project the same name as the module it's wrapping. However, this actually causes the (non-existent) project-level content to shadow the module with the same name, requiring the project to actually have a _different_ name.

This shadowing is a natural consequence of projects having their own content and is not a bug, but it _is_ surprising behavior that can cause confusion when tasks/plans don't show up as expected.

## Describe the Solution You Would Like

We should warn at some point during Bolt's lifecycle, either every time a command is run or only when using tasks/plans or only when installing modules, that the project name will conflict with a module of the same name on the modulepath.

We should raise this warning when we call PAL#in_bolt_compiler when loading modules.

beechtom:
Closed by #2143

PR #2143, opened 2020-08-31
(GH-2108) Warn when Bolt project shadows module of the same name

nicklewis:
!feature

  * **Bolt will now warn if a project has the same name as a module**
     ([#2108](https://github.com/puppetlabs/bolt/issues/2108))

    Any module with the same name as the Bolt project will be ignored,
    and Bolt will now issue a warning to indicate that.

### DIFF
--- a/lib/bolt/pal.rb
+++ b/lib/bolt/pal.rb
@@ -141,6 +141,14 @@ module Bolt
       end
     end
 
+    def detect_project_conflict(project, modules)
+      return unless project
+      if modules.any? { |mod| mod.name == project.name }
+        Bolt::Logger.warn_once("project shadows module",
+                               "The project '#{project.name}' shadows an existing module of the same name")
+      end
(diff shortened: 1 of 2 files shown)


lib/bolt/pal.rb
```diff
@@ -141,6 +141,14 @@ def register_resource_types(loaders)
       end
     end
 
+    def detect_project_conflict(project, modules)
+      return unless project
+      if modules.any? { |mod| mod.name == project.name }
+        Bolt::Logger.warn_once("project shadows module",
+                               "The project '#{project.name}' will shadow an existing module of the same name")
```
beechtom, line 148:
```suggestion
                               "The project '#{project.name}' shadows an existing module of the same name")
```

I think removing the tense makes this read a little clearer and avoids any potential confusion around 'will shadow' being in the future.

PR #2145, opened 2020-08-31
(GH-2108) Don't report project as conflicting with itself

nicklewis:
This was always warning about the project shadowing a module with the
same name, because we load the project as a module and it thus
"conflicted" with itself. We now explicitly check that the path to the
module is different than the path to the project.

We also bypass the check entirely if we're not loading the project as a
module. This is just a short-circuit that doesn't change the correctness
of the method, since no module would ever have matched the name `nil` to
issue this warning in that case.

!no-release-note

### DIFF
--- a/lib/bolt/pal.rb
+++ b/lib/bolt/pal.rb
@@ -141,9 +141,14 @@ module Bolt
       end
     end
 
-    def detect_project_conflict(project, modules)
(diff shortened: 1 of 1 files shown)


nicklewis:
This doesn't work...

PR #2153, opened 2020-09-02
Extract Puppet interaction to a manager service

nicklewis:
This pulls out the code from PAL that handles interacting with Puppet
and building a compiler into a new CompilerService class. That class
builds a compiler context in a dedicated thread and accepts operations
to perform via a work queue, posting results via a promise.

This allows us to only instantiate a single compiler to perform multiple
operations, rather than getting a new one each time we need to use
Puppet.

It also helps separate the code that knows how to interact with Puppet
on a basic level (CompilerService) from the code that knows how to do the particular
operations we care about in Bolt (Bolt::PAL itself).

### DIFF
--- a/lib/bolt/pal.rb
+++ b/lib/bolt/pal.rb
@@ -3,6 +3,7 @@
 require 'bolt/applicator'
 require 'bolt/executor'
 require 'bolt/error'
+require 'bolt/pal/compiler_service'
 require 'bolt/plan_result'
 require 'bolt/util'
 require 'etc'
@@ -52,10 +53,6 @@ module Bolt
 
     def initialize(modulepath, hiera_config, resource_types, max_compiles = Etc.nprocessors,
                    trusted_external = nil, apply_settings = {}, project = nil)
-      # Nothing works without initialized this global state. Reinitializing
-      # is safe and in practice only happens in tests
-      self.class.load_puppet
-
       @user_modulepath = modulepath
       @modulepath = [BOLTLIB_PATH, *modulepath, MODULES_PATH]
       @hiera_config = hiera_config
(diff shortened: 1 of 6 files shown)


lib/bolt/pal.rb
```diff
@@ -70,134 +67,38 @@ def initialize(modulepath, hiera_config, resource_types, max_compiles = Etc.npro
...
-      # The environment modulepath has stripped out non-existent directories,
-      # so we don't need to check for them
-      modules = environment.modulepath.flat_map do |path|
-        Dir.children(path).select { |name| Puppet::Module.is_module_directory?(name, path) }
-      end
-      if modules.include?(project.name)
-        Bolt::Logger.warn_once("project shadows module",
-                               "The project '#{project.name}' shadows an existing module of the same name")
-      end
+      @compiler_service = Bolt::PAL::CompilerService.new(project, @modulepath)
```
McdonaldSeanp, line 70: Is this a chance to refactor the PAL class so the initializer isn't the thing that takes the modulepath, but instead we pass instances of applicator and compiler service?

Something we talked about in slack yesterday with respect to how PE builds on top of Bolt::PAL is that one of the things we need to override in PE::PAL is the Bolt::PAL initializer, and we only need to do that because we need to control the modulepath differently.

In that discussion you mentioned:
> ideally the applicator would probably be an argument to the PAL constructor and then we could avoid this situation entirely

Potentially I feel like that applies here: we can avoid the modulepath/user_modulepath logic if we just have the init method for PAL take instances of CompilerService and Applicator as params.

lib/bolt/pal.rb
```diff
@@ -479,16 +347,20 @@ def generate_types
...
+          compiler.call_function('run_task', task_name, targets, description, params)
+        end
       end
     end
 
-    def run_plan(plan_name, params, executor = nil, inventory = nil, pdb_client = nil, applicator = nil)
-      in_plan_compiler(executor, inventory, pdb_client, applicator) do |compiler|
-        r = compiler.call_function('run_plan', plan_name, params.merge('_bolt_api_call' => true))
-        Bolt::PlanResult.from_pcore(r, 'success')
+    def run_plan(plan_name, params, executor = nil, inventory = nil, pdb_client = nil)
```
McdonaldSeanp, line 358: Removing the ability to pass applicator here will force PE to re-implement this function. We pass the PE version of the applicator to this function, where the PE version implements to correct functionality to pass the apply through Orch.

Refactoring the PAL init method to take applicator would be a way to solve this.

nicklewis: Ahh that makes sense. I couldn't find anywhere in Bolt where we passed an applicator through, so I removed it. I do think making the Applicator an argument to either PAL init or the `with_bolt_executor` method itself would probably make the most sense. That would remove _most_ of the parameters to the PAL constructor, which are primarily there just to be later passed to the Applicator.

McdonaldSeanp: Here's the function in PE where we actually create our PAL and call run_plan: https://github.com/puppetlabs/orchestrator/blob/master/ruby/bolt-plan-executor-lib/bolt_plan_executor/jruby_plan_executor.rb#L44

Hopefully that provides a little context on our use-case. From PE's perspective I don't think it matters too much specifically where the applicator can be passed, as long as 1. we can pass it to a PAL function that's a public part of the PAL class and 2. it's easy to explicitly manage the full modulepath

Issue #2124, opened 2020-08-26
Document Bolt logs


hestonhoffman:
Now that we're creating a log file, and we're working on cleaning up the logs, I think it's time to document logs in more detail. 

This should be a dedicated doc for logging in the configuring Bolt section (perhaps named "Configuring logs"):

The doc should cover:
- The log file: Name. When is it created? How to disable it?
- Log levels: What are the different log levels, what do they log, and when would you want to use each one?
- How + where to configure logs (CLI and config files)

Also add the log file to the table on the projects page and link to this doc from there.

PR #2155, opened 2020-09-03
(GH-2124)(Docs) Logging doc

hestonhoffman:
Adds a doc for Bolt logs.

closes GH-2124

!no-release-note

### DIFF
--- a/documentation/logs.md
+++ b/documentation/logs.md
@@ -1,0 +1,152 @@
+# Logs
+
+Bolt supports multiple log levels. You can configure the log level from the CLI,
+or in a project configuration file. Supported logging levels, in order from most
+to least information logged, are `trace`, `debug`, `info`, `warn`, and `error`.
+
+By default, Bolt logs a debug-level log to a `bolt-debug.log` file in the root
+of your project directory. You can also configure custom logs from your project
+configuration file.
+
+## The `bolt-debug.log` file
+ 
+Each time you run a Bolt command, Bolt prints a debug-level log to a
+`bolt-debug.log` file in the root of your project directory.
+
+You can disable the log file by specifying the following in your
+`bolt-project.yaml` file:
+
+```yaml
+log:
+  bolt-debug.log: disable
+```
+
+## Setting log level
+
+You can set Bolt's log level from the CLI or use a
+configuration file.
+
+### Setting log level on the CLI
+
+To set the log level from the CLI, use the `--log-level` option along with the
+desired level. Available log levels are `trace`, `debug`, `info`, `warn`, and
+`error`. For example:
+
+```console
+bolt command run whoami -t target1 --log-level trace
+```
+
+### Setting log level in a configuration file
+
+To set the log level for the console, add a `log` map with a `console` mapping
+to your [project configuration file](configuring_bolt.md#project-level-configuration).
+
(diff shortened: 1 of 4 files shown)


documentation/logs.md
```diff
@@ -0,0 +1,152 @@
+# Logs
+
+Bolt supports multiple log levels. You can configure the log level from the CLI,
+or in a project configuration file. Supported logging levels, in order from most
+to least information logged, are `trace`, `debug`, `info`, `warn`, and `error`.
```
lucywyman, line 5: Technically speaking `fatal` is also a valid log level, but it's only useful if you only want to see `emerg` and `critical` messages from a Puppet manifest or apply block. We don't actually use it in Bolt code. I'm not sure if that means we should include it or not - I think it's fine to leave it out.

documentation/logs.md
```diff
@@ -0,0 +1,152 @@
...
+> permission to write to it.
+
+## Log levels
+
+### `trace`
+
+Trace logs contain the most detailed information for actions run on individual
+targets. For example, the logs include each command run on a target, each step
+of compiling a manifest for an apply block, and information about establishing
+connections with targets.
```
lucywyman, line 86: I'm not sure readers will find this distinction useful, but the team thinks of trace as "why Bolt is doing what it's doing" while debug is just "what Bolt's doing"

documentation/logs.md
```diff
@@ -0,0 +1,152 @@
...
+
+### `warn`
+
+Warn logs contain warnings about deprecations and potentially harmful situations
+that might affect your Bolt run, even though they don't prevent Bolt from
+executing. For example, Bolt warns you if your inventory file includes an
+unsupported transport configuration that might result in Bolt connecting to
+targets in a way that you don’t expect.
+
+Bolt always prints warnings to the console.
```
lucywyman, line 123:
```suggestion
Bolt prints warnings to the console by default.
```

documentation/logs.md
```diff
@@ -0,0 +1,152 @@
...
+Error logs contain messages for errors that Bolt encountered during execution.
+For example, Bolt prints an error if a PuppetDB query fails.
+
+Bolt always prints errors to the console.
+
+Use an error log if you want to see errors that Bolt raised during a Bolt run.
+
+## Using `verbose` logging
+
+Bolt's `run`, `file upload`, and `file download` commands include the
```
lucywyman, line 140:
```suggestion
Bolt's `(command | plan | script | task) run` and `file (upload | dowload)` commands include the
```

lucywyman: The original wording just made it sound like there was a `bolt run` command - I'm not sure if this is clearer or not, but some other syntax might be more useufl.

hestonhoffman: Okay, let me take another swing at this

documentation/logs.md
hestonhoffman: Yeah, Tom mentioned that we don't use it, so I opted to leave it out. The fact that it will return stuff from a puppet manifest makes me wonder if we should include it though.... Maybe I'll add a note with a link to the puppet docs for that... I'll do some digging and figure that out.

hestonhoffman: Okay, I've added in a section for fatal logs

lucywyman (approved):

PR #2157, opened 2020-09-04
(PE-30023) Add additional "allowed" metadata key for project

donoghuc:
Include whether plans/task from a bolt-project have been allowlisted in bolt-project.yaml

### DIFF
--- a/developer-docs/bolt-api-servers.md
+++ b/developer-docs/bolt-api-servers.md
@@ -516,10 +516,12 @@ This returns a JSON object of this shape:
 ```
 [
   {
-    "name": "facts"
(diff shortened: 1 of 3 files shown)


steveax (approved):
👍

nicklewis:
I don't think "allowed" captures the meaning of the tasks/plans list in the project. The context is more that those are the tasks and plans which are enumerated as belonging to the UI presented by the project. In particular, nothing about not being on that list makes a task or plan _disallowed_.

donoghuc:
In the context of PE this bit of information is actually part of permissions checks, so for this case there is more information than just "display or not".

PR #2158, opened 2020-09-04
(maint) Only require release notes tag for bolt source code

donoghuc:
This commit updates the commit message tag check to only run when the commit touches a file that contains a change to bolt source code. This excludes files in tests, bolt-server source, and other files that where changes will not need to be added to the changelog.


donoghuc:
Did this really just fail bc there was a change in `.github` :rofl: :rofl: :rofl: touche github.

PR #2159, opened 2020-09-04
(GH-2156) Don't leak scope between targets

### LINKED ISSUE #2156: bolt apply applies a catalog from another target caused by race condition in compile-concurrency code
## Describe the Bug
I have ~500 nodes (transport SSH) migrated from ansible to puppet bolt.
From begin we have a strange behavior during bolt apply -t all.
A node got a catalog from another node. Not persistent, randomly. Some nodes got wrong configurations and lost SSH service.

Some research shows a race condition in /opt/puppetlabs/bolt/lib/ruby/gems/2.5.0/gems/bolt-2.26.0/lib/bolt/applicator.rb compile function. Variable **scope** from method **apply_ast** has been shared to all pool threads

```ruby
module Bolt
  class Applicator
...

def apply_ast(raw_ast, targets, options, plan_vars = {})
  ...
      scope = {
         ...
      }
  ...
      r = @executor.log_action(description, targets) do
        futures = targets.map do |target|
          Concurrent::Future.execute(executor: @pool) do
            Thread.current[:name] ||= Thread.current.name
            @executor.with_node_logging("Compiling manifest block", [target]) do
              compile(target, scope) # Ruby shares variable scope between all threads in pool
...

    def compile(target, catalog_input)
    ...
      catalog_input[:target] = {
        name: target.name,
        facts: @inventory.facts(target).merge('bolt' => true),
        variables: @inventory.vars(target),
        trusted: trusted.to_h
      }
      ...
      # catalog_input can be modified from another thread
      ...
      out, err, stat = Open3.capture3('ruby', bolt_catalog_exe, 'compile', stdin_data: catalog_input.to_json)
```
So "'ruby', bolt_catalog_exe, 'compile'" can get  catalog_input[:target] from another thread from another target.

## Steps to Reproduce
```bash
$ cat inventory.yaml
config:
  transport: local
targets:
  - test1
  - test2
  - test3
  - test4
  - test5
  - test6
  - test7
  - test8
  - test9
$ sudo patch /opt/puppetlabs/bolt/lib/ruby/gems/2.5.0/gems/bolt-2.26.0/lib/bolt/applicator.rb <<EOF
--- applicator.rb-
+++ applicator.rb
@@ -88,11 +88,14 @@
         trusted: trusted.to_h
       }

+name_before = catalog_input[:target][:name]
       bolt_catalog_exe = File.join(libexec, 'bolt_catalog')
       old_path = ENV['PATH']
       ENV['PATH'] = "#{RbConfig::CONFIG['bindir']}#{File::PATH_SEPARATOR}#{old_path}"
       out, err, stat = Open3.capture3('ruby', bolt_catalog_exe, 'compile', stdin_data: catalog_input.to_json)
       ENV['PATH'] = old_path
+name_after = catalog_input[:target][:name]
+STDERR.puts "before:#{name_before}, after:#{name_after}" if name_before != name_after

       # If bolt_catalog does not return valid JSON, we should print stderr to
       # see what happened
EOF
$ bolt apply -e '' --compile-concurrency 3 --noop -t all | fgrep before
before:test3, after:test1
before:test1, after:test4
before:test2, after:test5
before:test4, after:test7
before:test5, after:test8
before:test8, after:test9
before:test7, after:test9
```

As a fast dirty fix can be used this patch:
```
--- applicator.rb-
+++ applicator.rb
@@ -218,7 +218,7 @@
           Concurrent::Future.execute(executor: @pool) do
             Thread.current[:name] ||= Thread.current.name
             @executor.with_node_logging("Compiling manifest block", [target]) do
-              compile(target, scope)
+              compile(target, scope.dup)
             end
           end
         end
```

## Environment
 - puppet-bolt 2.26.0
 - CentOS 8, Fedora 32

nicklewis:
When compiling catalogs, we use a single scope object to capture the
current plan scope at the time of the compilation, and then merge in
target-specific information for each compile. However, the compile
method was *modifying* the single scope object to add its
target-specific info. This caused a race condition whereby the scope
could be modified by another thread and the wrong target inserted and
used for compilation.

This appears to be relatively uncommon at low levels of concurrency,
because the critical section only lasts until the scope has been
serialized to JSON to be passed to the compilation process. Once that's
happened, any external modifications won't affect the current compile.
Due to the GIL, it's impossible for multiple threads to be running that
code at the same time, so the race condition requires a context switch
to happen in the middle of execution. With high concurrency or many
targets being compiled, this is not unlikely.

We now _merge_ the target-specific data rather than inserting it
directly, so each target gets a copy of the scope. We also `freeze` the
scope for good measure, to avoid any such issues popping up in the
future. This does not have a significant impact on memory usage as each
scope hash still refers to a single copy of the actual underlying
_data_.

Fixes #2156

!bug

  * **Prevent data for the wrong target being used when compiling `apply()` blocks**
     ([#2156](https://github.com/puppetlabs/bolt/issues/2156))

    Previously, a race condition allowed `apply()` blocks to use data
    for the wrong target during compilation. This could cause targets to
    apply incorrect or invalid catalogs.

### DIFF
--- a/lib/bolt/applicator.rb
+++ b/lib/bolt/applicator.rb
@@ -75,23 +75,24 @@ module Bolt
       end
     end
 
-    def compile(target, catalog_input)
+    def compile(target, scope)
       # This simplified Puppet node object is what .local uses to determine the
       # certname of the target
       node = Puppet::Node.from_data_hash('name' => target.name,
                                          'parameters' => { 'clientcert' => target.name })
       trusted = Puppet::Context::TrustedInformation.local(node)
-      catalog_input[:target] = {
+      target_data = {
         name: target.name,
         facts: @inventory.facts(target).merge('bolt' => true),
         variables: @inventory.vars(target),
         trusted: trusted.to_h
       }
+      catalog_request = scope.merge(target: target_data)
 
       bolt_catalog_exe = File.join(libexec, 'bolt_catalog')
       old_path = ENV['PATH']
       ENV['PATH'] = "#{RbConfig::CONFIG['bindir']}#{File::PATH_SEPARATOR}#{old_path}"
-      out, err, stat = Open3.capture3('ruby', bolt_catalog_exe, 'compile', stdin_data: catalog_input.to_json)
+      out, err, stat = Open3.capture3('ruby', bolt_catalog_exe, 'compile', stdin_data: catalog_request.to_json)
       ENV['PATH'] = old_path
 
       # If bolt_catalog does not return valid JSON, we should print stderr to
@@ -192,7 +193,7 @@ module Bolt
         plan_vars: plan_vars,
         # This data isn't available on the target config hash
         config: @inventory.transport_data_get
-      }
+      }.freeze
       description = options[:description] || 'apply catalog'
 
       required_modules = options[:required_modules].nil? ? nil : Array(options[:required_modules])


nicklewis:
I'm thinking about how to effectively test this in a way that isn't `it should not have this specific bug`...

donoghuc (approved):

adreyer (approved):

PR #2161, opened 2020-09-10
(PUP-10653) Remove dependency on win32/dir

GabrielNagy:
For Puppet 7 we are dropping the `win32/dir` dependency as we only used
constants from it, which we replaced with environment variables (see:
https://github.com/puppetlabs/puppet/pull/8314).

This would become breaking in bolt when it switches to Puppet 7 and the
win32 gems are no longer pulled.

For AIO bolt, the Puppet change won't break anything since bolt uses its
own runtime which still packages the win32 gems, but it would still be
an improvement in speed if the dependency is dropped.

The `win32/dir` gem also monkey patches some `Dir` methods, most notably
`glob` and `pwd` to use backslashes which end up causing more trouble as
Ruby uses `/` as the default separator for Windows.

Switch to using the `ALLUSERSPROFILE` environment variable instead of
the `Dir::COMMON_APPDATA` constant, and remove a `Dir.pwd` workaround.

### DIFF
--- a/lib/bolt/config.rb
+++ b/lib/bolt/config.rb
@@ -97,11 +97,8 @@ module Bolt
     end
 
     def self.system_path
-      # Lazy-load expensive gem code
-      require 'win32/dir' if Bolt::Util.windows?
-
       if Bolt::Util.windows?
-        Pathname.new(File.join(Dir::COMMON_APPDATA, 'PuppetLabs', 'bolt', 'etc'))
+        Pathname.new(File.join(ENV['ALLUSERSPROFILE'], 'PuppetLabs', 'bolt', 'etc'))
       else
         Pathname.new(File.join('/etc', 'puppetlabs', 'bolt'))
       end
--- a/lib/bolt/puppetdb/config.rb
+++ b/lib/bolt/puppetdb/config.rb
@@ -18,8 +18,7 @@ module Bolt
       end
 
       def self.default_windows_config
-        require 'win32/dir'
-        File.expand_path(File.join(Dir::COMMON_APPDATA, 'PuppetLabs/client-tools/puppetdb.conf'))
+        File.expand_path(File.join(ENV['ALLUSERSPROFILE'], 'PuppetLabs/client-tools/puppetdb.conf'))
       end
 
       def self.load_config(options, project_path = nil)


GabrielNagy:
Hmm, I guess `win32/dir` is still being loaded through puppet, so I'll leave the `Dir.pwd` as is for now.

lucywyman:
This looks great @GabrielNagy ! Since we're pinned to Puppet versions, I'm just gonna give this a Puppet 7 label and we'll merge it when we update.

GabrielNagy:
Ah cool @lucywyman. I'll remove this `Dir.pwd` workaround here then to avoid confusion: https://github.com/puppetlabs/bolt/blob/4742cf14e9a9841bb2a994bfc9e6ca9b92437365/lib/bolt/cli.rb#L794

Just a heads up that PR testing will fail until puppet 7, as puppet 6 still loads `win32/dir` even if not explicitly required in bolt. Just ping me whenever and I'll rebase the PR 😄.

lib/bolt/cli.rb
```diff
@@ -787,11 +787,7 @@ def generate_types
     # Initializes a specified directory as a Bolt project and installs any modules
     # specified by the user, along with their dependencies
     def initialize_project
-      # Dir.pwd will return backslashes on Windows, but Pathname always uses
-      # forward slashes to concatenate paths. This results in paths like
-      # C:\User\Administrator/modules, which fail module install. This ensure
-      # forward slashes in the cwd path.
-      dir = File.expand_path(Dir.pwd)
+      dir = Dir.pwd
```
lucywyman, line 790: @GabrielNagy Sorry just coming back to this to prep for Puppet 7 release. I'm not sure I follow why this change is needed? I think it may be causing the failure in the Windows tests in GH actions.

GabrielNagy: `Dir.pwd` is monkey-patched by the `win32/dir` gem to return a backslash-separated path. Even though the gem is not required here, it could be that it's required through puppet 6 here https://github.com/puppetlabs/puppet/blob/master/lib/puppet/util/windows.rb#L30.

With puppet 7 `Dir.pwd` uses forward slashes, as we removed all references to `win32/dir`, so this workaround here is not needed anymore.

I assumed the bolt version is tightly coupled with the puppet version due to the dependency being somewhat strict (https://github.com/puppetlabs/bolt/blob/main/bolt.gemspec#L51). After bumping puppet to 7.0.0 in the gemspec, PR tests should no longer fail.

I can revert this hunk so the code will work with both puppet 6 and 7. Let me know your thoughts.

lucywyman: Hm, it sounds like we aren't planning to pick up Puppet 7 in Bolt proper until our next major version, which we're planning for early next year. Once we pick up Puppet 7 though we won't need to support puppet 6, since we ship the Puppet that Bolt uses with Bolt. Long way of saying we should leave it, sorry for the hassle!

GabrielNagy: No worries, left it in and rebased. Seems that PR tests are passing now, so this should be good to merge.

GabrielNagy:
@lucywyman fwiw this can be merged anytime as it's a backwards-compatible replacement. We did the same for the stdlib module and p-e-m.

lucywyman:
Oh awesome - sorry, I didn't realize that!

PR #2162, opened 2020-09-10
(GH-1835) Document plan_hierarchy usage

### LINKED ISSUE #1835: Don't fail when looking up data outside an apply block from a hiera.yaml that includes interpolations

lucywyman:
Previously, Bolt would use the same Hiera hierarchy for lookups both
inside apply blocks and outside apply blocks, and simply removed facts
and vars from the PAL ScriptCompiler to prevent users from using
interpolations in their hierarchies for lookups outside apply blocks.
This meant that hierarchies that used interpolations would error even if
the user wasn't looking up data that included an interpolation. To
mitigate this, we added a separate hierarchy under the key
`plan_hierarchy` in puppetlabs/puppet#8232.

This documents using the plan_hierarchy Hiera config key, including a
short example.

Closes #1835

!feature

* **Define Hiera data to be looked up outside apply blocks under plan_hierarchy key** ([#1835](https://github.com/puppetlabs/bolt/issues/1835))

  Previously, Bolt used the same Hiera hierarchy for lookups inside and
  outside apply blocks. Interpolations are only supported in apply blocks,
  so if a hierarchy included interpolations the user could not look up
  data outside an apply block. Users can now define a separate, statically
  configured hierarchy in their Hiera config to be used outside apply
  blocks.

### DIFF
--- a/documentation/applying_manifest_blocks.md
+++ b/documentation/applying_manifest_blocks.md
@@ -252,6 +252,44 @@ If a custom data provider is used, such as `hiera-eyaml`, which allows you to
 encrypt your data, the gem dependencies must be available to Bolt. See [Install
 gems with Bolt packages](bolt_installing.md#).
 
+[Hiera interpolations](https://puppet.com/docs/puppet/latest/hiera_merging.html#hiera_interpolation)
+are not supported in Bolt outside of apply blocks, because Hiera data is looked up per-target and
+plans do not run in a per-target context. If you want to use Hiera to look up data outside of an apply
+block and have an existing Hiera config that contains interpolations, you can add the data you want
+to use outside of the apply block to the top-level `plan_hierarchy` key. If present, Bolt uses the data 
+under the `plan_hierarchy` key outside of apply blocks, and uses the the typical hierarchy key inside 
+of apply blocks. This allows you to use your existing Hiera configs in Bolt plans without encountering 
+an error if interpolation exists and your plan tries to look up data outside of an apply block.
+
+For example, with this Hiera configuration at `<project>/hiera.yaml`:
+```yaml
+version: 5
+
+hierarchy:
+  - name: "Per-OS defaults"
+    path: "os/%{facts.os.family}.yaml"
+  - name: Common
+    path: hierarchy.yaml
+
+plan_hierarchy:
+  - name: Common
+    path: plan_hierarchy.yaml
+```
+
+Bolt uses the `plan_hierarchy` data for the first `lookup()` call, and the regular `hierarchy`
+data for the second `lookup()` call:
+```
+plan plan_lookup(
+  TargetSpec $targets
+) {
+  $outside_apply = lookup('pop')
+  $in_apply = apply($targets) {
+    notify { lookup('pop'): }
+  }
+  ...
+}
+```
+
 📖 **Related information**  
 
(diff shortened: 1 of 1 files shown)


beechtom (approved):

documentation/applying_manifest_blocks.md
```diff
@@ -252,6 +252,44 @@ If a custom data provider is used, such as `hiera-eyaml`, which allows you to
...
 gems with Bolt packages](bolt_installing.md#).
 
+[Hiera interpolations](https://puppet.com/docs/puppet/latest/hiera_merging.html#hiera_interpolation)
+are not supported in Bolt outside of apply blocks, since Hiera data is looked up per-target but
+plans do not run in a per-target context. If you want to use Hiera to look up data outside of apply
+block and have an existing Hiera config that contains interpolations, you can add the data you want
+to use outside of apply blocks to the top-level `plan_hierarchy` key. If present, the data under
+this key will be used outside of apply blocks, while the typical hierarchy key will be used inside
+apply blocks. This allows users to use their existing Hiera configs in Bolt plans without erroring
+if any interpolation exists and they look up data outside an apply block.
```
hestonhoffman, line 262:
```suggestion
[Hiera interpolations](https://puppet.com/docs/puppet/latest/hiera_merging.html#hiera_interpolation)
are not supported in Bolt outside of apply blocks, because Hiera data is looked up per-target and
plans do not run in a per-target context. If you want to use Hiera to look up data outside of an apply
block and have an existing Hiera config that contains interpolations, you can add the data you want
to use outside of the apply block to the top-level `plan_hierarchy` key. If present, Bolt uses the data 
under the `plan_hierarchy` key outside of apply blocks, and uses the the typical hierarchy key inside 
of apply blocks. This allows you to use your existing Hiera configs in Bolt plans without encountering 
an error if interpolation exists and your plan tries to look up data outside of an apply block.
```

documentation/applying_manifest_blocks.md
```diff
@@ -252,6 +252,44 @@ If a custom data provider is used, such as `hiera-eyaml`, which allows you to
...
+  - name: Common
+    path: hierarchy.yaml
+
+plan_hierarchy:
+  - name: Common
+    path: plan_hierarchy.yaml
+```
+
+The `plan_hierarchy` data will be used for the first `lookup()` call, and the regular `hierarchy`
+data will be used for the second `lookup()` call:
```
hestonhoffman, line 280:
```suggestion
Bolt uses the `plan_hierarchy` data for the first `lookup()` call, and the regular `hierarchy`
data for the second `lookup()` call:
```

hestonhoffman (approved):

PR #2163, opened 2020-09-10
(GH-2131) Add bolt module subcommands

### LINKED ISSUE #2131: Copy bolt puppetfile to bolt module
## Use Case
- Rename `bolt puppetfile show-modules` to `bolt module show`.
- Rename `bolt puppetfile generate-types` to `bolt module generate-types`
We should not deprecate `bolt puppetfile *` yet, we can do so closer to Bolt 3.0.

This should still show modules from the Puppetfile, not from the `modules` data in bolt-project.yaml.
Behavior should be the same as `bolt puppetfile`.

lucywyman:
This adds the `bolt module generate-types` and `bolt module show`
subcommands, which are analagous to `bolt puppetfile generate-types` and
`bolt puppetfile show-modules` respectively. These commands have the
exact same behavior as the puppetfile subcommands. The environment
variable `BOLT_MODULE_FEATURE` must be set to use the `bolt module`
subcommand. The puppetfile commands aren't deprecated yet - we'll do
that closer to 3.0. This also doesn't include a release note, as we're
launching the module feature dark until the entire workflow is complete.

Closes #2131

!no-release-note

### DIFF
--- a/lib/bolt/bolt_option_parser.rb
+++ b/lib/bolt/bolt_option_parser.rb
@@ -69,6 +69,12 @@ module Bolt
         when 'install'
           { flags: OPTIONS[:global] + %w[configfile force project],
             banner: MODULE_INSTALL_HELP }
+        when 'show'
+          { flags: OPTIONS[:global] + OPTIONS[:global_config_setters],
+            banner: MODULE_SHOW_HELP }
+        when 'generate-types'
+          { flags: OPTIONS[:global] + OPTIONS[:global_config_setters],
+            banner: MODULE_GENERATETYPES_HELP }
         else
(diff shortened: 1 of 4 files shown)


beechtom (changes requested):
This should include the help text for the new commands in `bolt_option_parser` so it isn't missed when the workflow goes live.

beechtom (changes requested):
`show` and `generate-types` need to be added as actions in the `module` help text: https://github.com/puppetlabs/bolt/blob/main/lib/bolt/bolt_option_parser.rb#L363

beechtom (approved):

PR #2165, opened 2020-09-11
(GH-2148) Fix JSON ordering in pester test

### LINKED ISSUE #2148: Invoke-BoltTask -Params returns json in different order than given
## Describe the Bug

When using a powershell hash as a value for the `params` parameter in invoke-bolttask, the order of the json hash given to bolt is different.

## Expected Behavior

Invoke-BoltTask should return the json version of the powershell hash in the original order given

## Steps to Reproduce

https://github.com/puppetlabs/bolt/pull/2147/checks?check_run_id=1057739739

```
Starting discovery in 2 files.
Discovery finished in 3.74s.
[+] D:\a\bolt\bolt\pwsh_module\autogenerated.tests.ps1 20.99s (6.56s|10.85s)
WARNING: The module manifest member 'ModuleToProcess' has been deprecated. Use the 'RootModule' member instead.
WARNING: this works but order is not same
WARNING: this works but order is not same
WARNING: this works but is not 100%
WARNING: requires params to not be positionl...is that a problem
WARNING: requires params to not be positionl...is that a problem
WARNING: Verify this
[-] test all bolt command examples.bolt task.bolt task run package --targets target1,target2 action=status name=bash 91ms (88ms|3ms)
 Expected strings to be the same, but they were different.
 String lengths are both 96.
 Strings differ at index 64.
 Expected: '...s '{"name"...'
 But was:  '...s '{"actio...'
 at $results | Should -Be "bolt task run 'package' --targets 'target1,target2' --params '{`"name`":`"bash`",`"action`":`"status`"}'", D:\a\bolt\bolt\pwsh_module\command.tests.ps1:267
 at <ScriptBlock>, D:\a\bolt\bolt\pwsh_module\command.tests.ps1:267
Tests completed in 23.21s
Tests Passed: 1112, Failed: 1, Skipped: 0 NotRun: 0
##[error]Process completed with exit code 1.
```

## Environment
 - Version [2.24.1]
 - Platform [Windows]

lucywyman:
The ConvertTo-JSON cmdlet respects JSON ordering in Powershell 5.1 and
7.1, but not version 7.0. This caused the test that verifies passing a
JSON object to Invoke-BoltTask to fail, as the ordering of JSON
parameters was wrong and the test simply verifies the resulting Bolt
command as a string. This updates the test to accept the JSON object
elements in any order. Because users and their tasks don't care about
the order, it doesn't make sense to fix this in the code itself - just
to update the test to also not care about the order.

Closes #2148

!no-release-note

### DIFF
--- a/pwsh_module/command.tests.ps1
+++ b/pwsh_module/command.tests.ps1
@@ -263,9 +263,12 @@ Describe "test all bolt command examples" {
       $results = Invoke-BoltTask -name 'package' -targets 'target1,target2' -params '{"name":"bash","action":"status"}'
       $results | Should -Be "bolt task run 'package' --targets 'target1,target2' --params '{`"name`":`"bash`",`"action`":`"status`"}'"
 
-      # Addressing in https://github.com/puppetlabs/bolt/issues/2148
-      # $results = Invoke-BoltTask -name 'package' -targets 'target1,target2' -params @{ 'name' = 'bash'; 'action' = 'status' }
-      # $results | Should -Be "bolt task run 'package' --targets 'target1,target2' --params '{`"name`":`"bash`",`"action`":`"status`"}'"
+      $results = Invoke-BoltTask -name 'package' -targets 'target1,target2' -params @{ 'name' = 'bash'; 'action' = 'status' }
+      # We don't care about the order of JSON keys, and they might become out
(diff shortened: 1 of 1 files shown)


jpogran (approved):
Neat use of pester and array checking

PR #2170, opened 2020-09-14
(GH-2168) Add version requirements for module declarations

### LINKED ISSUE #2168: Support version requirements for module declarations
## Use Case

When specifying modules in the project configuration file, a user might want to install a specific version of a module or restrict a module to a version range.

Module declarations under the `modules` key in a project configuration file should accept a `version_requirement` key. This key should accept any version that is considered valid by semantic-puppet, including specific versions (`1.0.0`), version shorthand (`1.x`), and version ranges (`>= 1.0.0 < 2.0.0`).

## Additional Context

This work requires `puppetfile-resolver >= 0.4.0`, as earlier versions of the gem included a bug that did not allow version ranges.

beechtom:
This adds support for specifying version requirements for a module
declaration under the `modules` key of a project config file. To specify
a version, the module declaration should have a `version_requirement`
key. Version requirements should be valid according to semantic-puppet,
and includes support for specific versions (`1.0.0`), version shorthand
(`1.x`), and version ranges (`>= 1.0.0 < 3.0.0`).

Depends on https://github.com/puppetlabs/puppet-runtime/pull/371
Closes #2168 

!no-release-note

### DIFF
--- a/lib/bolt/config/options.rb
+++ b/lib/bolt/config/options.rb
@@ -239,11 +239,21 @@ module Bolt
               "name" => {
                 description: "The name of the module.",
                 type: String
+              },
+              "version_requirement" => {
+                description: "The version requirement for the module. Accepts a specific version (1.2.3), version "\
+                             "shorthand (1.2.x), or a version range (>= 1.2.0).",
+                type: String
               }
             }
           },
           _plugin: false,
-          _example: [{ "name" => "puppetlabs-mysql" }, { "name" => "puppetlabs-apache" }]
+          _example: [
+            { "name" => "puppetlabs-mysql" },
+            { "name" => "puppetlabs-apache", "version_requirement" => "5.5.0" },
+            { "name" => "puppetlabs-puppetdb", "version_requirement" => "7.x" },
+            { "name" => "puppetlabs-firewall", "version_requirement" => ">= 1.0.0 < 3.0.0" }
+          ]
         },
         "name" => {
           description: "The name of the Bolt project. When this option is configured, the project is considered a "\
--- a/lib/bolt/puppetfile/module.rb
+++ b/lib/bolt/puppetfile/module.rb
@@ -30,7 +30,7 @@ module Bolt
           raise Bolt::ValidationError, "Module name #{mod['name']} must include both the owner and module name."
         end
 
-        new(owner, name)
+        new(owner, name, mod['version_requirement'])
       end
 
       # Returns the module's title.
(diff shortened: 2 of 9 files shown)


nicklewis:
This doesn't play well with the step where we check whether the existing Puppetfile already includes the modules you need. I think we also need to check whether that Puppetfile satisfies the `version_requirements` you've specified.

beechtom:
Updated to check if versions intersect.

lib/bolt/config/options.rb
```diff
@@ -239,11 +239,21 @@ module Options
               "name" => {
                 description: "The name of the module.",
                 type: String
+              },
+              "version_requirement" => {
+                description: "The version requirement for the module. Accepts either a specific version, version "\
```
nicklewis, line 244: It might be helpful to have examples of those formats.

`Accepts a specific version (1.2.3), version shorthand (1.2.x), or a version range (>= 1.2.0)`

lib/bolt/puppetfile/module.rb
```diff
@@ -42,16 +42,39 @@ def title
       # Checks two modules for equality.
       #
       def eql?(other)
-        self.class == other.class && @owner == other.owner && @name == other.name
+        self.class == other.class &&
+          @owner == other.owner &&
+          @name == other.name &&
+          versions_intersect?(other)
```
nicklewis, line 48: What do we use `eql?` for? It's weird to say that two modules with version _range_ specifications are equal just because there is overlap between their ranges. But how weird it is depends on the context in which it's used.

nicklewis (commented):
I'm a little concerned about the fact that we're conflating the concepts of version range specifications with actual resolved version, but I don't have a good solution for it. We might want to represent the content of `modules` in bolt-project.yaml with a `ModuleDependency` object or something, separate from a `Module` that exists in a resolved Puppetfile (and which therefore has a specific version). We can always do that later if needed though.

lib/bolt/puppetfile/module.rb
beechtom: It's used by `Set` to determine whether the object is already part of a set / for comparison of sets. At least that's my understanding.

beechtom:
Ya, I'm not completely happy with the `Module` class at the moment. I think it works for this first iteration, but as soon as we add support for git modules I think it'll need some reworking...

lib/bolt/puppetfile/module.rb
beechtom: Specifically, checking if the versions intersect allows `existing_puppetfile.modules.superset? puppetfile.modules` to check if versions for each module in the Puppetfile fall within the version requirement specified in project config.

beechtom: An alternative I was looking at would remove this logic in favor of a `Bolt::Puppetfile#superset?` (or similar) method, which would compare the modules in two Puppetfiles directly, instead of relying on a comparison of sets. That would probably give some more flexibility with checking if modules in a Puppetfile conform to the version requirements without conflating the two?

nicklewis: If we had a ModuleDependency class separate from Module, then something like `Bolt::Puppetfile.satisfies?([ModuleDependency, ModuleDependency, ModuleDependency])` would make sense.

nicklewis (approved):

Issue #2134, opened 2020-08-27
Add bolt module add command


lucywyman:
## Use Case
"I want to add a module to my project and install it"

- [x] This adds a `bolt module add <module>` command that will add the module to the user's `modules` key in `bolt-project.yaml`. The command only accepts one module.
- [x] If none are provided, we should error.
- [x] If multiple modules are provided, we should error.
- [x] ~If the project has a bolt.yaml we should error and tell you to run bolt project migrate~ Only error if there is no bolt-project.yaml
- [x] If we don't find a bolt-project.yaml we should error and tell you to run bolt project init
- [x] If the `bolt-project.yaml` doesn't have a `modules` key we should create one with the specified module under it.
- [x] If `bolt-project.yaml` can't be written to we should error.
- [x] If the module is already specified in `bolt-project.yaml` we should print a nice message and do nothing.

- [x] If `bolt-project.yaml` has a modules key but does not have the specified module listed, we should add it as just `{ name: <name> }`.
- [x] We should validate that the module exists and can be installed before adding it if possible. If it does not exist or cannot be added for some reason we should error. If this ends up being hard or impossible we can just add it to modules and skip this.

## Questions
- Is there other metadata we should automatically add for the module? Not for now, since there's no other metadata.
- Should this cause other modules to be resolved + upgraded, not just the one that was added? We should try pinning everything that's already installed and install the new thing. If that fails, we should upgrade existing modules.

PR #2172, opened 2020-09-14
(GH-2134) Add 'bolt module add' command

beechtom:
This adds a new `bolt module add` command that adds a new module to the
project. The command accepts a single module name, resolves its
dependencies, updates the project configuration file and `Puppetfile`,
and installs the project's modules.

If the project configuration file already includes a declaration for the
new module, Bolt will display a helpful message and exit successfully
without doing anything.

When resolving modules, Bolt will first attempt to resolve the new
module with each of the installed modules pinned to a specific version.
This is intended to prevent any installed modules from being updated
unintentionally. If resolving with this approach fails, Bolt falls back
to re-resolving all of the module declarations with the new module,
which may result in insalled modules being updated to different
versions.

Lastly, this includes a new `ModuleInstaller` class which moves most of
the logic for installing modules from the `CLI` class.

This feature is shipping dark and does not have an associated release
note.

!no-release-note

### DIFF
--- a/lib/bolt/bolt_option_parser.rb
+++ b/lib/bolt/bolt_option_parser.rb
@@ -66,15 +66,18 @@ module Bolt
           banner: GUIDE_HELP }
       when 'module'
         case action
+        when 'add'
+          { flags: OPTIONS[:global] + %w[configfile force project],
+            banner: MODULE_ADD_HELP }
+        when 'generate-types'
+          { flags: OPTIONS[:global] + OPTIONS[:global_config_setters],
+            banner: MODULE_GENERATETYPES_HELP }
         when 'install'
           { flags: OPTIONS[:global] + %w[configfile force project],
             banner: MODULE_INSTALL_HELP }
         when 'show'
           { flags: OPTIONS[:global] + OPTIONS[:global_config_setters],
             banner: MODULE_SHOW_HELP }
-        when 'generate-types'
-          { flags: OPTIONS[:global] + OPTIONS[:global_config_setters],
-            banner: MODULE_GENERATETYPES_HELP }
         else
           { flags: OPTIONS[:global],
             banner: MODULE_HELP }
@@ -364,12 +367,40 @@ module Bolt
           bolt module <action> [options]
 
       DESCRIPTION
-          Install and list modules and generate type references
+          Manage Bolt project modules
 
       ACTIONS
+          add                   Add a module to the project
           generate-types        Generate type references to register in plans
           install               Install the project's modules
           show                  List modules available to the Bolt project
+    HELP
+
+    MODULE_ADD_HELP = <<~HELP
+      NAME
+          add
+      
+      USAGE
+          bolt module add <module> [options]
+
+      DESCRIPTION
+          Add a module to the project.
+
+          Adds a module declaration to the project's configuration file.
+          Bolt will automatically resolve the new module's dependencies,
+          update any existing project modules if necessary, generate a
+          new Puppetfile, and install the modules.
+    HELP
+
+    MODULE_GENERATETYPES_HELP = <<~HELP
+      NAME
+          generate-types
+
+      USAGE
+          bolt module generate-types [options]
+
+      DESCRIPTION
+          Generate type references to register in plans.
     HELP
 
     MODULE_INSTALL_HELP = <<~HELP
@@ -385,17 +416,6 @@ module Bolt
           Module declarations are loaded from the project's configuration
           file. Bolt will automatically resolve all module dependencies,
           generate a Puppetfile, and install the modules.
-    HELP
-
-    MODULE_GENERATETYPES_HELP = <<~HELP
-      NAME
-          generate-types
-
-      USAGE
-          bolt module generate-types [options]
-
-      DESCRIPTION
-          Generate type references to register in plans.
     HELP
 
     MODULE_SHOW_HELP = <<~HELP
--- a/lib/bolt/cli.rb
+++ b/lib/bolt/cli.rb
@@ -25,6 +25,7 @@ require 'bolt/pal'
 require 'bolt/target'
 require 'bolt/version'
 require 'bolt/secret'
+require 'bolt/module_installer'
 
 module Bolt
   class CLIExit < StandardError; end
@@ -41,7 +42,7 @@ module Bolt
                  'project' => %w[init migrate],
                  'apply' => %w[],
                  'guide' => %w[],
-                 'module' => %w[install show generate-types] }.freeze
+                 'module' => %w[add generate-types install show] }.freeze
 
     attr_reader :config, :options
 
@@ -291,6 +292,10 @@ module Bolt
         raise Bolt::CLIError, "Must specify a plan name."
       end
 
+      if options[:subcommand] == 'module' && options[:action] == 'add' && !options[:object]
+        raise Bolt::CLIError, "Must specify a module name."
+      end
+
       if options.key?(:debug) && options.key?(:log)
         raise Bolt::CLIError, "Only one of '--debug' or '--log-level' may be specified"
       end
@@ -452,8 +457,10 @@ module Bolt
         end
       when 'module'
         case options[:action]
+        when 'add'
+          code = add_project_module(options[:object], config.project)
         when 'install'
-          code = install_project_modules
+          code = install_project_modules(config.project, options[:force])
         when 'generate-types'
           code = generate_types
         end
@@ -837,7 +844,8 @@ module Bolt
                 "project with modules."
         end
 
-        install_modules(puppetfile, {}, moduledir, options[:modules])
+        installer = Bolt::ModuleInstaller.new(outputter, pal)
+        installer.install(options[:modules], puppetfile, moduledir)
       end
 
       # If either bolt.yaml or bolt-project.yaml exist, the user has already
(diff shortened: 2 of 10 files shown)


lib/bolt/bolt_option_parser.rb
```diff
@@ -367,24 +370,26 @@ def get_help_text(subcommand, action = nil)
...
 
-    MODULE_INSTALL_HELP = <<~HELP
+    MODULE_ADD_HELP = <<~HELP
       NAME
-          install
+          add
       
       USAGE
-          bolt module install [options]
+          bolt module add [options]
```
lucywyman, line 384:
```suggestion
          bolt module add <module> [options]
```

lib/bolt/cli.rb
```diff
@@ -291,6 +291,16 @@ def validate(options)
         raise Bolt::CLIError, "Must specify a plan name."
       end
 
+      if options[:subcommand] == 'module' && options[:action] == 'add'
+        unless options[:object]
+          raise Bolt::CLIError, "Must specify a module name."
+        end
+
+        if options[:leftovers].any?
```
lucywyman, line 299: It looks like this is never hit because we error on line 241 first:
```
lucy ~/githubs/bolt ((36c8295...)) $ be bolt module add puppetlabs-apache puppetlabs-aws_inventory --project project/
Unknown argument(s) puppetlabs-aws_inventory
```

You could just add an `elsif subcommand == 'module'` there, or leave that message as is and remove this clause.

lib/bolt/cli.rb
```diff
@@ -858,77 +873,57 @@ def initialize_project
 
     # Installs modules declared in the project configuration file.
     #
-    def install_project_modules
-      if config.project.modules.nil?
-        outputter.print_message "Project configuration file '#{config.project.project_file}' "\
-                                "does not specify any module dependencies. Nothing to do."
+    def install_project_modules(project, force)
+      require 'bolt/module_installer'
```
lucywyman, line 877: If this is expensive to load it seems easier to move requires into the actions in the required files (like in the puppetfile/installer) than to have requires all over the place here.

lib/bolt/cli.rb
```diff
@@ -831,13 +843,16 @@ def initialize_project
       # Puppetfile, install the specified modules. The module installer will
       # resolve dependencies, generate a Puppetfile, and install the modules.
       if options[:modules]
+        require 'bolt/module_installer'
```
lucywyman, line 846: If we want to leave all these requires in, it'd be best to move them to after we validate the puppetfile doesn't exist so we don't load this then immediately error.

lib/bolt/cli.rb
```diff
@@ -858,77 +873,57 @@ def initialize_project
...
+      installer.add(name, modules, project.puppetfile, moduledir, project.project_file)
+    end
+
+    # Asserts that there is a project configuration file.
+    #
+    def assert_project_file(project)
+      unless project.project_file?
+        raise Bolt::Error.new(
+          "Could not find project configuration file #{project.project_file}, unable "\
+          "to install modules. To create a Bolt project with a project configuration "\
```
lucywyman, line 913:
```suggestion
          "to install modules. To create a Bolt project run 'bolt project init'."\
```

lucywyman: This just seems a little verbose...`bolt project init` won't actually make a bolt-project.yaml if bolt.yaml exists, so I don't think we need to worry about differentiating that. If we really want to make it a good experience, we could have a separate message about renaming bolt.yaml to bolt-project.yaml if it exists.

lib/bolt/cli.rb
beechtom: Er, not sure how I missed that...

lib/bolt/cli.rb
```diff
@@ -858,77 +866,54 @@ def initialize_project
...
         return 0
       end
 
-      install_modules(
-        config.puppetfile,
-        config.puppetfile_config,
-        config.project.path + '.modules',
-        config.project.modules
-      )
+      moduledir = project.path + '.modules'
```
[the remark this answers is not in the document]
beechtom: I actually added this in #2189 (see [here](https://github.com/puppetlabs/bolt/pull/2189/files#diff-5fb12f608b24f29bc6e72995332d2c24R184)). I'll port that over to this PR.

spec/bolt/cli_spec.rb
```diff
@@ -189,75 +183,67 @@ def stub_config(file_content = {})
...
-      expect(result).to be
-      expect((project + 'Puppetfile').exist?).to eq(true)
-      expect(File.read(project + 'Puppetfile')).to match(/mod "puppetlabs-yaml"/)
-      expect((project + '.modules').exist?).to eq(true)
-      expect(Dir.children(project + '.modules')).to include('yaml')
-      expect(Dir.children(project + '.modules').size).to be > 1
-    end
+      it 'errors without a module' do
+        cli = Bolt::CLI.new(%W[module add --project #{project}])
+        expect { cli.parse }.to raise_error(Bolt::CLIError)
```
lucywyman, line 196: Checking the message would also be good, so that if this raises a *different* error than we expect we catch it.

lib/bolt/module_installer.rb
```diff
@@ -0,0 +1,148 @@
...
+      @outputter.print_message "Updating project configuration file at #{config}"
+
+      data = Bolt::Util.read_yaml_hash(config, 'project')
+      data['modules'] ||= []
+      data['modules'] <<  { 'name' => new_module.title }
+
+      begin
+        File.write(config, data.to_yaml)
+      rescue SystemCallError => e
+        raise Bolt::FileError(
```
lucywyman, line 48:
```suggestion
        raise Bolt::FileError.new(
```

spec/bolt/module_installer_spec.rb
```diff
@@ -0,0 +1,95 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+require 'bolt/module_installer'
+
+describe Bolt::ModuleInstaller do
+  let(:path)      { @project + 'Puppetfile' }
```
lucywyman, line 7: I think calling this var puppetfile would make it a bit more readable

spec/bolt/module_installer_spec.rb
```diff
@@ -0,0 +1,95 @@
...
+  let(:moduledir) { @project + '.modules' }
+  let(:config)    { @project + 'bolt-project.yaml' }
+  let(:name)      { 'puppetlabs-pkcs7' }
+  let(:modules)   { [{ 'name' => 'puppetlabs-yaml' }] }
```
lucywyman, line 11:
```suggestion
  let(:modules_config)   { [{ 'name' => 'puppetlabs-yaml' }] }
```

spec/bolt/module_installer_spec.rb
```diff
@@ -0,0 +1,95 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+require 'bolt/module_installer'
+
+describe Bolt::ModuleInstaller do
+  let(:path)      { @project + 'Puppetfile' }
+  let(:moduledir) { @project + '.modules' }
+  let(:config)    { @project + 'bolt-project.yaml' }
+  let(:name)      { 'puppetlabs-pkcs7' }
```
lucywyman, line 10:
```suggestion
  let(:new_module)      { 'puppetlabs-pkcs7' }
```

lib/bolt/module_installer.rb
```diff
@@ -0,0 +1,148 @@
...
+module Bolt
+  class ModuleInstaller
+    def initialize(outputter, pal)
+      @outputter = outputter
+      @pal       = pal
+    end
+
+    # Adds a single module to the project.
+    #
+    def add(name, modules, path, moduledir, config)
```
lucywyman, line 14:
```suggestion
    def add(name, modules, puppetfile_path, moduledir, config_path)
```

spec/bolt/cli_spec.rb
```diff
@@ -189,75 +183,70 @@ def stub_config(file_content = {})
...
+          /Must specify a module name/
+        )
+      end
 
-    context 'with existing Puppetfile' do
-      before(:each) do
-        File.write(File.join(@project_dir, 'Puppetfile'), "mod 'puppetlabs-yaml', '0.2.0'")
+      it 'errors with multiple modules' do
+        cli = Bolt::CLI.new(%W[module add foo bar --project #{project}])
+        expect { cli.parse }.to raise_error(Bolt::CLIError)
```
lucywyman, line 204: Check message

spec/bolt/cli_spec.rb
```diff
@@ -189,75 +183,70 @@ def stub_config(file_content = {})
...
+    context 'install' do
+      let(:flags)       { %W[--project #{project}] }
+      let(:cli)         { Bolt::CLI.new(%w[module install] + flags) }
+      let(:config_data) { { 'modules' => [{ 'name' => 'puppetlabs-yaml' }] } }
 
-      it 'installs the Puppetfile if it is not missing specifications' do
-        File.write(File.join(@project_dir, 'bolt-project.yaml'), config.to_yaml)
+      it 'errors if missing a bolt-project.yaml' do
+        FileUtils.rm(config)
+        expect { cli.execute(cli.parse) }.to raise_error(Bolt::Error)
```
lucywyman, line 227: Check message

spec/bolt/module_installer_spec.rb
```diff
@@ -0,0 +1,95 @@
...
+    it 'returns early if the module is already declared' do
+      result = installer.add('puppetlabs-yaml', modules_config, puppetfile, moduledir, config)
+      expect(result).to eq(0)
+      expect(puppetfile.exist?).to eq(false)
+      expect(moduledir.exist?).to eq(false)
+    end
+
+    it 'errors if Puppetfile is not managed by Bolt' do
+      File.write(puppetfile, '')
+      expect { installer.add(new_module, modules_config, puppetfile, moduledir, config) }.to raise_error(Bolt::Error)
```
lucywyman, line 38: Check message

spec/bolt/module_installer_spec.rb
```diff
@@ -0,0 +1,95 @@
...
+
+      expect(result).to eq(0)
+      expect(File.read(puppetfile)).not_to match(/#{spec}/)
+    end
+  end
+
+  context '#install' do
+    it 'errors if Puppetfile is not managed by Bolt' do
+      File.write(puppetfile, '')
+      expect { installer.install(modules_config, puppetfile, moduledir) }.to raise_error(Bolt::Error)
```
lucywyman, line 77: Check message

spec/bolt/module_installer_spec.rb
```diff
@@ -0,0 +1,95 @@
...
+    end
+
+    it 'installs modules forcibly' do
+      File.write(puppetfile, '')
+      expect { installer.install(modules_config, puppetfile, moduledir, force: true) }.not_to raise_error
```
lucywyman, line 82: We should also check that it actually installs the modules, not just doesn't error.

spec/bolt/module_installer_spec.rb
```diff
@@ -0,0 +1,95 @@
...
+    end
+
+    it 'writes a Puppetfile' do
+      installer.install(modules_config, puppetfile, moduledir)
+      expect(puppetfile.exist?).to be(true)
+    end
+
+    it 'installs a Puppetfile' do
+      installer.install(modules_config, puppetfile, moduledir)
+      expect(moduledir.exist?).to be(true)
```
lucywyman, line 92: And a check that one of the expected modules is there too

spec/bolt/module_installer_spec.rb
beechtom: Added a check that the moduledir exists, but checking for specific modules causes problems in windows CI, as `Dir.children(moduledir)` doesn't seem to return anything. This seems more of an r10k / testing environment issue, as I've run into this exact problem before when moving our tests to github.

lucywyman: We use `Dir.children` quite a bit in tests, i.e.: https://github.com/puppetlabs/bolt/blob/main/spec/bolt/project_migrate_spec.rb#L77 on both platforms. Is the issue that modules don't install to the right directory on Windows? Did you verify manually this works as expected on Windows?

beechtom: No, the issue is that the modules do not seem to install to the moduledir at all in Windows CI. Running the command on Windows correctly installs modules to the expected location.

It may make more sense to just have this test confirm that `#install_puppetfile` is called.

spec/bolt/cli_spec.rb
lucywyman: I don't think this got addressed?

beechtom: Whoops, I made this change to a similar test further up.

beechtom: Updated.

nicklewis (commented):
I started this review a while ago so I'm not sure what will happen with a couple of the comments... Guess we'll find out!

lib/bolt/module_installer.rb
```diff
@@ -0,0 +1,144 @@
...
+      rescue SystemCallError => e
+        raise Bolt::FileError(
+          "Unable to update project configuration file: #{e.message}",
+          config
+        )
+      end
+
+      # Write the Puppetfile.
+      @outputter.print_message "Writing Puppetfile at #{path}"
+      puppetfile.write(path, force: true)
```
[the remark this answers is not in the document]
nicklewis: This one is "outdated" because the line it was written against changed, but I think the comment itself is still valid.

lib/bolt/module_installer.rb
```diff
@@ -0,0 +1,144 @@
...
+      # If there is an existing Puppetfile, add the new module and attempt
+      # to resolve. This will not update the versions of any installed modules.
+      if path.exist?
+        puppetfile = Bolt::Puppetfile.parse(path)
+        puppetfile.add_modules(new_module)
+
+        begin
+          puppetfile.resolve
+          return puppetfile
+        rescue Bolt::Error
```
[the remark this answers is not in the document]
beechtom: Ah, good idea. Added a debug statement in the rescue. To me it makes a bit more sense there than after, since if the second resolve fails it will issue an error detailing which modules have incompatible version requirements. I think that error makes less sense without the debug message saying Bolt will attempt to re-resolve from the config.

lib/bolt/module_installer.rb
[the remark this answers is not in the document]
beechtom: Ya, I think the `force` option is pretty useless here. :) Removed.

lib/bolt/cli.rb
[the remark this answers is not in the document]
lucywyman: I think porting the managed_moduledir bit here would be helpful, since I think this will get in first.

lucywyman (changes requested):
I'd like to move the `managed_moduledir` project class attribute change to this PR so we don't have to immediately change it back. Then I think this is gtg.

lib/bolt/module_installer.rb
```diff
@@ -0,0 +1,144 @@
...
+    def add(name, modules, path, moduledir, config)
+      # If the project configuration file already includes this module,
+      # exit early.
+      puppetfile  = Bolt::Puppetfile.new(modules)
+      new_module  = Bolt::Puppetfile::Module.from_hash('name' => name)
+
+      if puppetfile.modules.include?(new_module)
+        @outputter.print_message "Project configuration file #{config} already includes "\
+                                 "module #{new_module}. Nothing to do."
+        return 0
```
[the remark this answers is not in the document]
lucywyman: This wasn't addressed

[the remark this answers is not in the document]
beechtom: Updated that one `return 0`. This should now only return `true/false`.

Issue #2166, opened 2020-09-11
Bolt versions > 2.23.0 is failing on Windows


daianamezdrea:
## Describe the Bug
For example, for inifile module the litmus rake task:bundle exec rake litmus:acceptance:parallel  fails with Bolt 2.26.0, 2.25.0, 2.24.0 on Windows after maximum 10 seconds(tests are not running at all)

## Expected Behavior
To run tests 

## Steps to Reproduce
1.Take a module which has Windows tests
2.provision, install agent, install module, run acceptance tests

```
An error occurred while loading ./spec/acceptance/ini_setting_spec.rb.
On host `keen-profession.delivery.puppetlabs.net'
Failure/Error: LitmusHelper.instance.apply_manifest(pp)
NoMethodError:
  undefined method `trace' for #<Logging::Logger:0x00007fa9e690c1e8>

# ./.bundle/gems/ruby/2.5.0/gems/bolt-2.27.0/lib/bolt/analytics.rb:228:in `event'
# ./.bundle/gems/ruby/2.5.0/gems/bolt-2.27.0/lib/bolt/executor.rb:208:in `report_transport'
# ./.bundle/gems/ruby/2.5.0/gems/bolt-2.27.0/lib/bolt/executor.rb:121:in `block in queue_execute'
# ./.bundle/gems/ruby/2.5.0/gems/bolt-2.27.0/lib/bolt/executor.rb:119:in `each'
# ./.bundle/gems/ruby/2.5.0/gems/bolt-2.27.0/lib/bolt/executor.rb:119:in `flat_map'
# ./.bundle/gems/ruby/2.5.0/gems/bolt-2.27.0/lib/bolt/executor.rb:119:in `queue_execute'
# ./.bundle/gems/ruby/2.5.0/gems/bolt-2.27.0/lib/bolt/executor.rb:174:in `batch_execute'
# ./.bundle/gems/ruby/2.5.0/gems/bolt-2.27.0/lib/bolt/executor.rb:317:in `block in upload_file'
# ./.bundle/gems/ruby/2.5.0/gems/bolt-2.27.0/lib/bolt/executor.rb:182:in `log_action'
# ./.bundle/gems/ruby/2.5.0/gems/bolt-2.27.0/lib/bolt/executor.rb:314:in `upload_file'
# ./.bundle/gems/ruby/2.5.0/gems/bolt-2.27.0/lib/bolt_spec/run.rb:225:in `upload_file'
[...]

PR #2174, opened 2020-09-15
(GH-2166) Ensure logging is initialized in inventory

nicklewis:
Commit 546b5554 replaced all calls to `Logging.logger` with a wrapper
method `Bolt::Logger.logger` that initializes logging before returning a
logger object, to ensure that the log methods we need are created.
However, commit 23bb8b33 was in flight at the same time and added a
*new* call to `Logging.logger`. These two changes logically conflicted
but not in a way that could be identified by Git. We therefore wound up
with an undesired call directly to `Logging.logger`, which can cause
failures related to the logging not being initialized when using Bolt as
a library.

This commit changes the `Logging.logger` call to the proper
`Bolt::Logger.logger`.

!no-release-note

### DIFF
--- a/lib/bolt/inventory.rb
+++ b/lib/bolt/inventory.rb
@@ -46,7 +46,7 @@ module Bolt
     end
 
     def self.from_config(config, plugins)
-      logger = Logging.logger[self]
+      logger = Bolt::Logger.logger(self)
 
       if ENV.include?(ENVIRONMENT_VAR)
         begin


beechtom (approved):

lucywyman:
Oh ugh - I think this needs a rebase for the tests to go green.

nicklewis:
Rebased

PR #2185, opened 2020-09-16
(maint) Automatically compute next version number

nicklewis:
This checks whether the changelog includes any new features,
deprecations or removals and increments the y version if so, and the z
version if not. It still allows the version to be set explicitly if
needed, for major releases or cases where the heuristic is wrong.

!no-release-note

### DIFF
--- a/scripts/generate_changelog.rb
+++ b/scripts/generate_changelog.rb
@@ -6,12 +6,7 @@ require 'octokit'
 class ChangelogGenerator
   attr_reader :version, :entries
 
-  def initialize(version)
-    unless version
-      warn "Usage: generate_changelog.rb VERSION"
-      exit 1
-    end
-
+  def initialize(version = nil)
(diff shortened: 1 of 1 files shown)


scripts/generate_changelog.rb
```diff
@@ -131,9 +136,11 @@ def generate
 
     if entries.each_value.all? { |type| type[:entries].empty? }
       warn "No release notes for #{latest}..main"
-      exit 0
```
beechtom, line 134: I'd leave this in, just because if there aren't any release notes we don't actually know what the next version will be. Other than that, lgtm! 👍

nicklewis: Whoops, good catch. I removed that line for testing (since we just did a release and there were no changes 😄) and forgot to put it back.

beechtom (approved):

PR #2189, opened 2020-09-17
(GH-2182) Use new modulepath when 'modules' is configured

### LINKED ISSUE #2182: Update Modulepath for new module workflow users
## Use Case
When bolt detects a `modules` key in `bolt-project.yaml` it should use the new cleaner modulepath logic.

- The default modulepath should be a single directory `modules`
- The `.modules` directory should be appended to the modulepath.
- bolt should error if `.modules` appears on the user configured part of the modulepath.
- Users who attempt to use `bolt module` without a `modules` key should be informed they need to update their project and linked to documentation when done.
- Users who attempt to use `bolt puppetfile` with a `modules` key present should be informed that it is deprecated and bolt should error.
- Appending should happen inside `Bolt::Config` so it can be picked up transparently by PE.

beechtom:
This updates `Bolt::Project` and `Bolt::Config` to use a new modulepath
when `modules` is configured in `bolt-project.yaml`. This work is part
of the new module management workflow.

When `modules` is configured, the default modulepath is `['modules']`
and the full modulepath will have `.modules` appended to it, whether
Bolt uses the default modulepath or the user-configured modulepath. If
the user-configured modulepath includes `.modules`, Bolt will error and
display a helpful message that the modulepath cannot be configured with
`.modules` as it is automatically appended.

If `modules` is not configured, Bolt's behavior remains the same and the
default modulepath is `['modules', 'site-modules', 'site']`.

This also adds new `BoltSpec::Project` library functions that will
create a temporary project. This is useful for spec tests and prevents
Bolt from unintentionally loading configuration on the system running
the tests.

This change is shipping dark until the full module management workflow
is complete.

---

This also deprecates the `bolt puppetfile install` command when
`modules` is configured, and disables the `bolt module *` commands
when `modules` is not configured.

---

Closes #2182

### DIFF
--- a/lib/bolt/cli.rb
+++ b/lib/bolt/cli.rb
@@ -190,6 +190,9 @@ module Bolt
 
       warn_inventory_overrides_cli(options)
 
+      # Disable the puppetfile subcommand when 'modules' is configured.
+      assert_puppetfile_command(config.project.modules)
+
       options
     rescue Bolt::Error => e
       outputter.fatal_error(e)
@@ -474,7 +477,11 @@ module Bolt
         when 'generate-types'
           code = generate_types
         when 'install'
-          code = install_puppetfile(config.puppetfile_config, config.puppetfile, config.modulepath.first)
+          code = install_puppetfile(
+            config.puppetfile_config,
+            config.puppetfile,
+            config.modulepath.first
+          )
         end
       when 'secret'
         code = Bolt::Secret.execute(plugins, outputter, options)
@@ -920,6 +927,19 @@ module Bolt
       installer = Bolt::ModuleInstaller.new(outputter, pal)
       ok = installer.install_puppetfile(puppetfile, moduledir, config)
       ok ? 0 : 1
+    end
+
+    # Raises an error if the 'puppetfile install' command is deprecated due to
+    # modules being configured.
+    #
+    def assert_puppetfile_command(modules)
+      if modules
+        raise Bolt::CLIError,
+              "Unable to use command 'bolt puppetfile #{options[:action]}' when "\
+              "'modules' is configured in bolt-project.yaml. Use the 'module' command "\
+              "instead. For a list of available actions for the 'module' command, run "\
+              "'bolt module --help'."
+      end
     end
 
     def pal
--- a/lib/bolt/cli.rb
+++ b/lib/bolt/cli.rb
@@ -30,19 +30,20 @@ require 'bolt/module_installer'
 module Bolt
   class CLIExit < StandardError; end
   class CLI
-    COMMANDS = { 'command' => %w[run],
-                 'script' => %w[run],
-                 'task' => %w[show run],
-                 'plan' => %w[show run convert new],
-                 'file' => %w[download upload],
-                 'puppetfile' => %w[install show-modules generate-types],
-                 'secret' => %w[encrypt decrypt createkeys],
-                 'inventory' => %w[show],
-                 'group' => %w[show],
-                 'project' => %w[init migrate],
-                 'apply' => %w[],
-                 'guide' => %w[],
-                 'module' => %w[add generate-types install show] }.freeze
+    COMMANDS = {
+      'command'    => %w[run],
+      'script'     => %w[run],
+      'task'       => %w[show run],
+      'plan'       => %w[show run convert new],
+      'file'       => %w[download upload],
+      'puppetfile' => %w[install show-modules generate-types],
+      'secret'     => %w[encrypt decrypt createkeys],
+      'inventory'  => %w[show],
+      'group'      => %w[show],
+      'project'    => %w[init migrate],
+      'apply'      => %w[],
+      'guide'      => %w[]
+    }.freeze
 
     attr_reader :config, :options
 
@@ -59,6 +60,14 @@ module Bolt
     end
     private :inventory
 
+    def commands
+      if ENV['BOLT_MODULE_FEATURE']
+        COMMANDS.merge('module' => %w[add generate-types install show])
+      else
+        COMMANDS
+      end
+    end
+
     def help?(remaining)
       # Set the subcommand
       options[:subcommand] = remaining.shift
@@ -70,7 +79,7 @@ module Bolt
 
       # This section handles parsing non-flag options which are
       # subcommand specific rather then part of the config
-      actions = COMMANDS[options[:subcommand]]
+      actions = commands[options[:subcommand]]
       if actions && !actions.empty?
         options[:action] = remaining.shift
       end
@@ -100,6 +109,10 @@ module Bolt
       # This part aims to handle both `bolt <mode> --help` and `bolt help <mode>`.
       remaining = handle_parser_errors { parser.permute(@argv) } unless @argv.empty?
       if @argv.empty? || help?(remaining)
+        # If the subcommand is not enabled, display the default
+        # help text
+        options[:subcommand] = nil unless commands.include?(options[:subcommand])
+
         # Update the parser for the subcommand (or lack thereof)
         parser.update
         puts parser.help
@@ -190,8 +203,9 @@ module Bolt
 
       warn_inventory_overrides_cli(options)
 
-      # Disable the puppetfile subcommand when 'modules' is configured.
-      assert_puppetfile_command(config.project.modules)
+      # Assert whether the puppetfile/module commands are available depending
+      # on whether 'modules' is configured.
+      assert_puppetfile_or_module_command(config.project.modules)
 
       options
     rescue Bolt::Error => e
@@ -220,17 +234,13 @@ module Bolt
     end
 
     def validate(options)
-      # Disables the 'module' subcommand unless the module feature flag is set.
-      commands = COMMANDS.dup
-      commands.delete('module') unless ENV['BOLT_MODULE_FEATURE']
-
       unless commands.include?(options[:subcommand])
         raise Bolt::CLIError,
               "Expected subcommand '#{options[:subcommand]}' to be one of " \
               "#{commands.keys.join(', ')}"
       end
 
-      actions = COMMANDS[options[:subcommand]]
+      actions = commands[options[:subcommand]]
       if actions.any?
         if options[:action].nil?
           raise Bolt::CLIError,
@@ -932,13 +942,18 @@ module Bolt
     # Raises an error if the 'puppetfile install' command is deprecated due to
     # modules being configured.
     #
-    def assert_puppetfile_command(modules)
-      if modules
+    def assert_puppetfile_or_module_command(modules)
+      if modules && options[:subcommand] == 'puppetfile'
         raise Bolt::CLIError,
               "Unable to use command 'bolt puppetfile #{options[:action]}' when "\
               "'modules' is configured in bolt-project.yaml. Use the 'module' command "\
               "instead. For a list of available actions for the 'module' command, run "\
               "'bolt module --help'."
+      elsif modules.nil? && options[:subcommand] == 'module'
+        raise Bolt::CLIError,
+              "Unable to use command 'bolt module #{options[:action]}'. To use "\
+              "this command, update your project configuration to manage module "\
+              "dependencies."
       end
     end
 
--- a/lib/bolt/config.rb
+++ b/lib/bolt/config.rb
@@ -453,7 +453,7 @@ module Bolt
     def modulepath
       path = @data['modulepath'] || @project.modulepath
 
-      if @project.modules
+      if @project.modules && ENV['BOLT_MODULE_FEATURE']
         path + [@project.managed_moduledir]
       else
         path
--- a/lib/bolt/config.rb
+++ b/lib/bolt/config.rb
@@ -391,6 +391,12 @@ module Bolt
         @logs << { warn: msg }
       end
 
+      if @project.modules && @data['modulepath']&.include?(@project.managed_moduledir)
+        raise Bolt::ValidationError,
+              "Found invalid path in modulepath: #{@project.managed_moduledir}. This path "\
+              "is automatically appended to the modulepath and cannot be configured."
+      end
+
       keys = OPTIONS.keys - %w[plugins plugin_hooks puppetdb]
       keys.each do |key|
         next unless Bolt::Util.references?(@data[key])
@@ -445,7 +451,13 @@ module Bolt
     end
 
     def modulepath
-      @data['modulepath'] || @project.modulepath
+      path = @data['modulepath'] || @project.modulepath
+
+      if @project.modules
+        path + [@project.managed_moduledir]
+      else
+        path
+      end
     end
 
     def modulepath=(value)
--- a/lib/bolt/project.rb
+++ b/lib/bolt/project.rb
@@ -114,7 +114,7 @@ module Bolt
 
       # If the 'modules' key is present in the project configuration file,
       # use the new, shorter modulepath.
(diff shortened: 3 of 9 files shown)


spec/bolt/cli_spec.rb
```diff
@@ -164,6 +165,69 @@ def stub_config(file_content = {})
...
+    context 'with modules configured' do
+      let(:project_config) { { 'modules' => [] } }
+      let(:command)        { %w[module show] }
+
+      it 'does not error' do
+        expect { cli.parse }.not_to raise_error
+      end
+    end
+
+    describe 'install' do
```
beechtom, line 210: There's a set of tests for 'module install' starting at 231, but I've left them as-is for now since https://github.com/puppetlabs/bolt/pull/2172 makes some changes to them. Depending on which PR makes it in first, I'll rebase the other and update the tests accordingly.

lib/bolt/cli.rb
```diff
@@ -184,6 +197,14 @@ def finalize_setup
 
       warn_inventory_overrides_cli(options)
 
+      # Error if the module command is disabled
+      if options[:subcommand] == 'module' && config.project.modules.nil?
+        raise Bolt::CLIError,
+              "Unable to use command 'bolt module #{options[:action]}'. To use "\
+              "this command, update your project configuration to manage module "\
+              "dependencies."
```
beechtom, line 205: This message will need a link to the proper docs once those are written. I left it out for now since I'm not sure what the URL for that page will be.

lucywyman: This hits the `validate()` function before getting here, which checks the subcommands on line 237. I think this message needs to be checked and raised there.
```
lucy ~/githubs/bolt ((1ab94b3...)) $ be bolt module help
Expected subcommand 'module' to be one of command, script, task, plan, file, puppetfile, secret, inventory, group, project, apply, guide
```

lucywyman: Similarly, it looks like if `bolt module help` or `bolt module --help` is run there's not message that the user needs to enable the module subcommand, which I think would be useful.

spec/bolt/cli_spec.rb
```diff
@@ -165,6 +165,68 @@ def stub_config(file_content = {})
...
+      ENV['BOLT_MODULE_FEATURE'] = original
+    end
+
+    before(:each) do
+      allow(Bolt::Project).to receive(:new).and_return(project)
+    end
+
+    it 'errors without BOLT_MODULE_FEATURE being set' do
+      ENV.delete('BOLT_MODULE_FEATURE')
+      expect { cli.parse }.to raise_error(Bolt::CLIError)
```
lucywyman, line 191: Check message

spec/bolt/cli_spec.rb
```diff
@@ -165,6 +165,68 @@ def stub_config(file_content = {})
...
+    end
+
+    context 'without modules configured' do
+      it 'errors' do
+        expect { cli.parse }.to raise_error(Bolt::CLIError)
```
lucywyman, line 196: For me, this raises a different error than I think is expected: 
```
From: /home/lucy/githubs/bolt/spec/bolt/cli_spec.rb:197 :

    192:     end
    193: 
    194:     context 'without modules configured' do
    195:       it 'errors' do
    196:         require 'pry'; binding.pry
 => 197:         expect { cli.parse }.to raise_error(Bolt::CLIError)
    198:       end
    199:     end
    200: 
    201:     context 'with modules configured' do
    202:       let(:project_config) { { 'modules' => [] } }

[1] pry(#<RSpec::ExampleGroups::BoltCLI::Module::WithoutModulesConfigured>)> cli.parse
Bolt::CLIError: Expected an action of the form 'bolt module <action>'
from /home/lucy/githubs/bolt/lib/bolt/cli.rb:244:in `validate
```

spec/bolt/cli_spec.rb
```diff
@@ -165,6 +165,68 @@ def stub_config(file_content = {})
...
+    before(:each) do
+      allow(Bolt::Project).to receive(:new).and_return(project)
+    end
+
+    it 'errors without BOLT_MODULE_FEATURE being set' do
+      ENV.delete('BOLT_MODULE_FEATURE')
+      expect { cli.parse }.to raise_error(Bolt::CLIError)
+    end
+
+    context 'without modules configured' do
```
lucywyman, line 194: I don't think we need to wrap this in a context, the `it` should just have a more verbose message.

spec/bolt/cli_spec.rb
```diff
@@ -165,6 +165,68 @@ def stub_config(file_content = {})
...
+        expect { cli.parse }.to raise_error(Bolt::CLIError)
+      end
+    end
+
+    context 'with modules configured' do
+      let(:project_config) { { 'modules' => [] } }
+      let(:command)        { %w[module show] }
+
+      it 'does not error' do
+        expect { cli.parse }.not_to raise_error
```
lucywyman, line 205: We should check expected output from the command.

spec/bolt/cli_spec.rb
```diff
@@ -165,6 +165,68 @@ def stub_config(file_content = {})
...
+          'modules' => [
+            { 'name' => 'puppetlabs-yaml' }
+          ]
+        }
+      end
+
+      it 'does nothing if project config has no module declarations' do
+        allow(project).to receive(:modules).and_return([])
+        result = cli.execute(cli.parse)
+        expect(result).to be
```
lucywyman, line 223:
```suggestion
        expect(result).to eq(0)
```

lib/bolt/cli.rb
beechtom: We can't perform this check as part of `validate()` without changing when/how the command is validated. The validations happens before config is loaded, and we need config to be loaded to check whether `modules` is configured.

Also, the `module` command is still feature flagged in this PR and `help` isn't a valid action for any subcommand, so that may be part of what you're running into? Setting the flag and running a `module` command without `modules` configured will get you to this message and not error during validation.

I have updated the help text for the `module` commands to say that `modules` needs to be configured.

spec/bolt/cli_spec.rb
```diff
@@ -2454,20 +2451,26 @@ def stub_config(file_content = {})
...
         expect(modules.values.first.map { |h| h[:name] }).to eq(%w[boltlib ctrl dir file out prompt system])
         expect(modules.values[1].map { |h| h[:name] })
           .to include("aggregate", "canary", "puppetdb_fact", "puppetlabs/yaml")
       end
+
+      it 'errors when modules is configured' do
+        allow(Bolt::Util).to receive(:read_yaml_hash).and_call_original
+        allow(Bolt::Util).to receive(:read_optional_yaml_hash).and_call_original
+
+        expect { cli.execute(cli.parse) }.to raise_error(Bolt::Error)
```
lucywyman, line 2518: Check message

spec/bolt/cli_spec.rb
beechtom: Updated to check that Bolt exits with a successful code instead.

spec/integration/plan_spec.rb
```diff
@@ -79,9 +79,8 @@
       end
 
       it 'runs registers types defined in $project/.resource_types', ssh: true do
-        ENV['BOLT_MODULE_FEATURE'] = 'true'
         # generate types based and save in project (based on value of --configfile)
-        run_cli(%w[module generate-types] + config_flags)
+        run_cli(%w[puppetfile generate-types] + config_flags)
```
lucywyman, line 83: We should just have two separate tests for each of these cases.

lib/bolt/project.rb
```diff
@@ -167,6 +166,25 @@ def modules
       @data['modules']
     end
 
+    # Returns the default modulepath for the project.
+    # If the 'modules' key is present in the project configuration file,
+    # use the new, shorter modulepath.
+    #
+    def modulepath
+      if modules
```
lucywyman, line 174: This seems like it's really saying `unless modules.nil?`

beechtom: Tomato tomato?

lucywyman: It's just more readable. This also seems like we should still be setting the attribute, either in this method as a `@modulepath ||=`, or in the init function. Recomputing it each time seems unnecessary..

lib/bolt/project.rb
```diff
@@ -167,6 +166,25 @@ def modules
...
+        [(@path + 'modules').to_s]
+      else
+        [(@path + 'modules').to_s, (@path + 'site-modules').to_s, (@path + 'site').to_s]
+      end
+    end
+
+    # The path to the managed moduledir, where Bolt installs modules
+    # when using the 'module' command.
+    #
+    def managed_moduledir
```
lucywyman, line 184: This also seems like an attribute of the class, not a function.

lib/bolt/cli.rb
```diff
@@ -919,9 +941,17 @@ def install_modules(puppetfile_path, config, moduledir, modules)
 
     # Loads a Puppetfile and installs its modules.
     #
-    def install_puppetfile(config, puppetfile, moduledir)
+    def install_puppetfile(config, puppetfile, moduledir, modules)
       require 'bolt/puppetfile/installer'
 
+      if modules
```
lucywyman, line 947: This seems like it'd make more sense directly in the `when 'puppetfile'` clause before the case statement so that this is less prone to getting raised in situations where we're not running a puppetfile subcommand.

lib/bolt/project.rb
nicklewis: Given there's an `else`, `if` is clearer. `unless/else` is really confusing.

lib/bolt/cli.rb
```diff
@@ -474,7 +491,12 @@ def execute(options)
         when 'generate-types'
           code = generate_types
         when 'install'
-          code = install_puppetfile(config.puppetfile_config, config.puppetfile, config.modulepath.first)
+          assert_puppetfile_install_command(config.project.modules)
```
lucywyman, line 494: Do we also want to show a deprecation warning for other puppetfile commands if modules is configured? Mostly it seems confusing that only one action is deprecated while others are valid. If I was a user I'd wonder if only `puppetfile install` was eventually getting deprecated, not the whole subcommand.

lib/bolt/cli.rb
```diff
@@ -100,6 +109,10 @@ def parse_command
       # This part aims to handle both `bolt <mode> --help` and `bolt help <mode>`.
       remaining = handle_parser_errors { parser.permute(@argv) } unless @argv.empty?
       if @argv.empty? || help?(remaining)
+        # If the subcommand is not enabled, display the default
+        # help text
+        options[:subcommand] = nil unless commands.include?(options[:subcommand])
```
lucywyman, line 114: I think it'd be a better experience if we instructed users on how to enable the module subcommand and linked to docs, rather than just acting like it doesn't exist. If I had heard of the module subcommand (through slack or the docs or something) and went to try it, I wouldn't know where to go to learn more about why it wasn't showing up at all.  We can include caveats about it being experimental both in the message and the docs.

spec/bolt/cli_spec.rb
```diff
@@ -207,15 +230,6 @@ def stub_config(file_content = {})
         )
       end
 
-      it 'errors if missing a bolt-project.yaml' do
```
lucywyman, line 210: Not clear why we're removing this, rather than just updating the message?

spec/bolt/cli_spec.rb
```diff
@@ -224,39 +238,21 @@ def stub_config(file_content = {})
     end
 
     context 'install' do
-      let(:flags)       { %W[--project #{project}] }
-      let(:cli)         { Bolt::CLI.new(%w[module install] + flags) }
-      let(:config_data) { { 'modules' => [{ 'name' => 'puppetlabs-yaml' }] } }
+      let(:command)        { %W[module install --project #{project}] }
+      let(:project_config) { { 'modules' => [] } }
 
-      it 'errors if missing a bolt-project.yaml' do
```
lucywyman, line 231: Same here

spec/bolt/cli_spec.rb
```diff
@@ -224,39 +238,21 @@ def stub_config(file_content = {})
...
       end
 
-      it 'installs project modules' do
+      it 'runs' do
         expect(installer).to receive(:install)
         cli.execute(cli.parse)
       end
-
-      it 'installs project modules forcibly' do
-        cli = Bolt::CLI.new(%W[module install --project #{project} --force])
```
lucywyman, line 252: And here.

lib/bolt/cli.rb
beechtom: This is only preventing users from seeing the help text for a feature-flagged command. It won't prevent users from seeing the help text for the `module` command if `modules` is not configured.

As for updating the help text itself to include links to the docs, that can happen once the docs are finished and we know what the URL to the page is.

spec/bolt/cli_spec.rb
beechtom: This is covered by this test: https://github.com/puppetlabs/bolt/blob/main/spec/bolt/module_installer_spec.rb#L87

spec/bolt/cli_spec.rb
beechtom: Removed accidentally.

spec/bolt/cli_spec.rb
beechtom: Removed accidentally.

lib/bolt/cli.rb
beechtom: The ticket called for just deprecating `puppetfile install`, but I agree that the entire `puppetfile` command makes more sense.

spec/bolt/cli_spec.rb
beechtom: Ah, scratch that last comment. Removed them because use of the `module` command requires `modules` be configured, which requires you have a `bolt-project.yaml`. `Bolt::CLI.assert_project_file` is dead code now. 😛

spec/integration/plan_spec.rb
```diff
@@ -68,23 +68,34 @@
...
         result = run_cli(['plan', 'run', 'resource_types', '--targets', target] + config_flags)
         expect(JSON.parse(result)).to eq('built-in' => 'success', 'core' => 'success', 'custom' => 'success')
       end
+
+      it 'runs registers types defined in $project/.resource_types', ssh: true do
+        Dir.mktmpdir(nil, Dir.pwd) do |tmpdir|
```
lucywyman, line 90: Can this just use `with_project`?

lib/bolt/config.rb
```diff
@@ -445,7 +451,13 @@ def puppetfile
     end
 
     def modulepath
-      @data['modulepath'] || @project.modulepath
+      path = @data['modulepath'] || @project.modulepath
+
+      if @project.modules
```
lucywyman, line 456: Should this also only be enabled if `BOLT_MODULE_FEATURE` is set?

lib/bolt/project.rb
```diff
@@ -113,6 +112,14 @@ def initialize(raw_data, path, type = 'option', logs = [])
 
       @data = raw_data.reject { |k, _| Bolt::Config::INVENTORY_OPTIONS.include?(k) }
 
+      # If the 'modules' key is present in the project configuration file,
+      # use the new, shorter modulepath.
+      @modulepath = if @data.key?('modules')
```
lucywyman, line 117: And BOLT_MODULE_FEATURE?

Issue #2127, opened 2020-08-27
Implement parallelism / threading capability in plans


reidmv:
## Use Case

In the early days of Bolt, plan execution was entirely procedural and each procedure needed to finish before the next could start. This meant performing similar-but-not-identical actions across a large set of nodes could be very slow, as it effectively needed to be serial.

`run_task_with` provided a means of quickly running a _task_ across a set of nodes which all needed slightly different parameters. Provided you only needed to run one task with differing parameters and parallelism, this was useful. Leveraging it can result in code that [isn't very readable though](https://github.com/puppetlabs/puppetlabs-peadm/blob/fe75751b7ed35c652e13b1c9d4edf2292e75305c/plans/action/install.pp#L309-L343), compared to what it would simplistically look like to call `run_task()` multiple times with different parameters.

Suppose you want to run a sequence of tasks though, or a plan, across a set of nodes, with slightly different parameters for each. Today this requires you to go back to serial execution one node at a time, waiting for each node to finish before moving on to the next.

## Describe the Solution You Would Like

Provide real asynchronous constructs in the Bolt plan language to accelerate execution of more complicated actions across nodes in parallel.

Suggested implementation:

* `run_task_async` function
* `run_plan_async` function
* `AsyncResultSet` type
* `AsyncPlanResult` type

The `*_async` functions should immediately return their respective `AsyncResult*` type, and begin async processing of the plan or task in a separate thread.

The `AsyncResult*` types should support `.get()` methods, which when called will block until the separate thread completes and return a `ResultSet`, `PlanResult`, or `Error` object.

```puppet
# Kick off arbitrarily complex stuff you need to do in the background
$async_plan_results = $targets.each |$target| {
  run_plan_async('plan::mcplanface', $target,
    important_parameter => $hash[$target],
  }
}

# Do other stuff if you need to

# When you need the plan results
$plan_results = $async_plan_results.map |$async_result| { $async_result.get }
```

adreyer:
We've talked about this for a long time and it seems reasonable to do some exploratory work to figure out exactly how hard it is.  We should do a spike to experiment with parallel plan execution and discover and see how we need to manage the puppet compiler.

Scope:
Implement a `plan_future` function that starts a plan and returns a `Future` datatype. The `Future` datatype should allow the plan author to:
- check the status of a future
- block until the future is resolved
- fetch the result of the future(or raise its error)
- return an error raised in the future without raising and catching it.

reidmv:
[...]

PR #2190, opened 2020-09-17
(GH-2127) Parallelize plan steps across targets

lucywyman:
This implements a new (to-be-named) `parallelize` function which
accepts an array and a block and maps the given block to the array in
parallel. That is to say, the block is executed in sequence for each
element in the array, with the array elements executing in parallel.
This is in contrast to typical plan steps which execute each step in
parallel across targets and wait for all targets to finish before
continuing to the next step.

The block will wait for the sequence to finish on all elements of the
array before returning an array of Results - one per element in the
array. The results can be either the value of a `return` statement, the
result of the last line in the block (can be nil), or an error that was
raised during the block for that element. The results are in the same
order the original array was in, regardless of which elements finish the
block 'first'. Unless `catch_errors()` wraps the block (either inside or
outside), if any element errored during the block the plan will fail and
all results (successful and error) will be included in the failure
details. If `catch_errors()` wraps the block then the array will contain
Error objects for errored results, as usual.

Parallel execution is managed by the executor. When a parallel block
begins, we make a copy of the existing variable scope for each element
in the array so that any variables defined in the scope won't get
overridden by parallel execution. Then we create a Ruby Fiber for each
element of the array executing the block, and send the array of Fibers
to the executor. The executor goes into "parallel mode", where it will
resume each fiber in turn until the block has completed for all elements
of the array. Certain plan function can then be "Fiber aware" by
starting a new thread and yielding their Fiber once the thread has
started, waiting for the thread to finish before returning the result
back to the executor. This allows certain functions to not block on
other array elements continuing to execute the block, while other
functions (or just certain parts of functions, like loading tasks) will
block.

!feature

* **Experimental: Add `parallelize()` plan function**

  This adds a new function `parallelize()` that accepts an array and a
  block and executes the block on each element of the array in parallel.
  This allows plan authors to continue executing a block on other targets
  if a particular plan function may take a long time on one target. This
  features is experimental.

### DIFF
--- a/bolt-modules/boltlib/lib/puppet/functions/parallelize.rb
+++ b/bolt-modules/boltlib/lib/puppet/functions/parallelize.rb
@@ -1,0 +1,94 @@
+# frozen_string_literal: true
+
+require 'bolt/yarn'
+
+# Map a code bock onto an array, where each array element executes in parallel.
+# This function is experimental.
+#
+# > **Note:** Not available in apply block.
+Puppet::Functions.create_function(:parallelize, Puppet::Functions::InternalFunction) do
+  # Map a block onto an array, where each array element executes in parallel.
+  # This function is experimental.
+  # @param data The array to apply the block to.
+  # @return [Array] An array of PlanResult objects. Each input from the input
+  #   array returns a corresponding PlanResult object.
+  # @example Execute two tasks on multiple targets. Once the task finishes on one
+  #   target, that target can move to the next step without waiting for the task
+  #   to finish on the second target.
+  # $targets = get_targets(["host1", "host2"])
+  # $result = parallelize ($targets) |$t| {
+  #   run_task('a', $t)
+  #   run_task('b', $t)
+  # }
+  dispatch :parallelize do
+    scope_param
+    param 'Array[Any]', :data
+    block_param 'Callable[Any]', :block
+    return_type 'Array[Boltlib::PlanResult]'
+  end
+
+  def parallelize(scope, data, &block)
+    unless Puppet[:tasks]
+      raise Puppet::ParseErrorWithIssue
+        .from_issue_and_stack(Bolt::PAL::Issues::PLAN_OPERATION_NOT_SUPPORTED_WHEN_COMPILING, action: 'parallelize')
+    end
+
+    executor = Puppet.lookup(:bolt_executor)
+    executor.report_function_call(self.class.name)
+
+    skein = data.each_with_index.map do |object, index|
+      fiber = Fiber.new do
+        # Create the new scope
+        newscope = Puppet::Parser::Scope.new(scope.compiler)
+        local = Puppet::Parser::Scope::LocalScope.new
+
+        # Compress the current scopes into a single vars hash to add to the new scope
+        current_scope = scope.effective_symtable(true)
+        until current_scope.nil?
+          current_scope.instance_variable_get(:@symbols)&.each_pair { |k, v| local[k] = v }
+          current_scope = current_scope.parent
+        end
+        newscope.push_ephemerals([local])
+
+        begin
+          result = catch(:return) do
+            args = { block.parameters[0][1].to_s => object }
+            block.closure.call_by_name_with_scope(newscope, args, true)
+          end
+
+          # If we got a return from the block, get it's value
+          # Otherwise the result is the last line from the block
+          result = result.value if result.is_a?(Puppet::Pops::Evaluator::Return)
+
+          # Validate the result is a PlanResult
+          unless Puppet::Pops::Types::TypeParser.singleton.parse('Boltlib::PlanResult').instance?(result)
+            raise Bolt::InvalidParallelResult.new(result.to_s, *Puppet::Pops::PuppetStack.top_of_stack)
+          end
+
+          result
+        rescue Puppet::PreformattedError => e
+          if e.cause.is_a?(Bolt::Error)
+            e.cause
+          else
+            raise e
+          end
+        end
+      end
+
+      Bolt::Yarn.new(fiber, index)
+    end
+
+    result = executor.round_robin(skein)
+
+    failed_indices = result.each_index.select do |i|
+      result[i].is_a?(Bolt::Error)
+    end
+
+    # TODO: Inner catch errors block?
+    if failed_indices.any?
+      raise Bolt::ParallelFailure.new(result, failed_indices)
+    end
+
+    result
+  end
+end
--- a/bolt-modules/boltlib/lib/puppet/functions/parallelize.rb
+++ b/bolt-modules/boltlib/lib/puppet/functions/parallelize.rb
@@ -37,45 +37,7 @@ Puppet::Functions.create_function(:parallelize, Puppet::Functions::InternalFunct
     executor.report_function_call(self.class.name)
 
     skein = data.each_with_index.map do |object, index|
-      fiber = Fiber.new do
-        # Create the new scope
-        newscope = Puppet::Parser::Scope.new(scope.compiler)
-        local = Puppet::Parser::Scope::LocalScope.new
-
-        # Compress the current scopes into a single vars hash to add to the new scope
-        current_scope = scope.effective_symtable(true)
-        until current_scope.nil?
-          current_scope.instance_variable_get(:@symbols)&.each_pair { |k, v| local[k] = v }
-          current_scope = current_scope.parent
-        end
-        newscope.push_ephemerals([local])
-
-        begin
-          result = catch(:return) do
-            args = { block.parameters[0][1].to_s => object }
-            block.closure.call_by_name_with_scope(newscope, args, true)
-          end
-
-          # If we got a return from the block, get it's value
-          # Otherwise the result is the last line from the block
-          result = result.value if result.is_a?(Puppet::Pops::Evaluator::Return)
-
-          # Validate the result is a PlanResult
-          unless Puppet::Pops::Types::TypeParser.singleton.parse('Boltlib::PlanResult').instance?(result)
-            raise Bolt::InvalidParallelResult.new(result.to_s, *Puppet::Pops::PuppetStack.top_of_stack)
-          end
-
-          result
-        rescue Puppet::PreformattedError => e
-          if e.cause.is_a?(Bolt::Error)
-            e.cause
-          else
-            raise e
-          end
-        end
-      end
-
-      Bolt::Yarn.new(fiber, index)
+      executor.create_yarn(scope, block, object, index)
     end
 
     result = executor.round_robin(skein)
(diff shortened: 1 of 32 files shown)


lucywyman:
This is the plan I've been using to test:
```
plan bolt(
  #TargetSpec $targets
) {
  out::message("Hello")
  $future = future() || {
    run_task('bolt', 'localhost', val => 'a', time => 5)
    run_task('bolt', 'localhost', val => 'b', time => 1)
    run_task('bolt', 'localhost', val => 'c', time => 1)
  }
  $f2 = future() || {
    run_task('bolt', 'localhost', val => 'd', time => 3)
  }
  run_task('bolt', 'localhost', val => 'e', time => 1)
  run_task('bolt', 'localhost', val => 'f', time => 10) 
  run_task('bolt', 'localhost', val => 'g', time => 1)
  run_task('bolt', 'localhost', val => 'h', time => 1)
}
```

With this task:
```
lucy ~/githubs/bolt (GH-2127) $ cat tasks/init.sh 
#!/bin/sh

sleep $PT_time
echo $PT_val
```
```
lucy ~/githubs/bolt (GH-2127) $ cat tasks/init.json 
{
  "description": "something",
  "parameters": {
    "val": {
      "type": "String"
    },
    "time": {
      "type": "Integer"
    }
  },
  "supports_noop": false
}
```

lib/bolt/executor.rb
```diff
@@ -347,6 +357,41 @@ def run_plan(scope, plan, params)
...
+        skein.each do |yarn|
+          # Check if stored a value
+          if yarn.alive?
+            yarn.resume
+          else
+            results[yarn.index] = yarn.value
+            skein.delete(yarn)
+          end
+        end
+        sleep(1) until @thread_completed
```
beechtom, line 388:
```suggestion
        skein.each do |yarn|
          # Check if stored a value
          if yarn.alive?
            yarn.resume
          else
            results[yarn.index] = yarn.value
            skein.delete(yarn)
          end
        end
```

I'm not sure what `@thread_completed` is doing, and it seems like it can result in things running slowly _and_ cause the plan to get stuck in an infinite loop. Since `@thread_completed` can only be set to `true` by the executor receiving a `node_result` event, the block would need to end with a function call that publishes that event. For example, the following would cause an infinite loop:

```
plan bolt () {
  $numbers = [1, 2, 3]
  $result = parallelize($numbers) |$number| {
    $number * 2
  }
  return $result
}
```

I also ran into an infinite loop a few other times with blocks that ended with a `run_*` function call.

bolt-modules/boltlib/lib/puppet/functions/parallelize.rb
```diff
@@ -0,0 +1,90 @@
...
+    unless Puppet[:tasks]
+      raise Puppet::ParseErrorWithIssue
+        .from_issue_and_stack(Bolt::PAL::Issues::PLAN_OPERATION_NOT_SUPPORTED_WHEN_COMPILING, action: 'parallelize')
+    end
+
+    executor = Puppet.lookup(:bolt_executor)
+    executor.report_function_call(self.class.name)
+
+    skein = data.each_with_index.map do |object, index|
+      fiber = Fiber.new do
```
nicklewis, line 37: I think it would make sense to have an executor method that does basically everything inside this block (the equivalent of the former `future()` function).

bolt-modules/boltlib/lib/puppet/functions/parallelize.rb
```diff
@@ -0,0 +1,90 @@
...
+
+require 'bolt/yarn'
+
+# Accepts an array and a block and maps the block to the array in parallel
+#
+# > **Note:** Not available in apply block
+Puppet::Functions.create_function(:parallelize, Puppet::Functions::InternalFunction) do
+  # Map a block to an array in parallel
+  # @param data The array to apply the block to
+  # @return [Array] An array of Result or Error objects from the block for each array element
```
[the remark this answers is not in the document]
hestonhoffman: That's a tough one. One approach would be to start at the other end:
"An array of PlanResult objects. The input from the input array returns a corresponding PlanResult object." I don't know that block is important here, as I think the user can infer from the previous descriptions that they're executing a block of code..."

hestonhoffman:
Probably a bit of a preemptive suggestion (and I might have missed it), but I'm assuming this is going to be introduced as experimental. In that case, we should probably include a "This function is experimental" wherever the function is surfaced in the docs.

lib/bolt/executor.rb
lucywyman: Hm - we need @thread_completed to tell when we're waiting for several tasks to complete so we don't thrash while waiting for several threads to finish. I think I can yield an argument from functions that yield the Fiber and then check for the argument in the loop, so that anything that doesn't yield back `round_robin` won't wait for the node_result call.

lucywyman: This is exactly the kind of thing I wanted to catch now! Thanks Tom!!

lib/bolt/executor.rb
```diff
@@ -347,6 +357,88 @@ def run_plan(scope, plan, params)
...
+
+    def handle_event(event)
+      case event[:type]
+      when :node_result
+        # Give the result a chance to return back to the function
+        # If we resume the fiber before the result makes it there, this will
+        # spin forever
+        #
+        # TODO: Is there a way to avoid this?
+        sleep(1)
```
beechtom, line 410:
```suggestion
        sleep(0.1)
```

A full second sleep seems a little long. Does it need to be this high?

lib/bolt/executor.rb
```diff
@@ -347,6 +357,88 @@ def run_plan(scope, plan, params)
...
+          if yarn.alive?
+            r = yarn.resume
+          else
+            results[yarn.index] = yarn.value
+            skein.delete(yarn)
+          end
+        end
+
+        next unless r == 'unfinished'
+        sleep(1) until @thread_completed || skein.empty?
```
beechtom, line 434:
```suggestion
        sleep(0.1) until @thread_completed || skein.empty?
```

Same as above.

beechtom (approved):
This looks really good, and everything seems to be working as expected. 👍 Just a couple small things to address.

lib/bolt/executor.rb
lucywyman: Oo, in rerunning this a few times I'm not seeing the same issue I was when I wrote this (which was that it would hang waiting for the `:node_result` signal. Unless you see the same behavior I'd like to remove it for now and we can reassess if it pops up again (good thing this is experimental :grimacing: )

lucywyman:
Closes #2127

beechtom (approved):

documentation/experimental_features.md
```diff
@@ -8,6 +8,99 @@ API may change, requiring the user to update their code or configuration. The
...
+  run_command('hostname', $targets)
+  run_command('whoami', $targets)
+}
+```
+
+The experimental `parallelize()` function accepts an array and a block, and runs the entire block on
+each array element in parallel. In the example above, target3 will have to wait for `hostname` to
+finish on target1 and target2 before it can run `whoami`. Inside a parallelize block targets will be
+able to run subsequent plan functions before all targets have finished each step. Let's see the same
+plan with a parallelize block:
```
hestonhoffman, line 33:
```suggestion
In the example above, `target3` has to wait for `hostname` to finish on `target1` and `target2` 
before it can run `whoami`. The experimental `parallelize()` function accepts an array and a block, 
and runs the entire block on each array element in parallel. Inside a parallelize block, targets can
run subsequent plan functions before all targets have finished each step. For example, here is the 
same plan with a parallelize block:
```

documentation/experimental_features.md
```diff
@@ -8,6 +8,99 @@ API may change, requiring the user to update their code or configuration. The
...
+
+  parallelize($ts) |$target| {
+    run_command('hostname', $target)
+    run_command('whoami', $target)
+  }
+}
+```
+
+Here, if target3 completes running `hostname` before target1 or target2 it can continue directly to
+running `whoami`.
```
hestonhoffman, line 49:
```suggestion
Here, if `target3` completes running `hostname` before `target1` or `target2`, it can continue directly to
running `whoami`.
```

documentation/experimental_features.md
```diff
@@ -8,6 +8,99 @@ API may change, requiring the user to update their code or configuration. The
...
+
+This functionality is particularly useful for plan functions that may take a long time on certain
+targets but not others, or for plans where some long running process may fail on a target but the
```
hestonhoffman, line 52:
```suggestion
targets but not on others, or for plans where some long running process may fail on a target but the
```

documentation/experimental_features.md
```diff
@@ -8,6 +8,99 @@ API may change, requiring the user to update their code or configuration. The
...
+plan author wants the plan to be able to continue quickly on successful targets.
+
+Within the parallelize block only certain functions are able to run in parallel. These functions are
+`run_command`, `run_task`, `run_task_with`, `run_script`, `upload_file`, and `download_file`. Other
+functions can be run from a parallelize block, but those functions will block execution
+on other targets until they complete. For example, in the following plan Bolt can start running
+`task2` and `task3` while `task1` is still executing. However, it cannot start `task4` while
+`out::message` is executing on any of the targets.
```
hestonhoffman, line 60:
```suggestion
Within the parallelize block, only the following functions can run in parallel: 
- `run_command`
- `run_task`
- `run_task_with`
- `run_script`
- `upload_file`
- `download_file`. 

You can run other functions from a parallelize block, but those functions will block execution
on other targets until they complete. For example, in the following plan, Bolt can start running
`task2` and `task3` while `task1` is still executing. However, it cannot start `task4` while
`out::message` is executing on any of the targets.
```

documentation/experimental_features.md
```diff
@@ -8,6 +8,99 @@ API may change, requiring the user to update their code or configuration. The
...
+
+# This will print ["Don't run the task on this target", "target2", "target3"]
+out::message($result)
+```
+
+If any step of the block errors, Bolt will stop executing the block for that target but
+will continue executing for all other targets from the input array. When the block finishes, if
+there is an error in the result array the plan will throw a `PlanFailure` as usual and will include
+the entire result array in the `details` key of the failure. If the block is wrapped in a
+`catch_errors()` block the `PlanFailure` will be caught and execution of the plan will continue.
```
hestonhoffman, line 102:
```suggestion
If any step of the block errors, Bolt stops executing the block for that target, but
continues executing for all other targets from the input array. When the block finishes, if
there is an error in the result array, the plan throws a `PlanFailure` and includes
the entire result array in the `details` key of the failure. If the block is wrapped in a
`catch_errors()` block, Bolt catches the `PlanFailure` and continues executing the plan.
```

hestonhoffman (approved):
LGTM! 👍

beechtom (changes requested):
Just needs a rebase and then we can merge :D

beechtom (approved):

PR #2192, opened 2020-09-17
(PE-30092) Dont write a bolt-debug.log when listing project content

donoghuc:
This commit updates bolt-server metadata endpoints for serving project content to not create a bolt-debug.log file at the root of the project directory.


donoghuc:
Given we mock out the project and config objects with instance doubles in our tests here I think we will have to just verify this in orch acceptance tests.

barriserloth (approved):
LGTM

PR #2198, opened 2020-09-21
Target env_vars parameter

faucct:
This makes it possible to set environment variables only for duration of bolt-script.
A valid usage example would be configuring a SSH tunnel options with a `http_proxy` env variable:

```
ssh-command:
- ssh
- "-L"
- "3128:proxy:3128"

env_vars:
  http_proxy: "http://localhost:3128"
```

As the variable only makes sense while the script is running, it is better to define it in target instead of exporting it via remote files.

The changes are mostly a copy-paste of `vars`. I hope, I didn't miss anything. I could also copy tests if necessary.

### DIFF
--- a/bolt-modules/boltlib/lib/puppet/datatypes/target.rb
+++ b/bolt-modules/boltlib/lib/puppet/datatypes/target.rb
@@ -16,6 +16,7 @@ Puppet::DataTypes.create_type('Target') do
       target_alias => { type => Optional[Variant[String[1], Array[String[1]]]], kind => given_or_derived },
       config => { type => Optional[Hash[String[1], Data]], kind => given_or_derived },
       vars => { type => Optional[Hash[String[1], Data]], kind => given_or_derived },
+      env_vars => { type => Optional[Hash[String[1], String[1]]], kind => given_or_derived },
       facts => { type => Optional[Hash[String[1], Data]], kind => given_or_derived },
       features => { type => Optional[Array[String[1]]], kind => given_or_derived },
       plugin_hooks => { type => Optional[Hash[String[1], Data]], kind => given_or_derived },
--- a/lib/bolt/apply_target.rb
+++ b/lib/bolt/apply_target.rb
@@ -2,7 +2,7 @@
 
 module Bolt
   class ApplyTarget
-    ATTRIBUTES = %i[uri name target_alias config vars facts features
+    ATTRIBUTES = %i[uri name target_alias config vars env_vars facts features
                     plugin_hooks resources safe_name].freeze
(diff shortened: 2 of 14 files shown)


adreyer:
I have two concerns with this. 

1.  It only impacts some actions, scripts and commands, which is not clear from is probably counterintuitive to users. Adding envar support for tasks is not desirable since tasks have specific other methods of accepting inputs.
2. The example use case is for variables specific to a script while the implementation applies to all scripts and commands.

faucct:
> I have two concerns with this.
> 
> 1. It only impacts some actions, scripts and commands, which is not clear from is probably counterintuitive to users. Adding envar support for tasks is not desirable since tasks have specific other methods of accepting inputs.

With my change you could decide whether to use any `http_proxy` hack or not during the definition of targets in an inventory file. Without the change I guess you would have to either implement tasks for http_proxy both being on and off, either somehow switch between them based on user parameter or even per target parameters if only some targets should use a proxy.

> 2. The example use case is for variables specific to a script while the implementation applies to all scripts and commands.

A `http_proxy` env variable is a tacit rule used by lots of libraries, including apt, lynx, wget, curl, perl LWP, perl WWW::Search, python urllib/urllib2, and etc. I believe making it possible to enable it for all scripts is a nice feature.

BTW, could you suggest a way to wrap all calls in a plan, so specific targets would have this variable? I have found this puppet option, but I am not sure, how can I set it.
https://github.com/puppetlabs/puppet/blob/e8d183a7774677331632dbbb6b83a51ca49b07d9/lib/puppet/defaults.rb#L525
Though for my specific needs, it would be even enough to have this variable set forever, as there will always be an available proxy host and there is no need for SSH tunnel. I guess I could set it in `/etc/environment`

lucywyman:
Hm, I think this puts us in a tough situation with tasks in general. We don't currently have a solution for provisioning target environments before running a task, and while this is a good case for that this lacks a per-task approach that might be needed later down the line. Unfortunately I think we don't want to paint ourselves into a corner by accepting this solution, and we also aren't ready to prioritize a more general solution particularly since we haven't had many other environment-provisioning use cases come up.

Is provisioning your environment with a plan or another tool an option before running the task? If you're ok with it I think the next steps for us are to open an issue with this desired workflow with the 'Request for Comment' label and see if anyone else has a similar use case. I also let our Project Manager know, so they know this is something in our backlog.

beechtom:
Closing since this is fairly outdated and we're not ready to commit to this particular implementation. If you are still interested in a similar solution, feel free to open an issue as Lucy mentioned above.

PR #2200, opened 2020-09-22
(docs) Clarify Hiera data loading docs

lucywyman:
This clarifies that while Bolt sets a default data dir and default Hiera
config file, it does not have any content in that Hiera config and will
not load any Hiera data unless the user configures it to.

!no-release-note

### DIFF
--- a/documentation/applying_manifest_blocks.md
+++ b/documentation/applying_manifest_blocks.md
@@ -240,12 +240,12 @@ plan do_thing() {
 ```
 
 Manifest block compilation can access Hiera data that you add to your Bolt
-configuration. The default location for Hiera config is `<project-directory>/hiera.yaml`.
+configuration. The default location for Hiera config is `<PROJECT DIRECTORY>/hiera.yaml`.
 You can change this with the `hiera-config` key in a Bolt config file or the
 `--hiera-config` CLI option.
 
 Following the Hiera 5 convention, the default data directory is relative to
-`hiera.yaml` at `<project-directory>/data`. For configuration file examples, see [Configuring
+`hiera.yaml` at `<PROJECT DIRECTORY>/data`. For configuration file examples, see [Configuring
 Hiera](https://puppet.com/docs/puppet/latest/hiera_config_yaml_5.html). Bolt 
 will not automatically load any Hiera data from this directory. If you want to load data from a
 data file you must set the relevant paths in your Hiera config.
@@ -263,7 +263,7 @@ under the `plan_hierarchy` key outside of apply blocks, and uses the the typical
 of apply blocks. This allows you to use your existing Hiera configs in Bolt plans without encountering 
 an error if interpolation exists and your plan tries to look up data outside of an apply block.
 
-For example, with this Hiera configuration at `<project>/hiera.yaml`:
+For example, with this Hiera configuration at `<PROJECT DIRECTORY>/hiera.yaml`:
 ```yaml
 version: 5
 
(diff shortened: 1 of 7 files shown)


documentation/applying_manifest_blocks.md
```diff
@@ -246,7 +246,9 @@ You can change this with the `hiera-config` key in a Bolt config file or the
 
 Following the Hiera 5 convention, the default data dir is relative to
```
hestonhoffman, line 247:
```suggestion
Following the Hiera 5 convention, the default data directory is relative to
```

documentation/applying_manifest_blocks.md
```diff
@@ -246,7 +246,9 @@ You can change this with the `hiera-config` key in a Bolt config file or the
...
 `hiera.yaml` at `$BOLTDIR/data`. For configfile examples, see [Configuring
```
hestonhoffman, line 248:
```suggestion
`hiera.yaml` at `$BOLTDIR/data`. For configuration file examples, see [Configuring
```

hestonhoffman: When we use `$BOLTDIR` here, are we talking specifically about an embedded project directory, or is $BOLTDIR an environment variable? If we're just talking about a Bolt project directory, then maybe we can just use <PROJECT_DIRECTORY>?

documentation/applying_manifest_blocks.md
```diff
@@ -246,7 +246,9 @@ You can change this with the `hiera-config` key in a Bolt config file or the
...
-Hiera](https://puppet.com/docs/puppet/latest/hiera_config_yaml_5.html).
+Hiera](https://puppet.com/docs/puppet/latest/hiera_config_yaml_5.html). Bolt 
+will not automatically load any Hiera data from this directory, you must
+set paths that you want loaded in your Hiera config.
```
hestonhoffman, line 251:
```suggestion
Hiera](https://puppet.com/docs/puppet/latest/hiera_config_yaml_5.html). Bolt 
does not automatically load Hiera data from this directory. If you want want to 
load data from a Hiera directory, you must set the relevant paths in your Hiera 
configuration file.
```

fccagou: There is 2 times want on line 250

hestonhoffman: Ah thanks!

documentation/applying_manifest_blocks.md
lucywyman: No, I think it's what we used to use for `<project_directory>`. Should it be all caps?

lucywyman: It looks like elsewhere we use PROJECT_NAME https://puppet.com/docs/bolt/latest/bolt_running_plans.html#how-bolt-locates-plans

lucywyman: And <project-directory> https://puppet.com/docs/bolt/latest/configuring_bolt.html#bolt-projectyaml

hestonhoffman: Good point. Let's standardize it as `<PROJECT_DIRECTORY>` so we're following Puppet style. We should change `<PROJECT_NAME>` anyway, because technically your project directory name and project name don't have to match.

lucywyman: Ended up with `<PROJECT DIRECTORY>` which seems clearest and inline with style guidelines.

hestonhoffman (approved):
Awesome!

PR #2202, opened 2020-09-22
(GH-2186) Add resolve and force options for module subcommand

### LINKED ISSUE #2186: Add --no-resolve option for 'bolt module install'
## Use Case
If a project configuration file include module declarations that are not present in the `Puppetfile`, Bolt will error. Currently, a user can tell Bolt to re-resolve the declarations and rewrite the `Puppetfile` using the `--force` option. However, there is no way for a user to tell Bolt to install the `Puppetfile` as-is without re-resolving the module declarations.

The `bolt module install` command should accept a `--no-resolve` option that will install the project's `Puppetfile` as-is without re-resolving module declarations.

## Questions
- Should `--force` be replaced with `--resolve`?

lucywyman:
This adds a new `--[no-]resolve` flag, which defaults to true, that will
allow users to install the Puppetfile without resolving the modules
enumerated in `modules` key in Bolt config. No release notes as
this will ship dark until the full workflow is implemented.

Closes #2186

!no-release-note

### DIFF
--- a/lib/bolt/cli.rb
+++ b/lib/bolt/cli.rb
@@ -478,7 +478,7 @@ module Bolt
         when 'add'
           code = add_project_module(options[:object], config.project)
         when 'install'
-          code = install_project_modules(config.project, options[:force])
+          code = install_project_modules(config.project, options[:force], options[:resolve])
         when 'generate-types'
           code = generate_types
         end
@@ -888,7 +888,7 @@ module Bolt
 
     # Installs modules declared in the project configuration file.
     #
-    def install_project_modules(project, force)
+    def install_project_modules(project, force, resolve)
       assert_project_file(project)
 
       unless project.modules
@@ -899,7 +899,11 @@ module Bolt
 
       installer = Bolt::ModuleInstaller.new(outputter, pal)
 
-      ok = installer.install(project.modules, project.puppetfile, project.managed_moduledir, force: force)
+      ok = installer.install(project.modules,
+                             project.puppetfile,
+                             project.managed_moduledir,
+                             force: force,
+                             resolve: resolve)
       ok ? 0 : 1
     end
 
@@ -911,7 +915,11 @@ module Bolt
       modules   = project.modules || []
       installer = Bolt::ModuleInstaller.new(outputter, pal)
 
-      ok = installer.add(name, modules, project.puppetfile, project.managed_moduledir, project.project_file)
+      ok = installer.add(name,
+                         modules,
+                         project.puppetfile,
+                         project.managed_moduledir,
+                         project.project_file)
       ok ? 0 : 1
     end
 
--- a/lib/bolt/module_installer.rb
+++ b/lib/bolt/module_installer.rb
@@ -94,7 +94,7 @@ module Bolt
 
     # Installs a project's module dependencies.
     #
-    def install(modules, path, moduledir, force: false)
+    def install(modules, path, moduledir, force: false, resolve: true)
       require 'bolt/puppetfile'
 
       puppetfile = Bolt::Puppetfile.new(modules)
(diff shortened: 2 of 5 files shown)


lib/bolt/module_installer.rb
```diff
@@ -94,23 +94,26 @@ def add(name, modules, puppetfile_path, moduledir, config_path)
...
-      if path.exist? && !force
-        assert_managed_puppetfile(puppetfile, path)
-      else
-        @outputter.print_message "Resolving module dependencies, this may take a moment"
-        puppetfile.resolve
-
-        @outputter.print_message "Writing Puppetfile at #{path}"
-        puppetfile.write(path)
+      # overwrite an existing Puppetfile with the '--force' option, or opt to
+      # install the Puppetfile as-is with --no-resolve
```
beechtom, line 106:
```suggestion
      # Overwrite an existing Puppetfile with the '--force' option, or opt to
      # install the Puppetfile as-is with --no-resolve.
```

lib/bolt/cli.rb
```diff
@@ -463,9 +463,9 @@ def execute(options)
       when 'module'
         case options[:action]
         when 'add'
-          code = add_project_module(options[:object], config.project)
+          code = add_project_module(options[:object], config.project, options[:force])
         when 'install'
-          code = install_project_modules(config.project, options[:force])
+          code = install_project_modules(config.project, options[:force], options[:resolve])
```
beechtom, line 468: This results in `bolt module install` never resolving module declarations, as `resolve` will be `nil` unless you use the `--[no]-resolve` flag. Since you're explicitly passing `nil` as a parameter, `Bolt::ModuleInstaller.install` will not default to `true`.

lib/bolt/module_installer.rb
```diff
@@ -28,7 +28,7 @@ def add(name, modules, puppetfile_path, moduledir, config_path)
       end
 
       # If the Puppetfile exists, make sure it's managed by Bolt.
-      if puppetfile_path.exist?
+      if puppetfile_path.exist? && !force
```
beechtom, line 31: I don't think we want to forcibly install here. When adding a single module, the installer resolves dependencies in one of two ways:

- Resolves the Puppetfile's modules with the new module
- Resolves the module declarations with the new module

If there is a mismatch between the module declarations and the Puppetfile, then we can't be sure what the user actually intends by trying to add a module. Are they trying to add a module to a manually edited Puppetfile? Then they should manually edit the Puppetfile instead of using the `module add` command. Are they trying to add a module to a Bolt-managed Puppetfile? Then they should be told to re-install their modules with `module install --force`.

beechtom: I think another option may be to skip the initial attempt to resolve with the Puppetfile's modules when forcibly adding a module, but I still think if you are using Bolt to manage your modules and want to use the `module add` command, then your Puppetfile should be managed by Bolt.

lib/bolt/module_installer.rb
lucywyman: This isn't the start of a sentence...

beechtom: D'oh. Sorry, I think my eyes skipped over the previous lines. :|

lib/bolt/module_installer.rb
lucywyman: I only initially added this because I saw that `force` is listed as an option for `module add` https://github.com/puppetlabs/bolt/blob/main/lib/bolt/bolt_option_parser.rb#L70 but to me it makes sense to have the force option available here to both add the module to the modules list and transition the Puppetfile to a managed Puppetfile, just like with module install. We would know the user is trying to add a module to a Bolt-managed Puppetfile when they ran with `--force`, as they simply wouldn't run this with that flag if they wanted to manually add a module to their Puppetfile.

beechtom: Ah, that's listed as an option because I originally added support for `--force` but then removed it for the above reasons.

If we move forward with `module add --force` then this would need to be updated to skip resolving with the Puppetfile's modules: https://github.com/puppetlabs/bolt/blob/main/lib/bolt/module_installer.rb#L71-L84

lucywyman: M'k - we can discuss at standup tomorrow

lib/bolt/module_installer.rb
```diff
@@ -144,8 +149,8 @@ def install_puppetfile(path, moduledir, config = {})
         raise Bolt::Error.new(
```
beechtom, line 149: Ugh, this error was updated as part of #2168 , but wasn't carried over in #2134 . Would you mind adding it back in here? https://github.com/puppetlabs/bolt/commit/b0db40014e28b4f07f78392554619dd894dba664#diff-b4ccc59dee0e054966d665d51e08d67eR895-R903

beechtom (commented):
Aside from the discussion tomorrow on `module add --force`, this lgtm.
I do still question whether `--force` should be replaced by `--resolve`, but I don't really have a strong opinion since there are pros/cons for both... That might be something we can get feedback on from users if we're interested.

beechtom (approved):

Issue #2184, opened 2020-09-16
Module management migration tool


adreyer:
## Use Case
In many cases we should be able to help users migrate to the new module management workflow. 

## Describe the Solution You Would Like
- update bolt project init to add `modules` key populated with any specified modules or empty.
- add a module migraton step to `bolt project migrate`

## Describe Alternatives You've Considered
A fully manual process.

## Questions
- Should we use an interactive interface? 
- Can we move `site-modules` to `modules` location?
   In some cases see below.
- Can we update git for the user?
   git out of scope for this release. 
- How can we determine which modules are direct dependencies?
   Ask the user interactively
- How can we handle `bolt.yaml` -> `bolt-project.yaml`?
   This is the first step of the project migrate command.
- Should this be `bolt project migrate`? does it need a flag? 
   start with bolt project migrate until it starts causing issues. 

## Solution
Cases:
- Standard Module Path
  - if standard module path and no Puppetfile:
        merge site-modules into modules
        remove explicitly set modulepath
  - if Puppetfile:
     query user for what modules in the puppetfile are direct dependencies and update `modules`
     recompute puppetfile and show the user the changes. 
     delete non-local modules from `modules`
     merge site-modules into modules.
     remove explicitly set module path
- NonStandard Module Path
   - refer user to docs.

PR #2203, opened 2020-09-22
(GH-2184) Update 'project migrate' for module management workflow

beechtom:
(GH-2184) Refactor ProjectMigrate class

This refactors the `ProjectMigrate` class so that the logic for each
step of the migration process is contained in a separate class.
Migration for inventory files is now handled by the
`Bolt::ProjectMigrate::Inventory` class, while migration for
configuration files is now handled by the `Bolt::ProjectMigrate::Config`
class.

The intent behind this refactor is to break up the `ProjectMigrate`
class into more discrete components, and have the class instead defer
the work to specialized migrators. This will also allow for adding a
migrator for project modules without making the `ProjectMigrate` class
even more monolithic.

!no-release-note

---

(GH-2184) Update 'project migrate' to migrate modules

This updates the `bolt project migrate` command to migrate a project's
modules to the new module management workflow. If the project already
configures `modules` or configures `modulepath` to a non-default
modulepath, then the migrator will do nothing.

If the project has a `Puppetfile`, the migrator will parse the file for
modules, and then prompt the user to select direct dependencies for the
project. Any modules selected as dependencies will be used to generate a
new `Puppetfile` that includes the modules and their dependencies. Bolt
will then install the `Puppetfile` to the managed moduledir `.modules`
and add the selected modules to the project configuration. Any modules
on the modulepath that are not selected as dependencies will be
consolidated into the `modules` directory, while all remaining
moduledirs will be deleted.

If the project does not have a `Puppetfile`, then Bolt will consolidate
all modules on the modulepath into the `modules` directory and delete
all remaining moduledirs.

This feature is feature flagged and is shipping dark until the module
management workflow is complete.

!no-release-note


lib/bolt/cli.rb
```diff
@@ -449,9 +449,8 @@ def execute(options)
         when 'init'
           code = initialize_project
         when 'migrate'
-          inv = config.inventoryfile
-          path = config.project.path
-          code = Bolt::ProjectMigrate.new(path, outputter, inv).migrate_project
+          ok = Bolt::ProjectMigrate.new(config, outputter).migrate_project
+          code = ok ? 0 : 1
```
lucywyman, line 453: I think since this is called directly from the CLI class it's appropriate for the migrate_project method to just return 0 or 1. It's only methods that aren't directly returned by the CLI where that doesn't make sense.

lib/bolt/outputter.rb
```diff
@@ -27,6 +27,10 @@ def indent(indent, string)
       string.gsub(/^/, indent.to_s)
     end
 
+    def wrap(string, width = 80)
```
lucywyman, line 30: This probably just needs to be in the human outputter, since we shouldn't wrap JSON output or logs.

beechtom: Whoops, that shouldn't be there at all. Removed.

lib/bolt/project_migrate.rb
```diff
@@ -1,138 +1,61 @@
...
+    # Migrates the project-level configuration file to the latest version.
+    #
+    private def migrate_config
+      migrator = Bolt::ProjectMigrate::Config.new(@outputter)
+
+      migrator.migrate(
+        @config.project.config_file,
+        @config.project.project_file,
+        @config.inventoryfile || @config.project.inventory_file,
+        @config.project.path + '.bolt-bak'
```
lucywyman, line 35: I think this can be an attribute of the project

lib/bolt/project_migrate/base.rb
```diff
@@ -0,0 +1,52 @@
...
+require 'bolt/error'
+
+module Bolt
+  class ProjectMigrate
+    class Base
+      def initialize(outputter)
+        @outputter = outputter
+      end
+
+      protected def print_step(step)
```
lucywyman, line 13: This seems like it belongs in the human outputter class, especially since we're already passing the outputter around. That's where we handle things like formatting and printing, and would ensure we don't have to redefine `indent` here. I think indent can use the `Bolt::Outputter#indent` method, and wrap can go in that `Bolt::Outputter` as well.

lib/bolt/project_migrate/base.rb
```diff
@@ -0,0 +1,52 @@
...
+      protected def wrap(string)
+        string.gsub(/(.{1,76})(\s+|\Z)/, "\\1\n")
+      end
+
+      protected def indent(string, size = 2)
+        indent = "\s" * size
+        string.gsub(/^/, indent.to_s)
+      end
+
+      protected def backup_file(origin_path, backup_dir)
```
lucywyman, line 30: This could potentially go in the Project class if we wanted to get rid of this base class altogether, but I'm also fine with leaving it here.

lib/bolt/project_migrate/inventory.rb
```diff
@@ -0,0 +1,69 @@
+# frozen_string_literal: true
+
+require 'bolt/project_migrate/base'
+
+module Bolt
+  class ProjectMigrate
+    class Inventory < Base
+      def migrate(inventory_file, backup_dir)
+        @outputter.print_message "Migrating inventory\n\n"
```
lucywyman, line 9: Hm, I'm not actually sure we should print everything we try to do if we don't need to do it, mostly because it's a lot to read (and admittedly because in my terminal `\n` seems to be expanding weirdly). For example, this is the output if there's nothing to do:
```
lucy ~/githubs/bolt/project ((0da2c5b...)) $ be bolt project migrate
Migrating inventory

  → Bolt inventory is already on the latest version. Nothing to do.


Migrating project configuration

  → Found project configuration file

    /home/lucy/githubs/bolt/project/bolt-project.yaml. Nothing to do.


Migrating project modules

  → Project is already using module management workflow. Nothing to do.


Project successfully migrated
```
Which seems like a lot of scanning just to be like "Ok it didn't do anything".

beechtom: Ya, I definitely think cutting down on the messages is warranted. :)

lib/bolt/outputter.rb
lucywyman: Uhh, I still see it?

beechtom: I might have misinterpreted this comment https://github.com/puppetlabs/bolt/pull/2203#discussion_r493065255 and put it back. I'll move it to `Bolt::Outputter::Human`.

lib/bolt/project_migrator/modules.rb
```diff
@@ -0,0 +1,212 @@
...
+        modules = select_modules(puppetfile.modules)
+
+        # Create new Puppetfile object
+        puppetfile = Bolt::Puppetfile.new(modules)
+
+        # Attempt to resolve dependencies
+        begin
+          @outputter.print_message('')
+          @outputter.print_migrate_step("Resolving module dependencies, this may take a moment")
+          puppetfile.resolve
```
lucywyman, line 72: Hm, this seems to stack trace if we can't resolve the Puppetfile - for example, I tried migrating the Bolt puppetfile which errored on the local modules (canary, aggregate, etc.):
```
lucy ~/githubs/bolt/project ((e445890...)) $ be bolt project migrate
Migrating project modules

  → Parsing Puppetfile at /home/lucy/githubs/bolt/project/Puppetfile

  → Select modules that are direct dependencies of your project. Bolt will
    automatically manage dependencies for each module selected, so do not select
    a module's dependencies unless you use content from it directly in your
    project.

    Select all modules? ([y]es/[n]o) y

  → Resolving module dependencies, this may take a moment

bundler: failed to load command: bolt (/home/lucy/githubs/bolt/vendor/bundle/ruby/2.6.0/bin/bolt)
ArgumentError: Module name (-canary) must match either 'modulename' or 'owner/modulename'
  /home/lucy/githubs/bolt/vendor/bundle/ruby/2.6.0/gems/puppetfile-resolver-0.4.0/lib/puppetfile-resolver/puppetfile/base_module.rb:57:in `parse_title'
  /home/lucy/githubs/bolt/vendor/bundle/ruby/2.6.0/gems/puppetfile-resolver-0.4.0/lib/puppetfile-resolver/puppetfile/base_module.rb:39:in `initialize'
  /home/lucy/githubs/bolt/vendor/bundle/ruby/2.6.0/gems/puppetfile-resolver-0.4.0/lib/puppetfile-resolver/puppetfile/forge_module.rb:9:in `initialize'
  /home/lucy/githubs/bolt/lib/bolt/puppetfile.rb:76:in `new'
  /home/lucy/githubs/bolt/lib/bolt/puppetfile.rb:76:in `block in resolve'
  /home/lucy/.rvm/rubies/ruby-2.6.3/lib/ruby/2.6.0/set.rb:338:in `each_key'
  /home/lucy/.rvm/rubies/ruby-2.6.3/lib/ruby/2.6.0/set.rb:338:in `each'
  /home/lucy/githubs/bolt/lib/bolt/puppetfile.rb:74:in `resolve'
  /home/lucy/githubs/bolt/lib/bolt/project_migrator/modules.rb:72:in `migrate_modules_from_puppetfile'
  /home/lucy/githubs/bolt/lib/bolt/project_migrator/modules.rb:27:in `migrate'
  /home/lucy/githubs/bolt/lib/bolt/project_migrator.rb:55:in `migrate_modules'
  /home/lucy/githubs/bolt/lib/bolt/project_migrator.rb:15:in `migrate'
  /home/lucy/githubs/bolt/lib/bolt/cli.rb:465:in `execute'
  /home/lucy/githubs/bolt/exe/bolt:11:in `<top (required)>'
  /home/lucy/githubs/bolt/vendor/bundle/ruby/2.6.0/bin/bolt:23:in `load'
  /home/lucy/githubs/bolt/vendor/bundle/ruby/2.6.0/bin/bolt:23:in `<top (required)>'
```
We should catch that and return just the error.

lib/bolt/project_migrator.rb
```diff
@@ -0,0 +1,61 @@
...
+
+module Bolt
+  class ProjectMigrator
+    def initialize(config, outputter)
+      @config    = config
+      @outputter = outputter
+    end
+
+    def migrate
+      ok = migrate_inventory && migrate_config && migrate_modules
```
lucywyman, line 15: I think we only want to do `migrate_modules` if BOLT_MODULE_FEATURE is enabled, for now.

lib/bolt/project_migrator/modules.rb
beechtom: Huh, that's weird... it should be catching all errors during the parse/resolve steps. It might be when it's building the document for resolving. Good find!

beechtom: Error stems from `Bolt::Puppetfile.parse` assuming all parsed modules are Forge modules. Updated the parser to error when a non-Forge module is present, unless the `skip_unsupported_modules` flag is set, in which case it will just ignore the unsupported modules. This is primarily to allow `project migrate` to still migrate a Puppetfile that has non-Forge modules.

lib/bolt/project_migrator/inventory.rb
```diff
@@ -0,0 +1,66 @@
...
+        data.delete('version') if data['version'] != 2
+        migrated = migrate_group(data)
+
+        return true unless migrated
+
+        @outputter.print_message "Migrating inventory\n\n"
+
+        backup_file(inventory_file, backup_dir)
+
+        if File.write(inventory_file, data.to_yaml)
```
lucywyman, line 29: `File.write` returns the number of bytes written, which will always be an integer unless an error is raised writing the file. We should try writing and rescue an error instead of this `if...else` statement:
```
lucy ~/githubs/bolt/project ((570719b...)) $ be bolt project migrate
Migrating inventory

  → Backing up inventory.yaml from
    /home/lucy/githubs/bolt/project/inventory.yaml to
    /home/lucy/githubs/bolt/project/.bolt-bak/inventory.yaml.20200924_131921810.bak

bundler: failed to load command: bolt (/home/lucy/githubs/bolt/vendor/bundle/ruby/2.6.0/bin/bolt)
Errno::EACCES: Permission denied @ rb_sysopen - /home/lucy/githubs/bolt/project/inventory.yaml
  /home/lucy/githubs/bolt/lib/bolt/project_migrator/inventory.rb:29:in `write'
  /home/lucy/githubs/bolt/lib/bolt/project_migrator/inventory.rb:29:in `inventory_1_to_2'
  /home/lucy/githubs/bolt/lib/bolt/project_migrator/inventory.rb:9:in `migrate'
  /home/lucy/githubs/bolt/lib/bolt/project_migrator.rb:48:in `migrate_inventory'
  /home/lucy/githubs/bolt/lib/bolt/project_migrator.rb:15:in `migrate'
  /home/lucy/githubs/bolt/lib/bolt/cli.rb:465:in `execute'
  /home/lucy/githubs/bolt/exe/bolt:11:in `<top (required)>'
  /home/lucy/githubs/bolt/vendor/bundle/ruby/2.6.0/bin/bolt:23:in `load'
  /home/lucy/githubs/bolt/vendor/bundle/ruby/2.6.0/bin/bolt:23:in `<top (required)>'
```
```

File: lib/bolt/project_migrator.rb [hunk 10]
```diff
@@ -0,0 +1,65 @@
...
+    def initialize(config, outputter)
+      @config    = config
+      @outputter = outputter
+    end
+
+    def migrate
+      ok = migrate_inventory && migrate_config
+
+      if ok && ENV['BOLT_MODULE_FEATURE']
+        ok = migrate_modules
```
lucywyman, line 18: This should be a new `modules_ok` var so we don't overwrite if any of the previous migrations failed. Then on line 21, should be `if ok && modules_ok`. Right now this will print that the project migrated successfully even if migrate_inventory or migrate_config returns false.

beechtom: I don't think that's accurate. if either `migrate_inventory` or `migrate_config` is `false`, then `migrate_modules` will never be called since the condition `ok && ENV['BOLT_MODULE_FEATURE']` is never met.

lucywyman: Oh yes, I missed the 'ok &&' conditional.

lib/bolt/project_migrator/modules.rb
```diff
@@ -0,0 +1,219 @@
...
+          @outputter.print_prompt(prompt)
+          response = $stdin.gets.to_s.downcase.chomp
+          return choices[response] if choices.key?(response)
+        end
+      end
+
+      # Consolidates all modules on the modulepath to 'modules'.
+      #
+      private def consolidate_modules(modulepath)
+        @outputter.print_migrate_step("Consolidating non-managed modules at #{modulepath.first}")
```
lucywyman, line 157: It might be clearer to say 'Moving modules in 'site-modules' and 'site' to the 'modules' directory'

lib/bolt/project_migrator/modules.rb
```diff
@@ -0,0 +1,222 @@
...
+          @outputter.print_migrate_step(
+            "stdin is not a tty, unable to automatically migrate project modules."\
+            "To migrate project modules manually, see http://pup.pt/bolt-modules"
+          )
+          return false
+        end
+
+        begin
+          @outputter.print_migrate_step("Parsing Puppetfile at #{puppetfile_path}")
+          puppetfile = Bolt::Puppetfile.parse(puppetfile_path, skip_unsupported_modules: true)
```
lucywyman, line 56: This just deletes local modules, which *definitely* doesn't seem like what we want. I think erroring here if we can't mirgate modules is the right thing to do.

lib/bolt/project_migrator/modules.rb
```diff
@@ -0,0 +1,222 @@
...
+          'y'   => true,
+          'yes' => true,
+          'n'   => false,
+          'no'  => false
+        }
+
+        loop do
+          @outputter.print_prompt(prompt)
+          response = $stdin.gets.to_s.downcase.chomp
+          return choices[response] if choices.key?(response)
```
lucywyman, line 150: This is a small thing but it might be nice to have an else here if we get anything other than 'y | yes | n | no':
```
    Select puppetlabs-ruby_task_helper? ([y]es/[n]o) y
    Select puppetlabs-ruby_plugin_helper? ([y]es/[n]o) y
    Select puppetlabs-stdlib? ([y]es/[n]o) f
    Select puppetlabs-stdlib? ([y]es/[n]o) f
    Select puppetlabs-stdlib? ([y]es/[n]o) 
    Select puppetlabs-stdlib? ([y]es/[n]o) f
    Select puppetlabs-stdlib? ([y]es/[n]o) f
    Select puppetlabs-stdlib? ([y]es/[n]o) f
    Select puppetlabs-stdlib? ([y]es/[n]o) f
    Select puppetlabs-stdlib? ([y]es/[n]o) f
    Select puppetlabs-stdlib? ([y]es/[n]o) f
    Select puppetlabs-stdlib? ([y]es/[n]o) yno
    Select puppetlabs-stdlib? ([y]es/[n]o) y
    Select puppetlabs-aws_inventory? ([y]es/[n]o) y
```

Just something in red that's `'Please use 'y' or 'n''` kind of thing

lib/bolt/project_migrator/modules.rb
```diff
@@ -0,0 +1,222 @@
...
+          message += puppetfile.modules.map(&:to_spec).join("\n").to_s
+          @outputter.print_migrate_step(message)
+
+          # Write Puppetfile
+          @outputter.print_migrate_step("Updating Puppetfile at #{puppetfile_path}")
+          puppetfile.write(puppetfile_path)
+
+          # Install Puppetfile
+          @outputter.print_migrate_step("Syncing modules from #{puppetfile_path} to #{moduledir}")
+          Bolt::Puppetfile::Installer.new({}).install(puppetfile_path, moduledir)
```
lucywyman, line 107: Should we actually be threading config through here?

lib/bolt/project_migrator/modules.rb
```diff
@@ -0,0 +1,222 @@
...
+        if modulepath != project.modulepath
+          @outputter.print_migrate_step(
+            "Project has a non-default configured modulepath, unable to automatically "\
+            "migrate project modules. To migrate project modules manually, see "\
+            "http://pup.pt/bolt-modules"
+          )
+          true
+        # Migrate modules from Puppetfile
+        elsif File.exist?(puppetfile)
+          migrate_modules_from_puppetfile(config, puppetfile, moduledir, modulepath)
```
lucywyman, line 27: It seems safer to pass `project.modulepath` here :sparkles: just in case :sparkles:

lib/bolt/project_migrator/modules.rb
```diff
@@ -0,0 +1,222 @@
...
+        loop do
+          @outputter.print_prompt(prompt)
+          response = $stdin.gets.to_s.downcase.chomp
+          return choices[response] if choices.key?(response)
+        end
+      end
+
+      # Consolidates all modules on the modulepath to 'modules'.
+      #
+      private def consolidate_modules(modulepath)
```
lucywyman, line 156: We only actually get here if the modulepath is exactly the old default modulepath, right? I actually think that might be good to hard code here, since `project.modulepath` will get updated to just 'modules' after 3.0 and we'll still want to check for site-module and site here. That also ensure we don't accidentally move any other paths if we make changes elsewhere in the codebase.

lib/bolt/project_migrator/modules.rb
```diff
@@ -0,0 +1,222 @@
...
+        moduledir, *sources = modulepath
+
+        sources.select! { |source| Dir.exist?(source) }
+
+        if sources.any?
+          @outputter.print_migrate_step(
+            "Moving modules from #{sources.join(', ')} to #{modulepath.first}"
```
lucywyman, line 163:
```suggestion
            "Moving modules from #{sources.join(', ')} to #{moduledir}"
```

lib/bolt/project_migrator/modules.rb
```diff
@@ -0,0 +1,222 @@
...
+      end
+
+      # Consolidates all modules on the modulepath to 'modules'.
+      #
+      private def consolidate_modules(modulepath)
+        moduledir, *sources = modulepath
+
+        sources.select! { |source| Dir.exist?(source) }
+
+        if sources.any?
```
lucywyman, line 161: We shouldn't need to do this check.

lib/bolt/project_migrator/modules.rb
beechtom: The puppetfile config? No. `puppetfile-resolver` only handles resolving from the Puppet Forge. We shouldn't install modules from any other Forge.

lib/bolt/project_migrator/modules.rb
beechtom: Having this check prevents Bolt from saying it's moving any modules when it's not, and from possibly creating an empty `modules` directory.

lib/bolt/project_migrator/modules.rb
lucywyman: This is still 'modulepath' it looks like?

lib/bolt/project_migrator/modules.rb
beechtom: It should just move local modules into `modules`... Anything that isn't parsed _and_ selected as a managed module will just be moved. The only modules that are deleted are those that are parsed and selected as managed modules.

lib/bolt/project_migrator/modules.rb
lucywyman: In thinking more about this, I can't think of a scenario where we want to use this for any modulepath other than `['modules', 'site-modules', 'site']`. Setting that as a constant in this class and using it directly instead of passing the modulepath around would be more clear, would simplify the signatures, and would prevent us from accidentally operating on any other modulepaths unintentionally. I think it'd also make the class more readable and easier to follow.

lucywyman: Obvious caveat we still need to check that the user modulepath is that modulepath, and exit if not. I think `moduledir` can also be hardcoded as 'modules' ?

beechtom: I think using a hardcoded modulepath is fine here, but I don't think having it as a constant simplifies much, as the directories need to be expanded relative to the project path.

Updated.

lib/bolt/project_migrator/modules.rb
lucywyman: Ah, ok, I see what  Idid - project migrate then puppetfile install. I'm not sure how common that will be - do we want to raise an error if you have the 'modules' key and try to puppetfile install

beechtom: I think this PR just needed a rebase. The work in #2189 disables `puppetfile install` if `modules` is configured.

lib/bolt/project_migrator/modules.rb
lucywyman: I don't think this got addressed? If there's a valid case other than the old default path we can keep this, I just can't think of any other case where we'd use this class.

PR #2206, opened 2020-09-23
(maint) Update docs on Puppet logging functions

barriserloth:
This commit updates the documentation around using Puppet's built-in
logging functions within Bolt plans. Since the mapping for Puppet's
logging was changed, the docs should be updated to match the new levels
that logging functions will log at.

### DIFF
--- a/documentation/writing_plans.md
+++ b/documentation/writing_plans.md
@@ -822,11 +822,22 @@ to the log file.
 
 ### Puppet log functions
 
-To generate log messages from a plan, use the Puppet log function that corresponds to the level you
-want to track: `error`, `warn`, `notice`, `info`, or `debug`. Configure the log level for both log
-files and console logging in `bolt-project.yaml`.  The default log level for the console is `warn`.
-For log files, the default is `notice`.  Use the `--log-level debug` flag to set the console log
-level to `debug` for a single run.
+To generate log messages from a plan, you can use Puppet's built-in log functions and configure
+your Bolt log level accordingly. Be aware that Bolt logs messages at slightly different levels than
+Puppet. Puppet log levels map to Bolt log levels in the following way:
+
+| Puppet log level | Bolt log level |
+| --- | --- |
+| `debug` | `trace` |
+| `info` | `debug` |
+| `notice` | `info` |
+| `warning` | `warn` |
+| `err` | `error` |
+
+Configure the log level for both log files and console logging in `bolt-project.yaml`. The default
+log level for both the console and log files is `warn`. Use the `--log-level debug` flag to set
+the console log level to `debug` for a single run.  For more information on configuring log
+levels, see [Logs](logs.md).
 
 ### Default action logging
 


barriserloth:
I _think_ the info in here is correct, but please let me know if the phrasing is bad or if something's wrong!

documentation/writing_plans.md
```diff
@@ -822,11 +822,20 @@ to the log file.
...
+To generate log messages from a plan, you can use Puppet's built-in log functions. However, within
+Bolt, these log at slightly different levels than in Puppet (generally, the levels are demoted 
+by one). The new levels are:
+```
+debug: Logged at trace level
+info: Logged at debug level
+notice: Logged at info level
+warning: Logged at warn level
+err: Logged at error level
+```
```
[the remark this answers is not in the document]
barriserloth: That does seem better, thanks!

documentation/writing_plans.md
```diff
@@ -822,11 +822,21 @@ to the log file.
 
 ### Puppet log functions
```
hestonhoffman, line 823: So is this section specifically for logging Puppety stuff in Bolt? Or for logging from plans in general?

barriserloth: I was under the impression that it was for using specifically built-in puppet functions as a way to log from plans, does that sound right? I think there are other ways to log like `out::message` (which I guess technically doesn't log, but does output things) as well that aren't affected by the difference between Puppet vs. Bolt log levels

hestonhoffman: Reading through the doc, it looks like this is just to log from plans in general. The section should really just tell people how to use plan logs and then link off to log.md. Let's move the comparison between puppet log files and Bolt log files to logs.md

hestonhoffman: I think I took too long to refresh and missed your comment @barriserloth. I'll drop you a line tomorrow

barriserloth: Sounds good! Thanks for the feedback so far, I appreciate the help

hestonhoffman (commented):
Thanks Barr!

barriserloth:
@hestonhoffman Thanks for the review, I think the text should be updated now!

hestonhoffman (approved):
LGTM!

documentation/writing_plans.md
```diff
@@ -822,11 +822,22 @@ to the log file.
...
+| Puppet log level | Bolt log level |
+| --- | --- |
+| `debug` | `trace` |
+| `info` | `debug` |
+| `notice` | `info` |
+| `warning` | `warn` |
+| `err` | `error` |
+
+Configure the log level for both log files and console logging in `bolt-project.yaml`. The default
+log level for both the console and log files is `warn`. Use the `--log-level debug` flag to set
```
donoghuc, line 838: Suggesting users set the log level to debug in this section to surface puppet log level debug seems confusing (and  wrong). We show in the table how we "demote" debug to trace in puppet. So if users want to see *puppet* debug level logs in the console they would need to set the log level to trace.

donoghuc: @hestonhoffman @barriserloth

hestonhoffman: Okay, let me take another stab at this. I'll open a separate PR and ping both of you

barriserloth: Ah, ok. That was part of the docs already, and so I didn't mess with it

PR #2211, opened 2020-09-24
(GH-2210) Configure 'modules' with 'bolt project init'

### LINKED ISSUE #2210: Configure 'modules' with 'bolt project init'
## Use Case
To simplify the process of getting started with the module workflow, the `bolt project init` command should be updated to configure `modules` in `bolt-project.yaml`.

- If run with `--modules`, `modules` should have an array of module declarations
- If run without `--modules`, `modules` should be an empty array
- If running in a directory that already has a `bolt-project.yaml`, but no `Puppetfile`, the `bolt-project.yaml` should be updated to configure `modules`

## Questions
- Updating an existing project config file when there is no `Puppetfile` seems a bit odd. Do we want to take this opportunity to have `bolt project init` only run if there is no `bolt-project.yaml`? Maybe not until 3.0, as this is a breaking change?

## Additional Context
- We should not merge this change until the module workflow is stable.

beechtom:
This updates the `bolt project init` command to configure `modules` in
`bolt-project.yaml`, allowing users with new projects to immediately
start using the module workflow.

- If `bolt project init` is run, `modules` will be an empty array `[]`
- If `bolt project init --modules` is run, `modules` will be an array of
  declarations for the specified modules

This also updates the command so that it can only be run once per
project. Previously, if a project had a configuration file, but no
`Puppetfile`, the `bolt project init --modules` command would still
install modules. Now, Bolt will error in this case and display a helpful
error that the user should instead use the `bolt module add` command.

Closes #2210 

!feature

* **Configure `modules` with `bolt project init`**
  ([#2110](https://github.com/puppetlabs/bolt/issues/2210))

  The `bolt project init` command will now configure the `modules` key
  in the `bolt-project.yaml` file, enabling the `bolt module` command.


.rubocop.yml
```diff
@@ -239,6 +239,11 @@ Style/RedundantRegexpEscape:
 Style/RedundantSelfAssignment:
   Enabled: true
 
+Style/RegexpLiteral:
+  Enabled: true
```
nicklewis, line 243: It looks like we want `EnforcedStyle: mixed` here? That will enforce the use of `//` for single line literals and `%r{}` for multi line.

nicklewis (approved):

lib/bolt/cli.rb
```diff
@@ -818,71 +818,81 @@ def generate_types
     end
 
     # Initializes a specified directory as a Bolt project and installs any modules
-    # specified by the user, along with their dependencies
+    # specified by the user, along with their dependencies.
+    #
     def initialize_project
```
lucywyman, line 823: I think it's probably time to move this into it's own class, similar to ProjectMigrate. I think it might make sense to combine those classes into a ProjectManager class that both initializes and migrates projects, but I don't think it'd be wrong per se to keep them separate, just more files.

lucywyman (commented):
This also needs a rebase :P

lib/bolt/cli.rb
beechtom: Updated!

lib/bolt/project_manager.rb
```diff
@@ -0,0 +1,159 @@
...
+            "Found existing project directory with #{config.basename} at #{project}, "\
+            "unable to initialize project.",
+            'bolt/existing-project-error'
+          )
+        end
+      elsif old_config.exist?
+        raise Bolt::Error.new(
+          "Found existing project directory with #{old_config.basename} at #{project}, "\
+          "unable to initialize project. #{old_config.basename} is deprecated. To "\
+          "update the project to current best practices, run 'bolt project migrate'.",
```
lucywyman, line 47: This should calculate the command for Powershell

spec/bolt/cli_spec.rb
```diff
@@ -2698,123 +2698,6 @@ def stub_config(file_content = {})
     end
   end
 
-  describe 'project' do
```
lucywyman, line 2701: Hmm am I missing where these got replaced? They still seem useful.

beechtom: https://github.com/puppetlabs/bolt/pull/2211/files#diff-7fadf6af77342a81135f665eb01ac144c3eb19193ec8761e0143e723a85e9ae2R34-R122

lucywyman: Oh I see. Those are unit tests though - we should have integration tests, either here or in the integration tests proper that cover these use cases.

beechtom: Those tests are 1:1 what was in cli_spec, granted they're rewritten to use `with_project`. I'm not sure what benefit having multiple integration tests gives us over the unit tests, since the CLI just instantiates a `ProjectManager` and calls `create`. I can add one that just makes sure a project name and `--modules` are passed along, but outside of that it'd just be duplicating tests.

lucywyman: Ah, I missed that we were making a project around each example. You're right - I think adding a happy-path integration test would be good to ensure calling from the CLI works, but everything else is already tested.

beechtom: Updated with a happy path integration test. 👍

spec/integration/project_manager_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+  include BoltSpec::Integration
+  include BoltSpec::Project
+
+  context 'creating a project' do
+    let(:command) { %w[project init myproject --modules puppetlabs-yaml] }
+
+    # Execute from a temporary project directory.
+    around(:each) do |example|
+      with_project do
+        Dir.chdir(project.path) do
```
beechtom, line 16: I'm wondering if we should just make this `Dir.chdir` part of `with_project` by default...

lucywyman: Could be a separate `in_project` function that uses `with_project` then does the chdir

PR #2213, opened 2020-09-24
(maint) docs - Clarify puppet log functions

hestonhoffman:
Had a chat with Cas and Barr about this. Because puppet log functions in Bolt do not map 1:1, it seems better to recommend out::message instead. To try and make that a bit clearer, I've simplified the hierarchy a bit and moved the bit about Puppet log functions down a bit. I also added this info to the apply doc, seeing as though people who are used to applying Puppet code might be confused about why their log functions aren't working as expected.

### DIFF
--- a/documentation/applying_manifest_blocks.md
+++ b/documentation/applying_manifest_blocks.md
@@ -292,6 +292,20 @@ plan plan_lookup(
 }
 ```
 
+## Puppet log functions in Bolt
+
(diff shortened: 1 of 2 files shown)


barriserloth (approved):
I think this makes sense, but also I imagine I have less context around Bolt logging than Cas or Bolt team members

PR #2214, opened 2020-09-25
(GH-2135) Add Powershell 'module' subcommands and un-feature-flag

### LINKED ISSUE #2135: Add PowerShell cmdlet Install-BoltModule and Add(?)-BoltModule
## Use Case
Support `bolt module install` and `bolt module add` in the Powershell cmdlet. We should wait for `bolt module add` to do this to avoid needing to feature-flag the addition.

As part of this we should also unfeature flag the modules feature by removing all instances where we check `BOLT_MODULE_FEATURE`, including where we set it in tests.

Blocked by #2134

lucywyman:
### Add 'module' subcommands to pwsh cmdlets module
This adds the `bolt module` subcommands to the Powershell cmdlets module
as `Add-BoltModule`, `Register-BoltModuleTypes`, `Install-BoltModule`,
and `Get-BoltModule`. This call directly into the respective `bolt
module` subcommands and should exhibit the same behavior.

This additionally fixes up the parameters to some Bolt commands in the
powershell module, ensuring that parameters are only included if they
are valid and marking them as optional or required appropriately.
    
Closes #2135
    
!no-release-note

---
### Remove BOLT_MODULE feature flag
This removes the feature flag for the `bolt module` command, now that
the workflow is ready to be marked experimental and shipped. The module
workflow is now only enabled by the presence of the `modules` key in a
`bolt-project.yaml` file.
    
!feature
    
* **New Bolt module subcommand**
    
   The new 'bolt module' subcommand and 'modules' key in project config
   can be used to manage a project's module dependencies, including
   resolving dependencies and version ranges. Use 'bolt project migrate' to
   migrate to the new workflow. See http://pup.pt/bolt-modules for
   documentation. This feature is experimental.

### DIFF
--- a/lib/bolt/project_migrator.rb
+++ b/lib/bolt/project_migrator.rb
@@ -31,11 +31,7 @@ module Bolt
 
       @outputter.print_message('')
 
-      ok = migrate_inventory && migrate_config
-
-      if ok && ENV['BOLT_MODULE_FEATURE']
-        ok = migrate_modules
-      end
+      ok = migrate_inventory && migrate_config && migrate_modules
 
       if ok
         @outputter.print_message("Project successfully migrated")
--- a/rakelib/pwsh.rake
+++ b/rakelib/pwsh.rake
@@ -9,6 +9,7 @@ namespace :pwsh do
   task :generate_module do
     # https://docs.microsoft.com/en-us/powershell/scripting/developer/cmdlet/approved-verbs-for-windows-powershell-commands?view=powershell-7
     @pwsh_verbs = {
+      'add'        => 'Add',
       'apply'      => 'Invoke',
       'convert'    => 'Convert',
       'createkeys' => 'New',
@@ -25,17 +26,21 @@ namespace :pwsh do
     }
 
     @hardcoded_cmdlets = {
-      'createkeys' => {
+      'secret:createkeys' => {
         'verb' => 'New',
         'noun' => 'BoltSecretKey'
       },
-      'show-modules' => {
+      'puppetfile:show-modules' => {
         'verb' => 'Get',
         'noun' => 'BoltPuppetfileModules'
       },
-      'generate-types' => {
+      'puppetfile:generate-types' => {
         'verb' => 'Register',
         'noun' => 'BoltPuppetfileTypes'
+      },
+      'module:generate-types' => {
+        'verb' => 'Register',
+        'noun' => 'BoltModuleTypes'
       }
     }
 
@@ -47,9 +52,7 @@ namespace :pwsh do
     Bolt::CLI::COMMANDS.each do |subcommand, actions|
       # The 'bolt guide' command is handled by PowerShell's help system, so
       # don't create a cmdlet for it.
-      # The 'bolt module' command is currently feature-flagged and should
-      # not be visible to users, so don't create a PowerShell cmdlet for now.
-      next if %w[guide module].include?(subcommand)
+      next if %w[guide].include?(subcommand)
 
       actions << nil if actions.empty?
       actions.each do |action|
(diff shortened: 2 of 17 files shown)


lucywyman:
@jpogran I'm not sure why it's not picking up the Register-BoltModuleTypes command?

lib/bolt/project_migrator.rb
```diff
@@ -33,7 +33,7 @@ def migrate
 
       ok = migrate_inventory && migrate_config
```
beechtom, line 34:
```suggestion
      ok = migrate_inventory && migrate_config && migrate_modules
```

We should just remove the below `if` block entirely.

jpogran:
@lucywyman The reason is the hardcoded special case for the puppetfile commands. I'll be able to push up a commit that addresses this to your PR today

beechtom:
The default help text needs to be updated to include the `module` command: https://github.com/puppetlabs/bolt/blob/main/lib/bolt/bolt_option_parser.rb#L190

rakelib/pwsh.rake
```diff
@@ -49,7 +58,7 @@ namespace :pwsh do
       # don't create a cmdlet for it.
       # The 'bolt module' command is currently feature-flagged and should
```
jpogran, line 59: TODO: Remove this comment

rakelib/pwsh.rake
```diff
@@ -26,13 +26,25 @@ namespace :pwsh do
...
+        'noun' => 'BoltPuppetfileTypes'
+      },
+      'module:generate-types' => {
+        'verb' => 'Register',
+        'noun' => 'BoltModuleTypes'
+      },
+      'module:show-modules' => {
+        'verb' => 'Get',
+        'noun' => 'BoltModule'
       }
```
beechtom, line 48: We shouldn't need to hardcode this, since the command is `bolt module show`

lucywyman: I think that's from @jpogran

jpogran: I'll try without it, but coulda sworn..

jpogran: Good catch @beechtom , fixing

beechtom:
I believe we want to add a `Module` parameter for `Add-BoltModule`, otherwise the cmdlet can't specify a module to add with a named parameter: https://github.com/puppetlabs/bolt/blob/main/rakelib/pwsh.rake#L88

beechtom (changes requested):
I believe we want to add a `Module` parameter for `Add-BoltModule`, otherwise the cmdlet can't specify a module to add with a named parameter: https://github.com/puppetlabs/bolt/blob/main/rakelib/pwsh.rake#L88

rakelib/pwsh.rake
```diff
@@ -219,6 +222,20 @@ namespace :pwsh do
...
+            name:                       'Module',
+            ruby_short:                 'md',
+            help_msg:                   'The name of the module to add to the Bolt project',
+            mandatory:                  false,
+            type:                       'string',
+            switch:                     false,
+            position:                   0,
+            ruby_arg:                   'bare',
+            validate_not_null_or_empty: true
+          }
```
beechtom, line 238: This will add `Module` as an optional parameter to all of the `BoltModule` cmdlets. This should be wrapped in a case/if so that it only applies to the `add` action and can be listed as a mandatory parameter.

beechtom (changes requested):

rakelib/pwsh.rake
lucywyman: Ah, for other commands we just set it as not-mandatory if any subcommands didn't require it. I fixed those up as well.

rakelib/pwsh.rake
```diff
@@ -219,6 +223,21 @@ namespace :pwsh do
             ruby_arg:                   'bare',
             validate_not_null_or_empty: true
           }
+        when 'module'
+          # bolt module install
+          # bolt module add [module]
+          mod_param_mandatory = (@pwsh_command[:verb] == 'Add')
```
beechtom, line 229: This will still result in `Get-BoltModule`, `Install-BoltModule`, and `Register-BoltModuleTypes` having an optional `Module` parameter, which is not correct:

```
function Register-BoltModuleTypes
{
  [CmdletBinding()]
  param (
    [Parameter(Position=0)]
    [ValidateNotNullOrEmpty()]
    [Alias('md')]
    [string]
    $Module,
```

beechtom (changes requested):

beechtom (approved):

Issue #2208, opened 2020-09-24
Error when user supplies extra parameter to bolt module install


lucywyman:
## Use Case
Some users may accidentally try to use `bolt module install <module>` to install an individual module. We should error in this case and instruct to use `bolt module add` to add and install new modules.

beechtom:
Closed by #2215

PR #2215, opened 2020-09-25
(GH-2208) Error with extra arguments to 'bolt module install'

beechtom:
This change causes Bolt to error if any additional arguments are
provided to the `bolt module install` command. This is to prevent
potential confusion for users who might use the command to install a
single module, e.g. `bolt module install puppetlabs/apt`. The error
includes a suggestion to use `bolt module add` instead.

!no-release-note

### DIFF
--- a/lib/bolt/cli.rb
+++ b/lib/bolt/cli.rb
@@ -254,12 +254,6 @@ module Bolt
         end
       end
 
-      if options[:subcommand] != 'file' && options[:subcommand] != 'script' &&
-         !options[:leftovers].empty?
-        raise Bolt::CLIError,
-              "Unknown argument(s) #{options[:leftovers].join(', ')}"
-      end
-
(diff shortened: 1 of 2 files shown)


lib/bolt/cli.rb
```diff
@@ -314,6 +314,11 @@ def validate(options)
         raise Bolt::CLIError, "Must specify a module name."
       end
 
+      if options[:subcommand] == 'module' && options[:action] == 'install' && options[:object]
```
lucywyman, line 317: We may want to join this with `options[:leftovers]` as well, so that `bolt module install puppetlabs-apache puppetlabs-yaml` raises the same error and doesn't raise `Unknown argument(s)`

beechtom: I'm just gonna move that check for `:leftovers` to happen _after_ the check for `:object`. Seems like the correct logical ordering anyways.

PR #2216, opened 2020-09-25
(maint) Have modules key accept strings

lucywyman:
This adds support for the 'modules' key accepting strings in addition to
hashes as part of the modules array. If a module in the array is a
barestring it will be converted to the hash `{ name: <string> }`,
similar to how targets accept a string and treat it as the URI. This
provides a nice shorthand for users listing modules manually. The `bolt
module add` command will still add the hash version.

!no-release-note

### DIFF
--- a/lib/bolt/config/options.rb
+++ b/lib/bolt/config/options.rb
@@ -233,7 +233,7 @@ module Bolt
                        "install` command.",
           type: Array,
           items: {
-            type: Hash,
+            type: [Hash, String],
             required: ["name"],
             properties: {
               "name" => {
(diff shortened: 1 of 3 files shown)


beechtom:
The schema needs to be updated to accept a string: https://github.com/puppetlabs/bolt/blob/main/lib/bolt/config/options.rb#L236

It should be `type: [Hash, String]`.

lib/bolt/config/options.rb
```diff
@@ -233,8 +233,7 @@ module Options
                        "install` command.",
           type: Array,
           items: {
-            type: Hash,
-            required: ["name"],
```
beechtom, line 237: We should keep `required`, otherwise the schema won't flag hashes that are missing it.

beechtom (approved):

Issue #2217, opened 2020-09-27
keyword parameters is deprecated error message upon invocation 


logicminds:
## Describe the Bug
I am using bolt as a gem and when I run task show I get the following message.

```
sudo BOLT_GEM=true /opt/crossbelt/embedded/bin/bolt task show --modulepath=/opt/crossbelt/puppet_modules
/opt/crossbelt/embedded/lib/ruby/gems/2.7.0/gems/bolt-2.29.0/lib/bolt/cli.rb:392: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
/opt/crossbelt/embedded/lib/ruby/gems/2.7.0/gems/bolt-2.29.0/lib/bolt/analytics.rb:121: warning: The called method `screen_view' is defined here
Enum parameters must be identifiers or strings
   (file & line not available)

Guessing in ruby 2.7 `def screen_view(screen, **kwargs)` the **kwargs is deprecated. 

```

## Expected Behavior
My expectation is that no error messages would be returned.

## Steps to Reproduce
1. Install ruby 2.7.1
2. Install bolt gem
3. Run a bolt command with BOLT_GEM=true set

## Environment
- Ubuntu 18
- Ruby 2.7.1
- Bolt as a gem

## Additional Context
Since I package my own ruby and puppet and bolt I don't want the packaged version of bolt since it would bring another version of ruby to the environment.

lucywyman:
@logicminds That error is actually coming from Puppet https://github.com/puppetlabs/puppet/blob/main/lib/puppet/pops/types/type_parser.rb#L425 Is your task accepting an Enum parameter that isn't strings? Or are you trying to pass in a parameter that isn't a string to an Enum? I think this is coming up in analytics because we submit the analytics before executing the Bolt command, so that's why the error gets raised where it does.

nicklewis:
I don't think this is about the error from Puppet but rather the Ruby warnings. It looks like we have a few places that need to be updated.

lucywyman:
Well, I think we've at least addressed the warnings in this issue. Going to close this for now and we can address the same warning if it comes up elsewhere.

PR #2221, opened 2020-09-28
(GH-2217) Explicitly pass kwargs to methods that accept them

nicklewis:
In some places, we were passing keyword arguments as a hash as the last
argument to the method. This syntax is deprecated in Ruby 2.7 in favor
of explicitly passing them as keyword arguments using the ** operator.
Unfortunately, it's hard to programmatically find places where this
issue occurs since it depends on the definitin of the method, so this
commit is just a first stab.

!no-release-note

### DIFF
--- a/lib/bolt/catalog.rb
+++ b/lib/bolt/catalog.rb
@@ -97,7 +97,7 @@ module Bolt
       }
 
       with_puppet_settings(puppet_settings) do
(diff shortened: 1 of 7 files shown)


lucywyman (commented):
Missed one at https://github.com/puppetlabs/bolt/blob/main/lib/bolt/plugin/module.rb#L19

nicklewis:
Good catch! Updated.

PR #2224, opened 2020-09-28
(bug) Do not allow missing modules when resolving module dependencies

beechtom:
This fixes a bug in `Bolt::Puppetfile.resolve` that would allow the
resolver to return a dependency graph with missing module dependencies.
This would result in a graph that was difficult to parse and create a
helpful error for the user that informed them if any modules or their
dependencies could not be found. The resolver now does not permit any
missing module dependencies, which will cause the resolver to error as
soon as it encounters a missing module and notify the user that it could
not be found.

!bug

* **Show missing module dependencies when resolving modules**

  Bolt now correctly displays the names of missing module dependencies
  when resolving module dependencies errors. Previously, if a module
  dependency was missing, Bolt did not display the name of the missing
  module.

### DIFF
--- a/lib/bolt/puppetfile.rb
+++ b/lib/bolt/puppetfile.rb
@@ -104,40 +104,18 @@ module Bolt
           cache:                 nil,
           ui:                    nil,
           module_paths:          [],
-          allow_missing_modules: true
+          allow_missing_modules: false
         )
       rescue StandardError => e
         raise Bolt::Error.new(e.message, 'bolt/puppetfile-resolver-error')
       end
 
-      # Validate that the modules exist.
-      missing_graph = result.specifications.select do |_name, spec|
-        spec.instance_of? PuppetfileResolver::Models::MissingModuleSpecification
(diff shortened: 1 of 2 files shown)


nicklewis:
```
Puppetfile Resolver could not find compatible versions for possibility named "dockerinstall":
  In Puppetfile:
    aursu-kubeinstall >= 0 was resolved to Forge aursu-kubeinstall-0.2.1, which depends on
      aursu-dockerinstall >= 0.3.0 < 1.0.0
```

But where on earth does this message _come_ from? I can't find it in `puppetfile-resolver` anywhere...

nicklewis:
Ah that's from CocoaPods/Molinillo. It appears there's an option to pass to tell it what to call the "possibility" (so it could say "module" instead). Not sure whether that would be provided to some hash that puppetfile-resolver _already_ passes to molinillo or if we need to update puppetfile-resolver itself.

lucywyman (approved):

PR #2225, opened 2020-09-29
(docs) Standardize docs on modulepath, not module path

lucywyman:
This standardizes referring to the modulepath as 'modulepath', not
module path, in our documentation and code. This is just for clarify, and
avoid having to decide how to refer to the path of modules when writing
docs.

!no-release-note


hestonhoffman (approved):
Thanks Lucy!

Issue #2226, opened 2020-09-29
Bolt project migrate adds a `:latest` versionrequirement to modules which breaks bolt


adreyer:
## Describe the Bug
If a module in the Puppetfile specifies no version or `:latest` bolt will add `:latest` as the version requirement in `bolt-project.yaml` this will later prevent bolt from loading the file since symbols are not safe.

## Expected Behavior
`:latest` and no specified version should result in no versionrequirement in bolt-project.yaml.

PR #2229, opened 2020-09-29
(GH-2226) Do not set module version to latest when migrating

beechtom:
This prevents a module's version requirement in `bolt-project.yaml` from
being set to the string `:latest` when migrating a project. Previously,
if a Puppetfile had a module specification with either a version
`:latest` or no version at all, the module's version requirement would
be set to `:latest`, which is not valid. Now, in these instances the
module will not have a version requirement in the `bolt-project.yaml`.

!no-release-note


lucywyman:
:+1:

PR #2204, opened 2020-09-22
(GH-2183) Document new module workflow

hestonhoffman:
- Adds "Managing modules" page
- Add experimental callouts on various pages to link to the
  managing modules page.
- Edits to "Bolt projects" page to add .modules and a note to
  the Puppetfile description. Also added `bolt project init` to section
  on creating a Bolt project.

!no-release-note

PR #2238, opened 2020-09-30
(GH-2234) Add moduledir to generated Puppetfile

### LINKED ISSUE #2234: Generated Puppetfiles should specify the moduledir
## Use Case
The main advantage of continuing to use the Puppetfile format with the module workflow is that it allows the puppetfile to be committed and used from legacy tools. Adding `moduledir '.modules'` to the Puppetfile will improve this support by preventing r10k(or old versions of bolt) from deleting the modules directory and still installing the modules where new versions of bolt expect them.

## Additional Context
This will mean that new style projects cannot be run with older bolt versions.
We should document that users who want to use new style projects as environments in PE need to specify the correct modulepath in `environment.conf` .

lucywyman:
When generating the Puppetfile from the list of modules in
`bolt-project.yaml`, we now add the line `moduledir '.modules'` to
prevent R10K and old versions of Bolt from deleting the `modules/`
directory and installing modules from the Puppetfile to the wrong
location. This essentially ensures that using the new module management
workflow is safe with legacy tools, as the Puppetfile knows to install
modules to the new modules directory `.modules/` instead of installing
to `modules/` which should now be treated as a place for users' custom
modules. This line is only added when generating a new Puppetfile using
the new module management workflow - behavior is not modified if using
`bolt project init --modules`, which still uses the old workflow.

This additionally documents configuring the modulepath in PE
environments to pick up the new modulepath.

Closes #2234

!no-release-note

Note: I rebased this on #2204, so that needs to go in first in order to add the docs.

### DIFF
--- a/documentation/managing_modules.md
+++ b/documentation/managing_modules.md
@@ -48,7 +48,7 @@ the `modules` key in your `bolt-project.yaml`, resolves the modules and their
 dependencies, generates a Puppetfile listing all of the modules to install, and
 then installs the modules.
 
-If your project needs another module, you can use the `bolt module add` command to add the module to your
+If your project needs another module, you can use Bolt to add the module to your
 project configuration, generate a new Puppetfile that includes the new module
 and its dependencies, and install the modules.
 
--- a/documentation/managing_modules.md
+++ b/documentation/managing_modules.md
@@ -48,7 +48,7 @@ the `modules` key in your `bolt-project.yaml`, resolves the modules and their
 dependencies, generates a Puppetfile listing all of the modules to install, and
 then installs the modules.
 
-If your project needs another module, you can use Bolt to add the module to your
+If your project needs another module, you can use the `bolt module add` command to add the module to your
 project configuration, generate a new Puppetfile that includes the new module
 and its dependencies, and install the modules.
 
@@ -238,6 +238,19 @@ message if it cannot resolve a dependency due to a version requirement.
 > 🔩 **Tip**: For information on how to specify module versions, see the Puppet
 > documentation on [Specifying
 > versions](https://puppet.com/docs/puppet/latest/modules_metadata.html#specifying-versions).
+
+## Compatibility with Bolt versions
+
+The new module management style is incompatible with older Bolt versions, since it sets the
+moduledir in the Puppetfile to `.modules/` which is not on the modulepath in versions < 2.30.0. If
+you're using the new module management system in an environment in Puppet Enterprise, you need to
+specify `.modules` on the modulepath in your [environment.conf](https://puppet.com/docs/puppet/latest/config_file_environment.html#example), like so:
+
+```
+# /etc/puppetlabs/code/environments/test/environment.conf
+
+modulepath = site:dist:modules:$basemodulepath:.modules
+```
 
 ## Examples
 
--- a/lib/bolt/module_installer.rb
+++ b/lib/bolt/module_installer.rb
@@ -55,7 +55,7 @@ module Bolt
 
       # Write the Puppetfile.
       @outputter.print_message "Writing Puppetfile at #{puppetfile_path}"
-      puppetfile.write(puppetfile_path)
+      puppetfile.write(puppetfile_path, moduledir)
 
       # Install the modules.
       install_puppetfile(puppetfile_path, moduledir)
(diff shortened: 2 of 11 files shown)


documentation/managing_modules.md
```diff
@@ -0,0 +1,332 @@
...
+
+The new module management style is incompatible with older Bolt versions, since it sets the
+moduledir in the Puppetfile to `.modules/` which is not on the modulepath in versions < 2.30.0. If
+you're using the new module management system in an environment in Puppet Enterprise, you need to
+specify `.modules` on the modulepath in your [environment.conf](https://puppet.com/docs/puppet/latest/config_file_environment.html#example), like so:
+
+```
+# /etc/puppetlabs/code/environments/test/environment.conf
+
+modulepath = .modules:site:dist:modules:$basemodulepath
```
beechtom, line 252: Won't this cause `.modules` to take precedence over every other directory?

lucywyman: Woops yes.

lib/bolt/module_installer.rb
```diff
@@ -114,7 +114,16 @@ def install(modules, path, moduledir, force: false, resolve: true)
           puppetfile.resolve
 
           @outputter.print_message "Writing Puppetfile at #{path}"
-          puppetfile.write(path)
+          # We get here either through 'bolt module install' which uses the
+          # managed modulepath (which isn't configurable) or through bolt
+          # project init --modules, which uses the default modulepath. This
+          # should be safe to assume that if `.modules/` is the moduledir the
+          # user is using the new workflow
+          if moduledir.to_s =~ /\.modules/
```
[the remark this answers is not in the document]
lucywyman: :facepalm: This is literally a pathname, that's much easier than regex.

lib/bolt/module_installer.rb
```diff
@@ -114,7 +114,16 @@ def install(modules, path, moduledir, force: false, resolve: true)
           puppetfile.resolve
 
           @outputter.print_message "Writing Puppetfile at #{path}"
-          puppetfile.write(path)
+          # We get here either through 'bolt module install' which uses the
+          # managed modulepath (which isn't configurable) or through bolt
+          # project init --modules, which uses the default modulepath. This
+          # should be safe to assume that if `.modules/` is the moduledir the
+          # user is using the new workflow
+          if moduledir.basename == '.modules'
```
beechtom, line 122:
```suggestion
          if File.basename(moduledir) == '.modules'
```

`moduledir` (and all other parts of the modulepath) are strings.

lib/bolt/puppetfile.rb
```diff
@@ -56,9 +56,10 @@ def self.parse(path, skip_unsupported_modules: false)
     # Writes a Puppetfile that includes specifications for each of the
     # modules.
     #
-    def write(path)
+    def write(path, moduledir = nil)
       File.open(path, 'w') do |file|
         file.puts '# This Puppetfile is managed by Bolt. Do not edit.'
+        file.puts "moduledir '#{moduledir}'" if moduledir
```
beechtom, line 62:
```suggestion
        file.puts "moduledir '#{File.basename(moduledir)}'" if moduledir
```

We should put the basename here, not the full path, otherwise the project can't be shared between users.

beechtom (changes requested):

lib/bolt/module_installer.rb
lucywyman: ... ?
```
lucy ~/githubs/bolt/test (GH-2234) $ be bolt project init --modules puppetlabs-apache
Resolving module dependencies, this may take a moment
Writing Puppetfile at /home/lucy/githubs/bolt/test/Puppetfile

Frame number: 0/23

From: /home/lucy/githubs/bolt/lib/bolt/module_installer.rb:123 Bolt::ModuleInstaller#install:

    118:           # managed modulepath (which isn't configurable) or through bolt
    119:           # project init --modules, which uses the default modulepath. This
    120:           # should be safe to assume that if `.modules/` is the moduledir the
    121:           # user is using the new workflow
    122:           require 'pry'; binding.pry
 => 123:           if moduledir.basename == '.modules'
    124:             puppetfile.write(path, moduledir)
    125:           else
    126:             puppetfile.write(path)
    127:           end
    128:         end

[1] pry(#<Bolt::ModuleInstaller>)> moduledir.class
=> Pathname
```

lucywyman: Oh UGH. It's a pathname if it gets there through `bolt project init --modules`, but a string if it gets there through `bolt module install` since apparently the moduledir on the project is the *only* path that's a string?

beechtom (approved):

PR #2246, opened 2020-10-08
Add moduledir directive to Puppetfile with 'module install'

beechtom:
This fixes a bug in the `Bolt::ModuleInstaller` class that does not
correctly pass the managed moduledir to `Bolt::Puppetfile#write`. This
resulted in Bolt generating Puppetfiles with the `bolt module install`
command that did not include the `moduledir` directive.

!bug

* **Add moduledir directive to generated Puppetfile**

  Puppetfiles generated using the `bolt module install` command did not
  include the `moduledir` directive.

### DIFF
--- a/lib/bolt/puppetfile.rb
+++ b/lib/bolt/puppetfile.rb
@@ -61,8 +61,15 @@ module Bolt
     #
     def write(path, moduledir = nil)
       File.open(path, 'w') do |file|
-        file.puts '# This Puppetfile is managed by Bolt. Do not edit.'
-        file.puts "moduledir '#{moduledir.basename}'" if moduledir
+        if moduledir
+          file.puts "# This Puppetfile is managed by Bolt. Do not edit."
+          file.puts "# For more information, see https://pup.pt/bolt-modules"
+          file.puts
+          file.puts "# The following directive installs modules to the managed moduledir."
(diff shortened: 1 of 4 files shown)


lib/bolt/puppetfile.rb
```diff
@@ -61,8 +61,13 @@ def self.parse(path, skip_unsupported_modules: false)
     #
     def write(path, moduledir = nil)
       File.open(path, 'w') do |file|
-        file.puts '# This Puppetfile is managed by Bolt. Do not edit.'
-        file.puts "moduledir '#{moduledir.basename}'" if moduledir
+        if moduledir
+          file.puts "# This Puppetfile is managed by Bolt. Do not edit."
+          file.puts "# For more information, see https://pup.pt/bolt-modules\n\n"
+          file.puts "# The following directive installs modules to the managed moduledir."
+          file.puts "moduledir '#{moduledir.basename}'\n\n"
```
nicklewis, line 68: It might be clearer to have a couple extra empty `file.puts` calls below this line rather than append `\n\n`. Or this could be a heredoc. Or both!

beechtom: I'll do the extra `file.puts`. Anything to avoid heredocs. :P

nicklewis (approved):

Issue #2205, opened 2020-09-23
Diagnose inventory issues with `bolt inventory show`


nicklewis:
## Use Case
When trying to use `bolt inventory show` to diagnose inventory problems, it's helpful to understand where inventory information is coming from and why it's not showing the expected results.

## Describe the Solution You Would Like
• Print the inventory path that was loaded 
• Print a warning (or other indicator) if the specified target argument doesn't exist in the inventory
- Print 'Could not find config for target in inventory' as a warning for bolt inventory show command.

PR #2248, opened 2020-10-08
(GH-2205) Improve 'bolt inventory show' output

beechtom:
This improves the output for the `bolt inventory show` command.

When using the `human` outputter, Bolt will now display each target's source
alongside the target. If the target was loaded from inventory, it will
show the path to the inventory file; if the target is an adhoc target,
it will say it was not loaded from inventory. The output will also show
where inventory was loaded from, and tell the user if the inventory file
it attempted to load does not exist. Lastly, the target count now lists
the total number of targets, the number loaded from inventory, and the
number of adhoc targets.

When using the `json` outputter, Bolt will now include `inventory` and
`adhoc` keys. The `inventory` key includes a list of targets loaded from
inventory, the number of targets loaded from inventory, and the path to
the inventory file. The `adhoc` key lists the names of adhoc targets,
and the total number of adhoc targets.

!feature

* **Improved output for `bolt inventory show` and `Get-BoltInventory`**
  ([#2205](https://github.com/puppetlabs/bolt/issues/2205))

  The `bolt inventory show` command and `Get-BoltInventory` cmdlet now
  show where each target was loaded from alongside the target's name.
  Output also includes the path to the loaded inventory file and the
  number of inventory targets and adhoc targets.

### DIFF
--- a/lib/bolt/cli.rb
+++ b/lib/bolt/cli.rb
@@ -587,8 +587,24 @@ module Bolt
     end
 
     def list_targets
+      inventoryfile = config.inventoryfile || config.default_inventoryfile
+
+      # Retrieve the known group and target names. This needs to be done before
+      # updating targets, as that will add adhoc targets to the inventory.
+      known_names = inventory.target_names
+
       update_targets(options)
-      outputter.print_targets(options[:targets])
+
+      inventory_targets, adhoc_targets = options[:targets].partition do |target|
+        known_names.include?(target.name)
+      end
+
+      target_list = {
+        inventory: inventory_targets,
+        adhoc:     adhoc_targets
+      }
+
+      outputter.print_targets(target_list, inventoryfile)
     end
 
     def show_targets
--- a/lib/bolt/outputter/human.rb
+++ b/lib/bolt/outputter/human.rb
@@ -331,10 +331,28 @@ module Bolt
         end
       end
 
-      def print_targets(targets)
-        count = "#{targets.count} target#{'s' unless targets.count == 1}"
-        @stream.puts targets.map(&:name).join("\n")
-        @stream.puts colorize(:green, count)
+      def print_targets(target_list, inventoryfile)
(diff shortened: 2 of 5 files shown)


lib/bolt/cli.rb
```diff
@@ -587,8 +591,27 @@ def list_plans
     end
 
     def list_targets
+      inventoryfile = config.inventoryfile || config.default_inventoryfile
+
+      # Check target args against group and target names in the inventory. This
+      # needs to be done separately from resolving the inventory as update_targets
+      # adds targets to the inventory.
+      targets       = resolve_target_args(options).first&.split(',') || []
```
nicklewis, line 599: Can we just build the `known_names` set before calling `update_targets`? I don't quite follow why we _have_ to do this step separately.

lib/bolt/outputter/human.rb
```diff
@@ -331,10 +331,29 @@ def print_module_list(module_list)
...
-        @stream.puts colorize(:green, count)
+      def print_targets(target_list, inventoryfile)
+        inventory = colorize(:cyan, inventoryfile.to_s)
+        adhoc     = colorize(:cyan, "Not found in inventory file")
+
+        targets  = []
+        targets += target_list[:inventory].map { |target| [inventory, target.name] }
+        targets += target_list[:adhoc].map { |target| [adhoc, target.name] }
+
+        if targets.any?
```
nicklewis, line 342: Given that there are only two values for the first column and we already group the targets by them, it seems redundant to print that column every time. Would it make more sense for  it to be a header? Or maybe just an annotation for targets that _don't_ exist in the inventory (and nothing for the ones that do, since we report the inventory file below)?

lib/bolt/cli.rb
beechtom: Err, definitely. Updated. :)

nicklewis (approved):

PR #2251, opened 2020-10-09
(docs) Update 'Using plugins' page, add 'Supported plugins' page

beechtom:
This updates the 'Using plugins' page to describe using plugins in more
detail. It also moves the list of bundled plugins to a new 'Supported
plugins' page.

!no-release-note

### DIFF
--- a/documentation/bolt.ditamap
+++ b/documentation/bolt.ditamap
@@ -16,9 +16,6 @@
         <topicref href="projects.md" format="markdown"/>
         <topicref href="inventory_file_v2.md" format="markdown"/>
         <topicref href="configuring_bolt.md" format="markdown"/>
-        <topicref href="using_plugins.md" format="markdown">
-            <topicref href="writing_plugins.md" format="markdown"/>
-        </topicref>
         <topicref href="logs.md" format="markdown"/>
         <topicref href="privilege_escalation.md" format="markdown"/>
         <topicref href="vscode_and_bolt.md" format="markdown"/>
@@ -80,6 +77,11 @@
             </topicmeta>
         </topicref>
     </topicref>
+    <topichead navtitle="Plugins">
+      <topicref href="using_plugins.md" format="markdown"/>
+      <topicref href="supported_plugins.md" format="markdown"/>
+      <topicref href="writing_plugins.md" format="markdown"/>
+    </topichead>
     <topichead navtitle="Using Bolt with Puppet">
         <topicref href="applying_manifest_blocks.md" format="markdown"/>
         <topicref href="bolt_configure_orchestrator.md" format="markdown"/>
--- a/documentation/supported_plugins.md
+++ b/documentation/supported_plugins.md
@@ -1,0 +1,211 @@
+# Supported plugins
+
+The following plugins are supported and maintained by Bolt. Supported plugins
+are shipped with Bolt packages and do not need to be installed separately.
+
+## Reference plugins
+
+Reference plugins fetch data from an external source and store it in a static
+data object. You can use reference plugins in configuration files, inventory
+files, and plans.
+
+| Plugin | Description | Documentation |
+| --- | --- | --- |
+| `aws_inventory` | Generate targets from Amazon Web Services EC2 instances. | [aws_inventory](https://forge.puppet.com/puppetlabs/aws_inventory) |
+| `azure_inventory` | Generate targets from Azure VMs and VM scale sets. | [azure_inventory](https://forge.puppet.com/puppetlabs/azure_inventory) |
+| `env_var` | Read values stored in environment variables. | [env_var](#env-var) |
+| `gcloud_inventory` | Generate targets from Google Cloud compute engine instances. | [gcloud_inventory](https://forge.puppet.com/puppetlabs/gcloud_inventory) |
+| `pkcs7` | Decrypt ciphertext. | [pkcs7](https://forge.puppet.com/puppetlabs/pkcs7) |
+| `prompt` | Prompt the user for a sensitive value. | [prompt](#prompt) |
+| `puppetdb` | Query PuppetDB for a group of targets. | [puppetdb](#puppetdb) |
+| `task` | Run a task as a plugin. | [task](#task) |
+| `terraform` | Generate targets from local and remote Terraform state files. | [terraform](https://forge.puppet.com/puppetlabs/terraform) |
+| `vault` | Access secrets from a Key/Value engine on a Hashicorp Vault server. | [vault](https://forge.puppet.com/puppetlabs/vault) |
+| `yaml` | Compose multiple YAML files into a single file. | [yaml](#yaml) |
+
+## Secret plugins
+
+Use secret plugins to create keys for encryption and decryption, to encrypt
+plaintext, or to decrypt ciphertext. Secret plugins are used by Bolt's `secret`
+command.
+
+| Plugin | Description | Documentation |
+| --- | --- | --- |
+| `pkcs7` | Generate key pairs, encrypt plaintext, and decrypt ciphertext. | [pkcs7](https://forge.puppet.com/puppetlabs/pkcs7) |
+
+## Puppet library plugins
+
+Puppet library plugins ensure that the Puppet library is installed on a target
+when a plan calls the `apply_prep` function.
+
+| Plugin | Description | Documentation |
+| --- | --- | --- |
+| `puppet_agent` | Install Puppet libraries on target nodes when a plan calls `apply_prep`. | [puppet_agent](https://forge.puppet.com/puppetlabs/puppet_agent) |
+
+## Built-in plugins
+
+The following plugins are built into Bolt and are not available in modules.
+
+### `env_var`
+
+The `env_var` plugin allows users to read values stored in environment variables
+and load them into an inventory or configuration file.
+
+#### Parameters
+
+The following parameters are available to the `env_var` plugin:
+
+| Parameter | Description | Type | Default |
+| --- | ----------- | ---- | ------- |
+| `var` | **Required.** The name of the environment variable to read from. | `String` | None |
+| `default` | A value to use if the environment variable `var` isn't set. | `String` | None |
+| `optional` | Unless `true`, `env_var` raises an error when the environment variable `var` does not exist.  When `optional` is `true` and `var` does not exist, env_var returns `nil`. | `Boolean` | `false` |
+
+#### Example usage
+
+Looking up a value from an environment variable in an inventory file:
+
+```yaml
+targets:
+  - target1.example.com
+config:
+  ssh:
+    user: bolt
+    password:
+      _plugin: env_var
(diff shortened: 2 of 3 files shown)


documentation/using_plugins.md
```diff
@@ -1,339 +1,213 @@
...
-multiple hooks. The way in which a plugin is used varies depending on the type
-of hook used.
-
-> **Note:** Plugins are only available in configuration files and version 2
-> inventory files.
-
+Bolt supports the use of plugins to dynamically load information during a
+Bolt run and change how some Bolt actions are executed. Plugins either ship
+with Bolt or are installed as Puppet modules that have the same name as the
+plugin.
```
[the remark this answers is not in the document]
hestonhoffman: Actually, this makes me wonder if we need a short section here on installing a plugin from another user? Even if it's a case of linking off to the installing modules page. 
Edit: It looks like users might have to use the built-in task plugin?

documentation/supported_plugins.md
```diff
@@ -0,0 +1,207 @@
+# Supported plugins
+
+The following plugins are supported and maintained by Bolt. Supported plugins
+are shipped with Bolt packages and do not need to be installed separately.
+
+## Reference plugins
+
+Reference plugins are used to fetch data from an external source and store it in
+a static data object. Reference plugins can be used in configuration files,
+inventory files, and plans.
```
hestonhoffman, line 10:
```suggestion
Reference plugins fetch data from an external source and store it in
a static data object. You can use reference plugins in configuration files,
inventory files, and plans.
```

documentation/bolt.ditamap
```diff
@@ -16,9 +16,11 @@
         <topicref href="projects.md" format="markdown"/>
         <topicref href="inventory_file_v2.md" format="markdown"/>
         <topicref href="configuring_bolt.md" format="markdown"/>
-        <topicref href="using_plugins.md" format="markdown">
+        <topichead navtitle="Plugins">
```
hestonhoffman, line 19: After reading through this stuff, I think we can move plugins out of the configuring section and just list them in the main nav. Their application is much broader than just configuration. Maybe this section can go after modules.

documentation/supported_plugins.md
```diff
@@ -0,0 +1,207 @@
...
+| `prompt` | Prompt the user for a sensitive value. | [prompt](#prompt) |
+| `puppetdb` | Query PuppetDB for a group of targets. | [puppetdb](#puppetdb) |
+| `task` | Run a task as a plugin. | [task](#task) |
+| `terraform` | Generate targets from local and remote Terraform state files. | [terraform](https://forge.puppet.com/puppetlabs/terraform) |
+| `vault` | Access secrets from a Key/Value engine on a Hashicorp Vault server. | [vault](https://forge.puppet.com/puppetlabs/vault) |
+| `yaml` | Compose multiple YAML files into a single file. | [yaml](#yaml) |
+
+## Secret plugins
+
+Secret plugins are used to create keys for encryption and decryption, encrypt
```
hestonhoffman, line 28:
```suggestion
Use secret plugins to create keys for encryption and decryption, encrypt
```

documentation/supported_plugins.md
```diff
@@ -0,0 +1,207 @@
...
+plaintext, and decrypt ciphertext. Secret plugins are used by Bolt's `secret`
+command.
+
+| Plugin | Description | Documentation |
+| --- | --- | --- |
+| `pkcs7` | Generate key pairs, encrypt plaintext, and decrypt ciphertext. | [pkcs7](https://forge.puppet.com/puppetlabs/pkcs7) |
+
+## Puppet library plugins
+
+Puppet library plugins are used to make sure the Puppet library is installed
```
hestonhoffman, line 38:
```suggestion
Puppet library plugins ensure that the Puppet library is installed
```

documentation/supported_plugins.md
```diff
@@ -0,0 +1,207 @@
...
+on a target when the `apply_prep` function is called.
```
hestonhoffman, line 39:
```suggestion
on a target when a plan calls the `apply_prep` function.
```

documentation/supported_plugins.md
```diff
@@ -0,0 +1,207 @@
...
+and load them into an inventory or configuration file.
+
+#### Parameters
+
+The following parameters are available to the `env_var` plugin:
+
+| Parameter | Description | Type | Default |
+| --- | ----------- | ---- | ------- |
+| `var` | **Required.** The name of the environment variable to read from. | `String` | None |
+| `default` | A value to use if the environment variable `var` isn't set | `String` | None |
```
hestonhoffman, line 61:
```suggestion
| `default` | A value to use if the environment variable `var` isn't set. | `String` | None |
```

documentation/supported_plugins.md
```diff
@@ -0,0 +1,207 @@
...
+| `optional` | Unless `true`, env_var will raise an error when the environment variable `var` does not exist.  When `optional` is `true` and `var` does not exist, env_var returns `nil` | `Boolean` | `false` |
```
hestonhoffman, line 62:
```suggestion
| `optional` | Unless `true`, `env_var` raises an error when the environment variable `var` does not exist.  When `optional` is `true` and `var` does not exist, `env_var` returns `nil`. | `Boolean` | `false` |
```

documentation/supported_plugins.md
```diff
@@ -0,0 +1,207 @@
...
+The `puppetdb` plugin queries PuppetDB for a group of targets. 
+
+If target-specific configuration is required, the `puppetdb` plugin can be used
+to lookup configuration values for the `alias`, `config`, `facts`, `features`,
+`name`, `uri` and `vars` inventory options for each target. These values can be
+set in the `target_mapping` field. The fact lookup values can be either
+`certname` to reference the `[certname]` of the target, or a [PQL dot
+notation](https://puppet.com/docs/puppetdb/latest/api/query/v4/ast.html#dot-notation)
+facts string such as `facts.os.family` to reference a fact value. Dot notation
+is required for both structured and unstructured facts.
```
hestonhoffman, line 119:
```suggestion
If you require target-specific configuration, you can use the `puppetdb` plugin
to look up configuration values for the `alias`, `config`, `facts`, `features`,
`name`, `uri` and `vars` inventory options for each target. Set these values
in the `target_mapping` field. The fact lookup values can be either
`certname` to reference the `[certname]` of the target, or a [PQL dot
notation](https://puppet.com/docs/puppetdb/latest/api/query/v4/ast.html#dot-notation)
facts string such as `facts.os.family` to reference a fact value. Dot notation
is required for both structured and unstructured facts.
```

documentation/supported_plugins.md
```diff
@@ -0,0 +1,207 @@
...
+
+The following parameters are available to the `puppetdb` plugin:
+
+| Parameter | Description | Type | Default |
+| --- | ----------- | ---- | ------- |
+| `query` | **Required.** A string containing a [PQL query](https://puppet.com/docs/puppetdb/latest/api/query/v4/pql.html) or an array containing a [PuppetDB AST format query](https://puppet.com/docs/puppetdb/latest/api/query/v4/ast.html). | `String` | None |
+| `target_mapping` | **Required.** A hash of target attributes (`name`, `uri`, `config`) to populate with fact lookup values. | `Hash` | None |
+
+> **Note:** If neither `name` nor `uri` is specified in `target_mapping`, then
+> `uri` will be set to `certname`.
```
hestonhoffman, line 131:
```suggestion
> `uri` is set to `certname`.
```

documentation/supported_plugins.md
```diff
@@ -0,0 +1,207 @@
...
+`target_mapping` field:
+
+| Key | Description |
+| --- | ----------- |
+| `certname` | The certname of the node returned from PuppetDB. This is short hand for doing: `facts.trusted.certname`. |
+| `facts.*` | [PQL dot notation](https://puppet.com/docs/puppetdb/latest/api/query/v4/ast.html#dot-notation) facts string such as `facts.os.family` to reference fact value. Dot notation is required for both structured and unstructured facts. |
+
+#### Example usage
+
+Lookup targets with the fact `osfamily: RedHat` and setting:
```
hestonhoffman, line 145:
```suggestion
Look up targets with the fact `osfamily: RedHat` and setting:
```

hestonhoffman, line 145:
```suggestion
Lookup targets with the fact `osfamily: RedHat` and the following 
configuration values:
```

documentation/supported_plugins.md
```diff
@@ -0,0 +1,207 @@
...
+ * The alias with the fact `hostname`
+ * The name with the fact `certname`
+ * A target fact called `custom_fact` with the `custom_fact` from PuppetDB
+ * A feature from the fact `custom_feature`
+ * The SSH hostname with the fact `networking.interfaces.en0.ipaddress`
+ * The puppetversion var from the fact `puppetversion`
```
hestonhoffman, line 151:
```suggestion
 * The puppetversion variable from the fact `puppetversion`
```

documentation/supported_plugins.md
```diff
@@ -0,0 +1,207 @@
...
+      vars:
+        puppetversion: facts.puppetversion
+```
+
+### `task`
+
+The `task` plugin lets Bolt run a task as a plugin. Tasks will run on the
+`localhost` target without access to any configuration defined in an inventory
+file, but with access to any parameters that are configured. The plugin extracts
+the `value` key from the task output and uses that as the value.
```
hestonhoffman, line 176:
```suggestion
The `task` plugin lets Bolt run a task as a plugin and extracts
the `value` key from the task output to use as the plugin value. 
Plugin tasks run on the `localhost` target without access to any 
configuration defined in an inventory file, but with access to 
any parameters that you've configured.

For example, you could run a `task` plugin that collects target 
names from a JSON file and interpolates them into a `target` array 
in your inventory file.
```

documentation/using_plugins.md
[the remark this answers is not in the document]
beechtom: I think a link to "Installing modules" would suffice here. Third-party plugins should be written as plugins, so there shouldn't be a need to use the `task` plugin to run them.

documentation/using_plugins.md
```diff
@@ -1,339 +1,213 @@
...
-  - target1.example.com
-config:
+# bolt-defaults.yaml
+inventory-config:
   ssh:
-    user: bolt
-    password:
-      _plugin: env_var
-      var: BOLT_PASSWORD
+    password: bolt
```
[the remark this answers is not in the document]
beechtom: Removed. I agree that it's a bit confusing.

documentation/bolt.ditamap
beechtom: Done and done! :)

hestonhoffman (approved):
This looks good - I've added a small paragraph to Using plugins

hestonhoffman (approved):

PR #2252, opened 2020-10-09
Refactor modulepath parsing in PAL to new class

McdonaldSeanp:
This PR refactors the modulepath parsing we do in PAL in to a new Bolt::Config::Modulepath class, that can hold the value of the modulepath and provide transformation functions to PAL functionality.

As part of this change, we also remove the PE::PAL class used in bolt-server, and instead use a child class of the Modulepath object that can take an environment.


McdonaldSeanp:
This refactor enables us here: https://github.com/puppetlabs/orchestrator/pull/1503

lib/bolt/pal.rb
```diff
@@ -157,7 +171,7 @@ def detect_project_conflict(project, environment)
     def in_bolt_compiler
       # TODO: If we always call this inside a bolt_executor we can remove this here
       setup
-      r = Puppet::Pal.in_tmp_environment('bolt', modulepath: @modulepath, facts: {}) do |pal|
+      r = Puppet::Pal.in_tmp_environment('bolt', modulepath: @modulepath.bolt_compiler_modulepath, facts: {}) do |pal|
```
donoghuc, line 174: why not just use the `modulepath` method? https://github.com/puppetlabs/bolt/pull/2252/files#diff-6f471e6eed4bf0e0390c0a6a5aa5304cfd0d7e22a034795ea51c18da15381b90R79

lib/bolt/pal.rb
```diff
@@ -212,11 +226,11 @@ def with_bolt_executor(executor, inventory, pdb_client = nil, applicator = nil,
         apply_executor: applicator || Applicator.new(
           inventory,
           executor,
-          @modulepath,
+          @modulepath.bolt_compiler_modulepath,
```
donoghuc, line 229: same question for both of these https://github.com/puppetlabs/bolt/pull/2252/files#r504079462

lib/bolt/pal.rb
```diff
@@ -45,16 +43,24 @@ def initialize(msg, details = {})
...
 
-      @user_modulepath = modulepath
-      @modulepath = [BOLTLIB_PATH, *modulepath, MODULES_PATH]
+      # Build a Config::Modulepath object that can perform modulepath transformations
+      # for normal bolt and apply compilation
+      if modulepath.is_a?(Array) || modulepath.is_a?(String) || modulepath.nil?
+        @modulepath = Bolt::Config::Modulepath.new(modulepath)
+      # Specifically check the superclass, instead of the class itself,
+      # so that a child class can also be sent to PAL.
+      elsif modulepath.class.superclass == Bolt::Config::Modulepath
```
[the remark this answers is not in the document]
McdonaldSeanp: Yeah, I could see replacing the equality with `is_a?`

As far as the awkwardness: I think we need to find a way in PAL to allow for extending the inputs so that they can match our needs in a way that we avoid things like monkeypatching or redefining PAL. I was thinking it made sense to use superclass because inheritance struck me as a reasonable implementation of the requirements of both Bolt and PE, since bolt can always use the parent, and PE can use pretty standard inheritance for its needs.

If there's another approach that you think is more appropriate I am certainly on board to look at other options though.

lib/bolt/pal.rb
```diff
@@ -63,13 +69,21 @@ def initialize(modulepath, hiera_config, resource_types, max_compiles = Etc.npro
       @project = project
 
       @logger = Bolt::Logger.logger(self)
-      if modulepath && !modulepath.empty?
-        @logger.debug("Loading modules from #{@modulepath.join(File::PATH_SEPARATOR)}")
+      unless @modulepath.apply_compiler_modulepath.empty?
```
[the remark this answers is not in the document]
McdonaldSeanp: Yeah, I think this needs a rename. I think the bolt_compiler_modulepath makes sense as a name, since it is actually that: the modulepath that the bolt compiler needs. But I think this other modulepath needs another name.

spec/bolt_server/transport_app_spec.rb
```diff
@@ -411,91 +333,67 @@ def mock_plan_info(full_name)
...
-
-      context 'with non-existant task' do
-        let(:path) { '/tasks/foo/bar?environment=production' }
-        it 'returns 400 if an unknown plan error is thrown' do
-          expect(BoltServer::PE::PAL).to receive(:new).and_return(fake_pal)
-          expect(fake_pal).to receive(:get_task).with('foo::bar').and_raise(Bolt::Error.unknown_task('foo::bar'))
-          get(path)
-          expect(last_response.status).to eq(400)
-        end
-      end
```
donoghuc, line 430: In general in this file I think its good that we are not mocking PAL. Maybe i'm missing it but it looks like we lost all our tests around non-existant content?

spec/fixtures/bolt_server/environments/production/modules/bolt_server_test/tasks/init.sh
```diff
@@ -0,0 +1 @@
+whoami
```
donoghuc, line 1: for all the task implemnetations it might be better to just have empty files rather than tasks that wont work (missing shebang).

spec/bolt_server/transport_app_spec.rb
McdonaldSeanp: I removed several of these because they are redundant. Most of the functionality uses the same helper functions, and it didn't seem particularly helpful to retest those helpers over and over.

lib/bolt/pal.rb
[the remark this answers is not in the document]
McdonaldSeanp: IMO there isn't any extra value in having this handle explicit types vs. checking the ancestry of the class. Is there a specific concern you have with the check?

My thinking here is if we check the ancestry chain, the child classes in bolt-server/PE can do whatever they want with their names, and we don't have to make any changes in bolt if the name of the classes change. We also don't have to explicitly add checks specific to only PE in to the bolt codebase, which means in the future if another tool started building on top of bolt they could re-use this architecture.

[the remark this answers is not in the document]
nicklewis:
```rb 
@modulepath = case modulepath
              when Array, String, nil
                Bolt::Config::Modulepath.new(modulepath)
              when Bolt::Config::Modulepath
                modulepath
              else
                msg = "Type error in PAL: modulepath must be one of Array, Bolt::Config::Modulepath"
                raise Bolt::Error.new(msg, "bolt/execution-error")
              end
```

But I think it might be better to just lean into duck typing here and require the caller to provide a `Bolt::Config::Modulepath` or equivalent object and avoid explicit type checking or conversion. That might have implications for existing consumers of `Bolt::PAL` though.

lib/bolt/config/modulepath.rb
```diff
@@ -0,0 +1,38 @@
...
+require 'bolt/config'
+
+module Bolt
+  class Config
+    class Modulepath
+      BOLTLIB_PATH = File.expand_path('../../../bolt-modules', __dir__)
+      MODULES_PATH = File.expand_path('../../../modules', __dir__)
+
+      def initialize(modulepath)
+        @modulepath = if modulepath.nil?
```
nicklewis, line 12: This block is equivalent to `@modulepath = Array(modulepath)`

lib/bolt/config/modulepath.rb
```diff
@@ -0,0 +1,38 @@
...
+
+require 'bolt/config'
+
+module Bolt
+  class Config
+    class Modulepath
+      BOLTLIB_PATH = File.expand_path('../../../bolt-modules', __dir__)
+      MODULES_PATH = File.expand_path('../../../modules', __dir__)
+
+      def initialize(modulepath)
```
nicklewis, line 11: I think it would be better if this were a _little_ less purpose-built. Maybe methods like `boltlib`, `user_modules`, and `builtin_modules`? We could then make the constructor take those three pieces and have a factory method that 
 supplies the default locations for `boltlib` and `builtin_modules`. That might avoid the need for a PE fork altogether?

lib/bolt/config/modulepath.rb
```diff
@@ -0,0 +1,38 @@
...
+                        [modulepath]
+                      else
+                        modulepath
+                      end
+      end
+
+      # The bolt_compiler_modulepath includes both the BOLTLIB
+      # path and the MODULES_PATH to ensure bolt functions and
+      # built-in content are available in the compliler
+      def bolt_compiler_modulepath
```
nicklewis, line 24: Maybe `full_modulepath`?

lib/bolt/config/modulepath.rb
```diff
@@ -0,0 +1,38 @@
...
+        [BOLTLIB_PATH, *@modulepath, MODULES_PATH]
+      end
+
+      # The apply_compiler_modulepath only includes the original
+      # modulepath, since we don't want to pluginsync any of the
+      # content from BOLT_MODULES since that content includes
+      # core modules that can conflict with modules installed
+      # with an agent
+      def apply_compiler_modulepath
```
nicklewis, line 33: This one is a little weird. It's really more like `plugin_modules`, since actual _compilation_ of apply blocks uses the entire modulepath. It's only when gathering facts and when applying the catalog that we use a subset of them. But if we had a `user_modules` method then maybe that's what we would want to use here.

lib/bolt/pal.rb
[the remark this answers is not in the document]
McdonaldSeanp: Are there any consumers other than PE and bolt?

Changing all of the bolt PAL.new calls isn't too hard

donoghuc (approved):

lib/bolt/config/modulepath.rb
```diff
@@ -0,0 +1,33 @@
...
+        @user_modulepath = Array(user_modulepath)
+        @boltlib_path = Array(boltlib_path)
+        @builtin_content_path = Array(builtin_content_path)
+      end
+
+      # The full_modulepath includes both the BOLTLIB
+      # path and the MODULES_PATH to ensure bolt functions and
+      # built-in content are available in the compliler
+      def full_modulepath
+        [*@boltlib_path, *@user_modulepath, *@builtin_content_path]
```
nicklewis, line 29: Since we ensure these are all arrays, it could just be `@boltlib_path + @user_modulepath + @builtin_content_path`

lib/bolt/config/modulepath.rb
```diff
@@ -0,0 +1,33 @@
...
+    class Modulepath
+      BOLTLIB_PATH = File.expand_path('../../../bolt-modules', __dir__)
+      MODULES_PATH = File.expand_path('../../../modules', __dir__)
+
+      # The user_modulepath only includes the original
+      # modulepath, and is used during pluginsync
+      # since we don't want to pluginsync any of the
+      # content from BOLT_MODULES since that content includes
+      # core modules that can conflict with modules installed
+      # with an agent
```
beechtom, line 16:
```suggestion
      # The user_modulepath only includes the original modulepath and is used during pluginsync.
      # We don't want to pluginsync any of the content from BOLT_MODULES since that content 
      # includes core modules that can conflict with modules installed with an agent.
```

Just some punctuation to make this comment a little easier to understand.

beechtom (approved):

PR #2262, opened 2020-10-15
Fix transport_app_spec tests

McdonaldSeanp:
Previously, the transport_app_spec tests were broken due to two simultaneous
code changes that didn't merge well.

This commit updates the previous tests to use fixtures, and adds a new test
for directories when using /file_metadatas


donoghuc (approved):

spec/bolt_server/transport_app_spec.rb
```diff
@@ -787,82 +788,184 @@ def post_over_transport(transport, action, body_content, multiple: false)
...
         expect(last_response.status).to eq(400)
       end
 
       it 'returns 400 if project_ref does not exist' do
-        allow(Dir).to receive(:exist?).with("/tmp/foo/#{project_ref}").and_return(false)
-        get(path)
+        get("/project_file_metadatas/bar/foo?project_ref=not_a_real_project")
         error = last_response.body
-        expect(error).to eq("`project_ref`: /tmp/foo/#{project_ref} does not exist")
+        expect(error).to include("`project_ref`: #{project_dir}/not_a_real_project does not exist")
```
ekinanp, line 806: Yeah, propagating bad consistency is probably not a good idea (e.g. https://github.com/puppetlabs/bolt/blob/main/spec/bolt_server/transport_app_spec.rb#L236 and other places also use `eq`).

I think we use `include`, then the match should be shorter? e.g. `include(%r{project_ref.*#{project_dir}/not_a_real_project)` ?

spec/bolt_server/transport_app_spec.rb
```diff
@@ -787,82 +788,184 @@ def post_over_transport(transport, action, body_content, multiple: false)
...
+          end
+          expect(file_metadatas).to eq(expected_response)
+          expect(last_response.status).to eq(200)
+        end
       end
 
-      context "when the file path contains '/'" do
-        let(:file) { "foo/bar" }
+      context "when the file path contains '/'", ssh: true do
+        skip "The transport App does not run on windows" if Bolt::Util.windows?
```
ekinanp, line 869: This message tells me that more tests should be skipped, but we are not doing that. Maybe reword to explain why _this_ test is skipped?

McdonaldSeanp: Oh, whoops

I'm just going to remove this

ekinanp (approved):

Issue #2187, opened 2020-09-17
Add support for declaring git modules in project config


beechtom:
## Use Case
The `modules` key only supports modules that can be installed from the Puppet Forge. However, a user may have a module that is a git repo that they would like to add as a dependency of their project.

## Describe the Solution You Would Like
The `modules` key should accept modules that declare a git module. A git module accepts the following keys:

- `name`: The name of the module
- `git`: The URL to the public git repo
- `ref`: Specifies the Git reference to check out

## Questions
- If none of the reference keys is present, should Bolt error or default to the same branch as r10k (i.e. master)? We should error if no ref key is present (or if the ref key is present without a git key).
- We should also error if we parse a Puppetfile and it has a git module without a ref.
- If a user supplies a version requirement and git we should fail
- `module add` does not support git modules for now
- name is valid, used to find mod in puppetfile
- Puppetfile should always be created with a commit or tag, not a branch (if the ref is a branch it should be resolved in puppetfile and pinned). (explain module.lock thoughts in docs)
- When checking if a Puppetfile satisfies the module declarations, only compare `name` and `git`. Comparing the `ref` is not supported.

PR #2264, opened 2020-10-16
(GH-2187) Add Git module support for module management feature

beechtom:
This adds support for Git modules to the module management feature. Git
modules can now be specified in `bolt-project.yaml` using the following
format:

```
modules:
  - git: <GIT REPO URL>
    ref: <GIT REF>
```

When a Git module is specified, Bolt will first resolve its name from
the module's `metadata.json` and resolve the SHA for the specified
`ref`. If the URL to the repo does not point to a valid module, Bolt
will error.

After resolving the Git module's name and SHA, Bolt will resolve modules
using the `puppetfile-resolver` library. If resolution succeeds, a
Puppetfile with the Git module will be generated.

This also includes a significant refactoring of the `Bolt::Puppetfile`
class and the introduction of new `Spec` and `Module` classes. These
latter classes are used to represent module specs in `bolt-project.yaml`
and resolved modules in the Puppetfile, respectively. The
`Bolt::Puppetfile` class uses these two classes to determine which
modules need to be resolved, and whether any specs from project config
are unsatisfied. The `Bolt::Puppetfile` class needs access to both
classes primarily to support Git modules, as the `puppetfile-resolver`
library does not return the Git repo URL or SHA after resolving modules.

> **Note:** Only modules from GitHub are supported, as the
> `puppetfile-resolver` library only supports GitHub modules.

!feature

* **Git module support for module management feature**
  ([#2187](https://github.com/puppetlabs/bolt/issues/2187))

  Git modules can now be specified in `bolt-project.yaml` and used with
  the module management feature. Only GitHub modules are supported.


lib/bolt/puppetfile.rb
```diff
@@ -1,27 +1,32 @@
...
 module Bolt
   class Puppetfile
-    attr_reader :modules
+    attr_reader :modules, :unsatisfied_specs
 
     def initialize(modules = [])
-      @modules = Set.new
-      add_modules(modules)
+      @modules           = modules
+      @satisfied_specs   = []
```
nicklewis, line 21: It definitely seems a bit strange to hang the specs on the puppetfile object, but I don't have a great alternative.

It almost seems like `ForgeModule`/`GitModule` are to `Puppetfile` as `ForgeSpec`/`GitSpec` are to... some non-existent class _`Foo`_. And the process of resolving _`Foo`_ results in a `Puppetfile`, which contains modules objects.

lib/bolt/puppetfile.rb
```diff
@@ -33,37 +38,41 @@ def self.parse(path, skip_unsupported_modules: false)
...
       modules = parsed.modules.each_with_object([]) do |mod, acc|
-        unless mod.instance_of? PuppetfileResolver::Puppetfile::ForgeModule
-          next if skip_unsupported_modules
-
-          raise Bolt::ValidationError,
-                "Module '#{mod.title}' is not a Puppet Forge module. Unable to "\
-                "parse Puppetfile #{path}."
+        case mod.module_type
+        when :forge
+          version = mod.version.is_a?(String) ? mod.version[1..-1] : mod.version
```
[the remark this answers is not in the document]
beechtom: `[1..]` doesn't seem to work with Ruby 2.5.1 and raises a syntax error.

Non-string versions can be `nil` (no version in the Puppetfile) or `:latest`. The parser picks those two options up when parsing a Puppetfile.

lib/bolt/puppetfile/git_module.rb
```diff
@@ -0,0 +1,77 @@
+# frozen_string_literal: true
+
+require 'bolt/error'
+
+# This class represents a resolved Git module.
+#
+module Bolt
+  class Puppetfile
+    class GitModule
+      NAME_REGEX = %r{\A(?<first>[a-z][a-z0-9_]*)([-/](?<second>[a-z][a-z0-9_]*))?\z}.freeze
```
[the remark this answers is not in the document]
beechtom: 👍 That's a good improvement. I was originally capturing the owner and name, and setting both as class variables. I decided to drop owner altogether and didn't think to update the regex.

lib/bolt/puppetfile.rb
```diff
@@ -114,29 +118,98 @@ def resolve
...
-      @modules
+      # This shouldn't happen, but better safe than sorry.
+      unless git_spec
+        raise Bolt::Error.new(
+          "Could not find Git module specification with name #{name}.",
+          "bolt/missing-module-spec-error"
+        )
+      end
+
+      yield git_spec
```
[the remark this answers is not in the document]
beechtom: I was up late finishing this and didn't realize how silly that was? :P Will update.

lib/bolt/puppetfile.rb
```diff
@@ -114,29 +118,98 @@ def resolve
...
+    end
+
+    # Returns a list of modules and specs used to resolve. Raises an error if
+    # any specs share names. If a module and unsatisfied spec share a name, the
+    # unsatisfied spec takes precedence.
+    #
+    private def to_resolve
+      assert_unique_spec_names
+
+      modules = @modules.reject do |mod|
```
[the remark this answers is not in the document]
beechtom: Heh, I actually explored doing that for another reason while I was working on this. It didn't work well for what I intended, so I abandoned the idea, but it'd definitely be good here. 👍

lib/bolt/puppetfile.rb
beechtom: Ya, I'm definitely not keen about having the specs on the puppetfile object. The idea of a separate class that handles the specs and resolving (`Specifications` or something?) is interesting, though... I'll explore that a bit.

documentation/managing_modules.md
```diff
@@ -96,18 +95,72 @@ bolt module install --force
...
+| --- | --- | :-: |
+| `git` | The URI to the GitHub repository. URI must begin with either `https://github.com`or `git@github.com`. |	✓ |
+| `ref` | The Git reference to checkout. Can be either a branch, commit, or tag. | ✓ |
+
+The following example specifies the `puppetlabs/puppetlabs-puppetdb` Git module
+with a specific tag:
+
+```yaml
+modules:
+  - git: https://github.com/puppetlabs/puppetlabs-puppetdb'
```
lucywyman, line 149:
```suggestion
  - git: https://github.com/puppetlabs/puppetlabs-puppetdb
```

documentation/managing_modules.md
```diff
@@ -96,18 +95,72 @@ bolt module install --force
 Install-BoltModule -Force
 ```
 
+## Specify modules in a Bolt project
```
hestonhoffman, line 98: I think this is a little confusing if it comes before the "installing modules" section - unless we're expecting all users to manually specify their modules? Perhaps we can move it below that section?
```suggestion
## Manually specify modules in a Bolt project
```

lib/bolt/puppetfile/forge_spec.rb
```diff
@@ -0,0 +1,84 @@
...
+        @version_requirement = parse_version_requirement(version_requirement)
+        @type                = :forge
+      end
+
+      def self.from_hash(hash)
+        new(hash['name'], hash['version_requirement'])
+      end
+
+      def self.implements?(hash)
+        (hash.keys & REQUIRED_KEYS) == REQUIRED_KEYS && (hash.keys - KNOWN_KEYS).empty?
```
lucywyman, line 28: This will also verify the order which I don't think we want, since:
```
modules:
  - ref: 'main'
    git: https://github.com/mycool/module
```
will error. I think converting to sets an comparing those is most reliable, but sorting and comparing is also common ([benchmarks](https://stackoverflow.com/a/56739603/2708055)) (ooo or there's a new `difference` function when we upgrade to 2.6, though I'm not sure it's faster just maybe more readable).

documentation/managing_modules.md
```diff
@@ -96,18 +95,72 @@ bolt module install --force
...
+modules:
+  - git: https://github.com/puppetlabs/puppetlabs-puppetdb'
+    ref: '7.0.0'
+```
+
+Bolt only supports installing Git modules from GitHub.
+
 ## Install modules in a Bolt project
 
 After you've opted in, you can use the `bolt module add` and `bolt module
```
hestonhoffman, line 157:
```suggestion
After you've opted in, you can install your modules from the Bolt command line.
```
Removing *nix-specific commands from intro

documentation/managing_modules.md
```diff
@@ -96,18 +95,72 @@ bolt module install --force
...
-install` commands to manage your Forge modules. 
-
-> **Note:** Bolt only supports managing modules from the public Puppet
-> Forge. Git modules and private Forge modules are not supported.
+install` commands to manage your modules.
```
hestonhoffman, line 158:
```suggestion
```

lib/bolt/puppetfile/forge_spec.rb
lucywyman: It also seems like a lot of this is copied from the ForgeSpec class - it seems like we should either have a parent class that can handle both specifications, or tailor each class to it's needs - so this for example wouldn't need to have KNOWN_KEYS

documentation/managing_modules.md
```diff
@@ -96,18 +95,72 @@ bolt module install --force
...
-> **Note:** Bolt only supports managing modules from the public Puppet
-> Forge. Git modules and private Forge modules are not supported.
+install` commands to manage your modules. 
 
 ### Add a Forge module and its dependencies to an existing project
 
-To add a single module and its dependencies to your Bolt project, use the `bolt
-module add` command. For example, to add the puppetlabs/apt module:
+To add a single Forge module and its dependencies to your Bolt project, use the
+`bolt module add` command. For example, to add the puppetlabs/apt module:
```
hestonhoffman, line 163:
```suggestion
To add a single Forge module and its dependencies to your Bolt project:
```

documentation/managing_modules.md
```diff
@@ -119,7 +172,7 @@ bolt module add puppetlabs/apt
 Add-BoltModule -Module puppetlabs/apt
 ```
 
-Bolt adds a declaration to your project configuration file, resolves
+Bolt adds a specification to your project configuration file, resolves
 dependencies, updates your project's Puppetfile, and installs the modules to the
 project's `.modules` directory.
```
hestonhoffman, line 178:
```suggestion
> **Note:** This command only works with Forge modules. To add a git module, [manually specify the module in your project configuration file](#git-modules).
```

documentation/managing_modules.md
```diff
@@ -96,18 +95,72 @@ bolt module install --force
 Install-BoltModule -Force
 ```
 
+## Specify modules in a Bolt project
+
```
hestonhoffman, line 99:
```suggestion
In most cases, you can use the Bolt command line to install a new module. 
However, there are times when you might want to manually specify a module
instead of using the Bolt command line. For example, you might 
want to add a git module, which is not supported through the command 
line, or you might want to add a large number of modules to your project 
at once and you don't want to string together multiple commands.
```
This is just to provide a little context between using `add` or manually specifying

documentation/managing_modules.md
```diff
@@ -12,11 +12,10 @@ update](developer_updates.md#september-2020).
 - **Managed module installation directory:** Bolt now installs modules it
   manages into the `.modules/` directory instead of `modules/`. Avoid committing
   `.modules/` to source control. Your users can download your Bolt project and
-  use the `bolt module install` command to download the required modules from
-  the Forge.
+  use the `bolt module install` command to download the required modules.
```
hestonhoffman, line 15:
```suggestion
  use the `bolt module install` *nix shell command, or `Install-BoltModule` cmdlet to download the required modules.
```

documentation/managing_modules.md
```diff
@@ -96,18 +95,72 @@ bolt module install --force
...
+Forges are not supported.
+
+### Git modules
+
+To specify a Git module, use the following keys in the specification:
+
+| Key | Description | Required |
+| --- | --- | :-: |
+| `git` | The URI to the GitHub repository. URI must begin with either `https://github.com`or `git@github.com`. |	✓ |
+| `ref` | The Git reference to checkout. Can be either a branch, commit, or tag. | ✓ |
```
hestonhoffman, line 142:
```suggestion
| `ref` | The git reference to checkout. Can be either a branch, commit, or tag. | ✓ |
```

documentation/managing_modules.md
```diff
@@ -96,18 +95,72 @@ bolt module install --force
...
+
+The following example specifies the `puppetlabs/puppetlabs-puppetdb` Git module
```
hestonhoffman, line 144:
```suggestion
The following example specifies the `puppetlabs/puppetlabs-puppetdb` git module
```

documentation/managing_modules.md
```diff
@@ -96,18 +95,72 @@ bolt module install --force
...
+with a specific tag:
+
+```yaml
+modules:
+  - git: https://github.com/puppetlabs/puppetlabs-puppetdb'
+    ref: '7.0.0'
+```
+
+Bolt only supports installing Git modules from GitHub.
```
hestonhoffman, line 153:
```suggestion
Bolt only supports installing git modules from GitHub.
```

documentation/managing_modules.md
```diff
@@ -96,18 +95,72 @@ bolt module install --force
...
+modules:
+  - puppetlabs/apache
+```
+
+Bolt only supports installing Forge modules from the Puppet Forge. Private
+Forges are not supported.
+
+### Git modules
+
+To specify a Git module, use the following keys in the specification:
```
hestonhoffman, line 137:
```suggestion
To specify a git module, use the following keys in the specification:
```

documentation/managing_modules.md
hestonhoffman, line 163:
```suggestion
`bolt module add` *nix shell command or the `Add-BoltModule` Powershell cmdlet. 
For example, to add the puppetlabs/apt module:
```

lib/bolt/cli.rb
```diff
@@ -922,6 +925,8 @@ def install_project_modules(project, force, resolve)
     # Adds a single module to the project.
     #
     def add_project_module(name, project)
+      require 'bolt/module_installer'
```
lucywyman, line 928: Is this pulling in something really huge? It seems fine to have this at the top...

lib/bolt/module_installer.rb
```diff
@@ -1,7 +1,11 @@
 # frozen_string_literal: true
 
-require 'bolt/error'
-require 'bolt/logger'
+require_relative 'error'
```
lucywyman, line 3: Why the require_relatives?

lib/bolt/cli.rb
beechtom: We have a test for the CLI that running a `show` or `help` command does not load the SSL library, which this does via the `GitSpec` class.

I can rework it so that SSL is only loaded when a request is actually made.

lib/bolt/module_installer.rb
beechtom: Ah, that made testing the classes in IRB easier. :P Will update.

lib/bolt/module_installer.rb
```diff
@@ -70,130 +78,92 @@ def add(name, modules, puppetfile_path, moduledir, config_path)
...
+
+        # If forcibly installing or if there is no Puppetfile, resolve
+        # and write a Puppetfile.
+        if force || !path.exist?
           @outputter.print_action_step("Resolving module dependencies, this may take a moment")
-          puppetfile.resolve
+          puppetfile = Resolver.new.resolve(specs)
 
           @outputter.print_action_step("Writing Puppetfile at #{path}")
-          # We get here either through 'bolt module install' which uses the
```
lucywyman, line 187: I think it's useful to leave this here, so that if any of the assumptions this makes changes we have some hint to know we need to change this too.

lib/bolt/module_installer/puppetfile.rb
```diff
@@ -0,0 +1,146 @@
+# frozen_string_literal: true
+
+require 'puppetfile-resolver'
+
+require_relative '../error'
+require_relative 'puppetfile/forge_module'
```
lucywyman, line 6:
```suggestion
require 'bolt/module_installer/puppetfile/forge_module'
```

lib/bolt/module_installer.rb
```diff
@@ -13,45 +17,49 @@ def initialize(outputter, pal)
...
+          "#{name}. Nothing to do."
         )
         return true
       end
 
+      @outputter.print_message("Adding module #{name} to project\n\n")
+
+      # Create a Puppetfile object. If there isn't an existing Puppetfile,
+      # this won't include any resolved modules.
+      existing_puppetfile = Puppetfile.parse(puppetfile_path)
```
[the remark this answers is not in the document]
lucywyman: `new_specs` and `existing_specs` or `old_specs` might also be better names? But up to you

lib/bolt/module_installer.rb
```diff
@@ -216,27 +184,5 @@ def install_puppetfile(path, moduledir, config = {})
 
       ok
     end
-
-    # Asserts that an existing Puppetfile is managed by Bolt.
-    #
-    private def assert_managed_puppetfile(puppetfile, path)
```
lucywyman, line 222: I like that this logic is moved to the Puppetfile object itself, so this class is really just an interface between the CLI and the Puppetfile

lib/bolt/module_installer.rb
nicklewis: If you `bundle exec irb`, regular `require` statements should work.

beechtom: TIL!

lib/bolt/module_installer/puppetfile.rb
```diff
@@ -0,0 +1,146 @@
...
+            This may not be a Puppetfile managed by Bolt.
+          MSG
+        end
+
+        new(serialize_parsed_modules(parsed.modules, skip_unsupported_modules))
+      end
+
+      # Serializes the modules returned from the parser.
+      #
+      private_class_method def self.serialize_parsed_modules(modules, skip_unsupported)
```
nicklewis, line 49: Serialization isn't quite what this is doing.

beechtom: `hashify_parsed_modules`? `parsed_modules_to_data`? 🤔

lucywyman: I'm not entirely clear on why we turn these into data instead of directly into module objects. It also seems like we only need to turn them into our own module objects because from the puppetfile-resolver they don't have all the attributes we need, but it might be easier to just PR the resolver than do all this munging?

lib/bolt/module_installer/puppetfile.rb
```diff
@@ -0,0 +1,146 @@
...
+      def assert_satisfies(specs)
+        unsatisfied_specs = specs.specs.reject do |spec|
+          @modules.any? do |mod|
+            spec.satisfied_by?(mod)
+          end
+        end
+
+        return if unsatisfied_specs.empty?
+
+        command = Bolt::Util.windows? ? 'Install-BoltModule -Force' : 'bolt module install --force'
```
lucywyman, line 131: Nice.

lib/bolt/module_installer/specs.rb
```diff
@@ -0,0 +1,93 @@
...
+          else
+            @specs.unshift spec
+          end
+        end
+      end
+
+      # Parses a spec hash into a spec object.
+      #
+      private def spec_from_hash(hash)
+        return ForgeSpec.from_hash(hash) if ForgeSpec.implements?(hash)
```
lucywyman, line 51: It seems like using a hash as an intermediary representation of this data might be unnecessary - translating objects from the resolver directly into our own objects would probably be best.

lib/bolt/module_installer/specs.rb
```diff
@@ -0,0 +1,93 @@
...
+        return GitSpec.from_hash(hash)   if GitSpec.implements?(hash)
+
+        raise Bolt::ValidationError, <<~MESSAGE.chomp
+          Unknown module specification:
```
lucywyman, line 55:
```suggestion
          Invalid module configuration:
```

lucywyman: It'd be especially great if we could have a more specific error, like which keys are invalid or which keys aren't provided, but that may be too difficult / too much magic to be worth it. This at least seems more clear.

lib/bolt/module_installer/specs.rb
beechtom: This is for the specs from project config, which are already hashes and don't come from the resolver.

lib/bolt/module_installer/puppetfile/forge_module.rb
```diff
@@ -0,0 +1,80 @@
...
+        #
+        private def parse_version(version)
+          return unless version.is_a?(String)
+
+          unless SemanticPuppet::Version.valid?(version)
+            raise Bolt::ValidationError,
+                  "Invalid version for Forge module #{@full_name}: #{version.inspect}"
+          end
+
+          [version, SemanticPuppet::Version.parse(version)]
```
lucywyman, line 56: Will 'version' ever be different from `SemanticPuppet::Version.parse(version).to_s`?

lib/bolt/module_installer.rb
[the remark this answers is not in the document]
beechtom: This change would make `existing_puppetfile` inaccessible, which is needed as a parameter to `print_puppetfile_diff` after resolving.

lib/bolt/module_installer/puppetfile.rb
beechtom: The thought was to have a single place where those objects are instantiated, but it doesn't really matter. 🤷 And they're converted to our own module objects because the resolver returns a generic `ModuleSpecification` object that includes a limited set of attributes and does not have methods that are helpful for our purposes such as `to_spec` or `to_hash`.

lib/bolt/module_installer/puppetfile/forge_module.rb
```diff
@@ -0,0 +1,80 @@
...
+        def self.implements?(hash)
+          REQUIRED_KEYS == hash.keys.to_set
+        end
+
+        # Formats the full name and extracts the module name.
+        #
+        private def parse_name(name)
+          unless (match = name.match(NAME_REGEX))
+            raise Bolt::ValidationError,
+                  "Invalid name for Forge module : #{name}. Name must match 'owner/name', "\
```
lucywyman, line 38:
```suggestion
                  "Invalid name for Forge module: #{name}. Name must match 'owner/name', "\
```

lucywyman: Wait don't we already do this in the ForgeSpecs class? How would we get here and still have this name be invalid?

lib/bolt/module_installer/puppetfile/forge_module.rb
beechtom: No. Updated to just use a single `version` var.

lib/bolt/module_installer/puppetfile.rb
beechtom: Updated to just create the module objects directly.

lib/bolt/cli.rb
```diff
@@ -17,6 +17,7 @@
 require 'bolt/inventory'
 require 'bolt/rerun'
 require 'bolt/logger'
+require 'bolt/module_installer'
```
lucywyman, line 20: This seems unnecessary - can you just `git checkout puppet/main -- lib/bolt/cli.rb`?

lib/bolt/module_installer/puppetfile.rb
```diff
@@ -0,0 +1,120 @@
...
+            Unable to parse Puppetfile #{path}:
+            #{parsed.validation_errors.join("\n\n")}.
+            This may not be a Puppetfile managed by Bolt.
+          MSG
+        end
+
+        new(to_modules(parsed.modules, skip_unsupported_modules))
+      end
+
+      # Serializes the modules returned from the parser.
```
lucywyman, line 46:
```suggestion
      # Converts PuppetfileResolver::Document::*Module to Bolt module objects
```

lib/bolt/module_installer/puppetfile.rb
```diff
@@ -0,0 +1,120 @@
...
+          when :git
+            acc << GitModule.new(
+              mod.name,
+              mod.remote,
+              mod.ref || mod.commit || mod.tag
+            )
+          else
+            unless skip_unsupported
+              raise Bolt::ValidationError,
+                    "Module '#{mod.title}' is not a Puppet Forge or Git module."
```
lucywyman, line 65:
```suggestion
                    "Cannot parse Puppetfile at #{path}, module '#{mod.title}' is not a Puppet Forge or Git module."
```
Or something, just to indicate what we were trying to do and where something went wrong.

lib/bolt/module_installer.rb
```diff
@@ -13,45 +17,49 @@ def initialize(outputter, pal)
...
-      print_puppetfile_diff(existing, puppetfile)
+      print_puppetfile_diff(existing_puppetfile, puppetfile)
 
       # Add the module to the project configuration.
       @outputter.print_action_step("Updating project configuration file at #{config_path}")
 
       data = Bolt::Util.read_yaml_hash(config_path, 'project')
       data['modules'] ||= []
-      data['modules'] <<  { 'name' => new_module.title }
+      data['modules'] <<  name
```
lucywyman, line 62: Isn't `data['modules']` the same as the original `specs` here? We shouldn't need to read the project file twice. Separately, we should rename `specs` to `project_specs`, and I think `existing_specs` to `puppetfile_specs`.

lib/bolt/module_installer.rb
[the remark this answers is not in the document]
lucywyman: Couldn't `print_puppetfile_diff` just handle the old puppetfile not existing - probably it could just default to making `old` an empty hash if it gets nil from the old puppetfile.

lib/bolt/module_installer/puppetfile.rb
```diff
@@ -0,0 +1,121 @@
...
+            This may not be a Puppetfile managed by Bolt.
+          MSG
+        end
+
+        new(to_modules(parsed.modules, path, skip_unsupported_modules))
+      end
+
+      # Converts PuppetfileResolver::Document::*Module to Bolt module objects
+      #
+      private_class_method def self.to_modules(modules, path, skip_unsupported)
```
lucywyman, line 48: Can this just get inlined to the function above since that's the only place it's used? This seems so particular to dealing with modules from the parser that I don't think it makes sense outside of the `parsed` function.

hestonhoffman (approved):
Docs look good!

lib/bolt/module_installer.rb
beechtom: We can't update `bolt-project.yaml` without loading all of it and modifying the `modules` key...

lucywyman: But haven't we already read it when we loaded the project? `specs` here should be the same as `data[modules]`

lucywyman: Oh I see what you're saying - I mean, we're passing so many things from the project class here already. We could pass data instead of `specs` and get specs from that. That'd also make it more clear where specs is coming from.

lib/bolt/module_installer/installer.rb
```diff
@@ -1,19 +1,20 @@
...
 module Bolt
-  class Puppetfile
+  class ModuleInstaller
     class Installer
       def initialize(config = {})
         @config = config
       end
 
       def install(path, moduledir)
+        require 'r10k/cli'
```
lucywyman, line 16: Do we use this? Why do we need to require it now?

lib/bolt/module_installer/puppetfile/module.rb
```diff
@@ -0,0 +1,26 @@
...
+      class Module
+        attr_reader :full_name, :name, :type
+
+        def initialize(name)
+          @full_name, @name = parse_name(name)
+        end
+
+        # Formats the full name and extracts the module name.
+        #
+        protected def parse_name(name)
```
lucywyman, line 17: I don't understand why this is different from the one in forge_module?

lib/bolt/module_installer.rb
beechtom: It's a minor thing, but `data` doesn't necessarily have everything that's in the file since we reject transport config.

Also not super keen on passing all of the project config. Reading the file a second time seems fairly minor.

lib/bolt/module_installer/puppetfile/module.rb
beechtom: Just forgot to remove it from forge_module...

lib/bolt/module_installer/installer.rb
beechtom: It's always been required (https://github.com/puppetlabs/bolt/blob/main/lib/bolt/puppetfile/installer.rb#L3), it just needs to be loaded lazily since it requires some larger gems that we don't always want to load for every Bolt command. Alternatively, we can lazily load `bolt/module_installer/installer` as part of `Bolt::ModuleInstaller.install_puppetfile`.

lib/bolt/module_installer/resolver.rb
```diff
@@ -0,0 +1,79 @@
...
+          )
+        rescue StandardError => e
+          raise Bolt::Error.new(e.message, 'bolt/module-resolver-error')
+        end
+
+        # Create the Puppetfile object.
+        Bolt::ModuleInstaller::Puppetfile.new(make_modules(result.specifications.values, specs))
+      end
+
+      # Serializes the specs returned from the resolver. This makes it a little
```
lucywyman, line 54:
```suggestion
      # Converts the specs returned from the resolver. This makes it a little
```

lib/bolt/module_installer/resolver.rb
```diff
@@ -0,0 +1,75 @@
...
+            module_paths:          [],
+            allow_missing_modules: false
+          )
+        rescue StandardError => e
+          raise Bolt::Error.new(e.message, 'bolt/module-resolver-error')
+        end
+
+        # Convert the specs returned from the resolver into Bolt module objects.
+        modules = result.specifications.values.each_with_object([]) do |mod, acc|
+          next unless mod.is_a? PuppetfileResolver::Models::ModuleSpecification
```
lucywyman, line 52: What else could this be? Modules can only have other forge modules as their dependencies.

beechtom: `PuppetfileResolver::Models::PuppetSpecification`. puppetfile-resolver automatically adds that in to the resolved specs.

lib/bolt/module_installer/specs/forge_spec.rb
```diff
@@ -0,0 +1,88 @@
...
+        attr_reader :full_name, :name, :semantic_version, :type
+
+        def initialize(name, version_requirement = nil)
+          @full_name, @name                       = parse_name(name)
+          @version_requirement, @semantic_version = parse_version_requirement(version_requirement)
+          @type                                   = :forge
+        end
+
+        def self.from_hash(hash)
+          new(hash['name'], hash['version_requirement'])
```
lucywyman, line 27: So I believe `Target` datatype has this function so it can be called directly from plans. Since this isn't a public API and this function isn't doing any validation or data munging it seems simpler to call `.new()` directly.

lib/bolt/module_installer/specs/git_spec.rb
```diff
@@ -0,0 +1,173 @@
...
+                          "variable with a GitHub personal access token."
+              end
+
+              raise Bolt::Error.new(message, 'bolt/github-api-rate-limit-error')
+            when Net::HTTPNotFound
+              raise Bolt::Error.new(
+                "#{git} is not a Git repository.",
+                "bolt/missing-git-repository-error"
+              )
+            else
```
lucywyman, line 143: This can also be a 401 `Net::HTTPUnauthorized ` if the `GITHUB_TOKEN` is invalid

spec/bolt/module_installer/puppetfile/forge_module_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+  end
+
+  context '#to_spec' do
+    it 'returns a Puppetfile spec' do
+      expect(mod.to_spec).to eq("mod '#{name}', '#{version}'")
+    end
+  end
+
+  context '#to_hash' do
+    it 'serializes the module' do
```
lucywyman, line 41:
```suggestion
    it 'converts the module' do
```

spec/bolt/module_installer/puppetfile/git_module_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+require 'bolt/module_installer/puppetfile/git_module'
+
+describe Bolt::ModuleInstaller::Puppetfile::GitModule do
+  let(:name) { 'yaml' }
+  let(:git)  { 'https://github.com/puppetlabs/puppetlabs-yaml' }
+  let(:ref)  { '79f98ffd3faf8d3badb1084a676e5fc1cbac464e' }
+  let(:mod)  { described_class.new(name, git, ref) }
+
+  context '#to_spec' do
+    it 'returns a Puppetfile spec' do
```
lucywyman, line 13: Haven't finished reviewing yet, but we should have a test that also verifies that this line gets added to a Puppetfile and then that puppetfile installs correctly if we can.

spec/bolt/module_installer/puppetfile/git_module_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+      expect(mod.to_spec).to eq(<<~SPEC.chomp)
+        mod '#{name}',
+          git: '#{git}',
+          ref: '#{ref}'
+      SPEC
+    end
+  end
+
+  context '#to_hash' do
+    it 'serializes the module' do
```
lucywyman, line 23:
```suggestion
    it 'converts the module' do
```

spec/bolt/module_installer/puppetfile/module_spec.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
```
lucywyman, line 3: I don't think this gets used here

spec/bolt/module_installer/puppetfile/git_module_spec.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
```
lucywyman, line 3: Doesn't seem to get used here

spec/bolt/module_installer/puppetfile/module_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+require 'bolt/module_installer/puppetfile/module'
+
+describe Bolt::ModuleInstaller::Puppetfile::Module do
+  let(:name) { 'puppetlabs/yaml' }
+  let(:mod)  { described_class.new(name) }
+
+  context '#initialize' do
+    it 'normalizes the full name' do
```
lucywyman, line 11: I think this should verify that `puppetlabs-yaml` gets converted to `puppetlabs/yaml`

lucywyman (commented):
I'll finish up in the morning!

spec/bolt/module_installer/puppetfile_spec.rb
```diff
@@ -0,0 +1,112 @@
...
+  let(:path)       { @project + 'Puppetfile' }
+  let(:moduledir)  { @project + '.modules' }
+  let(:name)       { 'puppetlabs/yaml' }
+  let(:version)    { '0.1.0' }
+  let(:mod)        { [double('mod', name: name, version: version, to_spec: "mod '#{name}'")] }
+  let(:puppetfile) { described_class.new(mod) }
+
+  around(:each) do |example|
+    Dir.mktmpdir(nil, Dir.pwd) do |project|
+      @project = Pathname.new(project)
```
lucywyman, line 17: Can't this use `with_project`?

spec/bolt/module_installer/resolver_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+  let(:full_name) { "puppetlabs/#{name}" }
+  let(:version)   { '0.1.0' }
+  let(:git)       { "https://github.com/puppetlabs/puppetlabs-#{name}" }
+  let(:ref)       { '0.1.0' }
+  let(:mod)       { forge_module }
+  let(:resolver)  { described_class.new }
+  let(:spec)      { double('spec', to_resolver_module: mod, name: name, git: git, sha: ref) }
+  let(:specs)     { double('specs', specs: [spec]) }
+
+  let(:forge_module) do
```
lucywyman, line 18: This doesn't seem to get used?

spec/bolt/module_installer/resolver_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+require 'spec_helper'
+require 'puppetfile-resolver'
+require 'bolt/module_installer/resolver'
+
+describe Bolt::ModuleInstaller::Resolver do
+  let(:name)      { 'yaml' }
+  let(:full_name) { "puppetlabs/#{name}" }
+  let(:version)   { '0.1.0' }
+  let(:git)       { "https://github.com/puppetlabs/puppetlabs-#{name}" }
+  let(:ref)       { '0.1.0' }
```
lucywyman, line 12: It'd also be good to test that an invalid ref, invalid git repo, and invalid version numbers fail as expected.

spec/bolt/module_installer/specs/git_spec_spec.rb
```diff
@@ -0,0 +1,151 @@
...
+      it 'resolves and returns a SHA' do
+        expect(spec.sha).to be_a(String)
+      end
+    end
+
+    context 'with a valid tag' do
+      let(:ref) { '0.2.0' }
+
+      it 'resolves and returns a SHA' do
+        expect(spec.sha).to be_a(String)
```
lucywyman, line 117: What string do we expect it to be?

spec/bolt/module_installer/specs/git_spec_spec.rb
```diff
@@ -0,0 +1,151 @@
...
+      end
+    end
+  end
+
+  context '#resolve_sha' do
+    context 'with a valid commit' do
+      let(:ref) { '79f98ffd3faf8d3badb1084a676e5fc1cbac464e' }
+
+      it 'resolves and returns a SHA' do
+        expect(spec.sha).to be_a(String)
```
lucywyman, line 109: Same here, we should check the content

lucywyman: Heh, inspecting this I get `Bolt::Error: Ref 79f98ffd3faf8d3badb1084a676e5fc1cbac464e at https://github.com/puppetlabs/puppetlabs-yaml is not a commit, tag, or branch.` which *definitely* doesn't seem like what we want.

spec/bolt/project_migrator/modules_spec.rb
```diff
@@ -133,7 +133,7 @@ def make_moduledirs
       expect(migrate).to be(true)
       data = Bolt::Util.read_yaml_hash(project.project_file, 'project')
       expect(data['modules']).to match_array([
-                                               { 'name' => 'puppetlabs-yaml', 'version_requirement' => '0.1.0' }
+                                               { 'name' => 'puppetlabs/yaml', 'version_requirement' => '0.1.0' }
```
lucywyman, line 136: It's probably useful to verify that both formats work, since they're both supported. It's probably best to have a test for that in on of the module_installer spec tests, but doing it here would also work.

lucywyman (changes requested):
I think we should also add some integration tests for this - they can stub the API calls so that they don't take as long (and we don't hit rate limits), but should verify a few full codepaths. Having a test that verifies both a forge module and git module in the `modules` key would be good and successfully installs the modules would be good, and I think verifying that a nonexistent git repo fails correctly would also be good. I'd also add one that verifies that a combination of repos that can't be resolved fails as expected.

spec/bolt/module_installer/resolver_spec.rb
beechtom: Non-existent repo/ref is already tested here: https://github.com/puppetlabs/bolt/pull/2264/files#diff-318373336c66bcd156170adb3db5045a6eb94dadb645af6503d936a04781c9a7R129-R149

spec/bolt/module_installer/specs/git_spec_spec.rb
beechtom: Not sure why you're getting that... it's definitely a valid ref.

spec/bolt/project_migrator/modules_spec.rb
beechtom: That's not what this test is checking. Since the module classes normalize the name, the project migrator will _always_ output names that match `owner/name`, it's not a matter of supporting `owner-name`, too.

The base `module` class already tests that `owner-name` is normalized, which seems sufficient for testing on our end, as `puppetfile-resolver` is what ultimately needs to support both formats and test that it works. The name is just normalized on our end so the Puppetfile output at least looks consistent. :P

spec/bolt/module_installer/resolver_spec.rb
beechtom: Line 13 sets `forge_module` as the value for `mod`, and line 15 sets `mod` as the return value for the `to_resolver_module` stub.

spec/bolt/module_installer/puppetfile_spec.rb
beechtom: Ya. :) I think this is just left over from before I made `with_project`.

spec/bolt/module_installer/puppetfile/git_module_spec.rb
beechtom: Added as part of the integration test in cli_spec

beechtom:
Added a few integration tests to cli_spec that test the full code path. Decided not to stub the API calls since the modules being used have few dependencies and don't take much time to resolve.

bolt-modules/boltlib/spec/functions/run_task_spec.rb
```diff
@@ -356,9 +356,9 @@ def mock_task(executable, input_method)
 
       is_expected.to run.with_params(task_name, hostname, task_params).and_raise_error(
         Puppet::ParseError,
-        /Task\ test::params:\n
+        %r{Task\ test::params:\n
          \s*has\ no\ parameter\ named\ 'foo'\n
-         \s*has\ no\ parameter\ named\ 'bar'/x
+         \s*has\ no\ parameter\ named\ 'bar'}x
```
beechtom, line 361: These changes are just to make cli_spec a little more readable for VS Code users, as there's a bug with the Ruby plugin's syntax highlighting when you use multi-line regex literals.

spec/bolt/cli_spec.rb
```diff
@@ -269,7 +269,7 @@ def stub_config(file_content = {})
         cli.execute(cli.parse)
       end
 
-      it 'install modules from Puppetfile with resolving' do
+      it 'install modules from Puppetfile without resolving' do
```
lucywyman, line 272: I think a test in `spec/integration` might help keep this file a bit more manageable, especially since we're just checking file contents and errors. But basically the same idea!

beechtom: Moved the new tests to `integration/module_installer_spec`

PR #2265, opened 2020-10-16
(docs) Add PowerShell cmdlet reference

beechtom:
This adds a new PowerShell cmdlet reference page that is similar in
structure to the shell command reference page.

!no-release-note


rakelib/lib/pwsh.rb
```diff
@@ -0,0 +1,347 @@
+# frozen_string_literal: true
```
beechtom, line 1: I'm not sure if this file should be somewhere other than `rakelib/lib/pwsh.rb`...

documentation/README.md
```diff
@@ -28,7 +28,7 @@ Here are each of the documenation pages that are generated, the files
 that you may need to modify to make changes to the generated documentation,
 and the rake task used to generate the documentation.
 
-### Bolt command reference
+### Shell command reference
```
lucywyman, line 31:
```suggestion
### *nix command reference
```

Or something? Powershell is also a shell :P

rakelib/lib/pwsh.rb
lucywyman: I think the 'right' way to do this is to put this in it's own rake task, and make that task a prerequisite for the two pwsh tasks.

```
task cmdlet_reference: :generate_pwsh_cmdlets do
...
```

documentation/README.md
beechtom: Tbf, 'Shell command' isn't inaccurate since they can still be used in PowerShell. :P Probably best to have consistency with the command snippets in the docs, though.

rakelib/pwsh.rake
```diff
@@ -1,12 +1,44 @@
...
-  task :generate_module do
+  task generate_module: :generate_powershell_cmdlets do
+    # Move 'guides' to 'en-US' directory in module, renaming the text files
+    # so they are recognized by the PowerShell help system
+    source = File.expand_path(File.join(__dir__, '..', 'guides'))
+    dest   = File.expand_path(File.join(__dir__, '..', 'pwsh_module', 'en-US'))
+
+    FileUtils.mkdir(dest) unless File.exist?(dest)
+
+    Dir.children(source).each do |file|
```
lucywyman, line 16: I think this can just be `Dir.glob("*.txt", source)`

rakelib/pwsh.rake
```diff
@@ -1,12 +1,44 @@
...
+  task generate_module: :generate_powershell_cmdlets do
+    # Move 'guides' to 'en-US' directory in module, renaming the text files
+    # so they are recognized by the PowerShell help system
+    source = File.expand_path(File.join(__dir__, '..', 'guides'))
+    dest   = File.expand_path(File.join(__dir__, '..', 'pwsh_module', 'en-US'))
+
+    FileUtils.mkdir(dest) unless File.exist?(dest)
+
+    Dir.glob('*.txt', base: source).each do |file|
+      next if file !~ /\.txt\z/
```
lucywyman, line 17: Don't need this if we're globbing

beechtom: 🤦

PR #2269, opened 2020-10-20
(bug-fix) Use file methods instead of Pathname methods in the outputter

lucywyman:
Previously, the inventoryfile passed to the outputter for printing when
running `bolt inventory show` with the Human outputter could be either a
string or a Pathname, depending on whether the user had configured the 
inventory location (string) or was using the default project inventory
(Pathname). We erroneously assumed it would always be a
Pathname, which resulted in a 'method not found' error when we tried
calling `exist?` on the parameter. This uses the `File.exist?` method to
verify the path exists which will work for either object passed to the 
method.

!bug

* **Fix 'method not found' error running 'bolt inventory show'**

  Previously, when running 'bolt inventory show' with a configured
  inventory path and the human format a 'method not found' error was 
  raised. This now correctly prints the targets in the inventory.

### DIFF
--- a/lib/bolt/outputter/human.rb
+++ b/lib/bolt/outputter/human.rb
@@ -344,7 +344,7 @@ module Bolt
         end
 
         @stream.puts "INVENTORY FILE:"
-        if inventoryfile.exist?
+        if File.exist?(inventoryfile)
           @stream.puts inventoryfile
         else
           @stream.puts wrap("Tried to load inventory from #{inventoryfile}, but the file does not exist")
--- a/spec/bolt/outputter/human_spec.rb
+++ b/spec/bolt/outputter/human_spec.rb
@@ -374,7 +374,7 @@ describe "Bolt::Outputter::Human" do
   end
 
   context '#print_targets' do
-    let(:inventoryfile) { double('inventoryfile', to_s: '/path/to/inventory', exist?: true) }
+    let(:inventoryfile) { '/path/to/inventory' }
 
     let(:target_list) do
       {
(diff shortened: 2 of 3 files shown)


nicklewis:
Given that we can use the `File` methods and support both strings and Pathname objects, it seems better for methods that don't control their inputs to use those versions rather than to place a more strict requirement on the caller. That's doubly true for methods we expect to be used as part of the public API for Bolt. I would prefer to only use Pathname methods in cases where the lifecycle is controlled by the object using the path, so we can be sure it will always be a Pathname, rather than in some cases shifting that onus onto the caller.

lucywyman:
Hm I didn't think about using the outputter publicly - where is it used from outside of Bolt? My thinking was that we should standardize on using pathnames everywhere, and so this was just a stop-gap until we could do that work. Maybe not though, and we just need to always assume that a path could be a pathname *or* a string unless we know for sure and it's an internal method?

lib/bolt/project.rb
```diff
@@ -25,7 +25,7 @@ def self.default_project(logs = [])
       create_project(File.expand_path(File.join('~', '.puppetlabs', 'bolt')), 'user', logs)
     # If homedir isn't defined use the system config path
     rescue ArgumentError
-      create_project(Bolt::Config.system_path, 'system', logs)
+      create_project(Bolt::Config.system_ath, 'system', logs)
```
nicklewis, line 28: That looks like a typo

lib/bolt/outputter/human.rb
```diff
@@ -344,7 +344,7 @@ def print_targets(target_list, inventoryfile)
         end
 
         @stream.puts "INVENTORY FILE:"
-        if inventoryfile.exist?
+        if File.exist?(inventoryfile.to_s)
```
nicklewis, line 347: `to_s` shouldn't be necessary here.

nicklewis: I think I now understand your inclination toward Pathname. It would indeed be a pain if we had to call `to_s` everywhere (the same pain as having to call `Pathname.new` everywhere).

`Pathname` can actually be used _interchangeably_ with strings for the `File` methods, which is why I prefer it for the cases where we aren't using anything that's `Pathname`-specific (`parent`, `relative_path_from`, `+`).

lucywyman: Ah, ok - yeah, I do hate that when I print a pathname it doesn't know that I just want the path, not `<Pathname:/path>`. Assumed it needed the to_s, since it doesn't make we're fine using both mixed for now. I was thinking we could just ensure all paths in the Config class are pathnames, similar to in Project, but it's definitely unnecessary there since we aren't building paths. It would just let us make more assumptions like the one here.

nicklewis (approved):

beechtom (approved):

PR #2270, opened 2020-10-20
(maint) Don't generate pwsh docs in pwsh rake task

lucywyman:
We now generate the PowerShell cmdlet reference as part of the
documentation generation rake task instead of as part of generating the
Powershell module, since all docs generation happens as part of that
task (`docs:all`). We removed the old powershell reference docs template
in favor of the new one that matches other command reference templates,
but neglected to delete the section of the powershell task that fills in
the now-deleted template. This removes the docs generation from the
powershell module rake task, so that it doesn't error.

!no-release-note(maint) Don't generate pwsh docs in pwsh rake task


beechtom (approved):

beechtom (changes requested):
https://github.com/puppetlabs/bolt/blob/main/.github/workflows/pwsh.yaml#L9

We should update the paths to include `rakelib/pwsh.rake`

PR #2271, opened 2020-10-20
(Docs) Harmful terminology changes

hestonhoffman:
I grepped through the docs and found a couple of lingering terms
that we need to change. Most are just examples. One is a DB
filename that is going to change soon. I'm on the fence about merging
this now or waiting for the Puppet 7 release when all of the other docs
changes (and a lot of the engineering work) will land.

!no-release-note

### DIFF
--- a/documentation/bolt_installing.md
+++ b/documentation/bolt_installing.md
@@ -194,9 +194,10 @@ Use one of the supported *nix installation methods to install Bolt.
 
 > **CAUTION:** These instructions include enabling the Puppet Tools repository.
 > While Bolt can also be installed from the Puppet 6 or 5 platform repositories,
-> adding these repositories to a Puppet-managed target, especially a PE master,
-> might result in an unsupported version of a package like `puppet-agent` being
-> installed. This can cause downtime, especially on a PE master.
+> adding these repositories to a Puppet-managed target, especially a 
+> Puppet server, might result in an unsupported version of a package like 
+> `puppet-agent` being installed. This can cause downtime, especially on a 
+> Puppet server.
 
 ### Install Bolt on Debian or Ubuntu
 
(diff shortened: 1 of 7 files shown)


hestonhoffman:
Also, no idea why that config schema changed so much 😮

documentation/bolt_installing.md
```diff
@@ -194,9 +194,10 @@ Use one of the supported *nix installation methods to install Bolt.
 
 > **CAUTION:** These instructions include enabling the Puppet Tools repository.
 > While Bolt can also be installed from the Puppet 6 or 5 platform repositories,
-> adding these repositories to a Puppet-managed target, especially a PE master,
-> might result in an unsupported version of a package like `puppet-agent` being
-> installed. This can cause downtime, especially on a PE master.
+> adding these repositories to a Puppet-managed target, especially a primary
```
nicklewis, line 197: I think is actually true for "Puppet server", without the "primary" qualification.

nicklewis:
These changes are small enough that I don't think it's necessary to wait to merge them. We're not really adopting any _new_ terminology here.

nicklewis:
Oh I missed the reference to `certificate-allowlist` which does introduce something new. :shrug:

beechtom:
Looks like something funky happened with the `bolt-config` schema and the rake task needs to be run again.

hestonhoffman:
@beechtom Running the rake task again fixed that schema, thanks!

hestonhoffman:
> Oh I missed the reference to `certificate-allowlist` which does introduce something new. 🤷

The change should be live in today's PuppetDB release, but not sure whether installer has adopted it yet

nicklewis:
I'm not sure how much the whitelist -> allowlist change here matters with respect to whether the other docs/tools have been updated. Even if they have, most users are still going to be on an older version of PuppetDB where it's called whitelist. So either way, the docs are going to be using bleeding edge terminology that may not match their experience. So may as well :shipit: I think.

beechtom (approved):

hestonhoffman:
Okay sweet, let me rebase this quick and then

PR #2273, opened 2020-10-21
(PE-30382) Add project to task schema in bolt-server

donoghuc:
This commit updates the task metadata schema in bolt server to accept either an environment or project source for a task.

### DIFF
--- a/lib/bolt_server/schemas/partials/task.json
+++ b/lib/bolt_server/schemas/partials/task.json
@@ -63,9 +63,24 @@
                   "environment": {
                     "description": "Environment the task is in",
                     "type": "string"
+                  },
+                  "project": {
+                    "description": "Project the task is in",
+                    "type": "string"
                   }
                 },
-                "required": ["environment"],
+                "oneOf": [
+                  {
+                    "required": [
+                      "environment"
+                    ]
+                  },
+                  {
+                    "required": [
+                      "project"
+                    ]
+                  }
+                ],              
                 "additionalProperties": true
               }
             },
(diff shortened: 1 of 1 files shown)


McdonaldSeanp (approved):

lib/bolt_server/schemas/partials/task.json
```diff
@@ -63,9 +63,24 @@
                   "environment": {
                     "description": "Environment the task is in",
                     "type": "string"
+                  },
+                  "project": {
```
ekinanp, line 67: Do we need to be this thorough with our validation? I'd say it's good enough to make sure there is a `params` key, kind of like what we're doing in https://github.com/puppetlabs/bolt/blob/main/lib/bolt_server/schemas/action-run_script.json#L19. Also, if we go this route, we'll need to remember to change `project` => `project_ref` once Puppetserver (hopefully) does the same.

I also don't have strong feelings on this so feel free to ignore me too.

donoghuc: Given `environmetn` or `project` (at least currently its `project`) are required for the `FileCache` to get the file from puppetserver i think it makes sense to enforce that in the schema. An example below where relaxing the schema is https://github.com/puppetlabs/bolt/pull/2273/files#diff-2bc7fe7bbffc46aac3159f6cb11559859adb0875e0dd46408350c36a6df26ab8R108 in that case we are allowing the `allowed` key through as it can safely be ignored.

ekinanp: Ok, like I said I don't have strong feelings about this so I'll resolve the comment.

barriserloth (approved):
Looks like the linting issue is unrelated to this PR

PR #2274, opened 2020-10-21
(maint) Bump puppet version to 6.19.0

lucywyman:
This bumps the maximum allowed Puppet gem version to 6.19, while still allowing for 6.18 to be installed as there are no required changes for us to pick up in the new Puppet version.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -113,6 +113,9 @@ Lint/TopLevelReturnWithArgument:
 Lint/TrailingCommaInAttributeDeclaration:
   Enabled: true
 
+Layout/TrailingWhitespace:
+  AllowInHeredoc: true
+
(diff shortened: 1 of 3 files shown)


.rubocop.yml
```diff
@@ -113,6 +113,11 @@ Lint/TopLevelReturnWithArgument:
 Lint/TrailingCommaInAttributeDeclaration:
   Enabled: true
 
+Layout/TrailingWhitespace:
+  AllowInHeredoc: true
+  Exclude:
+    - lib/bolt/bolt_option_parser.rb
```
beechtom, line 119:
```suggestion
```

This isn't needed with `AllowInHeredoc: true`.
We should also update the Gemfile to use rubocop ~> 1.0.

Issue #2241, opened 2020-10-07
Printing errors from catch_errors fails with 'Stack level too deep'


lucywyman:
## Describe the Bug
When printing errors returned from `catch_errors`, instead of printing the error it fails with "stack level too deep". This is a bug in Puppet when printing `Puppet::PreformattedError` types and converting them to a string.

## Expected Behavior
The actual error message should be printed.

## Steps to Reproduce
Plan
```
plan bolt(
  TargetSpec $targets = 'localhost'
) {
  $f = catch_errors() || {
    run_task('bolt::err', $targets)
  }
  out::message($f)
}
```

Task
```
exit 1
```

For objects the outputter has a special output for, the outputter should have the format for that. For right now, that is only plan result objects. Everything else should be a string. We should convert anything that isn't one of those two classes to the class in the `out::message` function

Issue #2111, opened 2020-08-24
Fact collisions with topscope plan vars can cause hard to debug apply failures


donoghuc:
## Describe the Bug
With an inventory like:
```yaml
targets:
  - name: localhost
    facts:
      collision: foo
```
and a plan like:
```
plan test() {
  $collision = [one, two]
  $local_ref = $collision
  return apply(localhost){
    notify { String($local_ref): }
  }
}
```
The apply fails with:
```
cas@cas-ThinkPad-T460p:~/working_dir/bolt$ bolt plan run test
Starting: plan test
Starting: apply catalog on localhost
Finished: apply catalog with 1 failure in 3.09 sec
Finished: plan test in 3.1 sec
Failed on localhost:
  Apply failed to compile for localhost: Evaluation Error: Empty string title at 0. Title strings must have a length greater than zero. (file: /home/cas/working_dir/bolt/Boltdir/site/test/plans/init.pp, line: 577, column: 14) on node localhost
Failed on 1 target: localhost
Ran on 1 target
```
The serialized plan var `$local_ref` points to the `$collision` in plan scope which is blown away by conflicting fact value 
```
["local_ref", {"__ptype"=>"LocalRef", "__pvalue"=>"$['collision']"}]
```
## Expected Behavior
We attempt to warn when a fact conflicts with a plan scope variable https://github.com/puppetlabs/bolt/blob/main/lib/bolt/catalog.rb#L129 however that warning is never surfaced when https://github.com/puppetlabs/bolt/blob/main/lib/bolt/catalog.rb#L110 errors. 

Maybe we should iterate through the plan vars and determine if they point to a `LocalRef` that will be shadowed by a fact and raise an error if that is the case. 

## Solution

We should at least find somewhere to print a better warning that this is happening.
We should add this to our known issues page for now.
We should timebox fixing this in Puppet to 2 days, and if an end is not in sight we can leave it with just  warning. Expected behavior is to have the $local_ref plan variable reference `['one', 'two']`

PR #2277, opened 2020-10-21
(GH-2111) Shadow fact/variable collisions in PAL catalog compiler

beechtom:
This updates the `Bolt::Catalog.compile_catalog` method to pass target
and plan variables to the catalog compiler, which will then deserialize
plan variables and shadow any variables that collide with facts of the
same name. This change fixes a bug where a plan variable 'foo' that
collided with a fact 'foo' would not be passed to the compiler, so any
plan variable 'bar' that had a local reference to 'foo' would not be
deserialized correctly.

Blocked by https://github.com/puppetlabs/puppet/pull/8412

!bug

* **Correctly shadow fact/variable collisions in apply blocks**
  ([#2111](https://github.com/puppetlabs/bolt/issues/2111))

  Bolt now correctly shadows target and plan variables that collide with
  facts of the same name when running apply blocks.

### DIFF
--- a/lib/bolt/catalog.rb
+++ b/lib/bolt/catalog.rb
@@ -76,31 +76,27 @@ module Bolt
       # is the only way to log a message that will make it back to Bolt
       # to be printed.
       target = request['target']
-      plan_vars = shadow_vars('plan', request['plan_vars'], target['facts'])
-      target_vars = shadow_vars('target', target['variables'], target['facts'])
 
-      # Merge plan vars with target vars, while maintaining the order of the plan
-      # vars. It's critical that the order of plan vars is not changed, as Puppet
-      # will deserialize the variables in the order they appear. Variables may
-      # contain local references to variables that appear earlier in a plan. If
-      # these variables are moved before the variable they reference, Puppet will
-      # be unable to deserialize the data and raise an error.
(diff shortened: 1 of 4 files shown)


lucywyman:
Closes #2241

.github/workflows/docs.yaml
```diff
@@ -31,7 +31,7 @@ jobs:
           path: vendor/bundle
           key: ${{ runner.os }}-gems-${{ hashFiles('lib/bolt/version.rb') }}-${{ hashFiles('Gemfile') }}-${{ hashFiles('bolt.gemspec') }}
       - name: Install gems
-        if: steps.cache.outputs.cache-hit != 'true'
+        # if: steps.cache.outputs.cache-hit != 'true'
```
lucywyman, line 34: This seems leftover from testing?

lucywyman:
Woooo I'm not sure when Puppet 7 is getting released today, but once it is we should be able to get this in finally!

lucywyman:
This should be mergeable once Puppet 6.20 is released, if you want to do a quick rebase so we can merge once we update.

lucywyman:
And actually can just remove the `bolt.gemspec` changes, and this will just depend on getting 6.20 in to puppet-runtime.

Issue #2219, opened 2020-09-28
Add 'installing-modules' doc for 'bolt guide'


lucywyman:
## Use Case
Add a short guide for 'module' describing bolt module subcommands and workflow and linking to docs

PR #2279, opened 2020-10-22
(GH-2219) Add 'module' and 'modulepath' topic guides

beechtom:
This adds a new 'module' topic guide. It gives a brief description of
what a module is and how Bolt can manage a project's modules.

This also adds a new 'modulepath' topic guide that describes what the
modulepath is, precedence of modules on the modulepath, and how Bolt
adds directories to the modulepath.

!no-release-note

### DIFF
--- a/guides/module.txt
+++ b/guides/module.txt
@@ -1,0 +1,19 @@
+TOPIC
+    module
+
+DESCRIPTION
+    Modules are shareable, reusable packages of Puppet content. They can include
+    tasks, plans, functions, and other types of content that you can use in your
+    project. You can download and install modules to your project from the
+    Puppet Forge or write your own modules. Bolt also ships with several helpful
+    modules pre-installed that are available to all of your projects.
+
+    Bolt makes it easy to manage the modules that your project depends on. You
+    can use Bolt commands to install a project's modules, add new modules to a
+    project, and view the modules that are available to the project.
+    
+    To learn more about managing modules in a project, see the documentation.
+    To learn how modules are loaded by Bolt, see the 'modulepath' guide.
+
+DOCUMENTATION
+    https://pup.pt/bolt-modules
--- a/guides/modulepath.txt
+++ b/guides/modulepath.txt
@@ -1,0 +1,25 @@
+TOPIC
+    modulepath
+
+DESCRIPTION
+    The modulepath is an ordered list of directories that Bolt loads modules
+    from. When Bolt runs a command, it automatically loads modules from the
+    modulepath.
+    
+    While Bolt has a default modulepath, you can also configure your own
+    modulepath, which can include directories within the project or directories
+    elsewhere on your system. Regardless of whether your project uses a default
+    or configured modulepath, Bolt automatically adds directories to the
+    modulepath. This includes modules containing core Bolt content, which is
+    added to the beginning of the modulepath, and bundled content, which is
+    added to the end of the modulepath.
+
+    Modules loaded from a directory listed earlier in the modulepath take
+    precedence over modules with the same name loaded from a directory later in
+    the modulepath. Bolt will not warn or error when two modules share a name
+    and instead will ignore modules with a lower precedence.
+
+    To learn more about modules, see the 'module' guide.
+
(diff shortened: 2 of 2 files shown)


guides/module.txt
```diff
@@ -0,0 +1,23 @@
...
+    module
+
+DESCRIPTION
+    Modules are shareable, reusable packages of Puppet content. They can include
+    tasks, plans, functions, and other types of content that you can use in your
+    project. You can download and install modules to your project from the
+    Puppet Forge or write your own modules. Bolt also ships with several helpful
+    modules pre-installed that are available to all of your projects.
+
+    When Bolt runs a command, it automatically loads modules from specific
```
lucywyman, line 11: It might be useful to just have a separate modulepath guide. As a new person, it seems like it'd be confusing to think about both the `modules` configured in the project and the modulepath. Something like "to learn more about how modules are loaded, run 'bolt guide modulepath'" and then getting more in depth about the default path and what special paths are (i.e. `modules` is for custom modules, `.modules` is where modules get installed, etc.).

hestonhoffman: Oh that's a great idea!

lucywyman (changes requested):
Breakout modulepath into separate guide.

beechtom:
Added a modulepath guide. Left out listing specific directories on the default modulepath since that gets more into documentation and is also made a little more confusing because we currently have two default modulepaths.

guides/modulepath.txt
```diff
@@ -0,0 +1,25 @@
...
+
+DESCRIPTION
+    The modulepath is an ordered list of directories that Bolt loads modules
+    from. When Bolt runs a command, it automatically loads modules from the
+    modulepath.
+    
+    While Bolt has a default modulepath, you can also configure your own
+    modulepath, which can include directories within the project or directories
+    elsewhere on your system. Regardless of whether your project uses a default
+    or configured modulepath, Bolt automatically adds a couple directories to
```
hestonhoffman, line 12:
```suggestion
    or configured modulepath, Bolt automatically adds two directories to
```
It's just the two now right? If it's a small, but undefined number, I would maybe use different phrasing - For example "certain directories" / "specific directories" or forego the descriptor and just use "...adds directories...".

beechtom: _ish_. It adds `.modules` if you're using the module management workflow, and the project if you're using a named project. I'll be less specific.

guides/modulepath.txt
```diff
@@ -0,0 +1,25 @@
...
+
+DESCRIPTION
+    The modulepath is an ordered list of directories that Bolt loads modules
+    from. When Bolt runs a command, it automatically loads modules from the
+    modulepath.
+    
+    While Bolt has a default modulepath, you can also configure your own
+    modulepath, which can include directories within the project or directories
+    elsewhere on your system. Regardless of whether your project uses a default
+    or configured modulepath, Bolt automatically adds  directories to the
```
hestonhoffman, line 12:
```suggestion
    or configured modulepath, Bolt automatically adds directories to the
```

guides/modulepath.txt
```diff
@@ -0,0 +1,25 @@
...
+    modulepath. This includes modules containing core Bolt content, which is
+    added to the beginning of the modulepath, and bundled content, which is
+    added to the end of the modulepath.
+
+    Modules loaded from a directory listed earlier in the modulepath take
+    precedence over modules with the same name loaded from a directory later in
+    the modulepath. Bolt will not warn or error when two modules share a name
+    and instead will shadow modules with a lower precedence.
```
lucywyman, line 20:
```suggestion
    and instead will ignore modules with a lower precedence.
```

lucywyman (commented):
One nit, otherwise LGTM!

lucywyman (approved):

lucywyman:
Will merge once Heston approves!

guides/modulepath.txt
hestonhoffman: Ah, good catch!

hestonhoffman (approved):

PR #2280, opened 2020-10-23
(GH-2241) Handle printing preformatted Puppet errors with out::message

beechtom:
This updates the `out::message` plan function to handle printing
preformatted Puppet errors (and other data types that may not stringify
correctly after being passed to the executor). Previously, if a
preformatted Puppet error was printed using `out::message`, Bolt would
error with a "Stack level too-deep" error. Now, the process of
stringifying data types is handled in the `out::message` function itself
before being sent to the executor, which prevents this error from
occuring.

!bug

* **Handle printing preformatted Puppet errors with `out::message`**
  ([#2241](https://github.com/puppetlabs/bolt/issues/2241))

  The `out::message` plan function now correctly prints preformatted
  Puppet errors. Previously, printing preformatted Puppet errors would
  result in a 'stack level too deep' error.


lucywyman:
Closes #2241  (I hope this works? It might need to be in the commit body :P)

bolt-modules/out/spec/functions/out/message_spec.rb
```diff
@@ -2,12 +2,36 @@
...
 require 'bolt/executor'
-require 'bolt/target'
+require 'bolt/inventory'
 
 describe 'out::message' do
-  let(:executor) { Bolt::Executor.new }
-  let(:events) { [] }
-  let(:outputter) { stub('outputter') }
+  let(:executor)  { Bolt::Executor.new }
+  let(:events)    { [] }
```
lucywyman, line 9: I don't think this gets used

lucywyman (approved):

PR #2282, opened 2020-10-23
(docs) Verbose is an output level, not logging

lucywyman:
This clarifies that `--verbose` doesn't indicate a log level but
instructs Bolt to output more data through the outputter. Since #955,
the Bolt outputter and logger have a clear division of responsibility:
the outputter is responsible for user-facing messages, like progress and
results, while the logger is responsible for log messages like
warnings, errors, and logs from plans. Verbose is an outputter-specific
concept that prints more messages (or fewer with `--no-verbose`) through
the outputter.

!no-release-note


hestonhoffman (approved):
LGTM!

beechtom (approved):

PR #2284, opened 2020-10-26
(GH-2281) Add logging guide

### LINKED ISSUE #2281: Write bolt guide on logging
A clear and concise description of what you would like changed.

lucywyman:
This adds a brief guide to logging in Bolt, clarifying that there are
two output methods that are separate, and both configurable. This should
help clarify the difference between `--verbose` and `--log-level` for
users.

Note: This is from the hack day, I just forgot to demo + push

Closes #2281

!no-release-note

### DIFF
--- a/guides/logging.txt
+++ b/guides/logging.txt
@@ -1,0 +1,18 @@
+TOPIC
+    logging
+
+DESCRIPTION
+    Bolt prints messages both to the console and to log files. Messages can
+    either come from Bolt's 'outputter', which logs user-facing messages like
+    progress and results, or from the 'logger', which logs warnings, errors, and
+    log-structured output to log files. Both of these message streams are
+    configurable.
+
+    By default, Bolt logs to the console at 'warn' level and writes a log file to
+    '<project>/bolt-debug.log' at 'debug' level. Unless you are running a plan,
+    Bolt runs in verbose mode by default.
+
+    To learn more about projects, see the 'project' guide.
+
+DOCUMENTATION
+    https://pup.pt/bolt-logging


guides/logging.txt
```diff
@@ -0,0 +1,24 @@
...
+    streams are configurable.
+
+    By default, Bolt logs to the console at 'warn' level, writes a log file to
+    '<project>/bolt-debug.log' at 'debug' level, and runs with the outputter in
+    verbose mode unless you are running a plan, in which case it runs without
+    verbose (you still with me?). You can configure logging output in your
+    project configuration, and can configure console log levels (but not file log
+    levels) with `--log-level`. For example, we recommend running `--log-level
+    debug` to quickly debug things like connection issues and unexpected run
+    results. You can configure the user-facing output only on the command line with
```
beechtom, line 18: I think we discussed moving towards encouraging users to check `bolt-debug.log` over running `--log-level debug`, since users wouldn't need to run a command again in debug mode.

guides/logging.txt
```diff
@@ -0,0 +1,24 @@
...
+    `--[no]-verbose`.
```
beechtom, line 19:
```suggestion
    verbose (you still with me?). You can configure logging output in your
    project configuration, and can configure console output from the command
    line.
```

As with the other guides, I think we should avoid documenting specific commands, config options, etc., as that's more documentation. It also gets a little weird showing command-line options, as they're not the same between *nix shells and PowerShell.

guides/logging.txt
```diff
@@ -0,0 +1,24 @@
+TOPIC
+    logging
+
+DESCRIPTION
+    Bolt logs messages both to the console and to log files. Messages can
+    either come from Bolt's internal 'outputter' class, which logs user-facing
+    messages like progress and results, or from the 'logger' class, which logs
```
beechtom, line 7:
```suggestion
    either come from Bolt's outputter, which logs user-facing
    messages like progress and results, or from the logger, which logs
```

I think mentioning classes is more technical detail than a user needs to understand logging/output.

guides/logging.txt
```diff
@@ -0,0 +1,24 @@
...
+    warnings, errors, and log-structured output to log files. Both of these message
+    streams are configurable.
+
+    By default, Bolt logs to the console at 'warn' level, writes a log file to
+    '<project>/bolt-debug.log' at 'debug' level, and runs with the outputter in
+    verbose mode unless you are running a plan, in which case it runs without
+    verbose (you still with me?). You can configure logging output in your
```
beechtom, line 14:
```suggestion
    By default, Bolt logs to the console at 'warn' level and writes a log file to
    '<project>/bolt-debug.log' at 'debug' level. It also runs in verbose mode
    verbose mode unless you are running a plan, in which case it runs without
    verbose mode. You can configure logging output in your
```

guides/logging.txt
lucywyman: Oh hm yeah....I think I'll just remove that line, I'm not sure what else to recommend (recommending using `--log-level error` to silence warnings seems....not great.

guides/logging.txt
lucywyman: Aww, I like the "you still with me?". Too casual or negative?

hestonhoffman: It's a bit risky. It's an attempt to be inclusive, but a reader could find it a bit condescending. Also, sometimes phrases like this read as an admission of guilt (ie. "We're sorry this is so complicated").

beechtom: I do like the casual-ness, but agree with Heston's latter point. To me it makes it seem like the sentence is harder to follow / the concept is more difficult than it really is.

hestonhoffman: I think we could take the exception for plans to a new sentence (and maybe explain why):

"By default, Bolt logs to the console at 'warn' level and writes a log file to
'<project>/bolt-debug.log' at 'debug' level. Unless you are running a plan, 
Bolt runs in verbose mode by default. Bolt Plans use special formatting
depending on a plan's output, and do not use verbose mode by default."

guides/logging.txt
lucywyman: I'm starting to wonder if all of this is really just documentation - this is mainly to address Dimitri's question of whether `verbose` is a log level or not, but...there's not much to say that's *not* explanatory besides "Verbose is different from logging".

lucywyman: Maybe we could have this *just* link to the docs? We don't do that with other guides, but this could be the first

hestonhoffman: After reading this, I do think the logging doc needs some updates. I don't think it hurts to have a guide though 🤷

guides/logging.txt
```diff
@@ -0,0 +1,18 @@
...
+    logging
+
+DESCRIPTION
+    Bolt prints messages both to the console and to log files. Messages can
+    either come from Bolt's 'outputter', which logs user-facing messages like
+    progress and results, or from the 'logger', which logs warnings, errors, and
+    log-structured output to log files. Both of these message streams are
+    configurable.
+
+    By default, Bolt logs to the console at 'warn' level, writes a log file to
```
beechtom, line 11:
```suggestion
    By default, Bolt logs to the console at 'warn' level and writes a log file to
```

hestonhoffman (approved):
Good to go after Tom's change 👍

beechtom (approved):

Issue #2260, opened 2020-10-14
Add 'apply' function to function reference


beechtom:
## Describe the Change You Would Like
The `apply` function should be listed on the [function reference page](https://puppet.com/docs/bolt/latest/plan_functions.html). We should either duplicate the information [here](https://puppet.com/docs/bolt/latest/applying_manifest_blocks.html#applying-manifest-blocks-from-a-puppet-plan) or link to that page from the function reference page.

lucywyman:
Closed by #2285

PR #2285, opened 2020-10-26
(GH-2260) Add 'apply' function to function reference

beechtom:
This adds the `apply` function to the function reference doc. It also
slightly reformats by removing most bolded subheaders and listing params
and options in tables, improving the readability of the page.

!no-release-note

### DIFF
--- a/documentation/templates/plan_functions.md.erb
+++ b/documentation/templates/plan_functions.md.erb
@@ -1,53 +1,58 @@
-# Bolt functions
+# Plan functions
+
+Plans can use functions that are built into Bolt and Puppet, or custom
+functions included in modules. This reference
+page includes a list of built-in Bolt functions. To see a list of built-in
+Puppet functions, see Puppet's [built-in function
+reference](https://puppet.com/docs/puppet/latest/function.html). To learn how
+to write custom Puppet functions, see [the Puppet documentation on writing
+functions](https://puppet.com/docs/puppet/latest/writing_custom_functions.html).
 
 <% for func in @functions %>
 ## `<%= func['name'] %>`
 
-<%= func['text'] %>
+<%= func['desc'] %>
 
-<% for sig in func['signatures'] %>
-<% if sig['text'] != func['text'] -%>
-### <%= sig['text'] %>
-
+<% for sig in func['signatures'] -%>
+<% unless func['signatures'].count == 1 -%>
+**<%= sig['desc'].lines.first.strip.chomp('.') %>**
+<%= sig['desc'].lines.drop(1).join %>
 <% end -%>
+
+
 ```
 <%= sig['signature'] %>
 ```
 
-<% if sig['returns'] -%>
+This function<%= func['signatures'].count == 1 ? '' : ' signature' %> returns
+an object with the type `<%= sig['return'] %>` and accepts the following
+parameter<%= sig['params'].count == 1 ? '' : 's' %>:
 
-**Returns**
+| Parameter | Type | Description |
+| --- | --- | --- |
+<% sig['params'].each do |name, data| -%>
+| `<%= name %>` | `<%= data['type'] %>` | <%= data['desc'] %> |
+<% end %>
 
-<% for ret in sig['returns'] -%>
-`<%= ret['types'].first %>` <%= ret['text'] %>
+<% if sig['options'].any? -%>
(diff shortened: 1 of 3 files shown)


rakelib/docs.rake
```diff
@@ -342,3 +420,8 @@ end
 def format_links(text)
   text.gsub(/{([^}]+)}/, '[`\1`](#\1)')
 end
+
+def make_signature(function_name, params, return_type)
```
beechtom, line 424: I've personally found signatures with types a little hard to read, so I wanted to see what it would look like to just show how the function would be called and have the parameter list show what types are expected.

beechtom:
Preview site currently shows these changes.

documentation/templates/plan_functions.md.erb
```diff
@@ -1,53 +1,55 @@
-# Bolt functions
+# Plan functions
+
+Plans can use functions that are built into Bolt and Puppet. This reference
```
lucywyman, line 3:
```suggestion
Plans can use functions that are built into Bolt and Puppet, or custom functions included in modules on the modulepath. This reference
```

documentation/templates/plan_functions.md.erb
```diff
@@ -1,53 +1,55 @@
...
+page includes a list of built-in Bolt functions. To see a list of built-in
+Puppet functions, see Puppet's [built-in function
+reference](https://puppet.com/docs/puppet/latest/function.html).
```
lucywyman, line 6:
```suggestion
reference](https://puppet.com/docs/puppet/latest/function.html). To learn how to write custom Puppet functions, see [this guide](https://puppet.com/docs/puppet/latest/writing_custom_functions.html).
```

rakelib/docs.rake
lucywyman: This takes away pretty meaningful information from the signature, but we also have the parameters and types immediately below the signature. I'd advocate that between the parameter table and examples we can just remove the signature altogether. If we don't, I think we should leave them as they are.

nicklewis: I think in many cases the _types_ aren't that useful, since they tend to be what you would expect, but the signature lets you quickly see the _order_ to pass the parameters. For instance, I personally can never remember if `file::write` is `file::write(filename, content)` or `file::write(content, filename)`, but I do know that both of those are strings.

documentation/templates/plan_functions.md.erb
```diff
@@ -1,53 +1,58 @@
-# Bolt functions
+# Plan functions
+
+Plans can use functions that are built into Bolt and Puppet, or custom
+functions included in modules. This reference
+page includes a list of built-in Bolt functions. To see a list of built-in
+Puppet functions, see Puppet's [built-in function
+reference](https://puppet.com/docs/puppet/latest/function.html). To learn how
+to write custom Puppet functions, see [this
+guide](https://puppet.com/docs/puppet/latest/writing_custom_functions.html).
```
hestonhoffman, line 9:
```suggestion
to write custom Puppet functions, see [the Puppet documentation on writing functions](https://puppet.com/docs/puppet/latest/writing_custom_functions.html).
```
For accessibility

rakelib/docs.rake
hestonhoffman: I don't think the current signatures (with the types) offer much value because they're so hard to parse. Compare the current `download_file` example:
```
download_file(String[1] $source, String[1] $destination, Boltlib::TargetSpec $targets, Optional[Hash[String[1], Any]] $options)
```
vs the updated version:
```
download_file($source, $destination, $targets, $description, $options) => ResultSet
```
The new version seems much more useful to me simply because it's easier to figure out what's going on. Granted, I don't use this stuff every day. If I did, it would probably be easier for me to parse that first example, but I think I'd still wonder why it's included if the relevant information for each parameter is in the table below. 

I do think those initial signatures have value though. I think a lot of more experienced users would only need to look at that first signature (and maybe a quick glance at the table) so know what they need to do. Generally in the docs, we try to follow a similar structure: basic syntax, reference info you need, examples.

hestonhoffman (approved):
These changes look really good to me. I think the doc is a lot easier to read.

rakelib/docs.rake
lucywyman: Ok, if other people find it useful we can leave it.

lucywyman (approved):

PR #2294, opened 2020-10-28
(GH-2057) Add file and line number to Bolt errors

### LINKED ISSUE #2057: Spike: Expose file and line number for errors originating during plan runs.
## Use Case
When debugging plan code authors need to know where in the plan an error originated. For plan authors seeing the file name and line number in the plan code is a quick way to locate the source of the error.

For example, if a bolt plan function (ie. `run_task`) fails, the user doesn't know where the run_task call that failed is.

Bolt functions should add the calling file name and line number to  error details when they fail.
These should be visible in the debug log created for the plan run but not always included in the error message.
We should not worry about stacks traces for nested plans at this point.
We should try to preserve the original error when fail_plan is called with an Error object.
This work should focus just on fail_plan and run_* functions.

lucywyman:
This gets the file and line number from the Puppet plan function and
passes it all the way through to the Result object, so that if an error
occurs we can include the file and line number it came from under the
`details` key of the result. This ensures that users can find where a
particular failure is coming from not just from Puppet errors (ie. the
wrong number of arguments to a function) but also to Bolt errors (ie.  a
task that exits 1). This gets the file and line number simply using
`Puppet::Pops::PuppetStack.top_of_stack`. This only modifies the Result
objects for Results that build an error hash: `run_task`, `run_command`,
and `run_script`, as well as failures from `upload_file` and
`download_file`. This does not currently work for YAML plans.

Closes #2057

!feature

* **File and line number included in plan function errors** ([2057](https://github.com/puppetlabs/bolt/issues/2057))

  If the plan functions `run_command`, `run_script`, or `run_task` fail
  they will now include the file and line number in the `details` key of
  the Result object. This information will also be printed when run with
  info level logging or higher.

### DIFF
--- a/documentation/writing_tasks.md
+++ b/documentation/writing_tasks.md
@@ -421,6 +421,10 @@ end
 puts result.to_json
 ```
 
+Error results typically contain the file and line number under the `details` key of an
+error. Bolt merges the 'file' and 'line' keys with results unless 'file' is present in the
+`_error` hash returned from a task.
+
 ### Returning sensitive data
 
 To return secrets from a task, use the `_sensitive` key in the output. Bolt
--- a/lib/bolt/error.rb
+++ b/lib/bolt/error.rb
@@ -24,6 +24,10 @@ module Bolt
       h
     end
 
+    def add_filelineno(details)
+      @details.merge!(details) unless @details['file']
+    end
+
     def to_json(opts = nil)
       to_h.to_json(opts)
     end
--- a/lib/bolt/result.rb
+++ b/lib/bolt/result.rb
@@ -7,47 +7,58 @@ module Bolt
   class Result
     attr_reader :target, :value, :action, :object
 
-    def self.from_exception(target, exception, action: 'action')
+    def self.from_exception(target, exception, action: 'action', position: [])
+      details = create_details(position)
       if exception.is_a?(Bolt::Error)
-        error = exception.to_h
+        error = Bolt::Util.deep_merge({ 'details' => details }, exception.to_h)
       else
+        details['class'] = exception.class.to_s
         error = {
           'kind' => 'puppetlabs.tasks/exception-error',
           'issue_code' => 'EXCEPTION',
           'msg' => exception.message,
-          'details' => { 'class' => exception.class.to_s }
+          'details' => details
         }
         error['details']['stack_trace'] = exception.backtrace.join('\n') if exception.backtrace
       end
       Result.new(target, error: error, action: action)
     end
 
-    def self.for_command(target, stdout, stderr, exit_code, action, command)
+    def self.create_details(position)
+      %w[file line].zip(position).to_h.compact
+    end
+
+    def self.for_command(target, stdout, stderr, exit_code, action, command, position)
       value = {
         'stdout' => stdout,
         'stderr' => stderr,
         'exit_code' => exit_code
       }
+
+      details = create_details(position)
       unless exit_code == 0
+        details['exit_code'] = exit_code
         value['_error'] = {
           'kind' => 'puppetlabs.tasks/command-error',
           'issue_code' => 'COMMAND_ERROR',
           'msg' => "The command failed with exit code #{exit_code}",
-          'details' => { 'exit_code' => exit_code }
+          'details' => details
         }
       end
       new(target, value: value, action: action, object: command)
     end
 
-    def self.for_task(target, stdout, stderr, exit_code, task)
+    def self.for_task(target, stdout, stderr, exit_code, task, position)
       stdout.force_encoding('utf-8') unless stdout.encoding == Encoding::UTF_8
+
+      details = create_details(position)
       value = if stdout.valid_encoding?
                 parse_hash(stdout) || { '_output' => stdout }
               else
                 { '_error' => { 'kind' => 'puppetlabs.tasks/task-error',
                                 'issue_code' => 'TASK_ERROR',
                                 'msg' => 'The task result contained invalid UTF-8 on stdout',
-                                'details' => {} } }
+                                'details' => details } }
               end
 
       if exit_code != 0 && value['_error'].nil?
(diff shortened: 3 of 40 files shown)


lucywyman:
Recommended testing:
- Verify this works with the orch transport as expected
- Verify this works with the remote transport as expected
- Verify this works when using powershell, mostly as the target shell but as the controller shell would also be good
- Verify this prints errors from run_task_with
- Verify this works with the docker transport
- Verify this works with the local transport

lucywyman:
Woops - had those files in a separate commit :P This might break some tests but works for me now

lib/bolt/result.rb
```diff
@@ -7,47 +7,61 @@ module Bolt
   class Result
     attr_reader :target, :value, :action, :object
 
-    def self.from_exception(target, exception, action: 'action')
+    def self.from_exception(target, exception, action: 'action', position: [])
+      details = create_details(position)
       if exception.is_a?(Bolt::Error)
-        error = exception.to_h
+        error = { 'details' => details }.merge(exception.to_h)
```
beechtom, line 13: Should this be a deep merge? This will wipe out any existing details that the error might have already had.

lib/bolt/transport/orch.rb
```diff
@@ -63,25 +63,33 @@ def process_run_results(targets, results, task_name)
...
           if state == 'finished' || (result && result['_error'])
+            if result['_error']
+              file_lineno = position.any? ? { 'file' => position[0], 'line' => position[1] } : {}
+              result['_error']['details'].merge!(file_lineno) unless result['_error']['details']['file']
+            end
+
             Bolt::Result.new(target, value: result, action: 'task', object: task_name)
           elsif state == 'skipped'
+            details = {}
+            details['file'] = position[0] if position.any?
```
beechtom, line 74:
```suggestion
            details['file'] = position[0] if position[0]
```

lib/bolt/error.rb
```diff
@@ -24,6 +24,10 @@ def to_h
       h
     end
 
+    def add_filelineno(details)
+      @details.merge!(details) unless @details['file']
+    end
```
beechtom, line 29: It might be helpful to document that plans will add `file` and `line` to `details` unless `file` is already defined.

lib/bolt/result.rb
```diff
@@ -7,47 +7,61 @@ module Bolt
...
       end
       Result.new(target, error: error, action: action)
     end
 
-    def self.for_command(target, stdout, stderr, exit_code, action, command)
+    def self.create_details(position)
+      details = {}
+      details['file'] = position[0] if position.any?
+      details['line'] = position[1] if position[1]
+      details
```
beechtom, line 31:
```suggestion
      details = {}
      details['file'] = position[0] if position[0]
      details['line'] = position[1] if position[1]
      details
```

It looks like the stack will either always have [both file and line](https://github.com/puppetlabs/puppet/blob/eadd5474c26e7d28d90de00b9d7a7545ac10e55d/lib/puppet/pops/puppet_stack.rb#L33-L34) or return an [empty array](https://github.com/puppetlabs/puppet/blob/eadd5474c26e7d28d90de00b9d7a7545ac10e55d/lib/puppet/pops/puppet_stack.rb#L57), so I'm not sure there needs to be a check that both values exist.

Could even do a fun little one-liner: `%w[file line].zip(position).to_h.compact` :P

beechtom (commented):
It looks like `NoImplementationError` for missing task implementations don't get the file/line added to the error.

nicklewis:
Would it perhaps be cleaner to add the `top_of_stack` call to the `Result` class rather than passing it through _everywhere_? That's kind of magic, but it also seems like the stack location is a concern that the transports shouldn't have to be aware of in the first place.

lucywyman:
@nicklewis I'm mostly nervous about how that will interact with parallel blocks - each block will have it's own scope so that might just work, but this just seemed more certain.

lib/bolt/error.rb
lucywyman: I dunno - to me it feels like expected behavior and kind of clutters the docs, but I added it for consideration.

lib/bolt/result.rb
lucywyman: Yike yes good catch!

lucywyman:
@beechtom Can you post the plan/task that doesn't have the file and line number? For missing implementations I get:
```
lucy ~/githubs/bolt (GH-2057) $ be bolt plan run bolt::error
Starting: plan bolt::error
Finished: plan bolt::error in 0.01 sec
{
  "kind": "bolt/pal-error",
  "msg": "Task metadata for task bolt::err specifies missing implementation err.ps1",
  "details": {
    "file": "/home/lucy/githubs/bolt/plans/error.pp",
    "line": 4,
    "column": 3
  }
}
```

nicklewis:
Hmm, the stack is a thread-local variable. That means as far as multiple fibers go it should work fine. But it also means that since the transport operations happen on a threadpool, it wouldn't work at all.

Perhaps we could make the `run_*` functions annotate the ResultSet they get back with this stack information? It just seems really backward to take information that is known at one point, pass it all the way down, only for it to be returned all the way back up.

beechtom:
@lucywyman Ah, it actually is showing in json output. I think it's just because that's from an exception whereas fail results will print details to stderr.

lucywyman:
AH! I remember now why we had to thread it through everything - we log the result in the executor at info level, and in order for the file and line to be in that message (which I think they should be) we need to add them when the error is created, not afterwards. https://github.com/puppetlabs/bolt/blob/main/lib/bolt/executor.rb#L252

beechtom (approved):

documentation/writing_tasks.md
```diff
@@ -421,6 +421,10 @@ end
 puts result.to_json
 ```
 
+Error results typically contain the file and line number under the `details` key of an
+error. The 'file' and 'line' keys will be merged with results unless 'file' is present in the
+`_error` hash returned from a task.
```
hestonhoffman, line 426:
```suggestion
Error results typically contain the file and line number under the `details` key of an
error. Bolt merges the 'file' and 'line' keys with results unless 'file' is present in the
`_error` hash returned from a task.
```

hestonhoffman (approved):
One small change, but docs LGTM

Issue #2171, opened 2020-09-14
Receive "unable to get local issuer certificate" error when attempting to run bolt puppetfile install


rychannel:
## Describe the Bug

When I run bolt puppetfile install, I get the following error:

SSL_connect returned=1 errno=0 state=error: certificate verify failed (unable to get local issuer certificate)

## Expected Behavior
I expect the modules to actually install

## Steps to Reproduce

1. Run bolt puppetfile install from Windows

## Environment
 - Version 2.27.0
 - Platform: Windows 10

jpogran:
Hi @rychannel thanks for the report. Are you using a proxy? Can you paste output from an example run?

rychannel:
I found a support article on the Puppet Support site that said how to fix it.

jpogran:
That is great! Does that mean I can close this issue, or is there anything else we can help with?

rychannel:
Yep

jpogran:
The crux of the issues lies in these lines: https://github.com/puppetlabs/bolt/blob/b097bd10f94016a2115fb82e1bd6613ac9aee5b3/pwsh_module/pwsh_bolt_internal.ps1#L23-L25

We set these inside the bolt function, but instruct users that they can override it by setting it directly in the shell like this:

```
❯ $env:SSL_CERT_FILE = 'C:\path\to\ssl\cert.pem'
❯ bolt puppetfile install
```

This will never work, because the function always sets those variables. The long term fix is to have the function first check if those variables are set, and if set do not assign the default value to them.

The short term workaround is to have the user copy the following function and add it to their `$PROFILE`, which will override the function declared in the PuppetBolt module, and then delete these lines when the fix lands

```
function bolt {
  $fso = New-Object -ComObject Scripting.FileSystemObject

  $script:BOLT_BASEDIR = (Get-ItemProperty -Path "HKLM:\Software\Puppet Labs\Bolt").RememberedInstallDir
  # Windows API GetShortPathName requires inline C#, so use COM instead
  $script:BOLT_BASEDIR = $fso.GetFolder($script:BOLT_BASEDIR).ShortPath
  $script:RUBY_DIR = $script:BOLT_BASEDIR
  # Set SSL variables to ensure trusted locations are used
   if(!$env:SSL_CERT_FILE){
    $env:SSL_CERT_FILE = "$($script:BOLT_BASEDIR)\ssl\cert.pem"
  }
  if(!$env:SSL_CERT_DIR){
[...]

PR #2297, opened 2020-10-28
(GH-2171) Do not override SSL variables in PowerShell module

beechtom:
This updates the PowerShell module to not override the `SSL_CERT_FILE`
and `SSL_CERT_DIR` environment variables if they are already set.
Previously, the PowerShell module would always override these
environment variables, which could lead to SSL errors if a user
attempted to provide a cert on the command line.
are not already set.

!bug

* **Do not override SSL variables in PowerShell module**
  ([#2171](https://github.com/puppetlabs/bolt/issues/2171))

  The PowerShell module no longer overrides the `SSL_CERT_FILE` and
  `SSL_CERT_DIR` environment variables if they are already set.

### DIFF
--- a/documentation/troubleshooting.md
+++ b/documentation/troubleshooting.md
@@ -124,6 +124,27 @@ Use the Bolt-provided gem command to reinstall/install these gems. For example:
     sudo /opt/puppetlabs/bolt/bin/gem pristine unf_ext --version 0.0.7.5
 ```
 
+
+## Certificate verify failed when installing modules
+
+When running on Windows, Bolt automatically sets the `SSL_CERT_DIR` and
+`SSL_CERT_FILE` environment variables if they are not already set. Assuming a
+default install location, the variables are set to the following directory and
+certificate, which are installed with the Bolt package:
+
+- `SSL_CERT_DIR = C:\Program Files\Puppet Labs\Bolt\ssl\certs`
+- `SSL_CERT_FILE = C:\Program Files\Puppet Labs\Bolt\ssl\cert.pem`
+
+If you see an SSL connection error similar to the following when running on
+Windows:
+
+```
+SSL_connect returned=1 errno=0 state=error: certificate verify failed (unable to get local issuer certificate)
+```
+
+Set the `SSL_CERT_DIR` and `SSL_CERT_FILE` environment variables to use a valid
+certificate and certificate directory.
+
 ## I still need help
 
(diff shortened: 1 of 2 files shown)


lucywyman:
I think this might need a rebase to not be red.

lucywyman (approved):
@jpogran I'd love to get your review before merging!

pwsh_module/pwsh_bolt_internal.ps1
```diff
@@ -21,8 +26,12 @@ function Invoke-BoltCommandline {
   $script:BOLT_BASEDIR = $fso.GetFolder($script:BOLT_BASEDIR).ShortPath
   $script:RUBY_DIR = $script:BOLT_BASEDIR
   # Set SSL variables to ensure trusted locations are used
```
jpogran, line 28: This comment doesn't make sense anymore with these changes

pwsh_module/pwsh_bolt_internal.ps1
```diff
@@ -6,8 +6,13 @@ function bolt {
...
```
jpogran, line 8: This comment doesn't make sense anymore with these changes

jpogran (approved):
I think we should add a short example to the documentation showing that you can now set these, as there isn't a parameter or help text to show you this is possible. If that should be a separate PR, that's fine by me. Otherwise 👍

beechtom:
Added a short bit about setting the env vars to the Troubleshooting page.

documentation/troubleshooting.md
```diff
@@ -124,6 +124,27 @@ Use the Bolt-provided gem command to reinstall/install these gems. For example:
...
+
+When running on Windows, Bolt automatically sets the `SSL_CERT_DIR` and
+`SSL_CERT_FILE` environment variables if they are not already set. Assuming a
+default install location, the variables are set to the following directory and
+certificate, which are installed with the Bolt package:
+
+- `SSL_CERT_DIR = C:\Program Files\Puppet Labs\Bolt\ssl\certs`
+- `SSL_CERT_FILE = C:\Program Files\Puppet Labs\Bolt\ssl\cert.pem`
+
+If you see a SSL connection error similar to the following when running on
```
hestonhoffman, line 138:
```suggestion
If you see an SSL connection error similar to the following when running on
```

hestonhoffman (approved):
One small change, but looks good. Glad we're adding to the troubleshooting page!

Issue #2272, opened 2020-10-21
Invoke-BoltCommand fails to run PowerShell command with mixed quotes


AUsatenko:
## Describe the Bug
when I execute
```Invoke-BoltCommand -Command  "Get-WMIObject Win32_Service -Filter ""Name like '%mon'"" | SELECT NAME,STARTNAME"  -Targets server```

Get an error: Unknown argument(s) like, %mon | SELECT NAME,STARTNAME'

## Expected Behavior
should return name and startname of the service on remote server

## Steps to Reproduce
1. Run powershell cmdlet in bug description

## Environment
 - Version 2.31
 - Platform Windows 10

**Comment by jpogran** [comment]
The reason this is occurring is that there are 3 levels of argument parsing happening: the parent PowerShell process invoking `Invoke-BoltCommand`, the bolt ruby argument parser, and the final PowerShell process on the target.

The `Invoke-BoltCommand` does some parsing to build a command line for the ruby bolt, which results in the following: 

```
❯ Invoke-BoltCommand -Command "Get-WMIObject Win32_Service -Filter ""Name like '%mon'""" -Targets localhost -Verbose
VERBOSE: Examining Command
VERBOSE: Parsing Command as default
VERBOSE: Examining Targets
VERBOSE: Parsing Targets as default
VERBOSE: Examining Verbose
VERBOSE: Parsing Verbose as switch parameter
VERBOSE: Executing C:\PROGRA~1\PUPPET~1\Bolt\bin\ruby -S -- C:\PROGRA~1\PUPPET~1\Bolt\bin\bolt command run 'Get-WMIObject Win32_Service -Filter "Name like '%mon'"' --targets 'localhost' --verbose
0 command
1 run
2 'Get-WMIObject Win32_Service -Filter Name
3 like
4 %mon'
5 --targets
6 localhost
7 --verbose
Unknown argument(s) like, %mon'
```

The problem is that the `command` to run now has single quotes around it, which invalidate the un-escaped single quotes inside, causing ruby ARGV to break apart the string into 2 more parts.

To compare:

```
❯ bolt command run "Get-WMIObject Win32_Service -Filter ""Name like '%rm'""" -t localhost
0 command
1 run
2 Get-WMIObject Win32_Service -Filter "Name like '%rm'"
3 -t
4 localhost
Started on localhost...
Finished on localhost:
  STDOUT:
    ExitCode  : 0
    Name      : WinRM
    ProcessId : 15724
    StartMode : Auto
[...]

PR #2300, opened 2020-10-29
(GH-2272) Fix complicated quoting in PowerShell Cmdlets

jpogran:
This commit changes the default argument parsing inside Invoke-BoltCommandline to not add single quotes to every argument. It also special cases the `params` and `execute` parameters to quote the input specifically.

### DIFF
--- a/documentation/running_bolt_commands.md
+++ b/documentation/running_bolt_commands.md
@@ -27,6 +27,31 @@ Invoke-BoltCommand -Command 'Get-Location' -Targets servers
 > 🔩 **Tip:** If a command contains spaces or special shell characters, wrap
 > the command in single quotation marks.
 
+### Run a quoted command
+
+If you need to run a command that uses quotation marks, you must properly escape
+the quotations. The way you escape the quotations depends on whether you're
+using Bash or PowerShell.
+
+In a Bash shell, use backslashes `\` or double the the quotation marks:
+
+_\*nix shell command_
+
+```shell
+bolt command run "Get-WMIObject Win32_Service -Filter ""Name like '%mon'""" -t localhost
+```
+
+In a PowerShell shell, use a combination of backslashes `\` and doubling of
+quotation marks. The example below uses two double quotation marks to quote the
+value being passed to `Filter`, however the example also uses a backslash so
+that Bolt's underlying Ruby argument parser accepts the command.
+
+_PowerShell cmdlet_
+
+```powershell
+Invoke-BoltCommand -Command "Get-WMIObject Win32_Service -Filter \""Name like '%mon'\""" -Targets localhost
+```
+
 ### Read a command from a file
 
 Reading a command from a file is useful when you need to run a script on a target 
--- a/pwsh_module/command.tests.ps1
+++ b/pwsh_module/command.tests.ps1
@@ -118,18 +118,22 @@ Describe "test all bolt command examples" {
   Context "bolt apply" {
     It "bolt apply manifest.pp -t target" {
       $result = Invoke-BoltApply -manifest 'manifest.pp' -target 'target'
-      $result | Should -Be "bolt apply 'manifest.pp' --targets 'target'"
+      $result | Should -Be "bolt apply manifest.pp --targets target"
     }
     It "bolt apply -e `"file { '/etc/puppetlabs': ensure => present }`" -t target" {
       $result = Invoke-BoltApply -execute "file { '/etc/puppetlabs': ensure => present }" -targets 'target'
-      $result | Should -Be "bolt apply --execute 'file { '/etc/puppetlabs': ensure => present }' --targets 'target'"
+      $result | Should -Be "bolt apply --execute 'file { '/etc/puppetlabs': ensure => present }' --targets target"
     }
   }
 
   Context "bolt command" {
     It "bolt command run 'uptime' -t target1,target2" {
       $result = Invoke-BoltCommand -command 'uptime' -target 'target1,target2'
-      $result | Should -Be "bolt command run 'uptime' --targets 'target1,target2'"
+      $result | Should -Be "bolt command run uptime --targets target1,target2"
+    }
+    It "bolt command run complicated quoting" {
+      $result = Invoke-BoltCommand -Command "Get-WMIObject Win32_Service -Filter \""Name like '%mon'\""" -Targets 'target1,target2'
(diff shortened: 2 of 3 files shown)


documentation/running_bolt_commands.md
```diff
@@ -27,6 +27,26 @@ Invoke-BoltCommand -Command 'Get-Location' -Targets servers
 > 🔩 **Tip:** If a command contains spaces or special shell characters, wrap
 > the command in single quotation marks.
 
+### Run a quoted command
+
+If you need to run a command with quotes inside, there are a few considerations to keep in mind. You will need properly escape the quotes according to the whether you are running in `bash` or in `powershell`.
```
hestonhoffman, line 32:
```suggestion
If you need to run a command that uses quotation marks, you must properly escape the quotations. The way you escape the quotations depends on whether you're using Bash or PowerShell.
```

documentation/running_bolt_commands.md
```diff
@@ -27,6 +27,26 @@ Invoke-BoltCommand -Command 'Get-Location' -Targets servers
...
+
+In a `bash` shell, you'll need to use backslashes `\` or double the quote you're using in order for the quotes to pass through.
```
hestonhoffman, line 34:
```suggestion
In a Bash shell, use backslashes `\` or double the the quotation marks:
```

documentation/running_bolt_commands.md
```diff
@@ -27,6 +27,26 @@ Invoke-BoltCommand -Command 'Get-Location' -Targets servers
...
+
+_\*nix shell command_
+
+```shell
+bolt command run "Get-WMIObject Win32_Service -Filter ""Name like '%mon'""" -t localhost
+```
+
+In a `powershell` shell, you'll need to use a combination of escape characters and doubling quotes in order for the quotes to pass through. In the example below, we use two double quotes to quote the value being passed to `Filter`, however we need to pass a backslash in order to have the ruby code below it accept the command.
```
hestonhoffman, line 42:
```suggestion
In a PowerShell shell, use a combination of backslashes `\` and doubling of quotation marks. The example below uses two double quotation marks to quote the value being passed to `Filter`, however the example also uses a backslash so that the Ruby code below the value accepts the command.
```

hestonhoffman: I'm not 100% on "The Ruby code below it". Are we talking about the underlying code that makes up Bolt here?

documentation/running_bolt_commands.md
hestonhoffman: Does the Bash stuff apply to all *nix style shells? Just wondering if we should sub Bash for *nix

documentation/running_bolt_commands.md
jpogran: The ruby argument parser will properly accept the quotes if they are escaped. This is an artifact of passing tokens through two levels of programs. 

Maybe: `so that the Ruby argument parser accepts the command`

jpogran:
pwsh github action failure is from the powershell gallery being [down](https://github.com/PowerShell/PowerShellGallery/blob/master/psgallery_status.md)

documentation/running_bolt_commands.md
hestonhoffman: Okay, I made a small edit. How does that look?

hestonhoffman (approved):
LGTM!

lucywyman:
@jpogran Would you mind squashing the commits, and updating the release note to something like:
```
!bug

* **Fix complicated quoting in PowerShell Cmdlets** ([#2272](#2272))
    When using the Powershell cmdlets module, Bolt no longer wraps each command in single quotes allowing users to successfully use more complicated quoting patterns.
```

documentation/running_bolt_commands.md
```diff
@@ -27,6 +27,26 @@ Invoke-BoltCommand -Command 'Get-Location' -Targets servers
...
+```shell
+bolt command run "Get-WMIObject Win32_Service -Filter ""Name like '%mon'""" -t localhost
+```
+
+In a PowerShell shell, use a combination of backslashes `\` and doubling of quotation marks. The example below uses two double quotation marks to quote the value being passed to `Filter`, however the example also uses a backslash so that Bolt's underlying Ruby argument parser accepts the command.
+
+_PowerShell cmdlet_
+
+```powershell
+Invoke-BoltCommand -Command "Get-WMIObject Win32_Service -Filter \""Name like '%mon'\"""
```
beechtom, line 47:
```suggestion
Invoke-BoltCommand -Command "Get-WMIObject Win32_Service -Filter \""Name like '%mon'\""" -Targets localhost
```

documentation/running_bolt_commands.md
```diff
@@ -27,6 +27,26 @@ Invoke-BoltCommand -Command 'Get-Location' -Targets servers
...
+
+In a Bash shell, use backslashes `\` or double the the quotation marks:
+
+_\*nix shell command_
+
+```shell
+bolt command run "Get-WMIObject Win32_Service -Filter ""Name like '%mon'""" -t localhost
+```
+
+In a PowerShell shell, use a combination of backslashes `\` and doubling of quotation marks. The example below uses two double quotation marks to quote the value being passed to `Filter`, however the example also uses a backslash so that Bolt's underlying Ruby argument parser accepts the command.
```
beechtom, line 42:
```suggestion
In a PowerShell shell, use a combination of backslashes `\` and doubling of
quotation marks. The example below uses two double quotation marks to quote the
value being passed to `Filter`, however the example also uses a backslash so
that Bolt's underlying Ruby argument parser accepts the command.
```

documentation/running_bolt_commands.md
```diff
@@ -27,6 +27,26 @@ Invoke-BoltCommand -Command 'Get-Location' -Targets servers
 > 🔩 **Tip:** If a command contains spaces or special shell characters, wrap
 > the command in single quotation marks.
 
+### Run a quoted command
+
+If you need to run a command that uses quotation marks, you must properly escape the quotations. The way you escape the quotations depends on whether you're using Bash or PowerShell.
```
beechtom, line 32:
```suggestion
If you need to run a command that uses quotation marks, you must properly escape
the quotations. The way you escape the quotations depends on whether you're
using Bash or PowerShell.
```

beechtom (approved):

Issue #2278, opened 2020-10-22
Better Line numbers for evaluated code


adreyer:
## Use Case

When puppet code snippets fail in a yaml plan the location of the error should be in relation to the larger yaml plan rather than the snippet itself. What is the best way to make it clear to users what the location is in the YAML plan.

PR #2306, opened 2020-10-29
(GH-2278) Include file and line number in YAML plan eval errors

beechtom:
This adds the file path and line number of an error for evaluated code
in a YAML plan. Previously, if evaluated code resulted in an error, the
error would include no file path and the line number of the error within
the code itself (as though the code were in its own file). This made
these errors less helpful, as they did not correctly indicate where an
error in a YAML plan had occurred.

!feature

* **Include file and line number in YAML plan code evaluation errors**
  ([#2278](https://github.com/puppetlabs/bolt/issues/2278))

  Errors raised when evaluating code in a YAML plan now include the path
  to the YAML plan and the line number that the error occurred on in the
  plan.

!removal

* **Folded scalar values in YAML plans no longer evaluated**

  Folded scalar values in YAML plans are no longer evaluated and are
  instead treated as string literals.

### DIFF
--- a/lib/bolt/pal/yaml_plan.rb
+++ b/lib/bolt/pal/yaml_plan.rb
@@ -98,10 +98,12 @@ module Bolt
       # subclasses this parent class in order to implement its own evaluation
       # logic.
       class EvaluableString
-        attr_reader :value
+        attr_reader :file, :line, :value
 
-        def initialize(value)
+        def initialize(value, file = nil, line = nil)
           @value = value
+          @file  = file
+          @line  = line
         end
 
         def ==(other)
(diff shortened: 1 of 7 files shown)


lucywyman:
If we add a deprecation to the changelog do we also plan to add a deprecation warning about it when users use it?

beechtom:
@lucywyman We could, but it would also result in a warning whenever you use a folded scalar. Given the number of YAML plan users we have, and the likelihood that any of those users is using folded scalars (low), I feel that a deprecation warning isn't necessary and documenting the change in the changelog will suffice.

lucywyman (approved):

lucywyman:
I mean, why *not* add a deprecation warning even for those users? I assume we're going to wait until 3.0 and not just remove it.

beechtom:
We're not removing support for folded scalars, we're just not evaluating them as Puppet code anymore. There's also no removing anything with 3.0, as this _is_ the permanent change.

nicklewis:
To be strict then, maybe this _is_ a removal rather than a deprecation? We don't have a good way to flag straight up breaking changes which is what this _really_ is.

beechtom:
Updated the commit to say removal.

PR #2308, opened 2020-10-30
(docs) Document using target name in Hiera configuration

lucywyman:
This supersedes #2136

"This PR adds some additional points to the docs on using hiera with bolt
apply, specifically how to use the target name as part of a hiera path."

!no-release-note


hestonhoffman (approved):
LGTM!

PR #2310, opened 2020-10-31
(maint) Release the pal_mutex after fetching the file

ekinanp:
This way, the project_file_metadatas endpoint doesn't block other pal
operations (useful for large directories where fetching the file
metadata could take some time).

### DIFF
--- a/lib/bolt_server/transport_app.rb
+++ b/lib/bolt_server/transport_app.rb
@@ -357,23 +357,6 @@ module BoltServer
       plans.map { |plan_name| { 'name' => plan_name } }
     end
 
-    def file_metadatas(pal, module_name, file)
-      pal.in_bolt_compiler do
-        mod = Puppet.lookup(:current_environment).module(module_name)
-        raise ArgumentError, "`module_name`: #{module_name} does not exist" unless mod
-        abs_file_path = mod.file(file)
-        raise ArgumentError, "`file`: #{file} does not exist inside the module's 'files' directory" unless abs_file_path
-        fileset = Puppet::FileServing::Fileset.new(abs_file_path, 'recurse' => 'yes')
-        Puppet::FileServing::Fileset.merge(fileset).collect do |relative_file_path, base_path|
-          metadata = Puppet::FileServing::Metadata.new(base_path, relative_path: relative_file_path)
-          metadata.checksum_type = 'sha256'
-          metadata.links = 'follow'
-          metadata.collect
-          metadata.to_data_hash
-        end
-      end
-    end
-
     get '/' do
       200
     end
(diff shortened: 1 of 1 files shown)


McdonaldSeanp (approved):

lib/bolt_server/transport_app.rb
```diff
@@ -621,11 +604,40 @@ def make_winrm_target(target_hash)
...
+      # would clean-up some of this code.
+      error_response = in_bolt_project(project_ref) do |context|
+        abs_file_path = context[:pal].in_bolt_compiler do
+          mod = Puppet.lookup(:current_environment).module(module_name)
+          raise ArgumentError, "`module_name`: #{module_name} does not exist" unless mod
+          mod.file(file)
+        end
+        nil
       end
+      next error_response if error_response
```
donoghuc, line 624: Why is this `next` and not `return`? We are not in iteration at this point right?

donoghuc: Also, what would the exit code be for that? maybe we should just `raise` here if we get the error respones.

ekinanp: `next` is the idiomatic way of returning from a block, like `return` is for a function.

The `error_response` bit is b/c of how `in_bolt_project` is structured. Specifically, its current form is

(1) Do some validation. If validation fails, then return an error response (e.g. like for a nonexistent project ref).

(2) Execute the block, the block should also return a response. However here, I only need the lock to fetch the file path so I don't need a full response object. It would also be confusing if I had something like

```
abs_file_path = in_bolt_project <..>
```

b/c then `abs_file_path` is either an error response (like one for nonexistent project ref), or the file path itself.

Ideally, `in_bolt_project` and similar methods should be refactored so that they don't set the `pal_mutex` (letting the caller do it instead) and _maybe_ not expect a response to be returned, but I don't want to do that refactor in this PR.

donoghuc: What i dont understand is how ruby knows that the next in this case is for the block we use to assign to the variable `error_response`. I also, dont understand what happens if we hit that `raise` in the block? it seems like we could just do
```ruby
    in_bolt_project(project_ref) do |context|
        abs_file_path = context[:pal].in_bolt_compiler do
          mod = Puppet.lookup(:current_environment).module(module_name)
          raise ArgumentError, "`module_name`: #{module_name} does not exist" unless mod
          mod.file(file)
        end
    end
```

and let that raise be handled just like any other exception (see https://github.com/puppetlabs/bolt/pull/2310/files#diff-9d9a096c76154ac144a49d1ae6a2f5d317ef1b644a1d444bc3eb76c85e81d12bR629 )

ekinanp: Yeah, that won't work because `in_bolt_project` doesn't raise an error for e.g. nonexistent project-ref. It returns the error response -- see https://github.com/puppetlabs/bolt/blob/main/lib/bolt_server/transport_app.rb#L265 -- which is an array.

I also didn't want to change that to raise `ArgumentError` instead since I just wanted this to be a quick PR. But I do agree that it's current form is pretty confusing and that confusion's caused by a non-intuitive implementation of `in_bolt_project`.

I think I'll just file a ticket for this PR ("Release the pal_mutex in project_file_metadatas after fetching the file") and look at it again _after_ some of the inventory parsing work's been finished (since part of that work requires a slight refactor of `in_bolt_project` anyways).

donoghuc: Absolutely. Thanks @ekinanp

ekinanp: Filed https://tickets.puppetlabs.com/browse/PE-30449.

ekinanp:
Closing this PR in favor of https://tickets.puppetlabs.com/browse/PE-30449, which will probably come after some slight bolt-server refactoring caused by the upcoming inventory parsing work.

PR #2312, opened 2020-11-03
Fixes #2289

### LINKED ISSUE #2289: SSH Config breaks bolt despite native-ssh transport
## Describe the Bug

A valid ssh client config breaks bolt's ssh transport in native-ssh mode, even with `load-config: false`.

## Expected Behavior

Bolt running ssh as is and leaving the client config alone

## Steps to Reproduce

1. Have an SSH config with a `Match host canonical all` block
2. Set up bolt with this config:

```yaml
config:
  ssh:
    ssh-command: 'ssh'
    native-ssh: true
    load-config: false
    user: root
    host-key-check: false
```

3. Run the `facts` plan on all hosts:

```
$ bolt plan run facts targets=all
Starting: plan facts
Starting: task facts on hosta
Finished: task facts with 1 failure in 0.06 sec
Finished: plan facts in 0.07 sec
Failed on shibboleth:
  all cannot be mixed with other conditions
Failed on 1 target: hosta
Ran on 1 target
```

## Environment

 - Bolt version 2.32.0
 - OpenSSH_8.4p1, OpenSSL 1.1.1h  22 Sep 2020
 - Arch Linux

fuero:
Fixes #2289 by disabling loading the ssh config.

### DIFF
--- a/lib/bolt/transport/ssh/exec_connection.rb
+++ b/lib/bolt/transport/ssh/exec_connection.rb
@@ -12,8 +12,12 @@ module Bolt
           raise Bolt::ValidationError, "Target #{target.safe_name} does not have a host" unless target.host
 
           @target = target
-          ssh_config = Net::SSH::Config.for(target.host)
-          @user = @target.user || ssh_config[:user] || Etc.getlogin
+          begin
+            ssh_config = Net::SSH::Config.for(target.host)
+            @user = @target.user || ssh_config[:user] || Etc.getlogin
+          rescue StandardError
+            @user = @target.user || Etc.getlogin
+          end
           @logger = Bolt::Logger.logger(self)
         end
 


lib/bolt/transport/ssh/exec_connection.rb
```diff
@@ -12,8 +12,12 @@ def initialize(target)
           raise Bolt::ValidationError, "Target #{target.safe_name} does not have a host" unless target.host
 
           @target = target
-          ssh_config = Net::SSH::Config.for(target.host)
-          @user = @target.user || ssh_config[:user] || Etc.getlogin
+          unless target.transport_config['native-ssh']
```
lucywyman, line 15: We only ever use this transport if `native-ssh` is true for a target, so we should be able to skip the `if` and just use `@user = @target.user || Etc.getlogin` directly.

lucywyman (commented):
Dang great find! Thanks so much for fixing this.

The only other thing to do before merging is add a release note to the commit. You should be able to do `git commit --amend` and then add:
```
!bug

* **Don't load SSH Config through net-ssh when using native-ssh** ([#2289 ](https://github.com/puppetlabs/bolt/issues/2289)
  We now skip loading SSH config through the net-ssh gem when using the native-ssh transport to avoid limitations imposed by the net-ssh gem, which isn't used for the native-ssh transport.
```

* **Descriptive title of changes** ([#issue_number](issue_url))

  Descriptive summary of changes.

fuero:
Ok, done.

I tried signing the CLA, but it seems that something is broken, as it keeps redirecting me:

https://cla.puppet.com/?error=redirect_uri_mismatch&error_description=The+redirect_uri+MUST+match+the+registered+callback+URL+for+this+application.&error_uri=https%3A%2F%2Fdocs.github.com%2Fapps%2Fmanaging-oauth-apps%2Ftroubleshooting-authorization-request-errors%2F%23redirect-uri-mismatch&state=7af90619b4dda44d29e6f0f471bbb5ccdbb9266646b59f84

lib/bolt/transport/ssh/exec_connection.rb
```diff
@@ -12,8 +12,7 @@ def initialize(target)
           raise Bolt::ValidationError, "Target #{target.safe_name} does not have a host" unless target.host
 
           @target = target
-          ssh_config = Net::SSH::Config.for(target.host)
-          @user = @target.user || ssh_config[:user] || Etc.getlogin
+          @user = @target.user || Etc.getlogin
```
nicklewis, line 15: I _think_ we want to omit `Etc.getlogin` here too and just rely on `ssh` doing the right thing. Otherwise this will _never_ respect the username set in SSH config and will always use the login user.

I'm not sure whether there's some reason I'm not thinking of why it's better to set the user to the login user explicitly...

nicklewis: Ah this conversation has some context: https://github.com/puppetlabs/bolt/pull/1806#discussion_r423246069

So we _do_ need to know the SSH user for sure and can't just rely on `ssh` figuring it out.

lucywyman: !!! I think we might be able to use `sshd -T` instead of `ssh -G` to do exactly that! It was introduced in OpenSSH 5.1, and I'm double checking our supported OS's all have at least that version but I'd be surprised if they didn't.

nicklewis: It looks like that checks the sshd config file, which is different than the ssh config file.

fuero: Sorry if that's a dumb question, but can't you just use $USER on the host ssh connects to for the things referenced in #1806? Let the destination shell evaluate it?

lucywyman: @fuero There are some cases where that won't work, like some network devices and maybe powershell environments (which we have a feature for SSHing into, though it's experimental). @nicklewis had a great idea which was to try loading the config through Net::SSH and if it fails fall back to using `$USER`. Does that seem reasonable? I know it's a bit more work than what's here - let me know if you want to:
1. Make the changes yourself, 
2. Have me push the changes to your branch (I might need permission)
3. Have me put up a new PR with that behavior

lucywyman: @fuero No rush, just wanted to check in and make sure this doesn't slip through the cracks. We'd like to get it fixed!

fuero: @lucywyman Sorry, was busy at work. Here's another take, it works for my use case. Have a look!

lib/bolt/transport/ssh/exec_connection.rb
```diff
@@ -12,8 +12,12 @@ def initialize(target)
           raise Bolt::ValidationError, "Target #{target.safe_name} does not have a host" unless target.host
 
           @target = target
-          ssh_config = Net::SSH::Config.for(target.host)
-          @user = @target.user || ssh_config[:user] || Etc.getlogin
+          begin
+            ssh_config = Net::SSH::Config.for(target.host)
+            @user = @target.user || ssh_config[:user] || Etc.getlogin
+          rescue => e
```
lucywyman, line 18:
```suggestion
          rescue StandardError
```
This will make rubocop happy (specifying an error class, and no assigning to a var without using it)

lib/bolt/transport/ssh/exec_connection.rb
lucywyman: Perfect! One small syntax comment, then this looks good to me!

lib/bolt/transport/ssh/exec_connection.rb
lucywyman: Oh sorry no `=>`, just the `rescue StandardError`

Issue #2314, opened 2020-11-03
The bolt module add command fails when a forge module with an uppercase has been added


martezr:
## Describe the Bug
The new module management in Bolt has issues when dealing with Puppet Forge modules that include an uppercase letter in the username part of the module slug. This prevents modules from being added to the Bolt project if a module that was created by a Forge user with an upper case character in their name.

In an attempt to understand how wide reaching the implications of this bug are I queried the Puppet forge for such modules. There are roughly 286 modules on the Puppet Forge with an uppercase letter in the username of the module's author. The following python script was used to fetch the module data.

```
import requests

moduleinit = 0
badmodules = []
while moduleinit != 6700:
    url = "https://forgeapi.puppet.com/v3/modules"
    PARAMS = {'limit':100, 'offset': moduleinit}
    r = requests.get(url, params = PARAMS)
    data = r.json()
    for module in data['results']:
        if module['owner']['slug'].islower():
            pass
        else:
            print(module['slug'])
            badmodules.append(module['slug'])
    moduleinit += 100

print(len(badmodules))
```

## Expected Behavior
The module should be installed and added to the Puppetfile as it would if the invalid module wasn't added to the Puppetfile.

## Steps to Reproduce
1. Create an empty directory for the new bolt project 
```
mkdir modulebug
```
2. Initialize the bolt project
```
bolt project init
```
3. Migrate the project to support adding modules
```
bolt project migrate
```
4. Add a module that has been created by a user with an uppercase letter in their username
```
bolt module add kyleanderson-consul
```
5. Attempt to add another module to the project
```
bolt module add puppetlabs-vcsrepo
```

The following error message is thrown when attempting to install the vcsrepo module.

```
Adding module puppetlabs-vcsrepo to project

Puppetfile does not include modules that satisfy the following specifications:

- name: kyleanderson/consul

This may not be a Puppetfile managed by Bolt. To forcibly overwrite the
Puppetfile, run 'bolt module install --force'.
```

The error message is being thrown because the entry in the Puppetfile doesn't match due to the case mismatch.

```
# This Puppetfile is managed by Bolt. Do not edit.
# For more information, see https://pup.pt/bolt-modules

# The following directive installs modules to the managed moduledir.
moduledir '.modules'

mod 'KyleAnderson/consul', '6.1.0'
mod 'puppetlabs/stdlib', '6.5.0'
mod 'puppet/archive', '4.6.0'
mod 'camptocamp/systemd', '2.10.0'
[...]

PR #2315, opened 2020-11-03
(GH-2314) Fix module name regex for Forge/Git specifications

beechtom:
This updates the regex used to validate module names for Forge and Git
specifications. Previously, the regex for validating the owner segment
of the module name was `[a-z][a-z0-9_]`, which is the same regex used
for module name segments. However, owner names follow different
requirements and can only include letters (uppercase and lowercase) and
digits.

!bug

* **Fix module name validation for Forge and Git module specifications**
  ([#2314](https://github.com/puppetlabs/bolt/issues/2314))

  Forge and Git module specifications now correctly validate the
  module's name and permit uppercase letters in the owner segment of the
  module name. Previously, if the owner segment of a module name
  included uppercase letters, Bolt would raise an error.

### DIFF
--- a/lib/bolt/module_installer/specs/forge_spec.rb
+++ b/lib/bolt/module_installer/specs/forge_spec.rb
@@ -11,7 +11,7 @@ module Bolt
   class ModuleInstaller
     class Specs
       class ForgeSpec
-        NAME_REGEX    = %r{\A[a-z][a-z0-9_]*[-/](?<name>[a-z][a-z0-9_]*)\z}.freeze
+        NAME_REGEX    = %r{\A[a-zA-Z0-9]+[-/](?<name>[a-z][a-z0-9_]*)\z}.freeze
         REQUIRED_KEYS = Set.new(%w[name]).freeze
         KNOWN_KEYS    = Set.new(%w[name version_requirement]).freeze
 
@@ -33,8 +33,9 @@ module Bolt
           unless (match = name.match(NAME_REGEX))
             raise Bolt::ValidationError,
                   "Invalid name for Forge module specification: #{name}. Name must match "\
-                  "'owner/name', must start with a lowercase letter, and may only include "\
-                  "lowercase letters, digits, and underscores."
+                  "'owner/name'. Owner segment may only include letters or digits. Name "\
+                  "segment must start with a lowercase letter and may only include lowercase "\
+                  "letters, digits, and underscores."
           end
 
           [name.tr('-', '/'), match[:name]]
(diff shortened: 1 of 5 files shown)


lib/bolt/module_installer/specs/forge_spec.rb
```diff
@@ -11,7 +11,7 @@ module Bolt
   class ModuleInstaller
     class Specs
       class ForgeSpec
-        NAME_REGEX    = %r{\A[a-z][a-z0-9_]*[-/](?<name>[a-z][a-z0-9_]*)\z}.freeze
+        NAME_REGEX    = %r{\A[a-zA-Z0-9]+[-/](?<name>[a-z][a-z0-9_]*)\z}.freeze
```
beechtom, line 14: Based on username requirements for forge.puppet.com - only letters and digits permitted.

lib/bolt/module_installer/specs/git_spec.rb
```diff
@@ -11,7 +11,7 @@ module Bolt
   class ModuleInstaller
     class Specs
       class GitSpec
-        NAME_REGEX    = %r{\A(?:[a-z][a-z0-9_]*[-/])?(?<name>[a-z][a-z0-9_]*)\z}.freeze
+        NAME_REGEX    = %r{\A(?:[a-zA-Z0-9]+[-/])?(?<name>[a-z][a-z0-9_]*)\z}.freeze
```
beechtom, line 14: I'm not really sure if this should be the same as the Forge specs, or if it should be a bit more permissive:

```
\A(?:\w+[-/])?(?<name>[a-z][a-z0-9_]*)\z
```

lucywyman (changes requested):
This solution only works if the user correctly cases the module name when adding it to `bolt-project.yaml`. Could we also `.downcase` module specs before writing them to the Puppetfile so that adding modules isn't case sensitive?

```
 78             file.puts "moduledir '#{moduledir.basename}'"
 79             file.puts
 80           end
 81 
 82           @modules.each { |mod| file.puts mod.to_spec.downcase }
 83         end
```

I think this will require accounting for casing in a few other places:
```
  → Adding the following modules:
    puppetlabs/vcsrepo 3.1.1
    KyleAnderson/consul 6.1.0

  → Removing the following modules:
    kyleanderson/consul 6.1.0
```

beechtom:
Ah, good catch. It looks like this really only needs to be accounted for in `ForgeSpec.satisfied_by?`, so I updated that to downcase the spec and module names when comparing them. I'd prefer to keep the actual case of the module name that is returned by the resolver.

lib/bolt/module_installer/specs/git_spec.rb
lucywyman: This works just fine - git urls aren't case sensitive.

PR #2324, opened 2020-11-06
(GH-2140) Document escalating privilege in powershell scripts on Windows

### LINKED ISSUE #2140: Run-as administrator windows won't work.
Hi,

Since I can't find any documentation on this.
I'm currenly using bolt version 2.25.0

In my usecase I want to winrm to a windows node and run a script as administrator. But I login as a normal user who has administrator rights and not as Admin itself.

Here is my command: 

```
bolt script run windows_puppet_agent_6_17.ps1 --password-prompt --targets winrm://by1ack --no-ssl --user mjbjf@adpa3.local --run-as Administrator
```

When I run this command it executes it without admin priviliges. Is there a way like run-as to become admin when logging in as a normal user?
Or is it for this moment only possible by directly logging in as the adminstrator user.

Kind Regards

lucywyman:
The WinRM transport does not - and will not - support something akin to
a `run-as` command. This is because the way WinRM works is by connecting
with the user that has permissions to perform the actions, not by
elevating or changing to another user after connection. Put another way,
WinRM operates with a session level token and has no functionality to
change to another user after authentication.

Instead of escalating the user once connected, users can pass
credentials to scripts or plans running on the targets and use them
directly in commands. There are a number of ways to do this, each ideal
for different use cases. This adds a section to the 'Escalating
Privilege' document that describes the different ways to pass
credentials securely to either a script or plan to be used for
escalating on Windows. This includes using sensitive task parameters,
and describes setting and retrieving variables on Target objects
securely using Hiera data, Inventory plugins, and Remote tasks.

Closes #2140

!no-release-notes

### DIFF
--- a/documentation/templates/privilege_escalation.md.erb
+++ b/documentation/templates/privilege_escalation.md.erb
@@ -102,70 +102,72 @@ This enables using any executable and command-line options to change users. Keep
 
 ## Escalating privilege over WinRM
 
-
-- **Sensitive task parameters** are best if need your secrets obfuscated in logs. You can either
-  pass these as static values to a task directly, or can compute them per target and use a plan to
-  pass them to the task.
-- **Remote tasks** are best if your secrets are unique to each target and you don't want to write a
-  plan.
-
 The WinRM transport does not support `run-as`. This is because WinRM operates
 with a session level token and has no functionality to change to another user
 after authentication. Once you connect with a user, you can only perform
-actions that you have permission to perform with that user.  Instead of
-switching users, you can connect with a user, and then provide credentials to a
-program that accepts them to perform another action. The behavior of providing
+actions that you have permission to perform with that user. Instead of
+switching users, you can connect with one user, and then provide a secondary
+user's credentials to a program to perform another action. The behavior of providing
 credentials to a program is not part of WinRM.
 
 For example, to get the processes running on a Windows target, you could
 connect to the target with WinRM using `Invoke-BoltScript -Script
 .\myscript.ps1 -Targets winrm://mytarget -User userA -Password passwordA` and
-then use the following Powershell script to pass credentials to the
-`Get-WmiObject` command:
-```
+then use a Powershell script to pass credentials to the `Get-WmiObject` command:
+
+```powershell
 $Cred = New-Object System.Management.Automation.PsCredential('userB',$passwordB)
 Get-WmiObject -Class Win32_BIOS -Computer SERVER1 -Credential $Cred
 ```
 
-If you want to securely pass your userB credentials into a powershell script
-you have a few options.  The best way to pass parameters securely into a script
-is by [converting it to a task](writing_tasks.md#converting-scripts-to-tasks)
-and [defining sensitive
-parameters](writing_tasks.md#defining-sensitive-parameters) for that task. This
-will obfuscate the secret in logs and API responses.
+If you want to securely pass credentials into a Powershell script, you have two options: 
+- **[Convert your script into a task](writing_tasks.md#converting-scripts-to-tasks) and
+  use [sensitive task parameters](writing_tasks.md#defining-sensitive-parameters):** 
+  This is the best option if you need your secret obfuscated in logs. You can either 
+  pass the parameters as static values to a task, or compute them per target and use a
+  plan to pass them to the task.
+- **Use a remote task:** If your secrets are unique to each target and you don't want
+  to write a Bolt plan, you can use a remote task. This option requires you to manage
+  the connection to the target yourself, rather than letting Bolt manage the connection.
 
-You can pass sensitive parameters to a task as static values if your secrets are the
-same for every target. If your secrets are unique to each target, or need to
-be computed or fetched you will want to set your secret as data on the target
-in the inventory. For the purposes of escalating as a secondary user you can
-set arbitrary variables on targets in the inventory, which you can then access
-from a Bolt plan.
+If you're not sure where to start, we recommend using a task with sensitive
+parameters. If your secrets are unique to targets, you must also use an
+inventory plugin and a Bolt plan. [This method is demonstrated in the example
+below](#example).
 
-There are a few ways of setting and retrieving variables on targets. You can set variables on a
-target using an [inventory plugin](using_plugins.md). You can then access those
-variables in a plan, or directly in the task using a [remote
+If your secrets are the same for every target, you can pass sensitive
+parameters to a task as static values. If your secrets are unique to each
+target, or the secrets need to be computed or fetched, set the secrets as data
+on the target in the inventory. For the purposes of escalating as a secondary
+user, you can set arbitrary variables on targets in the inventory which you
+can then access from a Bolt plan.
+
+You can set [variables on the target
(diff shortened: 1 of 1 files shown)


lucywyman:
I wasn't entirely sure of what examples to provide and where, since most of these things aren't specific to windows and the workflows are so distinct. I don't think writing an example for each possible workflow would be a good use of time, and there are already smaller (platform-agnostic) examples of each element of the workflow on their respective pages. I tested each workflow on a Windows machine to verify it worked as expected, so happy to include examples anywhere they would fit!

beechtom (changes requested):
@lucywyman This is a generated doc, so the changes should be in the template.

lucywyman:
Booooo updated

documentation/templates/privilege_escalation.md.erb
```diff
@@ -99,3 +99,63 @@ This enables using any executable and command-line options to change users. Keep
...
 - Bolt will not use the `run-as-command` option if `run-as` is not set.
+
+## Escalating privilege over WinRM
+
+As you may have noted we do not support `run-as` for the WinRM transport. This is because the way
+WinRM works is by connecting with the user that has permissions to perform the actions, not by
+elevating or changing to another user after connection. WinRM operates with a session level token
+and has no functionality to change to another user after authentication. Instead, you can connect
+with one user, then provide credentials to a program that accepts them to perform another action,
+but that is not part of WinRm.
```
hestonhoffman, line 110:
```suggestion
The WinRM transport does not support `run-as`. This is because
WinRM operates with a session level token and has no functionality to change to another user after
authentication. Once you connect with a user, you can only perform actions that you have permission to
perform with that user.  Instead of switching users, you can connect with a user, and then provide credentials
to a program that accepts them to perform another action. The behavior of providing credentials to a 
program is not part of WinRM.
```

documentation/templates/privilege_escalation.md.erb
```diff
@@ -99,3 +99,63 @@ This enables using any executable and command-line options to change users. Keep
...
+
+For example, you can use `Invoke-BoltScript -Script .\myscript.ps1 -Targets winrm://mytarget -User
+userA -Password passwordA` to connect as a user you have privilege to connect as, then within
+`myscript.ps1` you can run:
```
hestonhoffman, line 114:
```suggestion
For example, to get the processes running on a Windows target, you could connect to the
target with WinRM using 
`Invoke-BoltScript -Script .\myscript.ps1 -Targets winrm://mytarget -User userA -Password passwordA`
and then use the following Powershell script to pass credentials to the `Get-WmiObject` command:
```

hestonhoffman (commented):
I have a couple comments, but I've run out of time! I'll get back to this on Monday :-)

documentation/templates/privilege_escalation.md.erb
```diff
@@ -99,3 +99,65 @@ This enables using any executable and command-line options to change users. Keep
...
+Get-WmiObject -Class Win32_BIOS -Computer SERVER1 -Credential $Cred
+```
+
+If you want to securely pass your userB credentials into a powershell script you have a few options.
+The best way to pass parameters securely into a script is by [converting it to a
+task](writing_tasks.md#converting-scripts-to-tasks) and [defining sensitive
+parameters](writing_tasks.md#defining-sensitive-parameters) for that task. This will obfuscate the
+secret in logs and API responses.
+
+You can pass sensitive parameters to a task directly if your secrets are the same for every target.
```
hestonhoffman, line 129: Does "directly" here mean via the command line?

documentation/templates/privilege_escalation.md.erb
```diff
@@ -99,3 +99,65 @@ This enables using any executable and command-line options to change users. Keep
...
+If your secrets are unique to each target, or need to be computed, fetched, or obfuscated when you
+pass them to the task you will want to set your secret as data on the target in the inventory. For
```
hestonhoffman, line 131: I think a small inventory example here would be really helpful. Would you set this under the `vars` key?

documentation/templates/privilege_escalation.md.erb
```diff
@@ -99,3 +99,65 @@ This enables using any executable and command-line options to change users. Keep
...
+the purposes of escalating as a secondary user you can set arbitrary variables on targets in the
+inventory, which you can then access.
+
+There are a few ways of setting and retrieving variables on targets. You can set variables on a
```
hestonhoffman, line 135: I'm a little confused here. I think it's just cos there's a lot going on in this paragraph and I'm not sure if it's describing an alternate way to do the same thing as the previous paragraph, or an additional thing that you need to do. I'm guessing it's an alternate way based on the summary later.

So far, we've covered using a task, defining sensitive parameters, and passing sensitive parameters into the task using an inventory file or directly. Is this paragraph covering a new use case - using an inventory plugin or Hiera data along with a remote task instead of defining sensitive parameters?

documentation/templates/privilege_escalation.md.erb
```diff
@@ -99,3 +99,65 @@ This enables using any executable and command-line options to change users. Keep
...
+target using an [inventory plugin](using_plugins.md), or using [Hiera
+data](applying_manifest_blocks.md#using-hiera-data-in-a-manifest-block). You can then access those
+variables in a plan, or if you're using inventory plugins you can access them directly in the task
+using a [remote task](writing_tasks.md#writing-remote-tasks). Remote tasks run on localhost and pass
+the entire target object as json to the task as a metaparameter `_target`. You would write a task to
+get the secret, then connect to the remote target and use the secret there. Put another way, you
+have to manage the connection to the remote target in your task rather than letting Bolt manage it
+for you.
+
+To summarize:
```
hestonhoffman, line 145: Instead of summarizing at the end, let's move the summary bit to the top - that's a good way to signpost to the reader that you're going to discuss two possible options/use cases

documentation/templates/privilege_escalation.md.erb
lucywyman: Yeah, at the time I was thinking "without having to do anything fancy involving plans or inventory", but I think 'via the command line or directly to `run_task`' would also work?

documentation/templates/privilege_escalation.md.erb
lucywyman: I'm just going to remove the Hiera bit. You can definitely use it along these lines, but it's so complicated and I think anyone who isn't familiar with Hiera can just do the inventory workflow

lucywyman:
Ok, I rewrote this to be more clear and tried to make it more clear how to choose which solution was best. I removed the Hiera bit because it's a whole other can of worms and I think this solution is simpler and will work for most users, but using Hiera is really powerful and kind of cool.

I feel like I said the same thing like 3 times, could definitely use a good editor!

hestonhoffman:
Hi Lucy, I was having trouble editing this in the GUI, so I pushed up a PR to your fork instead: https://github.com/lucywyman/bolt/pull/11

lucywyman:
Merged and made some small edits

beechtom (approved):

hestonhoffman (approved):
LGTM!

Issue #2303, opened 2020-10-29
Support alternate Forge for 'bolt module' workflow


nossralf:
## Use Case
There is currently no way of specifying an alternate Forge URL when using the new `bolt module` workflow for managing module dependencies. This is [supported](https://puppet.com/docs/bolt/latest/bolt_project_reference.html#forge) in the older `bolt puppetfile` workflow, and something that is almost a necessity in a corporate environment where Bolt plans may rely on internal modules that cannot be published to the Puppet Forge.

We should make a new setting with the exact same structure and keys as `puppetfile` config option. It's called `module-install`.

We should also update bolt project migrate to update this setting to update `puppetfile` to `module-install`.

If you have the (`modules` or `module-install`) key and the `puppetfile` key you will get a warning to update the `puppetfile` key to `module-install` with `bolt project migrate`. It will not honor the puppetfile key. `module-install` should not be honored by `puppetfile` subcommands.

op-ct:
> "In a corporate environment where Bolt plans may rely on internal modules that cannot be published to the Puppet Forge" 

It would be even more valuable to support git repos (and refs) for individual modules in these environments (as the older `bolt puppetfile` does)—it is much more common for corporate environments to serve modules from internal git repos than from private Puppet Forges.

PR #2325, opened 2020-11-06
(GH-2303) Add module-install config option

beechtom:
This adds a new `module-install` config option that is meant to replace
`puppetfile`. The config option accepts the same sub-options as
`puppetfile` does.

Currently, this config option is not respected when resolving module
dependencies, as changes need to be made to the `puppetfile-resolver`
library to support proxies and an alternate Forge. However, this config
is still passed to r10k, which will respect it when installing modules
from a Puppetfile.

If a user has `module-install` configured and then runs `bolt module
add|install` or `Add|Install-BoltModule`, Bolt will issue a warning
letting the user know that the config option will not be respected when
resolving modules, and then provides a link to a known issue with more
information.

This also adds a handful of warnings that are issued depending on
whether `puppetfile` or `module-install` are configured and if the user
runs `puppetfile install` or `module install`. If both options are
configured, Bolt will also warn and use `module-install`.

This also updates the project migrator to update the `puppetfile` option
to `module-install`.

!feature

* **Add `module-install` configuration option**
  ([#2303](#2303))

  Bolt now supports a `module-install` configuration option in
  `bolt-project.yaml` and `bolt-defaults.yaml`. This option is used to
  configure proxies and an alternate forge when installing modules using
  the `bolt module add|install` commands or `Add|Install-BoltModule`
  cmdlets.

  _This option is not currently supported when resolving module
  dependencies._

### DIFF
--- a/documentation/bolt_known_issues.md
+++ b/documentation/bolt_known_issues.md
@@ -1,5 +1,94 @@
 # Known issues
 
+## Resolving module dependencies does not support proxies or alternate Forge
+
+When running the `bolt module add|install` commands or `Add|Install-BoltModule`
+cmdlets, Bolt does not support a configured proxy or alternate Forge when it
+resolves module dependencies, even if the `module-install` configuration
+option is set. Support for resolving module dependencies with proxies or an
+alternate Forge requires changes to one of Bolt's gem dependencies, which is
+currently in progress.
+
+While Bolt will not resolve module dependencies with proxies or an alternate
+Forge, it will respect this configuration when installing modules.
+
+If your project configures the `module-install` option, you may experience
+one of the following issues:
+
+- On a restricted network where a proxy is required, resolving modules may
+  cause Bolt to fail.
+
+- When an alternate Forge is specified, Bolt may resolve and install a
+  different set of modules than expected.
+
+### Install modules without resolving dependencies
+
+To avoid this limitation and any potential errors, you can manually edit your
+Puppetfile to add and install modules without resolving dependencies. This is a
+similar workflow to using the `bolt puppetfile install` and
+`Install-BoltPuppetfileModules` commands.
+
+For example, if your project requires the `puppetlabs/apache` module, but
+you need to download modules hosted on an alternate Forge, you should write
+your Puppetfile manually. That Puppetfile might look similar to this:
+
+```ruby
+mod 'puppetlabs/apache', '5.7.0'
+mod 'puppetlabs/stdlib', '6.5.0'
+mod 'puppetlabs/concat', '6.3.0'
+mod 'puppetlabs/translate', '2.2.0'
+```
+
+Once you have a Puppetfile, you can install modules without resolving
+dependencies using the `no-resolve` command-line option:
+
+_\*nix command_
+
+```shell
+bolt module install --no-resolve
+```
+
+_PowerShell cmdlet_
+
+```powershell
+Install-BoltModule -NoResolve
+```
+
+If you need to add a module to your project, manually add the module and
+its dependencies to your Puppetfile and then run the above command again.
+
+### Set the HTTP proxy environment variables
+
+If you only need to configure a proxy to work around network restrictions,
+you can set the HTTP proxy environment variables when you run Bolt. For
+example, if you only configure the `proxy` option under `module-install` like
+this:
+
+```yaml
+# bolt-project.yaml
+module-install:
+  proxy: http://myproxy.com
+```
+
+Then you can set the `HTTP_PROXY` and `HTTPS_PROXY` environment variables
+to this value when you run `bolt module add|install` or `Add|Install-BoltModule`.
+This will configure a proxy that Bolt will use when it makes requests to the
+Puppet Forge and GitHub when it resolves modules.
+
+_\*nix command_
+
+```shell
+HTTP_PROXY=http://myproxy.com HTTPS_PROXY=http://myproxy.com bolt module install
+```
+
+_PowerShell cmdlet_
+
+```powershell
+SET HTTP_PROXY=http://myproxy.com
+SET HTTPS_PROXY=http://myproxy.com
+Install-BoltModule
+```
+
 ## `facts` task fails on Windows targets with Facter 3 installed
 
 When running the `facts` task on a Windows target that has Facter 3 installed,
--- a/lib/bolt/cli.rb
+++ b/lib/bolt/cli.rb
@@ -491,9 +491,9 @@ module Bolt
       when 'module'
         case options[:action]
         when 'add'
-          code = add_project_module(options[:object], config.project)
+          code = add_project_module(options[:object], config.project, config.module_install)
         when 'install'
-          code = install_project_modules(config.project, options[:force], options[:resolve])
+          code = install_project_modules(config.project, config.module_install, options[:force], options[:resolve])
         when 'generate-types'
           code = generate_types
         end
@@ -743,7 +743,7 @@ module Bolt
 
     # Installs modules declared in the project configuration file.
     #
-    def install_project_modules(project, force, resolve)
+    def install_project_modules(project, config, force, resolve)
       assert_project_file(project)
       assert_puppetfile_or_module_command(project.modules)
 
@@ -753,11 +753,21 @@ module Bolt
         return 0
       end
 
+      if resolve != false && config.any?
+        @logger.warn(
+          "Detected configuration for 'module-install'. This configuration is currently "\
+          "only supported when installing modules, not when resolving module dependencies. "\
+          "For more information, see https://pup.pt/bolt-module-install"
+        )
+      end
+
+      modules   = project.modules || []
       installer = Bolt::ModuleInstaller.new(outputter, pal)
 
-      ok = installer.install(project.modules,
+      ok = installer.install(modules,
                              project.puppetfile,
                              project.managed_moduledir,
+                             config,
                              force: force,
                              resolve: resolve)
       ok ? 0 : 1
@@ -765,9 +775,17 @@ module Bolt
 
     # Adds a single module to the project.
     #
-    def add_project_module(name, project)
+    def add_project_module(name, project, config)
       assert_project_file(project)
       assert_puppetfile_or_module_command(project.modules)
+
+      if config.any?
+        @logger.warn(
+          "Detected configuration for 'module-install'. This configuration is currently "\
+          "only supported when installing modules, not when resolving module dependencies. "\
+          "For more information, see https://pup.pt/bolt-module-install"
+        )
+      end
 
       modules   = project.modules || []
       installer = Bolt::ModuleInstaller.new(outputter, pal)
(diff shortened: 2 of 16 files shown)


lucywyman (commented):
I think if the user has `module-install` without `modules`, or has `puppetfile` with `modules`, we should error and say that config isn't getting picked up.

hestonhoffman (approved):
Docs LGTM!

schemas/bolt-defaults.schema.json
```diff
@@ -222,7 +252,7 @@
               "format": "uri"
             },
             "proxy": {
-              "description": "The HTTP proxy to use for Git and Forge operations.",
+              "description": "The HTTP proxy to use for Forge operations.",
```
lucywyman, line 255: I believe this can still be used for Git modules manually declared in the Puppetfile.

schemas/bolt-project.schema.json
```diff
@@ -298,7 +328,7 @@
               "format": "uri"
             },
             "proxy": {
-              "description": "The HTTP proxy to use for Git and Forge operations.",
```
lucywyman, line 301: Same here - this is still valid for Git operations

lib/bolt/config/options.rb
```diff
@@ -375,19 +411,19 @@ module Options
                   _example: "https://forge.example.com"
                 },
                 "proxy" => {
-                  description: "The HTTP proxy to use for Git and Forge operations.",
```
lucywyman, line 378: Still valid for Git

lucywyman (changes requested):
This should include unit tests for the config conflict cases (`module_install`, `modules`, and `puppetfile`) and it'd be great to add an integration test that verifies that the config gets picked up and used. I actually think the easiest way to do that is to use a fake proxy/baseurl and ensure it errors correctly.

schemas/bolt-defaults.schema.json
beechtom: This is under `forge` and only applies to Forge operations. The `proxy` setting one level higher is the global setting that applies to both git and Forge operations.

lucywyman: Ah ok - sorry, the git diff folded the surrounding lines!

beechtom:
Updated.

lucywyman (approved):
I think some tests still need fixing, but once it's green + the Puppetfile Resolver is merged this is gtg!

lucywyman:
Resolve from the Forge but download from proxies. For resolution, use the module-install proxy. Create a known issue around resolving from proxies. If resolution doesn't work, just install with no-resolve. We can always recommend users use the old workflow, even if it's deprecated.

beechtom:
Ok, I think I've got this where we want it until `puppetfile-resolver` is updated. Things to note:

- Migrator updates `puppetfile` to `module-install`
- Bolt warns when an option is configured but the opposite command is used (i.e. `puppetfile` + `module install` warns)
- Bolt warns when `module-install` is configured and `module add|install` is run, informing the user it's not supported for resolving dependencies
- Added a known issue describing workarounds
- Left in most of the work needed to support this option when resolving, including marking spec tests as pending. When `puppetfile-resolver` is updated we just need to pass the config along in `Bolt::ModuleInstall::Resolver.resolve`.

lib/bolt/config.rb
```diff
@@ -497,6 +497,25 @@ def validate
       unless TRANSPORT_CONFIG.include?(transport)
         raise UnknownTransportError, transport
       end
+
+      # Warn the user how they should be using the 'puppetfile' or
+      # 'module-install' config options. We don't error here since these
+      # settings can be set at the user or system level.
+      if @project.modules && puppetfile_config.any? && module_install.empty?
+        command = Bolt::Util.powershell? ? 'bolt project migrate' : 'Update-BoltProject'
```
lucywyman, line 505: I think these are backwards

lib/bolt/module_installer/resolver.rb
```diff
@@ -9,7 +9,7 @@ class ModuleInstaller
     class Resolver
       # Resolves module specs and returns a Puppetfile object.
       #
-      def resolve(specs)
+      def resolve(specs, _config = {})
```
lucywyman, line 12: Can we wait to add this when the config gets used?

lib/bolt/cli.rb
```diff
@@ -753,30 +753,49 @@ def install_project_modules(project, force, resolve)
...
-    def add_project_module(name, project)
+    def add_project_module(name, project, config)
       assert_project_file(project)
       assert_puppetfile_or_module_command(project.modules)
 
+      if config.any?
+        @logger.warn(
+          "Detected configuration for 'module-install'. This configuration is currently "\
+          "only supported when installing modules, not when resolving module dependencies. "\
+          "For more information, see https://pup.pt/bolt-module-install"
```
lucywyman, line 786: Might want to set up the redirect before merging just so we don't forget (speaking from experience :grimacing: )

lib/bolt/config.rb
beechtom: But what if I _want_ to confuse people? :P Fixed.

lib/bolt/module_installer.rb
```diff
@@ -49,28 +49,28 @@ def add(name, specs, puppetfile_path, moduledir, config_path)
 
       begin
         resolve_specs.add_specs('name' => name)
-        puppetfile = Resolver.new.resolve(resolve_specs)
+        puppetfile = Resolver.new.resolve(resolve_specs, config)
       rescue Bolt::Error
         project_specs.add_specs('name' => name)
-        puppetfile = Resolver.new.resolve(project_specs)
+        puppetfile = Resolver.new.resolve(project_specs, config)
```
lucywyman, line 55: I think we also want to catch if this fails and install module as if the user had passed `--no-resolve`

beechtom: I'm not sure what you mean... `bolt module add` doesn't support `--no-resolve`, and it doesn't really make sense to me to support loading modules from the Puppetfile and adding a new module to it. If you're using `--no-resolve` to install modules, you should be adding new modules manually.

lucywyman: Can we catch an error here then and, if `module-install` is configured recommend the manual workflow?

lib/bolt/module_installer.rb
```diff
@@ -155,7 +155,7 @@ def install(specs, path, moduledir, force: false, resolve: true)
         # and write a Puppetfile.
         if force || !path.exist?
           @outputter.print_action_step("Resolving module dependencies, this may take a moment")
-          puppetfile = Resolver.new.resolve(specs)
+          puppetfile = Resolver.new.resolve(specs, config)
```
lucywyman, line 158: This we want to catch when it errors and install as if `--no-resolve` was passed ([per comment](https://github.com/puppetlabs/bolt/pull/2325#issuecomment-737528220))

lib/bolt/module_installer.rb
beechtom: While config is currently being passed to `Bolt::ModuleInstaller::Resolver`, it doesn't actually get passed to `puppetfile-resolver`. If resolution is failing here, it's because the modules cannot be resolved at all. I don't think we should be recommending the manual workflow in that case.

lib/bolt/module_installer.rb
beechtom: I think the only reasonable way to catch an error here is if we can distinguish between resolution errors and errors caused by a restricted network (I think that's the only kind of error a user might run into by `module-install` not being passed along, anyways). Considering a user will receive a warning that `module-install` does not work for resolution yet, and it links them to fairly detailed documentation on getting around that, I lean more towards letting this fail and the user deciding how to get around it.

documentation/bolt_known_issues.md
```diff
@@ -1,5 +1,94 @@
...
+- On a restricted network where a proxy is required, resolving modules may
+  cause Bolt to fail.
+
+- When an alternate Forge is specified, Bolt may resolve and install a
+  different set of modules than expected.
+
+### Install modules without resolving dependencies
+
+To avoid this limitation and any potential errors, you can manually edit your
+Puppetfile and install modules without resolving dependencies. This is a similar
```
lucywyman, line 27:
```suggestion
Puppetfile to add and install modules without resolving dependencies. This is a similar
```

lib/bolt/module_installer.rb
lucywyman: But the resolution might fail because users have a module published to their alternate Forge (that they've configured) that isn't available on the public Forge, which can cause resolution to fail. In that case users can manually add the module to their Puppetfile and still use `module install --no-resolve` to install from the correct place, even though resolution will fail.

beechtom: I'm not particularly keen on falling back to `--no-resolve` behavior if resolution fails, since I think it results in somewhat unintuitive behavior. I think there are two approaches we could take instead:

- Only pass `module-install` config if a user runs `module install --no-resolve`. Otherwise, warn that the config is not used and link to the known issue. Since we currently _only_ support this config for installing modules, not resolving, restricting the config to only being used when using the `--no-resolve` flag makes sense. This also puts users who need this config in the habit of running `--no-resolve` instead of expecting Bolt to try and resolve modules and potentially succeed with a different set of modules than they're expecting.

- Immediately fall back to `--no-resolve` behavior for `module install` if `module-install` is configured. This way Bolt won't even attempt to resolve modules. We can add a warning here as well that Bolt is skipping resolution because `module-install` is configured, with a link to the known issue. This allows users to still run `module install` to set up a project they downloaded, though it does still add what I think is slightly unintuitive behavior to the command for project authors.

lucywyman: I think the best thing to do for both `module add` (here) and `module install` (below) is to try to resolve, catch errors and print a message telling the user they can try manually adding the module to the Puppetfile and installing with `--no-resolve` if they have `module-install.forge.baseurl` configured.

documentation/bolt_known_issues.md
```diff
@@ -1,5 +1,94 @@
...
+  cause Bolt to fail.
+
+- When an alternate Forge is specified, Bolt may resolve and install a
+  different set of modules than expected.
+
+### Install modules without resolving dependencies
+
+To avoid this limitation and any potential errors, you can manually edit your
+Puppetfile to add and install modules without resolving dependencies. This is a
+similar workflow as the `bolt puppetfile install` and
```
hestonhoffman, line 28:
```suggestion
similar workflow to using the `bolt puppetfile install` and
```

documentation/bolt_known_issues.md
```diff
@@ -1,5 +1,94 @@
...
+`Install-BoltPuppetfileModules` commands.
+
+For example, if your project requires the `puppetlabs/apache` module, but
+you need to download modules hosted on an alternate Forge, you should write
+your Puppetfile manually. That Puppetfile may look similar to this:
```
hestonhoffman, line 33:
```suggestion
your Puppetfile manually. That Puppetfile might look similar to this:
```
Just a style thing - we prefer 'might' over 'may'

hestonhoffman (approved):
Couple small changes, but LGTM!

PR #2326, opened 2020-11-06
Show PowerShell cmdlets in output when running in PowerShell

beechtom:
When running certain commands or raising certain errors, Bolt will
output a command to run. This change checks whether Bolt is running in
PowerShell or not, and sets the command in the output to a PowerShell
cmdlet if it is or a *nix shell command if it is not.

!no-release-note

### DIFF
--- a/lib/bolt/cli.rb
+++ b/lib/bolt/cli.rb
@@ -197,10 +197,6 @@ module Bolt
 
       warn_inventory_overrides_cli(options)
 
-      # Assert whether the puppetfile/module commands are available depending
-      # on whether 'modules' is configured.
-      assert_puppetfile_or_module_command(config.project.modules)
-
       options
     rescue Bolt::Error => e
       outputter.fatal_error(e)
@@ -229,9 +225,9 @@ module Bolt
 
     def validate(options)
       unless COMMANDS.include?(options[:subcommand])
+        command = Bolt::Util.powershell? ? 'Get-Command -Module PuppetBolt' : 'bolt help'
         raise Bolt::CLIError,
-              "Expected subcommand '#{options[:subcommand]}' to be one of " \
-              "#{COMMANDS.keys.join(', ')}"
+              "'#{options[:subcommand]}' is not a Bolt command. See '#{command}'."
       end
 
       actions = COMMANDS[options[:subcommand]]
@@ -286,8 +282,9 @@ module Bolt
       end
 
       if options[:subcommand] == 'module' && options[:action] == 'install' && options[:object]
+        command = Bolt::Util.powershell? ? 'Add-BoltModule -Module' : 'bolt module add'
         raise Bolt::CLIError, "Invalid argument '#{options[:object]}'. To add a new module to "\
-                              "the project, run 'bolt module add #{options[:object]}'."
+                              "the project, run '#{command} #{options[:object]}'."
       end
 
       if options[:subcommand] != 'file' && options[:subcommand] != 'script' &&
@@ -466,7 +463,8 @@ module Bolt
       when 'plan'
         case options[:action]
         when 'new'
-          @logger.warn("Command 'bolt plan new' is experimental and subject to changes.")
+          command = Bolt::Util.powershell? ? 'New-BoltPlan' : 'bolt plan new'
+          @logger.warn("Command '#{command}' is experimental and subject to changes.")
           plan_name = options[:object]
 
           # If this passes validation, it will return the path to the plan to create
@@ -717,10 +715,12 @@ module Bolt
     end
 
     def list_modules
+      assert_puppetfile_or_module_command(config.project.modules)
       outputter.print_module_list(pal.list_modules)
     end
 
     def generate_types
+      assert_puppetfile_or_module_command(config.project.modules)
       # generate_types will surface a nice error with helpful message if it fails
       pal.generate_types
       0
@@ -741,8 +741,9 @@ module Bolt
             "project name must begin with a lowercase letter and can include lowercase "\
             "letters, numbers, and underscores."
         else
+          command = Bolt::Util.powershell? ? 'New-BoltProject -Name <NAME>' : 'bolt project init <NAME>'
           raise Bolt::ValidationError, "The current directory name '#{name}' is an invalid "\
-            "project name. Please specify a name using 'bolt project init <name>'."
+            "project name. Please specify a name using '#{command}'."
         end
       end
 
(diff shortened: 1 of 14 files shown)


lib/bolt/cli.rb
```diff
@@ -229,9 +229,9 @@ def update_targets(options)
 
     def validate(options)
       unless COMMANDS.include?(options[:subcommand])
+        command = Bolt::Util.powershell? ? 'Get-Command -Module PuppetBolt' : 'bolt help'
```
lucywyman, line 232: This doesn't work if the user hasn't loaded the powershell module. I think we can see if the user has the module loaded with `Get-Module -ListAvailable -Name PuppetBolt` and switch on that instead?

lib/bolt/cli.rb
```diff
@@ -229,9 +229,9 @@ def update_targets(options)
...
         raise Bolt::CLIError,
-              "Expected subcommand '#{options[:subcommand]}' to be one of " \
-              "#{COMMANDS.keys.join(', ')}"
+              "'#{options[:subcommand]}' is not a Bolt command. See '#{command}'."
```
lucywyman, line 234: Dang much better phrasing :clap:

lucywyman (commented):
I think a few places were also missed:
https://github.com/puppetlabs/bolt/blob/main/lib/bolt/bolt_option_parser.rb#L199
https://github.com/puppetlabs/bolt/blob/main/lib/bolt/project_migrator/config.rb#L60
https://github.com/puppetlabs/bolt/blob/main/lib/bolt/outputter/human.rb#L299
https://github.com/puppetlabs/bolt/blob/main/lib/bolt/inventory/group.rb#L247
And in cli.rb at lines:
290
469
869
872
and 875

lib/bolt/cli.rb
beechtom: Shouldn't the powershell module be loaded automatically when Bolt is installed as a package?

lib/bolt/cli.rb
beechtom: Thank you git for the inspo ;)

beechtom:
Updated.

I left out https://github.com/puppetlabs/bolt/blob/main/lib/bolt/bolt_option_parser.rb#L199 because that help text isn't used in the Powershell module and is already specific to the *nix shell commands. It seems incorrect to show a Powershell cmdlet in the help text for *nix shell commands.

Similarly, https://github.com/puppetlabs/bolt/blob/main/lib/bolt/outputter/human.rb#L299 was left out since there is no Powershell cmdlet for the topic guides and this list won't be shown when using the Powershell cmdlet `Get-Help about_bolt_*`.

lib/bolt/cli.rb
lucywyman: https://puppet.com/docs/bolt/latest/bolt_installing.html#add-the-bolt-module-to-powershell I doubt many people are using Powershell 2 or 3 but it generally just seems better to verify the cmdlets are loaded before recommending a cmdlet? Does `Get-Module` add a lot of time? There might be more efficient ways to check...

lib/bolt/cli.rb
```diff
@@ -861,17 +866,35 @@ def install_puppetfile(config, puppetfile, moduledir)
...
+        when 'generate-types'
+          old_command = 'Register-BoltPuppetfileTypes'
+          new_command = 'Register-BoltModuleTypes'
+        when 'install'
+          old_command = 'Install-BoltPuppetfile'
+          new_command = 'Install-BoltModule'
+        when 'show-modules'
+          old_command = 'Get-BoltPuppetfileModules'
+          new_command = 'Get-BoltModule'
+        end
```
lucywyman, line 880: This also needs a clause for 'add' and 'show' if you're using the `bolt module` command without `modules` in config.

lib/bolt/cli.rb
```diff
@@ -837,11 +840,13 @@ def add_project_module(name, project)
     def assert_project_file(project)
```
lucywyman, line 840: Huh! This is actually dead code, because if the user doesn't have a projectfile this will be raised:
```
Unable to use command 'bolt module install' when 'modules' is not configured in bolt-project.yaml. Use 'bolt puppetfile install' instead.
```
From line 202 first. I think I'd move `assert_puppfile_or_module` to be called in the `puppetfile` and `module` handlers, maybe even just after `assert_projectfile`, but you might have better ideas of how that flow should work.

lib/bolt/module_installer.rb
```diff
@@ -194,5 +194,29 @@ def install_puppetfile(path, moduledir, config = {})
 
       ok
     end
+
+    # Asserts that an existing Puppetfile is managed by Bolt.
+    #
+    private def assert_managed_puppetfile(puppetfile, path)
```
lucywyman, line 200: I don't see where this gets called?

lib/bolt/outputter/human.rb
```diff
@@ -225,20 +226,26 @@ def print_task_info(task)
...
+                else
+                  +"bolt task run #{task.name} --targets <targets>"
+                end
 
         task.parameters&.each do |k, v|
           pretty_params << "- #{k}: #{v['type'] || 'Any'}\n"
           pretty_params << "    Default: #{v['default'].inspect}\n" if v.key?('default')
           pretty_params << "    #{v['description']}\n" if v['description']
-          usage << if v['type'].is_a?(Puppet::Pops::Types::POptionalType)
+          usage << if v['type'].start_with?("Optional")
```
lucywyman, line 239: Why this change?

lib/bolt/util.rb
```diff
@@ -273,6 +273,11 @@ def windows?
         !!File::ALT_SEPARATOR
       end
 
+      # Returns true if running in PowerShell.
+      def powershell?
+        !!ENV['PSModulePath']
```
lucywyman, line 278: @jpogran Does this seem like a reasonable way to know if we're in powershell vs. CMD or *sh?

beechtom: This is actually his suggestion!

lib/bolt/outputter/human.rb
beechtom: Because it's a bug. `v['type']` is the stringified type, so `is_a?(Puppet::Pops::Types::POptionalType)` will always be false.

lib/bolt/module_installer.rb
beechtom: Leftover from before the refactor of this class.

lib/bolt/cli.rb
lucywyman: Ok, it looks like in the last month, out of 278 Windows users we've had 7 on Windows 7 and Windows 2012 (non-r2), which have Powershell3 by default. I feel ok leaving this as is and modifying if anyone complains.

PR #2328, opened 2020-11-09
(feature) Allow plugins to return sensitive data

lucywyman:
This gives users the ability to return sensitive data from configuration
plugins. Users can specify sensitive data by wrapping the `value` key in
a `_sensitive` key in the result of the plugin. Bolt will then
automatically unwrap the sensitive value and use the `value` hash as the
result. This prevents the output of the plugin from being logged when we
log task results in the executor. This is particularly useful for users
who want to return passwords and other sensitive data from plugins.

!feature

* **Specify sensitive value in plugins**

  Users can specify their plugin returns as 'sensitive', which will
  prevent the results from being logged.


adreyer:
It seems like enough plugins are sensitive that we should not log them at info rather than require plugin authors to modify their plugins.

lucywyman:
Closing in favor of #2329

PR #2334, opened 2020-11-11
(GH-2304) Set default ports for Puppet services if not specified

### LINKED ISSUE #2304: Include ports in example URLs

Everywhere that we document example URLs for services that require ports, we should include the default port. This is 
at least an issue for [service-url for the pcp transport](https://puppet.com/docs/bolt/latest/bolt_transports_reference.html#pcp). We should audit and make sure we update any other examples like that.

As part of this we can raise a warning when puppetdb urls and orchestrator urls don't have a port and instruct users to add a port. Silence this warning by just putting 443.

lucywyman:
Previously, the docs contained a few examples of the `service-url`
config needed to connect Bolt to the Orchestrator that didn't include
the port. Users assumed this meant the port didn't need to be set in
config, and tried to connect to the orchestrator using the default HTTPS
port 443 instead of the typical Orchestrator port 8143. This updates the
documentation to include the port for URLs that don't have standard
ports (80 for HTTP, 443 for HTTPS). It also sets the default ports for
both PuppetDB URLs and Orchestrator URLs to 8081 and 8143 respectively
if the user does not specify a port.

Closes #2304 

!feature

* **Set default ports for PuppetDB and Orchestrator** ([2304](https://github.com/puppetlabs/bolt/issues/2304))

  We now set the ports for PuppetDB `server_urls` and Orchestrator
  `service-url` to 8081 and 8143 respectively if the port is not set in
  config.


lib/bolt/transport/orch/connection.rb
```diff
@@ -17,13 +17,16 @@ def self.get_key(opts)
...
           @logger = logger
           @key = self.class.get_key(opts)
-          client_keys = %w[service-url token-file cacert job-poll-interval job-poll-timeout]
-          client_opts = client_keys.each_with_object({}) do |k, acc|
-            acc[k] = opts[k] if opts.include?(k)
-          end
+          client_opts = opts.slice('token-file', 'cacert', 'job-poll-interval', 'job-poll-timeout')
+          uri = Addressable::URI.parse(opts['service-url'])
+          uri&.port ||= 8143
+          client_opts['service-url'] = uri.to_s
```
beechtom, line 27:
```suggestion
          if opts.key?('service-url')
            uri = Addressable::URI.parse(opts['service-url'])
            uri&.port ||= 8143
            client_opts['service-url'] = uri.to_s
          end
```

This will result in `service-url` being set to an empty string (`nil.to_s == ""`) if it's not set or not parseable, resulting in a misleading error:

```
  '' is an invalid service-url
```

We could just error here if `service-url` is not set or is not parseable since it's a required option, but that seems less preferable.

hestonhoffman (approved):
👍

beechtom (approved):

beechtom (approved):

Issue #2337, opened 2020-11-11
Bolt shouldn't stack trace if config is invalid


lucywyman:
## Describe the Bug
Bolt will stack trace with the following config file (or any instance of `puppetfile` with no content) when running any command:
```
---
puppetfile:
```
```
lucy ~/githubs/bolt/project ((0ed604c...)) $ be bolt puppetfile install
bundler: failed to load command: bolt (/home/lucy/githubs/bolt/vendor/bundle/ruby/2.6.0/bin/bolt)
NoMethodError: undefined method `slice' for nil:NilClass
  /home/lucy/githubs/bolt/lib/bolt/config.rb:337:in `finalize_data'
  /home/lucy/githubs/bolt/lib/bolt/config.rb:254:in `initialize'
  /home/lucy/githubs/bolt/lib/bolt/config.rb:51:in `new'
  /home/lucy/githubs/bolt/lib/bolt/config.rb:51:in `from_project'
  /home/lucy/githubs/bolt/lib/bolt/cli.rb:177:in `load_config'
  /home/lucy/githubs/bolt/lib/bolt/cli.rb:95:in `parse'
  /home/lucy/githubs/bolt/exe/bolt:10:in `<top (required)>'
  /home/lucy/githubs/bolt/vendor/bundle/ruby/2.6.0/bin/bolt:23:in `load'
  /home/lucy/githubs/bolt/vendor/bundle/ruby/2.6.0/bin/bolt:23:in `<top (required)>'
```
Yes - we should have a more general solution.

PR #2345, opened 2020-11-12
(GH-2337) Validate config using schemas

beechtom:
This adds a new `Bolt::Config::Validator` class which validates config
against a schema. The validator will recurse through the config and
schema, validating each value against the corresponding definition in
the schema. If validation fails for a given value, the validator will
add a message to a list of errors. After the entire config has been
validated, and if any invalid values were detected, the validator will
raise an error that lists all of the invalid values.

The validator provides basic validation for config values, such as type
checking, ensuring required keys are present, and that items in an array
are unique. It does not provide more complex validation such as ensuring
a file exists at a given path or that a URL can be reached.

!feature

* **Improved config validation**
  ([#2337](https://github.com/puppetlabs/bolt/issues/2337))

  Bolt now validates config files against Bolt's schemas and indicates
  which config file an error is found in.

### DIFF
--- a/lib/bolt/config.rb
+++ b/lib/bolt/config.rb
@@ -7,6 +7,7 @@ require 'bolt/project'
 require 'bolt/logger'
 require 'bolt/util'
 require 'bolt/config/options'
+require 'bolt/config/validator'
 
 module Bolt
   class UnknownTransportError < Bolt::Error
@@ -73,6 +74,16 @@ module Bolt
       new(project, data, overrides)
     end
 
+    def self.defaults_schema
+      base = OPTIONS.slice(*BOLT_DEFAULTS_OPTIONS)
+      base['inventory-config'][:properties] = TRANSPORT_CONFIG.transform_values(&:schema)
+      base
+    end
+
+    def self.bolt_schema
+      OPTIONS.slice(*BOLT_OPTIONS).merge(TRANSPORT_CONFIG.transform_values(&:schema))
+    end
+
     def self.system_path
       # Lazy-load expensive gem code
       require 'win32/dir' if Bolt::Util.windows?
@@ -105,6 +116,10 @@ module Bolt
           "will be ignored."
         )
       end
+
+      # Validate the config against the schema. This will raise a single error
+      # with all validation errors.
+      Validator.new.validate(data, defaults_schema, filepath)
 
       # Remove project-specific config such as hiera-config, etc.
       project_config = data.slice(*(BOLT_PROJECT_OPTIONS - BOLT_DEFAULTS_OPTIONS))
@@ -160,6 +175,10 @@ module Bolt
       deprecations = [{ type: 'Using bolt.yaml for system configuration',
                         msg: "Configuration file #{filepath} is deprecated and will be removed in a future version "\
                         "of Bolt. Use '#{dir + BOLT_DEFAULTS_NAME}' instead." }]
+
+      # Validate the config against the schema. This will raise a single error
+      # with all validation errors.
+      Validator.new.validate(data, bolt_schema, filepath)
 
       { filepath: filepath, data: data, logs: logs, deprecations: deprecations }
     end
@@ -271,7 +290,7 @@ module Bolt
 
       # Set console log to debug if in debug mode
       if options[:debug]
-        overrides['log'] = { 'console' => { 'level' => :debug } }
+        overrides['log'] = { 'console' => { 'level' => 'debug' } }
       end
 
       if options[:puppetfile_path]
@@ -279,6 +298,9 @@ module Bolt
       end
 
       overrides['trace'] = opts['trace'] if opts.key?('trace')
+
+      # Validate the overrides
+      Validator.new.validate(overrides, OPTIONS, 'command line')
 
       overrides
     end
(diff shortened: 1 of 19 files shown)


beechtom:
Questions:
- Should `Bolt::Project.validate` and `Bolt::Config.validate` take a second pass at validating config? This would really only apply / be beneficial for testing, since a normal Bolt run will load config from files.
- Should we separately test basic validation for each config option, or are we comfortable with the coverage provided in unit and integration tests for the validator?
- Do we want to include validation of transport config in this first pass? If so, do we want to validate in files + the transport config classes, or just the transport config classes? The latter is simpler, but the former will catch errors earlier and give the file where the error is.

lucywyman:
* *Should Bolt::Project.validate and Bolt::Config.validate take a second pass at validating config?*
I can't think of any cases where we would need this outside of testing - CLI arguments should be validated by optparse, and I can't think of anything else that would change that would require revalidation? To skip to the question below, I think we can just test the Validator itself and not worry about structural config validation in tests. We obviously may still need to have tests around invalid values, but those wouldn't be caught by the validator anyhow.
* *Should we separately test basic validation for each config option, or are we comfortable with the coverage provided in unit and integration tests for the validator?* Just testing the validator seems fine
* *Do we want to include validation of transport config in this first pass? If so, do we want to validate in files + the transport config classes, or just the transport config classes?*
I think we probably need to validate once the config is resolved so that invalid config that gets overridden doesn't raise errors (which it seems like is the current behavior at least for config). We could maybe try to do something fancy later on that provides the file an error came from, but since we don't currently have that I think it can come later. It'd be great to include the transport config now, but if it's a lot more work we can get this in and do it separately :shrug: Whichever you prefer!

lib/bolt/config.rb
```diff
@@ -397,35 +416,11 @@ def validate
...
-              "#{compile_concurrency}"
-      end
-
       compile_limit = 2 * Etc.nprocessors
       unless compile_concurrency < compile_limit
         raise Bolt::ValidationError, "Compilation is CPU-intensive, set concurrency less than #{compile_limit}"
       end
 
-      unless %w[human json rainbow].include? format
-        raise Bolt::ValidationError, "Unsupported format: '#{format}'"
```
lucywyman, line 426: Previously this would validate CLI args as well as config options. It seems like most CLI args get caught either by optparse or in later validation, but at least this stacktraces:
```
lucy ~/githubs/bolt ((f8c19af...)) $ be bolt command run whoami -t docker --format 2
bundler: failed to load command: bolt (/home/lucy/githubs/bolt/vendor/bundle/ruby/2.6.0/bin/bolt)
NoMethodError: undefined method `print_head' for nil:NilClass
  /home/lucy/githubs/bolt/lib/bolt/cli.rb:512:in `execute'
  /home/lucy/githubs/bolt/exe/bolt:11:in `<top (required)>'
  /home/lucy/githubs/bolt/vendor/bundle/ruby/2.6.0/bin/bolt:23:in `load'
  /home/lucy/githubs/bolt/vendor/bundle/ruby/2.6.0/bin/bolt:23:in `<top (required)>'

lucy ~/githubs/bolt ((f8c19af...)) $ bolt command run whoami -t docker --format 2
Unsupported format: '2'
```
It seems like the easiest way to address that is to just check CLI options against the schema when we initialize config (I think right after calling  `normalize_overrides`) - that way we still get location for each schema but also validate the CLI.

lib/bolt/util.rb
```diff
@@ -295,6 +295,10 @@ def references?(input)
         end
       end
 
+      def reference?(input)
```
lucywyman, line 298: This seems like it will only be used in the validator class - I think you can bake it into the `plugin_reference?` method.

beechtom: Ah, I think I had meant for this to replace `Bolt::Plugin.reference?`, I just forgot to change the method calls in `Bolt::Plugin`. I'll just bake it into the validator's `plugin_reference?` method.

lucywyman (commented):
It seems like we want to also set `_plugins` to false for the `plugins` key since top-level keys there cannot be references themselves, and the `_plugins: false` isn't recursive on the data. And then can remove validation for that in plugins.rb

beechtom:
Ooo, good catch. That was definitely a bug in the schema before. 😬

lib/bolt/config/validator.rb
```diff
@@ -0,0 +1,179 @@
...
+          validate_value(item, definition[:items])
+        ensure
+          @path.pop
+        end
+      end
+
+      # Validates a string value, logging errors for any validations that fail.
+      #
+      private def validate_string(value, definition)
+        if definition.key?(:enum) && !definition[:enum].include?(value)
```
lucywyman, line 117: Hm...it seems like the transport option isn't getting picked up here and I can't figure out why...:
```
Unknown transport foo found for localhost:20022
```
That's coming from `lib/bolt/inventory/target.rb#154` I think

beechtom: Ah, `transport` isn't getting added into the schema. 😬 I'll fix that.

lib/bolt/config/validator.rb
```diff
@@ -0,0 +1,179 @@
...
+        case value
+        when Hash
+          validate_hash(value, definition)
+        when Array
+          validate_array(value, definition)
+        when String
+          validate_string(value, definition)
+        when Numeric
+          validate_number(value, definition)
+        end
```
lucywyman, line 66: This seems like it should have an else that errors, just in case none of these are matched....I don't think now we would ever reach the else, but it'd be hard to identify if anything in Psych ever changed or when upgrading Ruby and we could reach an else.

lib/bolt/config/validator.rb
beechtom: Oh, actually, the target-specific `transport` option is handled a bit differently, since the transport can be set in multiple locations and the transport config object only handles settings under `ssh`, `winrm`, etc. I think for now we can leave the validation here as-s: https://github.com/puppetlabs/bolt/blob/main/lib/bolt/inventory/target.rb#L153-L155

lucywyman: Ok, that seems fine - it only stuck out because it's different but I don't think it matters at all for the user experience.

lib/bolt/config/validator.rb
beechtom: We do reach the end if the value is a boolean. However, in that case it makes sense to fall through the case statement without a match since there's nothing to validate about booleans other than their type.

I don't think that adding an `else` adds any benefit, since any potential unknown type needs to be part of an option's types in order for validation to progress past `valid_type?`.

lucywyman (approved):
One last comment, then this lgtm!

Issue #2180, opened 2020-09-16
Allow easy way to show all project level content


npwalker:
## Use Case
When using the plan / task list in `bolt-project.yaml` I would like to add new content to the project and still have that content show up automatically in `bolt plan show` or `bolt task show`. 

From a project author perspective, I definitely want to curate a list of tasks/plans for my consumers but I do not want to be burdened by extra steps to show each new task/plan I make during a development cycle.  

## Describe the Solution You Would Like
My first thought is an option to provide globs then I'd just make my list 

```
plan:
 - project_name::*
task 
 - project_name::*
```

## Describe Alternatives You've Considered
I considered either or both of the following flags

show_builtin_content: false # to hide bolt built-in content but show all module or project content
show_project_content_only: true # to show only tasks/plans in the project but not modules or built-in content 

but they seem less useful while I assume adding more technical complexity.  

## Solution
* Should this be a wildcard or a glob? A glob solves for both issues, and the use case that inspired this ticket would require a glob, so...glob
* Does the glob match colons? So is `project_name::subdir::*` could include just top-level content and not anything "below" that. But it doesn't seem like most people would expect globbing to work this way. So yes, should match globs (basically `project_name::*` should match all subdirs)

npwalker:
this problem also applies to modules.  I spoke with a user today that wanted to restrict access to tasks/plans in certain modules and would like the ability to do allow access to all tasks/plans in a module OR to a set of task/plan names that match a pattern like `module_name::team_l1` or `module_name::team_l3`

PR #2350, opened 2020-11-13
(GH-2180) Filter project tasks and plans with glob patterns

beechtom:
The `plans` and `tasks` options in `bolt-project.yaml` now accept glob
patterns in addition to plan and task names. Plans and tasks that match
a glob pattern will be show in `bolt plan|task show` and
`Get-Bolt(Plan|Task)` output.

!feature

* **Filter project plans and tasks with glob patterns**
  ([#2180](https://github.com/puppetlabs/bolt/issues/2180))

  The `plans` and `tasks` options in `bolt-project.yaml` now support
  glob patterns in addition to plan and task names. Plans and tasks that
  match a glob pattern will appear in `bolt plan|task show` and
  `Get-Bolt(Plan|Task)` output.

### DIFF
--- a/documentation/projects.md
+++ b/documentation/projects.md
@@ -63,22 +63,24 @@ For a list of all the available project configuration options, see
 
 ### Limiting displayed plans and tasks
 
-Projects allow you to limit which plans and
-tasks a user can see when running `bolt plan show` or `bolt task show`. 
+Projects allow you to limit which plans and tasks a user can see when running
+`bolt plan|task show` and `Get-Bolt(Plan|Task)`. 
 
 Limiting tasks and plans is useful for the following reasons:
+
 - Bolt is bundled with several plans and tasks that might not be useful in your
   project. 
 - You might have written a task or plan that is only used by another task or
   plan, and you don't want your users to run that task or plan directly.
-- Displaying only specific content in the `show` commands makes it easier for
-  your users to find what they're looking for.
+- Displaying only specific content in the UI makes it easier for your users to
+  find what they're looking for.
 
-To control what plans and tasks appear when your users run `bolt plan show` or
-`bolt task show`, add `plans` and `tasks` keys to your `bolt-project.yaml` and
-include an array of plan and task names. For example, to surface a
-plan named `myproject::myplan`, and a task named `myproject::mytask`, you would
-use the following `bolt-project.yaml` file:
+To control which plans and tasks appear when your users show project content,
+add `plans` and `tasks` keys to your `bolt-project.yaml`. Both keys accept a list
+of names and glob patterns to filter content by.
+
+For example, to surface a plan named `myproject::myplan`, and a task named
+`myproject::mytask`, you would use the following `bolt-project.yaml` file:
 
 ```yaml
 name: myproject
@@ -87,8 +89,9 @@ plans:
 tasks:
 - myproject::mytask
 ```
-If your user runs the `bolt plan show` command, they'll get similar output to
-this:
+
+If your user runs the `bolt plan show` command or `Get-BoltPlan` PowerShell
+cmdlet, they'll get similar output to this:
 
 ```console
 $ bolt plan show
@@ -99,6 +102,33 @@ MODULEPATH:
 
 Use `bolt plan show <plan-name>` to view details and parameters for a specific plan.
 ```
+
+You can also use glob patterns to match multiple plan or task names. For
+example, to surface all tasks and plans in a project named `myproject`, you
+would use the following `bolt-project.yaml` file:
+
+```yaml
+name: myproject
+plans:
+- myproject::*
+tasks:
+- myproject::*
+```
+
+Glob patterns that begin with a metacharacter might cause problems when the
+configuration file is loaded and parsed, because the pattern might not be
+recognized as a string. To avoid this parsing issue, wrap any glob pattern that
+begins with a metacharacter in quotes. For example, you would write
+`"[abc]_module::*"` instead of `[abc]_module::*`.
+
+The following metacharacters can be used in a glob pattern:
+
+| Metacharacter | Description |
+| --- | --- |
+| `*` | Matches any number of characters. |
+| `?` | Matches any one character. |
+| `[set]` | Matches any one character in the set. |
+| `{a,b}` | Matches pattern a and pattern b. |
 
 ## Common files and directories in a project
 
--- a/lib/bolt/cli.rb
+++ b/lib/bolt/cli.rb
@@ -575,10 +575,15 @@ module Bolt
       outputter.print_task_info(pal.get_task(task_name))
     end
 
+    # Filters a list of content by matching substring.
+    #
+    private def filter_content(content, filter)
+      return content unless content && filter
+      content.select { |name,| name.include?(filter) }
+    end
+
     def list_tasks
-      tasks = pal.list_tasks
-      tasks.select! { |task| task.first.include?(options[:filter]) } if options[:filter]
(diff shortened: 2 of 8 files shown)


documentation/projects.md
```diff
@@ -100,6 +103,28 @@ MODULEPATH:
...
 ```
 
+You can also use glob patterns to match multiple plan or task names. For
+example, to surface all tasks and plans in a project named `myproject`, you
+would use the following `bolt-project.yaml` file:
+
+```yaml
+name: myproject
+plans:
+- myproject*
```
nicklewis, line 113: I think `myproject::*` is a clearer way to express this (and it avoids pulling in things from  a module named `myproject2`).

lib/bolt/cli.rb
```diff
@@ -575,10 +575,26 @@ def show_task(task_name)
       outputter.print_task_info(pal.get_task(task_name))
     end
 
+    # Filters a list of content by a filter and a list of patterns. The filter
+    # is a substring that matches the content name, whereas the patterns are
+    # glob patterns.
+    #
+    private def filter_content(content, filter, patterns)
+      if filter
+        content = content.select { |name,| name.include?(filter) }
```
nicklewis, line 584: Can we just turn this into a pattern as well by wrapping it in `*`? That would have the side effect of allowing users to pass a glob filter, which doesn't seem undesirable.

nicklewis: Ah I guess the problem is that `patterns` has to match _any_, but `filter` _must_ match.

nicklewis:
`bolt_server` has its own implementation of this filtering logic, to mark tasks/plans as "allowed". I'm not sure whether that should be updated with this PR or not.

nicklewis:
Maybe a clearer solution for this would be to give the PAL `list_tasks` and `list_plans` methods a boolean argument indicating whether they should return only the tasks/plans listed in the project config. Then the CLI filter would be an entirely separate concept. And bolt server would be able to take advantage of that argument as well, since we're already separately planning on making it _not_ return content that isn't listed in the project (rather than marking the content as `allowed` or not).

beechtom:
Updated to do the project config filtering in PAL.

documentation/projects.md
```diff
@@ -100,6 +103,28 @@ MODULEPATH:
...
+```
+
+The following metacharacters can be used in a glob pattern:
+
+| Metacharacter | Description |
+| --- | --- |
+| `*` | Matches any number of characters. |
+| `?` | Matches any one character. |
+| `[set]` | Matches any one character in the set. |
+| `\` | Escapes the next metacharacter. |
```
lucywyman, line 125: I don't think this is necessary since none of the metacharacters are valid parts of a modulename anyway.

lib/bolt/cli.rb
```diff
@@ -575,10 +575,15 @@ def show_task(task_name)
       outputter.print_task_info(pal.get_task(task_name))
     end
 
+    # Filters a list of content by matching substring.
+    #
+    private def filter_content(content, filter)
```
lucywyman, line 580: Why are we filtering content twice? It seems like just filtering it in PAL should work.

lib/bolt/pal.rb
```diff
@@ -286,15 +286,26 @@ def parse_manifest(code, filename)
       raise Bolt::PAL::PALError, "Failed to parse manifest: #{e}"
     end
 
-    def list_tasks
+    # Filters content by a list of names and glob patterns specified in project
+    # configuration.
+    def filter_content(content, patterns)
+      return content unless patterns
```
lucywyman, line 292: Or if content is empty :shrug:

lib/bolt/cli.rb
beechtom: It will get filtered twice regardless, though I don't really care _where_ that happens. Leaving the `--filter`... filtering in the CLI made sense per Nick's comment above about it being a separate concept entirely. 🤷

nicklewis: Restricting the content to only what belongs to the project is an API level operation. Picking a subset of that to view is a UI operation. So I think it's helpful to keep them separate.

lucywyman: Yeah sorry I didn't realize we had a CLI filter option.

lib/bolt/pal.rb
```diff
@@ -286,15 +286,26 @@ def parse_manifest(code, filename)
...
-        tasks = compiler.list_tasks
-        tasks.map(&:name).sort.each_with_object([]) do |task_name, data|
+        tasks = compiler.list_tasks.map(&:name).sort.each_with_object([]) do |task_name, data|
           task_sig = compiler.task_signature(task_name)
           unless task_sig.task_hash['metadata']['private']
             data << [task_name, task_sig.task_hash['metadata']['description']]
           end
         end
+
+        filter_content ? filter_content(tasks, @project&.tasks) : tasks
```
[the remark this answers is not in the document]
nicklewis: I think any _other_ kind of filtering ought to just be done by the caller. The caller could always use the `filter_content` method directly to get that behavior if they want it. But generic filtering seems out of the scope of _this_ method, otherwise all it's doing is taking a value to pass through to another method, which seems like an unnecessary inversion of control.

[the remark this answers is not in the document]
lucywyman: So, sorry, just to make sure I follow: PE needs to use this to filter tasks as well, but PE doesn't have projects right? I'm confused about this line: "And bolt server would be able to take advantage of that argument as well, since we're already separately planning on making it not return content that isn't listed in the project (rather than marking the content as allowed or not).". Are y'all using the same config settings as a blocklist instead of allowlist?

documentation/projects.md
```diff
@@ -100,6 +103,27 @@ MODULEPATH:
...
+You can also use glob patterns to match multiple plan or task names. For
+example, to surface all tasks and plans in a project named `myproject`, you
+would use the following `bolt-project.yaml` file:
+
+```yaml
+name: myproject
+plans:
+- myproject::*
+tasks:
+- myproject::*
```
lucywyman, line 115:
```suggestion
- "[mno]yproject::*"
```
Starting the regex with special characters can be a little tricky, but wrapping the phrase in a string should always work. Either stating that or having an example seems useful. This also applies for anything that starts with `*`.

lib/bolt/pal.rb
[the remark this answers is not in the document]
nicklewis: PE does have projects. Currently, requests to bolt-server to list tasks\* in a project include _all_ tasks and include a field `"allowed": true/false` for each item to indicate whether the task is on the list in `bolt-project.yaml`. We want to change it to _only_ return the things on the list, rather than listing everything with an `allowed` field. If that filtering logic is in PAL, then bolt-server can just  call `pal.list_tasks(filter_content: true)` rather than having to do the filter itself. So if we make a change to how the filtering works, as this PR does, bolt-server will automatically be able to honor those rules.

\* This all applies to plans as well but I didn't want to repeat tasks/plans everywhere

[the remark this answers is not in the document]
lucywyman: Ok that makes sense. I had thought y'all were adding a new way of filtering and didn't load project, which was why passing in the filter regex seemed useful.

This seems to do the right thing if `@project` is nil too, so if there are tools that try to use this without a project it should behave correctly.

documentation/projects.md
beechtom: Added a short paragraph on wrapping patterns that start with a metacharacter.

lucywyman (approved):

PR #2351, opened 2020-11-16
(maint) Don't use set-env to set env vars in Github Actions

lucywyman:
Github Actions environment [recently deprecated the use of set-env] in
their Windows environment due to a security vulnerability. We were using
this command to set the `BOLT_WINRM_PASSWORD` env var to a generated
password in our CI environment provisioning script. The script now
appends the env var and value to a file located at `$ENV:GITHUB_ENV`,
which Github Actions uses to set environment variables. This should set
the variable correctly without making us restructure how we provision
and run Github Actions.

This also fixes up some rubocop errors from the latest update

!no-release-note

### DIFF
--- a/scripts/ci.ps1
+++ b/scripts/ci.ps1
@@ -155,4 +155,4 @@ Enable-PSRemoting
 Set-WSManQuickConfig -Force
 Set-WinRMHostConfiguration
 Test-WinRMConfiguration @User | Out-Null
-Write-Output "::set-env name=BOLT_WINRM_PASSWORD::$pass"
+Add-Content -Path $ENV:GITHUB_ENV -Value "BOLT_WINRM_PASSWORD=$pass"
--- a/spec/fixtures/scripts/bolt_cli_loader.rb
+++ b/spec/fixtures/scripts/bolt_cli_loader.rb
@@ -7,13 +7,11 @@ require 'bolt'
 require 'bolt/cli'
 
 def suppress_outputs
-  # rubocop:disable Style/NegatedIfElseCondition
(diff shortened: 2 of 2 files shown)


scripts/ci.ps1
```diff
@@ -155,4 +155,4 @@ Enable-PSRemoting
 Set-WSManQuickConfig -Force
 Set-WinRMHostConfiguration
 Test-WinRMConfiguration @User | Out-Null
-Write-Output "::set-env name=BOLT_WINRM_PASSWORD::$pass"
\ No newline at end of file
+$ENV:GITHUB_ENV=@{ BOLT_WINRM_PASSWORD = $pass}
```
nicklewis, line 158: I think the value of `$ENV:GITHUB_ENV` is a file that we should append our environment variable assignments to, not a variable that we should assign.

scripts/ci.ps1
```diff
@@ -155,4 +155,4 @@ Enable-PSRemoting
 Set-WSManQuickConfig -Force
 Set-WinRMHostConfiguration
 Test-WinRMConfiguration @User | Out-Null
-Write-Output "::set-env name=BOLT_WINRM_PASSWORD::$pass"
\ No newline at end of file
+$ENV:GITHUB_ENV="BOLT_WINRM_PASSWORD=$pass"
```
beechtom, line 158:
```suggestion
Write-Output "BOLT_WINRM_PASSWORD=$pass" >> $env:GITHUB_ENV
```

beechtom (approved):

Issue #1650, opened 2020-03-04
Error with bolt when running Concat module against windows machines


nheaney18:
The IAC team have recently been seeing an error appear whenever we apply certain manifests on a windows machine via Bolt. Originally we felt that this may be originating from a puppet agent change however we now feel that it is likely originating from bolt, more information on the investigations into Puppet Agent can be found in [https://tickets.puppetlabs.com/browse/PA-3065|https://cdn1.evernote.com/ion/dist/ce/].
 
The IAC Team have made multiple applications of a select manifest against the windows machines using bolt via litmus, bolt directly and through logging onto the machine and applying it manually. When applying the manifest via bolt in either way the result is an exit code of 1, while applying it manually gives an exit code of 0. I have attached screenshots of the results of each run onto this
 ticket and will paste the output below:
  
 When run locally
```
Administrator@wooden-debacle /cygdrive/c/tmp
$ '../Program Files/Puppet Labs/Puppet/bin/puppet.bat' apply /tmp/manifest_20200129_56833_19lr8ku.pp --detailed-exitcodes
Notice: Compiled catalog for wooden-debacle.delivery.puppetlabs.net in environment production in 0.11 seconds
Error: /Stage[main]/Main/Concat[c:/concat_test/file]/Concat_file[c:/concat_test/file]: Failed to generate additional resources using 'eval_generate': Duplicate key 'one' found with values 'foo' and bar'. Use 'force' attribute to merge keys.
Notice: Applied catalog in 0.02 secondsAdministrator@wooden-debacle /cygdrive/c/tmp
$  echo $?
0
```
When run via litmus
```
  1) force merge of file when run should not force applies manifest twice with stderr check
     On host `wooden-debacle.delivery.puppetlabs.net'
     Failure/Error: expect(apply_manifest(pp, catch_failures: true).stderr).to match("Duplicate key 'one' found with values 'foo' and bar'. Use 'force' attribute to merge keys.")
     RuntimeError:
       apply manifest failed
       ` puppet apply /tmp/manifest_20200129_56833_cptso6.pp --detailed-exitcodes`
       with exit code 1 (expected: [0, 2])
       ====== Start output of failed Puppet apply ======
       puppet.bat : Error: /Stage[main]/Main/Concat[c:/concat_test/file]/Concat_file[c:/concat_test/file]: Failed to generate additional resources using 'eval_generate': Duplicate key 'one' found with values 'foo' and bar'. Use 'force' attribute to merge keys.
[...]

PR #2354, opened 2020-11-17
(GH-2299) Write Windows local transport commands to file

### LINKED ISSUE #2299: PowerShell commands with pipes return no output
## Describe the Bug

If you attempt to use any PowerShell command and pipe the results to another, the results are lost

## Expected Behavior

Bolt should return a string with the results of the next piped command.

## Steps to Reproduce

```
❯ bolt command run "Get-WMIObject Win32_Service -Filter ""Name like '%rm'"" | select name,servicename" -t localhost
Started on localhost...
Finished on localhost:
Ran on 1 target in 1.0 sec
```

## Environment
 - Version 0.23.X
 - Platform Windows

## Additional Context

If we use a Format-* cmdlet instead of relying on Out-Default, we actually get a return:

```
❯ bolt command run "Get-WMIObject Win32_Service -Filter ""Name like '%rm'"" | select name,servicename | fl" -t localhost
Started on localhost...
Finished on localhost:
  STDOUT:
    name        : WinRM
    servicename :
Ran on 1 target in 1.0 sec
```

This is an artifact of how we are detecting the `$LASTEXITCODE`. I'll open a separate bug for that.

jpogran:
This modifies the PowerShell shell to write commands to a file for 
execution when running over the local transport. This uses the same
logic we use when running commands greater than the max command length
to wrap the command in a script. This enables two necessary behaviors:
returning correct exit codes and allowing users to run commands with
pipes.

PowerShell will always exit 0 or 1 when running commands directly, even
when the command returns a different exit code (for example, running
`puppet agent --detailed-exitcodes`). However when executing a file with
`-File`, PowerShell will return the exit code returned by the script
[source](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_powershell_exe?view=powershell-5.1#-command).
Wrapping the command in a script also works around syntactic weirdness
(waves hands) that prevents users from easily running commands with
pipes.

Closes #2299

!bug

* **Windows local transport returns correct exit codes and accepts pipes** ([#2299](https://github.com/puppetlabs/bolt/issues/2299))

  When running commands over the local transport on Windows machines,
  Bolt now returns the exit code returned by the command as opposed to
  just 0 or 1. It also accepts pipes as part of the command.

### DIFF
--- a/lib/bolt/shell/powershell.rb
+++ b/lib/bolt/shell/powershell.rb
@@ -177,7 +177,8 @@ module Bolt
       def run_command(command, options = {}, position = [])
         command = [*env_declarations(options[:env_vars]), command].join("\r\n") if options[:env_vars]
 
-        output = execute(command)
+        wrap_command = conn.is_a?(Bolt::Transport::Local::Connection)
+        output = execute(command, wrap_command)
         Bolt::Result.for_command(target,
                                  output.stdout.string,
                                  output.stderr.string,
@@ -268,8 +269,9 @@ module Bolt
         end
       end
 
-      def execute(command)
-        if conn.max_command_length && command.length > conn.max_command_length
+      def execute(command, wrap_command = false)
+        if (conn.max_command_length && command.length > conn.max_command_length) ||
+           wrap_command
           return with_tmpdir do |dir|
             command += "\r\nif (!$?) { if($LASTEXITCODE) { exit $LASTEXITCODE } else { exit 1 } }"
             script_file = File.join(dir, "#{SecureRandom.uuid}_wrapper.ps1")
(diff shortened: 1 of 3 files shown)


lucywyman:
I verified this returns the correct exit code running `bolt command run --targets localhost 'puppet apply --trace --detailed-exitcodes -e "notify { foo: }"'` (#1846), and `bolt command run "Get-WMIObject Win32_Service -Filter ""Name like '%rm'"" | select name,servicename" -t localhost` (#2299).

I verified this at least works with Litmus at least for the current puppetlabs-concat module test suite [see passing tests here](https://github.com/puppetlabs/puppetlabs-concat/pull/671), but I'm not sure that's strong enough to prove #1650 is resolved. Open to more testing scenarios!

beechtom (approved):

Issue #2355, opened 2020-11-17
Disallowed class in YAML files stacktraces


lucywyman:
## Describe the Bug
When Psych raises a DisallowedClass error, such as if one of the values in config is a symbol, the error does not have a `line` method which causes a stacktrace. We should only include the error location if the error has the location.

bolt.yaml
```
color: :true
```

```
lucy ~/githubs/bolt ((4a98cad...)) $ bolt command run whoami -t docker
Traceback (most recent call last):
	9: from /opt/puppetlabs/bolt/bin/bolt:23:in `<main>'
	8: from /opt/puppetlabs/bolt/bin/bolt:23:in `load'
	7: from /opt/puppetlabs/bolt/lib/ruby/gems/2.5.0/gems/bolt-2.35.0/exe/bolt:10:in `<top (required)>'
	6: from /opt/puppetlabs/bolt/lib/ruby/gems/2.5.0/gems/bolt-2.35.0/lib/bolt/cli.rb:95:in `parse'
	5: from /opt/puppetlabs/bolt/lib/ruby/gems/2.5.0/gems/bolt-2.35.0/lib/bolt/cli.rb:175:in `load_config'
	4: from /opt/puppetlabs/bolt/lib/ruby/gems/2.5.0/gems/bolt-2.35.0/lib/bolt/project.rb:41:in `find_boltdir'
	3: from /opt/puppetlabs/bolt/lib/ruby/gems/2.5.0/gems/bolt-2.35.0/lib/bolt/project.rb:77:in `create_project'
	2: from /opt/puppetlabs/bolt/lib/ruby/gems/2.5.0/gems/bolt-2.35.0/lib/bolt/util.rb:48:in `read_optional_yaml_hash'
	1: from /opt/puppetlabs/bolt/lib/ruby/gems/2.5.0/gems/bolt-2.35.0/lib/bolt/util.rb:25:in `read_yaml_hash'
/opt/puppetlabs/bolt/lib/ruby/gems/2.5.0/gems/bolt-2.35.0/lib/bolt/util.rb:41:in `rescue in read_yaml_hash': undefined method `line' for #<Psych::DisallowedClass:0x000000000174ca10> (NoMethodError)
```

## Environment
 - Version [e.g. 1.27.0]
 - Platform [e.g. Ubuntu 18.04]

PR #2360, opened 2020-11-17
(GH-2355) Do not error when Psych errors do not have a line number

beechtom:
This updates the `Bolt::Util.read_yaml_hash` method to not error with a
stacktrace when handling Psych errors that do not include a line number.
Previously, if a Psych error did not include a line number, such as the
`Psych::DisallowedClass` error, Bolt would raise an undefined method
error with a stacktrace.

!no-release-note


lucywyman:
I kind of don't think this needs a release note

Issue #2338, opened 2020-11-11
Ensure Bolt is compatible with Puppet agent 7


lucywyman:
## Use Case
The Puppet 7 release is coming up, and it'd be great if we could ensure Bolt is usable with Puppet 7 when it's released or shortly after. This includes:
- [x] Test what gem versions are compatible with bolt. Update the maximum Puppet versions in our gemfile if possible. 
- [x] Validate how the release will impact `apply_prep` and `apply`? Will puppet 7 be installed and work seamlessly? Should we pin the default puppet_library to use puppet 6 for now? What about macos and windows?
  *To answer this question a little bit, Puppet 7 will be installed by the `puppet_agent::install` task by default if we do nothing.*
  *It sounds like there aren't any breaking Puppet language changes. The main changes I would be worried about are the following*:
    - Apt provider  ensure: held  (replaced by  mark: hold DirectoryService#write_to_file) (PUP-10597)
    - NameService.listbyname (replaced by NameService.instances) (PUP-10488)
    - Puppet::Util::Execution.execpipe method (deprecated) (PUP-7584)
    - Puppet::Resource.set_default_parameters (PUP-10474)
    - Puppet::Resource.validate_complete (PUP-10474)
    - Puppet::Resource::Type.assign_parameter_values (PUP-10474)

Do we also want to do an hour of exploratory testing off the main Puppet branch to anticipate any other changes to add to the 3.0 doc? It seems like a good idea, but we can wait until closer to releasing 3.0 to do it.

lucywyman:
Plan is to leave the `puppet_agent::install` to pull in latest (Puppet 7), but wait to update the Puppet we ship with Bolt until Bolt 3.0.

PR #2366, opened 2020-11-17
(GH-2338) Rename 'sensitive' plan in tests

lucywyman:
### Rename 'sensitive' plan in tests
'Sensitive' is a reserved word in Puppet 7 and cannot be used as a plan
name. We have a plan named 'sensitive' used for testing the `run_plan`
plan function. This renames the test fixture to 'sensitive_test'.

### Remove maximum version for Puppet gem
We will continue to ship Bolt with the puppet 6 gem, but we shouldn't
need to limit the maximum version of the Puppet gem for now. It seems
like Bolt is compatible with Puppet 7, and we don't have any breaking
changes going in to Puppet that we need to gate against. This also means
tools that use Bolt as a dependency won't be limited to using Puppet 6.x.

!no-release-note

### DIFF
--- a/bolt-modules/boltlib/spec/fixtures/modules/sensitive_test/plans/complex.pp
+++ b/bolt-modules/boltlib/spec/fixtures/modules/sensitive_test/plans/complex.pp
@@ -1,4 +1,4 @@
-plan sensitive::complex (
+plan sensitive_test::complex (
   Variant[Sensitive[String], Array[String]] $complex
 ) {
 }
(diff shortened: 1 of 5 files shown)


beechtom (approved):

scotje:
@lucywyman Did you consider just bumping the upper bound to for the Puppet dependency to `< 8.0.0`? Theoretically Bolt 2.36.0 now claims it will be compatible with all future release of Puppet. Bundler and Rubygems generally don't seem to do much backtracking to resolve dependencies so it's probably not a big deal as long as an upper bound gets added back in at some point if an incompatibility arises.

PR #2369, opened 2020-11-18
(PE-30464) Change 'GET' => 'POST' for project_inventory_targets

ekinanp:
This endpoint will eventually accept a request body. Since it's
unidiomatic for 'GET' endpoints to accept a request body, we change the
project_inventory_targets to a 'POST' endpoint instead.


ekinanp:
I noticed that we needed to make this change while writing the secrets management tickets. Figured it's best to get it out of the way now so we can get it in the initial sprint than defer it to later since it is a simple change.

donoghuc (approved):

PR #2371, opened 2020-11-18
(GH-2302) Send analytics about how many plan types exist in each project

### LINKED ISSUE #2302: Track project composition analytics
We'd like to know how many yaml plans and puppet plans users have in their project. The data needs to be a single blob of yaml plans and puppet plans together so we can correlate the data.
This is to answer:
Are yaml plans and puppet plans used together? Is one more common than the other?

lucywyman:
This adds two new GA dimensions to the data we send for user screenviews
when Bolt is run on the CLI: puppet plan count and yaml plan count.
These dimensions simply count the number of files in the project's
`plans` directory (including subdirs) to provide analytics for how many
project-level plans in each language the user has. The dimension is a
user-level metric, meaning that whenever the values for a particular
user change they will get updated in GA and the old value will be
removed. This poses a problem for users with multiple projects, as each
time they run Bolt from a different project their values will get
updated. However the tradeoff is that we get per-project values as
opposed to aggregated values, and we don't get multiple values when a
user adds a new plan.

Closes #2302

!no-release-note

### DIFF
--- a/lib/bolt/analytics.rb
+++ b/lib/bolt/analytics.rb
@@ -2,6 +2,7 @@
 
 require 'bolt/util'
 require 'bolt/version'
+require 'find'
 require 'json'
 require 'logging'
 require 'securerandom'
@@ -23,14 +24,16 @@ module Bolt
       plan_steps: :cd8,
       return_type: :cd9,
       inventory_version: :cd10,
-      boltdir_type: :cd11
+      boltdir_type: :cd11,
+      puppet_plan_count: :cd12,
+      yaml_plan_count: :cd13
     }.freeze
 
     def self.build_client
       logger = Bolt::Logger.logger(self)
       begin
-        config_file = config_path(logger)
-        config = load_config(config_file, logger)
+        config_file = config_path
+        config = load_config(config_file)
       rescue ArgumentError
         config = { 'disabled' => true }
       end
@@ -51,7 +54,7 @@ module Bolt
       NoopClient.new
     end
 
-    def self.config_path(logger)
+    def self.config_path
       path     = File.expand_path(File.join('~', '.puppetlabs', 'etc', 'bolt', 'analytics.yaml'))
       old_path = File.expand_path(File.join('~', '.puppetlabs', 'bolt', 'analytics.yaml'))
 
@@ -59,7 +62,7 @@ module Bolt
         if File.exist?(old_path)
           message = "Detected analytics configuration files at '#{old_path}' and '#{path}'. Loading "\
                     "analytics configuration from '#{path}'."
-          logger.warn(message)
+          Bolt::Logger.warn_once('duplicate_analytics', message)
         end
 
         path
@@ -70,12 +73,12 @@ module Bolt
       end
     end
 
-    def self.load_config(filename, logger)
+    def self.load_config(filename)
       if File.exist?(filename)
         Bolt::Util.read_optional_yaml_hash(filename, 'analytics')
       else
         unless ENV['BOLT_DISABLE_ANALYTICS']
-          logger.warn <<~ANALYTICS
+          Bolt::Logger.warn_once('analytics_opt_out', <<~ANALYTICS)
             Bolt collects data about how you use it. You can opt out of providing this data.
 
             To disable analytics data collection, add this line to ~/.puppetlabs/etc/bolt/analytics.yaml :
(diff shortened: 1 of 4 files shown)


lib/bolt/analytics.rb
```diff
@@ -139,6 +141,15 @@ def report_bundled_content(mode, name)
         end
       end
 
+      def plan_counts(project_path)
```
lucywyman, line 144: This function can kind of go anywhere. I opted to put it here since we're using it for analytics and putting it here indicates that, not because it relies on anything in this class.

lib/bolt/analytics.rb
```diff
@@ -51,15 +53,15 @@ def self.build_client
...
+    def self.config_path
       path     = File.expand_path(File.join('~', '.puppetlabs', 'etc', 'bolt', 'analytics.yaml'))
       old_path = File.expand_path(File.join('~', '.puppetlabs', 'bolt', 'analytics.yaml'))
 
       if File.exist?(path)
         if File.exist?(old_path)
           message = "Detected analytics configuration files at '#{old_path}' and '#{path}'. Loading "\
                     "analytics configuration from '#{path}'."
-          logger.warn(message)
+          Bolt::Logger.warn_once('duplicate_analytics', message)
```
lucywyman, line 64: This change was originally because I was re-loading the configfile to write to it, which resulted in double warnings. The final solution doesn't do that, but I think these changes are still more in line with modern Bolt code so I left them.

lib/bolt/analytics.rb
```diff
@@ -139,6 +141,15 @@ def report_bundled_content(mode, name)
         end
       end
 
+      def plan_counts(project_path)
+        pp_count, yaml_count = %w[pp yaml].map do |extension|
+          Find.find(File.join(project_path, 'plans')).grep(/.*\.#{extension}/).length
```
nicklewis, line 146: This fails if there's no `plans` directory in the project. We should probably both accommodate that and wrap this in an error handler that returns `{}` so analytics don't cause a crash.

lib/bolt/cli.rb
```diff
@@ -392,7 +392,7 @@ def execute(options)
         output_format: config.format,
         # For continuity
         boltdir_type: config.project.type
-      }
+      }.merge!(analytics.plan_counts(config.project.path))
```
beechtom, line 395:
```suggestion
      }.merge!(analytics.plan_counts(config.project.plans_path))
```

lib/bolt/analytics.rb
beechtom, line 146:
```suggestion
      def plan_counts(plans_path)
        pp_count, yaml_count = %w[pp yaml].map do |extension|
          Find.find(plans_path).grep(/.*\.#{extension}/).length
```

lib/bolt/analytics.rb
```diff
@@ -139,6 +141,15 @@ def report_bundled_content(mode, name)
...
+        end
+
+        # Hash.new(%w[puppet_plan_count yaml_plan_count].zip(pp_count, yaml_count))
```
beechtom, line 149:
```suggestion
```

I'm guessing this is leftover?

lib/bolt/cli.rb
lucywyman: I don't remember why I didn't do this initially but I know there was a reason....anyway, maybe it wasn't important

lib/bolt/analytics.rb
lucywyman: Mostly to deter comments on how to calculate it, since it's more characters and operations :P I'll remove!

lib/bolt/analytics.rb
lucywyman: Hmm how does it fail? I don't seem to be able to reproduce that...
```
lucy ~/githubs/bolt (GH-2302) $ ls | grep 'plan'
lucy ~/githubs/bolt (GH-2302) $ be bolt command run whoami -t localhost
Started on localhost...
Finished on localhost:
  STDOUT:
    lucy
Ran on 1 target in 0.01 sec
lucy ~/githubs/bolt (GH-2302) $ ls | grep 'bolt-project.yaml'
bolt-project.yaml
```

nicklewis: `bundle exec` disables analytics by default, is that why it's not failing for you?

lib/bolt/analytics.rb
```diff
@@ -139,6 +141,18 @@ def report_bundled_content(mode, name)
         end
       end
 
+      def plan_counts(plans_path)
+        pp_count, yaml_count = if File.exist?(plans_path)
+                                 %w[pp yaml].map do |extension|
+                                   Find.find(plans_path.to_s).grep(/.*\.#{extension}/).length
```
beechtom, line 147: The `Find` class is only available because we require the module in the applicator and orch transport. I think we should require it in this file as well just in case that changes in the future.

beechtom (approved):
Nitpick, but otherwise lgtm

PR #2380, opened 2020-11-19
(GH-2362) Deprecate boltdir, configfile, puppetfile, and description CLI flags

### LINKED ISSUE #2362: Deprecate `--boltdir`, `--configfile`, `--puppetfile`, and `--description` cli options
## Use Case
Print a deprecation warning when these CLI flags are used. Make sure to use the logger method so we get analytics.

We should also mark them as deprecated in the help text and docs.

lucywyman:
This adds deprecation warnings for the `--boltdir`, `--configfile`,
`--puppetfile`, and `--description` command line flags. Whenever users
use any of these command line options - and only if the option is
specified on the CLI - a warning should be printed. This doesn't update
spec tests or other internal references to deprecated options, but it
does update the help text to not include deprecated flags.

Closes #2362

!deprecation

* **Add deprecation warnings for boltdir, configfile, puppetfile, and description** ([#2362](https://github.com/puppetlabs/bolt/issues/2362))

  We are planning to remove the `--boltdir`, `--configfile`,
  `--puppetfile`, and `--description` command line flags in the next major
  version of Bolt. This adds deprecation warnings that are printed when
  users specify any of these flags.

### DIFF
--- a/lib/bolt/bolt_option_parser.rb
+++ b/lib/bolt/bolt_option_parser.rb
@@ -6,11 +6,12 @@ require 'optparse'
 
 module Bolt
   class BoltOptionParser < OptionParser
+    PROJECT_PATHS = %w[project configfile boltdir].freeze
     OPTIONS = { inventory: %w[targets query rerun description],
                 authentication: %w[user password password-prompt private-key host-key-check ssl ssl-verify],
                 escalation: %w[run-as sudo-password sudo-password-prompt sudo-executable],
                 run_context: %w[concurrency inventoryfile save-rerun cleanup],
-                global_config_setters: %w[modulepath project configfile],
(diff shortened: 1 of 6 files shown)


lib/bolt/bolt_option_parser.rb
```diff
@@ -794,6 +796,9 @@ def initialize(options)
       end
       define('--description DESCRIPTION',
              'Description to use for the job') do |description|
```
beechtom, line 798:
```suggestion
             'Deprecated. Description to use for the job.') do |description|
```

beechtom (changes requested):
The release note tag for this should be !deprecation

lucywyman:
Nice, updated the commit and comment!

beechtom (approved):

Issue #2344, opened 2020-11-12
Bolt error caused by a Facter warning on Windows


carabasdaniel:
## Describe the Bug

Issue seems to be caused by this warning returned from puppetlabs-facts task call `"_error"=>{"msg"=>"C:/Program Files/Puppet Labs/Puppet/puppet/bin/libfacter.so: warning: already initialized constant Facter::FACTERVERSION\n", "kind"=>"bolt/error", "details"=>{}}}}`

For more details on the failures see logs at: 
https://github.com/puppetlabs/puppetlabs-iis/runs/1390513520?check_suite_focus=true#step:10:37

## Expected Behavior
Return the correct facts without error. 

## Steps to Reproduce
1. Workflow can be re-triggered on: https://github.com/puppetlabs/puppetlabs-iis/actions/runs/359686201

Gemfile pointed to puppetlabs/bolt main branch in PR: https://github.com/puppetlabs/puppetlabs-iis/pull/295

## Environment
Details in log.

beechtom:
@carabasdaniel This seems to be caused by a bug in Facter 3 on Windows. The `facts` task runs the following command for Facter 3:

```
/path/to/facter.exe -p --json --show-legacy
```

When resolving Puppet facts, Facter is unable to load a Ruby library, resulting in the following warning being issued:

```
    facter.exe : 2020-11-19 20:22:30.253491 WARN  puppetlabs.facter - could not locate a ruby library: facts requiring Ruby will not be resolved.
        + CategoryInfo          : NotSpecified: (2020-11-19 20:2...ot be resolved.:String) [], RemoteException
        + FullyQualifiedErrorId : NativeCommandError
```

And then Facter is terminated by a segmentation violation (11) signal, which results in a non-zero exit code. The `facts` task returns an error object whenever Facter returns a non-zero exit code, resulting in the failure that you're seeing.

We're adding this is a known issue to Bolt's documentation and have added a comment regarding this bug in a ticket opened a while back with a similar looking issue: https://tickets.puppetlabs.com/browse/FACT-1349

carabasdaniel:
Thanks @beechtom for doing the investigation on this. I'll reach out to the facter team about this so you can close this issue.

PR #2381, opened 2020-11-19
(GH-2344) Document Facter 3 failure on Windows as known issue

beechtom:
This adds a known issue where the `facts` task fails when running Facter
3 on Windows targets. The failure is caused by a bug in Facter 3 that
results in a sigsegv (11) signal when attempting to resolve Puppet
facts.

!no-release-note

### DIFF
--- a/documentation/bolt_known_issues.md
+++ b/documentation/bolt_known_issues.md
@@ -1,5 +1,33 @@
 # Known issues
 
+## `facts` task fails on Windows targets with Facter 3 installed
+
+When running the `facts` task on a Windows target that has Facter 3 installed,
+the task will fail but still return facts for the target. Output may look
+similar to the following:
+
+```shell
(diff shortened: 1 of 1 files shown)


hestonhoffman (approved):
LGTM!

documentation/bolt_known_issues.md
```diff
@@ -1,5 +1,33 @@
 # Known issues
 
+## `facts` task fails on Windows targets with Facter 3 installed
+
+When running the `facts` task on a Windows target that has Facter 3 installed,
```
lucywyman, line 5: Does this happen on all Windows versions consistently?

beechtom: I tried with 2012 r2, 2016, and 2019 images and it showed on all 3.

PR #2383, opened 2020-11-19
(GH-2313) Add plugin caching

### LINKED ISSUE #2313: Cache plugin for dynamic inventory.yaml
## Use Case
We're using a inventory.yaml which is fully dynamic (based on PuppetDB queries) and creates based on the dynamic output Bolt groups. As we have 3 external connection types (LDAP, IPA and external IPA) we have 3 PuppetDB queries to encapsulate our system landscape. As the PuppetDB is used we're always complete with our inventory.yaml, the current downside of this setup is that we have a huge delay before the Bolt command runs, as it needs to query and compile the data.

To minimize the wait time it would be greate to have a cache plugin with a TTL on it, so reducing the need at each Bolt command to run the queries and the delay which it causes. In the Bolt slack channel the following example was suggested:
```yaml
- targets:
    _plugin: cache
    ttl: 600m
    cache: cache/inventory.yaml
    source: source/puppetdb.yaml
```

As alternative I was thinking about to have an option to export the dynamic output to a static inventory.yaml to be used. This static file would be used for execution by Bolt and on the background run a task to update the static file when necessary. But this is a complicated method of creating a cache.

## Additional Context
n/a

lucywyman:
This adds some basic caching for Bolt plugins. Caching is enabled by
configuring `cache`, either as a default cache for all plugins in the
`plugin-cache` config option, or as configuration on the plugin
itself with the `_cache` key. When the plugin reference is evaluated, if
caching is enabled Bolt will check the cache first to see if the plugin
result has been cached and is fresh.

Bolt stores cached results in the `.plugin_cache.json` file of the
bolt project and assigning each plugin result a unique ID by hashing a
stringified, sorted version of the hash. This means that if users have
the exact same plugin with the same configuration in multiple places in
the inventory they can share the same cache files. Bolt will check the
cache and see if cache is past it's ttl. If it's not, Bolt will first
clear out all outdated cache results from the cache file before
returning the cached result for this plugin. This ensures that orphaned
cache entries (results whose plugins have been changed or removed) don't
stay forever. Bolt is able to clear out orphaned results by storing the
original ttl alongside the results in the cache file. This does have the
side effect that a plugin may be cleared even if it's ttl is updated, if
it has already reached it's previous ttl.

If the cache is past it's ttl or if no cache has been configured Bolt
will evaluate the plugin as normal and cache the result. It stores the
current ttl with the result so the results can be cleaned out.

This also provides `--clear-cache` and `-ClearCache` which
clears all cached results.

There are a lot of other features that would be interesting to add:
- Encrypted caching
- Refreshing only certain plugins

Closes #2313

!feature

* **Plugin caching**

  **This feature is experimental**. Bolt plugins can now be configured
  to cache their results. Users can either configure a default cache
  time-to-live for all plugins, or configure each plugins ttl
  individually. See https://pup.pt/bolt-cache for more information


lucywyman:
I may add more testing as needed, but I think this is ready for review.

lib/bolt/config/options.rb
```diff
@@ -172,6 +183,18 @@ module Options
...
+          description: "This feature is experimental. Enable plugin caching and set the time-to-live.",
+          type: Hash,
+          _plugin: false,
+          _example: { "ttl" => 3600 },
+          properties: {
+            "ttl" => {
+              description: "Time in seconds to keep the plugin cache.",
+              type: Integer
+            }
+          }
```
beechtom, line 196:
```suggestion
          required: ["ttl"],
          properties: {
            "ttl" => {
              description: "Time in seconds to keep the plugin cache.",
              type: Integer
            }
          },
          _plugin: false,
          _example: { "ttl" => 3600 }
```

Make sure `ttl` is required & keeping key order consistent between options.

lib/bolt/config/options.rb
```diff
@@ -494,6 +517,7 @@ module Options
         inventoryfile
         log
         modulepath
+        plugin-cache
```
beechtom, line 520:
```suggestion
```

Since this file is deprecated we shouldn't enable new features for it.

lib/bolt/project.rb
```diff
@@ -116,6 +116,7 @@ def initialize(raw_data, path, type = 'option', logs = [])
       @plans_path        = @path + 'plans'
       @managed_moduledir = @path + '.modules'
       @backup_dir        = @path + '.bolt-bak'
+      @cache_file        = @path + '.cache' + 'plugin_cache.json'
```
beechtom, line 119:
```suggestion
      @cache_file        = @path + '.plugin_cache.json'
```

Since the cache is just a single file in this iteration, I think it shouldn't be saved in a separate directory, similar to .rerun.json.

lib/bolt/config.rb
```diff
@@ -459,6 +459,10 @@ def validate
         Bolt::Util.validate_file('inventory file', default_inventoryfile)
       end
 
+      if @data['plugin-cache'] && @data['plugin-cache']['ttl'] < 0
+        raise Bolt::ValidationError, "plugin-cache.ttl must be a positive integer"
+      end
```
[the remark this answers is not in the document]
lucywyman: I think it might, like configuring a directory or file to store the cache in. It gives us the most flexibility if it takes a hash for now.

lib/bolt/plugin.rb
```diff
@@ -288,16 +313,55 @@ def resolve_single_reference(reference)
...
         # Evaluate the plugin and then recursively evaluate any plugin returned by it.
         hook.call(reference)
       rescue StandardError => e
         loc = "resolve_reference in #{plugin_name}"
         raise PluginError::ExecutionError.new(e.message, plugin_name, loc)
       end
+
+      if cache?(reference)
+        cache = read_cache
+        cache.merge!({ id => { 'result' => result,
```
beechtom, line 326:
```suggestion
        cache.merge!({ id => { 'result' => result,
```

This saves the result in plaintext, which will expose any sensitive information returned by the plugin. Is that something we're concerned about?

lib/bolt/bolt_option_parser.rb
```diff
@@ -54,6 +54,15 @@ def get_help_text(subcommand, action = nil)
           { flags: OPTIONS[:global],
             banner: INVENTORY_HELP }
         end
+      when 'cache'
```
beechtom, line 57: This command needs to be added to the default banner text (`bolt --help`): https://github.com/puppetlabs/bolt/blob/main/lib/bolt/bolt_option_parser.rb#L187

lib/bolt/bolt_option_parser.rb
```diff
@@ -54,6 +54,15 @@ def get_help_text(subcommand, action = nil)
...
+        case action
+        when 'refresh'
```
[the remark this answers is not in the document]
lucywyman: Hm, I think that makes sense. We can always add a separate command afterwards if people ask for it.

lib/bolt/project.rb
lucywyman: Hm, I initially made this a directory assuming we may want to cache other things in the future. I still think that might be the case, but I'm not sure what other cached elements we'll add short term (maybe commonly used modules so that using Puppet is faster?). Having this be a file for now seems fine and we can make it a directory later since users should interact with it directly.

lib/bolt/config/options.rb
lucywyman: I guess this just seemed like a nice thing to do for people who hadn't upgraded yet, since it would handle `project migrate`, but I think you're right.

lib/bolt/plugin.rb
lucywyman: Yes, I have a note in the commit message to encrypt the cache as a next step. We can do that as part of this iteration if it seems unusable without it, but I figured this was a good mvp.

beechtom: Ah, missed that part, my bad. 😅  I think having it as a next step is okay since it's experimental + opt-in.

lucywyman: Np np - it definitely *is* unusable for some users without that.

documentation/experimental_features.md
```diff
@@ -8,6 +8,49 @@ API may change, requiring the user to update their code or configuration. The
...
+project, and all expired cache entries will be removed whenever Bolt uses a cache result. This
+prevents cache entries from getting orphaned and never removed. Bolt identifies a cache result based
+on a hash (delightfully named 'bubblebabble') of the entire plugin invocation, minus the `_cache`
+key. If any element of the plugin invocation changes, Bolt will reload the plugin and update the
+cache.
+
+**NOTE**: Plugin results are currently cached in plaintext. Encrypted caching is planned for a
+future release.
+
+Users can clear all cache entries using the `--clear-cache` or `-ClearCache` with any Bolt command.
```
beechtom, line 35:
```suggestion
Users can clear all cache entries using the `--clear-cache` or `-ClearCache` command-line
option with any Bolt command.
```

lib/bolt/cli.rb
```diff
@@ -45,6 +45,7 @@ class CLI
       'project'    => %w[init migrate],
       'module'     => %w[add generate-types install show],
       'apply'      => %w[],
+      'cache'      => %w[refresh],
```
beechtom, line 48:
```suggestion
```

lib/bolt/cli.rb
```diff
@@ -197,6 +198,7 @@ def finalize_setup
       @parser_deprecations.each { |dep| Bolt::Logger.deprecation_warning(dep[:type], dep[:msg]) }
       config.deprecations.each { |dep| Bolt::Logger.deprecation_warning(dep[:type], dep[:msg]) }
 
+      FileUtils.rm_f(config.project.cache_file) if options[:clear_cache]
```
beechtom, line 201:
```suggestion
      FileUtils.rm(config.project.cache_file) if options[:clear_cache] && config.project.cache_file.exist?
```

`FileUtils.rm_f` will suppress all StandardErrors, which we don't want to do. See Nick's comment regarding `FileUtils.rm_rf` here: https://github.com/puppetlabs/bolt/pull/1979#issuecomment-658942621

lib/bolt/cli.rb
```diff
@@ -210,7 +212,9 @@ def update_targets(options)
       target_string = "'--targets', '--rerun', or '--query'"
       if target_opts.length > 1
         raise Bolt::CLIError, "Only one targeting option #{target_string} may be specified"
-      elsif target_opts.empty? && options[:subcommand] != 'plan'
+      elsif target_opts.empty? &&
+            options[:subcommand] != 'plan' &&
+            options[:action] != 'refresh'
```
beechtom, line 217:
```suggestion
      elsif target_opts.empty? && options[:subcommand] != 'plan'
```

lib/bolt/cli.rb
```diff
@@ -450,6 +454,12 @@ def execute(options)
                else
                  list_topics
                end
+      when 'cache'
+        case options[:action]
+        when 'refresh'
+          FileUtils.rm_f(@config.project.cache_file)
+          code = 0
+        end
```
beechtom, line 462:
```suggestion
```

lib/bolt/config.rb
```diff
@@ -484,6 +484,10 @@ def validate
...
```
beechtom, line 489:
```suggestion
      if @data['plugin-cache'] && @data['plugin-cache']['ttl'] < 0
        raise Bolt::ValidationError, "plugin-cache.ttl must be a positive integer"
      end
```

This can be handled by setting `:minimum` in the schema.

lib/bolt/config/options.rb
```diff
@@ -33,6 +33,17 @@ module Options
...
               type: "string"
+            },
+            "_cache" => {
+              description: "This feature is experimental. Enable plugin caching and set a time-to-live.",
+              type: "object",
+              required: ["ttl"],
+              properties: {
+                "ttl" => {
+                  description: "Time in seconds to keep the plugin cache.",
+                  type: "integer"
```
beechtom, line 44:
```suggestion
                  type: "integer",
                  minimum: 1
```

lib/bolt/config/options.rb
```diff
@@ -175,6 +186,19 @@ module Options
...
           _default: "project/inventory.yaml"
         },
+        "plugin-cache" => {
+          description: "This feature is experimental. Enable plugin caching and set the time-to-live.",
+          type: Hash,
+          required: ["ttl"],
+          properties: {
+            "ttl" => {
+              description: "Time in seconds to keep the plugin cache.",
+              type: Integer
```
beechtom, line 196:
```suggestion
              type: Integer,
              minimum: 1
```

rakelib/pwsh.rake
```diff
@@ -41,16 +41,17 @@ namespace :pwsh do
...
-      'init'       => 'New',
-      'install'    => 'Install',
       'encrypt'    => 'Protect',
       'decrypt'    => 'Unprotect',
+      'download'   => 'Receive',
+      'init'       => 'New',
+      'install'    => 'Install',
       'migrate'    => 'Update',
+      'new'        => 'New',
+      'refresh'    => 'Clear',
```
beechtom, line 51:
```suggestion
```

lib/bolt/plugin.rb
```diff
@@ -288,16 +313,55 @@ def resolve_single_reference(reference)
...
+      end
+      result
     end
     private :resolve_single_reference
 
+    def read_cache
+      if File.zero?(@config.project.cache_file)
+        {}
+      else
+        # This will fail if the file is empty
```
beechtom, line 339:
```suggestion
```

`File.zero?` returns `true` if the file is empty.

lib/bolt/plugin.rb
```diff
@@ -288,16 +313,55 @@ def resolve_single_reference(reference)
...
+        JSON.parse(File.read(@config.project.cache_file))
```
beechtom, line 340: Might want to catch JSON parse errors here. It's unlikely that the file _won't_ be valid JSON, but just in case.

lib/bolt/plugin.rb
```diff
@@ -277,6 +278,30 @@ def resolve_top_level_references(data)
...
+          # First remove any cache entries past their ttl
+          # This prevents removing plugins from leaving orphaned cache entries
+          cache.delete_if do |_, entry|
+            Time.now - Time.parse(entry['mtime']) >= entry['ttl']
+          end
+          File.write(@config.project.cache_file, cache.to_json)
+
+          return cache[id]['result']
+        end
+      end
```
beechtom, line 303: I think there's enough going on here that it should be separated out into a separate method.

lib/bolt/config.rb
```diff
@@ -519,6 +519,10 @@ def modulepath=(value)
       @data['modulepath'] = value
     end
 
+    def plugin_cache
+      @data['plugin-cache']
```
beechtom, line 523: This should have a default of `{}` in the initializer, similar to `apply_settings`, `plugin_hooks`, etc.

lib/bolt/plugin.rb
```diff
@@ -288,16 +313,55 @@ def resolve_single_reference(reference)
...
+        unless r.key?('ttl')
+          raise Bolt::ValidationError, "_cache must set 'ttl' key"
+        end
+
+        unless r['ttl'].is_a?(Integer) && r['ttl'] >= 0
+          raise Bolt::ValidationError,
+                "_cache.ttl must be a positive Integer, received #{r.class}: #{r.inspect}"
+        end
+      end
+      reference.dig('_cache', 'ttl') || @config.plugin_cache&.fetch('ttl', nil)
```
beechtom, line 361:
```suggestion
      reference.dig('_cache', 'ttl') || @config.plugin_cache['ttl']
```

If `plugin-cache` has a default in config you can get rid of the safe operator.

lib/bolt/config/options.rb
```diff
@@ -33,6 +33,18 @@ module Options
...
+            },
+            "_cache" => {
+              description: "This feature is experimental. Enable plugin caching and set a time-to-live.",
+              type: "object",
+              required: ["ttl"],
+              properties: {
+                "ttl" => {
+                  description: "Time in seconds to keep the plugin cache.",
+                  type: "integer",
+                  minimum: 0
```
beechtom, line 45: It seems like setting a minimum of 0 would effectively make the cache useless?

lib/bolt/plugin.rb
```diff
@@ -288,16 +313,55 @@ def resolve_single_reference(reference)
...
+    def read_cache
+      if File.zero?(@config.project.cache_file)
+        {}
+      else
+        # This will fail if the file is empty
+        JSON.parse(File.read(@config.project.cache_file))
+      end
+    end
+
+    def cache?(reference)
```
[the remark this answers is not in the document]
lucywyman: I mean it's kind of used as a boolean? I can change it.

lib/bolt/plugin.rb
lucywyman: Right the comment is meant to say "We have to check if the file is empty first before reading it because otherwise it will fail".

lib/bolt/plugin.rb
lucywyman: It shouldn't have a default though, we need to know if the user set it or not. I guess we could set plugin-cache to `ttl => nil` but I don't see how that's better?

lib/bolt/config/options.rb
lucywyman: It's a way to turn off a default cache. Say I want to cache all my plugins except one, setting it to 0 lets users disable cache for one or a few plugins.

beechtom: Ah, that's helpful. I would add that to the docs so users can discover that behavior more easily.

lucywyman: Will do! I also added a check for 0 so we don't waste time reading/writing cache if it's 0.

lucywyman:
Ok, I'm gonna pull the caching functions into their own class. Converting back to a draft for now.

lib/bolt/plugin_cache.rb
```diff
@@ -0,0 +1,70 @@
+# frozen_string_literal: true
+
+require 'fileutils'
+require 'bolt/error'
+require 'bolt/util'
+
+module Bolt
+  class PluginCache
```
beechtom, line 8: Small thing, but I think this should be a nested class of `Bolt::Plugin`, since its behavior is specific to plugins and it already has 'Plugin' in its name.

lib/bolt/config/options.rb
```diff
@@ -34,6 +34,20 @@ module Options
               description: "The name of the plugin.",
               type: "string"
             }
+          },
+          additionalProperties: {
+            "_cache" => {
```
beechtom, line 39: This should be under `:properties` since it's a known property. `:additionalProperties` is meant to validate properties that are unknown (such as in `log` where we don't know the filenames used for keys, but we do know that they should be hashes with only two possible keys).

lib/bolt/plugin_cache.rb
```diff
@@ -0,0 +1,70 @@
...
+      return if reference['_cache'].nil?
+      r = reference['_cache']
+      unless r.is_a?(Hash)
+        raise Bolt::ValidationError,
+              "_cache must be a Hash, received #{r.class}: #{r.inspect}"
+      end
+
+      unless r.key?('ttl')
+        raise Bolt::ValidationError, "_cache must set 'ttl' key."
+      end
```
beechtom, line 63: We'll want to validate that `ttl` is >= 0 here, since the config validator won't validate references.

lib/bolt/util.rb
```diff
@@ -22,6 +22,28 @@ def read_arg_file(file)
         raise Bolt::FileError.new("Error attempting to read #{file}: #{e}", file)
       end
 
+      def read_json_file(path, filename)
```
beechtom, line 25: I wonder if we could just alias `read_yaml_hash` and `read_optional_yaml_hash`? Psych syntax errors also seem to be a bit more helpful than JSON's:

```
Psych::SyntaxError ((test.json): did not find expected ',' or '}' while parsing a flow mapping at line 1 column 1)
```

```
JSON::ParserError (784: unexpected token at '{)
```

lib/bolt/plugin_cache.rb
lucywyman: All the current plugin subclasses are plugins themselves, so I think that'd make it seem like this is a caching plugin and not a builtin feature.

lib/bolt/config/options.rb
lucywyman: Hm ok...this was raising a "_cache required" error, but I think I solved that in the code itself.

lib/bolt/plugin_cache.rb
beechtom: `Bolt::Plugin::Module` isn't a plugin itself 🤷

lib/bolt/util.rb
lucywyman: We'd have to refactor those anyway as `.rerun.json` reads and writes an array, not a hash. Those functions are really used for reading configuration files, while the json function is used for internal files (for example we don't read task metadata with it). I don't expect users to even encounter this error. I guess we can load the file with Psych, but there is some level of risk as I'm not entirely sure JSON and YAML are always [guaranteed to be strict supersets](https://yaml.org/spec/1.2/spec.html#id2759572) (for example JSON allows duplicate keys while YAML doesn't), and it generally seems weird to load json with a yaml loader. I don't think it makes sense for the file to be in YAML either, as YAML is really more for data that requires human interaction and I don't expect anyone to ever look at this file.

lib/bolt/config/options.rb
beechtom: It looks like docs are missing info about setting `ttl` to 0 to disable caching when `plugin-cache` is configured.

spec/integration/cache_spec.rb
```diff
@@ -0,0 +1,102 @@
...
+require 'bolt_spec/config'
+require 'bolt_spec/integration'
+require 'bolt_spec/project'
+
+describe 'caching plugins' do
+  include BoltSpec::Config
+  include BoltSpec::Integration
+  include BoltSpec::Project
+
+  let(:cache_file) { File.join(project_path, '.plugin_cache.json') }
```
beechtom, line 13: You can just call `project.cache_file`, since `BoltSpec::Project` includes a method that returns the project.

spec/lib/bolt_spec/project.rb
```diff
@@ -43,7 +44,7 @@ def config_path
     end
 
     def project_config
-      {}
+      { 'name' => @project_name }
```
beechtom, line 47: We probably want to update `with_project` to do something like:

```
    def with_project(name = 'project')
      Dir.mktmpdir(nil, Dir.pwd) do |tmpdir|
        @tmpdir       = Pathname.new(tmpdir).expand_path
        @project_path = (@tmpdir + name).expand_path
        @config_path  = (@project_path + 'bolt-project.yaml').expand_path

        FileUtils.mkdir_p(@project_path)
        File.write(@config_path, { 'name' => name }.merge(project_config).to_yaml)  <--- THIS
        yield
      end
    end
```

That way the project has a name even if `project_config` is overridden for a specific test.

spec/lib/bolt_spec/project.rb
```diff
@@ -3,6 +3,7 @@
 module BoltSpec
   module Project
     def with_project(name = 'project')
+      @project_name = name
```
beechtom, line 6:
```suggestion
```

I don't think this is necessary if we ensure the name is written to the config file in `with_project`.

lib/bolt/plugin.rb
```diff
@@ -288,16 +299,24 @@ def resolve_single_reference(reference)
...
       end
+
+      plugin_cache.write_cache(result) if cache?(reference)
+
+      result
     end
     private :resolve_single_reference
 
+    private def cache?(reference)
+      reference.key?('_cache') || @config.plugin_cache.key?('ttl')
```
beechtom, line 317:
```suggestion
      reference.key?('_cache') || @config.plugin_cache.key?('ttl')
```

`Bolt::Config` doesn't receive config from `bolt-project.yaml` if that file only includes keys not found in `bolt.yaml`. In that case, `@config.plugin_cache` returns an empty hash, even if `plugin-cache` is configured.

spec/lib/bolt_spec/project.rb
lucywyman: What if the test specifically doesn't want to include a name? It seems more clear to let devs set the entire config than to have some config magically set.

beechtom (approved):
This is _amazing_! 🥳

documentation/experimental_features.md
```diff
@@ -8,6 +8,58 @@ API may change, requiring the user to update their code or configuration. The
...
+This feature was introduced in [Bolt TODO]()
+
+Bolt supports the use of plugins to dynamically load information during a Bolt run and change how
+Bolt executes certain actions. Bolt also loads all configuration plugins for most Bolt commands, and
+all inventory plugins for any action that requires the inventory, even if the action only uses a
+subset of targets from the inventory. Plugins can sometimes take a long time to execute, and several
+plugin invocations can add quite a bit of start up time to any Bolt command regardless of whether it
+uses the plugin results.
+
+To mitigate this we've introduced plugin caching. Plugin caching is enabled by configuring a
```
hestonhoffman, line 22:
```suggestion
To mitigate the time it takes for Bolt to load plugins, we've introduced plugin caching. 
Plugin caching is enabled by configuring a
```

documentation/experimental_features.md
```diff
@@ -8,6 +8,58 @@ API may change, requiring the user to update their code or configuration. The
...
+time-to-live (ttl) for the cache, either by setting a default ttl for all plugins with the
```
hestonhoffman, line 23:
```suggestion
Time to Live (TTL) for the cache, either by setting a default TTL for all plugins with the
```

documentation/experimental_features.md
```diff
@@ -8,6 +8,58 @@ API may change, requiring the user to update their code or configuration. The
...
+`plugin-cache` project configuration option or by setting `ttl` under the `_cache` option for an
+individual plugin invocation. The ttl is always in seconds. Plugin results will be cached inside the
```
hestonhoffman, line 25:
```suggestion
individual plugin invocation. The TTL is always in seconds. Bolt caches plugin results inside the
```

documentation/experimental_features.md
```diff
@@ -8,6 +8,58 @@ API may change, requiring the user to update their code or configuration. The
...
+Bolt project, and all expired cache entries will be removed whenever Bolt uses a cache result. This
```
hestonhoffman, line 26:
```suggestion
Bolt project, and removes all expired cache entries whenever it uses a cache result. This
```

documentation/experimental_features.md
```diff
@@ -8,6 +8,58 @@ API may change, requiring the user to update their code or configuration. The
...
+prevents cache entries from getting orphaned and never removed. Bolt identifies a cache result based
+on a hash (delightfully named 'bubblebabble') of the entire plugin invocation, minus the `_cache`
+key. If any element of the plugin invocation changes, Bolt will reload the plugin and update the
```
hestonhoffman, line 29:
```suggestion
key. If any element of the plugin invocation changes, Bolt reloads the plugin and updates the
```

documentation/experimental_features.md
```diff
@@ -8,6 +8,58 @@ API may change, requiring the user to update their code or configuration. The
...
+cache.
+
+You can set `ttl` to 0 to disable caching for a particular plugin. Caching is disabled by default,
+so this is mostly useful if you have a default cache ttl configured under `plugin-cache` in your
```
hestonhoffman, line 33:
```suggestion
so this is mostly useful if you have a default cache `ttl` configured under `plugin-cache` in your
```

documentation/experimental_features.md
```diff
@@ -8,6 +8,58 @@ API may change, requiring the user to update their code or configuration. The
...
+`bolt-project.yaml` or `bolt-defaults.yaml` and want to disable caching for individual plugin
+invocations.
+
+**NOTE**: Plugin results are currently cached in plaintext. Encrypted caching is planned for a
```
hestonhoffman, line 37:
```suggestion
> **NOTE**: Plugin results are currently cached in plaintext. Encrypted caching is planned for a
```

documentation/experimental_features.md
```diff
@@ -8,6 +8,58 @@ API may change, requiring the user to update their code or configuration. The
...
+future release.
+
+Users can clear all cache entries using the `--clear-cache` or `-ClearCache` command-line option
+with any Bolt command.
+
+This inventory sets a ttl of 1 hour, or 3600 seconds, for the PuppetDB plugin:
```
hestonhoffman, line 43:
```suggestion
This inventory sets a TTL of 1 hour, or 3600 seconds, for the PuppetDB plugin:
```

documentation/experimental_features.md
```diff
@@ -8,6 +8,58 @@ API may change, requiring the user to update their code or configuration. The
...
+targets:
+  - _plugin: puppetdb
+    _cache:
+      ttl: 3600
+    query: "inventory[certname] { facts.osfamily = 'RedHat' }"
+    target_mapping:
+      name: certname
+```
+
+This config file sets a default ttl of 30 minutes, or 1800 seconds, for all plugins:
```
hestonhoffman, line 54:
```suggestion
This config file sets a default TTL of 30 minutes, or 1800 seconds, for all plugins:
```

documentation/experimental_features.md
```diff
@@ -8,6 +8,58 @@ API may change, requiring the user to update their code or configuration. The
...
+```
+plugin-cache:
+  ttl: 1800
+```
+
+**NOTE**: The same cache and cache configuration is used for the `resolve_reference()` plan
```
hestonhoffman, line 60:
```suggestion
> **NOTE**: The same cache and cache configuration is used for the `resolve_reference()` plan
```

hestonhoffman (approved):
Some small style changes, but the doc looks good to me :)

Issue #2382, opened 2020-11-19
Generate packaged modules page


hestonhoffman:
## Generate a packaged_modules.md

It would be nice if we generated a list of packaged modules so we could link to it. Having a separate page would also make discoverability a bit simpler cos at the moment it doesn't really fit in either the overview or the installing pages.

PR #2384, opened 2020-11-19
(GH-2382) Add 'Packaged modules' docs page

beechtom:
This adds a generated page to Bolt docs that lists the modules that are
shipped with Bolt packages. The list of modules is read directly from
Bolt's Puppetfile, while descriptions are retrieved from each module's
metadata or README file.

!no-release-note


hestonhoffman (approved):
Awesome! I'll add a link to this in my modules PR

lucywyman (approved):

PR #2388, opened 2020-11-23
(maint) Validate inventoryfile config

lucywyman:
Previously we tried to use `Bolt::Util.file_stat` to verify that a
configured inventoryfile existed before warning that CLI options may be
overridden by the inventoryfile. This would stacktrace if the file did
not exist. We now validate that the file exists and can be read when
loading config if `inventoryfile` is configured, validate
`default_inventoryfile` if the file exists, and use `File.exist?` in the
CLI when seeing if we need to warn that inventoryfile may override CLI
options. This provides more standard errors and prevents stacktracing.

!no-release-note


beechtom (approved):

beechtom:
How did we not handle this for _so long_?? 👍  for picking it up.

Issue #2364, opened 2020-11-17
Create an inventoryfile as part of bolt project init


lucywyman:
## Use Case
If an inventoryfile already exists, do nothing.
Proposed default content (all commented out). This should include top-level config as well, and includes updating the documentation to have the same example under 'basic example'
```
---
# This is an example inventory.yaml
# Link to docs
groups:
  - name: linux
    targets:
      - target1.example.com
      - target2.example.com
    config:
      transport: ssh
      ssh:
        user: myuser
  - name: windows
    targets:

      - uri: target4.example.com
        alias: windows2
    config:
      transport: winrm
```
Question:
- Should this accept an argument to configure adding to it? I can imagine taking a list of bare hostnames and adding them to a top-level `targets` key. _This is a little too complex and doesn't provide enough value to conside_.

PR #2391, opened 2020-11-23
(GH-2364) Create inventory.yaml when creating new projects

beechtom:
The `bolt project init` and `New-BoltProject` commands now create an
`inventory.yaml` file with an example inventory commented out. If a
directory already includes an `inventory.yaml` file, Bolt will not
create a new one.

!feature

* **Create `inventory.yaml` file when creating new projects**
  ([#2364](https://github.com/puppetlabs/bolt/issues/2364))

  The `bolt project init` and `New-BoltProject` commands now create an
  `inventory.yaml` file in the new project.

### DIFF
--- a/documentation/inventory_file_v2.md
+++ b/documentation/inventory_file_v2.md
@@ -294,6 +294,9 @@ targets to the SSH protocol.
 The `windows` group lists its targets and sets the default transport for the
 targets to the WinRM protocol.
 
+The inventory file sets top-level configuration that applies to the `all` group.
+It sets default configuration for the `ssh` and `winrm` transports.
+
 ```yaml
 groups:
   - name: linux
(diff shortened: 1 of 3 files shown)


documentation/inventory_file_v2.md
```diff
@@ -304,12 +307,19 @@ groups:
...
-      - uri: target4.example.com
-        alias: windows2
+      - name: win1
+        uri: target3.example.com
+      - name: win2
+        uri: target4.example.com
     config:
       transport: winrm
+config:
+  ssh:
```
lucywyman, line 317: It seems useful to set one of these at the group level as well so people know that you can set more than just `transport` there and also to demo that the transport needs to be a key under `config`.

lib/bolt/project_manager.rb
```diff
@@ -17,12 +45,13 @@ def initialize(config, outputter, pal)
     def create(path, name, modules)
       require 'bolt/module_installer'
 
-      project      = Pathname.new(File.expand_path(path))
-      old_config   = project + 'bolt.yaml'
-      config       = project + 'bolt-project.yaml'
-      puppetfile   = project + 'Puppetfile'
-      moduledir    = project + '.modules'
-      project_name = name || File.basename(project)
+      project       = Pathname.new(File.expand_path(path))
```
lucywyman, line 48: This is why I hate aligned operators :P

beechtom: Different strokes :P I just prefer them since it makes things more legible.

Issue #2376, opened 2020-11-18
Warn when unknown keys are found in configuration files


lucywyman:
## Use Case
We should issue a warning when unknown keys are found in certain configuration data - namely anything where there is a known set of respected values. We should not warn, for example, when parameters are passed to a plugin, but we should warn when an unknown key exists under the `ssh:` config key.

PR #2386, opened 2020-11-20
(GH-2363) Deprecate inventoryfile config option

beechtom:
This deprecates the `inventoryfile` config option, which is slated for
removal in Bolt 3.0.

!deprecation

* **Deprecate `inventoryfile` configuration option**
  ([#2363](https://github.com/puppetlabs/bolt/issues/2363))

  The `inventoryfile` configuration option has been deprecated and will
  be removed in Bolt 3.0. Users should move contents from non-default
  inventory files to the `inventory.yaml` file in a Bolt project, or can
  use the `--inventoryfile` command-line option to load a non-default
  inventory file.

PR #2392, opened 2020-11-23
(GH-2376) Warn about unknown configuration options

beechtom:
This updates the config validator to issue warnings for unknown
config options. When the validator encounters an unknown option, it
stores a warning indicating where the unknown option is specified, which
is later issued by the logger. Options are only reported as unknown if
the schema for the option defines the `:properties` key but _not_ the
`:additionalProperties` key.

!feature

* **Warn about unknown configuration options**
  ([#2376](https://github.com/puppetlabs/bolt/issues/2376))

  Bolt now issues a warning when it detects an unknown configuration
  option. The warning will indicate where the configuration option is
  located.

### DIFF
--- a/lib/bolt/config.rb
+++ b/lib/bolt/config.rb
@@ -92,10 +92,8 @@ module Bolt
 
     def self.defaults_schema
       base      = OPTIONS.slice(*BOLT_DEFAULTS_OPTIONS)
-      inventory = INVENTORY_OPTIONS.each do |option, definition|
-        if TRANSPORT_CONFIG.key?(option)
-          definition[:properties] = TRANSPORT_CONFIG[option].schema
-        end
+      inventory = INVENTORY_OPTIONS.each_with_object({}) do |(option, definition), acc|
+        acc[option] = TRANSPORT_CONFIG.key?(option) ? definition.merge(TRANSPORT_CONFIG[option].schema) : definition
       end
 
       base['inventory-config'][:properties] = inventory
@@ -103,10 +101,8 @@ module Bolt
     end
 
     def self.bolt_schema
-      inventory = INVENTORY_OPTIONS.each do |option, definition|
-        if TRANSPORT_CONFIG.key?(option)
-          definition[:properties] = TRANSPORT_CONFIG[option].schema
-        end
+      inventory = INVENTORY_OPTIONS.each_with_object({}) do |(option, definition), acc|
+        acc[option] = TRANSPORT_CONFIG.key?(option) ? definition.merge(TRANSPORT_CONFIG[option].schema) : definition
       end
 
       OPTIONS.slice(*BOLT_OPTIONS).merge(inventory)
--- a/lib/bolt/config.rb
+++ b/lib/bolt/config.rb
@@ -38,6 +38,14 @@ module Bolt
                project.data
              else
                c = Bolt::Util.read_optional_yaml_hash(project.config_file, 'config')
+
+               # Validate the config against the schema. This will raise a single error
+               # with all validation errors.
+               Validator.new.tap do |validator|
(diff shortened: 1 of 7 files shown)


lib/bolt/config.rb
```diff
@@ -128,7 +128,10 @@ def self.load_bolt_defaults_yaml(dir)
 
       # Validate the config against the schema. This will raise a single error
       # with all validation errors.
-      Validator.new.validate(data, defaults_schema, filepath)
```
lucywyman, line 131: I know it's going away soon but we should also add this to the `from_file` handler so that `--configfile` also issues this warning

lucywyman (changes requested):

lucywyman:
Would it also make sense to start validating the inventory here? We already validate inventory-config from the defaults file.

lucywyman:
Hm, this might also be a separate issue but I noticed the message if users have misconfigured `log` is a little misleading:
```
lucy ~/githubs/bolt/project ((f69845b...)) $ be bolt command run whoami -t localhost
Invalid configuration at /home/lucy/githubs/bolt/project/bolt-project.yaml:
  Value at 'log.foo' must be one of disable
```

We might need to have a special handler for that.

beechtom:
@lucywyman 

What config are you using for `log`? If it's set as a string other than disable, the validator will validate it according to the string definition, not the hash one. That will apply to any config option that has can be more than one type, and is a little harder to provide a concise error message for.

This class can't really be used to validate the inventory file, since the schema would need to be recursive since groups can be nested... which isn't really feasible with how the schemas are set up in Ruby. We can definitely make an inventory validator, though, and was something I was thinking about earlier today! :)

lib/bolt/config.rb
beechtom: Added to https://github.com/puppetlabs/bolt/pull/2386 already.

lucywyman:
Right I understand how the validator is working, but in the case of `log` (and anything else that takes multiple types) the message is misleading because it doesn't instruct users that they need to provide a hash. For example it's easy to see someone writing:
```
log:
  mylog: debug
```
And being confused by the error message. I think the message should read something like:
```
Invalid configuration at /home/lucy/githubs/bolt/project/bolt-project.yaml:
  Value at 'log.mylog' must be either the String 'disable' or a Hash
```

beechtom:
Updated the validator to append alternate types to an error message. Currently this is only used by `#validate_string`, but I think that should suffice for now since none of the other multi-type options will lead to potential confusion. For example, if you're getting an error for `modulepath` it's not really helpful for us to say "or it must be a String" since someone using an array probably meant to use an array.

lib/bolt/config/validator.rb
```diff
@@ -169,6 +195,18 @@ def validate(data, schema, location = nil)
         end
       end
 
+      # Adds an error that includes additional helpful information for values
+      # that accept multiple types.
+      #
+      private def multitype_error(message, value, definition)
+        if Array(definition[:type]).count > 1
+          types    = Array(definition[:type]) - [value.class]
+          message += " or must be of type #{types.join(', ')}"
```
lucywyman, line 204: This feels pedantic, but what do you think about joining this with `or` instead? It looks like all multitypes take at most 2 types, and I feel like this can be misinterpreted as an "Array of Strings" or "Hash of strings":
```
  Value at 'modulepath' must be of type Array, String
```

Plus, I think having 3+ linked by "or" is like...annoying but still clear:
```
  Value at 'modulepath' must be of type Array or Hash or String
```

lucywyman (approved):
One small comment, looks good to me!

lib/bolt/config/validator.rb
beechtom: Fine by me 🤷  Updated.

Just need to figure out why that one test is failing now...

PR #2394, opened 2020-11-24
(PE-30636) Add load_plugins option to Bolt::Plugin.setup

ekinanp:
When set to true, the returned Bolt::Plugin object will load
plugins on-demand. This is what Bolt is already doing and hence
the default behavior.

When set to false, the returned Bolt::Plugin object will raise a new
PluginLoadingDisabled error indicating that plugin loading is disabled.
However, users can still add their statically loaded plugins via
Bolt::Plugin#add_plugin.

!no-release-note

### DIFF
--- a/lib/bolt/plugin.rb
+++ b/lib/bolt/plugin.rb
@@ -36,6 +36,13 @@ module Bolt
           super("Plugin #{plugin_name} does not support #{hook}", 'bolt/unsupported-hook')
         end
       end
+
+      class LoadingDisabled < PluginError
+        def initialize(plugin_name)
+          msg = "Cannot load plugin #{plugin_name}: plugin loading is disabled"
+          super(msg, 'bolt/plugin-loading-disabled', { 'plugin_name' => plugin_name })
+        end
+      end
     end
 
     class PluginContext
@@ -119,8 +126,8 @@ module Bolt
       end
     end
 
-    def self.setup(config, pal, analytics = Bolt::Analytics::NoopClient.new)
-      plugins = new(config, pal, analytics)
+    def self.setup(config, pal, analytics = Bolt::Analytics::NoopClient.new, **opts)
+      plugins = new(config, pal, analytics, **opts)
 
       config.plugins.each_key do |plugin|
         plugins.by_name(plugin)
@@ -141,12 +148,13 @@ module Bolt
 
     private_class_method :new
 
-    def initialize(config, pal, analytics)
+    def initialize(config, pal, analytics, load_plugins: true)
       @config = config
       @analytics = analytics
       @plugin_context = PluginContext.new(config, pal, self)
       @plugins = {}
       @pal = pal
+      @load_plugins = load_plugins
       @unknown = Set.new
       @resolution_stack = []
       @unresolved_plugin_configs = config.plugins.dup
@@ -169,6 +177,8 @@ module Bolt
     end
 
     def add_ruby_plugin(plugin_name)
+      raise PluginError::LoadingDisabled, plugin_name unless @load_plugins
+
       cls_name = Bolt::Util.snake_name_to_class_name(plugin_name)
       filename = "bolt/plugin/#{plugin_name}"
       require filename
@@ -185,10 +195,17 @@ module Bolt
     def add_module_plugin(plugin_name)
       opts = {
         context: @plugin_context,
+        # Make sure that the plugin's config is validated _before_ the unknown-plugin
+        # and loading-disabled checks. This way, we can fail early on invalid plugin
+        # config instead of _after_ loading the modulepath (which can be expensive).
         config: config_for_plugin(plugin_name)
       }
 
-      plugin = Bolt::Plugin::Module.load(plugin_name, modules, opts)
+      mod = modules[plugin_name]
+      raise PluginError::Unknown, plugin_name unless mod&.plugin?
+      raise PluginError::LoadingDisabled, plugin_name unless @load_plugins
+
+      plugin = Bolt::Plugin::Module.load(mod, opts)
       add_plugin(plugin)
     end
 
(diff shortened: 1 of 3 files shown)


ekinanp:
Marking this merged until @nicklewis has had a chance to review this since this PR was based off his suggestion.

lib/bolt/plugin.rb
```diff
@@ -36,6 +36,12 @@ def initialize(plugin_name, hook)
           super("Plugin #{plugin_name} does not support #{hook}", 'bolt/unsupported-hook')
         end
       end
+
+      class PluginLoadingDisabled < PluginError
```
ekinanp, line 40: I feel like `LoadingDisabled` would feel weird b/c the name does not explicitly tell me _what_ is being loaded. Hence, I feel like `PluginLoadingDisabled` may be the better name here.

barriserloth (approved):

lib/bolt/plugin.rb
```diff
@@ -36,6 +36,12 @@ def initialize(plugin_name, hook)
...
+        def initialize(plugin_name)
+          super("Cannot load plugin #{plugin_name}: plugin loading is disabled", 'bolt/plugin-loading-disabled')
```
donoghuc, line 42: Can we put the plugin name in structured data (details hash maybe) so that we can use it easily when we convert it when we catch this for `PluginNotSupported` errors?

ekinanp: Done.

lib/bolt/plugin.rb
```diff
@@ -188,6 +198,13 @@ def add_module_plugin(plugin_name)
         config: config_for_plugin(plugin_name)
       }
 
+      # Make sure that the plugin's config is validated _before_ these checks.
```
nicklewis, line 201: I don't understand the context of this comment. It seems like it's referring to the fact that we call `config_for_plugin` above. Is that the correct interpretation? If so, I think it should be moved to that line and rephrased because it seems ambiguous here.

lib/bolt/plugin.rb
```diff
@@ -188,6 +198,13 @@ def add_module_plugin(plugin_name)
...
+      # This way, we can fail early on invalid plugin config instead of _after_
+      # loading the modulepath (which can be expensive).
+      mod = modules[plugin_name]
+      raise PluginError::Unknown, plugin_name unless mod&.plugin?
```
nicklewis, line 205: Since we've now done this check, `Bolt::Plugin::Module.load` can just take the single `module` instead of the `modules` list it currently takes.

lib/bolt/plugin.rb
ekinanp: Yeah, I can rephrase it. It's mostly for people (like me) who may wonder why we're validating config for a potentially unknown plugin.

ekinanp:
Updated to address Nick L's comments.

lib/bolt/plugin/module.rb
```diff
@@ -12,15 +12,14 @@ def initialize(msg, plugin)
         end
       end
 
-      def self.load(name, modules, opts)
-        mod = modules[name]
+      def self.load(mod, opts)
         if mod&.plugin?
```
nicklewis, line 16: The `&` is unnecessary here. We've guarded against `nil` at the single call site and can therefore incorporate that 
as a precondition of this method. And as the code stands now, if this argument _were_ ever `nil`, the `raise` would itself fail, because it calls `mod.name` directly.

ekinanp: Updated.

nicklewis:
Just so I understand it, the expected usage of this will look like:

```rb
plugins = Bolt::Plugin.setup(config, pal, load_plugins: false)
plugin.add_plugin(Bolt::Plugin::Module.load(plugins.modules['connect'], pal))
```

Is that correct (aside from the actual name of the plugin)?

It might be helpful, though not necessarily as part of this PR, to rework `add_ruby_plugin` and `add_module_plugin` to only be responsible for _loading_ the plugins. Then `by_name` could call `add_plugin(load_module_plugin(plugin_name))`, and we could duplicate that logic in bolt server. The trouble with that approach is that our checking the plugin's existence _before_ checking whether plugin loading is disabled makes it difficult to reuse the method in two contexts. We of course wouldn't want loading the plugin to fail when we call it explicitly.

ekinanp:
Close. Its use will be something like

```
plugins = Bolt::Plugin.setup(config, pal, load_plugins: false)
plugins.add_plugin(PuppetConnectData.new(data_hash))
```

(`PuppetConnectData`'s just a Ruby plugin, so a Ruby class).

Re: the extra refactor, I don't think it's worth spending extra time doing that in this PR since the current implementation is still pretty clean and the bolt-server use is straightforward-enough that the refactor wouldn't add much more clarity to it.

nicklewis:
Ah right, I forgot it will be a Ruby plugin, so that particular use case is a lot easier. :thumbsup:

McdonaldSeanp (approved):

lib/bolt/plugin.rb
```diff
@@ -36,6 +36,13 @@ def initialize(plugin_name, hook)
           super("Plugin #{plugin_name} does not support #{hook}", 'bolt/unsupported-hook')
         end
       end
+
+      class LoadingDisabled < PluginError
+        def initialize(plugin_name)
+          msg = "Cannot load plugin #{plugin_name}: plugin loading is disabled"
+          super(msg, 'bolt/plugin-loading-disabled', { plugin_name: plugin_name })
```
donoghuc, line 43: hmm, are there any implications to using a keyword for the key here? I feel like everywhere else in bolt errors we use a string...

ekinanp: That's just me not knowing Bolt conventions. Updated to use a string.

donoghuc (approved):

PR #2395, opened 2020-11-24
(GH-2365) Deprecate Powershell 2 support

### LINKED ISSUE #2365: Deprecate Powershell 2.0
## Use Case
We should warn users who are using Powershell 2.0 on the Bolt controller or the remote target that we'll be removing support for PS2 soon. This includes checking `PSVersionTable` on the remote target for the version and issuing the warning based on the output. If we can't parse PSVersionTable output we should not print the deprecation.

lucywyman:
This adds a deprecation warning when users run Bolt from Powershell
version less than 3, and for each target users run against in Powershell
less than 3. This only applies to the shell Bolt is run from or
connected to, not all Powershell versions on the machine. A warning is
issued for each target the user connects to with a deprecated version so
that users know which targets have deprecated PS versions.

Closes #2365

!deprecation

* **Deprecate Powershell 2 support** ([2365](https://github.com/puppetlabs/bolt/issues/2365))

  Support for Powershell 2 on both Bolt targets and controllers is
  deprecated, and will be dropped in Bolt 3.0.

### DIFF
--- a/lib/bolt/shell/powershell.rb
+++ b/lib/bolt/shell/powershell.rb
@@ -14,6 +14,22 @@ module Bolt
         extensions = [target.options['extensions'] || []].flatten.map { |ext| ext[0] == '.' ? ext : '.' + ext }
         extensions += target.options['interpreters'].keys if target.options['interpreters']
         @extensions = DEFAULT_EXTENSIONS + extensions
+        validate_ps_version
+      end
+
+      def validate_ps_version
+        version = execute("$PSVersionTable.PSVersion.Major").stdout.string.chomp
+        if !version.empty? && version.to_i < 3
+          # This lets us know how many targets have Powershell 2, and lets the
+          # user know how many targets they have with PS2
+          msg = "Detected PowerShell 2 on one or more targets.\nPowerShell 2 "\
+            "is deprecated, and support will be removed in Bolt 3.0. See "\
+            "bolt-debug.log or run with '--log-level debug' to see the full "\
+            "list of targets with PowerShell 2."
+
+          Bolt::Logger.deprecation_warning("PowerShell 2", msg)
+          @logger.debug("Detected PowerShell 2 on #{target}.")
+        end
       end
 
       def provided_features
--- a/lib/bolt/cli.rb
+++ b/lib/bolt/cli.rb
@@ -198,11 +198,23 @@ module Bolt
       config.deprecations.each { |dep| Bolt::Logger.deprecation_warning(dep[:type], dep[:msg]) }
 
       warn_inventory_overrides_cli(options)
+      validate_ps_version
 
       options
     rescue Bolt::Error => e
       outputter.fatal_error(e)
       raise e
+    end
+
+    private def validate_ps_version
+      if Bolt::Util.powershell?
+        target = inventory.get_target('localhost')
+        Bolt::Transport::Local.new.with_connection(target) do |conn|
+          # This will automatically validate the powershell version on the Bolt
+          # controller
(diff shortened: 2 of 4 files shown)


lib/bolt/shell/powershell.rb
```diff
@@ -14,6 +14,22 @@ def initialize(target, conn)
...
+        if !version.empty? && version.to_i < 3
+          # This lets us know how many targets have Powershell 2, and lets the
+          # user know how many targets they have with PS2
+          msg = "Detected Powershell 2 on one or more targets.\nPowershell 2 "\
+            "is deprecated, and support will be removed in Bolt 3.0. See "\
+            "bolt-debug.log or run with '--log-level debug' to see the full "\
+            "list of targets with Powershell 2."
+
+          Bolt::Logger.deprecation_warning("Powershell 2", msg)
+          @logger.debug("Detected Powershell 2 on #{target}.")
```
beechtom, line 31:
```suggestion
          @logger.debug("Detected PowerShell 2 on #{target}. Support for PowerShell is deprecated.")
```

The debug message on its own doesn't say much currently.

lib/bolt/shell/powershell.rb
```diff
@@ -14,6 +14,22 @@ def initialize(target, conn)
...
+
+      def validate_ps_version
+        version = execute("$PSVersionTable.PSVersion.Major").stdout.string.chomp
+        if !version.empty? && version.to_i < 3
+          # This lets us know how many targets have Powershell 2, and lets the
+          # user know how many targets they have with PS2
+          msg = "Detected Powershell 2 on one or more targets.\nPowershell 2 "\
+            "is deprecated, and support will be removed in Bolt 3.0. See "\
+            "bolt-debug.log or run with '--log-level debug' to see the full "\
+            "list of targets with Powershell 2."
```
beechtom, line 28:
```suggestion
          msg = "Detected PowerShell 2 on one or more targets.\nPowerShell 2 "\
            "is deprecated, and support will be removed in Bolt 3.0. See "\
            "bolt-debug.log or run with '--log-level debug' to see the full "\
            "list of targets with PowerShell 2."
```

lib/bolt/shell/powershell.rb
lucywyman: Oh it's not meant to. Users will see a warning that Powershell 2 is deprecated, but they'll only be warned once if any targets have Powershell 2. The intention is that users will see the warning and look in debug logs for which targets have Powershell 2. They should already know it's deprecated if they're seeing these logs. I wanted to keep the message short to avoid overloading debug output.

lucywyman: I could maybe have it say "Detected unsupported Powershell version 2 on #{target}."

beechtom: I think that would be helpful. Even though there is the warning that gives more context, I think each individual debug message should speak for itself.

beechtom: Also, the check is for < 3, which means PowerShell 1 will also probably trigger this warning. I'm not sure if it's too much of a concern that the warnings say PowerShell 2 though, since 1 is fairly old.

beechtom (approved):

lib/bolt/shell/powershell.rb
lucywyman: *Actually* Powershell one doesn't have `PSVersionTable`, and this won't raise a warning if nothing is returned. Since we don't support Powershell 1 and none of the systems from analytics have Powershell 1 I don't think we should add special handling for it.

Issue #2357, opened 2020-11-17
Rename 'apply_settings' to 'apply-settings'


lucywyman:
## Use Case
We should update the `apply_settings` config option to `apply-settings` to standardize on using hyphens for config settings where possible. This should include updating `bolt project migrate` to update the setting. `apply_settings` should still be valid until we release 3.0, but we should issue a deprecation warning if it's present. If both `apply-settings` and `apply_settings` are present we should ignore `apply_settings` and issue the deprecation warning as usual. If this is set in bolt.yaml we should under the hood update it to use the hyphen, issue the warning, and still use the hyphenated version. If you have both the hyphen will win and we'll still warn.

This setting isn't present in bolt-defaults.yaml so we don't need to worry about that.

PR #2399, opened 2020-11-24
(GH-2357) Deprecate 'apply_settings' in favor of 'apply-settings'

beechtom:
This updates Bolt to issue a deprecation warning when a config file
includes the `apply_settings` option, which has been renamed to
`apply-settings`. When config is set for both of these options, Bolt
will issue another warning that `apply_settings` will be ignored.

!deprecation

* **Deprecate `apply_settings` in favor of `apply-settings`**
  ([#2357](https://github.com/puppetlabs/bolt/issues/2357))

  The `apply_settings` configuration option has been deprecated in favor
  of `apply-settings`.

### DIFF
--- a/lib/bolt/config.rb
+++ b/lib/bolt/config.rb
@@ -276,6 +276,7 @@ module Bolt
       @config_files = []
 
       default_data = {
+        'apply-settings'      => {},
         'apply_settings'      => {},
         'color'               => true,
         'compile-concurrency' => Etc.nprocessors,
@@ -370,7 +371,7 @@ module Bolt
           when *TRANSPORT_CONFIG.keys
             Bolt::Util.deep_merge(val1, val2)
           # Hash values are shallow merged
-          when 'puppetdb', 'plugin-hooks', 'plugin_hooks', 'apply_settings', 'log'
+          when 'puppetdb', 'plugin-hooks', 'plugin_hooks', 'apply-settings', 'apply_settings', 'log'
             val1.merge(val2)
           # All other values are overwritten
           else
@@ -407,8 +408,9 @@ module Bolt
       end
 
       # Filter hashes to only include valid options
-      @data['apply_settings'] = @data['apply_settings'].slice(*OPTIONS['apply_settings'][:properties].keys)
-      @data['puppetfile'] = @data['puppetfile'].slice(*OPTIONS['puppetfile'][:properties].keys)
+      %w[apply-settings apply_settings puppetfile].each do |opt|
+        @data[opt] = @data[opt].slice(*OPTIONS.dig(opt, :properties).keys)
+      end
     end
 
     private def normalize_log(target)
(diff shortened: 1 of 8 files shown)


lib/bolt/config.rb
```diff
@@ -276,6 +276,7 @@ def initialize(project, config_data, overrides = {})
       @config_files = []
 
       default_data = {
+        'apply-settings'      => {},
```
lucywyman, line 279: It seems weird to set both settings and then force users not to, but on the other hand setting the default once we've resolved which one we're using means we'll have to change it again when we remove the option so...I think this is the best way.

lib/bolt/config.rb
```diff
@@ -406,8 +407,9 @@ def deep_clone
       end
 
       # Filter hashes to only include valid options
-      @data['apply_settings'] = @data['apply_settings'].slice(*OPTIONS['apply_settings'][:properties].keys)
-      @data['puppetfile'] = @data['puppetfile'].slice(*OPTIONS['puppetfile'][:properties].keys)
+      %w[apply-settings apply_settings puppetfile].each do |opt|
```
lucywyman, line 410: I like this much better.

lucywyman (changes requested):
This needs to be added to `BOLT_PROJECT_OPTIONS` and `BOLT_OPTIONS`, and be added to bolt project migrate.

beechtom:
Updated the migrator and threw in a test for fun. The new option is already a part of `BOLT_OPTIONS` and `BOLT_PROJECT_OPTIONS` though.

lib/bolt/project_manager/config_migrator.rb
```diff
@@ -72,9 +72,10 @@ def migrate(config_file, project_file, inventory_file, backup_dir)
         data     = Bolt::Util.read_yaml_hash(project_file, 'config')
         modified = false
 
-        if data.key?('plugin_hooks') && !data.key?('plugin-hooks')
-          @outputter.print_action_step("Updating 'plugin_hooks' option to 'plugin-hooks'")
-          data['plugin-hooks'] = data.delete('plugin_hooks')
+        [%w[apply_settings apply-settings], %w[plugin_hooks plugin-hooks]].each do |old, new|
+          next unless data.key?(old) && !data.key?(new)
```
lucywyman, line 76: If users have both should we just delete the old key? It will be ignored anyway.

lucywyman: I thought about it with plugin-hooks and decided it was too destructive, but looking at it again I don't see a good reason *not* to do that.

spec/bolt/project_manager/config_migrator_spec.rb
```diff
@@ -2,73 +2,113 @@
...
 require 'bolt/project_manager/config_migrator'
 
 describe Bolt::ProjectManager::ConfigMigrator do
+  include BoltSpec::Project
+
   def migrate
     migrator.migrate(config_file, project_file, inventory_file, backup_dir)
   end
 
+  let(:inventory_file) { project_dir + 'inventory.yaml' }
```
lucywyman, line 15: Not alphabetized...why the order change? (Just curious, I don't mind it)

lib/bolt/project_manager/config_migrator.rb
beechtom: Updated to remove the deprecated option when both exist.

spec/bolt/project_manager/config_migrator_spec.rb
beechtom: Moved them temporarily while I was updating the tests and just copy/pasted them back into the wrong spot. :P

lucywyman (approved):

Issue #2370, opened 2020-11-18
Example incorrect


woter1832:
## Describe the Change You Would Like
In your example, you say that `out::message` should return an object:
```
[
  {
    "target": "target0",
    "action": "task",
    "object": "package",
    "status": "success",
    "value": {
      "status": "uninstalled",
      "version": ""
    }
  }
]
```

When I try this from Puppet Enterprise, the console reports the error:
```
Plan output

{
  "msg" : "'out::message' parameter 'message' expects a String value, got ResultSet",
  "kind" : "bolt/plan-failure",
  "details" : {
    "class" : "Bolt::PAL::PALError"
  }
}
```
This suggests there are differences or changes that need to be highlighted in the documentation.

The section of documentation I refer to is: https://puppet.com/docs/bolt/latest/writing_plans.html#using-outmessage-to-debug-a-plan

> plan website::test(
>  TargetSpec $targets
> ) {
>   $check_mysql = run_task('package', $targets, "Check for MySQL", {'action' => 'status', 'name' => 'mysql'})
>   out::message($check_mysql)
> }

The specific line: `out::message($check_mysql)`
Thanks

lucywyman:
Hello! That's probably because the version of PE you have is using an older version of Bolt from before we had added outputting objects instead of just strings. You can see [which version of Bolt is running in PE here](https://puppet.com/docs/pe/2019.8/component_versions_in_recent_pe_releases.html#puppet-enterprise-agent-and-server-components), and see the [documentation for debugging using out::message in PE here](https://puppet.com/docs/pe/2019.8/writing_plans_in_puppet_language_pe.html#outputting-to-the-cli-or-console) which indicates to pass `out::message` a string.

woter1832:
@lucywyman ,
Thank you for your reply. You closed it before I had a chance to reply. I am pretty sure this is not the case. I am running the very latest version of PE (2019.8.4).

### Plan
```
plan base_windows::testplan (
  TargetSpec $targets,
  Optional[String] $contents = undef,
  String $filename,
){
  $targets.apply_prep

  $apply_results = apply($targets, '_catch_errors' => true) {
    class { 'base_windows::testpp':
      filename => $filename,
      contents => $contents,
    }
  }

  #return $apply_results

   $apply_results.each |$result| {
    $target = $result.target.name
    #notice( "testplan result: ${target}, ${result}")

    if $result.ok {
      notice("${target} returned a value: ${result.value}")
[...]

PR #2401, opened 2020-11-24
(GH-2370) Add value() function for ApplyResult datatype

beechtom:
This adds a new `value()` function to the `ApplyResult` data type, which
returns a hash with the Puppet report under a `report` key.

!feature

* **Added `value()` function to `ApplyResult` datatype**
  ([#2370](https://github.com/puppetlabs/bolt/issues/2370))

  The `ApplyResult` datatype has a new `value()` function that returns a
  hash that includes the Puppet report from an apply under the `report`
  key.

### DIFF
--- a/lib/bolt/apply_result.rb
+++ b/lib/bolt/apply_result.rb
@@ -96,9 +96,9 @@ module Bolt
       @target = target
       @value = {}
       @action = 'apply'
-      value['report'] = report if report
-      value['_error'] = error if error
-      value['_output'] = metrics_message if metrics_message
+      @value['report'] = report if report
+      @value['_error'] = error if error
+      @value['_output'] = metrics_message if metrics_message
     end
 
     def event_metrics
--- a/bolt-modules/boltlib/lib/puppet/datatypes/applyresult.rb
+++ b/bolt-modules/boltlib/lib/puppet/datatypes/applyresult.rb
@@ -12,6 +12,7 @@ Puppet::DataTypes.create_type('ApplyResult') do
       message => Callable[[], Optional[String]],
       action => Callable[[], String],
       to_data => Callable[[], Hash],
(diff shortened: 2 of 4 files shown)


lib/bolt/apply_result.rb
```diff
@@ -96,9 +96,9 @@ def initialize(target, error: nil, report: nil)
       @target = target
       @value = {}
       @action = 'apply'
-      value['report'] = report if report
-      value['_error'] = error if error
-      value['_output'] = metrics_message if metrics_message
+      @value['report'] = report if report
```
lucywyman, line 99: Oh lol so this was...really more of a bug?

beechtom: I guess. :P The implementation class already had the `value` function since it inherits from `Result`. Just needed to expose it on the Puppet side.

lucywyman (changes requested):
This needs to have an accessor in `lib/bolt/apply_result.rb` - it currently isn't available:
```
lucy ~/githubs/bolt (GH-2313) $ be bolt plan run bolt::apply -t localhost
Unknown option 'plugin-cache' at /home/lucy/githubs/bolt/bolt-project.yaml.
Starting: plan bolt::apply
Starting: install puppet and gather facts on localhost
Finished: install puppet and gather facts with 0 failures in 2.51 sec
Finished: apply catalog with 0 failures in 4.57 sec
Finished: plan bolt::apply in 7.27 sec
{
  "kind": "bolt/pal-error",
  "msg": "Evaluation Error: Unknown function: 'value'. (file: /home/lucy/githubs/bolt/plans/apply.pp, line: 10, column: 23)",
  "details": {
    "file": "/home/lucy/githubs/bolt/plans/apply.pp",
    "line": 10,
    "column": 23
  }
}
```

PR #2402, opened 2020-11-25
(GH-2379) Suggest noexec may be an issue with exit code 126

### LINKED ISSUE #2379: Suggest `noexec` may be an issue if the task is not executable
## Use Case
Users sometimes have tempdirs that have `noexec` configured, meaning when we copy the task executable there it cannot be executed. We should catch executable errors and suggest that `noexec` may be the issue, and point to docs instructing users to set `tmpdir` to avoid the issue.

Current error message:
```
The task failed with exit code 126 and no stdout but stderr contained: 
.... <temp path to task>>.rb: Permission denied
```
Make sure we only do this for 'script' and 'task', and catch the issue before making a result object and modify the message (stderr) to include the suggestion. This should happen in the shell connections.

Docs: Add a section to troubleshooting and quote the new error message to improve the chances that someone gets to it from Google.

lucywyman:
Exit code 126 indicates that a file Bolt has written to the temporary
directory can't be executed because the user Bolt is running as doesn't
have permission to execute it. This can be caused by a variety of
issues, but one common cause is that administrators have mounted the
default tmpdir with `noexec`, meaning that users can't execute files in
the tmpdir. The recommended workaround for this is to configure an
alternative `tmpdir` for the transport(s) being used, or to contact the
administrator about modifying the permissions (though usually they're in
place for a reason). This appends a brief message that `noexec` may be
the issue to stderr output whenever exit code 126 is returned. This is
appended for bash shells. The message directs users
to our troubleshooting doc, which has been updated to include a note
about `noexec` and it's mitigation.

Closes #2379

!no-release-note

### DIFF
--- a/documentation/troubleshooting.md
+++ b/documentation/troubleshooting.md
@@ -36,6 +36,18 @@ system. For example, if you write a Python task and include the line:
 `#!/usr/bin/env python`, Bolt attempts to execute the script using the default
 `python` executable on the target system.
 
+If your task fails with the following error, the issue may be that your temporary directory (tmpdir)
+is [mounted with noexec](https://superuser.com/questions/728127/what-does-noexec-flag-mean-when-mounting-directories-on-rhel).
+
+```
+The task failed with exit code 126 and no stdout but stderr contained: 
+.... <temp path to task>>.rb: Permission denied
+```
+
+You can resolve this by [configuring an alternate tmpdir](bolt_transports_reference.md) for the
+transport you're using, or by talking to your administrator about updating permissions for the
+directory.
+
 ## Bolt can't connect to my hosts over SSH
 
 ### Host key verification failures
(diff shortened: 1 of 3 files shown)


beechtom (approved):

lib/bolt/shell/powershell.rb
```diff
@@ -300,6 +300,12 @@ def execute(command)
         stderr.join
         result.exit_code = t.value.respond_to?(:exitstatus) ? t.value.exitstatus : t.value
 
+        if result.exit_code == 126
```
nicklewis, line 303: I don't think this applies to powershell

lucywyman: D'oh. Thanks Nick

hestonhoffman (approved):

Issue #2403, opened 2020-11-25
Windows Bolt with apply block in plans fails silently and logs "Something's gone terribly wrong"


kev-in-shu:
## Describe the Bug
Running a plan with apply block, from a Windows Bolt Controller, fails silently .
There is only an INFO log entry appearing in bolt-debug.log

```
2020-11-25T09:05:21.567312 INFO   [] [Bolt::Executor] {"message":"Something's gone terribly wrong! STDERR is logged."}
```

## Expected Behavior
Apply blocks that work perfectly fine when run from a Linux Bolt controller should work the same way on Windows.

Also, regardless of the issue, if there is an error that's gone terribly wrong, Bolt should show the error instead of hiding it in the bolt-debug.log file.

## Steps to Reproduce
1. Create a bolt-project.yaml file with the following piece of yaml
```
---
name: boltcode
```
2. Create a file named plans/test_apply.pp with the following piece of code
```
plan boltcode::test_apply(
  TargetSpec $targets
) {
  apply($targets) {
    file {'c:/tempapplyblock':
      ensure => file
    }
  }
}
```
3. Run the plan on a target
4. Check that the standard output of Bolt shows no sign of errors like "apply catalog with 0 failures"
5. Check on the target that no resource was applied
6. Check in bolt-debug.log the "Something's gone terribly wrong" message

## Environment
 - Version: 2.35.0
 - Platform: Windows Bolt controller
 - Platform: Windows or Linux target

## Additional Context

- Debugging the issue shows that libexec/bolt_catalog writes warnings in its standard output (see below). For this reason, the resulting JSON cannot be parsed.

```
E, [2020-11-25T09:04:53.025679 #31348] ERROR -- : Facter::InternalFactManager - "\xE9" to UTF-8 in conversion from ASCII-8BIT to UTF-8 to UTF-16LE
```

- The above warning about UTF conversion also appears systematically at the start of every bolt command run from a Windows Bolt Controller
  - So far, I saw that this warning about UTF conversion disappears when I delete the bolt-project.yaml... But I could not run my plan under these conditions

lucywyman:
@kev I'm not able to reproduce this issue, can you post a (redacted) version of your plan, and the output from Bolt? It seems to me like something else is going wrong and causing an undiagnosed error. I could be wrong and if Puppet is returning useless information at stderr we'll probably want to figure out what is being returned and remove it rather than just ignore it.

[...]

PR #2405, opened 2020-11-26
(GH-2403) Fix apply block on Windows

kev-in-shu:
On Windows, bolt plans fail to run apply blocks, because of a bug that makes bolt_catalog start its outputs with unwanted warnings about encoding issues.

This PR fixes https://github.com/puppetlabs/bolt/issues/2403


lucywyman:
I don't think we want to ignore non-JSON output on stdout since it generally means something has gone wrong (in this case it seems like an error was raised). Closing.

PR #2408, opened 2020-11-30
(PE-30638) Raise PluginNotSupported error for unsupported plugins

donoghuc:
Update bolt-servers `project_inventory_targets` to raise informative errors when unsupported or unknown plugins are encountered in a project's inventoryfile.

### DIFF
--- a/lib/bolt_server/transport_app.rb
+++ b/lib/bolt_server/transport_app.rb
@@ -8,6 +8,7 @@ require 'bolt/inventory'
 require 'bolt/project'
 require 'bolt/target'
 require 'bolt_server/file_cache'
+require 'bolt_server/plugin'
 require 'bolt/task/puppet_server'
 require 'json'
 require 'json-schema'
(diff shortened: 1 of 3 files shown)


lib/bolt_server/transport_app.rb
```diff
@@ -647,16 +655,23 @@ def make_winrm_target(target_hash)
...
+           context[:config].inventoryfile
+          raise Bolt::ValidationError, "Project inventory must be defined in the " \
+            "inventory.yaml file at the root of the project directory"
+        end
+        plugins = Bolt::Plugin.setup(context[:config], context[:pal], load_plugins: false)
+        begin
+          inventory = Bolt::Inventory.from_config(context[:config], plugins)
+          target_list = inventory.get_targets('all').map { |targ| targ.to_h.merge({ 'transport' => targ.transport }) }
+        rescue Bolt::Plugin::PluginError::PluginLoadingDisabled => e
+          raise Bolt::PluginNotSupported.new(e.msg, 'stub')
```
donoghuc, line 670: @ekinanp This is where i would use the structured data i asked about in https://github.com/puppetlabs/bolt/pull/2394/files#r532886010

donoghuc:
Just pushed this to collaborate with Enis on 30638. Will rebase and clean up when that is merged. Added DMN for now.

lib/bolt_server/transport_app.rb
ekinanp: Yup, I updated my PR to include `plugin_name` in `details`.

lib/bolt/error.rb
```diff
@@ -166,4 +166,10 @@ def initialize(msg, path)
       super(msg, 'bolt/file-error', { "path" => path })
     end
   end
+
+  class PluginNotSupported < Bolt::Error
```
ekinanp, line 170: Let's put this in `lib/bolt-server` for now since there is currently no way for Bolt's plugin code to raise this error.

lib/bolt_server/transport_app.rb
```diff
@@ -647,16 +655,23 @@ def make_winrm_target(target_hash)
...
-      inventory = Bolt::Inventory.from_config(bolt_config, plugins)
-      target_list = inventory.get_targets('all').map { |targ| targ.to_h.merge({ 'transport' => targ.transport }) }
+      in_bolt_project(params['project_ref']) do |context|
+        if context[:config].inventoryfile &&
+           context[:config].project.inventory_file.to_s !=
+           context[:config].inventoryfile
+          raise Bolt::ValidationError, "Project inventory must be defined in the " \
+            "inventory.yaml file at the root of the project directory"
+        end
+        plugins = Bolt::Plugin.setup(context[:config], context[:pal], load_plugins: false)
```
ekinanp, line 665: Ah, I think we'll want to put `Plugin.setup` in the `begin/rescue` too because that also resolves some plugin references. Bit subtle, but chain is `setup` => `by_name` => `add_module_plugin` => `config_for_plugin`. See https://github.com/puppetlabs/bolt/blob/main/lib/bolt/plugin.rb#L122.

ekinanp: May be worth having a test case for "config with plugin references" to capture that use-case too? On top of the existing "inventory file with plugin references" test cases.

ekinanp (approved):

Issue #2098, opened 2020-08-19
Add scripts to projects


adreyer:
## Use Case
In many cases adding a script to a project is a better choice than converting it to a task or wrapping it in a plan. For example, if the script takes no arguments or simple arguments that cannot be easily validated users will be better served by simply adding the script to a project and exposing it to users.

## Describe the Solution You Would Like
When scripts are added to a project they should be runnable with `bolt script run`.
`bolt script show` should list the scripts in a project.

## Questions
- Can scripts live in modules or only in a project? They should probably live in modules so they can be run from environments.
- Should we add a `scripts` key to bolt-project.yaml similar to `tasks` and `plans`? Yes
- Where should scripts live in the project/module structure? `scripts/` `files/scripts`? A configured directory?
  Running scripts from the CLI should have the same interface as running them from a plan (ie. we can't support `modulename/script` syntax on the CLI for scripts nested under `files`).
- How does `bolt script run` differentiate between a path and a script from the project?

PR #2410, opened 2020-12-01
(bug) Add 'notice' log level to schema and deprecate

beechtom:
This adds the `notice` log level back to the config schema, which was
removed in #2098. Because we now use the schema for config validation,
the absence of `notice` as a valid option resulted in Bolt raising an
error for any user that had that log level configured. Now, Bolt will
instead issue a deprecation warning that the log level is deprecated and
will be removed in 3.0.

!bug

* **Support `notice` log level**

  Log levels can now be set to `notice`. Previously, Bolt would raise an
  error saying that `notice` was not a supported log level.

!deprecation

* **Deprecate `notice` log level**

  The `notice` log level is deprecated and will be removed in Bolt 3.0.
  Use the `info` log level instead.

### DIFF
--- a/rakelib/docs.rake
+++ b/rakelib/docs.rake
@@ -176,6 +176,10 @@ begin
       # 'console' and filepath
       @opts['log'][:properties] = @opts['log'][:additionalProperties][:properties]
 
+      # Remove 'notice' log level. This is soft-deprecated and shouldn't appear
+      # in documentation.
+      @opts['log'][:properties]['level'][:enum].delete('notice')
+
       # Stringify data types
       @opts.transform_values! { |data| stringify_types(data) }
       @inventory.transform_values! { |data| stringify_types(data) }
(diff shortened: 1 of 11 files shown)


rakelib/docs.rake
```diff
@@ -200,6 +204,14 @@ begin
       inventory   = Bolt::Config::INVENTORY_OPTIONS.dup
       @transports = Bolt::Config::TRANSPORT_CONFIG.keys
 
+      # Move sub-options for 'log' option up one level, as they're nested under
+      # 'console' and filepath
+      @opts['log'][:properties] = @opts['log'][:additionalProperties][:properties]
+
+      # Remove 'notice' log level. This is soft-deprecated and shouldn't appear
+      # in documentation.
```
lucywyman, line 212: While we're here, can we update https://github.com/puppetlabs/bolt/blob/main/documentation/templates/bolt_defaults_reference.md.erb#L9 to have 1 space instead of 2?

rakelib/docs.rake
```diff
@@ -200,6 +204,14 @@ begin
       inventory   = Bolt::Config::INVENTORY_OPTIONS.dup
       @transports = Bolt::Config::TRANSPORT_CONFIG.keys
 
+      # Move sub-options for 'log' option up one level, as they're nested under
+      # 'console' and filepath
+      @opts['log'][:properties] = @opts['log'][:additionalProperties][:properties]
```
lucywyman, line 209: I don't see the difference in generated outputs for this? I generated the doc on `main` and on this branch, and the results were the same...

beechtom: Run `rake docs:defaults_reference`. The template pulls from the `:properties` key, without this change running that task on its own will show `console` instead of `level` and `append` in the docs.

This change isn't needed when running `rake docs:all` since `docs:config_reference` modifies the options hash first.

lucywyman: Ah I see. Ok.

PR #2416, opened 2020-12-03
(bug) Validate groups are an Array

donoghuc:
After accidentally declaring a single element `group` as a Hash instead of a single element Array, The stack trace that resulted was difficult to troubleshoot. This commit adds validation that groups are always arrays.

!bug

* **Validate inventory groups are Arrays**
    Raise a validation error when a group resolves to a data structure other than an Array instead.

### DIFF
--- a/lib/bolt/inventory/group.rb
+++ b/lib/bolt/inventory/group.rb
@@ -68,7 +68,7 @@ module Bolt
         # though, since that will resolve any nested references and we want to
         # leave it to the group to do that lazily.
         groups = @plugins.resolve_top_level_references(groups)
-
+        raise ValidationError.new("Groups must be an Array", nil) unless groups.is_a?(Array)
         @groups = Array(groups).map { |g| Group.new(g, plugins) }
       end
 
--- a/spec/bolt/inventory/group_spec.rb
+++ b/spec/bolt/inventory/group_spec.rb
@@ -469,6 +469,11 @@ describe Bolt::Inventory::Group do
       data['plugin_hooks'] = 'puppet_library'
       expect { group }.to raise_error(/Expected plugin_hooks to be of type Hash/)
     end
+
+    it 'fails if plugin_hooks is not a hash' do
+      data['groups'] = { 'forgot' => 'to make this a single element list' }
+      expect { group }.to raise_error(/Groups must be an Array/)
+    end
   end
 
   describe 'with aliases' do


donoghuc:
With an inventory like:
```
groups:
  name: ssh
  targets:
    - name: nix_target
      uri: example.com
  config:
    transport: ssh
    ssh:
      user: root
      password: password
```

This was hard to diagnose:
```
NoMethodError: undefined method `key?' for ["name", "ssh"]:Array
  /home/cas/working_dir/bolt/lib/bolt/inventory/group.rb:27:in `initialize'
  /home/cas/working_dir/bolt/lib/bolt/inventory/group.rb:72:in `new'
  /home/cas/working_dir/bolt/lib/bolt/inventory/group.rb:72:in `block in initialize'
  /home/cas/working_dir/bolt/lib/bolt/inventory/group.rb:72:in `map'
  /home/cas/working_dir/bolt/lib/bolt/inventory/group.rb:72:in `initialize'
  /home/cas/working_dir/bolt/lib/bolt/inventory/inventory.rb:25:in `new'
  /home/cas/working_dir/bolt/lib/bolt/inventory/inventory.rb:25:in `initialize'
  /home/cas/working_dir/bolt/lib/bolt/inventory.rb:87:in `new'
  /home/cas/working_dir/bolt/lib/bolt/inventory.rb:87:in `create_version'
  /home/cas/working_dir/bolt/lib/bolt/inventory.rb:77:in `from_config'
  /home/cas/working_dir/bolt/lib/bolt/cli.rb:62:in `inventory'
  /home/cas/working_dir/bolt/lib/bolt/cli.rb:410:in `execute'
  /home/cas/working_dir/bolt/exe/bolt:11:in `<top (required)>'
  /home/cas/working_dir/bolt/vendor/bundle/ruby/2.5.0/bin/bolt:23:in `load'
  /home/cas/working_dir/bolt/vendor/bundle/ruby/2.5.0/bin/bolt:23:in `<top (required)>'
```

lib/bolt/inventory/group.rb
```diff
@@ -68,7 +68,7 @@ def initialize(input, plugins)
         # though, since that will resolve any nested references and we want to
         # leave it to the group to do that lazily.
         groups = @plugins.resolve_top_level_references(groups)
-
+        raise ValidationError.new("Groups must be an Array", nil) unless groups.is_a?(Array)
```
nicklewis, line 71: Interesting. The call to `Array(groups)` below this would seem to indicate this doesn't _have_ to be an array. I think the problem is actually that passing a hash to `Array()` actually converts the hash to an array of key/value pairs rather than wrapping it in an array. So maybe `Array(groups)` should be `groups = [groups] unless groups.is_a?(Array)`.

nicklewis: I was under the impression that `Array(...)` _always_ wrapped the argument in an array if it wasn't one, but I guess that's not the case. I wonder how many other places in Bolt this same bug arises from the same incorrect assumption.

nicklewis: It appears there will be a similar bug for individual targets. https://github.com/puppetlabs/bolt/pull/2416/files#diff-297edc81f841f0d72ea66094838373e2eae905ddfcf1335e5f020411b3e615deR48

beechtom:
@donoghuc I think this needs a rebase to pick up some changes to CI.

lib/bolt/inventory/group.rb
donoghuc: Yeah, this is a bigger issue than I originally expected. Ticketed for Bolt team https://github.com/puppetlabs/bolt/issues/2430

donoghuc:
This is an incomplete solution. Filed a proper bug ticket. https://github.com/puppetlabs/bolt/issues/2430

PR #2418, opened 2020-12-03
Specify puppet6 collection when defaulting version to 6.2.0

nicklewis:
These tests set the default puppet version for the puppet_library hook
to 6.2.0 but don't set the default collection to puppet6 accordingly.
Now that the default collection 'puppet' points to puppet7, installing
the package fails. We now set the collection to match the version
explicitly.

!no-release-note


beechtom:
Closed in favor of #2419

PR #2419, opened 2020-12-03
(maint) Fix broken apply integration tests

beechtom:
This includes several changes to Bolt's spec tests and CI environment:

- **Fix broken apply integration tests**

  With the release of Puppet 7, the `puppet_agent::install` task now
  installs `puppet7` packages by default. This broke several of the
  apply integration tests that expected `puppet6` packages, or attempted
  to downgrade to the `puppet6` repo after having already installed the
  `puppet7` repo. This fixes the tests to assume `puppet7` is installed.

  Additionally, the apply integration tests were refactored slightly to
  make it easier to tell what is happening in setup, testing, and
  teardown for each spec test. Previously, the agent package was
  repeatedly installed and uninstalled between spec tests and inventory
  was sourced in multiple different ways. Now, agentful targets have the
  agent installed before all spec tests are run, and then the agent is
  uninstalled after the tests have all run. Additionally, inventory is
  sourced in one way for each group of spec tests.

- **Add puppet_7_node container**

  This adds a new `puppet_7_node` container that is used in apply
  integration testing and will have the agent package installed from the
  `puppet7` repo.

- **Update containers to Ubuntu 18.04**

  With Ubuntu 16.04 reaching EOL in 2021, I took the opportunity to
  update the containers to an Ubuntu 18.04 image.

- **Update gems when restoring gem cache**

  Bolt's CI caches installed gems after running `bundle install` in
  order to speed up setup for testing. However, this occasionally causes
  issues when running `bundle exec` and a newer version of a gem is
  expected than what is installed. To get around this, a `bundle update`
  step is now run whenever the cache is hit, ensuring gems are
  up-to-date.

!no-release-note

### DIFF
--- a/spec/integration/apply_spec.rb
+++ b/spec/integration/apply_spec.rb
@@ -4,178 +4,93 @@ require 'spec_helper'
 require 'bolt_spec/conn'
 require 'bolt_spec/files'
 require 'bolt_spec/integration'
+require 'bolt_spec/project'
 require 'bolt_spec/puppet_agent'
 require 'bolt_spec/run'
 
-describe "apply", expensive: true do
+TEST_VERSIONS = [
+  [5, 'puppet5'],
+  [6, 'puppet6'],
+  [7, 'puppet']
+].freeze
+
+describe 'apply', expensive: true do
   include BoltSpec::Conn
-  include BoltSpec::Files
   include BoltSpec::Integration
+  include BoltSpec::Project
   include BoltSpec::PuppetAgent
   include BoltSpec::Run
 
-  let(:modulepath) { File.join(__dir__, '../fixtures/apply') }
-  let(:hiera_config) { File.join(__dir__, '../fixtures/configs/empty.yml') }
-  let(:config_flags) { %W[--format json --targets #{uri} --password #{password} --modulepath #{modulepath}] + tflags }
-  let(:apply_settings) { { 'show_diff' => true } }
+  let(:project_config) do
+    {
+      'apply_settings' => { 'show_diff' => true },
+      'hiera-config'   => File.join(__dir__, '../fixtures/configs/empty.yml'),
+      'modulepath'     => File.join(__dir__, '../fixtures/apply')
+    }
+  end
+
+  # The following are run with both *nix and Windows targets.
+  shared_examples 'agentful tests' do |targets|
+    it 'runs a ruby task' do
+      results = run_cli_json(%W[task run basic::ruby_task -t #{targets}], project: project)
+
+      results['items'].each do |result|
+        expect(result).to include('status' => 'success')
+        expect(result['value']).to eq('ruby' => 'Hi')
+      end
+    end
+
+    it 'runs an apply plan' do
+      results = run_cli_json(%W[plan run basic::notify -t #{targets}], project: project)
+
+      results.each do |result|
+        expect(result).to include('status' => 'success')
+        expect(result.dig('value', 'report', 'resource_statuses')).to include("Notify[Apply: Hi!]")
+      end
+    end
+
+    it 'succeeds with an empty hiera config' do
+      results = run_cli_json(%W[plan run prep -t #{targets}], project: project)
+
+      results.each do |result|
+        expect(result['status']).to eq('success')
+        expect(result.dig('value', 'report', 'resource_statuses')).to include(/Notify\[Hello .*\]/)
+      end
+    end
+  end
 
   describe 'over ssh', ssh: true do
-    let(:uri) { conn_uri('ssh') }
-    let(:user) { conn_info('winrm')[:user] }
+    let(:uri)      { conn_uri('ssh') }
+    let(:user)     { conn_info('ssh')[:user] }
     let(:password) { conn_info('ssh')[:password] }
-    let(:tflags) { %W[--no-host-key-check --run-as root --sudo-password #{password}] }
 
-    def root_config
-      { 'modulepath' => File.join(__dir__, '../fixtures/apply') }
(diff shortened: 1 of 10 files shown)


spec/integration/apply_spec.rb
```diff
@@ -289,7 +289,7 @@ def task_plugin_inventory
     context "with a puppet_agent installed" do
       before(:all) do
         # Deferred must use puppet >= 6
-        target = 'puppet_6'
+        target = 'ubuntu_node'
```
lucywyman, line 292: I think we actually want to install the agent on the `puppet_6` node and verify against that target. Using the `agent_version_inventory` it should pick up the correct config.

lucywyman: I think we want to set `uri` to the puppet_6 node. I kind of think the easiest way to do that is to at the top have:
```
def puppet_6
  @puppet_6 ||= docker_inventory(root: true).get_target('puppet_6_node')
end
```
and then set the uri to `puppet_6` the 'with puppet_agent' context. But you could also just add 'puppet_6' and puppet_5 functions to the conn helper that do the same thing.

spec/integration/apply_spec.rb
```diff
@@ -4,266 +4,288 @@
...
-        result = run_task('puppet_agent::version', 'puppet_5', {}, inventory: agent_version_inventory)
-        expect(result.count).to eq(1)
-        expect(result[0]).to include('status' => 'success')
-        expect(result[0]['value']['version']).to match(/^5/)
+      # Uninstall agents after running all tests.
+      after(:all) do
+        TEST_VERSIONS.each do |version,|
+          uninstall("puppet_#{version}_node", inventory: docker_inventory(root: true))
+        end
+      end
```
beechtom, line 82: Dunno if we _really_ need to do this, but left it in since it doesn't seem to hurt. 🤷

spec/integration/apply_spec.rb
```diff
@@ -4,266 +4,288 @@
...
   include BoltSpec::Integration
+  include BoltSpec::Project
   include BoltSpec::PuppetAgent
   include BoltSpec::Run
 
-  let(:modulepath) { File.join(__dir__, '../fixtures/apply') }
-  let(:hiera_config) { File.join(__dir__, '../fixtures/configs/empty.yml') }
-  let(:config_flags) { %W[--format json --targets #{uri} --password #{password} --modulepath #{modulepath}] + tflags }
-  let(:apply_settings) { { 'show_diff' => true } }
+  let(:project_config) do
```
lucywyman, line 24: We should pass this transparently to `with_project` so it's not appearing magically.

spec/integration/apply_spec.rb
lucywyman: Hm, I don't think so - mostly because 1. It'd be annoying if there was an unrelated error uninstalling the agents that caused them to fail, 2. These containers are named such and explicitly intended to be containers with the agent installed, and for local development it'd be annoying if a test expected the agent to be installed and it wasn't.

spec/integration/apply_spec.rb
```diff
@@ -4,266 +4,288 @@
...
+      around(:each) do |example|
+        with_project do
+          File.write(project.inventory_file, docker_inventory(root: true).to_yaml)
+          example.run
+        end
+      end
 
-        result = run_task('puppet_agent::version', 'puppet_6', {}, inventory: agent_version_inventory)
-        expect(result.count).to eq(1)
+      include_examples 'agentful tests', 'agent_targets'
```
lucywyman, line 93: While we're here can we rename this to `nix_agents`?

spec/integration/apply_spec.rb
```diff
@@ -4,266 +4,288 @@
...
+    }
+  end
 
-  describe 'over ssh', ssh: true do
-    let(:uri) { conn_uri('ssh') }
-    let(:user) { conn_info('winrm')[:user] }
-    let(:password) { conn_info('ssh')[:password] }
-    let(:tflags) { %W[--no-host-key-check --run-as root --sudo-password #{password}] }
+  # The following are run with both *nix and Windows targets.
+  shared_examples 'agentful tests' do |targets|
```
lucywyman, line 33: It seems like we lost the `does not create project` test?

spec/integration/apply_spec.rb
```diff
@@ -4,266 +4,288 @@
...
-      it 'runs an apply plan' do
-        with_tempfile_containing('inventory', YAML.dump(agent_version_inventory), '.yaml') do |inv|
-          results = run_cli_json(%W[plan run basic::notify --targets agent_targets
-                                    --modulepath #{modulepath} --inventoryfile #{inv.path}])
-          results.each do |result|
-            expect(result['status']).to eq('success')
-            report = result['value']['report']
-            expect(report['resource_statuses']).to include("Notify[Apply: Hi!]")
-          end
+      it "applies a manifest" do
```
lucywyman, line 150: I think it's still useful to wrap this in a `bolt apply command` context so we can better pinpoint failures.

lucywyman (commented):
This change looks great! A few small comments then excited to get this in

spec/integration/apply_spec.rb
beechtom: I actually meant to ask about that test... I'm not really sure why we have that test to begin with or if it's providing any value. As far as I'm aware, installing the agent will never create `~/.puppetlabs` on a target.

lucywyman: Hm I think I'm ok with that, since I'd be surprised if it also made a bolt directory there.

PR #2420, opened 2020-12-03
(PE-30656) Allow project_inventory_targets to use PuppetConnectData

donoghuc:
This commit updates the `project_inventory_targets` endpoint to accept data from puppet connect and uses it to enable the connect plugin.

### DIFF
--- a/developer-docs/bolt-api-servers.md
+++ b/developer-docs/bolt-api-servers.md
@@ -380,7 +380,25 @@ This endpoint behaves identically to the /ssh/check_node_connections endpoint, b
 
 This endpoint parses a project inventory and returns a list of target hashes. Note that the only accepetable inventoryfile location is the default `inventory.yaml` at the root of the project.
 
+#### Query parameters
+
 - `project_ref`: String, *required* - Reference to the bolt project (in the form [PROJECT NAME]\_[REF])
+
+#### POST body
+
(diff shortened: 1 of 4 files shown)


lib/bolt_server/schemas/connect-data.json
```diff
@@ -0,0 +1,19 @@
...
+  "title": "project_inventory_targets connect plugin data",
+  "description": "POST project_inventory_targets connect plugin data",
+  "type": "object",
+  "properties": {
+    "puppet_connect_data": {
+      "type": "object",
+      "properties": {
+        "type": "object",
+        "properties": {
+          "key": {}
```
donoghuc, line 12: Should we enforce this as a string here?

developer-docs/bolt-api-servers.md
```diff
@@ -380,8 +380,26 @@ This endpoint behaves identically to the /ssh/check_node_connections endpoint, b
 
 This endpoint parses a project inventory and returns a list of target hashes. Note that the only accepetable inventoryfile location is the default `inventory.yaml` at the root of the project.
 
+### Query parameters
```
barriserloth, line 383: This is a nitpick, but these new sections looks like they should be subsections of the `POST /project_inventory_targets` section, so maybe these headers should be the level below the proj_inv_targets header so they don't all look like separate sections?

barriserloth (approved):
Besides the one small comment, this LGTM!

beechtom:
@donoghuc I think this needs a rebase to pick up changes in CI.

PR #2425, opened 2020-12-04
(GH-2400) Add bundled-ruby option for local transport

### LINKED ISSUE #2400: New local Transport configuration option: "use-localhost-magic"
## Use Case

From inventory, I want to be able to dynamically configure additional `local` Transport Targets with the same "magical" settings as the built-in `localhost` Target get.

A new local [Transport configuration](https://puppet.com/docs/bolt/latest/bolt_transports_reference.html) option: **`use-localhost-magic`**

* This suggestion came from @adreyer [over in the #bolt Slack](https://puppetcommunity.slack.com/archives/C7GP57T50/p1606237744029800) (:sparkles:thanks!:sparkles:). 
*  Setting this option would configure a `local` Transport with the same options & features as the `localhost` Target magically gets.  That includes:
   * Dynamically taking care of any platform-dependent settings
   * Setting interpreters to use the Bolt package's built-in executables, when available.

* Currently, the only way to do something like this is from inside a Plan's logic, where you can get_target('localhost') and copy its config directly into new Target objects.
  * But it would be much nicer (and more portable) to take care of that within the inventory so everything can use it.

* I initially considered a Ruby API bolt plugin (like the built-in [RUBY_PLUGINS](https://github.com/puppetlabs/bolt/blob/main/lib/bolt/plugin.rb#L134) such as [env_var](https://github.com/puppetlabs/bolt/blob/main/lib/bolt/plugin/env_var.rb), [prompt](https://github.com/puppetlabs/bolt/blob/main/lib/bolt/plugin/prompt.rb), etc.,) to return the `localhost` Target's configuration.  
   * However, there isn't a mechanism for integrating third-party Ruby API bolt plugins.
   * Even if implemented, the actual `_plugin` usage would probably be quite cumbersome.

## Additional Context

Slack discussion at https://puppetcommunity.slack.com/archives/C7GP57T50/p1606237461023300

lucywyman:
This introduces the `bundled-ruby` config option for the local
transport. The option enables merging default config values currently
used just for a target with name `localhost`. Those default config
values are:

```
config:
  local:
    interpreters:
      '.rb' => RbConfig.ruby # This is Bolt's bundled Ruby
features:
  - puppet-agent
```

The default config values are merged at the target config level, meaning
they are overridden by target-level config but not group-level config.
The 'features' array is appended, not overridden, and is a set so the
value will not be added twice.  The `localhost` target additionally
defaults to using the `local` transport as part of the "magic config",
however the `bundled-ruby` option will only apply to targets that are
already configured to use the local transport.

`bundled-ruby` currently defaults to false, but will default to `true`
in Bolt 3.0. A warning is issued for users who don't set the
conifguration and use the local transport that the default config will
be updated.

Closes #2400

!feature

* **bundled-ruby local transport config option enables local defaults** ((#2400)[https://github.com/puppetlabs/bolt/issues/2400)

  Set `bundled-ruby` in the local transport config to enable or disable
  the default config currently used for the 'localhost' target.

### DIFF
--- a/lib/bolt/transport/local.rb
+++ b/lib/bolt/transport/local.rb
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 
+require 'bolt/logger'
 require 'bolt/transport/simple'
 
 module Bolt
@@ -10,6 +11,18 @@ module Bolt
       end
 
       def with_connection(target)
+        if target.transport_config['bundled-ruby'] || target.name == 'localhost'
+          target.set_local_defaults
+        end
+
+        if target.name != 'localhost' &&
+           !target.transport_config.key?('bundled-ruby')
+          msg = "The local transport will default to using Bolt's Ruby interpreter and "\
+            "setting the 'puppet-agent' feature in Bolt 3.0. Enable or disable these "\
+            "defaults by setting 'bundled-ruby' in the local transport config."
+          Bolt::Logger.warn_once('local default config', msg)
+        end
+
         yield Connection.new(target)
       end
     end
--- a/documentation/developer_updates.md
+++ b/documentation/developer_updates.md
@@ -53,6 +53,24 @@ compiled a list of expected changes and removals.
   should instead use the built-in [PowerShell
   cmdlets](bolt_cmdlet_reference.md).
 
+- **New default configuration for the local transport**
+
+  In Bolt 2.x, Bolt applied the following settings to targets named `localhost` by default:
+  ```
+  targets:
+    - name: localhost
+      config:
+        transport: local
+        local:
+          interpreters:
+            .rb: RbConfig.ruby # This uses the Ruby we ship with Bolt
+      features:
+        - puppet-agent
+  ```
+  These settings will now be applied to all targets using the local transport. Settings can be
+  overridden at the target level in inventory. They can also be enabled or disabled starting in Bolt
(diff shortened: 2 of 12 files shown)


lucywyman:
One fun edge case with this is that `localhost` default config cannot be unset by the `bundled-ruby` config option - it still needs to be overridden at the target config level. That should change in Bolt 3.0, but until then it's the exception to the `bundled-ruby` option.

lib/bolt/transport/local.rb
```diff
@@ -10,6 +11,18 @@ def connected?(_target)
       end
 
       def with_connection(target)
+        if target.transport_config['bundled-ruby'] || target.name == 'localhost'
+          target.set_local_defaults
+        end
+
+        if target.name != 'localhost' &&
+           !target.transport_config.key?('bundled-ruby')
+          msg = "The local transport will default to using Bolt's ruby interpreter and "\
```
beechtom, line 20:
```suggestion
          msg = "The local transport will default to using Bolt's Ruby interpreter and "\
```

beechtom (approved):
Built packages and confirmed it works as expected. 👍  Just one small nitpick on a warning message.

beechtom (approved):

hestonhoffman (approved):

PR #2432, opened 2020-12-04
(feature) Print spinner to the console for long operations

### LINKED ISSUE #606: Running 'bolt puppetfile install' should tell the user what is happening
As a user with a large Puppetfile, I need to know whether or not the command is hanging or I'm just being impatient.

Having the output simply clear the first output line and use a spinner or something would be great.

lucywyman:
(feature) Print spinner to the console for long operations

This introducs a spinner which spins while Bolt performs long
operations. Currently the spinner is added for installing modules,
running plans, and running tasks.

The spinner works by creating a new thread and continuously printing a
series of messages, where each message includes a backspace and the 
next character from an array of characters. The spinner is a function of
the outputters, which each have their own `start_spin` and `stop_spin`
functions that open and close the spinner thread respectively.

The spinner is enabled by default, and can be disabled by setting
`spinner: false` in any of `bolt.yaml`, `bolt-project.yaml`, or
`bolt-default.yaml`.

Closes #606

!feature

* **CLI spinner for long running operations**

  Bolt now has a spinner printed to the CLI for long-running operations,
  so that users know the Bolt process has not hung. Disable the spinner
  by setting `spinner: false` in any Bolt configuration file.

### DIFF
--- a/lib/bolt/cli.rb
+++ b/lib/bolt/cli.rb
@@ -538,12 +538,15 @@ module Bolt
               validate_file('script', script)
               executor.run_script(targets, script, options[:leftovers], executor_opts)
             when 'task'
-              pal.run_task(options[:object],
-                           targets,
-                           options[:task_options],
-                           executor,
-                           inventory,
-                           options[:description])
+              r = outputter.spin do
+                pal.run_task(options[:object],
+                             targets,
+                             options[:task_options],
+                             executor,
+                             inventory,
+                             options[:description])
+              end
+              r
             when 'file'
               src = options[:object]
               dest = options[:leftovers].first
@@ -678,7 +681,9 @@ module Bolt
 
       executor.subscribe(log_outputter)
       executor.start_plan(plan_context)
-      result = pal.run_plan(plan_name, plan_arguments, executor, inventory, puppetdb_client)
+      result = outputter.spin do
+        pal.run_plan(plan_name, plan_arguments, executor, inventory, puppetdb_client)
+      end
 
       # If a non-bolt exception bubbles up the plan won't get finished
       executor.finish_plan(result)
@@ -764,12 +769,15 @@ module Bolt
       modules   = project.modules || []
       installer = Bolt::ModuleInstaller.new(outputter, pal)
 
-      ok = installer.install(modules,
-                             project.puppetfile,
-                             project.managed_moduledir,
-                             config,
-                             force: force,
-                             resolve: resolve)
+      ok = outputter.spin do
+        installer.install(modules,
+                          project.puppetfile,
+                          project.managed_moduledir,
+                          config,
+                          force: force,
+                          resolve: resolve)
+      end
+
       ok ? 0 : 1
     end
 
@@ -790,12 +798,15 @@ module Bolt
       modules   = project.modules || []
       installer = Bolt::ModuleInstaller.new(outputter, pal)
 
-      ok = installer.add(name,
-                         modules,
-                         project.puppetfile,
-                         project.managed_moduledir,
-                         project.project_file,
-                         config)
+      ok = outputter.spin do
+        installer.add(name,
+                      modules,
+                      project.puppetfile,
+                      project.managed_moduledir,
+                      project.project_file,
+                      config)
+      end
+
       ok ? 0 : 1
     end
 
@@ -824,7 +835,10 @@ module Bolt
 
       outputter.print_message("Installing modules from Puppetfile")
       installer = Bolt::ModuleInstaller.new(outputter, pal)
-      ok = installer.install_puppetfile(puppetfile, moduledir, puppetfile_config)
+      ok = outputter.spin do
+        installer.install_puppetfile(puppetfile, moduledir, puppetfile_config)
+      end
+
       ok ? 0 : 1
     end
 
@@ -930,7 +944,11 @@ module Bolt
     end
 
     def outputter
-      @outputter ||= Bolt::Outputter.for_format(config.format, config.color, options[:verbose], config.trace)
+      @outputter ||= Bolt::Outputter.for_format(config.format,
+                                                config.color,
+                                                options[:verbose],
+                                                config.trace,
+                                                config.spinner)
     end
 
     def log_outputter
--- a/lib/bolt/outputter.rb
+++ b/lib/bolt/outputter.rb
@@ -2,24 +2,25 @@
 
 module Bolt
   class Outputter
-    def self.for_format(format, color, verbose, trace)
+    def self.for_format(format, color, verbose, trace, spin)
       case format
       when 'human'
-        Bolt::Outputter::Human.new(color, verbose, trace)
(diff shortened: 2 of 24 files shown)


lib/bolt/module_installer/installer.rb
```diff
@@ -28,14 +29,17 @@ def install(path, moduledir)
           moduledir:  moduledir.to_s
         }
 
+        @outputter.start_spin
```
beechtom, line 32: I think it'd be helpful to have the outputter spin for _all_ of the `module add|install` command, not just the install step. Resolving modules in particular can take a while, at times even longer than installing them.

lib/bolt/outputter.rb
```diff
@@ -2,24 +2,25 @@
...
       end
     end
 
-    def initialize(color, verbose, trace, stream = $stdout)
+    def initialize(color, verbose, trace, spin, stream = $stdout)
       @color = color
       @verbose = verbose
       @trace = trace
       @stream = stream
+      @spin = spin
```
beechtom, line 23: Since there's an instance variable set to indicate whether the spinner is on, we could use that to 'clear' the spinner character when printing subsequent messages, which would help with console messages having the spinner at the beginning. For example, `print_message` could be updated like so:

```ruby
      def print_message(message)
        @stream.print("\b") if @spin
        @stream.puts(message)
      end
```

lib/bolt/outputter/human.rb
```diff
@@ -14,12 +14,14 @@ class Human < Bolt::Outputter
...
       def print_head; end
 
-      def initialize(color, verbose, trace, stream = $stdout)
+      def initialize(color, verbose, trace, spin, stream = $stdout)
         super
         # Plans and without_default_logging() calls can both be nested, so we
         # track each of them with a "stack" consisting of an integer.
         @plan_depth = 0
         @disable_depth = 0
+        @dots = %w[⠋ ⠙ ⠹ ⠸ ⠼ ⠴ ⠦ ⠧ ⠇ ⠏]
```
beechtom, line 23: Is this going to be configurable? If not, it should just be removed for now.

lib/bolt/cli.rb
```diff
@@ -678,7 +681,9 @@ def run_plan(plan_name, plan_arguments, nodes, options)
 
       executor.subscribe(log_outputter)
       executor.start_plan(plan_context)
+      outputter.start_spin
       result = pal.run_plan(plan_name, plan_arguments, executor, inventory, puppetdb_client)
+      outputter.stop_spin
```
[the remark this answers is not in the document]
lucywyman: Oh no, I think this was just to demo that both worked the same. Updated.

lib/bolt/module_installer/installer.rb
lucywyman: Hm, yeah. I was initially thinking it'd be best to have the spinner spin closest to the long running operation themselves, so that any codepaths that got to that operation would use the spinner and there was less risk of having two spinners at once. That means passing the outputter around a lot though, and while I'm not sure anyone else is going to be using the module_installer and project migrator it doesn't quite feel right to *require* the outputter there. Then again, it does feel much less readable and discoverable when the calling methods are wrapped in spins and not the "end" of the codepath itself....I don't know I'm still on the fence.

lib/bolt/outputter.rb
lucywyman: I don't follow how that improves the feature or code?

beechtom: It removes the spinner character from the beginning of the line. For example:

image

With a change similar to this:

lucywyman: Are you sure you're on the latest? The spinner itself prints a backspace character that should remove it from any output, and I can't reproduce what you're seeing....

lib/bolt/outputter/human.rb
lucywyman: Yeah - I ended up going with the pinwheel for now but it'd be fun to add more later :)

lib/bolt/outputter.rb
beechtom: I'm pretty sure I was... I've been trying to always fetch when testing changes. It seems to be working now, though. Was it an issue with not printing the backspace for the rainbow outputter?

lib/bolt/module_installer/installer.rb
beechtom: I think having the spinners all in the CLI wrapping the entire operation makes things more discoverable, instead of needing to hunt down where the spinner is. It also seems a bit cleaner - as you noted with the outputter needing to be passed around a lot, not having to do that would keep function signatures simpler.

I think either approach is fine and don't feel super strongly one way or the other. 🤷

lucywyman: I don't think we can wrap in the CLI for things like the migrator because it will write the spinner to the prompt when we prompt for user input. But the more I think about it the more I like keeping the outputter out of the installer and resolver classes, so I like the approach I have here.

lib/bolt/outputter.rb
lucywyman: Yeah I think I just hadn't pushed it :sweat_smile:

spec/fixtures/configs/.plan_cache.json
```diff
@@ -0,0 +1 @@
...ernal facts"},"targets":{"type":"TargetSpec","sensitive":false,"description":"The targest the collect and set facts on"}},"module":"/home/lucy/githubs/bolt/modules/facts"},"canary":{"name":"canary","description":"Run a task, command or script on canary targets before running it on all targets.","parameters":{"task":{"type":"Optional[String[0]]","sensitive":false,"default_value":"undef","description":"The name of the task to run. Mutually exclusive with command and script."},"command":{"type":"Optional[String[0]]","sensitive":false,"default_value":"undef","description":"The command to run. Mutually exclusive with task and script."},"script":{"type":"Optional[String[0]]","sensitive":false,"default_value":"undef","description":"The script to run. Mutually exclusive with task and command."},"targets":{"type":"TargetSpec","sensitive":false,"description":"The target to run on."},"params":{"type":"Hash[String, Data]","sensitive":false,"default_value":"{}","description":"The parameters to use for the task."},"canary_size":{"type":"Integer","sensitive":false,"default_value":"1","description":"How many targets to use in the canary group."}},"module":"/home/lucy/githubs/bolt/modules/canary"}}
```
beechtom, line 1: Looks like this accidentally got added.

spec/bolt/module_installer/installer_spec.rb
```diff
@@ -1,8 +1,8 @@
 # frozen_string_literal: true
 
 require 'spec_helper'
-
 require 'bolt/module_installer/installer'
+require 'bolt/outputter'
```
beechtom, line 5:
```suggestion
```

beechtom (changes requested):

beechtom (approved):

PR #2437, opened 2020-12-07
(PE-30693) change project_ref to versioned_project

steveax:
This commit changes the project_ref parameter to versioned_project
to match the parameter in puppetserver.

!no-release-note

### DIFF
--- a/lib/bolt_server/transport_app.rb
+++ b/lib/bolt_server/transport_app.rb
@@ -51,8 +51,8 @@ module BoltServer
     # See the `orchestrator.bolt.codedir` tk config setting.
     DEFAULT_BOLT_CODEDIR = '/opt/puppetlabs/server/data/orchestration-services/code'
 
-    MISSING_PROJECT_REF_RESPONSE = [
-      400, Bolt::ValidationError.new('`project_ref` is a required argument').to_json
+    MISSING_VERSIONED_PROJECT_RESPONSE = [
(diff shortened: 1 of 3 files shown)


lib/bolt_server/transport_app.rb
```diff
@@ -514,10 +514,10 @@ def make_winrm_target(target_hash)
 
     # Fetches the metadata for a single plan
     #
-    # @param project_ref [String] the project to fetch the plan from
+    # @param versioned_project [String] the project to fetch the plan from
     get '/project_plans/:module_name/:plan_name' do
-      return MISSING_PROJECT_REF_RESPONSE if params['project_ref'].nil?
-      in_bolt_project(params['project_ref']) do |context|
+      return MISSING_PROJECT_REF_RESPONSE if params['versioned_project'].nil?
```
McdonaldSeanp, line 519: Nit: we could probably change this to `MISSING_VERSIONED_PROJECT_RESPONSE` to be a little clearer

But definitely don't block on that :)

McdonaldSeanp (approved):
👍

lib/bolt_server/transport_app.rb
steveax: Yep, good call, let me change that

steveax: Updated

donoghuc (approved):

Issue #2413, opened 2020-12-02
Validate inventory with config validator


beechtom:
## Use Case
Now that Bolt validates config files using the config schemas, we should validate inventory using the inventory schema. This would allow us to more easily validate inventories and provide more detailed error messages instead of failing at the first error encountered.

## Other opportunities

We could potentially add a `bolt inventory validate` command for users to validate their inventory without needing to run commands that load the inventory. This could be particularly helpful for users who use plugins in their inventory, as the command could resolve all references in the inventory and then validate it.

PR #2440, opened 2020-12-07
(GH-2413) Validate inventory with validator

beechtom:
This updates the `Bolt::Inventory` class to validate loaded inventory
against a schema using the validator. If the loaded inventory has any
errors, Bolt will raise an error describing each error, as well as its
location.

This also updates the `Bolt::Config::Validator` class. First, the class
is now just `Bolt::Validator`, since it is used to validate more than
just config files. Additionally, the validator now expects a schema
describing the _entire_ data structure that was loaded, not just the
various options available. This means each schema needs to describe the
type of data passed in. Because of this change, each of the schema
methods on `Bolt::Config` and `Bolt::Project` have been updated to
describe the full data structure.

To accommodate nested `groups` in inventory, the validator now handles a
new `:_ref` key in the schema. When the validator comes across a `:_ref`
key, it will use a definition from the schema to validate the value
against, pulling the definition from `schema[:definitions][ref]`. This
allows the validator to perform recursive validation.

!feature

* **Improved inventory validation**
  ([#2413](https://github.com/puppetlabs/bolt/issues/2413))

  Bolt now validates inventory against Bolt's inventory schema and
  indicates where errors are found.


lib/bolt/config/options.rb
```diff
@@ -50,70 +50,8 @@ module Options
         }
       }.freeze
 
-      # The following constants define the various configuration options available to Bolt.
```
beechtom, line 53: Moved all of this to the schemas README so as not to duplicate information across the 3 options files.

lib/bolt/inventory/options.rb
```diff
@@ -0,0 +1,136 @@
...
+        },
+        "name" => {
+          description: "A human-readable name to refer to the group or target. Names "\
+                       "cannot conflict with the name of a group, the name of a target, "\
+                       "or the alias of a target. A name is required for a group and is "\
+                       "required for a target unless the uri option is set.",
+          type: String,
+          _plugin: true
+        },
+        "plugin_hooks" => {
```
beechtom, line 81: Ugh. It'd be nice to rename this, but that seems like an inventory v3 thing?

lib/bolt/validator.rb
```diff
@@ -0,0 +1,224 @@
+# frozen_string_literal: true
```
beechtom, line 1: Moved this out of the `Bolt::Config` namespace since it's used by more than just that class.

rakelib/schemas.rake
```diff
@@ -129,45 +129,21 @@ namespace :schemas do
   desc 'Generate inventory.yaml JSON schema'
   task :inventory do
```
beechtom, line 130: This task is greatly simplified now since we have `Bolt::Inventory.schema`. We can likely do something similar for the other schemas, but I'm leaving that for another PR.

lib/bolt/config/options.rb
```diff
@@ -494,7 +432,7 @@ module Options
                        "specified in the URI.",
           type: String,
           enum: TRANSPORT_CONFIG.keys,
-          _plugin: false,
+          _plugin: true,
```
lucywyman, line 435: Was this just a bug?

beechtom: I believe so. I don't think we wanted to disallow plugins for the transport.

lib/bolt/inventory/options.rb
lucywyman: We can do that as part of 3.0, we obviously need to do it soon.

lib/bolt/validator.rb
```diff
@@ -0,0 +1,224 @@
...
+      @path         = []
+    end
+
+    # This is the entry method for validating data against the schema.
+    #
+    def validate(data, schema, location = nil)
+      @schema   = schema
+      @location = location
+
+      validate_value(data, schema)
```
lucywyman, line 25: The name is a bit funny now, since "value" can be the entire config/inventory contents. It's values all the way down :turtle:

lucywyman (changes requested):
Whoops I think we lost the deprecations in the new validator:
```
lucy ~/githubs/bolt ((76fac9e8...)) $ be bolt command run whoami -t localhost
Started on localhost...
Finished on localhost:
  STDOUT:
    lucy
Ran on 1 target in 0.0 sec
lucy ~/githubs/bolt ((76fac9e8...)) $ bolt command run whoami -t localhost
Option 'apply_settings' at /home/lucy/githubs/bolt/bolt-project.yaml is deprecated. This option will be removed in Bolt 3.0. Use `apply-settings` instead.
Started on localhost...
Finished on localhost:
  STDOUT:
    lucy
Ran on 1 target in 0.0 sec
```

beechtom:
Good catch 😬 Looks like it was happening for values that had a reference to a definition. Updated.

lib/bolt/inventory/options.rb
```diff
@@ -0,0 +1,137 @@
...
+    module Options
+      # Top-level options available in the inventory.
+      OPTIONS = %w[
+        config
+        facts
+        features
+        groups
+        targets
+        vars
+        version
```
lucywyman, line 16: I think we can remove this - it'll just show up as an unknown key, not error.

lib/bolt/inventory/options.rb
```diff
@@ -0,0 +1,137 @@
...
+          type: String,
+          format: "uri",
+          _plugin: true
+        },
+        "vars" => {
+          description: "A map of variables for the group or target.",
+          type: Hash,
+          _plugin: true
+        },
+        "version" => {
```
lucywyman, line 127: And this.

lib/bolt/inventory/options.rb
```diff
@@ -0,0 +1,137 @@
...
+          items: {
+            type: String,
+            _plugin: true
+          },
+          _plugin: true
+        },
+        "config" => {
+          description: "A map of configuration options.",
+          type: Hash,
+          properties: {},
```
lucywyman, line 36: Can we add a comment that these are generated elsewhere?

schemas/bolt-config.schema.json
```diff
@@ -317,14 +317,21 @@
...
       "description": "The default transport to use when the transport for a target is not specified in the URI.",
-      "type": "string",
-      "enum": [
-        "ssh",
-        "winrm",
-        "pcp",
-        "local",
-        "docker",
-        "remote"
+      "oneOf": [
```
lucywyman, line 320: Oo...do you know why this changed? I don't think it matters but it seems weird.

beechtom: Ya, it allows a user to use a plug-in reference for transport without the schema marking it as invalid. The rake task makes this same change everywhere we have :_plugin: true.

lib/bolt/inventory/options.rb
```diff
@@ -0,0 +1,130 @@
...
+        facts
+        features
+        groups
+        targets
+        vars
+      ].freeze
+
+      # Definitions used to validate the data.
+      # https://github.com/puppetlabs/bolt/blob/main/schemas/README.md
+      DEFINITIONS = {
```
lucywyman, line 20: I think this needs to merge in the transport definitions as well, so that unknown keys in transport config are caught:
```
lucy ~/githubs/bolt ((b79def82...)) $ cat inventory.yaml 
---
groups:
  - name: docker
    targets:
      - uri: localhost:20022
      - uri: localhost:20023
      - uri: localhost:20024
    config:
      ssh:
        user: bolt
        ssl-verify: false
        foo: bar
lucy ~/githubs/bolt ((b79def82...)) $ be bolt command run whoami -t localhost
Started on localhost...
Finished on localhost:
  STDOUT:
    lucy
Ran on 1 target in 0.01 sec
```

beechtom: The definitions are there, Bolt just wasn't logging the warnings on the validator. I swear I added that at some point, but guess not... anyway, it should warn now. :)

spec/integration/validator_spec.rb
```diff
@@ -114,6 +116,136 @@
...
+                  }
+                }
+              }
+            ]
+          }
+        ],
+        'config' => {
+          'transport' => 'ssh',
+          'ssh' => {
+            'host-key-check' => false
```
lucywyman, line 164: I think it'd be useful to add examples here of keys that are valid for other transports but not `ssh`, and a key that isn't valid for any transport.

spec/integration/validator_spec.rb
```diff
@@ -114,6 +116,136 @@
...
+                'winrm' => {
+                  'ssl' => false
+                }
+              }
+            }
+          }
+        ],
+        'config' => {
+          'transport' => 'foo',
+          'ssh' => {
```
lucywyman, line 220: I think it'd be useful to add examples here of keys that are valid for other transports but not `ssh`, and a key that isn't valid for any transport.

spec/integration/validator_spec.rb
```diff
@@ -114,6 +116,136 @@
...
+      expect { run_cli(command, project: project) }.not_to raise_error
+    end
+  end
+
+  context 'with invalid inventory' do
+    let(:inventory) do
+      {
+        'groups' => [
+          {
+            'targets' => [
```
lucywyman, line 184: Same here - verify that non-group-config (like `alias`) raises a warning.

lib/bolt/inventory/options.rb
```diff
@@ -0,0 +1,130 @@
...
+                       "cannot conflict with the name of a group, the name of a target, "\
+                       "or the alias of a target. A name is required for a group and is "\
+                       "required for a target unless the uri option is set.",
+          type: String,
+          _plugin: true
+        },
+        "plugin_hooks" => {
+          description: "Configuration for the Puppet library plugin used to install the "\
+                       "Puppet agent on the target. For more information, see "\
+                       "https://pup.pt/bolt-plugin-hooks",
```
lucywyman, line 85: This 404s

spec/integration/validator_spec.rb
lucywyman: Sorry I don't see the updates? Once those are tested I think this is GtG!

beechtom: There's a net new test "with unknown inventory options" at the bottom of the file. Can't test for warnings when there are errors, since Bolt errors before issuing the warnings.

lucywyman: Oh weird I get some warnings before errors:
```
lucy ~/githubs/bolt ((b8269dae...)) $ be bolt command run whoami -t localhost
Unknown option 'ssl' at 'inventory-config.ssh' at /home/lucy/.puppetlabs/etc/bolt/bolt-defaults.yaml.
Invalid configuration at /home/lucy/githubs/bolt/inventory.yaml:
  Value at 'groups.0.config.ssh.user' must be of type String
```
:shrug:  Thanks!

beechtom: Ya, that first one is for default config, which is validated before inventory. :)

PR #2444, opened 2020-12-08
(maint) Update puppet dep to 6.19.1

McdonaldSeanp:
PE services are failing to start up in the main branches because of a mismatch
between the puppet installed with bolt-server and the puppet.conf file changes
allowed in newer puppet versions.

This commit updates the puppet version dependency to the latest in the 6
series of puppet


donoghuc (changes requested):
I think all we need to do is just ship the newer gem in runtime

Issue #2445, opened 2020-12-08
Document plugin_hooks for Target objects in inventory


lucywyman:
## Describe the Change You Would Like
https://puppet.com/docs/bolt/latest/bolt_inventory_reference.html#target-objects

PR #2446, opened 2020-12-08
(GH-2445) Document 'plugin_hooks' option for inventory

beechtom:
This adds documentation for the `plugin_hooks` option to the inventory
reference, as `plugin_hooks` is a valid option for groups and targets.

!no-release-note

### DIFF
--- a/documentation/bolt_inventory_reference.md
+++ b/documentation/bolt_inventory_reference.md
@@ -192,6 +192,33 @@ groups:
   - name: linux
 ```
 
+### `plugin_hooks`
+
+A map of [plugin hooks](writing_plugins.md#plugin-hooks) and which plugins a
+hook should use for targets in the group. The only configurable plugin hook is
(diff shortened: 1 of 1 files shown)


hestonhoffman (approved):

documentation/bolt_inventory_reference.md
```diff
@@ -192,6 +192,30 @@ groups:
   - name: linux
 ```
 
+### `plugin_hooks`
+
+A map of [plugin hooks](writing_plugins.md#plugin-hooks) and which plugins a
+hook should use for targets in the group. The only configurable plugin hook is
+`puppet_library`, which configures the plugin used to install the Puppet agent
+on the targets. The `puppet_library` plugin hook can use one of two plugins:
```
lucywyman, line 200: 'When plans call `apply_prep`' - that's the only time this is used.

PR #2458, opened 2020-12-10
(maint) Consolidate BoltSpec::Config and BoltSpec::Files

lucywyman:
Previously, the `BoltSpec::Config` helper had a method `fixture_path()`
which was a stict subset of behavior from
`BoltSpec::Files#fixtures_path()` - the latter returns just the path to
`fixtures/` if no arguments are provided. This removes
`BoltSpec::Config#fixture_path()` in favor of
`BoltSpec::Files#fixtures_path()`. This also meant that the 
`BoltSpec::Config` class didn't have any methods that were being used,
however there was a method defined in a few different tests titled
`make_config()` which seemed like it fit that helper. This moves that
function to the `BoltSpec::Config` class and removes it from the spec
tests that defined it. The exception to this is
`spec/integration/transport/winrm_spec.rb`, which uses a custom
`mk_config()` function to return just the transport config as a hash,
not a `Bolt::Config` object. This also removes the `modulepath` helper
method, as it would opaquely set the modulepath (particularly in files
where the modulepath would sometimes come from the helper, then later be
overridden by an actual `let(:modulepath)` setting).

This also adds the following helpers to the BoltSpec::PAL helper:
- `make_pal()`: which accepts an optional modulepath. The modulepath
  defaults to `['spec/fixtures/modules/']`
- `make_plugins()`: which accepts an optional Bolt::Config instance. The 
  default config has no data, and uses a project with the CWD.
- `make_inventory()`: which accepts an optional data hash and uses the 
  default config and plugins to make a Bolt::Inventory::Inventory object.

!no-release-note

### DIFF
--- a/spec/bolt/inventory/group_spec.rb
+++ b/spec/bolt/inventory/group_spec.rb
@@ -11,10 +11,11 @@ require 'bolt_spec/plugins'
 describe Bolt::Inventory::Group do
   include BoltSpec::Config
 
-  let(:data) { { 'name' => 'all' } }
-  let(:pal) { nil } # Not used
+  let(:data)    { { 'name' => 'all' } }
+  let(:pal)     { nil } # Not used
+  let(:config)  { make_config }
   let(:plugins) { Bolt::Plugin.setup(config, nil) }
-  let(:group) {
+  let(:group)   {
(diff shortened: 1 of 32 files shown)


beechtom (approved):
🎉  Looks great! Happy to see that spec tests are finally getting some TLC.

Issue #2449, opened 2020-12-09
Remove deprecated CLI flags from spec tests


lucywyman:
--boltdir --configfile --description --puppetfile

Issue #2428, opened 2020-12-04
Add 'config' parameter to with_project spec helper


lucywyman:
## Use Case
The project spec_helper function `with_project` currently reads a variable `project_config` from the rspec test it's used in, which can be opaque when reading the tests. We should pass `config` to with_project explicitly.

beechtom:
Closed by #2466

PR #2465, opened 2020-12-11
(GH-2449) Remove deprecated CLI flags from the codebase

lucywyman:
This removes `--configfile`, `--boltdir`, and `--puppetfile` (and
`--description`, but that wasn't used anywhere) from our spec tests. It
leaves tests that specifically test those flags to be removed when the
flag is removed, but refactors tests that relied on the flags for
setting Bolt config to use Bolt projects instead.

Note: This relies on changes from #2428

!no-release-note


lucywyman:
This also includes changes from #2458, just to avoid conflicts.

beechtom (changes requested):
Needs a rebase

beechtom (approved):

PR #2466, opened 2020-12-11
(GH-2428) Update with_project method to be more transparent

beechtom:
This updates the `BoltSpec::Project` method `with_project` to be more
transparent. Previously, the method used some magic to configure a
project by calling the `project_config` method. Additionally, the method
would also set several instance variables for spec tests to use.

The `with_project` method now includes a few keyword arguments that can
be used to modify the project directory structure that it creates:

- `config`

  A hash of config to write to `bolt-project.yaml`. If `nil`, writes a
  default config file with the project name set.

- `inventory`

  A hash of inventory data to write to `inventory.yaml`. If `nil`,
  writes an empty inventory file.

- `boltdir`

  A boolean. When true, creates the project inside a `Boltdir`.

The `with_project` method now yields a `Bolt::Project` instance for the
project directory. The `in_project` method was also updated to accept
the same keyword arguments and yield a `Bolt::Project` instance.

This also adds two new methods to `BoltSpec::Project`:
`with_project_directory` and `make_project`. The
`with_project_directory` method accepts all the same arguments as
`with_project`, but yields the project path instead of a `Bolt::Project`
instance. This method is helpful when writing tests that rely on errors
being raised when project config is loaded. The `make_project` method
accepts a path and returns a `Bolt::Project` instance for the project
directory.

!no-release-note

### DIFF
--- a/spec/bolt/cli_spec.rb
+++ b/spec/bolt/cli_spec.rb
@@ -170,12 +170,15 @@ describe "Bolt::CLI" do
   end
 
   context 'module' do
-    let(:cli)       { Bolt::CLI.new(command + %W[--project #{project_path}]) }
-    let(:command)   { %w[module show] }
-    let(:installer) { double('installer', add: true, install: true) }
+    let(:cli)            { Bolt::CLI.new(command) }
+    let(:command)        { %w[module show] }
+    let(:installer)      { double('installer', add: true, install: true) }
+    let(:project_config) { nil }
+    let(:project)        { @project }
 
     around(:each) do |example|
-      with_project do
+      in_project(config: project_config) do |project|
+        @project = project
         example.run
       end
     end
@@ -204,7 +207,7 @@ describe "Bolt::CLI" do
       let(:project_config) { { 'modules' => [] } }
 
       it 'errors without a module' do
-        cli = Bolt::CLI.new(%W[module add --project #{project}])
+        cli = Bolt::CLI.new(%w[module add])
         expect { cli.parse }.to raise_error(
           Bolt::CLIError,
           /Must specify a module name/
@@ -212,7 +215,7 @@ describe "Bolt::CLI" do
       end
 
       it 'errors with multiple modules' do
-        cli = Bolt::CLI.new(%W[module add foo bar --project #{project}])
+        cli = Bolt::CLI.new(%w[module add foo bar])
         expect { cli.parse }.to raise_error(
           Bolt::CLIError,
           /Unknown argument/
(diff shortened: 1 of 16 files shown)


spec/bolt/cli_spec.rb
```diff
@@ -170,12 +170,15 @@ def stub_config(file_content = {})
...
 
   context 'module' do
-    let(:cli)       { Bolt::CLI.new(command + %W[--project #{project_path}]) }
-    let(:command)   { %w[module show] }
-    let(:installer) { double('installer', add: true, install: true) }
+    let(:cli)            { Bolt::CLI.new(command) }
+    let(:command)        { %w[module show] }
+    let(:installer)      { double('installer', add: true, install: true) }
+    let(:project_config) { nil }
+    let(:project)        { @project }
```
lucywyman, line 177: I don't think this gets used?

spec/bolt/cli_spec.rb
```diff
@@ -2328,13 +2327,13 @@ def stub_config(file_content = {})
...
-      let(:output) { StringIO.new }
-      let(:puppetfile) { @puppetfile }
-      let(:modulepath) { (project_path + 'modules').to_s }
-      let(:action_stub) { double('r10k_action_puppetfile_install') }
-      let(:project_config) { { 'modules' => [] } }
-
-      let(:cli) { Bolt::CLI.new(%W[puppetfile install --project #{project_path} -m #{modulepath}]) }
+      let(:output)         { StringIO.new }
+      let(:puppetfile)     { project.puppetfile }
+      let(:modulepath)     { (project.path + 'modules').to_s }
```
lucywyman, line 2332: Isn't this `project.modulepath`?

spec/bolt/plan_creator_spec.rb
```diff
@@ -79,8 +73,8 @@
 
     it 'catches existing file errors when creating directories' do
       plan_name = "#{project.name}::foo::bar"
-      FileUtils.mkdir(File.join(@project_path, 'plans'))
-      FileUtils.touch(File.join(@project_path, 'plans', 'foo'))
+      FileUtils.mkdir(File.join(project.path, 'plans'))
```
lucywyman, line 76:
```suggestion
      FileUtils.mkdir(project.plans_path)
```

spec/integration/validator_spec.rb
```diff
@@ -8,14 +8,20 @@
   include BoltSpec::Project
 
   around(:each) do |example|
-    with_project do
-      File.write((@project_path + 'inventory.yaml'), inventory.to_yaml)
+    with_project_directory(config: project_config, inventory: inventory) do |project_path|
```
lucywyman, line 11: I think this can be `in_project`?

lucywyman: Ah, I think I see - with `run_cli` you either need to specify a project or run with `--project`. Honestly now that we have the project helpers it seems like removing the `project` key from run_cli is best so we can standardize on using `with_project` and specifying `--project` on the CLI in cases where it needs it. But if that will take too long (maybe more than half a day) to do, we can do that separately and leave this as is for now.

Issue #2438, opened 2020-12-07
Don't print `no project name` warning when `bolt-project.yaml` doesn't exist


lucywyman:
## Describe the Bug
We should not print the warning if the project does not have any directories to load content from: `plans/`, `tasks`/ and `files/`, and not if those directories are not present.

PR #2468, opened 2020-12-11
(GH-2438) Don't warn that project content won't be loaded if there's no project content

lucywyman:
This limits the warning that project content won't be loaded if the
project doesn't have a name to only be raised if the project has a
`tasks/`, `plans/`, or `files/` directory.

!bug

* **Only warn that project content won't be loaded if there's project content**

  Bolt will now only warn that project content won't be loaded if the
  proejct directory has a `tasks/`, `plans/`, or `files/` directory that
  may contain content.

### DIFF
--- a/lib/bolt/project.rb
+++ b/lib/bolt/project.rb
@@ -227,7 +227,10 @@ module Bolt
           raise Bolt::ValidationError, "The project '#{name}' will not be loaded. The project name conflicts "\
             "with a built-in Bolt module of the same name."
         end
-      else
+      elsif name.nil? &&
+            (File.directory?(plans_path) ||
+            File.directory?(@path + 'tasks') ||
+            File.directory?(@path + 'files'))
         message = "No project name is specified in bolt-project.yaml. Project-level content will not be available."
         @logs << { warn: message }
       end
--- a/spec/bolt/project_spec.rb
+++ b/spec/bolt/project_spec.rb
@@ -83,6 +83,29 @@ describe Bolt::Project do
     end
   end
 
+  describe "with no name set" do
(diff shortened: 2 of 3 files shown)


lib/bolt/project.rb
```diff
@@ -227,7 +227,10 @@ def validate
           raise Bolt::ValidationError, "The project '#{name}' will not be loaded. The project name conflicts "\
             "with a built-in Bolt module of the same name."
         end
-      else
+      elsif name.nil? &&
+            (File.directory?(plans_path) ||
+            File.directory?(@path + 'tasks') ||
+            File.directory?(@path + 'files'))
```
beechtom, line 233: Why are we including the `files/` directory in this check? I know `tasks/` and `plans/` make sense, since you can reference content in those directories from the CLI, but I don't think you can for `files/` (unless I'm missing a command).

lucywyman: You could reference a file from the project in a manifest being applied, for (very contrived) example:
```
bolt apply -e " file { '/tmp/foobar': content => file('myproject/service.conf'), ensure => present }"
```
Or more likely from a manifest file you specify with `bolt apply`.

I think we opted not to include manifests for the reason you said, that usually if you're consuming manifests it's probably in a plan (in which case you'd get the error). We could add `manifests/` too to err on the side of zealously warning, in case people are consuming the project from a module on the modulepath or something.

beechtom: I'm not really worried about it, and it's easy enough to add to the check later if needed. :) I was more just curious why `files/` was included, and that example makes perfect sense!

beechtom (approved):

Issue #2454, opened 2020-12-10
Remove validation that is checked by the validator


lucywyman:
We have some custom validation that should now be covered by the Validator class that we can remove, for example: https://github.com/puppetlabs/bolt/tree/main/lib/bolt/config.rb#L464.

PR #2476, opened 2020-12-14
(GH-2454) Remove redundant config validation

beechtom:
This removes some redundant validation for the `Bolt::Config` class that
is handled by the validator.

!no-release-note

### DIFF
--- a/lib/bolt/config.rb
+++ b/lib/bolt/config.rb
@@ -448,39 +448,14 @@ module Bolt
         next if val == 'disable'
 
         name = normalize_log(key)
+        acc[name] = val.slice('append', 'level').transform_keys(&:to_sym)
 
-        # But otherwise it has to be a Hash
-        unless val.is_a?(Hash)
-          raise Bolt::ValidationError,
-                "config of log #{name} must be a Hash, received #{val.class} #{val.inspect}"
-        end
+        next unless acc[name][:level] == 'notice'
 
-        acc[name] = val.slice('append', 'level')
-                       .transform_keys(&:to_sym)
-
-        if (v = acc[name][:level])
-          unless v.is_a?(String) || v.is_a?(Symbol)
-            raise Bolt::ValidationError,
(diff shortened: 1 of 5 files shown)


lucywyman (changes requested):
We also need to do this for the transport configurations, inventory, and plugin config. For example we validate that `login-shells` is one of `LOGIN_SHELLS` in `lib/bolt/config/transport/ssh.rb#L114` when we don't need to

beechtom:
Updated to remove validation for `login-shells`, which was the only manual validation that didn't get pulled out when I first added the validator to those classes.

We can't really pull out any other manual validation in the inventory classes, since the data needs to be validated after plugins are resolved... we can probably validate that data with the validator class after plugins are resolved, but that would require additional work.

I'm not really sure what you mean by plugin config... we don't have schemas for plugins. Again, we can probably update the built-in plugins to use the validator, but that seems outside the scope of this ticket.

lucywyman:
There's also redundant timeout validation in `lib/bolt/puppetdb/config.rb`, we can remove `lib/bolt/inventory/group.rb#L253`, I think `lib/bolt/inventory/group.rb#L258` is invalid, or at least I'm not sure how to trigger it? I haven't finished going through all the files but that's at least a few.

beechtom:
lib/bolt/inventory/group.rb#L253 can't be removed currently because the data it checks might come from a plugin reference, which is not checked by the validator.

lib/bolt/inventory/group.rb#L258 ensures the key itself cannot be a plugin reference. I'm not sure exactly sure why we have that test in the first place, though... I'm fine getting rid of it if we consider it an unlikely scenario that we need to test for.

lucywyman:
Yeah mostly I'm not sure *how* to make the key a plugin reference, since it has to be a hash and `walk_keys` will walk the keys of the `_plugin` hash and not the whole hash (if that makes sense).

I see what you mean about a lot of validation being after plugins are resolved (like the PuppetDB timeout config).

lucywyman (approved):
GtG once `lib/bolt/inventory/group.rb#L258` is removed!

PR #2480, opened 2020-12-15
(GH-2477) Allow any keys under `remote` transport config

### LINKED ISSUE #2477: Unknown options warning messages with bolt 2.38.0
## Describe the Bug
When running Bolt `2.38.0` i'm seeing the following warning messages:

```
Unknown option 'version' at xxx/bolt/patching/inventory.yaml.
Unknown option 'port' at 'groups.0.config.remote' at xxx/bolt/patching/inventory.yaml.
Unknown option 'username' at 'groups.0.config.remote' at xxx/bolt/patching/inventory.yaml.
Unknown option 'password' at 'groups.0.config.remote' at xxx/bolt/patching/inventory.yaml.
```

## Expected Behavior
No warning messages shown.

## Steps to Reproduce
Here is my `inventory.yaml`
```yaml
---
version: 2
groups:
  - name: xxx
    config:
      transport: remote
      remote:
        port: 1234
        username: 'user@domain.tld'
        password: xxx
```

## Environment
 - Version `2.38.0`

lucywyman:
There was a bug in the configuration schema that warned when users would
set any configuration under the remote transport key. The key now
accepts a hash with any keys and values and does not warn.

Closes #2477

!bug

* **Fix bug warning about keys under 'remote' transport** ([#2477](https://github.com/puppetlabs/bolt/issues/2477))

  Bolt now will not warn when keys are configured for the `remote`
  transport in inventory.

### DIFF
--- a/lib/bolt/config/options.rb
+++ b/lib/bolt/config/options.rb
@@ -509,6 +509,7 @@ module Bolt
         "remote" => {
           description: "A map of configuration options for the remote transport.",
           type: Hash,
+          additionalProperties: true,
           _plugin: true,
           _example: { "run-on" => "proxy_target" }
         },
--- a/schemas/bolt-inventory.schema.json
+++ b/schemas/bolt-inventory.schema.json
@@ -352,6 +352,7 @@
               "oneOf": [
                 {
                   "type": "object",
+                  "additionalProperties": true,
                   "properties": {
                     "run-on": {
                       "description": "The proxy target that the task executes on.",
--- a/lib/bolt/config/transport/options.rb
+++ b/lib/bolt/config/transport/options.rb
@@ -143,7 +143,8 @@ module Bolt
                          "`task.py`) and the extension is case sensitive. When a target's name is `localhost`, "\
                          "Ruby tasks run with the Bolt Ruby interpreter by default.",
             additionalProperties: {
-              type: String
+              type: String,
+              _plugin: false
             },
             propertyNames: {
               pattern: "^.?[a-zA-Z0-9]+$"
--- a/lib/bolt/validator.rb
+++ b/lib/bolt/validator.rb
@@ -87,7 +87,7 @@ module Bolt
         if properties.include?(key)
           check_deprecated(key, definition[:properties][key])
           validate_value(val, definition[:properties][key])
-        elsif definition[:additionalProperties]
+        elsif definition[:additionalProperties].is_a?(Hash)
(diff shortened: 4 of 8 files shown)


lib/bolt/config/options.rb
```diff
@@ -509,6 +509,11 @@ module Options
         "remote" => {
           description: "A map of configuration options for the remote transport.",
           type: Hash,
+          additionalProperties: {
+            description: "Parameters to pass to the remote task",
```
beechtom, line 513:
```suggestion
            description: "Parameter to pass to the remote task.",
```

beechtom (approved):

schemas/bolt-inventory.schema.json
```diff
@@ -352,6 +352,23 @@
...
                 {
                   "type": "object",
+                  "additionalProperties": {
+                    "description": "Parameter to pass to the remote task",
+                    "oneOf": [
+                      {
+                        "type": [
+                          "string",
+                          "object",
+                          "integer",
```
adreyer, line 362: This should be `number` right? Is there a reason to specify a list of types instead of leaving it empty to accept any type?

lucywyman: Yeah, we validate that all definitions have a `type` defined. It might be safer to remove that check instead of doing this.

lucywyman: Ok, I ended up just explicitly skipping this in the test, which I think is the best of both worlds. That way future options that don't have a type still fail the test (which seems especially useful for community PRs) but we don't have to define the type here.

lib/bolt/config/options.rb
```diff
@@ -509,6 +509,10 @@ module Options
         "remote" => {
           description: "A map of configuration options for the remote transport.",
           type: Hash,
+          additionalProperties: {
+            description: "Parameter to pass to the remote task",
+            _plugin: true
+          },
```
beechtom, line 515:
```suggestion
          additionalProperties: true,
```

JSON Schema Draft 07 lets `additionalProperties` be a boolean. By default it's set to `true`, but we can explicitly set it to `true` to tell the validator to skip the `validate_keys` check.

Then the validator just needs a small update [here](https://github.com/puppetlabs/bolt/blob/main/lib/bolt/validator.rb#L90):

```
elsif definition[:additionalProperties].is_a?(Hash)
```

That'll let us skip the extra logic in the spec test, too.

beechtom (approved):

PR #2481, opened 2020-12-15
Support plugins under 'private-key.key-data' and 'interpreters' in transport config

beechtom:
This allows the `key-data` suboption for the `private-key` config option
to be a plugin reference.

!bug

* **Allow plugins under `key-data` in `private-key` config**

  The `key-data` suboption for the `private-key` config option once
  again supports plugins.

This allows plugins to be set under individual interpreters in transport
config.

!bug

* **Allow plugins under individual interpreters in transport config**

  Individual interpreters under the `interpreters` transport config
  option once again support plugins.


lucywyman:
Lol https://github.com/puppetlabs/bolt/pull/2482 - I also explicitly set it to false for some properties -not sure which solution we'd prefer?

PR #2482, opened 2020-12-15
(bug-fix) Set _plugin for sub-keys in config

lucywyman:
Previously, we didn't set `_plugin` for most `properties` definitions in
the configuration schema. Those sub-keys for top-level config options
were assumed to inherit the outer-configs `_plugin` key, which isn't the 
case. This was already addressed for the `puppetdb` configuration
options, but missed `transport.private-key.key-data` and 
`transport.interpreters`. This change explicitly sets `_plugin` to match
the 'outer' `_plugin` config for all `properties` of configuration
options.

!bug

* **Allow plugins for sub-option `key-data` and `interpreters**

  The configuration for SSH private key using `key-data` now accepts a
  plugin.

### DIFF
--- a/lib/bolt/config/options.rb
+++ b/lib/bolt/config/options.rb
@@ -61,6 +61,7 @@ module Bolt
             "show_diff" => {
               description: "Whether to log and report a contextual diff.",
               type: [TrueClass, FalseClass],
+              _plugin: false,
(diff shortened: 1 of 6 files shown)


lucywyman:
Closing in favor of #2483

PR #2483, opened 2020-12-15
(bug) Support plugins for suboptions when superoption supports plugins

beechtom:
This updates the validator to allow plugins for all items and suboptions
of a given option if that option supports plugins. Items and suboptions
do not need to explicitly set the `:_plugin` key to enable this
behavior.

!bug

* **Support plugins for suboptions under options that allow plugins**

  All suboptions for config options that support plugins once again
  support plugins. For example, the `key-data` suboption for the
  `private-key` option can use plugins again.

### DIFF
--- a/rakelib/schemas.rake
+++ b/rakelib/schemas.rake
@@ -223,7 +223,7 @@ def add_plugin_reference(definition)
 end
 
 # Recurses through a definition and JSON-ifies the data types.
-def to_schema(data)
+def to_schema(data, plugin = false)
   return data unless data.is_a?(Hash)
 
   # Creates JSON references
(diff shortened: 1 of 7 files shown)


rakelib/schemas.rake
```diff
@@ -236,22 +236,26 @@ def to_schema(data)
...
+  # Flip plugin to true if the definition allows a plugin reference.
+  # This will allow plugin references for all sub-options and items.
+  plugin = metadata[:_plugin] if metadata[:_plugin]
+
   # Recurse through some properties, as each can have their own defintions.
-  %i[items additionalProperties propertyNames].each do |key|
+  %i[items additionalProperties].each do |key|
     next unless data.key?(key)
-    data[key] = to_schema(data[key])
+    data[key] = to_schema(data[key], plugin)
```
lucywyman, line 246: This should use the `_plugin` definition for the suboption if it's defined, otherwise default to `_plugin` for the parent key.

lucywyman:
Note for posterity: `_plugin` is still respected per-definition, even if it differs from the value of `_plugin` for the "parent" definition. The parent's `_plugin` is a default.

spec/bolt/validator_spec.rb
```diff
@@ -134,6 +134,83 @@ def validate
         /Value at 'option' is a plugin reference, which is unsupported at this location/
       )
     end
+
+    context 'with suboptions' do
```
lucywyman, line 138: Can we test this with the actual options? `plugins` doesn't allow `_plugin`, but any suboptions of `plugins` do. And then we can use the cases that raised this (i.e. key-data) to verify that defaults work.

beechtom: Added integration tests for `plugins`, `plugins.*`, `private-key.key-data`, `interpreters`, and `interpreters.*`. That should cover `_plugin` inheritance from parent options and the cases that raised this issue.

I'm wondering if we should just explicitly test each option or not...

lucywyman: I think general testing is good, we can manually test each option but don't want to do that for every spec run.

PR #2487, opened 2020-12-16
(bug) Correctly compute Cache ID for plugin invocations

lucywyman:
Previously, we computed a unique cache ID for each plugin invocation
based just on the value of `_cache` for the invocation. This meant that
any plugin invocations with the same cache `ttl` value would override
each other, causing the wrong cached values to be loaded for plugin
invocations. We now correctly compute the cache ID to be the plugin
invocation hash except for the `_cache` data, so that the same plugin
with different cache `ttl` values will load the same data.

!bug

* **Load the correct data for plugin invocations**

  Bolt now correctly loads data for the plugin invocation based on the
  plugin data, not just cache `ttl`. Previously, any plugins with the
  same cache configuration would collide in the cache data and overwrite
  each other, causing the wrong data to be loaded.


beechtom (approved):
Checked with multiple plugins, including plugins with `_cache` configured, and plugins without `_cache` configured and `plugin-cache` configured.

PR #2488, opened 2020-12-16
(Docs) Improve troubleshooting doc

hestonhoffman:
Moved the bit on tempdirs into its own section so it's easier
for people to scan and find the info they're looking for

!no-release-note

### DIFF
--- a/documentation/troubleshooting.md
+++ b/documentation/troubleshooting.md
@@ -26,15 +26,7 @@ You need to upgrade your project so Bolt can manage your modules and
 dependencies. For more information, see [migrate a Bolt
 project](./projects.md#migrate-a-bolt-project).
 
-## My task fails mysteriously
-
-Try running Bolt with `--log-level debug` to see the exact output from your task.
-
-Make sure your task executable starts with a shebang (`#!`) line indicating the
-interpreter to use and verify that the executable is present on the target
-system. For example, if you write a Python task and include the line:
-`#!/usr/bin/env python`, Bolt attempts to execute the script using the default
-`python` executable on the target system.
+## My task fails with a "permission denied" error (`noexec` issue)
 
 If your task fails with the following error, the issue may be that your temporary directory (tmpdir)
 is [mounted with noexec](https://superuser.com/questions/728127/what-does-noexec-flag-mean-when-mounting-directories-on-rhel).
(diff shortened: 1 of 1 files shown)


documentation/troubleshooting.md
```diff
@@ -26,15 +26,7 @@ You need to upgrade your project so Bolt can manage your modules and
...
-## My task fails mysteriously
-
-Try running Bolt with `--log-level debug` to see the exact output from your task.
-
-Make sure your task executable starts with a shebang (`#!`) line indicating the
-interpreter to use and verify that the executable is present on the target
-system. For example, if you write a Python task and include the line:
-`#!/usr/bin/env python`, Bolt attempts to execute the script using the default
-`python` executable on the target system.
+## My task fails with a "permission denied" error
```
lucywyman, line 29: It sounds like Dan might have been specifically looking for `noexec` here...I'm not sure how to phrase this such that it includes that? Something like `## My task fails with a "permission denied" error - could be that tmpdir is noexec` but...obviously shorter :grimacing:

hestonhoffman: Doesn't that mean you'd be looking for the problem using the solution?

hestonhoffman: Okay, I think I understand the issue. Let me take another stab at it

beechtom: Considering we've had several users run into this issue recently, I wonder if it's worth adding a section to the [Running tasks](https://puppet.com/docs/bolt/latest/bolt_running_tasks.html) page that explains how tasks are uploaded to `/tmp` and how you can configure it if that directory is mounted noexec.

That can definitely be in another ticket if we do want to go that route.

hestonhoffman:
@lucywyman I don't love this solution, but at least it keeps the heading down to a manageable length. Judging from past feedback, some users do not know that the directory has `noexec` applied. So I want to make sure that we're addressing those users. Kind of a chicken and egg scenario

documentation/troubleshooting.md
hestonhoffman: Yeah that's true. I think there's a ticket somewhere for improving the tasks docs. I'll add this to that ticket.

lucywyman:
Definitely - this solution seems ideal in that people who *suspect* noexec may be an issue can find it easily, but those who don't will still read it

lucywyman (approved):

beechtom (approved):

Issue #2447, opened 2020-12-09
Ensure that Bolt-level plugin-hooks config is picked up by Target object


lucywyman:
We need to ensure that configuration from `bolt-*.yaml` for `plugin-hooks` is picked up by Target objects as `plugin_hooks`.

PR #2498, opened 2020-12-17
(GH-2447) Ensure plugin-hooks config picked up by targets

beechtom:
This adds a test to ensure that the `plugin-hooks` config option is
picked up by targets, which use the `plugin_hooks` field.

!no-release-note

### DIFF
--- a/spec/integration/inventory_spec.rb
+++ b/spec/integration/inventory_spec.rb
@@ -4,15 +4,16 @@ require 'spec_helper'
 require 'bolt_spec/conn'
 require 'bolt_spec/files'
 require 'bolt_spec/integration'
+require 'bolt_spec/project'
 require 'bolt_spec/puppetdb'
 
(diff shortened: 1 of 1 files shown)


spec/integration/inventory_spec.rb
```diff
@@ -58,24 +59,25 @@
...
+      'modulepath' => fixtures_path('modules')
+    }
+  end
 
-  let(:run_plan) { ['plan', 'run', 'inventory', "command=#{shell_cmd}", "host=#{target}"] + config_flags }
+  let(:config)       { base_config }
+  let(:config_flags) { %W[--project #{@project.path} --password #{conn[:password]}] }
+  let(:conn)         { conn_info('ssh') }
+  let(:target)       { conn[:host] }
+  let(:modulepath)   { fixtures_path('modules') }
```
lucywyman, line 74: This doesn't seem to get used

Issue #2013, opened 2020-07-22
Change ExecutionPolicy instructions when 2.20.0 is released


jpogran:
## Describe the Change You Would Like

When 2.20.0 is released, the PowerShell module files will be signed, which means users are not required to set the `ExecutionPolicy` to `RemoteSigned` in order to use Bolt as described [here](https://puppet.com/docs/bolt/latest/bolt_installing.html#change-execution-policy-restrictions)

jpogran:
We can update the explanation in https://puppet.com/docs/bolt/latest/bolt_installing.html#change-execution-policy-restrictions

PowerShell's [execution policy](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_execution_policies?view=powershell-7) is a safety feature that controls the conditions under which PowerShell loads configuration files and runs scripts. This feature helps prevent the execution of malicious scripts. The default policy is `Restricted` (allows no scripts to run) for Windows *clients* and `RemoteSigned` (allows signed scripts and non-signed scripts not from the internet) for Windows *servers*. Some environments change this to `AllSigned`, which only allows scripts to run as long as they are signed from with a trusted publisher.

As of 2.21.0 we sign the Bolt PowerShell Module files with a Puppet code signing cert. This means that environments that have `AllSigned` set will allow the bolt command to work without user input so long as they trust our publisher. If they do not implicitly trust the publisher, then can accept the publisher without having to change `ExecutionPolicy`. Environments that have it set to `Restricted` will still need to change their policy to `RemoteSigned` or `AllSigned`.

PR #2501, opened 2020-12-18
(GH-2013) Update docs on PowerShell execution policy

beechtom:
This updates our documentation on PowerShell execution policy when
installing Bolt on Windows. Since Bolt's PowerShell module is now
signed, users on Bolt >= 2.21.0 will likely not need to change their
script execution policy to use Bolt.

!no-release-note

**Note:** The text for this is taken directly from [James' comment](https://github.com/puppetlabs/bolt/issues/2013#issuecomment-675027455) in the issue.

### DIFF
--- a/documentation/bolt_installing.md
+++ b/documentation/bolt_installing.md
@@ -101,12 +101,27 @@ profile.
 
 ### Change execution policy restrictions
 
-Some Windows installations have security restrictions that do not allow Bolt to
-run. These restrictions are easy to change, but check with your security team
-first.
+PowerShell's execution policy is a safety feature that controls the conditions
+under which PowerShell loads configuration files and runs scripts. This feature
+helps prevent the execution of malicious scripts. The default policy is
+`Restricted` (allow no scripts to run) for Windows _clients_ and `RemoteSigned`
+(allow signed scripts and non-signed scripts not from the internet) for Windows
+_servers_. Some environments change this to `AllSigned`, which only allows
+scripts to run as long as they are signed by a trusted publisher.
+
+As of Bolt 2.21.0, we sign Bolt PowerShell module files with a Puppet code
+signing certificate. If your PowerShell environment uses an `AllSigned`
+execution policy and you add Puppet as a trusted publisher, the `bolt` command
+works without any further input. If you're using the `AllSigned` policy and you
+have not added Puppet as a trusted publisher, you can accept the publisher
+without having to change your execution policy.
+
+If your environment uses a `Restricted` policy, you must change your policy to
+`RemoteSigned` or `AllSigned`. Check with your security team before you make any
+policy changes.
 
 If you see this or a similar error when trying to run Bolt, you probably need to
-change your script execution policy restrictions, as described here.
+change your script execution policy restrictions:
 
 ```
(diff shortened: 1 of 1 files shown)


lucywyman (approved):
LGTM! I'll wait for review from @hestonhoffman and @jpogran .

jpogran (approved):
LGTM!

documentation/bolt_installing.md
```diff
@@ -101,9 +101,22 @@ profile.
...
+scripts to run as long as they are signed from with a trusted publisher.
+
+As of Bolt 2.21.0, we sign the Bolt PowerShell Module files with a Puppet code
+signing cert. This means that environments that have `AllSigned` set will allow
+the `bolt` command to work without user input so long as they trust our
+publisher. If they do not implicitly trust the publisher, they can accept the
+publisher without having to change `ExecutionPolicy`. Environments that have it
+set to `Restricted` will still need to change their policy to `RemoteSigned` or
+`AllSigned`. These restrictions are easy to change, but check with your security
+team first.
```
hestonhoffman, line 119:
```suggestion
As of Bolt 2.21.0, we sign Bolt PowerShell module files with a Puppet code
signing certificate. If your PowerShell environment uses an `AllSigned` execution 
policy and you add Puppet as a trusted publisher, the `bolt` command works 
without any further input. If you're using the `AllSigned` policy and you have 
not added Puppet as a trusted publisher, you can accept the publisher without 
having to change your execution policy.

If your environment uses a `Restricted` policy, you must change your policy to 
`RemoteSigned` or`AllSigned`. Check with your security team before you make 
any policy changes.
```

documentation/bolt_installing.md
```diff
@@ -101,9 +101,22 @@ profile.
...
 
 If you see this or a similar error when trying to run Bolt, you probably need to
 change your script execution policy restrictions, as described here.
```
hestonhoffman, line 122:
```suggestion
If you see this or a similar error when trying to run Bolt, you probably need to
change your script execution policy restrictions:
```

documentation/bolt_installing.md
```diff
@@ -118,8 +131,10 @@ For more information, run 'Import-Module PuppetBolt'.
                 n
                 + FullyQualifiedErrorId : CouldNotAutoloadMatchingModule
 ```
+
```
hestonhoffman, line 134:
```suggestion
To change your script execution policy:
```

documentation/bolt_installing.md
hestonhoffman: Question: How would one add Puppet as a trusted publisher? Talk to your security team?

beechtom: My understanding is that you're considered a trusted publisher if the script is signed with a valid, unexpired digital signature issued by a certificate authority. @jpogran would know better since he's much more familiar with Windows, but I don't think we need to cover that since it only applies to users running old versions of Bolt.

hestonhoffman: @beechtom After consulting with James, I don't think we need to add anything else to this edit 👍

hestonhoffman (approved):

Issue #2475, opened 2020-12-14
Don't merge `name => all` to inventoryfile before resolving reference


lucywyman:
## Describe the Bug
Bolt should support an inventoryfile that only contains a plugin reference:
```
---
_plugin: yaml
filepath: ../bolt/inventory.yaml
```

## Additional context

The inventory schema will also need to be updated to accept a plugin at the top level.

PR #2503, opened 2020-12-18
(GH-2475) Allow entire inventory to be a plugin reference

beechtom:
This change allows an entire inventory to be a plugin reference. For
exmaple, the following inventory was previously invalid:

```yaml
---
_plugin: yaml
filepath: /path/to/inventory_partial.yaml
```

A top-level plugin like this used to be invalid as Bolt automatically
added `'name' => 'all'` to the top of the loaded inventory, which would
cause a plugin to receive a `name` parameter and (likely) error. Now,
Bolt will resolve a top-level plugin reference, validate that the
resolved data does not include a `name` key, and then merge in `'name'
=> 'all'`.

!bug

* **Allow entire inventory to be specified with a plugin**
  ([#2475](https://github.com/puppetlabs/bolt/issues/2475))

  Inventory files can now be specified with a plugin. For example, the
  following inventory file is now valid:

  ```yaml
  ---
  _plugin: yaml
  filepath: /path/to/inventory_partial.yaml
  ```

### DIFF
--- a/lib/bolt/inventory/group.rb
+++ b/lib/bolt/inventory/group.rb
@@ -18,11 +18,19 @@ module Bolt
       GROUP_KEYS = DATA_KEYS + %w[name groups targets]
       CONFIG_KEYS = Bolt::Config::INVENTORY_OPTIONS.keys
 
-      def initialize(input, plugins)
+      def initialize(input, plugins, all_group: false)
         @logger = Bolt::Logger.logger(self)
         @plugins = plugins
 
         input = @plugins.resolve_top_level_references(input) if @plugins.reference?(input)
+
+        if all_group
+          if input.key?('name') && input['name'] != 'all'
+            @logger.warn("Top-level group '#{input['name']}' cannot specify a name, using 'all' instead.")
+          end
+
+          input = input.merge('name' => 'all')
+        end
 
         raise ValidationError.new("Group does not have a name", nil) unless input.key?('name')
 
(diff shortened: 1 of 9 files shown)


lucywyman:
Ugh, needs a rebase :( Sorry I totally forgot about this!

lib/bolt/inventory/group.rb
```diff
@@ -18,12 +18,18 @@ class Group
...
 
-      def initialize(input, plugins)
+      def initialize(input, plugins, all_group: false)
         @logger = Bolt::Logger.logger(self)
         @plugins = plugins
 
         input = @plugins.resolve_top_level_references(input) if @plugins.reference?(input)
 
+        if all_group && input.key?('name')
+          raise ValidationError.new("Top-level group 'all' cannot specify name '#{input['name']}'", nil)
```
lucywyman, line 28: I don't understand why we want to prevent this - it's currently allowed at the top level, and will get overwritten by the merge in `inventory.rb`. This seems unnecessary and risky?

lucywyman, line 28: Can we warn that the name will be overridden and have this error in 3.0? Just because the current behavior allows it and while it seems like a bug, it'd be a pain for people to rewrite around.

lucywyman (approved):
Awesome! Verified this works with a top-level plugin and nested plugins and doesn't use additional keys (like a top level `targets` key). Also verified the group name merging warns if a name is set from the plugin, and that any group named all raises the collision error.

Issue #2505, opened 2020-12-18
Do not add localhost target to all group in PowerShell


beechtom:
## Describe the Bug
https://github.com/puppetlabs/bolt/pull/2395 introduced a bug that added the `localhost` target to the inventory on Windows. The `localhost` target should not be added to the inventory by default.

PR #2506, opened 2020-12-18
(GH-2505) Do not add localhost to the all group by default in PowerShell

beechtom:
This fixes a bug introduced in #2395 that adds the `localhost` target to
the inventory by default. This results in any use of the `all` group
when running in PowerShell to include the `localhost` target in
inventory. Now, Bolt will not add the `localhost` target to inventory by
default.

!bug

* **Do not add `localhost` target to the `all` group by default in
  PowerShell**
  ([#2505](https://github.com/puppetlabs/bolt/issues/2505))

  Bolt no longer adds the `localhost` target to the `all` group by
  default. Previously, when running Bolt in PowerShell, the `localhost`
  target would be added to the `all` group unintentionally.

### DIFF
--- a/lib/bolt/cli.rb
+++ b/lib/bolt/cli.rb
@@ -212,12 +212,15 @@ module Bolt
 
     private def validate_ps_version
       if Bolt::Util.powershell?
-        target = inventory.get_target('localhost')
-        Bolt::Transport::Local.new.with_connection(target) do |conn|
-          # This will automatically validate the powershell version on the Bolt
-          # controller
-          Bolt::Shell::Powershell.new(target, conn)
-        end
+        command = "powershell.exe -NoProfile -NonInteractive -NoLogo -ExecutionPolicy "\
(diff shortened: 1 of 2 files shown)


lib/bolt/cli.rb
```diff
@@ -212,12 +212,13 @@ def finalize_setup
 
     private def validate_ps_version
       if Bolt::Util.powershell?
-        target = inventory.get_target('localhost')
-        Bolt::Transport::Local.new.with_connection(target) do |conn|
-          # This will automatically validate the powershell version on the Bolt
-          # controller
-          Bolt::Shell::Powershell.new(target, conn)
-        end
+        stdout, _stderr, _status = Open3.capture3('pwsh -Command $PSVersionTable.PSVersion.Major')
```
lucywyman, line 215: The way this currently works, the command is actually `powershell.exe -NoProfile -NonInteractive -NoLogo -ExecutionPolicy Bypass -Command $PSVersionTable.PSVersion.Major`.

PR #2512, opened 2021-01-04
(PE-30862) Delete transport config keys that resolved to nil

ekinanp:
We need this for the upcoming PuppetConnectData plugin where we want to
enforce the semantic that "this key resolves to nil" == "this key was
not specified".

### DIFF
--- a/lib/bolt/inventory/group.rb
+++ b/lib/bolt/inventory/group.rb
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 
+require 'bolt/config/options'
 require 'bolt/inventory/group'
 require 'bolt/inventory/inventory'
 require 'bolt/inventory/target'
@@ -315,7 +316,26 @@ module Bolt
           'features' => @plugins.resolve_references(data.fetch('features', [])),
           'plugin_hooks' => @plugins.resolve_references(data.fetch('plugin_hooks', {}))
         }
+
         validate_data_keys(result, target)
+
+        Bolt::Config::Options::TRANSPORT_CONFIG.each_key do |transport|
+          next unless result['config'].key?(transport)
+          transport_config = result['config'][transport]
+          next unless transport_config.is_a?(Hash)
+          transport_config = Bolt::Util.postwalk_vals(transport_config) do |val|
+            if val.is_a?(Hash)
+              val = val.compact
+              val = nil if val.empty?
+            end
+            val
+          end
+          # the transport config is user-specified data so we
+          # still want to preserve it even if it exclusively
+          # contains nil-resolved keys
+          result['config'][transport] = transport_config || {}
+        end
+
         result['features'] = Set.new(result['features'].flatten)
         result
       end
--- a/spec/bolt/inventory/inventory_spec.rb
+++ b/spec/bolt/inventory/inventory_spec.rb
@@ -1351,4 +1351,90 @@ describe Bolt::Inventory::Inventory do
       expect(target.transport).to eq('local')
     end
   end
+
+  context 'when the transport config contains keys that resolved to nil' do
+    let(:transport_config) do
+      {
+        'transport' => 'ssh',
+        'ssh' => {
+          'user' => {
+            '_plugin' => 'mock_plugin'
+          },
+          'private-key' => {
+            'key-data' => {
+              '_plugin' => 'mock_plugin'
+            }
+          },
+          'password' => 'foo_password'
+        }
+      }
+    end
+
+    before(:each) do
+      mock_plugin = double('mock_plugin')
+      allow(mock_plugin).to receive(:name).and_return('mock_plugin')
+      allow(mock_plugin).to receive(:hooks).and_return([:resolve_reference])
+      allow(mock_plugin).to receive(:resolve_reference).and_return(nil)
+      plugins.add_plugin(mock_plugin)
+    end
+
+    context 'for a target' do
+      let(:targets) do
+        [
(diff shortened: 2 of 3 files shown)


ekinanp:
Does this need a release note? If so, what's an example commit that shows the right formatting for this specific PR?

lib/bolt/inventory/group.rb
```diff
@@ -315,6 +316,13 @@ def resolve_data_keys(data, target = nil)
           'features' => @plugins.resolve_references(data.fetch('features', [])),
           'plugin_hooks' => @plugins.resolve_references(data.fetch('plugin_hooks', {}))
         }
+
+        # Delete keys that resolved to nil from the transport config
+        Bolt::Config::Options::TRANSPORT_CONFIG.each_key do |transport|
+          next unless result['config'].key?(transport)
+          result['config'][transport].compact!
```
donoghuc, line 323: does `compact` work on nested nils?

beechtom:
@ekinanp If there are user-facing changes there should be a release note. More info on the format is in the [contributing docs](https://github.com/puppetlabs/bolt/blob/main/CONTRIBUTING.md#pull-requests).

If there aren't user-facing changes, then you can just use the `!no-release-note` label in your commit.

ekinanp:
@beechtom Sounds good. Do you think these are user-facing?

lib/bolt/inventory/group.rb
ekinanp: Do we need to account for nested nils? If so, an example would help.

beechtom:
@ekinanp I would say they are since config values that are `nil` will no longer be returned in a config hash. Compare the return for `$target.config` before this change:

```
{
  "ssh": {
    "host-key-check": false
  },
  "transport": "winrm",
  "winrm": {
    "user": null
  }
}
```

And with this change:

```
{
  "ssh": {
    "host-key-check": false
  },
  "transport": "winrm",
  "winrm": {
  }
}
```

lib/bolt/inventory/group.rb
```diff
@@ -315,6 +316,17 @@ def resolve_data_keys(data, target = nil)
...
+        if result['config'].is_a?(Hash)
+          Bolt::Config::Options::TRANSPORT_CONFIG.each_key do |transport|
+            next unless result['config'].key?(transport)
+            transport_config = result['config'][transport]
+            next unless transport_config.is_a?(Hash)
+            result['config'][transport].compact!
+          end
+        end
+
         validate_data_keys(result, target)
```
beechtom, line 330: I think we can just move this line further up to validate each value's type before compacting the transport config. Then there's no need for a separate check that `config` is a hash.

lib/bolt/inventory/group.rb
donoghuc: I think a common one would be:

```
config:
  ssh:
    private-key:
      key-data: 
        _plugin: foo
```
```
(base) cas@cas-ThinkPad-T460p:~/working_dir/bolt$ irb
irb(main):001:0> nested = {'private-key' => { 'key-data' => nil}}
=> {"private-key"=>{"key-data"=>nil}}
irb(main):002:0> nested.compact
=> {"private-key"=>{"key-data"=>nil}}
```

ekinanp: Ok, that makes sense. Updated.

lib/bolt/inventory/group.rb
ekinanp: Updated.

donoghuc (approved):

lib/bolt/inventory/group.rb
```diff
@@ -315,7 +316,26 @@ def resolve_data_keys(data, target = nil)
...
+
+        Bolt::Config::Options::TRANSPORT_CONFIG.each_key do |transport|
+          next unless result['config'].key?(transport)
+          transport_config = result['config'][transport]
+          next unless transport_config.is_a?(Hash)
+          transport_config = Bolt::Util.postwalk_vals(transport_config) do |val|
+            if val.is_a?(Hash)
+              val.compact!
```
nicklewis, line 328: I'm concerned about making in-place updates to an object that we didn't necessarily construct. Specifically I'm unsure how this will affect future targets getting their config from the same group. We should make a copy here I think.

nicklewis: I also think we don't need to prune empty hashes, since we do a deep merge. Empty hashes won't have any affect on the final merged result in that case. Combined with the previous comment, that would make this block just

```rb
if val.is_a?(Hash)
  val.compact
else
  val
end
```

Potentially we would prefer not to make a copy of the hash in the common case where it _doesn't_ contain any `nil` values, but I don't think that matters too much.

ekinanp: I'm fine with changing to a copy.

Re: the second comment, would returning an empty hash still be OK for _target_ config? E.g. will a target-config that's something like `{"ssh": {"private-key": {}}}` cause any issues?

donoghuc: Yeah i think the private-key case is somewhat special. currently if you allow it to be an empty hash you get a schema error. That is a weird one but i imagine it would be a common one for connect.

ekinanp: In that case, I lean towards the current behavior of deleting keys that correspond to empty hashes.

nicklewis (approved):

donoghuc (approved):

Issue #2452, opened 2020-12-09
Move test container provisioning out of spec tests.


lucywyman:
## Use Case
We currently do a lot of provisioning of our docker containers that we use for testing in the tests themselves. We should move that out to the Docker config files themselves - specifically installing the correct Puppet agent version to the Puppet 5, Puppet 6, and Puppet 7 containers.

PR #2513, opened 2021-01-04
(GH-2452) Provision Docker containers with puppet-agent package

beechtom:
This updates the Dockerfile and Docker compose files to provision the
agentful containers used in spec tests with the appropriate
`puppet-agent` package. Previously, the `puppet-agent` packages were
installed as part of the spec tests themselves.

!no-release-note

### DIFF
--- a/spec/integration/apply_spec.rb
+++ b/spec/integration/apply_spec.rb
@@ -69,13 +69,6 @@ describe 'apply', expensive: true do
     # Run tests that require the puppet-agent package to be installed on the target.
     # Each test is run against all agent targets unless otherwise noted.
     context 'with puppet installed' do
-      # Install agents on the agent targets.
-      before(:all) do
-        TEST_VERSIONS.each do |version, collection|
-          install("puppet_#{version}_node", collection: collection, inventory: docker_inventory(root: true))
-        end
-      end
-
       # Set up a project directory for the tests. Include an inventory file so Bolt
       # can actually connect to the targets.
       around(:each) do |example|
--- a/spec/integration/plan_spec.rb
+++ b/spec/integration/plan_spec.rb
@@ -5,14 +5,12 @@ require 'bolt_spec/conn'
 require 'bolt_spec/files'
 require 'bolt_spec/integration'
 require 'bolt_spec/project'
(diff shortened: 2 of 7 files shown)


spec/integration/apply_spec.rb
```diff
@@ -67,13 +67,6 @@
     # Run tests that require the puppet-agent package to be installed on the target.
     # Each test is run against all agent targets unless otherwise noted.
     context 'with puppet installed' do
-      # Install agents on the agent targets.
-      before(:all) do
-        TEST_VERSIONS.each do |version, collection|
```
lucywyman, line 72: I think this can also refactor tests that install and uninstall the agent on the main container to use the containers with agents instead, like I see that in `spec/integration/plan_spec.rb#L73` we install and uninstall the agent around each test instead of just using the containers with the agent installed. Or same in `spec/integration/device_spec.rb`. I think we still need the `uninstall` utility to uninstall after tests that verify installing the agent with apply_prep works, but we should be able to remove the `install` spec helper function with this PR.

beechtom: The `install` helper is still needed by the apply integration tests on Windows. Updated the plan and device integration tests to use the agentful containers.

spec/integration/plan_spec.rb
```diff
@@ -70,28 +68,35 @@
...
-          config_flags = %W[--format json -m #{modulepath} --project #{project.path} --no-host-key-check]
+      context 'with puppetfile subcommand' do
+        let(:project_config) { {} }
+        let(:subcommand)     { 'puppetfile' }
 
-          run_cli(%w[module generate-types] + config_flags)
-          result = run_cli(['plan', 'run', 'resource_types', '--targets', target] + config_flags)
-          expect(JSON.parse(result)).to eq('built-in' => 'success', 'core' => 'success', 'custom' => 'success')
-        end
+        include_examples 'registered types'
```
lucywyman, line 92: So fancy, yet so short-lived. I guess some stars are brightest before they die :star2:

beechtom: 😂

PR #2521, opened 2021-01-07
(PE-30766) Add the puppet_connect_data plugin to Bolt

ekinanp:
!no-release-note

### DIFF
--- a/lib/bolt/plugin.rb
+++ b/lib/bolt/plugin.rb
@@ -139,7 +139,7 @@ module Bolt
       plugins
     end
 
-    RUBY_PLUGINS = %w[task prompt env_var puppetdb].freeze
+    RUBY_PLUGINS = %w[task prompt env_var puppetdb puppet_connect_data].freeze
     BUILTIN_PLUGINS = %w[task terraform pkcs7 prompt vault aws_inventory puppetdb azure_inventory
                          yaml env_var gcloud_inventory].freeze
     DEFAULT_PLUGIN_HOOKS = { 'puppet_library' => { 'plugin' => 'puppet_agent', 'stop_service' => true } }.freeze
--- a/lib/bolt/plugin/puppet_connect_data.rb
+++ b/lib/bolt/plugin/puppet_connect_data.rb
@@ -1,0 +1,35 @@
+# frozen_string_literal: true
+
+module Bolt
+  class Plugin
+    class PuppetConnectData
+      def initialize(context:, **_opts)
+        puppet_connect_data_yaml_path = File.join(context.boltdir, 'puppet_connect_data.yaml')
+        @data = Bolt::Util.read_optional_yaml_hash(
+          puppet_connect_data_yaml_path,
+          'puppet_connect_data.yaml'
+        )
+      end
+
+      def name
+        'puppet_connect_data'
+      end
+
+      def hooks
+        %i[resolve_reference validate_resolve_reference]
+      end
+
+      def resolve_reference(opts)
+        key = opts['key']
+        @data[key]
+      end
+
+      def validate_resolve_reference(opts)
+        unless opts['key']
+          raise Bolt::ValidationError,
+                "puppet_connect_data plugin requires that 'key' be specified"
+        end
+      end
+    end
+  end
+end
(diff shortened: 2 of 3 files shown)


documentation/supported_plugins.md
```diff
@@ -209,3 +210,152 @@ config:
...
+We plan on adding caching support in a later iteration of the plugin. Please
+file an issue if you'd like us to expedite this process.
+
+**Note:** You do not have to specify auto-loaded, `puppet_connect_data`-referenced
+SSH config in your `puppet_connect_data.yaml` file. The `puppet_connect_data` plugin
+resolves all unspecified keys to `nil`, which tells Bolt to remove the corresponding
+config keys from the target's transport config (meaning SSH config will be auto-loaded
+if that was the previous, pre-`puppet_connect_data` plugin behavior).
+
+#### Stored connection data
```
donoghuc, line 335: I dont think this detail belongs in the bolt docs.

donoghuc: I wonder if this whole section belongs in experimental? https://github.com/lucywyman/bolt/blob/main/documentation/experimental_features.md

donoghuc (commented):
The only thing that is a bit strange about this is that the plugin works everywhere you would expect a plugin to work, however we only get the "special" delete keys that are nil functionalitly if the plugin is being used to resolve config. I'm not sure what the alternative is or if this is actually a problem.

As a super simple example:
puppet_connect_data.yaml
```yaml
target:
  - name: foo
    config:
      transport: local
```
inventory.yaml
```yaml
targets:
  - _plugin: puppet_connect_data
    key: target
```
This works fine, but if you delete the key in puppet_connect_data.yaml you get an error.

documentation/supported_plugins.md
ekinanp: Hmm, perhaps that could be shortened to defining what we consider to be "secret" data. Don't really have any opinions on that.

Re: experimental, not sure. Maybe it could go to  a special `Puppet Connect` section? Or be included in the Puppet Connect docs? Probably worth asking about during product check-in later today.

ekinanp:
@donoghuc The "delete `nil` transport config keys" is not a Puppet Connect thing, it's what Bolt does w/ my recent change. Thus, it will occur for any plugin that resolves a transport config key to `nil`. I don't think that is a problem since for _transport config_, I don't see how adding any other meaning for `nil` could be useful to the end user (and I also made sure to note in the docs that `nil` keys will only be deleted in transport config. It might be worth making that a more general note in plugin reference resolution since it does affect all plugins, don't know what would be a good spot for that info).

Re: your example, I'm not sure what you mean by the error.

donoghuc (approved):
The plugin implementation works as described. Dont block on the docs question or the general behavior of squashing nils.

donoghuc:
@ekinanp the "error" in the example i posted is that you cant have `nil` as a target. If that list had compacted it would have not errored and the entry would be ignored completely. I dont know how we would implement that though.

ekinanp:
@donoghuc Ah ok. Isn't that already an error with the current setup though? Seems fine to me to leave it as-is. Although that introduces an inconsistency with transport config, I don't think it's a bad one because nil-deletion makes sense there while it may not make sense for a `targets` array (and even if it does, it's not worth doing unless a user asks for it and gives us a good reason to implement it).

documentation/supported_plugins.md
```diff
@@ -209,3 +210,152 @@ config:
       parameters:
         key: ssh_password
 ```
+
+### `puppet_connect_data`
```
ekinanp, line 214: @hestonhoffman This plugin's meant to be used by Puppet Connect users. Although it is a supported plugin that we (Skeletor) do want to ship with Bolt, I'm not sure if we should document this here, in Puppet Connect-specific docs, or both. I've been told you're a good person to ask about this, so am welcoming any ideas/suggestions here.

lucywyman: Do we expect Bolt users who aren't also Connect users to use this plugin? If not, I don't think it should be included here and should just be in Connect docs with a note that the plugin can still be used with Bolt without breaking.

ekinanp: I think this plugin will only be used by Connect users, so putting it in the Connect docs seems like the way to go. I'll remove them from this PR.

Should the plugin still have a release note in the commit?

lucywyman: Probably not - you can do `!no-release-note` so it doesn't fail the check.

ekinanp: Updated to remove the docs + add `!no-release-note`.

donoghuc (approved):

nicklewis (approved):

Issue #2450, opened 2020-12-09
Update deprecated YAML plan steps in spec tests


lucywyman:
source key for YAML plan upload step renamed to upload
target key for YAML plan steps renamed to targets

PR #2524, opened 2021-01-07
(GH-2450) Remove deprecated yaml plan keys from spec tests 

lucywyman:
This replaces deprecated YAML plan keys `source` with `upload` and
`target` with `targets` in the spec tests, so that these tests continue
to work once the deprecated keys are removed.

!no-release-note


beechtom:
I think https://github.com/puppetlabs/bolt/blob/main/spec/bolt/pal/yaml_plan/step_spec.rb#L84 should be `upload`

beechtom (approved):

PR #2526, opened 2021-01-07
(docs) List which config options are pluggable

lucywyman:
This updates the `Using Plugins` document to tell users that not all 
configuration options are pluggable, and how to request a new option be
made pluggable. This also updates the configuration reference templates
to add a `Pluggable` list item for configuration options that are 
pluggable. The `Using Plugins` page points to this reference for users
that want to see which options are pluggable.

!no-release-note

### DIFF
--- a/documentation/using_plugins.md
+++ b/documentation/using_plugins.md
@@ -78,6 +78,13 @@ soon as the inventory file is loaded, whereas reference plugins under data keys
 such as `config` or `facts` are resolved once Bolt starts running an action on
 that target.
 
+While plugins are supported for configuration files, they're disabled for some settings. You can see
+which options are available on configuration references pages for
+[bolt-project.yaml](bolt_project_reference.md) and [bolt-defaults.yaml](bolt_defaults_reference.md).
+If you'd like another setting to have plugins enabled, let us know [in #bolt in
+slack](https://slack.puppet.com) or by [making an
+issue](https://github.com/puppetlabs/bolt/issues/new/choose).
+
 📖 **Related information**
 
 - [Configuring Bolt](configuring_bolt.md)
--- a/documentation/templates/bolt_configuration_reference.md.erb
+++ b/documentation/templates/bolt_configuration_reference.md.erb
@@ -26,6 +26,9 @@ _This option is deprecated. <%= data[:_deprecation] %>_
 <% if data.key?(:_default) -%>
 - **Default:** <%= data[:_default] %>
 <% end -%>
+<% if data[:_plugin] == true -%>
+- **Pluggable:** true
+<% end -%>
 
 <% if data.key?(:properties) -%>
 <% data[:properties].each do |option, data| -%>
@@ -39,6 +42,9 @@ _This option is deprecated. <%= data[:_deprecation] %>_
 <% end -%>
 <% if data.key?(:_default) -%>
 - **Default:** <%= data[:_default] %>
+<% end -%>
+<% if data[:_plugin] == true -%>
+- **Pluggable:** true
 <% end -%>
 <% end %>
 <% end -%>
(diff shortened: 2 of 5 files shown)


documentation/using_plugins.md
```diff
@@ -78,6 +78,11 @@ soon as the inventory file is loaded, whereas reference plugins under data keys
 such as `config` or `facts` are resolved once Bolt starts running an action on
 that target.
 
+While plugins are supported for configuration files, they're disable for some settings. Plugins are
```
hestonhoffman, line 81:
```suggestion
While plugins are supported for configuration files, they're disabled for some settings. Plugins are
```

documentation/using_plugins.md
```diff
@@ -78,6 +78,11 @@ soon as the inventory file is loaded, whereas reference plugins under data keys
 such as `config` or `facts` are resolved once Bolt starts running an action on
 that target.
 
+While plugins are supported for configuration files, they're disabled for some settings. Plugins are
+enabled for `plugin-hooks`, `plugins`, `puppetdb` and subsettings, and `inventory-config` and any
+subsettings. If you'd like another setting to have plugins enabled, let us know [in #bolt in
+slack](https://slack.puppet.com) or by making an issue.
```
beechtom, line 84:
```suggestion
slack](https://slack.puppet.com) or by [making an issue](https://github.com/puppetlabs/bolt/issues/new/choose).
```

documentation/using_plugins.md
```diff
@@ -78,6 +78,11 @@ soon as the inventory file is loaded, whereas reference plugins under data keys
 such as `config` or `facts` are resolved once Bolt starts running an action on
 that target.
 
+While plugins are supported for configuration files, they're disabled for some settings. Plugins are
+enabled for `plugin-hooks`, `plugins`, `puppetdb` and subsettings, and `inventory-config` and any
+subsettings. If you'd like another setting to have plugins enabled, let us know [in #bolt in
```
beechtom, line 83:
```suggestion
enabled for `plugin-hooks`, `plugins`, `puppetdb` and subsettings, and transport configuration
under `inventory-config`. If you'd like another setting to have plugins enabled, let us know [in #bolt in
```

`inventory-config` itself is not pluggable.

beechtom (commented):
I'm wondering if it might be helpful for the ref pages to say whether or not a given option is pluggable, and then link to those. Then we don't have to keep this page updated.

lucywyman:
That's a great idea! Updated to include a `Pluggable` item for options that are pluggable in the reference docs - I think adding this just for options that are pluggable instead of for everything makes it easier to search the page to see what's pluggable, and helps make the page less huge and overwhelming.

beechtom:
Looks good! 👍 I think the transport config reference could use this change as well.

lucywyman:
Sounds good - I initially didn't since I think all options are pluggable for the transports, but it does seem confusing if `Pluggable` is set on some config pages and not others. Added!

beechtom (approved):

documentation/using_plugins.md
```diff
@@ -78,6 +78,13 @@ soon as the inventory file is loaded, whereas reference plugins under data keys
 such as `config` or `facts` are resolved once Bolt starts running an action on
 that target.
 
+While plugins are supported for configuration files, they're disabled for some settings. You can see
+which options are available on configuration references pages for
+[bolt.yaml](bolt_configuration_reference.md), [bolt-project.yaml](bolt_project_reference.md), and
```
hestonhoffman, line 83: Do we want bolt.yaml here? My vote would be to leave it out cos it's one less thing to whackamole when we hit 3.0

lucywyman: Great point, removed it!

hestonhoffman (approved):

PR #2533, opened 2021-01-11
(maint) Pin ffi to < 1.14.0

beechtom:
This updates the gemspec to pin ffi to < 1.14.0, which includes some
breaking changes on Windows. Previously, ffi was pinned to ~> 1.13.0,
which caused our packaging pipeline to fail, as bolt-runtime only
includes ffi 1.9.25.

!no-release-note

### DIFF
--- a/bolt.gemspec
+++ b/bolt.gemspec
@@ -46,7 +46,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency "aws-sdk-ec2", '~> 1'
   spec.add_dependency "CFPropertyList", "~> 2.2"
   spec.add_dependency "concurrent-ruby", "~> 1.0"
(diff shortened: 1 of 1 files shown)


kaorihinata:
Just a note that pinning `ffi` to `< 1.14.0` will block future Apple Silicon support. `1.14.0` was the first release to include support for the new architecture (see: https://github.com/ffi/ffi/issues/841.)

lucywyman:
Oh, that is great to know! I think we'll get lucky with this one, as we can unpin this once we release Bolt 3.0 with Ruby 2.7 minimum version, which should be before we're able to support Big Sur or Silicon (as we're waiting on the infrastructure for testing the platform and hardware). But if either of those doesn't happen it's great to be aware of!

PR #2537, opened 2021-01-12
(PE-30856) Add and use the builtin-content-dir setting

ekinanp:

### DIFF
--- a/lib/bolt_server/transport_app.rb
+++ b/lib/bolt_server/transport_app.rb
@@ -280,7 +280,8 @@ module BoltServer
         bolt_config = config_from_project(versioned_project)
         modulepath_object = Bolt::Config::Modulepath.new(
           bolt_config.modulepath,
-          boltlib_path: [PE_BOLTLIB_PATH, Bolt::Config::Modulepath::BOLTLIB_PATH]
+          boltlib_path: [PE_BOLTLIB_PATH, Bolt::Config::Modulepath::BOLTLIB_PATH],
+          builtin_content_path: @config['builtin-content-dir']
         )
         pal = Bolt::PAL.new(modulepath_object, nil, nil, nil, nil, nil, bolt_config.project)
         context = {
--- a/lib/bolt_server/config.rb
+++ b/lib/bolt_server/config.rb
@@ -9,7 +9,7 @@ module BoltServer
     def config_keys
       super + %w[concurrency cache-dir file-server-conn-timeout
                  file-server-uri projects-dir environments-codedir
-                 environmentpath basemodulepath]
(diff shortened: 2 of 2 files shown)


nicklewis (approved):

lib/bolt_server/transport_app.rb
```diff
@@ -280,7 +280,8 @@ def in_bolt_project(versioned_project)
         bolt_config = config_from_project(versioned_project)
         modulepath_object = Bolt::Config::Modulepath.new(
           bolt_config.modulepath,
-          boltlib_path: [PE_BOLTLIB_PATH, Bolt::Config::Modulepath::BOLTLIB_PATH]
+          boltlib_path: [PE_BOLTLIB_PATH, Bolt::Config::Modulepath::BOLTLIB_PATH],
+          builtin_content_path: @config['builtin-content-dir']
```
donoghuc, line 284: Will this mess up the built in modules that were previously the default? 
https://github.com/puppetlabs/bolt/blob/e765bdd69468de3c73dbd3f65c210c5711abc868/lib/bolt/config/modulepath.rb#L16

ekinanp: It will mean that those modules aren't loaded anymore. Right now I see:

```
[root@voltaic-blemish bolt-2.40.2]# ls modules/
aggregate  canary  puppetdb_fact
```

If we do want those modules with our builtin content, then we should probably talk about whether they should be added to the `pe-builtin-bolt-content` package (to keep it separate from Bolt) or if we should just accept that some of our builtin content will include stuff that Bolt ships. Latter seems to make more sense to me.

donoghuc: I think we 100% need to make sure those built ins are available. We have floated the idea of breaking them out into modules outside of bolt. Maybe this is the time for that.

ekinanp: @McdonaldSeanp thoughts on ^? Should these modules be in `pe-builtin-bolt-content` or should we just include `MODULES` along with our `builtin-content-dir` in the modulepath's `builtin_content_path` option?

nicklewis: We could either do the latter or we could configure the `builtin-content-dir` setting to include what is currently set as `MODULES`.

ekinanp: We agreed to move it over to the setting, so am resolving this.

ekinanp:
@donoghuc Since we agreed in standup to move the default `modules` builtin content to the `builtin-content-dir` setting, I think the code here remains unchanged so this should be good to go.

donoghuc (approved):

Issue #2511, opened 2021-01-04
Don't print the spinner while prompting from a plan


lucywyman:
## Describe the Bug
We print a spinner while running plans, including if the `prompt` or `resolve_references` plan functions are called:
```
lucy ~/githubs/bolt (main) $ be bolt plan run bolt -t docker
Starting: plan bolt
Prompt: no spin\
Finished: plan bolt in 5.3 sec
Plan completed successfully with no result
```
This doesn't affect the user input as the spinner spins in a separate process, so isn't input into the prompt. But it is annoying and potentially confusing to have the spinner present in the input area on the CLI.

## Expected Behavior
Don't print the spinner while prompting the user for input

## Steps to Reproduce
```
plan bolt (){
  prompt("Prompt")

  $ref =  {
    "targets" => [
      "_plugin" => "prompt",
      "message" => "RR prompt"
    ]   
  }
  resolve_references($ref)
}
```

## Solution
- We should just start/stop the spinner from the executor for certain functions so that this isn't printed while running the whole plan but just while executing individual functions.
- Ensure that we do the spinning while plan code isn't executing, to avoid something like spinning while in `run_task` but `run_task` is paused and another fiber is prompting.

PR #2538, opened 2021-01-12
(GH-2511) Only spin while executing run_* plan functions

lucywyman:
Previously the spinner would spin for the entirety of a plan or task
run, starting and stopping from the Bolt CLI manager. This meant that if
the user was prompted during a plan run the spinner would spin in the
area of the terminal the user was supposed to input text. While this
didn't write to stdin or impact user input, it was annoying and
definitely wrong.

This modifies the spinner to be started and stopped from the executor,
impacting calls from the command line and plan functions. The executor
only starts and stops the spinner as part of `with_node_logging()`,
which affects `run_*` functions and `wait_until_available`. The spinner
still spins for module management and project updating commands, which
are unaffected by this change. The spinner will only be started and
stopped if Bolt is not executing inside a `parallelize` block - instead
the parallelize block itself (specifically `Bolt::Executor#round_robin`)
will manage starting and stopping the spinner. This is so that the first
element in the array that finishes doesn't signal to stop spinning
before other functions have completed, but the spinner will spin for the
entire parallelize block as expected.

Test cases:
- [x] The spinner spins while executing executor plan functions
- [x] The spinner spins while executing `bolt command run`, `bolt script
  run`, `bolt task run`, `bolt file upload`, and `bolt file download`,
  and module management commands
- [x] The spinner spins normally while running multiple functions in a
  parallelize block across multiple targets
- [x] Spinner does not spin when running other plan functions or Bolt
  commands (in particular when prompting for user input)

!bug

* **Only spin while executing run_* plan functions** ([#2511](https://github.com/puppetlabs/bolt/issues/2511))
  Bolt will now only print the spinner while executing `run_*`,
  `file_upload`, `file_download`, and`wait_until_available` plan
  functions. It also now spins while running those functions equivalent
  commandline commands. This prevents the spinner from spinning while
  prompting for output from a plan.

### DIFF
--- a/bolt-modules/boltlib/lib/puppet/functions/parallelize.rb
+++ b/bolt-modules/boltlib/lib/puppet/functions/parallelize.rb
@@ -12,14 +12,12 @@ Puppet::Functions.create_function(:parallelize, Puppet::Functions::InternalFunct
   # @param data The array to apply the block to.
   # @return [Array] An array of PlanResult objects. Each input from the input
   #   array returns a corresponding PlanResult object.
-  # @example Execute two tasks on multiple targets. Once the task finishes on one
-  #   target, that target can move to the next step without waiting for the task
-  #   to finish on the second target.
-  # $targets = get_targets(["host1", "host2"])
-  # $result = parallelize ($targets) |$t| {
-  #   run_task('a', $t)
-  #   run_task('b', $t)
-  # }
+  # @example Execute two tasks on two targets.
+  #   $targets = get_targets(["host1", "host2"])
+  #   $result = parallelize ($targets) |$t| {
+  #     run_task('a', $t)
+  #     run_task('b', $t)
+  #   }
   dispatch :parallelize do
     scope_param
     param 'Array[Any]', :data
--- a/lib/bolt/outputter/human.rb
+++ b/lib/bolt/outputter/human.rb
@@ -32,8 +32,8 @@ module Bolt
       end
 
       def start_spin
-        return unless @spin && @stream.isatty
-        @spin = true
+        return unless @spin && @stream.isatty && !@spinning
+        @spinning = true
         @spin_thread = Thread.new do
           loop do
             sleep(0.1)
@@ -43,9 +43,9 @@ module Bolt
       end
 
       def stop_spin
-        return unless @spin && @stream.isatty
+        return unless @spin && @stream.isatty && @spinning
+        @spinning = false
         @spin_thread.terminate
-        @spin = false
         @stream.print("\b")
       end
 
@@ -81,6 +81,10 @@ module Bolt
             print_plan_start(event)
           when :plan_finish
             print_plan_finish(event)
+          when :start_spin
+            start_spin
+          when :stop_spin
+            stop_spin
           end
         end
       end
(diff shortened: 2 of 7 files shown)


bolt-modules/boltlib/lib/puppet/functions/parallelize.rb
```diff
@@ -12,14 +12,12 @@
...
   #   array returns a corresponding PlanResult object.
-  # @example Execute two tasks on multiple targets. Once the task finishes on one
-  #   target, that target can move to the next step without waiting for the task
-  #   to finish on the second target.
-  # $targets = get_targets(["host1", "host2"])
-  # $result = parallelize ($targets) |$t| {
-  #   run_task('a', $t)
-  #   run_task('b', $t)
-  # }
+  # @example Execute two tasks on across two targets.
```
beechtom, line 15:
```suggestion
  # @example Execute two tasks on two targets.
```

lib/bolt/outputter/human.rb
```diff
@@ -43,9 +43,9 @@ def start_spin
       end
 
       def stop_spin
-        return unless @spin && @stream.isatty
+        return unless @spin && @stream.isatty && @spinning
```
beechtom, line 46:
```suggestion
        return unless @spin && @stream.isatty && @spinning
```

There's something wonky happening with checking the state of the spinner and starting/stopping it when using the `parallelize` function. Running a plan that calls `parallelize`, followed by the `prompt` function, results in the spinner showing for a prompt. Here's an example plan that shows this behavior:

```puppet
plan bolt (
  TargetSpec $targets = 'localhost',
) {
  $_targets = get_targets($targets)

  $results = parallelize($_targets) |$target| {
    out::message('hello')
  }

  $prompt = prompt('prompt')

  return $results
}
```

Removing `@spinning` from the conditional fixes it, so I suspect that's just not being set correctly when calling `parallelize`?

lucywyman: Hmmm, I'm not seeing that...maybe we can talk about it at standup tomorrow? The `@spinning` needs to be there so that multiple spinners aren't printed when in a parallelize block.

beechtom:
Hrm, I'm still seeing a spinner when prompting, whether it's in a `parallelize` block or after.

image

lucywyman:
Ok, I modified this to spin pessimistically while in parallel: that is, it sends the `stop_spinning` signal before doing anything that might not be a long-running process (like run_task). If that doesn't work, I'm interested in seeing the plan that's causing it - it could just be a race condition.

beechtom:
I've been using the plan that I posted a bit above in a comment. It just has a `parallelize` block with a single `out::message` call, and then the `prompt` function call _after_ the `parallelize` block. That's no longer printing a spinner when prompting. However, adding a function call to the `parallelize` block that triggers the spinner causes the spinner to show up when prompting:

image

Here's the plan I'm using:

```
plan bolt (
  TargetSpec $targets = 'localhost',
) {
  $_targets = get_targets($targets)

  $results = parallelize($_targets) |$target| {
    out::message('hello')
    run_task('facts', $target, '_catch_errors' => true)
  }

  $prompt = prompt('prompt')

  return $results
}
```

beechtom (approved):

Issue #2522, opened 2021-01-07
Support non-github git repos


nmaludy:
## Use Case
We are not allowed to host our code in external infrastructure, so all of our private code lives in on-prem git repos.

It would be great if `bolt module add` and `bolt module install` supported non-github git repos.

## Describe the Solution You Would Like
The ability to specify non-github git repos in the `modules` setting of `bolt-project.yaml`

lucywyman:
## Solution for now
It may be a while before we can support all the module declarations people are interested in using (or modules that don't resolve). For now, a good workaround for this would be to add an option to individual modules to not resolve them with the `puppetfile-resolver`.

This should be an option for both Forge and Git module types that will omit the module from being part of the dependency resolution step when installing modules but will include the module in the final Puppetfile.
```
modules:
  - git: https://mygithmodule
    ref: main
    resolve: false
```
- What happens if a dependency has the same name as a module that doesn't resolve? Check existing behavior.

PR #2544, opened 2021-01-14
(GH-2522) Add 'resolve' key to modules config

beechtom:
This adds support for a new `resolve` key for both Forge and git module
specs in `bolt-project.yaml`. It also adds support for a `name` key for
git module specs, which is required when setting the `resolve` key to
`false`.

When `resolve` is `false`, Bolt will skip dependency resolution for the
module. This allows users to include modules that would otherwise break
Bolt's dependency resolution in their Bolt projects, avoiding the need
to manually manage a Puppetfile. This includes modules with broken
metadata and git modules hosted in a repository other than a public
GitHub repository.

If a unresolved module shares the name of another module's dependency,
Bolt will error indicating that there is a name conflict. This is in
line with r10k's behavior when installing a Puppetfile that has
duplicate module names.

!feature

* **Add `resolve` key for Forge and git module specifications**
  ([#2522](https://github.com/puppetlabs/bolt/issues/2522))

  Forge and git module specifications in `bolt-project.yaml` now support
  a `resolve` key. When setting `resolve: false`, Bolt will skip
  dependency resolution for the module, allowing users to include
  modules with broken metadata or modules hosted in a repository other
  than a public GitHub repository in their project configuration.

---

### Test cases

- [ ] Requires git modules to have a `name` if `resolve: false`
- [ ] Skips resolution for modules with `resolve: false` when installing and adding modules
- [ ] Allows non-public or non-GitHub git modules to be installed without `--no-resolve`
- [ ] Allows modules with broken metadata to be installed
- [ ] Errors if no-resolve modules have a name conflict with resolved dependencies

### DIFF
--- a/documentation/bolt_installing_modules.md
+++ b/documentation/bolt_installing_modules.md
@@ -185,6 +185,7 @@ To specify a Forge module, use the following keys in the specification:
 | --- | --- | :-: |
 | `name` | The full name of the module. Can be either `<OWNER>-<NAME>` or `<OWNER>/<NAME>`. | ✓ |
 | `version_requirement` | The version requirement that the module must satisfy. Can be either a specific semantic version (`1.0.0`), a version range (`>= 1.0.0 < 3.0.0`), or version shorthand (`1.x`).  |
+| `resolve` | Boolean. Whether to resolve the module's dependencies when installing modules. | |
 
 After you've made changes to the module specification, [update your
 modules](#update-a-projects-modules).
@@ -214,7 +215,9 @@ To specify a git module, use the following keys in the specification:
 | Key | Description | Required |
 | --- | --- | :-: |
 | `git` | The URI to the GitHub repository. URI must begin with either `https://github.com`or `git@github.com`. | ✓ |
+| `name` | The name of the module. Bolt uses this name for the module in the Puppetfile, the directory that the module's contents are downloaded to, and as a namespace for the module's content. To avoid errors, make sure this name matches the name specified in the module's `metadata.json`. **Required if `resolve` is `false`.** | |
 | `ref` | The git reference to checkout. Can be either a branch, commit, or tag. | ✓ |
+| `resolve` | Boolean. Whether to resolve the module's dependencies when installing modules. | |
 
 After you've made changes to the module specification, [update your
 modules](#update-a-projects-modules).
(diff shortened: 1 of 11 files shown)


lucywyman:
Woops, quick rebase.

lib/bolt/module_installer.rb
```diff
@@ -195,8 +195,10 @@ def install_puppetfile(path, moduledir, config = {})
       @outputter.stop_spin
 
       # Automatically generate types after installing modules
-      @outputter.print_action_step("Generating type references")
-      @pal.generate_types(cache: true)
+      if ok
```
lucywyman, line 198: When is this false?

beechtom: Anytime installing modules fails. I specifically ran into this when a Puppetfile has duplicate module names -- r10k will error and fail to install modules, but Bolt will continue to try and generate types.

lucywyman: It seems like we should raise an error here if `ok` is false in that case.

lib/bolt/config/options.rb
```diff
@@ -260,6 +260,10 @@ module Options
                     description: "The name of the module.",
                     type: String
                   },
+                  "resolve" => {
+                    description: "Whether to resolve the module's dependencies when installing modules.",
+                    type: [TrueClass, FalseClass]
```
lucywyman, line 265: Hm, this weirdly doesn't seem to error with non-boolean values are provided. Do we not run the validator when doing `module install`?

beechtom: No, the validator would need to be updated to handle the `oneOf` key since there are multiple valid definitions for a module spec. I can add in a separate check as part of the module objects to ensure that it's a boolean.

lib/bolt/module_installer.rb
beechtom: My only hesitation with doing that is we explicitly have a check in the outputter to print a message stating that installing the modules failed https://github.com/puppetlabs/bolt/blob/main/lib/bolt/outputter/human.rb#L428-L434 . I'm fine with making the change, though, if we want to error instead.

lib/bolt/config/options.rb
lucywyman: That'd be great.

lucywyman: Actually, it seems like the validator should be checking every key that's there against it's type even if there are multiple possible configurations (basically should be able to handle `oneOf`). It looks like this is the only instance of `oneOf`, so it might be appropriate to add here.

documentation/bolt_installing_modules.md
```diff
@@ -329,9 +332,98 @@ module-install:
...
+name: myproject
+modules:
+  - puppetlabs/ntp
+  - name: private_module
+    git: git@github.com:puppetlabs/private_module
+    ref: 1.0.0
+    resolve: false
+```
+
+> **Note:** When setting `resolve: false` for a git module, you must include
```
lucywyman, line 371: I think it's also worth calling out (here or elsewhere) that the value of `name` is how the module will be named on the modulepath as well.

beechtom: Added some more info to the table further up the page that lists the available keys.

spec/bolt/module_installer/specs/git_spec_spec.rb
```diff
@@ -44,6 +44,15 @@
         /Invalid git source/
       )
     end
+
+    it 'errors if resolve is false and missing a name' do
```
lucywyman, line 48: Since it's a separate check it'd also be good to verify that `resolve: <non-boolean>` errors for both forge and git modules.

lucywyman (changes requested):

documentation/bolt_installing_modules.md
```diff
@@ -214,7 +215,9 @@ To specify a git module, use the following keys in the specification:
 | Key | Description | Required |
 | --- | --- | :-: |
 | `git` | The URI to the GitHub repository. URI must begin with either `https://github.com`or `git@github.com`. | ✓ |
+| `name` | The name of the module. This name is used to name the module in the Puppetfile and the directory that the module's contents are downloaded to. Required if `resolve` is `false`. | |
```
adreyer, line 218: This should be more explicit about whether it's namespaced or not.

documentation/bolt_installing_modules.md
```diff
@@ -329,9 +332,98 @@ module-install:
...
+
+Sometimes you might want to skip dependency resolution for a specific module,
+usually because Bolt is unable to resolve its dependencies. This allows you to
+take advantage of Bolt's module dependency resolution while still using modules
+that Bolt cannot resolve dependencies for. You might want to skip dependency
+resolution for a module if:
+
+- The module has outdated or incorrect metadata.
+- The module is a git module hosted in a repository other than a public GitHub
+  repository.
```
adreyer, line 345: + The modules dependencies cannot be cleanly resolved

documentation/bolt_installing_modules.md
beechtom: Updated. Not 100% sure on the phrasing, though... for example, if the name doesn't match the module's _actual_ name, tasks can still be run but not plans.

lucywyman:
- [x] Does not install dependencies when `resolve: false`
- [x] Installs dependencies when `resolve` is:
  - [x] Unset
  - [x] True
- [x] Errors if resolve isn't boolean
- [x] Errors if resolve is set for a git module and name isn't set
- [x] Git module with a name different from actual name uses provided name
- [x] Adding a fake module with `resolve: false` doesn't fail until installing modules
  - [x] Adding a fake module with `resolve: false` doesn't prevent other modules from being installed
- [x] Errors when `resolve: false` module has the same name as a module dependency

lucywyman (approved):
Just needs docs review.

documentation/bolt_installing_modules.md
```diff
@@ -185,6 +185,7 @@ To specify a Forge module, use the following keys in the specification:
 | --- | --- | :-: |
 | `name` | The full name of the module. Can be either `<OWNER>-<NAME>` or `<OWNER>/<NAME>`. | ✓ |
 | `version_requirement` | The version requirement that the module must satisfy. Can be either a specific semantic version (`1.0.0`), a version range (`>= 1.0.0 < 3.0.0`), or version shorthand (`1.x`).  |
+| `resolve` | Whether to resolve the module's dependencies when installing modules. | |
```
hestonhoffman, line 188:
```suggestion
| `resolve` | Boolean. Whether to resolve the module's dependencies when installing modules. | |
```

documentation/bolt_installing_modules.md
```diff
@@ -214,7 +215,9 @@ To specify a git module, use the following keys in the specification:
 | Key | Description | Required |
 | --- | --- | :-: |
 | `git` | The URI to the GitHub repository. URI must begin with either `https://github.com`or `git@github.com`. | ✓ |
+| `name` | The name of the module. This name is used to name the module in the Puppetfile and the directory that the module's contents are downloaded to and is used as a namespace for the module's content. To avoid errors, make sure this name matches the name specified in the module's `metadata.json`. **Required if `resolve` is `false`.** | |
```
hestonhoffman, line 218:
```suggestion
| `name` | The name of the module. Bolt uses this name for the module in the Puppetfile, the directory that the module's contents are downloaded to, and as a namespace for the module's content. To avoid errors, make sure this name matches the name specified in the module's `metadata.json`. **Required if `resolve` is `false`.** | |
```

documentation/bolt_installing_modules.md
```diff
@@ -214,7 +215,9 @@ To specify a git module, use the following keys in the specification:
...
 | `ref` | The git reference to checkout. Can be either a branch, commit, or tag. | ✓ |
+| `resolve` | Whether to resolve the module's dependencies when installing modules. | |
```
hestonhoffman, line 220:
```suggestion
| `resolve` | Boolean. Whether to resolve the module's dependencies when installing modules. | |
```

documentation/bolt_installing_modules.md
```diff
@@ -329,9 +332,99 @@ module-install:
 - [bolt-project.yaml options](bolt_project_reference.md#module-install)
 - [bolt-defaults.yaml options](bolt_defaults_reference.md#module-install)
 
+## Skip dependency resolution for a module
+
+Sometimes you might want to skip dependency resolution for a specific module,
+usually because Bolt is unable to resolve its dependencies. This allows you to
+take advantage of Bolt's module dependency resolution while still using modules
+that Bolt cannot resolve dependencies for. You might want to skip dependency
+resolution for a module if:
```
hestonhoffman, line 341:
```suggestion
Skipping dependency resolution for a module allows you to
take advantage of Bolt's module dependency resolution while still using modules
that Bolt cannot resolve dependencies for.

You might want to skip dependency resolution for a module if:
```
Just streamlining this a bit because it's a little redundant with the list

documentation/bolt_installing_modules.md
```diff
@@ -329,9 +332,99 @@ module-install:
...
+
+- The module has outdated or incorrect metadata.
+- The module is a git module hosted in a repository other than a public GitHub
+  repository.
+- The module's dependencies cannot be cleanly resolved.
```
hestonhoffman, line 346:
```suggestion
- Bolt can't cleanly resolve the module's dependencies.
```

lucywyman (approved):
Will merge once Heston approves as well (no rush Heston!) :sparkle:

hestonhoffman (approved):
Docs LGTM!

PR #2563, opened 2021-01-15
(maint) Refactor the Docker transport to use Shell class

lucywyman:
This refactors the Docker transport to use the same code paths that
other transports do, in an effort to prove the viability of making
transports pluggable. The behavior of the transport should not change at
all.

The `Bolt::Transport::Docker` class now inherits from the `Simple`
transport class, and only overloads the `with_connection` function to
create a new `Bolt::Transport::Docker::Connection` instance and yield
it.

The `Bolt::Transport::Docker::Connection` class provides the functions
that the shell classes need to interact with Docker containers:
`connect`, `upload_file`, `download_file`, and `execute`. These are the 
functions we're hoping are needed for a stable API for making transports
pluggable. This update largely consists of renaming functions to match
the API, but also includes distinguishing the `execute` function that the 
shell classes use that must return pipes and the 
`execute_local_docker` functions that are used internally by the 
Connection class and just return strings.

!no-release-note

### DIFF
--- a/lib/bolt/transport/docker.rb
+++ b/lib/bolt/transport/docker.rb
@@ -6,7 +6,7 @@ require 'bolt/transport/base'
 
 module Bolt
   module Transport
-    class Docker < Base
+    class Docker < Simple
       def provided_features
         ['shell']
       end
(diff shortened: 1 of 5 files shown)


lucywyman:
I think this needs particularly special attention to:
- [ ] Running tasks that rely on environment variables
  - [ ] From the CLI
  - [ ] From plans
  - [ ] With `run_as`
  - [ ] Are there any special environment variables that break this?
- [ ] Running tasks that rely on stdin
  - [ ] With `run_as`
  - [ ] Without `run_as`

beechtom:
This needs a quick rebase.

lib/bolt/transport/docker.rb
```diff
@@ -16,130 +16,6 @@ def with_connection(target)
...
-          Bolt::Result.for_download(target, source, destination, download)
-        end
-      end
-
-      def run_command(target, command, options = {}, position = [])
-        execute_options = {}
-        execute_options[:tty] = target.options['tty']
-        execute_options[:environment] = options[:env_vars]
-
-        if target.options['shell-command'] && !target.options['shell-command'].empty?
```
nicklewis, line 54: It looks like this is no longer honored. It seems wrong that it would only have been used for `run_command` and not all `execute` calls though...

lucywyman: Good catch! Updated to use this in the `execute` function, so it should be applied to all commands. Also added a test so it'll get caught next time.

lib/bolt/transport/docker/connection.rb
```diff
@@ -29,203 +48,113 @@ def connect
...
-        def execute(*command, options)
-          command.unshift(options[:interpreter]) if options[:interpreter]
-          # Build the `--env` parameters
-          envs = []
-          if options[:environment]
-            options[:environment].each { |env, val| envs.concat(['--env', "#{env}=#{val}"]) }
+        # @param command [string] The command to run
+        def execute(command)
+          env_vars = []
+          # CODEREVIEW: Is there a better way to handle `docker exec ... PT_foo` not working well?
```
beechtom, line 61: This definitely seems strange, and I think the weirdness comes from the shell class primarily handling getting the command in the right format for the shell, but the bash shell has some built-in assumptions for passing env vars that don't play well with the `docker exec` command. I'm not sure how to get around it without

1. Passing `options` and building the full command from `conn.execute` instead of the shell class, or
1. Having the shell class call a method on the connection class to format the command correctly, or
1. Having a separate 'docker' shell that builds docker-specific commands

lib/bolt/transport/docker/connection.rb
```diff
@@ -29,203 +48,113 @@ def connect
...
-          command_options = []
-          # Need to be interactive if redirecting STDIN
-          command_options << '--interactive' unless options[:stdin].nil?
-          command_options << '--tty' if options[:tty]
-          command_options.concat(envs) unless envs.empty?
-          command_options << container_id
-          command_options.concat(command)
+          args = []
+          # CODEREVIEW: Is it always safe to pass --interactive?
+          args += %w[--interactive]
```
beechtom, line 70: It doesn't seem like it'd be an issue. If we're really concerned about always keeping STDIN open, we could either pass `options` to this method or handle building the command a different way (see above comment).

beechtom (commented):
Works as expected, just some remaining questions re: building the command string.

lib/bolt/transport/docker/connection.rb
lucywyman: Mmk, I think part of the problem here is that I'm not sure how many user-provided transports are going to need to define their own environment handling or not. Without overengineering anything, I ended up with logic that says "If the connection define and environment variable handler then use it, otherwise use the default" which is the handler that our existing transports use.This seems to work well enough for now without the obscene parsing-env-vars-out-of-the-string, and we can adjust it based on peoples needs if we ever get to the point of writing transport plugins.

lib/bolt/transport/docker/connection.rb
lucywyman: Yeah, I think similar to above it's hard to know if any other transports will need to know about 'options' or not, and if *this* was the only thing that ended up needing it that'd make me sad to have to pass them around and accept them *everywhere* just for this. Unless you (or anyone else) is concerned I think we can leave it here for now and can change it if it causes issues or if other transports need to know about options.

lucywyman:
This is ready for re-review whenever y'all get a chance! I'd love to unblock the streaming PR and LXD transport.

beechtom (approved):
Runs all the basic CLI commands, `run_*` functions from plans, and passes environment variables correctly.

Issue #2542, opened 2021-01-13
Let users silence warnings


lucywyman:
## Todo
- Ensure warnings have 1-word IDs
- When issuing a warning include the warning ID at the end of the warning
- Users can silence warnings with the following in either `bolt-project.yaml` or `bolt-defaults.yaml`
  - Lists should be merged if set in both (in merge_config_layers)
```
disable-warnings:
  - puppet_5
```

PR #2568, opened 2021-01-21
(GH-2542) Add disable-warnings config option to disable warnings

beechtom:
This adds a new `disable-warnings` config option that is available in
both `bolt-project.yaml` and `bolt-defaults.yaml`. This option accepts
an array of warning IDs that are used to suppress warnings issued by
Bolt. If a warning would otherwise be issued, and its ID is under the
`disable-warnings` option, Bolt will not print the warning to either the
CLI or a log file.

If `disable-warnings` is configured in multiple config layers, Bolt will
merge the arrays between config layers.

To enable this feature, a new `Bolt::Logger.warn` method was added that
accepts an ID and message. If the ID is present in the
`disable-warnings` config, the message will not be logged.

Lastly, this adds a queuing mechanic to `Bolt::Logger`. This work
includes the new methods `deprecate`, `deprecate_once`, `info`, and
`debug`; and updates the existing `warn_once` method. These methods now
log messages via the `log` method. If `Bolt::Logger` has been
configured, the messages are immediately logged. Otherwise, the messages
are added to a queue to be logged once `Bolt::Logger` is configured. The
`flush_queue` method can be called to flush the message queue and log
all messages in the queue.

!feature

* **Suppress warnings with `disable-warnings` config option**
  ([#2542](#2542))

  The `disable-warnings` configuration option accepts an array of
  warning IDs that are used to suppress warnings in both the CLI and log
  files. This configuration option is supported in both
  `bolt-project.yaml` and `bolt-defaults.yaml`.

---

### Acceptance critera

- [ ] Suppresses deprecation warnings with a matching ID
  - [ ] Does not print deprecation to console
  - [ ] Does not print deprecation to log file
- [ ] Suppresses regular warnings with a matching ID
  - [ ] Does not print deprecation to console
  - [ ] Does not print deprecation to log file
- [ ] Does not suppress warnings without a matching ID
- [ ] Merges warning IDs between multiple config layers
- [ ] Does not deadlock

### DIFF
--- a/documentation/logs.md
+++ b/documentation/logs.md
@@ -178,6 +178,31 @@ Bolt to output additional information in a human-readable format. Verbose output
 useful for debugging your tasks and plans - if you're not sure why something is failing, try running
 it with `--verbose` to get more information.
 
+## Suppress warnings
+
+You can suppress warning messages from being logged by Bolt. To disable specific
+warnings, add a list of warning IDs to the `disable-warnings` configuration
+option in `bolt-project.yaml` or `bolt-defaults.yaml`. If defined in both files,
+the lists of warnings are concatenated per the [configuration merge
+strategy](configuring_bolt.md#merge-strategy). For example, if Bolt issues the
+following warning:
+
+```shell
+The configuration option 'apples' is deprecated. Use 'oranges' instead. [ID: fruit_option]
+```
+
+You could add the ID `fruit_option` under the `disable-warnings` configuration option:
+
+```
+# bolt-project.yaml
+---
+name: myproject
+disable-warnings:
+  - fruit_option
+```
+
+The next time you run Bolt, the warning message will not be logged.
+
 📖 **Related information**  
 
 - [Debugging tasks](writing_tasks.md#debugging-tasks)
--- a/lib/bolt/cli.rb
+++ b/lib/bolt/cli.rb
@@ -147,10 +147,6 @@ module Bolt
       end
 
       validate(options)
-
-      # Deprecation warnings can't be issued until after config is loaded, so
-      # store them for later.
-      @parser_deprecations = parser.deprecations
     rescue Bolt::Error => e
       fatal_error(e)
       raise e
@@ -185,17 +181,13 @@ module Bolt
 
     # Completes the setup process by configuring Bolt and log messages
     def finalize_setup
-      Bolt::Logger.configure(config.log, config.color)
+      Bolt::Logger.configure(config.log, config.color, config.disable_warnings)
       Bolt::Logger.analytics = analytics
+      Bolt::Logger.flush_queue
 
       # Logger must be configured before checking path case and project file, otherwise logs will not display
       config.check_path_case('modulepath', config.modulepath)
       config.project.check_deprecated_file
-
-      # Log messages created during parser and config initialization
-      config.logs.each { |log| @logger.send(log.keys[0], log.values[0]) }
-      @parser_deprecations.each { |dep| Bolt::Logger.deprecation_warning(dep[:type], dep[:msg]) }
-      config.deprecations.each { |dep| Bolt::Logger.deprecation_warning(dep[:type], dep[:msg]) }
 
       if options[:clear_cache] && File.exist?(config.project.plugin_cache_file)
         FileUtils.rm(config.project.plugin_cache_file)
@@ -220,7 +212,7 @@ module Bolt
 
         msg = "Detected PowerShell 2 on controller. PowerShell 2 is deprecated and "\
               "support will be removed in Bolt 3.0."
-        Bolt::Logger.deprecation_warning("PowerShell 2 controller", msg)
+        Bolt::Logger.deprecate("powershell_2_controller", msg)
       end
     end
 
@@ -376,7 +368,10 @@ module Bolt
       conflicting_options = Set.new(opts.keys.map(&:to_s)).intersection(inventory_cli_opts)
 
       if inventory_source && conflicting_options.any?
-        @logger.warn("CLI arguments #{conflicting_options.to_a} may be overridden by Inventory: #{inventory_source}")
+        Bolt::Logger.warn(
+          "cli_overrides",
+          "CLI arguments #{conflicting_options.to_a} may be overridden by Inventory: #{inventory_source}"
+        )
       end
     end
 
@@ -662,7 +657,7 @@ module Bolt
         if node_param && target_param
           msg = "Plan parameters include both 'nodes' and 'targets' with type 'TargetSpec', " \
                 "neither will populated with the value for --nodes or --targets."
-          @logger.warn(msg)
+          Bolt::Logger.warn("nodes_targets_parameters", msg)
         elsif node_param
           plan_arguments['nodes'] = nodes.join(',')
         elsif target_param
@@ -709,7 +704,7 @@ module Bolt
                   "about defining and declaring classes and types in the Puppet documentation at "\
                   "https://puppet.com/docs/puppet/latest/lang_classes.html and "\
                   "https://puppet.com/docs/puppet/latest/lang_defined_types.html"
-        @logger.warn(message)
+        Bolt::Logger.warn("empty_manifest", message)
       end
 
       executor = Bolt::Executor.new(config.concurrency, analytics, noop, config.modified_concurrency)
@@ -861,7 +856,7 @@ module Bolt
       elsif modules.nil? && options[:subcommand] == 'puppetfile'
         msg = "Command '#{old_command}' is deprecated and will be removed in Bolt 3.0. Update your project to use "\
               "the module management feature. For more information, see https://pup.pt/bolt-module-migrate."
-        Bolt::Logger.deprecation_warning('puppetfile command', msg)
+        Bolt::Logger.deprecate("puppetfile_command", msg)
       elsif modules.nil? && options[:subcommand] == 'module'
         msg  = "Unable to use command '#{new_command}' when 'modules' is not configured in "\
                "bolt-project.yaml. "
@@ -966,7 +961,7 @@ module Bolt
           set the BOLT_GEM environment variable.
         MSG
 
-        @logger.warn(msg)
+        Bolt::Logger.warn("gem_install", msg)
       end
 
       # We only need to enumerate bundled content when running a task or plan
--- a/lib/bolt/config.rb
+++ b/lib/bolt/config.rb
@@ -20,7 +20,7 @@ module Bolt
   class Config
     include Bolt::Config::Options
 
-    attr_reader :config_files, :logs, :data, :transports, :project, :modified_concurrency, :deprecations
+    attr_reader :config_files, :data, :transports, :project, :modified_concurrency
 
     BOLT_CONFIG_NAME = 'bolt.yaml'
     BOLT_DEFAULTS_NAME = 'bolt-defaults.yaml'
@@ -33,9 +33,6 @@ module Bolt
     end
 
     def self.from_project(project, overrides = {})
-      logs         = []
-      deprecations = []
-
       conf = if project.project_file == project.config_file
                project.data
              else
@@ -45,31 +42,26 @@ module Bolt
                # with all validation errors.
                Bolt::Validator.new.tap do |validator|
                  validator.validate(c, bolt_schema, project.config_file.to_s)
-
-                 validator.warnings.each { |warning| logs << { warn: warning } }
-
-                 validator.deprecations.each do |dep|
-                   deprecations << { type: "#{BOLT_CONFIG_NAME} #{dep[:option]}", msg: dep[:message] }
-                 end
+                 validator.warnings.each { |warning| Bolt::Logger.warn(warning[:id], warning[:msg]) }
+                 validator.deprecations.each { |dep| Bolt::Logger.deprecate(dep[:id], dep[:msg]) }
                end
 
-               logs << { debug: "Loaded configuration from #{project.config_file}" } if File.exist?(project.config_file)
+               if File.exist?(project.config_file)
+                 Bolt::Logger.debug("Loaded configuration from #{project.config_file}")
+               end
                c
              end
+
       data = load_defaults(project).push(
-        filepath:     project.config_file,
-        data:         conf,
-        logs:         logs,
-        deprecations: deprecations
+        filepath: project.config_file,
+        data: conf
       )
 
       new(project, data, overrides)
     end
 
     def self.from_file(configfile, overrides = {})
-      project      = Bolt::Project.create_project(Pathname.new(configfile).expand_path.dirname)
-      logs         = []
-      deprecations = []
+      project = Bolt::Project.create_project(Pathname.new(configfile).expand_path.dirname)
 
       conf = if project.project_file == project.config_file
                project.data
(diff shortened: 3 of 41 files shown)


lucywyman (changes requested):
So the way this currently works, only `warn_once` or `deprecation_warning` warnings are able to be disabled, which excludes a lot of warnings that come from per-target loggers that can't use either of those methods. I think for this to work for all warnings we'll need to add a function `warn` to Bolt::Logger that does *basically* what `warn_once` does without checking to see if we've already warned:

```
    def self.warn(type, msg)
      return if @disabled_warnings.include?(type)

      @mutex.synchronize do
        @logger ||= Bolt::Logger.logger(self)
        @logger.warn("#{msg} [ID: #{type}]")
      end 
    end
```

lib/bolt/logger.rb
```diff
@@ -123,18 +130,54 @@ def self.reset_logging
...
+      @message_queue.each do |message|
+        case message[:type]
+        when :deprecation
+          deprecation_warning(id: message[:id], msg: message[:msg], src: message[:src])
+        when :warn
+          warn(id: message[:id], msg: message[:msg], src: message[:src])
+        else
+          logger(message[:src]).send(message[:type], message[:msg])
+        end
+      end
```
beechtom, line 180: While not strictly necessary to enable this change, these methods help to simplify handling messages logged before the logger is configured. I don't believe this should cause any issues / race conditions, as the `Logging` library handles creating `Logging` instances.

lib/bolt/logger.rb
```diff
@@ -123,18 +130,54 @@ def self.reset_logging
...
 
-    def self.deprecation_warning(type, msg)
-      @analytics&.event('Warn', 'deprecation', label: type)
-      warn_once(type, msg)
+    # Issues a deprecation warning unless the user has suppressed
+    # it. Submits analytics for the deprecation regardless of
+    # whether the warning is suppressed.
+    def self.deprecation_warning(id:, msg:, src: self)
+      @analytics&.event('Warn', 'deprecation', label: id)
+      warn(id: id, msg: msg, src: src)
```
beechtom, line 159: I believe this was originally `warn_once` to handle deprecations logged when connecting to targets (specifically PowerShell 2 deprecations). I switched this over to just use `warn` to handle cases where a user has a deprecated config option in multiple config layers.

Leaving this as `warn_once` for the config option case would result in the warning for each config layer having its own ID (`apply_settings_user`, `apply_settings_system`, `apply_settings_default`), otherwise users wouldn't see a warning for each config layer. However, using `warn` results in _every_ target with PowerShell 2 logging a deprecation warning... I wonder if we might want a separate `deprecation_warn_once` method?

lib/bolt/logger.rb
```diff
@@ -123,18 +130,54 @@ def self.reset_logging
...
-        @logger ||= Bolt::Logger.logger(self)
-        if @warnings.add?(type)
-          @logger.warn(msg)
+        Bolt::Logger.logger(src).warn("#{msg} [ID: #{id}]")
+      end
+    end
+
+    # Issues a warning unless the user has suppressed it,
+    # or a warning with the same ID has already been issued.
+    def self.warn_once(id:, msg:, src: self)
```
beechtom, line 144: Adding in `src` fixes a (very minor) bug in the logger that shows all `warn_once` and `deprecation_warning` messages as coming from `Bolt::Logger` in a log file.

beechtom:
@nicklewis If you have some spare time, I'd appreciate some feedback on the changes to `Bolt::Logger`, specifically `Bolt::Logger.queue_messages` and `Bolt::Logger.issue_queued_messages`. The intent is to simplify logging messages that come up before the logger is configured (i.e. when parsing the command and loading the project/config) so we don't need to pass around arrays of messages between multiple classes.

documentation/logs.md
```diff
@@ -178,6 +178,28 @@ Bolt to output additional information in a human-readable format. Verbose output
 useful for debugging your tasks and plans - if you're not sure why something is failing, try running
 it with `--verbose` to get more information.
 
+## Suppress warnings
+
+You can suppress warning messages from being logged by Bolt. To disable specific warnings, add a list
+of warning IDs to the `disable-warnings` configuration option in `bolt-project.yaml` or
+`bolt-defaults.yaml`. For example, if Bolt issues the following warning:
```
lucywyman, line 185:
```suggestion
`bolt-defaults.yaml`. If defined in both files, the list of warnings is shallowly merged per the [configuration merge strategy](configuring_bolt.md#merge-strategy). For example, if Bolt issues the following warning:
```

lib/bolt/cli.rb
```diff
@@ -186,17 +182,16 @@ def load_config
     # Completes the setup process by configuring Bolt and log messages
     def finalize_setup
       Bolt::Logger.configure(config.log, config.color)
+      Bolt::Logger.disable_warnings = config.disable_warnings
```
lucywyman, line 185: This seems like it should either be part of `Bolt::Logger.configure` or `disable-warnings` should be under the `log` key in config.

lib/bolt/logger.rb
```diff
@@ -5,8 +5,11 @@
 module Bolt
   module Logger
     LEVELS = %w[trace debug info notice warn error fatal].freeze
-    @mutex = Mutex.new
-    @warnings = Set.new
+
+    @mutex            = Mutex.new
+    @warnings         = Set.new
+    @disable_warnings = Set.new
```
lucywyman, line 11: Nick made a good point that we're basically treating this module as a global singleton. I think that's probably OK here - it needs to be global since multiple classes need to access these variables, and all need to access the same instance of those variables. We need to be very careful that accessing the variables is thread-safe though, so any reference to any of these variables needs to be wrapped in a `@mutex.synchronize`.

I'd leave a comment above these variables explaining that we're treating the module like a global singleton so that multiple classes can add warnings.

We also might eventually want to make these their own class that can be instantiated if this gets too hard to test or track, but I think for now this is simple enough we don't need that.

lucywyman: Honestly since all of the messages we queue follow a single path to get here, I think it's simpler to just keep passing them through Config to the CLI to here, and avoid having global vars that have to be synched and introduce risk.

lib/bolt/config.rb
```diff
@@ -593,8 +651,9 @@ def plugins
     def plugin_hooks
       if @data['plugin-hooks'].any? && @data['plugin_hooks'].any?
         Bolt::Logger.warn_once(
-          "plugin-hooks and plugin_hooks set",
-          "Detected configuration for 'plugin-hooks' and 'plugin_hooks'. Bolt will ignore 'plugin_hooks'."
+          id:  "plugin_hooks_conflict",
+          msg: "Detected configuration for 'plugin-hooks' and 'plugin_hooks'. Bolt will ignore 'plugin_hooks'.",
+          src: self
```
lucywyman, line 656: Definitely wish there was a better way to just get the calling class from the method instead of having to do this *everywhere* :P

beechtom: Ya :(

nicklewis: Would it be cleaner to just extend to `Logging::Logger` class with a `warn_once` method? That does make sharing the cache a little tricky though...

lib/bolt/logger.rb
```diff
@@ -86,6 +91,10 @@ def self.analytics=(analytics)
       @analytics = analytics
     end
 
+    def self.disable_warnings=(disable_warnings)
```
lucywyman, line 94: I don't think we need this anymore

lib/bolt/logger.rb
lucywyman: `deprecation_warn_once` is probably the simplest way to solve this. I think we may eventually want to have a warning hierarchy of sorts, where one ID can refer to multiple warnings (so `apply_settings` would map to `apply_settings_user`, `apply_settings_system`, and `apply_settings_default`, but users could also disable any of those warnings individually if they wanted to suppress `apply_settings_default` but not `apply_settings_user`) but adding that now really complicates things and I'm not sure it's needed yet.

lib/bolt/logger.rb
```diff
@@ -123,18 +131,59 @@ def self.reset_logging
...
+    def self.deprecation_warning(id:, msg:, src: self)
+      @analytics&.event('Warn', 'deprecation', label: id)
+      warn(id: id, msg: msg, src: src)
+    end
+
+    # Queues a log message to be issued at a later time. This is used when
+    # parsing the command and loading the project/config, as the logger is not
+    # configured until after that happens.
+    #
+    def self.queue_message(type:, msg:, id: nil, src: nil)
```
lucywyman, line 166: I don't think 'id' can ever be nil.

lib/bolt/logger.rb
beechtom: I disagree. `Bolt::Logger` is already treated as a global singleton due to the `warn_once`/`deprecation_warning` methods and the `@warnings`/`@disable_warnings` variables. In my opinion, passing around log messages between multiple methods in multiple classes adds more complexity by making the code more difficult to understand and maintain in the long-term. Adding an extra `@mutex.synchronize` block around the new methods seems really low risk considering we already do that in this module. 🤷

lib/bolt/logger.rb
beechtom: https://github.com/puppetlabs/bolt/pull/2568/files#diff-69231076f18e469d86a14ebeb845fecfaec1d2adf4543f5513363c8123b051c3R51-R56

lib/bolt/logger.rb
```diff
@@ -123,18 +131,59 @@ def self.reset_logging
       Logging.reset
     end
 
-    def self.warn_once(type, msg)
+    # Issues a warning unless the user has suppressed it.
+    #
+    def self.warn(id:, msg:, src: self)
```
nicklewis, line 136: If you made this method take a block which is expected to return the message when yielded (the other log methods support that syntax as well), then you could use `block.binding.receiver` to get the caller. Kind of weird, but should work.

beechtom: Hrm... that'd require calling the method with a block, wouldn't it? I.e.

```
Bolt::Logger.warn(...) do; end
```

spec/bolt/cli_spec.rb
```diff
@@ -775,14 +775,14 @@ def stub_config(file_content = {})
       end
 
       it "warns when using debug" do
-        expect(Bolt::Logger).to receive(:deprecation_warning)
-          .with(anything, /Command line option '--debug' is deprecated/)
+        expect(Bolt::Logger).to receive(:queue_message)
```
lucywyman, line 778: This only verifies that the message gets queued, not printed. It's better to verify the behavior in this case:
```
expect(@log_output.readlines).to include(/Command line option '--debug' is deprecated/)
```

spec/bolt/logger_spec.rb
```diff
@@ -45,7 +45,7 @@ def initialize(*args)
     it 'submits an analytics event' do
       expect(analytics).to receive(:event).with('Warn', 'deprecation', { label: "We've got clearance Clarence" })
       Bolt::Logger.analytics = analytics
-      Bolt::Logger.deprecation_warning("We've got clearance Clarence", 'Roger Roger')
```
lucywyman, line 48: This should also include a test for the logic in `self.issue_queued_messages`, particularly that the message queue gets cleared at the end.

spec/spec_helper.rb
```diff
@@ -71,6 +71,12 @@
     # Ignore user's known hosts and ssh config files
     conf = { user_known_hosts_file: '/dev/null/', global_known_hosts_file: '/dev/null' }
     allow(Net::SSH::Config).to receive(:for).and_return(conf)
+
+    # Allow logger to receive messages
+    allow(Bolt::Logger).to receive(:queue_message).and_call_original
```
lucywyman, line 76: These shouldn't be needed, what was failing?

lucywyman (changes requested):

lib/bolt/logger.rb
lucywyman: :mask: Ok, can you add a comment though?

lib/bolt/logger.rb
```diff
@@ -123,18 +131,59 @@ def self.reset_logging
...
       end
     end
 
-    def self.deprecation_warning(type, msg)
-      @analytics&.event('Warn', 'deprecation', label: type)
-      warn_once(type, msg)
+    # Log messages in the message queue.
+    #
+    def self.log_queued_messages
+      @message_queue.each do |message|
```
lucywyman, line 175: This needs a mutex sync.

lib/bolt/logger.rb
nicklewis: Yeah it would be something like `Bolt::Logger.warn('problem_occurred') { "Something terrible has happened" }`

hestonhoffman (commented):
Docs look good. There's a bit of passive voice, but in the context of these docs, I think that's okay. 

One interesting style thing. A while back, we went through all the docs and tried to remove the foo/bar/baz placeholders. IIRC, it was part of a larger push to improve our examples. It kinda gives us an opportunity to be a little more creative and create use groupings that might be a bit easier for non-coding folks to grasp (see https://puppet.com/docs/pe/latest/nodes_endpoint.html).

documentation/logs.md
```diff
@@ -178,6 +178,31 @@ Bolt to output additional information in a human-readable format. Verbose output
...
+option in `bolt-project.yaml` or `bolt-defaults.yaml`. If defined in both files,
+the lists of warnings are concatenated per the [configuration merge
+strategy](configuring_bolt.md#merge-strategy). For example, if Bolt issues the
+following warning:
+
+```shell
+The configuration option 'foo' is deprecated. Use 'bar' instead. [ID: foo_option]
+```
+
+You could add the ID `foo_option` under the `disable-warnings` configuration option:
```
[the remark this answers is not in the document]
lucywyman: :banana: :apple:

lib/bolt/logger.rb
```diff
@@ -131,59 +136,104 @@ def self.reset_logging
       Logging.reset
     end
 
-    # Issues a warning unless the user has suppressed it.
+    def self.warn(msg, id)
```
lucywyman, line 139: Haha, so tempting to metaprogram.

lib/bolt/task.rb
```diff
@@ -149,7 +149,7 @@ def validate_metadata
       if unknown_keys.any?
         msg = "Metadata for task '#{@name}' contains unknown keys: #{unknown_keys.join(', ')}."
         msg += " This could be a typo in the task metadata or may result in incorrect behavior."
-        @logger.warn(msg)
+        Bolt::Logger.warn(msg, "unknown_task_metadata_keys")
```
lucywyman, line 152: I think these are still backwards.

lucywyman:
- [x] Disables warnings when set in bolt-project.yaml and bolt-defaults.yaml
- [x] Merges setting from bolt-defauls.yaml and bolt-project.yaml
- [x] Errors if not provided an array
- [x] Accepts an empty array
- [x] Accepts strings that are not valid IDs 
- [x] Queues warnings from multiple classes and prints correctly
- [x] Includes warning ID next to any warnings
- [x] All warnings can be disabled

lucywyman (approved):

hestonhoffman (approved):
LGTM!

Issue #2559, opened 2021-01-15
Remove deprecated command-line options


beechtom:
## Use Case
The following command-line options should be removed:
- `--boltdir`
- `--configfile`
- `--debug`
- `--description`
- `--puppetfile`

PR #2569, opened 2021-01-21
(GH-2559) Remove deprecated command-line options

beechtom:
This removes support for deprecated command-line options:

- `--boltdir`
- `--configfile`
- `--debug`
- `--description`
- `--puppetfile`

This also removes the `Bolt::Config.from_file` factory method, since it
is only used when using the `--configfile` command-line option.

!removal

* **Remove deprecated command-line options**
  ([#2559](#2559))

  The `--boltdir`, `--configfile`, `--debug`, `--description`, and
  `--puppetfile` command-line options have been removed.

### DIFF
--- a/lib/bolt/cli.rb
+++ b/lib/bolt/cli.rb
@@ -159,25 +159,19 @@ module Bolt
     # Loads the project and configuration. All errors that are raised here are not
     # handled by the outputter, as it relies on config being loaded.
     def load_config
-      @config = if ENV['BOLT_PROJECT']
-                  project = Bolt::Project.create_project(ENV['BOLT_PROJECT'], 'environment')
-                  Bolt::Config.from_project(project, options)
-                elsif options[:configfile]
-                  Bolt::Config.from_file(options[:configfile], options)
+      project = if ENV['BOLT_PROJECT']
+                  Bolt::Project.create_project(ENV['BOLT_PROJECT'], 'environment')
+                elsif options[:project]
+                  dir = Pathname.new(options[:project])
+                  if (dir + Bolt::Project::BOLTDIR_NAME).directory?
+                    Bolt::Project.create_project(dir + Bolt::Project::BOLTDIR_NAME)
+                  else
+                    Bolt::Project.create_project(dir)
+                  end
                 else
-                  cli_flag = options[:project] || options[:boltdir]
-                  project = if cli_flag
-                              dir = Pathname.new(cli_flag)
-                              if (dir + Bolt::Project::BOLTDIR_NAME).directory?
-                                Bolt::Project.create_project(dir + Bolt::Project::BOLTDIR_NAME)
-                              else
-                                Bolt::Project.create_project(dir)
-                              end
-                            else
-                              Bolt::Project.find_boltdir(Dir.pwd)
-                            end
-                  Bolt::Config.from_project(project, options)
+                  Bolt::Project.find_boltdir(Dir.pwd)
                 end
+      @config = Bolt::Config.from_project(project, options)
     rescue Bolt::Error => e
       fatal_error(e)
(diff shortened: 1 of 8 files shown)


lucywyman (changes requested):
I think we should also update `resources/bolt_bash_completion.sh`. This also needs to include `--inventoryfile`, and I think should also include `--puppetfile` (which doesn't seem to be in the developer update? It'd be good to add that here too https://puppet.com/docs/bolt/latest/developer_updates.html).

Tests:
- [x] Using any of the commandline options raises an error

beechtom:
@lucywyman We aren't removing support for either `--inventoryfile` or `--puppetfile`.

lucywyman:
Sorry, we aren't for `--inventoryfile`, but we are removing `--puppetfile`:
```
lucy ~/githubs/bolt/project (refactor-transport) $ be bolt module install --puppetfile Puppetfile --force
Unknown option 'disable-warnings' at /home/lucy/githubs/bolt/project/bolt-project.yaml.
Command line option '--puppetfile' is deprecated, and will be removed in Bolt 3.0. You can migrate to using the new module management workflow using 'bolt project migrate'.
```

resources/bolt_bash_completion.sh
```diff
@@ -12,11 +12,11 @@ _bolt() {
...
+	local all_options="--cleanup --color -c --concurrency --connect-timeout --format -h --help --host-key-check -i --inventoryfile --log-level -m --modulepath --no-cleanup --no-color --no-host-key-check --no-ssl --no-ssl-verify --no-tty --noop -p --password --password-prompt --private-key --project -q --query --run-as --ssl --ssl-verify --sudo-executable --sudo-password --sudo-password-prompt --tmpdir --trace --transport --tty -u --user -v --verbose --version"
 
-	local general_opts="-h --help --debug --format --version"
+	local general_opts="-h --help --format --version"
 	local targeting_opts="-t --targets -q --query --rerun --save-rerun --no-save-rerun"
-	local project_config_opts="--project --configfile"
+	local project_config_opts="--project"
```
lucywyman, line 19: File also needs to have --puppetfile reference removed on line 65.

lib/bolt/cli.rb
```diff
@@ -159,25 +159,22 @@ def parse_command
...
     def load_config
-      @config = if ENV['BOLT_PROJECT']
-                  project = Bolt::Project.create_project(ENV['BOLT_PROJECT'], 'environment')
-                  Bolt::Config.from_project(project, options)
-                elsif options[:configfile]
-                  Bolt::Config.from_file(options[:configfile], options)
+      project = if ENV['BOLT_PROJECT']
+                  Bolt::Project.create_project(ENV['BOLT_PROJECT'], 'environment')
                 else
                   cli_flag = options[:project] || options[:boltdir]
```
lucywyman, line 165: `options[:boltdir]` can't be set anymore, this can be inlined to the `if` below.

lucywyman (approved):
Few small comments, looks good!

- [x] Errors when CLI option is used:
  - [x] --configfile
  - [x] --boltdir
  - [x] --description
  - [x] --debug
  - [x] --puppetfile
- [x]  Does not autocomplete
  - [x] --configfile
  - [x] --boltdir
  - [x] --description
  - [x] --debug
  - [x] --puppetfile
- [x] No references in codebase
  - [x] --configfile
  - [x] --boltdir
  - [x] --description
  - [x] --debug - This is still referenced at `spec/fixtures/linuxdev/bolt-kerberos-test.sh`, but I don't think we use that.
  - [x] --puppetfile
- [x] Still respects `--project`
- [x] Does not call `Config.from_file`
- [x] Respects `description` set in a plan

PR #2574, opened 2021-01-22
(maint) Update Puppet modules

lucywyman:
This pulls in updates to the Puppet modules that include allowing them
to be used with Puppet 7.x. Updates are to the maximum Puppet version,
not the minimum, so this is safe to merge to the 2.x series.

!feature

* **Bolt modules usable with Puppet 7.x**

  Modules owned by the Bolt team now have a maximum Puppet version of 8,
  so are usable with Puppet 7 on the Bolt controller.


lucywyman:
Ready once `secure_env_vars` is out.

beechtom (approved):

Issue #2467, opened 2020-12-11
Clarify using interpolations in plan_hierarchy


beechtom:
## Describe the Change You Would Like
Our documentation on `plan_hierarchy` in Hiera suggests that Hiera lookups outside of apply blocks do not support interpolations. However, Hiera lookups outside of apply blocks _do_ support interpolations, as long as the variable being interpolated is defined. We should clarify that outside of apply blocks, target-level data such as facts is not available, so interpolations that would normally work under `hierarchy` won't work under `plan_hierarchy` unless the variable being interpolated is explicitly defined.

## Example

Following is a basic example of interpolation with a plan variable under `plan_hierarchy`:

```yaml
# hiera.yaml
version: 5
defaults:
  datadir: data
  data_hash: yaml_data

plan_hierarchy:
  - name: "OS data"
    path: "%{os}.yaml"
  - name: "Common data"
    path: common.yaml
```

```yaml
# data/ubuntu.yaml
foo: ubuntu.yaml
```

```yaml
# data/common.yaml
foo: common.yaml
```

```puppet
# init.pp
plan bolt(String $os) {
  return lookup('foo')
}
```

```shell
$ bolt plan run bolt os=ubuntu                                                                                                      
Starting: plan bolt
Finished: plan bolt in 0.01 sec
"ubuntu.yaml"
```

```shell
$bolt plan run bolt os=windows
Starting: plan bolt
Finished: plan bolt in 0.01 sec
"common.yaml"
```

## TODO
Add philosophy around why this is undefined behavior (unlike how HIera is used elsewhere, incidental to using HIera, plan calling plan example). to this issue.

adreyer:
I don't think it's a good idea for us to have an example of interpolations since they're probably not a best practice. If we are going to have an example I think it's critical that we chose an interpolation that doesn't look like a fact. Perhaps "application" or something?  I think a caveat instead of an example along the lines of

Interpolations are not well supported in plan hierarchy. Target level data such as facts will not be available so the normal hierachy interpolation will not work. The `lookup` function will interpolate based on variables currently in scope.

PR #2576, opened 2021-01-26
(GH-2467) Clarify using interpolations in plan_hierarchy

lucywyman:
This introduces a new section about using Hiera interpolations with the
`plan_hierarchy` key. It clarifies that plans are not run in
environment- or node-specific contexts, and so only have access to
variables in scope. It also provides an example of using static
variable lookups in a plan.

!no-release-note

### DIFF
--- a/documentation/writing_plans.md
+++ b/documentation/writing_plans.md
@@ -210,14 +210,14 @@ compiles Puppet catalogs in a per-target context, and has unfettered access to H
 use the same Hiera config and data you would for Puppet, and look up data as you would expect
 inside apply blocks. Outside apply blocks, Bolt is essentially executing a script. It doesn't have a
 concept of a target or context, and thus cannot load per-target data. This breaks common Hiera
-features like interpolations.
+features like interpolating target facts.
 
 To look up static data outside of apply blocks, you can add a `plan_hierarchy` key to your Hiera
 configuration. The `plan_hierarchy` key is specified at the same level as the `hierarchy` key and is
 used whenever you look up data outside an apply block. By specifying both keys in the same Hiera
 configuration, you can look up data inside and outside apply blocks in the same plan. This allows
 you to use your existing Hiera configuration in Bolt plans without encountering an error if
-interpolations exist and your plan tries to look up data outside an apply block.
+per-target interpolations exist and your plan tries to look up data outside an apply block.
 
 Using static Hiera data is particularly useful for user-specific data that you want the plan to look
 up using Hiera (for example, to take advantage of Hiera eyaml). For example, with this Hiera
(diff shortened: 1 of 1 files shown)


documentation/writing_plans.md
```diff
@@ -268,6 +268,69 @@ plan plan_lookup(
 }
 ```
 
+### Using interpolations in the `plan_hierarchy` hierarchy
+
+Interpolations are not well supported in the `plan_hierarchy` hierarchy. Puppet plans run in a
+custom, temporary Puppet environment and are not run in the context of a particular target.
```
hestonhoffman, line 274:
```suggestion
Interpolations are not well supported in the `plan_hierarchy` hierarchy. Bolt runs 
Puppet plans in a custom, temporary Puppet environment, and does not run the plans 
in the context of a particular target.
```

documentation/writing_plans.md
```diff
@@ -268,6 +268,69 @@ plan plan_lookup(
...
+Therefore target level data such as facts will not be available to look ups outside of apply blocks,
```
hestonhoffman, line 275:
```suggestion
Target level data such as facts are not available to lookups outside of apply blocks,
```
I think "lookup" is noun here, so one word is fine

documentation/writing_plans.md
```diff
@@ -268,6 +268,69 @@ plan plan_lookup(
...
+so the normal hierachy interpolation will not work. The `lookup()` function will interpolate based on
+variables currently in scope.
```
hestonhoffman, line 277:
```suggestion
so the normal hierarchy interpolation does not work. The `lookup()` function 
only interpolates based on the variables currently in scope.
```

documentation/writing_plans.md
```diff
@@ -268,6 +268,69 @@ plan plan_lookup(
...
+plan plan_lookup(
+  TargetSpec $targets,
+  String $application = 'doggos'
+) {
+  $site_path = lookup('site_path')
+  run_task("deploy_site", $targets, 'path' => $site_path)
+}
+```
+
+Note that if you try to call `lookup('site_path')` from a subplan of `plan_lookup`, like so:
```
hestonhoffman, line 312:
```suggestion
Note that if you tried to call `lookup('site_path')` from a subplan of `plan_lookup`, like so:
```

documentation/writing_plans.md
```diff
@@ -268,6 +268,69 @@ plan plan_lookup(
...
+```
+plan other_plan(
+  TargetSpec $targets
+) {
+  lookup('site_path')
+  ...
+}
+```
+
+This will error, as `$application` is not in scope in `other_plan`.
```
hestonhoffman, line 332:
```suggestion
This would error, because `$application` is not in scope in `other_plan`.
```

hestonhoffman (commented):
🐶 🐈

hestonhoffman:
Oh, one more thing. None of this is contradicting line 212, right? ("This breaks common Hiera features like interpolations.")

lucywyman:
Good call, I updated those lines to be more accurate.

hestonhoffman (approved):

beechtom (approved):

Issue #2558, opened 2021-01-15
Remove puppetfile subcommand


beechtom:
## Use Case
Bolt should remove support for the `puppetfile` subcommand.

beechtom:
Closed by #2579

PR #2579, opened 2021-01-26
(GH-2558) Remove 'bolt puppetfile' subcommand

lucywyman:
This removes the `bolt puppetfile` subcommand entirely, including from
the powershell module and autocomplete script.

!removal

* **Remove bolt puppetfile subcommand**

  Removes the 'bolt puppetfile' and '-BoltPuppetfile' subcommands in
  favor of the module workflow.


lib/bolt/cli.rb
```diff
@@ -725,19 +708,17 @@ def generate_types
     #
     def install_project_modules(project, config, force, resolve)
       assert_project_file(project)
-      assert_puppetfile_or_module_command(project.modules)
 
-      unless project.modules
+      unless project.modules.any?
```
beechtom, line 712: This will error if `modules` is not configured. If we want `modules` to have a default `[]`, then `Bolt::Project.modules` needs to be updated.

lib/bolt/config.rb
```diff
@@ -251,6 +251,7 @@ def initialize(project, config_data, overrides = {})
         'format'              => 'human',
         'log'                 => { 'console' => {} },
         'module-install'      => {},
+        'modules'             => [],
```
beechtom, line 254: This will always be `[]` and isn't accessible outside `Bolt::Config`.

beechtom (changes requested):

lib/bolt/cli.rb
lucywyman: Ok, updated to set the default in the project as well.I think we'll end up removing most references to `project.modules` once we don't need to modify the modulepath, but we'll need it for the migrator.

lib/bolt/config.rb
beechtom, line 254:
```suggestion
```

beechtom (changes requested):

lib/bolt/config.rb
lucywyman: Why? This should be the default.

lucywyman: Oh, because it can't be set in `bolt-defaults.yaml`. Hm...ok.

beechtom (approved):

Issue #2560, opened 2021-01-15
Remove notice log level


beechtom:
## Use Case
Bolt should drop support for the deprecated `notice` log level.

beechtom:
Closed by #2580

PR #2580, opened 2021-01-27
(GH-2560) Remove 'notice' log level

lucywyman:
This removes the 'notice' log level, which means Bolt will error if the
logger is configured to 'notice' either via the command line or
configuration files.

!removal

* **Remove 'notice' log level**

  Bolt no longer accepts 'notice' as a log level, via the command line
  or configuration. Use 'info' instead.

### DIFF
--- a/documentation/running_bolt_in_docker.md
+++ b/documentation/running_bolt_in_docker.md
@@ -120,7 +120,7 @@ This is a basic [project configuration](configuring_bolt.md) file.
 ```yaml
 log:
   console:
-    level: notice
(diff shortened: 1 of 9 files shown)


beechtom (approved):

hestonhoffman:
There are a couple of mentions to notice here: https://github.com/puppetlabs/bolt/blob/3.0/documentation/applying_manifest_blocks.md#puppet-log-functions-in-bolt

hestonhoffman (approved):

Issue #2557, opened 2021-01-15
Remove support for bolt.yaml config file


beechtom:
## Use Case
Bolt should drop support for the deprecated `bolt.yaml` configuration file.

lucywyman:
Boooo github.

PR #2581, opened 2021-01-27
(GH-2557) Remove bolt.yaml support

beechtom:
This removes support for the `bolt.yaml` configuration file.

!removal

* **Remove support for the `bolt.yaml` configuration file**
  ([#2557](#2557))

  The `bolt.yaml` configuration file is no longer supported by Bolt. Use
  `bolt-project.yaml` and `bolt-defaults.yaml` instead.


beechtom:
There's some more cleanup I'd like to do of the config and project classes, since removing `bolt.yaml` support makes things _much_ simpler. I think we can leave that for a later time, though, so it doesn't hold this work up.

lib/bolt/config/options.rb
```diff
@@ -501,26 +501,17 @@ module Options
...
-        apply-settings
-        color
+      # Options that are available on the command line
+      CLI_OPTIONS = %w[
         compile-concurrency
         concurrency
         format
-        hiera-config
         log
-        modulepath
```
lucywyman, line 513: `modulepath` can be specified on the CLI - I know it gets converted to a string no matter what, so is valid, but it seems weird to leave out here.

lucywyman (approved):
One v small nit, this is so exciting! :raised_hands:

Issue #2556, opened 2021-01-15
Stop building and testing Debian 8 packages


beechtom:
## Use Case
We should stop building and testing Debian 8 packages, as the platform has reached EOL.

- [ ] [ci-job-configs PR](https://github.com/puppetlabs/ci-job-configs/pull/7562) This can't be merged until we're ready to release 3.0
- [x] [Bolt PR](https://github.com/puppetlabs/bolt/pull/2583)
- [x] [Bolt vanagon PR](https://github.com/puppetlabs/bolt-vanagon/pull/151)

PR #2583, opened 2021-01-27
(GH-2556) Remove support for Debian 8

lucywyman:
This removes the install documentation for Debian 8, and additionally
removes some unneeded logic for skipping Debian 8 in our acceptance
tests. This is related to puppetlabs/ci-job-configs#7562 which removes
testing and building on Debian 8.

!removal

* **Remove support for Debian 8**

  Bolt will no longer build or test packages for the Debian 8 platform.

### DIFF
--- a/acceptance/tests/apply_plan_ssh.rb
+++ b/acceptance/tests/apply_plan_ssh.rb
@@ -7,13 +7,7 @@ test_name "bolt plan run should apply manifest block on remote hosts via ssh" do
   extend Acceptance::BoltCommandHelper
 
   ssh_nodes = select_hosts(roles: ['ssh'])
-  skip_targets = select_hosts(platform: [/debian-8/])
   targets = "ssh_nodes"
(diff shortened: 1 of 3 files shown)


hestonhoffman (approved):

acceptance/tests/apply_plan_ssh.rb
```diff
@@ -7,7 +7,6 @@
   extend Acceptance::BoltCommandHelper
 
   ssh_nodes = select_hosts(roles: ['ssh'])
-  skip_targets = select_hosts(platform: [/debian-8/])
```
beechtom, line 10: Since this var is no longer defined we should remove the `if skip_targets.any?` bit below.

lucywyman: Damn, eagle-eye Beech!

beechtom (approved):

PR #2584, opened 2021-01-27
(PE-31004) Support input data env var for the PuppetConnectData plugin

ekinanp:


lib/bolt/plugin/puppet_connect_data.rb
```diff
@@ -3,12 +3,32 @@
 module Bolt
   class Plugin
     class PuppetConnectData
+      INPUT_DATA_VAR = 'PUPPET_CONNECT_INPUT_DATA'
+
       def initialize(context:, **_opts)
-        puppet_connect_data_yaml_path = File.join(context.boltdir, 'puppet_connect_data.yaml')
+        data_path = nil
```
nicklewis, line 9: This initialization isn't necessary.

lib/bolt/plugin/puppet_connect_data.rb
```diff
@@ -3,12 +3,32 @@
...
-          puppet_connect_data_yaml_path,
-          'puppet_connect_data.yaml'
+          data_path,
+          File.basename(data_path)
         )
+
+        if @input_data_path
+          # Validate that the data does not contain any plugin-reference
+          # values
+          @data.each do |k, v|
```
nicklewis, line 25: This should check for nested values:

```yaml
mykey:
  another_key:
    _plugin: myplugin
```

I think there are some helpers for that in `Bolt::Util`.

lib/bolt/plugin/puppet_connect_data.rb
```diff
@@ -29,6 +49,13 @@ def validate_resolve_reference(opts)
           raise Bolt::ValidationError,
                 "puppet_connect_data plugin requires that 'key' be specified"
         end
+        if @input_data_path && !@data.key?(opts['key'])
```
nicklewis, line 52: I think we should use a more explicit variable (or have a wrapper method) to indicate that what we've turned on is behavior to fail on unspecified keys.

That would make the initialize method like

```rb
if ENV.key?(INPUT_DATA_VAR)
  @fail_on_missing_key = true
  data_path = ENV[INPUT_DATA_VAR]
else
  @fail_on_missing_key = false
  data_path = File.join(context.boltdir, 'puppet_connect_data.yaml')
end
```

nicklewis (changes requested):

lib/bolt/plugin/puppet_connect_data.rb
ekinanp: I thought about that. The reason I didn't choose it was b/c that variable doesn't explain why we're failing on missing keys, while `input_data_path` does (because when it's input data, you want to fail there). That is also a personal preference so if you think the more explicit variable would help I can do that.

Think I'll still need to store the input data path to properly generate the error message.

nicklewis: I think making `data_path` an instance variable is fine so we store the path in either case. I would be okay leaving this line as-is with a comment explaining. I worry that it will become unclear in the future why the behavior is different depending on whether you have an explicit path set or not.

ekinanp: Updated with a comment.

lib/bolt/plugin/puppet_connect_data.rb
ekinanp: Thanks, updated.

lib/bolt/plugin/puppet_connect_data.rb
ekinanp: Thanks, updated.

lib/bolt/plugin/puppet_connect_data.rb
```diff
@@ -3,12 +3,34 @@
 module Bolt
   class Plugin
     class PuppetConnectData
+      INPUT_DATA_VAR = 'PUPPET_CONNECT_INPUT_DATA'
+
       def initialize(context:, **_opts)
-        puppet_connect_data_yaml_path = File.join(context.boltdir, 'puppet_connect_data.yaml')
+        if ENV.key?(INPUT_DATA_VAR)
```
McdonaldSeanp, line 9: I think we should comment here on what what this var changes about the plugin and why. The way the code interacts with the variables set makes sense to me, but leaving the extra breadcrumb of what/why here should be helpful in the future.

ekinanp: Done.

McdonaldSeanp (approved):

Issue #2555, opened 2021-01-15
Remove deprecated pkcs7 plugin parameters


beechtom:
## Use Case
The following pkcs7 plugin parameters should be removed:
- `private-key`
- `public-key`

beechtom:
Closed by #2585

PR #2585, opened 2021-01-27
(GH-2555) Remove public-key, private-key params for pkcs7 plugin

beechtom:
This removes support for the `public-key` and `private-key` parameters
in the pkcs7 plugin. Previously, Bolt would change these parameters to
the correct `public_key` and `private_key` parameters.

!removal

* **Remove support for `private-key`, `public-key` parameters in pkcs7
  plugin**
  ([#2555](https://github.com/puppetlabs/bolt/issues/2555))

  Support for the `private-key` and `public-key` parameters in the pkcs7
  plugin has been removed. Use the `private_key` and `public_key`
  parameters instead.

### DIFF
--- a/lib/bolt/plugin/module.rb
+++ b/lib/bolt/plugin/module.rb
@@ -30,10 +30,6 @@ module Bolt
         @module = mod
         @config = config
         @context = context
-
-        if @module.name == 'pkcs7'
-          @config = handle_deprecated_pkcs7_keys(@config)
-        end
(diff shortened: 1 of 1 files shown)


lucywyman (approved):

lucywyman:
Successfully does not recognize either key, and I can't find any other references in the codebase (at least for `public-key`, `private-key` is much harder to search for).

Issue #2554, opened 2021-01-15
Remove deprecated YAML plan step keys


beechtom:
## Use Case
The following YAML plan step keys should be removed:
- `target`
- `upload`

lucywyman:
Closed by #2586

PR #2586, opened 2021-01-27
(GH-2554) Remove source, target YAML plan step keys

beechtom:
This removes support for the `source` and `target` keys in YAML plan
steps. Previously, the `source` key could be used as an action instead
of `upload`, while `target` could be used instead of `targets`.

!removal

* **Remove `source` and `target` YAML plan step keys**
  ([#2554](https://github.com/puppetlabs/bolt/issues/2554))

  Support for the `source` and `target` keys in YAML plans has been
  removed. Use `upload` and `targets` instead.

### DIFF
--- a/lib/bolt/pal/yaml_plan/step.rb
+++ b/lib/bolt/pal/yaml_plan/step.rb
@@ -9,19 +9,17 @@ module Bolt
         attr_reader :name, :type, :body, :targets
 
         def self.allowed_keys
-          Set['name', 'description', 'target', 'targets']
+          Set['name', 'description', 'targets']
         end
 
         STEP_KEYS = %w[
           command
-          destination
           download
           eval
           message
           plan
           resources
           script
-          source
           task
           upload
         ].freeze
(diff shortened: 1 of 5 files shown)


lib/bolt/pal/yaml_plan/step.rb
```diff
@@ -96,19 +88,6 @@ def self.validate_step_keys(body, step_number)
...
 
-          # Handle cases where steps with a required 'targets' key are using the deprecated
-          # 'target' key instead.
-          # TODO: Remove this when 'target' is removed
-          if body.include?('target')
-            missing_keys -= ['targets']
-          end
-
-          # Handle cases where upload step uses deprecated 'source' key instead of 'upload'
-          # TODO: Remove when 'source' is removed
```
lucywyman, line 107: I can feel, through the screen, how excited we were to remove these.

lucywyman:
:+1: `target` and `source` are not recognized for any steps that formerly accepted them, and can't find any other references in the codebase.

Issue #2549, opened 2021-01-15
Update Bolt to only use new default modulepath


beechtom:
## Use Case
Update Bolt to only use the new default modulepath `['modules']`.

beechtom:
Closed by #2587

PR #2587, opened 2021-01-27
(GH-2549) Use new default modulepath

beechtom:
This updates Bolt to use the new default modulepath `['modules']` and
append the `.modules/` directory to the end of both the default
modulepath and configured modulepath.

!feature

* **Update default modulepath**
  ([#2549](https://github.com/puppetlabs/bolt/issues/2549))

  Bolt's default modulepath is now `['modules']` instead of
  `['modules', 'site', 'site-modules']`. Bolt will also automatically
  append the project's `.modules/` directory to all modulepaths,
  whether a project uses the default modulepath or a configured
  modulepath.

### DIFF
--- a/lib/bolt/project_manager.rb
+++ b/lib/bolt/project_manager.rb
@@ -195,7 +195,7 @@ module Bolt
 
       migrator.migrate(
         @config.project,
-        @config.modulepath
+        @config.modulepath[0...-1]
       )
     end
   end
--- a/lib/bolt/project_manager/module_migrator.rb
+++ b/lib/bolt/project_manager/module_migrator.rb
@@ -6,19 +6,20 @@ module Bolt
   class ProjectManager
     class ModuleMigrator < Migrator
       def migrate(project, configured_modulepath)
-        return true unless project.modules.empty?
+        return true if project.managed_moduledir.exist?
 
         @outputter.print_message "Migrating project modules\n\n"
 
         config            = project.project_file
         puppetfile        = project.puppetfile
         managed_moduledir = project.managed_moduledir
-        modulepath        = [(project.path + 'modules').to_s,
+        new_modulepath    = [(project.path + 'modules').to_s]
+        old_modulepath    = [(project.path + 'modules').to_s,
                              (project.path + 'site-modules').to_s,
                              (project.path + 'site').to_s]
 
         # Notify user to manually migrate modules if using non-default modulepath
-        if configured_modulepath != modulepath
+        if configured_modulepath != new_modulepath && configured_modulepath != old_modulepath
           @outputter.print_action_step(
             "Project has a non-default configured modulepath, unable to automatically "\
             "migrate project modules. To migrate project modules manually, see "\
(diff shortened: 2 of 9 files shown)


hestonhoffman:
QQ: Does this update the generated reference docs (like bolt_project_reference) with the new defaults? When I checked this branch out and ran the rake task, I'm still getting 
```
- **Type:** Array, String
- **Default:** ["project/modules", "project/site-modules", "project/site"]
```

But I might be doing it wrong 😊

beechtom:
Ah, nope, forgot to update the default. Good catch! 👍

lib/bolt/project_manager.rb
```diff
@@ -195,7 +195,7 @@ def migrate
 
       migrator.migrate(
         @config.project,
-        @config.modulepath
+        @config.modulepath[0...-1]
```
lucywyman, line 198:
```suggestion
        @config.modulepath[0..-1]
```
I'm not even sure what `...` does, but this results in `modulepath` being removed from `bolt-project.yaml` entirely, but *only* if it was either `[modules]` or `[modules, site-modules, site]`.

beechtom: It makes sure we drop `.modules/` from the modulepath, since `Bolt::Config.modulepath` automatically appends it to the default modulepath _and_ the configured modulepath.

The current behavior for the migrator is to remove `modulepath` from `bolt-project.yaml`, so that's expected behavior...

lib/bolt/project_manager/module_migrator.rb
```diff
@@ -27,14 +28,21 @@ def migrate(project, configured_modulepath)
...
-          consolidate_modules(modulepath)
+          consolidate_modules(old_modulepath)
           update_project_config([], config)
         end
       end
 
+      # Returns true if the project's modules can be migrated, false otherwise.
+      # To migrate a project's modules, the current project must use either
+      # the old or new default modulepath.
+      def migrateable?
```
lucywyman, line 42: I don't see where we use this?

beechtom: Hah. Just a direction I abandoned and forgot to remove. 😅

hestonhoffman (approved):
Docs LGTM!

lucywyman:
- [x] Updates the default `modulepath` to `['modules']` if `modulepath` isn't configured
- [x] Appends `.modules/` to the modulepath
  - [x] When `modulepath` is configured
  - [x] When `modulepath` isn't configured
- [x] Uses the default modulepath even if `modulepath` is "set" in `bolt-defaults.yaml`
- [x] `project migrate` moves modules from the previous default modulepath `['modules', 'site-modules', 'site']` to the `modules` directory **Note: this does mean that if modules have the same name, one will get overridden**
  - [x] If `modulepath` isn't configured
  - [x] If `modulepath` is configured to `['modules']`
  - [x] If `modulepath` is configured to `'modules'`
  - [x] Does not move modules if `modulepath` is configured to anything other than `['modules']`
  - [x] If `modulepath` is `['modules', 'site-modules', 'site']`
- [x] Still appends the `project` to the modulepath if named
- [x] Is still overridden by `--modulepath`

Issue #2552, opened 2021-01-15
Apply localhost magic to all targets using the local transport


beechtom:
## Use Case
Default configuration for the `local` transport should be updated to match the same defaults used for the `localhost` target.

beechtom:
Closed by #2589

PR #2589, opened 2021-01-28
(GH-2552) bundled-ruby transport config now defaults to true

lucywyman:
The local transport's `bundled-ruby` configuration option, which
determines whether to use the Ruby bundled with Bolt packages for local
targets, now defaults to 'true' instead of 'false'. The option can still
be configured as before.

!feature

* **Local transport's 'bundled-ruby' option defaults to true**
  The local transport's `bundled-ruby` configuration option, which
  determines whether to use the Ruby bundled with Bolt packages for local
  targets, now defaults to 'true' instead of 'false'. The option can still
  be configured as before.

### DIFF
--- a/lib/bolt/config/transport/local.rb
+++ b/lib/bolt/config/transport/local.rb
@@ -17,6 +17,7 @@ module Bolt
         OPTIONS = WINDOWS_OPTIONS.dup.concat(RUN_AS_OPTIONS).sort.freeze
 
         DEFAULTS = {
+          'bundled-ruby' => true,
           'cleanup' => true
         }.freeze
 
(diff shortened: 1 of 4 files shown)


lucywyman:
Relies on #2597

beechtom (changes requested):
I think https://github.com/puppetlabs/bolt/blob/main/lib/bolt/transport/local.rb#L18-L24 needs to be removed.

lucywyman:
This does have one weird edge case I found, which is that previously we merged in the `localhost` defaults for any target named localhost. In order for `bundled-ruby` to also be able to configure `localhost` (which is still the only target that automatically uses the `local` transport), we need to, well, only apply the defaults if `bundled-ruby` is true. That means that previously `ssh://localhost` would have picked up the `puppet-agent` feature, and it won't anymore. I doubt anyone will run in to that, but it's just good to know and record.

beechtom (approved):

Issue #2547, opened 2021-01-15
Ship with Puppet 7


beechtom:
## Use Case
Ship Bolt packages with Puppet 7

- [x] Update puppet-runtime rubygem https://github.com/puppetlabs/puppet-runtime/pull/417

lucywyman:
This needs to go into `puppet-runtime#master`, and would make us unshippable. Blocking until we're ready to ship 3.0.

PR #2600, opened 2021-01-29
Revert "(GH-2547) Use Puppet 7"

lucywyman:
Reverts puppetlabs/bolt#2598


lucywyman:
Gah sorry - we should only need to update the puppet-runtime to pull in the Puppet 7 gem, but the current rule here should still allow for Puppet 7 to be used and we don't want to be unnecessarily restrictive.

PR #2603, opened 2021-02-01
(GH-102) Prototype streaming remote output

lucywyman:
This depends on #2563. Targetting 3.0 for now, but I expect this will get retargeted to 'main' after releasing 3.0.


lucywyman:
- [x] It is disabled by default, and can be enabled
  - Enabled as part of the transport so we have access on the target (not ideal, but...people will live)
  - New log level 'stream', which is only enabled for logging stream messages.
  - Cannot be sent to file because the validator won't allow it to be
    configured to be sent to a file
- [x] It streams both stdout and stderr
- [x] It does not print to debug or trace logs
- [x] It does not print other non-stream log messages
- [x] It streams elegantly with multiple nodes
  - I only tested on ~4, this would definitely need more testing
  - Though it's also opt-in, so...
- [x] It streams over local
- [x] It streams over Docker
  - This relies on the Docker transport refactor
- [x] Can print strings more than 4096
  - This ends up printing them across multiple outputs, but that kind of seems
    fine? This should definitely not ever be consumed by machines.
- [x] Works elegantly with spinner
  - The spinner does not seems to "stick" to the output at all (kind of
    unsurprising, I don't think this is a big risk as we log other data from
    the Shell class)
- 

Later work:
- [ ] Add a CLI flag

Test:
- [ ] Works using the PowerShell shell
- [ ] It isn't unbearably slow
- [ ] It streams over SSH
- [ ] It streams both before and after a sudo prompt
- [ ] Does it do weird things with newlines with strings above 4096?
- [ ] Make sure stream can't be set as log level

PR #2608, opened 2021-02-02
 (GH-102) Prototype streaming remote output

lucywyman:
This adds a new configuration option `stream` that will enable
streaming output back to the console while Bolt is executing actions on
remote targets.

Streaming works by registering a new Ruby logger named 'stream' that is
not additive (so does not send log messages to the parent logger, which
prevents messages from being logged to files or to other Bolt logging
streams). It configures the stream logger to only print to the console
without much formatting. The stream logger is only created if `stream`
is configured. Bolt then prints output to the stream logger as it's
read, in addition to sending messages to the Result object to be
returned and printed when the action is finished.

Streaming messages are prepended with the target's safe name and the 
stream (stdout or stderr) the message came from. Streaming messages are 
generally printed twice, once when streaming and again as part of the 
result message.

This feature is experimental.

!feature

* **Add 'stream' option** ([#102](https://github.com/puppetlabs/bolt/issues/102))

  **This option is experimental**. New `stream` configuration option and
  `--stream` CLI flag enable streaming output from Bolt actions as they are running.

### DIFF
--- a/documentation/experimental_features.md
+++ b/documentation/experimental_features.md
@@ -8,6 +8,28 @@ API may change, requiring the user to update their code or configuration. The
 Bolt team attempts to make these changes painless by providing useful warnings
 around breaking behavior where possible. 
 
+## Streaming output
+
+This feature was introduced in [Bolt 3.2.0](https://github.com/puppetlabs/bolt/blob/main/CHANGELOG.md#bolt-310-2021-3-08).
+
+You can set the new `stream` output option in `bolt-project.yaml` or `bolt-defaults.yaml`, or
+specify the option on the command line as `--stream`. Bolt streams results back to the console as
+they are received, with the target's safe name (the URI without the password included) and the
+stream (either 'out' or 'err') appended to the message, like so:
+```
+Started on docker://puppet_6_node...
+Started on docker://puppet_7_node...
+[docker://puppet_7_node] out: Hello!
+[docker://puppet_6_node] out: Hello!
+Finished on docker://puppet_7_node:
+  Hello!
+Finished on docker://puppet_6_node:
+  Hello!
+```
+
+As you can see, when you configure output to stream, Bolt might print to the console twice:
+once as the actions are running, and again after Bolt prints the results.
+
 ## LXD Transport
 
 This feature was introduced in [Bolt 3.2.0](https://github.com/puppetlabs/bolt/blob/main/CHANGELOG.md#bolt-310-2021-3-08).
--- a/lib/bolt/cli.rb
+++ b/lib/bolt/cli.rb
@@ -175,6 +175,7 @@ module Bolt
     # Completes the setup process by configuring Bolt and log messages
     def finalize_setup
       Bolt::Logger.configure(config.log, config.color, config.disable_warnings)
+      Bolt::Logger.stream = config.stream
       Bolt::Logger.analytics = analytics
       Bolt::Logger.flush_queue
 
--- a/lib/bolt/logger.rb
+++ b/lib/bolt/logger.rb
@@ -91,6 +91,14 @@ module Bolt
       Logging.logger[:root].appenders.any?
     end
 
+    def self.stream
+      @stream
+    end
+
+    def self.stream=(stream)
+      @stream = stream
+    end
+
     # A helper to ensure the Logging library is always initialized with our
     # custom log levels before retrieving a Logger instance.
     def self.logger(name)
--- a/lib/bolt/shell/bash.rb
+++ b/lib/bolt/shell/bash.rb
@@ -12,7 +12,6 @@ module Bolt
         super
 
         @run_as = nil
-
         @sudo_id = SecureRandom.uuid
         @sudo_password = @target.options['sudo-password'] || @target.password
       end
@@ -390,14 +389,23 @@ module Bolt
           # See if we can read from out or err, or write to in
           ready_read, ready_write, = select(read_streams.keys, write_stream, nil, timeout)
 
-          # Read from out and err
           ready_read&.each do |stream|
+            stream_name = stream == out ? 'out' : 'err'
             # Check for sudo prompt
-            read_streams[stream] << if use_sudo
-                                      check_sudo(stream, inp, options[:stdin])
-                                    else
-                                      stream.readpartial(CHUNK_SIZE)
-                                    end
+            to_print = if use_sudo
+                         check_sudo(stream, inp, options[:stdin])
+                       else
+                         stream.readpartial(CHUNK_SIZE)
+                       end
+
+            if !to_print.chomp.empty? && @stream_logger
+              formatted = to_print.lines.map do |msg|
+                "[#{@target.safe_name}] #{stream_name}: #{msg.chomp}"
+              end.join("\n")
+              @stream_logger.warn(formatted)
+            end
+
+            read_streams[stream] << to_print
           rescue EOFError
           end
 
--- a/lib/bolt/bolt_option_parser.rb
+++ b/lib/bolt/bolt_option_parser.rb
@@ -13,7 +13,7 @@ module Bolt
                 run_context: %w[concurrency inventoryfile save-rerun cleanup],
                 global_config_setters: PROJECT_PATHS + %w[modulepath],
                 transports: %w[transport connect-timeout tty native-ssh ssh-command copy-command],
-                display: %w[format color verbose trace],
+                display: %w[format color verbose trace stream],
                 global: %w[help version log-level clear-cache] }.freeze
 
     ACTION_OPTS = OPTIONS.values.flatten.freeze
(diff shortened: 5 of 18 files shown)


lib/bolt/config/transport/local.rb
```diff
@@ -14,7 +14,7 @@ class Local < Base
           tmpdir
         ].freeze
 
-        OPTIONS = WINDOWS_OPTIONS.dup.concat(RUN_AS_OPTIONS).sort.freeze
+        OPTIONS = WINDOWS_OPTIONS.dup.concat(RUN_AS_OPTIONS).concat(%w[stream]).sort.freeze
```
beechtom, line 17: Is there a reason to not have this available to the `local` transport on Windows controllers?

lib/bolt/shell/bash.rb
```diff
@@ -390,14 +389,20 @@ def execute(command, sudoable: false, **options)
...
-                                      stream.readpartial(CHUNK_SIZE)
-                                    end
+            to_print = if use_sudo
+                         check_sudo(stream, inp, options[:stdin])
+                       else
+                         stream.readpartial(CHUNK_SIZE)
+                       end
+
+            if !to_print.chomp.empty? && @stream_logger
+              @stream_logger.warn("[#{@target.safe_name}] #{stream_name}: #{to_print.chomp}")
```
beechtom, line 402: When the output has newlines the target & stream names get lost pretty easily. I wonder if adding these to the beginning of each line, or indenting each line would help with readability and quickly figuring out where output is coming from?

documentation/experimental_features.md
```diff
@@ -8,6 +8,28 @@ API may change, requiring the user to update their code or configuration. The
 Bolt team attempts to make these changes painless by providing useful warnings
 around breaking behavior where possible. 
 
+## Streaming output
+
+This feature was introduced in [Bolt 3.2.0](https://github.com/puppetlabs/bolt/blob/main/CHANGELOG.md#bolt-310-2021-3-08).
+
+The new `stream` output option can be set in [transport
+configuration](bolt_transports_reference.md). Bolt will stream results back to the console as they
```
beechtom, line 16: It would be nice to have this is a CLI option as well, so it can be toggled easily. Should be as simple as adding a switch to the option parser since the logic for translating it to transport config is already there.

beechtom (commented):
Still playing around with it, but overall this is awesome! :D

lib/bolt/config/transport/local.rb
lucywyman: Oh no I just forgot!

documentation/experimental_features.md
lucywyman: Yeah - I also realized that now that it's added to all the transports (it was originally just Bash shell) it can also be a Bolt config option, instead of transport config. Refactored to suit.

lib/bolt/shell/bash.rb
lucywyman: Oo definitely. Great idea.

documentation/experimental_features.md
beechtom: Ahh, that feels _much_ better than having it be a transport option. :) Good idea.

documentation/experimental_features.md
```diff
@@ -8,6 +8,28 @@ API may change, requiring the user to update their code or configuration. The
...
 around breaking behavior where possible. 
 
+## Streaming output
+
+This feature was introduced in [Bolt 3.2.0](https://github.com/puppetlabs/bolt/blob/main/CHANGELOG.md#bolt-310-2021-3-08).
+
+The new `stream` output option can be set in either `bolt-project.yaml` or `bolt-defaults.yaml`, or
+specified on the command line as `--stream`. Bolt will stream results back to the console as they
+are received, with the target's safe name (the URI without the password included) and the stream
+(either 'out' or 'err') appended to the message, like so:
```
hestonhoffman, line 18:
```suggestion
You can set the new `stream` output option in `bolt-project.yaml` or `bolt-defaults.yaml`, or
specify the option on the command line as `--stream`. Bolt streams results back to the console as they
are received, with the target's safe name (the URI without the password included) and the stream
(either 'out' or 'err') appended to the message, like so:
```

documentation/experimental_features.md
```diff
@@ -8,6 +8,28 @@ API may change, requiring the user to update their code or configuration. The
...
+[docker://puppet_7_node] out: Hello!
+[docker://puppet_6_node] out: Hello!
+Finished on docker://puppet_7_node:
+  Hello!
+Finished on docker://puppet_6_node:
+  Hello!
+```
+
+As you can see above, when streaming is configured output may be printed to the console twice: as
+the actions are running, and again when results are printed.
```
hestonhoffman, line 31:
```suggestion
As you can see, when you configure output to stream, Bolt might print to the console twice:
once as the actions are running, and again after Bolt prints the results.
```

hestonhoffman (approved):
Some small changes, but LGTM!

lib/bolt/shell/bash.rb
```diff
@@ -390,14 +389,25 @@ def execute(command, sudoable: false, **options)
...
+                         stream.readpartial(CHUNK_SIZE)
+                       end
+
+            if !to_print.chomp.empty? && @stream_logger
+              formatted = ""
+              to_print.split("\\n").each_with_index do |msg, i|
+                indent = i == 0 ? '' : '  '
+                formatted += "[#{@target.safe_name}] #{stream_name}: #{indent}#{msg.chomp}\n"
+              end
+              @stream_logger.warn(formatted)
```
beechtom, line 407:
```suggestion
              to_print.lines.each_with_index do |msg, i|
                indent = i == 0 ? '' : '  '
                formatted += "[#{@target.safe_name}] #{stream_name}: #{indent}#{msg}"
              end
              @stream_logger.warn(formatted.chomp)
```

I think this will fix avoid any extra newlines. Before change:

```shell
$ be bolt command run 'for v in 1 2 3 4 5; do sleep 1; echo "$v"; done' -t nix --stream
Started on nix...
The 'stream' option is experimental, and might include breaking changes between minor versions. [ID: stream_experimental]
[nix] out: 1

[nix] out: 2

[nix] out: 3

[nix] out: 4

[nix] out: 5

Finished on nix:
  STDOUT:
    1
    2
    3
    4
    5
```

With change:

```shell
$ be bolt command run 'for v in 1 2 3 4 5; do sleep 1; echo "$v"; done' -t nix --stream
Started on nix...
The 'stream' option is experimental, and might include breaking changes between minor versions. [ID: stream_experimental]
[nix] out: 1
[nix] out: 2
[nix] out: 3
[nix] out: 4
[nix] out: 5
Finished on nix:
  STDOUT:
    1
    2
    3
    4
    5
```

lib/bolt/shell/bash.rb
```diff
@@ -390,14 +389,25 @@ def execute(command, sudoable: false, **options)
...
+                         check_sudo(stream, inp, options[:stdin])
+                       else
+                         stream.readpartial(CHUNK_SIZE)
+                       end
+
+            if !to_print.chomp.empty? && @stream_logger
+              formatted = ""
+              to_print.split("\\n").each_with_index do |msg, i|
+                indent = i == 0 ? '' : '  '
+                formatted += "[#{@target.safe_name}] #{stream_name}: #{indent}#{msg.chomp}\n"
```
beechtom, line 405: Not super sure about the indent here, mostly because it's not really noticeable with just two spaces. Maybe just the source being added to each line is enough?

lib/bolt/cli.rb
```diff
@@ -175,6 +175,7 @@ def load_config
     # Completes the setup process by configuring Bolt and log messages
     def finalize_setup
       Bolt::Logger.configure(config.log, config.color, config.disable_warnings)
+      Bolt::Logger.stream = options[:stream] || config.stream
```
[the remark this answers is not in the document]
lucywyman: Yeah, I think that `configure` is more meant to configure the Logger itself, whereas this variable is only attached to the logger so that's it's available basically globally (and because it's vaguely logger-related). I actually originally had it as a var to `configure` then pulled it out.

lib/bolt/logger.rb
```diff
@@ -91,6 +91,14 @@ def self.configured?
       Logging.logger[:root].appenders.any?
     end
 
+    def self.stream
+      @stream
+    end
+
+    def self.stream=(stream)
+      @stream = stream
+    end
```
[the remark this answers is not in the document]
lucywyman: Yeah - I think I prefer this because having instance variables on modules already feels weird, and explicitly defining a getter and setter...makes it feel better? It to me reads as more clear that we know we're doing something weird but we're doing it anyway.

lib/bolt/shell/bash.rb
lucywyman: Ah nice - I always forget about `lines`

lib/bolt/shell/bash.rb
lucywyman: Yeah - the spec example is especially weird since it's JSON. I honestly don't mind it but I can definitely see how it's confusing rather than clarifying. We can always reformat if The People ask for it.

beechtom (approved):
🙌

Issue #2565, opened 2021-01-19
Add yard docs for built in plan content


donoghuc:
## Describe the Change You Would Like

The built in `canary` plan has nice docs https://github.com/puppetlabs/bolt/blob/61a691cbac0e0f87def2e9316f9f776808d40f8c/modules/canary/plans/init.pp#L1-L27 (this is very helpful in GUIs that show all built in content)

Update the built in `aggregate` and `puppetdb_fact` plan description and parameter descriptions.

PR #2609, opened 2021-02-02
(GH-2565) Add plan metadata to `aggregate` and `puppetdb_fact` plans

beechtom:
This adds plan metadata to the plans in the `aggregate` and
`puppetdb_fact` modules, which will show in `bolt plan show <plan>` and
`Get-BoltPlan -Name <plan>` output.

This also removes the `aggregate::nodes` plan, which has been superseded
by `aggregate::targets`.

!removal

* **Remove `aggregate::nodes` plan**

  Bolt no longer ships with the `aggregate::nodes` plan. Use the
  `aggregate::targets` plan instead.


lucywyman (approved):
LGTM once CI is green!

PR #2611, opened 2021-02-03
(maint) Required ruby version 2.5

lucywyman:
We accidentally bumped the minimum ruby version in the bolt.gemspec from
ruby 2.7 to ruby 2.5. While Bolt will ship with Ruby 2.7 packages, Bolt
server will continue to ship with 2.5 and we will rely on PE testing to
catch any issues that arise from that.

!no-release-note

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -13,7 +13,7 @@ AllCops:
     - site-modules/**/*
     - tasks/**/*
   SuggestExtensions: false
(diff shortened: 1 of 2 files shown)


beechtom (approved):

beechtom (changes requested):
Looks like rubocop needs to use 2.5 too.

PR #2612, opened 2021-02-03
(PE-31005) Add puppet_connect::test_input_data plan

ekinanp:
This plan's used to test that the Puppet Connect input data is complete,
meaning that all consuming inventory targets are connectable.

!no-release-note

### DIFF
--- a/modules/puppet_connect/README.md
+++ b/modules/puppet_connect/README.md
@@ -1,0 +1,26 @@
+# puppet_connect
+
+#### Table of Contents
+
+1. [Description](#description)
+2. [Requirements](#requirements)
+3. [Usage](#usage)
+4. [Reference](#reference)
+
+## Description
+
+This module provides the puppet_connect::test_input_data plan, which is used to test that the provided Puppet Connect input data is complete. The module should be used exactly as specified in the Puppet Connect docs since it is designed for a highly specialized workflow.
+
+## Requirements
+
+This module is compatible with the version of Puppet Bolt it ships with.
+
+## Usage
+
+See the Puppet Connect docs for the module's usage.
+
+### Parameters
+
+* **targets** - The set of targets to test. Usually this should be 'all', the default.
+
+## Reference
--- a/modules/puppet_connect/plans/test_input_data.pp
+++ b/modules/puppet_connect/plans/test_input_data.pp
@@ -1,0 +1,31 @@
+# @summary
+#   Tests that the provided Puppet Connect input data is complete, meaning that all consuming inventory targets are connectable.
+#
+# This plan should only be used as part of the copy-pastable "test input data"
+# workflow specified in the Puppet Connect docs.
+#
+# @param targets
+#   The set of targets to test. Usually this should be 'all', the default.
+#
+# @return ResultSet the result of invoking the 'is connectable?' query on all
+# the targets. Note that this query currently consists of running the 'echo'
+# command.
+#
+plan puppet_connect::test_input_data(TargetSpec $targets = 'all') {
+  $targs = get_targets($targets)
+  $targs.each |$target| {
+    if $target.transport != 'ssh' and $target.transport != 'winrm' {
+      fail_plan("Inventory contains target ${target} with unsupported transport, must be ssh or winrm")
+    }
+    if $target.transport == 'ssh' {
+      # Disable SSH autoloading to prevent false positive results
+      # (input data is wrong but target is still connectable due
+      # to autoloaded config)
+      set_config($target, ['ssh', 'load-config'], false)
+    }
+  }
+  # The SSH/WinRM transports will report an 'unknown host' error for targets where
+  # 'host' is unknown so run_command's implementation will take care of raising that
+  # error for us.
+  return run_command('echo Connected', $targs)
+}
--- a/modules/puppet_connect/spec/plans/test_input_data.rb
+++ b/modules/puppet_connect/spec/plans/test_input_data.rb
@@ -1,0 +1,82 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+require 'bolt_spec/plans'
+require 'bolt/target'
+
+describe 'puppet_connect::test_input_data' do
+  include BoltSpec::Plans
+
+  let(:inventory_data) do
+    {
+      'version' => 2,
+      'targets' => [
+        {
+          'name' => 'ssh_target',
+          'uri'  => 'ssh_uri',
+          'config' => {
+            'transport' => 'ssh',
+            'ssh'  => {
+              'load-config' => true
+            }
+          }
+        },
+        {
+          'name'   => 'winrm_target',
+          'uri'    => 'winrm_uri',
+          'config' => {
+            'transport' => 'winrm'
+          }
+        }
+      ]
+    }
+  end
+  let(:ssh_target) { inventory.get_target('ssh_target') }
+  let(:winrm_target) { inventory.get_target('winrm_target') }
+
+  context 'when the inventory contains an unsupported Puppet Connect transport' do
+    let(:inventory_data) do
+      sup = super()
+      sup['targets'].first['config']['transport'] = 'docker'
+      sup
+    end
+
+    it 'returns an error result' do
+      result = run_plan('puppet_connect::test_input_data', {})
+      expect(result.ok?).to be(false)
+      expect(result.value.msg).to match(%r{ssh_target.*ssh.*winrm})
+    end
+  end
+
+  it 'sets load-config to false for ssh targets' do
+    allow_command('echo Connected')
+      .always_return({})
+
+    winrm_config_before = winrm_target.config
+    run_plan('puppet_connect::test_input_data', {})
+
+    expect(ssh_target.config).to include('ssh' => { 'load-config' => false })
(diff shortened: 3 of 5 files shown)


modules/puppet_connect/README.md
```diff
@@ -0,0 +1,26 @@
...
+#### Table of Contents
+
+1. [Description](#description)
+2. [Requirements](#requirements)
+3. [Usage](#usage)
+4. [Reference](#reference)
+
+## Description
+
+This module provides the puppet_connect::test_input_data plan, which is used to test that the provided Puppet Connect input data is complete. The module should be used exactly as specified in the Puppet Connect docs since it is designed for a highly specialized workflow.
```
ekinanp, line 12: I don't think the README needs to be more detailed than this since all its possible viewers (user and maintainer) will likely be viewing the Puppet Connect docs when interacting with this module.

ekinanp:
Looks like CI failure's a known issue -- cache_dir transient in bolt-server.

modules/puppet_connect/plans/test_input_data.pp
```diff
@@ -0,0 +1,31 @@
...
+#   The inventory targets. This should never be set by the user. It is only here for compatibility with Bolt's rerun workflow.
+#
+# @return ResultSet the result of invoking the 'is connectable?' query on all
+# the targets. Note that this query currently consists of running the 'whoami'
+# command.
+#
+plan puppet_connect::test_input_data(TargetSpec $targets = 'all') {
+  $targs = get_targets($targets)
+  $targs.each |$target| {
+    if $target.transport != 'ssh' and $target.transport != 'winrm' {
```
ekinanp, line 17: I would like this validation to be included in v1 because without it, the additional UX of "inventory sync returns invalid inventory error" => "fix inventory error" => "commit and push to remote" => "deploy" => "sync inventory" is annoying.

modules/puppet_connect/plans/test_input_data.pp
```diff
@@ -0,0 +1,31 @@
+# @summary
```
ekinanp, line 1:
```
bolt plan show puppet_connect::test_input_data

puppet_connect::test_input_data - Tests that the provided Puppet Connect input data is complete, meaning that all consuming inventory targets are connectable.

USAGE:
bolt plan run puppet_connect::test_input_data [targets=<value>]

PARAMETERS:
- targets: TargetSpec
    Default: 'all'
    The inventory targets. This should never be set by the user. It is only here for compatibility with Bolt's rerun workflow.

MODULE:
built-in module
```

modules/puppet_connect/plans/test_input_data.pp
```diff
@@ -0,0 +1,31 @@
...
+#   Tests that the provided Puppet Connect input data is complete, meaning that all consuming inventory targets are connectable.
+#
+# This plan should only be used as part of the copy-pastable "test input data"
+# workflow specified in the Puppet Connect docs.
+#
+# @param targets
+#   The inventory targets. This should never be set by the user. It is only here for compatibility with Bolt's rerun workflow.
```
nicklewis, line 8: I don't know that we need to be quite so strong about this. Nothing bad will happen if you set it, it's just probably not what you want. Maybe "The set of targets to test. Usually this should be 'all', which is the default."

modules/puppet_connect/plans/test_input_data.pp
```diff
@@ -0,0 +1,31 @@
...
+# the targets. Note that this query currently consists of running the 'whoami'
+# command.
+#
+plan puppet_connect::test_input_data(TargetSpec $targets = 'all') {
+  $targs = get_targets($targets)
+  $targs.each |$target| {
+    if $target.transport != 'ssh' and $target.transport != 'winrm' {
+      fail_plan("Inventory contains target ${target} with unsupported transport, must be ssh or winrm")
+    }
+    if $target.host == undef {
```
nicklewis, line 20: I _think_ this check should be unnecessary since the SSH/WinRM transports will report that the target doesn't have a host. Having a `fail_plan` here means you will only see the error for the _first_ target that's missing a host.

modules/puppet_connect/plans/test_input_data.pp
```diff
@@ -0,0 +1,31 @@
...
+      fail_plan("Inventory contains invalid target ${target}, the target's 'host' key is unknown")
+    }
+    if $target.transport == 'ssh' {
+      # Disable SSH autoloading to prevent false positive results
+      # (input data is wrong but target is still connectable due
+      # to autoloaded config)
+      set_config($target, ['ssh', 'load-config'], false)
```
nicklewis, line 27: We can set this unconditionally. If the target isn't using ssh, it will just ignore the ssh portion of the config.

modules/puppet_connect/plans/test_input_data.pp
ekinanp: Hmm, but it is why the parameter was added. I don't have strong feelings about making this more flexible, but I see no good reason to do that when this plan is meant for a very specialized workflow.

modules/puppet_connect/plans/test_input_data.pp
ekinanp: Actually on second thought, I'd rather leave this as-is because it is more in-line with how you'd explain it to someone ("we set load-config to false for SSH targets" => "if target is ssh, set load-config to false"). Knowing that it will still do the right thing for WinRM targets is a nice implementation detail, but an implementation detail that doesn't result in significant cleanup.

modules/puppet_connect/plans/test_input_data.pp
ekinanp: Yup, you are correct. Updated.

modules/puppet_connect/plans/test_input_data.pp
nicklewis: It's more an issue of perception than making it more flexible. Having something that is _exposed_ to the user but is not _intended_ to be exposed to the user feels sloppy and I think it will make this feature feel less finished. But if we reframe it as this is a feature that is available but you probably don't need to use it, it doesn't feel like there's a hole.

ekinanp: Ok, that reasoning makes sense. Updated.

modules/puppet_connect/plans/test_input_data.pp
```diff
@@ -0,0 +1,31 @@
...
+    }
+  }
+  # The SSH/WinRM transports will report an 'unknown host' error for targets where
+  # 'host' is unknown so run_command's implementation will take care of raising that
+  # error for us.
+  return run_command('whoami', $targs)
```
donoghuc, line 30: nit, but in orch we do https://github.com/puppetlabs/orchestrator/blob/760c1f665101ee20247e021dbd24f828ec004cb2/src/puppetlabs/orchestrator/connection_tests.clj#L41-L51 

perhaps we should remain consistent? `echo 'connected'`?

ekinanp: Well, `whoami` was chosen b/c we think it'll be on all SSH and Windows targets. Can the same be said for echo?

McdonaldSeanp: Looks like echo exists on windows too: https://docs.microsoft.com/en-us/windows-server/administration/windows-commands/echo

Either is likely fine, whoami might be good b/c it doesn't require params so any nasty shell escaping bugs can be avoided in windows land.

ekinanp: I am fine with changing this to echo if we can agree that the probability of encountering shell escaping bugs is 0%. Otherwise, I'd like to use `whoami`.

nicklewis: If we use echo, we should avoid any punctuation and only print a single word. `echo Connected` would work. Anything else gets into either shell issues (bash vs. powershell) or differences in the implementations (`echo Successfully connected` will print on two lines in powershell).

If we go with `whoami` we should just make sure the doc about the workflow explains what the expected output is.

ekinanp: Thanks, updated to `echo Connected`.

modules/puppet_connect/spec/plans/test_input_data.rb
```diff
@@ -0,0 +1,82 @@
...
+    expect(winrm_target.config).to eql(winrm_config_before)
+  end
+
+  it 'checks if the targets are connectable' do
+    expect_command('echo Connected')
+      .be_called_times(1)
+      .with_targets(['ssh_target', 'winrm_target'])
+      .return_for_targets({
+        'ssh_target' => {
+          'stdout' => 'root'
```
donoghuc, line 68: In the spirit of keeping mocks in line with reality, i think these should be `Connected`.

ekinanp: Ah thanks. That's what I get for doing find and replace. Updated.

donoghuc (approved):

modules/puppet_connect/plans/test_input_data.pp
ekinanp: Updated the workflow doc.

PR #2621, opened 2021-02-04
(maint) Ensure Bolt PATH environment variable in beaker

jpogran:
This adds a step that adds the Bolt installation path to the PATH environment variable. This happens after the bolt msi install, so that subsequent commands will know where to find bolt.


lucywyman (approved):
Ran locally on a Windows controller and Windows node and it succeeded!

Issue #2463, opened 2020-12-11
Enumerate `report` hash keys for ApplyResult


lucywyman:
## Describe the Change You Would Like
ApplyResults have a `report` method that returns a lengthy report on what Puppet applied. This hash can be pretty large and difficult to traverse and discover for users. It's the same structure documented in Puppet at https://puppet.com/docs/puppet/7.0/format_report.html#puppet::transaction::report. We should link to the Puppet documentation so users can discover what keys are available when parsing apply results, and include an example of printing all the log messages. Clarify this is a hash, not an object.

PR #2623, opened 2021-02-04
(GH-2463) Document ApplyResult report keys

beechtom:
This adds documentation for the keys that are in an `ApplyResult` report
hash, including an example of accessing the log messages in a report.

!no-release-note

### DIFF
--- a/documentation/applying_manifest_blocks.md
+++ b/documentation/applying_manifest_blocks.md
@@ -223,6 +223,57 @@ $results.each |$result| {
 }
 ```
 
+### Report keys
+
+The [`ApplyResult` object](bolt_types_reference.md#applyresult) includes a
+`report` method that returns a hash representation of the
+[`Puppet::Transaction::Report`](https://puppet.com/docs/puppet/7.3/format_report.html)
+object. Each property on the object corresponds to a key with the same name in
+the report hash. However, not every property is presented in the hash for every
+result. Only properties that have a value from the apply are present in the
+hash.
+
+For example, to access and print the logs from a report in your plan, you can
+access the `logs` key in the report hash:
+
+```puppet
+plan example (
+  TargetSpec $targets
+) {
+  # Install the puppet-agent package and gather facts
+  $targets.apply_prep
+
+  # Apply Puppet code
+  $apply_results = apply($targets) {
+    file { '/etc/puppetlabs': 
+      ensure => present
+    }
+  }
+
+  # Print log messages from the report
+  $apply_results.each |$result| {
+    $result.report['logs'].each |$log| {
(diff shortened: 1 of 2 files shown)


documentation/applying_manifest_blocks.md
```diff
@@ -223,6 +223,55 @@ $results.each |$result| {
 }
 ```
 
+### Report keys
+
+The [`ApplyResult`
+object](https://puppet.com/docs/bolt/latest/bolt_types_reference.html#applyresult)
+includes a `report` method that returns a hash representation of the
+[`Puppet::Transaction::Report`](https://puppet.com/docs/puppet/7.3/format_report.html)
+object. Each property on the object corresponds to a key with the same name in
```
lucywyman, line 232: I think it's worth noting that not every property will be represented in the hash for every result, I believe it's only those that have values from the apply (for example I don't think `job_id` will ever be in the report in Bolt)

lucywyman (approved):

documentation/applying_manifest_blocks.md
```diff
@@ -223,6 +223,58 @@ $results.each |$result| {
 }
 ```
 
+### Report keys
+
+The [`ApplyResult`
+object](https://puppet.com/docs/bolt/latest/bolt_types_reference.html#applyresult)
```
hestonhoffman, line 229:
```suggestion
object](bolt_types_reference.md#applyresult)
```

documentation/applying_manifest_blocks.md
```diff
@@ -223,6 +223,58 @@ $results.each |$result| {
...
+includes a `report` method that returns a hash representation of the
+[`Puppet::Transaction::Report`](https://puppet.com/docs/puppet/7.3/format_report.html)
+object. Each property on the object corresponds to a key with the same name in
+the report hash. However, not every property will be presented in the hash for
+every result. Only properties that have a value from the apply will be present
```
hestonhoffman, line 234:
```suggestion
the report hash. However, not every property is presented in the hash for
every result. Only properties that have a value from the apply are present
```

documentation/applying_manifest_blocks.md
```diff
@@ -223,6 +223,58 @@ $results.each |$result| {
...
+in the hash.
+
+For example, to access and print the logs from a report in your plan you can
+access the `logs` key in the report hash:
```
hestonhoffman, line 238:
```suggestion
For example, to access and print the logs from a report in your plan, you can
access the `logs` key in the report hash:
```

documentation/applying_manifest_blocks.md
beechtom: 🤦

hestonhoffman (approved):

PR #2628, opened 2021-02-08
Added documentation on how to configure Bolt to install Puppet agent without internet

nmaludy:
The default `puppet_agent::install` task goes out to the internet and downloads things. This breaks in "air gapped" environments. I thought it would be an interesting example to add to the documentation in order to showcase the `plugin-hooks` feature using existing tasks.

### DIFF
--- a/documentation/using_plugins.md
+++ b/documentation/using_plugins.md
@@ -183,6 +183,19 @@ plugin-hooks:
     task: <plugin name>
 ```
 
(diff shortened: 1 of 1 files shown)


beechtom (approved):

hestonhoffman (approved):

hestonhoffman:
Thanks @nmaludy!

PR #2631, opened 2021-02-09
(bug) Install Puppetfile without 'modules' configured

beechtom:
This fixes a bug where Bolt would not install a Puppetfile with `bolt
module install --no-resolve` if the `modules` key is not configured or
is an empty array.

!bug

* **Install Puppetfile without `modules` configured**

  Bolt now correctly installs a Puppetfile with `bolt module install
  --no-resolve` and `Install-BoltModule -NoResolve` even if the
  `modules` key is not configured or is an empty array.

### DIFF
--- a/documentation/bolt_installing_modules.md
+++ b/documentation/bolt_installing_modules.md
@@ -438,15 +438,6 @@ manually manage your project's Puppetfile.
 To manually manage a project's Puppetfile and install modules without resolving
 dependencies, follow these steps:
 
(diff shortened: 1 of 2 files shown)


hestonhoffman (approved):

lucywyman (approved):
Installs Puppetfile correctly when `modules` is not set or is an empty array. Not a lot of edge cases here that I can think of.

PR #2633, opened 2021-02-09
(GH-2633) Add Bolt Task directory to PSModulePath

jpogran:
This commit adds the bolt task target directory to the PSModulePath environment variable for the currently executing session. This will allow Bolt tasks to ship powershell modules that can be automatically imported during runtime instead of having to calculate relative paths.

To use this feature add any powershell module to the `files` metadata in the task.json file:

```
"files": [
  "<BOLT MODULE NAME>/files/<PWSH MODULE NAME>/"
],
```

Then inside your Bolt PowerShell task start using the content from that PowerShell module without having to write `Import-Module` statements with fully qualified paths to the temp directory Bolt operates in.

### DIFF
--- a/documentation/writing_tasks.md
+++ b/documentation/writing_tasks.md
@@ -967,6 +967,91 @@ Successful on 1 target: localhost
 Ran on 1 target in 0.12 seconds
 ```
 
+### PowerShell example
+
+Create task and metadata in your project at
+`<PROJECT DIRECTORY>/tasks/mytask.{json,ps1}`.
+
+**Metadata**
+
+```json
+{
+  "files": [
+    "powershell_task_helper/files/BoltPwshHelper/"
+  ],
+  "input_method": "powershell",
+  "parameters": {
+    "name": {
+      "description": "Name of user to run command",
+      "type": "String"
+    }
+  }
+}
+```
+
+**Task**
+
+```powershell
+#!/usr/bin/env pwsh
+[CmdletBinding()]
+Param(
+  [Parameter(Mandatory = $True)]
+  [String]
+  $Name
+)
+
+<#
+If using PowerShell 3.0, you will need to add an import statement here:
+Import-Module BoltPwshHelper
+#>
+​
+<#
+Handle unhandled exceptions using the `trap` keyword
+#>
+trap {
+  Write-BoltError -Message "Generic trap handler" -Exception $_
+}
+
+<#
+A example of a custom error messages based on a specific use case
+#>
+if ($name -eq 'Robert') {
+  Write-BoltError -Message "User ${name} not allowed"
+}
+else {
+  # TODO
+}
+
+<#
+An exmaple of returning a full exception stacktrace in Bolt formatted json
+You can add a `-Message` if you want
+#>
+try {
+  Write-Output (@{ "greeting" = "Hi, my name is ${name}"} | ConvertTo-JSON)
+}
+catch {
+  Write-BoltError -Exception $_
+}
+```
+
+> **Note**: If you're using a version of PowerShell more recent than 3.0, 
+  you don't have to write an import statement to load your PowerShell module. 
+  Bolt automatically loads the module by adding your task directory to 
+  `$env:PSModulePath`.
+
+**Output**
+
+```console
+$ bolt task run mymodule::mytask -n localhost name="Robert'); DROP TABLE Students;--"
+Started on localhost...
+Finished on localhost:
+  {
+    "greeting": "Hi, my name is Robert'); DROP TABLE Students;--"
+  }
+Successful on 1 target: localhost
+Ran on 1 target in 0.12 seconds
+```
+
 ## Secure coding practices for tasks
 
 Use secure coding practices when you write tasks and help protect your system.
--- a/lib/bolt/shell/powershell.rb
+++ b/lib/bolt/shell/powershell.rb
@@ -274,7 +274,12 @@ module Bolt
                               []
                             end
 
-          output = execute([Snippets.shell_init, *env_assignments, command].join("\n"))
+          output = execute([
+            Snippets.shell_init,
+            Snippets.append_ps_module_path(dir),
+            *env_assignments,
+            command
+          ].join("\n"))
 
           Bolt::Result.for_task(target, output.stdout.string,
                                 output.stderr.string,
--- a/lib/bolt/shell/powershell/snippets.rb
+++ b/lib/bolt/shell/powershell/snippets.rb
@@ -67,6 +67,12 @@ module Bolt
             PS
           end
 
+          def append_ps_module_path(directory)
+            <<~PS
+            $env:PSModulePath += ";#{directory}"
+            PS
+          end
+
           def ps_task(path, arguments)
             <<~PS
             $private:tempArgs = Get-ContentAsJson (


documentation/writing_tasks.md
```diff
@@ -955,6 +955,54 @@ Successful on 1 target: localhost
...
+#!/usr/bin/env pwsh
+[CmdletBinding()]
+param (
+  # Name or list of SCCM Trigger to run
+  [Parameter(Mandatory=$true)]
+  [string]$Name
+)
+
+<#
+If using PowerShell 3.0, you will need to add an import statement here:
```
hestonhoffman, line 984:
```suggestion
If using PowerShell 3.0, add an import statement here:
```

documentation/writing_tasks.md
```diff
@@ -955,6 +955,54 @@ Successful on 1 target: localhost
...
+Import-Module BoltPwshHelper
+#>
+​
+Write-BoltResult -Result @{ greeting = "Hi, my name is ${name}" }
+```
+
+> Note: Bolt automatically adds your task directory to the `$env:PSModulePath`, so PowerShell module auto loading will work without having to write out the import statement
```
hestonhoffman, line 991:
```suggestion
> **Note**: You don't have to write an import statement to load your PowerShell module. 
  Bolt automatically loads the module by adding your task directory to 
  `$env:PSModulePath`.
```

lib/bolt/shell/powershell.rb
```diff
@@ -274,7 +274,12 @@ def run_task(task, arguments, _options = {}, position = [])
                               []
                             end
 
-          output = execute([Snippets.shell_init, *env_assignments, command].join("\n"))
+          output = execute([
+            Snippets.shell_init,
+            Snippets.append_ps_module_path(task_dir),
```
beechtom, line 279: See comment for `#append_ps_module_path`. If making that change, this would be updated to take `dir` and `extra_files` as parameters.

documentation/writing_tasks.md
```diff
@@ -955,6 +955,58 @@ Successful on 1 target: localhost
...
+Create task and metadata in a new module at
+`~/.puppetlabs/bolt/site-modules/mymodule/tasks/mytask.{json,ps1}`.
+
+**Metadata**
+
+```json
+{
+  "files": ["powershell_task_helper/files/boltpwsh"],
+  "input_method": "powershell"
+}
```
beechtom, line 969:
```suggestion
  "parameters": {
    "name": {
      "description": "Name or list of SCCM Trigger to run",
      "type": "String"
    }
  }
}
```

Metadata needs to list the parameters the task accepts.

documentation/writing_tasks.md
```diff
@@ -955,6 +955,58 @@ Successful on 1 target: localhost
...
+#>
+​
+trap {
+  Write-Error -Message "Generic trap handler" -Exception $_
+}
+```
+
+> **Note**: You don't have to write an import statement to load your PowerShell module. 
+  Bolt automatically loads the module by adding your task directory to 
+  `$env:PSModulePath`.
```
hestonhoffman, line 995:
```suggestion
> **Note**: If you're using a version of PowerShell more recent than 3.0, 
  you don't have to write an import statement to load your PowerShell module. 
  Bolt automatically loads the module by adding your task directory to 
  `$env:PSModulePath`.
```

documentation/writing_tasks.md
```diff
@@ -955,6 +955,58 @@ Successful on 1 target: localhost
...
+### PowerShell example
+
+Create task and metadata in a new module at
+`~/.puppetlabs/bolt/site-modules/mymodule/tasks/mytask.{json,ps1}`.
+
+**Metadata**
+
+```json
+{
+  "files": ["powershell_task_helper/files/boltpwsh"],
```
beechtom, line 967:
```suggestion
  "files": ["powershell_task_helper/files/BoltPwshHelper/"],
```

beechtom (commented):
It looks like the example task is incomplete.

documentation/writing_tasks.md
```diff
@@ -955,6 +955,58 @@ Successful on 1 target: localhost
 Ran on 1 target in 0.12 seconds
 ```
 
+### PowerShell example
+
+Create task and metadata in a new module at
+`~/.puppetlabs/bolt/site-modules/mymodule/tasks/mytask.{json,ps1}`.
```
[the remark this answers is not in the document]
jpogran: This was a copy pasta from the ruby/python example so we have more to change

lib/bolt/shell/powershell/snippets.rb
```diff
@@ -67,6 +67,13 @@ def run_script(arguments, script_path)
             PS
           end
 
+          def append_ps_module_path(directory)
+            <<~PS
+            $parent = Split-Path "#{directory}"
+            $env:PSModulePath += ";${parent}"
+            PS
+          end
```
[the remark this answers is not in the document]
jpogran: If we add the task directory itself, PowerShell traverses the tree to find any module there. So any powershell module included in the puppet module that the task is in is available to autoload. We don't have to add directories specifically. This works like Puppet or Bolt in that respect.

We could be more restrictive if we only add the `files` directory, then only content in that folder is available for autoload.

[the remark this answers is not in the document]
beechtom: Makes sense. Should the above be `Split-Path "#{directory}" -Parent` in that case? Currently it's just adding the `tasks/` directory to the path, which won't include the extra files listed in the metadata.

Also, do we eventually want to support this behavior for files from other modules (like the powershell task helper)?

documentation/writing_tasks.md
```diff
@@ -967,6 +967,67 @@ Successful on 1 target: localhost
 Ran on 1 target in 0.12 seconds
 ```
 
+### PowerShell example
+
+Create task and metadata in your project at
+`<PROJECT>/tasks/mytask.{json,ps1}`.
```
lucywyman, line 973:
```suggestion
`<PROJECT DIRECTORY>/tasks/mytask.{json,ps1}`.
```
Asked about this on one of Tom's PRs too - we standardized on this a few months ago, but I'm not sure if `PROJECT` is better....

documentation/writing_tasks.md
```diff
@@ -967,6 +967,67 @@ Successful on 1 target: localhost
...
+)
+
+<#
+If using PowerShell 3.0, add an import statement here:
+Import-Module BoltPwshHelper
+#>
+​
+trap {
+  Write-BoltError -Message "Generic trap handler" -Exception $_
+}
```
beechtom, line 1010: We should use the complete task and metadata example from https://github.com/jpogran/puppetlabs-powershell_task_helper/tree/pwsh-content/examples , since this example doesn't produce the output shown below.

beechtom (approved):
Just needs an updated example, but otherwise lgtm 👍

documentation/writing_tasks.md
jpogran: Updated

beechtom (approved):
Do we want to add the powershell_task_helper module to the Puppetfile in this PR or a separate one? Either works.

jpogran:
@beechtom I can add it, but the powershell_task_helper hasn't been pushed to the forge yet so that needs to be done before we could merge this PR

beechtom:
Let's go ahead and merge this and then add the module separately, since ci-job-configs also needs to be updated.

hestonhoffman (approved):

Issue #2240, opened 2020-10-06
Document task parameter constraints


adreyer:
## Describe the Change You Would Like
We don't clearly document the constraints on parameter names in the writing tasks page. We should document that https://puppet.com/docs/bolt/latest/writing_tasks.html#defining-parameters-in-tasks

tdtheautomator:
would like to see how this progresses as it becomes crucial to adopting tasks at a wider level else it comes from trial and error

PR #2634, opened 2021-02-09
(GH-2240) Document task parameter constraints

beechtom:
This adds documentation for constraints around task parameter names and
types.

!no-release-note

### DIFF
--- a/documentation/writing_tasks.md
+++ b/documentation/writing_tasks.md
@@ -185,6 +185,18 @@ If your task must receive parameters only in a certain way, such as only
 `input_method`. For more information on `input_method`, see [Task metadata
 reference](#task-metadata-fields).
 
+### Task parameter constraints
+
+There are a couple important constraints to keep in mind when defining your
+task's parameters:
+
+- Task parameter names must begin with a lowercase letter, and can only include
+  lowercase letters (`a-z`), digits (`0-9`), and underscores (`_`).
+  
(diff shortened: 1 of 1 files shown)


documentation/writing_tasks.md
```diff
@@ -185,6 +185,18 @@ If your task must receive parameters only in a certain way, such as only
 `input_method`. For more information on `input_method`, see [Task metadata
 reference](#task-metadata-fields).
 
+### Task parameter constraints
+
+There are a couple important constraints to keep in mind when defining your
+task's parameters:
+
+- Task parameter names must begin with a lowercase letter, and may only include
```
hestonhoffman, line 193:
```suggestion
- Task parameter names must begin with a lowercase letter, and can only include
```
😁

hestonhoffman (approved):
One nit, but looks good!

documentation/writing_tasks.md
beechtom: Noooo! 😂 I've been trying so hard not to use "may".

hestonhoffman: You're doing great! This is kind of a tricky one cos you can't use might here. It's actually part of the reason we don't like"may" - the ambiguity for non-native speakers.

beechtom: Makes total sense 👍  Thanks Heston!

PR #2636, opened 2021-02-10
(GH-2636) Fix PowerShell Cmdlet Version detection

jpogran:
This removes the PowerShell cmdlets to call `bolt --version` on windows platforms. This was originally done for speed reasons, but can be removed to support cross platform use of the module since the ruby runtime will be patched to run faster on windows.

### DIFF
--- a/pwsh_module/command.tests.ps1
+++ b/pwsh_module/command.tests.ps1
@@ -34,7 +34,7 @@ Describe "test bolt module" {
 
     it "has the correct number of exported functions" {
       # should count of pwsh functions
-      @($commands).Count | Should -Be 22
+      @($commands).Count | Should -Be 23
     }
   }
 }
--- a/pwsh_module/pwsh_bolt_internal.ps1
+++ b/pwsh_module/pwsh_bolt_internal.ps1
@@ -68,3 +68,11 @@ function Get-BoltCommandline {
 
   Write-Output $params
 }
+
+function Get-BoltVersion{
+  [CmdletBinding()]
+  param()
+
+  $module = Get-Module -Name PuppetBolt
+  Write-Output [string]($module.Version)
+}
--- a/pwsh_module/pwsh_bolt.psd1.erb
+++ b/pwsh_module/pwsh_bolt.psd1.erb
@@ -12,6 +12,7 @@
   Copyright         = '(c) 2017 Puppet, Inc. All rights reserved'
   PowerShellVersion = '3.0'
   FunctionsToExport = @(
+    'Get-BoltVersion',
     '<%= @commands.sort_by { |cmd| cmd[:cmdlet] }.map{|cmd| cmd[:cmdlet] }.join("',\n    '") %>'
   )
   CmdletsToExport   = @()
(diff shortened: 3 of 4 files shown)


pwsh_module/pwsh_bolt_internal.ps1
```diff
@@ -68,3 +68,17 @@ function Get-BoltCommandline {
 
   Write-Output $params
 }
+
+function Get-BoltVersion{
```
beechtom, line 72: Does it make sense to export this function so it can be used on its own?

pwsh_module/pwsh_bolt_internal.ps1
```diff
@@ -68,3 +68,17 @@ function Get-BoltCommandline {
...
+  [CmdletBinding()]
+  param()
+  $IsLinuxEnv = (Get-Variable -Name "IsLinux" -ErrorAction Ignore) -and $IsLinux
+  $IsMacOSEnv = (Get-Variable -Name "IsMacOS" -ErrorAction Ignore) -and $IsMacOS
+  $IsWinEnv = !$IsLinuxEnv -and !$IsMacOSEnv
+  if($IsWinEnv){
```
beechtom, line 78: Is checking the module's version not guaranteed to work on macOS/Linux? Importing the module on macOS and running `[string](Get-Module -Name PuppetBolt).Version` seems to work fine at least.

beechtom (commented):
Works as intended! 👍 Just a couple questions for my own curiosity, but otherwise lgtm.

beechtom (approved):

pwsh_module/pwsh_bolt_internal.ps1
jpogran: I think it makes sense to make it a module export because the only way to get the version currently is to use a `get` cmdlet like `Get-BoltTask -version` and that doesn't make much sense.

pwsh_module/pwsh_bolt_internal.ps1
jpogran: The original intent was to provide a faster way to get the version of Bolt without executing ruby. Since the `version.txt` isn't updated reliably, moving to the `PuppetBolt` module version makes sense, so long as we always keep the module the same version as the released bolt version.

Since the module can be released without making a bolt MSI release, theoretically the two could be versioned separately. If we commit to always have a 1:1 release of the MSI and the module, then I think we can safely use the module version.

lucywyman (approved):

pwsh_module/command.tests.ps1
```diff
@@ -5,6 +5,10 @@ BeforeAll {
     return "bolt " + $params -join " "
   }
 
+  Mock -ModuleName 'PuppetBolt' -Verifiable -CommandName Get-BoltVersion -MockWith {
+    return "3.0.0"
```
lucywyman, line 9: Are we going to need to update this every time we release, or can it be mocked?

jpogran: Thanks for catching this. This was needed in a prior version of the PR, but is not needed with the final version of the `Get-BoltVersion` function. Removing this now

If this were staying in, it wouldn't matter that the version was hardcoded because the tests were testing parameter presence, not version correctness.

PR #2642, opened 2021-02-16
(maint) Update powershell helpers to escape correctly for linux

McdonaldSeanp:
This commit updates the PuppetBolt powershell module to correctly escape hash
parameters for linux platforms. Conversation in this thread:
https://github.com/MicrosoftDocs/PowerShell-Docs/issues/2361 describe the
circumstances of why interpolation breaks.

The fix is to include backslash escape chars for all quotes in hash parameters
if the powershell function is running on linux


pwsh_module/pwsh_bolt_internal.ps1
```diff
@@ -42,7 +42,19 @@ function Get-BoltCommandline {
...
+      #
+      # See https://github.com/MicrosoftDocs/PowerShell-Docs/issues/2361
+      #
+      # $IsLinux is an automatic variable from powershellcore:
+      # https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_automatic_variables?view=powershell-7.1#islinux
+      if ($IsLinux) {
+        $params += $v -replace '"', '\"'
+      } else {
+        $params += "'$($v)'"
+      }
```
jpogran, line 57:
```suggestion
     $IsLinuxEnv = (Get-Variable -Name "IsLinux" -ErrorAction Ignore) -and $IsLinux
     $IsMacOSEnv = (Get-Variable -Name "IsMacOS" -ErrorAction Ignore) -and $IsMacOS
     $IsWinEnv = !$IsLinuxEnv -and !$IsMacOSEnv
     if($IsWinEnv){
       $params += "'$($v)'"
     } else {
       $params += $v -replace '"', '\"'
     }
```

jpogran: I think we'll need the get-variable pattern to handle being run in 5.1, and we'll need to account for macox not being detected as linux.

jpogran (changes requested):

pwsh_module/pwsh_bolt_internal.ps1
McdonaldSeanp: Updated

jpogran (approved):

jpogran:
@McdonaldSeanp just need to update CI to run these tests on nix and macos, and adjust the pester tests to account for the text changes and we're good to go!

PR #2645, opened 2021-02-17
(maint) Update Puppetfile to include puppet_connect module

McdonaldSeanp:
This commit updates the Puppetfile bolt uses during packaging steps to include
the puppet_connect module so that it isn't overwritten by r10k during builds

### DIFF
--- a/lib/bolt/cli.rb
+++ b/lib/bolt/cli.rb
@@ -899,7 +899,8 @@ module Bolt
     # Gem installs include the aggregate, canary, and puppetdb_fact modules, while
     # package installs include modules listed in the Bolt repo Puppetfile
     def incomplete_install?
-      (Dir.children(Bolt::Config::Modulepath::MODULES_PATH) - %w[aggregate canary puppetdb_fact secure_env_vars]).empty?
+      builtin_module_list = %w[aggregate canary puppetdb_fact secure_env_vars puppet_connect]
+      (Dir.children(Bolt::Config::Modulepath::MODULES_PATH) - builtin_module_list).empty?
     end
 
     # Mimicks the output from Outputter::Human#fatal_error. This should be used to print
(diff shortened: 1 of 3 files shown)


donoghuc (approved):

nicklewis:
Do we also need a corresponding update to the hardcoded list of bundled content used for detecting rubygems installs?

donoghuc:
I dont think so. https://github.com/puppetlabs/bolt/blob/8e14757f6ea72f09c8bc5f512b8a7d7dbe4cb979/bolt.gemspec#L24-L36 looks like we are covered

nicklewis:
This is what I was referring to:

https://github.com/puppetlabs/bolt/blob/main/lib/bolt/cli.rb#L902

I think this PR as it stands will cause the warning about gem installs to never trigger.

donoghuc:
Oh you mean this https://github.com/puppetlabs/bolt/blob/8e14757f6ea72f09c8bc5f512b8a7d7dbe4cb979/lib/bolt/cli.rb#L899-L903

lib/bolt/cli.rb
```diff
@@ -899,7 +899,7 @@ def bundled_content
     # Gem installs include the aggregate, canary, and puppetdb_fact modules, while
     # package installs include modules listed in the Bolt repo Puppetfile
     def incomplete_install?
-      (Dir.children(Bolt::Config::Modulepath::MODULES_PATH) - %w[aggregate canary puppetdb_fact secure_env_vars]).empty?
+      (Dir.children(Bolt::Config::Modulepath::MODULES_PATH) - %w[aggregate canary puppetdb_fact secure_env_vars puppet_conenct]).empty?
```
nicklewis, line 902:
```suggestion
      (Dir.children(Bolt::Config::Modulepath::MODULES_PATH) - %w[aggregate canary puppetdb_fact secure_env_vars puppet_connect]).empty?
```

nicklewis (changes requested):

lib/bolt/cli.rb
McdonaldSeanp: damnit

beechtom (approved):

donoghuc (approved):

PR #2647, opened 2021-02-17
(maint) Ignore non-built-in modules

beechtom:
This updates the `.gitignore` file to ignore all contents in the
`modules/` directory except for built-in modules (aggregate, canary,
puppet_connect, puppetdb_fact).

!no-release-note

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -44,36 +44,12 @@ project.yaml
 
 acceptance/hosts.yaml
 
-modules/apply
-modules/facts
-modules/package
-modules/puppet_agent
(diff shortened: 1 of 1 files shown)


beechtom:
Thought this would be helpful as it (very slightly) lessens the number of changes needed when adding a new bundled module. The trade-off is that any new built-in modules will be ignored, but I don't know how likely it is that we'll add more of those.

jpogran (approved):

lucywyman (approved):

Issue #2615, opened 2021-02-03
Document unit testing Bolt plans with BoltSpec


lucywyman:
## Use Case
Pull the documentation from the BoltSpec::Plans library into the actual docs, and detail how to unit test plans (including mocking subplan outputs) with Bolt.

Include examples!

PR #2649, opened 2021-02-18
(GH-2615) Add BoltSpec documentation

beechtom:
This adds documentation for the `BoltSpec` library, including a
reference for available mocks/stubs, examples, and instructions on
setting up the testing environment.

This also adds a new `config_data` helper to set configuration for
tests, similar to the existing `inventory_data` helper.

!feature

* **Added `config_data` helper to `BoltSpec` library**
  ([#2615](https://github.com/puppetlabs/bolt/issues/2615))

  The `BoltSpec` library includes a new `config_data` helper which can
  be used to set Bolt configuration in your plan unit tests.


beechtom:
Putting this up now to get the review process started since it's fairly lengthy. Still working on adding at least one more example for testing a plan with sub-plans.

hestonhoffman (commented):
I'm not done yet, but I didn't want to lose my comments :)

documentation/boltspec_reference.md
```diff
@@ -0,0 +1,1117 @@
...
+      Bolt::Result.new(target, value: { 'path' => File.join(destination, source) })
+    end
+
+    Bolt::ResultSet.new(results)
+  end
+  ```
+
+- `error_with(error)`
+
+  Sets the error hash for each target's `Result` when the file is downloadedn.
```
hestonhoffman, line 744:
```suggestion
  Sets the error hash for each target's `Result` when the file is downloaded.
```

hestonhoffman:
Okay, I'm done!

documentation/testing_plans.md
```diff
@@ -0,0 +1,783 @@
...
+expect_task('configure')
+expect_task('configure').with_targets(['servers'])
+```
+
+`BoltSpec` does not match the `run_task` invocation to the third mock because it
+uses the `with_params` modifier to add more specificity. Because `BoltSpec`
+found multiple matching mocks, it will use the last one defined in the test:
+`expect_task('configure').with_targets(['servers'])`. The test will then fail,
+as the first mock is not fulfilled, as the plan does not run the `configure`
+task a second time.
```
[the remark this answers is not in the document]
beechtom: All mocks need their assertions to be fulfilled for a test to pass since a mock is an assertion that something will happen. When `BoltSpec` uses a mock, it increments a counter on the mock to indicate how many times it has been met. Because `BoltSpec` only uses the last matching mock, only the last matching mock will have its counter incremented.

documentation/testing_plans.md
```diff
@@ -0,0 +1,783 @@
+# Testing plans
+
+Bolt ships with a library of helpers, named `BoltSpec`, intended to be used for
+writing unit tests for plans with the `RSpec` testing tool. Before writing unit
+tests for plans, get familiar with [RSpec](https://rspec.info/).
+
+## Set up the test environment
```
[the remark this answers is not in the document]
beechtom: Hm, good question... I'm not sure what the best approach is since I don't think we should imply plans can only be tested in modules. Technically you can add simplified `Gemfile`, `Rakefile`, and `spec_helper.rb` files to your project manually, and then the tests will work normally. You should also be able to convert your project to a module to get those files from PDK, though you'd get a ton of extraneous files with them.

Maybe two subsections: "Testing plans in a module" and "Testing plans in a project"?

documentation/testing_plans.md
```diff
@@ -0,0 +1,783 @@
...
+# Configure Puppet and Bolt for testing
+BoltSpec::Plans.init
+
+# Additional spec_helper configuration . . .
+```
+
+### Include BoltSpec in test files
+
+To expose the `BoltSpec` helper functions and configure Puppet and Bolt for
+a single test file, add the following lines to the test file:
```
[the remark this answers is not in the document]
beechtom: There's a comment in BoltSpec that says running `BoltSpec::Plans.init` probably isn't safe to do in modules that also test manifest code. I _think_ that only applies if you're doing that in `spec_helper.rb`, since it would apply it to all tests.

We could add a line that suggests doing it from `spec_helper.rb` unless you're also testing manifest code?

documentation/testing_plans.md
```diff
@@ -0,0 +1,783 @@
...
+
+- _Stubs_ are used to allow the invocation of executor functions and set their
+  return values.
+
+- _Mocks_ are used to make assertions about executor function invocations and
+  set their return values.
+
+A helpful way to remember the difference between a stub and a mock, and when to
+use one or the other, is that stubs _allow_ you to invoke executor functions,
+while mocks _expect_ executor functions to be invoked. Each stub and mock
```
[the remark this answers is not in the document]
beechtom: The sad thing is I read that line multiple times and knew it was off :|

documentation/testing_plans.md
```diff
@@ -0,0 +1,783 @@
...
+Executor function invocations are matched to any stub or mock that is as
+specific or less specific than the function invocation, but not to any stub or
+mock that is more specific than the function invocation. If the function
+invocation matches multiple stubs or mocks, it will use the last stub or mock
+that matched.
+
+For example, the following plan invokes the `run_task` function to run the
+`configure` task on a list of targets:
+
+And the following plan which runs the `configure` task on a list of targets:
```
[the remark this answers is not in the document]
beechtom: Lol, ya, it refers to the same plan. :P I was moving the example around and forgot to delete that extra line.

documentation/testing_plans.md
```diff
@@ -0,0 +1,783 @@
...
+Save the following plan to `patch/plans/init.pp`:
+
+```puppet
+plan patch (
+  TargetSpec $targets
+) {
+  # Group targets by puppet-agent status
+  $_targets = run_plan('patch::get_targets', 'targets' => $targets)
+
+  # Run the patching task on agentful targets
```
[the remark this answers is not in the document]
beechtom: 👍 Opted to "targets with agents". Hopefully that is okay in lieu of "nodes with agents" since Bolt terminology does not use "nodes"?

documentation/testing_plans.md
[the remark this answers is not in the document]
hestonhoffman: Yeah, I think 2 sections makes sense

documentation/testing_plans.md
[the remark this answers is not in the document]
hestonhoffman: Cool - yeah I language around preferring spec_helper unless you're testing manifests makes sense

documentation/boltspec_reference.md
```diff
@@ -0,0 +1,1117 @@
...
+tests for plans. For more information about writing unit tests for plans, and
+how to use these functions, see [Testing plans](testing_plans.md).
+
+## Mocks
+
+Mocks serve two purposes when you are writing tests for your plans: they allow
+assertions about executor functions invoked during a plan run and also allow you
+to set their return values. For example, you might use a mock to make an
+assertion that the `run_task` function is invoked exactly one time during a
+test.
```
lucywyman, line 13:
```suggestion
test, and/or that it returns 'foo' on stdout.
```

documentation/boltspec_reference.md
```diff
@@ -0,0 +1,1117 @@
+# BoltSpec reference
+
+The `BoltSpec` helper library includes several functions to help write unit
+tests for plans. For more information about writing unit tests for plans, and
+how to use these functions, see [Testing plans](testing_plans.md).
+
+## Mocks
+
+Mocks serve two purposes when you are writing tests for your plans: they allow
+assertions about executor functions invoked during a plan run and also allow you
```
lucywyman, line 10:
```suggestion
assertions that executor functions are invoked during a plan run and also allow you
```

documentation/boltspec_reference.md
```diff
@@ -0,0 +1,1117 @@
...
+to set their return values. For example, you might use a mock to make an
```
lucywyman, line 11:
```suggestion
to mock their return values. For example, you might use a mock to make an
```

lucywyman: idk if this is just more confusing, like using a word in it's own definition :P

documentation/boltspec_reference.md
```diff
@@ -0,0 +1,1117 @@
...
+assertion that the `run_task` function is invoked exactly one time during a
+test.
+
+### `expect_command`
```
lucywyman, line 15: This page overall seems pretty verbose, and like it includes a lot of duplication of expectations. What do you think about having a separate `Expectations` page that lists all the expectations, and then each function can just list the expectations it accepts and link to the Expectations reference page?

beechtom: I don't think verbosity/length should be a concern for reference pages since they're not meant to read straight through and have a ToC for easy navigation. Some of the modifiers also have different behavior (`be_called_times`) or have different parameters (`always_return`, `return_for_targets`, `return`) depending on the mock/stub used. We _could_ enumerate the exceptions for each mock/stub under the modifier, but I think that would make the page more confusing than anything.

documentation/boltspec_reference.md
beechtom: I think using mock in two contexts here could be confusing, so I'll leave it as 'set'.

documentation/testing_plans.md
```diff
@@ -0,0 +1,808 @@
...
+
+## Include BoltSpec functions
+
+To use the `BoltSpec` library functions, include them in either the
+`spec_helper.rb` file or in the individual test files, and configure Puppet and
+Bolt for the testing environment.
+
+> 🔩 **Tip**: Unless your module also includes tests for manifest code, add the
+> `BoltSpec` helpers and configure Puppet and Bolt in `spec_helper.rb`. This makes
+> the helpers available to all of your tests.
```
hestonhoffman, line 73:
```suggestion
> 🔩 **Tip**: Unless your module also includes tests for manifest code, add the
> `BoltSpec` helpers and configure Puppet and Bolt in `spec_helper.rb` instead
> of including them in the individual test files. This makes the helpers available to 
> all of your tests.
```

documentation/testing_plans.md
```diff
@@ -0,0 +1,808 @@
...
+  TargetSpec $databases
+) {
+  $server_results   = run_task('facts', $servers)
+  $database_results = run_task('facts', $databases)
+  $results          = ResultSet.new($server_results.results + $database_results.results)
+  return $results
+}
+```
+
+Modifiers can also be chained, allowing you to add as much or as little
```
hestonhoffman, line 325:
```suggestion
You can chain modifiers, allowing you to add as much or as little
```

documentation/testing_plans.md
```diff
@@ -0,0 +1,808 @@
...
+
+### Matching stubs and mocks
+
+When you run your tests and a plan invokes an executor function, `BoltSpec`
+matches the function invocation to any stubs or mocks created for the test.
+Executor function invocations are matched to any stub or mock that is as
+specific or less specific than the function invocation, but not to any stub or
+mock that is more specific than the function invocation. If the function
+invocation matches multiple stubs or mocks, it uses the last stub or mock that
+matched.
```
hestonhoffman, line 352:
```suggestion
When you run your tests and a plan invokes an executor function, `BoltSpec`
matches the function invocation to any stubs or mocks created for the test.
`BoltSpec` matches executor function invocations to any stub or mock that is as
specific or less specific than the function invocation, but not to any stub or
mock that is more specific than the function invocation. If the function
invocation matches multiple stubs or mocks, `BoltSpec` uses the last stub or mock 
that matched.

For a test to pass, `BoltSpec` must find a match for each stub or mock in the test. 
```
Adding that last bit because I think it's what I got tripped up on later.

documentation/testing_plans.md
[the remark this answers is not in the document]
hestonhoffman: Okay cool, I updated this edit and added a sentence above to clarify that.

documentation/testing_plans.md
```diff
@@ -0,0 +1,808 @@
...
+    run_plan('patch', 'targets' => 'agents')
+  end
+end
+```
+
+Run the tests:
+
+```shell
+bundle exec rake spec
+```
```
hestonhoffman, line 808: Could you add a related info section here? The backticks are throwing my suggestion off:

```
📖 Related information

- For more information on the available mocks and stubs, see the [BoltSpec reference](boltspec_reference.md).
```

documentation/boltspec_reference.md
hestonhoffman: Yeah, I agree. I think the way we expect users to interact with these reference pages is by using the TOC or using CMD+F, and it's reasonable for reference pages to be fairly repetitive for the sake of completeness. I'm also a little wary of adding more reference pages than we have to cos the references section is starting to get kinda chonky.

documentation/boltspec_reference.md
```diff
@@ -0,0 +1,1117 @@
...
+  ```
+
+- `error_with(error)`
+
+  Sets the error hash for each target's `Result` when the file is uploaded.
+  Returns a `Bolt::ResultSet` object.
+
+  ```ruby
+  allow_upload('sshd_config').error_with('msg' => 'Not authorized')
+  ```
```
hestonhoffman, line 1117: Similar to the testing doc, I think it's nice to bookend these with a related info section:

```
📖 Related information

- For more information on using `BoltSpec`, see [Testing plans](testing_plans.md).
```

hestonhoffman (approved):
A couple more suggestions, but this is looking good!

documentation/testing_plans.md
beechtom: 👍 Added. Modified it to only say 'each mock', since stubs don't make assertions and thus don't need to be matched for a test to pass.

documentation/boltspec_reference.md
```diff
@@ -0,0 +1,1121 @@
+# BoltSpec reference
```
lucywyman, line 1: Is there a reason not to include `execute_no_plan` and `execute_any_plan` here? [This PR](https://github.com/puppetlabs/bolt/commit/6fdbaf25f3d94d54cfff0e6605c6154e0365618b#diff-f84e1d7c038c6793ef936447c02fa9c43dbf03afe70213d0b3f7eb75dc71509bR176) includes an example and documentation.

beechtom: I did have that originally, but removed it cause it was just a duplicate of what was in `testing_plans.md`. I can add it back.

beechtom: Here it is in `testing_plans.md` btw: https://github.com/puppetlabs/bolt/pull/2649/files#diff-2c26325655047f2d520c02a0f4cb9101097e114b65d6db4ed0c551940b502e42R444

hestonhoffman: Yeah, I think it's best to have them in the reference as well for completeness. And then it probably wouldn't hurt to link to testing_plans from each

documentation/testing_plans.md
```diff
@@ -0,0 +1,815 @@
...
+
+Bolt ships with a library of helpers, named `BoltSpec`, intended to be used for
+writing unit tests for plans with the `RSpec` testing tool. Before writing unit
+tests for plans, get familiar with [RSpec](https://rspec.info/).
+
+`BoltSpec` requires other Puppet testing tools like `rspec-puppet`. The [Puppet
+Development Kit (PDK)](https://puppet.com/docs/pdk/latest/pdk.html) provides
+commands for installing and configuring these tools in new or existing modules.
+If your plan is part of a Bolt project, you can either move the plan to a module
+created with PDK or convert your project to a module.
```
lucywyman, line 11:
```suggestion
created with PDK or [manually setup rspec-puppet](https://rspec-puppet.com/documentation/setup/) in your Bolt project.
```

documentation/testing_plans.md
```diff
@@ -0,0 +1,815 @@
...
+
+```shell
+bundle exec rake spec
+```
+
+This command runs a rake task that is defined in the `Rakefile` created by PDK,
+and should execute successfully before you have written any tests. After you've
+written tests and run the rake task, the task prints a list of your tests to the
+console together with each test's pass or fail status.
+
```
lucywyman, line 141: To run an individual spec test, run:
```
bundle exec rake spec_prep
bundle exec rspec spec/plans/myplan_spec.rb
...
bundle exec rake spec_clean
```

documentation/testing_plans.md
```diff
@@ -0,0 +1,815 @@
...
+function. This function should return an array of strings, where each string is
+the path to a directory that includes modules.
+
+```ruby
+def modulepath
+  [File.expand_path('../fixtures/modules')]
+end
+```
+
+## Running plans
```
lucywyman, line 196:
```suggestion
## Acceptance Testing for Bolt Plans
```
Since this actually runs the plan with the provided parameters, I think it's safe to call this acceptance testing.

documentation/testing_plans.md
```diff
@@ -0,0 +1,815 @@
...
+```ruby
+it 'asks who you are' do
+  result = run_plan('whoami', {})
+
+  expect(result.ok?).to be(true)
+  expect(result.value).to eq('Who are YOU?')
+end
+```
+
+## Stubs and mocks
```
lucywyman, line 235:
```suggestion
## Unit testing for Bolt plans
```

documentation/testing_plans.md
```diff
@@ -0,0 +1,815 @@
...
+
+The `BoltSpec` library includes several functions used to mock and stub executor
```
lucywyman, line 237:
```suggestion
The `BoltSpec` library includes several functions used to mock and stub plan
```

lucywyman: While these go through the executor I think users are used to interacting with them through plans.

documentation/testing_plans.md
```diff
@@ -0,0 +1,815 @@
...
+```
+
+## Execution modes
+
+Plans often execute sub-plans with the `run_plan` function to build complex
+workflows. When testing these plans, it might be helpful to execute any
+sub-plans as well without needing to stub or mock the plan. To support this,
+`BoltSpec` offers two different execution modes:
+
+- `execute_any_plan`
```
lucywyman, line 451: Ok, I *swear* I searched both this page and my checkout before commenting above :P Grep has failed me.

documentation/testing_plans.md
```diff
@@ -0,0 +1,815 @@
...
+expect_task('configure')
+expect_task('configure').with_targets(['servers'])
+```
+
+`BoltSpec` does not match the `run_task` invocation to the third mock because it
+uses the `with_params` modifier to add more specificity. Because `BoltSpec`
+found multiple matching mocks, it will use the last one defined in the test:
+`expect_task('configure').with_targets(['servers'])`. The test will then fail,
+as the first mock is not fulfilled, as the plan does not run the `configure`
+task a second time.
```
hestonhoffman, line 393:
```suggestion
`expect_task('configure').with_targets(['servers'])`. Because plan does
not run the `configure` task a second time, the first mock is not fulfilled, 
and the test fails.
```
Just removing the future + passive

documentation/testing_plans.md
```diff
@@ -0,0 +1,825 @@
...
+  # Configure Puppet and Bolt before running any tests
+  before(:all) do
+    BoltSpec::Plans.init
+  end
+
+  # Set the return value of puppet_agent::version for each target
+  # Stub the patch and puppet_agent::install tasks
+  before(:each) do
+    allow_task('puppet_agent::version').return_for_targets({
+      'agent-target'     => { 'version' => '7.0.0' },
```
lucywyman, line 768: This can only stub `stdout` or `stderr` :thinking: it's a bit of a pickle since we still want the task to be runnable on its own too...you could maybe try `puppet --version` and filter on if that fails or not?

documentation/testing_plans.md
```diff
@@ -0,0 +1,825 @@
...
+require 'bolt_spec/plans'
+
+describe 'test' do
+  include BoltSpec::Plans
+
+  # Configure Puppet and Bolt before running any tests
+  before(:all) do
+    BoltSpec::Plans.init
+  end
+
```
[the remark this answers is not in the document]
beechtom: This shouldn't be required, since that's the default modulepath provided by rspec-puppet.

documentation/testing_plans.md
```diff
@@ -0,0 +1,825 @@
...
+```ruby
+expect_task('configure')
+expect_task('configure').with_targets(['servers'])
+```
+
+`BoltSpec` does not match the `run_task` invocation to the third mock because it
+uses the `with_params` modifier to add more specificity. Because `BoltSpec`
+found multiple matching mocks, it will use the last one defined in the test:
+`expect_task('configure').with_targets(['servers'])`. The test will then fail,
```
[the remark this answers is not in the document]
beechtom: That mock isn't matched, since it's more specific than the function invocation in the plan.

Cleared up the language in this paragraph since I think it wasn't as clear as it could be.

hestonhoffman (approved):

documentation/testing_plans.md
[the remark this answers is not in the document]
lucywyman: Ack apologies, I meant to delete the comment!

documentation/testing_plans.md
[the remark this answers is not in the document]
lucywyman: Oh, hm...it didn't work for me locally without adding this.

documentation/testing_plans.md
lucywyman: Hm, they're failing for me locally and if I'm reading this right it applies to `return_for_targets` as well https://github.com/puppetlabs/bolt/blob/main/lib/bolt_spec/plans.rb#L106

beechtom: That says for command and script, while this is for the `allow_task` stub. Further down in that comment we also have several examples of task stubs/mocks returning values without `stdout` or `stderr` keys.

I'm not sure why these are failing for you since I'm able to run them successfully locally...

documentation/testing_plans.md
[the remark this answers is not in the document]
beechtom: Are you modifying the modulepath elsewhere, like in your spec_helper? I'm working in a module made with `pdk new module` and there's no need to configure the modulepath to get the tests to run successfully.

documentation/testing_plans.md
```diff
@@ -0,0 +1,831 @@
...
+```puppet
+plan patch::get_targets (
+  TargetSpec $targets
+) {
+  # Get puppet-agent version on all targets
+  $_targets = get_targets($targets) 
+  $results  = run_task('puppet_agent::version', $_targets)
+
+  # Group targets by puppet-agent status
+  $agents    = $results.filter_set |$result| { $result['version'] != undef }.targets
```
lucywyman, line 677:
```suggestion
  $agents    = $results.filter_set |$result| { $result['stdout']['version'] != undef }.targets
```

lucywyman (approved):

Issue #2504, opened 2020-12-18
Referencing files with Puppet file (puppet:///module/file) fails


lucywyman:
## Describe the Bug
While some plan functions can specifically lookup files given their Puppet file path (such as `file::read()`), referencing `puppet:///mymodule/file` for files in a project in an apply block fails with:
```
Failed on localhost:
  Resources failed to apply for localhost
    File[/home/user/foo]: Could not retrieve information from environment production source(s) puppet:///bolt/myfile
```

## Expected Behavior
We should properly bundle module files when executing tasks so that module content, like files, are available from apply blocks.

## Steps to Reproduce
```
plan bolt (
  TargetSpec $targets
){
  apply($targets) {
    file { "/home/user/foo":
           source => "puppet:///modules/bolt/myfile"
    }
  }
}
```

```
bolt plan run bolt -t myhost
```

nicklewis:
This may be the culprit:

https://github.com/puppetlabs/bolt/blob/main/lib/bolt/applicator.rb#L296

We override the modulepath to avoid syncing builtin content, but we therefore also lose the project directory.

zbentley:
This is affecting me as well. The workaround is to modify any bolt-imported Puppet language modules so that their files are loaded via a `template` call, and move the files into from the `files` to the `templates` directory inside each module.

This is unfortunate, since third-party modules may also have issues, and because some of my files happen to have valid template interpolations in them, which breaks ERB/EPP.

zbentley:
Confirmed working, thanks!!

PR #2655, opened 2021-02-18
(GH-2504) Support Puppet file syntax for project files in apply blocks

beechtom:
This adds support for using Puppet file syntax
(`puppet://modules/<MODULE>/<FILE>`) in an apply block for files that are
in a Bolt project.

This also removes the `puppet_5_node` Docker container from spec tests,
since Bolt no longer supports Puppet 5.

!bug

* **Support Puppet file syntax for files in a Bolt project**
  ([#2504](#2504))

  Bolt now supports Puppet file syntax
  (`puppet:///modules/<MODULE>/<FILE>`) in apply blocks for files in a
  Bolt project. Previously, apply blocks would not compile if using this
  syntax for files in a Bolt project.

### DIFF
--- a/lib/bolt/pal.rb
+++ b/lib/bolt/pal.rb
@@ -215,6 +215,7 @@ module Bolt
     def with_bolt_executor(executor, inventory, pdb_client = nil, applicator = nil, &block)
       setup
       opts = {
+        bolt_project: @project,
         bolt_executor: executor,
         bolt_inventory: inventory,
         bolt_pdb_client: pdb_client,
--- a/spec/docker-compose.yml
+++ b/spec/docker-compose.yml
@@ -6,15 +6,6 @@ services:
     ports:
       - "20022:22"
 
-  puppet_5_node:
-    build:
-      context: .
-      args:
-        PUPPET_COLLECTION: puppet5
-    container_name: puppet_5_node
-    ports:
-      - "20023:22"
-
   puppet_6_node:
     build:
       context: .
--- a/.github/workflows/linux.yaml
+++ b/.github/workflows/linux.yaml
@@ -45,8 +45,8 @@ jobs:
           echo 'runner:runner' | sudo chpasswd
           sudo sh -c "echo 'Defaults authenticate' >> /etc/sudoers"
           sudo sh -c "echo 'runner  ALL=(ALL) PASSWD:ALL' >> /etc/sudoers"
-          docker-compose -f spec/docker-compose.yml build --parallel ubuntu_node puppet_5_node puppet_6_node puppet_7_node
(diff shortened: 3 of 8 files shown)


spec/docker-compose.yml
```diff
@@ -6,15 +6,6 @@ services:
...
       - "20022:22"
 
-  puppet_5_node:
-    build:
-      context: .
-      args:
-        PUPPET_COLLECTION: puppet5
-    container_name: puppet_5_node
-    ports:
-      - "20023:22"
```
lucywyman, line 16: I think in the future if anyone asks why we don't have `20023` we should say it's an unlucky number.

beechtom: And if they ask why, just respond "think about it". ;P

lib/bolt/pal.rb
```diff
@@ -215,6 +215,7 @@ def in_bolt_compiler
     def with_bolt_executor(executor, inventory, pdb_client = nil, applicator = nil, &block)
       setup
       opts = {
+        bolt_project: @project,
```
lucywyman, line 218: I think similar to line 176 here we'll want to call `detect_project_conflict` after overriding this.

lucywyman (approved):
Just a little thinking out loud here to make sure this doesn't have any unexpected side-effects:
We call `with_bolt_executor` in the following places:
- [x] `in_plan_compiler`, which immediately calls `in_bolt_compiler` which similarly overrides the project, so this should not introduce any behavior changes for that function or anything that calls it.
- [x] Same story with `in_task_compiler` - immediately calls `in_bolt_compiler`
- [x] In `BoltSpec::Run#apply_manifest`, which is correct so that BoltSpec works
- [x] In `cli.rb` when applying manifests using  `bolt apply`, which is also correct so that that behavior works 

To test this I verified that:
- [x] This works in apply blocks in plans loaded from the project itself
- [x] This works in manifests applied using `bolt apply`
- [x] Using the BoltSpec::Run#apply_manifest function to apply a manifest that relies on a project-level file passes.

lucywyman:
One minor improvement, otherwise LGTM!

lib/bolt/pal.rb
beechtom: It looks like `detect_project_conflict` will already be called whether using `bolt apply` or an apply block. I'm also not sure how to squeeze that function call into the block that's being passed to `Puppet.override` on line 237. :|

lucywyman: Oh I see - we call `in_plan_compiler` to call the `apply_prep` function in cli.rb, which raises the project conflict error :+1:

PR #2658, opened 2021-02-19
(maint) Remove testResults.xml

beechtom:
This removes a file that was erroneously added to the repo.

!no-release-note


lucywyman:
Woops! Good catch Tom.

PR #2663, opened 2021-02-23
(PE-31142) Serialize sensitive output from task Result

donoghuc:
This commit updates the `to_data` method for the Result class to serialize sensitive task output by calling the `to_s` value explictly on the sensitive output.

### DIFF
--- a/lib/bolt/result.rb
+++ b/lib/bolt/result.rb
@@ -203,12 +203,17 @@ module Bolt
     end
 
     def to_data
+      serialized_value = safe_value
+      if serialized_value.key?('_sensitive') &&
+         serialized_value['_sensitive'].is_a?(Puppet::Pops::Types::PSensitiveType::Sensitive)
+        serialized_value['_sensitive'] = serialized_value['_sensitive'].to_s
+      end
       {
         "target" => @target.name,
         "action" => action,
         "object" => object,
         "status" => status,
-        "value" => safe_value
+        "value" => serialized_value
       }
     end
 
--- a/spec/bolt/result_spec.rb
+++ b/spec/bolt/result_spec.rb
@@ -3,6 +3,7 @@
 require 'spec_helper'
 require 'json'
 require 'bolt'
+require 'bolt/target'
 require 'bolt/result'
 
 describe Bolt::Result do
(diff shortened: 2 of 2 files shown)


lib/bolt/result.rb
```diff
@@ -203,12 +203,14 @@ def safe_value
...
     def to_data
+      serialized_value = safe_value
+      serialized_value['_sensitive'] = serialized_value['_sensitive'].to_s if serialized_value.key?('_sensitive')
       {
         "target" => @target.name,
         "action" => action,
         "object" => object,
         "status" => status,
-        "value" => safe_value
+        "value" => serialized_value
```
nicklewis, line 213: I think the trouble is that this method is called by `to_json`, which we use for the JSON outputter as well, and we _don't_ want to print sensitive data in that case.

donoghuc: I'm not following. I dont see how this is different? Calling `to_s` on the `Sensitive` type would be `Sensitive [value redacted]` so wouldnt the json look the same in the outputter?

nicklewis: Ah I see, I completely ignored the lines above whoops :|

How do we get the original sensitive value when this task was run from a plan?

donoghuc: I believe we only use the `to_data` serialization when we https://github.com/puppetlabs/orchestrator/blob/7b26b219b47867c865aebeaa25da297876e901d2/ruby/bolt-plan-executor-lib/bolt_plan_executor/jruby_plan_executor_next.rb#L222 return from a top level plan (so any run_task actions or nested plan returns would have the native Result object).

nicklewis: I guess what I'm not sure about is how we wind up returning sensitive data from a task run via bolt-server in the first place. What's the path that data takes?

donoghuc: I think bolt-server gets a ResultSet object (with a single Result) and with the PR that was just merged makes sure the data is unwrapped so that orch can handle re-wrapping it or obfuscating it. https://github.com/puppetlabs/bolt/pull/2662/files#diff-9d9a096c76154ac144a49d1ae6a2f5d317ef1b644a1d444bc3eb76c85e81d12bR136

We only call the `to_data` method when we are done running a top level plan, so Result/ResultSets should not be serialized intermediately (and therefore will correctly maintain the non-serialized Sensitive type in plan context).

nicklewis (approved):

ekinanp:
Nit: This commit message is a bit confusing... Why do we want to serialize sensitive output from a task result? I am guessing it is to fix a bug that is there if we were not doing that, right? If so, it would help if the commit message could explain that bug and why this fixes it.

donoghuc:
Updated commit msg

beechtom (approved):

Issue #2659, opened 2021-02-22
Improve error messaging for "bolt script run" on Powershell


jpogran:
## Describe the Bug

When the PowerShell `ExecutionPolicy` is set to `Restricted` or `AllSigned`, Bolt won’t be able to execute an unsigned powershell script via bolt script run. 

## Expected Behavior

A clear generic error returned to the user or an error that explains the restricted execution policy that is preventing Bolt from running.

## Steps to Reproduce
1. Create a Windows target with the PowerShell Execution Policy set to `Restricted` or `AllSigned`
2. Execute a powershell script on target

## Environment
 - Version 3.0.0
 - Platform Windows

## Additional Context

Example output:

```
Started on winrm://wsus.dreamworx.nl...
--
Failed on winrm://wsus.dreamworx.nl:
The command failed with exit code 1
STDERR:
$invokeArgs = @{
: System.Management.Automation.ParameterBindingValidationException: Cannot validate argument on parameter 'ScriptBlock'. The argument is null. Provide a valid value for the argument, and then try running the command again. ---> System.Management.Automation.ValidationMetadataException: The argument is null. Provide a valid value for the argument, and then try running the command again.
at System.Management.Automation.ValidateNotNullAttribute.Validate(Object arguments, EngineIntrinsics engineIntrinsics)
at System.Management.Automation.ParameterBinderBase.BindParameter(CommandParameterInternal parameter, CompiledCommandParameter parameterMetadata, ParameterBindingFlags flags)
--- End of inner exception stack trace ---
at System.Management.Automation.ExceptionHandlingOps.CheckActionPreference(FunctionContext funcContext, Exception exception)
at System.Management.Automation.Interpreter.ActionCallInstruction`2.Run(InterpretedFrame frame)
at System.Management.Automation.Interpreter.EnterTryCatchFinallyInstruction.Run(InterpretedFrame frame)
at System.Management.Automation.Interpreter.EnterTryCatchFinallyInstruction.Run(InterpretedFrame frame)
+ CategoryInfo          : NotSpecified: (:) [Write-Error], WriteErrorException
+ FullyQualifiedErrorId : Microsoft.PowerShell.Commands.WriteErrorException
Failed on 1 target: winrm://wsus.dreamworx.nl
Ran on 1 target in 2.09 sec
```

Reported by : @kreeuwijk

jpogran:
Note that Bolt ships it's PowerShell module signed, so this problem occurs on the target not the host running Bolt.

[...]

PR #2664, opened 2021-02-23
(GH-2659 Improve error messages for "bolt script run" on PowerShell

jpogran:
This PR modifies the error printed when a target system has an execution policy of `Restricted` or `AllSigned`

When the target has a `Restricted` or `AllSigned` execution policy, it prevents running scripts and prevents pulling info from a script, resulting in an empty `ScriptBlock`. We weren't checking whether it was null or not, so the attempt blows up with a non helpful message. This changes the message to point to the root cause instead of a long stacktrace without any error information.

- [ ] Ports the exception handling and formatting method from bolt task to bolt script.
- [ ] Improve error messages for bolt script in PowerShell when target system has an execution policy of `Restricted` or `AllSigned`
 

> Note: This is very hard to test on localhost due to the way execution policies work and how they have to be edited. Best to test using the winrm connection

### DIFF
--- a/lib/bolt/shell/powershell/snippets.rb
+++ b/lib/bolt/shell/powershell/snippets.rb
@@ -55,6 +55,27 @@ module Bolt
             }
             #{build_arg_list}
 
+            switch -regex ( Get-ExecutionPolicy )
+            {
+              '^AllSigned'
+              {
+                if ((Get-AuthenticodeSignature -File "#{script_path}").Status -ne 'Valid') {
+                  $Host.UI.WriteErrorLine("Error: Target host Powershell ExecutionPolicy is set to ${_} and script '#{script_path}' does not contain a valid signature.")
+                  exit 1;
+                }
+              }
+              '^Restricted'
+              {
+                $Host.UI.WriteErrorLine("Error: Target host Powershell ExecutionPolicy is set to ${_} which denies running any scripts on the target.")
+                exit 1;
+              }
+            }
+
+            if([string]::IsNullOrEmpty($invokeArgs.ScriptBlock)){
+              $Host.UI.WriteErrorLine("Error: Failed to obtain scriptblock from '#{script_path}'. Running scripts may be disabled on this system. For more information, see about_Execution_Policies at https:/go.microsoft.com/fwlink/?LinkID=135170");
+              exit 1;
+            }
+
             try
             {
               Invoke-Command @invokeArgs
--- a/lib/bolt/shell/powershell/snippets.rb
+++ b/lib/bolt/shell/powershell/snippets.rb
@@ -61,8 +61,8 @@ module Bolt
             }
             catch
             {
-              Write-Error $_.Exception
-              exit 1
+              $Host.UI.WriteErrorLine("[$($_.FullyQualifiedErrorId)] Exception $($_.InvocationInfo.PositionMessage).`n$($_.Exception.Message)");
+              exit 1;
             }
             PS
           end
@@ -144,7 +144,7 @@ module Bolt
               [Parameter(Mandatory = $true)] $Text,
               [Parameter(Mandatory = $false)] [Text.Encoding] $Encoding = [Text.Encoding]::UTF8
             )
-          
+
             $Text | ConvertFrom-Json | ConvertFrom-PSCustomObject
             }
             PS
--- a/lib/bolt/shell/powershell/snippets.rb
+++ b/lib/bolt/shell/powershell/snippets.rb
@@ -72,7 +72,7 @@ module Bolt
             }
 
             if([string]::IsNullOrEmpty($invokeArgs.ScriptBlock)){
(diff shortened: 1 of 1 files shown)


lib/bolt/shell/powershell/snippets.rb
```diff
@@ -55,14 +55,19 @@ def run_script(arguments, script_path)
             }
             #{build_arg_list}
 
+            if([string]::IsNullOrEmpty($invokeArgs.ScriptBlock)){
+              $Host.UI.WriteErrorLine("Error: Failed to obtain scriptblock from '#{script_path}'. Running scripts may be disabled on this system. For more information, see about_Execution_Policies at https:/go.microsoft.com/fwlink/?LinkID=135170");
+              exit 1;
+            }
```
jpogran, line 61: @puppetlabs/bolt I need help with the error message here.

This checks if `$invokeArgs.ScriptBlock` is null or empty and tells the user the likely cause is execution policy. However there could be other reasons, but I won't be able to tell why exactly. Is there a better way to phrase this?

jpogran:
@kreeuwijk if you could try this out and LMK if the error is easier to understand I'd appreciate it

lib/bolt/shell/powershell/snippets.rb
adreyer: You could link to a page in our docs with a more in depth explanation. These are always hard problems.
would "The system may have an execution policy that prevents running unsigned scripts"  be slightly more informative?

kreeuwijk:
I've added more specific validation and handling, see the output of test runs under different policies:
*AllSigned*
```
  STDERR:
    Error: Powershell ExecutionPolicy is set to AllSigned and script 'C:\Users\adm_winrm.DREAMWORX\AppData\Local\Temp\zlbptgy2.n3j\myscript.ps1' does not contain a valid signature.
Failed on 1 target: winrm://wsus.dreamworx.nl
```
*Restricted*
```
  STDERR:
    Error: Powershell ExecutionPolicy is set to Restricted which denies running any scripts.
Failed on 1 target: winrm://wsus.dreamworx.nl
```
Also added `File-Unblock` when the policy is `RemoteSigned` or `Unrestricted`, ensuring the local script is allowed to run. As we clean up the file afterwards, this should not pose a security threat.

jpogran:
Thanks for the additional work. I agree with the execution policy detection approach, that adds specific actionable info before the exception happens, but still leaves the 'catch' to deal with unknown errors.

I am not sold on the `Unblock-File` addition however. `Unblock-File` only removes the [Zone3 Identifier](https://docs.microsoft.com/en-us/powershell/module/Microsoft.PowerShell.Utility/Unblock-File), which gets added when files are downloaded. Our files are uploaded using the WinRm bolt transport, which write byte streams. The local transport use Ruby methods to create the file and write a byte stream to it. Have you run into a case using the either the local or WinRM transport where the target file has the alternate data streams written to it? I'm not concerned so much about security of the file, more so about adding an additional command to run that doesn't affect anything

kreeuwijk:
I wasn't sure how Bolt uploaded the file, that's why I added it. If we can be certain Bolt never sets the Zone3 identifier on files it creates, we can remove `'^(RemoteSigned|Unrestricted)'` case altogether.

lib/bolt/shell/powershell/snippets.rb
lucywyman: I think as long as you include `may be` (like you do here) this is fine. We do this a few other places: lib/bolt/shell/bash.rb#L448, lib/bolt/transport/ssh/connection.rb#L233.

lib/bolt/shell/powershell/snippets.rb
```diff
@@ -55,14 +55,35 @@ def run_script(arguments, script_path)
...
+              }
+              '^Restricted'
+              {
+                $Host.UI.WriteErrorLine("Error: Target host Powershell ExecutionPolicy is set to ${_} which denies running any scripts on the target.")
+                exit 1;
+              }
+            }
+
+            if([string]::IsNullOrEmpty($invokeArgs.ScriptBlock)){
+              $Host.UI.WriteErrorLine("Error: Failed to obtain scriptblock from '#{script_path}'. Running scripts may be disabled on this system. For more information, see about_Execution_Policies at https:/go.microsoft.com/fwlink/?LinkID=135170");
```
beechtom, line 75:
```suggestion
              $Host.UI.WriteErrorLine("Error: Failed to obtain scriptblock from '#{script_path}'. Running scripts might be disabled on this system. For more information, see about_Execution_Policies at https:/go.microsoft.com/fwlink/?LinkID=135170");
```

Docs team prefers the use of 'might' since it's less ambiguous for non-native speakers.

lib/bolt/shell/powershell/snippets.rb
jpogran: I think @kreeuwijk addition addresses @adreyer  suggestion nicely, and @lucywyman suggestion goes in line with the edit @beechtom provided

beechtom (approved):

lucywyman (approved):

Issue #2595, opened 2021-01-28
Unpin PuppetDB container and diagnose why it's exiting early


lucywyman:
## Describe the Bug
The 7.0.1 tag of the PuppetDB container does not seem to stay alive as 6.x tags did. The container itself builds fine from the `docker-compose.yml`, but then exits before tests are able to use it. Containers typically have one process or service running, and once that process or service exits the container does as well. It's unclear what changed between tags. This is to investigate what's happening and unpin the container.

I tried adding `CMD ['foreground']` which should keep the container itself running, but PuppetDB was still unreachable, which to me indicates that the service itself is shutting down. It sounds like some configuration may have changed, so maybe there's an error there?

PR #2665, opened 2021-02-23
(GH-2595) Unpin PDB container in CI

beechtom:
This unpins the puppetdb container in CI. Previously, the 7.x series of
the puppetdb container would immediately close due to an error raised
by a missing certfile when starting puppetdb.

!no-release-note

### DIFF
--- a/spec/Dockerfile.puppetdb
+++ b/spec/Dockerfile.puppetdb
@@ -1,7 +1,8 @@
-FROM puppet/puppetdb:6.13.1
+FROM puppet/puppetdb
 
 # Use our own certs so this doesn't have to wait for puppetserver startup
 COPY fixtures/ssl/ca.pem /opt/puppetlabs/server/data/puppetdb/certs/certs/ca.pem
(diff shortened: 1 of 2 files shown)


spec/Dockerfile.puppetdb
```diff
@@ -1,7 +1,8 @@
-FROM puppet/puppetdb:6.13.1
+FROM puppet/puppetdb
 
 # Use our own certs so this doesn't have to wait for puppetserver startup
 COPY fixtures/ssl/ca.pem /opt/puppetlabs/server/data/puppetdb/certs/certs/ca.pem
-COPY fixtures/ssl/cert.pem /opt/puppetlabs/server/data/puppetdb/certs/certs/pdb.pem
+COPY fixtures/ssl/cert.pem /opt/puppetlabs/server/data/puppetdb/certs/certs/server.crt
```
lucywyman, line 5: Do these fixtures need to be renamed to avoid confusion?

lucywyman (approved):

spec/Dockerfile.puppetdb
beechtom: It looks like they're reused in a couple locations, so I don't think there's a need to.

PR #2670, opened 2021-02-25
(PE-31182)(PE-31192) Improvements to puppet_connect::test_input_data

ekinanp:
This PR adds the following to puppet_connect::test_input_data:

* It maintains transport config parity with bolt-server
* It validates that the SSH private-key does not point to a filepath

### DIFF
--- a/modules/puppet_connect/plans/test_input_data.pp
+++ b/modules/puppet_connect/plans/test_input_data.pp
@@ -16,6 +16,32 @@ plan puppet_connect::test_input_data(TargetSpec $targets = 'all') {
   $targs.each |$target| {
     case $target.transport {
       'ssh': {
+        $private_key_config = dig($target.config, 'ssh', 'private-key')
+        if $private_key_config =~ String {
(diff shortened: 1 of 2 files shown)


donoghuc (approved):

modules/puppet_connect/plans/test_input_data.pp
```diff
@@ -14,14 +14,50 @@
...
+            Next is the corresponding entry in the input data file:
+              ...
+              ssh_private_key:
+                key-data:
+                  <private_key_contents>
+              ...
+            | END
+
+          out::message($msg)
+          fail_plan("The SSH private key of the ${$target.name} target points to a filepath on disk")
```
nicklewis, line 42: Some places use `${target}` and some use `${$target.name}`. I think `${target}` is fine everywhere. That's equivalent to `${target.safe_name}`.

ekinanp: Thanks, updated.

Issue #2672, opened 2021-02-26
@param without description causes `bolt plan show` to fail


op-ct:
## Describe the Bug

Running `bolt plan show` when any plan in the modulepath contains a Puppet Strings `@param` comment with an empty description causes Bolt to fail (without indicating what went wrong or which plan was responsible):

```console
$ bolt plan show
Traceback (most recent call last):
        12: from /opt/puppetlabs/bolt/bin/bolt:23:in `<main>'
        11: from /opt/puppetlabs/bolt/bin/bolt:23:in `load'
        10: from /opt/puppetlabs/bolt/lib/ruby/gems/2.7.0/gems/bolt-3.0.0/exe/bolt:11:in `<top (required)>'
         9: from /opt/puppetlabs/bolt/lib/ruby/gems/2.7.0/gems/bolt-3.0.0/lib/bolt/cli.rb:416:in `execute'
         8: from /opt/puppetlabs/bolt/lib/ruby/gems/2.7.0/gems/bolt-3.0.0/lib/bolt/cli.rb:573:in `list_plans'
         7: from /opt/puppetlabs/bolt/lib/ruby/gems/2.7.0/gems/bolt-3.0.0/lib/bolt/pal.rb:373:in `list_plans_with_cache'
         6: from /opt/puppetlabs/bolt/lib/ruby/gems/2.7.0/gems/bolt-3.0.0/lib/bolt/pal.rb:373:in `each_with_object'
         5: from /opt/puppetlabs/bolt/lib/ruby/gems/2.7.0/gems/bolt-3.0.0/lib/bolt/pal.rb:373:in `each'
         4: from /opt/puppetlabs/bolt/lib/ruby/gems/2.7.0/gems/bolt-3.0.0/lib/bolt/pal.rb:374:in `block in list_plans_with_cache'
         3: from /opt/puppetlabs/bolt/lib/ruby/gems/2.7.0/gems/bolt-3.0.0/lib/bolt/pal.rb:442:in `get_plan_info'
         2: from /opt/puppetlabs/bolt/lib/ruby/gems/2.7.0/gems/bolt-3.0.0/lib/bolt/pal.rb:442:in `each_with_object'
         1: from /opt/puppetlabs/bolt/lib/ruby/gems/2.7.0/gems/bolt-3.0.0/lib/bolt/pal.rb:442:in `each'
[...]

PR #2676, opened 2021-02-26
(GH-2672) Handle plan parameter tags without descriptions

beechtom:
This fixes a bug where a plan with a Puppet strings `@param` tag that
did not include a description would cause a plan to error.

!bug

* **Handle plan parameter tags without descriptions**
  ([#2672](https://github.com/puppetlabs/bolt/issues/2672))

  Bolt no longer errors if a plan includes a Puppet strings `@param` tag
  that does not have a description.


lucywyman (approved):

lucywyman:
Nice - fixes the issue for Puppet plans, and parameter parsing seems to be otherwise pretty robust. I tried `@params` without the name of the parameter and that was handled fine, as well as misspelled `@para targets foo` which also correctly ignores the param.

PR #2678, opened 2021-02-26
(maint) Add plan descriptions to 'bolt plan show' output

beechtom:
This adds plan descriptions to `bolt plan show` and `Get-BoltPlan`
output.

!feature

* **Show plan descriptions in plan list**

  Plan descriptions now appear in `bolt plan show` and `Get-BoltPlan`
  output.

### DIFF
--- a/lib/bolt/outputter/human.rb
+++ b/lib/bolt/outputter/human.rb
@@ -53,8 +53,19 @@ module Bolt
         string.sub(/\s\z/, '')
       end
 
+      # Wraps a string to the specified width. Lines only wrap
+      # at whitespace.
+      #
       def wrap(string, width = 80)
+        return string unless string.is_a?(String)
         string.gsub(/(.{1,#{width}})(\s+|\Z)/, "\\1\n")
+      end
+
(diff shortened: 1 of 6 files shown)


beechtom:
This adds the plan description in JSON output as well. I'm not sure if that would be considered breaking or not, but it's easy enough to remove if we're concerned about it.

`[['plan1'],['plan2']]` => `[['plan1', 'desc'],['plan2','desc']]`

lucywyman:
Hm - I don't think that's breaking because this is purely additive, and users should already be accessing the plan name by index.

Tests:
- [x] Does not print a description if one isn't provided
- [x] Does not print a description if it's an empty string
- [x] Does not wrap descriptions less than 80 characters
- [x] Wraps descriptions more than 80 characters
- [x] Prints 'no available tasks' if no tasks are available
- [x] Prints 'no available plans' if no plans are available
- [x] Does not apply formatting to descriptions in JSON
- [x] Includes descriptions in JSON output

lucywyman (approved):

PR #2682, opened 2021-03-01
(maint) Test YAML plans with BoltSpec

beechtom:
This enables testing YAML plans with BoltSpec by adding a
`report_yaml_plan` method to the mock executor.

!bug

* **Test YAML plans with BoltSpec**

  YAML plans can now be tested with BoltSpec.


lucywyman (approved):
Nice! Verified this fix locally. It looks like there's a few other functions that we don't support and should probably add :grimacing:

PR #2683, opened 2021-03-02
(maint) Prefer 'might' to 'may' where appropriate

lucywyman:
We learned recently that 'might' is preferable to 'may' when expressing
possibility or making suggestions. This replaces any use of 'may' that
can be replaced with 'might' in messages printed by Bolt and the
documentation. It ignores cases where 'may' refers to ability (ie. "You
may include lowercase letters"), and comments in the Bolt codebase.

!no-release-note


hestonhoffman:
This is awesome, thanks Lucy :) 
A note on "may" to describe the ability to do something. Can I suggest using "can" instead? From the Puppet style guide:
> May: avoid using because the meaning can be unclear, particularly for non-native English speakers. Instead, use "can" to convey that the user is able to do something. Use "might" instead of "may" to convey that something might or might not happen or that the user might or might not be able to do something.

beechtom (approved):

hestonhoffman (approved):
🎉

Issue #2651, opened 2021-02-18
Support powershell parameters when passing a script. 


adreyer:
## Use Case
Powershell has a standard method of passing parameters to a script. Bolt should expose this to the run_script function and script step.

## Describe the Solution You Would Like
Bolt's run_script should accept a hash of `pwsh_params` and pass those parameters to the script. 

`run_script(my_script.ps, $targets, pwsh_params => { 'this' => 1, 'that' => argument} )`

Bolt should support accepting arguments along with params.

## Questions

- Should we support typed or structured data? 
  Not at this point string only for now.
- Should we support passing powershell params from the `bolt script run` cli?
  This is less important because you don't have filter out parameters.
- Does it make escaping less awful?
  Not at this point but we should look for how we can leverage this after it's merged in.

PR #2685, opened 2021-03-02
(GH-2651) Add pwsh_params option to run_script function

beechtom:
This adds a `pwsh_params` option to the `run_script` plan function.
The option accepts a hash of named parameters to pass to a PowerShell
script, where each key is the name of the parameter. If the `run_script`
function is called with both the `arguments` and `pwsh_params` options,
Bolt will error and provide a helpful message that both cannot be
specified.

This also adds a `pwsh_params` key to the YAML plan script step.

!feature

* **Add `pwsh_params` option to `run_script` plan function**
  ([#2651](https://github.com/puppetlabs/bolt/issues/2651))

  The `run_script` plan function now accepts a `pwsh_params` option
  which can be used to pass named parameters to a PowerShell script.


documentation/writing_yaml_plans.md
```diff
@@ -189,7 +189,9 @@ Script steps use these fields:
 -   `script`: The script to run
 -   `targets`: A target or list of targets to run the script on
 -   `arguments`: (Optional) An array of command-line arguments to pass to the
-    script
+    script. Mutually exclusive with `pwsh_params`.
```
lucywyman, line 192: I'll let Heston rule on this, but I'm not sure everyone will know what 'mutually exclusive' means? "Can't be used with" might be friendlier.

lucywyman, line 192: I'll let Heston rule on this, but I'm not sure everyone will know what 'mutually exclusive' means? "Can't be used with" might be friendlier.

documentation/templates/plan_functions.md.erb
```diff
@@ -41,7 +40,7 @@ This function<%= func['signatures'].count == 1 ? '' : ' signature' %> accepts th
 | --- | --- | --- |
 <% sig['options'].each do |name, data| -%>
 | `<%= name %>` | `<%= data['type'] %>` | <%= data['desc'] %> |
-<% end -%>
+<% end %>
```
lucywyman, line 43: Er, doesn't this need the `-%>`?

lib/bolt/shell/powershell/snippets.rb
```diff
@@ -48,6 +48,7 @@ def run_script(arguments, script_path)
             build_arg_list = arguments.map do |a|
               "$invokeArgs.ArgumentList += @'\n#{a}\n'@"
             end.join("\n")
+
```
lucywyman, line 51: This seems incidental - would you be ok resetting just this file to keep the Git history more navigable?

documentation/templates/plan_functions.md.erb
beechtom: Not really. The page will still render fine on the docs site, but this fixes an issue where the markdown wasn't displaying correctly because there wasn't a blank line between the end of the table and the next line of text.

documentation/writing_yaml_plans.md
hestonhoffman: Yeah, I agree. I like your replacement Lucy. We use this term quite a bit in the Puppet docs. Might be a good one for a future docs pothole :)

lucywyman:
Not sure if this isn't updating just because I'm on plane wifi, but I think this needs a rebase.

lib/bolt/pal/yaml_plan/step/script.rb
```diff
@@ -20,8 +20,12 @@ def self.required_keys
           # Returns an array of arguments to pass to the step's function call
           #
           private def format_args(body)
+            args        = body['arguments'] || []
+            pwsh_params = body['pwsh_params'] || {}
+
             opts = format_options(body)
-            opts = opts.merge('arguments' => body['arguments'] || []) if body.key?('arguments')
+            opts = opts.merge('arguments' => args) if args.any?
```
lucywyman, line 27: For both of these we should check that the user passed in a hash before calling `empty?` and raise an error if the value isn't a hash (I'm not sure if this gets called before or after we evaluate Puppet code, but using variables should also work and error if the variable evaluates to a non-hash).

lucywyman (commented):
- [x] Accepts `pwsh_params` in YAML plan language
- [x] `pwsh_params` is mutually exclusive with `arguments`
  - [x] In Puppet plan language
  - [x] With a description
  - [x] In YAML plan language
  - [x] With a description
- [x] Coerces Puppet types to PowerShell types
  - [x] [PS converts strings to more specific datatypes](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_functions_advanced_parameters?view=powershell-7.1#type-conversion-of-parameter-values) Test this with timespans and regexes.
  - [x] String
  - [x] Integer
  - [x] Float
  - [x] Boolean
  - [x] Hash
  - [x] Mixed array
- [x] Errors if `pwsh_params` is not a hash
  - [x] Puppet plan: kind of?:
    ```
    ...
    Failed on localhost:
      STDERR:
        [MissingMandatoryParameter,2lkdm4yz.upu\complex_params.ps1] Exception At line:61 char:3
        +   & "/tmp/2lkdm4yz.upu\complex_params.ps1" @taskArgs
        +   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~.
        Cannot process command because of one or more missing mandatory parameters: ArgString.
    ``` 
  - [x] YAML plan - see comment

Overall looking great! It's very handy that PowerShell just takes care of converting stringified parameters to "odd" datatypes.

beechtom:
Updated to error if `arguments` or `pwsh_params` is the wrong type in both Puppet and YAML plans.

hestonhoffman (approved):
LGTM!

lucywyman (approved):

PR #2686, opened 2021-03-02
(maint) Improve plan and task info output

beechtom:
This changes the format of plan and task info output for the human and
rainbow formatter to be more organized and easier to read.

!feature

* **Improved plan and task info output**

  Plan and task info output shown with `bolt plan|task show` and
  `Get-BoltPlan|Task` has been reformatted to be more organized.

### DIFF
--- a/lib/bolt/outputter/human.rb
+++ b/lib/bolt/outputter/human.rb
@@ -229,7 +229,7 @@ module Bolt
         @stream.puts total_msg
       end
 
-      def format_table(results, padding_left = 0, padding_right = 3)
+      def format_table(results, padding_left = 0, padding_right = 3, separators: false)
         # lazy-load expensive gem code
         require 'terminal-table'
 
@@ -242,7 +242,8 @@ module Bolt
             padding_left: padding_left,
             padding_right: padding_right,
             border_top: false,
-            border_bottom: false
+            border_bottom: false,
+            all_separators: separators
           }
         )
       end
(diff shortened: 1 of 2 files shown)


spec/bolt/outputter/human_spec.rb
```diff
@@ -130,65 +130,10 @@
...
-      PARAMETERS:
-      - icing: Cream cheese
-          Rich, tangy, sweet
-
-      MODULE:
-      /path/to/cinnamony/goodness
-    TASK_OUTPUT
-  end
-
-  it 'converts Data (undef) to Any' do
```
beechtom, line 149: I'm not really sure this test was doing what it claimed to be doing... unless I'm misunderstanding it?

lib/bolt/outputter/human.rb
```diff
@@ -269,78 +270,104 @@ def print_tasks(tasks, modulepath)
...
-                     " [#{k}=<value>]"
-                   else
-                     " #{k}=<value>"
-                   end
-        end
+        module_path = task.files.first['path'].chomp("/tasks/#{task.files.first['name']}")
+        module_path = 'built-in module' if module_path.start_with?(Bolt::Config::Modulepath::MODULES_PATH)
 
         if task.supports_noop
           usage << Bolt::Util.powershell? ? '[-Noop]' : '[--noop]'
```
lucywyman, line 285: Sorry I know this isn't part of your changes, but I'm getting a stacktrace here (with Ruby 2.6.3) that I verified in IRB, where Ruby does not like this syntax:
```
/opt/puppetlabs/bolt/lib/ruby/gems/2.7.0/gems/bolt-3.1.0/lib/bolt/outputter/human.rb:285:in `print_task_info': no implicit conversion of false into String (TypeError)
```
It looks like saving that conditional output to a var and then append it, it works. This is also failing in current Bolt, definitely not related to your changes - it just seems worth fixing here.

lucywyman: This is my task metadata:
```
{
  "description": "Task with noop",
  "supports_noop": true
}
```

lib/bolt/outputter/human.rb
```diff
@@ -269,78 +270,104 @@ def print_tasks(tasks, modulepath)
...
-        path = task.files.first['path'].chomp("/tasks/#{task.files.first['name']}")
-        task_info << if path.start_with?(Bolt::Config::Modulepath::MODULES_PATH)
-                       "built-in module"
-                     else
-                       path
-                     end
-        @stream.puts(task_info)
+          description = +""
+          description << wrap(info['description']) if info['description']
+          description << "\n#{type}"
```
lucywyman, line 293: It seems like the descriptions already have a newline, so this results in 2 newlines which makes the output pretty big. I think we don't need another newline here (or can do something like  chomp the description and then add the newline)

lucywyman: Hm, I think my terminal may be adding extra space as well, since this looks more like 3 lines to me than 2...
```
  retry               The number of retries in case of network connectivity failures                                                         
                                                                                                                                             
                      Optional[Integer]                                                                                                      
                      Default: 5
```
![task_show_output](https://user-images.githubusercontent.com/5055819/110021135-3a253400-7cdf-11eb-861e-596ece6310a8.png)

spec/bolt/outputter/human_spec.rb
lucywyman: This was added when we first added the `task show <task>` subcommand to ensure the formatting was correct - I think that replacing the heredoc string with the new string makes sense, to make sure format doesn't drift, rather than checking individual regexes (for example, adding unnecessary newlines to output wouldn't be caught the new test but would be caught with the old one).

lucywyman: Ack sorry, comment applies to the bit *above*. *That* test does seem useless now :sweat_smile: https://github.com/puppetlabs/bolt/commit/e3c61a2a64440ebdcd2dfbc6368d2c730d0cc505 was the original, but it looks like we don't coerce the datatype anymore.

lucywyman: Ok, *actually* actually it seems like that functionality should be there and has gotten lost - I think the test was updated erroneously, when it was really catching a bug. I'll make a ticket and we can discuss - I think you can remove the test for now.

lib/bolt/outputter/human.rb
beechtom: Nice catch! 🙌

lib/bolt/outputter/human.rb
beechtom: @lucywyman Updated to format a string and print to the stream once. Hopefully that fixes the spacing issues. 🤞

lucywyman:
Ok, I've made some progress on what's happening. Terminal Table is using `Unicode::DisplayWidth.of(@value)` to determine the width the cell needs to be, and for ~~ some reason ~~ that is returning the full length of the string in Ruby including newline characters, instead of the max-width of any line without newline characters. For example, `"The version of puppet-agent to install (defaults to latest when no agent is\ninstalled)\n\nOptional[String]"` has a width of 107 according to Displaywidth, as opposed to a width of 80. You actually *can* set the width of a cell as its style, but that does not seem to affect the output at all.

This matters because when the cell is 116 characters long it thinks the whole cell needs to be 116 characters (plus 20 or so characters for the other column), which is more than the screen width so wraps, which is what causes it to look like new-newlines.

I think this hasn't been an issue previously since the data we print tends to be "single file" lists or lists that get wrapped. I honestly feel like I've sunk enough time into this already that I'd kind of like to table it for now (haha) since I'm fairly sure this will be an issue for a significant number of users.

PR #2687, opened 2021-03-03
(PE-31273) Collect analytics on source files for plan actions

donoghuc:
Collect analytics on whether absolute paths are uses as the source file for run_script, upload_file, or file::read.

!no-relase-note

### DIFF
--- a/bolt-modules/boltlib/lib/puppet/functions/run_script.rb
+++ b/bolt-modules/boltlib/lib/puppet/functions/run_script.rb
@@ -80,7 +80,7 @@ Puppet::Functions.create_function(:run_script, Puppet::Functions::InternalFuncti
         Puppet::Pops::Issues::NOT_A_FILE, file: script
       )
     end
-
+    executor.report_file_source(self.class.name, script)
     # Ensure that given targets are all Target instances)
     targets = inventory.get_targets(targets)
 
--- a/documentation/analytics.md
+++ b/documentation/analytics.md
@@ -38,6 +38,8 @@ with a randomly generated, non-identifiable user UUID:
 - Which built-in plugins Bolt is using
 - Topics viewed with `bolt guide <TOPIC>`
 - IDs of any deprecation warnings
+- Whether source file for `upload_file`, `run_script`, or `file::read` plan
+  functions use an absolute path or a module path.
 
 ## Viewing collected data
 
--- a/documentation/analytics.md
+++ b/documentation/analytics.md
@@ -38,8 +38,8 @@ with a randomly generated, non-identifiable user UUID:
 - Which built-in plugins Bolt is using
 - Topics viewed with `bolt guide <TOPIC>`
 - IDs of any deprecation warnings
-- Whether source file for `upload_file`, `run_script`, or `file::read` plan
-  functions use an absolute path or a module path.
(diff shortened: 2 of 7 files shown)


adreyer:
We need to add this to the list of data we collect: https://github.com/puppetlabs/bolt/blob/main/documentation/analytics.md

lib/bolt/executor.rb
```diff
@@ -232,6 +233,11 @@ def report_bundled_content(mode, name)
       @analytics.report_bundled_content(mode, name)
     end
 
+    def report_file_source(plan_function, source)
+      label = Pathname.new(source).absolute? ? 'absolute' : 'module'
+      @analytics&.event('Plan', plan_function, label: label)
+    end
```
beechtom, line 239: BoltSpec's mock executor will need this method as well.

beechtom (commented):
We should add a line that we collect this data to https://github.com/puppetlabs/bolt/blob/main/documentation/analytics.md#data-collected

bolt-modules/boltlib/lib/puppet/functions/run_script.rb
```diff
@@ -80,7 +80,7 @@ def run_script_with_description(scope, script, targets, description = nil, optio
         Puppet::Pops::Issues::NOT_A_FILE, file: script
       )
     end
-
+    executor.report_file_source(self.class.name, script)
```
donoghuc, line 83: In general I chose to report once the file has actually been found (so paths that do not exist, possibly due to user error) will not be reported. Is this the correct behavior?

documentation/analytics.md
```diff
@@ -38,6 +38,8 @@ with a randomly generated, non-identifiable user UUID:
 - Which built-in plugins Bolt is using
 - Topics viewed with `bolt guide <TOPIC>`
 - IDs of any deprecation warnings
+- Whether source file for `upload_file`, `run_script`, or `file_read` plan
```
beechtom, line 41:
```suggestion
- Whether source file for `upload_file`, `run_script`, or `file::read` plan
```

beechtom (approved):

lib/bolt/executor.rb
```diff
@@ -232,6 +233,11 @@ def report_bundled_content(mode, name)
       @analytics.report_bundled_content(mode, name)
     end
 
+    def report_file_source(plan_function, source)
+      label = Pathname.new(source).absolute? ? 'absolute' : 'module'
+      @analytics&.event('Plan', plan_function, label: label)
```
nicklewis, line 238: This is a little weird because it means we'll get both "Plan / call_function / file::read" and "Plan / file::read / absolute". That namespacing seems a bit off, but I'm not sure there's a better option...

documentation/analytics.md
```diff
@@ -38,6 +38,8 @@ with a randomly generated, non-identifiable user UUID:
 - Which built-in plugins Bolt is using
 - Topics viewed with `bolt guide <TOPIC>`
 - IDs of any deprecation warnings
+- Whether source file for `upload_file`, `run_script`, or `file::read` plan
+  functions use an absolute path or a module path.
```
hestonhoffman, line 42:
```suggestion
- Whether the source file for `upload_file`, `run_script`, or `file::read` plan
  functions uses an absolute path or a module path.
```

Issue #2652, opened 2021-02-18
Standard script path reference for `bolt script run` and `run_script`


adreyer:
## Use Case
currently the CLI uses a relative or absolute path to the script while `run_script` uses a puppet file reference format. From the top level of the project this means `files/myscript.sh` vs `my_project/myscript.sh` this is confusing and annoying.

There should be a single method of referring to a script located in a project or module that works directly from the commandline as well as from a plan. 

## Describe the Solution You Would Like
The best solution is probably to fall back to `module path` based resolution when a scripts path doesn't resolve. In a future release we can prefer module path when the first part of the scripts path matches the name of an installed module. Error messages should be clear about what paths failed to find the module.

This should use bolt's puppetless module classes to avoid a performance hit. those classes may need to be updated to support the project module.

## Questions
How can we limit the performance hit for using puppet paths?
Is there a better long term option than using module path based resolution? Perhaps we should have more magic about file level content but we need usertesting to justify that inconsistency.

PR #2688, opened 2021-03-04
(GH-2652) Support Puppet paths in 'bolt script run'

beechtom:
This adds support for using Puppet paths (`<MODULE NAME>/<FILE NAME>`)
in `bolt script run` and `Invoke-BoltScript`. The default behavior is to
still locate the script file using an absolute/relative path, but Bolt
will now fall back to checking if the path is a Puppet path if that
fails.

!feature

* **Support Puppet paths when running scripts from the CLI**
  ([#2652](https://github.com/puppetlabs/bolt/issues/2652))

  You can now use Puppet paths (`<MODULE NAME>/<FILE NAME>`) to specify
  the path to a script when running `bolt script run` or
  `Invoke-BoltScript`.

### DIFF
--- a/lib/bolt/cli.rb
+++ b/lib/bolt/cli.rb
@@ -249,11 +249,14 @@ module Bolt
         end
       end
 
-      if %w[task plan].include?(options[:subcommand]) && options[:action] == 'run'
+      if %w[task plan script].include?(options[:subcommand]) && options[:action] == 'run'
         if options[:object].nil?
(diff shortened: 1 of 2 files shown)


lucywyman (approved):
Nice!
- [x] Finds files from the builtin modulepath
- [x] Finds files from the user-defined modulepath
- [x] Errors if the script isn't in the module (with a very fancy path even!)
- [x] Errors if the script is in the module but isn't under `files/`
- [x] Errors if the script is unreadable (ie. owned by another user with 600 permissions)

PR #2696, opened 2021-03-09
(maint) Don't include password for Windows target

lucywyman:
The Windows testing password includes characters that aren't allowed in
target uris that are looked up in the inventory file. When we use the
full target uri (including password) from the `BoltSpec::Conn` module
along with a project created by `BoltSpec::Project#with_project` which
laid down an empty inventoryfile, Bolt would try to lookup the uri in
the empty inventory and fail.

This makes two changes to address this:
- First, it modifies the failing test to use `--password` instead of
  including the password in the URI.
- Second, it modifies `with_project` to only create an inventoryfile if
  any inventory data is passed to the function. This will prevent Bolt
  from trying to lookup targets in the inventory unnecessarily.

!no-release-note

### DIFF
--- a/spec/lib/bolt_spec/project.rb
+++ b/spec/lib/bolt_spec/project.rb
@@ -18,12 +18,11 @@ module BoltSpec
         # Set default config and inventory if not provided
         cfg = { 'name' => name }
         cfg.merge!(config) if config
-        inventory ||= {}
 
         # Create project directory and files
         FileUtils.mkdir_p(project_path)
         File.write(config_path, cfg.to_yaml)
-        File.write(inventory_path, inventory.to_yaml)
+        File.write(inventory_path, inventory.to_yaml) unless inventory.nil?
 
         yield(project_path)
       end
--- a/spec/integration/logging_spec.rb
+++ b/spec/integration/logging_spec.rb
@@ -179,19 +179,19 @@ shared_examples 'streaming output' do
     end
 
     it 'streams stdout to the console' do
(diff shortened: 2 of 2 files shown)


spec/lib/bolt_spec/project.rb
```diff
@@ -13,17 +13,16 @@ def with_project_directory(name = 'project', config: nil, inventory: nil, boltdi
         tmpdir         = Pathname.new(tmpdir).expand_path
         project_path   = boltdir ? (tmpdir + name + 'Boltdir') : (tmpdir + name)
         config_path    = project_path + 'bolt-project.yaml'
-        inventory_path = project_path + 'inventory.yaml'
+        inventory_path = project_path + 'inventory.yaml' unless inventory.nil?
```
beechtom, line 16: This change isn't necessary.

spec/lib/bolt_spec/project.rb
```diff
@@ -13,17 +13,16 @@ def with_project_directory(name = 'project', config: nil, inventory: nil, boltdi
...
         # Set default config and inventory if not provided
         cfg = { 'name' => name }
         cfg.merge!(config) if config
-        inventory ||= {}
 
         # Create project directory and files
         FileUtils.mkdir_p(project_path)
         File.write(config_path, cfg.to_yaml)
-        File.write(inventory_path, inventory.to_yaml)
+        File.write(inventory_path, inventory.to_yaml) unless inventory.nil?
```
beechtom, line 25: 🙌

PR #2697, opened 2021-03-10
(bug) Fix 'Linux fast' workflow to run tests

beechtom:
This fixes a bug in the linux spec tests that was preventing any tests
from running in GitHub Actions. The issue specifically came about from
using the `:lxd` tag to mark LXD tests. Updating the tag to
`:lxd_transport` fixes the issue.

!no-release-note


lucywyman:
Ugh - I'd really like to understand why this caused a problem. This makes absolutely no sense.

Issue #2679, opened 2021-03-01
Support `run_as` user when downloading files


lucywyman:
## Describe the Bug
When users specify a `run_as` user when using a supporting when using the `download_files` plan function or CLI subcommand we should ensure the file is downloaded as the privileged user, not as the connecting user. This allows users to download files that are owned by the `run_as` user or otherwise not accessible by the connecting user.

nmaludy:
My use case is retrieving OpenSCAP scanning results. We are not allowed to SSH as the root user, so we login with a service account, then sudo to root to run the scans (this is setup very simply using `ssh.run-as` in our inventory).

I would then like to obtain the scan results using `download_file`, however the logged-in-user doesn't have access to the root user's home directory (who ran the scan).

The workaround right now is that we have an intermediate step where we copy the scan results to a world-readable directory like `/tmp` and then obtain them with our SSH user (not ideal, but workable).

lucywyman:
## Solution
Basically the reverse process of uploading files:
1. Copy the existing file to the `tmpdir/some_random_numbers_and_letters/myfile`
2. Chown to the connecting user
3. SCP to Bolt controller
4. Cleanup the temporary instance of the file

Most important is to make sure the file is minimally accessible at each stage, and the intermediary files always get cleaned up.

krisavi:
I think mistake was made when this was implemented. I will make another ticket with same text and refer to this ticket. However it seems the tmp dir is created with another user, copying files to that temp dir may fail if user that created the tmp dir is privileged user. Issue is `mkdir -m 700` which makes that dir accessible only by the user that creates the directory. Maybe it worked and tmp dir creation was changed after this issue.

PR #2700, opened 2021-03-10
(GH-2679) Support run-as configuration when downloading files

beechtom:
This updates the SSH transport to respect `run-as` configuration when
downloading files. Specifically, when using `run-as`, Bolt executes the
following steps:

- Create a temporary directory as the run-as user
- Copy the source file to the temporary directory as the run-as user
- Chown the temporary directory and its contents to the connecting user
- Download the copy of the source file as the connecting user
- Delete the temporary directory and contents

This is a workaround to a limitation in the net-ssh library, which only
allows for downloading files as the connecting user. Because Bolt users
might not be able to connect as the root user, the source file needs to
be made available to the connecting user.

Because the temporary directory / file copy are just implementations of
this workaround, the temporary directory is _always_ deleted after
downloading the file, regardless of the user's `cleanup` configuration.

Lastly, this removes some logic for getting the source file's basename,
depending on whether the source file is a *nix or Windows path, as it is
unlikely that a user would be using a Windows-style path (`\`
separators) when connecting to a Windows target over SSH. This is
because a user is most likely to use CMD or PowerShell when connecting
to a Windows target. If the user has SSH configured to use a bash shell,
then it is likely through WSL, which uses *nix bash and expects *nix
paths.

!bug

* **Support `run-as` configuration when downloading files**
  ([#2679](https://github.com/puppetlabs/bolt/issues/2679))

  The `run-as` configuration for the SSH transport is now supported when
  downloading files.

### DIFF
--- a/lib/bolt/shell/bash.rb
+++ b/lib/bolt/shell/bash.rb
@@ -53,19 +53,35 @@ module Bolt
 
       def download(source, destination, options = {})
         running_as(options[:run_as]) do
-          # Target OS may be either Unix or Windows. Without knowing the target OS before-hand
-          # we can't assume whether the path separator is '/' or '\'. Assume we're connecting
-          # to a target with Unix and then check if the path exists after downloading.
           download = File.join(destination, Bolt::Util.unix_basename(source))
 
-          conn.download_file(source, destination, download)
+          # If using run-as, the file is copied to a tmpdir and chowned to the
+          # connecting user. This is a workaround for limitations in net-ssh that
+          # only allow for downloading files as the connecting user, which is a
+          # problem for users who cannot connect to targets as the root user.
+          # This temporary copy should *always* be deleted.
+          if run_as
+            with_tmpdir(force_cleanup: true) do |dir|
+              tmpfile = File.join(dir.to_s, Bolt::Util.unix_basename(source))
 
-          # If the download path doesn't exist, then the file was likely downloaded from Windows
-          # using a source path with backslashes (e.g. 'C:\Users\Administrator\foo'). The file
-          # should be saved to the expected location, so update the download path assuming a
-          # Windows basename so the result shows the correct local path.
-          unless File.exist?(download)
-            download = File.join(destination, Bolt::Util.windows_basename(source))
+              result = execute(['cp', '-r', source, dir.to_s], sudoable: true)
+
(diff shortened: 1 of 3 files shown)


lib/bolt/shell/bash.rb
```diff
@@ -53,19 +53,35 @@ def upload(source, destination, options = {})
...
       def download(source, destination, options = {})
         running_as(options[:run_as]) do
-          # Target OS may be either Unix or Windows. Without knowing the target OS before-hand
-          # we can't assume whether the path separator is '/' or '\'. Assume we're connecting
-          # to a target with Unix and then check if the path exists after downloading.
           download = File.join(destination, Bolt::Util.unix_basename(source))
 
-          conn.download_file(source, destination, download)
+          # If using run-as, the file is copied to a tmpdir and chowned to the
+          # connecting user. This is a workaround for limitations in net-ssh that
```
lucywyman, line 59: I'm honestly not sure that `net-ssh` should support a run-as user, this use case is somewhat specific to Bolt.

lib/bolt/shell/bash.rb
```diff
@@ -53,19 +53,35 @@ def upload(source, destination, options = {})
...
 
-          # If the download path doesn't exist, then the file was likely downloaded from Windows
-          # using a source path with backslashes (e.g. 'C:\Users\Administrator\foo'). The file
-          # should be saved to the expected location, so update the download path assuming a
-          # Windows basename so the result shows the correct local path.
-          unless File.exist?(download)
-            download = File.join(destination, Bolt::Util.windows_basename(source))
+              result = execute(['cp', '-r', source, dir.to_s], sudoable: true)
+
+              if result.exit_code != 0
```
lucywyman, line 69: Does it make sense to try to determine if slashes might be the issue, and suggest that here if there's a backslash anywhere in the path (using 'might be' language since it could always be something else)?

lucywyman (approved):
- [x] It downloads a file owned by the root user on the local transport
- [x] It downloads a file owned by the root user over the SSH transport
- [x] It downloads a file owned by the root user over the native SSH transport
- [x] It errors when the `run-as` user does not have permission to chown
  - [x] It cleansup the tmpdir even when there are errors midstream

PR #2701, opened 2021-03-10
(PE-31357) Update bolt-server to return the target's plugin hooks

ekinanp:
The puppet_library plugin hook defaults to the task plugin with the task
set to the puppet_agent::install task.

### DIFF
--- a/lib/bolt_server/transport_app.rb
+++ b/lib/bolt_server/transport_app.rb
@@ -690,11 +690,23 @@ module BoltServer
         Bolt::Util.validate_file('inventory file', context[:config].project.inventory_file)
 
         begin
+          # Set the default puppet_library plugin hook if it has not already been
+          # set
+          context[:config].data['plugin-hooks']['puppet_library'] ||= {
+            'plugin'     => 'task',
+            'task'       => 'puppet_agent::install',
+            'parameters' => {
+              'stop_service' => true
+            }
+          }
+
           connect_plugin = BoltServer::Plugin::PuppetConnectData.new(body['puppet_connect_data'])
           plugins = Bolt::Plugin.setup(context[:config], context[:pal], load_plugins: false)
           plugins.add_plugin(connect_plugin)
           inventory = Bolt::Inventory.from_config(context[:config], plugins)
-          target_list = inventory.get_targets('all').map { |targ| targ.to_h.merge({ 'transport' => targ.transport }) }
+          target_list = inventory.get_targets('all').map do |targ|
+            targ.to_h.merge({ 'transport' => targ.transport, 'plugin_hooks' => targ.plugin_hooks })
+          end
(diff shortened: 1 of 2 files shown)


lib/bolt_server/transport_app.rb
```diff
@@ -690,11 +690,23 @@ def make_winrm_target(target_hash)
         Bolt::Util.validate_file('inventory file', context[:config].project.inventory_file)
 
         begin
+          # Set the default puppet_library plugin hook if it has not already been
+          # set
+          context[:config].data['plugin-hooks']['puppet_library'] ||= {
```
ekinanp, line 695: Think this matches https://github.com/puppetlabs/orchestrator/blob/2019.8.x/src/puppetlabs/orchestrator/project_gather_target_facts.clj#L57

lib/bolt_server/transport_app.rb
```diff
@@ -690,11 +690,23 @@ def make_winrm_target(target_hash)
...
+            }
+          }
+
           connect_plugin = BoltServer::Plugin::PuppetConnectData.new(body['puppet_connect_data'])
           plugins = Bolt::Plugin.setup(context[:config], context[:pal], load_plugins: false)
           plugins.add_plugin(connect_plugin)
           inventory = Bolt::Inventory.from_config(context[:config], plugins)
-          target_list = inventory.get_targets('all').map { |targ| targ.to_h.merge({ 'transport' => targ.transport }) }
+          target_list = inventory.get_targets('all').map do |targ|
+            targ.to_h.merge({ 'transport' => targ.transport, 'plugin_hooks' => targ.plugin_hooks })
```
ekinanp, line 708: [to_h](https://github.com/puppetlabs/bolt/blob/main/lib/bolt/target.rb#L93) doesn't include `plugin_hooks`, which is why this was needed.

barriserloth (approved):

lib/bolt_server/transport_app.rb
donoghuc: I forget why we narrowed down what we send back... I think eventually we just want everything so maybe just
`inventory.get_targets('all').map(&:detail)` That way when we go to do the next thing (like perhaps facts) we get them without changing bolt code?

ekinanp: I'm not sure either.. I think it was to get `options` as top-level keys (since that is not the case in `detail`)?

I agree that we probably do want to go with `detail`, but I think that should be deferred for later since it requires some changes (e.g. moving options keys to top-level is one such change).

nicklewis: Things like `host` and `port` can be supplied in the uri, in config or _both_, so we would need to do some math to figure those values out. I agree on deferring it for now.

donoghuc (approved):

Issue #2653, opened 2021-02-18
bolt {script,command} run output should show stdout and stderr together as if it was run from a terminal


adreyer:
## Use Case
Some scripts provide useful ordered messages on both stdout and stderr. Bolt script run should support displaying these streams together.

"As a user running Bolt on the CLI I want to see stdout and stderr messages in the order they were printed, as if I was running the command or script directly on the command line"
"As a new user, I want to see the output from Bolt as if I had just run the script or command on the target directly"

## Describe the Solution You Would Like
This should be done in the transports as the streams are read from. This should be added to the result object as `value.merged_output`. We should both print the messages and also sort them into 'stdout' and 'stderr' arrays. The whole result object (`stdout`, `stderr`, and `merged_output`) should be available using the JSON format. Better document the keys available in Result objects.

We can add a flag later to disable this if folks run into issues.

For PowerShell, we should already get the interleaved output. We need to ensure all the same keys are available in PowerShell.

## Questions
* Should we default to interleaving? Yes, since it's a separate key.

PR #2702, opened 2021-03-10
(GH-2653) Merge stdout and stderr for commands and scripts

beechtom:
This updates the `bolt command|script run` commands to display
merged output from stdout and stderr on the CLI. When Bolt logs the
output from these two streams, it also adds each line to a
`merged_output` value in the order the lines are received. This
`merged_output` value is passed to the result for the command or script,
and can be accessed from both plans and the JSON output format. Lastly,
the human outputter has been updated to display the merged output by
default, resulting in output resembling what a user would see if they
ran the command or script directly on the target.

!feature

* **Display merged stdout and stderr output for commands and scripts**
  ([#2653](#2653))

  The `bolt command|script run` commands and `Invoke-BoltCommand|Script`
  cmdlets now display merged output from stdout and stderr in the CLI.
  This merged output is also available to the `Result` object in plans
  and in the JSON output format under the `merged_output` key.

### DIFF
--- a/documentation/bolt_types_reference.md
+++ b/documentation/bolt_types_reference.md
@@ -84,6 +84,27 @@ The following functions are available to `Result` objects.
 | `to_data` | `Hash` | A serialized representation of `Result`. |
 | `value` | `Hash` | The output or return of executing on the target. |
 
+#### Command and script result value keys
+
+The `Result` object returned by the `run_command` and `run_script` plan function
+includes the following keys on the `value` hash:
+
+| Key | Type | Description |
+| --- | --- | --- |
+| `exit_code` | `Number` | The command's or script's exit code. |
+| `merged_output` | `String` | Output written to both standard error (stderr) and standard out (stdout) in the order that Bolt received the output. |
+| `stderr` | `String` | Output written to standard error (stderr). |
+| `stdout` | `String` | Output written to standard out (stdout). |
+
+#### Download result value keys
+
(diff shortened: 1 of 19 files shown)


lucywyman (changes requested):
Hm I think at least one test should verify that values to the stdout and stderr streams get merged in `merged_output`.

lucywyman:
- It logs both stdout and stderr in the same output stream
  - [x] Human outputter on *nix
  - [x] Human outputter on windows
  - [x] Includes `merged_output`, `stdout`, and `stderr` in JSON outputter
- [x] Adds 'merged_output' key to Result object with the merged streams in order
  - [x] Value can be accessed by `$result.value['merged_output']`
- [x] Maintains `stdout` as it's own key on Result objects
- [x] Maintains `stderr` as it's own key on Result objects
- [x] Returns `merged_output` key from BoltSpec `run_command` function
- [x] Returns `merged_output` key from BoltSpec `run_script` function
- [x] It doesn't append extra characters if either stream has no content
- [x] It returns an empty string if neither stream has output
- [x] Works with `--tty` specified

hestonhoffman (approved):
Docs LGTM. One small thing. I wonder if we can include a cross link to writing_plans.md#returning-results-from-plans in this section (if that's the right place for people to learn about using these results).

lucywyman (approved):
:worried: This is definitely a big change, I'm honestly nervous about what folks will think. :crossed_fingers: It's well received.

Issue #2680, opened 2021-03-01
Apply blocks always report failure when inventory has tty: true


kreeuwijk:
## Describe the Bug
If the `inventory.yaml` contains `tty: true` for the SSH config, any `apply{}` block in a Bolt Plan will report a failure and cause the Plan to abort after that step. The actual resources get applied correctly, but Bolt thinks the step failed for some reason.
When `tty: true` is removed from the `inventory.yaml`, the same `apply{}` block returns success.

## Expected Behavior
Enabling or disabling the `tty` option should not affect the return code logic of `apply{}` blocks.

## Steps to Reproduce
1. Create a Bolt Plan with a simple apply block
2. Create an inventory file for a remote SSH node and set `tty: true` in the SSH config
3. Run the plan and notice Bolt reports the apply step failed, even though the puppet resources were successfully applied
4. Remove `tty: true` from the SSH config in the inventory file
5. Re-run the plan and notice Bolt now reports success for the apply step.

## Environment
 - Reproduced on Bolt 2.42, 2.44 and 3.0.1
 - Reproduced on MacOS and CentOS 7 as Bolt workstations
 - Reproduced with CentOS 7 and CentOS 8 as SSH targets

## Additional Context
It seems the return code logic breaks for some reason when `tty: true`

lucywyman:
Timebox to 1 day

PR #2708, opened 2021-03-11
(GH-2680) Do not send task parameters over stdin when using a tty

beechtom:
This fixes a bug with running tasks using the `stdin` input method when
using a `tty`. When the `tty` config option is set, Bolt creates a
wrapper script that executes the task and passes the task's parameters
over stdin. However, Bolt would still invoke this wrapper script on the
target and attempt to pass the parameters over stdin to the wrapper
script, even though the wrapper script already included the task
parameters. This resulted in the parameters being passed to the wrapper
script to print to stdout and be incorrectly returned as part of the
task's output.

!bug

* **Do not send task parameters over stdin when using a tty**
  ([#2680](https://github.com/puppetlabs/bolt/issues/2680))

  Tasks with a `stdin` input method that are run on targets with `tty:
  true` configuration no longer return the task's parameters as part of
  the task output. Previously, Bolt was sending these parameters to the
  task twice, causing them to be printed to standard out (stdout) and
  returned in the task output.

### DIFF
--- a/lib/bolt/shell/bash.rb
+++ b/lib/bolt/shell/bash.rb
@@ -144,7 +144,10 @@ module Bolt
 
             dir.chown(run_as)
 
-            execute_options[:stdin] = stdin
+            # Don't pass parameters on stdin if using a tty, as the parameters are
+            # already part of the wrapper script.
+            execute_options[:stdin] = stdin unless stdin && target.options['tty']
+
             execute_options[:sudoable] = true if run_as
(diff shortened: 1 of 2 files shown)


lucywyman (commented):
- [x] Apply blocks applied to targets with `tty` enabled do not fail
  - [x] When run using run-as
- [x] Other tasks with stdin do not receive stdin when using a tty

PR #2709, opened 2021-03-11
(docs) Fix broken transports reference links

beechtom:
This fixes some incorrect links to the transports reference that are in
the inventory reference.

!no-release-note


hestonhoffman (approved):

lucywyman (approved):
Nice! Did a quick search and looks like these are the only two mistakes that match this pattern.

Issue #2689, opened 2021-03-04
Ensure env_vars is a hash in plans


beechtom:
## Describe the Bug
Passing the `env_vars` option to scripts and commands in a plan will error with a stacktrace if the value is not a hash. Bolt should validate the option's type and error with a helpful message if it is not a hash.

Should it be string => any. We should convert to JSON in the plan function for validation, if we decide to do something special with powershell later we can handle that.

PR #2710, opened 2021-03-11
(GH-2689) Ensure env_vars is a hash, convert values to JSON

beechtom:
This updates the `run_command` and `run_script` plan functions to ensure
the `_env_vars` option is a hash. It also transforms any hash values in
the `env_vars` hash to JSON before calling the executor function,
ensuring that the environment variable is not a Ruby-style hash.

!bug

* **Ensure `env_vars` is a hash in commands and scripts**
  ([#2689](#2689))

  Bolt now ensures that the `env_vars` option passed to commands and
  scripts in plans is a hash and will raise a helpful error message
  otherwise.

* **Convert `env_vars` hash values to JSON**
  ([#2689](#2689))

  Bolt now converts hash values for an environment variable passed to a
  command or script to JSON. Previously, a hash value would be passed
  with Ruby-style syntax.

### DIFF
--- a/bolt-modules/boltlib/lib/puppet/functions/run_command.rb
+++ b/bolt-modules/boltlib/lib/puppet/functions/run_command.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require 'bolt/error'
+require 'json'
 
 # Runs a command on the given set of targets and returns the result from each command execution.
 # This function does nothing if the list of targets is empty.
@@ -13,7 +14,7 @@ Puppet::Functions.create_function(:run_command) do
   # @param options A hash of additional options.
   # @option options [Boolean] _catch_errors Whether to catch raised errors.
   # @option options [String] _run_as User to run as using privilege escalation.
-  # @option options [Hash] _env_vars Map of environment variables to set
+  # @option options [Hash[String, Any]] _env_vars Map of environment variables to set
   # @return A list of results, one entry per target.
   # @example Run a command on targets
   #   run_command('hostname', $targets, '_catch_errors' => true)
@@ -31,7 +32,7 @@ Puppet::Functions.create_function(:run_command) do
   # @param options A hash of additional options.
   # @option options [Boolean] _catch_errors Whether to catch raised errors.
   # @option options [String] _run_as User to run as using privilege escalation.
-  # @option options [Hash] _env_vars Map of environment variables to set
+  # @option options [Hash[String, Any]] _env_vars Map of environment variables to set
   # @return A list of results, one entry per target.
   # @example Run a command on targets
   #   run_command('hostname', $targets, 'Get hostname')
@@ -55,6 +56,23 @@ Puppet::Functions.create_function(:run_command) do
 
     options = options.transform_keys { |k| k.sub(/^_/, '').to_sym }
     options[:description] = description if description
+
+    # Ensure env_vars is a hash and that each hash value is transformed to JSON
+    # so we don't accidentally pass Ruby-style data to the target.
+    if options[:env_vars]
+      unless options[:env_vars].is_a?(Hash)
+        raise Bolt::ValidationError, "Option 'env_vars' must be a hash"
+      end
+
+      if (bad_keys = options[:env_vars].keys.reject { |k| k.is_a?(String) }).any?
+        raise Bolt::ValidationError,
+              "Keys for option 'env_vars' must be strings: #{bad_keys.map(&:inspect).join(', ')}"
+      end
+
+      options[:env_vars] = options[:env_vars].transform_values do |val|
+        [Array, Hash].include?(val.class) ? val.to_json : val
+      end
+    end
 
     executor = Puppet.lookup(:bolt_executor)
     inventory = Puppet.lookup(:bolt_inventory)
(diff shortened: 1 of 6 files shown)


beechtom:
Opted to just convert hash values to JSON, since converting other values results in extra quoting that would likely affect users. For example, the string `foo` would become the string `"foo"`.

bolt-modules/boltlib/lib/puppet/functions/run_command.rb
```diff
@@ -56,6 +57,18 @@ def run_command_with_description(command, targets, description = nil, options =
     options = options.transform_keys { |k| k.sub(/^_/, '').to_sym }
     options[:description] = description if description
 
+    # Ensure env_vars is a hash and that each hash value is transformed to JSON
+    # so we don't accidentally pass Ruby-style data to the target.
+    if options[:env_vars]
```
lucywyman, line 62: We also need to validate that the keys are strings here

bolt-modules/boltlib/lib/puppet/functions/run_command.rb
```diff
@@ -56,6 +57,18 @@ def run_command_with_description(command, targets, description = nil, options =
...
+      unless options[:env_vars].is_a?(Hash)
+        raise Bolt::ValidationError, "Option 'env_vars' must be a hash"
+      end
+
+      options[:env_vars] = options[:env_vars].transform_values do |val|
+        val.is_a?(Hash) ? val.to_json : val
```
lucywyman, line 68: Is there a reason not to do this for all types? This becomes weird when, for example, you have an array that includes a hash and that *doesn't* get converted:
```
lucy ~/githubs/bolt ((40f60cdb...)) $ be bolt plan run bolt -t localhost
...
Finished on localhost:
  STDOUT:
    ["hi", "there", {"hi"=>"there"}]
```

bolt-modules/boltlib/lib/puppet/functions/run_script.rb
```diff
@@ -85,6 +85,18 @@ def run_script_with_description(scope, script, targets, description = nil, optio
     options[:description] = description if description
     options[:pwsh_params] = pwsh_params if pwsh_params
 
+    # Ensure env_vars is a hash and that each hash value is transformed to JSON
+    # so we don't accidentally pass Ruby-style data to the target.
+    if options[:env_vars]
```
lucywyman, line 90: Same changes here.

lucywyman:
- Enforces env_vars being a hash in run_command
  - [x] YAML plan
  - [x] Puppet plan
- Enforces hash keys being strings
  - [ ] YAML plan
  - [ ] Puppet plan
- Allows non-string types as values
  - [x] YAML plan
  - [x] Puppet plan
- Enforces env_vars being a hash in run_script
  - [x] YAML plan
  - [x] Puppet plan
- [x] Converts hash values to JSON
  - Only if they're not embedded in other types

bolt-modules/boltlib/lib/puppet/functions/run_command.rb
beechtom: I think we actually do want to do it for arrays, too. My hesitance with doing it for all types is that it changes existing behavior for string values. Converting strings to JSON will add quotes to the beginning and end of the string, which might affect users using this option...

lucywyman: Sorry I completely missed your comment above! Agreed.

bolt-modules/boltlib/lib/puppet/functions/run_command.rb
```diff
@@ -56,6 +57,22 @@ def run_command_with_description(command, targets, description = nil, options =
...
 
+    # Ensure env_vars is a hash and that each hash value is transformed to JSON
+    # so we don't accidentally pass Ruby-style data to the target.
+    if options[:env_vars]
+      unless options[:env_vars].is_a?(Hash)
+        raise Bolt::ValidationError, "Option 'env_vars' must be a hash"
+      end
+
+      unless options[:env_vars].keys.all? { |k| k.is_a?(String) }
+        raise Bolt::ValidationError, "Keys for option 'env_vars' must be strings"
```
lucywyman, line 68: Hm, can you add the specific value(s) that are failing here? I know other validation errors here don't have identifying information, and we should probably fix those too - but in lieu of line numbers (in YAML plans) it'd be handy to have the value that's failing so users can find it.

lucywyman (approved):
One small todo, looks great!

bolt-modules/boltlib/lib/puppet/functions/run_command.rb
```diff
@@ -56,6 +57,23 @@ def run_command_with_description(command, targets, description = nil, options =
     options = options.transform_keys { |k| k.sub(/^_/, '').to_sym }
     options[:description] = description if description
 
+    # Ensure env_vars is a hash and that each hash value is transformed to JSON
+    # so we don't accidentally pass Ruby-style data to the target.
+    if options[:env_vars]
+      unless options[:env_vars].is_a?(Hash)
+        raise Bolt::ValidationError, "Option 'env_vars' must be a hash"
```
lucywyman, line 64: Hmm in thinking more about it I think the right thing to do for this and the other validation errors here is to additionally validate them as part of the validate functions for the YAML steps in the 'step key' classes - that way they'll fail fast for YAML plans and include the line number there. It looks like that's what we do for other validation errors in plan functions.

beechtom: Updated to validate the type in the step itself. This doesn't include the line number, but does include the step name/number.

lucywyman (approved):

PR #2711, opened 2021-03-12
(hack) Add welcome message when users first run Bolt

lucywyman:
This adds a nice welcome message when users first run the Bolt command.
The message is printed to stdout, and is only printed if the following
is true:
- The stdout stream is a tty
- The user runs `bolt`, `bolt help` or `bolt --help`. This should not
  print for any other Bolt commands, such as `bolt --version` or `bolt
  command --help`.
- The user has not run Bolt before. This is determined by the presence
  of a file `~/.puppetlabs/etc/bolt/.first_runs_free`.

!feature

* **Add welcome message when users first run Bolt**
  Bolt now prints a friendly welcome message when users first run Bolt
  if they run `bolt`, `bolt --help`, or `bolt help`.

### DIFF
--- a/lib/bolt/cli.rb
+++ b/lib/bolt/cli.rb
@@ -96,6 +96,48 @@ module Bolt
       finalize_setup
     end
 
+    # Prints a welcome message when users first install Bolt and run `bolt`, `bolt help` or `bolt --help`
+    def welcome_message
+      bolt = <<~BOLT
+                   `.::-`
+              `.-:///////-.`
+           `-:////:.  `-:///:-  /ooo.                        .ooo/
+       `.-:///::///:-`   `-//:  ymmm-                        :mmmy  .---.
+      :///:-.   `.:////.  -//:  ymmm-                        :mmmy  +mmm+
+      ://.          ///.  -//:  ymmm--/++/-       `-/++/:`   :mmmy-:smmms::-
+      ://.          ://. .://:  ymmmdmmmmmmdo`  .smmmmmmmmh: :mmmysmmmmmmmms
+      ://.          ://:///:-.  ymmmh/--/hmmmy -mmmd/-.:hmmm+:mmmy.-smmms--.
+      ://:.`      .-////:-`     ymmm-     ymmm:hmmm-    `dmmm/mmmy  +mmm+
+      `-:///:-..:///:-.`        ymmm-     ommm/dmmm`     hmmm+mmmy  +mmm+
+         `.-:////:-`            ymmm+    /mmmm.ommms`   /mmmh:mmmy  +mmmo
+             `-.`               ymmmmmhhmmmmd:  ommmmhydmmmy`:mmmy  -mmmmdhd
+                                oyyy+shddhs/`    .+shddhy+-  -yyyo   .ohddhs
+
+
+      BOLT
+      example_cmd = if Bolt::Util.windows?
+                      "Invoke-BoltCommand -Command 'hostname' -Targets localhost"
+                    else
+                      "bolt command run 'hostname' --target localhost"
+                    end
+      prev_cmd = String.new("bolt")
+      prev_cmd << " #{@argv[0]}" unless @argv.empty?
+
+      message = <<~MSG
+      🎉 Welcome to Bolt #{VERSION}
+      😌 We're here to help bring order to the chaos
+      📖 Find our documentation at https://bolt.guide
+      🙋 Ask a question in #bolt on https://slack.puppet.com/
+      🔩 Contribute at https://github.com/puppetlabs/bolt/
+      💡 Not sure where to start? Try "#{example_cmd}"
+
+      We only print this message once. Run "#{prev_cmd}" again for help text.
+      MSG
+
+      $stdout.print "\033[36m#{bolt}\033[0m"
+      $stdout.print message
+    end
+
     # Parses the command and validates options. All errors that are raised here
     # are not handled by the outputter, as it relies on config being loaded.
     def parse_command
@@ -106,6 +148,16 @@ module Bolt
         # If the subcommand is not enabled, display the default
         # help text
         options[:subcommand] = nil unless COMMANDS.include?(options[:subcommand])
+
+        if Bolt::Util.first_run?
+          FileUtils.mkdir_p(Bolt::Util.first_runs_free.dirname)
+          FileUtils.touch(Bolt::Util.first_runs_free)
+
+          if options[:subcommand].nil? && $stdout.isatty
+            welcome_message
+            raise Bolt::CLIExit
+          end
+        end
 
         # Update the parser for the subcommand (or lack thereof)
         parser.update
--- a/lib/bolt/util.rb
+++ b/lib/bolt/util.rb
@@ -77,6 +77,14 @@ module Bolt
         File.exist?(path) ? read_yaml_hash(path, file_name) : {}
       end
 
+      def first_runs_free
+        Bolt::Config.user_path + '.first_runs_free'
+      end
+
+      def first_run?
+        Bolt::Config.user_path && !File.exist?(first_runs_free)
+      end
+
       # Accepts a path with either 'plans' or 'tasks' in it and determines
       # the name of the module
       def module_name(path)


lucywyman:
This ends up being hard to write automated tests for since `$stdout.isatty` is false in RSpec, and mocking it gets kind of hairy. I think just manual testing is enough.

lib/bolt/util.rb
```diff
@@ -77,6 +77,14 @@ def read_optional_yaml_hash(path, file_name)
         File.exist?(path) ? read_yaml_hash(path, file_name) : {}
       end
 
+      def first_runs_free
+        Pathname.new(File.expand_path(File.join('~', '.puppetlabs', 'etc', 'bolt', '.first_runs_free')))
```
beechtom, line 81: This will blow up if there's no homedir. We should use [`Bolt::Config.user_path`](https://github.com/puppetlabs/bolt/blob/main/lib/bolt/config.rb#L73) and handle the case where that returns `nil`.

lib/bolt/cli.rb
```diff
@@ -96,6 +96,48 @@ def parse
...
+ `.-:///::///:-`   `-//:  ymmm-                        :mmmy  .---.
+:///:-.   `.:////.  -//:  ymmm-                        :mmmy  +mmm+
+://.          ///.  -//:  ymmm--/++/-       `-/++/:`   :mmmy-:smmms::-
+://.          ://. .://:  ymmmdmmmmmmdo`  .smmmmmmmmh: :mmmysmmmmmmmms
+://.          ://:///:-.  ymmmh/--/hmmmy -mmmd/-.:hmmm+:mmmy.-smmms--.
+://:.`      .-////:-`     ymmm-     ymmm:hmmm-    `dmmm/mmmy  +mmm+
+`-:///:-..:///:-.`        ymmm-     ommm/dmmm`     hmmm+mmmy  +mmm+
+   `.-:////:-`            ymmm+    /mmmm.ommms`   /mmmh:mmmy  +mmmo
+       `-.`               ymmmmmhhmmmmd:  ommmmhydmmmy`:mmmy  -mmmmdhd
+                          oyyy+shddhs/`    .+shddhy+-  -yyyo   .ohddhs
```
beechtom, line 114: Can this be indented a few times so it's in line with the `bolt` variable?

lib/bolt/util.rb
```diff
@@ -77,6 +77,14 @@ def read_optional_yaml_hash(path, file_name)
         File.exist?(path) ? read_yaml_hash(path, file_name) : {}
       end
 
+      def first_runs_free
+        Bolt::Config.user_path + '.first_runs_free'
```
beechtom, line 81: `Bolt::Config.user_path` returns `nil` if there's no homedir, so this will still fail.

beechtom, line 81: Would using `analytics.yaml` instead of a new file work here? That'd prevent the welcome message displaying for existing users.

lucywyman: We don't write `analytics.yaml` if the help text is printed, because then we have to load config which will significantly slow down this first help text run kind of unnecessarily. I don't think it's worth the tradeoff.

lucywyman:
- [ ] Does not print if `~/.puppetlabs/etc/bolt/.first_runs_free` is present
- [ ] Prints when running `bolt`
- [ ] Prints when running `bolt --help`
- [ ] Prints when running `bolt help`
- [ ] Doesn't print if they run `bolt --version` first
- [ ] Doesn't print if they run `bolt command run --help`

lib/bolt/util.rb
```diff
@@ -77,6 +77,15 @@ def read_optional_yaml_hash(path, file_name)
         File.exist?(path) ? read_yaml_hash(path, file_name) : {}
       end
 
+      def first_runs_free
+        path = Bolt::Config.user_path || Bolt::Config.system_path
```
beechtom, line 81: I'm not too familiar with the cases where the running user does not have a homedir, but are they guaranteed to have permissions to write to the system path?

lib/bolt/cli.rb
```diff
@@ -96,6 +96,48 @@ def parse
...
+                    end
+      prev_cmd = String.new("bolt")
+      prev_cmd << " #{@argv[0]}" unless @argv.empty?
+
+      message = <<~MSG
+      🎉 Welcome to Bolt #{VERSION}
+      😌 We're here to help bring order to the chaos
+      📖 Find our documentation at https://bolt.guide
+      🙋 Ask a question in #bolt on https://slack.puppet.com/
+      🤖 Contribute at https://github.com/puppetlabs/bolt/
```
nicklewis, line 131: This emoji doesn't work for me in Terminal. It does show up in iTerm 2 though. Strange...

lib/bolt/util.rb
nicklewis: Probably not.

lib/bolt/cli.rb
lucywyman: Oo actually what does it look like when it doesn't work? If it just doesn't show up that seems fine, but if there's weird characters I can remove them. Or if there's some way to check if they're supported first we can do that

nicklewis: Actually I zoomed in to take a screenshot and it _is_ a robot! Just not one that matches the rest of the emoji on my system. It looks like my font (Source Code Pro) just includes its own robot character that doesn't look much like a robot at 10pt.

image

lib/bolt/util.rb
lucywyman: Ok, well, I think if the user doesn't have a homedir it's also most likely a case where we don't need to print a welcome message anyway. Rather than trying to catch errors and write the file somewhere else I just made `Bolt::Config.user_path` succeeding a requirement for printing the text to begin with.

lib/bolt/cli.rb
lucywyman: Oh hmm...to be honest I don't love the robot for the "Contribute" line anyway, if we want to replace that...maybe with 🔩 ?

lib/bolt/util.rb
```diff
@@ -77,6 +77,15 @@ def read_optional_yaml_hash(path, file_name)
         File.exist?(path) ? read_yaml_hash(path, file_name) : {}
       end
 
+      def first_runs_free
+        path = Bolt::Config.user_path
+        path + '.first_runs_free'
+      end
+
+      def first_run?
+        !File.exist?(first_runs_free)
```
beechtom, line 86:
```suggestion
        Bolt::Config.user_path && !File.exist?(first_runs_free)
```

I think this should cover cases where homedir doesn't exist and return `false`.

lib/bolt/cli.rb
```diff
@@ -107,6 +149,16 @@ def parse_command
         # help text
         options[:subcommand] = nil unless COMMANDS.include?(options[:subcommand])
 
+        if Bolt::Util.first_run? && Bolt::Config.user_path
```
beechtom, line 152:
```suggestion
        if Bolt::Util.first_run?
```

This will still fail if the homedir doesn't exist, since `first_run?` will call `first_runs_free` and try to concatenate with `nil`.

lib/bolt/cli.rb
```diff
@@ -107,6 +149,16 @@ def parse_command
...
+          FileUtils.mkdir_p(Bolt::Util.first_runs_free.dirname)
+          FileUtils.touch(Bolt::Util.first_runs_free)
+
+          if options[:subcommand].nil? && $stdout.isatty
```
beechtom, line 156: Could `$stdout.isatty` be part of `Bolt::Util.first_run?` ? Not sure if there's a need to create the first_runs_free file if stdin isn't a tty.

lucywyman: I think there might be other cases where we want to know if this is the first time Bolt has been run or not that won't rely on whether stdout is a tty, so I'd like to leave that here as it's specific to the welcome message.

lib/bolt/util.rb
```diff
@@ -77,6 +77,15 @@ def read_optional_yaml_hash(path, file_name)
         File.exist?(path) ? read_yaml_hash(path, file_name) : {}
       end
 
+      def first_runs_free
+        path = Bolt::Config.user_path
+        path + '.first_runs_free'
```
beechtom, line 82:
```suggestion
        Bolt::Config.user_path + '.first_runs_free'
```

beechtom (approved):
One small change, but otherwise lgtm!

lib/bolt/util.rb
lucywyman: Oh my gosh I actually only did this because the `noop` bug scared me off :sweat:

PR #2712, opened 2021-03-15
Run 'plan convert' with plan name

beechtom:
This adds support for calling `bolt plan convert` and `Convert-BoltPlan`
with the name of a plan instead of just the path to a plan.

!feature

* **Convert YAML plans by name**

  The `bolt plan convert` and `Convert-BoltPlan` commands now accept the
  name of a YAML plan to convert instead of just a path to a YAML plan.

### DIFF
--- a/lib/bolt/pal.rb
+++ b/lib/bolt/pal.rb
@@ -521,10 +521,30 @@ module Bolt
       end
     end
 
-    def convert_plan(plan_path)
+    def convert_plan(plan)
+      path = File.expand_path(plan)
+
+      # If the path doesn't exist, check if it's a plan name
+      unless File.exist?(path)
+        in_bolt_compiler do |compiler|
+          sig = compiler.plan_signature(plan)
+
+          # If the plan was loaded, look for it on the module loader
+          # There has to be an easier way to do this...
+          if sig
+            type = compiler.list_plans.find { |p| p.name == plan }
+            path = sig.instance_variable_get(:@plan_func)
+                      .loader
+                      .find(type)
+                      .origin
+                      .first
+          end
+        end
+      end
+
       Puppet[:tasks] = true
       transpiler = YamlPlan::Transpiler.new
-      transpiler.transpile(plan_path)
+      transpiler.transpile(path)
     end
 
     # Returns a mapping of all modules available to the Bolt compiler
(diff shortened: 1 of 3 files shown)


spec/integration/transpiler_spec.rb
```diff
@@ -47,12 +47,18 @@
   }
   PLAN
 
-  it 'transpiles a yaml plan' do
+  it 'transpiles a YAML plan from a path' do
```
lucywyman, line 50: Can this also include a quick test for transpiling by name?

beechtom: I added a test below this one that converts by name.

beechtom: https://github.com/puppetlabs/bolt/blob/89278e92f48ae0b7ca92ee9025fcc9a9bd198871/spec/integration/transpiler_spec.rb#L56

lib/bolt/pal.rb
```diff
@@ -521,10 +521,28 @@ def get_plan_mtime(path)
       end
     end
 
-    def convert_plan(plan_path)
+    def convert_plan(plan)
+      # If the path doesn't exist, check if it's a plan name
+      unless File.exist?(plan)
```
lucywyman, line 526: It looks like we immediately expand this path in the `transpile` method and then don't reuse the relative path - I'd definitely sleep a little better if we just expand it here before checking if the file exists and then pass the absolute path to the transpile method (whether it's from a path or plan name)

lib/bolt/pal.rb
```diff
@@ -521,10 +521,30 @@ def get_plan_mtime(path)
...
+
+      # If the path doesn't exist, check if it's a plan name
+      unless File.exist?(path)
+        in_bolt_compiler do |compiler|
+          sig = compiler.plan_signature(plan)
+
+          # If the plan was loaded, look for it on the module loader
+          # There has to be an easier way to do this...
+          if sig
+            type = compiler.list_plans.find { |p| p.name == plan }
```
lucywyman, line 535: Hmm I spent some time looking to see if there's a better way, but I think that might have diminishing returns - ultimately I don't think it's going to be a whole lot more efficient or readable than this :/

lucywyman (approved):

lib/bolt/pal.rb
nicklewis: I think you can do something like `Puppet.lookup(:loaders).private_environment_loader.find(Puppet::Pops::Parser::TypedName.new(:plan, plan))`

Issue #2704, opened 2021-03-11
Prompt: Support default values


fetzerms:
## Use Case
Running a plan, the user needs to input a username. In 95% of the cases, the username can be derived from a environmental variable ($USER). But in the other cases, the user needs to specify a different username.

## Describe the Solution You Would Like
I want to be able to specify a default value when prompt()-ing in a plan. The default value should be used, when the user does not enter any data.

```
  $username = prompt("Enter your username", {
    'default_value' => "almightyadminuser",
    'sensitive' => true
  })
```

Results to:

```
  "Enter your username [almightyadminuser]: "
```

And in case the user simply presses enter, the default gets assigned to $username. I already did some experiments in the bolt source code. It seems pretty feasible.

PR #2713, opened 2021-03-15
(GH-2704) Add default value for 'prompt' plan function

beechtom:
This adds a new `default` option for the `prompt` plan function. If a
user does not provide input when prompted, and the `default` option is
set, the function will return the `default` value. The prompt message
itself will display the default value, unless the `sensitive` value is
also set.

!feature

* **Add default value for `prompt` plan function**
  ([#2704](https://github.com/puppetlabs/bolt/issues/2704))

  The `prompt` plan function has a new `default` option which can be
  used to return a default value when a user does not provide input.

### DIFF
--- a/bolt-modules/prompt/lib/puppet/functions/prompt.rb
+++ b/bolt-modules/prompt/lib/puppet/functions/prompt.rb
@@ -11,12 +11,24 @@ Puppet::Functions.create_function(:prompt) do
   # @option options [Boolean] sensitive Disable echo back and mark the response as sensitive.
   #   The returned value will be wrapped by the `Sensitive` data type. To access the raw
   #   value, use the `unwrap` function (i.e. `$sensitive_value.unwrap`).
+  # @option options [String] default The default value to return if the user does not provide
+  #   input or if stdin is not a tty.
   # @return The response to the prompt.
   # @example Prompt the user if plan execution should continue
   #   $response = prompt('Continue executing plan? [Y\N]')
   # @example Prompt the user for sensitive information
   #   $password = prompt('Enter your password', 'sensitive' => true)
   #   out::message("Password is: ${password.unwrap}")
+  # @example Prompt the user and provide a default value
+  #   $user = prompt('Enter your login username', 'default' => 'root')
+  # @example Prompt the user for sensitive information, returning a sensitive default if one is not provided
+  #   $token = prompt('Enter token', 'default' => lookup('default_token'), 'sensitive' => true)
+  #   out::message("Token is: ${token.unwrap}")
+  # @example Prompt the user and fail with a custom message if no input was provided
+  #   $response = prompt('Enter your name', 'default' => '')
+  #   if $response.empty {
+  #     fail_plan('Must provide your name')
+  #   }
   dispatch :prompt do
     param 'String', :prompt
     optional_param 'Hash[String[1], Any]', :options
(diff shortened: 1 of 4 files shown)


bolt-modules/prompt/lib/puppet/functions/prompt.rb
```diff
@@ -30,14 +34,20 @@ def prompt(prompt, options = {})
...
+    options  = options.transform_keys(&:to_sym)
     executor = Puppet.lookup(:bolt_executor)
+
     # Send analytics report
     executor.report_function_call(self.class.name)
 
+    # Require default to be a string value
+    if options.key?(:default) && !options[:default].is_a?(String)
+      raise Bolt::ValidationError, "Option 'default' must be a string"
+    end
```
beechtom, line 46: Should this allow `nil`/`Undef` as a default?

lucywyman: Hmm I'd be interested in waiting until someone asks for it to understand their use case better - I think an empty string will work for basically making the prompt "optional". It'd be useful to include that use case as an example I think:
```
$prompt = prompt("Password please", "default" => "")
if $prompt.empty {
  # Fail the plan with a custom message
  fail_plan("Please provide a password to the prompt")
} elsif {
# Some validation here...
} else {
  run_script(...)
}
```
Something along those lines.

bolt-modules/prompt/lib/puppet/functions/prompt.rb
```diff
@@ -11,12 +11,16 @@
   # @option options [Boolean] sensitive Disable echo back and mark the response as sensitive.
   #   The returned value will be wrapped by the `Sensitive` data type. To access the raw
   #   value, use the `unwrap` function (i.e. `$sensitive_value.unwrap`).
+  # @option options [String] default The default value to return if the user does not provide
+  #   input or if standard in (stdin) is not a tty.
```
lucywyman, line 15: To be perfectly honestly I've never seen it referred to as standard in - @hestonhoffman is that actually helpful to expand, or can we just say 'stdin'?

bolt-modules/prompt/lib/puppet/functions/prompt.rb
```diff
@@ -30,14 +34,20 @@ def prompt(prompt, options = {})
...
     executor.report_function_call(self.class.name)
 
+    # Require default to be a string value
+    if options.key?(:default) && !options[:default].is_a?(String)
+      raise Bolt::ValidationError, "Option 'default' must be a string"
+    end
+
     response = executor.prompt(prompt, options)
 
+    # If sensitive, wrap it
```
lucywyman, line 50: An example of this would also be great, since you can't set the default to an actual `Sensitive` type and I expect this will be a pretty common use case.

bolt-modules/prompt/spec/functions/prompt_spec.rb
```diff
@@ -28,6 +28,20 @@
     expect(result.unwrap).to eq(response)
   end
 
+  it 'returns a default value if no input is provided' do
```
lucywyman, line 31: A test to verify wrapping + not printing if `:sensitive` is provided would also be good, since that'd be a pretty bad bug.

lib/bolt/executor.rb
```diff
@@ -485,21 +485,29 @@ def wait_until(timeout, retry_interval)
 
     def prompt(prompt, options)
       @prompting = true
+
       unless $stdin.tty?
```
lucywyman, line 489: Hm, I don't entirely grok this behavior but setting `@prompting` to true when `$stdin.tty?` is false seems to cause this to spin forever. If we move this check to be before `@prompting = true` then this succeeds as expected.

lucywyman:
- [x] Sets the default value
- [x] Errors if the default is not a string
- [x] Accepts an empty string as a default
- [x] Does not print the default if it's sensitive
- [ ] Uses the default if running without a tty

bolt-modules/prompt/lib/puppet/functions/prompt.rb
hestonhoffman: Technically, stdin is an abbreviation for "standard input", so best practice would be to write it out the first time and use (stdin) in parenthesis (We've used it that way here https://puppet.com/docs/bolt/latest/running_bolt_commands.html). 

On the other hand, this is a code comment, and will most likely show up in reference docs, and as you've suggested, it's a very common abbreviation in this field. I think it's okay to use the abbreviation in the code comments and stick to the best practice in our "long-form" docs. How does that sound? There are probably instances in the current docs where we're not doing this. I can root those out in a separate PR.

lucywyman (approved):
Going to let Heston do a quick docs review :+1:

hestonhoffman (approved):
Sorry, took a while to get to this cos I had to go on an rbenv adventure to get the rake task to run. LGTM!

PR #2714, opened 2021-03-15
(hack) Add 'prompt::menu' plan function

beechtom:
This adds a new `prompt::menu` plan function. It accepts a prompt and
list of options to choose from, and prompts the user to choose one of
the options. The function loops until a valid option is selected, or
returns an optional default value.

!feature

* **`prompt::menu` plan function**

  The new `prompt::menu` plan function can be used to prompt the user to
  select an option from a menu of options.

---

Blocked by #2713

### DIFF
--- a/bolt-modules/prompt/lib/puppet/functions/prompt/menu.rb
+++ b/bolt-modules/prompt/lib/puppet/functions/prompt/menu.rb
@@ -1,0 +1,103 @@
+# frozen_string_literal: true
+
+require 'bolt/error'
+
+# Display a menu prompt and wait for a response. Continues to prompt
+# until an option from the menu is selected.
+#
+# > **Note:** Not available in apply block
+Puppet::Functions.create_function(:'prompt::menu') do
+  # Select from a list of options.
+  # @param prompt The prompt to display.
+  # @param menu A list of options to choose from.
+  # @param options A hash of additional options.
+  # @option options [String] default The default option to return if the user does not provide
+  #   input or if standard in (stdin) is not a tty. Must be an option present in the menu.
+  # @return The selected option.
+  # @example Prompt the user to select from a list of options
+  #   $selection = prompt::menu('Select a fruit', ['apple', 'banana', 'carrot'])
+  # @example Prompt the user to select from a list of options with a default value
+  #   $selection = prompt::menu('Select environment', ['development', 'production'], 'default' => 'development')
+  dispatch :prompt_menu_array do
+    param 'String', :prompt
+    param 'Array[Variant[Target, Data]]', :menu
+    optional_param 'Hash[String[1], Variant[Target, Data]]', :options
+    return_type 'Variant[Target, Data]'
+  end
+
+  # Select from a list of options with custom inputs.
+  # @param prompt The prompt to display.
+  # @param menu A hash of options to choose from, where keys are the input used to select a value.
+  # @param options A hash of additional options.
+  # @option options [String] default The default option to return if the user does not provide
+  #   input or if standard in (stdin) is not a tty. Must be an option present in the menu.
+  # @return The selected option.
+  # @example Prompt the user to select from a list of options with custom inputs
+  #   $menu = { 'y' => 'yes', 'n' => 'no' }
+  #   $selection = prompt::menu('Install Puppet?', $menu)
+  dispatch :prompt_menu do
+    param 'String', :prompt
+    param 'Hash[String[1], Variant[Target, Data]]', :menu
+    optional_param 'Hash[String[1], Variant[Target, Data]]', :options
+    return_type 'Variant[TargetSpec, Data]'
+  end
+
+  def prompt_menu_array(prompt, menu, options = {})
+    menu_hash = menu.map.with_index { |value, index| [(index + 1).to_s, value] }.to_h
+    prompt_menu(prompt, menu_hash, options)
+  end
+
+  def prompt_menu(prompt, menu, options = {})
+    unless Puppet[:tasks]
+      raise Puppet::ParseErrorWithIssue
+        .from_issue_and_stack(Bolt::PAL::Issues::PLAN_OPERATION_NOT_SUPPORTED_WHEN_COMPILING,
+                              action: 'prompt::menu')
+    end
+
+    options  = options.transform_keys(&:to_sym)
+    executor = Puppet.lookup(:bolt_executor)
+
+    # Send analytics report
+    executor.report_function_call(self.class.name)
+
+    # Error if there are no options
+    if menu.empty?
+      raise Bolt::ValidationError, "Menu cannot be empty"
+    end
+
+    # Error if the default value is not on the menu
+    if options.key?(:default) && !menu.value?(options[:default])
+      raise Bolt::ValidationError, "Default value '#{options[:default]}' is not one of the provided menu options"
+    end
+
+    # The first prompt should include the menu
+    to_prompt = format_menu(menu) + prompt
+
+    # Request input from the user until they provide a valid option
+    loop do
+      selection = executor.prompt(to_prompt, options)
+
+      return menu[selection] if menu.key?(selection)
+      return selection       if options.key?(:default) && options[:default] == selection
+
+      # Only reprint the prompt, not the menu
+      to_prompt = "Invalid option, try again. #{prompt}"
+    end
+  end
+
+  # Builds the menu string. Aligns all the values by padding input keys.
+  #
+  private def format_menu(menu)
+    # Find the longest input and add 2 for wrapping parenthesis
+    key_length = menu.keys.max_by(&:length).length + 2
+
+    menu_string = +''
+
+    menu.each do |key, value|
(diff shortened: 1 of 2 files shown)


bolt-modules/prompt/lib/puppet/functions/prompt/menu.rb
```diff
@@ -0,0 +1,103 @@
...
+  #   input or if standard in (stdin) is not a tty. Must be an option present in the menu.
+  # @return The selected option.
+  # @example Prompt the user to select from a list of options
+  #   $selection = prompt::menu('Select a fruit', ['apple', 'banana', 'carrot'])
+  # @example Prompt the user to select from a list of options with a default value
+  #   $selection = prompt::menu('Select environment', ['development', 'production'], 'default' => 'development')
+  dispatch :prompt_menu_array do
+    param 'String', :prompt
+    param 'Array[Any]', :menu
+    optional_param 'Hash[String[1], Any]', :options
```
lucywyman, line 24: Should this be `String[1], String`?

beechtom: That would limit the values that you can choose from to strings, which I don't see a good reason for doing. Only the keys need to be strings, since you need to be able to type the input.

lucywyman: Hm, I guess since the menu can accept any types we're not beholden to what the user types in (like with `prompt`), so this actually could be `Any`... :thinking: I think it'd be simpler if everything accepted strings, but I don't think that's a good enough reason to limit it since I'm not sure what we'd do different once someone asked for this feature.

beechtom: Heh, whoops, ignore that last comment. I was thinking this was for the `menu` param. Since the values can be anything, the `default` option also needs to have the `Any` type.

bolt-modules/prompt/lib/puppet/functions/prompt/menu.rb
```diff
@@ -0,0 +1,103 @@
...
+
+    # Request input from the user until they provide a valid option
+    loop do
+      selection = executor.prompt(to_prompt, options)
+
+      return menu[selection] if menu.key?(selection)
+      return selection       if options.key?(:default) && options[:default] == selection
+
+      # Only reprint the prompt, not the menu
+      to_prompt = prompt
```
lucywyman, line 84: It'd be a nice touch if this prepended something like "Woops, that input is invalid. Please try again.\nPrompt:" or something.

bolt-modules/prompt/lib/puppet/functions/prompt/menu.rb
```diff
@@ -0,0 +1,103 @@
...
+  #
+  private def format_menu(menu)
+    # Find the longest input and add 2 for wrapping parenthesis
+    key_length = menu.keys.max_by(&:length).length + 2
+
+    menu_string = +''
+
+    menu.each do |key, value|
+      key = "(#{key})".ljust(key_length)
+      menu_string << "#{key} #{value}\n"
```
lucywyman, line 98: It'd be nice to use `safe_name` here if the type is a Target.

beechtom: Stringified targets use the safe name already: https://github.com/puppetlabs/bolt/blob/main/lib/bolt/target.rb#L63-L65 .

```
$ be bolt plan run bolt -t ssh://user:password@example.com:22
Starting: plan bolt
(1) ssh://user@example.com:22
Pick a target:
```

bolt-modules/prompt/lib/puppet/functions/prompt/menu.rb
```diff
@@ -0,0 +1,103 @@
...
+  #   input or if standard in (stdin) is not a tty. Must be an option present in the menu.
+  # @return The selected option.
+  # @example Prompt the user to select from a list of options with custom inputs
+  #   $menu = { 'y' => 'yes', 'n' => 'no' }
+  #   $selection = prompt::menu('Install Puppet?', $menu)
+  dispatch :prompt_menu do
+    param 'String', :prompt
+    param 'Hash[String[1], Any]', :menu
+    optional_param 'Hash[String[1], Any]', :options
+    return_type 'Data'
```
lucywyman, line 42: Ah, ok, if we're going to limit it to `Data` (which I think is reasonable) then we should do that in the `menu` parameter and `default` too.

Notably, this excludes sensitive values, but those won't be printed anyway and it's hard to think of a case where you'd want to choose between a bunch of values where they're all `Sensitive [redacted]`.

bolt-modules/prompt/spec/functions/prompt/menu_spec.rb
```diff
@@ -0,0 +1,96 @@
...
+describe 'prompt::menu' do
+  let(:executor)      { Bolt::Executor.new }
+  let(:tasks_enabled) { true }
+
+  around(:each) do |example|
+    Puppet[:tasks] = tasks_enabled
+    Puppet.override(bolt_executor: executor) { example.run }
+  end
+
+  # before(:each) do
```
lucywyman, line 15: Leftover?

bolt-modules/prompt/lib/puppet/functions/prompt/menu.rb
beechtom: Updated the return type to `Any`. I think there's a compelling use case for selecting targets, and `Target` objects aren't `Data`.

bolt-modules/prompt/lib/puppet/functions/prompt/menu.rb
```diff
@@ -0,0 +1,103 @@
...
+    executor.report_function_call(self.class.name)
+
+    # Error if there are no options
+    if menu.empty?
+      raise Bolt::ValidationError, "Menu cannot be empty"
+    end
+
+    # Error if the default value is not on the menu
+    if options.key?(:default) && !menu.value?(options[:default])
+      raise Bolt::ValidationError, "Default value '#{options[:default]}' is not a valid option"
```
lucywyman, line 70:
```suggestion
      raise Bolt::ValidationError, "Default value '#{options[:default]}' is not one of the provided menu options."
```

bolt-modules/prompt/lib/puppet/functions/prompt/menu.rb
lucywyman: Hmm, I actually think limiting this to `Data` instead of `Any` makes perfect sense - I don't understand the use case where people would want any other types, and that prevents this from having too much undefined behavior.

beechtom: See https://github.com/puppetlabs/bolt/pull/2714#discussion_r595563927

bolt-modules/prompt/lib/puppet/functions/prompt/menu.rb
lucywyman: Huh that's what I thought - I neglected to call `get_targets` so they were still just strings :P

lucywyman (commented):
- [x] Displays the options
  - [x] Handles printing mixed types (Numbers, Strings, Booleans, Array, Hash, undef)
  - [ ] Only accepts `Data` types
- [x] Selects based on user input
- [x] Reprompts user if they input an invalid option
- [x] Reprompts user if they just hit enter
- [x] Errors if the provided array is empty
- [x] Errors if the provided hash is empty
- [x] Errors if a hash is provided with duplicate keys
- [x] Doesn't error if an array is provided with duplicate values
- [x] Accepts a default with array values
  - [x] Accepts non-string default (tested with an array and hash)
- [x] Accepts a default with hash values
  - [x] Accepts non-string default
- [x] Errors if the default is not in the menu
- [x] Prints the safe_name if targets are provided

Sweet - I feel pretty strongly we should limit this to `Data` instead of `Any` for the types and return, but beyond that this is great! And happy to talk more about the types.

bolt-modules/prompt/lib/puppet/functions/prompt/menu.rb
lucywyman: Hm, what about `Variant[Target, Data]`? I just don't want to have to think about all the other datatypes and how they'll print to the console. Even Targets I'd say can be strings first and then call `get_target` after choosing.

beechtom: I think requiring targets to be strings and then calling `get_target` after is unnecessarily complicating things for users, especially if they want to pass the name of a group and select from the targets in that group.

As for using `Variant[Target, Data]` over `Any`, I'm not strictly opposed to it... I am concerned that it adds an unnecessary restriction for the function just in the off chance a user wants to select from types that don't print nicely to the console. I think maybe we can discuss with the rest of the team in standup/planning?

lucywyman: I think we decided on `Variant[Target, Data]` here for now.

PR #2715, opened 2021-03-15
(hack) Add YAML plan 'prompt' step

beechtom:
This adds a new YAML plan `prompt` step, which can be used to prompt the
user for input.

* **YAML plan `prompt` step**

  YAML plans now support a `prompt` step, which can be used to prompt
  the user for input.

---

Blocked by #2713 and #2714

### DIFF
--- a/bolt-modules/prompt/lib/puppet/functions/prompt/menu.rb
+++ b/bolt-modules/prompt/lib/puppet/functions/prompt/menu.rb
@@ -12,7 +12,7 @@ Puppet::Functions.create_function(:'prompt::menu') do
   # @param menu A list of options to choose from.
   # @param options A hash of additional options.
   # @option options [String] default The default option to return if the user does not provide
-  #   input or if standard in (stdin) is not a tty. Must be an option present in the menu.
+  #   input or if stdin is not a tty. Must be an option present in the menu.
   # @return The selected option.
   # @example Prompt the user to select from a list of options
   #   $selection = prompt::menu('Select a fruit', ['apple', 'banana', 'carrot'])
@@ -30,7 +30,7 @@ Puppet::Functions.create_function(:'prompt::menu') do
   # @param menu A hash of options to choose from, where keys are the input used to select a value.
   # @param options A hash of additional options.
   # @option options [String] default The default option to return if the user does not provide
-  #   input or if standard in (stdin) is not a tty. Must be an option present in the menu.
+  #   input or if stdin is not a tty. Must be an option present in the menu.
   # @return The selected option.
   # @example Prompt the user to select from a list of options with custom inputs
   #   $menu = { 'y' => 'yes', 'n' => 'no' }
--- a/documentation/writing_yaml_plans.md
+++ b/documentation/writing_yaml_plans.md
@@ -372,15 +372,15 @@ For example:
 ```yaml
 steps:
   - resources:
-    # This resource is type 'package' and title 'nginx'
-    - package: nginx
-      parameters:
-        ensure: latest
-    # This resource is type 'service' and title 'nginx'
-    - type: service
-      title: nginx
-      parameters:
-        ensure: running
+      # This resource is type 'package' and title 'nginx'
+      - package: nginx
+        parameters:
+          ensure: latest
+      # This resource is type 'service' and title 'nginx'
+      - type: service
+        title: nginx
+        parameters:
+          ensure: running
     targets:
       - web1.example.com
       - web2.example.com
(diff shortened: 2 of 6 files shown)


bolt-modules/prompt/lib/puppet/functions/prompt/menu.rb
```diff
@@ -0,0 +1,91 @@
...
+# until an option from the menu is selected.
+#
+# > **Note:** Not available in apply block
+Puppet::Functions.create_function(:'prompt::menu') do
+  # Select from a list of options.
+  # @param prompt The prompt to display.
+  # @param menu A list of options to choose from.
+  # @param options A hash of additional options.
+  # @option options [String] default The default option to return if the user does not provide
+  #   input or if standard in (stdin) is not a tty. Must be an option present in the menu.
```
hestonhoffman, line 15:
```suggestion
  #   input or if stdin is not a tty. Must be an option present in the menu.
```

bolt-modules/prompt/lib/puppet/functions/prompt/menu.rb
```diff
@@ -0,0 +1,91 @@
...
+    optional_param 'Hash[String[1], Any]', :options
+    return_type 'Data'
+  end
+
+  # Select from a list of options with custom inputs.
+  # @param prompt The prompt to display.
+  # @param menu A hash of options to choose from, where keys are the input used to select a value.
+  # @param options A hash of additional options.
+  # @option options [String] default The default option to return if the user does not provide
+  #   input or if standard in (stdin) is not a tty. Must be an option present in the menu.
```
hestonhoffman, line 33:
```suggestion
  #   input or if stdin is not a tty. Must be an option present in the menu.
```

schemas/bolt-yaml-plan.schema.json
```diff
@@ -180,6 +180,29 @@
...
       }
     },
+    "prompt": {
+      "description": "Displays a prompt to the user and waits for input.",
+      "type": "object",
+      "additionalProperties": false,
+      "required": ["prompt"],
+      "properties": {
+        "default": {
+          "description": "The default value to use if the user does not provide input or standard in (stdin) is not a tty.",
```
hestonhoffman, line 190:
```suggestion
          "description": "The default value to use if the user does not provide input or stdin is not a tty.",
```

bolt-modules/prompt/lib/puppet/functions/prompt.rb
```diff
@@ -11,12 +11,16 @@
   # @option options [Boolean] sensitive Disable echo back and mark the response as sensitive.
   #   The returned value will be wrapped by the `Sensitive` data type. To access the raw
   #   value, use the `unwrap` function (i.e. `$sensitive_value.unwrap`).
+  # @option options [String] default The default value to return if the user does not provide
+  #   input or if standard in (stdin) is not a tty.
```
hestonhoffman, line 15:
```suggestion
  #   input or if stdin is not a tty.
```

documentation/writing_yaml_plans.md
```diff
@@ -416,6 +416,50 @@ steps:
...
 
+### Prompt step
+
+The `prompt` step displays a prompt to the user and waits for input.
+
+Prompt steps support the following keys:
+
+| Key | Type | Description | Required |
+| --- | --- | --- | --- |
+| `default` | `String` | The default value to use if the user does not provide input or if standard in (stdin) is not a tty. | |
```
hestonhoffman, line 427:
```suggestion
| `default` | `String` | The default value to use if the user does not provide input or if standard input (stdin) is not a tty. | |
```
In this case, we should keep "standard input (stdin)"

documentation/writing_yaml_plans.md
```diff
@@ -416,6 +416,50 @@ steps:
...
+| `menu` | `Array`, `Hash` | A list of options to choose from, or a hash of options to choose from where each key is the input used to select a value. | |
+| `name` | `String` | The name of the variable to save the step result to. | |
+| `prompt` | `String` | The prompt to display to the user. | ✓ |
+| `sensitive` | `Boolean` | Whether to disable echo back and mark the response as sensitive. The returned value will be wrapped by the Sensitive data type. | |
```
hestonhoffman, line 431:
```suggestion
| `sensitive` | `Boolean` | Whether to redact the response and mark it as sensitive. Bolt wraps sensitive values in the Sensitive data type. | |
```
I think this reads a little clearer. Not sure if there is a distinction here between echoing back to logs and the result. Let me know if this needs to be reworded further. This text appears elsewhere. I can open a separate ticket to fix if you accept this change.

schemas/bolt-yaml-plan.schema.json
```diff
@@ -180,6 +180,29 @@
...
+        },
+        "name": {
+          "$ref": "#/definitions/name"
+        },
+        "prompt": {
+          "description": "The prompt to display to the user.",
+          "type": "string"
+        },
+        "sensitive": {
+          "description": "Whether to disable echo back and mark the response as sensitive. The returned value will be wrapped by the Sensitive data type.",
```
hestonhoffman, line 201: See above

lucywyman:
Sweet I think this is ready for a rebase-arooni

lucywyman:
Oh woops - it looks like there's an extra `Add default values` commit.

documentation/writing_yaml_plans.md
```diff
@@ -416,6 +416,50 @@ steps:
...
 
+### Prompt step
+
+The `prompt` step displays a prompt to the user and waits for input.
+
+Prompt steps support the following keys:
+
+| Key | Type | Description | Required |
+| --- | --- | --- | --- |
+| `default` | `String` | The default value to use if the user does not provide input or if standard input (stdin) is not a tty. | |
```
lucywyman, line 427: This can be `Variant[Data, Target]` right? If that's what the keys of the menu hash accept.

beechtom: Updated to `Array`, `Boolean`, `Hash`, `Number`, `String`

schemas/bolt-yaml-plan.schema.json
```diff
@@ -180,6 +180,29 @@
         }
       }
     },
+    "prompt": {
+      "description": "Displays a prompt to the user and waits for input.",
+      "type": "object",
+      "additionalProperties": false,
+      "required": ["prompt"],
+      "properties": {
```
lucywyman, line 188: Does this need to include "menu" too?

lucywyman (changes requested):
Ah, I see how this is complicated. If the user doesn't have a `menu` set then this has to be a string, otherwise it can be `Data` or `Target`. Those are two different functions in Puppet plans, but are one step here. I think initially that divergence didn't seem like a big deal - providing menu options as an option to `prompt` is a little cumbersome in Puppet plans but isn't here - but now that their differences are coming more to light I think that we need to unify the implementations to minimize confusion when transitioning between the two, and prevent cases like this where one function takes one type and the other takes a different type but they're both represented here.

I don't feel particularly strongly whether we move the `prompt::menu` function to be a `menu` option in Puppet plans, or if we add a new `prompt_menu` step to YAML plans in addition to `prompt`, but I think having 1 YAML step represent 2 Puppet plan functions is a bit of a recipe for disaster.

A third option would be to much more closely align the `prompt` and `prompt::menu` plan functions so that their options all have the same types and we can avoid these issues, but I think having a 1 to 1 or 2 to 2 mapping will be less risky and confusing.

PR #2716, opened 2021-03-15
(GH-2680) Keep wrapper script for tty logic grouped

donoghuc:
The use of a wrapper script for passing input to tasks over stdin avoids issues emulating a tty. The wrapper script takes care of setting the interpreter and pipeing in stdin. This commit keeps the logic to conditionally use the wrapper script grouped in a single conditional.

!no-release-note


beechtom (approved):
I swear I can code. 🙃

PR #2718, opened 2021-03-16
(docs) Correct modulepath precedence documentation

lucywyman:
The Modules overview document listed the wrong precedence of paths on
the modulepath, stating that `<project>/.modules/` would override
`<project>/modules`. This corrects the order to state that any modules
in `<project>/modules` will be used if they have the same name as
`<project>/.modules` - basically that user-created modules will override
installed modules.

!no-release-note


beechtom (approved):

hestonhoffman (commented):
Oops, I thought I approved this already!

Issue #2719, opened 2021-03-16
"Pretty" printing parameter metadata stack traces when "type" is not set in metadata


donoghuc:
## Describe the Bug
When the "type" of a parameter is not explicitly added in task metadata ruby stack traces on https://github.com/puppetlabs/bolt/blob/b3b99d432c65ba4248c2eb75be98d90e7e9f2484/lib/bolt/outputter/human.rb#L285 (`v['type'].starts_with`). 

Example:

```
[root@outstate-table modules]# cat foo/tasks/init.json 
{
  "parameters": {
    "foo": {
      "default": "bar"
    }
  }
}
[root@outstate-table modules]# bolt task show foo --modulepath . 
Traceback (most recent call last):
	7: from /opt/puppetlabs/bolt/bin/bolt:23:in `<main>'
	6: from /opt/puppetlabs/bolt/bin/bolt:23:in `load'
	5: from /opt/puppetlabs/bolt/lib/ruby/gems/2.7.0/gems/bolt-3.3.0/exe/bolt:11:in `<top (required)>'
	4: from /opt/puppetlabs/bolt/lib/ruby/gems/2.7.0/gems/bolt-3.3.0/lib/bolt/cli.rb:412:in `execute'
	3: from /opt/puppetlabs/bolt/lib/ruby/gems/2.7.0/gems/bolt-3.3.0/lib/bolt/cli.rb:556:in `show_task'
	2: from /opt/puppetlabs/bolt/lib/ruby/gems/2.7.0/gems/bolt-3.3.0/lib/bolt/outputter/human.rb:281:in `print_task_info'
	1: from /opt/puppetlabs/bolt/lib/ruby/gems/2.7.0/gems/bolt-3.3.0/lib/bolt/outputter/human.rb:281:in `each'
/opt/puppetlabs/bolt/lib/ruby/gems/2.7.0/gems/bolt-3.3.0/lib/bolt/outputter/human.rb:285:in `block in print_task_info': undefined method `start_with?' for nil:NilClass (NoMethodError)
```

PR #2720, opened 2021-03-16
(GH-2719) Correctly show task info if task parameters don't have a type

lucywyman:
Previously, Bolt assumed that task parameters included the `type` key
when checking to see if the `type` was optional to format it for
printing in `bolt task show` output. If the task included parameters
that didn't specify a type, Bolt would stacktrace. This now only adds
the `optional` parameter formatting if the parameter has a type defined
and that type includes 'Optional', otherwise it formats the type as
required. This prevents Bolt from stacktracing.

!bug

* **Don't stacktrace when showing tasks that include untyped parameters** ([#2719](https://github.com/puppetlabs/bolt/issues/2719))
  Bolt will now correctly show task details for tasks that include
  parameters that do not specify a type, instead of stacktracing.

### DIFF
--- a/lib/bolt/outputter/human.rb
+++ b/lib/bolt/outputter/human.rb
@@ -282,7 +282,7 @@ module Bolt
           pretty_params << "- #{k}: #{v['type'] || 'Any'}\n"
           pretty_params << "    Default: #{v['default'].inspect}\n" if v.key?('default')
           pretty_params << "    #{v['description']}\n" if v['description']
-          usage << if v['type'].start_with?("Optional")
+          usage << if v['type']&.start_with?("Optional")
                      " [#{k}=<value>]"
                    else
                      " #{k}=<value>"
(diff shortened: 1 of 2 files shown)


lib/bolt/outputter/human.rb
```diff
@@ -321,7 +321,7 @@ def print_plan_info(plan)
                 end
 
         plan['parameters'].each do |name, p|
-          pretty_params << "- #{name}: #{p['type']}\n"
+          pretty_params << "- #{name}: #{p['type'] || 'Any'}\n"
```
donoghuc, line 324: I think puppet does this for you (i dont think it is possible to not have a type here?)

lucywyman: Nice - this was intended for YAML plans, not Puppet plans, but it looks like it gets set there too :+1:

donoghuc (approved):

beechtom (approved):

PR #2722, opened 2021-03-16
(bug) Don't error when showing noop task info

beechtom:
This fixes a bug that caused Bolt to error when showing info for a task
that supports running in noop mode.

!bug

* **Do not error when showing 'noop' task info**

  Bolt no longer errors when printing task information for a task that
  supports running in no-operation mode.

### DIFF
--- a/lib/bolt/outputter/human.rb
+++ b/lib/bolt/outputter/human.rb
@@ -290,7 +290,7 @@ module Bolt
         end
 
         if task.supports_noop
-          usage << Bolt::Util.powershell? ? '[-Noop]' : '[--noop]'
+          usage << (Bolt::Util.powershell? ? ' [-Noop]' : ' [--noop]')
         end
 
         task_info << "\n#{task.name}"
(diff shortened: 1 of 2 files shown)


lib/bolt/outputter/human.rb
```diff
@@ -290,7 +290,7 @@ def print_task_info(task)
         end
 
         if task.supports_noop
-          usage << Bolt::Util.powershell? ? '[-Noop]' : '[--noop]'
+          usage << (Bolt::Util.powershell? ? '[-Noop]' : '[--noop]')
```
lucywyman, line 293:
```suggestion
          usage << (Bolt::Util.powershell? ? ' [-Noop]' : ' [--noop]')
```

lucywyman: Just so these don't butt right up against the last options.

lucywyman:
Oy vey - I've been thinking that we *really* should get in the habit of defining acceptance criteria. It won't catch everything like this, but this is a pretty bad bug and I think we could have caught it sooner.

PR #2723, opened 2021-03-16
(bug) Handle malformed `_error` values in task result in Orchestrator

lucywyman:
Previously, we assumed that if a task returned with an `_error` key that
it would be a hash and include the `details` key, and that that
`details` key would also have a hash value. Any of these assumptions
might be wrong. This amends the Orchestrator transport to handle
`_error` values that:
* Are not a hash, by placing the provided value under the `message` key
* Is a hash that doesn't have the `details` key by setting the details
  key to an empty hash
* Is a hash with a `details` key that is not itself a hash by not trying
  to merge our own keys into it.

!bug

* **Handle malformed `_error` values in task results in Orchestrator**

  Bolt now handles `_error` from task results in the Orchestrator
  transport when the value of the key is not a hash, does not include the
  `details` key, or the `details` key is not a hash. Previously Bolt would
  error if any of these conditions was true.

### DIFF
--- a/lib/bolt/transport/orch.rb
+++ b/lib/bolt/transport/orch.rb
@@ -59,6 +59,18 @@ module Bolt
           # the result otherwise make sure an error is generated
           if state == 'finished' || (result && result['_error'])
             if result['_error']
+              unless result['_error'].is_a?(Hash)
+                result['_error'] = { 'kind' => 'puppetlabs.tasks/task-error',
+                                     'issue_code' => 'TASK_ERROR',
+                                     'msg' => result['_error'],
+                                     'details' => {} }
+              end
+
+              result['_error']['details'] ||= {}
+              unless result['_error']['details'].is_a?(Hash)
+                deets = result['_error']['details']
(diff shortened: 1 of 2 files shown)


lib/bolt/transport/orch.rb
```diff
@@ -58,7 +58,15 @@ def process_run_results(targets, results, task_name, position = [])
...
-            if result['_error']
+            unless result['_error'].is_a?(Hash)
+              result['_error'] = { 'kind' => 'puppetlabs.tasks/task-error',
+                                   'issue_code' => 'TASK_ERROR',
+                                   'msg' => result['_error'],
+                                   'details' => {} }
+            end
+
+            result['_error']['details'] ||= {}
+            if result['_error']['details'].is_a?(Hash)
```
adreyer, line 69: We need to coerce details into a hash to match the puppet data type
https://puppet.com/docs/puppet/7.5/error-data-type.html

lucywyman:
Verified this isn't an issue over SSH, Windows, or Docker

Issue #2692, opened 2021-03-04
Don't stacktrace when calling unsupported plan functions through BoltSpec


lucywyman:
## Describe the Bug
How do we want to handle unsupported plan functions in BoltSpec, like `parallelize`? Require they be mocked?

For `run_task_with` just have an 'allow' and basically require that that be set, so that the return value is mocked. Either a limited mock, or something that's not too unwieldy (since run_task_with return objects can be pretty large)

For `parallelize`, when 'round_robin' gets called define in the mock_executor to run things in order. If it's easier to run through actual round_robin we can probably do that since tasks shouldn't be delayed.

Part of this should be verifying that there aren't any other executor functions that need to be defined or handled in the mock executor.

It would be good to see if we can automate scanning the executor and verifying that anything is either marked private or defined in the mock executor, but if that breaks tests or takes more than half a day to implement it isn't a strict requirement for this ticket.

PR #2724, opened 2021-03-17
(GH-2692) Support 'run_task_with' and 'parallelize' in BoltSpec

beechtom:
(GH-2692) Support 'parallelize' in BoltSpec

This adds support for testing plans that use the `parallelize()` plan
function in BoltSpec. The block passed to the function is evaluated
using each object sequentially instead of in parallel.

!feature

* **Test plans that use `parallelize()` plan function in BoltSpec**
  ([#2692](#2692))

  Plans that use the `parallelize()` plan function can now be tested
  with BoltSpec.

---

(GH-2692) Support 'run_task_with' in BoltSpec

This adds support for testing plans that use the `run_task_with()` plan
function to BoltSpec. When BoltSpec runs a plan with this function, it
will execute a separate task run for each target. Users making
assertions about tasks run with `run_task_with()` should make an
assertion about each individual target that the task runs against, or an
assertion about the number of times that the task is run. This includes
an example in the 'Testing plans' documentation that details how to test
these types of plans.

!feature

* **Test plans that use `run_task_with()` plan function in BoltSpec*
  ([#2692](#2692))

  Plans that use the `run_task_with()` plan function can now be tested
  with BoltSpec.

---

(GH-2692) Add test for BoltSpec mock executor method definitions

This adds a test that checks that all public methods on the
`Bolt::Executor` class are defined on the
`BoltSpec::Plans::MockExecutor` class. This is to ensure that if tests
run with BoltSpec invoke any methods on the executor that it will not
result in an "undefined method" error.

!no-release-note

---

### Acceptance criteria

- [x] Public methods on `Bolt::Executor` but not `BoltSpec::Plans::MockExecutor` result in failed spec test
- [x] Private methods on `Bolt::Executor` but not `BoltSpec::Plans::MockExecutor` do not result in failed spec test
- [x] Plans with a `run_task_with()` function can be tested in BoltSpec
- [x] Plans with a `parallelize()` function can be tested in BoltSpec

### DIFF
--- a/lib/bolt_spec/plans/mock_executor.rb
+++ b/lib/bolt_spec/plans/mock_executor.rb
@@ -91,6 +91,14 @@ module BoltSpec
         result
       end
 
+      def run_task_with(target_mapping, task, options = {}, _position = [])
+        resultsets = target_mapping.map do |target, arguments|
+          run_task([target], task, arguments, options)
+        end.compact
+
+        Bolt::ResultSet.new(resultsets.map(&:results).flatten)
+      end
+
       def download_file(targets, source, destination, options = {}, _position = [])
         result = nil
         if (doub = @download_doubles[source] || @download_doubles[:default])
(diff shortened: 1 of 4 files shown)


lib/bolt_spec/plans/mock_executor.rb
```diff
@@ -17,7 +17,7 @@ class UnexpectedInvocation < ArgumentError; end
 
     # Nothing on the executor is 'public'
     class MockExecutor
-      attr_reader :noop, :error_message, :in_parallel
+      attr_reader :noop, :error_message, :in_parallel, :transports
```
lucywyman, line 20: Sorry, I feel like I'm missing something - where does this get called?

beechtom: Spec tests.

hestonhoffman (approved):
Docs LGTM!

lucywyman (approved):
- [x] Matches expectations that a task will run with `run_task_with`
- [x] Stubs `run_task` apply to `run_task_with`
- [x] Can test plans with the `wait_until_avilable` function
- [x] Can test plans with the `parallelize` function, asserting parallelized contents
- [x] Can access variables from the outer scope
- [x] Does not modify variables in the outer scope
- [x] Returns results from a `return` in the parallelize block
- [x] Does not execute `parallelize` block in parallel

PR #2726, opened 2021-03-17
(PE-31344) Validate puppet_library plugin hook for connect

donoghuc:
The puppet_library plugin hook can be configured to run an arbitrary task (following the task plugin spec) for connect. Currently only the task plugin (ie not module plugin) is supported. For the validation plan, bolt uses the default puppet_agent module based plugin where connect will specify this same default using the task based plugin notation. The plan validates that either the default is being used or that a task based plugin is being used. The other current limitation is that for connect every target must have the same plugin configured. The validation plan validates this.

Finally, the test file for this module has been renamed so the tests contained in it will actually be run in the rake task triggered by github actions.

### DIFF
--- a/modules/puppet_connect/plans/test_input_data.pp
+++ b/modules/puppet_connect/plans/test_input_data.pp
@@ -13,6 +13,19 @@
 #
 plan puppet_connect::test_input_data(TargetSpec $targets = 'all') {
   $targs = get_targets($targets)
+  $unique_plugins = $targs.group_by |$t| {$t.plugin_hooks['puppet_library']}
+  if ($unique_plugins.keys.length > 1) {
+    out::message('Multiple puppet_library plugin hooks detected')
+    $unique_plugins.each |$plug, $target_list| {
+      $target_message = if ($target_list.length > 10) {
+                          "${target_list.length} targets"
+                        } else {
+                          $target_list.join(', ')
+                        }
+      out::message("Plugin hook ${plug} configured for ${target_message}")
+    }
+    fail_plan("The puppet_library plugin config must be the same across all targets")
+  }
   $targs.each |$target| {
     case $target.transport {
       'ssh': {
@@ -58,6 +71,15 @@ plan puppet_connect::test_input_data(TargetSpec $targets = 'all') {
       default: {
         fail_plan("Inventory contains target ${target} with unsupported transport, must be ssh or winrm")
       }
+    }
+
+    # Bolt defaults to using the "module" based form of the puppet_agent plugin. Connect defaults
+    # to using the "task" based form as *only* the task based form in supported in Connect. This check
+    # ensures that if the default is not being used, only task based plugins are allowed.
+    $plugin = $target.plugin_hooks["puppet_library"]
+    $user_configured_plugin = $plugin != { "plugin"=> "puppet_agent", "stop_service"=> true }
+    if ($user_configured_plugin and $plugin["plugin"] != "task"){
+      fail_plan("Only task plugins are acceptable for puppet_library hook")
     }
   }
   # The SSH/WinRM transports will report an 'unknown host' error for targets where
--- a/modules/puppet_connect/spec/plans/test_input_data_spec.rb
+++ b/modules/puppet_connect/spec/plans/test_input_data_spec.rb
@@ -13,10 +13,10 @@ describe 'puppet_connect::test_input_data' do
       'targets' => [
         {
           'name' => 'ssh_target',
-          'uri'  => 'ssh_uri',
+          'uri' => 'ssh_uri',
           'config' => {
             'transport' => 'ssh',
-            'ssh'  => {
+            'ssh' => {
               'load-config' => true
             }
           }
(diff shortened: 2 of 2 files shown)


steveax (approved):
👍

modules/puppet_connect/plans/test_input_data.pp
```diff
@@ -59,6 +60,16 @@
         fail_plan("Inventory contains target ${target} with unsupported transport, must be ssh or winrm")
       }
     }
+
+    $plugin = $target.plugin_hooks["puppet_library"]
+    $user_configured_plugin = $plugin != { "plugin"=> "puppet_agent", "stop_service"=> true }
```
nicklewis, line 65: A comment about how bolt-server uses a different default here would be helpful.

modules/puppet_connect/plans/test_input_data.pp
```diff
@@ -13,6 +13,7 @@
 #
 plan puppet_connect::test_input_data(TargetSpec $targets = 'all') {
   $targs = get_targets($targets)
+  $plugin_hook = $targs[0].plugin_hooks
```
nicklewis, line 16: This will fail if there are no targets. Should we guard against that?

nicklewis: Or maybe we should check this against the whole set of targets rather than comparing each individual target to the first target?

```puppet
$targs.map |$targ| { $targ.plugin_hooks }.unique.length > 1
```

modules/puppet_connect/plans/test_input_data.pp
ekinanp: Would it be worth changing the Bolt default for puppet-library to w/e bolt-server uses? So changing https://github.com/puppetlabs/bolt/blob/main/lib/bolt/plugin.rb#L145 to https://github.com/puppetlabs/bolt/blob/main/lib/bolt_server/transport_app.rb#L696. That would eliminate this check, and also the other case of what to do when the user-specified plugin _is_ the current Bolt default.

modules/puppet_connect/plans/test_input_data.pp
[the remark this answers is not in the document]
ekinanp: It also might be a good idea to add validation for the task params. Thinking a Puppet function that does https://github.com/puppetlabs/bolt/blob/main/lib/bolt/plugin.rb#L87 (unless there's already a way to do that with existing Bolt code).

That can definitely be a separate piece of work though.

modules/puppet_connect/plans/test_input_data.pp
```diff
@@ -59,6 +60,16 @@
...
     }
+
+    $plugin = $target.plugin_hooks["puppet_library"]
+    $user_configured_plugin = $plugin != { "plugin"=> "puppet_agent", "stop_service"=> true }
+
+    if ($user_configured_plugin and $plugin["plugin"] != "task"){
+      fail_plan("Only task plugins are acceptable for puppet_library hook")
+    }
+    if ($target.plugin_hooks != $plugin_hook){
+      fail_plan("The puppet_library plugin config must be the same across all targets")
```
ekinanp, line 71: We should probably include as much detail as possible and that is useful here, like in the Orch-side validation. At a minimum, I think the non-unique target's name would be good.

modules/puppet_connect/plans/test_input_data.pp
nicklewis: The issue is that what bolt-server is doing is technically less correct. The proper thing to do is use `plugin => puppet_agent` which will then reference [`bolt-plugin.json`](https://github.com/puppetlabs/puppetlabs-puppet_agent/blob/main/bolt_plugin.json) to determine which task plugin corresponds to the `puppet_library` hook. For bolt-server, we just hardcode that task to `puppet_agent::install` and 🤞 it doesn't change.

ekinanp: Right, so I wonder if we should now hard-code that default to `puppet_agent::install` ? Sounds like no.

modules/puppet_connect/plans/test_input_data.pp
```diff
@@ -13,6 +13,19 @@
 #
 plan puppet_connect::test_input_data(TargetSpec $targets = 'all') {
   $targs = get_targets($targets)
+  $unique_plugins = $targs.group_by |$t| {$t.plugin_hooks['puppet_library']}
+  if ($unique_plugins.keys.length > 1) {
+    out::message('Multiple puppet_library plugin hooks detected')
+    $unique_plugins.each |$plug, $target_list| {
+      $target_message = if ($target_list.length > 10) {
+                          $target_list.length
```
ekinanp, line 21: I think this means to condense the list?

nicklewis: I think it should be `"${target_list.length} targets"` and remove the word `targets` from the message below. Otherwise it will be "Plugin hook [...] configured for [foo.example.com, bar.example.com] targets" which is a weird one.

modules/puppet_connect/plans/test_input_data.pp
```diff
@@ -13,6 +13,19 @@
...
+                        } else {
+                          $target_list
```
nicklewis, line 23: `$target_list.join(", ")`?

barriserloth (approved):
LGTM, won't merge in case Nick is looking at this since he's assigned atm

nicklewis (approved):

Issue #2737, opened 2021-03-24
Warn when users try to set a fact with a dot in the name


lucywyman:
## Use Case
When users:
- Call `add_fact()` from a plan and specify a fact with a dot in the name
- Are setting facts in the inventory and try to set a fact with a dot in the name
- Are running with an older version of Facter on the agent and we get a dotted fact back from Facter (TODO: which version?) if they try to set a fact with a name that has a '.' character in it (solve in target.add_facts() method)

In any of these cases, we should raise a warning and instruct that using dots in the fact name are deprecated (we won't implement the structured fact magic that Facter does - we should probably document that somewhere, not sure where is most helpful). Include fact that failed in the message. Maybe Developer Update is a good place to document it, and then create a short link and include in the warning.

PR #2741, opened 2021-03-25
(GH-2737) Deprecate dotted fact names

beechtom:
This updates Bolt to log a deprecation warning when it detects a target
that has facts with dotted names. This warning is logged whether a
target is loaded with a dotted fact name (either in the inventory or
with `Target.new`) or when adding facts to a target (with
`$target.add_facts`).

!deprecation

* **Deprecate dotted fact names**
  ([#2737](https://github.com/puppetlabs/bolt/issues/2737))

  Dotted fact names (e.g. `foo.bar`) are now deprecated. Bolt issues a
  deprecation warning if it detects a target is loaded with these facts
  or has them added during a plan run.

--- 

### TODO

- [ ] Create pup.pt short link once approved

---

### Acceptance criteria

- [x] Warns when target has dotted facts in inventory
- [x] Warns when facter returns dotted facts, such as during an `apply_prep` (only applies to Facter 2 & 3)
- [x] Warns when adding facts in a plan with `$target.add_facts`

### DIFF
--- a/documentation/developer_updates.md
+++ b/documentation/developer_updates.md
@@ -3,6 +3,39 @@
 Find out what the Bolt team is working on and why we're making the decisions
 we're making.
 
+## March 2021
+
+### Deprecating dotted fact names
+
+With the release of Facter 4, dotted fact names are automatically converted to
+structured facts. For example, the custom fact `role.name = server` becomes the
+following structured fact:
+
+```
+{
+  "role" => {
+    "name" => "server"
+  }
+}
+```
+
+This is a significant change from earlier versions of Facter, which permitted
+dotted fact names.
+
+Because Bolt allows you to set facts in the inventory file or during a plan run,
+and uses Facter directly (such as when running the `apply_prep` function) and
+indirectly (such as when making a PuppetDB query), we've updated Bolt to warn
+you when it detects a dotted fact name.
+
+If Bolt detects that a target has a dotted fact name set in either the inventory
+or during a plan run, it displays a deprecation warning. Bolt does not
+automatically convert these facts to structured facts, which is inconsistent
+behavior with Facter 4.
+
+To suppress these deprecation warnings, you can either update your facts or add
+the `dotted_fact_name` ID to the `disable-warnings` option in your project
+configuration.
+
 ## November 2020
 
 ### Changes coming in Bolt 3.0
--- a/lib/bolt/inventory/target.rb
+++ b/lib/bolt/inventory/target.rb
@@ -92,6 +92,7 @@ module Bolt
       end
 
       def add_facts(new_facts = {})
+        validate_fact_names(new_facts)
         @facts = Bolt::Util.deep_merge(@facts, new_facts)
       end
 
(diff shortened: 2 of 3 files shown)


lib/bolt/inventory/target.rb
```diff
@@ -153,9 +154,24 @@ def validate
...
 
+        validate_fact_names(@facts)
+
         transport_config
       end
 
+      # Validate fact names and issue a deprecation warning if any fact names have a dot.
+      #
+      private def validate_fact_names(facts)
+        if (dotted = facts.keys.select { |name| name.include?('.') }).any?
```
nicklewis, line 165: I think we need to also dig into hashes and look for hash keys with `.`

nicklewis: Apparently this is not the case.

beechtom: Phew. I wasn't too sure at first, but the doc for this issue suggested that we had internal facts with dotted names that should stay that way.

lucywyman: Are only top-level dots considered structured facts?

lib/bolt/inventory/target.rb
```diff
@@ -153,9 +154,24 @@ def validate
           raise Bolt::UnknownTransportError.new(transport, uri)
         end
 
+        validate_fact_names(@facts)
```
lucywyman, line 157: This needs to be `facts` not `@facts` so that facts set at the group level are also caught, not just those set directly on the target.

lucywyman:
- [x] Warns when user calls `add_facts` with a fact with a dot in the name
- [x] Doesn't warn if the fact has a dot in the value
- [x] Warns when fact is set in the inventory on a target
- [x] Warns when fact is set in the inventory on a group
- [x] Doesn't warn if a fact is set on a target but the target is never loaded
- [x] Warns when facter returns a custom dotted fact and is facter 2 or 3
  - [x] With apply_prep
  - [x] With the facts plan
- [x] Doesn't warn when facter returns a custom dotted fact and is facter 4
- [x] Only warns once if the fact is set multiple places

lib/bolt/inventory/target.rb
lucywyman: Answer: yes, nested facts with dots in them aren't treated as structured facts with Facter 4. Do we still want to discourage people from using them in general though? I'm not sure if that will ever change.

lucywyman (approved):
I'll wait for Heston to review the Developer update, but LGTM!

documentation/developer_updates.md
```diff
@@ -3,6 +3,39 @@
...
+This is a significant change from earlier versions of Facter, which permitted
+dotted fact names.
+
+Because Bolt allows you to set facts in the inventory file or during a plan run,
+and uses Facter directly (such as when running the `apply_prep` function) and
+indirectly (such as when making a PuppetDB query), we've updated Bolt to warn
+you when it detects a dotted fact name.
+
+If Bolt detects that a target has a dotted fact name set in either the inventory
+or during a plan run, it will display a deprecation warning. Bolt will not
```
hestonhoffman, line 31:
```suggestion
If Bolt detects that a target has a dotted fact name set in either the inventory
or during a plan run, it displays a deprecation warning. Bolt does not
```

hestonhoffman (approved):
Couple small changes, but LGTM!

PR #2744, opened 2021-03-26
(PE-31347) Load cloud inventory plugins in bolt-server

nicklewis:
This slightly refactors the way that plugins are loaded so that the
`add_module_plugin` method can be called explicitly even if plugin
loading is disabled. Specifically, the `add_ruby_plugin` and
`add_module_plugin` methods now longer verify that the plugin exists or
that plugin loading is enabled. Those steps are now handled exclusively by
`by_name` and `get_hook`, which are the entry points into the plugin
system.

The resulting behavior is the same as before:
  1. If a plugin doesn't exist, `by_name` won't raise an error and
     `get_hook` will. This allows config to be set for a non-existent
     plugin as that will only call `by_name`, which will do nothing. If
     the plugin is later *used*, `get_hook` will raise an error.
  2. If a plugin does exist but loading is disabled, `by_name` will raise an
     error right away.

### DIFF
--- a/lib/bolt/plugin.rb
+++ b/lib/bolt/plugin.rb
@@ -178,8 +178,6 @@ module Bolt
     end
 
     def add_ruby_plugin(plugin_name)
-      raise PluginError::LoadingDisabled, plugin_name unless @load_plugins
-
       cls_name = Bolt::Util.snake_name_to_class_name(plugin_name)
       filename = "bolt/plugin/#{plugin_name}"
       require filename
@@ -203,8 +201,6 @@ module Bolt
       }
 
       mod = modules[plugin_name]
-      raise PluginError::Unknown, plugin_name unless mod&.plugin?
-      raise PluginError::LoadingDisabled, plugin_name unless @load_plugins
 
       plugin = Bolt::Plugin::Module.load(mod, opts)
       add_plugin(plugin)
@@ -224,6 +220,12 @@ module Bolt
       end
     end
 
+    def known_plugin?(plugin_name)
+      @plugins.include?(plugin_name) ||
+        RUBY_PLUGINS.include?(plugin_name) ||
+        (modules.include?(plugin_name) && modules[plugin_name].plugin?)
+    end
+
     def get_hook(plugin_name, hook)
       plugin = by_name(plugin_name)
       raise PluginError::Unknown, plugin_name unless plugin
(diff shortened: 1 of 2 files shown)


lib/bolt/plugin.rb
```diff
@@ -235,16 +237,16 @@ def get_hook(plugin_name, hook)
 
     # Calling by_name or get_hook will load any module based plugin automatically
     def by_name(plugin_name)
-      return @plugins[plugin_name] if @plugins.include?(plugin_name)
-      begin
-        if RUBY_PLUGINS.include?(plugin_name)
+      if known_plugin?(plugin_name)
+        if @plugins.include?(plugin_name)
```
barriserloth, line 241: This is a question for my own understanding more so than a review comment: Is the reason this `@plugins.include?` happens before the `!@load_plugins` check to allow us to load plugins even when it's technically disabled?

barriserloth: Does this have implications for users where they could end up loading plugins even with plugin loading disabled? I guess, is plugin loading something that they even have control over in the first place?

nicklewis: No, this only affects bolt-server's ability to explicitly load plugins. From a bolt user's point of view, all plugins are loaded magically and there's no way in bolt to disable plugin loading at all.

barriserloth (approved):

lib/bolt_server/transport_app.rb
```diff
@@ -703,6 +703,9 @@ def make_winrm_target(target_hash)
           connect_plugin = BoltServer::Plugin::PuppetConnectData.new(body['puppet_connect_data'])
           plugins = Bolt::Plugin.setup(context[:config], context[:pal], load_plugins: false)
           plugins.add_plugin(connect_plugin)
+          %w[aws_inventory azure_inventory gcloud_inventory].each do |plugin_name|
+            plugins.add_module_plugin(plugin_name) if plugins.modules.include?(plugin_name)
```
ekinanp, line 707: Oh, what happens if the user overrides one of these modules, the override is no longer a module plugin, and they're not using that module as an inventory plugin [meaning the override makes sense]? E.g. I have an `aws_inventory` module that is not a module plugin and I am not using the `aws_inventory` plugin in my inventory file. I think in this case, `add_module_plugin` would error whereas in Bolt, I think we'd just pass-thru b/c dynamic loading would ensure that the plugin's never loaded and hence `add_module_plugin(aws_inventory)` is never called.

If that is a valid case, do we want to handle it? If so, how?

Do we want to handle that here

nicklewis: I think that means this check should change from `plugins.modules.include?(plugin_name)` to `plugins.modules.include?(plugin_name) && plugins.modules[plugin_name].plugin?`. Or just `plugins.known_plugin?(plugin_name)` because we _know_ these aren't Ruby plugins.

nicklewis: Updated!

ekinanp (approved):

donoghuc (approved):
Resolved some aws targets in Connect with this :champagne:

PR #2746, opened 2021-03-29
(bug) Allow 'version' key in inventory files

lucywyman:
When running `bolt project migrate` we don't remove the `version` key if
the value is `2`, but we still raise a warning that the `version` key
isn't recognized. This amends the inventoryfile schema to include the
`version` key so that users don't see a warning.

!bug

* **Allow 'version' key in inventory files**

  Bolt now recognizes the `version` configuration in an inventoryfile
  and doesn't raise a warning that the key is unknown.

### DIFF
--- a/lib/bolt/inventory/options.rb
+++ b/lib/bolt/inventory/options.rb
@@ -14,6 +14,7 @@ module Bolt
         plugin_hooks
         targets
         vars
+        version
       ].freeze
 
       # Definitions used to validate the data.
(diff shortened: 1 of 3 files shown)


lib/bolt/inventory/options.rb
```diff
@@ -124,6 +125,12 @@ module Options
           description: "A map of variables for the group or target.",
           type: Hash,
           _plugin: true
+        },
+        "version" => {
+          description: "The version of the inventoryfile.",
```
beechtom, line 130:
```suggestion
          description: "The version of the inventory file.",
```

Docs have standardized on two separate words for 'inventory file'.

lib/bolt/inventory/options.rb
```diff
@@ -124,6 +125,12 @@ module Options
...
+          type: Integer,
+          _plugin: false,
+          _example: 2
```
beechtom, line 133:
```suggestion
          _example: 2,
          _default: 2
```

beechtom (approved):

PR #2751, opened 2021-03-30
(hack) Improve 'inventory show' output

beechtom:
This updates the output for `bolt inventory show` and
`Get-BoltInventory` to better match the other `show` commands. Targets,
the inventory source, and the target count are now listed under
colorized headings.

This also updates the `--detail` view so that targets are listed in
alphabetical order and use each target's name as a heading for easy
reading. The target's configuration and data is also now displayed in
YAML in the human outputter, to better match the format of the inventory
file. The `--detail` view also now includes the same inventory source
and target count that are in the undetailed view.

!feature

* **Improved output for `inventory show`**

  The `bolt inventory show` and `Get-BoltInventory` command now display
  `human` output in the same format as other `show` commands.

### DIFF
--- a/lib/bolt/outputter/human.rb
+++ b/lib/bolt/outputter/human.rb
@@ -452,29 +452,77 @@ module Bolt
         targets += target_list[:inventory].map { |target| [target.name, nil] }
         targets += target_list[:adhoc].map { |target| [target.name, adhoc] }
 
-        if targets.any?
-          @stream.puts format_table(targets, 0, 2)
-          @stream.puts
-        end
+        info = +''
 
-        @stream.puts "INVENTORY FILE:"
-        if File.exist?(inventoryfile)
-          @stream.puts inventoryfile
-        else
-          @stream.puts wrap("Tried to load inventory from #{inventoryfile}, but the file does not exist")
-        end
+        # Add target list
+        info << colorize(:cyan, "Targets\n")
+        info << if targets.any?
+                  format_table(targets, 2, 2).to_s
+                else
+                  indent(2, 'No targets')
+                end
+        info << "\n\n"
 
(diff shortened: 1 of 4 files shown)


lucywyman (approved):
- [x] Shows targets that are in the inventory and not in the inventory
- [x] Expands group names, queries, `--rerun`
- [x] Prints 'no targets' if the query doesn't match any targets
- [x] Does not change the behavior of the JSON format
- [x] Integrates config from user-level `bolt-defaults.yaml` file

So pretty! This is so much more readable.

lib/bolt/outputter/human.rb
```diff
@@ -452,29 +452,77 @@ def print_targets(target_list, inventoryfile)
...
+
+        print_inventory_summary(
+          target_list[:inventory].count,
+          target_list[:adhoc].count,
+          inventoryfile
+        )
+      end
+
+      def print_target_info(target_list, inventoryfile)
+        adhoc_targets     = target_list[:adhoc].map(&:safe_name).to_set
```
lucywyman, line 476: Hm, admittedly I think this is kind of moot since the non-safe-name will get printed right below, unless the idea is to be more readable?
```
lucy ~/githubs/bolt ((f4d3f755...)) $ be bolt inventory show -t all --detail
bolt@localhost:20024
  name: bolt:bolt@localhost:20024
  uri: bolt:bolt@localhost:20024
```

lucywyman: :chefs-kiss: Yeah, I think that's what users will be expecting to look for. Can you also update the commit message, just we don't also confuse our future selves? Then this is GtG!

beechtom: done

PR #2753, opened 2021-03-31
(maint) Address puppetlabs_spec_helper deprecation warning

lucywyman:
The `puppetlabs_spec_helper` ruby library we use as part of testing
embedded Bolt modules was raising a deprecation warning that the default
mocking framework was `mocha` but that to silence the warning the tests
themselves should specify which framework was being used. While I'd
honestly love to switch to `rspec` it's a lot of work to do so for
relatively little gain. For now setting the configuration to `mocha` to
silence the warning is easiest.

!no-release-note

### DIFF
--- a/bolt-modules/boltlib/spec/spec_helper.rb
+++ b/bolt-modules/boltlib/spec/spec_helper.rb
@@ -7,4 +7,7 @@ require 'bolt/pal'
 # so we get task loaders.
 Puppet[:tasks] = true
 Bolt::PAL.load_puppet
+RSpec.configure do |c|
+  c.mock_with :mocha
+end
 require 'puppetlabs_spec_helper/module_spec_helper'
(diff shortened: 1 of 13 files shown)


beechtom (changes requested):
`bolt_spec_spec` and the modules in `modules/` also need this config change.

lucywyman:
Good call, updated

beechtom (changes requested):
https://github.com/puppetlabs/bolt/blob/main/bolt_spec_spec/spec/functions/with_datatype_spec.rb#L3

RSpec needs to be configured before `puppetlabs_spec_helper/module_spec_helper` is loaded, so this line should be moved after `require 'spec_helper'` or to the end of `spec_helper.rb` similar to the other modules.

beechtom (approved):

PR #2754, opened 2021-03-31
(hack) Improve plan and task info output

beechtom:
This updates the output for `bolt plan|task show <name>` to better align
with the formatting in other `show` commands. Information is now listed
under colorized headings and parameter information is colorized and
spaced out to improve readability. This also wraps the usage string and
descriptions so each line is a maximum of 80 characters long, which
makes it easier to read long descriptions.

!feature

* **Improved plan and task information output**

  The `bolt plan|task show <name>` and `Get-Bolt(Plan|Task) -Name
  <name>` commands now display `human` output in a similar format to
  other `show` commands.

### DIFF
--- a/lib/bolt/outputter/human.rb
+++ b/lib/bolt/outputter/human.rb
@@ -6,6 +6,7 @@ module Bolt
   class Outputter
     class Human < Bolt::Outputter
       COLORS = {
+        dim:    "2", # Dim, the other color of the rainbow
         red:    "31",
         green:  "32",
         yellow: "33",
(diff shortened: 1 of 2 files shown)


lucywyman (approved):
Looks great when:
  - [x] Task or plan has a description
  - [x] Task or plan doesn't have a description
  - [x] Includes noop flag if the task supports noop
  - [x] Doesn't include noop flag if the task doesn't support noop
  - [x] Marks optional parameters as optional in usage
  - [x] With parameters
    - [x] With a default
    - [x] With a description
    - [x] With a type
    - [x] With all 3
    - [x] With extra long description that wraps
  - [x] Without parameters
  - [x] With the `puppet_agent::install` task
  - [x] With complex peadm plans, like `peadm::action::configure`

This looks so much better! Dramatic improvement :clap: . My only concern that sticks out is the dim text - I tried a few other colors, and I think blue (`34`), magenta (`35`), and light blue (`94`) don't look too bad, but we can also wait and see if folks think it's hard to read before changing.

Issue #2756, opened 2021-03-31
Don't stack trace when running 'bolt module' without a bolt-project.yaml


lucywyman:
## Describe the Bug
Running 'bolt module add' subcommand without a bolt-project.yaml results in a stacktrace:
```
Traceback (most recent call last):
	5: from /opt/puppetlabs/bolt/bin/bolt:23:in `<main>'
	4: from /opt/puppetlabs/bolt/bin/bolt:23:in `load'
	3: from /opt/puppetlabs/bolt/lib/ruby/gems/2.7.0/gems/bolt-3.5.0/exe/bolt:11:in `<top (required)>'
	2: from /opt/puppetlabs/bolt/lib/ruby/gems/2.7.0/gems/bolt-3.5.0/lib/bolt/cli.rb:529:in `execute'
	1: from /opt/puppetlabs/bolt/lib/ruby/gems/2.7.0/gems/bolt-3.5.0/lib/bolt/cli.rb:792:in `add_project_module'
/opt/puppetlabs/bolt/lib/ruby/gems/2.7.0/gems/bolt-3.5.0/lib/bolt/cli.rb:812:in `assert_project_file': undefined method `config_file' for #<Bolt::Project:0x00007f8568987598> (NoMethodError)
```

PR #2758, opened 2021-03-31
(GH-2756) Do not stack trace when missing project config file

beechtom:
This fixes a bug that arises from running the `bolt module add|install`
command when a project does not have a `bolt-project.yaml` file.
Previously, Bolt would call the undefined method
`Bolt::Project.config_file`, result in a stacktrace. Since this method
was removed in Bolt 3.0, the condition that called this method has been
removed.

!bug

* **Do not stack trace when missing project configuration file**
  ([#2756](https://github.com/puppetlabs/bolt/issues/2756))

  Bolt no longer stack traces when installing modules if the project
  does not have a `bolt-project.yaml` configuration file.

### DIFF
--- a/lib/bolt/cli.rb
+++ b/lib/bolt/cli.rb
@@ -822,15 +822,10 @@ module Bolt
     #
     def assert_project_file(project)
       unless project.project_file?
-        msg = if project.config_file.exist?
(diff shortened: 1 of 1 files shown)


lucywyman (approved):
- [x] No other calls to `.config_file` in the codebase
  - [x] No other methods removed during 3.0 release are still around either
- [x] Raises warning when no project file exists to write to

Issue #2731, opened 2021-03-19
Fallback to finding files in `scripts/` if not found in `files/`


lucywyman:
## Use Case
Puppet should now look for files in the `scripts/` directory if they aren't found in the `files/` directory. This behavior should only be enabled if the feature flag in #2730 is set.

### Questions
Should this also be gated on `Puppet[:tasks]`, or another setting to ensure this only ever works in Bolt?
Should this only fallback if we know we're trying to run a script (either through the function or the CLI)? It seems like 'no', but I just want to be sure.
Are there any other cases where we shouldn't do this?
To avoid unexpected behavior in Puppet (like loading a script when it's not expected) we'll want to add the `find_file_or_script` method as a helper in Bolt that deals with loading different syntaxes correctly.

All the places to verify loading or not-loading works as expected:
- run_script, file_upload, other file loading plan functions
- File resources should not support this
- bolt script run command line, bolt file upload, other CLI commands 
- bolt-module functions called from an apply

PR #2760, opened 2021-03-31
(GH-2731) Load files from specified Puppet paths

lucywyman:
This introduces a new syntax for loading files from Puppet modules.
Previously, files in Puppet modules could be loaded if the file existed
in the `files/` subdirectory of a module on the modulepath, and the user
provided `mymodule/file.sh` to load a file located at
`mymodule/files/file.sh`.

The new loading behavior is gated behind a new project-level
configuration option. The option is specified under the `future` key
which is used enable different features that may introduce breaking
changes in Bolt. Right now the only suboption that `future` has is
`file_paths`, which enables the following loading behavior.

The new loading behavior loads with the following precedence depending
on what syntax is provided. Note that files can now be loaded from the
`scripts/` directory of a module in addition to the `files/` directory:

Provided: **module/myscript.sh**
- Load from `files/myscript.sh`

Provided: **module/scripts/myscript.sh**
- Load from `files/scripts/myscript.sh`
- Fall back to `scripts/myscript.sh`

Provided: **module/files/myscript.sh**
- Load from `files/files/myscript.sh`
- Fall back to `files/myscript.sh`

!feature

* **Load files from specified Puppet paths** ([#2731](https://github.com/puppetlabs/bolt/issues/2731))

  If the project-level `future.file_paths` configuration is enabled,
  Puppet files can be loaded using the new loading syntax. For more information see https://pup.pt/bolt-loading-files.

### DIFF
--- a/bolt-modules/boltlib/lib/puppet/functions/run_script.rb
+++ b/bolt-modules/boltlib/lib/puppet/functions/run_script.rb
@@ -108,7 +108,11 @@ Puppet::Functions.create_function(:run_script, Puppet::Functions::InternalFuncti
     # Send Analytics Report
     executor.report_function_call(self.class.name)
 
-    found = Puppet::Parser::Files.find_file(script, scope.compiler.environment)
+    future = executor&.future || {}
+    fallback = future.fetch('file_paths', false)
+
+    # Find the file path if it exists, otherwise return nil
+    found = Bolt::Util.find_file_from_scope(script, scope, fallback)
     unless found && Puppet::FileSystem.exist?(found)
       raise Puppet::ParseErrorWithIssue.from_issue_and_stack(
         Puppet::Pops::Issues::NO_SUCH_FILE_OR_DIRECTORY, file: script
--- a/bolt-modules/boltlib/lib/puppet/functions/upload_file.rb
+++ b/bolt-modules/boltlib/lib/puppet/functions/upload_file.rb
@@ -70,7 +70,11 @@ Puppet::Functions.create_function(:upload_file, Puppet::Functions::InternalFunct
     # Send Analytics Report
     executor.report_function_call(self.class.name)
 
-    found = Puppet::Parser::Files.find_file(source, scope.compiler.environment)
+    future = executor&.future || {}
+    fallback = future.fetch('file_paths', false)
+
+    # Find the file path if it exists, otherwise return nil
+    found = Bolt::Util.find_file_from_scope(source, scope, fallback)
     unless found && Puppet::FileSystem.exist?(found)
       raise Puppet::ParseErrorWithIssue.from_issue_and_stack(
         Puppet::Pops::Issues::NO_SUCH_FILE_OR_DIRECTORY, file: source
--- a/bolt-modules/file/lib/puppet/functions/file/exists.rb
+++ b/bolt-modules/file/lib/puppet/functions/file/exists.rb
@@ -12,14 +12,20 @@ Puppet::Functions.create_function(:'file::exists', Puppet::Functions::InternalFu
   #   file::exists('example/VERSION')
   dispatch :exists do
     scope_param
-    required_param 'String', :filename
+    required_param 'String[1]', :filename
     return_type 'Boolean'
   end
 
   def exists(scope, filename)
     # Send Analytics Report
-    Puppet.lookup(:bolt_executor) {}&.report_function_call(self.class.name)
-    found = Puppet::Parser::Files.find_file(filename, scope.compiler.environment)
+    executor = Puppet.lookup(:bolt_executor) {}
+    executor&.report_function_call(self.class.name)
+
+    future = executor&.future || Puppet.lookup(:future) || {}
+    fallback = future.fetch('file_paths', false)
+
+    # Find the file path if it exists, otherwise return nil
+    found = Bolt::Util.find_file_from_scope(filename, scope, fallback)
     found ? Puppet::FileSystem.exist?(found) : false
   end
 end
--- a/bolt-modules/file/lib/puppet/functions/file/read.rb
+++ b/bolt-modules/file/lib/puppet/functions/file/read.rb
@@ -11,7 +11,7 @@ Puppet::Functions.create_function(:'file::read', Puppet::Functions::InternalFunc
   #   file::read('example/VERSION')
   dispatch :read do
     scope_param
-    required_param 'String', :filename
+    required_param 'String[1]', :filename
     return_type 'String'
   end
 
@@ -20,7 +20,11 @@ Puppet::Functions.create_function(:'file::read', Puppet::Functions::InternalFunc
     executor = Puppet.lookup(:bolt_executor) {}
     executor&.report_function_call(self.class.name)
 
-    found = Puppet::Parser::Files.find_file(filename, scope.compiler.environment)
+    future = executor&.future || Puppet.lookup(:future) || {}
+    fallback = future.fetch('file_paths', false)
+
+    # Find the file path if it exists, otherwise return nil
+    found = Bolt::Util.find_file_from_scope(filename, scope, fallback)
     unless found && Puppet::FileSystem.exist?(found)
       raise Puppet::ParseErrorWithIssue.from_issue_and_stack(
         Puppet::Pops::Issues::NO_SUCH_FILE_OR_DIRECTORY, file: filename
--- a/bolt-modules/file/lib/puppet/functions/file/readable.rb
+++ b/bolt-modules/file/lib/puppet/functions/file/readable.rb
@@ -12,15 +12,20 @@ Puppet::Functions.create_function(:'file::readable', Puppet::Functions::Internal
   #   file::readable('example/VERSION')
   dispatch :readable do
     scope_param
-    required_param 'String', :filename
+    required_param 'String[1]', :filename
     return_type 'Boolean'
   end
 
   def readable(scope, filename)
     # Send Analytics Report
-    Puppet.lookup(:bolt_executor) {}&.report_function_call(self.class.name)
+    executor = Puppet.lookup(:bolt_executor) {}
+    executor&.report_function_call(self.class.name)
 
-    found = Puppet::Parser::Files.find_file(filename, scope.compiler.environment)
+    future = executor&.future || Puppet.lookup(:future) || {}
+    fallback = future.fetch('file_paths', false)
+
+    # Find the file path if it exists, otherwise return nil
+    found = Bolt::Util.find_file_from_scope(filename, scope, fallback)
     found ? File.readable?(found) : false
   end
 end
--- a/lib/bolt/cli.rb
+++ b/lib/bolt/cli.rb
@@ -563,7 +563,8 @@ module Bolt
             when 'command'
               executor.run_command(targets, options[:object], executor_opts)
             when 'script'
-              script_path = find_file(options[:object])
+              script_path = find_file(options[:object], executor.future&.fetch('file_paths', false))
+              validate_file('script', script_path)
               executor.run_script(targets, script_path, options[:leftovers], executor_opts)
             when 'task'
               pal.run_task(options[:object],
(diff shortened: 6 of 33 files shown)


lucywyman:
Just a quick note this can't go out until 7.6.0 is released later this month.

lib/bolt/util.rb
```diff
@@ -85,6 +85,38 @@ def first_run?
         Bolt::Config.user_path && !File.exist?(first_runs_free)
       end
 
+      def find_file_in_mod(mod, module_file)
+        subdir = module_file.split(File::SEPARATOR).first
```
nicklewis, line 89: You can use something like `subdir, relative_path = module_file.split(File::SEPARATOR, 2)` to split the path in only one place.

lib/bolt/util.rb
```diff
@@ -85,6 +85,38 @@ def first_run?
...
+        case subdir
+        when 'scripts'
+          script_path = module_file.split(File::SEPARATOR).drop(1).join(File::SEPARATOR)
```
nicklewis, line 92: `script_path` will be `""` if the original path was just `mymodule/scripts` (or `nil` if you use the `split(..., 2)` suggestion above) . In that case, we probably shouldn't fall back to `mod.script`. Ditto for the `files` case.

lib/bolt/util.rb
```diff
@@ -85,6 +85,44 @@ def first_run?
...
+        when 'files'
+          unless relative_path.nil?
+            mod.file(module_file) || mod.file(File.join(mod, relative_path))
+          else
+            mod.file(module_file)
+          end
+        else
+          mod.file(module_file)
+        end
+      end
```
beechtom, line 106: This method isn't the easiest to follow. I think this stems from a couple things:

- `module_file` is always split, which results in `subdir` sometimes being the name of a subdir, while other times it ends up being the filename. Just from a naming standpoint, it seems like `subdir` should be `nil` if the path doesn't have at least two parts.
- Nested logic. `||` operators inside `unless` statements inside a `case` statement make this logically a bit hard to follow. Part of the extra complexity comes from checking `relative_path.nil?`, but I'm not sure if there's a way around that. Maybe the way to approach that is with...
- Comments. It would be helpful to have comments describing what filepath is being checked in each statement and why.

lib/bolt/util.rb
```diff
@@ -85,6 +85,44 @@ def first_run?
         Bolt::Config.user_path && !File.exist?(first_runs_free)
       end
 
+      def find_file_in_mod(mod, module_file)
```
beechtom, line 88: This method, `find_module_manually`, and `search_module` should have comments describing what the method does. While they're not the most complex methods, I think it's preferable to have a couple sentences describing something instead of requiring someone to parse the code.

lib/bolt/util.rb
```diff
@@ -85,6 +85,44 @@ def first_run?
...
+            mod.file(module_file)
+          end
+        when 'files'
+          unless relative_path.nil?
+            mod.file(module_file) || mod.file(File.join(mod, relative_path))
+          else
+            mod.file(module_file)
+          end
+        else
+          mod.file(module_file)
```
beechtom, line 104:
```suggestion
          mod.file(module_file) || mod.script(module_file)
```

I think this needs to fallback to `scripts`?

bolt-modules/boltlib/lib/puppet/functions/run_script.rb
```diff
@@ -108,7 +110,13 @@ def run_script_with_description(scope, script, targets, description = nil, optio
...
     executor.report_function_call(self.class.name)
 
-    found = Puppet::Parser::Files.find_file(script, scope.compiler.environment)
+    found = if executor.future&.fetch('file_paths', false)
+              Puppet::Parser::Files.find_in_module(script, scope.compiler.environment) do |mod, module_file|
+                Bolt::Util.find_file_in_mod(mod, module_file)
+              end
+            else
+              Puppet::Parser::Files.find_file(script, scope.compiler.environment)
+            end
```
beechtom, line 119: A comment describing what's happening here would be helpful.

bolt-modules/boltlib/lib/puppet/functions/upload_file.rb
```diff
@@ -70,7 +71,13 @@ def upload_file_with_description(scope, source, destination, targets, descriptio
...
     executor.report_function_call(self.class.name)
 
-    found = Puppet::Parser::Files.find_file(source, scope.compiler.environment)
+    found = if executor.future&.fetch('file_paths', false)
+              Puppet::Parser::Files.find_in_module(source, scope.compiler.environment) do |mod, module_file|
+                Bolt::Util.find_file_in_mod(mod, module_file)
+              end
+            else
+              Puppet::Parser::Files.find_file(source, scope.compiler.environment)
+            end
```
beechtom, line 80: Ditto on the comment.

bolt-modules/file/lib/puppet/functions/file/exists.rb
```diff
@@ -18,8 +18,16 @@
...
+    executor = Puppet.lookup(:bolt_executor)
+    executor.report_function_call(self.class.name)
+
+    found = if executor.future&.fetch('file_paths', false)
+              Puppet::Parser::Files.find_in_module(filename, scope.compiler.environment) do |mod, module_file|
+                Bolt::Util.find_file_in_mod(mod, module_file)
+              end
+            else
+              Puppet::Parser::Files.find_file(filename, scope.compiler.environment)
+            end
```
beechtom, line 30: Ditto on the comment.

bolt-modules/file/lib/puppet/functions/file/read.rb
```diff
@@ -17,10 +17,16 @@
...
+    executor.report_function_call(self.class.name)
 
-    found = Puppet::Parser::Files.find_file(filename, scope.compiler.environment)
+    found = if executor.future&.fetch('file_paths', false)
+              Puppet::Parser::Files.find_in_module(filename, scope.compiler.environment) do |mod, module_file|
+                Bolt::Util.find_file_in_mod(mod, module_file)
+              end
+            else
+              Puppet::Parser::Files.find_file(filename, scope.compiler.environment)
+            end
```
beechtom, line 29: Ditto on the comment.

bolt-modules/file/lib/puppet/functions/file/readable.rb
```diff
@@ -18,9 +18,16 @@
...
+    executor.report_function_call(self.class.name)
 
-    found = Puppet::Parser::Files.find_file(filename, scope.compiler.environment)
+    found = if executor.future&.fetch('file_paths', false)
+              Puppet::Parser::Files.find_in_module(filename, scope.compiler.environment) do |mod, module_file|
+                Bolt::Util.find_file_in_mod(mod, module_file)
+              end
+            else
+              Puppet::Parser::Files.find_file(filename, scope.compiler.environment)
+            end
```
beechtom, line 30: image

lucywyman:
Ok, did quite a bit of refactoring so that everything ends up using the same function to search modules. I looked through Puppet's file loading to make sure it doesn't do any other magic to load from other paths. This way there's a guaranteed consistent experience between the plan function path loading and the CLI loading.

beechtom:
- [ ] `module/script.sh`
  - [ ] Finds `module/files/script.sh`
  - [ ] Errors if `module/files/script.sh` is missing
- [ ] `modules/scripts/script.sh`
  - [ ] Finds `module/files/scripts/script.sh`
  - [ ] Falls back to `module/scripts/script.sh` with `file_paths` configured
  - [ ] Errors if both files are missing with `file_paths` configured
  - [ ] Errors if `module/files/scripts/script.sh` is missing without `file_paths` configured
- [ ] `modules/files/script.sh`
  - [ ] Finds `module/files/files/script.sh`
  - [ ] Falls back to `module/files/script.sh` with `file_paths` configured
  - [ ] Errors if both files are missing with `file_paths` configured
  - [ ] Errors if `module/files/files/script.sh` is missing without `file_paths` configured
- `run_script`, `upload_file`, and `file::*` plan functions use new behavior
- `bolt script run` and `bolt file upload` commands use new behavior

bolt-modules/boltlib/lib/puppet/functions/run_script.rb
```diff
@@ -108,7 +110,22 @@ def run_script_with_description(scope, script, targets, description = nil, optio
...
+              # Fall back to new-style paths if enabled.
+              fallback = executor.future&.fetch('file_paths', false)
+              # Split the module_name from the pattern used to find the file
+              module_name, file_pattern = Bolt::Util.split_file_path(script)
+              # Get the absolute path to the module root from Puppet
+              mod_path = scope.compiler.environment.module(module_name)&.path
+              # Search the module for the file, optionally falling back to new-style
+              # paths if enabled.
+              Bolt::Util.find_file(mod_path, file_pattern, fallback) if mod_path
+            end
```
beechtom, line 127: Could this be a method on `Bolt::Util` since it's used in multiple plan functions? I don't imagine this will need to change, but just in case it does that would reduce risk.

bolt-modules/file/lib/puppet/functions/file/exists.rb
```diff
@@ -12,14 +12,31 @@
...
+    required_param 'String[1]', :filename
     return_type 'Boolean'
   end
 
   def exists(scope, filename)
     # Send Analytics Report
-    Puppet.lookup(:bolt_executor) {}&.report_function_call(self.class.name)
-    found = Puppet::Parser::Files.find_file(filename, scope.compiler.environment)
+    executor = Puppet.lookup(:bolt_executor)
+    executor.report_function_call(self.class.name)
```
beechtom, line 22: Why this change? Don't functions available in apply blocks need to have the safe operator for the executor?

lib/bolt/cli.rb
```diff
@@ -563,7 +563,8 @@ def execute(options)
             when 'command'
               executor.run_command(targets, options[:object], executor_opts)
             when 'script'
-              script_path = find_file(options[:object])
+              script_path = find_file(options[:object], executor.future&.fetch('file_paths', false))
```
beechtom, line 566: We could update `Bolt::Config` to set `'future' => {}` by default so we can get rid of the safe operator.

beechtom (commented):
Couple small comments, but otherwise lgtm! The refactor of the `Bolt::Util` methods makes it a lot easier to understand what's happening as well. :)

bolt-modules/boltlib/lib/puppet/functions/run_script.rb
lucywyman: Yeah, that's a good idea. I moved this to be on the executor, which feels more correct for passing scope around and will also let any other executors easily override this function.

lib/bolt/cli.rb
lucywyman: I thought about that, but I think at some point it might useful to know if `future` was set at all or not. We may not end up needing that, but I don't want to rule it out so early.

bolt-modules/file/lib/puppet/functions/file/exists.rb
lucywyman: Oh my gosh thank you for catching this! I had completely forgotten these could be called from inside apply blocks. Ended up reworking everything so that Bolt can locate these files without an executor.

bolt.gemspec
```diff
@@ -55,7 +55,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency "net-ssh", ">= 4.0"
   spec.add_dependency "net-ssh-krb", "~> 0.5"
   spec.add_dependency "orchestrator_client", "~> 0.5"
-  spec.add_dependency "puppet", ">= 6.18.0"
+  # spec.add_dependency "puppet", ">= 6.18.0"
```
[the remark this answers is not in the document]
lucywyman: Haha at least you know I actually forgot to touch this last week 😅

beechtom (approved):

Issue #2759, opened 2021-03-31
Disable analytics in bolt-project.yaml or bolt-defaults.yaml


beechtom:
## Use Case
- I'm a user who wants to provide analytics, but disable it for specific projects that I do not wish Bolt to receive analytics for.
- I'm a system administrator who does not wish for any user running Bolt to submit analytics, and do not want to rely on users disabling analytics in their analytics.yaml file.

## Solution
Bolt loads configuration before creating the analytics client, which allows us to add a configuration option to both `bolt-project.yaml` and `bolt-defaults.yaml` that will disable analytics. The simplest option name might be `analytics` which accepts a boolean value. If it's false at any level it should disable analytics.

```
# bolt-project.yaml
analytics: false
```

PR #2763, opened 2021-04-01
(GH-2759) Add 'analytics' config option to disable analytics

beechtom:
This adds a new configuration option called `analytics` that disables
data collection in Bolt when set to `false`. The option is supported in
both `bolt-defaults.yaml` and `bolt-project.yaml`. If this option is set
to `false`, it cannot be overriden in another configuration file. For
example, setting `analytics: true` in `bolt-project.yaml` will not
enable data collection if `analytics: false` is set in
`bolt-defaults.yaml`.

!feature

* **Disable analytics in system, user, and project config files**
  ([#2759](https://github.com/puppetlabs/bolt/issues/2759))

  The new `analytics` configuration option can be used to disable data
  collection in Bolt and is supported in both `bolt-defaults.yaml` and
  `bolt-project.yaml`. Disabling data collection cannot be overridden by
  enabling it in another configuration file.

### DIFF
--- a/documentation/analytics.md
+++ b/documentation/analytics.md
@@ -5,13 +5,33 @@ about how to improve it. You can opt out of providing this data.
 
 ## Opt out of data collection
 
-You can opt out of data collection by modifying Bolt's analytics configuration
-file. This file is located in the user's home directory at `<HOME
-DIRECTORY>/.puppetlabs/etc/bolt/analytics.yaml`. To opt out of data collection,
-add the following line to the file:
+You can opt out of data collection by setting an option in Bolt's configuration
+or by setting an environment variable.
+
+### Bolt configuration
+
+To disable data collection, set `analytics: false` in your [configuration
+file](configuring_bolt.md). This option is supported in the system-wide,
+user-level, and project configuration files.
 
 ```yaml
-disabled: true
+# bolt-defaults.yaml
+analytics: false
+```
+
+Setting the `analytics: false` option in a configuration file disables data
+collection universally. You cannot override the option by setting it to `true`
+in another configuration file. For example, setting `analytics: true` in a
+project configuration file does not enable data collection if you've set
+`analytics: false` in the system-wide or user-level configuration file.
+
+### Environment variable
+
+To disable data collection, set the `BOLT_DISABLE_ANALYTICS` environment
+variable to any value.
+
+```
+export BOLT_DISABLE_ANALYTICS=true
 ```
 
 ## Data collected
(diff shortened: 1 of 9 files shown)


documentation/analytics.md
```diff
@@ -5,13 +5,32 @@ about how to improve it. You can opt out of providing this data.
...
 ```yaml
-disabled: true
+# bolt-defaults.yaml
+analytics: false
+```
+
+Setting the `analytics: false` option cannot be overridden in another
+configuration file. For example, setting `analytics: true` in a project
+configuration file will not enable data collection if you have set
+`analytics: false` in the system-wide or user-level configuration file.
```
hestonhoffman, line 25:
```suggestion
Setting the `analytics: false` option in a configuration file disables
data collection universally. You cannot override the option by setting
it to `true` in another configuration file. For example, setting 
`analytics: true` in a project configuration file does not enable data 
collection if you've set `analytics: false` in the system-wide or user-level 
configuration file.
```

lib/bolt/config.rb
```diff
@@ -263,6 +264,8 @@ def merge_config_layers(*config_data)
           # Disabled warnings are concatenated
           when 'disable-warnings'
             val1.concat(val2)
+          when 'analytics'
+            val1 == false ? false : val2
```
lucywyman, line 268:
```suggestion
            val1 && val2
```

lucywyman (approved):
- [x] Disables analytics when configured to false in user-level configuration
- [x] Disables analytics when configured to false in project-level configuration
- [x] `false` is not overridden by project-level config
- [x] Does not disable analytics if all config files set to `true`
- [x] Respects `analytics.yaml` configuration
- [x] Disables analytics if configured true and env var is set
- [x] Only accepts boolean values for `analytics`

documentation/analytics.md
```diff
@@ -5,13 +5,32 @@ about how to improve it. You can opt out of providing this data.
 
 ## Opt out of data collection
 
-You can opt out of data collection by modifying Bolt's analytics configuration
-file. This file is located in the user's home directory at `<HOME
```
lucywyman, line 9: We also need to update https://github.com/puppetlabs/bolt/pull/2763/files#diff-eadede6f594353cfaf24138e4b22335b58e711eff9ca3121a84d67c5fda19bb5R84 to similarly recommend setting `analytics: false` in user-level configuration instead.

lucywyman (approved):

hestonhoffman (approved):

PR #2766, opened 2021-04-05
(hack) Improve 'group show' output

beechtom:
This updates the output for the `bolt group show` / `Get-BoltGroup`
commands to better align with other `show` commands. The command now
lists information under colorized headers and also lists the source of
the inventory file if one was loaded.

!feature

* **Improved group information output**

  The `bolt group show` and `Get-BoltGroup` commands now display `human`
  output in a similar format to other `show` commands. The `json` output
  now includes the path to the inventory file that the groups are loaded
  from.

### DIFF
--- a/lib/bolt/cli.rb
+++ b/lib/bolt/cli.rb
@@ -642,8 +642,12 @@ module Bolt
         options[:targets] = 'all'
       end
 
-      inventoryfile = config.inventoryfile || config.default_inventoryfile
-      outputter.print_targets(group_targets_by_source, inventoryfile, target_flag)
+      outputter.print_targets(
+        group_targets_by_source,
+        inventory.source,
+        config.default_inventoryfile,
+        target_flag
+      )
     end
 
     def show_targets
@@ -653,8 +657,12 @@ module Bolt
         options[:targets] = 'all'
       end
 
-      inventoryfile = config.inventoryfile || config.default_inventoryfile
-      outputter.print_target_info(group_targets_by_source, inventoryfile, target_flag)
+      outputter.print_target_info(
+        group_targets_by_source,
+        inventory.source,
+        config.default_inventoryfile,
+        target_flag
+      )
     end
 
     # Returns a hash of targets sorted by those that are found in the
@@ -675,8 +683,7 @@ module Bolt
     end
 
     def list_groups
-      groups = inventory.group_names
-      outputter.print_groups(groups)
+      outputter.print_groups(inventory.group_names.sort, inventory.source, config.default_inventoryfile)
     end
 
     def run_plan(plan_name, plan_arguments, nodes, options)
--- a/lib/bolt/outputter/human.rb
+++ b/lib/bolt/outputter/human.rb
@@ -483,7 +483,7 @@ module Bolt
         end
       end
 
-      def print_targets(target_list, inventoryfile, target_flag)
+      def print_targets(target_list, inventory_source, default_inventory, target_flag)
         adhoc = colorize(:yellow, "(Not found in inventory file)")
 
         targets  = []
@@ -501,18 +501,13 @@ module Bolt
                 end
         info << "\n\n"
 
-        @stream.puts info
+        info << format_inventory_source(inventory_source, default_inventory)
+        info << format_target_summary(target_list[:inventory].count, target_list[:adhoc].count, target_flag, false)
 
-        print_inventory_summary(
-          target_list[:inventory].count,
-          target_list[:adhoc].count,
-          inventoryfile,
-          target_flag,
-          false
-        )
+        @stream.puts info
       end
 
-      def print_target_info(target_list, inventoryfile, target_flag)
+      def print_target_info(target_list, inventory_source, default_inventory, target_flag)
         adhoc_targets     = target_list[:adhoc].map(&:name).to_set
         inventory_targets = target_list[:inventory].map(&:name).to_set
         targets           = target_list.values.flatten.sort_by(&:name)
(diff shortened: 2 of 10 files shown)


lib/bolt/cli.rb
```diff
@@ -661,8 +661,9 @@ def show_targets
     end
 
     def list_groups
-      groups = inventory.group_names
-      outputter.print_groups(groups)
+      groups = inventory.group_names.sort
+      inventoryfile = config.inventoryfile || config.default_inventoryfile
```
lucywyman, line 665: Hmm, it actually looks like this and the `target show` commands are missing `ENV[Bolt::Inventory::ENVIRONMENT_VAR]` as a possible source. I think the simplest way to reliably get the inventory source would be to have the inventory itself know where it comes from (since it can only have one source), so add a `source` instance var to `Bolt::Inventory::Inventory` then we can call `inventory.source` whenever we need this info.

lib/bolt/outputter/json.rb
```diff
@@ -125,10 +125,11 @@ def print_target_info(target_list, _inventoryfile)
         )
       end
 
-      def print_groups(groups)
+      def print_groups(groups, inventoryfile)
         count = groups.count
         @stream.puts({ groups: groups,
-                       count: count }.to_json)
+                       count: count,
+                       inventoryfile: inventoryfile.to_s }.to_json)
```
lucywyman, line 132: In `bolt target show` output this is nested under `inventory: { file: }` I think following that same structure here would good for consistency and listing the groups there as well.

lib/bolt/cli.rb
beechtom: Opted to put this logic directly in the outputter. While storing this info on the inventory class does make sense, I don't think it makes things simpler here, since we need some way to tell if 1) the source is the environment variable or a file path and 2) check if the source exists and inventory was actually loaded from it and 3) if not, where we attempted to load it from. For now the simplest approach seems to be just checking if the env var is set in the outputter.

lib/bolt/outputter/json.rb
beechtom: The `inventory` section in JSON is meant for showing which targets are loaded from inventory. I'm not sure that really applies here, since all groups (except `all`) _must_ come from inventory. Removed adding the inventory file for now and can be addressed in a separate ticket.

lucywyman: Consistent JSON API response formats are a [pretty basic tenet](https://stackoverflow.com/questions/50148694/rest-api-json-format-consistency) of API design so that users can predict and understand responses easily without having to fail first, run the command manually and study the output, and adjust. I was actually thinking we should also have the `groups` listed both under `inventory` and at the top level similar to `inventory show` output as well for this reason. Plus, *not* all groups are defined in the inventory file: the 'all' group is a notable exception, and if users are loading inventory from an environment variable I think that'd probably be a different key under the 'inventory' key.

Don't take it from me, here are some other sources:
- [“consistency underlies all principles. A good architecture is consistent in the sense that, given a partial knowledge of the system, one can predict the remainder.”](https://cloud.google.com/blog/products/api-management/api-design-best-practices-common-pitfalls)
- ["It should be simple, intuitive, and consistent to make adoption not only easy but pleasant"](https://www.vinaysahni.com/best-practices-for-a-pragmatic-restful-api)
- ["it’s best to work toward consistency when designing new aspects of an existing API."](https://www.oreilly.com/content/best-design-practices-to-get-the-most-out-of-your-api/)

No one's asked for this specifically yet so it's fine to leave it out so we don't paint ourselves into a corner.

lib/bolt/cli.rb
lucywyman: I don't follow - the inventory object already does all of that checking for you, so you wouldn't have to do it here. If you set 'source' in `Bolt::Inventory.from_config`, pass that through to `create_version` (and include `source` from sources that call `create_version`). That seems reliable to me and wouldn't require any checking here.

beechtom: Maybe we could pass `nil` as the source to indicate inventory couldn't be found anywhere, though we'd also need a way to get the path to the default inventory file... so maybe we pass that along too when calling the outputter functions? :/ 

```
outputter.print_groups(
  groups,
  inventory.source,
  config.default_inventoryfile
)
```

lucywyman:
The commit message should also list fixing the bug that the inventory file was printed even if the inventory was loaded from an env var.

lib/bolt/outputter/human.rb
```diff
@@ -532,41 +529,54 @@ def print_target_info(target_list, inventoryfile)
...
         # Add inventory file source
-        info << colorize(:cyan, "Inventory file\n")
-        info << if File.exist?(inventoryfile)
+        info << colorize(:cyan, "Inventory source\n")
+        info << if ENV[Bolt::Inventory::ENVIRONMENT_VAR]
+                  indent(2, "#{Bolt::Inventory::ENVIRONMENT_VAR} environment variable\n")
+                elsif File.exist?(inventoryfile)
                   indent(2, "#{inventoryfile}\n")
                 else
                   indent(2, wrap("Tried to load inventory from #{inventoryfile}, but the file does not exist\n"))
```
lucywyman, line 548: In what situation is this not already caught when we load the inventory in the CLI?

spec/bolt/outputter/human_spec.rb
```diff
@@ -421,18 +424,60 @@
...
+    end
+
+    it 'prints the inventory environment variable' do
+      with_env_vars('BOLT_INVENTORY' => '') do
+        outputter.print_groups(groups, inventoryfile)
+        expect(output.string).to match(/Inventory source.*BOLT_INVENTORY environment variable/m)
+      end
+    end
+
+    it 'prints that the inventorfile does not exist' do
```
lucywyman, line 472:
```suggestion
    it 'prints that the inventory file does not exist' do
```

spec/bolt/outputter/human_spec.rb
```diff
@@ -421,18 +424,60 @@
     it 'prints the inventory file path' do
       expect(File).to receive(:exist?).with(inventoryfile).and_return(true)
       outputter.print_targets(target_list, inventoryfile)
-      expect(output.string).to match(/Inventory file.*#{inventoryfile}/m)
+      expect(output.string).to match(/Inventory source.*#{inventoryfile}/m)
+    end
+
```
lucywyman, line 429: It'd be great to include a smoke test for this in `spec/integration/inventory_spec`.

lucywyman:
#### bolt group show
- [x] Always shows the 'all' group
- [x] Prints reasonably when no inventory file can be loaded
- [x] Lists the environment variable as a source if set
- [x] Lists inventory groups loaded from plugins
#### bolt inventory show
- [x] Only modifies the 'Inventory source' header for `inventory show`
  - [x] Lists the environment variable as a source if set 
#### bolt inventory show --detail
- [x] Only modifies the 'Inventory source' header for `inventory show`
  - [x] Lists the environment variable as a source if set

lib/bolt/outputter/human.rb
beechtom: Can you clarify? Not quite sure what you're asking here.

spec/bolt/outputter/human_spec.rb
beechtom: Added for both `inventory show` and `group show`.

lib/bolt/cli.rb
beechtom: ^ Went with the above approach.

lib/bolt/outputter/human.rb
lucywyman: Ah, sorry, I could definitely have asked more clearly! I was wondering if this was dead code because we check for inventory file existence when loading from the CLI, but it's not because this gets hit when Bolt falls back to the default inventory file and it doesn't exist :+1:

lucywyman (approved):

PR #2770, opened 2021-04-06
(bugfix) Ensure all messages are streamed and saved to merged_output

lucywyman:
In the Bash shell class we collect messages from stdout and stderr in a
few places: when using `run-as` after checking for a prompt, during the
read-loop as the command is executing, and then reading anything that's
unread after the thread exits. When we added streaming and merged_output
features we did this for the main read-loop, but not for the initial
sudo-prompt check or the cleanup loop that executes once the thread has
finished. This resulted in some scripts and commands not having their
results printed since we now print merged_output, which was missing
messages from the final stream reading.

Now, any messages read during the initial sudo-prompt check or once the
thread has finished will be read and logged to the stream output and
saved to `merged_output` so that it can be printed.

!bug

* **Ensure all messages printed, even after thread finishes**

  Bolt now ensures that all messages from a command or script are
  printed back to the user. Previously, some messages would be lost
  if they were read after the thread finished executing or when Bolt had
  been prompted for a sudo password.

### DIFF
--- a/spec/integration/local_spec.rb
+++ b/spec/integration/local_spec.rb
@@ -123,6 +123,11 @@ describe "when running over the local transport" do
       expect(result.map { |r| r['stdout'].strip }).to eq(%w[root root])
     end
 
+    it 'returns merged output', :reset_puppet_settings do
+      result = run_nodes(["command", "run", ">&2 echo Hello $USER && whoami"] + config_flags)
+      expect(result.map { |r| r['merged_output'].strip }).to eq(["Hello root\nroot", "Hello root\nroot"])
+    end
+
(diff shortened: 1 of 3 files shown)


spec/integration/local_spec.rb
```diff
@@ -123,6 +123,11 @@
       expect(result.map { |r| r['stdout'].strip }).to eq(%w[root root])
     end
 
+    it 'returns merged output', :reset_puppet_settings do
```
lucywyman, line 126: Verified this fails *without* these changes.

beechtom (approved):

Issue #2747, opened 2021-03-29
Default to showing all of the inventory


genebean:
## Use Case
`bolt inventory show` should default to showing something instead of just complaining. I would think that without additional limiting commands that it would show me my entire inventory. Also, the output of `bolt inventory show --help` does not indicate that `-t all` is a valid thing... I only learned about it via Slack.

## Solution
- Default to show the inventory with `-t all`
- Include an additional information section with the cue to run with flags to filter the output, similar to `module show`
- Audit existing bolt option parser output to ensure that required options are marked as required, and ensure that the usage string include required options.
- Add verbiage around requiring a 'targeting key' and cue to run 'bolt guide targets' for more information about targeting flags.
- Add bolt guide on targets

genebean:
Also, something in or related to this should break things up by the groups used in the config file, even if it requires a flag to do so.

PR #2775, opened 2021-04-07
(GH-2747) Default to showing all targets with 'inventory show'

beechtom:
(GH-2747) Default to showing all targets with 'inventory show'

This updates the `bolt inventory show` and `Get-BoltInventory` commands
to display the 'all' group if a targetting option is not provided to the
command. Additionally, the human outputter now displays additional
information to the user, indicating that they can view specific targets
using a targetting option or view target config and data with
`--detail`, assuming the options are not provided by the user.

!feature

* **Default to showing all targets with `bolt inventory show`**
  ([#2747](#2747))

  The `bolt inventory show` and `Get-BoltInventory` commands now default
  ot showing all targets in the inventory if a targetting option
  (`--targets`, `--query`, `--rerun`) are not provided.

---

(GH-2747) Show required options in help text

This updates the help text to show required options for each command.

!no-release-note

---

(GH-2747) Add 'targets' guide

This adds a new 'targets' guide to the `bolt guide` feature.

!no-release-note

### DIFF
--- a/guides/targets.txt
+++ b/guides/targets.txt
@@ -1,0 +1,31 @@
+TOPIC
+    targets
+
+DESCRIPTION
+    A target is a device that Bolt connects to and runs actions on. Targets can
+    be physical, such as servers, or virtual, such as containers or virtual
+    machines.
+
+    Several of Bolt's commands connect to targets and run actions on them. When
+    you run these commands, Bolt requires a list of targets to run the action
+    on. For example, the `bolt script run` command and `Invoke-BoltScript`
+    PowerShell cmdlet require you to provide a list of targets to run your
+    script on. You can provide a list of targets to a command using one of the
+    following command-line options:
+
+    Shell commands
+        -t, --targets TARGETS
+        -q, --query   QUERY
+            --rerun   FILTER
+
+    PowerShell cmdlets
+        -T, -Targets TARGETS
+        -Q, -Query   QUERY
+            -Rerun   FILTER
+
+    Typically, targets and their configuration and data are listed in a
+    project's inventory file. For more information about inventory files,
+    see 'bolt guide inventory'.
+
+DOCUMENTATION
+    https://pup.pt/bolt-commands
--- a/lib/bolt/bolt_option_parser.rb
+++ b/lib/bolt/bolt_option_parser.rb
@@ -190,7 +190,8 @@ module Bolt
           apply
 
       USAGE
-          bolt apply [manifest.pp] [options]
+          bolt apply [manifest] {--targets TARGETS | --query QUERY | --rerun FILTER}
+            [options]
 
       DESCRIPTION
           Apply Puppet manifest code on the specified targets.
@@ -219,7 +220,8 @@ module Bolt
           run
 
       USAGE
-          bolt command run <command> [options]
+          bolt command run <command> {--targets TARGETS | --query QUERY | --rerun FILTER}
+            [options]
 
       DESCRIPTION
           Run a command on the specified targets.
@@ -248,7 +250,8 @@ module Bolt
           download
 
       USAGE
-          bolt file download <src> <dest> [options]
+          bolt file download <source> <destination> {--targets TARGETS | --query QUERY | --rerun FILTER}
+            [options]
 
       DESCRIPTION
           Download a file or directory from one or more targets.
@@ -267,7 +270,8 @@ module Bolt
           upload
 
       USAGE
-          bolt file upload <src> <dest> [options]
+          bolt file upload <source> <destination> {--targets TARGETS | --query QUERY | --rerun FILTER}
+            [options]
 
       DESCRIPTION
           Upload a local file or directory.
@@ -343,7 +347,12 @@ module Bolt
           bolt inventory show [options]
 
       DESCRIPTION
-          Show the list of targets an action would run on.
+          Show the list of targets an action would run on. This command will list
+          all targets in the project's inventory by default.
+
+          To filter the targets in the list, use the --targets, --query, or --rerun
+          options. To view detailed configuration and data for targets, use the
+          --detail option.
     HELP
 
     MODULE_HELP = <<~HELP
@@ -432,7 +441,7 @@ module Bolt
           plan
 
       USAGE
-          bolt plan <action> [parameters] [options]
+          bolt plan <action> [options]
 
       DESCRIPTION
           Convert, create, show, and run Bolt plans.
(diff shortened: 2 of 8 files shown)


beechtom:
Updates to usage strings is based on https://developers.google.com/style/code-syntax . Open to adding the angled brackets back in for required arguments, though I think we should stick with the capitalization of required and optional arguments since that's what we do in docs.

guides/targets.txt
```diff
@@ -0,0 +1,32 @@
+TOPIC
+    targets
+
+DESCRIPTION
+    A target is a device that Bolt connects to and runs actions on. Targets can
+    be physical, such as servers, or virtual, such as containers or virtual
+    machines.
+
+    Several of Bolt's commands connect to targets and run actions on them. For
+    example, the `bolt script run` command runs a script on a target. When you
```
lucywyman, line 10:
```suggestion
    example, the `bolt script run` command and `Invoke-BoltScript` PowerShell cmdlet run a script on a target. When you
```

guides/targets.txt
```diff
@@ -0,0 +1,32 @@
...
+    run these commands, Bolt requires a list of targets to run the action on.
+    You can provide this list of targets using one of the following command-line
+    options:
+
+        -t, --targets TARGETS
```
lucywyman, line 15: This feels more like reference material than a guide. I think it'd be tidier to list the options (possibly even generate them from `lib/bolt/cli.rb#L266`) and then either link to reference docs or instruct to run `bolt <subcommand> --help` to see details. Defer to @hestonhoffman if he thinks the expanded version will be more helpful.

lib/bolt/bolt_option_parser.rb
```diff
@@ -161,7 +161,7 @@ def get_help_text(subcommand, action = nil)
           bolt
 
       USAGE
-          bolt <subcommand> [action] [options]
+          bolt SUBCOMMAND [ACTION] [OPTIONS]
```
lucywyman, line 164: I think all-caps makes sense for things like `-t TARGETS` to clarify which part is the flag and which part isn't, but otherwise lowercase is standard for CLI tools. The `<subcommand>` brackets are also important to instruct the user that they need to fill that part in - so `bolt subcommand --help` isn't valid.

- [docopt standard](http://docopt.org/)
- [Microsoft CLI syntax](https://docs.microsoft.com/en-us/windows-server/administration/windows-commands/command-line-syntax-key)

lib/bolt/bolt_option_parser.rb
```diff
@@ -340,18 +343,23 @@ module            Manage Bolt project modules
           show
 
       USAGE
-          bolt inventory show [options]
+          bolt inventory show [OPTIONS]
```
lucywyman, line 346: This and `bolt inventory` also require targetting flags.

lib/bolt/bolt_option_parser.rb
```diff
@@ -467,7 +476,7 @@ module            Manage Bolt project modules
           new
       
       USAGE
-          bolt plan new <plan> [options]
+          bolt plan new NAME [OPTIONS]
```
lucywyman, line 479: Hm, to me 'plan' is more clear than 'name', but 'plan name' is probably most clear.

lib/bolt/bolt_option_parser.rb
```diff
@@ -702,15 +714,8 @@ def initialize(options)
...
-             'Enter a comma-separated list of target URIs or group names.',
-             "Or read a target list from an input file '@<file>' or stdin '-'.",
-             'Example: --targets localhost,target_group,ssh://nix.com:23,winrm://windows.puppet.com',
-             'URI format is [protocol://]host[:port]',
-             "SSH is the default protocol; can be #{TRANSPORTS.keys.join(', ')}",
-             'For Windows targets, specify the winrm:// protocol if it has not be configured',
-             'For SSH, port defaults to `22`',
-             'For WinRM, port defaults to `5985` or `5986` based on the --[no-]ssl setting') do |targets|
+             'Specify the targets or group of targets to run on.',
+             "For more information, see 'bolt guide targets'.") do |targets|
```
lucywyman, line 718: Aha, ok, I see now where the guide came from. It seems like all this text belongs in a docs section somewhere that we can link to from the guide.

lib/bolt/cli.rb
```diff
@@ -634,13 +634,25 @@ def list_plans
     end
 
     def list_targets
+      if options[:targets] || options[:rerun] || options[:query]
```
lucywyman, line 637: We also use these targetting options on line 265 - making those a constant and replacing the references would be good so if they change they're all in one place.

lib/bolt/bolt_option_parser.rb
beechtom: I'll switch back to using brackets. Fwiw, this is the doc I had been referencing, which suggested all caps: https://developers.google.com/style/code-syntax . Everyone's gotta have their own standard. :P

lib/bolt/bolt_option_parser.rb
beechtom: Most of this info is present on the 'Run Bolt' page: https://puppet.com/docs/bolt/latest/running_bolt_commands.html#specify-a-transport
I'm going to leave this in for now, but I'd like to revisit it later since this text is from the very early days of Bolt and there might be better ways to surface this info to users rather than put it in the description for the option.

guides/targets.txt
beechtom: Removed the description of each and just left the options themselves in.

guides/targets.txt
```diff
@@ -0,0 +1,30 @@
...
+DESCRIPTION
+    A target is a device that Bolt connects to and runs actions on. Targets can
+    be physical, such as servers, or virtual, such as containers or virtual
+    machines.
+
+    Several of Bolt's commands connect to targets and run actions on them. For
+    example, the `bolt script run` command and `Invoke-BoltScript` PowerShell
+    cmdlet run a script on a target. When you run these commands, Bolt requires
+    a list of targets to run the action on. You can provide this list of targets
+    using one of the following command-line options:
```
hestonhoffman, line 13:
```suggestion
    Several of Bolt's commands connect to targets and run actions on them. When
    you run these commands, Bolt requires a list of targets to run the action on. For
    example, the `bolt script run` command and `Invoke-BoltScript` PowerShell
    cmdlet require you to provide a list of targets to run your script on. You can 
    provide a list of targets to a command using one of the following 
    command-line options:
```

lucywyman (approved):
#### bolt guide
- [x] Shows new 'targets' guide in 'bolt guide' output
- [x] Prints new 'targets' guide correctly

#### Adds targetting flags to help text:
- [x] `bolt apply`
- [x] `bolt command run`
- [x] `bolt file download`
- [x] `bolt file upload`
- [x] `bolt script run`
- [x] `bolt task run`

#### bolt inventory show
- [x] Defaults to the 'all' group
- [x] Only shows detail help text if not passed `--detail`
- [x] Only shows filtering help text if not passed
  - [x] `-t`
  - [x] `--rerun`
  - [x] `-q`

giphy

hestonhoffman (approved):
LGTM!

Issue #2777, opened 2021-04-08
Metaparameters for task step in yaml plans raise an error


adreyer:
## Describe the Bug
Trying to set a metaparameter in a yaml plan task step raises an error about

## Expected Behavior
The following should be a valid task step
```
steps:
  -
     name: command_step 
    task: service
    targets: $targets
    catch_errors: false
    parameters:
      action: status
      name: puppet
```

instead it fails with 
```
Parse error in step "command_step": 
 Cannot specify metaparameters when using top-level keys with same name: _catch_errors

```

### Environment
 - Version  3.5

PR #2778, opened 2021-04-08
(GH-2777) Do not error when using metaparameters in YAML plans

beechtom:
This fixes a bug that was causing YAML plan plan and task steps that
specified a metaparameter under the `parameters` key to error. The
correct behavior for Bolt is to only error when the metaparameter is
present as both a top-level key _and_ under the `parameters` key. For
example, specifying the `catch_errors` key and the
`parameters._catch_errors` key.

!bug

* **Do not error when using metaparameters in YAML plans**
  ([#2777](https://github.com/puppetlabs/bolt/issues/2777))

  Bolt no longer errors for YAML plans that include a plan or task step
  that includes an additional option (e.g. `_catch_errors`) under the
  `parameters` key.

### DIFF
--- a/lib/bolt/pal/yaml_plan/step.rb
+++ b/lib/bolt/pal/yaml_plan/step.rb
@@ -122,9 +122,11 @@ module Bolt
               raise StepError.new("Parameters key must be a hash", body['name'], step_number)
             end
 
-            metaparams = option_keys.map { |key| "_#{key}" }
+            metaparams = body['parameters'].keys
+                                           .select { |key| key.start_with?('_') }
+                                           .map { |key| key.sub(/^_/, '') }
 
-            if (dups = body['parameters'].keys & metaparams).any?
+            if (dups = body.keys & metaparams).any?
               raise StepError.new(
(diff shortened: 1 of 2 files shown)


lib/bolt/pal/yaml_plan/step.rb
```diff
@@ -122,9 +122,11 @@ def self.validate(body, step_number)
               raise StepError.new("Parameters key must be a hash", body['name'], step_number)
             end
 
-            metaparams = option_keys.map { |key| "_#{key}" }
+            metaparams = body['parameters'].keys
```
lucywyman, line 125:
```suggestion
            metaparams = body['parameters'].transform_keys { |k| k.sub(/^_/, '') }
```

lucywyman (approved):

lucywyman:
- [x] Accepts `_catch_errors` as a parameter
- [x] Errors when both `_catch_errors` and top-level `catch_errors` are provided
- [x] Accepts top-level parameters
- [x] Accepts `_run_as` as a parameter

Issue #2757, opened 2021-03-31
`bolt plan new` should default to use the current project name instead of failing


op-ct:
## Describe the Bug

`bolt plan new` fails unless the plan's first name segment matches the project name.  

However: Bolt already knows the project name, so it should just add it if only one name segment is given.

## Steps to Reproduce

1. Run `bolt plan new foo --pp` from inside a project directory ("boltybolt"):

```console
$ bolt plan new foo --pp
First segment of plan name 'foo' must match project name 'boltybolt'. Did you mean 'boltybolt::foo'?
```

## Expected Behavior

```console
$ bolt plan new foo --pp
Created plan 'boltybolt::foo' at '/path/to/boltybolt/plans/foo.pp'

Show this plan with:
    bolt plan show boltybolt::foo
Run this plan with:
    bolt plan run boltybolt::foo
```

## Environment
 - Version: 3.5.0
 - Platform: Debian 10

lucywyman:
Hm, I think the team agrees that the happy path here would be a better experience, but it introduces some cases where Bolt has to guess what the user meant (for example, in a project named "project" if the user runs "bolt plan new project::foo" do they want to create plans/foo.yaml or plans/project/foo.yaml?). Forcing users to include the project namespace also reinforces that the `namespace::planname` is the name of the plan (for example for `plan show` or `plan run`).

hestonhoffman:
How about:
> Incomplete plan name: Plan names consist of the project or module name followed by the plan name. Did you mean: {project}::{name}

beechtom:
Closed by #2779

PR #2779, opened 2021-04-08
(GH-2757) Update error message when 'plan new' receives incomplete name

lucywyman:
This updates the error message when `bolt plan new` is run and receives
an invalid plan name to clarify that plan names include the project or
module that the plan is loaded from, so that users understand that plan
names must always include the module namespace in other contexts of Bolt
(for example `plan run` or `plan show`).

!no-release-note

### DIFF
--- a/lib/bolt/plan_creator.rb
+++ b/lib/bolt/plan_creator.rb
@@ -36,8 +36,8 @@ module Bolt
       prefix, _, basename = segment_plan_name(plan_name)
 
       unless prefix == project.name
-        message = "First segment of plan name '#{plan_name}' must match project name '#{project.name}'. "\
-                  "Did you mean '#{project.name}::#{plan_name}'?"
+        message = "Incomplete plan name: A plan name must be prefixed with the name of the "\
+          "project or module. Did you mean '#{project.name}::#{plan_name}'?"
 
         raise Bolt::ValidationError, message
       end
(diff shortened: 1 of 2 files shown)


lib/bolt/plan_creator.rb
```diff
@@ -36,8 +36,8 @@ def self.validate_input(project, plan_name)
       prefix, _, basename = segment_plan_name(plan_name)
 
       unless prefix == project.name
-        message = "First segment of plan name '#{plan_name}' must match project name '#{project.name}'. "\
-                  "Did you mean '#{project.name}::#{plan_name}'?"
+        message = "Incomplete plan name: Plan names consist of the project or module name "\
+          "followed by the specific plan. Did you mean '#{project.name}::#{plan_name}'?"
```
beechtom, line 40: "specific plan name"? "rest of the plan name"? Something sounds a bit odd about "specific plan". @hestonhoffman , any thoughts?

hestonhoffman: Hmmm. I see the issue with the original suggestion. "Plan name" is repetitive. How about something like this:
"Incomplete plan name: A plan name must be prefixed with the name of the project or module. Did you mean..."

lucywyman: Yeah that seems better - I wasn't sure what to call it either :P

beechtom (approved):

Issue #2771, opened 2021-04-07
Push Facts to PDB from a bolt plan


adreyer:
## Use Case
I want to use PuppetDB from bolt without managing long-running puppet-server and regular agent check-ins.

PuppetDB is a powerful tool to track inventory that can be leveraged for selecting targets from Bolt with `--query` flag or the `pdb_query` function. Unfortunately, that workflow is only available for targets that are part of a larger classic puppet installation not agentless targets or environments without a puppet-server and CA set managed. If Bolt has the ability to push facts directly into pdb from a plan users will be able to leverage the power of PuppetDB without having to run agents or manage more services than required, pdb can even be run on-demand locally inside a container.

## Describe the Solution You Would Like
New function `pdb_command` that posts to PDB's command endpoint.

# Accepts a command, a version, and a payload
`pdb_command('replace_facts', 5, { 'certname' => 'myhost.com'})`

This function should be implemented through a `send_command` method on the pdb_client that can be implemented in orchestrators client implementation. It should fail cleanly if that method is missing.

## Describe Alternatives You've Considered
`pdb::replace_facts` It seems better to abstract over pdb less. We can write puppet language functions or plans on top of a general-purpose command function.

## Solution
- [API spec](https://puppet.com/docs/puppetdb/latest/api/command/v1/commands.html). What can we 'automate' vs making the user set everything? For now, let the user set everything so we don't have to manage different versions, let pdb error appropriately, etc. Can potentially eventually have a puppetdb module that adds a nicer interface around that.
- Don't need to worry about blocking for now.
- Require `command`, `version`, and `payload`. Positional or kwargs? Let's do positional, and we can add a non-positional one later if more keys are needed.
- For the command API, it's command API version 1 which requires certname to be one of the query parameters (along with the command, version, and payload). We should pull the certname out of the payload and pass it to PDB. If the user doesn't specify a certname in their payload just don't include it in the request and let PDB give us an error.
- Returns an object with a UUID key, we should pull that out of the return and return that from the function (which should be just a string).
- Focus testing on `replace_facts` version 5 - document as only supported command and other commands may work
- If it errors, create a Bolt::Error with the error from PDB in the details (or whatever we have for pdb errors already). We've had some issues with stringified JSON inside JSON, so....try to avoid that 😬 
- Document as experimental
- Ping skeletor on the PR (Sean or someone)

Due date: release next week

beechtom:
**Update**

[...]

PR #2781, opened 2021-04-09
(GH-2771) Add puppetdb_command plan function

beechtom:
This adds a new `puppetdb_command` plan function which can be used to
invoke commands in PuppetDB. The function accepts three parameters: the
command, the command version, and the payload. Bolt uses these values to
build a POST request to the command v1 endpoint for PuppetDB. If the
request is successful, the command returns the UUID for the command
invocation, which can be used to reference the command in logs.

Currently, this function only supports and tests the `replace_facts`
command, though other commands might still work.

!feature

* **`puppetdb_command` plan function**
  ([#2771](https://github.com/puppetlabs/bolt/issues/2771))

  The `puppetdb_command` plan function can be used to invoke commands in
  PuppetDB. Currently, only the `replace_facts` command is officially
  tested and supported, though other commands might work as well.

  _This feature is experimental and subject to change._

---

## Bolt acceptance criteria

- [x] Is able to replace facts for a target with an agent
- [x] Is able to add facts for an agentless target
- [x] Does not update facts with a malformed payload
- [x] Returns a UUID for the command on success
- [x] Tries additional server URLs if unable to connect to PDB
- [x] Errors with a helpful message if the PDB client does not implement `:send_message`
- [x] Errors with a helpful message if missing `certname` in the payload

### DIFF
--- a/bolt-modules/boltlib/lib/puppet/functions/puppetdb_command.rb
+++ b/bolt-modules/boltlib/lib/puppet/functions/puppetdb_command.rb
@@ -1,0 +1,66 @@
+# frozen_string_literal: true
+
+require 'bolt/error'
+
+# Send a command with a payload to PuppetDB.
+#
+# The `pdb_command` function only supports version 5 of the `replace_facts`
+# command. Other commands might also work, but are not tested or supported
+# by Bolt.
+#
+# See the [commands endpoint](https://puppet.com/docs/puppetdb/latest/api/command/v1/commands.html)
+# documentation for more information about available commands and payload
+# format.
+#
+# _This function is experimental and subject to change._
+#
+# > **Note:** Not available in apply block
+#
+Puppet::Functions.create_function(:puppetdb_command) do
+  # @param command The command to invoke.
+  # @param version The version of the command to invoke.
+  # @param payload The payload to the command.
+  # @return The UUID identifying the response sent by PuppetDB.
+  # @example Replace facts for a target
+  #   $payload = {
+  #     'certname'           => 'localhost',
+  #     'environment'        => 'dev',
+  #     'producer'           => 'bolt',
+  #     'producer_timestamp' => '1970-01-01',
+  #     'values'             => { 'orchestrator' => 'bolt' }
+  #   }
+  #
+  #   puppetdb_command('replace_facts', 5, $payload)
+  dispatch :puppetdb_command do
+    param 'String[1]', :command
+    param 'Integer', :version
+    param 'Hash[Data, Data]', :payload
+    return_type 'String'
+  end
+
+  def puppetdb_command(command, version, payload)
+    # Disallow in apply blocks.
+    unless Puppet[:tasks]
+      raise Puppet::ParseErrorWithIssue.from_issue_and_stack(
+        Bolt::PAL::Issues::PLAN_OPERATION_NOT_SUPPORTED_WHEN_COMPILING,
+        action: 'puppetdb_command'
+      )
+    end
+
+    # Send analytics report.
+    Puppet.lookup(:bolt_executor).report_function_call(self.class.name)
+
+    puppetdb_client = Puppet.lookup(:bolt_pdb_client)
+
+    # Error if the PDB client does not implement :send_command
+    unless puppetdb_client.respond_to?(:send_command)
+      raise Bolt::Error.new(
+        "PuppetDB client #{puppetdb_client.class} does not implement :send_command, "\
+        "unable to invoke command.",
+        'bolt/pdb-command'
+      )
+    end
+
+    puppetdb_client.send_command(command, version, payload)
+  end
+end
--- a/lib/bolt/puppetdb/client.rb
+++ b/lib/bolt/puppetdb/client.rb
@@ -95,6 +95,60 @@ module Bolt
         make_query(query, path)
       end
 
+      # Sends a command to PuppetDB using version 1 of the commands API.
+      # https://puppet.com/docs/puppetdb/latest/api/command/v1/commands.html
+      #
+      # @param command [String] The command to invoke.
+      # @param version [Integer] The version of the command to invoke.
+      # @param payload [Hash] The payload to send with the command.
+      # @return A UUID identifying the submitted command.
+      #
+      def send_command(command, version, payload)
+        command = command.dup.force_encoding('utf-8')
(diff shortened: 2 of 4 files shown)


beechtom:
First pass at the new function. Known issue is that an invalid payload will return a response with a `200` status code and a UUID for the command, but no changes will be made in PDB.

bolt-modules/boltlib/lib/puppet/functions/puppetdb_command.rb
```diff
@@ -0,0 +1,58 @@
...
+#
+# See the [commands endpoint](https://puppet.com/docs/puppetdb/latest/api/command/v1/commands.html)
+# documentation for more information about available commands and payload
+# format.
+#
+# _This function is experimental and subject to change._
+#
+Puppet::Functions.create_function(:puppetdb_command) do
+  # @param command The command to invoke.
+  # @param version The version of the command to invoke.
```
lucywyman, line 19:
```suggestion
  # @param version The version of the API endpoint to target.
```

bolt-modules/boltlib/lib/puppet/functions/puppetdb_command.rb
```diff
@@ -0,0 +1,58 @@
...
+
+    # Send analytics report. Bolt executor not expected when invoked from
+    # apply block.
+    executor = Puppet.lookup(:bolt_executor) { nil }
+    executor&.report_function_call(self.class.name)
+
+    # Error if the PDB client does not implement :send_command
+    unless puppetdb_client.respond_to?(:send_command)
+      raise Bolt::Error.new(
+        "PuppetDB client #{puppetdb_client.class} does not implement :send_command, "\
```
lucywyman, line 50: Maybe include something like "To use this plan function, please upgrade Bolt to 3.7.0 or later"?

bolt-modules/boltlib/lib/puppet/functions/puppetdb_command.rb
beechtom: This is taken directly from the PDB documentation: https://puppet.com/docs/puppetdb/latest/api/command/v1/commands.html

```
version is a JSON integer describing what version of the given command you’re attempting to invoke.
```

bolt-modules/boltlib/lib/puppet/functions/puppetdb_command.rb
```diff
@@ -0,0 +1,58 @@
...
+  def puppetdb_command(command, version, payload)
+    puppetdb_client = Puppet.lookup(:bolt_pdb_client)
+
+    # Send analytics report. Bolt executor not expected when invoked from
+    # apply block.
+    executor = Puppet.lookup(:bolt_executor) { nil }
+    executor&.report_function_call(self.class.name)
+
+    # Error if the PDB client does not implement :send_command
+    unless puppetdb_client.respond_to?(:send_command)
```
nicklewis, line 48: In what case could this happen?

bolt-modules/boltlib/lib/puppet/functions/puppetdb_command.rb
```diff
@@ -0,0 +1,58 @@
...
+    param 'Integer', :version
+    param 'Hash[Data, Data]', :payload
+    return_type 'String'
+  end
+
+  def puppetdb_command(command, version, payload)
+    puppetdb_client = Puppet.lookup(:bolt_pdb_client)
+
+    # Send analytics report. Bolt executor not expected when invoked from
+    # apply block.
```
nicklewis, line 43: This function probably should be disallowed in an apply block.

bolt-modules/boltlib/lib/puppet/functions/puppetdb_command.rb
beechtom: Part of the ticket was to fail cleanly if `:send_command` isn't implemented on the client, and the orchestrator client implementation of the client was called out specifically. I can remove this bit if that's not actually a concern.

nicklewis: I don't think this version of boltlib will get into a PE version that doesn't have an implementation of the function, so it should be fine.

lucywyman: I had this same question, but now that Bolt usage has proliferated quite a bit within Puppet I'm not confident that this variable will always be returning Bolt's own `puppetdb_client` with the function defined, so it seems safer to leave it in (for now and later).

lib/bolt/puppetdb/client.rb
```diff
@@ -95,6 +95,60 @@ def make_query(query, path = nil)
...
+      # @return A UUID identifying the submitted command.
+      #
+      def send_command(command, version, payload)
+        command = command.dup.force_encoding('utf-8')
+        body    = JSON.generate(payload)
+
+        # PDB requires the following query parameters to the POST request.
+        # Error early if there's no certname, as PDB does not return a
+        # message indicating it's required.
+        unless payload['certname']
```
lucywyman, line 113: In testing the endpoint it seems like all keys are required and should be validated either here or in the plan function itself since PDB doesn't raise an error if the request doesn't include them or has the wrong type. While I'm inclined to say this will most likely be used by people who have used the API before, I also think part of Bolt's mission is to make things easy for users new to the Puppet ecosystem for whom the required keys here might not be obvious.

Required:
- environment, string
- producer_timestamp, string
- values, keys are strings
- optional producer, if not provided add it automatically as 'nil' (since it's required by pdb)

lib/bolt/puppetdb/client.rb
```diff
@@ -95,6 +95,60 @@ def make_query(query, path = nil)
...
+            "Payload must include 'certname', unable to invoke command.",
+            'bolt/pdb-command'
+          )
+        end
+
+        url = uri.tap do |u|
+          u.path         = 'pdb/cmd/v1'
+          u.query_values = { 'command'  => command,
+                             'version'  => version,
+                             'certname' => body['certname'] }
```
lucywyman, line 124:
```suggestion
                             'certname' => payload['certname'] }
```
body here is a JSON string, payload is the actual ruby object. This will always return the string "certname".

beechtom: D'oh. Fixed. I changed the original var name and forgot to update it here.

Issue #2772, opened 2021-04-07
Document using `--no-verbose` with `--stream` for non-duplicated output


lucywyman:
## Describe the Change You Would Like
When using the `--stream` flag run results will be printed twice, once when streaming and again once the run has finished. This can be turned off by setting `--no-verbose` in addition to `--stream`.

### Questions:
- Should we just have verbose be off by default if using `--stream`, and users can override with `--verbose`? Perhaps the solution is to have human-verbose, human-streaming, etc. outputters.  Kind of going down a rabbit hole for now so let's just document `--no-verbose` and leave it at that.
- Document this in the help text for `--stream` specifically, and anywhere else that's useful.

PR #2783, opened 2021-04-12
(GH-2772) Document using --no-verbose with --stream

beechtom:
This adds documentation and updates Bolt's help text to suggest using
`--no-verbose` with `--stream` to prevent Bolt from printing output from
a target twice.

!no-release-note


lucywyman (approved):

lucywyman:
"standard out" still feels weird but it's not a hill I want to die on.

hestonhoffman (approved):

Issue #2721, opened 2021-03-16
Add links to documentation to the `bolt help` output


lucywyman:
## Use Case
It'd be useful to include helpful links to documentation and instructions for making an issue to the help text so that users can easily access those resources.

## Questions
* Should we also have a `bolt guide links` or `bolt guide help` or something to print this information?

PR #2786, opened 2021-04-13
(GH-2721) Add links and prompts to run 'guide' to Bolt help text

lucywyman:
This adds relevant links and recommendations to run 'bolt guide' to
several Bolt help text outputs in an effort to help Bolt users find
relevant documentation more quickly, and on the CLI if they can.

!no-release-note

### DIFF
--- a/guides/guide.txt
+++ b/guides/guide.txt
@@ -1,0 +1,17 @@
+TOPIC
+    guide
+
+DESCRIPTION
+    A guide is a person (or CLI tool ;D) who leads travelers through unknown or
+    unfamiliar locations. The term can also be applied to a person who leads others
+    to more abstract goals such as knowledge or wisdom. 
+
+    Etymology: Originated sometime between 1325 and 1375. From Middle English
+    guide, from the Old French guide, from Old Occitan guida, from guidar, from
+    Frankish *wītan (“to show the way, lead”), from Proto-Germanic *wītaną (“to
+    see, know; go, depart”), from Proto-Indo-European *weyd- (“to see, know”).
+    Related also to English wit. 
+
+DOCUMENTATION
+    https://en.wikipedia.org/wiki/Guide
+    https://en.wiktionary.org/wiki/guide
--- a/guides/targets.txt
+++ b/guides/targets.txt
@@ -6,22 +6,20 @@ DESCRIPTION
     be physical, such as servers, or virtual, such as containers or virtual
     machines.
 
-    Several of Bolt's commands connect to targets and run actions on them. When
-    you run these commands, Bolt requires a list of targets to run the action
-    on. For example, the `bolt script run` command and `Invoke-BoltScript`
-    PowerShell cmdlet require you to provide a list of targets to run your
-    script on. You can provide a list of targets to a command using one of the
-    following command-line options:
+    Several of Bolt's commands connect to targets and run actions on them.
+    These commands require a target or targets to run on. You can specify
+    targets to a command using one of the following command-line options:
 
-    Shell commands
-        -t, --targets TARGETS
-        -q, --query   QUERY
-            --rerun   FILTER
+    *nix options                Powershell options
+     -t, --targets TARGETS      -T, -Targets  TARGETS
+     -q, --query   QUERY        -Q, -Query    QUERY
+         --rerun   FILTER       -Rerun        FILTER
 
-    PowerShell cmdlets
-        -T, -Targets TARGETS
-        -Q, -Query   QUERY
-            -Rerun   FILTER
+    The 'targets' option accepts a comma-separated list of target URIs or group
+    names, or can read a target list from an input file '@<file>' or stdin '-'.
+    URIs can be specified with the format [protocol://][user@]host[:port]. To
+    learn more about available protocols and their defaults, run 'bolt guide
+    transports'.
 
     Typically, targets and their configuration and data are listed in a
     project's inventory file. For more information about inventory files,
--- a/guides/transports.txt
+++ b/guides/transports.txt
@@ -1,0 +1,23 @@
+TOPIC
+    transports
+
+DESCRIPTION
+    Bolt uses transports (also known as protocols) to establish a connection
+    with a target in order to run actions on the target. The default transport is
+    SSH, and you can see available transports along with their configuration
+    options and defaults at http://pup.pt/bolt-reference.
+
+    You can specify a transport for a target by prepending '<transport>://' to
+    the target's URI. For example, to connect to a target with hostname
+    'example.com' as user 'Administrator' using the WinRM transport, you would
+    pass the following to the target flag:
+        winrm://Administrator@example.com
+
+    You can also specify a default transport for all targets by passing the
+    '--transport' flag on *nix systems and the '-Transport' flag in Powershell.
+    Finally, you can set the transport for a target in the inventory. For more
+    information about the Bolt inventory, run 'bolt guide inventory'.
+
+DOCUMENTATION
+    https://pup.pt/bolt-commands#specify-a-transport
+    http://pup.pt/bolt-inventory#transport-configuration
--- a/lib/bolt/bolt_option_parser.rb
+++ b/lib/bolt/bolt_option_parser.rb
@@ -156,18 +156,30 @@ module Bolt
       end
     end
 
+    COLORS = {
+      cyan: "36"
+    }.freeze
+
+    def self.colorize(color, string)
+      if $stdout.isatty
+        "\033[#{COLORS[color]}m#{string}\033[0m"
+      else
+        string
+      end
+    end
+
     BANNER = <<~HELP
-      NAME
+      #{colorize(:cyan, 'Name')}
           bolt
 
-      USAGE
+      #{colorize(:cyan, 'Usage')}
           bolt <subcommand> [action] [options]
 
-      DESCRIPTION
+      #{colorize(:cyan, 'Description')}
           Bolt is an orchestration tool that automates the manual work it takes to
           maintain your infrastructure.
 
-      SUBCOMMANDS
+      #{colorize(:cyan, 'Subcommands')}
           apply             Apply Puppet manifest code
           command           Run a command remotely
           file              Copy files between the controller and targets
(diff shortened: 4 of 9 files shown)


guides/targets.txt
```diff
@@ -6,26 +6,24 @@ DESCRIPTION
...
-    following command-line options:
+    Several of Bolt's commands connect to targets and run actions on them.
+    These commands require a target or targets to run on. You can specify
+    targets to a command using one of the following flags:
 
-    Shell commands
-        -t, --targets TARGETS
-        -q, --query   QUERY
-            --rerun   FILTER
+    *nix flags                  Powershell flags
```
beechtom, line 13: We refer to these as command-line options in docs.

guides/guide.txt
```diff
@@ -0,0 +1,18 @@
+TOPIC
```
beechtom, line 1: While fun, I think this clutters the guides list more than anything.

guides/targets.txt
```diff
@@ -6,26 +6,24 @@ DESCRIPTION
...
 
-    PowerShell cmdlets
-        -T, -Targets TARGETS
-        -Q, -Query   QUERY
-            -Rerun   FILTER
+    The 'targets' flag accepts a comma-separated list of target URIs or group
+    names, or can read a target list from an input file '@<file>' or stdin '-'.
+    URIs can be specified with the format [protocol://][user@]host[:port]. To
+    learn more about available protocols and their defaults, run 'bolt guide
+    transports'.
```
beechtom, line 22: This information is already present in https://puppet.com/docs/bolt/latest/running_bolt_commands.html . I think we should prefer linking to specific information like this when we can so we don't duplicate information.

guides/transports.txt
```diff
@@ -0,0 +1,24 @@
+TOPIC
+    transports
+
+DESCRIPTION
+    Bolt uses transports (also known as protocols) to establish a connection
+    with a target in order to run actions on the target. The default transport is
+    SSH, with available transports docker, local, lxd, pcp, podman, remote,
+    ssh, and winrm. Each transport has it's own set of configuration options and
```
beechtom, line 8:
```suggestion
    with a target in order to run actions on the target. The default transport is
    SSH. Each transport has it's own set of configuration options and
```

Just so we don't need to continually update this guide when we add transports.

beechtom, line 8:
```suggestion
    with a target in order to run actions on the target. The default transport is
    SSH, with available transports docker, local, lxd, pcp, podman, remote,
    ssh, and winrm. Each transport has its own set of configuration options and
```

guides/transports.txt
```diff
@@ -0,0 +1,24 @@
...
+    defaults, which you can see in our reference documentation at
```
beechtom, line 9:
```suggestion
    defaults, which you can see in the reference documentation at
```

guides/transports.txt
```diff
@@ -0,0 +1,24 @@
...
+
+    You can specify a transport for a target by prepending '<transport>://' to
+    the target's URI. For example, to connect to a target with hostname
+    'example.com' as user 'Administrator' using the WinRM transport, you would
+    pass the following to the target flag:
+        winrm://Administrator@example.com
+    You can also specify a default transport for all targets by passing the
+    '--transport' flag on *nix systems and the '-Transport' flag in Powershell.
+    Finally, you can set the transport for a target in the inventory. For more
+    information about the Bolt inventory, run 'bolt guide inventory'.
```
beechtom, line 20: This information is also covered in https://puppet.com/docs/bolt/latest/running_bolt_commands.html . A link to that page might be better.

lib/bolt/bolt_option_parser.rb
```diff
@@ -156,18 +156,30 @@ def get_help_text(subcommand, action = nil)
...
+        "\033[#{COLORS[color]}m#{string}\033[0m"
+      else
+        string
+      end
+    end
+
     BANNER = <<~HELP
-      NAME
+      #{colorize(:cyan, 'NAME')}
           bolt
```
beechtom, line 173:
```suggestion
      #{colorize(:cyan, 'Name')}
        bolt
```

I'm guessing colorizing the headers is to bring this more in line with the show commands? If so, we should use sentence casing for the headers and decrease the indent for the rest of the text to 2 instead of 4.

lib/bolt/bolt_option_parser.rb
```diff
@@ -181,334 +193,381 @@ module            Manage Bolt project modules
...
+      #{colorize(:cyan, 'USAGE')}
           bolt apply [manifest] {--targets TARGETS | --query QUERY | --rerun FILTER}
             [options]
 
-      DESCRIPTION
+      #{colorize(:cyan, 'DESCRIPTION')}
           Apply Puppet manifest code on the specified targets.
 
-      EXAMPLES
+      #{colorize(:cyan, 'DOCS')}
```
beechtom, line 212:
```suggestion
      #{colorize(:cyan, 'Documentation')}
```

Prefer spelling the entire word out.

guides/guide.txt
lucywyman: I thought about removing it and making it even more of an easter egg - would that work?

beechtom: I think that'd work :)

guides/transports.txt
lucywyman: Hm, I don't see the point of the guides being entirely separate from documentation, but rather providing just enough information to hopefully answer user questions without having to go to the docs and then linking to docs if users need to know more. For example targeting flags are also documented https://puppet.com/docs/bolt/latest/running_bolt_commands.html#specify-targets but we opted to include those flags in the `targets` guide.

guides/guide.txt
lucywyman: Also definitely open to more whimsical text here - I mostly wanted to avoid getting too close to https://forge.puppet.com/modules/binford2k/easter_egg

guides/transports.txt
```diff
@@ -0,0 +1,22 @@
...
+    with a target in order to run actions on the target. The default transport is
+    SSH, and you can see available transports along with their configuration
+    options and defaults at http://pup.pt/bolt-reference.
+
+    You can specify a transport for a target by prepending '<transport>://' to
+    the target's URI. For example, to connect to a target with hostname
+    'example.com' as user 'Administrator' using the WinRM transport, you would
+    pass the following to the target flag:
+        winrm://Administrator@example.com
+    You can also specify a default transport for all targets by passing the
```
beechtom, line 15:
```suggestion

    You can also specify a default transport for all targets by passing the
```

lib/bolt/bolt_option_parser.rb
```diff
@@ -714,120 +812,108 @@ def initialize(options)
...
-             'Enter a comma-separated list of target URIs or group names.',
-             "Or read a target list from an input file '@<file>' or stdin '-'.",
-             'Example: --targets localhost,target_group,ssh://nix.com:23,winrm://windows.puppet.com',
-             'URI format is [protocol://]host[:port]',
-             "SSH is the default protocol; can be #{TRANSPORTS.keys.join(', ')}",
-             'For Windows targets, specify the winrm:// protocol if it has not be configured',
-             'For SSH, port defaults to `22`',
-             'For WinRM, port defaults to `5985` or `5986` based on the --[no-]ssl setting',
+      separator "\n#{self.class.colorize(:cyan, 'Inventory options')}"
+      define('-t', '--targets TARGETS', 'Identifies the targets of command.',
```
beechtom, line 816:
```suggestion
      define('-t', '--targets TARGETS', 'Identifies the targets of the command.',
```

beechtom (approved):
Two small nits, but otherwise lgtm!

beechtom:
Needs a rebase

guides/guide.txt
```diff
@@ -0,0 +1,18 @@
...
+    guide
+
+DESCRIPTION
+    A guide is a person who leads travelers, sportsmen, or tourists through
+    unknown or unfamiliar locations. The term can also be applied to a person who
+    leads others to more abstract goals such as knowledge or wisdom.
```
hestonhoffman, line 7:
```suggestion
    A guide is a person who leads travelers through
    unknown or unfamiliar locations. The term can also be applied to a person who
    leads others to more abstract goals such as knowledge or wisdom. 
```
I think just leaving this at travelers is more effective (and I want to remove the gendered "sportsmen")

guides/guide.txt
```diff
@@ -0,0 +1,18 @@
...
+    A guide is a person who leads travelers, sportsmen, or tourists through
+    unknown or unfamiliar locations. The term can also be applied to a person who
+    leads others to more abstract goals such as knowledge or wisdom. 
+
+    Etymology: Originated sometime between 1325 and 1375. From Middle English
+    guide, from the Old French guide, from Old Occitan guida, from guidar, from
+    Frankish *wītan (“to show the way, lead”), from Proto-Germanic *wītaną (“to
+    see, know; go, depart”), from Proto-Indo-European *weyd- (“to see, know”).
+    Cognate with Old English wītan (“to see, take heed to, watch after, guard,
+    keep”). Related also to English wit.
```
hestonhoffman, line 14:
```suggestion
    Etymology: Originated sometime between 1325 and 1375. From Middle English
    guide, from the Old French guide, from Old Occitan guida, from guidar, from
    Frankish *wītan (“to show the way, lead”), from Proto-Germanic *wītaną (“to
    see, know; go, depart”), from Proto-Indo-European *weyd- (“to see, know”).
    Cognate with Old English wītan (“to see, take heed to, watch after, guard,
    keep”). Related also to English wit. 
```
I feel like we're going into the weeds a little here? It's whimsical, so if that's the intent, I'm fine with it, but I wonder if we should put a paragraph above this one that tells people what Bolt guides are for? 
Something like:
> Several of Bolt's commands offer guides that include basic usage information, command options, and links to documentation.

hestonhoffman: Maybe I'm just missing the joke cos it's Monday morning and I shouldn't do PR reviews first thing on a Monday morning :D

lucywyman: Hmm, the intent is definitely to be whimsical more than to be helpful - this guide won't even show in the `bolt guide` output that lists available guides, so it's intended to be an easter egg. Is there a way to make that more clear do you think? Maybe some ASCII art? Definitely still open to a short graf about bolt commands!

lucywyman: Haha, I mean, there's not *that* much of a joke to get, this is just copied from the linked wikis. I think as an easter egg it could use some more creativity 😅

hestonhoffman: Yeah, gotcha :D I figured it was an easter egg after I'd had some coffee. ASCII art would be amazing TBH!

lucywyman: Hm, I noodled on this for a little bit and at this point am out of creativity - I think we should either cut this and add it later, or just merge as is :/ @beechtom thoughts on which?

hestonhoffman: You can always merge it as is and then iterate on it for a hack day or something. This is the kind of thing I could hack at too :D

beechtom: Agree with Heston, I think we can merge this for now and iterate on it later, especially since it's just an easter egg. :)

hestonhoffman (approved):

PR #2787, opened 2021-04-13
(hack) Autogenerate type reference doc

beechtom:
This adds puppet-strings documentation to Bolt's built-in data types and
adds some missing tags for built-in functions. With the addition of
puppet-strings to Bolt's data types, the type reference page is now
autogenerated.

!no-release-note

---

This is a hack time project from a couple weeks back that needed updating to include the `ContainerResult` data type. :)


rakelib/docs.rake
```diff
@@ -64,6 +66,32 @@ begin
...
+        FileUtils.mkdir_p('tmp')
+        tmpfile = 'tmp/boltlib.json'
+
+        PuppetStrings.generate(
+          ['bolt-modules/*'],
+          markup: 'markdown',
+          json: true,
+          path: tmpfile,
+          yard_args: ['bolt-modules/boltlib',
+                      'bolt-modules/ctrl',
```
lucywyman, line 84: Whoops, I think we missed `dir` the first time around here.

bolt-modules/boltlib/lib/puppet/datatypes/applyresult.rb
```diff
@@ -1,5 +1,31 @@
 # frozen_string_literal: true
 
+# An [apply action](applying_manifest_blocks.md#return-value-of-apply-action)
+# returns an `ApplyResult`. An `ApplyResult` is part of a `ResultSet` object and
+# contains information about the apply action.
+#
+# @param report
+#   The Puppet report from the apply action. Equivalent to calling `$apply_result.value['report']`.
+#   The report is a hash representation of the [`Puppet::Transaction::Report`
+#   object](https://puppet.com/docs/puppet/7.0/format_report.html), where each property
```
lucywyman, line 10:
```suggestion
#   object](https://puppet.com/docs/puppet/latest/format_report.html), where each property
```

rakelib/docs.rake
```diff
@@ -196,25 +224,69 @@ begin
...
+            'type' => param['types'].first
+          }
+        end
+
+        # Add functions
+        data['functions'].each do |func|
+          if data['name'] == 'ResultSet' && func['name'] == '[]'
+            type = 'Variant[Result, ApplyResult, Array[Variant[Result, ApplyResult]]]'
+          else
+            type = func.dig('docstring', 'tags').first['types']
```
lucywyman, line 259: There are a number of other cases where the first type is the type of a parameter to the function, not the return type. I think it's safest to get `.last` here instead and verify that the `tag_name` is `return`.

rakelib/docs.rake
beechtom: :ohmy: Added

bolt-modules/boltlib/lib/puppet/datatypes/applyresult.rb
```diff
@@ -1,5 +1,31 @@
...
+#   The Puppet report from the apply action. Equivalent to calling `$apply_result.value['report']`.
+#   The report is a hash representation of the [`Puppet::Transaction::Report`
+#   object](https://puppet.com/docs/puppet/latest/format_report.html), where each property
+#   corresponds to a key in the report hash. For more information, see [Result
+#   keys](applying_manifest_blocks.md#result-keys).
+# @param target
+#   The target the result is from.
+#
+# @!method action
+#   The action performed. `ApplyResult.action` always returns the string `apply`.
```
lucywyman, line 17: Hm, this got updated from `$apply_result.action` to `ApplyResult.action`, but we didn't do the same above for `$apply_result.value['report']`. I honestly don't have a preference, but they should be consistent.

rakelib/docs.rake
```diff
@@ -196,25 +225,71 @@ begin
...
+            val
+          end
+        end
+
+        # Add attributes, which are listed under the type's params
+        data.dig('docstring', 'tags').each do |param|
+          functions << {
+            'name' => param['name'],
+            'desc' => param['text'].tr("\n", ' '),
+            'type' => param['types'].first
```
lucywyman, line 251: It'd be nice to strip the `[1]` from Strings here if we can, including in types like `Variant[String[1], Type[Resource]]`. I think the best way to achieve that is with a regex: `.gsub(/String\[1\]/)`. I think we should only do that for `String[1]` though, if we have a type that requires a minimum character length of more than that that seems important to call out here.

rakelib/docs.rake
```diff
@@ -196,25 +225,71 @@ begin
...
+            type = func.dig('docstring', 'tags')
+                       .find { |tag| tag['tag_name'] == 'return' }
+                       .fetch('types')
+            type = type.first if type.is_a?(Array)
+          end
+
+          functions << {
+            'name' => func['name'],
+            'desc' => func.dig('docstring', 'text').tr("\n", ' '),
+            'type' => type
```
lucywyman, line 269: Same `[1]` stripping here.

lucywyman (approved):
One small improvement, this is awesome Tom!

lucywyman (approved):

Issue #2740, opened 2021-03-25
Win32API deprecation warning


beechtom:
## Describe the Bug

Running Bolt on a Windows controller and connecting to a target with the SSH transport results in the following warning being displayed to the console:

```
C:/Program Files/Puppet Labs/Bolt/lib/ruby/gems/2.7.0/gems/bolt-3.4.0/lib/bolt/transport/ssh.rb:34: warning: Win32API is deprecated after Ruby 1.9.1; use fiddle directly instead
```

PR #2790, opened 2021-04-15
(GH-2740) Use fiddle instead of Win32API

beechtom:
This replaces the use of the deprecated `Win32API` gem in SSH
connections on Windows with the use of `fiddle` instead.

!no-release-note

---

### Testing

Testing this PR requires running Bolt on a Windows machine and connecting to a target with the SSH transport.

**On main branch**

- Confirm that connecting to a target using the SSH transport results in the following warning:
   ```
   C:/Users/Administrator/bolt/lib/bolt/transport/ssh/connection.rb:118: warning: Win32API is deprecated after Ruby 1.9.1; use fiddle directly instead
   ```

**On feature branch**

- Connect to a target using the SSH transport while Pageant is not running. Check the debug logs to confirm that `use_agent = false` is passed to net-ssh.
- Install PuTTY and then start the Pageant process (you should see an icon of a computer with a hat in the system tray). Connect to a target using the SSH transport. Check the debug logs to confirm that `use_agent` was not disabled.

### Acceptance criteria

- [x] Does not warn that Win32API is deprecated
- [x] Passes `use_agent = false` to net-ssh if Pageant process is not running
- [x] Does not print a warning from net-ssh that Pageant process is not running
- [x] Does not pass `use_agent = false` to net-ssh if Pageant process is running
- [x] Picks up SSH configuration from Pageant

### DIFF
--- a/lib/bolt/transport/ssh/connection.rb
+++ b/lib/bolt/transport/ssh/connection.rb
@@ -39,8 +39,6 @@ module Bolt
           end
         end
 
-        PAGEANT_NAME = "Pageant\0".encode(Encoding::UTF_16LE)
-
         def connect
           options = {
             logger: @transport_logger,
@@ -115,10 +113,7 @@ module Bolt
                 options[:use_agent] = false
               end
             elsif Bolt::Util.windows?
-              require 'Win32API' # case matters in this require!
-              # https://docs.microsoft.com/en-us/windows/desktop/api/winuser/nf-winuser-findwindoww
-              @find_window ||= Win32API.new('user32', 'FindWindowW', %w[P P], 'L')
-              if @find_window.call(nil, PAGEANT_NAME).to_i == 0
+              unless pageant_running?
                 @logger.debug { "Disabling use_agent in net-ssh: pageant process not running" }
                 options[:use_agent] = false
               end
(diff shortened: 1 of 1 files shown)


lib/bolt/transport/ssh/connection.rb
```diff
@@ -318,6 +313,45 @@ def validate_ssh_version
...
+            )
+
+            # Pack the name of the window into a pointer, and then decode that
+            # pointer as a native-endian signed long.
+            # https://apidock.com/ruby/Array/pack
+            # https://apidock.com/ruby/String/unpack
+            pageant, = ["Pageant\0".encode(Encoding::UTF_16LE)].pack("p").unpack("l!*")
+
+            # Find the 'Pageant' process using the 'FindWindowW' function.
+            status, = find_window.call(0, pageant)
```
lucywyman, line 348: Is there a case where this returns an array? It seems like it should always be just the handle to the window.

lucywyman (approved):
\o/ Awesome job on this!!

lib/bolt/transport/ssh/connection.rb
beechtom: Ah, no, it should just be the handle. Updated.

Issue #2791, opened 2021-04-16
Bolt fails to find containers when given a full ID


Sharpie:
## Describe the Bug

When told to connect to containers using a full ID instead of a truncated one, e.g. `81f3897a794b8a0e5411cf3cfd168ce4f7430bb32c96f8446048c642853030fe` instead of `81f3897a794b`, Bolt fails to find the container. This appears to be caused by Bolt running `docker ps --format {{json .}}` which returns the short ID unless `--no-trunc` is used. The IDs are then compared using an equality check, which fails:

https://github.com/puppetlabs/bolt/blob/3.7.0/lib/bolt/transport/docker/connection.rb#L49-L51

Short IDs are common, but provisioning tools like Terraform will return long IDs.

## Expected Behavior

Bolt is able to find containers using both short and long IDs.

## Steps to Reproduce

  - Set up a docker container:

```
docker run -d --name bolt-test docker/getting-started
```

  - Try to run a command on the container using its long ID:

```
$ bolt command run -t "docker://$(docker ps -q --no-trunc --filter 'name=bolt-test')" 'hostname -f'

Started on docker://8703dc222e64d87b8d639e9aec06c37942c9497a86abea118870149ebbb73895...
Failed on docker://8703dc222e64d87b8d639e9aec06c37942c9497a86abea118870149ebbb73895:
  Failed to connect to docker://8703dc222e64d87b8d639e9aec06c37942c9497a86abea118870149ebbb73895: Could not find a container with name or ID matching '8703dc222e64d87b8d639e9aec06c37942c9497a86abea118870149ebbb73895'
Failed on 1 target: docker://8703dc222e64d87b8d639e9aec06c37942c9497a86abea118870149ebbb73895
```

 - Omitting the `--no-trunc` flag causes `docker ps` to return the short ID, which succeeds:

```
$ bolt command run -t "docker://$(docker ps -q --filter 'name=bolt-test')" 'hostname -f'
Started on docker://8703dc222e64...
Finished on docker://8703dc222e64:
  8703dc222e64
Successful on 1 target: docker://8703dc222e64
Ran on 1 target in 0.61 sec
```
## Environment
 - Version: 3.7.0
 - Platform: macOS 10.14

Sharpie:
I'm guessing this affects Podman as well. Not sure about LXC.

Sharpie:
Looks like the podman connection class handles this by using `start_with?`

[...]

PR #2792, opened 2021-04-16
Allow Docker to connect with long IDs

Sharpie:
Currently, Docker connections only succeed when the 12 character
"short ID" of a container is used. When a full container ID is
provided, connections fail despite the container being present
and running. This is because the Docker transport searches for the
container ID in the output of `docker ps` and `docker ps` only returns
short ID values unless the `--no-trunc` flag is used.

This commit updates the connection logic to pass the `--no-trunc` flag
and match container IDs using `start_with?` instead of an equality
check.

Fixes puppetlabs/bolt#2791.

### DIFF
--- a/lib/bolt/transport/docker/connection.rb
+++ b/lib/bolt/transport/docker/connection.rb
@@ -46,8 +46,8 @@ module Bolt
         def connect
           # We don't actually have a connection, but we do need to
           # check that the container exists and is running.
-          output = execute_local_json_command('ps')
-          index = output.find_index { |item| item["ID"] == target.host || item["Names"] == target.host }
+          output = execute_local_json_command('ps', ['--no-trunc'])
+          index = output.find_index { |item| item["ID"].start_with?(target.host) || item["Names"] == target.host }
           raise "Could not find a container with name or ID matching '#{target.host}'" if index.nil?
           # Now find the indepth container information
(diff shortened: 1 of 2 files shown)


lucywyman:
@Sharpie This looks great, works for me! Can you add a release note to the commit message (instructions in [CONTRIBUTING](https://github.com/puppetlabs/bolt/blob/main/CONTRIBUTING.md#pull-requests), [example](https://github.com/puppetlabs/bolt/pull/2778)). I think for this it would be:

```
!bug

* **Allow Docker to connect when provided long container ID** ([#2791](https://github.com/puppetlabs/bolt/issues/2791))
  Previously, Docker connections only succeed when the 12-character "short ID" of a container is used. Bolt now accepts the long form of the container ID as the container host.
```

lucywyman (approved):
- [x] Connects to a target when provided the long id: `docker://61c0d09a2e2f84f91e4a67b43c2215d9244fb33c27e2ea58139a3b1958765196`
- [x] Connects to a target when provided the short id: `docker://61c0d09a2e2f`
- [x] Connects to a target when provided the container name: `docker://puppet_7_node`

Sharpie:
The commit message already has a `!bug` tag --- it's missing from the PR description because I added it after the check failed.

Issue #2785, opened 2021-04-13
Improve git module errors when installing modules


beechtom:
## Use Case
Since introducing the `resolve` key for module specs, Bolt can install modules from private and non-GitHub repositories. However, this information can be hard to find for users who may not read through the module documentation completely. We should update some of the error messages in Bolt to direct users to the appropriate documentation for using the `resolve` key.

The following error messages should be updated:
- https://github.com/puppetlabs/bolt/blob/main/lib/bolt/module_installer/specs/git_spec.rb#L71
- https://github.com/puppetlabs/bolt/blob/main/lib/bolt/module_installer/specs/git_spec.rb#L161

The messages should include a short link to https://puppet.com/docs/bolt/latest/bolt_installing_modules.html#skip-dependency-resolution-for-a-module

beechtom:
Closed by #2793

PR #2793, opened 2021-04-16
(GH-2785) Improve error message when provided git repo isn't available

lucywyman:
When specifying modules to install, when a user provides a git module if
the module cannot be retrieved we raise an error. Previously the error
simply stated that the provided repo was invalid, without specifying
that private repositories cannot have their dependencies resolved and
must follow a documented workflow. This updates the error message to
state that the repo is either not a Git repo or is private, and link to
documentation describing how to install modules that include a private
Git repo without installing dependencies.

!no-release-note

### DIFF
--- a/lib/bolt/module_installer/specs/git_spec.rb
+++ b/lib/bolt/module_installer/specs/git_spec.rb
@@ -67,8 +67,7 @@ module Bolt
                  elsif git.start_with?('https://github.com')
                    git.split('https://github.com/').last.split('.git').first
                  else
-                   raise Bolt::ValidationError,
-                         "Invalid git source: #{git}. Only GitHub modules are supported."
+                   raise Bolt::ValidationError, invalid_git_msg(git)
                  end
 
           [git, repo]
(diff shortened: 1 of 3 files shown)


lib/bolt/module_installer/specs/git_spec.rb
```diff
@@ -89,6 +88,15 @@ def to_hash
           }
         end
 
+        # Returns an error message that the provided repo is not a git repo or
+        # is private.
+        #
+        private def invalid_git_msg(repo_name)
+          "#{repo_name} is not a git repository or is private. If the repository is private, "\
+            "configure Bolt to not resolve dependencies for this module. See "\
+            "https://pup.pt/no-resolve for more information."
```
beechtom, line 97: This error will show for public non-GitHub repos as well and might be misleading since it implies that only private repos can use this approach. Maybe updating the message to say "is not a public GitHub repository" instead of "is not a git repository or is private"?

lucywyman: Hm, I can see that. Gave it another shot!

beechtom (approved):
One failing test that needs its regex updated, but otherwise lgtm! :D

Issue #2499, opened 2020-12-18
Command to lookup value with Hiera


beechtom:
## Use Case

Bolt supports the `lookup()` Puppet function in plans, both inside and outside of apply blocks. However, checking which value Hiera will return for a given key can be difficult with Bolt, as there is no command available to perform a lookup. Being able to do a lookup with Hiera with a Bolt command would make it easier for users to know which values are returned given specific inputs. It would be especially helpful if a user could also use the Bolt command to check what value a lookup using the `hierarchy` config would return vs. the `plan_hierarchy` config, as they can return different values.

## Describe the Solution You Would Like

Similar to the `puppet lookup` command, Bolt should have a `bolt lookup` command. This command should accept a positional argument that is the `key` to lookup.

```shell
bolt lookup foo
```

### Lookups using `plan_hierarchy`

**Should this command support lookups using `plan_hierarchy`?**
Most likely. Looking up values for plan_hierarchy seems like the most common use case for this command, as the `puppet lookup` command is not a substitute.

**Lookups using `plan_hierarchy` currently support interpolation. How should Bolt handle specifying variables for interpolation?**
Similar to specifying parameters to a plan. These can then be passed along to the compiler and set in scope. Variables could be passed using `var=value` syntax or as a JSON file.

```shell
bolt lookup foo foo=value1 bar=value2
```

```shell
bolt lookup foo --vars vars.json
```

### Lookups using `hierarchy`

**Should this command support lookups using `hierarchy`?**
Possibly. `puppet lookup --confdir <project>` may be a suitable substitute. However, this does require users who only use Bolt to use Puppet, which isn't great if we want Bolt to have a 'complete' experience.

**Lookups using `hierarchy` support interpolation using facts. How should Bolt handle specifying facts for interpolation?**
A few ways to do this. If using real facts, Bolt will need to fetch facts for the targets first. This is easy enough to do by running the facts task and setting them on the targets before performing the lookups. We could also support specifying facts, which would be helpful for users to test return values for specific facts.

To pass facts we could potentially pass them similar to passing parameters to a plan. We could also support passing facts as a JSON file, which is a pattern already used by `puppet lookup`.

```shell
bolt lookup foo --targets all os.version=1 os.name=linux
```

```shell
bolt lookup foo --targets all --facts facts.json
```

[...]

PR #2801, opened 2021-04-22
(GH-2499) Add 'lookup' command

beechtom:
This adds a new `lookup` command to Bolt that looks up values with
Hiera. The command first runs `apply_prep` on each target to install the
puppet-agent package and collect facts, then sets up Hiera and looks for
the specified key in the hierarchy. The command requires a targeting
option, and the Hiera lookup is performed in the context of each target
passed to the command, allowing for interpolations using facts.

!feature

* **`lookup` command to look up values with Hiera**
  ([#2499](https://github.com/puppetlabs/bolt/issues/2499))

  The new `bolt lookup` and `Invoke-BoltLookup` commands can be used to
  look up values with Hiera.

### DIFF
--- a/documentation/hiera.md
+++ b/documentation/hiera.md
@@ -45,11 +45,79 @@ data set by a user overrides the default data set by the module’s author.
 
 ## Look up data from the command line
 
-If you have used Puppet, you might be familiar with the [`puppet lookup`
-command](https://puppet.com/docs/puppet/latest/hiera_automatic.html#using_puppet_lookup),
-which is the command line interface for the `lookup()` function. Bolt does not
-have an equivalent command line interface, so it is not possible to look up data
-with Hiera outside of running your code in Bolt.
+You can use the `bolt lookup` command and `Invoke-BoltLookup` PowerShell cmdlet
+to look up data from the command line. The `lookup` command looks up data in the
+context of a target, allowing you to interpolate target facts and variables in
+your hierarchy.
+
+> **Note:** The `bolt lookup` and `Invoke-BoltLookup` commands only look up data
+> using the `hierarchy` key in the Hiera configuration file. `plan_hierarchy`
+> is not supported from the command line.
+
+When you run the `bolt lookup` and `Invoke-BoltLookup` commands, Bolt first
+runs an `apply_prep` on each of the targets specified. This installs the
+`puppet-agent` package on the target, collects facts, and then stores the facts
+on the target to be used in interpolations.
+
+Looking up data from the command line is particularly useful if you need to
+debug a plan that includes calls to the `lookup()` function, or if you need to
+look up target-specific data such as a password for authenticating connections
+to the target.
+
+Given the following Hiera configuration at `<PROJECT DIRECTORY>/hiera.yaml`:
+
+```yaml
+# hiera.yaml
+version: 5
+
+hierarchy:
+  - name: "Per-OS defaults"
+    path: "os/%{facts.os.name}.yaml"
+  - name: "Common data"
+    path: "common.yaml"
+```
+
+And the following data source at `<PROJECT DIRECTORY>/data/os/Windows.yaml`:
+
+```yaml
+# data/os/Windows.yaml
+password: Bolt!
+```
+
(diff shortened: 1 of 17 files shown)


hestonhoffman (approved):
Docs LGTM!

documentation/hiera.md
```diff
@@ -45,11 +45,79 @@ data set by a user overrides the default data set by the module’s author.
...
+```
+
+And the following data source at `<PROJECT DIRECTORY>/data/os/Windows.yaml`:
+
+```yaml
+# data/os/Windows.yaml
+password: Bolt!
+```
+
+And the following data source at `<PROJECT DIRECTORY>/data/os/Ubuntu.yaml`:
```
lucywyman, line 87:
```suggestion
And the following data source at `<PROJECT DIRECTORY>/data/os/Debian.yaml`:
```

documentation/hiera.md
```diff
@@ -45,11 +45,79 @@ data set by a user overrides the default data set by the module’s author.
...
+
+```yaml
+# data/os/Ubuntu.yaml
```
lucywyman, line 90:
```suggestion
# data/os/Debian.yaml
```

lib/bolt/cli.rb
```diff
@@ -196,7 +197,7 @@ def parse_command
 
       # Default to verbose for everything except plans
       unless options.key?(:verbose)
-        options[:verbose] = options[:subcommand] != 'plan'
+        options[:verbose] = !%w[lookup plan].include?(options[:subcommand])
```
lucywyman, line 200: Does `verbose` actually impact the output? If not I think we can omit this, we don't exclude it for other commands that don't use `verbose`.

spec/bolt/cli_spec.rb
```diff
@@ -169,6 +169,42 @@ def stub_config(file_content = {})
     end
   end
 
+  context 'lookup' do
+    let(:pal)     { double('pal', lookup: results) }
+    let(:results) { Bolt::ResultSet.new([]) }
+
+    around(:each) do |example|
```
lucywyman, line 176: I don't think this needs to be in a custom project since it doesn't use config.

spec/bolt/outputter/human_spec.rb
```diff
@@ -249,6 +249,12 @@
     expect(lines).to match(/^    "key": "val"$/)
   end
 
+  it 'prints lookup results' do
+    result = Bolt::Result.for_lookup(target, 'key', 'value')
+    outputter.print_result(result)
+    expect(output.string).to match(/value/)
```
lucywyman, line 255:
```suggestion
    expect(output.string).to eq("Finished on target1:\n  value\n")
```
Or can do `#{target.name}` if you're worried about the name changing, but I wouldn't be.

lib/bolt/pal.rb
```diff
@@ -615,5 +615,50 @@ def run_plan(plan_name, params, executor = nil, inventory = nil, pdb_client = ni
...
+      end
+
+      overrides = {
+        bolt_inventory: inventory,
+        bolt_project:   @project
+      }
+
+      # Do a lookup with a catalog compiler, which uses the 'hierarchy' key in
+      # Hiera config.
+      results = targets.map do |target|
```
lucywyman, line 633: This seems fine since the use case we're solving for is looking up a value on a limited set of targets for debugging, but we may want to parallelize this if people are using it to lookup lots of values.

documentation/hiera.md
beechtom: Meant for the fact to be `os.name`, so updated that instead.

lib/bolt/cli.rb
beechtom: Nope. Removed.

lib/bolt/pal.rb
beechtom: I think we can save that for future work when it's requested by a user, especially since I think it's unlikely users will be running this command frequently with large lists of targets that would impact performance.

lucywyman (approved):
Approved once CI is green!

Issue #2764, opened 2021-04-02
Enable more intuitive parallelism for varied operations


adreyer:
### Varied Actions
The existing parallelize function iterates over a collection executing the same block of code. This is well suited to iterating over an array of targets and doing the same thing on each one but can be non-intuitive to use in cases where the plan needs to do multiple different things at the same time, for example when provisioning a multi-node application you may want to take some actions on one target to set up the db while taking different actions on other targets to set up the application server. This can be accomplished with if/else or case statements but is non-intuitive. 
```
parralellize([db, app]) |$action| {
  case $action {
     'db': { run_plan(install_db, $db_targets) }
     'app': {run_plan(install_app, $app_targets) }
   }
}
```

### Fire and Forget
Sometimes nothing else in a plan depends on an action completing and it's useful to start a slow action then complete the rest of the plan without worrying about its result. This is challenging with the current `parallelize` function as the entire "rest of the plan" needs to be inside a conditional inside a `parallelize` block. If the action is happening in a sub-plan the parent plan cannot continue execution until the long-running action is complete.

## Describe the Solution You Would Like
A `background` function that accepts a block of code to execute in the background returns a `future` object that can be used to wait on or access the result of the block. 
A `wait` function that accepts a `future` or array of futures and returns an array of results. 

- Errors in a backgrounded block should not impact execution.
- Plans should not block returning if a backgrounded block is still running. If a future is unhandled we need to surface errors at a higher logging level (info maybe?).
- New entry point in Pal like 'start_plan' that starts the first thread, and then returns the first fiber result (the Main Plan) and keeps background tasks running.
- Pal should block after the top-level plan has returned until all backgrounded blocks have completed.
- Pal should log any errors from unresolved backgrounded blocks before exiting.
- The background function should accept a name.
- The wait function should return a parallel execution error if any futures in it failed 

The following statements should be functionally equivalent

```
parallelize([1,2,3]) |$val| { .... }
map([1,2,3]) |$val| { background($val) || { .... } }.wait()
```

## Questions
What should the `wait` function be called? `wait` with an alias for `value`? Not for now, just have `wait` that accepts a timeout and can add `value` once we understand how people are using this more.
What methods should the future object have? Should we have a 'state' function? Yes, since we can't have question marks in function names. `running`, `done`, `error` are possible states.

## Solution
[...]

PR #2810, opened 2021-04-30
(GH-2764) Add 'background' and 'wait' plan functions

lucywyman:
This fundamentally changes how plans are run in Bolt in order to support
two new plan functions, `background()` and `wait()`. Previously, users
who wanted to parallelize a Bolt plan had to use the `parallelize()`
plan function, which was built for fan-out parallelization. With
`parallelize()`, fire-and-forget parallelization was difficult because
users would have to include "the rest of the plan" in the body of the
`parallelize()` block, which was hard to read and understand.

To support fire-and-forget parallelism this change introduces
`background()` and `wait()` plan functions. `background()` runs the
provided block in parallel with the rest of the plan, and does not block
unless `wait()` is called. `wait()` blocks on the results of a
backgrounded block and returns them. `background()` creates a new
Bolt::Future object, which `wait()` accepts as an argument.

With `parallelize()` the `wait()` functionality was builtin to the
function, so the main plan itself could be executed serially with only
the parallelized blocks running in parallel (along with the main plan).
However with the `background()` function the main plan needs to run in
parallel with backgrounded blocks.

Bolt now starts the main plan in a Fiber and continuously resumes that
Fiber, even when there are no other blocks running (which could be a
future improvement). This shouldn't effect serial plan execution, and 
means that Bolt doesn't need to have a serial and parallel execution
mode - it just always executes in "parallel" even if there's only one 
"thread".

!feature

* **Add 'background()' and 'wait()' plan functions** ([#2764](https://github.com/puppetlabs/bolt/issues/2764))
  This introduces two new plan functions, `background()` and `wait()`,
  to support fire-and-forget parallelism in Bolt. Read more at
  https://pup.pt/bolt-parallelism.


lucywyman:
### Background function
- [x] Immediately returns a Future object
- [ ] Includes variables from calling scope
  - [ ] Once the scope has 'finished' (for example, background is last call of a plan)
  - [ ] In inner scopes (background inside background)
  - [ ] In subplan scopes (background in a subplan does not have access to 
- [ ] Returns from a `return` statement
- [ ] Does not fail the plan if an error occurs in the backgrounded block
- [ ] Errors if multiple fibers have the same name
- [ ] Plan waits for all futures to finish, even if some error (don't immediately fail the plan)

### Wait function
- [ ] Blocks until a future has finished
- [ ] Blocks until an array of futures has finished
- [ ] Returns results in the same order as the array passed in
- [ ] Returns only the results from the waited-on futures if they are in the 'middle' of the future queue
- [ ] Waits for a Future with an inner `background` call to finish
- [ ] Waits on all futures even if one errors
- [ ] Returns fibers that timedout as errors

### Parallelize
- [ ] Doesn't error when passed an array with duplicate objects
- [ ] Assigns a unique name to each fiber

### Future objects
- [x] Accurately retrieve `.state` as
  - [x] running
  - [x] done
  - [x] error

### Timing is everything
- [ ] A subplan with a background returns even if a backgrounded task is still running
- [ ] Logs errors even if the background task outlives the main plan
- [ ] The main plan does not return until all backgrounded tasks have finished
- [ ] The main plan returns immediately if there are no backgrounded tasks running

### Logging
- [ ] Logs trace messages while round-robining


- [ ] Errors in a backgrounded block should not impact execution.
- [ ] Subplans should not block returning if a backgrounded block is still running.
- [ ] If a future is unhandled we need to surface errors at a higher logging level (info maybe?).
    New entry point in Pal like 'start_plan' that starts the first thread, and then returns the first fiber result (the Main Plan) and keeps background tasks running.
- [ ] Pal should block after the top-level plan has returned until all backgrounded blocks have completed.
- [ ] Pal should log any errors from unresolved backgrounded blocks before exiting.
- [ ] The wait function should return a parallel execution error if any futures in it failed
- [ ] Need to have trace level messages in the executor while we're round-robinning for debugging, which would also make round-robinning with one thread annoying.
- [ ] We should ensure that errors in wait don't exit the plan, and are returned from the function
- [ ] As part of this, have parallelize return an error instead of raising it if the error is returned from the block, instead of thrown.
- [ ] If you call wait and it has errors, return a ParallelFailure object and raise unless in a catch_errors.

lib/bolt/executor.rb
```diff
@@ -373,83 +382,156 @@ def run_plan(scope, plan, params)
...
+
+    # Check on all the Fibers and see if any can be resumed.
+    def round_robin(rrfibers = plan_fibers.all)
+      r = nil
+
+      rrfibers.each do |fiber|
+        # If the Fiber is still running and can be resumed, then resume it and
+        # save the output to check if the task has finished or not.
+        @logger.trace("Checking future '#{fiber.name}'") if in_parallel
+        if fiber.alive?
```
adreyer, line 463: In general fibers should only become not alive during a resume call. I think this if-else confuses things and forces us into an "extra" loop when things finish. It seems like we should expect fibers to be alive, call `resume`, and then check to see if they completed and do the cleanup each loop rather than doing one then the other the next loop.

lib/bolt/executor.rb
```diff
@@ -373,83 +382,156 @@ def run_plan(scope, plan, params)
...
-      @in_parallel = true
-      publish_event(type: :stop_spin)
+    def plan_complete?
+      plan_fibers.fibers.empty? && main_plan.nil?
+    end
 
-      until skein.empty?
-        @thread_completed = false
-        r = nil
+    def wait(fibers, timeout, catch_errors: false, **_kwargs)
```
adreyer, line 500: I think we should already be round_robining so we don't need to call round_robin in this. We could just loop over all the provided fibers checking if they're alive but this will be much easier if each  `Bolt::Future` contained a concurrent ruby future we could check once without polling.

lib/bolt/executor.rb
```diff
@@ -373,83 +382,156 @@ def run_plan(scope, plan, params)
...
+          break
         end
+        round_robin(plan_fibers.fibers)
+      end
 
-        next unless r == 'unfinished'
-        sleep(0.1) until @thread_completed || skein.empty?
+      # Collect and return results from the specific fibers waited on in order.
+      result = names.each_with_object([]) do |name, acc|
+        acc << if (f = fiber_results.fibers[name])
```
adreyer, line 519: This is an example where I don't think we need the `fiber_results` why not just ask the completed fibers what their `value` is?

lib/bolt/pal.rb
```diff
@@ -607,11 +607,33 @@ def run_task(task_name, targets, params, executor, inventory, description = nil)
...
+        # other Fibers created during the plan run in the round_robin, with the
+        # main plan always taking precedence in being resumed.
+        fiber = Fiber.new do
+          r = compiler.call_function('run_plan', plan_name, params.merge('_bolt_api_call' => true))
+          Bolt::PlanResult.from_pcore(r, 'success')
+        rescue Bolt::Error => e
+          Bolt::PlanResult.new(e, 'failure')
+        end
+        # Give the main plan Fiber a name
+        name = "Main plan #{plan_name}"
```
adreyer, line 625: I think these messages are going to be more legible if we treat this more as a name than a description. The user probably knows what plan they called let's just call this `main`.

lib/bolt/yarn.rb
```diff
@@ -3,12 +3,13 @@
 require 'fiber'
 
 module Bolt
-  class Yarn
-    attr_reader :fiber, :value, :index
+  class Future
```
adreyer, line 6: Don't forget to move this file Also I think `PlanFuture` may be a better name.

bolt-modules/boltlib/lib/puppet/functions/parallelize.rb
```diff
@@ -35,21 +35,15 @@ def parallelize(scope, data, &block)
...
+    # Create a map of duplicates { "key" => 0 } to keep track of incrementing
+    # the number for the Fiber's name
+    duplicates = data.select { |e| data.count(e) > 1 }.to_a.uniq.product([0]).to_h
 
-    # TODO: Inner catch errors block?
-    if failed_indices.any?
-      raise Bolt::ParallelFailure.new(result, failed_indices)
+    futures = data.map do |object|
+      number = duplicates&.key?(object) ? duplicates[object] += 1 : ''
+      executor.create_fiber(scope, block, object, "Parallelize block for #{object}#{number}")
```
adreyer, line 44: I'm concerned this is a bad name but I don't know what better. For example if iterating over a series of hashes every log message would have the full hash..

adreyer (changes requested):
Overall this looks pretty close but I think it can be a lot cleaner if we move some code around, limit state, and take advantage of some concurrent ruby features.

The requirements look good other than what we discussed about changing the default behavior to fail by default. We should definitely bring that up to known `parallelize` users and in the channel.

lib/bolt/future.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+require 'fiber'
+
+module Bolt
+  class Future
```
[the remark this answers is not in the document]
lucywyman: Hm, so it seems like `round_robin` isn't running in a separate thread, so this just blocks forever. I suspect I'm doing something wrong (or more likely *not* doing something I should be) but for now I think doing this more manually will get this working and we can refactor

lib/bolt/fiber_executor.rb
```diff
@@ -0,0 +1,163 @@
...
+        # If the Fiber is still running and can be resumed, then resume it
+        @logger.trace("Checking future '#{future.name}'")
+        if future.alive?
+          @logger.trace("Resuming future '#{future.name}'")
+          future.resume
+        end
+
+        # Once we've restarted the Fiber, check to see if it's finished again
+        # and cleanup if it has.
+        next if future.alive?
```
beechtom, line 106: Still trying to dig into what the exact issue is, but there's some sort of race condition or something going on here. The following plan never finishes:

```puppet
plan bolt(
  TargetSpec $targets = 'localhost'
) {
  $futures = background() || {
    run_command('whoami', $targets)
  }

  $results = wait($futures)

  return $results
}
```

Specifically, everything runs as it should up until the "Main plan Bolt" future is all that's left. Putting a short sleep between the two `future.alive?` checks fixes this, though that's obviously a naive solution. 😬

lib/bolt/fiber_executor.rb
```diff
@@ -0,0 +1,168 @@
...
+          Bolt::Logger.warn('errored-futures', "Error in future '#{future.name}': #{future.value}")
+        end
+
+        # Remove the PlanFuture from the FiberExecutor.
+        plan_futures.delete(future)
+      end
+
+      # If the Fiber immediately returned wait for long enough that something
+      # can execute before checking again. This mitigates CPU thrashing.
+      return unless plan_futures.all? { |f| f.value.is_a?(Symbol) && f.value == :returned_immediately }
```
beechtom, line 121:
```suggestion
      return unless plan_futures.all? { |f| f.value == :returned_immediately }
```

I'm not sure that the class check is needed, since the equality operator should do the same?

lib/bolt/fiber_executor.rb
```diff
@@ -0,0 +1,168 @@
...
+    # Block until the provided PlanFuture objects have finished, or the timeout is reached.
+    #
+    def wait(futures, timeout, catch_errors: false, **_kwargs)
+      if timeout
+        start = Time.now
+        Fiber.yield(:unfinished) until (Time.now - start > timeout) || futures.map(&:alive?).none?
+        # Raise an error for any futures that are still alive
+        futures.each do |f|
+          if f.alive?
+            f.raise(Bolt::FutureTimeoutError.new(f.name, timeout))
```
beechtom, line 143: Fibers aren't returning this exception as their values and are instead returning symbols (specifically `:returned_immediately`). This causes the `wait` function to error, as the return type is not a valid `PlanResult`, and is instead a `RunTime` object.

Example plan:

```puppet
plan bolt(
  TargetSpec $targets = 'localhost'
) {
  $futures = background() || {
    run_command('sleep 5', $targets)
  }

  return wait($futures, 2)
}
```

And the error:

```json
{
  "kind": "bolt/pal-error",
  "msg": "value returned from function 'wait' has wrong type, index 0 expects a Boltlib::PlanResult = Optional[Variant[Boolean, Numeric, String, Object[{name => 'Error', type_parameters => {'kind' => Variant[String, Regexp, Type[Enum], Type[Pattern], Type[NotUndef], Type[Undef]], 'issue_code' => Variant[String, Regexp, Type[Enum], Type[Pattern], Type[NotUndef], Type[Undef]]}, attributes => {'msg' => String[1], 'kind' => {type => Optional[String[1]], value => undef}, 'details' => {type => Optional[Hash[String[1], Data]], value => undef}, 'issue_code' => {type => Optional[String[1]], value => undef}}, functions => {'message' => Callable[[0, 0], String[1]]}}], Object[{name => 'Result', attributes => {'value' => Hash[String[1], Data], 'target' => Target}, functions => {'error' => Callable[[0, 0], Optional[Error]], 'message' => Callable[[0, 0], Optional[String]], 'sensitive' => Callable[[0, 0], Optional[Sensitive[Data]]], 'action' => Callable[[0, 0], String], 'status' => Callable[[0, 0], String], 'to_data' => Callable[[0, 0], Hash], 'ok' => Callable[[0, 0], Boolean], '[]' => Callable[[String[1]], Variant[Data, Sensitive[Data]]]}}], Object[{name => 'ApplyResult', attributes => {'report' => Hash[String[1], Data], 'target' => Target}, functions => {'error' => Callable[[0, 0], Optional[Error]], 'ok' => Callable[[0, 0], Boolean], 'message' => Callable[[0, 0], Optional[String]], 'action' => Callable[[0, 0], String], 'to_data' => Callable[[0, 0], Hash], 'value' => Callable[[0, 0], Hash]}}], Object[{name => 'ResultSet', attributes => {'results' => Array[Variant[Result, ApplyResult]]}, functions => {'count' => Callable[[0, 0], Integer], 'empty' => Callable[[0, 0], Boolean], 'error_set' => Callable[[0, 0], ResultSet], 'filter_set' => Callable[[0, 0, Callable], ResultSet], 'find' => Callable[[String[1]], Optional[Variant[Result, ApplyResult]]], 'first' => Callable[[0, 0], Optional[Variant[Result, ApplyResult]]], 'names' => Callable[[0, 0], Array[String[1]]], 'ok' => Callable[[0, 0], Boolean], 'ok_set' => Callable[[0, 0], ResultSet], 'targets' => Callable[[0, 0], Array[Target]], 'to_data' => Callable[[0, 0], Array[Hash]], '[]' => Variant[Callable[[Integer], Optional[Variant[Result, ApplyResult, Array[Variant[Result, ApplyResult]]]]], Callable[[Integer, Integer], Optional[Variant[Result, ApplyResult, Array[Variant[Result, ApplyResult]]]]]]}}], Object[{name => 'Target', attributes => {'uri' => {type => Optional[String[1]], kind => given_or_derived}, 'name' => {type => Optional[String[1]], kind => given_or_derived}, 'safe_name' => {type => Optional[String[1]], kind => given_or_derived}, 'target_alias' => {type => Optional[Variant[String[1], Array[String[1]]]], kind => given_or_derived}, 'config' => {type => Optional[Hash[String[1], Data]], kind => given_or_derived}, 'vars' => {type => Optional[Hash[String[1], Data]], kind => given_or_derived}, 'facts' => {type => Optional[Hash[String[1], Data]], kind => given_or_derived}, 'features' => {type => Optional[Array[String[1]]], kind => given_or_derived}, 'plugin_hooks' => {type => Optional[Hash[String[1], Data]], kind => given_or_derived}, 'resources' => {type => Optional[Hash[String[1], ResourceInstance]], kind => given_or_derived}}, functions => {'host' => Callable[[0, 0], Optional[String]], 'password' => Callable[[0, 0], Optional[String[1]]], 'port' => Callable[[0, 0], Optional[Integer]], 'protocol' => Callable[[0, 0], Optional[String[1]]], 'transport' => Callable[[0, 0], String[1]], 'transport_config' => Callable[[0, 0], Hash[String[1], Data]], 'user' => Callable[[0, 0], Optional[String[1]]]}}], Object[{name => 'ResourceInstance', attributes => {'target' => Target, 'type' => Variant[String[1], Type[Resource]], 'title' => String[1], 'state' => {type => Optional[Hash[String[1], Data]], value => undef}, 'desired_state' => {type => Optional[Hash[String[1], Data]], value => undef}, 'events' => {type => Optional[Array[Hash[String[1], Data]]], value => undef}}, functions => {'add_event' => Callable[[Hash[String[1], Data]], Array[Hash[String[1], Data]]], 'set_state' => Callable[[Hash[String[1], Data]], Hash[String[1], Data]], 'overwrite_state' => Callable[[Hash[String[1], Data]], Hash[String[1], Data]], 'set_desired_state' => Callable[[Hash[String[1], Data]], Hash[String[1], Data]], 'overwrite_desired_state' => Callable[[Hash[String[1], Data]], Hash[String[1], Data]], 'reference' => Callable[[0, 0], String], '[]' => Callable[[String[1]], Data]}}], Object[{name => 'ContainerResult', attributes => {'value' => Hash[String[1], Data]}, functions => {'[]' => Callable[[String[1]], Data], 'error' => Callable[[0, 0], Optional[Error]], 'ok' => Callable[[0, 0], Boolean], 'status' => Callable[[0, 0], String], 'stdout' => Callable[[0, 0], String], 'stderr' => Callable[[0, 0], String], 'to_data' => Callable[[0, 0], Hash]}}], Object[{name => 'Future', functions => {'state' => Callable[[0, 0], Enum['done', 'error', 'running']]}}], Array[Boltlib::PlanResult], Hash[String, Boltlib::PlanResult]]] value, got Runtime",
  "details": {
    "file": "/Users/tom/.puppetlabs/bolt/plans/init.pp",
    "line": 8,
    "column": 10
  }
}
```

lib/bolt/fiber_executor.rb
```diff
@@ -0,0 +1,168 @@
...
+      end
+
+      # PlanFutures are assigned an ID, which is just a global incrementing
+      # integer. The main plan should always have ID 0.
+      @id += 1
+      Bolt::PlanFuture.new(future, @id, name)
+    end
+
+    # Register a PlanFuture with the FiberExecutor to be executed while round_robining.
+    def add_future(future)
```
adreyer, line 79: separating `create` and `add` in the API doesn't make much sense. Is it only there for top level that doesn't need to copy the scope? Is there a clean way to make a single create/add method to expose as API?

lib/bolt/executor.rb
```diff
@@ -373,83 +374,54 @@ def run_plan(scope, plan, params)
...
+    def plan_futures
+      @fiber_executor.plan_futures
     end
 
-    def round_robin(skein)
-      subscribe(self, [:node_result])
-      results = Array.new(skein.length)
-      @in_parallel = true
-      publish_event(type: :stop_spin)
+    def run_in_thread(cmd, params)
```
adreyer, line 409: I think this would be easier to understand/more idiomatic if it accepted a block rather than a method name and partial params. We already handling the stack in the boltlib functions for the non-parallel case so abstracting it here makes it harder to understand. 
A block accepting function would also be more general purpose and could be implemented on the `fiber_executor` perhaps.

spec/fixtures/parallel/error/tasks/fail.ps1
```diff
@@ -0,0 +1,2 @@
+Write-Output "failing"
+exit 1
```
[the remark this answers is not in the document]
lucywyman: Why?

lib/bolt/fiber_executor.rb
```diff
@@ -0,0 +1,168 @@
...
+
+      results = futures.map(&:value)
+
+      failed_indices = results.each_index.select do |i|
+        results[i].is_a?(Bolt::Error)
+      end
+
+      if failed_indices.any?
+        if !catch_errors
+        raise Bolt::ParallelFailure.new(results, failed_indices)
```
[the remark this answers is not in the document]
lucywyman: Huh surprised rubocop didn't catch this ❓

lib/bolt/fiber_executor.rb
```diff
@@ -0,0 +1,168 @@
...
+    # Creates a new Puppet scope from the current Plan scope so that variables
+    # can be used inside the block and won't interact with the outer scope.
+    # Then creates a new Fiber to execute the block, wraps the Fiber in a
+    # Bolt::PlanFuture, and returns the Bolt::PlanFuture. Because the main plan
+    # doesn't need the scoping and is called by the plan compiler in PAL, this
+    # is separate from `add_future` which adds the future to the list of
+    # futures to round_robin on. This is so that the main plan can create it's
+    # own PlanFuture and add it, while all other PlanFutures are created here.
+    #
+    def create_future(scope, block, name: nil, object: nil)
```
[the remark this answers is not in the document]
lucywyman: I talked with Alex about it, since each Future has a unique ID and users might not know what futures have been created we decided that Futures sharing a name was ok (just not sharing an ID). I'm a little on the fence about it myself, but does that seem reasonable to you?

lib/bolt/fiber_executor.rb
lucywyman: It's also because the plan and the block are called pretty different - the main plan is called through the compiler directly, whereas the code block is a `Puppet::Pops::Evaluator::PuppetProc` that has to be called with the new scope. I think I can have create_future accept a block that goes in the fiber perhaps...

spec/fixtures/parallel/error/tasks/fail.ps1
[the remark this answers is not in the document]
beechtom: There's no newline at the end of the script. 
image

lib/bolt/fiber_executor.rb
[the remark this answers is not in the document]
beechtom: Ya, fine with me. We should remove it from the acceptance criteria in that case.

lib/bolt/fiber_executor.rb
lucywyman: Ok, I think what I've got not makes a lot more sense and provides a cleaner API

lib/bolt/executor.rb
lucywyman: 100%, not sure why I didn't think of that.

lib/bolt/fiber_executor.rb
lucywyman: Oh my gosh, `Fiber.raise` wasn't added until Ruby 2.7 https://rubyreferences.github.io/rubychanges/2.7.html#fiberraise ughghghghghhg

lucywyman: Ah, I missed setting the `value` of the PlanFuture when raising errors. Nice!

lib/bolt/pal.rb
```diff
@@ -607,7 +607,36 @@ def run_task(task_name, targets, params, executor, inventory, description = nil)
...
+        executor.round_robin until executor.plan_complete?
+
+        # Return the result from the main plan
+        future.value
+      end
+    rescue Bolt::Error => e
+      Bolt::PlanResult.new(e, 'failure')
+    end
+
+    def boltspec_run_plan(plan_name, params, executor = nil, inventory = nil, pdb_client = nil, applicator = nil)
```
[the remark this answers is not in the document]
lucywyman: Hm, initially this was so that the plan itself would also execute serially since it doesn't call `create_future` which in BoltSpec is what executes each step, but now that the main plan does call `create_future` this shouldn't be needed anymore.

spec/fixtures/parallel/error/tasks/fail.ps1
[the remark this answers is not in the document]
lucywyman: Does there need to be?

lib/bolt/fiber_executor.rb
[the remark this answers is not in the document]
lucywyman: Good call, removed

documentation/experimental_features.md
```diff
@@ -173,7 +174,48 @@ plan myplan(TargetSpec $targets) {
...
+of plan steps to execute in the background while other parts of the plan execute in parallel:
+* The *background* plan function will begin executing a block of code in parellel with the main plan
+  and other backgrounded code blocks. This is great for use cases where you want to start a process
+  and don't care about the results, or don't need the results until much later in the plan. This
+  function returns a [Future](bolt_types_reference.md#Future) object so that the result can be
+  accessed later in the plan.
+* The *wait* function is a sister to `background`, it accepts a Future or array of Futures and
+  blocks until they are finished, then returns the results.
+* The *parallelize* function accepts an array and a block of plan steps, and then creates a Future
+  for each
```
hestonhoffman, line 189:
```suggestion
* The `wait` function is a sister to `background`, it accepts a Future or array of Futures and
  blocks until they are finished, then returns the results.
* The `parallelize` function accepts an array and a block of plan steps, and then creates a Future
  for each block in the array.
```

hestonhoffman (commented):
Quick initial edit. Note on style for this one - for obvious reasons, I think it's going to be okay to use a bit of future tense and passive voice on this one (although we should still try to avoid the passive where possible).

lib/bolt/fiber_executor.rb
lucywyman: Hm, it's not hanging for me. Is this still an issue after some refactoring? Are you running on multiple targets or just localhost?

beechtom: Still an issue after refactoring. Runs fine when the only target is `localhost`, but hangs with anything else (single target or multiple).

beechtom: Adding any `run_*` function between `background` and `wait` fixes the hanging issue.

lib/bolt/plan_future.rb
```diff
@@ -0,0 +1,48 @@
...
+    def to_s
+      "Future '#{name}'"
+    end
+
+    def alive?
+      fiber.alive?
+    end
+
+    def raise(exception)
+      fiber.raise(exception)
```
beechtom, line 30: I can't remember, but did we need to maintain compatibility with Ruby 2.5? `Fiber.raise` is private before 2.7. If not, please ignore. :)

beechtom (commented):
Still the issue with it hanging, and I'm still doing some testing, but overall this is looking really good :D

lib/bolt/plan_future.rb
lucywyman: We do in general need to maintain compatibility with Ruby 2.5, but the whole feature won't be supported in other projects for the foreseeable future - I'll leave a note this will only work in Ruby 2.7, but I don't expect it to be an issue for a while. There's also no risk of this getting called by running the main plan because this is only ever used to raise a Timeout error for `wait`, which is used by the `wait()` and `parallelize()` plan functions.

bolt-modules/boltlib/lib/puppet/functions/wait.rb
```diff
@@ -0,0 +1,43 @@
...
+  #   run_task("deploy", $targets)
+  #   # Wait for the file upload to finish
+  #   $results = wait($futures)
+  dispatch :wait do
+    param 'Variant[Future, Array[Future]]', :futures
+    # optional_param does not set Integer to optional
+    optional_param 'Optional[Numeric]', :timeout
+    optional_param 'Hash[String[1], Any]', :options
+    return_type 'Array[Boltlib::PlanResult]'
+  end
```
beechtom, line 26: I think this needs separate dispatches since both `timeout` and `options` are optional parameters. The following function invocation:

```
wait($future, '_catch_errors' => true)
```

Results in the following error:

```
"'wait' parameter 'timeout' expects a value of type Undef or Numeric, got Struct"
```

bolt-modules/boltlib/lib/puppet/functions/wait.rb
```diff
@@ -0,0 +1,43 @@
...
+  #     upload_file("./very_large_file", "/opt/jfrog/artifactory/var/etc/artifactory", $targets)
+  #   }
+  #   # Run an unrelated task
+  #   run_task("deploy", $targets)
+  #   # Wait for the file upload to finish
+  #   $results = wait($futures)
+  dispatch :wait do
+    param 'Variant[Future, Array[Future]]', :futures
+    # optional_param does not set Integer to optional
+    optional_param 'Optional[Numeric]', :timeout
```
beechtom, line 23: Should this be restricted to positive values?

bolt-modules/boltlib/lib/puppet/functions/wait.rb
```diff
@@ -0,0 +1,43 @@
...
+  end
+
+  def wait(futures, timeout = nil, options = {})
+    unless Puppet[:tasks]
+      raise Puppet::ParseErrorWithIssue
+        .from_issue_and_stack(Bolt::PAL::Issues::PLAN_OPERATION_NOT_SUPPORTED_WHEN_COMPILING, action: 'wait')
+    end
+
+    options = options.transform_keys { |k| k.sub(/^_/, '').to_sym }
+    options[:timeout] = timeout if timeout
```
beechtom, line 35: This lets you pass `timeout` to the `options` hash. We should either filter out invalid options or pass `timeout` as a parameter to `executor.wait`.

lib/bolt/executor.rb
```diff
@@ -77,6 +76,8 @@ def initialize(concurrency = 1,
 
       @concurrency = concurrency
       @warn_concurrency = modified_concurrency
+      # TODO: Bikeshed name
```
beechtom, line 79:
```suggestion
```

Delete comment before merging.

lib/bolt/fiber_executor.rb
beechtom: This issue has been fixed 👍

lib/bolt/fiber_executor.rb
```diff
@@ -0,0 +1,141 @@
...
+
+        # Once we've restarted the Fiber, check to see if it's finished again
+        # and cleanup if it has.
+        next if future.alive?
+        @logger.trace("Cleaning up future '#{future.name}'")
+
+        # If the future errored and the main plan has already exited, log the
+        # error at warn level.
+        unless plan_futures.map(&:id).include?(0) || future.state == "done"
+          Bolt::Logger.warn('errored-futures', "Error in future '#{future.name}': #{future.value}")
```
beechtom, line 82:
```suggestion
          Bolt::Logger.warn('errored_futures', "Error in future '#{future.name}': #{future.value}")
```

Warning IDs use underscores.

beechtom, line 82: This will log `RunFailure` error messages which state the plan was aborted. This is a bit misleading since errors in backgrounded blocks don't actually abort the plan. Here's some example output:

```
Starting: plan bolt
Starting: task facts on nix1, nix2, win
Starting: command 'tree' on nix1, nix2, win
Starting: command 'sleep 5' on nix1, nix2, win
Finished: command 'tree' with 2 failures in 1.69 sec
Error in future '2': Plan aborted: run_command 'tree' failed on 2 targets [ID: errored-futures]
skip_cleanup [ID: Skipping cleanup of tmpdir 'C:\Users\Administrator\AppData\Local\Temp\ld1ioqpz.gx1']
Finished: task facts with 0 failures in 3.64 sec
Finished: plan bolt in 4.07 sec
Finished on nix1:
  ...
Finished on nix2:
  ...
Finished on win:
  ...
Successful on 3 targets: nix1,nix2,win
Ran on 3 targets
```

bolt-modules/boltlib/lib/puppet/functions/wait.rb
lucywyman: Oh, nice catch! Definitely needs two dispatches.

bolt-modules/boltlib/lib/puppet/functions/wait.rb
lucywyman: Hmm, I mostly like having it as a keyword arg to wait so that the function is more readable and is more clear that it's optional. I'll opt for filtering and see how clunky it feels.

lib/bolt/fiber_executor.rb
lucywyman: Ugh, I think this is actually going to be trickier to work around than I thought....It's hard to know when we raise the RunFailure in the plan functions if the function is being run in a background block, and other teams are [raising RunFailure errors](https://github.com/puppetlabs/pe-bolt-modules/blob/52a1c9a16aa036458524b527eb4f00177614cf24/pe-bolt-modules/pe_boltlib/lib/puppet/functions/apply_prep.rb#L23) so modifying the function signature beyond adding an optional parameter isn't a good optoin.

lucywyman: I think this is an improvement we can make, but doing it right is going to be a bit complicated. I'd like to get this in first and do that as a followup to avoid scope-creep too much.

lucywyman: <5 hours later>

bolt-modules/boltlib/lib/puppet/functions/wait.rb
```diff
@@ -0,0 +1,75 @@
+# frozen_string_literal: true
+
+require 'bolt/logger'
+require 'bolt/target'
+
+# Wait for a Future or array of Futures to finish and return results.
+#
+# > **Note:** Not available in apply block
+Puppet::Functions.create_function(:wait, Puppet::Functions::InternalFunction) do
+  # Wait for a Future or array of Futures to finish and return results.
```
beechtom, line 10: We should have a brief summary of this specific dispatch instead of a description of the function, as this is used as the header for the function signature. Maybe something like "Wait for Futures to finish" ?

See [run_command](https://github.com/puppetlabs/bolt/blob/main/bolt-modules/boltlib/lib/puppet/functions/run_command.rb#L6-L11) as an example.

bolt-modules/boltlib/lib/puppet/functions/wait.rb
```diff
@@ -0,0 +1,75 @@
...
+  #   run_task("deploy", $targets)
+  #   # Wait for the file upload to finish
+  #   $results = wait($futures)
+  dispatch :wait do
+    param 'Variant[Future, Array[Future]]', :futures
+    optional_param 'Hash[String[1], Any]', :options
+    return_type 'Array[Boltlib::PlanResult]'
+  end
+
+  # Wait for a Future or array of Futures to finish and return results.
```
beechtom, line 28: Same as above, this should be a brief summary of the specific dispatch. Something like "Wait for futures to finish, timing out after a set time".

bolt-modules/boltlib/lib/puppet/functions/wait.rb
```diff
@@ -0,0 +1,75 @@
...
+  # @option options [Boolean] _catch_errors Whether to catch raised errors.
+  # @return A Result or Results from the Futures
+  # @example Upload a large file in the background, then wait until it's loaded
+  #   $futures = background() || {
+  #     upload_file("./very_large_file", "/opt/jfrog/artifactory/var/etc/artifactory", $targets)
+  #   }
+  #   # Run an unrelated task
+  #   run_task("deploy", $targets)
+  #   # Wait for the file upload to finish
+  #   $results = wait($futures)
```
beechtom, line 40: This example should include a timeout in the `wait` invocation and.

bolt-modules/boltlib/lib/puppet/functions/wait.rb
```diff
@@ -0,0 +1,75 @@
...
+        .from_issue_and_stack(Bolt::PAL::Issues::PLAN_OPERATION_NOT_SUPPORTED_WHEN_COMPILING, action: 'wait')
+    end
+
+    valid, invalid = options.partition { |k, _v| %w[_catch_errors].include?(k) }.map(&:to_h)
+    if invalid.any?
+      file, line = Puppet::Pops::PuppetStack.top_of_stack
+      msg = "The wait() function call in #{file}#L#{line} received invalid options "\
+        "#{invalid.keys}. Removing invalid options and continuing..."
+      Bolt::Logger.warn("wait_function_options", msg)
+    end
```
beechtom, line 64: I don't think we have this pattern anywhere else, at least in plan functions. I think we should either just silently ignore unknown options or use a more general warning ID like `plan_function_options`.

lucywyman: Yeah, we don't, but it seemed like it could be useful? I can remove it if we want to add this to more functions as it's own ticket (this PR is certainly big enough :P)

bolt-modules/boltlib/lib/puppet/functions/wait.rb
lucywyman: Oof I just totally forgot to do this, thanks!

bolt-modules/boltlib/lib/puppet/functions/wait.rb
beechtom: Ya, I don't mind having it at all (as long as the ID is more generic). Your call if it goes in this ticket or another one!

bolt-modules/boltlib/lib/puppet/functions/background.rb
```diff
@@ -0,0 +1,60 @@
...
+  # @return A Bolt Future object
+  # @example Start a long-running process
+  #   background() || {
+  #     run_task('superlong::task', $targets)
+  #   }
+  #   run_command("echo 'Continue immediately'", $targets)
+  dispatch :background do
+    scope_param
+    optional_param 'String[1]', :name
+    block_param 'Callable[0, 0]', :block
```
beechtom, line 20: `block` needs a `@param` tag.

bolt-modules/boltlib/lib/puppet/functions/wait.rb
```diff
@@ -0,0 +1,88 @@
...
+  #   $futures = background() || {
+  #     upload_file("./very_large_file", "/opt/jfrog/artifactory/var/etc/artifactory", $targets)
+  #   }
+  #   # Run an unrelated task
+  #   run_task("deploy", $targets)
+  #   # Wait for the file upload to finish
+  #   $results = wait($futures)
+  dispatch :wait do
+    param 'Variant[Future, Array[Future]]', :futures
+    optional_param 'Hash[String[1], Any]', :options
```
beechtom, line 25: Needs a `@param` tag.

bolt-modules/boltlib/lib/puppet/functions/wait.rb
```diff
@@ -0,0 +1,88 @@
...
+  #   }
+  #   # Run an unrelated task
+  #   run_task("deploy", $targets)
+  #   # Wait for the file upload to finish
+  #   $results = wait($futures, 30, '_catch_errors' => true)
+  dispatch :wait_with_timeout do
+    param 'Variant[Future, Array[Future]]', :futures
+    # PUP-11066
+    optional_param 'Optional[Variant[Integer[0], Float[0.0]]]', :timeout
+    optional_param 'Hash[String[1], Any]', :options
```
beechtom, line 57: Needs a `@param` tag.

bolt-modules/boltlib/lib/puppet/functions/wait.rb
```diff
@@ -0,0 +1,88 @@
...
+  #     upload_file("./very_large_file", "/opt/jfrog/artifactory/var/etc/artifactory", $targets)
+  #   }
+  #   # Run an unrelated task
+  #   run_task("deploy", $targets)
+  #   # Wait for the file upload to finish
+  #   $results = wait($futures, 30, '_catch_errors' => true)
+  dispatch :wait_with_timeout do
+    param 'Variant[Future, Array[Future]]', :futures
+    # PUP-11066
+    optional_param 'Optional[Variant[Integer[0], Float[0.0]]]', :timeout
```
beechtom, line 56:
```suggestion
    param 'Variant[Integer[0], Float[0.0]]', :timeout
```

This doesn't need to be optional since we have separate dispatches.

bolt-modules/boltlib/lib/puppet/functions/wait.rb
```diff
@@ -0,0 +1,88 @@
...
+  #   # Wait for the file upload to finish
+  #   $results = wait($futures)
+  dispatch :wait do
+    param 'Variant[Future, Array[Future]]', :futures
+    optional_param 'Hash[String[1], Any]', :options
+    return_type 'Array[Boltlib::PlanResult]'
+  end
+
+  # Wait for a Future or array of Futures to finish, timing out on any Futures
+  # don't finish within the provided timeout. Returns the results of the Futures.
```
beechtom, line 30: This feels a bit verbose for a section header. Here's how it's rendered in docs:

image

I would have the description at the top of the file be the verbose description and the dispatch descriptions be simple, one-line summaries (e.g. "Wait for Futures to finish with a timeout").

bolt-modules/boltlib/lib/puppet/functions/wait.rb
```diff
@@ -0,0 +1,88 @@
...
+  # @param futures A Bolt Future object or array of Bolt Futures to wait on.
+  # @param timeout How long to wait for Futures to finish before raising a Timeout error.
+  # @option options [Boolean] _catch_errors Whether to catch raised errors.
+  # @return A Result or Results from the Futures
+  # @example Upload a large file in the background, then wait up to 30 seconds
+  # until it's loaded
```
beechtom, line 36: Strings does not like multi-line descriptions for examples.

image

bolt-modules/boltlib/lib/puppet/functions/wait.rb
```diff
@@ -0,0 +1,88 @@
...
+  #   $futures = background() || {
+  #     upload_file("./very_large_file", "/opt/jfrog/artifactory/var/etc/artifactory", $targets)
+  #   }
+  #   # Run an unrelated task
+  #   run_task("deploy", $targets)
+  #   # Wait for the file upload to finish
+  #   $results = wait($futures, 30)
+  # @example Upload a large file in the background, then wait up to 30 seconds
+  # until it's loaded and catch any errors that occur
```
beechtom, line 45: Same here for the multi-line example description.

bolt-modules/boltlib/lib/puppet/functions/wait.rb
```diff
@@ -0,0 +1,88 @@
...
+    unless Puppet[:tasks]
+      raise Puppet::ParseErrorWithIssue
+        .from_issue_and_stack(Bolt::PAL::Issues::PLAN_OPERATION_NOT_SUPPORTED_WHEN_COMPILING, action: 'wait')
+    end
+
+    valid, invalid = options.partition { |k, _v| %w[_catch_errors].include?(k) }.map(&:to_h)
+    if invalid.any?
+      file, line = Puppet::Pops::PuppetStack.top_of_stack
+      msg = "The wait() function call in #{file}#L#{line} received invalid options "\
+        "#{invalid.keys}. Removing invalid options and continuing..."
```
beechtom, line 75: Maybe 'unknown' instead of 'invalid', since that's what we use elsewhere in Bolt? Don't have a strong preference.

bolt-modules/boltlib/lib/puppet/functions/wait.rb
lucywyman: Definitely - updated the ID, but leaving this here.

bolt-modules/boltlib/lib/puppet/functions/wait.rb
```diff
@@ -0,0 +1,87 @@
...
+
+require 'bolt/logger'
+require 'bolt/target'
+
+# Wait for a Future or array of Futures to finish and return results,
+# optionally with a timeout.
+#
+# > **Note:** Not available in apply block
+Puppet::Functions.create_function(:wait, Puppet::Functions::InternalFunction) do
+  # Wait for Future(s) to finish with timeout
```
beechtom, line 11:
```suggestion
  # Wait for Future(s) to finish
```

bolt-modules/boltlib/lib/puppet/functions/wait.rb
```diff
@@ -0,0 +1,87 @@
...
+    optional_param 'Hash[String[1], Any]', :options
+    return_type 'Array[Boltlib::PlanResult]'
+  end
+
+  # Wait for Future(s) to finish with timeout
+  # @param futures A Bolt Future object or array of Bolt Futures to wait on.
+  # @param timeout How long to wait for Futures to finish before raising a Timeout error.
+  # @param options A hash of additional options.
+  # @option options [Boolean] _catch_errors Whether to catch raised errors.
+  # @option options [Boolean] _catch_errors Whether to catch raised errors.
```
beechtom, line 35:
```suggestion
```

Duplicated from the line above.

documentation/experimental_features.md
```diff
@@ -246,11 +306,30 @@ $result = parallelize($ts) |$target| {
...
+out::message($result)
+```
+
 If any step of the block errors, Bolt stops executing the block for that target, but continues
 executing for all other targets from the input array. When the block finishes, if there is an error
 in the result array, the plan throws a `PlanFailure` and includes the entire result array in the
 `details` key of the failure. If the block is wrapped in a `catch_errors()` block, Bolt catches the
-`PlanFailure` and continues executing the plan.
+`PlanFailure` and continues executing the plan. For `wait()`, if `_catch_errors => true` is
+provided then any errors raised will be returned and the plan will continue.
```
hestonhoffman, line 332:
```suggestion
`PlanFailure` and continues to execute the plan. If you've provided `_catch_errors => true`
to `wait()`, Bolt returns any errors raised and the plan continues to execute.
```

documentation/experimental_features.md
```diff
@@ -173,7 +174,62 @@ plan myplan(TargetSpec $targets) {
...
+}
+
+# Is equivalent to
+['./file1', './file2'].map |$file| {
+  background() || {
+    file_upload($file,...)
+    ...
+  }
+}
+```
```
beechtom, line 206: Shouldn't the second example have a `wait` call to make these functionally equivalent?

```
['./file1', './file2'].map |$file| {
  background() || {
    file_upload($file,...)
    ...
  }
}.wait
```

lucywyman: Oh my gosh good eye! Definitely.

beechtom (approved):

documentation/experimental_features.md
```diff
@@ -154,15 +154,16 @@ plugin-cache:
 > **NOTE**: The same cache and cache configuration is used for the `resolve_reference()` plan
 function.
 
-## `Parallelize()` function
+## Parallelism in Bolt plans
 
-This feature was introduced in [Bolt
+The `parallelize()` plan function was introduced in [Bolt
 2.35.0](https://github.com/puppetlabs/bolt/tree/main/CHANGELOG.md#bolt-2350-2020-11-16).
+The `background()` and `wait()` plan functions were introduced in [Bolt 3.9.0](https://github.com/puppetlabs/bolt/tree/main/CHANGELOG.md#bolt-390-2021-05-24).
```
hestonhoffman, line 161:
```suggestion
The `background()` and `wait()` plan functions were introduced in [Bolt 3.9.0](https://github.com/puppetlabs/bolt/tree/main/CHANGELOG.md#bolt-390-2021-05-25).
```

documentation/experimental_features.md
```diff
@@ -173,7 +174,62 @@ plan myplan(TargetSpec $targets) {
...
+before it can run `whoami`.
+
+While useful, this form of parallelism is limited. Bolt plans have a few methods that allow a block
+of plan steps to execute in the background while other parts of the plan execute in parallel:
+* The `background` plan function begins executing a block of code in parellel with the main plan
+  and other backgrounded code blocks. This is great for use cases where you want to start a process
+  and don't care about the results, or don't need the results until much later in the plan. This
+  function returns a [Future](bolt_types_reference.md#Future) object so that the result can be
+  accessed later in the plan.
+* The `wait` function is a sister to `background`, it accepts a Future or array of Futures and
```
hestonhoffman, line 186:
```suggestion
* The `wait` function is a sister to `background`. It accepts a Future or array of Futures and
```

documentation/experimental_features.md
```diff
@@ -173,7 +174,62 @@ plan myplan(TargetSpec $targets) {
...
+```
+
+### `background()` and `wait()` plan functions
+
+The `background()` plan function creates a new Future object, and begins running the code block in
+the background in parallel with the main plan and other backgrounded code blocks. The function
+accepts an optional name for the Future to make log messages easier to follow.
+
+The code block has access to all existing variables when it's created, and has its own scope so any
+new variables are not accessible outside the code block.
```
hestonhoffman, line 215:
```suggestion
The code block has access to all existing variables when it's created and has its own scope, so any
new variables are not accessible outside the code block.
```

hestonhoffman (approved):
Couple small changes, but this is looking good!

Issue #2794, opened 2021-04-19
Add a note about encrypted private keys to transport configuration + Troubleshooting


lucywyman:
## Describe the Change You Would Like
Bolt cannot read encrypted private SSH keys, and if users aren't using an SSH agent this can result in a cryptic message `: Could not parse PKey: no start line`. We should add a note about encrypted private keys to the transport configuration page and anywhere else we mention connecting using private keys, as well as a section to the Troubleshooting page. Using `openssl rsa -in ` to decrypt the key works around the issue, but may be unusable in some environments for security reasons.

PR #2811, opened 2021-04-30
(GH-2794) Document encrypted keys with 'key-data' issue

lucywyman:
This adds a note to the Transport configuration reference page and the
Troubleshooting page that providing an encrypted private SSH key with
`key-data` will not work, and includes instructions on the
Troubleshooting page with how to decrypt the key.

!no-release-note

### DIFF
--- a/documentation/troubleshooting.md
+++ b/documentation/troubleshooting.md
@@ -179,12 +179,6 @@ SSL_connect returned=1 errno=0 state=error: certificate verify failed (unable to
 Set the `SSL_CERT_DIR` and `SSL_CERT_FILE` environment variables to use a valid
 certificate and certificate directory.
 
-## I still need help
-
-Visit the **#bolt** channel in the [Puppet Community
-Slack](https://slack.puppet.com) to find a whole community of people waiting
-to help!
-
 ## PowerShell does not recognize Bolt cmdlets
 
 PowerShell 3.0 cannot automatically discover and load the Bolt module. If you're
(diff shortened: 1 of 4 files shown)


documentation/troubleshooting.md
```diff
@@ -251,3 +251,11 @@ To change your script execution policy:
     documentation about [execution
     policies](http://go.microsoft.com/fwlink/?LinkID=135170) and [how to set
     them](https://msdn.microsoft.com/en-us/powershell/reference/5.1/microsoft.powershell.security/set-executionpolicy).
+
+## Getting error 'Could not parse PKey: no start line' using SSH private key
+
+This error indicates that the SSH private key provided is encrypted, which is unsupported by Bolt
+if the key is provided under `key-data`. You can manually decrypt the key by running `openssl rsa -in
+<key_file>` and providing your passphrase, if providing a decrypted key is feasible for your use
+case and security practices. You can also provide a path to the key instead of `key-data` and add it
```
nicklewis, line 260: Adding it to SSH agent is the part that makes this work. Providing a path is unnecessary if you're using the SSH agent and _doesn't_ work if you aren't.

lucywyman: Ah good point!

documentation/troubleshooting.md
```diff
@@ -251,3 +251,11 @@ To change your script execution policy:
...
     policies](http://go.microsoft.com/fwlink/?LinkID=135170) and [how to set
     them](https://msdn.microsoft.com/en-us/powershell/reference/5.1/microsoft.powershell.security/set-executionpolicy).
+
+## Getting error 'Could not parse PKey: no start line' using SSH private key
+
+This error indicates that the SSH private key provided is encrypted, which is unsupported by Bolt
+if the key is provided under `key-data`. You can manually decrypt the key by running `openssl rsa -in
+<key_file>` and providing your passphrase, if providing a decrypted key is feasible for your use
+case and security practices. You can also add the key to your SSH agent and *not* specify a
+`private-key` for Bolt to use, and Bolt will use the agent to authenticate your connection.
```
hestonhoffman, line 261:
```suggestion
## 'Could not parse PKey: no start line' error message when using SSH private key

Bolt does not support encrypted SSH private keys if the keys are provided using the
`key-data` field in your transport configuration. If providing a decrypted key is feasible
for your use case and security practices, you can manually decrypt the key by running
`openssl rsa -in <KEY FILE>` and providing your passphrase. Alternatively, you can
add the key to your SSH agent and *not* specify a `private-key` for Bolt to use. Bolt
will use the agent to authenticate your connection.
```

hestonhoffman: @lucywyman Could you move the "I still need help" section down? That should be the last section on the page.

beechtom (approved):

hestonhoffman (approved):

Issue #2601, opened 2021-01-29
Clarify that using  `_sensitive` in task return can only be used at the top level


lucywyman:
## Describe the Change You Would Like
We should [clarify here](https://puppet.com/docs/bolt/latest/writing_tasks.html#returning-sensitive-data) that the `_sensitive` key can only be used at the top level of the hash returned from a task, and that it's value must be a hash.

PR #2812, opened 2021-04-30
(GH-2601) Clarify that _sensitive can only be used at the top-level

lucywyman:
This clarifies in our documentation that when returning sensitive values
from a task, the `_sensitive` key must be at the top level to make it's
value sensitive. It provides an example using a nested `_sensitive` key
that would not be made sensitive.

!no-release-note


documentation/writing_tasks.md
```diff
@@ -439,9 +439,10 @@ error. Bolt merges the 'file' and 'line' keys with results unless 'file' is pres
 
 ### Returning sensitive data
 
-To return secrets from a task, use the `_sensitive` key in the output. Bolt
-will treat the result as sensitive and won't allow it to be printed to the
-console or log.
+To return secrets from a task, use the `_sensitive` key at the top level of the output. Bolt will
+treat the result as sensitive and won't allow it to be printed to the console or log. Note that this
+will not obfuscate data under `_sensitive` if `_sensitive` is not at the top level, for example `{
+user: _sensitive: { carmen: sandiego } }` will not be made sensitive.
```
hestonhoffman, line 445:
```suggestion
does not obfuscate data under `_sensitive` if `_sensitive` is not at the top level. 
For example, Bolt does not treat this `user` as sensitive: 
`{user: _sensitive: { carmen: sandiego } }`.
```
Could we also provide the "correct" example here?

lucywyman: The correct example is directly below, I can make that more clear!

documentation/writing_tasks.md
```diff
@@ -439,10 +439,12 @@ error. Bolt merges the 'file' and 'line' keys with results unless 'file' is pres
...
 
-To return secrets from a task, use the `_sensitive` key in the output. Bolt
-will treat the result as sensitive and won't allow it to be printed to the
-console or log.
+To return secrets from a task, use the `_sensitive` key at the top level of the output. Bolt treats
+the result as sensitive and does not allow it to be printed to the console or log. Note that Bolt
+does not obfuscate data under `_sensitive` if `_sensitive` is not at the top level.  For example,
+Bolt does not treat this `user` data as sensitive: `{user: _sensitive: { carmen: sandiego } }`.
 
+This task will generate a random password for user 'someone', making the password hash sensitive.
```
beechtom, line 447:
```suggestion
This task generates a random password for user 'someone', making the password hash sensitive.
```

documentation/writing_tasks.md
```diff
@@ -457,6 +459,17 @@ result = { user: user_name, _sensitive: { password: user_password } }
 puts result.to_json
 ```
 
+Running the task will print:
```
beechtom, line 462:
```suggestion
Running the task prints:
```

documentation/writing_tasks.md
lucywyman: 💀

beechtom: I use future tense _all the time_ too haha. It's the most common correction Heston makes to my docs changes. 😅

beechtom (approved):

PR #2814, opened 2021-05-03
(bug-fix) Don't stacktrace if welcome message file can't be written

lucywyman:
In order to determine whether to print the Bolt welcome message,
Bolt inspects a directory for a file, and if the file does not exist
prints the welcome message and lays down the file so that future runs
don't also print the welcome message. This allows the welcome message to
be printed on a mostly-per-user basis, and avoids environment variable
instability.

Previously the path for the file was based on `Bolt::Config.user_path`,
but when that fails to be created Bolt will stacktrace. Now, Bolt
catches `StandardError` and `Errno::ENOENT` error types and falls back
first to `Bolt::Config.system_path`, and then to returning `nil` and
avoiding printing the welcome message entirely. This ensures that trying
to print a welcome message doesn't result in Bolt being unusable due to
erroring out trying to create a file in an invalid location.

!bug

* **Don't stacktrace if welcome message file can't be written**

  Previously, Bolt would stacktrace if it failed to make the directory
  to store the welcome message file in, which relies on expanding `~`.
  Bolt now falls back to a system-level path, and then omits the welcome
  message entirely if the system-level path also fails to be created or
  written to.


lib/bolt/util.rb
```diff
@@ -78,11 +78,25 @@ def read_optional_yaml_hash(path, file_name)
...
+        rescue StandardError
+          nil
+        end
       end
 
       def first_run?
-        Bolt::Config.user_path && !File.exist?(first_runs_free)
+        # If the first_run file can be written, and hasn't already been written
+        !first_runs_free.nil? &&
+          File.writable?(first_runs_free) &&
```
nicklewis, line 98: `File.writable?` checks whether the file has the `write` permission, not whether it could be created. It should be impossible for a file to both not exist but also be writable.

lib/bolt/util.rb
```diff
@@ -78,11 +78,23 @@ def read_optional_yaml_hash(path, file_name)
...
       def first_runs_free
+        # If this fails, use the system path instead
+        FileUtils.mkdir_p(Bolt::Config.user_path)
         Bolt::Config.user_path + '.first_runs_free'
+      rescue StandardError
+        begin
+          # If using the system path fails, then don't bother with the welcome
+          # message
+          FileUtils.mkdir_p(Bolt::Config.system_path)
+          Bolt::Config.system_path + '.first_runs_free'
```
beechtom, line 89: Since it's fairly unlikely that a user will be running with permissions to write to this directory, I think it might make sense to not fall back to the system path if the user path fails.

lucywyman: I don't see why we wouldn't at least try, and give up if it's not writable? This is the same fallback we have for default projects, which is essentially what I'm trying to do here - write to the default project without actually loading config. Is there a risk you're concerned about?

Issue #2773, opened 2021-04-07
Document JSON output for Bolt commands


beechtom:
## Describe the Change You Would Like
Currently, users need to run commands with the JSON format to know how the output is structured. We should document this output (expected keys, types, etc.) so users can know what to expect without needing to run commands.

PR #2817, opened 2021-05-05
(GH-2773) Document JSON output

beechtom:
This adds a new reference page that documents the JSON output format for
Bolt commands.

!feature

* **JSON output documentation**
  ([#2773](https://github.com/puppetlabs/bolt/issues/2773))

  The format for JSON output for each of Bolt's commands is [now
  documented](https://puppet.com/docs/bolt/latest/json_output_reference.md).


documentation/json_output_reference.md
```diff
@@ -0,0 +1,763 @@
+# JSON output
+
+Bolt output can be returned as JSON by specifying the `--format json` or `-Format json`
```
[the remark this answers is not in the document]
lucywyman: Or something, mostly just so *nix folks don't try `-Format json` :P

documentation/json_output_reference.md
```diff
@@ -0,0 +1,763 @@
...
+command-line options. Bolt commands return JSON output with the following format:
```
lucywyman, line 4:
```suggestion
command-line options. Bolt commands return JSON output guaranteed to have the following structures. Note that some commands may have additional data depending on if the command failed, whether the command was run with certain command-line options (like `--detail`), etc.
```

documentation/json_output_reference.md
```diff
@@ -0,0 +1,763 @@
...
+{
+  "topics": [
+    "inventory",
+    "logging",
+    "module"
+  ]
+}
+```
+
+## `inventory show`
```
lucywyman, line 266: Hm, I think this is one of the trickier commands given the note about different forms of the command having different outputs, since `--detail` only has `targets` and `count`. Maybe including `adhoc` and `inventory` in a separate subheader that's explicitly without `--detail` is the simplest? I think the risk is that I suspect `inventory show --detail` is pretty useful, especially using JSON, and it'd be frustrating to see the keys documented but then have them not be in the output and not know why they weren't in the output.

Issue #2755, opened 2021-03-31
Noop for bolt


m0un10:
## Use Case

I want to test my plan safely.

As bolt has `noop` for tasks, I was surprised to learn that `bolt plan` doesn't have this feature.

## Describe the Solution You Would Like

Running `bolt plan run <planName> --noop` runs with noop on all the tasks it calls so it doesn't actually make changes to the target environments. If the tasks don't support `noop` then the plan should fail with a message indicating that the plan doesn't support `noop` due tasks X,Y,Z etc.

## Describe Alternatives You've Considered

Ansible Playbooks with `check` mode.

## Additional Context

This is a bit of a showstopper for using bolt at the moment

lucywyman:
## Solution
- This should only be supported as a flag, and cannot be set in the plan itself. No `--no-noop` flag, or `_noop` for the `run_plan` plan function.
- This should error if any functions other than `apply` or `run_task` are called, or if any tasks that are run don't support noop. Consider a noop executor that's swapped out for the regular executor when this flag is provided that errors if any functions besides run_task are called.
- We should ensure that the puppet_agent::install task supports noop so that `apply_prep` can be called safely.
- For both puppet and yaml plans, run until we hit a step that doesn't support noop and then error out.
- Investigate bug around `run_task_with` respecting executor noop setting.

op-ct:
BoltSpec seems like the right tool for this.  I'm skeptical that a `bolt plan run --noop` could be reliable enough to be useful:

* Plans can include logic that makes decisions based on the results of `run_task` and `apply`.   
* If those functions run as `--noop`, they won't return the same ResultSets—which could cause the plan logic to behave differently than it would have with a real environment.  
* If `bolt plan run --noop` is restricted to only two plan functions, requires all tasks to support noop, and may even try to do different things, it's hard to see it being very useful as a noop.

On the other hand: 

* BoltSpec can mock `run_task`, `apply`, and [many other plan functions](https://puppet.com/docs/bolt/latest/boltspec_reference.html#mocks) to return expected ResultSets before test-running a plan and verifying that it does the right things—without ever touching a real environment.  
* Even better: BoltSpec can use different ResultSets to model various conditions and validate that a plan's behavior is correct under all of them.  
* IMO, this accomplishes the goal of running plans under noop better than an ambiguous `--noop` flag could.

beechtom:
[...]

PR #2822, opened 2021-05-07
(GH-2755) Add support for running plans in noop mode

beechtom:
This adds basic support for running plans in noop mode. When running
`plan run --noop` or `Invoke-BoltPlan -Noop`, Bolt sets the executor
into noop mode. This automatically runs all tasks and apply blocks in a
plan in noop mode.

When running in noop mode, only a subset of plan functions are
supported. The plan functions that are not supported in noop mode are
`run_command`, `run_container`, `run_script`, `download_file`,
`upload_file`, `write_file`, and `file::write`, as they make changes to
either the controller or the specified targets and do not have a defined
noop mode. When any of these functions are encountered, the plan
immediately fails.

This also updates the `apply_prep` function to run the `puppet_library`
plugin in noop mode, which affects both the `apply --noop` command as
well as the `apply_prep` plan function. Because running this plugin in
noop mode might result in a target not having the puppet-agent package
installed, this also disables running the custom facts task in noop
mode, as the custom facts task depends on both the puppet-agent package
and Puppet's Ruby interpreter being available to the target.

!feature

* **Support running plans in noop mode**
  ([#2755](https://github.com/puppetlabs/bolt/issues/2755))

  Plans can now be run in no-operation mode from the command line using
  the `--noop` command-line option. Running in no-operation mode will
  automatically run all tasks and apply blocks in no-operation mode.
  Plans that use other plan functions to make changes to targets, such
  as `run_script`, will fail in no-operation mode, as these plan
  functions are not supported in this mode.

### DIFF
--- a/bolt-modules/boltlib/lib/puppet/functions/apply_prep.rb
+++ b/bolt-modules/boltlib/lib/puppet/functions/apply_prep.rb
@@ -74,6 +74,7 @@ Puppet::Functions.create_function(:apply_prep) do
 
     # Unfreeze this
     options = options.slice(*%w[_run_as _required_modules])
+    options['_noop'] = true if executor.noop
 
     applicator = Puppet.lookup(:apply_executor)
 
@@ -112,6 +113,13 @@ Puppet::Functions.create_function(:apply_prep) do
             # Give plan function options precedence over inventory options
             { 'target' => t,
               'hook_proc' => hook.call(opts.merge(options), t, self) }
+          # Catch and immediately re-raise noop errors so the entire plan fails,
+          # otherwise we'll get a failing result for each target. This error is
+          # raised by the plugin module, since that's where the plugin task is
+          # loaded and where Bolt checks if it can be run in noop mode.
+          rescue Bolt::Plugin::PluginError::NoopError
+            raise
+          # Catch all other errors and create a failing result.
           rescue StandardError => e
             Bolt::Result.from_exception(t, e)
           end
@@ -133,24 +141,30 @@ Puppet::Functions.create_function(:apply_prep) do
           need_install_targets.each { |target| set_agent_feature(target) }
         end
 
-        task = applicator.custom_facts_task
-        arguments = { 'plugins' => Puppet::Pops::Types::PSensitiveType::Sensitive.new(plugins) }
-        results = run_task(targets, task, arguments, options)
+        # If running in noop mode, skip the custom facts task. This task uses
+        # the Puppet Ruby interpreter by default, which won't be available
+        # unless the target already has the puppet-agent package installed as
+        # apply_prep does not install the puppet-agent package in noop mode.
+        unless executor.noop
+          task = applicator.custom_facts_task
+          arguments = { 'plugins' => Puppet::Pops::Types::PSensitiveType::Sensitive.new(plugins) }
+          results = run_task(targets, task, arguments, options)
 
-        # TODO: Standardize RunFailure type with error above
-        raise Bolt::RunFailure.new(results, 'run_task', task.name) unless results.ok?
+          # TODO: Standardize RunFailure type with error above
+          raise Bolt::RunFailure.new(results, 'run_task', task.name) unless results.ok?
 
-        results.each do |result|
-          # Log a warning if the client version is < 6
-          if unsupported_puppet?(result['clientversion'])
-            Bolt::Logger.deprecate(
-              "unsupported_puppet",
-              "Detected unsupported Puppet agent version #{result['clientversion']} on target "\
-              "#{result.target}. Bolt supports Puppet agent 6.0.0 and higher."
-            )
+          results.each do |result|
+            # Log a warning if the client version is < 6
+            if unsupported_puppet?(result['clientversion'])
+              Bolt::Logger.deprecate(
+                "unsupported_puppet",
+                "Detected unsupported Puppet agent version #{result['clientversion']} on target "\
+                "#{result.target}. Bolt supports Puppet agent 6.0.0 and higher."
+              )
+            end
+
+            inventory.add_facts(result.target, result.value)
           end
-
-          inventory.add_facts(result.target, result.value)
         end
       end
     end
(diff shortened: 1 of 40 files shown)


lib/bolt/bolt_option_parser.rb
```diff
@@ -847,9 +847,6 @@ def initialize(options)
              "Available filters are 'all', 'failure', and 'success'.") do |rerun|
         @options[:rerun] = rerun
       end
-      define('--noop', 'See what changes Bolt will make without actually executing the changes.') do |_|
```
lucywyman, line 850: I think this moved by accident?

beechtom: Specifically because I was making sure it showed up in the help text and couldn't find it because I didn't think to look under 'inventory options'. :|

lib/bolt/plugin/module.rb
```diff
@@ -267,6 +273,14 @@ def puppet_library(opts, target, apply_prep)
         options = {}
         options[:run_as] = meta_params['_run_as'] if meta_params['_run_as']
 
+        if meta_params['_noop']
```
lucywyman, line 276: This won't validate parameters for folks using the `task` plugin, this check needs to be there too.

bolt-modules/boltlib/lib/puppet/functions/apply_prep.rb
```diff
@@ -112,6 +113,13 @@ def apply_prep(target_spec, options = {})
             # Give plan function options precedence over inventory options
             { 'target' => t,
               'hook_proc' => hook.call(opts.merge(options), t, self) }
+          # Catch and immediately re-raise noop errors so the entire plan fails,
+          # otherwise we'll get a failing result for each target. This error is
+          # raised by the plugin module, since that's where the plugin task is
+          # loaded and where Bolt checks if it can be run in noop mode.
+          rescue Bolt::Plugin::Module::NoopError
```
lucywyman, line 120: Since this can also come from the Task plugin this probably needs to go in the Plugin class

lucywyman (changes requested):

bolt-modules/boltlib/lib/puppet/functions/apply_prep.rb
lucywyman: Also sorry I tried to leave these as comments and Github wanted it to be a review :P

bolt-modules/file/lib/puppet/functions/file/write.rb
```diff
@@ -15,8 +15,15 @@
   end
 
   def write(filename, content)
+    executor = Puppet.lookup(:bolt_executor)
```
lucywyman, line 18:
```suggestion
    executor = Puppet.lookup(:bolt_executor) {}
```
This will error if we don't provide it a default.

lucywyman (changes requested):
Last thing I caught: I think `BoltSpec::Run.run_plan` should respect running in noop, since it uses a real executor. Just need to thread the noop parameter through.

## Running apply_prep
### Using a plugin without noop support
- [x] Raises a single error
- [x] Errors if the task doesn't have metadata
- [x] Errors if the task specifies `supports_noop: false`

### Using a plugin with noop support
- [x] Runs apply_prep in noop

## Running tasks
- [x] Runs all tasks in noop mode
- [x] --noop overrides `_noop: false` plan metaparameter

## From a plan
- [x] `run_plan` accepts `_noop` metaparameter and runs in noop

## On the CLI
- [x] Runs in noop with `--noop`
- [x] Errors with `--no-noop`
- [x] --noop overrides `_noop: false` plugin configuration
- [x] --noop overrides `_noop: false` plan metaparameter
- [x] Errors if provided to a command other than `task run`, `apply`, or `plan run`
- [x] Runs subplans in noop as well

## Errors
- [x] Raises an error when non-noop plan functions are called:

## BoltSpec
- [ ] BoltSpec::Run can run a plan in noop mode

documentation/testing_plans.md
```diff
@@ -242,6 +242,52 @@ it 'asks who you are' do
 end
 ```
 
+### Running plans in no-operation mode
+
+The `run_plan` function has a `noop` option which can be used to run a plan in
+no-operation (no-op) mode. To run a plan in no-op mode, set the `noop` option
+to `true`:
```
hestonhoffman, line 249:
```suggestion
Use the `noop` option on the `run_plan` function to run a plan in
no-operation (no-op) mode. To run a plan in no-op mode, set the `noop` option
to `true`:
```

documentation/testing_plans.md
```diff
@@ -242,6 +242,52 @@ it 'asks who you are' do
...
+
+```ruby
+run_plan('myplan', {}, noop: true)
+```
+
+When you run a plan in no-op mode, all tasks, apply blocks, and sub-plans are
+automatically run in no-op mode. If a plan run in no-op mode invokes any of
+the following functions, then the plan automatically fails:
```
hestonhoffman, line 257:
```suggestion
the following functions, the plan automatically fails:
```

documentation/testing_plans.md
```diff
@@ -242,6 +242,52 @@ it 'asks who you are' do
...
+```puppet
+plan command (
+  TargetSpec $targets = 'localhost'
+) {
+  $result = run_command('touch example.txt', $targets)
+}
+```
+
+You can write a test that ensures that the `run_command` function is not
+invoked and that the plain fails with a helpful message:
```
hestonhoffman, line 280:
```suggestion
You can write a test that ensures that the plan does not invoke
`run_command` and that the plan fails with a helpful message:
```

hestonhoffman (approved):

lucywyman:
Just for clarity, we're going to hold off on moving forward with noop plans until we gather more information about what users want to use noop plans for, and how useful noop mode is in general for tasks.

beechtom:
Based on the infrequent use of no-op tasks (approximately 0.12% of all tasks run), and how likely users are to use and benefit from no-op plans, we're going to close this for now.

PR #2826, opened 2021-05-12
(bug) Do not lookup `:future` in Puppet for `file::*` functions

beechtom:
This fixes a bug in the `file::*` plan functions that was introduced
in #2731. Previously, these functions would look for future
configuration with a Puppet lookup (i.e. `Puppet.lookup(:future)`).
However, Bolt does not set this value in Puppet, causing the functions
to error. This updates the functions to only check the Bolt executor for
future configuration.

!bug

* **Do not error when looking up future configuration in `file::*` plan
  functions**

  The `file::read`, `file::readable`, and `file::exists` plan functions
  no longer error when checking for `future` configuration in Bolt.

### DIFF
--- a/bolt-modules/file/lib/puppet/functions/file/exists.rb
+++ b/bolt-modules/file/lib/puppet/functions/file/exists.rb
@@ -21,7 +21,7 @@ Puppet::Functions.create_function(:'file::exists', Puppet::Functions::InternalFu
     executor = Puppet.lookup(:bolt_executor) {}
     executor&.report_function_call(self.class.name)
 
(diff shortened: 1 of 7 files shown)


lucywyman:
This breaks `file::` plan functions inside apply blocks that are using the new script loading behavior. What we want is `Puppet.lookup(:future) { {} }` instead so that it defaults to an empty hash instead of `|| {}`

beechtom:
@lucywyman Already done 😅  https://github.com/puppetlabs/bolt/pull/2828

PR #2828, opened 2021-05-12
(bug) Do not error in `file::*` functions when future not configured

beechtom:
This fixes a bug where the `file::exists`, `file::read`, and
`file::readable` plan functions would error outside of an apply block if
the `future` configuration was not set.

!bug

* **Do not error in `file::*` plan functions when `future` is not
  configured**

  The `file::exists`, `file::read`, and `file::readable` plan functions
  no longer error when invoked outside of an apply block when `future`
  is not configured.


lucywyman:
:+1: Verified that calling `file::read` for example outside an apply block without `future` set raises the error. This both fixes the error, and makes it so that `file::` functions that use the future configuration both in and out of apply blocks also work.

lucywyman (approved):

Issue #2806, opened 2021-04-26
LXD transport: run_command does not accept _run_as param (and perhaps others)


dontlaugh:
## Describe the Bug
In a Bolt plan, the LXD transport `run_command` function does not accept the '_run_as' parameter, and throws an exception.

## Steps to Reproduce

Here is an example function call.
```pp
  run_command('systemctl daemon-reload', $targets, '_run_as' => 'root')
```

## Environment
 - Version 3.3.0
 - Platform Manjaro Linux

## Additional Context
Passing --run_as at the command line doesn't throw an error.

beechtom:
Fixed by #2838

PR #2838, opened 2021-05-19
(GH-2806) Support run_as for container transports using Bash

lucywyman:
This adds support for using `run-as` and associated configuration
options to the Docker, Podman, and LXD transports when running on *nix
systems. Originally, 'run-as' was not supported for the Docker transport
because it could be running on either Windows or *nix, and run-as is not
supported when running on Windows systems. At the time it seemed
confusing for the option to only be supported sometimes, but now that
this is the case for the SSH transport when using the PowerShell
`login-shell` there's more of a precedent for users understanding when
it is and isn't supported. This was intended to be
addressed after the great [shell-transport split of 2020](https://github.com/puppetlabs/bolt/pull/1702),
and just fell through the cracks.

In addition to adding support for `run-as` to Docker when running on
*nix, this adds support for the option on the related LXD and Podman
transports under the same circumstances, since those transports are
largely based on the Docker transport and operate in the same way. LXD
specifically does not support PowerShell, so `run-as` support requires
no caveats there.

This also fixes a bug where Bolt would stacktrace if `run-as` was set
for any of the above transports, because the Bash shell would try to
call `reset_cwd?` on the transport which was not defined.

!feature

* **Support run-as for container transports when running on \*nix** ([#2806](https://github.com/puppetlabs/bolt/issues/2806))
  The Docker, LXD, and Podman transports now support `run-as`
  configuration and related configuration options when running on *nix
  systems. `run-as` is not supported for any Windows systems or the
  PowerShell shell over SSH.

### DIFF
--- a/lib/bolt/config/transport/docker.rb
+++ b/lib/bolt/config/transport/docker.rb
@@ -15,7 +15,7 @@ module Bolt
           shell-command
           tmpdir
           tty
-        ].freeze
+        ].concat(RUN_AS_OPTIONS).sort.freeze
 
         DEFAULTS = {
           'cleanup' => true
@@ -26,6 +26,10 @@ module Bolt
 
           if @config['interpreters']
             @config['interpreters'] = normalize_interpreters(@config['interpreters'])
+          end
+
+          if Bolt::Util.windows? && @config['run-as']
+            raise Bolt::ValidationError, "run-as is not supported when using PowerShell"
           end
         end
       end
--- a/.github/workflows/schemas.yaml
+++ b/.github/workflows/schemas.yaml
@@ -7,7 +7,7 @@ on:
       - 'rakelib/schemas.rake'
       - 'schemas/*.json'
       - 'lib/bolt/config/options.rb'
-      - 'lib/bolt/config/transport/options.rb'
+      - 'lib/bolt/config/transport/*'
   pull_request:
     types: [opened, reopened, edited, synchronize]
     paths:
--- a/lib/bolt/config/transport/lxd.rb
+++ b/lib/bolt/config/transport/lxd.rb
@@ -11,7 +11,7 @@ module Bolt
           cleanup
           remote
           tmpdir
-        ].freeze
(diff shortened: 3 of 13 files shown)


lib/bolt/config/transport/docker.rb
```diff
@@ -27,6 +27,10 @@ class Docker < Base
           if @config['interpreters']
             @config['interpreters'] = normalize_interpreters(@config['interpreters'])
           end
+
+          if Bolt::Util.windows? && @config['run-as']
+            raise Bolt::ValidationError, "run-as is not supported when using PowerShell"
+          end
```
beechtom, line 33: Won't this prevent users running on Windows but connecting to a remote Docker host running Linux containers from using `run-as`?

Also, I think it might be more accurate to say "when using Windows containers" here.

lib/bolt/config/transport/docker.rb
```diff
@@ -15,7 +15,7 @@ class Docker < Base
           shell-command
           tmpdir
           tty
-        ].freeze
+        ].concat(RUN_AS_OPTIONS).sort.freeze
```
beechtom, line 18: The inventory schema needs to be regenerated to include these options. Looks like the CI job didn't get triggered since the paths don't include `lib/bolt/transport/**`. 😬

lib/bolt/config/transport/docker.rb
beechtom: Hrm, I guess `run-as` wouldn't be available anyways since we use the Powershell shell class when running Bolt on Windows.

lucywyman: Yeah, we already don't support `run-as` if you're connecting to *nix containers from Windows

lib/bolt/config/transport/docker.rb
lucywyman: Oh good call, updated the trigger for that workflow as well.

beechtom (approved):

PR #2842, opened 2021-05-20
(PE-31709) Structured errors for missing environment

donoghuc:
When asking for information about plans/tasks in bolt-server `environment` is required. Previously an unstructured string response which does not match other serialized error objects was returned. This commit updates that error to conform to the expected structure.

### DIFF
--- a/lib/bolt_server/transport_app.rb
+++ b/lib/bolt_server/transport_app.rb
@@ -686,7 +686,13 @@ module BoltServer
     end
 
     error Bolt::Error do |err|
-      [500, err.to_json]
+      # In order to match the request code pattern, unknown plan/task content should 400. This also
+      # gives us an opportunity to trim the message instructing users to use CLI to show available content.
+      if ['bolt/unknown-plan', 'bolt/unknown-task'].include?(err.kind)
+        [404, BoltServer::RequestError.new(err.msg.split('.').first).to_json]
+      else
+        [500, err.to_json]
+      end
     end
 
     error StandardError do
--- a/lib/bolt_server/transport_app.rb
+++ b/lib/bolt_server/transport_app.rb
@@ -10,6 +10,7 @@ require 'bolt/target'
 require 'bolt_server/file_cache'
 require 'bolt_server/plugin'
 require 'bolt_server/plugin/puppet_connect_data'
+require 'bolt_server/request_error'
 require 'bolt/task/puppet_server'
 require 'json'
 require 'json-schema'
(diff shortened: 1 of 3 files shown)


lib/bolt_server/transport_app.rb
```diff
@@ -257,7 +257,10 @@ def modulepath_from_environment(environment_name)
     end
 
     def in_pe_pal_env(environment)
-      return [400, '`environment` is a required argument'] if environment.nil?
+      if environment.nil?
+        err = Bolt::ValidationError.new("'environment' is a required argument")
```
nicklewis, line 261: If we added an `error Bolt::ValidationError` handler, we could just raise this error normally. Although it looks like there are a couple of places we _currently_ raise ValidationError that actually represent misconfigurations of bolt-server itself and properly should return 500 (they do now), so those would probably also need to change.

lib/bolt_server/transport_app.rb
```diff
@@ -84,40 +81,29 @@ def scrub_stack_trace(result)
       result
     end
 
-    def error_result(error)
```
donoghuc, line 87: The biggest change in this PR is *not* making schema errors into "result" like objects and instead treating them as any other RequestError. This has run against the task/plan acceptance tests in orch and no issues have come up. Still investigating manually. Primary: `fertile-massuh.delivery.puppetlabs.net`

steveax (approved):
👍

lib/bolt_server/transport_app.rb
```diff
@@ -727,6 +681,20 @@ def make_winrm_target(target_hash)
...
 
+    error BoltServer::RequestError do |err|
+      [400, err.to_json]
+    end
+
+    error Bolt::Error do |err|
+      # In order to match the request code pattern, unknown plan/task content should 400. This also
+      # gives us an opportunity to trim the message instructing users to use CLI to show available content.
+      if ['bolt/unknown-plan', 'bolt/unknown-task'].include?(err.kind)
+        [404, BoltServer::RequestError.new(err.msg.split('.').first).to_json]
```
ekinanp, line 692: This feels a bit too clever. Could we just hardcode this message so we remove the dependency on `err`'s message? E.g. `Unknown task/plan ....`

ekinanp (commented):
Looks good, just one comment.

ekinanp:
Nit: PR title should also change to reflect that this work also standardizes the bolt-server "invalid request" errors.

lib/bolt_server/transport_app.rb
donoghuc: that would be more work because we dont actually have the requested plan/task (that data is only in the error message). One option would be to change this in bolt. Perhaps have the unknown task/plan error only give the message about the unknown requested content and then the CLI handler print CLI specific messaging about running `show`.

ekinanp: 👍 Using `err.msg` to extract the unknown task/plan since that is the only way to do so makes sense. I'd add a comment explaining that reasoning.

Re: options to fix, probably extending Bolt to include it in `details` should be enough for us b/c that's where I'd expect that info to be.

PR #2844, opened 2021-05-20
(PE-31710) Remove backticks from bolt-server error messages

donoghuc:
Standardize on using single quotes instead of backticks in error messages.


donoghuc:
Folded in to https://github.com/puppetlabs/bolt/pull/2842/files

Issue #2832, opened 2021-05-14
Determine project name independent of the directory when loading files


lucywyman:
## Describe the Bug
When tarring modules to pass into an apply context, Bolt uses `Pathname.new(file).relative_path_from(parent)` to determine the "module_name" to store the files in once untarred. This assumes that the directory name is the same as the module name, which is not true in a few cases such as when using a `Boltdir` or when a project has a different name than its directory.

In `lib/bolt/applicator.rb#L309`, we need to get the path to the module and replace the name of the last element of the path (the directory containing the module) with the name that Puppet has for the module `mod.name`. We don't need to do this in `lib/bolt/transport/orch.rb` when packing files for upload because that's used for the `upload` functionality, which is expected to simply take a directory as is and upload it. You wouldn't expect to try to upload a file from a module and then load that file through Puppet without changing the names if they mismatched.

PR #2845, opened 2021-05-20
(GH-2832) Use module name for plugin tar directory

beechtom:
This updates the applicator to use the name of a module as the name of
the directory that plugin files are saved in when building the plugin
tarball. Previously, the applicator would use name of the module
directory as name of the directory that plugin files were saved to. This
caused issues when referencing a file in a project that had a configured
name that is different than the project directory's name.

For example, in a project named `foo` that uses a `Boltdir`, a file
might be located at `foo/Boltdir/files/somefile.txt`. When building the
plugin tarball, the applicator would set the path of this file to
`Boltdir/files/somefile.txt` instead of the correct path
`foo/files/somefile.txt`. This would cause any plan that referenced that
file using the correct syntax `puppet:///modules/foo/somefile.txt` to
error.

!feature

* **Upload project plugin files to correct directory when running an
  apply**
  ([#2832](https://github.com/puppetlabs/bolt/issues/2832))

  Project plugin files are now uploaded to the correct directory when
  running an apply. Previously, if a project used a `Boltdir` or had a
  directory name that did not match the project's configured name, apply
  blocks could not correctly reference files in the project using Puppet
  file syntax (`puppet:///modules/<project name>/<file name>`).

### DIFF
--- a/lib/bolt/applicator.rb
+++ b/lib/bolt/applicator.rb
@@ -306,11 +306,11 @@ module Bolt
       Puppet.lookup(:current_environment).override_with(modulepath: @plugin_dirs).modules.each do |mod|
         search_dirs = yield mod
 
-        parent = Pathname.new(mod.path).parent
+        tar_dir = Pathname.new(mod.name) # goes great with fish
         files = Find.find(*search_dirs).select { |file| File.file?(file) }
 
         files.each do |file|
-          tar_path = Pathname.new(file).relative_path_from(parent)
+          tar_path = tar_dir + Pathname.new(file).relative_path_from(mod.path)
           @logger.trace("Packing plugin #{file} to #{tar_path}")
           stat = File.stat(file)
           content = File.binread(file)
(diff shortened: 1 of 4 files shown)


lib/bolt/applicator.rb
```diff
@@ -306,11 +306,11 @@ def build_plugin_tarball
       Puppet.lookup(:current_environment).override_with(modulepath: @plugin_dirs).modules.each do |mod|
         search_dirs = yield mod
 
-        parent = Pathname.new(mod.path).parent
+        tar_dir = Pathname.new(mod.name) # goes great with fish
```
lucywyman, line 309: 😂

nicklewis: ooooooooof

beechtom: :D

lucywyman (approved):
- [x] Can reference files from a Boltdir project in an apply block
- [x] Can reference files from a project with a different name than it's directory

I also confirmed tests fail as expected when on the main branch, and everything looks correct to me!

lucywyman:
Oh, uh, noticed as I was merging this that I think the tag should be 'bug' instead of 'feature', but I think we can manually update that in the CHANGELOG

beechtom:
Ya, just noticed that too. I usually need to move stuff around and make edits anyways, so not a big deal. 🤷

PR #2846, opened 2021-05-20
(PE-31711) Make projects-dir a required bolt-server configuration value

donoghuc:
Bolt server endpoints that handle computing information about projects assume the projects-dir setting is configured. Make this a required setting to make that assumption safe.


steveax (approved):
👍

PR #2851, opened 2021-05-24
(bug) Use Pathname object when building plugin tar ball path

beechtom:
This fixes a bug in `Bolt::Applicator.build_plugin_tarball` that was
causing Bolt to stacktrace when running an apply block on a Windows
controller. When Bolt builds the tar ball, it joins the module's name
with the path to the module file relative to the module directory (i.e.
`my_module` + `files/some_file.txt`). To get the relative path to the
module file, Bolt calls the Ruby function `Pathname.relative_path_from`.
Previously, the string representation of the file's full path was passed
to this function, which then errored as it called `Pathname.cleanpath`
internally, which is not valid for string objects. Bolt now passes a
`Pathname` object to this function, fixing the stacktrace issue.

!no-release-note

### DIFF
--- a/lib/bolt/applicator.rb
+++ b/lib/bolt/applicator.rb
@@ -307,10 +307,11 @@ module Bolt
         search_dirs = yield mod
 
         tar_dir = Pathname.new(mod.name) # goes great with fish
+        mod_dir = Pathname.new(mod.path)
         files = Find.find(*search_dirs).select { |file| File.file?(file) }
 
         files.each do |file|
-          tar_path = tar_dir + Pathname.new(file).relative_path_from(mod.path)
+          tar_path = tar_dir + Pathname.new(file).relative_path_from(mod_dir)
           @logger.trace("Packing plugin #{file} to #{tar_path}")
           stat = File.stat(file)
           content = File.binread(file)


beechtom:
For reference, this is what caused the stacktrace in acceptance tests:

- https://github.com/puppetlabs/bolt/blob/main/lib/bolt/applicator.rb#L313 calls `Pathname.relative_path_from` where `mod.path` is a `String`.
- https://ruby-doc.org/stdlib-2.5.0/libdoc/pathname/rdoc/Pathname.html#method-i-relative_path_from is _supposed to_ convert that string to a `Pathname` object:
  ```ruby
  base_directory = Pathname.new(base_directory) unless base_directory.is_a? Pathname
  ```
- `Pathname.relative_path_from` errors with an undefined method error here:
  ```ruby
  base_directory = base_directory.cleanpath.to_s
  ```

And here's the error Bolt returns:
```json
{
  "kind": "bolt/pal-error",
  "msg": "undefined method `cleanpath' for \"C:/cygwin64/tmp/apply_ssh.oQXeNU/modules/example_apply\":String",
  "details": {
    "file": "C:/cygwin64/tmp/apply_ssh.oQXeNU/modules/example_apply/plans/init.pp",
    "line": 7,
    "column": 11
  }
}
```

beechtom:
Ran acceptance tests on the same set of targets that CI uses (`windows2019-64bolt%2Cwinrm.-debian10-64ssh.-windows2019-64winrm.`) and confirmed that the `apply_plan_ssh` tests are now passing:

```shell
tests/apply_plan_ssh.rb passed in 82.07 seconds
      Test Suite: tests @ 2021-05-24 14:04:16 -0700

       Total Suite Time: 82.07 seconds
      Average Test Time: 82.07 seconds
              Attempted: 1
                 Passed: 1
                 Failed: 0
                Errored: 0
                Skipped: 0
                Pending: 0
                  Total: 1
```

PR #2852, opened 2021-05-25
(Docs) Adds docs test

hestonhoffman:
This is a small test that generates the Bolt docs and then runs
the DITA OT in strict mode to check for any issues.

Strict mode fails early if it encounters an error.

### DIFF
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -3,10 +3,10 @@ name: Docs
 on:
   push:
     branches: [main]
-    paths-ignore: ['**.md', 'schemas/*']
+    paths-ignore: ['schemas/*']
   pull_request:
     types: [opened, reopened, edited, synchronize]
-    paths-ignore: ['**.md', 'schemas/*']
+    paths-ignore: ['schemas/*']
 
 jobs:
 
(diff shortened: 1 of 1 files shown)


.github/workflows/docs-test.yaml
```diff
@@ -0,0 +1,29 @@
...
+          bundler-cache: true
+      - name: Update gems
+        run: bundle update
+      - name: Generate docs
+        run: bundle exec rake docs:all
+      - name: Run DITA
+        uses: dita-ot/dita-ot-action@master
+        with:
+          build: |
+            dita -i documentation/puppetdb.ditamap -o out -f html5 --processing-mode=strict
```
beechtom, line 29: We should just add this bit to the existing docs workflow (https://github.com/puppetlabs/bolt/blob/main/.github/workflows/docs.yaml).

hestonhoffman: I did originally, but the `paths-ignore: ['**.md', 'schemas/*']` confused me. Does that only ignore `.md` files in the root directory?

beechtom: No, that'll ignore all markdown files in the repo. Since we want to make sure the dita build succeeds, I think removing `**.md` here is fine.

lucywyman (approved):
Looks like this pretty well matches the command run in production, and passed!

beechtom (approved):

PR #2858, opened 2021-05-28
(bug) Provide project for BoltSpec::Run::Runner#pal

jpartlow:
Ran into a case where using BoltSpec::Run.run_plan with a plan
containing a download_file() call would fail here:

https://github.com/puppetlabs/bolt/blob/3.9.0/bolt-modules/boltlib/lib/puppet/functions/download_file.rb#L99

because Puppet.lookup(:bolt_project) would otherwise be null at this
point.

Providing the config.project to the Bolt::PAL instance created by
BoltSpec::Run::Runner#pal fixes this.

### DIFF
--- a/lib/bolt_spec/run.rb
+++ b/lib/bolt_spec/run.rb
@@ -182,7 +182,10 @@ module BoltSpec
         @pal ||= Bolt::PAL.new(Bolt::Config::Modulepath.new(config.modulepath),
                                config.hiera_config,
                                config.project.resource_types,
-                               config.compile_concurrency)
+                               config.compile_concurrency,
+                               config.trusted_external,
+                               config.apply_settings,
+                               config.project)
       end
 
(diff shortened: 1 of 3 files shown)


jpartlow:
Not sure that this patch is in the right spot, or that the spec would be wanted, but I think it at least illustrates what I bumped into.

beechtom (changes requested):
Definitely the right place for this fix! Thanks for catching this.
When you have a chance, can you add a release note to the end of your commit? It should look like this:

```
!bug

* **Descriptive title of changes**
  ([#2858](https://github.com/puppetlabs/bolt/pull/2858))

  Descriptive summary of changes.
```

jpartlow:
Yep, I'll add that in.

beechtom (approved):

jpartlow:
Thanks! o/

Issue #2840, opened 2021-05-19
Add analytics around noop to see how it's being used in tasks and plans


lucywyman:
## Use Case
In order to understand how having plans run in noop would be useful, we'd like to know how many distinct users are actually running tasks or plan functions in noop mode, and how often. It'd also be useful to ask in channel and see if anyone is using it.

beechtom:
Closed by #2860

PR #2860, opened 2021-06-01
(GH-2840) Report whether tasks are run in noop mode

beechtom:
This adds a new analytics event that reports whether or not a task is
run in no-operation mode.

!feature

* **Report whether tasks are run in no-operation mode**
  ([#2840](https://github.com/puppetlabs/bolt/issues/2840))

  Bolt now reports whether or not a task is run in no-operation mode
  when it collects analytics.

### DIFF
--- a/bolt-modules/boltlib/lib/puppet/functions/run_task.rb
+++ b/bolt-modules/boltlib/lib/puppet/functions/run_task.rb
@@ -130,6 +130,9 @@ Puppet::Functions.create_function(:run_task) do
       end
     end
 
+    # Report whether the task was run in noop mode.
+    executor.report_noop_mode(executor.noop || options[:noop])
+
     if targets.empty?
       Bolt::ResultSet.new([])
     else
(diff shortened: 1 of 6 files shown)


beechtom:
Ran multiple tasks with and without noop and confirmed that they are showing up in analytics:
image

lucywyman:
Since this is such a small change can you also add this function call to the Orchestrator plan executor?
https://github.com/puppetlabs/orchestrator/blob/main/ruby/bolt-plan-executor-lib/bolt_plan_executor/executor.rb#L330 similarly with an empty function body.

lucywyman (approved):
### Test cases
- [x] Reports noop when running a task from the CLI with `--noop`
- [x] Reports when running a task from the CLI without `--noop`
- [x] Reports when running `run_task` plan function with noop function option
- [x] Reports when running `run_task` plan function with `--noop`
- [x] Reports when running `run_task` plan function without noop
- [x] Reports when running `run_task_with` plan function with noop

PR #2861, opened 2021-06-02
(maint) Standardize persisting scope for current plan

donoghuc:
THis commit updates the mock executor to match the fiber executor when persisting current plan scope.

### DIFF
--- a/lib/bolt_spec/plans/mock_executor.rb
+++ b/lib/bolt_spec/plans/mock_executor.rb
@@ -274,11 +274,7 @@ module BoltSpec
           local = Puppet::Parser::Scope::LocalScope.new
 
           # Compress the current scopes into a single vars hash to add to the new scope
-          current_scope = scope.effective_symtable(true)
-          until current_scope.nil?
-            current_scope.instance_variable_get(:@symbols)&.each_pair { |k, v| local[k] = v }
-            current_scope = current_scope.parent
-          end
+          scope.to_hash.each_pair { |k, v| local[k] = v }
           newscope.push_ephemerals([local])
         end
 


lib/bolt_spec/plans/mock_executor.rb
```diff
@@ -274,12 +274,7 @@ def create_future(scope: nil, name: nil)
           local = Puppet::Parser::Scope::LocalScope.new
 
           # Compress the current scopes into a single vars hash to add to the new scope
-          current_scope = scope.effective_symtable(true)
```
donoghuc, line 277: It almost looked like this would recursively pass through parent plan scope, but i did not actually see that in my (limited) testing. The main motivation is to standardize across the mock executor (which we are modeling after in our orch executor) and the fiber executor https://github.com/puppetlabs/bolt/blob/main/lib/bolt/fiber_executor.rb#L36L38 

It is a bit unclear to me if these are effectively the same, or if they both kind of do the same thing. I'm not quite sure how to get nested scopes meaningfully. Why were we previously trying to recurse through "parent" scopes if we only want the current scope? Also, does `to_hash` already do that recursion?

lucywyman: I can at least tell you why it's like that - I don't think it's wrong, but even if it is at least it'll be wrong in the same way as the fiber executor itself.

Before the fiber executor refactor that was how the executor would collect variables https://github.com/puppetlabs/bolt/blob/66d2908c4e0b9d7b0833e1823d9dd0d064859c59/lib/bolt/executor.rb#L385 (I looked through my notes, I remember there being an issue with parallel MatchScopes being loaded into blocks which would make variables from other futures available in all futures, but I can't reproduce it now). I think we realized they were equivalent and updated the fiber executor with the refactor https://github.com/puppetlabs/bolt/pull/2810/files#diff-893ebf9461c76f97dee50a194ee4d6a3529c2383a8cac903c83a2a883312837eR37 but just didn't update BoltSpec.

lucywyman (approved):
Test cases:
- [x] FiberExecutor and BoltSpec::MockExecutor now handle copying scope the exact same way
- [x] BoltSpec can successfully test plans with background blocks that:
  - [x] Use a variable defined outside the background block
  - [x] Set a variable in the background block, which is then not available outside the block
  - [x] Test a subplan with a background block that uses variables defined in the subplan outside the background block

Since this is already what we do in the FiberExecutor, I'm pretty comfortable with those abbreviated tests. Thanks Cas!

PR #2862, opened 2021-06-02
SOLARCH-278 updating bundle instructions to pdk

davidsandilands:
Since the setup uses pdk it would reduce complexity to use bundle via pdk for running and testing plans.

### DIFF
--- a/documentation/testing_plans.md
+++ b/documentation/testing_plans.md
@@ -31,17 +31,11 @@ Before you can write and run tests, you need to set up the test environment.
 1. Ensure your module has `Gemfile`, `Rakefile`, and `spec/spec_helper.rb`
    files.
 
-1. Install the `bundler` gem with Ruby:
-
-    ```shell
-    gem install bundler
-    ```
-
 1. Install the module's gem dependencies, which include Bolt and other testing
    tools:
 
     ```shell
-    bundle install
+    pdk bundle install
     ```
 
 ## Directory and file structure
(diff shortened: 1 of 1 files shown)


lucywyman (approved):
Thanks for this update!

hestonhoffman (approved):
🙌

lucywyman:
Ah, ok, I see now why we didn't initially include this command: it's experimental https://github.com/puppetlabs/pdk#experimental-features I'd like to wait until it's not an experimental change to merge this so that users don't run into unhandled errors or anything. I'll keep an eye on the feature though and I think this change would be great once the command is fully baked in!

jpogran:
`pdk bundle` is marked as experimental to indicate that anything executed that way is outside of the 'easy mode' path of the pdk command structure. Think 'advanced mode' instead of 'this does what I need for me without me typing things'.

documentation/testing_plans.md
```diff
@@ -31,17 +31,11 @@ Before you can write and run tests, you need to set up the test environment.
...
-    ```shell
-    gem install bundler
-    ```
-
 1. Install the module's gem dependencies, which include Bolt and other testing
    tools:
 
     ```shell
-    bundle install
+    pdk bundle install
```
jpogran, line 38: I haven't tested with bolt plans, but if you go the route of using the PDK then this should be the workflow:

```
> pdk new module foo
> cd foo
> # add tests
> pdk validate  # does a gem install using the PDK env, runs spec_prep, etc
> pdk test unit  # runs all tests, ensuring gem install, etc. equivalent of all the manual rake spec commands
```

davidsandilands: @jpogran you are absolutely right. I'll update the pull to use full pdk instructions. I wasn't looking at big picture. Thanks :-)

beechtom (approved):

PR #2865, opened 2021-06-02
(PT-317) Add project_facts_plugin_tarball endpoint to bolt-server

ekinanp:
This returns the base64 encoded tar archive of plugin code that is needed to
calculate custom facts.

### DIFF
--- a/lib/bolt_server/transport_app.rb
+++ b/lib/bolt_server/transport_app.rb
@@ -21,6 +21,10 @@ require 'puppet'
 # Needed by the `/project_file_metadatas` endpoint
 require 'puppet/file_serving/fileset'
 
+# Needed by the 'project_facts_plugin_tarball' endpoint
+require 'minitar'
+require 'zlib'
+
 module BoltServer
   class TransportApp < Sinatra::Base
     # This disables Sinatra's error page generation
@@ -70,6 +74,8 @@ module BoltServer
 
       # This is needed until the PAL is threadsafe.
       @pal_mutex = Mutex.new
+
+      @logger = Bolt::Logger.logger(self)
 
       super(nil)
     end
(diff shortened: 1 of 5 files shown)


lib/bolt_server/transport_app.rb
```diff
@@ -675,6 +679,51 @@ def make_winrm_target(target_hash)
...
 
+    # Returns the base64 encoded tar archive of plugin code that is needed to calculate
+    # custom facts
+    #
+    # @param versioned_project [String] the versioned_project to build the plugin tarball from
+    post '/project_facts_plugin_tarball' do
+      raise BoltServer::RequestError, "'versioned_project' is a required argument" if params['versioned_project'].nil?
+      content_type :json
+      in_bolt_project(params['versioned_project']) do |context|
+        # Inspired by Bolt::Applicator.build_plugin_tarball
```
ekinanp, line 690: `build_plugin_tarball` is basically "tar-up all of this stuff". That code will probably not change too much in the future and it seems basic enough that it's not worth introducing an implicit API requirement for Applicator to ensure that `build_plugin_tarball` isn't broken. So, I went with basically copy-pasting that code here.

lib/bolt_server/transport_app.rb
```diff
@@ -675,6 +679,51 @@ def make_winrm_target(target_hash)
...
+        sio = StringIO.new
+
+        begin
+          output = Minitar::Output.new(Zlib::GzipWriter.new(sio))
+
+          # TODO: Should this be full_modulepath
+          plugin_modulepath = context[:pal].user_modulepath
```
ekinanp, line 697: See above TODO.

ekinanp: Also, there is some logging in `build_plugin_tarball` -- see https://github.com/puppetlabs/bolt/blob/main/lib/bolt/applicator.rb#L315. Do we want to include that here?

donoghuc: I cant imagine we would need anything from the "builtin content dir". I though at one point in bolt we explicitly did not sync anything from our packaged builtin content (the core types and providers) but maybe i am not remembering that correctly. I know we allow users to configure what they want to sync https://github.com/puppetlabs/bolt/commit/3e0fe7f39533efc99da0102d6806872daabc901b but that is out of scope here. 

I'm a bit torn, i think maybe just foll modulepath would be better for now given we dont actually ship the core types/providers that bolt does?

donoghuc: as far as logging, I think logging the time it takes to pack and serialize the code would help us understand how important caching may be.

ekinanp: Ok, based on discussion in Slack we'll continue using the user modulepath.

I've updated the PR with logging, had to verify that the message appears at the expected level. Not sure if that should still be trace like in build_plugin_tarball or if we want to move it up more, maybe to INFO. I guess TRACE seems fine.

McdonaldSeanp (approved):

lib/bolt_server/transport_app.rb
```diff
@@ -675,6 +681,56 @@ def make_winrm_target(target_hash)
       end
     end
 
+    # Returns the base64 encoded tar archive of plugin code that is needed to calculate
+    # custom facts
+    #
+    # @param versioned_project [String] the versioned_project to build the plugin tarball from
+    post '/project_facts_plugin_tarball' do
```
donoghuc, line 688: In the ticket I specified this as `GET` why did we implement with POST?

lib/bolt_server/transport_app.rb
```diff
@@ -675,6 +681,56 @@ def make_winrm_target(target_hash)
...
+        start_time = Time.now
+        sio = StringIO.new
+
+        begin
+          output = Minitar::Output.new(Zlib::GzipWriter.new(sio))
+
+          # Bolt also sets this to user modulepath so do it here too
+          # for consistency
+          plugin_modulepath = context[:pal].user_modulepath
+          Puppet.lookup(:current_environment).override_with(modulepath: plugin_modulepath).modules.each do |mod|
```
donoghuc, line 702: Can we release the PAL lock quicker than this (maybe just look up all the files to pack, then pack them after we are out of the project context?

lib/bolt_server/transport_app.rb
ekinanp: Oh yeah, in case we extend this with a request body that may include additional input (similar to why project_inventory_targets was turned into a POST endpoint since GET w/ request bodies are weird). My gut feeling tells me that could be a possibility, but it is also a gut feeling so happy to switch back to a GET.

lib/bolt_server/transport_app.rb
ekinanp: Sure, updated.

lib/bolt_server/transport_app.rb
steveax: I vote GET, but based on ^^ that comment noticed that `project_inventory_targets` is a POST but retained the query param. We should move that into the POST body at some point.

ekinanp: Ok, updated to GET.

lib/bolt_server/transport_app.rb
```diff
@@ -675,6 +681,68 @@ def make_winrm_target(target_hash)
...
+        files
+      end
+
+      # Pack the plugin files
+      sio = StringIO.new
+      begin
+        output = Minitar::Output.new(Zlib::GzipWriter.new(sio))
+
+        plugin_files.each do |mod, files|
+          parent = Pathname.new(mod.path).parent
```
donoghuc, line 722: I have a hunch that there is a bug here somewhere around not using the module names (instead the paths) to build the archive structure. I have not fully formed this idea yet though... This looks intentional https://github.com/puppetlabs/bolt/blob/7b1caf6e314fabb21f84cdcf5900da5a2412869f/lib/bolt/applicator.rb#L309-L314 (using the name instead of the path)

ekinanp: I think I missed this because my version of Bolt was out-of-date.. Thanks, updated.

donoghuc (approved):

Issue #2866, opened 2021-06-03
Undef variables not set in Future scope


beechtom:
## Describe the Bug
Plan variables explicitly set to `undef` are not included in the scope for a Future. If code in a `background` or `parallelize` block references these variables, Bolt will error because the variable is not defined.

## Expected Behavior
The Future's scope should include variables explicitly set to `Undef`.

## Steps to Reproduce
Run the following plan, which sets a variable to `Undef` and references the variable from a `parallelize` block:

```puppet
plan example () {
  $foo = undef
  parallelize(['localhost']) |$target| {
    if $foo {
      out::message('MESSAGE')
    }
  }
  return('Done')
}
```

## Solution

When [creating the Fiber scope](https://github.com/puppetlabs/bolt/blob/main/lib/bolt/fiber_executor.rb#L37), Bolt needs to explicitly include variables with an `undef` variable in the call to the original scope's [`to_hash` function](https://github.com/puppetlabs/puppet/blob/2f14ddd269c26cba36771d9b3464c2c29422e46a/lib/puppet/parser/scope.rb#L667). 

It might be necessary to [filter out certain variables](https://github.com/reidmv/reidmv-change_risk/blob/main/lib/puppet/functions/change_risk.rb#L124-L128) from the new scope. Investigate further in testing.

## Additional context

- ~This seems to be an issue when running Bolt as a gem, but not as a package. Investigate further.~
- Regression introduced in 3.9.0 as part of the parallelism refactor. Previously symbols were pulled directly from the scope ( https://github.com/puppetlabs/bolt/blob/3.8.0/lib/bolt/executor.rb#L382-L387 ).

PR #2867, opened 2021-06-04
(GH-2866) Include undef variables in parallelize blocks

lucywyman:
When creating a new Future block, Bolt must copy any variables that
should be in scope for the block into a new scope for the block, so that
if variables redefined inside the block they don't affect other blocks
running in parallel. Gathering variables in scope uses the
`Puppet::Parser::Scope#to_hash` method, which accepts 2 positional
parameters:
- The first parameter tells the function whether to recursively get
  variables from parent (aka enclosing) scopes. This defaults to true
  and we want it to be true.
- The second parameter tells the function whether to include variables
  set to 'undef', which defaults to false and which we want to be true.

Previously we had not enabled including undef variables in the new
scope, which resulted in defined variables not being passed to the new
block as users expect. This changes how we call `to_hash` to correctly
include variables that are `undef`.

!bug

* **Undef variables are now included in Future block scopes** ([#2866](https://github.com/puppetlabs/bolt/issues/2866))
  Previously, Bolt would not include variables with a value `undef` in
  Future block scopes, leading them to be undefined. Bolt now includes
  those variables when creating the new scope for Future blocks.

### DIFF
--- a/lib/bolt/fiber_executor.rb
+++ b/lib/bolt/fiber_executor.rb
@@ -34,7 +34,7 @@ module Bolt
         local = Puppet::Parser::Scope::LocalScope.new
 
         # Compress the current scopes into a single vars hash to add to the new scope
-        scope.to_hash.each_pair { |k, v| local[k] = v }
+        scope.to_hash(true, true).each_pair { |k, v| local[k] = v }
         newscope.push_ephemerals([local])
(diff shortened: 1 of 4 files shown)


beechtom (approved):
Lgtm! 👍

Issue #2550, opened 2021-01-15
Move PowerShell module to PowerShell Gallery


beechtom:
## Use Case
The PowerShell module should be hosted on PowerShell Gallery

## Steps
- [x] Create PowerShell Gallery account
  - [x] File ticket with help desk to create account
- [x] Validate module is Gallery-compatible by following [publishing guidelines](https://docs.microsoft.com/en-us/powershell/scripting/gallery/concepts/publishing-guidelines?view=powershell-7.1)
- [x] Manually publish the module to the Gallery
- [x] Update Bolt docs to describe installing the PowerShell module
- [ ] Add a deprecation warning when the module is imported to the MSI package

### In Bolt 4.0
- [ ] Update bolt-vanagon to not include the module in the MSI

PR #2868, opened 2021-06-04
(GH-2550) Add deprecation warning to PuppetBolt module import

beechtom:
This adds a warning to the PuppetBolt module that the module will no
longer ship with Bolt packages in a future Bolt release. It also
provides information for installing the module through PowerShell
Gallery. The warning is only displayed the first time the module is
used, after which a 'first run' file is created in the module to
indicate that the message should not be printed.

This also adds an environment variable to the module publishing workflow
to not add this warning to the module that is published to PowerShell
Gallery.

!deprecation

* **PuppetBolt PowerShell module will not ship with Bolt packages in a
  future release**
  ([#2550](https://github.com/puppetlabs/bolt/issues/2550))

  The PuppetBolt PowerShell module will no longer ship with Bolt
  packages on Windows in a future release. The PuppetBolt PowerShell
  module should instead be installed from PowerShell Gallery.

### DIFF
--- a/pwsh_module/pwsh_bolt.psm1.erb
+++ b/pwsh_module/pwsh_bolt.psm1.erb
@@ -90,3 +90,20 @@ end
 
 }
 <% end %>
+
+<% unless ENV['BOLT_DISABLE_PWSH_WARNING'] -%>
+function Write-DeprecationWarning
+{
+  $FirstRunFile = "$HOME/.puppetlabs/etc/bolt/.first_runs_free"
+
+  if (-not(Test-Path -Path $FirstRunFile)) {
+    New-Item -ItemType File -Path $FirstRunFile -Force -ErrorAction SilentlyContinue
+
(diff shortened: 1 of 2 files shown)


pwsh_module/pwsh_bolt.psm1.erb
```diff
@@ -90,3 +90,20 @@ end
 
 }
 <% end %>
+
+<% unless ENV['BOLT_DISABLE_PWSH_WARNING'] -%>
+function Write-DeprecationWarning
+{
+  $FirstRunFile = "$PSScriptRoot/Private/first_run"
```
jpogran, line 97: You're going to have to put this in the user's bolt directory (`~/.puppetlabs/bolt`). You won't be able to write to the `Program Files\WindowsPowerShell\Modules` directory on a production system.

lucywyman (approved):

PR #2859, opened 2021-06-01
(GH-2815) Support plan-hierarchy lookups on the CLI

lucywyman:
This adds a new flag `--plan-hierarchy` to the `bolt lookup` CLI
command, which will perform a lookup as if in a plan outside an apply
block rather than in an apply block. The command optionally accepts
variable definitions for plan variable interpolation in hiera config.
The flag is mutually exclusive with targetting options, and either a
targetting option or `--plan-hierarchy` are required. The command
returns the bare value of the lookup, rather than a Result object.

This also updates the parallelize test to verify that `return`
statements return to only run on SSH infrastructure, since it keeps
falsely failing on Windows.

!feature

* **Lookup hiera plan_hierarchy values from the CLI** ([#2815](https://github.com/puppetlabs/bolt/issues/2815))

  The `bolt lookup` command now has a `--plan-hierarchy` flag that will
  lookup values from Hiera's `plan_hierarchy`.

PR #2869, opened 2021-06-05
(hack) Bash tab completion for Bolt commands

lucywyman:
This updates our tab completion script to be accurate with modern Bolt
commands and flags. It updates the script to only produce the options
available for a particular Bolt subcommand, and to only include the
flags after the entire Bolt command has been entered. This also adds
support for tab completing Bolt tasks and plans when they are cached in
the users current directory or the default Bolt project. This means Bolt
can't tab complete tasks and plans if the user is using a Boltdir, or if
they haven't run a command that caches tasks and plans yet.

In order to support tab completing task names, this also adds caching
for Task data that behaves exactly like Plan data caching. The cache is
registered or updated whenever types are generated, or when `bolt task
show` has been run without a cache. Bolt will check the mtime of all
task files to verify that the cache is up to date, and use the cache
data if available.

!feature

* **Builtin tab completion**

  Bolt now supports tab completion for Bolt commands, subcommands, and
  any flags available for a command. It also supports tab completing tasks
  and plans for certain project types when tasks and plans have been
  cached.


resources/bolt_bash_completion.sh
```diff
@@ -6,134 +6,171 @@
...
+      ;;
+    plan)
+      if [ $COMP_CWORD -eq 2 ]; then
+        next="convert new run show"
+      elif [[ $COMP_CWORD -eq 3 && $action != "new" ]]; then
+        if [ -f "${PWD}/.plan_cache.json" ]; then
+          # Use Puppet's ruby instead of jq since we know it will be there.
+          next=$(get_json_keys "${PWD}/.plan_cache.json")
+        elif [ -f "${HOME}/.puppetlabs/etc/bolt/.plan_cache.json" ]; then
+          next=$(get_json_keys "${HOME}/.puppetlabs/etc/bolt/.plan_cache.json")
```
beechtom, line 104:
```suggestion
        elif [ -f "${HOME}/.puppetlabs/bolt/.plan_cache.json" ]; then
          next=$(get_json_keys "${HOME}/.puppetlabs/bolt/.plan_cache.json")
```

resources/bolt_bash_completion.sh
```diff
@@ -6,134 +6,171 @@
...
+      ;;
+    task)
+      if [ $COMP_CWORD -eq 2 ]; then
+        next="run show"
+      elif [ $COMP_CWORD -gt 2 ]; then
+        if [ -f "${PWD}/.task_cache.json" ]; then
+          # Use Puppet's ruby instead of jq since we know it will be there.
+          next=$(get_json_keys "${PWD}/.task_cache.json")
+        elif [ -f "${HOME}/.puppetlabs/etc/bolt/.task_cache.json" ]; then
+          next=$(get_json_keys "${HOME}/.puppetlabs/etc/bolt/.task_cache.json")
```
beechtom, line 153:
```suggestion
        elif [ -f "${HOME}/.puppetlabs/bolt/.task_cache.json" ]; then
          next=$(get_json_keys "${HOME}/.puppetlabs/bolt/.task_cache.json")
```

resources/bolt_bash_completion.sh
```diff
@@ -6,134 +6,171 @@
...
+        fi
```
beechtom, line 154: Not sure if there's anything that can be done about this, but if the cache file for either plans or tasks doesn't exist then bash will try to autocomplete with the contents of the cwd:

image

beechtom (commented):
Task caching works as expected.

It might be helpful to include documentation for installing the `bash-completion` package, since not all systems will have it available. If the package is missing, the output from the autocompletion script is pretty bad:

image

resources/bolt_bash_completion.sh
lucywyman: I updated this to show flag options if the tasks aren't found - I don't want to show nothing, since I'm afraid folks will think that the tab-complete stopped working or something, but if there's a better thing to do (maybe print a message below or something about needing to run `bolt module generate-types` to tab complete task names?) I can try to make it happen.

resources/bolt_bash_completion.sh
```diff
@@ -6,134 +6,195 @@
...
-			local next=$(cat ${inventory_list_cache_file})
-		else
-			local jsondata=$(bolt --detail --targets all inventory show | sed -e '$d')
-			jsondata=${jsondata//\"/\\\"}
+          if [ -f "${PWD}/.plan_cache.json" ]; then
+            # Use Puppet's ruby instead of jq since we know it will be there.
+            next=$(get_json_keys "${PWD}/.plan_cache.json")
+          elif [ -f "${HOME}/.puppetlabs/bolt/.plan_cache.json" ]; then
+            next=$(get_json_keys "${HOME}/.puppetlabs/bolt/.plan_cache.json")
+          fi
```
beechtom, line 122: I think this also needs the same `else` statement as the `task` command so the options show if `.plan_cache.json` doesn't exist.

lucywyman: Good catch!

resources/bolt_bash_completion.sh
beechtom: I think just falling back to listing options is enough, especially since it seems like most users won't run into missing cache files frequently (especially considering the plan cache is generated whenever a module is installed).

resources/bolt_bash_completion.sh
```diff
@@ -6,134 +6,202 @@
...
+        ;;
+      inventory)
+        if [ $COMP_CWORD -eq 2 ]; then
+          next="show"
+        elif [ $action == 'show' ]; then
+          next="${inventory_opts} --project --format --inventoryfile --detail"
+        fi
+        ;;
+      lookup)
+        next="${action_opts} --hiera-config"
```
beechtom, line 96:
```suggestion
        next="${action_opts} --hiera-config --plan-hierarchy"
```

Now that https://github.com/puppetlabs/bolt/pull/2859 is merged.

resources/bolt_bash_completion.sh
```diff
@@ -6,134 +6,202 @@
...
+        else
+          case $action in
+            convert) next="${context_opts}";;
+            new) next="--project --pp";;
+            run) next="${action_opts} ${apply_opts} --params --tmpdir";;
+            show) next="${context_opts} --filter --format";;
+          esac
+        fi
+        ;;
+      project)
```
beechtom, line 139: This needs a case for `plugin show` above now that https://github.com/puppetlabs/bolt/pull/2872 is merged.

beechtom (changes requested):

beechtom (approved):

PR #2872, opened 2021-06-07
(hack) Add 'plugin show' command

beechtom:
This adds a new `plugin show` command, which lists the plugins available
to the project. Plugins are sorted by available hooks and include
descriptions if available.

!feature

* **Add `plugin show` command**

  The `bolt plugin show` command and `Get-BoltPlugin` PowerShell cmdlet
  can be used to list plugins available to the current project.


lib/bolt/outputter/json.rb
```diff
@@ -60,6 +60,10 @@ def print_tasks(tasks, modulepath)
         print_table('tasks' => tasks, 'modulepath' => modulepath)
       end
 
+      def print_plugin_list(plugins, modulepath)
```
lucywyman, line 63: I think this should also filter out `validate_resolve_reference` hooks.

lib/bolt/plugin.rb
```diff
@@ -250,6 +250,44 @@ def by_name(plugin_name)
       end
     end
 
+    # Loads all plugins and returns a map of plugin names to hooks.
+    #
+    def list_plugins
+      load_all_plugins
+
+      hooks = KNOWN_HOOKS.map { |hook| [hook, {}] }.to_h
```
lucywyman, line 258: Less readable but also less characters: `KNOWN_HOOKS.product([{}]).to_h`

lib/bolt/plugin.rb
```diff
@@ -250,6 +250,44 @@ def by_name(plugin_name)
...
+
+      @plugins.sort.each do |name, plugin|
+        # Don't show the Puppet Connect plugin for now.
+        next if name == 'puppet_connect_data'
```
lucywyman, line 262: Heh, I literally added 'Add private metadata for plugins' right after you demoed this to my list of hacks. Someone from Skeletor might be interested in taking that up.

spec/fixtures/plugin_modules/my_plug/tasks/resolve_reference.json
```diff
@@ -1,7 +1,8 @@
 {
-    "parameters": {
-        "optional_config": {
-            "type": "Optional[String]"
-        }
+  "description": "My custom plugin",
```
lucywyman, line 2: Is there an integration test attached to these?

lucywyman:
- [x] Shows the current project if it's a plugin
- [x] Shows plugins from the modulepath
- [x] Truncates long task descriptions
- [x] Raises a plugin loading disabled error if plugin loading is disabled
      **NOTE**: This is a little weird since it will error for the first plugin it tries to load, not for all plugins in general, but it at least fails fast and I don't think is worth making a new error type.
- [x] Errors if the hook specified does not exist
- [x] Formats correctly with JSON output
- [x] Formats reasonably with rainbow output

I did find one....kind of quirk? in testing this, I definitely don't think it's wrong but it's a little funny. If you have a task called `mymodule::resolve_reference` and configure that in your `bolt_plugin.json` to be a plugin for puppet_library say, then the plugin can be run both as a resolve_reference plugin and as a puppet_library plugin (and, I guess correctly, shows up in both places in the list of plugins).

lib/bolt/plugin.rb
beechtom: I considered using `product` but opted for readability. 😅

spec/fixtures/plugin_modules/my_plug/tasks/resolve_reference.json
lucywyman: Ah, no, I just assumed the fixtures were for an integration test but I see where they're used https://github.com/puppetlabs/bolt/pull/2872/files#diff-22978a4806ecd2ae6fd64f939c4ee442249a4fe838a399c43d459536f078ca1bR53

PR #2874, opened 2021-06-08
(bug) Do not error when using `--run-as` on Windows controller

beechtom:
This fixes a bug introduced by #2806 that causes Bolt to error on a
Windows controller whenever the `--run-as` command-line option is used.
Previously, the `docker` and `podman` transport config classes included
validation that would raise an error whenever the `run-as` option was
set for the transport and the controller was running Windows. Because
Bolt automatically sets the `run-as` configuration for all transports
that support it when using the `--run-as` command-line option, this
resulted in any command using that option to raise this error, even if
there weren't any targets using that transport.

This removes the relevant validation in the transport config classes.
This will not result in any errors when running on targets using these
transports, as Bolt automatically sets the transport's shell to
`powershell`, which does not include logic for running as another user
and cannot be overridden by the user.

!bug

* **Do not error when using `--run-as` on a Windows controller**

  Bolt no longer raises an error when using the `--run-as` command-line
  option on a Windows controller.

### DIFF
--- a/lib/bolt/config/transport/docker.rb
+++ b/lib/bolt/config/transport/docker.rb
@@ -27,10 +27,6 @@ module Bolt
           if @config['interpreters']
             @config['interpreters'] = normalize_interpreters(@config['interpreters'])
           end
-
-          if Bolt::Util.windows? && @config['run-as']
-            raise Bolt::ValidationError, "run-as is not supported when using PowerShell"
-          end
         end
       end
     end
--- a/lib/bolt/config/transport/podman.rb
+++ b/lib/bolt/config/transport/podman.rb
@@ -26,10 +26,6 @@ module Bolt
           if @config['interpreters']
             @config['interpreters'] = normalize_interpreters(@config['interpreters'])
           end
-
-          if Bolt::Util.windows? && @config['run-as']
-            raise Bolt::ValidationError, "run-as is not supported when using PowerShell"
-          end
         end
       end
     end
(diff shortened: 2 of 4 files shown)


lib/bolt/config/transport/docker.rb
```diff
@@ -27,10 +27,6 @@ class Docker < Base
           if @config['interpreters']
             @config['interpreters'] = normalize_interpreters(@config['interpreters'])
           end
-
-          if Bolt::Util.windows? && @config['run-as']
-            raise Bolt::ValidationError, "run-as is not supported when using PowerShell"
-          end
```
beechtom, line 33: We could potentially log a message here instead that indicates the transport does not support `run-as` on Windows and will be ignored, just in case users expect it to and are surprised. Since it would _always_ be logged when using `--run-as` on Windows, even when the transport isn't being used, it would probably want to be either at debug or trace level.

lucywyman: We should be doing this validation in the PowerShell shell instead of the transports, when the shell is initialized. I still think raising an error is probably best because performing an action as a user that you don't expect could be dangerous.

lib/bolt/shell/powershell.rb
```diff
@@ -11,6 +11,13 @@ class Powershell < Shell
       def initialize(target, conn)
         super
 
+        if target.options['run-as']
```
lucywyman, line 14: Ok, it turns out for transports that don't support run-as (so Winrm and even the `local` transport if we're running on Windows) we filter out all unsupported configuration options, so if `run-as` is set and using a transport that doesn't support it the configuration will never even get here to raise the warning. I think you can just remove this to avoid confusion and we can merge as is and have a separate ticket for deciding how we want to message this to users.

lucywyman (approved):

Issue #2788, opened 2021-04-14
Invoke-BoltScript Arguments Parameter Fails


glsc251:
## Describe the Bug
PowerShell CMDLET Invoke-BoltScript does not pass Arguments parameter to bolt correctly. You have to enclose them in single quotes and the underlying code should split that apart and pass it to bolt. It does not matter what the script is doing with the PS command it passes it as a single string not three parameters.

## Expected Behavior
The following works correctly under Windows or Linux
```sh
bolt script run path/to/script.sh arg1 arg2 arg3 --targets 10.0.68.18 --user root --password-prompt
```
The script is copied to the target and the three parameters are passed to the script.

## Steps to Reproduce
```PowerShell
Invoke-BoltScript -Script C:\path\to\script.sh -Arguments 'arg1 arg2 arg3' -Targets 10.0.68.18 -User root -PasswordPrompt -Transport ssh
```

## Environment
 - Version [3.6.1]
 - Platform [Windows 10 and Debian 10]

jpogran:
We'll first need to adjust the generator to make `-Arguments` have `ValueFromRemainingArguments` set to `$true`. This will tell the pwsh parser to take all non-bound args and pass them to `-Arguments`. 

The next step is to investigate https://github.com/puppetlabs/bolt/blob/main/pwsh_module/pwsh_bolt_internal.ps1#L32-L81 and determine if we need to add an array statement there so that the bolt commandline is generated correctly.

PR #2878, opened 2021-06-09
(GH-2788) Pass multiple arguments to PowerShell `-Arguments` parameter

beechtom:
This fixes a bug in the PuppetBolt PowerShell module that was preventing
multiple arguments from being passed to the `-Arguments` parameter. This
updates the parameter's `ValueFromRemainingArguments` parameter to
`true`, so any arguments not bound to another parameter are passed to
`-Arguments`. This supports the following cmdlet invocations:

- `Invoke-BoltScript -Script script.sh -Arguments foo,bar -Targets all`
- `Invoke-BoltScript -Script script.sh foo bar -Targets all`

!bug

* **Pass remaining arguments to PowerShell `-Arguments` parameter**
  ([#2788](https://github.com/puppetlabs/bolt/issues/2788))

  The PuppetBolt PowerShell module now correctly supports passing
  multiple arguments to the `-Arguments` parameter. Previously, any
  unbound arguments were not passed to this parameter, causing the
  PowerShell parser to error.

### DIFF
--- a/documentation/running_bolt_commands.md
+++ b/documentation/running_bolt_commands.md
@@ -416,10 +416,17 @@ bolt script run ./scripts/configure.sh --targets servers arg1 arg2
 
 _PowerShell cmdlet_
 
-To pass arguments to a script, use the `-Arguments` parameter:
+To pass arguments to a script, specify them after the command:
 
 ```powershell
-Invoke-BoltScript -Script ./scripts/configure.sh -Targets servers -Arguments arg1 arg2
+Invoke-BoltScript -Script ./scripts/configure.sh -Targets servers arg1 arg2
+```
+
+You can also use the `-Arguments` parameter and provide a comma-separated
+list of arguments:
+
+```powershell
+Invoke-BoltScript -Script ./scripts/configure.sh -Targets servers -Arguments arg1,arg2
 ```
 
 > 🔩 **Tip:** If an argument contains spaces or special characters, wrap them
--- a/pwsh_module/command.tests.ps1
+++ b/pwsh_module/command.tests.ps1
@@ -266,6 +266,16 @@ Describe "test all bolt command examples" {
       # This does work without quotes being explicitly added here
       $result | Should -Be "bolt script run myscript.sh echo hello --targets target1,target2"
     }
+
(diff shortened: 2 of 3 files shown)


beechtom:
I'm having trouble finding out if this is expected behavior, or if there's something else that needs to be changed to permit it, but the following doesn't work:

```powershell
 Invoke-BoltScript -Script test.sh -Targets nix -Arguments foo bar
Invoke-BoltScript : A positional parameter cannot be found that accepts argument 'bar'.
At line:1 char:1
+ Invoke-BoltScript -Script test.sh -Targets nix -Arguments foo bar
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidArgument: (:) [Invoke-BoltScript], ParameterBindingException
    + FullyQualifiedErrorId : PositionalParameterNotFound,Invoke-BoltScript
```

jpogran:
@beechtom try

```
Invoke-BoltScript -Script test.sh -Targets nix foo bar
```

`ValueFromRemainingArguments` is meant to fill one parameter will all non- bound arguments. If you specify `Arguments`, this means there are no parameters left unbound and then PowerShell thinks you mean every argument *should* have a matching parameter. Leaving `Arguments` out tells PowerShell "we matched these parameters, but these left over go to Arguments"

beechtom:
@jpogran Thanks for the extra info! I had tested the command you gave and it worked, just wasn't sure if `-Arguments foo bar` was or not. 😅

lucywyman (commented):
Confirmed the bug, and that both `Invoke-BoltScript -Script script.sh -Arguments foo,bar -Targets all` and `Invoke-BoltScript -Script script.sh foo bar -Targets all` work now. What *doesn't* work is passing arguments that are space-separated to `-Arguments`, which is expected but is documented wrong here https://puppet.com/docs/bolt/latest/running_bolt_commands.html#pass-arguments-to-a-script

beechtom:
Oh, good catch @lucywyman . Updated the docs example.

PR #2882, opened 2021-06-11
(bug) Fix `BoltSpec::Plans::MockExecutor#wait` method signature

beechtom:
This removes the `timeout` argument from
`BoltSpec::Plans::MockExecutor#wait`'s method signature so it matches
the same method signature on `Bolt::Executor`, which sets `timeout` as a
keyword argument.

!bug

* **Test plans that use `parallelize()` plan function in BoltSpec**

  BoltSpec no longer fails when testing plans that include the
  `parallelize()` plan function. Previously, testing plans that used
  this function would cause the plan to fail with an argument mismatch
  error.

### DIFF
--- a/lib/bolt_spec/plans/mock_executor.rb
+++ b/lib/bolt_spec/plans/mock_executor.rb
@@ -286,7 +286,7 @@ module BoltSpec
         future
       end
 
-      def wait(results, _timeout, **_kwargs)
+      def wait(results, **_kwargs)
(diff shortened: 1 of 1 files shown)


lucywyman:
:+1:  Verified the bug and that this makes plans with `parallelize` testable with BoltSpec. Also verified that using `background()` and `wait()` don't error.

Issue #2839, opened 2021-05-19
Refactor the Bolt::CLI class to be more focused


lucywyman:
## Use Case
Bolt's `Bolt::CLI` class takes input from the user and delegates it to the rest of Bolt to run what the user asked. Our CLI class has become a catch-all for lots of "quick" tasks (like showing guides, a lot of failing fast and doing validation that could be farmed out to other classes, etc.). It has grown to the point where it's difficult to add simple changes, such as calling `update_targets` based on a flag to a command instead of action (in the case of `lookup --plan-hierarchy`). Having such a large class that does so many things also makes it hard to understand and reason about.

[Sourcemaking](https://sourcemaking.com/refactoring) has a niceish guide on refactoring concepts, though I think it's important to not be dogmatic about our refactor. We can still have a 200+ line class and I don't think it's a problem as long as it's easy for the team members now to follow. It's hard to know what future developers may find easy or hard to understand, so while we should try to make the class generally "easy to follow" there's no strict definition of what that means and I don't want to get caught up in whether something is "understandable" or not as long as it's understandable for the team we have now and flexible to build on.

#### Class responsibilities
- Parse input from the user
- Validate that input from the user fits CLI-level validation (ie. `bolt command show` and `bolt task run --env-var` should error)
- Load Bolt project, configuration, and inventory if needed
- Log any warnings from loading config
- Case statement to call Bolt based on command the user is running
- Call into the outputter to print the output from running whatever we ran in Bolt
Basically, the CLI should just have a `parse` method and `execute` method that's the big switch statement, and then call into something else (not run another function on the CLI, with the exception of a helper).
Anything where the CLI class is doing some work on a result it got back from another class seems like a smell that the CLI is doing too much work. Preprocessing similarly seems like we should at least try to move that into the function the CLI calls, rather than have it live in the CLI.
Another way to think of this is as "how to make the backend implementation swappable" (and with that in mind, do we actually need to load config + project, and if we do make sure that that's an isolated call that can be swapped out).

### What is the CLI doing that it shouldn't be?
A good heuristic for this is if we had an interface other than a CLI, what would that also need to do (ie. API specific vs. interface specific). 

#### Some ideas
- We could have a CLI validator class (`Bolt::Validator::CLI`) that would validate the input, and do checks like `config.check_path_case` and `warn_inventory_overrides_cli` and `validate_ps_version`?
  - Commands as data. This would let us better define the different positional args and options that each command accepts and do some auto-validation.
[...]

PR #2885, opened 2021-06-15
(GH-2839) Move logic from `Bolt::CLI` to new `Bolt::Application` class

beechtom:
This moves a bulk of the logic for executing Bolt's commands from the
`Bolt::CLI` class to a new `Bolt::Application` class. Previously, the
`Bolt::CLI` class had multiple responsibilities:

- Parsing the command
- Validating the command
- Initializing all of Bolt's classes
- Determing which command to execute
- Executing each command, including much of the logic to do so

This left the `Bolt::CLI` class overloaded, as it was essentially the
sole interface for running Bolt. This made it difficult at times to
modify existing behavior or add new commands, as the class was
incredibly tangled. Classes needed to be loaded and methods needed to be
invoked in a very particular order, otherwise Bolt would break.

This marks the first step in a larger refactoring of the `Bolt::CLI`
class. Further refactoring might include:

- Move responsibilities for validating commands to the parser
- Split the `Bolt::Application` class into several smaller subclasses
  that are each responsible for running a specific subcommand
- Offload resposibility for loading some classes, such as `Bolt::Rerun`
  and `Bolt::PAL`, to `Bolt::Application`
- Defering resolving inventory until targets are needed

!no-release-note


beechtom:
Response to Nick's comments from the previous PR:

**Rerun seems like it belongs in the CLI class, since it involves storing and retrieving data between runs**

Agreed. Paired with the comment below re: exit codes, it makes sense for the application to return the actual result which can be processed by the CLI. This also makes sense for an eventual API so the application returns something useful other than a number. :P

**More broadly, resolving targets also feels like a CLI responsibility, since it's a CLI convenience to be able to specify them in various formats. Plus, a method signature like `command_run(command, targets, options)` makes a lot of sense.**

Agreed on it being a CLI convenience. The one part I'm a little iffy on here is that requires the CLI to load `Bolt::PAL` and `Bolt::Plugins`, which I was hoping to eventually have the application be responsible for. The intent was for an eventual API to not require users to initialize these classes and pass them in, but if we have some sort of `Bolt::Engine` class that is the API equivalent to the CLI, then that could do effectively the same thing.

**Related, any method on Application that is taking `options` should probably instead take specific positional or keyword arguments to make it explicit what it uses.**

Agreed. I was originally going to hold off on this until we could update the parser to filter out invalid options from the hash, but I think it'd also make things clearer to just pass specific kwargs now.

**The `SUCCESS`/`FAILURE` constants are great, but exit code is definitely a CLI concern. The Application should either return true/false for status or return a proper result object that is interpreted by the CLI (it can do either for a given method, doesn't have to be consistent)**

Agreed. See above.

**Invoking the outputter should happen in the CLI, with the Application just returning the data that will be output. Methods that call `executor.subscribe` should either be `application.subscribe` in the CLI or pass `outputter` as an argument to the Application method.**

Agreed that the CLI should invoke the outputter and print results. Having an `application.subscribe` method would be nice. Running plans and applies has some special handling for subscribing the outputter -- I'll explore whether there's a way to do that without passing the outputter to the `plan_run`/`apply` methods.

**`executor.shutdown` should probably be `application.shutdown` in the CLI (which presumably just calls `executor.shutdown` :smile:). Conceptually you should be able to build the Application once and call multiple methods on it, even if there's currently some state that makes that not work in practice.**

Agreed.

beechtom:
Made some changes per Nick & Lucy's comments in the [previous PR](https://github.com/puppetlabs/bolt/pull/2876). It creates a clearer split between the responsibilities of the CLI and Application classes. Specifically, the CLI is responsible for parsing and processing the command string, configuring the application, outputting results, updating the rerun file, and interpreting results as exit codes. The Application is only responsible for actually doing the 'work' for the command and returning data (or in some cases just a boolean), and does not need to know about implementations details of the CLI (except when we explicitly pass an outputter, like for `project migrate`).

There's definitely still more work to be done, and I'm not 100% sure about the current split.

lib/bolt/outputter/human.rb
```diff
@@ -492,11 +492,11 @@ def print_module_list(module_list)
         end
       end
 
-      def print_plugin_list(plugin_list, modulepath)
+      def print_plugin_list(plugin_data)
```
lucywyman, line 495: In general I'd kind of like to move towards named keyword args for things like this so that it's more clear what the data is and what parameters the function takes, so here it'd be (similar to above) `plugins:, modulepath:` and then called `print_plugin_list(**plugin_data)`

lib/bolt/plan_creator.rb
```diff
@@ -85,17 +85,16 @@ def self.create_plan(plans_path, plan_name, outputter, is_puppet)
         run_command  = 'bolt plan run'
       end
 
-      output = <<~OUTPUT
+      # TODO: This should probably return the name and path to the new
```
lucywyman, line 88: For sure - it doesn't seem like the JSON outputter needs something this readable.

lib/bolt/project.rb
```diff
@@ -209,12 +209,5 @@ def validate
         Bolt::Logger.warn("missing_project_name", message)
       end
     end
-
-    def check_deprecated_file
```
lucywyman, line 213: Can this get moved to `Bolt::Project#validate`?

lib/bolt/cli.rb
```diff
@@ -24,14 +25,15 @@
...
 require 'bolt/rerun'
-require 'bolt/secret'
 require 'bolt/target'
 require 'bolt/version'
 
 module Bolt
   class CLIExit < StandardError; end
 
   class CLI
+    attr_reader :outputter, :rerun
```
lucywyman, line 35: Is the outputter just for testing? I'm not sure where rerun gets called?

beechtom: Both are set as instance variables in `#execute` and the readers are called in `#process_command`. Mostly set these as instance variables since they're specific to the CLI, but we can pass them as parameters to `#process_command` and get rid of the readers.

lib/bolt/cli.rb
```diff
@@ -379,710 +309,563 @@ def validate(options)
...
+                    else
+                      Bolt::Project.find_boltdir(Dir.pwd)
+                    end
 
-    def query_puppetdb_nodes(query)
-      puppetdb_client.query_certnames(query)
-    end
+          config = Bolt::Config.from_project(project, options)
+
+          # TODO: Should this be an instance variable? Outputter is specific
```
lucywyman, line 359: I don't know that it matters too much? Having it be an instance variable makes sense to me :shrug:

lib/bolt/cli.rb
```diff
@@ -379,710 +309,563 @@ def validate(options)
...
+            config.trace,
+            config.spinner
+          )
+
+          # TODO: Should this be an instance variable? Rerun is specific
+          #       to the CLI, so might make sense to have it be a property of
+          #       the class.
+          @rerun = Bolt::Rerun.new(config.rerunfile, config.save_rerun)
+
+          # TODO: Should this be an input to the executor? Might make sense
```
lucywyman, line 375: Yeah I think having Analytics subscribe to the executor makes a lot of sense, and would mean that Skeletor wouldn't necessarily have to override all of these functions.

lib/bolt/cli.rb
```diff
@@ -379,710 +309,563 @@ def validate(options)
...
-    def warn_inventory_overrides_cli(opts)
-      inventory_source = if ENV[Bolt::Inventory::ENVIRONMENT_VAR]
-                           Bolt::Inventory::ENVIRONMENT_VAR
-                         elsif config.inventoryfile
-                           config.inventoryfile
-                         elsif File.exist?(config.default_inventoryfile)
-                           config.default_inventoryfile
-                         end
+          # TODO: Configure logger with a separate method call?
+          Bolt::Logger.configure(config.log, config.color, config.disable_warnings)
```
lucywyman, line 385: 🤷🏼  Up to you, either is fine by me.

lib/bolt/cli.rb
```diff
@@ -379,710 +309,563 @@ def validate(options)
...
-    end
+    # Execute a Bolt command. The +options+ hash includes the subcommand and
+    # action to be run, as well as any additional arguments and options for the
+    # command.
+    #
+    # @param options [Hash] The CLI options.
+    #
+    def execute(options)
+      with_signal_handling do
+        with_error_handling do
```
lucywyman, line 335: I think this with_error_handling can be moved into the `exe/bolt` rescue statement so we don't have to wrap parse and execute in it, and that can just call `cli.fatal_error`.

lib/bolt/cli.rb
```diff
@@ -379,710 +309,563 @@ def validate(options)
...
+
+          log_outputter = Bolt::Outputter::Logger.new(options[:verbose], config.trace)
+
+          #
+          # FINALIZING SETUP
+          #
+
+          check_gem_install
+
+          # TODO: Discuss possibility of Bolt 4.0 having CLI overrides take
```
lucywyman, line 420: I think the big issue here is target-specific config vs. "global" config, but since we resolved the inventory all at once we don't know what came from target-specific config or top-level config, so everything has to override the CLI. Now that top-level inventory config is supported in the defaults directory which does get overridden by the CLI, documenting that inventory config can go there and then be overridden seems good + then we can point to those docs in the message. But I don't think having "top-level" config from the CLI override specific options makes sense.

lib/bolt/cli.rb
```diff
@@ -379,710 +309,563 @@ def validate(options)
...
-      inventory_cli_opts = %i[authentication escalation transports].each_with_object([]) do |key, acc|
-        acc.concat(Bolt::BoltOptionParser::OPTIONS[key])
-      end
+          # TODO: Include logic for handling special cases when subscribing the
+          #       outputter (e.g. `plan run` and `apply`).
+          executor.subscribe(outputter)
+          executor.subscribe(log_outputter)
 
-      inventory_cli_opts.concat(%w[no-host-key-check no-ssl no-ssl-verify no-tty])
+          # TODO: Figure out where this should really go. It doesn't seem to
```
lucywyman, line 447: Yeah, this and the clear cache lines above + script line below don't really have a clear organizational place. I think you could group them in a separate function call if you want, or they're fine here :shrug: I personally don't mind long functions especially since a lot of this is a single function call with newlined arguments.

lib/bolt/project.rb
beechtom: I think this could probably be removed altogether since `project.yaml` was only present in 2.8.0 and 2.9.0 and was listed as experimental at the time.

lib/bolt/outputter/human.rb
```diff
@@ -308,7 +308,7 @@ def format_table(results, padding_left = 0, padding_right = 3)
         )
       end
 
-      def print_tasks(tasks, modulepath)
+      def print_tasks(tasks:, modulepath:)
```
lucywyman, line 311: Yessssssss

lucywyman: Also just checked this is only ever called from Bolt https://github.com/search?q=org%3Apuppetlabs+print_tasks&type=code :+1:

lucywyman (commented):
Made at least a little bit of progress - more tomorrow!

lib/bolt/outputter/human.rb
```diff
@@ -533,12 +533,31 @@ def print_plugin_list(plugin_list, modulepath)
...
+          Created plan '#{name}' at '#{path}'
+  
+          Show this plan with:
+              #{show_command} #{name}
+          Run this plan with:
+              #{run_command} #{name}
+        OUTPUT
+      end
+
+      def print_targets(target_data, target_flag)
```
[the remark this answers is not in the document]
beechtom: Yup yup!

spec/bolt/cli_spec.rb
```diff
@@ -1,1120 +1,1204 @@
...
+    allow(Bolt::Executor).to receive(:new).and_return(executor)
+    allow(Bolt::Inventory).to receive(:from_config).and_return(inventory)
+    allow(Bolt::Outputter).to receive(:for_format).and_return(outputter)
+    allow(Bolt::PAL).to receive(:new).and_return(pal)
+    allow(Bolt::Plugin).to receive(:setup).and_return(plugin)
+    allow(Bolt::Project).to receive(:create_project).and_return(project)
+    allow(Bolt::Rerun).to receive(:new).and_return(rerun)
+
+    # Allow the stubbed outputter to yield from its spinner.
+    allow(outputter).to receive(:spin) { |&block| block.call }
```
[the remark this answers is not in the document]
beechtom: I'm not sure I understand? These tests stub the outputter, and this line is specifically just to allow it to execute the block, otherwise it just returns the stub.

lib/bolt/application.rb
```diff
@@ -0,0 +1,612 @@
...
+    # @param destination [String] The destination path on the targets.
+    # @param targets [Array[String]] The targets to run on.
+    # @return [Bolt::ResultSet]
+    #
+    def file_upload(source, destination, targets)
+      future  = executor.future&.fetch('file_paths', false)
+      source  = find_file(source, future)
+      targets = inventory.get_targets(targets)
+
+      Bolt::Util.validate_file('source file', source)
```
lucywyman, line 124: This needs to pass `true` as a third argument so that directories can be uploaded.
```
lucy ~/githubs/bolt ((afd2cdd0...)) $ be bolt file upload documentation/ ~/ -t localhost
The source file 'documentation/' is not a file
lucy ~/githubs/bolt ((afd2cdd0...)) $ bolt file upload documentation/ ~/ -t localhost
Started on localhost...
Finished on localhost:
  Uploaded 'documentation/' to 'localhost:/home/lucy/'
Ran on 1 target in 0.02 sec
```

beechtom: Yep, caught this while working on getting tests working again. Good catch! :D

lib/bolt/application.rb
```diff
@@ -0,0 +1,612 @@
...
+        count: inventory.group_names.count,
+        groups: inventory.group_names.sort,
+        inventory: {
+          default: config.default_inventoryfile.to_s,
+          source: inventory.source
+        }
+      }
+    end
+
+    # TODO: Should this take an outputter or return data that the CLI
```
lucywyman, line 146: I think returning data to be printed by the CLI makes more sense. It's what we do for things like `group show` and aligns more with each classes' role.

lib/bolt/application.rb
```diff
@@ -0,0 +1,612 @@
...
+    #       prints with the outputter?
+    #
+    # Show Bolt guides.
+    #
+    # @param topic [String, NilClass] The topic to show.
+    # @param outputter [Bolt::Outputter] An outputter instance.
+    # @return [Boolean]
+    #
+    def guide(topic, outputter)
```
lucywyman, line 155: Seems like topic can be an optional param + default to nil

lib/bolt/application.rb
```diff
@@ -0,0 +1,612 @@
...
+      end
+      true
+    end
+
+    # Show inventory information.
+    #
+    # @param targets [Array[String], NilClass] The targets to show.
+    # @return [Hash]
+    #
+    def inventory_show(targets)
```
lucywyman, line 207: This also seems like targets could default to nil.

lib/bolt/cli.rb
```diff
@@ -379,710 +341,601 @@ def validate(options)
...
-      # submit a different screen for `bolt task show` and `bolt task show foo`
-      if options[:action] == 'show' && options[:object]
-        screen += '_object'
-      end
+      when 'guide'
+        app.guide(options[:object], outputter)
+        SUCCESS
+
+      when 'inventory'
+        targets = app.inventory_show(options[:targets])
```
lucywyman, line 593: Ah, this is a tricky one: we need to validate that multiple targeting options aren't passed here, since the dreaded `if` statement on line 348 doesn't check when `inventory show` is run since it *can* have no targeting options, but still can't have *multiple* targeting options. I think that check can go here similar to how the other check is done in this class.

```
lucy ~/githubs/bolt ((afd2cdd0...)) $ be bolt inventory show --targets localhost --rerun all
Targets
  localhost    (Not found in inventory file)

Inventory source
  /home/lucy/githubs/bolt/inventory.yaml

Target count
  1 total, 0 from inventory, 1 adhoc

Additional information
  Use the '--detail' option to view target configuration and data
lucy ~/githubs/bolt ((afd2cdd0...)) $ bolt inventory show --targets localhost --rerun all
Only one targeting option '--targets', '--rerun', or '--query' can be specified
```

lib/bolt/cli.rb
```diff
@@ -379,710 +341,601 @@ def validate(options)
...
-
-      update_targets(options)
-
-      inventory_targets, adhoc_targets = options[:targets].partition do |target|
-        known_names.include?(target.name)
-      end
-
-      { inventory: inventory_targets, adhoc: adhoc_targets }
-    end
+    private def process_target_list(pdb_client, rerun, options)
```
lucywyman, line 729: Ah I see - can we check for multiple targeting options here? Targeting options can be optional, but anything that does accept targeting options should only ever have one specified.

lib/bolt/application.rb
```diff
@@ -0,0 +1,612 @@
...
+          targets: targets[:inventory].map(&:detail),
+          file: (inventory.source || config.default_inventoryfile).to_s,
+          default: config.default_inventoryfile.to_s
+        },
+        targets: targets.values.flatten.map(&:detail),
+        count: targets.values.flatten.count
+      }
+    end
+
+    # TODO: This should probably be separate #lookup and #plan_lookup methods.
```
lucywyman, line 226: Heh, 110%

lib/bolt/application.rb
```diff
@@ -0,0 +1,612 @@
...
+    #
+    # @param outputter [Bolt::Outputter] An outputter instance.
+    # @param force [Boolean] Forcibly install modules.
+    # @param resolve [Boolean] Resolve module dependencies.
+    # @return [Boolean]
+    #
+    def module_install(outputter, force: false, resolve: true)
+      assert_project_file(config.project)
+
+      if config.project.modules.empty? && force
```
lucywyman, line 289: Why did this get changed from `resolve` to `force`? It seems like this should still get skipped if resolve is false, and not be based on `force`...

lucywyman: Sorry, the previous statement for reference:
```
if project.modules.empty? && resolve != false
```

lib/bolt/application.rb
```diff
@@ -0,0 +1,612 @@
...
+
+    # Run a plan.
+    #
+    # @param plan [String] The plan to run.
+    # @param targets [Array[String], NilClass] The targets to pass to the plan.
+    # @param params [Hash] Parameters to pass to the plan.
+    # @return [Bolt::PlanResult]
+    #
+    def plan_run(plan, targets, params: {})
+      if targets
```
lucywyman, line 343: Hm, targets can still be empty here - most likely from a PDB query, but also from rerun (or you could probably even get there with JSON params). I think `if targets && !targets.empty?`?

lib/bolt/application.rb
```diff
@@ -0,0 +1,612 @@
...
+      plan_context = { plan_name: plan, params: params }
+
+      executor.start_plan(plan_context)
+      result = pal.run_plan(plan, params, executor, inventory, plugins.puppetdb_client)
+      executor.finish_plan(result)
+
+      result
+    end
+
+    # TODO: This should probably be separate #show_plan and #list_plans methods.
```
lucywyman, line 376: Also 110%.

lib/bolt/application.rb
```diff
@@ -0,0 +1,612 @@
...
+    # @param env_vars [Hash] Environment variables to set on the target.
+    # @return [Bolt::ResultSet]
+    #
+    def script_run(script, targets, arguments: [], env_vars: {})
+      future = executor.future&.fetch('file_paths', false)
+      script = find_file(script, future)
+
+      Bolt::Util.validate_file('script', script)
+
+      with_benchmark do
```
lucywyman, line 436: I just realized that previously we were doing all the file lookup and validation as part of the benchmark, and now it only wraps the actual script running. I personally don't particularly care one way or the other, but it'll be interesting to see if people notice Bolt *reporting* it ran a bit faster when it didn't...

lucywyman: I feel like the real question is why we benchmark + report how quickly Bolt ran in the first place, and I think that it's mostly so people can verify and report issues when Bolt takes a long time without having to do their own benchmarking, which seems like a case for wrapping everything in a benchmark....but then again we only wrap commands that connect remotely in a benchmark, and finding the file definitely isn't part of that remote connection...Like, subscribing the outputted to the executor also used to be part of the benchmark and while that shouldn't take any noticeable amount of time it's also not meaningful to benchmark - basically there are some obvious cases, and I feel like file loading is kind of a gray area.

lib/bolt/application.rb
```diff
@@ -0,0 +1,612 @@
...
+    #
+    # @param task [String, NilClass] The name of the task to show.
+    # @param filter [String, NilClass] A string to filter tasks by.
+    # @return [Boolean]
+    #
+    def task_show(task, filter: nil)
+      if task
+        pal.get_task(task)
+      else
+        tasks = filter_content(pal.list_tasks(filter_content: true), filter)
```
lucywyman, line 512: It'd be nice if this was `list_tasks_with_cache` just so that the task cache stays up to date...but strictly speaking it doesn't need to use the cache.

lib/bolt/outputter/json.rb
```diff
@@ -115,35 +119,38 @@ def print_puppetfile_result(success, puppetfile, moduledir)
...
+      end
+
+      # Print target names and where they came from.
+      #
+      # @param adhoc [Hash] Adhoc targets provided on the command line.
+      # @param inventory [Hash] Targets provided from the inventory.
+      # @param targets [Array] All targets.
+      # @param count [Integer] Number of targets.
+      #
+      def print_target_info(adhoc:, inventory:, targets:, count:, **_kwargs)
```
lucywyman, line 143: Definitely justified in saying this is less readable, but this could be:
```
      def print_target_info(**kwargs)
        @stream.puts(kwargs.slice(*%i[adhoc inventory targets count]).to_json)
     end
```

Or honestly probably just
```
def print_target_info(**kwargs)
  @stream.puts(kwargs.to_json)
end
```

lib/bolt/application.rb
beechtom: Ya, my reasoning for wrapping just the executor portion is that it seems like the only part you'd really care to benchmark. File lookup and validation and subscribing to the executor should have a negligible impact on execution time, and we already were doing a fair amount of setup/validation in the CLI prior to reaching the benchmarking anyway. 🤷 Honestly I don't really feel strongly one way or the other.

lib/bolt/application.rb
beechtom: Derp. The logic needed updating since we don't need to account for `resolve` being `nil` anymore, but I accidentally switched it to `force`. 🤦

lib/bolt/application.rb
beechtom: Updated to return data for `group show` and error if given an unknown topic. Previously we showed the same output as `bolt guide`, but that gets a bit trickier to handle if `bolt guide <topic>` can return different data depending on if the topic is known or unknown. Erroring also seems to be in line with `show` commands that receive an unknown object.

lib/bolt/application.rb
```diff
@@ -0,0 +1,613 @@
...
+      Bolt::ProjectManager.new(config, outputter, pal)
+                          .create(Dir.pwd, name, modules)
+    end
+
+    # Migrate a project to current best practices.
+    #
+    # @param [Bolt::Outputter] An outputter instance.
+    # @return [Boolean]
+    #
+    def project_migrate(outputter)
```
artlawson, line 419: Not really sure what this is

lib/bolt/application.rb
artlawson: It doesn't seem to matter in the grand scheme, but for my 2 cents, I'd vote keeping it this way (wrapping just the executor). I'm sure users would adjust pretty easily and it seems like a good house-cleaning move.

lib/bolt/cli.rb
beechtom: I think this can be solved by just checking for multiple targeting options _before_ checking the specific command being run. Updated.

lib/bolt/outputter/human.rb
```diff
@@ -310,7 +310,7 @@ def format_table(results, padding_left = 0, padding_right = 3)
         )
       end
 
-      def print_tasks(tasks, modulepath)
+      def print_tasks(tasks:, modulepath:)
```
artlawson, line 313: What is the difference here? I presume it has something to do with default values?

artlawson (commented):
Got my first real read-through of a couple of the files changed. A few Ruby things I wasn't sure of, and a few Bolt things I wasn't sure of, but I am sure of all the hard work that is happening here :). Looking forward to checking it out more tomorrow!

lib/bolt/outputter/human.rb
beechtom: This was updated to use keyword arguments: https://medium.com/@tyler_brewer2/utilizing-ruby-keyword-arguments-462ba8366483

A few of the benefits are we now explicitly label what the arguments are, position doesn't matter, and we can easily add and remove new args in the future.

lib/bolt/application.rb
beechtom: The method itself? It implements `bolt project migrate`, which is a command that will update your Bolt project to be in line with what the current version of Bolt is expecting. More info is here: https://puppet.com/docs/bolt/latest/projects.html#migrate-a-bolt-project

lib/bolt/application.rb
beechtom: Hmm... in that case I think all the methods should be updated to verb-noun names instead of noun-verbs.

lib/bolt/application.rb
```diff
@@ -0,0 +1,619 @@
...
+        plans:      filter_content(pal.list_plans_with_cache(filter_content: true), filter),
+        modulepath: pal.user_modulepath
+      }
+    end
+
+    # Show available plugins.
+    #
+    # @return [Hash]
+    #
+    def plugin_show
```
lucywyman, line 402: The smallest nit, but should this be `list_plugins` for consistency?

lib/bolt/application.rb
```diff
@@ -0,0 +1,619 @@
...
+          msg = "Plan parameters include both 'nodes' and 'targets' with type 'TargetSpec', " \
+                "neither will populated with the value for --nodes or --targets."
+          Bolt::Logger.warn("nodes_targets_parameters", msg)
+        elsif node_param
+          params['nodes'] = targets.join(',')
+        elsif target_param
+          params['targets'] = targets.join(',')
+        end
+
+        inventory.get_targets(targets)
```
lucywyman, line 365: This seems like it doesn't get stored anywhere - and actually, we already [get targets inside of run_plan function itself](https://github.com/puppetlabs/bolt/blob/main/bolt-modules/boltlib/lib/puppet/functions/run_plan.rb#L114) that ends up getting called, so that string targets from plans also work. I don't think this is needed.unless it's for validation.

lib/bolt/application.rb
```diff
@@ -0,0 +1,619 @@
...
+    end
+
+    # Get a list of Bolt guides.
+    #
+    private def load_guides
+      root_path = File.expand_path(File.join(__dir__, '..', '..', 'guides'))
+      files     = Dir.children(root_path).sort
+
+      files.each_with_object({}) do |file, guides|
+        next if file !~ /\.txt\z/
```
lucywyman, line 581: Ah I think this needs to be updated based on the hack day changes.

lib/bolt/cli.rb
```diff
@@ -379,710 +341,584 @@ def validate(options)
...
-                           config.inventoryfile
-                         elsif File.exist?(config.default_inventoryfile)
-                           config.default_inventoryfile
-                         end
+    # Validates that only one targeting option is provided and that commands
+    # requiring a targeting option received one.
+    #
+    # @param options [Hash] The CLI options.
+    #
+    private def validate_targeting_options(options)
```
lucywyman, line 353: This seems like it should return immediately for subcommands that don't take any targets, like `guide` and `group`, mostly because it seems odd to error if it's passed invalid options that aren't used.

beechtom: That would require having [this](https://github.com/puppetlabs/bolt/blob/main/lib/bolt/cli.rb#L444-L446), which I'm trying to avoid. Ideally, I'd like to have the parser just error when a command is given an invalid option, which theoretically can be done automatically by comparing its [flags](https://github.com/puppetlabs/bolt/blob/main/lib/bolt/bolt_option_parser.rb#L60) with the keys in the `options` hash. That'd be a "next step" of the refactor, though.

lib/bolt/cli.rb
```diff
@@ -379,710 +341,584 @@ def validate(options)
...
+          log_outputter = Bolt::Outputter::Logger.new(options[:verbose], config.trace)
+
+          #
+          # FINALIZING SETUP
+          #
+
+          check_gem_install
+          warn_inventory_overrides_cli(config, options)
+          submit_screen_view(analytics, config, inventory, options)
+          process_target_list(plugins.puppetdb_client, @rerun, options)
```
lucywyman, line 456: Instead of this modifying the options in the function as a side effect, this is more clear if it's:
```
options[:target] = process_target_list(plugins.puppetdb_client, @rerun, options)
```

lib/bolt/cli.rb
```diff
@@ -379,710 +341,584 @@ def validate(options)
...
+            FileUtils.rm(config.project.plugin_cache_file)
+          end
 
-      case options[:action]
-      when 'show'
-        case options[:subcommand]
-        when 'task'
-          if options[:object]
-            show_task(options[:object])
+          # TODO: Include logic for handling special cases when subscribing the
```
lucywyman, line 465: I think can remove this todo?

lib/bolt/cli.rb
```diff
@@ -379,710 +341,584 @@ def validate(options)
...
-        when 'plan'
-          if options[:object]
-            show_plan(options[:object])
-          else
-            list_plans
+
+          executor.subscribe(log_outputter)
+
+          # TODO: Figure out where this should really go. It doesn't seem to
+          #       make sense in the application, since the params should already
```
lucywyman, line 485: Actually, since there's only 2 cases where we do this it seems like they could just go in the `task run` and `plan run` handlers. Similarly the `options[:argument]` for script seems unnecessary.

lib/bolt/cli.rb
```diff
@@ -379,710 +341,584 @@ def validate(options)
...
+        app.shutdown
+        outputter.print_apply_result(results)
+        results.ok? ? SUCCESS : FAILURE
 
-      if %w[task plan].include?(options[:subcommand]) && options[:task_options] && !options[:params_parsed] && pal
-        options[:task_options] = pal.parse_params(options[:subcommand], options[:object], options[:task_options])
-      end
+      when 'command'
+        outputter.print_head
+        results = outputter.spin do
```
lucywyman, line 536: Hmm, right now this is started and stopped by `with_node_logging` in the executor by publishing the start spin events. I think having just that might be better so that the spinner still spins when running plans, and then we shouldn't need to manager the spinner here (at least for executor functions).

lib/bolt/cli.rb
```diff
@@ -379,710 +341,584 @@ def validate(options)
...
+          app.command_run(options[:object], options[:targets], **options.slice(:env_vars))
+        end
+        rerun.update(results)
+        app.shutdown
+        outputter.print_summary(results, results.elapsed_time)
```
lucywyman, line 541: It seems a little funny now that we publish messages to the outputter from the executor, but then for results we *return* the results and pass them to the outputter. I think this is too much work to be in scope for this, but it'd be nice if `with_node_logging` published the results instead of returning them, then the outputter could just see the result and print it according to the type. That'd also be nice because then `app.shutdown` could just be in an `ensure` in this function instead of in every case statement.

lib/bolt/cli.rb
```diff
@@ -379,710 +341,584 @@ def validate(options)
...
+      options[:targets] = if options[:query]
+                            pdb_client.query_certnames(options[:query])
+                          elsif options[:rerun]
+                            rerun.get_targets(options[:rerun])
+                          end
     end
 
-    # Looks up a value with Hiera as if in a plan outside an apply block, using
-    # provided variable values for interpolations
+    # TODO: Discuss whether we still want / need to collect analytics for
```
lucywyman, line 722: I think so - it's useful when determining how urgent a bug is, or if we should continue to support bundled content (or bundle it).

lib/bolt/cli.rb
```diff
@@ -379,710 +341,584 @@ def validate(options)
...
-      end
+          # TODO: Subscribe this to the executor.
+          analytics = begin
+            client = Bolt::Analytics.build_client(config.analytics)
+            client.bundled_content = bundled_content(options)
+            client
+          end
 
-      analytics.screen_view(screen, **screen_view_fields)
+          # TODO: Configure logger with a separate method call?
```
lucywyman, line 419: I think can remove this todo

lib/bolt/cli.rb
beechtom: This seems similar to `query` in that it's a CLI implementation detail. From what I understand, `pal.parse_params` converts JSON values to the expected parameter type. Ideally, the application receives parameters as actual objects, not JSON, hence they should be parsed before reaching the application, i.e. params would always be passed as:

```ruby
application.run_plan(..., params: { 'foo' => [1, 2, 3] })
```

and not

```ruby
application.run_plan(..., params: { 'foo' => "[1, 2, 3]" })
```

spec/fixtures/modules/sample/plans/success.pp
```diff
@@ -0,0 +1,5 @@
+plan sample::successful_task (
```
lucywyman, line 1: This has the wrong name and seems like a duplicate of `successful_task` - remove?

spec/fixtures/modules/sample/tasks/error.ps1
```diff
@@ -0,0 +1,2 @@
+Write-Output 'error'
```
lucywyman, line 1: I think these are duplicates of `modules/error/tasks/fail.sh` and `modules/error/tasks/fail.ps1` - can those be reused for the CLI tests?

lib/bolt/cli.rb
beechtom: Hm, I like that idea. We can ticket and discuss another time.

lucywyman (commented):
### apply_manifest
- `bolt apply` no longer verifies the path it's passed isn't `nil`
  - [x] Errors if not provided a path or manifest code
  - [x] Raises a nice error if the path is `''`
- [x] Ensure the outputter and log_outputter are always subscribed
- [x] Otherwise the same codepaths as in `Bolt::CLI`
- [x] Ensure shutdown + print + update rerun all happen

### command_run
- [x] Same codepath as before

### file_download
- [x] Errors if no source is specified
- [x] Errors if no destination is specified
- [x] Same codepath

### file_upload
- [x] Errors if no source
- [x] Errors if no dest
- [ ] Can upload directories
- [x] Same codepath

### guide (list + show)
- [x] Same codepaths

### inventory show
- [x] Errors with multiple targetting options
- [x] Accepts `rerun` and `query`

### lookup
- [x] Validates no key to lookup
- [x] Errors with `-t` and `--plan-hierarchy`
- [x] Subscribes outputters
- [x] Spins

### module add
- [x] assert_project_file unchanged
- [x] spins

### plan run
- [x] Errors when target list is empty
- [x] Same codepaths

### script run
- [x] Uses same functions for loading and validation

### One liners
- [x] module generate-types unchanged
- [x] module show unchanged
- [x] plan new
- [x] plan convert unchanged
- [x] plan show unchanged
- [x] project init
- [x] run task
- [x] plugin show
- [x] project migrate
- [x] secret createkeys
- [x] secret encrypt
- [x] secret decrypt
- [x] task run
- [x] task show

## Functions unchanged
- [x] assert_project_file
- [x] filter_content
- [x] find_file
- [x] group_targets_by_source

## CLI class
- [x] handle_parser_errors() moved to OptionParser#permute
- [x] rename target_options to params
- [x] validate function adds checks for additional objects to many subcommands
- [x] Puppet and yaml plan counts inlined from analytics
- [x] Calls `outputter.print_head` for `command run`, `file upload`, `file download`, `script run`, `task run`
- [x] Calls `rerun.update` for `command run`, `file upload`, `file download`, `script run`, `task run`

### CLI functions unchanged
Validated all of these using `diff -w <(head -n <line> ./lib/bolt/cli.rb | tail -n <line>) <(head -n <line> ./old_cli.rb | tail -n <line>)`
- [x] validate_ps_version
- [x] check_gem_install
- [x] warn_inventory_overrides
- [x] Submitting analytics data (submit_screen_view)

lib/bolt/application.rb
```diff
@@ -0,0 +1,617 @@
...
+    end
+
+    # Show available guides.
+    #
+    # @param guides [Hash] A map of topics to paths to guides.
+    # @param outputter [Bolt::Outputter] An outputter instance.
+    # @return [Boolean]
+    #
+    def list_guides
+      { topics: load_guides.keys - ['guide'] }
```
lucywyman, line 153:
```suggestion
      { topics: load_guides.keys }
```

lib/bolt/cli.rb
```diff
@@ -379,725 +337,570 @@ def validate(options)
...
-        plan_vars = Hash[options[:leftovers].map { |a| a.split('=', 2) }]
-        # Validate functions verifies one of these was passed
-        if options[:targets]
-          code = lookup(options[:object], options[:targets], plan_vars: plan_vars)
-        elsif options[:plan_hierarchy]
-          code = plan_lookup(options[:object], plan_vars: plan_vars)
+        SUCCESS
+
+      when 'inventory'
+        targets = app.show_inventory(options[:targets])
```
lucywyman, line 562: `options[:targets]` will actually be `[]` here if no targetting options are provided, not `nil`, which means `bolt inventory show` is actually always empty...I think checking for `empty?` instead below makes sense, but for this maybe something like:
`options[:targets].empty? ? nil : options[:targets]`

Also vaguely concerning we don't have a test that caught this....

lucywyman (changes requested):

lib/bolt/cli.rb
lucywyman: Also verified this is the only place we assume that `options[:targets]` can be nil.

beechtom: Fixed and added an integration test for this. :)

spec/integration/application_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+      expect(output.keys).to match_array(%w[moduledir puppetfile success])
+      expect(output['moduledir']).to eq((@project.path + '.modules').to_s)
+      expect(output['puppetfile']).to eq((@project.path + 'Puppetfile').to_s)
+      expect(output['success']).to be
+    end
+  end
+
+  describe 'module install' do
+    it 'installs project modules' do
+    end
```
lucywyman, line 63: Woops 😅

lucywyman:
It's a little messy but I started [tracking test coverage in this gist](https://gist.github.com/lucywyman/a44c69c7bd1a962418d0dae7ecce3866) - running out of steam so I may come back to it tomorrow.

spec/integration/application_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+
+    it 'installs modules from a Puppetfile without resolving' do
+    end
```
artlawson, line 66: double whoops 🙃

spec/integration/application_spec.rb
beechtom: Whoops. I think this was an early iteration, and I meant to move them elsewhere and finish them. :P Moved the `module add|install` integration tests to `spec/integration/module_installer_spec.rb`.

lucywyman (approved):
Phew, ok! Never closer. Updated the testing gist - I can more nicely format it if folks are interested. There are a few tests that it would may be helpful to add back from `spec/bolt/cli_spec.rb`, but I would be fine with merging this and then adding them depending on how you're feeling @beechtom:
- Unit test for `validate_targeting_opts`
- File upload fails with `--env-var` I mostly think this is important so that users know that setting an env var (like `HOME`) won't get used for this command

🙌🏼

beechtom:
@lucywyman Added those tests and rebased off `main`.

Issue #2877, opened 2021-06-09
Let `wait()` with no futures argument wait for all un-joined background jobs


reidmv:
## Use Case

When writing plans using `background() || { ... }` in Puppet's immutable-variables DSL,  naming and keeping vars for background jobs is often onerous and cluttered. It is common to set up a series of background jobs and at a particular point in a plan, simply want to wait for ALL in-scope background jobs that have been initiated since the last call to  `wait()` to complete.

## Describe the Solution You Would Like

The `wait()` function should accept a call with no futures arguments, and be able to dynamically wait on and resolve futures for all in-scope background jobs not previously collected by a `wait()` call. E.g.

```
background() || { "one" }
background() || { "two" }

$three = "Something synchronous"

background() || { $three }

wait()
```

This aligns with the relative simplicity of shell scripting's `&` background and `wait` functions.

## Additional Context

With things as they are, the cleanest I've been able to hack together is using arrays to reduce variables collected for waiting, but it's still not as elegant as an argument-less call to `wait()` would be. E.g. today, I can do this:

```
$bg_jobs = [
  background() || { "one" },
  background() || { "two" },
]

wait($bg_jobs)
```

But it's not easy to do synchronous calculation in-between background calls, and the `$bg_jobs` variable(s) are still cluttered.

## Solution
- This should return the results from all futures for this plan, even futures that have finished, in order to be predictable by users and return results from everything
- The futures should only be from the calling plan, not from any subplans
  - This is going to require attaching a unique identifier (plan name? hash of the scope? what about recursive plans or circular plan calls?) to the Future object so they can be filtered to only wait for plans in the calling `wait`'s scope
- Return should be the same as a normal `wait()` function call with an array passed. The order of Results can be whatever the order the Futures are currently in, since there's no way to really make it predictable to the user

PR #2887, opened 2021-06-15
(GH-2877) `wait()` with no arguments waits on all Futures from the plan

lucywyman:
This amends the `wait()` plan function to wait on all the Futures that
were spawned from the same plan as the `wait()` call if the function is
not provided a list of futures. The function still accepts a `timeout`
or `options` parameter if it is not passed any futures. To support this,
Bolt's `PlanFuture` object now has a `plan_id` attribute, which
determines which **invocation** of a plan the Future was created from
(since the same plan can be invoked multiple times in parallel). When
`wait()` is called without Futures, it selects the Futures from the
FiberExecutor with a matching `plan_id` to itself, and waits on those.

    
This also removes a flaky test around whether the cache file is updated
when plans in modules are modified. The test doesn't seem to be failing
legitimately based on local testing.

- [ ] puppetlabs/orchestrator#1738
- [ ] puppetlabs/pe-bolt-modules#20

!feature

* **`wait()` without arguments waits on all Futures from plan** ([#2877](https://github.com/puppetlabs/bolt/issues/2877))

  Calling the `wait()` plan function without a list of Futures will now
  wait on all Futures created so far in the plan.

### DIFF
--- a/bolt-modules/boltlib/lib/puppet/functions/background.rb
+++ b/bolt-modules/boltlib/lib/puppet/functions/background.rb
@@ -31,7 +31,8 @@ Puppet::Functions.create_function(:background, Puppet::Functions::InternalFuncti
     executor = Puppet.lookup(:bolt_executor)
     executor.report_function_call(self.class.name)
 
-    executor.create_future(scope: scope, name: name) do |newscope|
+    plan_id = executor.get_current_plan_id(fiber: Fiber.current)
+    executor.create_future(scope: scope, name: name, plan_id: plan_id) do |newscope|
       # Catch 'return' calls inside the block
       result = catch(:return) do
         # Execute the block. Individual plan steps in the block will yield
--- a/bolt-modules/boltlib/lib/puppet/functions/parallelize.rb
+++ b/bolt-modules/boltlib/lib/puppet/functions/parallelize.rb
@@ -34,7 +34,11 @@ Puppet::Functions.create_function(:parallelize, Puppet::Functions::InternalFunct
     executor.report_function_call(self.class.name)
 
     futures = data.map do |object|
-      executor.create_future(scope: scope) do |newscope|
+      # We're going to immediately wait for these futures, *and* don't want
+      # their results to be returned as part of `wait()`, so use a 'dummy'
+      # value as the plan_id. This could also be nil, though in general we want
+      # to require Futures to have a plan stack so that they don't get lost.
+      executor.create_future(scope: scope, plan_id: 'parallel') do |newscope|
         # Catch 'return' calls inside the block
         result = catch(:return) do
           # Add the object to the block parameters
--- a/bolt-modules/boltlib/lib/puppet/functions/run_plan.rb
+++ b/bolt-modules/boltlib/lib/puppet/functions/run_plan.rb
@@ -125,6 +125,17 @@ Puppet::Functions.create_function(:run_plan, Puppet::Functions::InternalFunction
       params = wrap_sensitive_parameters(params, closure.parameters)
     end
 
+    # This can be anything as long as it's unique
+    plan_instance_id = SecureRandom.uuid
+
+    # Add the plan invocation ID to the plan_stack for the PlanFuture the plan is
+    # running in so that we know the PlanFuture is running in a new plan
+    # invocation. This can be nil in test cases and when `wait()` isn't
+    # supported.
+    current_future = executor.get_current_future(fiber: Fiber.current)
+    # Safe operator to make testing easier
+    current_future&.plan_stack&.unshift(plan_instance_id)
+
     # wrap plan execution in logging messages
     executor.log_plan(plan_name) do
       result = nil
@@ -150,6 +161,8 @@ Puppet::Functions.create_function(:run_plan, Puppet::Functions::InternalFunction
           raise e
         end
       ensure
+        # Pop the plan invocation ID off of the plan_id stack for the Future.
+        current_future&.plan_stack&.shift
         if run_as
           executor.run_as = old_run_as
         end
--- a/bolt-modules/boltlib/lib/puppet/functions/wait.rb
+++ b/bolt-modules/boltlib/lib/puppet/functions/wait.rb
@@ -1,14 +1,13 @@
 # frozen_string_literal: true
 
 require 'bolt/logger'
-require 'bolt/target'
 
 # Wait for a Future or array of Futures to finish and return results,
 # optionally with a timeout.
 #
 # > **Note:** Not available in apply block
 Puppet::Functions.create_function(:wait, Puppet::Functions::InternalFunction) do
-  # Wait for Future(s) to finish
+  # Wait for Futures to finish.
   # @param futures A Bolt Future object or array of Bolt Futures to wait on.
   # @param options A hash of additional options.
   # @option options [Boolean] _catch_errors Whether to catch raised errors.
@@ -27,7 +26,37 @@ Puppet::Functions.create_function(:wait, Puppet::Functions::InternalFunction) do
     return_type 'Array[Boltlib::PlanResult]'
   end
 
-  # Wait for Future(s) to finish with timeout
+  # Wait for all Futures in the current plan to finish.
+  # @param options A hash of additional options.
+  # @option options [Boolean] _catch_errors Whether to catch raised errors.
+  # @return A Result or Results from the Futures
+  # @example Perform multiple tasks in the background, then wait for all of them to finish
+  #   background() || { upload_file("./large_file", "/opt/jfrog/...", $targets) }
+  #   background() || { run_task("db::migrate", $targets) }
+  #   # Wait for all futures in the plan to finish and return all results
+  #   $results = wait()
+  dispatch :wait_for_all do
+    optional_param 'Hash[String[1], Any]', :options
+    return_type 'Array[Boltlib::PlanResult]'
+  end
+
+  # Wait for all Futures in the current plan to finish with a timeout.
+  # @param timeout How long to wait for Futures to finish before raising a Timeout error.
+  # @param options A hash of additional options.
+  # @option options [Boolean] _catch_errors Whether to catch raised errors.
+  # @return A Result or Results from the Futures
+  # @example Perform multiple tasks in the background, then wait for all of them to finish with a timeout
+  #   background() || { upload_file("./large_file", "/opt/jfrog/...", $targets) }
+  #   background() || { run_task("db::migrate", $targets) }
+  #   # Wait for all futures in the plan to finish and return all results
+  #   $results = wait(30)
+  dispatch :wait_for_all_with_timeout do
+    param 'Variant[Integer[0], Float[0.0]]', :timeout
+    optional_param 'Hash[String[1], Any]', :options
+    return_type 'Array[Boltlib::PlanResult]'
+  end
+
+  # Wait for Futures to finish with timeout.
   # @param futures A Bolt Future object or array of Bolt Futures to wait on.
   # @param timeout How long to wait for Futures to finish before raising a Timeout error.
   # @param options A hash of additional options.
@@ -58,14 +87,22 @@ Puppet::Functions.create_function(:wait, Puppet::Functions::InternalFunction) do
   end
 
   def wait(futures, options = {})
-    inner_wait(futures, nil, options)
+    inner_wait(futures: futures, options: options)
+  end
+
+  def wait_for_all(options = {})
+    inner_wait(options: options)
+  end
+
+  def wait_for_all_with_timeout(timeout, options = {})
+    inner_wait(timeout: timeout, options: options)
   end
 
   def wait_with_timeout(futures, timeout, options = {})
-    inner_wait(futures, timeout, options)
+    inner_wait(futures: futures, timeout: timeout, options: options)
   end
 
-  def inner_wait(futures, timeout = nil, options = {})
+  def inner_wait(futures: nil, timeout: nil, options: {})
     unless Puppet[:tasks]
       raise Puppet::ParseErrorWithIssue
         .from_issue_and_stack(Bolt::PAL::Issues::PLAN_OPERATION_NOT_SUPPORTED_WHEN_COMPILING, action: 'wait')
@@ -85,7 +122,10 @@ Puppet::Functions.create_function(:wait, Puppet::Functions::InternalFunction) do
     executor = Puppet.lookup(:bolt_executor)
     executor.report_function_call(self.class.name)
 
-    futures = Array(futures)
+    # If we get a single Future, make sure it's an array. If we didn't get any
+    # futures pass that on to wait so we can continue collecting any futures
+    # that are created while waiting on existing futures.
+    futures = Array(futures) unless futures.nil?
     executor.wait(futures, **valid)
   end
 end
--- a/documentation/experimental_features.md
+++ b/documentation/experimental_features.md
@@ -79,7 +79,7 @@ Finished on docker://puppet_6_node:
   Hello!
 ```
 
-As you can see, when you configure output to stream, Bolt might print to the console twice:
+As you can see, when you configure output to stream, Bolt may print to the console twice:
 once as the actions are running, and again after Bolt prints the results. You can prevent
 Bolt from printing the results once the action has completed by specifying the `--no-verbose`
 command-line option.
@@ -190,7 +190,6 @@ of plan steps to execute in the background while other parts of the plan execute
 
 These two function invocations are equivalent:
 ```
-# 
 parallelize(['./file1', './file2']) |$file| {
   file_upload($file, '/home/user/', $targets)
   ...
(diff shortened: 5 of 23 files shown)


lucywyman:
Still manually testing, but I think this is ready for review. I'll move it back to draft if I find any issues.

### Test cases
- [x] Only waits on futures from the calling plan, not subplans or parent plans
- [x] Accepts just a timeout with no futures
- [x] Accepts a timeout and options
- [x] Accepts just options
- [x] Doesn't wait for itself when called inside `background()` block
- [ ] Waits on Futures created inside a `background()` *after* the `wait()` has been called
- [ ] Only waits on futures from a particular invocation
  - [ ] If the same subplan has been invoked multiple times from a plan
  - [ ] If the plan is invoked recursively
  - [ ] Waits on futures created inside a new scope, like `each`
  - [ ] Does not wait on futures created in a subplan
  - [ ] Does not wait on futures created in a subplan run from a new scope

### BoltSpec
- [ ] Accepts just a timeout with no futures
- [ ] Accepts a timeout and options
- [ ] Accepts just options

lucywyman:
### Tests
- [x] Only waits on futures from the calling plan, not subplans or parent plans
- [x] Accepts just a timeout with no futures
- [x] Accepts a timeout and options
- [x] Accepts just options
- [x] Doesn't wait for itself when called inside `background()` block
- [x] Does not wait on Futures created inside a `background()` in the calling plan
- [x] Only waits on Futures from a particular plan invocation
  - [x] If a subplan is called multiple times which creates Futures, each subplan will only wait for it's own Futures
  - [x] If the plan is invoked recursively
  - [x] Waits on futures created inside a new scope, like `each`
  - [x] Does not wait on futures created in a subplan
  - [x] Does not wait on futures created in a subplan run from a new scope

### BoltSpec
- [ ] Accepts just a timeout with no futures
- [ ] Accepts a timeout and options
- [ ] Accepts just options

bolt-modules/boltlib/lib/puppet/functions/background.rb
```diff
@@ -31,7 +33,10 @@ def background(scope, name = nil, &block)
     executor = Puppet.lookup(:bolt_executor)
     executor.report_function_call(self.class.name)
 
-    executor.create_future(scope: scope, name: name) do |newscope|
+    # Use the Ruby object ID of the nested scopes to identify an individual
+    # plan invocation. This has to include
```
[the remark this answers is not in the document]
lucywyman:
```
    # Use the Ruby object ID of the nested scopes to identify an individual
    # plan invocation. Puppet scopes are a stack (not hierarchy), where there's a
    # top-level scope object with several 'layers' of scope underneath it. The
    # ephemeral instance variable contains the stack. The plan invocation ID
    # needs to include the object IDs for the entire scope stack because the
    # top-scope ID is persistent across new plan invocations, and additionally
    # new scopes are created during the course of a plan invocation. By
    # ensuring the scope stack for this plan invocation matches at least the
    # beginning of the scope stack in another plan function call, we can assert
    # that those functions belong in the same plan invocation.
```
I think I just didn't want to write it all out if this wasn't the solution we ended up using :P

bolt-modules/boltlib/lib/puppet/functions/parallelize.rb
```diff
@@ -34,7 +34,8 @@ def parallelize(scope, data, &block)
     executor.report_function_call(self.class.name)
 
     futures = data.map do |object|
-      executor.create_future(scope: scope) do |newscope|
+      plan_id = scope.instance_variable_get(:@ephemeral).last.object_id
```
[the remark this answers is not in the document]
lucywyman: I think I just forgot to update it. It seems like using just the "closest" scope and making sure that's in the stack for the future would work...

lib/bolt/fiber_executor.rb
```diff
@@ -101,7 +105,16 @@ def round_robin
...
-      plan_futures.empty?
+      active_futures.empty?
+    end
+
+    # Get the Future objects associated with a particular plan.
+    #
+    def get_futures_for_plan(plan_id:)
+      plan_futures = active_futures + finished_futures
+      plan_futures.select do |future|
+        future.plan_id&.slice(0, plan_id.length) == plan_id
```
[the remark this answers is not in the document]
lucywyman: No, the scopes for the `background()` call and `wait()` call can have different lengths if one is called inside an `each` block for example, or another construct that's part of the same plan invocation but has a different scope stack. That can go both directions though, and I think I only tested this with `background()` being called in an `each` block and `wait` being at the top level. I think this needs more testing and I suspect there may be issues....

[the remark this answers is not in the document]
beechtom: D'oh. Definitely see now that `future.plan_id` and `plan_id` are different. 🤦

documentation/experimental_features.md
```diff
@@ -219,13 +219,18 @@ Futures will continue to execute in parallel. However, Bolt itself will not exit
...
-instead of raising them by passing `_catch_errors => true` to `wait()`. The `wait()`
-function returns the results from the Future blocks once they've all finished. If a
-Future errors, Bolt only raises the error after all other Futures finish executing and
-return to `wait()`.
+The `wait()` plan function accepts either a single Future object, an array of Futures, or no
+Futures. If not provided any Futures, the function will implicitly wait for all Futures created up
+to that point in the calling plan. Any Futures created in subplans or inside of `background()`
+blocks will not be waited on, even if the subplan is called inside a `background()` block. Users
+must call `wait()` in the subplan or `background()` block directly, or return Future objects to the
+calling plan to be explicitly passed to `wait()`.
```
hestonhoffman, line 227:
```suggestion
The `wait()` plan function accepts either a single Future object, an array of Futures, or no
Futures. If not provided any Futures, the function implicitly waits for all Futures created up
to that point in the calling plan. The `wait()` function does not wait on any Futures created
in subplans or inside of `background()` blocks, even if the subplan is called inside a 
`background()` block. If you want a plan to wait on a Future object, you must call `wait()` 
in the subplan or `background()` block directly, or return Future objects to the calling plan 
to be explicitly passed to `wait()`.
```

documentation/experimental_features.md
```diff
@@ -219,13 +219,18 @@ Futures will continue to execute in parallel. However, Bolt itself will not exit
...
+
+The function blocks until the Futures finish, with an optional timeout. If provided a timeout, any
+unfinished Futures will raise a timeout error if they have not completed within the timeout. You can
+return errors instead of raising them by passing `_catch_errors => true` to `wait()`. The `wait()`
+function returns the results from the Future blocks once they've all finished. If a Future errors,
+Bolt only raises the error after all other Futures finish executing and return to `wait()`.
```
hestonhoffman, line 233:
```suggestion
The function blocks until the Futures finish, with an optional timeout. If provided a timeout,
unfinished Futures raise a timeout error if they have not completed within the timeout. You can
return errors instead of raising them by passing `_catch_errors => true` to `wait()`. The `wait()`
function returns the results from the Future blocks once they've all finished. If a Future errors,
Bolt only raises the error after all other Futures finish executing and return to `wait()`.
```

bolt-modules/boltlib/lib/puppet/functions/background.rb
```diff
@@ -16,22 +16,21 @@
...
       raise Puppet::ParseErrorWithIssue
         .from_issue_and_stack(Bolt::PAL::Issues::PLAN_OPERATION_NOT_SUPPORTED_WHEN_COMPILING, action: 'background')
     end
 
     executor = Puppet.lookup(:bolt_executor)
     executor.report_function_call(self.class.name)
-
-    executor.create_future(scope: scope, name: name) do |newscope|
+    plan_id = Puppet.lookup(:plan_stack).last
+    executor.create_future(scope: closure_scope, name: name, plan_id: plan_id) do |newscope|
```
artlawson, line 33: What is closure scope?

bolt-modules/boltlib/lib/puppet/functions/wait.rb
```diff
@@ -58,22 +87,30 @@
...
+    inner_wait(timeout: timeout, options: options)
   end
 
   def wait_with_timeout(futures, timeout, options = {})
-    inner_wait(futures, timeout, options)
+    inner_wait(futures: futures, timeout: timeout, options: options)
   end
 
-  def inner_wait(futures, timeout = nil, options = {})
+  def inner_wait(futures: nil, timeout: nil, options: {})
```
artlawson, line 105: What's the difference between using `=` or `:`  here? It seems that they both set default values for the arguments.

lucywyman: Yeah `:` makes these [keyword arguments](https://thoughtbot.com/upcase/videos/ruby-keyword-arguments) or "named parameters" instead of just optional parameters. The main difference is when you call the functions.

With optional parameters (`=`), the parameter is optional (has a default value) but when you call the function you still have to know what parameters the function takes and position the arguments correctly to call the function. This makes calling the function confusing if you don't already know the function definition.
```
def function(breakfast = 'oatmeal', lunch = 'a sandwich')
  puts "Today I had #{breakfast} for breakfast and #{lunch} for lunch"
end

function() # Uses the defaults
function('waffles') # Sets 'waffles' as breakfast
function('oatmeal', 'tacos') # We have to manually pass in the first default so that Ruby knows we're setting *lunch*, not breakfast. Plus, you have to infer or Just Know that you had oatmeal for breakfast and tacos for lunch, and not the other way around.
```

Whereas with named arguments the function calls are self documenting so we know what parameters are being passed, and we can pass them in any order (and even do fancy things like [make a hash of arguments and pass them all at once](https://www.justinweiss.com/articles/fun-with-keyword-arguments/).
```
def function(breakfast: 'oatmeal', lunch: 'a sandwich')
  puts "Today I had #{breakfast} for breakfast and #{lunch} for lunch"
end

function() # Uses the defaults
function(breakfast: 'waffles') # Yum
function(lunch: 'tacos') # Such readability, so wow
```

Honestly, this wasn't something I really thought about until recently so I've just been updating our methods as I make changes. We can't just do it willy-nilly since some other tools use our functions directly or override them, but here it's safe to rewrite the definition.

lib/bolt/fiber_executor.rb
```diff
@@ -78,19 +82,19 @@ def round_robin
...
         # If the future errored and the main plan has already exited, log the
         # error at warn level.
-        unless plan_futures.map(&:id).include?(0) || future.state == "done"
+        unless active_futures.map(&:id).include?(0) || future.state == "done"
           Bolt::Logger.warn('errored_futures', "Error in future '#{future.name}': #{future.value}")
         end
 
         # Remove the PlanFuture from the FiberExecutor.
-        plan_futures.delete(future)
+        finished_futures.push(active_futures.delete(future))
```
artlawson, line 90: A little confused. What is being returned by `active_futures.delete(future)` ? Is this a more concise way of deleting the future from `active_futures` in one line and adding the same future to `finished_futures` in another?

bolt-modules/boltlib/lib/puppet/functions/background.rb
lucywyman: It's a special variable that gets [injected into the function by Puppet when creating a custom function in Ruby](https://puppet.com/docs/puppet/6/functions_ruby_implementation.html#functions_ruby_implementation-accessing-puppet-vars) that represents the Puppet scope object. It's generally useful for accessing variables from the manifest or Bolt plan. Here we were trying to use it to give a unique ID to each Bolt plan invocation, since in theory the scope should be different for each invocation and it has been tricky to distinguish each plan *invocation* when the plan *name* can be the same (ie. calling the same plan multiple times). But this ended up not working because the scope can change based on any number of things (a new block, like `each {}`, etc.) and it was impossible to determine from the scope if function calls were in the same plan invocation or not.

lib/bolt/fiber_executor.rb
lucywyman: Yeah, here `active_futures` is an Array, and [Ruby's Array#delete method](https://ruby-doc.org/core-2.6.2/Array.html#method-i-delete) removes the element(s) from the Array and returns the element.

documentation/experimental_features.md
```diff
@@ -219,13 +219,19 @@ Futures will continue to execute in parallel. However, Bolt itself will not exit
...
-function returns the results from the Future blocks once they've all finished. If a
-Future errors, Bolt only raises the error after all other Futures finish executing and
-return to `wait()`.
+The `wait()` plan function accepts either a single Future object, an array of Futures, or no
+Futures. If not provided any Futures, the function implicitly waits for all Futures created up
+to that point in the calling plan. The `wait()` function does not wait on any Futures created
+in subplans or inside of `background()` blocks, even if the subplan is called inside a 
+`background()` block. If you want a plan to wait on a Future object, you must call `wait()` 
+in the subplan or `background()` block directly, or return Future objects to the calling plan 
+to be explicitly passed to `wait()`.
```
beechtom, line 228: I think an example describing how to properly wait on futures from sub-plans would be helpful.

bolt-modules/boltlib/lib/puppet/functions/run_plan.rb
```diff
@@ -150,6 +162,8 @@ def run_inner_plan(scope, plan_name, targets, args = {})
           raise e
         end
       ensure
+        # Pop the plan invocation ID off of the plan_id stack for the Future.
+        current_future&.plan_stack&.shift
```
beechtom, line 166: This doesn't really matter since the effect is the same, but why not use `#pop` to pop elements off the stack? :P Took a second to realize the `plan_stack` is backwards from what I expected when examining it.

lib/bolt/fiber_executor.rb
```diff
@@ -101,12 +105,54 @@ def round_robin
...
+    end
+
+    def all_futures
+      active_futures + finished_futures
+    end
+
+    # Get the PlanFuture object that is currently executing
+    #
+    def get_current_future(fiber:)
+      # If no future is found, then this can just be nil.
```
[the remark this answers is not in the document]
lucywyman: Hm, I can delete the comment. I was worried about matching with multiple Futures or no Futures (so much that initially I had a check for those cases as part of this function). That's true right now in Bolt, but if this gets used in other contexts (ie. PE) it might not always be true. I think matching multiple Futures is actually a much bigger issue if it ever does happen than matching no Futures, hence the comment, but....we can deal with that if it ever actually becomes an issue which seems unlikely.

bolt-modules/boltlib/lib/puppet/functions/run_plan.rb
lucywyman: Honestly only did it this way because I spent *so long* looking at the [Pops::PuppetStack](https://github.com/puppetlabs/puppet/blob/main/lib/puppet/pops/puppet_stack.rb) class + stack while trying to make the scope work for the Plan ID. I can change it if it's more readable, I think it was just on the brain.

documentation/experimental_features.md
lucywyman: Ended up including a different (simpler) example, especially since I think now that blocks are less complicated (since we're not using the scope to determine wait ID) the futures from subplans bit is less confusing.

bolt-modules/boltlib/lib/puppet/functions/run_plan.rb
beechtom: Nah, I don't think it really matters. 🤷

beechtom:
Just want to confirm if this is expected behavior, since the criteria above states that Futures created inside a `background()` are not waited on from the calling plan.

Given this plan:

```puppet
plan bolt (
  TargetSpec $targets = 'localhost'
) {
  background('outer') || {
    run_command('whoami', $targets)

    background('inner') || {
      run_command('echo hi', $targets)
    }
  }

  $results = wait()
  out::message("Waited on ${results.length} futures")
  return $results
}
```

The output will be:

```shell
[
  [
    {
      "target": "localhost",
      "action": "command",
      "object": "echo bye",
      "status": "success",
      "value": {
        "stdout": "bye\n",
        "stderr": "",
        "merged_output": "bye\n",
        "exit_code": 0
      }
    },
    {
      "target": "local://127.0.0.1",
      "action": "command",
      "object": "echo bye",
      "status": "success",
      "value": {
        "stdout": "bye\n",
        "stderr": "",
        "merged_output": "bye\n",
        "exit_code": 0
      }
    }
  ],
  "Future 'inner'"
]
```

Both Futures are waited on, and `wait()` returns the value for each. In this case the first value is the result of the inner `background()` and the second value is the Future for the inner `background()`. Logically this seems like correct behavior?

lucywyman:
That is expected, I just didn't update the comment. The only thing that matters is that whether the inner backgrounds run or not is *predictable* and not just based on whether they've been started or not. Now that we're using an actual Plan ID to identify futures, inner Futures will be included in that if they've already started so we *need* to include all futures even ones that start after calling `wait()`. This is explicitly accomplished by recursively calling `wait()` and rechecking for new Futures until all the Futures have finished.

Updated the comment above and the docs.

bolt-modules/boltlib/lib/puppet/functions/run_plan.rb
```diff
@@ -57,6 +57,8 @@ def run_plan(scope, plan_name, args = {})
   end
 
   def run_inner_plan(scope, plan_name, targets, args = {})
+    # TODO: This is the only way that anyone can run a Bolt plan from a plan, right??
+
```
beechtom, line 61:
```suggestion
```

As far as I'm aware, yes.

documentation/experimental_features.md
```diff
@@ -219,20 +218,55 @@ Futures will continue to execute in parallel. However, Bolt itself will not exit
...
-Future errors, Bolt only raises the error after all other Futures finish executing and
-return to `wait()`.
+The `wait()` plan function accepts either a single Future object, an array of Futures, or no
+Futures. If not provided any Futures, the function implicitly waits for all Futures created up
+to that point in the calling plan. The `wait()` function does not wait on any Futures created
+in subplans, even if the subplan is called inside a `background()` block. If you want a plan to wait
+on a Future object created in a subplan then you must call `wait()` in the subplan directly, or
+return Future objects to the calling plan to be explicitly passed to `wait()`.
+
+For example, the following plan will begin running the code block with three `run_command()` calls on
```
beechtom, line 228:
```suggestion
For example, the following plan runs the code block with three `run_command()` calls on
```

documentation/experimental_features.md
```diff
@@ -219,20 +218,55 @@ Futures will continue to execute in parallel. However, Bolt itself will not exit
...
+the provided targets in the background, then executes a task in parallel. Once the task has
+finished, the commands may still be running in the background. Another task is kicked off in the
```
beechtom, line 230:
```suggestion
finished, the commands might still be running in the background. Another task is kicked off in the
```

documentation/experimental_features.md
```diff
@@ -219,20 +218,55 @@ Futures will continue to execute in parallel. However, Bolt itself will not exit
...
+background. Then the backgrounded commands and tasks are waited on before continuing.
+
+```
+# $targets = target1,target2,target3
+plan myplan(TargetSpec $targets) {
+  $binary_future = background("Run mycoolbinary") || {
+    # These commands will still execute sequentially per target. So Bolt connects to the targets in
```
beechtom, line 237:
```suggestion
    # These commands execute sequentially per target. Bolt connects to the targets in 
```

documentation/experimental_features.md
```diff
@@ -219,20 +218,55 @@ Futures will continue to execute in parallel. However, Bolt itself will not exit
...
+  # This task runs while the commands above are running
+  run_task("mytask", $targets)
+  
+  # Once the task 'mytask' has finished, start more commands in the background
+  $status_future = background("Check service status") || {
+    $r = run_command("systemctl status myservice", $targets)
+    out::message($r)
+  }
+
+  wait([$binary_future, $status_future]) # This is equivalent to wait() in this plan
```
beechtom, line 253:
```suggestion
  # This is equivalent to wait() in this plan
  wait([$binary_future, $status_future])
```

bolt-modules/boltlib/lib/puppet/functions/run_plan.rb
```diff
@@ -125,6 +127,16 @@ def run_inner_plan(scope, plan_name, targets, args = {})
...
 
+    # This can be anything as long as it's unique
+    plan_instance_id = SecureRandom.uuid
+
+    # Add the plan invocation ID to the plan_stack for the PlanFuture the plan is
+    # running in so that we know the PlanFuture is running in a new plan
+    # invocation. This can be nil in test cases and when `wait()` isn't
+    # supported.
+    current_future = executor.get_current_future(fiber: Fiber.current)
+    current_future&.plan_stack&.unshift(plan_instance_id)
```
beechtom, line 138: I think we can safely remove the safe nav operators?

bolt-modules/boltlib/lib/puppet/functions/run_plan.rb
beechtom, line 166: Same here?

beechtom:
Overall this is looking good and seems to be working as intended. Still planning on doing a bit more testing, but I'm feeling comfortable with where this is at, especially since parallelism in Bolt is still experimental and existing behavior doesn't break.

lucywyman:
Definitely no rush - I'll also do a little more manual testing later today.

bolt-modules/boltlib/lib/puppet/functions/run_plan.rb
beechtom: @lucywyman Heh, apparently not.

lucywyman: I mean, I'm just too lazy to update all the tests :P

beechtom (approved):

Issue #2863, opened 2021-06-02
Stacktrace when command or script errors from a plan


lucywyman:
## Describe the Bug
With the following plan:
```
plan bolt(
  TargetSpec $targets = 'localhost'
) {
  run_command('whoami', $targets, '_run_as' => 'root')
}
```
Bolt raises a stack trace when `merged_output` is nil due to the command or script having raised an error (ie. '  Sudo password for user lucy was not provided for localhost'):
```
lucy ~/githubs/bolt (main) $ bolt plan run bolt -t localhost 
Starting: plan bolt
Starting: command 'whoami' on localhost
Finished: command 'whoami' with 1 failure in 0.01 sec
Finished: plan bolt in 0.01 sec
Failed on localhost:
  Sudo password for user lucy was not provided for localhost
Traceback (most recent call last):
	11: from /opt/puppetlabs/bolt/bin/bolt:23:in `<main>'
	10: from /opt/puppetlabs/bolt/bin/bolt:23:in `load'
	 9: from /opt/puppetlabs/bolt/lib/ruby/gems/2.7.0/gems/bolt-3.9.0/exe/bolt:11:in `<top (required)>'
	 8: from /opt/puppetlabs/bolt/lib/ruby/gems/2.7.0/gems/bolt-3.9.0/lib/bolt/cli.rb:532:in `execute'
	 7: from /opt/puppetlabs/bolt/lib/ruby/gems/2.7.0/gems/bolt-3.9.0/lib/bolt/cli.rb:778:in `run_plan'
	 6: from /opt/puppetlabs/bolt/lib/ruby/gems/2.7.0/gems/bolt-3.9.0/lib/bolt/outputter/human.rb:614:in `print_plan_result'
	 5: from /opt/puppetlabs/bolt/lib/ruby/gems/2.7.0/gems/bolt-3.9.0/lib/bolt/outputter/human.rb:627:in `print_result_set'
	 4: from /opt/puppetlabs/bolt/lib/ruby/gems/2.7.0/gems/bolt-3.9.0/lib/bolt/result_set.rb:41:in `each'
	 3: from /opt/puppetlabs/bolt/lib/ruby/gems/2.7.0/gems/bolt-3.9.0/lib/bolt/result_set.rb:41:in `each'
	 2: from /opt/puppetlabs/bolt/lib/ruby/gems/2.7.0/gems/bolt-3.9.0/lib/bolt/result_set.rb:41:in `block in each'
[...]

PR #2890, opened 2021-06-16
(GH-2863) Stacktrace when plan throws error

artlawson:
This resolves a bug where previously, Bolt raised a stack trace when merged_output is nil due to the command or script having raised an error.

!bug

* Bolt no longer raises stack trace with plan error (#2863)

### DIFF
--- a/lib/bolt/outputter/human.rb
+++ b/lib/bolt/outputter/human.rb
@@ -174,11 +174,22 @@ module Bolt
           if result.message?
             @stream.puts(remove_trail(indent(2, result.message)))
           end
-
           case result.action
           when 'command', 'script'
             safe_value = result.safe_value
-            @stream.puts(indent(2, safe_value['merged_output'])) unless safe_value['merged_output'].strip.empty?
+            if safe_value["merged_output"]
+              @stream.puts(indent(2, safe_value['merged_output'])) unless safe_value['merged_output'].strip.empty?
+
+            else # output stdout or stderr
+              unless safe_value['stdout'].nil? || safe_value['stdout'].strip.empty?
+                @stream.puts(indent(2, "STDOUT:"))
+                @stream.puts(indent(4, safe_value['stdout']))
+              end
+              unless safe_value['stderr'].nil? || safe_value['stderr'].strip.empty?
+                @stream.puts(indent(2, "STDERR:"))
+                @stream.puts(indent(4, safe_value['stderr']))
+              end
+            end
           when 'lookup'
             @stream.puts(indent(2, result['value']))
           else
--- a/spec/bolt/outputter/human_spec.rb
+++ b/spec/bolt/outputter/human_spec.rb
@@ -255,6 +255,19 @@ describe "Bolt::Outputter::Human" do
     expect(output.string).to match(/Finished on #{target}.*value/m)
   end
 
+  it "doesn't stacktrace when merged_output is nil" do
+    value = {
+      'stdout'        => 'stdout',
+      'stderr'        => 'stderr',
+      'merged_output' => nil,
+      'exit_code'     => 2
+    }
+    expect {
+      outputter.print_result(Bolt::Result.for_command(target, value, 'command', "executed", []))
+    }.not_to raise_error
+    expect(output.string).to match(/stdout.*stderr/m)
+  end
+
   it "prints empty results from a plan" do
     outputter.print_plan_result(Bolt::PlanResult.new([], 'success'))
     expect(output.string).to eq("[\n\n]\n")


spec/bolt/outputter/human_spec.rb
```diff
@@ -255,6 +255,39 @@
...
+  it "matches standard error output" do
+    value = {
+      'stdout'        => nil,
+      'stderr'        => 'stderr',
+      'merged_output' => "stdout\nstderr",
+      'exit_code'     => 2
+    }
+    outputter.print_result(Bolt::Result.for_command(target, value, 'command', "executed", []))
+    expect(output.string).to match(/stderr/m)
+  end
```
beechtom, line 289: Because `merged_output` is set in both of these tests, the outputter will print what's in `merged_output` and not what's in `stdout`/`stderr`. While these tests both pass, they're not quite testing what we want: that the outputter falls back to printing `stdout`/`stderr` when `merged_output` is `nil`. The test on line 258 already prints a result like this, so we can either update that test to also check the output string or write a separate test that checks the output string.

lib/bolt/outputter/human.rb
```diff
@@ -174,11 +174,22 @@ def print_result(result)
...
-
           case result.action
           when 'command', 'script'
             safe_value = result.safe_value
-            @stream.puts(indent(2, safe_value['merged_output'])) unless safe_value['merged_output'].strip.empty?
+            if safe_value["merged_output"]
+              @stream.puts(indent(2, safe_value['merged_output'])) unless safe_value['merged_output'].strip.empty?
+
+            else # output stdout or stderr
+              unless safe_value['stdout'].strip.empty?
```
lucywyman, line 184:
```suggestion
              unless  safe_value['stdout']&.strip.empty?
```
This is the [Ruby safe navigation operator](https://mitrev.net/ruby/2015/11/13/the-operator-in-ruby/) which will just ensure that calling `strip()` doesn't raise an error if `safe_value['stdout']` returns nil.

lib/bolt/outputter/human.rb
```diff
@@ -174,11 +174,22 @@ def print_result(result)
...
+                @stream.puts(indent(2, "STDOUT:"))
+                @stream.puts(indent(4, safe_value['stdout']))
+              end
+              unless safe_value['stderr'].strip.empty?
```
lucywyman, line 188:
```suggestion
              unless safe_value['stderr']&.strip.empty?
```

spec/bolt/outputter/human_spec.rb
```diff
@@ -255,6 +255,18 @@
...
 
+  it "doesn't stacktrace when merged_output is nil" do
+    value = {
+      'stdout'        => 'stdout',
+      'stderr'        => 'stderr',
+      'merged_output' => nil,
+      'exit_code'     => 2
+    }
+    outputter.print_result(Bolt::Result.for_command(target, value, 'command', "executed", []))
+    expect { output.string }.not_to raise_error
```
beechtom, line 266: We want to wrap the `expect` block around the part of the code that we want to ensure does not error, which is the line above (`outputter.print_result`).

lib/bolt/outputter/human.rb
artlawson: Got it!

beechtom: I think we actually want this to be `unless safe_value['stdout'].nil? || safe_value['stdout'].strip.empty?`, otherwise a `nil` value will still print `STDOUT:`. Same for `stderr`.

artlawson: That makes sense. Just put it in updated commit

lib/bolt/outputter/human.rb
```diff
@@ -174,11 +174,22 @@ def print_result(result)
...
-
           case result.action
           when 'command', 'script'
             safe_value = result.safe_value
-            @stream.puts(indent(2, safe_value['merged_output'])) unless safe_value['merged_output'].strip.empty?
+            if safe_value["merged_output"]
+              @stream.puts(indent(2, safe_value['merged_output'])) unless safe_value['merged_output'].strip.empty?
+
+            else # output stdout or stderr
+              unless safe_value['stdout'].nil? || safe_value['stdout']&.strip&.empty?
```
beechtom, line 184:
```suggestion
              unless safe_value['stdout'].nil? || safe_value['stdout'].strip.empty?
```

If this is a non-nil value, it'll be a string. No need for safe operators then.

lib/bolt/outputter/human.rb
```diff
@@ -174,11 +174,22 @@ def print_result(result)
...
+                @stream.puts(indent(2, "STDOUT:"))
+                @stream.puts(indent(4, safe_value['stdout']))
+              end
+              unless safe_value['stderr'].nil? || safe_value['stderr']&.strip&.empty?
```
beechtom, line 188:
```suggestion
              unless safe_value['stderr'].nil? || safe_value['stderr'].strip.empty?
```

Ditto.

artlawson: Is that because the nil check means we don't need the safe operators anymore?

lib/bolt/outputter/human.rb
artlawson: I read them backwards, you can ignore my question. Thanks for clarification!

spec/bolt/outputter/human_spec.rb
beechtom: @artlawson I think this got missed.

beechtom (approved):

Issue #2886, opened 2021-06-15
Update `bolt guide` human output


beechtom:
## Use Case
We've been updating the output of all `show` commands to have more consistent output. The last command that needs updating is `bolt guide`.

Current output:
```
Available topics are:
inventory
links
logging
module
modulepath
project
targets
transports

Use 'bolt guide <TOPIC>' to view a specific guide.
```

## Describe the Solution You Would Like
The human output should be updated to have colored headers, an indented topic list, and an 'Additional information' section detailing how to view a specific guide. Output should look similar to this:

```
Available topics
  inventory
  links
  logging
  module
  modulepath
  project
  targets
  transports

Additional information
  Use 'bolt guide <TOPIC>' to view a specific guide.
```

PR #2902, opened 2021-06-24
(GH-2886) Update `bolt guide` human output

beechtom:
This updates the human output for the `bolt guide` command to use a
similar format to the `show` commands.

!no-release-note

### DIFF
--- a/lib/bolt/outputter/human.rb
+++ b/lib/bolt/outputter/human.rb
@@ -462,9 +462,11 @@ module Bolt
       end
 
       def print_topics(topics)
-        print_message("Available topics are:")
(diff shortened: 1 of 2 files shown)


lucywyman (approved):
- [x] Output is reasonable when passed a guide that's not available (`bolt guide foo`)
- [x] Output with the human outputter is formatted

FWIW it looks like this method, in particular, is overridden in the rainbow outputter so that outputs a different format and color scheme, while commands like `bolt task show` output the same thing with the rainbow outputter as the human outputter. I don't think it's worth trying to align the outputs or anything, but if anyone brings it up it's that we didn't think about it just that it wasn't worth the time.

Issue #2889, opened 2021-06-16
Parallelize causes plan variable to be 'forgotten'


benridley:
## Describe the Bug
In some cases, using parallelize in a plan causes variables declared previously in that plan to be 'forgotten', and attempting to use them results in an error of the form "Evaluation Error: Unknown variable: '\<variable\>.'"

## Expected Behavior
Parallelize has no effect on variables declared in a higher scope.

## Steps to Reproduce
I am having trouble reproducing this easily, however I have managed to reproduce it reliably with a complex set of plans and tasks. Some of these rely on internal network services, so it's difficult for me to disclose details. 
 
In this case, I'm using parallelize to shutdown a number of clustered applications at once. I've narrowed the culprit down to a single Puppet function, which is a custom function that calls a task on localhost to lookup secrets from an external system. When this function is called in specific circumstances, it appears to *sometimes* cause this bug, but not all the time. This function is called at multiple places in the parallel blocks and they all complete successfully. Only after completing and exiting the block is this problem occurring. 

Accessing the variable inside the parallelize blocks, even as the final step of the block, works successfully in all cases. It's only once we exit the parallelize function that the variable seems to disappear. To test this I added an `out:message` to simply print the variable before and after executing a plan inside parallelize. This always prints the expected value. Immediately after exiting the `parallelize` function, trying to use the variable causes the above error. 

## Environment
 - Version: 3.9.2
 - Platform: Red Hat Enterprise Linux 8
 
## Additional Context
I initially suspected a child plan might be overwriting a top level plan or something like that. I've tried renaming the variable to a unique name which didn't fix the problem. When I use the `each` function to iterate instead of `parallelize`, the problem does not occur.

lucywyman:
Hm, is there any chance you can provide the custom function you're using that's causing the issue? Or some pseudocode for the plan you have that's forgetting variables?

benridley:
Thanks @lucywyman, I'll do my best to provide details.

```puppet
function resolve_secret(
  Integer $secret_id
){
  run_task('get_secret', localhost, {'secret_id' => $secret_id}.first.sensitive
}
```

The task `get_secret` is just a Python script to call out to the service that stores our secrets. It looks like this (simplified):
```python
import json
import requests

params = json.load(stdin)
with open('credentials.json', 'r') as f:
  creds = json.load(f)

sid = params['secret_id']
resp = requests.get(f"{secret_endpoint}/api/secrets/{sid}")
[...]

Issue #2873, opened 2021-06-07
Catch exported resources warning and raise a Bolt warning


lucywyman:
## Use Case
[Exported resources](https://puppet.com/docs/puppet/latest/lang_exported.html) are not supported in Bolt manifests because Bolt does not load Puppet config, which is required to set up exported resources. [While documented](https://puppet.com/docs/bolt/latest/applying_manifest_blocks.html#manifest-block-limitations) this means that exported resource raise a Puppet warning, even when the user knows that exported resources won't work (and the warning is specific to Puppet rather than Bolt). To make this a bit friendly for folks who are expecting the warning, or for those who don't understand why the warning is being issued, we can catch this log [in the PAL log handler](https://github.com/puppetlabs/bolt/blob/main/lib/bolt/pal/logging.rb#L14) and raise a Bolt warning instead, stating that Puppet configuration cannot be loaded and exported resources aren't supported. This warning can then be turned off using the warning configuration by users that are expecting it.

PR #2906, opened 2021-06-29
(GH-2873) Disable 'exported resources' warning from Puppet

beechtom:
This adds an ID to 'exported resources' warnings that come from Puppet
when applying a manifest that creates an exported resource or collects
exported resources. This allows these warnings to be disabled by listing
the ID under the `disable-warnings` config option.

!feature

* **Disable 'exported resources' warnings from Puppet**
  ([#2889](https://github.com/puppetlabs/bolt/issues/2889))

  Warnings logged by Puppet when declaring or collecting exported
  resources in a manifest can now be disabled. To disable these warnings
  add the `exported_resources` ID under the `disable-warnings`
  configuration option.

### DIFF
--- a/lib/bolt/applicator.rb
+++ b/lib/bolt/applicator.rb
@@ -122,7 +122,13 @@ module Bolt
         logs.each do |log|
           bolt_level = Bolt::Util::PuppetLogLevel::MAPPING[log['level'].to_sym]
           message = log['message'].chomp
-          @logger.send(bolt_level, "#{target.name}: #{message}")
+
+          case bolt_level
+          when :warn
+            handle_warning(target, message)
+          else
+            @logger.send(bolt_level, "#{target.name}: #{message}")
+          end
         end
       end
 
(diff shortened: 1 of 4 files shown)


lib/bolt/applicator.rb
```diff
@@ -138,6 +144,20 @@ def compile(target, scope)
       result
     end
 
+    # Handles logging Puppet warnings, some of which are suppressable.
+    #
+    # @param target [Bolt::Target] The target the apply ran on.
+    # @param message [String] The log message.
+    #
+    private def handle_warning(target, message)
+      if message.include?(Puppet::Pops::Issues::RT_NO_STORECONFIGS_EXPORT.format) ||
```
nicklewis, line 153: This could use a comment about what these messages actually are/why they happen.

beechtom: Commented. 👍

lucywyman (approved):
### From Apply blocks
- [x] Prints the warning by default
- [x] Warning can be disabled
### From Catalogs
- [x] Prints the warning by default
- [x] Warning can be disabled

Issue #2899, opened 2021-06-23
Add a logging module to Bolt


lucywyman:
## Use Case
Bolt plans are only able to print to the console using `out` module functions (`message` and now `verbose`), and logging using Puppet functions is difficult because we downgrade Puppet logs because Puppet is an implementation detail of Bolt and we don't know if the logs are coming explicitly from the plan or from Puppet itself. In order to distinguish between logs from Puppet and logs from plan authors we can add a `log` module with the same log levels as Bolt which just sends message to the Bolt logger at the specified level.

`log::info`, `log::warn`, etc. 

These functions should not use Puppet's logging functions or affect Puppet's logs.

PR #2908, opened 2021-06-30
(GH-2899) Add built-in log module

beechtom:
This adds a new built-in `log` module that includes plan functions for
logging messages at each of Bolt's log levels. Functions include
`log::trace`, `log::debug`, `log::warn`, `log::debug`, `log::error`, and
`log::fatal`. Each of the functions publishes a `:log` event to the
executor. The logger outputter has been udpated to handle this event
type and passes the message directly to the configured logger instance.

!feature

* **Add built-in `log` module**
  ([#2899](#2899))

  Bolt now ships with a new built-in `log` module that includes plan
  functions for logging messages at each of Bolt's log levels. The new
  plan functions include `log::trace`, `log::debug`, `log::warn`,
  `log::info`, `log::error`, and `log::fatal`.

---

This adds a new `Bolt::Util::Format` module that includes utility
methods for formatting. Specifically, this moves the `stringify()`
method that used to live on the `out::message` function to a reusable
module, as the `log::*` functions need this functionality as well.

Note: These methods cannot live on `Bolt::Outputter`, as they must be
able to format Puppet object types in plans. Attempting to stringify a
Puppet object type outside the scope of a plan results in a "stack trace
too deep" error.

!no-release-note

### DIFF
--- a/bolt-modules/log/lib/puppet/functions/log/debug.rb
+++ b/bolt-modules/log/lib/puppet/functions/log/debug.rb
@@ -1,0 +1,37 @@
+# frozen_string_literal: true
+
+# Log a debugging message.
+#
+# Messages logged at this level typically include detailed information about
+# what a plan is doing. For example, you might log a message at the `debug`
+# level that shows what value is returned from a function invocation.
+#
+# See [Logs](logs.md) for more information about Bolt's log levels.
+#
+# > **Note:** Not available in apply block
+Puppet::Functions.create_function(:'log::debug') do
+  # Log a debugging message.
+  # @param message The message to log.
+  # @example Log a debugging message
+  #   log::trace("Function frogsay returned: ${result}")
+  dispatch :log_debug do
+    param 'Any', :message
+    return_type 'Undef'
+  end
+
+  def log_debug(message)
+    unless Puppet[:tasks]
+      raise Puppet::ParseErrorWithIssue.from_issue_and_stack(
+        Bolt::PAL::Issues::PLAN_OPERATION_NOT_SUPPORTED_WHEN_COMPILING,
+        action: 'log::debug'
+      )
+    end
+
+    Puppet.lookup(:bolt_executor).tap do |executor|
+      executor.report_function_call(self.class.name)
+      executor.publish_event(type: :log, level: :debug, message: message)
+    end
+
+    nil
+  end
+end
--- a/bolt-modules/log/lib/puppet/functions/log/debug.rb
+++ b/bolt-modules/log/lib/puppet/functions/log/debug.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require 'bolt/util/format'
+
 # Log a debugging message.
 #
 # Messages logged at this level typically include detailed information about
@@ -13,7 +15,7 @@ Puppet::Functions.create_function(:'log::debug') do
   # Log a debugging message.
   # @param message The message to log.
   # @example Log a debugging message
-  #   log::trace("Function frogsay returned: ${result}")
+  #   log::debug("Function frogsay returned: ${result}")
   dispatch :log_debug do
     param 'Any', :message
     return_type 'Undef'
@@ -29,7 +31,7 @@ Puppet::Functions.create_function(:'log::debug') do
 
     Puppet.lookup(:bolt_executor).tap do |executor|
       executor.report_function_call(self.class.name)
-      executor.publish_event(type: :log, level: :debug, message: message)
+      executor.publish_event(type: :log, level: :debug, message: Bolt::Util::Format.stringify(message))
     end
 
     nil
--- a/bolt-modules/log/lib/puppet/functions/log/warn.rb
+++ b/bolt-modules/log/lib/puppet/functions/log/warn.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require 'bolt/util/format'
+
 # Log a warning message.
 #
 # Messages logged at this level typically include messages about deprecated
@@ -31,7 +33,7 @@ Puppet::Functions.create_function(:'log::warn') do
 
     Puppet.lookup(:bolt_executor).tap do |executor|
       executor.report_function_call(self.class.name)
-      executor.publish_event(type: :log, level: :warn, message: message)
+      executor.publish_event(type: :log, level: :warn, message: Bolt::Util::Format.stringify(message))
     end
 
     nil
--- a/bolt-modules/log/lib/puppet/functions/log/warn.rb
+++ b/bolt-modules/log/lib/puppet/functions/log/warn.rb
@@ -1,0 +1,39 @@
+# frozen_string_literal: true
+
+# Log a warning message.
+#
+# Messages logged at this level typically include messages about deprecated
+# behavior or potentially harmful situations that might affect the plan run.
+# For example, you might log a message at the `warn` level if you are
+# planning to make a breaking change to your plan in a future release and
+# want to notify users.
+#
+# See [Logs](logs.md) for more information about Bolt's log levels.
+#
+# > **Note:** Not available in apply block
+Puppet::Functions.create_function(:'log::warn') do
+  # Log a warning message.
+  # @param message The message to log.
+  # @example Log a warning message
+  #   log::warn('This plan will no longer install the package in a future release.')
+  dispatch :log_warn do
+    param 'Any', :message
+    return_type 'Undef'
+  end
+
+  def log_warn(message)
+    unless Puppet[:tasks]
+      raise Puppet::ParseErrorWithIssue.from_issue_and_stack(
+        Bolt::PAL::Issues::PLAN_OPERATION_NOT_SUPPORTED_WHEN_COMPILING,
+        action: 'log::warn'
+      )
+    end
+
+    Puppet.lookup(:bolt_executor).tap do |executor|
+      executor.report_function_call(self.class.name)
+      executor.publish_event(type: :log, level: :warn, message: message)
+    end
+
+    nil
+  end
+end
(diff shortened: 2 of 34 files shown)


lib/bolt/util/format.rb
```diff
@@ -0,0 +1,68 @@
+# frozen_string_literal: true
+
+module Bolt
+  module Util
+    module Format
```
beechtom, line 5: Not really attached to having smaller utility modules like this. Thought it'd be nice just from an organizational standpoint (for example we could have `Bolt::Util::Files` as well), but I'm also fine with keeping everything in `Bolt::Util`.

nicklewis: I think the word `util` is a code smell, which means having a single monolithic `util` namespace is especially a code smell. So breaking it up seems like an improvement.

bolt-modules/out/lib/puppet/functions/out/message.rb
```diff
@@ -22,55 +24,11 @@ def output_message(message)
...
-  def format_apply_result(result)
-    logs = result.resource_logs&.map do |log|
-      # Omit low-level info/debug messages
-      next if %w[info debug].include?(log['level'])
-      indent(2, format_log(log))
-    end
-    hash = result.to_data
-    hash['logs'] = logs unless logs.empty?
-    hash
+    nil
```
artlawson, line 32: Is `nil` serving as a default return value? Not sure why `hash` and `nil` are at the end of the function.

beechtom: Yep. The last evaluated statement in a method is its return value. We specifically don't want the `out::*` and `log::*` functions to return anything, so we return `nil` here.

lib/bolt/util/format.rb
beechtom: Good point. I found [this](https://www.jmgundersen.net/blog/utility-and-helper-classes-a-code-smell) blog post which argues the same, and I think they make some very valid points as well.

bolt-modules/log/lib/puppet/functions/log/debug.rb
```diff
@@ -0,0 +1,39 @@
...
+    unless Puppet[:tasks]
+      raise Puppet::ParseErrorWithIssue.from_issue_and_stack(
+        Bolt::PAL::Issues::PLAN_OPERATION_NOT_SUPPORTED_WHEN_COMPILING,
+        action: 'log::debug'
+      )
+    end
+
+    Puppet.lookup(:bolt_executor).tap do |executor|
+      executor.report_function_call(self.class.name)
+      executor.publish_event(type: :log, level: :debug, message: Bolt::Util::Format.stringify(message))
```
lucywyman, line 34: Hm, I think a more specific event type like `plan_function_log` might be helpful if we want to add more types of events in the future. `plan_log` might work but is still kind of ambiguous...

bolt-modules/log/lib/puppet/functions/log/debug.rb
```diff
@@ -0,0 +1,39 @@
...
+# level that shows what value is returned from a function invocation.
+#
+# See [Logs](logs.md) for more information about Bolt's log levels.
+#
+# > **Note:** Not available in apply block
+Puppet::Functions.create_function(:'log::debug') do
+  # Log a debugging message.
+  # @param message The message to log.
+  # @example Log a debugging message
+  #   log::trace("Function frogsay returned: ${result}")
```
lucywyman, line 18:
```suggestion
  #   log::debug("Function frogsay returned: ${result}")
```

bolt-modules/log/lib/puppet/functions/log/warn.rb
```diff
@@ -0,0 +1,41 @@
...
+# planning to make a breaking change to your plan in a future release and
+# want to notify users.
+#
+# See [Logs](logs.md) for more information about Bolt's log levels.
+#
+# > **Note:** Not available in apply block
+Puppet::Functions.create_function(:'log::warn') do
+  # Log a warning message.
+  # @param message The message to log.
+  # @example Log a warning message
```
lucywyman, line 19:
```suggestion
  # @example Log a warn message
```
CC @hestonhoffman no idea which is right.

lib/bolt/bolt_option_parser.rb
```diff
@@ -1088,7 +1088,7 @@ def initialize(options)
       end
       define('--log-level LEVEL',
              "Set the log level for the console. Available options are",
-             "trace, debug, info, warn, error, fatal, any.") do |level|
+             "trace, debug, info, warn, error, fatal.") do |level|
```
lucywyman, line 1091: I think this also needs to be removed from `resources/bolt_bash_completion.sh`, `rakelib/pwsh.rake`, `lib/bolt_spec/plans.rb`, and `lib/bolt/config/options.rb`.

bolt-modules/log/lib/puppet/functions/log/debug.rb
beechtom: I'm not sure the extra specificity is needed? All the event does is send the message directly to the logger; it doesn't really care or need to know that the message came from a plan function. If we were doing something like sending analytics that the log came from a plan function, then I can see the extra specificity making sense.

lib/bolt/util/format.rb
lucywyman: Breaking this out is definitely great, and like...there are other ways to make these functions available in Puppet plan functions but to me a utility class seems fine.

lucywyman (commented):
Looks like ` spec/fixtures/configs/empty/bolt-debug.log ` got added by accident.

- [x] Logging functions send log messages to all configured log locations
- [x] The Bolt log level aligns with the level log function messages are logged at (ie. `log::debug` logs at Bolt's debug level)
- [x] `log::debug` logs a message at debug level
- [x] Can't be called from an apply block
- [x] Sends analytics for function calls
- [x] Functions return `nil`
- [x] Works if `bolt-debug.log` is disabled
- [x] Log module spec tests run in Github CI

bolt-modules/log/lib/puppet/functions/log/debug.rb
lucywyman: Hm, yeah, I also realized we can just update this type if we want something else to have `log` type in the future.

lucywyman (changes requested):
Heh, I was looking around for mentions of `out::message` in the docs reviewing #2914 and saw this section https://puppet.com/docs/bolt/latest/troubleshooting.html#puppet-log-functions-are-not-logging-to-the-console which calls out using `out::message` instead of Puppet log functions, which we can definitely update with this change!

lib/bolt/bolt_option_parser.rb
beechtom: I wish I knew why I included `any` in the first place... :|

lucywyman (approved):

bolt-modules/log/lib/puppet/functions/log/warn.rb
hestonhoffman: I'd lean towards "warning" here. We should change the references to "warn logs" in logs.md too. These log entries are kinda weird cos they're all different parts of speech.

documentation/troubleshooting.md
```diff
@@ -128,23 +128,29 @@ the following ways:
...
-the `--log-level debug` command-line option, use the `notice` Puppet log
-function. The `notice` function sets the console log level to `debug` for that
-run.
-
-For more information, see the docs for configuring [Bolt's log
-level](https://puppet.com/docs/bolt/latest/bolt_configuration_options.html#log-file-configuration-options).
+Puppet logs might not be printing to the console because they are logged at a
+lower level than Bolt is configured to print at. By default, Bolt prints all
+logs at the `warn` level or higher to the console. If you use the `notice()`
+Puppet log function, then Bolt will not print it to the console by default.
```
hestonhoffman, line 134:
```suggestion
logs at the `warn` level or higher to the console.
```
Moving this bit down

documentation/troubleshooting.md
```diff
@@ -128,23 +128,29 @@ the following ways:
...
+
+Additionally, Puppet logs are printed at a different level in Bolt than they
+would be in Puppet. You can see which Bolt log level each Puppet log level maps
+to in [Puppet log functions in
+Bolt](writing_plans.md#puppet-log-functions-in-bolt).
```
hestonhoffman, line 139:
```suggestion
Additionally, Puppet logs are printed at a different level in Bolt than they
would be in Puppet. For example, Puppet `notice` level logs are equivalent
to the `info` level in Bolt. As a result, if you use the `notice()`
Puppet log function, Bolt does not print the contents of `notice` to the
console by default. You can see which Bolt log level each Puppet log level
maps to in [Puppet log functions in
Bolt](writing_plans.md#puppet-log-functions-in-bolt).
```
I think the `notice` example makes more sense here.

documentation/writing_plans.md
```diff
@@ -896,11 +896,25 @@ without_default_logging { run_command('echo hi', $targets) }
...
+| `error` | [`log::error`](plan_functions.md#logtrace) |
+| `fatal` | [`log::fatal`](plan_functions.md#logfatal) |
+
 ### Puppet log functions in Bolt
 
 You can use Puppet log functions in Bolt plans, but Bolt log levels do not map
 directly to Puppet log levels. For example, a `notice` function in a plan logs
-at the `info` level in Bolt. Log levels map as follows:
+at the `info` level in Bolt. Whenever possible, use Bolt's log functions instead
+of Puppet's. Log levels map as follows:
```
hestonhoffman, line 917:
```suggestion
at the `info` level in Bolt. Whenever possible, use Bolt log functions instead
of Puppet log functions. Log levels map as follows:
```
Generally, we try to avoid using product names in the possessive. It's a style thing that I tend to be kind of lenient on, but in this case I think it's better to avoid it.

hestonhoffman (commented):
Couple small edits, but the docs LGTM

hestonhoffman (approved):

PR #2911, opened 2021-06-30
(PE-32176) Use versioned_project from body for /project_inventory_targets

nicklewis:
At one time, this endpoint used a GET request and therefore expected
versioned_project as a query param. It was later changed to a POST
request with puppet_connect_data included in the request body, but the
versioned_project query param remained. For consistency, we now expect
both pieces of data to be supplied in the request body.

### DIFF
--- a/lib/bolt_server/schemas/connect-data.json
+++ b/lib/bolt_server/schemas/connect-data.json
@@ -16,7 +16,10 @@
         }
       },
       "additionalProperties": false
+    },
+    "versioned_project": {
+      "type": "string"
     }
   },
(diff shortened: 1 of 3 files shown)


nicklewis:
This shouldn't be merged until https://github.com/puppetlabs/orchestrator/pull/1745 is merged and promoted. That PR adds compatibility to orchestrator for _both_ the query param and body param, so it can be safely merged before this. Once that's done, we can merge this PR. Once this one is released and in PE, we can update orchestrator to no longer set the query param.

donoghuc (approved):

PR #2912, opened 2021-06-30
(bug) Fix stacktrace when cache includes missing file

beechtom:
This fixes a bug where Bolt would stacktrace when listing plans or tasks
and the plugin or task cache included a file that no longer exists.
Specifically, this error is raised by the `File.mtime` method receiving
a non-existent filepath.

The conditions that cause this bug are:

- Moving a module to a different location in the modulepath.
- Deleting a module that shadows another module of the same name on the
  modulepath.

If the user then runs `plan|task show`, Bolt will error.

!bug

* **Do not stacktrace with non-existent files in plan or task cache**

  Bolt no longer errors with a stacktrace when listing plans or tasks
  when the plan or task cache include paths to files that no longer
  exist.

### DIFF
--- a/lib/bolt/pal.rb
+++ b/lib/bolt/pal.rb
@@ -324,7 +324,7 @@ module Bolt
           data['files'].each do |f|
             # If any file has been updated since we last cached, update the
             # cache
-            next if File.mtime(f['path']).to_s == f['mtime']
+            next unless file_modified?(f['path'], f['mtime'])
             data = get_task_info(task_name, with_mtime: true)
             data = Bolt::Util.walk_keys(data, &:to_s)
             # Tell Bolt to write to the cache file once we're done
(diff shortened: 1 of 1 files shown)


lib/bolt/pal.rb
```diff
@@ -443,6 +440,15 @@ def list_plans_with_cache(filter_content: false)
       filter_content ? filter_content(plan_list, @project&.plans) : plan_list
     end
 
+    # Returns true if a file has been modified, false otherwise.
```
lucywyman, line 443:
```suggestion
    # Returns true if a file exists and has been modified, false otherwise.
```

lucywyman (approved):
Ah, cache invalidation. I filed https://github.com/puppetlabs/bolt/issues/2915 since I really don't see any good way to detect that a new module with the same name has been added further up the modulepath without spending a lot of time looking for each plan.

lib/bolt/pal.rb
beechtom: Updated.

lucywyman (approved):

PR #2916, opened 2021-07-06
(maint) Fix broken Windows acceptance tests

beechtom:
Starting with Bolt 3.0, we no longer include the `bolt` function in the
PuppetBolt PowerShell module and instead shipped with a batch file for
Windows packages that would run the Bolt executable. When this change
was made acceptance tests were broken as the `PATH` did not include the
path to Bolt's `bin/` directory and the test would error with a "bolt is
not a recognized command" message.

During the beaker setup process (specifically step 032), beaker
updates the `PATH` to include Bolt's `bin/` directory. Previously, the
script responsible for this used the `Join-Path` cmdlet with multiple
positional arguments, which is not supported in PowerShell < 6.0,
resulting in the `PATH` not getting updated. Updating this line to not
error correctly sets the `PATH` for the step that sets it, but the
`PATH` is not preserved between PowerShell sessions. This resulted in
the tests continuing to fail since `bolt` was not recognized as a
command.

This change updates the `bolt_command_on` helper method to set the
`PATH` prior to invoking `bolt`, ensuring that the `PATH` is correctly
configured every time Bolt is run.

!no-release-note

### DIFF
--- a/acceptance/lib/acceptance/bolt_command_helper.rb
+++ b/acceptance/lib/acceptance/bolt_command_helper.rb
@@ -18,7 +18,16 @@ module Acceptance
 
       case host['platform']
       when /windows/
-        execute_powershell_script_on(host, bolt_command, opts)
+        # Set the PATH environment variable before running the Bolt command. Ideally this
+        # would be set during the setup process, but the PATH is not preserved between
+        # sessions on the SUT.
+        script = <<~PS
+          $boltpath = [IO.Path]::Combine($env:ProgramFiles, 'Puppet Labs', 'Bolt', 'bin')
+          $env:Path = $boltpath + ";" + $env:Path
+          #{bolt_command}
+        PS
+
+        execute_powershell_script_on(host, script, opts)
       when /osx/
         # Ensure Bolt runs with UTF-8 under macOS. Otherwise we get issues with
         # UTF-8 content in task results.
--- a/acceptance/setup/common/pre-suite/032_configure_windows_profile.rb
+++ b/acceptance/setup/common/pre-suite/032_configure_windows_profile.rb
@@ -5,36 +5,35 @@ require 'bolt_setup_helper'
 test_name "Configure Windows Profile" do
   extend Acceptance::BoltSetupHelper
 
+  # TODO: The PATH environment variable is not preserved between sessions on the SUT.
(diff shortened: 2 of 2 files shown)


beechtom:
Ran locally and confirmed that this results in acceptance tests passing with a Windows controller:

```shell
$ bundle exec beaker-hostgenerator windows2019-64bolt%2Cwinrm. --hypervisor abs --disable-default-role --osinfo-version 1 > hosts.yaml
$ bundle exec beaker --log-level debug --hosts hosts.yaml --tests tests --preserve-hosts never --keyfile ~/.ssh/id_rsa-acceptance --options config/package/options.rb

...

      Test Suite: tests @ 2021-07-06 14:27:15 -0700

       Total Suite Time: 347.57 seconds
      Average Test Time: 19.31 seconds
              Attempted: 18
                 Passed: 10
                 Failed: 0
                Errored: 0
                Skipped: 8
                Pending: 0
                  Total: 18

      - Specific Test Case Status -
```

acceptance/lib/acceptance/bolt_command_helper.rb
```diff
@@ -18,7 +18,16 @@ def bolt_command_on(host, command, flags = {}, opts = {})
 
       case host['platform']
       when /windows/
-        execute_powershell_script_on(host, bolt_command, opts)
+        # Set the PATH environment variable before running the Bolt command. Ideally this
```
lucywyman, line 21: I think you can do `host.add_env_var('PATH', new_path)` which actually it looks like we're one of the only ones that already does this in our setup 😅 [in Bolt acceptance](https://github.com/puppetlabs/bolt/blob/b77d7e7f2f458824fcbccad78367a0f113638463/acceptance/setup/common/pre-suite/010_install_ruby.rb), [Github search](https://github.com/search?q=org%3Apuppetlabs+add_env_var+path+path%3Aacceptance%2Fsetup&type=Code)

beechtom: @lucywyman Ya, that doesn't work. I'm not sure if it's some sort of config for the VM, but the path isn't preserved between PS sessions. Each beaker method that connects to the host runs in a new PS session, so the path won't have whatever changes were made with `host.add_env_var` when we run the Bolt command.

Issue #2926, opened 2021-07-12
Add `verbose` YAML plan key


lucywyman:
## Use Case
To match the new `out::verbose` Puppet plan function, we should also add a `verbose` step key to YAML plans. This should behave the same as the `out::verbose` plan function in Puppet, and similar to the `message` step accepts a string.

### To do
- [x] Ensure the step prints when run with `--verbose`
- [x] Ensure the step doesn't print when run without `--verbose` or with `--no-verbose`
- [x] Ensure the step prints when using both the JSON and human outputter
- [x] Ensure that `bolt plan convert` correctly converts the `verbose` step to an `out::verbose` function call

PR #2928, opened 2021-07-13
(GH-2926) Add YAML plan verbose step

artlawson:
This adds a YAML plan verbose step that prints a message when run in verbose mode.

!feature

* **YAML plan verbose step**
  ([#2926](https://github.com/puppetlabs/bolt/issues/2926))

  YAML plans now support a verbose step that prints a message when run in verbose mode.

### DIFF
--- a/documentation/writing_yaml_plans.md
+++ b/documentation/writing_yaml_plans.md
@@ -107,7 +107,7 @@ Message steps support the following keys:
 
 | Key | Type | Description | Required |
 | --- | --- | --- | --- |
-| `message` | `String` | The message to print. | ✓ |
+| `message` | `Any` | The message to print. | ✓ |
 
 For example:
 
@@ -123,6 +123,30 @@ prints the object as a string.
 
 For information on printing a step result with `message`, see [Debugging
 plans](#debugging-plans).
+
+### Verbose step
+
+Use a `verbose` step to print a message in verbose mode. The step prints a message to standard
+out (stdout) when using the `human` output format, and prints to standard error
+(stderr) when using the `json` output format. 
+
+Verbose steps support the following keys:
+
+| Key | Type | Description | Required |
+| --- | --- | --- | --- |
+| `verbose` | `Any` | The message to print. | ✓ |
+
+For example:
+
+```yaml
+steps:
+  - verbose: hello world
+```
+
+You can pass variables to the verbose step to print them to stdout. If the 
+variable is a [Bolt datatype](bolt_types_reference.md) it will be formatted 
+as a Hash. Once the object is formatted, if it's a Hash or Array it is printed
+as JSON, otherwise Bolt prints the object as a string.
 
 ### Command step
 
--- a/lib/bolt/pal/yaml_plan/step/verbose.rb
+++ b/lib/bolt/pal/yaml_plan/step/verbose.rb
@@ -4,25 +4,25 @@ module Bolt
   class PAL
     class YamlPlan
       class Step
-        class Message < Step
+        class Verbose < Step
           def self.allowed_keys
-            super + Set['message']
+            super + Set['verbose']
           end
 
           def self.required_keys
-            Set['message']
+            Set['verbose']
           end
 
           # Returns an array of arguments to pass to the step's function call
           #
           private def format_args(body)
-            [body['message']]
+            [body['verbose']]
           end
 
           # Returns the function corresponding to the step
           #
           private def function
-            'out::message'
+            'out::verbose'
           end
         end
       end
--- a/schemas/bolt-yaml-plan.schema.json
+++ b/schemas/bolt-yaml-plan.schema.json
@@ -320,7 +320,24 @@
           "type": "string"
         }
       }
-    }
+    },
+    "verbose": {
+      "description": "Print a message only when run in verbose mode. This will print a message to standard out (stdout) when using the 'human' output format or standard error (stderr) when using the 'json' output format.",
+      "type": "object",
+      "additionalProperties": false,
+      "required": ["verbose"],
+      "properties": {
+        "description": {
+          "$ref": "#/definitions/description"
+        },
+        "verbose": {
+          "description": "The message to print."
+        },
+        "name": {
+          "$ref": "#/definitions/name"
+        }
+      }
+    },
   },
   "definitions": {
     "catch_errors": {
--- a/spec/fixtures/modules/yaml/plans/verbose.yaml
+++ b/spec/fixtures/modules/yaml/plans/verbose.yaml
@@ -1,0 +1,9 @@
+parameters:
+  message:
+    type: String
+    default: hershey
+
+steps:
+  - name: verbose
+    verbose: $message
+    description: 'This will call out::verbose(hershey)'
--- a/spec/integration/plan_spec.rb
+++ b/spec/integration/plan_spec.rb
@@ -77,24 +77,18 @@ describe 'plans' do
 
       let(:config_flags) { ['--no-host-key-check'] }
       let(:opts)         { { outputter: Bolt::Outputter::Human, project: @project } }
+      let(:project)      { @project }
 
-      context 'output' do
-        let(:output)    { StringIO.new }
-        let(:outputter) { Bolt::Outputter::Human.new(true, true, true, true, output) }
-
-        before(:each) do
-          allow(Bolt::Outputter::Human).to receive(:new).and_return(outputter)
+      context 'out::verbose' do
+        it 'outputs verbose messages in verbose mode' do
+          result = run_cli(%w[plan run output::verbose --verbose], outputter: Bolt::Outputter::Human,
+                                                                   project: project)
+          expect(result).to match(/Hi, I'm Dave/)
         end
 
-        it 'outputs message with verbose flag' do
-          run_cli(%W[plan run output::verbose --targets #{target} --verbose] + config_flags,
-                  outputter: Bolt::Outputter::Human, project: @project)
-          expect(output.string).to match(/Hi, I'm Dave/)
-        end
-
-        it 'doesnt output without verbose flag' do
-          result = run_cli(%W[plan run output::verbose --targets #{target}] + config_flags,
-                           outputter: Bolt::Outputter::Human, project: @project)
+        it 'does not output verbose messages when not in verbose mode' do
+          result = run_cli(%w[plan run output::verbose], outputter: Bolt::Outputter::Human,
+                                                         project: project)
           expect(result).not_to match(/Hi, I'm Dave/)
         end
       end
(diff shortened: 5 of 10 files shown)


spec/integration/plan_spec.rb
```diff
@@ -95,7 +95,9 @@
         it 'doesnt output without verbose flag' do
           result = run_cli(%W[plan run output::verbose --targets #{target}] + config_flags,
                            outputter: Bolt::Outputter::Human, project: @project)
-          expect(result).not_to match(/Hi, I'm Dave/)
+          #see yaml_plan_spec.rb
+          expect(config_flags).not_to include '--verbose' #:(
+          expect(output.string).not_to match(/Hi, I'm Dave/)
```
artlawson, line 100: Correcting previous verbose PR. I'm pretty sure we don't want `expect(result)` here. Not passing, but I think once we figure out this issue on the current PR then it'll fix both.

nicklewis:
Would it make more sense for this to just be a `verbose: <true/false>` parameter to the existing `message` step, rather than a new step altogether?

beechtom:
@artlawson We've discussed that in the past (in relation to creating a `prompt` step that executes either the `prompt` function or `prompt::menu` function depending on the keys present), and thought it was best to keep a 1:1 relationship for steps and plan functions. 

If `out::verbose` had instead been implemented as a `verbose` parameter to the `out::message` plan function, then I think having a `verbose` key for the `message` step would have been appropriate.

lib/bolt/pal/yaml_plan/step/verbose.rb
```diff
@@ -0,0 +1,39 @@
...
+            'out::verbose'
+          end
+
+          # Transpiles the step into the plan language
+          #
+          def transpile
+            code = String.new("  ")
+            code << function_call(function, format_args(body))
+            code << "\n"
+          end
```
beechtom, line 34: It looks like this actually results in a small bug when transpiling plans, and is present in the `message` step as well. 

This YAML plan:

```yaml
steps:
  - name: verbose
    verbose: verbosity

return: $verbose
```

Becomes this Puppet plan:

```puppet
plan bolt::yaml() {
  out::verbose('verbosity')

  return $verbose
}
```

`name` keys in YAML plans are equivalent to variable declarations in Puppet. However, the transpile method for `message` and `verbose` isn't declaring these variables, so the above Puppet plan will actually error. We should remove the `transpile` method override in `message` and `verbose`, since the base class' method will do what we want.

beechtom (changes requested):
The [YAML plan schema](https://github.com/puppetlabs/bolt/blob/main/schemas/bolt-yaml-plan.schema.json) needs to be updated to include a `verbose` step.

The [Writing YAML plans doc page](https://github.com/puppetlabs/bolt/blob/main/documentation/writing_yaml_plans.md#steps) also needs to be updated to include the `verbose` step.

artlawson:
@beechtom  thanks for the additional step/function context

lib/bolt/pal/yaml_plan/step/verbose.rb
artlawson: Would we need any additional testing to make sure that this works, or would the existing testing (plus checking the conversion) be sufficient? Where would I check to see how to manually test the conversion.

beechtom: I don't think an extra test is needed here, just a manual check should be fine.

documentation/writing_yaml_plans.md
```diff
@@ -124,6 +124,34 @@ prints the object as a string.
 For information on printing a step result with `message`, see [Debugging
 plans](#debugging-plans).
 
+
+### Verbose step
```
artlawson, line 128: @hestonhoffman  Currently this is just a port of `message` step right above it. I figured a lot of it would be able to be "similar to `message` step...." but I wasn't sure how much. Any advice?

documentation/writing_yaml_plans.md
```diff
@@ -124,6 +124,34 @@ prints the object as a string.
...
+  - verbose: hello world
+```
+
+You can pass variables to the verbose step to print them to stdout. If the
+variable contains a valid plan result, Bolt formats the plan result using a JSON
+representation of the result object. If the object is not a plan result, Bolt
+prints the object as a string.
+
+For information on printing a step result with `message`, see [Debugging
+plans](#debugging-plans).
```
artlawson, line 153: Also planned on deleting this, unless there is a useful debugging verbose link to point to?

beechtom: We can probably add using the `verbose` step [here](https://puppet.com/docs/bolt/latest/writing_yaml_plans.html#using-the---verbose-option-for-debugging).

beechtom (approved):
Other than a small comment on docs, lgtm 👍

documentation/writing_yaml_plans.md
hestonhoffman: hi @artlawson, it's better not to use "similar to..." language. It's not a great user experience to find the thing you're looking for and then have to go somewhere else. It's also problematic over time because the parity between things can shift and create a maintenance burden. Copying another section and changing the relevant information is a great approach for something like this. It keeps the docs consistent, and it gives the user the info they're looking for 👍

hestonhoffman: Having said that, if the two sections are very similar and are linked in functionality, we might have an opportunity to combine the two sections 🤔 I'll take another look when I review.

artlawson: Ok great! Thanks for the insight :)

documentation/writing_yaml_plans.md
```diff
@@ -124,6 +124,34 @@ prints the object as a string.
...
+
+Use a `verbose` step to print a message in verbose mode. The step prints a message to standard
+out (stdout) when using the `human` output format, and prints to standard error
+(stderr) when using the `json` output format. 
+
+Verbose steps support the following keys:
+
+| Key | Type | Description | Required |
+| --- | --- | --- | --- |
+| `verbose` | `String` | The message to print. | ✓ |
```
lucywyman, line 138:
```suggestion
| `verbose` | `Any` | The message to print. | ✓ |
```

lucywyman: We should update this for `message` in this doc too - originally this function *only* accepted Strings, but we [added support for printing any object](https://github.com/puppetlabs/bolt/pull/2055/files) and just missed updating these docs.

lucywyman (approved):
- [x] Prints with verbose
- [x] Doesn't print without verbose
- [x] Doesn't print with --no-verbose
- [x] Converts to a Puppet plan
- [x] Prints objects
- [x] Accepts an empty string
- [x] Uses YAML plan string interpolation + behavior

documentation/writing_yaml_plans.md
```diff
@@ -124,6 +124,34 @@ prints the object as a string.
...
+
+For example:
+
+```yaml
+steps:
+  - verbose: hello world
+```
+
+You can pass variables to the verbose step to print them to stdout. If the
+variable contains a valid plan result, Bolt formats the plan result using a JSON
```
lucywyman, line 148: This also isn't *quite* true - if the types we get are Boltlib datatypes (or structures like Arrays or Hashes that include Bolt datatypes) then we'll convert them to their hash representation. We then print any Hash or Array we get as JSON after formatting Bolt datatypes. I think something like:
```
If the variable is a [Bolt datatype](bolt_types_reference.md) it will be formatted as a Hash. Once the object is formatted, if it's a Hash or Array it is printed as JSON, otherwise Bolt prints the object as a string.
```

schemas/bolt-yaml-plan.schema.json
```diff
@@ -320,7 +320,24 @@
...
+    "verbose": {
+      "description": "Print a message only when run in verbose mode. This will print a message to standard out (stdout) when using the 'human' output format or standard error (stderr) when using the 'json' output format.",
+      "type": "object",
+      "additionalProperties": false,
+      "required": ["verbose"],
+      "properties": {
+        "description": {
+          "$ref": "#/definitions/description"
+        },
+        "message": {
```
lucywyman, line 333:
```suggestion
        "verbose": {
```

schemas/bolt-yaml-plan.schema.json
```diff
@@ -320,7 +320,24 @@
...
+          "description": "The message to print. Messages will be specially formatted if they are valid PlanResult objects, otherwise they will be stringified."
```
lucywyman, line 334:
```suggestion
          "description": "The message to print."
```

spec/fixtures/modules/yaml/plans/verbose.yaml
```diff
@@ -0,0 +1,11 @@
...
+  message:
+    type: String
+    default: hershey
+
+steps:
+  - name: verbose
+    verbose: $message
+    description: 'This will call out::verbose(hershey)'
+
+return: $verbose
```
lucywyman, line 11: I don't think this return is meaningful since the function itself returns nil.

artlawson:
Waiting on Heston's review for 'documentation/writing_yaml_plans.md' suggestion, but otherwise should be gtg.

documentation/writing_yaml_plans.md
lucywyman: Hm, I think we could but I don't think we have to. The purpose of the function for debugging is mostly to be able to print variables, results, etc. I made a separate issue for pulling the debugging and logging in plans docs into their own page https://github.com/puppetlabs/bolt/issues/2933 and I think we can add `verbose` to the debugging page then, but for now I think we should just remove this line from the verbose step section.

documentation/writing_yaml_plans.md
```diff
@@ -124,6 +124,34 @@ prints the object as a string.
...
+steps:
+  - verbose: hello world
+```
+
+You can pass variables to the verbose step to print them to stdout. If the 
+variable is a [Bolt datatype](bolt_types_reference.md) it will be formatted 
+as a Hash. Once the object is formatted, if it's a Hash or Array it is printed
+ as JSON, otherwise Bolt prints the object as a string.
+
+For information on printing a step result with `message`, see [Debugging
```
lucywyman, line 152:
```suggestion

```

lucywyman (commented):
Ok, one last comment, then :+1:

documentation/writing_yaml_plans.md
```diff
@@ -124,6 +124,30 @@ prints the object as a string.
...
+
+```yaml
+steps:
+  - verbose: hello world
+```
+
+You can pass variables to the verbose step to print them to stdout. If the 
+variable is a [Bolt datatype](bolt_types_reference.md) it will be formatted 
+as a Hash. Once the object is formatted, if it's a Hash or Array it is printed
+ as JSON, otherwise Bolt prints the object as a string.
```
beechtom, line 149:
```suggestion
as JSON, otherwise Bolt prints the object as a string.
```

lucywyman (approved):

Issue #2900, opened 2021-06-23
Send out::message and out::verbose messages to logs


lucywyman:
## Use Case
We should add messages printed with `out::message` and `out::verbose` plan functions to the logs, since users already have to parse logs and having these messages may be helpful or can be ignored. `out::message` should be logged at info log level and `verbose` should be logged at debug level. This should happen after we add the `log` module.

PR #2929, opened 2021-07-14
(GH-2900) Log messages from `out::message` and `out::verbose`

beechtom:
This updates the `out::message` plan function to log messages at the
`info` level and the `out::verbose` plan function to log messages at the
`debug` level. Both functions continue to print to the console as
before.

!feature

* **Log messages from `out::message` and `out::verbose`**
  ([#2900](https://github.com/puppetlabs/bolt/issues/2900))

  The `out::message` plan function now logs messages at the `info` level
  and the `out::verbose` plan function now logs messages at the `debug`
  level. Both functions continue to print messages to the console.

### DIFF
--- a/bolt-modules/out/lib/puppet/functions/out/message.rb
+++ b/bolt-modules/out/lib/puppet/functions/out/message.rb
@@ -5,7 +5,9 @@ require 'bolt/util/format'
 # Output a message for the user.
 #
 # This will print a message to stdout when using the human output format,
-# and print to stderr when using the json output format
+# and print to stderr when using the json output format. Messages are
+# also logged at the `info` level. For more information about logs, see
+# [Logs](logs.md).
 #
 # > **Note:** Not available in apply block
 Puppet::Functions.create_function(:'out::message') do
@@ -26,7 +28,7 @@ Puppet::Functions.create_function(:'out::message') do
 
     Puppet.lookup(:bolt_executor).tap do |executor|
       executor.report_function_call(self.class.name)
-      executor.publish_event(type: :message, message: Bolt::Util::Format.stringify(message))
+      executor.publish_event(type: :message, message: Bolt::Util::Format.stringify(message), level: :info)
     end
 
     nil
--- a/spec/integration/plan_spec.rb
+++ b/spec/integration/plan_spec.rb
@@ -79,17 +79,35 @@ describe 'plans' do
       let(:opts)         { { outputter: Bolt::Outputter::Human, project: @project } }
       let(:project)      { @project }
 
-      context 'out::verbose' do
-        it 'outputs verbose messages in verbose mode' do
-          result = run_cli(%w[plan run output::verbose --verbose], outputter: Bolt::Outputter::Human,
-                                                                   project: project)
-          expect(result).to match(/Hi, I'm Dave/)
+      context 'output' do
+        context 'out::message' do
+          it 'outputs messages' do
+            result = run_cli(%w[plan run output], outputter: Bolt::Outputter::Human, project: project)
+            expect(result).to match(/Outputting a message/)
+          end
+
+          it 'logs messages at the info level' do
+            run_cli(%w[plan run output], outputter: Bolt::Outputter::Human, project: project)
+            expect(@log_output.readlines).to include(/INFO.*Outputting a message/)
+          end
         end
 
-        it 'does not output verbose messages when not in verbose mode' do
-          result = run_cli(%w[plan run output::verbose], outputter: Bolt::Outputter::Human,
-                                                         project: project)
-          expect(result).not_to match(/Hi, I'm Dave/)
+        context 'out::verbose' do
+          it 'outputs verbose messages in verbose mode' do
+            result = run_cli(%w[plan run output::verbose --verbose], outputter: Bolt::Outputter::Human,
+                                                                     project: project)
+            expect(result).to match(/Hi, I'm Dave/)
+          end
+
+          it 'does not output verbose messages' do
+            result = run_cli(%w[plan run output::verbose], outputter: Bolt::Outputter::Human, project: project)
+            expect(result).not_to match(/Hi, I'm Dave/)
+          end
+
+          it 'logs verbose messages at the debug level' do
+            run_cli(%w[plan run output::verbose], outputter: Bolt::Outputter::Human, project: project)
+            expect(@log_output.readlines).to include(/DEBUG.*Hi, I'm Dave/)
+          end
         end
       end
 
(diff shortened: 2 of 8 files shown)


spec/integration/plan_spec.rb
```diff
@@ -77,25 +77,37 @@
...
+        context 'out::verbose' do
+          it 'outputs verbose messages in verbose mode' do
+            result = run_cli(%w[plan run output::verbose --verbose], outputter: Bolt::Outputter::Human,
+                                                                     project: project)
+            expect(result).to match(/Hi, I'm Dave/)
+          end
+
+          it 'does not output verbose messages' do
+            result = run_cli(%w[plan run output::verbose], outputter: Bolt::Outputter::Human, project: project)
+            expect(result).not_to match(/Hi, I'm Dave/)
```
artlawson, line 104: I'm not sure what `project` represents here.

beechtom: `project` is a `let` that returns the instance variable `@project`. That gets defined [here](https://github.com/puppetlabs/bolt/blob/main/spec/integration/plan_spec.rb#L71-L76) using the `with_project` helper defined [here](https://github.com/puppetlabs/bolt/blob/main/spec/lib/bolt_spec/project.rb). `with_project` creates a temporary Bolt project with configuration and inventory that we can use as the project for the spec test. It simplifies a lot of the setup/cleanup when running integration tests.

We need to explicitly pass the project to `run_cli`, otherwise it creates its own project.

bolt-modules/out/lib/puppet/functions/out/message.rb
```diff
@@ -24,9 +26,12 @@ def output_message(message)
         .from_issue_and_stack(Bolt::PAL::Issues::PLAN_OPERATION_NOT_SUPPORTED_WHEN_COMPILING, action: 'out::message')
     end
 
+    stringified_message = Bolt::Util::Format.stringify(message)
+
     Puppet.lookup(:bolt_executor).tap do |executor|
       executor.report_function_call(self.class.name)
-      executor.publish_event(type: :message, message: Bolt::Util::Format.stringify(message))
+      executor.publish_event(type: :message, message: stringified_message)
+      executor.publish_event(type: :log, level: :info, message: stringified_message)
```
lucywyman, line 34: I don't think this is strictly necessary, but it would be nice (and doable) if this didn't print twice if the log level was such that the logger logs to the console. I think the best way to accomplish that is to have a helper method in util that inspects `Logging.logger[:root].appenders.select { |a| a.name == 'console' }.first.level` to get the level of the console log (in theory there could be a logger for stdout and stderr but that doesn't ever happen in Bolt), and use that here to see if `Bolt::Util.console_log_level =< Logging::level_num(:info)` and *not* log in that case (the levels go down for more verbosity, so `info = 2`, `warn = 3`, `error = 4`, etc.)

If that doesn't ✨ just work ✨ I don't think we have to worry about it.

lucywyman (changes requested):
We should probably mention this functionality [in the debugging section of Writing Plans](https://puppet.com/docs/bolt/latest/writing_plans.html#debugging-plans), but more importantly we explicitly say we *don't* log `out::message` [here](https://puppet.com/docs/bolt/latest/writing_plans.html#plan-logging) 😅

bolt-modules/out/lib/puppet/functions/out/message.rb
beechtom: That would prevent the message from printing in log files as well, which I don't think we want. We could probably skip the `:message`/`:verbose` events, though I'm not 100% sure if that might mess things up anywhere. It seems like it should work, since the logger prints to stderr with the json format...

Fwiw, Bolt already prints some messages to the console twice depending on the log level:

```shell
$be bolt plan run bolt --log-level info

Loaded default project from '/Users/tom/.puppetlabs/bolt'
Starting: plan bolt
Starting: plan bolt
hello
hello
Finished: plan bolt in 0.0 sec
Finished: plan bolt in 0.0 sec
```

I'm fine with suppressing the `:message`/`:verbose` events if you are.

lucywyman: No no, I meant to only print the log event and not the message event, not not do the log event. That check should also happen in the Outputter instead of here, so we always send both events and the outputter knows to not print a message twice if it's already been printed.

I know it's not a big deal, but like I said seems doable and makes sense for users.

beechtom: I guess the last thing we should be aware of is that this means `out::*` will print to _either_ stdout or stderr, depending on the configured log level.

beechtom: Just messing around with the layout for the console logs, I wonder if adding the log level before log messages might help?

```console
INFO: Loaded default project from '/Users/tom/.puppetlabs/bolt'
WARN: CLI arguments ["host-key-check"] might be overridden by Inventory: /Users/tom/.puppetlabs/bolt/inventory.yaml [ID: cli_overrides]
Starting: plan bolt
INFO: Starting: plan bolt
hello
INFO: hello
Finished: plan bolt in 0.01 sec
INFO: Finished: plan bolt in 0.01 sec
```

Just throwing out ideas at this point. :P

lucywyman: That's definitely not a bad idea... https://github.com/beechtom/bolt/pull/22 this is along the lines of what I'm thinking. The only potential issue I see is that we don't know if stdout or stderr are getting redirected or used in a meaningful way, so inspecting both and then "merging" could mean that we don't log something to stdout based on the level of stderr, even if stderr is getting sent somewhere else, so the message isn't seen. I think a better but also riskier tact would be to close and reopen the stderr logger around the message event if it has type `:message`, but Bolt is multithreaded and I'm not confident that won't have it's own issues. I think we can leave it as is for now and if people complain we can go back to the drawing board.

beechtom: I know what you meant and did a [similar implementation](https://github.com/puppetlabs/bolt/commit/e3e4724562d3610d50c64390c289760a56786c40) last week. :) After writing that I realized the weirdness with inconsistent printing between stdout/stderr.

I would definitely prefer to leave it printing twice for now and see if users would prefer the duplicate message to be suppressed.

lib/bolt/logger.rb
```diff
@@ -147,6 +147,18 @@ def self.reset_logging
       Logging.reset
     end
 
+    # Checks if the specified level logs to the console.
+    #
+    def self.log_to_console?(level)
```
lucywyman, line 152: Seems like we can remove these until they're needed.

lucywyman (approved):

lucywyman (approved):

PR #2932, opened 2021-07-14
(PE-32485) Allow plugin_hooks to be passed to bolt-server actions

nicklewis:
This adds the plugin_hooks key to the SSH and WinRM target schemas,
allowing it to be passed to the API. This will let the key be used
internally within orchestrator without worrying about whether it gets
passed to bolt-server.

This key, along with the existing special 'hostname' and
'private-key-content' keys, will automatically be ignored by Bolt's
transport config handling. However, for clarity we now explicitly filter
the options we pass to the config to include only those that are
accepted by the transport. While not strictly necessary, this makes it
more obvious why those keys don't cause a problem and it provides an
extra layer of certainty in case the underlying implementation shifts in
the future.

### DIFF
--- a/lib/bolt_server/transport_app.rb
+++ b/lib/bolt_server/transport_app.rb
@@ -441,7 +441,7 @@ module BoltServer
         'uri' => target_hash['hostname'],
         'config' => {
           'transport' => 'ssh',
-          'ssh' => opts
+          'ssh' => opts.slice(*Bolt::Config::Transport::SSH.options)
         }
       }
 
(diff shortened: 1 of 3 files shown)


ekinanp (approved):

lib/bolt_server/transport_app.rb
```diff
@@ -479,7 +479,7 @@ def make_winrm_target(target_hash)
         'uri' => target_hash['hostname'],
         'config' => {
           'transport' => 'winrm',
-          'winrm' => opts
+          'winrm' => opts.slice(*Bolt::Config::Transport::WinRM.options)
```
ekinanp, line 482: What does this do? The `slice` I mean.

nicklewis: It's equivalent to `select-keys` in clojure. It returns the subset of the hash that corresponds to the keys in the array (ignoring keys that aren't present in the hash).

beechtom: It's also the best method ever created.

donoghuc (approved):

PR #2937, opened 2021-07-16
(hack) Convert guides to YAML and pretty print

lucywyman:
When we first implemented the `guide` subcommand, we chose to make the
file format of guides flat text and documented how users could add their
own guides to Bolt to be loaded by the `bolt guide` subcommand. While
useful, we haven't seen anyone use this functionality, and further it
does not seem like the file format has much impact on whether users add
their own guides or not. This converts the file format from flat text to
YAML, which supports both being read as data and formatting multiline
strings. This lets us format the `bolt guide <guide>` output to be more
readable and inline with other bolt output, make the guide text easily
editable (as opposed to JSON), and still lets users add their own guides
if they choose.

!feature

* **Beautify `bolt guide` output** (hack)
  This converts our existing guides from flat text to YAML so that they
  can be pretty printed.


beechtom:
Since these now need to be parsed before printing, should the guides just be data in Bolt instead of separate files?

lucywyman:
I don't think so - this still allows people to add their own guides if they'd like, plus having so much text as code is kind of awful, plus having these be JSON makes them more difficult to format so keeping them as YAML is preferable. I don't see any advantage to changing where they're loaded from.

guides/guide.txt
```diff
@@ -1,17 +0,0 @@
-TOPIC
```
beechtom, line 1: RIP 💀

guides/debugging.yaml
```diff
@@ -1,10 +1,9 @@
...
-    debugging
-
-DESCRIPTION
+---
+topic: debugging
+guide: |
     When Bolt isn't behaving as expected, there are a few helpful commands and
     logs that can help identify common issues. The first place to look is in
-    `<PROJECT>/bolt-debug.log`, which contains debug-level logs from the last Bolt
+    `<PROJECT|/bolt-debug.log`, which contains debug-level logs from the last Bolt
```
beechtom, line 6: Why the change to a pipe?

guides/logging.yaml
```diff
@@ -1,18 +1,16 @@
...
+guide: |
     Bolt prints messages both to the console and to log files. Messages can
     either come from Bolt's 'outputter', which logs user-facing messages like
     progress and results, or from the 'logger', which logs warnings, errors, and
     log-structured output to log files. Both of these message streams are
     configurable.
 
     By default, Bolt logs to the console at 'warn' level and writes a log file to
-    '<project>/bolt-debug.log' at 'debug' level. Unless you are running a plan,
+    '<project|/bolt-debug.log' at 'debug' level. Unless you are running a plan,
```
beechtom, line 10: Same here; why the change to a pipe?

guides/module.yaml
```diff
@@ -14,6 +13,5 @@ DESCRIPTION
     
     To learn more about managing modules in a project, see the documentation.
     To learn how modules are loaded by Bolt, see the 'modulepath' guide.
-
-DOCUMENTATION
-    https://pup.pt/bolt-modules
+documentation:
```
beechtom, line 16:
```suggestion

documentation:
```

Just a small nit for consistency, since the rest of the guides have an extra newline before documentation. 😬

guides/targets.yaml
```diff
@@ -16,7 +15,7 @@ DESCRIPTION
          --rerun   FILTER       -Rerun        FILTER
 
     The 'targets' option accepts a comma-separated list of target URIs or group
-    names, or can read a target list from an input file '@<file>' or stdin '-'.
+    names, or can read a target list from an input file '@<file|' or stdin '-'.
```
beechtom, line 18: Why a pipe?

guides/transports.yaml
```diff
@@ -1,13 +1,12 @@
...
+---
+topic: transports
+guide: |
     Bolt uses transports (also known as protocols) to establish a connection
     with a target in order to run actions on the target. The default transport is
     SSH, and you can see available transports along with their configuration
     options and defaults at http://pup.pt/bolt-reference.
 
-    You can specify a transport for a target by prepending '<transport>://' to
+    You can specify a transport for a target by prepending '<transport|://' to
```
beechtom, line 9: Pipe?

rakelib/pwsh.rake
```diff
@@ -6,16 +6,33 @@ require 'fileutils'
...
+        # Make sure both topic and guide keys are defined
+        unless (%w[topic guide] - info.keys).empty?
+          raise "Guide file #{file} must have a 'topic' key and 'guide' key, but has #{info.keys} keys."
+        end
+
+        txt = +"#{info['topic']}\n"
+        txt << info['guide'].gsub(/^/, '  ')
+
+        if info['documentation']
+          txt << "\nDocumentation"
```
beechtom, line 28:
```suggestion
          txt << "\nDocumentation\n"
```

guides/debugging.yaml
lucywyman: Oh, ha, definitely a typo.

guides/logging.yaml
lucywyman: I think I used a sed command to find + replace that must have modified these....

guides/guide.txt
lucywyman: Gotta think of a better one :P

guides/logging.yaml
```diff
@@ -1,7 +1,5 @@
-TOPIC
-    logging
-
-DESCRIPTION
+topic: logging
```
beechtom, line 1:
```suggestion
---
topic: logging
```

beechtom (approved):

PR #2941, opened 2021-07-19
(hack) Format PlanResult if it's a Bolt datatype

lucywyman:
Bolt will format a PlanResult as JSON when we don't have a handler for
the class of the result. Despite having functions for printing
`Bolt::Result`, `Bolt::ApplyResult`, and `Bolt::Error` types, the plan
result printer in the human outputter would not use these functions and
would format the types as JSON, which often makes them difficult to
read. Bolt will now correctly format the result types if returned from a
plan.

!feature

* **Format PlanResults if they are Bolt datatypes** (hack)

  Bolt will now print a more human readable message for plan results
  that are Bolt datatypes when using the human output format, rather than
  printing JSON.

### DIFF
--- a/lib/bolt/outputter/human.rb
+++ b/lib/bolt/outputter/human.rb
@@ -665,6 +665,12 @@ module Bolt
           print_container_result(value.result)
         when Bolt::ResultSet
           print_result_set(value)
+        when Bolt::Result
+          print_result(value)
+        when Bolt::ApplyResult
+          print_apply_result(value)
+        when Bolt::Error
+          print_bolt_error(**value.to_h.transform_keys(&:to_sym))
         else
           @stream.puts(::JSON.pretty_generate(plan_result, quirks_mode: true))
         end
@@ -702,6 +708,17 @@ module Bolt
 
       def print_error(message)
         @stream.puts(colorize(:red, message))
+      end
+
+      def print_bolt_error(msg:, details:, **_kwargs)
+        err = msg
+        if (f = details[:file])
+          err += "\n  (file: #{f}"
+          err += ", line: #{details[:line]}" if details[:line]
+          err += ", column: #{details[:column]}" if details[:column]
+          err += ")"
+        end
+        @stream.puts(colorize(:red, err))
       end
 
       def print_prompt(prompt)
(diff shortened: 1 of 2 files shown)


lib/bolt/outputter/human.rb
```diff
@@ -665,6 +665,12 @@ def print_plan_result(plan_result)
           print_container_result(value.result)
         when Bolt::ResultSet
           print_result_set(value)
+        when Bolt::Result
+          print_result(value)
+        when Bolt::ApplyResult
+          print_apply_result(value)
+        when Bolt::Error
+          print_error(value)
```
beechtom, line 673: `print_error` only prints the message, which means we lose the error location under `details`. I think this info is helpful for users and should be available using human output.

lucywyman: Hmm, I think often times that info is already in the message:
```
lucy ~/githubs/bolt (hack/better-output) $ be bolt plan run bolt::error
Starting: plan bolt::error
Finished: plan bolt::error in 0.0 sec
Evaluation Error: Unknown variable: 'target'. (file: /home/lucy/githubs/bolt/plans/error.pp, line: 4, column: 30)
  (file: /home/lucy/githubs/bolt/plans/error.pp, line: 4, column: 30)
```

But can't hurt to have it twice.

lib/bolt/outputter/human.rb
```diff
@@ -704,6 +710,17 @@ def print_error(message)
         @stream.puts(colorize(:red, message))
       end
 
+      def print_bolt_error(msg:, details:, **_kwargs)
+        err = msg
+        if (f = details[:file])
+          err << "\n  (file: #{f}"
+          err << ", line: #{details[:line]}" if details[:line]
+          err << ", column: #{details[:column]}" if details[:column]
+          err << ")"
```
beechtom, line 719:
```suggestion
          err += "\n  (file: #{f}"
          err += ", line: #{details[:line]}" if details[:line]
          err += ", column: #{details[:column]}" if details[:column]
          err += ")"
```

beechtom (changes requested):

lib/bolt/outputter/human.rb
nicklewis: or `err = msg.dup`

beechtom (approved):

PR #2947, opened 2021-07-21
(bug) Do not warn about analytics when analytics is disabled

beechtom:
This fixes a bug that was causing Bolt to warn about analytics
collection when analytics is disabled in a config file and the
`analytics.yaml` file does not exist.

!bug

* **Do not warn about analytics when analytics is disabled**

  Bolt no longer displays a warning about analytics collection when
  analytics is disabled and the `analytics.yaml` file does not exist.


lucywyman (approved):
:+1: Reproduced the issue and verified the fix.

PR #2948, opened 2021-07-22
(maint) Remove unsupported platforms

lucywyman:
The release engineering team is removing vanagon support for Fedora 30,
Fedora 31, and MacOS 10.14, in addition to removing support for the Puppet
Agent. This means that Bolt can no longer build packages for these
platforms, so is removing them.

!removal

* **Remove support for Fedora 30, Fedora 31, and MacOS 10.14**

### DIFF
--- a/documentation/bolt_installing.md
+++ b/documentation/bolt_installing.md
@@ -10,8 +10,8 @@ and Microsoft Windows.
 | Operating system          | Versions            |
 | ------------------------- | ------------------- |
 | Debian                    | 9, 10               |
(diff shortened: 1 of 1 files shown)


beechtom (changes requested):
Can you update the release note to say "macOS 10.14" instead of "OSX 10.14"?

beechtom (approved):

beechtom (changes requested):
The release note tag for this should be `!removal`

hestonhoffman (approved):

beechtom (approved):

PR #2950, opened 2021-07-23
(GH-2889) Temporarily override scope when resuming future

beechtom:
This updates `PlanFuture` instances to keep track of their own scope
object. When a `PlanFuture` is resumed by the `FiberExecutor`, the
`global_scope` in Puppet is temporarily overridden with the `PlanFuture`
scope. This prevents ephemerals from being popped off in the wrong order
due to race conditions that arise from invoking custom Puppet language
functions that invoke Bolt executor functions.

!bug

* **Do not remove variables from scope when using `parallelize()`**
  ([#2889](https://github.com/puppetlabs/bolt/issues/2889))

  Bolt no longer removes variables from a plan's scope when invoking
  a custom Puppet language function that invokes a `run_*` function from
  a `parallelize()` block. Previously, a race condition would result in
  the current plan's variables to be removed from the plan's scope.

### DIFF
--- a/lib/bolt/fiber_executor.rb
+++ b/lib/bolt/fiber_executor.rb
@@ -52,7 +52,7 @@ module Bolt
       # tracking which Futures to wait on when `wait()` is called without
       # arguments.
       @id += 1
-      future = Bolt::PlanFuture.new(future, @id, name: name, plan_id: plan_id)
+      future = Bolt::PlanFuture.new(future, @id, name: name, plan_id: plan_id, scope: newscope)
       @logger.trace("Created future #{future.name}")
 
       # Register the PlanFuture with the FiberExecutor to be executed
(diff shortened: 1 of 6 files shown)


lucywyman:
This testing seems very minimal, did you run manual tests with nested Bolt plans with and without parallelize blocks to verify that scoping doesn't misbehave?

lucywyman (approved):
- [x] Variables from parent plan aren't available in subplans
- [x] Variables are "restored" after switching Futures
- [x] Variables from subplans aren't available in parent plans
- [x] Variables from other Futures aren't available
- [x] Runs a plan from the PE console
- [x] Restores variables after running a Ruby plan function
- [x] Restores variables after running a Puppet plan function
- [x] Doesn't regress on ignoring global_scope modifications in Ruby Puppet functions (this already errors)

Looking at Puppet to see what might be using or modifying the global scope in ways that this might impact, Puppet language and Ruby Puppet functions are my main concerns, which are definitely improved with this PR. This can also potentially prevent other scoping issues, given Puppet isn't expecting multithreaded use, and having the Fibers manage their own scopes circumvents that.

Issue #2925, opened 2021-07-12
add _catch_errors handler to bolt apply command


vchepkov:
## Use Case
Currently, bolt apply command fails if any of the targets in inventory are not available, for example
```
$ bolt apply -e "notify { 'test': }" -t linux 2>&1|head
Starting: install puppet and gather facts on 362 targets
Plan aborted: run_task 'apply_helpers::custom_facts' failed on 2 targets
```

## Describe the Solution You Would Like
This happens because under the hood apply is running the agent_install and fact_collecting tasks, but to the user apply is a single operation. This should default to always continuing to execute the full 'operation' if targets fail for any of the tasks, and include failed targets and their errors in the result. 

For the results, we should save any errors from running the tasks and then merge them with the resultset from the apply so that there's one result object.

PR #2953, opened 2021-07-26
(GH-2925) Support `_catch_errors` in `apply_prep`

beechtom:
This adds support for using the `_catch_errors` metaparameter in the
`apply_prep` plan function. When set, the function will return a result
set that includes results for all targets that successfully installed
targets and retrieved custom facts, and any errors for targets that were
not able to do both.

This also updates the `bolt apply` and `Invoke-BoltApply` commands to
set `_catch_errors` when running the `apply_prep` function. This allows
the command to complete the apply on as many targets as possible instead
of erroring if even one target fails the `apply_prep` step.

!feature

* **Support `_catch_errors` in `apply_prep`**
  ([#2925](https://github.com/puppetlabs/bolt/issues/2925))

  The `apply_prep` function now accepts the `_catch_errors` parameter.
  Additionally, the `bolt apply` and `Invoke-BoltApply` commands set
  `_catch_errors` to `true`, allowing an apply action to complete on as
  many targets as possible instead of erroring if even one target fails
  the `apply_prep` portion of the command.

### DIFF
--- a/bolt-modules/boltlib/lib/puppet/functions/apply_prep.rb
+++ b/bolt-modules/boltlib/lib/puppet/functions/apply_prep.rb
@@ -8,62 +8,23 @@ require 'bolt/task'
 # installed using either the configured plugin or the `task` plugin with the
 # `puppet_agent::install` task.
 #
-# Agent installation will be skipped if the target includes the `puppet-agent` feature, either as a
-# property of its transport (PCP) or by explicitly setting it as a feature in Bolt's inventory.
+# Agent installation will be skipped if the target includes the `puppet-agent`
+# feature, either as a property of its transport (PCP) or by explicitly setting
+# it as a feature in Bolt's inventory.
 #
 # > **Note:** Not available in apply block
 Puppet::Functions.create_function(:apply_prep) do
   # @param targets A pattern or array of patterns identifying a set of targets.
   # @param options Options hash.
+  # @option options [Boolean] _catch_errors Whether to catch raised errors.
   # @option options [Array] _required_modules An array of modules to sync to the target.
   # @option options [String] _run_as User to run as using privilege escalation.
-  # @return [nil]
+  # @return [Bolt::ResultSet]
   # @example Prepare targets by name.
   #   apply_prep('target1,target2')
   dispatch :apply_prep do
     param 'Boltlib::TargetSpec', :targets
     optional_param 'Hash[String, Data]', :options
-  end
-
-  def script_compiler
-    @script_compiler ||= Puppet::Pal::ScriptCompiler.new(closure_scope.compiler)
-  end
-
-  def inventory
-    @inventory ||= Puppet.lookup(:bolt_inventory)
-  end
-
-  def get_task(name, params = {})
-    tasksig = script_compiler.task_signature(name)
-    raise Bolt::Error.new("Task '#{name}' could not be found", 'bolt/apply-prep') unless tasksig
-
-    errors = []
-    unless tasksig.runnable_with?(params) { |msg| errors << msg }
-      # This relies on runnable with printing a partial message before the first real error
-      raise Bolt::ValidationError, "Invalid parameters for #{errors.join("\n")}"
-    end
-
-    Bolt::Task.from_task_signature(tasksig)
-  end
-
-  # rubocop:disable Naming/AccessorMethodName
-  def set_agent_feature(target)
-    inventory.set_feature(target, 'puppet-agent')
-  end
-  # rubocop:enable Naming/AccessorMethodName
-
-  def run_task(targets, task, args = {}, options = {})
-    executor.run_task(targets, task, args, options)
-  end
-
-  # Returns true if the target has the puppet-agent feature defined, either from inventory or transport.
-  def agent?(target, executor, inventory)
-    inventory.features(target).include?('puppet-agent') ||
-      executor.transport(target.transport).provided_features.include?('puppet-agent') || target.remote?
-  end
-
-  def executor
-    @executor ||= Puppet.lookup(:bolt_executor)
   end
 
   def apply_prep(target_spec, options = {})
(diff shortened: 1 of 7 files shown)


bolt-modules/boltlib/lib/puppet/functions/apply_prep.rb
```diff
@@ -8,155 +8,206 @@
...
-# property of its transport (PCP) or by explicitly setting it as a feature in Bolt's inventory.
+# Agent installation will be skipped if the target includes the `puppet-agent`
+# feature, either as a property of its transport (PCP) or by explicitly setting
+# it as a feature in Bolt's inventory.
 #
 # > **Note:** Not available in apply block
 Puppet::Functions.create_function(:apply_prep) do
   # @param targets A pattern or array of patterns identifying a set of targets.
   # @param options Options hash.
   # @option options [Array] _required_modules An array of modules to sync to the target.
```
lucywyman, line 19: Add `_catch_errors`

bolt-modules/boltlib/lib/puppet/functions/apply_prep.rb
```diff
@@ -8,155 +8,206 @@
...
     @inventory ||= Puppet.lookup(:bolt_inventory)
   end
 
+  def script_compiler
+    @script_compiler ||= Puppet::Pal::ScriptCompiler.new(closure_scope.compiler)
+  end
+
+  # Retrieves a task by name. This method is called by the puppet_library hook.
+  #
   def get_task(name, params = {})
```
lucywyman, line 68: https://github.com/puppetlabs/bolt/commit/f4db81f56d6cbee8aa3b04554f886de6d6754327 I don't remember *why* we initially had to load the task here instead of using `pal.get_task` besides for nicer error messages, but I think it'd make more sense to:
* Call `pal.get_task` in the puppet_library plugin
* Call `tasksig.runnable_with?(params)` there as well

The PluginExecution error already includes the information that this is running as a plugin, so better error messaging doesn't seem necessary, and then we're not duplicating this code with `pal.get_task`.

bolt-modules/boltlib/lib/puppet/functions/apply_prep.rb
beechtom: I swear I added this 💀

lib/bolt/plugin/task.rb
```diff
@@ -59,7 +59,7 @@ def puppet_library(opts, target, apply_prep)
         run_opts = {}
         run_opts[:run_as] = opts['_run_as'] if opts['_run_as']
         begin
-          task = apply_prep.get_task(opts['task'], params)
+          task = @context.get_validated_task(opts['task'], params)
```
lucywyman, line 62: `apply_prep.run_task` also seems like it can just use `@context.run_task`

bolt-modules/boltlib/lib/puppet/functions/apply_prep.rb
```diff
@@ -8,155 +8,188 @@
...
+    executor.report_function_call(self.class.name)
 
-    errors = []
-    unless tasksig.runnable_with?(params) { |msg| errors << msg }
-      # This relies on runnable with printing a partial message before the first real error
-      raise Bolt::ValidationError, "Invalid parameters for #{errors.join("\n")}"
+    executor.log_action('install puppet and gather facts', targets) do
+      executor.without_default_logging do
+        install_results = install_agents(targets, options)
+        facts_results   = get_facts(install_results.ok_set.targets, options)
```
lucywyman, line 44: This also needs to collect facts for targets that already had the agent installed (`agent_targets`), not just the targets we successfully installed on.

lib/bolt/plugin/task.rb
beechtom: There's no `PluginContext.run_task` method, and I don't think writing one is really in the scope of this ticket.

bolt-modules/boltlib/lib/puppet/functions/apply_prep.rb
beechtom: It does. `install_results` includes successful results for targets with agents installed: https://github.com/puppetlabs/bolt/pull/2953/files#diff-d0014ca5267115a969a290f0cdf83b618e6e87d085f5a7229f459ca74ad0af8aR105

bolt-modules/boltlib/lib/puppet/functions/apply_prep.rb
```diff
@@ -8,155 +8,188 @@
...
+  end
+
+  # Return a list of targets and their puppet_library hooks.
+  #
+  private def get_hooks(targets, options)
+    hooks  = []
+    errors = []
+
+    targets.each do |target|
+      plugin_opts = target.plugin_hooks&.fetch('puppet_library').dup
```
lucywyman, line 164:
```suggestion
      plugin_opts = target.plugin_hooks.fetch('puppet_library').dup
```
We definitely don't handle this being nil, so...it must never be nil (and it does get set to empty hash in `inventory/target.rb`).

bolt-modules/boltlib/lib/puppet/functions/apply_prep.rb
lucywyman: Ah, ok.

bolt-modules/boltlib/spec/functions/apply_prep_spec.rb
```diff
@@ -60,36 +60,18 @@
       Puppet::Pal::ScriptCompiler.any_instance.stubs(:task_signature).with('service').returns(task3)
     end
 
-    it 'sets feature and gathers facts' do
```
lucywyman, line 63: Am I reading this right, did we literally just have the same test here twice?

lib/bolt/plugin/task.rb
lucywyman: Oh, gosh, sorry I was looking at `Bolt::Task::Run.run_task()`, which we can use here. I don't think it matters that much, it just seems odd to have so many functions that just wrap `executor.run_task`. Reduce reuse recycle, etc.

bolt-modules/boltlib/spec/functions/apply_prep_spec.rb
beechtom: Yep

bolt-modules/boltlib/spec/functions/apply_prep_spec.rb
```diff
@@ -282,13 +264,51 @@
...
+    before(:each) do
+      applicator.stubs(:build_plugin_tarball).returns(:tarball)
+      applicator.stubs(:custom_facts_task).returns(custom_facts_task)
+
+      plugins.expects(:get_hook)
+             .with("puppet_agent", :puppet_library)
+             .returns(task_hook)
+    end
+
+    context 'with failing hook' do
```
lucywyman, line 293: Can this include a test that results are merged as well?

beechtom: I'm having trouble stubbing things and testing the return result, since rspec-puppet only lets you compare equality for the return value. :/

lib/bolt/plugin/task.rb
beechtom: Hrm... to use `Bolt::Task::Run.run_task()` here we'd need to pass the executor to `Bolt::Plugin::Module.puppet_library` and `Bolt::Plugin::Task.puppet_library`. At that point why not just call `executor.run_task` directly?

lucywyman (dismissed):
## apply_prep
- [x] Catches errors when `_catch_errors` is set 
- [x] Does not catch errors when `_catch_errors` is false
- [x] Reports the function call to analytics
- [x] Combines install_agent and custom_facts result sets
- [x] Does not run the install task if `puppet-agent` feature is set 
- [x] Does not run the install task if using the PCP transport
- [x] Doesn't need to be updated in pe-bolt-modules
- [x] `_catch_errors` overrides plugin configuration
- [x] Continues to collect facts on successful nodes if install fails on some nodes

## apply
- [x] Always runs with `catch_errors`

bolt-modules/boltlib/spec/functions/apply_prep_spec.rb
lucywyman: Maybe an integration test that runs against a nonexistent target and a real target, and ensures both results are there.

beechtom: D'oh. Added. :P

lucywyman (approved):

PR #2954, opened 2021-07-26
(maint) Update bolt to version 3.14.1

artlawson:


beechtom (changes requested):
This needs to be rebased on `main`.

beechtom (approved):

Issue #2945, opened 2021-07-21
Bolt doesn't honor the `@api private` docstring


reidmv:
## Use Case

From the [Bolt docs](https://puppet.com/docs/bolt/latest/writing_plans.html#:~:text=unlike%20tasks%2C%20plans%20do%20not%20have%20a%20corresponding%20metadata.json%20file.%20instead%2C%20bolt%20pulls%20documentation%20directly%20from%20the%20plan%20using%20puppet%20types%20and%20puppet%20strings.):

> Bolt pulls documentation directly from the plan using Puppet Types and Puppet Strings.

Puppet Strings has used [`# @api private` docstring](https://puppet.com/docs/puppet/7/puppet_strings.html#puppet_strings_reference-strings-tags) to designate Puppet code elements as private for a long time. Bolt has the concept of private plans, but currently only respects a [different docstring `# @private true`](https://puppet.com/docs/bolt/latest/writing_plans.html#making-plans-private) to designate them as such.

Because Bolt is a consumer/user of the Puppet DSL, and documented to use Puppet Strings for its inline plan documentation, Bolt should honor the `# @api private` tag.

## Describe the Solution You Would Like

Bolt should check for and respect `# @api private` the same way it does `# @private true`.

Note that the real pain point to solve for is that users of Puppet Strings, and people who read the Puppet Strings documentation, should be able to use the tags documented there. An alternative approach would be to change Puppet Strings to use, and document, `# @private true`, but that's a huge lift and changing the more established Puppet Strings tool would be highly disruptive.

It would have been best to align this from the beginning, but A) ship sailed, and B) speculating, there was probably a UX desire to use a boolean flag to match how Task metadata works, and to make the switch cleaner in the Yaml implementation of plans. Given that, the cleanest fix at this point is probably to keep what we have, but silently honor `# @api private` as equivalent to `# @private true` when encountered in the DSL.

## Solution
We can start supporting both `@api private` and `@private true` now
- Update the documentation to prefer `@api private` to `@private true`
- Issue a deprecation warning for `@private`
- As part of Bolt 4.0, remove `@private` flag

PR #2955, opened 2021-07-27
(GH-2945) Respect `@api` strings tag, deprecate `@private` tag

beechtom:
This adds support for using the `@api private|false` tag in plans and
deprecates the `@private false|true` tag. Originally, plans were marked
as private using the `@private` tag, but this created differing behavior
from what is established and documented by puppet-strings.

When a plan uses the `@private` tag, Bolt will now issue a deprecation
warning, and suggest the user use the `@api` tag instead.

If both tags are present, Bolt will consider the plan private if either
`@api private` or `@private true` are present. This is because
puppet-strings sets the `@api` tag to `public` by default, whether it's
actually present in the plan or not.

!feature

* **Respect `@api` tag in Puppet language plans**
  ([#2945](https://github.com/puppetlabs/bolt/issues/2945))

  Bolt now respects the `@api` tag in Puppet language plans. If the
  `@api private` tag is present, the plan will be hidden from `bolt plan
  show` and `Get-BoltPlan` output.

!deprecation

* **Deprecate `@private` tag in Puppet language plans**
  ([#2945](https://github.com/puppetlabs/bolt/issues/2945))

  The `@private` tag for Puppet language plans is deprecated. Use the
  `@api` tag instead.


lib/bolt/pal.rb
```diff
@@ -554,14 +549,39 @@ def get_plan_info(plan_name, with_mtime: false)
...
+    #
+    # @param plan [PuppetStrings::Yard::CodeObjects::Plan] The puppet-strings plan documentation.
+    # @return [Boolean]
+    #
+    private def private_plan?(plan)
+      if plan.tag(:private)
+        value     = plan.tag(:private).text
+        api_value = value.downcase == 'true' ? 'private' : 'public'
+
+        Bolt::Logger.deprecate(
```
lucywyman, line 571:
```suggestion
        Bolt::Logger.deprecate_once(
```

beechtom: This would mean only the first plan to have a `@private` tag would get a warning. I think having a separate warning for each plan is helpful so you know which ones to make changes to.

beechtom: Also realizing that the plan cache means you won't see these warnings unless the plans are modified...

beechtom: That's because of a bug in `file_modified?` that I just pushed a fix for. The mtime from the plan data can be either a string if loaded from the cache or a time object if loaded from `get_plan_info`. When the file's current mtime was compared to the cache mtime, the comparison would be falsey if the mtime came from `get_plan_info` since the class is different.

lucywyman (approved):

Issue #2922, opened 2021-07-08
Provide a way to not leak bolt runner ruby environment to tasks run over local transport


donoghuc:
## Use Case
When I run a ruby task over the local transport I dont want to necessarily run under the same context as the ruby env running bolt. 

## Describe the Solution You Would Like
When i disable [bundled-ruby](https://puppet.com/docs/bolt/latest/bolt_transports_reference.html#bundled-ruby) i do not want any ruby specific environment variables set when running the bolt process (maybe something like:
```
unset GEM_HOME
unset GEM_PATH
unset DLN_LIBRARY_PATH
unset RUBYLIB
unset RUBYLIB_PREFIX
unset RUBYOPT
unset RUBYPATH
unset RUBYSHELL
```
)

## Describe Alternatives You've Considered
Does this generalizes to being able to configure environment variables when running a task. For example maybe something like is done for interpreters:
```
interpreters:
  rb: /my/ruby
  env:
    GEM_HOME: /some/thing
```

## Additional Context
INnthe context of running a remote task over ace, we do not want to use the settings for GEM_PATH and GEM_HOME used to start the ace process. This is a specific issue within general weirdness executing ruby tasks over the local transport given the bolt runner will also be a ruby process. 

Also, somewhat related is https://github.com/puppetlabs/puppet/pull/8685 which is another approach at solving this particular issue in ACE>

lucywyman:
Hmm, I'm having some trouble reproducing this in just Bolt. We [use Open3.popen3](https://ruby-doc.org/stdlib-2.7.2/libdoc/open3/rdoc/Open3.html#method-c-popen3) without any additional environment variables to open the subprocess, which should just preserve the existing user environment variables including GEM_HOME and GEM_PATH (even though we unset those in the [Bolt bin wrapper](
https://github.com/puppetlabs/bolt-vanagon/blob/main/resources/files/posix/bolt_env_wrapper) popen3 seems to reset to the user environment before adding those). The only place I can see at least GEM_PATH get set is [in the actual pe-plan-executor service definition](https://github.com/puppetlabs/bolt-vanagon/blob/main/resources/systemd/pe-plan-executor.service). Am I missing something?

donoghuc:
The values of GEM_* we use to start the service that you linked to are the issue. We explicitly do not want those set when ruining the remote ruby tasks. 

It would be surprising to me if 
> which should just preserve the existing user environment variables including GEM_HOME and GEM_PATH
[...]

PR #2956, opened 2021-07-27
(GH-2922) Attempt to restore Ruby environment on local

lucywyman:
When running over the `local` transport with `bundled-ruby` set to
false, Bolt will attempt to restore the original Ruby environment
variables, specifically:
* GEM_PATH
* GEM_HOME
* RUBYLIB
* RUBYLIB_PREFIX
* RUBYOPT
* RUBYPATH
* RUBYSHELL

The Bolt binary wrapper explicitly unsets these variables to provide a
clean environment for the Bolt process to run in, which affects running
Bolt from a system package, and Bundler can also modify these
environment variables, saving their original values to
`BUNDLER_ORIG_<env_var>`. Changes to the Ruby environment that Bolt runs
in are incidentally picked up by Ruby tasks and plans running over the
local transport, which opens a new subprocess to execute. The binary
wrapper now stores those environment variables in `BOLT_ORIG_<env_var>`
environment variables, and when executing with `bundled-ruby` explicitly set
to false (not unconfigured), Bolt will attempt to restore the
environment variables first from `BOLT_ORIG_<env_var>`, then from
`BUNDLER_ORIG_<env_var>`.

!bug

* **Attempt to restore Ruby environment on local with bundled-ruby false** ([#2922](https://github.com/puppetlabs/bolt/issues/2922))
  When executing, Bolt will unset several Ruby environment variables in
  it's own environment to avoid unexpected behavior. This can interfere
  with running Ruby on the local transport with Bolt though. Bolt will now
  attempt to restore Ruby environment variables like `GEM_HOME` and
  `GEM_PATH` when running over the local transport with `bundled-ruby` set
  to false.

### DIFF
--- a/lib/bolt/transport/local/connection.rb
+++ b/lib/bolt/transport/local/connection.rb
@@ -10,6 +10,8 @@ module Bolt
   module Transport
     class Local < Simple
       class Connection
+        RUBY_ENV_VARS = %w[GEM_PATH GEM_HOME RUBYLIB RUBYLIB_PREFIX RUBYOPT RUBYPATH RUBYSHELL].freeze
+
         attr_accessor :user, :logger, :target
 
         def initialize(target)
@@ -68,7 +70,21 @@ module Bolt
             end
           end
 
-          Open3.popen3(*command)
+          # Only do this if bundled-ruby is set to false, not nil
+          ruby_env_vars = if target.transport_config['bundled-ruby'] == false
+                            RUBY_ENV_VARS.each_with_object({}) do |e, acc|
+                              acc[e] = ENV["BOLT_ORIG_#{e}"] if ENV["BOLT_ORIG_#{e}"]
+                            end
+                          end
+
+          if target.transport_config['bundled-ruby'] == false &&
+             Gem.loaded_specs.keys.include?('bundler')
+            Bundler.with_unbundled_env do
+              Open3.popen3(ruby_env_vars || {}, *command)
+            end
+          else
+            Open3.popen3(ruby_env_vars || {}, *command)
+          end
         end
 
         # This is used by the Bash shell to decide whether to `cd` before
--- a/acceptance/Rakefile
+++ b/acceptance/Rakefile
@@ -9,7 +9,7 @@ extend Acceptance::BoltSetupHelper
 
 desc "Generate Beaker Host config"
 rototiller_task :host_config do |task|
-  unless ENV['BEAKER_HOSTS']
+  unless ENV['BEAKER_HOSTS_FILE']
     task.add_env(name: 'BOLT_CONTROLLER', default: 'debian10-64')
     task.add_env(name: 'BOLT_NODES',
                  default: 'centos7-64,osx1012-64,windows10ent-64')
--- a/acceptance/tests/task_local.rb
+++ b/acceptance/tests/task_local.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require 'bolt_command_helper'
+require 'bolt_setup_helper'
 
 test_name "bolt task run should execute tasks on localhost via local transport" do
   extend Acceptance::BoltCommandHelper
@@ -32,7 +33,7 @@ test_name "bolt task run should execute tasks on localhost via local transport" 
       assert_equal(profile_pre.stdout, profile_post.stdout, 'Profile was loaded')
     end
   else
-    step "create task on bolt controller" do
+    step "create whoami task on bolt controller" do
       on(bolt, "mkdir -p #{dir}/modules/test/tasks")
       create_remote_file(bolt, "#{dir}/modules/test/tasks/whoami_nix", <<-FILE)
       #!/bin/sh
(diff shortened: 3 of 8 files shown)


lib/bolt/transport/local/connection.rb
```diff
@@ -68,7 +68,10 @@ def execute(command)
             end
           end
 
-          Open3.popen3(*command)
+          # It honestly seems like we should just always do this.
+          Bundler.with_unbundled_env do
```
donoghuc, line 72: I think we would have to start shipping the bundler gem with bolt for this.

beechtom: Isn't bundler a default gem?

lucywyman: Gah didn't even think of that...I think I can wrap this in a check that Bundler is available and not do this if it's not, since it'll be moot in that case.

lucywyman:
Relies on https://github.com/puppetlabs/bolt-vanagon/pull/170

lib/bolt/transport/local/connection.rb
beechtom: Boo, looks like it's only default starting with ruby 2.6. :( https://stdgems.org/compare/

lucywyman:
Phew, got a successful acceptance run (with an actual value for GEM_PATH):
```
  * execute `bolt task run test::ruby_env` on localhost via local transport
    
    numb-separation.delivery.puppetlabs.net (numb-separation.delivery.puppetlabs.net) 10:50:08$ GEM_PATH=$(gem env GEM_PATHS) bolt task run test::ruby_env --targets localhost --modulepath /tmp/local_task.5LiGPf/modules
      Started on localhost...
      Finished on localhost:
        /root/.gem/ruby/2.5.0:/var/lib/gems/2.5.0:/usr/lib/x86_64-linux-gnu/rubygems-integration/2.5.0:/usr/share/rubygems-integration/2.5.0:/usr/share/rubygems-integration/all
      Ran on 1 target in 0.08 sec
    
    numb-separation.delivery.puppetlabs.net (numb-separation.delivery.puppetlabs.net) executed in 1.77 seconds
    
    numb-separation.delivery.puppetlabs.net (numb-separation.delivery.puppetlabs.net) 10:50:10$ gem env GEM_PATHS
      /root/.gem/ruby/2.5.0:/var/lib/gems/2.5.0:/usr/lib/x86_64-linux-gnu/rubygems-integration/2.5.0:/usr/share/rubygems-integration/2.5.0:/usr/share/rubygems-integration/all
    
    numb-separation.delivery.puppetlabs.net (numb-separation.delivery.puppetlabs.net) executed in 0.21 seconds
```

lib/bolt/shell/powershell.rb
```diff
@@ -167,6 +167,25 @@ def with_tmpdir
...
+            end
+            acc[e] ||= ""
+          end
+          options[:env_vars] ||= {}
+          # Let any user-defined env vars override our attempts to restore the
+          # Ruby environment
+          options[:env_vars] = ruby_env_vars.merge(options[:env_vars])
+        end
+        options
+      end
```
beechtom, line 187: Can this be a method on the base `Bolt::Shell` class so it can be reused? It looks like the bash shell uses `options[:environment]` over `options[:env_vars]` so that'd need to be updated, too.

lucywyman: Hmm, I'd honestly rather just remove this from PowerShell altogether. We don't unset these env vars in the MSI, only in *nix, so this is only for unsetting the bundled environment which I just realized should already be covered by `Bundler.without_bundled` in the local transport.

lib/bolt/shell/bash.rb
```diff
@@ -345,6 +345,23 @@ def execute(command, sudoable: false, **options)
         # together multiple commands into a single sh invocation
         commands = [inject_interpreter(options[:interpreter], command)]
 
+        # Only do this for the local transport, not if bundled-ruby is nil
+        if target.transport_config['bundled-ruby'] == false
+          ruby_env_vars = RUBY_ENV_VARS.each_with_object({}) do |e, acc|
+            acc[e] = ENV["BOLT_ORIG_#{e}"] if ENV["BOLT_ORIG_#{e}"]
```
donoghuc, line 351: Wont this "pollute" the ENV for the rest of the bolt process? I was thinking that the `Connection` would just set these when we shell out with Open3.

donoghuc: oh, i see, we are not actually setting these, just reading. Nevermind.

lucywyman: Er, this is merging the values of `BOLT_ORIG` with the user-specified environment variables to get prepended to the command itself, not setting it *in* the Bolt environment. Or am I misunderstanding what you mean?

I initially had it in `Open3`, but was testing with Bundler using the `GEM_PATH` env var which [it explicitly unsets unless otherwise configured](https://github.com/rubygems/rubygems/blob/39771027ac3727a3234118a9795c733323d2188d/bundler/lib/bundler.rb#L671), but I didn't figure that out until I had moved it to be merged with the user-specified environment variables (thinking open3 might be the issue). I didn't think it mattered when I put this up, but I realized this *will* set the environment variables when using `run-as` on the local transport which I don't think we want, so I'll move it back to be in the connection later this morning (it honestly makes more sense there anyway).

donoghuc: My first comment was a mistake. I thought we were *setting* ENV, but we are just reading from it. Sorry i was kind of half looking at this PR yesterday. I really want to get this in a build with ACE server but I wont have time to get that done this week. This general thing should totally work there though :)

beechtom (approved):
LGTM. 👍

Issue #2910, opened 2021-06-30
Investigate using `require_relative`


beechtom:
## Use Case
Puppet recently switched to using `require_relative` in lieu of `require`, which resulted in performance improvements when loading in Puppet. We should investigate whether doing the same in Bolt would result in performance improvements as well.

PR #2960, opened 2021-07-29
(GH-2910) Relativize  bolt paths

artlawson:
Use `require_relative` in place of `require` for Bolt performance improvements.

!no-release-note


beechtom (changes requested):
The paths in the `bolt_spec` and `bolt_server` files are incorrect. I think we should avoid changing these anyways and focus on just `bolt` files (unless @lucywyman has other thoughts?).

beechtom (changes requested):
This needs to be rebased off `main`.

donoghuc:
I would love it if somebody builds this into a PE install and runs the orch tests against it. Given the issues we saw with puppet moving to relative_require i want to double check there wont be strange code loading issues.

lucywyman:
Sweet, finally got a green run for this in the Orchestrator!

lib/bolt/secret.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
```
beechtom, line 1: Need to delete this file.

beechtom (changes requested):
1. This also needs to update paths in `lib/bolt/application.rb`.
2. The commit message has two release note tags. We should probably remove `!feature`.

beechtom (approved):

PR #2961, opened 2021-07-30
(docs) Fix ruby error in an example

igalic:
the divide-by-zero example uses `divisor` as bare word that's never
asigned. We fix the resulting:

```
div-by-zero.rb:14:in `rescue in <main>': undefined local variable or method `divisor' for main:Object (NameError)
```

and provide a sensible error should this case ever occur.

!no-release-note

### DIFF
--- a/documentation/writing_tasks.md
+++ b/documentation/writing_tasks.md
@@ -421,7 +421,7 @@ rescue ZeroDivisionError
   result[:_error] = { msg: "Cannot divide by zero",
                       # namespace the error to this module
                       kind: "puppetlabs-example_modules/dividebyzero",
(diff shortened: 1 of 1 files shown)


beechtom:
@igalic Thanks for bringing this to our attention. Looking at the code example, it looks like the intent was actually to have line 424 be:

```ruby
details: { divisor: params['divisor'] }
```

instead of:

```ruby
details: { divisor: divisor }
```

If you want, we can make this change in a separate PR, or you can make it to your PR. Let us know.

igalic:
@beechtom, done!

beechtom (approved):
@igalic Awesome, thank you! :)

PR #2962, opened 2021-07-30
(PE-32492, PE32496) Support for apply/apply_prep for plans sourced from projects

donoghuc:
This commit adds two new actions to bolt-server. The apply_prep action runs a configurable task to install an agent and the `apply_helpers::custom_facts` task. The apply action runs the `apply_helpers::apply_catalog` task. These actions support apply/apply_prep from project plans run via orchestrator. Given a plugin tarbal is calculated for every invocation, this commit adds caching for plugin tarbals. The caching is "easy" here because a versioned project is unique and will not change over time.

### DIFF
--- a/lib/bolt_server/transport_app.rb
+++ b/lib/bolt_server/transport_app.rb
@@ -43,6 +43,8 @@ module BoltServer
       transport-ssh
       transport-winrm
       connect-data
+      action-apply_prep
+      action-apply
     ].freeze
 
     # PE_BOLTLIB_PATH is intended to function exactly like the BOLTLIB_PATH used
@@ -118,12 +120,7 @@ module BoltServer
       end
     end
 
-    def run_task(target, body)
-      validate_schema(@schemas["action-run_task"], body)
-
-      task_data = body['task']
-      task = Bolt::Task::PuppetServer.new(task_data['name'], task_data['metadata'], task_data['files'], @file_cache)
-      parameters = body['parameters'] || {}
+    def task_helper(target, task, parameters)
       # Wrap parameters marked with '"sensitive": true' in the task metadata with a
       # Sensitive wrapper type. This way it's not shown in logs.
       if (param_spec = task.parameters)
@@ -140,6 +137,84 @@ module BoltServer
         next unless value.key?('_sensitive')
         value['_sensitive'] = value['_sensitive'].unwrap
       end
+    end
+
+    def run_task(target, body)
+      validate_schema(@schemas["action-run_task"], body)
+
+      task_data = body['task']
+      task = Bolt::Task::PuppetServer.new(task_data['name'], task_data['metadata'], task_data['files'], @file_cache)
+      task_helper(target, task, body['parameters'] || {})
+    end
+
+    def extract_install_task(target)
+      unless target.plugin_hooks['puppet_library']['task']
+        raise BoltServer::RequestError,
+              "Target must have 'task' plugin hook"
+      end
+      install_task = target.plugin_hooks['puppet_library']['task'].split('::', 2)
+      install_task << 'init' if install_task.count == 1
+      install_task
+    end
+
+    def apply_prep(target, body)
+      validate_schema(@schemas["action-apply_prep"], body)
+      plugins_tarball = if (tarball = @file_cache.get_cached_plugin_tarball(body['versioned_project'], 'fact_plugins'))
+                          tarball
+                        else
+                          new_tarball = build_project_plugins_tarball(body['versioned_project']) do |mod|
+                            search_dirs = []
+                            search_dirs << mod.plugins if mod.plugins?
+                            search_dirs << mod.pluginfacts if mod.pluginfacts?
+                            search_dirs
+                          end
+                          @file_cache.cache_plugin_tarball(body['versioned_project'], 'fact_plugins', new_tarball)
+                          new_tarball
+                        end
+      task_datas = in_bolt_project(body['versioned_project']) do |context|
+        ps_parameters = {
+          'versioned_project' => body['versioned_project']
+        }
+        install_task = extract_install_task(target.first)
+        {
+          'install_task_data' => pe_task_info(context[:pal], install_task[0], install_task[1], ps_parameters),
+          'facts_task_data' => pe_task_info(context[:pal], 'apply_helpers', 'custom_facts', ps_parameters)
+        }
+      end
+      task_data = task_datas['install_task_data']
+      task = Bolt::Task::PuppetServer.new(task_data['name'], task_data['metadata'], task_data['files'], @file_cache)
+      install_task_result = task_helper(target, task, target.first.plugin_hooks['puppet_library']['parameters'] || {})
+      return install_task_result unless install_task_result.ok
+      task_data = task_datas['facts_task_data']
+      task = Bolt::Task::PuppetServer.new(task_data['name'], task_data['metadata'], task_data['files'], @file_cache)
+      task_helper(target, task, { 'plugins' => plugins_tarball })
+    end
+
+    def apply(target, body)
+      validate_schema(@schemas["action-apply"], body)
+      plugins_tarball = if (tarball = @file_cache.get_cached_plugin_tarball(body['versioned_project'], 'all_plugins'))
+                          tarball
+                        else
+                          new_tarball = build_project_plugins_tarball(body['versioned_project']) do |mod|
+                            search_dirs = []
+                            search_dirs << mod.plugins if mod.plugins?
+                            search_dirs << mod.pluginfacts if mod.pluginfacts?
+                            search_dirs << mod.files if mod.files?
+                            type_files = "#{mod.path}/types"
+                            search_dirs << type_files if File.exist?(type_files)
+                            search_dirs
+                          end
+                          @file_cache.cache_plugin_tarball(body['versioned_project'], 'all_plugins', new_tarball)
+                          new_tarball
+                        end
+      task_data = in_bolt_project(body['versioned_project']) do |context|
+        ps_parameters = {
+          'versioned_project' => body['versioned_project']
+        }
+        pe_task_info(context[:pal], 'apply_helpers', 'apply_catalog', ps_parameters)
+      end
+      task = Bolt::Task::PuppetServer.new(task_data['name'], task_data['metadata'], task_data['files'], @file_cache)
+      task_helper(target, task, body['parameters'].merge({ 'plugins' => plugins_tarball }))
     end
 
     def run_command(target, body)
@@ -476,6 +551,8 @@ module BoltServer
       run_task
       run_script
       upload_file
+      apply
+      apply_prep
     ].freeze
 
     def make_ssh_target(target_hash)
@@ -499,7 +576,8 @@ module BoltServer
         'config' => {
           'transport' => 'ssh',
           'ssh' => opts.slice(*Bolt::Config::Transport::SSH.options)
-        }
+        },
+        'plugin_hooks' => target_hash['plugin_hooks']
       }
 
       inventory = Bolt::Inventory.empty
@@ -537,7 +615,8 @@ module BoltServer
         'config' => {
           'transport' => 'winrm',
           'winrm' => opts.slice(*Bolt::Config::Transport::WinRM.options)
-        }
+        },
+        'plugin_hooks' => target_hash['plugin_hooks']
       }
 
       inventory = Bolt::Inventory.empty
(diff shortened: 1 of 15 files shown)


spec/Dockerfile.puppetserver
```diff
@@ -1,4 +1,4 @@
-FROM puppet/puppetserver
+FROM docker.io/puppet/puppetserver:edge
```
donoghuc, line 1: depends on https://github.com/puppetlabs/puppetserver/pull/2571

spec/docker-compose.yml
```diff
@@ -55,3 +55,4 @@ services:
       - "8140:8140"
     volumes:
       - ./fixtures/modules:/etc/puppetlabs/code/modules
+      - ./fixtures/bolt_server/projects:/etc/puppetlabs/code/projects
```
donoghuc, line 58: depends on https://github.com/puppetlabs/puppetserver/pull/2571

lib/bolt_server/transport_app.rb
```diff
@@ -142,6 +139,84 @@ def run_task(target, body)
...
+              "Target must have 'task' plugin hook"
+      end
+      install_task = target.plugin_hooks['puppet_library']['task'].split('::', 2)
+      install_task << 'init' if install_task.count == 1
+      install_task
+    end
+
+    def apply_prep(target, body)
+      validate_schema(@schemas["action-apply_prep"], body)
+      plugins_tarball = if (tarball = @file_cache.get_cached_plugin_tarball(body['versioned_project'], 'fact_plugins'))
```
ekinanp, line 162: Since this logic is also duplicated in `apply`, I think it would be good to add a `build_cached_project_plugins_tarball`-esque helper that centralizes this stuff.

ekinanp: Also, I think the code here risks running into weird "simultaneous file-write" edge-cases resulting from at least two threads trying to write to the cache. Fortunately, file_cache has a nice pattern for that via serial_execute (see e.g. https://github.com/puppetlabs/bolt/blob/main/lib/bolt_server/transport_app.rb#L182). Reading the linked example, I think wrapping all of this plugin-tarball-computation code up in a `file_cache.serial_execute` should handle that edge-case (including the code that reads from the cache).

^ is also further motivation for the centralized `build_cached_project_plugins_tarball` helper.

lib/bolt_server/transport_app.rb
```diff
@@ -142,6 +139,84 @@ def run_task(target, body)
...
+                          new_tarball = build_project_plugins_tarball(body['versioned_project']) do |mod|
+                            search_dirs = []
+                            search_dirs << mod.plugins if mod.plugins?
+                            search_dirs << mod.pluginfacts if mod.pluginfacts?
+                            search_dirs
+                          end
+                          @file_cache.cache_plugin_tarball(body['versioned_project'], 'fact_plugins', new_tarball)
+                          new_tarball
+                        end
+      task_datas = in_bolt_project(body['versioned_project']) do |context|
```
ekinanp, line 174: Hmm, this is a bit concerning because it means we'll be acquiring then releasing a _global_ mutex O(N) times, where N is the # of targets and could be thousands... I don't know how much of a performance hit that may have, if any, but it is probably worth investigating sooner rather than later.

ekinanp: If we find that there could be a performance hit, then one option would be to have these endpoints accept the task data in the request. That would mean Orch would need to calculate that info, meaning we'd need to update the apply-* transport methods to include that in the signature.. At that point, it feels like we'd be defeating what is to me the key purpose of doing this in bolt-server -- to avoid cluttering the apply-* transport method's signatures.

Another option would be to cache the task data too, but that may have its own issues if the tasks are from the builtin content since that can change (which begs the Q. if we can assume that a versioned project is truly immutable due to its dependency on the built-in content, which is mutable).

ekinanp: Hmm, my point re: caching the task data may be a non-issue if we cache everything on the job ID instead of the versioned project since it's much less likely for e.g. the built-in content to change during a running job than it is during the lifetime of a versioned project.

Given that we already cache something on job ID (e.g. [upload_file](https://github.com/puppetlabs/bolt/blob/main/lib/bolt_server/transport_app.rb#L163)), I kind of lean towards doing that here for all of the apply stuff (plugin tarballs, task datas, etc.). Spiritually, each of those bits of computation represent something happening in a specific _job_ so it makes sense to cache said computation under that job (like upload_file).

donoghuc: yeah i thought about this. Couple thoughts. 

1. Project_ref is great for caching. It is unique and immutable. So by caching on that we dont risk dupicating work on jobs run against the same project. 
2. It would be great to remove puppetserver from this equation. This would remove the need to use that mutex at all. I will look in to caching the task metadata because i think that should be easy enough, but i think the better long term goal will be to use file content locally for projects.

spec/Dockerfile.puppetserver
donoghuc: got this released!

lib/bolt_server/transport_app.rb
donoghuc: oh jeeze i thought ruby had atomic writes. yeah i for sure need to fix this up. Thanks

lib/bolt_server/transport_app.rb
donoghuc: I guess the task data is a bit different than plugin tarballs though because the plugin tarball and (for now) the custom facts tasks will be the same per-project, but i guess it is possible for different targets to have different agent install tasks. For those i would likely need to cache by job id.

donoghuc: actually the combination of project_ref and task name should be unique. Implemented that.

lib/bolt_server/transport_app.rb
```diff
@@ -773,15 +854,7 @@ def make_winrm_target(target_hash)
...
-      plugins_tarball = build_project_plugins_tarball(params['versioned_project']) do |mod|
-        search_dirs = []
-        search_dirs << mod.plugins if mod.plugins?
-        search_dirs << mod.pluginfacts if mod.pluginfacts?
-        search_dirs << mod.files if mod.files?
-        type_files = "#{mod.path}/types"
-        search_dirs << type_files if File.exist?(type_files)
-        search_dirs
-      end
+      plugins_tarball = plugin_tarball(params['versioned_project'], 'all_plugins')
```
donoghuc, line 857: We can probably just delete this endpoint eventually.

lib/bolt_server/transport_app.rb
donoghuc: Updated and added tests. Verified that the tests fail without the serial_execute. Great catch Enis!

lib/bolt_server/transport_app.rb
```diff
@@ -142,6 +145,101 @@ def run_task(target, body)
...
+        end
+      end
+    end
+
+    # This helper is responsible for computing or retrieving task metadata for a project.
+    # It expects task name in segments and uses the combination of task name and versioned_project
+    # as a unique identifier for caching in addition to the job_id. The job id is added to protect against
+    # a case where the buildtin content is update (where the apply_helpers would be managed)
+    def project_task_metadata(versioned_project, task_name_segments, job_id)
+      cached_file_name = "#{task_name_segments.join('_')}_#{job_id}"
```
ekinanp, line 203: If we're including job ID as part of the name here, then why not just cache everything under https://github.com/puppetlabs/bolt/blob/main/lib/bolt_server/transport_app.rb#L168 ? Thinking of something like

```
<job_id>/fact_plugins_tarball
<job_id>/all_plugins_tarball
<job_id>/<task_name>_metadata
...
```

(We'll need to change upload_file's caching to something like <job_id>/upload_file_cache/<....> to avoid conflicts where an uploaded file could have the same name as one of the cached components)

I'm OK with deferring that to a subsequent PR though, just thought I would bring it up here.

spec/fixtures/bolt_server/projects/bolt_server_test_project/modules/apply_helpers/tasks/apply_catalog.rb
```diff
@@ -0,0 +1,117 @@
+#! /opt/puppetlabs/puppet/bin/ruby
```
ekinanp, line 1: Wondering if there's a Puppetfile for tests that we can put these external deps in. I'm guessing no?

spec/bolt_server/app_integration_spec.rb
```diff
@@ -171,6 +176,191 @@ def app
...
+        }
+        post(path, JSON.generate(body), 'CONTENT_TYPE' => 'text/json')
+        expect(last_response).to be_ok
+        expect(last_response.status).to eq(200)
+        result = JSON.parse(last_response.body)
+        expect(result['status']).to eq('success')
+        expect(result['value'].keys).to include('os')
+      end
+
+      it 'apply_prep fails when install task fails' do
```
ekinanp, line 207: Should probably add an "apply_prep fails when custom facts task" fails test as well if we are including this here.

spec/bolt_server/app_integration_spec.rb
```diff
@@ -171,6 +176,191 @@ def app
...
+        }
+        post(path, JSON.generate(body), 'CONTENT_TYPE' => 'text/json')
+        expect(last_response).to be_ok
+        expect(last_response.status).to eq(200)
+        result = JSON.parse(last_response.body)
+        expect(result['status']).to eq('failure')
+        expect(result['object']).to eq('fake_puppet_agent::install')
+      end
+
+      it 'apply_prep fails when target does not define suitable puppet_library plugin_hook' do
```
ekinanp, line 232: Wonder if we can just expect callers to handle this or encode this in the schema somewhere...

spec/bolt_server/app_integration_spec.rb
```diff
@@ -171,6 +176,191 @@ def app
...
+              }
+            ],
+            "classes" => [
+              "settings"
+            ]
+          }
+        }
+      end
+
+      it 'applies a catalog' do
```
ekinanp, line 333: Do we want a test to see what this does if the catalog fails to apply?

lib/bolt_server/transport_app.rb
donoghuc: Sure, For this iteration i wanted to keep per versioned project for plugin tarballs (as nothing we are shipping in bundled content needs to go in there)  I only wanted per versioned project plus job id for task data (as the apply_helpers are distributed there). I'm still trying to wrap my head around how practically caching will even work in kubernetes. Like is there even a scenario where there would be a deploy that includes bundled content that *also* would preserve the cache?

spec/fixtures/bolt_server/projects/bolt_server_test_project/modules/apply_helpers/tasks/apply_catalog.rb
donoghuc: I thought about shelling out to r10k or even using bolt to install these as part of test setup. But i think its more work than its worth. I think for now just shipping these as duplicates is OK. If they become problematic we can update. The way i see we get the "real" test coverage in orch acceptance where we run against the actual builtin bolt content.

spec/bolt_server/app_integration_spec.rb
donoghuc: I"m not sure that is worth it. The reason I added that test is because we need to ensure the correct behavior (we do not try to run the facts task when the install task fails). Just asserting the facts task fails (which would not really be trivial to do) seems unnecessary.

spec/bolt_server/app_integration_spec.rb
donoghuc: I thought of that too. I think it would require a new "target" schema though special for apply/apply_prep. I suppose i could do that but i'de rather just do an explicit check for now to avoid duplicating target.

spec/bolt_server/app_integration_spec.rb
donoghuc: I dont think we have anything at this level valuable to test. I have coverage for this case in orch https://github.com/puppetlabs/orchestrator/pull/1778/files#diff-5c92a78fec2c5e883d5e84edbbd419974aaf1a39ace4f4464dfa0a5720469d41R109-R117

steveax (approved):
👍

lib/bolt_server/transport_app.rb
ekinanp: Not sure. But worth documenting this discussion somewhere.

ekinanp (approved):

PR #2967, opened 2021-08-03
(docs) update /puppet/ links to point to latest

igalic:
some links were still using a specific version, rather than the
`/latest/` meta-version.
In most cases that did redirect to `/7/` but let's just make sure this
always works.

!no-release-note


beechtom (approved):

beechtom:
@igalic Thanks for finding and fixing this! :)

Issue #2964, opened 2021-08-02
Restart a task running over PCP if it errors with 'puppetlabs.orchestrator/plan-already-finished'


lucywyman:
## Use Case
When running a task through the Orchestrator, the task gets added to PuppetDB with a plan ID for it to run as part of, in order to work with the Orchestrator API. If the Orchestrator is restarted as part of the task (for example as part of the `peadm::pe_install` task), it will fail with the error `"The plan has already finished"`. This happens because when Orchestrator restarts it marks any running plans as failed, which also means they are interpreted as "finished". This is correct, however Bolt is using Orchestrator as a transport and doesn't care about whether the associated plan ID still exists or what it's state is, it just needs to use the Orchestrator to connect to targets and run the task.

To solve this, we can catch the `plan-finished` error and rerun the task with a new plan ID.

### Questions
* How many times should we catch the error? Just once? Make it configurable?
* Are there other instances where we shouldn't catch the error and retry? (I can't think of any).

lucywyman:
## Test cases
- [ ] Reruns the task with a new Plan ID
- [ ] Doesn't rerun the task if it doesn't fail with `plan-already-failed`
- [ ] Only reruns once, if the error is raised twice then the second error will error

### 2019.8.6
- [ ] Reruns a task when the Orchestrator restarts

### 2019.8.7
- [ ] Reruns a task when the Orchestrator restarts

### 2021.4.0
- [ ] Reruns a task when the Orchestrator restarts

PR #2971, opened 2021-08-04
(GH-2964) Restart a task over PCP if it fails with 'plan-already-finished'

lucywyman:
When running a plan over the PCP transport, Bolt issues a
`command/plan_start` API request and gets a Plan ID to use throughout
the Bolt run. The plan ID is then associated with every subsequent
`run_task` request, so that the task is associated with the plan it's
running as part of. If the Orchestrator is restarted as part of the
plan, it (correctly) marks the plan as 'failed' and returns a
`plan-already-finished` error for subsequent task runs using the same
ID. This makes sense, but Bolt can continue to run the tasks with a new
ID given that restarting the Orchestrator is a somewhat common case.
When Bolt gets a `plan-already-finished` error from the Orchestrator it
will start a new plan to generate a new Plan ID, and will rerun the task
and continue executing.

!feature

* **Continue executing when a plan over PCP restarts the Orchestrator**

  Bolt will now retry a task once when running over PCP as part of a
  plan, if the task errors with `plan-already-finished`.

Still manually verifying this, since it's high priority I'm fine with someone else putting up a PR or pushing to my branch if this isn't over the line by Friday.

### DIFF
--- a/lib/bolt/transport/orch/connection.rb
+++ b/lib/bolt/transport/orch/connection.rb
@@ -36,7 +36,8 @@ module Bolt
           end
           logger.debug("Creating orchestrator client for #{client_opts}")
           @client = OrchestratorClient.new(client_opts, true)
-          @plan_job = start_plan(plan_context)
+          @plan_context = plan_context
+          @plan_job = start_plan(@plan_context)
           logger.debug("Started plan #{@plan_job}")
           @environment = opts["task-environment"]
         end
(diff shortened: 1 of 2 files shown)


lucywyman:
Sweet, reproduced the same error + verified that we at least get a *different* error on PE 2019.8.7:
```
[root@plush-file ~]# bolt plan run bolt -t pcp://myhost
Starting: plan bolt
Starting: command 'whoami' on pcp://myhost
Finished: command 'whoami' with 0 failures in 1.94 sec
Starting: command 'systemctl restart pe-orchestration-services' on localhost
Finished: command 'systemctl restart pe-orchestration-services' with 0 failures in 12.68 sec
Starting: command 'whoami' on pcp://myhost
Finished: command 'whoami' with 1 failure in 3.2 sec
Finished: plan bolt in 17.82 sec
Failed on pcp://myhost:
  This node (myhost) is either disconnected or does not have a connection type that supports running tasks/plans.
Failed on 1 target: pcp://myhost
Ran on 1 target
```

lucywyman:
Mmk, also confirmed for 2019.8.6 (and the disconnect error is fixed by waiting for Orchestrator to come back up).

beechtom (approved):

Issue #2966, opened 2021-08-03
Stacktrace when adding module when Puppetfile includes same module without a version


beechtom:
## Describe the Bug

When adding a module to a project:

```shell
$ bolt module add puppet/cron
```

When the project Puppetfile already includes the module without a version requirement:

```ruby
mod 'puppet/cron'
```

Bolt will stacktrace because the `ModuleInstaller` does not handle Puppetfile modules without a version:

```shell
Adding module puppet/cron to project

  → Resolving module dependencies, this might take a moment

Traceback (most recent call last):
	13: from /Users/tom/.rvm/gems/ruby-2.6.3/bin/ruby_executable_hooks:24:in `<main>'
	12: from /Users/tom/.rvm/gems/ruby-2.6.3/bin/ruby_executable_hooks:24:in `eval'
	11: from /Users/tom/git/bolt/vendor/bundle/ruby/2.6.0/bin/bolt:23:in `<main>'
	10: from /Users/tom/git/bolt/vendor/bundle/ruby/2.6.0/bin/bolt:23:in `load'
	 9: from /Users/tom/git/bolt/exe/bolt:11:in `<top (required)>'
	 8: from /Users/tom/git/bolt/lib/bolt/cli.rb:559:in `execute'
	 7: from /Users/tom/git/bolt/lib/bolt/cli.rb:912:in `add_project_module'
	 6: from /Users/tom/git/bolt/lib/bolt/outputter.rb:46:in `spin'
	 5: from /Users/tom/git/bolt/lib/bolt/cli.rb:913:in `block in add_project_module'
	 4: from /Users/tom/git/bolt/lib/bolt/module_installer.rb:61:in `add'
	 3: from /Users/tom/git/bolt/lib/bolt/module_installer.rb:113:in `print_puppetfile_diff'
	 2: from /Users/tom/git/bolt/lib/bolt/module_installer.rb:113:in `select'
	 1: from /Users/tom/git/bolt/lib/bolt/module_installer.rb:115:in `block in print_puppetfile_diff'
/Users/tom/git/bolt/lib/bolt/module_installer.rb:115:in `>': comparison of SemanticPuppet::Version with nil failed (ArgumentError)
```

## Solution

[...]

PR #2981, opened 2021-08-11
(GH-2966) Fix stacktrace for conflicting `module add` and `Puppetfile`

artlawson:
When a project's `Puppetfile` has a forge module without a version specified, Bolt now offers a more helpful error when attempting to do a `module install/add` of the same module.

!bug

* **Fix stacktrace for conflicting module add and Puppetfile**
[(#2966)](https://github.com/puppetlabs/bolt/issues/2966)

  When a project's `Puppetfile` has a forge module without a version specified,
  Bolt now offers a more helpful error when attempting to do a `module install/add` of 
  the same module.

### DIFF
--- a/lib/bolt/module_installer/puppetfile.rb
+++ b/lib/bolt/module_installer/puppetfile.rb
@@ -97,20 +97,34 @@ module Bolt
           end
         end
 
-        return if unsatisfied_specs.empty?
-
+        versionless_mods = @modules.select { |mod| mod.is_a?(ForgeModule) && mod.version.nil? }
         command = Bolt::Util.windows? ? 'Install-BoltModule -Force' : 'bolt module install --force'
 
-        message = <<~MESSAGE.chomp
-          Puppetfile does not include modules that satisfy the following specifications:
+        if unsatisfied_specs.any?
+          message = <<~MESSAGE.chomp
+            Puppetfile does not include modules that satisfy the following specifications:
 
-          #{unsatisfied_specs.map(&:to_hash).to_yaml.lines.drop(1).join.chomp}
-          
-          This Puppetfile might not be managed by Bolt. To forcibly overwrite the
-          Puppetfile, run '#{command}'.
-        MESSAGE
+            #{unsatisfied_specs.map(&:to_hash).to_yaml.lines.drop(1).join.chomp}
+            
+            This Puppetfile might not be managed by Bolt. To forcibly overwrite the
+            Puppetfile, run '#{command}'.
+          MESSAGE
 
-        raise Bolt::Error.new(message, 'bolt/missing-module-specs')
+          raise Bolt::Error.new(message, 'bolt/missing-module-specs')
+        end
+
+        if versionless_mods.any?
+          message = <<~MESSAGE.chomp
+            Puppetfile includes Forge modules without a version requirement:
+            
+            #{versionless_mods.map(&:to_spec).join.chomp}
+            
+            This Puppetfile might not be managed by Bolt. To forcibly overwrite the
+            Puppetfile, run '#{command}'.
+          MESSAGE
+
+          raise Bolt::Error.new(message, 'bolt/missing-module-version-specs')
+        end
       end
     end
   end
--- a/spec/bolt/module_installer/puppetfile_spec.rb
+++ b/spec/bolt/module_installer/puppetfile_spec.rb
@@ -112,5 +112,21 @@ describe Bolt::ModuleInstaller::Puppetfile do
         )
       end
     end
+
+    context 'with missing module version' do
+      let(:spec)    { double('spec', satisfied_by?: true) }
+      let(:version) { nil }
+
+      let(:mod) do
+        [double('mod', name: name, version: version, is_a?: true, to_spec: "mod '#{name}'")]
+      end
+
+      it "errors" do
+        expect { puppetfile.assert_satisfies(specs) }.to raise_error(
+          Bolt::Error,
+          /Puppetfile includes Forge modules without a version/
+        )
+      end
+    end
   end
 end
--- a/spec/bolt/module_installer_spec.rb
+++ b/spec/bolt/module_installer_spec.rb
@@ -54,6 +54,16 @@ describe Bolt::ModuleInstaller do
       )
     end
 
+    it 'errors if module is already in puppetfile without version' do
+      File.write(puppetfile, "mod '#{new_module}'")
+      expect {
+        installer.add(new_module, [], puppetfile, moduledir, project_file, install_config)
+      }.to raise_error(
+        Bolt::Error,
+        /Puppetfile includes Forge modules without a version/
+      )
+    end
+
     it 'updates files and installs modules' do
       expect(installer).to receive(:install_puppetfile)
       installer.add(new_module, specs, puppetfile, moduledir, project_file, install_config)


artlawson:
Still need to add testing, but wanted feedback on what I've done so far

lib/bolt/module_installer/puppetfile.rb
```diff
@@ -97,20 +97,43 @@ def assert_satisfies(specs)
...
-
+        #versionless_mods = @modules.any? do |mod|    
+        #  mod.version.nil? 
+        #end
+        versionless_mods = []
+        @modules.each {|mod|
+          if mod.version.nil?
+            versionless_mods << mod
+          end
+        }
```
artlawson, line 108: On lines 100-102, I attempted to do the pretty ruby version of what I do on 103-107. Syntax help would be appreciated :)

beechtom, line 108:
```suggestion
        versionless_mods = @modules.select { |mod| mod.version.nil? }
```

[Array#select](https://ruby-doc.org/core-2.7.1/Array.html#method-i-select) lets you create an array of items that meet a given condition.

Also, if the block is a single line then you can use the bracket notation. If the block is multiple lines, use `do...end`.

lib/bolt/module_installer/puppetfile.rb
```diff
@@ -97,20 +97,43 @@ def assert_satisfies(specs)
...
+        
+        return if unsatisfied_specs.empty? && versionless_mods.empty?
```
beechtom, line 110:
```suggestion
```

Since the errors below are now gated by conditionals, this return statement is redundant.

lib/bolt/module_installer/puppetfile.rb
```diff
@@ -97,20 +97,43 @@ def assert_satisfies(specs)
...
         command = Bolt::Util.windows? ? 'Install-BoltModule -Force' : 'bolt module install --force'
+        
+        if !unsatisfied_specs.empty?
```
beechtom, line 113:
```suggestion
        if unsatisfied_specs.any?
```

Prefer not to negate in a conditional, since it can make the logic a bit harder to understand. Another alternative is `unless unsatisfied_spec.empty?`.

lib/bolt/module_installer/puppetfile.rb
```diff
@@ -97,20 +97,43 @@ def assert_satisfies(specs)
...
+
+            #{unsatisfied_specs.map(&:to_hash).to_yaml.lines.drop(1).join.chomp}
+            
+            This Puppetfile might not be managed by Bolt. To forcibly overwrite the
+            Puppetfile, run '#{command}'.
+          MESSAGE
+
+          raise Bolt::Error.new(message, 'bolt/missing-module-specs')
+
+        else
```
beechtom, line 125:
```suggestion
        if versionless_mods.any?
```

Having this be a separate conditional instead of an `else` clause makes the code a little clearer. It'll also be needed if the `return` statement above is deleted.

lib/bolt/module_installer/puppetfile.rb
```diff
@@ -97,20 +97,43 @@ def assert_satisfies(specs)
...
+          message = <<~MESSAGE.chomp
+            Puppetfile includes Forge modules without a version requirement:
+            #{versionless_mods.map(&:to_hash).to_yaml.lines.drop(1).join.chomp}
```
beechtom, line 128:
```suggestion
            #{versionless_mods.map(&:name).join.chomp}
```

`to_hash` will format the modules similar to how they'd be written in a `bolt-project.yaml` file, which is a bit misleading. Instead, we just want to list the module names that are missing the version requirement.

beechtom (changes requested):
This will need a test or two that asserts Bolt errors when adding/installing modules and a module in the Puppetfile is missing a version requirement. You can create a unit test in `spec/bolt/module_installer/puppetfile_spec.rb` and/or an integration test in `spec/bolt/module_installer_spec.rb`.

lib/bolt/module_installer/puppetfile.rb
beechtom: Oops, actually, this needs to make sure the module is a Forge module, too. Git modules don't have a `version` method, so you'll get an "undefined method" error otherwise. Updated the suggestion above.

spec/bolt/module_installer/puppetfile_spec.rb
```diff
@@ -112,5 +112,26 @@
...
+      let(:spec) { double('spec', satisfied_by?: false, to_hash: {})}
+      let(:pal)            { double('pal', generate_types: nil) }
+      let(:outputter) do
+        double('outputter', print_message: nil,
+                            print_puppetfile_result: nil,
+                            start_spin: nil,
+                            stop_spin: nil,
+                            print_action_step: nil)
+      end
+      let(:installer)      {  Bolt::ModuleInstaller.new(outputter, pal) }
```
artlawson, line 125: I'm sure there is a better/easier way to stub for this behavior, but I wanted to try and get a working test by myself first before asking for help. I used `spec/bolt/module_installer_spec.rb` to help out.

artlawson: which I realized soon after uploading this still isn't working 100% lol, but its close enough. Will pick up tomorrow with "cleaner" version.

beechtom: The tests in this file are unit tests for the `Bolt::ModuleInstaller::Puppetfile` class, so we shouldn't be creating an instance of `Bolt::ModuleInstaller` here. A test in this file should likely follow this structure:

- Configure the `puppetfile` instance (defined on line 16) to include a module without a version. This module is already defined (line 15), and sets its version using `version` (line 14). You can override `version` by redefining it in this test's `context`.
- Set the `spec` double to return `true` for `satisfied_by?`. We want to only test that a missing module version raises an error. Setting `satisfied_by?` to `false` would also raise an error.
- Call `puppetfile.assert_satisifies(spec)`, similar to the test above (line 108), and assert that it raises the correct error.

The test you have here is pretty close to what we'd want for an integration test, though. It should probably stay in `spec/bolt/module_installer_spec.rb` in the `#add` context. Overall, the text will probably look almost identical to [this one](https://github.com/puppetlabs/bolt/blob/main/spec/bolt/module_installer_spec.rb#L47-L55), with the only difference being the content written to the Puppetfile (like what you have here on line 129).

lib/bolt/module_installer/puppetfile.rb
```diff
@@ -97,20 +97,33 @@ def assert_satisfies(specs)
           end
         end
 
-        return if unsatisfied_specs.empty?
-
+        versionless_mods = @modules.select { |mod| mod.is_a?(ForgeModule) && mod.version.nil? }
```
artlawson, line 100: @beechtom currently the test is failing because the module isn't being recognized as a `ForgeModule`. I tried with the ticket example `puppet/cron` and the pre-existing name `puppetlabs/yaml`

spec/bolt/module_installer/puppetfile_spec.rb
```diff
@@ -112,5 +112,20 @@
         )
       end
     end
+    context 'with missing module version' do
+      let(:spec) { double('spec', satisfied_by?: true)}
+      let(:version) {nil}
+      let(:name) {'puppet/cron'}
+      let(:mod)  { [double('mod', name: name, version: version, to_spec: "mod '#{name} #{version}'")] } #i know having version is probably unnecessary
+      let(:puppetfile) { described_class.new(mod) }
```
beechtom, line 120:
```suggestion
```

These are all defined on lines 13-16 already.

spec/bolt/module_installer/puppetfile_spec.rb
```diff
@@ -112,5 +112,20 @@
...
+      
+      it "errors" do
+        puppetfile.write(path)
```
beechtom, line 123:
```suggestion
```

This is not needed since we're not testing writing a Puppetfile.

spec/bolt/module_installer/puppetfile_spec.rb
```diff
@@ -112,5 +112,20 @@
         )
       end
     end
+    context 'with missing module version' do
+      let(:spec) { double('spec', satisfied_by?: true)}
+      let(:version) {nil}
```
beechtom, line 117:
```suggestion

    context 'with missing module version' do
      let(:spec)    { double('spec', satisfied_by?: true) }
      let(:version) { nil }
```

Changes to whitespace for improved readability / consistency with other tests.

beechtom (changes requested):
Because we only check `ForgeModule` instances when looking for modules without a version, we need to stub the `mod` double so it 'looks' like a `ForgeModule` instance to Ruby. We can do this by stubbing the `is_a?` method to return `true` on the `mod` double on line 15.

The double already stubs several methods by providing a key that is the name of the method and a value that that method returns. Stubbing methods looks like this:

```ruby
double('name_of_double', stubbed_method: return_value)
```

lib/bolt/module_installer/puppetfile.rb
```diff
@@ -97,20 +97,33 @@ def assert_satisfies(specs)
...
-          Puppetfile, run '#{command}'.
-        MESSAGE
+          raise Bolt::Error.new(message, 'bolt/missing-module-specs')
+        end
 
-        raise Bolt::Error.new(message, 'bolt/missing-module-specs')
+        if versionless_mods.any?
+          message = <<~MESSAGE.chomp
+            Puppetfile includes Forge modules without a version requirement:
+            #{versionless_mods.map(&:name).join.chomp}
```
beechtom, line 119:
```suggestion

            #{versionless_mods.map(&:to_spec).join.chomp}
```

To keep the formatting between the two error messages the same.
Also, this should be `to_spec` instead of `name`. Forgot about that method yesterday, my bad. 😬

spec/bolt/module_installer/puppetfile_spec.rb
```diff
@@ -112,5 +112,21 @@
...
       end
     end
+
+    context 'with missing module version' do
+      let(:spec)    { double('spec', satisfied_by?: true) }
+      let(:version) { nil }
+      let(:mod)       {
+        [double('mod', name: name, version: version,
+                       is_a?: true, to_spec: "mod '#{name}'")]
+      }
```
beechtom, line 122:
```suggestion

      let(:mod) do
        [double('mod', name: name, version: version, is_a?: true, to_spec: "mod '#{name}'")]
      end
```

Since this is too long for one line wrap it in `do .. end`.

spec/bolt/module_installer_spec.rb
```diff
@@ -53,6 +53,15 @@
         /managed by Bolt/
       )
     end
+    it 'errors if module is already in puppetfile without version' do
+      installer.add(new_module, specs, puppetfile, moduledir, project_file, install_config)
```
artlawson, line 57: Since this errors with a `File.write` call, I'm not sure how to put `new_module` in the puppetfile without the version automatically being appended. I believe this is the last thing for is.

beechtom:
```suggestion
      installer.add(new_module, specs, puppetfile, moduledir, project_file, install_config)
```

You don't want to use the installer here. Instead, you should write to the Puppetfile directly like the test above does on line 48.

spec/bolt/module_installer_spec.rb
```diff
@@ -54,6 +54,17 @@
       )
     end
 
+    it 'errors if module is already in puppetfile without version' do
+      File.write(puppetfile, new_module)
```
beechtom, line 58:
```suggestion
      File.write(puppetfile, "mod '#{new_module}'")
```

We want to write a valid Puppetfile instead of writing just the module name. [This](https://puppet.com/docs/pe/latest/puppetfile.html) doc goes over the Puppetfile format.

spec/bolt/module_installer_spec.rb
```diff
@@ -54,6 +54,17 @@
...
+      expect {
+        installer.add(new_module, specs, puppetfile, moduledir, project_file, install_config)
```
beechtom, line 60:
```suggestion
        installer.add(new_module, [], puppetfile, moduledir, project_file, install_config)
```

We should pass an empty array for specs here. The specs are the modules listed in project config (see [this doc](https://puppet.com/docs/bolt/latest/bolt_installing_modules.html#manually-specify-modules-in-a-bolt-project)). By passing an empty array for the specs, we won't hit the first error that ensures all specs are present in the Puppetfile.

beechtom (changes requested):
Needs a release note, but otherwise lgtm 👍

beechtom (changes requested):
The release note needs to be added to the end of the commit message.

artlawson:
@beechtom I changed the message so the top and bottom match, (the top description wasn't very clear in the first place). Is that fine? Not sure if I fully understand your last comment

beechtom:
@artlawson The release note needs to be added to the commit message itself (i.e. the message you write when you run `git commit --amend`): https://github.com/puppetlabs/bolt/pull/2981/commits/40e60007a1f5719ec7bd168510d762092f418221 .

beechtom (approved):

lucywyman:
Hm, this still raises the error even if `bolt module add puppetlabs/cron --force` is provided, as the error recommends.

PR #2984, opened 2021-08-12
(maint) Bump http_request module to 0.3.1

beechtom:
This bumps the http_request module to 0.3.1.

!bug

* **Handle empty response bodies in `http_request` task**

  The `http_request` task no longer errors when a response does not
  include a body.


artlawson (approved):
👍

PR #2990, opened 2021-08-18
(maint) Fix broken integration tests

beechtom:
This fixes a shareable task integration test that was caused by the release of
Puppet 7.10.0, which includes changes to an error message that the test
matches against.

This also fixes parallel integration tests on winrm that were failing
due to connection timeout errors. The connection timeout seemed to be
caused by running multiple `sleep` commands on the winrm target.

!no-release-note


artlawson (approved):
Passed local testing & LGTM!

Issue #2992, opened 2021-08-19
"bolt module install" no longer works when there's a task plugin in the bolt-project file


jschoewe:
## Describe the Bug
When running "bolt module install" on a project, installing the modules from the Puppetfile needs to be the first thing that happens because tasks from those modules sometimes need to be used as plugins in the bolt-project.yaml file.

## Steps to Reproduce

Puppetfile:
```
mod 'puppetlabs-stdlib'
mod 'encore-inventory_utils'
```

bolt-project.yaml
```yaml
---
name: deploy

# paths are relative to this bolt.yaml file
hiera-config: "./hiera.yaml"
modulepath: "./modules_extern:./modules"

log:
  console:
    level: info

plugins:
  pkcs7:
    public_key:
      _plugin: task
      task: inventory_utils::pick_path
      parameters:
        paths:
          - ~/.ssh/public_key.pkcs7.pem
          - ~/.puppetlabs/bolt/keys/public_key.pkcs7.pem
    private_key:
      _plugin: task
      task: inventory_utils::pick_path
      parameters:
        paths:
          - ~/.ssh/private_key.pkcs7.pem
          - ~/.puppetlabs/bolt/keys/private_key.pkcs7.pem
```

Running "bolt module install" on a project with these files results in the following error:
```
Loaded project from '/path/to/project/deploy'
Could not find a task named 'inventory_utils::pick_path'. For a list of available tasks, run 'bolt task show'.
```

## Expected Behavior
Prior to version 3.16, these modules were being installed fine and we were able to use their tasks in the bolt-project file.

## Environment
RHEL 8

PR #2993, opened 2021-08-19
(GH-2992) Defer resolving plugin references

beechtom:
This updates the Bolt to only resolve plugin config (`plugins`) and
plugin hooks (`plugin-hooks`) as needed. Previously, when Bolt setup the
`Bolt::Plugin` class, it resolved all plugin config and hooks as part of
the setup process. Because Bolt now loads the `Bolt::Plugin` class for
all commands, this resulted in Bolt erroring in cases where a plugin
module was not available, even if the command run didn't need access to
any plugins. This is particularly problematic for the `module install`
command, as users installing the project for the first time would be
attempting to install the very plugin modules that Bolt will say are
missing when the command is run.

---

This updates the inventory to defer resolving plugins until the
inventory is needed. During the CLI refactor, Bolt was changed to always
load the inventory, regardless of whether the inventory was actually
used. This resulted any plugins in an inventory being resolved whenever
a Bolt command was run, potentially leading to a significant degradation
in performance.

Now, when the inventory is instantiated, Bolt only stores the raw
inventory data. Once Bolt requests access to any of the groups, targets,
or config in the inventory, the inventory parses the data, resolves
plugin references, and loads the groups and targets into inventory.

---

!bug

* **Resolve plugins as needed**
  ([#2992](#2992))

  Bolt now only resolves plugin configuration, plugin hooks, and
  inventory plugins as needed. Previously, Bolt resolved all plugins
  each time a command was run, causing some commands to fail when
  they should not.


lib/bolt/plugin.rb
```diff
@@ -167,6 +163,20 @@ def initialize(config, pal, analytics, load_plugins: true)
       end
       @unresolved_plugin_configs['puppetdb'] = config.puppetdb if config.puppetdb
       @plugin_hooks = DEFAULT_PLUGIN_HOOKS.dup
+      @unresolved_plugin_hooks = config.plugin_hooks
```
lucywyman, line 166: This seems like it should be a boolean instead of data, and you can just use `resolve_references(config.plugin_hooks)` instead of `resolve_references(unresolved_plugin_hooks)`.

lib/bolt/plugin.rb
```diff
@@ -145,9 +143,7 @@ def self.setup(config, pal, analytics = Bolt::Analytics::NoopClient.new, **opts)
     DEFAULT_PLUGIN_HOOKS = { 'puppet_library' => { 'plugin' => 'puppet_agent', 'stop_service' => true } }.freeze
 
     attr_reader :pal, :plugin_context
-    attr_accessor :plugin_hooks
-
-    private_class_method :new
+    attr_writer :plugin_hooks
```
lucywyman, line 146: Hmm, definitely seems safer to keep this but I also don't see where we write to it...

lib/bolt/plugin.rb
beechtom: 👍 Good suggestion. Opted for lazily merging the config and assigning it to `@plugin_hooks`.

lib/bolt/cli.rb
```diff
@@ -434,7 +434,7 @@ def execute(options)
             config.project
           )
 
-          plugins = Bolt::Plugin.setup(config, pal, analytics)
+          plugins = Bolt::Plugin.new(config, pal, analytics)
```
lucywyman, line 437: So, this is a bit weird. `Plugin.setup` will resolve any plugins called under the `plugins` key in configuration so that any plugin defaults are already loaded when we go to resolve plugins. However, we already call `config_for_plugin` on plugins before resolving them to load the config, which will resolve the config `plugins` key, and I assume we do that for inventory plugins too.

This change only works because all plugins already call `config_for_plugin` themselves, which resolves `config.plugins`. By bypassing this we require any other configuration options that we make pluggable to also call `config_for_plugin` on their own, which could be confusing to figure out if we ever make other options pluggable. That's a big if though, since we haven't heard of anyone wanting to make other config options pluggable in *looks at watch* years, plus I can't imagine a situation where you wouldn't call `config_for_plugin` since that's how you get the plugin config to begin with.

If requiring any config options to call `config_for_plugin` on their own is what we want to do (which I feel pretty comfortable with), then it seems like we should remove `setup` altogether so that we don't *sometimes* load all plugins that are configured, and sometimes require the plugins to be resolved by calling `by_name` manually. It also seems like `setup` is just unnecessary at this point, we added it around having PuppetDB client config be configurable with plugins, and plugin loading has changed so much since hten. It seems like all `Plugin.setup` calls can be replaced with `Plugin.new`:
* In `BoltSpec::Context` this gets passed to the real Inventory, which resolves any plugins (if the above works, then the call in `BoltSpec::Context` should too)
* Same in `BoltSpec::Run`
* The empty inventory....might be fine? I initially thought it wouldn't matter but need to test if you have an empty inventory but then call `resolve_references` from a plan
* `BoltServer::TransportApp` also adds all their own modules manually

lib/bolt/plugin.rb
beechtom: I assumed it was for some tests way back or something. Also decided to not risk removing it, though. :P

lib/bolt/cli.rb
beechtom: I was initially going to drop usage of `setup` altogether, but wasn't sure of the ramifications elsewhere (particularly BoltServer). If you're comfortable dropping it then I can pick up those changes tomorrow.

Issue #2997, opened 2021-08-25
Remove OMI, Podman, and LXD tests


lucywyman:
## Use Case
These features are experimental, and with no plans to move them out of experimental we can remove automated tests for them. We should verify that manual testing instructions work on Mac and Linux, and that manual testing is documented. And we can move any files that support testing OMI, LXD, or Podman (such as `spec/Dockerfile.omiserver`, `spec/lxd_config.yaml`) into a new `dev-resources` directory and update any manual testing instructions to use that new directory.

PR #3002, opened 2021-08-26
(GH-2997) Remove OMI, Podman, and LXD tests

lucywyman:
Due to lack of prioritization and usage, we aren't expecting the Windows
Kerberos support, Podman transport, or LXD transport to become stable
anytime in Bolt's future. If these features are permanently
experimental, we don't need to test them as part of CI. This removes
tests for Podman, LXD, and Kerberos on Windows, and moves testing
infrastructure and setup directions to a new `dev-resources` directory
so that devs can still troubleshoot any issues that users run into.

This additionally removes some long-running tests that duplicate test coverage.
Integration tests that verify installing the Puppet agent take up to a
minute to run on nix, and a few of the tests are now duplicates (such
as verifying a task plugin and the `puppet_agent` plugin, given the
puppet_agent plugin *is* the task plugin). This removes long running
tests that seem unnecessary.

!no-release-note


dev-resources/lxd.md
```diff
@@ -0,0 +1,18 @@
+### Provisioning for LXD tests
```
beechtom, line 1: Can this be under `dev-resources/lxd/README.md` instead?

Issue #2905, opened 2021-06-29
BoltSpec `expect_upload` assumes source file resides inside the module


reidmv:
## Describe the Bug

Currently, the expect_upload BoltSpec function assumes the uploaded file resides inside the module. This limits its usefulness as the following example shows:

https://github.com/puppetlabs/puppetlabs-peadm/blob/main/spec/plans/util/retrieve_and_upload_spec.rb#L17-L27

The above example implements a work-around for mocking the upload_file call as defined here: https://github.com/puppetlabs/puppetlabs-peadm/blob/main/plans/util/retrieve_and_upload.pp#L32  

## Expected Behavior

When mocking `upload_file` with an absolute path source argument, the mock should not fail because the file is not present on the filesystem. This is a mock.

## Steps to Reproduce

1. Create a plan that calls `upload_file` with an absolute source path such as "/tmp/not-a-real-file"
2. Try and write a spec test for that using `expect_upload`

PR #3004, opened 2021-08-27
(GH-2905) Update BoltSpec functions to allow absolute paths

MikaelSmith:
Updates BoltSpec `allow/expect_upload` and `allow/expect_script` functions to accept an absolute path. The path referred to must still exist, but that's relatively easy to mock with temporary files.

----

!bug

* **Update BoltSpec functions to allow absolute paths** ([puppetlabs#2905](puppetlabs#2905))

  BoltSpec's `allow_upload`, `expect_upload`, `allow_script`, and `expect_script` functions now support passing absolute paths that rather than a module reference. The referenced path must refer to a real file or upload and script functions will still error.

### DIFF
--- a/lib/bolt_spec/bolt_context.rb
+++ b/lib/bolt_spec/bolt_context.rb
@@ -39,12 +39,14 @@ require 'bolt/plugin'
 #
 # Stubs:
 # - allow_command(cmd), expect_command(cmd): expect the exact command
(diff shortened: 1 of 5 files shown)


lucywyman (approved):
Verified this works for all `expect_upload`, `allow_upload`, `expect_script`, and `allow_script`

PR #3005, opened 2021-08-28
(maint) Break out long running test jobs in GH Actions

lucywyman:
Now that the limit of concurrent Github Actions jobs for the Bolt repo
is 180 instead of 20, we can break the integration tests into more
separate jobs so that they don't take 50+ minutes to run. Running the 
tests concurrently with rspec would also be an option, but many of these
tests use the same infrastructure and we want to avoid multiple
access errors as much as possible.

This moves all of the transport integration tests into their own Github Actions jobs, which
profiled high in the 100 longest-running jobs. It also moves all of the apply tests into their own 
jobs for each platform for similar reasons, and the bolt_spec and bolt_server tests into their own jobs.

This also moves common setup steps into composite Github Action steps so that we aren't duplicating YAML everywhere.

!no-release-note


.github/actions/common_setup/action.yaml
```diff
@@ -0,0 +1,23 @@
+name: Common setup steps
+description: Install Ruby into the runner environment, install gem dependencies, and install Puppet modules
+runs:
+  using: composite
+  steps:
+    - name: Checkout repository
+      uses: actions/checkout@v2
```
beechtom, line 7: I don't think this action is available until you checkout the repo. Removing this step should allow us to use this as an action in the rest of the workflows for setting up Ruby/gems/modules, unless there's something I'm missing.

Issue #2921, opened 2021-07-07
Support new file reference syntax for the `project_file_metadatas` endpoint


lucywyman:
## Use Case
Modify the `project_file_metadatas` BoltServer endpoint to have the same behavior for looking up scripts that Bolt has:
* First try the currently supported, original path
* If not found, parse the path to see if BoltServer should look in `files/` or `scripts/`
* Return 400 if the file still isn't found

PR #3007, opened 2021-08-31
(GH-2921) Support loading from scripts/ in BoltServer

lucywyman:
This modifies the `project_file_metadatas` BoltServer endpoint use the
behavior currently under the `future.file_paths` feature flag in Bolt by
default.

When a user sends a file request to BoltServer, in includes a Puppet
file reference. BoltServer can't know if that file reference uses the
old, 'generic' Puppet file reference syntax `<module>/<path>` or the new
`<module>/files/<path>` syntax. To support both, BoltServer will now
first look in the old location for the given path, and if the file isn't
found it will parse the provided path to determine whether it can look
in `files/` or `scripts/` for the provided subpath. If the file isn't
found in either the old or new expected location, it will return 400 as
it did previously when a file wasn't found.

This also includes tarring the scripts directory in the plugin tarballs,
as scripts can be referenced from task metadata so need to be available
as part of the plugin tarball.

This additionally fixes a bug in the Applicator where `scripts` files
were not included in the plugins tarball to apply Puppet code, meaning
that using the scripts mount in apply blocks and manifests didn't work.

!bug

* **Allow using the scripts file mount in Puppet manifests**

  This fixes a bug where `scripts/` directory files needed for the new 
  Puppet scripts mount were not included in the tarball of dependencies
  for applying manifests. Users can now use the scripts mount from Puppet
  manifests.


lucywyman:
The unit test failure is due to AWS having issues right now, which is impacting installing modules from the Forge :/

lib/bolt_server/transport_app.rb
```diff
@@ -184,6 +184,7 @@ def plugin_tarball(versioned_project, tarball_type)
             search_dirs << mod.pluginfacts if mod.pluginfacts?
             if tarball_type == 'all_plugins'
               search_dirs << mod.files if mod.files?
+              search_dirs << mod.scripts if mod.scripts?
```
nicklewis, line 187: I'm not sure this is necessary. We don't do it [in the applicator](https://github.com/puppetlabs/bolt/blob/main/lib/bolt/applicator.rb#L241) and Puppet manifests shouldn't be using anything from the scripts directory.

lucywyman: Ah, I think not having it in the Applicator is actually the bug. Users can use the `scripts` mount in Puppet catalogs, which works now if that bug is fixed. I think fixing that here makes sense.

nicklewis (approved):

Issue #2977, opened 2021-08-10
Print error result sets from parallelize blocks


beechtom:
## Use Case
When running an executor command in a parallelize block, Bolt will not print detailed error messages for any failing results. For example, given the following plan:

```puppet
plan bolt (
  TargetSpec $targets
) {
  $results = parallelize(get_targets($targets)) |$target| {
    run_command('foobar', $target)
  }

  return $results
}
```

A failing result will look like:

```shell
Starting: plan bolt
Starting: command 'foobar' on localhost
Finished: command 'foobar' with 1 failure in 0.01 sec
Error in future '2': run_command 'foobar' failed on 1 target [ID: errored_futures]
Finished: plan bolt in 1.04 sec
parallel block failed on 1 target
  (file: /Users/tom/.puppetlabs/bolt/plans/init.pp, line: 14, column: 10)
```

When running this function outside of a parallelize block, the error result is much more detailed:

```shell
Failed on localhost:
  No such file or directory - foobar
```

If an executor action fails in a parallelize block, the error message should include this additional information.

lucywyman:
Ah, so when we create the main plan Fiber [we assign it ID 1](https://github.com/puppetlabs/bolt/blob/main/lib/bolt/pal.rb#L739), but when we check to see if the main plan is still running to know whether to raise a warning or not [we assume the main plan has ID 0](https://github.com/puppetlabs/bolt/blob/main/lib/bolt/fiber_executor.rb#L89), which means that this warning is always raised for any errors when it was intended to only be raised if the main plan had finished.

I think for this we should fix that bug. Once that's fixed, the behavior is the same as running `run_command('foo', $targets, '_catch_errors' => true)`, which assumes that users will run with `--verbose` to see actual error messages.

For example with this plan:
```
plan bolt () {
  run_command('foo', 'localhost', '_catch_errors' => true)
}
```

Will result in this output:
```
lucy ~/githubs/bolt (main) $ be bolt plan run bolt -t localhost
Starting: plan bolt
Starting: command 'foo' on localhost
Finished: command 'foo' with 1 failure in 0.0 sec
Finished: plan bolt in 0.01 sec
Plan completed successfully with no result


lucy ~/githubs/bolt (main) $ be bolt plan run bolt -t localhost --verbose
Starting: plan bolt
Starting: command 'foo' on localhost
Started on localhost...
Failed on localhost:
  No such file or directory - foo
Finished: command 'foo' with 1 failure in 0.0 sec
[...]

PR #3014, opened 2021-09-09
(GH-2977) Set main plan ID correctly

beechtom:
This fixes a small bug where the main plan ID was being set incorrectly.
Previously, the main plan ID was set to 1, while the fiber executor
expected the main plan ID to be 0. The main plan ID is now correctly set
to 0 as expected by the fiber executor.

This also adds documentation for running plans in verbose mode to view
error results from a parallel block.

!bug

* **Correctly set main plan ID**
  ([#2977](https://github.com/puppetlabs/bolt/issues/2977))

  Bolt now sets the main plan's ID to the correct value. Previously, the
  main plan ID was set incorrectly, causing Bolt to issue warnings
  during plan runs when it should not have.

### DIFF
--- a/documentation/experimental_features.md
+++ b/documentation/experimental_features.md
@@ -320,7 +320,8 @@ plan myplan(TargetSpec $targets) {
 }
 ```
 
-#### Getting Results from parallel blocks
+#### Getting results from parallel blocks
+
 The `parallelize()` and `wait()` functions return an array that contains the results of executing
 the block in the same order as the input array. You can think of them as `map` functions
 that run in parallel. The `result` of the block for a particular input is either a value passed to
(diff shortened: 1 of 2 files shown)


documentation/experimental_features.md
```diff
@@ -366,6 +367,82 @@ in the result array, the plan throws a `PlanFailure` and includes the entire res
 `PlanFailure` and continues to execute the plan. If you've provided `_catch_errors => true`
 to `wait()`, Bolt returns any errors raised and the plan continues to execute.
 
+#### Viewing failing results in parallel blocks
+
+When an action fails in a parallel block, such as a task returning an error, Bolt
+does not print the error result to the console. For example, the following plan
```
lucywyman, line 373: Hm, it prints the error as a warning...
```
lucy ~/githubs/bolt ((6f8e3c60...)) $ be bolt plan run bolt

Starting: plan bolt
Starting: command 'badcommand' on localhost
Finished: command 'badcommand' with 1 failure in 0.0 sec
Error in future '2': run_command 'badcommand' failed on 1 target [ID: errored_futures]
Finished: plan bolt in 1.07 sec
parallel block failed on 1 target
  (file: /home/lucy/githubs/bolt/plans/init.pp, line: 5, column: 14)
```

lucywyman (approved):

documentation/experimental_features.md
```diff
@@ -366,6 +367,82 @@ in the result array, the plan throws a `PlanFailure` and includes the entire res
...
+  $results = parallelize($_targets) |$target| {
+    run_command('badcommand', $target)
+  }
+
+  return $results
+}
+```
+
+The error result from each target the command is run is not printed to the console,
+making it difficult to know why the command failed:
```
hestonhoffman, line 391:
```suggestion
Bolt does not print the error result from the targets the command is run on,
making it difficult to know why the command failed:
```

hestonhoffman: I think "to the console" is clear from the context here, so we can remove it to shorten this sentence a bit

PR #2291, opened 2020-10-27
(GH-2291) Enable bootsnap loadpath caching

jpogran:
Depends on

- [x] https://github.com/puppetlabs/puppet-runtime/pull/384
- [ ] https://github.com/puppetlabs/bolt-vanagon/pull/156
- [x] Change bolt bootsnap cache dir to $env:ProgramData/Puppetlabs/bolt in bolt project

PR #3020, opened 2021-09-13
(GH-2291) Enable Bootsnap

lucywyman:
This enables Bootsnap's compile cache for Bolt when running on Windows, which caches the compiled Ruby bytecode and YAML Messagepack format results for faster loading. When using Bootsnap we disable the load path cache, since Bolt modifies the load path from the user's environment which results in the cache being recomputed and rewritten every Bolt run, which negates any performance improvements from compilation caching.

!no-release-note

### DIFF
--- a/exe/bolt
+++ b/exe/bolt
@@ -1,6 +1,13 @@
 #!/usr/bin/env ruby
 # frozen_string_literal: true
 
+if RbConfig::CONFIG['host_os'] =~ /mswin|mingw|cygwin/
+  require 'bootsnap'
+  Bootsnap.setup(
+    cache_dir:       File.join(ENV['ProgramData'], 'PuppetLabs', 'bolt'),
+    load_path_cache: false
+  )
+end
 require 'bolt'
 require 'bolt/cli'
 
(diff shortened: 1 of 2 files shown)


exe/bolt
```diff
@@ -1,6 +1,11 @@
 #!/usr/bin/env ruby
 # frozen_string_literal: true
 
+if RbConfig::CONFIG['host_os'] =~ /mswin|mingw|cygwin/
+  ENV['BOOTSNAP_CACHE_DIR'] = File.join(ENV['ProgramData'], 'PuppetLabs', 'bolt')
+  ENV['DISABLE_BOOTSNAP_LOAD_PATH_CACHE'] = true
+  require 'bootsnap/setup'
```
beechtom, line 7:
```suggestion
  require 'bootsnap'
  Bootsnap.setup(
    cache_dir:       File.join(ENV['ProgramData'], 'PuppetLabs', 'bolt'),
    load_path_cache: false
  )
```

Bootsnap recommends manual setup if you're not using Rails (see [here](https://github.com/Shopify/bootsnap#usage)). Personally, I'd prefer to have more direct control over setup instead of relying on env vars, but am fine leaving this as-is if you have concerns.

lucywyman: I actually had some trouble getting this working when it was packaged, let me test it first.

lucywyman: Ok, confirmed this works - the original issue was that `setup` would only accept all five options, so was throwing `compile_cache_iseq not found` (or something along those lines). But this seems to work :+1:

beechtom (approved):

PR #3022, opened 2021-09-13
(PE-32839) Support r10k authorization_token setting

beechtom:
This adds a new `module-install.forge.authorization_token` setting which
can be used to configure r10k to use a token for authorizing requests to
the Forge when installing modules.

!feature

* **Support Forge authorization token when installing modules**

  Bolt now supports using a token to authorizing requests to the Forge
  when installing modules using the
  `module-install.forge.authorization_token` setting.

### DIFF
--- a/documentation/bolt_installing_modules.md
+++ b/documentation/bolt_installing_modules.md
@@ -329,6 +329,27 @@ module-install:
     proxy: https://forge-proxy.com:8080
 ```
 
+## Install modules using an authorization token
+
+If you need to install modules that require an authorization token to download,
+you can configure Bolt to pass a token as part of the request it makes to the
+Forge host. To configure Bolt to use an authorization token, set the
+`module-install` configuration option in either your project configuration file,
+`bolt-project.yaml`, or the default configuration file, `bolt-defaults.yaml`.
+
+To set an authorization token that is used when installing Forge modules,
+set the `authorization_token` key under the `forge` section of the
+`module-install` option. When you set the `authorization_token` key, you
+must also set the `baseurl` key.
+
+```yaml
+# bolt-project.yaml
+module-install:
+  forge:
+    authorization_token: Bearer eyJhbGciOiJIUzI1NiIsInR5c...
+    baseurl: https://forge.example.com
+```
+
 📖 **Related information**
 
 - [bolt-project.yaml options](bolt_project_reference.md#module-install)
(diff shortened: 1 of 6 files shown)


lib/bolt/application.rb
```diff
@@ -258,7 +258,7 @@ def add_module(name, outputter)
                     config.project.puppetfile,
                     config.project.managed_moduledir,
                     config.project.project_file,
-                    config.module_install)
+                    @plugins.resolve_references(config.module_install))
```
lucywyman, line 261: If we resolve here, isn't the whole `module_install` config setting or any subkeys also pluggable? I think that's totally fine, just want to make sure that that's known, and we should also update the data in `options.rb` for those options

beechtom: Technically, they can be. They still currently error when validating config, though, since only `module-install.forge.authorization_key` is marked as pluggable.

hestonhoffman (approved):

lucywyman (approved):
## Tests
- [x] Is pluggable
- [x] Resolves from defaults and project
      - Note: doesn't merge config, choose project or default
- [x] Works when using a proxy
- [x] Fails if the token is incorrect
- [x] Fails if `baseurl` isn't configured
- [x] Works for `bolt module install`
- [x] Works for `bolt module add`
- [x] Installs content that doesn't require a token if the token is invalid
- [x] Installs premium content when it's a dependency
- [x] Fails installing premium content dependencies if the token is invalid
- [x] Only works with `Bearer` prefix
- [x] Ignores r10k configuration
- [x] Ignores puppet.conf configuration

It feels so weird to have no automated tests, but...it does kind of seem like they aren't needed.

lib/bolt/application.rb
lucywyman: Ah, ok :+1:

documentation/bolt_installing_modules.md
```diff
@@ -329,6 +329,27 @@ module-install:
     proxy: https://forge-proxy.com:8080
 ```
 
+## Install modules using an authorization token
```
lucywyman, line 332: Hm, maybe worth mentioning that R10K and Puppet configuration is ignored, so you must include this configuration when using Bolt to install modules that require authorization.

beechtom: I lean towards not mentioning r10k or Puppet config. This section (and the previous two) already states that if you need to use an auth token to install modules you need to configure Bolt. Adding that you need to do this in Bolt because we don't load r10k or Puppet config seems redundant.

PR #3025, opened 2021-09-15
(PE-32844) Deprecate `future.file_paths` config option

beechtom:
This deprecates the `future.file_paths` config option and enables
loading scripts from the `scripts` directory of a project or module by
default.

!deprecation

* **Deprecate `future.file_paths` configuration option**

  The `future.file_paths` configuration option has been deprecated.

!feature

* **Look for scripts in a `scripts` directory**

  Bolt now looks for scripts in a project's or module's `scripts`
  directory when provided a Puppet file path with the form
  `<module>/scripts/<file>`. Previously, this behavior was opt-in only
  by setting the `future.file_paths` configuration option.

### DIFF
--- a/lib/bolt/config/options.rb
+++ b/lib/bolt/config/options.rb
@@ -150,7 +150,9 @@ module Bolt
               description: "Load scripts from the `scripts/` directory of a module.",
               type: [TrueClass, FalseClass],
               _example: true,
-              _default: false
+              _default: false,
+              _deprecation: "Bolt no longer honors this option and enables loading scripts from the scripts "\
+                            "directory by default."
             },
             "script_interpreter" => {
               description: "Use a target's [`interpreters` configuration](bolt_transports_reference.md#interpreters) "\
@@ -161,7 +163,7 @@ module Bolt
             }
           },
           _plugin: false,
-          _example: { 'file_paths' => true, 'script_interpreter' => true }
+          _example: { 'script_interpreter' => true }
         },
         "hiera-config" => {
           description: "The path to the Hiera configuration file.",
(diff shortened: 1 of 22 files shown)


lib/bolt/config/options.rb
```diff
@@ -150,7 +150,8 @@ module Options
               description: "Load scripts from the `scripts/` directory of a module.",
               type: [TrueClass, FalseClass],
               _example: true,
-              _default: false
+              _default: false,
+              _deprecation: true
```
lucywyman, line 154: In thinking more about this, I think the messaging for this is a little tricky. "Deprecation" makes it sound like it's *still* supported, but won't be in the future. I think we could either:
1. Just leave the deprecation warning, it's close enough to the truth
2. Deprecate the option and also raise a warning explaining that it's not honored anymore. This is nice because it'd still be documented.
3. Remove the option altogether. That'll raise a warning that `file_paths` isn't recognized. We could either:
  a. Raise an additional warning explaining it isn't honored anymore
  b. My preference, raise a custom warning that `file_paths` will be ignored, and Bolt will behave as if it's true.

I don't feel too strongly about which  we do, but just "deprecating" the option doesn't feel 100% right.

lucywyman (approved):
Looks like some spec tests still set this:
```
spec/integration/winrm_spec.rb
100:        'file_paths' => true
228:            'file_paths'         => true,

spec/integration/ssh_spec.rb
97:        'file_paths' => true
173:            'file_paths'         => true,
```

Ah, we can also remove:
- `lib/bolt/catalog.rb#L69` setting 'future' during catalog compilation, that was for the file functions

beechtom:
@lucywyman Just updated the deprecation warning to be more clear. Forgot that the validator already had logic for specifying a custom message. 🤷‍♂️

PR #3026, opened 2021-09-21
(PE-32930) Load task dependency files from the scripts directory

ekinanp:

### DIFF
--- a/lib/bolt_server/transport_app.rb
+++ b/lib/bolt_server/transport_app.rb
@@ -463,7 +463,7 @@ module BoltServer
     end
 
     def file_metadatas(versioned_project, module_name, file)
-      abs_file_path = @pal_mutex.synchronize do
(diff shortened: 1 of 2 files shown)


ekinanp:
Don't know if I should be referencing the GH issue in the first commit.

ekinanp:
I've also verified that this is all we'll need on the bolt-server side to get the load scripts bit working for projects. (Verification involved manually scp'ing the changed files in Puppet containing the scripts work over to the primary then restarting bolt-server).

ekinanp:
This PR's blocked on https://tickets.puppetlabs.com/browse/PE-32938

ekinanp:
Looks like https://github.com/puppetlabs/puppet-runtime/pull/496 has been merged so I think this should be unblocked?

lucywyman (approved):

PR #3029, opened 2021-10-04
(bug-fix) Return resolved config object to inventory

lucywyman:
When resolving inventory plugins for transport configuration the
inventory would call `resolve()` on each transport object in it's config
map (which maps the transport name "ssh" to the class
Bolt::Transport::SSH object). Previously the `resolve()` function
returned `nil` since it's last method is to call `validate`, which
caused the inventory value for the config key to be nil instead of the
class itself. This amends the `resolve()` function to return the
modified transport class so that the inventory config map is correctly
updated.

!bug

* **Correctly set inventory config when resolving plugins**

  Previously, Bolt would resolve plugins then set the config in the
  Inventory to `nil`, causing errors like `undefined method 'merge' for
  nil:NilClass`. Bolt now correctly returns the resolved config to the
  inventory.

### DIFF
--- a/lib/bolt/inventory/inventory.rb
+++ b/lib/bolt/inventory/inventory.rb
@@ -71,7 +71,7 @@ module Bolt
           @config
         else
           @config_resolved = true
-          @config.transform_values! { |t| t.resolved? ? t : t.resolve(@plugins) }
+          @config.each_value { |t| t.resolve(@plugins) unless t.resolved? }
         end
       end
 
--- a/spec/unit/inventory/inventory_spec.rb
+++ b/spec/unit/inventory/inventory_spec.rb
@@ -1320,6 +1320,38 @@ describe Bolt::Inventory::Inventory do
     end
   end
 
+  context 'resolving transport config' do
+    let(:config_data) do
+      {
+        'ssh' => {
+          'user' => {
+            '_plugin' => 'mock_plugin'
+          }
+        }
+      }
+    end
+    let(:config)    { make_config(config_data) }
+    let(:inventory) do
+      Bolt::Inventory::Inventory.new({ 'config' => config_data },
+                                     config.transport,
+                                     config.transports,
+                                     plugins)
+    end
+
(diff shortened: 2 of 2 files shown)


lib/bolt/config/transport/base.rb
```diff
@@ -77,6 +77,7 @@ def resolve(plugins)
           @resolved = true
 
           validate
+          self
```
donoghuc, line 80: Is this not a bug in how we are resolving? https://github.com/puppetlabs/bolt/blob/e3a7529dc289db69fa39db6cd629ab8a23f25d2a/lib/bolt/inventory/inventory.rb#L74 It seems like maybe instead of `transform_values` that should just be `each_value` and the `resolve` will mutate the actual instance of the transport config?

donoghuc: Something like this:
```diff
@@ -77,7 +77,6 @@ module Bolt
           @resolved = true
 
           validate
-          self
         end
 
         def resolved?
diff --git a/lib/bolt/inventory/inventory.rb b/lib/bolt/inventory/inventory.rb
index e3327231..bccad996 100644
--- a/lib/bolt/inventory/inventory.rb
+++ b/lib/bolt/inventory/inventory.rb
@@ -71,7 +71,8 @@ module Bolt
           @config
         else
           @config_resolved = true
-          @config.transform_values! { |t| t.resolved? ? t : t.resolve(@plugins) }
+          @config.each_value { |t| t.resolved? ? t : t.resolve(@plugins) }
+          @config
         end
       end
 
```

lucywyman: Ok, I *tried* this first too, and was still seeing the same issue - which seemed weird, but the other solution worked. Of course now it's working, I must have just been doing something else wrong :P

donoghuc (approved):
ah, didnt know `each_value` returns the hash. Nice.

beechtom (approved):

PR #3031, opened 2021-10-05
Permit newer versions of the terminal-table gem

reidmv:
Newer versions have features that are desirable to use in Bolt plans, such as unicode table borders. This commit still allows the older 1.8 version of terminal-table to be used, but opens up the dependency requirement to permit new 3.x versions as well.

If the team is open to it, I'd just as soon update the dependence to `~> 3`. Not actually sure what this gem is used for in Bolt besides being available for the puppet/format module to make use of.


beechtom:
@reidmv terminal-table is used by the human outputter for the `bolt * show` commands. It looks like 3.0+ doesn't break anything, so I have no problem updating the dependency to `~> 3.0`.

Also, can you please add the `!no-release-note` tag to the end of your commit message?

reidmv:
@beechtom tag added, and changed the dependency to a simple twiddle-wakka 3.0 instead of keeping any of the 1.8 stuff.

reidmv:
Not something I'm familiar with.

https://github.com/puppetlabs/puppet-runtime/blob/master/configs/components/rubygem-terminal-table.rb

Is updating this a manual process, or is there another way?

beechtom (approved):
Ya, we'll need to bump the rubygem component in puppet-runtime as well. I can take care of that prior to the next release. For now I'm going to label this as "do not merge" so our CI doesn't fail. :)

beechtom:
Blocked pending updating puppet-runtime to use terminal-table 3.0+ : https://github.com/puppetlabs/puppet-runtime/blob/master/configs/components/rubygem-terminal-table.rb

PR #3041, opened 2021-12-07
(PT-738) Add `policy` subcommands and `policies` config option

beechtom:
This adds the new `policy apply|new|show` subcommands and the `policies` config option.

### DIFF
--- a/lib/bolt/application.rb
+++ b/lib/bolt/application.rb
@@ -413,6 +413,94 @@ module Bolt
       { plugins: plugins.list_plugins, modulepath: pal.user_modulepath }
     end
 
+    # Add a new policy to the project.
+    #
+    # @param name [String] The name of the new policy.
+    # @return [Hash]
+    #
+    def new_policy(name)
+      # Validate the policy name
+      unless name =~ Bolt::Module::CONTENT_NAME_REGEX
+        message = <<~MESSAGE.chomp
+          Invalid policy name '#{name}'. Policy names are composed of one or more name segments
+          separated by double colons '::'.
+
+          Each name segment must begin with a lowercase letter, and can only include lowercase
+          letters, digits, and underscores.
+
+          Examples of valid policy names:
+              - #{@config.project.name}
+              - #{@config.project.name}::my_policy
+        MESSAGE
+
+        raise Bolt::ValidationError, message
+      end
+
+      prefix, *name_segments, basename = name.split('::')
+
+      # Error if name is not namespaced to project
+      unless prefix == @config.project.name
+        raise Bolt::ValidationError,
+              "Policy name '#{name}' must begin with project name '#{@config.project.name}'. Did "\
+              "you mean '#{@config.project.name}::#{name}'?"
+      end
+
+      # If the policy name is just the project name, use the special init.pp class
+      basename ||= 'init'
+
+      # Policies can be saved in subdirectories in the 'manifests/' directory
+      policy_dir = File.expand_path(File.join(name_segments), @config.project.manifests)
+      policy     = File.expand_path("#{basename}.pp", policy_dir)
+
+      # Ensure the policy does not already exist
+      if File.exist?(policy)
+        raise Bolt::Error.new(
+          "A policy with the name '#{name}' already exists at '#{policy}', nothing to do.",
+          'bolt/existing-policy-error'
+        )
+      end
+
+      # Create the policy directory structure in the current project
+      begin
+        FileUtils.mkdir_p(policy_dir)
+      rescue Errno::EEXIST => e
+        raise Bolt::Error.new(
+          "#{e.message}; unable to create manifests directory '#{policy_dir}'",
+          'bolt/existing-file-error'
+        )
+      end
+
+      # Create the new policy
+      begin
+        File.write(policy, <<~POLICY)
+          class #{name} {
+
+          }
+        POLICY
+      rescue Errno::EACCES => e
+        raise Bolt::FileError.new("#{e.message}; unable to create policy", policy)
+      end
+
+      # Update the project configuration to include the new policy
+      project_config = Bolt::Util.read_yaml_hash(@config.project.project_file, 'project config')
+
+      # Add the 'policies' key if it does not exist and de-dupiclate entries
+      project_config['policies'] ||= []
+      project_config['policies'] <<  name
+      project_config['policies'].uniq!
+
+      begin
+        File.write(@config.project.project_file, project_config.to_yaml)
+      rescue Errno::EACCES => e
+        raise Bolt::FileError.new(
+          "#{e.message}; unable to update project configuration",
+          @config.project.project_file
+        )
+      end
+
+      { name: name, path: policy }
+    end
+
     # Initialize the current directory as a Bolt project.
     #
     # @param name [String] The name of the project.
--- a/lib/bolt/application.rb
+++ b/lib/bolt/application.rb
@@ -413,6 +413,60 @@ module Bolt
       { plugins: plugins.list_plugins, modulepath: pal.user_modulepath }
     end
 
+    # Applies one or more policies to the specified targets.
+    #
+    # @param policies [String] A comma-separated list of policies to apply.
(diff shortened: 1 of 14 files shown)


lib/bolt/application.rb
```diff
@@ -413,6 +413,170 @@ def list_plugins
...
+        raise Bolt::Error.new(
+          "The following policies cannot be loaded: '#{unloadable_policies.join("', '")}'. "\
+          "Policies must be a Puppet class saved to a project's or module's manifests directory.",
+          'bolt/unloadable-policy-error'
+        )
+      end
+
+      # Execute a single include statement with all the policies to apply them
+      # to the targets. Yay, reusable code!
+      # CODEREVIEW: Is there a meaningful difference between the one-line include
```
nicklewis, line 467: Nope! One line is perfect.

lib/bolt/project_manager.rb
```diff
@@ -119,7 +119,7 @@ def create(path, name, modules)
         @outputter.stop_spin
       end
 
-      data = { 'name' => project_name }
+      data = { 'name' => project_name, 'profiles' => [] }
```
nicklewis, line 122: Should be `policies`

beechtom: Removed this altogether. Per a discussion in the design doc, we felt that while this feature is experimental we should not add it to projects by default.

nicklewis: Oh yeah that makes way more sense 👍

nicklewis (approved):

lib/bolt/application.rb
```diff
@@ -413,6 +413,170 @@ def list_plugins
...
+      unavailable_policies = policies.reject do |policy|
+        @config.policies.any? do |known_policy|
+          File.fnmatch?(known_policy, policy, File::FNM_EXTGLOB)
+        end
+      end
+
+      if unavailable_policies.any?
+        command = Bolt::Util.powershell? ? 'Get-BoltPolicy' : 'bolt policy show'
+
+        # CODEREVIEW: Phrasing
```
lucywyman, line 436: @hestonhoffman If you get a chance?

lib/bolt/application.rb
```diff
@@ -413,6 +413,170 @@ def list_plugins
...
+      rescue Errno::EEXIST => e
+        raise Bolt::Error.new(
+          "#{e.message}; unable to create manifests directory '#{policy_dir}'",
+          'bolt/existing-file-error'
+        )
+      end
+
+      # Create the new policy
+      begin
+        # CODEREVIEW: Do we want a more detailed, commented out example of what a policy might look like?
```
lucywyman, line 531: I think this is fine for now, that might be something we add later.

lib/bolt/cli.rb
```diff
@@ -287,6 +288,26 @@ def parse
         raise Bolt::CLIError, "Must specify a plan."
       end
 
+      if options[:subcommand] == 'policy'
+        if options[:action] == 'apply' && !options[:object]
```
lucywyman, line 292: Need to add `subcommand policy` to line 348 below for `--noop` to work.

lib/bolt/bolt_option_parser.rb
```diff
@@ -653,6 +669,61 @@ module            Manage Bolt project modules
...
+          policy apply
+
+      #{colorize(:cyan, 'Usage')}
+          bolt policy apply <policy> [options]
+
+      #{colorize(:cyan, 'Description')}
+          Apply a policy to the specified targets.
+    HELP
+
+    # CODEREVIEW: Do we want to have a more detailed description for the command
```
lucywyman, line 701: Definitely not here. If we do anything it should be adding docs and linking to them here like we do for `bolt task` 
```
Documentation
    Learn more about Bolt tasks at http://pup.pt/bolt-tasks.
```

lib/bolt/cli.rb
```diff
@@ -631,6 +652,26 @@ def execute(options)
         outputter.print_plugin_list(**app.list_plugins)
         SUCCESS
 
+      when 'policy'
```
lucywyman, line 655: I mean so much hard work but this code block is beautiful 👨🏼‍🍳 💋

lib/bolt/application.rb
```diff
@@ -413,6 +413,170 @@ def list_plugins
...
+      rescue Errno::EACCES => e
+        raise Bolt::FileError.new("#{e.message}; unable to create policy", policy)
+      end
+
+      # Update the project configuration to include the new policy
+      project_config = Bolt::Util.read_yaml_hash(@config.project.project_file, 'project config')
+
+      # Add the 'policies' key if it does not exist
+      project_config['policies'] ||= []
+      project_config['policies'] <<  name
```
lucywyman, line 546: There's a funny case here where if the *file* doesn't exist, but the policy is still listed here, it'll be included twice. I don't think it's a big deal, and definitely we shouldn't try to de-duplicate globs, but if it bugs you we can de-duplicate here too as well as when listing.

lucywyman (approved):
## Misc
- [x] JSON output works for all commands

## Help
- [x] `policy` shows up in `bolt help` list of subcommands
- [x] `bolt policy --help`
- [x] Warns when using the new subcommands
- [x] PowerShell cmdlets are generated
- [x] PowerShell cmdlets execute the underlying bolt commands

## `bolt policy show`
- [x] Prints what is in `bolt-project.yaml`
- [x] Errors if policies key is missing
- [x] Loads from the specified project path

## `bolt policy new`
- [x] Errors if the policy doesn't match the project name
- [x] Adds policy key if not present
- [x] Deduplicates the available policy list
- [x] Errors if the provided name is invalid
- [x] Supports subdirectories with `module::foo::bar` and creates the directory

## `bolt policy apply`
- [x] Errors if the policy specified doesn't exist
- [x] Errors if the policy specified isn't under `policies` key 
- [x] Loads policies from projects
- [x] Loads policies from the modulepath
- [x] Handles apply_prep failures
- [x] Handles apply failures
- [x] Supports globbing for allowing policies
- [x] Applies multiple policies in order
- [x] Uses `--noop`

I don't feel like this is exhaustive but I think it's close to good enough. Amazing work as always Tom!!

lib/bolt/application.rb
beechtom: Updated to de-duplicate entries.

lib/bolt/application.rb
```diff
@@ -413,6 +413,168 @@ def list_plugins
...
+        end
+      end
+
+      if unavailable_policies.any?
+        command = Bolt::Util.powershell? ? 'Get-BoltPolicy' : 'bolt policy show'
+
+        # CODEREVIEW: Phrasing
+        raise Bolt::Error.new(
+          "The following policies are not available to the project: '#{unavailable_policies.join("', '")}'. "\
+          "Policies must be listed in a project's 'policies' setting before they are applied to targets. "\
```
hestonhoffman, line 439:
```suggestion
          "You must list policies in a project's 'policies' setting before Bolt can apply them to targets. "\
```
Just taking out some passive voice

steveax (approved):

lib/bolt/application.rb
```diff
@@ -413,6 +413,168 @@ def list_plugins
...
+    # @param targets [Array[String]] The list of targets to apply the policies to.
+    # @param noop [Boolean] Whether to apply the policies in no-operation mode.
+    # @return [Bolt::ResultSet]
+    #
+    def apply_policies(policies, targets, noop: false)
+      policies = policies.split(',')
+
+      # Validate that the policies are available to the project.
+      unavailable_policies = policies.reject do |policy|
+        @config.policies.any? do |known_policy|
```
lucywyman, line 428: I don't *think* people will do this, but if you run `be bolt policy apply project::file, -t docker` this stacktraces

lib/bolt/application.rb
lucywyman: I think can probably remove the CODEREVIEWs?

lib/bolt/application.rb
beechtom: Good catch. Updated.

lucywyman (approved):

Issue #3049, opened 2022-01-21
Make rerunfile adjustable at project


crazymind1337:
## Use Case
We are using bolt on our jumphost host to run pre-configured commands on a global project. The bolt-project is place at `/etc/puppetlabs/bolt/<project>`. Each user uses this project with an environment variable set like `BOLT_PROJECT=/etc/puppetlabs/bolt/<project>`.

We switched to a global project path because we had way too many users (~100) to manage each project in their home directories. Since this switch we are not able to use the `rerun`-feature anymore, because the user are not allowed to write the the global project directory and even if they would be allowed, they would overwrite the file on parallel runs.

Therefore I have checked the code and it looks like there is no possibility to configure the path of the `rerunfile`: `lib/bolt/project.rb::110`.

## Describe the Solution You Would Like
I would like to have the possibility to configure the rerunfile at the project file like:
`rerunfile: $HOME/.bolt/rerun.json`

## Describe Alternatives You've Considered
Using an environment variable to set the path of the rerunfile would also be an alternative:
`export BOLT_RERUNFILE=$HOME/.bolt/rerun.json`

## Additional Context
%

crazymind1337:
@beechtom thanks for taking care of this!

PR #3052, opened 2022-02-06
Allow to set a new rerunfile path via environment variable

crazymind1337:
* **Add env var to set rerunfile** ([#3049](https://github.com/puppetlabs/bolt/issues/3049)

  Checks if the environment variable BOLT_RERUNFILE is set.
  If it is set it uses the content of the env var.
  If not is uses the default value within project.


donoghuc:
In general we have avoided adding configuration via environment variable unless there is a very specific reason. Instead additions like this are implemented with a combination of configuration files and/or cli flags. That aside... What is the use case this is trying to solve? Are you trying to write a `rerun` file *outside* of a project? Or do you just want a different name/path for `.rerun.json` *inside* a project?

crazymind1337:
>  In general we have avoided adding configuration via environment variable unless there is a very specific reason. Instead additions like this are implemented with a combination of configuration files and/or cli flags.

I see. In first place I tried to make this config available in the project. But I did not succed. Therefore I tried the simple way of using the env var. I really would prefer the setting via config on project level, but for my coding skills, the code is way too complex.

>  What is the use case this is trying to solve? Are you trying to write a rerun file outside of a project? Or do you just want a different name/path for .rerun.json inside a project?

As @McdonaldSeanp said, thats written in the issue, but I might say it again here:
The project directory is own by root. User cannot write any files there, any even if they would be allowed, they would overwrite the rerunfile.

baurmatt:
This, and the possibility to configure rerunfile in bolt-project.yaml, is also interesting for #3054.

beechtom:
Superseded by #3062

PR #3057, opened 2022-02-15
bolt apply: allow reading manifest from a named pipe

lollipopman:
Prior to this commit bolt threw an error when given a manifest which
resolves to a named pipe:

    $ bolt apply -t butter.example.com <(cat manifests/*.pp)

After this commit bolt allows reading from the named pipes!


lollipopman:
@puppetlabs/skeletor kindly review

lib/bolt/util.rb
```diff
@@ -344,8 +344,8 @@ def validate_file(type, path, allow_dir = false)
 
         if !stat.readable?
           raise Bolt::FileError.new("The #{type} '#{path}' is unreadable", path)
-        elsif !stat.file? && (!allow_dir || !stat.directory?)
-          expected = allow_dir ? 'file or directory' : 'file'
+        elsif !(stat.file? || stat.pipe?) && (!allow_dir || !stat.directory?)
```
nicklewis, line 347: I think in the spirit of "everything is a file", we can change this check to simply `elsif !allow_dir && stat.directory?`. What we really care about is just that we can read from it like a file, and the closest thing to that check we have is just that it's not a directory.

Similarly, it's probably simplest to still just report we expect it to be a "file or directory".

nicklewis (commented):
This is great, I love named pipes! I have just one comment inline about generalizing this logic a bit more.

lib/bolt/util.rb
lollipopman: makes sense to me, I updated the pull request

lollipopman:
> @lollipopman Can you clarify the use case for using named pipes in a `bolt apply` command? The example you provide would not actually work, since `bolt apply` expects a path to a single manifest file, whereas your example passes the code from potentially multiple manifest files.
> 
> ```
> $ bolt apply -t butter.example.com <(cat manifests/*.pp)
> ```

@beechtom thanks for reviewing. For my use case I am concatenating two manifests into a single file, so bolt will read them both as does puppet server in a typical setup. I can attest that at least in my use case it works successfully as bolt happily reads the concatenated file. I sure there are other reasons to support pipes, for instance perhaps you want to grep something out before applying?

```
$ bolt apply -t butter.example.com <(grep -v password site.pp)
```

lib/bolt/util.rb
lollipopman: @nicklewis can you take another look when you have a moment?

donoghuc: Nick will be out for a bit, I think this is a change that we can get in. I just allowed some tests to run on this and some other checks. We will figure out a path forward for getting this in and released.

lollipopman: @donoghuc thanks!

lollipopman: @donoghuc some of the checks flagged errors, but some of them seem unrelated to this commit. Also, happy to add a release note, if desired?

lollipopman: @donoghuc any chance you could take a look at the CI failures?

lollipopman:
@nicklewis any chance you could review?

nicklewis:
This looks good. Can you add a release note to the commit please? Then this should be ready to merge.

The [contributing doc](https://github.com/puppetlabs/bolt/blob/main/CONTRIBUTING.md#pull-requests) has info about the format. I'm also happy to take care of it if you want.

PR #3062, opened 2022-03-07
(GH-3049) Configure path to rerun file

### LINKED ISSUE #3049: Make rerunfile adjustable at project

beechtom:
This adds a new `rerunfile` configuration option that allows users to
configure the path to Bolt's rerun file.

!feature

* **Configure path to the rerun file**
  ([#3049](https://github.com/puppetlabs/bolt/issues/3049))

  You can now configure the path to Bolt's rerun file at the project,
  user, and system-wide levels with the new `rerunfile` config option.

Supersedes #3052
Closes #3049

### DIFF
--- a/lib/bolt/config/options.rb
+++ b/lib/bolt/config/options.rb
@@ -461,6 +461,13 @@ module Bolt
           },
           _plugin: true
         },
+        "rerunfile" => {
+          description: "The path to the project's rerun file. The rerun file is used to store information "\
+                       "about targets from the most recent run. Expands relative to the project directory.",
+          type: String,
+          _example: "/Users/bolt/project/rerun.json",
+          _plugin: true
+        },
         "save-rerun" => {
           description: "Whether to update `.rerun.json` in the Bolt project directory. If "\
                        "your target names include passwords, set this value to `false` to avoid "\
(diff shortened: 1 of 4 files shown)


donoghuc (commented):
1. Should the rerunfile be relative to `@path` if not absolute?
2. Can we not just set the instance variable and expose it with the attribute accessor? I dont quite get why we need a class method to expose it explicitly.

donoghuc (approved):

PR #3067, opened 2022-03-11
(maint) Pin to windows-2019 for gh action runner

donoghuc:
The latest windows runner https://github.com/actions/virtual-environments/issues/4856 appears to be breaking our winrm setup. Pin while we figure that out.

!no-release-note

### DIFF
--- a/.github/workflows/apply.yaml
+++ b/.github/workflows/apply.yaml
@@ -20,7 +20,7 @@ jobs:
     name: Tests
     strategy:
       matrix:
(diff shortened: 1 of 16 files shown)


rileynewton (approved):

donoghuc:
This showed me that the winrm service is in fact configured and running and bolt can (eventually) connect. There is just a change such that whatever is happening with the new image, we need to increase connect timeouts to reliably connect in our tests. opened https://github.com/puppetlabs/bolt/pull/3071 for that effort.

PR #3071, opened 2022-03-15
(maint) Increase winrm connection timeouts for windows-latest

donoghuc:
For any integration test that runs against a live winrm connection, increase the timeout as the new GH runner is slower to connect.

### DIFF
--- a/spec/fixtures/modules/error/plans/winrm_disconnect.pp
+++ b/spec/fixtures/modules/error/plans/winrm_disconnect.pp
@@ -2,6 +2,7 @@ plan error::winrm_disconnect(
   TargetSpec $targets
 ) {
   $result = run_command('restart-service winrm', $targets, _catch_errors => true)
+  wait_until_available($targets, _catch_errors => true)
   return run_command('get-service |? name -like winrm', $targets)
 }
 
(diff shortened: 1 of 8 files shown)


spec/integration/transport/winrm_spec.rb
```diff
@@ -78,7 +78,7 @@ def stub_winrm_to_raise(klass, message)
 
       # The connection should fail immediately; this timeout helps ensure that
       # and avoids a hang
-      Timeout.timeout(3) do
+      Timeout.timeout(45) do
```
donoghuc, line 81: These timeouts probably dont need to be this long. I think maybe the three seconds might have been just a little bit too slow to get the connect error for the closed port.

spec/fixtures/modules/error/plans/winrm_disconnect.pp
```diff
@@ -2,6 +2,7 @@
   TargetSpec $targets
 ) {
   $result = run_command('restart-service winrm', $targets, _catch_errors => true)
+  wait_until_available($targets, _catch_errors => true)
```
donoghuc, line 5: I *think* this is still in line with what this test is for? Not sure how we got away with this for so long... Maybe this is actually exposing a real issue?

donoghuc:
Two green runs in a row now... I think this is ready for some review. Its certainly better than it was.

beechtom (approved):

PR #3072, opened 2022-03-15
(maint) Add option to parse env_var values as json

donoghuc:
I want to use json encoded data fron ENV with the env_var plugin. This adds and option to the plugin to parse values looked up from environment as JSON.

!feature
* **Option to parse data as JSON looked up with `env_var` plugin**

  The env_var plugin now accepts a `parse-as-json` option to allow
  the plugin to use data encoded as JSON stored in environment
  variables.

### DIFF
--- a/documentation/supported_plugins.md
+++ b/documentation/supported_plugins.md
@@ -60,6 +60,7 @@ The following parameters are available to the `env_var` plugin:
 | `var` | **Required.** The name of the environment variable to read from. | `String` | None |
 | `default` | A value to use if the environment variable `var` isn't set. | `String` | None |
 | `optional` | Unless `true`, `env_var` raises an error when the environment variable `var` does not exist.  When `optional` is `true` and `var` does not exist, env_var returns `nil`. | `Boolean` | `false` |
+| `json` | The environment variable value is encoded in a json string, parse it and use the resolved data. | `Boolean` | `false` |
 
 #### Example usage
 
--- a/lib/bolt/plugin/env_var.rb
+++ b/lib/bolt/plugin/env_var.rb
@@ -1,8 +1,16 @@
 # frozen_string_literal: true
 
+require 'json'
 module Bolt
   class Plugin
     class EnvVar
+      class InvalidPluginData < Bolt::Plugin::PluginError
+        def initialize(msg, plugin)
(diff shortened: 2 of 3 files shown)


lib/bolt/plugin/env_var.rb
```diff
@@ -31,7 +32,8 @@ def validate_resolve_reference(opts)
       end
 
       def resolve_reference(opts)
-        ENV[opts['var']] || opts['default']
+        reference = ENV[opts['var']] || opts['default']
+        opts['parse-as-json'] && reference ? JSON.parse(reference) : reference
```
beechtom, line 36: This should have a rescue for `JSON::ParserError` so we can raise a nicer message.

documentation/supported_plugins.md
```diff
@@ -60,6 +60,7 @@ The following parameters are available to the `env_var` plugin:
 | `var` | **Required.** The name of the environment variable to read from. | `String` | None |
 | `default` | A value to use if the environment variable `var` isn't set. | `String` | None |
 | `optional` | Unless `true`, `env_var` raises an error when the environment variable `var` does not exist.  When `optional` is `true` and `var` does not exist, env_var returns `nil`. | `Boolean` | `false` |
+| `parse-as-json` | The environment variable value is encoded in a json string, parse it and use the resolved data. | `Boolean` | `false` |
```
beechtom, line 63: Not a huge fan of verbose option names. Why not just `json`? It's succinct and easier to remember.

nicklewis, line 63: Plugin parameters need to use underscores. Plugins implemented as tasks _must_ use underscores, so we want to be consistent in the pure ruby ones.

lib/bolt/plugin/env_var.rb
```diff
@@ -31,7 +32,8 @@ def validate_resolve_reference(opts)
       end
 
       def resolve_reference(opts)
-        ENV[opts['var']] || opts['default']
+        reference = ENV[opts['var']] || opts['default']
```
nicklewis, line 35: I don't think we want to parse the default. If you want structured data as the default, you should just pass structured data as the value, not a JSON-encoded string version of it.

beechtom (approved):

Issue #2974, opened 2021-08-09
`bolt plan show <plan>` hides untagged description block, unlike Puppet Strings


op-ct:
## Describe the Bug

When a Bolt Plan is documented with a `@summary` tag and a longer, untagged block of descriptive text, `bolt plan show` and `bolt plan show <specific::plan>` only display  the `@summary` text and ignore the content of the untagged block. This behavior is inconsistent with Puppet Strings, which renders the `REFERENCE.md` TOC using the plan's `@summary` **and** the plan's details using the content of the untagged block.

## Expected Behavior
When a plan is documented with both `@summary` and an untagged text block,  `bolt plan show <specific::plan>` should include the content of the untagged text block.

## Steps to Reproduce
1. Review the `canary` plan's inline documentation at https://github.com/puppetlabs/bolt/blob/main/modules/canary/plans/init.pp, noting the `@summary` tag + untagged description block.
2. Run `bolt plan show canary`
3. All of the untagged description block's documentation is missing

## Environment
 - Version 3.13.0

## Additional Context

There don't seem to be any good options to add detailed documentation to Bolt Plans in a way that displays well in both Puppet Strings and `bolt plan show`.

I've pushed an example Bolt project to https://github.com/op-ct/bolt_doc_example.  It contains 3 small plans that demonstrate what goes wrong in `bolt plan show` and/or Puppet Strings' `REFERENCE.md` when rendering plan docs with each combination of:

* [Short `@summary` with untagged description block](https://github.com/op-ct/bolt_doc_example/blob/main/plans/a/mixed_docs.pp)
* [Multi-line `@summary`, without an untagged block](https://github.com/op-ct/bolt_doc_example/blob/main/plans/b/multiline_summary.pp)
* [Untagged description block, without `@summary`](https://github.com/op-ct/bolt_doc_example/blob/main/plans/c/untagged_only.pp)

Each plan's docs summarize what happens, and the resulting [`REFERENCE.md`][0] shows how Puppet Strings renders each example.  You can validate how bolt renders each plan's docs with `bolt plan show` and `bolt plan show <plan::name>`.

[0]: https://github.com/op-ct/bolt_doc_example/blob/main/REFERENCE.md

lucywyman:
[...]

PR #3077, opened 2022-04-08
(GH-2974) Include plan docstring in `plan show` output

beechtom:
This updates the `bolt plan show <PLAN>` and `Get-BoltPlan -Name <PLAN>`
commands to include a plan's docstring in the output. Previously, these
commands would show just a plan's summary when a `@summary` tag was
used, or the plan's docstring if the `@summary` tag was not used. This
resulted in Bolt not showing all of a plan's documentation when using
these commands (see the `canary` plan for an example).

!bug

* **Include a plan's docstring in `plan show` output**
  ([#2974](https://github.com/puppetlabs/bolt/issues/2974))

  The `bolt plan show <PLAN>` and `Get-BoltPlan -Name <PLAN>` commands
  now include a plan's docstring in the output, even when the plan
  includes a `@summary` tag. Previously, Bolt would only display a
  plan's summary or docstring, but not both.


nicklewis (approved):
I don't really understand what's going on here from a puppet-strings point of view but it works!

PR #3081, opened 2022-04-17
Dockerfile: Install gems in parallel

bastelfreak:
This allows bundler to use all CPU cores during gem installation.

### DIFF
--- a/Dockerfile.bolt-server
+++ b/Dockerfile.bolt-server
@@ -10,7 +10,7 @@ RUN mkdir /bolt-server
 # Gemfile requires gemspec which requires bolt/version which requires bolt
 ADD . /bolt-server
 WORKDIR /bolt-server
-RUN bundle install --no-cache --path vendor/bundle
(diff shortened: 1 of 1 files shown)


Dockerfile.bolt-server
```diff
@@ -10,7 +10,7 @@ RUN mkdir /bolt-server
 # Gemfile requires gemspec which requires bolt/version which requires bolt
 ADD . /bolt-server
 WORKDIR /bolt-server
-RUN bundle install --no-cache --path vendor/bundle
+RUN bundle install --no-cache --path vendor/bundle --jobs=$(nproc)
```
nicklewis, line 13: This will conflict with #3083 but :+1: for the change.

bastelfreak: yes, I will rebase one after the merge. I thought splitting it up into multiple PRs gives a better history.

beechtom (approved):

PR #3083, opened 2022-04-17
Dockerfile: dont use legacy bundler syntax

bastelfreak:
the old --path option throws a warning.

### DIFF
--- a/Dockerfile.bolt-server
+++ b/Dockerfile.bolt-server
@@ -10,7 +10,8 @@ RUN mkdir /bolt-server
 # Gemfile requires gemspec which requires bolt/version which requires bolt
 ADD . /bolt-server
 WORKDIR /bolt-server
-RUN bundle install --no-cache --path vendor/bundle --jobs=$(nproc)
+RUN bundle config set path 'vendor/bundle' \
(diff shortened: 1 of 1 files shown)


Dockerfile.bolt-server
```diff
@@ -10,7 +10,8 @@ RUN mkdir /bolt-server
 # Gemfile requires gemspec which requires bolt/version which requires bolt
 ADD . /bolt-server
 WORKDIR /bolt-server
-RUN bundle install --no-cache --path vendor/bundle
+RUN bundle config set path 'vendor/bundle' \
```
nicklewis, line 13: I think this needs `&&`? `--no-cache` is _also_ deprecated, so seems like we should set that here as well.

bastelfreak: Added the `&&`. I'm not sure about no-cache. That option is a bit unusual because it doesn't accept any values. Also I didn't see any deprecation message during bundler.

beechtom:
@bastelfreak With #3081 merged this is ready for a rebase.

bastelfreak:
@beechtom should be good now.

beechtom (approved):

PR #3084, opened 2022-04-18
Document testing plans with PuppetDB functions

seanmil:
Add some basic documentation explaining how to stub/mock calls to
the puppetdb_* functions when unit testing Plans that use them.

### DIFF
--- a/documentation/testing_plans.md
+++ b/documentation/testing_plans.md
@@ -450,6 +450,39 @@ it 'configures servers and databases' do
 end
 ```
 
+### Stubbing and Mocking PuppetDB Calls
+
+Plans that use the `puppetdb_*` set of functions can stub and mock values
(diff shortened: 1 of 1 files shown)


seanmil:
Note that this documentation is what I was able to piece together by poking around in the Bolt code and trial and error on the rspec side. It seems to work for me in my environment.

bastelfreak (approved):

donoghuc (approved):
Thanks for adding this documentation. We note that the mock client "is provided as a class so expectations can be placed on it"  in the source but don't explicitly say which methods would need to be mocked 😅 
 https://github.com/puppetlabs/bolt/blob/1343c12b819a001182e83d4e0758f1911f475566/lib/bolt_spec/plans.rb#L104-L111

Issue #1527, opened 2020-01-13
Support multiple PuppetDB connections in inventory.yaml


tspeigner:
## Use Case
Currently you can only define the puppetdb connection information in the bolt.yaml configuration.  It is currently designed to only allow one db configuration.  It would be nice if it could be added to the inventory file as part of the config for the plugin on targets.

## Describe the Solution You Would Like
Something like this in the inventory file:

`version: 2

groups:

  - name: cluster1

    targets:

      - _plugin: puppetdb

        query: "inventory[certname] {}"

    config:

      puppetdb:

        server_urls: ["https://cluster1.domain:8081"]

        cacert: /home/user/.puppetlabs/cluster1.ca.crt

        token: /home/user/.puppetlabs/cluster1.token

      transport: pcp

      pcp:

        service-url: "https://cluster1.domain:8143"

        cacert: /home/user/.puppetlabs/cluster1.ca.crt

        token-file: /home/user/.puppetlabs/cluster1.token

  - name: cluster2

    targets:

      - _plugin: puppetdb

        query: "inventory[certname] {}"

    config:

      puppetdb:

        server_urls: ["https://cluster2.domain:8081"]

        cacert: /home/user/.puppetlabs/cluster2.ca.crt

        token: /home/user/.puppetlabs/cluster2.token

      transport: pcp

      pcp:

        service-url: "https://cluster2.domain:8143"

        cacert: /home/user/.puppetlabs/cluster2.ca.crt

        token-file: /home/user/.puppetlabs/cluster2.token`

adreyer:
This won't work well because the puppetdb instance is configured on the group and won't be available to the plugin. Further while it could be used by the `facts` plan and during apply it would not "just work" for most uses of the `pdb_query` function or `--query`

I can see two ways to implement this:
1. Support it just for populating groups by making it an argument to the plugin.
ie
```
version: 2
groups:
  - name: cluster1
    targets:
      - _plugin: puppetdb
        query: "inventory[certname] {}"
        connection:
          server_urls: ["https://cluster1.domain:8081"]
          cacert: /home/user/.puppetlabs/cluster1.ca.crt
          token: /home/user/.puppetlabs/cluster1.token
      transport: pcp
      pcp:
        service-url: "https://cluster1.domain:8143"
        cacert: /home/user/.puppetlabs/cluster1.ca.crt
        token-file: /home/user/.puppetlabs/cluster1.token
```

[...]

PR #3092, opened 2022-05-18
(GH-1527) Connect to multiple PuppetDB instances

beechtom:
This adds support for connecting to multiple PuppetDB instances and specifying a named PuppetDB instance whenever Bolt connects to PuppetDB.

See commits for more info.


beechtom:
Questions:

1. Does it make sense to have a separate `puppetdb-instances` option, or should it be `puppetdb.instances` instead?
2. Does specifying a named instance in each of the places you can specify one (plan functions, command line, plugin) seem intuitive?
3. Should named instances pick up default config from puppetdb.conf?
4. Should named instances pick up default config from the `puppetdb` section?
5. Should there be a 'default' instance? If yes, how is that determined? (Currently it's whatever is under `puppetdb`.)

donoghuc:
1. Does it make sense to have a separate `puppetdb-instances` option, or should it be `puppetdb.instances` instead?
 - My vote is separate. Having it be "under" puppetdb could be confusing if we are not "sharing/merging" config. 
3. Does specifying a named instance in each of the places you can specify one (plan functions, command line, plugin) seem intuitive?
- Yes, i think what you have here is perfect (if you dont use the feature at all you wont even notice this change)
4. Should named instances pick up default config from puppetdb.conf?
- No, this should only apply to the "default" instance (assuming it already does)
6. Should named instances pick up default config from the `puppetdb` section?
- No, each instance should be complete and independent (if there are named conflicts, the one at the most specific level should win, no merging). 
8. Should there be a 'default' instance? If yes, how is that determined? (Currently it's whatever is under `puppetdb`.)
- Yes, I think this is the best way to support the existing workflows for this and provide the new features for those users who choose to take it.

bolt-modules/boltlib/lib/puppet/functions/puppetdb_query.rb
```diff
@@ -16,15 +18,35 @@
...
+
+  def make_query_with_instance(query, instance)
     puppetdb_client = Puppet.lookup(:bolt_pdb_client)
     # Bolt executor not expected when invoked from apply block
     executor = Puppet.lookup(:bolt_executor) { nil }
     # Send Analytics Report
     executor&.report_function_call(self.class.name)
 
-    puppetdb_client.make_query(query)
+    puppetdb_client.make_query(query, instance)
```
nicklewis, line 50: This change will cause trouble in PE where we only support the `make_query(query)` signature. We could either implement support for this new signature (with an error, since `instance` doesn't make sense in that context) or fork this function in PE and add the error there.

documentation/bolt_connect_puppetdb.md
```diff
@@ -91,6 +94,133 @@ puppetdb:
...
+To specify a PuppetDB instance to the `puppetdb_*` plan functions, pass the
+PuppetDB instance name as the last positional argument to the function:
+
+```puppet
+plan example() {
+  puppetdb_fact(['host.example.com'], 'instance-1')
+}
+```
+
+To specify a PuppetDB instance to the `apply` plan function, use the `_pdb_instance`
```
donoghuc, line 138: Should this just be `_puppetdb` to match the cli option?

donoghuc, line 138: Should this just be `_puppetdb` to match the cli option?

lib/bolt/plugin.rb
```diff
@@ -151,7 +151,8 @@ def initialize(config, pal, analytics = Bolt::Analytics::NoopClient.new, load_pl
         msg = "Configuration for the PuppetDB plugin must be in the 'puppetdb' config section, not 'plugins'"
         raise Bolt::Error.new(msg, 'bolt/plugin-error')
       end
-      @unresolved_plugin_configs['puppetdb'] = config.puppetdb if config.puppetdb
+      @unresolved_plugin_configs['puppetdb'] = config.puppetdb.merge('default'   => config.default_puppetdb,
```
donoghuc, line 154: What was that `if config.puppetdb` guarding against? Is there a case that could be `nil`?

donoghuc, line 154: What was that `if config.puppetdb` guarding against? Is there a case that could be `nil`?

beechtom: I think it's just stale code. The default config for `puppetdb` is `{}`, so `config.puppetdb` will never be `nil`.

documentation/bolt_connect_puppetdb.md
beechtom: Updated. Also added some documentation I missed to the plan functions page.

lib/bolt/plugin.rb
donoghuc: cool, that looks correct. thanks.

donoghuc: cool, that looks correct. thanks.

donoghuc (approved):
Connected this to a couple PE instances. Everything works as advertised. Very cool. 😎

Issue #2767, opened 2021-04-06
Cannot install bolt with ruby v3.0.0


renich:
## Describe the Bug
When trying to install bolt in Fedora 34; with ruby 3.0.0, iit fails; informing me about the ruby version required by ruby 3.5.0.

## Expected Behavior
It should just work.

## Steps to Reproduce
1. gem -Vu install bolt

## Environment
 - Version 3.5.0
 - Platform Fedora 34

## Additional Context
Here's what happens when I run it:

```
[renich@introdesk ~]$ gem install -Vu bolt
HEAD https://index.rubygems.org/
200 OK
GET https://index.rubygems.org/info/bolt
200 OK
Successfully installed bolt-0.0.1
1 gem installed

[renich@introdesk ~]$ gem install -Vuv 3.5.0 bolt
HEAD https://index.rubygems.org/
200 OK
GET https://index.rubygems.org/info/bolt
200 OK
ERROR:  Error installing bolt:
	bolt-3.5.0 requires Ruby version ~> 2.5. The current ruby version is 3.0.0.0.
```

lucywyman:
Hi! Thanks for raising this. As you might have seen Bolt supports Ruby 2.5+ and is developed, built, tested, and ships with Ruby 2.7. We need to maintain support for 2.5+ because Bolt is a dependency of many other Puppet projects that are on the Ruby 2.x series, and adding support for Ruby 3.0 (which presumably has breaking changes with Ruby 2.x, given it's a major version bump) would be difficult. This will probably happen eventually, but I don't see it being prioritized anytime soon.

In the meantime, like I said Bolt ships with it's own Ruby that should work alongside your system Ruby. We're typically pretty good at picking up new versions as soon as our internal CI allows, hopefully we'll have Fedora 33 available soon (it looks like Fedora 34 won't be released until about 2 weeks from now?). You could try installing the Fedora 32 package https://puppet.com/docs/bolt/latest/bolt_installing.html#install-bolt-on-fedora and see if it works, I know users have done that in the past successfully.

renich:
I hope you do soon.

lucywyman:
Ok, after more consideration we think it's worth the effort to add a CI matrix cell for Ruby 3.0 and loosen the restrictions. We'll make sure that Bolt works with Ruby 3 first, and if it does update the gem file.

lucywyman:
[...]

PR #3093, opened 2022-05-23
(GH-2767) Allow Bolt to run with Ruby 3.0

donoghuc:
Previously, our gemspec limited the Ruby version Bolt could run with to 2.x,
as we didn't know if there would be breaking changes in Ruby 3. Now
that 3 has been released and there don't appear to be any changes that
will make Bolt unusable with Ruby 3, we've modified the gemspec to allow
any Ruby greater than 2.5. Bolt will continue to ship with Ruby 2.7, so
this only applies to gem installations. This adds test matrices for Ruby
3 to the unit and integration tests.

!feature

* **Bolt can now be run with Ruby 3** ([#2767](https://github.com/puppetlabs/bolt/issues/2767))
  Required Ruby version for running Bolt is now simply greater than 2.5,
  rather than between 2.5 and 3.0, allowing Bolt gem installs to run with
  Ruby 3.


.github/workflows/linux.yaml
```diff
@@ -12,13 +12,16 @@ jobs:
   integration:
     name: Integration
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        ruby: [2.7, 3.0]
```
bastelfreak, line 17: since Ruby 3.1 is already out, should that be added as well?

donoghuc:
https://github.com/puppetlabs/puppetlabs-ruby_task_helper/pull/21 is another piece of this puzzle. Will need to get a release out of that and take it up to unblock this.

.github/workflows/publish-powershell-module.yaml
```diff
@@ -12,6 +12,9 @@ jobs:
   publish_module:
     name: Publish
     runs-on: windows-latest
+    strategy:
+      matrix:
+        os: [ubuntu-latest, windows-latest]
```
beechtom, line 17: This isn't needed.

.github/workflows/schemas.yaml
```diff
@@ -21,13 +21,16 @@ jobs:
   generate:
     name: Generate
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        ruby: [2.7, 3.1]
```
beechtom, line 26: The changes in this file aren't necessary as it just generates the schemas and ensures they match the ones in the commit history.

.github/workflows/pwsh.yaml
```diff
@@ -12,14 +12,16 @@ jobs:
 
   test:
     name: Test Generated PowerShell Cmdlets
-    runs-on: windows-latest
+    strategy:
+      matrix:
+        ruby: [2.7, 3.1]
```
beechtom, line 17: The changes in this file shouldn't be necessary since it's just testing the generated cmdlets.

.github/workflows/modules.yaml
```diff
@@ -28,6 +31,9 @@ jobs:
   windows:
     name: Windows
     runs-on: windows-latest
+    strategy:
+      matrix:
+        ruby: [2.7, 3.0]
```
beechtom, line 36:
```suggestion
        ruby: [2.7, 3.1]
```

Puppetfile
```diff
@@ -27,7 +27,7 @@ mod 'puppetlabs-powershell_task_helper', '0.1.0'
 mod 'puppetlabs-puppet_conf', '1.2.0'
 mod 'puppetlabs-python_task_helper', '0.5.0'
 mod 'puppetlabs-reboot', '4.1.0'
-mod 'puppetlabs-ruby_task_helper', '0.6.0'
+mod 'puppetlabs-ruby_task_helper', '0.6.1'
```
beechtom, line 30: Can you add a release note to this commit that the `ruby_task_helper` module was updated to `0.6.1`?

beechtom (changes requested):

lib/bolt/applicator.rb
```diff
@@ -162,7 +162,13 @@ def compile(target, scope)
 
     def validate_hiera_config(hiera_config)
       if File.exist?(File.path(hiera_config))
-        data = File.open(File.path(hiera_config), "r:UTF-8") { |f| YAML.safe_load(f.read, [Symbol]) }
+        data = File.open(File.path(hiera_config), "r:UTF-8") do |f|
+          if Psych.method(:safe_load).parameters.flatten.include?(:permitted_classes)
```
nicklewis, line 166: I think instead of `flatten.include?` we can just do `[...].parameters.rassoc(:permitted_classes)`

beechtom (approved):

Issue #3063, opened 2022-03-09
Tasks run over pcp do not support `noop`


donoghuc:
## Describe the Bug
Given bolt does not know the metadata of the task deployed on the primary this check https://github.com/puppetlabs/bolt/blob/3726f0927d4ae9559bd9c5141702c6246b3ceb85/bolt-modules/boltlib/lib/puppet/functions/run_task.rb#L126 will fail when attempting to pass through `noop` (even when a task does support noop). 

## Expected Behavior
When using pcp transport, options that depend on metadata are passed through without failing. 

## Steps to Reproduce
`puppet task run puppet_agent::install -t foo --noop`

## Additional Context
I suppose the best thing would be to expose an endpoint bolt can talk to to get the metadata. 

This shows a hacky workaround:

```diff
diff --git a/bolt-modules/boltlib/lib/puppet/functions/run_task.rb b/bolt-modules/boltlib/lib/puppet/functions/run_task.rb
index f5138ced..73a9675a 100644
--- a/bolt-modules/boltlib/lib/puppet/functions/run_task.rb
+++ b/bolt-modules/boltlib/lib/puppet/functions/run_task.rb
@@ -76,8 +76,10 @@ Puppet::Functions.create_function(:run_task) do

     options[:description] = description if description

+    # Tasks run over pcp wont have any "real" metadata
+    pcp_transport = targets.all? { |t| t.transport == 'pcp' }
     # Don't bother loading the local task definition if all targets use the 'pcp' transport.
-    if !targets.empty? && targets.all? { |t| t.transport == 'pcp' }
+    if !targets.empty? && pcp_transport
       # create a fake task
       task = Bolt::Task.new(task_name, {}, [{ 'name' => '', 'path' => '' }])
     else
@@ -123,7 +125,7 @@ Puppet::Functions.create_function(:run_task) do
     # executor.noop is set when run task is called from the CLI
     # options[:noop] is set when it's called from a plan
     if executor.noop || options[:noop]
-      if task.supports_noop
+      if task.supports_noop || pcp_transport
         params['_noop'] = true
       else
         raise with_stack(:TASK_NO_NOOP, 'Task does not support noop')
```

mcatkpncom:
I have the exact same issue with NOOP

PR #3094, opened 2022-05-24
(GH-3063) Support running tasks in no-op mode over pcp

beechtom:
Tasks run over pcp can now be run in no-op mode. Previously, running
tasks in no-op mode would cause Bolt to error with a message saying
the task does not support running in noop mode as task metadata is not
available when running over the pcp transport. Now, if the task is run
in no-op mmode, the `_noop` metaparameter will be added to the task
input before it is run.

!bug

* **Support running tasks in no-op mode over pcp**
  ([#3063](https://github.com/puppetlabs/bolt/issues/3063))

  Bolt no longer errors when running tasks in no-op mode over the pcp
  transport.


bolt-modules/boltlib/lib/puppet/functions/run_task.rb
```diff
@@ -77,7 +77,7 @@ def run_task_with_description(task_name, targets, description, args = {})
     options[:description] = description if description
 
     # Don't bother loading the local task definition if all targets use the 'pcp' transport.
-    if !targets.empty? && targets.all? { |t| t.transport == 'pcp' }
+    if (pcp_only = targets.any? && targets.all? { |t| t.transport == 'pcp' })
```
donoghuc, line 80: I wish there was a more idomatic way to do this in ruby...

bolt-modules/boltlib/lib/puppet/functions/run_task.rb
```diff
@@ -123,7 +123,9 @@ def run_task_with_description(task_name, targets, description, args = {})
     # executor.noop is set when run task is called from the CLI
     # options[:noop] is set when it's called from a plan
     if executor.noop || options[:noop]
-      if task.supports_noop
+      # If using the pcp transport, we don't have the task metadata. Set the noop
+      # metaparameter in this case.
+      if task.supports_noop || pcp_only
```
donoghuc, line 128: I'm starting to think we should just do this in the block where we create the fake task (and alway support noop there by specifying it in the fake task metadata). That better encapsulates that bit of logic and does not bleed in to this part of the code as much.

bolt-modules/boltlib/lib/puppet/functions/run_task.rb
nicklewis: What if we move the `if targets.empty?` check up near the top of the function. All this intermediate logic is superfluous in that case anyway so it makes sense for it to be an early test.

bolt-modules/boltlib/lib/puppet/functions/run_task.rb
beechtom: That makes way more sense. :) Done.

donoghuc (approved):
niceeee

PR #3096, opened 2022-06-03
(BOLT-1586) Remove ssh role from bolt controller in acceptance tests

beechtom:
This updates the acceptance test Rakefile to not add the `ssh` role to
the Bolt controller. Previously, the `ssh` role was added to the Bolt
controller, which caused some tests run on the controller to connect
over SSH. This caused problems when installing the puppet-agent package
on the controller during acceptance tests, as the
`puppet_agent::install` task added logic to not install the puppet-agent
package when a gem install of the agent is detected in
`puppet_agent >= 4.10.0`. Removing the `ssh` role results in Bolt using
the puppet gem instead of the puppet-agent package.

---

Run the following command from the `acceptance/` directory to see the failing tests:

```shell
bundle exec rake test:preserve
```

Then run the following command to see the tests passing with this change:

```shell
GIT_FORK='beechtom/bolt' GIT_BRANCH='1586/acceptance' bundle exec rake test:preserve
```

### DIFF
--- a/Puppetfile
+++ b/Puppetfile
@@ -6,7 +6,7 @@ moduledir File.join(File.dirname(__FILE__), 'modules')
 
 # Core modules used by 'apply'
 mod 'puppetlabs-service', '2.2.0'
-mod 'puppetlabs-puppet_agent', '4.9.0'
+mod 'puppetlabs-puppet_agent', '4.11.0'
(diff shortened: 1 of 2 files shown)


beechtom:
This work also includes a change for Jenkins: https://github.com/puppetlabs/ci-job-configs/pull/8438

Issue #2618, opened 2021-02-03
Return catalog in ApplyResult


adreyer:
## Use Case
Debugging dependency issues
uploading a catalog to be run by the agent. 

## Describe the Solution You Would Like
The ApplyResult should include the the catalog that was applied in addition to the report.
Should this always be included? Should it be inlcuded by default or only when an option is passed?

PR #3102, opened 2022-06-07
(GH-2618) Add `catalog` to `ApplyResult` objects

beechtom:
This adds a `catalog` attribute to the `ApplyResult` object, which
includes the catalog used during an apply. The catalog is masked using
the `Sensitive` data type to prevent any sensitive information from
being printed to the console or logs.

!feature

* **Include catalog in `ApplyResult` objects**
  ([#2618](https://github.com/puppetlabs/bolt/issues/2618))

  The `ApplyResult` object now includes a `catalog` attribute that
  includes the catalog used during an apply. The catalog is masked using
  the `Sensitive` data type to prevent sensitive information in the
  catalog from being printed to the catalog or Bolt logs.

### DIFF
--- a/bolt-modules/boltlib/lib/puppet/datatypes/applyresult.rb
+++ b/bolt-modules/boltlib/lib/puppet/datatypes/applyresult.rb
@@ -4,6 +4,13 @@
 # returns an `ApplyResult`. An `ApplyResult` is part of a `ResultSet` object and
 # contains information about the apply action.
 #
+# @param catalog
+#   The Puppet catalog used to configure the target. The catalog describes the desired state
+#   of the target. The catalog is masked with the `Sensitive` data type to protect any
+#   sensitive information in the catalog from being printed to the console or logs. To access
+#   the catalog, unwrap it using `$apply_result.catalog.unwrap`. For more information
+#   about catalogs and the catalog compilation process, see [Catalog
+#   compilation](https://puppet.com/docs/puppet/latest/subsystem_catalog_compilation.html).
 # @param report
 #   The Puppet report from the apply action. Equivalent to calling `ApplyResult.value['report']`.
 #   The report is a hash representation of the [`Puppet::Transaction::Report`
@@ -29,6 +36,7 @@
 Puppet::DataTypes.create_type('ApplyResult') do
   interface <<-PUPPET
     attributes => {
+      'catalog' => Sensitive[Hash[String[1], Data]],
       'report' => Hash[String[1], Data],
       'target' => Target
     },
--- a/bolt-modules/boltlib/lib/puppet/datatypes/applyresult.rb
+++ b/bolt-modules/boltlib/lib/puppet/datatypes/applyresult.rb
@@ -4,13 +4,6 @@
 # returns an `ApplyResult`. An `ApplyResult` is part of a `ResultSet` object and
 # contains information about the apply action.
 #
-# @param catalog
-#   The Puppet catalog used to configure the target. The catalog describes the desired state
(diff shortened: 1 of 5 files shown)


donoghuc:
For the release note:
> includes the catalog used during an apply. The catalog is masked using
the Sensitive data type to prevent sensitive information in the
catalog from being printed to the catalog or Bolt logs.

I think there is a subtle, but confusing typo in the last sentence: "being printed to the **catalog** or bolt logs" should be "being printed to the **console** or bolt logs" ?

beechtom:
@donoghuc  Totally fine closing this if it's going to be difficult to add in PE. :) I was just looking through the backlog for future work and thought this would be an easy win.

donoghuc:
I think its absolutely do-able in PE. I like the feature. It will just be a fun journey of serialization between ruby, clojure, and JSON ;)

nicklewis:
The extension of support for sensitive objects to more than the `_sensitive` keyword may cause trouble in PE. I think we currently avoid saving the `_sensitive` portion of task results, but with this implementation we would probably wind up saving the catalog to the database, which is a bit concerning. I think the correct thing to do here is actually to put the `catalog` key _under_ `_sensitive`. That matches the intended use case of `_sensitive`.

For ease of access, we could consider introducing an `unwrap()` function to Result that basically unwraps and merges the keys from `_sensitive` with the rest of the result. Then you would access this field like `$result.unwrap['catalog']` rather than `$result.sensitive.unwrap['catalog']`. Or of course you could just assign the unwrapped version to a new variable.

beechtom:
Updated to store the catalog under `_sensitive.catalog` and have the `catalog` function unwrap the sensitive value and just return the catalog.

bolt-modules/boltlib/lib/puppet/datatypes/applyresult.rb
```diff
@@ -33,6 +41,7 @@
       'target' => Target
     },
     functions => {
+      catalog => Callable[[], Optional[Hash[String[1], Data]]],
```
nicklewis, line 44: Making this a function is a little weird, because it means the `ApplyResult` object can't round trip through JSON. Of course it couldn't anyway because the catalog is sensitive. But it does seem to break one of the rules of pcore data types, which is that functions should only return data that can be derived from the attributes. In this case, we effectively sideload some additional data into the report that is not one of the attributes. I'm not sure what to make of that though.

Separately, I wonder if the specificity of this type declaration has a performance impact. When the function is called, puppet should verify that the result conforms to the type. In this case, that means verifying that every hash value is Data, which means traversing the entire catalog down to the leaves. Since the catalog is quite a large object, that could be significant. Of course, it only affects users who actually _call_ the `catalog()` function, so it's not a critical problem. It would be interesting to test though. If it's a problem, just declaring this as a `Hash` ought to be sufficient.

beechtom: @nicklewis Wouldn't that also apply to `error`? Or am I misunderstanding what is considered an attribute in this case?

I agree that just declaring the catalog a `Hash` would be sufficient in this case.

nicklewis: Hmm yes that would also apply to `error`. We do return `error` from `_pcore_init_hash`, so I wonder why that doesn't cause a problem when we try to serialize and deserialize an ApplyResult. For instance, if you run multiple `apply()` blocks in a single plan, such that the result of one is available within the next one.

nicklewis (approved):

PR #3103, opened 2022-06-08
(BOLT-1570) Support `extensions` in `local` transport on Windows

beechtom:
This adds support for the `local.extensions` transport config option
when running Bolt from a Windows controller. Previously, only the
default extensions (`.rb`, `.pp`, `.ps1`) and any extensions listed
under `local.interpreters` were supported.

!feature

* **Support `extensions` in `local` transport on Windows controllers**

  The `local` transport now supports the `extensions` config option when
  running Bolt on a Windows controller.


lib/bolt/config/transport/local.rb
```diff
@@ -10,6 +10,7 @@ class Local < Base
         WINDOWS_OPTIONS = %w[
           bundled-ruby
           cleanup
+          extensions
```
beechtom, line 13: This makes `extensions` available when running on a *nix controller, it just won't do anything. We _could_ limit this to just Windows controllers, but if the project is run on both Windows and *nix then an "unknown config option" warning would show up when run on *nix.

nicklewis (approved):

lib/bolt/config/transport/local.rb
donoghuc: I think this is fine.

donoghuc (approved):

PR #3105, opened 2022-06-13
Support extended globs for target matching

jay7x:
!feature

* **Support extended globs for target matching**
  Bolt now supports extended globs for target matching in CLI
  `--targets` option.

### DIFF
--- a/lib/bolt/inventory/inventory.rb
+++ b/lib/bolt/inventory/inventory.rb
@@ -8,6 +8,12 @@ module Bolt
     class Inventory
       attr_reader :plugins, :source, :targets, :transport
 
+      # Getting targets from the inventory using '--targets' supports extended glob pattern
+      # matching. In this case, use the extended regex so Bolt only uses commas outside
+      # brackets and braces as delimiters.
+      EXTENDED_TARGET_REGEX = /[[:space:],]+(?=[^\]}]*(?:[\[{]|$))/.freeze
+      TARGET_REGEX          = /[[:space:],]+/.freeze
+
       class WildcardError < Bolt::Error
         def initialize(target)
           super("Found 0 targets matching wildcard pattern #{target}", 'bolt.inventory/wildcard-error')
@@ -85,8 +91,8 @@ module Bolt
       # alias for analytics
       alias node_names target_names
 
-      def get_targets(targets)
-        target_array = expand_targets(targets)
+      def get_targets(targets, ext_glob: false)
+        target_array = expand_targets(targets, ext_glob: ext_glob)
         if target_array.is_a? Array
           target_array.flatten.uniq(&:name)
         else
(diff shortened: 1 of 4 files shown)


beechtom:
@jay7x Thanks for the contribution to Bolt! Please make sure to sign our CLA (more information above), and we'll take a look at getting this reviewed as soon as we can. :)

jay7x:
Checking wrt CLA with my company.. I guess it should be ok but better to recheck always..

jay7x:
I just signed the CLA JFYI :)

spec/unit/inventory/inventory_spec.rb
```diff
@@ -285,11 +285,26 @@ def get_target(inventory, name, alia = nil)
...
+
+        it 'should match wildcard selector []' do
+          targets = inventory.get_targets('target[0-9]')
+          expect(targets.map(&:name).sort).to eq(%w[target1 target2 target3 target4 target5 target6 target7 target9])
+        end
+
+        it 'should match wildcard selector {}' do
+          targets = inventory.get_targets('{target,node}[0-9]')
+          expect(targets.map(&:name).sort).to eq(%w[target1 target2 target3 target4 target5 target6 target7 target9])
+        end
```
beechtom, line 306: This test doesn't confirm that using the `{}` wildcard selector works, since the inventory doesn't include any targets named `node[0-9]`. I'd add at least one node with a matching name to the inventory (starting on line 104) and update this test to include that node in the expected list.

I'd also add a test that uses a combination of glob patterns with commas and target names (e.g. `{target,node}[1-3],target9` or similar) to test that the regex on line 150 works as expected.

lib/bolt/inventory/inventory.rb
```diff
@@ -140,7 +146,8 @@ def expand_targets(targets)
           targets.map { |tish| expand_targets(tish) }
         when String
           # Expand a comma-separated list
-          targets.split(/[[:space:],]+/).reject(&:empty?).map do |name|
+          # Regex magic below is required to workaround `{foo,bar}` glob syntax
+          targets.scan(/[^\s,]*(?:(?<={).*?(?=}))[^\s,]*|[^\s,]*/).reject(&:empty?).map do |name|
```
beechtom, line 150: Let's use a named constant (e.g. `TARGET_REGEX` or similar) and use that here instead of inline regex.

beechtom (changes requested):
Couple small changes, but overall this looks good and works as expected. 👍

jay7x:
I found weird issue with `{target,node}[0-3,9]` pattern which breaks the parsing.. It seems my regex knowledge is limited here as I cannot imagine any short-enough regex which can work for the case.. If it was C language I'd just do parsing by iterating over the string with FSM but I _hope_ Ruby has some special magic for this..

beechtom:
@jay7x Good catch on those strings not getting parsed correctly. 👍 

Instead of matching tokens, it'll likely be easier to use any commas not inside a wildcard as a delimiter. The regex for that is a lot simpler and _should_ cover those more complex examples.

Additionally, after talking with the team, we'll want to limit support for extended glob patterns just to the CLI. This change would also affect plans that use the `get_targets` plan function, which we'd like to avoid because of the risk associated with parsing strings using regex.

I've opened a PR against this branch with suggested changes: https://github.com/jay7x/bolt/pull/1 . Let me know if you have any questions! Thanks again for suggesting this feature, I think users will love it! :)

jay7x:
Thank you for the PR! Now I know how to split such hard cases! :)

jay7x:
I updated the case JFYI. Pushed from wrong email so had to force-push after to fix 😞

beechtom:
@jay7x Ugh, made a silly mistake in that PR from earlier. It passed the `ext_glob` option to the rerun file instead of to the inventory. I made a second PR to your branch to pass it where needed in `Bolt::Application`.

Also, good idea to disable `File::FNM_EXTGLOB` when targets aren't passed from the CLI. I think in that case we can fall back to the old regex matching in your `match_wildcard?` method.

beechtom:
@jay7x Once `ext_glob` is added to the `target_aliases`, feel free to squash everything down to a single commit. Once tests are all passing, I'll get this merged. :)

beechtom:
@jay7x Looks like we've got some failing spec tests. Updating these two lines to include the `ext_glob: true` parameters should fix them:

https://github.com/puppetlabs/bolt/blob/main/spec/unit/application_spec.rb#L150
https://github.com/puppetlabs/bolt/blob/main/spec/unit/application_spec.rb#L155

Example:
```ruby
expect(inventory).to receive(:get_targets).with(targets, ext_glob: true).and_return([target])
```

Also, no worries about crediting. Just happy to help. :)

jay7x:
I updated the failing test as requested JFYI. It looks good for me locally

beechtom (approved):

Issue #3106, opened 2022-06-14
PuppetDB plugin doesn't honor specified instance when retrieving fact values


nicklewis:
## Describe the Bug

When using the `puppetdb` inventory plugin with multiple PuppetDB instances, the `query` for nodes is performed against the specified instance, but the follow-up query to retrieve referenced facts uses the default instance. That results in facts not actually being found.

PR #3107, opened 2022-06-14
(GH-3106) Retrieve fact values using named PDB instance

beechtom:
This fixes a bug where the `puppetdb` was correctly querying certnames
from a named PuppetDB instance, but was not collecting fact values from
the named PuppetDB instance. The `puppetdb` plugin now correctly
connects to the named instance when querying certnames and collecting
fact values.

!bug

* **Collect fact values from a named PuppetDB instance using `puppetdb`
  plugin**
  ([#3106](https://github.com/puppetlabs/bolt/issues/3106))

  The `puppetdb` plugin now correctly connects to a named PuppetDB
  instance when collecting fact values. Previously, the plugin only
  connected to the named PuppetDB instance when querying certnames.

### DIFF
--- a/lib/bolt/plugin/puppetdb.rb
+++ b/lib/bolt/plugin/puppetdb.rb
@@ -88,7 +88,7 @@ module Bolt
 
         facts.uniq!
         # Returns {'mycertname' => [{'path' => ['nested', 'fact'], 'value' => val'}], ... }
-        fact_values = @puppetdb_client.fact_values(targets, facts)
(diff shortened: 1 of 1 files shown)


donoghuc (approved):
Is there a similar bug when using `--query` on cli? https://github.com/puppetlabs/bolt/blob/a97c71b8cf870fa566ecd372d88f2b2936474f55/lib/bolt/cli.rb#L742

nicklewis (approved):

PR #3111, opened 2022-06-21
(BOLT-1527) Execute commands over winrm in separate shell instances

beechtom:
This updates the `winrm` transport to execute each command from Bolt in
a separate shell instance. Previously, the `winrm` transport opened a
single shell instance and ran all commands through it. This prevented
tasks and scripts that load DLLs from executing Bolt's cleanup command,
as DLLs cannot be unloaded when the shell instance that loaded them is
still running. Now, any task or script executed over `winrm` is executed
in a separate shell instance from the cleanup command, allowing DLLs to
be unloaded and removed from the temp directory when they are no longer
needed.

!bug

* **Execute commands over `winrm` in separate shell instances**

  Bolt now opens a new shell instance each time it runs a command over
  `winrm`. Previously, Bolt used a single shell instance to run all
  commands, preventing Bolt from properly cleaning up temporary files in
  some instances.

### DIFF
--- a/lib/bolt/transport/winrm/connection.rb
+++ b/lib/bolt/transport/winrm/connection.rb
@@ -53,9 +53,11 @@ module Bolt
             @connection = ::WinRM::Connection.new(options)
             @connection.logger = @transport_logger
 
-            @session = @connection.shell(:powershell)
-            @session.run('$PSVersionTable.PSVersion')
-            @logger.trace { "Opened session" }
+            @connection.shell(:powershell) do |session|
+              session.run('$PSVersionTable.PSVersion')
+            end
+
+            @logger.trace { "Opened connection" }
           end
         rescue Timeout::Error
           # If we're using the default port with SSL, a timeout probably means the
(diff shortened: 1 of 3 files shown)


lib/bolt/transport/winrm/connection.rb
```diff
@@ -116,12 +117,15 @@ def execute(command)
...
-            out_wr.close
-            err_wr.close
-            result.exitcode
+
+            @connection.shell(:powershell) do |session|
+              result = session.run(command)
+              out_wr << result.stdout
+              err_wr << result.stderr
+              out_wr.close
+              err_wr.close
```
beechtom, line 126: Just checking, but we should be able to remove these two lines due to the `ensure` below?

lib/bolt/transport/winrm/connection.rb
```diff
@@ -116,12 +117,13 @@ def execute(command)
...
             # they will be unhandled, so the default stack trace is unneeded.
             Thread.current.report_on_exception = false
-            result = @session.run(command)
-            out_wr << result.stdout
-            err_wr << result.stderr
-            out_wr.close
-            err_wr.close
-            result.exitcode
+
+            @connection.shell(:powershell) do |session|
```
nicklewis, line 121: Definitely a comment about why we do this will be useful in 3-5 years.

nicklewis (approved):

PR #3112, opened 2022-06-21
(maint) Properly configure bundler in bolt-server build

nicklewis:
Previously we were setting bundler config options at the user level in
the build container, which meant they didn't get copied over to the
runtime container. That caused gems not to be found at runtime. We now
set them in the local .bundle/config file so they get copied over to the
final build.

### DIFF
--- a/Dockerfile.bolt-server
+++ b/Dockerfile.bolt-server
@@ -2,15 +2,15 @@
 FROM alpine:3.14 as build
 
 RUN \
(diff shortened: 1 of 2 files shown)


nicklewis:
I believe the reason this was missed during local development is that a local `.bundle/config` will be copied into the container, so if you _already_ have `BUNDLE_PATH` set in that config, it just works.

donoghuc (approved):

Issue #3109, opened 2022-06-15
Bump puppetfile-resolver to 0.6.0


beechtom:
## Use Case
The `puppetfile-resolver` gem, which Bolt uses to resolve module dependencies, released [version 0.6.0](https://github.com/glennsarti/puppetfile-resolver/blob/main/CHANGELOG.md#060---2022-06-13). This version includes direct support for GitLab repos and searching for metadata files using `git clone`. Using this version of the gem in Bolt would allow users to install modules from a wider ranger of git repos.

PR #3115, opened 2022-06-23
(BOLT-1589) Support installing any git-based module

beechtom:
This updates Bolt to support installing git-based modules from any
source. Now, Bolt will attempt to download metadata directly from GitHub
or GitLab for public repositories and fall back to cloning the git
repo's metadata if that fails.

This also updates the spec resolver to no longer calculate a SHA based
on the ref provided in project configuration. Previously, Bolt would
calculate a SHA and write it to the Puppetfile to pin the module to a
specific commit when modules are installed. However, doing this is
difficult and time-consuming with the move to supporting any git-based
module.

!feature

* **Support installing any git-based module**
  ([#3109](https://github.com/puppetlabs/bolt/issues/3109))

  Bolt now supports installing any git-based module. Previously, Bolt
  only supported installing and resolving dependencies for modules
  hosted in a public GitHub repository. Now, Bolt will check both GitHub
  and GitLab before falling back to cloning a module's metadata using
  the `git` executable. This allows users to install modules from
  private repositories, or from locations other than GitHub.

---

This fixes a bug where the proxy configured in `module-install.proxy`
was not being set when retrieving metadata for git-based modules.
Previously, the proxy was only configured when resolving and
installing modules.

!bug

* **Configure proxy when retrieving metadata for git-based modules**

  Bolt now configures a proxy when `module-install.proxy` is set and it
  retrieves metadata for git-based modules. Previously, Bolt was only
  configuring a proxy when it resolved and installed modules.

### DIFF
--- a/lib/bolt/module_installer/specs/git_spec.rb
+++ b/lib/bolt/module_installer/specs/git_spec.rb
@@ -20,12 +20,13 @@ module Bolt
 
         attr_reader :git, :ref, :resolve, :type
 
-        def initialize(init_hash)
+        def initialize(init_hash, config = {})
           @logger  = Bolt::Logger.logger(self)
           @resolve = init_hash.key?('resolve') ? init_hash['resolve'] : true
           @git     = init_hash['git']
           @ref     = init_hash['ref']
           @name    = parse_name(init_hash['name'])
+          @proxy   = config.dig('proxy')
           @type    = :git
 
           unless @resolve == true || @resolve == false
@@ -108,7 +109,7 @@ module Bolt
 
           unless data
             raise Bolt::Error.new(
-              "Unable to locate metadata.json for module at #{@git}. This may not be a valid module. "\
+              "Unable to locate metadata.json for module at #{loc(@git)}. This may not be a valid module. "\
               "For more information about how Bolt attempted to locate the metadata file, check the "\
               "debugging logs.",
               'bolt/missing-module-metadata-error'
@@ -119,14 +120,14 @@ module Bolt
 
           unless data.is_a?(Hash)
             raise Bolt::Error.new(
-              "Invalid metadata.json at #{@git}. Expected a Hash, got a #{data.class}.",
+              "Invalid metadata.json at #{loc(@git)}. Expected a Hash, got a #{data.class}.",
               'bolt/invalid-module-metadata-error'
             )
           end
 
           unless data.key?('name')
             raise Bolt::Error.new(
-              "Invalid metadata.json at #{@git}. Metadata must include a 'name' key.",
+              "Invalid metadata.json at #{loc(@git)}. Metadata must include a 'name' key.",
               'bolt/missing-module-name-error'
             )
           end
@@ -134,7 +135,7 @@ module Bolt
           data
         rescue JSON::ParserError => e
           raise Bolt::Error.new(
-            "Unable to parse metadata.json for module at #{@git}: #{e.message}",
+            "Unable to parse metadata.json for module at #{loc(@git)}: #{e.message}",
             'bolt/metadata-parse-error'
           )
         end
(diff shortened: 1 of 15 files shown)


nicklewis:
Not calculating a SHA means the Puppetfile no longer refers to a specific version of the module, correct? In the case where the ref is a branch, that is.

lib/bolt/module_installer/specs/id/gitclone.rb
```diff
@@ -0,0 +1,99 @@
...
+          private_class_method def self.name_and_sha(git, ref, proxy)
+            require 'open3'
+
+            unless git?
+              Bolt::Logger.debug("'git' executable not found, unable to use git clone resolution.")
+              return nil
+            end
+
+            # Clone the repo into a temp directory that will be automatically cleaned up.
+            Dir.mktmpdir do |dir|
```
beechtom, line 25: This will cause problems for users who can't write to the system tmpdir. We might want to consider making this configurable either now or in the future. `puppetfile-resolver` supports config for doing this, so we'd be able to pass it through to that as well. We could name the config option `module-install.tmpdir` or something else.

nicklewis: I don't think that's actually a problem in this case, because `Dir.mktmpdir` will try a bunch of directories until it finds one that's writable, including `.` if it has to. We run into problems when we're trying to write to a tmp dir on a _remote_ system, because we only have one option to try there.

beechtom: Oh, perfect! 👍

donoghuc (approved):
Thanks for the "in person" walkthrough on this. Looks good, and we have some thoughts for improving this moving forward.

lib/bolt/module_installer/specs/id/base.rb
```diff
@@ -0,0 +1,105 @@
...
+          end
+
+          # Request the name and SHA for a module and ref.
+          #
+          # @param git [String] The URL to the git repo.
+          # @param ref [String] The ref to checkout.
+          # @param proxy [String] A proxy to use when making requests.
+          #
+          def self.request(git, ref, proxy)
+            name, sha = name_and_sha(git, ref, proxy)
```
donoghuc, line 31: Should the base class define a stub for this method to indicate that it is required for any sub class that uses the base?

lib/bolt/module_installer/specs/git_spec.rb
```diff
@@ -107,90 +96,50 @@ def to_resolver_module
...
+          id.sha
+        end
+
+        # Gets the ID for the module based on the specified ref and git URL.
+        # This is lazily resolved since Bolt does not always need this information,
+        # and requesting it is expensive.
+        #
+        private def id
+          @id ||= begin
+            module_id = Bolt::ModuleInstaller::Specs::ID::GitHub.request(@git, @ref, @proxy) ||
```
donoghuc, line 117: So it looks like the way we decide which `spec` class instance is to find the first non-`nil` request. I think this is ok, but a comment here and maybe some documentation in the spec implementations `request` method indicating that it is crucial that the request either return a new object or nil would be helpful down the line.

donoghuc (commented):
Couple of minor suggestions, but overall this looks good!

beechtom:
@donoghuc Updated with suggestions.

donoghuc (approved):
Looks great! How do we want to handle bumping the puppetfile resolver dep in the runtime? https://github.com/puppetlabs/puppet-runtime/blob/master/configs/components/rubygem-puppetfile-resolver.rb 

I think it may be simplest to bump it in the runtime and then once that promotes merge this.

beechtom:
@donoghuc Here's the PR to bump puppetfile-resolver (and other gem components): https://github.com/puppetlabs/puppet-runtime/pull/561

PR #3117, opened 2022-06-28
Stop submitting sensitive param values to plan_start command

nicklewis:
Previously, we sent all supplied parameter values to the plan_start
command, which meant Orchestrator stored and returned sensitive values.
We now use the plan metadata to remove sensitive values and send them in
a separate `sensitive` array. This allows the names to be stored and
returned without the values, so we can indicate *which* parameters were
used without exposing sensitive data.

When run against versions of PE that don't support the `sensitive`
array, it will just look like those params were never supplied.

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -76,6 +76,7 @@ RSpec.configure do |config|
   config.after :each do
     Bolt::Logger.stream    = nil
     Bolt::Logger.analytics = nil
+    YARD::Registry.clear if defined?(YARD::Registry)
   end
 
   # This will be default in future rspec, leave it on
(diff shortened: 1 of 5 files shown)


donoghuc (approved):
I like this idea. Looks like unconditionally attempting to load plan metadata breaks some test assumptions.

spec/spec_helper.rb
```diff
@@ -76,6 +76,7 @@
   config.after :each do
     Bolt::Logger.stream    = nil
     Bolt::Logger.analytics = nil
+    YARD::Registry.clear if defined?(YARD::Registry)
```
beechtom, line 79: Just curious why this was added?

beechtom: Cool cool. 👍

beechtom (approved):

donoghuc (approved):
I like that `run_plan` is better about returning a PlanResult now on error.

PR #3119, opened 2022-06-28
(maint) Pin net-ssh to < 7.0

beechtom:
This pins the max version of net-ssh to < 7.0. net-ssh >= 7.0
requires Ruby >= 2.6, which we cannot ship since Bolt needs to be
compatible with Ruby >= 2.5.

!no-release-note

### DIFF
--- a/bolt.gemspec
+++ b/bolt.gemspec
@@ -53,7 +53,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency "logging", "~> 2.2"
   spec.add_dependency "minitar", "~> 0.6"
   spec.add_dependency "net-scp", "~> 1.2"
-  spec.add_dependency "net-ssh", ">= 4.0"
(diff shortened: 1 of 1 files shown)


beechtom:
@donoghuc I'm not sure I agree. Most all of our gems are pinned to a version range, with the notable exceptions being puppet and net-ssh. Additionally, net-ssh 7.0 introduces some breaking changes (specifically, [removing a ton of algorithms](https://github.com/net-ssh/net-ssh#supported-algorithms)) that might adversely affect Bolt gem installs (see [here](https://github.com/puppetlabs/bolt/blob/main/lib/bolt/transport/ssh/connection.rb#L79-L84) and [here](https://github.com/puppetlabs/bolt/blob/main/lib/bolt/transport/ssh/connection.rb#L101-L103)).

donoghuc (approved):

PR #3120, opened 2022-06-28
(bug) Correctly print secret key paths when using rainbow format

beechtom:
This fixes a bug that caused bolt to error when running `bolt secret
createkeys --format rainbow`.

!bug

* **Correctly print secret key paths when using rainbow format**

  Bolt no longer errors when running `bolt secret createkeys --format
  rainbow` or `New-BoltSecretKey -Format rainbow`.

### DIFF
--- a/lib/bolt/outputter/human.rb
+++ b/lib/bolt/outputter/human.rb
@@ -2,6 +2,7 @@
 
 require_relative '../../bolt/container_result'
 require_relative '../../bolt/pal'
+require_relative '../../bolt/util/format'
 
 module Bolt
   class Outputter
@@ -890,7 +891,7 @@ module Bolt
       end
 
       def print_message(message)
-        @stream.puts(message)
+        @stream.puts(Bolt::Util::Format.stringify(message))
       end
 
       def print_error(message)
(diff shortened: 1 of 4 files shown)


rileynewton (approved):

beechtom:
@nicklewis That's definitely doable, though `print_message` is _intended_ to only print strings. This is the one location in Bolt where we currently pass a non-String value.

We could use `Bolt::Util::Format.stringify` to stringify everything first, though that would result in a slightly different output:

```shell
$ bolt secret createkeys --force --format rainbow
[
  "/Users/tom/.puppetlabs/bolt/keys/public_key.pkcs7.pem",
  "/Users/tom/.puppetlabs/bolt/keys/private_key.pkcs7.pem"
]
```

beechtom:
Updated to stringify messages before printing them.

donoghuc (approved):

nicklewis:
We could also just call `to_s` which is basically what the human outputter does (implicitly by calling `puts`)

beechtom:
@nicklewis True. I think I'd prefer to print JSON instead of stringified Ruby, though. It keeps things consistent across Bolt (for example, `out::message` converts to JSON).

nicklewis (approved):

Issue #3013, opened 2021-09-05
Support interpreters in lxd transport


dontlaugh:
## Use Case
An interpreters config is needed for parity with [the other container-based transports](https://puppet.com/docs/bolt/latest/bolt_transports_reference.html).

I believe having interpreters config will also allow using `bolt apply` on unsupported platforms like Arch Linux, e.g. the agent install can be skipped with `features` (and a manual install via `pacman`)

## Describe the Solution You Would Like
The same **interpreters** config options as the other transports.

## Additional Context
Here is an error message that led me to this feature request.

<details>
<summary>debug logs</summary>

```text
0 % bolt apply --execute "file { '/etc/puppetlabs': ensure => present }" --log-level=debug --targets nomad-server-02
Loaded project from '/home/coleman/Code/ops'
Started with 100 max thread(s)
Loading modules from /opt/puppetlabs/bolt/lib/ruby/gems/2.7.0/gems/bolt-3.17.0/bolt-modules:/home/coleman/Code/ops/modules:/home/coleman/Code/ops/.modules:/opt/puppetlabs/bolt/lib/ruby/gems/2.7.0/gems/bolt-3.17.0/modules
Loaded inventory from /home/coleman/Code/ops/inventory.yaml
Starting: plan 
Starting: install puppet and gather facts on nomad-server-02
Starting: install puppet and gather facts on nomad-server-02
Starting: task apply_helpers::custom_facts on nomad-server-02
Running task apply_helpers::custom_facts with '{"plugins":"Sensitive [value redacted]","_task":"apply_helpers::custom_facts"}' on ["nomad-server-02"]
Running task 'apply_helpers::custom_facts' on nomad-server-02
The LXD transport is experimental, and might include breaking changes between minor versions. [ID: lxd_experimental]
{"target":"nomad-server-02","action":"task","object":"apply_helpers::custom_facts","status":"failure","value":{"_output":"","_error":{"kind":"puppetlabs.tasks/task-error","issue_code":"TASK_ERROR","msg":"The task failed with exit code 126 and no stdout, but stderr contained:\nsh: /tmp/aa749c48-b472-4ee5-89b3-0231e92b3676/custom_facts.rb: /opt/puppetlabs/puppet/bin/ruby: bad interpreter: No such file or directory\n\n\nThis might be caused by the default tmpdir being mounted using 'noexec'. See http://pup.pt/task-failure for details and workarounds.","details":{"exit_code":126}}}}
```

</details>

[...]

PR #3123, opened 2022-06-29
Add interpreters config option to LXD transport

jonasdemoor:
This allows setting a custom Ruby path, which may come in handy when
using Puppet and Ruby packages provided by the distro itself (e.g. Arch
Linux) instead of the official Puppetlabs packages.

This change, together with setting `features: ['puppet-agent']`, allows
`bolt apply` to run against LXD containers with non-default Puppet and
Ruby paths.

!feature

* **Support interpreters in LXD transport configuration** ([#3013](https://github.com/puppetlabs/bolt/issues/3013))

Bolt now also supports setting the interpreters configuration option on LXD
transports.

### DIFF
--- a/lib/bolt/config/transport/lxd.rb
+++ b/lib/bolt/config/transport/lxd.rb
@@ -9,6 +9,7 @@ module Bolt
       class LXD < Base
         OPTIONS = %w[
           cleanup
+          interpreters
           remote
           tmpdir
         ].concat(RUN_AS_OPTIONS).sort.freeze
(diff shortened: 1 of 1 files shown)


nicklewis:
We already support this setting for the docker transport so this makes perfect sense. Thanks for filing this!

beechtom (approved):

jonasdemoor:
I tested this against a remote LXD server running in a local libvirt VM. Both the VM and test container are running the latest Arch Linux updates.

Bolt version (installed via `bundle install` from this PR's branch):
```
$ bolt --version
3.23.1
```

LXD server version:
```
$ sudo lxc version
Server version: 5.3
```

LXD client version:
```
$ lxc version
Client version: 5.1
```

I'd be happy to provide more details if needed :)

PR #3126, opened 2022-06-30
(maint) Do not fail fast for unit tests

beechtom:
This removes the `--fail-fast` flag when running RSpec unit tests.
Previously, the unit test suite would fail as soon as a single test case
failed, which prevented us from seeing all tests that would have failed
in a run. Removing this flag will now let RSpec run all tests and show
all failures in the test suite, allowing for faster development.

!no-release-note


nicklewis:
Can you add to the commit message _why_ we're making this change? Since it was unclear why it was the way it was before

beechtom:
@nicklewis Updated with a commit message.

Issue #2818, opened 2021-05-06
Target group globbing


breca:
## Use Case
I would like to be able to glob target groups

## Describe the Solution You Would Like
`bolt inventory show --targets group:'foo*'`
`bolt command run 'echo "Bolt is great"' -t group:'*bar*'`

## Describe Alternatives You've Considered
Creating groups of groups, which is nice on paper but can become a bit painful when dealing with dynamic inventories at large scales

reidmv:
Suggestion:

No need to include a `group:` prefix. The Bolt inventory already knows group names, in addition to target names, and treats them identically for almost all other interfaces except for globbing. So, globbing shouldn't be special. Make globbing work for and find group names in addition to target names. It should work exactly as it already does for targets, just match group names as well.

PR #3137, opened 2022-07-27
(GH-2818) Support globbing group names

beechtom:
This updates the inventory to support globbing group names in addition
to target names and aliases.

!feature

* **Support globbing group names**
  ([#2818](https://github.com/puppetlabs/bolt/issues/2818))

  Bolt now supports selecting group names with glob patterns when
  selecting targets.

### DIFF
--- a/lib/bolt/inventory/inventory.rb
+++ b/lib/bolt/inventory/inventory.rb
@@ -125,16 +125,25 @@ module Bolt
       end
 
       # If target is a group name, expand it to the members of that group.
-      # Else match against targets in inventory by name or alias.
+      # Else match against groups and targets in inventory by name or alias.
       # If a wildcard string, error if no matches are found.
       # Else fall back to [target] if no matches are found.
       def resolve_name(target, ext_glob: false)
         if (group = group_lookup[target])
-          group.all_targets
+          group.all_targets.to_a
         else
-          # Try to wildcard match targets in inventory
-          # Ignore case because hostnames are generally case-insensitive
-          targets = groups.all_targets.select { |targ| match_wildcard?(target, targ, ext_glob: ext_glob) }
+          targets = []
+
+          # Find groups that match the glob
+          group_lookup.each do |name, grp|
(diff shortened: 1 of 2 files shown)


lib/bolt/inventory/inventory.rb
```diff
@@ -125,16 +125,25 @@ def match_wildcard?(wildcard, target_name, ext_glob: false)
...
         else
-          # Try to wildcard match targets in inventory
-          # Ignore case because hostnames are generally case-insensitive
-          targets = groups.all_targets.select { |targ| match_wildcard?(target, targ, ext_glob: ext_glob) }
+          targets = []
+
+          # Find groups that match the glob
+          group_lookup.each do |name, grp|
+            next unless match_wildcard?(target, name, ext_glob: ext_glob)
+            targets += grp.all_targets.to_a
```
justinstoller, line 140: It looks like this behavior existed prior to this PR. But am I reading this correctly in that: `Group#all_targets` returns a `Set` and this method (`resolve_name`) will return a `Set` of targets if the `target` parameter exactly matches a group name, or an array of targets otherwise?

justinstoller (approved):

lib/bolt/inventory/inventory.rb
beechtom: Ya, that's correct. I'm not really sure why it was written like that, but since it's only called in one place and immediately mapped it probably doesn't matter too much. I can update it to always return either an `Array` or `Set` if we want, though.

justinstoller: I assume that's unintentional but I understand not wanting to change a method that may have public users so I'm fine if you are merging as-is, since it honors the existing weirdness. Though if we want to keep it might b nice to add a comment regarding return values at the top of the method.

beechtom: Updated to just have it always return an `Array`. Easy enough fix. :) Thanks for pointing out the weirdness!

PR #3141, opened 2022-07-29
(hack) Add `bolt script show` and `Get-BoltScript` commands

beechtom:
This adds new `bolt script show` and `Get-BoltScript` commands.
When used without a script specified, they list all scripts available to
the project. When a script is specified, they print the script to the
console.

These commands load scripts from the `scripts/` and `files/scripts/`
directories in a project and modules on the modulepath, similar to the
`bolt script run` and `Invoke-BoltScript` commands. It does not load
scripts that are in the `files/` directory, as Bolt does not know
whether a file in that directory is a script or a file used elsewhere in
a module.

!feature

* **Add `bolt script show` and `Get-BoltScript` commands**

  You can now view scripts available to a project using the new `bolt
  script show` and `Get-BoltScript` commands. These commands show
  scripts in the `scripts/` and `files/scripts/` directories for
  projects and modules.

---

This adds a new `scripts` project configuration option that allows users
to create an allowlist of scripts that are available to the project.
This option supports globbing. Any scripts that are not present in the
allowlist will not appear in `bolt script show` or `Get-BoltScript`
output, though they can still be run by users directly.

!feature

* **Add `scripts` project configuration option**

  This new configuration option allows users to create an allowlist of
  scripts that are available to the project. Glob patterns are
  supported. Any scripts not in the allowlist are hidden from `bolt
  script show` and `Get-BoltScript` output, but can still be run
  directly by users.

### DIFF
--- a/lib/bolt/pal.rb
+++ b/lib/bolt/pal.rb
@@ -478,14 +478,16 @@ module Bolt
 
     # List the scripts available to a project. This includes any scripts in a
     # project's or module's 'files/scripts/' and 'scripts/' directories.
-    def list_scripts
+    def list_scripts(filter_content: false)
       in_bolt_compiler do
-        Puppet.lookup(:current_environment).modules.collect do |mod|
+        scripts = Puppet.lookup(:current_environment).modules.collect do |mod|
           files_dir   = Pathname.new(mod.files) + 'scripts'
           scripts_dir = Pathname.new(mod.scripts)
-          scripts     = Bolt::Util.collect_files(files_dir) + Bolt::Util.collect_files(scripts_dir)
-          scripts.map { |path| File.join(mod.name, path.relative_path_from(mod.path)) }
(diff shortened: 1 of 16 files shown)


lib/bolt/pal.rb
```diff
@@ -476,6 +476,29 @@ def list_plans(filter_content: false)
...
     end
 
+    def list_scripts
+      in_bolt_compiler do
+        scripts = []
+
+        Puppet.lookup(:current_environment).modules.each do |mod|
+          if mod.files?
+            if File.directory?(file_scripts = File.join(mod.files, 'scripts'))
+              Dir.children(file_scripts).each do |script|
```
donoghuc, line 486: Does script discovery need to descend into nested directories (and *not* list directory names themselves?).

beechtom: Ah, good catch, it does.

beechtom: Updated to recursively search for scripts. 👍

PR #3143, opened 2022-08-09
Work around PuppetDB bug involving large array queries

nicklewis:
PuppetDB has degenerate performance when parsing a query that contains a 
long array. When querying facts for a list of certnames, we generate a 
query that enumerates all the certnames to consider. This triggers the 
degenerate performance when querying for a large number (on the order of 
tens of thousands) of nodes. We now follow a common pattern of breaking 
this single array into many nested arrays, capping the length of any 
individual array at 100 elements.

We also now use an `in` query instead of an `or` query, which doesn't have
a performance impact but is cleaner to generate.

This also fixes a bug where we would always query PuppetDB for facts even if we
were only referencing "certname", which is not a fact, in the target_mapping
section.

### DIFF
--- a/lib/bolt/plugin/puppetdb.rb
+++ b/lib/bolt/plugin/puppetdb.rb
@@ -79,8 +79,9 @@ module Bolt
 
         Bolt::Util.walk_vals(template) do |value|
           # This is done in parts instead of in place so that we only need to
-          # make one puppetDB query
-          if value.is_a?(String)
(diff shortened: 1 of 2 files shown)


nicklewis:
For future reference, the PuppetDB bug in question here is [PDB-3171](https://tickets.puppetlabs.com/browse/PDB-3171).

donoghuc (approved):

PR #3156, opened 2022-09-27
(PE-34483) Ensure plan execution is halted when sleep is interrupted

donoghuc:
The ruby sleep function will handle an interrup signal and allow plan execution to continue when run in PE (using jruby interpreter). This commit updates the `ctrl` functions (which use sleep) to handle an interrupt signal by raising an error when interrupted so that plan execution is stopped.

### DIFF
--- a/bolt-modules/ctrl/lib/puppet/functions/ctrl/sleep.rb
+++ b/bolt-modules/ctrl/lib/puppet/functions/ctrl/sleep.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require_relative '../../../shared/sleep_signal'
+
 # Sleeps for specified number of seconds.
 Puppet::Functions.create_function(:'ctrl::sleep') do
   # @param period Time to sleep (in seconds)
@@ -13,8 +15,7 @@ Puppet::Functions.create_function(:'ctrl::sleep') do
   def sleeper(period)
     # Send Analytics Report
     Puppet.lookup(:bolt_executor) {}&.report_function_call(self.class.name)
-
-    sleep(period)
+    SleepSignal.sleep_with_signal_handler(period)
     nil
   end
 end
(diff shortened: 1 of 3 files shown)


donoghuc:
Open questions:
1. Should this land in bolt or pe-bolt-modules? (I chose bolt to avoid duplication and to keep consistency). 
2. What spec tests do we need for this? I was thinking maybe showing the handler is restored, but its probably not worth it...

bolt-modules/ctrl/lib/puppet/functions/ctrl/sleep.rb
```diff
@@ -1,5 +1,15 @@
 # frozen_string_literal: true
 
+def sleep_with_signal_handler(period)
+  # When running with a jruby interpreter sleep will handle an iterrupt and allow the plan to keep
+  # running. In order to have consistent behavior between plans in bolt and PE and to ensure plans
+  # are able to be reliably stopped we add a signal handler to raise an exception on interrupt.
+  handler = Signal.trap(:INT) { raise Puppet::Error, "interrupt signal received during sleep" }
```
justinstoller, line 7: Shouldn't the handler just be the function? like
```rb
handler = ->() { raise Puppet::Error, ".." }
```
?

bolt-modules/ctrl/lib/puppet/functions/ctrl/sleep.rb
```diff
@@ -1,5 +1,15 @@
 # frozen_string_literal: true
 
+def sleep_with_signal_handler(period)
```
justinstoller, line 3: It feels like this should go into a utility class and be required by both functions, like PuppetX utils in types & providers

donoghuc:
Thanks @justinstoller unfortunately this is not going to work as I thought. Cancelling the future on our job tracker does not result in a signal we can trap at this level. Going to have to try something else.

PR #3158, opened 2022-10-13
Helpful task fail message for CRLF line endings issue

zoojar:


lib/bolt/shell/bash.rb
```diff
@@ -480,7 +480,8 @@ def execute(command, sudoable: false, **options)
           @logger.trace { "Command `#{command_str}` returned successfully" }
         when 126
           msg = "\n\nThis might be caused by the default tmpdir being mounted "\
-            "using 'noexec'. See http://pup.pt/task-failure for details and workarounds."
+            "using 'noexec', or the task file has CRLF (not LF) line endings."\
```
nicklewis, line 483:
```suggestion
            "using 'noexec', or the task file has CRLF (not LF) line endings. "\
```

donoghuc:
Thanks @zoojar ! Would you mind squshing commits and adding a `!no-release-note` commit message (see [contributing guidelines](https://github.com/puppetlabs/bolt/blob/main/CONTRIBUTING.md#pull-requests) )

PR #3164, opened 2022-11-10
(maint) Make hostname_nix task more specific

donoghuc:
Some new platforms (namely fedora 36) will resolve `$HOSTNAME` as a fqdn. Instead of relying on this environment variable use the `hostname -s` command to compute the hostname. This will work on all of the unix systems we test ssh against in our acceptance test matrix.

### DIFF
--- a/acceptance/tests/plan_ssh.rb
+++ b/acceptance/tests/plan_ssh.rb
@@ -18,7 +18,7 @@ test_name "C100553: \
     on(bolt, "mkdir -p #{dir}/modules/test/{tasks,plans}")
     create_remote_file(bolt, "#{dir}/modules/test/tasks/hostname_nix.sh", <<-FILE)
 #!/bin/bash
-if [ ! -f /tmp/retry.txt ] && [ "$HOSTNAME" = '#{first_node}' ]; then
+if [ ! -f /tmp/retry.txt ] && [ "$(hostname -s)" = '#{first_node}' ]; then
   touch /tmp/retry.txt;
   exit 1
 elif [ -f /tmp/retry.txt ]; then


acceptance/tests/plan_ssh.rb
```diff
@@ -18,7 +18,7 @@
     on(bolt, "mkdir -p #{dir}/modules/test/{tasks,plans}")
     create_remote_file(bolt, "#{dir}/modules/test/tasks/hostname_nix.sh", <<-FILE)
 #!/bin/bash
-if [ ! -f /tmp/retry.txt ] && [ "$HOSTNAME" = '#{first_node}' ]; then
+if [ ! -f /tmp/retry.txt ] && [[ "$HOSTNAME" =~ '#{first_node}' ]]; then
```
justinstoller, line 21: I don't know how anal this is, but are we guaranteed to have first name only be something that could match the first portion of the hostname? ie, `=~` treats the matcher as being `/.*#{first_node}.*/`, so if `first_node` is `lab` it will match `lab01.google.com` or `foo.puppetlabs.net`.

If that's _not_ too nickpicky then either `[[ "$(hostname -s)" = '#{first_node}" ]]` (on linux systems) or `[[ "$(echo $HOSTNAME | cut -d. -f1)" = '#{first_name}' ]]` (all unixy systems including solaris and is the bash equivalent of the Ruby transform we do to `first_node`).

donoghuc: you do have a valid point, but i think its pretty unlikely given how we assign hostnames ({random word}-{random word}). I updated to just use `hostname -f` and an exact match to help prevent against any issues of a non-specific regex. I looked through our test matrix and I dont see any platforms where we would need the second option. Thanks for the review!

justinstoller (approved):

PR #3166, opened 2022-11-12
Allow latest version of puppet-strings

smortex:
[puppet-strings 3.0.0] was released on 2022-10-21.  It bumps some
minimal requirements which seems to be in-line with those of Bolt so
allow this version too.

[puppet-strings 3.0.0]:https://github.com/puppetlabs/puppet-strings/blob/main/CHANGELOG.md#v300---2022-10-21


donoghuc (approved):

donoghuc:
Thanks @smortex will be working on getting a release out soon.

PR #3168, opened 2022-11-16
(maint) Update acceptance config of OSX/Win for puppet-strings 2.9

jpartlow:
Following up on 806852ae, two of our internal acceptance test platforms, osx12 and windows19 have older Rubies and Rubygems (2.6, 2.5) that are neither compatible with puppet-strings 3, nor capable of recognizing that they need to downshift to 2.x. So this commit updates the presuite to force installation of puppet-strings 2.9.0 ahead of installing the test bolt gem on those platforms.


jpartlow:
This cleared up our two acceptance failures installing the bolt gem in my local manual testing.

donoghuc (approved):

PR #3182, opened 2023-01-23
(PUP-11722) Restrict concurrent ruby to 1.1.x

donoghuc:
The 1.2.x series of concurrent ruby is incompatable with existing puppet 7.x. This commit restricts concurrent ruby while we figure out how to take up 1.2.


donoghuc:
Similar patch in puppet https://github.com/puppetlabs/puppet/pull/8996

justinstoller (approved):

PR #3183, opened 2023-01-24
(PE-34846) Remove POST /project_inventory_targets endpoint

nicklewis:
This endpoint was intended for support for a non-existent feature in PE,
leaving it now unused and unnecessary.

This commit leaves the other project-related methods because they are
generic and useful for any project-based functionality in the future.
The /project_inventory_targets endpoint on the other hand was much more
purpose-built and is unlikely to be helpful in the future.

!no-release-note


donoghuc (approved):
🔥

PR #3204, opened 2023-04-27
gemsp[ec: bump dependency on puppet-strings

b4ldr:
This is required so that projects that depend on bolt can make use of the more recent puppet-strings release.

I ran the doc:all rake task and there where no changes

### DIFF
--- a/bolt.gemspec
+++ b/bolt.gemspec
@@ -59,7 +59,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency "puppet", ">= 6.18.0"
   spec.add_dependency "puppetfile-resolver", ">= 0.6.2", "< 1.0"
   spec.add_dependency "puppet-resource_api", ">= 1.8.1"
-  spec.add_dependency "puppet-strings", ">= 2.3.0", "< 4.0"
+  spec.add_dependency "puppet-strings", ">= 2.3.0", "< 5.0"
   spec.add_dependency "r10k", "~> 3.10"
   spec.add_dependency "ruby_smb", "~> 1.0"
   spec.add_dependency "terminal-table", "~> 3.0"


donoghuc:
Hi @b4ldr is there a particular feature in puppet-strings you are looking to take advantage of?

b4ldr:
Yes the @deprecated flag

donoghuc:
Thanks! Looking in to a bunch of gem bumps as we work on getting bolt ready to be compatible with puppet 8. I will make sure this gets looked at as part of that.

b4ldr:
any update on this, its preventing any project that depends on bolt from making use of the recent improvements in puppet-strings

donoghuc:
Are you managing bolt's gem dependencies in your workflow or are you depending on the gems we ship with bolt system packages? Trying to get a sense of whether a gem release with this relaxed dependency will unblock you or if you need us to ship the new puppet-strings in bolt packages?

b4ldr:
> Are you managing bolt's gem dependencies in your workflow or are you depending on the gems we ship with bolt system packages?

My main motivation is to start using the deprecated tag in  the `puppetlabs-stdlib` repo but the same flow applies for any puppet module that use bolt or litmnus for testing.  stdlib is bringing this dependency in via [puppet_litmus](https://github.com/puppetlabs/puppetlabs-stdlib/blob/main/Gemfile#L40) from its Gemfile

We have had to do similar patches in the [past](https://github.com/puppetlabs/bolt/pull/3166)

donoghuc:
Cool. This should work for your use case. We merged this change in https://github.com/puppetlabs/bolt/pull/3216 . I would expect this to eventually make it in to bolt system packages, but for now if you are using bolt as a gem, we can get this released and that should help.

b4ldr:
great thanks <3

PR #3206, opened 2023-05-02
(maint) Pin sematic-puppet for component acceptance

donoghuc:
The bolt runners in the component acceptance tests have ruby interpreters < 2.7. There is a new sematic puppet release that requires ruby 2.7 and the gem install commmand is not smart enough to pull in dependencies based on ruby version requirements. This commit updates the test runners to pre-install a suitable sematic puppet version before installing the bolt gem.


donoghuc:
MEant to address https://jenkins-platform.delivery.puppetlabs.net/view/bolt/job/platform_bolt_vanagon-integration-component_main_project-main/932/

jpartlow (approved):
:+1:

PR #3213, opened 2023-06-02
(PE-35920) Error when using interpolations for hiera config in plan scope

donoghuc:
Set `:avoid_hiera_interpolation_errors` to `false` so puppet will error when intepolations are attempted in plan scope.


donoghuc:
It turns out this was built on a bug in puppet 7. When using strict mode failing to find values for interpolations should not actually raise an error, instead the appropriate default should be used for lookups. Closing this, we will update the tests here when we move to taking up puppet 8.

PR #3216, opened 2023-06-20
(MAINT) Bump puppet-strings dependency

chelnak:
Prior to this change, the latest major version of puppet-strings was not being pulled in to bolt.

This change relaxes the requirement to allow versions less than v5.


donoghuc (approved):

GSPatton:
could we get release of bolt to reflect this change?

PR #3218, opened 2023-07-14
(PE-36345) Relax CFPropertyList version

mcdonaldseanp:
In preperation for new ruby versions in PE relax the CFPropertyList requirement so later versions can install

### DIFF
--- a/bolt.gemspec
+++ b/bolt.gemspec
@@ -45,7 +45,7 @@ Gem::Specification.new do |spec|
 
   spec.add_dependency "addressable", '~> 2.5'
   spec.add_dependency "aws-sdk-ec2", '~> 1'
-  spec.add_dependency "CFPropertyList", "~> 2.2"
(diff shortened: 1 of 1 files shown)


bolt.gemspec
```diff
@@ -45,7 +45,7 @@ Gem::Specification.new do |spec|
 
   spec.add_dependency "addressable", '~> 2.5'
   spec.add_dependency "aws-sdk-ec2", '~> 1'
-  spec.add_dependency "CFPropertyList", "~> 2.2"
+  spec.add_dependency "CFPropertyList", ">= 2.2"
```
justinstoller, line 48: For posterity it looks like 3.0.6 is necessary for Ruby 3.2 support and the big breaking change between 2.x and 3.x was dropping support for Ruby 1.8 (and we complain about supporting 2.76!).

justinstoller (approved):

donoghuc (approved):

Issue #3217, opened 2023-07-11
Bolt logs sensitive data by default in certain contexts


Ziaunys:
## Describe the Bug
With the default logging configuration, Bolt logs sensitive parameters of a class that is applied during a Puppet apply block when running a plan.

## Expected Behavior
I don't think sensitive data should be logged period, but I would settle for not logging any by default. Additionally, I would not overload the log level to indicate when sensitive data is logged because it seems like an unexpected side effect. Perhaps we there could be a separate setting to determine if sensitive data should be logged.

## Steps to Reproduce
1. Create a class that looks like this:
```pp
class my_module::sensitive_test (Sensitive[String[1]] $password, Hash[String[1], Sensitive[String[1]]] $secret_hash) {
  notify { 'This class has a sensitive parameter': }
}
```
2. Declare this class in a Bolt plan:
```pp
plan my_module::sensitive_param_test() {
  apply('test_target', '_run_as' => 'root') {
    class { 'my_module::sensitive_test':
      password    => Sensitive('this_is_a_secret'),
      secret_hash => { 'secret_key' => Sensitive('secret_value') },
    }
  }
}
```
3. Run the bolt plan.
4. Check the bolt-debug.log to see the sensitive params logged. You should see a log line like the following
```
2023-07-11T16:03:57.367699 INFO   [apply-worker-1] [Bolt::Executor] {"tags":["settings","my_module::sensitive_test","my_module","sensitive_test","class"],"name":"zounese-dev-madding-larvae","version":1689116637,"code_id":null,"catalog_uuid":"1b8edb44-e6fa-4a88-bb8c-75cb4a75b9dc","catalog_format":2,"environment":"bolt_catalog",
"resources":[{"type":"Stage","title":"main","tags":["stage"],"exported":false,"kind":"compilable_type","parameters":{"name":"main"}},{"type":"Class","title":"Settings","tags":["class","settings"],"exported":false,"kind":"unknown"},{"type":"Class","title":"main","tags":["class"],"exported":false,"kind":"unknown","parameters":{"name":"main"}},
[...]

PR #3222, opened 2023-08-10
(GH-3217) Only log compiled catalogs at trace level

donoghuc:
Previously compiled catalogs would be logged per target at INFO level. This clogs up logs and can expose Sensitive data in catalogs. This commit moves logging of compiled catalogs to the TRACE log level only.

!bug

* **Only log compiled catalogs at trace level** ([#3217](#3217))

  Previously compiled catalogs would be logged per target at INFO level. This clogs up logs and can expose Sensitive data in catalogs. This commit moves logging of compiled catalogs to the TRACE log level only.

### DIFF
--- a/lib/bolt/applicator.rb
+++ b/lib/bolt/applicator.rb
@@ -256,7 +256,7 @@ module Bolt
         futures = targets.map do |target|
           Concurrent::Future.execute(executor: @pool) do
             Thread.current[:name] ||= Thread.current.name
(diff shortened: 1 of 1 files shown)


jpartlow (approved):
:+1: Works in my local testing. Compiled catalog is no longer logged by default, so the secrets no longer show; but you can get the catalog info back with --log-level trace.

PR #3242, opened 2023-10-05
(maint) stop restricting bundler and ruby in Gemfile

donoghuc:
Bolt's gemspec allows ruby 2.5 and up. This is the source of truth for ruby compatability. This commit removes the versions in the Gemfile. The restriction was added to try to get rid of a warning message when using a bundler version that is too old. This is something bundler users can manage themselves if the warning is annoying them.

!bug

* **Stop restricting ruby version in Gemfile**

  The source of truth for minimum ruby version is in the gemspec. Stop duplicating that and risking conflicts in the Gemfile.


mcdonaldseanp (approved):

tlehman (commented):
LGTM

Issue #1987, opened 2020-07-13
Unable to authenticate with ed25519 keys over SSH transport on Windows


beechtom:
## Describe the Bug
Attempting to connect to a target with ed25519 keys using the SSH transport on a
Windows host causes Bolt to raise an exception that net-ssh requires gem
dependencies for ed25519 support.

## Expected Behavior
Bolt should be able to use ed25519 keys using the SSH transport on a Windows host.

## Environment
 - **Platform:** Windows

## Additional Context
By default, Bolt uses the `net-ssh` Ruby libary to connect to targets over
SSH. The `net-ssh` library requires the `ed25519` and `bcrypt_pbkdf` gems as
dependencies, which are not supported in Bolt's packaging process due to
issues with compiling native extensions.

Attempting to authenticate with ed25519 keys over SSH on Windows will result
in an error message similar to this:

```
unsupported key type `ssh-ed25519'
 net-ssh requires the following gems for ed25519 support:
  * ed25519 (>= 1.2, < 2.0)
  * bcrypt_pbkdf (>= 1.0, < 2.0)
```

## Workaround
A workaround is to use native SSH when you need to authenticate with
ed25519 keys. When native SSH is enabled, Bolt will use a specified SSH
client to connect to targets instead of the `net-ssh` Ruby library. To
learn more about native SSH, see [native SSH
transport](https://puppet.com/docs/bolt/latest/experimental_features.html#native-ssh-transport).

PR #3248, opened 2023-11-22
(maint) add ed25519 support

h0tw1r3:
rubygems has compiled versions of bcrypt_pbkdf 1.1.0 for windows

addresses #1987

- [x] https://github.com/puppetlabs/puppet-runtime/pull/844
- [x] https://github.com/net-ssh/bcrypt_pbkdf-ruby/pull/22


donoghuc:
@h0tw1r3 did you build a bolt package with some combo of these gems on windows? If so, could you update puppet-runtime with that corresponding change? If this builds we can certainly take this up.

h0tw1r3:
I did built it, will add it to runtime and send it through the standard pipeline.

donoghuc:
Thanks, yeah we will need that to be merged before we require these gems in the gemspec :)

PR #3249, opened 2023-11-27
r10k: Require 4.x & octokit: Allow newer versions & puppetlabs_spec_helper: Require 7.x

bastelfreak:
This is a requirement to get newer dependencies for modules testing. pdk configures puppet_litmus, which in turn depends on bolt, which depends on outdated other gems.

### DIFF
--- a/bolt.gemspec
+++ b/bolt.gemspec
@@ -60,7 +60,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency "puppetfile-resolver", ">= 0.6.2", "< 1.0"
   spec.add_dependency "puppet-resource_api", ">= 1.8.1"
   spec.add_dependency "puppet-strings", ">= 2.3.0", "< 5.0"
-  spec.add_dependency "r10k", "~> 3.10"
+  spec.add_dependency "r10k", "~> 4.0"
   spec.add_dependency "ruby_smb", "~> 1.0"
   spec.add_dependency "terminal-table", "~> 3.0"
   spec.add_dependency "winrm", "~> 2.0"
--- a/bolt.gemspec
+++ b/bolt.gemspec
@@ -68,7 +68,7 @@ Gem::Specification.new do |spec|
 
   spec.add_development_dependency "bundler", ">= 1.14"
   spec.add_development_dependency "octokit", "~> 4.0", '< 9'
-  spec.add_development_dependency "puppetlabs_spec_helper", "~> 5.0"
+  spec.add_development_dependency "puppetlabs_spec_helper", "~> 7.0", '>= 7.0.1'
   spec.add_development_dependency "rake", "~> 12.0"
   spec.add_development_dependency "rspec", "~> 3.0"
 end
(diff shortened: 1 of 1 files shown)


bolt.gemspec
```diff
@@ -60,15 +60,15 @@ Gem::Specification.new do |spec|
   spec.add_dependency "puppetfile-resolver", ">= 0.6.2", "< 1.0"
   spec.add_dependency "puppet-resource_api", ">= 1.8.1"
   spec.add_dependency "puppet-strings", ">= 2.3.0", "< 5.0"
-  spec.add_dependency "r10k", "~> 3.10"
+  spec.add_dependency "r10k", "~> 4.0"
```
donoghuc, line 63: We cant drop support for 3. We can update to support both 3 and 4 if that is helpful.

bolt.gemspec
```diff
@@ -60,15 +60,15 @@ Gem::Specification.new do |spec|
...
   spec.add_dependency "ruby_smb", "~> 1.0"
   spec.add_dependency "terminal-table", "~> 3.0"
   spec.add_dependency "winrm", "~> 2.0"
   spec.add_dependency "winrm-fs", "~> 1.3"
 
   spec.add_development_dependency "bundler", ">= 1.14"
-  spec.add_development_dependency "octokit", "~> 4.0"
-  spec.add_development_dependency "puppetlabs_spec_helper", "~> 5.0"
+  spec.add_development_dependency "octokit", "~> 4.0", '< 9'
```
donoghuc, line 70: Wouldnt `~> 4.0` imply `>= 4.0 < 5` ?

bolt.gemspec
bastelfreak: Hi,
I took a bit of a break in responding here because the whole Ruby situation in the Puppet/Perforce tooling is a bit frustrating.

In theory we could do `spec.add_dependency "r10k", ">= 3.10", "< 5"`. But bundler showed in the past that it had some trouble to figure out correct versions / it didn't always hounor the minimal ruby version in the dependencies. I don't know if this keeps working on ruby 2.5 / 2.6 if we do the change.

bolt.gemspec
bastelfreak: indeed, it was a typo, should have been `">= 4.0", '< 9'`

bolt.gemspec
donoghuc: `spec.add_dependency "r10k", ">= 3.10", "< 5"` seems reasonable to me

bastelfreak: But in order to actuallly use the new r10k version, we need a release of https://github.com/puppetlabs/orchestrator_client-ruby/pull/38 .

donoghuc:
I think we can drop 2.5 for bolt 4.

bolt.gemspec
donoghuc: Yep.

bastelfreak:
For bolt 4? Why can't we drop it now? Or do you mean for Bolt 3?

donoghuc:
The only reason we've kept 2.5 is for some internal testing on older OS. I was thinking of updating all that when we move to making the default ruby 3 based (keeping ruby 2.7 compatability for 2021.7 PE stream). What is the min version being 2.5 preventing for you now?

PR #3255, opened 2023-12-14
support using the target agent puppet config

h0tw1r3:
modules like puppetlabs-puppetdb which make use of Puppet.runtime[:http] require the agent ssl certificate and key. this change supports an optional flag which simply skips the confdir setting override when initializing the Puppet object.


donoghuc:
Apply blocks intentionally isolate from existing agent settings. A generalized option like this I think will have many unintended consequences. Perhaps we can brainstorm some other way to pipe through specific settings for a use case like this.

PR #3262, opened 2024-01-29
add shell-command and tty support to the LXD transport

h0tw1r3:
!feature

* **add shell-command and tty support to the LXD transport** ([#3262](https://github.com/puppetlabs/bolt/pull/3262))

Previously the LXD transport would always execute commands on the target with `sh -c`. The shell-command and tty options are added to provide a more consistent experience between the docker, podman and lxd transports.

### DIFF
--- a/lib/bolt/config/transport/lxd.rb
+++ b/lib/bolt/config/transport/lxd.rb
@@ -11,7 +11,9 @@ module Bolt
           cleanup
           interpreters
           remote
+          shell-command
           tmpdir
+          tty
         ].concat(RUN_AS_OPTIONS).sort.freeze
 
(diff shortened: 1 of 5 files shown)


h0tw1r3:
End goal is the implement litmus testing with the lxd (or incus) transport.

This is a blocker because the static shell command `sh -c` will not allow the `litmus:install_agent` rake task to complete. Docker and podman work with litmus because it's able to set shell-command to `sh -lc`.

h0tw1r3:
reminder to self, there are no tests for lxd as it's considered experimental.

donoghuc (approved):

PR #3271, opened 2024-02-18
(maint) trace docker transport run_cmd

h0tw1r3:
while tracing uploading files with the docker transport, I noticed many of the docker commands being run were not trace logged.


donoghuc:
Are you saying that the trace logging with upload_file https://github.com/puppetlabs/bolt/blob/ca25a6bed9b1ea70e8b03d88597c7ad935140dfe/lib/bolt/transport/docker/connection.rb#L100 is not working? I dont think we log run_command in the other transports in the connection class, was this a parity feature or something you wanted generally for all the transports?

h0tw1r3:
> Are you saying that the trace logging with upload_file
> 
> https://github.com/puppetlabs/bolt/blob/ca25a6bed9b1ea70e8b03d88597c7ad935140dfe/lib/bolt/transport/docker/connection.rb#L100
> is not working? I dont think we log run_command in the other transports in the connection class, was this a parity feature or something you wanted generally for all the transports?

The executed docker commands are being trace logged in all cases (that I can see) except the upload command. This command is very helpful to see when tracing and reproducing upload issues.

PR #3287, opened 2024-03-30
(GH-3286) Add `file::delete()` function

### LINKED ISSUE #3286: Add `file::delete()` function
## Use Case
Sometimes I miss the `file::delete()` function to delete a file on the Bolt controller node. So far I'm using `run_command("rm '${file}'", 'localhost')`, though I wouldn't call it an elegant solution.

jay7x:
Closes #3286

!feature

* **Add `file::delete()` function** Delete a file on localhost using ruby's `File.delete`. This will only delete files on the machine you run Bolt on.

### DIFF
--- a/bolt-modules/file/lib/puppet/functions/file/delete.rb
+++ b/bolt-modules/file/lib/puppet/functions/file/delete.rb
@@ -1,0 +1,21 @@
+# frozen_string_literal: true
+
+# Delete a file on localhost using ruby's `File.delete`. This will only delete
+# files on the machine you run Bolt on.
+Puppet::Functions.create_function(:'file::delete') do
+  # @param filename Absolute path.
+  # @example Delete a file from disk
+  #   file::delete('C:/Users/me/report')
+  dispatch :delete do
+    required_param 'String[1]', :filename
+    return_type 'Undef'
+  end
+
+  def delete(filename)
+    # Send Analytics Report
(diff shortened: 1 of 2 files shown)


jay7x:
One more note wrt this.. I guess it'd be nice to add another function signature like `file::delete(String $file, Boolean $catch_errors = false)`. This is functional equivalent of `rm -f`, which is quite useful if one doesn't really want to check if the file exists before deleting it.

bolt-modules/file/lib/puppet/functions/file/delete.rb
```diff
@@ -0,0 +1,21 @@
...
+    required_param 'String[1]', :filename
+    return_type 'Undef'
+  end
+
+  def delete(filename)
+    # Send Analytics Report
+    Puppet.lookup(:bolt_executor) {}&.report_function_call(self.class.name)
+
+    File.delete(filename)
+    nil
```
bastelfreak, line 19: why `nil` here, do you always want to return `nil` and not the result of `delete()`?

jay7x: That was copied from `file::write()` actually.. In our case `delete()` will always return 1 (number of filenames passed as the arguments). Not sure how useful it is.

donoghuc:
This looks good. I'll need to look in to the integration tests and docs actions. I'll do that in a separate PR.

PR #3288, opened 2024-03-30
Add `file::with_tmpfile()` function

jay7x:
!feature

* **Add `file::with_tmpfile()` function** Create a temporary file and execute the block with the filename created as an argument. The file is deleted after the block execution. This will only operate on the machine you run Bolt on.

### DIFF
--- a/bolt-modules/file/lib/puppet/functions/file/with_tmpfile.rb
+++ b/bolt-modules/file/lib/puppet/functions/file/with_tmpfile.rb
@@ -1,0 +1,39 @@
+# frozen_string_literal: true
+
+require 'tempfile'
+
+# Create a temporary file and execute the block with the filename created as an
+# argument. The file is deleted after the block execution. This will only
+# operate on the machine you run Bolt on.
+Puppet::Functions.create_function(:'file::with_tmpfile') do
+  # @param basename Determines the name of the temporary file (see ruby's Tempfile.new)
+  # @param tmpdir Directory to place the temporary file in (ruby's Dir.tmpdir is default value)
+  # @param block The block to execute with a temporary filename
+  # @example Run a command with a temporary file
+  #   # NOTE: This will work on the Bolt controller node only!
+  #   $res = file::with_tmpfile('foo') |$filename| {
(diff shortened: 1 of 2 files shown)


jay7x:
`Tempfile.new()` has another signature allowing to pass Array[String, 2] as the basename. Not sure if it'd be useful to implement here also..

donoghuc:
Interesting. We have something similar in BoltSpec https://github.com/puppetlabs/bolt/blob/4b30c391d97ad903bdfc27fcaf0c2a910371e898/spec/lib/bolt_spec/files.rb#L7 Seems like most cases you would know what content you want in the file and only want the file around for another process to read from it. Would porting over the BoltSpec function solve your use case?

jay7x:
> Seems like most cases you would know what content you want in the file and only want the file around for another process to read from it. Would porting over the BoltSpec function solve your use case?

Consider the case:
```
$res = file::with_tmpfile('foo') |$filename| {
  run_command("do_something_with '${filename}'", 'localhost')
  upload_file($filename, $targets)
}
```

There are 2 things:
1. The file content might generated by a command which takes the filename as an argument.
2. BoltSpec function yields a block with a File object as I see. In my case it was a file name.

If we can deal with both cases I'm fine to reuse the BoltSpec code :-D

Actually I'm totally ok to not include this function into the Boltlib (`file::delete()` should be enough). I can just use my own module with when needed.

donoghuc:
I dont understand the example? What do you need with an empty file in run_command? I guess, for me the only use case I can think of for a "with temp file" you would know the content of the file (or be generating it) explicitly for consumption by something else. Therefore, why would we need a temp file to exist with no way to modify the content? 

As far as the return type etc, yeah that would be changed, i was more getting at the use case here and pointing out that we probably would want to have the argument to write content.

jay7x:
> I dont understand the example? What do you need with an empty file in run_command?

I mean that the file content may be generated by a command, taking the file name as an argument. In this case the file content is not known in advance. Though, to be honest, for now it's purely "synthetic" case. I'm not sure how many ppl might have this requirement.

PR #3292, opened 2024-04-02
Improve add_facts performance

seanmil:
When making multiple add_facts calls to add facts to an already present and large set of target facts performance declines as more are added due to the initialization and copying of the large hashes each time.

This change merges the added facts in place to the existing facts, which slightly improves performance.

### DIFF
--- a/lib/bolt/util.rb
+++ b/lib/bolt/util.rb
@@ -212,6 +212,17 @@ module Bolt
         hash1.merge(hash2, &recursive_merge)
       end
 
+      def deep_merge!(hash1, hash2)
+        recursive_merge = proc do |_key, h1, h2|
+          if h1.is_a?(Hash) && h2.is_a?(Hash)
+            h1.merge!(h2, &recursive_merge)
+          else
+            h2
+          end
+        end
+        hash1.merge!(hash2, &recursive_merge)
+      end
+
       # Accepts a Data object and returns a copy with all hash keys
       # modified by block. use &:to_s to stringify keys or &:to_sym to symbolize them
       def walk_keys(data, &block)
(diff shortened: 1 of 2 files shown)


lib/bolt/util.rb
```diff
@@ -212,6 +212,17 @@ def deep_merge(hash1, hash2)
         hash1.merge(hash2, &recursive_merge)
       end
 
+      def deep_merge!(hash1, hash2)
+        recursive_merge = proc do |_key, h1, h2|
+          if h1.is_a?(Hash) && h2.is_a?(Hash)
+            h1.merge(h2, &recursive_merge)
```
donoghuc, line 218: Should nested hashes be merged in place too?

seanmil: Ah, yes, probably. I was initially uncertain if that would make sense (or a difference) but I think nested ones should be merged in place too.

seanmil:
I will fix this up and re-push, but just so you know - the original symptom of VERY slow `add_facts` calls which sent me down this path seem to have actually been due to the Bolt analytics associated collected for that call.  In the scenario I was testing I have some 3,300 targets which I'm trying to add 6 specific facts from PuppetDB each to (the whole set of facts is unneeded and would be way too much) - which results in a little more than 20,000 add_facts calls.  It seems when analytics is enabled it just falls over at some point when processing them.  For a smaller but still large number of calls it just makes the add_facts calls start to take progressively (and significantly) more time to return.  When I added `analytics: false` to the `bolt-project.yaml` it immediately started working and the performance was quite acceptable.

I was misled in my troubleshooting because it seems that analytics is not enabled when running bolt from a Git checkout instead of from the normal package.  So, this change is probably still a good idea, but is quite incremental vs. revolutional as I had initially thought.

donoghuc:
Thanks for reporting that. I've filed a bug issue and I intend on disabling analytics submissions by default moving forward. https://github.com/puppetlabs/bolt/issues/3293

donoghuc (approved):

Issue #3293, opened 2024-04-02
Analytics collection is not feasible with some plan functions


donoghuc:
## Describe the Bug
Analytics submissions for `add_facts` calls on thousands of targets bogs down analytics submissions.

## Expected Behavior
Analytics submissions should not degrade performance.

## Additional Context
Turning off analytics submission is a good work around for existing installs. Stop collecting analytics by default.

donoghuc:
https://github.com/puppetlabs/bolt/pull/3294

PR #3294, opened 2024-04-02
(maint) Turn off analytics submissions

donoghuc:
Collecting analytics is now turned off by default.

!removal

* **Stop collecting bolt analytics** ([#3293](#3293))

  Bolt no longer collects analytics by default.

### DIFF
--- a/lib/bolt/shell/powershell/snippets.rb
+++ b/lib/bolt/shell/powershell/snippets.rb
@@ -55,21 +55,25 @@ module Bolt
             }
             #{build_arg_list}
 
-            switch -regex ( Get-ExecutionPolicy )
+            try
             {
-              '^AllSigned'
+              switch -regex ( Get-ExecutionPolicy )
               {
-                if ((Get-AuthenticodeSignature -File "#{script_path}").Status -ne 'Valid') {
-                  $Host.UI.WriteErrorLine("Error: Target host Powershell ExecutionPolicy is set to ${_} and script '#{script_path}' does not contain a valid signature.")
+                '^AllSigned'
+                {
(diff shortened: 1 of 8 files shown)


tlehman (approved):
LGTM

lib/bolt/shell/powershell/snippets.rb
```diff
@@ -55,21 +55,25 @@ def run_script(arguments, script_path)
             }
             #{build_arg_list}
 
-            switch -regex ( Get-ExecutionPolicy )
+            try
```
donoghuc, line 58: @mcdonaldseanp I think the path is messed up for module loading in our test runners:

```
       +"merged_output" => "Get-ExecutionPolicy : The 'Get-ExecutionPolicy' command was found in the module 'Microsoft.PowerShell.Security', but \r\nthe module could not be loaded. For more information, run 'Import-Module Microsoft.PowerShell.Security'.\r\nAt line:10 char:17\r\n+ switch -regex ( Get-ExecutionPolicy )\r\n+                 ~~~~~~~~~~~~~~~~~~~\r\n    + CategoryInfo          : ObjectNotFound: (Get-ExecutionPolicy:String) [], CommandNotFoundException\r\n    + FullyQualifiedErrorId : CouldNotAutoloadMatchingModule\r\n \r\nparam\n",
```
I think that the intent of this was simply to provide a helpful error message and that we should not error if we cant load that module. In practice i doubt this comes up much because task running all works due to other powershell setup we do. 

Do you think a `catch` here is reasonable, should it be more specific on objects it catches?

donoghuc: https://github.com/puppetlabs/bolt/issues/2659 for reference

Issue #3296, opened 2024-04-08
Puppetdb Client *ALWAYS* uses token header even when cert/key is configured


donoghuc:
## Describe the Bug
In the case a `token` AND cert based auth methods are configured for puppetdb, the client always includes the token in the header for puppetdb requests. This results in the case where puppetdb queries fail even when a valid cert/key pair is configured.

## Expected Behavior
Puppetdb client should be able to use *either* token *or* cert/key auth. 

Options:
1. Client could have cert/key take precedence over token
2. Client could try with token, on 401/403 response could retry with cert/key

donoghuc:
I think what is a bit confounding is that if cert/key are invalid it will fail regardless of the token. If this were net-new I would just choose to prefer cert based auth when both are specified (IE *only* include token if cert/key are not specified). But in order to not break anybody relying on token being included in header even if certs are configured I wanted to not disturb that (specifically thinking of the case where you would prefer the identity piece of an rbac token if possible). I bring this up in response to `2`, specifically i think that the configuration will get a bit out of control with complexity if we try to use token only even if cert/key are incorrectly configured. We are not currently at risk of breaking anybody with a change today in this regard (if you have bad certs configured, regardless of your token config its not going to connect).

donoghuc:
I'm tempted to simplify this to follow the pattern in the puppetdb CLI. Specifically, if cert is configured, do not use token. I think that really simplifies things and is easier to understand. It also probably helps as a forcing function to ensure there is not ambiguous config files in practice. I can put up a separate PR with that approach and make sure to add a log message to warn that token wont be used when cert based auth is configured.

PR #3299, opened 2024-04-11
(GH-3296) Prefer cert auth to token auth for puppetdb client

donoghuc:
Previously regardless of using certs any puppetdb token (either read from default location OR configured in settings) would be sent in x-authentication header for puppetdb requests. In the case a cert is configured, do not include this as the puppetdb endpoint will 401 in the case a valid cert but revoked token is presented.


tlehman (approved):
LGTM

PR #3300, opened 2024-04-12
(maint) Update packaged modules

donoghuc:
Update to latest modules for next bolt release.

!feature

* **Update puppet modules shipped with bolt packages**

  Modules shipped with bolt packages have been updated to latest.


donoghuc:
latest puppet_agent module drops support for puppet6. I think that is fine, will need to do some test cleanup.

tlehman (approved):

PR #3304, opened 2024-04-18
(maint) Fix ctrl::do_until docs

mcdonaldseanp:
The docs mischaracterized the return value of ctrl::do_until as `nil`. That function returns the value of the given code block's last iteration


donoghuc:
You might need a `!no-release-note` in commit message to make docs happy.

donoghuc (approved):

PR #3312, opened 2024-05-09
r10k: Allow >= 3.10 < 5

bastelfreak:
from the commit message:

    This also updates the orchestrator_client dependency. It requires
    faraday 1.x. Older versions of orchestrator_client also depend on
    faraday but didn't specify the version requirement. r10k 4 depends on
    a newer puppet_forge gem which in turn depends on faraday 2.
    If we only bump r10k we will downgrade orchestrator_client from 0.7.0 to
    0.5.3, which we probably don't want.

    Result => orchestrator_client needs to be updated to work with faraday
    2. (which is actually implemented but unreleased).

This also contains https://github.com/puppetlabs/bolt/pull/3311
This needs a release of https://github.com/puppetlabs/orchestrator_client-ruby/pull/38


donoghuc:
Thanks, right now we are looking towards packaging bolt with ruby 3. We already run bolt lib in PE for both streams (ruby 2.7 and ruby 3). Now, the bolt packages are built for ruby 2/ puppet 7. Therefore gems like r10k etc that are not really used in bolt's PE runtime are not yet updated. Re-factoring all the gem enviornments to satisfy both PE streams and bolt packages will have some implications outside of just bolt. We understand your desire for this change and I think we can get to a place where this will work, but at the moment its going to take getting prioritized from Perforce so we can account for the gem packaging etc we will likely need to do for this.

donoghuc:
Bolt packages going to ruby 3 / puppet 8 is for sure on the roadmap, the exact timeline is not yet finalized. Certainly before 2023 becomes new LTS

bastelfreak:
Would it be possible to move bolt to Ruby 3.3 before it enters the new LTS? Otherwise we probably again end up in a situation where the LTS depends on a Ruby version that's EoL upstream. Related issue: https://github.com/puppetlabs/puppet-enterprise_issues/issues/8

donoghuc:
Bolt will use the exact same ruby as puppet 8 agent. Whatever the resolution of that issue you linked is will apply to bolt.

alex501212 (approved):
@bastelfreak Approving this ahead of the Bolt v4 release

Issue #3309, opened 2024-05-08
Local transport defaults are not applied when using apply_prep 


donoghuc:
## Describe the Bug
When using the local transport for a target the default behavior is described as using the ruby env shipped with bolt and the puppet-agent feature is set. This only happend on `with_connection`. When the first action a plan does on a target is `apply_prep` the targets who need puppet_agents is decided *before* any actions are run on that target https://github.com/puppetlabs/bolt/blob/f52313e4522be19dbe1662a0c926595c4f43fed5/bolt-modules/boltlib/lib/puppet/functions/apply_prep.rb#L98

Therefore these two plans have different results on the local transport:

```
plan foo(TargetSpec $targets){
  run_command(whoami, $targets)
  apply_prep($targets)
}
```

```
plan foo(TargetSpec $targets){
  apply_prep($targets)
}
```

The first plan correctly assigns the `puppet-agent` feature to the  resolved`$targets`, then apply prep does not attempt to install an agent. The second plan attempts to install an agent on the target. 

## Expected Behavior
Defaults should be assigned when targets are resolved so that the expected state is available whenever the target is referenced.

PR #3313, opened 2024-05-09
(GH-3309) Ensure bundled-ruby defaults are applied at target init

donoghuc:
Previously defaults were not applied until an action was actually requested on a target. This made inventory resolution confusing and non-determanistic behavior with apply_prep. This commit moves application of defaults for a target to the init method for creating a target object.

!bug

* **Apply bundled-ruby defaults at target creation** ([#3309](3309))

  Ensure defaults associated with `bundled-ruby` transport config are applied at target initialization.

### DIFF
--- a/lib/bolt/inventory/target.rb
+++ b/lib/bolt/inventory/target.rb
@@ -34,7 +34,7 @@ module Bolt
           @name = @uri
           @safe_name = @uri_obj.omit(:password).to_str.sub(%r{^//}, '')
         end
-
+        # handle special localhost target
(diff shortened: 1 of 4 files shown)


mcdonaldseanp (approved):
I'm a little surprised we weren't doing this before

donoghuc:
Yeah, i think this is a subtle one that is easy to not be noticed. I think i understand the sentiment around deferring before adding the defaults, bug given it changes the state of the target we need to ensure it happens consistently.

Issue #3310, opened 2024-05-08
Core types/providers are not synced when using bundled-ruby for apply


donoghuc:
## Describe the Bug
When using bundled-ruby on the local transport core types are not pluginsynced. It looks like there used to be an issue with puppet 5. Given that is very long EOL we should be able to just pluginsync bolt's version of core types/providers. If that is a bad idea then for at least the local transport we should. https://github.com/puppetlabs/bolt/blob/f52313e4522be19dbe1662a0c926595c4f43fed5/lib/bolt/pal.rb#L227-L230

## Expected Behavior
When I run apply using the local transport I can use core types/providers.

PR #3314, opened 2024-05-10
(GH-3310) Ensure plugin code for core types is available on local transport

donoghuc:
When using the local transport defaulting to `bundled-ruby: true` it is assumed an agent is on the target. Generally when a `puppet-agent` feature is applied to a target this is assumed to be the case. This provides the core types and providers when an catalog is applied. We explicitly do not want to pluginsync all the core types and providers when we do not have to. This commit updates the apply_catalog task to take advantage of the fact that bolt packages have the core types in them. This works by sending as a task parameter the install location of bundled module content in bolt packages. After unpacking the plugins sent to the apply_catalog task, if the task finds any modules shipped with bolt, it will add those to the `LOAD_PATH`. This allows us to continue to avoid excessive pluginsync and also support the bundled-ruby config over the local transport.

### DIFF
--- a/lib/bolt/applicator.rb
+++ b/lib/bolt/applicator.rb
@@ -284,6 +284,9 @@ module Bolt
                   'catalog' => Puppet::Pops::Types::PSensitiveType::Sensitive.new(catalog),
                   'plugins' => Puppet::Pops::Types::PSensitiveType::Sensitive.new(plugins),
                   'apply_settings' => @apply_settings,
+                  # This should just be boltlib and modules dirs shipped with bolt packages
+                  # The apply_catalog task uses them to load core types if they exist
+                  'bolt_builtin_content' => @modulepath - @plugin_dirs,
                   '_task' => catalog_apply_task.name,
                   '_noop' => options[:noop]
                 }
--- a/libexec/apply_catalog.rb
+++ b/libexec/apply_catalog.rb
@@ -52,6 +52,19 @@ begin
     $LOAD_PATH << dir unless $LOAD_PATH.include?(dir)
   end
 
+  # In the case we are applying on a bolt runner and using bundled-ruby over local transport
+  # we will want to load code shipped with bolt. This is last on the load path and therefore
+  # explicitly packaged plugins should take precedence
+  args['bolt_builtin_content'].each do |builtin_dir|
+    next unless Dir.exist?(builtin_dir)
+    Dir.foreach(builtin_dir) do |dir|
+      unless ['.', '..'].include? dir
+        full_path = File.join(builtin_dir, dir, 'lib')
+        $LOAD_PATH << full_path unless $LOAD_PATH.include?(full_path)
+      end
+    end
+  end
+
   if (conn_info = args['_target'])
     unless (type = conn_info['remote-transport'])
       puts "Cannot execute a catalog for a remote target without knowing it's the remote-transport type."


donoghuc:
@joshcooper I think the concern here would be if bolt were syncing over core types for the version it is based off (in our case puppet 7) to an agent where those would be different. It looks like modern agents (7 and 8) are pretty similar for core types. I was thinking this will be the simplest solution. Do you think we risk diverging core types between agents?

mcdonaldseanp (approved):

mcdonaldseanp:
@donoghuc I think we're probably pretty safe with the core types, I can't imagine we ever want to change them much at all. If an issue came up the solution is trivial anyway, just install a compatible agent version on the target.

donoghuc:
Looks like the apply integration tests are failing due to the large plugin tarball being passed to the fake apply task. Going to see about how best to address that in the tests. Throwing a DNM on it for now.

donoghuc:
Ideally we could both compile and apply catalogs based on the plugins shipped in bolt's package for local transport. For apply, we prepare a single plugin tarball regardless of transports and apply it. There is no established or clear way to switch behavior based on whether or not we are running on the local transport. 

The local transport aside, I'm thinking perhaps it may be best anyway to have the plugins from bolt's package in the plugin tarball so that agents (regardless of transport) would use the same code as the catalog was compiled against.

lib/bolt/applicator.rb
```diff
@@ -284,6 +284,7 @@ def apply_ast(raw_ast, targets, options, plan_vars = {})
                   'catalog' => Puppet::Pops::Types::PSensitiveType::Sensitive.new(catalog),
                   'plugins' => Puppet::Pops::Types::PSensitiveType::Sensitive.new(plugins),
                   'apply_settings' => @apply_settings,
+                  'bolt_builtin_content' => @modulepath.full_modulepath - @modulepath.user_modulepath || [],
```
mcdonaldseanp, line 287: Instead of doing this calculation, we should add a new function to https://github.com/puppetlabs/bolt/blob/main/lib/bolt/config/modulepath.rb and document what it is there

mcdonaldseanp (changes requested):
Just one small thing

lib/bolt/applicator.rb
donoghuc: YEah, i will have to do a significant amount of cleanup i think around all the mocking etc in tests. Just wanted to show this approach (and test it).

mcdonaldseanp: Ah, I see.

Yeah this approach is a good way to go

donoghuc:
After some discussion on internal slack i think the best path forward for now is to go with the approach in https://github.com/puppetlabs/bolt/pull/3314/commits/a094b30bc7926d5e67bcc7be04d6a3e25e5393a3 as it avoids having to do any additional pluginsync and should be a cheap and effective way to ensure core types will be available during catalog application over the local transport with the default `bundled-ruby` config.

lib/bolt/applicator.rb
donoghuc: Actually, it looks like we just pass in resolved arrays into applicator anyway. I have tested this change on a fresh vmpooler image and it appears to work. Looks like our tests dont blow up with an extra task parameter so i'm just going to leave the comment and call it good.

mcdonaldseanp (approved):

PR #3315, opened 2024-05-10
Bump minimal ruby 2.5->2.7

bastelfreak:


donoghuc:
Is the min ruby version preventing you from doing something? I was thinking of moving it to ruby 2.7 when we start packaging bolt with ruby 3 and puppet 8. Its not clear now why we would drop 2.5 compatibility.

bastelfreak:
The problem is the dependency resolution. Ruby 2.4 and older didn't properly honour the `required_ruby_version` from dependencies. This got better with Ruby 2.5. I don't know if it got fixed completely.

Problem two: The versions of the dependencies that are pulled in will depend on the ruby version. For example we could allow r10k 4 now but then bundler will pull in an older version of the orchestrator_client. And if that works or not isn't clear.

I think the minimal version being 2.5 doesn't make sense. PE 2021 is the oldest supported "user" of bolt. It uses Ruby 2.7. And Ruby 2.5, 2.6 and 2.7 are EoL. So I think we should drop Ruby 2.5 and switch to 2.7 in Bolt. Now if that requires a major or minor release in bolt is another question. I'm fine with both. But even a major release shouldn't be a problem. Make a "small" bolt 4 release that drop Ruby 2.5 & 2.6. Then focus on bolt 5. We did that on other gem in the puppet ecosystem as well.

donoghuc:
Yeah, i dont think dropping min ruby version necessarily requires a major version bump based on our use cases. The dependency resolution story is one i deeply empathize with as i've been struggling with that when using `gem install` instead of bundler for internal workflows. We do have some internal CI that runs bolt on ruby 2.5, so i cant just merge this without doing some tactical changes on the back end. I understand you want it and I for sure see a path forward for making the min version 2.7 in the near term. Thank you for the input.

PR #3318, opened 2024-05-20
(maint) Fix bug in plan_spec integration test

donoghuc:
Previously the `target` argument was not being properly resolved in the argument array resulting in the literal string `#{target}` being used instead of the resolved value.

!no-release-note


mcdonaldseanp (approved):
I fucking hate ruby

PR #3328, opened 2024-06-26
Ensure bash shell executions close file descriptors

seanmil:
Currently Bolt::Shell::Bash will leaves open file descriptors at the conclusion of execute(). The file descriptions do fall out of scope at the conclusion of the execute() method and the Ruby GC eventually closes them. However, on a very busy Bolt invocation with lots of short-lived Tasks or a number of Tasks running in parallel (e.g. via background()) the number of open FDs before garbage collection can get moderately high, hitting problems on systems with low file descriptor limits.

In some simple testing in my environment this reduced the number of consistently open FDs from around 90-110 to somewhere in the 50s for a given test scenario.

### DIFF
--- a/lib/bolt/shell/bash.rb
+++ b/lib/bolt/shell/bash.rb
@@ -470,7 +470,10 @@ module Bolt
             result_output.merged_output << to_print
           }
         rescue Errno::EAGAIN, EOFError
+        ensure
+          stream.close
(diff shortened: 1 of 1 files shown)


donoghuc (approved):
Wow, nice! thanks for finding this.

Issue #3333, opened 2024-07-01
run_command: support array of commands


bastelfreak:
## Use Case

Due to security concerns, the best practice for commands in Puppet DSL are now arrays: https://www.puppet.com/docs/puppet/7/types/exec.html#exec-attribute-command

The `run_command()` only supports the String datatype as input and I think it makes sense that it supports array as well.

## Describe the Solution You Would Like

Array support for the commands in `run_command()`

PR #3336, opened 2024-07-22
run_command: support an array of commands

alex501212:
!feature

* **run_command: support an array of commands* ([#3333](https://github.com/puppetlabs/bolt/issues/3333))

  Added run_commands method which allows an array of commands to be passed to run_command

### DIFF
--- a/bolt-modules/boltlib/lib/puppet/functions/run_command.rb
+++ b/bolt-modules/boltlib/lib/puppet/functions/run_command.rb
@@ -25,6 +25,23 @@ Puppet::Functions.create_function(:run_command) do
     return_type 'ResultSet'
   end
 
+  # Run multiple commands.
+  # @param commands Commands to run on target.
+  # @param targets A pattern identifying zero or more targets. See {get_targets} for accepted patterns.
+  # @param options A hash of additional options.
+  # @option options [Boolean] _catch_errors Whether to catch raised errors.
+  # @option options [String] _run_as User to run as using privilege escalation.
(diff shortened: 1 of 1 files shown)


donoghuc (changes requested):
This does not address the use case described in the issue. The request is to mirror the pattern in puppet's exec to be more explicit about parameterization. The example they give is: `command => ['/bin/echo', 'hello world; rm -rf /']`. Your PR here would do:
`/bin/echo && hello world; rm -rf /` which is absolutely not what we want. 

Properly implementing this feature will involve making a similar change to how we use the underlying Open3 lib to take advantage of the explicit parameterization features.

alex501212:
> This does not address the use case described in the issue. The request is to mirror the pattern in puppet's exec to be more explicit about parameterization. The example they give is: `command => ['/bin/echo', 'hello world; rm -rf /']`. Your PR here would do: `/bin/echo && hello world; rm -rf /` which is absolutely not what we want.
> 
> Properly implementing this feature will involve making a similar change to how we use the underlying Open3 lib to take advantage of the explicit parameterization features.

Wondering if it would  be acceptable splitting commands, then escaping each argument and then lastly joining arguments again into a single command?  This way in [‘/bin/echo', 'hello world; rm -rf /'] the second command won’t run at all

donoghuc:
I think you are misunderstanding the feature request. it is not to be able to run multiple commands in the same invocation of the function. Instead it is to support explicit parameters to a single command.

PR #3343, opened 2024-08-20
(PE-39083) Upgrade module versions to latest

tlehman:


donoghuc:
The changelog is auto generated from commit mesages. You shouldnt check in changes directly to it. Here is a formatted commit message i used last time, https://github.com/puppetlabs/bolt/pull/3300

donoghuc:
That script looks useful! feel free to check it in here under `dev-resources`. thanks

lib/bolt/version.rb
```diff
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
 
 module Bolt
-  VERSION = '3.30.0'
+  VERSION = '3.31.0'
```
donoghuc, line 4: This is also handled by CI. You can leave this alone.

donoghuc (changes requested):

donoghuc (approved):

PR #3352, opened 2024-10-17
Update puppet-agent module to 4.21.0

alex501212:
Updating puppetlabs-puppet_agent to 4.21.0


donoghuc (changes requested):
Thanks for the update. Can you update the commit message so our release note automation includes a callout for updated module content? See https://github.com/puppetlabs/bolt/blob/main/CONTRIBUTING.md#pull-requests

Thanks!

donoghuc (approved):

PR #3353, opened 2024-10-22
Remove puppet version pin from gemfile

alex501212:
Since PE-35920 is resolved I have removed the puppet version pin in the gemfile

### DIFF
--- a/spec/integration/lookup_spec.rb
+++ b/spec/integration/lookup_spec.rb
@@ -94,9 +94,13 @@ describe 'lookup' do
 
     context 'with interpolations' do
       let(:hiera_config) { File.join(project, 'hiera_interpolations.yaml') }
-      it 'returns a value' do
+
+      it 'returns an error' do
         result = run_cli_json(cli_command + %w[key=test::interpolations])
-        expect(result).to eq('test::interpolations data/common.yaml')
+        expect(result).to include(
+          'kind' => 'bolt/pal-error',
+          'msg'  => /Interpolations are not supported in lookups/
+        )
       end
     end
 
--- a/spec/integration/lookup_spec.rb
+++ b/spec/integration/lookup_spec.rb
@@ -94,13 +94,9 @@ describe 'lookup' do
 
     context 'with interpolations' do
       let(:hiera_config) { File.join(project, 'hiera_interpolations.yaml') }
-
-      it 'returns an error' do
+      it 'returns a value' do
         result = run_cli_json(cli_command + %w[key=test::interpolations])
-        expect(result).to include(
-          'kind' => 'bolt/pal-error',
-          'msg'  => /Interpolations are not supported in lookups/
-        )
+        expect(result).to eq('test::interpolations data/common.yaml')
       end
     end
 
--- a/spec/integration/lookup_spec.rb
+++ b/spec/integration/lookup_spec.rb
@@ -92,15 +92,12 @@ describe 'lookup' do
       end
     end
 
-    context 'with interpolations' do
+    context 'when context is not available' do
       let(:hiera_config) { File.join(project, 'hiera_interpolations.yaml') }
 
-      it 'returns an error' do
+      it 'returns the default value' do
         result = run_cli_json(cli_command + %w[key=test::interpolations])
-        expect(result).to include(
-          'kind' => 'bolt/pal-error',
-          'msg'  => /Interpolations are not supported in lookups/
-        )
+        expect(result).to eq('test::interpolations data/common.yaml')
       end
     end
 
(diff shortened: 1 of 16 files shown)


donoghuc:
Looks like while that ticket is closed it is not actually resolved. We will probably want to address this issue. Specifically we will want a readable error when an unsupported Hiera lookup is attempted.

spec/integration/lookup_spec.rb
```diff
@@ -94,13 +94,9 @@
...
       let(:hiera_config) { File.join(project, 'hiera_interpolations.yaml') }
-
-      it 'returns an error' do
+      it 'returns a value' do
         result = run_cli_json(cli_command + %w[key=test::interpolations])
-        expect(result).to include(
-          'kind' => 'bolt/pal-error',
-          'msg'  => /Interpolations are not supported in lookups/
-        )
+        expect(result).to eq('test::interpolations data/common.yaml')
```
donoghuc, line 99: We want to ensure bolt errors here. Returning a success here is a regression.

donoghuc (changes requested):

spec/integration/lookup_spec.rb
donoghuc: Sorry, I completely forgot about https://github.com/puppetlabs/bolt/pull/3213 I think you can update the test to validate that the default value is used in the lookup.

alex501212: is this similar to other tests were we add options with a default value to the run_cli_json call?

```
    it 'accepts a default value' do
      options = { 'default_value' => 'default' }.to_json
      result  = run_cli_json(cli_command + %W[key=foo::bar::baz options=#{options}])
      expect(result).to eq('default')
    end
```

donoghuc: The core of the issue here is that we wanted to make it clear to users that in plan scope (specifically *not* in an apply block) there is not going to be the context (for example `facts`) necessary to resolve a hierachy that relies on that data for resolution. Previously we relied on a behavior in puppet whereby a lookup without necessary context would result in an error. Turns out that was not necessarily the correct assumption (see https://github.com/puppetlabs/bolt/pull/3213 ). The path forward here is that a lookup attempted in plan scope (outside of an apply block) should be able to do a lookup without error using the defaults that the underlying puppet/hiera comes up with. We will need to update our tests to codify our expectations about what this behavior should be. Generally, does the test code in these fixtures https://github.com/puppetlabs/bolt/tree/main/spec/fixtures/hiera make sense to you? The idea is that we have some plan code and a bunch of hiera configurations used in the tests to validate assumptions about what lookups will resolve to.

alex501212: would https://github.com/puppetlabs/bolt/blob/main/spec/fixtures/hiera/modules/test/data/common.yaml need updated for this test? I'm not very familar with hiera so i'm wondering if theres any similar tests you could point me to?

donoghuc: The tests in this file largely cover the expectations for how lookups behave. The data itself in that file (`common.yaml`) should remain unchanged, the methods we use for doing the lookup (for example in top level plan scope) with the various configuration files we have there and the expected results will need to be refactored to ensure the defaults are expected. 

This section of our docs does a decent job of explaining the basics. https://www.puppet.com/docs/bolt/latest/hiera.html You can review that and make sure that the new behavior still tracks with the docs.

alex501212:
after running lookup_spec.rb it looks like quite alot of tests are failing when using Puppet8:

```
rspec ./spec/integration/lookup_spec.rb:98 # lookup plan function with interpolations returns an error
rspec ./spec/integration/lookup_spec.rb:193 # lookup command without --plan-hierarchy looks up a value
rspec ./spec/integration/lookup_spec.rb:238 # lookup command without --plan-hierarchy looks up a value in the module hierarchy
rspec ./spec/integration/lookup_spec.rb:246 # lookup command without --plan-hierarchy errors with a missing key
rspec ./spec/integration/lookup_spec.rb:254 # lookup command without --plan-hierarchy looks up a value with a built-in backend
rspec ./spec/integration/lookup_spec.rb:262 # lookup command without --plan-hierarchy looks up a value with a custom backend
rspec ./spec/integration/lookup_spec.rb:281 # lookup command without --plan-hierarchy looks up the same value as a plan lookup
rspec ./spec/integration/lookup_spec.rb:204 # lookup command without --plan-hierarchy with interpolations looks up a value with facts
rspec ./spec/integration/lookup_spec.rb:212 # lookup command without --plan-hierarchy with interpolations looks up a value with vars
rspec ./spec/integration/lookup_spec.rb:220 # lookup command without --plan-hierarchy with interpolations looks up a value with a trusted fact
rspec ./spec/integration/lookup_spec.rb:231 # lookup command without --plan-hierarchy with interpolations with plan vars interpolated looks up a value with plan vars interpolated
rspec ./spec/integration/lookup_spec.rb:273 # lookup command without --plan-hierarchy with a missing backend returns an error
rspec ./spec/integration/lookup_spec.rb:310 # lookup command with --plan-hierarchy with invalid plan_hierarchy raises a validation error
```

donoghuc:
Are you seeing those in CI? https://github.com/puppetlabs/bolt/actions/runs/11597999624/job/32350978455?pr=3353  I only see the three failures that are expecting error instead of using a default.

gavindidrichsen (approved):
LGTM

PR #3349, opened 2024-09-19
(maint) Fix component CI by pinning back yard

donoghuc:
See https://github.com/puppetlabs/puppet-strings/pull/401/files for similar approach.

PR #3354, opened 2024-10-31
Fix with latest version of yard

smortex:
Monkey patching libraries is a bad idea. Latest version of yard changed
some internal details, which broke bolt.

A workaround was proposed in #3349 but the root cause was not addressed.

Detect if the version of YARD we are using provide the
`YARD::Logger::Severity` class (part of the new interal API) and if so
use its constants to set the log level, otherwise fallback to the
previous way of setting the log level using a Symbol.

While here, revert the workaround that was committed in #3349 and #3350.


smortex:
Cc @donoghuc as he authored the workaround and will probably be interested in a fix.

donoghuc (approved):
Brilliant. Thanks so much @smortex

PR #3358, opened 2024-11-14
Revert "r10k: Allow >= 3.10 < 5"

alex501212:
Reverts puppetlabs/bolt#3312


steveax:
Can we have some context here about the motivation for the revert?

PR #3361, opened 2024-11-20
Update actions workflows to use ubuntu-22.04

alex501212:
Updated github actions workflows to use ubuntu-22.04 instead of ubuntu-latest, ubuntu-22.04 being the latest stable version of ubuntu for Bolt


gavindidrichsen (approved):
Thanks!

PR #3377, opened 2025-02-25
(CAT-2237) - Update Rubocop

danadoherty639:
This PR updates to Rubcop to 1.72.2 and fixes any errors


danadoherty639:
> Huge diffs in a single commit are hard to review. It would be nicer if you broke down each cop (or groups of cops) into commits.

yeah I didn't realise there was so many, will sort this out tomorrow.

PR #3402, opened 2026-04-05
fix: enable WinRM retry for transient WSMAN errors

### LINKED ISSUE #3385: Powershell execution fails sometimes with "WSMAN ERROR CODE: 1018" because of switched off retries
## Describe the Bug
Our powershell transport based bolt plans and tasks sometimes fail with  "WSMAN ERROR CODE: 1018". This seems to be caused by a temporary problem in the windows system. Unfortunately we were not able to find the cause nor a workaround.
The used ruby module WinRb/WinRM in bolt provides an automatic retry in this case: The shell call is embedded in "retryable". The default is 3 retries with waiting time 10 seconds (exception class: WinRM::WinRMWSManFault).
But in the bolt code
https://github.com/puppetlabs/bolt/blob/main/lib/bolt/transport/winrm/connection.rb#L45
the retries are explicitly switched off by setting the retry_limit to 1. The option value given here is valid for the whole connection.
After increasing the retry_limit to a higher value, we don't see any WSMAN errors any more.
What is the reason why the retry is explicitely switched off?

## Expected Behavior
Bolt plans and jobs must not fail because of temporary WSMAN problems in the windows system.

## Steps to Reproduce
As we did not find out the cause for the WSMAN problems in windows os, we cannot give any advice to forcing them.

## Environment
 - bolt-4.0.0
 - Windows Server 2022

raman1236:
### Description

Enable WinRM retries for transient WSMAN errors by increasing the retry_limit from 1 to 3.

### Problem

The WinRM connection in Bolt explicitly sets `retry_limit: 1`, which effectively disables the automatic retry mechanism provided by the WinRM gem. When a transient WSMAN error occurs (e.g. ERROR CODE: 1018), the connection fails immediately instead of retrying.

The WinRM gem's default retry behavior (3 retries with 10 second delays) is specifically designed to handle these transient errors, but Bolt overrides it.

### Fix

Change `retry_limit` from `1` to `3` and add `retry_delay: 10` to match the WinRM gem defaults. This allows the connection to recover from transient WSMAN errors without failing the entire task or plan.

Fixes #3385

### DIFF
--- a/lib/bolt/transport/winrm/connection.rb
+++ b/lib/bolt/transport/winrm/connection.rb
@@ -42,7 +42,8 @@ module Bolt
                       # https://github.com/WinRb/WinRM/issues/270
                       user: target.options['realm'] ? 'dummy' : @user,
                       password: target.options['realm'] ? 'dummy' : target.password,
-                      retry_limit: 1,
+                      retry_limit: 3,
+                      retry_delay: 10,
                       transport: transport,
                       basic_auth_only: target.options['basic-auth-only'],
                       ca_trust_path: cacert,


raman1236:
Friendly ping — this PR enables WinRM retry for transient WSMAN errors, improving reliability of remote task execution on Windows targets. Would appreciate a review when convenient. Thanks!

